PR #295, opened 2018-07-06
[FEATURE] Adds configurator interface for algorithms.

rrahn:
Blocked by #361 

Adds a configurator interface for a generic and unified configuration for algorithms.
The following code will be added:

* a generic configurator class
  * supports tuple interface
  * provides a meta-function to alter the composition of configs contained in the configuration.
  * supports pipeable interface.
  * supports apply_deferred_config invocation [used for runtime parameter to static type conversion]
* configurator concepts
* config_base class: configurations must implement this crtp-base class.
* deferred_config_base class: dynamic configurations must implement this crtp-base class.
* most of the docu.
* tests for all basic features.

### TODO
 - [x] Complete and fix documentation
 - [x] move configurator to detail namespace - the user won't directly interface with the configurator
 - [x] fix crash in execution of invoke deferred_config 

@cpockrandt, @h-2 please have a first look over this PR. I am mostly done. There are still some minor todos. Mostly it has to deal with some docu stuff that doesn't work as I'd like it.

I will clean the commits after I am done with the rest of the code. Right now this is still work in progress.


rrahn:
@h-2 ready to be merged

include/seqan3/core/algorithm/configurator.hpp
```diff
@@ -0,0 +1,596 @@
...
+    configurator_base & operator=(configurator_base &&)      = default;
+    ~configurator_base()                                     = default;
+
+    template <typename ... other_configs_t>
+        requires (common_reference_concept<configurator<other_configs_t ...>, remaining_configs_t> && ...)
+    constexpr explicit configurator_base(configurator<other_configs_t ...> const & cfg) :
+        remaining_configs_t{cfg}...
+    {}
+};
+//!\endcond
```
h-2: Did you consider just implementing this as a std::tuple (or inheriting from std::tuple) and using the various std utilities like tuple_cat to tie and untie the config objects?

include/seqan3/core/algorithm/concept_pre.hpp
```diff
@@ -0,0 +1,244 @@
...
+    { std::move(cfg).data() };
+    { std::move(cfg_c).data() };
+
+    requires std::is_lvalue_reference_v<decltype(cfg.data())>;
+    requires std::is_lvalue_reference_v<decltype(cfg_c.data())> &&
+             std::is_const_v<std::remove_reference_t<decltype(cfg_c.data())>>;
+    requires std::is_rvalue_reference_v<decltype(std::move(cfg).data())>;
+    requires std::is_rvalue_reference_v<decltype(std::move(cfg_c).data())> &&
+             std::is_const_v<std::remove_reference_t<decltype(std::move(cfg_c).data())>>;
+};
```
h-2: You are aware that this is satisfied by mayn opther types probably, too?

include/seqan3/core/algorithm/configurator.hpp
```diff
@@ -0,0 +1,596 @@
...
+ *
+ * ```cpp
+ * auto bar_state = get<bar>(my_cfg);
+ * ```
+ *
+ * In addition every algorithm specific config implementation can overload the get-interface with a enum based
+ * get-interface, such that the correct config is extracted based on some type-independent enum value.
+ * This is necessary for the algorithm implementation, as the algorithm neither knows the order nor the exact type of
+ * all configs (the config type could be a template class).
+ */
```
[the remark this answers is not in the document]
rrahn: a) `configurator`, because it indicates some handling of stored configurations. Though, for the name of the entities it could be renamed to config_element. That is fine with me.
b) I like the design idea and will change this.
c) This is one artefact of the current implementation. At some places we want to constraint the generic type to be a configurator and therefor I first introduced it as a private concept. But giving it more thoughts, it is nothing that other libraries need to interface with really as it is only needed for our specific algorithms and thus does not need a full fledged concept definition. 
I will clean this up and use a type_trait MF instead to check if it is the correct type.

include/seqan3/core/algorithm/deferred_config_base.hpp
```diff
@@ -0,0 +1,153 @@
...
+ * seqan3::detail::config_access struct. This class, then gives access to the underlying data via getter functions.
+ * For a dynamic dispatching of configurations, that should be translated to a static configuration for the target algorithm,
+ * the `invoke` function is triggered by the configuration system before the algorithm is executed. In this function, as
+ * can be seen above, the runtime parameter can be translated to a static config and then calls the passed callable `fn`, with
+ * the new configurator object. Please make sure, you pass the old configurator to the new one, such that the other configs are
+ * correctly passed to the new configurator.
+ * One can use the helper meta-function seqan3::detail::replace_config_with to easily create a new configurator with the deferred
+ * config replaced with the static one.
+ *
+ * \see seqan3::detail::config_base
```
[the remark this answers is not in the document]
rrahn: Yes and good point.

include/seqan3/core/algorithm/concept_pre.hpp
rrahn: After some cleanup I will have a more specific concept for this.
I needed a first feasibility draft that we can discuss before I invest to much time changing things afterwards.

include/seqan3/core/algorithm/configurator.hpp
```diff
@@ -0,0 +1,596 @@
...
+        remaining_configs_t{cfg}...
+    {}
+};
+//!\endcond
+
+// ----------------------------------------------------------------------------
+// configurator
+// ----------------------------------------------------------------------------
+
+/*!\brief Collection of configurations objects used to specify the run behavior of algorithms.
```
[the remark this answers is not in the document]
rrahn: 🏃

include/seqan3/core/algorithm/configurator.hpp
rrahn: At first I wanted to inherit the member functions of the config elements so that I can use the assembled configurator and call on this instance the member functions to set certain values. 
Consider that after initialisation some parameters might be changed. Like running the same config, but adapting the values of the band. But then I switched  to the `get` interface and afterwards I also thought to rather change it to the tuple implementation.

rrahn:
blocked by #311

h-2:
> configurator, because it indicates some handling of stored configurations. Though, for the name of the entities it could be renamed to config_element. That is fine with me.

IMHO "configurator" is a defined term that describes an entity that is external to the actual entity, but "configures" that other entity, e.g. "a setup wizard" or something like that. See also
https://en.wiktionary.org/wiki/configurator

"Configuration" on the other hand is exactly a composition/arrangement of multiple elements that together define the behaviour of something...

BUT I don't want to bikeshed this too much, it's your module :)

rrahn:
I gave it a thought, and I must agree. So I will change it.

rrahn:
Just rebased on master.

test/unit/core/algorithm/configuration_test.cpp
```diff
@@ -0,0 +1,576 @@
...
+
+    { // const r-value
+        detail::configuration<bax, bar> const cfg_rc{cfg};
+        EXPECT_EQ(std::get<bar>(std::move(cfg_rc)), 3);
+
+        EXPECT_TRUE((std::is_same_v<decltype(std::get<bar>(std::move(cfg_rc))), int const &&>));
+    }
+}
+
+TEST(configuration, construction_w_lvalue_cfg)
```
marehr: `_with_`?

test/unit/core/algorithm/deferred_config_element_base_test.cpp
```diff
@@ -0,0 +1,146 @@
...
+    friend class detail::config_element_access<bar>;
+
+    template <typename fn_t, typename configuration_t>
+    constexpr auto invoke(fn_t && fn, configuration_t && config) const
+    {
+        using new_cfg_t = detail::replace_config_with_t<std::remove_reference_t<configuration_t>,
+                                                        bar,
+                                                        bar_static<1>>;
+        return fn(new_cfg_t{config});
+    }
```
marehr: I'm a bit confused. The example above was not working so I don't know if you wanted to have here multiple static types depending on some runtime parameters?

rrahn:
@h-2 @marehr  I think I addressed all issues now.

include/seqan3/core/algorithm/deferred_config_element_base.hpp
```diff
@@ -0,0 +1,150 @@
...
+ * template <size_t I>
+ * class my_config : detail::config_element_base<my_config<I>>
+ * {
+ *     // Grant the base class access to the private member `value`.
+ *     friend class detail::config_element_access<my_config<I>>;
+ *
+ *     size_t value{I};  // Has to be named `value`.
+ * };
+ *
+ * class my_deferred_config = detail::deferred_config_element_base<my_deferred_config>
```
h-2: `:` instead of `=` ?

include/seqan3/core/algorithm/config_element_base.hpp
```diff
@@ -0,0 +1,152 @@
...
+    constexpr auto && get() && noexcept
+    {
+        return config_element_access<derived_t>::_get(std::move(static_cast<derived_t &&>(*this)));
+    }
+
+    //!\copydoc get()
+    constexpr auto const && get() const && noexcept
+    {
+        return config_element_access<derived_t>::_get(std::move(static_cast<derived_t const &&>(*this)));
+    }
```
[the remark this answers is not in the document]
rrahn: as discussed, I will adapt the configs to simple aggregate types.

include/seqan3/core/algorithm/configuration.hpp
```diff
@@ -0,0 +1,778 @@
...
+    //!\endcond
+    constexpr configuration(cfg_fn_t && cfg_fn) :
+        configuration{std::invoke(std::forward<cfg_fn_t>(cfg_fn), configuration<>{})}
+    {}
+    //!\}
+
+    //!\cond
+    template <size_t elem_no, typename ... _configs_t>
+        requires (std::is_same_v<configs_t, _configs_t> && ...)
+    friend constexpr auto & seqan3::get(configuration<_configs_t...> & cfg) noexcept;
```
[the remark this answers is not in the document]
rrahn: Because it is in a different namespace, and the only way to befriend them is to specify the namespace they are defined in.

include/seqan3/core/algorithm/configuration.hpp
```diff
@@ -0,0 +1,778 @@
...
+     */
+    template <typename configuration_t, std::size_t ... Is>
+    auto explode(configuration_t && cfg, std::index_sequence<Is...> const & /*unsued*/) &&
+    {
+        // Move out the elements from cache.
+        return derived_t{}(std::forward<configuration_t>(cfg), std::move(std::get<Is>(args_cache))...);
+    }
+
+    template <typename configuration_t, std::size_t ... Is>
+    auto explode(configuration_t && cfg, std::index_sequence<Is...> const & /*unsued*/) &
```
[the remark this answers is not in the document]
rrahn: I would rather ask why `unsued`? 🤣

include/seqan3/core/algorithm/configuration.hpp
```diff
@@ -0,0 +1,778 @@
...
+ * Thread-safe as long as the referenced data is not modified.
+ */
+template <size_t elem_no, typename ... configs_t>
+constexpr auto & get(detail::configuration<configs_t...> & cfg) noexcept
+{
+    using type_list_type = detail::tuple_type_list_t<typename detail::configuration<configs_t...>::base_type>;
+
+    static_assert(elem_no < meta::size<type_list_type>::value,
+                  "Index out of range.");
+    return std::get<elem_no>(cfg).get();
```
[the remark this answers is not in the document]
rrahn: as discussed.

rrahn:
@h-2 I adressed all your issues

include/seqan3/core/algorithm/concept.hpp
```diff
@@ -0,0 +1,149 @@
...
+ */
+/*!\var value
+ * \memberof seqan3::detail::config_element_concept
+ * \brief Member storing the configuration value.
+ */
+//!\}
+//!\cond
+template <typename config_t>
+concept bool config_element_concept = semi_regular_concept<std::remove_reference_t<config_t>> &&
+                                      assignable_concept<std::remove_reference_t<config_t> &, config_t> &&
```
h-2: I think assignable_concept with one parameter already does what you want, or not?

include/seqan3/core/algorithm/configuration.hpp
```diff
@@ -0,0 +1,778 @@
...
+    //!\copydoc seqan3::get<elem_no>(seqan3::detail::configuration)
+    template <size_t elem_no, typename ... _configs_t>
+        requires (std::is_same_v<configs_t, _configs_t> && ...)
+    friend constexpr auto && seqan3::get(configuration<_configs_t...> &&) noexcept;
+
+    //!\copydoc seqan3::get<elem_no>(seqan3::detail::configuration)
+    template <size_t elem_no, typename ... _configs_t>
+        requires (std::is_same_v<configs_t, _configs_t> && ...)
+    friend constexpr auto const && seqan3::get(configuration<_configs_t...> const &&) noexcept;
+    //!\endcond
```
h-2: This block is unconditionally excluded from doxygen, but you still have copydocs inside :thinking:

include/seqan3/core/algorithm/configuration.hpp
```diff
@@ -0,0 +1,778 @@
...
+        return detail::configuration{std::tuple_cat(std::move(prefix),
+                                                    std::tuple{std::forward<new_config_element_t>(new_element)},
+                                                    tuple_pop_front(std::move(remainder)))};
+    }
+
+    //!\copydoc replace_with
+    template <config_element_concept old_config_element_t,
+              config_element_concept new_config_element_t>
+    constexpr auto replace_with(old_config_element_t const & SEQAN3_DOXYGEN_ONLY(old_element),
+                                new_config_element_t && new_element) const &&
```
h-2: here also `&&` instead of `const &&`

include/seqan3/core/algorithm/configuration.hpp
```diff
@@ -0,0 +1,778 @@
...
+ */
+template <typename proxy_fn_t,
+          typename rhs_fn_t>
+constexpr auto operator|(proxy_fn_t && proxy_fn,
+                         rhs_fn_t && rhs_fn)
+//!\cond
+    requires is_type_specialisation_of_v<remove_cvref_t<proxy_fn_t>, configuration_fn_proxy> &&
+             (is_type_specialisation_of_v<remove_cvref_t<rhs_fn_t>, configuration_fn_proxy> ||
+             (std::is_base_of_v<configuration_fn_base<remove_cvref_t<rhs_fn_t>>, remove_cvref_t<rhs_fn_t>> &&
+              !std::is_same_v<remove_cvref_t<proxy_fn_t>, remove_cvref_t<rhs_fn_t>>))
```
h-2: these requirements are quite hard to parse for humans, can they maybe be simplified by a concept somewhere?

include/seqan3/core/algorithm/configuration.hpp
```diff
@@ -0,0 +1,778 @@
...
+     */
+    template <typename configuration_t, std::size_t ... Is>
+    auto explode(configuration_t && cfg, std::index_sequence<Is...> /*unused*/) &&
+    {
+        // Move out the elements from cache.
+        return derived_t{}(std::forward<configuration_t>(cfg), std::move(std::get<Is>(args_cache))...);
+    }
+
+    template <typename configuration_t, std::size_t ... Is>
+    auto explode(configuration_t && cfg, std::index_sequence<Is...> /*unused*/) &
```
h-2: const & qualified?

include/seqan3/core/algorithm/configuration.hpp
```diff
@@ -0,0 +1,778 @@
...
+{
+    using type_list_type = detail::tuple_type_list_t<typename detail::configuration<configs_t...>::base_type>;
+
+    static_assert(elem_no < meta::size<type_list_type>::value,
+                  "Index out of range.");
+
+    // TODO: It is unclear why return must be wrapped in std::move here.
+    return std::move(std::get<elem_no>(std::move(cfg)).value);
+}
+} // namespace seqan3
```
h-2: So if I understand this correctly, `std::get<0>` gives my the config_element and `seqan3::get<0>` gives me the cargo?

rrahn: I don’t like this but yes. Alternatively, the get could be a member function of the configuration and the global get is never overloaded. Because the deferred configs shouldn’t be accessed  directly. For the algorithm you want to check if a specific config is set but you are only interested in the cargo. Got alignment  we have a specific enum based get interface that I can only offer as global function.

include/seqan3/core/algorithm/configuration.hpp
rrahn: Maybe

rrahn: ok i removed it and joined both operator| overloads.

rrahn:
Open TODOs?

PR #297, opened 2018-07-09
[FEATURE] view::take_line

h-2:
This version respects the requirement for `it == sen` to be equality preserving (where `it` is of the iterator type and `sen` is of the sentinel type).

I have further added a generic iterator intermediate layer to ease writing iterators.

This view seems to work with any other view, except view::reverse, but I don't think this is our fault: ericniebler/range-v3#860


h-2:
this is blocked by #303

h-2:
TODO the views could be made const-iterable.

include/seqan3/range/view/take_line.hpp
```diff
@@ -0,0 +1,431 @@
...
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+// DAMAGE.
+//
+// ============================================================================
+
+/*!\file
+ * \author Hannes Hauswedell <hannes.hauswedell AT fu-berlin.de>
+ * \brief Provides seqan3::view::take_line.
```
h-2: and seqan3::view::take_line_or_throw

include/seqan3/range/view/take_line.hpp
```diff
@@ -0,0 +1,431 @@
...
+         */
+        iterator_type() = default;
+        constexpr iterator_type(iterator_type const & rhs) = default;
+        constexpr iterator_type(iterator_type && rhs) = default;
+        constexpr iterator_type & operator=(iterator_type const & rhs) = default;
+        constexpr iterator_type & operator=(iterator_type && rhs) = default;
+        ~iterator_type() = default;
+
+        //!\brief Constructor that delegates to the CRTP layer.
+        iterator_type(base_base_tit) : base_t_{it} {}
```
h-2: `base_base_t it`

include/seqan3/range/view/take_line.hpp
```diff
@@ -0,0 +1,431 @@
...
+     * Constant.
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    iterator begin() noexcept
+    {
+        return {ranges::begin(urange)};
+    }
```
h-2: add const versions

include/seqan3/range/view/take_line.hpp
```diff
@@ -0,0 +1,431 @@
...
+     *
+     * No-throw guarantee.
+     */
+    sentinel_type end() noexcept
+    {
+        return {ranges::end(urange)};
+    }
+    //!\}
+
+    /*!\brief Convert this view into a container implicitly.
```
h-2: document the template parameter and constraints

h-2:
@cpockrandt @joergi-w Can you have a look at this? It has high priority, because it's very useful for most of the IO.

h-2:
(I know some tests are still failing, I am investigating it)

test/unit/range/detail/inherited_iterator_base_test.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+#include <seqan3/range/detail/inherited_iterator_base.hpp>
+
+/* This class is extensively tested by the many views that use it, e.g. view::take_line */
+
+using namespace seqan3;
+
+//! [inherited_iterator_base def]
+
+class skip_uneven_it : public detail::inherited_iterator_base<skip_uneven_it, std::vector<int>::iterator>
```
joergi-w: Do you mean odd numbers (mod 2 = 1) when you use uneven? Then consider whether you want to rename it `skip_odd_number_it`.

include/seqan3/range/view/take_line.hpp
```diff
@@ -0,0 +1,469 @@
...
+        {
+            return static_cast<base_base_t>(*this) != static_cast<base_base_t>(rhs);
+        }
+
+        friend operator!=(sentinel_type const & lhs, iterator_type const & rhs) noexcept(!require_eol)
+        {
+            return rhs != lhs;
+        }
+        //!\}
+    };
```
joergi-w: comment that this is the end of the inner class (I was wondering why there is `public:` after `public:`)

include/seqan3/range/view/take_line.hpp
```diff
@@ -0,0 +1,469 @@
...
+//!\relates seqan3::detail::view_take_line
+template <typename urng_t, bool require_eol = false>
+view_take_line(urng_t) -> view_take_line<std::remove_reference_t<urng_t>, require_eol>;
+
+// ============================================================================
+//  take_line_fn (adaptor definition)
+// ============================================================================
+
+/*!\brief View adaptor definition for view::take_line and view::take_line_or_throw.
+ * \tparam require_eol Whether to throw or not.
```
joergi-w: whether to throw **if no eol characters are found before end of file**

include/seqan3/range/detail/inherited_iterator_base.hpp
```diff
@@ -0,0 +1,289 @@
...
+    {
+        derived_t cpy{*this_derived()};
+        return cpy -= skip;
+    }
+
+    //!\brief Non-member operator- delegates to non-friend operator-.
+    constexpr friend derived_t operator-(difference_type const skip, derived_t const & it) noexcept(noexcept(base_t{} -= skip))
+        requires random_access_iterator_concept<base_t>
+    {
+        return it - skip;
```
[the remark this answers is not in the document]
h-2: We don't have to, but it also doesnt hurt!

include/seqan3/range/view/take_line.hpp
```diff
@@ -0,0 +1,469 @@
...
+ * \ingroup view
+ *
+ * \details
+ *
+ * This adaptor returns a single line **excluding** the end-line character(s), *but moving the cursor behind them
+ * for single-pass ranges.* I.e. for all ranges that satisfy seqan3::forward_range_concept this is the same as calling
+ * ```cpp
+ * ranges::view::take_while([] (auto const & l) { return (l != '\r') && (l != '\n'); })
+ * ```
+ * but for ranges that are single pass this means that the endline is also consumed.
```
joergi-w: _single pass ranges_ might be easier to understand here instead of _ranges that are single pass_

include/seqan3/range/view/take_line.hpp
h-2: well that's what indention is for, right :laughing:

include/seqan3/range/view/take_line.hpp
```diff
@@ -0,0 +1,469 @@
...
+    }
+
+    //!\copydoc begin()
+    const_iterator begin() const noexcept
+        requires const_iterable_concept<urng_t>
+    {
+        return {ranges::cbegin(urange)};
+    }
+
+    //!\copydoc begin()
```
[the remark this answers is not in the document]
h-2: Yep, it's ok to copy in this case. Also the return type actually would be the same, if defined.

include/seqan3/range/detail/inherited_iterator_base.hpp
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    //!\brief Cast this to base type.
+    base_t * this_to_base()
+    {
+        return static_cast<base_t*>(this);
+    }
+
+    //!\copydoc this_derived
+    base_t const * this_to_base() const
```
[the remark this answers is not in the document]
h-2: good catch

include/seqan3/range/view/take_line.hpp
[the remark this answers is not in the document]
cpockrandt: Oops, didn't pay attention to the actual return type :baby_chick:

h-2:
I think I have addressed all the issues and all tests pass except the false positive jenkins error.

Issue #294, opened 2018-07-06
Getting started with alphabets, implement aa20


h-2:
1. Read through the general documentation on alphabets and see if you understand what's going on: http://docs.seqan.de/seqan/3.0.0-master-user/group__alphabet.html
2. Read through the general documentation on Nucleotides (http://docs.seqan.de/seqan/3.0.0-master-user/group__nucleotide.html) and look at the code of e.g. dna4 and corresponding tests. Do you understand what's going on and why things where done in that way? If not ask / take notes!
3. Look at the AminoAcid submodule and study the aa27 class. 
4. Implement the aa20 class as a class of just the canonical amino acids without the special ones:
  * you can copy'n'paste from aa27 and change the required parts, 
  * you also need to add a amino acid concept that convers all the amino acids
  * instead of making a new set of tests for `aa20`, make a generalised test (`TYPED_TEST`) for all aminoacids, like this is currently done for all nucleotides.

h-2:
@joshuak94 this one's for you :)

PR #299, opened 2018-07-11
Add aa20.hpp and aminoacid_concept

joshuak94:
Addresses issue #294.
Adds a new 20 letter amino acid alphabet. Also adds a new aminoacid_concept file, which defines
any amino acid to have all 27 letters of the alphabet.

### DIFF
--- a/include/seqan3/alphabet/aminoacid/aa20.hpp
+++ b/include/seqan3/alphabet/aminoacid/aa20.hpp
@@ -1,0 +1,424 @@
+// ============================================================================
+//                 SeqAn - The Library for Sequence Analysis
+// ============================================================================
+//
+// Copyright (c) 2006-2018, Knut Reinert & Freie Universitaet Berlin
+// Copyright (c) 2016-2018, Knut Reinert & MPI Molekulare Genetik
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//     * Redistributions of source code must retain the above copyright
+//       notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//     * Neither the name of Knut Reinert or the FU Berlin nor the names of
+//       its contributors may be used to endorse or promote products derived
+//       from this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL KNUT REINERT OR THE FU BERLIN BE LIABLE
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+// DAMAGE.
+//
+// ============================================================================
+
+/*!\file
+ * \author Joshua Kim <joshua.kim AT fu-berlin.de>
+ * \brief Contains seqan3::aa20, container aliases and string literals.
+ */
+
+#pragma once
+
+#include <cassert>
+#include <vector>
+
+#include <seqan3/core/platform.hpp>
+#include <seqan3/alphabet/detail/convert.hpp>
+#include <seqan3/alphabet/aminoacid/concept.hpp>
+
+namespace seqan3
+{
+/*!\brief The canonical amino acid alphabet.
+ * \ingroup aminoacid
+ * \implements seqan3::aminoacid_concept
+ *
+ * \details
+ * The alphabet consists of letters A, C, D, E, F, G, H, I, K, L, M, N, P, Q, R, S, T, V, W, Y
+ *
+ * The alphabet may be brace initialized from the static letter members (see above). Note that you cannot assign
+ * regular characters, but additional functions for this are available.
+ *
+ * *Note:* Letters which belong in the extended alphabet will be automatically converted based on the frequency
+ * of their options.\n Terminator characters are converted to W, because the most commonly occurring
+ * stop codon in higher eukaryotes is UGA<sup>2</sup>.
+ * Anything unknown is converted to S, because it occurs most frequently across 53 vertebrates<sup>1</sup>.
+ *
+ * |Input Letter  |Converts to  |
+ * |--------------|-------------|
+ * |B             |D<sup>1</sup>|
+ * |J             |L<sup>1</sup>|
+ * |O             |L<sup>1</sup>|
+ * |U             |C<sup>1</sup>|
+ * |Z             |E<sup>1</sup>|
+ * |X (Unknown)   |S<sup>1</sup>|
+ * |* (Terminator)|W<sup>2</sup>|
+ * <sup><b>1</b></sup>King, J. L., & Jukes, T. H. (1969). Non-Darwinian Evolution.
+ * Science, 164(3881), 788-798. doi:10.1126/science.164.3881.788\n
+ * <sup><b>2</b></sup>Trotta, E. (2016). Selective forces and mutational biases drive stop codon usage
+ * in the human genome: a comparison with sense codon usage.
+ * BMC Genomics, 17, 366. http://doi.org/10.1186/s12864-016-2692-4
+ *
+ * ### Example
+ *```cpp
+ *     aa20 my_letter{aa20::A};
+ *     // doesn't work:
+ *     // aa20 my_letter{'A'};
+ *
+ *     my_letter.assign_char('C'); // <- this does!
+ *
+ *     my_letter.assign_char('?'); // converted to A internally
+ *     if (my_letter.to_char() == 'A')
+ *        std::cout << "yeah\n"; // "yeah";
+ *```
+ */
+
+struct aa20
+{
+    //!\brief The type of the alphabet when converted to char (e.g. via to_char()).
+    using char_type = char;
+
+    //!\brief The type of the alphabet when represented as a number (e.g. via to_rank()).
+    using rank_type = uint8_t;
+
+
+    /*!\name Letter values
+     * \brief Static member "letters" that can be assigned to the alphabet or used in aggregate initialization.
+     * \details Similar to an Enum interface . *Don't worry about the `internal_type`.
+     */
+    //!\{
+    static const aa20 A;
+    static const aa20 B;
+    static const aa20 C;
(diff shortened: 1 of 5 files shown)


h-2:
@joergi-w Can you review this tomorrow?

include/seqan3/alphabet/aminoacid/aa20.hpp
```diff
@@ -0,0 +1,412 @@
...
+        {
+            using in_t = internal_type;
+            std::array<in_t, 256> ret{};
+
+            // initialize with UNKNOWN (std::array::fill unfortunately not constexpr)
+            for (auto & c : ret)
+                c = in_t::UNKNOWN;
+
+            ret['A'] = in_t::A; ret['a'] = in_t::A;
+            ret['B'] = in_t::D; ret['b'] = in_t::D; // Convert b (either D/N) to D, since D occurs more frequently. King, J. L. and T. H. Jukes. 1969. Non-Darwinian evolution. Science 164:788-798
```
joergi-w: These comments are very helpful! Sadly, they cannot be seen in the User Documentation in this way. Consider doxygen comments.

joergi-w: Line length

joergi-w:
Well done!

include/seqan3/alphabet/aminoacid/aa20.hpp
```diff
@@ -0,0 +1,425 @@
...
+#include <cassert>
+#include <vector>
+
+#include <seqan3/core/platform.hpp>
+#include <seqan3/alphabet/detail/convert.hpp>
+#include <seqan3/alphabet/aminoacid/concept.hpp>
+
+namespace seqan3
+{
+/*!\brief The canonical amino acid alphabet
```
h-2: Brief descriptions always need to end with a full stop.

include/seqan3/alphabet/aminoacid/aa20.hpp
```diff
@@ -0,0 +1,425 @@
...
+ *     aa20 my_letter{aa20::A};
+ *     // doesn't work:
+ *     // aa20 my_letter{'A'};
+ *
+ *     my_letter.assign_char('C'); // <- this does!
+ *
+ *     my_letter.assign_char('?'); // converted to A internally
+ *     if (my_letter.to_char() == 'A')
+ *        std::cout << "yeah\n"; // "yeah";
+ *```
```
h-2: I would move this section below the note on conversion and give it its own header `### Example`

include/seqan3/alphabet/aminoacid/aa20.hpp
```diff
@@ -0,0 +1,425 @@
...
+ *
+ * <b>Note</b>: Letters not belonging to the canonical alphabet will be automatically translated. Letters
```
h-2: "translated" is potentially confusing in the context of amino acids, I would use "converted" instead,

include/seqan3/alphabet/aminoacid/aa20.hpp
```diff
@@ -0,0 +1,425 @@
...
+ * <tr><td>Z                <td>E<sup>1</sup>
+ * <tr><td>X (Unknown)      <td>S<sup>1</sup>
+ * <tr><td>* (Terminator)   <td>W<sup>2</sup>
+ * </table>
+ * <b><sup>1</sup></b>King, J. L., & Jukes, T. H. (1969). Non-Darwinian Evolution. <br>
+ * Science, 164(3881), 788-798. doi:10.1126/science.164.3881.788<br>
+ * <b><sup>2</sup></b>Trotta, E. (2016). Selective forces and mutational biases drive stop codon usage <br>
+ * in the human genome: a comparison with sense codon usage. <br>
+ * BMC Genomics, 17, 366. http://doi.org/10.1186/s12864-016-2692-4
+ */
```
h-2: Although inline HTML works fine mostly, we prefer to use Markdown when possible, so `*foo*` instead of `<b>foo</b>` a.s.o. It should also work for tables, but I don't think there is markdown for sup so you can use `<sub>`.

h-2: Other than that a very good description :+1:

include/seqan3/alphabet/aminoacid/aa20.hpp
```diff
@@ -0,0 +1,425 @@
...
+
+public:
+    //!\privatesection
+    //!\brief The data member.
+    internal_type _value;
+    //!\publicsection
+};
+
+constexpr aa20 aa20::A{internal_type::A};
+constexpr aa20 aa20::B{internal_type::D};
```
h-2: You could do the mapping here, too, i.g. `constexpr aa20 aa20::B{aa20::D};`, because this is visible from the documentation and it makes it eaiser to distinguish remapped characters from fully supported ones.

include/seqan3/alphabet/aminoacid/concept.hpp
```diff
@@ -0,0 +1,133 @@
...
+#include <seqan3/alphabet/concept.hpp>
+
+// ============================================================================
+// concept
+// ============================================================================
+
+namespace seqan3
+{
+
+/*!\interface seqan3::aminoacid_concept <>
```
h-2: `amino_acid_concept`, because we introduce `_` whenerver there would be a space in English...

include/seqan3/alphabet/aminoacid/concept.hpp
```diff
@@ -0,0 +1,133 @@
...
+ * \extends seqan3::alphabet_concept
+ * \brief A concept that indicates whether an alphabet represents amino acids.
+ * \ingroup aminoacid
+ *
+ * In addition to the requirements for seqan3::alphabet_concept, the aminoacid_concept introduces
+ * a requirement for an accepted alphabet. All aminoacid_concept members should accept the 27
+ * amino acid letters, and convert them to fit their respective size constraints.
```
h-2: I would write instead:
```
In addition to the requirements for seqan3::alphabet_concept, the amino_acid_concept expects
conforming alphabets to provide an enum-like interface with all possible 27 amino acids as values
(although some may be mapped to others if the alphabet is smaller than 27).
```
Because it is more about requiring that the alphabet *provide* these than *accepting* these.

include/seqan3/alphabet/aminoacid/concept.hpp
h-2: OTOH we also called the submodule "aminoacid" and not "amino_acid", :thinking:

include/seqan3/alphabet/aminoacid/concept.hpp
```diff
@@ -0,0 +1,133 @@
...
+ * |    Tyrosine                | Tyr                         | Y                           | Y                          |
+ * |    Valine                  | Val                         | V                           | V                          |
+ * |    Selenocysteine          | Sec                         | C                           | U                          |
+ * |    Pyrrolysine             | Pyl                         | K                           | O                          |
+ * | Asparagine or aspartic acid| Asx                         | D                           | B                          |
+ * | Glutamine or glutamic acid | Glx                         | E                           | Z                          |
+ * |    Leucine or Isoleucine   | Xle                         | L                           | J                          |
+ * |    Unknown                 | Xaa                         | S                           | X                          |
+ * |    Stop Codon              | N/A                         | W                           | TERMINATOR                 |
+ */
```
h-2: This table is definitely helpful, but I would move it to the general description of the aminoacid submodule where we have such a table already.

I would merge them and make four columns:

| Full name | Three letter code | One letter code | Remapped in seqan3::aa20

And then write after the table something like:
```
All amino acid alphabets provide static value members (like an enum) for all amino acids in the form of the one-letter representation (except that the stop codon is called ::TERMINATOR). As shown above, alphabets smaller than 27 internally represent multiple amino acids as one.

For most cases it is highly recommended to use seqan3::aa27 as seqan3::aa20 provides no benefits in regard to space consumption (both need 5bits). Use it only when you know you need to interface with other software of formats that only support the canonical set.
```

test/unit/alphabet/aminoacid_test.cpp
```diff
@@ -83,55 +98,159 @@ TYPED_TEST(aminoacid, assign_char)
...
 }
 
 TYPED_TEST(aminoacid, stream_operator)
 {
     std::stringstream ss;
-    ss << TypeParam::A << TypeParam::C << TypeParam::G;
-    EXPECT_EQ(ss.str(), "ACG");
+    ss << TypeParam::A << TypeParam::C << TypeParam::G << TypeParam::B << TypeParam::J
+       << TypeParam::O << TypeParam::U << TypeParam::X << TypeParam::Z;
+    if constexpr (std::is_same_v<TypeParam, aa27>)
```
h-2: insert empty line

include/seqan3/alphabet/aminoacid/aa20.hpp
```diff
@@ -0,0 +1,426 @@
...
+ * \ingroup aminoacid
+ * \implements seqan3::aminoacid_concept
+ *
+ * \details
+ * The alphabet consists of letters A, C, D, E, F, G, H, I, K, L, M, N, P, Q, R, S, T, V, W, Y
+ *
+ * The alphabet may be brace initialized from the static letter members (see above). Note that you cannot assign
+ * regular characters, but additional functions for this are available.
+
+ *
```
h-2: remove the empty line up there

include/seqan3/alphabet/aminoacid/aa20.hpp
```diff
@@ -0,0 +1,426 @@
...
+ * *Note:* Letters not belonging to the canonical alphabet will be automatically converted. Letters
+ * which belong in the extended alphabet will be automatically converted based on the frequency of their options.\n
```
h-2: I think the first sentence is superfluous, or not ?

Oh wow, and does '\n' actually introduce a newline in the documentation? I didn't know that!

include/seqan3/alphabet/aminoacid/aa20.hpp
```diff
@@ -0,0 +1,426 @@
...
+ * |Z             |E<sup>1</sup>|
+ * |X (Unknown)   |S<sup>1</sup>|
+ * |* (Terminator)|W<sup>2</sup>|
+ * <sup><b>1</b></sup>King, J. L., & Jukes, T. H. (1969). Non-Darwinian Evolution.
+ * Science, 164(3881), 788-798. doi:10.1126/science.164.3881.788\n
+ * <sup><b>2</b></sup>Trotta, E. (2016). Selective forces and mutational biases drive stop codon usage
+ * in the human genome: a comparison with sense codon usage.
+ * BMC Genomics, 17, 366. http://doi.org/10.1186/s12864-016-2692-4
+ *
+ * ### Example:
```
h-2: No colon in headlines

include/seqan3/alphabet/aminoacid/all.hpp
```diff
@@ -55,38 +57,41 @@
...
+ * |    Tryptophan              | Trp               | W               | W                          |
+ * |    Tyrosine                | Tyr               | Y               | Y                          |
+ * |    Valine                  | Val               | V               | V                          |
+ * |    Selenocysteine          | Sec               | U               | C                          |
+ * |    Pyrrolysine             | Pyl               | O               | K                          |
+ * | Asparagine or aspartic acid| Asx               | B               | D                          |
+ * | Glutamine or glutamic acid | Glx               | Z               | E                          |
+ * |    Leucine or Isoleucine   | Xle               | J               | L                          |
+ * |    Unknown                 | Xaa               | X               | S                          |
+ * |    Stop Codon              | N/A               | TERMINATOR      | W                          |
```
h-2: I would mark those entries in the last column that are not changed just as `-`, because that makes spotting the changes easier. or maybe use colours? What do you think?

h-2: Also the "official" one-letter-code for the Stop Codon is `*`. The paragraph below states that we use `::TERMINATOR` but I would keep the official things in this table.

joshuak94: On second thought, maybe I will try the colors. I forgot that I grouped all of the extra alphabet letters at the bottom, so having two color groupings wouldn't be too bad.

joshuak94:
@h-2 I don't think there is anything wrong with calling it just aminoacid. Even though it is two words and should have the underscore, I think it's more consistent with the fact that nucleotide doesn't have any underscores and the initialization aa20/aa27 also don't have any underscores.

h-2:
Oh, and also make sure that all the continuous integration passes, it looks like you still have an error in the documentation!

PR #301, opened 2018-07-12
[FIX] Fixes pipeline to actually call deploy node.

rrahn:


rrahn:
!rebuild

h-2:
whats up wit the java error on jenkins?

rrahn:
Well, if only I would know. But I could at least narrow it down. Still looking for the actual cause, and then hopefully fix it

rrahn:
Blocked by #304

PR #304, opened 2018-07-13
[FIX] Remove tag expression in deploy step since it breaks the pipeline.

rrahn:


h-2:
This fixes the java issue of the other PRs?

PR #305, opened 2018-07-17
FASTQ

h-2:


include/seqan3/io/sequence/sequence_file_format_fasta.hpp
```diff
@@ -242,56 +221,49 @@ class sequence_file_format_fasta
...
+                                           {
+                                               throw parse_error{std::string{"Encountered an unexpected letter: "} +
+                                                                   is_legal_alph.msg.string() +
+                                                                   " evaluated to false on " +
+                                                                   detail::make_printable(c)};
+                                           }
+                                           return c;
+                                       })                                           // enforce legal alphabet
+                                     | view::char_to<value_type_t<seq_type>>,       // convert to actual target alphabet
+                         detail::make_conversion_output_iterator(seq));
```
joergi-w: It's not clear to me: why do we need `detail::make_conversion_output_iterator` instead of `ranges::back_inserter` here?

include/seqan3/io/sequence/sequence_file_format_fastq.hpp
```diff
@@ -0,0 +1,326 @@
...
+        }
+        else
+        {
+            ranges::copy(stream_view | view::take_line_or_throw,
+                         detail::make_conversion_output_iterator(id));
+        }
+
+        /* Sequence */
+        auto seq_view = stream_view | ranges::view::take_while(!is_char<'+'>{})    // until 2nd ID line TODO: or_throw
+                                    | ranges::view::remove_if(is_space);           // ignore whitespace
```
joergi-w: in fasta we ignore also numbers...

include/seqan3/io/sequence/sequence_file_format_fastq.hpp
```diff
@@ -0,0 +1,326 @@
...
+            if (ranges::empty(id)) //[[unlikely]]
+                throw std::runtime_error{"The ID field may not be empty when writing FASTQ files."};
+
+            stream_it = '@';
+            ranges::copy(id, stream_it);
+
+            if (options.add_carriage_return)
+                stream_it = '\r';
+
+            stream_it = '\n';
```
joergi-w: Shouldn't we define a shortcut, e.g. `add_eol` or `endl`, as we need this multiple times in every file format?

include/seqan3/io/sequence/sequence_file_format_fastq.hpp
```diff
@@ -0,0 +1,326 @@
...
+    //!\brief The valid file extensions for this format; note that you can modify this value.
+    static inline std::vector<std::string> file_extensions
+    {
+        { "fastq" },
+        { "fq"    }
+    };
+
+    //!\copydoc sequence_file_in_format_concept::read
+    template <typename stream_type,     // constraints checked by file
+              typename seq_legal_alph_type, bool seq_qual,
```
joergi-w: in the concept you call it `seq_qual_in_one`, please change either of these for consistency

include/seqan3/io/sequence/sequence_file_in_options.hpp
```diff
@@ -44,7 +44,7 @@
 namespace seqan3
 {
 //!\brief The options type defines various option members that influence the behaviour of all or some formats.
-template <typename sequence_legal_alphabet_>
+template <typename sequence_legal_alphabet_, bool seq_qual_in_one>
```
joergi-w: here is the counterpart: in the file you named this `seq_qual`

include/seqan3/io/sequence/sequence_file_out.hpp
```diff
@@ -841,32 +861,57 @@ class sequence_file_out
...
+                {
+                    f.write(stream, options, *seqs_it, *ids_it, *quals_it);
+
+                    ++seqs_it;
+                    ++ids_it;
+                    ++quals_it;
+                } while ((seqs_it != seqs_end)   &&
+                         (ids_it != ids_end)     &&
+                         (quals_it != quals_end));
+            }
```
joergi-w: You could use zip_iterator here, couldn't you?

test/unit/io/sequence/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
 seqan3_test(sequence_file_in_test.cpp)
 seqan3_test(sequence_file_out_test.cpp)
 seqan3_test(sequence_file_format_fasta_test.cpp)
+seqan3_test(sequence_file_format_fastq_test.cpp)
```
joergi-w: The file `sequence_file_format_fastq_test.cpp` is not contained in this PR!

include/seqan3/io/sequence/sequence_file_format_fasta.hpp
```diff
@@ -126,23 +128,16 @@ class sequence_file_format_fasta
 
     //!\copydoc sequence_file_in_format_concept::read
     template <typename stream_type,     // constraints checked by file
-              typename seq_legal_alph_type,
+              typename seq_legal_alph_type, bool seq_qual,
```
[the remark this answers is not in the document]
joergi-w: Ahh, probably you want to group them because they are the options. I guess it's fine then.

include/seqan3/io/sequence/sequence_file_format_fasta.hpp
h-2: In general the detail::make_conversion_output_iterator is deprecated and should be removed by @rrahn. The reason I am still using it in some places, is that we sometimes need an custom overload if the parameter is actually `std::ignore`. I guess we should should make our own back_insert_iterator that inherits from `ranges::back_insert_iterator` by default, but is specialised for `std::ignore`. Do you want to make a PR? :smiley:

include/seqan3/io/sequence/sequence_file_format_fastq.hpp
h-2: Yes, but AFAICT numbers are not legal for fastq

include/seqan3/io/sequence/sequence_file_out.hpp
h-2: yes, good catch :+1:

PR #307, opened 2018-07-20
[FEATURE] structured sequence I/O

joergi-w:
This PR adds functions and tests for structured sequence I/O. 
There is also the first file format implemented: Dot Bracket Notation.


include/seqan3/io/structure/structure_file_format_dot_bracket.hpp
```diff
@@ -0,0 +1,399 @@
...
+              structure_file_in_options<seq_legal_alph_type, structured_seq_combined> const & options,
+              seq_type & seq,
+              id_type & id,
+              bpp_type & bpp,
+              structure_type & structure,
+              energy_type & energy,
+              react_type & react,
+              react_type & react_err,
+              comment_type & comment,
+              offset_type & offset)
```
h-2: You probably get a lot of unused variable warning from this! I would wrap the unused ones in SEQAN3_DOXYGEN_ONLY

Aso you could align the `&` but you don't have to.

include/seqan3/io/structure/structure_file_format_dot_bracket.hpp
```diff
@@ -0,0 +1,399 @@
...
+#include <seqan3/range/view/to_char.hpp>
+#include <seqan3/range/view/take_line.hpp>
+#include <seqan3/std/concept/range.hpp>
+#include <seqan3/std/view/subrange.hpp>
+#include <seqan3/std/view/transform.hpp>
+
+namespace seqan3
+{
+/*!\brief       The Dot Bracket format.
+ * \implements  structure_file_format_concept
```
h-2: misses seqan3:: ?

include/seqan3/io/structure/structure_file_in.hpp
```diff
@@ -0,0 +1,1217 @@
...
+    /*!\name Member types
+     * \brief Definitions to satisfy seqan3::structure_file_in_traits_concept.
+     * \{
+     */
+    using seq_alphabet        = aa27;
+    using seq_legal_alphabet  = aa27;
+    using structure_alphabet       = dssp9;
+    using structure_legal_alphabet = dssp9;
+    template<typename _seq_alphabet, typename _structure_alphabet>
+    using structured_seq_alphabet        = structured_aa<_seq_alphabet, _structure_alphabet>;
```
h-2: either align the `=` or just use single spaces!

include/seqan3/io/structure/structure_file_in.hpp
```diff
@@ -0,0 +1,1217 @@
...
+ *     "> example\n"
+ *     "UUGGAGUACACAACCUGUACACUCUUUC\n"
+ *     "..(((((..(((...)))..)))))... (-3.71)\n"
+ * };
+ *
+ * std::istringstream iss(input);
+ *
+ * structure_file_in fin{std::move(iss), structure_file_format_dot_bracket{}};
+ * //              ^ no need to specify the template arguments
+ * ```
```
h-2: BTW, you can now use code-snippets from tests, as well, I don't know if that would be helpful for you. See https://github.com/seqan/seqan3/blob/master/include/seqan3/range/detail/inherited_iterator_base.hpp with an example of how it works.

include/seqan3/io/structure/structure_file_in.hpp
```diff
@@ -0,0 +1,1217 @@
...
+ * You can iterate over this file record-wise:
+ *
+ * ```cpp
+ * structure_file_in fin{"/tmp/my.dbn"};
+ *
+ * for (auto & rec : fin)
+ * {
+ *     std::cout << "ID:  " << get<field::ID>(rec) << '\n';
+ *     std::cout << "SEQ: " << (get<field::SEQ>(rec) | view::to_char) << '\n'; // sequence is converted to char on-the-fly
+ *     std::cout << "STRUCTURE: " << (get<field::STRUCTURE>(rec) | view::to_char) << '\n';
```
h-2: I think `field::STRUCTURE` is not default anymore, right? So this example wouldn't work. Please double-check also the other examples!

include/seqan3/io/structure/structure_file_in.hpp
```diff
@@ -0,0 +1,1217 @@
...
+
+        // buffer first record
+        read_next_record();
+    }
+    /* NOTE(h-2): Curiously we do not need a user-defined deduction guide for the above constructor.
+     * A combination of default template parameters and auto-deduction guides works as expected,
+     * independent of whether the second/optional parameter is specified or not, i.e. it is possible
+     * to auto-deduct and overwrite a single template parameter out of the four if the optional parameter
+     * is specified and use the default otherwise.
+     */
```
h-2: you can remove this comment, having in one file is definetely enough 😉

include/seqan3/io/structure/structure_file_in.hpp
```diff
@@ -0,0 +1,1217 @@
...
+template<size_t elem_no,
+         seqan3::structure_file_in_traits_concept traits_type,
+         seqan3::detail::fields_concept selected_field_ids,
+         seqan3::detail::type_list_of_structure_file_in_formats_concept valid_formats,
+         seqan3::istream_concept<char> stream_type>
+struct tuple_element<elem_no, seqan3::structure_file_in<traits_type, selected_field_ids, valid_formats, stream_type>>
+    : tuple_element<elem_no, typename seqan3::structure_file_in<traits_type,
+                                                                selected_field_ids,
+                                                                valid_formats,
+                                                                stream_type>::file_as_tuple_type>
```
h-2: I remember we fixed something here, can you make a PR for sequence_io or at least open an issue so we don't forget? Thanks!

include/seqan3/io/structure/structure_file_out_format_concept.hpp
```diff
@@ -0,0 +1,194 @@
...
+#include <seqan3/alphabet/structure/all.hpp>
+#include <seqan3/core/type_list.hpp>
+
+namespace seqan3
+{
+//!\brief The options type defines various option members that influence the behaviour of all or some formats.
+struct structure_file_out_options
+{
+    //!\brief Begin the ID line with ";" instead of ">" (not recommended).
+    bool fasta_legacy_id_marker = false;
```
h-2: There is no "FastA" format for structure_file_in so this definetely needs to be called different! From my POV we could just remove them, or do you think dbn actually has this?

include/seqan3/io/structure/structure_file_format_dot_bracket.hpp
```diff
@@ -0,0 +1,399 @@
...
+#include <seqan3/range/view/char_to.hpp>
+#include <seqan3/range/view/to_char.hpp>
+#include <seqan3/range/view/take_line.hpp>
+#include <seqan3/std/concept/range.hpp>
+#include <seqan3/std/view/subrange.hpp>
+#include <seqan3/std/view/transform.hpp>
+
+namespace seqan3
+{
+/*!\brief       The Dot Bracket format.
```
h-2: Maybe we should call it Vienna instead? Because we use "dot bracket" already for an alphabet and it could be confusing. Maybe it's clearer to say that the STRUCTURE field is in "dot bracket", but the file is "vienna"? Also it would look nice when we add stockholm, because of city names 😄

h-2:
And almost all your checks fail 😉

include/seqan3/io/structure/structure_file_format_dot_bracket.hpp
```diff
@@ -0,0 +1,399 @@
...
+ *
+ * \details
+ *
+ * ### Introduction
+ *
+ * Dot Bracket Notation is widely used for secondary structure annotation. Is is similar to the FastA format,
+ * containing an ID in the first line and a sequence in the second line. An additional third line represents the
+ * secondary structure, using brackets to denote interacting nucleotides or amino acids, and dots for unpaired sites.
+ * Optionally, the structure can be followed by a space character and the minimum free energy value enclosed
+ * in round brackets ().
```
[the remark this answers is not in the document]
joergi-w: True, thanks!

include/seqan3/io/structure/structure_file_in.hpp
joergi-w: Yes, I found some more stuff and I already have a separate branch for that.

include/seqan3/io/structure/structure_file_out_format_concept.hpp
joergi-w: I'll just remove this, as I never saw it in a structure file. I thought it wouldn't hurt, if we allow it. Basically the format is just FastA, extended by the structure line and without line breaks inside the sequence.

include/seqan3/io/structure/structure_file_format_dot_bracket.hpp
joergi-w: Vienna is a subset of this format, as Vienna doesn't allow for pseudo knots and therefore the alphabet must be limited to db3 instead of wuss51 and we have to drop energies. I'll think about it.

joergi-w:
Thank you for the immediate review and useful comments! I'll fix everything during next week.

include/seqan3/io/structure/structure_file_format_dot_bracket.hpp
```diff
@@ -0,0 +1,399 @@
...
+                                 if (!is_legal_structure(c))
+                                 {
+                                     throw parse_error{
+                                         std::string{"Encountered an unexpected letter: "} +
+                                         is_legal_structure.msg.string() +
+                                         " evaluated to false on " + detail::make_printable(c)};
+                                 }
+                                 return c;
+                             })                                  // enforce legal alphabet
+                           | view::char_to<alph_type>;           // convert to actual target alphabet
```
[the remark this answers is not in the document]
joergi-w: I need to call the function three times with different types (`value_type_t<structure_type>`, `typename value_type_t<structure_type>::structure_alphabet_type` and `wuss51`). As it is difficult to resolve the _alph_type_ beforehand, as the candidates just exist under certain conditions, I'm keeping this as a private function template.

include/seqan3/io/structure/structure_file_in.hpp
joergi-w: The field::STRUCTURE becomes default now, to be consistent with structure_file_out and also to be similar to sequence_io. I think it's more intuitive to use it this way.

include/seqan3/io/structure/structure_file_format_dot_bracket.hpp
joergi-w: I pushed a commit that switches the dot_bracket to vienna format.

include/seqan3/io/structure/structure_file_in.hpp
joergi-w: see PR #313

h-2:
One more thing, I though about the file naming and submodule structure and would propose the following change:
```
io/structure/                                       → io/structure_file/
io/structure/structure_file_in.hpp                  → io/structure_file/input.hpp
io/structure/structure_file_in_format_concept.hpp   → io/structure_file/input_format_concept.hpp
io/structure/structure_file_format_vienna.hpp       → io/structure_file/format_vienna.hpp
...
```

So the change would be
1. spell out `input` and `output` (also in the class names)
2. include `_file` in the folder name (because not all sub-modules of `io` are files)
3. strip `structure_file` from the filenames, because it is redundant (but keep in the class names)

I will make a PR to adjust the sequenc_file.

include/seqan3/io/structure_file/format_vienna.hpp
```diff
@@ -0,0 +1,425 @@
...
+    structure_file_format_vienna(structure_file_format_vienna const &) = delete;
+    structure_file_format_vienna & operator=(structure_file_format_vienna const &) = delete;
+    structure_file_format_vienna(structure_file_format_vienna &&) = default;
+    structure_file_format_vienna & operator=(structure_file_format_vienna &&) = default;
+    //!\}
+
+    //!\brief The valid file extensions for this format; note that you can modify this value.
+    static inline std::vector<std::string> file_extensions
+    {
+        { "dbn" }
```
h-2: in SeqAn2 we have `.dbv`, too, is that not used anymore?

include/seqan3/io/structure_file/detail.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+namespace seqan3::detail
+{
+/*!
+ * \brief Transforms a structure annotation string into a base pair probability matrix.
+ * \throws seqan3::parse_error if unpaired brackets are found in the structure annotation.
+ * \tparam structure_alph_type The type of the structure alphabet; must satisfy seqan3::rna_structure_concept.
+ * \tparam bpp_type            The type of the target matrix.
+ * \tparam structure_type      The range type of the structure annotation.
+ * \param[out] bpp             The target matrix that receives the base pair probabilities.
```
h-2: Why don't you return this instead of doing an out-parameter?

h-2: Ah, because you need to specify the type?

include/seqan3/io/structure_file/input_options.hpp
```diff
@@ -0,0 +1,60 @@
...
+ * same/combined.
+ */
+template<typename seq_legal_alphabet, bool structured_seq_combined>
+struct structure_file_input_options
+{
+    //!\brief Read the ID string only up until the first whitespace character.
+    bool truncate_ids = false;
+
+    //!\brief Switch this off, if you want to read files without structure.
+    bool file_has_structure = true;
```
h-2: I don't understand this. If you don't want the structure, don't request the field :thinking:

include/seqan3/io/structure_file/output.hpp
```diff
@@ -0,0 +1,967 @@
...
+                      energy_type && energy,
+                      react_type && react,
+                      react_type && react_error,
+                      comment_type && comment,
+                      offset_type && offset)
+    {
+        static_assert(detail::decays_to_ignore_v<structured_seq_type> ||
+                      (detail::decays_to_ignore_v<seq_type> && detail::decays_to_ignore_v<structure_type>),
+                      "You may not select field::STRUCTURED_SEQ and either of field::SEQ and field::STRUCTURE "
+                      "at the same time.");
```
h-2: I think the condition is wrong, or not?

include/seqan3/io/structure_file/output.hpp
```diff
@@ -0,0 +1,967 @@
...
+
+        assert(!format.valueless_by_exception());
+        std::visit([&] (auto & f)
+        {
+            if constexpr (!detail::decays_to_ignore_v<reference_t<structured_seq_type>>)
+            {
+                auto zipped = ranges::view::zip(structured_seq, id, bpp, energy, react, react_error, comment, offset);
+
+                for (auto && v : zipped)
+                    f.write(stream,
```
h-2: please wrap in `{ }` when the statement spans multiple lines even if it's just one statement.

include/seqan3/io/structure_file/output.hpp
```diff
@@ -0,0 +1,967 @@
...
+                            std::get<6>(v),  // comment
+                            std::get<7>(v)); // offset
+            }
+            else
+            {
+                auto zipped = ranges::view::zip(seq, id, bpp, structure, energy, react, react_error, comment, offset);
+
+                for (auto && v : zipped)
+                    f.write(stream, options, std::get<0>(v), std::get<1>(v), std::get<2>(v), std::get<3>(v),
+                             std::get<4>(v), std::get<5>(v), std::get<6>(v), std::get<7>(v), std::get<8>(v));
```
h-2: here, as well!

test/unit/io/structure_file/structure_file_input_test.cpp
```diff
@@ -0,0 +1,471 @@
...
+        auto interactions = bpp | ranges::view::remove_if([] (auto & set) { return set.size() != 1; });
+        for (auto & elem : interactions)
+        {
+            EXPECT_EQ(elem.begin()->second, bpp_comp[idx++]);
+        }
+        EXPECT_EQ(idx, bpp_comp.size());
+    }
+};
+
+struct structure_file_input_record_reading : public structure_file_input_read {};
```
h-2: why this?

include/seqan3/io/structure_file/format_vienna.hpp
joergi-w: It is not used anymore, but I added `.fa` and `.fasta` instead.

include/seqan3/io/structure_file/detail.hpp
joergi-w: Good idea, I give the type by template argument and return `bpp`.

include/seqan3/io/structure_file/output.hpp
joergi-w: I cannot find anything wrong here... Either structured_seq must be ignored or both seq and structure.

test/unit/io/structure_file/structure_file_input_test.cpp
joergi-w: I wanted to create a different test group that has access to the same struct. But I can remove this and rename the tests, if it looks weird.

include/seqan3/io/structure_file/output.hpp
h-2: Yep, my bad.

include/seqan3/io/structure_file/detail.hpp
h-2: Ah, no, sorry, I meant that if you need to able specify the type, than it's perfectly fine to use an out-parameter and likely a better solution that having to pass an extra template parameters.

test/unit/io/structure_file/structure_file_input_test.cpp
h-2: It's not important, really.

include/seqan3/io/structure_file/detail.hpp
h-2: Could you revert this change? Otherwise it looks good now.

Issue #306, opened 2018-07-18
mask-alphabet and masked-alphabet composition


h-2:
- [x] Read through http://docs.seqan.de/seqan/3.0.0-master-user/group__composition.html and the individual documentation pages; you can also look at the code of the the composition templates, but it is probably more difficult to understand
  - [ ] Instead or additionally look at the example alphabets that are based on them, like seqan3::gapped (a union_composition) and seqan3::quality_composition (a cartesian_composition) and try figure out what's going on
  - [ ] implement `seqan3::mask`, a "semi_alphabet" (only rank notation, no char notation) that has bool as rank type or an enum and the static members `::UNMASKED` and `::MASKED`
  - [ ] implement `seqan3::masked<>` template that is derived of seqan3::cartesian_composition and combines the seqan3::mask alphabet with another alphabet, in such a way that the character representation is that of the other alpahbet, except that if the seqan3::mask half of the combined alphabet is true, the character will be converted to lower-case ( you can use https://en.cppreference.com/w/cpp/locale/tolower for this)

In the end one should be able to do:
```cpp
masked<dna5> l;
l = dna5::G;
l = mask::UNMASKED;
std::cout << to_char(l); // prints 'G'
l = mask::MASKED;
std::cout << to_char(l); // prints 'g'
```

PS: you will need to relax the constraints on types that are accepted by cartesian_composition, currently they are required to alphabet, but they should only be required to be semi_alphabet.

marehr:
Can you give a rational for the naming `UNMASKED` and `MASKED`? It sounds odd to me.

h-2:
> Can you give a rational for the naming UNMASKED and MASKED? It sounds odd to me.

https://bioinformatics.stackexchange.com/questions/225/uppercase-vs-lowercase-letters-in-reference-genome

h-2:
@joshuak94 You can ask @marehr all the technical questions, but @joergi-w should also be able to help you out while I am on vacation and in the next days!

marehr:
Is this done?

joshuak94:
@marehr I think so... #308 introduces the mask and masked classes, both of which have been merged into the master.

PR #308, opened 2018-07-20
[FEATURE] Introduce the semialphabet mask and its test class.

joshuak94:
Regarding issue #306, introduces the mask semi alphabet along with its test case. Still to do: implement the mask composition.


include/seqan3/alphabet/mask/all.hpp
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+#include <seqan3/alphabet/mask/mask.hpp>
+
+/*!\defgroup mask Mask
+ * \brief Contains the mask alphabet and functionality for creating masked compositions.
+ * \ingroup alphabet
+ *
+ * \par Introduction
+ * Masks are useful as cartesian compositions when one wants to create a masked alphabet with
```
h-2: Maybe give a little more information on when/why/where sequence masking happens, that there is "hard-masking" which converts to the `::UNKNOWN` character and that there is "soft-masking" that is visualised by using lower-case instead of upper-case. However because regular nucleotide and aminoacid alphabets discard case on assignment, one needs to create additional alphabets to preserve this information (if desired). Finally that this alphabet in itself is not useful to users directly, but instead the composition `seqan3::masked` may be used to transform another alphabet into a new alphabet that can represent the original alphabet plus masking information...

include/seqan3/alphabet/mask/mask.hpp
```diff
@@ -0,0 +1,153 @@
...
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+// DAMAGE.
+//
+// ============================================================================
+// Author: Joshua Kim <joshua.kim AT fu-berlin.de>
+// ============================================================================
+
+#pragma once
```
h-2: remove the author lines up there and add file documentation like in the other file.

include/seqan3/alphabet/mask/mask.hpp
```diff
@@ -0,0 +1,153 @@
...
+#include <seqan3/alphabet/concept_pre.hpp>
+
+namespace seqan3
+{
+/*!\brief Implementation of a masked alphabet to be used for cartesian compositions.
+ * \ingroup mask
+ * \implements seqan3::semi_alphabet_concept
+ *
+ * \details
+ * Should only be used in conjunction with mask::masked_composition, to create a new
```
h-2: Maybe instead `This alphabet is not usually used directly, but instead via seqan3::masked. For more information see the \link mask Mask submodule \endlink.

test/unit/alphabet/mask_test.cpp
```diff
@@ -0,0 +1,77 @@
...
+    EXPECT_TRUE(mask::MASKED >= mask::UNMASKED);
+    EXPECT_TRUE(mask::UNMASKED <= mask::MASKED);
+
+    EXPECT_FALSE(mask::MASKED == mask::UNMASKED);
+    EXPECT_FALSE(mask::MASKED != mask::MASKED);
+    EXPECT_FALSE(mask::MASKED > mask::MASKED);
+    EXPECT_FALSE(mask::UNMASKED < mask::UNMASKED);
+    EXPECT_FALSE(mask::MASKED <= mask::UNMASKED);
+    EXPECT_FALSE(mask::UNMASKED >= mask::MASKED);
+}
```
h-2: Eventually we could think about splitting the current `alphabet_test` into a `semi_alphabet_test` and an `alphabet_test`, but for now this is sufficient!

include/seqan3/alphabet/mask/mask.hpp
```diff
@@ -0,0 +1,153 @@
...
+        assert(c < value_size);
+        _value = static_cast<internal_type>(c);
+        return *this;
+    }
+
+    //!\brief Assign from mask::MASKED or mask::UNMASKED.
+    constexpr mask & assign_rank(mask const & c)
+    {
+        assert(c == MASKED || c == UNMASKED);
+        return assign_rank(c == MASKED ? 1 : 0);
```
rrahn: why this comparison and not just call `c.to_rank()`?

test/unit/alphabet/mask_test.cpp
```diff
@@ -0,0 +1,77 @@
...
+    EXPECT_EQ(m1.assign_rank(1), mask::MASKED);
+    EXPECT_TRUE(m1.to_rank());
+    EXPECT_EQ(m1.assign_rank(0), mask::UNMASKED);
+    EXPECT_FALSE(m1.to_rank());
+    EXPECT_EQ(m1.assign_rank(true), mask::MASKED);
+    EXPECT_EQ(m1.assign_rank(false), mask::UNMASKED);
+    EXPECT_EQ(m1.assign_rank(mask::MASKED), mask::MASKED);
+    EXPECT_TRUE(m1.to_rank());
+    EXPECT_EQ(m1.assign_rank(mask::UNMASKED), mask::UNMASKED);
+    EXPECT_FALSE(m1.to_rank());
```
rrahn: Please also add tests for temporaries.
Following versions should be tested:

 reference type | `to_rank` | `assign_rank` 
--------------- | --------- | ------------
`&` | ✅ | ✅
`const &` | ✅  | 🚫 
`&&` | ✅ | ✅
`const &&` | ✅ | 🚫

test/unit/alphabet/mask_test.cpp
```diff
@@ -0,0 +1,77 @@
...
+// Author: Joshua Kim <joshua.kim AT fu-berlin.de>
+// ============================================================================
+
+#include <gtest/gtest.h>
+
+#include <seqan3/alphabet/concept_pre.hpp>
+#include <seqan3/alphabet/mask/all.hpp>
+
+using namespace seqan3;
+
```
rrahn: Please add tests to check if mask alphabet is standard layout and trivial.
See: https://en.cppreference.com/w/cpp/types/is_standard_layout and https://en.cppreference.com/w/cpp/types/is_trivial

include/seqan3/alphabet/mask/mask.hpp
```diff
@@ -0,0 +1,153 @@
...
+    //!\brief Assign from a numeric value or true/false.
+    constexpr mask & assign_rank(rank_type const c)
+    {
+        assert(c < value_size);
+        _value = static_cast<internal_type>(c);
+        return *this;
+    }
+
+    //!\brief Assign from mask::MASKED or mask::UNMASKED.
+    constexpr mask & assign_rank(mask const & c)
```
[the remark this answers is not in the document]
h-2: No, assignment should return reference to self. The free function interface also depends on this.

test/unit/alphabet/mask_test.cpp
h-2: The semi_alphabet concept already does these checks:
https://github.com/seqan/seqan3/blob/master/include/seqan3/alphabet/concept.hpp#L92

joshuak94:
I'm not really sure why the `assert(c < value_size)` keeps failing in the continuous integration tests...

marehr:
Try to build it in debug mode. `cmake -DCMAKE_BUILD_TYPE=Debug <seqan3>`

h-2:
Tests still fail. Did you push changes?

joshuak94:
@h-2 I haven't been able to push any changes yet, wasn't in today. But I ran into a problem before I left, where my own tests were running into compile time errors because `masked<dna4>` isn't recognized as a tuple, or something like that?

include/seqan3/alphabet/mask/mask.hpp
```diff
@@ -0,0 +1,154 @@
...
+     * \brief Static member "booleans" that can be assigned to the alphabet or used in aggregate initialization.
+     * \details Similar to an Enum interface . *Don't worry about the `internal_type`.
+     */
+    //!\{
+    static const mask UNMASKED;
+    static const mask MASKED;
+    //!\}
+
+    //!\brief The size of the alphabet, i.e. the number of different values it can take.
+    static constexpr rank_type value_size{2};
```
eseiler:
```cpp
using rank_type = bool;
static constexpr rank_type value_size{2};
```
The value of value_size will be 1, so in Debug mode the assertion `assert(c < value_size);` will fail for MASKED/1/true.

eseiler:
```cpp
using rank_type = bool;
static constexpr rank_type value_size{2};
```
The value of value_size will be 1, so in Debug mode the assertion `assert(c < value_size);` will fail for MASKED/1/true.
I also commented in the code.

include/seqan3/alphabet/composition/cartesian_composition.hpp
```diff
@@ -74,13 +74,13 @@ namespace seqan3
  * \endcond
  *
  * \sa quality_composition
```
h-2: \sa qualified

include/seqan3/alphabet/mask/masked.hpp
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    //!\brief Directly assign the mask letter.
+    constexpr masked & operator=(mask const l) noexcept
+    {
+        get<1>(*this) = l;
+        return *this;
+    }
+
+    //!\brief Assign from a character. This modifies the internal sequence letter.
```
h-2: copy'n'paste error, it modifies both internal letters so just delete the second sentence.

include/seqan3/alphabet/mask/masked.hpp
```diff
@@ -0,0 +1,118 @@
...
+        seqan3::assign_char(get<0>(*this), c);
+        seqan3::assign_rank(get<1>(*this), islower(c));
+        return *this;
+    }
+    //!\}
+
+    /*!\name Read functions
+     * \{
+     */
+    //!\brief Return a character. This reads the internal sequence letter.
```
h-2: same here

include/seqan3/alphabet/mask/masked.hpp
```diff
@@ -0,0 +1,118 @@
...
+    constexpr char_type to_char() const noexcept
+    {
+        if (seqan3::to_rank(get<1>(*this)))
+        {
+            return std::tolower(seqan3::to_char(get<0>(*this)));
```
h-2: I think there is a version of std::tolower that is local dependent, and one that is independent. Can you make sure this is the one that is independent or can you use one where you explicitly define the C local so that this function does what is expected also on weird platform settings?

include/seqan3/alphabet/mask/all.hpp
```diff
@@ -0,0 +1,56 @@
...
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL KNUT REINERT OR THE FU BERLIN BE LIABLE
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+// DAMAGE.
+
```
h-2: I think the copyright notices are missing their terminating "line"

test/unit/alphabet/mask/masked_test.cpp
```diff
@@ -0,0 +1,645 @@
...
+ #include <gtest/gtest.h>
+
+ #include <seqan3/alphabet/concept_pre.hpp>
+ #include <seqan3/alphabet/mask/all.hpp>
+ #include <seqan3/alphabet/nucleotide/dna4.hpp>
+ #include <seqan3/alphabet/aminoacid/aa20.hpp>
+
+ using namespace seqan3;
+
+ /************** TUPLE INHERITANCE **********************/
```
h-2: can you add `masked<dna5>` to the generic alphabet test, too?
alphabet_test.cpp:58 and alphabet_test.cpp:295

h-2:
Did you address my previous comments? (they are not marked as resolved, but maybe a gihub issue , too)

joshuak94:
Yeah, I addressed them. But I think it's because I messed up the rebase and had to get Svenja's help, somehow the commits got a little messed up.

Josh

include/seqan3/alphabet/mask/masked.hpp
h-2: It still looks like this wasn't changed :wink:

h-2:
Good job :tada:

PR #311, opened 2018-07-26
[FIX] Fixes template inspection for non template source types.

rrahn:
Fixes the following issue: 
When using our template inspection with non-template types the compiler will emit an error, rather than having a negative result.

### DIFF
--- a/include/seqan3/core/metafunction/template_inspection.hpp
+++ b/include/seqan3/core/metafunction/template_inspection.hpp
@@ -170,7 +170,8 @@ using transfer_template_vargs_onto_t = typename transfer_template_vargs_onto<sou
  * \tparam source_type      The source type.
  * \tparam target_template  The type template you wish to compare against (must take only types as template arguments).
  * \ingroup metafunction
- * \see seqan3::detail::is_value_specialisation_of_v
+ * \see seqan3::detail::is_value_specialisation_of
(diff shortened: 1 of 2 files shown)


include/seqan3/core/metafunction/template_inspection.hpp
```diff
@@ -176,6 +187,16 @@ using transfer_template_vargs_onto_t = typename transfer_template_vargs_onto<sou
...
+{};
+
+/*!\brief Helper variable template for seqan3::detail::is_type_specialisation_of.
+ * \tparam source_type      The source type.
+ * \tparam target_template  The type template you wish to compare against (must take only types as template arguments).
+ * \ingroup metafunction
+ */
+template <typename source_t, template <typename ...> typename target_template>
 constexpr bool is_type_specialisation_of_v =
     std::is_same_v<source_t, transfer_template_args_onto_t<source_t, target_template>>;
```
h-2: If you make it a full typetrait, I would make the shortcut refer to the base class to not duplicate the logic. And while you are at it, please add the `inline` keyword, i.e. 
```cpp
template <typename source_t, template <typename ...> typename target_template>
inline constexpr bool is_type_specialisation_of_v = is_specialisation_of<source_t, target_template>::value;
```

rrahn:
@h-2 done

PR #313, opened 2018-07-31
collected corrections for sequence_io

joergi-w:
When I implemented structure I/O, I found some issues with the sequence I/O, which are addressed in this PR.


include/seqan3/io/sequence/sequence_file_in_options.hpp
```diff
@@ -51,9 +51,6 @@ namespace seqan3
 template <typename sequence_legal_alphabet_, bool seq_qual_combined>
 struct sequence_file_in_options
 {
-    //!\brief Export the (possibly larger) legal alphabet to the format.
-    using sequence_legal_alphabet = sequence_legal_alphabet_;
-
```
h-2: If you remove this, please rename the template-parameter so that the trailing `_` is also removed.

include/seqan3/io/sequence/sequence_file_out.hpp
```diff
@@ -459,49 +459,6 @@ class sequence_file_out
...
-        requires tuple_like_concept<record_t> &&
-                 requires { requires detail::is_type_specialisation_of_v<remove_cvref_t<record_t>, record>; }
-    {
-        write_record(detail::get_or_ignore<field::SEQ>(r),
-                     detail::get_or_ignore<field::ID>(r),
-                     detail::get_or_ignore<field::QUAL>(r),
-                     detail::get_or_ignore<field::SEQ_QUAL>(r));
-
-    }
-
```
h-2: Why did you remove this? The overload below it does something else semantically:
It guesses the association values ←→ field-ID based on the current selection of the file. But for seqan::records we don't need to guess, we can just ask the record.
Maybe the tests still pass after this changed because the record and file configuration always matched. But they don't need to! Instead please create an example that illustrates this behaviour 😄

joergi-w: > Why did you remove this?

When this function is present, then all `push_back_tuple` tests compile just for the reason that in your tests you use only tuples of size 2 (seq, id). The call `push_back()` **always** executes the `push_back(record)` function, even if you pass a tuple. Compilation fails, whenever the tuple size is not 2, because record expects exactly 2 template args (it uses the 2 tuple types). 

To force the use of `push_back(tuple)`, which works for tuples of arbitrary size and records as well, I deleted the `push_back(record)` function. It also affects the structure_io code, where I usually have tuples of size 3 or 4, and fastq format, where you query also quality as a 3rd type. 

Can you please give me a hint how to fix that in a better way?

h-2: > The call push_back() always executes the push_back(record) function, even if you pass a tuple. 

Why? There clearly is a check that prevents this:
```
requires { requires detail::is_type_specialisation_of_v<remove_cvref_t<record_t>, record>; }
```

Also `detail::get_or_ignore` with a field as template parameter obviously isn't defined for a regular tuple so this could never compile.

> Compilation fails, whenever the tuple size is not 2, because record expects exactly 2 template args 

Both functions take one template argument so how is the size of the tuple related to anything?

> To force the use of push_back(tuple), which works for tuples of arbitrary size and records as well, 

As I said above, it **does not work** for records that contain a different field composition than the file. Please see this example from the documentation:
https://github.com/seqan/seqan3/blob/master/include/seqan3/io/sequence/sequence_file_out.hpp#L182
We need a test case that tests this behaviour.

> Can you please give me a hint how to fix that in a better way?

What is broken from your point of view?

joergi-w: I guess I understand now; please check the two additional tests.

test/unit/io/sequence/sequence_file_out_test.cpp
```diff
@@ -330,6 +330,27 @@ TEST(row, emplace_back)
...
+
+    std::string const expected_out
+    {
+        "> Test2\n"
+        "AGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGG\n"
+        "CTGNAGGCTGN\n"
+    };
+    EXPECT_EQ(fout.get_stream().str(), expected_out);
+}
+
```
h-2: :+1:

test/unit/io/sequence/sequence_file_out_test.cpp
```diff
@@ -330,6 +330,27 @@ TEST(row, emplace_back)
     });
 }
 
+TEST(row, different_fields)
```
h-2: maybe a little more verbose: `different_fields_in_record_and_file` and two comment lines describing the situation?

test/unit/io/sequence/sequence_file_out_test.cpp
```diff
@@ -405,6 +426,34 @@ TEST(rows, assign_sequence_file_pipes)
...
+        "> ID2\n"
+        "TATTA\n"
+    };
+
+    auto fout = sequence_file_in{std::istringstream{fastq_in}, sequence_file_format_fastq{}} |
+                sequence_file_out{std::ostringstream{}, sequence_file_format_fasta{}};
+    fout.get_stream().flush();
+    EXPECT_EQ(fout.get_stream().str(), fasta_out);
+}
+
```
h-2: I think this is covered by the previous test or not? Format specific things should not be tested here, but in the format tests...

test/unit/io/sequence/sequence_file_out_test.cpp
joergi-w: Okay, I added a comment and changed the name.

test/unit/io/sequence/sequence_file_out_test.cpp
joergi-w: Here I disagree, as it tests the sequence_file_out ability of converting file formats rather than fastq specific things. The fastq format test does not know about sequence_file_out. 
It is not covered in the previous test, because there the format and fields are the same; it rather extends it.

PR #320, opened 2018-08-02
[DOC] Introduce namespace seqan3::align.

rrahn:
fixes #320

### DIFF
--- a/include/seqan3/alignment/all.hpp
+++ b/include/seqan3/alignment/all.hpp
@@ -1,0 +1,49 @@
+// ============================================================================
+//                 SeqAn - The Library for Sequence Analysis
+// ============================================================================
+//
+// Copyright (c) 2006-2018, Knut Reinert & Freie Universitaet Berlin
+// Copyright (c) 2016-2018, Knut Reinert & MPI Molekulare Genetik
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//     * Redistributions of source code must retain the above copyright
+//       notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//     * Neither the name of Knut Reinert or the FU Berlin nor the names of
+//       its contributors may be used to endorse or promote products derived
+//       from this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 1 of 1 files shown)


include/seqan3/alignment/all.hpp
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * \todo Write detailed landing page.
+ */
+
+/*!\namespace seqan3::align
+ * \brief The SeqAn3 namespace for alignment related functions and classes.
+ *
+ * SeqAn3 implements various algorithms that are specific to bioinformatics and uses special configurations
+ * for these. In order to avoid collisions of configurations or classes and functions all alignment specific
+ * functions, concepts, configurations etc. are contained in there own namespace.
```
cpockrandt: their

include/seqan3/alignment/all.hpp
```diff
@@ -0,0 +1,61 @@
...
+// ============================================================================
+
+/*!\file
+ * \brief Meta-header for the \link align alignment module \endlink.
+ * \author Rene Rahn <rene.rahn AT fu-berlin.de>
+ */
+
+#pragma once
+
+/*!\defgroup align Alignment
```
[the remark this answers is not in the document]
h-2: Also the module names right now only differ in case from the folder names and I would strongly suggest to keep it that way so that the module hierarchy in the documentation corresponds to the file/folder structure.

include/seqan3/alignment/all.hpp
```diff
@@ -0,0 +1,61 @@
...
+ * \brief Contains alignment functionality.
+ *
+ * The align module contains concepts, functions and classes that are related to computation of
+ * pairwise and multiple sequence alignments.
+ *
+ * \todo Write detailed landing page.
+ */
+
+/*!\namespace seqan3::align
```
h-2: Let's discuss tomorrow if we want to have this as a general thing for the alignment module or only for the configs. If we use it only for the configs, I would also move it to whichever submodule the configs are in.

include/seqan3/alignment/all.hpp
```diff
@@ -0,0 +1,61 @@
...
+ * \brief The SeqAn3 namespace for alignment related functions and classes.
+ *
+ * SeqAn3 implements various algorithms that are specific to bioinformatics and uses special configurations
+ * for these. In order to avoid collisions of configurations or classes and functions all alignment specific
+ * functions, concepts, configurations etc. are contained in their own namespace.
+ *
+ * \attention
+ * This means you cannot use them, unless you explicitly add `using namespace seqan3::align;` (in addition
+ * to `using namespace seqan3;`).
```
h-2: I added this paragraph to the `seqan3::literal` namespace because there actually is no way to use the literals without the using decleration, but I don't think we need to add this for every sub-namespace. It should be clear that it's members can be used without an using decleration, simply by prepending `align::`.

rrahn:
@h-2 I think I addressed all points. As discussed we will have a special config namespace where only the config elements are contained.

PR #346, opened 2018-08-02
Feature/global align configuration

### LINKED ISSUE #319: Implement `align::global_config`
Create configuration for global alignment.
Based on #295

Configuration object for global alignment. 
Following config elements are supported:
`with_gaps`, `with_band`, `with_result`, `with_scoring`, `with_free_ends`

Implemented  by #346 

This needs to be reworked based on the discussion about the design decision.

rrahn:
@marehr and @cpockrandt so you can see how the usage is meant to be in combination with the configuration module.

Fixes #319 

Blocked by #295.
Please only focus on the last commit.

### DIFF
--- a/include/seqan3/alignment/configuration/global_config.hpp
+++ b/include/seqan3/alignment/configuration/global_config.hpp
@@ -1,0 +1,121 @@
+// ============================================================================
+//                 SeqAn - The Library for Sequence Analysis
+// ============================================================================
+//
+// Copyright (c) 2006-2018, Knut Reinert & Freie Universitaet Berlin
+// Copyright (c) 2016-2018, Knut Reinert & MPI Molekulare Genetik
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//     * Redistributions of source code must retain the above copyright
+//       notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//     * Neither the name of Knut Reinert or the FU Berlin nor the names of
+//       its contributors may be used to endorse or promote products derived
+//       from this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL KNUT REINERT OR THE FU BERLIN BE LIABLE
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+// DAMAGE.
+//
+// ============================================================================
+
+/*!\file
+ * \brief Provides seqan3::align::global_config
+ * \author Rene Rahn <rene.rahn AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <seqan3/core/algorithm/all.hpp>
+#include <seqan3/core/metafunction/basic.hpp>
+
+namespace seqan3::detail
+{
+
+// ----------------------------------------------------------------------------
+// global_alignment_cfg
+// ----------------------------------------------------------------------------
+
+//!\brief The global_alignment_configuration element to distinguish between alignment policies.
+class global_alignment_cfg : public config_element_base<global_alignment_cfg>
+{
+    //!\brief Grant access to private state.
+    friend class config_element_access<global_alignment_cfg>;
+
+    //!\brief State is ignored.
+    remove_cvref_t<decltype(std::ignore)> state;
+};
+}  // namespace seqan3::detail
+
+namespace seqan3::align
+{
+
+/*!\brief Global alignment configuration object.
+ * \ingroup align
+ */
+template <seqan3::detail::config_element_concept ... cfg_elements_t>
+class global_config : public detail::configuration<cfg_elements_t...>
+{
+    //!\brief The base type.
+    using base_type = detail::configuration<cfg_elements_t...>;
+
+public:
+    /*!\name Constructor, destructor and assignment
+     * \{
+     */
+    global_config()                                  = delete;
+    global_config(global_config const &)             = default;
+    global_config(global_config &&)                  = default;
+    global_config & operator=(global_config const &) = default;
+    global_config & operator=(global_config &&)      = default;
+    ~global_config()                                 = default;
+
+    //TODO(rrahn): Using the intermediate concept form here causes an ICE!
+    //!\brief Construction from another seqan3::detail::configuration object.
+    template <typename ... other_configs_t>
+    //!\cond
+        requires (detail::config_element_concept<other_configs_t> && ...)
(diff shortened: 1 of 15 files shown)


include/seqan3/alignment/configuration/global_config.hpp
```diff
@@ -0,0 +1,121 @@
...
+        requires (detail::config_element_concept<other_configs_t> && ...)
+    //!\endcond
+    global_config(detail::configuration<other_configs_t...> const & other) : base_type{other}
+    {}
+
+    //TODO(rrahn): Using the intermediate concept form here causes an ICE!
+    //!\brief Construction from another seqan3::detail::configuration object.
+    template <typename ... other_configs_t>
+    //!\cond
+        requires (detail::config_element_concept<other_configs_t> && ...)
```
h-2: why don't you just enforce this concept in `detail::config_element_concept` so you don't need to do it here everywhere?

include/seqan3/alignment/configuration/global_config.hpp
```diff
@@ -0,0 +1,121 @@
...
+//!\brief Deduces template arguments from constructing with another seqan3::detail::configuration.
+template <seqan3::detail::config_element_concept ... cfg_elements_t>
+global_config(detail::configuration<cfg_elements_t...> const &) ->
+    global_config<detail::global_alignment_cfg, cfg_elements_t...>;
+
+//!\brief Deduces template arguments from constructing with another seqan3::detail::configuration.
+template <seqan3::detail::config_element_concept ... cfg_elements_t>
+global_config(detail::configuration<cfg_elements_t...> &&) ->
+    global_config<detail::global_alignment_cfg, cfg_elements_t...>;
+//!\}
```
h-2: Shouldn't the previous one be sufficient?

include/seqan3/core/algorithm/concept.hpp
```diff
@@ -0,0 +1,150 @@
...
+ * \implements seqan3::detail::config_element_base
+ *
+ * \details
+ *
+ * Classes must extend the abstract crtp-base class seqan3::detail::config_element_base in order to satisfy this
+ * concept. This concept is merely used for internal purposes and shall not be exposed in public user interfaces.
+ */
+//!\cond
+template <typename config_t>
+concept bool config_element_concept = default_constructible_concept<std::remove_reference_t<config_t>> &&
```
h-2: Are you sure you don't want semi-regular or regular instead?

include/seqan3/core/algorithm/concept.hpp
```diff
@@ -0,0 +1,150 @@
...
+ * this concept. This concept is merely used for internal purposes and shall not be exposed in public user interfaces.
+ */
+//!\cond
+template <typename config_t>
+concept bool deferred_config_element_concept =
+    std::is_base_of_v<deferred_config_element_base<std::remove_reference_t<config_t>>, std::remove_reference_t<config_t>> &&
+    config_element_concept<config_t> &&
+    invocable_concept<std::remove_reference_t<config_t>,
+                      decltype(_callable_dummy),
+                      detail::configuration<std::remove_reference_t<config_t>>>;
```
h-2: when would this be false?

include/seqan3/core/algorithm/config_element_access.hpp
```diff
@@ -0,0 +1,116 @@
...
+class config_element_base;
+//!\endcond
+
+//!\cond
+template <typename derived_t>
+class deferred_config_element_base;
+//!\endcond
+
+/*!\brief An attorney class to grant the crtp-base classes (seqan3::detail::config_element_base and seqan3::detail::deferred_config_element_base)
+ *        access to the members of there derived classes.
```
h-2: Why don't the specialisations just befriend the CRTP-base?

include/seqan3/core/algorithm/configuration.hpp
```diff
@@ -0,0 +1,1071 @@
...
+ * auto bar_state = get<bar>(my_cfg);
+ * ```
+ *
+ * In addition every algorithm specific config implementation can overload the get-interface with a enum based
+ * get-interface, such that the correct config is extracted based on some type-independent enum value.
+ * This is necessary for the algorithm implementation, as the algorithm neither knows the order nor the exact type of
+ * all configs (the config type could be a template class).
+ */
+template <config_element_concept ... configs_t>
+class configuration
```
h-2: Why don't you inherit from std::tuple (or seqan3::pod_tuple)? Then you don't have to reimplement the get interfaces...

h-2: Also you can just use things like std::tuple_cat to recombine config with new config_elements

include/seqan3/core/algorithm/configuration.hpp
```diff
@@ -0,0 +1,1071 @@
...
+constexpr auto operator|(configuration_t && cfg,
+                         fn_t && fn)
+//!\cond
+    requires is_algorithm_configuration_v<remove_cvref_t<configuration_t>> &&
+             (is_type_specialisation_of_v<remove_cvref_t<fn_t>, configuration_fn_proxy> ||
+              std::is_base_of_v<configuration_fn_base<remove_cvref_t<fn_t>>, remove_cvref_t<fn_t>>)
+//!\endcond
+{
+    return fn(std::forward<configuration_t>(cfg));
+}
```
h-2: why not friend of the CRTP-base

include/seqan3/core/algorithm/configuration.hpp
```diff
@@ -0,0 +1,1071 @@
...
+{
+    return seqan3::get<target_t>(std::move(cfg));
+}
+
+//!\brief Provides std::get interface.
+template <typename target_t, typename ... configs_t>
+constexpr auto const && get(seqan3::detail::configuration<configs_t...> const && cfg) noexcept
+{
+    return seqan3::get<target_t>(std::move(cfg));
+}
```
h-2: As I said above, I think this code could be simplified dramatically by inheriting from std::tuple or seqan3::pod_tuple or am I missing something?

include/seqan3/alignment/configuration/global_config.hpp
rrahn: Can you please elaborate?

include/seqan3/alignment/configuration/global_config.hpp
rrahn: If I think about it, one could also have the one ctor taking by copy and always move. Requires all config elements to be movable. I think we can assume that. As long as std::reference_wrapper is movable. Have to check

include/seqan3/core/algorithm/concept.hpp
rrahn: Not sure. Actually I want to drop the default constructible requirement to pass config states per std::ref

include/seqan3/core/algorithm/concept.hpp
rrahn: If any of the conditions is false?
But of course since I require the base of I can drop the imvocable. Ah that’s what you meant. Well, I wasn’t sure about having a concept requiring a base class of the type🙄🤖. Then again it is only to use forwarding references in internal function declarations

include/seqan3/core/algorithm/config_element_access.hpp
rrahn: Because friendship is not inheritable and deferred_config_base and config_base must befriend both with derived_t

include/seqan3/core/algorithm/configuration.hpp
rrahn: I had a hell of issues with the get interface  when inheriting it from std::tuple. Also I don’t think that it will reduce much of the complexity but probably the push_front logic can be replaced by tuple_cat. Look into it

include/seqan3/core/algorithm/configuration.hpp
rrahn: Told you. Constraint friend templates won’t work.

include/seqan3/core/algorithm/configuration.hpp
rrahn: I am not sure it will reduce code dramatically. Actually not sure to expose the positional get Interface at all because of some other issues.  But let’s discuss it tomorrow

include/seqan3/alignment/configuration/global_config.hpp
h-2: I meant in `detail::configuration` sorry.

include/seqan3/alignment/configuration/global_config.hpp
h-2: But these are not constructors, they are deduction guides, and since both deduce to the same type, one should be sufficient (independent of anything else).

include/seqan3/core/algorithm/concept.hpp
h-2: Many styles recommend for library types to always be semi_regular if possible. It makes many things easier, it's the style of range-v3 and I think they want to adopt it for the core guidelines, too.

include/seqan3/core/algorithm/concept.hpp
h-2: yeah, I was referring to the invocable, it's confusing me :laughing:

rrahn:
Outdated after original design has been changed.

PR #357, opened 2018-08-09
Feature/scoring scheme

h-2:
A first working version of the scoring schemes. 

TODOs:
  - [x] some things missing in (developer) documentation
  - [x] constructors in the documentation
  - [x] examples in the documentation
  - [x] tests for amino acid similarity matrixes and serialisation
  - [x] proper integration with `alignment/data_structure/all.hpp` (depends on stuff by @smehringer )
  - [x] strong type parameters for simple scheme

### DIFF
--- a/include/seqan3/alignment/data_structure/all.hpp
+++ b/include/seqan3/alignment/data_structure/all.hpp
@@ -1,0 +1,51 @@
+// ============================================================================
+//                 SeqAn - The Library for Sequence Analysis
+// ============================================================================
+//
+// Copyright (c) 2006-2018, Knut Reinert & Freie Universitaet Berlin
+// Copyright (c) 2016-2018, Knut Reinert & MPI Molekulare Genetik
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//     * Redistributions of source code must retain the above copyright
+//       notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//     * Neither the name of Knut Reinert or the FU Berlin nor the names of
+//       its contributors may be used to endorse or promote products derived
+//       from this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL KNUT REINERT OR THE FU BERLIN BE LIABLE
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+// DAMAGE.
+//
+// ============================================================================
+
+/*!\file
+ * \brief Meta-header for the \link data_structure Data Structure sub-module \endlink.
+ * \author Hannes Hauswedell <hannes.hauswedell AT fu-berlin.de>
+ */
+
+#pragma once
+
+/*!\defgroup data_structure Data Structure
+ * \brief Provides the data structures used in alignment computation.
+ * \ingroup alignment
+ *
+ * \todo Write detailed landing page.
+ */
+
+#include <seqan3/alignment/data_structure/aminoacid_scoring_scheme.hpp>
+#include <seqan3/alignment/data_structure/nucleotide_scoring_scheme.hpp>
+#include <seqan3/alignment/data_structure/scoring_scheme_base.hpp>
--- a/include/seqan3/alignment/data_structure/aminoacid_scoring_scheme.hpp
+++ b/include/seqan3/alignment/data_structure/aminoacid_scoring_scheme.hpp
@@ -1,0 +1,355 @@
+// ============================================================================
+//                 SeqAn - The Library for Sequence Analysis
+// ============================================================================
+//
+// Copyright (c) 2006-2018, Knut Reinert & Freie Universitaet Berlin
+// Copyright (c) 2016-2018, Knut Reinert & MPI Molekulare Genetik
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//     * Redistributions of source code must retain the above copyright
+//       notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//     * Neither the name of Knut Reinert or the FU Berlin nor the names of
+//       its contributors may be used to endorse or promote products derived
+//       from this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL KNUT REINERT OR THE FU BERLIN BE LIABLE
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+// DAMAGE.
+//
+// ============================================================================
+
+/*!\file
+ * \brief Provides seqan3::aminoacid_scoring_scheme.
+ * \author Hannes Hauswedell <hannes.hauswedell AT fu-berlin.de>
+ * \author Rene Rahn <rene.rahn AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <range/v3/algorithm/copy.hpp>
+
+#include <seqan3/alignment/data_structure/scoring_scheme_base.hpp>
+#include <seqan3/alphabet/aminoacid/aa27.hpp>
+
+namespace seqan3
+{
+
+/*!\brief Identifiers for amino acid similarity matrixes.
+ * \ingroup data_structure
+ *
+ * \details
+ *
+ * This enum provides IDs for amino acid similarity matrixes of the
+ * [BLOSUM](https://en.wikipedia.org/wiki/BLOSUM) and [PAM](https://en.wikipedia.org/wiki/Point_accepted_mutation)
+ * families.
+ *
+ * \see seqan3::aminoacid_scoring_scheme::set_similarity_matrix
+ */
+enum class aminoacid_similarity_matrix
+{
+    //ATTENTION: when you change this, also update set_similarity_matrix() below
+    /*!\brief The BLOSUM30 matrix for very distantly related proteins.
+     * \details
+     * \snippet include/seqan3/alignment/data_structure/aminoacid_scoring_scheme.hpp matrix_data blosum30
+     */
+    BLOSUM30,
+    /*!\brief The BLOSUM45 matrix for distantly related proteins.
+     * \details
+     * \snippet include/seqan3/alignment/data_structure/aminoacid_scoring_scheme.hpp matrix_data blosum45
+     */
+    BLOSUM45,
+    /*!\brief The BLOSUM62 matrix recommended for most use-cases.
+     * \details
+     * \snippet include/seqan3/alignment/data_structure/aminoacid_scoring_scheme.hpp matrix_data blosum62
+     */
+    BLOSUM62,
+    /*!\brief The BLOSUM80 matrix for closely related proteins.
+     * \details
+     * \snippet include/seqan3/alignment/data_structure/aminoacid_scoring_scheme.hpp matrix_data blosum80
+     */
+    BLOSUM80
+};
+
+/*!\brief A data structure for managing and computing the score of two amino acids.
+ * \ingroup data_structure
+ *
+ * \details
+ *
+ * You can use an instance of this class to score two amino acids. The amino acids need not be of the same type.
+ * Different scoring behaviour can be set via the member functions.
+ *
+ * ### Example
+ *
+ * Score two letters:
+ * ```cpp
+ * aminoacid_scoring_scheme scheme{aminoacid_similarity_matrix::BLOSUM62};
+ * std::cout << scheme.get_score(aa27::T, aa27::S); // == 1
+ *
+ * scheme.set_similarity_matrix(aminoacid_similarity_matrix::BLOSUM80);
+ * std::cout << scheme.get_score(aa27::T, aa20::S); // == 2
+ * // you can score aa20 against aa27        ^
+ *
+ * scheme.set_scheme_hamming();
+ * std::cout << scheme.get_score(aa27::T, aa20::S); // == -1
+ * std::cout << scheme.get_score(aa27::T, aa20::T); // == 0
+ * ```
+ *
+ * You can "edit" a given matrix directly:
+ * ```cpp
+ * scheme.set_similarity_matrix(aminoacid_similarity_matrix::BLOSUM80);
+ * std::cout << scheme.get_score(aa27::T, aa27::S); // == 2
+ * auto & cell = scheme.get_score(aa27::T, aa27::S);
+ * cell = 3;
+ * std::cout << scheme.get_score(aa27::T, aa27::S); // == 3
+ * ```
+ *
+ * Score two sequences:
+ * ```cpp
+ * std::vector<aa27> one = "ALIGATOR"_aa27;
+ * std::vector<aa27> two = "ANIMATOR"_aa27;
+ *
+ * aminoacid_scoring_scheme scheme{aminoacid_similarity_matrix::BLOSUM62};
+ * int score = 0;
+ * for (auto pair : ranges::view::zip(one, two))
+ *     score += scheme.get_score(std::get<0>(pair), std::get<1>(pair));
+ *
+ * ```
+ */
+template <arithmetic_concept score_type = int8_t>
+class aminoacid_scoring_scheme : public scoring_scheme_base<aminoacid_scoring_scheme<score_type>, aa27, score_type>
+{
+private:
+    //!\brief Type of the CRTP-base.
+    using base_t = scoring_scheme_base<aminoacid_scoring_scheme<score_type>, aa27, score_type>;
+
+    //!\brief Make the base's private member visible.
+    using base_t::matrix;
+public:
+    //!\privatesection
+    //!\brief Inherit the base class's constructors.
+    using base_t::base_t;
+    //!\brief Make the base's private member type visible.
+    using typename base_t::matrix_t;
+    //!\publicsection
+
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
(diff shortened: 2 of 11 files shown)


include/seqan3/alignment/data_structure/scoring_scheme_base.hpp
```diff
@@ -0,0 +1,220 @@
...
+    {
+        set_scheme_hamming();
+    }
+
+    /*!\brief Constructor for the simple scheme (delegates to set_scheme_simple()).
+     * \copydetails set_scheme_simple()
+     */
+    constexpr scoring_scheme_base(match_score const ms, mismatch_score const mms) noexcept
+    {
+        set_scheme_simple(ms, mms);
```
rrahn: We could add an enum `unknown_policy` to treat matching against unknowns differently. 
Either `::matches_unknown`, `::matches_all`, `::matches_nothing`.

include/seqan3/alignment/data_structure/scoring_scheme_base.hpp
```diff
@@ -0,0 +1,220 @@
...
+    {
+        for (matrix_size_t i = 0; i < matrix_size; ++i)
+            for (matrix_size_t j = 0; j < matrix_size; ++j)
+                matrix[i][j] = (i == j) ? static_cast<score_t>(ms) : static_cast<score_t>(mms);
+    }
+
+    /*!\brief Set a custom scheme by passing a full matrix with arbitrary content.
+     * \param[in] _matrix A full matrix that is copied into the scheme.
+     */
+    constexpr void set_scheme_custom(matrix_t const & _matrix) noexcept
```
rrahn: In general I think we should call the function just `set_scheme` and use overload resolution to pick the correct behaviour.

include/seqan3/alignment/data_structure/scoring_scheme_base.hpp
```diff
@@ -0,0 +1,220 @@
...
+#include <seqan3/core/concept/cereal.hpp>
+
+#if SEQAN3_WITH_CEREAL
+#include <cereal/types/array.hpp>
+#endif // SEQAN3_WITH_CEREAL
+
+namespace seqan3
+{
+
+/*!\brief A strong type of underlying type `int8_t` that represents the score two matching characters.
```
smehringer: .. the score `of` two ..

include/seqan3/alignment/data_structure/scoring_scheme_base.hpp
```diff
@@ -0,0 +1,220 @@
...
+struct match_score : detail::strong_type<int8_t, match_score, detail::strong_type_skill::convert>
+{
+     using detail::strong_type<int8_t, match_score, detail::strong_type_skill::convert>::strong_type;
+};
+
+/*!\brief A strong type of underlying type `int8_t` that represents the score two different characters.
+ * \ingroup data_structure
+ * \see scoring_scheme_base::set_scheme_simple
+ */
+struct mismatch_score : detail::strong_type<int8_t, mismatch_score, detail::strong_type_skill::convert>
```
smehringer: Maybe `mismatch_penalty` instead of `score` since a negative value is expected?

include/seqan3/alignment/data_structure/scoring_scheme_base.hpp
```diff
@@ -0,0 +1,220 @@
...
+/*!\brief A CRTP base class for scoring schemes.
+ * \tparam  derived_t The derived type.
+ * \tparam alphabet_t Type of the largest target alphabet.
+ * \ingroup data_structure
+ *
+ * \details
+ *
+ * This type is never used directly, instead use seqan3::nucleotide_scoring_scheme or seqan3::aminoacid_scoring_scheme.
+ */
+template <typename derived_t, alphabet_concept alphabet>
```
smehringer: alphabet`_t` :nail_care:

include/seqan3/alignment/data_structure/scoring_scheme_base.hpp
```diff
@@ -0,0 +1,220 @@
...
+    constexpr void set_scheme_custom(matrix_t const & _matrix) noexcept
+    {
+        ranges::copy(_matrix, ranges::begin(matrix));
+    }
+    //!\}
+
+    /*!\name Accessors
+     * \{
+     */
+    /*!\brief Score two letters (either two nucleotids or two amino acids).
```
smehringer: > (either two nucleotids or two amino acids)

Other alphabets (modelling the alphabet concept) should be also fine or not?

test/unit/alignment/scoring_scheme_test.cpp
```diff
@@ -0,0 +1,275 @@
...
+#include <seqan3/core/type_list.hpp>
+
+#if SEQAN3_WITH_CEREAL
+#include <seqan3/test/tmp_filename.hpp>
+
+#include <fstream>
+
+#include <cereal/archives/binary.hpp>
+#include <cereal/archives/portable_binary.hpp>
+#include <cereal/archives/json.hpp>
```
smehringer: alphabetical ordering :nail_care:

include/seqan3/alignment/data_structure/aminoacid_scoring_scheme.hpp
```diff
@@ -0,0 +1,321 @@
...
+ * ### Example
+ *
+ * Score two letters:
+ * ```cpp
+ * aminoacid_scoring_scheme scheme{aminoacid_similarity_matrix::BLOSUM62};
+ * std::cout << scheme.get_score(aa27::T, aa27::S); // == 1
+ *
+ * scheme.set_scheme_similarity_matrix(aminoacid_similarity_matrix::BLOSUM80);
+ * std::cout << scheme.get_score(aa27::T, aa20::S); // == 2
+ * // you can score aa20 against aa27        ^
```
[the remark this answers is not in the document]
h-2: Yes :)

include/seqan3/alignment/data_structure/aminoacid_scoring_scheme.hpp
```diff
@@ -0,0 +1,321 @@
...
+    //!\publicsection
+
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    //NOTE(h-2): unfortunately these are not inherited in the documentation
+    //!\copydoc scoring_scheme_base::scoring_scheme_base()
+    SEQAN3_DOXYGEN_ONLY(( aminoacid_scoring_scheme() {} ))
+    //!\copydoc scoring_scheme_base::scoring_scheme_base(match_score const ms, mismatch_score const mms)
+    SEQAN3_DOXYGEN_ONLY(( aminoacid_scoring_scheme(match_score const ms, mismatch_score const mms) {} ))
```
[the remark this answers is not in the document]
h-2: It's not used often, but I wouldn't say it's meaning-less. Especially when we introduce reduced alphabets, there might be use-cased.

include/seqan3/alignment/data_structure/scoring_scheme_base.hpp
```diff
@@ -0,0 +1,220 @@
...
+#endif // SEQAN3_WITH_CEREAL
+
+namespace seqan3
+{
+
+/*!\brief A strong type of underlying type `int8_t` that represents the score two matching characters.
+ * \ingroup data_structure
+ * \see scoring_scheme_base::set_scheme_simple
+ */
+struct match_score : detail::strong_type<int8_t, match_score, detail::strong_type_skill::convert>
```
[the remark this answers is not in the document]
h-2: > We need them parameterized as also floating point scores will be needed.

image

include/seqan3/alignment/data_structure/scoring_scheme_base.hpp
h-2: I don't know what you mean by this and I didn't get it from your files either.

include/seqan3/alignment/data_structure/scoring_scheme_base.hpp
h-2: I prefer @smehringer 's proposal to just drop `_scheme` from the function names.

include/seqan3/alignment/data_structure/aminoacid_scoring_scheme.hpp
```diff
@@ -0,0 +1,321 @@
...
+ * std::cout << scheme.get_score(aa27::T, aa20::S); // == -1
+ * std::cout << scheme.get_score(aa27::T, aa20::T); // == 1
+ * ```
+ *
+ * You can "edit" a given matrix directly:
+ * ```cpp
+ * scheme.set_scheme_similarity_matrix(aminoacid_similarity_matrix::BLOSUM80);
+ * std::cout << scheme.get_score(aa27::T, aa27::S); // == 2
+ * auto & cell = scheme.get_score(aa27::T, aa27::S);
+ * cell = 3;
```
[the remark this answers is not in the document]
h-2: Yeah, well the setter functions set the whole scheme. And modifying an existing scheme by position is not used often I assume. So I think using references for this is a good way.

include/seqan3/alignment/data_structure/scoring_scheme_base.hpp
```diff
@@ -0,0 +1,220 @@
...
+     * \attention These functions are never called directly, see \ref serialisation for more details.
+     */
+    template <cereal_archive_concept archive_t>
+    void CEREAL_SERIALIZE_FUNCTION_NAME(archive_t & archive)
+    {
+        archive(matrix);
+    }
+    //!\endcond
+
+private:
```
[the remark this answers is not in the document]
h-2: there is a a friend decleration further up!

include/seqan3/alignment/data_structure/scoring_scheme_base.hpp
h-2: No, we shouldn't mix the terms. Either we use "score" **OR** we "cost" or "penalty". But mixing these is confusing because it implies inverting the sign (a penalty of 3 is a score of -3). If we have both we need to add two separate functions (`set_score` and `set_penalty`).

include/seqan3/alignment/data_structure/scoring_scheme_base.hpp
h-2: No, we only have these two and the scpecialisations inherit this documentation so I wanted to point out that it is either two of nucl or two of aa, but not one of each or random alphabets.

include/seqan3/alignment/data_structure/scoring_scheme_base.hpp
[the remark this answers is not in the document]
h-2: Your suggestion for solving this with type deduction guides really isn't feasible since our underlying type is `int8_t`, but `match_score{11}` resolves to `match_score<int>{11}` (and that of course is not convertible to `match_score<int8_t>` because strong types) so we would need to have our users call
```cpp
scheme1.set_simple_scheme(match_score<int8_t>{11}, mismatch_score<int8_t>{-7});
```
 which is ridiculously complex and error prone.
The same goes for deduction in the constructor.

The only thing I can think of would be to add a skill to the strong types that allows strong type of some integral type to be converted to another. But I think that is also error-prone (even if optional) and defeats the purpose of strong types.

I don't see a way to resolve this cleanly and am afraid we nee to drop strong types from the interface if the underlying type really needs to be configurable. For the concrete interface I think it would be ok since we are dealing with only two parameters and no sane person would think of writing the mismatch score before the match score. [although in general of course I would like to use strong types very much]

include/seqan3/alignment/data_structure/scoring_scheme_base.hpp
smehringer: So we will also call the gap config `gap_open_score` and `gap_extend_score` ?

include/seqan3/alignment/data_structure/scoring_scheme_base.hpp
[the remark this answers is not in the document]
rrahn: First, there is no rational to enforce the underlying score to be int8_t. We had use-cases where the score is set to values beyond the value range of int8_t to express for example to never allow mismatches.  The only thing we do know is, that the specific similarity matrices don't need a higher type and int8_t is for them a good choice. So the score type should be deduced from them, which we can do.
I really think we should keep the score type dependent of the user input. And well, in 99.9% the user just doesn't care about the specific type and it will be sufficient to just assume int/double for the underlying type so they will never have to write the expression from your comment. But, in case you really want to specify the type you can use regular c++ template declaration to do so. I honestly don't see the problem with that and how this interferes with strong types.
That would also allow us to use floating point scoring matrices for amino acids, if applications might need this. I mean at the moment I only know use-cases for rna/dna related scoring schemes.

include/seqan3/alignment/data_structure/aminoacid_scoring_scheme.hpp
```diff
@@ -0,0 +1,321 @@
...
+    SEQAN3_DOXYGEN_ONLY(( aminoacid_scoring_scheme(matrix_t const & _matrix) {} ))
+    //!\}
+
+    /*!\name Scheme selection
+     * \{
+     */
+    /*!\brief Set the similarity matrix scheme (e.g. BLOSUM62).
+     * \param[in] matrix_id The enum ID of the matrix, see seqan3::aminoacid_similarity_matrix.
+     */
+    constexpr void set_scheme_similarity_matrix(aminoacid_similarity_matrix const matrix_id)
```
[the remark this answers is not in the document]
rrahn: I don't like: `set_simple`. And I further think, that from the arguments, thanks to strong_types a simple `set_scheme` is very expressive.

include/seqan3/alignment/data_structure/scoring_scheme_base.hpp
```diff
@@ -0,0 +1,281 @@
...
+ */
+template <typename derived_t, alphabet_concept alphabet_t, arithmetic_concept score_type>
+class scoring_scheme_base
+{
+public:
+    /*!\name Member types
+     * \{
+     */
+    //!\brief Type of the score values.
+    using score_t = score_type;
```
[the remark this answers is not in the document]
h-2: you don't need it. the only type useful to users is the matrix (for `set_custom_matrix`) and possibly the types that the matrix depends on.

include/seqan3/alignment/data_structure/scoring_scheme_base.hpp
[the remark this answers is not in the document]
h-2: > I really think we should keep the score type dependent of the user input. And well, in 99.9% the user just doesn't care about the specific type and it will be sufficient to just assume int/double for the underlying type so they will never have to write the expression from your comment.

No, I think you didn't get my point. The strong types in the function interface are fixed for every scheme once the scheme is defined. Unless we a do a "pre-selection" like above, you will get conflicts, because numbers always deduce to int (unless postfixed by e.g. `l`) and this is incompatible with type expected.

> That would also allow us to use floating point scoring matrices for amino acids, if applications might need this. I mean at the moment I only know use-cases for rna/dna related scoring schemes.

Whether or not aminoacids can be floats is entirely independent of strong types. I just didn't templatise it, because I didn't see the need.

include/seqan3/alignment/data_structure/aminoacid_scoring_scheme.hpp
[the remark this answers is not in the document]
h-2: It's called `set_simple_scheme` right now...

h-2:
I have addressed all comments, implemented the requested changes, added tests for the constructors and added type deduction guides + tests.

h-2:
@rrahn And all tests :point_up:

include/seqan3/alignment/data_structure/aminoacid_scoring_scheme.hpp
```diff
@@ -0,0 +1,355 @@
...
+        {-2,  0, -7,  1,  6, -6, -4,  0, -6, -6,  1, -5, -3, -1, -1, -2,  5,  0, -1, -2, -1, -4, -5, -4,  6, -1, -8},//Y
+        {-1, -3, -4, -3, -2, -3, -3, -2, -2, -2, -2, -2, -2, -2, -2, -3, -2, -2, -1, -1, -2, -2, -5, -3, -1, -2, -8},//Z
+        {-8, -8, -8, -8, -8, -8, -8, -8, -8, -8, -8, -8, -8, -8, -8, -8, -8, -8, -8, -8, -8, -8, -8, -8, -8, -8,  1} //*
+        //! [matrix_data blosum80]
+    }};
+
+    //TODO(h-2): add more matrixes
+};
+
+/*!\name Template argument type deduction guides
```
rrahn: In cpp reference it is only called `Type Deduction`. 
I used it as well. I don't care much, but we should have it consistent.

include/seqan3/alignment/data_structure/nucleotide_scoring_scheme.hpp
```diff
@@ -0,0 +1,137 @@
...
+
+#pragma once
+
+#include <seqan3/alignment/data_structure/scoring_scheme_base.hpp>
+#include <seqan3/alphabet/nucleotide/dna15.hpp>
+
+namespace seqan3
+{
+
+/*!\brief An type for computing the score of two nucleotides.
```
rrahn: Like above: A scoring/similarity matrix?

rrahn: Just, it is clear that it is a type.

include/seqan3/alignment/data_structure/nucleotide_scoring_scheme.hpp
```diff
@@ -0,0 +1,137 @@
...
+    //!\copydoc scoring_scheme_base::scoring_scheme_base(matrix_t const & _matrix)
+    SEQAN3_DOXYGEN_ONLY(( constexpr nucleotide_scoring_scheme(matrix_t const & _matrix) noexcept {} ))
+    //!\}
+
+    //!\privatesection
+    //!\brief Inherit the base class's constructors.
+    using base_t::base_t;
+};
+
+/*!\name Template argument type deduction guides
```
rrahn: Same here: We need a unified description.

include/seqan3/alignment/data_structure/scoring_scheme_base.hpp
```diff
@@ -0,0 +1,279 @@
...
+    {
+        std::conditional_t<integral_concept<score_t>, int64_t, double> i_ms = static_cast<score_arg_t>(ms);
+        std::conditional_t<integral_concept<score_t>, int64_t, double> i_mms = static_cast<score_arg_t>(mms);
+        if ((i_ms  < std::numeric_limits<score_t>::lowest() || i_ms  > std::numeric_limits<score_t>::max()) ||
+            (i_mms < std::numeric_limits<score_t>::lowest() || i_mms > std::numeric_limits<score_t>::max()))
+        {
+            throw std::invalid_argument{"You passed a score value to set_simple_scheme that is out of range of the "
+                                        "scoring scheme's underlying type. Define your scoring scheme with a larger "
+                                        "template parameter; or construct match_score explicitly with a template "
+                                        "parameter to silence this warning."};
```
rrahn: Mhmm, it is not just a warning, but an exception causing a hard fail if it is not catched. So I would remove the last part of the last sentence.

include/seqan3/std/concept/core_language.hpp
```diff
@@ -90,11 +92,17 @@ concept bool common_reference_concept =             static_cast<bool>(ranges::Co
...
+ */
+template <typename t>
+concept bool arithmetic_concept =                   std::is_arithmetic_v<t>;
+
 /*!\brief Resolves to `ranges::Integral<type>()`
  * \sa http://en.cppreference.com/w/cpp/experimental/ranges/concepts/Integral
  */
 template <typename t>
-concept bool integral_concept =                     static_cast<bool>(ranges::Integral<t>());
+concept bool integral_concept =                     arithmetic_concept<t> && static_cast<bool>(ranges::Integral<t>());
```
rrahn: why not using `std::is_integral_v`?

rrahn: It would produce a better error message, as the constraint failure is not somewhere within the ranges library.

test/unit/alignment/scoring_scheme_test.cpp
```diff
@@ -0,0 +1,389 @@
...
+                                              nucleotide_scoring_scheme<float>,
+                                              aminoacid_scoring_scheme<>,
+                                              aminoacid_scoring_scheme<int>>;
+
+TYPED_TEST_CASE(generic, scoring_scheme_types);
+
+
+TEST(nucleotide_scoring_scheme, template_argument_deduction)
+{
+    {
```
rrahn: And maybe test the default construction without brace-initialisation.

test/unit/alignment/scoring_scheme_test.cpp
```diff
@@ -0,0 +1,389 @@
...
+        std::array<std::array<int16_t, 15>, 15> m;
+        nucleotide_scoring_scheme scheme{m};
+        EXPECT_TRUE((std::is_same_v<decltype(scheme), nucleotide_scoring_scheme<int16_t>>));
+    }
+}
+
+TEST(aminoacid_scoring_scheme, template_argument_deduction)
+{
+    {
+        aminoacid_scoring_scheme scheme{};
```
rrahn: same here

include/seqan3/alignment/data_structure/all.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+/*!\file
+ * \brief Meta-header for the \link data_structure Data Structure sub-module \endlink.
+ * \author Hannes Hauswedell <hannes.hauswedell AT fu-berlin.de>
+ */
+
+#pragma once
+
+/*!\defgroup data_structure Data Structure
+ * \brief Provides the data structures used in alignment computation.
```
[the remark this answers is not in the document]
h-2: :thinking:

include/seqan3/alignment/data_structure/aminoacid_scoring_scheme.hpp
h-2: Ok, I have grepped our repo and apparently "Type deduction guide(s)" is the most abundant name so I will use that, too, ok?

include/seqan3/alignment/data_structure/nucleotide_scoring_scheme.hpp
h-2: I think many people don't assume that a scoring scheme for nucleotides is actually a matrix. I have renamed it to "A data structure for managing and computing the score of two nucleotides.", ok?

include/seqan3/alignment/data_structure/scoring_scheme_base.hpp
h-2: I have changed the text so that it doesn't suggest a warning, but I think it still doesn't hurt to tell users how to circumvent this.

include/seqan3/std/concept/core_language.hpp
h-2: There is a bunch of concepts that are not yet implemented natively and not documented properly. I didn't want to do clean-up in the same PR as features.

rrahn: Ah ok, I missed, that the integral_concept was just extended.

PR #363, opened 2018-08-10
Fix minor issues and typos

### LINKED ISSUE #358: Collection of Minor issues and typos.
- [ ] File https://github.com/seqan/seqan3/blob/master/include/seqan3/alphabet/gap/all.hpp has no `#pragma once`

- [ ] The documentation group opened here has no closing bracket : https://github.com/seqan/seqan3/blob/master/include/seqan3/std/concept/core_language.hpp#L153

- [ ] The record code example is wrong (order of types and get according function) : http://docs.seqan.de/seqan/3.0.0-master-user/structseqan3_1_1record.html

- [ ] The `write_eol` function in IO has a wierd documentation as "\n" is displayed as an actual new line instead of the character: http://docs.seqan.de/seqan/3.0.0-master-dev/group__io.html#gaf387d6832c3e03849af06740694ada9c

- [ ] In many files the date is still `2016-2017` and needs to be updated.

eseiler:
This resolves #358

### DIFF
--- a/include/seqan3/std/concept/core_language.hpp
+++ b/include/seqan3/std/concept/core_language.hpp
@@ -162,6 +162,7 @@ concept bool nothrow_assignable_concept =           assignable_concept<t1, t2> &
  * \attention This is a concept requirement, not an actual function (however types satisfying this concept
  * will provide an implementation).
  */
+//!\}
 //!\cond
 template <typename t>
 concept bool swappable_concept =               requires (t && v1, t && v2) { std::swap(v1, v2); } ||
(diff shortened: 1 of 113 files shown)


include/seqan3/std/concept/core_language.hpp
```diff
@@ -150,7 +150,6 @@ concept bool nothrow_assignable_concept =           assignable_concept<t1, t2> &
  */
 /*!\name Requirements for seqan3::swappable_concept
  * \brief You can expect these functions on all types that implement seqan3::swappable_concept.
- * \{
```
h-2: instead of removing this, please add a closing brace after the next element

h-2:
Thanks for fixing all the stuff, one minor thing!

include/seqan3/std/concept/core_language.hpp
h-2: `//!\}`

PR #364, opened 2018-08-12
[FEATURE] Equality comparators for qualified composition against either of it's inner types.

### LINKED ISSUE #360: qualified class should be comparable to either of their inner types
Say you read a FASTQ file into a `quality_composiyion<dna4, phred42>` for convenient storage and access and now you want to parse the sequence and query a position for an `dna4::A`. This is currently not possible because clearly the types are different. It would be a handy feature to have equality comparators to compare the composition character value to another alphabet character while ignoring the attached quality value and vice versa.

smehringer:
Fixes #360

### DIFF
--- a/include/seqan3/alphabet/quality/qualified.hpp
+++ b/include/seqan3/alphabet/quality/qualified.hpp
@@ -98,6 +98,12 @@ struct qualified :
     public cartesian_composition<qualified<sequence_alphabet_t, quality_alphabet_t>,
                                  sequence_alphabet_t, quality_alphabet_t>
 {
+private:
+    //!\brief The base type
+    using base_type = cartesian_composition<qualified<sequence_alphabet_t, quality_alphabet_t>,
+                                            sequence_alphabet_t, quality_alphabet_t>;
+
+public:
     //!\brief First template parameter as member type.
     using sequence_alphabet_type = sequence_alphabet_t;
     //!\brief Second template parameter as member type.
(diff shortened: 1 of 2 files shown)


include/seqan3/alphabet/quality/qualified.hpp
```diff
@@ -173,4 +173,134 @@ template <typename sequence_alphabet_type, typename quality_alphabet_type>
 qualified(sequence_alphabet_type &&, quality_alphabet_type &&)
     -> qualified<std::decay_t<sequence_alphabet_type>, std::decay_t<quality_alphabet_type>>;
 
+/*!\name Comparison operators for the qualified composition and it's inner alphabet type.
```
h-2: Please implement them as member functions.

include/seqan3/alphabet/quality/qualified.hpp
```diff
@@ -173,4 +173,134 @@ template <typename sequence_alphabet_type, typename quality_alphabet_type>
...
+ * std::cout << (q2 == dna4::A) << std::endl; // 1 (true)
+ * std::cout << (q2 == dna4::T) << std::endl; // 0 (false)
+ *```
+ *
+ */
+//!\brief Returns `true` if \p std::get<alphabet_type>(lhs) == \p rhs, `false` otherwise.
+template<alphabet_concept alphabet_type, quality_concept quality_type>
+constexpr bool operator==(qualified<alphabet_type, quality_type> const & lhs, alphabet_type const & rhs)
+{
+    return std::get<alphabet_type>(lhs) == rhs;
```
h-2: Please don't use the type-based-std::get as potentially you could have two quality alphabets in the composition now and the get wouldn't know which one to take. Use the composition's member types instead.

include/seqan3/alphabet/quality/qualified.hpp
```diff
@@ -173,4 +173,134 @@ template <typename sequence_alphabet_type, typename quality_alphabet_type>
...
+}
+
+//!\brief Returns `true` if \p std::get<alphabet_type>(lhs) != \p rhs, `false` otherwise.
+template<alphabet_concept alphabet_type, quality_concept quality_type>
+constexpr bool operator!=(qualified<alphabet_type, quality_type> const & lhs, alphabet_type const & rhs)
```
h-2: all operators need to be `noexcept` too!

include/seqan3/alphabet/quality/qualified.hpp
eseiler: its

include/seqan3/alphabet/quality/qualified.hpp
```diff
@@ -173,4 +173,134 @@ template <typename sequence_alphabet_type, typename quality_alphabet_type>
...
+/*!\name Comparison operators for the qualified composition and it's inner alphabet type.
+ *\{
+ * \tparam alphabet_type The alphabet type of the qualified composition. Must
+ *                       model the seqan3::alphabet_concept.
+ * \tparam quality_type  The alphabet type of the qualified composition. Must
+ *                       model the seqan3::quality_type.
+ *
+ * This enables the comparison of a qualified composition value against a value
+ * of the inner alphabet type while ignoring it's quality score. The operators
+ * delegate the comparison operators of the alphabet type.
```
eseiler: delegate **to**  ? 🤔

include/seqan3/alphabet/quality/qualified.hpp
```diff
@@ -173,4 +173,134 @@ template <typename sequence_alphabet_type, typename quality_alphabet_type>
...
+
+//!\brief Returns `true` if \p std::get<alphabet_type>(lhs) >= \p rhs, `false` otherwise.
+template<alphabet_concept alphabet_type, quality_concept quality_type>
+constexpr bool operator>=(qualified<alphabet_type, quality_type> const & lhs, alphabet_type const & rhs)
+{
+    return std::get<alphabet_type>(lhs) >= rhs;
+}
+//!\}
+
+/*!\name Comparison operators for the qualified composition and it's inner quality type.
```
eseiler: its

include/seqan3/alphabet/quality/qualified.hpp
```diff
@@ -173,4 +173,134 @@ template <typename sequence_alphabet_type, typename quality_alphabet_type>
...
+ *\{
+ * \tparam alphabet_type The alphabet type of the qualified composition. Must
+ *                       model the seqan3::alphabet_concept.
+ * \tparam quality_type  The alphabet type of the qualified composition. Must
+ *                       model the seqan3::quality_type.
+ *
+ * This enables the comparison of a qualified composition value against a value
+ * of the inner quality type while ignoring it's alphabet character value.
```
eseiler: its

include/seqan3/alphabet/quality/qualified.hpp
```diff
@@ -173,4 +173,134 @@ template <typename sequence_alphabet_type, typename quality_alphabet_type>
...
+ * The operators delegate the comparison operators of the quality type.
```
eseiler: delegate **to**  ? 🤔

smehringer:
@h-2 

implementing the comparators as member functions forced me to implement the default comparison (qualified vs qualified). I wonder how the arithmetic comparison should be handled since it is ambiguous?
Consider this:
```
qualified<dna4, phred42> t0{dna4::C, 6};
qualified<dna4, phred42> t1{dna4::A, 7};
```
where `std::get<0>(t0) > std::get<0>(t1)` but `std::get<1>(t0) < std::get<1>(t1)`.

The behaviour right now (by a default comparison of the compiler) is the following:
```
t0 < t1  // false
t0 <= t1 // false
t0 == t1 // false
t0 >= t1 // true
t0 > t1  // true
```
which obviously only compares the character. Do we want this behaviour?

joergi-w:
> which obviously only compares the character. 

It is important that different quality characters lead to `!=`, even if the nucleotide character is equal... My suggestion is a behaviour  similar to std::pair comparison (decide on dna4; if dna4 is equal, decide on phred42).

include/seqan3/alphabet/quality/qualified.hpp
```diff
@@ -165,6 +165,270 @@ struct qualified :
...
+    {
+        return !(*this > rhs);
+    }
+
+    //!\brief Returns `true` if `*this` is not strictly smaller then \p rhs, `false` otherwise.
+    constexpr bool operator>=(qualified<sequence_alphabet_t, quality_alphabet_t> const & rhs) const noexcept
+    {
+        return !(*this < rhs);
+    }
+    //!\}
```
h-2: You don't need to add these, they are implemented in the base class, already!

include/seqan3/alphabet/quality/qualified.hpp
```diff
@@ -165,6 +165,270 @@ struct qualified :
...
+     */
+    //!\brief Returns `true` if \p std::get<0>(*this) == \p rhs, `false` otherwise.
+    constexpr bool operator==(sequence_alphabet_t const & rhs) const noexcept
+    {
+        return std::get<0>(*this) == rhs;
+    }
+
+    //!\brief Returns `true` if \p std::get<0>(rhs) == \p lhs, `false` otherwise.
+    friend constexpr bool operator==(sequence_alphabet_t const & lhs,
+                                     qualified<sequence_alphabet_t, quality_alphabet_t> const & rhs)
```
h-2: friends should be noexcept, too!

h-2: > `qualified<sequence_alphabet_t, quality_alphabet_t>`

You can write just `qualified` :nail_care:

include/seqan3/alphabet/quality/qualified.hpp
```diff
@@ -165,6 +165,270 @@ struct qualified :
...
+        return std::get<0>(*this) >= rhs;
+    }
+
+    //!\brief Returns `true` if \p std::get<0>(rhs) >= \p lhs, `false` otherwise.
+    friend constexpr bool operator>=(sequence_alphabet_t const & lhs,
+                                     qualified<sequence_alphabet_t, quality_alphabet_t> const & rhs)
+    {
+        return rhs >= lhs;
+    }
+    //!\}
```
h-2: I usually write the non-firends in one block and then the friends, but I guess it you could argue for either style...

include/seqan3/alphabet/quality/qualified.hpp
```diff
@@ -165,6 +165,270 @@ struct qualified :
...
+    constexpr bool operator>=(sequence_alphabet_t const & rhs) const noexcept
+    {
+        return std::get<0>(*this) >= rhs;
+    }
+
+    //!\brief Returns `true` if \p std::get<0>(rhs) >= \p lhs, `false` otherwise.
+    friend constexpr bool operator>=(sequence_alphabet_t const & lhs,
+                                     qualified<sequence_alphabet_t, quality_alphabet_t> const & rhs)
+    {
+        return rhs >= lhs;
```
h-2: I think you friends are not doing the correct comparisons :thinking:

PR #365, opened 2018-08-13
reduce complexity of RNA I/O

joergi-w:
omit concept checks with nested templates


h-2:
Please don't remove the tests! they are important and I thought it was only the concept checks that are expensive?

joergi-w:
Sure, let's try with omitting concept checks only. That already showed a significant memory reduction.

PR #368, opened 2018-08-14
[feature] start sam io

MitraDarja:
Start of the sam io
there are still some TODOs open and I think more tests are necessary to check if it really works. Specifically, a test case with actual qualities are necessary.


include/seqan3/io/sequence/sequence_file_format_sam.hpp
```diff
@@ -0,0 +1,352 @@
...
+              sequence_file_in_options<seq_legal_alph_type> const & options,
+              seq_type                                            & sequence,
+              id_type                                             & id,
+              qual_type                                           & qualities,
+              seq_qual_type                                       & seq_qual)
+    {
+        static_assert(detail::decays_to_ignore_v<seq_type> || detail::decays_to_ignore_v<seq_qual_type>,
+                      "Either the sequence field, or the seq_qual field need to be set to std::ignore.");
+        static_assert(detail::decays_to_ignore_v<qual_type> || detail::decays_to_ignore_v<seq_qual_type>,
+                      "Either the qualities field, or the seq_qual field need to be set to std::ignore.");
```
h-2: The stuff in sequence io has changed a little so that you don't need to worry about `seq_qual_type` too much anymore. But you need to adapt your code. Have a look at how the fastq code looks now!

include/seqan3/io/sequence/sequence_file_format_sam.hpp
```diff
@@ -0,0 +1,352 @@
...
+        auto stream_view = view::subrange<decltype(std::istreambuf_iterator<char>{stream}),
+                                          decltype(std::istreambuf_iterator<char>{})>
+                            {std::istreambuf_iterator<char>{stream},
+                             std::istreambuf_iterator<char>{}};
+        auto stream_it = ranges::begin(stream_view);
+        if(*stream_it != '@') { //Ignore comment lines
```
h-2: braces always go to newline in SeqAn

include/seqan3/io/sequence/sequence_file_format_sam.hpp
```diff
@@ -0,0 +1,352 @@
...
+          // make sure "buffer at end" implies "stream at end"
+          if ((std::istreambuf_iterator<char>{stream} == std::istreambuf_iterator<char>{}) &&
+              (!stream.eof()))
+          {
+              stream.get(); // triggers error in stream and sets eof
+          }
+        }
+        else{
+          while(*stream_it != '\n')
+          {++stream_it;}
```
h-2: instead do `detail::consume(stream_view | view::take_line_or_throw);` This makes sure an end-of-line is actually reached (throws an exception otherwise).

include/seqan3/io/sequence/sequence_file_format_sam.hpp
```diff
@@ -0,0 +1,352 @@
...
+    {
+        static_assert(detail::decays_to_ignore_v<seq_type> || detail::decays_to_ignore_v<seq_qual_type>,
+                      "Either the sequence field, or the seq_qual field need to be set to std::ignore.");
+        static_assert(detail::decays_to_ignore_v<qual_type> || detail::decays_to_ignore_v<seq_qual_type>,
+                      "Either the qualities field, or the seq_qual field need to be set to std::ignore.");
+        ranges::ostreambuf_iterator stream_it{stream};
+        // ID
+        if constexpr (detail::decays_to_ignore_v<id_type>)
+        {
+            throw std::logic_error{"The ID field may not be set to ignore when writing sam files."};
```
h-2: Are you sure? I think they can be set to `*` and the file will still be valid, or not?

include/seqan3/io/sequence/sequence_file_format_sam.hpp
```diff
@@ -0,0 +1,352 @@
...
+        }
+        else
+        {
+            if (ranges::empty(id)) //[[unlikely]]
+                throw std::runtime_error{"The ID field may not be empty when writing sam files."};
```
h-2: same here, if the file is still valid when the ID is set to `*` to that instead.

include/seqan3/io/sequence/sequence_file_format_sam.hpp
```diff
@@ -0,0 +1,352 @@
...
+            stream_it = '0'; //PNEXT
+            stream_it = '\t';
+            stream_it = '0'; //TLEN
+            stream_it = '\t';
+        }
+        // Sequence
+        if constexpr (!detail::decays_to_ignore_v<seq_type>) // sequence
+        {
+            if (ranges::empty(sequence)) //[[unlikely]]
+                throw std::runtime_error{"The SEQ field may not be empty when writing sam files."};
```
h-2: same here, I think the field could be `*`

include/seqan3/io/sequence/sequence_file_format_sam.hpp
```diff
@@ -0,0 +1,352 @@
...
+            write_seq(stream_it, options, seq_qual);
+        }
+        else
+        {
+            throw std::logic_error{"The SEQ and SEQ_QUAL fields may not both be set to ignore when writing sam files."};
+        }
+        // Qualities
+        if constexpr (!detail::decays_to_ignore_v<qual_type>) // qual
+        {
+            write_qual(stream_it, options, qualities);
```
h-2: check if empty and write `*` if so

include/seqan3/io/sequence/sequence_file_format_sam.hpp
```diff
@@ -0,0 +1,352 @@
...
+                                 | ranges::view::drop_while(!is_tab)
+                                 | ranges::view::drop_while(is_tab)
+                                 | ranges::view::drop_while(!is_tab)
+                                 | ranges::view::drop_while(is_tab)
+                                 | ranges::view::drop_while(!is_tab)
+                                 | ranges::view::drop_while(is_tab)
+                                 | ranges::view::take_while(!is_tab)                // until next
+                                 | ranges::view::drop_while(is_tab)
+                                 | ranges::view::remove_if(is_space || is_digit)   // ignore whitespace and numbers
+                                 | view::transform([is_legal_alph] (char const c)
```
h-2: Does this work? It shouldn't! The view doesn't actually start doing anything until you start iterating over it...

Instead do multiple `detail::consume(stream_view | ranges::view::take_while(is_tab);` beforehand

include/seqan3/io/sequence/sequence_file_format_sam.hpp
```diff
@@ -0,0 +1,352 @@
...
+        for (auto it = ranges::begin(seq_view); it != ranges::end(seq_view); ++it)
+        {}
+    }
+    template <typename stream_view_t,
+              typename qual_type>
+    void read_qual(stream_view_t                                       & stream_view,
+                  qual_type                                            & qualities)
+    {
+        auto const is_end = is_char<'\n'>{} || is_char<'\t'>{};
+        ranges::copy(stream_view | ranges::view::take_while(!is_char<'\n'>{} ), //TODO: What if we have tags?
```
h-2: `!is_space` maybe?

include/seqan3/io/sequence/sequence_file_format_sam.hpp
h-2: > Does this work? It shouldn't! 

Actually it should work, but it will be slower than doing the consumes I mentioned above.

h-2:
Instead of `ranges::view::take_while` you can now use `view::take_until` or `view::take_until_or_throw`

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,346 @@
...
+          detail::consume(stream_view | ranges::view::take_while(!is_tab));
+          detail::consume(stream_view | ranges::view::take_while(is_tab));
+          detail::consume(stream_view | ranges::view::take_while(!is_tab));
+          detail::consume(stream_view | ranges::view::take_while(is_tab));
+          detail::consume(stream_view | ranges::view::take_while(!is_tab));
+          detail::consume(stream_view | ranges::view::take_while(is_tab));
+
+
+
+          auto seq_view = stream_view | ranges::view::remove_if(is_free || is_digit)   // ignore whitespace and numbers
```
smehringer: @h-2 I see that skipping whitespaces and numbers is also done when reading fasta/fastq so I am adressing you: Shouldn't the format rather throw an invalid-char error when a number occurrs whithin the sequence? 
@MitraDarja you commented to skip whitespaces but `is_free` checks (only) for `\r`? (I am also not sure if whitespaces should be allowed in SAM format)

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,346 @@
...
+          if constexpr (seq_qual_combined)
+          {
+              // seq_qual field implies that they are the same variable
+              assert(std::addressof(sequence) == std::addressof(qualities));
+              ranges::copy(stream_view | view::char_to<typename value_type_t<qual_type>::quality_alphabet_type>,
+                           ranges::begin(qualities) + sequence_size_before);
+          }
+          else if constexpr (!detail::decays_to_ignore_v<qual_type>)
+          {
+            ranges::copy(stream_view | view::take_until_or_throw(is_cntrl) //TODO: What if we have tags?
```
smehringer: open todo.
> What if we have tags?

then you have a tab instead of a new line so can use `take_until_or_throw(is_cntrl || is_tab)` 

but than make sure to consume the rest.

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,346 @@
...
+          if ((std::istreambuf_iterator<char>{stream} == std::istreambuf_iterator<char>{}) &&
+              (!stream.eof()))
+          {
+              stream.get(); // triggers error in stream and sets eof
+          }
+
+          // Jump over the new line character in the end
+          detail::consume(stream_view | ranges::view::take_while(is_cntrl));
+        }
+        else{
```
smehringer: bracket into new line

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,346 @@
...
+              typename id_type,
+              typename qual_type>
+    void write(stream_type                     & stream,
+               sequence_file_output_options const & options,
+               seq_type                       && sequence,
+               id_type                        && id,
+               qual_type                      && qualities)
+    {
+        ranges::ostreambuf_iterator stream_it{stream};
+
```
smehringer: Note that you use a lot of functionality for the types that might not compile if you e.g. give an integer to `seq_type`. Therefore the types should be restricted with concepts. With @h-2 I discussed that this should be done as static asserts to give the user a helpful error message.
```
static_assert(std::ForwardRange<seq_type> && alphabet_concept<value_type_t<seq_type>>, 
              "The sequence must model std::ForwardRange and its value type must model seqan3::alphabet_concept. ")

File: test/unit/io/sequence_file/sequence_file_format_sam_test.cpp [hunk 14]
```diff
@@ -0,0 +1,285 @@
...
+TEST_F(read, newline_before_eof)
+{
+    std::string input
+    {
+      "@ Comment\n"
+      "ID1\t0\tBABABA\t200\t0\t*\tBABABA\t0\t0\tACGTTTTTTTTTTTTTTT\t!##$%&'()*+,-./++-\n"
+      "ID2\t0\t*\t0\t0\tBABA\t*\t30\t0\tACGTTTTTTTTTTTTT\rTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTT"
+      "TTTTT\t!##$&'()*+,-./+)*+,-)*+,-)*+,-)*+,BDEBDEBDEBDEBDEBDEBDEBDEBDEBDEBDEBDEBDEBDEBDEBDE\n"
+      "ID3 lala\t0\t*\t0\t0\tHAHAHA+\t*\t0\t0\tACGTTTA\t!!!!!!!\n"
+      "\n"
```
smehringer: is this a valid sam format with a newline in the end?

test/unit/io/sequence_file/sequence_file_format_sam_test.cpp
```diff
@@ -0,0 +1,285 @@
...
+            EXPECT_EQ(qual,expected_quals[i]);
+        }
+    }
+};
+TEST_F(read, standard)
+{
+    std::string input
+    {
+        "@ Comment\n"
+        "ID1\t0\t*\t0\t0\t*\t*\t0\t0\tACGTTTTTTTTTTTTTTT\t!##$%&'()*+,-./++-\n"
```
smehringer: add a tag in the end `\tAS:i:3\n` to check if it is also accepted

test/unit/io/sequence_file/sequence_file_format_sam_test.cpp
```diff
@@ -0,0 +1,285 @@
...
+{
+    std::string comp
+    {
+      "TEST 1\t0\t*\t0\t0\t*\t*\t0\t0\tACGT\t*\n"
+      "Test2\t0\t*\t0\t0\t*\t*\t0\t0\tAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGC"
+      "TGNAGGCTGN\t*\n"
+      "Test3\t0\t*\t0\t0\t*\t*\t0\t0\tGGAGTATAATATATATATATAT\t*\n"
+    };
+      for (unsigned i = 0; i < 3; ++i)
+        quals[i] = { "*"_phred42 };
```
smehringer: This is cheating :wink:. The test should have a write function that does not take any qualities and then automatically writes `*` for the quality fields!

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,346 @@
...
+   * All fields are required when writing.
+   *
+   * ### Implementation notes
+   *
+   */
+class sequence_file_format_sam
+{
+public:
+    /*!\name Constructors, destructor and assignment
+     * \brief Rule of five explicitly defaulted.
```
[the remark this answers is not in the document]
MitraDarja: I know not a good answer but I just copy&pasted this from format_fasta/fastaq, so does the brief there need to be changed as well?

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,346 @@
...
+        auto const is_tab = is_char<'\t'>;
+        auto const is_cntrl = is_char<'\n'>;
+        auto const is_free = is_char<'\r'>;
+        /* ID */
+        if (*stream_it != '@') //Ignore comment lines
+        {
+          // ID
+          if constexpr (!detail::decays_to_ignore_v<id_type>)
+          {
+            //TODO: Make truncate_ids also work if not all ids have a blank
```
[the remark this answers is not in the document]
MitraDarja: According to https://samtools.github.io/hts-specs/SAMv1.pdf, ids can consists of regexp of type [!-?A-~], so I would say no blanks. Should I throw an error if I encounter a blank or just ignore id? Should I check for writing beforehand that ids have no blank character?

If I see it correctly, there is also no blank in seq allowed, should I throw an error there as well?

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -136,18 +148,18 @@ class sequence_file_format_sam
             sequence_size_before = ranges::size(sequence);
 
         auto const is_tab = is_char<'\t'>;
-        auto const is_cntrl = is_char<'\n'>;
         auto const is_free = is_char<'\r'>;
+
+
+        while (*stream_it == '@')
```
smehringer: You can also use `is_char<'@'>(*stream_it)` here

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -160,40 +172,26 @@ class sequence_file_format_sam
...
-          detail::consume(stream_view | ranges::view::take_while(is_tab));
-          detail::consume(stream_view | ranges::view::take_while(!is_tab));
-          detail::consume(stream_view | ranges::view::take_while(is_tab));
-          detail::consume(stream_view | ranges::view::take_while(!is_tab));
-          detail::consume(stream_view | ranges::view::take_while(is_tab));
-          detail::consume(stream_view | ranges::view::take_while(!is_tab));
-          detail::consume(stream_view | ranges::view::take_while(is_tab));
-
-
+        for (int i = 0; i <8;i++)
```
smehringer: space before 8: `i < 8`

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -160,40 +172,26 @@ class sequence_file_format_sam
...
+        {
+            ++stream_it;
+            detail::consume(stream_view | ranges::view::take_while(!is_tab));
+        }
+        ++stream_it;
 
-          auto seq_view = stream_view | ranges::view::remove_if(is_free || is_digit)   // ignore whitespace and numbers
+        auto seq_view = stream_view
+        | ranges::view::remove_if(is_char<'\r'> || is_digit || is_char<'\v'> || is_char<'\f'>)
```
smehringer: whats `\f` and `\v`? :thinking:

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -331,13 +327,14 @@ class sequence_file_format_sam
         {
           throw std::logic_error{"The QUAL and SEQ_QUAL fields may not both be set to ignore when writing sam files."};
         }
+        else if (ranges::empty(qualities))
+        {
+          stream_it = '*';
+          stream_it = '\n';
```
smehringer: use `detail::write_eol(stream_it, options.add_carriage_return);` here too

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,343 @@
...
+            }
+            else
+            {
+              ranges::copy(stream_view | view::take_until_or_throw(is_tab)
+                                       | view::char_to<value_type_t<id_type>>,
+                                   ranges::back_insert_iterator{id});
+            }
+
+        }
+        else
```
smehringer: If the `if` statement has brackets, the `else` must too. Sorry for being picky, I learned that from h-2 :innocent:

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,343 @@
...
+                                          return c;
+                                      })
+                                    | view::char_to<value_type_t<seq_type>>,      // convert to actual target alphabet
+                           ranges::back_insert_iterator{sequence});
+
+              sequence_size_after = ranges::size(sequence);
+          }
+          else
+          {
+            auto it = ranges::begin(seq_view);
```
smehringer: indent 4 spaces not 2

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,343 @@
...
+          {
+              // seq_qual field implies that they are the same variable
+              assert(std::addressof(sequence) == std::addressof(qualities));
+              ranges::copy(stream_view | view::char_to<typename value_type_t<qual_type>::quality_alphabet_type>,
+                           ranges::begin(qualities) + sequence_size_before);
+          }
+          else if constexpr (!detail::decays_to_ignore_v<qual_type>)
+          {
+            ranges::copy(stream_view | view::take_until_or_throw(is_cntrl||is_tab)
+                        | view::char_to<value_type_t<qual_type>>,
```
smehringer: indent pipe `|` to align with the pipe above.

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,343 @@
...
+              typename qual_type>
+    void write(stream_type                     & stream,
+               sequence_file_output_options const & options,
+               seq_type                       && sequence,
+               id_type                        && id,
+               qual_type                      && qualities)
+    {
+      if constexpr (!(detail::decays_to_ignore_v<seq_type>))
+       {
+        static_assert(std::ranges::ForwardRange<seq_type> && alphabet_concept<value_type_t<seq_type>>,
```
smehringer: Missing indentation.
Also I think the surrounding brackets are not aligned to the `if`

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,343 @@
...
+            stream_it = '0'; //PNEXT
+            stream_it = '\t';
+            stream_it = '0'; //TLEN
+            stream_it = '\t';
+        }
+
+        // Sequence
+        if constexpr (detail::decays_to_ignore_v<seq_type>)
+        {
+          throw std::logic_error{"The SEQ and SEQ_QUAL fields may not both be set to ignore when writing sam files."};
```
smehringer: 4 indent instead of 2

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,347 @@
...
+        if constexpr (!detail::decays_to_ignore_v<seq_type>)
+            sequence_size_before = ranges::size(sequence);
+
+        auto const is_tab = is_char<'\t'>;
+        auto const is_free = is_char<'\r'>;
+
+        while (is_char<'@'>(*stream_it))
+        {
+            detail::consume(stream_view | view::take_line);
+            stream_it = ranges::begin(stream_view);
```
smehringer: why do you need to set the stream_it here while in line 186 (a similar loop) you do not?

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,347 @@
...
+          // Jump over all parts we are not interested in
+        for (int i = 0; i < 8;i++)
+        {
+            ++stream_it;
+            detail::consume(stream_view | ranges::view::take_while(!is_tab));
+        }
+        ++stream_it;
+
+        auto seq_view = stream_view
+        | ranges::view::remove_if(is_char<'\r'> || is_digit)
```
smehringer: put this line on the line above.

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,347 @@
...
+          }
+
+          // make sure "buffer at end" implies "stream at end"
+          if ((std::istreambuf_iterator<char>{stream} == std::istreambuf_iterator<char>{}) &&
+              (!stream.eof()))
+          {
+              stream.get(); // triggers error in stream and sets eof
+          }
+
+          if ((std::istreambuf_iterator<char>{stream} == std::istreambuf_iterator<char>{}) &&
```
smehringer: this code block is exactly the same as the one above (L 250-255) is it not? If so it can be removed.

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,347 @@
...
+        ranges::ostreambuf_iterator stream_it{stream};
+        // ID
+        if constexpr (detail::decays_to_ignore_v<id_type>)
+        {
+            throw std::logic_error{"The ID field may not be set to ignore when writing sam files."};
+        }
+        else
+        {
+            if (ranges::empty(id)) //[[unlikely]]
+                throw std::runtime_error{"The ID field may not be empty when writing sam files."};
```
smehringer: Actually, everything can be empty in the SAM format, so you can remove the exception and write out `*`

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,347 @@
...
+
+        // Sequence
+        if constexpr (detail::decays_to_ignore_v<seq_type>)
+        {
+            throw std::logic_error{"The SEQ and SEQ_QUAL fields may not both be set to ignore when writing sam files."};
+        }
+        else
+        {
+            if (ranges::empty(sequence)) //[[unlikely]]
+                throw std::runtime_error{"The SEQ field may not be empty when writing sam files."};
```
smehringer: Same as abouve. Replace exception by writing out `*`

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,347 @@
...
+            ranges::copy(sequence | view::to_char, stream_it);
+
+        }
+        stream_it = '\t';
+
+        // Quality line
+        if constexpr (detail::decays_to_ignore_v<qual_type>)
```
smehringer: please also remove the constexpr here because qual_type is allowed to be ignored.

include/seqan3/io/sequence_file/format_sam.hpp
MitraDarja: This is only necessary for g++7, in g++8 I would not need that line at all and I think the reason I need it here in g++7 but not in the line below is because here I jump over an entire line and need to keep stream_it updated, below I only jump over one letter at a time and adapt stream_it via ++stream_it.

include/seqan3/io/sequence_file/format_sam.hpp
MitraDarja: I am not sure what you mean by allowed to be ignored, I throw an error if it is set to ignore like I do when sequence is set to ignored. In fastaq for both ifs there is a constexpr and then an exception throw. Why is it different here? 
(Plus if I remove it I get a really long error messaged.)

MitraDarja:
@smehringer Did I oversee some of your requested changes or is this okay?

include/seqan3/io/sequence_file/format_sam.hpp
smehringer: then don't use the stream_it at all but directly ut `ranges::begin(stream_view)` into the while condition: `while (is_char<'@'>(*ranges::begin(stream_view)))`

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,343 @@
...
+        }
+        else
+        {
+             detail::consume(stream_view | ranges::view::take_while(!is_tab));
+        }
+
+          // Jump over all parts we are not interested in
+        for (int i = 0; i < 8;i++)
+        {
+            ++stream_it;
```
smehringer: Can you check if deleting this line changes anything? because I think it does not. The stream_iterator is invalidated as soon as you call `detail::consume`

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,343 @@
...
+            detail::consume(stream_view | ranges::view::take_while(!is_tab));
+        }
+        ++stream_it;
+
+        auto seq_view = stream_view | ranges::view::remove_if(is_char<'\r'> || is_digit)
+        // ignore whitespace and numbers
```
smehringer: Thos three lines should lool like this:
```cpp
auto seq_view = stream_view | ranges::view::remove_if(is_char<'\r'> || is_digit) // ignore whitespace and numbers
                            | view::take_until_or_throw(is_tab);                 // until next space
```

include/seqan3/io/sequence_file/format_sam.hpp
MitraDarja: It changes things, because then the tests fail and in my opinion that makes sense, the ++stream_it jumps over the tab, which is not consumed by detail::consume(stream_view | ranges::view::take_while(!is_tab));

MitraDarja:
@smehringer Did I miss something or all your requested changes included?

test/unit/io/sequence_file/sequence_file_format_sam_test.cpp
```diff
@@ -0,0 +1,281 @@
...
+#include <seqan3/alphabet/quality/all.hpp>
+#include <seqan3/io/sequence_file/input_format_concept.hpp>
+#include <seqan3/io/sequence_file/output_format_concept.hpp>
+#include <seqan3/io/sequence_file/format_sam.hpp>
+#include <seqan3/range/view/convert.hpp>
+
+using namespace seqan3;
+using namespace seqan3::literal;
+
+//TODO remove after #256 is merged
```
smehringer: remove this comment

test/unit/io/sequence_file/sequence_file_format_sam_test.cpp
```diff
@@ -0,0 +1,281 @@
...
+#include <range/v3/view/transform.hpp>
+
+#include <seqan3/alphabet/quality/all.hpp>
+#include <seqan3/io/sequence_file/input_format_concept.hpp>
+#include <seqan3/io/sequence_file/output_format_concept.hpp>
+#include <seqan3/io/sequence_file/format_sam.hpp>
+#include <seqan3/range/view/convert.hpp>
+
+using namespace seqan3;
+using namespace seqan3::literal;
```
smehringer: Remove this as the namespace does not exist anymore. I think if you rebase onto master you will notice those changes.

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,308 @@
...
+        auto constexpr is_cntrl= is_char<'\n'>;
+
+        while (is_char<'@'>(*ranges::begin(stream_view)))
+            detail::consume(stream_view | view::take_line);
+
+        /* ID */
+        std::string first; // Checking if id is given or set to *
+        ranges::copy(stream_view | view::take_until_or_throw(is_tab || is_blank), std::back_inserter(first));
+        if (first == "*")
+            throw parse_error{"The ID field may not be empty for SAM files."};
```
smehringer: This should ONLY be checked if the id_type does not decay to ignore!

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,308 @@
...
+
+        if constexpr (!detail::decays_to_ignore_v<id_type>)
+        {
+            ranges::copy(first | view::char_to<value_type_t<id_type>>, std::back_inserter(id)); //Add first to id
```
smehringer: Please do not use temporary strings. You can call begin() as often as you want without consuming the stream.
Replace the code by this
```c++
if(is_char<'*'>(*begin(stream_view)))
    throw parse_error{"The ID field may not be empty for SAM files."};

```

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,308 @@
...
+            detail::consume(stream_view | view::take_until_or_throw(is_tab));
+        }
+        ++begin(stream_view);
+
+        auto seq_view = stream_view | view::take_until_or_throw(is_tab); // until next tab
+
+         // Sequence
+         if constexpr (!detail::decays_to_ignore_v<seq_type>)
+         {
+             auto constexpr is_legal_alph = is_in_alphabet<seq_legal_alph_type>;
```
smehringer: don't you also need to check if the sequence is `*`?

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,308 @@
...
+              sequence_size_after = ranges::size(sequence);
+          }
+          else
+          {
+              for (auto it = begin(seq_view); it != end(seq_view); ++it)
+                  ++sequence_size_after;
+          }
+          ++begin(stream_view); // skip tab
+
+          std::string qual;
```
smehringer: Again do not use a temporary variable. Just query for `*` like for id and seq

test/unit/io/sequence_file/sequence_file_format_sam_test.cpp
```diff
@@ -0,0 +1,368 @@
...
+    };
+
+    sequence_file_format_sam format;
+    sequence_file_input_options<dna5, false> options;
+    std::string id;
+    dna5_vector seq;
+    std::vector<phred42> qual;
+    std::stringstream istream{input};
+
+    EXPECT_THROW((format.read(istream, options, seq, id, qual)), parse_error );
```
smehringer:
```suggestion
    EXPECT_THROW((format.read(istream, options, seq, std::ignore, std::ignore)), parse_error);
```

And then you can remove the variables. saves you some lines of code you do not need :) (check the no_id and no_wual as well)

include/seqan3/io/sequence_file/format_sam.hpp
MitraDarja: At the moment it will throw an parse_error{std::string{"Encountered an unexpected letter: "}. Do you think this error message is misleading? Then I will add another if before.

smehringer: Yes for consistency :)

include/seqan3/io/sequence_file/format_sam.hpp
MitraDarja: Another question about qualities: What if we have the quality length is greater than the sequence length? view::take_exactly_or_throw does accept this.

smehringer:
One more thing: Can you add one test that creates a `sequence_file` from a sam file stream and check if reading through the file works? I remember that I had to change something when adapting to the compression but I am not sure. The compiler will tell you.

Here is something you can probably copy and paste that needs only little adjustment from my alignment_file_tets:

```cpp
TEST(some_name, some_other_name)
{
    std::string input
    {
        "read1\t41\tref\t1\t61\t1S1M1D2M\tref\t10\t300\tACGT\t!##$\tAS:i:2\tNM:i:7\n"
        "read2\t42\tref\t2\t62\t7M1D1M1S\tref\t10\t300\tAGGCTGNAG\t!##$&'()*\txy:B:S,3,4,5\n"
        "read3\t43\tref\t3\t63\t1S1M1D4M1D1M1S\tref\t10\t300\tGGAGTATA\t!!*+,-./\n"
    };

    std::vector<dna5_vector> seq_comp
    {
        "ACGT"_dna5,
        "AGGCTGNAG"_dna5,
        "GGAGTATA"_dna5
    };

    std::vector<std::string> id_comp
    {
        "read1",
        "read2",
        "read3"
    };

    std::vector<std::vector<phred42>> qual_comp
    {
        { "!##$"_phred42 },
        { "!##$&'()*"_phred42 },
        { "!!*+,-./"_phred42 },
    };

    sequence_file_input fin{std::istringstream{input}, sequence_file_format_sam{}};

    size_t counter = 0;
    for (auto & [ id, seq, qual ] : fin)
    {
        EXPECT_TRUE((std::ranges::equal(seq, seq_comp[counter])));
        EXPECT_TRUE((std::ranges::equal(id,  id_comp[counter])));
        EXPECT_TRUE((std::ranges::equal(qual,  qual_comp[counter])));

        counter++;
    }

    EXPECT_EQ(counter, 3u);
}
```

MitraDarja:
@smehringer Is there a way to also write to a stream_file like to opposite of the code you wrote above? If so, should I add it?

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,283 @@
...
+
+    //!\copydoc sequence_file_input_format_concept::read
+    template <typename stream_type,     // constraints checked by file
+              typename seq_legal_alph_type, bool seq_qual_combined,
+              typename seq_type,        // other constraints checked inside function
+              typename id_type,
+              typename qual_type>
+    void read(stream_type                     & stream,
+              sequence_file_input_options<seq_legal_alph_type, seq_qual_combined> const & options,
+              seq_type                        & sequence,
```
smehringer: always align parameters at the `&`
```cpp
stream_type                                                               & stream,
sequence_file_input_options<seq_legal_alph_type, seq_qual_combined> const & options,
seq_type                                                                  & sequence,
```
check if this is the case everywhere

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,283 @@
...
+              qual_type                       & qualities)
+    {
+        auto stream_view = view::subrange<decltype(std::istreambuf_iterator<char>{stream}),
+                                          decltype(std::istreambuf_iterator<char>{})>
+                            {std::istreambuf_iterator<char>{stream},
+                             std::istreambuf_iterator<char>{}};
+
+        // cache the begin position so we write quals to the same position as seq in seq_qual case
+        size_t sequence_size_before = 0;
+        size_t sequence_size_after = 0;
```
smehringer:
```suggestion
        size_t sequence_size_after{0};
```

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,283 @@
...
+
+        auto constexpr is_tab = is_char<'\t'>;
+        auto constexpr is_cntrl= is_char<'\n'>;
+
+        while (is_char<'@'>(*ranges::begin(stream_view)))
+            detail::consume(stream_view | view::take_line);
+
+        /* ID */
+        if (is_char<'*'>(*begin(stream_view)))
+            throw parse_error{"The ID field may not be empty for SAM files."};
```
smehringer:
```suggestion
            throw parse_error{"The ID field may not be empty for sequence files."};
```

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,283 @@
...
+            ++begin(stream_view);
+            detail::consume(stream_view | view::take_until_or_throw(is_tab));
+        }
+        ++begin(stream_view);
+
+        auto seq_view = stream_view | view::take_until_or_throw(is_tab); // until next tab
+
+         // Sequence
+         if (is_char<'*'>(*begin(stream_view)))
+             throw parse_error{"The Sequence field may not be empty for SAM files."};
```
smehringer:
```suggestion
             throw parse_error{"The Sequence field may not be empty for sequence files."};
```

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,283 @@
...
+             {
+                  // seq_qual field implies that they are the same variable
+                  assert(std::addressof(sequence) == std::addressof(qualities));
+                  ranges::copy(qual_view | view::char_to<typename value_type_t<qual_type>::quality_alphabet_type>,
+                               ranges::begin(qualities) + sequence_size_before);
+             }
+             else if constexpr (!detail::decays_to_ignore_v<qual_type>)
+             {
+                  ranges::copy(qual_view | view::char_to<value_type_t<qual_type>>,
+                             std::back_inserter(qualities));
```
smehringer: fix indentation

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,342 @@
...
+          // Jump over the new line character in the end
+          detail::consume(stream_view | ranges::view::take_while(is_cntrl));
+    }
+
+    //!\copydoc sequence_file_output_format_concept::write
+    template <typename stream_type,     // constraints checked by file
+              typename seq_type,        // other constraints checked inside function
+              typename id_type,
+              typename qual_type>
+    void write(stream_type                     & stream,
```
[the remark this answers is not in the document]
smehringer: this is not resolved yet

include/seqan3/io/sequence_file/input.hpp
```diff
@@ -439,7 +440,8 @@ template <
                                                                                           field::ID,
                                                                                           field::QUAL>,
     detail::type_list_of_sequence_file_input_formats_concept valid_formats_      = type_list<sequence_file_format_fasta,
-                                                                                          sequence_file_format_fastq
+                                                                                          sequence_file_format_fastq,
```
smehringer: fix indentation :nail_care:

test/unit/io/sequence_file/sequence_file_format_sam_test.cpp
```diff
@@ -0,0 +1,493 @@
...
+#include <gtest/gtest.h>
+
+#include <seqan3/alphabet/quality/all.hpp>
+#include <seqan3/io/sequence_file/format_sam.hpp>
+#include <seqan3/io/sequence_file/input.hpp>
+#include <seqan3/io/sequence_file/input_format_concept.hpp>
+#include <seqan3/io/sequence_file/output.hpp>
+#include <seqan3/io/sequence_file/output_format_concept.hpp>
+#include <seqan3/test/pretty_printing.hpp>
+
```
smehringer: remove extra newline

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,278 @@
...
+#include <seqan3/range/view/take_exactly.hpp>
+#include <seqan3/range/view/take_line.hpp>
+#include <seqan3/range/view/take_until.hpp>
+#include <seqan3/std/ranges>
+#include <seqan3/std/view/subrange.hpp>
+#include <seqan3/std/view/transform.hpp>
+
+namespace seqan3
+{
+  /*!\brief       The sam format used as sequence file.
```
eseiler:
```suggestion
  /*!\brief       The SAM format used as sequence file.
```

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,278 @@
...
+   * \ingroup     sequence
+   *
+   * \details
+   *
+   * ### Introduction
+   *
+   * The SAM format is commonly used to store pairwise alignment information between a query sequence and its
+   * reference sequence, e.g. a read mapping result. As it provides information on the **id**, **sequence** and
+   * **quality** of each query, it can additionally be treated as a sequence file.
+   * See the [article on wikipedia](https://en.wikipedia.org/wiki/SAM_(file_format)) for an in-depth description of
```
eseiler:
```suggestion
   * See the [article on Wikipedia](https://en.wikipedia.org/wiki/SAM_(file_format)) for an in-depth description of
```

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,278 @@
...
+   * the format.
```
eseiler: I'd also link the technical specification https://samtools.github.io/hts-specs/SAMv1.pdf

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,279 @@
...
+   *
+   */
+class sequence_file_format_sam
+{
+public:
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+     //!\brief Default constructor is explicitly deleted, you need to give a stream or file name.
+    sequence_file_format_sam() = default;
```
h-2: the default constructor is defaulted, but the documentation says it's deleted...

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,279 @@
...
+            stream_it = '*';
+        else
+            ranges::copy(sequence | view::to_char, stream_it);
+        stream_it = '\t';
+
+        // Quality line
+        if constexpr (detail::decays_to_ignore_v<qual_type>)
+          stream_it = '*';
+        else if (ranges::empty(qualities))
+          stream_it = '*';
```
h-2: strange indent

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,279 @@
...
+        else
+            ranges::copy(qualities | view::to_char, stream_it);
+        detail::write_eol(stream_it, options.add_carriage_return);
```
h-2: insert empty line after if-else

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,279 @@
...
+   * ### Fields
+   *
+   * The SAM format provides the fields seqan3::field::SEQ, seqan3::field::ID and seqan3::field::SEQ_QUAL.
+   * All fields are allowed to be empty when writing.
+   *
+   * ### Implementation notes
+   *
+   * This implementation supports the following optional features of the format:
+   *
+   * line breaks and/or other whitespace characters in any part of the sequence (only when reading!), optional tags
```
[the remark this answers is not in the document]
MitraDarja: Actually, no, we don't allow that.  At the moment we don't even support whitespaces. Should we do that?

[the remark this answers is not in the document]
eseiler: As far as I can see, line breaks/whitespaces are not allowed. So just remove the docs for it.
I'd also remove the `This implementation supports the following optional features of the format:` and just state in the implementation details that the optional tags, or rather everything besides id, seq and qual,  are ignored (actually they are ignored in general, not just when reading)

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,280 @@
...
+ */
+
+#pragma once
+
+#include <iterator>
+#include <string>
+#include <string_view>
+#include <vector>
+
+#include <range/v3/algorithm/copy.hpp>
```
smehringer: I think you do not need this header and the one below.
If you use `ranges::copy` anywhere change it to `std::ranges::copy`.

MitraDarja:
@h-2 Do you also want a test for seq_qual combined in writing? Because at the moment seq_qual combined at writing is not an option and I am not sure if that has are reason or if I just forgot to add it.

h-2:
@MitraDarja For writing you don't need to handle SEQ_QUAL extra, because it is just split outside of the format already by sequence_file

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,277 @@
...
+        // cache the begin position so we write quals to the same position as seq in seq_qual case
+        size_t sequence_size_before{0};
+        size_t sequence_size_after{0};
+        if constexpr (!detail::decays_to_ignore_v<seq_type>)
+            sequence_size_before = ranges::size(sequence);
+
+        auto constexpr is_tab{is_char<'\t'>};
+        auto constexpr is_cntrl{is_char<'\n'>};
+
+        while (is_char<'@'>(*ranges::begin(stream_view)))
```
h-2: You still have some `ranges::` calls in the files, please replace them with just the function call or `seqan3::` or `std::ranges::`

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,277 @@
...
+                                          if (!is_legal_alph(c))
+                                          {
+                                              throw parse_error{std::string{"Encountered an unexpected letter: "} +
+                                                                is_legal_alph.msg.string() +
+                                                                " evaluated to false on " +
+                                                                detail::make_printable(c)};
+                                          }
+                                          return c;
+                                      })
+                                    | view::char_to<value_type_t<seq_type>>, // convert to actual target alphabet
```
h-2: this indented weirdly. Make the `|` align and make the lambas block align with the beginning of `view::` above.

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -0,0 +1,277 @@
...
+                           std::back_inserter(sequence));
+
+              sequence_size_after = ranges::size(sequence);
+          }
+          else
+          {
+              for (auto it = begin(seq_view); it != end(seq_view); ++it)
+                  ++sequence_size_after;
+          }
+          ++begin(stream_view); // skip tab
```
h-2: It's better to call `std::ranges::next(stream_view)` than to increment on begin.

h-2: Also in other places!

test/unit/io/sequence_file/sequence_file_format_sam_test.cpp
```diff
@@ -0,0 +1,524 @@
...
+    fout.get_stream().flush();
+
+    std::string comp
+    {
+R"(TEST 1	0	*	0	0	*	*	0	0	ACGT	!##$
+Test2	0	*	0	0	*	*	0	0	AGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGN	!##$&'()*+,-./+)*+,-)*+,-)*+,-)*+,BDEBDEBDEBDEBDEBDEBDEBDEBDEBDEBDEBDEBDEBDEBDBDDEBDBEEBEBE
+Test3	0	*	0	0	*	*	0	0	GGAGTATAATATATATATATAT	!!*+,-./+*+,-./+!!FF!!
+)"
+    };
+
```
h-2: I think it already is a lot more readable with the raw string :+1:

include/seqan3/io/sequence_file/format_sam.hpp
MitraDarja: Then, you have to merge this, please: https://github.com/seqan/seqan3/pull/724

h-2:
We'll merge this now, although probably somethings need to be adapted once the samformat for alignment in is mergend.

Thanks for working on this and sorry for the long review time 👍

PR #371, opened 2018-08-16
Feature/gap scheme

h-2:
TODOs:
- [ ] verify doxygen
- [ ] add scoring_scheme_concept


h-2:
Good to go from my PV

include/seqan3/alignment/scoring/gap_scheme.hpp
```diff
@@ -0,0 +1,290 @@
...
+        return gap_open;
+    }
+
+    //!\copydoc gap_open_score
+    constexpr score_t get_score_gap_open() const noexcept
+    {
+        return gap_open;
+    }
+
+    /*!\brief Compute the score of stretch of gap characters.
```
rrahn: of `a` stretch?

include/seqan3/alignment/scoring/scoring_scheme_concept.hpp
```diff
@@ -0,0 +1,97 @@
...
+ * will provide an implementation).
+ */
+//!\}
+//!\cond
+template <typename t, alphabet_concept alphabet_t, alphabet_concept alphabet2_t = alphabet_t>
+concept bool scoring_scheme_concept = requires (t scheme,
+                                                alphabet_t const alph1,
+                                                alphabet2_t const alph2)
+{
+    { scheme.score(alph1, alph2) } -> common_reference_concept<typename std::remove_reference_t<t>::score_type>;
```
rrahn: In my opinion, if this is a concept thought for score classes to be customisation points, we should offer them as global free functions.

h-2: but the scoring schemes are very specific to our algorithms. People might want to create their own, but I can't think of a place where they would want to integrate this with another library or need to wrap another library's type...

rrahn: On the contrary, people might actually have their own library, want to use our algorithms, but have different scoring schemes. Like if you would need a PSSM, which we don't offer (yet).  They either need to write a wrapper around their scheme or just need to provide an overload for `scrore(scheme, alph1, alph2)`.

PR #381, opened 2018-08-20
view::take, view::take_exactly and view::take_exactly_or_throw

h-2:
not blocked anymore


include/seqan3/range/view/take.hpp
```diff
@@ -0,0 +1,508 @@
...
+
+    //!\brief The sentinel type is identical to that of the underlying range.
+    using sentinel_type = sentinel_t<urng_t>;
+
+    //!\brief The iterator type inherits from the underlying type, but overwrite several operators.
+    class iterator_type : public inherited_iterator_base<iterator_type, iterator_t<urng_t>>
+    {
+    private:
+        //!\brief The iterator type of the underlying range.
+        using base_base_t = iterator_t<urng_t>;
```
rrahn: 🤦‍♂️

include/seqan3/range/view/take.hpp
```diff
@@ -0,0 +1,508 @@
...
+
+        /*!\name Arithmetic operators
+         * \brief seqan3::detail::inherited_iterator_base operators are used unless specialised here.
+         * \{
+         */
+        iterator_type & operator++() noexcept(noexcept(++base_base_t{}))
+        {
+            base_base_t::operator++();
+            ++pos;
+
```
rrahn: empty line

include/seqan3/range/view/take.hpp
```diff
@@ -0,0 +1,508 @@
...
+    template <view_concept urng_t>
+    static auto impl(urng_t urange, size_t const target_size)
+    {
+        return view_take<urng_t, exactly, or_throw>{std::move(urange), target_size};
+    }
+
+    /*!\brief       Call the view's constructor with the underlying range wrapped in seqan3::view::all as argument.
+     * \returns     An instance of seqan3::detail::view_take.
+     */
+    template <viewable_range_concept urng_t>
```
rrahn: Couldn't inherit this from the pipable_adaptor_base?

rrahn: Then we don't need to write this every time, which I think can be DRYed out.

h-2: > Couldn't inherit this from the pipable_adaptor_base?

It does.

> Then we don't need to write this every time, which I think can be DRYed out.

The problem here is that that functor type is templatised and that we use that parameter for deducing the returned view. Moving the `view::all` up into `pipable_adaptor` would work, but would break other specialisations of `pipable_adaptor` like `view::deep`. 

I have a large TODO for refactoring the pipable adaptor on my mental list, because it currently can't fold multiple adaptors into one without applying to a range and I will revisit this topic than, as well. But it is not trivial and I am trying to focus on more urgent things right now.

include/seqan3/range/view/take.hpp
h-2: It all about the bass :sound:

PR #384, opened 2018-08-21
re-activate fully virtualised builds because they have more memory

h-2:
high priority because the build failures on travis are annoying as hell and our builds are too long to profit from the short boot-up-time of the containers anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: false
+sudo: required
 dist: trusty
 language: cpp
 


marehr:
lgtm, but which ci failed?

h-2:
> lgtm, but which ci failed?

All the trouble with @joergi-w stuff and now i have failures sometimes as well.

joergi-w:
> All the trouble with @joergi-w stuff and now i have failures sometimes as well.

Is there still trouble somewhere that I should fix?

PR #362, opened 2018-08-10
[INTERNAL] Slacken the restriction on the qualified class from nucleotide_concept to alphabet_concept.

smehringer:
Fixes #359

Issue #359, opened 2018-08-10
qualified class should not be restricted to nucleotide alphabets


smehringer:
This especially hinders the gapped<alphabet> because it does not fulfil the nucleotide_concept as a `gap::GAP` does not have a complement. Hence, the quality_composition cannot be build over an gapped alphabet which would be desirable though IMPOV.

Also, the user might want to attach qualities to other strings..

PR #385, opened 2018-08-21
[FIX] completes PR#362.

smehringer:
PR #362, solving issue #359, slackened the restriction of qualified from nucleotide concept to alphabet concept so the complement function must be restricted also.

### DIFF
--- a/test/unit/alphabet/quality/qualified_test.cpp
+++ b/test/unit/alphabet/quality/qualified_test.cpp
@@ -305,6 +305,14 @@ TEST(qualified, cmp_qual)
 
 /************** ALPHABET and QUALITY concept **********************/
 
+TEST(qualified, concept)
+{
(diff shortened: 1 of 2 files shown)


test/unit/alphabet/quality/qualified_test.cpp
```diff
@@ -416,3 +424,11 @@ TEST(qualified, outstream)
 
     EXPECT_EQ(s.str(), "CA");
 }
+
+TEST(qualified, complement)
+{
+    qualified<dna4, phred42> t0{dna4::A, phred42{8}};
+    qualified<dna4, phred42> t0_c{dna4{dna4::A}.complement(), phred42{8}};
```
h-2: `dna4{dna4::A}` ?

smehringer:
!rebuild

PR #386, opened 2018-08-21
[FIX] Adds vector and sstream to include in seqan3/argument_parser/auxiliary.hpp

joshuak94:


include/seqan3/argument_parser/auxiliary.hpp
```diff
@@ -40,6 +40,8 @@
 #pragma once
 
 #include <seqan3/io/concept.hpp>
+#include <sstream>
+#include <vector>
```
h-2: please see https://github.com/seqan/seqan3/wiki/File-structure-and-naming#contents on how to group and order include directives!

PR #391, opened 2018-08-21
[FEATURE] introduce the snippets into the argument_parser module.

joshuak94:
Also adds type deduction guide in the validators.hpp regex_validator.


marehr:
can you rebase on top of master again?

test/snippet/argument_parser/validators_1.cpp
```diff
@@ -0,0 +1,32 @@
+//! [usage]
+#include <seqan3/argument_parser/all.hpp>
+int main(int argc, const char ** argv)
```
smehringer: add a new line before main function

joshuak94:
Rebased on current master.

marehr:
@smehringer We could, but there was always the problem that not every output is deterministic.

For example floats and doubles can have different outputs depending on the machine.

This feature request is out of scope of this pr. Can you create an issue?

PR #396, opened 2018-08-24
switch to std:: concept notation for std-concepts

h-2:
**this PR touches almost every file in the codebase**

It has a high priority for being merged, because anything else will break it.


h-2:
@seqan/all I have just merged major changes to concepts. All open PRs need to be rebased. Also have a look at the (soon updated) online documentation because some things have changed that affect many people.

h-2:
> merge without review? 

I discussed this with @rrahn last week. We agreed that the changes are just renaming and that it's hell to review 2500 lines of that. There are likely some things that were missed and maybe also some glitches and these can be fixed with smaller PRs next week.

> where did you split the std stuff into?

We now have three headers under `std/`:
  * `<concepts>` which corresponds to https://en.cppreference.com/w/cpp/concepts
  * `<iterator>` with the iterator concepts from https://en.cppreference.com/w/cpp/experimental/ranges#Iterators
  * `<ranges>` with the concepts from https://en.cppreference.com/w/cpp/experimental/ranges#Ranges

This corresponds 100% to the currently planned or accepted changes for C++20. You can look at the details in http://docs.seqan.de/seqan/3.0.0-master-user/group__std.html (sorry this wasn't properly updated up until now). There are still a number of things to do, I will create issues for them soon.

> what happened to the non-std concepts?

They are unchanged for now, we can discuss whether to keep our syntax or switch to std at a later point in time.

Issue #378, opened 2018-08-19
How does structured_rna::pseudoknot_id work?


marehr:
Hi Jörg,

maybe you can help me to understand this part of the code (especially line [194](#L194)):

https://github.com/seqan/seqan3/blob/7fbbefe94b542499ad58bd9293092fe9b84b2a9e/include/seqan3/alphabet/structure/structured_rna.hpp#L190-L198

I think it is strange that you can make a `constexpr` decision on the current state of a runtime value. I think this needs to be fixed, but I don't know what the expected outcome is.

Do you have test cases for `structure_alphabet_type::pseudoknot_support <= 1`?

Is `return 0;` a special id for opening and closing pairs?

Under which circumstances can  `structure_alphabet_type::pseudoknot_support <= 1` be true?

joergi-w:
Thank you for asking this – I'm happy to make this more clear. Let me divide your question into two topics, **(1)** `pseudoknot_support` and **(2)** `pseudoknot_id`. 

**(1)** The `pseudoknot_support` variable is the maximum allowed depth of pseudoknots for the structure alphabet, i.e. the nestedness of brackets (in biology: loops). Value 1 means that the format supports just one bracket type `(`/`)`, so no pseudoknots. Any value > 1 says that there is support for pseudoknots. As it is a static class variable (e.g. 1 for dot_bracket, 22 for wuss51), the constexpr decision is valid. 

**(2)** The `pseudoknot_id` gives an identifier for each bracket type, which is 0 for `()`, 1 for `[]`, and so on... It is defined for types that support pseudoknots and should return id = 0, if there is just one bracket type. The value is std::nullopt (undefined), iff the structure letter does not represent a bracket, but an unpaired site.

With the explanations above at hand, I hope you understand the code snippet and agree that it is correct. The doxygen documentation for these are directly in the formats, e.g. check `wuss.hpp` lines 247-257. I was thinking that a User, who uses a certain alphabet, would firstly look here. As you couldn't find or understand it, would you suggest that I put it to a different location or should I write even more detailed text?

marehr:
As discussed the line does not work:

```cpp
else if constexpr (is_pair_open() || is_pair_close()) 
```

joergi-w:
fixed with PR #402

PR #402, opened 2018-08-28
[FIX, TEST] fix a wrong constexpr condition and add a test

joergi-w:
Solves issue #378.

### DIFF
--- a/include/seqan3/alphabet/concept_pre.hpp
+++ b/include/seqan3/alphabet/concept_pre.hpp
@@ -246,6 +246,9 @@ struct pseudoknot_support{};
  * \attention Do not specialise this shortcut, instead specialise seqan3::pseudoknot_support.
  */
 template<typename alphabet_type>
+//!\cond
+    requires requires (alphabet_type alph) { pseudoknot_support<alphabet_type>::value; }
+//!\endcond
 constexpr uint8_t pseudoknot_support_v = pseudoknot_support<alphabet_type>::value;
 //!\}
 
--- a/include/seqan3/alphabet/concept_pre.hpp
+++ b/include/seqan3/alphabet/concept_pre.hpp
@@ -233,23 +233,23 @@ using underlying_char_t = typename underlying_char<alphabet_type>::type;
  * \snippet test/snippet/alphabet/concept_pre.cpp pseudoknot value retrieval
  *
  * \par Helper variable template
- *   seqan3::pseudoknot_support_v as a shorthand for `seqan3::pseudoknot_support<alphabet_type>::%value`
+ *   seqan3::max_pseudoknot_depth_v as a shorthand for `seqan3::max_pseudoknot_depth<alphabet_type>::%value`
  *
  * \attention This is the base template, it needs to be specialised.
  */
 template<typename alphabet_type>
-struct pseudoknot_support{};
+struct max_pseudoknot_depth{};
 
 /*!\brief The pseudoknot ability of the alphabet. [value metafunction shortcut]
  * \ingroup alphabet
  *
- * \attention Do not specialise this shortcut, instead specialise seqan3::pseudoknot_support.
+ * \attention Do not specialise this shortcut, instead specialise seqan3::max_pseudoknot_depth.
  */
 template<typename alphabet_type>
 //!\cond
-    requires requires (alphabet_type alph) { pseudoknot_support<alphabet_type>::value; }
+    requires requires (alphabet_type alph) { max_pseudoknot_depth<alphabet_type>::value; }
 //!\endcond
-constexpr uint8_t pseudoknot_support_v = pseudoknot_support<alphabet_type>::value;
+constexpr uint8_t max_pseudoknot_depth_v = max_pseudoknot_depth<alphabet_type>::value;
 //!\}
 
 } // namespace seqan3
--- a/include/seqan3/alphabet/structure/rna_structure_concept.hpp
+++ b/include/seqan3/alphabet/structure/rna_structure_concept.hpp
@@ -84,7 +84,10 @@ namespace seqan3
  * It is guaranteed to be smaller than seqan3::pseudoknot_support.
  */
 /*!\struct pseudoknot_support<structure_type>
- * \brief The pseudoknot ability of the structure_type.
+ * \brief The ability of this alphabet to represent pseudoknots, i.e. crossing interactions, up to a certain depth.
+ * \relates seqan3::rna_structure_concept
+ * \details It is the number of distinct pairs of interaction symbols the format supports. The value 1 denotes no
+ * pseudoknot support; any higher number gives the maximum nestedness. Value 0 is not allowed.
  */
 //!\cond
 template <typename structure_type>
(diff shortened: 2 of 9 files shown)


test/unit/alphabet/structure_test.cpp
```diff
@@ -164,8 +164,13 @@ TYPED_TEST(structure, concept)
 
 TYPED_TEST(structure, rna_structure_concept)
 {
-    EXPECT_TRUE(rna_structure_concept<dot_bracket3>);
-    EXPECT_TRUE(rna_structure_concept<wuss51>);
+    if constexpr (!std::is_same_v<TypeParam, dssp9>)
+    {
+        EXPECT_TRUE(rna_structure_concept<TypeParam>);
+        EXPECT_NE(pseudoknot_support_v<TypeParam>, 0);
+    }
```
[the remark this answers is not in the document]
joergi-w: Thank you – that's a good idea and it uncovered another issue with the concept check that gets fixed with this PR.

include/seqan3/alphabet/concept_pre.hpp
```diff
@@ -260,6 +260,9 @@ struct pseudoknot_support{};
  * \attention Do not specialise this shortcut, instead specialise seqan3::pseudoknot_support.
  */
 template<typename alphabet_type>
+//!\cond
+    requires requires (alphabet_type alph) { pseudoknot_support<alphabet_type>::value; }
+//!\endcond
```
marehr: `_v` helpers should not require concepts.

It should return `false` if it has no support.

Change the base class to:
```c++
template<typename alphabet_type>
struct pseudoknot_support :
    std::false_type
{};
```

include/seqan3/alphabet/structure/rna_structure_concept.hpp
```diff
@@ -100,7 +100,8 @@ concept bool rna_structure_concept = requires(structure_type val)
     { pseudoknot_id(val) } -> std::optional<uint8_t>;
 
     // this is delegated to a static class variable
-    { pseudoknot_support_v<structure_type> } -> uint8_t;
+    pseudoknot_support<std::remove_reference_t<structure_type>>::value;
+    pseudoknot_support_v<std::remove_reference_t<structure_type>>;
```
marehr: change it here to 
```c++
requires pseudoknot_support_v<std::remove_reference_t<structure_type>>;
```

test/unit/alphabet/structure_test.cpp
```diff
@@ -261,13 +270,15 @@ TEST(rna_structure, wuss)
...
         EXPECT_TRUE(vec[idx].pseudoknot_id());
+        EXPECT_EQ(*vec[idx].pseudoknot_id(), (idx - 7) / 2);
     }
     for (unsigned idx = 8; idx <= 22; idx+=2)
     {
         EXPECT_TRUE(vec[idx].is_pair_close());
         EXPECT_FALSE(vec[idx].is_unpaired());
         EXPECT_FALSE(vec[idx].is_pair_open());
         EXPECT_TRUE(vec[idx].pseudoknot_id());
+        EXPECT_EQ(*vec[idx].pseudoknot_id(), (idx - 8) / 2);
```
marehr: Why fo you need to dereference it? Is the id a pointer?
// Ah okay it is an optional

Can you make the right side the return type of `pseudoknot_id()`?

So change it to:
```c++
EXPECT_EQ(vec[idx].pseudoknot_id(), std::optional{(idx - 8) / 2});
```

test/unit/alphabet/structure_test.cpp
```diff
@@ -611,6 +622,22 @@ TEST(structured_rna, rna_structure_concept)
...
+    using type = structured_rna<rna5, dot_bracket3>;
+    EXPECT_EQ(type::pseudoknot_support, 1); // max bracket depth 1
+    type t0{rna5::A, dot_bracket3::PAIR_OPEN};
+    type t1{rna5::N, dot_bracket3::UNPAIRED};
+    type t2{rna5::U, dot_bracket3::PAIR_CLOSE};
+
+    EXPECT_TRUE (t0.pseudoknot_id());
+    EXPECT_FALSE(t1.pseudoknot_id());
+    EXPECT_TRUE (t2.pseudoknot_id());
+    EXPECT_EQ   (t0.pseudoknot_id().value(), 0);
```
marehr: Also add here

```c++
EXPECT_EQ   (t0.pseudoknot_id(), std::optional{0});
```

include/seqan3/alphabet/concept_pre.hpp
joergi-w: No, pseudoknot_support is a static class variable of type uint8_t. 
I implemented it in the same way as alphabet_size_v, which has a concept requirement.
Without the concept check, the rna_structure_concept would not evaluate to `false` on dssp9, but rather produce a compiler error when pseudoknot_support::value is not found.

include/seqan3/alphabet/structure/rna_structure_concept.hpp
joergi-w: As the value is not bool, I can test for
```cpp
requires pseudoknot_support_v<std::remove_reference_t<structure_type>> > 0;
```
Would this be better? Why do I need the `requires` here?

test/unit/alphabet/structure_test.cpp
joergi-w: Okay, I agree to make comparisons of same types.

include/seqan3/alphabet/concept_pre.hpp
marehr: The name of `alphabet_size_v` shows that it gives a size back. `pseudoknot_support` does not hint in any way that this will return a number. It sound to me like a boolean and the correlating question: `Has this thing pseudoknot support?`. Is this for a biologist / bioinformatician clear that this returns a number and what this number represents?

This is only a question and out of the scope of this pull request. Thank you for clarifying!

> Without the concept check, the rna_structure_concept would not evaluate to false on dssp9, but rather produce a compiler error when pseudoknot_support::value is not found.

This in itself is clearly a bug on gcc's side and knowing that `alphabet_size_v` uses the same trick, I think Hannes already reported it.

joergi-w: You are right and indeed formerly the variabe was boolean. But I need to query the allowed nestedness in structure I/O. I was already considering to name it `max_pseudoknot_depth`, but that sounds very technical. In your opinion, is that more intuitive?

include/seqan3/alphabet/structure/structured_rna.hpp
```diff
@@ -178,17 +178,17 @@ class structured_rna :
...
-     * It is guaranteed to be smaller than seqan3::pseudoknot_support.
+     * It is guaranteed to be smaller than seqan3::max_pseudoknot_depth.
      */
     constexpr std::optional<uint8_t> pseudoknot_id() const noexcept
     {
-        if constexpr (structure_alphabet_type::pseudoknot_support > 1)
+        if constexpr (structure_alphabet_type::max_pseudoknot_depth > 1)
             return get<1>(*this).pseudoknot_id();
-        else if constexpr (is_pair_open() || is_pair_close())
+        else if (is_pair_open() || is_pair_close())
```
rrahn: Can you change the second part of the expression to:
```cpp
if constexpr ...
{
    ...
}
else
{
    return (is_pair_open() || is_pair_close()) ? 0 : std::nullopt;
}

joergi-w: I have to cast the `0` to std::optional then to have the same type in the ?: operator.

PR #405, opened 2018-08-29
first steps towards clang support

h-2:


h-2:
> AND wait until clang has full support for concepts.

The experimental branch has full support, now, it's just not merged, yet. I think we could provide them with valuable testing data maybe resulting in a faster merge.

> I think we should postpone this after the user meeting

 I am just writing the longer "SeqAn3 announcement" and it would feel good to be able to write that we have experimental support for clang, too.

> We also should brief our group that we don't write concept bool anymore.

Yes, I am doing this also in the context that a lot of other concept stuff is changing right now and I would like to update corresponding docs once and not again-and-again. That will also make it easier for our devs :smile: 

> But, I think we can merge your first two commits out of the four.

The first two are the important ones, but the other two just correct minor stuff:
3) not using concept outside its scope as a keyword: we should have never done this anyway, I am surprised the concepts TS accepts it
4) it's just two semicolons and they are valid in GCC, too

marehr:
> 3) not using concept outside its scope as a keyword: we should have never done this anyway, I am surprised the concepts TS accepts it
> 4) it's just two semicolons and they are valid in GCC, too

I meant with `But, I think we can merge your first two commits out of the four.` exactly to merge those two comment.

BTW:
```c++
TEST(inherited_iterator_base, concept)
```
was always okay, because it will generate a similar function name like this `inherited_iterator_base_concept`.

PR #409, opened 2018-08-31
[FIX] track changes in google testing framework

h-2:
Recent travis and jenkings builds have been failing with:
```cpp
/home/travis/build/seqan/seqan3/test/performance/example/example_benchmark.cpp:56:34: error: ‘void benchmark::State::SetBytesProcessed(int64_t)’ is deprecated: The SetItemsProcessed()/items_processed()/SetBytesProcessed()/bytes_processed() will be removed in a future release. Please use custom user counters. [-Werror=deprecated-declarations]
     state.SetBytesProcessed(bytes);
                                  ^
In file included from /home/travis/build/seqan/seqan3/test/performance/example/example_benchmark.cpp:38:0:
/home/travis/build/seqan/seqan3-build/vendor/benchmark/include/benchmark/benchmark.h:549:8: note: declared here
   void SetBytesProcessed(int64_t bytes) { bytes_processed_ = bytes; }
        ^~~~~~~~~~~~~~~~~
```

This PR fixes the behaviour.

@marehr Can you have quick look at it?

### DIFF
--- a/test/performance/example/example_benchmark.cpp
+++ b/test/performance/example/example_benchmark.cpp
@@ -53,7 +53,7 @@ static void memcpy_benchmark(benchmark::State& state) {
         memcpy(dst, src, size);
 
     int64_t bytes = int64_t(state.iterations()) * int64_t(size);
-    state.SetBytesProcessed(bytes);
+    state.counters["bytes_processed"] = bytes;
     delete[] src;
     delete[] dst;
 }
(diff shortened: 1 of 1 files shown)


h-2:
> I would have pinned the version in seqan3-test.cmake as a quick fix 

Yeah, I wasn't sure about that. I guess we should decide once for all the test suites whether we pull a release or track master, but just using some commit-id seems arbitrary. Can you make an issue for this?

PR #414, opened 2018-09-05
[FEATURE] The cigar_op and cigar alphabet.

smehringer:
Introduces all that is needed to represent and stream a cigar string.


include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,326 @@
...
+    //!\brief The type combining both types into a streamable string.
+    using string_type = std::string;
+
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    cigar() = default;
+    constexpr cigar(cigar const &) = default;
+    constexpr cigar(cigar &&) = default;
+    constexpr cigar & operator =(cigar const &) = default;
```
rrahn: whitespace between `operator` and `=`

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,326 @@
...
+ * CIGAR elements in this context consisting of a CIGAR operator (seqan3::cigar_op)
+ * and a count value.
+ *
+ * Example usage:
+ * \snippet test/snippet/alphabet/cigar/cigar_op.cpp general
+ *
+ * \note Usually you do not want to manipulate cigar elements and vectors on
+ *       your own but convert an alignment to a cigar and back. See
+ *       seqan3::get_cigar_vector for how to convert two aligned sequences into
+ *       an cigar_vector.
```
joergi-w: ~an~ a cigar_vector

joergi-w: seqan3::cigar_vector

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,326 @@
...
+    //!\copydoc cartesian_composition::operator=(component_type const alph)
+    SEQAN3_DOXYGEN_ONLY(( constexpr cigar & operator=(component_type const alph) {} ))
+    //!\copydoc cartesian_composition::operator=(indirect_component_type const alph)
+    SEQAN3_DOXYGEN_ONLY(( constexpr cigar & operator=(indirect_component_type const alph) {} ))
+    //!\}
+
+    /*!\name Write functions
+     * \{
+     */
+    //!\brief Assign from a character. This modifies the internal sequence letter.
```
joergi-w: sequence letter? You probably mean the cigar operator.

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,326 @@
...
+    constexpr cigar & assign_char(char_type const c)
+    {
+        seqan3::assign_char(get<0>(*this), c);
+        return *this;
+    }
+
+    //!\brief Assign from a count value. This modifies the internal quality letter.
```
joergi-w: _quality letter_ is a copy'n'paste error. Below there is at least two more of this (please do a grep on _quality_).

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,326 @@
...
+    {
+        get<1>(*this) = c;
+        return *this;
+    }
+    //!\}
+
+    /*!\name Read functions
+     * \{
+     */
+    //!\brief Return a character. This reads the internal sequence letter.
```
joergi-w: ...and also sequence letter appears again. Please also do a search on this.

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,326 @@
...
+        return seqan3::to_char(get<0>(*this));
+    }
+
+    //!\brief Return the count value. This reads the internal quality letter.
+    constexpr uint32_t to_count() const noexcept
+    {
+        return get<1>(*this);
+    }
+
+    //!\brief Return the count value. This reads the internal quality letter.
```
joergi-w: Returns a string!

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,326 @@
...
+};
+
+//!\brief Type deduction guide enables usage of cigar without specifying template args.
+//!\relates cigar
+template <typename cigar_type, typename count_type>
+cigar(cigar_type &&, count_type &&)
+    -> cigar<std::decay_t<cigar_type>, std::decay_t<count_type>>;
+
+/*!\brief Specialized implementation of seqan3::alphabet_concept::operator<<()
+ *        for cigar delegates to `cig.to_string()`.
```
joergi-w: delegates to `seqan3::cigar::to_string()`
(it is a class member function and the param name should not appear in \brief)

include/seqan3/alphabet/cigar/all.hpp
```diff
@@ -0,0 +1,79 @@
...
+ * | Operator | Description |
+ * |----------|-------------|
+ * | D        | Deletion; the nucleotide is present in the reference but not in the read |
+ * | H        | Hard Clipping; the clipped nucleotides are not present in the read. |
+ * | I        | Insertion; the nucleotide is present in the read  but not in the reference. |
+ * | M        | Match; can be either an alignment match or mismatch. |
+ * | N        | Skipped region; a region of nucleotides is not present in the read |
+ * | P        | Padding; padded area in the read and not in the reference |
+ * | S        | Soft Clipping;  the clipped nucleotides are present in the read |
+ * | X        | Read Mismatch; the nucleotide is present in the reference |
```
rrahn: is not present?

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,326 @@
...
+ * \relates cigar
+ * \param os  The output stream you are printing to.
+ * \param cig The seqan3::cigar letter that you wish to print.
```
joergi-w: in your case, _letter_ might be confusing. Maybe use element or character.

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,326 @@
...
+
+namespace seqan3
+{
+
+//!\brief Alias for an std::vector of seqan3::cigar.
+//!\ingroup cigar
+using cigar_vector = std::vector<cigar<>>;
+
+/*!\brief Specialized implementation of the stream operator for a container of
+ *        cigar elements; delegates to `cig.to_string()`.
```
joergi-w: see above for `cig.to_string()`

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,326 @@
...
+} // namespace seqan3
+
+// ------------------------------------------------------------------
+// containers
+// ------------------------------------------------------------------
+
+namespace seqan3
+{
+
+//!\brief Alias for an std::vector of seqan3::cigar.
```
joergi-w: of `seqan3::cigar<cigar_op, uint32_t>`
Because you use always the default template args here.

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,326 @@
...
+ *
+ * Example usage:
+ * \snippet test/snippet/alphabet/cigar/cigar_op.cpp general
+ *
+ * \note Usually you do not want to manipulate cigar elements and vectors on
+ *       your own but convert an alignment to a cigar and back. See
+ *       seqan3::get_cigar_vector for how to convert two aligned sequences into
+ *       an cigar_vector.
+ */
+template<typename cigar_type = cigar_op, typename count_type = uint32_t>
```
rrahn: I would switch the order. First, cigar_type is kind of fixed, albeit we need to declare it as a template in order to inherit the tuple size, tuple element functions. I am wondering if we should drop it and reimplement the functions specifically for the cigar alphabet. Because it is kind of misleading to make the interface generic while it is not generic. It does not justify to save 6 lines of code. Unless it would be more complicated? Second, the user might want to use a smaller int type and `cigar<uint16_t>` is better than `cigar<cigar_op, uint16_t>`

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,326 @@
...
+{
+
+//!\brief Alias for an std::vector of seqan3::cigar.
+//!\ingroup cigar
+using cigar_vector = std::vector<cigar<>>;
+
+/*!\brief Specialized implementation of the stream operator for a container of
+ *        cigar elements; delegates to `cig.to_string()`.
+ * \relates cigar
+ * \tparam range_type Must model seqan3::std::ranges::ForwardRange and its
```
joergi-w: `seqan3::std::ranges::ForwardRange`
something is wrong here

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,326 @@
...
+    return os;
+}
+
+} // namespace seqan3
+
+// ------------------------------------------------------------------
+// create cigar_vector out of alignment
+// ------------------------------------------------------------------
+
+namespace seqan3
```
joergi-w: Do not close namespace, if you re-open it immediately.

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,326 @@
...
+    return os;
+}
+
+} // namespace seqan3
+
+// ------------------------------------------------------------------
+// containers
+// ------------------------------------------------------------------
+
+namespace seqan3
```
joergi-w: Do not close namespace, if you re-open it immediately.

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,326 @@
...
+// create cigar_vector out of alignment
+// ------------------------------------------------------------------
+
+namespace seqan3
+{
+
+/*!\brief Compares two aligned sequence values and returns their seqan3::cigar_op.
+ * \ingroup aligned_sequence
+ * \tparam reference_value_type Must be equality comparable to seqan3::gap.
+ * \tparam query_value_type     Must be equality comparable to seqan3::gap.
```
joergi-w: 2x: instead of `_value_type` rather use `_char_type` here?

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,326 @@
...
+ * \param  reference_value      Considered indirectly related to the cigar operation.
```
joergi-w: this param description is not clear

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,326 @@
...
+ * \param  query_value          The value the cigar operation is based on.
+ * \param  extended_cigar Whether to print the extended cigar alphabet or not. See seqan3::cigar_op.
```
joergi-w: align

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,326 @@
...
+        ? (query_value == gap::GAP)
+            ? cigar_op::P
+            : cigar_op::I
+        : (query_value == gap::GAP)
+            ? cigar_op::D
+            : (extended_cigar)
+                ? (query_value == reference_value)
+                    ? cigar_op::EQ
+                    : cigar_op::X
+                : cigar_op::M;
```
joergi-w: really?? This nested `?:` is hard to understand and debug. Can we use temporary bool variables to reduce/structure this or just have if statements? Is the efficiency so much better to justify this construct?

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,326 @@
...
+}
+
+/*!\brief Compares two aligned sequence values and returns their seqan3::cigar_op.
```
joergi-w: It returns a whole vector of complete cigar compositions!

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,326 @@
...
+ * \tparam ref_seq_type   Must model seqan3::aligned_sequence_concept.
+ * \tparam query_seq_type Must model seqan3::aligned_sequence_concept.
+ * \param  alignment      The alignment, represented by a pair of aligned sequences,
+ *                        to be transformed into seqan3::cigar_vector based on the
+ *                        second (query) sequence.
+ * \param  offset_start   The offset at the start of the query sequence indicating soft-clipping.
+ * \param  offset_end     The offset at the end of the query sequence indicating soft-clipping.
+ * \param  extended_cigar Whether to print the extended cigar alphabet or not. See seqan3::cigar_op.
+ * \returns A seqan3::cigar_op representing the alignment operation between the
+ *          two values.
```
joergi-w: It returns a whole vector of complete cigar compositions!

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,326 @@
...
+ * \attention Note that seqan3::cigar elements (respectively by their
+ *            seqan3::cigar_op) are always based on one of the two sequences in
+ *            a pairwise alignment. In this case, the resulting cigar_vector is
+ *            based on sequence at the second position of the \p alignment pair.
+ *
+ * Example:
+ * \snippet test/snippet/alphabet/cigar/cigar.cpp get_cigar_vector
+ */
+template<aligned_sequence_concept ref_seq_type, aligned_sequence_concept query_seq_type>
+cigar_vector get_cigar_vector(std::pair<ref_seq_type, query_seq_type> const & alignment,
```
rrahn: Mhmm, I really think we should offer this as a view::to_cigar
Let's consider this use case: 
```cpp
for (auto res : align(seq_rng, config)
{
   //res.alignment would be a zip view over the two aligned sequences.
   bam_record.cigar = res.alignment | to_cigar([soft_clipping_begin{}, soft_clipping_end{}]);
}
```
For the time being, I would make the utility function into `detail` and document it that it will be replaced by a view.

rrahn: Mhmm, I really think we should offer this as a view::to_cigar
Let's consider this use case: 
```cpp
for (auto res : align(seq_rng, config)
{
   //res.alignment would be a zip view over the two aligned sequences.
   bam_record.cigar = res.alignment | to_cigar([soft_clipping_begin{}, soft_clipping_end{}]);
}
```
For the time being, I would make the utility function into `detail` and document it that it will be replaced by a view.

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,326 @@
...
+
+/*!\brief Compares two aligned sequence values and returns their seqan3::cigar_op.
+ * \ingroup aligned_sequence
+ * \tparam ref_seq_type   Must model seqan3::aligned_sequence_concept.
+ * \tparam query_seq_type Must model seqan3::aligned_sequence_concept.
+ * \param  alignment      The alignment, represented by a pair of aligned sequences,
+ *                        to be transformed into seqan3::cigar_vector based on the
+ *                        second (query) sequence.
+ * \param  offset_start   The offset at the start of the query sequence indicating soft-clipping.
+ * \param  offset_end     The offset at the end of the query sequence indicating soft-clipping.
```
joergi-w: offset is not a good expression at the end of a sequence. 
Think about sth like `std::pair<uint,uint> aligned_region` or `uint align_start` + `align_end`...

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,326 @@
...
+ * \param  extended_cigar Whether to print the extended cigar alphabet or not. See seqan3::cigar_op.
+ * \returns A seqan3::cigar_op representing the alignment operation between the
+ *          two values.
+ *
+ * \details
+ *
+ * \attention Note that seqan3::cigar elements (respectively by their
+ *            seqan3::cigar_op) are always based on one of the two sequences in
```
joergi-w: based on -> relating to, with respect to

include/seqan3/alphabet/cigar/cigar.hpp
smehringer: I see your point for switching the template arguments but I think there is more than 6 lines of code to change the cigar alphabet to its own composition. The cartesian_composition base class includes 600 lines of codes that would need to be copy and pasted, since the functionality for assigning, constructing and comparing a compound type is redundant (e.g. `cigar<> c = 4;` sets only the count value). So the question is if it justifies 300 lines of redundant code? I'm not sure myself..

include/seqan3/alphabet/cigar/cigar.hpp
smehringer: What's the added value of a view compared to a free function in this context? 
I only know that it would be pipepable that way, but I cannot think of a use case where you want to modify a cigar string any further by another view..

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,326 @@
...
+#include <seqan3/std/ranges>
+
+#include <range/v3/view/zip.hpp>
+
+namespace seqan3
+{
+
+/*!\brief The cigar composition joins the seqan3::cigar_op alphabet with a count value.
+ * \ingroup cigar
+ * \tparam cigar_type Must be of type seqan3::cigar_op.
```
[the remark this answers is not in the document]
smehringer: I have to because of the cartesian_composition inheritance. If I don't do this I cannot use any std::tuple specific functions.. :( that's why Rene proposed to change the composition to it's own type.

include/seqan3/alphabet/cigar/cigar.hpp
smehringer: It's so much shorter than as many if clauses though :open_mouth: 
I don't know about debugging but I think with the indentation it is easy to read.

include/seqan3/alphabet/cigar/cigar.hpp
rrahn: Well, if it is a transformation of a range, than it should be a view, otherwise, who should know wether this can be transformed with a view and something else can be transformed with a free function. You might want to drop the first 4 cigar elements.

include/seqan3/alphabet/cigar/cigar.hpp
rrahn: No, I probably didn't express myself correctly.
The cigar should still be a composition and inherit from it. But my last understanding, was, that the std::tuple_element and std::tuple_size wouldn't work if cigar does not offer the exact same template arguments as the std::tuple it is inheriting from through the composition. But you could overload those two meta-functions for your cigar class. Still the other inherited functions should work.

smehringer: Ah I see now! I will check if this is the only 6 lines of code we would need to overload. If so, I agree to change it.

include/seqan3/alphabet/cigar/cigar.hpp
smehringer: It is a transformation but not a linear one working on every element of the range. One advancing step in the cigar view would mean a linear stream over the alignment. Does this not pose a problem?

Side questions: Is a std::pair of alignments usually transformable by views, like ` res.alignment | view::take(5) `? 

And I doubt that you want to drop or take certain cigar elements ... If anything you want to treat the cigar vector as an alignment and e.g. insert a deletion, but then you should actually create an alignment out of the cigar_vector. Do we also want to model this with a view? 
```
auto alignment = cigar_v | to_alignment(query_seq, ref_seq);
```
This could get very complicated I think..

include/seqan3/alphabet/cigar/cigar.hpp
smehringer: Done.

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,409 @@
...
+ *
+ * \note Usually you do not want to manipulate cigar elements and vectors on
+ *       your own but convert an alignment to a cigar and back. See
+ *       seqan3::get_cigar_vector for how to convert two aligned sequences into
+ *       a seqan3::cigar_vector.
+ */
+template<typename length_t = uint32_t>
+//!\cond
+    requires std::UnsignedIntegral<length_t>
+//!\endcond
```
h-2: `template <std::UnsignedIntegral length_t = uint32_t>`

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,409 @@
...
+    }
+
+    //!\brief Return the length value. This reads the internal length value.
+    constexpr length_type to_length() const noexcept
+    {
+        return get<1>(*this);
+    }
+
+    //!\brief Return the cigar element as a string. This reads both, the internal seqan3::cigar_op letter and length value.
+    string_type to_string() const noexcept
```
h-2: Hm, I am not really happy with the general design of this. I would argue that if you define two alphabets where to_char returns e.g. 'M' for the first and "20" for the second, then `to_char` should return "20M" and there shouldn't be a `to_string`. That's just super confusing. A string is always multiple letters of one alphabet and here we convert a single letter into multiple ones...

I think this can only be properly solved by adding a "small_string" class to our container that has variable size <= 16 and uses a fixed 16byte array internally, similar to the small-string-optimisation part of std::string. We could make it inherit from view_base so that views accept it as constant-time copyable.

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,409 @@
...
+ * \relates cigar
+ * \tparam length_type Type of the length value; must model std::UnsignedIntegral.
+ * \param  os          The output stream you are printing to.
+ * \param  cig         The seqan3::cigar element that you wish to print.
+ * \returns A reference to the output stream.
+ */
+template<typename length_type>
+//!\cond
+    requires std::UnsignedIntegral<length_type>
+//!\endcond
```
h-2: short template syntax...

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,409 @@
...
+template <typename length_type = uint32_t, std::ranges::ForwardRange range_type>
+//!\cond
+    requires std::UnsignedIntegral<length_type> &&
+             std::Same<reference_t<range_type>, cigar<length_type> &>
+//!\endcond
+std::ostream & operator<<(std::ostream & os, range_type const & cigv)
+{
+    std::for_each(cigv.begin(), cigv.end(), [&os] (auto cig) { os << cig; } );
+    return os;
+}
```
h-2: we shouldn't have this here at this point, instead we should add a generic one for debug_stream

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,409 @@
...
+ * seqan3::cigar_op::M, for the basic cigar the two bases are aligned, while
+ * in the extended CIGAR alphabet (\p extended_cigar = `true`) the function
+ * will return an seqan3::cigar_op::X since the bases are aligned but are not
+ * equal.
+ */
+template<typename reference_char_type, typename query_char_type>
+//!\cond
+    requires std::EqualityComparableWith<std::remove_reference_t<reference_char_type>, gap> &&
+             std::EqualityComparableWith<std::remove_reference_t<query_char_type>, gap>
+//!\endcond
```
h-2: you can also use shorter syntax here and don't forget the space between `template <typename`

h-2: shouldn't this function maybe go into detail?

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,409 @@
...
+                    ? cigar_op::P
+                    : cigar_op::I
+                : (query_char == gap::GAP)
+                    ? cigar_op::D
+                    : (extended_cigar)
+                        ? (query_char == reference_char)
+                            ? cigar_op::EQ
+                            : cigar_op::X
+                        : cigar_op::M;
+}
```
h-2: would this maybe be easier to read:
```cpp
    return (reference_char == gap::GAP) ? (query_char == gap::GAP ? cigar_op::P : cigar_op::I)        :
           (query_char == gap::GAP)     ? cigar_op::D                                                 :
           (extended_cigar)             ? (query_char == reference_char ? cigar_op::EQ : cigar_op::X) : cigar_op::M;
```

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,409 @@
...
+ * ATGCCCCGTTG--C
+ * ```
+ * In this case, the function seqan3::compare_aligned_values will return
+ * The following cigar string when printed: "4M2I5M2D1M". The extended cigar
+ * string would look like this: "3=1X2I3=1X1=2D1=".
+ *
+ * ### Code Example:
+ *
+ * \snippet test/snippet/alphabet/cigar/cigar.cpp get_cigar_vector
+ */
```
h-2: could we make a view of this instead?

include/seqan3/alphabet/cigar/cigar.hpp
rrahn: Probably you could use fixed size constexpr_string as a basis

rrahn:
Is this superseded by #861 or what is the state with this?

h-2:
superseded #887

PR #421, opened 2018-09-11
[INTERNAL] get cigar string from alignment.

smehringer:
replaces #414 until we decide what to do with the cigar string/alphabet/sequence in general.


h-2:
All of this is in detail and will be reworked/removed once the other stuff lands, right?
Then I think we can just merge this if all tests pass.

h-2:
documentation tests fail

smehringer:
they pass now

PR #429, opened 2018-09-13
[FEATURE] Adds alignment stream and two way alignment executor.

rrahn:
Adds an executor and steaming interface for range based iteration over a set of pairwise alignments.
The code is tested with a branch of the Myers and some integration test.
I will add more. in the future, but for now, I want to bring everything together. 
So also documentation is yet minimal but will be upgraded as soon as possible.

### DIFF
--- a/include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
+++ b/include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
@@ -1,0 +1,227 @@
+// ============================================================================
+//                 SeqAn - The Library for Sequence Analysis
+// ============================================================================
+//
+// Copyright (c) 2006-2018, Knut Reinert & Freie Universitaet Berlin
+// Copyright (c) 2016-2018, Knut Reinert & MPI Molekulare Genetik
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//     * Redistributions of source code must retain the above copyright
+//       notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//     * Neither the name of Knut Reinert or the FU Berlin nor the names of
+//       its contributors may be used to endorse or promote products derived
+//       from this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL KNUT REINERT OR THE FU BERLIN BE LIABLE
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+// DAMAGE.
+//
+// ============================================================================
+
+/*!\file
+ * \brief Provides seqan3::detail::alignment_executor_two_way.
+ * \author Rene Rahn <rene.rahn AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <functional>
+#include <optional>
+#include <type_traits>
+
+#include <seqan3/alignment/pairwise/execution/alignment_range.hpp>
+#include <seqan3/alignment/pairwise/execution/execution_handler_sequential.hpp>
+#include <seqan3/core/metafunction/range.hpp>
+#include <seqan3/std/ranges>
+
+namespace seqan3::detail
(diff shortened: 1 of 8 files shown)


include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
```diff
@@ -0,0 +1,227 @@
...
+    using buffer_pointer         = iterator_t<buffer_type>;
+    //!\}
+public:
+
+    //!\brief The result type of invoking the alignment instance.
+    using value_type = buffer_value_type;
+    //!\brief A reference to the alignment result.
+    using reference  = std::add_lvalue_reference_t<value_type>;
+    //!\brief The offset type for the buffer.
+    using off_type   = typename buffer_type::difference_type;
```
marehr: why not call it `difference_type`?

include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
```diff
@@ -0,0 +1,227 @@
...
+
+    /*!\name Buffer
+     * \{
+     */
+    //!\brief The result of invoking the alignment instance.
+    using buffer_value_type       = std::invoke_result_t<resource_value_type>;
+    //!\brief The internal buffer.
+    using buffer_type             = std::vector<buffer_value_type>;
+    //!\brief The pointer type of the buffer.
+    using buffer_pointer         = iterator_t<buffer_type>;
```
marehr: indentation

include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
```diff
@@ -0,0 +1,227 @@
...
+
+    /*!\name Resource
+     * \{
+     */
+    //!\brief The iterator over the resource.
+    using resource_iterator = iterator_t<std::remove_reference_t<align_instance_rng_t>>;
+    //!\brief The sentinel over the resource.
+    using resource_sentinel = sentinel_t<std::remove_reference_t<align_instance_rng_t>>;
+    //!\brief The value type of the resource.
+    using resource_value_type     = std::remove_reference_t<decltype(*std::declval<resource_iterator>())>;
```
marehr: indent

include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
```diff
@@ -0,0 +1,227 @@
...
+        // Advance the resource.
+        std::advance(resource_iter, in_avail());
+        return in_avail();
+    }
+    //!\}
+
+    /*!\name Miscellaneous
+     * \{
+     */
+
```
marehr: empty space?

include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
```diff
@@ -0,0 +1,227 @@
...
+    buffer_pointer egptr;
+};
+
+/*!\name Type deduction guide
+ * \{
+ */
+template <std::ranges::ViewableRange resource_rng_t>
+alignment_executor_two_way(resource_rng_t && res) ->
+    alignment_executor_two_way<resource_rng_t, execution_handler_sequential>;
+
```
marehr: empty space

include/seqan3/alignment/pairwise/execution/alignment_range.hpp
```diff
@@ -0,0 +1,243 @@
...
+{
+    static_assert(!std::is_const_v<stream_buffer_t>,
+                  "Cannot create an alignment stream over a const buffer.");
+
+    //!\brief The iterator of seqan3::detail::alignment_range.
+    class iterator_type
+    {
+    public:
+        //!\brief Type for distances between iterators.
+        using difference_type = typename alignment_range::off_type;
```
marehr: call it `difference_type` instead of `off_type`

include/seqan3/alignment/pairwise/execution/execution_handler_sequential.hpp
```diff
@@ -0,0 +1,80 @@
...
+    /*!\name Execution
+     * \{
+     */
+    //!\brief Invokes the passed alignment instance in a blocking manner.
+    template <typename alignment_instance_t,
+              typename delegate_t>
+    void execute(alignment_instance_t && alignment,
+                 delegate_t && delegate)
+    {
+        delegate(std::invoke(alignment));
```
marehr: why not simply call `alignment()`?

include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
rrahn: I probably change it back. I kind of see it as a stream, and wasn't sure if we might want to add seeking. So as if to skip alignment computations. But maybe that is over-engineered at the current point.

include/seqan3/alignment/pairwise/execution/execution_handler_sequential.hpp
rrahn: That's why:
```
Invoke the Callable object f with the parameters args. As by INVOKE(std::forward<F>(f), std::forward<Args>(args)...).
where INVOKE(f, t1, t2, ..., tN) is defined as follows:
If f is a pointer to member function of class T:
If std::is_base_of<T, std::decay_t<decltype(t1)>>::value is true, then INVOKE(f, t1, t2, ..., tN) is equivalent to (t1.*f)(t2, ..., tN)
If std::decay_t<decltype(t1)> is a specialization of std::reference_wrapper, then INVOKE(f, t1, t2, ..., tN) is equivalent to (t1.get().*f)(t2, ..., tN)
If t1 does not satisfy the previous items, then INVOKE(f, t1, t2, ..., tN) is equivalent to ((*t1).*f)(t2, ..., tN).
Otherwise, if N == 1 and f is a pointer to data member of class T:
If std::is_base_of<T, std::decay_t<decltype(t1)>>::value is true, then INVOKE(f, t1) is equivalent to t1.*f
If std::decay_t<decltype(t1)> is a specialization of std::reference_wrapper, then INVOKE(f, t1) is equivalent to t1.get().*f
If t1 does not satisfy the previous items, then INVOKE(f, t1) is equivalent to (*t1).*f
Otherwise, INVOKE(f, t1, t2, ..., tN) is equivalent to f(t1, t2, ..., tN) (that is, f is a FunctionObject)
```

marehr: If I don't oversee anything; `std::invoke(callable)` is in all cases equivalent to `callable()`

I would prefer `std::function` in both cases, because it does type erasure and this might have an in pact on compile time (drastically reducing it).

Not important right now.

PR #430, opened 2018-09-13
[FIX,DOC] Fixes group of aligned sequence submodule

rrahn:


marehr:
sorry; you are not self-referencing; I should go to bed :sleeping:

PR #431, opened 2018-09-13
[FEATURE] edit distance alignment

marehr:


include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
```diff
@@ -0,0 +1,551 @@
...
+//!\endcond
+} // namespase seqan3::detail
+
+namespace seqan3::detail
+{
+/*!\brief This calculates an alignment using the edit distance and without a band.
+ * \ingroup alignment
+ */
+template <typename database_t, typename query_t, typename align_config_t>
+struct pairwise_alignment_edit_distance_unbanded
```
rrahn: maybe not class.

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
```diff
@@ -0,0 +1,551 @@
...
+        }
+
+        return false;
+    }
+
+    /*!\todo describe me
+     */
+    inline bool on_hit()
+    {
+        // _setFinderEnd(finder);
```
rrahn: maybe we can throw an exception if some one wants to access this function.

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
```diff
@@ -0,0 +1,551 @@
...
+    static constexpr bool use_max_errors = detail::max_errors_concept<align_config_t>;
+    //!\brief Whether the alignment is a global alignment or not.
+    static constexpr bool is_global = detail::global_config_concept<align_config_t>;
+    //!\brief Whether the alignment is a semi-global alignment or not.
+    static constexpr bool is_semi_global = detail::semi_global_config_concept<align_config_t>;
+
+    //!\brief How to pre-initialize hp.
+    static constexpr word_type hp0 = is_global ? 1 : 0;
+
+    static_assert(8*sizeof(word_type) <= 64, "we assume at most uint64_t as word_type");
```
rrahn: whitespace between the multiply

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
```diff
@@ -0,0 +1,551 @@
...
+     */
+    pairwise_alignment_edit_distance_unbanded(
+        database_t _database,
+        query_t _query,
+        align_config_t _config)
+        : database{std::forward<database_t>(_database)},
+          query{std::forward<query_t>(_query)},
+          config{std::forward<align_config_t>(_config)},
+          _score{query.size()},
+          database_it{std::cbegin(database)},
```
rrahn: please use ranges::begin() also we cannot assure, or at least need to specify in the concept that it is const iteratable. We have a concept for this.

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
```diff
@@ -0,0 +1,551 @@
...
+    /*!\brief Constructor
+     * \param[in] _database \copydoc database
+     * \param[in] _query    \copydoc query
+     * \param[in] _config   \copydoc config
+     */
+    pairwise_alignment_edit_distance_unbanded(
+        database_t _database,
+        query_t _query,
+        align_config_t _config)
+        : database{std::forward<database_t>(_database)},
```
rrahn: forward is superfluous. Either we copy a view or a l-value ref for the sequences.

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
```diff
@@ -0,0 +1,551 @@
...
+};
+//!\endcond
+} // namespase seqan3::detail
+
+namespace seqan3::detail
+{
+/*!\brief This calculates an alignment using the edit distance and without a band.
+ * \ingroup alignment
+ */
+template <typename database_t, typename query_t, typename align_config_t>
```
rrahn: require ViewableRange concept?

include/seqan3/alignment/matrix/rowwise_matrix.hpp
```diff
@@ -0,0 +1,109 @@
...
+#include <vector>
+#include <seqan3/core/platform.hpp>
+
+namespace seqan3::detail
+{
+
+/*!\brief A matrix represented in a one-dimensional std::vector
+ * \ingroup alignment_matrix
+ * \implements seqan3::detail::matrix_concept
+ *
```
smehringer: \tparam missing.

include/seqan3/alignment/matrix/rowwise_matrix.hpp
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * This data structure stores the matrix in a flat way using the
+ * std::vector<#entry_type> data structure where each row is stored
+ * continuously.
+ */
+template <typename entry_t>
+struct rowwise_matrix
```
smehringer: I think with a custom constructor this should be a class no? and the default constructors should be explicelty defined?

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
```diff
@@ -0,0 +1,551 @@
...
+    //!\}
+
+private:
+
+    /*!\todo describe me
+     */
+    template <bool with_overflow_check>
+    inline void compute_step(word_type b, word_type & hp, word_type & hn, word_type & vp, word_type & vn, word_type & carry_d0, word_type & carry_hp, word_type & carry_hn)
+    {
+        word_type x, d0, t;
```
smehringer: :dizzy_face: I hope you will change the names of those variables for the final version? Otherwise I at least won't review this code..

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -0,0 +1,339 @@
...
+        throw std::runtime_error{"End trace must be NONE"};
+
+    while(row >= 0 && col >= 0)
+    {
+        trace_matrix_directions dir = matrix.at(row, col);
+        if ((dir & L) == L)
+        {
+            col = std::max<int>(col - 1, 0);
+            gapped_database.push_front(database[col]);
+            gapped_query.push_front(gap::GAP);
```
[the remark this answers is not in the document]
marehr: `insert_gap` assumes that I have an underlying sequence already. But in this case I always push a character to the front. Furthermore, I already know my `gapped_sequence` datatype.

I could change that to `insert_gap(gapped_database, std::begin(gapped_database))`, but I don't think this is better than simply calling `push_front`.

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
marehr: I would to prefer not; This change would require a lot of changes in the code.

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
marehr: I also want to bind containers.

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
marehr: they are directly taken from the paper every rename would be confusing to whoever has read the paper.

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
marehr: we should add a `on_hit` to the configuration which will be called from within this function

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
marehr: never mind went through

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
rrahn: I see the point now.

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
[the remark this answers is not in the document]
rrahn: In general I would construct the a temporary gapped_sequence in reverse order and then reverse move it into the target sequences value by value. It still would be 2n runtime but I think more efficient then constructing lists (not sure if it does small value optimisations on the nodes) and then converting them to gapped sequences.
But we will fix that later. Right now, it is important to have a working code base and that the interfaces can be streamlined.

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -0,0 +1,339 @@
...
+ *
+ * ### Output
+ * \include test/snippet/alignment/matrix/alignment_trace_matrix.out
+ */
+template <typename database_type, typename query_type, typename align_config_type, typename ...score_matrix_params_t>
+//!\cond
+    requires matrix_concept<alignment_score_matrix<score_matrix_params_t...>> &&
+             std::is_integral_v<typename alignment_score_matrix<score_matrix_params_t...>::entry_type>
+//!\endcond
+struct alignment_trace_matrix<database_type, query_type, align_config_type, alignment_score_matrix<score_matrix_params_t...>>
```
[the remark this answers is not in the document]
rrahn: Ok so just debug mode.

rrahn:
Ok, in general the important thing is the myers implementation and that we can compute the alignment from it. The details of how we refactor the matrix types and compute the trace must be done incrementally when we add the dp code.

include/seqan3/alignment/matrix/alignment_score_matrix.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+/*!\file
+ * \author Marcel Ehrhardt <marcel.ehrhardt AT fu-berlin.de>
+ * \brief Contains the declaration of seqan3::detail::alignment_score_matrix.
+ */
+
+#pragma once
+
+#include <seqan3/std/concepts>
+#include <seqan3/alignment/matrix/rowwise_matrix.hpp>
```
h-2: order

include/seqan3/alignment/matrix/alignment_score_matrix.hpp
```diff
@@ -0,0 +1,88 @@
...
+ * ## Example
+ *
+ * \snippet test/snippet/alignment/matrix/alignment_score_matrix.cpp code
+ *
+ * ### Output
+ * \include test/snippet/alignment/matrix/alignment_score_matrix.out
+ */
+template <std::Integral score_t>
+struct alignment_score_matrix<std::vector<score_t>>
+    : public rowwise_matrix<score_t>
```
h-2: `row_wise_matrix` :nail_care:

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -0,0 +1,333 @@
...
+             std::Same<typename remove_cvref_t<trace_matrix_t>::entry_type, trace_directions>
+//!\endcond
+inline std::pair<std::list<gapped_database_t>, std::list<gapped_query_t>>
+alignment_trace(database_t && database, query_t && query, trace_matrix_t && matrix)
+{
+    constexpr auto N = trace_directions::none;
+    constexpr auto D = trace_directions::diagonal;
+    constexpr auto L = trace_directions::left;
+    constexpr auto U = trace_directions::up;
+    int row = matrix.rows()-1;
```
h-2: spacing

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -0,0 +1,333 @@
...
+    typename database_t,
+    typename query_t,
+    typename trace_matrix_t,
+    typename gapped_database_t = gapped<std::decay_t<decltype(std::declval<database_t>()[0])>>,
+    typename gapped_query_t = gapped<std::decay_t<decltype(std::declval<query_t>()[0])>>>
+//!\cond
+    requires matrix_concept<remove_cvref_t<trace_matrix_t>> &&
+             std::Same<typename remove_cvref_t<trace_matrix_t>::entry_type, trace_directions>
+//!\endcond
+inline std::pair<std::list<gapped_database_t>, std::list<gapped_query_t>>
```
h-2: this returns a pair of lists of sequences? shouldn't it be pairs of numbers?

h-2: ah, ok, so `gapped_database_t` is actually `gapped_database_alphabet_t`? Could we rename it? It's very confusing.

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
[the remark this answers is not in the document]
h-2: > In general I would construct the a temporary gapped_sequence in reverse order and then reverse move it into the target sequences value by value. It still would be 2n runtime but I think more efficient then constructing lists

absolutely! a std::list of single byte type has an overhead of 16x over the actual size and bad cache behaviour and slower access because locality cannot be assumed.

> I could change that to insert_gap(gapped_database, std::begin(gapped_database)), but I don't think this is better than simply calling push_front.

Well we want to be able to specify other `aligned_sequence_concept` types as return values...

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -0,0 +1,333 @@
...
+ * \snippet test/snippet/alignment/matrix/alignment_trace_matrix_vector.cpp code
+ *
+ * ### Output
+ * \include test/snippet/alignment/matrix/alignment_trace_matrix_vector.out
+ */
+template <>
+struct alignment_trace_matrix<std::vector<trace_directions>>
+    : public rowwise_matrix<trace_directions>
+{
+    using rowwise_matrix<trace_directions>::rowwise_matrix;
```
h-2: are alignment computed row-wise or column-wise in our DP-code? This should match the matrix type...

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -0,0 +1,333 @@
...
+ *
+ * \snippet test/snippet/alignment/matrix/alignment_trace_matrix.cpp code
+ *
+ * ### Output
+ * \include test/snippet/alignment/matrix/alignment_trace_matrix.out
+ */
+template <typename database_type, typename query_type, typename align_config_type, typename ...score_matrix_params_t>
+//!\cond
+    requires matrix_concept<alignment_score_matrix<score_matrix_params_t...>> &&
+             std::is_integral_v<typename alignment_score_matrix<score_matrix_params_t...>::entry_type>
```
h-2: `std::Integral`

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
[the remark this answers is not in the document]
h-2: all of this is only required in debug mode? Why? And how is this a view?

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -0,0 +1,333 @@
...
+    //!\brief Access to the score_matrix.
+    inline score_matrix_type const & score_matrix() const noexcept
+    {
+        return *this;
+    }
+
+private:
+
+    //!\brief Does the trace come from the above entry?
+    inline bool is_trace_up(unsigned row, unsigned col) const noexcept
```
h-2: parameters should be const qualified and I think `size_t` would be more appropriate than `unsigned`? We always use fixed-width integer types or `size_t` in SeqAn...

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -0,0 +1,333 @@
...
+    {
+        entry_type direction{};
+
+        if(is_trace_diagonal(row, col))
+            direction |= entry_type::diagonal;
+
+        if(is_trace_up(row, col))
+            direction |= entry_type::up;
+
+        if(is_trace_left(row, col))
```
h-2: spacing

include/seqan3/alignment/matrix/matrix_concept.hpp
```diff
@@ -0,0 +1,141 @@
...
+     * \brief The entry of the matrix at position (\a row, \a col), e.g. `matrix[row][col]`.
+     * \memberof seqan3::detail::matrix_concept
+     */
+    { m.at(0u, 0u) } -> typename matrix_t::entry_type;
+//!\cond
+};
+//!\endcond
+//!\}
+
+/*!\brief `true`, if two alignment matrices are equal.
```
h-2: This is not a full sentence. Also please group the comparison operator into named group called "Comparison operators".

include/seqan3/alignment/matrix/matrix_concept.hpp
```diff
@@ -0,0 +1,141 @@
...
+ * \tparam    matrix2_t The type of the right hand side matrix.
+ * \param[in] lhs       Compare the left hand side matrix
+ * \param[in] rhs       with the right hand side matrix.
+ */
+template <matrix_concept matrix1_t, matrix_concept matrix2_t>
+//!\cond
+    requires std::EqualityComparableWith<
+                typename matrix1_t::entry_type,
+                typename matrix2_t::entry_type
+             >
```
h-2: no trailing `>` please (I think the whole thing would fit into one line).

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -0,0 +1,333 @@
...
+
+/*!\file
+ * \author Marcel Ehrhardt <marcel.ehrhardt AT fu-berlin.de>
+ * \brief Contains the declaration of seqan3::detail::alignment_trace_matrix.
+ */
+
+#pragma once
+
+#include <list>
+#include <seqan3/alphabet/gap/gapped.hpp>
```
h-2: move to other headers

include/seqan3/alignment/matrix/matrix_concept.hpp
```diff
@@ -0,0 +1,141 @@
...
+//!\endcond
+inline bool operator==(matrix1_t const & lhs, matrix2_t const & rhs)
```
h-2: mark `noexcept`

include/seqan3/alignment/matrix/rowwise_matrix.hpp
h-2: this was not resolved

h-2: Also rename to `row_wise_matrix`

include/seqan3/alignment/matrix/rowwise_matrix.hpp
```diff
@@ -0,0 +1,116 @@
...
+     rowwise_matrix(rowwise_matrix &&) = default;
+     rowwise_matrix & operator=(rowwise_matrix const &) = default;
+     rowwise_matrix & operator=(rowwise_matrix &&) = default;
+    /*!\brief Construct the matrix out of the *entries*, the *rows*,
+     *        and the *cols*.
+     * \param entries The entry values as a flat std::vector <#entry_type>.
+     * \param rows    The number of rows.
+     * \param cols    The number of columns.
+     */
+    rowwise_matrix(std::vector<entry_type> entries, size_t rows, size_t cols)
```
h-2: `size_t const`

include/seqan3/alignment/matrix/rowwise_matrix.hpp
```diff
@@ -0,0 +1,116 @@
...
+        : _entries{std::move(entries)}, _rows{rows}, _cols{cols}
+    {}
+    //!\}
+
+    //!\copydoc seqan3::detail::matrix_concept::rows
+    inline std::size_t rows() const noexcept
```
h-2: member functions that are defined with their declerarion are implicity `inline`.

you can ommit `std::` before size_t

include/seqan3/alignment/matrix/rowwise_matrix.hpp
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    //!\copydoc seqan3::detail::matrix_concept::cols
+    inline std::size_t cols() const noexcept
+    {
+        return _cols;
+    }
+
+    //!\copydoc seqan3::detail::matrix_concept::at
+    inline entry_type at(unsigned row, unsigned col) const noexcept
```
h-2: `size_t const`

include/seqan3/alignment/matrix/rowwise_matrix.hpp
```diff
@@ -0,0 +1,116 @@
...
+    {
+        return _entries[row * cols() + col];
```
h-2: `assert` that this is not out-of-bounds

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
```diff
@@ -0,0 +1,548 @@
...
+
+#include <algorithm>
+#include <bitset>
+#include <utility>
+#include <seqan3/alphabet/nucleotide/dna4.hpp>
+#include <seqan3/core/add_enum_bitwise_operators.hpp>
+#include <seqan3/core/metafunction/range.hpp>
+
+#include <seqan3/alignment/matrix/alignment_score_matrix.hpp>
+#include <seqan3/alignment/matrix/alignment_trace_matrix.hpp>
```
h-2: include order messed up

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
```diff
@@ -0,0 +1,548 @@
...
+template <typename align_config_t>
+concept semi_global_config_concept = requires()
+{
+    {align_config_t::is_semi_global} -> bool;
+};
+template <typename align_config_t>
+concept global_config_concept = requires()
+{
+    {align_config_t::is_global} -> bool;
+};
```
h-2: is this really necessary? Why is not documented?

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
```diff
@@ -0,0 +1,548 @@
...
+template <typename align_config_t>
+concept max_errors_concept = requires(align_config_t cfg)
+{
+    {cfg.max_errors};
+};
+//!\endcond
+} // namespase seqan3::detail
+
+namespace seqan3::detail
+{
```
h-2: closing and re-opening same namespace

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
```diff
@@ -0,0 +1,548 @@
...
+    using trace_matrix_type = detail::alignment_trace_matrix<pairwise_alignment_edit_distance_unbanded>;
+
+    //!\brief The size of one machine word.
+    static constexpr uint8_t word_size = sizeof(word_type) * 8;
+
+private:
+    //!\brief The type of an iterator of the database sequence.
+    using database_iterator = iterator_t<database_type>;
+    //!\brief The alphabet type of the query sequence.
+    using query_alphabet_type = std::remove_reference_t<decltype(query[0])>;
```
h-2: `reference_t`

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
```diff
@@ -0,0 +1,548 @@
...
+
+    //!\brief true - uses the ukkonen trick with the last active cell and bounds the error to config.max_errors.
```
h-2: "true - " ?

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
```diff
@@ -0,0 +1,548 @@
...
+    std::vector<word_type> vn{};
+    //!\brief The machine words which translate a letter of the query into a bit mask.
+    //!\details Each bit position which is true (=1) corresponds to a match of a letter in the query at this position.
+    std::vector<word_type> bit_masks{};
+
+    /*!\name Only used when use_max_errors is true
+     * \{
+     */
+    //!\brief Which score value is considered as a hit?
+    unsigned max_errors{255};
```
h-2: fixed width integer type or `size_t`

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
h-2: The comment was not addressed. the forward has no effect, because there is no forwarding reference. Also the indention is weird, maybe this would be more legible:
```cpp
    pairwise_alignment_edit_distance_unbanded(database_t _database, query_t _query, align_config_t _config) :
        database{std::forward<database_t>(_database)},
        query{std::forward<query_t>(_query)},
```

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
```diff
@@ -0,0 +1,548 @@
...
+        while (!(_score <= max_errors))
+        {
+            advance_score(vp[last_block], vn[last_block], score_mask);
+            if(!prev_last_active_cell())
+                break;
+        }
+
+        if ((score_mask == last_score_mask) && (last_block == vp.size() - 1))
+            return on_hit();
+        else
```
h-2: balance parantheses

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
```diff
@@ -0,0 +1,548 @@
...
+
+        score_mask = 1;
+        last_block++;
+    }
+
+    //!\brief Use the ukkonen trick and update the last active cell.
+    inline bool update_last_active_cell()
+    {
+        // updating the last active cell
+        while (!(_score <= max_errors))
```
h-2: `>` ?

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
```diff
@@ -0,0 +1,548 @@
...
+        last_block--;
+        if (is_global && last_block == (unsigned)-1)
+            return false;
+
+        score_mask = (word_type)1 << (word_size - 1);
+        return true;
+    }
+
+    //!\brief Increment the last active cell position.
+    inline void next_last_active_cell()
```
h-2: ~~inline~~

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
```diff
@@ -0,0 +1,548 @@
...
+    std::vector<state_type> states{};
+    //!\brief Add a computation step
+    void add_state()
+    {
+        states.push_back(state_type{vp, vn});
+    }
+
+public:
+
+    /*!\name Constructors
```
h-2: `Constructors, destructor and assignment`

Also please either default or delete the usual 5/6

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
```diff
@@ -0,0 +1,548 @@
...
+namespace seqan3::detail
+{
+/*!\brief This calculates an alignment using the edit distance and without a band.
+ * \ingroup alignment
+ * \tparam database_t     \copydoc pairwise_alignment_edit_distance_unbanded::database_type
+ * \tparam query_t        \copydoc pairwise_alignment_edit_distance_unbanded::query_type
+ * \tparam align_config_t The type of the alignment config.
+ */
+template <std::ranges::ViewableRange database_t, std::ranges::ViewableRange query_t, typename align_config_t>
+struct pairwise_alignment_edit_distance_unbanded
```
h-2: this should be a `class`

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
```diff
@@ -0,0 +1,548 @@
...
+    {
+        return score_matrix_type{*this};
+    }
+
+    //!\brief Return the trace matrix of the alignment.
+    trace_matrix_type trace_matrix()
+    {
+        return trace_matrix_type{*this};
+    }
+};
```
h-2: there are 6(!) switches between public and private sections in this class, please clean this up a little, it's very hard to follow

h-2: and put members into groups!

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
```diff
@@ -0,0 +1,548 @@
...
+template<typename database_t, typename query_t, typename config_t>
+pairwise_alignment_edit_distance_unbanded(database_t && database, query_t && query, config_t config)
+    -> pairwise_alignment_edit_distance_unbanded<database_t, query_t, config_t>;
+//!\}
+
+} // namespace seqan3::detail
+
+//!\cond
+namespace seqan3::detail
+{
```
h-2: closing and opening same namespace

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
```diff
@@ -0,0 +1,548 @@
...
+    using alignment_t = pairwise_alignment_edit_distance_unbanded<database_t, query_t, align_config_t>;
+    using score_matrix_t = alignment_score_matrix<alignment_t>;
+    using base_trace_matrix_t = alignment_trace_matrix<database_t const &, query_t const &, align_config_t, score_matrix_t>;
+
+    alignment_trace_matrix(alignment_t const & alignment) :
+        base_trace_matrix_t{alignment.database, alignment.query, alignment.config, score_matrix_t{alignment}}
+    {
+    }
+};
```
h-2: why do we need these many different matrix types?

rrahn:
Ok, I don't want to start a long discussion now. 
So here is my very short rationale:
I need the alignment code to finally wrap the interface around it. That I finished now, and I will spent the last two days that we have to clean it up. 
I do think we should not put this away, but discuss everything with a beer on Friday. Before that we have a greater goal to achieve.

marehr:
> I know we have a lot on our plate and we won't address all of the raised issues before Wednesday, but I hope we can at least discuss the raised issues so they can be resolved ASAP after the workshop.

This was the premisses of the merge and we will do it :)

I think that the most stuff is minor and will be easily addressed. And most of the existing code will change once a couple of features are implemented.

include/seqan3/alignment/matrix/alignment_matrix_formatter.hpp
```diff
@@ -0,0 +1,383 @@
...
+    {u8"⠀",u8"⠁",u8"⠈",u8"⠉",u8"⠄",u8"⠅",u8"⠌",u8"⠍"}, // 3bytes per char
+    2, 3
+};
+
+constexpr alignment_matrix_format alignment_matrix_format::unicode_arrows
+{
+    u8"ε", u8"║", u8"═", u8"╬",
+    {u8"↺",u8"↖",u8"↑",u8"↖↑",u8"←",u8"↖←",u8"↑←",u8"↖↑←"}, // 3bytes per char
+    2, 3
+};
```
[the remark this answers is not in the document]
marehr: This is 2 months old code. Way before the whole workshop hype started and it was incredible helpful for debugging. Furthermore, this will be a cool feature once the debug stream is in.

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
marehr: The forward is correct, because the type deduction:

https://github.com/seqan/seqan3/blob/ae2fad8773728cfc567fff15d7c0dde15a663a46/include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp#L464-L466

include/seqan3/alignment/matrix/alignment_matrix_formatter.hpp
```diff
@@ -0,0 +1,377 @@
...
+#include <seqan3/alignment/matrix/matrix_concept.hpp>
+#include <seqan3/alignment/matrix/alignment_trace_matrix.hpp>
+#include <seqan3/core/metafunction/template_inspection.hpp>
+
+namespace seqan3::detail
+{
+
+/*!\brief Format used by seqan3::detail::alignment_matrix_formatter
+ * \ingroup alignment_matrix
+ *
```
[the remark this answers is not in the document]
marehr: I addressed this point

include/seqan3/alignment/matrix/alignment_matrix_formatter.hpp
```diff
@@ -0,0 +1,383 @@
...
+    //!\brief row separator symbol (a single symbol)
+    char const * row_sep{};
+    //!\brief row separator symbol (a single symbol)
+    char const * row_col_sep{};
+
+    /*!\brief Eight symbols for each combination of directions a trace can have (each entry can have multiple symbols).
+     *
+     * * 1st bit: **D** = diagonal
+     * * 2nd bit: **U** = up
+     * * 3th bit: **L** = left
```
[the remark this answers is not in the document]
marehr: true, but this explains the data layout which is something different from the enum definition.

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
marehr: will be replaced with different code / is currently only a stub

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
marehr:
```
using query_alphabet_type = std::remove_reference_t<reference_t<query_type>>;
```

Works, but looks funny xD

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
marehr: I find this easier to read. Because, you know that after the while loop `_score <= max_errors` is true.

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
marehr: I don't understand this comment

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
marehr: I wanted to put them into groups, but this does not work apparently when mixing type definitions, functions and variables.

include/seqan3/alignment/pairwise_alignment_edit_distance_unbanded.hpp
marehr: There are only two matrix types and they are needed. The specialisations are needed, because of the differences in the stored data between the algorithms.

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
marehr: This will likely change to column-wise, but this will be addressed later.

include/seqan3/alignment/matrix/alignment_matrix_formatter.hpp
```diff
@@ -0,0 +1,377 @@
...
+     * \sa https://en.wikipedia.org/wiki/Comma-separated_values
+     *
+     *
+     * ### Score output
+     * \snippet test/snippet/alignment/matrix/alignment_matrix_formatter.out alignment_matrix_format::csv#score
+     *
+     * ### Trace output
+     * \snippet test/snippet/alignment/matrix/alignment_matrix_formatter.out alignment_matrix_format::csv#trace
+     */
+    static const alignment_matrix_format csv;
```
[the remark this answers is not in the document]
marehr: You can't declare static constexpr members of the same type (as the class) inside the class itself.

PR #432, opened 2018-09-13
Feature/align result

rrahn:
Adds output config and alignment result class.

Note the intended use is to set in the align config the wanted output. 
There are several possibilities defined in align_config_key. 
The keys indicate an include order, that is, when the user requests a trace, he will automatically get the score, the begin coordinates and the end coordinates of the alignment. If the user requests the end coordinate, he will get the end and the score only. In all configurations one can also access the alignment id.
This is needed, since in some cases the same alignment produces multiple results but the resulting alignment range would be flat. So it would be hard for the user to figure out, which result corresponds to which alignment.


include/seqan3/alignment/pairwise/all.hpp
```diff
@@ -0,0 +1,47 @@
...
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+// DAMAGE.
+//
+// ============================================================================
+
+/*!\file
+ * \brief Meta-header for the \link pairwise pairwise submodule \endlink.
```
joergi-w: 2x pairwise

rrahn:
Also fix:

there is a seqan3::scoring_scheme_concept that should probably be required here
and 
seqan3:: prefix in documentation of other pr.

include/seqan3/alignment/pairwise/all.hpp
rrahn: the second one is part of the description. The first one is the referenced symbol

PR #436, opened 2018-09-14
[FEATURE] The alignment_file_out class and SAM format

smehringer:
This is the raw version to discuss design decisions and semantic errors.

### DIFF
--- a/include/seqan3/io/alignment_file/format_sam.hpp
+++ b/include/seqan3/io/alignment_file/format_sam.hpp
@@ -1,0 +1,568 @@
+// ============================================================================
+//                 SeqAn - The Library for Sequence Analysis
+// ============================================================================
+//
+// Copyright (c) 2006-2018, Knut Reinert & Freie Universitaet Berlin
+// Copyright (c) 2016-2018, Knut Reinert & MPI Molekulare Genetik
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//     * Redistributions of source code must retain the above copyright
+//       notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//     * Neither the name of Knut Reinert or the FU Berlin nor the names of
+//       its contributors may be used to endorse or promote products derived
+//       from this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL KNUT REINERT OR THE FU BERLIN BE LIABLE
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+// DAMAGE.
+//
+// ============================================================================
+
+/*!\file
+ * \brief Provides the seqan3::alignment_file_format_sam class.
+ * \author Svenja Mehringer <avenja.mehringer AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <iterator>
+#include <string>
+#include <string_view>
+#include <vector>
+
+#include <range/v3/algorithm/copy.hpp>
+#include <range/v3/view/remove_if.hpp>
+
+#include <seqan3/core/concept/tuple.hpp>
+#include <seqan3/core/metafunction/range.hpp>
+#include <seqan3/io/alignment_file/detail.hpp>
+#include <seqan3/io/alignment_file/header.hpp>
+// #include <seqan3/io/alignment/input_options.hpp>
+#include <seqan3/io/alignment_file/output_options.hpp>
+#include <seqan3/io/alignment_file/sam_tag_dictionary.hpp>
+#include <seqan3/io/detail/ignore_output_iterator.hpp>
+#include <seqan3/io/detail/output_iterator_conversion_adaptor.hpp>
+#include <seqan3/io/detail/misc.hpp>
+#include <seqan3/io/stream/parse_condition.hpp>
+#include <seqan3/range/view/char_to.hpp>
+#include <seqan3/range/view/get.hpp>
+#include <seqan3/range/view/take_until.hpp>
+#include <seqan3/range/view/to_char.hpp>
+#include <seqan3/std/ranges>
+#include <seqan3/std/concepts>
+
+namespace seqan3
+{
+
+/*!\brief       The SAM format.
+ * \implements  alignment_file_format_concept
+ * \ingroup     alignment_file
+ *
+ * \details
+ *
+ * ### Introduction
+ *
+ * SAM is often used for storing alignments of several read sequences against one
+ * or more reference sequences. See the
+ * [article on wikipedia](https://en.wikipedia.org/wiki/SAM_(file_format)) for an
+ * introduction of the format or look into the official
+ * [SAM format specifications](https://samtools.github.io/hts-specs/SAMv1.pdf).
+ * **SeqAn implements version 1.6 of the SAM specification**.
+ *
+ * ### Fields
+ *
+ * The SAM format provides the following fields:
+ * seqan3::field::ALIGNMENT, seqan3::field::SEQ, seqan3::field::QUAL,
+ * seqan3::field::ID, seqan3::field::REF_SEQ, seqan3::field::REF_ID
+ * seqan3::field::REF_OSSFET, seqan3::field::OFFSET, seqan3::field::FLAG,
+ * seqan3::field::MAPQ and seqan3::field::MATE.
+ * In addition there is the seqan3::field::HEADER_PTR, which is usually not set but
+ * needed to provide the range-based functionality of the file.
+ *
+ * **None of the fields are required** when writing but will be defaulted
+ * to '0' for numeric fields and '*' for other fields.
+ *
+ * ### SAM format columns -> fields
+ *
+ * As many users will be accustomed to the columns of the SAM format, here is a
+ * mapping of the common SAM format columns to the SeqAn3 record fields:
+ *
+ * | #  | SAM Column ID |  FIELD name                                       |
+ * |:--:|:--------------|:--------------------------------------------------|
+ * | 1  | QNAME         | seqan3::field::ID                                 |
+ * | 2  | FLAG          | seqan3::field::FLAG                               |
+ * | 3  | RNAME         | seqan3::field::REF_ID                             |
+ * | 4  | POS           | seqan3::field::REF_OFFSET                         |
+ * | 5  | MAPQ          | seqan3::field::MAPQ                               |
+ * | 6  | CIGAR         | implicilty stored in seqan3::field::ALIGNMENT     |
+ * | 7  | RNEXT         | seqan3::field::MATE (tuple pos 0)                 |
+ * | 8  | PNEXT         | seqan3::field::MATE (tuple pos 1)                 |
+ * | 9  | TLEN          | seqan3::field::MATE (tuple pos 2)                 |
+ * | 10 | SEQ           | seqan3::field::SEQ                                |
+ * | 11 | QUAL          | seqan3::field::QUAL                               |
+ *
+ * The (read sequence/query) **OFFSET** will be required to store the soft
+ * clipping information at the read start (end clipping will be automatically
+ * deduced by how much the read sequence length + offset is larger than the
+ * alignment length).
+ *
+ * Note: SeqAn currently does not support hard clipping. When reading SAM,
+ * hard-clipping is discarded; but the resulting alignment/sequence combination
+ * is still valid.
+ *
+ * ### Format Check
+ *
+ * The format checks are implemented according to the official
+ * [SAM format specifications](https://samtools.github.io/hts-specs/SAMv1.pdf)
+ * in order to ensure correct SAM file output.
+ *
+ * If a non-recoverable format violation is encountered on reading, or you specify
+ * invalid values/combinations when writing, seqan3::format_error is thrown.
+ *
+ * Note: All sequence like fields in SAM (e.g. field::SEQ) are truncated at the
+ *       the first white space character (see seqan3::is_space) to ensure a
+ *       correct format.
+ *
+ * ### Header implementation
+ *
+ * The SAM header is printed once in the beginning, before the first record is
+ * written.
+ */
+class alignment_file_format_sam
+{
+public:
+    /*!\name Constructors, destructor and assignment
+     * \brief Rule of five explicitly defaulted.
+     * \{
+     */
+    alignment_file_format_sam() = default;
+    alignment_file_format_sam(alignment_file_format_sam const &) = delete;
+    alignment_file_format_sam & operator=(alignment_file_format_sam const &) = delete;
+    alignment_file_format_sam(alignment_file_format_sam &&) = default;
+    alignment_file_format_sam & operator=(alignment_file_format_sam &&) = default;
+    //!\}
+
+    //!\brief The valid file extensions for this format; note that you can modify this value.
+    static inline std::vector<std::string> file_extensions
+    {
+        { "sam" },
+    };
+
+    //!\copydoc alignment_file_output_format_concept::write
+    template <typename stream_type,
+              typename seq_type,
+              typename id_type,
+              typename offset_type,
+              typename ref_seq_type,
+              typename ref_id_type,
+              typename ref_offset_type,
+              typename align_type,
+              typename flag_type,
+              typename mapq_type,
+              typename qual_type,
+              typename mate_type,
+              typename tag_dict_type,
+              typename e_value_type,
+              typename bit_score_type>
+    void write(stream_type                            &  stream,
+               alignment_file_output_options const    &  options,
+               std::unique_ptr<alignment_file_header> &  header_ptr,
+               seq_type                               && seq,
+               qual_type                              && qual,
+               id_type                                && id,
+               offset_type                            && offset,
+               ref_seq_type                           && SEQAN3_DOXYGEN_ONLY(ref_seq),
+               ref_id_type                            && ref_id,
+               ref_offset_type                        && ref_offset,
+               align_type                             && align,
+               flag_type                              && flag,
+               mapq_type                              && mapq,
+               mate_type                              && mate,
+               tag_dict_type                          && tag_dict,
+               e_value_type                           && SEQAN3_DOXYGEN_ONLY(e_value),
+               bit_score_type                         && SEQAN3_DOXYGEN_ONLY(bit_score))
+    {
+        /* Note the following general things:
+         *
+         * - Given the SAM specifications, all fields may be empty
+         *
+         * - Arithmetic values default to 0 while all others default to '*'
+         *
+         * - Because of the former, arithmetic values can be directly streamed
+         *   into 'stream' as operator<< is defined for all arithmetic types
+         *   and the default value (0) is also the SAM default.
+         *
+         * - All other non-arithmetic values need to be checked for emptiness
+         */
+
+        // ---------------------------------------------------------------------
+        // Type Requirements (as static asserts for user friendliness)
+        // ---------------------------------------------------------------------
+        static_assert((std::ranges::ForwardRange<seq_type>        &&
+                      alphabet_concept<value_type_t<remove_cvref_t<seq_type>>>),
+                      "The seq object must be a std::ranges::ForwardRange over "
+                      "letters that model seqan3::alphabet_concept.");
+
+        static_assert((std::ranges::ForwardRange<id_type>         &&
+                      alphabet_concept<value_type_t<remove_cvref_t<id_type>>>),
+                      "The id object must be a std::ranges::ForwardRange over "
+                      "letters that model seqan3::alphabet_concept.");
+
+        static_assert(std::UnsignedIntegral<remove_cvref_t<offset_type>>,
+                      "The offset object must be a std::UnsignedIntegral.");
+
+        static_assert((std::ranges::ForwardRange<ref_seq_type>    &&
+                      alphabet_concept<value_type_t<remove_cvref_t<ref_seq_type>>>),
+                      "The ref_seq object must be a std::ranges::ForwardRange "
+                      "over letters that model seqan3::alphabet_concept.");
+
+        static_assert((std::ranges::ForwardRange<ref_id_type>     &&
+                      alphabet_concept<value_type_t<remove_cvref_t<ref_id_type>>>),
+                      "The ref_id object must be a std::ranges::ForwardRange "
+                      "over letters that model seqan3::alphabet_concept.");
+
+        static_assert(std::Integral<remove_cvref_t<ref_offset_type>>, // -1 is given default to evaluate to 0
+                      "The ref_offset object must be an std::Integral >= 0.");
+
+        if (((ref_offset + 1) < 0))
+            throw format_error("The ref_offset object must be an std::Integral >= 0.");
+
+        static_assert(tuple_like_concept<remove_cvref_t<align_type>>,
+                      "The align object must be a std::pair of two ranges whose "
+                      "value_type is comparable to seqan3::gap");
+
+        static_assert((std::tuple_size_v<remove_cvref_t<align_type>> == 2 &&
+                       std::EqualityComparableWith<gap, value_type_t<remove_cvref_t<decltype(std::get<0>(align))>>> &&
+                       std::EqualityComparableWith<gap, value_type_t<remove_cvref_t<decltype(std::get<1>(align))>>>),
+                      "The align object must be a std::pair of two ranges whose "
+                      "value_type is comparable to seqan3::gap");
+
+        static_assert(std::UnsignedIntegral<remove_cvref_t<flag_type>>,
+                      "The flag object must be a std::UnsignedIntegral.");
+
+        static_assert(std::UnsignedIntegral<remove_cvref_t<mapq_type>>,
+                      "The mapq object must be a std::UnsignedIntegral.");
+
+        static_assert((std::ranges::ForwardRange<qual_type>       &&
+                       alphabet_concept<value_type_t<remove_cvref_t<qual_type>>>),
+                      "The qual object must be a std::ranges::ForwardRange "
+                      "over letters that model seqan3::alphabet_concept.");
+
+        static_assert(tuple_like_concept<remove_cvref_t<mate_type>>,
+                      "The mate object must be a std::tuple of size 3 with "
+                      "1) a std::ranges::ForwardRange with a value_type modelling seqan3::alphabet_concept, "
+                      "2) an std::UnsignedIntegral, and"
+                      "3) an std::UnsignedIntegral.");
+
+        static_assert((std::ranges::ForwardRange<decltype(std::get<0>(mate))>                     &&
+                      alphabet_concept<value_type_t<remove_cvref_t<decltype(std::get<0>(mate))>>> &&
+                      std::UnsignedIntegral<remove_cvref_t<decltype(std::get<1>(mate))>>          &&
+                      std::UnsignedIntegral<remove_cvref_t<decltype(std::get<2>(mate))>>),
+                      "The mate object must be a std::tuple of size 3 with "
+                      "1) a std::ranges::ForwardRange with a value_type modelling seqan3::alphabet_concept, "
+                      "2) an std::UnsignedIntegral, and"
+                      "3) an std::UnsignedIntegral.");
+
+        static_assert(std::Same<remove_cvref_t<tag_dict_type>, sam_tag_dictionary>,
+                      "The tag_dict object must be of type seqan3::sam_tag_dictionary.");
+
+        // ---------------------------------------------------------------------
+        // logical Requirements
+        // ---------------------------------------------------------------------
+        if (!ranges::empty(get<1>(align)) && ranges::empty(seq))
+            throw format_error("If you specify an align object you must also specify the seq object. "
+                               "Hint: Check if offset needs to be set to if soft-clipping is present.");
+
+        if (options.sam_require_header && (header_ptr != nullptr) && !ranges::empty(ref_id))
+        {
+            if ((header_ptr->ref_dict).count(std::string(ref_id)) == 0) // no reference id matched
+                throw format_error(std::string("The ref_id '") + std::string(ref_id) +
+                                   "' was not in the list of references");
+        }
+
+        // ---------------------------------------------------------------------
+        // Writing the Header on first call
+        // ---------------------------------------------------------------------
+        if (options.sam_require_header && !written_header && (header_ptr != nullptr))
+        {
+            write_header(stream, options, header_ptr);
+            written_header = true;
+        }
+
+        // ---------------------------------------------------------------------
+        // Writing the Record
+        // ---------------------------------------------------------------------
+        ranges::ostreambuf_iterator stream_it{stream};
+        char const separator{'\t'};
+
+        write_range(stream_it, std::forward<id_type>(id));
+
+        stream << separator;
+
+        stream << std::forward<flag_type>(flag) << separator;
+
+        write_range(stream_it, std::forward<ref_id_type>(ref_id));
+
+        stream << separator;
+
+        stream << (ref_offset + 1) << separator; // SAM is 1 based
+
+        stream << std::forward<mapq_type>(mapq) << separator;
+
+        if (!ranges::empty(get<1>(align)))
+        {
+            // compute possible distance from alignment end to sequence end
+            // which indicates soft clipping at the end.
+            // This should be replace by a free count_gaps function for
+            // aligned sequences which is more efficient if possible.
+            size_t off_end{seq.size() - offset};
+            for (auto chr : get<1>(align))
+                if (chr == gap::GAP)
+                    ++off_end;
+            off_end -= (get<1>(align)).size();
+
+            write_range(stream_it,
+                        detail::get_cigar_string(std::forward<align_type>(align),
+                                                 std::forward<offset_type>(offset),
+                                                 off_end));
+        }
+        else
+        {
+            stream << '*';
+        }
+
+        stream << separator;
+
+        write_range(stream_it, get<0>(std::forward<mate_type>(mate)));
+
+        stream << separator;
+
+        stream << get<1>(std::forward<mate_type>(mate)) << separator;
+
+        stream << get<2>(std::forward<mate_type>(mate)) << separator;
+
+        write_range(stream_it, std::forward<seq_type>(seq));
+
+        stream << separator;
+
+        write_range(stream_it, std::forward<qual_type>(qual));
+
+        write_tag_fields(stream, std::forward<tag_dict_type>(tag_dict), separator);
+
+        detail::write_eol(stream_it, options.add_carriage_return);
+    }
+
+protected:
+    //!/brief The format version string.
+    static constexpr char format_version[4] = "1.6";
+
+    //!\brief A variable that tracks whether the header_ptr as been written or not.
+    bool written_header{false};
+
+    /*!\brief Writes a field value to the stream.
+     * \tparam stream_it_t The stream iterator type.
+     * \tparam field_type  The type of the field value. Must model std::ranges::ForwardRange.
+     *
+     * \param[in,out] stream_it   The stream iterator to print to.
+     * \param[in]     field_value The value to print.
+     */
+    template <typename stream_it_t, typename field_type>
+        requires std::ranges::ForwardRange<field_type>
+    void write_range(stream_it_t & stream_it, field_type && field_value)
+    {
+        if (ranges::empty(field_value))
+            stream_it = '*';
+        else
+            ranges::copy(field_value | view::to_char | view::take_until(is_space), stream_it);
+    }
+
+    /*!\brief Writes the optional fields of the seqan3::sam_tag_dictionary.
+     * \tparam stream_t   The stream type.
+     *
+     * \param[in,out] stream    The stream to print to.
+     * \param[in]     tag_dict  The tag dictionary to print.
+     * \param[in]     separator The field separator to append.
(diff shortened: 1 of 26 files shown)


include/seqan3/io/alignment_file/output.hpp
```diff
@@ -0,0 +1,940 @@
...
+     *     { "NATA"_dna5, "2nd" },
+     *     { "GATA"_dna5, "Third" }
+     * }; // a range of "records"
+     *
+     * fout = range; // will iterate over the records and write them
+     * ```
+     */
+    template <typename rng_t>
+    alignment_file_out & operator=(rng_t && range)
+        requires std::ranges::InputRange<rng_t> && tuple_like_concept<reference_t<rng_t>>
```
rrahn: value_type_t and please guard for reference inside the rng_t.

include/seqan3/io/alignment_file/output.hpp
```diff
@@ -0,0 +1,940 @@
...
+              std::ranges::InputRange align_type,
+              std::ranges::InputRange flag_type,
+              std::ranges::InputRange mapq_type,
+              std::ranges::InputRange qual_type,
+              std::ranges::InputRange seq_qual_type,
+              std::ranges::InputRange mate_type,
+              std::ranges::InputRange tags_type,
+              std::ranges::InputRange evalue_type,
+              std::ranges::InputRange bit_score_type>
+    void write_columns(header_type         && header,
```
rrahn: indent

include/seqan3/io/alignment_file/output.hpp
```diff
@@ -0,0 +1,940 @@
...
+                      ref_offset_type  && ref_offset,
+                      align_type       && align,
+                      flag_type        && flag,
+                      mapq_type        && mapq,
+                      qual_type        && qual,
+                      seq_qual_type    && seq_qual,
+                      mate_type        && mate,
+                      tags_type        && tags,
+                      evalue_type      && evalue,
+                      bit_score_type   && bit_score)
```
rrahn: could be simplified by parameter packs.

include/seqan3/io/alignment_file/output.hpp
```diff
@@ -0,0 +1,940 @@
...
+                      (detail::decays_to_ignore_v<seq_type> && detail::decays_to_ignore_v<qual_type>),
+                      "You may not select field::SEQ_QUAL and either of field::SEQ "
+                      "and field::QUAL at the same time.");
+
+        assert(!format.valueless_by_exception());
+        std::visit([&] (auto & f)
+        {
+            if constexpr (!detail::decays_to_ignore_v<seq_qual_type>)
+            {
+                f.write(stream,
```
rrahn: I think with some little helper functions it would also be possible to modify the expansion of the pack

include/seqan3/io/alignment_file/output.hpp
rrahn: Also here you could simplify with a parameter pack

include/seqan3/io/alignment_file/output.hpp
```diff
@@ -0,0 +1,940 @@
...
+                       flag_type        && flag,
+                       mapq_type        && mapq,
+                       qual_type        && qual,
+                       seq_qual_type    && seq_qual,
+                       mate_type        && mate,
+                       tags_type        && tags,
+                       evalue_type      && evalue,
+                       bit_score_type   && bit_score)
+    {
+        static_assert(!(detail::decays_to_ignore_v<reference_t<header_type>>     &&
```
rrahn: and this would become a one liner

include/seqan3/io/alignment_file/output.hpp
```diff
@@ -0,0 +1,940 @@
...
+                       detail::decays_to_ignore_v<reference_t<qual_type>>),
+                      "You may not select field::SEQ_QUAL and either of field::SEQ "
+                      "and field::QUAL at the same time.");
+
+        assert(!format.valueless_by_exception());
+        std::visit([&] (auto & f)
+        {
+            if constexpr (!detail::decays_to_ignore_v<reference_t<seq_qual_type>>)
+            {
+                auto zipped = ranges::view::zip(header,
```
rrahn: this too

include/seqan3/io/alignment_file/output.hpp
```diff
@@ -0,0 +1,940 @@
...
+                            std::get<1>(v) | view::convert<typename reference_t<seq_qual_type>::alignment_alphabet_type>,
+                            std::get<2>(v),
+                            std::get<3>(v),
+                            std::get<4>(v),
+                            std::get<5>(v),
+                            std::get<6>(v),
+                            std::get<7>(v),
+                            std::get<8>(v),
+                            std::get<9>(v),
+                            std::get<0>(v) | view::convert<typename reference_t<seq_qual_type>::quality_alphabet_type>,
```
rrahn: wrong index

include/seqan3/io/alignment_file/output.hpp
```diff
@@ -0,0 +1,940 @@
...
+                                                mate,
+                                                tags,
+                                                evalue,
+                                                bit_score);
+
+                for (auto && v : zipped)
+                    f.write(stream,
+                            options,
+                            std::get<0>(v),
+                            std::get<1>(v) | view::convert<typename reference_t<seq_qual_type>::alignment_alphabet_type>,
```
rrahn: Here is a working theory that compiles on godbolt ...
```cpp
template <size_t idx, typename element_t>
auto apply(element_t && elem)
{
    if constexpr (idx == 1)
        return elem + 1;
    else if constexpr (idx == 4)
        return elem + 4;
    else 
        return std::forward<element_t>(elem);
}

template <typename ...args_t, 
          size_t ... is>
void foo_(std::index_sequence<is...>,
          args_t && ... args)
{
    /*f.write(stream, options, */(apply<is>(args),...)/*)*/;
}

template <typename ...args_t>
void foo(args_t && ... args)
{
    foo_(std::make_index_sequence<sizeof...(args_t)>{}, std::forward<args_t>(args)...);
}

int main()
{
    foo(1, 2, 3, 4, 5, 6, 7, 8, 9);
}
```

If we would use this you could simplify the entire code and save a lot of redundancies.

rrahn: Of course adapted to your use case.

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,571 @@
...
+              typename ref_seq_type,
+              typename ref_id_type,
+              typename ref_offset_type,
+              typename align_type,
+              typename flag_type,
+              typename mapq_type,
+              typename qual_type,
+              typename mate_type,
+              typename tag_dict_type,
+              typename evalue_type,
```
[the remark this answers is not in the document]
smehringer: this will be needed for the blast format as is required by the concept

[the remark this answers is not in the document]
rrahn: But call it `e_value_type`. I first stumbled over evaluation something until I realised you mean the e value

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,571 @@
...
+        // ---------------------------------------------------------------------
+        // Type Requirements
+        // ---------------------------------------------------------------------
+        if constexpr (!detail::decays_to_ignore_v<header_type>)
+            static_assert(std::Same<remove_cvref_t<header_type>, alignment_file_header>,
+                          "The header object must be of type seqan3::alignment_file_header.");
+
+        if constexpr (!detail::decays_to_ignore_v<seq_type>)
+            static_assert((std::ranges::ForwardRange<seq_type>        &&
+                          alphabet_concept<value_type_t<seq_type>>),
```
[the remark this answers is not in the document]
rrahn: Or they work, as long as we delegate to the ranges implementation. Not sure, if that is actually a benefit and whether this changes once we remove the dependency to ranges.

include/seqan3/io/alignment_file/format_sam.hpp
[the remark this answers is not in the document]
smehringer: ok

include/seqan3/io/alignment_file/input_options.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+#include <seqan3/core/platform.hpp>
+
+namespace seqan3
+{
+
+//!\brief The options type defines various option members that influence the behaviour of all or some formats.
+struct alignment_file_in_options
+{
+    // options to define
```
h-2: header could also be a member of the options maybe?

include/seqan3/io/alignment_file/output.hpp
h-2: > value_type_t

why? dereferencing the iterator returns reference_t

> and please guard for reference inside the rng_t.

As noted before, our functions should work on references of ranges, it makes sense and simplifies the code.

include/seqan3/io/alignment_file/output.hpp
```diff
@@ -0,0 +1,940 @@
...
+    {
+        f = range;
+        return std::move(f);
+    }
+    //!\}
+
+    /*!\name Tuple interface
+     * \brief Provides functions for field-based ("column"-based) writing.
+     * \{
+     */
```
h-2: I don't think we really need the column-wise interface for alignment-files. It makes sense mostly only for sequence_files...

include/seqan3/io/alignment_file/output.hpp
h-2: I would suggest to drop column-wise support for alignment-file!

include/seqan3/io/exception.hpp
```diff
@@ -86,4 +86,18 @@ struct unexpected_end_of_input : std::runtime_error
...
+// write exceptions
+// ----------------------------------------------------------------------------
+
+//!\brief Thrown if information given to output format didn't match expectations.
+struct format_error : std::invalid_argument
+{
+    //!\brief Constructor that forwards the exception string.
+    format_error(std::string const & s) : std::invalid_argument{s}
+    {}
+};
```
h-2: when and why is this thrown?

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,571 @@
...
+              typename flag_type,
+              typename mapq_type,
+              typename qual_type,
+              typename mate_type,
+              typename tag_dict_type,
+              typename evalue_type,
+              typename bit_score_type>
+    void write(stream_type                      &  stream,
+               alignment_file_out_options const &  options,
+               header_type                      && header,
```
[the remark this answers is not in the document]
smehringer: because we decided on this in another discussion...
the main point was that otherwise the direct range-based functionality does not work. E. g. `file_out = file_in;`

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,571 @@
...
+        {
+            stream << '\t';
+
+            char char0 = tag/256;
+            char char1 = tag - ((tag/256) * 256);
+
+            stream << char0 << char1 << ':' << detail::sam_tag_type_char[variant.index()] << ':';
+
+            if (detail::sam_tag_type_char_extra[variant.index()] != '\0')
+                stream << detail::sam_tag_type_char_extra[variant.index()] << ',';
```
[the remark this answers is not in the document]
smehringer: The SAM specifications say that an optional tag field of type vector is composed of the following:
`TAG:B:EXTRA,...` where EXTRA is an additional char indicating the value type (one of `[c,C,s,S,i,I,f]`)

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,571 @@
...
+                    for (auto it = arg.begin(); it != (arg.end() - 1); ++it)
+                        stream << *it << ",";
+
+                    stream << *(arg.end() - 1); // write last value without trailing ','
+                }
+            }
+        };
+
+        // can I check at runtime which sam_tag_type_t tag has ?
+        // such that I can check if the value in variant was set correctly.
```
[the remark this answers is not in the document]
smehringer: but `index()` is a runtime parameter and I only have the sam_tag_type struct that needs to be specialized with a non-type template parameter

include/seqan3/io/alignment_file/header.hpp
```diff
@@ -0,0 +1,76 @@
...
+    {
+        std::string id;
+        std::string name;
+        std::string command_line_call;
+        std::string previous;
+        std::string description;
+        std::string version;
+    };
+
+    std::string file_format; // should be set by format
```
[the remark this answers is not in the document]
smehringer: yes. We follow the current SAM specs which sets the version to 1.6

include/seqan3/io/alignment_file/header.hpp
```diff
@@ -0,0 +1,76 @@
...
+    std::string sorting_order{"unknown"}; // enum? for SAM
+    std::string grouping{"none"};        // enum? for SAM
+
+    // {id, name, cmd, prev, desc, version}
+    std::vector<program_info_t> program_infos;
+
+    std::vector<std::string> comments;
+
+    // reference sequence dictionary of SAM format
+    std::vector<std::tuple<std::string, uint32_t, std::string>> ref_dict; // {ref_name, length, alt_loci, all optional fields as a simple string}
```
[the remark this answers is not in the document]
smehringer: because the header is not required and if not supplied the read records still need the reference name as string.

include/seqan3/io/exception.hpp
smehringer: when the header is ill-formed and thus a correct SAM file cannot be written

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,571 @@
...
+            if(!written_header)
+            {
+                check_header_information(header);
+                write_header(stream, header, options);
+                written_header = true;
+            }
+
+            // if all other fields decay to std::ignore, only the header
+            // should be printed and not an all-empty-record.
+            if constexpr (detail::decays_to_ignore_v<seq_type> &&
```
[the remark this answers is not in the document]
smehringer: the header is only required in BAM format but optional in SAM format. We wanted to keep the header in the record because we don't want to disrupt the range-based functionality by the need of first calling a read/write header function, especially because for BAM files, reading the header is a must as correct parsing is not possible otherwise.
But when the user only calls file_out = {header} it would be very unintuitive if the header and an empty record is printed.. don't you think?

include/seqan3/io/alignment_file/input_options.hpp
smehringer: But then how does the input file communicate the header with the output file as their options are distinct? Except if there is one step which we did not want to keep the range-based functionality..

include/seqan3/io/alignment_file/output.hpp
smehringer: does this consider blast files? I agree that SAM records, bloated as they are, are probably never used in column-based reading/writing. I don't know about blast so I just wanted to check.

include/seqan3/io/exception.hpp
smehringer: or if the reference name specified in the record does not match any reference name in the header.ref_dict

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,571 @@
...
+ *
+ *   * ID lines beginning with `;` instead of `>`
+ *   * line breaks and other whitespace characters in any part of the alignment
+ *   * character counts within the alignment (they are simply ignored)
+ *
+ * The following optional features are currently **not supported:**
+ *
+ *   * Multiple comment lines (starting with either `;` or `>`), only one ID line before the alignment line is accepted
+ *
+ */
```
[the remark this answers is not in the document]
smehringer: yes, i did not want to update any documentation if the design is not fixed yet

include/seqan3/io/alignment_file/format_sam.hpp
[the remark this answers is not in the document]
smehringer: This is especially important considering BAM files: Here, the header is required because the reference names are stored as numbers and the ref_dict is the only way to restore the names. When reading a BAM file without reading the header first, accessing the ref_name will result in the wrong behaviour.

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,571 @@
...
+         *
+         * - Given the SAM specifications, all fields may be empty
+         *
+         * - Arithmetic values default to 0 while all others default to '*'
+         *
+         * - Because of the former, arithmetic values can be directly streamed
+         *   into 'stream' as operator<< is defined for all arithmetic types
+         *   and the default value (0) is also the SAM default.
+         *
+         * - All other non-arithmetic values need to be checked for emptiness
```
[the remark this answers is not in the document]
smehringer: do you mean that the file class should pass default initialized fields?

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,571 @@
...
+
+        // ---------------------------------------------------------------------
+        // Type Requirements
+        // ---------------------------------------------------------------------
+        if constexpr (!detail::decays_to_ignore_v<header_type>)
+            static_assert(std::Same<remove_cvref_t<header_type>, alignment_file_header>,
+                          "The header object must be of type seqan3::alignment_file_header.");
+
+        if constexpr (!detail::decays_to_ignore_v<seq_type>)
+            static_assert((std::ranges::ForwardRange<seq_type>        &&
```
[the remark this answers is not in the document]
smehringer: so what now? change or leave it?

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,571 @@
...
+            {
+                auto ref = header.ref_dict | view::get<0>
+                                           | ranges::view::remove_if([&ref_id] (std::string & s) {return s != ref_id;});
+                if (ref.begin() == ref.end()) // no reference id matched
+                    throw format_error(std::string("The ref_id '") + ref_id + "' was not in the list of references");
+            }
+
+            if(!written_header)
+            {
+                check_header_information(header);
```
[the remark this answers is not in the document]
smehringer: The alignment_file_header object is quite unrestrictive and general as it takes mostly vectors and strings instead of a bloated enum/class/struct interface for simplicity. If we want to enforce that only valid SAM files are written with SeqAn we need to check the header object for correctness. Since this involves many things, I didn't do it yet for time reasons and just in case you have a different opinion on the checks.

include/seqan3/io/alignment_file/format_sam.hpp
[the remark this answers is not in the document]
h-2: > because we decided on this in another discussion...

Yes, I remember, thanks for reminding me! Still I would make the header a member of the file only include a pointer to it in the record. I would then just make a check directly at the beginning of the first write call so that it is written.  But there shouldn't be an extra write-call with std::ignores just for the header (I don't know if I maybe misunderstood that from your comments)

include/seqan3/io/alignment_file/format_sam.hpp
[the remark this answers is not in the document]
h-2: Yes, if the behaviour for `std::ignore` and "" is exactly the same and if there are no illegal combinations of unselected fields that could/should be detected at compile-time, we might as well get rid of the std::ignores!

include/seqan3/io/alignment_file/format_sam.hpp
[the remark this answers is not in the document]
h-2: I would implement the checks inside the `write_header()` function and throw exceptions there.

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,571 @@
...
+                return;
+        }
+
+        ranges::ostreambuf_iterator stream_it{stream};
+
+        if constexpr (detail::decays_to_ignore_v<id_type>)
+            stream_it = '*';
+        else
+        {
+            write_field(stream_it, std::string{id});
```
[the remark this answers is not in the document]
h-2: oh, but you are constructing a new string here? That's expensive if you do it often.

include/seqan3/io/alignment_file/format_sam.hpp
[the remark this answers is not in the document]
h-2: Hm, I don't understand this, but we can discuss it tomorrow

include/seqan3/io/alignment_file/header.hpp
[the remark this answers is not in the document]
h-2: then maybe call it file_format_revision or something like that? because the "file format" is the thing you are implementing in `format_sam.hpp`.

include/seqan3/io/alignment_file/header.hpp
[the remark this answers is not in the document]
h-2: Hm, but this is inconsisten. Why would I specify the header, if I need to specify the variable anyway later? What would happen if you specify two different values?

While it is not required for SAM to have the header in the file, nothing is stopping us from requiring it for our data structure (note that this is independent of whether we write it to the header or not).

include/seqan3/io/alignment_file/input_options.hpp
h-2: true, strike that comment.

include/seqan3/io/alignment_file/output.hpp
h-2: I don't think it's relevant for BLAST. And if it turns out that it is, we can still add it later.

include/seqan3/io/alignment_file/format_sam.hpp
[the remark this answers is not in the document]
smehringer: the problem was that for if passing a `const char *` then ranges::copy prints the null character.

include/seqan3/io/alignment_file/format_sam.hpp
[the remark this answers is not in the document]
smehringer: Ah now I see what you meant! Thanks for clarifying.

include/seqan3/io/alignment_file/output.hpp
smehringer: Yes but no because some of the types/values are treated special and would need to passed as single paramters (namely header, seq, qual, seq_qual). I could put all of those in the beginning and pack the rest ?

include/seqan3/io/alignment_file/format_sam.hpp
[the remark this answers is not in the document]
smehringer: Unfortunately it is not.. 
1) If the alignment type does not decay to ignore, the seq type and offset type mustn't either
2) If neither the header type nor the ref_id type decay to ignore, the ref_id must be checked against the ref_dict

In addition, I think default values are not possible since they differ between BLAST and SAM. The BLAST format will just not print the column whereas SAM print default values.

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,585 @@
...
+ *
+ * The format checks are implemented according to the official
+ * [SAM format specializations](https://samtools.github.io/hts-specs/SAMv1.pdf)
+ * in order to ensure correct SAM file output.
+ *
+ * If any error is detected, a seqan3::format_error is thrown. There is only
+ * **one exception**: If the sequences like `id`, `ref_id`, `seq`, etc. ...
+ * contain any illegal white space character (seqan3::is_space) they are truncated
+ * up until its occurrence instead of throwing an error. This avoids severe
+ * format errors for the tab separated SAM format.
```
h-2: I would instead write:
```
Note that SAM does not allow whitespace in the sequence ID so SeqAn truncates these strings at the first whitespace when writing SAM.
```
(it doesn't throw because it is not an error)

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,585 @@
...
+ *
+ * ### Header implementation
+ *
+ * The header_ptr (if set in the seqan3::alignment_file_output or given by a pointer
+ * in the seqan3::field::HEADER field) is only printed once in the beginning
```
h-2: the header_pointer is given by a pointer? Sounds kind of weird. Since all of the discussed logic is part of the file, I would explain it there and just mention here that the SAM header is automatically written with the first record based on the information stored in the alignment_file.

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,585 @@
...
+                      "The ref_id object must be a std::ranges::ForwardRange "
+                      "over letters that model seqan3::alphabet_concept.");
+
+        static_assert(std::Integral<remove_cvref_t<ref_offset_type>>, // -1 is given default to evaluate to 0
+                      "The ref_offset object must be an std::Integral >= 0.");
+
+        if (((ref_offset + 1) < 0))
+            throw format_error("The ref_offset object must be an std::Integral >= 0.");
+
+        static_assert(tuple_like_concept<align_type>,
```
h-2: and size == 2

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,585 @@
...
+
+        if (header_ptr != nullptr && !ranges::empty(ref_id))
+        {
+            auto ref = header_ptr->ref_dict | view::get<0>
+                                        | ranges::view::remove_if([&ref_id] (auto & s) {return s != ref_id;});
+
+            if (ref.begin() == ref.end()) // no reference id matched
+                throw format_error(std::string("The ref_id '") + std::string(ref_id) +
+                                   "' was not in the list of references");
+        }
```
h-2: std::unordered_map

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,585 @@
...
+
+        // ---------------------------------------------------------------------
+        // Writing the Header on first call
+        // ---------------------------------------------------------------------
+        if (!written_header_ptr && header_ptr != nullptr)
+        {
+            write_header_ptr(stream, options, header_ptr);
+            written_header_ptr = true;
+        }
```
h-2: throw if `(!written_header_ptr && header_ptr == nullptr)` ?

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,585 @@
...
+
+        // ---------------------------------------------------------------------
+        // Writing the Record
+        // ---------------------------------------------------------------------
+        ranges::ostreambuf_iterator stream_it{stream};
+        const char separator{'\t'};
```
h-2: `char const`

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,585 @@
...
+ *
+ * \details
+ *
+ * ### Introduction
+ *
+ * SAM is often used for storing alignments of several read sequences against one
+ * or more reference sequences. See the
+ * [article on wikipedia](https://en.wikipedia.org/wiki/SAM_(file_format)) for an
+ * introduction of the format or look into the official
+ * [SAM format specializations](https://samtools.github.io/hts-specs/SAMv1.pdf).
```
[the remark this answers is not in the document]
h-2: and add "SeqAn implements version 1.6 of the SAM specification."

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,585 @@
...
+     */
+    template <typename stream_it_t, typename field_type>
+        requires std::ranges::ForwardRange<field_type>
+    void write_field(stream_it_t & stream_it, field_type && field_value, char const separator)
+    {
+        write_field(stream_it, std::forward<field_type>(field_value));
+        stream_it = separator;
+    }
+
+
```
h-2: double newline

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,585 @@
...
+     template <typename stream_it_t, typename field_type>
+        requires std::ranges::ForwardRange<field_type>
+    void write_field(stream_it_t & stream_it, field_type && field_value)
+    {
+        if (ranges::empty(field_value))
+            stream_it = '*';
+        else
+            ranges::copy(field_value | view::to_char
+                                     | view::take_until(is_space)
+                         , stream_it);
```
h-2: just write this into one line maybe? If it doesnt fit, correct the indention/spacing and enclose the block in braces

h-2: I would also rename this to `write_range` maybe, because it can't write other fields. And it is also only called once and it is confusing that the above is called the same, but does something different semantically (usually if you omit one parameter, it means that this is defaulted, but here it means that it is not written; thats inconsistent).

Maybe just remove the first interface and print the separator locally.

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,585 @@
...
+     *         information or the contents are ill-formed.
+     *
+     * \details
+     *
+     * Before writing the header_ptr, the contents are checked for correctness
+     * according to the rules of the official
+     * [SAM format specializations](https://samtools.github.io/hts-specs/SAMv1.pdf).
+     */
+    template <typename stream_t>
+    void write_header_ptr(stream_t & stream,
```
h-2: just call this `write_header` not `write_header_ptr`

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,585 @@
...
+            // -----------------------------------------------------------------
+            // Check Header
+            // -----------------------------------------------------------------
+
+            // (@HD) Check header_ptr line
+            // The format version string will be taken from the local member variable
+            if (!(header_ptr->sorting != "unknown"   ||
+                  header_ptr->sorting != "unsorted"  ||
+                  header_ptr->sorting != "queryname" ||
+                  header_ptr->sorting != "coordinate" ))
```
h-2: not not

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,585 @@
...
+                throw format_error("SAM format error: The header_ptr->sorting member must be "
```
h-2: enclose in braces

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,585 @@
...
+                                   "one of [unknown, unsorted, queryname, coordinate].");
+
+            if (!(header_ptr->grouping != "none"   ||
+                  header_ptr->grouping != "query"  ||
+                  header_ptr->grouping != "reference"))
+                throw format_error("SAM format error: The header_ptr->grouping member must be "
+                                   "one of [none, query, reference].");
```
h-2: not not and braces

include/seqan3/io/alignment_file/output.hpp
```diff
@@ -0,0 +1,687 @@
...
+
+    //!\brief The options are public and its members can be set directly.
+    alignment_file_output_options options;
+
+    /*!\brief The file header object.
+     *
+     * ### Example
+     * \snippet test/snippet/io/alignment_file/output.cpp set_header
+     */
+    std::unique_ptr<alignment_file_header> header{nullptr};
```
h-2: make this private and add a member function `alignment_file_header & header()` that returns a reference to the contained header and initialises the pointer if it is `nullptr`. Also use this function when passing the header to the format, thereby we hide all the pointer semantics from the user and also the format.

include/seqan3/io/alignment_file/output.hpp
```diff
@@ -0,0 +1,687 @@
...
+
+    //!\brief The actual std::variant holding a pointer to the detected/selected format.
+    format_type format;
+
+    //!\brief Write record to format.
+    template <typename ...pack_type>
+    void write_record(alignment_file_header const *  hdr_ptr, pack_type && ...remainder)
+    {
+        if (header == nullptr && hdr_ptr != nullptr)
+            header = std::unique_ptr<alignment_file_header>(new alignment_file_header(*hdr_ptr));
```
h-2:
```
header = new alignment_file_header{*hdr_ptr};
```

include/seqan3/io/alignment_file/output.hpp
```diff
@@ -0,0 +1,687 @@
...
+
+    //!\brief Type of the format, an std::variant over the `valid_formats`.
+    using format_type = detail::transfer_template_args_onto_t<valid_formats, std::variant>;
+
+    //!\brief The actual std::variant holding a pointer to the detected/selected format.
+    format_type format;
+
+    //!\brief Write record to format.
+    template <typename ...pack_type>
+    void write_record(alignment_file_header const *  hdr_ptr, pack_type && ...remainder)
```
h-2: double space before hdr_ptr

include/seqan3/io/alignment_file/output_format_concept.hpp
```diff
@@ -0,0 +1,209 @@
...
+//!\endcond
+
+/*!\name Requirements for seqan3::alignment_file_output_format_concept
+ * \brief You can expect these **members** on all types that implement seqan3::alignment_file_output_format_concept.
+ * \memberof seqan3::alignment_file_output_format_concept
+ * \{
+ */
+
+/*!\fn void write(stream_type                            &  stream,
+                  alignment_file_output_options const       &  options,
```
h-2: indentions

include/seqan3/io/alignment_file/output_options.hpp
```diff
@@ -0,0 +1,56 @@
...
+namespace seqan3
+{
+
+//!\brief The options type defines various option members that influence the behaviour of all or some formats.
+struct alignment_file_output_options
+{
+    /*!\brief The default plain text line-ending is "\n", but on Windows an
+     *        additional carriage return is recommended ("\r\n" for line-ending).
+     */
+    bool        add_carriage_return     = false;
```
h-2: add some TODOs here, maybe `bool sam_write_header = true;`

test/unit/io/alignment_file/alignment_file_output_test.cpp
```diff
@@ -0,0 +1,415 @@
...
+                                                            alignment_file_format_sam{}} ));
+
+
+    /* stream + format_tag + fields */
+    EXPECT_NO_THROW(( alignment_file_output<fields<field::SEQ, field::ID, field::QUAL>,
+                                        type_list<alignment_file_format_sam>,
+                                        std::ostringstream>{std::ostringstream{},
+                                                            alignment_file_format_sam{},
+                                                            fields<field::SEQ, field::ID, field::QUAL>{}} ));
+}
```
h-2: many of the previous examples have wrong indent

include/seqan3/io/alignment_file/format_sam.hpp
smehringer: only in BAM format ?

include/seqan3/io/alignment_file/output.hpp
smehringer: does not compile

include/seqan3/io/alignment_file/format_sam.hpp
h-2: whether or not to write the header should be an option and not depend on whether or not a header is available.

include/seqan3/io/alignment_file/output.hpp
```diff
@@ -0,0 +1,706 @@
...
+ *   8. field::MAPQ
+ *   9. field::FLAG
+ *   10. field::QUAL
+ *   11. field::MATE
+ *   12. field::TAGS
+ *   13. field::EVALUE
+ *   14. field::BIT_SCORE
+ *
+ * There is an additional field called `field::HEADER_PTR`, it used to transfer
+ * header information from `alignment_file_input` to `alignment_file_output`,
```
h-2: it **is** used

no backticks around the stuff, but instead `seqan3::` before the files. always include namespace to make sure that links are generated :point_up:

include/seqan3/io/alignment_file/output.hpp
```diff
@@ -0,0 +1,706 @@
...
+ * For opening from file, `alignment_file_output<>` would have also worked, but for
+ * opening from stream it would not have.
+ *
+ * ### Writing record-wise
+ *
+ * \snippet test/snippet/io/alignment_file/record_based_writing.cpp all
+ *
+ * The easiest way to write to an alignment file is to use the push_back() member functions. These
+ * work similarly to how they work on an std::vector. If you pass a tuple to push_back() the seqan3::field ID
+ * of the i-th tuple-element/argument is assumed to be the i-th value of selected_field_ids.
```
h-2: like I said before, please point out that push_back(record) should be used and not the tuple interface.

include/seqan3/io/alignment_file/output.hpp
```diff
@@ -0,0 +1,706 @@
...
+    /*!\cond DEV
+     * \brief Expose a reference to the underlying stream object. [public, but not documented as part of the API]
+     */
+    stream_type & get_stream()
+    {
+        return stream;
+    }
+    //!\endcond
+
+    /*!\brief Return a reference to the header object of the file.
```
h-2: Maybe just "Access the file's header." ?

include/seqan3/io/alignment_file/output.hpp
```diff
@@ -0,0 +1,706 @@
...
+     *
+     * \details
+     *
+     * If the header object was not set before, the function will create a
+     * default seqan3::alignment_file_header object.
```
h-2: users don't need to know about any header objects, this is detail

include/seqan3/io/alignment_file/output.hpp
```diff
@@ -0,0 +1,706 @@
...
+     *
+     * \sa seqan3::alignment_file_header
+     */
+    alignment_file_header & header()
+    {
+        if (header_ptr == nullptr)
+            header_ptr = std::unique_ptr<alignment_file_header>(new alignment_file_header);
+
+        return *header_ptr;
+    }
```
h-2: :+1:

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,576 @@
...
+        }
+
+        // ---------------------------------------------------------------------
+        // Writing the Header on first call
+        // ---------------------------------------------------------------------
+        if (!written_header && header_ptr != nullptr)
+        {
+            write_header(stream, options, header_ptr);
+            written_header = true;
+        }
```
h-2: I think it should still throw, because it is easy to forget to edit the header. If you really don want to print it, you can set an extra option `sam_write_header = false;` as a suggested before.

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,576 @@
...
+     */
+    template <typename stream_it_t, typename field_type>
+        requires std::ranges::ForwardRange<field_type>
+    void write_field(stream_it_t & stream_it, field_type && field_value, char const separator)
+    {
+        write_range(stream_it, std::forward<field_type>(field_value));
+        stream_it = separator;
+    }
+
+
```
h-2: double newline

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -0,0 +1,576 @@
...
+    /*!\brief Writes a field value to the stream.
+     * \tparam stream_it_t The stream iterator type.
+     * \tparam field_type  The type of the field value. Must model std::ranges::ForwardRange.
+     *
+     * \param[in,out] stream_it   The stream iterator to print to.
+     * \param[in]     field_value The value to print.
+     */
+    template <typename stream_it_t, typename field_type>
+        requires std::ranges::ForwardRange<field_type>
+    void write_range(stream_it_t & stream_it, field_type && field_value)
```
h-2: now there are two different functions? write_range AND write_field?

I feel like you did not address my previous points, yet...

smehringer: I deleted the field function and now the separator is printed separately

smehringer:
:tada: :tada:

Issue #444, opened 2018-09-17
loosen alphabet_concept requirements?


h-2:
Although it makes sense to design alphabets as trivial types and standard-layout-types, I am no longer sure that we should require this. Especially when dealing with external datatypes it is really difficult to ensure triviality.

In fact I am deactivating the trivial requirement now, because of dealing with SDSL-types. 

We need to re-evaluate this after the workshops.

marehr:
> In fact I am deactivating the trivial requirement now, because of dealing with SDSL-types.

Can you give an example where we use sdsl-types as alphabets?

h-2:
we use sdsl::int_vector_reference inside the proxy type returned by bitcompressed_vector and I assume we will have more of this when mergen the gap_adaptors.

h-2:
(having a non-trivial type inside your alphabet also makes it non-trivial)

h-2:
Sure it is convertible, but that doesnt help when you have constrained type deduction and you can't infer which concrete alphabet it would be convertible to.

h-2:
changed via #535

PR #446, opened 2018-09-17
Fix/concatenated

### LINKED ISSUE #395: Error in sequence_file_in
```cpp
struct my_traits : public sequence_file_in_default_traits_dna
{
    using sequence_alphabet = dna4;                        // instead of dna5

    template <typename alph>
    using sequence_container = bitcompressed_vector<alph>; // must be defined as a template!
};

sequence_file_in<my_traits> fin{"/tmp/my.fasta"};

//...
```

Produces errors about sequence_container_concept, among other strange things.

h-2:
fixes #395 

(actually a bunch of different problems)

### DIFF
--- a/include/seqan3/alphabet/concept.hpp
+++ b/include/seqan3/alphabet/concept.hpp
@@ -91,7 +91,7 @@ namespace seqan3
 template <typename t>
 concept semi_alphabet_concept = std::Regular<std::remove_reference_t<t>> &&
                                      standard_layout_concept<std::remove_reference_t<t>> &&
-                                     trivial_concept<std::remove_reference_t<t>> &&
+//                                      trivial_concept<std::remove_reference_t<t>> &&
                                      std::StrictTotallyOrdered<t> &&
                                      requires (t t1, t t2)
 {
@@ -203,3 +203,4 @@ void CEREAL_LOAD_MINIMAL_FUNCTION_NAME(archive_t const &,
  */
 
 } // namespace seqan3
+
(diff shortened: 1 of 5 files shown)


h-2:
I advise going through the commits one-by-one (they are very small individually)

include/seqan3/range/container/concatenated_sequences.hpp
```diff
@@ -923,9 +926,18 @@ class concatenated_sequences
             value_len = std::distance(ranges::begin(value), ranges::end(value));
 
         data_values.reserve(data_values.size() + count * value_len);
+        auto placeholder = ranges::view::repeat_n(value_type_t<rng_type>{}, count * value_len)
+                         | view::common;
```
marehr: The commit message is `[FIX] ensure that ranges passed on are actually ForwardRanges `, but I can't find where this is actually required in your first commit.

rrahn: The pipe symbol on the previous line

include/seqan3/alphabet/concept.hpp
```diff
@@ -91,7 +91,7 @@ namespace seqan3
 template <typename t>
 concept semi_alphabet_concept = std::Regular<std::remove_reference_t<t>> &&
                                      standard_layout_concept<std::remove_reference_t<t>> &&
-                                     trivial_concept<std::remove_reference_t<t>> &&
+//                                      trivial_concept<std::remove_reference_t<t>> &&
```
rrahn: why is this necessary?

include/seqan3/range/container/concatenated_sequences.hpp
h-2: this calls the same functions on the underlying range and they require forward

include/seqan3/alphabet/concept.hpp
h-2: see #444

PR #442, opened 2018-09-17
Feature/align cfg edit

rrahn:
rebased on #441 and #440

PR #448, opened 2018-09-17
Feature/edit distance with align cfg

rrahn:
Fixes the edit distance to use the align_cfg object.

Introduces a traits class for the edit distance algorithm.
Frankly speaking, I don't see why the user should configure the word_type, so I added this config as a traits type which defaults to uint64_t so that the tests still work as expected. If we think this should be configured by the user, we can pull it up into the align_cfg.

This is rebased on #440 #441 #442. So it also supersedes them, and can be read commit by commit.


include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
```diff
@@ -236,6 +264,17 @@ struct pairwise_alignment_edit_distance_unbanded
...
+
+    pairwise_alignment_edit_distance_unbanded(
+        database_t _database,
+        query_t _query,
+        align_config_t _config,
+        traits_t /*_traits*/) :
+            pairwise_alignment_edit_distance_unbanded(std::forward<database_t>(_database),
+                                                      std::forward<query_t>(_query),
+                                                      std::forward<align_config_t>(_config))
+    {}
```
marehr: This constructor is not needed. When someone wants to use the `traits_t` parameter (which should only when testing the code) he should specify everything.

include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
```diff
@@ -190,8 +218,8 @@ struct pairwise_alignment_edit_distance_unbanded
      * \param[in] _config   \copydoc config
      */
     pairwise_alignment_edit_distance_unbanded(
-        database_t _database,
-        query_t _query,
+        database_t && _database,
+        query_t && _query,
```
marehr: I had trouble when using forwarding references, but if everything works now I'm okay with this.

include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
```diff
@@ -343,8 +382,42 @@ struct pairwise_alignment_edit_distance_unbanded
...
+        else if constexpr (std::tuple_size_v<result_type> == 4)
+        {
+            throw std::invalid_argument{"The current output setting is not supported!"};
+        }
+        else if constexpr (std::tuple_size_v<result_type> == 5)
+        {
+            //TODO:
+            auto trace = alignment_trace(database, query, trace_matrix());
+            ranges::copy(std::get<0>(trace), ranges::back_inserter(std::get<0>(get<align_result_key::trace>(res))));
+            ranges::copy(std::get<1>(trace), ranges::back_inserter(std::get<1>(get<align_result_key::trace>(res))));
```
marehr: Not sure why you copy here, because trace is already a sequence which satisfies `aligned_sequence_concept`.

include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
```diff
@@ -343,8 +382,42 @@ struct pairwise_alignment_edit_distance_unbanded
...
+            // TODO: Compute the end coordinate
+            // TODO: Compute the begin coordinate
+            get<align_result_key::score>(res) = score();
+        }
+        return res;
+    }
+
     //!\brief Compute the alignment.
-    pairwise_alignment_edit_distance_unbanded& run()
+    decltype(auto) run()
```
marehr: Because of shortness sake or for what?

include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
```diff
@@ -464,6 +537,10 @@ bool pairwise_alignment_edit_distance_unbanded<database_t, query_t, align_config
 template<typename database_t, typename query_t, typename config_t>
 pairwise_alignment_edit_distance_unbanded(database_t && database, query_t && query, config_t config)
     -> pairwise_alignment_edit_distance_unbanded<database_t, query_t, config_t>;
+
+template<typename database_t, typename query_t, typename config_t, typename traits_t>
+pairwise_alignment_edit_distance_unbanded(database_t && database, query_t && query, config_t config, traits_t)
+    -> pairwise_alignment_edit_distance_unbanded<database_t, query_t, config_t, traits_t>;
```
marehr: can be removed.

include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
rrahn: Mhmm, I mean we check the sequences to be of ViewableRange concept. So either they are views and are copied or they are l-value refs of ranges and the ref is by the deduction guide part of the type, right. So the ref can be removed here. I am always confused by this.

PR #454, opened 2018-09-18
edit_distance now computes begin position

marehr:
Changes:
* computes begin_position
* `edit_distance{}(result_type)` supports now all `result_type` modes and properly sets the return values
* `align_result` expects now the `alignment_coordinate` type
* `alignment_trace` won't use `lists` anymore

### DIFF
--- a/include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
+++ b/include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
@@ -46,8 +46,7 @@
 #include <seqan3/alignment/matrix/alignment_score_matrix.hpp>
 #include <seqan3/alphabet/gap/gapped.hpp>
 #include <seqan3/core/add_enum_bitwise_operators.hpp>
-#include <seqan3/core/metafunction/basic.hpp>
-#include <seqan3/core/metafunction/iterator.hpp>
+#include <seqan3/core/metafunction/range.hpp>
 
 namespace seqan3::detail
 {
--- a/include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
+++ b/include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
@@ -157,13 +157,17 @@ alignment_begin_coordinate(trace_matrix_t && matrix, alignment_coordinate const 
 
 /*!\brief Compute the trace from a trace matrix.
  * \ingroup alignment_matrix
- * \tparam    database_t           The type of the database sequence.
- * \tparam    query_t              The type of the query sequence.
- * \tparam    trace_matrix_t       The type of the trace matrix.
- * \param[in] database             The database sequence.
- * \param[in] query                The query sequence.
- * \param[in] matrix               The trace matrix.
- * \param[in] end_coordinate       Where the trace in the matrix ends.
+ * \tparam    database_t                 The type of the database sequence.
+ * \tparam    query_t                    The type of the query sequence.
+ * \tparam    trace_matrix_t             The type of the trace matrix.
+ * \cond DEV
+ * \tparam    gapped_database_alphabet_t The alphabet type of the gapped database sequence.
+ * \tparam    gapped_query_alphabet_t    The alphabet type of the gapped query sequence.
+ * \endcond
+ * \param[in] database                   The database sequence.
+ * \param[in] query                      The query sequence.
+ * \param[in] matrix                     The trace matrix.
+ * \param[in] end_coordinate             Where the trace in the matrix ends.
  */
 template <
     typename database_t,
@@ -176,11 +180,10 @@ template <
              std::Same<typename remove_cvref_t<trace_matrix_t>::entry_type, trace_directions>
 //!\endcond
 inline std::pair<std::vector<gapped_database_alphabet_t>, std::vector<gapped_query_alphabet_t>>
-alignment_trace(
-    database_t && database,
-    query_t && query,
-    trace_matrix_t && matrix,
-    alignment_coordinate const end_coordinate)
+alignment_trace(database_t && database,
+                query_t && query,
+                trace_matrix_t && matrix,
+                alignment_coordinate const end_coordinate)
 {
     using signed_size_t = std::make_signed_t<size_t>;
 
--- a/include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
+++ b/include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
@@ -105,7 +105,7 @@ struct alignment_coordinate
  * \tparam    trace_matrix_t The type of the trace matrix.
  * \param[in] matrix         The trace matrix.
  * \param[in] end_coordinate Where the trace in the matrix ends.
- * \return Returns the begin coordinate.
+ * \returns Returns the begin coordinate.
  */
  template <typename trace_matrix_t>
  //!\cond
@@ -169,7 +169,7 @@ inline alignment_coordinate alignment_begin_coordinate(trace_matrix_t && matrix,
  * \param[in] query                      The query sequence.
  * \param[in] matrix                     The trace matrix.
  * \param[in] end_coordinate             Where the trace in the matrix ends.
- * \return Returns a seqan3::aligned_sequence.
+ * \returns Returns a seqan3::aligned_sequence.
  */
 template <
     typename database_t,
--- a/include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
+++ b/include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
@@ -39,11 +39,12 @@
 
 #pragma once
 
-#include <list>
-#include <seqan3/alphabet/gap/gapped.hpp>
+#include <deque>
+#include <vector>
 
 #include <seqan3/alignment/matrix/matrix_concept.hpp>
 #include <seqan3/alignment/matrix/alignment_score_matrix.hpp>
+#include <seqan3/alphabet/gap/gapped.hpp>
 #include <seqan3/core/add_enum_bitwise_operators.hpp>
 #include <seqan3/core/metafunction/basic.hpp>
 
@@ -98,64 +99,128 @@ struct alignment_coordinate
     size_t seq2_pos;
 };
 
-/*!\brief Compute the trace from a trace matrix
+/*!\brief Compute the begin coordinate.
  * \ingroup alignment_matrix
- * \tparam    database_t     The type of the database sequence.
- * \tparam    query_t        The type of the query sequence.
  * \tparam    trace_matrix_t The type of the trace matrix.
- * \param[in] database       The database sequence.
- * \param[in] query          The query sequence.
  * \param[in] matrix         The trace matrix.
  * \param[in] end_coordinate Where the trace in the matrix ends.
+ */
+ template <typename trace_matrix_t>
+ //!\cond
+     requires matrix_concept<remove_cvref_t<trace_matrix_t>> &&
+              std::Same<typename remove_cvref_t<trace_matrix_t>::entry_type, trace_directions>
+ //!\endcond
+inline alignment_coordinate
+alignment_begin_coordinate(trace_matrix_t && matrix, alignment_coordinate const end_coordinate)
+{
+    using signed_size_t = std::make_signed_t<size_t>;
+
+    constexpr auto N = trace_directions::none;
+    constexpr auto D = trace_directions::diagonal;
+    constexpr auto L = trace_directions::left;
+    constexpr auto U = trace_directions::up;
+    signed_size_t row = end_coordinate.seq2_pos + 1;
+    signed_size_t col = end_coordinate.seq1_pos + 1;
+
+    assert(row < matrix.rows());
+    assert(col < matrix.cols());
+
+    while(true)
+    {
+        trace_directions dir = matrix.at(row, col);
+        if ((dir & L) == L)
+        {
+            col = std::max<signed_size_t>(col - 1, 0);
+        }
+        else if ((dir & U) == U)
+        {
+            row = std::max<signed_size_t>(row - 1, 0);
+        }
+        else if ((dir & D) == D)
+        {
+            row = std::max<signed_size_t>(row - 1, 0);
+            col = std::max<signed_size_t>(col - 1, 0);
+        }
+        else
+        {
+            if (row == 0 || col == 0)
+                break;
+            throw std::logic_error{"Trace not found"};
+        }
+    }
+    return
+    {
+        std::max<signed_size_t>(col - 1, 0),
+        std::max<signed_size_t>(row - 1, 0)
+    };
+}
+
+/*!\brief Compute the trace from a trace matrix.
+ * \ingroup alignment_matrix
+ * \tparam    database_t           The type of the database sequence.
+ * \tparam    query_t              The type of the query sequence.
+ * \tparam    trace_matrix_t       The type of the trace matrix.
+ * \param[in] database             The database sequence.
+ * \param[in] query                The query sequence.
+ * \param[in] matrix               The trace matrix.
+ * \param[in] end_coordinate       Where the trace in the matrix ends.
  */
 template <
     typename database_t,
     typename query_t,
     typename trace_matrix_t,
-    typename gapped_database_t = gapped<std::decay_t<decltype(std::declval<database_t>()[0])>>,
-    typename gapped_query_t = gapped<std::decay_t<decltype(std::declval<query_t>()[0])>>>
+    typename gapped_database_alphabet_t = gapped<std::decay_t<decltype(std::declval<database_t>()[0])>>,
+    typename gapped_query_alphabet_t = gapped<std::decay_t<decltype(std::declval<query_t>()[0])>>>
 //!\cond
     requires matrix_concept<remove_cvref_t<trace_matrix_t>> &&
              std::Same<typename remove_cvref_t<trace_matrix_t>::entry_type, trace_directions>
 //!\endcond
-inline std::pair<std::list<gapped_database_t>, std::list<gapped_query_t>>
-alignment_trace(database_t && database, query_t && query, trace_matrix_t && matrix, alignment_coordinate const end_coordinate)
+inline std::pair<std::vector<gapped_database_alphabet_t>, std::vector<gapped_query_alphabet_t>>
+alignment_trace(
+    database_t && database,
+    query_t && query,
+    trace_matrix_t && matrix,
+    alignment_coordinate const end_coordinate)
 {
+    using signed_size_t = std::make_signed_t<size_t>;
+
     constexpr auto N = trace_directions::none;
     constexpr auto D = trace_directions::diagonal;
     constexpr auto L = trace_directions::left;
     constexpr auto U = trace_directions::up;
-    size_t col = end_coordinate.seq1_pos + 1;
-    size_t row = end_coordinate.seq2_pos + 1;
+    signed_size_t col = end_coordinate.seq1_pos + 1;
+    signed_size_t row = end_coordinate.seq2_pos + 1;
 
-    assert(col <= database.size());
     assert(row <= query.size());
+    assert(col <= database.size());
+    assert(row < matrix.rows());
+    assert(col < matrix.cols());
 
-    std::list<gapped_database_t> gapped_database{};
-    std::list<gapped_query_t> gapped_query{};
+    std::deque<gapped_database_alphabet_t> gapped_database{};
+    std::deque<gapped_query_alphabet_t> gapped_query{};
 
     if (matrix.at(0, 0) != N)
-        throw std::runtime_error{"End trace must be NONE"};
+        throw std::logic_error{"End trace must be NONE"};
 
-    while(row >= 0 && col >= 0)
+    while(true)
     {
         trace_directions dir = matrix.at(row, col);
         if ((dir & L) == L)
         {
-            col = std::max<int>(col - 1, 0);
+            col = std::max<signed_size_t>(col - 1, 0);
             gapped_database.push_front(database[col]);
             gapped_query.push_front(gap::GAP);
         }
         else if ((dir & U) == U)
         {
-            row = std::max<int>(row - 1, 0);
+            row = std::max<signed_size_t>(row - 1, 0);
             gapped_database.push_front(gap::GAP);
             gapped_query.push_front(query[row]);
         }
         else if ((dir & D) == D)
         {
-            row = std::max<int>(row - 1, 0);
-            col = std::max<int>(col - 1, 0);
+            row = std::max<signed_size_t>(row - 1, 0);
+            col = std::max<signed_size_t>(col - 1, 0);
             gapped_database.push_front(database[col]);
             gapped_query.push_front(query[row]);
         }
@@ -163,10 +228,16 @@ alignment_trace(database_t && database, query_t && query, trace_matrix_t && matr
         {
             if (row == 0 || col == 0)
                 break;
-            throw std::runtime_error{"Trace not found"};
+            throw std::logic_error{"Trace not found"};
         }
+
     }
-    return {gapped_database, gapped_query};
+
+    return
+    {
+        {std::begin(gapped_database), std::end(gapped_database)},
+        {std::begin(gapped_query), std::end(gapped_query)}
+    };
 }
 
 /*!\brief A trace matrix represented in a one-dimensional std::vector
--- a/include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
+++ b/include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
@@ -145,17 +145,15 @@ inline alignment_coordinate alignment_begin_coordinate(trace_matrix_t && matrix,
         }
         else
         {
-            if (row == 0 || col == 0)
-                break;
-            throw std::logic_error{"Trace not found"};
+#ifndef NDEBUG
+            if (!(row == 0 || col == 0))
+                throw std::logic_error{"Unkown seqan3::trace_direction in an inner cell of the trace matrix."};
+#endif
+            break;
         }
     }
 
-    return
-    {
-        std::max<signed_size_t>(col - 1, 0),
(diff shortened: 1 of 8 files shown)


rrahn:
oh I fixed the alignment_selector already to use the alignment_coordinate.

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -131,40 +213,58 @@ alignment_trace(database_t && database, query_t && query, trace_matrix_t && matr
...
+    size_t i = gapped_seq_size - 1;
     while(row >= 0 && col >= 0)
     {
         trace_directions dir = matrix.at(row, col);
+
+        if (dir == N)
+        {
+            if (row == 0 || col == 0)
+                break;
+            throw std::runtime_error{"Trace not found"};
```
rrahn: logic_error? I mean the invariant of the trace matrix was not fulfilled.

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -118,9 +194,15 @@ template <
...
 //!\endcond
-inline std::pair<std::list<gapped_database_t>, std::list<gapped_query_t>>
-alignment_trace(database_t && database, query_t && query, trace_matrix_t && matrix, alignment_coordinate const end_coordinate)
+inline std::pair<std::vector<gapped_database_t>, std::vector<gapped_query_t>>
+alignment_trace(
+    database_t && database,
+    query_t && query,
+    trace_matrix_t && matrix,
+    alignment_coordinate const end_coordinate,
+    std::optional<size_t> gapped_sequence_size = std::nullopt)
```
rrahn: This seems to be over engineered. Why not just take a vector fill it reverse and finally reverse move it element wise into the result. That is exactly the same complexity and does not require the statistics stuff and pre knowledge of the size etc.

rrahn: It needs to be touched any way, when we have gap structures.

include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
```diff
@@ -426,27 +426,28 @@ struct pairwise_alignment_edit_distance_unbanded
...
-            // TODO: Compute the end coordinate
-            // TODO: Compute the begin coordinate
-            get<align_result_key::score>(res) = score();
+
+            if constexpr (std::tuple_size_v<result_type> >= 3)
+            {
+                alignment_coordinate end = end_coordinate();
+                get<align_result_key::end>(res) = end;
+
+                if constexpr (std::tuple_size_v<result_type> >= 4)
```
rrahn: don't nest them

marehr: If I don't nest it will be a lot of code duplication

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
marehr: Yeah lets do this later!

marehr: The thing is that you enforce to compute the `begin_coordinate` in align_result (i.e., traversing the trace matrix twice). Why not use the already know length of the `gapped_sequence`.

But, this has to be changed, because this only works for random access containers.

include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
rrahn: Why? If condition end is fulfilled we will write the end coordinate.
if condition begin is fulfilled we write begin coordinate.
If we need to store the state then we do it in the class. That is how the rest is done anyway.
if condition trace is fulfilled we write trace. 
This means, that the lower requirements are always included for the higher ones and the order gives us the right way to execute it without duplicating code

marehr: okay I don't nest it anymore

include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
```diff
@@ -430,20 +430,30 @@ struct pairwise_alignment_edit_distance_unbanded
         {
             get<align_result_key::score>(res) = score();
         }
-        if constexpr (std::tuple_size_v<result_type> >= 3)
+
+        if constexpr (std::tuple_size_v<result_type> == 3)
+        {
+            alignment_coordinate end = end_coordinate();
```
rrahn: why do you store it first in a temp?

include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
```diff
@@ -430,20 +430,30 @@ struct pairwise_alignment_edit_distance_unbanded
         {
             get<align_result_key::score>(res) = score();
         }
-        if constexpr (std::tuple_size_v<result_type> >= 3)
+
+        if constexpr (std::tuple_size_v<result_type> == 3)
```
rrahn: why not >=3?

include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
```diff
@@ -430,20 +430,30 @@ struct pairwise_alignment_edit_distance_unbanded
...
+        {
+            alignment_coordinate end = end_coordinate();
+            get<align_result_key::end>(res) = end;
+        }
+        else if constexpr (std::tuple_size_v<result_type> == 4)
         {
-            get<align_result_key::end>(res) = end_coordinate();
+            alignment_coordinate end = end_coordinate();
+            trace_matrix_type matrix = trace_matrix();
```
rrahn: is trace_matrix computing or getting the cached trace matrix?

include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
```diff
@@ -430,20 +430,30 @@ struct pairwise_alignment_edit_distance_unbanded
...
-            auto && [seq1_trace, seq2_trace] = alignment_trace(database, query, trace_matrix(), end_coordinate());
-            auto & [in1, in2] = get<align_result_key::trace>(res);
-            ranges::copy(seq1_trace, ranges::back_inserter(in1));
-            ranges::copy(seq2_trace, ranges::back_inserter(in2));
+            alignment_coordinate end = end_coordinate();
+            trace_matrix_type matrix = trace_matrix();
+            std::tuple<size_t, size_t, size_t> stats = alignment_trace_statistics(matrix, end);
+
+            get<align_result_key::end>(res) = end;
+            get<align_result_key::begin>(res) = alignment_begin_coordinate(end, stats);
```
rrahn: could be stored already in the contexts before

include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
marehr: computing

include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
marehr: true

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -98,29 +99,108 @@ struct alignment_coordinate
...
+    constexpr auto D = trace_directions::diagonal;
+    constexpr auto L = trace_directions::left;
+    constexpr auto U = trace_directions::up;
+    size_t col = end_coordinate.seq1_pos + 1;
+    size_t row = end_coordinate.seq2_pos + 1;
+    size_t inserts = 0;
+    size_t deletes = 0;
+    size_t substitutes = 0;
+
+    while(row >= 0 && col >= 0)
```
smehringer: `row` and `col` never drop below 0 but you break when the trace is not L/U/D and one is zero.. why this check then? you could just use `while(true)`

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -98,29 +99,108 @@ struct alignment_coordinate
...
+        }
+        else if ((dir & D) == D)
+        {
+            row = std::max<signed_size_t>(row - 1, 0);
+            col = std::max<signed_size_t>(col - 1, 0);
+            ++substitutes;
+        }
+        else
+        {
+            if (row == 0 || col == 0)
```
smehringer: Does this consider free end gaps ? (I suppose in free end gaps you have a trace != L/D/U in the respective column/row?)

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -98,29 +99,108 @@ struct alignment_coordinate
...
+                break;
+            throw std::logic_error{"Trace not found"};
+        }
+    }
+    return {inserts, deletes, substitutes};
+}
+
+/*!\brief Compute the begin coordinate.
+ * \ingroup alignment_matrix
+ * \tparam    trace_matrix_t The type of the trace matrix.
```
smehringer: copy and paste error?

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -98,29 +99,108 @@ struct alignment_coordinate
...
+ * \param[in] end_coordinate Where the trace in the matrix ends.
+ * \param[in] statistics     Statistics returned by alignment_trace_statistics.
+ */
+inline alignment_coordinate
+alignment_begin_coordinate(alignment_coordinate const end_coordinate, std::tuple<size_t, size_t, size_t> const & statistics)
```
smehringer: Is this a function only needed in combination with statistics? otherwise, you might want to add a function that takes the end coordinate and the matrix and computes the begin position

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -98,29 +99,108 @@ struct alignment_coordinate
     size_t seq2_pos;
 };
 
-/*!\brief Compute the trace from a trace matrix
+/*!\brief Compute the number of insertions, deletions and
+ * substitutions between begin coordinate and end coordinate.
```
smehringer: I think substitutes is the wrong name since you don't know if bases are substituted (they could be the same and just match). Maybe call it `aligned` or `aligned_bases` or something else since substitutes implies mismatches.

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -98,29 +99,108 @@ struct alignment_coordinate
...
  * \ingroup alignment_matrix
- * \tparam    database_t     The type of the database sequence.
- * \tparam    query_t        The type of the query sequence.
  * \tparam    trace_matrix_t The type of the trace matrix.
- * \param[in] database       The database sequence.
- * \param[in] query          The query sequence.
  * \param[in] matrix         The trace matrix.
  * \param[in] end_coordinate Where the trace in the matrix ends.
  */
```
smehringer: missing return statement!

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -98,29 +99,108 @@ struct alignment_coordinate
...
+    }
+    return {inserts, deletes, substitutes};
+}
+
+/*!\brief Compute the begin coordinate.
+ * \ingroup alignment_matrix
+ * \tparam    trace_matrix_t The type of the trace matrix.
+ * \param[in] end_coordinate Where the trace in the matrix ends.
+ * \param[in] statistics     Statistics returned by alignment_trace_statistics.
+ */
```
smehringer: missing return docu

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -98,29 +99,108 @@ struct alignment_coordinate
...
+/*!\brief Compute the trace from a trace matrix.
+ * \ingroup alignment_matrix
+ * \tparam    database_t           The type of the database sequence.
+ * \tparam    query_t              The type of the query sequence.
+ * \tparam    trace_matrix_t       The type of the trace matrix.
+ * \param[in] database             The database sequence.
+ * \param[in] query                The query sequence.
+ * \param[in] matrix               The trace matrix.
+ * \param[in] end_coordinate       Where the trace in the matrix ends.
+ */
```
smehringer: missing return

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -98,29 +99,108 @@ struct alignment_coordinate
...
+        static_cast<size_t>(std::max<signed_size_t>(end_coordinate.seq2_pos - deletes - substitutes, 0))
+    };
+}
+
+/*!\brief Compute the trace from a trace matrix.
+ * \ingroup alignment_matrix
+ * \tparam    database_t           The type of the database sequence.
+ * \tparam    query_t              The type of the query sequence.
+ * \tparam    trace_matrix_t       The type of the trace matrix.
+ * \param[in] database             The database sequence.
```
smehringer: missing \tparam for gapped `gapped_database_alphabet_t` and `gapped_query_alphabet_t`

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -98,29 +99,108 @@ struct alignment_coordinate
...
+    typename gapped_database_alphabet_t = gapped<std::decay_t<decltype(std::declval<database_t>()[0])>>,
+    typename gapped_query_alphabet_t = gapped<std::decay_t<decltype(std::declval<query_t>()[0])>>>
 //!\cond
     requires matrix_concept<remove_cvref_t<trace_matrix_t>> &&
              std::Same<typename remove_cvref_t<trace_matrix_t>::entry_type, trace_directions>
 //!\endcond
-inline std::pair<std::list<gapped_database_t>, std::list<gapped_query_t>>
-alignment_trace(database_t && database, query_t && query, trace_matrix_t && matrix, alignment_coordinate const end_coordinate)
+inline std::pair<std::vector<gapped_database_alphabet_t>, std::vector<gapped_query_alphabet_t>>
+alignment_trace(
```
smehringer: please move the database parameter into this line and indent the rest accordingly

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
marehr: Yeah it does. But, this will break if we have local alignments. Add this point we have to rewrite this.

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
marehr: changed this :)

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
marehr: those are non-public, I put them in between `\cond`

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -98,75 +99,148 @@ struct alignment_coordinate
...
  * \param[in] matrix         The trace matrix.
  * \param[in] end_coordinate Where the trace in the matrix ends.
  */
+ template <typename trace_matrix_t>
+ //!\cond
+     requires matrix_concept<remove_cvref_t<trace_matrix_t>> &&
+              std::Same<typename remove_cvref_t<trace_matrix_t>::entry_type, trace_directions>
+ //!\endcond
+inline alignment_coordinate
+alignment_begin_coordinate(trace_matrix_t && matrix, alignment_coordinate const end_coordinate)
```
h-2: this is is seqan3 stlye, instead 
```cpp
inline alignment_coordinate alignment_begin_coordinate(trace_matrix_t && matrix, 
                                                       alignment_coordinate const end_coordinate)
```

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -98,75 +99,148 @@ struct alignment_coordinate
...
+    constexpr auto D = trace_directions::diagonal;
+    constexpr auto L = trace_directions::left;
+    constexpr auto U = trace_directions::up;
+    signed_size_t row = end_coordinate.seq2_pos + 1;
+    signed_size_t col = end_coordinate.seq1_pos + 1;
+
+    assert(row < matrix.rows());
+    assert(col < matrix.cols());
+
+    while(true)
```
h-2: spacing

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -98,75 +99,148 @@ struct alignment_coordinate
...
+            if (row == 0 || col == 0)
+                break;
+            throw std::logic_error{"Trace not found"};
+        }
+    }
+    return
+    {
+        std::max<signed_size_t>(col - 1, 0),
+        std::max<signed_size_t>(row - 1, 0)
+    };
```
h-2: please introduce a blank line before the return statement and write the return statement into a single line, it looks really weird :see_no_evil:

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -98,75 +99,148 @@ struct alignment_coordinate
...
+ * \param[in] end_coordinate             Where the trace in the matrix ends.
+ */
 template <
     typename database_t,
     typename query_t,
     typename trace_matrix_t,
-    typename gapped_database_t = gapped<std::decay_t<decltype(std::declval<database_t>()[0])>>,
-    typename gapped_query_t = gapped<std::decay_t<decltype(std::declval<query_t>()[0])>>>
+    typename gapped_database_alphabet_t = gapped<std::decay_t<decltype(std::declval<database_t>()[0])>>,
+    typename gapped_query_alphabet_t = gapped<std::decay_t<decltype(std::declval<query_t>()[0])>>>
```
h-2: please use `reference_t<query_t>` instead of `decltype(std::declval<query_t>()[0])>>`

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -98,75 +99,148 @@ struct alignment_coordinate
...
         }
+
     }
-    return {gapped_database, gapped_query};
+
+    return
+    {
+        {std::begin(gapped_database), std::end(gapped_database)},
+        {std::begin(gapped_query), std::end(gapped_query)}
+    };
```
h-2: why this change?

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
marehr: What is with realllllllllllllllllllllllllllllllly long return statements?

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -98,75 +99,148 @@ struct alignment_coordinate
...
+            col = std::max<signed_size_t>(col - 1, 0);
             gapped_database.push_front(database[col]);
             gapped_query.push_front(query[row]);
         }
         else
         {
             if (row == 0 || col == 0)
                 break;
-            throw std::runtime_error{"Trace not found"};
+            throw std::logic_error{"Trace not found"};
```
smehringer: This is not a very expressive error message IMPOV. Maybe `Unkown seqan3::trace_direction in an inner cell of the trace matrix.`

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
marehr: yeah readability looks weird...

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
rrahn: decltype(auto) 🤣

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
rrahn: Also, please don't decay but use remove_cvref, or what was the rational for decay?

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -98,75 +99,148 @@ struct alignment_coordinate
...
-    typename gapped_query_t = gapped<std::decay_t<decltype(std::declval<query_t>()[0])>>>
+    typename gapped_database_alphabet_t = gapped<std::decay_t<decltype(std::declval<database_t>()[0])>>,
+    typename gapped_query_alphabet_t = gapped<std::decay_t<decltype(std::declval<query_t>()[0])>>>
 //!\cond
     requires matrix_concept<remove_cvref_t<trace_matrix_t>> &&
              std::Same<typename remove_cvref_t<trace_matrix_t>::entry_type, trace_directions>
 //!\endcond
-inline std::pair<std::list<gapped_database_t>, std::list<gapped_query_t>>
-alignment_trace(database_t && database, query_t && query, trace_matrix_t && matrix, alignment_coordinate const end_coordinate)
+inline std::pair<std::vector<gapped_database_alphabet_t>, std::vector<gapped_query_alphabet_t>>
```
rrahn: return a tuple?

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
```diff
@@ -98,75 +99,148 @@ struct alignment_coordinate
...
+ * \tparam    query_t                    The type of the query sequence.
+ * \tparam    trace_matrix_t             The type of the trace matrix.
+ * \cond DEV
+ * \tparam    gapped_database_alphabet_t The alphabet type of the gapped database sequence.
+ * \tparam    gapped_query_alphabet_t    The alphabet type of the gapped query sequence.
+ * \endcond
+ * \param[in] database                   The database sequence.
+ * \param[in] query                      The query sequence.
+ * \param[in] matrix                     The trace matrix.
+ * \param[in] end_coordinate             Where the trace in the matrix ends.
```
rrahn: return parameter.

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
rrahn: and this must also be on the same line, then???

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
rrahn: because the return and the local type does now differ.
So their is now a conversion from deque to vector.

include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
```diff
@@ -430,20 +430,23 @@ struct pairwise_alignment_edit_distance_unbanded
...
-        // }
-        if constexpr (std::tuple_size_v<result_type> == 5)
+
+        if constexpr (std::tuple_size_v<result_type> >= 5)
         {
-            auto && [seq1_trace, seq2_trace] = alignment_trace(database, query, trace_matrix(), end_coordinate());
-            auto & [in1, in2] = get<align_result_key::trace>(res);
-            ranges::copy(seq1_trace, ranges::back_inserter(in1));
-            ranges::copy(seq2_trace, ranges::back_inserter(in2));
+            get<align_result_key::trace>(res) = alignment_trace(database, query, matrix, end);
```
rrahn: you don't need to cache `end`, as it is already part of the res. You can just use res.end_coordinate() here.

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
marehr: > Also, please don't decay but use remove_cvref, or what was the rational for decay?

I just use `value_type`

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
marehr: tuple is not possible right now, because of #471. @smehringer has to fix the `aligned_sequence -> cigar` conversion function.

marehr:
someone else how want to do a review? I think I addressed all remarks.

include/seqan3/alignment/matrix/alignment_trace_matrix.hpp
h-2: Wouldn't it be faster (and easier to read) to push_back() on a vector, call reverse() in the end and move it back out?

marehr: I personally would not convert at all.

I doubt that this will be faster. And I'm not sure why it would be better to put something in reverse order just to reserve it again. From a data layout point it should be more expansive.

h-2: Well now you are allocating for a deque. A deck is non-contiguous so you will likely allocate mulitple dispersed blocks of memory. Then on return you iterate over these blocks jumping around through memory and then allocating another time for the vector and copying the values. Since deque is non-contiguous you can never use memcpy for this.

Calling reverse on a vector seems a lot faster because you have a contiguous block and you swap the element internally...

marehr: You know I had a vector and pushed it to the front, then @rrahn found it not cool. Then I changed it to a deque, because it has contiguous blocks, then @h-2 found it not cool. Whatever you say, I will not change this anymore for this pull request.

marehr: The thing is that this function will change probably anyway, once we use different gapped sequence implementation.

rrahn: First, there is nothing to argue about performance before measuring performance. So this is likely to change since, for affine gaps this will change anyway.

rrahn:
i think all main issues have been addressed.

PR #461, opened 2018-09-19
[FIX] align_result should give a std::pair

marehr:
@rrahn merge ASAP

### DIFF
--- a/include/seqan3/alignment/pairwise/alignment_selector.hpp
+++ b/include/seqan3/alignment/pairwise/alignment_selector.hpp
@@ -92,8 +92,8 @@ struct determine_result_type
                                               score_type,
                                               alignment_coordinate,
                                               alignment_coordinate,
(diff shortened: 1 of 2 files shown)


rrahn:
Why??? Actually the other parts should be tuples. 
The problem here is that tuple can be assigned from pair and tuple, but pair can only be piecewise constructe from tuple. Since the type is not part of the user interface, I wanted to keep it most generic.
So where did it cause problems?

h-2:
I agree with @rrahn here! Also some checks failed :flushed:

marehr:
failed checks was because of the `debug_stream` pr #460 

The reason for pair was:

https://github.com/seqan/seqan3/blob/8085063bee8e61c067f5e57ddbe77194d65912b6/include/seqan3/io/alignment_file/detail.hpp#L161-L165

This has to be changed.

PR #464, opened 2018-09-19
[FIX] fix short help printing.

cpockrandt:
now prints the name and hline instead of two empty lines

### DIFF
--- a/include/seqan3/argument_parser/detail/format_help.hpp
+++ b/include/seqan3/argument_parser/detail/format_help.hpp
@@ -680,6 +680,8 @@ public:
      */
     void parse(argument_parser_meta_data const & parser_meta)
     {
+        meta = parser_meta;
+
         print_header();
 
         if (!parser_meta.synopsis.empty())


include/seqan3/argument_parser/detail/format_help.hpp
```diff
@@ -680,6 +680,8 @@ class format_short_help : public format_help
      */
     void parse(argument_parser_meta_data const & parser_meta)
     {
+        meta = parser_meta;
+
```
h-2: why do you set this member and not directly print the argument?

cpockrandt: Yeah @smehringer ... why are you doing this? :smiling_imp:

marehr: Because I (@smehringer ) wanted to delegate to the `print_header` function for consistency. If we change something in the way we print the help page you would have to change it at two places.

PR #470, opened 2018-09-19
fix the debug stream

h-2:
Fixes printing of:
* 0-terminated strings
* io manipulators that are implemented as functions templates
* const ranges

### DIFF
--- a/include/seqan3/io/stream/debug_stream.hpp
+++ b/include/seqan3/io/stream/debug_stream.hpp
@@ -149,9 +149,16 @@ public:
      */
     //!\brief Forwards to the underlying stream object.
     template <typename t>
-    debug_stream_type & operator<<(t const & v)
+    debug_stream_type & operator<<(t && v)
     {
         *stream << v;
+        return *this;
+    }
+
+    //!\brief This overloads enables forwarding std::endl and other manipulators.
+    debug_stream_type & operator<<(std::ostream&(*fp)(std::ostream&))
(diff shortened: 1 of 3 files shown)


include/seqan3/io/stream/debug_stream.hpp
```diff
@@ -149,12 +149,19 @@ class debug_stream_type
...
     template <typename t>
-    debug_stream_type & operator<<(t const & v)
+    debug_stream_type & operator<<(t && v)
     {
         *stream << v;
         return *this;
     }
 
+    //!\brief This overloads enables forwarding std::endl and other manipulators.
+    debug_stream_type & operator<<(std::ostream&(*fp)(std::ostream&))
```
marehr: interesting syntax :laughing:

include/seqan3/io/stream/debug_stream.hpp
```diff
@@ -149,12 +149,19 @@ class debug_stream_type
      */
     //!\brief Forwards to the underlying stream object.
     template <typename t>
-    debug_stream_type & operator<<(t const & v)
+    debug_stream_type & operator<<(t && v)
```
rrahn: that I did already 😏

include/seqan3/io/stream/debug_stream.hpp
h-2: function pointer ;)

Issue #471, opened 2018-09-19
after-workshop feedback


marehr:
A list of encountered issues:
* [x] This should take a tuple:
https://github.com/seqan/seqan3/blob/8085063bee8e61c067f5e57ddbe77194d65912b6/include/seqan3/io/alignment_file/detail.hpp#L161-L165
* [x]  Revert #461 to use tuple again
* [x] `get` has a nameclash for `sequence_file_input`
    ```c++
    using namespace std;
    using namespace seqan3;
    get<fields::ID>(seq_file)
    ```
* [x] searching a `dna5q` query in a `dna5` database does not work
* [x] #465 index: `mode(true ? best : strata{4})` does not work (accepts no runtime parameter).
* [x] general thoughts:
     * allow `sequences_of_pair | align(cfg)` and
     * allow `queries | search(index, cfg) `
     * switch argument order in search; instead of `search(index, queries, cfg)` let it be `search(queries, index, cfg)`, because `index` and `cfg` are kinda fixed.

eseiler:
I think, I could write up a tutorial on how to implement a read mapper (like the Seqan2 tutorial style) - we have the code now, so we might as well use it as an example project.
But I would need to know where to put it (preferably we want to have CI on it, just like with the snippets) and what to use for writing the tutorial.

marehr:
We could setup the repo at `seqan/seqan3_simple_read_mapper` as an external project.

We can add a .travis.yml file and let Travis check daily whether the build still works. I think we even get a notification when the build fails.

In general I would add a branch for each single tutorial step. While the master only has the CmakeLists.txt and the reference and query files.

This approach has the advantage that people can catchup with the rest.

joergi-w:
I really like the idea of having the read mapper as a project. For this, I see two options: 
1. We really keep it small as a demo project and therefore call it seqan3_demo_read_mapper.
2. We consider it as our first (official) seqan3 app and can still use it for demonstration on how to write seqan3 applications. It's good to have a kind of app template at hand. 
In any case, I would not use the term simple.

h-2:
I have removed all points that were marked as fixed.

> searching a dna5q query in a dna5 database does not work

would work in some cases if #898 is adapted.

> switch argument order in search; instead of search(index, queries, cfg) let it be search(queries, index, cfg), because index and cfg are kinda fixed.

I agree, I also propose this in #898.

PR #475, opened 2018-09-20
[SNIPPETS] Fix code example to add newline after printing the exception message.

smehringer:
One fix for after user meeting feedback #471


smehringer:
But it is printing the exception message? Performance is not the issue anymore if an exception is thrown.

smehringer:
@rrahn I updated the newline printing.

PR #476, opened 2018-09-20
[INTERNAL] argument_parser now accepts also non-const char **.

### LINKED ISSUE #388: argument_parser requires const char ** argv
The argument_parser initialization requires a const char ** argv option, instead of just a char ** argv.

https://github.com/seqan/seqan3/blob/9b81a6c8403c5b79d4a6163dc6256c22f293ca48/include/seqan3/argument_parser/argument_parser.hpp#L212-L216

smehringer:
It is indirectly tested in the snippets that now mirror the "usual" main function parameters.

Fixes #388 and can be checked off in #471 after merging.

### DIFF
--- a/include/seqan3/argument_parser/argument_parser.hpp
+++ b/include/seqan3/argument_parser/argument_parser.hpp
@@ -166,7 +166,7 @@ public:
      * \param[in] argc     The number of command line arguments.
      * \param[in] argv     The command line arguments to parse.
      */
-    argument_parser(std::string const app_name, int const argc, const char ** argv)
+    argument_parser(std::string const app_name, int const argc, char const * const * const  argv)
     {
         info.app_name = std::move(app_name);
         init(argc, argv);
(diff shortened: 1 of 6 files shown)


include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -166,7 +166,7 @@ class argument_parser
      * \param[in] argc     The number of command line arguments.
      * \param[in] argv     The command line arguments to parse.
      */
-    argument_parser(std::string const app_name, int const argc, const char ** argv)
+    argument_parser(std::string const app_name, int const argc, char const * const * argv)
```
rrahn: `char const * const * const argv`
This would also cover const arrays of const arrays of const char, I think.

h-2:
Perhaps it would be wise to actually check the different combinations of `const` input? :nail_care:

PR #480, opened 2018-09-21
[FIX] various fixes for alignment_file

### LINKED ISSUE #457: alignment_file issues
Some of these will make the tutorial harder so it would be nice if you could look at them asap:

  * `alignment_file_output` and `alignment_file_output_options` are not part of the `alignment_file` module in the documentation
  * compare_aligned_values and get_cigar_string are not in detail as planned; constraints for get_cigar_string are not guarded
  * write_range constraints are not guarded
  * sam_format and alignment_file_output protected sections are not marked as private although it is detail

smehringer:
Fixes #457

### DIFF
--- a/include/seqan3/io/alignment_file/detail.hpp
+++ b/include/seqan3/io/alignment_file/detail.hpp
@@ -43,6 +43,7 @@
 #include <sstream>
 
 #include <seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp>
+#include <seqan3/core/concept/tuple.hpp>
 #include <seqan3/std/concepts>
 #include <seqan3/std/ranges>
 
@@ -50,10 +51,8 @@
 
 namespace seqan3::detail
 {
-
 /*!\brief Compares two aligned sequence values and returns their CIGAR operation.
  * \ingroup alignment_file
- * \relatesalso seqan3::aligned_sequence_concept
  * \tparam reference_char_type Must be equality comparable to seqan3::gap.
  * \tparam query_char_type     Must be equality comparable to seqan3::gap.
  * \param  reference_char      The aligned character of the reference to compare.
@@ -93,6 +92,7 @@ namespace seqan3::detail
  * in the extended CIGAR alphabet (\p extended_cigar = `true`) the function
  * will return an 'X' since the bases are aligned but are not
  * equal.
+ * \sa seqan3::aligned_sequence_concept
  */
 template<typename reference_char_type, typename query_char_type>
 //!\cond
@@ -119,26 +119,19 @@ char compare_aligned_values(reference_char_type const reference_char,
 /*!\brief Transforms an alignment represented by two aligned sequences into the
  *        corresponding CIGAR string.
  * \ingroup alignment_file
- * \relatesalso seqan3::aligned_sequence_concept
- * \tparam ref_seq_type    Must model seqan3::aligned_sequence_concept.
- * \tparam query_seq_type  Must model seqan3::aligned_sequence_concept.
- * \param  alignment       The alignment, represented by a pair of aligned sequences,
- *                         to be transformed into CIGAR_vector based on the
- *                         second (query) sequence.
+ *
+ * \tparam ref_seq_type    Must model std::ranges::ForwardRange. The value_type must
+ *                         be equality comparable to seqan3::gap.
+ * \tparam query_seq_type  Must model std::ranges::ForwardRange. The value_type must
+ *                         be equality comparable to seqan3::gap.
+ * \param  ref_seq         The reference sequence to compare against the query sequence.
+ * \param  query_seq       The query sequence to build the CIGAR string for.
  * \param  query_start_pos The start position of the alignment in the query
  *                         sequence indicating soft-clipping.
  * \param  query_end_pos   The end position of the alignment in the query
  *                         sequence indicating soft-clipping.
  * \param  extended_cigar  Whether to print the extended cigar alphabet or not. See CIGAR operation.
- * \returns An std::string representing the alignment.
- *
- * \details
- *
- * \attention Note that CIGAR elements (respectively by their
- *            CIGAR operation) are always related to on one of the two sequences
- *            in a pairwise alignment. In this case, the resulting cigar_vector
- *            is based on sequence at the second position of the \p alignment pair,
- *            namely the query sequence.
+ * \returns An std::string representing the alignment as a CIGAR string.
  *
  * ### Theoretical Example:
  *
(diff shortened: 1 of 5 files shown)


include/seqan3/io/alignment_file/detail.hpp
```diff
@@ -153,12 +157,15 @@ char compare_aligned_values(reference_char_type const reference_char,
...
-             std::EqualityComparableWith<gap, value_type_t<query_seq_type>>
-//\!endcond
-std::string get_cigar_string(std::pair<ref_seq_type, query_seq_type> const & alignment,
+template<tuple_like_concept alignment_type>
+//!\cond
+    requires requires {{std::tuple_size_v<alignment_type> == 2}} &&
+             std::ranges::ForwardRange<remove_cvref_t<decltype(get<0>(alignment_type{}))>> &&
+             std::ranges::ForwardRange<remove_cvref_t<decltype(get<1>(alignment_type{}))>> &&
+             std::EqualityComparableWith<gap, value_type_t<remove_cvref_t<decltype(get<0>(alignment_type{}))>>> &&
+             std::EqualityComparableWith<gap, value_type_t<remove_cvref_t<decltype(get<1>(alignment_type{}))>>>
```
h-2: 1) why do we need remove_cvref_t here?
2) use `std::tuple_element_t<>` instead of `decltype(get<`
3) use `alignment_type && alignment` instead of `const &`

But in general I think this whole signature could be greatly simplified by just taking two range arguments as `&&`. That would make the concept checks much easier to write (and to explain in the documentation) and also because they are named it would be immediately clear which one is the reference and which one the query.

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -381,7 +381,9 @@ class alignment_file_format_sam
      * \param[in]     field_value The value to print.
      */
     template <typename stream_it_t, typename field_type>
-        requires std::ranges::ForwardRange<field_type>
+    //!\cond
+        requires std::ranges::ForwardRange<remove_cvref_t<field_type>>
+    //!\endcond
```
h-2: why is this change necessary? I would even argue that it is dangerous, because not all ranges are const-iterable. If you pass a range that isn't it will still be accepted by this check, but than fail later on.

include/seqan3/io/alignment_file/detail.hpp
```diff
@@ -43,13 +43,15 @@
...
 #include <seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp>
+#include <seqan3/core/concept/tuple.hpp>
 #include <seqan3/std/concepts>
 #include <seqan3/std/ranges>
 
 #include <range/v3/view/zip.hpp>
 
 namespace seqan3::detail
 {
+//!\privatesection
```
h-2: I don't think this makes a difference? How can non-members be private?

smehringer: You need to discuss this with doxygen. it solves the issue for me

include/seqan3/io/alignment_file/detail.hpp
smehringer: 1) I thought that's what you meant in your comment "get_string constraints are not guarded". And I thought value_type does not always work on references.

2) why removing `const`?

general) well, but we decided to treat "an alignment" as a pair of gapped sequences. so I thought we at least make the related functions work on "an alignment".

h-2: > I thought that's what you meant in your comment "get_string constraints are not guarded"

no, I meant that `//!\cond` is missing around it :laughing: 

> And I thought value_type does not always work on references.

It should definitely work on ranges, otherwise that's a bug.

> why removing const?

you mean why remove `remove_cvref_t`? See my answer below, I think it's not only unnecessary, but actually wrong.

In general, please don't just `remove_cvref_t` everywhere that you are not sure, but actually try to understand what's going on :wink: In most cases you won't need it and in many others `remove_const_t` or `remove_reference_t` will be sufficient.

> general) well, but we decided to treat "an alignment" as a pair of gapped sequences. so I thought we at least make the related functions work on "an alignment".

we agreed internally that we wouldn't make an extra data type but that doesn't mean we have to use pairs or tuples everywhere. I think that depends on the situation, e.g if you compare two different "alignments", then I would take two tuples so that each row is clearly grouped into one of the two, but in this case the interface would just become simpler by individual args IMHO.
Especially since both rows of the alignment are not treated equally by this algorithm!

include/seqan3/io/alignment_file/detail.hpp
h-2: touché :rofl:

include/seqan3/io/alignment_file/format_sam.hpp
rrahn: Does ForwardRange check cbegin/cend?

include/seqan3/io/alignment_file/detail.hpp
smehringer: > you mean why remove remove_cvref_t?

No sorry that was meant for point 3) why passing `alignment_type && alignment` and not `alignment_type const && alignment`

h-2: because the tuple could contain ranges that are not const-iterable and if the tuple is const the elements will appear const, too, and it will break.

It's best to test all these functions that take "some range" with a range that is not const-iterable, like `view::drop`.

h-2: (but this would also disappear if the tuple disappears!

include/seqan3/io/alignment_file/format_sam.hpp
h-2: no, why should it?

h-2: free function cbegin is the same as free function begin with const & argument

include/seqan3/io/alignment_file/detail.hpp
rrahn: Just to add to it. The tuple interface makes the typical use case working with seqan alignment computation and much more easier, as it does not involve untying tuple before passing it's elements into the function and fairly I was hoping to have this interface at the end:
```cpp
debug_stream << aling_pairwise(...) | transform([](auto && res){ return res.trace(); }) | to_cigar << '\n';
```

include/seqan3/io/alignment_file/detail.hpp
```diff
@@ -207,4 +202,57 @@ std::string get_cigar_string(std::pair<ref_seq_type, query_seq_type> const & ali
...
+ * the following cigar string when printed: "4M2I5M2D1M". The extended cigar
+ * string would look like this: "3=1X2I3=1X1=2D1=".
+ * \sa seqan3::aligned_sequence_concept
+ */
+template<tuple_like_concept alignment_type>
+//!\cond
+    requires std::tuple_size_v<remove_cvref_t<alignment_type>> == 2
+//!\endcond
+std::string get_cigar_string(alignment_type && alignment,
+                             uint32_t query_start_pos = 0,
```
rrahn: const

include/seqan3/io/alignment_file/format_sam.hpp
rrahn: just need vacation 🍹

PR #486, opened 2018-09-21
constexpr_invocable check and constexpr alphabets

h-2:
All our alphabets are `constexpr` safe and some functionality (e.g. in the compositions) depends on this. But we require it nowhere (and it shouldnt be required in general).

Unfortunately there is no concept checking syntax for checking the constexpr'ness of a function, but I found a great hack on how to do it.

I have subsequently added constexpr concepts for alphabet and semi_alphabet and changed the requirements for the compositions. Currently these concepts reside in detail.

### DIFF
--- a/include/seqan3/alphabet/quality/phred42.hpp
+++ b/include/seqan3/alphabet/quality/phred42.hpp
@@ -173,9 +173,8 @@ struct phred42
      *
      * Constant.
      */
-    constexpr phred42 & assign_char(char_type const c)
+    constexpr phred42 & assign_char(char_type const c) noexcept
     {
-        assert(c >= offset_char && c < offset_char + max_value_size);
         _value = char_to_value[c];
         return *this;
     }
@@ -190,7 +189,7 @@ struct phred42
      *
      * Constant.
      */
-    constexpr phred42 & assign_phred(phred_type const p)
+    constexpr phred42 & assign_phred(phred_type const p) noexcept
     {
         // p >= 0 is implicitly always true
         assert(p < max_value_size);
@@ -213,7 +212,7 @@ struct phred42
      *
      * Guaranteed not to throw.
      */
-    constexpr phred42 & assign_rank(phred_type const p)
+    constexpr phred42 & assign_rank(phred_type const p) noexcept
     {
         return assign_phred(p);
     }
@@ -223,7 +222,7 @@ struct phred42
     //!\tparam other_qual_type The type to convert to; must satisfy seqan3::quality_concept.
     //
     template <quality_concept other_qual_type>
-    explicit constexpr operator other_qual_type() const //noexcept
+    explicit constexpr operator other_qual_type() const noexcept
     {
         return detail::convert_through_phred_representation<other_qual_type, std::decay_t<decltype(*this)>>[to_phred()];
     }
--- a/include/seqan3/alphabet/quality/phred42.hpp
+++ b/include/seqan3/alphabet/quality/phred42.hpp
@@ -54,6 +54,7 @@ namespace seqan3
 
 /*!\brief Quality type for traditional Sanger and modern Illumina Phred scores (typical range).
  * \implements seqan3::quality_concept
+ * \implements seqan3::detail::constexpr_alphabet_concept
  * \ingroup quality
  *
  * \details
--- a/include/seqan3/alphabet/quality/phred63.hpp
+++ b/include/seqan3/alphabet/quality/phred63.hpp
@@ -54,6 +54,7 @@ namespace seqan3
 
 /*!\brief Quality type for traditional Sanger and modern Illumina Phred scores (full range).
  * \implements seqan3::quality_concept
+ * \implements seqan3::detail::constexpr_alphabet_concept
  * \ingroup quality
  *
  * \details
--- a/include/seqan3/alphabet/quality/phred63.hpp
+++ b/include/seqan3/alphabet/quality/phred63.hpp
@@ -168,9 +168,8 @@ struct phred63
       *
       * Constant.
       */
-    constexpr phred63 & assign_char(char_type const c)
+    constexpr phred63 & assign_char(char_type const c) noexcept
     {
-        assert(c >= offset_char && c < offset_char + value_size);
         _value = char_to_value[c];
         return *this;
     }
@@ -185,7 +184,7 @@ struct phred63
      *
      * Constant.
      */
-    constexpr phred63 & assign_phred(phred_type const p)
+    constexpr phred63 & assign_phred(phred_type const p) noexcept
     {
         // p >= 0 is implicitly always true
         assert(p < value_size);
@@ -207,7 +206,7 @@ struct phred63
      *
      * Guaranteed not to throw.
      */
-    constexpr phred63 & assign_rank(phred_type const p)
+    constexpr phred63 & assign_rank(phred_type const p) noexcept
     {
         return assign_phred(p);
     }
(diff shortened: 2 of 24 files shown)


marehr:
```
/home/travis/build/seqan/seqan3/test/unit/alphabet/composition/composition_test.cpp: In member function ‘virtual void composition_custom_constructors_Test::TestBody()’:
/home/travis/build/seqan/seqan3/test/unit/alphabet/composition/composition_test.cpp:76:36: error: template constraint failure
     gapped<union_composition<dna4, phred42>> t61{dna4::C};
                                    ^~~~~~~
/home/travis/build/seqan/seqan3/test/unit/alphabet/composition/composition_test.cpp:76:36: note:   constraints not satisfied
/home/travis/build/seqan/seqan3/test/unit/alphabet/composition/composition_test.cpp:76:36: note: ‘(constexpr_alphabet_concept<alternative_types>&& ... && sizeof ... (alternative_types ...) >= 2)’ evaluated to false
/home/travis/build/seqan/seqan3/test/unit/alphabet/composition/composition_test.cpp:76:43: error: template argument 1 is invalid
     gapped<union_composition<dna4, phred42>> t61{dna4::C};
                                           ^~
```
on travis and on jenkins

https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/pipeline_unit_tests_unix/837/CEREAL=ON,label=ubuntu_1804/consoleFull

marehr:
It seems that `phred42` is not constexpr

h-2:
@marehr This is some compiler weirdness with the assertions in the qualitites' `assign_char`. However `assign_char` is not supposed to assert anyways. It's supposed to just convert silently so I have removed the assertions and now the test passes.

include/seqan3/core/metafunction/function.hpp
```diff
@@ -0,0 +1,70 @@
...
+// ----------------------------------------------------------------------------
+// is_constexpr
+// ----------------------------------------------------------------------------
+
+namespace seqan3::detail
+{
+
+//!\brief NO-OP helper for is_constexpr
+//!\ingroup metafunction
+constexpr auto is_constexpr_helper(int) {}
```
rrahn: is the return type evaluated in any case different then void?

include/seqan3/core/metafunction/function.hpp
```diff
@@ -0,0 +1,70 @@
...
+{
+
+/*!\brief Returns true if the expression passed to this macro can be evaluated at compile time, false otherwise.
+ * \ingroup metafunction
+ * \returns true or false.
+ */
+#if defined(__clang__) // bug: https://bugs.llvm.org//show_bug.cgi?id=15481
+#define is_constexpr(...) __builtin_constant_p( (__VA_ARGS__,0) )
+#else
+#define is_constexpr(...) noexcept(seqan3::detail::is_constexpr_helper( (__VA_ARGS__, 0) ))
```
rrahn: Mhmm, https://en.cppreference.com/w/cpp/language/constexpr suggests, that this is possible util c++17?
And the same states documentation for the https://en.cppreference.com/w/cpp/language/noexcept operator.
So as far as I understand, you simply check if any of the expressions added to __VA_ARGS__ can be evaluated at compile time and if so the noexcept returns true. But now I am irritated by the limitations of cppref, also because there wasn't any `since` clause what to expect since c++17.
Maybe you can elaborate?

h-2:
> And honestly, can you give an example, of where it is required or makes it easier for us to require that certain functions are in fact constexpr?

The current implementation of the compositions **requires** this already, it just never checked for it (I changed the check, see the diff). And the more diverse the alphabets get, it would probably be good to have a specialisation of the composition that works on non-constexpr alphabets, too (albeit more slowly).

This is something I discussed with @marehr before and I am also working on a type-erased alphabet where I would need to make this distinction.

rrahn:
I mean yes, it needs them to be constexpr, but it would cause a compiler error, that some functions cannot be used in a constexpr context I guess. So this is to make the error more readable for combining the alphabets. But as I said, I am not firm yet with the properties of noexcept as cppreference says something regarding it's behaviour until c++17. But it is not clear to me, if that means, that the behaviour remains valid or if we have to expect some changes in the functionality with respect to this.

h-2:
> I mean yes, it needs them to be constexpr, but it would cause a compiler error, that some functions cannot be used in a constexpr context I guess. So this is to make the error more readable for combining the alphabets

No, this is a prerequisite to making a specialisation that doesn't produce an error!

> But as I said, I am not firm yet with the properties of noexcept as cppreference says something regarding it's behaviour until c++17. But it is not clear to me, if that means, that the behaviour remains valid or if we have to expect some changes in the functionality with respect to this.

And as I said, it does say something about C++17:
poexc1
This links to:
https://en.cppreference.com/w/cpp/language/except_spec
which states that
"Each expression e has a set of potential exceptions, which is the union of the sets of potential exceptions of all immediate subexpressions of e [...]" which corresponds very closely to the old definition from my POV. The call to the helper is only noexcept if the sub-expression is constexpr so everything should be fine, I think.

h-2:
I have changed the return type of the helper to void and renamed the macro to be ugly (but consistent with other macros) since we cannot namespace it.

rrahn:
> No, this is a prerequisite to making a specialisation that doesn't produce an error!

That's right. 👍 

>"Each expression e has a set of potential exceptions, which is the union of the sets of potential exceptions of all immediate subexpressions of e [...]" which corresponds very closely to the old definition from my POV. The call to the helper is only noexcept if the sub-expression is constexpr so everything should be fine, I think.

Well, I read this and I think it should all work as it doesn't really make sense otherwise. It would be just helpful if somewhere would be written that the set of potential exceptions for core constant expressions is empty.

test/unit/core/metafunction/function_test.cpp
```diff
@@ -0,0 +1,105 @@
...
+    void get_i(int)
+    {}
+};
+
+TEST(metafunction, is_constexpr_invocable)
+{
+    int i = 32;
+    int constexpr j = 42;
+
+    EXPECT_TRUE((is_constexpr(constexpr_nonvoid_free_fun(3))));
```
marehr: rename?

include/seqan3/alphabet/quality/phred42.hpp
```diff
@@ -172,9 +173,8 @@ struct phred42
      *
      * Constant.
      */
-    constexpr phred42 & assign_char(char_type const c)
+    constexpr phred42 & assign_char(char_type const c) noexcept
     {
-        assert(c >= offset_char && c < offset_char + max_value_size);
```
marehr: Can we add this again? This should not have any conflict with noexcept, since it does not throw but terminate.

include/seqan3/alphabet/quality/phred63.hpp
```diff
@@ -167,9 +168,8 @@ struct phred63
       *
       * Constant.
       */
-    constexpr phred63 & assign_char(char_type const c)
+    constexpr phred63 & assign_char(char_type const c) noexcept
     {
-        assert(c >= offset_char && c < offset_char + value_size);
```
marehr: Here to

include/seqan3/alphabet/quality/phred68legacy.hpp
```diff
@@ -171,9 +172,8 @@ struct phred68legacy
       *
       * Constant.
       */
-    constexpr phred68legacy & assign_char(char_type const c)
+    constexpr phred68legacy & assign_char(char_type const c) noexcept
     {
-        assert(c >= offset_char && c < offset_char + value_size);
```
marehr: and here

PR #492, opened 2018-09-28
[INTERNAL] Remove the output_iterator_conversion_adaptor.

smehringer:
The conversion now happens directly in the formats by adding the `view::to_char` to the stream pipe.


h-2:
Hm, if we don't use the `ignore_output_iterator` anymore we can also think about removing it, but we can also discuss that at a later point in time.

In general I would use `ranges::back_insert_iterator{id}` instead of `ranges::back_inserter(id)`. The latter is just a convenience function for when there was no template argument type deduction, and the former states more clearly that a new object is being created.

smehringer:
> Hm, if we don't use the ignore_output_iterator anymore we can also think about removing it, but we can also discuss that at a later point in time.

I agree. Especially since `detail::consume()` already supplies a similar functionality and can often be used instead.

PR #496, opened 2018-10-05
improve detection of concept flag for clang

marehr:
This PR allows to use the concept version of clang++ directly with our cmake script.


h-2:
Hm, the Xclang stuff is a relic from the super-old clang implementation. Even on the experimental branch this shouldn't be necessary soon. For released clang it will definitely not be necessary so I am reluctant about adding this.

marehr:
Okay I will close this then.

Issue #239, opened 2018-04-25
create `std` module


h-2:
We will create a module called `std` that contains all wrapper and aliases of std and expected std functionality, i.e. implementations of standard library concepts and type_traits, as well as aliases for range-v3 functionality that is expected to be merged into the standard.

*Reason:* this signals clearly to developers that something called `foo` and defined in that module is identical to such `foo` in the standard library. It is also an indicator of "all this stuff is not biological, just infrastructure" and it also makes it easier to identify and remove the code in the distant future.

Checklist

  - [x] create module and at least basic documentation
  - [x] move most of `core/concept` there
  - [ ] update concepts to be interfaces, add interface documentation
    - [x] most things, but not all
  - [ ] move most of `core/metafunction` to `std/type_traits`
  - [ ] add aliases for `<algorithm>` from range-v3
  - [ ] add aliases for free frunctions for ranges from range-v3 (begin(), end(), size(), ...)
  - [ ] add aliases for view from range-v3

In general I would propose:
  * All usage of `ranges::` content must go through aliases in the std module.
  * Any of the things in std that will likely not land in c++20 should be reimplemented in our library (and moved out of std again)
  * this way we can ensure that SeqAn3 in combination with C++20 works without the ranges library.

marehr:
Do you mean by 

> move most of `core/metafunction` to `std/type_traits`

the same as in #243?

h-2:
This is the link to the current Ranges proposal: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0789r3.pdf

You can look up there whether something will (likely) be included or not.

h-2:
@smehringer @joshuak94 

I will try to explain this issue in more detail:

Currently we rely heavily on the range-v3 library for providing lots of functionality. Most of the functionality will be part of C++20, but quite a few things will not, and some of the things will be, but under different names. To avoid code-duplication and similar-but-different function names et cetera we decided to rely **only** on those parts of range-v3 that will be part of C++20. Furthermore we would like to only have to depend on the range-v3 library if we cannot actually use C++20. This also means we need to find out which things won't be part of C++20 and implement those ourselves anyway (I call this `CATEGORY0` for now).

[...]

PR #498, opened 2018-10-08
[INTERNAL] Update ::ranges::begin, end, size, copy, and empty to std::ranges::...

joshuak94:
More to come.

### DIFF
--- a/include/seqan3/io/structure_file/format_vienna.hpp
+++ b/include/seqan3/io/structure_file/format_vienna.hpp
@@ -61,6 +61,7 @@
 #include <seqan3/io/structure_file/detail.hpp>
 #include <seqan3/io/structure_file/input_options.hpp>
 #include <seqan3/io/structure_file/output_options.hpp>
+#include <seqan3/range/shortcuts.hpp>
 #include <seqan3/range/detail/misc.hpp>
 #include <seqan3/range/view/char_to.hpp>
 #include <seqan3/range/view/to_char.hpp>
@@ -161,24 +162,24 @@ public:
 
         // READ ID (if present)
         auto constexpr is_id = is_char<'>'>;
-        if (is_id(*ranges::begin(stream_view)))
+        if (is_id(*begin(stream_view)))
         {
             if constexpr (!detail::decays_to_ignore_v<id_type>)
             {
                 if (options.truncate_ids)
                 {
-                    ranges::copy(stream_view | ranges::view::drop_while(is_id || is_blank) // skip leading >
-                                             | view::take_until_or_throw(is_cntrl || is_blank)
-                                             | view::char_to<value_type_t<id_type>>,
-                                 std::back_inserter(id));
+                    std::ranges::copy(stream_view | ranges::view::drop_while(is_id || is_blank) // skip leading >
+                                                  | view::take_until_or_throw(is_cntrl || is_blank)
+                                                  | view::char_to<value_type_t<id_type>>,
+                                      std::back_inserter(id));
                     detail::consume(stream_view | view::take_line_or_throw);
                 }
                 else
                 {
-                    ranges::copy(stream_view | ranges::view::drop_while(is_id || is_blank) // skip leading >
-                                             | view::take_line_or_throw
-                                             | view::char_to<value_type_t<id_type>>,
-                                 std::back_inserter(id));
+                    std::ranges::copy(stream_view | ranges::view::drop_while(is_id || is_blank) // skip leading >
+                                                  | view::take_line_or_throw
+                                                  | view::char_to<value_type_t<id_type>>,
+                                      std::back_inserter(id));
                 }
             }
             else
@@ -189,11 +190,11 @@ public:
         else if constexpr (!detail::decays_to_ignore_v<id_type>)
         {
             auto constexpr is_legal_seq = is_in_alphabet<seq_legal_alph_type>;
-            if (!is_legal_seq(*ranges::begin(stream_view))) // if neither id nor seq found: throw
+            if (!is_legal_seq(*begin(stream_view))) // if neither id nor seq found: throw
             {
                 throw parse_error{std::string{"Expected to be on beginning of ID or sequence, but "} +
                                   is_id.msg.string() + " and " + is_legal_seq.msg.string() +
-                                  " evaluated to false on " + detail::make_printable(*ranges::begin(stream_view))};
+                                  " evaluated to false on " + detail::make_printable(*begin(stream_view))};
             }
         }
 
@@ -201,21 +202,21 @@ public:
         if constexpr (!detail::decays_to_ignore_v<seq_type>)
         {
             auto constexpr is_legal_seq = is_in_alphabet<seq_legal_alph_type>;
-            ranges::copy(stream_view | view::take_line_or_throw                      // until end of line
-                                     | ranges::view::remove_if(is_space || is_digit) // ignore whitespace and numbers
-                                     | ranges::view::transform([is_legal_seq](char const c)
-                                       {
-                                           if (!is_legal_seq(c))                    // enforce legal alphabet
-                                           {
-                                               throw parse_error{std::string{"Encountered an unexpected letter: "} +
-                                                                 is_legal_seq.msg.string() +
-                                                                 " evaluated to false on " +
-                                                                 detail::make_printable(c)};
-                                           }
-                                         return c;
-                                       })
-                                     | view::char_to<value_type_t<seq_type>>, // convert to actual target alphabet
-                         std::back_inserter(seq));
+            std::ranges::copy(stream_view | view::take_line_or_throw                      // until end of line
+                                          | ranges::view::remove_if(is_space || is_digit) // ignore whitespace and numbers
+                                          | ranges::view::transform([is_legal_seq](char const c)
+                                            {
+                                                if (!is_legal_seq(c))                    // enforce legal alphabet
+                                                {
+                                                    throw parse_error{std::string{"Encountered an unexpected letter: "} +
+                                                                      is_legal_seq.msg.string() +
+                                                                      " evaluated to false on " +
+                                                                      detail::make_printable(c)};
+                                                }
+                                              return c;
+                                            })
+                                          | view::char_to<value_type_t<seq_type>>, // convert to actual target alphabet
+                              std::back_inserter(seq));
         }
         else
         {
@@ -229,7 +230,7 @@ public:
             {
                 assert(std::addressof(seq) == std::addressof(structure));
                 using alph_type = typename value_type_t<structure_type>::structure_alphabet_type;
-                ranges::copy(read_structure<alph_type>(stream_view), ranges::begin(structure));
+                std::ranges::copy(read_structure<alph_type>(stream_view), begin(structure));
 
                 if constexpr (!detail::decays_to_ignore_v<bpp_type>)
                     detail::bpp_from_rna_structure<alph_type>(bpp, structure);
@@ -237,13 +238,13 @@ public:
             else
             {
                 using alph_type = value_type_t<structure_type>;
-                ranges::copy(read_structure<alph_type>(stream_view), std::back_inserter(structure));
+                std::ranges::copy(read_structure<alph_type>(stream_view), std::back_inserter(structure));
 
                 if constexpr (!detail::decays_to_ignore_v<bpp_type>)
                     detail::bpp_from_rna_structure<alph_type>(bpp, structure);
             }
             if constexpr (!detail::decays_to_ignore_v<seq_type>)
-                if (ranges::size(seq) != ranges::size(structure))
+                if (size(seq) != size(structure))
                     throw parse_error{"Found sequence and associated structure of different length."};
         }
         else if constexpr (!detail::decays_to_ignore_v<bpp_type>)
@@ -251,7 +252,7 @@ public:
             detail::bpp_from_rna_structure<wuss51>(bpp, read_structure<wuss51>(stream_view));
 
             if constexpr (!detail::decays_to_ignore_v<seq_type>)
-                if (ranges::size(seq) != ranges::size(bpp))
+                if (size(seq) != size(bpp))
                     throw parse_error{"Found sequence and associated structure of different length."};
         }
         else
@@ -309,16 +310,16 @@ public:
                comment_type && SEQAN3_DOXYGEN_ONLY(comment),
                offset_type && SEQAN3_DOXYGEN_ONLY(offset))
     {
-        ranges::ostreambuf_iterator stream_it{stream};
+        std::ranges::ostreambuf_iterator stream_it{stream};
 
         // WRITE ID (optional)
         if constexpr (!detail::decays_to_ignore_v<id_type>)
         {
-            if (!ranges::empty(id))
+            if (!empty(id))
             {
                 stream_it = '>';
                 stream_it = ' ';
-                ranges::copy(id, stream_it);
+                std::ranges::copy(id, stream_it);
                 detail::write_eol(stream_it, options.add_carriage_return);
             }
         }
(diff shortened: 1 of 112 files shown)


include/seqan3/io/structure_file/format_vienna.hpp
```diff
@@ -161,21 +161,21 @@ class structure_file_format_vienna
...
-        if (is_id(*ranges::begin(stream_view)))
+        if (is_id(*std::ranges::begin(stream_view)))
         {
             if constexpr (!detail::decays_to_ignore_v<id_type>)
             {
                 if (options.truncate_ids)
                 {
-                    ranges::copy(stream_view | ranges::view::drop_while(is_id || is_blank) // skip leading >
+                    std::ranges::copy(stream_view | ranges::view::drop_while(is_id || is_blank) // skip leading >
                                              | view::take_until_or_throw(is_cntrl || is_blank)
```
smehringer: fix indentation :nail_care:

include/seqan3/io/structure_file/format_vienna.hpp
```diff
@@ -161,21 +161,21 @@ class structure_file_format_vienna
...
                                              | view::char_to<value_type_t<id_type>>,
                                  ranges::back_insert_iterator{id});
                     detail::consume(stream_view | view::take_line_or_throw);
                 }
                 else
                 {
-                    ranges::copy(stream_view | ranges::view::drop_while(is_id || is_blank) // skip leading >
+                    std::ranges::copy(stream_view | ranges::view::drop_while(is_id || is_blank) // skip leading >
                                              | view::take_line_or_throw
```
smehringer: fix indentation :nail_care:

include/seqan3/io/structure_file/format_vienna.hpp
```diff
@@ -189,19 +189,19 @@ class structure_file_format_vienna
...
             }
         }
 
         // READ SEQUENCE
         if constexpr (!detail::decays_to_ignore_v<seq_type>)
         {
             auto constexpr is_legal_seq = is_in_alphabet<seq_legal_alph_type>;
-            ranges::copy(stream_view | view::take_line_or_throw                      // until end of line
+            std::ranges::copy(stream_view | view::take_line_or_throw                      // until end of line
                                      | ranges::view::remove_if(is_space || is_digit) // ignore whitespace and numbers
```
smehringer: fix indentation :nail_care:

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -232,8 +233,8 @@ class fm_index
         // * sdsl construction currently only works for int_vector, std::string and char *, not ranges in general
         // uint8_t largest_char = 0;
         sdsl::int_vector<8> tmp_text(text.size());
-        ranges::copy(text | view::reverse | view::to_rank | view::transform([] (uint8_t const r) { return r + 1; }),
-                     ranges::begin(tmp_text)); // reverse and increase rank by one
+        std::ranges::copy(text | view::reverse | view::to_rank | view::transform([] (uint8_t const r) { return r + 1; }),
+                     std::ranges::begin(tmp_text)); // reverse and increase rank by one
```
smehringer: fix indentation :nail_care:

include/seqan3/std/ranges
```diff
@@ -173,8 +179,171 @@ template <typename type>
...
+
+/*!\interface std::ranges::uncvref_t
+* \brief Alias for ranges::uncvref_t. Obtains the reference type of a top level cv.
+*/
+using ::ranges::uncvref_t;
+
+/*!\interface std::ranges::range_value_type_t
+* \brief Alias for ranges::range_value_type_t. Obtains the value type of a Range type.
+*/
+using ::ranges::range_value_type_t;
```
h-2: use our `value_type_t` from core/metafunction instead

include/seqan3/std/ranges
```diff
@@ -173,8 +179,171 @@ template <typename type>
...
+
+/*!\interface std::ranges::ostreambuf_iterator
+* \brief Alias for ranges::ostreambuf_iterator. Output iterator that writes to std::basic_streambuf.
+*/
+using ::ranges::ostreambuf_iterator;
+
+/*!\interface std::ranges::back_insert_iterator
+* \brief Alias for ranges::back_insert_iterator. Iterator adaptor for insertion at the end of a container.
+*/
+using ::ranges::back_insert_iterator;
```
h-2: We need to think about something here, because the above two should not be in the `std/ranges`, but in `std/iterator`, but then they will conflict with the standard library one :(

include/seqan3/std/ranges
```diff
@@ -173,8 +179,171 @@ template <typename type>
...
+
+/*!\interface std::ranges::make_iterator_range
+* \brief Alias for ranges::make_iterator_range. Makes the iterator adaptor.
+*/
+using ::ranges::make_iterator_range;
+
+/*!\interface std::ranges::back_inserter
+* \brief Alias for ranges::back_inserter. Creates the back inserter.
+*/
+using ::ranges::back_inserter;
```
h-2: same here

include/seqan3/std/ranges
```diff
@@ -173,8 +179,171 @@ template <typename type>
...
+
+/*!\interface std::ranges::any_view
+* \brief Alias for ranges::any_view.
+*/
+using ::ranges::any_view;
+
+/*!\interface std::ranges::reverse_view
+* \brief Alias for ranges::reverse_view.
+*/
+using ::ranges::reverse_view;
```
h-2: don't touch the views for now!

include/seqan3/std/ranges
```diff
@@ -173,8 +179,171 @@ template <typename type>
...
+
+/*!\interface std::ranges::ImplicitlyConvertibleTo
+* \brief Alias for ranges::ImplicitlyConvertibleTo.
+*/
+using ::ranges::ImplicitlyConvertibleTo;
+
+/*!\interface std::ranges::ExplicitlyConvertibleTo
+* \brief Alias for ranges::ExplicitlyConvertibleTo.
+*/
+using ::ranges::ExplicitlyConvertibleTo;
```
h-2: don't add these two, instead use our concepts from core/concept

h-2:
So, we have a little problem here, regarding the iterators:

We have chosen to use just define our own type/functions/templates in the `std::` namespace to make upgrades seamless and the code more readable. This works well for new stuff, because either the name exists already, then we don't define and include the header; or it doesn't exist and we define it, which also works well.

However for things like `std::istreambuf_iterator` it is already defined in the standard library, it's just being changed for c++20. We need to the new one or the one from `ranges::`, but we cannot *overwrite* the one from the standard library.

Maybe we should use `std20::` for those cases? I don't like it, because we create inconsistency again, but I don't see a better solution right now. @marehr @smehringer what do you think?

smehringer:
> We need to the new one or the one from ranges::

? (to include?)

Can you direct to the upcoming changes in `std::istreambuf_iterator`? Just to see an example.. in cppreference it just says `until C++17` but not how it will look like in C++20 :neutral_face:

include/seqan3/std/ranges
joshuak94: But in core/metafunction, should it still reference std::ranges::value_type_t (in core/metafunction/iterator.hpp)

h-2: No, it would be good to resolve to the actual functionality there. I will have a look if I can maybe get that done today.

h-2:
> Can you direct me to the upcoming changes in std::istreambuf_iterator?

You can look it up in the proposal:
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0789r3.pdf

It boils down to making the iterator model `std::InputIterator` which the current one doesn't. We need this, because we create views using the iterator and the view checks the concept.

include/seqan3/range/detail/random_access_iterator.hpp
```diff
@@ -76,7 +76,7 @@ class random_access_iterator_base
     //!\brief Iterator stores pointer to underlying container structure.
     typename std::add_pointer_t<range_type> host{nullptr};
     //!\brief Use container's size_type as a position.
-    using position_type =  ranges::v3::size_type_t<range_type>;
+    using position_type =  std::ranges::size_type_t<range_type>;
```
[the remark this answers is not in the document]
joshuak94: This doesn't work. I've included core/metafunction/all.hpp and it gives a whole bunch of errors. @marehr said it might be because of some cyclic dependency checks?

include/seqan3/std/ranges
```diff
@@ -173,8 +179,171 @@ template <typename type>
...
+
+/*!\interface std::ranges::common_type_t
+* \brief Alias for ranges::common_type_t. Determine the common type of a set of types.
+*/
+using ::ranges::common_type_t;
+
+/*!\interface std::ranges::uncvref_t
+* \brief Alias for ranges::uncvref_t. Obtains the reference type of a top level cv.
+*/
+using ::ranges::uncvref_t;
```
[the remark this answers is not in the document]
joshuak94: This is only in the concepts file. I'd have to change it to seqan3::remove_cvref_t instead of ::ranges::uncvref_t. Is that alright?

include/seqan3/std/ranges
joshuak94: But the implementation of `explicitly_convertible_to_concept` or `implicitly_convertible_to_concept` still rely on `std::ranges::ExplicitlyConvertibleTo`, right? The implementation in core/concept is used throughout seqan, but in core/concept/core_language.hpp, I need to use `std::ranges::ExplicitlyConvertibleTo` or something similar.

include/seqan3/std/ranges
```diff
@@ -173,8 +179,171 @@ template <typename type>
...
+
+/*!\interface std::ranges::difference_type_t
+* \brief Alias for ranges::difference_type_t. Obtains the difference type of a WeaklyIncrementable type.
+*/
+using ::ranges::difference_type_t;
+
+/*!\interface std::ranges::size_type_t
+* \brief Alias for ranges::size_type_t. Obtains the value type of a Size type.
+*/
+using ::ranges::size_type_t;
```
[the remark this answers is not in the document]
smehringer: Our metafunction redirect to the ranges library so we need to keep the alias still, correct?

include/seqan3/std/ranges
h-2: good point. You can change the definition of implicitly_convertible_concept to delegate to `std::is_convertible_v` and for the explicitly_convertible_to_concept check if a `static_cast`  would work. @smehringer can help you with this if something does not work.

include/seqan3/std/ranges
h-2: I have created #512

h-2:
This doesn't fix it for ostreambuf_iterator, though. I can not think of a clean solution, because inheriting conditionally is really ugly: have to add constraints, have to explicitly inherit constructors and all member types :nauseated_face: 

Easiest thing would probably to just add a convenience function `make_ostreambuf_iterator` and have that dispatch based on the cpp standard, but where would we put that?

include/seqan3/std/ranges
joshuak94: Should it be `std::is_convertible_v` or `std::is_nothrow_convertible_v`? (https://en.cppreference.com/w/cpp/types/is_convertible)

smehringer: Use 
```
template <typename t, typename u>
concept explicitly_convertible_to_concept = requires (t vt) { {static_cast<u>(vt)}; };
```

smehringer:
Thanks for clarifying this in a lot of detail in #239!

Although I am not sure if I fully comprehend the issue, I agree that inheriting conditionally is nasty and would complicate our code base. I think a convenience function is the least awful, albeit inconsistent, and can be easily understood and documented.

include/seqan3/std/ranges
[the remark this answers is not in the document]
joshuak94: What about semiregular_t?

include/seqan3/std/ranges
joshuak94: So std::ranges::back_insert_iterator is no longer used, but std::ranges::back_inserter is still used, correct? Do we have a solution for the fact that it should be in std/iterator?

joshuak94:
@h-2 Some things I am not quite sure about yet:

`semiregular_type_t` used in range/container/bitcompressed_vector.hpp, do we have our own implementation for this or should it be `std::ranges::`?
`std::ranges::size_type_t` is used in range/detail/random_access_iterator.hpp, and when I try to use `size_type` from core/metafunction, it doesn't work.
In std/iterator `::std::ranges::difference_type_t` is used, should I use `seqan3::difference_type` instead?

smehringer:
Can you check out the `seqan/performance/std/charconv_benchmark.cpp` file and change the ranges::begin call to std::ranges::begin too? (Newly merged)

test/snippet/io/stream/debug_stream.cpp
```diff
@@ -10,16 +10,16 @@ int main()
 {
 //! [usage]
 // This does not work:
-//std::cout << dna5::C;
+//debug_stream << dna5::C;
 // because the alphabet needs to be converted to char explicitly:
-std::cout << to_char(dna5::C);  // prints 'C'
+debug_stream << to_char(dna5::C);  // prints 'C'
```
h-2: don't change these two

include/seqan3/range/container/bitcompressed_vector.hpp
```diff
@@ -325,7 +326,7 @@ class bitcompressed_vector
         requires has_same_value_type_v<other_range_t>
     //!\endcond
     explicit bitcompressed_vector(other_range_t && range) :
-        bitcompressed_vector{ranges::begin(range), ranges::end(range)}
+        bitcompressed_vector{seqan3::begin(range), seqan3::end(range)}
```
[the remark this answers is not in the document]
joshuak94: Hmmmm, I think something in the file was throwing an error. There is this:
https://github.com/seqan/seqan3/blob/39c535150556944e59dab82612c08c834d6053fc/include/seqan3/range/container/bitcompressed_vector.hpp#L504

Not sure if that affects it.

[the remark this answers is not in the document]
joshuak94: Yeah if I replace either seqan3::begin or seqan3::end with just the function, it won't make.

test/unit/range/view/view_single_pass_input_test.cpp
```diff
@@ -193,7 +193,7 @@ TYPED_TEST(single_pass_input, iterator_pre_increment)
     detail::single_pass_input_view view{p};
 
     auto it = view.begin();
-    if constexpr (std::is_same_v<ranges::range_value_type_t<TypeParam>, char>)
+    if constexpr (std::is_same_v<std::ranges::range_value_type_t<TypeParam>, char>)
```
joshuak94: @h-2 `std::ranges::range_value_type_t` is still used here (and 3 more times in this file).

include/seqan3/std/ranges
```diff
@@ -173,8 +179,101 @@ template <typename type>
 concept ViewableRange = Range<type> && (std::is_lvalue_reference_v<type> || View<type>);
 //!\endcond
 
-//!\}
+/*!\interface std::ranges::begin
```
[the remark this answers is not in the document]
joshuak94: Neither \typedef nor deleting the line seem to work.

The error message is
```cpp
/srv/public/jjk/development/seqan3/include/seqan3/std/ranges:267: warning: documented symbol `std::ranges::ostream_iterator' was not declared or defined.
```

include/seqan3/range/container/bitcompressed_vector.hpp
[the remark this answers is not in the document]
h-2: Ok, then just leave it like it is.

include/seqan3/std/ranges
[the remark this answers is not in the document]
h-2: Hm, we need to investigate this.

test/unit/range/view/view_single_pass_input_test.cpp
h-2: Then just replace it with `value_type_t` from our namespace.

joshuak94:
@h-2 All range in first commit, debug_stream in second commit. Also rebased on current master as of now.

joshuak94:
@h-2 Updated to typedef instead of interface. Squashed down into two commits, one for ranges and one for debug_stream.

Issue #191, opened 2018-03-12
Argument Parser - File extension validator should check for file existence


smehringer:
Should also check for permissons

smehringer:
@MitraDarja Could you look into this issue?

The code of the file_ext_validator was written when the `filesystem` stl header was still experimental. So it would be great if you could also replace the `std::string` passed as an argument to a `filesytem::path` object [here](https://github.com/seqan/seqan3/blob/master/include/seqan3/argument_parser/validators.hpp#L331). You need to inlcude `<seqan3/io/filesystem.hpp>` to use the functionality.

I think [filesystem](https://en.cppreference.com/w/cpp/filesystem/exists) also offers something to check if the file exists. 

Thank you :)

smehringer:
Fixed by #499 introducing a file_existence validator

PR #499, opened 2018-10-08
Checks if file exits in file validation extension, if not throws error

MitraDarja:
#191 

This checks if file exists and if not throws an error, added a test in test file to check if error is thrown.

### DIFF
--- a/include/seqan3/argument_parser/validators.hpp
+++ b/include/seqan3/argument_parser/validators.hpp
@@ -361,7 +361,7 @@ private:
  * The struct then acts as a functor that throws a seqan3::parser_invalid_argument
  * exception whenever a given filename (string) does not exist.
  *
- * \snippet test/snippet/argument_parser/validators_5.cpp usage
+ * \snippet test/snippet/argument_parser/validators_file_existance.cpp usage
  */
 class file_existance_validator
 {


include/seqan3/argument_parser/validators.hpp
```diff
@@ -322,17 +323,21 @@ class file_ext_validator // TODO AFTER INITIAL MERGE: Check if file exists, also
         extensions{v}
     {}
 
-    /*!\brief Tests whether cmp lies inside extensions.
-     * \param cmp The input value to check.
+    /*!\brief Tests whether p lies inside extensions.
```
smehringer: please call it path or filename. One character names are nasty to grep and replace later on :skull_and_crossbones:

include/seqan3/argument_parser/validators.hpp
```diff
@@ -322,17 +323,82 @@ class file_ext_validator // TODO AFTER INITIAL MERGE: Check if file exists, also
...
     {
-        std::string ext{cmp.substr(cmp.find_last_of('.') + 1)};
+      if (!(filesystem::exists(path)))
+        throw parser_invalid_argument(detail::to_string("File ", path, " does not exist."));
 
-        if (!(std::find(extensions.begin(), extensions.end(), ext) != extensions.end()))
-            throw parser_invalid_argument(detail::to_string("Extension ", ext, " is not one of ",
+      std::string cmp = path.string();
+      std::string ext{cmp.substr(cmp.find_last_of('.') + 1)};
+      if (!(std::find(extensions.begin(), extensions.end(), ext) != extensions.end()))
```
smehringer: This can be removed from the file_existence_validator :) They should both only test what they say.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -322,17 +323,82 @@ class file_ext_validator // TODO AFTER INITIAL MERGE: Check if file exists, also
...
+class file_existance_validator // TODO AFTER INITIAL MERGE: Check if file exists, also allow filepath
+{
+public:
+    //!\brief Type of values that are tested by validator
+    using value_type = std::string;
+
+    /*!\brief Constructing from a vector.
+     * \param[in] v The vector of valid file extensions to test (e.g. {"fa", "fasta"}).
+     */
+    file_existance_validator(std::vector<std::string> const & v) :
```
smehringer: The file existence validator does not need constructors now because there are no extension to test. You can delete them and in the test just default initialize it `file_existance_validator my_validator{};`.

Also the private value `extensions` is not needed anymore.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -322,17 +323,82 @@ class file_ext_validator // TODO AFTER INITIAL MERGE: Check if file exists, also
...
+    {
+        return detail::to_string("File name extension must be one of ", view::all(extensions), ".");
+    }
+
+private:
+    //!\brief Stores valid file extensions.
+    std::vector<std::string> extensions;
+};
+
+/*!\brief A validator that checks if a filenames has one of the valid extensions and also exits.
```
smehringer: Adjust documentation

include/seqan3/argument_parser/validators.hpp
```diff
@@ -322,17 +323,82 @@ class file_ext_validator // TODO AFTER INITIAL MERGE: Check if file exists, also
...
+ *
+ * \details
+ *
+ * On construction, the validator must receive a list (vector) of valid file extensions.
+ * The struct than acts as a functor, that throws a seqan3::parser_invalid_argument
+ * exception whenever a given filename (string) is not in the given extension list.
+ *
+ * \snippet test/snippet/argument_parser/validators_3.cpp usage
+ */
+class file_existance_validator // TODO AFTER INITIAL MERGE: Check if file exists, also allow filepath
```
smehringer: The TODO comment can be removed now :sparkles:

include/seqan3/argument_parser/validators.hpp
```diff
@@ -322,16 +323,16 @@ class file_ext_validator // TODO AFTER INITIAL MERGE: Check if file exists, also
...
      */
-    void operator()(std::string const & cmp) const
+    void operator()(filesystem::path const & path) const
     {
-        std::string ext{cmp.substr(cmp.find_last_of('.') + 1)};
-
-        if (!(std::find(extensions.begin(), extensions.end(), ext) != extensions.end()))
-            throw parser_invalid_argument(detail::to_string("Extension ", ext, " is not one of ",
+      std::string cmp = path.string();
+      std::string ext{cmp.substr(cmp.find_last_of('.') + 1)};
```
smehringer: Having a filesystem path now, we can use it's member function [`filesystem::path::extension()`](https://en.cppreference.com/w/cpp/filesystem/path/extension) to get the extension. This ensures a correct extension even for some cases that we did not cover yet (e.g. `./` is a valid directory path but does not have a file extension. Our substr() command right now takes `/` as the extension while the fileysytem returns an empty extension).
Unfortunately `path.extension()` returns the extension WITH the dot so we need to drop that:
```
std::string ext{path.extension().string()};
ext = ext.substring(std::min(1, ext.size())); // drop '.' if extension is non-empty
```

include/seqan3/argument_parser/validators.hpp
```diff
@@ -352,6 +353,39 @@ class file_ext_validator // TODO AFTER INITIAL MERGE: Check if file exists, also
...
+ *
+ * \snippet test/snippet/argument_parser/validators_3.cpp usage
+ */
+class file_existance_validator
+{
+public:
+    //!\brief Type of values that are tested by validator
+    using value_type = std::string;
+
+    /*!\brief Tests whether path lies inside extensions.
```
smehringer: Tests whether path exists

include/seqan3/argument_parser/validators.hpp
```diff
@@ -352,6 +353,39 @@ class file_ext_validator // TODO AFTER INITIAL MERGE: Check if file exists, also
...
+     * \param path The input value to check.
+     * \throws parser_invalid_argument
+     */
+    void operator()(filesystem::path const & path) const
+    {
+      if (!(filesystem::exists(path)))
```
smehringer: indent 4 spaces not 2

include/seqan3/argument_parser/validators.hpp
```diff
@@ -352,6 +353,39 @@ class file_ext_validator // TODO AFTER INITIAL MERGE: Check if file exists, also
...
+        throw parser_invalid_argument(detail::to_string("File ", path, " does not exist."));
+    }
+
+    //!\brief Tests whether every value of v lies inside extensions.
```
smehringer: Tests whether every filename in list v exists.

h-2:
Beware that extension() includes the dot!

test/unit/argument_parser/format_parse_validators_test.cpp
```diff
@@ -52,6 +53,21 @@ TEST(validator_test, fullfill_concept)
     EXPECT_TRUE(validator_concept<file_ext_validator>);
 }
 
+TEST(validator_test, no_file)
+{
+  filesystem::path p{"./sandbox.fasta"};
```
smehringer: 4 spaces instead of 2 :nerd_face: 
Usually you can set up you editor to automatically indent by 4 when using tab.

Could you also add a string variable for each test:
```cpp
    filesystem::path p{"./sandbox.fasta"};
    std:;string s{"./stonebox.fasta"};
    file_existance_validator my_validator{};
    EXPECT_THROW(my_validator(p), parser_invalid_argument);
    EXPECT_THROW(my_validator(s), parser_invalid_argument);
``` 
because it should work on strings

include/seqan3/argument_parser/validators.hpp
```diff
@@ -322,16 +323,16 @@ class file_ext_validator // TODO AFTER INITIAL MERGE: Check if file exists, also
...
      * \throws parser_invalid_argument
      */
-    void operator()(std::string const & cmp) const
+    void operator()(filesystem::path const & path) const
     {
-        std::string ext{cmp.substr(cmp.find_last_of('.') + 1)};
-
-        if (!(std::find(extensions.begin(), extensions.end(), ext) != extensions.end()))
-            throw parser_invalid_argument(detail::to_string("Extension ", ext, " is not one of ",
+      std::string ext{path.extension().string()};
```
eseiler: Indentation

include/seqan3/argument_parser/validators.hpp
```diff
@@ -352,6 +353,39 @@ class file_ext_validator // TODO AFTER INITIAL MERGE: Check if file exists, also
     std::vector<std::string> extensions;
 };
 
+/*!\brief A validator that checks if a file exits.
+ * \ingroup argument_parser
+ *
+ * \details
+ *
+ * The struct than acts as a functor, that throws a seqan3::parser_invalid_argument
```
eseiler: th~~a~~ acts as a functor ~~,~~ that throws

include/seqan3/argument_parser/validators.hpp
```diff
@@ -352,6 +353,39 @@ class file_ext_validator // TODO AFTER INITIAL MERGE: Check if file exists, also
     std::vector<std::string> extensions;
 };
 
+/*!\brief A validator that checks if a file exits.
```
eseiler: exi**s**ts

test/unit/argument_parser/format_parse_validators_test.cpp
```diff
@@ -52,6 +53,26 @@ TEST(validator_test, fullfill_concept)
...
+
+TEST(validator_test, file_exists)
+{
+    filesystem::path p{"./"};
+    std::string s{"./"};
+    file_existance_validator my_validator{};
+    for (auto & file : filesystem::directory_iterator(p))
+        EXPECT_NO_THROW(my_validator(file));
+    for (auto & file : filesystem::directory_iterator(s))
+          EXPECT_NO_THROW(my_validator(file));
```
h-2: How do you know the current directory contains something?

MitraDarja: I assumed the current directory would be the directory the test file itself is in. Is that assumption not valid?

include/seqan3/argument_parser/validators.hpp
```diff
@@ -322,14 +323,14 @@ class file_ext_validator // TODO AFTER INITIAL MERGE: Check if file exists, also
...
+     * \param path The input value to check.
      * \throws parser_invalid_argument
      */
-    void operator()(std::string const & cmp) const
+    void operator()(filesystem::path const & path) const
     {
-        std::string ext{cmp.substr(cmp.find_last_of('.') + 1)};
-
+        std::string ext{path.extension().string()};
+        ext = ext.substr(std::min(1, static_cast<int>(ext.size()))); // drop '.' if extension is non-empty
```
h-2: I am not really hooked on filesystem::path::extension(), because it can return empty strings, too, which we never want, do we?

Getting the extension with views could look like this:
```cpp
auto v = path | view::reverse | view::take_until_or_throw(is_char<'.'>) | view::reverse;
```

We could even save this as `view::file_extension`...

include/seqan3/argument_parser/validators.hpp
```diff
@@ -352,6 +353,39 @@ class file_ext_validator // TODO AFTER INITIAL MERGE: Check if file exists, also
...
+     * \throws parser_invalid_argument
+     */
+    void operator()(filesystem::path const & path) const
+    {
+        if (!(filesystem::exists(path)))
+            throw parser_invalid_argument(detail::to_string("File ", path, " does not exist."));
+    }
+
+    //!\brief Tests whether every filename in list v exists.
+    void operator()(std::vector<std::string> const & v) const
```
h-2: `std::vector<filesystem::path>`?

test/unit/argument_parser/format_parse_validators_test.cpp
h-2: I don't know, the test need not be run in the directory of the binary, it could be another test directory. You could instead use our `tmp_filename()` to create a file for the test!

include/seqan3/argument_parser/validators.hpp
smehringer: > which we never want, do we?

I see no problem since an empty extension will not match any valid extensions.

Your example would need an extra try catch block (we only want validation exceptions to be thrown on failure).

And it would also return an empty string on `.` as a file name.

test/unit/argument_parser/format_parse_validators_test.cpp
```diff
@@ -52,6 +54,27 @@ TEST(validator_test, fullfill_concept)
...
+    EXPECT_THROW(my_validator(s), parser_invalid_argument);
+}
+
+TEST(validator_test, file_exists)
+{
+    test::tmp_filename("./sandbox2.fasta");
+    filesystem::path p{"./"};
+    std::string s{"./"};
+    file_existance_validator my_validator{};
+    for (auto & file : filesystem::directory_iterator(p))
```
smehringer: You need to actually use the `tmp_filename` to create a file that exists! 
```cpp
test::tmp_filename tmp_file_name{"./sandbox2.fasta"};
std::ofstream tmp_file(tmp_file_name.get_path());
tmp_file << "foo\n"; // I am actually not sure if you need this line because the constructor may already create the file
```
and then 
```suggestion
    for (auto & file : filesystem::directory_iterator(tmp_filename.get_path()))
```

include/seqan3/argument_parser/validators.hpp
smehringer: @h-2 Can I resolves this?

test/unit/argument_parser/format_parse_validators_test.cpp
smehringer: @MitraDarja ping :)

include/seqan3/argument_parser/validators.hpp
```diff
@@ -322,14 +323,14 @@ class file_ext_validator // TODO AFTER INITIAL MERGE: Check if file exists, also
         extensions{v}
     {}
 
-    /*!\brief Tests whether cmp lies inside extensions.
-     * \param cmp The input value to check.
+    /*!\brief Tests whether path lies inside extensions.
```
```suggestion
    /*!\brief Tests whether the filepath \p path ends with a valid extension.
```
:nail_care:

include/seqan3/argument_parser/validators.hpp
```diff
@@ -352,6 +353,39 @@ class file_ext_validator // TODO AFTER INITIAL MERGE: Check if file exists, also
...
 
+/*!\brief A validator that checks if a file exists.
+ * \ingroup argument_parser
+ *
+ * \details
+ *
+ * The struct then acts as a functor that throws a seqan3::parser_invalid_argument
+ * exception whenever a given filename (string) does not exist.
+ *
+ * \snippet test/snippet/argument_parser/validators_3.cpp usage
```
smehringer: I think this is the snippet from the file_extension_validator. There needs to be a new one for this new validator :nerd_face:

include/seqan3/argument_parser/validators.hpp
```diff
@@ -352,6 +353,39 @@ class file_ext_validator // TODO AFTER INITIAL MERGE: Check if file exists, also
...
+ */
+class file_existance_validator
+{
+public:
+    //!\brief Type of values that are tested by validator
+    using value_type = std::string;
```
smehringer: This needs to be `filesystem::path` I think! When you create the snippet I mentioned before this could otherwise be a source of a compiler error.

test/unit/argument_parser/format_parse_validators_test.cpp
```diff
@@ -52,6 +55,24 @@ TEST(validator_test, fullfill_concept)
...
 }
 
+TEST(validator_test, no_file)
+{
+    filesystem::path p{"./sandbox.fasta"};
+    std::string s{"./stonebox.fasta"};
+    file_existance_validator my_validator{};
+    EXPECT_THROW(my_validator(p), parser_invalid_argument);
+    EXPECT_THROW(my_validator(s), parser_invalid_argument);
+}
```
smehringer: Can you add for both tests an actual argument parser to test the functionality in connection with the parser?
The other tests are examples for this. Something like...
```cpp
    filesystem::path file_in_path;
    std::vector<filesystem::path> path_vector;
 
     // option
     const char * argv[] = {"./argument_parser_test", "-i", "./sandbox.fasta"};
     argument_parser parser("test_parser", 3, argv);
     parser.add_option(file_in_path, 'i', "int-option", "desc",
                       option_spec::DEFAULT, file_existence_validator());
 
     EXPECT_THROW(parser.parse(), parser_invalid_argument);
```

and again with std::string. There might a compiler warning with string... If so come and get me :)

test/unit/argument_parser/format_parse_validators_test.cpp
```diff
@@ -50,6 +53,36 @@ TEST(validator_test, fullfill_concept)
...
+     argument_parser parser("test_parser", 3, argv);
+     parser.add_option(file_in_path, 'i', "int-option", "desc",
+                       option_spec::DEFAULT, file_existance_validator());
+
+     EXPECT_THROW(parser.parse(), parser_invalid_argument);
+}
+
+TEST(validator_test, file_exists)
+{
+    test::tmp_filename tmp_file_name{""};
```
smehringer: The vary last things I think:

Can you give it a name instead of an empty string? 
and then I think you can remove the for loop because there will be only this one file in the directory.

Can you also add an argument_parser example here ?

test/unit/argument_parser/format_parse_validators_test.cpp
```diff
@@ -50,6 +53,46 @@ TEST(validator_test, fullfill_concept)
...
+    filesystem::path file_in_path;
+    std::vector<filesystem::path> path_vector;
+    
+    // option
+    const char * argv[] = {"./argument_parser_test", "-i", "./testbox.fasta"};
+    argument_parser parser("test_parser", 3, argv);
+    parser.add_option(file_in_path, 'i', "int-option", "desc",
+                      option_spec::DEFAULT, file_existance_validator());
+
+    EXPECT_THROW(parser.parse(), parser_invalid_argument);
```
smehringer: This should be EXPECT_NO_THROW? :hushed:

PR #500, opened 2018-10-09
[FEATURE] std::from_char and std::to_char functions, tests and benchmarks.

smehringer:
This PR introduces the std::from_chars functionality because

1. In gcc7 none of the std::from_chars overloads are implemented
2. In gcc8 only the **integral** std::from_chars overloads are implemented

For 1) The code of LLVM was copied and documented as such.
For 2) The functions `strto[f/d/ld]` were used as a workaround to provide the std::from_chars functionality for floating point values.

#### Issues
* The return type: The return type of std::from_chars is a struct containing an error code and a char pointer. The char pointer is documented to be set to the position where a character is NOT matching the correct parsing pattern (e.g. illegal character, or wrong format). As the functions  `strto[f/d/ld]` do not provide this functionality, this behaviour could only be mimiced by extensive string comparison. IMHO this functionality should not be provided by the workaround and documented as it is.

* The locale: std::from_chars is documented to be locale independent. The accepted patterns are identical to the one used by strtod in the defailt ("C") locale. The functions strto[d/f/ld] used here are locale dependent but setting the locale manually by std::setlocale is not thread safe. IMHO for the time being the workaround may remain locale dependent.

#### Todo
* [x] Include guard for from_chars for floating point to be only included if not implemented in the charcon header.

#### Design decisions
Copying from boost is not easily possible as the code includes large parts of the boost basic library and does some wierd stuff (if anyone is interested in understanding the code, [this file](https://github.com/boostorg/spirit/blob/b4c5ef702bf6c28e964a84c9e9abe1a6549bce69/include/boost/spirit/home/x3/support/numeric_utils/detail/extract_int.hpp) is one of the core implementations for integer parsing which is needed for float parsing).
Using the functions  `strto[f/d/ld]` is much faster then allocating a string stream each time. 
Using a manual workaround, e.g. parsing the string as integer ignoring the decimal point and then dividing by 10^[digits after decimal points] would be more efficient but a ton of work to also allow for scientific notation, hexadecimal numbers and so on..

#### Benchmarks
grafik

### DIFF
--- a/include/seqan3/std/charconv
+++ b/include/seqan3/std/charconv
@@ -1,0 +1,285 @@
+// ============================================================================
+//                 SeqAn - The Library for Sequence Analysis
+// ============================================================================
+//
+// Copyright (c) 2006-2018, Knut Reinert & Freie Universitaet Berlin
+// Copyright (c) 2016-2018, Knut Reinert & MPI Molekulare Genetik
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//     * Redistributions of source code must retain the above copyright
+//       notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//     * Neither the name of Knut Reinert or the FU Berlin nor the names of
+//       its contributors may be used to endorse or promote products derived
+//       from this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL KNUT REINERT OR THE FU BERLIN BE LIABLE
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+// DAMAGE.
+//
+// ============================================================================
+
+/*!\file
+ * \brief Provides std::from_chars and std::to_chars if not defined in the stl \<charconv\> header.
+ * \author Svenja Mehringer <svenja.mehringer AT fu-berlin.de>
+ */
+
+#pragma once
+
+#if __has_include(<version>)
+#include <version> // from C++20 all feature macros should be defined here
+#endif
+
+#include <utility> // __cpp_lib_to_chars may be defined here as currently documented
+
+#if __cpp_lib_to_chars >= 201611
+#include <charconv>
+#else // else include our own implementation below
+
+/*!\defgroup charconv charconv
+ * \ingroup std
+ * \brief The \<charconv\> header from C++17's standard library.
+ */
+
+// -----------------------------------------------------------------------------
+// structs: to_chars_result, from_chars_result and chars_format
+// -----------------------------------------------------------------------------
+
+namespace std
+{
+
+//!\brief Result type of std::to_chars
+//!\ingroup charconv
+struct to_chars_result
+{
+    char * ptr;  //!< A pointer to the output string.
+    errc ec;    //!< The error code.
+};
+
+//!\brief Result type of std::from_chars
+//!\ingroup charconv
+struct from_chars_result
+{
+    char const * ptr;  //!< A pointer to the input string.
+    errc ec;          //!< The error code.
+};
+
+//!\brief A BitmaskType used to specify floating-point formatting for std::to_chars and std::from_chars.
+//!\ingroup charconv
+enum class chars_format
+{
+    scientific = 0x1,             //!< Scientific notation, e.g. 3.991E-0003.
+    fixed = 0x2,                  //!< Fixed number of digits for precision.
+    hex = 0x4,                    //!< Hexadecimal notation. Ff specified in from_chars, prefix 0x,0X,x is not allowed.
+    general = fixed | scientific  //!< General use case.
+};
+
+} // namespace std
+
+#include <seqan3/std/charconv_detail.hpp> // structs definitions above are needed in this header
+
+namespace std
+{
+//!\cond
+// implementation detail taken from LLVM
+void to_chars(char *, char *, bool, int = 10) = delete;
+void from_chars(char const *, char const *, bool, int = 10) = delete;
+//!\endcond
+
+// -----------------------------------------------------------------------------
+// to_chars for integral types
+// -----------------------------------------------------------------------------
+
+/*!\brief Convert an integral into a char sequence.
+ * \ingroup charconv
+ * \tparam value_type The type to convert to a char sequence; Must model std::Integral.
+ * \param[in]      first The start of the range to fill.
+ * \param[in]      last  The end of the range to fill.
+ * \param[in, out] value The value to store the parsed result in.
+ * \param[in]      base  The integer base of the format of the string to parse.
+ *                       Must be a value between 2 and 36 (inclusive).
+ * \returns A std::to_char_result. See detail section return value for more information.
+ *
+ * \details
+ *
+ * Converts value into a character string by successively filling the range
+ * [first, last), where [first, last) is required to be a valid range.
+ *
+ * value is converted to a string of digits in the given base (with no redundant
+ * leading zeroes). Digits in the range 10..35 (inclusive) are represented as
+ * lowercase characters a..z. If value is less than zero, the representation
+ * starts with a minus sign. The library provides overloads for all signed and
+ * unsigned integer types and for the type char as the type of the parameter value.
+ *
+ * ### Return value
+ * On success, returns a value of type to_chars_result such that ec equals
+ * value-initialized std::errc and ptr is the one-past-the-end pointer of the
+ * characters written. Note that the string is not NUL-terminated.
+ *
+ * On error, returns a value of type to_chars_result holding
+ * std::errc::value_too_large in ec, a copy of the value last in ptr, and leaves
+ * the contents of the range [first, last) in unspecified state.
+ *
+ * \sa https://en.cppreference.com/w/cpp/utility/to_chars
+ */
+template <std::Integral value_type>
+inline std::to_chars_result to_chars(char * first, char * last, value_type value, int base)
+{
+    assert(2 <= base && base <= 36);
+    return seqan3::detail::to_chars_integral(first, last, value, base, is_signed<value_type>());
+}
+
+//!\brief std::to_chars overload with default base = 10.
+//!\ingroup charconv
+template <std::Integral value_type>
+inline std::to_chars_result to_chars(char * first, char * last, value_type value)
+{
+    return seqan3::detail::to_chars_itoa(first, last, value, is_signed<value_type>());
+}
+
+// -----------------------------------------------------------------------------
+// from_chars for integral types
+// -----------------------------------------------------------------------------
+
+/*!\brief Parse a char sequence into an integral.
+ * \ingroup charconv
+ * \tparam value_type The type to parse the string into; Must model std::Integral.
+ * \param[in]      first The start of the string to parse.
+ * \param[in]      last  The end of the string to parse.
+ * \param[in, out] value The value to store the parsed result in.
+ * \param[in]      base  The integer base of the format of the string to parse.
+ *                       Must be a value between 2 and 36 (inclusive).
+ * \returns A std::from_char_result. See detail section return value for more information.
+ *
+ * \details
+ *
+ * Analyzes the character sequence [first,last) for a pattern described below.
+ * If no characters match the pattern or if the value obtained by parsing the
+ * matched characters is not representable in the type of value, value is
+ * unmodified, otherwise the characters matching the pattern are interpreted as
+ * a text representation of an arithmetic value, which is stored in value.
+ *
+ * Expects the pattern identical to the one used by std::strtol in the default
+ * ("C") locale and the given non-zero numeric base, except that "0x" or "0X"
+ * prefixes are not recognized for base 16, and that only the minus sign is
+ * recognized (not the plus sign), and only for signed integer types of value.
+ * Digits in the range 10..35 (inclusive) are represented as lowercase or
+ * uppercase characters a..z/A...Z. The library provides overloads for all
+ * signed and unsigned integer types and char as the referenced type of the
+ * parameter value.
+ *
+ * ### Return value
+ * On success, returns a value of type from_chars_result such that ptr points at
+ * the first character not matching the pattern, or has the value equal to last
+ * if all characters match and ec is value-initialized.
+ *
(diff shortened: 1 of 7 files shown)


include/seqan3/std/charconv
```diff
@@ -0,0 +1,1010 @@
...
+ * \brief The \<charconv\> header from C++20's standard library.
+ */
+
+// -----------------------------------------------------------------------------
+// from_char for integral types
+// -----------------------------------------------------------------------------
+
+#if __has_include(<charconv>)
+#include <charconv>
+#else // implement
```
marehr: a TODO?

include/seqan3/std/charconv
```diff
@@ -0,0 +1,1010 @@
...
+};
+
+template <typename _Tp>
+struct traits_base<_Tp, decltype(void(uint32_t{declval<_Tp>()}))>
+{
+    using type = uint32_t;
+
+    static int width(_Tp v)
+    {
+        auto t = (32 - __builtin_clz(v | 1)) * 1233 >> 12;
```
marehr: okay this only works on gcc and clang.

include/seqan3/std/charconv
```diff
@@ -0,0 +1,1010 @@
...
+    auto c = a * b;
+    r = c;
+    return c > (numeric_limits<unsigned short>::max)();
+}
+
+template <typename _Tp>
+inline bool mul_overflowed(_Tp a, _Tp b, _Tp& r)
+{
+    static_assert(is_unsigned<_Tp>::value, "");
+    return __builtin_mul_overflow(a, b, &r);
```
marehr: again only in gcc and clang

include/seqan3/std/charconv
```diff
@@ -0,0 +1,1010 @@
...
+        for (; first1 < last1; ++first1, ++first2)
+            init = init + *first1 * *first2;
+        return init;
+    }
+};
+
+}  // namespace itoa
+
+template <typename _Tp>
+inline _Tp complement(_Tp x)
```
marehr: Hmm okay, so `complement` is now public API in std? And the following functions?

include/seqan3/std/charconv
```diff
@@ -0,0 +1,1010 @@
...
+#include <seqan3/core/concept/core_language.hpp>
+#include <seqan3/io/stream/parse_condition.hpp>
+#include <seqan3/std/concepts>
+
+namespace std
+{
+
+// The inline namespace will allow to use the std::chars_format from the stl if
+// available or fall back to the implementation below if not.
+// In addition, the function std::from_chars must be templetized further ini order
```
marehr: typo in `templetized`

include/seqan3/std/charconv
smehringer: no. the following code is the implement :laughing: 
I will change the misleading comment

include/seqan3/std/charconv
smehringer: Do we plan to add support for anything else sometime soon?

include/seqan3/std/charconv
smehringer: Hmm... that's what LLVM did but none of the functions is documented in cppreference so I guess they should not appear in the std namespace.
We could add them to the iota namespace? or wrap them in detail?

include/seqan3/std/charconv
```diff
@@ -0,0 +1,1010 @@
...
+
+// -*- C++ -*-
+//===------------------------------ charconv ------------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is dual licensed under the MIT and the University of Illinois Open
+// Source Licenses. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
```
h-2: You need to actually paste the contents of LICENSE.TXT since we are not shipping that!

include/seqan3/std/charconv
```diff
@@ -0,0 +1,1010 @@
...
+
+/*!\file
+ * \brief Implements std::from_char.
```
h-2: from_char**s** and also to_chars

include/seqan3/std/charconv
h-2: it shouldn't be else, because the file could contain only one implementation (from OR to) or none at all. You need to check the feature macros. I hope they have different ones for from/to

include/seqan3/std/charconv
```diff
@@ -0,0 +1,1010 @@
...
+    char const * ptr;  //!< A pointer to the input string.
+    errc ec;          //!< The error code.
+};
+
+//!\cond
+// implementation detail taken from LLVM
+void to_chars(char*, char*, bool, int = 10) = delete;
+void from_chars(char const *, char const *, bool, int = 10) = delete;
+
+namespace itoa
```
h-2: is this supposed to be a public namespace?

include/seqan3/std/charconv
```diff
@@ -0,0 +1,1010 @@
...
+{
+
+static constexpr uint64_t pow10_64[] =
+{
+    UINT64_C(0),
```
h-2: what does UINT64_C do?

include/seqan3/std/charconv
h-2: It's ok, because MSVC has the actual functions already implemented and they will be picked up.

include/seqan3/std/charconv
```diff
@@ -0,0 +1,1010 @@
...
+    else if (in_pattern(c))
+        return {true, c - '0'};
+    else if ('a' <= c && c < 'a' + base - 10)
+        return {true, c - 'a' + 10};
+    else
+        return {'A' <= c && c < 'A' + base - 10, c - 'A' + 10};
+}
+
+template <typename _It, typename _Tp, typename _Fn, typename... _Ts>
+inline from_chars_result subject_seq_combinator(_It first, _It last, _Tp& value, _Fn f, _Ts... args)
```
h-2: in theory we are not allowed to use names that start with underscore and then a capital letter.

include/seqan3/std/charconv
h-2: also British English!

include/seqan3/std/charconv
```diff
@@ -0,0 +1,1010 @@
...
+        return {last, std::errc::invalid_argument};
+    }
+
+    // Success.
+    value = tmp;
+    return {last, {}};
+}
+
+} // inline namespace charconv to trick the compiler
+
```
h-2: Depending on how much of this file is copy'n'paste from LLVM, it might be easier to copy the entire file as-is,  save it as charconv_llvm and include based on some condition OR we import it properly, but then we should adapt the naming of the variables so they don't violate anything and also move stuff to detail that is detail.

include/seqan3/std/charconv
smehringer: Look for the code below. The else is ONLY for the integral from_char implementation

The floating point from_char implementation is wrapped differently.. a little hacky though

include/seqan3/std/charconv
smehringer: I don't know.. No I guess since there is no such namespace documented in cpprefference

include/seqan3/std/charconv
smehringer: There is no feature macro to distinguish between different feature completeness (as far as I can tell).
We can use the same trick as for the lfoatin gpoint implementation like @marehr showed me.

include/seqan3/std/charconv
smehringer: [Appends the correct suffix to a 64-bit unsigned integer literal.](https://jameshfisher.com/2017/01/20/uint64_c_macro.html)

smehringer:
Please take a look at the new structure (before reviewing in detail).
I moved the LLVM copied code into our detail namespace in a separate file, while the structs and functions to appear public are in the `seqan3/std/charconv` header. The workaround to include an inline namespace is everything else than pretty and could cause conflicts IF the stl decides to also use inline namespaces (which none does so far, tested on gcc7 and gcc8, viewed clang source code) but I have no better solution?

h-2:
> Please take a look at the new structure (before reviewing in detail).
> I moved the LLVM copied code into our detail namespace in a separate file, while the structs and functions to appear public are in the seqan3/std/charconv header. 

Hm, now that you have renamed everything to our style and even put it into `seqan3::` it doesn't look like llvm anymore and maybe we should put all implementation in the detail file (also our implementation)? Then we could just have the wrapper-functions in the main header and it would be more readable.

> The workaround to include an inline namespace 

I still think we should just use the feature test macro `__cpp_lib_to_chars`. I just verified that GCC does not set this, because it doesn't offer full support, yet. So it's fairly easy to just check for the macro and not have to do other hacks. The only thing will be that we will use our own/llvm's integer conversion on GCC8, but that's not a big issue from my POV. And if we really want to use GCC8's function, we could use an extra macro check for that.

h-2:
> Do you mean moving the floating point implementation?

Yes! So only the small wrappers are in the file.

> I'll check LLVM again I thought they set the macro already.

Would be cool.

include/seqan3/std/charconv
```diff
@@ -0,0 +1,294 @@
...
+} // namespace std
+
+#include <seqan3/std/charconv_detail.hpp> // structs definitions above are needed in this header
+
+namespace std
+{
+//!\cond
+// implementation detail taken from LLVM
+void to_chars(char*, char*, bool, int = 10) = delete;
+void from_chars(char const *, char const *, bool, int = 10) = delete;
```
h-2: I don't think you need the extra constraints anymore, now, do you?

include/seqan3/std/charconv
```diff
@@ -0,0 +1,294 @@
...
+ * characters written. Note that the string is not NUL-terminated.
+ *
+ * On error, returns a value of type to_chars_result holding
+ * std::errc::value_too_large in ec, a copy of the value last in ptr, and leaves
+ * the contents of the range [first, last) in unspecified state.
+ *
+ * \sa https://en.cppreference.com/w/cpp/utility/to_chars
+ */
+template <typename char_type, std::Integral _Tp>
+    requires std::Same<char_type, char>
```
h-2: Same thing here. If you do end up needing, please use the shorter concept syntax

include/seqan3/std/charconv
```diff
@@ -0,0 +1,294 @@
...
+ * such that ec equals std::errc::result_out_of_range and ptr points at the first
+ * character not matching the pattern. value is unmodified.
+ *
+ * \sa https://en.cppreference.com/w/cpp/utility/from_chars
+ */
+template <typename char_type, std::Integral _Tp>
+    requires std::Same<char_type, char>
+inline std::from_chars_result from_chars(char_type const * first,
+                                         char_type const * last,
+                                         _Tp & value,
```
h-2: please replace `_Tp` since we are not allowed to use that :nail_care:

include/seqan3/std/charconv_detail.hpp
```diff
@@ -0,0 +1,843 @@
...
+// ============================================================================
+
+// -*- C++ -*-
+//===------------------------------ charconv ------------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is dual licensed under the MIT and the University of Illinois Open
+// Source Licenses. See LICENSE.TXT for details.
+//
```
h-2: this piece is not needed anymore since you pasted the license into the file.

include/seqan3/std/charconv_detail.hpp
```diff
@@ -0,0 +1,843 @@
...
+//       endorse or promote products derived from this Software without specific
+//       prior written permission.
+
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+// FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
+// CONTRIBUTORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS WITH THE
+// SOFTWARE.
```
h-2: This is the same license we are using so you can just compress it into one paragraph with multiple copyright holders.

include/seqan3/std/charconv_detail.hpp
```diff
@@ -0,0 +1,843 @@
...
+
+// Program             Directory
+// -------             ---------
+// Autoconf            llvm/autoconf
+//                     llvm/projects/ModuleMaker/autoconf
+//                     llvm/projects/sample/autoconf
+// CellSPU backend     llvm/lib/Target/CellSPU/README.txt
+// Google Test         llvm/utils/unittest/googletest
+// OpenBSD regex       llvm/lib/Support/{reg*, COPYRIGHT.regex}
+// =============================================================================
```
h-2: This part is not necessary, because the respective directories/files were not pasted into SeqAn.

include/seqan3/std/charconv_detail.hpp
```diff
@@ -0,0 +1,843 @@
...
+
+/*!\file
+ * \brief Copied code from LLVM for std::from_chars and std::to_chars.
```
h-2: "Provides LLVM's implementation of std::from_chars and std::to_chars for integral types and a custom solution for floating points types."

test/performance/io/charconv_benchmark.cpp
```diff
@@ -0,0 +1,257 @@
...
+static void to_char(benchmark::State& state)
+{
+    arithmetic_type val{120};
+    std::stringstream ss;
+    char buffer[10];
+
+    for (auto _ : state)
+    {
+        std::memset(buffer, 0, sizeof(buffer));
+        std::to_chars(&buffer[0], &buffer[0] + sizeof(buffer), val);
```
h-2: `..., &buffer[10], ...` ?

test/performance/io/charconv_benchmark.cpp
```diff
@@ -0,0 +1,257 @@
...
+        ss << buffer;
```
h-2: You might want to clear ss.rdbuf regularly so that it doesn't dynamically grow a buffer internally. I suspect it would be faster and show the influence of the actual operation more clearly, but I am not 100% sure.
Also, do you have to clear buffer everytime? It will be overwritten anyways, or not?

test/unit/std/charconv_test.cpp
```diff
@@ -0,0 +1,502 @@
...
+    char buffer[10];
+
+        // buffer.clear();
+    [[maybe_unused]] auto res = std::to_chars(&buffer[0], &buffer[0] + sizeof(buffer), val);
+
+}
+
+// =============================================================================
+// std::from_chars for float, double and long double
+// =============================================================================
```
h-2: maybe make two files instead?

test/unit/std/charconv_test.cpp
```diff
@@ -0,0 +1,502 @@
...
+
+        auto res = std::from_chars(&str[0], &str[0] + str.size(), value, 16);
+
+        EXPECT_EQ(res.ptr, &str[0]);
+        EXPECT_EQ(res.ec, std::errc::invalid_argument);
+        EXPECT_EQ(value, TypeParam{42});
+    }
+}
+
+// GCC and Clang have different behaviour in debug mode for an invalid base parameter
```
h-2: Hm, if the behaviour is not specified, we shouldn't test it I think.

If you really want to leave this code inside, do not comment it via `//`, instead use `#if 0` and clearly state under which condition it shall be activated again.

test/performance/io/charconv_benchmark.cpp
smehringer: > Also, do you have to clear buffer every time? It will be overwritten anyway, or not?

In this case, I would not have to clear the buffer as I always read the same number. But if this should be implemented to generically read to chars, I will not know if I read a smaller (in terms of digits) value afterwards and I do need to clear the buffer every time. So clearing it here I think better resembles the use case.

smehringer: > You might want to clear ss.rdbuf regularly so that it doesn't dynamically grow a buffer internally. I suspect it would be faster and show the influence of the actual operation more clearly, but I am not 100% sure.

Clearing the buffer decreases runtime.

test/unit/std/charconv_test.cpp
smehringer: fine

test/performance/io/CMakeLists.txt
```diff
@@ -1,2 +0,0 @@
-
-seqan3_benchmark(parse_condition_benchmark.cpp)
```
h-2: but why :cry:

smehringer: because I can :smiling_imp:

PR #501, opened 2018-10-09
[FEATURE] add seqan3::detail::transformation_trait_or

marehr:
A convenience function to give a default type if `::type` member is not defined.

example:
```c++
template <typename T>
struct A;

template <>
struct A<int>
{
    using type = int;
};

template <typename T>
struct B;

// B<unsigned>::type is not defined, thus falling back to `size_t`
static_assert(std::is_same_v<size_t, transformation_trait_or_t<B<unsigned>, size_t>>);
// A<int>::type is defined, use A<int>::type
static_assert(std::is_same_v<int, transformation_trait_or_t<A<int>, void>>);
```

### DIFF
--- a/include/seqan3/core/metafunction/transformation_trait_or.hpp
+++ b/include/seqan3/core/metafunction/transformation_trait_or.hpp
@@ -65,9 +65,9 @@ namespace seqan3::detail
  */
 template <typename type_t, typename default_t>
 using transformation_trait_or = std::conditional_t<meta::is_trait<type_t>::value,  // check if type_t::type exists
-                                        type_t, // if yes, return type_t
-                                        // otherwise return struct{using type = default_t};
-                                        std::enable_if<true, default_t>>;
+                                                   type_t, // if yes, return type_t
+                                                   // otherwise return struct{using type = default_t};
+                                                   std::enable_if<true, default_t>>;
 
 /*!\brief Helper type of seqan3::detail::transformation_trait_or
  * \ingroup metafunction
--- a/include/seqan3/core/metafunction/transformation_trait_or.hpp
+++ b/include/seqan3/core/metafunction/transformation_trait_or.hpp
@@ -32,36 +32,48 @@
 //
 // ============================================================================
 
-#include <gtest/gtest.h>
+/*!\file
+ * \author Marcel Ehrhardt <marcel.ehrhardt AT fu-berlin.de>
+ * \brief Provides seqan3::detail::transformation_trait_or.
+ */
 
-#include <seqan3/core/metafunction/default_type.hpp>
+#pragma once
 
-using namespace seqan3;
+#include <type_traits>
 
-struct A
+#include <meta/meta.hpp>
+
+namespace seqan3::detail
 {
-    using type = int;
-};
 
-struct B;
+/*!\brief This gives a fallback type if *type_t::type* is not defined.
+ * \ingroup metafunction
+ * \tparam type_t    The type to use if *type_t::type* is defined.
+ * \tparam default_t The type to use otherwise.
+ *
+ * \details
+ *
+ * Gives *type_t* back if *T::type* is a member type, otherwise *struct{using type = default_t}*.
+ *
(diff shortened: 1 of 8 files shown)


smehringer:
Hmm, IMHO I think the name is unintuitive or not expressive enough. If I read default_type<..., ...> in the code I would first be confused that it takes two arguments and then never associate this to the ::type member.

I am bad myself with namings though, so I don't have a very good alternative. Maybe `type_member_or`..? Anyone else has input on this?

marehr:
Yeah you are right. Maybe `type_trait_or`. I think the meta library calls them type traits. 

Am 10. Oktober 2018 09:40:29 MESZ schrieb smehringer <notifications@github.com>:
>Hmm, IMHO I think the name is unintuitive or not expressive enough. If
>I read default_type<..., ...> in the code I would first be confused
>that it takes two arguments and then never associate this to the ::type
>member.
>
>I am bad myself with namings though, so I don't have a very good
>alternative. Maybe `type_member_or`..? Anyone else has input on this?
>
>-- 
>https://github.com/seqan/seqan3/pull/501#issuecomment-428470337

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

marehr:
after a short discussion with @h-2, it seems that we settle for `transformation_trait_or`, because they are named `transformation_trait` in the stl.

include/seqan3/core/metafunction/transformation_trait_or.hpp
```diff
@@ -0,0 +1,79 @@
...
+ * functionality.
+ *
+ * \par Helper types
+ *   seqan3::detail::transformation_trait_or_t as a shorthand for *seqan3::detail::transformation_trait_or::type*
+ */
+template <typename type_t, typename default_t>
+using transformation_trait_or = std::conditional_t<meta::is_trait<type_t>::value,  // check if type_t::type exists
+                                        type_t, // if yes, return type_t
+                                        // otherwise return struct{using type = default_t};
+                                        std::enable_if<true, default_t>>;
```
h-2: indent to be under `meta::...`

Is there no other way than using `enable_if`? Seems a little obscure :candle:

test/unit/core/metafunction/transformation_trait_or_test.cpp
```diff
@@ -0,0 +1,67 @@
...
+    using a_type = detail::transformation_trait_or_t<A, void>;
+    using b_transformation_trait_or = detail::transformation_trait_or_t<B, void>;
+    using c_transformation_trait_or = detail::transformation_trait_or_t<C, double>;
+    using d_transformation_trait_or = detail::transformation_trait_or<D, B>::type;
+
+    EXPECT_TRUE((std::is_same_v<a_type, int>));
+    EXPECT_TRUE((std::is_same_v<b_transformation_trait_or, void>));
+    EXPECT_TRUE((std::is_same_v<c_transformation_trait_or, double>));
+    EXPECT_TRUE((std::is_same_v<d_transformation_trait_or, B>));
+}
```
h-2: Maybe use this directly as the snippet? :thinking:

marehr: I don't like `EXPECT_TRUE` in snippets.

include/seqan3/core/metafunction/transformation_trait_or.hpp
h-2: Actually, why not? Seems useful! But maybe in another PR!

rrahn:
Mhmm, I don't see the point of calling something after its category. Like only because it is a transformation trait I don't call it so. I am not sure that something like this occurs in the STL nor somewhere in  our code? I have some ideas, but iit s not important. I am just looking into all the stuff during my absence 😁

include/seqan3/core/metafunction/transformation_trait_or.hpp
FirstLoveLife: Ah, I have been using `template< class T > struct _t {     using type = T; };` for a long time just to simpify codes. But the sample on cppreference inspires me a lot XD

PR #504, opened 2018-10-11
[FIX] Added missing return values in search module. Fixes #494

### LINKED ISSUE #494: GCC8 testing issues
When using cmake with GCC8, the debug builds are all executed fine. However, the Release mode builds have a few errors:

`sequence_file_integration_test` fails with error: `C++ exception with description "The ID field may not be empty when writing FASTA files." thrown in the test body.`

`search_test` fails with errors:
In `search_strategy_best`:
 ```cpp
  search(this->index, "AAAA"_dna4, cfg)
    Which is: { 0 }
  (hits_result_t{})
    Which is: {}
```
Further, SEGFAULTS from the following tests:
```
	 56 - unit/io/alignment_file/alignment_file_output_test (SEGFAULT)
	 62 - unit/io/sequence_file/sequence_file_output_test (SEGFAULT)
	 69 - unit/io/structure_file/structure_file_output_test (SEGFAULT)
```

### PR DESCRIPTION
**Pull request by cpockrandt**
Resolves #494

### DIFF
--- a/include/seqan3/search/algorithm/detail/search.hpp
+++ b/include/seqan3/search/algorithm/detail/search.hpp
@@ -47,8 +47,8 @@ namespace seqan3::detail
 {
 
 //!\brief \todo Docs missing
-template <typename index_t, typename query_t, typename config_t>
-inline auto _search_single(index_t const & index, query_t const & query, config_t const & cfg)
+template <typename index_t>
+inline auto search_single(index_t const & index, auto & query, auto const & cfg)
 {
     // retrieve error numbers / rates
     detail::search_params max_error{0, 0, 0, 0};
@@ -141,7 +141,7 @@ inline auto _search_single(index_t const & index, query_t const & query, config_
             // only one iterator is reported but it might contain more than one text position
             if (!internal_hits.empty())
             {
-                auto const & text_pos = internal_hits[0].lazy_locate();
+                auto text_pos = internal_hits[0].lazy_locate();
                 hits.push_back(text_pos[0]);
             }
         }
@@ -160,12 +160,8 @@ inline auto _search_single(index_t const & index, query_t const & query, config_
 }
 
 //!\brief \todo Docs missing
-template <typename index_t, typename queries_t, typename config_t>
-    requires
-        (std::ranges::RandomAccessRange<queries_t> ||
-            (std::ranges::ForwardRange<queries_t> && std::ranges::RandomAccessRange<value_type_t<queries_t>>)) &&
-        detail::is_algorithm_configuration_v<remove_cvref_t<config_t>>
-inline auto _search(index_t const & index, queries_t const & queries, config_t const & cfg)
+template <typename index_t, typename queries_t>
+inline auto search_all(index_t const & index, queries_t & queries, auto const & cfg)
 {
     // return type: for each query: a vector of text_position (or iterators) and number of errors spent
     // delegate params: text_position (or iterator), number of errors spent and query id. (TODO: or return vector)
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: include/seqan3/search/algorithm/detail/search.hpp [hunk 1]
```diff
@@ -141,7 +141,7 @@ inline auto _search_single(index_t const & index, query_t const & query, config_
             // only one iterator is reported but it might contain more than one text position
             if (!internal_hits.empty())
             {
-                auto const & text_pos = internal_hits[0].lazy_locate();
+                auto && text_pos = internal_hits[0].lazy_locate();
```
h-2: what does lazy_locate return and why are you taking it by `&&`?

Beware that `auto &&` only extends life-time of prvalues, not of xvalues.

include/seqan3/search/algorithm/detail/search.hpp
```diff
@@ -48,7 +48,7 @@ namespace seqan3::detail
 
 //!\brief \todo Docs missing
 template <typename index_t, typename query_t, typename config_t>
-inline auto _search_single(index_t const & index, query_t const & query, config_t const & cfg)
+inline auto _search_single(index_t const & index, query_t & query, config_t const & cfg)
```
h-2: why do these function names begin with an underscore? This violates SeqAn3 style guide.

include/seqan3/search/algorithm/detail/search_trivial.hpp
```diff
@@ -150,14 +134,10 @@ auto _search_trivial(auto it, auto const & query, uint64_t const query_pos, sear
...
                     if (error_left.substitution == 0 && it.last_char() == query[query_pos])
                     {
-                        if constexpr (abort_on_hit)
-                        {
-                            if (_search_trivial<abort_on_hit>(it, query, query_pos + 1, error_left, delegate))
-                                return true;
-                        }
-                        else
+                        if (_search_trivial<abort_on_hit>(it, query, query_pos + 1, error_left, delegate) &&
+                            abort_on_hit)
```
h-2: Is abort_on_hit only every evaluated at run-time now? If yes, it should be a regular parameter.

include/seqan3/search/algorithm/detail/search_trivial.hpp
```diff
@@ -84,19 +84,15 @@ namespace seqan3::detail
 {
 
 template <bool abort_on_hit>
-auto _search_trivial(auto it, auto const & query, uint64_t const query_pos, search_params error_left,
-                     auto && delegate)
+bool _search_trivial(auto it, auto & query, uint64_t const query_pos, search_params error_left, auto && delegate)
```
h-2: Use of `auto` typed parameters is valid as of the new concepts syntax, too, but the use in this file is very inconsistent (sometime `typename query_type`, sometime `auto`).

h-2: Why was `const` removed on the query?

include/seqan3/search/algorithm/search.hpp
```diff
@@ -54,7 +54,7 @@ template <typename index_t, typename queries_t, typename config_t>
             (std::ranges::ForwardRange<queries_t> && std::ranges::RandomAccessRange<value_type_t<queries_t>>)) &&
         detail::is_algorithm_configuration_v<remove_cvref_t<config_t>>
 //!\endcond
-inline auto search(index_t const & index, queries_t const & queries, config_t const & cfg)
+inline auto search(index_t const & index, queries_t && queries, config_t const & cfg)
```
h-2: here you have `queries &&`, above it is only  a single `&`

include/seqan3/search/algorithm/detail/search.hpp
h-2: just `auto`

include/seqan3/search/algorithm/detail/search_trivial.hpp
cpockrandt: I might change that when I replace this file with optimum search schemes. :+1:

h-2:
Can you squash all commits except the update of the sub-module? Then I will merge it!

cpockrandt:
Done :)

PR #505, opened 2018-10-11
simd: add first batch of simd functionality

marehr:
This commit introduces the basic structure to create, fill and manage simd types.

Like in seqan2, simd_t is still a transformation trait that delegates to the right simd implementation (now called simd backend). We still support the gcc vector extension and UME::SIMD.

The interface is more flexible than the one in seqan2, because we can now use multiple backends at the same time. The backends are designed, such that you can always include them and that they only fail if you try to access the `type` in the transformation traits.

The next step is to back port all simd algorithms, test cases and build structure (namely SeqAnSimdUtility.cmake) from seqan2 to seqan3.

--------
blocked by #501


test/snippet/simd/detail/is_ume_simd.cpp
```diff
@@ -1,7 +1,11 @@
 #include <seqan3/simd/detail/ume_simd.hpp>
 using namespace seqan3::detail;
 
+#if __has_include(<umesimd/UMESimd.h>)
+
 using int8x16_t = ume_simd<int8_t, 16>::type;
```
smehringer: If this is in the `<umesimdu/UMESimd.h>` header don't you need to include it ?

marehr: It will be auto included in `seqan3/simd/detail/ume_simd.hpp`.

h-2:
Cool that this is coming together!

I have briefly scanned over this, before I look at the details:
1. IIRC UmeSIMD currently provides no benefits over native on any of our target platforms, right? Why should we include it at all then? I mean, I think it's good to have an abstraction layer in place in case that the situation changes or another SIMD-library pops up, but right now it would seem like we are importing a lot of complexity for nothing?
2. I am unsure of where to put the SIMD code. Making a top-level module seems a little too much since non of the code is used directly by users, right? OTOH making everything `detail` also seems a little extreme... Maybe a submodule of `core`?
3. The code still has a lot of MACROs. Are all of these really necessary? Can we try to reduce it a bit?

marehr:
> Cool that this is coming together!
> 
> I have briefly scanned over this, before I look at the details:
> 
>     1. IIRC UmeSIMD currently provides no benefits over native on any of our target platforms, right? Why should we include it at all then? I mean, I think it's good to have an abstraction layer in place in case that the situation changes or another SIMD-library pops up, but right now it would seem like we are importing a lot of complexity for nothing?

That's not completely true, it works on the windows compilers (wicc and msvc).
 
> 
>     2. I am unsure of where to put the SIMD code. Making a top-level module seems a little too much since non of the code is used directly by users, right? OTOH making everything `detail` also seems a little extreme... Maybe a submodule of `core`?

It seems wrong to put it into `detail`. I would prefer the `simd` namespace. Maybe also extracting it from seqan and to provide it as a submodule, like we thought to do it with the argument parser.

> 
>     3. The code still has a lot of MACROs. Are all of these really necessary? Can we try to reduce it a bit?

When we want to support clang in the future, we don't have any other chance. I tried to open a bug report at https://bugs.llvm.org/show_bug.cgi?id=34878. With this we could omit most of the macros.

h-2:
> That's not completely true, it works on the windows compilers (wicc and msvc).

Yes, but we also don't support these platforms, yet, and won't do so until C++20 at which point the standard library might have all the functionality that we need.
I am not saying that we should never use UmeSIMD, but I don't see any benefit right now, and considering that any benefit is still far away and we don't even know if UmeSIMD will still be maintained at that point (last changes more than a year old...), I would suggest to factor the UmeSIMD stuff out and only merge builtin at this point. What do you think?

> It seems wrong to put it into detail. I would prefer the simd namespace. Maybe also extracting it from seqan and to provide it as a submodule, like we thought to do it with the argument parser.

Hm, I am not against this per se, but I think it will be more maintenance overhead. For now I would suggest putting everything into `core/simd`. I would suggest not creating a new namespace. You would still need to decide which parts of the SIMD stuff shall be part of our API and which parts go into `detail`.

> When we want to support clang in the future, we don't have any other chance. I tried to open a bug report at https://bugs.llvm.org/show_bug.cgi?id=34878. With this we could omit most of the macros.

Hm, I feel you. I will ping people on the issue. Would be good to get this fixed!

marehr:
Okay I will take out the backend ume simd. But I'll keep the structure, s.t. we can opt it in later.

I'll put into core/simd. And if I'm allowed to, I would want to wrap everything in an inline namespace.

```c++
namespace seqan3
{
inline namespace simd
{
// Simd code goes here
} 
} 
```

This will leave me the option to extract it at a later point of time. 

Am 16. Oktober 2018 13:07:57 MESZ schrieb Hannes Hauswedell <notifications@github.com>:
>> That's not completely true, it works on the windows compilers (wicc
>and msvc).
>
>Yes, but we also don't support these platforms, yet, and won't do so
>until C++20 at which point the standard library might have all the
>functionality that we need.
>I am not saying that we should never use UmeSIMD, but I don't see any
>benefit right now, and considering that any benefit is still far away
>and we don't even know if UmeSIMD will still be maintained at that
>point (last changes more than a year old...), I would suggest to factor
>the UmeSIMD stuff out and only merge builtin at this point. What do you
>think?
>
>> It seems wrong to put it into detail. I would prefer the simd
>namespace. Maybe also extracting it from seqan and to provide it as a
>submodule, like we thought to do it with the argument parser.
>
>Hm, I am not against this per se, but I think it will be more
>maintenance overhead. For now I would suggest putting everything into
>`core/simd`. I would suggest not creating a new namespace. You would
>still need to decide which parts of the SIMD stuff shall be part of our
>API and which parts go into `detail`.
>
>> When we want to support clang in the future, we don't have any other
>chance. I tried to open a bug report at
>https://bugs.llvm.org/show_bug.cgi?id=34878. With this we could omit
>most of the macros.
>
>Hm, I feel you. I will ping people on the issue. Would be good to get
>this fixed!
>
>-- 
>https://github.com/seqan/seqan3/pull/505#issuecomment-430196416

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

h-2:
sounds like a good plan :+1:

include/seqan3/core/simd/detail/builtin_simd.hpp
```diff
@@ -0,0 +1,219 @@
...
+
+/*!\brief Helper struct for seqan3::detail::is_builtin_simd
+ * \ingroup simd
+ * \sa seqan3::detail::is_builtin_simd
+ */
+template <typename builtin_simd_t>
+//!\cond
+    requires requires(builtin_simd_t simd){ {simd[0]} }
+//!\endcond
+struct _is_builtin_simd<builtin_simd_t>
```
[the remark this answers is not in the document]
h-2: Can you just but the condition from below into the requires statement and have it inherit `std::true_type`?

include/seqan3/core/simd/detail/builtin_simd.hpp
```diff
@@ -0,0 +1,219 @@
...
+ * \sa seqan3::simd::simd_traits for more information
+ */
+template <typename builtin_simd_t>
+// \cond
+    requires detail::is_builtin_simd<builtin_simd_t>::value
+// \endcond
+struct simd_traits<builtin_simd_t>
+{
+    //!\copydoc seqan3::simd::simd_traits::scalar_type
+    using scalar_type = typename detail::_is_builtin_simd<builtin_simd_t>::scalar_type;
```
h-2: wait, this is really weird that something that is called `is_...` which implies bool now provides other member types like a type_traits object :thinking:

include/seqan3/core/simd/detail/builtin_simd.hpp
```diff
@@ -0,0 +1,219 @@
...
+    return min_viable_uint_v<32u>;
+#elif defined(__AVX2__)
+    return min_viable_uint_v<32u>;
+#elif defined(__SSE4_1__) && defined(__SSE4_2__)
+    return min_viable_uint_v<16u>;
+#else
+    // static_assert(false, "Your code uses simd, but you didn't provide flags "
+    // "(like `-march=native`, `-msse4`, `-mavx2`, etc) to your compiler "
+    // "to build special binaries that make use of the special vector/simd "
+    // "registers.");
```
h-2: why is this commented out?

include/seqan3/core/simd/detail/default_simd_max_length.hpp
```diff
@@ -0,0 +1,60 @@
...
+ * \include test/snippet/core/simd/detail/default_simd_max_length.cpp
+ *
+ * This value is influenced by compiler flags like `-march=native`, `-msse4`,
+ * `-mavx2`, etc and sets this value accordingly.
+ * \sa https://gcc.gnu.org/onlinedocs/gcc-7.2.0/gcc/x86-Options.html#x86-Options
+ * for simd instruction sets and their flags.
+ */
+template <template<typename, size_t> typename simd_backend_t>
+constexpr auto default_simd_max_length = 0u;
+} // namespace seqan3::detail
```
h-2: Do the previous three functions need to be in three separate files? They are all detail anyway...

include/seqan3/core/simd/detail/default_simd_max_length.hpp
```diff
@@ -0,0 +1,60 @@
...
+ * \tparam simd_backend_t The name of the simd backend.
+ *
+ * \include test/snippet/core/simd/detail/default_simd_max_length.cpp
+ *
+ * This value is influenced by compiler flags like `-march=native`, `-msse4`,
+ * `-mavx2`, etc and sets this value accordingly.
+ * \sa https://gcc.gnu.org/onlinedocs/gcc-7.2.0/gcc/x86-Options.html#x86-Options
+ * for simd instruction sets and their flags.
+ */
+template <template<typename, size_t> typename simd_backend_t>
```
h-2: space after template before `<typename` :nail_care:

include/seqan3/core/simd/detail/ume_simd.hpp
```diff
@@ -0,0 +1,172 @@
+// ============================================================================
```
h-2: This will be removed, right?

include/seqan3/core/simd/simd_algorithm.hpp
```diff
@@ -0,0 +1,111 @@
...
+
+#include <seqan3/core/simd/concept.hpp>
+#include <seqan3/core/simd/simd_traits.hpp>
+
+#include <utility>
+
+namespace seqan3::detail::simd
+{
+
+using namespace seqan3::simd;
```
h-2: No `using namespace` in the library please. It is a drawback of further nesting, that you cannot implicitly use the stuff from two levels up, but I recommend just using seqan3::detail here, like you did in other places.

include/seqan3/core/simd/simd_algorithm.hpp
```diff
@@ -0,0 +1,111 @@
...
+
+//!\brief Helper function for seqan3::simd::iota.
+//!\ingroup simd
+template <simd_concept simd_t, typename scalar_t, scalar_t... I>
+constexpr simd_t iota_impl(scalar_t const offset, std::integer_sequence<scalar_t, I...>)
+{
+    return simd_t{(offset + I)...};
+}
+
+} // namespace seqan3::detail
```
h-2: The closing doc even says that you do :-)

h-2:
Tests not reviewed, yet.

include/seqan3/core/simd/all.hpp
```diff
@@ -0,0 +1,66 @@
...
+#include <seqan3/core/simd/concept.hpp>
+#include <seqan3/core/simd/simd_algorithm.hpp>
+#include <seqan3/core/simd/simd_debug_stream.hpp>
+#include <seqan3/core/simd/simd_traits.hpp>
+#include <seqan3/core/simd/simd.hpp>
+
+/*!\namespace seqan3::simd
+ * \brief The SeqAn3 namespace for simd data types, algorithms and meta functions.
+ * \sa https://en.wikipedia.org/wiki/SIMD What is SIMD conceptually?
+ * \sa https://en.wikipedia.org/wiki/Streaming_SIMD_Extensions Which SIMD architectures exist?
```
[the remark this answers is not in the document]
marehr: They are already in the module description

include/seqan3/core/simd/concept.hpp
```diff
@@ -0,0 +1,99 @@
...
+    { a[0] };
+    requires std::is_convertible_v<decltype(a[0]), typename simd_traits<simd_t>::scalar_type>;
+
+    // require comparison operators
+    { a == b } -> typename simd_traits<simd_t>::mask_type;
+    { a != b } -> typename simd_traits<simd_t>::mask_type;
+    { a <  b } -> typename simd_traits<simd_t>::mask_type;
+    { a >  b } -> typename simd_traits<simd_t>::mask_type;
+    { a <= b } -> typename simd_traits<simd_t>::mask_type;
+    { a >= b } -> typename simd_traits<simd_t>::mask_type;
```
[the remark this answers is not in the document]
marehr: As discussed is this not possible for `simd` types, because they return a collection of bools

include/seqan3/core/simd/concept.hpp
```diff
@@ -0,0 +1,99 @@
...
+    typename simd_traits<simd_t>::mask_type;
+    typename simd_traits<simd_t>::swizzle_type;
+
+    // require that static member variables are defined
+    { simd_traits<simd_t>::length }
+    { simd_traits<simd_t>::max_length }
+
+    // assume array access that returns a scalar_type type
+    { a[0] };
+    requires std::is_convertible_v<decltype(a[0]), typename simd_traits<simd_t>::scalar_type>;
```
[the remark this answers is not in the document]
marehr: Yeah, I wanted to maybe add this later. We can leave it as a follow up issue.

marehr:
@h-2 I addressed your points

include/seqan3/core/simd/detail/default_simd_max_length.hpp
h-2: why again is this not possible? It clearly doesn't depend on anything...

include/seqan3/core/simd/simd_algorithm.hpp
```diff
@@ -0,0 +1,109 @@
...
+ * \brief Contains algorithms to modify seqan3::simd::simd_type.
+ * \author Marcel Ehrhardt <marcel.ehrhardt AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <seqan3/core/simd/concept.hpp>
+#include <seqan3/core/simd/simd_traits.hpp>
+
+#include <utility>
```
h-2: include order :nail_care:

include/seqan3/core/simd/simd_debug_stream.hpp
```diff
@@ -0,0 +1,69 @@
...
+#include <seqan3/core/simd/simd_traits.hpp>
+
+namespace seqan3
+{
+
+/*!\brief Overload for debug_stream for simd types.
+ * \ingroup simd
+ */
+template <typename simd_t>
+    requires simd::simd_concept<remove_cvref_t<simd_t>>
```
h-2: don't you need to guard this or put it below the signature for doxygen?

include/seqan3/core/simd/simd_debug_stream.hpp
```diff
@@ -0,0 +1,69 @@
...
+inline debug_stream_type & operator<<(debug_stream_type & s, simd_t && simd)
+{
+    using simd_type = remove_cvref_t<simd_t>;
+
+    s << '(';
```
h-2: maybe use `[` instead of `(` like we do for ranges?

PS: as soon as simd types model ranges we can remove this!

include/seqan3/core/simd/simd_traits.hpp
```diff
@@ -0,0 +1,95 @@
...
+     * underlying type would be *[u]int8_t* (is not defined if *simd_t* does not
+     * satisfies *seqan3::simd::simd_concept*)
+     */
+    static constexpr auto max_length = IMPLEMENTATION_DEFINED;
+    /*!\brief The type returned by comparison operators (is not defined if
+     * *simd_t* does not satisfies *seqan3::simd::simd_concept*)
+     */
+    using mask_type = IMPLEMENTATION_DEFINED;
+    /*!\brief The type used to define how to swizzle a simd vector (is not
+     * defined if *simd_t* does not satisfies *seqan3::simd::simd_concept*)
```
h-2: does not satisf**y** in all previous documentation blocks!

test/include/seqan3/test/simd_utility.hpp
```diff
@@ -0,0 +1,70 @@
...
+    auto [left, right] = std::make_tuple(__VA_ARGS__); \
+    static_assert(seqan3::simd::simd_concept<decltype(left)>, "The left argument of SIMD_EQ is not a simd_type"); \
+    static_assert(seqan3::simd::simd_concept<decltype(right)>, "The right argument of SIMD_EQ is not a simd_type"); \
+    static_assert(std::is_same_v<decltype(left), decltype(right)>, "The left and right argument of SIMD_EQ don't have the same type."); \
+    using _simd_traits_t = seqan3::simd::simd_traits<decltype(left)>; \
+    std::vector<typename _simd_traits_t::scalar_type> left_simd(_simd_traits_t::length), right_simd(_simd_traits_t::length); \
+    for (size_t i = 0; i < _simd_traits_t::length; ++i) \
+    { left_simd[i] = left[i]; right_simd[i] = right[i]; } \
+    EXPECT_EQ(left_simd, right_simd); \
+} while (false)
```
h-2: most ugly code award of 2018 :japanese_ogre: 
:laughing:

test/snippet/core/simd/fill.cpp
```diff
@@ -0,0 +1,13 @@
+#include <seqan3/core/simd/all.hpp>
+
+using namespace seqan3;
+using namespace seqan3::simd;
```
h-2: You don't need this, right?

test/unit/core/simd/detail/builtin_simd_test.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+#include <seqan3/core/simd/concept.hpp>
+#include <seqan3/core/simd/detail/builtin_simd.hpp>
+
+#include <iostream>
+#include <type_traits>
+
+using namespace seqan3;
+using namespace seqan3::detail;
+using namespace seqan3::simd;
```
h-2: I would prefer not to do `using namespace seqan3::detail` because it might hide errors that occur precisely because one forgot to right `detail::` in the library, too :wink:

test/include/seqan3/test/simd_utility.hpp
marehr: Thank you I'm honoured :clap: :tada: :1st_place_medal:

include/seqan3/core/simd/simd_debug_stream.hpp
marehr: good catch; I used the short notation before

marehr:
> One last general questions: If somebody builds without SIMD extensions, what is the behaviour?

I added this to https://github.com/seqan/seqan3/issues/531; I think I want scalar emulation, but the code should warn the user that he wants use simd optimized code, but does not provide the correct compiler flag.

> And should be we add placeholders, similar to an std::array, but with overloaded operators for +... ?

What placeholders do you mean?

marehr:
Slight update; due to the removal of the macros, we get scalar emulation for free. So the question remains whether we warn the user that he uses SIMD code, but does not provide the appropriate flags.

h-2:
!rebuild

test/unit/core/simd/simd_debug_stream_test.cpp
```diff
@@ -40,15 +40,14 @@
 #include <iostream>
 #include <sstream>
 
-using namespace seqan3;
 using namespace seqan3::simd;
```
h-2: no, I meant removing `seqan3::simd`, but keeping `seqan3` :see_no_evil:

marehr: I think I want to leave it like this, because IF we rip this as an extra module from seqan3. I can easily search and replace :)

h-2: but if you include seqan3 and not `seqan3::simd` you still don't need to prefix things with `simd::`, because it's an inline namespace. 

I am arguing that the tests should be similar to what people will use in their apps, and people will likely include `seqan3` but I see no-one including `seqan3::simd` and not `seqan3` :thinking:

marehr: I reverted it

h-2:
!rebuild

marehr:
hmm jenkins is still down

marehr:
even the docs build now :hugs: :yellow_heart:

PR #512, opened 2018-10-16
[INTERNAL] make the metafunctions not redirect to range-v3 metafunctions

h-2:
minor changes to not make these depend on range-v3 counterparts. They do however depend on ranges::begin, ranges::end, ranges::size and ranges::iter_move which should be available in our `std` module.

### DIFF
--- a/include/seqan3/core/metafunction/iterator.hpp
+++ b/include/seqan3/core/metafunction/iterator.hpp
@@ -39,6 +39,7 @@
 
 #pragma once
 
+#include <iterator>
 #include <type_traits>
 
 #include <range/v3/utility/iterator_traits.hpp>
@@ -65,7 +66,7 @@ template <std::InputIterator it_t>
 struct value_type<it_t>
 {
     //!\brief Return the member type as return type.
-    using type = ranges::value_type_t<it_t>;
+    using type = typename std::iterator_traits<std::remove_reference_t<it_t>>::value_type;
 };
 
 // see specialisation for ranges in core/metafunction/range.hpp
@@ -81,7 +82,7 @@ template <std::InputIterator it_t>
 struct reference<it_t>
 {
     //!\brief Return the member type as return type.
-    using type = ranges::reference_t<it_t>;
+    using type = typename std::iterator_traits<std::remove_reference_t<it_t>>::reference;
 };
 
 // see specialisation for ranges in core/metafunction/range.hpp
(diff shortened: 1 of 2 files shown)


include/seqan3/core/metafunction/iterator.hpp
```diff
@@ -65,7 +66,7 @@ template <std::InputIterator it_t>
 struct value_type<it_t>
 {
     //!\brief Return the member type as return type.
-    using type = ranges::value_type_t<it_t>;
+    using type = typename std::iterator_traits<std::remove_reference_t<it_t>>::value_type;
```
smehringer: As I understand std::iterator_traits is specialized with an Iterator type but you are passing a range type?
Shouldn't it be rather something like 
```suggestion
    using type = typename std::iterator_traits<std::remove_reference_t<decltype(it_t{}.begin())>>::value_type;
```

include/seqan3/core/metafunction/iterator.hpp
```diff
@@ -97,7 +98,7 @@ template <std::InputIterator it_t>
 struct rvalue_reference<it_t>
 {
     //!\brief Return the member type as return type.
-    using type = ranges::rvalue_reference_t<it_t>;
+    using type = decltype(ranges::iter_move(std::declval<it_t &>()));
```
smehringer: This still depends on the ranges library? Why is this any better?

include/seqan3/core/metafunction/range.hpp
```diff
@@ -71,7 +71,7 @@ namespace seqan3
  * \tparam t The type to operate on.
  */
 template <std::ranges::Range rng_t>
-using iterator_t = ranges::iterator_t<rng_t>;
+using iterator_t = decltype(ranges::begin(std::declval<rng_t &>()));
```
joshuak94: Does this have to be `std::ranges::begin` or can it just be `ranges::begin`?

include/seqan3/core/metafunction/iterator.hpp
h-2: `rvalue_reference_t` won't be part of C++20, `iter_move` will be.

h-2:
> In general, why don't we use std::begin?

Please see my comments in #239

include/seqan3/core/metafunction/range.hpp
h-2: After your PR is merged it should be `std::ranges::begin`, but right now that wouldn't work.

include/seqan3/core/metafunction/iterator.hpp
h-2: and pending the changes by @joshuak94 it would become `std::ranges::iter_move`

PR #513, opened 2018-10-16
factor out the use of ranges::back_insert_iterator

h-2:
as discussed in #498

### DIFF
--- a/include/seqan3/std/iterator
+++ b/include/seqan3/std/iterator
@@ -39,9 +39,9 @@
 
 #pragma once
 
-#if 0 //TODO add check for C++20 iterator library
 #include <iterator>
-#else // implement via range-v3
+
+#if __cplusplus <= 201703L // TODO feature_test_macro if available
 
 #include <range/v3/utility/iterator.hpp>
 
@@ -57,6 +57,33 @@ namespace std
 
 //!\addtogroup iterator
 //!\{
+
+// ------------------------------------------------------------------
+// auxiliary functions
+// ------------------------------------------------------------------
+
+/*!\brief Create a std::back_insert_iterator for the argument.
+ * \tparam container_t  Type of the parameter; must have a `push_back()` member function.
+ * \param container     The container on which to create the iterator.
+ * \returns The respective back insert iterator.
+ *
+ * \details
+ *
+ * This function delegates to ranges::back_inserter from range-v3; it is more constrained than a possibly
+ * outdated one from the standard library.
+ */
+template <typename container_t>
+//!\cond
+    requires requires (container_t & v) { { v.push_back(*v.begin()) }; }
+//!\endcond
+constexpr auto back_inserter(container_t & container)
+{
+    return ::ranges::back_inserter(container);
+}
+
+// ------------------------------------------------------------------
+// Concepts
+// ------------------------------------------------------------------
 
 /*!\interface std::Readable <>
  * \brief The concept Readable is satisfied by types that are readable by applying operator*, such as pointers, smart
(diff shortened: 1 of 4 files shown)


joshuak94:
>     1. why do we want to factor out all the ranges stuff again?
> 
>     2. Why did you replace the `back_insert_iterator` by the `back_inserter` helper?
>        I liked the argument that you should directly use the `back_insert_iterator` to show that an object is created. So why not add `std::back_insert_iterator` to auxiliary that redirects to `ranges::back_insert_iterator`?

I think we want to take out the ranges stuff so that when C++20 comes out, we don't need to go through it and replace it then?

Also I don't really understand the difference between back_inserter and back_inserter_iterator. `back_inserter` is essentially creating the iterator, isn't it?

h-2:
> why do we want to factor out all the ranges stuff again?

Please see #239 

> Also I don't really understand the difference between back_inserter and back_inserter_iterator. back_inserter is essentially creating the iterator, isn't it?

`std::back_inserter` is a convenience function that returns `std::back_insert_iterator`, but doesn't require that you specify the types yourself. Now that we have type deduction guides it is redundant, but for our case it is helpful (I will explain below). An example with similar behaviour, but maybe easier to understand is std::tuple:

```cpp
std::string s{};
int i{};

// valid pre C++17 (and still)
std::tuple<std::string, int> tup{s, i};

// valid pre C++17, the same thing but lazy
auto tup = std::make_tuple(s, i);

// valid only since C++17
std::tuple tup{s, i};
```
(All three examples result in the same object, but the last is only valid since C++17.

> Why did you replace the back_insert_iterator by the back_inserter helper?
I liked the argument that you should directly use the back_insert_iterator to show that an object is created. So why not add std::back_insert_iterator to auxiliary that redirects to ranges::back_insert_iterator?

As I explained in #239 the whole matter of "redirecting" is quite difficult for types and gets more difficult if the name already exists. You need to create a template specialisation of `std::back_insert_iterator` that appears specialised, but acutally isn't; that also inherits from `ranges::back_insert_iterator` and then you need to explicitly inherit all the constructors and all member types, because they are not inherited by default.
For ostreambuf_iterator we either have to do this, or make a small function. I currently lean towards the latter although it is less consistent.

marehr:
@h-2 If I understand you correctly, then

> To import types we can use using, but for type templates we loose type deduction guides, because they just don't work on typedefs. An alternative is to declare a "wrapper-type" like with the function, but then we need to add all sorts of cruft, because templates inheriting other templates don't inherit constructors and member types.

from https://github.com/seqan/seqan3/issues/239#issuecomment-430612391 is your main reason to use the `back_inserter` instead of the actual type.

h-2:
@marehr: No, you cannot declare an alias in this case even if you wanted to, because the name `std::back_insert_iterator` already exists. An alias templates cannot be specialised. In the terminology of #239 this case is 1b and not 1a.

marehr:
Okay thank you for clarifying. 

Am 17. Oktober 2018 23:46:05 MESZ schrieb Hannes Hauswedell <notifications@github.com>:

include/seqan3/std/iterator
```diff
@@ -217,4 +244,4 @@ concept RandomAccessIterator = BidirectionalIterator<i> &&
 
 }  // namespace std
 
-#endif
+#endif // C++17
```
smehringer:
```suggestion
#endif // < C++17
```

h-2: no, it's not "< C++17", it's "<= C++17" or "< C++20" which is "== C++17" because the C++17 is the minimum version SeqAn supports :wink

h-2:
> The TODO of the feature macro should maybe created as an issue so we do not forget about this?

I don't think I will forget it as I will go over this all again once Ranges is actually merged in San Diego in November (hopefully).

PR #515, opened 2018-10-17
Temporal use of boost::filesystem

qPCR4vir:
On top of https://github.com/seqan/seqan3/pull/514
Temporal workaround for GCC 8.1 not implementing filesystem for windows and a "bug" on gcc 8.1 fstream not taking a wchar_t filename in windows ([there is a commit for this in master branch of gcc - it will be fixed in future distributions](https://github.com/gcc-mirror/gcc/commit/d4fd5c4964cf1862c19c0caa876fa43a677bc04a))
In windows it need [minor changes in sdsl-lite](https://github.com/xxsds/sdsl-lite/pull/55) and also [a fix on one test](https://github.com/qPCR4vir/seqan3/commit/c210df206a0aa46e283719bbd8557798537c967b)

I'm in Windows 7pro 64b, compiling with [MinGW - gcc 8.1 - boost 1.67 - ZLIB-1.2.11 - BZip2-1.0.6](https://nuwen.net/mingw.html), using cmake 3.12.2 and CLion 2018.2.2. In windows it pass all the test you have in travis.


h-2:
In general I would be very happy to support MinGW as a target platform, but I am reluctant to add lots of "temporary changes" that end up sticking around for a lot longer than one always anticipates. 

According to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78870 everything works with MinGW 9 snapshots, can you confirm that? Maybe one could add some "pretty please" posts to that issue to motivate Jonathan to backport those changes to the gcc-8-branch!

If they don't backport, I would prefer right now that we just track MinGW-9 as target since SeqAn3 is not even released, yet, and adding workarounds that won't be necessary for the release means extra work for all of us (we need to review and maintain extra complexity, later again remove, you need to revise the PR et cetera).

qPCR4vir:
Yes, completely agree with you. These changes are mean to be in a separate temporal branch, not merged to master or any other branch (that is way I did two separate PR - I have 2 branch also). 
That gcc fix was the first I wanted to use, but I'm not sure I can compile gcc self (or the std library - [you need to build GCC with --enable-libstdcxx-filesystem-ts](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78870#c14) ). I will try.

h-2:
resolved by new MinGW version

PR #516, opened 2018-10-18
[FIX] ambigious constructors of specialised inherited_iterator_base

h-2:
Fixes a bug that prevented multiple instances of `view::take`, `view::take_until` and `view::take_line` (and their `_or_throw` versions to be chained.

The problem was found by @joshuak94 but I don't think there was an issue.

### DIFF
--- a/include/seqan3/range/view/take.hpp
+++ b/include/seqan3/range/view/take.hpp
@@ -111,7 +111,7 @@ private:
         ~iterator_type() = default;
 
         //!\brief Constructor that delegates to the CRTP layer.
-        iterator_type(base_base_t it) :
+        iterator_type(base_base_t const & it) :
             base_t{it}
         {}
 
--- a/include/seqan3/range/view/persist.hpp
+++ b/include/seqan3/range/view/persist.hpp
@@ -46,7 +46,6 @@
 #include <seqan3/io/exception.hpp>
 #include <seqan3/range/concept.hpp>
 #include <seqan3/range/container/concept.hpp>
-#include <seqan3/range/detail/inherited_iterator_base.hpp>
 #include <seqan3/range/view/detail.hpp>
 #include <seqan3/std/concepts>
 #include <seqan3/std/view/view_all.hpp>
--- a/include/seqan3/range/view/take_line.hpp
+++ b/include/seqan3/range/view/take_line.hpp
@@ -105,7 +105,7 @@ private:
         ~iterator_type() = default;
 
         //!\brief Constructor that delegates to the CRTP layer.
-        iterator_type(base_base_t it) : base_t{it} {}
+        iterator_type(base_base_t const & it) : base_t{it} {}
         //!\}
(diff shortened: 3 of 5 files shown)


include/seqan3/range/view/take.hpp
```diff
@@ -111,7 +111,7 @@ class view_take : public ranges::view_interface<view_take<urng_t, exactly, or_th
         ~iterator_type() = default;
 
         //!\brief Constructor that delegates to the CRTP layer.
-        iterator_type(base_base_t it) :
+        iterator_type(base_base_t const & it) :
```
smehringer: This will also bind rvalues correct?

h-2: Both bind everything, the old line by copy, the new one by `const &`. In situations where base_base_t is the same as iterator_type this is ambiguous with the explicitly defaulted copy-constructor (a few lines further up). But if we define it with `const &` the signature will become identical with the explicitly defaulted copy-constructor and it won't be defaulted anymore :smiley:

smehringer: Neat. But should we default the copy-constructor when it will never be used? This might cause confusion if you actually change the behaviour of this constructor you might still expect to default construct if you construct from iterator_type?

h-2: > Neat. But should we default the copy-constructor when it will never be used?

Well in 99.9% of the cases it is used. The types are only the same if you have two instances of `view::take` immediately after each other in a pipe.

> This might cause confusion if you actually change the behaviour of this constructor you might still expect to default construct if you construct from iterator_type?

What do you mean here exactly?

h-2: > The types are only the same if you have two instances of view::take immediately after each other in a pipe.

Which also doesn't make sense btw.

marehr: > But if we define it with `const &` the signature will become identical with the explicitly defaulted copy-constructor and it won't be defaulted anymore

So we use a property of defaulted constructors? Normally, when two functions have the same signature it will give you a compilation error. Can you point to where this behaviour is specified?

h-2: My previous analysis was wrong. `base_base_t` is never identical to iterator_type, but through the weird use of CRTP it might appear as if one inherits from the other, or at least from the same base class. Since we pass it it by reference the compiler doesn't know if to
1) cast to base class and take by `const &`; or
2) copy

This example illustrates the behaviour:
```cpp

struct A
{};

struct B : A
{};

bool foobar(A const &)
{
    return true;
}

bool foobar(B)
{
    return false;
}

int main()
{
    B const b;
    foobar(b);
}
```
The first overload is selected in this case. I am not 100% sure why the above case does not produce an error or warning while our case does, but I assume, because by using the 2nd parameters of the CRTP we are not using the exact type, or some other CRTP weirdness.

PR #518, opened 2018-10-24
[FIX] Piping const configuration object into empty configuration object

### LINKED ISSUE #456: Configuration: cannot pipe configuration object to empty configuration
This code snippet doesn't work and gives a compilation error:

```
detail::configuration cfg;
auto cfg2 = cfg | mode(all);
```

Currently we have to do a work-around:

```
if constexpr (std::Same<remove_cvref_t<decltype(cfg)>, detail::configuration<>>)
{
    detail::configuration const cfg2 = mode(all);
    return detail::_search(index, queries, cfg2);
}
else
{
   detail::configuration const cfg2 = cfg | mode(all);
   return detail::_search(index, queries, cfg2);
}
```

Maybe we can overload the pipe operator for empty configurations as well?

cpockrandt:
Fixes  #456

### DIFF
--- a/include/seqan3/core/algorithm/configuration.hpp
+++ b/include/seqan3/core/algorithm/configuration.hpp
@@ -318,7 +318,7 @@ configuration(cfg_fn_t &&) ->
  */
 template <typename target_t, typename query_t>
 //!\cond
-    requires !is_algorithm_configuration_v<std::remove_reference_t<target_t>>
+    requires !is_algorithm_configuration_v<remove_cvref_t<target_t>>
 //!\endcond
 struct is_configuration_combinable_with
 {
@@ -335,7 +335,7 @@ struct is_configuration_combinable_with
 //!\ingroup algorithm
 template <typename target_t, typename query_t>
 //!\cond
-    requires !is_algorithm_configuration_v<std::remove_reference_t<target_t>>
+    requires !is_algorithm_configuration_v<remove_cvref_t<target_t>>
 //!\endcond
 inline constexpr bool is_configuration_combinable_with_v = is_configuration_combinable_with<target_t, query_t>::value;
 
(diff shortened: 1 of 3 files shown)


include/seqan3/core/algorithm/configuration.hpp
```diff
@@ -459,7 +459,7 @@ template <typename lhs_fn_t,
 constexpr auto operator|(lhs_fn_t && lhs_fn,
                          rhs_fn_t && rhs_fn)
 //!\cond
-    requires is_configuration_combinable_with_v<lhs_fn_t, rhs_fn_t>
+    requires is_configuration_combinable_with_v<remove_cvref_t<lhs_fn_t>, remove_cvref_t<rhs_fn_t>>
```
h-2: You are also removing ref, although `is_configuration_combinable_with_v<` already removes ref. Maybe that would be a good place to also remove const instead of here? Also: are you sure all types of const configurations are combinable if their non-const counter-parts are? Maybe @rrahn explicitly only wanted to remove ref when comparing?

include/seqan3/search/algorithm/search.hpp
```diff
@@ -87,23 +87,11 @@ inline auto search(index_t const & index, queries_t && queries, config_t const &
     }
     else
     {
-        // TODO: overload pipe operator for empty config object
```
smehringer: Can you point out the difference between if else blocks that you removed? I don't see the diffetrence and the connection to piping an empty config object.

include/seqan3/core/algorithm/configuration.hpp
cpockrandt: As far as I know there should not be any difference whether a configuration is const or not.

PR #519, opened 2018-10-24
embl format

### LINKED ISSUE #506: The sequence file format emlb
Part of #310

[Here](https://www.genomatix.de/online_help/help/sequence_formats.html) is a nice overview over the formats including the **embl** format.

As a first step, ignore the header: 

* For reading just read the `ID` field in the beginning, skip until `SQ` tag and read the sequence from there.
Quality information like fastq are not provided so you can copy from fasta format.

* For writing for now write out `ID` and `SQ` seperated by one empty line (`XX` tag). Research in the literature has to be done about which fields in the header are required.

After that, we may include HEADER object in the sequence file format as well.. but this has to be discussed with @h-2 . The header would be also needed for the genbank format.

MitraDarja:
Resolves #506

### DIFF
--- a/include/seqan3/io/sequence_file/format_embl.hpp
+++ b/include/seqan3/io/sequence_file/format_embl.hpp
@@ -1,36 +1,9 @@
-// ============================================================================
-//                 SeqAn - The Library for Sequence Analysis
-// ============================================================================
-//
-// Copyright (c) 2006-2018, Knut Reinert & Freie Universitaet Berlin
-// Copyright (c) 2016-2018, Knut Reinert & MPI Molekulare Genetik
-// All rights reserved.
-//
-// Redistribution and use in source and binary forms, with or without
-// modification, are permitted provided that the following conditions are met:
-//
-//     * Redistributions of source code must retain the above copyright
-//       notice, this list of conditions and the following disclaimer.
-//     * Redistributions in binary form must reproduce the above copyright
-//       notice, this list of conditions and the following disclaimer in the
-//       documentation and/or other materials provided with the distribution.
-//     * Neither the name of Knut Reinert or the FU Berlin nor the names of
-//       its contributors may be used to endorse or promote products derived
-//       from this software without specific prior written permission.
-//
-// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
-// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
-// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
-// ARE DISCLAIMED. IN NO EVENT SHALL KNUT REINERT OR THE FU BERLIN BE LIABLE
-// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
-// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
-// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
-// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
-// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
-// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
-// DAMAGE.
-//
-// ============================================================================
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
+// -----------------------------------------------------------------------------------------------------
 
 /*!\file
  * \brief Provides the seqan3::sequence_file_format_embl class.
@@ -50,6 +23,7 @@
 #include <range/v3/view/drop_while.hpp>
 #include <range/v3/view/join.hpp>
 #include <range/v3/view/remove_if.hpp>
+#include <range/v3/view/repeat_n.hpp>
 #include <range/v3/view/take_while.hpp>
 
 #include <seqan3/alphabet/nucleotide/dna5.hpp>
@@ -71,6 +45,8 @@
 #include <seqan3/std/view/subrange.hpp>
 #include <seqan3/std/view/transform.hpp>
 
+#include <seqan3/std/charconv>
+
 namespace seqan3
 {
 /*!\brief       The EMBL format.
@@ -91,8 +67,8 @@ namespace seqan3
  *
  * ### Implementation notes
  *
- * When reading the ID-line, the ID is read until it encounters a ';'. Unless, the option truncate_ids is set to true,
- * then the id is read until it either sees a blank, ';' or a new line.
+ * When reading the ID-line, the ID is read until the stream encounters a ';'. Unless, the option truncate_ids is set to
+ * true, then the id is read until it either sees a blank, ';' or a new line.
  *
  * When writing the ID-line, the sequence length is appended.
  *
@@ -143,17 +119,17 @@ public:
                              std::istreambuf_iterator<char>{}};
         auto stream_it = ranges::begin(stream_view);
 
-        std::string identifier;
+        std::string idbuffer;
         ranges::copy(stream_view | view::take_until_or_throw(is_cntrl || is_blank),
-                     std::back_inserter(identifier));
-        if (identifier != "ID")
+                     std::back_inserter(idbuffer));
+        if (idbuffer != "ID")
             throw parse_error{"An entry has to start with the code word ID."};
 
         if constexpr (!detail::decays_to_ignore_v<id_type>)
         {
             if (options.complete_header)
             {
-                ranges::copy(identifier | view::char_to<value_type_t<id_type>>, std::back_inserter(id));
+                ranges::copy(idbuffer | view::char_to<value_type_t<id_type>>, std::back_inserter(id));
                 do
                 {
                     ranges::copy(stream_view | view::take_until_or_throw(is_char<'S'>)
@@ -163,7 +139,7 @@ public:
                     ++stream_it;
                 } while (*stream_it != 'Q');
                 id.pop_back(); // remove 'S' from id
-                identifier = "SQ";
+                idbuffer = "SQ";
             }
             else
             {
@@ -184,19 +160,18 @@ public:
                                  std::back_inserter(id));
                 }
             }
-
         }
 
         // Jump to sequence
-        detail::consume(stream_view | view::take_line_or_throw);
-        while (identifier != "SQ")
+        if (idbuffer !="SQ")
         {
-            identifier.clear();
-            ranges::copy(stream_view | view::take_until_or_throw(is_cntrl || is_blank),
-                         std::back_inserter(identifier));
-            detail::consume(stream_view | view::take_until(is_cntrl));
-            ++stream_it;
+            do
+            {
+                detail::consume(stream_view | view::take_until_or_throw(is_char<'S'>));
+                ++stream_it;
+            } while (*stream_it != 'Q');
         }
+        detail::consume(stream_view | view::take_line_or_throw); //Consume line with infos to sequence
 
         // Sequence
         auto constexpr is_end = is_char<'/'> ;
@@ -225,8 +200,9 @@ public:
         {
             detail::consume(stream_view | view::take_until(is_end));
         }
-        //Jump over //
-        detail::consume(stream_view | view::take_until(is_cntrl));
+        //Jump over // and cntrl
+        ++stream_it;
+        ++stream_it;
         ++stream_it;
 
         // make sure "buffer at end" implies "stream at end"
@@ -250,7 +226,8 @@ public:
     {
 
         ranges::ostreambuf_iterator stream_it{stream};
-        size_t sequence_size = 0;
+        [[maybe_unused]] size_t sequence_size = 0;
+        [[maybe_unused]] char buffer[50];
         if constexpr (!detail::decays_to_ignore_v<seq_type>)
             sequence_size = ranges::size(sequence);
 
@@ -270,13 +247,11 @@ public:
             }
             else
             {
-                stream_it = 'I';
-                stream_it = 'D';
-                stream_it = ' ';
+                ranges::copy(std::string_view{"ID "}, stream_it);
                 ranges::copy(id, stream_it);
-                stream_it = ';';
-                stream_it = ' ';
-                ranges::copy(std::to_string(sequence_size), stream_it);
+                ranges::copy(std::string_view{"; "}, stream_it);
+                auto res = std::to_chars(&buffer[0], &buffer[0] + sizeof(buffer), sequence_size);
+                std::copy(&buffer[0], res.ptr, stream_it);
                 ranges::copy(std::string_view{" BP.\n"}, stream_it);
             }
 
@@ -285,8 +260,7 @@ public:
         // Sequence
         if constexpr (detail::decays_to_ignore_v<seq_type>) // sequence
         {
-            throw std::logic_error{"The SEQ and SEQ_QUAL fields may not both be set to ignore when writing "
-            "embl files."};
+            throw std::logic_error{"The SEQ field may not be set to ignore when writing embl files."};
         }
         else
         {
@@ -294,23 +268,23 @@ public:
                 throw std::runtime_error{"The SEQ field may not be empty when writing embl files."};
 
             ranges::copy(std::string_view{"SQ Sequence "}, stream_it);
-            ranges::copy(std::to_string(sequence_size), stream_it);
-            ranges::copy(std::string_view{" BP;"}, stream_it);
-            stream_it = '\n';
-            auto seq = sequence | ranges::view::chunk(60);
+            auto res = std::to_chars(&buffer[0], &buffer[0] + sizeof(buffer), sequence_size);
+            std::copy(&buffer[0], res.ptr, stream_it);
+            ranges::copy(std::string_view{" BP;\n"}, stream_it);
+            auto seqChunk = sequence | ranges::view::chunk(60);
             unsigned int i = 0;
             size_t bp = 0;
-            while (bp < sequence_size)
+            for (auto chunk : seqChunk)
             {
-                ranges::copy(seq[i] | view::to_char
-                                    | ranges::view::chunk(10)
-                                    | ranges::view::join(' '), stream_it);
-                bp = std::min(sequence_size, bp + 60);
+                ranges::copy(chunk | view::to_char
+                                   | ranges::view::chunk(10)
+                                   | ranges::view::join(' '), stream_it);
                 ++i;
                 stream_it = ' ';
+                bp = std::min(sequence_size, bp + 60);
                 uint8_t num_blanks = 60 * i - bp;  // for sequence characters
                 num_blanks += num_blanks / 10;     // additional chunk separators
-                ranges::copy(std::string(num_blanks, ' '), stream_it);
+                ranges::copy(ranges::view::repeat_n(' ', num_blanks), stream_it);
                 ranges::copy(std::to_string(bp), stream_it);
                 stream_it = '\n';
             }
--- a/include/seqan3/io/sequence_file/format_embl.hpp
+++ b/include/seqan3/io/sequence_file/format_embl.hpp
@@ -1,0 +1,324 @@
+// ============================================================================
+//                 SeqAn - The Library for Sequence Analysis
+// ============================================================================
+//
+// Copyright (c) 2006-2018, Knut Reinert & Freie Universitaet Berlin
+// Copyright (c) 2016-2018, Knut Reinert & MPI Molekulare Genetik
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//     * Redistributions of source code must retain the above copyright
+//       notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//     * Neither the name of Knut Reinert or the FU Berlin nor the names of
+//       its contributors may be used to endorse or promote products derived
+//       from this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL KNUT REINERT OR THE FU BERLIN BE LIABLE
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+// DAMAGE.
+//
+// ============================================================================
+
+/*!\file
+ * \brief Provides the seqan3::sequence_file_format_embl class.
+ * \author Mitra Darvish <mitra.darvish AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <iterator>
+#include <string>
+#include <string_view>
+#include <vector>
+
+#include <range/v3/algorithm/copy.hpp>
+#include <range/v3/utility/iterator.hpp>
+#include <range/v3/view/chunk.hpp>
+#include <range/v3/view/drop_while.hpp>
+#include <range/v3/view/join.hpp>
+#include <range/v3/view/remove_if.hpp>
+#include <range/v3/view/take_while.hpp>
+
+#include <seqan3/alphabet/nucleotide/dna5.hpp>
+#include <seqan3/alphabet/quality/aliases.hpp>
+#include <seqan3/core/metafunction/range.hpp>
+#include <seqan3/io/detail/ignore_output_iterator.hpp>
+#include <seqan3/io/detail/misc.hpp>
+#include <seqan3/io/sequence_file/input_options.hpp>
+#include <seqan3/io/sequence_file/output_options.hpp>
+#include <seqan3/io/stream/parse_condition.hpp>
+#include <seqan3/range/detail/misc.hpp>
+#include <seqan3/range/view/char_to.hpp>
+#include <seqan3/range/view/to_char.hpp>
+#include <seqan3/range/view/take.hpp>
+#include <seqan3/range/view/take_exactly.hpp>
+#include <seqan3/range/view/take_line.hpp>
+#include <seqan3/range/view/take_until.hpp>
+#include <seqan3/std/ranges>
+#include <seqan3/std/view/subrange.hpp>
+#include <seqan3/std/view/transform.hpp>
+
+namespace seqan3
+{
+/*!\brief       The EMBL format.
+ * \implements  sequence_file_input_format_concept
+ * \implements  sequence_file_output_format_concept
+ * \ingroup     sequence
+ *
+ * \details
+ *
+ * ### Introduction
+ *
+ * embl is the format used in ENA sequence records. See the
+ * (ftp://ftp.ebi.ac.uk/pub/databases/embl/doc/usrman.txt) for a an in-depth description of the format.
+ *
+ * ### Fields
+ *
+ * The embl format provides the fields seqan3::field::SEQ and seqan3::field::ID. Both fields are required when writing.
+ *
+ * ### Implementation notes
+ *
+ * When reading the ID-line the ID is read until it encounters a ';'. Unless, the option truncate_ids is set to true,
+ * than the id is read until it either sees a blank, ';' or a new line.
+ *
+ * When writing the ID-line, the sequence length is appended.
+ *
+ * All other identifiers apart from 'ID' and 'SQ' are currently ignored.
+ *
+ * Passed qualities to either the read or write function are ignored.
+ *
+ */
+class sequence_file_format_embl
+{
+public:
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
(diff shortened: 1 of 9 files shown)


include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,310 @@
...
+                                    | ranges::view::chunk(10)
+                                    | ranges::view::join(' '), stream_it);
+                if (bp+60 > sequence_size)
+                    bp =  sequence_size;
+                else
+                    bp += 60;
+                ++i;
+                stream_it = ' ';
+                ranges::copy(std::to_string(bp), stream_it);
+                stream_it = '\n';
```
joergi-w: The numbers will not be properly right-aligned, especially the last line, if sequence_size is not a multiple of 60.

test/unit/io/sequence_file/sequence_file_format_embl_test.cpp
```diff
@@ -0,0 +1,419 @@
...
+        seq_qual.clear();
+
+        format.read(istream, options2, seq_qual, id, seq_qual);
+
+        EXPECT_TRUE((ranges::equal(id, expected_ids[i])));
+        EXPECT_TRUE((ranges::equal(seq_qual | view::convert<dna5>, expected_seqs[i])));
+    }
+}
+
+
```
joergi-w: double newline

test/unit/io/sequence_file/sequence_file_format_embl_test.cpp
```diff
@@ -0,0 +1,419 @@
...
+        "ID TEST 1; 4 BP.\n"
+        "XX\n"
+        "SQ Sequence 4 BP;\n"
+        "ACGT 4\n"
+        "//\n"
+        "ID Test2; 91 BP.\n"
+        "XX\n"
+        "SQ Sequence 91 BP;\n"
+        "AGGCTGNAGG CTGNAGGCTG NAGGCTGNAG GCTGNAGGCT GNAGGCTGNA GGCTGNAGGC 60\nTGNAGGCTGN AGGCTGNAGG CTGNAGGCTG N 91\n"
+        //                                             linebreak inserted after 60 char  ^
```
joergi-w: align? (also below)

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,310 @@
...
+
+        // ID
+        detail::consume(stream_view | ranges::view::take_while(is_blank));
+
+        // read id
+        if constexpr (!detail::decays_to_ignore_v<id_type>)
+        {
+            if (options.truncate_ids)
+            {
+                ranges::copy(stream_view | view::take_until_or_throw(is_blank || is_char<';'>)
```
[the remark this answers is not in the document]
MitraDarja: I don't enforce ';' any more because I want to truncate the ids, so as soon as I see a blank, I regard the rest. Should I mention that in a comment and if so, in the implementation notes or just in the line above the code?

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,310 @@
...
+            while (bp < sequence_size)
+            {
+                ranges::copy(seq[i] | view::to_char
+                                    | ranges::view::chunk(10)
+                                    | ranges::view::join(' '), stream_it);
+                if (bp+60 > sequence_size)
+                    bp =  sequence_size;
+                else
+                    bp += 60;
+                ++i;
```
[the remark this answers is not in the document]
MitraDarja: i is used in ranges::copy(seq[i] | view::to_char ...

include/seqan3/io/sequence_file/format_embl.hpp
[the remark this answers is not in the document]
joergi-w: Okay. Please mention it in the user documentation (implementation notes), because this affects how the user's file is parsed. In the case of the truncate option, consider not only checking for the blank character, but also for whitespace in general. My example above shows how confusing it behaves, if you only consider the blank character.

include/seqan3/io/sequence_file/format_embl.hpp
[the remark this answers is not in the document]
joergi-w: Ah, true! Sorry, I didn't spot it, because it's above...

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,352 @@
...
+            throw std::logic_error{"An entry has to start with the code word ID."};
+
+        if (options.complete_header)
+        {
+            while (identifier != "SQ")
+            {
+                if constexpr (!detail::decays_to_ignore_v<id_type>)
+                {
+                    ranges::copy(identifier  | view::char_to<value_type_t<id_type>>,
+                                 std::back_inserter(id));
```
joergi-w: Why do you copy the "ID" identifier into the id string for every line? I suggest only to copy the text itself; without any format-specific stuff. Imagine, you use the `id` and `sequence` variables for printing FASTA...

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,352 @@
...
+                    ranges::copy(stream_view | view::take_until_or_throw(is_char<'\n'>)
```
joergi-w: For this we have the shortcut `take_line_or_throw`, because additionally `is_char<'\r'>` should be considered.

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,352 @@
...
+                                             | view::char_to<value_type_t<id_type>>,
+                                 std::back_inserter(id));
+                    ranges::copy(std::string_view{"\n"} | view::char_to<value_type_t<id_type>>,
+                                 std::back_inserter(id));
```
joergi-w: Please do not insert newline characters into the id string. It is problematic for formats that allow only single-line ids.

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,352 @@
...
+                             std::back_inserter(identifier));
+            }
+        }
+        else
+        {
+            // ID
+            detail::consume(stream_view | ranges::view::take_while(is_blank));
+
+            // read id
+            if constexpr (!detail::decays_to_ignore_v<id_type>)
```
joergi-w: Please switch the conditions (should be much shorter then): 
```cpp
{
    if (options.complete_header)
        copy everything...
    else
        copy until whitespace or ';'...
}
else
{
    consume...
}
```

MitraDarja:
@joergi-w I discussed this with Svenja and the best solution for the header was to store all information if the user thinks the information in there is of some sort of value. I mean it is an option and the default is to not store everything.

include/seqan3/io/sequence_file/format_embl.hpp
MitraDarja: Again, if we want to store the complete header, this includes the newline characters. Doesn't it?

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,347 @@
...
+                    throw std::runtime_error{"The ID field may not be empty when writing embl files."};
+
+                stream_it = 'I';
+                stream_it = 'D';
+                stream_it = ' ';
+                ranges::copy(id, stream_it);
+                stream_it = ';';
+                stream_it = ' ';
+                ranges::copy(std::to_string(sequence_size), stream_it);
+                ranges::copy(std::string_view{" BP.\nXX\n"}, stream_it);
```
joergi-w: Question: Is the line consisting of two 'X' part of the format specification? 
You introduce **always** multi-line ids with this. And when you parse, you do not enforce them.

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,347 @@
...
+            }
+
+        }
+
+        // Sequence
+        if constexpr (detail::decays_to_ignore_v<seq_type>) // sequence
+        {
+            throw std::logic_error{"The SEQ and SEQ_QUAL fields may not both be set to ignore when writing embl files."};
```
joergi-w: line length should be ≤ 120

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,347 @@
...
+        }
+        else
+        {
+            if (ranges::empty(sequence)) //[[unlikely]]
+                throw std::runtime_error{"The SEQ field may not be empty when writing embl files."};
+
+            ranges::copy(std::string_view{"SQ Sequence "}, stream_it);
+            ranges::copy(std::to_string(sequence_size), stream_it);
```
joergi-w: Do we really need the sequence size twice – in ID and again in SQ?

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,347 @@
...
+                        stream_it = ' ';
+                        if (j % 10 == 0)
+                            stream_it = ' ';
+                    }
+                    ranges::copy(std::to_string(bp), stream_it);
+
+                }
+                else
+                {
+                    ranges::copy(std::to_string(bp), stream_it);
```
joergi-w: duplicate in both if/else branches: move below

test/unit/io/sequence_file/sequence_file_format_embl_test.cpp
joergi-w: please remove one of the empty lines

include/seqan3/io/sequence_file/format_embl.hpp
joergi-w: Actually, you can omit the whole condition and loop, because you can calculate the number of blanks with the following expression:
```cpp
uint8_t num_blanks = 60 * i - bp;  // for sequence characters
num_blanks += num_blanks / 10;     // additional chunk separators
```
For full line length (bp = 60*i) it will be 0, so no case distinction is needed.

include/seqan3/io/sequence_file/format_embl.hpp
MitraDarja: The XX are optional, I don't remember why I put them there, I will delete this.

include/seqan3/io/sequence_file/format_embl.hpp
MitraDarja: No, but there is no harm to have it at both positions and in most examples it is given twice, e.g. here: https://www.genomatix.de/online_help/help/sequence_formats.html

include/seqan3/io/sequence_file/format_embl.hpp
MitraDarja: I covered this in my test case and it seems fine to me.

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -274,31 +261,32 @@ class sequence_file_format_embl
...
             }
 
         }
 
         // Sequence
         if constexpr (detail::decays_to_ignore_v<seq_type>) // sequence
         {
-            throw std::logic_error{"The SEQ and SEQ_QUAL fields may not both be set to ignore when writing embl files."};
+            throw std::logic_error{"The SEQ and SEQ_QUAL fields may not both be set to ignore when writing"
+            "embl files."};
```
joergi-w: You need a space inside the string at the line break, otherwise you produce `writingembl`.

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -320,21 +308,11 @@ class sequence_file_format_embl
...
-
-                }
-                else
-                {
-                    ranges::copy(std::to_string(bp), stream_it);
-                }
+                uint8_t num_blanks = 60 * i - bp;  // for sequence characters
+                num_blanks += num_blanks / 10;     // additional chunk separators
+                for(int j = 0; j < num_blanks; j++)
+                    stream_it = ' ';
```
joergi-w: Please avoid unnecessary loops. See the interface of the std::string constructor (2) [here](https://en.cppreference.com/w/cpp/string/basic_string/basic_string) and use for instance:
```cpp
ranges::copy(std::string(num_blanks, ' '), stream_it);
```
Probably there is also a solution with the ranges library, if you prefer.

test/unit/io/sequence_file/sequence_file_format_embl_test.cpp
```diff
@@ -169,6 +168,7 @@ TEST_F(read, other_lines)
     do_read_test(input);
 }
 
+
```
joergi-w: you inserted a double newline here

test/unit/io/sequence_file/sequence_file_format_embl_test.cpp
```diff
@@ -357,6 +379,29 @@ TEST_F(read, seq_qual)
     }
 }
 
+TEST_F(read, illegal_alphabet)
+{
+    std::string input
+    {
+        "IK ID1;\tstuff\n"
```
joergi-w: This test throws already for the wrong ID identifier, right? So illegal alphabet characters are not actually tested...

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,324 @@
...
+        else
+        {
+            if (ranges::empty(sequence)) //[[unlikely]]
+                throw std::runtime_error{"The SEQ field may not be empty when writing embl files."};
+
+            ranges::copy(std::string_view{"SQ Sequence "}, stream_it);
+            ranges::copy(std::to_string(sequence_size), stream_it);
+            ranges::copy(std::string_view{" BP;"}, stream_it);
+            stream_it = '\n';
+            auto seq = sequence | ranges::view::chunk(60);
```
h-2: I would maybe rename this "chunks" so it is clearer that it is actually a range-of-range and not a range...

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,324 @@
...
+            unsigned int i = 0;
+            size_t bp = 0;
+            while (bp < sequence_size)
+            {
+                ranges::copy(seq[i] | view::to_char
+                                    | ranges::view::chunk(10)
+                                    | ranges::view::join(' '), stream_it);
+                bp = std::min(sequence_size, bp + 60);
```
h-2: instead of checking this while condition and accessing seq[i] I would recommend just iterating over the chunks

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,324 @@
...
+                ++i;
+                stream_it = ' ';
+                uint8_t num_blanks = 60 * i - bp;  // for sequence characters
+                num_blanks += num_blanks / 10;     // additional chunk separators
+                ranges::copy(std::string(num_blanks, ' '), stream_it);
```
h-2: constructing a new string on every line just to print blanks is terribly inefficient. You could use view::repeat_n instead

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,324 @@
...
+                ranges::copy(std::to_string(bp), stream_it);
```
h-2: use charconv again

test/unit/io/sequence_file/sequence_file_format_embl_test.cpp
```diff
@@ -0,0 +1,573 @@
...
+// ARE DISCLAIMED. IN NO EVENT SHALL KNUT REINERT OR THE FU BERLIN BE LIABLE
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+// DAMAGE.
+//
+// ==========================================================================
```
h-2: copyright header

test/unit/io/sequence_file/sequence_file_format_embl_test.cpp
```diff
@@ -0,0 +1,573 @@
...
+        "ID ID2;\n"
+        "SQ Sequence 1859 BP; 609 A; 314 C; 355 G; 581 T; 0 other;\n"
+        "  ACGTTTTTTT\nTTTTTTTTTT\nTTTTTTTTTT\vTTTTTTTTTT\tTTTTTTTTTT TTTTTTTTTT 60\n"
+        "TTTTTTTTTT TTTTTTTTTT TT        82\n"
+        "//\n"
+        "ID ID3 lala;\n"
+        "AC 2389ß\n"
+        "SQ Sequence 1859 BP; 609 A; 314 C; 355 G; 581 T; 0 other;\n"
+        "  ACGTTTA        7\n"
+        "//\n"
```
h-2: These would probably be more readable if you use raw string literals:
(6) on https://en.cppreference.com/w/cpp/language/string_literal

h-2:
I have added a note to the range-module that we should add a `view::interleave` that can replace `view::chunk | view::join`.

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,324 @@
...
+        auto constexpr is_end = is_char<'/'> ;
+        if constexpr (!detail::decays_to_ignore_v<seq_type>)
+        {
+            auto seq_view = stream_view | ranges::view::remove_if(is_space || is_digit)
+            // ignore whitespace and numbers
+                                        | view::take_until_or_throw(is_end);   // until //
+
+            auto constexpr is_legal_alph = is_in_alphabet<seq_legal_alph_type>;
+            ranges::copy(seq_view | view::transform([is_legal_alph] (char const c) // enforce legal alphabet
+                                    {
```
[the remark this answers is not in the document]
MitraDarja: Because it is the same for fastaq, should I change it here to do it all in once?

test/unit/io/sequence_file/sequence_file_format_embl_test.cpp
MitraDarja: I think this is more readable, that is how the file would look like for a human, right?

eseiler:
```cpp
    std::string input
    {
R"(ID ID1;  stuff
XX
SQ Sequence 1859 BP; 609 A; 314 C; 355 G; 581 T; 0 other;
  ACGTTTTTTT
TTTTTTTT        18
//
ID ID2;
SQ Sequence 1859 BP; 609 A; 314 C; 355 G; 581 T; 0 other;
  ACGTTTTTTT
TTTTTTTTTT
TTTTTTTTTTTTTTTTTTTT    TTTTTTTTTT TTTTTTTTTT 60
TTTTTTTTTT TTTTTTTTTT TT        82
//
ID ID3 lala;
AC 2389ß
SQ Sequence 1859 BP; 609 A; 314 C; 355 G; 581 T; 0 other;
  ACGTTTA        7
//)"
    };
```

would work.
Though I don't know how to write `\v`.

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,324 @@
...
+            throw std::logic_error{"The SEQ and SEQ_QUAL fields may not both be set to ignore when writing "
+            "embl files."};
+        }
+        else
+        {
+            if (ranges::empty(sequence)) //[[unlikely]]
+                throw std::runtime_error{"The SEQ field may not be empty when writing embl files."};
+
+            ranges::copy(std::string_view{"SQ Sequence "}, stream_it);
+            ranges::copy(std::to_string(sequence_size), stream_it);
```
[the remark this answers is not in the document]
MitraDarja: @h-2 , @smehringer , when I use the following:
```

char buffer[50];
 auto res = std::to_chars(&buffer[0], &buffer[0] + sizeof(buffer), sequence_size);
                ranges::copy(buffer, stream_it);
```

I get printed out the whole buffer, which might look like this: \xA3~\0\0\0\0\0\x2\0\0\0\x15\0\0\0)\xA5~\0\0\0\0\0\x1\xA4~\0\0\0\0\0\0\0\0\0\0\0\0@)\xA5~\0\0\0\0\00\xA5

Any idea, how to solve this problem?

test/unit/io/sequence_file/sequence_file_format_embl_test.cpp
MitraDarja: But the problem is that I than can't have any tabs before ID ID2; or ID ID3, because I am assuming I am always at the beginning of an ID, if I am not after reading one I get an error. Do you want me to consume this empty fields in the beginning? If not, then using raw string literals would mean to have it look like Enrico posted it, so the strings are not indented... Which would be inconsistent with other files...

h-2: It's ok that it's not indented, I think it will still be a lot clearer to see that this is "raw text"

include/seqan3/io/sequence_file/format_embl.hpp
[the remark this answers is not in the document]
h-2: No, just do it here for now!

include/seqan3/io/sequence_file/input_options.hpp
```diff
@@ -26,6 +26,8 @@ struct sequence_file_input_options
 {
     //!\brief Read the ID string only up until the first whitespace character.
     bool truncate_ids = false;
+    //!\brief Read the complete header to id.
+    bool complete_header = false;
```
h-2: please prefix the variable name with `embl_` so it is clear that its a NO-OP for other formats.

include/seqan3/io/sequence_file/output_options.hpp
```diff
@@ -37,6 +37,9 @@ struct sequence_file_output_options
      *        recommended ("\r\n" for line-ending).
      */
     bool        add_carriage_return     = false;
+
+    //!\brief Complete header given for embl or genbank
+    bool        complete_header  = false;
```
h-2: same here

test/unit/io/sequence_file/sequence_file_format_embl_test.cpp
eseiler: We tried this in one of @smehringer's snippets.
One problem is that you need to process the raw string to replace the spaces with tabs again.
We did it for a SAM file, where it is relatively easy since you don't have any spaces in the file we used and can just replace consecutive spaces with a tab.
This is quite handy if you want to print the SAM file and then use it since you only need to modify the raw string in one place.

But here we have mixed spaces and tabs in the format, so this might not be really applicable at all.

h-2: But you can put literal tabs into the raw string, than it works fine, or not?

eseiler: Mhm I guess you could, but then you need to make your editor behave....

And sometimes (like for SAM), you want to use extra spaces to align the columns for readability.

MitraDarja: @h-2 Just to clarify, should I change this just for this one example or also for all other embl inputs?

MitraDarja: @h-2 A question, which is more important for SAM than EMBL but concerns the Raw string, what do I do if my line gets too long due to a long sequence, before I could just break the sequence and put some parts in the next line but with a raw string this means I have a new line character in there.

h-2: Exceeding the line length is ok for raw strings. Raw string are just a ways of clearly communicating to the reader: "THIS IS NOT CPLUSPLUS, its' just a block of data"

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,290 @@
...
+#include <iterator>
+#include <string>
+#include <string_view>
+#include <vector>
+
+#include <range/v3/view/chunk.hpp>
+#include <range/v3/view/join.hpp>
+#include <range/v3/view/remove_if.hpp>
+#include <range/v3/view/repeat_n.hpp>
+#include <range/v3/view/take_while.hpp>
```
h-2: please remove all of the previous include except the one for chunk. All of these are now included in seqan3/std/ranges.

Use std::view::filter with negated condition instead of ranges::view::remove_if

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,290 @@
...
+#include <seqan3/range/view/take_line.hpp>
+#include <seqan3/range/view/take_until.hpp>
+#include <seqan3/std/charconv>
+#include <seqan3/std/ranges>
+
+namespace seqan3
+{
+/*!\brief       The EMBL format.
+ * \implements  SequenceFileInputFormat
+ * \implements  SequenceFileOutputFormat
```
h-2: Are links generated for this in the documentation? If not, please prefix the concept names with `seqan3::`

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,290 @@
...
+    void read(stream_type                                                               & stream,
+              sequence_file_input_options<seq_legal_alph_type, seq_qual_combined> const & options,
+              seq_type                                                                  & sequence,
+              id_type                                                                   & id,
+              qual_type                                                                 & SEQAN3_DOXYGEN_ONLY(qualities))
+    {
+        auto stream_view = std::ranges::subrange<decltype(std::istreambuf_iterator<char>{stream}),
+                                          decltype(std::istreambuf_iterator<char>{})>
+                            {std::istreambuf_iterator<char>{stream},
+                             std::istreambuf_iterator<char>{}};
```
h-2: indent

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,290 @@
...
+        auto stream_it = ranges::begin(stream_view);
```
h-2: std::ranges::begin
also in other places

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,290 @@
...
+
+        std::string idbuffer;
+        std::ranges::copy(stream_view | view::take_until_or_throw(is_cntrl || is_blank),
+                     std::back_inserter(idbuffer));
```
h-2: indent

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,290 @@
...
+
+        if constexpr (!detail::decays_to_ignore_v<id_type>)
+        {
+            if (options.embl_genbank_complete_header)
+            {
+                std::ranges::copy(idbuffer | view::char_to<value_type_t<id_type>>, std::back_inserter(id));
+                do
+                {
+                    std::ranges::copy(stream_view | view::take_until_or_throw(is_char<'S'>)
+                                             | view::char_to<value_type_t<id_type>>,
```
h-2: indent

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,290 @@
...
+                    id.push_back(*stream_it);
+                    ++stream_it;
+                } while (*stream_it != 'Q');
+                id.pop_back(); // remove 'S' from id
+                idbuffer = "SQ";
+            }
+            else
+            {
+                // ID
+                detail::consume(stream_view | ranges::view::take_while(is_blank));
```
h-2: `view::take_until(!is_blank)`

include/seqan3/io/sequence_file/format_embl.hpp
h-2: also in many other places where std::ranges::copy is used

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,290 @@
...
+        ++stream_it;
+        ++stream_it;
+        ++stream_it;
+
+        // make sure "buffer at end" implies "stream at end"
+        if ((std::istreambuf_iterator<char>{stream} == std::istreambuf_iterator<char>{}) &&
+            (!stream.eof()))
+        {
+            stream.get(); // triggers error in stream and sets eof
+        }
```
h-2: the previous few lines are not needed since #835

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,290 @@
...
+              typename id_type,
+              typename qual_type>
+    void write(stream_type                          & stream,
+               sequence_file_output_options const   & options,
+               seq_type                             && sequence,
+               id_type                              && id,
+               qual_type                            && SEQAN3_DOXYGEN_ONLY(qualities))
+    {
+
+        ranges::ostreambuf_iterator stream_it{stream};
```
h-2: std::ranges:...
and remove empty line before

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,290 @@
...
+            std::copy(&buffer[0], res.ptr, stream_it);
+            std::ranges::copy(std::string_view{" BP;\n"}, stream_it);
+            auto seqChunk = sequence | ranges::view::chunk(60);
+            unsigned int i = 0;
+            size_t bp = 0;
+            for (auto chunk : seqChunk)
+            {
+                std::ranges::copy(chunk | view::to_char
+                                   | ranges::view::chunk(10)
+                                   | ranges::view::join(' '), stream_it);
```
h-2: indent and std::view::join

h-2: or better view::interleave if we get #742 merged before

include/seqan3/io/sequence_file/format_embl.hpp
h-2: this is resolved?

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -0,0 +1,290 @@
...
+                num_blanks += num_blanks / 10;     // additional chunk separators
+                std::ranges::copy(ranges::view::repeat_n(' ', num_blanks), stream_it);
+                std::ranges::copy(std::to_string(bp), stream_it);
+                stream_it = '\n';
+            }
+            std::ranges::copy(std::string_view{"//"}, stream_it);
+            stream_it = '\n';
+        }
+    }
+
```
h-2: remove empty line

include/seqan3/io/sequence_file/input.hpp
```diff
@@ -333,11 +334,12 @@ struct sequence_file_input_default_traits_aa : sequence_file_input_default_trait
...
-                                                                                          field::QUAL>,
-    detail::TypeListOfSequenceFileInputFormats valid_formats_      = type_list<sequence_file_format_fasta,
-                                                                                             sequence_file_format_fastq
-                                                                                             /*, ...*/>,
+                                                                            field::ID,
+                                                                            field::QUAL>,
+    detail::TypeListOfSequenceFileInputFormats valid_formats_      = type_list<sequence_file_format_embl,
+                                                                               sequence_file_format_fasta,
+                                                                               sequence_file_format_fastq
+                                                                               /*, ...*/>,
```
h-2: this needs to be changed now that sam is also in

include/seqan3/io/sequence_file/output.hpp
```diff
@@ -31,6 +31,7 @@
 #include <seqan3/io/detail/misc_output.hpp>
 #include <seqan3/io/detail/out_file_iterator.hpp>
 #include <seqan3/io/detail/record.hpp>
+#include <seqan3/io/sequence_file/format_embl.hpp>
 #include <seqan3/io/sequence_file/format_fasta.hpp>
```
h-2: careful here as well now that sam is merged

include/seqan3/io/sequence_file/format_embl.hpp
MitraDarja: Stupid question, but how do I check that?

include/seqan3/io/sequence_file/format_embl.hpp
MitraDarja: I think so, Jörg approved this PR already and the tests are fine... Or do you disagree, @joergi-w ?

PR #522, opened 2018-10-28
forbid conversion operators into the type itself

marehr:
This is the main reason why clang seg faults. (There are still some other major defects, like memory leaks, (new) seg faults and other stuff).

saarraz/clang-concepts/issues/12

### DIFF
--- a/include/seqan3/alphabet/nucleotide/dna15.hpp
+++ b/include/seqan3/alphabet/nucleotide/dna15.hpp
@@ -152,7 +152,7 @@ struct dna15
     //!\tparam other_nucl_type The type to convert to; must satisfy seqan3::nucleotide_concept and have the same \link value_size \endlink.
     template <typename other_nucl_type>
     //!\cond
-        requires nucleotide_concept<other_nucl_type> && value_size == alphabet_size_v<other_nucl_type>
+        requires !std::Same<dna15, other_nucl_type> && nucleotide_concept<other_nucl_type> && value_size == alphabet_size_v<other_nucl_type>
     //!\endcond
     constexpr operator other_nucl_type() const noexcept
     {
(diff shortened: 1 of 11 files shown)


include/seqan3/alphabet/nucleotide/dna15.hpp
```diff
@@ -152,7 +152,7 @@ struct dna15
     //!\tparam other_nucl_type The type to convert to; must satisfy seqan3::nucleotide_concept and have the same \link value_size \endlink.
     template <typename other_nucl_type>
     //!\cond
-        requires nucleotide_concept<other_nucl_type> && value_size == alphabet_size_v<other_nucl_type>
+        requires !std::Same<dna15, other_nucl_type> && nucleotide_concept<other_nucl_type> && value_size == alphabet_size_v<other_nucl_type>
```
smehringer: maybe put the `nucleotide_concept` into the template declaration (short concept notation) so that this line becomes shorter? 
Hm I see that you explicitly changed the short notation to the long one below.. why ?

marehr: The order is important, because we have an infinite recursion (on clang) if we first go into `nucleotide_concept` and afterwards check the `Same` condition.

smehringer: Is this a bug in clang?

smehringer: or is this "The correct way of doing it"?

marehr: Sry for answering late, this apparently is the correct way in the current standard.

marehr:
@h-2 can you merge this, äh, slowly :grimacing:

h-2:
This PR is superseded by #535

PR #523, opened 2018-10-29
Make alphabet tests more modular

marehr:
While trying to get clang working, I figured out that it would be much easier if each alphabet would have its own test case. (Some alphabets build with clang, others don't; makes porting much easier.)

This PR does not change any code, it only moves it into several new files.

The underlying change is that we now have `<name>_testcase.hpp` files which contain all test cases for a logical subgroup, like alphabets, nucleotide, structure and so on, and that each alphabet includes all test cases to which subgroup it belongs.

For example: dna4 belongs to alphabet and nucleotide.


smehringer:
But the idea was that we can avoid code duplication with typed test cases and that we can easily change a test and check that it still works for every alphabet? 
Maybe we could instead make an `#ifdef` around the types that are checked depending on the compiler?

h-2:
Hm, I have mixed feelings about this. In the end, it's the same code being run and although it's not duplicated per se, we do have 1000 added lines of code....

Maybe we could split it, but not into as many bits? So that we have the alphabet basic stuff for nucleotides tested in the nucleotide test, but we don't add a test for each individual nucleotide? Especially with my current refactoring the different nucleotides/qualities/... each differ only by a line or two of code.
This would basically mean splitting up alphabet, but not the sub-modules. This also prevents some problems introduced now regarding conversion and the type_list of types that can easily get out of sync if the everything is split up.

Also I think `testcase` vs `test` is not ideal :thinking:
Is this the naming that gtest uses?

@marehr @smehringer what do you think? (don't change anything, yet, let's agree on something first)

smehringer:
Hmmm,
In this context, and I hope not completely out of place, I was thinking earlier that maybe we should have one test case for each hpp file.. then not only our sub/modules correspond to the documentation but also to the tests and vice versa? The corresponding tests should cover the functionality in the hpp file. 

Considering this, I even think that this restructering can be a cleaner way to keep track of what is tested because you can look into the dna4_test file and see that it is tested for alphabet and nucleotide stuff...

> This also prevents some problems introduced now regarding conversion and the type_list of types that can easily get out of sync if the everything is split up.

I don't understand what gets out of sync?

> Also I think testcase vs test is not ideal
Me neither. that's why I would introduce one test file for each header file named `_test` and include what ever is testing the file there.

marehr:
> Hm, I have mixed feelings about this. In the end, it's the same code being run and although it's not duplicated per se, we do have 1000 added lines of code....

20x50 lines of license shenanigans... We should introduce short headers for test cases, they only live in this repo, are not distributed and therefore a `license.txt` file would suffice.

> Hmmm,
> In this context, and I hope not completely out of place, I was thinking earlier that maybe we should have one test case for each hpp file.. then not only our sub/modules correspond to the documentation but also to the tests and vice versa? The corresponding tests should cover the functionality in the hpp file.
> 
> Considering this, I even think that this restructering can be a cleaner way to keep track of what is tested because you can look into the dna4_test file and see that it is tested for alphabet and nucleotide stuff...

This was exactly my thought. Consistency in the library and easier access for new people. It is also much more easy to see which files have test cases.


> > This also prevents some problems introduced now regarding conversion and the type_list of types that can easily get out of sync if the everything is split up.
> 
> I don't understand what gets out of sync?

I think conversion tests should be in special test cases which test the interactions between the base classes. They are **not** highly coupled with the base classes and therefore should have special places where they reside.

> 
> > Also I think `testcase` vs `test` is not ideal
> 
> Me neither. that's why I would introduce one test file for each header file named `_test` and include what ever is testing the file there.

Totally agree, I'm open for any suggestions.

How about
* `nucleotide_module_test.hpp` 
* `nucleotide_test_collection.hpp`
* `nucleotide_tests.hpp`
* `nucleotide_base_test.hpp`
* `nucleotide_shared_test.hpp`

It should be clear that they are shareable test case parts, because they are all in `.hpp` files.

h-2:
> 20x50 lines of license shenanigans... We should introduce short headers for test cases, they only live in this repo, are not distributed and therefore a license.txt file would suffice.

I agree. Let's discuss if we want to change this throughout the library or only in the tests.

> This was exactly my thought. Consistency in the library and easier access for new people. It is also much more easy to see which files have test cases.

Yes, these are benefits that might outweigh the scattering of information...

> I think conversion tests should be in special test cases which test the interactions between the base classes. They are not highly coupled with the base classes and therefore should have special places where they reside.

We could make extra integration-tests per module, but it does add extra possibility to forget to add new types...

I am not against these changes per se, let's discuss this with @rrahn when he's back.

test/unit/alphabet/aminoacid/aminoacid_conversion_test.cpp
```diff
@@ -0,0 +1,102 @@
+// ==========================================================================
```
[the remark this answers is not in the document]
marehr: I think i don't want to make them to big; maybe let them split and name them 

* `aminoacid_conversion_integration_test.cpp`
* `aminoacid_translation_integration_test.cpp`

test/unit/alphabet/structure/structured_aa_test.cpp
```diff
@@ -0,0 +1,328 @@
...
+{
+    structured_aa<aa27, dssp9> t0{aa27::C, dssp9::E};
+    structured_aa<aa27, dssp9> t1{aa27::C, dssp9::E};
+    structured_aa<aa27, dssp9> t2{};
+    structured_aa<aa27, dssp9> t3{};
+
+    std::swap(t1, t2);
+    EXPECT_EQ(t2, t0);
+    EXPECT_EQ(t1, t3);
+}
```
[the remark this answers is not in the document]
marehr: I opened a follow-up issue #561

marehr:
> Since quite a few new files are being introduced we could already introduce the short copyright header if we agree on that before this is merged (I don't suggest changing existing headers within this PR, but adding 600 lines of copyright headers just to remove them again in another PR 10min later seems like a foolish consistency)

I Agree, especially on the `if we agree on that before this is merged` part. I would pledge for merging as it is (without changing the copyright stuff) since this PR is ready from my POV and we haven't completely decided on #553, yet. Therefore this PR shouldn't be blocked.

test/unit/alphabet/structure/structured_aa_test.cpp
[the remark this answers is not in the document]
h-2: Ok, I have re-evaluated this, can you just remove **all** of the custom stuff for `structured_rna` and `structured_aa`? I will add a new template for the cartesian compositions as part of #535 and take care of the few specific tests.

test/unit/alphabet/structure/dot_bracket3_test.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+#include "../alphabet_test_template.hpp"
+#include "../alphabet_constexpr_test_template.hpp"
+#include "structure_test_template.hpp"
+
+INSTANTIATE_TYPED_TEST_CASE_P(dot_bracket3, alphabet, dot_bracket3);
+INSTANTIATE_TYPED_TEST_CASE_P(dot_bracket3, alphabet_constexpr, dot_bracket3);
+INSTANTIATE_TYPED_TEST_CASE_P(dot_bracket3, structure, dot_bracket3);
+
+TEST(dot_bracket3_literals, vector)
```
h-2: dot_bracket3, literals

test/unit/alphabet/structure/dot_bracket3_test.cpp
```diff
@@ -0,0 +1,61 @@
...
+    std::vector<dot_bracket3> vec1;
+    vec1.resize(5, dot_bracket3::PAIR_OPEN);
+    EXPECT_EQ(vec1, "((((("_db3);
+
+    std::vector<dot_bracket3> vec2{dot_bracket3::UNPAIRED, dot_bracket3::PAIR_OPEN, dot_bracket3::PAIR_OPEN,
+                                   dot_bracket3::PAIR_CLOSE, dot_bracket3::PAIR_CLOSE, dot_bracket3::UNKNOWN};
+    EXPECT_EQ(vec2, ".(())."_db3);
+}
+
+TEST(rna_structure, dot_bracket3)
```
h-2: dot_bracket3, dot_bracket3

test/unit/alphabet/structure/structure_test_template.hpp
```diff
@@ -0,0 +1,177 @@
...
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+// DAMAGE.
+//
+// ============================================================================
+
+/*!\file
+ * \author Joerg Winkler <j.winkler AT fu-berlin.de>
+ * \brief Tests for the structure alphabets.
+ */
```
h-2: remove this comment

test/unit/alphabet/structure/structure_test_template.hpp
```diff
@@ -0,0 +1,177 @@
...
+    }
+    else
+    {
+        EXPECT_FALSE(rna_structure_concept<TypeParam>);
+    }
+
+    // additional wuss test
+    EXPECT_TRUE(rna_structure_concept<wuss<>>);  // also wuss51
+    EXPECT_TRUE(rna_structure_concept<wuss<67>>);
+}
```
h-2: The stuff in this file is literally just three different tests. It was ok to group these if everything is in one file anyway, but if you split this, please also split these tests into the individual files.

test/unit/alphabet/structure/structured_aa_test.cpp
```diff
@@ -0,0 +1,328 @@
...
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+// DAMAGE.
+//
+// ============================================================================
+
+/*!\file
+ * \author Joerg Winkler <j.winkler AT fu-berlin.de>
+ * \brief Tests for the structure alphabets.
+ */
```
h-2: remove

marehr: we have an issue for this #533

test/unit/alphabet/quality/qualified_test.cpp
```diff
@@ -35,12 +35,27 @@
...
 
+using qualified_types = ::testing::Types<qualified<dna4, phred42>,
+                                         qualified<dna4, phred63>,
+                                         qualified<aa27, phred42>,
+                                         qualified<gapped<dna4>, phred42>,
+                                         dna4q>;
+
+INSTANTIATE_TYPED_TEST_CASE_P(qualified, alphabet, qualified_types);
+INSTANTIATE_TYPED_TEST_CASE_P(qualified, alphabet_constexpr, qualified_types);
+
```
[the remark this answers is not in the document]
marehr: Sry, I don't understand your question.

test/unit/alphabet/structure/structured_aa_test.cpp
marehr: I removed it

rrahn:
>I think the conversion tests would be better named integration tests immediately so we can add other integration stuff later if we want.

I don't think this is correct. IMO they are still unit tests, as they test a specific interface, just for many settings. An integration test is what our former apps test where about: testing the integration of multiple components. I will open an issue for this, because at some point we need to take care of this and I am proposing to use a spearate directory that lives parallel to unit tests, as unit test need to pass before integration tests can be executed (in the ideal world)

test/unit/alphabet/quality/qualified_test.cpp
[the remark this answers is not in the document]
h-2: You forgot the quality tests for the qualified alphabets:
```cpp
INSTANTIATE_TYPED_TEST_CASE_P(qualified, quality, quality_types);
```

test/unit/alphabet/structure/structured_aa_test.cpp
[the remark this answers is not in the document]
h-2: This is still open, please just remove the custom code here!

smehringer:
I just flew over some changes and I noticed some inconsistencies. I thought the agreement was to put everything generic, which must be provided by a module, should be tested in the corresponding template, while everything specific to the hpp file should only be tested in it's own test. But I see that you still need if constexpr in e.g. the aminoacid_template and nucleotode_template. And in the aa20/27 test file you would think there is no difference to those alphabets except the literals.

I'm sorry if this involves again more changes but I think we should be consistent throughout the tests now that we touch everything.

test/unit/alphabet/structure/structured_aa_test.cpp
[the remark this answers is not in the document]
h-2: Also for qualified and masked, all the custom code can go away because it's 90% copy'n'pasted code from the alphabet test.

marehr:
@smehringer Please make the comments at the files, s.t. I can change them

test/unit/alphabet/quality/qualified_test.cpp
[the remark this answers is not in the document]
marehr: Ah Okay! Yeah, because the original `quality_test` didn't include them

https://github.com/seqan/seqan3/blob/bf271cca3e414b0c3c35d0fc248d91cbbeb3de54/test/unit/alphabet/quality_test.cpp#L49-L54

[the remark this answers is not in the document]
marehr: Looking at the code

https://github.com/seqan/seqan3/blob/bf271cca3e414b0c3c35d0fc248d91cbbeb3de54/test/unit/alphabet/quality_test.cpp#L59-L74

I think it was not a `quality_test_template`, but a `phred_test_template` or a integration test between phreads

test/unit/alphabet/aminoacid/aminoacid_test_template.hpp
```diff
@@ -0,0 +1,130 @@
...
+    std::vector<char> input
+    {
+        'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M',
+        'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm',
+        'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z',
+        'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z',
+        '*', '!'
+    };
+    std::vector<TypeParam> cmp;
+    if constexpr (std::is_same_v<TypeParam, aa27>)
```
smehringer: Remove the aa27/aa20 specific `input` and `cmp` and ut them into the corresponding single tests.

test/unit/alphabet/aminoacid/aminoacid_test_template.hpp
```diff
@@ -0,0 +1,130 @@
...
+    EXPECT_EQ(to_char(TypeParam::P), 'P');
+    EXPECT_EQ(to_char(TypeParam::Q), 'Q');
+    EXPECT_EQ(to_char(TypeParam::R), 'R');
+    EXPECT_EQ(to_char(TypeParam::S), 'S');
+    EXPECT_EQ(to_char(TypeParam::T), 'T');
+    EXPECT_EQ(to_char(TypeParam::V), 'V');
+    EXPECT_EQ(to_char(TypeParam::W), 'W');
+    EXPECT_EQ(to_char(TypeParam::Y), 'Y');
+
+    if constexpr (std::is_same_v<TypeParam, aa27>)
```
smehringer: move to aa27.hpp

test/unit/alphabet/aminoacid/aminoacid_test_template.hpp
```diff
@@ -0,0 +1,130 @@
...
+        EXPECT_EQ(to_char(TypeParam::B), 'B');
+        EXPECT_EQ(to_char(TypeParam::J), 'J');
+        EXPECT_EQ(to_char(TypeParam::O), 'O');
+        EXPECT_EQ(to_char(TypeParam::U), 'U');
+        EXPECT_EQ(to_char(TypeParam::X), 'X');
+        EXPECT_EQ(to_char(TypeParam::Z), 'Z');
+        EXPECT_EQ(to_char(TypeParam::TERMINATOR), '*');
+        EXPECT_EQ(to_char(TypeParam::UNKNOWN), 'X');
+    }
+    else if constexpr (std::is_same_v<TypeParam, aa20>)
```
smehringer: move to aa20

test/unit/alphabet/nucleotide/nucleotide_test_template.hpp
```diff
@@ -0,0 +1,136 @@
...
+    {
+        'A', 'C', 'G', 'T', 'U', 'N',
+        'a', 'c', 'g', 't', 'u', 'n',
+        'R', 'Y', 'S', 'W', 'K', 'M', 'B', 'D', 'H', 'V',
+        'r', 'y', 's', 'w', 'k', 'm', 'b', 'd', 'h', 'v',
+        '!'
+    };
+
+    std::vector<TypeParam> cmp;
+    if constexpr (std::is_same_v<TypeParam, dna4> || std::is_same_v<TypeParam, rna4>)
```
smehringer: remove all those specific tests and put them into the corresponding dna4/... test files. Same for `to_char`

Instead I would first only test for `A,C,G` with to_char and from_char as they are common between all alphabets,
and then
```cpp
for (t::rank_type rnk = 0; rnk < alphabet_size_v<t>; rnk++)
{
    t tmp{};
    assign_rank(tmp, rnk); // assuming assign_rank and to_rank are tested with the alphabet test beforehand
    from_char(tmp, to_char(rnk))); 
    EXPECT_EQ(rnk, to_rank(tmp));
}
``` 
to test that the two functions are consistent.
What do you think? Same could be done for amino acids?

h-2:
I agree with making the quality test a phred test, because it tests implementation defined behaviour!

From my side nothing is open anymore except the points by @smehringer 

Oh, and as you anticipated the copyright header changed :rofl:  But we agreed on a final version in #553 !

test/unit/alphabet/mask/masked_test.cpp
```diff
@@ -1,76 +1,51 @@
...
+#include "../alphabet_test_template.hpp"
+#include "../alphabet_constexpr_test_template.hpp"
+
+using namespace seqan3;
+
+using masked_types = ::testing::Types<masked<dna4>, masked<dna5>>;
+
+INSTANTIATE_TYPED_TEST_CASE_P(masked, alphabet, masked_types);
+INSTANTIATE_TYPED_TEST_CASE_P(masked, alphabet_constexpr, masked_types);
+
```
h-2: remove below here:

test/unit/alphabet/quality/qualified_test.cpp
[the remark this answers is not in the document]
h-2: remove below here:

test/unit/alphabet/structure/structured_aa_test.cpp
```diff
@@ -0,0 +1,294 @@
...
+#include "../alphabet_constexpr_test_template.hpp"
+
+using namespace seqan3;
+using namespace seqan3::literal;
+
+using structured_aa_types = ::testing::Types<structured_aa<aa27, dssp9>>;
+
+INSTANTIATE_TYPED_TEST_CASE_P(structured_aa, alphabet, structured_aa_types);
+INSTANTIATE_TYPED_TEST_CASE_P(structured_aa, alphabet_constexpr, structured_aa_types);
+
```
h-2: remove below here:

test/unit/alphabet/structure/structured_rna_test.cpp
```diff
@@ -0,0 +1,368 @@
...
+#include "../alphabet_constexpr_test_template.hpp"
+
+using namespace seqan3;
+using namespace seqan3::literal;
+
+using structured_rna_types = ::testing::Types<structured_rna<rna5, dot_bracket3>, structured_rna<rna4, wuss51>>;
+
+INSTANTIATE_TYPED_TEST_CASE_P(structured_rna, alphabet, structured_rna_types);
+INSTANTIATE_TYPED_TEST_CASE_P(structured_rna, alphabet_constexpr, structured_rna_types);
+
```
h-2: remove below here:

h-2:
But can you squash this into seperate commits for the different splitting processes /module /submodules? It is not only "nicer", but it will also make my fixing a lot easier.

marehr:
Can you give me an example how fine grained you want this? when I can do it!

h-2:
I would suggest one for generic alphabet and one for every submodule, except composition where I would maintain two separate for cartesian and union.

marehr:
I'll do it later today!

marehr:
@h-2 Is this how you wanted it to be split?

smehringer:
:scream: :tada:

PR #507, opened 2018-10-12
Added parser_exception, what to do with parser_interruption?

MitraDarja:
Fixes #481

PR #527, opened 2018-10-30
[FEATURE] Updates different aspects of argument_parser

joshuak94:
1. Update integral_range_validator to arithmetic_range_validator, so that non-integer types (double, etc.) can still be validated. Also updates the snippet, and adds test cases for successful and unsuccessful uses of double.

### DIFF
--- a/include/seqan3/argument_parser/argument_parser.hpp
+++ b/include/seqan3/argument_parser/argument_parser.hpp
@@ -578,6 +578,8 @@ private:
     */
     bool id_exists(std::string const & long_id)
     {
+        if (long_id.empty())
+            return false;
         return(!(used_option_ids.insert(long_id)).second);
     }
 
@@ -588,6 +590,8 @@ private:
     */
     bool id_exists(char const short_id)
     {
+        if (short_id == '\0')
+            return false;
         return(!(used_option_ids.insert(std::string(1, short_id))).second);
     }
 
--- a/include/seqan3/argument_parser/argument_parser.hpp
+++ b/include/seqan3/argument_parser/argument_parser.hpp
@@ -54,6 +54,7 @@
 #include <seqan3/argument_parser/detail/format_man.hpp>
 #include <seqan3/argument_parser/detail/format_parse.hpp>
 #include <seqan3/io/stream/concept.hpp>
+#include <seqan3/io/stream/parse_condition.hpp>
 
 namespace seqan3
 {
--- a/include/seqan3/argument_parser/argument_parser.hpp
+++ b/include/seqan3/argument_parser/argument_parser.hpp
@@ -608,7 +608,8 @@ private:
                  detail::format_ctd*/> format{detail::format_help(0)};
 
     //!\brief List of option/flag identifiers that are already used.
-    std::set<std::string> used_option_ids;
+    std::set<std::string> used_option_ids{"h", "hh", "help", "advanced-help", "export-help", "version"};
+
 };
 
 } // namespace seqan3
--- a/include/seqan3/argument_parser/argument_parser.hpp
+++ b/include/seqan3/argument_parser/argument_parser.hpp
@@ -221,6 +221,8 @@ public:
             throw parser_design_error("Option Identifier '" + std::string(1, short_id) + "' was already used before.");
         if (id_exists(long_id))
             throw parser_design_error("Option Identifier '" + long_id + "' was already used before.");
+        if (long_id.length() == 1)
+            throw parser_design_error("Long IDs must be either empty, or longer than one character.");
         if (short_id == '\0' && long_id.empty())
             throw parser_design_error("Option Identifiers cannot both be empty.");
 
@@ -249,6 +251,8 @@ public:
             throw parser_design_error("Option Identifier '" + std::string(1, short_id) + "' was already used before.");
         if (id_exists(long_id))
             throw parser_design_error("Option Identifier '" + long_id + "' was already used before.");
+        if (long_id.length() == 1)
+            throw parser_design_error("Long IDs must be either empty, or longer than one character.");
         if (short_id == '\0' && long_id.empty())
             throw parser_design_error("Option Identifiers cannot both be empty.");
 
@@ -609,7 +613,6 @@ private:
 
     //!\brief List of option/flag identifiers that are already used.
     std::set<std::string> used_option_ids{"h", "hh", "help", "advanced-help", "export-help", "version"};
-
 };
 
 } // namespace seqan3
--- a/include/seqan3/argument_parser/argument_parser.hpp
+++ b/include/seqan3/argument_parser/argument_parser.hpp
@@ -592,6 +592,7 @@ private:
     void verify_identifiers(char const short_id, std::string const & long_id)
     {
         auto constexpr allowed = is_alnum || is_char<'_'> || is_char<'@'>;
+
         if (id_exists(short_id))
             throw parser_design_error("Option Identifier '" + std::string(1, short_id) + "' was already used before.");
         if (id_exists(long_id))
@@ -602,11 +603,12 @@ private:
             throw parser_design_error("Option identifiers may only contain alphanumeric characters, '_', or '@'.");
         if (long_id.size() > 0 && is_char<'-'>(long_id[0]))
             throw parser_design_error("First character of long ID cannot be '-'.");
+
         std::for_each(long_id.begin(), long_id.end(), [&allowed] (char c)
-            {
-                if (!(allowed(c) || is_char<'-'>(c)))
-                    throw parser_design_error("Long identifiers may only contain alphanumeric characters, '_', '-', or '@'.");
-            });
+                      {
+                          if (!(allowed(c) || is_char<'-'>(c)))
+                              throw parser_design_error("Long identifiers may only contain alphanumeric characters, '_', '-', or '@'.");
+                      });
         if (short_id == '\0' && long_id.empty())
             throw parser_design_error("Option Identifiers cannot both be empty.");
     }
--- a/include/seqan3/argument_parser/argument_parser.hpp
+++ b/include/seqan3/argument_parser/argument_parser.hpp
@@ -202,8 +202,6 @@ public:
      * \param[in]  desc      The description of the option to be shown in the help page.
      * \param[in]  spec      Advanced option specification. see seqan3::option_spec.
      * \param[in]  validator The validator applied to the value after parsing (callable).
-     *
-     * \throws seqan3::parser_design_error
      */
     template <typename option_type, validator_concept validator_type = detail::default_validator<option_type>>
     //!\cond
@@ -218,15 +216,7 @@ public:
                     option_spec const & spec = option_spec::DEFAULT,
                     validator_type validator = validator_type{}) // copy to bind rvalues
     {
-        if (id_exists(short_id))
-            throw parser_design_error("Option Identifier '" + std::string(1, short_id) + "' was already used before.");
-        if (id_exists(long_id))
-            throw parser_design_error("Option Identifier '" + long_id + "' was already used before.");
-        if (long_id.length() == 1)
-            throw parser_design_error("Long IDs must be either empty, or longer than one character.");
-        if (short_id == '\0' && long_id.empty())
-            throw parser_design_error("Option Identifiers cannot both be empty.");
-
+        verify_identifiers(short_id, long_id);
         // copy variables into the lambda because the calls are pushed to a stack
         // and the references would go out of scope.
         std::visit([=, &value] (auto & f) { f.add_option(value, short_id, long_id, desc, spec, validator); }, format);
@@ -239,8 +229,6 @@ public:
      * \param[in]  long_id  The long identifier for the flag (e.g. "integer").
      * \param[in]  desc     The description of the flag to be shown in the help page.
      * \param[in]  spec     Advanced flag specification. see seqan3::option_spec.
-     *
-     * \throws seqan3::parser_design_error
      */
     void add_flag(bool & value,
                   char const short_id,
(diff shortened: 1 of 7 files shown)


joshuak94:
@smehringer Regarding this line:
https://github.com/seqan/seqan3/blob/b8de35a06d103b06423bb4aa32c79778d1062e35/include/seqan3/argument_parser/argument_parser.hpp#L543
Does it actually work with hh? Because we can't make hh as a short_id.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -145,7 +146,7 @@ class integral_range_validator<option_value_type>
      * \param[in] min_ Minimum set for the range to test.
      * \param[in] max_ Maximum set for the range to test.
      */
-    integral_range_validator(inner_value_type const min_,
+    arithmetic_range_validator(inner_value_type const min_,
                              inner_value_type const max_) :
```
smehringer: adjust indentation :nail_care:

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -217,9 +217,9 @@ class argument_parser
                     option_spec const & spec = option_spec::DEFAULT,
                     validator_type validator = validator_type{}) // copy to bind rvalues
     {
-        if (id_exists(short_id))
+        if (id_exists(short_id) && short_id != '\0')
```
smehringer: Please remove this again and instead change the `id_exists` function : 
(This way will require only 2 code line changes instead of 4 and I think it belongs to the functionalty of `id_exists`)
example for long_id
```cpp
    bool id_exists(std::string const & long_id)
    {
        if (long_id.empty())
            return false;
        return(!(used_option_ids.insert(long_id)).second);
    }

```

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -223,6 +223,8 @@ class argument_parser
             throw parser_design_error("Option Identifier '" + long_id + "' was already used before.");
         if (short_id == '\0' && long_id.empty())
             throw parser_design_error("Option Identifiers cannot both be empty.");
+        if (id_forbidden(std::string(1, short_id)) || id_forbidden(long_id))
```
smehringer: This is a nice idea to check this but I think it might be easier to just insert the forbidden ids into the `used_option_ids` set? It would avoid having this extra if clause and the ones below. 
Add this in line 570 of `argument_parser.hpp` and see if it works. 
```cpp
used_option_ids.insert({"h", "help", "hh", ... });
```

joshuak94: Do you think it might be confusing if the error message is the same for the "forbidden IDs" and for the user set IDs?

joshuak94: Also, could we just instantiate used_option_ids with "h", "help", etc.? As such:
```cpp
std::set<std::string> used_option_ids{"h", "help", ...};
```

smehringer: > Do you think it might be confusing if the error message is the same for the "forbidden IDs" and for the user set IDs?

I think it's fine because it will tell the user that `an option with identifier -h is already taken`. Which is maybe even more true than just saying its forbidden since there IS an option -h for their program? What do you think?

 > Also, could we just instantiate used_option_ids with "h", "help", etc.? As such:

Yes!

smehringer:
> Does it actually work with hh? Because we can't make hh as a short_id.

This one is tricky but fine for now IMPOV:
We do not permit multi-character short identifier BUT therefore (and thats the reason why we forbid them) we allow clustering of flags. Which means that if you have a flag `f` and `g` you may combine them with `-fg`. Now this can be also used to specify a flag multiple times (like a count) which is often used for `-v` saying verbose level 1. So our `-hh` simply says it's the help page level 2. 

What works : We can justify that although allowing for only single-charcter flags and forbid `h` as an identifier we claim `-hh` is treating like calling `-h` twice. So you can leave it for now in this PR.

What needs to be done (Please include this in the issue list you have): We allow the above only for `-hh` right now and there is no functionalty for custom flags to list them. We should consider to add this feature.

joshuak94:
Rebased and squashed commits @smehringer

smehringer:
Sorry thats not exactly what I meant. Now there are changes in `[FIX] Prevents users from using h, hh, advanced-help, help, export-he` that actually belong to `[FEATURE] Allows multiple options to have empty short or long i...` 
Each commit should do exaclt and only what belongs to the commit message. Do you know what I mean? :innocent:

joshuak94:
Third time's a charm? @smehringer :D

joshuak94:
Rebased on current master.

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -604,7 +608,8 @@ class argument_parser
                  detail::format_ctd*/> format{detail::format_help(0)};
 
     //!\brief List of option/flag identifiers that are already used.
-    std::set<std::string> used_option_ids;
+    std::set<std::string> used_option_ids{"h", "hh", "help", "advanced-help", "export-help", "version"};
+
```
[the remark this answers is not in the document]
smehringer: Yes, it does mix them. Also, the ids in the add_option calls are appended to this one list... 
I see that this introduces the problem that having `-i` and `--i` simultaneously is not possible as it will through a `parser_design_error` that identifier  "i" was specified before.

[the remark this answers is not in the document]
smehringer: We also do not require long_ids to be longer than one char... should we ? `--i` seems wierd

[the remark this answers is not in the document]
joshuak94: I think if we require `long_id` to be longer than one character, that would be fine. And it would solve the issue with the IDs being appended to the same list.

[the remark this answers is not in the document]
joshuak94: We could add the check into the id_exists function for long_ids. And then maybe it also makes more sense to rename `id_exists` to `id_valid`?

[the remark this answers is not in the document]
joshuak94: Just kidding, I guess it's better to create a separate if statement to check for the length, so that we can output a relevant throw message.

smehringer:
Last thing:

Add 
```
if (!is_print(id))
 A non printible char is not allowed as option/flag identifier.
```

test/unit/argument_parser/argument_parser_design_error_test.cpp
```diff
@@ -86,6 +86,33 @@ TEST(parse_test, parser_design_error)
...
+    EXPECT_THROW(parser9.add_option(option_value, 'y', "z", "long identifier is one letter"),
+                 parser_design_error);
+    EXPECT_THROW(parser9.add_flag(flag_value, 'y', "z", "long identifier is one letter"),
+                 parser_design_error);
+
+    // using non-printable characters
+    argument_parser parser10("test_parser", 1, argv);
+    EXPECT_THROW(parser10.add_option(option_value, '\t', "no\n", "tab and newline don't work!"),
+                 parser_design_error);
+    EXPECT_THROW(parser10.add_flag(flag_value, '\t', "no\n", "tab and newline don't work!"),
```
smehringer:
```suggestion
    EXPECT_THROW(parser10.add_flag(flag_value, 'i', "no\n", "tab and newline don't work in long ids either!"),
```
Otherwise the short id is always checked first, the exception is thrown and you are never testing the long id

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -217,10 +218,22 @@ class argument_parser
...
+        auto constexpr allowed = is_alnum || is_char<'_'> || is_char<'@'>;
         if (id_exists(short_id))
             throw parser_design_error("Option Identifier '" + std::string(1, short_id) + "' was already used before.");
         if (id_exists(long_id))
             throw parser_design_error("Option Identifier '" + long_id + "' was already used before.");
+        if (long_id.length() == 1)
+            throw parser_design_error("Long IDs must be either empty, or longer than one character.");
+        if (!allowed(short_id))
+            throw parser_design_error("Option identifiers may only contain alphanumeric characters, '_', or '@'.");
+        if (is_char<'-'>(long_id[0]))
```
smehringer:
```suggestion
        if (long_id.size() > 0 && is_char<'-'>(long_id[0]))
```
otherwise this will segfault

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -217,10 +218,22 @@ class argument_parser
...
+            throw parser_design_error("First character of long ID cannot be '-'.");
+        std::for_each(long_id.begin() + 1, long_id.end(), [&allowed] (char c)
+            {
+                if (!(allowed(c) || is_char<'-'>(c)))
+                    throw parser_design_error("Option identifiers may only contain alphanumeric characters, '_', or '@'.");
```
smehringer:
```suggestion
                    throw parser_design_error("Long identifiers may only contain alphanumeric characters, '_', '-' or '@'.");
```

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -620,6 +584,36 @@ class argument_parser
         return(!(used_option_ids.insert(std::string(1, short_id))).second);
     }
 
+    /*!\brief Verifies that the short and the long identifiers are correctly formatted.
+     *        Specifically, checks that identifiers haven't been used before,
```
smehringer: This is not brief anymore :smile: Please put in into details.

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -620,6 +584,36 @@ class argument_parser
...
+     *        the length of long IDs is either empty or longer than one char,
+     *        the characters used in the identifiers are all valid,
+     *        and at least one of short_id or long_id is given.
+     * \param[in] short_id The short identifier of the command line option/flag.
+     * \param[in] long_id The long identifier of the command line option/flag.
```
smehringer: missing `\throws` documentation

test/unit/argument_parser/argument_parser_design_error_test.cpp
```diff
@@ -109,7 +109,7 @@ TEST(parse_test, parser_design_error)
     argument_parser parser10("test_parser", 1, argv);
     EXPECT_THROW(parser10.add_option(option_value, '\t', "no\n", "tab and newline don't work!"),
                  parser_design_error);
-    EXPECT_THROW(parser10.add_flag(flag_value, '\t', "no\n", "tab and newline don't work!"),
+    EXPECT_THROW(parser10.add_flag(flag_value, 'i', "no\n", "tab and newline don't work!"),
```
smehringer: Could you refactor this change to be part of the commit before? :)
Otherwise the change seems weird

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -600,6 +580,37 @@ class argument_parser
         return(!(used_option_ids.insert(std::string(1, short_id))).second);
     }
 
+    /*!\brief Verifies that the short and the long identifiers are correctly formatted.
+     * \details Specifically, checks that identifiers haven't been used before,
```
smehringer: details always goes to the end :nail_care:

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -600,6 +580,37 @@ class argument_parser
...
+     *          the length of long IDs is either empty or longer than one char,
+     *          the characters used in the identifiers are all valid,
+     *          and at least one of short_id or long_id is given.
+     * \param[in] short_id The short identifier of the command line option/flag.
+     * \param[in] long_id  The long identifier of the command line option/flag.
+     * \throws parser_design_error
```
smehringer: seqan3::parser_design_error

joshuak94:
Everything should be fixed now, @smehringer

smehringer:
Let's merge #507 first so this PR can be rebased on the correct exception names.

h-2:
@joshuak94 can you rebase this so we can merge it before the meeting today?

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -588,9 +575,42 @@ class argument_parser
...
+     * \throws seqan3::parser_design_error
+     * \details Specifically, checks that identifiers haven't been used before,
+     *          the length of long IDs is either empty or longer than one char,
+     *          the characters used in the identifiers are all valid,
+     *          and at least one of short_id or long_id is given.
+     */
+    void verify_identifiers(char const short_id, std::string const & long_id)
+    {
+        auto constexpr allowed = is_alnum || is_char<'_'> || is_char<'@'>;
+        if (id_exists(short_id))
```
h-2: insert newline before

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -588,9 +575,42 @@ class argument_parser
...
+            throw parser_design_error("Option Identifier '" + std::string(1, short_id) + "' was already used before.");
+        if (id_exists(long_id))
+            throw parser_design_error("Option Identifier '" + long_id + "' was already used before.");
+        if (long_id.length() == 1)
+            throw parser_design_error("Long IDs must be either empty, or longer than one character.");
+        if (!allowed(short_id) && !is_char<'\0'>(short_id))
+            throw parser_design_error("Option identifiers may only contain alphanumeric characters, '_', or '@'.");
+        if (long_id.size() > 0 && is_char<'-'>(long_id[0]))
+            throw parser_design_error("First character of long ID cannot be '-'.");
+        std::for_each(long_id.begin(), long_id.end(), [&allowed] (char c)
```
h-2: insert newline before

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -588,9 +575,42 @@ class argument_parser
...
+            {
+                if (!(allowed(c) || is_char<'-'>(c)))
+                    throw parser_design_error("Long identifiers may only contain alphanumeric characters, '_', '-', or '@'.");
+            });
```
h-2: align block with surrounding code (left indent by 2?),

PR #537, opened 2018-11-02
[FEATURE] Enable chaining of validators.

### LINKED ISSUE #192: Argument Parser - Make it possible to chain validators
It would be nice to do something like `validator1() | validator2()` or `validator1() + validator2()` or else

smehringer:
Fixes #192 

You can now do this

```
parser.add_option(..., validator1 | validator2);
```

### DIFF
--- a/include/seqan3/argument_parser/validators.hpp
+++ b/include/seqan3/argument_parser/validators.hpp
@@ -44,9 +44,10 @@
 
 #include <seqan3/argument_parser/auxiliary.hpp>
 #include <seqan3/argument_parser/exceptions.hpp>
+#include <seqan3/core/metafunction/basic.hpp>
 #include <seqan3/io/filesystem.hpp>
-#include <seqan3/std/concepts>
 #include <seqan3/range/container/concept.hpp>
+#include <seqan3/std/concepts>
 #include <seqan3/std/view/view_all.hpp>
 
 namespace seqan3
@@ -564,6 +565,113 @@ struct default_validator
     }
 };
 
+/*!\brief A helper struct to chain validators recursively via the pipe operator.
+ *\ingroup argument_parser
+ *\implements seqan3::validator_concept
+ *
+ *\details
+ *
+ * Note that both validators must operate on the same value_type in order to
+ * avoid unexpected behaviour and ensure that the seqan3::argument_parser::add_option
+ * call is well-formed. (add_option(val, ...., validator) requires
+ * that val is of same type as validator::value_type).
+ */
+template <validator_concept validator1_type, validator_concept validator2_type>
+//!\cond
+    requires std::Same<typename validator1_type::value_type, typename validator2_type::value_type>
+//!\endcond
+class validator_chain_adaptor
+{
+public:
+    //!\brief The underlying type in both validators.
+    using value_type = typename validator1_type::value_type;
+
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    //!\brief The default constructor is explicitly deleted.
+    validator_chain_adaptor() = delete;
+    validator_chain_adaptor(validator_chain_adaptor const & pf) = default;
+    validator_chain_adaptor & operator=(validator_chain_adaptor const & pf) = default;
+    validator_chain_adaptor(validator_chain_adaptor &&) = default;
+    validator_chain_adaptor & operator=(validator_chain_adaptor &&) = default;
+
+    /*!\brief Constructing from two validators.
+     * \param[in] vali1_ Some validator to be chained to vali2_.
+     * \param[in] vali2_ Another validator to be chained to vali1_.
+     */
+    validator_chain_adaptor(validator1_type vali1_, validator2_type vali2_) :
+        vali1{std::move(vali1_)}, vali2{std::move(vali2_)}
+    {}
+
+    //!\brief The destructor.
+    ~validator_chain_adaptor() = default;
+    //!\}
+
+    /*!\brief Calls the operator() of each validator on the value cmp.
+     * \param[in] cmp The value to validate.
+     *
+     * This function delegates to the validation of both of the chained validators
+     * by calling their operator() one after the other. The behaviour depends on
+     * the chained validators which may throw on input error.
+     */
+    void operator()(value_type const & cmp) const
+    {
+        vali1(cmp);
+        vali2(cmp);
+    }
+
+    //!\brief Returns a message that can be appended to the (positional) options help page info.
+    std::string get_help_page_message() const
+    {
+        return detail::to_string(vali1.get_help_page_message(), " ", vali2.get_help_page_message());
+    }
+
+private:
+    //!\brief The first validator in the chain.
+    validator1_type vali1;
+    //!\brief The second validator in the chain.
+    validator2_type vali2;
+};
+
 } // namespace detail
+
+/*!\brief Enables the chaining of validators.
+ *!\ingroup argument_parser
+ * \tparam validator1_type The type of the fist validator;
+ *                         Must satisfy the seqan3::validator_concept and the
+ *                         same value_type as the second validator type.
+ * \tparam validator2_type The type of the second validator;
+ *                         Must satisfy the seqan3::validator_concept and the
+ *                         same value_type as the fist validator type.
+ * \param[in] vali1 The first validator to chain.
+ * \param[in] vali2 The second validator to chain.
+ * \returns A new validator that tests a value for both vali1 and vali2.
+ *
+ * \details
+ *
+ * The pipe operator is the AND operation for two validators, which means that a
+ * value must pass both validators in order to be accepted by the new validator.
+ *
+ * For example you may want a file name that only accepts absolute paths but
+ * also must have one out of some given file extensions.
+ * For this purpose you can chain a seqan3::regex_validator to a
+ * seqan3::file_ext_validator like this:
+ *
+ * \include test/snippet/argument_parser/validators_chaining.cpp
+ *
+ * You can chain as many validators as you want which will be evaluated one after
+ * the other from left to right (first to last).
+ */
+template <validator_concept validator1_type, validator_concept validator2_type>
+//!\cond
+    requires std::Same<typename std::remove_reference_t<validator1_type>::value_type,
+                       typename std::remove_reference_t<validator2_type>::value_type>
+//!\endcond
+auto operator|(validator1_type && vali1, validator2_type && vali2)
+{
+    return detail::validator_chain_adaptor{std::forward<validator1_type>(vali1),
+                                           std::forward<validator2_type>(vali2)};
+}
 
 } // namespace seqan3
--- a/include/seqan3/argument_parser/validators.hpp
+++ b/include/seqan3/argument_parser/validators.hpp
@@ -61,14 +61,48 @@ namespace seqan3
  * The requirements for this concept are given as related functions and metafunctions.
  * Types that satisfy this concept are shown as "implementing this interface".
  */
+/*!\name Requirements for seqan3::validator_concept
+ * \brief You can expect these (meta-)functions on all types that implement seqan3::validator_concept.
+ * \{
+ */
+/*!\typedef     using value_type
+ * \brief       The type of value on which the validator is called on.
+ * \relates     seqan3::validator_concept
+ *
+ * \details
+ * \attention This is a concept requirement, not an actual typedef (however types satisfying this concept
+ * will provide an implementation).
+ */
+/*!\fn              void operator()(value_type const & cmp) const
+ * \brief           Validates the value 'cmp' and throws a seqan3::validation_error on failure.
+ * \tparam          value_type The type of the value to be validated.
+ * \param[in,out]   cmp The value to be validated.
+ * \relates         seqan3::validator_concept
+ * \throws          seqan3::validation_error if value 'cmp' does not pass validation.
+ *
+ * \details
+ * \attention This is a concept requirement, not an actual function (however types satisfying this concept
+ * will provide an implementation).
+ */
+/*!\fn              std::string get_help_page_message() const
+ * \brief           Returns a message that can be appended to the (positional) options help page info.
+ * \relates         seqan3::validator_concept
+ * \returns         A string that contains information on the performed validation.
+ *
+ * \details
+ * \attention This is a concept requirement, not an actual function (however types satisfying this concept
+ * will provide an implementation).
+ */
+//!\}
 //!\cond
 template <typename validator_type>
-concept validator_concept = std::Invocable<validator_type, typename validator_type::value_type> &&
-                                 requires(validator_type validator,
-                                          typename validator_type::value_type value)
+concept validator_concept = std::Copyable<remove_cvref_t<validator_type>> &&
+                            requires(validator_type validator,
+                                     typename std::remove_reference_t<validator_type>::value_type value)
 {
-    typename validator_type::value_type;
+    typename std::remove_reference_t<validator_type>::value_type;
 
+    { validator(value) } -> void;
     { validator.get_help_page_message() } -> std::string;
 };
 //!\endcond
@@ -80,6 +114,7 @@ class integral_range_validator;
 
 /*!\brief A validator that checks whether a number is inside a given range.
  * \ingroup argument_parser
+ * \implements seqan3::validator_concept
  *
  * \tparam option_value_type Must be a (container of) integral type(s).
  *
@@ -182,6 +217,7 @@ private:
 
 /*!\brief A validator that checks whether a value is inside a list of valid values.
  * \ingroup argument_parser
+ * \implements seqan3::validator_concept
  *
  * \details
  *
@@ -236,6 +272,7 @@ private:
 
 /*!\brief A validator that checks if each value in a container appears in a list of valid values.
  * \ingroup argument_parser
+ * \implements seqan3::validator_concept
  * \extends seqan3::value_list_validator
  *
  * \tparam option_value_type The container type. Must satisfy the seqan3::container_concept.
@@ -294,6 +331,7 @@ private:
 
 /*!\brief A validator that checks if a filenames has one of the valid extensions.
  * \ingroup argument_parser
+ * \implements seqan3::validator_concept
  *
  * \details
  *
@@ -399,6 +437,7 @@ class regex_validator;
 
 /*!\brief A validator that checks if a matches a regular expression pattern.
  * \ingroup argument_parser
+ * \implements seqan3::validator_concept
  *
  * \details
  *
@@ -450,6 +489,7 @@ private:
 
 /*!\brief A validator that checks if each value in a container satisfies a regex expression.
  * \ingroup argument_parser
+ * \implements seqan3::validator_concept
  * \extends seqan3::regex_validator
  */
 template <>
@@ -500,6 +540,7 @@ namespace detail
 
 /*!\brief Validator that always returns true.
  * \ingroup argument_parser
+ * \implements seqan3::validator_concept
  *
  * \details
  *
@@ -512,11 +553,9 @@ struct default_validator
     //!\brief Type of values that are tested by validator
     using value_type = option_value_type;
 
-    //!\brief Value cmp always passes validation.
-    bool operator()(option_value_type const & /*cmp*/) const noexcept
-    {
-        return true;
-    }
+    //!\brief Value cmp always passes validation because the operator never throws.
+    void operator()(option_value_type const & /*cmp*/) const noexcept
+    {}
 
     //!\brief Since no validation is happening the help message is empty.
     std::string get_help_page_message() const
--- a/test/snippet/argument_parser/validators_chaining.cpp
+++ b/test/snippet/argument_parser/validators_chaining.cpp
@@ -1,0 +1,33 @@
+#include <seqan3/argument_parser/all.hpp>
+
+int main(int argc, const char ** argv)
+{
+    seqan3::argument_parser myparser("Test", argc, argv); // initialize
+
+    std::string file_name;
+
+    seqan3::regex_validator<std::string> absolute_path_validator("(/[^/]+)+/.*\\.[^/\\.]+$");
+    seqan3::file_ext_validator my_file_ext_validator({"sa", "so"});
+
+    myparser.add_option(file_name, 'f', "file","Give me a file name/path.",
+                        seqan3::option_spec::DEFAULT, absolute_path_validator | my_file_ext_validator);
+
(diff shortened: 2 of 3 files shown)


include/seqan3/argument_parser/validators.hpp
```diff
@@ -485,6 +485,82 @@ struct default_validator
...
+ * \tparam validator_rhs_type The type of the second validator;
+ *                            Must satisfy the seqan3::validator_concept  and the
+ *                            same value_type as the fist validator type.
+ * \param[in] lhs The first validator to chain.
+ * \param[in] rhs The second validator to chain.
+ * \returns A new validator that tests a value for both lhs and rhs.
+ *
+ * \details
+ *
+ * The operator&() is the AND operation for two validator, which means that a
```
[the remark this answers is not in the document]
joergi-w: operator `+` !

include/seqan3/argument_parser/validators.hpp
```diff
@@ -485,6 +485,82 @@ struct default_validator
...
+template <validator_concept validator_lhs_type, validator_concept validator_rhs_type>
+    requires std::Same<typename validator_lhs_type::value_type, typename validator_rhs_type::value_type>
+struct validator_chain_adaptor
+{
+    //!\brief Type of value that are tested by each of the validators.
+    using value_type = typename validator_lhs_type::value_type;
+
+    /*!\brief Constructing from a vector.
+     * \param[in] lhs_ Some validator to be chained to rhs_.
+     * \param[in] rhs_ Another validator to be chained to lhs_.
```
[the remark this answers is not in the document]
smehringer: maybe lhs and rhs is confusing as the order does not matter at all. I will change that! 
If the first is false it will throw an exception so no, the second is not evaluated since the program aborts. The "constructing from a vector" is a copy and paste error sorry :see_no_evil:

include/seqan3/argument_parser/validators.hpp
```diff
@@ -485,6 +485,82 @@ struct default_validator
...
+    void operator()(value_type const & cmp) const
+    {
+        lhs(cmp);
+        rhs(cmp);
+    }
+
+    //!\brief Returns a message that can be appended to the (positional) options help page info.
+    std::string get_help_page_message() const
+    {
+        return detail::to_string(lhs.get_help_page_message(), " ", rhs.get_help_page_message());
```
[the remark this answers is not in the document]
smehringer: It is if the help message of the validator is properly formatted which must be relied on.
If I introduce a new line this may introduce wrong formatting when printing the help page because the there new lines are handled depending on the size of the terminal.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -485,6 +485,98 @@ struct default_validator
...
+    validator_chain_adaptor(validator_chain_adaptor const & pf) = default;
+    validator_chain_adaptor & operator=(validator_chain_adaptor const & pf) = default;
+    validator_chain_adaptor(validator_chain_adaptor &&) = default;
+    validator_chain_adaptor & operator=(validator_chain_adaptor &&) = default;
+
+    /*!\brief Constructing from two validators.
+     * \param[in] vali1_ Some validator to be chained to vali2_.
+     * \param[in] vali2_ Another validator to be chained to vali1_.
+     */
+    validator_chain_adaptor(validator1_type const & vali1_, validator2_type const & vali2_) :
```
rrahn: Are validator movable? If so take them by copy and move them into the object parameter

include/seqan3/argument_parser/validators.hpp
```diff
@@ -485,6 +485,98 @@ struct default_validator
     }
 };
 
+//!\brief A helper struct to chain validators recursively via operator+()
+//!\ingroup validators
+template <validator_concept validator1_type, validator_concept validator2_type>
+    requires std::Same<typename validator1_type::value_type, typename validator2_type::value_type>
+class validator_chain_adaptor
```
rrahn: does it implement the validator_concept?
Shouldn't the doxygen than say `//!\implements ...concept`?

include/seqan3/argument_parser/validators.hpp
```diff
@@ -485,6 +485,98 @@ struct default_validator
...
+    //!\brief The second validator in the chain.
+    validator2_type vali2;
+};
+
 } // namespace detail
 
+/*!\brief Enables the chaining of validators.
+ *!\ingroup validators
+ * \tparam validator1_type The type of the fist validator;
+ *                            Must satisfy the seqan3::validator_concept and the
```
rrahn: indentation

include/seqan3/argument_parser/validators.hpp
```diff
@@ -485,6 +485,98 @@ struct default_validator
...
+ * seqan3::file_ext_validator like this:
+ *
+ * \snippet test/snippet/argument_parser/validators_chaining.cpp all
+ *
+ * You can chain as many validators as you want which will be evaluated one after
+ * the other from left to right (first to last).
+ */
+template <validator_concept validator1_type, validator_concept validator2_type>
+    requires std::Same<typename validator1_type::value_type, typename validator2_type::value_type>
+auto operator+(validator1_type const & vali1, validator2_type const & vali2)
```
rrahn: Here you should use forwarding references and forward the objects into the operator.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -485,6 +485,98 @@ struct default_validator
     }
 };
 
+//!\brief A helper struct to chain validators recursively via operator+()
+//!\ingroup validators
```
rrahn: A documentation about the fact, that both value types must be the same would be good.
And is it required that they are the same or can we use a less strict concept, such as implicit_convertible_to_concept?

rrahn: Or better they must fulfill `CommonReference<T, U>` ...

include/seqan3/argument_parser/validators.hpp
```diff
@@ -485,6 +485,98 @@ struct default_validator
...
+    //!\}
+
+    /*!\brief Calls the operator() of each validator on the value cmp.
+     * \param[in] cmp The value to validate.
+     *
+     * This function delegates to the validation of both of the chained validators
+     * by calling their operator() one after the other. The behaviour depends on
+     * the chained validators which may throw on input error.
+     */
+    void operator()(value_type const & cmp) const
```
rrahn: ... here the `common_reference_t` should be used.

include/seqan3/argument_parser/validators.hpp
rrahn: As discussed: rather use the pipe symbol.

include/seqan3/argument_parser/validators.hpp
smehringer: :+1: I actually did not do that for any of the validators... I will change all of them

include/seqan3/argument_parser/validators.hpp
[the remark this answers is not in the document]
joergi-w: okay!

include/seqan3/argument_parser/validators.hpp
smehringer: My thoughts on this:

Before going into the problems, note that your proposed change would require another change for consistency: Currently in the `add_option(value, ..., validator)` call, the value type is required to be of the same type as the `validator::value_type` and and the functor of each validator has one parameter of that value_type. If we would relax this requirement to common reference type consider the following:

```
uint32_t val{};
arithmetic_range_validator<uint8_t> validator(3,10);
parser.add_option(val, ..., validator);
```
now the common type of uint8_t and uint32_t is uint32_t correct? so this code would be valid under your requirement. But comparing `val` to `uint8_t min = 3` will either trigger a compiler warning or probably yield wrong results, no? This could be solved by requiring that the type of `val` has to be implicelty castable to validator::value_type but I am not 100%sure if this would be a safe relaxation.

Now, if you chain validators a new validator with a value_type member is created. If I would only require that validator1 and validator2 hava a common reference, I would need to choose the common reference type for the new value_type, but I am not sure if this introduces complications when the adaptor is used in an `add_option` call.

Writing this down I think we could take the common reference type when chaining validators and then require the given value to be implicelty castable to the validator::value_type.... what do you think?

smehringer: @rrahn After discussing the issue can we leave this like it is for now? If so please re-review as I introduced all the other changes you requested besides this one and the common reference type below

include/seqan3/argument_parser/validators.hpp
```diff
@@ -62,12 +62,12 @@ namespace seqan3
  */
 //!\cond
 template <typename validator_type>
-concept validator_concept = std::Invocable<validator_type, typename validator_type::value_type> &&
-                                 requires(validator_type validator,
-                                          typename validator_type::value_type value)
+concept validator_concept = requires(validator_type validator,
+                                     typename std::remove_reference_t<validator_type>::value_type value)
```
rrahn: you requesting a type before it is checked in the concept. I am not sure, what happens when the type does not support a value_type member? Will this result in a compiler error or will it evaluate to false?

include/seqan3/argument_parser/validators.hpp
```diff
@@ -473,10 +480,8 @@ struct default_validator
     using value_type = option_value_type;
 
     //!\brief Value cmp always passes validation.
-    bool operator()(option_value_type const & /*cmp*/) const noexcept
-    {
-        return true;
-    }
+    void operator()(option_value_type const & /*cmp*/) const noexcept
+    {}
```
rrahn: Is the comment above still acurate with the change?

include/seqan3/argument_parser/validators.hpp
```diff
@@ -485,6 +490,109 @@ struct default_validator
     }
 };
 
+/*!\brief A helper struct to chain validators recursively via the pipe operator.
+ *\ingroup argument_parser
+ *\implements seqan3::validator_concept
+ *
+ *\details
+ *
+ * Note that both validators must operate on the same value_type in order to be
```
rrahn: `to be avoid...`

include/seqan3/argument_parser/validators.hpp
```diff
@@ -485,6 +490,109 @@ struct default_validator
...
+ * avoid unexpected behaviour and ensure that the seqan3::argument_parser::add_option
+ * call will continue to be well-formed. (add_option(val, ...., validator) requires
```
rrahn:
```diff
-seqan3::argument_parser::add_option
-call will continue to be well-formed.
+seqan3::argument_parser::add_option is well formed.
```

include/seqan3/argument_parser/validators.hpp
```diff
@@ -485,6 +490,109 @@ struct default_validator
...
+ * that val is of same type as validator::value_type).
+ */
+template <validator_concept validator1_type, validator_concept validator2_type>
+    requires std::Same<typename validator1_type::value_type, typename validator2_type::value_type>
```
rrahn: wrap in doxygen conditional

include/seqan3/argument_parser/validators.hpp
```diff
@@ -485,6 +490,109 @@ struct default_validator
...
+
 } // namespace detail
 
+/*!\brief Enables the chaining of validators.
+ *!\ingroup argument_parser
+ * \tparam validator1_type The type of the fist validator;
+ *                         Must satisfy the seqan3::validator_concept and the
+ *                         same value_type as the second validator type.
+ * \tparam validator2_type The type of the second validator;
+ *                         Must satisfy the seqan3::validator_concept  and the
```
rrahn: superflouos space

include/seqan3/argument_parser/validators.hpp
```diff
@@ -485,6 +490,109 @@ struct default_validator
...
+ * \param[in] vali2 The second validator to chain.
+ * \returns A new validator that tests a value for both vali1 and vali2.
+ *
+ * \details
+ *
+ * The pipe operator is the AND operation for two validators, which means that a
+ * value must pass both validators in order to be accepted by the new validator.
+ *
+ * For example you may want a file name that only accepts absolute paths but
+ * also must have one out of a some given file extensions.
```
rrahn: `a some`

include/seqan3/argument_parser/validators.hpp
```diff
@@ -485,6 +490,109 @@ struct default_validator
...
+ * For this purpose you can chain a seqan3::regex_validator to a
+ * seqan3::file_ext_validator like this:
+ *
+ * \snippet test/snippet/argument_parser/validators_chaining.cpp all
+ *
+ * You can chain as many validators as you want which will be evaluated one after
+ * the other from left to right (first to last).
+ */
+template <validator_concept validator1_type, validator_concept validator2_type>
+    requires std::Same<typename std::remove_reference_t<validator1_type>::value_type,
```
rrahn: docygen conditional

include/seqan3/argument_parser/validators.hpp
rrahn: I couldn't find a negative test case either.

rrahn: I think this type must be given as a second template parameter, like the invocable concepts from the STL.

smehringer: > you requesting a type before it is checked in the concept. I am not sure, what happens when the type does not support a value_type member? Will this result in a compiler error or will it evaluate to false?

tried and tested: This will not cause a compiler error but just result in the concept being not satisfied.

> I couldn't find a negative test case either.

I'll add one

> I think this type must be given as a second template parameter, like the invocable concepts from the STL.
 
no. See first answer.

smehringer:
@rrahn ping

include/seqan3/argument_parser/validators.hpp
```diff
@@ -485,6 +490,113 @@ struct default_validator
...
+    validator_chain_adaptor(validator_chain_adaptor const & pf) = default;
+    validator_chain_adaptor & operator=(validator_chain_adaptor const & pf) = default;
+    validator_chain_adaptor(validator_chain_adaptor &&) = default;
+    validator_chain_adaptor & operator=(validator_chain_adaptor &&) = default;
+
+    /*!\brief Constructing from two validators.
+     * \param[in] vali1_ Some validator to be chained to vali2_.
+     * \param[in] vali2_ Another validator to be chained to vali1_.
+     */
+    validator_chain_adaptor(validator1_type const vali1_, validator2_type const vali2_) :
```
rrahn: 🤔 The const here prevents the usage of the move constructor for rvalues.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -62,12 +62,12 @@ namespace seqan3
  */
 //!\cond
 template <typename validator_type>
-concept validator_concept = std::Invocable<validator_type, typename validator_type::value_type> &&
-                                 requires(validator_type validator,
-                                          typename validator_type::value_type value)
+concept validator_concept = requires(validator_type validator,
```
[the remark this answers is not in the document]
smehringer: Do you mean the validator type should be semi-regular?
Validators do not need to default constructible! They should be copyable, yes, but I think Semi-Regular is too restrictive concerning default constructibility.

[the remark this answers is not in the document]
smehringer: How do I document the value_type member? `\var` or `\typedef` ?

include/seqan3/argument_parser/validators.hpp
```diff
@@ -485,6 +490,113 @@ struct default_validator
...
+ *\ingroup argument_parser
+ *\implements seqan3::validator_concept
+ *
+ *\details
+ *
+ * Note that both validators must operate on the same value_type in order to
+ * avoid unexpected behaviour and ensure that the seqan3::argument_parser::add_option
+ * call is well-formed. (add_option(val, ...., validator) requires
+ * that val is of same type as validator::value_type).
+ */
```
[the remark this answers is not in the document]
smehringer: An example snippet of how to chain validators is given in the `operator|` documentation which is public.

include/seqan3/argument_parser/validators.hpp
smehringer: :thinking: True... does that mean it should not compile with temporaries? Because it does :eyes:

rrahn: Well it does compile with temporaries, but they aren't moved but copied 🤓

include/seqan3/argument_parser/validators.hpp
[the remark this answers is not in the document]
rrahn: `\typedef` I assume. You can see some examples at the scoring scheme concept
```cpp
/*!\typedef     typedef IMPLEMENTATION_DEFINED score_type;
 * \brief       The type returned by seqan3::scoring_scheme_concept::score(), usually a seqan3::arithmetic_concept.
 * \memberof seqan3::scoring_scheme_concept
 *
 * \details
 * \attention This is a concept requirement, not an actual typedef (however types satisfying this concept
 * will provide an implementation).
 */
```

smehringer:
@rrahn I also included tests for printing help page messages with the validators to get the code coverage up. (See one more commit and one change in the chaining feature commit)
Now the code coverage says I am missing out in search schemes... :thinking: :thinking: :thinking:

test/snippet/argument_parser/validators_chaining.cpp
```diff
@@ -0,0 +1,35 @@
+//! [all]
```
marehr: If you include the whole file don't use `\snippet`, but `\include`

smehringer:
Let's merge #507 first so this PR can be rebased on the correct exception names.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -525,6 +565,113 @@ struct default_validator
     }
 };
 
+/*!\brief A helper struct to chain validators recursively via the pipe operator.
+ *\ingroup argument_parser
+ *\implements seqan3::validator_concept
+ *
+ *\details
```
joergi-w: align doxygen keywords (space after `*`)

include/seqan3/argument_parser/validators.hpp
```diff
@@ -525,6 +565,113 @@ struct default_validator
...
+ *
+ * Note that both validators must operate on the same value_type in order to
+ * avoid unexpected behaviour and ensure that the seqan3::argument_parser::add_option
+ * call is well-formed. (add_option(val, ...., validator) requires
+ * that val is of same type as validator::value_type).
```
joergi-w: **the** same

include/seqan3/argument_parser/validators.hpp
```diff
@@ -525,6 +565,113 @@ struct default_validator
...
+    //!\brief The destructor.
+    ~validator_chain_adaptor() = default;
+    //!\}
+
+    /*!\brief Calls the operator() of each validator on the value cmp.
+     * \param[in] cmp The value to validate.
+     *
+     * This function delegates to the validation of both of the chained validators
+     * by calling their operator() one after the other. The behaviour depends on
+     * the chained validators which may throw on input error.
```
joergi-w: Do you mean: ...throws **an** input error?
Use throws keyword.

test/snippet/argument_parser/validators_chaining.cpp
```diff
@@ -0,0 +1,33 @@
...
+int main(int argc, const char ** argv)
+{
+    seqan3::argument_parser myparser("Test", argc, argv); // initialize
+
+    std::string file_name;
+
+    seqan3::regex_validator<std::string> absolute_path_validator("(/[^/]+)+/.*\\.[^/\\.]+$");
+    seqan3::file_ext_validator my_file_ext_validator({"sa", "so"});
+
+    myparser.add_option(file_name, 'f', "file","Give me a file name/path.",
```
joergi-w: space after comma

test/unit/argument_parser/format_parse_validators_test.cpp
```diff
@@ -163,6 +175,28 @@ TEST(validator_test, integral_range_validator_success)
...
+    std::string stdout = testing::internal::GetCapturedStdout();
+    std::string expected = "test_parser"
+                           "==========="
+                           "POSITIONAL ARGUMENTS"
+                           "    ARGUMENT-1 List of INT (32 bit)'s"
+                           "          desc Value must be in range [-20,20]."
+                           "VERSION"
+                           "    Last update: "
+                           "    test_parser version: "
+                           "    SeqAn version: 3.0.0";
```
joergi-w: Must this test be updated with every new SeqAn release?

smehringer: It's set by the SEQAN_VERSION_MACRO which will be updated every release and thus no extra maintenance is needed.

joergi-w: But here you do not use the macro, but a hard-coded `3.0.0` instead...

smehringer: ah true. good catch

include/seqan3/argument_parser/validators.hpp
```diff
@@ -525,6 +565,113 @@ struct default_validator
...
+{
+public:
+    //!\brief The underlying type in both validators.
+    using value_type = typename validator1_type::value_type;
+
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    //!\brief The default constructor is explicitly deleted.
+    validator_chain_adaptor() = delete;
```
rrahn: Just a general question. You explicitly delete the default construction here. And the general design is that it is not used somewhere else than for the  argumentparser. But are all your validators not default constructible?
So I would propose to allow validator to be a semi-regular type and in case of default construction of the adaptor use the default_validator which always passes.

smehringer: If we really want to allow default constructibility, I would propose to specifically create a validator that in the context allows every value (type dependent of course).
e.g. the regex_validator would be defaulted to `val{".*"}`

test/unit/argument_parser/format_parse_validators_test.cpp
```diff
@@ -35,16 +35,30 @@
...
+#include <seqan3/io/stream/parse_condition.hpp>
+#include <seqan3/range/view/persist.hpp>
 #include <seqan3/test/tmp_filename.hpp>
 
 using namespace seqan3;
 
+const std::string VERSION_STR = std::to_string(SEQAN3_VERSION_MAJOR) + "."
+                                + std::to_string(SEQAN3_VERSION_MINOR) + "."
+                                + std::to_string(SEQAN3_VERSION_PATCH);
+
```
h-2: we already have a string for this in version.hpp :point_up:

PR #543, opened 2018-11-07
[fix] clang: concept as typename is not allowed

marehr:


marehr:
upps; duplicate of #542

PR #544, opened 2018-11-07
[fix] clang: 'decltype(auto)' cannot be combined with other type specifiers

marehr:
```
/seqan3/include/seqan3/io/detail/record.hpp:212:1: error: 'decltype(auto)' cannot be combined with other type specifiers
decltype(auto) const get_or(record<field_types, field_ids> const & r, or_type && or_value)
^
```

See https://stackoverflow.com/questions/27766601/why-cannot-form-reference-to-decltypeauto/27766740#27766740


smehringer:
I really wonder why the tests don't fail... I remember that it did not work without the decltype(auto)

PR #547, opened 2018-11-07
[fix] clang: 'auto' not allowed in function prototype

marehr:
```c++
seqan3/include/seqan3/alignment/matrix/alignment_matrix_formatter.hpp:361:34: error: 'auto' not allowed in function prototype
    static std::string as_string(auto && entry) noexcept
                                 ^~~~
```

### DIFF
--- a/include/seqan3/alignment/matrix/alignment_matrix_formatter.hpp
+++ b/include/seqan3/alignment/matrix/alignment_matrix_formatter.hpp
@@ -358,7 +358,8 @@ private:
     }
 
     //!\brief Convert a matrix entry into a std::string
-    static std::string as_string(auto && entry) noexcept
+    template <typename entry_type>
+    static std::string as_string(entry_type && entry) noexcept
     {
         std::stringstream strstream;
         debug_stream_type stream{strstream};


h-2:
Hm, do you understand what the problem here is?

marehr:
nop, I wanted to here your opinion

FirstLoveLife:
Some discussions on stackoverflow https://stackoverflow.com/questions/30665506/auto-not-allowed-in-function-prototype-with-clang

> clang is correct in the sense that we don't yet have auto parameters. Concepts lite may bring those, but C++14 doesn't have them.

BTW, gcc will give a warning if concept is not turned on: 

```
a.cpp:3:30: warning: use of ‘auto’ in parameter declaration only available with -fconcepts
 static std::string as_string(auto&& entry) noexcept
```

h-2:
@FirstLoveLife Thanks for the comments!

In fact we are using the clang-branch that has WIP concepts support:
https://github.com/saarraz/clang-concepts

`auto` parameters work in other places with this branch, I think.

h-2:
> auto parameters work in other places with this branch, I think.

Actually, @marehr can you verify this?

marehr:
We normally use `auto` as parameter only in lambdas.

PR #550, opened 2018-11-08
[fix] clang: problems with fold-expressions

marehr:
```c++
/seqan3/include/seqan3/io/stream/parse_condition_detail.hpp:111:53: error: expected ')'
    (condition_head_t::msg + ... + (constexpr_string{" "} + condition_ts::msg)) +
                                                    ^
/seqan3/include/seqan3/io/stream/parse_condition_detail.hpp:111:36: note: to match this '('
    (condition_head_t::msg + ... + (constexpr_string{" "} + condition_ts::msg)) +
                                   ^
```

### DIFF
--- a/include/seqan3/io/stream/parse_condition_detail.hpp
+++ b/include/seqan3/io/stream/parse_condition_detail.hpp
@@ -108,8 +108,10 @@ template <char op, typename condition_head_t, typename ...condition_ts>
 inline constexpr_string constexpr condition_message_v
 {
     constexpr_string{"("} +
-    (condition_head_t::msg + ... +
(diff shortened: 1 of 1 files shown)


h-2:
Wow, this is pure C++17, not even concepts-related. Can you try the following:

```c++
(condition_head_t::msg + ((constexpr_string{" "} + condition_ts::msg) + ...))
```

That might work... if not I really think this is a bug. And the workaround is a lot less readable :(

If the above does not work, I would report a bug and temporarily do:
```cpp
#ifdef __clang__ // reference to bug report
hacky-hack solution
#else

marehr:
Okay, will do

marehr:
I think the main problem is that he is confused by `{}` initialization

marehr:
This seems to be fixed in the most recent clang-concepts.

PR #535, opened 2018-11-01
Refactoring of alphabets

h-2:
## major refactorisation of all alphabets

- there are now alphabet base classes that our alphabets inherit from, this reduces code duplication
- alphabets are no longer aggregate types fixing some weird issues
- our alphabets are still PODs although this is not required by the alphabet_concept anymore
- `constexpr` default construction has been fixed everywhere
- cartesian composition no longer inherits pod_tuple, but actually compresses into a single rank value
  - `gapped<dna5>` now consumes one byte, not two
  - all operations are now constant time, not linear in the number of elements
- introduction of alphabet_proxy, a generic proxy base that also models the concepts of it's emulated type
- fixed a heap of bugs I found on the way

## Reviews:

- @joshuak94 can you review the changes to the mask alphabet and gap?
- @sarahet can you review the changes regarding the amino acids? It's not much ;)
- @joergi-w can you review the changes to the structure alphabets and the dna alphabets?
- @mariehoffmann can you review the changes to the qualities?
- @smehringer @marehr Can you review rest, in particular the changes to cartesian_composition and the alphabet_proxy?

Follow-Up TODOs:
  - [ ] investigate the two ICEs
  - [ ] make union_composition require the same char_type on alternatives
  - [ ] rename `composition` module to `composite`
  - [ ] rename `cartesian_composition` to `alphabet_tuple`
  - [ ] rename `union_composition` to `alphabet_variant`
  - [ ] implement `alphabet_any`
  - [ ] switch all alphabets from enum-notation to using a char literal (only done for nucleotides)

PR #552, opened 2018-11-12
move travis from ubuntu14.04 to ubuntu16.04, add GCC8

h-2:
* switch from Ubuntu 14.0.4 to 16.04
* add a GCC8 test
* remove test of benchmark in debug

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: required
-dist: trusty
+dist: xenial
 language: cpp
 
 git:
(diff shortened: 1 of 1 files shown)


marehr:
Is recently looked if travis has a new ubuntu version. And the most that I found is that it was still in beta and that it might not always work. Is there some announcement that made this official?

h-2:
It's on the regular page: https://docs.travis-ci.com/user/reference/overview/

marehr:
https://blog.travis-ci.com/2018-11-08-xenial-release

okay, because I looked into it two weeks ago.

h-2:
They have also fixed long-standing issues with mac so we should be able to add mac, as well, but I wanted to get this in first to see if it resolves the issues with #535

marehr:
And they have windows support :fearful: 

Let's merge :full_moon_with_face:

PR #557, opened 2018-11-12
Updated concept names

MitraDarja:
#520

I didn't change concepts which had a connection to std::range.

Issue #520, opened 2018-10-26
Refactor our concept names to be CamelCase


h-2:
Since it touches a lot files, this shouldn't be attempted before #498 is merged.

Our concept names are currently `seqan3::foo_bar_concept`, but to be more in line with the standard library, they should be called `seqan3::FooBar` in the future. Someone needs to make lots of replacements!

While we are at it, I would like to change some names, beyond just case:
```
seqan3::nucleotide_concept -> seqan3::NucleotideAlphabet
seqan3::aminoacid_concept -> seqan3::AminoacidAlphabet
... (the alphabets should end in "...Alphabet")

seqan3::const_iterable_concept -> seqan3::ConstIterableRange
```

Either of @joshuak94 or @MitraDarja could have a look at this!

smehringer:
I like the name changes for alphabets :+1: (although I liked the `concept`-naming more to begin with :laughing:) 

> `seqan3::const_iterable_concept -> seqan3::ConstIterableRange`

This fits nicely with the Alphabet changes but it might be inconsistent with concept names where the object type (range in your case for const_iterable) is not known. e.g. what to do with `floating_point_concept -> FloatingPoint` vs `floating_point_concept -> FloatingPointValue`?

rrahn:
@smehringer @h-2 @MitraDarja what is the state here? Can you make a list of modules already done/need refactorisation? This needs to be finished before the release.

MitraDarja:
@rrahn I will work on this again, when I am done with fixing doxygen warnings due to missing default comments. 

Done:
aminoacid_concept -> AminoacidAlphabet
nucleotide_concept -> NucleotideAlphabet
quality_concept -> QualityAlphabet
semi_constexpr_alphabet_concept -> SemiConstexprAlphabet
constexpr_semi_alphabet_concept -> ConstexprSemialphabet
constexpr_alphabet_concept -> ConstexprAlphabet
semi_alphabet_concept -> Semialphabet
alphabet_concept -> Alphabet
rna_structure_concept -> RnaStructureAlphabet
char_adaption_concept -> CharAdaptationAlphabet
uint_adaption_concept -> UintAdaptionAlphabet
config_element_concept -> ConfigElement
cereal_archive_concept -> CerealArchive
cereal_output_archive_concept-> CerealOutputArchive
cereal_input_archive_concept -> CerealInputArchive
cerealisable_concept -> Cerealisable
[...]

PR #558, opened 2018-11-12
[fix] clang: use source-level subsumption

### LINKED ISSUE #524: [clang-concepts] normalization of constraints
Code like:

https://github.com/seqan/seqan3/blob/147a7111c16801dba01a32fd7d9301b9e9eaf044/include/seqan3/core/metafunction/range.hpp#L201-L222

and

https://github.com/seqan/seqan3/blob/147a7111c16801dba01a32fd7d9301b9e9eaf044/include/seqan3/core/metafunction/range.hpp#L242-L255

seem to be not concept-ts conforming.

I opened a discussion at https://github.com/saarraz/clang-concepts/issues/23 where the concept implementer  said that this is not standard conforming.

A suggested fix for this situation can also be found in https://github.com/saarraz/clang-concepts/issues/23.

marehr:
fixes #524; a problem with normalization of constaints

### DIFF
--- a/include/seqan3/core/metafunction/range.hpp
+++ b/include/seqan3/core/metafunction/range.hpp
@@ -231,7 +231,9 @@ using innermost_value_type_t = typename innermost_value_type<t>::type;
  * returns.
  */
 template <typename t>
+//!\cond
     requires detail::has_value_type<t>
+//!\endcond
 constexpr size_t dimension_v = 1;
 
 //!\cond
--- a/include/seqan3/core/metafunction/range.hpp
+++ b/include/seqan3/core/metafunction/range.hpp
@@ -231,13 +231,8 @@ using innermost_value_type_t = typename innermost_value_type<t>::type;
  * returns.
  */
 template <typename t>
-constexpr size_t dimension_v = 0;
-
-template <typename t>
-//!\cond
     requires detail::has_value_type<t>
-//!\endcond
-constexpr size_t dimension_v<t> = 1;
+constexpr size_t dimension_v = 1;
 
 //!\cond
 template <typename t>
--- a/include/seqan3/core/metafunction/range.hpp
+++ b/include/seqan3/core/metafunction/range.hpp
@@ -57,6 +57,16 @@
 //NOTE(h-2): for the range overloads we explicitly forbid that the type is iteratoer
 // because some types are actually both (e.g. std::directory_iterator)
 
+namespace seqan3::detail
+{
+
+//!\cond
+template <typename t>
+concept has_value_type = requires { typename value_type_t<remove_cvref_t<t>>; };
+//!\endcond
+
+} // namespace seqan3::detail
+
 namespace seqan3
 {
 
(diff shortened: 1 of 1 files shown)


marehr:
Since #557 addresses #520; What should the Concept name be in this case?

include/seqan3/core/metafunction/range.hpp
```diff
@@ -180,20 +190,20 @@ struct size_type<rng_t>
...
+struct innermost_value_type<t>
 {
     //!\brief The return type (recursion not shown).
     using type = value_type_t<remove_cvref_t<t>>;
 };
 
 //!\cond
 template <typename t>
-    requires requires (t)
+    requires detail::has_value_type<t> && requires
```
h-2: why not change this to just
```cpp
requires detail::has_value_type<t> && detail::has_value_type<value_type_t<t>>
```
Or does this fail because of GCC not lazy instantiating? Then you could still do:
```cpp
requires detail::has_value_type<t> && detail::has_value_type<detail::type_trait_or_t<value_type<t>, void>
```

include/seqan3/core/metafunction/range.hpp
```diff
@@ -220,18 +230,20 @@ using innermost_value_type_t = typename innermost_value_type<t>::type;
  * Attention, this metafunction implicitly removes cv-qualifiers and reference from the types it recurses on and
  * returns.
  */
+template <typename t>
+constexpr size_t dimension_v = 0;
+
```
h-2: Why do you want to add this anchor that makes the statement valid for types where it makes no sense? Is this required for something?

include/seqan3/core/metafunction/range.hpp
```diff
@@ -57,6 +57,16 @@
 //NOTE(h-2): for the range overloads we explicitly forbid that the type is iteratoer
 // because some types are actually both (e.g. std::directory_iterator)
 
+namespace seqan3::detail
+{
+
+//!\cond
+template <typename t>
+concept has_value_type = requires { typename value_type_t<remove_cvref_t<t>>; };
```
smehringer: 1) Why the extra concept? I remember from one of the meeting C++ talks that `requires requires` clauses are not recommended but is this the reason or was there a specialization issue?

2) If we do have an extra concept shouldn't this go into `/core` and be used in all similar places?

marehr: 1.
    Here, the problem is that we use the overload-resolution property of concepts. 

    If we don't use a single concept `detail::has_value_type` here, the compiler does not now that both statements are the same (the definition of "same" is if the same line number in the same file of a concept matches) and `innermost_value_type` would not see that the second definition is a more specialized version of the first one.

    ```c++
    template <typename t>
        requires detail::has_value_type<t>
    struct innermost_value_type

    template <typename t>
        requires detail::has_value_type<t> && detail::has_value_type<value_type_t<remove_cvref_t<t>>>
    struct innermost_value_type<t>
    ```

2. The extra concept is only local, therefore in the same file. If we need that in other file we can think about the location.

marehr:
@h-2  ping

PR #559, opened 2018-11-13
add head file

FirstLoveLife:
`std::logic_error` needs `<stdexcept>`

### DIFF
--- a/include/seqan3/alignment/exception.hpp
+++ b/include/seqan3/alignment/exception.hpp
@@ -39,6 +39,8 @@
 
 #pragma once
 
+#include <stdexcept>
+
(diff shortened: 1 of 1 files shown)


marehr:
Thank you for your contribution!

@h-2, @smehringer It is weird that the header test didn't catch this, because it should have.

HAHA What a stupid mistake; Who sees my mistake? (Note that `$header` is `seqan3/alignment/exception.hpp` in this case)

https://github.com/seqan/seqan3/blob/0fdd7cd47752373e6bb13dd230cff1757bb5d678/test/header/CMakeLists.txt#L93-L97

PR #566, opened 2018-11-16
[NOP] Adds googlemock.

rrahn:
Allows to use googlemock features.


h-2:
What does it do?

rrahn:
If adds functionality that gtest did not provide, e.g. testing elements of a container like this:

```cpp
using ::testing::GE;
using ::testing::_;
using ::testing::ElementsAre;

ASSERT_THAT(std::vector{0, 1, 2, 3}, ElementsAre(0, _, GE(0), 3))
```
Cheks if first element is 0, any value at the second, the third is greater or equal 0 and the last is 3.

But it comes with even more poweful features to create mock classes and interfaces that can be tested in unit tests.

PR #370, opened 2018-08-15
Feature/banded alignment

joergi-w:
Adds the banded alignment configuration and band_static element.
Solves issues #338 and #339.

PR #567, opened 2018-11-16
Feature/rework configuration

rrahn:
Major rework of the configuration system to reduce a lot of boiler plate code and maintain a more leightweight code basis.
Makes querying the configuration more intutive.
Also it will make it a whole lot aeasier to document the configurations for algorithms.

~This is still WIP, but I want to share it early, as it will change a lot of the exisiting configurations.~

### Blocked
* <s>#566</s>
* ~#370~
* ~#613~

###  TODO
* [x] Adapt existing alignment configurations
* [x] Adapt existing search configurations
* [x] reevaluate the replace_with function
* [x] reevaluate the deferred_config (maybe this is better done within the implementation of the algorithm.)
* [x] remove some unused code in configuration
* [x] check if all new files have a short header.
* [x] make some more types aggregate
* [x] Check commented block in test
* [x] Revert changes in scoring scheme implementation
* [x] make gap scheme concept work with score function only.
* [x] double view::persist in align_pairwise
* [x] Remove is_algorithm_configuration unary type trait.
* [x] rename the aligned_ends using `first` and `second` prefix

### State

Except for some smaller things like fixing doxygen or removing unused code this is in a final state.
I suggest going through commit by commit as they are separated per configuration element and some basic units. Also please note, that the docu will for this PR only cover very basic coverage. For the upcoming sprints I plan to make the documentation more verbose, but it depends on other stuff too.


include/seqan3/core/algorithm/configuration.hpp
```diff
@@ -44,92 +44,37 @@
...
- * \snippet test/snippet/core/algorithm/configuration.cpp access
- * Note, that the type based get should be used with caution. In the typical usage scenario, the user of the
- * configuration works with config adaptor types. Only these adaptors know what exact type was added to the
- * configuration.
- * To allow algorithm implementors to still access data for specific properties independent of their type and position,
- *  every algorithm will expose a enum based get interface, that allows to access the element associated with the
- * enum identifier. The details can be found in the respective algorithm description.
+ * This class provides a unified interface to create and query such
+ * configurations for a specific algorithm. It extends the standard tuple interface with some useful functions to modify
+ * and query the user configurations.
```
smehringer: Can you reference the detailed description of the module?

rrahn:
@h-2 I changed the requirements for `score` member function for scoring_schemes in 3f7b9e6 to also allow score computation if char types are passed to the function. This change makes checking for the concept feasible even if the alphabet type is not known (wich is the case for the configuration object). The alternatives I saw, were to either provide an additional template parameter to the config element which has to be specified by the user, which makes the code unnecessarily complex. To not constrain the template would be another way, but wrong instantiations would be triggered deeper in the code making error handling complicated. This change also implies, that user provided scoring schems also need to take this into account. But in my opinion, this is the least confusing option for the user as most of them won't deal with user defined scoring schemes. If you agree, I would update the concept description accordingly.

h-2:
I don't fully understand why this is necessary... only so that you can perform a concept check in the config? Why don't you just expose the alphabet from the scoring scheme as a member type?

h-2:
>But that would require that the scoring scheme concept requires that the scoring scheme exposes this type, 

Yes

> which would be a cycling dependency.

Why? Maybe I am stupid today :see_no_evil:  But you could just change the concept from:
```cpp
template <typename t, alphabet_concept alphabet_t, alphabet_concept alphabet2_t = alphabet_t>
concept scoring_scheme_concept = requires (t scheme,
                                                alphabet_t const alph1,
                                                alphabet2_t const alph2)
{
    { scheme.score(alph1, alph2) } -> std::CommonReference<typename std::remove_reference_t<t>::score_type>;
    { scheme.score(alphabet_t{}, alphabet2_t{}) }
        -> std::CommonReference<typename std::remove_reference_t<t>::score_type>;
};
```
to
```cpp
template <typename t, alphabet_concept alphabet_t, alphabet_concept alphabet2_t = alphabet_t>
concept scoring_scheme_concept = requires (t scheme,
                                                alphabet_t const alph1,
                                                alphabet2_t const alph2)
{
    typename t::alphabet_t;
    { scheme.score(alph1, alph2) } -> std::CommonReference<typename std::remove_reference_t<t>::score_type>;
    { scheme.score(alphabet_t{}, alphabet2_t{}) }
        -> std::CommonReference<typename std::remove_reference_t<t>::score_type>;
};
```
And then access that from the config checks? 

But OTOH we want to allow scoring schemes that are not alphabet specific, too :thinking: 

The reason I am making a big deal about this is that we deliberately have very strong typing for our alphabets throughout the library. There is no implicit conversion from char to alphabet **anywhere**, but this would introduce it and just for syntactical concept checks.

Maybe instead define `seqan3::defail::any` which does nothing except that it is convertible to any type implicitly (`std::any` is not). Then you can use that for concept checks, because our alphabet specific interfaces inside the scoring schemes will try to convert and that will work so it will verify the other requirements.

h-2:
> Now I am accessing a member type of something I only know it has it if it's concept was satisfied. Am I misunderstanding something here? thinking

Only if you pass in the scheme as an incomplete type... was that your plan?

> Mhmm, if I recall correctly we only disallow implicit/explicit casting in order to maintain the property for POD types, right?

No :scream: For safety! And to avoid confusion between the rank representation and the char representation. If your interface takes `foo(char c)` you can call `f(3)`, which is probably not what you want. Your interface would allow this and silently convert the number  `3` to some non-printable char which will silently be converted to `dna4::A` :boom: 

> And no it is not just for concept checking, but also a realistic desing for the API. I can imagine that programmers want to use libraries together. So if they use a c-library which exposes some sort of functionality it is likely that alphabets are simply chars. If they want to use this with our alignment then it will work without any wrapping or conversion etc.
> I mean our types should be strong but the algorithms should be generic and allow this kind of intercation.

We support people using `char` as the alphabet type if they want to that's why we have the adaptations. This is not want this PR is about, this PR allows silently converting `char` to any other alphabet under the hood, or not?

> Possible yes, but the question is, whether the concept should reflect the usage from above.

We do support this. It's just the nucleotide_scoring scheme that doesn't take `char`. You can easily define your own and the `identity_scheme` you once proposed would work, too, as it is independent of alphabet.

>In any case setting an incompatible scoring scheme cannot be checked at the configuration level but first in the call of the align function.

Yes, the algorithm needs to check that the alphabets implicitly and explicitly set on input data and various config elements are compatible.

> I am still more convinced about open up a generic interface. 

I don't see why our interfaces are not generic. The algorithm should support any type that models alphabet_concept. Certain scoring schemes that have certain requirements and make certain assumptions may only work on a subset of types. You are free to define other scoring schemes. If you want to use existing ones, you need to convert to that respective type. But you need to that for a multitude of other interfaces, as well.

> Alternatively, we could offer a view::assign_char<alphabet_t> (not sure we have this already) that converts a char range into the respective alphabet.

We have that, it's called `view::char_to<alphabet_t>` :smiley:

h-2:
> Let people explicitly use conversion so they know what they are doing.

Yes, I think this is really important for our other interfaces, as well. They should properly read their data or create a view which costs almost nothing.

> Change the concept to expose for the alphabet_type
> Use that type and live with the still readable error message.

The problem here is that, as we discussed before, there maybe be schemes that have no fixed alphabet type. For the existing ones it would just be choosing an example alphabet, anyway, or not? I am sorry that I didn't get this point earlier.

rrahn:
> Yes, I think this is really important for our other interfaces, as well. They should properly read their data or create a view which costs almost nothing.

agreed, I was to concerned about the safety of the config element.

>The problem here is that, as we discussed before, there maybe be schemes that have no fixed alphabet type. For the existing ones it would just be choosing an example alphabet, anyway, or not? I am sorry that I didn't get this point earlier.

Actually I am not happy with any of the suggested solutions.
Also the `any`type, which would be ok for our types. 
But it does not work if the user defined type doesn't check for convertible types, which is something we can't enforce. This wouldn't invalidate the scoring scheme concept.

So I am back to square one.
Actually I will for now remove that constraint completely and live with the fact, that it will be tested once the align function is called. Probably with a static assert, that gives a more specific error message. This problem can then be revisited later.

marehr:
@rrahn Is this still blocked?

marehr:
I added #370 as blocker.

rrahn:
@marehr this is all still in WIP and I am aware of #370, which will be merged before anyway.

include/seqan3/core/algorithm/all.hpp
```diff
@@ -38,13 +38,68 @@
...
-//!\cond DEV
+
 /*!\defgroup algorithm Algorithm
  * \ingroup core
  * \brief Contains core functionality used to configure algorithms.
- * \todo Write detailed description about usage of this module.
+ *
+ * \details
+ *
+ * In SeqAn there are many, e.g. alignment or search interfaces, that can be configured through
```
smehringer: there are many.... cool colleagues :sunglasses: ?

test/snippet/core/algorithm/configuration.cpp
```diff
@@ -1,88 +1,96 @@
...
-struct bar
+enum struct my_id : int
 {
-    float value = 0;
+    bar_id,
+    foo_id
 };
 
-struct foo
+struct bar : public pipeable_config_element
```
smehringer: Since `bar` has a non-default constructor it should be a class (you also talk about classes in your description). Same goes for foo then.

test/snippet/core/algorithm/configuration.cpp
```diff
@@ -1,88 +1,96 @@
...
 {
-    int value = 0;
+    static constexpr my_id id{my_id::bar_id};
+
+    bar()                        = default;
```
smehringer: can't you implement the default constructors in `pipeable_config_element` and then just `using` them here?

test/snippet/core/algorithm/configuration.cpp
```diff
@@ -1,88 +1,96 @@
...
+
+    bar(float v) : value(v)
+    {}
+
+    float value;
 };
 
-struct with_foo_adaptor : public detail::configuration_fn_base<with_foo_adaptor>
+template <typename t>
+struct foo : public pipeable_config_element
```
smehringer:
```suggestion
struct foo : public pipeable_config_element  // define a templatized configuration element 'foo'
```

test/snippet/core/algorithm/configuration.cpp
smehringer:
```suggestion
struct bar : public pipeable_config_element // define a configuration element 'bar'
```

include/seqan3/core/algorithm/all.hpp
```diff
@@ -38,13 +38,68 @@
...
+ *
+ * The basis of any algorithm configuration are configuration elements. These are objects that handle a specific
+ * setting and must satisfy the seqan3::detail::config_element_concept. The following snippet demonstrates a
+ * basic setup for such configuration elements.
+ *
+ * \snippet test/snippet/core/algorithm/configuration.cpp configuration_setup
+ *
+ * Here, two classes with the name `bar` and `foo` are created. They can be normal or template classes and must
+ * inherit from seqan3::pipeable_config_element and contain a member with the name `value` to satisfy the respective
+ * concept. The separate `value` member is used for a proper encapsulation from the actual setting parameter.
```
smehringer: link the actual concept?

include/seqan3/core/algorithm/all.hpp
```diff
@@ -38,13 +38,68 @@
...
+ * For example the alignment algorithms require a scoring scheme, but the scoring scheme itself should not be pipeable
```
smehringer: > The separate `value` member is used for a proper encapsulation from the actual setting parameter. For example the alignment algorithms require a scoring scheme, but the scoring scheme itself should not be pipeable with other settings.

I do not understand what you are trying to tell here... what is a setting parameter? Is the scoring scheme a configuration element? This is very specific and does not blend in well the explanations of what is going on so far.

test/snippet/core/algorithm/configuration.cpp
```diff
@@ -1,88 +1,96 @@
+#include <iostream>
+
+//! [configuration_setup]
 #include <seqan3/core/algorithm/configuration.hpp>
+#include <seqan3/core/algorithm/pipeable_config_element.hpp>
 
 using namespace seqan3;
 
-struct bar
+enum struct my_id : int
```
smehringer:
```suggestion
enum struct config_element_id : int // this enum will be needed for compatibility checks of config elements
```

include/seqan3/core/algorithm/all.hpp
```diff
@@ -38,13 +38,68 @@
...
+ * with other settings.
+ *
+ * In addition an enum type was defined that will be used later to allow for compatibility checks when combining
+ * settings in a configuration object. This enum assigns every configuration element a unique id.
```
smehringer: > This enum assigns every configuration element a unique id.

This enum stores a unique id for every configuration element. The id must be set in the respective classes.

include/seqan3/core/algorithm/all.hpp
```diff
@@ -38,13 +38,68 @@
...
+ * To provide compatibility checks the `seqan3::detail::compatibility_table` must be overloaded for the specific
```
smehringer: start a new paragraph here as it refers to the next snippet and end the sentence with `:`

test/snippet/core/algorithm/configuration.cpp
```diff
@@ -1,88 +1,96 @@
...
-    {
-        return std::forward<configuration_t>(cfg).push_front(bar{new_v});
-    }
-
-    template <typename configuration_t>
-    constexpr auto invoke(configuration_t && cfg) const
-        requires detail::is_algorithm_configuration_v<remove_cvref_t<configuration_t>>
     {
-        return std::forward<configuration_t>(cfg).push_front(bar{2});
+        {0, 1},
```
smehringer: This looks much nicer than the code before :+1: 
I would add some more comments so the code is more self-explanatory 
```suggestion
        {0, 1}, // 'bar' cannot be combined with itself but with `foo`
```

test/snippet/core/algorithm/configuration.cpp
```diff
@@ -1,88 +1,96 @@
...
+        {1, 0}
```
smehringer:
```suggestion
        {1, 0} // 'foo' can be combined with `bar` but not with itself
```

include/seqan3/core/algorithm/all.hpp
```diff
@@ -38,13 +38,68 @@
...
+ * algorithm configuration.
+ *
+ * \snippet test/snippet/core/algorithm/configuration.cpp compatibility
+ *
+ * The type for the configuration element ids is used to overload the bool table. In the example above, both
+ * elements can be combined with each other but not with themselves, to avoid inconsistent settings.
```
smehringer: Does the table always need to be symmetric?

test/snippet/core/algorithm/configuration.cpp
```diff
@@ -1,88 +1,96 @@
...
-
+} // namespace seqan3::detail
+//! [compatibility]
 int main()
 {
-
+{
 //! [combine]
-auto my_cfg = detail::configuration<bar>{} | with_foo(1);  // my_cfg is now of type configuration<foo, bar>
+configuration my_cfg = bar{1.3} | foo<int>{{4}};  // my_cfg is now of type configuration<bar, foo<int>>
```
smehringer: why `{{4}}` and not just `{4}`?

include/seqan3/core/algorithm/all.hpp
```diff
@@ -38,13 +38,68 @@
...
+ * position-based and type-based `get` interfaces. The `get` interface was extended to also support
+ * template template types as input template parameters to query the correct element:
+ *
+ * \snippet test/snippet/core/algorithm/configuration.cpp get
+ *
+ * The get interface returns a reference to the stored configuration element. In some cases, e.g. the implementor
+ * of the actual algorithm, one wants to have an easy access to the actual value of the setting. Since, the
+ * configuration must not contain all possible configuration elements the seqan3::configuration provides a
+ * seqan3::configuration::value_or interface, which provides direct access to the value of the respective
+ * configuration element or uses a default value if the queried type is not contained in the configuration.
```
smehringer: > The get interface returns a reference to the stored configuration element. In some cases, e.g. the implementor  of the actual algorithm, one wants to have an easy access to the actual value of the setting. Since, the
  configuration must not contain all possible configuration elements the seqan3::configuration provides a
  seqan3::configuration::value_or interface, which provides direct access to the value of the respective
 configuration element or uses a default value if the queried type is not contained in the configuration.

Maybe shorter: 
The get interface returns a reference to the stored configuration element. In order to avoid an error if an optional confugartion element was not given, you can use the seqan3::configuration::value_or interface to give a default value:

test/snippet/core/algorithm/configuration.cpp
rrahn: pipeable is an empty base class. In general I would have loved to keep the config elements aggregate types but, then they become ugly to initialize because you need to give an empty construciton argument for the empty base class.

rrahn: Ok, while thinking about it I was making my live more complicated than it had to. I changed the base class to a crtp base. Further the class is now an aggregate so that in general configs are agregates to. This solved all boilerplate code with the constructors and such.

test/snippet/core/algorithm/configuration.cpp
rrahn: I don't know. I mean the snippet is inserted in a documentation test and there is a proper documentation for it.

include/seqan3/core/algorithm/all.hpp
rrahn: yes. Why would you allow local with global but not global with local. The order should not matter, becuase this would be confusing for the users and the error detection very cumbersome.

include/seqan3/core/algorithm/all.hpp
rrahn: Well, except for bugs in the documentation I would prefer not to touch this stuff a lot. My plan is for the upcoming sprints to refactor the documentation as well, when the whole systems starts to grow together.

test/snippet/core/algorithm/configuration.cpp
smehringer: nice

test/snippet/core/algorithm/configuration.cpp
smehringer: Hm, yes but I think usually people like to look at examples and skip the text (unless they really want to dive into the topic), so I always prefer snippets to be mostly understandable without the wrapped text. But this is your call then.

include/seqan3/core/algorithm/all.hpp
smehringer: ok

include/seqan3/core/algorithm/configuration_utility.hpp
```diff
@@ -0,0 +1,153 @@
...
+    //!\brief Helper typedef.
+    using index_type        = std::underlying_type_t<algorithm_id_type>;
+
+    //!\publicsection
+    //!\brief The result of checking the compatibility.
+    static constexpr bool value =
+        (std::is_same_v<algorithm_id_type, remove_cvref_t<decltype(compare_types::id)>> && ...) &&
+        (compatibility_table<algorithm_id_type>[static_cast<index_type>(query_t::id)]
+                                             [static_cast<index_type>(compare_types::id)] && ...);
+};
```
h-2: I realise this would make reading it a lot more difficult, but we said that we want to make the value metafunction inherit std::integral_constat... can you refactor this a little so it works?

include/seqan3/core/algorithm/configuration_utility.hpp
```diff
@@ -0,0 +1,153 @@
...
+template <template <typename ...> typename query_t>
+struct is_same_configuration_f
+{
+    /*!\brief A type template that evaluates to std::true_type if the given type is a specialization of `query_t`,
+     *        otherwise std::false_type.
+     * \tparam compare_type The type to compare against `query_t`.
+     */
+    template <typename compare_type>
+    using invoke = is_type_specialisation_of<compare_type, query_t>;
+};
```
h-2: this cruft is just for the meta library?

include/seqan3/core/algorithm/configuration_utility.hpp
```diff
@@ -0,0 +1,153 @@
...
+template <typename ...config_elements_t>
+struct is_algorithm_configuration<seqan3::configuration<config_elements_t...>> : std::true_type
+{};
+//!\endcond
+
+/*!\brief Helper variable template for seqan3::detail::is_algorithm_configuration.
+ * \ingroup algorithm
+ */
+template <typename object_t>
+inline constexpr bool is_algorithm_configuration_v = is_algorithm_configuration<object_t>::value;
```
h-2: here again, can't we just use is_type_specialisation_of wherever this would be used?

include/seqan3/core/algorithm/pipeable_config_element.hpp
```diff
@@ -0,0 +1,150 @@
...
+     * \details
+     *
+     * Effectively calls seqan3::configuration::push_back on a temporary created seqan3::configuration from the
+     * left operand.
+     */
+    template <typename other_derived_t, typename other_value_t>
+    friend constexpr auto operator|(pipeable_config_element && lhs,
+                                    pipeable_config_element<other_derived_t, other_value_t> && rhs)
+    {
+        return configuration{static_cast<derived_t &&>(lhs)}.push_back(static_cast<other_derived_t &&>(rhs));
```
h-2: does push_back on a configuration modify the configuration or does it return a new one? I suspect the latter, but the syntax strongly suggests the former :thinking:

include/seqan3/core/algorithm/pipeable_config_element.hpp
```diff
@@ -0,0 +1,150 @@
...
+    template <typename ...configs_t>
+    friend constexpr auto operator|(configuration<configs_t...> && lhs,
+                                    pipeable_config_element const & rhs)
+    {
+        return std::move(lhs).push_back(static_cast<derived_t const &>(rhs));
+    }
+
+    //!\overload
+    template <typename ...configs_t>
+    friend constexpr auto operator|(configuration<configs_t...> const & lhs,
```
h-2: these are in fact functions that would usually be perceived as being part of the configuration's interface, not the config element's interface, or not?

include/seqan3/search/algorithm/configuration/max_error_common.hpp
```diff
@@ -57,6 +57,10 @@ template <typename value_t>
 struct total : detail::strong_type<value_t, total<value_t>, detail::strong_type_skill::convert>
 {
     using detail::strong_type<value_t, total<value_t>, detail::strong_type_skill::convert>::strong_type;
+
+    //!\privatesection
+    //!\brief Internal id to check for consistent configuration settings.
+    static constexpr std::integral_constant<uint8_t, 0> _id{};
```
h-2: should these id maybe be enum values instead?

test/unit/alignment/band/static_band_test.cpp
```diff
@@ -0,0 +1,34 @@
...
+    EXPECT_TRUE((std::is_same_v<decltype(bs), static_band<int>>));
+}
+
+TEST(static_band, various_integer_types)
+{
+    static_band b{lower_bound{static_cast<int8_t>(-5)}, upper_bound{static_cast<int8_t>(5)}};
+    EXPECT_TRUE((std::is_same_v<decltype(b), static_band<int8_t>>));
+
+    static_band b2{lower_bound{static_cast<int16_t>(-5)}, upper_bound{static_cast<int16_t>(5)}};
+    EXPECT_TRUE((std::is_same_v<decltype(b2), static_band<int16_t>>));
```
h-2: Is the type of these parameters down-the-line dependent on the input type? Why not always use uint64_t to reduce template instantiations? I guess you use different internal types for the kernel anyway?

test/unit/alignment/configuration/align_config_aligned_ends_test.cpp
```diff
@@ -0,0 +1,458 @@
...
+// ARE DISCLAIMED. IN NO EVENT SHALL KNUT REINERT OR THE FU BERLIN BE LIABLE
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+// DAMAGE.
+//
+// ============================================================================
```
h-2: shorter copyright headers

test/unit/alignment/configuration/align_config_mode_test.cpp
```diff
@@ -34,73 +34,56 @@
...
+//         return std::ignore;
+//     }
+// }
+//
+// TYPED_TEST(align_cfg_mode_test, construction_from_variable)
+// {
+//     configuration cfg{align_cfg::mode{get_inline_variable<TypeParam>()}};
+//     EXPECT_EQ((std::is_same_v<std::remove_reference_t<decltype(get<align_cfg::mode>(cfg).value)>,
+//                               TypeParam>), true);
+// }
```
h-2: ?

include/seqan3/alignment/band/static_band.hpp
```diff
@@ -117,35 +121,42 @@ struct band_static
...
+static_band(lower_bound<value_t>, upper_bound<value_t>) -> static_band<value_t>;
 //!\}
 
 } // namespace seqan3
+
+namespace seqan3::detail
+{
+//!\cond
+template <std::Integral value_t>
+struct is_band_config<static_band<value_t>> : public std::true_type
```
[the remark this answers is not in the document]
rrahn: Well, this won't be the only band implementation. Making `is_template_specialisation_of_v` not really usable. Nor, am I sure if there should  be a band concept. So the alternative is to make it a unary type trait. Albeit, the name should not include config but only refer to the band type.

include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
```diff
@@ -0,0 +1,600 @@
...
+     * \tparam    _value_t Either a std::integral_constant or bool.
+     * \param[in] v        The value to wrap.
+     */
+    template <typename _value_t>
+    explicit constexpr seq_end_gap_base(_value_t const v) noexcept
+    {
+        if constexpr (std::is_same_v<_value_t, bool>)
+            value = v;
+        else
+            value = v();
```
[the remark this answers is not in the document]
rrahn: >You can still just assign inside the constructor. 

That's true, that should work since it is implicitly convertible to bool. 

> Altogether, I think two different overloards for the class template would be more readable and probably also save a lot of code, because they could be aggregates.

Splitting the mixin is already tracked as followup issue but it would not change the semantics. So it does not need be split now. In general I am relatively certain that I can make this an aggregate as is and would bring the spliting into another sprint.

include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
```diff
@@ -0,0 +1,600 @@
...
+class seq1_leading : public seq_end_gap_base<value_t>
+{
+    //!\brief An alias for the base type.
+    using base_t = seq_end_gap_base<value_t>;
+
+public:
+
+    //!\privatesection
+    //!\brief An internal id to allow consistency checks with other gap specifiers.
+    static constexpr std::integral_constant<uint8_t, 0> id{};
```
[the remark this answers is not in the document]
rrahn: No they are not API. That's why they are documented in a privatesection. They are for internal use only, but I want to avoid making them private and befriendnig with the classes that access them.

include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
```diff
@@ -0,0 +1,600 @@
...
+    constexpr seq1_leading(seq1_leading &&)                  noexcept = default;
+    constexpr seq1_leading & operator=(seq1_leading const &) noexcept = default;
+    constexpr seq1_leading & operator=(seq1_leading &&)      noexcept = default;
+    ~seq1_leading()                                          noexcept = default;
+
+    //!\brief Sets the wrapped value.
+    seq1_leading(value_t const v) noexcept : base_t{v}
+    {}
+    //!\}
+};
```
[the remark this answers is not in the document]
rrahn: Should be. I'll make it work.

include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
```diff
@@ -0,0 +1,600 @@
...
+// end_gaps
+// ----------------------------------------------------------------------------
+
+/*!\brief Wraps the sequence end-gap specifiers and provides ordered access to the respective values.
+ * \ingroup configuration
+ * \tparam ends_t A parameter pack containing at most 4 sequence end-gap specifier.
+ *
+ * \details
+ *
+ * A wrapper for providing ordered access to the end-gap specifiers independent of the input order.
```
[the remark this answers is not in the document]
rrahn: :scream: this was so much more confusing and adding boiler plate. This solution is so much cleaner and more readable. Making it dry? Maybe but not in this sprint.

include/seqan3/alignment/band/static_band.hpp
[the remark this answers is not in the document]
h-2: I don't fully get it, but it doesn't matter, if you know what you are doing :smile:

include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
```diff
@@ -0,0 +1,600 @@
...
+/*!\name Fixed end gaps configurations
+ * \brief These variables are pre-configured end gaps that are frequently used in pairwise sequence alignments.
+ * \relates seqan3::end_gaps
+ */
+
+//!\brief Disables all end-gaps, such that they are not penalized in the pairwise alignment.
+inline constexpr end_gaps all_ends_free{seq1_leading<std::true_type>{},
+                                        seq1_trailing<std::true_type>{},
+                                        seq2_leading<std::true_type>{},
+                                        seq2_trailing<std::true_type>{}};
```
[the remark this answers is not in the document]
h-2: Ok, I don't care much either way, but this is something we should document in the wiki prominently. Also we should not use `std::pair`, but `std::tuple` even for tuples of size two. And not use the member variables.

include/seqan3/alignment/configuration/detail.hpp
```diff
@@ -0,0 +1,86 @@
...
+{
+    aligned_ends,
+    band,
+    gap,
+    global,
+    max_error,
+    result,
+    scoring,
+    SIZE
+};
```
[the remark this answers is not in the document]
h-2: Hm, I feel like separating documentation from implementation always leads to documentation not being done at all. This is not a criticism of you, it's the same for me: the TODO-list is never empty and "improving documentation from a three month old commit" usually ends up at the bottom and stays there... just saying...

include/seqan3/alignment/scoring/gap_scheme.hpp
```diff
@@ -264,9 +264,9 @@ class gap_scheme
     //!\endcond
 
 private:
-    //!\brief The score per gap character.
-    score_t gap = 0;
-    //!\brief The score per sequence of gaps.
+    //!\brief The score per gap character. Defaults to -1.
+    score_t gap = -1;
```
joergi-w: Is this safe, if the user defines score_t as unsigned?
Or do you make sure that unsigned types are prohibited?

include/seqan3/core/algorithm/all.hpp
```diff
@@ -38,14 +38,68 @@
...
+ *
+ * \snippet test/snippet/core/algorithm/configuration.cpp compatibility
+ *
+ * The type for the configuration element ids is used to overload the bool table. In the example above, both
+ * elements can be combined with each other but not with themselves, to avoid inconsistent settings.
+ *
+ * ### Combining Configurations
+ *
+ * To enable easy creation of algorithm settings the seqan3::configuration supports a pipeable interface for the
+ * different configuration elements which is added through the seqan3::pipeable_config_element base class.
```
joergi-w: which **are** added

include/seqan3/core/algorithm/configuration.hpp
```diff
@@ -195,398 +154,440 @@ class configuration : public std::tuple<configs_t...>
...
-     *                        seqan3::detail::configuration.
-     * \tparam args_t A template parameter pack with the arguments applied to the configuration functor call.
-     *
-     * \param configuration The configuration to be extended.
-     * \param args         The argument pack to be passed to the functor.
-     * \returns A new seqan3::detail::configuration with the extended configuration.
+    /*!\brief Creates a new configuration object by recursively adding the configs from the tuple.
+     * \tparam    tuple_t The tuple from which to create a new configuration object; must model
+                          seqan3::detail::tuple_like_concept and must at least have 1 element.
+     * \param[in] tpl     The tuple to crate the configuration from.
```
joergi-w: create

include/seqan3/search/algorithm/configuration/detail.hpp
```diff
@@ -0,0 +1,80 @@
...
+ * auto max_total_errors = get<search_cfg::id::max_total_errors>(cfg);  // max_total_errors = 3;
+ * ```
+ */
+enum struct search_config_id : uint8_t
+{
+    //!\brief Identifier for max_errors configuration.
+    max_error,
+    max_error_rate,
+    output,
+    mode,
```
joergi-w: aligned_ends and band missing?

include/seqan3/search/algorithm/configuration/max_error_rate.hpp
```diff
@@ -35,144 +35,135 @@
...
-            (0 > insertion_v || insertion_v > 1) || (0 > deletion_v     || deletion_v > 1))
+        //TODO Make to for_each_value once it is in
+        [[maybe_unused]] auto dummy = ((base_t::value[(std::remove_reference_t<errors_t>::_id())] = errors.get()), ..., 0);
+
+        // check correct values.
+        ranges::for_each(base_t::value, [](auto error_elem)
         {
-            throw std::invalid_argument("Error rates must be between 0 and 1.");
-        }
+            if (0.0 > error_elem  || error_elem > 1.0)
```
joergi-w: Should we add an epsilon value to avoid failures due to inaccuracies?

test/unit/core/algorithm/configuration_mock.hpp
```diff
@@ -0,0 +1,86 @@
...
+// ARE DISCLAIMED. IN NO EVENT SHALL KNUT REINERT OR THE FU BERLIN BE LIABLE
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+// DAMAGE.
+//
+// ============================================================================
```
joergi-w: short header

include/seqan3/alignment/configuration/align_config_max_error.hpp
```diff
@@ -39,73 +39,20 @@
...
-    //!\brief Type alias used by meta::find_if
-    template <config_element_concept t>
-    using invoke = typename std::is_same<t, align_config_max_error>::type;
-};
-
-//!\brief Mapping from the seqan3::detail::align_config_max_error type to its corresponding seqan3::align_cfg::id.
-//!\ingroup configuration
-template <>
-struct align_config_type_to_id<align_config_max_error>
+class max_error : public pipeable_config_element<max_error, uint32_t>
```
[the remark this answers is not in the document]
rrahn: It is used for the edit distance computation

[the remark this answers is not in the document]
rrahn: The detailed documentation will follow in a separate PR

include/seqan3/alignment/configuration/align_config_mode.hpp
```diff
@@ -35,75 +35,54 @@
...
-    template <config_element_concept t>
-    using invoke = typename std::is_same<t, align_config_global>::type;
-};
 
-//!\brief Mapping from the seqan3::detail::align_config_global type to its corresponding seqan3::align_cfg::id.
+//!\brief Selects global alignment mode.
 //!\ingroup configuration
-template <>
-struct align_config_type_to_id<align_config_global>
+inline constexpr detail::global_alignment_type global_alignment;
```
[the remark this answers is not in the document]
rrahn: No, it is currently not implemented. These things will be added iteratively.

include/seqan3/alignment/pairwise/alignment_selector.hpp
```diff
@@ -73,21 +73,21 @@ struct determine_result_type
...
                                               score_type,
                                               alignment_coordinate>>{};
-            else if constexpr (get<align_cfg::id::output>(configuration_t{}) ==
-                               align_result_key::begin)
+            else if constexpr (std::Same<remove_cvref_t<decltype(get<align_cfg::result>(configuration_t{}).value)>,
+                                         with_begin_position_type>)
                 return align_result<type_list<uint32_t,
                                               score_type,
                                               alignment_coordinate,
                                               alignment_coordinate>>{};
```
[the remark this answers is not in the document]
rrahn: sure. I'll need to rebase anyway before your commit get's in.

include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
```diff
@@ -52,23 +52,43 @@
...
+        decltype(cfg.template value_or<align_cfg::aligned_ends>(align_cfg::end_gaps{align_cfg::none_ends_free}))>::
+            template get_static<0>();
+    requires std::remove_reference_t<
+        decltype(cfg.template value_or<align_cfg::aligned_ends>(align_cfg::end_gaps{align_cfg::none_ends_free}))>::
+            template get_static<1>();
 };
 
 template <typename align_config_t>
-concept global_config_concept = has_align_cfg_v<align_cfg::id::global, std::remove_reference_t<align_config_t>>;
+concept global_config_concept = requires (align_config_t & cfg)
```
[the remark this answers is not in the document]
rrahn: This is some detail stuff to check which version of Merys to run. This is not exposed to the public. So I think the naming is still fine with the domain it is used on?!

include/seqan3/alignment/scoring/gap_scheme.hpp
rrahn: Well if the user makes them unsigned, he needs to do it explicitly. So the user should know what he is doing

include/seqan3/search/algorithm/configuration/detail.hpp
rrahn: No this is the search configuration, not the alignment configuration.

include/seqan3/search/algorithm/configuration/max_error_rate.hpp
rrahn: I think this is fine as is. It is basically just moved to another location.

include/seqan3/alignment/scoring/scoring_scheme_base.hpp
```diff
@@ -230,16 +233,24 @@ class scoring_scheme_base
...
 
     //!\copydoc score
-    template <explicitly_convertible_to_concept<alphabet_t> alph1_t,
-              explicitly_convertible_to_concept<alphabet_t> alph2_t>
+    template <typename alph1_t,
+              typename alph2_t>
+    //!\cond
+        requires (explicitly_convertible_to_concept<alph1_t, alphabet_t> &&
+                  explicitly_convertible_to_concept<alph2_t, alphabet_t>)
+    //!\endcond
```
h-2: why these changes?

h-2: I think you can discard all changes in this file or not?

rrahn: Good spot. I was just removing the char stuff. The only thing that will change is that I like at least our scoring schemes to expose the `alphabet_t` It is not required by the concept, but it makes some unit tests much easier to implement if I know what the alphabet of the scoring scheme will be. Is that ok for you?

h-2: We could, but I'd prefer to mark it private. The thing is that the alphabet doesn't necessarily reflect what you are putting in, e.g. nucleotide_scoring_scheme always has alphabet_type dna15 although you can put all nucleotides in.

rrahn: But making it private is not really helpful. I don't see any danger coming with this. The type would be the underlying alphabet type. This information can be helpful and is basically custom stuff anyway, because the concept does not require it so within an algorithm we cannot and won't rely on it.

h-2: Sure, but other people might rely on it and I think it would be better to not make something that is indeed implementation detail part of the API, after all the nucleotide_scoring_scheme might be changed to do something else internally, or not?

> But making it private is not really helpful.

You could make it private in the documentation maybe so you can still use it in tests?

But it's not that important. If you feel strongly about it, just add it :wink:

h-2:
@rrahn merge if you see fit!

include/seqan3/alignment/scoring/scoring_scheme_base.hpp
rrahn: I'll have a look in the evening. See if I can twerk around it.

PR #570, opened 2018-11-19
[feature] for_each on parameter packs

marehr:
Allow applying a function to each argument of a parameter pack.

An example:

```c++
#include <iostream>
#include <string>

#include <seqan3/core/algorithm/parameter_pack.hpp>
#include <seqan3/io/stream/debug_stream.hpp>

using namespace seqan3;

int main()
{
    auto fn = [](auto && a)
    {
        debug_stream << a;
    };

    // prints each argument, i.e. "0, 1, 2, 3\n"
    detail::parameter_pack_for_each(fn, 0, ", ", 1.0, ", ", std::string{"2, 3"}, '\n');

    // is the same as explicitly writing
    fn(0);
    fn(", ");
    fn(1.0);
    fn(", ");
    fn(std::string{"2, 3"});
    fn('\n');
    return 0;
}
```

### DIFF
--- a/include/seqan3/core/algorithm/parameter_pack.hpp
+++ b/include/seqan3/core/algorithm/parameter_pack.hpp
@@ -24,8 +24,9 @@ namespace seqan3::detail
  * \param  fn               The function to call on every argument.
  * \param  args             The arguments as parameter pack.
  *
- * \details This function behaves like std::for_each but on parameter packs.
- * The invocation will be done without any loop.
+ * \details
+ *
+ * This function behaves like std::for_each but on parameter packs. The invocation will be done without any loop.
  *
  * \include test/snippet/core/algorithm/for_each_value.cpp
  */
@@ -46,8 +47,10 @@ constexpr void for_each_value(unary_function_t && fn, args_t && ...args)
  *
  * \param  fn               The function to call on every std::type_identity<type>.
  *
- * \details This function behaves like std::for_each but on types. The type will be wrapped into std::type_identity and
- * passed as argument. The invocation will be done without any loop.
+ * \details
+ *
+ * This function behaves like std::for_each but on types. The type will be wrapped into std::type_identity and passed as
+ * argument. The invocation will be done without any loop.
  *
  * \include test/snippet/core/algorithm/for_each_type.cpp
  */
@@ -70,8 +73,10 @@ constexpr void for_each_type(unary_function_t && fn)
  * \param  fn               The function to call on every std::type_identity<type>.
  * \param  type_list        The list of types, i.e. seqan3::type_list, std::tuple. [unused]
  *
- * \details This function behaves like std::for_each but on types. The type will be wrapped into std::type_identity and
- * passed as argument. The invocation will be done without any loop.
+ * \details
+ *
+ * This function behaves like std::for_each but on types. The type will be wrapped into std::type_identity and passed as
+ * argument. The invocation will be done without any loop.
  *
  * \include test/snippet/core/algorithm/for_each_type_list.cpp
  */
--- a/include/seqan3/core/algorithm/parameter_pack.hpp
+++ b/include/seqan3/core/algorithm/parameter_pack.hpp
@@ -58,8 +58,7 @@ constexpr void for_each_value(unary_function_t && fn, args_t && ...args)
  * This function behaves like std::for_each but on types. The type will be wrapped into std::type_identity and passed as
  * argument. The invocation(s) will be done without any loop.
  *
- * In contrast to `meta::for_each`, this function can handle types which are incomplete, forward declared and not
- * std::Semiregular.
+ * This function can handle types which are incomplete, forward declared and not std::Semiregular.
  *
  * \include test/snippet/core/algorithm/for_each_type.cpp
  *
@@ -89,8 +88,7 @@ constexpr void for_each_type(unary_function_t && fn)
  * This function behaves like std::for_each but on types. The type will be wrapped into std::type_identity and passed as
  * argument. The invocation(s) will be done without any loop.
  *
- * In contrast to `meta::for_each`, this function can handle types which are incomplete, forward declared and not
- * std::Semiregular.
+ * This function can handle types which are incomplete, forward declared and not std::Semiregular.
  *
  * \include test/snippet/core/algorithm/for_each_type_list.cpp
  *
--- a/include/seqan3/core/algorithm/parameter_pack.hpp
+++ b/include/seqan3/core/algorithm/parameter_pack.hpp
@@ -26,9 +26,15 @@ namespace seqan3::detail
  *
  * \details
  *
- * This function behaves like std::for_each but on parameter packs. The invocation will be done without any loop.
+ * This function behaves like std::for_each but on parameter packs. The invocation(s) will be done without any loop.
  *
  * \include test/snippet/core/algorithm/for_each_value.cpp
+ *
+ * \attention The order of arguments is different to std::for_each. In std::for_each the first argument is a range and
+ * the second one is a callable. Due to limitations with the ordering for parameter packs we had to change this order in
+ * the other way around.
+ *
+ * \sa https://en.cppreference.com/w/cpp/language/parameter_pack
  */
 template <typename unary_function_t, typename ...args_t>
 //!\cond
@@ -50,9 +56,11 @@ constexpr void for_each_value(unary_function_t && fn, args_t && ...args)
  * \details
  *
  * This function behaves like std::for_each but on types. The type will be wrapped into std::type_identity and passed as
- * argument. The invocation will be done without any loop.
+ * argument. The invocation(s) will be done without any loop.
  *
  * \include test/snippet/core/algorithm/for_each_type.cpp
+ *
+ * \sa seqan3::detail::for_each_value
  */
 template <typename ...types, typename unary_function_t>
 //!\cond
@@ -76,9 +84,15 @@ constexpr void for_each_type(unary_function_t && fn)
  * \details
  *
  * This function behaves like std::for_each but on types. The type will be wrapped into std::type_identity and passed as
- * argument. The invocation will be done without any loop.
+ * argument. The invocation(s) will be done without any loop.
  *
  * \include test/snippet/core/algorithm/for_each_type_list.cpp
+ *
+ * \attention The order of arguments is different to std::for_each. In std::for_each the first argument is a range and
+ * the second one is a callable. To make the interface consistent with seqan3::detail::for_each_value we change the
+ * order here, too.
+ *
+ * \sa seqan3::detail::for_each_value
  */
 template <template <typename...> typename type_list_t, typename ...args_t, typename unary_function_t>
 //!\cond
--- a/include/seqan3/core/algorithm/parameter_pack.hpp
+++ b/include/seqan3/core/algorithm/parameter_pack.hpp
@@ -1,0 +1,81 @@
+
+#pragma once
+
+#include <utility>
+
+#include <seqan3/std/concepts>
+#include <seqan3/std/type_traits>
+
+namespace seqan3::detail
+{
+
+/*!\brief Applies a function to a range of arguments.
+ * \ingroup algorithm
+ *
+ * \tparam unary_function_t The function type, like function pointers, functors and lambdas.
+ * \tparam ...args_t        The parameter pack of the arguments. (Note that each argument type can be different)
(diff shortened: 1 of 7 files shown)


smehringer:
Nice :sparkles:

include/seqan3/core/algorithm/parameter_pack.hpp
```diff
@@ -0,0 +1,37 @@
...
+template <typename unary_function_t, typename ...args_t>
+//!\cond
+    requires (std::Invocable<unary_function_t, args_t> && ... && true)
+//!\endcond
+constexpr void parameter_pack_for_each(unary_function_t && fn, args_t && ...args)
+{
+    using expander_t = int[];
+    static_cast<void>(expander_t
+        {
+            0, (static_cast<void>(fn(std::forward<args_t>(args))), 0)...
```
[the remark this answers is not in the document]
marehr: this is the perfect reason why we need this, you have a different/better implementation

include/seqan3/core/algorithm/parameter_pack.hpp
```diff
@@ -0,0 +1,37 @@
...
+ * \details This function behaves like std::for_each but on parameter packs.
+ * The invocation will be done without any loop.
+ *
+ * \include test/snippet/core/algorithm/parameter_pack.cpp
+ */
+template <typename unary_function_t, typename ...args_t>
+//!\cond
+    requires (std::Invocable<unary_function_t, args_t> && ... && true)
+//!\endcond
+constexpr void parameter_pack_for_each(unary_function_t && fn, args_t && ...args)
```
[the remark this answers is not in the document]
marehr: First we already use it and second we can replace some of the meta:: library calls if we introduce this. I will add a `for_each_type` metafunction to this PR. I need this for example in a test case where I generate a lot of types for a base type and then test all generated types via a given function. This will be needed in a coming PR for SIMDs.

marehr:
I think the naming is still open, my suggestions are

* `for_each`
* `inline namespace parameter_pack::for_each`
* `pack_for_each`
* `for_each_in_pack`

rrahn:
I vote for `for_each_in_pack`

smehringer:
:raising_hand_woman:  `for_each`

marehr:
:raising_hand_woman:  `inline namespace parameter_pack::for_each`

rrahn:
Actually, we should not introduce inline namespaces in a to general fashion. For simd, it might be ok. But as far as I have read about it it is mainly useful for providing compatibility layers to API/ABI breaks. The latter is not even relevant for us.

rrahn:
The issue that I have with `for_each`, is that it has a different syntax due to the parameter pack. Usually the unary functor is added as last argument, though I am not sure if the pack can be put upfront? 
Actually I am thinking about:

```cpp
apply(fn, ...pack)
apply_unpacked(fn, ...pack) | apply_for_each(fn, ...pack)
```
So the first apply would call fn(pack...) and the second would call fn(pack) ... as seen above.

rrahn:
>The apply and apply_unpacked name is this context the wrong name, because std::apply does something completely different, we want std::for_each. std::apply(fn, args...) = fn(args...), whereas std::for_each(begin, end, fn) = for each elem in [begin, end): fn(elem)

That's what I wrote. 🤔  Still for_each **applies** the fn on every argument on the pack. So it is not so far fetched. But I see your point with the general naming scheme compared to the STL. 
Let us stick to `for_each_in_pack`.

include/seqan3/core/algorithm/parameter_pack.hpp
```diff
@@ -0,0 +1,87 @@
...
+ * \ingroup algorithm
+ *
+ * \tparam type_list_t      The name of the list type.
+ * \tparam ...types         The parameter pack of types within a list.
+ * \tparam unary_function_t The function type, like function pointers, functors and lambdas.
+ *
+ * \param  fn               The function to call on every std::type_identity<type>.
+ * \param  type_list        The list of types, i.e. seqan3::type_list, std::tuple. [unused]
+ *
+ * \details This function behaves like std::for_each but on types. The type will be wrapped into std::type_identity and
```
smehringer: newline after details

test/snippet/core/algorithm/for_each_type.cpp
```diff
@@ -0,0 +1,46 @@
...
+    // auto fn = []<typename type>(std::type_identity<type>)
+    // {
+    // ...
+    // };
+    auto fn = [](auto id)
+    {
+        using type = typename decltype(id)::type;
+
+        // id is of type std::type_identity<type>
+        using id_t = decltype(id);
```
smehringer: move this above of the type definition and then use
```cpp
using type = typename id_t::type;
```
I think it is easier to see the connection then

test/snippet/core/algorithm/for_each_type.cpp
```diff
@@ -0,0 +1,46 @@
...
+        static_assert(std::is_same_v<id_t, std::type_identity<type>>);
+
+        if constexpr(std::is_same_v<type, bool>)
+            debug_stream << "bool";
+
+        if constexpr(std::is_same_v<type, int>)
```
smehringer: else if ?

include/seqan3/core/algorithm/parameter_pack.hpp
```diff
@@ -0,0 +1,87 @@
...
+ * \ingroup algorithm
+ *
+ * \tparam ...types         The parameter pack of types.
+ * \tparam unary_function_t The function type, like function pointers, functors and lambdas.
+ *
+ * \param  fn               The function to call on every std::type_identity<type>.
+ *
+ * \details This function behaves like std::for_each but on types. The type will be wrapped into std::type_identity and
+ * passed as argument. The invocation will be done without any loop.
+ *
```
smehringer: Please explicitly explain the difference to meta::for_each

include/seqan3/core/algorithm/parameter_pack.hpp
```diff
@@ -0,0 +1,87 @@
...
+#include <seqan3/std/type_traits>
+
+namespace seqan3::detail
+{
+
+/*!\brief Applies a function to a range of arguments.
+ * \ingroup algorithm
+ *
+ * \tparam unary_function_t The function type, like function pointers, functors and lambdas.
+ * \tparam ...args_t        The parameter pack of the arguments. (Note that each argument type can be different)
```
rrahn:
```suggestion
 * \tparam ...args_t        The template parameter pack of the arguments. (Note that each argument type can be different)
```

include/seqan3/core/algorithm/parameter_pack.hpp
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * \param  fn               The function to call on every argument.
+ * \param  args             The arguments as parameter pack.
+ *
+ * \details This function behaves like std::for_each but on parameter packs.
+ * The invocation will be done without any loop.
```
rrahn: Add a note, that the order of fn and the `range` of values is different due to the parameter pack expansion.

include/seqan3/core/algorithm/parameter_pack.hpp
```diff
@@ -0,0 +1,87 @@
...
+//!\endcond
+constexpr void for_each_value(unary_function_t && fn, args_t && ...args)
+{
+    [[maybe_unused]] int r = (fn(std::forward<args_t>(args)), ..., 0);
+}
+
+/*!\brief Applies a function to a range of types.
+ * \ingroup algorithm
+ *
+ * \tparam ...types         The parameter pack of types.
```
rrahn:
```suggestion
 * \tparam ...types         The template parameter pack of types.
```

test/snippet/core/algorithm/for_each_type.cpp
```diff
@@ -0,0 +1,46 @@
...
+    // {
+    // ...
+    // };
+    auto fn = [](auto id)
+    {
+        using type = typename decltype(id)::type;
+
+        // id is of type std::type_identity<type>
+        using id_t = decltype(id);
+        static_assert(std::is_same_v<id_t, std::type_identity<type>>);
```
rrahn: please add a proper error message for what you expect here.

include/seqan3/core/algorithm/parameter_pack.hpp
```diff
@@ -0,0 +1,87 @@
...
+ * \param  fn               The function to call on every std::type_identity<type>.
+ *
+ * \details This function behaves like std::for_each but on types. The type will be wrapped into std::type_identity and
+ * passed as argument. The invocation will be done without any loop.
+ *
+ * \include test/snippet/core/algorithm/for_each_type.cpp
+ */
+template <typename ...types, typename unary_function_t>
+//!\cond
+    requires (std::Invocable<unary_function_t, std::type_identity<types>> && ... && true)
```
rrahn: I am not sure if I understand correctly why you have to use type_identity in this context. Can you please elaborate your rationale?

marehr: I wrap the type within `std::type_identity<>` and call the function. The functions must therefore be callable with `type_identity`

smehringer:
Did you investigate the code coverage miss? seems unrelated.

include/seqan3/core/algorithm/parameter_pack.hpp
```diff
@@ -0,0 +1,112 @@
...
+ * \param  args             The arguments as parameter pack.
+ *
+ * \details
+ *
+ * This function behaves like std::for_each but on parameter packs. The invocation(s) will be done without any loop.
+ *
+ * \include test/snippet/core/algorithm/for_each_value.cpp
+ *
+ * \attention The order of arguments is different to std::for_each. In std::for_each the first argument is a range and
+ * the second one is a callable. Due to limitations with the ordering for parameter packs we had to change this order in
```
rrahn:
```suggestion
 * the second one is a callable. Due to limitations with the ordering for parameter packs the order is the other way around.
```

include/seqan3/core/algorithm/parameter_pack.hpp
```diff
@@ -0,0 +1,112 @@
...
+ * This function behaves like std::for_each but on types. The type will be wrapped into std::type_identity and passed as
+ * argument. The invocation(s) will be done without any loop.
+ *
+ * In contrast to `meta::for_each`, this function can handle types which are incomplete, forward declared and not
+ * std::Semiregular.
+ *
+ * \include test/snippet/core/algorithm/for_each_type_list.cpp
+ *
+ * \attention The order of arguments is different to std::for_each. In std::for_each the first argument is a range and
+ * the second one is a callable. To make the interface consistent with seqan3::detail::for_each_value we change the
```
rrahn:
```suggestion
 * the second one is a callable. To make the interface consistent with seqan3::detail::for_each_value the order is changed.
```

include/seqan3/core/algorithm/parameter_pack.hpp
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * \param  fn               The function to call on every std::type_identity<type>.
+ * \param  type_list        The list of types, i.e. seqan3::type_list, std::tuple. [unused]
+ *
+ * \details
+ *
+ * This function behaves like std::for_each but on types. The type will be wrapped into std::type_identity and passed as
+ * argument. The invocation(s) will be done without any loop.
+ *
+ * In contrast to `meta::for_each`, this function can handle types which are incomplete, forward declared and not
```
rrahn: Remove the `meta::for_each` reference, because when we are going to remove the dependency, this comment get's kind of outdated.

include/seqan3/core/algorithm/parameter_pack.hpp
```diff
@@ -0,0 +1,112 @@
...
+ * \tparam unary_function_t The function type, like function pointers, functors and lambdas.
+ *
+ * \param  fn               The function to call on every std::type_identity<type>.
+ *
+ * \details
+ *
+ * This function behaves like std::for_each but on types. The type will be wrapped into std::type_identity and passed as
+ * argument. The invocation(s) will be done without any loop.
+ *
+ * In contrast to `meta::for_each`, this function can handle types which are incomplete, forward declared and not
```
rrahn: remove the reference to meta::for_each.

include/seqan3/core/algorithm/parameter_pack.hpp
marehr: @smehringer :sob: :cry:

PR #578, opened 2018-11-20
Added span.hpp from libc++. Not sure how to test it, and I'm not sure…

### LINKED ISSUE #577: Import std::span from libc++ into our std-module
Copy'n'aste from:
https://github.com/llvm-mirror/libcxx/blob/master/include/span

* include copyright holder in the header
* just remove `_LIBCPP_TEMPLATE_VIS` and `_LIBCPP_INLINE_VISIBILITY`
* adapt to our code-style an make sure stuff is not named `__foo`, but just `foo`
* don't import the comparison operators, because they are deprecated for span
* instead of `_LIBCPP_STD_VER > 17` check `__cplusplus <= 201703L`

joshuak94:
resolves #577

---

… what to do about the comments/documentation currently there...

### DIFF
--- a/include/seqan3/std/span.hpp
+++ b/include/seqan3/std/span.hpp
@@ -1,0 +1,551 @@
+// -*- C++ -*-
+//===------------------------------ span ---------------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is dual licensed under the MIT and the University of Illinois Open
+// Source Licenses. See LICENSE.TXT for details.
+//
+//===---------------------------------------------------------------------===//
+
+#ifndef _LIBCPP_SPAN
+#define _LIBCPP_SPAN
+
+/*
+    span synopsis
+
+namespace std {
+
+// constants
+inline constexpr ptrdiff_t dynamic_extent = -1;
+
+// [views.span], class template span
+template <class ElementType, ptrdiff_t Extent = dynamic_extent>
+    class span;
+
+// [span.comparison], span comparison operators
+template <class T, ptrdiff_t X, class U, ptrdiff_t Y>
+    constexpr bool operator==(span<T, X> l, span<U, Y> r);
+template <class T, ptrdiff_t X, class U, ptrdiff_t Y>
+    constexpr bool operator!=(span<T, X> l, span<U, Y> r);
+template <class T, ptrdiff_t X, class U, ptrdiff_t Y>
+    constexpr bool operator<(span<T, X> l, span<U, Y> r);
+template <class T, ptrdiff_t X, class U, ptrdiff_t Y>
+    constexpr bool operator<=(span<T, X> l, span<U, Y> r);
+template <class T, ptrdiff_t X, class U, ptrdiff_t Y>
+    constexpr bool operator>(span<T, X> l, span<U, Y> r);
+template <class T, ptrdiff_t X, class U, ptrdiff_t Y>
+    constexpr bool operator>=(span<T, X> l, span<U, Y> r);
+
+// [span.objectrep], views of object representation
+template <class ElementType, ptrdiff_t Extent>
+    span<const byte, ((Extent == dynamic_extent) ? dynamic_extent :
+        (static_cast<ptrdiff_t>(sizeof(ElementType)) * Extent))> as_bytes(span<ElementType, Extent> s) noexcept;
+
+template <class ElementType, ptrdiff_t Extent>
+    span<      byte, ((Extent == dynamic_extent) ? dynamic_extent :
+        (static_cast<ptrdiff_t>(sizeof(ElementType)) * Extent))> as_writable_bytes(span<ElementType, Extent> s) noexcept;
+
+
+namespace std {
+template <class ElementType, ptrdiff_t Extent = dynamic_extent>
+class span {
+public:
+    // constants and types
+    using element_type = ElementType;
+    using value_type = remove_cv_t<ElementType>;
+    using index_type = ptrdiff_t;
+    using difference_type = ptrdiff_t;
+    using pointer = element_type*;
+    using reference = element_type&;
+    using iterator = implementation-defined;
+    using const_iterator = implementation-defined;
+    using reverse_iterator = std::reverse_iterator<iterator>;
+    using const_reverse_iterator = std::reverse_iterator<const_iterator>;
+    static constexpr index_type extent = Extent;
+
+    // [span.cons], span constructors, copy, assignment, and destructor
+    constexpr span() noexcept;
+    constexpr span(pointer ptr, index_type count);
+    constexpr span(pointer firstElem, pointer lastElem);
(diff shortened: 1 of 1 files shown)


marehr:
Something does not work yet, the header tests are failing:

```c++
/home/travis/build/seqan/seqan3/include/seqan3/std/span.hpp:11:0: error: unterminated #ifndef
 #ifndef _LIBCPP_SPAN
```

and documentation is broken, because nothing is commented

include/seqan3/std/span.hpp
```diff
@@ -0,0 +1,551 @@
...
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is dual licensed under the MIT and the University of Illinois Open
+// Source Licenses. See LICENSE.TXT for details.
+//
+//===---------------------------------------------------------------------===//
+
+#ifndef _LIBCPP_SPAN
+#define _LIBCPP_SPAN
```
marehr: replace header guard with 
```suggestion
#pragma once
```

include/seqan3/std/span.hpp
```diff
@@ -0,0 +1,551 @@
...
+
```
marehr: add
```suggestion
//!\cond
```

to disable documentation stuff

include/seqan3/std/span.hpp
```diff
@@ -0,0 +1,551 @@
...
+    span(const array<span_tp, span_sz>&) -> span<const span_tp, span_sz>;
+
+template<class container>
+    span(container&) -> span<typename container::value_type>;
+
+template<class container>
+    span(const container&) -> span<const typename container::value_type>;
+} // namespace std
+
+#endif // __cplusplus <= 201703L
```
marehr: and close the no-documentation scope

```suggestion
#endif // __cplusplus <= 201703L
//!\endcond
```

marehr:
We decided that STL imported code does not need documentation. But I'm not sure with regards to tests. I'm inclined to say we don't need it, but I would want to clarify it with the core team first.

smehringer:
IMPOV I do think we should have a small test for it. Nothing extensive or thorough but the basic usage.

h-2:
> We decided that STL imported code does not need documentation. 

I think we decided on referencing the standard documentation, i.e.
```doxygen
\brief Provides std::foobar from the C++20 standard library.
\see https://en.cppreference.com/...
```
And then wrap the rest (all members and related functions) in `\cond` and `\endcond`.

> But I'm not sure with regards to tests. I'm inclined to say we don't need it, but I would want to clarify it with the core team first.

Well, I think for small things we don't need tests. For something as big as span (and since we did some renaming of variables and also smaller changes), maybe at least a small functionality test.

include/seqan3/std/span.hpp
```diff
@@ -0,0 +1,551 @@
...
+    : public true_type {};
+
+
+template <typename span_tp, ptrdiff_t span_extent>
+class span {
+public:
+//  constants and types
+    using element_type           = span_tp;
+    using value_type             = remove_cv_t<span_tp>;
+    using index_type             = ptrdiff_t;
```
h-2: this has now been changed to `size_t` so it's unsigned. We should change it in our version, too.

include/seqan3/std/span.hpp
```diff
@@ -0,0 +1,551 @@
...
+
+    span<byte, span_extent * sizeof(element_type)> as_writeable_bytes() const noexcept
+    { return {reinterpret_cast<byte *>(get_data()), size_bytes()}; }
+
+private:
+    pointer data;
+
+};
+
+
```
h-2: double newline

include/seqan3/std/span.hpp
```diff
@@ -0,0 +1,421 @@
...
+// #include <__config>
+#pragma once
+
+#include <cstddef>      // for ptrdiff_t
+#include <iterator>     // for iterators
+#include <array>        // for array
+#include <type_traits>  // for remove_cv, etc
+#include <cstddef>      // for byte
+
+#if __has_include(<span>)
```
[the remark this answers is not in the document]
smehringer: And you only want the above includes (array, type_traits.. etc.) IF you need to define the span here, so why are they above the if definition?

include/seqan3/std/span.hpp
```diff
@@ -0,0 +1,551 @@
...
+// #include <__config>
+#include <cstddef>      // for ptrdiff_t
+#include <iterator>     // for iterators
+#include <array>        // for array
+#include <type_traits>  // for remove_cv, etc
+#include <cstddef>      // for byte
+
+#pragma once
+
+#if __cplusplus <= 201703L
```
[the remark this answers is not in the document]
joshuak94: I'm confused now, should I be checking `#if __has_include(<span>)` or `if !__has_include(<span>)`?

include/seqan3/std/span.hpp
```diff
@@ -0,0 +1,421 @@
...
+//
+// This file is dual licensed under the MIT and the University of Illinois Open
+// Source Licenses. See LICENSE.TXT for details.
+//
+//===---------------------------------------------------------------------===//
+
+
+/*!\file
+// \brief Provides std::foobar from the C++20 standard library.
+// \see https://en.cppreference.com/...
```
h-2: Actually put the link here!

include/seqan3/std/span.hpp
```diff
@@ -0,0 +1,421 @@
...
+*/
+
+//!\cond
+// #include <__config>
```
h-2: whats this?

include/seqan3/std/span.hpp
```diff
@@ -0,0 +1,421 @@
...
+#pragma once
+
+#include <cstddef>      // for ptrdiff_t
+#include <iterator>     // for iterators
+#include <array>        // for array
+#include <type_traits>  // for remove_cv, etc
+#include <cstddef>      // for byte
```
h-2: sort alphbatically

include/seqan3/std/span.hpp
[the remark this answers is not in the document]
h-2: Yes, it should be 
```cpp
#if __has_include(<span>)
#include <span>
#else 
... rest of file
#endif
```

include/seqan3/std/span.hpp
```diff
@@ -0,0 +1,421 @@
...
+
+template <typename span_tp>
+class span<span_tp, dynamic_extent> {
+private:
+
+public:
+//  constants and types
+    using element_type           = span_tp;
+    using value_type             = remove_cv_t<span_tp>;
+    using index_type             = ptrdiff_t;
```
h-2: this should also be `size_t`

include/seqan3/std/span.hpp
joshuak94: Not sure, was in the original file and I didn't want to remove it, but removing it now!

PR #580, opened 2018-11-21
[FIX] Prevent recursive instantiation in debug_stream

### LINKED ISSUE #540: GCC8 ICE when trying to print a seqan3::filesystem::path with the debug_stream

eseiler:
Resolves #540

### DIFF
--- a/include/seqan3/io/stream/debug_stream.hpp
+++ b/include/seqan3/io/stream/debug_stream.hpp
@@ -312,14 +312,15 @@ inline debug_stream_type & operator<<(debug_stream_type & s, alphabet_t const l)
 template <std::ranges::InputRange rng_t>
 inline debug_stream_type & operator<<(debug_stream_type & s, rng_t && r)
 //!\cond
-    requires requires (reference_t<remove_cvref_t<rng_t>> l) { { debug_stream << l }; } &&
+    requires !std::Same<remove_cvref_t<reference_t<rng_t>>, remove_cvref_t<rng_t>> && // prevent recursive instantiation
+             requires (reference_t<rng_t> l) { { debug_stream << l }; } &&
              // exclude null-terminated strings:
              !(std::is_pointer_v<std::decay_t<rng_t>> &&
-               std::Same<remove_cvref_t<reference_t<std::remove_const_t<rng_t>>>, char>)
+               std::Same<remove_cvref_t<reference_t<rng_t>>, char>)
 //!\endcond
 {
-    if constexpr (alphabet_concept<reference_t<remove_cvref_t<rng_t>>> &&
-                  !detail::is_uint_adaptation_v<remove_cvref_t<reference_t<remove_cvref_t<rng_t>>>>)
+    if constexpr (alphabet_concept<remove_cvref_t<reference_t<rng_t>>> &&
+                  !detail::is_uint_adaptation_v<remove_cvref_t<reference_t<rng_t>>>)
     {
         for (auto && l : r)
             s << l;
--- a/test/unit/io/stream/debug_stream_test.cpp
+++ b/test/unit/io/stream/debug_stream_test.cpp
@@ -38,6 +38,7 @@
 #include <string>
 
 #include <seqan3/alphabet/nucleotide/all.hpp>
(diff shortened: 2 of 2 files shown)


include/seqan3/io/stream/debug_stream.hpp
```diff
@@ -295,6 +296,17 @@ inline debug_stream_type & operator<<(debug_stream_type & s, alphabet_t const l)
     return s << to_char(l);
 }
 
+/*!\brief A seqan3::filesystem::path can be printed to the seqan3::debug_stream by their string representation.
+ * \param s The seqan3::debug_stream.
+ * \param p The seqan3::filesystem::path.
+ * \relates seqan3::debug_stream_type
+ * \sa https://en.cppreference.com/w/cpp/filesystem/path
+ */
+inline debug_stream_type & operator<<(debug_stream_type & s, filesystem::path & p)
```
h-2: It should probably take `const &` of paths, too, or not?

include/seqan3/io/stream/debug_stream.hpp
```diff
@@ -312,7 +324,8 @@ inline debug_stream_type & operator<<(debug_stream_type & s, alphabet_t const l)
 template <std::ranges::InputRange rng_t>
 inline debug_stream_type & operator<<(debug_stream_type & s, rng_t && r)
 //!\cond
-    requires requires (reference_t<remove_cvref_t<rng_t>> l) { { debug_stream << l }; } &&
+    requires !std::Same<remove_cvref_t<reference_t<std::remove_const_t<rng_t>>>, filesystem::path> &&
+             requires (reference_t<remove_cvref_t<rng_t>> l) { { debug_stream << l }; } &&
```
h-2: This should not be required, the above overload is more specialised.

eseiler: When I tried it without it, it still produced an ICE. 
It still tries to evaluate the requires expression, right?

include/seqan3/io/stream/debug_stream.hpp
```diff
@@ -312,7 +324,8 @@ inline debug_stream_type & operator<<(debug_stream_type & s, alphabet_t const l)
 template <std::ranges::InputRange rng_t>
 inline debug_stream_type & operator<<(debug_stream_type & s, rng_t && r)
 //!\cond
-    requires requires (reference_t<remove_cvref_t<rng_t>> l) { { debug_stream << l }; } &&
+    requires !std::Same<remove_cvref_t<reference_t<std::remove_const_t<rng_t>>>, filesystem::path> &&
```
rrahn: Why not?
```suggestion
    requires !std::Same<remove_cvref_t<rng_t>, filesystem::path> &&
```

include/seqan3/io/stream/debug_stream.hpp
smehringer: @h-2 As The ICE is produced when evaluating the requires statement in line 328. As @eseiler said, this is a hack to avoid the ICE by taking advantage of the lazy evaluation of requires statements.

h-2: Ah, what a mess. But then please check instead whether the reference types is equal to the type itself. That's more generic than checking for path directly.

eseiler: I check now whether the reference type is equal to the type.

eseiler:
I would suggest to drop the overload for `filesystem::path`.
The change in the requires statement already resolves the ICE and makes the `<<` operator use the member function `debug_stream_type & operator<<(t && v)`.

The resulting behaviour is then the same as you would expect for printing to the stream directly:

```cpp
seqan3::filesystem::path const my_path{"/usr/file.txt"};
seqan3::debug_stream << my_path << '\n'; // "/usr/file.txt"
std::cerr << my_path << '\n'; // "/usr/file.txt"
```

If we'd opt for the overloads, we would also need two: one for `filesystem::path const &` and one for `filesystem::path &`. Otherwise, it will resolve to the member function for the missing overload.

eseiler:
I checked the `remove_cvref_t` stuff and changed it a bit - it still works.

PR #582, opened 2018-11-22
[FIX] Fixes scoring scheme matrix size type.

rrahn:
Some minor fix for scoring schemes.
Currntly the size type for the scoring scheme matrix is fixed to uint8_t, which won't work for char alphabets. 
To make it generic the matrix size type should be dependent to the size type that can hold the size of the alphabet. While char adaptions already exposed a type in the alphabet_size value meta function, the other alphabets did not. 
This fix adds the type alias for the alphabet_size value meta function.

### DIFF
--- a/include/seqan3/alignment/scoring/scoring_scheme_base.hpp
+++ b/include/seqan3/alignment/scoring/scoring_scheme_base.hpp
@@ -125,7 +125,7 @@ public:
     //!\brief Type of the score values.
     using score_type = score_t;
     //!\brief Size type that can hold the dimension of the matrix (i.e. size of the alphabet).
-    using matrix_size_type = uint8_t;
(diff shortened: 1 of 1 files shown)


rrahn:
I agree on inheriting from integral_constant, but I don't understand why this wouldn't work then as integral constants expose their types (of course it is called value_type), but the inconsistency was alreadt there. But anyway, decltype seems less invasive to me so I will change it.

h-2:
Btw, why is it not possible to use `underlying_rank_t` from your POV? Why do you need the type of the size?

h-2:
Can you change this quickly so we can merge it?=

rrahn:
I am looking at alphabet refactoring 😏

PR #583, opened 2018-11-23
Comparison seqan2 and seqan3 for fasta I/O

MitraDarja:
As you can see in the benchmark, I compare reading and writing of fasta files in seqan2 and seqan3. The results for the same input are (write2 and read2 are referring to seqan2):

write 31715 ns 31715 ns 19368
write2 506 ns 506 ns 1361641
read 38040 ns 38039 ns 18167
read2 2368 ns 2367 ns 300955

So, it looks like seqan3 is slower.

### DIFF
--- a/test/performance/io/format_fasta_benchmark.cpp
+++ b/test/performance/io/format_fasta_benchmark.cpp
@@ -1,0 +1,112 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
+// -----------------------------------------------------------------------------------------------------
+
+#if __has_include(<seqan/seq_io.h>)
+    #include <seqan/seq_io.h>
+#endif
+
+#include <algorithm>
+#include <cctype>
+#include <cstring>
+
+#include <benchmark/benchmark.h>
+
+#include <range/v3/algorithm/equal.hpp>
+#include <range/v3/view/transform.hpp>
+
+#include <seqan3/alphabet/quality/all.hpp>
+#include <seqan3/io/sequence_file/input.hpp>
+#include <seqan3/io/sequence_file/input_format_concept.hpp>
+#include <seqan3/io/sequence_file/output.hpp>
+#include <seqan3/io/sequence_file/output_format_concept.hpp>
+#include <seqan3/io/sequence_file/format_fasta.hpp>
+#include <seqan3/range/view/convert.hpp>
+
+#include <sstream>
+
+using namespace seqan3;
+
+void write3(benchmark::State & state)
+{
+    std::ostringstream ostream;
+    sequence_file_format_fasta format;
+    sequence_file_output_options options;
+    std::string id{"seq"};
+    dna5_vector seq{"ACTAGACTAGCTACGATCAGCTACGATCAGCTACGA"_dna5};
+
+    for (auto _ : state)
+    {
+        format.write(ostream, options, seq, id, std::ignore);
+    }
+}
+
+BENCHMARK(write3);
+
+#if __has_include(<seqan/seq_io.h>)
+
+void write2(benchmark::State & state)
+{
+    std::ostringstream outStream;
+    seqan::CharString meta = "seq";
+    seqan::Dna5String seq = "ACTAGACTAGCTACGATCAGCTACGATCAGCTACGA";
+
+    for (auto _ : state)
+    {
+        seqan::writeRecord(outStream, meta, seq, seqan::Fasta());
+    }
+}
+
+BENCHMARK(write2);
+#endif
+
+void read3(benchmark::State & state)
+{
+    std::string dummy_file{};
+    for (size_t idx = 0; idx < 10000000; idx++)
+        dummy_file += ">seq\nACTAGACTAGCTACGATCAGCTACGATCAGCTACGA\n";
+    std::istringstream istream{dummy_file};
+    sequence_file_format_fasta format;
+    sequence_file_input_options<dna5, false> options;
+    std::string id;
+    dna5_vector seq;
+    for (auto _ : state)
+    {
+        format.read(istream, options, seq, id, std::ignore);
+	    id.clear();
+	    seq.clear();
+    }
+}
+BENCHMARK(read3);
+
+#if __has_include(<seqan/seq_io.h>)
+
+#include <fstream>
+
+void read2(benchmark::State & state)
+{
+    seqan::CharString meta;
+    seqan::Dna5String seq;
+    std::string dummy_file{};
+    for (size_t idx = 0; idx < 10000000; idx++)
+        dummy_file += ">seq\nACTAGACTAGCTACGATCAGCTACGATCAGCTACGA\n";
+    std::istringstream istream{dummy_file};
+
+    seqan::VirtualStream<char, seqan::Input> comp;
+    open(comp, istream);
+    auto it = seqan::directionIterator(comp, seqan::Input());
+
+    for (auto _ : state)
+    {
+        readRecord(meta, seq,  it, seqan::Fasta{});
+        clear(meta);
+        clear(seq);
+    }
+}
+BENCHMARK(read2);
+#endif
+
+BENCHMARK_MAIN();
--- a/test/performance/io/CMakeLists.txt
+++ b/test/performance/io/CMakeLists.txt
@@ -1,3 +1,4 @@
 seqan3_benchmark(stream_input_benchmark.cpp)
 seqan3_benchmark(stream_output_benchmark.cpp)
 seqan3_benchmark(parse_condition_benchmark.cpp)
+seqan3_benchmark(format_fasta_benchmark.cpp)


test/performance/io/format_fasta_benchmark.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+    static void write2(benchmark::State& state)
+    {
+        seqan::CharString outStream;
+        seqan::CharString meta = "seq";
+        seqan::Dna5String seq = "ACGTAAAAAAAAAAAAAAAAAAAAAAAAAAAAA";
+
+        for (auto _ : state)
+        {
+            seqan::writeRecord(outStream, meta, seq, seqan::Fasta());
```
smehringer: please also make the writeRecord of seqan2 print to a stringstream ! Otherwise comparing both functions is not possible

test/performance/io/format_fasta_benchmark.cpp
```diff
@@ -0,0 +1,157 @@
...
+    {
+        seqan::CharString filename;
+        std::fstream *file = createFastAFile(filename);
+        seqan::SeqFileIn reader(*file);
+        seqan::CharString meta;
+        seqan::Dna5String seq;
+
+        for (auto _ : state)
+        {
+            seqan::readRecord(meta, seq, reader, seqan::Fasta());
```
smehringer: Same here. I am pretty sure that seqan2 readRecord also works with a string stream no?

smehringer: Keep The example with the file though and make one for seqan3 too

MitraDarja:
New numbers:

write3                       241 ns        241 ns    2451972
write2                       339 ns        339 ns    2065208
read3                     1150 ns      1150 ns      611523
read3_from_file     1229 ns       1229 ns     569913
read2_from_file        138 ns        138 ns   5076996

h-2:
Ah, nice, we are faster on writing! Do you have numbers for `read2` as well, without file? And did you repeat the test a few times, especially the `*_file`?

test/performance/io/format_fasta_benchmark.cpp
```diff
@@ -0,0 +1,202 @@
...
+    for (size_t idx = 0; idx < 10000000; idx++)
+        dummy_file += ">seq\nACTAGACTAGCTACGATCAGCTACGATCAGCTACGA\n";
+    std::stringstream istream{dummy_file};
+    sequence_file_format_fasta format;
+    sequence_file_input_options<dna5, false> options;
+    std::string id;
+    dna5_vector seq;
+    for (auto _ : state)
+    {
+        format.read(istream, options, seq, id, std::ignore);
```
smehringer: add
```
seq.clear();
id.clear();
```
and check if something changes on the runtime because I am pretty sure that seqan3 is appending and cashing large sequences may impact runtime.

test/performance/io/format_fasta_benchmark.cpp
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    sequence_file_input fin{filename};
+    auto file_it = fin.begin();
+    std::string id;
+    dna5_vector seq;
+
+    for (auto _ : state)
+    {
+        auto record = *file_it;
```
smehringer:
```suggestion
       [[maube_unused]] volatile auto record = *file_it;
```
just to make sure nothing is optimized away

test/performance/io/format_fasta_benchmark.cpp
```diff
@@ -0,0 +1,202 @@
...
+#if __has_include(<seqan/seq_io.h>)
+
+#include <fstream>
+
+std::fstream* createFastAFile(seqan::CharString &tempFilename)
+{
+    using namespace seqan;
+
+    tempFilename = SEQAN_TEMP_FILENAME();
+    char filenameBuffer[1000];
```
smehringer: I think your code for writing a file is a little complicated. You can just do this:
```cpp
  std::ofstream myfile;
  myfile.open (toCString(tempFilename));
  SEQAN_ASSERT(file->is_open());

  for (size_t idx = 0; idx < 10000000; idx++)
      myfile << ">seq\nACTAGACTAGCTACGATCAGCTACGATCAGCTACGA\n";
  
  myfile.close();
```

test/performance/io/format_fasta_benchmark.cpp
```diff
@@ -0,0 +1,202 @@
...
+        }
+
+    }
+    BENCHMARK(read2);
+*/
+    static void read2_from_file(benchmark::State& state)
+   {
+       seqan::CharString filename;
+       std::fstream *file = createFastAFile(filename);
+       seqan::SeqFileIn reader(*file);
```
smehringer: I actually think you can just use the file you already have written down:
```suggestion
       seqan::SeqFileIn reader("Test_Format_Fasta_Benchmark.fasta");
```

smehringer:
@h-2 Can you direct me to how Seqan2 can read from a stringstream? I did not find any documentation on this and I did not find a clue a good clue in the code base..

rrahn:
@smehringer in SeqAn2 all read interfaces can also take a stream iterator. So you should be able to call it with an stream_buf iterator over your string stream.

smehringer:
I saw this function but I am unfamiliar with the stream_buf iterator. Can you paste an example of how to create a stream_buf iterator from the stream?

rrahn:
```cpp
std::istringstream s;    
// s << fasta content
std::istreambuf_iterator<char> iter(s)
readRecord(id, seq,  iter, Fasta{});
```

test/performance/io/format_fasta_benchmark.cpp
MitraDarja: I tested it twice with and without it and there was no difference but I add it again. :)

test/performance/io/format_fasta_benchmark.cpp
MitraDarja: Okay, but that code is from a Seqan2 test file. Do you want it changed there as well?

MitraDarja:
Instead of creating the fasta file over and over again I commented that code out and used the same file, so I could run it a bit more often. Is there a possibility to add an example file in general or do you prefer having a tmp_file always created, @h-2 ?

Anyways, I added the stream reading for seqan2 and this are the new numbers:

write3                       526 ns         526 ns    1335735
write2                       453 ns         453 ns    1500982
read3                      1791 ns       1791 ns     390022
read3_from_file       2605 ns       2605 ns     335116
read2                        729 ns         729 ns      961222
read2_from_file        215 ns         215 ns    4154270

h-2:
@MitraDarja Now the performance for write2 and write3 has switched 🤔 Have you run this benchmark a few times to see if there are differences between the runs? Also maybe try not to run it on a laptop that does a lot of power-management and CPU scaling.

But in general I think we are getting a lot closer to the bare metal results now.

> Is there a possibility to add an example file in general or do you prefer having a tmp_file always created, @h-2 ?

I would propose not writing/reading files at all to be honest. There is so much operating system stuff going on, caching on so many different layers, it's just not meaningful here. We will have to do some file I/O at some point, but that will be independent of some format. I would suggest to just use stringstreams here.

>  ```cpp
> std::istringstream s;    
> // s << fasta content
> std::istreambuf_iterator<char> iter(s)
> readRecord(id, seq,  iter, Fasta{});
> ```

To be fair, I think SeqAn2 never actually does this. It usually goes through VirtualStream and directionIterator which my tests show is measurably slower that the above approach.

h-2:
@MitraDarja Now the performance for write2 and write3 has switched 🤔 Have you run this benchmark a few times to see if there are differences between the runs? Also maybe try not to run it on a laptop that does a lot of power-management and CPU scaling.

But in general I think we are getting a lot closer to the bare metal results now.

> Is there a possibility to add an example file in general or do you prefer having a tmp_file always created, @h-2 ?

I would propose not writing/reading files at all to be honest. There is so much operating system stuff going on, caching on so many different layers, it's just not meaningful here. We will have to do some file I/O at some point, but that will be independent of some format. I would suggest to just use stringstreams here.

>  ```cpp
> std::istringstream s;    
> // s << fasta content
> std::istreambuf_iterator<char> iter(s)
> readRecord(id, seq,  iter, Fasta{});
> ```

To be fair, I think SeqAn2 never actually does this. It usually goes through VirtualStream and directionIterator which my tests show is measurably slower that the above approach.

MitraDarja:
Okay, I now tried 10 times the script on two different computers, these are the means. (It wrote a python-script, I can increase easily the 10 to any other number but I was impatient...)

Computer 1 (8 X 3900 MHz CPU s):
write3                   200.9   200.9 3095566.0
write2                   339.6   339.6 2067232.3
read3                 1132.4 1132.3   618169.8
read3_from_file 1390.7 1390.3   392847.4
read2                   489.4   489.4 1430715.3
read2_from_file   118.1   118.0 5316102.6

Computer 2 (24 X 2393.85 MHz CPU s):
write3                   422.0  422.0  1600921.0
write2                   454.9  455.0  1498626.0
read3                 1792.4 1792.3   390781.2
read3_from_file 2544.3 2541.0   352657.0
read2                   756.3   756.3   925936.1
read2_from_file    213.0  207.7 4487121.0

h-2:
> Okay, I now tried 10 times the script on two different computers, these are the means. 

:+1: 

Although I think the median or even the best would be more accurate as the mean, because we want to disregard bat outliers completely.

> @h-2 So, what would be a fairer way to use streams in seqan2?

Well, like I said, use VirtualStream and directionIterator :laughing: Here's an example:
https://github.com/seqan/seqan3/pull/612/files#diff-ad7a91365f283d18fce9a6dc6f5d1bcaR215

MitraDarja:
@h-2 You mean like this?

        std::string dummy_file{};
        for (size_t idx = 0; idx < 10000000; idx++)
            dummy_file += ">seq\nACTAGACTAGCTACGATCAGCTACGATCAGCTACGA\n";
        std::istringstream istream{dummy_file};

        seqan::VirtualStream<char, seqan::Input> comp;
        open(comp, istream);
        auto it = seqan::directionIterator(comp, seqan::Input());

        for (auto _ : state)
        {
            [[maybe_unused]] volatile auto record = *it;
            ++it;
        }

Because then one iteration of read2 from stringstream takes only 2 ns, which can't be true.

h-2:
@MitraDarja No, you still need to do readRecord, just do it on the directionIterator instead of the streambuf iterator!

MitraDarja:
Now, the new numbers. This time I used the median. (I don't like the idea of using only the best value, which could also be an outlier...) Again, 10 test cases. 

Computer 1 (8 X 3900 MHz CPU s)
write3                   207.5 ns   207.5 ns 2995817.0
write2                   356.0 ns   356.0 ns 1985485.5
read3                 1106.5 ns 1106.5 ns   632377.5
read3_from_file 1417.0 ns 1417.0 ns   364720.5
read2                   114.0 ns   114.0 ns 6162158.0
read2_from_file   111.0 ns   111.0 ns 5962923.5

Computer 2 (24 X 2393.85 MHz CPU s)
write3                   566.5 ns  566.5 ns 1179196.0
write2                   477.5 ns  477.5 ns 1420207.5
read3                 1713.0 ns 1713.0 ns  408725.0
read3_from_file 2804.5 ns 2804.5 ns  321103.5
read2                   182.0 ns  182.0 ns 3835901.5
read2_from_file   203.0 ns  203.0 ns 4490582.0

test/performance/io/format_fasta_benchmark.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+BENCHMARK(read3_from_file);
+
+
+#if __has_include(<seqan/seq_io.h>)
+
+#include <fstream>
+
+static void read2(benchmark::State& state)
+    {
```
smehringer: no indentation here (the whole function)

test/performance/io/format_fasta_benchmark.cpp
```diff
@@ -0,0 +1,178 @@
...
+            dummy_file += ">seq\nACTAGACTAGCTACGATCAGCTACGATCAGCTACGA\n";
+        std::istringstream istream{dummy_file};
+
+        seqan::VirtualStream<char, seqan::Input> comp;
+        open(comp, istream);
+        auto it = seqan::directionIterator(comp, seqan::Input());
+
+        for (auto _ : state)
+        {
+            readRecord(meta, seq,  it, seqan::Fasta{});
```
smehringer:
```suggestion
            [[maybe_unused]] readRecord(meta, seq,  it, seqan::Fasta{});
```
add this to eevery read line just to make sure

test/performance/io/format_fasta_benchmark.cpp
```diff
@@ -0,0 +1,178 @@
...
+            //[[maybe_unused]] volatile auto record = *it;
+            //++it;
+        }
+    }
+    BENCHMARK(read2);
+
+    static void read2_from_file(benchmark::State& state)
```
smehringer: why the indentation?

test/performance/io/format_fasta_benchmark.cpp
MitraDarja: @smehringer  I get an error message, if I do that because I am not defining record here, so should I use this in front of the definitions of meta, seq and it/reader?

smehringer: ah I see. Then don't add it here.

h-2:
@MitraDarja Can you remove the file benchmarks for now, as discussed? Can you also give the exact gcc version for the benchmarks and format the tables in markdown or some other way so that they are readable?

I have run some tests and it seems like gcc8 is causing us some major slowdowns over gcc7 (it makes seqan2 reading measurabley faster and seqan3 reading slower by a factor of 4):

gcc7:
```
--------------------------------------------------
Benchmark           Time           CPU Iterations
--------------------------------------------------
write3            303 ns        303 ns    2237394
write2            563 ns        563 ns    1191895
read3             401 ns        401 ns    1746398
read2             113 ns        113 ns    6393161
```


gcc8:
```
--------------------------------------------------
Benchmark           Time           CPU Iterations
--------------------------------------------------
write3            298 ns        298 ns    2242296
write2            556 ns        556 ns    1253021
read3            1671 ns       1671 ns     419320
read2              89 ns         89 ns    7813023
```

marehr:
Can you add an output of how many bytes where processed? s.t. we can calculate what the transfer rate was, e.g. in `GB/s`.

test/performance/io/format_fasta_benchmark.cpp
```diff
@@ -0,0 +1,180 @@
...
+// ARE DISCLAIMED. IN NO EVENT SHALL KNUT REINERT OR THE FU BERLIN BE LIABLE
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+// DAMAGE.
+//
+// ==========================================================================
```
h-2: short copyright

test/performance/io/format_fasta_benchmark.cpp
```diff
@@ -0,0 +1,180 @@
...
+#include <seqan3/io/sequence_file/format_fasta.hpp>
+#include <seqan3/range/view/convert.hpp>
+
+#include <sstream>
+
+using namespace seqan3;
+
+static void write3(benchmark::State& state)
+{
+    std::stringstream ostream;
```
h-2: ostringstream

test/performance/io/format_fasta_benchmark.cpp
```diff
@@ -0,0 +1,180 @@
...
+    }
+}
+
+BENCHMARK(write3);
+
+#if __has_include(<seqan/seq_io.h>)
+
+    static void write2(benchmark::State& state)
+    {
+        std::stringstream outStream;
```
h-2: ostringstream

test/performance/io/format_fasta_benchmark.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+    BENCHMARK(write2);
+#endif
+
+static void read3(benchmark::State& state)
+{
+    std::string dummy_file{};
+    for (size_t idx = 0; idx < 10000000; idx++)
+        dummy_file += ">seq\nACTAGACTAGCTACGATCAGCTACGATCAGCTACGA\n";
+    std::stringstream istream{dummy_file};
```
h-2: istringstream

test/performance/io/format_fasta_benchmark.cpp
```diff
@@ -0,0 +1,180 @@
...
+    for (auto _ : state)
+    {
+        [[maybe_unused]] volatile auto record = *file_it;
+        ++file_it;
+    }
+}
+
+BENCHMARK(read3_from_file);
+
+
```
h-2: double newline

PR #233, opened 2018-04-23
Alignment class and stream operator

joergi-w:
This PR is waiting for aligned_sequence and should be updated before merge. 
Replaces #20

PR #20, opened 2017-03-23
[FEATURE] Alignment class

joergi-w:

PR #593, opened 2018-11-28
Stream sequence alignments in multi-line representation

joergi-w:
This PR adds debug_stream's `operator<<` for alignments, which are represented as tuple or pair of aligned sequences. Should I add support for std::vector of aligned sequences as well? 

Due to our file output policy of not checking user variables before printing, the output may be corrupted, if the sequences do not have equal length. 

The documentation of how to use zip_view as a column iterator for alignments will come in a separate PR. 

replaces PR #233 and #20

### DIFF
--- a/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
+++ b/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
@@ -299,7 +299,6 @@ template<tuple_like_concept alignment_t, size_t ...idx>
 void stream_alignment(debug_stream_type & stream, alignment_t const & align, std::index_sequence<idx...> const & /**/)
 {
     size_t const alignment_size = std::get<0>(align).size();
-    char const * indent = "        ";
 
     // split alignment into blocks of length 50 and loop over parts
     for (size_t begin_pos = 0; begin_pos < alignment_size; begin_pos += 50)
@@ -322,19 +321,19 @@ void stream_alignment(debug_stream_type & stream, alignment_t const & align, std
         }
 
         // write first sequence
-        stream << '\n' << indent;
+        stream << '\n' << std::setw(8) << "";
         ranges::for_each(std::get<0>(align) | ranges::view::slice(begin_pos, end_pos) | view::to_char,
                          [&stream] (char ch) { stream << ch; });
 
         auto stream_f = [&] (auto const & previous_seq, auto const & aligned_seq)
         {
             // write alignment bars
-            stream << '\n' << indent;
+            stream << '\n' << std::setw(8) << "";
             ranges::for_each(ranges::zip_view(previous_seq, aligned_seq) | ranges::view::slice(begin_pos, end_pos),
-                             [&stream] (auto ch) { stream << (std::get<0>(ch) == std::get<1>(ch) ? '|' : ' '); });
+                             [&stream] (auto && ch) { stream << (std::get<0>(ch) == std::get<1>(ch) ? '|' : ' '); });
 
             // write next sequence
-            stream << '\n' << indent;
+            stream << '\n' << std::setw(8) << "";
             ranges::for_each(aligned_seq | ranges::view::slice(begin_pos, end_pos) | view::to_char,
                              [&stream] (char ch) { stream << ch; });
         };
--- a/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
+++ b/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
@@ -40,10 +40,19 @@
 
 #pragma once
 
+#include <iomanip>
+#include <tuple>
+
+#include <range/v3/algorithm/for_each.hpp>
+#include <range/v3/view/slice.hpp>
+
 #include <seqan3/alignment/exception.hpp>
 #include <seqan3/alphabet/gap/all.hpp>
+#include <seqan3/core/concept/tuple.hpp>
 #include <seqan3/core/metafunction/all.hpp>
+#include <seqan3/io/stream/debug_stream.hpp>
 #include <seqan3/range/container/concept.hpp>
+#include <seqan3/range/view/to_char.hpp>
 #include <seqan3/std/ranges>
 
 namespace seqan3
@@ -204,7 +213,9 @@ template <sequence_container_concept seq_type>
 //!\cond
     requires weakly_assignable_concept<reference_t<seq_type>, gap const &>
 //!\endcond
-inline typename seq_type::iterator insert_gap(seq_type & seq, typename seq_type::const_iterator pos_it, typename seq_type::size_type count)
+inline typename seq_type::iterator insert_gap(seq_type & seq,
+                                              typename seq_type::const_iterator pos_it,
+                                              typename seq_type::size_type count)
 {
     return seq.insert(pos_it, count, value_type_t<seq_type>{gap::GAP});
 }
@@ -262,7 +273,9 @@ template <sequence_container_concept seq_type>
 //!\cond
     requires weakly_assignable_concept<reference_t<seq_type>, gap const &>
 //!\endcond
-inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::const_iterator first, typename seq_type::const_iterator last)
+inline typename seq_type::iterator erase_gap(seq_type & seq,
+                                             typename seq_type::const_iterator first,
+                                             typename seq_type::const_iterator last)
 {
     for (auto it = first; it != last; ++it)
         if (*it != gap::GAP) // [[unlikely]]
@@ -271,5 +284,98 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
     return seq.erase(first, last);
 }
 //!\}
+
+namespace detail
+{
+
+/*!
+ * Create the formatted alignment output and add it to the provided debug_stream.
+ * \tparam alignment_t  The type of the alignment, must satisfy tuple_like_concept.
+ * \tparam idx          An index sequence.
+ * \param stream        The output stream that receives the formatted alignment.
+ * \param align         The alignment that shall be streamed.
+ */
+template<tuple_like_concept alignment_t, size_t ...idx>
+void stream_alignment(debug_stream_type & stream, alignment_t const & align, std::index_sequence<idx...> const & /**/)
+{
+    size_t const alignment_length = std::get<0>(align).size();
+
+    // split alignment into blocks of length 50 and loop over parts
+    for (size_t used_length = 0; used_length < alignment_length; used_length += 50)
+    {
+        // write header
+        if (used_length != 0)
+            stream << std::endl;
+
+        stream << std::setw(7) << used_length << ' ';
+        for (size_t col = 1; col <= 50 && col + used_length <= alignment_length; ++col)
+        {
+            if (col % 10 == 0)
+                stream << ':';
+            else if (col % 5 == 0)
+                stream << '.';
+            else
+                stream << ' ';
+        }
+
+        // write sequences
+        const char * indent = "        ";
+        stream << std::endl << indent;
+        size_t const col_end = std::min(used_length + 50, alignment_length);
+        ranges::for_each(std::get<0>(align) | ranges::view::slice(used_length, col_end) | view::to_char,
+                         [&stream] (char ch) { stream << ch; });
+
+        auto stream_f = [&] (auto const & previous_sequence, auto const & aligned_sequence)
+        {
+            stream << std::endl << indent;
+            auto seq1 = previous_sequence.begin() + used_length;
+            auto seq2 = aligned_sequence.begin() + used_length;
+            for (auto it1 = seq1, it2 = seq2;
+                 it1 < previous_sequence.end() && it1 < seq1 + 50 &&
+                 it2 < aligned_sequence.end()  && it2 < seq2 + 50;
+                 ++it1, ++it2)
+            {
+                stream << (seqan3::to_char(*it1) == seqan3::to_char(*it2) ? '|' : ' ');
+            }
+            stream << std::endl << indent;
+            ranges::for_each(aligned_sequence | ranges::view::slice(used_length, col_end) | view::to_char,
+                             [&stream] (char ch) { stream << ch; });
+        };
+        (stream_f(std::get<idx>(align), std::get<idx + 1>(align)), ...);
+        stream << std::endl;
+    }
+}
+
+} // namespace detail
+
+/*!
+ * \brief True, if each type satisfies aligned_sequence_concept; false otherwise.
+ * \tparam elems The pack of types to be tested.
+ */
+template <typename ...elems>
+inline bool constexpr all_satisfy_aligned_seq = false;
+
+/*!
+ * \brief True, if each type satisfies aligned_sequence_concept; false otherwise.
+ * \tparam elems The pack of types to be tested.
+ */
+template <typename ...elems>
+inline bool constexpr all_satisfy_aligned_seq<type_list<elems...>> = (aligned_sequence_concept<elems> && ... && true);
+
+/*!
+ * \brief Streaming operator for alignments, which are represented as tuples of aligned sequences.
+ * \tparam tuple_t  The alignment type, must satisfy tuple_like_concept and its size must be at least 2.
+ * \param stream    The target stream for the formatted output.
+ * \param alignment The alignment that shall be formatted. All sequences must be equally long.
+ * \return          The given stream to which the alignment representation is appended.
+ */
+template <tuple_like_concept tuple_t>
+requires all_satisfy_aligned_seq<detail::tuple_type_list_t<tuple_t>>
+inline debug_stream_type & operator<<(debug_stream_type & stream, tuple_t const & alignment)
+{
+    static_assert(std::tuple_size_v<tuple_t> >= 2, "An alignment requires at least two sequences.");
+    detail::stream_alignment(stream, alignment, std::make_index_sequence<std::tuple_size_v<tuple_t> - 1> {});
+    return stream;
+}
 
 } // namespace seqan
--- a/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
+++ b/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
@@ -310,11 +310,11 @@ void stream_alignment(debug_stream_type & stream, alignment_t const & align, std
             stream << '\n';
 
         stream << std::setw(7) << begin_pos << ' ';
-        for (size_t rel_pos = 1; rel_pos <= (end_pos - 1) % 50 + 1; ++rel_pos)
+        for (size_t pos = begin_pos + 1; pos <= end_pos; ++pos)
         {
-            if (rel_pos % 10 == 0)
+            if (pos % 10 == 0)
                 stream << ':';
-            else if (rel_pos % 5 == 0)
+            else if (pos % 5 == 0)
                 stream << '.';
             else
                 stream << ' ';
--- a/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
+++ b/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
@@ -106,7 +106,7 @@ namespace seqan3
  *                         seqan3::aligned_sequence_concept.
  * \param[in,out] seq      The aligned sequence to modify.
  * \param[in]     pos_it   The iterator pointing to the position where to insert a gaps.
- * \param[in]     size     The number of gaps to insert.
+ * \param[in]     size     The number of gap symbols to insert (will result in a gap of length size).
  *
  * \details
  * \attention This is a concept requirement, not an actual function (however types
@@ -200,7 +200,7 @@ inline typename seq_type::iterator insert_gap(seq_type & seq, typename seq_type:
  *                         seqan3::gap.
  * \param[in,out] seq      The container to modify.
  * \param[in]     pos_it   The iterator pointing to the position where to insert gaps.
- * \param[in]     size     The number of gaps to insert.
+ * \param[in]     size     The number of gap symbols to insert (will result in a gap of length size).
  *
  * \relates seqan3::aligned_sequence_concept
  *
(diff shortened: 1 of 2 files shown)


include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -272,4 +285,97 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+
+/*!
+ * \brief Streaming operator for alignments, which are represented as tuples of aligned sequences.
+ * \tparam tuple_t  The alignment type, must satisfy tuple_like_concept and its size must be at least 2.
+ * \param stream    The target stream for the formatted output.
+ * \param alignment The alignment that shall be formatted. All sequences must be equally long.
+ * \return          The given stream to which the alignment representation is appended.
+ */
+template <tuple_like_concept tuple_t>
+requires all_satisfy_aligned_seq<detail::tuple_type_list_t<tuple_t>>
```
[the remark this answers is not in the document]
joergi-w: I tried, but the `quote` struct has a type/value mismatch, when `aligned_sequence_concept` is passed.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -272,4 +285,97 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+            if (col % 10 == 0)
+                stream << ':';
+            else if (col % 5 == 0)
+                stream << '.';
+            else
+                stream << ' ';
+        }
+
+        // write sequences
+        const char * indent = "        ";
```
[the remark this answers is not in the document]
joergi-w: true!

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -272,4 +285,97 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
 }
 //!\}
 
+namespace detail
+{
+
+/*!
```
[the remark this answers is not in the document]
joergi-w: yes ;)

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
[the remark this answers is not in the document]
smehringer: I feared as much. The concept is no meta function so it does not work... @rrahn Is there a way to treat a concept like a metafunction?

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -204,7 +213,9 @@ template <sequence_container_concept seq_type>
 //!\cond
     requires std::Assignable<reference_t<seq_type>, gap const &>
 //!\endcond
-inline typename seq_type::iterator insert_gap(seq_type & seq, typename seq_type::const_iterator pos_it, typename seq_type::size_type count)
+inline typename seq_type::iterator insert_gap(seq_type & seq,
+                                              typename seq_type::const_iterator pos_it,
+                                              typename seq_type::size_type count)
```
rrahn: call it size maybe instead of count? The semantic would be to insert a single gap of a certain size and not count many gap characters, right?

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -272,4 +285,99 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
 }
 //!\}
 
+namespace detail
+{
+
+/*!
+ * \brief               Create the formatted alignment output and add it to the provided debug_stream.
```
rrahn: ingroup missing?

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -272,4 +285,99 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+            if (col % 10 == 0)
+                stream << ':';
+            else if (col % 5 == 0)
+                stream << '.';
+            else
+                stream << ' ';
+        }
+
+        // write sequences
+        stream << std::endl << indent;
```
rrahn: why std::endl and not '\n' and why not using setw and then align right: see https://en.cppreference.com/w/cpp/io/manip for available manipulators.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -272,4 +285,99 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+{
+    size_t const alignment_length = std::get<0>(align).size();
+    char const * indent = "        ";
+
+    // split alignment into blocks of length 50 and loop over parts
+    for (size_t used_length = 0; used_length < alignment_length; used_length += 50)
+    {
+        // write header
+        if (used_length != 0)
+            stream << std::endl;
```
rrahn: Please try to avoid endl all the time. This will automatically flush the stream all the time.
Rather use std::flush in the end of your debug stream expression if you need it (so on user side).

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -272,4 +285,99 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+}
+
+} // namespace detail
+
+/*!
+ * \brief True, if each type satisfies aligned_sequence_concept; false otherwise.
+ * \tparam elems The pack of types to be tested.
+ */
+template <typename ...elems>
+inline bool constexpr all_satisfy_aligned_seq = false;
```
rrahn: this should be put into detail namespace.
And we need a generic concept that can differentiate between tuple/vector of aligned sequences, so we don't need to check this all the time. I made a card in the corresponding project.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -272,4 +285,99 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+        // write sequences
+        stream << std::endl << indent;
+        size_t const col_end = std::min(used_length + 50, alignment_length);
+        ranges::for_each(std::get<0>(align) | ranges::view::slice(used_length, col_end) | view::to_char,
+                         [&stream] (char ch) { stream << ch; });
+
+        auto stream_f = [&] (auto const & previous_sequence, auto const & aligned_sequence)
+        {
+            stream << std::endl << indent;
+            auto seq1 = previous_sequence.begin() + used_length;
```
rrahn: Not sure about the current aligned_sequence concept but does it require any container concept or a range concept. In the last case you need to use free function begin.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -272,4 +285,99 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+            auto seq2 = aligned_sequence.begin() + used_length;
```
rrahn: the name of the variables are misleading, as they are iterators. Maybe this should be reflected within the name.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -272,4 +285,99 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+            for (auto it1 = seq1, it2 = seq2;
```
rrahn:
```suggestion
            for_each(view::zip(seq1 | view::slice(used_length, min(size(seq1), used_length + 50)), 
                               seq2 | view::slice(used_length, min(size(seq2), used_length + 50))), 
                     [&stream](auto letters)
                     { 
                         stream << (get<0>(letters) == get<1>(letters) ? '|' : ' ');
                     });
```

I think you don't need the to_char for the comparison above. At the end the alignment container concept or whatever we will call it should make sure that the sequences in the alignmentfile are EqualityComparable.

rrahn: Note that I left out some namespace stuff and I did not test it but I have a feeling that this should work and removes all the entire temporary defintions of variables.

rrahn:
We don't need an option for vectors yet. Let's put it back until we get to the point where we need it (multiple sequence alignment)

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
smehringer: why do you need this?

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -272,4 +285,99 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+ */
+template <typename ...elems>
+inline bool constexpr all_satisfy_aligned_seq = false;
+
+/*!
+ * \brief True, if each type satisfies aligned_sequence_concept; false otherwise.
+ * \tparam elems The pack of types to be tested.
+ */
+template <typename ...elems>
+inline bool constexpr all_satisfy_aligned_seq<type_list<elems...>> = (aligned_sequence_concept<elems> && ... && true);
```
smehringer: isn't
```suggestion
inline bool constexpr all_satisfy_aligned_seq = (aligned_sequence_concept<elems> && ... );
```
enough to make it work?

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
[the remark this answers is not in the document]
joergi-w: Okay, so we stick to the current solution.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
joergi-w: Yes, I agree.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
joergi-w: Okay, I moved it to detail namespace. The concept will be implemented in a separate PR.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
joergi-w: Yes, very good idea! 
And I can apply the `slice` function on the `zip_view` to simplify this even more.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
joergi-w: exactly!

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
smehringer: please move the brief onto this line. This is done throughout the code base and should be consistent, no?
Same goes for all the other brief comments.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -89,15 +98,15 @@ namespace seqan3
...
+/*!\fn      inline seq_type::iterator insert_gap(seq_type & seq, typename seq_type::const_iterator pos_it, typename seq_type::size_type size)
  * \brief   Insert multiple seqan3::gap into an aligned sequence.
  * \relates seqan3::aligned_sequence_concept
  *
  * \tparam        seq_type Type of the range to modify; must model
  *                         seqan3::aligned_sequence_concept.
  * \param[in,out] seq      The aligned sequence to modify.
  * \param[in]     pos_it   The iterator pointing to the position where to insert a gaps.
- * \param[in]     count    The number of gaps to insert.
+ * \param[in]     size     The number of gaps to insert.
```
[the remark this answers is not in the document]
joergi-w: I agree.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -272,4 +285,93 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+            if (rel_pos % 10 == 0)
+                stream << ':';
+            else if (rel_pos % 5 == 0)
+                stream << '.';
+            else
+                stream << ' ';
+        }
+
+        // write first sequence
+        stream << '\n' << indent;
```
rrahn: still, why are you not using std::setw in combination with std::right?

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -272,4 +285,93 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+        ranges::for_each(std::get<0>(align) | ranges::view::slice(begin_pos, end_pos) | view::to_char,
+                         [&stream] (char ch) { stream << ch; });
+
+        auto stream_f = [&] (auto const & previous_seq, auto const & aligned_seq)
+        {
+            // write alignment bars
+            stream << '\n' << indent;
+            ranges::for_each(ranges::zip_view(previous_seq, aligned_seq) | ranges::view::slice(begin_pos, end_pos),
+                             [&stream] (auto ch) { stream << (std::get<0>(ch) == std::get<1>(ch) ? '|' : ' '); });
```
rrahn: the zip stream returns a rvalue. In order to safe a copy you should capture it per `&&`

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
joergi-w: I cannot use `std::right`, because I need left-aligned sequences with a constant indentation. However, now I apply `std::setw(8)` on the empty string to indent by 8 spaces.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -272,4 +285,92 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+    // split alignment into blocks of length 50 and loop over parts
+    for (size_t begin_pos = 0; begin_pos < alignment_size; begin_pos += 50)
+    {
+        size_t const end_pos = std::min(begin_pos + 50, alignment_size);
+
+        // write header line
+        if (begin_pos != 0)
+            stream << '\n';
+
+        stream << std::setw(7) << begin_pos << ' ';
```
marehr: Could we not make the indentation dependent on the value of `alignment_size`?

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -272,4 +285,92 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+        for (size_t rel_pos = 1; rel_pos <= (end_pos - 1) % 50 + 1; ++rel_pos)
```
marehr:
```suggestion
        for (size_t pos = begin_pos + 1; pos <= end_pos; ++pos)
```

You already do a modulo in the loop you don't need to start at position 1, because begin_pos is a multiple of 50,

Let's relax `rel_pos = 1 + k` (where `k in [0, 50)` is the current iteration) to

```
rel_pos' = begin_pos + 1 + k
```

Then

```
        rel_pos' % 10 == 0
<=>     ( begin_pos + 1 + k ) % 10 == 0
<=>     ( begin_pos % 10 + (1 + k) % 10 ) == 0 // begin_pos % 50 is zero, therefore also begin_pos % 10
<=>     ( (1 + k) % 10 ) == 0
<=>     ( rel_pos % 10 ) == 0  // the original condition holds
```
The same holds for `% 5`.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
marehr: Not sure how 'diffable' two outputs are, because we encode the position in the output. This is not a blocker, but I think we should open a follow-up issue to think about it.

joergi-w:
Yes, the implementation of `all_satisfy_aligned_seq` works in this way (as we discussed on the white board) and it may be replaced by an alignment concept in the future.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
joergi-w: I don't see a benefit, as we fix each sequence line to length 50 anyway. 
But fine, we can discuss this in a separate issue #614.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -89,15 +98,15 @@ namespace seqan3
...
+/*!\fn      inline seq_type::iterator insert_gap(seq_type & seq, typename seq_type::const_iterator pos_it, typename seq_type::size_type size)
  * \brief   Insert multiple seqan3::gap into an aligned sequence.
  * \relates seqan3::aligned_sequence_concept
  *
  * \tparam        seq_type Type of the range to modify; must model
  *                         seqan3::aligned_sequence_concept.
  * \param[in,out] seq      The aligned sequence to modify.
  * \param[in]     pos_it   The iterator pointing to the position where to insert a gaps.
- * \param[in]     count    The number of gaps to insert.
+ * \param[in]     size     The number of gap symbols to insert (will result in a gap of length size).
```
rrahn:
```suggestion
 * \param[in]     size     The number of gap symbols to insert (will result in a gap of length `size`).
```

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -191,22 +200,24 @@ inline typename seq_type::iterator insert_gap(seq_type & seq, typename seq_type:
  *                         seqan3::gap.
  * \param[in,out] seq      The container to modify.
  * \param[in]     pos_it   The iterator pointing to the position where to insert gaps.
- * \param[in]     count    The number of gaps to insert.
+ * \param[in]     size     The number of gap symbols to insert (will result in a gap of length size).
```
rrahn:
```suggestion
 * \param[in]     size     The number of gap symbols to insert (will result in a gap of length `size`).
```

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -191,22 +200,24 @@ inline typename seq_type::iterator insert_gap(seq_type & seq, typename seq_type:
...
  *
  * \relates seqan3::aligned_sequence_concept
  *
  * \details
  *
- * This function delegates to the member function `insert(iterator, count, value)`
+ * This function delegates to the member function `insert(iterator, size, value)`
```
rrahn:
```suggestion
 * This function delegates to the member function `insert(iterator, `size`, `value`)`
```

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -272,4 +285,92 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+ * \ingroup             aligned_sequence
+ * \tparam alignment_t  The type of the alignment, must satisfy tuple_like_concept.
+ * \tparam idx          An index sequence.
+ * \param[in] stream    The output stream that receives the formatted alignment.
+ * \param[in] align     The alignment that shall be streamed.
+ */
+template<tuple_like_concept alignment_t, size_t ...idx>
+void stream_alignment(debug_stream_type & stream, alignment_t const & align, std::index_sequence<idx...> const & /**/)
+{
+    size_t const alignment_size = std::get<0>(align).size();
```
rrahn: remove `std::` from `get`.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -272,4 +285,92 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+        stream << '\n' << std::setw(8) << "";
+        ranges::for_each(std::get<0>(align) | ranges::view::slice(begin_pos, end_pos) | view::to_char,
+                         [&stream] (char ch) { stream << ch; });
+
+        auto stream_f = [&] (auto const & previous_seq, auto const & aligned_seq)
+        {
+            // write alignment bars
+            stream << '\n' << std::setw(8) << "";
+            ranges::for_each(ranges::zip_view(previous_seq, aligned_seq) | ranges::view::slice(begin_pos, end_pos),
+                             [&stream] (auto && ch) { stream << (std::get<0>(ch) == std::get<1>(ch) ? '|' : ' '); });
```
rrahn:
```suggestion
                             [&stream] (auto && ch) { stream << (get<0>(ch) == get<1>(ch) ? '|' : ' '); });
```

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -272,4 +285,92 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+
+            // write next sequence
+            stream << '\n' << std::setw(8) << "";
+            ranges::for_each(aligned_seq | ranges::view::slice(begin_pos, end_pos) | view::to_char,
+                             [&stream] (char ch) { stream << ch; });
+        };
+        (stream_f(std::get<idx>(align), std::get<idx + 1>(align)), ...);
```
rrahn:
```suggestion
        (stream_f(get<idx>(align), get<idx + 1>(align)), ...);
```

PR #594, opened 2018-11-28
Index collections

### LINKED ISSUE #602: index collections
Implement a wrapper interface to allow indexing of sequence collections.

eseiler:
Resolves #602

**Main problem**
`bi_fm_index_iterator_collection_test` uses too much memory: ~9GB for me, Travis-VM only has 7.5GB.
<details><summary>Error message (click me)</summary><p>

```
[ 96%] Building CXX object search/fm_index_iterator/CMakeFiles/bi_fm_index_iterator_collection_coverage_test.dir/bi_fm_index_iterator_collection_test.cpp.o
virtual memory exhausted: Cannot allocate memory
search/fm_index_iterator/CMakeFiles/bi_fm_index_iterator_collection_coverage_test.dir/build.make:62: recipe for target 'search/fm_index_iterator/CMakeFiles/bi_fm_index_iterator_collection_coverage_test.dir/bi_fm_index_iterator_collection_test.cpp.o' failed
make[2]: *** [search/fm_index_iterator/CMakeFiles/bi_fm_index_iterator_collection_coverage_test.dir/bi_fm_index_iterator_collection_test.cpp.o] Error 1
make[2]: Target 'search/fm_index_iterator/CMakeFiles/bi_fm_index_iterator_collection_coverage_test.dir/build' not remade because of errors.
CMakeFiles/Makefile2:5969: recipe for target 'search/fm_index_iterator/CMakeFiles/bi_fm_index_iterator_collection_coverage_test.dir/all' failed
make[1]: *** [search/fm_index_iterator/CMakeFiles/bi_fm_index_iterator_collection_coverage_test.dir/all] Error 2
```
</p></details>
<br>

The test only uses `bi_fm_index_iterator<bi_fm_index<std::vector<std::vector<dna4>>>>`
Same problem for `fm_index_iterator_colletion_test` that tests the common functionality of `fm_index_iterator` and `bi_fm_index_iterator` (latter is currently commented out)
<details><summary>Some time -v and -fmem-report stuff (click me)</summary><p>

```
[ 75%] Building CXX object search/fm_index_iterator/CMakeFiles/bi_fm_index_iterator_collection_test.dir/bi_fm_index_iterator_collection_test.cpp.o
Number of expanded macros:                     24228
Average number of tokens per macro expansion:      8

Line Table allocations during the compilation process
Number of ordinary maps used:         1949 
Ordinary map used size:                 60k
Number of ordinary maps allocated:    4096 
Ordinary maps allocated size:          128k
Number of macro maps used:              21k
Macro maps used size:                  874k
Macro maps locations size:            1656k
Macro maps size:                      2531k
Duplicated maps locations size:        601k
Total allocated maps size:            2808k
Total used maps size:                 2591k
Ad-hoc table size:                      12M
Ad-hoc table entries used:           354735
optimized_ranges: 1226617
unoptimized_ranges: 326453

Memory still allocated at the end of the compilation process
Size      Allocated        Used    Overhead
8               684k        227k         20k
16             8564k       5577k        183k
32              137M         75M       2470k
64               39M         25M        638k
256              99M         92M       1388k
512            9060k       6227k        123k
1024             50M         37M        707k
2048             78M         67M       1099k
4096            134M        134M       1887k
8192            265M        265M       1861k
16384           292M        292M       1022k
32768           444M        444M        777k
65536           738M        738M        646k
131072         1042M       1042M        456k
262144          996M        996M        218k
524288          407M        407M         44k
1048576          22M         22M       1232 
2097152        2048k       2048k         56 
4194304        8192k       8192k        112 
8388608        8192k       8192k         56 
16777216         16M         16M         56 
24               75M         40M       1351k
40              172M        102M       2765k
48              186M         28M       2981k
56             9836k       3491k        153k
72               11M       3321k        166k
80               92M         92M       1301k
88               10M       5762k        140k
96               26M         23M        376k
112            7556k       2734k        103k
120            4368k       2232k         59k
152             242M        230M       3400k
128              84M         78M       1182k
144              25M       8527k        355k
168             590M        518M       8265k
176            2972k       2018k         40k
104              32M         28M        461k
368              37M         36M        527k
264             800k        441k         10k
Total          6415M       5892M         36M

String pool
entries		72150
identifiers	61302 (84.96%)
slots		131072
deleted		10848
bytes		8155k (17592186044408M overhead)
table size	1024k
coll/search	0.6983
ins/search	0.0721
avg. entry	115.75 bytes (+/- 194.90)
longest entry	1528
(No per-node statistics)
Type hash: size 262139, 176613 elements, 0.829618 collisions
DECL_DEBUG_EXPR  hash: size 1021, 0 elements, 0.000000 collisions
DECL_VALUE_EXPR  hash: size 4093, 314 elements, 0.992443 collisions
no search statistics
decl_specializations: size 1048573, 642941 elements, 1.817459 collisions
type_specializations: size 524287, 327955 elements, 1.682568 collisions
No gimple statistics
No RTX statistics

Heap vectors                                            Leak           Peak     Times       Leak items Peak items
--------------------------------------------------------------------------------------------------------------------------------------------
--------------------------------------------------------------------------------------------------------------------------------------------
Total                                                      0                        0                0
--------------------------------------------------------------------------------------------------------------------------------------------


Alias oracle query stats:
  refs_may_alias_p: 0 disambiguations, 0 queries
  ref_maybe_used_by_call_p: 0 disambiguations, 0 queries
  call_may_clobber_ref_p: 0 disambiguations, 0 queries
  TBAA oracle: 0 disambiguations 5 queries
               0 are in alias set 0
               0 queries asked about the same object
               5 queries asked about the same alias set
               0 access volatile
               0 are dependent in the DAG
               0 are aritificially in conflict with void *

PTA query stats:
  pt_solution_includes: 0 disambiguations, 2046 queries
  pt_solutions_intersect: 0 disambiguations, 0 queries
[100%] Linking CXX executable bi_fm_index_iterator_collection_test
[100%] Built target bi_fm_index_iterator_collection_test
	Command being timed: "make bi_fm_index_iterator_collection_test"
	User time (seconds): 524.51
	System time (seconds): 24.06
	Percent of CPU this job got: 94%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 9:41.67
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 8615476
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 8279
	Minor (reclaiming a frame) page faults: 4772575
	Voluntary context switches: 13195
	Involuntary context switches: 677196
	Swaps: 0
	File system inputs: 0
	File system outputs: 0
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 84
	Page size (bytes): 4096
	Exit status: 0
```

</p></details>
<br>

- [X] Snippets
- [x] Make Debug tests work on travis (coverage tests)
- - After above: Check if there's redundant code in test cases that can be refactored
- [x] Docs
- [X] FM
- [X] Bi-FM
- [x] Clean up code
- [x] Put debug stream for pairs into separate PR

<details><summary>Old questions (click me)</summary><p>

<details><summary>Regarding the output order (click me)</summary><p>

```cpp
int main()
{
    std::vector<std::vector<dna4>> vec { "ACAGCGAGC"_dna4, "GTTTCGAT"_dna4,
                                         "GCGATCGTAGCT"_dna4, "CACGCG"_dna4};
    fm_index index{vec};
    auto it = index.begin();
    it.extend_right("GCG"_dna4);
    debug_stream << it.query() << '\n';
    debug_stream << it.last_char() << '\n';
    for (auto && x : it.locate())
        debug_stream << x.first << '\t' << x.second << '\n';
}
```

Would print:
```
GCG    
G
0    3
3    3
2    0
```

Which is all correct, but the <sequence_no, sequence_pos> pairs are not ordered by sequence_no since the order in the index does not correspond with the order of the texts in the collection.

Should it be ordered?

**Answer**: It's fine but should be documented.

</p></details>
<br>

<details><summary>Copying the iterator multiple times in lazy_locate() (click me)</summary><p>

```cpp
return ranges::view::iota(node.lb, node.lb + count())
               | view::transform([*this, _offset = offset()] (auto sa_pos) { return _offset - index->index[sa_pos]; })
               | view::filter([*this, _query_length = query_length()] (auto loc)
               {
                   return (index->text_begin_rs.rank(loc + 1) == index->text_begin_rs.rank(loc + _query_length));
               })
               | view::transform([*this] (auto loc)
               {
                   size_type sequence_rank = index->text_begin_rs.rank(loc + 1);
                   size_type sequence_position = loc - index->text_begin_ss.select(sequence_rank);
                   return std::make_pair(sequence_rank-1, sequence_position);
               });
```

Do I need to copy the fm_index_iterator multiple times?

```cpp
| view::transform([*this, _offset = offset()] 
| view::filter([*this, _query_length = query_length()]
| view::transform([*this] 
```

**Answer**: It's fine since there's only stack memory involved when copying.

</p></details>
<br>
<details><summary>Boundary check (click me)</summary><p>

Currently, we are checking the validity of the match when calling `locate()` or `lazy_locate()`.
Actually, we should check for it, when moving in the index (e.g. `extend_right()`).
`locate()` will not always be called, so incorrect iterators might be used downstream.
But since the SA range is contiguous, it's hard to modify it to exclude matches crossing the boundaries that may be in the middle of the SA range.

**Answer**: Resolved by using delimiters.

</p></details>
<br>
</p></details>
<br>


include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -249,6 +263,44 @@ class fm_index
...
+        //!\cond
+        requires is_collection
+        //!\endcond
+    {
+        // text must not be empty
+        if (text.begin() == text.end())
+            throw std::invalid_argument("The text that is indexed cannot be empty.");
+        this->text = &text;
+
+        concatenated_sequences<value_type_t<text_t>> cvec{text};
```
smehringer: Check whether cvec is empty too? The above check only check whether the vector of vectors is non-empty but not if the contained vectors are also non-empty.

include/seqan3/io/stream/debug_stream.hpp
```diff
@@ -295,6 +295,32 @@ inline debug_stream_type & operator<<(debug_stream_type & s, alphabet_t const l)
...
+//!\endcond
+{
+    return s << '<' << p.first << ',' << p.second << '>';
+}
+
+//!\copydoc debug_stream_type & operator<<(debug_stream_type & s, std::pair<pair_t1, pair_t2> & p)
+template <typename pair_t1, typename pair_t2>
+inline debug_stream_type & operator<<(debug_stream_type & s, std::pair<pair_t1, pair_t2> const & p)
+//!\cond
+    requires requires (pair_t1 l) { { debug_stream << l }; } && requires (pair_t2 l) { { debug_stream << l }; }
```
marehr: Is this a new style that we put the requires in the trailing location?

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -223,6 +234,9 @@ class fm_index
      * No guarantees.
      */
     void construct(text_t const & text)
+        //!\cond
+        requires !is_collection
```
marehr: trailing location?

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -249,6 +263,44 @@ class fm_index
         // index.m_sigma = largest_char;
     }
 
+    //!\overload
+    void construct(text_t const & text)
+        //!\cond
+        requires is_collection
```
marehr: trailing location?

include/seqan3/io/stream/debug_stream.hpp
```diff
@@ -295,6 +295,32 @@ inline debug_stream_type & operator<<(debug_stream_type & s, alphabet_t const l)
...
+ * \param p The std::pair.
+ * \relates seqan3::debug_stream_type
+ */
+template <typename pair_t1, typename pair_t2>
+inline debug_stream_type & operator<<(debug_stream_type & s, std::pair<pair_t1, pair_t2> & p)
+//!\cond
+    requires requires (pair_t1 l) { { debug_stream << l }; } && requires (pair_t2 l) { { debug_stream << l }; }
+//!\endcond
+{
+    return s << '<' << p.first << ',' << p.second << '>';
```
marehr: I think `(,)` would be better delimiters. I would expect that a tuple (also the special case pair) returns `(a,b,c,d)`

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -166,6 +174,9 @@ class fm_index
     using iterator_type = fm_index_iterator<fm_index<text_t, fm_index_traits>>;
     //!\}
 
+    //!\brief Indicates whether index is built over a collection
+    static bool constexpr is_collection = dimension_v<text_t> == 2;
```
marehr: what is with dimension 3 and higher?

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -249,6 +263,44 @@ class fm_index
...
+
+        concatenated_sequences<value_type_t<text_t>> cvec{text};
+        // size + 1: for the rank query we need to go one to the right: rank=#1 BEFORE (i.e. excluding) the position
+        sdsl::bit_vector pos(cvec.concat_size()+1, 0);
+        size_t prefix_sum{0};
+        for (auto && x : cvec)
+        {
+            pos[prefix_sum] = 1;
+            prefix_sum += x.size();
+        }
```
marehr: Doesn't the `concatenated_sequences` already do this?!

marehr:
Tests are missing right now, for the collections extension.

include/seqan3/io/stream/debug_stream.hpp
eseiler: Maybe, I'll put it in another PR once I'm done. But I need it here for debugging and I might as well already get some feedback on it :D

include/seqan3/search/fm_index/fm_index.hpp
eseiler: Well, I wouldn't know what to do with it.
If it's an collection of collections, it would need to be flattened to a collection to build an index on it.
I'd guess currently it will crash during compiling, because something doesn't work - just like if you try to index a collection in the current version.

Another question would be, if this could be treated more elegantly, i.e. you actually know that it fails because you have to many dimensions.
Maybe adjust the `requires` to only be defined for dimension==1 and dimension==2?

include/seqan3/search/fm_index/fm_index.hpp
eseiler: Does it? In the docs, I only see `size` for the number of elements in the collection and `concat_size` for the cumulative sum.
But I need the individual sizes of the sequences.

include/seqan3/io/stream/debug_stream.hpp
eseiler: What does trailing location mean in this context?

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -152,6 +152,13 @@ class fm_index
     //!\brief Pointer to the indexed text.
     text_t const * text = nullptr;
 
+    //!\brief Bitvector storing begin positions for collections
```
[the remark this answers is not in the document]
eseiler: I guess you could say I missed the point.

include/seqan3/search/fm_index/fm_index.hpp
smehringer: Since there is no template I think the trailing location looks better here

include/seqan3/io/stream/debug_stream.hpp
rrahn: Well, I particularluy find it less readable, but @h-2 seems to use it as it does not need to be guarded with a doxygen condition. But honestly I prefer the other way, because I see immediately the constraints on the types. But we can discuss this on the strategy meeting.

h-2: For member functions of class templates (VS function templates) the trailing constraints notation is the only legal notation.
For function templates I also prefer the other notation, but if it means adding extra guards I usually switch to trailing notation, also because it means that the constraint is visible in doxygen.

eseiler:
Question:

The (bi)_fm_index_concept won't hold for collections:

```cpp
requires requires (t index, std::vector<typename t::char_type> const text)
    {
        { t(text) };
        { index.construct(text) } -> void;
    };
```

with `using char_type = innermost_value_type_t<text_t>;`.
So it will resolve, e.g. `std::vector<std::vector<dna4>>` to `dna4` and then try to construct an index with `std::vector<dna4>` which a fm_index for `std::vector<std::vector<dna4>>` cannot do.

So, either:

1) Change 
```cpp
using char_type = innermost_value_type_t<text_t>;
```
to
```cpp
using char_type = value_type_t<text_t>;
```

2) Change 
```cpp
requires requires (t index, std::vector<typename t::char_type> const text)
``` 
to 
```cpp
requires requires (t index, std::vector<value_type_t<typename t::text_type>> const text)
```

3) Change 
```cpp
requires requires (t index, std::vector<typename t::char_type> const text)
``` 
to 
```cpp
requires requires (t index, typename t::text_type const text)
```

include/seqan3/search/fm_index/fm_index.hpp
h-2: you can make a static_assert that the dimension is 1 or 2

include/seqan3/search/fm_index/fm_index.hpp
h-2: It is accessible via `.data()` but it shouldn't be so don't rely on that for now. Using view::join is the generic approach.

eseiler:
I transferred the debug_stream changes to #596

eseiler:
I added changes to use a delimiter for the texts.

➕iterator now won't cross the boundaries
➖increases alphabet size by 1 
➖char alphabet currently not indexable (256 + 1 > 8 bit unsigned int)

I also managed to reduce the number of text copies to 1 (only the copy text -> sdsl data type is done now).

smehringer:
Oh oh Code coverage misses :point_up:  (I really like this report)

eseiler:
I'll have a look at it once we can actually build the tests in debug mode.

h-2:
> Regarding the output order (click me)

Not sorting is the way to go I think. If people want it sorted, they can always sort it themselves. But it should be documented clearly.

> I added changes to use a delimiter for the texts.

This looks like the most reasonable thing to do, but will this affect the performance of the searches on wavelet-trees / depth of the wt? And do you know what the 0-rank is used for by the SDSL? Could you maybe hijack that?

For `char` you could scan the input for a character that is not used and use that as delimiter (and throw if all of the char characters are used).

> Copying the iterator multiple times in lazy_locate() (click me)

If you use delimiters, do you still need to filter? in general it is not a problem to copy the iterator/cursor, because it has only stack storage AFAIK.

> Boundary check (click me)

This is solved by using delimiters, right?

eseiler:
> > Regarding the output order (click me)
> 
> Not sorting is the way to go I think. If people want it sorted, they can always sort it themselves. But it should be documented clearly.

👍 

> 
> > I added changes to use a delimiter for the texts.
> 
> This looks like the most reasonable thing to do, but will this affect the performance of the searches on wavelet-trees / depth of the wt? And do you know what the 0-rank is used for by the SDSL? Could you maybe hijack that?
0 is used as sentinel. Using delimiters should affect the performance since we add another symbol to the alphabet.
> 
> For `char` you could scan the input for a character that is not used and use that as delimiter (and throw if all of the char characters are used).

Currently I'm reserving rank 255 for char. If the char is signed, 255 will never be used anyway, for unsigned chars it's `ÿ` 

> > Copying the iterator multiple times in lazy_locate() (click me)
> 
> If you use delimiters, do you still need to filter? in general it is not a problem to copy the iterator/cursor, because it has only stack storage AFAIK.

No I don't filter there anymore, it's an old snippet. 

> > Boundary check (click me)
> 
> This is solved by using delimiters, right?

Yes

include/seqan3/search/fm_index/bi_fm_index.hpp
```diff
@@ -222,7 +232,10 @@ class bi_fm_index
             throw std::invalid_argument("The text that is indexed cannot be empty.");
 
         this->text = &text;
-        rev_text = view::reverse(text);
+        if constexpr(is_collection)
+            rev_text = text | view::deep{view::reverse} | view::deep{view::persist} | view::reverse;
```
rrahn: could you not reuse the declared type from above to omit duplicate code?

include/seqan3/search/fm_index/bi_fm_index_iterator.hpp
```diff
@@ -929,6 +958,28 @@ class bi_fm_index_iterator
...
+        assert(index != nullptr);
+
+        std::vector<std::pair<size_type, size_type>> occ;
+        occ.reserve(count());
+        for (size_type i = 0; i < count(); ++i)
+        {
+            size_type loc = offset() - index->fwd_fm.index[fwd_lb + i];
+            size_type sequence_rank = index->fwd_fm.text_begin_rs.rank(loc + 1);
+            size_type sequence_position = loc - index->fwd_fm.text_begin_ss.select(sequence_rank);
+            occ.push_back(std::make_pair(sequence_rank - 1, sequence_position));
```
rrahn:
```suggestion
            occ.emplace_back(sequence_rank - 1, sequence_position);
```

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -223,10 +255,21 @@ class fm_index
      * No guarantees.
      */
     void construct(text_t const & text)
+        //!\cond
+        requires !is_collection
+        //!\endcond
     {
          // text must not be empty
-        if (text.begin() == text.end())
+        if (std::begin(text) == std::end(text))
```
rrahn:
```suggestion
        if (std::ranges::begin(text) == std::ranges::end(text))
```

include/seqan3/search/fm_index/bi_fm_index.hpp
eseiler: Mhm, I'd like to, but I didn't find any way to do it.

I need something like 
```cpp
static auto constexpr rev_text_view = std::conditional<is_collection, view::deep{view::reverse} | view::reverse, view::reverse>;
using rev_text_type = decltype (*text | rev_text_type);
```
But in a working version.

include/seqan3/search/fm_index/bi_fm_index.hpp
```diff
@@ -94,13 +95,22 @@ class bi_fm_index
...
+
     /*!\name Text types
      * \{
      */
     //!\brief The type of the forward indexed text.
     using text_type = text_t;
     //!\brief The type of the forward indexed text.
-    using rev_text_type = decltype(view::reverse(*text));
+    using rev_text_type = std::conditional_t<is_collection,
+                                            decltype(*text | view::deep{view::reverse} | view::deep{view::persist}
```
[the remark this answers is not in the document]
eseiler: Good catch 👍 

I know that I needed it there and that otherwise it wouldn't compile.
I probably changed something in the `fm_index` and did not revisit the `rev_text_type`.

include/seqan3/search/fm_index/fm_index.hpp
eseiler: What's the reasoning behind this? Is it more general?

include/seqan3/search/fm_index/bi_fm_index.hpp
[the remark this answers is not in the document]
eseiler: Actually I need it, because otherwise the `to_rev_iterator` function of the `bi_fm_index` cannot use the view.
I don't really understand why, because it's just a long stack trace, complaining that it cannot bind it without the `view::persist`.

include/seqan3/search/fm_index/fm_index.hpp
rrahn: `std::begin` is a function. The compiler would only look in the begin functions qualified by the std namespace. This means, that if you have an object with a global begin function that is not in std namespace, the correct begin cannot be found. The alternative would be to using declare `using std::begin` in the scope of the function and call it without the namespace qualifier. This would allow for ADL to kick in. The `std::ranges::begin` is a function object and allows for ADL without this using declaration.

smehringer: I think we aliased the `std::ranges::begin` so you can now just use `begin`

include/seqan3/search/fm_index/bi_fm_index.hpp
rrahn: Well yes, it is not so easy I guess, because you cannot propagate the object throught the views after you instantiated it. But in any case you could make a little detail helper function that makes the transformation. This can be reused to declare the type and to define the reverse text in the function

include/seqan3/search/fm_index/bi_fm_index.hpp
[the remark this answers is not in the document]
rrahn: Ok, we need to investigate this. Can you please add a card for this on the project site for the search module

rrahn:
@ I will give you a final review this day

include/seqan3/search/fm_index/fm_index.hpp
rrahn: Well someone should document them.

smehringer: http://docs.seqan.de/seqan/3.0.0-master-user/group__ranges.html#ga8b1352687dc82a2d4922da34778ca215

rrahn: That's the `std::ranges::` aliases. I meant the seqan3 versions.

h-2:
what is the status here?

eseiler:
Let's see if this works.
I adapted the view for fm-indices over collections to copy once
https://github.com/seqan/seqan3/blob/767c2d60be70932c8bf01c201b534c463e538f5d/include/seqan3/search/fm_index/fm_index.hpp#L346-L362
And it takes considerably less memory to build it in debug mode.

h-2:
Is this done from your POV?

eseiler:
From my POV this is ready to be merged.
For the additional stuff I want to have a look at, I already created cards.

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -78,7 +79,19 @@ class bi_fm_index_iterator;
...
+ * For an index over a text collection a delimiter is added inbetween the texts. This causes sigma to increase by 1.
+ * \attention For any alphabet, the symbol with rank 255 is not allowed to occur in the text. Addtionally,
+ *            rank 254 cannot occur when indexing text collections.
+ *
+ * \if DEV
+ * Rank 255 is generally not allowed. When constructing the index, we increase the rank of every letter by 1, since the
+ * SDSL uses 0 as a sentinel. Incrementing 255 by 1 will cause the rank (uint8_t) to overflow to 0.
+ * For text collections, 255 is reserved as delimiter between the individual texts. Therefore the letter with rank 254
+ * cannot occur in the text.
+ * \endif
```
h-2: are there any sanity checks for this in the code? When you do the copying you could actually check the value and throw if it is out-of-range.

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -223,10 +255,21 @@ class fm_index
      * No guarantees.
      */
     void construct(text_t const & text)
+        //!\cond
+        requires !is_collection
+        //!\endcond
     {
          // text must not be empty
-        if (text.begin() == text.end())
+        if (std::ranges::begin(text) == std::ranges::end(text))
```
h-2: you can do just begin and end

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -223,10 +255,21 @@ class fm_index
...
             throw std::invalid_argument("The text that is indexed cannot be empty.");
+
+        // Since we will later calculate rank + 1, the rank 255 + 1 will overflow to 0 which is reserved by sdsl
+        assert(!std::ranges::any_of(text,
+                                    [] (char_type r)
+                                    {
+                                        return to_rank(r) == static_cast<underlying_rank_t<char_type>>(255);
+                                    }));
```
h-2: ah, I see, here's the check. But the cast is unsafe again, see the previous comment regarding rank_t.

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -235,10 +278,10 @@ class fm_index
         // uint8_t largest_char = 0;
         sdsl::int_vector<8> tmp_text(text.size());
```
h-2: do you static_assert somewhere that the alphabet size can b 256 at most?

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -235,10 +278,10 @@ class fm_index
...
 
-        //TODO view::reverse is broken and can't be chained right now
-        std::vector<value_type_t<text_t>> another_copy = text | view::reverse;
-
-        std::ranges::copy(another_copy | view::to_rank | view::transform([] (uint8_t const r) { return r + 1; }),
+        std::ranges::copy(text
+                          | view::to_rank
+                          | view::transform([] (uint8_t const r) { return r + 1; })
```
h-2: I would instead insert into the lambda:
```cpp
| view::transform([] (uint8_t const r)
{
    if constexpr (alphabet_size_v<char_type> >= 256 - is_collection)
        if (r >= 256 - 1- is_collection)
            throw std::out_of_range("The input text cannot be indexed, because for full character alphabets the last one/two values are reserved (single sequence/collection).");
    return r + 1;
}
```

This way you make sure that the check is performed in release mode, too, but that you also only perform it if the underlying alphabet could possibly have such high values.

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -249,6 +292,78 @@ class fm_index
...
+
+        size_t concatenated_size{0};
+        // there must be at least one non-empty text in the collection
+        bool all_empty = true;
+        for (auto && t : text)
+        {
+            if (std::begin(t) != std::end(t))
+            {
+                all_empty = false;
+                concatenated_size += 1 + t.size(); // text size and delimiter
```
h-2: why do you filter empty sequences? Doesnt this make identifying the right sequence on locate() pretty difficult?

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -249,6 +292,78 @@ class fm_index
...
+            }
+        }
+        if (all_empty)
+            throw std::invalid_argument("A text collection that only contains empty texts cannot be indexed.");
+        this->text = &text;
```
h-2: it's better to just use different names for parameters and member variables...

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -249,6 +292,78 @@ class fm_index
...
+            prefix_sum += t.size() + 1;
+        }
+
+        text_begin    = sdsl::sd_vector(pos);
+        text_begin_ss = sdsl::select_support_sd<1>(&text_begin);
+        text_begin_rs = sdsl::rank_support_sd<1>(&text_begin);
+
+        sdsl::int_vector<8> tmp_text(concatenated_size - 1); // last text in collection needs no delimiter
+
+        uint8_t delimiter = alphabet_size_v<char_type> == 256 ? 255 : alphabet_size_v<char_type> + 1;
```
h-2: there are quite a few implicit conversions going on here that are dangerous. Like I said above, the type of alphabet_size_v could also be bool so please cast it explicitly up to `uint8_t` before doing the comparison and the addition.

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -249,6 +292,78 @@ class fm_index
...
+
+        std::vector<uint8_t> tmp = text
+                                   | view::filter([&] (auto && t) { return !std::empty(t); })
+                                   | view::deep{view::to_rank}
+                                   | view::deep{view::transform([] (uint8_t const r) { return r + 1; })}
+                                   | ranges::view::join(delimiter);
```
h-2: ah, I see that you have different functions for construct so you would need to apply my lambda from above here, too (and can replace the `is_collection` inside the lambda).

test/unit/search/fm_index/bi_fm_index_char_test.cpp
```diff
@@ -0,0 +1,43 @@
...
+    std::vector<std::vector<unsigned char>> text{{'a', 'b'}, {'a', 'u', ',', c, '0'}};
+
+    EXPECT_DEATH(bi_fm_index_t bi_index{text}, "Assertion failed: \\(!std::ranges::any_of\\(text \\|"
+                                               " ranges::view::join, \\[] \\(char_type r\\) { return to_rank\\(r\\) =="
+                                               " static_cast<underlying_rank_t<char_type>>\\(255\\) \\|\\|"
+                                               " to_rank\\(r\\) == static_cast<underlying_rank_t<char_type>>\\(254\\);"
+                                               " }\\)\\), function construct,.*");
+}
+#endif
+*/
```
h-2: why commented out?

include/seqan3/search/fm_index/bi_fm_index.hpp
```diff
@@ -288,9 +301,9 @@ class bi_fm_index
     // }
 
     /*!\brief Returns a seqan3::bi_fm_index_iterator on the index that can be used for searching.
-     *        \cond DEV
+     *        \if DEV
      *            Iterator is pointing to the root node of the implicit affix tree.
-     *        \endcond
+     *        \endif
```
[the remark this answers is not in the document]
eseiler: http://www.doxygen.nl/manual/commands.html#cmdif

`\cond` and `\endcond` don't work within comments.

include/seqan3/search/fm_index/bi_fm_index.hpp
```diff
@@ -326,6 +339,7 @@ class bi_fm_index
     /*!\brief Returns a unidirectional seqan3::fm_index_iterator on the reversed text of the bidirectional index that
      *        can be used for searching. Note that because of the text being reversed, extend_right() resp. cycle_back()
      *        correspond to extend_left() resp. cycle_front() on the bidirectional index iterator.
+     * \attention For text collections the text IDs are also reversed.
```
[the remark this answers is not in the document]
eseiler: I tried it once, but then other stuff broke.
I'll have another look.

include/seqan3/search/fm_index/bi_fm_index_iterator.hpp
```diff
@@ -126,6 +126,12 @@ class bi_fm_index_iterator
     size_type rev_rb;
     //\}
 
+    //!\brief Alphabet size of the index without delimiters
+    underlying_rank_t<typename index_type::char_type> sigma;
```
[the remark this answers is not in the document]
eseiler: Mhm, in theory yes. But why would you build an index over the mask alphabet 🤔 
But we can do something more fail safe here.

include/seqan3/search/fm_index/fm_index.hpp
eseiler: For the fm-index:
`std::Same<underlying_rank_t<innermost_value_type_t<text_t>>, uint8_t>`

... which also has the problems with the rank_type

include/seqan3/search/fm_index/bi_fm_index.hpp
```diff
@@ -94,13 +95,22 @@ class bi_fm_index
...
     /*!\name Text types
      * \{
      */
     //!\brief The type of the forward indexed text.
     using text_type = text_t;
     //!\brief The type of the forward indexed text.
-    using rev_text_type = decltype(view::reverse(*text));
+    using rev_text_type = std::conditional_t<is_collection,
+                                             decltype(*text | view::deep{view::reverse} | view::deep{view::persist}
+                                                            | view::reverse),
```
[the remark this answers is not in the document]
eseiler: Because for the bi_fm_index we need the text and the reversed text.
Example (ignoring the rank stuff)

text = `[ACGT, TGC]`
delimiter = `#`
flattened text = `ACGT#TGC`
reversed text = `CGT#TGCA`

text | deep_reverse = `[TGCA,CGT]`
^ | reverse = `[CGT,TGCA]`
^ flattened = `CGT#TGCA`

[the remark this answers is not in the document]
h-2: I know that the bifm needs the text in both direction, but why does "in both directions" mean that you need to reverse the order? I don't think this is being done in SeqAn2, is it?

include/seqan3/search/fm_index/bi_fm_index_iterator.hpp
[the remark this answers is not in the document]
h-2: the rank type of char is also uint8_t, but it's size cannot be represented by uint8_t!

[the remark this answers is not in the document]
eseiler: But it has to, since we work with `int_vector<8>`. That's why we have to throw if we find 255 (single) or 254/255 (collections), because we can otherwise not fit the alphabet into `uint8_t`.

include/seqan3/search/fm_index/bi_fm_index.hpp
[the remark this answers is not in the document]
eseiler: The sdsl bi_fm works on single texts, where one is forward and one is reversed. If we do not reverse the second one after deep_reversing it, it's not the true reverse of the text (seen as single sequence) and hence the synchronisation between both directions will fail.

include/seqan3/search/fm_index/bi_fm_index_iterator.hpp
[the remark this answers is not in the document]
h-2: Yes, but the alphabet_size of `char` is still 256 which doesn't fit into `uint8_t`!! You can choose to ignore/misuse the last one or two values, but that does not change the alphabet_size.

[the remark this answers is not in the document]
eseiler: Ah, yeah right.

I can just expose the sigma type via `using sdsl_sigma_type = typename sdsl_index_type::alphabet_type::sigma_type;` in the fm_index, and then use this (though it's always an uint16_t for the implemented alphabets.

include/seqan3/search/fm_index/bi_fm_index.hpp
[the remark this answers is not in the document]
h-2: Hm, yeah, I understand. :+1:

include/seqan3/search/fm_index/bi_fm_index_iterator.hpp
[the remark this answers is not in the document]
h-2: Or just use size_t?

include/seqan3/search/fm_index/fm_index.hpp
h-2: why is that?

h-2: (include seqan3/range/shortcuts.hpp)

include/seqan3/search/fm_index/fm_index.hpp
h-2: I would instead check that the size is <=256 with a comment that says that 256 and 255 are not really legally, but have a workaround that throws.

include/seqan3/search/fm_index/fm_index.hpp
h-2: I correct myself, don't case alphabet_size_v to uint8_t as we just established that it might not fit.

test/unit/search/fm_index/bi_fm_index_char_test.cpp
h-2: Hm, I don't understand :thinking:

include/seqan3/search/fm_index/bi_fm_index_iterator.hpp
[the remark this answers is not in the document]
eseiler: Would also work. I guess it's till cheap to copy?

include/seqan3/search/fm_index/fm_index.hpp
eseiler: So no cast at all?

test/unit/search/fm_index/bi_fm_index_char_test.cpp
eseiler: I also do not really understand.

If I do an assert and use `EXPECT_DEATH`, the debug unit test works fine, but the coverage test will actually be terminated by the assertion.
Internally, google test spawns a new process that then executes the death test. Maybe lcov intercepts this and handles the assertion? (it works without the death test)

Anyway, I now use the lambda with the throw and it works just fine 🎉

include/seqan3/search/fm_index/fm_index.hpp
eseiler: So, I can either do
`seqan3::begin`
`std::ranges::begin` or
`ranges::begin`

just `begin` can't be resolved
```
error: no matching function for call to 'seqan3::fm_index<std::vector<std::vector<seqan3::dna4> >, seqan3::fm_index_default_traits>::begin(const std::vector<std::vector<seqan3::dna4> >&)'
        if (begin(text) == end(text))
note: candidate: seqan3::fm_index<text_t, fm_index_traits>::cursor_type seqan3::fm_index<text_t, fm_index_traits>::begin() const [with text_t = std::vector<std::vector<seqan3::dna4> >; fm_index_traits = seqan3::fm_index_default_traits; seqan3::fm_index<text_t, fm_index_traits>::cursor_type = seqan3::fm_index_cursor<seqan3::fm_index<std::vector<std::vector<seqan3::dna4> >, seqan3::fm_index_default_traits> >]
        cursor_type begin() const noexcept
```

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -78,7 +79,19 @@ class bi_fm_index_cursor;
...
+ * For an index over a text collection a delimiter is added inbetween the texts. This causes sigma to increase by 1.
+ * \attention For any alphabet, the symbol with rank 255 is not allowed to occur in the text. Addtionally,
+ *            rank 254 cannot occur when indexing text collections.
+ *
+ * \if DEV
+ * Rank 255 is generally not allowed. When constructing the index, we increase the rank of every letter by 1, since the
+ * SDSL uses 0 as a sentinel. Incrementing 255 by 1 will cause the rank (uint8_t) to overflow to 0.
+ * For text collections, 255 is reserved as delimiter between the individual texts. Therefore the letter with rank 254
+ * cannot occur in the text.
+ * \endif
```
h-2: Add a paragraph about the limitations regarding alphabet size and which codeparts would need to be changed to make this more generic.

PR #597, opened 2018-12-01
[FIX] Fixes missing aligned_alloc on mac os

rrahn:
On macos glibc does not implement aligned_alloc 😖 so we need to fallback to posix_memalign.

### DIFF
--- a/include/seqan3/range/container/aligned_allocator.hpp
+++ b/include/seqan3/range/container/aligned_allocator.hpp
@@ -126,6 +126,12 @@ public:
         if (n > max_size)
             throw std::bad_alloc();
 
+        // NOTE: On macOS glibc does not implement aligned_alloc, so we need to fallback to posix_memalign instead.
+#if defined(__APPLE__) && (!defined(_GLIBCXX_HAVE_ALIGNED_ALLOC) && !defined(_ISOC11_SOURCE))
+        void * p{};
+        if (int res = posix_memalign(&p, alignment, n * sizeof(value_type)); res == 0 && p != nullptr)
+            return static_cast<pointer>(p);
+#else
         // NOTE:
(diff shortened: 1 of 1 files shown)


h-2:
I assume this is the same as on FreeBSD and its just a bug in GCC7 (GCC8 should work) and it has been fixed already for GCC7.4 which will be released very soon.

rrahn:
It does not seem to be a bug of gcc. It is a problem of the OS shipped glibc [see](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69680). And on mac os this is not available. Neither macports nor brew installed gcc support them. Note that I installed most recent libgcc version and tested with gcc8. I will install gcc9 as well and try it but I opened already a bug report to apple to see what they have to say.

rrahn:
as expected a newer version of gcc does not fix the issue.

h-2:
Hm, ok, just FYI, I was referring to this bug:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=83662
I am pretty sure macOS uses the freebsd libc, but it might be outdated...

h-2:
PS libc != libgcc

rrahn:
```
==89923== LEAK SUMMARY:
==89923==    definitely lost: 0 bytes in 0 blocks
==89923==    indirectly lost: 0 bytes in 0 blocks
==89923==      possibly lost: 72 bytes in 3 blocks
==89923==    still reachable: 72,904 bytes in 7 blocks
==89923==         suppressed: 22,424 bytes in 153 blocks
==89923== Reachable blocks (those to which a pointer was found) are not shown.
==89923== To see them, rerun with: --leak-check=full --show-leak-kinds=all
```

Does not seem to leak.

rrahn:
jenkins is not responding so I'll merge it now.

PR #598, opened 2018-12-03
[DOC] Move example in subrange.hpp to test/snippet/std/view/subrange.cpp

### LINKED ISSUE #581: Correct doc example for subrange view
The example for view::subrange should be exported into a snippet and then corrected.
Since it is a type alias template type deduction can't work.

joshuak94:
Fixes #581


test/snippet/std/view/subrange.cpp
```diff
@@ -0,0 +1,16 @@
...
+#include <seqan3/alphabet/nucleotide/dna4.hpp>
+#include <seqan3/range/view/to_char.hpp>
+#include <seqan3/io/stream/debug_stream.hpp>
+
+using namespace seqan3;
+
+int main()
+{
+//! [example]
+dna4_vector s{"ACTTTGATAN"_dna4};
```
rrahn: There is a `N` in a dna4 vector?

joshuak94: It automatically is converted to an A by default. I think this was intentional, since the comment also shows that it gets converted to an A? Or should I just get rid of the N.

smehringer: Since the `N` to `A` conversion of dna4 has nothing to do with the subrange view I would get rid of the N because it might be confusing if you aren't accustomed to our alphabets.

test/snippet/std/view/subrange.cpp
```diff
@@ -0,0 +1,16 @@
...
+auto v1 = view::subrange<decltype(begin(s)), decltype(end(s))>{begin(s) + 2, end(s)} | view::to_char; // == "TTTGATAA"
```
smehringer:
```suggestion
auto v1 = view::subrange<decltype(begin(s) + 2), decltype(end(s))>{begin(s) + 2, end(s)} | view::to_char; // == "TTTGATAA"
```
to be completely sure? @rrahn  Does it ever make a difference?

test/snippet/std/view/subrange.cpp
```diff
@@ -0,0 +1,16 @@
...
+//! [example]
+
+debug_stream << v1 << '\n';
```
smehringer: why this?

marehr: S.t. `v1` is not unused and for testing purposes.

test/snippet/std/view/subrange.cpp
marehr: depends on the implementer of `iterator+2`. In this case it makes no difference.

test/snippet/std/view/subrange.cpp
joshuak94: Yeah. I could have just done `(void) v1;` but I wanted to print it just to make sure it worked as expected.

marehr:
@rrahn can you approve?

marehr:
@rrahn please re-review

test/snippet/std/view/subrange.cpp
```diff
@@ -0,0 +1,16 @@
...
+#include <seqan3/alphabet/nucleotide/dna4.hpp>
+#include <seqan3/range/view/to_char.hpp>
+#include <seqan3/io/stream/debug_stream.hpp>
+
+using namespace seqan3;
+
+int main()
+{
+//! [example]
+dna4_vector s{"ACTTTGATAA"_dna4};
```
rrahn: For readability I propose using this:

```cpp
dna4_vector s{"ACTTTGATAA"_dna4};
using iterator = dna4_vector::iterator;
auto v1 = view::subrange<iterator, iterator>{s.begin() + 2, s.end()} | view::to_char; 
```

PR #599, opened 2018-12-03
[FEATURE] Adds metafunction to defer instantiation of crtp_base classes.

rrahn:
This helper meta-function allows augmentation of crtp classes with traits before the derived type is known.
At the moment I will use it for the alignment configuration but I saw it as general enough to put it into core metafunction.

### DIFF
--- a/test/snippet/core/metafunction/deferred_crtp_base.cpp
+++ b/test/snippet/core/metafunction/deferred_crtp_base.cpp
@@ -1,0 +1,51 @@
+#include <string>
+#include <type_traits>
+
+#include <seqan3/core/metafunction/deferred_crtp_base.hpp>
+
+// Defines a crtp_base class with an additional value type.
+template <typename derived_t, typename value_t>
+class base1
+{
+public:
+
+    value_t func1() const
+    {
+        return {"instance of base1"};
+    }
+};
+
+// Defines a crtp_base class with an additional value type and a parameter type.
+template <typename derived_t, typename value_t, typename parameter_t>
+class base2
+{
+public:
+
+    value_t func2(parameter_t const p) const
+    {
+        return static_cast<value_t>(p);
+    }
+};
+
+// The derived class that inherits from a variadic crtp pattern, which are augmented with additional trait types.
+// These types must be wrapped in a deferred layer, otherwise the compilation fails as incomplete types are not allowed.
+// But during the definition of the base classes, the derived class cannot be known.
+// In addition the deferred type must be invoked with the derived class using the `invoke_deferred_crtp_base` helper
+// template to instantiate the correct crtp base type.
+template <typename ...deferred_bases_t>
+class derived : public seqan3::detail::invoke_deferred_crtp_base<deferred_bases_t, derived<deferred_bases_t...>>...
+{};
+
+int main()
+{
+    // Wraps the actual classes into deferred crtp base classes.
+    // This step is necessary, since during their declaration the type of the derived class cannot be known.
+    using deferred_base1 = seqan3::detail::deferred_crtp_base<base1, std::string>;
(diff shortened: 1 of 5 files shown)


test/snippet/core/metafunction/deferred_crtp_base.cpp
```diff
@@ -0,0 +1,49 @@
...
+{
+public:
+
+    value_t func1() const
+    {
+        return {"instance of base1"};
+    }
+};
+
+// Defines a crtp_base class with an additional return type and a parameter_t.
```
smehringer:
```suggestion
// Defines a crtp_base class with an additional value type and a parameter_t.
```
?

test/snippet/core/metafunction/deferred_crtp_base.cpp
```diff
@@ -0,0 +1,49 @@
...
+{
+public:
+
+    value_t func2(parameter_t const p) const
+    {
+        return static_cast<value_t>(p);
+    }
+};
+
+// The derived class that inherits from a variadic crtp pattern.
```
smehringer: add
`// This declaration would not be possible without deferred crtp_base class instantiation as it would cause an incomplete type compiler error` (right?)
if I understood correct also add:
`// Note the derived class can now only be instantiated with deferred crtp base classes (created from normal crtp base classes by using seqan3::detail::deferred_crtp_base)`

test/snippet/core/metafunction/deferred_crtp_base.cpp
joergi-w: ... and a parameter type. (in a comment we don't need to write parameter_t)

test/snippet/core/metafunction/deferred_crtp_base.cpp
```diff
@@ -0,0 +1,49 @@
...
+template <typename ...bases_t>
```
marehr: `deferred_bases_t`

test/unit/core/metafunction/deferred_crtp_base_test.cpp
```diff
@@ -0,0 +1,110 @@
...
+// ARE DISCLAIMED. IN NO EVENT SHALL KNUT REINERT OR THE FU BERLIN BE LIABLE
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+// DAMAGE.
+//
+// ============================================================================
```
marehr: short headers

test/unit/core/metafunction/deferred_crtp_base_test.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+    value_t func2(parameter_t const p) const
+    {
+        return static_cast<value_t>(p);
+    }
+};
+
+// The derived class that inherits from a variadic crtp pattern.
+template <typename ...bases_t>
+class derived : public seqan3::detail::invoke_deferred_crtp_base<bases_t, derived<bases_t...>>... // Instantiate the crtp_base classes
```
marehr: isn't this longer than 120?

PR #603, opened 2018-12-04
[dry] test cases

marehr:
`seqan3::as_testing_list` allows to transform `seqan3::type_list` into `::testing::Types`. Which helps us to keep types in test cases dry and in synch if they are reused.

Example:
```c++
#include <seqan3/core/type_list.hpp>
#include <seqan3/test/testing_list.hpp>
using namespace seqan3;
template <typename T>
class my_typed_test : public ::testing::Test
{};
using types = type_list<int, char, double>;
TYPED_TEST_CASE(my_typed_test, as_testing_list<types>);
// same as TYPED_TEST_CASE(my_typed_test, ::testing::Types<int, char, double>);
```


test/include/seqan3/test/testing_list.hpp
```diff
@@ -0,0 +1,43 @@
...
+ * \tparam  type_list  The type list.
+ *
+ * \details
+ *
+ * You should use this if you would have to redefine every type in a second type list just to test the same types in
+ * gtest's `TYPED_TEST` and in one of your test cases. This reduces redundancy and errors when keeping the types
+ * up-to-date and in sync.
+ *
+ * \attention
+ * Google type lists can't easily be transformed by seqan3::detail::transfer_template_args_onto to a seqan3::type_list,
```
rrahn: Why would anyone want this?

test/unit/alphabet/aminoacid/aminoacid_conversion_integration_test.cpp
```diff
@@ -10,23 +10,23 @@
...
 template <typename T>
 class aminoacid_conversion : public ::testing::Test
 {};
 
-using aminoacid_types2 = meta::list<aa20, aa27>; // needed for some tests
-using aminoacid_types = detail::transfer_template_args_onto_t<aminoacid_types2, ::testing::Types>;
+using aminoacid_types = type_list<aa20, aa27>; // needed for some tests
 
-TYPED_TEST_CASE(aminoacid_conversion, aminoacid_types);
+TYPED_TEST_CASE(aminoacid_conversion, as_testing_list<aminoacid_types>);
```
rrahn: So `detail::transfer_template_args_onto_t<aminoacid_types2, ::testing::Types>` would also have worked right?

marehr: Yeah would have, I can revert to this. I thought it would be easier to read, but I can just use `transfer_template_args_onto_t`

test/include/seqan3/test/testing_list.hpp
marehr: To reuse the types.

smehringer:
What does it mean to "keep test cases dry"....? :sweat_smile:

eseiler:
> What does it mean to "keep test cases dry"....? 😅

https://stackoverflow.com/a/11837973

test/include/seqan3/test/testing_list.hpp
```diff
@@ -0,0 +1,43 @@
...
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
+// -----------------------------------------------------------------------------------------------------
+
+#pragma once
+
+#include <gtest/gtest.h>
+
+#include <seqan3/core/metafunction/template_inspection.hpp>
+
+namespace seqan3
```
smehringer: IMPOV this should be detail and no public interface? 

And maybe call this file utility or basic since it is just some convenience helper of which we might implement more and not each of them should go into a separate file.

marehr: My thought was, that it is public in the test environment, because it's only accessible in there.

smehringer: True, if you see it like that. But a whole file for this tiny helper is still megalomaniac :grin:  (yes I googled this word)

test/include/seqan3/test/testing_list.hpp
rrahn: But why would you transfer the types from the google interface to your own type? 
I think you can always go from your type to google type list and there is no need to convert the other way around. Or am I misreading the sentence?

test/unit/alphabet/aminoacid/aminoacid_conversion_integration_test.cpp
rrahn: It's just that for me it is just another name that is explicitly introduced in the testing environment. So I would prefer the  transfer_template_... thing as it is documented and we only need it at some few places.

marehr:
@rrahn Like I said, I will remove the alias and use it directly.

PR #607, opened 2018-12-05
[fix] cmake uses -isystem and not -I for seqan3 include path

### LINKED ISSUE #606: Fix warnings in unit tests, @marehr killed it.

marehr:
We observed the problem that warnings weren't triggered in our tests.

After some investigation, we figured out that seqan3 was included as `-isystem` in the unit tests.

After some more digging, I found out that this is due to using `add_library(name INTERFACE IMPORTED)`.

This PR changes the cmake script in such a way that seqan3 will always be included via `-I`, but our external dependencies will be still included as system libraries, i.e. `-isystem`.

This resolves #606.

-----

This PR only makes the changes in cmake, the following PRs fix the code:

* ~#616~ 
* ~#617~
* ~#618~
* ~#619~ 
* ~#627~
* ~#628~
* ~#629~

### DIFF
--- a/build_system/seqan3-config.cmake
+++ b/build_system/seqan3-config.cmake
@@ -170,9 +170,9 @@ find_path (SEQAN3_CLONE_DIR NAMES build_system/seqan3-config.cmake HINTS "${CMAK
 if (SEQAN3_CLONE_DIR)
     message (STATUS "  Detected as running from a repository checkout…")
 
-    if (IS_DIRECTORY "${SEQAN3_CLONE_DIR}/include")
+    if (NOT SEQAN3_BASEDIR AND IS_DIRECTORY "${SEQAN3_CLONE_DIR}/include")
         message (STATUS "  …adding SeqAn3 include:     ${SEQAN3_CLONE_DIR}/include")
-        set (SEQAN3_INCLUDE_DIRS "${SEQAN3_CLONE_DIR}/include" ${SEQAN3_INCLUDE_DIRS})
+        set (SEQAN3_BASEDIR "${SEQAN3_CLONE_DIR}/include")
     endif ()
 
     if (EXISTS "${SEQAN3_CLONE_DIR}/submodules")
@@ -187,13 +187,30 @@ if (SEQAN3_CLONE_DIR)
 endif ()
 
 # ----------------------------------------------------------------------------
+# Find SeqAn3 include path
+# ----------------------------------------------------------------------------
+
+if (NOT SEQAN3_BASEDIR)
+    find_path (SEQAN3_BASEDIR "seqan3/version.hpp" HINTS ${SEQAN3_INCLUDE_DIRS})
+endif ()
+
+mark_as_advanced (SEQAN3_BASEDIR)
+
+# find include directory
+if (SEQAN3_BASEDIR)
+    seqan3_config_print ("SeqAn3 include dir found:   ${SEQAN3_BASEDIR}")
+else ()
+    seqan3_config_error ("SeqAn3 include directory could not be found.")
+endif ()
+
+# ----------------------------------------------------------------------------
 # Options for CheckCXXSourceCompiles
 # ----------------------------------------------------------------------------
 
 # deactivate messages in check_*
 set (CMAKE_REQUIRED_QUIET       1)
 # use global variables in Check* calls
-set (CMAKE_REQUIRED_INCLUDES    ${CMAKE_INCLUDE_PATH} ${SEQAN3_INCLUDE_DIRS})
+set (CMAKE_REQUIRED_INCLUDES    ${CMAKE_INCLUDE_PATH} ${SEQAN3_BASEDIR} ${SEQAN3_INCLUDE_DIRS})
 set (CMAKE_REQUIRED_FLAGS       ${CMAKE_CXX_FLAGS})
 
 # ----------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


build_system/seqan3-config.cmake
```diff
@@ -615,13 +607,20 @@ if (NOT ${FIND_NAME}_FIND_QUIETLY)
...
-set_property (TARGET seqan3::seqan3 APPEND PROPERTY INTERFACE_LINK_LIBRARIES "${SEQAN3_LIBRARIES}")
-set_property (TARGET seqan3::seqan3 APPEND PROPERTY INTERFACE_INCLUDE_DIRECTORIES "${SEQAN3_INCLUDE_DIRS}")
+add_library (seqan3_seqan3 INTERFACE)
+target_compile_definitions (seqan3_seqan3 INTERFACE ${SEQAN3_DEFINITIONS})
+target_compile_options (seqan3_seqan3 INTERFACE ${SEQAN3_CXX_FLAGS_LIST})
+target_link_libraries (seqan3_seqan3 INTERFACE "${SEQAN3_LIBRARIES}")
+# include seqan3/include/ as -I, because seqan3 should never produce warnings.
+target_include_directories (seqan3_seqan3 INTERFACE "${SEQAN3_BASEDIR}")
+# include everything except seqan3/include/ as -isystem, i.e.
+# a system header which suppresses warnings of external libraries.
```
rrahn: so the `set_property` didn't work?

test/unit/alphabet/adaptation/char_test.cpp
```diff
@@ -72,15 +72,9 @@ TYPED_TEST(char_adaptation, to_rank)
...
-        unsigned char cmp{'A'};
-        EXPECT_EQ(to_rank(TypeParam{65}), cmp);
-    }
-    else
-    {
-        EXPECT_EQ(to_rank(TypeParam{65}), 'A');
-    }
+
+    unsigned char cmp{'A'};
+    EXPECT_EQ(to_rank(TypeParam{65}), cmp);
```
rrahn: why not to distinguish between signed and unsigned char anymore?

build_system/seqan3-config.cmake
marehr: It worked, but gave me the problem that cmake used `-isystem` instead of `-I`

test/unit/alphabet/adaptation/char_test.cpp
marehr: This is part of PR #617 which is already merged; I didn't rebase, yet.

I have to make it unsigned, because `to_rank`'s return type is `unsigned` in any case.

h-2:
is this still blocked by other PRs? Can you rebase it so we can merge it soon?

rrahn:
By #618 But my plan is to make the final revision for #618 today and then pull in #618 and rebase on my branch.

marehr:
@h-2, @rrahn There are more modules that need change. The current PRs were only a subset of the warnings.

h-2:
@rrahn @marehr I really would prefer if we can get this fixed in this sprint, because it causes a lot of problems down the line, it means lots of buggy code goes unnoticed or suddenly pops up in header tests. @marehr Maybe you can get @MitraDarja or @joshuak94 to help you with the fixes?

marehr:
@h-2 I was just rebuilding everything and tried to estimate the amount of work:

```
93% tests passed, 10 tests failed out of 140

Total Test time (real) =   0.30 sec

The following tests FAILED:
	 17 - unit/alignment/pairwise/alignment_selector_test (Not Run)
	 18 - unit/alignment/pairwise/edit_distance_unbanded_test (Not Run)
	 19 - unit/alignment/pairwise/execution/alignment_executor_two_way_test (Not Run)
	 97 - unit/io/stream/parse_condition_test (Not Run)
	 98 - unit/io/structure_file/format_vienna_test (Not Run)
	126 - unit/search/fm_index_test (Not Run)
	127 - unit/search/fm_index_iterator_test (Not Run)
	128 - unit/search/bi_fm_index_iterator_test (Not Run)
	131 - unit/search/search_scheme_algorithm_test (Not Run)
	132 - unit/search/search_test (Not Run)
```

It's not to much anymore. I think I can manage it until the of the sprint.

rrahn:
@marehr Oki if you feel that this is manageble then please do it. Otherwise, we finish it in the first week of next sprint (which should be next week :D)

eseiler:
I can offer to have a look at the search module

marehr:
@eseiler  Thank you, if I need help I come back to you!

marehr:
@h-2, @rrahn done (at least on gcc8 on my machine), if all PR's are merged this one should go through.

h-2:
@marehr Are we through with the other commits? Can you do another rebase?

marehr:
@rrahn, @h-2 rebased and let's see if we are warning free :trumpet:

marehr:
Okay still open things to fix:

```
/home/travis/build/seqan/seqan3/test/unit/alignment/scoring_scheme_test.cpp:76:46: error: ISO C++11 requires at least one argument for the "..." in a variadic macro [-Werror]
 TYPED_TEST_CASE(generic, scoring_scheme_types);
                                              ^
```
This is not a problem with our library, but with gtest, see  https://github.com/google/googletest/issues/1419.
I would propose to suppress gtest headers again.

build_system/seqan3-config.cmake
```diff
@@ -615,13 +607,20 @@ if (NOT ${FIND_NAME}_FIND_QUIETLY)
...
+target_include_directories (seqan3_seqan3 SYSTEM INTERFACE "${SEQAN3_INCLUDE_DIRS}")
+add_library (seqan3::seqan3 ALIAS seqan3_seqan3)
+
+# propagate SEQAN3_BASEDIR into SEQAN3_INCLUDE_DIRS
+set (SEQAN3_INCLUDE_DIRS ${SEQAN3_BASEDIR} ${SEQAN3_INCLUDE_DIRS})
```
marehr: I didn't rename, because I would have broken this. If you don't have any problem with it, I will change it.

h-2:
Thanks for fixing this! I am merging this now, but I still have some questions that we could maybe discuss tomorrow at the strategy meeting or later on.

PR #613, opened 2018-12-06
[FIX] Fixes transfer_template_args_onto when the transferred type would be ill-formed.

rrahn:
This fixes hard compiler errors that occur, when the transferred type is ill-formed due to constraints on the template arguments of the target type. This happens especially when trying to use `is_type_specialization_of` as it uses this meta-function internally.

### DIFF
--- a/include/seqan3/core/metafunction/template_inspection.hpp
+++ b/include/seqan3/core/metafunction/template_inspection.hpp
@@ -41,7 +41,9 @@
 
 #include <meta/meta.hpp>
 
-#include <seqan3/core/platform.hpp>
+#include <seqan3/core/metafunction/transformation_trait_or.hpp>
+
+#include <seqan3/std/concepts>
 
 namespace seqan3::detail
 {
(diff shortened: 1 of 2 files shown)


include/seqan3/core/metafunction/template_inspection.hpp
```diff
@@ -46,6 +46,29 @@
...
+/*!\brief Helper metafunction for seqan3::detail::transfer_template_args_onto to test whether transferring the source
+ *        types onto the target type does not generate an ill-formed type.
+ * \tparam The type to check whether it is not ill-formed.
+ *
+ * \details
+ *
+ * Returns `std::true_type` if a pointer to the type can be created, otherwise `std::false_type`.
+ */
+template <typename type>
+    requires requires (type * ptr){ ptr; }
```
marehr: Why suffices a pointer?

rrahn:
@h-2 do you want to look at it as well? Btw. Jenkins is currently not responding correctly, so we cannot wait for the builds to finish.

h-2:
The other `void` stuff in there was added also by you at some time, right? Or did I do that? :thinking: 

otherwise LGTM!

marehr:
But some tests are failing...

rrahn:
Ok I look at the failing tests. But I thought that the void cruft insight of it would be not necessary anymore, because if the type is ill-formed the type would not be generated anyway. But I will have a closer look giving, that also some stuff is broken

PR #615, opened 2018-12-07
Feature Dummy sequence

### LINKED ISSUE #586: Dummy Sequence that allows to be decorated with gaps but throws on access
This sequence is needed for the SAM format because of the following reasons:

The SAM format does not store the complete alignment as the reference sequence information is always missing. It only stores the query sequence and a cigar string. The cigar string holds the information about how the query is aligned to the reference.

The alignment_file_in class will be able to instantiate with additional reference sequence information but also wants to be able to read a SAM file **without** the reference information.

In order to be able to represent an alignment (currently a tuple of aligned sequences) without the reference sequence, we need a dummy sequence that is decorated with gaps. This way, I can use the alignment for e.g. statistics (how many gaps are in the reference? Is there a gap at position 3?) without the need for the actual sequence.

smehringer:
Resolves #586 

Some design decisions before you review the code:

1) The dummy_container mimics (and models) a random_access_container but to avoid access to it, all element access functions, and dereferencing of iterators **throw an exception**.

2) All other functions modifying functions are valid (insert, erase, push_back) although they merely change the size. It is also possible to iterate over the container (random access iterator), but you cannot dereference the iterator.

3) Even though no alphabet character is ever stored, the container is templetized over an alphabet, to be compatible with many requirements on containers of other functions (e.g. common reference type). Example1: The gap_dectorator will most likely need its container::value_type to mimic the container behaviour correctly. Example2: free functions, e.g. assign would probably require a common reference type even before they redirect to the member function.


h-2:
Wow, are you sure we need a full container for this task? Do we need insert and push_back etc? The adapters should work on any sized+random access range I think so the dummy could be just that, too, or not?

I thought of something like this originally, when we discussed this:
```cpp
#include <range/v3/view/repeat_n.hpp>
#include <range/v3/view/transform.hpp>

#include <iostream>
#include <cassert>

int main()
{
    auto v = ranges::view::repeat_n(int{1} /*anyvalue of alphabet*/, 7/*size*/) | ranges::view::transform([] (auto alph) { throw 42; return alph; });

    assert(ranges::size(v) == 7);
    static_assert(std::is_same_v<ranges::range_value_type_t<decltype(v)>, int>);

    std::cout << v[3]; // will throw
}
```
Wouldn't that work for our use-cases?

include/seqan3/range/container/dummy_container.hpp
```diff
@@ -0,0 +1,852 @@
...
+// ARE DISCLAIMED. IN NO EVENT SHALL KNUT REINERT OR THE FU BERLIN BE LIABLE
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+// DAMAGE.
+//
+// ============================================================================
```
rrahn: Short header

include/seqan3/range/container/dummy_container.hpp
```diff
@@ -0,0 +1,852 @@
...
+
+/*!\file
+ * \author Svenja Mehringer <svenja.mehringer AT fu-berlin.de>
+ * \brief Contains the seqan3::dummy_container that throws on access but otherwise
```
rrahn:
```suggestion
 * \brief Provides the seqan3::dummy_container that throws on access but otherwise
```

include/seqan3/range/container/dummy_container.hpp
```diff
@@ -0,0 +1,852 @@
...
+
+namespace seqan3
+{
+
+/*!\brief A dummy container that behaves like a standard sequences container except that it throws on access.
+ * \tparam alphabet_type The value type of the container, must satisfy seqan3::alphabet_concept.
+ * \implements seqan3::random_access_container_concept
+ * \ingroup container
+ *
+ * This sequence is primarily used for storing a semi-alignment. We define a semi-alignment to an alignment of known
```
rrahn: Please don’t use semi-alignment as it might confuse with semi-global alignment. Call it an incomplete alignment maybe?

include/seqan3/range/container/dummy_container.hpp
```diff
@@ -0,0 +1,852 @@
...
+ * structure, e.g. you know it's length and the number and position of all gaps, but you do not necessarily know the
+ * sequence information. The use case of a semi-alignment lies in the SAM format where alignments are stored without
+ * the reference sequence information. The dummy_container let's us read in the SAM format and still reconstruct the
```
rrahn:
```suggestion
 * the reference sequence information. The dummy_container  can be used to read in the SAM format and still reconstruct the
```

PR #617, opened 2018-12-10
[fix] warnings in alphabet tests

marehr:
Part of #607, fixes mainly `signed` != `unsigned` warnings.


test/unit/alphabet/adaptation/char_test.cpp
```diff
@@ -72,15 +72,9 @@ TYPED_TEST(char_adaptation, to_rank)
...
     EXPECT_TRUE((std::is_same_v<decltype(to_rank(TypeParam{'A'})), underlying_rank_t<TypeParam>>));
-    if constexpr (std::is_unsigned_v<TypeParam>)
-    {
-        unsigned char cmp{'A'};
-        EXPECT_EQ(to_rank(TypeParam{65}), cmp);
-    }
-    else
-    {
-        EXPECT_EQ(to_rank(TypeParam{65}), 'A');
-    }
```
h-2: it might be better to keep the if-check and do a `signed char cmp{'A'};` in the second case. I suspect this would be better for platforms where char is unsigned. Would actually emulate this in a nightly build with `-funsigned-char`, but I don't know if it's worth ith...

PR #621, opened 2018-12-10
Renaming

MitraDarja:
#520

All renamings which were necessary for file alphabet/pre_concept.hpp

Renames:
semi_constexpr_alphabet_concept -> SemiConstexprAlphabet
constexpr_semi_alphabet_concept -> ConstexprSemialphabet
constexpr_alphabet_concept -> ConstexprAlphabet
semi_alphabet_concept -> Semialphabet
alphabet_concept -> Alphabet
rna_structure_concept -> RnaStructureAlphabet
char_adaption_concept -> CharAdaptationAlphabet
uint_adaption_concept -> UintAdaptionAlphabet

### DIFF
--- a/include/seqan3/alphabet/composition/cartesian_composition.hpp
+++ b/include/seqan3/alphabet/composition/cartesian_composition.hpp
@@ -155,7 +155,7 @@ decltype(auto) get();
  * Short description:
  *   * combines multiple alphabets as independent component_list, similar to a tuple;
  *   * models seqan3::tuple_like_concept, i.e. provides a get interface to its component_list;
- *   * is itself a seqan3::Semialphabet, but most derived types implement the full seqan3::alphabet_concept;
+ *   * is itself a seqan3::Semialphabet, but most derived types implement the full seqan3::Alphabet;
  *   * its alphabet size is the product of the individual sizes;
  *   * constructible, assignable and comparable with each component type and also all types that
  *     these are constructible/assignable/comparable with;
--- a/include/seqan3/alphabet/composition/cartesian_composition.hpp
+++ b/include/seqan3/alphabet/composition/cartesian_composition.hpp
@@ -144,10 +144,10 @@ decltype(auto) get();
 
 /*!\brief The CRTP base for a combined alphabet that contains multiple values of different alphabets at the same time.
  * \ingroup composition
- * \implements seqan3::semi_alphabet_concept
+ * \implements seqan3::Semialphabet
  * \implements seqan3::detail::ConstexprSemialphabet
- * \tparam first_component_type Type of the first letter; must model seqan3::semi_alphabet_concept.
- * \tparam component_types      Types of further letters (up to 4); must model seqan3::semi_alphabet_concept.
+ * \tparam first_component_type Type of the first letter; must model seqan3::Semialphabet.
+ * \tparam component_types      Types of further letters (up to 4); must model seqan3::Semialphabet.
  *
  * This data structure is CRTP base class for combined alphabets, where the different
  * alphabet letters exist independently as component_list, similar to a tuple.
@@ -155,7 +155,7 @@ decltype(auto) get();
  * Short description:
  *   * combines multiple alphabets as independent component_list, similar to a tuple;
  *   * models seqan3::tuple_like_concept, i.e. provides a get interface to its component_list;
- *   * is itself a seqan3::semi_alphabet_concept, but most derived types implement the full seqan3::alphabet_concept;
+ *   * is itself a seqan3::Semialphabet, but most derived types implement the full seqan3::alphabet_concept;
  *   * its alphabet size is the product of the individual sizes;
  *   * constructible, assignable and comparable with each component type and also all types that
  *     these are constructible/assignable/comparable with;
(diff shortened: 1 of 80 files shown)


h-2:
:+1:

But like I said in #520, I would suggest that all the alphabet concept actually end in "Alphabet" so `rna_structure_concept` should become `RnaStructureAlphabet` IMHO. Can you change this for all of them?

h-2:
> > Can you change this for all of them?

the adaptations were not actually renamed, yet, right?

Also I think we should maybe switch to `Semialphabet` instead of `SemiAlphabet` (same for Constexpr*), because the standard also has `Semiregular` and not `SemiRegular`...

marehr:
> Also I think we should maybe switch to `Semialphabet` instead of `SemiAlphabet` (same for Constexpr*), because the standard also has `Semiregular` and not `SemiRegular`...

Please let's discuss this in the next strategy meeting (@smehringer can you add this?)

h-2:
@marehr If you feel this needs discussing we can just leave it like this. It's not important to me, other language use just suggest that in English "semi-foobar" is more like "semifoobar" than "semi foobar", thus suggesting CamelCase spelling of "Semifoobar" rather than "SemiFoobar".
But like I said, I don't think its important enough for four people to argue about if there is no immediate consensus on this.

MitraDarja:
> the adaptations were not actually renamed, yet, right?

No, I only updated the concepts here I mentioned in the first post. I will update more concepts next week. Maybe you can take a look at the renaming list in this pull request (https://github.com/seqan/seqan3/pull/621) because if not stated otherwise I will rename everything as I proposed there.

h-2:
> No, I only updated the concepts here I mentioned in the first post. 

But you did mention them in the first post, that's why I was confused...

MitraDarja:
> But you did mention them in the first post, that's why I was confused

Sorry, yeah updated those two adaptions but that were only one change each if I remember correctly.

include/seqan3/alphabet/composition/cartesian_composition.hpp
```diff
@@ -171,18 +171,18 @@ decltype(auto) get();
 
 /*!\brief The CRTP base for a combined alphabet that contains multiple values of different alphabets at the same time.
  * \ingroup composition
- * \implements seqan3::semi_alphabet_concept
- * \implements seqan3::detail::constexpr_semi_alphabet_concept
- * \tparam first_component_type Type of the first letter; must model seqan3::semi_alphabet_concept.
- * \tparam component_types      Types of further letters (up to 4); must model seqan3::semi_alphabet_concept.
+ * \implements seqan3::SemiAlphabet
+ * \implements seqan3::detail::ConstexprSemiAlphabet
+ * \tparam first_component_type Type of the first letter; must model seqan3::SemiAlphabet.
```
h-2: We re-discussed this and we decided on renaming `semi_alphabet_concept` to `Semialphabet` not `SemiAlphabet` (and for Constexpr, also).

h-2: (English language peculiarities)

include/seqan3/alphabet/concept_pre.hpp
```diff
@@ -242,7 +242,7 @@ constexpr uint8_t max_pseudoknot_depth_v = max_pseudoknot_depth<alphabet_type>::
 } // namespace seqan3
 
 // ------------------------------------------------------------------
-// seqan3::char_adaption_concept
+// seqan3::CharAdaption
```
h-2: this is also an alphabet so it should be CharAdaptationAlphabet

include/seqan3/alphabet/concept_pre.hpp
```diff
@@ -264,7 +264,7 @@ template <typename type>
 constexpr bool is_char_adaptation_v = is_char_adaptation<type>::value;
 
 // ------------------------------------------------------------------
-// seqan3::uint_adaption_concept
+// seqan3::UintAdaption
```
h-2: here as well

include/seqan3/range/view/to_rank.hpp
```diff
@@ -77,7 +77,7 @@ namespace seqan3::view
  * | std::ranges::OutputRange        |                                       | *lost*                                             |
  * | seqan3::const_iterable_concept  |                                       | *preserved*                                        |
  * |                                 |                                       |                                                    |
- * | seqan3::reference_t             | seqan3::alphabet_concept              | seqan3::underlying_rank_t<seqan3::value_type_t<urng_t>> |
+ * | seqan3::reference_t             | seqan3::Alphabet              | seqan3::underlying_rank_t<seqan3::value_type_t<urng_t>> |
```
h-2: please fix the alignment in the tables after the rename

include/seqan3/alphabet/structure/RnaStructureAlphabet.hpp
```diff
@@ -34,7 +34,7 @@
```
h-2: don't rename this file to RnaStructureAlphabet.hpp, files should always stay snake_case. Maybe just rename it to concept.hpp like in the other alphabet folders?

h-2:
has conflicts

Issue #608, opened 2018-12-05
alignment


rrahn:
Currently the alignment result object is implemented as a tuple with enum based access. To reduce the complexity of the interface and to allow type-erasure on the return type of alignment algorithms the result object should be refactored.

The file is here: `include/seqan3/alignment/pairwise/align_result.hpp`

Tasks:
 - [x] Remove tuple interface of the alignment result type
 - [x] the class can be queried to return the score*, the end position, the begin position, the traceback.
 - [x] the score must always be set
 - [x] The other values (except the score) are optionals
 - [x] On access of an non-set optional throw an invalid access error.
 - [x] Only return unmutables or copies of the respective value.
 - [x] The necessary types should be passed as a traits type  
 - [x] implement a concept for the traits class to check for valid configuration
 - [x] write tests
 - [x] write docu

###### estimate: 13

smehringer:
I thought we decided not to make meta issues but meta cards with single task issues?

joergi-w:
Implemented with PR #625.

PR #625, opened 2018-12-12
new Alignment Result class

joergi-w:
This PR re-implements the alignment result class. The tuple interface is removed and new get functions for all members are introduced. Tests and full documentation are included. 

Solves #608


test/unit/alignment/pairwise/alignment_result_test.cpp
```diff
@@ -32,17 +32,21 @@
 //
 // ==========================================================================
 
+// This file provides the unit tests for seqan3::align_result.
```
rrahn: please remove these lines from test files.

include/seqan3/alignment/pairwise/align_result.hpp
```diff
@@ -34,210 +34,154 @@
...
+
+    /*!\brief Constructor with identifier and score only.
+     * \param id_               The alignment identifier.
+     * \param score_            The alignment score.
+     */
+    align_result(id_t const id_,
+                 score_t const score_)
+        : id{id_}, score{score_}, end_coordinate{}, begin_coordinate{}, trace{}
+    {}
+
```
rrahn: Please add a default constructor. 
In this case the access would always throw and id would be -1. -> type deduced.

include/seqan3/alignment/pairwise/align_result.hpp
```diff
@@ -34,210 +34,154 @@
...
+    score_t score;
+    //! \brief The end coordinate of the alignment.
+    std::optional<coordinate_t> end_coordinate;
+    //! \brief The begin coordinate of the alignment.
+    std::optional<coordinate_t> begin_coordinate;
+    //! \brief The alignment trace, i.e. the actual base pair matching.
+    std::optional<trace_t> trace;
+
+public:
+    /*!\brief Constructor with all available fields.
```
rrahn: From the design I suggest a different approach.
Provide a `detail::alignment_result_value_type` which is an aggregate type (no constructors).
It has deduction guides for only the first, then the first and the second etc.

This is what I mean, but with more descriptive names.
The non set types will default to `void *`
```cpp
template <typename t1, typename t2 = void *>
struct alignment_result_value_type
{
    t1 v1{};
    t2 v2{};
};

template <typename t1>
alignment_result_value_type(t1) -> alignment_result_value_type<t1>;

template <typename t1, typename t2>
alignment_result_value_type(t1, t2) -> alignment_result_value_type<t1, t2>;

int main()
{
    alignment_result_value_type o{4};

    alignment_result_value_type o2{4, .5};
}
```

The alignment result class is templatized only over this struct, which we can constrain then with `is_specialization_of` as well. And there is only one constructor. The access functions could then static assert if the requested type is a `void *` instead of throwing. I realized yesterday while working on the algorithm implementation, that the type of the result must be and can determined anyway. 
The default constructor could deduce to a value type that only has void *.

include/seqan3/alignment/pairwise/align_result.hpp
```diff
@@ -34,210 +34,154 @@
 
 /*!\file
  * \brief Provides seqan3::align_result.
- * \author Rene Rahn <rene.rahn AT fu-berlin.de>
```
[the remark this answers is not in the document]
joergi-w: Oh, I didn't remove you on purpose. I created a new file from scratch and forgot to include you again – sorry!! 
I see it looks weird, because it shows the diff of the old and new file (with only a few brackets matching).

include/seqan3/alignment/pairwise/align_result.hpp
smehringer: Do you want to have the result members public or as private members with getters?

smehringer: Hm isn't this also too complicated? What about this:
```cpp
#include <optional>

template<typename t1, typename t2 = void *>
struct S
{

    S(t1 i2 = t1{}, t2 f2 = t2{}): i{i2}, f{f2}
    {} 

    t1 get_i()
    {
        static_assert(!std::is_same_v<t1, void *>);
        return i;
    }

    t2 get_f()
    {
        static_assert(!std::is_same_v<t2, void *>);
        return f;
    }

private:
    t1 i{};
    t2 f{};
};

int main()
{
    S mys{1, 1.0}; // for S mys{1}; the static assertion fails

    mys.get_f();
}

```
No multiple constructors, no multiple type deductions. no extra detail struct, static asserts for getter functions.

rrahn: All fine but, in your example all types are exposed in the user interface, rather then just one type.
Second, it forbids me to use `designated initializers` in the future. So within the alignment computation I would love to just do: 
```cpp
alignment_result{result_value_type{.id = 0, .score = sc, .end_coordinage = {30, 12}}}
``` 
Third, I can pass in the arguments in a wrong order, which I could prevent with strong types. In the above solution I can still mix the order, but at least I have a concrete naming for the variables and don't need to introduce all the strong types. 
Fourth, the implementation of the values is decoupled from the access object, which means I can reassign the values in the alignment computation. The user should not be able to do this.

smehringer: I see, I might still ask you out of interest later on as I don't understand all your points.
I just wanted to make a suggestion since the issue seemed easy enough at a first glance but as it looks like there is more to it I'll leave this design decision to you.

joergi-w: I implemented the proposed design, so you can have a look and compare.

include/seqan3/alignment/pairwise/align_result.hpp
```diff
@@ -34,210 +34,157 @@
...
+struct align_result_value_type
 {
+    //! \brief The alignment identifier.
+    id_t id{};
+    //! \brief The alignment score.
+    score_t score{};
+    //! \brief The end coordinate of the alignment.
+    coordinate_t end_coordinate{};
+    //! \brief The begin coordinate of the alignment.
+    coordinate_t begin_coordinate{};
```
rrahn: should be a different type name, as it would be possible to compute the end point without the begin point. But if they have the same type, I can not check with static_assert anymore, if it was computed or not.

include/seqan3/alignment/pairwise/align_result.hpp
```diff
@@ -34,210 +34,157 @@
...
-    using base_type = detail::transfer_template_args_onto_t<output_type_list_t, std::tuple>;
+//! \brief Type deduction for id, score and end coordinate.
+template <typename id_t, typename score_t, typename coordinate_t>
+align_result_value_type(id_t, score_t, coordinate_t) -> align_result_value_type<id_t, score_t, coordinate_t>;
 
-    //!\brief Inheriting the constructors of the base type.
-    using base_type::base_type;
+//! \brief Type deduction without the trace.
+template <typename id_t, typename score_t, typename coordinate_t>
+align_result_value_type(id_t, score_t, coordinate_t, coordinate_t) -> align_result_value_type<id_t, score_t,
```
rrahn: maybe wrap into next line?

include/seqan3/alignment/pairwise/align_result.hpp
```diff
@@ -34,210 +34,157 @@
...
+                                                                                              coordinate_t>;
 
-    /*!\name Access functions
-     * \brief Convenience helper functions to access elements of the alignment result type.
-     * \{
-     */
+//! \brief Type deduction with all available fields.
+template <typename id_t, typename score_t, typename coordinate_t, typename trace_t>
+align_result_value_type(id_t, score_t, coordinate_t, coordinate_t, trace_t) -> align_result_value_type<id_t,
```
rrahn: same here?

include/seqan3/alignment/pairwise/align_result.hpp
```diff
@@ -34,210 +34,157 @@
...
-    constexpr std::tuple_element_t<1, base_type> score() const noexcept
-    {
-        return std::get<1>(*this);
-    }
+/*!\brief Stores the alignment results and gives access to score, traceback and the begin and end coordinates.
+ * \ingroup pairwise
+ * \tparam align_result_traits The type of the traits object.
+ *
+ * \details
+ * Objects of this class are the result of an alignment computation.
```
rrahn: empty line

include/seqan3/alignment/pairwise/align_result.hpp
```diff
@@ -34,210 +34,157 @@
...
+ * It always contains and alignment identifier and the resulting score.
+ * Optionally – if the user requests – also the begin and end positions within
+ * the sequences and the trace can be calculated. When accessing a field that
+ * has not been calculated an std::bad_optional_access exception is thrown.
```
rrahn: we use now static_assert right?

include/seqan3/alignment/pairwise/align_result.hpp
```diff
@@ -34,210 +34,157 @@
...
+ */
+template <typename align_result_traits>
```
rrahn: constrain the template to be a speciailisation of the value type above

include/seqan3/alignment/pairwise/align_result.hpp
```diff
@@ -34,210 +34,157 @@
...
+class align_result
+{
+private:
+    //! \brief Traits object that contains the actual alignment result data.
+    align_result_traits data;
+
+    using id_t = decltype(data.id);
```
rrahn: inline documentation missing

rrahn: surround them also with a `/*!\name Member types ...` group.

include/seqan3/alignment/pairwise/align_result.hpp
```diff
@@ -34,210 +34,157 @@
...
+    using score_t = decltype(data.score);
+    using coord_t = decltype(data.end_coordinate);
+    using trace_t = decltype(data.trace);
+
+public:
+    /*!\brief Constructor to pass the alignment result traits.
+     * \param[in] traits The alignment results.
+     */
+    align_result(align_result_traits traits) : data(traits) {};
```
rrahn: value instead of traits? traits are used to augment the type.

include/seqan3/alignment/pairwise/align_result.hpp
```diff
@@ -34,210 +34,157 @@
...
+     * \details Guaranteed to be smaller than or equal to `get_end_coordinate()`.
      */
-    constexpr decltype(auto) trace() const & noexcept
+    constexpr coord_t get_begin_coordinate() const noexcept
     {
-        if constexpr (std::tuple_size_v<base_type> < 5)
-            return std::ignore;
-        else
-            return std::get<4>(*this);
+        // TODO Problem: Assertion does not fail, if end_coordinate exists, because the coordinate type is set.
```
rrahn: see my advice above

test/unit/alignment/pairwise/alignment_result_test.cpp
```diff
@@ -53,22 +54,24 @@ class align_result_test : public ::testing::Test
...
 public:
 
-    using res_t   = align_result<T>;
-    using id_t    = meta::at_c<T, 0>;
-    using score_t = meta::at_c<T, 1>;
-    using end_t   = meta::at_c<T, 2>;
-    using begin_t = meta::at_c<T, 3>;
-    using trace_t = meta::at_c<T, 4>;
-    using seq_t   = std::tuple_element_t<0, trace_t>;
+    using id_t     = meta::at_c<T, 0>;
```
rrahn: We want to drop the dependency to meta. I suggest you do:
```suggestion
    using id_t     = decltype(std::declval<T>().id);
```

include/seqan3/alignment/pairwise/align_result.hpp
```diff
@@ -34,210 +34,157 @@
...
-#include <seqan3/core/metafunction/template_inspection.hpp>
-
 namespace seqan3
 {
 
-/*!\brief Keys for different alignment results.
- * \ingroup pairwise
- */
-enum struct align_result_key : uint8_t
+namespace detail
```
h-2: don't define namespaces inside each other, use `namespace seqan3::detail {`

include/seqan3/alignment/pairwise/align_result.hpp
```diff
@@ -34,210 +34,157 @@
...
+    id_t id{};
+    //! \brief The alignment score.
+    score_t score{};
+    //! \brief The end coordinate of the alignment.
+    coordinate_t end_coordinate{};
+    //! \brief The begin coordinate of the alignment.
+    coordinate_t begin_coordinate{};
+    //! \brief The alignment trace, i.e. the actual base pair matching.
+    trace_t trace{};
+};
```
h-2: I think we never really had a design discussion on what we want this to look like... e.g. what type is the trace? don't we want actuall alignment (aligned sequences)? What about alignment statistics...

rrahn: So far the current interface represents the most basic things that the alignments should support and it could be augmented with additional information if we desire, without breaking the API (computing E-values etc.) The latest discussion on trace topic was to represent the traceback as a tuple of two gapped sequences. For the trace it would be good if we could formulate a concept which it models, such that either a tuple or a container or a graph or whatever of gapped sequences can be represented. Anyhow, the trace type will be determined by the algorithm and I'd like to omit the SeqAn2 style of allowing 5 different alignment representations for any algorithm. The algorithm should decide on the best output format. I think we can then always add transformation views that transform the range into another format if the user desires. We can discuss thisat the next strategy meeting.

include/seqan3/alignment/pairwise/align_result.hpp
joergi-w: okay!

test/unit/alignment/pairwise/alignment_result_test.cpp
joergi-w: Then `T` has to be of type `detail::align_result_value_type` and not `type_list` anymore, right? I'll adapt the implementation of this.

include/seqan3/alignment/pairwise/align_result.hpp
joergi-w: okay!

joergi-w:
I rebased on upstream/master, because of merge conflicts.

include/seqan3/alignment/pairwise/align_result.hpp
```diff
@@ -34,210 +34,177 @@
...
+ * \tparam end_coord_t   The type for the end coordinate, can be omitted.
+ * \tparam begin_coord_t The type for the begin coordinate, can be omitted.
+ * \tparam trace_t       The type for the alignment trace, can be omitted.
  */
-enum struct align_result_key : uint8_t
+template <typename id_t,
+          typename score_t,
+          typename end_coord_t = void *,
+          typename begin_coord_t = void *,
+          typename trace_t = void *>
```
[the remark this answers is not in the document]
joergi-w: It only works with `std::nullopt_t *` – is that fine as well?

include/seqan3/alignment/pairwise/align_result.hpp
```diff
@@ -34,210 +34,177 @@
...
+struct align_result_value_type
 {
-    id,     // cannot be set by the config
-    score,  // report the score
-    end,    // report the end position and score
-    begin,  // report the begin and end position and score
-    trace   // report the full trace and begin and end and score
+    //! \brief The alignment identifier.
+    id_t id{};
```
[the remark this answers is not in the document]
joergi-w: id is usually unsigned, so I think it's not a good idea...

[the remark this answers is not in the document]
joergi-w: Can 0 be the uninitialized id, and we start counting from 1?

test/unit/alignment/pairwise/alignment_result_test.cpp
```diff
@@ -34,287 +34,241 @@
...
-public:
-
-    using res_t   = align_result<T>;
-    using id_t    = meta::at_c<T, 0>;
-    using score_t = meta::at_c<T, 1>;
-    using end_t   = meta::at_c<T, 2>;
-    using begin_t = meta::at_c<T, 3>;
-    using trace_t = meta::at_c<T, 4>;
-    using seq_t   = std::tuple_element_t<0, trace_t>;
+    using traits_t = T;
```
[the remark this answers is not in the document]
joergi-w: Thanks!

rrahn:
@joergi-w the tests are failing. Can you have a look?

include/seqan3/alignment/pairwise/align_result.hpp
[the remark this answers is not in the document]
rrahn: I thought about it and even if I am favor of initializing it with -1 or `numeric_limits::max()`, I came to the conclusion, that there might be inputs where the Id is a tuple and hence the -1 initialization wouldn't work. So we leave it default initialized.

test/unit/alignment/pairwise/align_result_test.cpp
```diff
@@ -0,0 +1,274 @@
...
+// ARE DISCLAIMED. IN NO EVENT SHALL KNUT REINERT OR THE FU BERLIN BE LIABLE
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+// DAMAGE.
+//
+// ==========================================================================
```
rrahn: short header

include/seqan3/alignment/pairwise/align_result.hpp
```diff
@@ -34,210 +34,201 @@
...
 {
 
-/*!\brief Keys for different alignment results.
- * \ingroup pairwise
+/*!\brief A struct that contains the actual alignment result data.
+ * \tparam id_t          The type for the alignment identifier.
+ * \tparam score_t       The type for the resulting score.
+ * \tparam end_coord_t   The type for the end coordinate, can be omitted.
+ * \tparam begin_coord_t The type for the begin coordinate, can be omitted.
+ * \tparam trace_t       The type for the alignment trace, can be omitted.
```
rrahn: Can you please rename trace_t and all of its variants to alignment_t. Since we call the result an alignment.

include/seqan3/alignment/pairwise/align_result.hpp
```diff
@@ -34,210 +34,201 @@
...
     }
 
-    //!\overload
-    constexpr decltype(auto) begin_coordinate() const && noexcept
+    /*!\brief Returns the end coordinate of the alignment.
+     * \return A pair of positions in the respective sequences, where the calculated alignment ends.
+     * \attention This function with fail the compilation, if the end coordinate was not requested in the alignment
+     * configuration.
+     */
+    constexpr end_coord_t get_end_coordinate() const noexcept
```
rrahn:
```suggestion
    constexpr end_coord_t const & get_end_coordinate() const noexcept
```

include/seqan3/alignment/pairwise/align_result.hpp
```diff
@@ -34,210 +34,201 @@
...
-    /*!\brief Returns the traceback of the alignment if requested by the algorithm configuration,
-     *        otherwise `std::ignore`.
+    /*!\brief Returns the begin coordinate of the alignment.
+     * \return  A pair of positions in the respective sequences, where the calculated alignment starts.
+     * \details Guaranteed to be smaller than or equal to `get_end_coordinate()`.
+     * \attention This function with fail the compilation, if the begin coordinate was not requested in the alignment
+     * configuration.
      */
-    constexpr decltype(auto) trace() const & noexcept
+    constexpr begin_coord_t get_begin_coordinate() const noexcept
```
rrahn:
```suggestion
    constexpr begin_coord_t const & get_begin_coordinate() const noexcept
```

include/seqan3/alignment/pairwise/align_result.hpp
```diff
@@ -34,210 +34,201 @@
...
     }
 
-    //!\overload
-    constexpr decltype(auto) trace() const && noexcept
+    /*!\brief Returns the traceback of the alignment.
+     * \return At least two gapped sequences, which represent the alignment.
+     * \attention This function with fail the compilation, if the trace was not requested in the alignment
+     * configuration.
+     */
+    constexpr trace_t get_trace() const noexcept
```
rrahn:
```suggestion
    constexpr alignment_t const & get_alignment() const noexcept
```

joergi-w:
I rebased on master and solved the conflicts. The requested changes are included.

include/seqan3/alignment/pairwise/align_result.hpp
```diff
@@ -1,243 +1,206 @@
...
+    //! \brief The type for the end coordinate.
+    using end_coord_t   = decltype(data.end_coordinate);
+    //! \brief The type for the begin coordinate.
+    using begin_coord_t = decltype(data.begin_coordinate);
+    //! \brief The type for the alignment.
+    using alignment_t   = decltype(data.alignment);
+    //!\}
+
+public:
+    /*!\brief Constructor to pass the alignment result traits.
```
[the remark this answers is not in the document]
joergi-w: okay, done!

rrahn:
@joergi-w There are again some merge conflicts. Can you please look at this?

joergi-w:
rebased on master

PR #628, opened 2018-12-13
[fix] warnings in search tests

marehr:


include/seqan3/search/algorithm/configuration/max_error.hpp
```diff
@@ -122,10 +122,13 @@ struct search_config_max_error_adaptor : public configuration_fn_base<search_con
         // at least one specific error type specified
         else if constexpr (!total_set)
         {
+            uint8_t substitution_error = static_cast<uint8_t>(std::get<1>(out));
```
cpockrandt: Can you make all three `const`?

include/seqan3/search/algorithm/detail/search_scheme_algorithm.hpp
```diff
@@ -115,7 +115,7 @@ inline auto search_scheme_block_info(search_scheme_t const & search_scheme, size
 
     uint8_t const blocks      {search_scheme[0].blocks()};
     size_t  const block_length{query_length / blocks};
-    uint8_t const rest        {query_length - blocks * block_length};
+    uint8_t const rest        {static_cast<uint8_t>(query_length % blocks)};
```
cpockrandt: Are you sure we want to do this? `mod` has the complexity of `div` which should be much slower than the multiplication and substraction (since we already did the `div` in the line above)

include/seqan3/search/algorithm/detail/search_scheme_algorithm.hpp
```diff
@@ -404,7 +404,7 @@ inline bool search_ss(iterator_t it, query_t & query,
         return true;
     }
     // Exact search in current block.
-    else if ((max_error_left_in_block == 0) && (rb - lb - 1 != blocks_length[block_id]) ||
+    else if ((max_error_left_in_block == 0 && (rb - lb - 1 != blocks_length[block_id])) ||
```
cpockrandt: Good idea to put brackets around the `&&` term but I would additionally keep the brackets around the first term, i.e.

```suggestion
    else if (((max_error_left_in_block == 0) && (rb - lb - 1 != blocks_length[block_id])) ||
```

include/seqan3/search/algorithm/configuration/max_error.hpp
cpockrandt:
```suggestion
            uint8_t const substitution_error = static_cast<uint8_t>(std::get<1>(out));
            uint8_t const insertion_error = static_cast<uint8_t>(std::get<2>(out));
            uint8_t const deletion_error = static_cast<uint8_t>(std::get<3>(out));
```

(not sure whether multi-line suggestions work)

include/seqan3/search/algorithm/detail/search_scheme_algorithm.hpp
marehr: A division always delivers the rest on intel cpu's for free. See https://godbolt.org/z/KqID2a.

So you make it only more difficult for the compiler to see this.

include/seqan3/search/algorithm/configuration/max_error.hpp
marehr: will change! :+1:

include/seqan3/search/algorithm/detail/search_scheme_algorithm.hpp
marehr: It was not a good idea from me, it was a compiler warning :laughing: . I don't see the increase in readability to put extra brackets, but will do!

include/seqan3/search/algorithm/configuration/max_error.hpp
```diff
@@ -122,10 +122,13 @@ struct search_config_max_error_adaptor : public configuration_fn_base<search_con
...
+            uint8_t insertion_error = static_cast<uint8_t>(std::get<2>(out));
```
marehr:
```suggestion
            uint8_t const insertion_error = static_cast<uint8_t>(std::get<2>(out));
```

include/seqan3/search/algorithm/configuration/max_error.hpp
```diff
@@ -122,10 +122,13 @@ struct search_config_max_error_adaptor : public configuration_fn_base<search_con
...
+            uint8_t deletion_error = static_cast<uint8_t>(std::get<3>(out));
```
marehr:
```suggestion
            uint8_t const deletion_error = static_cast<uint8_t>(std::get<3>(out));
```

PR #631, opened 2018-12-13
[feature] add seqan3::detail::simd_transform

marehr:
This introduces `seqan3::detail::simd_transform`, which will help in testing and auto-vectorizing simd instructions.

```c++
using simd_type = seqan3::simd::simd_type_t<uint16_t, 8>;
using scalar_t = typename seqan3::simd::simd_traits<simd_type>::scalar_type; // same as uint16_t

auto iota = [i = scalar_t{0}] (size_t) mutable
{
    return i++;
};

simd_type result = seqan3::detail::simd_transform<simd_type>(iota);

// same as
result = simd_type
{
    iota(0),
    iota(1),
    iota(2),
    // ...
    iota(7)
};
```

### DIFF
--- a/include/seqan3/core/simd/simd_algorithm.hpp
+++ b/include/seqan3/core/simd/simd_algorithm.hpp
@@ -47,20 +47,90 @@
 namespace seqan3::detail
 {
 
-//!\brief Helper function for seqan3::simd::fill.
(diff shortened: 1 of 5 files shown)


marehr:
I forgot that I wanted to do this differently; I'll reopen when done.

PR #640, opened 2018-12-18
[FIX] minor fixes to view::persist

h-2:
Since we are always delegating to the underlying range's non-const iterator/sentinel, we don't need to check for const-iterability.


h-2:
> no changes to the test cases?

No, we were not testing on non-const-iterable ranges to begin with (and now it doesn't make a difference).

PR #646, opened 2018-12-19
[MISC] Update gap enum to gap char literal.

joshuak94:


include/seqan3/alphabet/gap/gap.hpp
```diff
@@ -149,6 +149,18 @@ struct gap
     //!\}
 };
 
-constexpr gap gap::GAP{};
+/*!\name Literals
+ * \{
+ */
+/*!\brief The seqan3::gap char literal.
+ * \param[in] c doesn't really matter, as gap only has one character.
```
smehringer: Does this mean you can assign any character? :thinking:

test/snippet/alphabet/composition/union_composition.cpp
```diff
@@ -14,11 +14,11 @@ union_composition<dna5, gap> letter2{'C'_dna5}; // constructed from alternative
 union_composition<dna5, gap> letter3{'U'_rna5}; // constructed from type that alternative is constructable from (== 'T'_dna5)
 
 letter2.assign_char('T');                      // == 'T'_dna5
-letter2.assign_char('-');                      // == gap::GAP
+letter2.assign_char('-');                      // == '-'_gap
 letter2.assign_char('K');                      // unknown characters map to the default/unknown
                                                // character of the first alternative type (== 'N'_dna5)
 
-letter2 = gap::GAP;                            // assigned from alternative (== gap::GAP)
+letter2 = '-'_gap;                            // assigned from alternative (== '-'_gap)
```
smehringer: indentation :nail_care:

include/seqan3/alphabet/gap/gap.hpp
smehringer: Hm alright, but then I would change this, as gap does not "have" a character.
```suggestion
 * \param[in] c Placeholder; A gap object will be created regardless of the given character.
```

smehringer: oh and add a test for this! assigning another char with he literal like `'f'_gap`. Although this is super wierd...

h-2:
> @h-2 Since you can also just use gap{}, I don't know if the literal is helpful as it is even longer? 

I thought the same thing when I saw in actual code :thinking: 

> Does this mean you can assign any character?

Yes, currently assign_char is guaranteed to never throw, it always converts. In this case all letter would be converted to '-'.

I agree that we maybe should just call the constructor for alphabets of size 1.

smehringer:
> The question is if we want to have "A-C-G-T"_gapped_dna15.

I'll save this for the next strategy meeting

@joshuak94 Sorry for the switch but can you please delete the literal again and change every `gap::GAP` to `gap{}`? Thank you :)

joshuak94:
@smehringer Re-committed, changing GAP::gap to gap{}.

smehringer:
Please tag your commit message with `"[MISC]" ... `.

joshuak94:
@smehringer Done!

PR #648, opened 2018-12-21
[FEATURE] Enable printing of std::variant for debug_stream.

smehringer:
Comes in handy when working with std::variant (e.g. sam_tag_dictionary)


include/seqan3/io/stream/debug_stream.hpp
```diff
@@ -334,6 +335,27 @@ inline debug_stream_type & operator<<(debug_stream_type & s, tuple_t && t)
...
+ *
+ * \details
+ *
+ * Note that in case the variant is valueless(_by_exception), nothing is printed.
+ */
+template <typename variant_type>
+//!\cond
+    requires detail::is_type_specialisation_of_v<remove_cvref_t<variant_type>, std::variant>
+//!\endcond
+inline debug_stream_type & operator<<(debug_stream_type & s, variant_type && v)
```
h-2: why not:
```cpp
template <typename ... ts>
inline debug_stream_type & operator<<(debug_stream_type & s, std::variant<ts...>  const & v)
```
... probably a lot easier on the compiler...

smehringer: That was my first intuition too, but if I do that, the compiler will not select this function because the generic member function is "equally suited" in overload resolution :disappointed: (and the generic one will then fail..)

h-2: and if you add `requires true` :rofl:

h-2: with which one does it compete? Maybe you need to add both `const &` and `&&`, but it should definitely be more constrained.

smehringer: It competes with the most generic member function in line 154
```cpp
    //!\brief Forwards to the underlying stream object.
    template <typename t>
    debug_stream_type & operator<<(t && v)
    {
        *stream << v;
        return *this;
    }
```
The compiler error is not an ambigous call by the way but it selects the generic member function and fails when passing the variant to the underlying stream. It doesn't work with `requires true` either, even if additionally add a const & overload.

smehringer:
@h-2 Your solution does not work with temporaries :( I added a test.

h-2:
Yeah, would need another overload with && ... Maybe the constrained solution is better after all? :unamused:

PR #652, opened 2019-01-07
[FEATURE] Char literals for structure alphabets

joergi-w:
Adds char literals to the following alphabets and removes the enum interface.
- [x] dot bracket
- [x] dssp
- [x] wuss

solves #644

### DIFF
--- a/include/seqan3/alphabet/structure/dot_bracket3.hpp
+++ b/include/seqan3/alphabet/structure/dot_bracket3.hpp
@@ -70,7 +70,7 @@ namespace seqan3
  *```
  *
  * \par Usage
- * The following code example creates a dot_bracket3 vector, modifies it, and prints the result to stdout.
+ * The following code example creates a dot_bracket3 vector, modifies it, and prints the result to stderr.
  * \snippet test/snippet/alphabet/structure/dot_bracket3.cpp general
  */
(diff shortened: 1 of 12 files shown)


include/seqan3/alphabet/structure/dot_bracket3.hpp
```diff
@@ -113,23 +102,23 @@ class dot_bracket3 : public alphabet_base<dot_bracket3, 3>
      */
     constexpr bool is_pair_open() const noexcept
     {
-        return *this == PAIR_OPEN;
+        return this->to_char() == '(';
```
smehringer: hm I wonder if the extra call to `to_char` makes a performance difference compared to directly comparing the rank with `this->rank == 1` :thinking:

include/seqan3/alphabet/structure/dot_bracket3.hpp
```diff
@@ -208,4 +183,20 @@ inline std::vector<dot_bracket3> operator""_db3(const char * str, std::size_t le
     return vec;
 }
 
+/*!
```
smehringer: put brief statement here :wink:

include/seqan3/alphabet/structure/dot_bracket3.hpp
joergi-w: Good point, so I will just compare the rank value.

PR #636, opened 2018-12-14
Feature/affine alignment

rrahn:
resolves #634 

### blocked by
 * #567

Work in Progress.

Implements the affine alignment kernel and the first selection.

PR #657, opened 2019-01-08
Feature/affine alignment 2

rrahn:
Depends on:
* #668 

Supersedes #636 

Implements the first version of an affine alignment kernel that computes the score.
It refines the alignment range and the two-way executor to be move only as they are dealing with a resource. Some small fixes of the single input pass view to allow default construction of it.
Refines the algorithm configuration to check pre conditions and throw if a configuration setting is yet not supported.

### DIFF
--- a/include/seqan3/alignment/pairwise/align_pairwise.hpp
+++ b/include/seqan3/alignment/pairwise/align_pairwise.hpp
@@ -25,7 +25,7 @@
 
 #include <seqan3/alignment/configuration/all.hpp>
 #include <seqan3/alignment/pairwise/align_result.hpp>
-#include <seqan3/alignment/pairwise/alignment_selector.hpp>
+#include <seqan3/alignment/pairwise/alignment_configurator.hpp>
 #include <seqan3/alignment/pairwise/execution/all.hpp>
 
 #include <seqan3/alphabet/gap/gapped.hpp>
@@ -46,9 +46,6 @@ template <std::ranges::InputRange sequence_t, typename alignment_config_t>
              tuple_like_concept<value_type_t<std::ranges::iterator_t<std::remove_reference_t<sequence_t>>>>
 constexpr auto align_pairwise(sequence_t && seq, alignment_config_t const & config)
 {
-    static_assert(std::tuple_size_v<value_type_t<std::ranges::iterator_t<std::remove_reference_t<sequence_t>>>> == 2,
-                  "Expects exactly two sequences for pairwise alignments.");
-
     // Wrap in persist to make the code also work with temporaries that are not views.
     auto seq_view = view::persist(std::forward<sequence_t>(seq));
     // Configure the alignment algorithm.
--- a/include/seqan3/alignment/pairwise/align_pairwise.hpp
+++ b/include/seqan3/alignment/pairwise/align_pairwise.hpp
@@ -44,19 +44,18 @@ namespace seqan3
 template <std::ranges::InputRange sequence_t, typename alignment_config_t>
     requires detail::is_type_specialisation_of_v<remove_cvref_t<alignment_config_t>, configuration> &&
              tuple_like_concept<value_type_t<std::ranges::iterator_t<std::remove_reference_t<sequence_t>>>>
-constexpr auto align_pairwise(sequence_t && seq, alignment_config_t && config)
+constexpr auto align_pairwise(sequence_t && seq, alignment_config_t const & config)
 {
     static_assert(std::tuple_size_v<value_type_t<std::ranges::iterator_t<std::remove_reference_t<sequence_t>>>> == 2,
                   "Expects exactly two sequences for pairwise alignments.");
 
-    using seq_tuple_t = std::remove_reference_t<decltype(*std::ranges::begin(seq))>;
+    // Make also work with temporaries.
+    auto seq_view = std::forward<sequence_t>(seq) | view::persist;
+    auto kernel = detail::alignment_configurator::configure(seq_view, config);
 
-    auto seq_view = seq | view::persist;
-    detail::alignment_selector<seq_tuple_t, remove_cvref_t<alignment_config_t>> selector{config};
-    using exec_type = detail::alignment_executor_two_way<decltype(seq_view), decltype(selector)>;
-    return alignment_range<exec_type>{seq_view, selector};
+    using exec_type = detail::alignment_executor_two_way<decltype(seq_view), decltype(kernel)>;
+    return alignment_range<exec_type>{seq_view, kernel};
 }
-//
 
 template <tuple_like_concept seq_t,
           typename alignment_config_t>
--- a/include/seqan3/alignment/pairwise/align_pairwise.hpp
+++ b/include/seqan3/alignment/pairwise/align_pairwise.hpp
@@ -49,12 +49,14 @@ constexpr auto align_pairwise(sequence_t && seq, alignment_config_t const & conf
     static_assert(std::tuple_size_v<value_type_t<std::ranges::iterator_t<std::remove_reference_t<sequence_t>>>> == 2,
                   "Expects exactly two sequences for pairwise alignments.");
 
-    // Make also work with temporaries.
-    auto seq_view = std::forward<sequence_t>(seq) | view::persist;
+    // Wrap in persist to make the code also work with temporaries that are not views.
+    auto seq_view = view::persist(std::forward<sequence_t>(seq));
+    // Configure the alignment algorithm.
     auto kernel = detail::alignment_configurator::configure(seq_view, config);
-
-    using exec_type = detail::alignment_executor_two_way<decltype(seq_view), decltype(kernel)>;
-    return alignment_range<exec_type>{seq_view, kernel};
+    // Create a two-way executor for the alignment.
+    detail::alignment_executor_two_way exec{std::move(seq_view), kernel};
+    // Return the range over the alignments.
+    return alignment_range{std::move(exec)};
 }
 
 template <tuple_like_concept seq_t,
@@ -65,7 +67,6 @@ constexpr auto align_pairwise(seq_t && seq, alignment_config_t && config)
     static_assert(std::tuple_size_v<std::remove_reference_t<seq_t>> == 2,
                   "Expects exactly two sequences for pairwise alignments.");
 
-    //TODO: Check the problem here.
     return align_pairwise(ranges::view::single(std::forward<seq_t>(seq)) | ranges::view::bounded,
                           std::forward<alignment_config_t>(config));
 }
--- a/include/seqan3/alignment/pairwise/alignment_algorithm.hpp
+++ b/include/seqan3/alignment/pairwise/alignment_algorithm.hpp
@@ -1,6 +1,6 @@
 // -----------------------------------------------------------------------------------------------------
-// Copyright (c) 2006-2018, Knut Reinert & Freie Universität Berlin
-// Copyright (c) 2016-2018, Knut Reinert & MPI für molekulare Genetik
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
 // This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
 // shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
 // -----------------------------------------------------------------------------------------------------
--- a/include/seqan3/alignment/pairwise/alignment_algorithm.hpp
+++ b/include/seqan3/alignment/pairwise/alignment_algorithm.hpp
@@ -1,0 +1,199 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2018, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2018, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides seqan3::detail::alignment_algorithm.
+ * \author Rene Rahn <rene.rahn AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <range/v3/algorithm/for_each.hpp>
+#include <range/v3/view/drop_exactly.hpp>
+#include <range/v3/view/zip.hpp>
+
+#include <seqan3/alignment/configuration/all.hpp>
+#include <seqan3/alignment/pairwise/policy/affine_gap_init_policy.hpp>
+#include <seqan3/alignment/pairwise/policy/affine_gap_policy.hpp>
+#include <seqan3/alignment/pairwise/policy/unbanded_dp_matrix_policy.hpp>
+#include <seqan3/alignment/pairwise/align_result_selector.hpp>
+
+#include <seqan3/alignment/scoring/gap_scheme.hpp>
+#include <seqan3/alignment/scoring/scoring_scheme_base.hpp>
+
+#include <seqan3/core/metafunction/deferred_crtp_base.hpp>
+#include <seqan3/io/stream/debug_stream.hpp>
+#include <seqan3/range/view/take_exactly.hpp>
+#include <seqan3/std/concepts>
+#include <seqan3/std/ranges>
+
+namespace seqan3::detail
+{
+
+/*!\brief The alignment algorithm type to compute standard pairwise alignment using dynamic programming.
+ * \implements std::Invocable
+ * \ingroup pairwise_alignment
+ * \tparam config_t             The configuration type; must be of type seqan3::configuration.
+ * \tparam algorithm_policies_t Template parameter pack with the policies to determine the execution of the algorithm;
+ *                              must be wrapped as seqan3::detail::deferred_crtp_base.
+ *
+ * \details
+ *
+ * ## Configuration
+ *
+ * The first template argument is the type of the alignment configuration which was used to configure the alignment
+ * algorithm type. They must be the same, otherwise it is possible that the output is not coherent with the expected
+ * result given the different configurations. The correct alignment is configured within the
+ * seqan3::detail::alignment_configurator and returned as a std::function object, which can be passed around and
+ * copied to create, for example multiple instances of the algorithm that can be executed in parallel.
+ *
+ * ## Policies
+ *
+ * This type uses multiple inheritance to configure a specific alignment computation, e.g. global or local alignments,
+ * using SIMD operations or scalar operations, computing the traceback or only the score etc.. These configurations
+ * are inherited using so-called `alignment policies`. An alignment policy is a type that implements a specific
+ * functionality through a common interface that is used by the alignment algorithm. These policies are also
+ * the customisation points of the algorithm which will be used to implement a specific behaviour. You can read more
+ * about the policies in \ref alignment_policy.
+ *
+ * Since some of the policies are augmented with traits to further refine the policy execution during the configuration,
+ * it is necessary to defer the template instantiation of the policies, which are modelled as CRTP-base classes.
+ * Therefore every added policy must be wrapped in a seqan3::detail::deferred_crtp_base class wrapper. This wrapper
+ * then is expanded during the final template instantiation of the alignment algorithm type using the corresponding
+ * template function seqan3::detail::invoke_deferred_crtp_base, which instantiates the CRTP policy types with the
+ * correct alignment algorithm type.
+ */
+template <typename config_t, typename ...algorithm_policies_t>
+class alignment_algorithm :
+    protected invoke_deferred_crtp_base<algorithm_policies_t, alignment_algorithm<algorithm_policies_t...>>...
+{
+public:
+    /*!\name Constructor, destructor and assignment
+     * \brief Defaulted all standard constructor.
+     * \{
+     */
+    constexpr alignment_algorithm()                                        = default;
+    constexpr alignment_algorithm(alignment_algorithm const &)             = default;
+    constexpr alignment_algorithm(alignment_algorithm &&)                  = default;
+    constexpr alignment_algorithm & operator=(alignment_algorithm const &) = default;
+    constexpr alignment_algorithm & operator=(alignment_algorithm &&)      = default;
+    ~alignment_algorithm()                                                 = default;
+
+    /*!\brief Constructs the algorithm with the passed configuration.
+     * \param cfg The configuration to be passed to the algorithm.
+     *
+     * \details
+     *
+     * Maintains a copy of the configuration object on the heap using a std::shared_ptr.
+     */
+    explicit constexpr alignment_algorithm(config_t const & cfg) : cfg_ptr{new config_t(cfg)}
+    {}
+    //!}
+
+    /*!\brief Invokes the actual alignment computation given two sequences.
+     * \tparam    first_batch_t  The type of the first sequence (or packed sequences); must model std::ForwardRange.
+     * \tparam    second_batch_t The type of the second sequence (or packed sequences); must model std::ForwardRange.
+     * \param[in] first_batch    The first sequence (or packed sequences).
+     * \param[in] second_batch   The second sequence (or packed sequences).
+     *
+     * \details
+     *
+     * ### Exception
+     *
+     * Strong exception guarantee.
+     *
+     * ### Thread-safety
+     *
+     * Calls to this functions in a concurrent environment are not thread safe. Instead use a copy of the alignment
+     * algorithm type.
+     *
+     * ### Complexity
+     *
+     * The code always runs in \f$ O(N^2) \f$ time and depending on the configuration requires at least \f$ O(N) \f$
+     * and at most \f$ O(N^2) \f$ space.
+     */
+    template <std::ranges::ForwardRange first_batch_t, std::ranges::ForwardRange second_batch_t>
+    auto operator()(first_batch_t const & first_batch, second_batch_t const & second_batch)
+    {
+        assert(cfg_ptr != nullptr);
+
+        // We need to allocate the score_matrix and maybe the trace_matrix.
+        this->allocate_score_matrix(first_batch, second_batch);
+
+        // Initialize cache variables to keep frequently used variables close to the CPU registers.
+        auto cache = this->make_cache(get<align_cfg::gap>(*cfg_ptr).value);
+
+        initialise_matrix(cache);
+
+        compute_matrix(first_batch, second_batch, cache);
+
+        using result_t = typename align_result_selector<first_batch_t, second_batch_t, config_t>::type;
+        result_t res{};
+
+        // Choose what needs to be computed.
+        if constexpr (config_t::template exists<align_cfg::result<detail::with_score_type>>())
+        {
+            auto last_col = this->current_column();
+            res.score = get<0>(*(std::ranges::end(last_col) - 1));
+        }
+        return align_result{res};
+    }
+
+private:
+
+    /*!\brief Initialises the first column of the dynamic programming matrix.
+     * \tparam         cache_t The cache type.
+     * \param[in,out]  cache   The cache holding hot variables.
+     */
+    template <typename cache_t>
+    void initialise_matrix(cache_t & cache)
+    {
+        auto col = this->current_column();
+
+        this->init_origin_cell(*std::ranges::begin(col), cache);
+
+        ranges::for_each(col | ranges::view::drop_exactly(1), [&cache, this](auto & cell)
+        {
+            this->init_column_cell(cell, cache);
+        });
+    }
+
+    /*!\brief Compute the alignment by iterating over the dynamic programming matrix in a column wise manner.
+     * \tparam        first_batch_t  The type of the first sequence (or packed sequences).
+     * \tparam        second_batch_t The type of the second sequence (or packed sequences).
+     * \tparam        cache_t        The type of the cache.
+     * \param[in]     first_batch    The first sequence.
+     * \param[in]     second_batch   The second sequence.
+     * \param[in,out] cache          The cache holding hot variables.
+     */
+    template <typename first_batch_t,
+              typename second_batch_t,
+              typename cache_t>
+    void compute_matrix(first_batch_t const & first_batch,
+                        second_batch_t const & second_batch,
+                        cache_t & cache)
+    {
+        auto const & score_scheme = get<align_cfg::scoring>(*cfg_ptr).value;
+        ranges::for_each(first_batch, [&, this](auto seq1_value)
+        {
+            auto col = this->current_column();
+            this->init_row_cell(*std::ranges::begin(col), cache);
+
+            auto second_batch_it = std::ranges::begin(second_batch);
+            ranges::for_each(col | ranges::view::drop_exactly(1), [&, this] (auto & cell)
+            {
+                this->compute_cell(cell, cache, score_scheme.score(seq1_value, *second_batch_it));
+                ++second_batch_it;
+            });
+        });
+    }
+
+    //!\brief The alignment configuration stored on the heap.
+    std::shared_ptr<config_t> cfg_ptr{};
+};
+
+} // namespace seqan3::detail
--- a/include/seqan3/alignment/pairwise/alignment_configurator.hpp
+++ b/include/seqan3/alignment/pairwise/alignment_configurator.hpp
@@ -1,0 +1,189 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2018, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2018, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
+// -----------------------------------------------------------------------------------------------------
+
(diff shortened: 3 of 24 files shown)


include/seqan3/alignment/pairwise/policy/affine_gap_policy.hpp
```diff
@@ -0,0 +1,101 @@
...
+        vt_score += gap_extend;
+        hz_score += gap_extend;
+        vt_score = std::max(vt_score, tmp);
+        hz_score = std::max(hz_score, tmp);
+    }
+
+    /*!\brief Creates the cache used for affine gap computation.
+     * \tparam    gap_scheme_t The type of the gap scheme.
+     * \param[in] scheme       The configured gap sheme.
+     * \returns The created cache.
```
joergi-w: Please document the meaning of the tuple elements. For instance, you redefine gap_open here as open_score + gap_score, which is different from the conventions and should therefore be clearly described. 
I suggest to assign only gap_open here and in the above function (compute_cell) do the sum.

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -71,19 +71,17 @@ namespace seqan3
 template <std::ranges::InputRange sequence_t, typename alignment_config_t>
     requires detail::is_type_specialisation_of_v<remove_cvref_t<alignment_config_t>, configuration> &&
              tuple_like_concept<value_type_t<std::ranges::iterator_t<std::remove_reference_t<sequence_t>>>>
-constexpr auto align_pairwise(sequence_t && seq, alignment_config_t && config)
+constexpr auto align_pairwise(sequence_t && seq, alignment_config_t const & config)
```
joergi-w: documentation missing (whole file)

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -0,0 +1,170 @@
...
+        result_t res{};
+
+        // Choose what needs to be computed.
+        if constexpr (config_t::template exists<align_cfg::result<detail::with_score_type>>())
+        {
+            auto last_col = this->active_column();
+            get<1>(res) = get<0>(*(std::ranges::end(last_col) - 1));
+            return res;
+        }
+        return res;
```
joergi-w: wrap in else {} statement

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+protected:
+
+    /*!\brief Initializes the first column of the dynamic programming matrix.
+     * \tparam         cache_t The cache type.
+     * \param[in,out]  cache   The cache holding hot variables.
+     */
+    template <typename cache_t>
+    void initialize(cache_t & cache)
```
joergi-w: `initialize` sounds like a constructor of alignment_algorithm. 
Consider a more descriptive name, e.g. `fill_1st_column`.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -0,0 +1,170 @@
...
+        });
+    }
+
+    /*!\brief Compute the alignment by iterating over the dynamic programming matrix in a column wise manner.
+     * \tparam        first_batch_t  The type of the first sequence (or packed sequences).
+     * \tparam        second_batch_t The type of the second sequence (or packed sequences).
+     * \tparam        cache_t        The type of the cache.
+     * \param[in]     first_batch    The first sequence.
+     * \param[in]     second_batch   The second sequence.
+     * \param[in,out] cache          The cache holding hot variables.
```
joergi-w: Please describe the cache better: which values are the hot ones?
(maybe it is enough to link to the actual cache class where the details are given)

include/seqan3/alignment/pairwise/alignment_configurator.hpp
```diff
@@ -0,0 +1,189 @@
...
+private:
+    /*!\brief Auxiliary member types
+     * \{
+     */
+    using first_seq_t  = std::tuple_element_t<0, value_type_t<std::ranges::iterator_t<range_type>>>;
+    using second_seq_t = std::tuple_element_t<1, value_type_t<std::ranges::iterator_t<range_type>>>;
+    //!\}
+
+public:
+    //!\brief Tests whether the value type of `range_type` is a tuple with exactly 2 members.
```
joergi-w: As said before, this could be an std::pair. In that case this test would be omitted.

include/seqan3/alignment/pairwise/select_result_type.hpp
```diff
@@ -0,0 +1,80 @@
...
+        using second_seq_value_type = value_type_t<second_batch_t>;
+        using score_type            = int32_t;
+
+        if constexpr (std::remove_reference_t<configuration_t>::template exists<align_cfg::result>())
+        {
+            if constexpr (configuration_t::template exists<align_cfg::result<with_end_position_type>>())
+                return align_result<type_list<uint32_t,
+                                              score_type,
+                                              alignment_coordinate>>{};
+            else if constexpr (configuration_t::template exists<align_cfg::result<with_begin_position_type>>())
```
joergi-w: use brackets for multi-line if statements

include/seqan3/alignment/pairwise/align_pairwise.hpp
rrahn: Yes, but this comes to a later point.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
rrahn: At the moment the cache is just internally defined by the gap policy but I have a tracking card to make this a separate class.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
rrahn: Mhmm, I am not against it per sé, but don't really see any improvements. The text books call this step initialization and the function is documented and no user will ever touch it.

rrahn: I will though make it private instead of protected.

include/seqan3/alignment/pairwise/alignment_configurator.hpp
rrahn: We cannot assume anything about the type here. So I cannot require that sequences always have to be passed as std::pair. For example using the ranges::view::zip (which is very likely in this context) will return a tuple not a pair.

include/seqan3/alignment/pairwise/alignment_configurator.hpp
```diff
@@ -0,0 +1,189 @@
...
+        // ----------------------------------------------------------------------------
+        // Test some basic preconditions
+        // ----------------------------------------------------------------------------
+
+        using alignment_contract_t = alignment_contract<remove_cvref_t<sequences_t>, config_t>;
+
+        static_assert(alignment_contract_t::expects_tuple_like_value_type(),
+                      "Alignment configuration error: "
+                      "The value type of the sequence ranges must fulfill the tupke_like_concept and must contain "
+                      "exactly 2 elements.");
```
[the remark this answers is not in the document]
rrahn: As I described above. This assumption would be to invasive.

include/seqan3/alignment/pairwise/alignment_configurator.hpp
```diff
@@ -0,0 +1,189 @@
...
+        return kernel_t{alignment_algorithm<config_t, dp_matrix_t, affine_t, init_t>{cfg}};
+    }
+
+private:
+
+    //!\brief Configure the edit distance algorithm.
+    template <typename kernel_t, typename config_t>
+    static constexpr auto configure_edit_distance(config_t const & cfg)
+    {
+        return kernel_t{edit_distance_wrapper<remove_cvref_t<config_t>>{cfg}};
```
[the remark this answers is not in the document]
rrahn: I need to refactor it to make it work with dynamically set end-gap specifiers. But this comes in another PR.

include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
```diff
@@ -604,6 +605,77 @@ pairwise_alignment_edit_distance_unbanded(database_t && database, query_t && que
...
+    {
+        using result_t = typename detail::select_result_type<remove_cvref_t<first_batch_t>,
+                                                             remove_cvref_t<second_batch_t>,
+                                                             remove_cvref_t<config_t>>::type;
+
+        pairwise_alignment_edit_distance_unbanded algo{std::forward<first_batch_t>(first_batch),
+                                                       std::forward<second_batch_t>(second_batch),
+                                                       *cfg_ptr};
+        result_t res{};
+        return algo(res);
```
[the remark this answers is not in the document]
rrahn: No RVO will kick in.

include/seqan3/alignment/pairwise/policy/affine_gap_policy.hpp
rrahn: Well, no the convention is actually that opening a gap is gap cost plus additional gap open costs. In SeqAn2 this was different. So I do the add in the cache generation in order to save the instruction in the hot kernel above.

rrahn: And yes as described before, I will make it a spearate class which is then documented and easier to follow.

include/seqan3/alignment/pairwise/select_result_type.hpp
rrahn: IIRC this follows our convention of single expression if-clauses don't necessarily require `{}`  🤔

rrahn:
@h-2 can you look at this PR? I addressed all issues from @joergi-w

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -0,0 +1,168 @@
...
+    /*!\brief Constructs the algorithm with the passed configuration.
+     * \param cfg The configuration to be passed to the algorithm.
+     *
+     * \details
+     *
+     * The configuration is copied once to the heap during construction and maintained by a std::shared_ptr.
+     * The configuration is not passed to the function-call-operator of this function object, in order to avoid
+     * incompatible configurations between the passed configuration and the one used during configuration of this
+     * class. Further will the function object be stored in a std::function which requires copyable objects and
+     * in parallel executions the function object must be copied as well.
```
h-2: I don't really understand this. Why do you pass a config to this object at all, and not just to `operator()`?

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -0,0 +1,168 @@
...
+    {
+        auto col = this->active_column();
+
+        this->init_origin_cell(*std::ranges::begin(col), cache);
+
+        ranges::for_each(col | ranges::view::drop_exactly(1), [&cache, this](auto & cell)
+        {
+            this->init_column_cell(cell, cache);
+        });
+    }
```
h-2: a lot of `this->` here, why?

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -0,0 +1,168 @@
...
+                this->compute_cell(cell, cache, score_scheme.score(seq1_value, *second_batch_it));
+                ++second_batch_it;
+            });
+        });
+    }
+
+private:
+
+    //!\brief The alignment configuration stored on the heap.
+    std::shared_ptr<config_t> cfg_ptr{};
```
h-2: why is the config a member of the type, but the cache a local variable? Are any of these things called in parallel?

include/seqan3/alignment/pairwise/execution/execution_handler_sequential.hpp
```diff
@@ -52,27 +52,18 @@ namespace seqan3::detail
...
-    /*!\name Constructors, destructor and assignment
-     * \{
-     */
-    execution_handler_sequential()                                                 = default;
-    execution_handler_sequential(execution_handler_sequential const &)             = default;
-    execution_handler_sequential(execution_handler_sequential &&)                  = default;
-    execution_handler_sequential & operator=(execution_handler_sequential const &) = default;
-    execution_handler_sequential & operator=(execution_handler_sequential &&)      = default;
-    ~execution_handler_sequential()                                                = default;
-    //!}
```
h-2: why are these removed? If this class is supposed to be an aggregate, please make it a struct

include/seqan3/alignment/pairwise/execution/execution_handler_sequential.hpp
```diff
@@ -52,27 +52,18 @@ namespace seqan3::detail
...
     //!\brief Invokes the passed alignment instance in a blocking manner.
-    template <typename func_result_t>
-    void execute(std::function<func_result_t(func_result_t &)> func,
-                 func_result_t & res,
-                 std::function<void(decltype(func(res)))> delegate)
+    template <typename func_result_t, typename first_batch_t, typename second_batch_t>
+    void execute(std::function<func_result_t(first_batch_t const &, second_batch_t const &)> func,
+                  first_batch_t const & first_batch,
+                  second_batch_t const & second_batch,
+                  std::function<void(decltype(func(first_batch, second_batch)))> delegate)
```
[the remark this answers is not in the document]
h-2: still open

include/seqan3/alignment/pairwise/policy/affine_gap_init_policy.hpp
```diff
@@ -0,0 +1,103 @@
...
+    /*!\brief Initializes the origin of the dynamic programming matrix.
+     * \tparam        cell_t      The underlying cell type.
+     * \tparam        cache_t     The type of the cache.
+     * \param[in,out] active_cell The current cell in the dynamic programming matrix.
+     * \param[in,out] cache       The cache storing hot helper variables.
+     */
+    template <typename cell_t, typename cache_t>
+    constexpr auto init_origin_cell(cell_t & active_cell, cache_t & cache) const noexcept
+    {
+        using std::get;
```
h-2: what for?

include/seqan3/alignment/pairwise/policy/affine_gap_policy.hpp
```diff
@@ -0,0 +1,101 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2018, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2018, Knut Reinert & MPI für molekulare Genetik
```
h-2: 2019 everywhere

include/seqan3/alignment/pairwise/policy/affine_gap_init_policy.hpp
```diff
@@ -0,0 +1,103 @@
...
+        auto & [main_score, hz_score] = active_cell;
+        auto & [prev_cell, gap_open, gap_extend] = cache;
+        auto & vt_score = get<1>(prev_cell);
+
+        main_score = 0;
+        hz_score = gap_open + gap_extend;
+        vt_score = gap_open + gap_extend;
+    }
+
+    /*!\brief Initializes the cells in the first column of the dynamic programming matrix.
```
h-2: in British English it's initiali**s**e

(often in your code!)

include/seqan3/alignment/pairwise/policy/affine_gap_policy.hpp
```diff
@@ -0,0 +1,101 @@
...
+    constexpr void compute_cell(score_cell_type & active_cell,
+                                cache_t & cache,
+                                score_t const score) const noexcept
+    {
+        // Unpack the cached variables.
+        auto & [main_score, hz_score] = active_cell;
+        auto & [prev_cell, gap_open, gap_extend] = cache;
+        auto & [prev_score, vt_score] = prev_cell;
+
+        // can we use GCC's builtin saturation arithmetics?
```
h-2: add a TODO maybe with a link what this is about?

include/seqan3/alignment/pairwise/policy/affine_gap_policy.hpp
h-2: maybe also not call it `setup`? I would assume that any member function called `setup` somehow changes the object, but this doesn't.

include/seqan3/alignment/pairwise/policy/unbanded_dp_matrix_policy.hpp
```diff
@@ -0,0 +1,84 @@
...
+     * \tparam first_batch_t   The type of the first sequence (or packed sequences).
+     * \tparam second_batch_t  The type of the second sequence (or packed sequences).
+     * \param[in] first_batch  The first sequence (or packed sequences).
+     * \param[in] second_batch The first sequence (or packed sequences).
+     */
+    template <typename first_batch_t, typename second_batch_t>
+    void allocate_score_matrix(first_batch_t && first_batch, second_batch_t && second_batch)
+    {
+        dimension_first_batch = std::ranges::size(std::forward<first_batch_t>(first_batch));
+        dimension_second_batch = std::ranges::size(std::forward<second_batch_t>(second_batch));
```
h-2: Why do you forward to size?

include/seqan3/alignment/pairwise/policy/unbanded_dp_matrix_policy.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+        // We use only one column to compute the score.
+        score_matrix.resize(dimension_second_batch + 1);
+    }
+
+    //!\brief Returns the active column that is computed next.
+    auto active_column()
```
h-2: is "active == current" or is "active == next" :thinking:

include/seqan3/alignment/pairwise/select_result_type.hpp
```diff
@@ -0,0 +1,83 @@
...
+            else if constexpr (std::Same<remove_cvref_t<decltype(get<align_cfg::result>(configuration_t{}).value)>,
+                                         with_trace_type>)
+                return align_result_value_type<uint32_t,
+                                               score_type,
+                                               alignment_coordinate,
+                                               alignment_coordinate,
+                                               std::tuple<std::vector<first_seq_value_type>,
+                                                          std::vector<second_seq_value_type>>>{};
+            else
+                return align_result_value_type<uint32_t, score_type>{};
```
h-2: braces around individual blocks because they are multiline

include/seqan3/alignment/pairwise/select_result_type.hpp
```diff
@@ -0,0 +1,83 @@
...
+ * \tparam second_batch_t  The type of the second sequence.
+ * \tparam configuration_t The configuration type. Must be of type seqan3::detail::configuration
+ */
+template <std::ranges::ForwardRange first_batch_t,
+          std::ranges::ForwardRange second_batch_t,
+          typename configuration_t>
+//!\cond
+    requires is_type_specialisation_of_v<remove_cvref_t<configuration_t>, configuration>
+//!\endcond
+struct select_result_type
```
h-2: not so important but the name sounds very general...

test/unit/alignment/pairwise/align_pairwise_test.cpp
```diff
@@ -118,11 +123,12 @@ TEST(align_pairwise, multiple_rng_lvalue)
...
-        EXPECT_EQ(std::string{gap2 | view::to_char}, "A-GTGATACT");
-    }
+    EXPECT_THROW(align_pairwise(vec, cfg), std::domain_error);
+    // for (auto && res : align_pairwise(vec, cfg))
+    // {
+    //     EXPECT_EQ(res.score(), -4);
+    //     auto && [gap1, gap2] = res.trace();
+    //     EXPECT_EQ(std::string{gap1 | view::to_char}, "ACGTGATG--");
+    //     EXPECT_EQ(std::string{gap2 | view::to_char}, "A-GTGATACT");
+    // }
```
h-2: all actual tests commented out?

test/unit/alignment/pairwise/alignment_selector_test.cpp
```diff
@@ -87,25 +90,38 @@ TEST(alignment_selector, determine_result_type)
...
+        auto res = fn(std::get<0>(*seqan3::begin(seq_rng)), std::get<1>(*seqan3::begin(seq_rng)));
+        std::cout << res.get_score() << std::endl;
+        // using _t = detail::alignment_selector<seq_t, decltype(cfg)>;
+
+        // using res_t = typename detail::select_result_type<seq1_t, seq2_t, decltype(cfg)>::type;
+        //
+        // seq_t s{};
+        // EXPECT_TRUE((std::is_same_v<res_t, typename _t::result_type>));
+        // EXPECT_TRUE((std::is_same_v<decltype(_t{cfg}.select(s)),
+        //                             std::function<res_t(res_t &)>>));
```
h-2: ...?

test/unit/alignment/pairwise/global_affine_unbanded_test.cpp
```diff
@@ -0,0 +1,214 @@
...
+//
+//     std::vector database = fixture.sequence1;
+//     std::vector query = fixture.sequence2;
+//
+//     auto alignment = edit_distance<word_type>(database, query, align_cfg);
+//
+//     auto && [gapped_database, gapped_query] = alignment.trace();
+//     EXPECT_EQ(std::string{gapped_database | view::to_char}, fixture.gapped_sequence1);
+//     EXPECT_EQ(std::string{gapped_query | view::to_char}, fixture.gapped_sequence2);
+// }
```
h-2: ...?

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -0,0 +1,168 @@
...
+    constexpr alignment_algorithm & operator=(alignment_algorithm &&)      = default;
+    ~alignment_algorithm()                                                 = default;
+
+    /*!\brief Constructs the algorithm with the passed configuration.
+     * \param cfg The configuration to be passed to the algorithm.
+     *
+     * \details
+     *
+     * The configuration is copied once to the heap during construction and maintained by a std::shared_ptr.
+     * The configuration is not passed to the function-call-operator of this function object, in order to avoid
```
eseiler:
```suggestion
     * The configuration is not passed to the function-call-operator of this function object in order to avoid
```

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -0,0 +1,168 @@
...
+    auto operator()(first_batch_t const & first_batch, second_batch_t const & second_batch)
+    {
+        assert(cfg_ptr != nullptr);
+
+        // We need to allocate the score_matrix and maybe the trace_matrix.
+        this->allocate_score_matrix(first_batch, second_batch);
+
+        // Initialize cache variables to keep frequently used variables close to the CPU registers.
+        auto cache = this->setup_cache(get<align_cfg::gap>(*cfg_ptr).value);
+        // auto cache = this->setup_cache(gap_scheme{gap_score{-1}, gap_open_score{-10}});
```
eseiler: Debug stuff?

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -0,0 +1,168 @@
...
+        {
+            auto last_col = this->active_column();
+            res.score = get<0>(*(std::ranges::end(last_col) - 1));
+        }
+        return align_result{res};
+    }
+
+protected:
+
+    /*!\brief Initializes the first column of the dynamic programming matrix.
```
eseiler: It should be BE everywhere... so `Initialises`

include/seqan3/alignment/pairwise/alignment_configurator.hpp
```diff
@@ -0,0 +1,189 @@
...
+                    return configure_edit_distance<kernel_t>(cfg);
+            }
+        }
+
+        // ----------------------------------------------------------------------------
+        // Unsupported configurations
+        // ----------------------------------------------------------------------------
+
+        if constexpr (config_t::template exists<align_cfg::aligned_ends>())
+            throw std::domain_error{"Configuring the aligned ends is yet not supported."};
```
eseiler: Just for understanding:

Why not `static_assert`?
It's static constexpr, so it needs to be known at compile time, right?

In any case, would a throw in a constexpr context occur during compiling? (Never tested it 😁 )

include/seqan3/alignment/pairwise/alignment_selector.hpp
```diff
@@ -45,102 +18,79 @@
...
-        using score_type      = int32_t;
+        return kernel_t{edit_distance_wrapper<remove_cvref_t<config_t>>{cfg}};
+    }
+
+    //!\brief Configure the algorithm.
+    template <std::ranges::View sequences_t, typename config_t>
+        requires is_type_specialisation_of_v<remove_cvref_t<config_t>, configuration>
+    static constexpr auto configure(sequences_t seq_range, config_t const & cfg)
+    {
+        using first_seq_t  = std::remove_reference_t<std::tuple_element_t<0, remove_cvref_t<decltype(*seqan3::begin(seq_range))>>>;
```
eseiler: Line lengths

include/seqan3/alignment/pairwise/policy/affine_gap_init_policy.hpp
```diff
@@ -0,0 +1,103 @@
...
+        auto & [main_score, hz_score] = active_cell;
+        auto & [prev_cell, gap_open, gap_extend] = cache;
+        auto & vt_score = get<1>(prev_cell);
+
+        main_score = vt_score;
+        hz_score = main_score + gap_open + gap_extend;
+        vt_score += gap_extend;
+    }
+
+    /*!\brief Initializes the cells in the first row of the dynamic programming matrix.
```
eseiler: Actually it initialises **a** cell of the first row, no?

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
rrahn: I think this is all valid according to our style guides

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -0,0 +1,168 @@
...
+/*!\brief A function object that computes pairwise alignments given two sequences.
+ * \ingroup pairwise_alignment
+ * \tparam config_t             The configuration type; must be of type seqan3::configuration.
+ * \tparam algorithm_policies_t Template parameter pack with the policies to determine the execution of the algorithm;
+ *                              must be wrapped as seqan3::detail::deferred_crtp_base.
+ *
+ * \details
+ *
+ * Computes the actual pairwise alignment given two sequences. The correct algorithm is configured while parsing
+ * the seqan3::configuration within the detail::seqan3::configurator.
```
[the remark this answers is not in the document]
rrahn: Ah now I see what you mean 🤦‍♂️

test/unit/alignment/pairwise/align_pairwise_test.cpp
rrahn: Yes they are going to be refined anyway with more code that I add. This PR is already old, and there are newer once, that resolve already parts of it.

rrahn:
@h-2 @eseiler I think I addressed all issues. I added documenation for the policies in the respective all header header and enhanced in general the documentation a little bit.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -0,0 +1,199 @@
...
+                ++second_batch_it;
+            });
+        });
+    }
+
+    //!\brief The alignment configuration stored on the heap.
+    std::shared_ptr<config_t> cfg_ptr{};
+};
+
+} // namespace seqan3::detail
```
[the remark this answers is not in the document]
rrahn: Mhmm, usually the editior does this automatically.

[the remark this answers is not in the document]
rrahn: No all files have a new line at the end. They are not displayed in the code diff

include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
```diff
@@ -27,105 +28,131 @@ namespace seqan3::detail
...
+     * \brief The class is not copy-constructible or copy-assignable but allows move construction and assignment.
      * \{
      */
-    alignment_executor_two_way()                                               = default;
-    alignment_executor_two_way(alignment_executor_two_way const &)             = default;
-    alignment_executor_two_way(alignment_executor_two_way &&)                  = default;
-    alignment_executor_two_way & operator=(alignment_executor_two_way const &) = default;
-    alignment_executor_two_way & operator=(alignment_executor_two_way &&)      = default;
-    ~alignment_executor_two_way()                                              = default;
+    alignment_executor_two_way() = delete;
```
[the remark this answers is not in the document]
rrahn: I had this in  my snippet but then decided that it is not worth it. So I changed it but not everywhere.

rrahn:
The failing test was due to time limits for the coverage builds. So everything else seems to work.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
h-2: :thinking:

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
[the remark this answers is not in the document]
eseiler: Mhm ok.
I looked at the raw file and there was none 🤔

include/seqan3/alignment/pairwise/policy/affine_gap_init_policy.hpp
h-2: ADL means only looking in associated namespaces (and friends and some other things). Writing `using ...` is not ADL, it means "consider the function from this namespace **in addition to ADL**". 
But you aren't call get anywhere anyway. And the structured bindings are explicitly required to not look at the current namespace or any `using ...` you might have done, they are only allowed to consider member get and ADL get.
If it doesn't compile without `using std::get;` than I am pretty sure it is a bug if it does with it.

include/seqan3/alignment/pairwise/policy/unbanded_dp_matrix_policy.hpp
h-2: You wouldn't want the size function to take ownership and destry your ranges :wink

test/unit/alignment/pairwise/global_affine_unbanded_test.cpp
h-2: But why commit dead code? If the other PRs change these lines anyway, why can't they add the code in the first place?

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
h-2: It's just that then the burder is on the user to take care of the cache, right?

include/seqan3/alignment/pairwise/policy/affine_gap_policy.hpp
h-2: yeah, make_cache is better, although I still feel that it is something that should be local to the object and created automatically, but that's up to you.

include/seqan3/alignment/pairwise/policy/all.hpp
```diff
@@ -0,0 +1,132 @@
...
+ *     This function gets the current cell dereferenced from the score matrix and the alignment algorithm cache.
+ *
+ * - init_row_cell:
+ *
+ *     Implements the initialisation of the cells in the first column of the matrix \f$ (M(0,j)) \f$.
+ *     This function gets the current cell dereferenced from the score matrix and the alignment algorithm cache.
+ *
+ * ### Existing gap init policies:
+ *
+ *  - seqan3::detail::affine_gap_init_policy
```
h-2: Grwat explanation! But is this really a costumisation point, i.e. can users just add their own policy, somehow squeeze it into the config and everything will work?

Maybe you could make it a little clearer what part of this documentation to refers to users adding custom behaviour in their app and which part only addresses developers of the library (most policies seem to be detail anyway?). And maybe mark respective things with `\if DEV` to it is only visible in the developer documentation?

rrahn: Oh, indeed it needs to be put into DEV. I forot this.

The intention was to provide an easier way to customise the alignment algorithms for library developers. Because most changes will involve structural changes in the algorithm and there
are depdencies between the data structures and layouts changing one thing means adapting others. Hence, the list of policies, what they are for, which interfaces they have to obey and that I need some more descriptive classes rather than plain tuples to interface with the different policies. So there are dpendencies but in general they are designed such that they can be replaced by some other implementation. That's why I would like to keep them noted as customisation points.

include/seqan3/alignment/pairwise/policy/affine_gap_policy.hpp
rrahn: I'll revisit this when I do more benchmarks. This is all detail code anyway so I can refactor it later without breaking API

include/seqan3/alignment/pairwise/select_result_type.hpp
rrahn: I changed it

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
rrahn: No? The user never sees nor touches any of this. This is all implementation specific stuff for library developers only.

test/unit/alignment/pairwise/global_affine_unbanded_test.cpp
rrahn: Well, well, I'll see that I remove it.

PR #658, opened 2019-01-08
String literals for the quality alphabets

smehringer:


include/seqan3/alphabet/quality/phred68legacy.hpp
```diff
@@ -116,6 +116,27 @@ constexpr phred68legacy operator""_phred68legacy(char const c) noexcept
...
+
+/*!\brief The seqan3::phred68legacy string literal.
+ * \param[in] s A pointer to the character sequence to assign from.
+ * \param[in] n The length of the character sequence to assign from.
+ * \relates seqan3::phred68legacy
+ * \returns seqan3::std::vector<phred68legacy>
+ *
+ * You can use this string literal to easily assign to std::vector<phred68legacy>:
+ *
+ * \include test/snippet/alphabet/nucleotide/phred68legacy_literal.cpp
```
joergi-w: Isn't they key word `\snippet`?

joergi-w: I looked it up and it includes the whole file. 
However, I'm still against this, as someone might add another snippet to that file later and then this code has to be found and updated, which can be an error source at some point.

include/seqan3/alphabet/quality/phred68legacy.hpp
```diff
@@ -116,6 +116,27 @@ constexpr phred68legacy operator""_phred68legacy(char const c) noexcept
...
+ */
+inline std::vector<phred68legacy> operator""_phred68legacy(char const * s, std::size_t n)
```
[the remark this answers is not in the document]
smehringer: The name is a design decision that is not part of this PR. But we maybe we do not want the phred68legacy string literal @h-2 ?

include/seqan3/alphabet/quality/phred68legacy.hpp
smehringer: Both are possible. I remember @marehr reviewing one of my PRs saying that if I include the whole file anyway , I would not use \snippet but \include. So I'm not sure how to handle this.

PR #659, opened 2019-01-09
[FIX] Fix from_chars for float 

smehringer:
Treat the second parameter as the end of the sequence to parse.

### DIFF
--- a/include/seqan3/std/charconv_detail.hpp
+++ b/include/seqan3/std/charconv_detail.hpp
@@ -80,13 +80,14 @@
 
 #pragma once
 
-#include <cstdlib>
-#include <type_traits>
-#include <limits>
-#include <cstdint>
-#include <cstring>
-#include <cmath>
+#include <algorithm>
 #include <cerrno>
+#include <cmath>
+#include <cstdint>
+#include <cstdlib>
+#include <cstring>
+#include <limits>
+#include <type_traits>
 
 #include <seqan3/core/concept/core_language.hpp>
 #include <seqan3/io/stream/parse_condition.hpp>
@@ -725,10 +726,8 @@ inline std::from_chars_result from_chars_floating_point(char const * first,
         return {last, std::errc::invalid_argument};
 
     float tmp{};
-    ptrdiff_t constexpr buffer_size = 10000;
-    char hex_buffer[buffer_size];
-    char * start;
-    char * end;
+    ptrdiff_t constexpr buffer_size = 100;
+    char buffer[buffer_size];
 
     if (fmt != std::chars_format::general)
     {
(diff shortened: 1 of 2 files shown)


include/seqan3/std/charconv_detail.hpp
```diff
@@ -724,11 +724,11 @@ inline std::from_chars_result from_chars_floating_point(char const * first,
     if (*first == '+') // + is permitted in function strto[d/f/ld] but not in from_chars
         return {last, std::errc::invalid_argument};
 
+    using ptrdiff_type = decltype(last - first);
```
h-2: why is this necessary?

include/seqan3/std/charconv_detail.hpp
```diff
@@ -758,40 +758,46 @@ inline std::from_chars_result from_chars_floating_point(char const * first,
...
 
+    // In contrast to std::from_chars, std::strto[f/d/ld] does not treat the second
+    // parameter (str_end) as "end of the sequence to parse" but merely as an out
+    // parameter to indicate where the parsing ended. Therefore, a buffer is
+    // needed to represent the truncated sequence and ensure correct from_chars
+    // functionality.
+    ptrdiff_type str_length{std::min(ptrdiff_type{buffer_size - offset}, last - first)};
+
+    for (ptrdiff_type i = 0; i < str_length; ++i)
+        buffer[i+offset] = first[i];
```
h-2: use std::copy?

include/seqan3/std/charconv_detail.hpp
```diff
@@ -758,40 +758,46 @@ inline std::from_chars_result from_chars_floating_point(char const * first,
...
 
     if (errno == ERANGE)
     {
         return {last, std::errc::result_out_of_range};
     }
-    else if (tmp == 0 && end == first)
+    else if (tmp == 0 && end == &buffer[0])
     {
         return {last, std::errc::invalid_argument};
     }
```
h-2: As discussed I would recommend only copying to the buffer when the input is not alread 0-terminated.

include/seqan3/std/charconv_detail.hpp
smehringer: I changed it back

include/seqan3/std/charconv_detail.hpp
```diff
@@ -87,6 +87,7 @@
 #include <cstring>
 #include <cmath>
 #include <cerrno>
+#include <algorithm>
```
h-2: order

test/unit/std/charconv_test.cpp
```diff
@@ -270,16 +280,18 @@ TYPED_TEST(from_char_real_test, real_numbers)
         TypeParam val{};
         std::string str = "1234";
         auto res = std::from_chars(&str[0], &str[0] + str.size(), val);
-        EXPECT_EQ(val, TypeParam{1234});
+        EXPECT_TRUE(double_equals(val, TypeParam{1234}));
```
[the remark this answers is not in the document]
smehringer: cool thanks! I didn't know they have this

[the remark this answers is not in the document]
smehringer: .... My tests fail with `EXPECT_DOUBLE_EQ` from google test since the epsilon is too small :scream: 
```
  val // from_chars output
    Which is: -0.30000001192092896
  TypeParam{-0.3} // direct initialization
    Which is: -0.29999999999999999
```

... @h-2 @joergi-w Do you think this is too severe? ... But I think if anything this is the fault of `std::strtof`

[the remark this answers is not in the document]
joergi-w: Yes, that's exactly what I would expect :)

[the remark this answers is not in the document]
smehringer: alright float then

PR #666, opened 2019-01-11
wuss alphabet: ensure size, rename test file

joergi-w:
This PR adds a constraint to the wuss alphabet size template parameter to avoid invalid types. 

The file containing the tests was renamed according to its corresponding hpp-file, where the (flexible) alphabet size is omitted.


include/seqan3/alphabet/structure/wuss.hpp
```diff
@@ -79,6 +79,9 @@ namespace seqan3
  * \snippet test/snippet/alphabet/structure/wuss.cpp general
  */
 template <uint8_t SIZE = 51>
+//!\cond
+    requires (SIZE >= 15 && SIZE <= 67 && SIZE % 2 == 1)
+//!\endcond
 class wuss : public alphabet_base<wuss<SIZE>, SIZE>
```
h-2: I would make a static_assert instead, because the condition does not participate in specialisation or overload resolution and with the static_assert you can give an error message to the user.

joergi-w: Okay, I agree!

PR #669, opened 2019-01-14
[MISC] Short headers and copyright year update

### LINKED ISSUE #553: Short license headers
In https://github.com/seqan/seqan3/pull/523#issuecomment-436177165 we shortly discussed that we might want to introduce shorter license headers.

For example, transforming

https://github.com/seqan/seqan3/blob/ffca425fecf8e2a490e1493399b866fe1c3ee689/test/unit/core/bit_manipulation_test.cpp#L1-L33

into

```c++
// This software may be modified and distributed under the terms of the BSD license.  See the LICENSE file for details.
```
This would reduce 33 lines down to 1 line.

I propose to do those changes only to files that we don't ship. Which means all files except for `include/` and `build_system/` should be changed.

----

Thoughts, suggestion, remarks?

eseiler:
Resolves #553

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,5 +1,5 @@
-Copyright (c) 2006-2018, Knut Reinert & Freie Universitaet Berlin
-Copyright (c) 2016-2018, Knut Reinert & MPI Molekulare Genetik
+Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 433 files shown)


eseiler:
CodeCov finds some changes not related to changes in this PR 🤷‍♂️

LICENSE
```diff
@@ -1,5 +1,5 @@
-Copyright (c) 2006-2018, Knut Reinert & Freie Universitaet Berlin
-Copyright (c) 2016-2018, Knut Reinert & MPI Molekulare Genetik
+Copyright (c) 2006-2019, Knut Reinert & Freie Universitaet Berlin
+Copyright (c) 2016-2019, Knut Reinert & MPI Molekulare Genetik
 All rights reserved.
```
h-2: Can you check the existing "new" copyright headers? We changed them to include Umlauts and also write "MPI für ..."

eseiler: Thanks for catching this. Was only the case in this file.

PR #671, opened 2019-01-14
[FIX] fix sequence_file_output writing of SEQ_QUAL field.

smehringer:
Caught by @joshuak94 when going through the snippets. This was never tested so it was never caught. I added tests for it.

### DIFF
--- a/test/unit/io/sequence_file/sequence_file_output_test.cpp
+++ b/test/unit/io/sequence_file/sequence_file_output_test.cpp
@@ -12,9 +12,9 @@
 #include <range/v3/view/zip.hpp>
 #include <range/v3/view/filter.hpp>
 
+#include <seqan3/alphabet/quality/phred42.hpp>
 #include <seqan3/io/sequence_file/output.hpp>
 #include <seqan3/range/shortcuts.hpp>
-#include <seqan3/range/view/convert.hpp>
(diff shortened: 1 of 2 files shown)


test/unit/io/sequence_file/sequence_file_output_test.cpp
```diff
@@ -372,6 +379,27 @@ TEST(row, different_fields_in_record_and_file)
...
+
+    std::vector<std::vector<qualified<dna5, phred42>>> seq_quals{3};
+    for (size_t i = 0; i < 3; ++i)
+    {
+        seq_quals[i].resize(quals[i].size());
+        for (size_t j = 0; j < quals[i].size(); ++j)
+        {
+            seq_quals[i][j] = seqs[i][j];
+            seq_quals[i][j] = quals[i][j];
+        }
```
h-2: std::copy?

Also you don't need the vector of seq_quals if you immediatly emplace_back the current value anyway, right?

test/unit/io/sequence_file/sequence_file_output_test.cpp
```diff
@@ -426,6 +454,28 @@ TEST(columns, assign_tuple_of_columns)
...
```
h-2: here you need the vector of vectors, but you might also be able to use std::copy

PR #676, opened 2019-01-15
[FIX] Adapts the weakly incrementable concept according to current standard draft.

rrahn:
Fixes the WeaklyIncrementable by adding the proposed iter_difference_t metafunction and the incrementable_traits class.

### DIFF
--- a/include/seqan3/std/iterator
+++ b/include/seqan3/std/iterator
@@ -54,9 +54,79 @@
 
 namespace std
 {
-
 //!\addtogroup iterator
 //!\{
+
+/*!\brief A type trait class that provides uniform interface to the properties of types that model the
+ *        WeaklyIncrementable concept.
+ * \tparam t The type to query.
+ * \sa https://en.cppreference.com/w/cpp/iterator/incrementable_traits
+ */
+template <typename t>
+struct incrementable_traits
+{};
+
+//!\cond
+template <typename t>
+    requires std::is_object_v<t>
+struct incrementable_traits<t*>
+{
+    using difference_type = ptrdiff_t;
+};
+
+template <typename t>
+struct incrementable_traits<const t> : incrementable_traits<t>
+{};
+
+template <typename t>
+    requires requires { typename t::difference_type; }
+struct incrementable_traits<t>
+{
+    using difference_type = typename t::difference_type;
+};
+
+template <typename t>
+    requires (!requires { typename t::difference_type; } &&
+              requires(t const & a, t const & b) { requires Integral<decltype(a - b)>; })
+struct incrementable_traits<t>
+{
+    using difference_type = std::make_signed_t<decltype(declval<t>() - declval<t>())>;
+};
+//!\endcond
+
+//!\cond
+namespace detail
+{
+// The type does not support incrementable concepts.
+template <typename t>
+struct incrementable_traits_or_iterator_traits
+{};
+
+// Use incrementable_traits<t>::difference_type if it is defined.
+template <typename t>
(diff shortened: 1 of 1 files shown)


include/seqan3/std/iterator
```diff
@@ -117,9 +184,9 @@ template <typename t>
 concept WeaklyIncrementable = Semiregular<t> &&
                                    requires (t v)
 {
-    typename std::iterator_traits<std::remove_reference_t<t>>::difference_type;
-    requires SignedIntegral<typename std::iterator_traits<std::remove_reference_t<t>>::difference_type>;
-    { ++v } -> t &; /* not required to be equality preserving */
+    typename iter_difference_t<std::remove_reference_t<t>>;
```
smehringer: Do you need the typename here? :thinking:

include/seqan3/std/iterator
```diff
@@ -55,6 +55,73 @@
...
+template <typename t>
+    requires requires { typename t::difference_type; }
+struct incrementable_traits<t>
+{
+    using difference_type = typename t::difference_type;
+};
+
+template <typename t>
+    requires (!requires { typename t::difference_type; } &&
+              requires { typename iterator_traits<t>::difference_type; })
```
smehringer: 1) Why this specialisation? I don't see it at cppreference (https://en.cppreference.com/w/cpp/iterator/incrementable_traits). 

2) Is this maybe ambiguous to the specialisation below? (If a type has `typename iterator_traits<t>::difference_type` but also fulfills `requires(t const & a, t const & b) { requires Integral<decltype(a - b)>; }`) ?

include/seqan3/std/iterator
```diff
@@ -55,6 +55,73 @@
...
+struct incrementable_traits<t>
+{
+    using difference_type = typename iterator_traits<t>::difference_type;
+};
+
+template <typename t>
+    requires (!requires { typename t::difference_type; } &&
+              requires(t const & a, t const & b) { requires Integral<decltype(a - b)>; })
```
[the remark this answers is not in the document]
h-2: the simpler thing would fail on types where the `operator-` is not defined.

include/seqan3/std/iterator
```diff
@@ -55,6 +55,73 @@
 namespace std
 {
 
+/*!\defgroup iterator_associated_types Associated types
+ * \ingroup iterator
+ * \brief Iterator requirements
+ */
```
h-2: you are not defining a group here, but a (sub-)module, is that intended?

h-2: In general we need to discuss if you shouldn't maybe remove all doc in the std module and finally get the plugin working that redirects to cppreference.com

rrahn: Yes, that would be good. The added subgroup follows the sections of the standard draft, in case other assoicated types will be added as well.

rrahn:
@smehringer @h-2 can you have another look.

include/seqan3/std/iterator
h-2: Yes, but what I wanted to say was: We have a very clear structure in our documentation:
Modules and sub-modules correspond to folders and sub-folders.

The std-module is breaking this.

rrahn: Yes, I am seeing your concern. But I can't group them otherwise, can I?

rrahn: Maybe `\section`?

rrahn: Ok, I couldn't find anything usful. I really don't care. I can revert this to make it plain in the iterator and we can change it later, if we decide for it.

PR #680, opened 2019-01-17
Aligned_sequence_concept - add the assign_ungapped_sequence function

smehringer:
Also adds (individual commits)

* is_gapped_alphabet helper and constrains the container functions to the ones with value type seqn3::gapped
* the ungapped sequence_type metafunction
* The assign_ungapped_sequence free function for containers over gapped alphabets.

### DIFF
--- a/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
+++ b/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
@@ -28,12 +28,68 @@
 #include <seqan3/range/view/to_char.hpp>
 #include <seqan3/std/ranges>
 
-namespace seqan3
+// ---------------------------------------------------------------------------------------------------------------------
+// unaligned_seq transformation trait
+// ---------------------------------------------------------------------------------------------------------------------
+
+namespace seqan3::detail
 {
 
-// -----------------------------------------------------------------------------
+//!\brief Helper function to deduce the unaligned sequence type from an aligned sequence container.
+template <template <typename...> typename container_type, typename seq_alph_t, typename ...rest_t>
+//!\cond
+    requires container_concept<container_type<gapped<seq_alph_t>, rest_t...>>
+//!\endcond
+constexpr auto remove_gap_from_value_type(container_type<gapped<seq_alph_t>, rest_t...>)
+    -> container_type<seq_alph_t, rest_t...>;
+
+//!\overload
+template <template <typename...> typename container_type,
+          template <typename...> typename allocator_type,
+          typename seq_alph_t, typename ...rest_t>
+//!\cond
+    requires container_concept<container_type<gapped<seq_alph_t>, allocator_type<gapped<seq_alph_t>>, rest_t...>>
+//!\endcond
+constexpr auto remove_gap_from_value_type(container_type<gapped<seq_alph_t>, allocator_type<gapped<seq_alph_t>>, rest_t...>)
+    -> container_type<seq_alph_t, allocator_type<seq_alph_t>, rest_t...>;
+
+//!\brief Default transformation trait that shall expose the unaligned sequence type of t when specialised.
+template <typename t>
+struct unaligned_seq
+{};
+
+//!\brief Exposes the unaligned sequence type given an aligned sequence container type.
+template <typename t>
+//!\cond
+    requires !requires { typename std::remove_reference_t<t>::unaligned_seq_type; } &&
+              requires { remove_gap_from_value_type(std::declval<t>()); }
+//!\endcond
+struct unaligned_seq<t>
+{
+    //!\brief The unaligned sequence type of t
+    using type = decltype(remove_gap_from_value_type(std::declval<t>()));
+};
+
+// customisation point for our gap decorators.
+//!\brief Exposes the unaligned sequence type if *t* exposes the type member `unaligned_seq_type`.
+template <typename t>
+//!\cond
+    requires requires { typename std::remove_reference_t<t>::unaligned_seq_type; }
+//!\endcond
+struct unaligned_seq<t>
+{
+    using type = typename std::remove_reference_t<t>::unaligned_seq_type; //!< The unaligned sequence type of t
+};
+
+//!\brief Helper type that delegates to seqan3::detail::unaligned_seq::type.
+template <typename t>
+using unaligned_seq_t = typename unaligned_seq<t>::type;
+
+} // namespace seqan3::detail
+
+// ---------------------------------------------------------------------------------------------------------------------
 // aligned_sequence_concept
-// -----------------------------------------------------------------------------
+// ---------------------------------------------------------------------------------------------------------------------
 
 /*!\interface seqan3::aligned_sequence_concept <>
  * \extends   std::ranges::ForwardRange
--- a/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
+++ b/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
@@ -58,41 +58,40 @@ namespace seqan3
  * \relates seqan3::aligned_sequence_concept
  * \{
  */
-/*!\fn      inline seq_type::iterator insert_gap(seq_type & seq, typename seq_type::const_iterator pos_it)
+/*!\fn      inline gapped_seq_type::iterator insert_gap(gapped_seq_type & seq, typename gapped_seq_type::const_iterator pos_it)
  * \brief   Insert a seqan3::gap into an aligned sequence.
  * \relates seqan3::aligned_sequence_concept
  *
- * \tparam        seq_type Type of the range to modify; must model
- *                         seqan3::aligned_sequence_concept.
- * \param[in,out] seq      The aligned sequence to modify.
- * \param[in]     pos_it   The iterator pointing to the position where to insert a gap.
+ * \tparam        gapped_seq_type Type of the range to modify; must model seqan3::aligned_sequence_concept.
+ * \param[in,out] seq             The aligned sequence to modify.
+ * \param[in]     pos_it          The iterator pointing to the position where to insert a gap.
  *
  * \details
  * \attention This is a concept requirement, not an actual function (however types
  *            modelling this concept will provide an implementation).
  */
-/*!\fn      inline seq_type::iterator insert_gap(seq_type & seq, typename seq_type::const_iterator pos_it, typename seq_type::size_type size)
+/*!\fn      inline gapped_seq_type::iterator insert_gap(gapped_seq_type & seq, typename gapped_seq_type::const_iterator
+ *          pos_it, typename gapped_seq_type::size_type size)
  * \brief   Insert multiple seqan3::gap into an aligned sequence.
  * \relates seqan3::aligned_sequence_concept
  *
- * \tparam        seq_type Type of the range to modify; must model
- *                         seqan3::aligned_sequence_concept.
- * \param[in,out] seq      The aligned sequence to modify.
- * \param[in]     pos_it   The iterator pointing to the position where to insert a gaps.
- * \param[in]     size     The number of gap symbols to insert (will result in a gap of length `size`).
+ * \tparam        gapped_seq_type Type of the range to modify; must model seqan3::aligned_sequence_concept.
+ * \param[in,out] seq             The aligned sequence to modify.
+ * \param[in]     pos_it          The iterator pointing to the position where to insert a gaps.
+ * \param[in]     size            The number of gap symbols to insert (will result in a gap of length `size`).
  *
  * \details
  * \attention This is a concept requirement, not an actual function (however types
  *            modelling this concept will provide an implementation).
  */
-/*!\fn      inline seq_type::iterator erase_gap(seq_type & seq, typename seq_type::const_iterator pos_it)
+/*!\fn      inline gapped_seq_type::iterator erase_gap(gapped_seq_type & seq, typename gapped_seq_type::const_iterator
+ *          pos_it)
  * \brief   Erase a seqan3::gap from an aligned sequence.
  * \relates seqan3::aligned_sequence_concept
  *
- * \tparam        seq_type Type of the range to modify; must model
- *                         seqan3::aligned_sequence_concept.
- * \param[in,out] seq      The aligned sequence to modify.
- * \param[in]     pos_it   The iterator pointing to the position where to erase a gap.
+ * \tparam        gapped_seq_type Type of the range to modify; must model seqan3::aligned_sequence_concept.
+ * \param[in,out] seq             The aligned sequence to modify.
+ * \param[in]     pos_it          The iterator pointing to the position where to erase a gap.
  *
  * \throws seqan3::gap_erase_failure if there is no seqan3::gap at \p pos_it.
  *
@@ -100,15 +99,15 @@ namespace seqan3
  * \attention This is a concept requirement, not an actual function (however types
  *            modelling this concept will provide an implementation).
  */
-/*!\fn      inline seq_type::iterator erase_gap(seq_type & seq, typename seq_type::const_iterator first, typename seq_type::const_iterator last)
+/*!\fn      inline gapped_seq_type::iterator erase_gap(gapped_seq_type & seq, typename gapped_seq_type::const_iterator
+ *          first, typename gapped_seq_type::const_iterator last)
  * \brief   Erase multiple seqan3::gap from an aligned sequence.
  * \relates seqan3::aligned_sequence_concept
  *
- * \tparam        seq_type Type of the range to modify; must model
- *                         seqan3::aligned_sequence_concept.
- * \param[in,out] seq      The aligned sequence to modify.
- * \param[in]     first    The iterator pointing to the position where to start erasing gaps.
- * \param[in]     last     The iterator pointing to the position where to stop erasing gaps.
+ * \tparam        gapped_seq_type Type of the range to modify; must model seqan3::aligned_sequence_concept.
+ * \param[in,out] seq             The aligned sequence to modify.
+ * \param[in]     first           The iterator pointing to the position where to start erasing gaps.
+ * \param[in]     last            The iterator pointing to the position where to stop erasing gaps.
  *
  * \throws seqan3::gap_erase_failure if one of the characters in [\p first, \p last) no seqan3::gap.
  *
@@ -137,19 +136,15 @@ SEQAN3_CONCEPT aligned_sequence_concept =
 // -----------------------------------------------------------------------------
 
 /*!\name Aligned sequence interface for containers
- * \brief Enables containers to model seqan3::aligned_sequence_concept if they
- * model seqan3::sequence_container_concept and have a reference type assignable
- * from seqan3::gap (e.g. std::vector<seqan3::gapped<seqan3::dna4>>).
+ * \brief Enables containers to model seqan3::aligned_sequence_concept if they model seqan3::sequence_container_concept
+ * and have a value type of the seqan3::gapped alphabet.
  * \{
  */
 /*!\brief An implementation of seqan3::aligned_sequence_concept::insert_gap for sequence containers.
- * \ingroup seqan3::aligned_sequence_concept
- * \tparam        seq_type Type of the container to modify; must model
- *                         seqan3::sequence_container_concept; the reference type
- *                         (seqan3::reference_t<seq_type>) must be assignable from
- *                         seqan3::gap.
- * \param[in,out] seq      The container to modify.
- * \param[in]     pos_it   The iterator pointing to the position where to insert a gap.
+ * \tparam        gapped_seq_type Type of the container to modify; must model seqan3::sequence_container_concept;
+ *                                The value type must be a seqan3::gapped alphabet.
+ * \param[in,out] seq             The gapped container to modify.
+ * \param[in]     pos_it          The iterator pointing to the position where to insert a gap.
  *
  *
  * \details
@@ -157,23 +152,21 @@ SEQAN3_CONCEPT aligned_sequence_concept =
  * This function delegates to the member function `insert(iterator, value)` of
  * the container.
  */
-template <sequence_container_concept seq_type>
+template <sequence_container_concept gapped_seq_type>
 //!\cond
-    requires WeaklyAssignable<reference_t<seq_type>, gap const &>
+    requires detail::is_gapped_alphabet<value_type_t<gapped_seq_type>>
 //!\endcond
-inline typename seq_type::iterator insert_gap(seq_type & seq, typename seq_type::const_iterator pos_it)
+inline typename gapped_seq_type::iterator insert_gap(gapped_seq_type & seq,
+                                                     typename gapped_seq_type::const_iterator pos_it)
 {
-    return seq.insert(pos_it, value_type_t<seq_type>{gap{}});
+    return seq.insert(pos_it, value_type_t<gapped_seq_type>{gap{}});
 }
 
 /*!\brief An implementation of seqan3::aligned_sequence_concept::insert_gap for sequence containers.
- * \ingroup seqan3::aligned_sequence_concept
- * \tparam        seq_type Type of the container to modify; must model
- *                         seqan3::sequence_container_concept; the reference type
- *                         (seqan3::reference_t<seq_type>) must be assignable from
- *                         seqan3::gap.
- * \param[in,out] seq      The container to modify.
- * \param[in]     pos_it   The iterator pointing to the position where to insert gaps.
+ * \tparam        gapped_seq_type Type of the container to modify; must model seqan3::sequence_container_concept;
+ *                                The value type must be a seqan3::gapped alphabet.
+ * \param[in,out] seq             The gapped container to modify.
+ * \param[in]     pos_it          The iterator pointing to the position where to insert gaps.
  * \param[in]     size     The number of gap symbols to insert (will result in a gap of length `size`).
  *
  * \details
@@ -181,25 +174,22 @@ inline typename seq_type::iterator insert_gap(seq_type & seq, typename seq_type:
  * This function delegates to the member function `insert(iterator, `size`, `value`)`
  * of the container.
  */
-template <sequence_container_concept seq_type>
+template <sequence_container_concept gapped_seq_type>
 //!\cond
-    requires WeaklyAssignable<reference_t<seq_type>, gap const &>
+    requires detail::is_gapped_alphabet<value_type_t<gapped_seq_type>>
 //!\endcond
-inline typename seq_type::iterator insert_gap(seq_type & seq,
-                                              typename seq_type::const_iterator pos_it,
-                                              typename seq_type::size_type size)
+inline typename gapped_seq_type::iterator insert_gap(gapped_seq_type & seq,
+                                                     typename gapped_seq_type::const_iterator pos_it,
+                                                     typename gapped_seq_type::size_type size)
 {
-    return seq.insert(pos_it, size, value_type_t<seq_type>{gap{}});
+    return seq.insert(pos_it, size, value_type_t<gapped_seq_type>{gap{}});
 }
 
 /*!\brief An implementation of seqan3::aligned_sequence_concept::erase_gap for sequence containers.
- * \ingroup seqan3::aligned_sequence_concept
- * \tparam        seq_type Type of the container to modify; must model
- *                         seqan3::sequence_container_concept; the reference type
- *                         (seqan3::reference_t<seq_type>) must be assignable from
- *                         seqan3::gap.
- * \param[in,out] seq      The container to modify.
- * \param[in]     pos_it   The iterator pointing to the position where to erase a gap.
+ * \tparam        gapped_seq_type Type of the container to modify; must model seqan3::sequence_container_concept;
+ *                                The value type must be a seqan3::gapped alphabet.
+ * \param[in,out] seq             The gapped container to modify.
+ * \param[in]     pos_it          The iterator pointing to the position where to erase a gap.
  *
  * \throws seqan3::gap_erase_failure if there is no seqan3::gap at \p pos_it.
  *
@@ -209,11 +199,12 @@ inline typename seq_type::iterator insert_gap(seq_type & seq,
  * container. Before delegating, the function checks if the position pointed to
  * by \p pos_it is an actual seqan3::gap and throws an exception if not.
  */
-template <sequence_container_concept seq_type>
+template <sequence_container_concept gapped_seq_type>
 //!\cond
-    requires WeaklyAssignable<reference_t<seq_type>, gap const &>
+    requires detail::is_gapped_alphabet<value_type_t<gapped_seq_type>>
 //!\endcond
-inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::const_iterator pos_it)
+inline typename gapped_seq_type::iterator erase_gap(gapped_seq_type & seq,
+                                                    typename gapped_seq_type::const_iterator pos_it)
 {
     if (*pos_it != gap{}) // [[unlikely]]
         throw gap_erase_failure("The position to be erased does not contain a gap.");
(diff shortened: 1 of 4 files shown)


include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -242,20 +285,45 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+ *                                  the value type must be a seqan3::gapped alphabet.
+ * \tparam        ungapped_seq_type Type of the container to assign from; must model std::ranges::InputRange;
+ * \param[in,out] gapped            The gapped sequence container to assign to.
+ * \param[in,out] ungapped          The ungapped sequence container to assign from.
+ *
+ * \relates seqan3::aligned_sequence_concept
+ *
+ * \details
+ *
+ * This function clears the content of the `gapped` container and reassign the content of the `ungapped` container by
```
eseiler:
```suggestion
 * This function clears the content of the `gapped` container and reassigns the content of the `ungapped` container by
```

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -30,6 +30,44 @@
...
+ *
+ * This specializations works for sequence containers of gapped alphabet and returns a std::vector over the ungapped
+ * alphabet. For example, `ungapped_sequence_type<std::vector<gapped<dna4>>>::type` will be of the type
+ * `std::vector<dna>`, but `ungapped_sequence_type<bitcompressed_vector<gapped<aa20>>>::type` too.
+ */
+template <sequence_container_concept gapped_seq_type>
+//!\cond
+    requires detail::is_gapped_alphabet<value_type_t<gapped_seq_type>>
+//!\endcond
+struct ungapped_sequence_type<gapped_seq_type>
```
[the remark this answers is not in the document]
smehringer: well I guess you can just use `value_type<ungapped_sequence>`

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -30,6 +30,47 @@
...
+};
+
+//!\brief Helper typedef for the metafunction seqan3::ungapped_sequence_type
+template <typename t>
+using ungapped_sequence_type_t = typename ungapped_sequence_type<t>::type;
+
+/*!\brief Transformation trait metafunction that returns the ungapped sequence type.
+ * \ingroup aligned_sequence
+ * \tparam gapped_seq_type The gapped sequence container; must model seqan3::sequence_container_concept and its value
+ *                         type must be a seqan3::gapped alphabet.
```
h-2: Hm, this is really not ideal. If it always returns a vector, it's not a generic metafunction and I would maybe just move it to detail and clearly state that it is a helper for verifying the concept. It's not used anywhere else, right?

smehringer: I agree. It is not used anywhere else yet.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -242,20 +249,45 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+ */
+template <sequence_container_concept gapped_seq_type, std::ranges::InputRange ungapped_sequence_type>
+//!\cond
+    requires detail::is_gapped_alphabet<value_type_t<gapped_seq_type>> &&
+             weakly_assignable_concept<reference_t<gapped_seq_type>, reference_t<ungapped_sequence_type>>
+//!\endcond
+void assign_ungapped_sequence(gapped_seq_type & gapped, ungapped_sequence_type && ungapped)
+{
+    gapped.resize(ungapped.size());
+    std::copy(ungapped.begin(), ungapped.end(), gapped.begin());
```
h-2: `begin(gapped)` as gapped need not have a member begin

include/seqan3/alphabet/gap/gapped.hpp
```diff
@@ -41,4 +41,46 @@ template <typename alphabet_t>
...
+struct ungapped<gapped<t>>
+{
+    using type = t;
+};
+
+//!\brief Helper typedef that delegates to the type member of seqan3::ungapped.
+//!       (e.g. seqan3::ungapped_t<seqan3::gapped<seqan3::dna4>> == seqan3::dna4).
+//!\ingroup gap
+template<typename t>
+using ungapped_t = typename ungapped<t>::type;
```
h-2: It's better to also constrain this with a check that the type has `::type`.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -242,20 +249,45 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+ * using std::copy.
+ */
+template <sequence_container_concept gapped_seq_type, std::ranges::InputRange ungapped_sequence_type>
+//!\cond
+    requires detail::is_gapped_alphabet<value_type_t<gapped_seq_type>> &&
+             weakly_assignable_concept<reference_t<gapped_seq_type>, reference_t<ungapped_sequence_type>>
+//!\endcond
+void assign_ungapped_sequence(gapped_seq_type & gapped, ungapped_sequence_type && ungapped)
+{
+    gapped.resize(ungapped.size());
```
rrahn:
```suggestion
    gapped.resize(std::distance(seqan3::begin(ungapped), seqan3::end(ungapped));
```

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -242,20 +249,45 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+    std::copy(begin(ungapped), end(ungapped), begin(gapped));
```
rrahn:
```suggestion
    std::copy(seqan3::begin(ungapped), seqan3::end(ungapped), seqan3::begin(gapped));
```

rrahn: You can't rely on unqualified lookup here.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
rrahn: But also note, that this can't be an InputRange anymore, but must be at least model ForwardRange.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -242,20 +249,45 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+ *
+ * This function clears the content of the `gapped` container and reassigns the content of the `ungapped` container by
+ * using std::copy.
+ */
+template <sequence_container_concept gapped_seq_type, std::ranges::InputRange ungapped_sequence_type>
+//!\cond
+    requires detail::is_gapped_alphabet<value_type_t<gapped_seq_type>> &&
+             weakly_assignable_concept<reference_t<gapped_seq_type>, reference_t<ungapped_sequence_type>>
+//!\endcond
+void assign_ungapped_sequence(gapped_seq_type & gapped, ungapped_sequence_type && ungapped)
```
rrahn: Now, here is the issue that I see. This function should only work for owning ranges. But the aligned_sequence_decorator does not own the data, hence it should not do a deep assign to the underlying range and therefore not offering this function. At least I am rather against it. So please explain what the use case of this function is.

include/seqan3/alphabet/gap/gapped.hpp
```diff
@@ -41,4 +41,54 @@ template <typename alphabet_t>
...
 using gapped = union_composition<alphabet_t, gap>;
 
+// ---------------------------------------------------------------------------------------------------------------------
+// ungapped metafunction
+// ---------------------------------------------------------------------------------------------------------------------
+
+//!\brief Transformation trait that exposes the underlying (ungapped) alphabet type [**default**: no type member].
+//!\ingroup gap
+template <typename t>
+struct ungapped {};
```
rrahn: braces on a new line

rrahn: I am a little confused by the name. There is  a gapped alphabet and a ungapped type trait? 
I would rather follow the STL here and call it `remove_gap`.

smehringer:
> I need more explanation about the use cases for assign_ungapped_sequence function.

I will need this in the `alignment_file_input` where I need a convenient way to assign the read/reference sequence to the alignment pair as they can differ in their type. Consider that you want only the alignment  from the SAM file, then your alignment pair needs to hold two gapped sequences containers (e.g. `std::vector<gapped<dna4>>`). On the other hand you might have the reference sequences in memory and just want the alignment pair to hold a subrange view over part of the reference.
So now when reading in the sequence how do I assign the sequence to my alignment pair.. For  `std::vector<gapped<dna4>>` I need to copy over the elements one by one but for a decorated subrange I need to reassign the underlying sequence.

h-2:
> Ok, so in general what you want is a reset function. But I would call it plainly reset.

Reset implies that something has an inherent default state that you reset it to. I have never seen `a.reset(b);` only `a.reset()`. But you actually want to change the state of the object to something new. 

We also cannot get rid of the different semantics for std::vector<gapped<>> (assign) and a decorator ("re-apply and reset") so every word will sound better for one case and worse for the other. Also we can actually do the following which has very similar semantics to our use case:
```cpp
std::string s{"foo"};
std::string t{"bar"};

auto v = s | view::take(2); // "fo"
v = view::all(t); // "ba"
v = view::all(s); // "fo"
```
→ You can change the underlying range of a view via assignment. And I suggested that the decorators also allow this. The "assign" function in addition to `operator=` is only necessary, because we cannot overload `operator=` for std::vector.

rrahn:
https://en.cppreference.com/w/cpp/memory/unique_ptr/reset

h-2:
Can you please reply to my other comment as well? We already have "assignment" as a way for doing this operation on views, in which way to you think it is different for decorators?

h-2:
@rrahn Please discuss this with @smehringer . @smehringer and I had a long discussions on all the pros and cons and we agreed on a design for decorators that is very close to views and I think we also discussed this briefly a strategy meeting. 

* Decorators always store the underlying range wrapped in a view, they do not store pointers anymore. They can be constructed from copies of views or lvalue-references to non-views (which will be wrapped in `view::all`).
* Copy-assignment replaces the underlying view with that of the RHS and it also copies the gap information. Assigning the underlying range type (wrapped in a view or not) also replaces the underlying view with the RHS and clears the gap information.
* Equality comparison can easily be implemented in the model so that `a = b` implies `a == b` after that.
* Nothing ever changes the contents of the underlying range.

> I mean, you could try to make the decorator only take a new ref_view to the assigned vector, 

It could take the underlying type by lvalue reference as well and just view-wrap it itself.

> but I think this only would work with a type_erased any_view stored in the decorator or some other major restrictions to the decorated type.

Why? It's the same as for views. If you construct over vector<int> you can only assign vector<int> and not array<int>, but there is no need for type erasure.

If you disagree with this design in general, we need to re-evaluate it at the next strategy meeting, but I am against changing it ad-hoc in the discussion of the PR.

rrahn:
> * Assigning the underlying range type (wrapped in a view or not) also replaces the underlying view with the RHS and clears the gap information.

That's the part that I have a problem with from the semantics point of view. Assign means I copy assign stuff. For the vector we say, everything that is an InputRange and whose value types are implicit convertible can be used. For the decorator this won't hold anymore. The requirements are more strict. So for the decorator, I cannot create from a vector and later assign from an array, right?
Not without type-erasure or with writing into the underlying resource. The types must be the same. So the function would be assign_from_underlying/host whatever and must be restricted for the vector as well.
What am I not seeing here? 😕 

> If you disagree with this design in general, we need to re-evaluate it at the next strategy meeting, but I am against changing it ad-hoc in the discussion of the PR.

In general, I just think that this assignment operation should not be part of the concept. But let us discuss this next strategy meeting.

h-2:
> That's the part that I have a problem with from the semantics point of view. Assign means I copy assign stuff.

Yes and no. It depends on the type of type. For `int * p` the `operator=` just changes where p points to.

> For the vector we say, everything that is an InputRange and whose value types are implicit convertible can be used. 

That's specific for std::vector, it doesn't hold for e.g std::array and it's not part of any concepts.

> So for the decorator, I cannot create from a vector and later assign from an array, right?

Correct. The same way you cannot create an array and later assign from a vector.

> So the function would be assign_from_underlying/host whatever and must be restricted for the vector as well. What am I not seeing here?

You mean it should only be possible to assign() `std::vector<dna4>` to `std::vector<gapped<dna4>>` and not `std::array<dna4>` to `std::vector<gapped<dna4>>`. I am totally fine with this restriction and it was the original design. However @smehringer and I found out that it's not trivial to implement this restriction. You can do some magic with template templates if you assume that the value_type is always the first template parameter, but it's not generic. We don't even know if a container will be a template, it's not required anywhere.
Therefore we decided to just check for Range and the value type.

> In general, I just think that this assignment operation should not be part of the concept. But let us discuss this next strategy meeting.

Unfortunately it is required from my POV, but we can discuss it of course.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
smehringer: why?

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
rrahn: you should also make a `constexpr if ` and check if the range is a sized range. In this case you can use seqan3::size instead. This will be more efficient for non randomaccess ranges which are sized.

h-2: std::distance() is in O(1) for sized ranges already!

h-2:
@rrahn can you mark your comments as resolved that were resolved in the last strategy meeting?

rrahn:
> std::distance() is in O(1) for sized ranges already!

@h-2 ah ok, I looked it up in the standard, you are right.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -58,64 +110,77 @@ namespace seqan3
...
  * \attention This is a concept requirement, not an actual function (however types
  *            modelling this concept will provide an implementation).
  */
+/*!\fn      void assign_unaligned_sequence(aligned_seq_t & aligned_seq, unaligned_sequence_type && unaligned_seq)
+ *!\brief Assign an ungapped sequence to a gapped sequence.
+ * \relates seqan3::aligned_sequence_concept
+ *
+ * \tparam        aligned_seq_t     Type of the container to reassign; must model seqan3::sequence_container_concept;
+ *                                  the value type must be a seqan3::gapped alphabet.
+ * \tparam        unaligned_seq_t   Type of the container to assign from; must model std::ranges::InputRange;
```
rrahn:
```suggestion
 * \tparam        unaligned_seq_t   Type of the container to assign from; must model std::ranges::ForwardRange;
```
?

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -28,12 +28,69 @@
...
+
+//!\brief Default transformation trait that shall expose the unaligned sequence type of t when specialised.
+template <typename t>
+struct unaligned_seq
+{};
+
+//!\brief Exposes the unaligned sequence type given an aligned sequence container type.
+template <typename t>
+//!\cond
+    requires container_concept<t>
```
rrahn: better be paranoid and remove_reference here as well.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -28,12 +28,69 @@
...
+    requires requires { typename std::remove_reference_t<t>::unaligned_seq_type; }
+//!\endcond
+struct unaligned_seq<t>
 {
+    using type = typename std::remove_reference_t<t>::unaligned_seq_type; //!< The unaligned sequence type of t
+};
+
+//!\brief Helper typdef that delegates to seqan3::detail::unaligned_seq::type.
+template <typename t>
+using unaligned_seq_t = typename unaligned_seq<std::remove_reference_t<t>>::type;
```
rrahn: you actually don't need remove_ref here, as the implementations should guard it

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
rrahn: and it should not have a unaligned_seq_type member type. Otherwise the requirements might lead to an ambiguous overload if there is a container with this member type.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
rrahn: I think this statement is to generic. we need to formulate the requirements more clearly, that the type is expected to be the same as the aligned_seq_t without the gap information, I think. At least this is what the concept checks.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -211,25 +271,24 @@ inline typename seq_type::iterator insert_gap(seq_type & seq,
...
  */
-template <sequence_container_concept seq_type>
+template <sequence_container_concept aligned_seq_t>
 //!\cond
-    requires weakly_assignable_concept<reference_t<seq_type>, gap const &>
+    requires detail::is_gapped_alphabet<value_type_t<aligned_seq_t>>
 //!\endcond
-inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::const_iterator pos_it)
+inline typename aligned_seq_t::iterator erase_gap(aligned_seq_t & aligned_seq,
+                                                typename aligned_seq_t::const_iterator pos_it)
```
rrahn: indentation

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -242,19 +301,44 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+ * using std::copy.
+ */
+template <sequence_container_concept aligned_seq_t, std::ranges::ForwardRange unaligned_sequence_type>
+//!\cond
+    requires detail::is_gapped_alphabet<value_type_t<aligned_seq_t>> &&
+             weakly_assignable_concept<reference_t<aligned_seq_t>, reference_t<unaligned_sequence_type>>
+//!\endcond
+void assign_unaligned(aligned_seq_t & aligned_seq, unaligned_sequence_type && unaligned_seq)
+{
+    aligned_seq.resize(std::ranges::size(unaligned_seq));
```
rrahn: Ok, if we require InputRange, you need two approaches. 
First, if I correctly understood the standard then size works only if the range supports it, otherwise it is ill-formed. 
Hence you need to use distance. But for InputRanges this won't work so you need a case distinction and for input ranges you need to use a push_back iterator.
Or am I missing something?

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -28,12 +28,69 @@
...
+{};
+
+//!\brief Exposes the unaligned sequence type given an aligned sequence container type.
+template <typename t>
+//!\cond
+    requires container_concept<t>
+//!\endcond
+struct unaligned_seq<t>
+{
+    using type = decltype(get_ungapped_type(std::declval<t>())); //!< The unaligned sequence type of t
```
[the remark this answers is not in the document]
smehringer: The function parameter type is specialized for the gapped alphabet type so there is no need to require it?
`constexpr auto remove_gap_from_value_type(container_type<gapped<seq_alph_t>, rest_t...>)`

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
smehringer: a gap_decorator will never model the container_concept.. but let's be paranoid

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
smehringer: I would propose to require ForwardRange for now and add a case distinction if a use case for forward ranges shows up.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
rrahn: but this does not mean that there could be an l-value reference to a container, does it? So we assume that container_concept strips references, but we should be paranoid and not assume this.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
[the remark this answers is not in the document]
rrahn: I think the compiler error will be not as easy to read, as he will complain about not finding viable overloads. 
But I am not sure. Could you test it maybe?

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
rrahn: ok

smehringer:
@rrahn do not merge yet, because #681 is about to be merged and this will need to be rebased.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -242,19 +308,55 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+//!\cond
+    requires detail::is_gapped_alphabet<value_type_t<aligned_seq_t>> &&
+             weakly_assignable_concept<reference_t<aligned_seq_t>, reference_t<unaligned_sequence_type>>
+//!\endcond
+void assign_unaligned(aligned_seq_t & aligned_seq, unaligned_sequence_type && unaligned_seq)
+{
+    aligned_seq_t tmp;
+    tmp.resize(std::ranges::size(unaligned_seq));
+    std::copy(begin(unaligned_seq), end(unaligned_seq), begin(tmp));
+    std::swap(aligned_seq, tmp);
```
rrahn: Please use unqualified lookup and add `using std::swap` at the beginning of the function body

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -242,19 +308,55 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+ */
+template <sequence_container_concept aligned_seq_t, std::ranges::ForwardRange unaligned_sequence_type>
+//!\cond
+    requires detail::is_gapped_alphabet<value_type_t<aligned_seq_t>> &&
+             weakly_assignable_concept<reference_t<aligned_seq_t>, reference_t<unaligned_sequence_type>>
+//!\endcond
+void assign_unaligned(aligned_seq_t & aligned_seq, unaligned_sequence_type && unaligned_seq)
+{
+    aligned_seq_t tmp;
+    tmp.resize(std::ranges::size(unaligned_seq));
```
rrahn: According to [this](http://eel.is/c++draft/range.prim.size) I really think that you can't use `size` without requiring a `SizedRange`, can you? 
You need to use std::ranges::distance(unaligned_seq), which will resolve to size if it can otherwise call distance(end, begin), while doing the most performant operation for this (see [here](http://eel.is/c++draft/range.iter.op.distance))

test/unit/alignment/aligned_sequence_test_template.hpp
```diff
@@ -26,6 +26,21 @@ TYPED_TEST_P(aligned_sequence, fulfills_concept)
...
+    TypeParam aligned{};
+
+    assign_unaligned(aligned, unaligned);
+
+    EXPECT_EQ(aligned.size(), unaligned.size());
+    EXPECT_EQ(*aligned.begin(), *unaligned.begin());
+    EXPECT_EQ(*(aligned.begin() + 1), *(unaligned.begin() + 1));
+}
+
+
```
rrahn: empty line

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
rrahn: Well during unqualified lookup the compiler will only look in the class (friend member function) and the namespace of the class for a possible candidate, but not in other namespaces. So if you rely on std::begin to be picked up it won't unless you declare `using std::begin` before in the function body. Then also the namespace std:: will be searched for a possible candidate. To remedy this additional `using ...` declaration the ranges library introduced function objects as customisation points instead of free functions. These "functors" will do the proper lookup for you, but you need to use the qualified name (in our case our alias is fine). Thus, begin for ranges from other namespaces (not seqan3) will be resolved correctly.

include/seqan3/alphabet/gap/gapped.hpp
rrahn: @smehringer This is gone right?

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
h-2: We should add a point "frequently used stuff from std::ranges::" to our next strategy meeting. And I would also propose making at least the free functions for our alphabets be "Niebloids" (function objects that are not found by ADL, but that perform ADL look-up if not delegating to members).

include/seqan3/alphabet/gap/gapped.hpp
smehringer: yes

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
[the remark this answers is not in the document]
smehringer: <details> <summary>Error message like it is now: (click me) </summary> 

```
In file included from /home/svenja/AGABI/seqan3/test/unit/alignment/../alignment/aligned_sequence_test_template.hpp:12:0,
                 from /home/svenja/AGABI/seqan3/test/unit/alignment/aligned_sequence_test.cpp:14:
/home/svenja/AGABI/seqan3/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp: In instantiation of ‘struct seqan3::detail::unaligned_seq<std::vector<seqan3::dna4> >’:
/home/svenja/AGABI/seqan3/test/unit/alignment/aligned_sequence_test.cpp:43:58:   required from here
/home/svenja/AGABI/seqan3/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp:71:53: error: no matching function for call to ‘remove_gap_from_value_type(std::vector<seqan3::dna4>)’
     using type = decltype(remove_gap_from_value_type(std::declval<t>()));
                           ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~
/home/svenja/AGABI/seqan3/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp:43:16: note: candidate: template<template<class ...> class container_type, class seq_alph_t, class ... rest_t>  requires  container_concept<container_type<seqan3::union_composition<seq_alph_t, seqan3::gap>, rest_t ...> > constexpr container_type<seq_alph_t, rest_t ...> seqan3::detail::remove_gap_from_value_type(container_type<seqan3::union_composition<seq_alph_t, seqan3::gap>, rest_t ...>)
 constexpr auto remove_gap_from_value_type(container_type<gapped<seq_alph_t>, rest_t...>)
                ^~~~~~~~~~~~~~~~~~~~~~~~~~
/home/svenja/AGABI/seqan3/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp:43:16: note:   template argument deduction/substitution failed:
/home/svenja/AGABI/seqan3/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp:71:53: note:   mismatched types ‘seqan3::union_composition<seq_alph_t, seqan3::gap>’ and ‘seqan3::dna4’
     using type = decltype(remove_gap_from_value_type(std::declval<t>()));
```

</details>

<details> <summary>Error message with an additional requires "is_gapped_alphabet": (click me) </summary> 

```
/home/svenja/AGABI/seqan3/test/unit/alignment/aligned_sequence_test.cpp: In member function ‘virtual void hi_hi_Test::TestBody()’:
/home/svenja/AGABI/seqan3/test/unit/alignment/aligned_sequence_test.cpp:43:60: error: ‘type’ in ‘struct seqan3::detail::unaligned_seq<std::vector<seqan3::dna4> >’ does not name a type
     using type = detail::unaligned_seq<std::vector<dna4>>::type;
```

I am not sure but the first is maybe even more descriptive as it tells you that dna4 is no gapped alphabet while the second just fails because the chosen struct (chosen by overload resolution when the requires fails) is the default one without a type member.

@rrahn what do you think?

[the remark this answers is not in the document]
rrahn: Mhmm, I thought that the requires clause will emit a constraint not satisfied compiler error pointing to the is_gapped_alphabet expression. Does this only works with concepts?

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -239,19 +302,56 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+template <sequence_container_concept aligned_seq_t, std::ranges::ForwardRange unaligned_sequence_type>
+//!\cond
+    requires detail::is_gapped_alphabet<value_type_t<aligned_seq_t>> &&
+             WeaklyAssignable<reference_t<aligned_seq_t>, reference_t<unaligned_sequence_type>>
+//!\endcond
+void assign_unaligned(aligned_seq_t & aligned_seq, unaligned_sequence_type && unaligned_seq)
+{
+    using std::swap;
+    aligned_seq_t tmp;
+    tmp.resize(std::ranges::distance(begin(unaligned_seq), end(unaligned_seq)));
```
rrahn:
```suggestion
    tmp.resize(std::ranges::distance(unaligned_seq));
```

rrahn: This allows to use size if it is a sized range.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -239,19 +302,56 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+//!\cond
+    requires detail::is_gapped_alphabet<value_type_t<aligned_seq_t>> &&
+             WeaklyAssignable<reference_t<aligned_seq_t>, reference_t<unaligned_sequence_type>>
+//!\endcond
+void assign_unaligned(aligned_seq_t & aligned_seq, unaligned_sequence_type && unaligned_seq)
+{
+    using std::swap;
+    aligned_seq_t tmp;
+    tmp.resize(std::ranges::distance(unaligned_seq));
+    std::copy(begin(unaligned_seq), end(unaligned_seq), begin(tmp));
```
rrahn:
```suggestion
    std::copy(std::ranges::begin(unaligned_seq), std::ranges::end(unaligned_seq), std::ranges::begin(tmp));
```

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -239,19 +302,56 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
...
+ *
+ * Strong exception guarantee.
+ *
+ */
+template <sequence_container_concept aligned_seq_t, std::ranges::ForwardRange unaligned_sequence_type>
+//!\cond
+    requires detail::is_gapped_alphabet<value_type_t<aligned_seq_t>> &&
+             WeaklyAssignable<reference_t<aligned_seq_t>, reference_t<unaligned_sequence_type>>
+//!\endcond
+void assign_unaligned(aligned_seq_t & aligned_seq, unaligned_sequence_type && unaligned_seq)
```
rrahn:
```suggestion
inline void assign_unaligned(aligned_seq_t & aligned_seq, unaligned_sequence_type && unaligned_seq)
```

smehringer:
@rrahn  Did I address all of your issues?

PR #681, opened 2019-01-17
The Anchor Set Gap decorator

smehringer:
This supersedes #726 

This is to take some work load of marie.

### DIFF
--- a/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
+++ b/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
@@ -143,6 +143,7 @@ SEQAN3_CONCEPT aligned_sequence_concept =
  * \{
  */
 /*!\brief An implementation of seqan3::aligned_sequence_concept::insert_gap for sequence containers.
+ * \ingroup seqan3::aligned_sequence_concept
  * \tparam        seq_type Type of the container to modify; must model
  *                         seqan3::sequence_container_concept; the reference type
  *                         (seqan3::reference_t<seq_type>) must be assignable from
@@ -150,7 +151,6 @@ SEQAN3_CONCEPT aligned_sequence_concept =
  * \param[in,out] seq      The container to modify.
  * \param[in]     pos_it   The iterator pointing to the position where to insert a gap.
  *
- * \relates seqan3::aligned_sequence_concept
  *
  * \details
  *
@@ -167,6 +167,7 @@ inline typename seq_type::iterator insert_gap(seq_type & seq, typename seq_type:
 }
 
 /*!\brief An implementation of seqan3::aligned_sequence_concept::insert_gap for sequence containers.
+ * \ingroup seqan3::aligned_sequence_concept
  * \tparam        seq_type Type of the container to modify; must model
  *                         seqan3::sequence_container_concept; the reference type
  *                         (seqan3::reference_t<seq_type>) must be assignable from
@@ -174,8 +175,6 @@ inline typename seq_type::iterator insert_gap(seq_type & seq, typename seq_type:
  * \param[in,out] seq      The container to modify.
  * \param[in]     pos_it   The iterator pointing to the position where to insert gaps.
  * \param[in]     size     The number of gap symbols to insert (will result in a gap of length `size`).
- *
- * \relates seqan3::aligned_sequence_concept
  *
  * \details
  *
@@ -194,6 +193,7 @@ inline typename seq_type::iterator insert_gap(seq_type & seq,
 }
 
 /*!\brief An implementation of seqan3::aligned_sequence_concept::erase_gap for sequence containers.
+ * \ingroup seqan3::aligned_sequence_concept
  * \tparam        seq_type Type of the container to modify; must model
  *                         seqan3::sequence_container_concept; the reference type
  *                         (seqan3::reference_t<seq_type>) must be assignable from
@@ -202,8 +202,6 @@ inline typename seq_type::iterator insert_gap(seq_type & seq,
  * \param[in]     pos_it   The iterator pointing to the position where to erase a gap.
  *
  * \throws seqan3::gap_erase_failure if there is no seqan3::gap at \p pos_it.
- *
- * \relates seqan3::aligned_sequence_concept
  *
  * \details
  *
@@ -224,6 +222,7 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
 }
 
 /*!\brief An implementation of seqan3::aligned_sequence_concept::erase_gap for sequence containers.
+ * \ingroup seqan3::aligned_sequence_concept
  * \tparam        seq_type Type of the container to modify; must model
  *                         seqan3::sequence_container_concept; the reference type
  *                         (seqan3::reference_t<seq_type>) must be assignable from
@@ -233,8 +232,6 @@ inline typename seq_type::iterator erase_gap(seq_type & seq, typename seq_type::
  * \param[in]     last     The iterator pointing to the position where to stop erasing gaps.
  *
  * \throws seqan3::gap_erase_failure if one of the characters in [\p first, \p last) no seqan3::gap.
- *
- * \relates seqan3::aligned_sequence_concept
  *
  * \details
  *
@@ -255,6 +252,87 @@ inline typename seq_type::iterator erase_gap(seq_type & seq,
             throw gap_erase_failure("The range to be erased contains at least one non-gap character.");
 
     return seq.erase(first, last);
+}
+//!\}
+
+/*!\name Aligned sequence interface for ranges that have the corresponding member functions.
+ * \brief Enables ranges to model seqan3::aligned_sequence_concept if they have the member functions
+ *        insert_gap() and erase_gap().
+ * \{
+ */
+/*!\brief An implementation of seqan3::aligned_sequence_concept::insert_gap for ranges with the corresponding
+ *        member function insert_gap(it, size).
+ * \ingroup seqan3::aligned_sequence_concept
+ * \tparam range_type   Type of the range to modify; must have an insert_gap(it, size) member function.
+ * \param[in,out] rng   The range to modify.
+ * \param[in]     it    The iterator pointing to the position where to start inserting gaps.
+ * \param[in]     size  The number of gaps to insert as an optional argument, default is 1.
+ *
+ * \details
+ *
+ * This function delegates to the member function `insert(iterator, size)` of the range.
+ */
+template <typename range_type>
+//!\cond
+    requires requires (range_type v)
+        {
+            v.insert_gap(typename range_type::iterator{});
+            v.insert_gap(typename range_type::iterator{}, typename range_type::size_type{});
+        }
+//!\endcond
+typename range_type::iterator insert_gap(range_type & rng,
+                                         typename range_type::iterator const it,
+                                         typename range_type::size_type const size = 1)
+{
+    return rng.insert_gap(it, size);
+}
+
+/*!\brief An implementation of seqan3::aligned_sequence_concept::erase_gap for ranges with the corresponding
+ *        member function erase_gap(it).
+ * \ingroup seqan3::aligned_sequence_concept
+ * \tparam range_type   Type of the range to modify; must have an erase_gap(it) member function.
+ * \param[in,out] rng   The range to modify.
+ * \param[in] it        The iterator pointing to the position where to erase one gap.
+ *
+ * \details
+ *
+ * This function delegates to the member function `erase(it)` of
+ * the range.
+ */
+template <typename range_type>
+//!\cond
+    requires requires (range_type v) { v.erase_gap(typename range_type::iterator{}); }
+//!\endcond
+typename range_type::iterator erase_gap(range_type & rng,
+                                        typename range_type::iterator const it)
+{
+    return rng.erase_gap(it);
+}
+
+/*!\brief An implementation of seqan3::aligned_sequence_concept::erase_gap for ranges with the corresponding
+ *        member function erase_gap(first, last).
+ * \ingroup seqan3::aligned_sequence_concept
+ * \tparam range_type   Type of the range to modify; must have an erase_gap(first, last) member function.
+ * \param[in,out] rng   The range to modify.
+ * \param[in] first     The iterator pointing to the position where to start erasing gaps.
+ * \param[in] last      The iterator pointing to the position where to stop erasing gaps.
+ *
+ * \throws seqan3::gap_erase_failure if one of the characters in [\p first, \p last) is no seqan3::gap.
+ *
+ *
+ * \details
+ *
+ * This function delegates to the member function `erase(first, last)` of the range.
+ */
+template <typename range_type>
+//!\cond
+    requires requires (range_type v) { v.erase_gap(typename range_type::iterator{}, typename range_type::iterator{}); }
+//!\endcond
+typename range_type::iterator erase_gap(range_type & rng,
+                                        typename range_type::iterator const first,
+                                        typename range_type::iterator const last)
+{
+    return rng.erase_gap(first, last);
 }
 //!\}
 
--- a/include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
+++ b/include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
@@ -1,0 +1,505 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides seqan3::gap_decorator_anchor_set.
+ * \author Marie Hoffmann <marie.hoffmann AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <set>
+#include <tuple>
+#include <type_traits>
+
+#include <range/v3/all.hpp>
+
+#include <seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp>
+#include <seqan3/alphabet/concept.hpp>
+#include <seqan3/alphabet/gap/gap.hpp>
+#include <seqan3/alphabet/gap/gapped.hpp>
+#include <seqan3/range/container/concept.hpp>
+#include <seqan3/range/detail/random_access_iterator.hpp>
+
+namespace seqan3
+{
+
+/*!\brief A gap decorator allows the annotation of sequences with gap symbols
+ * leaving the underlying sequence unmodified.
+ * \tparam inner_type The type of alphabet sequences that will be stored.
+ * \implements seqan3::aligned_sequence_concept
+ * \ingroup decorator
+ *
+ * This class may be used whenever you want to store or compute an alignment.
+ * The underlying (ungapped sequence) remains unmodified, and is augmented by
+ * gap information. Iterating over a gap_decorator sequences behaves as if you
+ * had a rolled out, aligned sequence with alphabet and gap symbols.
+ *
+ * \details
+ * The gap_decorator_anchor_set is a modified anchor list approach -- instead
+ * of storing tuples of anchor positions and gap lengths relative to the
+ * underlying sequence position, the anchor addresses are virtual and gap
+ * lengths accumulated from left to right, i.e. an anchor gap stores at the 2nd
+ * position all previous gap lengths including its own.
+ * This reduces the lookup time to log (k) with k
+ * being the number of continuous gaps (not gap symbols) and comes at the price
+ * of O(k) worst-case runtime for gap modifications - when inserting or erasing
+ * gaps the tailing gaps have to be updated by the resulting offset.
+ * The gap maintaining structure is a set that is given a gap comparing structure.
+ * Sets are implemented as red-black trees and perform random read operations in
+ * log (k) time. The anchor set approach provides a good trade-off when using
+ * both operator[] and gap insertion/erasure.
+ *
+ */
+template <std::ranges::RandomAccessRange inner_type>
+//!\cond
+    requires container_concept<inner_type>
+//!\endcond
+class gap_decorator_anchor_set
+{
+    //!\brief The gap type as a tuple storing position and accumulated gap lengths.
+    using gap_t = typename std::pair<size_t, size_t>;
+
+    /*!\brief Structure allowing the comparison of gaps.
+     * \tparam        gap_t Type of the gap storing position an length accumulator.
+     * \details It is assumed that the gap structure is always in a consistent
+     * state, i.e. there are no two gaps that are overlapping. The ordering of gaps
+     * structures is exclusively dependent on the gap starting position which is
+     * the ordering criterion for the anchor set implemented as ordered red-black tree.
+     */
+    template <typename gap_t>
+    struct gap_compare {
+        /*!\brief The operator allowing gap_t comparison.
+         * \param[in] lhs   The left-hand gap to compare.
+         * \param[in] rhs   The right-hand side gap to be compared.
+         * \returns A boolean flag indicating whether the gap starting position
+         *          of \p lhs is smaller than the one of \p rhs.
+         */
+         bool operator() (const gap_t& lhs, const gap_t& rhs) const {
+             return lhs.first < rhs.first;
+         }
+     };
+    //!\brief The iterator type for an anchor set.
+    using set_iterator = typename std::set<gap_t, gap_compare<gap_t>>::iterator;
+
+public:
+    /*!\name Member types
+     * \{
+     */
+    //!\brief The alphabet type of the underlying sequence.
+    //!\hideinitializer
+    using alphabet_type = typename ranges::v3::value_type_t<inner_type>;
+    //!\brief The union type of the alphabet type and gap symbol type.
+    //!\hideinitializer
+    using value_type = gapped<alphabet_type>;
+    //!\brief Use the value type as reference type.
+    //!\hideinitializer
+    using reference = value_type;
+    //!\brief Use the const value type as reference type.
+    //!\hideinitializer
+    using const_reference = value_type const;
+    //!\brief Use the size_type of the underlying sequence.
+    //!\hideinitializer
+    using size_type = size_type_t<inner_type>;
+    //!\brief Use the difference_type of the underlying sequence.
+    //!\hideinitializer
+    using difference_type = typename ranges::v3::difference_type_t<inner_type>;
+    //!\brief The iterator type of this container (a random access iterator).
+    //!\hideinitializer
+    using iterator = detail::random_access_iterator<gap_decorator_anchor_set>;
+    //!\hideinitializer
+    using const_iterator = detail::random_access_iterator<gap_decorator_anchor_set const>;
+    //!\}
+
+    /*!\name Constructors/Destructors
+     * \{
+     */
+    //!\brief Default constructor.
+    constexpr gap_decorator_anchor_set() = default;
+    //!\brief Copy constructor.
+    constexpr gap_decorator_anchor_set(gap_decorator_anchor_set const &) = default;
+    //!\brief Copy construction via assignment.
+    constexpr gap_decorator_anchor_set & operator=(gap_decorator_anchor_set const &) = default;
+    //!\brief Move constructor.
+    constexpr gap_decorator_anchor_set (gap_decorator_anchor_set && rhs) = default;
+    //!\brief Move assignment.
+    constexpr gap_decorator_anchor_set & operator=(gap_decorator_anchor_set && rhs) = default;
+    //!\brief Construct by host sequence.
+    constexpr gap_decorator_anchor_set(inner_type const & sequence): sequence{&sequence} {};
+    //!\brief Direct sequence assignment resets previously inserted gaps.
+    constexpr gap_decorator_anchor_set & operator=(inner_type const & sequence)
+    {
+        sequence = sequence;
+        anchors.clear();
+    };
+    //!\brief Use default deconstructor.
+    ~gap_decorator_anchor_set() = default;
+    //!\}
+
+    /*!\brief Returns the total length of the aligned sequence.
+     * \returns The total length of the aligned sequence.
+     *
+     * \par Complexity
+     *
+     * Constant.
+     *
+     * \par Exceptions
+     *
+     * No-throw guarantee.
+     */
+    size_type size() const noexcept
+    {
+        if (anchors.size())
+            return (*(anchors.rbegin())).second + sequence->size();
+        return sequence->size();
+    }
+
+    /*!\brief Insert a gap of length size at the aligned sequence iterator position.
+     * \param it    Iterator indicating the gap start position in the aligned sequence.
+     * \param size  Number of gap symbols to be inserted.
+     * \returns     An iterator pointing to the start position of the insertion.
+     *
+     * \par Complexity
+     *
+     * Average and worst case (insertion before last gap): o(k),
+     * Best case (back insertion): o(log k).
+     *
+     * \par Exceptions
+     * Throws assertion if \p it points beyond the end position.
+     */
+    iterator insert_gap(iterator const it, size_type const size = 1)
+    {
+        if (!size)
+            return it;
+        size_type const pos = it - begin();
+        assert(pos <= this->size());
+
+        set_iterator it_set = anchors.begin();
+        // case 1: extend previous/surrounding gap already existing
+        if ((pos < this->size()) && (((value_type)(*this)[pos] == gap{}.to_char()) || (pos > 0 && (*this)[pos-1] == gap{}.to_char())))
+        {
+            it_set = anchors.lower_bound(gap_t{pos, 0/*Unused*/});
+            if (it_set == anchors.end() || (*it_set).first > pos)
+                it_set = std::prev(it_set);
+            gap_t gap{(*it_set).first, (*it_set).second + size};
+            // merge with successor
+            auto it_next = it_set;
+            ++it_next;
+            if ((*it_set) < (*std::prev(anchors.end())) && (*it_next).first <= (*it_set).first + size - 1)
+            {
+                // extend gap for *it_next, delete *(it_next+1)
+                gap.second += (*it_next).second;
+                anchors.erase(it_next);
+            }
+            anchors.erase(it_set);
+            anchors.insert(gap);
+        }
+        // case 2: create new anchor gap
+        else
+        {
+            gap_t gap{pos, size};
+            // pre: pos not in anchor set, find preceeding gap to add accumulated gaps
+            if (anchors.size())
+            {
+                auto it_aux = anchors.lower_bound(gap_t{pos, 0/*Unused*/});
+                if (it_aux != anchors.begin())
+                    gap.second += (*--it_aux).second;
+            }
+            anchors.insert(gap);
+        }
+        // post-processing: reverse update of succeeding gaps
+        rupdate(pos, size);
+        return it;
+    }
+
+   /*!\brief Erase one gap symbol at the indicated iterator postion.
+    * \param it     Iterator indicating the gap to be erased.
+    * \returns      An iterator pointing to starting position of the gap erasure.
+    *
+    * \par Complexity
+    *
+    * O(log k)
+    *
+    * \par Exceptions
+    * \throws seqan3::gap_erase_failure if character is no seqan3::gap.
+    */
+    iterator erase_gap(iterator const it)
+    {
+        // check if [it, it+gap_len[ covers [first, last[
+        if ((*it) != gap{}.to_char()) // [[unlikely]]
+            throw gap_erase_failure("The range to be erased does not corresponds to a consecutive gap.");
+        return erase_gap(it, it + 1);
+    }
+
+    /*!\brief Erase gap symbols at the iterator postions [first, last[.
+     * \param[in]   first    The iterator pointing to the position where to start inserting gaps.
+     * \param[in]   last     The iterator pointing to the position where to stop erasing gaps.
+     * \returns     An iterator pointing to starting position of the gap erasure.
+     *
+     * \par Complexity
+     *
+     * O(log k)
+     *
+     * \par Exceptions
+     *
+     * \throws seqan3::gap_erase_failure if [\p first, \p last[ does not correspond
+     * to a consecutive seqan3::gap range.
+     */
+    iterator erase_gap(iterator const first, iterator const last)
+    {
+        size_type pos1 = first - begin(), pos2 = last - begin();
+        set_iterator it = anchors.lower_bound(gap_t{pos1, 0/*Unused*/});
+        size_type gap_len = get_gap_length(it);
+
+        if (it == anchors.end() || (*it).first > pos1)
+            it = std::prev(it);
+        // check if [it, it+gap_len[ covers [first, last[
+        if (!(((*it).first <= pos1) && (((*it).first + gap_len) >= pos2))) // [[unlikely]]
+            throw gap_erase_failure("The range to be erased does not corresponds to a consecutive gap.");
+        // case 1: complete gap is deleted
+        if (((*it).first == pos1) && (gap_len == pos2-pos1))
+            anchors.erase(it);
+        // case 2: gap to be deleted in tail or larger than 1 (equiv. to shift tail left, i.e. pos remains unchanged)
+        else
+        {
+            gap_t gap{(*it).first, (*it).second - pos2 + pos1};
+            anchors.erase(it);
+            anchors.insert(gap);
+        }
+        // post-processing: forward update of succeeding gaps
+        update(pos1, pos2-pos1);
+        return first;
+    }
+
+    /*!\name Iterators
+     * \{
+     */
+    /*!\brief Returns an iterator to the first element of the container.
+     * \returns Iterator to the first element.
+     *
+     * If the container is empty, the returned iterator will be equal to end().
+     *
+     * \par Complexity
+     *
+     * Constant.
+     *
+     * \par Exceptions
+     *
+     * No-throw guarantee.
+     */
+    iterator begin() noexcept
+    {
+        return iterator{*this};
+    }
+
+    //!\copydoc begin()
+    const_iterator begin() const noexcept
+    {
+        return const_iterator{*this};
+    }
+
+    //!\copydoc begin()
+    const_iterator cbegin() const noexcept
+    {
+        return const_iterator{*this};
+    }
+
+    /*!\brief Returns an iterator to the element following the last element of the decorator.
+     * \returns Iterator to the behind last element.
+     *
+     * This element acts as a placeholder; attempting to dereference it results in undefined behaviour.
+     *
+     * \par Complexity
+     *
+     * Constant.
+     *
+     * \par Exceptions
+     *
+     * No-throw guarantee.
+     */
+    iterator end() noexcept
+    {
+        return iterator{*this, size()};
+    }
+
+    //!\copydoc end()
+    const_iterator end() const noexcept
+    {
+        return const_iterator{*this, size()};
+    }
+
+    //!\copydoc end()
+    const_iterator cend() const noexcept
+    {
+        return const_iterator{*this, size()};
+    }
+    //!\}
+
+    /*!\name Element access
+     * \{
+     */
+    /*!\brief Return the i-th element as a reference.
+     * \param i     The element to retrieve.
+     * \returns     A reference of the gapped alphabet type.
+     *
+     * \par Complexity
+     *
+     * Constant.
+     *
+     * \par Exceptions
+     *
+     * Throws std::out_of_range exception if \p i is out of range.
+     */
+    reference at(size_type const i)
+    {
+        //TODO add SEQAN_UNLIKELY
+        if (i >= size())
+            throw std::out_of_range{"Trying to access element behind the last in gap_decorator."};
+        return (*this)[i];
+    }
+
+    //!\copydoc at()
+    const_reference at(size_type const i) const
+    {
+        //TODO add SEQAN_UNLIKELY
+        if (i >= size())
+            throw std::out_of_range{"Trying to access element behind the last in gap_decorator."};
+        return (*this)[i];
+    }
+
+    //!\copydoc at()
+    constexpr reference operator[](size_type const i) const
+    {
+        assert(i < size());
+        // case 1: there are no gaps
+        if (!anchors.size())
+            return value_type((*sequence)[i]);
+        // case 2: there are gaps
+        set_iterator it = anchors.upper_bound(gap_t{i, 0/*Unused*/});
+        if (it == anchors.begin())
+            return value_type((*sequence)[i]); // since no gaps happen before i
+        it = std::prev(it);
+        size_type gap_len{(*it).second};
+        if (it != anchors.begin())
+            gap_len -= (*(std::prev(it, 1))).second;
+        if (i < (*it).first + gap_len)
+           return gap{}.to_char();
+        else
+           return value_type((*sequence)[i - (*it).second]);
+    }
+    //!\}
+
+    /*!\name Comparison operators
+     * \{
+     */
+    /*!\brief Compare gap decorators by underlying sequence and gaps.
+     * \param[in] rhs   The right-hand side gap decorator to compare.
+     * \returns A boolean flag indicating (in)equality of the aligned sequences.
+     *
+     * |par Complexity
+     * Worst case: O(n*log k)
+     *
+     * \par Exceptions
+     *
+     * No-throw guarantee. Does not modify the aligned sequences.
+     */
+    template <typename inner_type2>
+    //!\cond
+        requires std::is_same_v<std::remove_const_t<inner_type>, std::remove_const_t<inner_type2>>
+    //!\endcond
+    constexpr bool operator==(gap_decorator_anchor_set<inner_type2> const & rhs) const noexcept
+    {
+        if (anchors.size() != rhs.anchors.size() || sequence->size() != rhs.sequence->size())
+            return false;
+
+        return std::ranges::equal(*this, rhs);
+    }
+
+    //!\copydoc operator==
+    template <typename inner_type2>
+    //!\cond
+        requires std::is_same_v<std::remove_const_t<inner_type>, std::remove_const_t<inner_type2>>
+    //!\endcond
+    constexpr bool operator!=(gap_decorator_anchor_set<inner_type2> const & rhs) const noexcept
+    {
+        return !(operator==(rhs));
+    }
+    //!\}
+private:
+    /*!\brief Helper function to compute the length of the gap indicated by the
+     * input iterator.
+     * \param[in] it    Iterator over the internal gap set.
+     * \returns The gap length corresponding to the gap pointed at by \p it.
+     * \details The length of a gap is difference of the accumulator pointed at
+     * and the one of its predecessor (if existing).
+     *
+     * \par Exceptions
+     * No-throw guarantee.
+     */
+    constexpr size_type get_gap_length(set_iterator it) const noexcept
+    {
+        if (it == anchors.begin())
+            return (*it).second;
+        return (*it).second - (*std::prev(it)).second;
+    }
+
+    /*!\brief Update all anchor gaps after the indicated position by adding an offset.
+     * \param[in] pos   Gap index after which to perform the update.
+     * \param[in] size  Offset to be added to the virtual gap positions and its accumulators.
+     *
+     * |par Complexity
+     * Linear in the number of gaps.
+     *
+     * \details For not invalidating the iterator over the ordered set, the
+     * update is done in reverse manner excluding the indicated gap.
+     */
+    void rupdate(size_type const pos, size_type const size)
+    {
+        size_type new_key, new_val;
+        for (auto it = std::prev(anchors.end(), 1); (*it).first > pos;)
+        {
+            new_key = (*it).first + size;
+            new_val = (*it).second + size;
+            anchors.emplace_hint(it, gap_t{new_key, new_val});
+            anchors.erase(*it--);
+        }
+    }
+
+    /*!\brief Update all anchor gaps after indicated position by substracting an offset.
+     * \param[in] pos   Gap index after which to perform the update.
+     * \param[in] size  Offset to be removed from the virtual gap positions and its accumulators.
+     *
+     * |par Complexity
+     * Linear in the number of gaps.
+     *
+     * \details For not invalidating the iterator over the ordered set, the
+     * decreasing is done in a forward manner excluding the indicated gap.
+     *
+     * \par Exceptions
+     * Throws assert when initial update position is out of range.
+     */
+    void update(size_type const pos, size_type const size)
+    {
+        assert(pos < size);
+        auto it = anchors.lower_bound(gap_t{pos + size + 1, 0/*Unused*/});
+        while (it != anchors.end())
+        {
+            gap_t gap{(*it).first - size, (*it).second - size};
+            anchors.insert(gap);
+            set_iterator it_next = std::next(it);
+            anchors.erase(it);
+            it = it_next;
+        }
+    }
+
+    //!\brief Stores a pointer to the ungapped, underlying sequence.
+    inner_type const * sequence{};
+    //!\brief Set storing the anchor gaps.
+    std::set<gap_t, gap_compare<gap_t>> anchors{};
+};
+
+} // namespace seqan3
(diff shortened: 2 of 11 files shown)


include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -347,4 +347,76 @@ inline debug_stream_type & operator<<(debug_stream_type & stream, tuple_t const
...
+ *
+ * This function delegates to the member function `erase(iterator, iterator)` of
+ * the gap decorator. Before delegating, the function checks if the range
+ * [\p first, \p last) contains only seqan3::gap symbols.
+ */
+template<typename gap_decorator_type>
+//!\cond
+    requires requires (gap_decorator_type v) { v.erase_gap(v.begin(), v.end()); }
+//!\endcond
+typename gap_decorator_type::iterator erase_gap(gap_decorator_type & gd, typename gap_decorator_type::iterator const first, typename gap_decorator_type::iterator const last)
```
eseiler: line length

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,525 @@
...
+ * \tparam inner_type The type of alphabet sequences that will be stored.
+ * \implements seqan3::aligned_sequence_concept
+ * \ingroup decorator
+ *
+ * This class may be used whenever you want to store or compute an alignment.
+ * The underlying (ungapped sequence) remains unmodified, and is augmented by
+ * gap information. Iterating over a gap_decorator sequences behaves as if you
+ * had a rolled out, aligned sequence with alphabet and gap symbols.
+ *
+ * \details
```
eseiler: Could you use latex here for the O annotations and so on?
e.g.
https://github.com/seqan/seqan3/blob/8df4ff7e5ce54aeaa8604db623e621b144180829/include/seqan3/search/fm_index/fm_index.hpp#L73

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,525 @@
...
+ * The gap_decorator_anchor_set is a modified anchor list approach -- instead
```
eseiler:
```suggestion
 * The gap_decorator_anchor_set is a modified anchor list approach - instead
```

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,525 @@
...
+     * No-throw guarantee.
+     */
+    size_type size() const noexcept
+    {
+        if (anchors.size())
+            return (*(anchors.rbegin())).second + sequence->size();
+        return sequence->size();
+    }
+
+    /*!\brief Insert a gap of length size at the aligned sequence iterator position.
```
eseiler:
```suggestion
    /*!\brief Insert a gap of length count at the aligned sequence iterator position.
```

?

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,525 @@
...
+    *
+    * \par Complexity
+    *
+    * O(log k)
+    */
+    iterator erase_gap(iterator const it)
+    {
+        // check if [it, it+gap_len[ covers [first, last[
+        if ((*it) != gap{}) // [[unlikely]]
+            throw gap_erase_failure("The range to be erased does not corresponds to a consecutive gap.");
```
eseiler:
```suggestion
            throw gap_erase_failure("The range to be erased does not correspond to a consecutive gap.");
```

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,525 @@
...
+            throw gap_erase_failure{"There is no gap to erase in range [" + std::to_string(pos1) + "," +
+                                    std::to_string(pos2) + "]."};
+
+        --it;
+        size_type const gap_len = get_gap_length(it);
+
+        // check if [it, it+gap_len[ covers [first, last[
+        if (((*it).first + gap_len) < pos2) // [[unlikely]]
+        {
+            throw gap_erase_failure{"The range to be erased does not corresponds to a consecutive gap."};
```
eseiler:
```suggestion
            throw gap_erase_failure{"The range to be erased does not correspond to a consecutive gap."};
```

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,525 @@
...
+    //!\copydoc begin()
+    const_iterator cbegin() const noexcept
+    {
+        return const_iterator{*this};
+    }
+
+    /*!\brief Returns an iterator to the element following the last element of the decorator.
+     * \returns Iterator to the behind last element.
+     *
+     * This element acts as a placeholder; attempting to dereference it results in undefined behaviour.
```
eseiler: `\attention` ?

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,525 @@
...
+     * Constant.
+     *
+     * \par Exceptions
+     *
+     * Throws std::out_of_range exception if \p i is out of range.
+     */
+    reference at(size_type const i)
+    {
+        //TODO add SEQAN_UNLIKELY
+        if (i >= size())
```
eseiler:
```suggestion
        if (i >= size()) // [[unlikely]]
```
and remove comment in line above

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,525 @@
...
+            throw std::out_of_range{"Trying to access element behind the last in gap_decorator."};
+        return (*this)[i];
+    }
+
+    //!\copydoc at()
+    const_reference at(size_type const i) const
+    {
+        //TODO add SEQAN_UNLIKELY
+        if (i >= size())
```
eseiler:
```suggestion
        if (i >= size()) // [[unlikely]]
```
and remove comment in line above

test/unit/range/decorator/gap_decorator_anchor_set_test.cpp
```diff
@@ -0,0 +1,156 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2018, Knut Reinert & Freie Universität Berlin
```
eseiler: copyright year

test/unit/range/decorator/gap_decorator_test_template.hpp
```diff
@@ -0,0 +1,156 @@
+// ============================================================================
+//                 SeqAn - The Library for Sequence Analysis
+// ============================================================================
+//
+// Copyright (c) 2006-2018, Knut Reinert & Freie Universitaet Berlin
```
eseiler: copyright header

smehringer:
@h-2 One design decision that is not handled yet: What if the underlying sequence pointer is not set, but the user tries to insert gaps?

h-2:
> @h-2 One design decision that is not handled yet: What if the underlying sequence pointer is not set, but the user tries to insert gaps?

Either assert (handled in debug mode, UB in release mode) or do an if check and throw. You can decide based on whether you think this is performance relevant and/or whether the case that the ptr is actually 0 happens often or not. In any case, please make sure that the behaviour is well documented and that the ptr is actually initialised to `nullptr` when the adaptor is default constructed.

smehringer:
Why do we want the decorator to be default constructible at all? It should really only be used to decorate an object, a default constructed decorator is not valid IMPOV. Next to insert_gap, the size function also does not make sense...

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,537 @@
...
+    //!\copydoc operator==
+    template <typename inner_type2>
+    //!\cond
+        requires std::is_same_v<std::remove_const_t<inner_type>, std::remove_const_t<inner_type2>>
+    //!\endcond
+    constexpr bool operator!=(gap_decorator_anchor_set<inner_type2> const & rhs) const noexcept
+    {
+        return !(operator==(rhs));
+    }
+    //!\}
```
h-2: implement comparison operators as friends, add lexicographical comparison. I think you can use things like std::less to make this whole thing easier.

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,537 @@
...
+        /*!\brief The operator allowing gap_t comparison.
+         * \param[in] lhs   The left-hand gap to compare.
+         * \param[in] rhs   The right-hand side gap to be compared.
+         * \returns A boolean flag indicating whether the gap starting position
+         *          of \p lhs is smaller than the one of \p rhs.
+         */
+         bool operator() (const gap_t& lhs, const gap_t& rhs) const {
+             return lhs.first < rhs.first;
+         }
+     };
```
h-2: Why do we have this? Why not just a lambda function?

h-2: If we need this, correct placement of braces

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,537 @@
...
+     * \tparam        gap_t Type of the gap storing position an length accumulator.
+     *
+     * \details
+     *
+     * It is assumed that the gap structure is always in a consistent
+     * state, i.e. there are no two gaps that are overlapping. The ordering of gaps
+     * structures is exclusively dependent on the gap starting position which is
+     * the ordering criterion for the anchor set implemented as ordered red-black tree.
+     */
+    template <typename gap_t>
```
h-2: confusing usage of `gap_t` as a template parameter and as a member type before

h-2: Also gap_t sounds like an alphabet, I would really use something different here. We need to have consistent naming throughout SeqAn. 

gap = single gap symbol
? = continuous stretch of gap symbols

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,537 @@
...
+     * \par Exceptions
+     * Throws assert when initial update position is out of range.
+     */
+    void update(set_iterator_type it, size_type const offset)
+    {
+        while (it != anchors.end())
+        {
+            gap_t gap{(*it).first - offset, (*it).second - offset};
+            it = anchors.erase(it);
+            it = anchors.insert(it, gap);
```
h-2: no hint here?

smehringer: There is not `insert_hint` but the `it` parameter is the hint :)

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
smehringer: I removed it in favour of a `dummy`_bound` that can be used when searching with lower_bound() because we need to make sure that the second pair value is never smaller. This is the same as overloading the comparator.

include/seqan3/std/ranges
```diff
@@ -239,4 +239,37 @@ using SEQAN3_DOXYGEN_ONLY(iter_move =) ::ranges::iter_move;
 //!\}
 } // namespace std::ranges
 
+// -----------------------------------------------------------------------------------------------------
+// enable_view alias
+// -----------------------------------------------------------------------------------------------------
+
+/* The ranges::enable_view type trait indicates whether a type is a View. In order to create an alias
+ * for the std library that we can specialize for seqan3 types (we cannot specialize using declarations)
```
eseiler:
```suggestion
 * for the std library that we can specialise for seqan3 types (we cannot specialise using declarations)
```

include/seqan3/std/ranges
```diff
@@ -239,4 +239,37 @@ using SEQAN3_DOXYGEN_ONLY(iter_move =) ::ranges::iter_move;
...
+ * we create a new std::ranges::enable_view struct. Additionally we now need an ranges::enable_view overload
```
eseiler:
```suggestion
 * we create a new std::ranges::enable_view struct. Additionally, we now need an ranges::enable_view overload
```

include/seqan3/std/ranges
```diff
@@ -239,4 +239,37 @@ using SEQAN3_DOXYGEN_ONLY(iter_move =) ::ranges::iter_move;
...
+ * that delegates to std::ranges::enable_view<type> whenever a type trait specialization exists.
```
eseiler:
```suggestion
 * that delegates to std::ranges::enable_view<type> whenever a type trait specialisation exists.
```

include/seqan3/std/ranges
```diff
@@ -239,4 +239,37 @@ using SEQAN3_DOXYGEN_ONLY(iter_move =) ::ranges::iter_move;
...
+ */
+
+namespace std::ranges
+{
+
+//!\brief Type trait that indicated wither `type` is a view.
```
eseiler:
```suggestion
//!\brief Type trait that indicates whether `type` is a view.
```

include/seqan3/std/ranges
```diff
@@ -239,4 +239,37 @@ using SEQAN3_DOXYGEN_ONLY(iter_move =) ::ranges::iter_move;
...
+template <typename type>
+struct enable_view
+{};
+
+} // namespace std::ranges
+
+namespace ranges
+{
+
+//!\brief Type trait that indicated wither `type` is a view.
```
eseiler:
```suggestion
//!\brief Type trait that indicates whether `type` is a view.
```

test/unit/range/decorator/gap_decorator_anchor_set_test.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+#include "../../alignment/aligned_sequence_test_template.hpp"
+
+using namespace seqan3;
+
+template <typename inner_type_>
+class aligned_sequence<gap_decorator_anchor_set<inner_type_>> : public ::testing::Test
+{
+public:
+    // Initializer function is needed for the typed test because the gapped_decorator
```
eseiler: intiali**s**e everywhere?

test/unit/range/decorator/gap_decorator_anchor_set_test.cpp
```diff
@@ -0,0 +1,243 @@
...
+        TypeParam dec{};
+        TypeParam dec2(dec);
+        EXPECT_EQ(dec, dec2);
+
+        TypeParam const dec_const{};
+        TypeParam const dec2_const(dec_const);
+        EXPECT_EQ(dec_const, dec2_const);
+    }
+
+    // copy assignment
```
eseiler: This is still copy constructor.

```
        TypeParam dec{};
        TypeParam dec2;
        dec2 = dec;
        EXPECT_EQ(dec, dec2);
```
would be assignment.

Then the const version will also not work (copy assignment should not work for a const type).

See https://en.cppreference.com/w/cpp/language/copy_constructor

Same for move assignment.

test/unit/range/decorator/gap_decorator_anchor_set_test.cpp
```diff
@@ -0,0 +1,243 @@
...
+    // from another decorator
+    {
+        TypeParam dec{};
+        TypeParam dec2;
+        dec2 = dec;
+        EXPECT_EQ(dec, dec2);
+
+        // const
+        TypeParam const dec_const{};
+        TypeParam dec2_const;
```
eseiler: Here you don't use `const` because it wouldn't work otherwise 😉

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,510 @@
...
+     * \par Exceptions
+     *
+     * No-throw guarantee. Does not modify the aligned sequences.
+     */
+    constexpr friend bool operator==(gap_decorator_anchor_set const & lhs, gap_decorator_anchor_set const & rhs)
+    {
+        if (lhs.size() != rhs.size())
+            return false;
+
+        for (size_t i = 0; i < std::min<size_type>(lhs.size(), rhs.size()); ++i)
```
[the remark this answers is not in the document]
smehringer: :woman_facepalming:

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,523 @@
...
+
+    //!\copydoc at()
+    const_reference at(size_type const i) const
+    {
+        if (i >= size()) // [[unlikely]]
+            throw std::out_of_range{"Trying to access element behind the last in gap_decorator."};
+        return (*this)[i];
+    }
+
+    //!\copydoc at()
```
eseiler: This means `operator[]` is documented with the `std::out_of_range` exception, even though it does not throw (or at least shouldn't throw).

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,523 @@
...
+    constexpr reference operator[](size_type const i) const
```
eseiler:
```suggestion
    constexpr reference operator[](size_type const i) const noexcept
```

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,523 @@
...
+     * \returns A boolean flag indicating (in)equality of the aligned sequences.
+     *
+     * \par Complexity
+     * Worst case: \f$O(n*\log k)\f$
+     *
+     * \par Exceptions
+     *
+     * No-throw guarantee. Does not modify the aligned sequences.
+     */
+    constexpr friend bool operator==(gap_decorator_anchor_set const & lhs, gap_decorator_anchor_set const & rhs)
```
eseiler:
```suggestion
    constexpr friend bool operator==(gap_decorator_anchor_set const & lhs, gap_decorator_anchor_set const & rhs) noexcept
```

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,523 @@
...
+        for (size_t i = 0; i < lhs.size(); ++i)
+            if (lhs[i] != rhs[i])
+                return false;
+
+        return true;
+        // return ranges::equal(lhs, rhs); // does not work
+    }
+
+    //!\copydoc operator==
+    constexpr friend bool operator!=(gap_decorator_anchor_set const & lhs, gap_decorator_anchor_set const & rhs)
```
eseiler:
```suggestion
    constexpr friend bool operator!=(gap_decorator_anchor_set const & lhs, gap_decorator_anchor_set const & rhs) noexcept
```

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,523 @@
...
+     * \details
+     *
+     * In order to avoid key conflicts when inserting into anchors, the
+     * decreasing is done in a forward manner excluding the indicated gap.
+     *
+     * \par Complexity
+     * Linear in the number of gaps.
+     *
+     * \par Exceptions
+     * Throws assert when initial update position is out of range.
```
eseiler: Do we document assertions?
Or does it throw an exception?

test/unit/range/container/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
-seqan3_test(aligned_allocator_test.cpp)
-seqan3_test(container_concept_test.cpp)
-seqan3_test(container_of_container_test.cpp)
-seqan3_test(container_test.cpp)
+#seqan3_test(aligned_allocator_test.cpp)
+#seqan3_test(container_concept_test.cpp)
+#seqan3_test(container_of_container_test.cpp)
+#seqan3_test(container_test.cpp)
```
eseiler: what's with these tests?

test/unit/range/decorator/gap_decorator_anchor_set_test.cpp
```diff
@@ -0,0 +1,216 @@
...
+    void initialise_typed_test_container(gap_decorator_anchor_set<inner_type_> & container, dna4_vector const & target)
+    {
+        container = target;
+    }
+};
+
+template <typename t>
+class gap_decorator_anchor_set_f : public ::testing::Test {};
+
+const std::vector<dna4> dummy_obj{}; // dummy lvalue for type declaration of views
```
eseiler: This currently doesn't work with `bitcompressed_vector` (because of some bug) but should, right?

test/unit/range/container/CMakeLists.txt
smehringer: I don't know ...

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -258,6 +258,92 @@ inline typename seq_type::iterator erase_gap(seq_type & seq,
...
+    return rng.insert_gap(it, size);
+}
+
+/*!\brief An implementation of seqan3::aligned_sequence_concept::erase_gap for ranges with the corresponding
+ *        member function erase_gap(it).
+ * \tparam range_type   Type of the range to modify; must have an erase_gap(it) member function.
+ * \param[in,out] rng   The range to modify.
+ * \param[in] it        The iterator pointing to the position where to erase one gap.
+ *
+ * \relates seqan3::aligned_sequence_concept
```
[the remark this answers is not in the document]
smehringer: where else? I think they are generic overloads that make ranges that feature the member function fulfil the concept.

smehringer:
@h-2 Can you take a new look at the iterator that no increments and decrements in constant time?

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,766 @@
...
+ * Resolving random access requires logarithmic access into the set and inserting or removing a gap symbol entails
+ * updating all subsequent elements in the same to preserve correct cumulative sizes.
+ *
+ */
+
+template <std::ranges::View inner_type>
+    requires std::ranges::RandomAccessRange<inner_type> && std::ranges::SizedRange<inner_type>
+class gap_decorator_anchor_set
+{
+public:
```
h-2: definition of the iterator type should be private.

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,766 @@
...
+            return *this;
+        }
+
+        //!\brief Decrement iterator by skip.
+        constexpr gap_decorator_anchor_set_iterator & operator-=(difference_type const skip) noexcept
+        {
+            this->pos -= skip;
+            update(*this->host, this->pos);
+            return *this;
+        }
```
h-2: BidirectionalIterators don't have `-=` and `+=`

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,766 @@
...
+        //!\brief Import the parent's constructors.
+        using base_t::base_t;
+
+        //!\brief Construct by host, delegate to CRTP layer and initialize members.
+        explicit constexpr gap_decorator_anchor_set_iterator(gap_decorator_t & host) noexcept :
+            base_t(host), ungapped_view_it{host.ungapped_view.begin()}, anchor_set_it{host.anchors.begin()}
+        {
+            if (host.anchors.size() && (*host.anchors.begin()).first == 0) // there are gaps at the very front
+            {
+                --ungapped_view_it; // hacky hack hack ? (definitely will work with an index instead of an iterator)
```
h-2: This is undefined behaviour unfortunately.

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,766 @@
...
+            {
+                ++ungapped_view_it;
+            }
+            else
+            {
+                prev_gap_end = (*anchor_set_it).first + (*anchor_set_it).second -
+                               ((anchor_set_it != (*this->host).anchors.begin()) ?
+                                    (*std::prev(anchor_set_it)).second : 0);
+                ++anchor_set_it;
+            }
```
h-2: Can you put a one line comment above each case documenting it? Like "We are still within gap-stretch and will stay in gap-stretch"

h-2: Cool that you could reduce the branches :+1:

h-2: also can you annotate branches with `// [[unlikely]]` if they are edge-cases?
[edge-cases "before-begin" or "at-end"]

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,766 @@
...
+    {
+        if (lhs.size() != rhs.size())
+            return false;
+
+        for (size_t i = 0; i < lhs.size(); ++i)
+            if (lhs[i] != rhs[i])
+                return false;
+
+        return true;
+        // return ranges::equal(lhs, rhs); // does not work
```
h-2: 1) first compare size
2) compare equality of the std::set
3) return ranges::equal on the underlying range

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,766 @@
...
+                return false;
+            else
+                return true;          // lhs prefix of rhs
+        }
+        else if (i == rhs.size())
+        {
+            return false;             // rhs prefix of lhs
+        }
+
+        return lhs[i] < rhs[i];
```
h-2: don't use [] here either... I think you can just do ranges::less?

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,766 @@
...
+     *
+     * \details
+     *
+     * The length of a gap is difference of the accumulator pointed at
+     * and the one of its predecessor (if existing).
+     *
+     * ### Exceptions
+     * No-throw guarantee.
+     */
+    constexpr size_type get_gap_length(set_iterator_type it) const noexcept
```
h-2: I would recommend to use `get` and `set` only for getters and setters, not for other function names.

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,766 @@
...
+  */
+
+#pragma once
+
+#include <limits>
+#include <set>
+#include <tuple>
+#include <type_traits>
+
+#include <seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp>
```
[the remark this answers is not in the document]
smehringer: if not and you are ONLY including the gap_decorator_anchor_set.hpp the free functions erase/insert_gap will not be available to the user..

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,766 @@
...
+#include <seqan3/alphabet/concept.hpp>
+#include <seqan3/alphabet/gap/gapped.hpp>
```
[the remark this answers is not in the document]
smehringer: But it is used in 
```
using value_type = gapped<value_type_t<inner_type>>;
```
... does this work with forward declarations?

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
smehringer: there are no edge cases anymore.

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,830 @@
...
+     * \{
+     */
+    /*!\brief Insert a gap of length count at the aligned sequence iterator position.
+     * \param it     Iterator indicating the gap start position in the aligned sequence.
+     * \param count  Number of gap symbols to be inserted.
+     * \returns      An iterator pointing to the start position of the insertion.
+     *
+     * ### Complexity
+     *
+     * Average and worst case (insertion before last gap): \f$o(k)\f$,
```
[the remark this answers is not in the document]
smehringer: Are you sure? This was written by Marie and I have not put a lot of thought into the runtime analysis yet

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,830 @@
...
+     *
+     * ### Complexity
+     * Worst case: \f$O(n*\log k)\f$
+     * Constant in case the decorators have not the same number of (consecutive) gaps.
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee. Does not modify the aligned sequences.
+     */
+    constexpr friend bool operator==(gap_decorator_anchor_set const & lhs, gap_decorator_anchor_set const & rhs) noexcept
```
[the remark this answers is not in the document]
smehringer: good point!

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,830 @@
...
+    inline static size_t const bound_dummy{std::numeric_limits<size_t>::max()};
+
+    /*!\brief Helper function to compute the length of the gap indicated by the
+     * input iterator.
+     * \param[in] it    Iterator over the internal gap set.
+     * \returns The gap length corresponding to the gap pointed at by \p it.
+     *
+     * \details
+     *
+     * The length of a gap is difference of the accumulator pointed at
```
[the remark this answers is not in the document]
smehringer: I changed it to `The length of a gap pointed to by \p it is the difference of the current sum of gaps (second tuple position) and the one of its predecessor (if existing).`, better ?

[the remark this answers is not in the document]
eseiler: Yes, I like it

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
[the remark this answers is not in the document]
eseiler: In the main docs in this file, we write O(k). I think o(k) wouldn't hold, since if we have 0 or 1 gaps, we don't have the update penalty of k-1, and for o(k) it would have to hold for all k >= 0, whereas for O(k) there is a number x such that it hold for all k >= x.

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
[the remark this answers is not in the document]
h-2: This is the same for the alphabets. If you just include dna4, you just get what dna4 contains. If you included the concept you also get the member exposure stuff (in that regard it is somehow associated with the concept :roll_eyes: ).

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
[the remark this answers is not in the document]
h-2: Have you tried?

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,767 @@
...
+ * ### Performance
+ *
+ * **n** The length of the underlying sequence.
+ * **k** The number of contiguous gaps (not gap symbols).
+ * **l** The total number of gap symbols.
+ *
+ * |            | access next | random access    | gap insert/erase at end | gap insert/erase random |
+ * |------------|-------------|----------------- |-------------------------|-------------------------|
+ * | decorator  | \f$O(1)\f$  | \f$O(\log(k))\f$ | \f$O(\log(k))\f$        | \f$O(k)\f$              |
+ * | vector     | \f$O(1)\f$  | \f$O(1)\f$       | \f$O(n)\f$              | \f$O(n)\f$              |
```
h-2: I don't think insert/erase at end is in O(n) for the vector, but in O(1) or not?

h-2: You could put the size overhead into the table, too, I guess, with a small note below it that explains that it refers to the size required in addition to the original ungapped sequence storage.

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,767 @@
...
+    //!\brief The iterator that moves over the seqan3::gap_decorator_anchor_set.
+    class gap_decorator_anchor_set_iterator
+    {
+    private:
+        //!\brief Pointer to the underlying container structure.
+        typename std::add_pointer_t<gap_decorator_anchor_set const> host{nullptr};
+        //!\brief Stores the virtual position index for the seqan3::gap_decorator_anchor_set.
+        typename gap_decorator_anchor_set::size_type pos{0u};
+        //!\brief Stores the physical position in the ungapped/underlying view.
+        int32_t ungapped_view_pos{0u};
```
h-2: Why is this not of size_type?

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,767 @@
...
+                    {
+                        left_gap_end = 0;
+                    }
+                }
+            }
+            else if (anchor_set_it == (*host).anchors.end() || pos < (*anchor_set_it).first)
+            {   // we are neither at the left nor right gap
+                --ungapped_view_pos;
+            }
+            // else -> no op (we are still within the right gap stretch)
```
h-2: How did we end up with so many branches in decrement when increment needs only three? (In the end, I would say it's preferable to have them in decrement, just wondering)

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,767 @@
...
+        }
+        //!\}
+
+        /*!\name Reference/Dereference operators
+         * \{
+        */
+        //!\brief Dereference operator returns element currently pointed at.
+        constexpr reference operator*() const noexcept
+        {
+            if (pos >= left_gap_end && (anchor_set_it == (*host).anchors.end() || pos < (*anchor_set_it).first))
```
h-2: This has to be evaluated on every deref now? Could/should this be cached?

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
smehringer: > I don't think insert/erase at end is in O(n) for the vector, but in O(1) or not?

but in the worst case push_back causes a resize of the vector no?

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
smehringer: Because in case there are gaps at the beginning I want to set `ungapped_view_pos` to -1 on construction which saves me an if branch in the increment operator.

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,767 @@
...
+                return value_type{(*host).ungapped_view[ungapped_view_pos]};
+            else
+                return gap{};
+        }
+
+        //!\brief Returns pointer to this iterator.
+        constexpr pointer operator->() const noexcept
+        {
+            // ?
```
h-2: Wow, this took me longer to research than I thought. 

This question sums up your use-case:
https://stackoverflow.com/questions/27582845/how-to-implement-operator-for-an-iterator-that-constructs-its-values-on-demand

In short: Strictly speaking you cannot make this iterator model anything stronger than InputIterator **according to the old language specification**. AFAICT the new requirements (post ranges) concepts make no assumptions about `operator->` at all.

There are two ways to deal with this:
1. just remove the operator and also the `::pointer` member type
2. store a member a variable of type `value_type` in the iterator and have every increment/decrement store the current value in that; have `operator->` return the address of that. 

The benefit of solution 2 is that old algorithms would still consider this an iterator and also you can just return that value on `operator*` and don't need the check.

You could also ask on stack overflow of course.

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,767 @@
...
+        }
+        //!\}
+
+        /*!\name Comparison operators
+         * \brief Compares iterators by virtual position.
+         * \{
+         */
+        constexpr bool operator==(gap_decorator_anchor_set_iterator const & rhs) const noexcept
```
h-2: implement comparison operators as friends

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,767 @@
...
+        }
+        //!\}
+    };
+
+public:
+    /*!\name Member types
+     * \{
+     */
+    //!\brief The underlying ungapped range type.
+    using ungapped_range_type = inner_type;
```
h-2: I would not group this with the others because it's not one of the typically "range-associated" types.

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,767 @@
...
+    using anchor_gap_t = typename std::pair<size_t, size_t>;
+
+    //!\brief The type of set to store the anchor gaps.
+    using anchor_set_type = std::set<anchor_gap_t>;
+
+    //!\brief The iterator type for an anchor set.
+    using set_iterator_type = typename anchor_set_type::iterator;
+
+    //!\brief The maximum value is needed for a correct search with upper_bound() in the anchor set.
+    inline static size_t const bound_dummy{std::numeric_limits<size_t>::max()};
```
h-2: constexpr?

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
h-2: It's amortised constant. All run-times given are amortised, but you can also write `~O(1)` to make this clear if you want to.

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
h-2: Ok, but shouldn't it still be 64bit?

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
smehringer: acutally it has only two outer branches.. I can replace on branch by a ternary expression but thats it.

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
smehringer: I guess this can be cashed..

smehringer: when I cash it I need again more if branches in the increment/decrement

smehringer:
@h-2 please re-review

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -83,34 +79,42 @@ class gap_decorator_anchor_set
         //!\brief Stores the virtual position index for the seqan3::gap_decorator_anchor_set.
         typename gap_decorator_anchor_set::size_type pos{0u};
         //!\brief Stores the physical position in the ungapped/underlying view.
-        int32_t ungapped_view_pos{0u};
+        int64_t ungapped_view_pos{0u};
```
eseiler:
```suggestion
        gap_decorator_anchor_set::size_type ungapped_view_pos{0u};
```

?

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -83,34 +79,42 @@ class gap_decorator_anchor_set
...
 
-                if (prev != (*host).anchors.begin())
+                if (prev != host->anchors.begin())
                     gap_len -= (*(std::prev(prev))).second;
 
                 ungapped_view_pos -= (*prev).second;
                 left_gap_end = (*prev).first + gap_len;
             }
+
+            if (ungapped_view_pos != static_cast<int64_t>(host->ungapped_view.size()) &&
```
eseiler: if `ungapped_view_pos` is of type `gap_decorator_anchor_set::size_type` you don't need the `static_cast` anymore

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -176,23 +183,26 @@ class gap_decorator_anchor_set
...
 
             if (pos < left_gap_end)
             {   // we stay within the preceding gap stretch
                 // no op but must precede other cases
             }
-            else if (anchor_set_it == (*host).anchors.end() || pos < (*anchor_set_it).first)
+            else if (anchor_set_it == host->anchors.end() || pos < (*anchor_set_it).first)
             {   // proceed within the view since we are right of the previous gap but didn't arrive at the right gap yet
                 ++ungapped_view_pos;
+                if (ungapped_view_pos != static_cast<int64_t>(host->ungapped_view.size()))
```
eseiler: Same here

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
smehringer: no! because this value might become -1

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
smehringer: see above

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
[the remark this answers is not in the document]
h-2: Still open

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,782 @@
...
+ * **n** The length of the underlying sequence.
+ * **k** The number of contiguous gaps (not gap symbols).
+ * **l** The total number of gap symbols.
+ *
+ * |            | access next | random access    | gap insert/erase at end | gap insert/erase random | size overhead |
+ * |------------|-------------|----------------- |-------------------------|-------------------------|---------------|
+ * | decorator  | \f$O(1)\f$  | \f$O(\log(k))\f$ | \f$O(\log(k))\f$        | \f$O(k)\f$              | O(k)          |
+ * | vector     | \f$O(1)\f$  | \f$O(1)\f$       | \f$O(n)\f$              | \f$O(n)\f$              | O(n)          |
+ *
+ * The *size overhead* refers to the additional space that is needed when using each of the data structures.
```
h-2: :nail_care: 

```
The *size overhead* refers to the space that is needed when using each of the data structures in addition to an already existing ungapped sequence.
```

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,782 @@
...
+ * ### Performance
+ *
+ * **n** The length of the underlying sequence.
+ * **k** The number of contiguous gaps (not gap symbols).
+ * **l** The total number of gap symbols.
+ *
+ * |            | access next | random access    | gap insert/erase at end | gap insert/erase random | size overhead |
+ * |------------|-------------|----------------- |-------------------------|-------------------------|---------------|
+ * | decorator  | \f$O(1)\f$  | \f$O(\log(k))\f$ | \f$O(\log(k))\f$        | \f$O(k)\f$              | O(k)          |
+ * | vector     | \f$O(1)\f$  | \f$O(1)\f$       | \f$O(n)\f$              | \f$O(n)\f$              | O(n)          |
```
h-2: latex escapes around the last column's entries

h-2: I still insist that insert at end of a vector is amortised constant. That's the point of a vector.

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,782 @@
...
+ *
+ * This decorator stores a std::set over tuples of `(pos, cumulative_size)` where every entry represents one
+ * contiguous stretch of gaps. `pos` is the (virtual) insert position in the underlying range and `cumulative_size`
+ * is the length of that contiguous stretch of gaps plus the length of all preceding elements.
+ * Resolving random access requires logarithmic access into the set and inserting or removing a gap symbol additionally
+ * entails updating all subsequent elements in the set to preserve correct cumulative sizes.
+ *
+ */
+template <std::ranges::RandomAccessRange inner_type>
+    requires std::ranges::SizedRange<inner_type>
```
h-2: dont you need to escape these for the documentation?

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,782 @@
...
+        typename gap_decorator_anchor_set::size_type pos{0u};
+        //!\brief Stores the physical position in the ungapped/underlying view.
+        int64_t ungapped_view_pos{0}; // must be signed because we need this value to be -1 in case of leading gaps.
+        //!\brief Stores the position (inkl. gaps) where the last (consecutive) gap that is still before the current
+        //!       iterator position ends.
+        typename gap_decorator_anchor_set::size_type left_gap_end{0};
+        //!\brief A pointer to the current anchor gap node. Note that the current tuple value at position 0 is the
+        //!       start of the right gap that is still behind the current iterator position.
+        typename gap_decorator_anchor_set::set_iterator_type anchor_set_it{};
+        //!\brief Cashes the current value pointed at.
```
h-2: :dollar: :heavy_dollar_sign: :moneybag: :money_mouth_face:

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,782 @@
...
+            gap_decorator_anchor_set_iterator cpy{*this};
+            --(*this);
+            return cpy;
+        }
+        //!\}
+
+        /*!\name Reference/Dereference operators
+         * \{
+        */
+        //!\brief Dereference operator returns element currently pointed at.
```
h-2: Maybe explicitly state here again that it copies:
`pointed at by copy (reference type is value_type).`

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
h-2: still open

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,767 @@
...
+
+        //!\brief Pre-decrement, returns updated iterator.
+        constexpr gap_decorator_anchor_set_iterator & operator--() noexcept
+        {
+            assert((host)); // host is set
+            --pos;
+
+            if (pos < left_gap_end)
+            {   // there was no gap before but we arrive at the left gap and have to update the variables.
+                if (anchor_set_it != (*host).anchors.begin())
```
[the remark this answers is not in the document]
h-2: This is also not special feature of the host variable but works for `(*prev).first` :wink:

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,782 @@
...
+        /*!\name Reference/Dereference operators
+         * \{
+        */
+        //!\brief Dereference operator returns element currently pointed at.
+        constexpr reference operator*() const noexcept
+        {
+            return current_val;
+        }
+
+        //!\brief Returns pointer to this iterator.
```
h-2: No, it doesn't :wink: 

`Returns a pointer to the currently cached value.`

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,782 @@
...
+ * entails updating all subsequent elements in the set to preserve correct cumulative sizes.
+ *
+ */
+template <std::ranges::RandomAccessRange inner_type>
+    requires std::ranges::SizedRange<inner_type>
+class gap_decorator_anchor_set
+{
+private:
+    //!\brief The iterator that moves over the seqan3::gap_decorator_anchor_set.
+    class gap_decorator_anchor_set_iterator
```
h-2: Please add a details section that contains something like the following sentence:

```
This iterator returns values when dereferenced, not references, i.e. it does not satisfy the semantic requirements of [LegacyForwardIterator](https://en.cppreference.com/w/cpp/named_req/ForwardIterator). It does model the C++20 std::BidirectionalIterator (and std::ForwardIterator implicitly).
```

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,782 @@
...
+     * ### Exceptions
+     *
+     * No-throw guarantee. Does not modify the aligned sequences.
+     */
+    constexpr friend bool operator==(gap_decorator_anchor_set const & lhs, gap_decorator_anchor_set const & rhs) noexcept
+    {
+        if (lhs.size()  == rhs.size()  &&
+            lhs.anchors == rhs.anchors &&
+            std::ranges::equal(lhs.ungapped_view, rhs.ungapped_view))
+            return true;
```
h-2: enclose in braces for readability :nail_care:

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -0,0 +1,782 @@
...
+        return false;
+    }
+
+    //!\copydoc operator==
+    constexpr friend bool operator!=(gap_decorator_anchor_set const & lhs, gap_decorator_anchor_set const & rhs) noexcept
+    {
+        return !(lhs == rhs);
+    }
+
+    friend bool operator<(gap_decorator_anchor_set const & lhs, gap_decorator_anchor_set const & rhs) noexcept
```
h-2: This isn't constexpr, but the previous one is?

smehringer:
There is one more open issue: that the free functions in aligned_sequence_concept that delegate to the member function should not be related to the concept. To whom else should they be related? ... IMO they do not belong anywhere else since they are generic overloads that make containers that have the member functions model the aligned_sequence concept?

include/seqan3/std/ranges
```diff
@@ -230,4 +233,35 @@ using SEQAN3_DOXYGEN_ONLY(iter_move =) ::ranges::iter_move;
...
+namespace ranges
+{
+
+//!\brief Type trait that indicates whether `type` is a view.
+template <typename type>
+//!\cond
+    requires true
+//!\endcond
+constexpr bool enable_view<type> = std::ranges::enable_view<type>;
+
```
h-2: I thought about this again, and in this model `std::ranges::enable_view` will not return to for local overloads of `::ranges::enable_view`. 

I think the only solution to this is:
  * declare `std::ranges::enable_view` a plain alias for `::ranges::enable_view` here
  * define `seqan3::enable_view` somewhere in our ranges module, but define it as `int` with default value of `-1`
  * in the same file overload `std::ranges::enable_view` with constraint `seqan3::enable_view == 0 || == 1` and then set it to the corresponding value
  * whenever we want to override a value, we just specify `seqan3::enable_view` to be 0 or 1.

It's not pretty, but it's the only thing that will work I think.

test/unit/alignment/aligned_sequence_test.cpp
```diff
@@ -22,19 +26,17 @@ class aligned_sequence<container_type> : public ::testing::Test
...
-    void initialize_typed_test_container(container_type & container, std::string const && target)
+    void initialise_typed_test_container(container_type & container, dna4_vector const & target)
     {
+        container.clear();
         for (auto & val : target)
         {
-            typename container_type::value_type cval{};
-            assign_char(cval, val);
-            container.push_back(cval);
+            container.push_back(typename container_type::value_type{}.assign_char(to_char(val)));
```
h-2: Don't rely on member functions in generic contexts!

test/unit/alignment/aligned_sequence_test_template.hpp
```diff
@@ -29,108 +31,186 @@ TYPED_TEST_P(aligned_sequence, fulfills_concept)
...
     EXPECT_EQ(*it, val);
-    EXPECT_EQ(aligned_seq, aligned_seq_expected);
 
     // 2) Removing a non-gap
-    TypeParam aligned_seq_fail;
-    TestFixture::initialize_typed_test_container(aligned_seq_fail, "A-CTA");
+    TestFixture::initialise_typed_test_container(aligned_seq, seq); // reset
 
-    EXPECT_THROW(erase_gap(aligned_seq_fail, begin(aligned_seq_fail) + 1, begin(aligned_seq_fail) + 3),
+    EXPECT_THROW(erase_gap(aligned_seq, std::next(begin(aligned_seq), 1), std::next(begin(aligned_seq), 3)),
```
h-2: I would recommen `std::ranges::next`, I know it's tedious. Maybe should think about giving up our own aliases and just offering `namespace seqan3::_sr` as an alias for `std::ranges`. That way we could just right `_sr::next` `_sr::begin` ... everywhere.

test/unit/alignment/aligned_sequence_test.cpp
smehringer: but it is in our test suite with only our alphabets...

test/unit/alignment/aligned_sequence_test_template.hpp
smehringer: whats wrong with std::next?

h-2: https://en.cppreference.com/w/cpp/iterator/next:
"ForwardIt must meet the requirements of LegacyForwardIterator. "

 We just established above that the exact iterator you are testing here does not. :roll_eyes:

test/unit/alignment/aligned_sequence_test.cpp
h-2: not all our alphabets do :point_up:

smehringer: :expressionless:

test/unit/alignment/aligned_sequence_test_template.hpp
smehringer: :see_no_evil:

include/seqan3/std/ranges
smehringer: :weary:

include/seqan3/std/ranges
```diff
@@ -230,4 +233,47 @@ using SEQAN3_DOXYGEN_ONLY(iter_move =) ::ranges::iter_move;
...
+namespace ranges
+{
+
+//!\brief Customises ranges type trait that indicates whether `type` is a view.
+template <typename type>
+    requires (seqan3::detail::enable_view<type> == 0) || (seqan3::detail::enable_view<type> == 1)
+constexpr bool enable_view<type> = static_cast<bool>(seqan3::detail::enable_view<type>);
+
+} // namespace ranges
```
h-2: you can delete this one

include/seqan3/std/ranges
```diff
@@ -230,4 +233,47 @@ using SEQAN3_DOXYGEN_ONLY(iter_move =) ::ranges::iter_move;
...
+
+namespace std::ranges
+{
+
+//!\brief Type trait that indicates whether `type` is a view.
+template <typename type>
+constexpr bool enable_view = ::ranges::enable_view<type>;
+
+} // namespace std::ranges
```
h-2: use an alias, don't declare a second variable (because then they can have different values).

include/seqan3/std/ranges
```diff
@@ -230,4 +233,47 @@ using SEQAN3_DOXYGEN_ONLY(iter_move =) ::ranges::iter_move;
...
+
+#if __cpp_lib_ranges
+
+namespace std::ranges
+{
+
+//!\brief Customises std::ranges type trait that indicates whether `type` is a view.
+template <typename type>
+    requires (seqan3::detail::enable_view<type> == 0) || (seqan3::detail::enable_view<type> == 1)
+constexpr bool enable_view<type> = static_cast<bool>(seqan3::detail::enable_view<type>);
```
h-2: move this to the definition of seqan3::detail::enable_view

include/seqan3/std/ranges
h-2: (the other overload will work for both cases (with and without range-v3).

smehringer:
@h-2 I had to make the comparison operators of the decorator non constexpr because GCC7 doesn't like the non-literal `auto lit = lhs.begin();`

smehringer:
@h-2 Since Jenkins cannot be trusted right now I think this is good to go and be merged.

h-2:
> Since Jenkins cannot be trusted right now

Why is Jenkins broken again? :angry:

rrahn:
Because I am playing with it. 😬

PR #682, opened 2019-01-22
[FEATURE] Adds performance test for affine alignment.

rrahn:
Adds the first benchmark test for the alignment module:

Preliminary results for score only:

```
--------------------------------------------------------------
Benchmark                       Time           CPU Iterations
--------------------------------------------------------------
seqan3_affine_dna4         613038 ns     612646 ns       1015
seqan2_affine_dna4         590937 ns     590440 ns       1169
seqan3_affine_dna4_set   59400658 ns   59376083 ns         12
seqan2_affine_dna4_set   59811930 ns   59782583 ns         12
```

### DIFF
--- a/test/performance/alignment/affine_alignment_benchmark.cpp
+++ b/test/performance/alignment/affine_alignment_benchmark.cpp
@@ -1,0 +1,169 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
+// -----------------------------------------------------------------------------------------------------
+
+#include <benchmark/benchmark.h>
+
+#include <iostream>
+#include <memory>
+#include <random>
+#include <utility>
+#include <vector>
+
+#include <seqan3/alignment/pairwise/align_pairwise.hpp>
+#include <seqan3/alphabet/aminoacid/aa20.hpp>
+#include <seqan3/alphabet/nucleotide/dna4.hpp>
+
+#if __has_include(<seqan/align.h>)
+    #define SEQAN3_HAS_SEQAN2 1
+#endif
+
+#ifdef SEQAN3_HAS_SEQAN2
+    #include <seqan/align.h>
+    #include <seqan/basic.h>
+    #include <seqan/sequence.h>
+#endif
+
+using namespace seqan3;
+
+template <typename alphabet_t>
+auto generate_sequence_seqan3(size_t const len = 500,
+                              size_t const variance = 0,
+                              size_t const seed = 0)
+{
+    std::mt19937 gen(seed);
+    std::uniform_int_distribution<uint8_t> dis_alpha(0, alphabet_size_v<alphabet_t>);
+    std::uniform_int_distribution<size_t> dis_length(len - variance, len + variance);
+
+    std::vector<alphabet_t> sequence;
+
+    size_t length = dis_length(gen);
+    for (size_t l = 0; l < length; ++l)
+        sequence.push_back(alphabet_t{}.assign_rank(dis_alpha(gen)));
+
+    return sequence;
+}
+
+#ifdef SEQAN3_HAS_SEQAN2
+template <typename alphabet_t>
+auto generate_sequence_seqan2(size_t const len = 500,
+                              size_t const variance = 0,
+                              size_t const seed = 0)
+{
+    std::mt19937 gen(seed);
+    std::uniform_int_distribution<uint8_t> dis_alpha(0, seqan::ValueSize<alphabet_t>::VALUE);
+    std::uniform_int_distribution<size_t> dis_length(len - variance, len + variance);
+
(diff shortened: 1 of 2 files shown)


test/performance/alignment/affine_alignment_benchmark.cpp
```diff
@@ -0,0 +1,167 @@
...
+// -----------------------------------------------------------------------------------------------------
+
+#include <benchmark/benchmark.h>
+
+#include <iostream>
+#include <memory>
+#include <random>
+#include <vector>
+#include <utility>
+
```
eseiler: sort lexicographically 💅

test/performance/alignment/affine_alignment_benchmark.cpp
```diff
@@ -0,0 +1,167 @@
...
+    #include <seqan/align.h>
+    #include <seqan/basic.h>
+    #include <seqan/sequence.h>
+#endif
+
+using namespace seqan3;
+
+template <typename alphabet_t>
+auto generate_sequence_seqan3(size_t const len = 500,
+                               size_t const variance = 0,
```
eseiler: align parameters

test/performance/alignment/affine_alignment_benchmark.cpp
```diff
@@ -0,0 +1,167 @@
...
+    for (size_t l = 0; l < length; ++l)
+        sequence.push_back(alphabet_t{}.assign_rank(dis_alpha(gen)));
+
+    return sequence;
+}
+
+#ifdef SEQAN3_HAS_SEQAN2
+template <typename alphabet_t>
+auto generate_sequence_seqan2(size_t const len = 500,
+                               size_t const variance = 10,
```
eseiler: For seqan3 you use `variance = 0` but here you use 10 ?

test/performance/alignment/affine_alignment_benchmark.cpp
```diff
@@ -0,0 +1,167 @@
...
+#ifdef SEQAN3_HAS_SEQAN2
+
+void seqan2_affine_dna4(benchmark::State & state)
+{
+    auto seq1 = generate_sequence_seqan2<seqan::Dna>();
+    auto seq2 = generate_sequence_seqan2<seqan::Dna>();
+
+    for (auto _ : state)
+    {
+        seqan::globalAlignmentScore(seq1, seq2, seqan::Score<int>{4, -5, -1, -11});
```
eseiler: IIRC, you need -11 here because in seqan3 the `gap_open_penalty = gap_open_score + gap_extend_score` (the textbook way(?)) while in seqan2  `gap_open_penalty = gap_open_score`.
Maybe add a comment for this.

test/performance/alignment/affine_alignment_benchmark.cpp
```diff
@@ -0,0 +1,167 @@
...
+// ============================================================================
+
+void seqan3_affine_dna4(benchmark::State & state)
+{
+    auto cfg = align_cfg::mode{align_cfg::global_alignment} |
+               align_cfg::gap{gap_scheme{gap_score{-1}, gap_open_score{-10}}} |
+               align_cfg::scoring{nucleotide_scoring_scheme{match_score{4}, mismatch_score{-5}}} |
+               align_cfg::result{align_cfg::with_score};
+
+    auto seq1 = generate_sequence_seqan3<seqan3::dna4>();
```
eseiler: I'm wondering, and this applies to all the test cases, if you call the `generate_sequence_seqan3` or `generate_sequence_seqan2` function, the generator is constructed each time with the same seed again - I'd think that would mean that `seq1 == seq2` - which kinda defies the whole point of generating random sequences.

test/performance/alignment/affine_alignment_benchmark.cpp
rrahn: 😕 I can add a comment, but I don't understand what you mean with i need -11 here. This code tests seqan2

test/performance/alignment/affine_alignment_benchmark.cpp
rrahn: That is right, also it doesn't reallt matter here. But I modify it.

test/performance/alignment/affine_alignment_benchmark.cpp
eseiler: If you'd put `-10` here, you wouldn't compute the same alignment since you'd be using a different config. 
For someone who doesn't know how SeqAn2 alignment works this might be confusing.

test/performance/alignment/affine_alignment_benchmark.cpp
```diff
@@ -0,0 +1,169 @@
...
+               align_cfg::scoring{nucleotide_scoring_scheme{match_score{4}, mismatch_score{-5}}} |
+               align_cfg::result{align_cfg::with_score};
+
+    using sequence_t = decltype(generate_sequence_seqan3<seqan3::dna4>());
+
+    std::vector<std::pair<sequence_t, sequence_t>> vec;
+    for (unsigned i = 0; i < 100; ++i)
+    {
+        sequence_t seq1 = generate_sequence_seqan3<seqan3::dna4>(500, 0, i);
+        sequence_t seq2 = generate_sequence_seqan3<seqan3::dna4>(500, 0, i + 100);
```
h-2: Can you reduce the individual sizes for the sets? Then we can get more than 6 iterations which is definetely too little for a benchmark

rrahn:
@h-2 can this be merged now?

PR #683, opened 2019-01-22
[FIX] builtin simd for macos

### LINKED ISSUE #663: mac os simd builtin test fails
On mac os the simd built in tests fail and need to be fixed.

###### estimate 8

```console
4/6 Test #4: unit/core/simd/detail/builtin_simd_test .......................***Failed    0.01 sec
Running main() from /Users/rmaerker/Development/seqan3/seqan3-build/unit/vendor/googletest/googletest/src/gtest_main.cc
[==========] Running 5 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 5 tests from builtin_simd
[ RUN      ] builtin_simd.builtin_simd
[       OK ] builtin_simd.builtin_simd (0 ms)
[ RUN      ] builtin_simd.is_builtin_simd
[       OK ] builtin_simd.is_builtin_simd (0 ms)
[ RUN      ] builtin_simd.simd_traits
[       OK ] builtin_simd.simd_traits (0 ms)
[ RUN      ] builtin_simd.default_simd_max_length
[       OK ] builtin_simd.default_simd_max_length (0 ms)
[ RUN      ] builtin_simd.simd_concept
/Users/rmaerker/Development/seqan3/seqan3-src/test/unit/core/simd/detail/builtin_simd_test.cpp:170: Failure
Value of: simd_concept<int64x2_t>
  Actual: false
Expected: true
/Users/rmaerker/Development/seqan3/seqan3-src/test/unit/core/simd/detail/builtin_simd_test.cpp:173: Failure
Value of: simd_concept<uint64x4_t>
  Actual: false
Expected: true
[  FAILED  ] builtin_simd.simd_concept (0 ms)
[----------] 5 tests from builtin_simd (0 ms total)

[----------] Global test environment tear-down
[==========] 5 tests from 1 test case ran. (0 ms total)
[  PASSED  ] 4 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] builtin_simd.simd_concept

 1 FAILED TEST
```

eseiler:
Resolves #663 

Related error message (I locally added a struct in the tests that requires simd_concept so that I get the actual constraint violations):

```console
[...]/test/unit/core/simd/detail/builtin_simd_test.cpp:141:16: note:   constraints not satisfied
In file included from [...]/test/unit/core/simd/detail/builtin_simd_test.cpp:10:
[...]/include/seqan3/core/simd/concept.hpp:37:9: note: within 'template<class simd_t> concept const bool seqan3::simd::simd_concept<simd_t> [with simd_t = __vector(2) long long int]'
 concept simd_concept = requires (simd_t a, simd_t b)
         ^~~~~~~~~~~~
[...]/include/seqan3/core/simd/concept.hpp:37:9: note:     with '__vector(2) long long int a'
[...]/include/seqan3/core/simd/concept.hpp:37:9: note:     with '__vector(2) long long int b'
In file included from[...]/include/seqan3/core/simd/concept.hpp:18,
                 from[...]/test/unit/core/simd/detail/builtin_simd_test.cpp:10:
[...]/include/seqan3/std/concepts:53:9: note: within 'template<class T, class U> concept const bool std::Same<T, U> [with T = __vector(2) long int; U = __vector(2) long long int]'
 concept Same = std::is_same_v<T, U>;
         ^~~~
[...]/include/seqan3/std/concepts:53:9: note: 'std::is_same_v' evaluated to false
```

`decltype(a == b)` is ` __vector(2) long int` while the given type is `__vector(2) long long int`.

I'm not sure if I should check if it is an integral type (will there be float simd types?).
I could add something like
```cpp
std::Integral<typename simd_traits<decltype(a == b)>::mask_type>
```

Also, the Integral requirement is already static_assert'ed for the builtin simd types.


include/seqan3/core/simd/concept.hpp
```diff
@@ -48,12 +48,12 @@ concept simd_concept = requires (simd_t a, simd_t b)
...
-    requires std::Same<decltype(a <  b), typename simd_traits<simd_t>::mask_type>;
-    requires std::Same<decltype(a >  b), typename simd_traits<simd_t>::mask_type>;
-    requires std::Same<decltype(a <= b), typename simd_traits<simd_t>::mask_type>;
-    requires std::Same<decltype(a >= b), typename simd_traits<simd_t>::mask_type>;
+    requires sizeof(decltype(a == b)) == sizeof(typename simd_traits<simd_t>::mask_type);
+    requires sizeof(decltype(a != b)) == sizeof(typename simd_traits<simd_t>::mask_type);
+    requires sizeof(decltype(a <  b)) == sizeof(typename simd_traits<simd_t>::mask_type);
+    requires sizeof(decltype(a >  b)) == sizeof(typename simd_traits<simd_t>::mask_type);
+    requires sizeof(decltype(a <= b)) == sizeof(typename simd_traits<simd_t>::mask_type);
+    requires sizeof(decltype(a >= b)) == sizeof(typename simd_traits<simd_t>::mask_type);
```
rrahn: This would not be the right fix for the problem. Testing here only the size of the types is wrong, as it could cause issues, if the mask type was defined not correctly but happens to have the same size.
To solve this we need to fix the mask_type. At the moment only gcc's builtin extensions are used and they only specify that the result of comparison results in a type of the same size and the same packing. But this does not mean it must be the same type. The mask_type should be decltype(a == b) in the definition of the simd_traits. I looked into the code and I think you can expose this in the simd_traits_helper and let simd_traits use this.  @marehr anything I am not seeing here?

eseiler: Mhm so I basically set mask_type to decltype(a == b)?

If `they only specify that the result of comparison results in a type of the same size and the same packing` I could also require that the sizeof the scalar_type and the length is the both for decltype(a==b) and simd_t, i.e. 
```cpp
requires (simd_traits<decltype(a == b)>::length == simd_traits<simd_t>::length) && 
               (sizeof(typename simd_traits<decltype(a == b)>::scalar_type) == 
                sizeof(typename simd_traits<simd_t>::scalar_type));
```

rrahn: No, we cannot require the mask type to be a vector. It could be a special mask type (introduced in AVX512). Any simd backend libary might implement a special type for this. Thus, simd_traits might not be valid. I think we need anyway a special concept for the mask type and then check if the mask type fulfills this concept, but I need to discuss this with @marehr.

PR #684, opened 2019-01-22
[MISC] Pretty Printing of our types through debug_stream.

smehringer:


test/include/seqan3/test/pretty_printing.hpp
```diff
@@ -0,0 +1,32 @@
...
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Contains an hacky overload for the googletest PrintTo function.
+ * \author Svenja Mehringer <svenja.mehringer AT fu-berlin.de>
+ */
```
joergi-w: Now we have also doxygen documentation in `test/`? I think it's useful here. Is there already a `make` target to build this?

test/include/seqan3/test/pretty_printing.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+#pragma once
+
+
```
joergi-w: double newline

test/include/seqan3/test/pretty_printing.hpp
```diff
@@ -0,0 +1,32 @@
...
+namespace seqan3
+{
+
+//!\cond DEV
+//!\brief Hacky overload for the googletest PrintTo function that always delegates to our debug_stream.
+template <typename t>
+    requires true
```
joergi-w: Why this "constraint"?

test/include/seqan3/test/pretty_printing.hpp
```diff
@@ -0,0 +1,32 @@
...
+void PrintTo (t const & v, std::ostream * out)
+{
+    std::ostringstream o;
+    debug_stream_type my_stream{o};
+    my_stream << v;
+    o.flush();
+    *out << o.str();
+}
+//!\endcond
+}
```
joergi-w: add `// namespace seqan3` (seems to me like a convention in our code for closing namespaces)

test/include/seqan3/test/pretty_printing.hpp
```diff
@@ -0,0 +1,32 @@
...
+template <typename t>
+    requires true
+void PrintTo (t const & v, std::ostream * out)
+{
+    std::ostringstream o;
+    debug_stream_type my_stream{o};
+    my_stream << v;
+    o.flush();
+    *out << o.str();
+}
```
h-2: Why the extra string stream, why not create the debug_stream from `out`?

test/include/seqan3/test/pretty_printing.hpp
eseiler: For the stuff in `test/include` we do have docs

joergi-w: Oh, good!

test/include/seqan3/test/pretty_printing.hpp
joergi-w: Okay – that's a clever way of overloading. And it works even if this function lives in seqan3 namespace?

eseiler: I'm not that familiar with the look-up rules, but [in the docs of googletest](https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#teaching-googletest-how-to-print-your-values) they also write that it needs to be in the same namespace as the type you want to print. (In the seconds snippet where they declare `printTo` as global function).

Though I think it only applies if you specialise it for a single type.
It will actually find both `printTo` functions (in `googletest` and `seqan3` namespace) and complain about the ambiguity if you don't use the constraint.

joergi-w: Thanks! Then let's just add a (non-doxy) comment that this line must not be removed. Otherwise, someone might wonder about the same questions again.

smehringer:
code coverage says I did not test this... duh. I see no way to test this as it is called only when a test fails :roll_eyes:

joergi-w:
You can explicitly call `seqan3::PrintTo()` from a cpp file...

PR #688, opened 2019-01-25
[FIX] Implement the --copyright flag, along with appropriate test cases.

### LINKED ISSUE #687: Flag --copyright does not exist, even though it's referenced in the code.
Resolve by adding a --copyright flag which prints the right copyright/license info.

joshuak94:
Resolves #687 by creating a format_copyright and the appropriate tests.

### DIFF
--- a/include/seqan3/argument_parser/detail/format_help.hpp
+++ b/include/seqan3/argument_parser/detail/format_help.hpp
@@ -765,4 +765,76 @@ public:
     }
 };
 
+/*!\brief The format that prints the copyright information to std::cout.
+ * \ingroup argument_parser
+ *
+ * \details
+ *
+ * The copyright message printing is not done immediately, because the user cannot provide
+ * meta information (e.g. long_copyright) on construction of the parser. Thus the meta information is collected
+ * and only evaluated when calling seqan3::format_version::parse.
+ */
+class format_copyright : public format_help
+{
+public:
+    /*!\brief Initiates the printing of the copyright message to std::cout.
+     * \param[in] parser_meta The meta information that are needed for a detailed version information.
+     */
+    void parse(argument_parser_meta_data const & parser_meta)
+    {
+        meta = parser_meta;
+        std::string seqan_license{
+        R"(Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of Knut Reinert or the FU Berlin nor the names of
+      its contributors may be used to endorse or promote products derived
+      from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ARE DISCLAIMED. IN NO EVENT SHALL KNUT REINERT OR THE FU BERLIN BE LIABLE
+FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+DAMAGE.)"};
+
+        std::cout << std::string(80, '=') << to_text("\n\\fB") << "Copyright information for "
+                  << meta.app_name << ":\n" << to_text("\\fP") << std::string(80, '-') << '\n';
+
+        if (!empty(meta.long_copyright))
+        {
+            std::cout << to_text("\\fP") << meta.long_copyright << "\n\n";
+        }
+        else if (!empty(meta.short_copyright))
+        {
+            std::cout << to_text("\\fP") << meta.app_name << " full copyright information not available. Displaying"
+                      << " short copyright information instead:\n" << to_text("\\fP") << meta.short_copyright << "\n\n";
+        }
+        else
+        {
+            std::cout << to_text("\\fP") << meta.app_name << " copyright information not available.\n\n";
+        }
+
+        std::cout << std::string(80, '=') << to_text("\n\\fB")
+                  << "This program contains SeqAn3 code licensed under the following terms:\n" << to_text("\\fP")
+                  << std::string(80, '-') << '\n' << seqan_license << '\n';
+
+        throw parser_interruption();
+    }
+};
+
 } // namespace seqan3::detail
--- a/test/unit/argument_parser/detail/CMakeLists.txt
+++ b/test/unit/argument_parser/detail/CMakeLists.txt
@@ -1,2 +1,3 @@
+add_definitions(-DSEQAN_INCLUDE_DIR="${SEQAN3_INCLUDE_DIR}")
 seqan3_test(format_help_test.cpp)
 seqan3_test(format_html_test.cpp)
--- a/test/unit/argument_parser/detail/format_help_test.cpp
+++ b/test/unit/argument_parser/detail/format_help_test.cpp
@@ -224,34 +224,10 @@ TEST(test_add_copyright, copyright)
     // Tests the --copyright call.
     const char * argvCopyright[] = {"./copyright", "--copyright"};
     argument_parser copyright("myApp", 2, argvCopyright);
-    std::string seqan_license = std::string(
-        "Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin"
-        "Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik"
-        "All rights reserved."
 
-        "Redistribution and use in source and binary forms, with or without"
-        "modification, are permitted provided that the following conditions are met:"
-
-        "   * Redistributions of source code must retain the above copyright"
-        "     notice, this list of conditions and the following disclaimer."
-        "   * Redistributions in binary form must reproduce the above copyright"
-        "     notice, this list of conditions and the following disclaimer in the"
-        "     documentation and/or other materials provided with the distribution."
-        "   * Neither the name of Knut Reinert or the FU Berlin nor the names of"
-        "     its contributors may be used to endorse or promote products derived"
-        "     from this software without specific prior written permission."
-
-        "THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS \"AS IS\""
-        "AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE"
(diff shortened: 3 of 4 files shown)


include/seqan3/argument_parser/detail/format_help.hpp
```diff
@@ -764,4 +765,84 @@ class format_version : public format_help
...
+            std::cout << to_text("\\fP") << meta.short_copyright << "\n\n";
+        }
+        else
+            std::cout << to_text("\\fP") << meta.app_name << "copyright information not available.\n\n";
+
+        std::cout << to_text("\\fB") << "SeqAn3 Copyright:\n";
+        std::string line;
+        while (std::getline(seqan_license, line))
+        {
+            std::fill_n(out, layout.leftPadding, ' ');
```
smehringer: I am pretty sure that for this line and the one below you can just use the print_line function of the format
```cpp
print_line(line);
```
Maybe you need `print_line(line, false);`.

joshuak94: I tried it, but with `print_line(line)` it treats every line as a paragraph and makes the license info huge. But with `print_line(line, false)` it treats the whole thing as one paragraph and removes the spacing in between... So maybe it's better to just keep it as is?

include/seqan3/argument_parser/detail/format_help.hpp
```diff
@@ -764,4 +765,84 @@ class format_version : public format_help
...
+ARE DISCLAIMED. IN NO EVENT SHALL KNUT REINERT OR THE FU BERLIN BE LIABLE
+FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+DAMAGE.)"};
+
+        print_header();
```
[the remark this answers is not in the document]
joshuak94: Would it be better to give a width of `layout.screenWidth`?

joshuak94:
@smehringer Implemented the license file.

test/unit/argument_parser/detail/format_help_test.cpp
```diff
@@ -105,16 +105,67 @@ TEST(help_add_test, add_option)
...
-    // const char * argvCopyright[] = {"./copyright", "--copyright"};
-    // argument_parser copyright("copyright", 2, argvCopyright);
-    // copyright.info.long_copyright = "long copyright";
-    // testing::internal::CaptureStdout();
-    // EXPECT_THROW(copyright.parse(), parser_interruption);
-    // stdout = testing::internal::GetCapturedStdout();
-    // expected = "whatever is expected";
-    // EXPECT_TRUE(ranges::equal((stdout   | ranges::view::remove_if(is_space)),
-                               // expected | ranges::view::remove_if(is_space)));
+    // Tests the --copyright call.
```
smehringer: Please put this into its own test case instead of adding it to the `TEST(help_add_test, add_option)` test.

test/unit/argument_parser/detail/format_help_test.cpp
```diff
@@ -105,16 +105,67 @@ TEST(help_add_test, add_option)
...
+                               "--------------------------------------------------------------------------------"
+                               "copyright copyright information not available."
+                               "================================================================================"
+                               "This program contains SeqAn3 code licensed under the following terms:"
+                               "--------------------------------------------------------------------------------"
+                               + license_string);
+        EXPECT_TRUE(ranges::equal((stdout   | ranges::view::remove_if(is_space)),
+                                   expected | ranges::view::remove_if(is_space)));
+    }
+    copyright.info.short_copyright = "short copyright\nshortcopyright";
```
smehringer: swap this with line below so the explanation comes first and introduce a newline between the tests. :nail_care:  makes reading easier

joshuak94:
@smehringer Updated!

test/unit/argument_parser/detail/format_help_test.cpp
```diff
@@ -105,31 +107,20 @@ TEST(help_add_test, add_option)
...
-
     // Empty help call with -hh
     const char * argv2[] = {"./help_add_test", "-hh"};
     argument_parser parser2("test_parser_2", 2, argv2);
     testing::internal::CaptureStdout();
     EXPECT_THROW(parser2.parse(), parser_interruption);
     stdout = testing::internal::GetCapturedStdout();
     expected = std::string("test_parser_2"
-                           "============="
-                           "VERSION"
```
smehringer: remove this change again :nail_care:

test/unit/argument_parser/detail/format_help_test.cpp
```diff
@@ -138,13 +129,13 @@ TEST(help_add_test, add_option)
     EXPECT_THROW(parser3.parse(), parser_interruption);
     stdout = testing::internal::GetCapturedStdout();
     expected = std::string("version"
-                           "======="
-                           "VERSION"
-                           "Last update:"
```
smehringer: here too

test/unit/argument_parser/detail/format_help_test.cpp
```diff
@@ -156,15 +147,15 @@ TEST(help_add_test, add_option)
     EXPECT_THROW(parser4.parse(), parser_interruption);
     stdout = testing::internal::GetCapturedStdout();
     expected = std::string("versionURL"
-                           "=========="
-                           "VERSION"
-                           "Last update:"
-                           "versionURL version:"
```
smehringer: here too :nerd_face:

test/unit/argument_parser/detail/format_help_test.cpp
```diff
@@ -174,14 +165,14 @@ TEST(help_add_test, add_option)
     EXPECT_THROW(parser5.parse(), parser_interruption);
     stdout = testing::internal::GetCapturedStdout();
     expected = std::string("hidden"
-                           "======"
-                           "OPTIONS"
-                           "VERSION"
-                           "Last update:"
```
smehringer: aaaaand here

test/unit/argument_parser/detail/format_help_test.cpp
```diff
@@ -199,30 +190,96 @@ TEST(help_add_test, add_option)
...
-                           "full synopsis"
-                           "full synopsis2"
-                           "DESCRIPTION"
-                           "description"
-                           "description2"
-                           "POSITIONAL ARGUMENTS"
-                           "ARGUMENT-1 List of STRING's"
-                           "this is a positional option."
-                           "OPTIONS"
-                           "-i, --int INT (32 bit)"
```
smehringer: and since i commented everywhere.. here too :grin:

joshuak94:
@smehringer **now** everything should be okay!

test/unit/argument_parser/detail/format_help_test.cpp
```diff
@@ -223,6 +214,72 @@ TEST(help_add_test, add_option)
...
+                                std::istreambuf_iterator<char>()};
+    // Test --copyright with empty short and long copyright info.
+    {
+        testing::internal::CaptureStdout();
+        EXPECT_THROW(copyright.parse(), parser_interruption);
+        stdout = testing::internal::GetCapturedStdout();
+        expected = std::string("================================================================================"
+                               "Copyright information for copyright:"
+                               "--------------------------------------------------------------------------------"
+                               "copyright copyright information not available."
```
smehringer: there is a double copyright? If this is the app name please change the app name to something like `MyApp` so it is clearer and also I would propose a `:`
-> `MayApp: copyright information not available.`

test/unit/argument_parser/detail/format_help_test.cpp
```diff
@@ -223,6 +214,72 @@ TEST(help_add_test, add_option)
...
+    {
+        testing::internal::CaptureStdout();
+        EXPECT_THROW(copyright.parse(), parser_interruption);
+        stdout = testing::internal::GetCapturedStdout();
+        expected = std::string("================================================================================"
+                               "Copyright information for copyright:"
+                               "--------------------------------------------------------------------------------"
+                               "copyright full copyright information not available. Displaying short copyright information instead:"
+                               "short copyright"
+                               "short copyright"
```
smehringer: double short copy right ? ah I see that this is intentional. can you change it to 
`short copyright line 1\n short copy right line 2`? otherwise people (like me) might think this is an error.

test/unit/argument_parser/detail/format_help_test.cpp
```diff
@@ -223,6 +214,72 @@ TEST(help_add_test, add_option)
...
+    // Test --copyright with a non-empty short copyright and a non-empty long copyright.
+    copyright.info.long_copyright = "long copyright\nlong copyright";
+    {
+        testing::internal::CaptureStdout();
+        EXPECT_THROW(copyright.parse(), parser_interruption);
+        stdout = testing::internal::GetCapturedStdout();
+        expected = std::string("================================================================================"
+                               "Copyright information for copyright:"
+                               "--------------------------------------------------------------------------------"
+                               "copyright Copyright:"
```
smehringer: actually why do you have `copyright Copyright:` here if in the line above you already have `Copyright information for copyright:"`? isn't that redundant?

joshuak94: Good point. Removed it!

test/unit/argument_parser/detail/format_help_test.cpp
```diff
@@ -223,6 +214,71 @@ TEST(help_add_test, add_option)
...
+}
+
+TEST(test_add_copyright, copyright)
+{
+    std::string stdout;
+    std::string expected;
+
+    // Tests the --copyright call.
+    const char * argvCopyright[] = {"./copyright", "--copyright"};
+    argument_parser copyright("copyright", 2, argvCopyright);
```
smehringer:
```suggestion
    argument_parser copyright("myApp", 2, argvCopyright);
```
thats probably why your test fails

joshuak94: Whoops, was a weird thing that happened while rebasing. Thanks!

smehringer:
@h-2 review ping

test/unit/argument_parser/detail/CMakeLists.txt
```diff
@@ -1,2 +1,3 @@
+add_definitions(-DSEQAN_INCLUDE_DIR="${SEQAN3_INCLUDE_DIR}")
```
h-2: why was this change necessary?

test/unit/argument_parser/detail/format_help_test.cpp
```diff
@@ -45,11 +47,11 @@ TEST(help_add_test, add_option)
...
-                                       "==========="
-                                       "VERSION"
-                                       "Last update:"
-                                       "test_parser version:"
-                                       "SeqAn version: 3.0.0");
+                           "==========="
+                           "VERSION"
+                           "Last update:"
+                           "test_parser version:"
+                           "SeqAn version: 3.0.0");
```
h-2: include the actual seqan3 version from seqan3/version.hpp instead of hard-coding

test/unit/argument_parser/detail/format_help_test.cpp
```diff
@@ -223,6 +214,71 @@ TEST(help_add_test, add_option)
                            "SeqAn version: 3.0.0");
     EXPECT_TRUE(ranges::equal((stdout   | ranges::view::remove_if(is_space)),
                                expected | ranges::view::remove_if(is_space)));
```
h-2: use view::filter(!...) instead view::remove_if(...) because the latter is deprecated.

test/unit/argument_parser/detail/format_help_test.cpp
```diff
@@ -223,6 +214,71 @@ TEST(help_add_test, add_option)
...
-   // EXPECT_THROW(parser6.parse(), parser_interruption);
+    std::ifstream license_file{std::string{{SEQAN_INCLUDE_DIR}} + "/../LICENSE"};
+    std::string license_string{(std::istreambuf_iterator<char>(license_file)),
+                                std::istreambuf_iterator<char>()};
+    // Test --copyright with empty short and long copyright info.
+    {
+        testing::internal::CaptureStdout();
+        EXPECT_THROW(copyright.parse(), parser_interruption);
+        stdout = testing::internal::GetCapturedStdout();
+        expected = std::string("================================================================================"
```
h-2: why is this wrapped in `std::string()`?

h-2: also in other places

test/unit/argument_parser/detail/CMakeLists.txt
smehringer: see the commit message. Because we want to read the licence file (and therefore we need the repo path) in order to compare it to the hard-coded copyright printing and keep things updated.

PR #696, opened 2019-01-28
[FIX] Make charconv functions inline

eseiler:
Fixes linkage for these two functions.
(We didn't notice because we don't use these functions yet, but we have PRs that'll use them)

### DIFF
--- a/include/seqan3/std/charconv
+++ b/include/seqan3/std/charconv
@@ -109,7 +109,7 @@ void from_chars(char const *, char const *, bool, int = 10) = delete;
  * \sa https://en.cppreference.com/w/cpp/utility/to_chars
  */
 template <std::Integral value_type>
-inline std::to_chars_result to_chars(char * first, char * last, value_type value, int base)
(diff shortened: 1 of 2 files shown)


eseiler:
I made them noexcept.
I reran the charconv benchmark for `from_char`. (100 repetitions, before = without noexcept, after = with noexcept)

|              Test                |   before  |   after   |
|----------------------------------|-----------|-----------|
| from_char<int8_t>_mean           | 193752101 | 190782481 |
| from_char<uint8_t>_mean          | 244282055 | 239220551 |
| from_char<int16_t>_mean          |  95139719 | 103068496 |
| from_char<uint16_t>_mean         | 122070312 | 128804328 |
| from_char<int32_t>_mean          |  98998699 |  99577507 |
| from_char<uint32_t>_mean         | 121025606 | 120923162 |
| from_char<int64_t>_mean          |  72476523 |  84005376 |
| from_char<uint64_t>_mean         | 100940186 | 100691897 |
| from_chars_to_float<float>_mean  |   9199269 |   9198665 |
| from_chars_to_float<double>_mean |  21673907 |  22162559 |

PR #707, opened 2019-01-29
[INFRA] Exclude std and contrib from coverage tests

### LINKED ISSUE #697: Exclude std and contrib module from coverage

eseiler:
Resolves #697

### DIFF
--- a/test/coverage/CMakeLists.txt
+++ b/test/coverage/CMakeLists.txt
@@ -24,6 +24,8 @@ endif ()
 # Files which should not be included in the coverage report
 set (TEST_COVERAGE_EXCLUDE_FILES
     "'/usr/*'"
+    "'${SEQAN3_CLONE_DIR}/include/seqan3/contrib/*'"
+    "'${SEQAN3_CLONE_DIR}/include/seqan3/std/*'"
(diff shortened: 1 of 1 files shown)


rrahn:
@h-2 this has nothing to do with the coverage tests. Every then and now, they are not passing because of timeouts. Triggering the build again mostly resolves this.

h-2:
Put this PR does in fact touch coverage builds, that's why I'd like for the test to pass before merging it. Especially since the coverage test in Jenkins is fake (can you fix that with the other changes?).

rrahn:
For sure, the test should go through. But I was just saying, that I experienced it to happen quite often lately. What do you mean with `is fake`?

h-2:
It says it does coverage tests, but it doesn't.

I recommended @eseiler do a rebase so the coverage tests move up in priority, but that's not the problem here right now.

rrahn:
Ah in Jenkins. 🤦‍♂️  Sorry I misread this. Yes, they will be added there as well. Maybe a better place for the coverage tests anyway.

eseiler:
I rebased it.
I also noticed that sometimes the coverage tests time out and then just restart them.

PR #668, opened 2019-01-11
[MISC] Update single_pass_input_view so that there is no more view_struct, and instead it is all directly in the class.

joshuak94:

PR #710, opened 2019-01-30
[FIX] Restrict PrinTo overload to objects that are streamable with our debug_stream

smehringer:
upsi

### DIFF
--- a/test/include/seqan3/test/pretty_printing.hpp
+++ b/test/include/seqan3/test/pretty_printing.hpp
@@ -20,8 +20,16 @@ namespace seqan3
 //!\cond DEV
 //!\brief Overload for the googletest PrintTo function that always delegates to our debug_stream.
 template <typename t>
-    requires true // tricks the compiler to consider this as more specialized than googletests generic PrintTo
+    requires requires (t const & v) { {debug_stream << v}; }
 void PrintTo (t const & v, std::ostream * out)
+{
+    debug_stream_type my_stream{*out};
+    my_stream << v;
+}
+
+template <typename t>
+    requires requires (t && v) { {debug_stream << v}; }
+void PrintTo (t && v, std::ostream * out)
 {
     debug_stream_type my_stream{*out};
     my_stream << v;
--- a/include/seqan3/io/stream/debug_stream.hpp
+++ b/include/seqan3/io/stream/debug_stream.hpp
@@ -125,6 +125,7 @@ public:
      */
     //!\brief Forwards to the underlying stream object.
     template <typename t>
+        requires requires (t & v, std::ostream * out) { *out << v; }
     debug_stream_type & operator<<(t && v)
     {
         *stream << v;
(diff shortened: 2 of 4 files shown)


smehringer:
@rrahn Can we merge this ASAP :pray:

smehringer:
dammit

test/include/seqan3/test/pretty_printing.hpp
```diff
@@ -20,7 +20,7 @@ namespace seqan3
 //!\cond DEV
 //!\brief Overload for the googletest PrintTo function that always delegates to our debug_stream.
 template <typename t>
-    requires true // tricks the compiler to consider this as more specialized than googletests generic PrintTo
+    requires requires (t v) { debug_stream << v; }
```
rrahn:
```suggestion
    requires requires (t & v) { debug_stream << v; }
```

smehringer:
ok please have another look.

test/include/seqan3/test/pretty_printing.hpp
```diff
@@ -20,8 +20,8 @@ namespace seqan3
 //!\cond DEV
 //!\brief Overload for the googletest PrintTo function that always delegates to our debug_stream.
 template <typename t>
-    requires true // tricks the compiler to consider this as more specialized than googletests generic PrintTo
-void PrintTo (t const & v, std::ostream * out)
+    requires requires (t && v) { debug_stream << v; }
```
rrahn: I think, if you want to test that both l-value and r-values can be put to the debug_stream you'll need to change  `debug_stream << std::forward<t>(v);` as well.

smehringer: I just wanted to capture `v` by `&&` so I don't add a const qualifier before passing it to the debug_stream. Since within PrinTo `v` is always an l-value I do not need to test for debug-streaming r-values do I?

rrahn: I don't know the signature of PrintTo from gtest though

smehringer: The signature from gtest has const & in it... so you might be right. I'll check this tomorrow morning

smehringer:
Blocked by [sdsl fix](https://github.com/xxsds/sdsl-lite/pull/59)

smehringer:
re oben to rebuild tests

eseiler:
Don't you need to push the submodule update for the tests to build? 🤔

smehringer:
There is still an ICE when printing `debug_stream << v | view::single_pass_input | view::take_line;`
waiting for #668 to be merged before tackling the issue as it might be resolved then.

rrahn:
@smehringer is there any news about the ICE? Is this already reported?

smehringer:
@rrahn I opened [an issue](https://github.com/ericniebler/range-v3/issues/1105) in the ranges library with no response as of now. I will file a bug report on GCC.

smehringer:
This can be reevaluated when we talk about how we will handling pretty printing

PR #711, opened 2019-02-01
[feature] auto link std:: to cppreference

marehr:


h-2:
Thanks for working on this on short notice!

PR #717, opened 2019-02-04
[Fix] alphabet proxy

smehringer:


include/seqan3/alphabet/detail/alphabet_proxy.hpp
```diff
@@ -249,14 +249,14 @@ class alphabet_proxy : public alphabet_base<derived_type,
         requires alphabet_concept<alphabet_type>
     {
```
h-2: Maybe add a one-line code comment that explains the necessity?

rrahn: Sorry for jumping in, but could you please explain how/why an additional constructor call is avoided?

smehringer: Sure, unfortunately I could not reproduce the error in a test case since some compiler wodoo lets it only occur when including `concatenated_sequences.hpp` and then using `get<1>(seq_qual_alph)` which should return the quality of the qualified composition.
The problem: The cartesian composition qualified returns a component_proxy which inherits from alphabet base. The qualified alphabet itself inherits from quality_base.
Now when accessing `get<1>(seq_qual_alph)` we want to call `to_phred` at some point because we want the quality, therefore the `to_phred` function from alphabet_proxy is called, but this function did a static_cast to the derived type which is calling the constructor from quality_base. Unfortunately now, the generic quality_base constructor uses `assign_phred(static_cast<derived_type &>(*this), to_phred(other));` ([here](https://github.com/seqan/seqan3/blob/master/include/seqan3/alphabet/quality/quality_base.hpp#L85)) which again tries to call to_phred of the alphabet_proxy => **infinite loop** :boom:

rrahn: Mhmm, ok. So there is somewhere a constructor that constructs the type from an alphabet_proxy, which is competing with the conversion operator of the proxy class? Sounds like fun.  But maybe you use exactly this explanation as a comment.

smehringer: All of it? :boom:

rrahn: preferrably, yes

PR #719, opened 2019-02-05
fix GCC9 part 2

h-2:
Enables building with GCC9 support as of this weeks snapshot, but only in c++17 mode. C++20 will be supported as soon as the respective errors are fixed in GCC.


h-2:
@rrahn It's fairly small!

h-2:
> Can you explain, why u8 prefix is obsolete now?

It's not obsolete, the problem is that in C++20 `u8' '` is of type char8_t and not char. They are distinct types which causes all sorts of problems when combined with each other.

rrahn:
Ok, but it feels like not solving the origin of the problem. As far as I know, the matrix printing facility should be completely self-contained and not interfere with anything else of the library. 

I am not an expert on encodings, but after some reading it seems that everything that can handle byte sequences should be able to handle utf-8 encoded byte sequences. So this change seems legit.

PR #724, opened 2019-02-08
Adding std::ranges::next

MitraDarja:
Alias for std::ranges::next


rrahn:
Hi, thanks for your work. Unfortunately, this will be already included by #712 which is going to be merged in the next couple of minutes. Also note, that these are iterator operations and are put int the `std/iterator` header according to the current standard draft.

smehringer:
@rrahn this was 27 minutes ago :p I need the alias too... If this might take longer I think the small PR should be favoured and you can rebase afterwards.

eseiler:
Resolved by #712 🎉

PR #732, opened 2019-02-12
performance units in benchmarks

marehr:
This introduces convenient functions to represent common measurable units in google/benchmark.

An example is bytes/s, which is introduced in this PR.  We introduce this, because SetBytesProcessed was marked as deprecated in favour for a more general benchmark::Counter approach. (see https://github.com/google/benchmark/pull/654)

In a later PR I will introduce CUPS (cell updates per second) for alignment benchmarks as another common benchmark unit.

### DIFF
--- a/test/include/seqan3/test/performance/units.hpp
+++ b/test/include/seqan3/test/performance/units.hpp
@@ -19,8 +19,7 @@
 namespace seqan3::test
 {
 
-/**
- * This returns a counter which represents how many bytes were processed per second.
+/*!\brief This returns a counter which represents how many bytes were processed per second.
  *
  * \param  bytes The total number of bytes processed of a complete benchmark run.
  * \return       Returns a benchmark Counter which represents bytes/s.
--- a/test/include/seqan3/test/performance/units.hpp
+++ b/test/include/seqan3/test/performance/units.hpp
@@ -1,0 +1,35 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Contains predefined custom units for google benchmark.
+ * \author Marcel Ehrhardt <marcel.ehrhardt AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <benchmark/benchmark.h>
+
+#include <seqan3/core/platform.hpp>
+
+namespace seqan3::test
+{
+
+/**
+ * This returns a counter which represents how many bytes were processed per second.
+ *
+ * \param  bytes The total number of bytes processed of a complete benchmark run.
+ * \return       Returns a benchmark Counter which represents bytes/s.
+ */
+inline benchmark::Counter bytes_per_second(size_t bytes)
+{
+    return benchmark::Counter(bytes,
+                              benchmark::Counter::kIsIterationInvariantRate,
+                              benchmark::Counter::OneK::kIs1024);
+}
+
+} // namespace seqan3::test
--- a/test/performance/example/example_benchmark.cpp
+++ b/test/performance/example/example_benchmark.cpp
@@ -10,6 +10,10 @@
 
 #include <benchmark/benchmark.h>
 
+#include <seqan3/test/performance/units.hpp>
+
+using namespace seqan3::test;
+
 static void vector_copy_benchmark(benchmark::State& state) {
     std::vector<int> x = {15, 13, 12, 10};
     for (auto _ : state)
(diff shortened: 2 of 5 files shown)


test/performance/example/example_benchmark.cpp
```diff
@@ -25,8 +29,7 @@ static void memcpy_benchmark(benchmark::State& state) {
     for (auto _ : state)
         memcpy(dst, src, size);
 
-    int64_t bytes = int64_t(state.iterations()) * int64_t(size);
-    state.counters["bytes_processed"] = bytes;
+    state.counters["bytes_per_second"] = bytes_per_second(state.range(0));
```
eseiler:
```suggestion
    state.counters["bytes_per_second"] = bytes_per_second(size);
```

?

test/performance/example/example_benchmark.cpp
```diff
@@ -40,8 +43,7 @@ static void copy_benchmark(benchmark::State& state) {
     for (auto _ : state)
         std::copy_n(src, size, dst);
 
-    int64_t bytes = int64_t(state.iterations()) * int64_t(size);
-    state.counters["bytes_processed"] = bytes;
+    state.counters["bytes_per_second"] = bytes_per_second(state.range(0));
```
eseiler:
```suggestion
    state.counters["bytes_per_second"] = bytes_per_second(size);
```

test/seqan3-test.cmake
```diff
@@ -126,7 +126,7 @@ macro (seqan3_require_benchmark)
         gbenchmark_project
         PREFIX gbenchmark_project
         GIT_REPOSITORY "https://github.com/google/benchmark.git"
-        GIT_TAG "v1.4.1"
+        GIT_TAG "b8ca0c42179b7b5d656494e61dda8b861057122f"
```
h-2: Why are switching away from a release?

test/include/seqan3/test/performance/units.hpp
```diff
@@ -0,0 +1,33 @@
...
+namespace seqan3::test
+{
+
+/**
+ * This returns a counter which represents how many bytes were processed per second.
+ *
+ * \param  bytes The total number of bytes processed of a complete benchmark run.
+ * \return       Returns a benchmark Counter which represents bytes/s.
+ */
+inline benchmark::Counter bytes_per_second(size_t bytes)
```
h-2: wouldn't KiB/s maybe be more readable?

marehr: If a number is big enough it will automatically displayed as K/s, M/s, G/s.

test/seqan3-test.cmake
marehr: `benchmark::Counter::kIsIterationInvariantRate` is only available in a more recent version (from August 2018, last release was May 2018).

I asked nicely for a new version number. :angel:

marehr:
> I would propose to put the `units.hpp` header into `test/include` like the other helper files we have for tests now.

@h-2 Your opinion? I have no hard feeling for the file location.

marehr:
@h-2, @eseiler, @smehringer  can you re-review?

test/performance/io/format_fasta_benchmark.cpp
```diff
@@ -42,6 +44,12 @@ void write3(benchmark::State & state)
     {
         format.write(ostream, options, seq, id, std::ignore);
     }
+
+    ostream = std::ostringstream{};
```
smehringer: why do you need to write to a flushed sotream again? can't you just use `ostream.size()` because we have already written to it?

test/performance/io/format_fasta_benchmark.cpp
```diff
@@ -42,6 +44,12 @@ void write3(benchmark::State & state)
...
+    format.write(ostream, options, seq, id, std::ignore);
+    size_t bytes_per_run = ostream.str().size();
+    state.counters["bytes_per_run"] = bytes_per_run;
```
smehringer: I'd change three lines to two in each of the four benchmarks:
```diff
- size_t bytes_per_run = ostream.str().size();
- state.counters["bytes_per_run"] = bytes_per_run;
- state.counters["bytes_per_second"] = bytes_per_second(bytes_per_run);
+ state.counters["bytes_per_run"] = ostream.str().size();
+ state.counters["bytes_per_second"] = bytes_per_second(state.counters["bytes_per_run"]);
```

test/include/seqan3/test/performance/units.hpp
```diff
@@ -0,0 +1,34 @@
...
+{
+    return benchmark::Counter(bytes,
+                              benchmark::Counter::kIsIterationInvariantRate,
+                              benchmark::Counter::OneK::kIs1024);
+}
+
+} // namespace seqan3::test
```
h-2: So every thing this file does, one could also achieve by just writing this single  line into the benchmark:
`benchmark::Counter(bytes, benchmark::Counter::kIsIterationInvariantRate, benchmark::Counter::OneK::kIs1024);`
?

test/performance/io/format_fasta_benchmark.cpp
```diff
@@ -76,9 +90,13 @@ void read3(benchmark::State & state)
     for (auto _ : state)
     {
         format.read(istream, options, seq, id, std::ignore);
-	    id.clear();
-	    seq.clear();
+        id.clear();
+        seq.clear();
     }
+
+    size_t bytes_per_run = dummy_file.size();
```
h-2: you can't assume the entire file was read. In fact, it shouldn't be!

test/seqan3-test.cmake
rrahn: and?

test/performance/io/format_fasta_benchmark.cpp
rrahn: I don't know if this is really relevant.

h-2:
LGTM

can be merged if the changes to the fasta benchmark are transparent

marehr:
@h-2 ping ping

PR #737, opened 2019-02-14
[TEST] Add all cpp file in seqan3/doc/ to snippet tests.

smehringer:
Is this ok?
It works for me and creates targets "[cpp_file_name]_snippet" and the binaries are stored in the directory structure of /doc/. E.g. `doc/tutorial/arg_parse/sol1.cpp` adds the target `sol1_snippet` and the binary appears in the current build directory under `tutorial/arg_parse/sol1_snippet`.

### DIFF
--- a/test/snippet/CMakeLists.txt
+++ b/test/snippet/CMakeLists.txt
@@ -19,18 +19,18 @@ macro (seqan3_snippet target_cpp snippet_full_path)
     # NOTE(marehr): ".+/test/" REGEX is greedy, that means
     # /test/test/test/hello_snippet.cpp will result in an empty `test_path`
     string (REGEX REPLACE ".cpp$" "" target_name ${target_cpp})
-    string (REGEX REPLACE ".+/test/" "" test_path ${snippet_full_path})
-    string (REGEX REPLACE ".+/test/snippet/" "" source_path ${snippet_full_path})
+    string (REGEX REPLACE ".+/(doc|test)/" "" test_path ${snippet_full_path})
+    string (REGEX REPLACE ".+/(doc|test/snippet)/" "" source_path ${snippet_full_path})
     set (target "${target_name}_snippet")
 
-    add_executable (${target} ${source_path}/${target_cpp})
+    add_executable (${target} ${snippet_full_path}/${target_cpp})
     target_link_libraries (${target} seqan3::test::unit)
     set_target_properties(${target}
         PROPERTIES
         RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/${source_path}"
     )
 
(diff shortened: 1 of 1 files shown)


test/snippet/CMakeLists.txt
```diff
@@ -37,14 +37,41 @@ macro (seqan3_snippet target_cpp snippet_full_path)
...
+    set_target_properties(${target}
+        PROPERTIES
+        RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/${source_path}"
+    )
+
+    add_test (NAME "${snippet_full_path}/${target}" COMMAND ${target})
+
+    unset (target_name)
+    unset (target)
+endmacro ()
```
marehr: why not combine `seqan3_doc_snippet` and `seqan3_snippet`?

test/snippet/CMakeLists.txt
```diff
@@ -37,14 +37,41 @@ macro (seqan3_snippet target_cpp snippet_full_path)
...
     seqan3_snippet("${target_cpp}" "${snippet_full_path}")
 endforeach()
+
+# add all snippets in doc/
+file(GLOB_RECURSE doc_snippets "../../doc/*.cpp")
+foreach(doc_snippet ${doc_snippets})
+    get_filename_component(target_cpp "${doc_snippet}" NAME)
+    get_filename_component(snippet_full_path "${doc_snippet}" DIRECTORY)
+
+    seqan3_doc_snippet("${target_cpp}" "${snippet_full_path}")
```
marehr:
```suggestion
    seqan3_snippet("${target_cpp}" "${snippet_full_path}")
```

test/snippet/CMakeLists.txt
```diff
@@ -37,14 +37,41 @@ macro (seqan3_snippet target_cpp snippet_full_path)
     unset (target)
 endmacro ()
 
+macro (seqan3_doc_snippet target_cpp snippet_full_path)
+    string (REGEX REPLACE ".cpp$" "" target_name ${target_cpp})
+    string (REGEX REPLACE ".+/doc/" "" source_path ${snippet_full_path})
```
marehr:
```suggestion
    string (REGEX REPLACE ".+/(doc|test)/" "" test_path ${snippet_full_path})
    string (REGEX REPLACE ".+/(doc|test/snippet)/" "" source_path ${snippet_full_path})
```

smehringer: I tried something similar but the problem is that when adding the executable you use the relative path $(source_path) but this does not work for doc snippets instead I needed the $(snippet_full_path) to make it work... not sure how to adjust this in a generic way as I am inexperienced with cmake :/

PR #738, opened 2019-02-14
[MISC] Replace stream operator by std::from_chars for Integral types.

smehringer:
long overdue and also has better handling for "out of range" errors.

### DIFF
--- a/include/seqan3/argument_parser/detail/format_parse.hpp
+++ b/include/seqan3/argument_parser/detail/format_parse.hpp
@@ -18,6 +18,7 @@
 
 #include <seqan3/argument_parser/detail/format_base.hpp>
 #include <seqan3/std/concepts>
+#include <seqan3/std/charconv>
 
 namespace seqan3::detail
 {
@@ -323,10 +324,8 @@ private:
         value.push_back(tmp);
     }
 
-    /*!\brief Tries to cast an input string into a (integral) value and tests for overflow errors.
-     *
-     * \tparam option_t Must be stream-convertible.
-     *
+    /*!\brief Tries to cast an input string into an arithmetic value.
+     * \tparam option_t  The optiona value type; must model seqan3::Arithmetic.
      * \param[out] value Stores the casted value.
      * \param[in]  in    The input argument to be casted.
      *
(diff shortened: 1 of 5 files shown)


include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -344,25 +345,20 @@ class format_parse : public format_base
...
     void retrieve_value(option_t & value, std::string const & in)
     {
-        int64_t tmp;
-        std::istringstream stream{in};
-        stream >> tmp;
-
-        if (stream.fail() || !stream.eof()) // !stream.eof() catches 13a as wrong
-            throw type_conversion_failed("Argument " + in + " could not be casted to type " +
-                                         get_type_name_as_string(value) + ".");
+        auto res = std::from_chars(&in[0], &in[0] + in.size(), value);
```
eseiler: Would `&in[in.size()]` also work?

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -344,25 +345,20 @@ class format_parse : public format_base
...
 
-        if (tmp > static_cast<int64_t>(std::numeric_limits<option_t>::max()) ||
-            tmp < static_cast<int64_t>(std::numeric_limits<option_t>::min()))
+        if (res.ec == std::errc::result_out_of_range)
             throw overflow_error_on_conversion("Argument " + in + " is not in integer range [" +
                                                std::to_string(std::numeric_limits<option_t>::min()) + "," +
                                                std::to_string(std::numeric_limits<option_t>::max()) + "].");
-
-        value = static_cast<option_t>(tmp);
+        else if (res.ec == std::errc::invalid_argument || res.ptr != &in[0] + in.size())
```
eseiler: same here 🙈

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -344,25 +345,20 @@ class format_parse : public format_base
      */
     template <std::Integral option_t>
     //!\cond
-        requires IStream<std::istringstream, option_t>
+        requires IStream<std::istringstream, option_t> &&
+                 requires (option_t & v, const char * ptr) { std::from_chars(ptr, ptr, v); }
```
h-2: You are already constraining option_t to be integral, do you need this additional check?

smehringer: hmm if you define a type that models Integral it does not automatically work with from_chars does it? (But I actually noticed that this should reqire `arithmetic_concept` instead anyway.

h-2: You cannot define a type that models std::Integral. std::Integral just covers the built-in types. Same for arithmetic

smehringer: I removed the check.

PR #742, opened 2019-02-14
[Feature] View::interleave

joshuak94:
Implementation for view::interleave.


include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,352 @@
...
+    /*!\name Associated types
+     * \{
+     */
+    //!\brief The reference_type.
+    using reference         = reference_t<urng_t>;
+    //!\brief The const_reference type is equal to the reference type.
+    using const_reference   = detail::transformation_trait_or_t<seqan3::reference<urng_t const>, void>;
+    //!\brief The value_type (which equals the reference_type with any references removed).
+    using value_type        = value_type_t<urng_t>;
+    //!\brief If the underliying range is Sized, this resolves to range_type::size_type, otherwise void.
```
h-2: wrong description, underlying range is guaranteed to be sized as per the requirements

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,352 @@
...
+        urange{_urange}, step_size{_step_size}, inserted_range{_inserted_range}
+    {}
+
+    /*!\brief Construct from a ViewableRange urange and inserted_range by wrapping in a view::all.
+     * \param[in] _urange The underlying range.
+     * \param[in] _step_size The step size for the insertion.
+     * \param[in] _inserted_range The range to be inserted.
+     */
+    template <std::ranges::ViewableRange orng_t, std::ranges::ViewableRange oirng_t>
+    view_interleave(orng_t _urange, size_t const _step_size, oirng_t _inserted_range) :
```
h-2: don't take these by copy, but by `&&`

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,352 @@
...
+        view_interleave{view::all(_urange), _step_size, view::all(_inserted_range)}
```
h-2: then also wrap _range in `std::forward<orng_t>(_range=`
 (same for inserted_range)

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,352 @@
...
+     * \param[in] _urange The underlying range.
+     * \param[in] _step_size The step size for the insertion.
+     * \param[in] _inserted_range The range to be inserted.
+     */
+    view_interleave(urng_t _urange, size_t const _step_size, inserted_rng_t _inserted_range) :
+        urange{_urange}, step_size{_step_size}, inserted_range{_inserted_range}
+    {}
+
+    /*!\brief Construct from a ViewableRange urange and inserted_range by wrapping in a view::all.
+     * \param[in] _urange The underlying range.
```
h-2: add \tparam description

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,352 @@
...
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    iterator begin() noexcept
+    {
+        return {*this, 0};
+    }
+
+    //!\copydoc begin()
```
h-2: you can use `\overload` here, because the functions have the same name

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,352 @@
...
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    iterator end() noexcept
+    {
+        return {*this, size()};
+    }
+
+    //!\copydoc end()
```
h-2: same here

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,352 @@
...
+     *
+     * Constant.
+     *
+     */
+    size_type size()
+    {
+        return urange.size() + (std::floor(urange.size() / step_size) * inserted_range.size());
+    }
+
+    //!\copydoc size()
```
h-2: same here

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,352 @@
...
+    {
+        size_t combined_size = step_size + inserted_range.size();
+        assert(i < size());
+        if (i % (combined_size) < step_size)
+            return urange[i - (std::floor(i/(combined_size)) * inserted_range.size())];
+        else
+            return inserted_range[(i % (combined_size)) - step_size];
+    }
+
+    //!\copydoc operator[]()
```
h-2: and here

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,352 @@
...
+    const_reference operator[](size_type const i) const
+    {
+        size_t combined_size = step_size + inserted_range.size();
+        assert(i < size());
+        if (i % (combined_size) < step_size)
+            return urange[i - (std::floor(i/(combined_size)) * inserted_range.size())];
+        else
+            return inserted_range[(i % (combined_size)) - step_size];
+    }
+};
```
h-2: The class still needs a cast to container type. You can probably copy'n'paste this from one of the other views.

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,352 @@
...
+
+//!\brief Template argument type deduction guide that differentiates between View and ViewableRange inputs.
+//!\relates seqan3::detail::view_interleave
+template <std::ranges::RandomAccessRange urng_t, std::ranges::RandomAccessRange inserted_rng_t>
+    //!\cond
+    requires std::ranges::View<urng_t> && std::ranges::SizedRange<urng_t> &&
+             std::ranges::View<inserted_rng_t> && std::ranges::SizedRange<inserted_rng_t> &&
+             std::Same<reference_t<urng_t>, reference_t<inserted_rng_t>>
+    //!\endcond
+view_interleave(urng_t &&, size_t, inserted_rng_t &&) -> view_interleave<urng_t, inserted_rng_t>;
```
h-2: I think you can just remove the first deduction guide altogether :thinking:

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,352 @@
...
+
+private:
+    //!\brief Befriend the base class so it can call impl().
+    friend base_t;
+
+    /*!\brief       For ranges that are viewable, delegate to std::view::all.
+     * \returns     An instance of std::view::all.
+     */
+    template <std::ranges::RandomAccessRange urng_t, std::ranges::RandomAccessRange inserted_rng_t,
+              std::Integral size_type>
```
h-2: please make the order correspond to the  parameter order

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,352 @@
...
+        requires std::ranges::SizedRange<urng_t> &&
+                 std::ranges::SizedRange<inserted_rng_t>
+    static auto impl(urng_t && urange, size_type size, inserted_rng_t && i)
+    {
+        return view_interleave{urange, size, i};
```
h-2: std::forward around the ranges, see above

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    /*!\brief       For ranges that are not views and not lvalue-references, call view_interleave's constructor.
+     * \returns     An instance of seqan3::detail::view_interleave.
+     */
+    template <std::ranges::ForwardRange urng_t, std::ranges::ForwardRange inserted_rng_t, std::Integral size_type>
+    static auto impl(urng_t && urange, size_type size, inserted_rng_t && i)
+    {
+        return ranges::view::split(urange, size) | ranges::view::join(i);
```
h-2: std::forward

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,352 @@
...
+namespace seqan3::detail
+{
+
+// ============================================================================
+//  view_interleave
+// ============================================================================
+
+/*!\brief The type returned by seqan3::view::interleave.
+ * \tparam urng_t The type of the underlying range, must model std::ranges::RandomAccessRange.
+ * \implements std::ranges::View
```
h-2: document other tparams

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,352 @@
...
+ * \tparam inserted_rng_t The type of the range being inserted.
+ * \param[in] urange The range being processed.
+ * \param[in] inserted_range The range being inserted.
+ * \param[in] step_size A value of size_type which indicates the interval to insert the inserted_range.
+ * \returns A range with the second range inserted at regular intervals. See below for properties of said range.
+ * \ingroup view
+ *
+ * \details
+ *
+ * This view can be used to insert one range into another range at regular intervals.
```
h-2: add:

It behaves essentially like `| ranges::view::split(step_size) | ranges::view::join(inserted_range)` except that for input that models std::ranges::RandomAccessRange and std::ranges::SizedRange a more efficient data structure is returned (otherwise it returns exactly the above combination of views).

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,352 @@
...
+ * | std::ranges::ViewableRange      | *required*                            | *guaranteed*                                       |
+ * | std::ranges::View               |                                       | *guaranteed*                                       |
+ * | std::ranges::SizedRange         |                                       | *preserved*                                        |
+ * | std::ranges::CommonRange        |                                       | *preserved*                                        |
+ * | std::ranges::OutputRange        |                                       | *preserved*                                        |
+ * | seqan3::const_iterable_concept  |                                       | *preserved*                                        |
+ * |                                 |                                       |                                                    |
+ * | seqan3::reference_t             |                                       | seqan3::reference_t<urng_t>                                        |
```
h-2: I would make two tables:

 | range concepts and reference_t  | `urng_t` (underlying range type)      | `rrng_t` (returned range type)                     |
  |---------------------------------|:-------------------------------------:|:--------------------------------------------------:|
  | std::ranges::InputRange         | *required*                            | *preserved*                                        |
  | std::ranges::ForwardRange       | *required*                            | *preserved*                                        |
  | std::ranges::BidirectionalRange |   *required*                  | *preserved*                                        |
  | std::ranges::RandomAccessRange  |     *required*                   | *preserved*                                        |
  | std::ranges::ContiguousRange    |                                       | *lost*                                          |
  |                                 |                                       |                                                    |
  | std::ranges::ViewableRange      | *required*                            | *guaranteed*                                       |
  | std::ranges::View               |                                       | *guaranteed*                                       |
  | std::ranges::SizedRange         |    *required*                  | *preserved*                                        |
  | std::ranges::CommonRange        |                                       | *preserved*                                        |
  | std::ranges::OutputRange        |                                       | *preserved*                                        |
  | seqan3::const_iterable_concept  |                                       | *preserved*                                        |
  |                                 |                                       |                                                    |
  | seqan3::reference_t             |                                       | seqan3::reference_t<urng_t>                                        |


If above requirements are not met, this adaptor forwards to `| ranges::view::split(step_size) | ranges::view::join(inserted_range)` which returns a view with the following properties:

 | range concepts and reference_t  | `urng_t` (underlying range type)      | `rrng_t` (returned range type)                     |
  |---------------------------------|:-------------------------------------:|:--------------------------------------------------:|
  | std::ranges::InputRange         | *required*                            | *preserved*                                        |
  | std::ranges::ForwardRange       | *required*                            | *lost*                                        |
  | std::ranges::BidirectionalRange |                         | *lost*                                           |
  | std::ranges::RandomAccessRange  |                          | *lost*                                          |
  | std::ranges::ContiguousRange    |                                       | *lost*                                           |
  |                                 |                                       |                                                    |
  | std::ranges::ViewableRange      | *required*                            | *guaranteed*                                       |
  | std::ranges::View               |                                       | *guaranteed*                                       |
  | std::ranges::SizedRange         |                         | *lost*                                           |
  | std::ranges::CommonRange        |                                       | *lost*                                          |
  | std::ranges::OutputRange        |                                       | *lost*                                          |
  | seqan3::const_iterable_concept  |                                       | *lost*                                          |
  |                                 |                                       |                                                    |
  | seqan3::reference_t             |                                       | seqan3::value_type_t<urng_t>                                        |

h-2: Please make sure the spacing in the table is correct (likely not in my code snippet)

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,352 @@
...
+ * * `urng_t` is the type of the range modified by this view (input).
+ * * `rrng_type` is the type of the range returned by this view.
+ * * for more details, see \ref view.
+ *
+ * ### Example
+ *
+ * \snippet test/snippet/range/view/interleave.cpp general
+ * \hideinitializer
+ */
+inline constexpr auto interleave =detail::generic_pipable_view_adaptor<detail::view_interleave>{};
```
h-2: You aren't actually using your adaptor :laughing:

test/snippet/range/view/interleave.cpp
```diff
@@ -0,0 +1,26 @@
...
+std::string i{"in"};
+size_t s = 3;
+std::string cmp{"FOOinBARinBAXinBATin"};
+std::string cmp_rev{"niTAB"};
+
+// pipe notation
+// explicitly call view::all
+auto v0 = u | view::interleave(s, i);
+// or, via constructor.
+auto v1 = detail::view_interleave(u, s, i);
```
h-2: don't call detail:: here, instead `auto v1 = view::interleave(u, s, i);` should work, too!

test/unit/range/view/view_interleave_test.cpp
```diff
@@ -0,0 +1,58 @@
...
+    auto v1 = detail::view_interleave(view::all(u), s, view::all(i));
+
+    EXPECT_TRUE(std::ranges::InputRange<decltype(v1)>);
+    EXPECT_TRUE(std::ranges::ForwardRange<decltype(v1)>);
+    EXPECT_TRUE(std::ranges::BidirectionalRange<decltype(v1)>);
+    EXPECT_TRUE(std::ranges::RandomAccessRange<decltype(v1)>);
+    EXPECT_TRUE(std::ranges::View<decltype(v1)>);
+    EXPECT_TRUE(std::ranges::SizedRange<decltype(v1)>);
+    EXPECT_TRUE(std::ranges::CommonRange<decltype(v1)>);
+}
```
h-2: Please remove all the `view::all` in here, that should be taken care of automatically. Also pick a second example as original input that does not mode RandomAccess or Sized so that you can check that the other path (that forwards to view::split...) is actually taken. Also verify the properties on that view as I documented them in the table above!

joshuak94: Would you happen to have an example of something which doesn't model RandomAccess or Sized?

include/seqan3/range/view/interleave.hpp
joshuak94: Hmmmm, I'm not sure what cast to container type you're referring to?

test/unit/range/view/view_interleave_test.cpp
h-2: std::list doesn't model randomaccess, view::filter models neither randomaccess nie sized!

joshuak94: So, testing this, I now have a question. We implemented it using view::split. I think we should have used view::chunk instead? I think view::split splits matching a pattern, not at a specified interval.

joshuak94:
@h-2 I fixed everything, except for the casting to container type, because I wasn't sure what you meant by that. Could you point me to an example?

Also I think we were using view::split when we should have been using view::chunk. I switched it, let me know if that was wrong!

joshuak94:
@h-2 Resubmitted with requested changes.

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,396 @@
...
+
+    /*!\brief Construct from a ViewableRange urange and inserted_range by wrapping in a std::view::all.
+     * \param[in] _urange         The underlying range.
+     * \param[in] _step_size      The step size for the insertion.
+     * \param[in] _inserted_range The range to be inserted.
+     * \tparam    orng_t          The type of the underlying range, must be a std::ranges::ViewableRange.
+     * \tparam    oirng_t         The type of the inserted range, must be a std::ranges::ViewableRange.
+     */
+    template <std::ranges::ViewableRange orng_t, std::ranges::ViewableRange oirng_t>
+    view_interleave(orng_t && _urange, size_t const && _step_size, oirng_t && _inserted_range) :
```
h-2:
```suggestion
    view_interleave(orng_t && _urange, size_t const _step_size, oirng_t && _inserted_range) :
```

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,396 @@
...
+public:
+    //!\brief Inherit the base class's Constructors.
+    using base_t::base_t;
+
+private:
+    //!\brief Befriend the base class so it can call impl().
+    friend base_t;
+
+    /*!\brief       For ranges that are viewable, delegate to std::std::view::all.
+     * \returns     An instance of std::std::view::all.
```
h-2: ~~std::~~ std::

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,396 @@
...
+ * |                                 |                                       |                                                    |
+ * | std::ranges::ViewableRange      | *required*                            | *guaranteed*                                       |
+ * | std::ranges::View               |                                       | *guaranteed*                                       |
+ * | std::ranges::SizedRange         |                                       | *lost*                                             |
+ * | std::ranges::CommonRange        |                                       | *lost*                                             |
+ * | std::ranges::OutputRange        |                                       | *lost*                                             |
+ * | seqan3::const_iterable_concept  |                                       | *lost*                                             |
+ * |                                 |                                       |                                                    |
+ * | seqan3::reference_t             |                                       | seqan3::value_type_t<urng_t>                       |
```
h-2: empty line after end of table

test/snippet/range/view/interleave.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+// pipe notation
+// explicitly call view::all
+auto v0 = u | view::interleave(s, i);
+// or, via constructor.
+auto v1 = view::interleave(u, s, i);
+
+if (ranges::equal(v0, v1))
+    debug_stream << "Success!\n";
+//![general]
```
h-2: This looks more like a test than a snippets, maybe instead just a little shorter:
```cpp
std::string u{"FOOBARBAXBAT"};
std::string i{"in"};
size_t s = 3;

auto v = u | view::interleave(s, i);

debug_stream << v << '\n'; // prints FOOinBARinBAXinBATin
```
(main stuff around it of course, too)

test/unit/range/view/view_interleave_test.cpp
```diff
@@ -0,0 +1,98 @@
...
+    EXPECT_TRUE(ranges::equal(cmp, v6));
+}
+
+TEST(view_interleave, concepts)
+{
+    // Test with a string, modeling RandomAccess and Sized
+    std::string u{"FOOBARBAXBAT"};
+    std::string i{"in"};
+    size_t s = 3;
+    auto v1 = detail::view_interleave(u, s, i);
```
h-2: why detail?

test/unit/range/view/view_interleave_test.cpp
```diff
@@ -0,0 +1,98 @@
...
+    std::string i{"ff"};
+    size_t s = 1;
+    std::list<std::string> cmp{"aa", "ff", "bb", "ff", "cc", "ff", "dd", "ff", "ee"};
+    auto v1 = view::interleave(u, s, i);
+
+    EXPECT_TRUE(ranges::equal(cmp, v1));
+    EXPECT_TRUE(std::ranges::InputRange<decltype(v1)>);
+    EXPECT_TRUE(std::ranges::ViewableRange<decltype(v1)>);
+    EXPECT_TRUE(std::ranges::View<decltype(v1)>);
+}
```
h-2: please explicitly EXPECT_FALSE on the concepts that you expect to be false 😃

h-2:
Please have a look at the test failures.

Do you know what's up with the coverage statistics above?

h-2:
There is still a test failure, in general, please fix these before (re-)requesting a review (or state that you need help fixing them :wink: ).

joshuak94:
Blocked by #883

joshuak94:
@h-2 Any idea why aligned_sequence_test fails on travis, but it builds fine on my computer?

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,349 @@
...
+ * \author Joshua Kim <joshua.kim AT fu-berlin.de>
+ * \brief Provides seqan3::view::interleave.
+ */
+
+#pragma once
+
+#include <cmath>
+
+#include <range/v3/view/join.hpp>
+#include <range/v3/view/split.hpp>
```
h-2: These are not required anymore.

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,349 @@
...
+     */
+    size_type size()
+    {
+        return urange.size() + (std::floor(urange.size() / step_size) * inserted_range.size());
+    }
+
+    //!\copydoc size()
+    size_type size() const
+    {
+        return urange.size() + (std::floor(urange.size() / step_size) * inserted_range.size());
```
h-2: do not use `.size()` member functions, instead use `std::ranges::size()` free function.

h-2: also in other places

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,349 @@
...
+            "The range to be inserted by view::interleave must model std::ranges::ForwardRange.");
+        if constexpr (std::ranges::RandomAccessRange<urng_t> && std::ranges::SizedRange<urng_t> &&
+                      std::ranges::RandomAccessRange<inserted_rng_t> && std::ranges::SizedRange<inserted_rng_t>)
+        {
+            return detail::view_interleave{std::forward<urng_t>(urange), size, std::forward<inserted_rng_t>(i)};
+        }
+        else
+        {
+            return ranges::view::split(std::forward<urng_t>(urange), size) |
+                   ranges::view::join(std::forward<inserted_rng_t>(i));
```
h-2:
```cpp
            return std::view::split(std::forward<urng_t>(urange), size)
                 | std::view::join(std::forward<inserted_rng_t>(i));
```

test/snippet/range/view/interleave.cpp
h-2: this was never resolved

include/seqan3/range/view/interleave.hpp
h-2: this was not resolved

test/snippet/range/view/interleave.cpp
h-2: I think the policy is now to include the full snippet files in the API doc with the includes and main function, not a snippet from it.

test/unit/range/view/view_interleave_test.cpp
h-2: > We implemented it using view::split. I think we should have used view::chunk instead? I think view::split splits matching a pattern, not at a specified interval.

True, and since there is no test testing this, we also didnt realise now.

joshuak94:
Regarding the `detail::view_interleave...`, I was under the impression that it had to have the detail since the view_interleave is in the seqan3::detail namespace...

h-2:
Yes and no. If you want to construct the class directly you would need to do that, but you can invoke the adaptor in the same way! Note that I refer to `view::interleave` above, not `view_interleave`.
Also the regular user-visible API documentation should not contain references to `detail::` because those won't resolve anyway.

joshuak94:
Okay! So I'm a bit stuck on something.

The chunk&join vs the actual interleave function perform differently in this scenario:
DNA vector AAAAAA (length = 6)
Inserted vector GG (length = 2)

Should the output be: AAGGAAGGAAGG (what our interleave returns) OR AAGGAAGGAA (what chunk&join return). I guess chunk&join leaves out the last GG because chunk will split the original into AA, AA, AA, and then join just inserts the GG between each one.

smehringer:
The intuitive result would be `AAGGAAGGAA` with joins only in between

joshuak94:
Alright, I changed it so that it doesn't add the inserted range at the end.

h-2:
There are a bunch of review comments by me that are not yet addressed (I will *resolve* all those that I think are). Please have a look at those (or debate them with me :wink: ) before re-requesting review.

include/seqan3/range/view/interleave.hpp
joshuak94: So, when I take by reference, the test at this line fails to compile:
```cpp
    // don't call view::all
    auto v1 = u | view::interleave(s, i);
    EXPECT_TRUE(ranges::equal(cmp, v1));
```
The compiler says the constructor delegates to itself. I traced the issue and it seems that I can fix it by deleting the calls to `std::forward` in the `interleave_fn` struct... Is that alright?

joshuak94: Just kidding, deleting the `std::forward` builds the test, but then that line will seg fault when I execute the test.

joshuak94:
Alright, I think I've addressed mostly everything (haven't pushed yet). First was hoping you could help me with adding the `&&` causing compiler errors (commented in the review above).

include/seqan3/range/view/interleave.hpp
h-2: It's correct to forward!

Also we can simplify the constraints on the constructor (and make them more correct). The whole thing should look like this:
```cpp
     template <typename orng_t, typename oirng_t>
         //!\cond
         requires std::Constructible<urng_t, decltype(view::all(std::declval<orng_t>()))> &&
                  std::Constructible<inserted_rng_t, decltype(view::persist(std::declval<oirng_t>()))>
        //!\endcond
    view_interleave(orng_t && _urange, size_t const _step_size, oirng_t && _inserted_range) :
        view_interleave{view::all(std::forward<orng_t>(_urange)), _step_size, view::persist(std::forward<oirng_t>(_inserted_range))}
```

joshuak94:
Alright, now I hope I haven't missed anything!

include/seqan3/range/view/interleave.hpp
h-2: The types are not the underlying types strictly speaking... I would propose:
`\tparam orng_t A type that is the same as `urng_t` once wrapped in a view (e.g. reference to container).`

Also please use either no alignment of \tparam | parameter | description, or some consistent one :nail_care:

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,382 @@
...
+            "The underlying range parameter in view::interleave must model std::ranges::ForwardRange.");
+        static_assert(std::ranges::ViewableRange<urng_t>,
+            "The underlying range parameter in view::interleave must model std::ranges::ViewableRange.");
+        static_assert(std::ranges::ForwardRange<inserted_rng_t>,
+            "The range to be inserted by view::interleave must model std::ranges::ForwardRange.");
+        if constexpr (std::ranges::RandomAccessRange<urng_t> && std::ranges::SizedRange<urng_t> &&
+                      std::ranges::RandomAccessRange<inserted_rng_t> && std::ranges::SizedRange<inserted_rng_t>)
+        {
+            return detail::view_interleave{std::forward<urng_t>(urange), size, std::forward<inserted_rng_t>(i)};
+            // return detail::view_interleave{urange, size, i};
```
h-2: remove this line

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,382 @@
...
+ * * `urng_t` is the type of the range modified by this view (input).
+ * * `rrng_type` is the type of the range returned by this view.
+ * * for more details, see \ref view.
+ *
+ * ### Example
+ *
+ * \snippet test/snippet/range/view/interleave.cpp general
+ * \hideinitializer
+ */
+inline auto constexpr interleave = detail::interleave_fn{};
```
h-2: the convention here changed again so it`s `inline constexpr auto interleave = ...`

test/snippet/range/view/interleave.cpp
```diff
@@ -0,0 +1,22 @@
...
+
+int main()
+{
+std::string u{"FOOBARBAXBAT"};
+std::string i{"in"};
+size_t s = 3;
+
+// pipe notation
+auto v0 = u | seqan3::view::interleave(s, i);
+// or, via constructor.
```
h-2:
```suggestion
// or, via function notation
```
(it's not a constructor strictly speaking; you invoke the operator() on the functor and then that invokes a constructor)

test/snippet/range/view/interleave.cpp
h-2: Now that the full file is included, please indent the contents of main() like you would normally!

test/unit/range/view/view_interleave_test.cpp
h-2: Please also test explicitly that the other concepts are false :nail_care:

h-2:
Please note that there are (very few) unresolved issues from the previous reviews still!

include/seqan3/range/view/interleave.hpp
joshuak94: Just wondering, does the order actually make a difference? Or is this more for consistency's sake.

joshuak94:
@h-2 I hope I got all of the unresolved issues now!

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,383 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
```
h-2:
```suggestion
// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
```

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,383 @@
...
+ *
+ * Note that most members of this class are generated by ranges::view_interface which is not yet documented here.
+ */
+template <std::ranges::RandomAccessRange urng_t, std::ranges::RandomAccessRange inserted_rng_t>
+    //!\cond
+    requires std::ranges::View<urng_t> && std::ranges::SizedRange<urng_t> &&
+             std::ranges::View<inserted_rng_t> && std::ranges::SizedRange<inserted_rng_t> &&
+             std::Same<reference_t<urng_t>, reference_t<inserted_rng_t>>
+    //!\endcond
+class view_interleave : public ranges::view_interface<view_interleave<urng_t, inserted_rng_t>>
```
h-2:
```suggestion
class view_interleave : public std::ranges::view_interface<view_interleave<urng_t, inserted_rng_t>>
```

test/snippet/range/view/interleave.cpp
h-2: it seems you forgot to include the final `}`. Note that you can get rid of these `//![general]` ans just `\include` instead of `\snippet` above to get the whole.

test/unit/range/view/view_interleave_test.cpp
```diff
@@ -0,0 +1,100 @@
+#include <forward_list>
```
h-2: copyright missing

test/unit/range/view/view_interleave_test.cpp
h-2: :+1:

include/seqan3/range/view/interleave.hpp
h-2: No, it's just a matter of style.

include/seqan3/range/view/interleave.hpp
```diff
@@ -0,0 +1,383 @@
...
+ * | std::ranges::BidirectionalRange | *required*                            | *preserved*                     |
+ * | std::ranges::RandomAccessRange  | *required*                            | *preserved*                     |
+ * | std::ranges::ContiguousRange    |                                       | *lost*                          |
+ * |                                 |                                       |                                 |
+ * | std::ranges::ViewableRange      | *required*                            | *guaranteed*                    |
+ * | std::ranges::View               |                                       | *guaranteed*                    |
+ * | std::ranges::SizedRange         | *required*                            | *preserved*                     |
+ * | std::ranges::CommonRange        |                                       | *preserved*                     |
+ * | std::ranges::OutputRange        |                                       | *preserved*                     |
+ * | seqan3::const_iterable_concept  |                                       | *preserved*                     |
```
h-2:
```suggestion
 * | seqan3::ConstIterableRange      |                                       | *preserved*                     |
```

This was recently renamed by @MitraDarja.... Also beloew

PR #747, opened 2019-02-15
return values of view::get

### LINKED ISSUE #745: Nested zip views in combination with the get_view causes stack-use-after-scope errors
The following code compiles but leads to memory issues:

```cpp
    std::vector vec1{0, 1, 2, 3, 4, 5, 6, 7, 8, 9};

    auto get_view = ranges::view::zip(ranges::view::zip(vec1, vec1), vec1) | get<0>;

    for (auto && elem : get_view)
    {
        std::get<0>(std::forward<decltype(elem)>(elem)) = -1;
    }

    assert(vec1[0] == -1);
```

Compiling with address sanitizer causes stack-use-after-scope errors in release mode only.
Valgrind triggers invalid read in debug mode:

```console
==10149== Invalid read of size 8
==10149==    at 0x171766: int& std::__pair_get<0ul>::__move_get<int&, int&>(std::pair<int&, int&>&&) (utility:177)
==10149==    by 0x1705D1: std::tuple_element<0ul, std::pair<int&, int&> >::type&& std::get<0ul, int&, int&>(std::pair<int&, int&>&&) (utility:222)
==10149==    by 0x16B80D: view_get_defect_invalid_read_Test::TestBody() (view_get_test.cpp:177)
==10149==    by 0x1A3158: void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) (gtest.cc:2443)
==10149==    by 0x19D01E: void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) (gtest.cc:2479)
==10149==    by 0x17BA43: testing::Test::Run() (gtest.cc:2517)
==10149==    by 0x17C3A0: testing::TestInfo::Run() (gtest.cc:2693)
==10149==    by 0x17CA23: testing::TestCase::Run() (gtest.cc:2811)
==10149==    by 0x18780D: testing::internal::UnitTestImpl::RunAllTests() (gtest.cc:5177)
==10149==    by 0x1A423D: bool testing::internal::HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) (gtest.cc:2443)
==10149==    by 0x19DE10: bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) (gtest.cc:2479)
==10149==    by 0x186291: testing::UnitTest::Run() (gtest.cc:4786)
==10149==  Address 0x1fff000460 is on thread 1's stack
==10149==  384 bytes below stack pointer
==10149==
```

Note that following code runs without any issues:

```cpp
    std::vector vec1{0, 1, 2, 3, 4, 5, 6, 7, 8, 9};

    auto get_view = ranges::view::zip(ranges::view::zip(vec1, vec1), vec1);

    for (auto && elem : get_view)
    {
        auto && tmp = get<0>(std::forward<decltype(elem)>(elem));
        std::get<0>(std::forward<decltype(tmp)>(tmp)) = -1;
    }

    assert(vec1[0] == -1);
```

h-2:
fixes #745

### DIFF
--- a/include/seqan3/core/metafunction/basic.hpp
+++ b/include/seqan3/core/metafunction/basic.hpp
@@ -34,6 +34,28 @@ namespace seqan3
 template <typename t>
 using remove_cvref_t = std::remove_cv_t<std::remove_reference_t<t>>;
 
+// ----------------------------------------------------------------------------
+// remove_rvalue_reference
+// ----------------------------------------------------------------------------
+
+/*!\brief Return the input type with `&&` removed, but lvalue references preserved [Type metafunction].
+ * \tparam t The type to operate on.
+ * \see seqan3::remove_rvalue_reference_t
+ */
+template <typename t>
+struct remove_rvalue_reference
+{
+    //!\brief The return type is the input type with any `&&` stripped.
+    using type = std::conditional_t<std::is_rvalue_reference_v<t>, std::remove_reference_t<t>, t>;
+};
+
+/*!\brief Return the input type with `&&` removed, but lvalue references preserved [Type metafunction, shortcut].
+ * \tparam t The type to operate on.
+ * \see seqan3::remove_rvalue_reference
+ */
+template <typename t>
+using remove_rvalue_reference_t = typename remove_rvalue_reference<t>::type;
+
 //!\}
 
 } // namespace seqan3
--- a/include/seqan3/range/view/get.hpp
+++ b/include/seqan3/range/view/get.hpp
@@ -64,7 +64,12 @@ inline auto const get = view::transform([] (auto && in) -> decltype(auto)
     using seqan3::get;
     static_assert(tuple_like_concept<decltype(in)>,
                   "You may only pass ranges to view::get whose reference_t models the tuple_like_concept.");
-    return get<index>(std::forward<decltype(in)>(in));
+
+    // we need to explicitly remove && around temporaries to return values as values (and not as rvalue references)
+    // we cannot simply cast to std::tuple_element_t (or set that as return value), because some tuples, like
+    // our cartesian_composition alphabets do not return that type when get is called on them (they return a proxy)
+    using ret_type = remove_rvalue_reference_t<decltype(get<index>(std::forward<decltype(in)>(in)))>;
+    return static_cast<ret_type>(get<index>(std::forward<decltype(in)>(in)));
 });
 
 //!\}
--- a/test/unit/range/view/view_get_test.cpp
+++ b/test/unit/range/view/view_get_test.cpp
@@ -9,14 +9,15 @@
 
 #include <gtest/gtest.h>
 
-#include <seqan3/std/view/reverse.hpp>
+#include <range/v3/view/zip.hpp>
 
+#include <seqan3/alphabet/mask/masked.hpp>
+#include <seqan3/alphabet/quality/aliases.hpp>
 #include <seqan3/range/concept.hpp>
 #include <seqan3/range/view/get.hpp>
 #include <seqan3/range/view/to_char.hpp>
 #include <seqan3/range/view/complement.hpp>
-#include <seqan3/alphabet/quality/aliases.hpp>
-#include <seqan3/alphabet/mask/masked.hpp>
+#include <seqan3/std/view/reverse.hpp>
 
 using namespace seqan3;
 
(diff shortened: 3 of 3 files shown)


include/seqan3/range/view/get.hpp
```diff
@@ -64,7 +64,12 @@ inline auto const get = view::transform([] (auto && in) -> decltype(auto)
     using seqan3::get;
     static_assert(tuple_like_concept<decltype(in)>,
                   "You may only pass ranges to view::get whose reference_t models the tuple_like_concept.");
-    return get<index>(std::forward<decltype(in)>(in));
+
+    // we need to explicitly remove && around temporaries to return values as values (and not as rvalue references)
+    // we cannot simply cast to std::tuple_element_t (or set that as return value), because some tuples, like
+    // our cartesian_composition alphabets do not return that type when get is called on them (they return a proxy)
```
h-2: We could discuss at the strategy meeting if this is the expected behaviour or if the tuple_element_type should actually be the proxy

include/seqan3/range/view/get.hpp
```diff
@@ -64,7 +64,12 @@ inline auto const get = view::transform([] (auto && in) -> decltype(auto)
...
+    using ret_type = remove_rvalue_reference_t<decltype(get<index>(std::forward<decltype(in)>(in)))>;
```
rrahn:
```suggestion
    using r_type = std::conditional_t<std::is_rvalue_reference_v<decltype(get<index>(std::forward<decltype(in)>(in)))>,
                                      std::remove_reference_t<decltype(get<index>(std::forward<decltype(in)>(in)))>,
                                      decltype(get<index>(std::forward<decltype(in)>(in)))>;
```

rrahn: This was my solution but I couldn't grasp yesterday why perfect forwarding using glvalues did not work. Now, I think I got it. The inner tuple was moved into the outer one but it was returned as xvalue, which is a dangling reference after the lambda finished I suppose? 
In general, I do not see a general purpose use case for `remove_rvalue_referene_t`?

h-2: > In general, I do not see a general purpose use case for `remove_rvalue_referene_t`?

There were other situations were I defined two overloads where I could now use only one, e.g. the two overloads of assign_rank/assign_char free function: https://github.com/seqan/seqan3/blob/master/include/seqan3/alphabet/detail/member_exposure.hpp#L113

> This was my solution but I couldn't grasp yesterday why perfect forwarding using glvalues did not work. Now, I think I got it. The inner tuple was moved into the outer one but it was returned as xvalue, which is a dangling reference after the lambda finished I suppose?

Something like that.  The reference type of the nested zip:
```cpp
// originally:
tuple<tuple<int &, int &> &&, int &> &&
// now:
tuple<tuple<int &, int &>, int &>
```
The problem here is with the inner tuple: `&&` (and `const &` for that matter) extend the life-time of temporaries only until until their own lifetime ends. If you declare another reference pointing to the first, this doesn't extend the lifetime further. But this happens already when storing the reference (in our case inner tuple) in the (outer) tuple, that's why it's never safe to use `&&` (or owning `const &`)  as tuple element type.

I have added this small example to document this for someone else stumbling on the code comment or issue later:
```cpp
#include <tuple>

std::tuple<int const&> foobar()
{
        return {int{7}};
}

int main()
{
        auto t = foobar();
        return std::get<0>(t);
}
```
The return value is random on linux, always 255 on FreeBSD for some reason. Only address-sanatizer in release mode detects this as "stack-after-scope".

rrahn: As a matter of fact the reference type of the zip is: 
```cpp
// originally:
tuple<tuple<int &, int &>, int &>
```
But the lambda captures this as: `tuple<tuple<int &, int &>, int &> &&`  and the `get` will forward `tuple<int &, int &> &&` which causes the problem. And yes it is because the lifetime get's extended while it shouldn't. 👍 
But ok, if we need the remove_rvalue_reference in other places keep it, but perhaps instead of the overloads you could use the conditional statement to reduce boiler plate code a little?

h-2: done

include/seqan3/core/metafunction/basic.hpp
```diff
@@ -34,6 +34,26 @@ namespace seqan3
 template <typename t>
 using remove_cvref_t = std::remove_cv_t<std::remove_reference_t<t>>;
 
+// ----------------------------------------------------------------------------
+// remove_rvalue_reference
+// ----------------------------------------------------------------------------
+
+/*!\brief Return the input type with `&&` removed, but lvalue references preserved [Type metafunction].
+ * \tparam t The type to operate on.
+ */
```
rrahn: Please add a link to the shortcut, since they are not grouped together in the api documentation

test/unit/range/view/view_get_test.cpp
```diff
@@ -9,14 +9,15 @@
 
 #include <gtest/gtest.h>
 
-#include <seqan3/std/view/reverse.hpp>
+#include <range/v3/view/zip.hpp>
 
+#include <seqan3/alphabet/quality/aliases.hpp>
+#include <seqan3/alphabet/mask/masked.hpp>
```
rrahn: alphabetical order

PR #751, opened 2019-02-19
Argument parser nail polish

smehringer:
Not urgent to be merged but also not much to review :nail_care: 

Best to look at the commits individually.


include/seqan3/argument_parser/detail/format_base.hpp
```diff
@@ -54,25 +56,27 @@ class format_base
...
+                                       "signed 32 bit integer",
+                                       "unsigned 32 bit integer",
+                                       "signed 64 bit integer",
+                                       "unsigned 64 bit integer",
+                                       "double",
+                                       "float",
+                                       "bool",
+                                       "char",
+                                       "std::string",
+                                       "filesystem::path"};
```
rrahn: why is this not `std::filesystem::path`?

smehringer: because I in gcc7 `<filesystem>` is not available yet so I include `<seqan3/io/filesystem.hpp>` instead which is not in the std namespace..?

rrahn: It is only inconsistent with the std::string. And it seems like some user output which might lead to confusions about whether filesystem is a seqan3 thing or not. 
So filesystem is only our namespace alias for std::filesystem/std::experimental::filesystem. I am wondering if we should define the  experimental alias in std namespace if it is not available otherwise. But I am not sure if the standard permits this and then always refer to std::filesystem instead.

smehringer: I agree that we should alias std::filesystem. 
The following works with GCC7 and 8 for our filesystem.hpp:

```cpp
#pragma once

#if __has_include(<filesystem>)
#include <filesystem>
#else
#include <experimental/filesystem>
#endif // __has_include(experimental/filesystem)

#if !__has_include(<filesystem>)
namespace std
{
namespace filesystem = std::experimental::filesystem;
} // namespace std
#endif // __has_include(experimental/filesystem)

```
But we should probably put this into `/std/` then instead.
What do you think?

rrahn: Makes sense.

smehringer: I will change this if @h-2 agrees

PR #753, opened 2019-02-19
[MISC] Refactor argument_parser to exit

smehringer:
instead of throwing seqan3::parser_interruption on special requests.


smehringer:
well, I grep'ed for parser_interruption and changed all the places where it appeared. I think most things can stay like they are because we still need to "always wrap parsing into a try-catch block because the parser throws".

smehringer:
code coverage is confused :roll_eyes: man page printing was not tested before either (this is a todo still)

PR #759, opened 2019-02-21
[FIX] argument parsing of option with empty long identifier.

smehringer:
The test I added formerly would throw a `cast_conversion` error because it would recognize `--` as the long identifier for the `i` option and then parse `unknown` as the option value.

### DIFF
--- a/include/seqan3/argument_parser/detail/format_parse.hpp
+++ b/include/seqan3/argument_parser/detail/format_parse.hpp
@@ -173,6 +173,16 @@ public:
     void add_list_item(std::string const &, std::string const &) {}
     //!\endcond
 
+    //!\brief Checks whether \p id is empty.
+    template <typename id_type>
+    static bool is_empty_id(id_type const & id)
+    {
+        if constexpr (std::Same<remove_cvref_t<id_type>, std::string>)
+            return id.empty();
+        else // char
+            return is_char<'\0'>(id);
+    }
+
 private:
     /*!\brief Initializes the format_parse on construction.
      *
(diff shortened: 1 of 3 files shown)


test/unit/argument_parser/format_parse_test.cpp
```diff
@@ -768,6 +768,15 @@ TEST(parse_test, multiple_empty_options)
         EXPECT_EQ(1, option_value);
     }
 
+    {
+        const char * argv[]{"./empty_long", "-s=1", "--unkown"};
```
eseiler:
```suggestion
        const char * argv[]{"./empty_long", "-s=1", "--unknown"};
```

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -224,6 +224,10 @@ class format_parse : public format_base
     template <typename id_type>
     std::vector<std::string>::iterator find_option_id(std::vector<std::string>::iterator const begin_it, id_type const & id)
     {
+        if constexpr (std::Same<remove_cvref_t<id_type>, std::string>)
+            if (id.empty())
+                return end_of_options_it;
```
h-2: what if id_type is not a string? 🤔

smehringer: Then it is a char and never empty but maybe the null character which cannot be supplied via the command line so there is never a '-\0' option and the fine command is fine.

h-2: > Then it is a char and never empty but maybe the null character which cannot be supplied via the command line 

Sure you can supply it, e.g. by copy-n-paste or by invocation from a script... That's why we need fuzzy testing of the argument_parser!

smehringer:
@eseiler review ping :)

PR #691, opened 2019-01-26
[Feature] affine kernel with traceback

rrahn:
Supersedes #675 

Blocked  by #686 

Implements the traceback for generating the alignment.
Here are some benchmarks 
gcc9
```console
---------------------------------------------------------------------------
Benchmark                                    Time           CPU Iterations
---------------------------------------------------------------------------
seqan3_affine_dna4                      806558 ns     804354 ns        802
seqan2_affine_dna4                      722358 ns     715522 ns       1028
seqan3_affine_dna4_trace               3057949 ns    3054982 ns        223
seqan2_affine_dna4_trace               2710810 ns    2707381 ns        247
seqan3_affine_dna4_collection          3282436 ns    3278531 ns        209
seqan2_affine_dna4_collection          3055412 ns    3051580 ns        226
seqan3_affine_dna4_trace_collection   12665911 ns   12656358 ns         53
seqan2_affine_dna4_trace_collection   11679496 ns   11672161 ns         56
```

gcc8

```console
---------------------------------------------------------------------------
Benchmark                                    Time           CPU Iterations
---------------------------------------------------------------------------
seqan3_affine_dna4                      703024 ns     702296 ns        890
seqan2_affine_dna4                      673800 ns     673459 ns        947
seqan3_affine_dna4_trace               3170673 ns    3170121 ns        214
seqan2_affine_dna4_trace               2779336 ns    2777941 ns        253
seqan3_affine_dna4_collection          3243861 ns    3241996 ns        224
seqan2_affine_dna4_collection          3635063 ns    3632989 ns        190
seqan3_affine_dna4_trace_collection   13125700 ns   13119235 ns         51
seqan2_affine_dna4_trace_collection   11904751 ns   11895446 ns         56
```
The traceback in seqan3 is slower than in seqan2. 
This performance regression must be analysed to see where it comes from.

gcc7
```console
---------------------------------------------------------------------------
Benchmark                                    Time           CPU Iterations
---------------------------------------------------------------------------
seqan3_affine_dna4                      607930 ns     600999 ns       1007
seqan2_affine_dna4                      606012 ns     599308 ns       1021
seqan3_affine_dna4_trace               2873277 ns    2872426 ns        235
seqan2_affine_dna4_trace               2659906 ns    2659506 ns        245
seqan3_affine_dna4_collection          2452780 ns    2452078 ns        283
seqan2_affine_dna4_collection          2554415 ns    2553559 ns        263
[...]

PR #760, opened 2019-02-21
Feature affine traceback computation for alignments

### LINKED ISSUE #661: Implement alignment traceback
Implement the traceback kernel for the alignment algorithm.
Use the traceback computation to compute the end position/begin position option at the moment. 
This must be refined later on.

rrahn:
Implements the traceback computation for the banded and unbanded alignment case.

Superseeds #691 

Fixes #661 

This PR adds the traceback computation for the banded and unbanded case. There are still some open tasks that are all documented in the alignment module project. But this will handle ~99% of the normal alignment cases and is everything we need for the tutorials. 

Updated performance results for gcc-7

```console
---------------------------------------------------------------------------
Benchmark                                    Time           CPU Iterations
---------------------------------------------------------------------------
seqan3_affine_dna4                      578966 ns     578363 ns       1043
seqan2_affine_dna4                      581657 ns     581027 ns       1153
seqan3_affine_dna4_trace               2841675 ns    2837551 ns        243
seqan2_affine_dna4_trace               2560076 ns    2555956 ns        274
seqan3_affine_dna4_collection          2360674 ns    2357736 ns        295
seqan2_affine_dna4_collection          2512771 ns    2507039 ns        280
seqan3_affine_dna4_trace_collection   11702101 ns   11680847 ns         59
seqan2_affine_dna4_trace_collection   11049155 ns   11032645 ns         62
```

### DIFF
--- a/include/seqan3/alignment/matrix/alignment_coordinate.hpp
+++ b/include/seqan3/alignment/matrix/alignment_coordinate.hpp
@@ -15,6 +15,7 @@
 #include <type_traits>
 
 #include <seqan3/core/detail/strong_type.hpp>
+#include <seqan3/io/stream/debug_stream.hpp>
 #include <seqan3/std/concepts>
 #include <seqan3/std/iterator>
 
@@ -370,5 +371,25 @@ public:
     //!\brief The begin/end position of the alignment in the second sequence.
     SEQAN3_DOXYGEN_ONLY(size_t second_seq_pos;)
 };
+
+/*!\brief A seqan3::alignment_coordinate can be printed to the seqan3::debug_stream.
+ * \tparam    coordinate_type The alignment coordinate type.
+ * \param[in] s               The seqan3::debug_stream.
+ * \param[in] c               The alignment coordinate to print.
+ * \relates seqan3::debug_stream_type
+ *
+ * \details
+ *
+ * Prints the alignment coordinate as a tuple.
+ */
+template <typename coordinate_type>
+//!\cond
+    requires std::Same<remove_cvref_t<coordinate_type>, alignment_coordinate>
+//!\endcond
+inline debug_stream_type & operator<<(debug_stream_type & s, coordinate_type && c)
+{
+    s << std::tie(c.first_seq_pos, c.second_seq_pos);
+    return s;
+}
 
 } // namespace seqan3
--- a/include/seqan3/alignment/matrix/alignment_coordinate.hpp
+++ b/include/seqan3/alignment/matrix/alignment_coordinate.hpp
@@ -37,7 +37,7 @@ struct row_index_type : detail::strong_type<size_t, row_index_type>
     using detail::strong_type<size_t, row_index_type>::strong_type;
 };
 
-/*!\brief Represents a state to specify the implementation of the seqan3::detail::advanceable_alignment_coordinate
+/*!\brief Represents a state to specify the implementation of the seqan3::detail::advanceable_alignment_coordinate.
  * \ingroup alignment_matrix
  *
  * \details
@@ -94,21 +94,21 @@ public:
     ~advanceable_alignment_coordinate() noexcept = default;
 
     //!\brief Copy-constructs from another advanceable_alignment_coordinate with a different policy.
-    template <advanceable_alignment_coordinate_state _state>
+    template <advanceable_alignment_coordinate_state other_state>
     //!\cond
-        requires !std::Same<_state, state>
+        requires !std::Same<other_state, state>
     //!\endcond
-    constexpr advanceable_alignment_coordinate(advanceable_alignment_coordinate<_state> const & other) :
+    constexpr advanceable_alignment_coordinate(advanceable_alignment_coordinate<other_state> const & other) :
         first_seq_pos{other.first_seq_pos},
         second_seq_pos{other.second_seq_pos}
     {}
 
     //!\brief Move-constructs from another advanceable_alignment_coordinate with a different policy.
-    template <advanceable_alignment_coordinate_state _state>
+    template <advanceable_alignment_coordinate_state other_state>
     //!\cond
-        requires !std::Same<_state, state>
+        requires !std::Same<other_state, state>
     //!\endcond
-    constexpr advanceable_alignment_coordinate(advanceable_alignment_coordinate<_state> && other) :
+    constexpr advanceable_alignment_coordinate(advanceable_alignment_coordinate<other_state> && other) :
         first_seq_pos{std::move(other.first_seq_pos)},
         second_seq_pos{std::move(other.second_seq_pos)}
     {}
@@ -252,7 +252,7 @@ public:
               not available if the policy was set to seqan3::detail::advanceable_alignment_coordinate_state::none.
      */
     constexpr friend advanceable_alignment_coordinate operator+(difference_type const offset,
-                                         advanceable_alignment_coordinate const & me) noexcept
+                                                                advanceable_alignment_coordinate const & me) noexcept
     //!\cond
         requires state != advanceable_alignment_coordinate_state::none
     //!\endcond
@@ -275,11 +275,14 @@ namespace seqan3
 /*!\brief Represents the begin/end of the pairwise alignment in the respective sequences.
  * \ingroup alignment_matrix
  *
- * \details
  * \if DEV
+ * \details
  * This class only gives access to the respective positions of the sequences and is meant for
- * the user interface. For the implementation of the alignment algorithms the
- * seqan3::detail::advanceable_alignment_coordinate is used.
+ * the user interface. The additional complexity of an advanceable coorindate using the
+ * seqan3::detail::advanceable_alignment_coordinate is only necessary for the implementation of the pairwise
+ * alignment algorithm. Within in the algorithm the coordinate is used in combination with a seqan3::view::iota to
+ * keep track of the current position within the alignment matrix. For the user, however, this interface adds no
+ * benefit as they are only interested in the begin/end coordinates for the respective alignment.
  * \endif
  */
 class alignment_coordinate
@@ -305,20 +308,15 @@ public:
     ~alignment_coordinate() = default;
 
     //!\cond DEV
-    //!\brief Constructs from the seqan3::detail::advanceable_alignment_coordinate base class.
+    //!\brief Inherit the constructor from the base class.
+    using base_t::base_t;
+
+    // !\brief Constructs from the seqan3::detail::advanceable_alignment_coordinate base class.
     constexpr alignment_coordinate(base_t const & base) : base_t{base}
     {}
 
     //!\brief Constructs from the seqan3::detail::advanceable_alignment_coordinate base class.
     constexpr alignment_coordinate(base_t && base) : base_t{std::move(base)}
-    {}
-
-    /*!\brief Construction from the respective column and row indices.
-     * \param[in] c_idx The respective column index within the matrix. Of type seqan3::detail::column_index_type.
-     * \param[in] r_idx The respective row index within the matrix. Of type seqan3::detail::row_index_type.
-     */
-    constexpr alignment_coordinate(detail::column_index_type const c_idx,
-                                   detail::row_index_type const r_idx) noexcept : base_t{c_idx, r_idx}
     {}
     //!\endcond
     //!\}
--- a/include/seqan3/alignment/matrix/alignment_coordinate.hpp
+++ b/include/seqan3/alignment/matrix/alignment_coordinate.hpp
@@ -12,20 +12,363 @@
 
 #pragma once
 
-#include <seqan3/core/platform.hpp>
+#include <type_traits>
+
+#include <seqan3/core/detail/strong_type.hpp>
+#include <seqan3/std/concepts>
+#include <seqan3/std/iterator>
 
 namespace seqan3::detail
 {
-
-/*!\brief Represents the begin/end of the pairwise alignment in the respective sequences.
- * This class can for example be used to represent the coordinate where the best alignment score is located.
+/*!\brief A strong type for designated initialisation of the column index of the seqan3::detail::alignment_coordinate.
+ * \ingroup alignment_matrix
+ * \tparam index_t The type of the index; must model std::UnsignedIntegral.
  */
-struct alignment_coordinate
+template <std::UnsignedIntegral index_t>
+struct column_index_type : detail::strong_type<index_t, column_index_type<index_t>>
 {
-    //!\brief The position in the first sequence.
-    size_t seq1_pos;
-    //!\brief The position in the second sequence.
-    size_t seq2_pos;
+    //!!\brief Introduce base class constructor into this type's definition.
+    using detail::strong_type<index_t, column_index_type<index_t>>::strong_type;
 };
 
+/*!\name Type deduction guides
+ * \{
+ */
+//!\brief The default constructed element deduces to `size_t`.
+column_index_type() -> column_index_type<size_t>;
+//!\brief Deduces the position type from the constructor argument.
+template <std::UnsignedIntegral index_t>
+column_index_type(index_t const) -> column_index_type<size_t>;
+//!\}
+
+/*!\brief A strong type for designated initialisation of the row index of the seqan3::detail::alignment_coordinate.
+ * \ingroup alignment_matrix
+ * \tparam index_t The type of the index; must model std::UnsignedIntegral.
+ */
+template <std::UnsignedIntegral index_t>
+struct row_index_type : detail::strong_type<index_t, row_index_type<index_t>>
+{
+    //!!\brief Introduce base class constructor into this type's definition.
+    using detail::strong_type<index_t, row_index_type<index_t>>::strong_type;
+};
+
+/*!\name Type deduction guides
+ * \{
+ */
+//!\brief The default constructed element deduces to `size_t`.
+row_index_type() -> row_index_type<size_t>;
+//!\brief Deduces the position type from the constructor argument.
+template <std::UnsignedIntegral index_t>
+row_index_type(index_t const) -> row_index_type<size_t>;
+//!\}
+
+/*!\brief Represents a state to specify the implementation of the seqan3::detail::advanceable_alignment_coordinate
+ * \ingroup alignment_matrix
+ *
+ * \details
+ *
+ * The class seqan3::detail::alignment_coordinate can be extended with an incrementable and decrementable policy,
+ * such that it can be used as a value type inside of a iota_view. This state offers three policies: none, which
+ * leaves the functionality of seqan3::detail::alignment_coordinate untouched; column, which adds the respective
+ * functionality only for the column index and row, which adds the respective functionality only for the row index.
+ */
+enum struct advanceable_alignment_coordinate_state : uint8_t
+{
+    //!\brief The corresponding alignment coordinate will not be incrementable/decrementable.
+    none,
+    //!\brief The corresponding alignment coordinate will be incrementable/decrementable in the column index.
+    column,
+    //!\brief The corresponding alignment coordinate will be incrementable/decrementable in the row index.
+    row
+};
+
+/*!\brief Implements an internal alignment coordinate that can be used as an argument to the std::ranges::iota_view.
+ * \ingroup alignment_matrix
+ * \implements std::EqualityComparable
+ * \tparam index_t  The type of the sequence index; must model std::UnsignedIntegral.
+ * \tparam states   A non-type template flag to enable a specific advanceable policy.
+ *                  See seqan3::detail::advanceable_alignment_coordinate_state
+ *
+ * \details
+ *
+ * This class provides all members to make the `advanceable_alignment_coordinate` be usable in a std::ranges::iota_view.
+ * For the purpose of alignments modelling only incrementable and decrementable would fully suffice.
+ * Unfortunately, the current range implementation does not preserve std::ranges::BidirectionalRange properties,
+ * so we need to model the full advanceable concept in order to preserve the std::ranges::RandomAccessRange properties.
+ * This however can be relaxed if the range implementation fully complies with the current standard draft for Ranges,
+ * and increment and decrement would be enough.
+ */
+template <std::UnsignedIntegral index_t,
+          advanceable_alignment_coordinate_state state = advanceable_alignment_coordinate_state::none>
+class advanceable_alignment_coordinate
+{
+public:
+
+    //!\name Member types
+    //!\{
+    //!\brief Defines the difference type to model the std::WeaklyIncrementable concept.
+    using difference_type = std::make_signed_t<index_t>;
+    //!\}
+
+    /*!\name Constructor, destructor and assignment
+     * \{
+     */
+    constexpr advanceable_alignment_coordinate() noexcept = default;
+    constexpr advanceable_alignment_coordinate(advanceable_alignment_coordinate const &) noexcept = default;
+    constexpr advanceable_alignment_coordinate(advanceable_alignment_coordinate &&) noexcept = default;
+    constexpr advanceable_alignment_coordinate & operator=(advanceable_alignment_coordinate const &) noexcept = default;
+    constexpr advanceable_alignment_coordinate & operator=(advanceable_alignment_coordinate &&) noexcept = default;
+    ~advanceable_alignment_coordinate() noexcept = default;
+
+    //!\brief Copy-constructs from another advanceable_alignment_coordinate with a different policy.
+    template <typename _index_t, advanceable_alignment_coordinate_state _state>
+        requires std::Same<_index_t, index_t> && !std::Same<_state, state>
+    constexpr advanceable_alignment_coordinate(advanceable_alignment_coordinate<_index_t, _state> const & other) :
+        first_seq_pos{other.first_seq_pos},
+        second_seq_pos{other.second_seq_pos}
+    {}
+
+    //!\brief Move-constructs from another advanceable_alignment_coordinate with a different policy.
+    template <typename _index_t, advanceable_alignment_coordinate_state _state>
+        requires std::Same<_index_t, index_t> && !std::Same<_state, state>
+    constexpr advanceable_alignment_coordinate(advanceable_alignment_coordinate<_index_t, _state> && other) :
+        first_seq_pos{std::move(other.first_seq_pos)},
+        second_seq_pos{std::move(other.second_seq_pos)}
+    {}
+
+    /*!\brief Save construction from column and row indices.
+     * \param c_idx The respective column index within the matrix. Of type seqan3::detail::column_index_type.
+     * \param r_idx The respective row index within the matrix. Of type seqan3::detail::row_index_type.
+     */
+    constexpr advanceable_alignment_coordinate(column_index_type<index_t> const c_idx,
+                                               row_index_type<index_t> const r_idx) noexcept :
+        first_seq_pos{c_idx.get()},
+        second_seq_pos{r_idx.get()}
+    {}
+    //!\}
+
+     //!\cond
+     constexpr friend bool operator==(advanceable_alignment_coordinate const & lhs,
+                                      advanceable_alignment_coordinate const & rhs) noexcept
+     {
+         return (lhs.first_seq_pos == rhs.first_seq_pos) &&
+                (lhs.second_seq_pos == rhs.second_seq_pos);
+     }
+
+     constexpr friend bool operator!=(advanceable_alignment_coordinate const & lhs,
+                                      advanceable_alignment_coordinate const & rhs) noexcept
+     {
+         return !(lhs == rhs);
+     }
+     //!\endcond
+
+    /*!\name Member functions
+     * \brief Advances or decrements the respective column or row coordinate depending on the set policy.
+     *        These member functions are not available if the non-type template parameter `state` was set to
+     *        seqan3::detail::advanceable_alignment_coordinate_state::none.
+     * \{
+     */
+
+    constexpr advanceable_alignment_coordinate & operator++(/*pre-increment*/) noexcept
+    //!\cond
+        requires state != advanceable_alignment_coordinate_state::none
+    //!\endcond
+    {
+        if constexpr (state == advanceable_alignment_coordinate_state::column)
+            ++this->first_seq_pos;
+        else
+            ++this->second_seq_pos;
+        return *this;
+    }
+
+    constexpr advanceable_alignment_coordinate operator++(int /*post-increment*/) noexcept
+    //!\cond
+        requires state != advanceable_alignment_coordinate_state::none
+    //!\endcond
+    {
+        advanceable_alignment_coordinate tmp{*this};
+        ++(*this);
+        return tmp;
+    }
+
+    constexpr advanceable_alignment_coordinate & operator--(/*pre-decrement*/) noexcept
+    //!\cond
+        requires state != advanceable_alignment_coordinate_state::none
+    //!\endcond
+    {
+        if constexpr (state == advanceable_alignment_coordinate_state::column)
+            --this->first_seq_pos;
+        else
+            --this->second_seq_pos;
+        return *this;
+    }
+
+    constexpr advanceable_alignment_coordinate operator--(int /*post-decrement*/) noexcept
+    //!\cond
+        requires state != advanceable_alignment_coordinate_state::none
+    //!\endcond
+    {
+        advanceable_alignment_coordinate tmp{*this};
+        --(*this);
+        return tmp;
+    }
+
+    constexpr advanceable_alignment_coordinate & operator+=(difference_type const offset) noexcept
+    //!\cond
+        requires state != advanceable_alignment_coordinate_state::none
+    //!\endcond
+    {
+        if constexpr (state == advanceable_alignment_coordinate_state::column)
+            this->first_seq_pos += offset;
+        else
+            this->second_seq_pos += offset;
+        return *this;
+    }
+
+    constexpr advanceable_alignment_coordinate & operator-=(difference_type const offset) noexcept
+    //!\cond
+        requires state != advanceable_alignment_coordinate_state::none
+    //!\endcond
+    {
+        if constexpr (state == advanceable_alignment_coordinate_state::column)
+            this->first_seq_pos -= offset;
+        else
+            this->second_seq_pos -= offset;
+        return *this;
+    }
+
+    constexpr advanceable_alignment_coordinate operator+(difference_type const offset) const noexcept
+    //!\cond
+        requires state != advanceable_alignment_coordinate_state::none
+    //!\endcond
+    {
+        advanceable_alignment_coordinate tmp{*this};
+        tmp += offset;
+        return tmp;
+    }
+
+    constexpr advanceable_alignment_coordinate operator-(difference_type const offset) const noexcept
+    //!\cond
+        requires state != advanceable_alignment_coordinate_state::none
+    //!\endcond
+    {
+        advanceable_alignment_coordinate tmp{*this};
+        tmp -= offset;
+        return tmp;
+    }
+
+    constexpr difference_type operator-(advanceable_alignment_coordinate const & other) const noexcept
+    //!\cond
+        requires state != advanceable_alignment_coordinate_state::none
+    //!\endcond
+    {
+        if constexpr (state == advanceable_alignment_coordinate_state::column)
+            return this->first_seq_pos - other.first_seq_pos;
+        else
+            return this->second_seq_pos - other.second_seq_pos;
+    }
+    //!\}
+
+    //!\brief Non-member function
+    //!\{
+
+    /*!\brief Advances the respective coordinate depending on the set policy by the given offset. This function is
+              not available if the policy was set to seqan3::detail::advanceable_alignment_coordinate_state::none.
+     */
+    constexpr friend advanceable_alignment_coordinate operator+(difference_type const offset,
+                                         advanceable_alignment_coordinate const & me) noexcept
+    //!\cond
+        requires state != advanceable_alignment_coordinate_state::none
+    //!\endcond
+    {
+        return me + offset;
+    }
+    //!\}
+
+    //!\brief The begin/end position of the alignment in the first sequence.
+    index_t first_seq_pos{};
+    //!\brief The begin/end position of the alignment in the second sequence.
+    index_t second_seq_pos{};
+};
+
+/*!\name Type deduction guides
+ * \{
+ */
+advanceable_alignment_coordinate() -> advanceable_alignment_coordinate<size_t>;
+//!\brief Deduces the position type from the constructor.
+template <std::UnsignedIntegral index_t>
+advanceable_alignment_coordinate(column_index_type<index_t> const, row_index_type<index_t> const) ->
+    advanceable_alignment_coordinate<size_t>;
+//!\}
+
 } // namespace seqan3::detail
+
+namespace seqan3
+{
+
+/*!\brief Represents the begin/end of the pairwise alignment in the respective sequences.
+ * \ingroup alignment_matrix
+ *
+ * \details
+ * \if DEV
+ * This class only gives access to the respective positions of the sequences and is meant for
+ * the user interface. For the implementation of the alignment algorithms the
+ * seqan3::detail::advanceable_alignment_coordinate is used.
+ * \endif
+ */
+class alignment_coordinate
+//!\cond DEV
+    : public detail::advanceable_alignment_coordinate<size_t>
+//!\endcond
+{
+    //!\cond DEV
+    //!\brief The type of the base class.
+    using base_t = detail::advanceable_alignment_coordinate<size_t>;
+    //!\endcond
+
+public:
+
+    /*!\name Constructor, destructor and assignment
+     * \{
+     */
+    constexpr alignment_coordinate() = default;
+    constexpr alignment_coordinate(alignment_coordinate const &) = default;
+    constexpr alignment_coordinate(alignment_coordinate &&) = default;
+    constexpr alignment_coordinate & operator=(alignment_coordinate const &) = default;
+    constexpr alignment_coordinate & operator=(alignment_coordinate &&) = default;
+    ~alignment_coordinate() = default;
+
+    //!\cond DEV
+    //!\brief Constructs from the seqan3::detail::advanceable_alignment_coordinate base class.
+    constexpr alignment_coordinate(base_t const & base) : base_t{base}
+    {}
+
+    //!\brief Constructs from the seqan3::detail::advanceable_alignment_coordinate base class.
+    constexpr alignment_coordinate(base_t && base) : base_t{std::move(base)}
+    {}
+
+    /*!\brief Save construction from column and row indices.
+     * \tparam index_t The index type used to store the column and row position.
+     * \param[in] c_idx The respective column index within the matrix. Of type seqan3::detail::column_index_type.
+     * \param[in] r_idx The respective row index within the matrix. Of type seqan3::detail::row_index_type.
+     */
+    template <typename index_t>
+    constexpr alignment_coordinate(detail::column_index_type<index_t> const c_idx,
+                                   detail::row_index_type<index_t> const r_idx) noexcept :
+        base_t{c_idx, r_idx}
+    {}
+    //!\endcond
+    //!\}
+
+    //!\brief The begin/end position of the alignment in the first sequence.
+    using base_t::first_seq_pos;
+    //!\brief The begin/end position of the alignment in the second sequence.
+    using base_t::second_seq_pos;
+
+    //!\brief The begin/end position of the alignment in the first sequence.
+    SEQAN3_DOXYGEN_ONLY(size_t first_seq_pos;)
+    //!\brief The begin/end position of the alignment in the second sequence.
+    SEQAN3_DOXYGEN_ONLY(size_t second_seq_pos;)
+};
+
+} // namespace seqan3
--- a/include/seqan3/alignment/matrix/alignment_coordinate.hpp
+++ b/include/seqan3/alignment/matrix/alignment_coordinate.hpp
@@ -21,55 +21,29 @@
 
 namespace seqan3::detail
 {
-/*!\brief A strong type for designated initialisation of the column index of the seqan3::detail::alignment_coordinate.
- * \ingroup alignment_matrix
- * \tparam index_t The type of the index; must model std::UnsignedIntegral.
- */
-template <std::UnsignedIntegral index_t>
-struct column_index_type : detail::strong_type<index_t, column_index_type<index_t>>
+//!\brief A strong type for designated initialisation of the column index of the seqan3::detail::alignment_coordinate.
+//!\ingroup alignment_matrix
+struct column_index_type : detail::strong_type<size_t, column_index_type>
 {
     //!!\brief Introduce base class constructor into this type's definition.
-    using detail::strong_type<index_t, column_index_type<index_t>>::strong_type;
+    using detail::strong_type<size_t, column_index_type>::strong_type;
 };
 
-/*!\name Type deduction guides
- * \{
- */
-//!\brief The default constructed element deduces to `size_t`.
-column_index_type() -> column_index_type<size_t>;
-//!\brief Deduces the position type from the constructor argument.
-template <std::UnsignedIntegral index_t>
-column_index_type(index_t const) -> column_index_type<size_t>;
-//!\}
-
-/*!\brief A strong type for designated initialisation of the row index of the seqan3::detail::alignment_coordinate.
- * \ingroup alignment_matrix
- * \tparam index_t The type of the index; must model std::UnsignedIntegral.
- */
-template <std::UnsignedIntegral index_t>
-struct row_index_type : detail::strong_type<index_t, row_index_type<index_t>>
+//!\brief A strong type for designated initialisation of the row index of the seqan3::detail::alignment_coordinate.
+//!\ingroup alignment_matrix
+struct row_index_type : detail::strong_type<size_t, row_index_type>
 {
     //!!\brief Introduce base class constructor into this type's definition.
-    using detail::strong_type<index_t, row_index_type<index_t>>::strong_type;
+    using detail::strong_type<size_t, row_index_type>::strong_type;
 };
-
-/*!\name Type deduction guides
- * \{
- */
-//!\brief The default constructed element deduces to `size_t`.
-row_index_type() -> row_index_type<size_t>;
-//!\brief Deduces the position type from the constructor argument.
-template <std::UnsignedIntegral index_t>
-row_index_type(index_t const) -> row_index_type<size_t>;
-//!\}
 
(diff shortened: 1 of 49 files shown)


rrahn:
@joergi-w can you review asap, as it depends on my tutorial which I have to do next.

joergi-w:
@rrahn Sure, I can start immediately, but can we split and share the review somehow? For stepping through 3000 code lines I need a lot of time, and as the review is urgent for you it probably makes sense...

rrahn:
@joergi-w of course. Can you arrange with @eseiler and @smehringer?

include/seqan3/alignment/pairwise/policy/unbanded_score_trace_dp_matrix_policy.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+#include <range/v3/view/zip.hpp>
+
+#include <seqan3/alignment/matrix/trace_directions.hpp>
+#include <seqan3/alignment/pairwise/policy/unbanded_dp_matrix_policy.hpp>
+#include <seqan3/std/span.hpp>
+
+namespace seqan3::detail
+{
+/*!\brief Manages the allocation and provision of an unbanded dynamic programming matrix.
```
smehringer: what do you mean with `provision`? This is the exact same brief as `unbanded_dp_matrix_policy`... maybe add some difference like `that stores the trace and score information`..

include/seqan3/alignment/pairwise/policy/unbanded_score_trace_dp_matrix_policy.hpp
```diff
@@ -0,0 +1,114 @@
...
+    using cell_type = typename score_allocator_t::value_type;
+    //!\brief The underlying cell type of the trace matrix.
+    using trace_type = typename trace_allocator_t::value_type;
+    //!\brief The type of the score matrix.
+    using score_matrix_type = std::vector<cell_type, score_allocator_t>;
+    //!\brief The type of the trace matrix.
+    using trace_matrix_type = std::vector<trace_type, trace_allocator_t>;
+    //!\}
+
+    /*!\name Constructor, destructor and assignment
```
smehringer: Constructor`s` :nail_care:

include/seqan3/alignment/pairwise/policy/unbanded_score_trace_dp_matrix_policy.hpp
```diff
@@ -0,0 +1,114 @@
...
+     * \{
+     */
+    constexpr unbanded_score_trace_dp_matrix_policy() = default;
+    constexpr unbanded_score_trace_dp_matrix_policy(unbanded_score_trace_dp_matrix_policy const &) = default;
+    constexpr unbanded_score_trace_dp_matrix_policy(unbanded_score_trace_dp_matrix_policy &&) = default;
+    constexpr unbanded_score_trace_dp_matrix_policy & operator=(unbanded_score_trace_dp_matrix_policy const &) = default;
+    constexpr unbanded_score_trace_dp_matrix_policy & operator=(unbanded_score_trace_dp_matrix_policy &&) = default;
+    ~unbanded_score_trace_dp_matrix_policy() = default;
+    //!}
```
smehringer: `//!\}` !

include/seqan3/alignment/pairwise/policy/affine_gap_banded_policy.hpp
```diff
@@ -72,25 +72,53 @@ class affine_gap_banded_policy : public affine_gap_policy<affine_gap_banded_poli
...
-        auto const & prev_hz_score = get<1>(next_entry);
+        auto & [score_entry, coordinate, trace_value] = current_cell;
+        auto & [current_entry, next_entry] = score_entry;
+        auto & [main_score, hz_score, hz_trace] = current_entry;
+        auto const & [_ignore, prev_hz_score, prev_hz_trace]= next_entry;
         auto & [prev_cell, gap_open, gap_extend, opt] = cache;
-        auto & [tmp, vt_score] = prev_cell;
+        auto & [tmp, vt_score, vt_trace] = prev_cell;
+
+        std::ignore = _ignore;
```
joergi-w: Do you overwrite std::ignore here??

include/seqan3/alignment/pairwise/policy/unbanded_score_trace_dp_matrix_policy.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+    /*!\brief Allocates the memory for the dynamic programming matrix given the two sequences.
+     * \tparam first_range_t   The type of the first sequence (or packed sequences).
+     * \tparam second_range_t  The type of the second sequence (or packed sequences).
+     * \param[in] first_range  The first sequence (or packed sequences).
+     * \param[in] second_range The second sequence (or packed sequences).
+     */
+    template <typename first_range_t, typename second_range_t>
```
smehringer: no constrains? you call them ranges/sequences so at least a `std::Range`

include/seqan3/alignment/pairwise/policy/unbanded_score_trace_dp_matrix_policy.hpp
```diff
@@ -0,0 +1,114 @@
...
+    constexpr void allocate_matrix(first_range_t && first_range, second_range_t && second_range)
+    {
+        dimension_first_range = seqan3::size(std::forward<first_range_t>(first_range)) + 1;
```
smehringer: 1. `std::ranges::::size`?
2. You do not require that the ranges are sized, so maybe you should use `distance`

include/seqan3/alignment/pairwise/policy/affine_gap_banded_policy.hpp
```diff
@@ -72,25 +72,53 @@ class affine_gap_banded_policy : public affine_gap_policy<affine_gap_banded_poli
...
         // allows us to check if overflow/underflow would happen. Not sure, if this helps with the performance though.
         // (see https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html)
         main_score += score;
-        main_score = std::max(main_score, vt_score);
-        main_score = std::max(main_score, prev_hz_score);
+        // Compute where the max comes from.
+        if constexpr (decays_to_ignore_v<decltype(trace_value)>) // Don't compute a traceback
+        {
+            main_score = (main_score < vt_score) ? vt_score : main_score;
+            main_score = (main_score < prev_hz_score) ? prev_hz_score : main_score;
```
joergi-w: use `std::max(main_score, std::max(vt_score, prev_hz_score))`

include/seqan3/alignment/pairwise/policy/unbanded_score_trace_dp_matrix_policy.hpp
```diff
@@ -0,0 +1,114 @@
...
+    //!}
+
+    /*!\brief Allocates the memory for the dynamic programming matrix given the two sequences.
+     * \tparam first_range_t   The type of the first sequence (or packed sequences).
+     * \tparam second_range_t  The type of the second sequence (or packed sequences).
+     * \param[in] first_range  The first sequence (or packed sequences).
+     * \param[in] second_range The second sequence (or packed sequences).
+     */
+    template <typename first_range_t, typename second_range_t>
+    constexpr void allocate_matrix(first_range_t && first_range, second_range_t && second_range)
```
smehringer: question: will you ever receive rvalues here?

include/seqan3/alignment/pairwise/policy/unbanded_score_trace_dp_matrix_policy.hpp
```diff
@@ -0,0 +1,114 @@
...
+        score_matrix.resize(dimension_second_range);
+        // We use the full matrix to store the trace direction.
+        trace_matrix.resize(dimension_first_range * dimension_second_range);
+        trace_matrix_iter = trace_matrix.begin();
+    }
+
+    //!\brief Returns the current column of the alignment matrix.
+    constexpr auto current_column()
+    {
+        return ranges::view::zip(std::span{score_matrix}, std::span{&(*trace_matrix_iter), dimension_second_range});
```
smehringer: `&(*trace_matrix_iter)`   ?

include/seqan3/alignment/pairwise/policy/affine_gap_banded_policy.hpp
```diff
@@ -72,25 +72,53 @@ class affine_gap_banded_policy : public affine_gap_policy<affine_gap_banded_poli
...
         tmp = main_score + gap_open;
-        vt_score = std::max(vt_score + gap_extend, tmp);
-        hz_score = std::max(prev_hz_score + gap_extend, tmp);
+        vt_score += gap_extend;
+        hz_score = prev_hz_score + gap_extend;
+
+        if constexpr (decays_to_ignore_v<decltype(trace_value)>)
+        {
+            vt_score = (vt_score < tmp) ? tmp : vt_score;
+            hz_score = (hz_score < tmp) ? tmp : hz_score;
```
joergi-w: also here std::max

include/seqan3/alignment/pairwise/policy/affine_gap_banded_policy.hpp
```diff
@@ -72,25 +72,53 @@ class affine_gap_banded_policy : public affine_gap_policy<affine_gap_banded_poli
...
-        auto & [current_entry, next_entry] = get<0>(current_cell);
-        auto & [main_score, hz_score] = current_entry;
-        auto const & prev_hz_score = get<1>(next_entry);
+        auto & [score_entry, coordinate, trace_value] = current_cell;
+        auto & [current_entry, next_entry] = score_entry;
+        auto & [main_score, hz_score, hz_trace] = current_entry;
+        auto const & [_ignore, prev_hz_score, prev_hz_trace]= next_entry;
         auto & [prev_cell, gap_open, gap_extend, opt] = cache;
-        auto & [tmp, vt_score] = prev_cell;
+        auto & [tmp, vt_score, vt_trace] = prev_cell;
```
joergi-w: Find a better name for tmp.

include/seqan3/alignment/pairwise/policy/affine_gap_banded_policy.hpp
```diff
@@ -107,21 +135,40 @@ class affine_gap_banded_policy : public affine_gap_policy<affine_gap_banded_poli
...
         using std::get;
         // Unpack the cached variables.
-        auto & [current_entry, next_entry] = get<0>(current_cell);
+        auto & [score_entry, coordinate, trace_value] = current_cell;
+        auto & [current_entry, next_entry] = score_entry;
         auto & main_score = get<0>(current_entry);
         auto const & prev_hz_score = get<1>(next_entry);
-        auto & vt_score = get<1>(get<0>(cache));
+        auto const & prev_hz_trace = get<2>(next_entry);
+        auto & [tmp, vt_score, vt_trace] = get<0>(cache);
```
joergi-w: rename tmp

test/unit/alignment/pairwise/policy/unbanded_score_trace_dp_matrix_policy_test.cpp
```diff
@@ -0,0 +1,136 @@
...
+    using score_alloc_t = std::allocator<std::tuple_element_t<0, alloc_types>>;
+    using trace_alloc_t = std::allocator<std::tuple_element_t<1, alloc_types>>;
+
+    auto fixture()
+    {
+        return policy_mock<score_alloc_t, trace_alloc_t>{};
+    }
+};
+
+using test_type = std::tuple<std::tuple<int32_t, int32_t>, seqan3::detail::trace_directions>;
```
smehringer:
```diff
- using test_type = std::tuple<std::tuple<int32_t, int32_t>, seqan3::detail::trace_directions>;
+ using test_types = ::testing::Types<std::tuple<std::tuple<int32_t, int32_t>, seqan3::detail::trace_directions>>;
```

its done this way everywhere else and is confusing to read

include/seqan3/alignment/pairwise/policy/unbanded_score_trace_dp_matrix_policy.hpp
```diff
@@ -0,0 +1,114 @@
...
+    //!\brief The base type
+    using base_t = unbanded_dp_matrix_policy<unbanded_score_trace_dp_matrix_policy<derived_t,
+                                                                                   score_allocator_t,
+                                                                                   trace_allocator_t>,
+                                             score_allocator_t>;
+
+    //!\brief Befriends the derived class to grant it access to the private members.
+    friend derived_t;
+    //!\brief Make dimension_first_range visible in this class.
+    using base_t::dimension_first_range;
```
smehringer: why not `horizontal` and `vertical` range?

test/unit/alignment/pairwise/policy/unbanded_score_trace_dp_matrix_policy_test.cpp
```diff
@@ -0,0 +1,136 @@
...
+TYPED_TEST(unbanded_score_trace_test, next_column)
+{
+    std::string seq1{"garfieldthecat"};
+    std::string seq2{"garfieldthefatcat"};
+
+    auto mock = this->fixture();
+
+    mock.allocate_matrix(seq1, seq2);
+
+    // Get the active column
```
smehringer:
```diff
- // Get the active column
+ // Assign to the active column
```

test/unit/alignment/pairwise/policy/unbanded_score_trace_dp_matrix_policy_test.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+    // Get the same active column again
+    auto zip_view_2 = mock.current_column();
+    for (auto const entry : zip_view_2)
+    {
+        EXPECT_TRUE((std::get<0>(entry) == std::tuple{10, -10}));
+        EXPECT_TRUE((std::get<1>(entry) == seqan3::detail::trace_directions::diagonal));
+    }
+
+    // Go to next active column and check if active column has moved.
```
smehringer: how do you notice if it moved? it has the same score values as zip_view_2?

include/seqan3/alignment/pairwise/policy/banded_score_trace_dp_matrix_policy.hpp
```diff
@@ -0,0 +1,144 @@
...
+    // Import member functions from base class
+    using base_t::current_band_size;
+    using base_t::second_range_begin_offset;
+    using base_t::band_touches_last_row;
+    using base_t::trim_sequences;
+
+    /*!\name Member types
+     * \{
+     */
+    //!\brief The underlying cell type of the dynamic programming matrix.
```
joergi-w: doxygen annotation without a member type (forgot to delete this?)

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -134,7 +134,7 @@ class alignment_algorithm :
         assert(cfg_ptr != nullptr);
 
         // We need to allocate the score_matrix and maybe the trace_matrix.
-        this->allocate_score_matrix(first_range, second_range);
+        this->allocate_matrix(first_range, second_range);
```
smehringer: so both matrices (if present) are allocated here? Maybe we should call it `allocate_matrices`

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -276,6 +276,7 @@ class alignment_algorithm :
     void initialise_matrix(cache_t & cache)
     {
         auto col = this->current_column();
+        auto score_col = col | view_get_score_column;
```
smehringer: Doesn't the traceback matrix need to be initialized too? (what's the default btw?)

include/seqan3/alignment/pairwise/policy/banded_score_trace_dp_matrix_policy.hpp
```diff
@@ -0,0 +1,144 @@
...
+    /*!\name Constructor, destructor and assignment
+     * \{
+     */
+    constexpr banded_score_trace_dp_matrix_policy() = default;
+    constexpr banded_score_trace_dp_matrix_policy(banded_score_trace_dp_matrix_policy const &) = default;
+    constexpr banded_score_trace_dp_matrix_policy(banded_score_trace_dp_matrix_policy &&) = default;
+    constexpr banded_score_trace_dp_matrix_policy & operator=(banded_score_trace_dp_matrix_policy const &) = default;
+    constexpr banded_score_trace_dp_matrix_policy & operator=(banded_score_trace_dp_matrix_policy &&) = default;
+    ~banded_score_trace_dp_matrix_policy() = default;
+    //!}
```
joergi-w: escape the `}` for doxygen

include/seqan3/alignment/pairwise/alignment_configurator.hpp
```diff
@@ -295,6 +295,20 @@ struct alignment_configurator
...
     static constexpr function_wrapper_t configure_free_ends_optimum_search(config_t const & cfg);
 
+    /*!\brief Determines the trace type.
+     * \tparam config_t The configuration type.
+     */
+    template <typename config_t>
+    struct configure_trace_type
+    {
+        //!\brief If traceback is enabled resolves to seqan3::detail::trace_directions,
+        //!\      otherwise seqan3::detail::ignore_t.
```
joergi-w: Why not std::ignore?

eseiler:
My review is for the 3rd commit

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -288,6 +288,30 @@ class alignment_algorithm :
         {
             res.score = get<3>(cache).score;
         }
+        if constexpr (config_t::template exists<align_cfg::result<with_end_position_type>>())
+        {
+            res.score = get<3>(cache).score;
```
joergi-w: At some point, we should have enum values to identify the correct tuple positions.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -436,6 +463,61 @@ class alignment_algorithm :
...
+    template <typename first_range_t, typename second_range_t>
+    auto compute_traceback(first_range_t const & first_range,
+                           second_range_t const & second_range,
+                           alignment_coordinate const & end_coordinate)
+    {
+        using first_seq_value_type = value_type_t<first_range_t>;
+        using second_seq_value_type = value_type_t<second_range_t>;
+
+        // Parse the traceback
+        auto [begin_coordinate, first_gap_segments, second_gap_segments] = this->parse_traceback(end_coordinate);
```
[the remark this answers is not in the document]
joergi-w: Ah, found it. It refers to the first (horizontal) and second (vertical) sequence.
Think again about these names, because you often use directional names, as up/left/horizontal etc. and it should be consistent all over the alignment module.

include/seqan3/alignment/matrix/alignment_coordinate.hpp
```diff
@@ -12,20 +12,363 @@
...
+
+    //!\brief Move-constructs from another advanceable_alignment_coordinate with a different policy.
+    template <typename _index_t, advanceable_alignment_coordinate_state _state>
+        requires std::Same<_index_t, index_t> && !std::Same<_state, state>
+    constexpr advanceable_alignment_coordinate(advanceable_alignment_coordinate<_index_t, _state> && other) :
+        first_seq_pos{std::move(other.first_seq_pos)},
+        second_seq_pos{std::move(other.second_seq_pos)}
+    {}
+
+    /*!\brief Save construction from column and row indices.
```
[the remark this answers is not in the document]
rrahn: No, safe in the sense that it uses a strong interface.

include/seqan3/alignment/matrix/alignment_coordinate.hpp
```diff
@@ -12,20 +12,363 @@
...
+    //!\cond
+        requires state != advanceable_alignment_coordinate_state::none
+    //!\endcond
+    {
+        advanceable_alignment_coordinate tmp{*this};
+        tmp -= offset;
+        return tmp;
+    }
+
+    constexpr difference_type operator-(advanceable_alignment_coordinate const & other) const noexcept
```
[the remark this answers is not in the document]
rrahn: No, in this case I am using the same template as *this. Comparing different types (different states) would cause a compiler error.

include/seqan3/alignment/matrix/alignment_coordinate.hpp
```diff
@@ -12,20 +12,363 @@
...
+    //!\cond DEV
+    //!\brief Constructs from the seqan3::detail::advanceable_alignment_coordinate base class.
+    constexpr alignment_coordinate(base_t const & base) : base_t{base}
+    {}
+
+    //!\brief Constructs from the seqan3::detail::advanceable_alignment_coordinate base class.
+    constexpr alignment_coordinate(base_t && base) : base_t{std::move(base)}
+    {}
+
+    /*!\brief Save construction from column and row indices.
```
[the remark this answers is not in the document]
rrahn: I removed the `safe` as it seems to add unnecessary confusion.

include/seqan3/alignment/matrix/alignment_coordinate.hpp
```diff
@@ -371,4 +372,24 @@ class alignment_coordinate
...
+ * \param[in] c               The alignment coordinate to print.
+ * \relates seqan3::debug_stream_type
+ *
+ * \details
+ *
+ * Prints the alignment coordinate as a tuple.
+ */
+template <typename coordinate_type>
+//!\cond
+    requires std::Same<remove_cvref_t<coordinate_type>, alignment_coordinate>
```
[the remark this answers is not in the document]
rrahn: Well, it is missing something. Good spot.

include/seqan3/alignment/matrix/alignment_optimum.hpp
```diff
@@ -0,0 +1,86 @@
...
+alignment_optimum(score_t const, alignment_coordinate const) ->
+    alignment_optimum<std::remove_reference_t<score_t>>;
+//!\}
+
+/*!\brief A less than comparator for two seqan3::detail::alignment_optimum objects.
+ * \ingroup alignment_matrix
+ *
+ * \details
+ *
+ * This function object is used in std::max functions to compare two seqan3::detail::alignment_optimum objects.
```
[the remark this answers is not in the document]
rrahn: Nop, the design pattern is officially called a function object.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
rrahn: Well, just because the internal layout uses two vectors it does not mean that the interface just allocates the dynamic programming matrix, what ever that is in detail. And in fact for the score only computation only one matrix is allocated which is even just one column. So I would like to keep the name as is.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
rrahn: There is already a ticket for refactoring the tuple types into aggregates with named member access, as it is really hard to follow the code with just the tuples.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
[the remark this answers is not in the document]
rrahn: So the entire public API will refer to first and second range as for the pairwise alignment the user is not interested in the internal layout. I kept this naming scheme for most of the internal implementation whenever I refer to the ranges. When I refer to horizontal or vertical then only at the places where the layout of the matrix is relevant or can be easier understood. (AFAICT this is only in comments)
For the trace directions: up, left and diagonal. They were introduced for the edit distance alignment and I think it is ok to give them the name in the way they walk. Like go left, up or diagonal. One could rename them to horizontal and vertical but I don't see any benefit in this. If you think otherwise, please open an issue, so the group can discuss it. It would be another PR anyway.

include/seqan3/alignment/pairwise/alignment_configurator.hpp
rrahn: std::ignore is not a type but a variable. ignore_t is `remove_cv_t<decltype(std::ignore)>`

include/seqan3/alignment/pairwise/alignment_configurator.hpp
```diff
@@ -307,8 +307,8 @@ constexpr function_wrapper_t alignment_configurator::configure_free_ends_initial
     // score and cell type
     // ----------------------------------------------------------------------------
 
-    using score_type = int;
-    using cell_type = std::tuple<score_type, score_type>;
+    using score_type = int32_t;
```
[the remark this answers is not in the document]
rrahn: We are not there yet 😉

include/seqan3/alignment/pairwise/policy/affine_gap_banded_policy.hpp
rrahn: No, assignment to std::ignore will hopefully be a noop and removed from the compiler. You can assign anything to std::ignore. I hope it is the same as `(void) _ignore`. But I need to tests this more in detail anyway. Remember that the is a ticket with `improve performance` 🤣

include/seqan3/alignment/pairwise/policy/affine_gap_banded_policy.hpp
rrahn: well in SeqAn2 it made a performance difference, also I am not sure why.
Second it is likely that I remove this when dealing with the performance thing, as I hope that in case trace_value is std::ignore, assigning to it will be optimised away

include/seqan3/alignment/pairwise/policy/affine_gap_banded_policy.hpp
rrahn: see above

include/seqan3/alignment/pairwise/policy/affine_gap_banded_policy.hpp
rrahn: But it is a value that is not propagated out of the function and is used to temporarily store an intermediate result. So that is what `tmp` suggests

include/seqan3/alignment/pairwise/policy/affine_gap_banded_policy.hpp
rrahn: see above

include/seqan3/alignment/pairwise/policy/all.hpp
```diff
@@ -82,6 +83,7 @@
  * ### Existing matrix policies:
  *
  *  - seqan3::detail::unbanded_dp_matrix_policy
+ *  - seqan3::detail::unbanded_score_trace_dp_matrix_policy
```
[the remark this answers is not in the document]
rrahn: Well, yes, but I like to do these in another PR. For the tutorial documentation I'd like to make these here interface documentations anyway.

include/seqan3/alignment/pairwise/policy/unbanded_score_trace_dp_matrix_policy.hpp
rrahn: Again for me the dynamic programming matrix is a global property that might be represented internally in different vectors. I added a detail description.

include/seqan3/alignment/pairwise/policy/unbanded_score_trace_dp_matrix_policy.hpp
rrahn: I tried to keep the naming with respect to the user interface, which uses first_range and second_range. Only in some specific places where I refer to the matrix layout I have a comment talking about horizontal and vertical direction. I see I could do it for the dimensions as well, but mostly when using the dimensions the ranges are affected by this.

include/seqan3/alignment/pairwise/policy/unbanded_score_trace_dp_matrix_policy.hpp
rrahn: Not in the internal interface. This will be already checked when invoking the class.

include/seqan3/alignment/pairwise/policy/unbanded_score_trace_dp_matrix_policy.hpp
rrahn: This will be refactored anyway. as I should not use forwarding references here, since we cannot steal the memory anyway. But I make this in a clean PR.

test/unit/alignment/pairwise/policy/unbanded_score_trace_dp_matrix_policy_test.cpp
rrahn: yes it is the expected behavior to have the same score, since it uses only on column and recycles this.

rrahn: The other values have changed though.

include/seqan3/alignment/pairwise/policy/unbanded_score_trace_dp_matrix_policy.hpp
smehringer: maybe you should make a list of follow up issues then...

include/seqan3/alignment/pairwise/policy/unbanded_score_trace_dp_matrix_policy.hpp
smehringer: ?

include/seqan3/alignment/pairwise/policy/all.hpp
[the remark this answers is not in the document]
rrahn: Ok i changed it right away

include/seqan3/alignment/pairwise/policy/unbanded_score_trace_dp_matrix_policy.hpp
rrahn: @smehringer  can this be resolved?

include/seqan3/alignment/pairwise/policy/unbanded_score_trace_dp_matrix_policy.hpp
rrahn: I included this now in this PR as well.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
rrahn: The traceback matrix will be written to within the computation.

rrahn:
@eseiler @smehringer I removed the template argument from the alignment coordinate. I had it originally planned to be generic, but in the current use case it is really not used. So, this can be refactored later if we realize it is used in other contexts as well.

include/seqan3/alignment/pairwise/policy/affine_gap_banded_policy.hpp
```diff
@@ -72,24 +72,53 @@ class affine_gap_banded_policy : public affine_gap_policy<affine_gap_banded_poli
...
-        auto const & prev_hz_score = get<1>(next_entry);
+        auto & [score_entry, coordinate, trace_value] = current_cell;
+        auto & [current_entry, next_entry] = score_entry;
+        auto & [main_score, hz_score, hz_trace] = current_entry;
+        auto const & [prev_main_score, prev_hz_score, prev_hz_trace]= next_entry;
         auto & [prev_cell, gap_open, gap_extend, opt] = cache;
-        auto & [tmp, vt_score] = prev_cell;
+        auto & [tmp, vt_score, vt_trace] = prev_cell;
+
+        std::ignore = prev_main_score;
```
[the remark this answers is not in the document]
rrahn: The variable is not used and would otherwise create an unused warning. This is now a just more expressive way to say `(void) prev_main_score`, but with the same effect and optimised away by the compiler anyway.

include/seqan3/alignment/pairwise/policy/banded_score_dp_matrix_policy.hpp
```diff
@@ -89,9 +94,16 @@ class banded_score_dp_matrix_policy :
...
 
+        band_size = band_column_index + band_row_index + 1;
+
         // Reserve one more cell to deal with last cell in the banded column which needs only the diagonal and up cell.
         // TODO: introduce specific named cell types with initialisation values.
-        score_matrix.resize(band_column_index + band_row_index + 2, cell_type{INF, INF});
+        score_matrix.resize(band_size + 1);
+
+        using std::get;
+        get<0>(score_matrix.back()) = INF;
```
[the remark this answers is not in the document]
rrahn: No it is actually the lowest() / 2. It is some detail used inside the algorithm to compute the last value of a column inside a band. It is necessary to avoid underflows in the kernel and to avoid unnecessary instructions.

include/seqan3/alignment/pairwise/execution/execution_handler_sequential.hpp
```diff
@@ -29,14 +30,30 @@ struct execution_handler_sequential
...
-    template <typename func_result_t, typename first_range_t, typename second_range_t>
-    void execute(std::function<func_result_t(first_range_t const &, second_range_t const &)> func,
-                 first_range_t const & first_range,
-                 second_range_t const & second_range,
-                 std::function<void(decltype(func(first_range, second_range)))> delegate)
+    /*!\brief Invokes the passed alignment instance in a blocking manner.
+     * \tparam fn_type           The callable that needs to be invoked; must model std::Invocable with first_range_type
+     *                           and second_range_type.
+     * \tparam first_range_type  The type of the first range.
+     * \tparam second_range_type The type of the first range.
```
joergi-w: ...the *second* range.

test/unit/alignment/pairwise/alignment_configurator_test.cpp
```diff
@@ -18,63 +18,49 @@ using namespace seqan3;
...
-TEST(alignment_configurator, configure_edit)
+template <typename config_t>
+bool run_test(config_t const & cfg)
 {
-    auto cfg = align_cfg::edit;
-
     auto r = setup();
     auto fn = detail::alignment_configurator::configure(r, cfg);
-    auto [seq1, seq2] = *seqan3::begin(r);
+    auto && [seq1, seq2] = *seqan3::begin(r);
```
smehringer:
```suggestion
    auto && [seq1, seq2] = *std::ranges::begin(r);
```

test/unit/alignment/pairwise/policy/banded_score_dp_matrix_policy_test.cpp
```diff
@@ -81,8 +81,9 @@ TEST(banded_score_dp_matrix_policy, allocate_matrix)
     EXPECT_EQ(mock.band_row_index, 3u);
     EXPECT_NE(mock.current_matrix_iter, seqan3::begin(mock.score_matrix));
 
-    auto last_cell = *(--seqan3::end(mock.score_matrix));
-    EXPECT_EQ(last_cell, (std::tuple{decltype(mock)::INF, decltype(mock)::INF}));
+    auto last_cell = *std::ranges::prev(seqan3::end(mock.score_matrix));
```
smehringer:
```suggestion
    auto last_cell = *std::ranges::prev(std::ranges::end(mock.score_matrix));
```

test/unit/alignment/pairwise/policy/banded_score_dp_matrix_policy_test.cpp
```diff
@@ -149,19 +150,27 @@ TEST(banded_score_dp_matrix_policy, current_column)
...
+    for (auto && tpl : col | view::take_exactly(3u))
+    {
+        int first;
+        int second;
+        std::tie(first, second, std::ignore) = std::get<1>(tpl);
+        EXPECT_EQ((std::tie(first, second)), (std::tuple{-1, -1}));
+    }
 
-    EXPECT_EQ(std::get<1>(*(--seqan3::end(col))), (std::tuple{decltype(mock)::INF, decltype(mock)::INF}));
+    EXPECT_EQ(std::get<0>(std::get<1>(*std::ranges::prev(seqan3::end(col)))), decltype(mock)::INF);
```
smehringer: std::ranges::end

test/unit/alignment/pairwise/policy/unbanded_score_dp_matrix_policy_test.cpp
```diff
@@ -0,0 +1,144 @@
...
+    std::string seq1{"garfieldthecat"};
+    std::string seq2{"garfieldthefatcat"};
+
+    auto mock = this->fixture();
+
+    mock.allocate_matrix(seq1, seq2);
+
+    auto zip_view = mock.current_column();
+
+    EXPECT_EQ(seqan3::size(zip_view), seq2.size() + 1);
```
smehringer: std::ranges::size

test/unit/alignment/pairwise/policy/unbanded_score_dp_matrix_policy_test.cpp
```diff
@@ -0,0 +1,144 @@
...
+{
+    auto mock = this->fixture();
+
+    // Get the active column
+    auto zip_view = mock.current_column();
+
+    EXPECT_TRUE(std::ranges::BidirectionalRange<decltype(zip_view)>);
+    EXPECT_TRUE(std::ranges::SizedRange<decltype(zip_view)>);
+
+    auto it = seqan3::begin(zip_view);
```
smehringer: std::ranges::begin

test/unit/alignment/pairwise/policy/unbanded_score_dp_matrix_policy_test.cpp
```diff
@@ -0,0 +1,144 @@
...
+    EXPECT_TRUE(std::BidirectionalIterator<decltype(it)>);
+
+    auto it_e = seqan3::end(zip_view);
```
smehringer: std::ranges::end

test/unit/alignment/pairwise/policy/unbanded_score_trace_dp_matrix_policy_test.cpp
```diff
@@ -0,0 +1,149 @@
...
+    std::string seq1{"garfieldthecat"};
+    std::string seq2{"garfieldthefatcat"};
+
+    auto mock = this->fixture();
+
+    mock.allocate_matrix(seq1, seq2);
+
+    auto zip_view = mock.current_column();
+
+    EXPECT_EQ(seqan3::size(zip_view), seq2.size() + 1);
```
smehringer: std::ranges::end

test/unit/alignment/pairwise/policy/unbanded_score_trace_dp_matrix_policy_test.cpp
```diff
@@ -0,0 +1,149 @@
...
+    row_index = 0;
+    for (auto const entry : zip_view_3)
+    {
+        EXPECT_TRUE((std::get<0>(entry) == std::tuple{10, -10}));
+        EXPECT_EQ(std::get<1>(entry).first_seq_pos,  1u);
+        EXPECT_EQ(std::get<1>(entry).second_seq_pos,  row_index++);
+        EXPECT_TRUE((std::get<2>(entry) == seqan3::detail::trace_directions::none));
+    }
+
+    EXPECT_EQ(seqan3::size(zip_view_3), seq2.size() + 1);
```
smehringer: std::ranges::size

test/unit/alignment/pairwise/alignment_configurator_test.cpp
rrahn: Woah, and I thought I did a regex replace on the entire subdirectory 😲

smehringer: I guessed so but I checked nonetheless :nerd_face:

rrahn: Oh it is in the tests, right 🤦‍♂️

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -301,26 +370,41 @@ class alignment_algorithm :
...
+            // Move internal matrix to next column.
+            this->next_column();
+
             auto col = this->current_column();
-            this->init_row_cell(*seqan3::begin(col), cache);
+            this->init_row_cell(*std::ranges::begin(col), cache);
 
             auto second_range_it = std::ranges::begin(second_range);
-            ranges::for_each(col | ranges::view::drop_exactly(1), [&, this] (auto & cell)
+            ranges::for_each(col | ranges::view::drop_exactly(1), [&, this] (auto && cell)
```
h-2: in which case does a column generate items on access?

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -347,10 +431,9 @@ class alignment_algorithm :
         {
             this->next_column(); // Move to the next column.
             auto col = this->current_column();
+            this->init_row_cell(*std::ranges::begin(col), cache); // initialise first row of dp matrix.
 
-            this->init_row_cell(*seqan3::begin(col), cache); // initialise first row of dp matrix.
-
-            auto second_range_it = seqan3::begin(second_range);
+            auto second_range_it = std::ranges::begin(second_range);
```
h-2: I think these changes are not necessary since we will replace all these lines anyway?

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -360,7 +443,13 @@ class alignment_algorithm :
             });
 
             if (this->band_touches_last_row())  // TODO [[unlikely]]
-                this->check_score_last_row(get<0>(get<0>(*std::ranges::prev(seqan3::end(col)))), get<3>(cache));
+            {
+                auto [cell, coordinate, trace] = *std::ranges::prev(std::ranges::end(col));
+                std::ignore = trace;
```
h-2: ?

h-2: to ignore warnings rather do `(void*)trace;`

include/seqan3/alignment/pairwise/policy/affine_gap_banded_policy.hpp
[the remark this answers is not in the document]
h-2: `(void) prev_main_score` is guaranteed to not generate code, while the assignment is only likely to not generate code. I would prefer the former...

include/seqan3/alignment/pairwise/policy/affine_gap_policy.hpp
```diff
@@ -56,38 +57,70 @@ class affine_gap_policy
...
+     * \param[in]     score        The score of comparing the respective letters of the first and the second sequence.
      */
-    template <typename score_cell_type, typename cache_t>
-    constexpr void compute_cell(score_cell_type & current_cell,
+    template <typename matrix_entry_type, typename cache_t>
+    constexpr void compute_cell(matrix_entry_type && current_cell,
                                 cache_t & cache,
                                 score_t const score) const noexcept
     {
+        using std::get;
```
h-2: Why do we have this so often? Souldn't we just rely on ADL?

include/seqan3/alignment/pairwise/policy/banded_score_dp_matrix_policy.hpp
```diff
@@ -172,32 +198,46 @@ class banded_score_dp_matrix_policy :
...
+     * \param coordinate The coordinate to refine.
+     */
+    constexpr auto map_banded_coordinate_to_range_position(alignment_coordinate coordinate) const noexcept
+    {
+        // Refine the row coordinate to match the original sequence coordinates since the first position of the
+        // trace matrix is shifted by the value of the band_column_index, i.e. the upper bound of the band.
+        //
+        // case 1: ends in column before the band_column_index: subtract the offset from the actual row coordinate.
+        // case 2: ends in column after the band_column_index: add the offset to the actual row coordinate.
+        coordinate.second_seq_pos += static_cast<int32_t>(coordinate.first_seq_pos - band_column_index);
```
h-2: where does the `int32_t` come from? Shouldn't this be parametrized some how?

include/seqan3/alignment/pairwise/policy/banded_score_trace_dp_matrix_policy.hpp
```diff
@@ -0,0 +1,289 @@
...
+            // parse vertical gap -> record gap in first_segments (will be translated into gap of first sequence)
+            if (static_cast<bool>(*direction_iter & trace_directions::up) ||
+                static_cast<bool>(*direction_iter & trace_directions::up_open))
+            {
+                // Get the current column index (note the column based layout)
+
+                gap_segment gap{col_pos, 0u};
+
+                // Follow gap until open signal is detected.
+                while (!static_cast<bool>(*direction_iter & trace_directions::up_open))
```
h-2: Do you really need this cast? Souldn't this be contextually convertile to bool anyway?

include/seqan3/alignment/pairwise/policy/banded_score_trace_dp_matrix_policy.hpp
```diff
@@ -0,0 +1,289 @@
...
+            }
+
+            size_t col_pos = std::ranges::distance(std::ranges::begin(trace_matrix), direction_iter) / band_size;
+            // parse vertical gap -> record gap in first_segments (will be translated into gap of first sequence)
+            if (static_cast<bool>(*direction_iter & trace_directions::up) ||
+                static_cast<bool>(*direction_iter & trace_directions::up_open))
+            {
+                // Get the current column index (note the column based layout)
+
+                gap_segment gap{col_pos, 0u};
```
h-2: I have said this in other places already, as well:
It is very confusing that we refer to both a single gap character and a "stretch" of gap characters as "gap". Can you brainstorm this in the next strategy meeting to come up with distinct terms that we can use consistently in the library?

include/seqan3/alignment/matrix/alignment_coordinate.hpp
```diff
@@ -12,20 +12,346 @@
...
+
+    //!\brief Move-constructs from another advanceable_alignment_coordinate with a different policy.
+    template <advanceable_alignment_coordinate_state _state>
+    //!\cond
+        requires !std::Same<_state, state>
+    //!\endcond
+    constexpr advanceable_alignment_coordinate(advanceable_alignment_coordinate<_state> && other) :
+        first_seq_pos{std::move(other.first_seq_pos)},
+        second_seq_pos{std::move(other.second_seq_pos)}
+    {}
```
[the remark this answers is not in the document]
rrahn: So it is at the moment. Maybe it will change someday. if not moving it, it could have side effects if the position types change to something that is not builtin anymore. So from this perspective this is a very valid implementation of the move constructor.

include/seqan3/alignment/matrix/alignment_optimum.hpp
```diff
@@ -0,0 +1,86 @@
+// -----------------------------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
rrahn: Well, this is debatable. I put this on the list for the retreat.

include/seqan3/alignment/matrix/alignment_coordinate.hpp
```diff
@@ -12,20 +12,346 @@
...
+    ~alignment_coordinate() = default;
+
+    //!\cond DEV
+    //!\brief Constructs from the seqan3::detail::advanceable_alignment_coordinate base class.
+    constexpr alignment_coordinate(base_t const & base) : base_t{base}
+    {}
+
+    //!\brief Constructs from the seqan3::detail::advanceable_alignment_coordinate base class.
+    constexpr alignment_coordinate(base_t && base) : base_t{std::move(base)}
+    {}
```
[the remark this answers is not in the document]
rrahn: But it saves only the last custom ctor. the ones with the base class need to be implemented anyway.

include/seqan3/alignment/matrix/alignment_optimum.hpp
```diff
@@ -0,0 +1,86 @@
...
+alignment_optimum(score_t const, alignment_coordinate const) ->
+    alignment_optimum<std::remove_reference_t<score_t>>;
+//!\}
+
+/*!\brief A less than comparator for two seqan3::detail::alignment_optimum objects.
+ * \ingroup alignment_matrix
+ *
+ * \details
+ *
+ * This function object is used in std::max functions to compare two seqan3::detail::alignment_optimum objects.
```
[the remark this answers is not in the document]
rrahn: It is not a lambda, because it would be the same and it is a better style if lambdas are used within scoped contexts and not declared globally. Second, I might overload it for different scoring types. Keeping it encapsulated in a class (possibly class template) seems to be the right way to go.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -262,7 +298,31 @@ class alignment_algorithm :
...
+          // This can be later optimised by computing the reverse alignment in linear memory from the maximum.
+          // Especially for the SIMD version this might be more efficient.
+            res.score = get<3>(cache).score;
+            res.end_coordinate = this->map_banded_coordinate_to_range_position(get<3>(cache).coordinate);
+            res.begin_coordinate =
+                get<0>(compute_traceback(first_range,
+                                         second_range,
+                                         get<3>(cache).coordinate));
+        }
+        if constexpr (config_t::template exists<align_cfg::result<with_trace_type>>())
```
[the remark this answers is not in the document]
rrahn: Not yet, but I added already a card to change the internal representation to be able to nest it and don't repeat the expressions, while working on this.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
rrahn: Always. it is internally a zip_view.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
rrahn: IDC. I regexed. Why would I change them back?

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
rrahn: I did some reading and unfortunately it is not intended to be used this way. Also it is really an expressive way to ignore unused variables. So I might as well change it back to ugly.

include/seqan3/alignment/pairwise/policy/affine_gap_policy.hpp
rrahn: But std::get is not a niebloid. So we are relying on ADL but to do it in a safe way we need to additionally search in the std namespace if there is not get in the namespace of the variable.

include/seqan3/alignment/pairwise/policy/banded_score_trace_dp_matrix_policy.hpp
rrahn: nop

include/seqan3/alignment/pairwise/policy/banded_score_trace_dp_matrix_policy.hpp
rrahn: i opened up a discussion card for this.

rrahn:
From MPOV everything was addressed.

PR #672, opened 2019-01-14
[TEST] IO module snippets

joshuak94:
Moves examples in the IO module to the snippets subdirectory.

Some exceptions:

In io/alignment_file/output.hpp: Four examples dependent on seqan3::alignment_file_in implementation.

In io/sequence_file/output.hpp: Some snippets fail writing SEQ_QUAL. (Fixed by #671)

PR #762, opened 2019-02-21
supersedes [TEST] I/O module snippets

### LINKED ISSUE #769: Structure file bugs
While working on the io snippets #672 I encountered two bugs in structure file output:

1. Trying to write a structured_seq object causes an compiler error (same as #671 )
2. The `push_pack` overload for records was missing so outputting records lead to incorrect ordering. 

Both issues will be fixed in the snippets PR

smehringer:
supersedes #672 

Resolves #769


smehringer:
@h-2 Everything works now. There is only the alignment_file snippet that relies on alignment file input

test/snippet/io/alignment_file/alignment_file_output.cpp
```diff
@@ -0,0 +1,126 @@
...
+{
+//! [set_header]
+alignment_file_output fout{tmp_dir/"my.sam"};
+
+// add information to the header of the file.
+fout.header().comments.push_back("This is a comment");
+//! [set_header]
+}
+
+// \todo TODO: uncomment once seqan3::alignment_file_input is implemented.
```
h-2: This really has the potential to be forgotten...

test/snippet/io/alignment_file/alignment_file_output.cpp
```diff
@@ -0,0 +1,126 @@
...
+        { "GATA"_dna5, "Third" }
+    }; // a range of "records"
+
+    fout = range; // will iterate over the records and write them
+}
+
+{
+//! [input_range]
+// file format conversion in one line:
+alignment_file_output{"/tmp/output.sam"} = sequence_file_input{"/tmp/input.fasta"};
```
h-2: Shouldn't this example be 
```cpp
alignment_file_output{"/tmp/output.sam"} = alignment_file_input{"/tmp/input.bam"};
```

i.e. conversion between two formats of the same file type. I mean in the above example, it actually works to transforma sequence_file to an alignment_file, but it doesn't really make a lot of sense.

Also you still use `/tmp` here and not a real temporary directory.

test/snippet/io/sequence_file/sequence_file_input.cpp
```diff
@@ -0,0 +1,222 @@
...
+fout.emplace_back("GGAGTATAATATATATATATATAT"_dna4, "Test3");
+}
+
+{
+// Create a /tmp/my_seq_qual.fasta file.
+sequence_file_output fout{tmp_dir/"my.fastq", fields<field::ID, field::SEQ_QUAL>{}};
+std::vector<qualified<dna4, phred42>> my_qualified{qualified<dna4, phred42>{'A'_dna4, phred42{7}},
+                                                   qualified<dna4, phred42>{'C'_dna4, phred42{6}},
+                                                   qualified<dna4, phred42>{'G'_dna4, phred42{5}},
+                                                   qualified<dna4, phred42>{'T'_dna4, phred42{4}}};
```
h-2: wouldn't this also work (an be shorter):
```cpp
std::vector<qualified<dna4, phred42>> my_qualified{{'A'_dna4, phred42{7}},
                                                   {'C'_dna4, phred42{6}},
                                                   {'G'_dna4, phred42{5}},
                                                   {'T'_dna4, phred42{4}}};
```

Also the phreds have literals now ☝️

test/snippet/io/structure_file/structure_file_output.cpp
```diff
@@ -0,0 +1,254 @@
...
+//
+// for (auto & r : fin)
+// {
+//     if (criteria)// r fulfills some criterium
+//     {
+//         fout.push_back(r);
+//     }
+// }
+// //! [pass_rec]
+// }
```
h-2: whats this?

test/snippet/io/alignment_file/alignment_file_output.cpp
smehringer: well maybe FASTA -> SAm does not make sense but SAM -> FASTA actually does. There is even a samtools command especially for this. But of course coversion within the same type makes more sense in the first place... we just don't have another format :trollface:

test/snippet/io/alignment_file/alignment_file_output.cpp
smehringer: I could create an issue for it...?

test/snippet/io/alignment_file/alignment_file_output.cpp
h-2: > SAM -> FASTA actually does. There is even a samtools command especially for this. 

Yes, but that's why we have the sequence_file_format_sam in addition to te alignment_file_format_sam, or not?

test/snippet/io/alignment_file/alignment_file_output.cpp
h-2: Or leave it commented out so its visible in the docs? Or add \todo to docs?

test/snippet/io/structure_file/structure_file_output.cpp
smehringer: This just caught a fat bug.. Structure file was missing the push_back overload for records (not mere tuples) so the order fields was not correctly parsed.

smehringer:
@h-2 Can you take another look?

h-2:
Can you stash your last commits?

smehringer:
sure I justed wanted to do that after you reviewd

smehringer:
Can we merge this or shall I request another review?

PR #771, opened 2019-02-26
[FIX] Index and Search

eseiler:
- Indices do not work on `std::string`, but they should (adapt concept to check that the text is a sequence container)
- Search does not work for text collections (define type of results as `size_t` or `std::pair<size_t, size_t>` depending on whether the index is built over a single text or text collection)

Todo:

- [x] Add tests

### DIFF
--- a/include/seqan3/search/algorithm/detail/search.hpp
+++ b/include/seqan3/search/algorithm/detail/search.hpp
@@ -13,6 +13,7 @@
 #pragma once
 
 #include <seqan3/core/metafunction/pre.hpp>
+#include <seqan3/search/algorithm/configuration/all.hpp>
 #include <seqan3/search/algorithm/detail/search_scheme_algorithm.hpp>
 #include <seqan3/search/algorithm/detail/search_trivial.hpp>
 #include <seqan3/search/fm_index/concept.hpp>
@@ -129,7 +130,11 @@ inline auto search_single(index_t const & index, query_t & query, configuration_
     }
     else
     {
-        std::vector<typename index_t::size_type> hits;
+        using hit_t = std::conditional_t<index_t::is_collection,
+                                         std::pair<typename index_t::size_type, typename index_t::size_type>,
+                                         typename index_t::size_type>;
+        std::vector<hit_t> hits;
+
         if constexpr (cfg_t::template exists<search_cfg::mode<detail::search_mode_best>>())
         {
             // only one cursor is reported but it might contain more than one text position
@@ -179,9 +184,12 @@ inline auto search_all(index_t const & index, queries_t & queries, configuration
     // delegate params: text_position (or cursor). we will withhold all hits of one query anyway to filter
     //                  duplicates. more efficient to call delegate once with one vector instead of calling
     //                  delegate for each hit separately at once.
+    using text_pos_t = std::conditional_t<index_t::is_collection,
+                                          std::pair<typename index_t::size_type, typename index_t::size_type>,
+                                          typename index_t::size_type>;
     using hit_t = std::conditional_t<cfg_t::template exists<search_cfg::output<detail::search_output_index_cursor>>(),
                                      typename index_t::cursor_type,
-                                     typename index_t::size_type>;
+                                     text_pos_t>;
 
     if constexpr (std::ranges::ForwardRange<queries_t> && std::ranges::RandomAccessRange<value_type_t<queries_t>>)
     {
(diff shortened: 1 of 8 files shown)


test/unit/search/fm_index/fm_index_test_template.hpp
```diff
@@ -100,9 +101,10 @@ TYPED_TEST_P(fm_index_test, serialization)
 
 TYPED_TEST_P(fm_index_test, concept_check)
 {
-    EXPECT_TRUE(FmIndex<TypeParam>);
-    EXPECT_TRUE(FmIndexTraits<fm_index_default_traits>);
-    EXPECT_TRUE(BiFmIndexTraits<bi_fm_index_default_traits>);
+    if constexpr (detail::is_type_specialisation_of_v<TypeParam, fm_index>)
+        EXPECT_TRUE(FmIndex<TypeParam>);
+    else
+        EXPECT_TRUE(BiFmIndex<TypeParam>);
```
joergi-w: Does a `bi_fm_index` also satisfy the `FmIndex` concept? (which I would expect)
If yes, then both if/else branches should be executed for this type.

eseiler: `bi_fm_index` also satisfies the `FmIndex`, but the `BiFmIndex` is a refinement of `FmIndex`, i.e. it requires that the index is an `FmIndex` and then some more stuff.
So I think it is fine this way?

joergi-w: yes, okay!

include/seqan3/search/algorithm/detail/search.hpp
```diff
@@ -179,9 +184,12 @@ inline auto search_all(index_t const & index, queries_t & queries, configuration
     // delegate params: text_position (or cursor). we will withhold all hits of one query anyway to filter
     //                  duplicates. more efficient to call delegate once with one vector instead of calling
     //                  delegate for each hit separately at once.
+    using text_pos_t = std::conditional_t<index_t::is_collection,
```
rrahn: like here it seems a frequent pattern to ask such a thing. I am wondering if there is something like `index_traits<index_t>::...` If not, I think it seems a good thing to do. If relevant can you make a card for the strategy meeting.

include/seqan3/search/fm_index/concept.hpp
```diff
@@ -129,10 +129,12 @@ SEQAN3_CONCEPT FmIndex = std::Semiregular<t> && requires (t index)
     typename t::size_type;
     typename t::cursor_type;
 
+    sequence_container_concept<typename t::text_type>;
```
rrahn:
```suggestion
    requires sequence_container_concept<typename t::text_type>;
```

test/unit/search/CMakeLists.txt
```diff
@@ -6,3 +6,4 @@ seqan3_test (search_configuration_test.cpp)
 seqan3_test (search_scheme_test.cpp)
 seqan3_test (search_scheme_algorithm_test.cpp)
 seqan3_test (search_test.cpp)
+seqan3_test (search_collection_test.cpp)
```
rrahn: alphabetical order

test/unit/search/fm_index/fm_index_test_template.hpp
rrahn: So why not just testing:
```cpp    
EXPECT_TRUE(FmIndex<TypeParam>);
if constexpr (detail::is_type_specialisation_of_v<TypeParam, bi_fm_index>)
    EXPECT_TRUE(BiFmIndex<TypeParam>);
```
?

test/unit/search/search_collection_test.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+template <typename T>
+class search_test : public ::testing::Test
+{
+public:
+    std::vector<std::vector<dna4>> text{"ACGTACGTACGT"_dna4, "ACGTACGTACGT"_dna4};
+    T index{text};
+};
+
+using fm_index_types = ::testing::Types<fm_index<std::vector<std::vector<dna4>>>>;
```
rrahn: and bi_fm_index?

include/seqan3/search/algorithm/detail/search.hpp
eseiler: I'll add a card for now, but it's more a trait of the cursor than the index itself

PR #772, opened 2019-02-26
Feature/pairwise combine view

### LINKED ISSUE #768: Feature:  add pairwise_combine_view
Add a pairwise_combine_view, that given an input range produces a view with all pairwise combinations of it's elements. 

Requirements:

The view requires a bounded view and at least ForwardRange.
The view preserves the iterator category of the underlying Range.
The view always guarantees a bounded range
The output is a tuple with exactly two elements.
The tuple contains the return values from dereferencing the respective iterator of the underlying range and propagates the reference type through the returned tuple.

Example:

```cpp
std::vector v{'a', 'b', 'c', 'd'};
for (auto res : v | view::pairwise_combine)
{
    debug_stream << res << "\n";
}
```

The expected output is:
```console
[a, b]
[a, c]
[a, d]
[b, c]
[b, d]
[c, d]
```

rrahn:
Fixes #768 

This is the first implementation of the pairwise_combine view to generate all pairwise generations. 
In the current state it is the basic implementation to support forward ranges. 
The following things will be added successively:

### TODO 
 - [x] Make view model bidirectional if source range is bidirectional
 - [x] Make view model random access if source range is random access
 - [x] Make view sized if source ranges is sized/random access
 - [x] Implement view adaptor
 - [x] make a snippet


test/unit/range/view/view_pairwise_combine_test.cpp
```diff
@@ -0,0 +1,270 @@
...
+class pairwise_combine_base_test : public ::testing::Test
+{
+public:
+
+    using view_t       = decltype(seqan3::detail::pairwise_combine_view{std::declval<t &>()});
+    using const_view_t = decltype(seqan3::detail::pairwise_combine_view{std::declval<t const &>()});
+
+    auto create_view()
+    {
+        return view_t{container | seqan3::view::all};
```
smehringer: I thought the constructor will wrap a container in view all itself. You should be able to construct from any container then without the pipe

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,307 @@
...
+{
+/*!\brief Generates all pairwise combinations of the elements in the source range.
+ * \ingroup view
+ * \tparam source_range_type The type of the source range; must model std::ranges::View, std::ranges::ForwardRange
+ *
+ * \details
+ *
+ * This view will provide a convenient way to iterate over all pairwise combinations of a given source range
+ * while the order of selection does not matter. A source range with `n` elements will therefore have `n choose 2`
+ * possible combinations.
```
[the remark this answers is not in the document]
rrahn: Sure, it is on the todo list for the adaptor anyway.

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,307 @@
...
+        constexpr iterator_type() = default;
+        constexpr iterator_type(iterator_type const &) = default;
+        constexpr iterator_type(iterator_type &&) = default;
+        constexpr iterator_type & operator=(iterator_type const &) = default;
+        constexpr iterator_type & operator=(iterator_type &&) = default;
+        ~iterator_type() = default;
+
+        /*!\brief Constructs the iterator from the iterator pointing to the last element and the iterator to the first
+         *        position.
+         * \param[in] i The iterator pointing to the first element of the underlying range.
```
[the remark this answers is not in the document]
rrahn: i think of a more verbose name

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,307 @@
...
+        /*!\name Accessors
+         * \{
+         */
+         //!\brief Accesses the pointed-to element
+        constexpr reference operator*() const
+            noexcept(noexcept(*std::declval<source_iterator_type>()))
+        {
+            return {*first_it, *second_it};
+        }
+        //!\}
```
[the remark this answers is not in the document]
rrahn: No we cannot, since we are generating temporary tuples. The operator-> returns a pointer to the pointed-to element but in this case we cannot take the address of the temporary

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,307 @@
...
+         * \{
+         */
+        //!\brief Pre-increment operator.
+        constexpr iterator_type & operator++(/*pre-increment*/)
+            noexcept(noexcept(++std::declval<source_iterator_type>()))
+        {
+            if (++second_it == end_it)
+            {
+                ++first_it;
+                second_it = first_it;
```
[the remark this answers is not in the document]
rrahn: Right now, there are 2 increments and 1 copy-assignment.
Both of your solutions would introduce an additional move-assignment. The compiler will be probably smart enough to sort it out anyway, but is it more readable/clear to what happens here?

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,307 @@
...
+     *
+     * \details
+     *
+     * During construction the iterator pointing to the last element of the view is cached. This optimises
+     * the call to end if the underlying range models only ForwardRange. Otherwise the call to end will be
+     * linear in the number of elements of the underlying range.
+     */
+    template <typename other_range_type>
+        requires std::is_same_v<other_range_type, source_range_type>
+    explicit constexpr pairwise_combine_view(other_range_type && range) : src_range{std::move(range)}
```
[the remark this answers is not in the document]
rrahn: But I need to use the template argument. The issues with this is that there is some race with the move constructor but if it is not constrained  then anything can be put here. So I constrain it to be the same type as the source_range_type which is anyway only relevant if someone tries to call the constructor without deduction guides and out of its actual use with the view adaptor.

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,307 @@
...
+    /*!\brief Constructs from a range.
+     * \param[in] range A lvalue reference to the underlying range.
+     *
+     * \details
+     *
+     * Wraps the passed range in a ref_view in order to store lvalue references to ranges (non-views) as a view
+     * and delegate to the view constructor above.
+     */
+    template <typename other_range_type>
+    explicit constexpr pairwise_combine_view(other_range_type & range) : pairwise_combine_view{view::all(range)}
```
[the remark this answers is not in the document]
rrahn: Yes

test/unit/range/view/view_pairwise_combine_test.cpp
rrahn: good point 😅

include/seqan3/range/view/pairwise_combine.hpp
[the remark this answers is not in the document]
smehringer: Maybe we should document then that the iterator does not model the LegacyForwardIterator from the stl but the ranges::ForwardIterator? I just had a similar case

include/seqan3/range/view/pairwise_combine.hpp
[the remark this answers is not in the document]
smehringer: ah ok. a comment would be good then :)
and you need to wrap the requires in `//!\cond`

include/seqan3/range/view/pairwise_combine.hpp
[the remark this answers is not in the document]
smehringer: How :open_mouth: ? I would think it will cause a `no function to call, candidate: casting const other_range to other_range discards qualifiers`..

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,329 @@
...
+         * \param[in] iter   The iterator pointing to current element within the source range.
+         * \param[in] end_it The iterator pointing to end of the source range.
+         *
+         * \details
+         *
+         * Constructs the iterator by caching the end of the source range and setting the first iterator to the
+         * given position and the second to the first incremented by one.
+         */
+        constexpr iterator_type(source_iterator_type iter,
+                                source_iterator_type end_it) noexcept :
```
smehringer: Pff you thumbed :+1: my review request but the indentation is still weird...

include/seqan3/range/view/pairwise_combine.hpp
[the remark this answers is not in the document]
rrahn: I kind of have to correct my previous statement. I thought I need the addtional template because of the forward reference, but in the case if I use the class template it is only an rvalue reference and this should be fine.

include/seqan3/range/view/pairwise_combine.hpp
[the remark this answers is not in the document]
rrahn: But the const is part of the template and thus propagated to the source_range_type.

include/seqan3/range/view/pairwise_combine.hpp
rrahn: Everything seems perfectly aligned to me? I just wrapped each initialiser on its own line. Since when is this weird?

include/seqan3/range/view/pairwise_combine.hpp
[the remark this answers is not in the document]
rrahn: Since we do not require LegacyIterators I don't see why this is necessary.

rrahn:
@smehringer I added now all the remaining features.

include/seqan3/range/view/pairwise_combine.hpp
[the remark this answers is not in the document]
smehringer: not resolved :nerd_face:

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,597 @@
...
+            noexcept(noexcept(*std::declval<source_iterator_type>()))
+        {
+            return {*first_it, *second_it};
+        }
+
+        /*!\brief Access the element at the given index
+         * \param[in] index The index of the element to be returned.
+         */
+        constexpr reference operator[](size_t const index)
+            // noexcept(noexcept(std::declval<iterator_type &>().from_index()))
```
smehringer: ?

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,597 @@
...
+        //!\endcond
+        {
+            size_t src_size = std::ranges::distance(begin_it, end_it);
+            size_t index_i = src_size - 2 -
+                             std::floor(std::sqrt(-8 * index + 4 * src_size * (src_size - 1) - 7)/2.0 - 0.5);
+            size_t index_j = index + index_i + 1 - src_size * (src_size - 1)/2 + (src_size - index_i) *
+                             ((src_size - index_i) - 1)/2;
+            first_it = begin_it;
+            second_it = begin_it;
+            std::ranges::advance(first_it, index_i);
```
smehringer: If you require randomAccessIterator you don't need advance?

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,597 @@
...
+            noexcept(noexcept(std::ranges::distance(std::declval<source_iterator_type &>(),
+                                                    std::declval<source_iterator_type &>())) &&
+                     noexcept(std::ranges::advance(std::declval<source_iterator_type &>(), 1)))
+        //!\cond
+            requires std::RandomAccessIterator<source_iterator_type>
+        //!\endcond
+        {
+            size_t src_size = std::ranges::distance(begin_it, end_it);
+            size_t index_i = src_size - 2 -
+                             std::floor(std::sqrt(-8 * index + 4 * src_size * (src_size - 1) - 7)/2.0 - 0.5);
```
smehringer: pfew.. did you calculate that yourself and can comment this or is there a reference you can link?

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,597 @@
...
+     * No-throw guarantee.
+     */
+    constexpr iterator_type begin() noexcept
+    {
+        return iterator_type{std::ranges::begin(src_range), std::ranges::begin(src_range), std::ranges::end(src_range)};
+    }
+
+    //!\copydoc begin()
+    constexpr iterator_type begin() const noexcept
+        requires const_iterable_concept<source_range_type>
```
smehringer: did you check if doxygen ignores the requires? sometimes it does, sometimes not..

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,597 @@
...
+ * \details
+ *
+ * This view generates two-element tuples representing all possible combinations of the elements of the source range
+ * while ignoring the order of the elements. If the source range has less than two elements the returned range is empty,
+ * otherwise the size of the returned view corresponds to the binomial coefficient `n choose 2`, where `n` is the
+ * size of the source range. The reference type of this range is a tuple over the reference type of the source range.
+ * Constness will not be propagated to the reference types of the source range.
+ *
+ * ### View properties
+ *
```
smehringer: remove newline

include/seqan3/range/view/pairwise_combine.hpp
smehringer: well now that its three parameters that don't fit in one line it's fine, but if function parameters fit into one line we usually put them in one line.. and I never saw the parameter initialisation in multiple lines either when they fit into one line. But I guess that's not a style guide requirement anyway so sorry for the request.

rrahn: No it is ok, but I always put them on single line once they cannot be put immediately behind the colon. But as you said we don’t have a guide for this

include/seqan3/range/view/pairwise_combine.hpp
rrahn: There is a link to the thread in the details section from the to_index function, which this function copies

include/seqan3/range/view/pairwise_combine.hpp
rrahn: I removed it completely, since the view does not propagate const to the source range.

include/seqan3/range/view/pairwise_combine.hpp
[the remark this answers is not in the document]
smehringer: @h-2 ? You just made me document it so we should decide on one or the other

rrahn:
@h-2 travis is done and green but somehow the status was not updated.

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,595 @@
...
+ * \returns           A range over all pairwise combinations. See below for the properties of the returned range.
+ * \ingroup view
+ *
+ * \details
+ *
+ * This view generates two-element tuples representing all possible combinations of the elements of the source range
+ * while ignoring the order of the elements. If the source range has less than two elements the returned range is empty,
+ * otherwise the size of the returned view corresponds to the binomial coefficient `n choose 2`, where `n` is the
+ * size of the source range. The reference type of this range is a tuple over the reference type of the source range.
+ * Constness will not be propagated to the reference types of the source range.
```
h-2: I think it does propagate const-ness of views (but not of non-views). This is also the behaviour of range-v3.

You are relying on `using source_iterator_type = std::ranges::iterator_t<source_range_type>;` which gives you a const_iterator if the underlying range type is const. 

Btw. is there a reason you chose not to use "underlying range type" like in all our other view and instead "source range type"?

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,595 @@
...
+ *
+ * ### View properties
+ *
+ * | range concepts and reference_t  | `urng_t` (underlying range type)      | `rrng_t` (returned range type)                                       |
+ * |---------------------------------|:-------------------------------------:|:--------------------------------------------------------------------:|
+ * | std::ranges::InputRange         |                                       | *undefined*                                                          |
```
h-2: The convention for other views was to write *required* | *preserved* for these cases, too. 
Of course Input is required, if Forward is required. It is debatable whether it is then also guaranteed, but like I said the style for the others is to write the above.

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,595 @@
...
+ * | std::ranges::RandomAccessRange  |                                       | *preserved*                                                          |
+ * | std::ranges::ContiguousRange    |                                       | *lost*                                                               |
+ * |                                 |                                       |                                                                      |
+ * | std::ranges::ViewableRange      | *required*                            | *guaranteed*                                                         |
+ * | std::ranges::View               |                                       | *guaranteed*                                                         |
+ * | std::ranges::SizedRange         |                                       | *guaranteed* iff urng_t models std::RandomAccessRange                |
+ * | std::ranges::CommonRange        |                                       | *guaranteed*                                                         |
```
h-2: As I mentioned previously, I think you are requiring this implicitly.

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,595 @@
...
+ * | std::ranges::BidirectionalRange |                                       | *preserved*                                                          |
+ * | std::ranges::RandomAccessRange  |                                       | *preserved*                                                          |
+ * | std::ranges::ContiguousRange    |                                       | *lost*                                                               |
+ * |                                 |                                       |                                                                      |
+ * | std::ranges::ViewableRange      | *required*                            | *guaranteed*                                                         |
+ * | std::ranges::View               |                                       | *guaranteed*                                                         |
+ * | std::ranges::SizedRange         |                                       | *guaranteed* iff urng_t models std::RandomAccessRange                |
```
h-2: This is not required, but you guarantee it? How? What about an infinite random access range as underlying type?

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,595 @@
...
+ * | std::ranges::CommonRange        |                                       | *guaranteed*                                                         |
+ * | std::ranges::OutputRange        |                                       | *lost*                                                               |
```
h-2: Actually I think you are preserving this, because you are reference type can be promoted to an `xvalue`:

```cpp
#include <tuple>

int i = 1;
int j = 2;

std::tuple<int&, int&> foo()
{
    return {i, j};
}

int main()
{
    int k = 3;
    int l = 4;
    foo() = std::tie(k, l);
    return i; // returns 3
}
```

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,595 @@
...
+ * | seqan3::const_iterable_concept  |                                       | *preserved*                                                          |
+ * |                                 |                                       |                                                                      |
+ * | seqan3::reference_t             |                                       | std::tuple<seqan3::reference_t<urng_t>, seqan3::reference_t<urng_t>> |
+ *
+ * See the \link view view submodule documentation \endlink for detailed descriptions of the view properties.
+ *
+ * ### Thread safety
+ *
+ * Concurrent access to this view, e.g. while iterating over it, is not thread-safe and must be protected externally.
```
h-2: This sounds like it is not safe to iterate over the view concurrently, but it is, or not? If not, you can't mark this as preserving const-iterable.

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,595 @@
...
+namespace seqan3::detail
+{
+/*!\brief Generates all pairwise combinations of the elements in the source range.
+ * \ingroup view
+ * \tparam source_range_type The type of the source range; must model std::ranges::View, std::ranges::ForwardRange
+ *
+ * \details
+ *
+ * This view will provide a convenient way to iterate over all pairwise combinations of a given source range
+ * while the order of selection does not matter. A source range with `n` elements will therefore have `n choose 2`
```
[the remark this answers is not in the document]
rrahn: Actually, while has many meanings and is not necessarily a synonym for as long as. But I will see that I revise the sentence as it sounds clumpy anyway.

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,595 @@
...
+         * \param[in] index The index of the element to be returned.
+         */
+        constexpr reference operator[](size_t const index)
+            noexcept(noexcept(std::declval<iterator_type &>().from_index(1)))
+        //!\cond
+            requires std::RandomAccessIterator<source_iterator_type>
+        //!\endcond
+        {
+            from_index(index);
+            return this->operator*();
```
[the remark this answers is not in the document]
rrahn: IDC

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,595 @@
...
+ * while the order of selection does not matter. A source range with `n` elements will therefore have `n choose 2`
+ * possible combinations.
+ *
+ * ### Example
+ *
+ * \include test/snippet/range/view/pairwise_combine.cpp
+ */
+template <std::ranges::View source_range_type>
+//!\cond
+    requires  std::ranges::ForwardRange<source_range_type>
```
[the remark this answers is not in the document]
rrahn: Right. Actually I had this before but then dropped it because I thought this restriction wouldn't be necessary. But I also did not implement it right then. But after doing some more reading on the topic I came to the conclusion that this requirement does not really end up to be a restriction, given that only non-common bidirectional ranges would be downgraded to forward ranges if converted to a common range, but need a linear construction anyway. Just they can't be decremented anymore. I can live with that.

include/seqan3/range/view/pairwise_combine.hpp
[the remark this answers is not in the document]
rrahn: 🤦‍♂️ I was just irritated because I was looking at the operator->, since it won't be used inside of stl::algorithms anyway. But now I remember, that the Cpp17ForwardIterator requires that the operator* returns a reference to the value type, which is in fact not guaranteed. So my first impression was that the users should not be bugged with even more concepts/requirements and dependencies that they have to understand, but this has far-reaching implications.

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,595 @@
...
+        //!\brief Pre-increment operator.
+        constexpr iterator_type & operator++(/*pre-increment*/)
+            noexcept(noexcept(++std::declval<source_iterator_type &>()))
+        {
+            if (++second_it == end_it)
+            {
+                ++first_it;
+                second_it = first_it;
+                ++second_it;
+            }
```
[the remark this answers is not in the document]
rrahn: I don't understand. The begin is always modeled as `it_first = begin_it` and `it_second = next(it_first)` So the above code guarantees that `it_second is always > `it_first`. So as long as `++second_it != end_it` we don't have to do anything, but note, that `it_second` was incremented in this expression and it holds `it_second > next(it_first)`. If `it_second == end_it` after incrementing `it_second` we increment `it_first` and update `it_second` to `next(it_first)`. So only when `it_second` reached the end of the range both iterators are consecutive

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,595 @@
...
+
+        /*!\name Comparison operators
+         * \brief These operators are available if the `source_iterator_type` models std::EqualityComparable or
+         *        std::StrictTotallyOrdered respectively.
+         * \{
+         */
+        constexpr friend bool operator==(iterator_type const & lhs, iterator_type const & rhs)
+            noexcept(noexcept(std::declval<source_iterator_type &>() == std::declval<source_iterator_type &>()))
+        //!\cond
+            requires std::EqualityComparable<source_iterator_type>
```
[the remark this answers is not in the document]
rrahn: that is unfortunately true and I did not thought about it when implementing the operators. But since the operators constraints are not participating in any overload resolution it works without the constraints as well. Only the emitted compiler error will be not as expected until the bug is resolved. So I tend to keep it this way as it is the suggested way to implement the operators.

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,595 @@
...
+        constexpr size_t to_index() const
+            noexcept(noexcept(std::ranges::distance(std::declval<source_iterator_type &>(),
+                                                    std::declval<source_iterator_type &>())))
+        //!\cond
+            requires std::RandomAccessIterator<source_iterator_type>
+        //!\endcond
+        {
+            size_t src_size = std::ranges::distance(begin_it, end_it);
+            size_t index_i = std::ranges::distance(begin_it, first_it);
+            size_t index_j = std::ranges::distance(begin_it, second_it);
```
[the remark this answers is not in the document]
rrahn: Basically yes. I thought it is more readable but IDC and change it to -. Yet it is inconsistent with from_index, so I will change it to make it at least consistent.

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,595 @@
...
+                     noexcept(std::declval<source_iterator_type &>() + 1))
+        //!\cond
+            requires std::RandomAccessIterator<source_iterator_type>
+        //!\endcond
+        {
+            size_t src_size = end_it - begin_it;
+            size_t index_i = src_size - 2 -
+                             std::floor(std::sqrt(-8 * index + 4 * src_size * (src_size - 1) - 7)/2.0 - 0.5);
+            size_t index_j = index + index_i + 1 - src_size * (src_size - 1)/2 + (src_size - index_i) *
+                             ((src_size - index_i) - 1)/2;
```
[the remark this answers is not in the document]
rrahn: Be my guest and find one. 👀

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,595 @@
...
+            if constexpr (std::ranges::BidirectionalRange<source_range_type> &&
+                          std::ranges::CommonRange<source_range_type>)
+            {
+                back_iterator = std::ranges::prev(std::ranges::end(src_range));
+            }
+            else
+            {  // Cannot decrement so we invoke linear pass until we get to the last element.
+                size_t size = std::distance(src_range);
+                back_iterator = std::ranges::begin(src_range);
+                std::advance(back_iterator, size - 1);
```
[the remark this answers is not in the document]
rrahn: But this would only be true for non-sized forward ranges, and it is the question if it is worth the optimisation to do this.

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,595 @@
...
+    /*!\brief Constructs from a range.
+     * \param[in] range A lvalue reference to the source range.
+     *
+     * \details
+     *
+     * Wraps the passed range in a ref_view in order to store lvalue references to ranges (non-views) as a view
+     * and delegate to the view constructor above.
+     */
+    template <typename other_range_type>
+    explicit constexpr pairwise_combine_view(other_range_type & range) : pairwise_combine_view{view::all(range)}
```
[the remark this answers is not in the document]
rrahn: Well explicit is a habit. They are not constrained because if they are it breaks a lot of stuff since it won't be move constructible anymore. I don't have the time to investigate this so we need to live with this for the moment.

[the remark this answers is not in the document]
rrahn: Also note, that this constructor was added because of the C++20 standard. But AFAIK these are not implemented in the current range versions (at least random lookups didn't reveal any hint on how/where they might be implemented. Don't know if this is tackled in future versions of the ranges library or if you know more about, then please enlighten me. But I can also just leave this ctor away because it just gives me PITA that I don't have time for.

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,595 @@
...
+ */
+
+//!\brief Deduces the source range type from the passed view.
+template <std::ranges::View other_type>
+pairwise_combine_view(other_type view) -> pairwise_combine_view<std::remove_reference_t<other_type>>;
+
+//!\brief Deduces the source range type from the passed range.
+template <std::ranges::Range range_type>
+    requires !std::ranges::View<range_type>
+pairwise_combine_view(range_type & range) -> pairwise_combine_view<decltype(std::declval<range_type &>() | view::all)>;
```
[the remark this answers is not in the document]
rrahn: What do you mean with they are not fully constrained? They are anyway a result of the annoying move-constructable thing that I rather don't have.

include/seqan3/range/view/pairwise_combine.hpp
rrahn: This case is explicitly prevented by the WritableConcept! See http://eel.is/c++draft/iterator.concepts#iterator.concept.writable-6

include/seqan3/range/view/pairwise_combine.hpp
rrahn: Copy'n'paste error.

include/seqan3/range/view/pairwise_combine.hpp
rrahn: How about distance of an infinite random access range?

include/seqan3/range/view/pairwise_combine.hpp
rrahn: Const is not propagated! Calling begin on a const pairwise_combine_view does not invoke a const_iterator of this view (an iterator where the source range is explicitly marked as const). So the const qualification of the source range will not be changed. Did I misunderstand this somehow? Is this covered differently in the standard/range?
Please enlighten me.

rrahn: >Btw. is there a reason you chose not to use "underlying range type" like in all our other view and instead "source range type"?
This seems to be the typical wording in the user documentation of the ranges library.

include/seqan3/range/view/pairwise_combine.hpp
rrahn: IDC

include/seqan3/range/view/pairwise_combine.hpp
rrahn: Explicitly even 😉

include/seqan3/range/view/pairwise_combine.hpp
rrahn: ja, ja, ja, ... I know it does not mean that it is sized.
But honestly, this view won't work with infinity ranges anyway, because upon construction it will take forever to get to the end - 1 position unless it was made finite before.

include/seqan3/range/view/pairwise_combine.hpp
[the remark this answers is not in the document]
rrahn: Ok, I did even more digging and the strangest things happen. I did find the biggest problem and resolved it, but there are still some open investigations that need to be done. See the todo comment that I left in the code.

include/seqan3/range/view/pairwise_combine.hpp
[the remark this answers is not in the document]
rrahn: These are now fixed after I fixed the second constructor.

include/seqan3/range/view/pairwise_combine.hpp
rrahn: I did changed it back in the public API. I don't want to start a general discussion about this now.

include/seqan3/range/view/pairwise_combine.hpp
rrahn: I added a block to make clear that it won't work with infinite ranges .

include/seqan3/range/view/pairwise_combine.hpp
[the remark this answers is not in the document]
h-2: Having interdependencies between the different range concepts is difficult to understand for the user and error-prone to program. I agree though that it is often easier to do and our focus should be in getting the features out.
Do I understand correctly that currently preserving size and anything above forward-range requires CommonRange? Then I would propose one of the following:
1. clearly documenting this in the table by writing *preserved iff CommonRange<urng_t>* into the column; and adding tests that guarantee that the view still works for all types of input (i.e. properly downgrades a Bidirectional-but-not-Common-Range to Forward).
2. Just require CommonRange for everything. That will make it easier to test and document. People can still use `| view::common` at the cost of some performance...

Regarding the preservedness of SizedRange: 
I still don't think you can guarantee it, you can only preserve it in certain cituations. What are the rules for preserving? Is it only preserved for RandomAccess (
RandomAccessRange+CommonRange implies SizedRange so you don't need extra documentation for that). What about Bidirectional+Sized+Common?

include/seqan3/range/view/pairwise_combine.hpp
h-2: > Calling begin on a const pairwise_combine_view does not invoke a const_iterator of this view

I expect this will cause problems. Have you tried const-iterating your view? Does it work? You expose references to the underlying range sometimes, I expect this will fail in a const-context, because you can't expose non-const-references (which the non-const iterator creates) to a const-member.
You can have a look at other view implementations like view::take to see how this is handled. If you don't change this it will result in your view non being const-iterable unless the underlying range is by chance also of const-type (in which case it would work, because your iterator delegates to a const-iterator anyway and the exposed references would be const references).

> This seems to be the typical wording in the user documentation of the ranges library.

But it is different from the wording we agreed on for SeqAn3 very early (I remember that we had a long discussion on this). And we use "underlying range" consistently throughout all of the view documentation, the range landing page etc etc. We can discuss if we want to change it, but I would really like to keep it consistent for now...

include/seqan3/range/view/pairwise_combine.hpp
h-2: Out of curiosity: in theory it could work for non-common infinite random access ranges, though, or not? It would just creates infinit pairs with the first element, though...

In practice: Is there a way to check from the code and throw an exception on construction? Infinite run-time loop is quite hard to catch. Then again if you formally require CommonRange, we exclude inifinite ranges anyway, or not? 

Actually no. We could have a common forward view where end is just some iterator of the same type that always compares false thus modelling an infinite common range... 🤔

include/seqan3/range/view/pairwise_combine.hpp
h-2: :+1:

Although there will be some hack before C++20 so exactly this will work for view::zip. The hack might be a custom type or actually adding const-qualified assignment to tuples of references 😮

include/seqan3/range/view/pairwise_combine.hpp
rrahn: >In practice: Is there a way to check from the code and throw an exception on construction? Infinite run-time loop is quite hard to catch. Then again if you formally require CommonRange, we exclude inifinite ranges anyway, or not?
Actually no. We could have a common forward view where end is just some iterator of the same type that always compares false thus modelling an infinite common range... 🤔

Exactly. CommonRange just wraps the iterator and the sentinel but resolves to the comparison of the wrapped objects. Thus, an infinite range can be a common range. I could refactor the code, such that I don't need to scan for the last element before, but only for the sake of efficiency for the increment and decrement operators. Though, the runtime should be compared before making premature assumptions. But since it does not break API I would postpone this.

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,618 @@
...
+     * Constant if `other_range_t` models std::ranges::BidirectionalRange, otherwise linear.
+     */
+    template <typename other_range_t>
+    //!\cond
+        requires !std::Same<remove_cvref_t<other_range_t>, pairwise_combine_view> &&
+                 std::ranges::ViewableRange<other_range_t> &&  // Must come after self type check to avoid conflicts with the move constructor.
+                 std::Constructible<source_range_type, ranges::ref_view<std::remove_reference_t<other_range_t>>>
+            //TODO: Investigate: the following expression is equivalent to the one above but raises a weird assertion in
+            //      the ranges adaptor suggesting that the pairwise_combine_view is not a ViewableRange.
+            //      std::Constructible<source_range_type, decltype(view::all(std::declval<other_range_t &&>()))>
```
h-2: Don't know if that was the issue, but did you try to `remove_reference_t<other_range_t>`? Otherwise reference collapsing will remove your `&&` here.

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,618 @@
...
+ * \tparam urng_t     The type of the range being processed. See below for requirements.
+ * \param[in] urange  The range being processed.
+ * \returns           A view over all pairwise combinations of the elements of the underlying range.
+ *                    See below for the properties of the returned range.
+ * \ingroup view
+ *
+ * \details
+ *
+ * This view generates two-element tuples representing all unique combinations of the elements of the underlying range
+ * (the order of the elements does not matter). If the underlying range has less than two elements the returned range is
```
h-2: `(the order of elements is undefined)` ?
"does not matter" is very colloquial. 💅

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,618 @@
...
+ * the size of the underlying range. The reference type of this range is a tuple over the reference type of the
+ * underlying range. This range is const-iterable as long as the underlying range is iterable (const-ness of this range
+ * will not be propagated to the underlying range).
+ * In order to receive the end iterator in constant time an iterator pointing to the last element of the underlying
+ * range will be cached upon construction of this view. This construction takes linear time for underlying ranges that
+ * do not model std::ranges::BidirectionalRange.
+ *
+ * ### Iterator
+ *
+ * The returned iterator from begin does not model std::LegacyIterator, also known as `Cpp17Iterator` since it
```
h-2: see above, I think it would be better to explicitly link to cppreferences as "std::LegacyIterator" is not really an entity.

include/seqan3/range/view/pairwise_combine.hpp
h-2: I did some reading and we can at least assume that a bidirectional common iterator is not infinite, because decrementing the end would be undefined. For this reason `view::common` also downgrades everything to at most ForwardRange.

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,618 @@
...
+ * ### Thread safety
+ *
+ * Concurrent access to this view, e.g. while iterating over it, is thread-safe and must not be protected externally.
+ *
+ * ### Example
+ *
+ * \include test/snippet/range/view/pairwise_combine.cpp
+ *
+ * \attention This view cannot be chained immediately with an infinite range, because upon construction it will
+ *            take forever to reach the last element of the view.
```
h-2: Seeing all the difficulties we already have with specifiying the behaviour of this view, can we create a TODO-card to remove the requirement for CommonRange and the inability to handle infinite ranges?

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,618 @@
...
+ *
+ * ### Example
+ *
+ * \include test/snippet/range/view/pairwise_combine.cpp
+ */
+template <std::ranges::View source_range_type>
+//!\cond
+    requires  std::ranges::ForwardRange<source_range_type> && std::ranges::CommonRange<source_range_type>
+// !\endcond
+class pairwise_combine_view : public ranges::view_interface<pairwise_combine_view<source_range_type>>
```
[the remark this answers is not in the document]
rrahn: We are missing quite a lot in our std module.

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,618 @@
...
+        constexpr friend bool operator<=(iterator_type const & lhs, iterator_type const & rhs)
+            noexcept(noexcept(std::declval<source_iterator_type &>() <= std::declval<source_iterator_type &>()))
+        //!\cond
+            requires std::StrictTotallyOrdered<source_iterator_type>
+        //!\endcond
+        {
+            return std::tie(lhs.first_it, lhs.second_it) <= std::tie(rhs.first_it, rhs.second_it);
+        }
+
+        constexpr friend bool operator>=(iterator_type const & lhs, iterator_type const & rhs)
```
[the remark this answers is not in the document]
rrahn: Can you be please a little bit more explicit. This code is conform with the standard implementation of iterators. Why would the compiler deoptimise when dealing with read-only values?

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,618 @@
...
+     * \param[in] range The source range to be wrapped. Of type `source_range_type`.
+     *
+     * \details
+     *
+     * During construction the iterator pointing to the last element of the view is cached (not the end of the range).
+     * This optimises the call to end if the source range models only ForwardRange. Otherwise the call to end will be
+     * linear in the number of elements of the source range.
+     *
+     * \attention This view cannot be chained immediately after an infinite range, because upon construction it will
+     *            take forever to reach the last element of the view.
```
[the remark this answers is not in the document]
rrahn: >The range library has this issue as well, and they have an undocumented is_infinite<> check that detects infinity for some ranges. You could use this to throw for those...

But this seems ranges specific and there is no guarantee that another library with their own infinite range would do this, right?

>(but also maybe we should not use internal stuff of range-v3 anymore)

We most certainly should not!

include/seqan3/range/view/pairwise_combine.hpp
rrahn: The `&&` does exactly what it should. It gives me a glvalue that I pass to view::all. 
Btw. the decltype is identical with the one above. But I believe there is a weird cycle between the constructor definition and the constraint using the view::all adaptor, which might invoke the constructor again or something. I wasted a lot of time on this and the current implementation is used by the C++20 views per draft as well. Note, that neither the ranges views nor our old views implement this and there will be enough time to rereview this later. But for now, I'd really like to get this ... PR done 🙏

include/seqan3/range/view/pairwise_combine.hpp
rrahn: Then wikipedia is colloquial 😱

include/seqan3/range/view/pairwise_combine.hpp
rrahn: yes

include/seqan3/range/view/pairwise_combine.hpp
rrahn: I'll make a card for it.

include/seqan3/range/view/pairwise_combine.hpp
rrahn: > I did some reading and we can at least assume that a bidirectional common iterator is not infinite, because decrementing the end would be undefined. For this reason view::common also downgrades everything to at most ForwardRange.

Well, all natively common iterators would work. The problem that I wasn't aware of really is that many views are not common and that common properties are really just a way of providing access to legacy code but very limited as it mostly downgrades to forward ranges. So by the way of how views work they should in general do not require common ranges. But also note, that this little difference will touch probably only a very few use cases.

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,690 @@
...
+        underlying_iterator_type second_it{};
+        //!\brief The begin of the underlying range.
+        underlying_iterator_type begin_it{};
+        //!\brief The end of the underlying range.
+        underlying_iterator_type end_it{};
+    };
+
+public:
+
+    //!\name Member types
```
h-2: "Associated types"

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,690 @@
...
+    //!\{
+    //!\brief The iterator type.
+    using iterator = iterator_type<underlying_range_type>;
+    //!\brief The const iterator type. Evaluates to void if the underlying range is not const iterable.
+    using const_iterator = transformation_trait_or_t<std::type_identity<iterator_type<underlying_range_type const>>,
+                                                     void>;
+    //!\}
```
h-2: what about the other associated types, like `value_type`, `reference` ...? We provide those for all our views and containers (although not strictly required by range concepts).

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,690 @@
...
+     * linear in the number of elements of the underlying range.
+     *
+     * \attention This view cannot be chained immediately after an infinite range, because upon construction it will
+     *            take forever to reach the last element of the view.
+     *
+     * ### Complexity
+     *
+     * Constant if `underlying_range_type` models std::ranges::BidirectionalRange, otherwise linear.
+     */
+    explicit constexpr pairwise_combine_view(underlying_range_type range) : src_range{std::move(range)}
```
h-2: If you want to keep this explicit, can you make a card for the range module that we need to decide on doing this eerywhere or nowhere?

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,690 @@
...
+     *
+     * Constant.
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    constexpr iterator begin() noexcept
+    {
+        return iterator{std::ranges::begin(src_range), std::ranges::begin(src_range), std::ranges::end(src_range)};
```
h-2: you can omit `iterator` here as you did a few lines down...

include/seqan3/range/view/pairwise_combine.hpp
```diff
@@ -0,0 +1,690 @@
...
+
+} // namespace seqan3::detail
+
+namespace seqan3::view
+{
+/*!\name General purpose views
+ * \{
+ */
+
+/*!\brief             A view adapter that generates all pairwise combinations of the elements of the underlying range.
```
h-2: adapt **o** r
:man_shrugging:

test/unit/range/view/view_pairwise_combine_test.cpp
```diff
@@ -0,0 +1,506 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
+// -----------------------------------------------------------------------------------------------------
+
```
h-2: Can you add a test that verifies all the Range concepts?

include/seqan3/range/view/pairwise_combine.hpp
rrahn: I agree, that we should keep it consistent. And I propose removing them as they can introduce subtle errors and it is not necessary as in most cases you access them via the meta-functions, through the iterator interface.

test/unit/range/view/view_pairwise_combine_test.cpp
rrahn: Which range concepts do you mean?

include/seqan3/range/view/pairwise_combine.hpp
h-2: > I agree, that we should keep it consistent.

Ok, then please add them for now.

> And I propose removing them as they can introduce subtle errors and it is not necessary as in most cases you access them via the meta-functions, through the iterator interface.

Let's discuss it at the next developer meeting. I don't have strong feelings about it, although I have a slight preference for keeping them, especially since our iterators are usually not publicly specified, while the ranges are.

rrahn: I added it already

h-2: I just saw you already did!

test/unit/range/view/view_pairwise_combine_test.cpp
```diff
@@ -0,0 +1,511 @@
...
+    if constexpr (std::ranges::SizedRange<TypeParam>)
+    {
+        EXPECT_TRUE(std::ranges::SizedRange<typename TestFixture::view_t>);
+    }
+
+    if constexpr (std::ranges::RandomAccessRange<TypeParam>)
+    {
+        EXPECT_TRUE(std::ranges::RandomAccessRange<typename TestFixture::view_t>);
+    }
```
h-2: Can you EXPECT_EQ here instead (equivalence)? It's a stronger check, because your's is just implication.

h-2: :nail_care:

PR #774, opened 2019-02-27
[DOC] fix documentation for Alphabet and Semialphabet

h-2:
This should do what you are asking with little overhead.


h-2:
> Can you also add for `char_is_valid_for` a hint that it is actually a function template? I have problems with overloading it for a custom alphabet.

You are right, it doesn't work the way it is currently implemented :( I work on something tonight.

PR #777, opened 2019-03-01
Gap decorator update

smehringer:


include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -98,8 +103,8 @@ class gap_decorator_anchor_set
         //!\brief A pointer to the current anchor gap node. Note that the current tuple value at position 0 is the
         //!       start of the right gap that is still behind the current iterator position.
         typename gap_decorator_anchor_set::set_iterator_type anchor_set_it{};
-        //!\brief Caches the current value pointed at.
```
h-2: It's "caching", not "cashing" :money_mouth:

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -98,8 +103,8 @@ class gap_decorator_anchor_set
...
-        typename gap_decorator_anchor_set::value_type current_val{gap{}};
```
h-2: Hm, is the iterator *on* a gap or *at* a gap? :thinking:

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -206,14 +211,14 @@ class gap_decorator_anchor_set
...
             else
             {   // we arrived at the right gap and have to update the variables. ungapped_view_pos remains unchanged.
                 left_gap_end = anchor_set_it->first + anchor_set_it->second -
                                ((anchor_set_it != host->anchors.begin()) ? (std::prev(anchor_set_it))->second : 0);
                 ++anchor_set_it;
-                current_val = gap{};
+                is_at_gap = true;
             }
 
             return *this;
```
h-2: Why the comment?

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -369,7 +369,8 @@ class gap_decorator_anchor_set
     ~gap_decorator_anchor_set() = default;
 
     //!\brief Construct with the ungapped range type.
-    constexpr gap_decorator_anchor_set(inner_type const & range) : ungapped_view{view::all(range)} {}
+    gap_decorator_anchor_set(inner_type range) : ungapped_view{view::all(std::forward<inner_type>(range))}
+    {}
     // !\}
```
h-2: Is it safe to use one interface here? This function also shouldn't accept container temporaries...

smehringer: this was your code ^^

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
smehringer: I'd say `at`... but I'm not sure

smehringer: `points_to_gap`?

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
smehringer: left over from refactoring. thanks

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
smehringer: If I try to give him a temporary it says `type deduction failed`.. not a very pretty error

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
rrahn: Why not `is_gap`. Is it public API?

rrahn: Is this public API? If not who cares?

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
rrahn: It cannot accept container temporaries otherwise the inner_type would be ill-formed due to the ViewableRange constraint. 🤔

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -783,12 +786,31 @@ class gap_decorator_anchor_set
...
+gap_decorator_anchor_set(urng_t && range) -> gap_decorator_anchor_set<std::remove_reference_t<urng_t> const &>;
+
+//!\brief Views always deduce to their respective type because they are copied.
+template <std::ranges::View urng_t>
+//!\cond
+    requires std::ranges::RandomAccessRange<urng_t> && std::ranges::SizedRange<urng_t>
+//!\endcond
+gap_decorator_anchor_set(urng_t range) -> gap_decorator_anchor_set<urng_t>;
+//!\}
+
```
rrahn: empty line

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -783,12 +786,31 @@ class gap_decorator_anchor_set
...
 
+/*!\name Type deduction guides
+ * \{
+ */
+//!\brief Ranges (non views!) always deduce to `const & range_type` since they are access-only anyway.
+template <std::ranges::ViewableRange urng_t>
+//!\cond
+    requires std::ranges::RandomAccessRange<urng_t> && std::ranges::SizedRange<urng_t>
+//!\endcond
+gap_decorator_anchor_set(urng_t && range) -> gap_decorator_anchor_set<std::remove_reference_t<urng_t> const &>;
```
rrahn: Does this not mean, that a lvalue view will use this deduction guide? And then you don't want to add const to the type don't you.

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -67,10 +66,16 @@ namespace seqan3
...
  * entails updating all subsequent elements in the set to preserve correct cumulative sizes.
  *
+ * ### The seqan3::gap_decorator_anchor_set::iterator type
+ *
+ * \attention The iterator of the seqan3::gap_decorator_anchor_set does not model the std::LegacyInputIterator because
+ * it has no operator->. Note that it does model the std::ranges::InputIterator.
+ *
  */
-template <std::ranges::RandomAccessRange inner_type>
+template <std::ranges::ViewableRange inner_type>
```
rrahn: I would propose constraining this to std::ranges::View and use two constructors. The first takes this type as a copy and moves into the respective data member. The other is a ref_view wrapped non-view range deduced by the deduction guide. Then you need only one. Or was there a specific reason not to do it this way?

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
smehringer: yes and that is the behaviour we want

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -196,7 +196,7 @@ SEQAN3_CONCEPT aligned_sequence_concept =
     Alphabet<value_type_t<t>> &&
     WeaklyAssignable<reference_t<t>, gap const &> &&
     requires { typename detail::unaligned_seq_t<t>; } &&
-    requires (t v, detail::unaligned_seq_t<t> & unaligned)
+    requires (t v, detail::unaligned_seq_t<t> unaligned)
```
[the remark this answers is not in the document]
smehringer: because the ref is in the type now

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -67,10 +66,16 @@ namespace seqan3
  * Resolving random access requires logarithmic access into the set and inserting or removing a gap symbol additionally
  * entails updating all subsequent elements in the set to preserve correct cumulative sizes.
  *
+ * ### The seqan3::gap_decorator_anchor_set::iterator type
+ *
+ * \attention The iterator of the seqan3::gap_decorator_anchor_set does not model the std::LegacyInputIterator because
+ * it has no operator->. Note that it does model the std::ranges::InputIterator.
```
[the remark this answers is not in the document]
smehringer: well, what exactly is a proxy? I thought proxys behave "as if I am working on the actual sequence".. which means I will receive a reference to something I can alter, like for normal iterators. In the gap decorator case, we return a copy of an `gapped<alphabet` character..

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
smehringer: I don't think so because the type requirements from the deduction guide below are `std::View` which is more constrained than `std::ViewableRange` so this should kick in before overload resolution considers && vs copy no?

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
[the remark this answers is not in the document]
rrahn: Well, the proxy behaves like a reference but you return it as a prvalue. The same when you return a copy of gapped<alphabet>. Both of them make the LegacyForwardIterator invalid as it requires a reference to the value type. The operator-> is anyway not used in std::algorithms because they wouldn't know the member variables/functions of the pointed-to object. But they might very well capture the dereferenced result as `auto & ref_elem = *it` which won't work when the iterator returns a prvalue.

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -369,7 +369,12 @@ class gap_decorator_anchor_set
     ~gap_decorator_anchor_set() = default;
 
     //!\brief Construct with the ungapped range type.
-    constexpr gap_decorator_anchor_set(inner_type const & range) : ungapped_view{view::all(range)} {}
+    template <typename other_range_t>
+         requires !std::Same<other_range_t, gap_decorator_anchor_set> &&
+                  std::Same<remove_cvref_t<other_range_t>, remove_cvref_t<inner_type>> &&
+                  std::ranges::ViewableRange<other_range_t> // at end, otherwise it competes with the move ctor
+    gap_decorator_anchor_set(other_range_t && range) : ungapped_view{view::all(std::forward<inner_type>(range))}
+    {} // (@smehringer) why does this not work with other_range_t ---------------------------^
```
rrahn: please mark it as a TODO so we can find it in the code.

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
rrahn: It works as long as the view overload below does not have &&, which it doesn't. So this is all fine.

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -369,7 +369,12 @@ class gap_decorator_anchor_set
     ~gap_decorator_anchor_set() = default;
 
     //!\brief Construct with the ungapped range type.
-    constexpr gap_decorator_anchor_set(inner_type const & range) : ungapped_view{view::all(range)} {}
+    template <typename other_range_t>
+         requires !std::Same<other_range_t, gap_decorator_anchor_set> &&
+                  std::Same<remove_cvref_t<other_range_t>, remove_cvref_t<inner_type>> &&
+                  std::ranges::ViewableRange<other_range_t> // at end, otherwise it competes with the move ctor
+    gap_decorator_anchor_set(other_range_t && range) : ungapped_view{view::all(std::forward<inner_type>(range))}
+    {} // TODO (@smehringer) why does this not work with other_range_t ---------------------------^
```
h-2: because you forgot to add const?

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
h-2: Yes, we removed that explicitly for that reason

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
h-2: Since you are now holding `const &` to all non-views (something I am pretty sure we need to change in the long run 😢 ), can't you just do:
```cpp
gap_decorator_anchor_set(inner_type const & range) : ungapped_view{view::all(range)} {}
gap_decorator_anchor_set(inner_type && range) = delete;
```
?

h-2: Ah, no, this covers initialisation from view and non-view? This whole thing is even more complex than the views 🤦‍♂️ 

I really think we should change it where the inner_type is the template parameter. Then we can model everything similar to views...

Anyway, what you need up there is to add conditional const-ness inside the cast. But only for non-views of course. It will be ugly. It might be easier to make separate constructors for view and non-view. Or just change the design as mentioned above straight away.

smehringer: But everything works like this :D can't we leave it for now?

h-2: I think this overload is always preferred over the other one, also for view-temporaries. Since you cast to inner_type, you will invoke a copy operation of the view at that point. This will break for views that are move-only.
If you cannot come up with a fix now, please clearly document this behaviour and add a card for changing it later on.

Maybe we can sit together some time this week and make a plan for the ranges/view/decorator stuff that needs to be done and in which order.

smehringer: Is there a move only view I can test?

I would document it for now and add a cart because this PR blocks alignment file input and we need this next week. Is that fine?

smehringer: And yes we should make a plan for the views.

h-2: > Is there a move only view I can test?

No, ATM copyability is still required, but this has been changed so some views, e.g. istream_range or our files will be move-only.

> I would document it for now and add a card because this PR blocks alignment file input and we need this next week. Is that fine?

Yep!

smehringer: Done

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
rrahn: Why do you need to comment this?

rrahn:
@smehringer should this be re-reviewed?

smehringer:
Yes please :)

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -67,10 +66,16 @@ namespace seqan3
  * Resolving random access requires logarithmic access into the set and inserting or removing a gap symbol additionally
  * entails updating all subsequent elements in the set to preserve correct cumulative sizes.
  *
+ * ### The seqan3::gap_decorator_anchor_set::iterator type
+ *
+ * \attention The iterator of the seqan3::gap_decorator_anchor_set does not model the std::LegacyInputIterator because
```
rrahn:
```suggestion
 * \attention The iterator of the seqan3::gap_decorator_anchor_set does not model the [Cpp17InputIterator](https://en.cppreference.com/w/cpp/named_req/InputIterator) because
```

rrahn: I had to learn that there is no std::LegacyInputIterator entry in cppreference. So it needs to be linked explicitly and by doing it we can also give it the name as used in the standard.

rrahn: Also I think you still should mention that the return type is not  a reference but a proxy. I think it is more important than a missing operator->.

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -369,7 +369,12 @@ class gap_decorator_anchor_set
     ~gap_decorator_anchor_set() = default;
 
     //!\brief Construct with the ungapped range type.
-    constexpr gap_decorator_anchor_set(inner_type const & range) : ungapped_view{view::all(range)} {}
+    template <typename other_range_t>
+         requires !std::Same<other_range_t, gap_decorator_anchor_set> &&
+                  std::Same<remove_cvref_t<other_range_t>, remove_cvref_t<inner_type>> &&
+                  std::ranges::ViewableRange<other_range_t> // at end, otherwise it competes with the move ctor
+    gap_decorator_anchor_set(other_range_t && range) : ungapped_view{view::all(std::forward<inner_type>(range))}
+    {} // TODO (@smehringer) only works for copyable views. Has to be changes once views are required to be movable.
```
rrahn:
```suggestion
    {} // TODO (@smehringer) only works for copyable views. Has to be changed once views are not required to be copyable anymore.
```

include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -67,10 +66,18 @@ namespace seqan3
  * Resolving random access requires logarithmic access into the set and inserting or removing a gap symbol additionally
  * entails updating all subsequent elements in the set to preserve correct cumulative sizes.
  *
+ * ### The seqan3::gap_decorator_anchor_set::iterator type
+ *
+ * \attention The iterator of the seqan3::gap_decorator_anchor_set does not model the
+ *            [Cpp17LegacyInputIterator](https://en.cppreference.com/w/cpp/named_req/InputIterator) requirements of the
```
rrahn: Have I written Cpp17LegacyInputIterator?
```suggestion
 *            [Cpp17InputIterator](https://en.cppreference.com/w/cpp/named_req/InputIterator) requirements of the
```

rrahn:
once it is green I will merge it

PR #779, opened 2019-03-06
[DOC] Setup

eseiler:
Maybe we should do the directory basically the same way we do the snippets/solutions of the manual?
That is:

- Have one subdirectory per section
- Have CMake fetch all cpp files and prefix them with the section (unique target name)

### DIFF
--- a/doc/setup/index.md
+++ b/doc/setup/index.md
@@ -1,1 +1,149 @@
 # Installing SeqAn3 {#setup}
+
+<b>Learning Objective:</b><br>
+In this short guide you will learn to set up SeqAn3 and how to compile a small example to test whether everything works.
+
+\tutorial_head{Easy, 30 Minutes, No prerequisites, }
+
+[TOC]
+
+<br>
+
+# Software
+Requirements:
+  - gcc >= 7
+  - cmake >= 3.4
+  - git
+
+SeqAn3 requires a compiler with full C++20 support *or* GCC >= 7. Current versions of LLVM/Clang and VisualStudio/MSVC are **not supported**.
+We will briefly explain how to install GCC-7 on some popular operating systems, but we recommend using the latest version of GCC available. For more information refer to your operating system's documentation.
+
+**Ubuntu >= 18.04**
+```
+sudo apt install g++
+```
+**Ubuntu < 18.04**
+```
+sudo add-apt-repository ppa:ubuntu-toolchain-r/test
+sudo apt update
+sudo apt install g++-7
+```
+**MacOS** using [Homebrew](https://brew.sh/)
+```
+brew install gcc@7
+```
+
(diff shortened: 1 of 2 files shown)


doc/setup/index.md
```diff
@@ -1 +1,169 @@
...
+
+```bash
+cmake ../source
+make
+./hello_world
+```
+
+The output should be `Hello world`.
+
+\note Depending on the standard C++ on your system, you may need to specify the compiler via `-DCMAKE_CXX_COMPILER=`, e.g. `-DCMAKE_CXX_COMPILER=g++-7` or the full path to the executable.
```
smehringer:
```suggestion
\note Depending on the standard C++ on your system, you may need to specify the compiler via `-DCMAKE_CXX_COMPILER=`, for example:
```bash
cmake ../source -DCMAKE_CXX_COMPILER=/path/to/executable/g++-7`
\```
```

doc/setup/index.md
```diff
@@ -1 +1,169 @@
...
+
+<!--
+<h4 id='setup--what-happens'>What happens</h4>
```
smehringer:
```suggestion
What happens {#section_what_happens}
```
can be references than: `[Here](#section_what_happens)`

doc/setup/index.md
```diff
@@ -1 +1,169 @@
...
+
+We use cmake as out build software.
```
smehringer:
```suggestion
We use cmake as our build software.
```

doc/setup/index.md
```diff
@@ -1 +1,149 @@
...
+
+**Ubuntu >= 18.04**
+```
+sudo apt-get install gcc
+```
+**Ubuntu < 18.04**
+```
+sudo add-apt-repository ppa:ubuntu-toolchain-r/test
+sudo apt-get update
+sudo apt-get install gcc-7
```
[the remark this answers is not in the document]
h-2: also `apt` is recommended over `apt-get` nowadays, I think.

doc/setup/index.md
```diff
@@ -1 +1,149 @@
...
+
+\note If you encounter the error <code>/usr/lib/x86_64-linux-gnu/libstdc++.so.6: version 'CXXABI_1.3.11' not found</code>, you have to set the LD_LIBRARY_PATH:
+```
+export LD_LIBRARY_PATH=/home/user/miniconda3/envs/gcc7/lib/
+```
+where `/home/user/miniconda3/` is the path to your conda installation.
+
+\attention After installing, `g++ --version` should print the desired version. If not, you may have to use, for example, `g++-7 --version` or even specify the full path to your executable.
+
+Similarly you need to install cmake and git.
```
h-2: Maybe make just one command line that they can copy'n'paste, like:
```bash
apt install g++-7 cmake git
```

PR #780, opened 2019-03-07
[DOC] Tutorial: First Steps with SeqAn3

joergi-w:
This PR adds the *First Steps with SeqAn3* tutorial to the documentation.


doc/tutorial/00_first_example/index.md
```diff
@@ -1,39 +0,0 @@
...
-
-  * "A Tour of C++", Second Edition, Bjarne Stroustrup, 2018
-
-# First head line
-
-`some inline code`
-
-## A second head line
-
-\note This is a note.
```
eseiler: Mhm, I'll make a card to make a wiki post where all the custom commands are explained (after the pmsb).

doc/tutorial/introduction/index.md
```diff
@@ -0,0 +1,119 @@
...
+
+***These tutorials try to briefly introduce C++ features not well known, however they do not teach programming in C++!
+If you know how to program in another language, but are not familiar with C++ and/or the significant
+changes in the language in recent years, we recommend the following resources:***
+
+  * Bjarne Stroustrup: "A Tour of C++", Second Edition, 2018.
+
+[TOC]
+
+<br/>
```
eseiler: Since we build html and not xhtml, we can use `<br>` everywhere

doc/tutorial/introduction/index.md
```diff
@@ -0,0 +1,119 @@
...
+Write a program that creates a std::vector of type `int` and initialise the vector with a few values.
+Then print the vector with seqan3::debug_stream. Does your program also work with std::cerr?
+\endassignment
+\solution
+\snippet introduction_debug_stream.cpp debug
+\endsolution
+
+\note
+You may have spotted that the blue coloured keywords link you directly to the respective API documentation.
+This is helpful, if you need further information on a function, concept or class. We recommend you to open them
```
eseiler:
```suggestion
This is helpful if you need further information on a function, concept or class. We recommend you to open them
```

doc/tutorial/introduction/index.md
```diff
@@ -0,0 +1,119 @@
...
+---
+
+# Hello World!
+
+Most good tutorials start with an easy *Hello World!* program. So have a look:
+
+\snippet introduction_hello_world.cpp hello
+
+Now you may ask, why we do not use std::cout or std::cerr for console output.
+Actually, for the given text it does not make a difference, as seqan3::debug_stream prints to std::cerr as well.
```
eseiler:
```suggestion
Actually, for the given text it does not make a difference since seqan3::debug_stream prints to std::cerr as well.
```

doc/tutorial/introduction/index.md
```diff
@@ -0,0 +1,119 @@
...
+in separate browser tabs such that you can easily switch back to the tutorial.
+
+<br/>
+
+---
+
+# Read sequence files
+
+SeqAn3 provides fast and easy access to biological file formats.
+The following code example demonstrates the interface of seqan3::sequence_file_input.
```
eseiler: Maybe put the first snippet here?

doc/tutorial/introduction/index.md
```diff
@@ -0,0 +1,119 @@
...
+Can you imagine anything easier? After you have initialised the instance with a filename,
+you can simply query the data of interest using the get function. The available fields are
+`SEQ`, `ID` and `QUAL` to retrieve sequences, ids and qualities, respectively. The latter is empty,
```
eseiler:
```suggestion
`SEQ`, `ID` and `QUAL` to retrieve sequences, ids and qualities, respectively. The latter is empty
```

doc/tutorial/introduction/index.md
```diff
@@ -0,0 +1,119 @@
...
+~~~
+
+<br/>
+
+---
+
+# Align two sequences
+
+We have two sequences from the file above now – so let us align them.
+The pairwise sequence alignment is one of the core algorithms in SeqAn, and used by several library components
```
eseiler:
```suggestion
The pairwise sequence alignment is one of the core algorithms in SeqAn and used by several library components
```

doc/tutorial/introduction/index.md
```diff
@@ -0,0 +1,119 @@
...
+and apps. It is strongly optimised for speed and parallel execution, while providing exact results and a
```
eseiler:
```suggestion
and apps. It is strongly optimised for speed and parallel execution while providing exact results and a
```

doc/tutorial/introduction/index.md
```diff
@@ -0,0 +1,119 @@
...
+generic interface.
+
+\snippet introduction_align.cpp alignment_include
+\snippet introduction_align.cpp alignment
+
+The algorithm returns a range of result objects – which is the reason for the loop here (in this case the range
+has length 1). Instead of passing a single pair of sequences, we could give a vector of sequence pairs to the
+algorithm, which executes all alignments in parallel and stores the results in various seqan3::align_result
```
eseiler:
```suggestion
algorithm which then executes all alignments in parallel and stores the results in various seqan3::align_result
```

doc/tutorial/introduction/index.md
```diff
@@ -0,0 +1,119 @@
...
+objects. The second argument to seqan3::align_pairwise is the *configuration*, which allows you to specify
```
eseiler:
```suggestion
objects. The second argument to seqan3::align_pairwise is the *configuration* which allows you to specify
```

doc/tutorial/introduction/index.md
```diff
@@ -0,0 +1,119 @@
...
+a lot of parameters for the alignment computation, for instance score functions, banded alignment and whether
+you wish to compute a traceback or not. The configurations have their own namespace seqan3::align_cfg and can
+be piped for building combinations. Check the alignment tutorial, if you want to learn more.
```
eseiler:
```suggestion
be piped for building combinations. Check the alignment tutorial if you want to learn more.
```

doc/tutorial/introduction/index.md
```diff
@@ -0,0 +1,119 @@
...
+
+\note
+Every directory in the SeqAn sources contains an `all.hpp` file, which includes all the functionality
```
eseiler:
```suggestion
Every directory in the SeqAn sources contains an `all.hpp` file which includes all the functionality
```

doc/tutorial/introduction/index.md
```diff
@@ -0,0 +1,119 @@
...
+include `seqan3/alignment/all.hpp` instead of naming each file individually.
+
+<br/>
+
+---
+
+# Indexed searching
+At the end of this introduction tutorial, we want to present you another efficient algorithm in SeqAn:
+The following code example shows you the steps for performing rapid searches in an indexed genome sequence.
+As index we use the FM index here. After creation we want to query the positions of the motiv *TAG* inside
```
eseiler:
```suggestion
As index we use the FM index here. After creation, we want to query the positions of the motif *TAG* inside
```

doc/tutorial/introduction/index.md
```diff
@@ -0,0 +1,119 @@
...
+a given sequence.
+
+\snippet introduction_align.cpp index_search_include
+\snippet introduction_align.cpp index_search
+
+For more functions on that we refer you to the *Index Tutorial*. We hope, you have got a rough overview on
```
eseiler:
```suggestion
For more information on indexing and searching we refer you to the *Index Tutorial*. We hope you got a rough overview on
```

doc/tutorial/introduction/introduction_align.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+//! [sequence_input_include]
+#include <seqan3/io/sequence_file/input.hpp> // for sequence_file_input
+#include <seqan3/io/stream/debug_stream.hpp> // for debug_stream
+//! [sequence_input_include]
+
+//! [alignment_include]
+#include <tuple>                                                          // for std::make_pair
+#include <seqan3/alignment/pairwise/align_pairwise.hpp>                   // for align_pairwise
+#include <seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp> // for alignment stream operator
```
eseiler: switch the two alignment includes

doc/tutorial/introduction/introduction_align.cpp
```diff
@@ -0,0 +1,74 @@
...
+//! [alignment_include]
+
+//! [index_search_include]
+#include <seqan3/search/algorithm/all.hpp> // for fm_index, search
+//! [index_search_include]
+
+using namespace seqan3;
+
+int main()
```
eseiler: Id' call them `file_out` and `file_in`, but both works.

doc/tutorial/introduction/introduction_align.cpp
```diff
@@ -0,0 +1,74 @@
...
+    // Define a genomic sequence.
+    std::vector<dna4> genome{"ATCGATCGAAGGCTAGCTAGCTAAGGGA"_dna4};
+
+    // Build an FM index for the given sequence.
+    fm_index index{genome};
+
+    // Search the TAG motiv in the index and print the positions, where it is found.
+    debug_stream << search(index, "TAG"_dna4) << '\n';   // => [13,17]
+//! [index_search]
+}
```
eseiler: Since it's all just snippets (so it won't show up in the docs), can you call `std::filesystem::remove(tmp_dir/"seq.fasta")` or similar at the very end?

doc/tutorial/introduction/introduction_align.cpp
```diff
@@ -0,0 +1,74 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
+// -----------------------------------------------------------------------------------------------------
```
eseiler: Usually we do not put the copyright for snippets.

doc/tutorial/introduction/index.md
eseiler: I'll also make a todo in my PR to change this to link to the tutorial

doc/tutorial/00_first_example/index.md
joergi-w: Thank you, that's a great idea!

doc/tutorial/introduction/index.md
joergi-w: Cool! 👍

doc/tutorial/introduction/introduction_align.cpp
joergi-w: Yes, cleaning up is a good idea.

doc/tutorial/introduction/introduction_align.cpp
joergi-w: The snippets and tests belong to our library and must be copyrighted as well.
If we have the copyright already missing in other tutorial's snippets, we should add them... especially to explicitly grant our users the right to copy and use the snippet code.
@smehringer what do you think?

eseiler: No more `\include` :(

But while the tests include the copyright, none the snippets contain a copyright.

doc/tutorial/introduction/index.md
```diff
@@ -0,0 +1,120 @@
...
+The following code example shows you the steps for performing rapid searches in an indexed genome sequence.
+As index we use the FM index here. After creation we want to query the positions of the motif *TAG* inside
+a given sequence.
+
+\snippet introduction_align.cpp index_search_include
+\snippet introduction_align.cpp index_search
+
+For more information on indexing and searching we refer you to the *Index Tutorial*. We hope you got a rough
+overview on how the SeqAn3 interface has developed since version 2 and you are now free to continue with the
+tutorials of your interest.
```
h-2: I think this Tutorial is written well, but I would have preferred something more "aus-einem-Guss", e.g. it is not clear why we do indexing after alignment. Usually it is the other way around. Also the arg-parser is not used... One program as a whole, with annotated steps would be easier, or not?

Also some crucial "meta" things are missing. I think the "first steps" should explain a little bit more about SeqAn, e.g.
  - that SeqAn3 is comprised of Modules, what these modules are and how they relate to includes. The current text is very short....
  - it should teach a little about doxygen, e.g. "If you remember the name of a function or class, but don't know which (sub-)module it belongs to, you can enter it in the search bar (top-right)."...
  - it should teach about the SeqAn style, e.g.
    - "SeqAn3 uses `snake_case` for almost everything, also classes. Only C++ concepts are named using `CamelCase`."
    - "We use a lot of Modern C++ in SeqAn3 so some things might look alien at first, e.g. type templates are used like ordinary types in many situations (no `<>`). We also always use `{}` to initialise objects and not `()` which is only used for function calls. In general the style should be much easier for newcomers."

doc/tutorial/introduction/introduction_align.cpp
h-2: The snippets should not contain this copyright, because it would force anyone copy'n'pasting from the snippets to include our license which is a little too much, I think. 

We should either 
  * prefix every snippet file with
```cpp
// This file is licensed under CreativeCommons - Zero 1.0 Universal / released into the Public Domain.
// You may use this file as wish, including removing this copyright notice.
// This does not affect obligations to include a notice to SeqAn3's copyright when distributing the SeqAn3 library.
```
  * Or not have any note in these files (so they can be `\include`d and instead put above notice into the LICENSE file in our repository root (with "This file is" changed to "Any C++ source code in the `doc` and `test/snippet`-folders is").

h-2: The legal implications of copy'n'pasting from the manual should also be covered in the first article. In general I would put more emphasis on the "meta" side of things.

E.g. "This is a code snippet. You will see many code snippets in our documentation. Most of them are compilable as-is, but some are only valid in their context, e.g. they depend on other code snippets given before/after the current one or other statements implied by the text. You can copy'n'paste freely from these examples, this implies no copyright-obligations (however distributing SeqAn3 or an application using it does, see "Copyright and Citing").

[I just realised we need an article like that]

doc/tutorial/introduction/index.md
```diff
@@ -0,0 +1,120 @@
...
+---
+
+# Read sequence files
+
+SeqAn3 provides fast and easy access to biological file formats.
+The following code example demonstrates the interface of seqan3::sequence_file_input.
+
+\snippet introduction_align.cpp sequence_input_include
+\snippet introduction_align.cpp sequence_input
+
```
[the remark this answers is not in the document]
h-2: seqan3::record s can also be printed as-is with the debug_stream, btw. We could even make more pretty ^^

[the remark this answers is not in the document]
smehringer: I agree. I would chose the structured bindings example because it is rather `pythonic` and might appeal to users.

doc/tutorial/introduction/introduction_align.cpp
smehringer: :+1:

PR #782, opened 2019-03-10
concepts tutorial

h-2:
This is a standalone tutorial for learning the concepts basics. If we merge this we can remove some concepts stuff from the alphabets tutorial so that gets a little easier.

### DIFF
--- a/doc/tutorial/concepts/custom_validator_solution1.cpp
+++ b/doc/tutorial/concepts/custom_validator_solution1.cpp
@@ -1,0 +1,21 @@
+#include <seqan3/argument_parser/all.hpp>
+
+struct custom_validator
+{
+    using value_type = double; // used for all arithmetic types
+
+    void operator() (value_type const &) const
+    {
+        // add implementation later
+    }
+
+    std::string get_help_page_message () const
+    {
+        // add real implementation later
+        return "";
+    }
+};
+
+static_assert(seqan3::validator_concept<custom_validator>); // does not cause compile error
+
+int main() {}
--- a/doc/tutorial/concepts/custom_validator_solution2.cpp
+++ b/doc/tutorial/concepts/custom_validator_solution2.cpp
@@ -1,0 +1,55 @@
+#include <seqan3/argument_parser/all.hpp>
+//![validator]
+#include <cmath>
+
+struct custom_validator
+{
+    using value_type = double; // used for all arithmetic types
+
+    void operator() (value_type const & val) const
+    {
+        if ((std::round(val)                         != val) ||  // not an integer
+            (std::pow(std::round(std::sqrt(val)), 2) != val))    // not a square
+        {
+            throw seqan3::parser_invalid_argument{"The provided number is not an arithmetic square."};
+        }
+    }
+
+    std::string get_help_page_message () const
+    {
+        return "Value must be the square of an integral number.";
+    }
+};
+//![validator]
+
+static_assert(seqan3::validator_concept<custom_validator>);
+
+//![main]
+int main(int argc, char ** argv)
+{
+    seqan3::argument_parser myparser("Test Parser", argc, argv);
+
+    int32_t variable{};
+    int16_t variable2{};
+
+    myparser.add_option(variable, 'i', "", "An int that is a square", seqan3::option_spec::DEFAULT,
+                        custom_validator{}); // ← your validator is used!
+
+    myparser.add_option(variable, 'j', "", "An int that is a square and within [0,20].", seqan3::option_spec::DEFAULT,
+                        custom_validator{} | seqan3::arithmetic_range_validator{0, 20}); // ← now it's chained
+
+    try
+    {
+         myparser.parse(); // trigger command line parsing
+    }
+    catch (seqan3::parser_invalid_argument const & ext)
+    {
+        seqan3::debug_stream << ext.what() << '\n';
+        return -1;
+    }
+
+    seqan3::debug_stream << "Yeah!" << std::endl;
+
+    return 0;
+}
+//![main]
--- a/doc/tutorial/concepts/index.md
+++ b/doc/tutorial/concepts/index.md
@@ -1,0 +1,319 @@
+# C++ Concepts {#tutorial_concepts}
+
+[TOC]
+
+This tutorial introduces "C++ Concepts", a feature of C++20 (and available to some extent in older GCC versions).
+You will learn the terminology used in the context of concepts and how to use SeqAn3's concepts in your application.
+
+\tutorial_head{Moderate, 60 min, \ref setup\, \ref tutorial_argument_parser,}
+
+This tutorial teaches the very basics of working with concepts. For more background and information on how to implement
+your own concepts, we recommend:
+  * A well-readable [paper](http://www.stroustrup.com/good_concepts.pdf) with motivation and historical background.
+  * The (rather dense) [documentation on cppreference](https://en.cppreference.com/w/cpp/language/constraints).
+
+# Constraints
+
+## Motivation
+
+One central design goal of SeqAn is to provide generic algorithms and data structures which can be used for different
+types without reimplementing the same algorithms over and over again for particular types.
+This has multiple benefits: improved maintainability due to an additional level of abstraction
+and more importantly the ability to reuse the code with user provided types.
+A familiar example for generic code is std::vector and the algorithms in the standard library.
+They are *templates* which means that they can be *instantiated* with other types.
+Most often the type cannot be arbitrary, because the template expects a particular interface from the type.
+
+A SeqAn example is the local alignment algorithm.
+It computes the best local match between two sequences over a finite alphabet.
+The algorithm is generic in so far that it allows any alphabet that offers the minimal interface which
+is used inside the algorithm (e.g. objects of the alphabet type must be equality comparable).
+Before C++20, this could not be checked easily and using the interface with non-conforming types would result in
+very hard to read compiler errors and consequently frustration of the user.
+In the following part of the tutorial you will learn how to *constrain* such template arguments of generic functions
+and data structures and how this can have a huge impact on your code.
+
+Here's a shorter example:
+
+```cpp
+template <typename t>
+t add(t const v1, t const v2)
+{
+    return v1 + v2;
+}
+
+int main()
+{
+    return add(1, 3); // instantiates add<int>()
+}
+```
+
+The template parameter `t` is said to be *unconstrained*, in theory it can be instantiated with any type.
+But of course it won't actually compile for all types, because the function template **implicitly requires** that
+types provide a `+` operator.
+If a type is used that does not have a `+` operator, this implicitness causes the compiler to fail at the place where
+such operator is used – and not at the place the template is instantiated.
+This leads to very complex error messages for deeply nested code.
+
+*Constraints* are a way of making requirements of template arguments **explicit**.
+Constraints can be formulated ad-hoc, but this tutorial only covers *concepts*.
+The interested reader can check the [documentation](https://en.cppreference.com/w/cpp/language/constraints) to learn
+about ad-hoc definitions.
+Concepts are a set of constraints with a given name.
+Let's assume there is a concept called `Addable` that requires the existence of a `+` operator (as previously mentioned
+the syntax for defining concepts is not covered here).
+The following snippet demonstrates how we can constrain our function template, i.e. make the template immediately
+reject any types that don't satisfy the requirement:
+
+```cpp
+template <Addable t>
+t add(t const v1, t const v2)
+{
+    return v1 + v2;
+}
+
+int main()
+{
+    return add(1, 3); // instantiates add<int>()
+}
+```
+
+The only difference is that we have replaced `typename` with `Addable`.
+If you plug in a type that does not model `Addable`, you will get a message stating exactly that and not a cryptic
+template backtrace.
+
+The standard library provides a set of [predefined concepts](https://en.cppreference.com/w/cpp/concepts).
+For our example above, the std::Integral concept could have been used.
+
+## Syntax variants
+
+Depending on the complexity of your constraint statements, three different syntaxes are available to enforce
+constraints; all of the following are equivalent.
+
+(1) The "verbose syntax", especially useful when enforcing multiple constraints:
+
+```cpp
+template <typename t1, typename t2>
+    requires std::Integral<t1> && std::Integral<t2> // && MyOtherConcept<t1>
+auto add(t1 const v1, t2 const v2)
+{
+    return v1 + v2;
+}
+```
+
+(2) The "intermediate syntax":
+```cpp
+template <std::Integral t1, std::Integral t2>                   // one constraint per type
+auto add(t1 const v1, t2 const v2)
+{
+    return v1 + v2;
+}
+```
+
+(3) The "terse syntax":
+```cpp
+auto add(std::Integral auto const v1, std::Integral const v2)   // one constraint per type
+{
+    return v1 + v2;
+}
+```
+
+\attention The terse syntax in this form is not yet available in GCC7, GCC8 and GCC9.
+
+Different constraints can be applied to different template parameters and a single template parameter can be constrained
+by multiple concepts.
+Syntaxes can also be combined:
+```cpp
+template <std::Integral t1, std::Integral t2>
+    // requires MyOtherConcept<t1>
+auto add(t1 const v1, t2 const v2)
+{
+    return v1 + v2;
+}
+```
+
+# Terminology
+
+  * Template arguments can be ***constrained***.
+  * A named set of constraints is a ***concept***.
+  * A type that satisfies all requirements of a concept is said to ***model*** said concept.
+  * A *concept* that is composed of another concept and additional constraints is said to ***refine*** said concept(s).
+
+Some people confuse concepts with *interfaces*.
+Both can be used as an abstraction of concrete types, but interfaces have to be inherited from. → the abstraction
+is explicit in the definition of the type.
+Concepts on the other hand "describe properties from the outside". → types don't need to be related and don't need
+to "know about the concept" to model it.
+
+Furthermore, the polymorphism possible with concepts (see below) is faster, because it is resolved at compile-time while
+interface inheritance is resolved at run-time.
+
+# Overloading and specialisation
+
+In generic programming, "function overloading" and "template specialisation" play an important role.
+They allow providing generic interfaces and (gradually) more specialised implementations for specific types or groups
+of types.
+
+## Function (template) overloading
+
+When a function is overloaded and multiple overloads are valid for a given/deduced template argument, the
+*most-refined* overload is chosen:
+
+\include doc/tutorial/concepts/overloading1.cpp
+
+But as soon as we introduce another overload, the compiler will pick the "best" match:
+
+\include doc/tutorial/concepts/overloading2.cpp
+
+\assignment{Exercise: Static polymorphism with alphabets I}
+Write a small program, similar to the one above with the following "skeleton":
+```cpp
+// which includes?
+
+using namespace seqan3;
+
+// Add one or more `void print` function template(s) here //
+
+int main()
+{
+    auto d = 'A'_dna5;
+    auto a = 'L'_aa27;
+    auto g = gap{};
+
+    print(d);
+    print(a);
+    print(g);
+}
+```
+
+The `print` function (template) should print for every object `v` passed to it the result of `to_char(v)` and it should
+be constrained to only accepts types that model seqan3::Alphabet.
+Try calling `print` with a different type, e.g. `int` to make sure that it does.
+\endassignment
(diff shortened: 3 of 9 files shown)


doc/tutorial/concepts/index.md
```diff
@@ -0,0 +1,214 @@
...
+Both can be used as an abstraction of concrete types, but interfaces have to be inherited from. → the abstraction
+is explicit in the definition of the type.
+Concepts on the other hand "describe properties from the outside". → types don't need to be related and don't need
+to "know about the concept" to model it.
+
+Furthermore the polymorphism possible with concepts (see below) is faster, because it is resolved at compile-time while
+interface inheritance is resolved at run-time.
+
+# Overloading and specialisation
+
```
smehringer: Maybe add something along the lines.

`Since constraints limit the set of types a function or class template may be instantiated with, it takes an important part in overloading and specialisation. `

doc/tutorial/concepts/index.md
```diff
@@ -0,0 +1,214 @@
...
+\include doc/tutorial/concepts/overloading2.cpp
+
+\assignment{Exercise: Static polymorphism with alphabets I}
+Write a small program, similar to the one above with the following "skeleton":
+```cpp
+// which includes?
+
+using namespace seqan3;
+
+// ??? void print ????
```
smehringer:
```suggestion
// One or more `void print` function template(s)
```

doc/tutorial/concepts/index.md
```diff
@@ -0,0 +1,214 @@
...
+
+    print(d);
+    print(a);
+    print(g);
+}
+```
+
+The `print` function (template) should print for every object `v` passed to it the result of `to_char(v)` and it should
+be constrained to only accepts types that model seqan3::Alphabet.
+Try with a different type, e.g. `int` to make sure that it does.
```
smehringer:
```suggestion
Try calling `print` with a different type, e.g. `int`, to make sure that it does not work (note: char does model our Alphabet concept!).
```

doc/tutorial/concepts/index.md
```diff
@@ -0,0 +1,214 @@
...
+int main()
+{
+    return add(1, 3); // instantiates add<int>()
+}
+```
+
+The only difference is that we have replaced `typename` with `Addable`. The standard library provides a set of
+[predefined concepts](https://en.cppreference.com/w/cpp/concepts), for our example above, the std::Integral concept
+could have been used.
+
```
smehringer: Maybe add. 
```
You can test if a type models a concept by using `static_assert(Concept<type>)`

snippet example with integral.
```

doc/tutorial/concepts/index.md
```diff
@@ -0,0 +1,214 @@
...
+(3) The "terse syntax":
+```cpp
+auto add(std::Integral auto const v1, std::Integral const v2)
+{
+    return v1 + v2;
+}
+```
+
+\attention The terse syntax in this form is not yet available in GCC7, GCC8 and GCC9.
```
rrahn: Please add a small paragraph stating how we use it in SeqAn.
Maybe also add an example for a the intermediate syntax with a constrained that uses at least two template arguments.

doc/tutorial/concepts/overloading1.cpp
```diff
@@ -0,0 +1,19 @@
+#include <iostream>                 // for std::cout
+
+#include <seqan3/std/concepts>      // GCC7 - GCC9 or
+//#include <concepts>               // compilers will full C++20 support
```
rrahn:
```suggestion
//#include <concepts>               // compilers with full C++20 support
```

doc/tutorial/concepts/overloading2.cpp
```diff
@@ -0,0 +1,25 @@
+#include <iostream>                 // for std::cout
+
+#include <seqan3/std/concepts>      // GCC7 - GCC9 or
+//#include <concepts>               // compilers will full C++20 support
```
rrahn: will->with

doc/tutorial/concepts/index.md
```diff
@@ -0,0 +1,214 @@
...
+```cpp
+template <std::Integral t1, std::Integral t2>
+    // requires MyOtherConcept<t1>
+auto add(t1 const v1, t2 const v2)
+{
+    return v1 + v2;
+}
+```
+
+# Terminology
```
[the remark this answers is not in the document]
h-2: But then they won't know what the terms mean when they read the overview. We should probably find a general style for "terminology section" though, as many tutorials seems to have them.

doc/tutorial/concepts/index.md
```diff
@@ -0,0 +1,214 @@
...
+
+  * Template arguments can be ***constrained***.
```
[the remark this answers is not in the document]
h-2: This is described in the text above. I tried to keep the bullet points as minimal as possible...

doc/tutorial/concepts/index.md
h-2: I have added a sentence!

doc/tutorial/concepts/index.md
h-2: This is explained futher down on the page.

doc/tutorial/concepts/index.md
```diff
@@ -0,0 +1,290 @@
...
+}
+
+int main()
+{
+    return add(1, 3); // instantiates add<int>()
+}
+```
+
+Of course this will not actually compile for all types, because the function template **implicitly requires** that
+types provide a `+` operator.
```
rrahn: I would also add a small paragraph of how this improves readability of compiler errors when constraining the type with a concept.

doc/tutorial/concepts/specialisation.cpp
```diff
@@ -0,0 +1,23 @@
...
+};
+
+template <std::UnsignedIntegral t>
+struct square_root_type<t>
+{
+    using type = float;                     // doesn't need imaginary part
+};
+
+// `int` models std::Integral but not std::UnsignedIntegral:
+static_assert(std::Same<typename square_root_type<int>::type,           std::pair<float, float>>);
```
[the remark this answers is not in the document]
h-2: to make it clearer that there are two arguments to the concept (all concepts mentioned previously had only one argument).

doc/tutorial/concepts/index.md
```diff
@@ -0,0 +1,214 @@
...
+your own concepts, we recommend:
+  * A well-readable [paper](http://www.stroustrup.com/good_concepts.pdf) with motivation and historical background.
+  * The (rather dense) [documentation on cppreference](https://en.cppreference.com/w/cpp/language/constraints).
+
+
+# Constraints
+
+## Introduction
+
+In C++20 a typical pre-C++20 template is said to be *unconstrained*, i.e. it can be instantiated with any type:
```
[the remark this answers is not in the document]
h-2: Very good point. I have added a paragraph.

doc/tutorial/concepts/index.md
h-2: > Please add a small paragraph stating how we use it in SeqAn.

I am not sure if this soo relevant for people that just use our interfaces to write apps (and don't read our code). There is a section further down on how concepts are documented in our docs.

> Maybe also add an example for a the intermediate syntax with a constrained that uses at least two template arguments.

Hm, I'd like to keep this short and really focus the necessary bits to get along and write your app. The syntax you are referring to is cool, but rarely used, especially in apps I think.

doc/tutorial/concepts/index.md
h-2: yep, good idead

doc/tutorial/concepts/index.md
```diff
@@ -0,0 +1,214 @@
...
+
+```cpp
+template <typename t>
```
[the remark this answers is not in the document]
h-2: those two are really small and the second one is not valid code anyway...

h-2:
I have followed almost all your suggestions!

doc/tutorial/concepts/index.md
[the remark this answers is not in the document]
smehringer: Hm IMHO terminology always comes first, even if it explained in more detail further down...

doc/tutorial/concepts/custom_validator_solution1.cpp
```diff
@@ -0,0 +1,21 @@
...
+    using value_type = double; // used for all arithmetic types
+
+    void operator() (value_type const &) const
+    {
+        // add implementation later
+    }
+
+    std::string get_help_page_message () const
+    {
+        // add real implementation later
```
smehringer:
```suggestion
        // add implementation later
```

doc/tutorial/concepts/custom_validator_solution1.cpp
```diff
@@ -0,0 +1,21 @@
...
+        return "";
+    }
+};
+
+static_assert(seqan3::validator_concept<custom_validator>);
```
smehringer:
```suggestion
static_assert(seqan3::validator_concept<custom_validator>); // fine
```

doc/tutorial/concepts/custom_validator_solution2.cpp
```diff
@@ -0,0 +1,55 @@
+#include <seqan3/argument_parser/all.hpp>
+#include <cmath>
```
smehringer: I would include this into the snippet! Especially the cmath include

doc/tutorial/concepts/custom_validator_solution2.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+//![validator]
+struct custom_validator
+{
+    using value_type = double; // used for all arithmetic types
+
+    void operator() (value_type const & val) const
+    {
+        if ((static_cast<double>(static_cast<uint64_t>(val))     != val) ||  // not an integer
```
smehringer: use `std::round` instead of double cast

doc/tutorial/concepts/index.md
```diff
@@ -0,0 +1,214 @@
...
+int main()
+{
+    return add(1, 3); // instantiates add<int>()
+}
+```
+
+Of course this will not actually compile for all types, because the function template **implicitly requires** that
+types provide a `+` operator.
+*Constraints* are a way of making requirements of template arguments **explicit**.
+Constraints can be formulated ad-hoc, but this tutorial only covers *concepts*.
```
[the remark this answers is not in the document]
smehringer: But only the interested reader 
grafik

doc/tutorial/concepts/index.md
```diff
@@ -0,0 +1,214 @@
...
+Concepts are a set of constraints with a given name.
```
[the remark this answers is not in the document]
smehringer: I agree with @h-2

doc/tutorial/concepts/index.md
```diff
@@ -0,0 +1,311 @@
...
+If a type is used that does not have a `+` operator, this implicitness causes the compiler to fail at the place where
+such operator is used – and not at the place the template is instantiated.
+This leads to very complex error messages for deeply nested code.
+
+*Constraints* are a way of making requirements of template arguments **explicit**.
+Constraints can be formulated ad-hoc, but this tutorial only covers *concepts*.
+The interested reader can check the [documentation](https://en.cppreference.com/w/cpp/language/constraints) to learn
+about ad-hoc definitions.
+Concepts are a set of constraints with a given name.
+Let's assume there is a concept called `Addable` that requires the existance of a `+` operator (as previously mentioned
```
smehringer:
```suggestion
Let's assume there is a concept called `Addable` that requires the existence of a `+` operator (as previously mentioned,
```

doc/tutorial/concepts/index.md
```diff
@@ -0,0 +1,311 @@
...
+}
+
+int main()
+{
+    return add(1, 3); // instantiates add<int>()
+}
+```
+
+The only difference is that we have replaced `typename` with `Addable`.
+If you plug in a type that does not model `Addable` you will get a message stating exactly that and not a cryptic
```
smehringer:
```suggestion
If you plug in a type that does not model `Addable`, you will get a message stating exactly that and not a cryptic
```

doc/tutorial/concepts/index.md
```diff
@@ -0,0 +1,311 @@
...
+template trace.
```
smehringer:
```suggestion
template stack trace.
```
?

doc/tutorial/concepts/index.md
```diff
@@ -0,0 +1,311 @@
...
+
+The standard library provides a set of [predefined concepts](https://en.cppreference.com/w/cpp/concepts).
+For our example above, the std::Integral concept could have been used.
+
+## Syntax variants
+
+Depending on the complexity of your constraint statements, three different syntaxes are available to enforce
+constraints; all of the following are equivalent.
```
smehringer:
```suggestion
constraints; all of the following are equivalent:
```

doc/tutorial/concepts/index.md
```diff
@@ -0,0 +1,311 @@
...
+    requires std::Integral<t1> && std::Integral<t2> // && MyOtherConcept<t1>
+auto add(t1 const v1, t2 const v2)
+{
+    return v1 + v2;
+}
+```
+
+(2) The "intermediate syntax":
+```cpp
+template <std::Integral t1, std::Integral t2>
```
smehringer:
```suggestion
template <std::Integral t1, std::Integral t2> // only one contraint per type possible
```

doc/tutorial/concepts/index.md
smehringer: I would say the  "How we use it in seqan3" is maybe rather for developers and contributors than for users of the library

doc/tutorial/concepts/index.md
```diff
@@ -0,0 +1,311 @@
...
+But in fact both are constrained as the detailled documentation reveals.
+
+Now, follow the link to seqan3::validator_concept. We will check in the next section whether you understand the
+documentation for the concept.
+
+# How to make your own type model a concept
+
+## seqan3::validator_concept
+
+Remember the the tutorial on \ref tutorial_argument_parser ? Let's implement our own validator that checks
```
smehringer:
```suggestion
Remember the tutorial on \ref tutorial_argument_parser? Let's implement our own validator that checks
```

doc/tutorial/concepts/index.md
```diff
@@ -0,0 +1,311 @@
...
+specify all public interfaces.
+Unfortunately, doxygen, the system used to generate this documentation, does not handle C++ concepts very well, yet.
+In some parts of the documentation concepts are called "interfaces", please don't let this confuse you.
+And the "verbose syntax" introduced above is not visible at all in the automatically generated documentation.
+That's why it's important to read the detailed documentation section where all requirements are documented.
+
+Have a look at the documentation of seqan3::argument_parser::add_positional_option().
+It has two template parameters, one seems unconstrained (`typename` in the signature) and one is constrained
+(`validator_concept` in the signature).
+But in fact both are constrained as the detailled documentation reveals.
```
smehringer:
```suggestion
But in fact both are constrained as the detailed documentation reveals.
```

doc/tutorial/concepts/custom_validator_solution1.cpp
h-2: but there is an implementation, because there is a return statement

doc/tutorial/concepts/custom_validator_solution1.cpp
h-2: fine is for the fine society? :crown:

doc/tutorial/concepts/index.md
h-2: hm, not sure if it's a stack strictly speaking

doc/tutorial/concepts/custom_validator_solution1.cpp
h-2: i have added a comment

doc/tutorial/concepts/index.md
h-2: but there are colons again. double colon is weird

doc/tutorial/concepts/custom_validator_solution2.cpp
h-2: this should be outdates

doc/tutorial/concepts/index.md
```diff
@@ -0,0 +1,311 @@
...
+This tutorial teaches the very basics of working with concepts. For more background and information on how to implement
+your own concepts, we recommend:
+  * A well-readable [paper](http://www.stroustrup.com/good_concepts.pdf) with motivation and historical background.
+  * The (rather dense) [documentation on cppreference](https://en.cppreference.com/w/cpp/language/constraints).
+
+# Constraints
+
+## Motivation
+
+Generic programming is all about making your algorithms work with different types, e.g. the local alignment algorithm
```
rrahn: One central design goal of SeqAn is to provide generic algorithms and data structures which can be used for different types without reimplementing the same algorithms over and over again for particular types. This has multiple benefits, among which the maintainability will be improved due to an additional level of abstraction and most importantly the code can be reused with user provided types. A familiar example for generic code is the std::vector and the algorithms in the STL. The vector and the algorithms are templatised. This means that they do not depend on a specific type but you could rather put in any type. This sounds awesome, but very often this is not completely true. While the given type can be used arbitrarily it very often must offer a specific interface. To give an example in SeqAn consider the local alignment algorithm. This algorithm computes the best local match between two sequences over a finite alphabet. The algorithm is generic as such that it allows any alphabet type as long as this type offers a minimal interface which is used inside of the algorithm (e.g. objects of the alphabet type must be equality comparable). 
Before C++20, this couldn't be checked easily such that using the interface with non-conformant types would result in very hard to read error messages emitted by the compiler and further would lead to a huge frustration by the user. In the following part of the tutorial you will learn how to *constraint* such template arguments of generic functions and data structures and how this can have a huge benificial impact on your code quality and quantity.

doc/tutorial/concepts/index.md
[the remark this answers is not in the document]
rrahn: 😭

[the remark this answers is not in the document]
rrahn: I still would at least (maybe in a note box) mention that this is only a compile time decision that has no influence on the runtime what so ever and that the concept is just a yes/no decision depending on whether all constraints are satisfied by the queried type.

doc/tutorial/concepts/index.md
rrahn: Gcc calls this template backtrace IIRC. At least they a compiler option called `-ftemplate-backtrace-limit`.

doc/tutorial/concepts/index.md
rrahn: But the terse syntax is not used either, since it is not yet available. And for me it is less the fact that users write this code but need to read the code and the documentation. So for them it is important to understand of how we use it.  But it is true that there is a separate section with regard to doxygen and this would be the better place to provide such information. I'll check this section again

doc/tutorial/concepts/index.md
h-2: Done.

doc/tutorial/concepts/index.md
h-2: Thanks for this text, I have imported it with few changes.

doc/tutorial/concepts/index.md
```diff
@@ -0,0 +1,319 @@
...
+```
+
+\attention The terse syntax in this form is not yet available in GCC7, GCC8 and GCC9.
+
+Different constraints can be applied to different template parameters and a single template parameter can be constrained
+by multiple concepts.
+Syntaxes can also be combined:
+```cpp
+template <std::Integral t1, std::Integral t2>
+    // requires MyOtherConcept<t1>
```
rrahn: why is this commented out? It is not a compiled snippet anyway. 
I would prefer a real example.

rrahn: Actually this is already an example of how we use things within the library. So maybe add a sentence like: `Within SeqAn we always use the intermediate syntax and additionally use the verbose expressions if the intermediate constraint needs to be refined further with more constraints.`

doc/tutorial/concepts/index.md
```diff
@@ -0,0 +1,319 @@
...
+     with `value_type = double` to work on all arithmetic values.
+     \attention In order to be chainable, the validators need to share the same value_type!
+  2. It has to be a [functor](https://stackoverflow.com/questions/356950/what-are-c-functors-and-their-uses), which
+     basically means it must provide `operator()`.
+  3. It has to have a member function `std::string get_help_page_message() const` that returns a string that can be
+     displayed on the help page.
+
+\endhint
+
+
```
rrahn: double new line?

doc/tutorial/concepts/specialisation.cpp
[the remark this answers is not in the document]
rrahn: Mhmm, I still think it is odd and rather not intended, but ok. Let the people that work through it decide.

PR #784, opened 2019-03-11
add -Wno-unused-variable to snippet builds

h-2:


doc/tutorial/introduction/introduction_read_fasta.cpp
```diff
@@ -32,8 +32,6 @@ int main(int argc, char * argv[])
     for (auto & [ seq, id, qual ] : file_in)
```
rrahn:
```suggestion
    for ([[maybe_unused]] auto & [ seq, id, qual ] : file_in)
```
Does this work?

h-2: No. There is no way to use [[maybe_unused]] with structured bindings in C++17.

h-2:
> In general I don't agree on this since we showing code that we would reject otherwise.

I am not sure what you mean by reject. Our cmake stuff does not activate any warnings for user projects.

> If only one variable is used why do you bother to capture as structured binding and not as record data structure?

We used structured bindings for the more complicated examples and also in all of the API docs. I think it is the right way to go, but we can can re-discuss that. Just not now, because the tutorial needs to be done basically in 20min.

h-2:
(everybody that is in the office an working tutorials right now agrees that it makes writing them a lot easier to have this warning deactivated. In some cases it's the only way to get a clearly readable file "en-bloc")

doc/tutorial/introduction/introduction_read_fasta.cpp
rrahn: At least not for gcc-7: https://stackoverflow.com/questions/41404001/structured-binding-with-maybe-unused I am afraid. 
So in this case don't do structured binding.

h-2:
merging this for now, we can re-evaluate it before the retreat.

rrahn:
We not just want to teach the people how to use seqan but also what a good coding style is. And I (working on the tutorials but not currently in the office) am against this premature lifting of rules we originally fixed for the project just for the sake of "en-bloc" files.

doc/tutorial/introduction/introduction_read_fasta.cpp
h-2: sorry, I didn't see this reply when merging.

PR #789, opened 2019-03-11
Adding performance benchmark framework for the gap_decorator.

mariehoffmann:
Currently only called with the `gap_decorator_anchor_set` and some sequence sizes between 2^2 and 2^15.


smehringer:
Can this be also tested with vector<gapped<dna5>>? (of course this is fine in a later PR but just asking for the possibility)
The comparison will be interesting :D

mariehoffmann:
> Can this be also tested with vector<gapped>? (of course this is fine in a later PR but just asking for the possibility)
> The comparison will be interesting :D

Sure, with a slim mock class having the same interface as the gap_decorator.

mariehoffmann:
These are the results using the Google benchmark (built in RELEASE mode). Runtime and actual CPU time are so close, that only the solid line is visible. There is currently only the `gap_decorator_anchor_set` compared with an `std::vector` over the union of gap and dna4. The anchor block version will follow soon.

## Read from left to right
read_left2right

## Read at random position
read_random

## Insert left to right 
insert_left2right

## Insert right to left  
insert_right2left

## Insert at random position
insert_random

## Delete at random position
delete_random

mariehoffmann:
Travis interrupts the performance benchmark after 10 min. Can we re-configure the time to life or should a benchmark finish within 10 min? I could cut at length = `1<<15`.
Details: I am testing currently 2 types of gap decorators with sequence sizes [4, 8, 64, 512, 4096, 32768, 262144] which is already sparse and some advantages (ptr versus gapped vector) show up only for very large sequence lengths. The total execution time of the benchmark on my Mac 2,8 GHz Intel Core i7, 16 GB RAM is less than 15 min. The benchmark contains 6 micro benchmarks.

h-2:
You can also split the benchmark into multiple benchmarks, e.g. one for sequential_read and one for random_access.

h-2:
in general they should still be "micro-"benchmarks, though. Keep in mind that all of us need to run these everytime we change something!

test/performance/range/gap_decorator_benchmark.cpp
```diff
@@ -0,0 +1,460 @@
...
+#include <utility>
+
+#include <benchmark/benchmark.h>
+
+#include <seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp>
+#include <seqan3/alignment/exception.hpp>
+#include <seqan3/alphabet/all.hpp>
+#include <seqan3/range/container/concept.hpp>
+#include <seqan3/range/decorator/all.hpp>
+#include <seqan3/range/detail/random_access_iterator.hpp>
```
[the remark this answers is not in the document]
mariehoffmann: Some, `alphabet/all.hpp` and `decorator/all.hpp`

test/performance/range/gap_decorator_benchmark.cpp
```diff
@@ -0,0 +1,460 @@
...
+ *
+ * Arguments:
+ * gap_vector   reference to empty vector storing the sampled gap lengths
+ * size         final (aligned) sequence size
+ */
+template<typename size_type>
+void sample(std::vector<size_type> * gap_vector, size_type size)
+{
+    std::default_random_engine generator;
+    generator.seed(std::chrono::system_clock::now().time_since_epoch().count());
```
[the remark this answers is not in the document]
mariehoffmann: I'd prefer to keep it random, because of code coverage. There is currently no randomization used inside the unit tests (I know performance is supposed to measure only performance). Apart from this I think the gap sampling here should not produce significant different results between runs (for larger sequence lengths). But it is a point to rethink when the unit tests are done.

test/performance/range/gap_decorator_benchmark.cpp
```diff
@@ -0,0 +1,460 @@
...
+ *
+ * Parameter:
+ * size_type    size type of underlying sequence range
+ *
+ * Arguments:
+ * gap_vector   reference to empty vector storing the sampled gap lengths
+ * size         final (aligned) sequence size
+ */
+template<typename size_type>
+void sample(std::vector<size_type> * gap_vector, size_type size)
```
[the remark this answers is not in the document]
mariehoffmann: No specific reason, replaced now by reference.

test/performance/range/gap_decorator_benchmark.cpp
```diff
@@ -0,0 +1,460 @@
...
+ */
+template<typename size_type, typename iterator_type, typename gap_decorator_t>
+void insert_gaps(std::vector<size_type> & gaps, gap_decorator_t & gap_decorator)
+{
+    [[maybe_unused]] size_type gap_acc = 0;
+    for (size_type i = 0; i < gaps.size(); ++i)
+    {
+        if (gaps[i])
+        {
+            iterator_type it = std::ranges::begin(gap_decorator);
```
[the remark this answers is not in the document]
mariehoffmann: Whenever the type is known I prefer to give or pass it and not to use `auto`, but handing it over as a template parameter is indeed tedious.

test/performance/range/gap_decorator_benchmark.cpp
```diff
@@ -0,0 +1,460 @@
...
+    using sequence_type = remove_cvref_t<detail::unaligned_seq_t<gap_decorator_t>>;
+    sequence_type seq(seq_len, 'A'_dna4);
+
+    // vector of sampled gap lengths for each position
+    std::vector<size_type> gaps(seq_len, 0);
+    sample<size_type>(&gaps, seq_len);
+
+    // determine sum of gaps and non-gap symbols for not exceeding targeted sequence length
+    if constexpr(gapped_flag)
+        resize<size_type, sequence_type>(gaps, seq, seq_len);
```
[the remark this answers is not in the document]
mariehoffmann: Because first the aligned sequence length (some power of 2) is fixed, and then I sample the gap lengths. Hence, I cannot know in advance how long the ungapped sequence has to be.

test/performance/range/gap_decorator_benchmark.cpp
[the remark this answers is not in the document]
smehringer: yes, I meant the random_access_iterator specifically :)So this can be deleted

test/performance/range/gap_decorator_benchmark.cpp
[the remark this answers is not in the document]
smehringer: Do you need the gapped length explicitly? Otherwise, it seems easier to fix the ungapped length and then just insert gaps in the aligned sequence.

test/performance/range/gap_decorator_benchmark.cpp
```diff
@@ -0,0 +1,460 @@
...
+    if constexpr(gapped_flag)
+        insert_gaps<size_type, iterator_type, gap_decorator_t>(gaps, gap_decorator);
+
+    size_t op_ctr = 0;
+    for (auto _ : state)
+    {
+        state.PauseTiming();
+        size_t pos = op_ctr % seq_len;
+        state.ResumeTiming();
+        benchmark::DoNotOptimize(gap_decorator[pos]);
```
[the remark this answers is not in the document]
mariehoffmann: As I understand `DoNotOptimize` forces the compiler to store the result, but it is not applied to intermediate commands. So, the compilation of `gap_decorator.someOp()` is not changed to something like `-O0`, it will be compiled with whatever flags had been given (e.g. `-O3`). The `DoNotOptimize` only forces execution of the right side.

test/performance/range/gap_decorator_benchmark.cpp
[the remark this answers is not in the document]
mariehoffmann: But it is working and is not a performance issue. It is clearer for benchmarking to fix the aligned sequence length compared to fixing only the aligned sequence length and not giving any guarantees how long the actual sequence length will be (due to gap sampling). It makes a huge difference if there are only 1% gaps (relative to unaligned sequence length) inserted or 90%. If I obfuscate this, the interpretation and comparison of the runtime results would become harder.

test/performance/range/gap_decorator_random_benchmark.cpp
```diff
@@ -0,0 +1,283 @@
...
+BENCHMARK_TEMPLATE(insert_random, gap_decorator_anchor_set<const std::vector<dna4> &>, true)->Range(1<<2, 1<<15);
+BENCHMARK_TEMPLATE(insert_random, std::vector<gapped<dna4>>, true)->Range(1<<2, 1<<15);
+
+// ============================================================================
+//  delete at random position
+// ============================================================================
+template <typename gap_decorator_t, bool gapped_flag>
+static void delete_random(benchmark::State& state)
+{
+    unsigned int seq_len = state.range(0);
```
smehringer: The next 13 lines of code are exactly the same as in the other benchmarks, right? Can you also outsource them into a helper function to reduce code duplication?

In addition, (If I understand correctly) you only need to call `sample<size_type>(gaps, seq_len);` if you really want to insert gaps? You could then include this in the `if constexpr` branch which probably reduces (preparation) runtime in the ungapped case

```cpp
    if constexpr(gapped_flag)
    {
        // vector of sampled gap lengths for each position
        std::vector<size_type> gaps(seq_len, 0);
        sample<size_type>(gaps, seq_len);
        resize<size_type, sequence_type>(gaps, seq, seq_len);
    }
```

test/performance/range/gap_decorator_random_benchmark.cpp
```diff
@@ -0,0 +1,283 @@
...
+ * gap_decorator_t  gap decorator type, e.g. gap_decorator_anchor_set
+ *
+ * Arguments:
+ * gaps             reference to gap vector
+ * gap_decorator    reference to gap decorator
+ */
+template<typename size_type, typename gap_decorator_t>
+void insert_gaps(std::vector<size_type> & gaps, gap_decorator_t & gap_decorator)
+{
+    [[maybe_unused]] size_type gap_acc = 0;
```
smehringer: why is this variable maybe unused?

test/performance/range/gap_decorator_random_benchmark.cpp
```diff
@@ -0,0 +1,283 @@
...
+ * Parameters:
+ * size_type        size type of the sequence
+ * gap_decorator_t  gap decorator type, e.g. gap_decorator_anchor_set
+ *
+ * Arguments:
+ * gaps             reference to gap vector
+ * gap_decorator    reference to gap decorator
+ */
+template<typename size_type, typename gap_decorator_t>
+void insert_gaps(std::vector<size_type> & gaps, gap_decorator_t & gap_decorator)
```
smehringer:
```suggestion
void insert_gaps(std::vector<typename gap_decorator_t::size_type> & gaps, gap_decorator_t & gap_decorator)
```
Then you don't need the first template paramter

test/performance/range/gap_decorator_sequential_benchmark.cpp
```diff
@@ -0,0 +1,270 @@
...
+ * size         final (aligned) sequence size
+ */
+template<typename size_type>
+void sample(std::vector<size_type> & gap_vector, size_type size)
+{
+    std::default_random_engine generator;
+    generator.seed(std::chrono::system_clock::now().time_since_epoch().count());
+    std::uniform_real_distribution<double> uni(0.0, 1.0);
+    std::array<double,10> cumsum = {0.6395, 0.8263, 0.8871, 0.9257, 0.9544, 0.9709, 0.9813, 0.9890, 0.9955, 1.0000};
+    for (size_type i = 0; i < size; ++i){
```
smehringer: brace on newline :see_no_evil: :nail_care: sorry

test/performance/range/gap_decorator_sequential_benchmark.cpp
```diff
@@ -0,0 +1,270 @@
...
+void resize(std::vector<size_type> & gaps, sequence_type & seq, unsigned int seq_len)
+{
+    size_type letter_acc = 0;
+    size_type gap_pos = 0;
+    size_type gap_acc = 0;
+
+    while (gap_pos < gaps.size() && gap_acc + letter_acc < seq_len)
+    {
+        if (!gaps[gap_pos])
+            ++letter_acc;
```
smehringer: is else has braces, if shall too :nail_care:

test/performance/range/gap_decorator_sequential_benchmark.cpp
mariehoffmann: damn, how could this have happen :)

test/performance/range/gap_decorator_random_benchmark.cpp
mariehoffmann: Compiler was complaining, but I guess this was when it was WIP. So, I am removing the tag.

test/performance/range/gap_decorator_sequential_benchmark.cpp
mariehoffmann: hmm, na gut.

test/performance/range/gap_decorator_benchmark.cpp
```diff
@@ -0,0 +1,452 @@
...
+    using sequence_type = remove_cvref_t<detail::unaligned_seq_t<gap_decorator_t>>;
+    sequence_type seq(seq_len, 'A'_dna4);
+
+    // vector of sampled gap lengths for each position
+    std::vector<size_type> gaps(seq_len, 0);
+
+    // determine sum of gaps and non-gap symbols for not exceeding targeted sequence length
+    if constexpr(gapped_flag)
+    {
+        sample<size_type>(gaps, seq_len);
```
smehringer:
```suggestion
        sample(gaps, seq_len);
```
I think the size_type is auto-deducted so you do not need it here and also not in the resize call :)

test/performance/range/gap_decorator_benchmark.cpp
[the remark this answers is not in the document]
mariehoffmann: Initially, I designed the benchmark to have a final gap proportion of roughly 40-45 %. For each position gap lengths are sampled according to a distribution that I found in a paper where they looked at  a typical use case and counted the gap lengths. I wrote this when posting the first benchmark results. But I agree, that we should be flexible here.
Comparing few long gaps vs. many small gaps is another issue. That would mean having now two distributions that are being sampled and would open again another parameter dimension in the performance benchmark. 
I doubt that those results give essential insights for the following reason: it is improbable that we ever use a bit vector version (due to bad performance in previous benchmarks), hence, the current approaches store gaps as anchors, so they always manage same sized data sets. Though, they have different theoretical runtimes for random access and those differences show up through simple scaling of the target length - double length means double amount of gaps. 
So, we measure few versus many gaps, but do not vary gap lengths. The latter is something to consider when using approaches storing gaps as single symbols, e.g. with bit vectors.
Varying gap density shifts the crossing points (where A outperforms B) on the x-axis, e.g. the anchor block approach outperforms the anchor set approach already at half the sequence length compared to a run with halvened gap density.
If everyone agrees I would then add the gap density parameter. In this regard we should think about some plotting functionality or csv output, s.t. results can be transformed by e.g. a tabular program to better investigate the performance.

test/performance/range/gap_decorator_benchmark.cpp
```diff
@@ -0,0 +1,452 @@
...
+ * gap_vector   reference to empty vector storing the sampled gap lengths
+ * size         final (aligned) sequence size
+ */
+template<typename size_type>
+void sample(std::vector<size_type> & gap_vector, size_type size)
+{
+    std::default_random_engine generator;
+    generator.seed(std::chrono::system_clock::now().time_since_epoch().count());
+    std::uniform_real_distribution<double> uni(0.0, 1.0);
+    std::array<double,10> cumsum = {0.6395, 0.8263, 0.8871, 0.9257, 0.9544, 0.9709, 0.9813, 0.9890, 0.9955, 1.0000};
```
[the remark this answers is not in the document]
mariehoffmann: The numbers are from the distribution of the initially mentioned paper (see https://www.ncbi.nlm.nih.gov/pmc/articles/PMC419611/). From the publication I derived the discrete cumulative distribution function which is the double vector you see above and I use it sample.

test/performance/range/gap_decorator_benchmark.cpp
```diff
@@ -0,0 +1,452 @@
...
+    for (size_type i = 0; i < size; ++i)
+    {
+        double y = uni(generator);
+        gap_vector[i] = y;
+        auto it = std::find_if(cumsum.begin(), cumsum.end(), [y](double i){return y <= i;});
```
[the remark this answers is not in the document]
mariehoffmann: Line 45 is obsolete. Otherwise, this is how you sample from an arbitrary cumulative distribution function (cdf) with a uniform number generator, i.e. you have the cdf of the probability density function (from the paper), which is basically a step function and you sample now uniformly distributed the y-value in the range [0;1]. You then determine which step is "hit" by the y-value and that's the first one which is greater or equal to your uniformly sampled double.  I rename the lambda-local `i` for improving readability.

test/performance/range/gap_decorator_benchmark.cpp
mariehoffmann: Hmm, didn't work out. Looks like I have to explicitly write it, the compiler is not able to deduce it.

test/performance/range/gap_decorator_benchmark.cpp
```diff
@@ -0,0 +1,452 @@
...
+    size_t op_ctr = 0;
+    for (auto _ : state)
+    {
+        state.PauseTiming();
+        pos = uni_dis(generator);
+        state.ResumeTiming();
+        benchmark::DoNotOptimize(gap_decorator[pos]);
+        state.PauseTiming();
+        ++op_ctr;
+        state.ResumeTiming();
```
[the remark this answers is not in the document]
mariehoffmann: Is there a guarantee that the state loop is not executed more than, e.g., 1,000,000 times (for avoiding an if-clause with reset or modulo computation)?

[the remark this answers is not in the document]
mariehoffmann: Hmm, some benchmarks are performed more than 5 M times, so I cannot avoid the conditional resetting.

test/performance/range/gap_decorator_benchmark.cpp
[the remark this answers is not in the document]
mariehoffmann: One last remark: due to sampling of gap lengths I cannot now in advance the ungapped sequence length if a fixed aligned sequence length is targeted. Even if I now will aim for 5 %, 10 %, ... relative gap amount, it is still sampling involved. Insertion would then stop, when roughly 5 %, 10 %, etc. are reached. So, I recommend to keep the resizing of the underlying sequence, but introduce varying gap densities. Limiting the random behaviour also reduces the expressiveness of the benchmark.

mariehoffmann:
With the varying gap densities (1, 5, 25, 50, 75 %) the benchmark is running for 40 minutes! Splitting sequential and random access functions won't be sufficient, esp. when the `anchor_block` approach will join the suite. Tested sequence lengths are 4, 16, 64, 256, 1024, 4096, 16384, 65536, and 262144 (= 1<<15). Any proposals where to reduce the test cases?

smehringer:
> Tested sequence lengths are 4, 16, 64, 256, 1024, 4096, 16384, 65536, and 262144 (= 1<<15)

I think the benchmarks for the CI may only run with small sequence lengths (maybe until 1024) and 
you could then add a comment or a commented-out-line-of-code, for people that are interested in more robust benchmarks, to run the benchmark on longer sequences?

mariehoffmann:
> > Tested sequence lengths are 4, 16, 64, 256, 1024, 4096, 16384, 65536, and 262144 (= 1<<15)
> 
> I think the benchmarks for the CI may only run with small sequence lengths (maybe until 1024) and
> you could then add a comment or a commented-out-line-of-code, for people that are interested in more robust benchmarks, to run the benchmark on longer sequences?

What do you, @h-2, think? What sequence lengths should be tested on a regular basis? Is 2^10 sufficient?

test/performance/range/gap_decorator_benchmark.cpp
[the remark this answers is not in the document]
h-2: wow, ok, in general I think this is cool from science POV, but maybe a little too much for performance regression tests in the library. In any case, now that it's there we can keep it I guess.

test/performance/range/gap_decorator_benchmark.cpp
```diff
@@ -0,0 +1,452 @@
...
+    {
+        state.PauseTiming();
+        size_t pos = op_ctr % seq_len;
+        auto it = std::ranges::begin(gap_decorator);
+        std::ranges::advance(it, pos);
+        state.ResumeTiming();
+        insert_gap(gap_decorator, it, 1);
+        state.PauseTiming();
+        ++op_ctr;
+        state.ResumeTiming();
```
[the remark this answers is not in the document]
h-2: this is not yet resolved

test/performance/range/gap_decorator_benchmark.cpp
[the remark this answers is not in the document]
h-2: You can do a loop inside the loop, e.g. perform 100K access per iteration of the outer loop and then always reset the counter/iterator to beginning.

h-2:
> With the varying gap densities (1, 5, 25, 50, 75 %) the benchmark is running for 40 minutes! Splitting sequential and random access functions won't be sufficient, esp. when the anchor_block approach will join the suite. Tested sequence lengths are 4, 16, 64, 256, 1024, 4096, 16384, 65536, and 262144 (= 1<<15). Any proposals where to reduce the test cases?

* I would do a check for `SEQAN3_LONG_TESTS` and if that is defined, perform the full set of tests, otherwise only a subset. This macro could be added to the nightlies or just activated on-deman.
* I would choose the subset as sequence lengths: 64, 256, 4096 and gap densities: 1, 5, 50; If this still takes to long, we can change these further.
* To further reduce the run-time, I would also generate the test sequence/gap selection once for each parameter combination (but independent of the decorator/vector). This also improves comparability.

test/performance/range/gap_decorator_benchmark.cpp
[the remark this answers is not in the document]
h-2: This also mitigates the effect the check/reset.

test/performance/range/gap_decorator_rand_read_benchmark.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+// Apply benchmarks with custom ranges for grid parameters sequence length and gap proportions
+static void CustomArguments(benchmark::internal::Benchmark* b) {
+    std::array<long long int, 5> gap_percentages = {1, 5, 25, 50, 75};
+    for (long long int seq_len = 4; seq_len <= (1 << 18); seq_len <<= 2)
+    {
+        for (auto gap_percentage : gap_percentages)
+            b->Args({seq_len, gap_percentage});
+    }
+}
```
h-2: please make this conform to our style-guide ( snake_case, no static, { to newline, fixed width integers...)

test/performance/range/gap_decorator_rand_read_benchmark.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+// ============================================================================
+//  read at random position
+// ============================================================================
+/* Parameters:
+ * gap_decorator_t      gap decorator class, e.g. gap_decorator_anchor_set
+ * gapped_flag          operate on already gapped (true) or ungapped sequence (false)
+ */
+template <typename gap_decorator_t, bool gapped_flag>
+static void read_random(benchmark::State& state)
```
h-2:
```suggestion
void read_random(benchmark::State & state)
```

test/performance/range/gap_decorator_rand_read_benchmark.cpp
```diff
@@ -0,0 +1,93 @@
...
+        insert_gaps<gap_decorator_t>(gaps, gap_decorator, seq_len);
+
+    std::mt19937 generator(time(0)); //Standard mersenne_twister_engine seeded with current time
+    std::uniform_real_distribution<> uni_dis{0.0, static_cast<double>(seq_len)};
+
+    // sample read positions in advance
+    std::vector<size_t> access_positions;
+    access_positions.resize(1 << 18);
+    for (size_t i = 0; i < access_positions.size(); ++i)
+        access_positions[i] = uni_dis(generator);
```
h-2: i would generate fewer positions and loop over them more often...

test/performance/range/gap_decorator_rand_write_benchmark.cpp
```diff
@@ -0,0 +1,150 @@
...
+    std::vector<size_t> access_positions;
+    access_positions.resize(1 << 18);
+    for (size_t i = 0; i < access_positions.size(); ++i)
+        access_positions[i] = uni_dis(generator);
+    size_t j = 0;
+    for (auto _ : state)
+    {
+        state.PauseTiming();
+        auto it = std::ranges::begin(gap_decorator);
+        std::ranges::advance(it, access_positions[j++]);
```
h-2: see above; do not advance every time, instead just increment the iterator every iteration which is constant time. also remove pausing of timer

test/performance/range/gap_decorator_benchmark.cpp
```diff
@@ -0,0 +1,452 @@
...
+    size_t op_ctr = 0;
+    for (auto _ : state)
+    {
+        state.PauseTiming();
+        size_t pos = op_ctr % seq_len;
+        state.ResumeTiming();
+        benchmark::DoNotOptimize(gap_decorator[pos]);
+        state.PauseTiming();
+        ++op_ctr;
+        state.ResumeTiming();
```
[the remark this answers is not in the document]
mariehoffmann: This guard has been removed.

test/performance/range/gap_decorator_benchmark.cpp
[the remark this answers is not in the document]
mariehoffmann: Yes, it is. The above file is outdated. The benchmark is split into 4 separate files. There is no pausing anymore for operations like counter increment.

test/performance/range/gap_decorator_benchmark.cpp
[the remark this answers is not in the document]
mariehoffmann: Done.

test/performance/range/gap_decorator_benchmark.cpp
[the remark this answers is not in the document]
h-2: there is still pausing for advance, see my comments below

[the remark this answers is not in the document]
mariehoffmann: In which file? The above one doesn't exist anymore.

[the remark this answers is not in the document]
h-2: See the last comment on this page. The file is test/performance/range/gap_decorator_rand_write_benchmark.cpp

test/performance/range/gap_decorator_rand_read_benchmark.cpp
mariehoffmann: > > With the varying gap densities (1, 5, 25, 50, 75 %) the benchmark is running for 40 minutes! Splitting sequential and random access functions won't be sufficient, esp. when the anchor_block approach will join the suite. Tested sequence lengths are 4, 16, 64, 256, 1024, 4096, 16384, 65536, and 262144 (= 1<<15). Any proposals where to reduce the test cases?
> 
>     * I would do a check for `SEQAN3_LONG_TESTS` and if that is defined, perform the full set of tests, otherwise only a subset. This macro could be added to the nightlies or just activated on-deman.
> 
>     * I would choose the subset as sequence lengths: 64, 256, 4096 and gap densities: 1, 5, 50; If this still takes to long, we can change these further.
> 
>     * To further reduce the run-time, I would also generate the test sequence/gap selection once for each parameter combination (but independent of the decorator/vector). This also improves comparability.

The first two points I implemented. The last one is tricky. You would typically use a fixture then, but the gap percentages and lengths are encoded in the state which is not available for the fixture class! Also, from my POV (or better state of knowledge) using a templated, fixture does not allow to have the test function externally defined (how it is currently). This allowed me to call it with various `gap_decorator versions. I could make it only running with defining the function body directly after the `BENCHMARK_TEMPLATE_DEFINE_F` macro like shown in google benchmark examples. However, this solution is inconvenient when testing with various types. Let's discuss this when you are back.

test/performance/range/gap_decorator_rand_write_benchmark.cpp
mariehoffmann: I don't get it :/. 
> do not advance every time, instead just increment the iterator every iteration

"every time" = "every iteration" or what am I missing here?

mariehoffmann: Not sure, if I understand this remark, because this is a random write benchmark and not a sequential write one (e.g. insert from left to right), that's why for each measured operation the iterator is moved to a random position.

test/performance/range/gap_decorator_rand_read_benchmark.cpp
mariehoffmann: Basically, I switched back to the previous version, where the data structures are initialized around the state loop. Template + CustomArguments + Test Class is not possible, because the definition needs to be inline (see answer here: https://stackoverflow.com/questions/56006025/how-to-use-a-templated-google-benchmark-fixture-with-custom-arguments-correctly/56010875?noredirect=1#comment98736674_56010875).

test/performance/range/gap_decorator_helper.hpp
```diff
@@ -0,0 +1,162 @@
...
+#include <seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp>
+#include <seqan3/alignment/exception.hpp>
+#include <seqan3/alphabet/all.hpp>
+#include <seqan3/range/decorator/all.hpp>
+#include <seqan3/std/ranges>
+
+using namespace seqan3;
+
+#define SEQ_LEN_LONG 1<<18
+#define SEQ_LEN_SHORT 1<<12
```
h-2:
```suggestion
#define SEQAN3_LEN_SHORT 1<<12
```

All macros should begin with "SEQAN3"

test/performance/range/gap_decorator_helper.hpp
```diff
@@ -0,0 +1,162 @@
...
+ * Apply benchmarks with custom ranges for grid parameters sequence length and gap
+ * proportions. When setting the SEQAN3_LONG_TESTS macro, the longer benchmark
+ * version will be executed.
+ *
+ *  | flag                | sequence lengths | gap proportions
+ *  | --------------------|------------------|---------------------
+ *  | !SEQAN3_LONG_TESTS  |  1 << [4:2:12]   | [1, 5, 50] %
+ *  | SEQAN3_LONG_TESTS   |  1 << [4:2:18]   | [1, 5, 25, 50, 75] %
+ */
+static void CustomArguments(benchmark::internal::Benchmark* b) {
```
h-2:
```suggestion
void custom_arguments(benchmark::internal::Benchmark * b)
{
```

no static, camel_case and style guide

test/performance/range/gap_decorator_helper.hpp
```diff
@@ -0,0 +1,162 @@
...
+    std::vector<long long int> gap_percentages{1, 5, 50};
+    long long int seq_len_max = SEQ_LEN_SHORT;
+
+    #ifdef SEQAN3_LONG_TESTS
+        seq_len_max = SEQ_LEN_LONG;
+        gap_percentages = {1, 5, 25, 50, 75};
+    #endif
+
+    for (long long int seq_len = 16; seq_len <= seq_len_max; seq_len <<= 2)
```
h-2:
```suggestion
    for (size_t seq_len = 16; seq_len <= seq_len_max; seq_len <<= 2)
```

size_t instead of `long long` also in the other cases

test/performance/range/gap_decorator_helper.hpp
```diff
@@ -0,0 +1,162 @@
...
+    {
+        for (auto gap_percentage : gap_percentages)
+            b->Args(
+                {
+                    seq_len, gap_percentage
+                }
+            );
```
h-2: the above can go in one line I think

h-2: If not, wrap another `{ }` around it

test/performance/range/gap_decorator_rand_read_benchmark.cpp
```diff
@@ -0,0 +1,83 @@
...
+    #endif
+
+    for (size_t i = 0; i < access_positions.size(); ++i)
+        access_positions[i] = uni_dis(generator);
+    size_t j = 0;
+    for (auto _ : state)
+    {
+        benchmark::DoNotOptimize(gap_decorator[access_positions[j++]]);
+        if (j == access_positions.size())
+            j = 0;
```
h-2: I expect the if check to be a lot more expensive then the look-up. Can you ensure that j is divisable by 10 and perform 10 random access lookups ever loop iteration?

test/performance/range/gap_decorator_rand_write_benchmark.cpp
h-2: Well, the problem is that advance always goes linearly from the beginning to the positions. This is very likely a lot slower than the insertion. We need to wait for the changes by @smehringer so that you can use the random_access iterator so you can jump in log-time instead of linear time at least.

test/performance/range/gap_decorator_rand_write_benchmark.cpp
```diff
@@ -0,0 +1,141 @@
...
+BENCHMARK_TEMPLATE(insert_random, std::vector<gapped<dna4>>, false)->Apply(CustomArguments);
+// Insert gaps of length 1 at random position into GAPPED sequence
+BENCHMARK_TEMPLATE(insert_random, gap_decorator_anchor_set<const std::vector<dna4> &>, true)->Apply(CustomArguments);
+BENCHMARK_TEMPLATE(insert_random, std::vector<gapped<dna4>>, true)->Apply(CustomArguments);
+
+// ============================================================================
+//  delete at random position
+// ============================================================================
+template <typename gap_decorator_t, bool gapped_flag>
+void delete_random(benchmark::State& state)
```
h-2:
```suggestion
void delete_random(benchmark::State & state)
```

also in other places

test/performance/range/gap_decorator_seq_read_benchmark.cpp
```diff
@@ -0,0 +1,71 @@
...
+    // insert gaps before starting benchmark
+    if constexpr(gapped_flag)
+        insert_gaps<gap_decorator_t>(gaps, gap_decorator, seq_len);
+
+    size_t op_ctr = 0;
+
+    for (auto _ : state)
+    {
+        size_t pos = op_ctr % seq_len;
+        benchmark::DoNotOptimize(gap_decorator[pos]);
```
h-2: calling the random access operator always invokes log behaviour. Instead use iterators that you increment and dereference.

test/performance/range/gap_decorator_seq_write_benchmark.cpp
```diff
@@ -0,0 +1,119 @@
...
+    if (gapped_flag)
+        insert_gaps<gap_decorator_t>(gaps, gap_decorator, seq_len);
+
+    auto it = std::ranges::begin(gap_decorator);
+    for (auto _ : state)
+    {
+        if (it == std::ranges::end(gap_decorator))
+            it = std::ranges::begin(gap_decorator);
+        it = insert_gap(gap_decorator, it, 1);
+        ++it;
```
h-2: maybe skip one more / increment twice?

test/performance/range/gap_decorator_seq_write_benchmark.cpp
```diff
@@ -0,0 +1,119 @@
...
+    // end() != begin() due to assert above on sequence length
+    auto it = std::ranges::end(gap_decorator);
+    for (auto _ : state)
+    {
+        it = insert_gap(gap_decorator, it, 1);
+        state.PauseTiming();
+        if (it == std::ranges::begin(gap_decorator))
+            it = std::ranges::end(gap_decorator);
+        --it;
+        state.ResumeTiming();
```
h-2: pause and resume likely more expensive that the actual operations...

test/performance/range/gap_decorator_rand_write_benchmark.cpp
h-2: alternatively you can generate X number of iterators before the loop, instead of X number of access_positions. That's definitely better for the benchmark.

test/performance/range/gap_decorator_rand_write_benchmark.cpp
```diff
@@ -0,0 +1,141 @@
...
+        auto first = std::ranges::begin(gap_decorator);
+        std::ranges::advance(first, access_positions[j++]);
+        first = insert_gap(gap_decorator, first, 2);
+        auto last = std::ranges::begin(gap_decorator);
+        last = first;
+        std::ranges::advance(last, 2);
+        if (j == access_positions.size())
+            j = 0;
+        state.ResumeTiming();
+        erase_gap(gap_decorator, first, last);
```
h-2: see above

test/performance/range/gap_decorator_rand_write_benchmark.cpp
mariehoffmann: Or just pause the timer, but this would still consume more CPU cycles then having iterators prepared. My new commit will implement your last proposal. We are getting close to the most environmental friendly solution ;)

mariehoffmann: Wait, I cannot prepare iterators for the same gap_decorator instance. Because after a couple of insertions, the decorator might be moved in memory and hereby invalidating the pre-computed iterators (=> segfault).

test/performance/range/gap_decorator_helper.hpp
mariehoffmann: In a previous comment, you insisted that curly brackets go to newlines.

test/performance/range/gap_decorator_rand_read_benchmark.cpp
mariehoffmann: I made this benchmark consistent with the other one in the same file, so I use `j %= 1 << 10` instead of if-clause.

mariehoffmann: Do you mean 10 random access ops per state iteration, right?

test/performance/range/gap_decorator_helper.hpp
mariehoffmann: `benchmark->Args` prescribes `long long int` and `size_t` (`long unsigned int`) to `long long int` conversion throws a `narrowing conversion` error.

test/performance/range/gap_decorator_rand_write_benchmark.cpp
h-2: Yes, your are right. For the write benchmarks, you cannot use this approach. You should still use the random access iterator. @smehringer What is the state of #1139 ?

h-2: And yes, if it cannot be avoided, stop the timer.

test/performance/range/gap_decorator_rand_write_benchmark.cpp
mariehoffmann: This is true for 17 files!

test/performance/range/gap_decorator_helper.hpp
h-2: Except where they can go in one 😉 
https://github.com/seqan/seqan3/wiki/General-formatting#brace

test/performance/range/gap_decorator_rand_read_benchmark.cpp
h-2: > Do you mean 10 random access ops per state iteration, right?

Yes!

test/performance/range/gap_decorator_rand_write_benchmark.cpp
h-2: Well then let's not add an 18th!

test/performance/range/gap_decorator_seq_read_benchmark.cpp
h-2: sequential read can definitely be done simply by incrementing the iterator!

mariehoffmann: Then we should clarify what is supposed to be benchmarked here. Do we do code optimization or do we benchmark the []-operator or the iterator over the aligned sequence?

test/performance/range/gap_decorator_seq_write_benchmark.cpp
mariehoffmann: Yes, for not having one contiguous gap.

mariehoffmann: But this also adds another condition in the if-clause (for which the timer is not paused) in order to avoid the iterator to jump over the end.

mariehoffmann: But this also adds another condition into the if-clause for avoiding that the iterator jumps over the end and will therefore affect the timer.

test/performance/range/gap_decorator_seq_read_benchmark.cpp
h-2: Hm, I am not sure what you mean by code optimisation. Reading from left to right usually means taking an iterator, reading from it, and incrementing it. That's what happens when you do a `for (auto a : v)` loop for example.
Why would you want to use the random-access-operator when it's not about random access? Especially in this case it will have a dramatic run-time impact!

test/performance/range/gap_decorator_seq_write_benchmark.cpp
h-2: Well, right now the check is not needed, is it? Because you would never reach the end...

test/performance/range/gap_decorator_seq_write_benchmark.cpp
mariehoffmann: But we want this to be independent from the iterator implementation. E.g. in iterator over an anchor set-based gap decorator depends on the increment/decrement runtimes of an `std::ordered_set`, which is in the worst logarithmic in the number of contiguous gaps, whereas the anchor block-based approach is expected to be constant. I therefore, would pause the timer here.

test/performance/range/gap_decorator_rand_write_benchmark.cpp
mariehoffmann: Since we modify the gap_decorator, addresses become invalidated (see comment above), so we cannot prepare iterators.

h-2: Yes, that's correct

test/performance/range/gap_decorator_seq_write_benchmark.cpp
h-2: We have decorators where `++` is not in O(1) ? 😱 

Also here: if you insert in every iteration and `insert` inserts before the given iterator and returns the newly inserted value... how is it ever possible to reach the end?

test/performance/range/gap_decorator_rand_read_benchmark.cpp
mariehoffmann: I implemented this change, but it obfuscates a bit what is tested in the benchmark and makes  less comparable with other benchmarks.

test/performance/range/gap_decorator_rand_write_benchmark.cpp
mariehoffmann: The read benchmarks are currently testing the []-operator, so we are measuring the time of its iterator's private `jump` method. Pre-computing the iterators, means we measure `operator*` of the iterator, which should be constant for all approaches.

test/performance/range/gap_decorator_seq_read_benchmark.cpp
mariehoffmann: Got you. So, I would rather call `*it` surrounded by the `DoNotOptimize` tag for the benchmark, right?

test/performance/range/gap_decorator_seq_write_benchmark.cpp
mariehoffmann: ++ on iterator: Maybe I miss here something, but the `std::ordered_set` storing the gaps is a red-black tree. In case the iterator has to jump from one consecutive gap to another one, it has to walk the branch up and down and the depth of the tree is in log(k).
reaching the end: I start from back to front.

mariehoffmann: However, amortized it is constant. But since the red-black tree has this worst case and another approach may not (like anchor block), I thought it's worth to exclude the iterator decrement.

test/performance/range/gap_decorator_seq_write_benchmark.cpp
mariehoffmann: With `++++it` I will need the check, except I insert 2 gaps. Should I?

mariehoffmann: I'll do it.

test/performance/range/gap_decorator_rand_write_benchmark.cpp
mariehoffmann: Hihi, posed a PR for this correction.

h-2:
Arg. Can't merge because there are CI failures. Can you fix those?

mariehoffmann:
Hmm, runs through on my mac. Jenkins errors are due to this outdated license or/and upload script error occurring everywhere. The other one is about the just happened renaming of `gap_decorator_anchor_set` to `gap_decorator`. When re-designing the gap decorator interface, it should always be considered that there will be multiple gap_decorator implementation. I guess you will them by a tag or something(?). Fixed with latest commit.

PR #795, opened 2019-03-12
reduce compile time on travis for header tests

marehr:
This is achieved by only testing seqan3 headers by default.

One can enable header tests of external libraries by calling

```
cmake -DSEQAN3_FULL_HEADER_TEST .../seqan3/test/header
```

Which could be used in nightlies.

### DIFF
--- a/test/header/CMakeLists.txt
+++ b/test/header/CMakeLists.txt
@@ -10,6 +10,8 @@ project (seqan3_header_test CXX)
 
 include (../seqan3-test.cmake)
 
+option (SEQAN3_FULL_HEADER_TEST "Test seqan3 headers as well as the headers of external libraries" OFF)
+
 # We compile each header twice in separate compilation units. Each alone is
 # sufficient to test that the header is functional, but both are needed to check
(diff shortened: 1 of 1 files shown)


h-2:
I wouldn't even mind removing those tests altogether, but I have no strong feelings about it.

test/header/CMakeLists.txt
```diff
@@ -95,6 +97,8 @@ seqan3_require_test ()
 seqan3_header_test (seqan3 "${SEQAN3_CLONE_DIR}/include")
 seqan3_header_test (seqan3_test "${SEQAN3_CLONE_DIR}/test/include")
 
+if (SEQAN3_FULL_HEADER_TEST)
+
```
rrahn: should the part below the if not be indented?

rrahn:
And can you please amend your commit message. `[todo]` should be `[INFRA]`

h-2:
@marehr only minor stuff, can you have a look so we can merge it? Thanks!

marehr:
@h-2 done

PR #796, opened 2019-03-12
[DOC] Update the documentation for aligned_ends.

rrahn:
Updates the doc for aligned ends configuration.

### DIFF
--- a/include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
+++ b/include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
@@ -72,19 +72,18 @@ public:
 // first_seq_leading
 // ----------------------------------------------------------------------------
 
-/*!\brief The penalty configuration for a leading gap in the first sequence of a pairwise sequence alignment.
+/*!\brief The penalty configuration for aligning the front of the first sequence with a gap.
  * \ingroup configuration
  * \tparam value_t The type of the value to be wrapped. Can be of type std::true_type, std::false_type or bool.
  *
  * \details
  *
- * This configuration element enables/disables penalties for the respective gaps in the pairwise
- * sequence alignment. If the wrapped value evaluates to true, leading gaps in the first sequence will be not penalized.
+ * This strong type enables (`false`) or disables (`true`) penalties for aligning the respective sequence end with gaps.
  * If one constructs this element with a std::integral_constant it will convert to a static type such that
- * compile time optimizations can be enabled. If the type is constructed from a bool it will convert to a dynamic type
- * but will be later during the instantiation of the pairwise alignment converted to a static type.
+ * compile time optimisations can be used. If the type is constructed from a bool it will convert to a dynamic type
+ * but will be converted to a static type during the configuration of the pairwise alignment algorithm.
  * Using a `bool` allows to dynamically set the value if the option is only known at runtime. If the option is
- * already known at compile time the static version will be a better option.
+ * already known at compile time the static version will be the preferred option.
  *
  * \see first_seq_trailing
  * \see second_seq_leading
@@ -111,7 +110,7 @@ first_seq_leading(value_t) -> first_seq_leading<value_t>;
 // first_seq_trailing
 // ----------------------------------------------------------------------------
 
-/*!\brief The penalty configuration for a trailing gap in the first sequence of a pairwise sequence alignment.
+/*!\brief The penalty configuration for a aligning the back of the first sequence with a gap.
  * \ingroup configuration
  * \tparam value_t The type of the value to be wrapped. Can be of type std::true_type, std::false_type or bool.
  *
@@ -142,7 +141,7 @@ first_seq_trailing(value_t) -> first_seq_trailing<value_t>;
 // second_seq_leading
 // ----------------------------------------------------------------------------
 
-/*!\brief The penalty configuration for a leading gap in the second sequence of a pairwise sequence alignment.
+/*!\brief The penalty configuration for a aligning the front of the second sequence with a gap.
  * \ingroup configuration
  * \tparam value_t The type of the value to be wrapped. Can be of type std::true_type, std::false_type or bool.
  *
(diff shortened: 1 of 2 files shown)


include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
```diff
@@ -211,11 +211,35 @@ second_seq_trailing(value_t) -> second_seq_trailing<value_t>;
  * \details
  *
  * A wrapper for providing ordered access to the end-gap specifiers independent of the input order.
- * The possible input types can be: seqan3::first_seq_leading, seqan3::first_seq_trailing, seqan3::second_seq_leading and
- * seqan3::second_seq_trailing.
+ * The possible input types can be: seqan3::first_seq_leading, seqan3::first_seq_trailing, seqan3::second_seq_leading
```
eseiler: These need to be `seqan3::align_cfg::first_seq_leading` and so on

include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
```diff
@@ -211,11 +211,35 @@ second_seq_trailing(value_t) -> second_seq_trailing<value_t>;
...
+ * If a specifier is not set it will default to `false` and thus the respective end-gap will be penalised in the
  * pairwise alignment.
+ *
+ * ### Static vs runtime configuration
+ *
+ * The end_gaps class preserves the static/non-static property of the respective end-gap specifier. Those specifiers
+ * can, depending on how they are constructed, contain a static information or a runtime information whether or not a
+ * specific end-gap is enabled. To check whether the information was static the function end_gaps::is_static can be
+ * used. If it was static the function end_gaps::get_static can be used to obtain the respective value at compile time.
+ * Note, that both functions are static and thus need to be called on the type and not on an instance of the type.
```
eseiler:
```suggestion
 * Note that both functions are static and thus need to be called on the type and not on an instance of the type.
```

include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
```diff
@@ -211,11 +211,35 @@ second_seq_trailing(value_t) -> second_seq_trailing<value_t>;
...
+ *
+ * \snippet snippet/alignment/configuration/aligned_ends_example.cpp access
+ *
+ * To get the respective value at runtime use `[]`-operator. This function returns always the respective value
```
eseiler:
```suggestion
 * To get the respective value at runtime use the \ref operator[]() "[]-operator". This function returns always the respective value
```

include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
```diff
@@ -211,11 +211,35 @@ second_seq_trailing(value_t) -> second_seq_trailing<value_t>;
...
+ * independent of whether the value was provided by a static variable or a runtime variable.
+ * Also note that static and non-static end-gap specifier can be mixed within the seqan3::end_gaps class template.
```
eseiler:
```suggestion
 * Also note that static and non-static end-gap specifier can be mixed within the end_gaps class template.
```

In this file you can either link with `end_gaps` (because it's in the file) or `seqan3::align_cfg::end_gaps` (the full identifier).

include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
```diff
@@ -390,7 +414,25 @@ end_gaps(ends_t && ...) -> end_gaps<std::remove_reference_t<ends_t>...>;
...
+ *
+ * \details
+ *
+ * This configuration element configures the aligned ends to further refine the global alignment algorithm.
+ * Particularly, the ends of the alignment can be penalised with gap costs or not. For example, the semi-global
+ * alignment does not penalise the leading and trailing gaps of one sequence while it does for the other sequence.
+ *
+ * The class is instantiated with an object of seqan3::align_cfg::end_gaps. The user can configure each of the
+ * gap specifier separately, such that it allows for maximal flexibility to configure the alignment algorithm.
+ * However, there are also predefined seqan3::align_cfg::end_gaps configurations: all_ends_free, none_ends_free,
```
eseiler: I don't know if you want them to link, but if you do, please use `seqan3::align_cfg::all_ends_free` and so on

include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
```diff
@@ -390,7 +414,25 @@ end_gaps(ends_t && ...) -> end_gaps<std::remove_reference_t<ends_t>...>;
...
+ * seq1_ends_free, and seq2_ends_free. Use these predefined values whenever possible.
+ *
+ * If this configuration element is not specified for the alignment algorithm it will automatically default to
```
eseiler:
```suggestion
 * If this configuration element is not specified for the alignment algorithm, it will automatically default to
```

include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
```diff
@@ -390,7 +414,25 @@ end_gaps(ends_t && ...) -> end_gaps<std::remove_reference_t<ends_t>...>;
...
+ * none_ends_free, which computes a global alignment.
```
eseiler:
```suggestion
 * seqan3::align_cfg::none_ends_free, which computes a global alignment.
```

If you want to link

include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
rrahn: But they won't link for whatever reason. So I removed it. Anyway, I changed it to link to the member group.

include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
rrahn: No I don't because it does not work

eseiler: Literally the only changes it didn't try myself do not work 😬 
It has to be `seqan3::align_cfg::aligned_ends::none_ends_free` (I tried it)
same for the `all_ends_free` and so on

rrahn: Ok, because he thinks it is a part of the aligned_ends class. But this is ugly, so I keep it as it is in the new version. (update in a second)

eseiler: I agree, it's a way too long link

rrahn:
@eseiler many thanks for playing around with the doxygen stuff. Can you have another look?

PR #803, opened 2019-03-13
[DOC] Update API documentation of align_cfg::scoring.

rrahn:


include/seqan3/alignment/configuration/align_config_scoring.hpp
```diff
@@ -21,8 +21,21 @@
...
+ *
+ * The scoring scheme allows to specify how two symbols of an alphabet are scored inside of the alignment algorithm.
+ * The scheme depends on the alphabet type of the passed sequences and must be chosen accordingly.
+ * During the configuration of the pairwise alignment algorithm a static assert is triggered if the scoring scheme
+ * is not compatible with the given alphabet types. Accordingly, this configuration cannot
+ * be defaulted since it depends on the sequences and must be given as a minimal configuration.
+ *
+ * ### Example
+ *
+ * \include snippet/alignment/configuration/minimal_alignment_configuration.cpp
```
joergi-w: minimal_alignment_config.cpp
(actually we do not need to specify the whole path, right?)

rrahn: It works, that is right, but I am pretty sure, this can't be good if there are multiple files with the same name. The example path only points to the source directory so giving the full path should be necessary to avoid conflicts.

PR #807, opened 2019-03-13
[MISC] Improve file_open_error message.

smehringer:
If you handle multiple files this is crucial to know which file cannot be openend


smehringer:
we just test that it throws a file open error but not the exact message.

PR #810, opened 2019-03-13
Cleanup views

h-2:
* moves all the views that will be part of C++20 into our std-module and adds a couple of new ones. 
* correct the naming throughout the library


h-2:
This touches a lot so it would be nice get it in soon :mouse:

include/seqan3/std/ranges
```diff
@@ -280,4 +295,646 @@ constexpr bool enable_view<type> = static_cast<bool>(seqan3::detail::enable_view
...
+template <std::Iterator it_t, std::Sentinel<it_t> sen_t>
+using subrange = ::ranges::subrange<it_t, sen_t>;
+
+} // namespace std::ranges
+
+// -----------------------------------------------------------------------------------------------------
+// View adaptor
+// -----------------------------------------------------------------------------------------------------
+
+namespace std::ranges::view
```
smehringer: I thought `std::view`? .... I don't like namespaces :weary:

include/seqan3/std/ranges
```diff
@@ -280,4 +295,646 @@ constexpr bool enable_view<type> = static_cast<bool>(seqan3::detail::enable_view
...
+
+/*!\brief A range adaptor that forwards views and "view-wraps" containers (you will likely not need to use this unless defining a new view).
+ * \tparam    urng_t The type of the range being processed. See below for requirements.
+ * \param[in] urange The range being processed.
+ * \returns A view over the elements of the underlying range.
+ * \ingroup ranges
+ *
+ * ### View properties
+ *
+ * This view is **source-only**, it can only be at the beginning of a pipe of range transformations. The
```
smehringer: Really? Can't I have it on the right hand side? (even if it does not make much sense)

include/seqan3/std/ranges
```diff
@@ -280,4 +295,646 @@ constexpr bool enable_view<type> = static_cast<bool>(seqan3::detail::enable_view
...
+ * |                                 |                                       |                                                    |
+ * | seqan3::reference_t             |                                       | seqan3::reference_t<urng_t>                        |
+ *
+ * See the \link view view submodule documentation \endlink for detailed descriptions of the view properties.
+ *
+ * \hideinitializer
+ */
+inline constexpr auto drop = ::ranges::view::drop;
+
+/*!\brief               A range adaptor that returns elements from the underlying range until the functor evaluates to
```
smehringer: Isn't this wrong for drop_while? 
Returns elements from the underlying range after the functor evaluated to true for the first time?

smehringer:
Can you please reference the daft? seems like I am looking at the wrong thing

include/seqan3/std/ranges
```diff
@@ -280,4 +295,646 @@ constexpr bool enable_view<type> = static_cast<bool>(seqan3::detail::enable_view
...
+ * \hideinitializer
+ */
+inline constexpr auto filter = ::ranges::view::filter;
+
+} // namespace std::ranges::view
+
+namespace std::ranges
+{
+
+/*!\brief               A view over an input stream.
```
smehringer:
```suggestion
/*!\brief               Makes the underlying range behave like an input range.
```
? 
The brief right now does not really tell you anything. You can also use view::all to make a view over an input stream.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -550,7 +549,7 @@ class alignment_algorithm :
         auto it_first_seq_end = std::ranges::begin(first_range);
         std::ranges::advance(it_first_seq_end, end_coordinate.first_seq_pos);
 
-        using first_subrange_type = seqan3::view::subrange<decltype(it_first_seq_begin), decltype(it_first_seq_end)>;
+        using first_subrange_type = std::ranges::subrange<decltype(it_first_seq_begin), decltype(it_first_seq_end)>;
```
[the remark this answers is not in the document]
h-2: All views, as in types that model std::ranges::View are in std::ranges.

Only the view adaptors are in std::ranges::view / std::view.

subrange is a type and not an adaptor (which is also why we can alias it easily).

include/seqan3/io/sequence_file/format_fasta.hpp
```diff
@@ -197,7 +195,7 @@ class sequence_file_format_fasta
         {
             if (options.truncate_ids)
             {
-                std::ranges::copy(stream_view | ranges::view::drop_while(is_id || is_blank)     // skip leading >
+                std::ranges::copy(stream_view | std::view::drop_while(is_id || is_blank)     // skip leading >
```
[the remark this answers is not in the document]
h-2: It was merged in Kona, but has not yet appeared in the unofficial draft document. See https://wg21.link/p1035r4 all of this was basically merged.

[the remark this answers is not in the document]
h-2: (wording might change a little still)

include/seqan3/range/view/take_until.hpp
```diff
@@ -410,9 +410,9 @@ namespace seqan3::view
  * See the \link view view submodule documentation \endlink for detailed descriptions of the view properties.
  *
  * This view only *preserves* certain concepts if the specified functor models
- * seqan3::regular_std::Invocable<fun_t, reference_t<urng_t>, i.e.
+ * std::RegularInvocable<fun_t, reference_t<urng_t>, i.e.
```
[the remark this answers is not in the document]
h-2: this was just broken documentation as you can see :wink:

include/seqan3/std/ranges
h-2: everything is defined in std::ranges::view and then std::view is declared an alias of std::ranges::view. This enables view stuff to use stuff from std::ranges without prefix, but still appear in a different namespace.

include/seqan3/std/ranges
h-2: yep, copy'n'paste error

include/seqan3/std/ranges
h-2: > Really?

nope

include/seqan3/std/ranges
h-2: > The brief right now does not really tell you anything. 

It tells you all you need :laughing: 

> You can also use subrange to make a view over an input stream.

This is a hack though. You can see that in our I/O, that we always need to do the weird check at the end, whether its actually at end, and increment once more et cetera.

Once these things are documented on en.cppreference.com we will remove our documentation anyway and link there.

smehringer: what does it do to no-input ranges?

include/seqan3/std/ranges
```diff
@@ -286,4 +301,645 @@ constexpr bool enable_view<type> = static_cast<bool>(seqan3::detail::enable_view
...
+ * |                                 |                                       |                                                    |
+ * | seqan3::reference_t             |                                       | seqan3::reference_t<urng_t>                        |
+ *
+ * See the \link view view submodule documentation \endlink for detailed descriptions of the view properties.
+ *
+ * \hideinitializer
+ */
+inline constexpr auto drop = ::ranges::view::drop;
+
+/*!\brief               A range adaptor that elements from the underlying range as long as the functor evaluates to
```
smehringer: I like those elementing adaptors :laughing:

include/seqan3/std/ranges
h-2: I understand neither of these questions :see_no_evil:

smehringer: Ignore the first.

2. what happens: `istream_view(std::vector<int>)`?

h-2: You can't. The documentation states that the view is source-only and that it has exactly one parameter:
```
 * \param[in] istr      The object of type std::istream.
```

smehringer: ah ok.

include/seqan3/std/ranges
smehringer: If this is changes I will approve

h-2: done

PR #811, opened 2019-03-13
Fix/span

h-2:


include/seqan3/std/span
```diff
@@ -253,137 +253,137 @@ class span<span_tp, dynamic_extent> {
...
                           nullptr_t> = nullptr) noexcept
-        : data{other.get_data()}, size{other.get_size()} {}
+        : data_{other.data()}, size_{other.size()} {}
 
 //    ~span() noexcept = default;
 
     template <ptrdiff_t count>
     inline constexpr span<element_type, count> first() const noexcept
     {
-        static_assert(count >= 0, "Count must be >= 0 in span::first()");
```
eseiler: This is in an constexpr context, so it could be `static_asserts` (?)
So why change it?

h-2: i was a little over-ambitious with removing the wrong asserts maybe

h-2:
> I guess changing all static_asserts to asserts is fine since the STL documents out of range errors to be undefined behaviour.

No, those were actual code errors in LLVM's STL, you cannot static assert on `data()`, even if `data()` is constexpr, because `data()` is not always evaluated in a constexpr-context.

smehringer:
Well then next time a comment in the PR would be very helpful for the review :pray:

h-2:
> Well then next time a comment in the PR would be very helpful for the review pray

:+1:

PR #812, opened 2019-03-13
[FIX] Also serialise support structures

eseiler:
We also need to store/load the data structures that enable the determination of the position in a text collection

This will be hopefully soon replaced by a way shorter code with cereal


h-2:
for real? 

:man_facepalming: 

We really need to get that cereal support into sdsl

PR #816, opened 2019-03-18
[FIX] Fixes construction of end_gaps from lvalues.

rrahn:
Before construction of the end_gaps from lvalues of the end gap specified was not working because of explicit rvalue semantics in the constructor.

Also renames the end gap specifiers and the predefined end gap values to be more consistent with the general naming scheme of the alignment module.

### DIFF
--- a/include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
+++ b/include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
@@ -85,7 +85,7 @@ public:
  * Using a `bool` allows to dynamically set the value if the option is only known at runtime. If the option is
  * already known at compile time the static version will be the preferred option.
  *
- * \see seqan3::first_seq_trailing
+ * \see seqan3::back_end_first
  * \see seqan3::second_seq_leading
  * \see seqan3::second_seq_trailing
  */
@@ -107,10 +107,10 @@ front_end_first(value_t) -> front_end_first<value_t>;
 //!\}
 
 // ----------------------------------------------------------------------------
-// first_seq_trailing
+// back_end_first
 // ----------------------------------------------------------------------------
 
-/*!\brief The penalty configuration for a aligning the back of the first sequence with a gap.
+/*!\brief The penalty configuration for aligning the back of the first sequence with a gap.
  * \ingroup alignment_configuration
  * \tparam value_t The type of the value to be wrapped. Can be of type std::true_type, std::false_type or bool.
  *
@@ -121,7 +121,7 @@ front_end_first(value_t) -> front_end_first<value_t>;
  * \see second_seq_trailing
  */
 template <typename value_t>
-struct first_seq_trailing : public sequence_end_gap_specifier_base<value_t>
+struct back_end_first : public sequence_end_gap_specifier_base<value_t>
 {
     //!\privatesection
     //!\brief An internal id to allow consistency checks with other gap specifiers.
@@ -129,12 +129,12 @@ struct first_seq_trailing : public sequence_end_gap_specifier_base<value_t>
 };
 
 /*!\name Type deduction guides
- * \relates seqan3::first_seq_trailing
+ * \relates seqan3::back_end_first
  * \{
  */
 //!\brief Deduces the template argument from the type of the wrapped value.
 template <typename value_t>
-first_seq_trailing(value_t) -> first_seq_trailing<value_t>;
+back_end_first(value_t) -> back_end_first<value_t>;
 //!\}
 
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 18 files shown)


include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
```diff
@@ -547,9 +500,60 @@ inline constexpr end_gaps seq1_ends_free{first_seq_leading<std::true_type>{},
...
+ * \tparam end_gaps_t The type of the end-gaps. Must be of type seqan3::end_gaps.
+ *
+ * \details
+ *
+ * This configuration element configures the aligned ends to further refine the global alignment algorithm.
+ * Particularly, the ends of the alignment can be penalised with gap costs or not. For example, the semi-global
+ * alignment does not penalise the leading and trailing gaps of one sequence while it does for the other sequence.
+ *
+ * The class is instantiated with an object of seqan3::end_gaps. The user can configure each of the
+ * gap specifier separately, such that it allows for maximal flexibility to configure the alignment algorithm.
```
eseiler:
```suggestion
 * gap specifier separately allowing for maximal flexibility when configuring the alignment algorithm.
```

include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
```diff
@@ -547,9 +500,60 @@ inline constexpr end_gaps seq1_ends_free{first_seq_leading<std::true_type>{},
...
+ * However, there are also predefined \ref predefined_end_gap_configurations "configurations", which should be
```
eseiler:
```suggestion
 * However, there are also predefined \ref predefined_end_gap_configurations "configurations" which should be
```

include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
```diff
@@ -547,9 +500,60 @@ inline constexpr end_gaps seq1_ends_free{first_seq_leading<std::true_type>{},
...
+ * preferred whenever possible.
+ *
+ * If this configuration element is not specified for the alignment algorithm, it will automatically default to
+ * \ref seqan3::end_gaps::none_ends_free "none_ends_free", which computes a global alignment.
```
eseiler:
```suggestion
 * \ref seqan3::end_gaps::none_ends_free "none_ends_free" which computes a global alignment.
```

include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
```diff
@@ -69,7 +69,7 @@ struct seq_end_gap_base
 };
 
 // ----------------------------------------------------------------------------
-// first_seq_leading
+// front_end_first
```
smehringer: does `first` correspond to the "first" sequence or sequence 1? Because you call free ends `first_ends_free` and now `first` is at the end.. maybe make this consistent? But I have no overview on all the member names

include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
```diff
@@ -292,13 +292,13 @@ class end_gaps
     ~end_gaps()                                      noexcept = default;
 
     //!\brief Construction from at least one sequence end-gap specifier.
-    constexpr end_gaps(ends_t && ...args) noexcept
+    constexpr end_gaps(ends_t const ...args) noexcept
```
smehringer: Why do you copy the arguments?

include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
```diff
@@ -547,9 +500,60 @@ inline constexpr end_gaps seq1_ends_free{first_seq_leading<std::true_type>{},
...
+namespace seqan3::align_cfg
+{
+
+// ----------------------------------------------------------------------------
+// aligned_ends
+// ----------------------------------------------------------------------------
+
+/*!\brief The configuration for aligned sequence ends.
+ * \ingroup alignment_configuration
+ * \tparam end_gaps_t The type of the end-gaps. Must be of type seqan3::end_gaps.
```
smehringer:
```suggestion
 * \tparam end_gaps_t The type of the end-gaps. Must be a specialisation of seqan3::end_gaps.
```

include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
rrahn: i see your concern.

include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
rrahn: because  they are only one byte big and those cheaper then using lvalues.

PR #817, opened 2019-03-18
[DOC] Add documentation for align_pairwise.

rrahn:


include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -38,10 +38,92 @@
...
 {
 
+/*!\brief Computes a pairwise alignment over a sequence pair or a range of sequence pairs.
+ * \ingroup pairwise_alignment
+ *
+ * \tparam sequence_t         The type of sequence pairs. Either a type modelling seqan3::tuple_like_concept
+ *                            with a tuple size of exactly two or std::ranges::InputRange whose value type models the
+ *                            before mentioned tuple concept.
+ * \tparam alignment_config_t The type of the alignment configuration; must be a seqan3::configuration.
+ * \param[in] seq             A tuple with two sequences or an input range over such tuples.
```
h-2: Is it really just InputRange?

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -38,10 +38,92 @@
...
+ *                            "alignment configurations".
+ * \return Depending on the configuration, the return type of align_pairwise is either `void` if the alignment was
+ * configured to call a delegate on each hit or a seqan3::alignment_range over the alignment results.
+ *
+ * \details
+ *
+ * This function computes the pairwise alignment of a sequence pair or a range of sequence pairs.
+ * The type of the elements stored in the tuples must model std::ranges::ForwardRange.
+ * The value category of these elements is either an lvalue or if the passed tuple or range is an lvalue, xvalues are
+ * also allowed as tuple elements.
```
h-2: What is the intention of this sentence? Which input are not allowed, can you give me an example?
And does the user understand this?

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -38,10 +38,92 @@
...
+ *
+ * For each alignment one or more alignment results can be computed depending on the configuration.
+ * Before computing the alignment the respective alignment algorithm is configured depending on the given alignment
+ * configuration. If the configuration enables a two-way execution (the default case), a seqan3::alignment_range is
```
h-2: What's two-way execution and why should I care? :thinking:

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -38,10 +38,92 @@
...
+ * returned. The alignment is first computed when calling begin on the returned range or the respective iterator is
+ * incremented.
```
h-2: I would maybe say that "alignments are computed on-demand  when iterating over the results", maybe that's easier to understand?

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -38,10 +38,92 @@
...
+ * ### Example
+ *
+ * The following snippets demonstrate the single element and the range based interface.
+ *
+ * \include test/snippet/alignment/pairwise/align_pairwise.cpp
+ *
+ * ### Exception
+ *
+ * Might throw std::bad_alloc if it fails to allocate the alignment matrix or seqan3::invalid_alignment_configuration
+ * if the configuration is invalid.
```
h-2: What kind of guarantee?

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -38,10 +38,92 @@
...
+ *
+ * The complexity depends on the configured algorithm. For the \ref seqan3::align_cfg::edit "edit distance"
+ * the following worst cases over two input sequences of size \f$ N \f$ can be assumed:
+ *
+ * | Component        | Runtime           | Space            |
+ * |------------------|-------------------|------------------|
+ * | Score only       | \f$ O(N*N/w) \f$  | \f$ O(w) \f$     |
+ * | Back coordinate  | \f$ O(N*N/w) \f$  | \f$ O(w) \f$     |
+ * | Begin coordinate | \f$ O(N*N/w) \f$  | \f$ O(N*N/w) \f$ |
+ * | Alignment        | \f$ O(N*N/w) \f$  | \f$ O(N*N/w) \f$ |
```
h-2: Did we agree on whether to call "the alignment" just "alignment" or "alignment trace" or ...?

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -38,10 +38,92 @@
...
+ * \f$ w \f$ is the size of a machine word.
+ *
+ * For all other algorithms that compute the standard dynamic programming algorithm the following worst cases hold:
+ *
+ * | Component        | Runtime         | Space          |
+ * |------------------|-----------------|----------------|
+ * | Score only       | \f$ O(N^2) \f$  | \f$ O(N) \f$   |
+ * | Back coordinate  | \f$ O(N^2) \f$  | \f$ O(N) \f$   |
+ * | Begin coordinate | \f$ O(N^2) \f$  | \f$ O(N^2) \f$ |
+ * | Alignment        | \f$ O(N^2) \f$  | \f$ O(N^2) \f$ |
```
h-2: You use `N^2` here, but `N*N` in the previous table :nail_care:

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -38,10 +38,92 @@
...
+ *
+ * ### Thread safety
+ *
+ * As long as no shared state is passed to the algorithm via the configuration object, the algorithm is thread safe and
```
h-2: How can you "pass a shared state to the algorithm via the configuration object"? When would I want to do that and why?

I would be maybe write:

This function is *re-entrant*, i.e. it is always safe to call in parallel **with different inputs**. It is *thread-safe*, i.e. it is safe to call in parallel **with the same input** under the condition that 
1. The input sequences do not change when being iterated over.
2. The configuration ... [does not contain config_element_xy?]

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -38,10 +38,92 @@
...
 //!\cond
+    requires tuple_like_concept<sequence_t> &&
+             detail::is_type_specialisation_of_v<remove_cvref_t<alignment_config_t>, configuration>
+//!\endcond
+constexpr auto align_pairwise(sequence_t && seq, alignment_config_t && config)
+{
+    static_assert(std::tuple_size_v<std::remove_reference_t<sequence_t>> == 2,
+                  "Expects exactly two sequences for pairwise alignments.");
+
+    return align_pairwise(ranges::view::single(std::forward<sequence_t>(seq)) | std::view::common,
```
h-2: please add view::single to std/ranges once my other changes there are merged

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -38,10 +38,92 @@
...
+
+//!\copydoc align_pairwise
 template <std::ranges::InputRange sequence_t, typename alignment_config_t>
+//!\cond
     requires detail::is_type_specialisation_of_v<remove_cvref_t<alignment_config_t>, configuration> &&
              tuple_like_concept<value_type_t<std::ranges::iterator_t<std::remove_reference_t<sequence_t>>>>
+//!\endcond
 constexpr auto align_pairwise(sequence_t && seq, alignment_config_t const & config)
 {
     // Wrap in persist to make the code also work with temporaries that are not views.
```
h-2: Why are you allowing this? Won't the result range outlive the temporaries?

include/seqan3/alignment/pairwise/align_pairwise.hpp
rrahn: It really is. The range that provides the sequence only needs to be iterated once.

include/seqan3/alignment/pairwise/align_pairwise.hpp
rrahn: Something like std::tie(rng1, rng2) gives you a prvalue with lvalue references wrapped inside. 
Something like std::tuple{"ACGT", "ACGT"} is only allowed, if the tuple itself is not a xvalue. That is because accessing the values of the tuple via get inside the alignment will return lvalues but they become dangling references after the invocation of align_pairwise. But I agree that the sentence might be confusing.

include/seqan3/alignment/pairwise/align_pairwise.hpp
rrahn: We agreed on not calling it trace or traceback. For the align_result class we decided on alignment. So yes, align_pairwise computes an alignment.

include/seqan3/alignment/pairwise/align_pairwise.hpp
rrahn: No it is N*N/w in the previous table

include/seqan3/alignment/pairwise/align_pairwise.hpp
rrahn: Well, because if I only compute the score or the coordinates I don't need the sequences afterwards. But this needs more checking on the call site. But this will be part of other fixes and details

include/seqan3/alignment/pairwise/align_pairwise.hpp
rrahn: yes

include/seqan3/alignment/pairwise/align_pairwise.hpp
rrahn: Because it can be different in the future. But I see that I rephrase it

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -38,32 +36,125 @@
...
 
+/*!\brief Computes a pairwise alignment over a sequence pair or a range of sequence pairs.
+ * \ingroup pairwise_alignment
+ *
+ * \tparam sequence_t         The type of sequence pairs. Either a type modelling seqan3::tuple_like_concept
+ *                            with a tuple size of exactly two or std::ranges::InputRange whose value type models the
+ *                            before mentioned tuple concept.
+ * \tparam alignment_config_t The type of the alignment configuration; must be a seqan3::configuration.
+ * \param[in] seq             A tuple with two sequences or an input range over such tuples.
+ * \param[in] config          A configuration object containing \ref alignment_configuration
```
smehringer:
```suggestion
 * \param[in] config          A configuration object containing \ref seqan3::alignment_configuration
```

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -38,32 +36,125 @@
...
+ *                            "alignment configurations".
+ * \return Depending on the configuration, the return type of align_pairwise is either `void` if the alignment was
+ * configured to call a delegate on each hit or a seqan3::alignment_range over the alignment results.
```
smehringer: what do you mean by `calling a delegate`?

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -38,32 +36,125 @@
...
+ *
+ * \details
+ *
+ * This function computes the pairwise alignment of a sequence pair or a range over sequence pairs.
+ * The pairs can be any class template that models the seqan3::tuple_like_concept.
+ * In case of calling the alignment with a single tuple, the value is passed as a std::view::single to the range
+ * aware interface. In this case the passed class must be std::CopyConstructible and both element types of the tuple
+ * must model in std::ranges::ViewableRange because it is copied into the std::view::single. The same
```
smehringer:
```suggestion
 * must model std::ranges::ViewableRange because it is copied into the std::view::single. The same
```

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -38,32 +36,125 @@
...
+ * ```cpp
+ * std::pair p{"ACGTAGC", "AGTACGACG"};
+ * auto rng = align_pairwise(p, config);
+ * ```
+ * You can use std::tie instead as shown in the example below.
+ *
+ * ### Accessing the alignment results
+ *
+ * For each sequence pair one or more \ref seqan3::align_result "seqan3::align_result"s can be computed.
+ * The align_pairwise function returns an seqan3::alignment_range which can be used to iterate over the
```
smehringer:
```suggestion
 * The seqan3::align_pairwise function returns an seqan3::alignment_range which can be used to iterate over the
```

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -38,32 +36,125 @@
...
+ * alignments. The alignments are then computed on-demand when iterating over the results.
```
smehringer:
```suggestion
 * alignments. The computation is lazy-evaluated which means that the alignments are computed on-demand when iterating over the result range.
```

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -38,32 +36,125 @@
...
+ *
+ * Strong exception guarantee.
+ *
+ * Might throw std::bad_alloc if it fails to allocate the alignment matrix or seqan3::invalid_alignment_configuration
+ * if the configuration is invalid.
+ *
+ * ### Complexity
+ *
+ * The complexity depends on the configured algorithm. For the \ref seqan3::align_cfg::edit "edit distance"
+ * the following worst cases over two input sequences of size \f$ N \f$ can be assumed:
```
smehringer:
```suggestion
 * the following worst case scenarios over two input sequences of size \f$ N \f$ can be assumed:
```

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -38,32 +36,125 @@
...
+ * | Computing        | Runtime           | Space            |
+ * |------------------|-------------------|------------------|
+ * | score            | \f$ O(N^2/w) \f$  | \f$ O(w) \f$     |
+ * | back coordinate  | \f$ O(N^2/w) \f$  | \f$ O(w) \f$     |
+ * | front coordinate | \f$ O(N^2/w) \f$  | \f$ O(N^2/w) \f$ |
+ * | alignment        | \f$ O(N^2/w) \f$  | \f$ O(N^2/w) \f$ |
+ *
+ * \f$ w \f$ is the size of a machine word.
+ *
+ * For all other algorithms that compute the standard dynamic programming algorithm the following worst cases hold:
```
smehringer:
```suggestion
 * For all other algorithms that compute the standard dynamic programming algorithm the following worst case scenarios hold:
```

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -38,32 +36,125 @@
...
+ * | score            | \f$ O(N*k) \f$  | \f$ O(k) \f$   |
+ * | back coordinate  | \f$ O(N*k) \f$  | \f$ O(k) \f$   |
+ * | front coordinate | \f$ O(N*k) \f$  | \f$ O(N*k) \f$ |
+ * | alignment        | \f$ O(N*k) \f$  | \f$ O(N*k) \f$ |
+ *
+ * \f$ k \f$ is the size of the band.
+ *
+ * ### Thread safety
+ *
+ * This function is re-entrant, i.e. it is always safe to call in parallel with different inputs. It is thread-safe,
```
smehringer:
```suggestion
 * This function is re-entrant, i.e. it is always safe to call it in parallel with different inputs. It is thread-safe,
```

include/seqan3/alignment/pairwise/align_pairwise.hpp
rrahn: now this is safe because it allows rvalue non-view ranges if the element types of the returned tuple are ViewableRanges.

include/seqan3/alignment/pairwise/alignment_configurator.hpp
```diff
@@ -33,6 +33,36 @@
...
+// This covers also transforming and non-transforming views (e.g. std::ranges::view::zip, or view::take).
+// Only a temporary non-view range piped with view::persist can't be handled securely.
+template <typename range_t>
+SEQAN3_CONCEPT align_pairwise_range_input_concept =
+    std::ranges::InputRange<std::remove_reference_t<range_t>> &&
+    align_pairwise_value_concept<value_type_t<std::remove_reference_t<range_t>>> &&
+    ((std::ranges::ViewableRange<range_t> && std::is_lvalue_reference_v<reference_t<range_t>>) ||
+     align_pairwise_single_input_concept<std::remove_reference_t<reference_t<range_t>>>);
+
+//!\endcond
```
h-2: why is this \cond? Please add at least short developer documentation.
Also concepts are CamelCase now.

test/snippet/alignment/pairwise/align_pairwise.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+#include <seqan3/alignment/configuration/all.hpp>
+#include <seqan3/alignment/pairwise/align_pairwise.hpp>
+#include <seqan3/alphabet/nucleotide/dna4.hpp>
+
+using namespace seqan3;
+
+int main()
+{
+
```
h-2: remove empty line

test/snippet/alignment/pairwise/align_pairwise.cpp
```diff
@@ -0,0 +1,24 @@
...
+                    std::pair{"AGTTACGAC"_dna4, "AGTAGCGATCG"_dna4}};
+
+    // Compute the alignment of a single pair.
+    for (auto const & res : align_pairwise(std::tie(vec[0].first, vec[0].second), align_cfg::edit))
+        debug_stream << "The score: " << res.get_score() << "\n";
+
+    // Compute the alignment over a range of pairs.
+    for (auto const & res : align_pairwise(vec, align_cfg::edit))
+        debug_stream << "The score: " << res.get_score() << "\n";
+
```
h-2: remove empty line

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -38,32 +36,124 @@
...
+/*!\brief Computes a pairwise alignment over a sequence pair or a range of sequence pairs.
+ * \ingroup pairwise_alignment
+ *
+ * \tparam sequence_t         The type of sequence pairs. Either a type modelling seqan3::tuple_like_concept
+ *                            with a tuple size of exactly two or std::ranges::InputRange whose value type models the
+ *                            before mentioned tuple concept.
+ * \tparam alignment_config_t The type of the alignment configuration; must be a seqan3::configuration.
+ * \param[in] seq             A tuple with two sequences or an input range over such tuples.
+ * \param[in] config          The object storing the alignment configuration.
+ *                            "alignment configurations".
```
smehringer: Did you want to reference this? Otherwise I would delete this..

include/seqan3/alignment/pairwise/alignment_configurator.hpp
```diff
@@ -33,6 +33,45 @@
...
+    AlignPairwiseValue<value_t> &&
+    std::ranges::ViewableRange<std::tuple_element_t<0, value_t>> &&
+    std::ranges::ViewableRange<std::tuple_element_t<1, value_t>>;
+
+/*!\brief A helper concept to test for correct range input in seqan3::align_pairwise.
+ * \ingroup pairwise_alignment
+ *
+ * \details
+ *
+ * Only use input ranges whose value type models AlignPairwiseValue and
```
smehringer:
```suggestion
 * Only use input ranges whose value type models seqan3::AlignPairwiseValue and
```

include/seqan3/alignment/pairwise/alignment_configurator.hpp
```diff
@@ -33,6 +33,45 @@
...
+ * whose reference type is an lvalue reference and the range itself models ViewableRange or
```
smehringer:
```suggestion
 * whose reference type is an lvalue reference and the range itself models std::ranges::ViewableRange or
```

include/seqan3/alignment/pairwise/alignment_configurator.hpp
```diff
@@ -33,6 +33,45 @@
...
+ * the reference type is a prvalue and it models AlignPairwiseSingleInput.
```
smehringer:
```suggestion
 * the reference type is a prvalue and it models seqan3::AlignPairwiseSingleInput.
```

include/seqan3/alignment/pairwise/alignment_configurator.hpp
```diff
@@ -73,14 +114,21 @@ struct alignment_contract
             using scoring_type = std::remove_reference_t<
                                     decltype(get<align_cfg::scoring>(std::declval<alignment_config_type>()).value)
                                  >;
-            return static_cast<bool>(scoring_scheme_concept<scoring_type, value_type_t<first_seq_t>,
-                                                                          value_type_t<second_seq_t>>);
+            return static_cast<bool>(scoring_scheme_concept<scoring_type,
+                                     value_type_t<std::remove_reference_t<first_seq_t>>,
```
smehringer: Isn't `value_type_t` independent of a reference in the type?

PR #820, opened 2019-03-22
Misc/align result

rrahn:
Only renaming stuff.


doc/tutorial/introduction/introduction_align.cpp
```diff
@@ -41,11 +41,11 @@ int main()
...
     // Call a pairwise alignment with edit distance and traceback.
     for (auto && res : align_pairwise(std::make_pair(sequences[0], sequences[1]),
-                                      align_cfg::edit | align_cfg::result{align_cfg::with_trace}))
+                                      align_cfg::edit | align_cfg::result{with_alignment}))
     {
         // Print the resulting score and the alignment.
-        debug_stream << res.get_score() << '\n';              // => -4
-        debug_stream << res.get_alignment() << '\n';          // =>       0     .    :
+        debug_stream << res.score() << '\n';              // => -4
+        debug_stream << res.alignment() << '\n';          // =>       0     .    :
```
joergi-w: Please indent the comment according to the following lines. Otherwise you destroy the expected output.

test/unit/alignment/pairwise/align_pairwise_test.cpp
```diff
@@ -33,21 +33,21 @@ TEST(align_pairwise, single_rng_lvalue)
...
-        configuration cfg = align_cfg::edit | align_cfg::result{align_cfg::with_score};
+        configuration cfg = align_cfg::edit | align_cfg::result{with_score};
         for (auto && res : align_pairwise(p, cfg))
         {
-            EXPECT_EQ(res.get_score(), -4);
+            EXPECT_EQ(res.score(), -4);
         }
     }
 
     {  // the trace
```
joergi-w: the alignment (traceback)

include/seqan3/alignment/matrix/alignment_coordinate.hpp
```diff
@@ -99,8 +99,8 @@ class advanceable_alignment_coordinate
         requires !std::Same<other_state, state>
     //!\endcond
     constexpr advanceable_alignment_coordinate(advanceable_alignment_coordinate<other_state> const & other) :
-        first_seq_pos{other.first_seq_pos},
-        second_seq_pos{other.second_seq_pos}
+        first{other.first},
+        second{other.second}
```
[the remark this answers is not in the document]
rrahn: Actually, a good point. We could make it inherit from std::pair and offer the tuple_element and tuple_size and the get interface as well. But this is not yet really needed. I put it on a card for the alignment module to think about it. But the intention is a) keep it close to familiar naming and b) the class is called alignment_coordinate and calling the member first and second is already very expressive. I think first_seq_pos does not add any to the understanding as it is redundant in this context.

doc/tutorial/introduction/introduction_align.cpp
rrahn: Yes, I have seen this also, but was to lazy to adapt it right away 😄 👍

PR #828, opened 2019-03-24
argument_parser no longer uses app_name for synopsis.

tloka:
Instead, the name of the binary has to be set manually in the synopsis definition.
This is necessary for a distinction between app name and binary name which are usually different.
For printing on command line, man or HTML, the first word of the synopsis (detected by the first ' ' or '\t') is still displayed in bold font.

### DIFF
--- a/include/seqan3/argument_parser/detail/format_help.hpp
+++ b/include/seqan3/argument_parser/detail/format_help.hpp
@@ -309,9 +309,8 @@ protected:
         for (unsigned i = 0; i < meta.synopsis.size(); ++i)
         {
             std::string text = "\\fB";
-            text.append(meta.app_name);
-            text.append("\\fP ");
             text.append(meta.synopsis[i]);
+            text.insert(text.find_first_of(" \t"), "\\fP");
 
             print_line(text, false);
         }
--- a/include/seqan3/argument_parser/auxiliary.hpp
+++ b/include/seqan3/argument_parser/auxiliary.hpp
@@ -99,7 +99,9 @@ struct argument_parser_meta_data // holds all meta information
      *        is separated by a new line.
      */
     std::vector<std::string> description;
-    //!\brief Add lines of usage to the synopsis section of the help page (e.g. "[OPTIONS] FILE1 FILE1").
+    /*!\brief Add lines of usage to the synopsis section of the help page (e.g. 
+     *        "./my_read_mapper [OPTIONS] FILE1 FILE1").
+     */
     std::vector<std::string> synopsis;
     /*!\brief Provide some examples on how to use your tool and what standard
      *        parameters might be appropriate in different cases (e.g.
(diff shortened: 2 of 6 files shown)


include/seqan3/argument_parser/detail/format_help.hpp
```diff
@@ -309,9 +309,8 @@ class format_help : public format_base
         for (unsigned i = 0; i < meta.synopsis.size(); ++i)
         {
             std::string text = "\\fB";
-            text.append(meta.app_name);
-            text.append("\\fP ");
             text.append(meta.synopsis[i]);
+            text.replace(text.find_first_of(" \t"), 0, "\\fP");
```
smehringer: why `\t` ?

tloka: I oriented on `is_blank` function that considers tab and space.
When used properly it's probably not necessary here. However, it's also no significant extra cost afaik. So I decided to include it.

tloka:
btw, is there a reason there is a leading underscore in `format_html::_print_synopsis(...)`?
I know that sometimes this is used for private functions (which is also the case here), but it doesn't seem to be consistent with the naming of the other functions?

include/seqan3/argument_parser/detail/format_help.hpp
smehringer: Hm I don't know if we need this but I guess it is fine.
What's the second parameter to replace doing?

smehringer:
> reason there is a leading underscore in format_html::_print_synopsis(...)?

nope, seems like I forgot to change it when transferring code from seqan2.

> I know that sometimes this is used for private functions (which is also the case here)

You usually use it if you cannot make your variable private in the class scope but you want to mark it as "should be private" to discourage its usage.

include/seqan3/argument_parser/detail/format_help.hpp
tloka: > What's the second parameter to replace doing?

The length of replacement, e.g.:

```
std::string str "hello world";
str.replace(str.find_first_of(" "), 0, ""); // hello world
str.replace(str.find_first_of(" "), 1, ""); // helloworld
str.replace(str.find_first_of(" "), 3, ""); // hellorld
```

Actually, now I see it, std::string::insert should also work and could be slightly faster (I guess), although theoretical complexity is the same:
```
std::string str "hello world";
str.insert(str.find_first_of(" ")+1, "new "); // hello new world
```

shall I change this?

smehringer: Well if one is concerned about performance we would not create temporary strings at all :grin: 
but we are not in the argument parser. `insert` is nicer though because the interface is better known so yes that would be nice

tloka:
ping @smehringer

PR #829, opened 2019-03-24
[FIX] Change parentheses initialization to brace initialization for v…

joshuak94:
…arious snippet examples in the argument parser module.

### DIFF
--- a/include/seqan3/argument_parser/validators.hpp
+++ b/include/seqan3/argument_parser/validators.hpp
@@ -267,18 +267,6 @@ public:
         extensions = c ? v : std::vector<std::string>{v | view::to_lower};
     }
 
-    /*!\brief Constructing from an initializer_list.
-     * \param[in] v The initializer_list of valid file extensions to test (e.g. {"fa", "fasta"}).
-     * \param[in] c Case sensitivity flag. Set true for case sensitivity. Default: false (case insensitive).
-     *
-     * For case insensitivity, everything is converted to lower case characters.
-     */
-    file_ext_validator(std::initializer_list<std::string> const & v, bool const c = false) :
-        case_sensitive{c}
-    {
-        extensions = c ? v : std::vector<std::string>{v | view::to_lower};
-    }
-
     /*!\brief Tests whether the filepath \p path ends with a valid extension.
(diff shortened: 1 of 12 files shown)


h-2:
Why was the second commit (explicit string) necessary?

joshuak94:
There was a bug where parentheses works with the implicit string, but changing it to curly braces all of a sudden caused a run time error... I think @smehringer knows more?

h-2:
@smehringer what's the problem here?

joshuak94:
@smehringer is this still good to merge?

smehringer:
I want to wait for the unit tests introduced in #848 so I can investigate why we need to construct explicitly from std::string for the file_ext_validator before we merge this PR.

joshuak94:
So, brace initialization of something like the `file_ext_validator` without explicitly giving the file extensions as strings will cause a runtime error:
```
terminate called after throwing an instance of 'std::length_error'
  what():  basic_string::_M_create
[1]    17242 abort      argument_parser/validators_3_snippet
```

But with an explicit `std::string{}` call, or by using parenthesis initialization, there is no error. Any thoughts? @smehringer

joshuak94:
Can I bump this? @smehringer :)

h-2:
Snippet tests still fail.

smehringer:
@joshuak94 Please remove the `file_ext_validator` initializer_list constructor. Then everything will work :)
We don't need it anyway, now that we have the extra "case sensitive* parameter

joshuak94:
So then should I also remove the tests which use an initializer list?

smehringer:
no, just leave everything as-is!
I'll explain:

With the extra `case_sensitive` parameter, the initializer list is obsolete, because an initializer list must contain values of the same type.
```cpp
file_ext_validator{"hi", "ho", true}; // never possible
file_ext_validator{"hi", "ho"}; // right now possible because of initializer list but weird ()  
file_ext_validator{{"hi", "ho"}, true}; // possible, but delegates to initializer list which tries to initialize a list of list of strings, which we do not want :(  
```
The good thing: The third one is also possible with the vector because the vector has an initializer list constructor :) so it will just work (worked on my branch :D).

PR #833, opened 2019-03-24
[FIX,DOC] seq5 has also at least length 5

JensUweUlrich:
in excercise 'fun with file ranges' seq5 is also has a size of at least 5


smehringer:
Yes but the exercise is two **only print the first two** that have a minimum length of five :)

PR #835, opened 2019-03-24
[FIX] empty sequence files are valid files that result into an empty range.

### LINKED ISSUE #823: parse_error when loading empty FASTA file
Problem
---
Loading empty FASTA file using sequence_file_input causes:
```
terminate called after throwing an instance of 'seqan3::parse_error'
  what():  Expected to be on beginning of ID, but (is_char<'>'> || is_char<';'>) evaluated to false on '�'
```

For me this is unexpected behavior, since e.g. pipelines could be negatively affected by throwing an exception (imagine a pipeline giving FASTA as intermediate analysis result, this file being empty (no matches => expected behavior) leading to a crash of the complete pipeline).

Proposed Solution
---
I would rather consider to return an empty field. The app developer can still decide to throw an error on empty field or to handle internally.

smehringer:
Resolves #823

### DIFF
--- a/test/unit/io/structure_file/structure_file_input_test.cpp
+++ b/test/unit/io/structure_file/structure_file_input_test.cpp
@@ -44,6 +44,7 @@ struct structure_file_input_class : public ::testing::Test
     test::tmp_filename create_file(char const * contents)
     {
         test::tmp_filename filename{"structure_file_input_constructor.dbn"};
+
         {
             std::ofstream filecreator{filename.get_path(), std::ios::out | std::ios::binary};
             filecreator << contents; // must contain at least one record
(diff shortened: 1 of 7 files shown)


test/unit/io/structure_file/structure_file_input_test.cpp
```diff
@@ -66,7 +66,10 @@ TEST_F(structure_file_input_class, construct_by_filename)
 {
     /* just the filename */
     {
-        test::tmp_filename filename = create_file("> ID\nACGU\n....\n");
```
h-2: what is create_file is it still used somewhere else?

test/unit/io/structure_file/structure_file_input_test.cpp
```diff
@@ -66,7 +66,10 @@ TEST_F(structure_file_input_class, construct_by_filename)
...
+        test::tmp_filename filename{"structure_file_input_constructor.dbn"};
+        {
```
h-2: please introduce a newline before beginning of scope

test/unit/io/structure_file/structure_file_input_test.cpp
smehringer: `create_file` is used in a lot of places, I removed it here to test an empty file :point_up:

smehringer:
> You have added tests for reading empty compressed files, but you didn't add tests for reading uncompressed empty files, right?

Yes I did. I remove the "must contain at least one record" lines that write into a temporary file which results in an empty file being tested.

test/unit/io/structure_file/structure_file_input_test.cpp
smehringer: :sleepy: Now I need to update the PR because of one newline. Is there anything else?

PR #836, opened 2019-03-25
[FIX] out-of-range in random access benchmarks

h-2:
I broke the documentation debug nightly builds yesterday 🏆

### DIFF
--- a/test/performance/range/view/view_drop_view_take_benchmark.cpp
+++ b/test/performance/range/view/view_drop_view_take_benchmark.cpp
@@ -106,7 +106,7 @@ void random_access(benchmark::State & state)
     std::vector<size_t> access_positions;
     access_positions.resize(1'000'000);
     std::mt19937 gen(42);
-    std::uniform_int_distribution<size_t> dis(0, 1'000'000);
+    std::uniform_int_distribution<size_t> dis(0, 998'000 - 1);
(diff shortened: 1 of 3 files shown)


test/performance/range/view/view_drop_view_take_benchmark.cpp
```diff
@@ -106,7 +106,7 @@ void random_access(benchmark::State & state)
     std::vector<size_t> access_positions;
     access_positions.resize(1'000'000);
     std::mt19937 gen(42);
-    std::uniform_int_distribution<size_t> dis(0, 1'000'000);
+    std::uniform_int_distribution<size_t> dis(0, 900'000);
```
eseiler:
```suggestion
    std::uniform_int_distribution<size_t> dis(0, 1'000'000 - 1);
```
The resulting range should be `1'000'000` in size?

h-2: No, this is just the sampling range, not the number of elements! And view::drop reduces the size of the final range, so we cannot jump to position 999'999

eseiler: But we drop 3*1000 from 1'003'000 ... so it's 1'000'000?

PR #839, opened 2019-03-25
[FIX] Allow view::complement on const containers

eseiler:
We need to *really* strip the const when checking for the `NucleotideAlphabet` concept.

### DIFF
--- a/include/seqan3/range/view/complement.hpp
+++ b/include/seqan3/range/view/complement.hpp
@@ -64,7 +64,7 @@ namespace seqan3::view
 
 inline auto const complement = deep{std::view::transform([] (auto && in)
 {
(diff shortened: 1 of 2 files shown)


smehringer:
@h-2 @eseiler  
As I see it, a `const dna4` should model `NulceotideAlphabet`, no? Don't we want to strip the constness in the concept then instead of in every usage?

eseiler:
Depends, `AlphabetConcept` also checks assignability. So it is correct that `dna4 const &` fails since it is not assignable. Thats why, in this case, we want to strip the constness.

This is ok for views, since we are not going to change the underlying data.
In general, we expect to be able to assign when requiring the `AlphabetConcept` - so if we just strip the const in the concept, we will run into problems if we actually want to assign something.

An alternative would be to split the concept into a *read-only* part and then the rest that checks assignability and such.

PR #842, opened 2019-03-25
[FEATURE] Adds buffer_queue from google-concurrency-library.

rrahn:
This queue is an excerpt from the google concurrency library and implements the interface as proposed by [P0260R3](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0260r3.html)

It is not clear if we can use this. We first need a license change of the library. Might be that we have to port our old queue from SeqAn2 and adapt the interface to use the proposal interface.


rrahn:
@marehr there is still the license issue. I am currently porting the SeqAn2 concurrent queue. Which should be done today as well. Our implementation has some advantages. First, non-blocking behavior to push and pop. Second,  dynamically resizable.

rrahn:
Superseded by #949

PR #848, opened 2019-03-27
Case (in-)sensitive file extension validator

tloka:
Created case insensitivity option (optional construction parameter) for file_ext_validator using to_lower view.
Introduced tests for file_ext_validator including case (in-)sensitivity tests.

### DIFF
--- a/include/seqan3/argument_parser/validators.hpp
+++ b/include/seqan3/argument_parser/validators.hpp
@@ -20,9 +20,10 @@
 #include <seqan3/core/concept/core_language.hpp>
 #include <seqan3/core/metafunction/basic.hpp>
 #include <seqan3/core/metafunction/pre.hpp>
-#include <seqan3/std/filesystem>
 #include <seqan3/range/container/concept.hpp>
+#include <seqan3/range/view/to_lower.hpp>
 #include <seqan3/std/concepts>
+#include <seqan3/std/filesystem>
 #include <seqan3/std/ranges>
 
 namespace seqan3
@@ -256,17 +257,27 @@ public:
 
     /*!\brief Constructing from a vector.
      * \param[in] v The vector of valid file extensions to test (e.g. {"fa", "fasta"}).
+     * \param[in] c Case sensitivity flag. Set true for case sensitivity. Default: false (case insensitive).
+     *              
+     * For case insensitivity, everything is converted to lower case characters.
      */
-    file_ext_validator(std::vector<std::string> const & v) :
-        extensions{v}
-    {}
+    file_ext_validator(std::vector<std::string> const & v, bool const c = false) :
+        case_sensitive{c}
+    {
+        extensions = c ? v : std::vector<std::string>{v | view::to_lower};
+    }
 
     /*!\brief Constructing from an initializer_list.
      * \param[in] v The initializer_list of valid file extensions to test (e.g. {"fa", "fasta"}).
+     * \param[in] c Case sensitivity flag. Set true for case sensitivity. Default: false (case insensitive).
+     *              
+     * For case insensitivity, everything is converted to lower case characters.
      */
-    file_ext_validator(std::initializer_list<std::string> const & v) :
-        extensions{v}
-    {}
+    file_ext_validator(std::initializer_list<std::string> const & v, bool const c = false) :
+        case_sensitive{c}
+    {
+        extensions = c ? v : std::vector<std::string>{v | view::to_lower};
+    }
 
     /*!\brief Tests whether the filepath \p path ends with a valid extension.
      * \param path The input value to check.
@@ -275,7 +286,12 @@ public:
     void operator()(std::filesystem::path const & path) const
     {
         std::string ext{path.extension().string()};
-        ext = ext.substr(std::min(1, static_cast<int>(ext.size()))); // drop '.' if extension is non-empty
+        ext = ext.substr(std::min(1, std::max(0, static_cast<int>(ext.size()) - 1))); // drop '.' if ext is non-empty
+
+        // extensions were transformed to lower case during construction,so do the same for input path
+        if (!case_sensitive)
+            ext = std::string{ext | view::to_lower};
+
         if (!(std::find(extensions.begin(), extensions.end(), ext) != extensions.end()))
             throw parser_invalid_argument(detail::to_string("Extension ", ext, " is not one of ",
                                                             std::view::all(extensions), "."));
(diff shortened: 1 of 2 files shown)


include/seqan3/argument_parser/validators.hpp
```diff
@@ -256,17 +257,38 @@ class file_ext_validator
 
     /*!\brief Constructing from a vector.
      * \param[in] v The vector of valid file extensions to test (e.g. {"fa", "fasta"}).
+     * \param[in] c Case sensitivity flag. Set true for case sensitivity.
+     *              Default: false (case insensitive).
+     *              For case insensitivity, the initializer list is transformed to lower case characters.
```
smehringer: this can go into details :)

```
     * \param[in] c Compare file extensions case sensitive or not. Default: false (case insensitive).
     *
     * For case insensitivity, everything is converted to lower case characters.
```

include/seqan3/argument_parser/validators.hpp
```diff
@@ -256,17 +257,38 @@ class file_ext_validator
...
      */
-    file_ext_validator(std::vector<std::string> const & v) :
-        extensions{v}
-    {}
+    file_ext_validator(std::vector<std::string> const & v, bool const c = false) :
+        case_sensitive{c}
+    {
+        if (c)
+        {
+            extensions = std::vector<std::string>(v);
```
```suggestion
            extensions = v;
```
This should be fine?

include/seqan3/argument_parser/validators.hpp
```diff
@@ -256,17 +257,38 @@ class file_ext_validator
...
+        }
+        else
+        {
+            auto ext_view = v | view::to_lower;
+            extensions = std::vector<std::string>(ext_view);
```
```suggestion
            extensions = std::vector<std::string>(v | view::to_lower);
```
This does not work?

include/seqan3/argument_parser/validators.hpp
```diff
@@ -275,10 +297,24 @@ class file_ext_validator
     void operator()(std::filesystem::path const & path) const
     {
         std::string ext{path.extension().string()};
-        ext = ext.substr(std::min(1, static_cast<int>(ext.size()))); // drop '.' if extension is non-empty
-        if (!(std::find(extensions.begin(), extensions.end(), ext) != extensions.end()))
-            throw parser_invalid_argument(detail::to_string("Extension ", ext, " is not one of ",
-                                                            std::view::all(extensions), "."));
+        ext = ext.substr(std::min(1, std::max(0,static_cast<int>(ext.size())-1))); // drop '.' if extension is non-empty
```
```suggestion
        ext = ext.substr(std::min(1, std::max(0, static_cast<int>(ext.size()) - 1))); // drop '.' if extension is non-empty
```

eseiler:
🧀

include/seqan3/argument_parser/validators.hpp
```diff
@@ -275,10 +297,24 @@ class file_ext_validator
...
+
+        // Case sensitive comparison
+        if (case_sensitive)
```
smehringer: You can reduce some duplicated code :) 
```cpp
std::string ext_tmp;

if (case_sensitive)
    ext_tmp = ext;
else
    ext_tmp = std::string{ext | view::to_lower}

if (!(std::find(extensions.begin(), extensions.end(), ext_tmp) != extensions.end()))
    throw parser_invalid_argument(detail::to_string("Extension ", ext, " is not one of ",                                                             
                                  std::view::all(extensions), "."));

```

test/unit/argument_parser/format_parse_validators_test.cpp
```diff
@@ -76,6 +76,118 @@ TEST(validator_test, file_exists)
...
+    {
+        const char * argv[] = {"./argument_parser_test", "-s", "/absolute/path/file.txt"};
+        argument_parser parser("test_parser", 3, argv);
+        parser.add_option(option_value, 's', "string-option", "desc",
+                          option_spec::DEFAULT, no_extension_file_ext_validator);
+
+        EXPECT_THROW(parser.parse(), validation_failed);
+    }
+
+        // check trailing dot with no_ext_validator => failure
```
smehringer:
```suggestion
    // check trailing dot with no_ext_validator => failure
```

include/seqan3/argument_parser/validators.hpp
tloka: I can reduce even more and omit the tmp variable ;)
I think like this the error message is even mor intuitive because you can directly see what was internally compared:

```
        if ( !case_sensitive )
            ext = std::string{ext | view::to_lower};

        if (!(std::find(extensions.begin(), extensions.end(), ext) != extensions.end()))
                throw parser_invalid_argument(detail::to_string("Extension ", ext, " is not one of ",
                                                                std::view::all(extensions), "."));
```

include/seqan3/argument_parser/validators.hpp
tloka: Of course it does :weary:

tloka:
ping @smehringer

include/seqan3/argument_parser/validators.hpp
```diff
@@ -256,17 +257,33 @@ class file_ext_validator
...
+     *              
+     * For case insensitivity, everything is converted to lower case characters.
      */
-    file_ext_validator(std::vector<std::string> const & v) :
-        extensions{v}
-    {}
+    file_ext_validator(std::vector<std::string> const & v, bool const c = false) :
+        case_sensitive{c}
+    {
+        if (c)
```
smehringer: I think we can even do that in one line :sunglasses: 
```suggestion
        extensions = (c) ? std::vector<std::string>{std::move(v)} : std::vector<std::string>{v | view::to_lower};
```
(then you need the cast to vector though to return the same type, but the move will prevent the copy)
if the line length is too long though I would just leave it as you have it now :)

include/seqan3/argument_parser/validators.hpp
```diff
@@ -275,7 +292,12 @@ class file_ext_validator
     void operator()(std::filesystem::path const & path) const
     {
         std::string ext{path.extension().string()};
-        ext = ext.substr(std::min(1, static_cast<int>(ext.size()))); // drop '.' if extension is non-empty
+        ext = ext.substr(std::min(1, std::max(0, static_cast<int>(ext.size())-1))); // drop '.' if ext is non-empty
```
```suggestion
        ext = ext.substr(std::min(1, std::max(0, static_cast<int>(ext.size()) - 1))); // drop '.' if ext is non-empty
```

include/seqan3/argument_parser/validators.hpp
```diff
@@ -275,7 +292,12 @@ class file_ext_validator
...
+
+        // extensions were transformed to lower case during construction,so do the same for input path
+        if ( !case_sensitive )
```
```suggestion
        if (!case_sensitive)
```
:nail_care: sorry :speak_no_evil:

tloka: > :nail_care: sorry :speak_no_evil:

:unamused:

PR #849, opened 2019-03-27
[FIX,DOC] nucleotide complement table

### LINKED ISSUE #740: The complement table in the nucleotide documentation is wrong
... better check the implementation, too.

eseiler:
Resolves #740 

The implementation is correct.


h-2:
Was this only wrong in the doc or also the implementation?

eseiler:
> The implementation is correct.

PR #850, opened 2019-03-27
[FIX,DOC] Revise Setup tutorial

### LINKED ISSUE #786: Add "How to add another cpp to your cmake file" to the setup tutorial
We do not document how to proceed with the following tutorials after hello_world.cpp compiles and runs.
(http://docs.seqan.de/seqan/3.0.0-master-user/setup.html)
EDIT: maybe we should just alter the CMake script so it copmiles every cpp file in `source`

eseiler:
Resolves #786

### DIFF
--- a/doc/setup/index.md
+++ b/doc/setup/index.md
@@ -38,14 +38,14 @@ brew install gcc@7
 sudo port install gcc-7
 ```
 
-**Linux/MacOS** using [conda](https://conda.io)
+**Linux** using [conda](https://conda.io)
 ```
 conda create -n gcc7 -c quantstack gcc-7 libgcc-7
 conda activate gcc7
 ```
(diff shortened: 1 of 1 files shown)


doc/setup/index.md
```diff
@@ -147,3 +147,24 @@ In this case you can try to export the Path:
 export PATH=/util/bin:$PATH
 ```
 and try running cmake again.
+
+# Adding a new source file to your project
+
+If create a new `cpp` file and want to compile it, you need to add another `add_executable` and `target_link_libraries`
```
smehringer: If change this, would be nice.

```suggestion
If you create a new `cpp` file and want to compile it, you need to add another `add_executable` and `target_link_libraries`
```

PR #852, opened 2019-03-27
Fix the take creating a subrange over a temp. iterator. Also remove a…

joshuak94:
… type on random_access_range.hpp.

### DIFF
--- a/include/seqan3/range/view/drop.hpp
+++ b/include/seqan3/range/view/drop.hpp
@@ -44,10 +44,19 @@ private:
     //!\brief Befriend the base class so it can call impl().
     friend base_t;
 
+    template <std::ranges::ViewableRange urng_t>
+    static auto impl(urng_t && urange, size_t drop_size)
+    {
+        return std::forward<urng_t>(urange) | std::view::drop(drop_size);
+    }
+
     /*!\brief       Call the view's constructor with the underlying view as argument.
      * \returns     An instance of std::ranges::subrange.
      */
     template <std::ranges::ViewableRange urng_t>
+    //!\cond
+        requires std::is_lvalue_reference_v<urng_t> && !std::ranges::ContiguousRange<urng_t>
+    //!\endcond
     static auto impl(urng_t && urange, size_t drop_size)
     {
         auto b = std::ranges::begin(urange);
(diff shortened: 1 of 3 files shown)


include/seqan3/range/view/take.hpp
```diff
@@ -457,7 +457,8 @@ class take_fn : public pipable_adaptor_base<take_fn<exactly, or_throw>>
      */
     template <std::ranges::ViewableRange urng_t>
     //!\cond
-        requires std::ranges::RandomAccessRange<urng_t> && std::ranges::SizedRange<urng_t>
+        requires std::ranges::RandomAccessRange<urng_t> && std::ranges::SizedRange<urng_t> &&
+                 std::is_lvalue_reference_v<urng_t>
```
smehringer: So lvalue views are fine too?
If this is not used for overload resolution we decided in one of the strategy meetings that it should be done as a static assert :point_up:

joshuak94: From what I'm understanding, this is indeed for overload resolution... Because we want viewable ranges which are RandomAccess and Sized but not lvalue to go through the call above this one. @rrahn is that right?

rrahn: That is correct. We cannot take the subrange over a temporary view. So in this case we need to resolve to the most generic case which is returning a take_view.

rrahn:
@joshuak94 there is also the view drop that has the same problem. I just checked and we need a new overload that resolves to the std::view::drop if type-erasure can't be done via subrange.

include/seqan3/range/view/take.hpp
```diff
@@ -591,15 +592,17 @@ namespace seqan3::view
...
+ * |:-------------------------------------------------------------------------------------:|:-----------------------------:|
+ * | std::basic_string *or* std::basic_string_view                                         | std::basic_string_view        |
+ * | std::ranges::SizedRange && std::ranges::ContiguousRange                               | std::span                     |
+ * | std::ranges::SizedRange && std::ranges::RandomAccessRange && std::is_lvalue reference | std::ranges::subrange         |
+ * | *else*                                                                                | *implementation defined type* |
  *
  * This adaptor is different from std::view::take in that it performs type erasure for some underlying ranges.
- * It returns exactly the type specified above.
+ * It returns exactly the type specified above. For a std::ranges::SizedRange and std::ranges::RandomAccessRange, the `urng_t`
+ * must satisfy std::is_lvalue_reference_v, as otherwise, a subrange over a temporary iterator is created, which has
```
```suggestion
 * must satisfy std::is_lvalue_reference_v, as otherwise, a subrange over a temporary iterator is created which has
```

joshuak94:
@rrahn I added the view::drop fix as a second commit...

h-2:
What is the status here? CI still fails.

include/seqan3/range/view/drop.hpp
```diff
@@ -179,14 +192,16 @@ namespace seqan3::view
...
- * |:---------------------------------------------------------:|:--------------------------------------------------:|
- * | std::basic_string *or* std::basic_string_view             | std::basic_string_view                             |
- * | std::ranges::SizedRange && std::ranges::ContiguousRange   | std::span                                          |
- * | *else*                                                    | std::ranges::subrange                              |
+ * | `urng_t` (underlying range type)                             | `rrng_t` (returned range type)                     |
+ * |:------------------------------------------------------------:|:--------------------------------------------------:|
+ * | std::basic_string *or* std::basic_string_view                | std::basic_string_view                             |
+ * | std::ranges::SizedRange && std::ranges::ContiguousRange      | std::span                                          |
+ * | std::is_lvalue_reference_v && ! std::ranges::ContiguousRange | std::ranges::subrange                              |
```
rrahn:
```suggestion
 * | std::is_lvalue_reference_v && !std::ranges::ContiguousRange  | std::ranges::subrange                              |
```

include/seqan3/range/view/drop.hpp
```diff
@@ -44,10 +44,23 @@ class drop_fn : public pipable_adaptor_base<drop_fn>
     //!\brief Befriend the base class so it can call impl().
     friend base_t;
 
+    /*!\brief       This constructor is used when urange is not an l-value reference
+     *              and is not a ContiguousRange.
```
rrahn:
```suggestion
     *              and is not a std::ranges::ContiguousRange.
```

include/seqan3/range/view/drop.hpp
```diff
@@ -44,10 +44,23 @@ class drop_fn : public pipable_adaptor_base<drop_fn>
...
+     * \returns     An instance of std::view::drop.
+    */
+    template <std::ranges::ViewableRange urng_t>
+    static auto impl(urng_t && urange, size_t drop_size)
```
rrahn:
```suggestion
    static auto impl(urng_t && urange, size_t const drop_size)
```

joshuak94:
I think Hannes submitted a new pull request superceding this one!

rrahn:
yes I just saw it.

h-2:
Yep, there was some related stuff that needed fixing so I did it all in one, I hope that was ok.

PR #856, opened 2019-03-27
[WIP] Update concept documentation + tests for views

tloka:
Content of this PR
---

I introduced a new test structure to test the concepts for all views that didn't look too complicated for me to touch. The new structure should give a more intuitive overview of preservance/guarantee/loss.
Thereby, I also added missing tests for ViewableRange and ContiguousRange.

The test structure
---

I created an enum for all different range concepts.
Using that enum, it is now possible to check for fulfillment, preserving, guarantee or loss via functions using initialization lists:

```c++
// Example for preserved, other types (fulfill, guarantee, loss) work in a similar way
using namespace seqan3::test;
std::string vec{"HelloWorld"};
auto view = vec | std::view::reverse;
EXPECT_TRUE((preserved<decltype(vec),decltype(view)>({Input, Forward})));
```
For lost and guaranteed concepts that cannot be properly tested for the input range, there are functions weak_lost(...) and weak_guaranteed(...) that only test the correct output type.

If `EXPECT_TRUE` fails, the error message gives a combined output of the underlying test functions and the GoogleTest output which would look like this:
```
Preserved check of concept 'OutputRange' failed:
 IN_TYPE: 1
OUT_TYPE: 0
are expected to be equal.
/home/lokat/workspace/seqan3-fork/test/unit/range/view/view_complement_test.cpp:58: Failure
Value of: (preserved<decltype(vec), decltype(v1)>({Input, Forward, Bidirectional, RandomAccess, Sized, Common, ConstIterable, Output}))
  Actual: false
Expected: true
```

With this combintion of output, it is ensured to get all necessary information about what failed (which was actually not too trivial to achieve): The concept type, which expectations were not fulfilled and the original line where the FAIL() occured.

The following views are already included and using the new functions:

- char_to
- complement
- convert
- deep
- get
- kmer_hash
- persist
- to_char
- to_rank
- translation
- trim

The following tests remain unchanged:

- pairwise_combine
- single_pass_input
- take
- take_line
- take_until

Issues
---

- `view_trim_test.cpp`: ContiguousRange is not preserved (test fails when uncommented)

WIP
---

- The remaining tests should be updated before merging PR.
- Test range type of input and output?

### DIFF
--- a/include/seqan3/range/view/get.hpp
+++ b/include/seqan3/range/view/get.hpp
@@ -1,4 +1,4 @@
-// -----------------------------------------------------------------------------------------------------
+ // -----------------------------------------------------------------------------------------------------
 // Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
 // Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
 // This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
@@ -33,22 +33,22 @@ namespace seqan3::view
  *
  * ### View properties
  *
- * | range concepts and reference_t  | `urng_t` (underlying range type)      | `rrng_t` (returned range type)                          |
- * |---------------------------------|:-------------------------------------:|:-------------------------------------------------------:|
- * | std::ranges::InputRange         | *required*                            | *preserved*                                             |
- * | std::ranges::ForwardRange       |                                       | *preserved*                                             |
- * | std::ranges::BidirectionalRange |                                       | *preserved*                                             |
- * | std::ranges::RandomAccessRange  |                                       | *preserved*                                             |
- * | std::ranges::ContiguousRange    |                                       | *lost*                                                  |
- * |                                 |                                       |                                                         |
- * | std::ranges::ViewableRange      | *required*                            | *preserved*                                             |
- * | std::ranges::View               |                                       | *preserved*                                             |
- * | std::ranges::SizedRange         |                                       | *preserved*                                             |
- * | std::ranges::CommonRange        |                                       | *preserved*                                             |
- * | std::ranges::OutputRange        |                                       | *preserved*                                             |
- * | seqan3::const_iterable_concept  |                                       | *preserved*                                             |
- * |                                 |                                       |                                                         |
- * | seqan3::reference_t             | seqan3::tuple_like_concept            | std::tuple_element_t<index, seqan3::reference_t<urng_t>>|
+ * | range concepts and reference_t  | `urng_t` (underlying range type) | `rrng_t` (returned range type)                           |
+ * |---------------------------------|:--------------------------------:|:--------------------------------------------------------:|
+ * | std::ranges::InputRange         | *required*                       | *preserved*                                              |
+ * | std::ranges::ForwardRange       |                                  | *preserved*                                              |
+ * | std::ranges::BidirectionalRange |                                  | *preserved*                                              |
+ * | std::ranges::RandomAccessRange  |                                  | *preserved*                                              |
+ * | std::ranges::ContiguousRange    |                                  | *lost*                                                   |
+ * |                                 |                                  |                                                          |
+ * | std::ranges::ViewableRange      | *required*                       | *guaranteed*                                             |
+ * | std::ranges::View               |                                  | *guaranteed*                                             |
+ * | std::ranges::SizedRange         |                                  | *preserved*                                              |
+ * | std::ranges::CommonRange        |                                  | *preserved*                                              |
+ * | std::ranges::OutputRange        |                                  | *preserved*                                              |
+ * | seqan3::const_iterable_concept  |                                  | *preserved*                                              |
+ * |                                 |                                  |                                                          |
+ * | seqan3::reference_t             | seqan3::tuple_like_concept       | std::tuple_element_t<index, seqan3::reference_t<urng_t>> |
  *
  * See the \link view view submodule documentation \endlink for detailed descriptions of the view properties.
  *
--- a/include/seqan3/range/view/get.hpp
+++ b/include/seqan3/range/view/get.hpp
@@ -1,4 +1,4 @@
- // -----------------------------------------------------------------------------------------------------
+// -----------------------------------------------------------------------------------------------------
 // Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
 // Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
 // This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
--- a/test/unit/range/view/view_concept_check.hpp
+++ b/test/unit/range/view/view_concept_check.hpp
@@ -1,0 +1,295 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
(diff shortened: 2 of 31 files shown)


include/seqan3/range/view/get.hpp
```diff
@@ -1,4 +1,4 @@
-// -----------------------------------------------------------------------------------------------------
+ // -----------------------------------------------------------------------------------------------------
```
rrahn: superfluous space

test/unit/range/view/view_persist_test.cpp
```diff
@@ -79,26 +79,36 @@ TEST(view_persist, const_)
...
+    
+    // Required
+    EXPECT_TRUE(std::ranges::InputRange<decltype(std::string{"foo"})>);
 
-    EXPECT_TRUE(std::ranges::InputRange<decltype(v1)>);
-    EXPECT_TRUE(std::ranges::ForwardRange<decltype(v1)>);
-    EXPECT_TRUE(std::ranges::BidirectionalRange<decltype(v1)>);
-    EXPECT_TRUE(std::ranges::RandomAccessRange<decltype(v1)>);
+    // Not Required (But important to be tested)
+    EXPECT_FALSE(std::ranges::ViewableRange<decltype(std::string("foo"))>);
```
rrahn: but why not simply std::string? It is literally the same expression.

tloka:
@rrahn I did it in a slightly different way than you suggested which is more intuitive to read in the test code (you directly see which concepts are tested to be preserved etc.).
What do you think about this solution (see updated PR comment)?

test/unit/range/view/view_concept_check.hpp
```diff
@@ -0,0 +1,295 @@
...
+    }
+}
+
+/*!\brief Expression to check the fulfillment of a range concept.
+ * \tparam conc    The range concept to check as seqan3::test::ConceptType object.
+ * \return True if TYPE fulfills conc. False if not.
+ * \throws std::logic_error    Throws logic error if the concept type given (concept) is unknown.
+ *
+ * ### Example
+ * using namespace seqan3::test;
```
eseiler: I'm not sure if we need snippets for this, but you could wrap it in a 
```console
```cpp
source code
\```
``` 
block (without the \)

test/unit/range/view/view_concept_check.hpp
```diff
@@ -0,0 +1,295 @@
...
+ * std::string vec{"HelloWorld"};
+ * EXPECT_TRUE(fulfilled<decltype(vec)>(ConceptType::Input));
+ */
+template<typename TYPE>
```
eseiler:
```suggestion
template <typename rng_t>
```

test/unit/range/view/view_concept_check.hpp
```diff
@@ -0,0 +1,295 @@
...
+        case ConstIterable:
+            return const_iterable_concept<TYPE>;
+        default:
+            throw std::logic_error("Unexpected ConceptType in view_concept_check::fulfilled(...).");
+    }
+}
+
+/*!\brief Checks if the fulfillment of a concept is preserved from IN_TYPE to OUT_TYPE.
+ * \tparam concepts    A list of range concepts to check as seqan3::test::ConceptType object.
+ * \return True if fulfillment of conc is preserved from IN_TYPE to OUT_TYPE. False if not.
```
eseiler:
```suggestion
 * \return True if in_t and out_t model the same concept. False otherwise.
```

test/unit/range/view/view_concept_check.hpp
```diff
@@ -0,0 +1,295 @@
...
+ * std::string vec{"HelloWorld"};
+ * auto view = vec | std::view::reverse;
+ * EXPECT_TRUE((preserved<decltype(vec),decltype(view)>({Input, Forward}))); //double brackets required
+ *
+ *\details
+ *
+ * This function is designed for use inside EXPECT_TRUE(...). Using other test scenarios (e.g., EXPECT_FALSE)
+ * can lead to unintended command line output via debug_stream.
+ */
+template<std::ranges::Range IN_TYPE, typename OUT_TYPE>
```
eseiler:
```suggestion
template<std::ranges::Range in_t, typename out_t>
```

test/unit/range/view/view_concept_check.hpp
```diff
@@ -0,0 +1,295 @@
...
+ *
+ * \details
+ *
+ * The weak_guaranteed(...) function weakly checks a guarantee, i.e. only true for OUT_TYPE. 
+ * You can use guaranteed(...) function to check true for IN_TYPE and true for OUT_TYPE.
+ * This function is designed for use inside EXPECT_TRUE(...). Using other test scenarios (e.g., EXPECT_FALSE)
+ * can lead to unintended command line output via debug_stream.
+ */
+template<typename OUT_TYPE>
+constexpr bool weak_guaranteed(std::initializer_list<ConceptType> const & concepts)
```
eseiler:
```suggestion
constexpr bool weakly_guaranteed(std::initializer_list<ConceptType> const & concepts)
```
I'd say it's an adverb in this case.
(How is it guaranteed? Weakly.)

test/unit/range/view/view_kmer_hash_test.cpp
```diff
@@ -42,3 +45,16 @@ TEST(kmer_hash, view)
...
+    using namespace seqan3::test;
+    std::vector<dna4> vec{"ACGTG"_dna4};
+    auto v1 = vec | view::kmer_hash(3);
+
+    EXPECT_TRUE((preserved<decltype(vec), decltype(v1)>({Input, Forward, Bidirectional, RandomAccess, Sized, Common, 
+                                                         ConstIterable})));
+    EXPECT_TRUE((guaranteed<decltype(vec), decltype(v1)>({View})));
+    EXPECT_TRUE(weak_guaranteed<decltype(v1)>({Viewable}));
+    EXPECT_TRUE((lost<decltype(vec), decltype(v1)>({Contiguous, Output})));
+}
```
eseiler: Missing newline at end of file?
(Github put some weird symbol here)

test/unit/range/view/view_trim_test.cpp
```diff
@@ -68,19 +70,14 @@ TEST(view_trim, qualified)
...
-    EXPECT_TRUE(!std::ranges::SizedRange<decltype(v1)>);
-}
+
+    //TODO Contiguous should be preserved, but isn't
+    EXPECT_TRUE((preserved<decltype(vec), decltype(v1)>(
+        {Input, Forward, Bidirectional, RandomAccess/*, Contiguous*/, Output, ConstIterable})));
+    EXPECT_TRUE((guaranteed<decltype(vec), decltype(v1)>({View})));
+    EXPECT_TRUE(weak_guaranteed<decltype(v1)>({Viewable}));
+    EXPECT_TRUE((lost<decltype(vec), decltype(v1)>({Sized, Common})));
+}
```
eseiler: Missing new line at end of file?

test/unit/range/view/view_concept_check.hpp
tloka: Nope, it's a weak test of the guarantee ;)
I don't think this should be an adverb, there is nothing like a weak guarantee so this would be confusing.

tloka:
Waiting for comments of @rrahn or @h-2 about the general way of the implementation before changing the requests of @eseiler , modifying the remaining tests and resolving conflicts.

h-2:
We are sorry this went stale for so long and thank you for the PR. There are currently conflicts with master branch and there are incoming large changes to ranges, see #1267 .

Let's wait for those and then we can re-evaluate this if you would still be willing to update it.

rrahn:
@tloka many thanks for this contribution and your work on that. We have put some emphasis on other things in the past months and will continue in a more module based approach for the future. At the moment we will put our focus on search. I would therefor close this PR at the moment and reevaluate it later. Especially since a lot of things have changed. Thank you very much for understanding.

PR #860, opened 2019-03-28
Repeat view

smehringer:
and small changes to the random_access_iterator


include/seqan3/range/view/repeat.hpp
```diff
@@ -0,0 +1,331 @@
...
+ * | std::ranges::ContiguousRange    |                                                    |
+ * |                                 |                                                    |
+ * | std::ranges::ViewableRange      | *guaranteed*                                       |
+ * | std::ranges::View               | *guaranteed*                                       |
+ * | std::ranges::SizedRange         |                                                    |
+ * | std::ranges::CommonRange        |                                                    |
+ * | std::ranges::OutputRange        | *guaranteed*                                       |
+ * | seqan3::const_iterable_concept  | *guaranteed*                                       |
+ * |                                 |                                                    |
+ * | seqan3::reference_t             | seqan3::value_type_t<it_t>                         |
```
rrahn:
```suggestion
 * | seqan3::reference_t             | &value_t                                           |
```

include/seqan3/range/view/repeat.hpp
```diff
@@ -0,0 +1,331 @@
...
+
+// ---------------------------------------------------------------------------------------------------------------------
+// view::repeat (factory instance)
+// ---------------------------------------------------------------------------------------------------------------------
+
+/*!\name General purpose views
+ * \{
+ */
+/*!\brief                A view factory that repeats a given value infinitely.
+ * \tparam value_t       The type of value to repeat wrapped in a std::view::single.
```
rrahn:
```suggestion
 * \tparam value_t       The type of value to repeat wrapped in a std::view::single; must model std::CopyConstructible.
```

test/unit/range/view/view_repeat_test.cpp
```diff
@@ -0,0 +1,116 @@
...
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
+// -----------------------------------------------------------------------------------------------------
+
+#include <gtest/gtest.h>
+
+#include <seqan3/core/metafunction/range.hpp>
+#include <seqan3/core/metafunction/iterator.hpp>
```
rrahn: alphabetical order

test/unit/range/view/view_repeat_test.cpp
```diff
@@ -0,0 +1,116 @@
...
+    EXPECT_TRUE((std::is_copy_assignable_v<decltype(v_const)>));
+    EXPECT_TRUE((std::is_move_assignable_v<decltype(v_const)>));
+}
+
+TEST(general, concept)
+{
+    char chr{'A'};
+    auto v = view::repeat(chr);
+
+    EXPECT_TRUE((std::ranges::RandomAccessRange<decltype(v)>));
```
rrahn: Could you also test the other concepts explicitly.

test/unit/range/view/view_repeat_test.cpp
```diff
@@ -0,0 +1,116 @@
...
+    auto it = v.begin();
+    EXPECT_EQ(*it, 'A');
+
+    ++it;
+    EXPECT_EQ(*it, 'A');
+    EXPECT_NE(it, v.end());
+
+    // assignment
+    *it = 'X';
+    EXPECT_EQ(*it, 'X');
```
rrahn: You should test all other operators as well. Like decrement, distance, postscript and weakly ordered operators.

include/seqan3/range/detail/random_access_iterator.hpp
```diff
@@ -15,6 +15,7 @@
 #include <cassert>
 #include <type_traits>
 
+#include <seqan3/core/metafunction/transformation_trait_or.hpp>
```
rrahn: it seems you don't need it anymore?

test/unit/range/view/view_repeat_test.cpp
smehringer: Well, I would just test the random_access_iterator since I inherit all the functionality?

test/unit/range/view/view_repeat_test.cpp
```diff
@@ -0,0 +1,135 @@
...
+}
+
+TEST(general, bracket_operator)
```
rrahn: subscript_operator?

include/seqan3/range/view/repeat.hpp
```diff
@@ -0,0 +1,337 @@
...
+    repeat_view(repeat_view &&) = default;                  //!< Default.
+    repeat_view & operator=(repeat_view &&) = default;      //!< Default.
+    ~repeat_view() = default;                               //!< Default.
+
+    //!\brief Construct from any type (Note: the value will be copied into view::single).
+    repeat_view(value_t & value) : single_value{std::view::single(value)}
+    {}
+
+    //!\overload
+    repeat_view(value_t && value) : single_value{std::view::single(std::move(value))}
```
rrahn:
```suggestion
    constexpr explicit repeat_view(value_t && value) : single_value{std::view::single(std::move(value))}
```

include/seqan3/range/view/repeat.hpp
```diff
@@ -0,0 +1,337 @@
...
+     *
+     * ### Complexity
+     *
+     * Constant.
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    iterator begin() noexcept
```
rrahn: constexpr to the functions

include/seqan3/range/view/repeat.hpp
```diff
@@ -0,0 +1,337 @@
...
+// ---------------------------------------------------------------------------------------------------------------------
+// repeat (factory)
+// ---------------------------------------------------------------------------------------------------------------------
+
+//!\brief View factory definition for view::repeat.
+struct repeat_fn
+{
+    //!\brief Returns an instance of seqan3::detail::repeat_view constructed with \p value.
+    template <typename value_type>
+    auto operator()(value_type && value) const
```
rrahn:
```suggestion
    constexpr auto operator()(value_type && value) const
```

test/snippet/range/view/repeat.cpp
```diff
@@ -0,0 +1,21 @@
...
+#include <seqan3/range/view/repeat.hpp>
+#include <seqan3/std/ranges>
+
+using namespace seqan3;
+
+int main()
+{
+    auto v = view::repeat('A');
+
+    debug_stream << *v.begin() << std::endl; // prints 'A'
```
rrahn:
```suggestion
    debug_stream << *std::ranges::begin(v) << std::endl; // prints 'A'
```

test/snippet/range/view/repeat.cpp
```diff
@@ -0,0 +1,21 @@
...
+    debug_stream << v[12355]   << std::endl; // also prints 'A'. It always prints 'A'
```
rrahn:
```suggestion
    debug_stream << v[12355] << std::endl; // also prints 'A'. It always prints 'A'
```

test/snippet/range/view/repeat.cpp
```diff
@@ -0,0 +1,21 @@
...
+
+    v[1345] = 'C';
+
+    // Now it always prints 'C'
+    debug_stream << *v.begin() << std::endl; // prints 'C'
```
rrahn:
```suggestion
    debug_stream << *std::ranges::begin(v) << std::endl; // prints 'C'
```

include/seqan3/range/view/repeat.hpp
```diff
@@ -0,0 +1,337 @@
...
+        constexpr repeat_view_iterator & operator=(repeat_view_iterator const &) = default; //!< Default.
+        constexpr repeat_view_iterator (repeat_view_iterator &&) = default;                 //!< Default.
+        constexpr repeat_view_iterator & operator=(repeat_view_iterator &&) = default;      //!< Default.
+        ~repeat_view_iterator() = default;                                                  //!< Default.
+
+        //!\brief Construct by host, default position pointer with 0.
+        constexpr repeat_view_iterator(parent_type & host) noexcept : base_t{host}
+        {}
+        //!\brief Construct by host and explicit position.
+        constexpr repeat_view_iterator(parent_type & host, position_type const pos) noexcept : base_t{host, pos}
```
rrahn: but why?

include/seqan3/range/view/repeat.hpp
```diff
@@ -0,0 +1,337 @@
...
+         */
+        constexpr repeat_view_iterator() = default;                                         //!< Default.
+        constexpr repeat_view_iterator(repeat_view_iterator const &) = default;             //!< Default.
+        constexpr repeat_view_iterator & operator=(repeat_view_iterator const &) = default; //!< Default.
+        constexpr repeat_view_iterator (repeat_view_iterator &&) = default;                 //!< Default.
+        constexpr repeat_view_iterator & operator=(repeat_view_iterator &&) = default;      //!< Default.
+        ~repeat_view_iterator() = default;                                                  //!< Default.
+
+        //!\brief Construct by host, default position pointer with 0.
+        constexpr repeat_view_iterator(parent_type & host) noexcept : base_t{host}
```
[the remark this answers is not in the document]
smehringer: But then I can't construct a const_iterator from a non-const repeat_view object anymore :( ? Why does it have to be explicit?

include/seqan3/range/view/repeat.hpp
```diff
@@ -0,0 +1,337 @@
...
+        //!\}
+    };
+
+public:
+    /*!\name Associated types
+     * \{
+     */
+    //!\brief The value type (equals the value_t with any references removed).
+    using value_type = std::remove_reference_t<value_t>;
+    //!\brief The reference_type (value_type &).
```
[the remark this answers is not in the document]
smehringer: What's wrong with this?

include/seqan3/range/view/repeat.hpp
smehringer: removed it

test/unit/range/view/view_repeat_test.cpp
smehringer: But why should someone change the RA-iterator to be **not** RA? o.O

eseiler: What's the `random_access_iterator_test` for then?

include/seqan3/range/view/repeat.hpp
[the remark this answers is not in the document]
rrahn: I general nothing, but why would you need it? First, the exact definition of the reference type can be seen in the documentation, so it is superfluous to write again into the brief. Second repeating it in the brief, can introduce inconsistencies if the reference type is changed some time but not the documentation. I would keep it as simple as possible and only expand the reference brief section if it is not immediately clear to the reader what the reference type is.

include/seqan3/range/view/repeat.hpp
```diff
@@ -0,0 +1,337 @@
...
+        using typename base_t::reference;
+        //!\brief Pointer type is pointer of container element type.
+        using typename base_t::pointer;
+        //!\brief Tag this class as a random access iterator.
+        using typename base_t::iterator_category;
+
+        /*!\name Constructors, destructor and assignment
+         * \{
+         */
+        constexpr repeat_view_iterator() = default;                                         //!< Default.
```
rrahn: I think we finally agreed on Defaulted. 😅

test/unit/range/view/view_repeat_test.cpp
h-2: > So you expect your class to behaves like a RAiterator, then you should test, that it in fact, it does.

I don't think this is necessary in all cases. If you have tests that verify that your Range is RandomAccessRange, you automatically verify that the iterator satisfies all syntactic requirements of RandomAccessIterator. The same holds for SizedRange and SizedSentinel behaviour of the iterator.
If you have further tests that do something like ranges::equal, you automatically check that the semantic requirements of `++` are met; and if you then also have checks for the subscript, I think you are very much on the safe side.
Yes, if someone now changes the random_access_iterator_base to change the behavior of `--` than this might not be detected here, but that's why random_access_iterator has dedicated tests that handle these situations.

include/seqan3/range/view/repeat.hpp
```diff
@@ -0,0 +1,325 @@
...
+        using typename base_t::value_type;
+        //!\brief Use reference type defined by container.
+        using typename base_t::reference;
+        //!\brief Pointer type is pointer of container element type.
+        using typename base_t::pointer;
+        //!\brief Tag this class as a random access iterator.
+        using typename base_t::iterator_category;
+
+        //!\brief Inherit constructors from base type.
+        using base_t::base_t;
```
rrahn: I am afraid, this only works because the crtp pattern of the ra-iterator-base is wrong.  It should not have public constructors but it has. Can you make a ticket for it? It does not have high priority for the release though. But obviously when we change this, this class will not work as well. So it would be better to have the constructors working anyway. Also, our coding style requires us to also specify all implicitly defined constructors. Note, that the corresponding base class constructors are hidden anyway by the derived class' constructors

include/seqan3/range/view/repeat.hpp
```diff
@@ -0,0 +1,325 @@
...
+
+        //!\brief Inequality comparison to the sentinel always returns true on an infinite view.
+        constexpr bool operator!=(std::ranges::default_sentinel_t const &) const noexcept
+        {
+            return true;
+        }
+
+        //!\brief Equality comparison to the sentinel always returns false on an infinite view.
+        friend constexpr bool operator==(std::ranges::default_sentinel_t const &,
+                                         repeat_view_iterator const &)
```
rrahn:
```suggestion
                                         repeat_view_iterator const &) noexcept
```

include/seqan3/range/view/repeat.hpp
```diff
@@ -0,0 +1,325 @@
...
+        {
+            return false;
+        }
+
+        //!\brief Inequality comparison to the sentinel always returns true on an infinite view.
+        friend constexpr bool operator!=(std::ranges::default_sentinel_t const &,
+                                         repeat_view_iterator const &)
```
rrahn:
```suggestion
                                         repeat_view_iterator const &) noexcept
```

test/unit/range/view/view_repeat_test.cpp
rrahn: Guys, I am not talking about that someone changes the random_access_iterator_base. I am talking about someone that might (for what ever reasons) changes the iterator of the repeat_n view. And then it is clearly not tested at all; will be green in the tests, although it might be wrong because this someone else thought, everything is decently tested. 
I know that it sometimes feels over the edge, because we are trying to achieve something big companies have entire teams for. But even if we have some redundancies, this is even better. Because we stumbled more than once over things that worked in the one place but not in the other.

smehringer: fine. take a look at the test now.

include/seqan3/range/view/repeat.hpp
```diff
@@ -0,0 +1,352 @@
...
+        //!\brief Copy construction via assignment.
+        constexpr repeat_view_iterator & operator=(repeat_view_iterator const &) = default;
+        //!\brief Move constructor.
+        constexpr repeat_view_iterator (repeat_view_iterator &&) = default;
+        //!\brief Move assignment.
+        constexpr repeat_view_iterator & operator=(repeat_view_iterator &&) = default;
+        //!\brief Use default deconstructor.
+        ~repeat_view_iterator() = default;
+
+        //!\brief Construct by host.
```
rrahn: please also document the parameter.

include/seqan3/range/view/repeat.hpp
```diff
@@ -0,0 +1,352 @@
...
+        //!\brief Pointer type is pointer of container element type.
+        using typename base_t::pointer;
+        //!\brief Tag this class as a random access iterator.
+        using typename base_t::iterator_category;
+
+        /*!\name Constructors, destructor and assignment
+         * \{
+         */
+        //!\brief Default constructor.
+        constexpr repeat_view_iterator() = default;
```
rrahn: please use inline documentation

include/seqan3/range/view/repeat.hpp
```diff
@@ -0,0 +1,337 @@
...
+
+// ---------------------------------------------------------------------------------------------------------------------
+// repeat (factory)
+// ---------------------------------------------------------------------------------------------------------------------
+
+//!\brief View factory definition for view::repeat.
+struct repeat_fn
+{
+    //!\brief Returns an instance of seqan3::detail::repeat_view constructed with \p value.
+    template <typename value_type>
```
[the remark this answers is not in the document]
rrahn: but this isn't resolved?

smehringer:
@h-2 Do you want to take a look or can I merge this?
(Jenkins cannot be trusted)

PR #863, opened 2019-03-29
Feature/seq file expose valid formats

### LINKED ISSUE #892: Add valid_file_extensions function to alignment file and structured files. (see sequence_file)

rrahn:
fixes  #892

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,13 +21,24 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 
 ## New features
 
-* Traits for "metaprogramming" with `seqan3::type_list` and type packs have been added.
+#### Argument parser
+* Simplified reading file extensions from formatted files in the input/output file validators.
+
+#### Core
+* Added traits for "metaprogramming" with `seqan3::type_list` and type packs.
 
 ## API changes
 
-### The `type_list` header has moved
+#### Argument parser
 
-If you included `<seqan3/core/type_list.hpp>` you need to change the path to `<seqan3/core/type_list/type_list.hpp>`.
+* **Changed class signature of input/output file validators:**
+  Most user code will be unaffected; to fix possible compiler errors you need to add an empty template parameter list to
+  the respective instances (e.g. change `input_file_validator` to `input_file_validator<>`).
+
+#### Core
+
+* **The `type_list` header has moved:**
+  If you included `<seqan3/core/type_list.hpp>` you need to change the path to `<seqan3/core/type_list/type_list.hpp>`.
 
 ## Notable Bug-fixes
 
--- a/include/seqan3/argument_parser/validators.hpp
+++ b/include/seqan3/argument_parser/validators.hpp
@@ -21,6 +21,7 @@
 #include <seqan3/core/detail/to_string.hpp>
 #include <seqan3/core/type_traits/basic.hpp>
 #include <seqan3/core/type_traits/pre.hpp>
+#include <seqan3/io/detail/misc.hpp>
 #include <seqan3/io/detail/safe_filesystem_entry.hpp>
 #include <seqan3/range/container/concept.hpp>
 #include <seqan3/range/view/drop.hpp>
@@ -236,11 +237,15 @@ value_list_validator(std::initializer_list<const char *>) -> value_list_validato
 
 /*!\brief An abstract base class for the file and directory validators.
  * \ingroup argument_parser
+ * \tparam file_t The type of the file to get the valid extensions for; `void` on default.
  *
  * \details
  *
  * This class provides a common interface for seqan3::input_file_validator and the seqan3::output_file_validator as
  * well as the seqan3::input_directory_validator and seqan3::output_directory_validator.
+ *
+ * The type can be further specialised for the seqan3::input_file_validator and the seqan3::output_file_validator
+ * using the template argument to determine the valid extensions from the given file type.
  */
 class file_validator_base
 {
@@ -258,12 +263,6 @@ public:
     file_validator_base & operator=(file_validator_base const &) = default; //!< Defaulted.
     file_validator_base & operator=(file_validator_base &&) = default;      //!< Defaulted.
     virtual ~file_validator_base() = default;                               //!< Virtual destructor.
-
-    /*!\brief Constructs from a set of valid extensions.
-     * \param[in] extensions The valid extensions to validate for.
-     */
-    explicit file_validator_base(std::vector<std::string> extensions) : extensions{std::move(extensions)}
-    {}
     //!\}
 
     /*!\brief Tests if the given path is a valid input, respectively output, file or directory.
@@ -384,6 +383,7 @@ protected:
 /*!\brief A validator that checks if a given path is a valid input file.
  * \ingroup argument_parser
  * \implements seqan3::Validator
+ * \tparam file_t The type of the file to get the valid extensions for; `void` on default.
  *
  * \details
  *
@@ -394,23 +394,65 @@ protected:
  *
  * \include test/snippet/argument_parser/validators_input_file.cpp
  *
+ * The valid extensions can also be obtained from a seqan3 formatted file type, e.g. seqan3::sequence_input_file, if
+ * it is given as template argument to this class. The following snippet demonstrates the different ways to instantiate
+ * the seqan3::output_file_validator.
+ *
+ * \include test/snippet/argument_parser/validators_input_file_ext_from_file.cpp
+ *
  * \note The validator works on every type that can be implicitly converted to std::filesystem::path.
  */
+template <typename file_t = void>
 class input_file_validator : public file_validator_base
 {
 public:
+
+    static_assert(std::Same<file_t, void> || detail::has_type_valid_formats<file_t>,
+                  "Expected either a template type with a static member called valid_formats (a file type) or void.");
+
     // Import from base class.
-    using file_validator_base::value_type;
+    using typename file_validator_base::value_type;
 
     /*!\name Constructors, destructor and assignment
      * \{
      */
-    input_file_validator() = default;                                         //!< Defaulted.
+
+    /*!\brief Default constructor.
+     *
+     * \details
+     *
+     * If the class' template argument `file_t` names a valid seqan3 file type that contains a
+     * static member `valid_formats`, e.g. seqan3::sequence_input_file::valid_formats, then it generates the
+     * list of valid extensions from this file. Otherwise the extensions list is empty.
+     */
+    input_file_validator()
+    {
+        if constexpr (!std::Same<file_t, void>)
+            file_validator_base::extensions = detail::valid_file_extensions<typename file_t::valid_formats>();
+    }
+
     input_file_validator(input_file_validator const &) = default;             //!< Defaulted.
     input_file_validator(input_file_validator &&) = default;                  //!< Defaulted.
     input_file_validator & operator=(input_file_validator const &) = default; //!< Defaulted.
     input_file_validator & operator=(input_file_validator &&) = default;      //!< Defaulted.
     virtual ~input_file_validator() = default;                                //!< Virtual destructor.
+
+    /*!\brief Constructs from a given collection of valid extensions.
+     * \param[in] extensions The valid extensions to validate for.
+     *
+     * \details
+     *
+     * This constructor is only available if `file_t` does not name a valid seqan3 file type that contains a
+     * static member `valid_formats`.
+     */
+    explicit input_file_validator(std::vector<std::string> extensions)
+    //!\cond
+        requires std::Same<file_t, void>
+    //!\endcond
+        : file_validator_base{}
+    {
+        file_validator_base::extensions = std::move(extensions);
+    }
 
     // Import base class constructor.
     using file_validator_base::file_validator_base;
@@ -450,15 +492,16 @@ public:
     //!\brief Returns a message that can be appended to the (positional) options help page info.
     std::string get_help_page_message() const
     {
-        return detail::to_string("Valid input file formats: ",
-                                 file_validator_base::extensions | std::view::join(std::string{", "}),
-                                 ".");
+        return detail::to_string("Valid input file formats: [",
+                                 extensions | std::view::join(std::string{", "}),
+                                 "]");
     }
 };
 
 /*!\brief A validator that checks if a given path is a valid output file.
  * \ingroup argument_parser
  * \implements seqan3::Validator
+ * \tparam file_t The type of the file to get the valid extensions for; `void` on default.
  *
  * \details
  *
@@ -469,23 +512,51 @@ public:
  *
  * \include test/snippet/argument_parser/validators_output_file.cpp
  *
+ * The valid extensions can also be obtained from a seqan3 formatted file type, e.g. seqan3::sequence_input_file, if
+ * it is given as template argument to this class. The following snippet demonstrates the different ways to instantiate
+ * the seqan3::output_file_validator.
+ *
+ * \include test/snippet/argument_parser/validators_output_file_ext_from_file.cpp
+ *
  * \note The validator works on every type that can be implicitly converted to std::filesystem::path.
  */
+template <typename file_t = void>
 class output_file_validator : public file_validator_base
 {
 public:
+
+    static_assert(std::Same<file_t, void> || detail::has_type_valid_formats<file_t>,
+                "Expected either a template type with a static member called valid_formats (a file type) or void.");
+
     // Import from base class.
-    using file_validator_base::value_type;
+    using typename file_validator_base::value_type;
 
     /*!\name Constructors, destructor and assignment
      * \{
      */
-    output_file_validator() = default;                                            //!< Defaulted.
+
+    //!\copydoc seqan3::input_file_validator::input_file_validator()
+    output_file_validator()
+    {
+        if constexpr (!std::Same<file_t, void>)
+            file_validator_base::extensions = detail::valid_file_extensions<typename file_t::valid_formats>();
+    }
+
     output_file_validator(output_file_validator const &) = default;               //!< Defaulted.
     output_file_validator(output_file_validator &&) = default;                    //!< Defaulted.
     output_file_validator & operator=(output_file_validator const &) = default;   //!< Defaulted.
     output_file_validator & operator=(output_file_validator &&) = default;        //!< Defaulted.
     virtual ~output_file_validator() = default;                                   //!< Virtual Destructor.
+
+    //!\copydoc seqan3::input_file_validator::input_file_validator(std::vector<std::string>)
+    explicit output_file_validator(std::vector<std::string> extensions)
+    //!\cond
+        requires std::Same<file_t, void>
+    //!\endcond
+        : file_validator_base{}
+    {
+        file_validator_base::extensions = std::move(extensions);
+    }
 
     // Import base constructor.
     using file_validator_base::file_validator_base;
@@ -524,8 +595,9 @@ public:
     //!\brief Returns a message that can be appended to the (positional) options help page info.
     std::string get_help_page_message() const
     {
-        return detail::to_string("Valid output file formats: ",
-                                 file_validator_base::extensions | std::view::join(std::string{", "}), ".");
+        return detail::to_string("Valid output file formats: [",
+                                 extensions | std::view::join(std::string{", "}),
+                                 "]");
     }
 };
 
@@ -547,7 +619,7 @@ class input_directory_validator : public file_validator_base
 {
 public:
     // Import from base class.
-    using file_validator_base::value_type;
+    using typename file_validator_base::value_type;
 
     /*!\name Constructors, destructor and assignment
      * \{
@@ -619,7 +691,7 @@ class output_directory_validator : public file_validator_base
 {
 public:
     // Imported from base class.
-    using file_validator_base::value_type;
+    using typename file_validator_base::value_type;
 
     /*!\name Constructors, destructor and assignment
      * \{
--- a/include/seqan3/io/detail/magic_header.hpp
+++ b/include/seqan3/io/detail/magic_header.hpp
@@ -14,82 +14,94 @@
 
 #include <array>
 
+#include <seqan3/core/type_traits/basic.hpp>
+#include <seqan3/core/type_traits/template_inspection.hpp>
 #include <seqan3/core/platform.hpp>
+#include <seqan3/core/type_list/type_list.hpp>
+#include <seqan3/core/type_list/traits.hpp>
+#include <seqan3/std/type_traits>
 
 namespace seqan3::detail
 {
 
-//!\brief Defines a magic byte sequence to disambiguate different compression formats. Default is empty.
-//!\ingroup io
-template <typename header_tag_t>
-inline constexpr std::array<char, 0> magic_header{};
-
 //!\brief A tag signifying a gz compressed file.
 //!\ingroup io
 struct gz_compression
-{};
+{
+    //!\brief The valid file extension for gz compression.
+    static inline std::vector<std::string> file_extensions
+    {
+        {"gz"}
+    };
 
-/*!\brief The magic byte sequence to disambiguate gz compressed files.
- * \ingroup io
- *
- * \details
- *
- * Specialises seqan3::detail::magic_header for seqan3::detail::gz_compression.
- */
-template <>
-inline constexpr std::array<char, 3> magic_header<gz_compression>{'\x1f', '\x8b', '\x08'};
+    //!\brief The magic byte sequence to disambiguate gz compressed files.
+    static constexpr std::array<char, 3> magic_header{'\x1f', '\x8b', '\x08'};
+};
 
 //!\brief A tag signifying a bz2 compressed file.
 //!\ingroup io
 struct bz2_compression
-{};
+{
+    //!\brief The valid file extension for bz2 compression.
+    static inline std::vector<std::string> file_extensions
+    {
+        {"bz2"}
+    };
 
-/*!\brief The magic byte sequence to disambiguate bz2 compressed files.
- * \ingroup io
- *
- * \details
- *
- * Specialises seqan3::detail::magic_header for seqan3::detail::bz2_compression.
- */
-template <>
-inline constexpr std::array<char, 3> magic_header<bz2_compression>{'\x42', '\x5a', '\x68'};
+    //!\brief The magic byte sequence to disambiguate bz2 compressed files.
+    static constexpr std::array<char, 3> magic_header{'\x42', '\x5a', '\x68'};
+};
 
 //!\brief A tag signifying a zstd compressed file.
 //!\ingroup io
 struct zstd_compression
-{};
+{
+    //!\brief The valid file extension for zstd compression.
+    static inline std::vector<std::string> file_extensions
+    {
+        {"zst"}
+    };
 
-/*!\brief The magic byte sequence to disambiguate zstd compressed files.
- * \ingroup io
- *
- * \details
- *
- * Specialises seqan3::detail::magic_header for seqan3::detail::zstd_compression.
- */
-template <>
-inline constexpr std::array<char, 4> magic_header<zstd_compression>{'\x28', '\xb5', '\x2f', '\xfd'};
+    //!\brief The magic byte sequence to disambiguate zstd compressed files.
+    static constexpr std::array<char, 4> magic_header{'\x28', '\xb5', '\x2f', '\xfd'};
+};
 
 //!\brief A tag signifying a bgzf compressed file.
 //!\ingroup io
 struct bgzf_compression
-{};
-
-/*!\brief The magic byte sequence to disambiguate bgzf compressed files.
- * \ingroup io
- *
- * \details
- *
- * Specialises seqan3::detail::magic_header for seqan3::detail::bgzf_compression.
- */
-template <>
-inline constexpr std::array<char, 18> magic_header<bgzf_compression>
 {
-//  ID1                              ID2                              CM
-    magic_header<gz_compression>[0], magic_header<gz_compression>[1], magic_header<gz_compression>[2],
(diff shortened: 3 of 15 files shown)


include/seqan3/io/sequence_file/input.hpp
```diff
@@ -691,6 +692,20 @@ class sequence_file_input
     }
     //!\}
 
+    //!\brief Receives a list of valid file extensions.
```
smehringer:
```suggestion
    //!\brief Returns a list of valid file extensions.
```

include/seqan3/io/sequence_file/input.hpp
```diff
@@ -691,6 +692,20 @@ class sequence_file_input
...
+    //!\returns std::vector over std::string with all valid file extensions specified by `valid_formats`.
+    static std::vector<std::string> valid_file_extensions()
+    {
+        std::vector<std::string> extensions;
+        detail::for_each_type([&extensions](auto t_identity)
```
smehringer:
```suggestion
        detail::for_each_type([&extensions] (auto t_identity)
```
:nail_care:

include/seqan3/io/sequence_file/input.hpp
```diff
@@ -691,6 +692,20 @@ class sequence_file_input
...
+        {
+            using format_t = typename decltype(t_identity)::type;
+            for (std::string ext : format_t::file_extensions)
```
smehringer: `std::ranges::copy(format_t::file_extensions, std::back_inserter(extensions))`
?

include/seqan3/io/sequence_file/input.hpp
```diff
@@ -800,6 +815,28 @@ class sequence_file_input
...
  * \{
  */
+
+//!\brief Deduces the sequence input file type from the stream and the format.
+template <IStream2                           stream_type,
+          SequenceFileInputFormat            file_format>
+sequence_file_input(stream_type & stream,
+                    file_format const &)
+    -> sequence_file_input<typename sequence_file_input<>::traits_type,       // actually use the default
+                           fields<field::SEQ, field::ID, field::QUAL>,        // default field ids.
```
smehringer:
```suggestion
                           typename sequence_file_input<>::selected_field_ids, // actually use the default
```

check the other ones below too

include/seqan3/io/sequence_file/input.hpp
```diff
@@ -800,6 +815,28 @@ class sequence_file_input
...
+                           type_list<file_format>,
+                           typename std::remove_reference_t<stream_type>::char_type>;
+
+//!\brief Deduces the sequence input file type from the stream and the format.
+template <IStream2                           stream_type,
+          SequenceFileInputFormat            file_format>
+sequence_file_input(stream_type && stream,
+                   file_format const &)
```
smehringer:
```suggestion
                    file_format const &)
```
:nail_care:  :grin:

test/unit/io/sequence_file/sequence_file_input_test.cpp
```diff
@@ -507,3 +501,36 @@ TEST_F(sequence_file_input_f, read_empty_bz2_file)
...
+        auto all_extensions = sequence_file_input<>::valid_file_extensions();
+
+        // define testing lambda
+        auto cmp_lambda = [&all_extensions](auto & source)
+        {
+            return std::find(all_extensions.begin(), all_extensions.end(), source);
+        };
+
+        // Test fasta extensions
+        for (std::string ext : sequence_file_format_fasta::file_extensions)
```
smehringer:
```suggestion
        for (std::string & ext : sequence_file_format_fasta::file_extensions)
```

test/unit/io/sequence_file/sequence_file_output_test.cpp
```diff
@@ -202,6 +196,39 @@ TEST(general, default_template_args_and_deduction_guides)
...
+        auto all_extensions = sequence_file_output<>::valid_file_extensions();
+
+        // define testing lambda
+        auto cmp_lambda = [&all_extensions](auto & source)
+        {
+            return std::find(all_extensions.begin(), all_extensions.end(), source);
+        };
+
+        // Test fasta extensions
+        for (std::string ext : sequence_file_format_fasta::file_extensions)
```
smehringer:
```suggestion
        for (std::string & ext : sequence_file_format_fasta::file_extensions)
```
also for all the other ones

test/unit/io/sequence_file/sequence_file_output_test.cpp
```diff
@@ -202,6 +196,39 @@ TEST(general, default_template_args_and_deduction_guides)
...
+        EXPECT_TRUE(std::ranges::equal(decltype(fout)::valid_file_extensions(),
+                                       sequence_file_format_fasta::file_extensions));
+    }
+
+    {  // All formats
+        // get all extensions.
+        auto all_extensions = sequence_file_output<>::valid_file_extensions();
+
+        // define testing lambda
+        auto cmp_lambda = [&all_extensions](auto & source)
```
smehringer:
```suggestion
        auto cmp_lambda = [&all_extensions] (auto & source)
```

test/unit/io/sequence_file/sequence_file_input_test.cpp
```diff
@@ -507,3 +501,36 @@ TEST_F(sequence_file_input_f, read_empty_bz2_file)
...
+        EXPECT_TRUE(std::ranges::equal(decltype(fin)::valid_file_extensions(),
+                                       sequence_file_format_fasta::file_extensions));
+    }
+
+    {  // All formats
+        // get all extensions.
+        auto all_extensions = sequence_file_input<>::valid_file_extensions();
+
+        // define testing lambda
+        auto cmp_lambda = [&all_extensions](auto & source)
```
smehringer:
```suggestion
        auto cmp_lambda = [&all_extensions] (auto & source)
```

include/seqan3/io/sequence_file/input.hpp
smehringer: This was not resolved?

h-2:
What's the status here?

smehringer:
my request is still not resolved :point_up:

rrahn:
🦏 ...

rrahn:
Somehow I have overwritten the submitted changes from your review and by that undo the changes. Please check if now everything is as expected.

include/seqan3/io/sequence_file/input.hpp
```diff
@@ -691,6 +692,19 @@ class sequence_file_input
     }
     //!\}
 
+    //!\brief Returns a list of valid file extensions.
+    //!\returns std::vector over std::string with all valid file extensions specified by `valid_formats`.
+    static std::vector<std::string> valid_file_extensions()
+    {
+        std::vector<std::string> extensions;
```
smehringer: we will need this code in each of the files (6 in total currently) so to reduce code it would be great if this can be outsourced into a function in `io/detail/misc` :pray:

rrahn: I already proposed a small mixin class that adds this feature. But @h-2 was proposing to keep it simple. I am in general in favour for DRYing this out and don't see any points against it.

h-2: Well, I think we should sit down and think about what else we can DRY about the files and do it in one step altogether; but if you really want to factor it out now, I would prefer not to do a CRTP-mixin, but rather just a free function in misc that is called by the member function.
CRTP-mixins have the drawback that they are difficult to document and I would rather hide the implementation details from the user, the class is complicated enough.

smehringer: Why a CRTP-mixin? I was aiming all along for a free function that takes the vallid_formats and returns a list of extensions.

rrahn: > I would prefer not to do a CRTP-mixin, but rather just a free function in misc that is called by the member function.

The interface still needs to be repeated in each file and still there can go a lot of  things wrong (e.g. diverging documentation and probably other subtle errors). And even if the interface is very simple not much can go wrong it suggests a wrong example that could be copied in other places for more complex problems. 

> CRTP-mixins have the drawback that they are difficult to document and I would rather hide the implementation details from the user, the class is complicated enough.

I think we don't need a full CRTP-mixin. Something that just get's the valid_formats type would be sufficient. I think this can be cleanly documented and it does not need to be detail stuff. Regarding further refactorings. We should discuss it, but given the huge amount of tasks and the few weeks we have left, I wouldn't go for something big now.

h-2: I am also not in favour of discussing this too long, I would propose the following free functions in misc
```cpp
template <typename ... format_ts>
std::vector<std::string> valid_file_format_extensions()
{
    // impl
}

inline std::vector<std::string> valid_file_compression_extensions()
{
   // impl based on macros
#ifdef SEQAN3_HAS_GRZIP 
}
```

And then just use the free functions in the argParse and display something like
```
--input   A file with the input sequences. Valid format extensions are [fasta, fa, fastq, fq], possibly combined with [gz, bz2, bgzf].
```

rrahn:
I have now revised everything to use a global detail function and also added a compression_formats type to determine the compression.

include/seqan3/io/detail/misc.hpp
```diff
@@ -96,4 +97,65 @@ void set_format(format_variant_type & format,
...
+ *
+ * thread-safe. Only reads from shared memory.
+ *
+ * ### Exception
+ *
+ * Strong exception guarantee. No input is modified. Might throw std::bad_alloc.
+ */
+template <typename formats_t>
+//!\cond
+    requires (all_formats_have_file_extensions(formats_t{}))
```
h-2: Do you need this? Aren't the formats already required by concept to have the extensions static member?

include/seqan3/io/sequence_file/input.hpp
h-2: not resolved

include/seqan3/io/sequence_file/input.hpp
```diff
@@ -819,7 +820,27 @@ class sequence_file_input
...
+template <IStream2                           stream_type,
+          SequenceFileInputFormat            file_format>
+sequence_file_input(stream_type && stream,
+                   file_format const &)
+   -> sequence_file_input<typename sequence_file_input<>::traits_type,         // actually use the default
+                          typename sequence_file_input<>::selected_field_ids,  // default field ids.
+                          type_list<file_format>,
+                          typename std::remove_reference_t<stream_type>::char_type>;
+
+//!\brief Deduces the sequence input file type from the stream, the format and the field ids.
```
h-2: Why are new overloads needed?

test/unit/io/sequence_file/sequence_file_input_test.cpp
```diff
@@ -186,11 +186,7 @@ TEST_F(sequence_file_input_f, default_template_args_and_deduction_guides)
         using t = decltype(fin);
         EXPECT_TRUE((std::is_same_v<typename t::traits_type,        comp0>));
         EXPECT_TRUE((std::is_same_v<typename t::selected_field_ids, comp1>));
-        EXPECT_TRUE((std::is_same_v<typename t::valid_formats,      type_list<format_embl,
-                                                                              format_fasta,
-                                                                              format_fastq,
-                                                                              format_genbank,
-                                                                              format_sam>>));                // changed
+        EXPECT_TRUE((std::is_same_v<typename t::valid_formats,      type_list<format_fasta>>));// changed
```
h-2: :+1: This is why the other deduction is necessary?

include/seqan3/io/detail/magic_header.hpp
```diff
@@ -72,7 +93,13 @@ inline constexpr std::array<int8_t, 4> magic_header<zstd_compression>{'\x28', '\
...
 //!\ingroup io
 struct bgzf_compression
-{};
+{
+    //!\brief The valid file extension for bgzf compression.
+    static inline const std::vector<std::string> file_extensions
+    {
+        {"bgzf"}
+    };
+};
```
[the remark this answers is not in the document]
rrahn: well, I am not yet convinced of how that could benefit. Like I can add my own format and say this is also gzipped? Not sure that we want that for compressions. I would at least put that on ice if this is really ever gonna requested.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -412,6 +413,20 @@ class input_file_validator : public file_validator_base
...
     virtual ~input_file_validator() = default;                                //!< Virtual destructor.
 
+    /*!\brief Constructs from a given file.
+     * \tparam    file_t The type of the file; must have a member type called `valid_formats` containing the formats.
+     * \param[in] file   The file to get the valid extensions for.
+     *
+     * \details
+     *
+     * Uses the extensions from the formats of the file to initialise the valid extensions.
+     */
```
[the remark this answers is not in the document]
rrahn: yes, I'll do

test/unit/io/sequence_file/sequence_file_input_test.cpp
rrahn: I stumbled over this issue with the missing deduction guide when fixing this. But it was a long time ago so I guess yes that's why we need the deduction guide. 😏

include/seqan3/io/sequence_file/input.hpp
rrahn: Because the test was testing wrong behavior and this guide was necessary if only the stream and the format is given.

include/seqan3/io/detail/misc.hpp
rrahn: Do we have a concept for the formats? I thought not but I'll have a second look or you point me there?

include/seqan3/io/detail/magic_header.hpp
```diff
@@ -92,4 +119,21 @@ inline constexpr std::array<int8_t, 18> magic_header<bgzf_compression>
...
+        std::tuple{gz_compression{}, bgzf_compression{}},
+        #endif // SEQAN3_HAS_ZLIB
+        #if SEQAN3_HAS_BZIP2
+        std::tuple{bz2_compression{}},
+        #endif // SEQAN3_HAS_BZIP2
+        #if SEQAN3_HAS_ZSTD
+        std::tuple{zstd_compression{}},
+        #endif // SEQAN3_HAS_BZIP2
+        std::tuple{}
+    )), type_list>;
```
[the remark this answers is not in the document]
rrahn: Well, I was really trying to get something working but in the end if you have this macro stuff  it is not trivial to figure out if the previous macro was enabled or not. This allows me to specify any order and always get rid of the was there a comma or not issue. 

Beautiful right 😆

include/seqan3/io/detail/magic_header.hpp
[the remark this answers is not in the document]
rrahn: > While you are at it, why not move the magic header variable into the tag? I think extensions and magic header are similarly "connected" to the header.

That's a good point. And it looks nicer.

include/seqan3/io/detail/misc.hpp
rrahn: Ah ok, now I understand what you mean. Well, I thought since it is now a free function which could be potentially used otherwise it is good to check if the stuff added is what we want. But since the PR is quite old I did not checked this via a static assert. But I would keep it for this reasons but make it a static assert instead

include/seqan3/argument_parser/validators.hpp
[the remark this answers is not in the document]
rrahn: well, if you consider it they are not usable at all. You would need to create a file but obviously you don't have a path yet since nothing was parsed yet. So the only way seems to to provide the file information as template argument.

include/seqan3/io/detail/magic_header.hpp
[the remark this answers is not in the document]
h-2: It's ugly, but oh well. Can you add a note that this should be replaced once the proper mechanics are available?

include/seqan3/io/detail/misc.hpp
```diff
@@ -96,4 +97,67 @@ void set_format(format_variant_type & format,
...
+ * \tparam    args_t...   A template parameter pack over the expected formats.
+ * \param[in] type_list   The type list containing the expected formats.
+ * \returns `true` if all formats are valid, otherwise `false`.
+ */
+template <template <typename ...> typename type_list_t, typename ... args_t>
+constexpr bool all_formats_have_file_extensions(type_list_t<args_t...> const & SEQAN3_DOXYGEN_ONLY(type_list))
+{
+    return (format_has_extension<args_t>::value && ... && true);
+}
+
```
h-2: So much boilerplate :cry:

include/seqan3/io/detail/misc_input.hpp
```diff
@@ -99,44 +99,57 @@ inline auto make_secondary_istream(std::basic_istream<char_t> & primary_stream,
     for (size_t i = 0 ; i < read_chars; ++i)
         primary_stream.unget();
 
-    std::string const extension = filename.extension().string();
+    std::string extension{};
+    if (filename.has_extension())
+        extension = filename.extension().string().substr(1);
 
     // set return value appropriately
     if (read_chars == magic_number.size() && contrib::_bgzfCheckHeader(magic_number.data())) // BGZF
```
h-2: Is this now valid when Gzip is not available? I remember we had an issue on that.

include/seqan3/io/detail/misc_input.hpp
```diff
@@ -99,44 +99,57 @@ inline auto make_secondary_istream(std::basic_istream<char_t> & primary_stream,
...
+    else if (starts_with(magic_number, gz_compression::magic_header)) // GZIP
     {
     #ifdef SEQAN3_HAS_ZLIB
-        if ((extension == ".gz") || (extension == ".bgzf"))
-            filename.replace_extension();
-
+        if ((std::ranges::find(gz_compression::file_extensions, extension) !=
+             std::ranges::end(gz_compression::file_extensions)) ||
+            (std::ranges::find(bgzf_compression::file_extensions, extension) !=
+             std::ranges::end(bgzf_compression::file_extensions)))
```
h-2: Should we maybe add a `contains()` free function? :see_no_evil:

include/seqan3/io/detail/misc_input.hpp
```diff
@@ -99,44 +99,57 @@ inline auto make_secondary_istream(std::basic_istream<char_t> & primary_stream,
...
+        {
+            (filename.replace_extension());
```
h-2: Why the parens?

include/seqan3/io/detail/magic_header.hpp
```diff
@@ -13,83 +13,94 @@
...
-template <typename header_tag_t>
-inline constexpr std::array<char, 0> magic_header{};
-
 //!\brief A tag signifying a gz compressed file.
 //!\ingroup io
 struct gz_compression
-{};
+{
+    //!\brief The valid file extension for gz compression.
+    static inline const std::vector<std::string> file_extensions
```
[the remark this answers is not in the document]
rrahn: This is the question. Do you want to provide another extension for the compressed formats? At the  moment I don't see little use for this. I think one can change this later without API breakage if it turns out to be needed after all.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -432,10 +481,10 @@ class input_file_validator : public file_validator_base
                 throw parser_invalid_argument(detail::to_string("The file ", file, " does not exist!"));
 
             // Check if file is regular and can be opened for reading.
-            validate_readability(file);
+            this->validate_readability(file);
 
             // Check extension.
-            validate_filename(file);
+            this->validate_filename(file);
```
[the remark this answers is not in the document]
rrahn: because, the base class is now a template class. 😒

include/seqan3/argument_parser/validators.hpp
```diff
@@ -236,12 +237,17 @@ value_list_validator(std::initializer_list<const char *>) -> value_list_validato
...
  *
  * \details
  *
  * This class provides a common interface for seqan3::input_file_validator and the seqan3::output_file_validator as
  * well as the seqan3::input_directory_validator and seqan3::output_directory_validator.
+ *
+ * The type can be further specialised for the seqan3::input_file_validator and the seqan3::output_file_validator
+ * using the template argument to determine the valid extensions from the given file type.
  */
+template <typename file_t = void>
```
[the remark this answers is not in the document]
rrahn: Because a constructor cannot have an explicit template argument. Thus you would need to pass in an instance of a file. But since it is not default constructible you need to feed it with a stream or file. Yet, it is a weird circular dependency as you need to parse the arguments before you know the filename.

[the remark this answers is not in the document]
rrahn: Note, that these changes have no implication on the current version, due to type deduction guides. That means, you can provide the file type as a template argument or you call the constructors like before, without any template argument.

include/seqan3/io/detail/magic_header.hpp
```diff
@@ -13,83 +13,94 @@
...
 
 //!\brief A tag signifying a bgzf compressed file.
 //!\ingroup io
 struct bgzf_compression
-{};
+{
+    //!\brief The valid file extension for bgzf compression.
+    static inline const std::vector<std::string> file_extensions
+    {
+        {"bgzf"}
```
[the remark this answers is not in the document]
rrahn: For output we write bgzf format if we detect .gz extension. But the extension is not replaced. Thus there is no reason to put the gz extension here as well. The same goes for gz format which could also read bgzf files since they are compatible.

include/seqan3/io/detail/magic_header.hpp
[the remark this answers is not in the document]
rrahn: I will try to give it 10 minutes more love and see if I make some meta programming interface for this. But for now I would not spend to much time on it.

include/seqan3/io/detail/misc_input.hpp
rrahn: There was an issue with some functionality not available if Gzip was not found. I'll fix this in another PR this or next week

include/seqan3/io/detail/misc_input.hpp
rrahn: not a free function but I'll dry it into a lambda

include/seqan3/io/detail/misc_input.hpp
rrahn: i don't know ^^

include/seqan3/argument_parser/validators.hpp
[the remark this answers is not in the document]
rrahn: Yes, I can do that but than have to copy the constructors for the output file. This kind of diminish the use of a base class to DRY some code. Just because I need to use this to call the correct functions?

include/seqan3/argument_parser/validators.hpp
```diff
@@ -450,15 +499,27 @@ class input_file_validator : public file_validator_base
...
+/*!\name Type deduction guides
+ * \relates seqan3::input_file_validator
+ * \{
+ */
+//!\brief Default constructor deduces to default template argument.
+input_file_validator() -> input_file_validator<>;
+
+//!\brief Construction from extension collection deduces to the default template argument.
+input_file_validator(std::vector<std::string>) -> input_file_validator<>;
+//!\}
```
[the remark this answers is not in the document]
rrahn: Mhmm, haven't tried without (at least for the first one this should be true). But for the second I guess he tries to infer some template argument. I'll check

include/seqan3/argument_parser/validators.hpp
```diff
@@ -450,15 +499,27 @@ class input_file_validator : public file_validator_base
...
+
 /*!\brief A validator that checks if a given path is a valid output file.
  * \ingroup argument_parser
  * \implements seqan3::Validator
+ * \tparam file_t The type of the file to get the valid extensions for; `void` on default.
```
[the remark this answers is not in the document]
rrahn: Well it all would go away if files could be default constructible.  But I agree that this does also make not a lot of sense.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -469,30 +530,41 @@ class input_file_validator : public file_validator_base
...
+    //!\copydoc seqan3::file_validator_base::file_validator_base()
+    output_file_validator() = default;
     output_file_validator(output_file_validator const &) = default;               //!< Defaulted.
     output_file_validator(output_file_validator &&) = default;                    //!< Defaulted.
     output_file_validator & operator=(output_file_validator const &) = default;   //!< Defaulted.
     output_file_validator & operator=(output_file_validator &&) = default;        //!< Defaulted.
     virtual ~output_file_validator() = default;                                   //!< Virtual Destructor.
 
+    //!\copydoc seqan3::file_validator_base::file_validator_base(std::vector<std::string>)
+    SEQAN3_DOXYGEN_ONLY(output_file_validator(std::vector<std::string> extensions){})
```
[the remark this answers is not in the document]
rrahn: But than you have ugly copy'n'paste code. But I am wondering why have this in the first place. The base is not a crtp and the ctors are public. I'll check it again

include/seqan3/io/detail/magic_header.hpp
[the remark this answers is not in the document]
rrahn: Ok, I remove the const, also I am less convinced that this is a good idea

include/seqan3/io/detail/magic_header.hpp
```diff
@@ -13,83 +13,95 @@
...
- * \ingroup io
- *
- * \details
- *
- * Specialises seqan3::detail::magic_header for seqan3::detail::gz_compression.
- */
-template <>
-inline constexpr std::array<char, 3> magic_header<gz_compression>{'\x1f', '\x8b', '\x08'};
+    //!\brief The magic byte sequence to disambiguate gz compressed files.
+    static inline constexpr std::array<char, 3> magic_header{'\x1f', '\x8b', '\x08'};
```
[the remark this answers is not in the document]
rrahn: that's true

include/seqan3/io/detail/magic_header.hpp
```diff
@@ -13,83 +13,95 @@
...
+                                                    #if SEQAN3_HAS_ZLIB
+                                                        ,gz_compression
+                                                        ,bgzf_compression
+                                                    #endif // SEQAN3_HAS_ZLIB
+                                                    #if SEQAN3_HAS_BZIP2
+                                                        ,bz2_compression
+                                                    #endif // SEQAN3_HAS_BZIP2
+                                                    #if SEQAN3_HAS_ZSTD
+                                                        ,zstd_compression
+                                                    #endif // SEQAN3_HAS_ZSTD
```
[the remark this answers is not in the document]
rrahn: but it looked so much more ugly with commas

rrahn:
@h-2 I think I addressed all issues now.

include/seqan3/io/detail/magic_header.hpp
```diff
@@ -13,83 +13,95 @@
 #pragma once
 
 #include <array>
+#include <tuple>
```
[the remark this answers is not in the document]
h-2: still there :nail_care:

rrahn:
Oh yes, good point with the changelog. 👍 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

CHANGELOG.md
```diff
@@ -25,9 +25,15 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 
 ## API changes
 
-### The `type_list` header has moved
+### Argument parser
```
h-2: Do we want to group things by module? I don't mind, but we should then do it for features, bugs and API changes, right?

h-2: This gets very nested then altogether. Also, some changes are not module specific, like renaming of Concepts that is going to happen :thinking:

Maybe we can just mention the module in the name of the entry and sort by module?

CHANGELOG.md
```diff
@@ -25,9 +25,15 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
...
 
-If you included `<seqan3/core/type_list.hpp>` you need to change the path to `<seqan3/core/type_list/type_list.hpp>`.
+* The signature of input file and output file validator have been modified to template classes.
+  It is now possible to get the list of file extensions from the type of a seqan input file, respectively output file. Old code will likely not be affected unless the type was used explicitly.
```
h-2: This is mixing up the description of a feature and an API Change. Also `Old code will likely not be affected` does not help people in actually fixing it, if it is affected. I think the API Changes section should provide at most minimal context and specify exactly what to do to make code compile.

I would propose to write:
```md

## New features

### Argument Parser: file validators accept files as template arguments

Passing a file as template argument to a file validator, e.g. `seqan3::input_file_validator<seqan3::sequence_file_input<>> validator{}` will automatically set all supported extensions.

[...]

## API Changes

### Argument Parser: file validators are now class templates

Most code should be unaffected; places that are affected need to explicitly provide empty template arguments, e.g. `seqan3::input_file_validator<>` instead of `seqan3::input_file_validator`.
```

CHANGELOG.md
```diff
@@ -25,9 +25,15 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
...
+
+### Core
+
+* The `type_list` header has moved
+  If you included `<seqan3/core/type_list.hpp>` you need to change the path to `<seqan3/core/type_list/type_list.hpp>`.
```
h-2: This will be joined into one line with the previous line in our doxygen documentation. I would prefer to keep  the `###` per item so that we have enough space to describe a feature or API workaround.

CHANGELOG.md
rrahn: In general the information should be grouped. The most natural grouping would be by module. Then people can review by module whether or not the changes affect their code.

CHANGELOG.md
rrahn: But no one is looking at our changelog through doxygen. I propose highlighting the begin of the point with bold and adding a colon. This will be rendered properly in markdown and doxygen too.

rrahn:
@h-2 ping

PR #864, opened 2019-03-29
adaptor apocalypse

h-2:
Rework of the adaptor machinery:
  * it is now possible to store two adaptor closure objects into a new one and use that
  * adaptor definition for adaptors that take arguments is a little more complex (not just one template), but the handling of args is also more correct now
  * empty parens allowed for proto-adaptors, e.g. `auto v = vec | view::translate()`
  * view::take_line is now expressed as a special case of view::take_until

After this PR is merged I will look at fixing the recently introduced broken drop/take.


include/seqan3/range/view/detail.hpp
```diff
@@ -322,6 +315,79 @@ class generic_pipable_view_adaptor : public pipable_adaptor_base<generic_pipable
...
+ * \ingroup view
+ *
+ * \details
+ *
+ * This template is particularly useful in combination with range adaptor objects that are not closure
+ * objects (a.k.a. proto adaptors).
+ * These objects take other parameters in addition to the range parameter that they might need store
+ * in which case they can return an object of this type with the stored arguments.
+ *
+ * Additionally this template delegates back to the proto adaptor for handling range input since
```
eseiler:
```suggestion
 * Additionally, this template delegates back to the proto adaptor for handling range input since
```

include/seqan3/range/view/take.hpp
```diff
@@ -16,6 +16,7 @@
 
 #include <seqan3/core/metafunction/iterator.hpp>
 #include <seqan3/core/metafunction/range.hpp>
+#include <seqan3/core/metafunction/template_inspection.hpp>
```
eseiler: Needed?

include/seqan3/range/view/translation.hpp
```diff
@@ -20,6 +20,7 @@
 #include <seqan3/alphabet/aminoacid/translation.hpp>
 #include <seqan3/core/add_enum_bitwise_operators.hpp>
 #include <seqan3/core/metafunction/range.hpp>
+#include <seqan3/core/metafunction/template_inspection.hpp>
```
eseiler: Needed?

include/seqan3/range/view/trim.hpp
```diff
@@ -25,113 +24,48 @@ namespace seqan3::detail
 /*!\brief The underlying type of seqan3::view::trim.
  * \ingroup view
  *
- * Under the hood this delegates to ranges::view::take_while.
+ * Under the hood this delegates to std::view::take_while.
```
eseiler:
```suggestion
 * Under the hood this delegates to std::view::take_until.
```

test/unit/range/view/view_deep_test.cpp
```diff
@@ -13,6 +13,7 @@
 #include <range/v3/view/take.hpp>
 
 #include <seqan3/alphabet/nucleotide/all.hpp>
+#include <seqan3/io/stream/debug_stream.hpp>
```
eseiler: 👀

include/seqan3/range/view/detail.hpp
```diff
@@ -13,301 +13,294 @@
...
-//  view_base
+//  forwards
 // ============================================================================
 
-/*!\brief An empty base class that our views inherit from so they are treated as views by the STL/range-v3.
- * \ingroup view
- */
-struct view_base : public ranges::view_base
-{};
+template <typename left_adaptor_t, typename right_adaptor_t>
```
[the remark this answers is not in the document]
h-2: ?

include/seqan3/range/view/detail.hpp
```diff
@@ -13,301 +13,294 @@
...
-    auto operator()(urng_t && urange, arg_types && ...args) const
+    // a default constructor is not provided, however the constructor below might be one.
+    constexpr adaptor_base(adaptor_base const &)                noexcept = default; //!< Defaulted.
+    constexpr adaptor_base(adaptor_base &&)                     noexcept = default; //!< Defaulted.
+    constexpr adaptor_base & operator=(adaptor_base const &)    noexcept = default; //!< Defaulted.
+    constexpr adaptor_base & operator=(adaptor_base &&)         noexcept = default; //!< Defaulted.
+    ~adaptor_base()                                             noexcept = default; //!< Defaulted.
+
+    //!\brief Constructor with possible arguments; becomes a default constructor for adaptors without args.
+    constexpr adaptor_base(stored_args_ts ... args) :
```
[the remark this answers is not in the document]
h-2: What do you mean?

include/seqan3/range/view/take.hpp
h-2: Yes

include/seqan3/range/view/translation.hpp
h-2: No

h-2:
The commits still need cleanup, but all major issues are resolved from my POV.

h-2:
Travis passes, Jenkins seems to be completely broken now, let's ignore it.

include/seqan3/range/view/detail.hpp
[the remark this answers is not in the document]
eseiler: `combined_adaptor` is not documented

include/seqan3/range/view/detail.hpp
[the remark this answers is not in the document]
eseiler: What I was trying to say at this very late hours was:
Does the compiler still provide a default `adaptor_base` constructor for adaptors with args?

test/unit/range/view/adaptor_base_test.cpp
```diff
@@ -0,0 +1,180 @@
...
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
+// -----------------------------------------------------------------------------------------------------
+
+#include <iostream>
+#include <memory>
+
+#include <gtest/gtest.h>
```
eseiler: newline?

test/unit/range/view/view_take_line_test.cpp
```diff
@@ -159,8 +159,7 @@ TEST(view_take_line, reverse_bug)
     EXPECT_TRUE(const_iterable_concept<decltype(v1)>);
     EXPECT_TRUE((std::ranges::OutputRange<decltype(v1), char>));
 
-    // Either of these build-fail on GCC7 and GCC8
-//     auto v2 = v1 | ranges::view::reverse;
-//     auto v2 = ranges::reverse_view{v1};
+    // No build failure, but wrong results:
```
eseiler: 👀 What's the result?

include/seqan3/range/view/detail.hpp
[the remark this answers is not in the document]
h-2: No, but it's not supposed to. That's what I tried to say with the comment a few lines further up.

include/seqan3/range/view/detail.hpp
```diff
@@ -13,301 +13,296 @@
...
- * A full view implementation consists of three entities:
+ * ## "Direct derivation"
  *
- *   1. the actual view, e.g. `view_foo`;
- *   2. an adaptor that returns instances of 1. and enables usage in pipes, e.g. `foo_fn`
- *   3. an instance of the adaptor, e.g. `view::foo` that is usable by consumers of your view
+ * From include/seqan3/range/view/persist.hpp:
  *
- * For more details, see the HowTo on writing views in the SeqAn3 wiki: https://github.com/seqan/seqan3/wiki
+ * \snippet include/seqan3/range/view/persist.hpp adaptor_def
```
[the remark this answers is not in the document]
h-2: Yes, I also don't know why and it's one of the reasons I propose we put all of the code-links into back-ticks.

include/seqan3/range/view/detail.hpp
```diff
@@ -13,301 +13,296 @@
...
+ * This adaptor directly derives from adaptor_base so that it can decide between
+ * actually calling the respective view's constructor and delegating to a different view
+ * (in this case `std::view::all`) based on the type of input.
  *
- * This CRTP base class can be used to simplify the implementation of the adaptor type. For most use cases it is
- * sufficient to just use the specialisation seqan3::detail::generic_pipable_view_adaptor, but other specialisations
- * like seqan3::detail::deep_view_adaptor_wrapper also make use of this base template.
+ * ## Derived templates
+ *
+ * See `seqan3::detail::combined_adaptor`, `seqan3::detail::adaptor_for_view_without_args` and
```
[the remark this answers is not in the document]
h-2: Sometimes we do this, sometimes we don't. But I think we should do it everywhere. It's more consistent, because "it is C++" and not "description text". Also it highlights broken links.

include/seqan3/range/view/detail.hpp
```diff
@@ -13,301 +13,296 @@
...
+ * A special case is `seqan3::view::deep` which is a template for range adaptor object that
+ * always wraps other range adaptor objects.
  */
-template <typename derived_type>
-class pipable_adaptor_base
+template <typename derived_type, typename ... stored_args_ts>
+class adaptor_base
 {
+private:
+    //!\brief Store the arguments.
```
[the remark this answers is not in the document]
h-2: Well it could be an instruction to the user, too: "Store the arguments in this thing!!" :laughing:

include/seqan3/range/view/detail.hpp
```diff
@@ -13,301 +13,296 @@
...
+        return static_cast<derived_type &>(*this).impl(
+            std::forward<urng_t>(urange),
+            std::tuple_element_t<Is, std::tuple<stored_args_ts...>>(std::get<Is>(arguments))...);
+    }
+
+    //!\overload
+    template <typename urng_t, size_t... Is>
+    constexpr auto explode(urng_t && urange, std::index_sequence<Is...> const &) &&
+    {
+        // move out values, because we don't need them anymore (this is temporary)
```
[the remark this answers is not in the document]
h-2: what else? :thinking:

include/seqan3/range/view/take_line.hpp
```diff
@@ -395,7 +38,7 @@ namespace seqan3::view
  * \details
  *
  * This adaptor returns a single line **excluding** the end-line character(s), *but moving the cursor behind them
- * for single-pass ranges.* I.e. for all ranges that satisfy std::ranges::ForwardRange this is the same as calling
+ * for single-pass ranges.* I.e. for all ranges that satisfy `std::ranges::ForwardRange` this is the same as calling
```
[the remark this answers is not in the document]
h-2: I copied stuff around and probably changed it. But I can revert it, since it's doc-only.

test/unit/range/view/view_take_line_test.cpp
h-2: The result is the whole underlying range reversed, instead of the sub-string. I tried to debug it for an hour, but didn't come up with a fix for range-v3. Working around the issue by using a distinct sentinel and storing end in the iterator, leads to problems with GCC. I will open a card for it.

Context: The problem is that range-v3 decrements the end iterator if possible to go to the last element, but it should only do so if the range is a CommonRange.

include/seqan3/range/view/detail.hpp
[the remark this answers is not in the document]
eseiler: Could be also read like `temporary` in the sense of "this behaviour will change"

h-2:
@rrahn Can you review this soon-ish? I have multiple other (smaller) PRs that touch the same files... thanks!

include/seqan3/range/view/detail.hpp
```diff
@@ -13,301 +13,296 @@
...
+             | std::forward<left_adaptor_t_>(left_adaptor)
+             | std::forward<right_adaptor_t_>(right_adaptor);
+    }
+
+public:
+    //!\brief Inherit the base type's constructors.
+    using base_type::base_type;
+
+    //!\brief Store both arguments in the adaptor.
+    combined_adaptor(left_adaptor_t l, right_adaptor_t r) :
```
smehringer: You copy here, why do you forward `l` and `r` in the member initialization?

include/seqan3/range/view/detail.hpp
```diff
@@ -13,301 +13,296 @@
...
 
-/*!\brief   A class template template that makes a view pipeable.
+/*!\brief Template for range adaptor closure objects that store no arguments and always delegate to the
+ *        view constructor.
+ * \tparam view_type The view type template.
  * \ingroup view
- * \tparam  view_type The view template that your wish to define the adaptor for.
  *
  * \details
  *
```
smehringer: Can you shortly state why and when you should prefer to inherit from this class instead of `adaptor_base`?

include/seqan3/range/view/detail.hpp
```diff
@@ -13,301 +13,296 @@
...
- * A full view implementation consists of three entities:
+ * # Example
  *
- *   1. the actual view, e.g. `view_foo`;
- *   2. an adaptor that returns instances of 1. and enables usage in pipes, e.g. `foo_fn`
- *   3. an instance of the adaptor, e.g. `view::foo` that is usable by consumers of your view
+ * From include/seqan3/range/view/single_pass_input.hpp:
```
smehringer:
```suggestion
 * Given a view class template that inherits from std::ranges::view_interface (include/seqan3/range/view/single_pass_input.hpp):
```

include/seqan3/range/view/detail.hpp
```diff
@@ -13,301 +13,296 @@
...
  *
- * For more details, see the HowTo on writing views in the SeqAn3 wiki: https://github.com/seqan/seqan3/wiki
+ * \snippet include/seqan3/range/view/single_pass_input.hpp view_def
  *
- * This template generates the adaptor for you:
+ * This is the signature of the view type template in namespace seqan3::detail.
```
smehringer: Ah ok the sentence here belongs to the snippet above? This is confusing because you read it as if 
`This ... :` and now comes this. It is much easier to understand if you put this sentence before the snippet.

include/seqan3/range/view/detail.hpp
```diff
@@ -322,6 +317,79 @@ class generic_pipable_view_adaptor : public pipable_adaptor_base<generic_pipable
...
+public:
+    //!\brief Inherit the base type's constructors.
+    using base_type::base_type;
+};
+
+// ============================================================================
+//  adaptor_from_functor
+// ============================================================================
+
+/*!\brief Template for range adaptor closure objects that store arguments and wrap a proto-adaptor.
```
smehringer: maybe include a proto-adaptor definition in the tutorial since you reference this class in the tutorial anyway...

so 
```
                    adaptor closure object 
                  /
adaptor object --- 
                  \
                    proto adaptor
```
?

include/seqan3/range/view/detail.hpp
```diff
@@ -322,6 +317,79 @@ class generic_pipable_view_adaptor : public pipable_adaptor_base<generic_pipable
...
+ * \tparam functor_type Type of the proto-adaptor functor.
+ * \tparam stored_args_ts Types of the stored arguments.
+ * \ingroup view
+ *
+ * \details
+ *
+ * This template is particularly useful in combination with range adaptor objects that are not closure
+ * objects (a.k.a. proto adaptors).
+ * These objects take other parameters in addition to the range parameter that they might need store
```
smehringer:
```suggestion
 * These objects take other parameters in addition to the range parameter that they might need to store
```

include/seqan3/range/view/detail.hpp
```diff
@@ -322,6 +317,79 @@ class generic_pipable_view_adaptor : public pipable_adaptor_base<generic_pipable
...
+ * in which case they can return an object of this type with the stored arguments.
+ *
+ * Additionally, this template delegates back to the proto adaptor for handling range input since
```
smehringer: I don't know what you are trying to say?

include/seqan3/range/view/detail.hpp
```diff
@@ -322,6 +317,79 @@ class generic_pipable_view_adaptor : public pipable_adaptor_base<generic_pipable
...
+ * it needs to be able to handle that in any case.
+ *
+ * # Example
+ *
+ * From include/seqan3/range/view/kmer_hash.hpp:
+ *
+ * \snippet include/seqan3/range/view/kmer_hash.hpp adaptor_def
+ *
+ * This is the full proto-adaptor, first look at the second member: it handles range **and** argument
```
smehringer:
```suggestion
 * This is the full proto-adaptor, first look at the second member function: it handles range **and** argument
```

include/seqan3/range/view/drop.hpp
```diff
@@ -30,25 +31,19 @@ namespace seqan3::detail
 
 /*!\brief View adaptor definition for view::drop and view::drop_or_throw.
  */
-class drop_fn : public pipable_adaptor_base<drop_fn>
+struct drop_fn
```
smehringer: are the constructors for aggregate types automatically constexpr?

include/seqan3/range/view/drop.hpp
```diff
@@ -30,25 +31,19 @@ namespace seqan3::detail
...
+    {
+        return detail::adaptor_from_functor{*this, drop_size};
+    }
 
     /*!\brief       Call the view's constructor with the underlying view as argument.
      * \returns     An instance of std::ranges::subrange.
      */
     template <std::ranges::ViewableRange urng_t>
-    static auto impl(urng_t && urange, size_t drop_size)
+    constexpr auto operator()(urng_t && urange, size_t drop_size) const
```
smehringer: why not `noexcept`? anything in here that might throw?

include/seqan3/range/view/drop.hpp
```diff
@@ -110,26 +105,26 @@ class drop_fn : public pipable_adaptor_base<drop_fn>
      * \returns     A std::basic_string_view over the input.
      */
     template <typename char_t, typename traits_t, typename alloc_t>
-    static std::basic_string_view<char_t, traits_t> impl(std::basic_string<char_t, traits_t, alloc_t> & urange,
-                                                         size_t drop_size)
+    std::basic_string_view<char_t, traits_t> operator()(std::basic_string<char_t, traits_t, alloc_t> & urange,
```
smehringer: `constexpr` ?

include/seqan3/range/view/take_line.hpp
```diff
@@ -427,10 +70,8 @@ namespace seqan3::view
  *
  * On single pass std::ranges::InputRange it can be used to tokenise the input stream line-wise:
  * \snippet test/snippet/range/view/take_line.cpp tokenise
- *
- * \hideinitializer
  */
-inline auto constexpr take_line = detail::take_line_fn<false>{};
+inline auto constexpr take_line = view::take_until_and_consume(is_char<'\r'> || is_char<'\n'>);
```
smehringer: Are you sure that this consumes `\r\n` fully and not just `\r` in case of windows line endings?

include/seqan3/range/view/take_until.hpp
```diff
@@ -109,7 +106,7 @@ class view_take_until : public ranges::view_interface<view_take_until<urng_t, fu
         {}
 
         //!\brief Constructor that delegates to the CRTP layer and initialises the callable.
-        iterator_type(base_base_t it, fun_ref_t _fun) :
+        iterator_type(base_base_t it, fun_ref_t _fun, sentinel_type) :
```
smehringer: why does the constructor need a tag parameter of a type that is not a template argument? :eyes: 
EDIT: 
```suggestion
        iterator_type(base_base_t it, fun_ref_t _fun, sentinel_type /*for comppatibility with iterator_type_consume_input*/) :
```

include/seqan3/range/view/take_until.hpp
```diff
@@ -170,6 +167,135 @@ class view_take_until : public ranges::view_interface<view_take_until<urng_t, fu
...
+        /*!\name Constructors, destructor and assignment
+         * \{
+         */
+        iterator_type_consume_input()                                                         = default; //!< Defaulted.
+        constexpr iterator_type_consume_input(iterator_type_consume_input const & rhs)        = default; //!< Defaulted.
+        constexpr iterator_type_consume_input(iterator_type_consume_input && rhs)             = default; //!< Defaulted.
+        //!\brief Defaulted.
+        constexpr iterator_type_consume_input & operator=(iterator_type_consume_input const & rhs) = default;
+        constexpr iterator_type_consume_input & operator=(iterator_type_consume_input && rhs) = default; //!< Defaulted.
+        ~iterator_type_consume_input() = default;
```
smehringer: missing doc

include/seqan3/range/view/take_until.hpp
```diff
@@ -170,6 +167,135 @@ class view_take_until : public ranges::view_interface<view_take_until<urng_t, fu
...
+        using pointer               = typename std::iterator_traits<base_base_t>::pointer;          //!< From base.
+        using iterator_category     = std::input_iterator_tag;                                      //!< Always input.
+        //!\}
+
+        /*!\name Arithmetic operators
+         * \brief seqan3::detail::inherited_iterator_base operators are used unless specialised here.
+         * \{
+         */
+        //!\brief Override pre-increment to implement consuming behaviour.
+        iterator_type_consume_input & operator++() noexcept(!or_throw)
```
smehringer: why is this not constexpr?

include/seqan3/range/view/take_until.hpp
```diff
@@ -170,6 +167,135 @@ class view_take_until : public ranges::view_interface<view_take_until<urng_t, fu
...
+        {
+            ++static_cast<base_base_t &>(*this);
+
+            if ((static_cast<base_base_t &>(*this) != stored_end) && fun(**this))
+            {
+                at_end_gracefully = true;
+                ++(*this); // call this function recursively
```
smehringer: Why are you calling it recursively ? If it is to fully consume `\r\n` does it also consume `\n\n`? This could be a problem for multiline parsing as I would not expect to consume more than one newline.

include/seqan3/range/view/take_until.hpp
```diff
@@ -170,6 +167,135 @@ class view_take_until : public ranges::view_interface<view_take_until<urng_t, fu
...
+
+            if (static_cast<base_base_t const &>(*this) == rhs)
+            {
+                if constexpr (or_throw)
+                    throw unexpected_end_of_input{"Reached end of input before functor evaluated to true."};
+                else
+                    return true;
+            }
+
+            return fun(**this);
```
smehringer: Is this ever true? because the moviing operations (++ etc.) always check `fun` and will set `at_end_gracefully`...

include/seqan3/range/view/take_until.hpp
```diff
@@ -207,6 +339,20 @@ class view_take_until : public ranges::view_interface<view_take_until<urng_t, fu
     view_take_until(urng_t _urange, fun_t _fun)
         : urange{std::move(_urange)}, fun{std::forward<fun_t>(_fun)}
     {}
+
+    /*!\brief Construct from another ViewableRange.
+     * \tparam rng_t      Type of the passed range; `urng_t` must be constructible from this.
+     * \param[in] _urange The underlying range.
+     * \param[in] _fun    The functor that acts as termination criterium.
+     * \throws unexpected_end_of_input If `or_throw && seqan3::sized_range_concept<urng_t>`.
```
smehringer: why only on sized ranges?

include/seqan3/range/view/take_until.hpp
```diff
@@ -308,8 +454,8 @@ class view_take_until : public ranges::view_interface<view_take_until<urng_t, fu
 
 //!\brief Type deduction guide that strips references.
 //!\relates seqan3::detail::view_take_until
-template <typename urng_t, typename fun_t, bool or_throw = false>
-view_take_until(urng_t, fun_t) -> view_take_until<std::remove_reference_t<urng_t>, fun_t, or_throw>;
+template <typename urng_t, typename fun_t, bool or_throw = false, bool and_consume = false>
+view_take_until(urng_t &&, fun_t) -> view_take_until<std::ranges::all_view<urng_t>, fun_t, or_throw, and_consume>;
```
smehringer: So now `&` is captured within `urng_t`, is this change intended?

include/seqan3/range/view/trim.hpp
```diff
@@ -25,113 +24,48 @@ namespace seqan3::detail
...
-    //!\cond
-        requires std::ranges::InputRange<irng_t> && QualityAlphabet<value_type_t<std::decay_t<irng_t>>> &&
-                 (std::is_same_v<std::decay_t<threshold_t>,
-                                 std::decay_t<value_type_t<std::decay_t<irng_t>>>> ||
-                  std::is_convertible_v<std::decay_t<threshold_t>,
-                                        underlying_phred_t<std::decay_t<value_type_t<std::decay_t<irng_t>>>>>)
-    //!\endcond
-    friend auto operator|(irng_t && irange,
-                          seqan3::detail::trim_fn::delegate<threshold_t> const & bound_view)
+    template <std::ranges::InputRange irng_t, typename threshold_t>
```
smehringer: There is no constraint on the range but you are calling `to_phred()` on its values. I would add a constraint. 
Is `to_phred` defined for all Integral types?

h-2: We said that we would enforce these constraints via static_assert now which is exactly what I am doing below.

include/seqan3/range/view/deep.hpp
```diff
@@ -97,72 +97,135 @@ namespace seqan3::view
...
      */
+    template <std::ranges::InputRange urng_t>
+    //!\cond
+        requires std::ranges::InputRange<reference_t<urng_t>>
+    //!\endcond
+    constexpr auto operator()(urng_t && urange) &
+    {
+        return std::forward<urng_t>(urange) | std::view::transform([me = *this] (auto && e)
+        {
+            return std::forward<decltype(e)>(e) | me;
```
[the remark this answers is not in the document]
smehringer: ok then I did not understand the exact behaviour.
```
template <typename t>
void foo (t && v)
{
    // within the scope of this function v behaves like an lvalue unless explicitily forwarded
    // so I thought: t != decltype(v)
    // as in t -> int && while decltype(v) -> int &
    // and that's why I need std::move(v) which casts the type back to && ... 
}
```

include/seqan3/range/view/detail.hpp
```diff
@@ -13,301 +13,296 @@
...
 {
+private:
+    //!\brief Stores the arguments.
+    std::tuple<stored_args_ts...> arguments;
+
+    //!\brief Helper function to unpack the tuple.
+    template <typename urng_t, size_t... Is>
+    constexpr auto explode(urng_t && urange, std::index_sequence<Is...> const &) &
+    {
+        // std::get returns lvalue-reference to value, but we need to copy the values
```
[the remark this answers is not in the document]
smehringer: maybe add this as a comment

include/seqan3/range/view/detail.hpp
smehringer: does `std::forward` move a copy automatically? I thought this only works if you pass the parameter by `&&`

include/seqan3/range/view/detail.hpp
smehringer: so proto adaptor was dropped? Is there another word for it? because it is confusing if only we use it now.

include/seqan3/range/view/detail.hpp
smehringer: ?

include/seqan3/range/view/deep.hpp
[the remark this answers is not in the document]
h-2: * v is always in lvalue, remember that rvalue references are lvalues, too.
* in your example 
```cpp
int i = 4;
f(i); // t == int & and decltype(v) also == int &

f(3); // t == int and decltype(v)  == int &&

// there are no other combinations not considering const
``` 

but independent of the type, if you call `bar(v)` inside foo, it will always pass `int &` unless you explicitly cast `&&` onto the type again with std::forward. In the case where the type already contains `&` casting `&&` onto it has no effect because of reference collapsing.

include/seqan3/range/view/detail.hpp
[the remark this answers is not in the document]
h-2: But it is a comment?

include/seqan3/range/view/take_until.hpp
smehringer: hm this really unintuitive because our take_until stops at the FIRST occurrence right?

Additionally I see the following problems for IO:
1) there might be formats where empty (e.g. tab separated) fields are allowed (luckily it is not in SAM), what about BLAST?
2) we automatically allow wrongly formatted files which include an extra newline

Is this recursion only because of the possibility of (`\r\n` as new line)?

include/seqan3/range/view/detail.hpp
h-2: What do you mean by moving a copy? The only thing that `std::forward<T>(t)` does is `static_cast<T&&>(t) ` that's the full implementation.

It has no knowledge whatsoever about how your function took any of it's parameters, it just makes use of reference collapsing so that if `T` was e.g. `int &` than the cast will not change the type. But if the type was `int` it will become `int &&`.

That's exactly the behaviour we want in that situation or not?

include/seqan3/range/view/detail.hpp
h-2: I will do some changes to the toturials next week and include something like this.

include/seqan3/range/view/detail.hpp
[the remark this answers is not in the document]
smehringer: the reason is not

include/seqan3/range/view/detail.hpp
smehringer: your explanation above resolved my misunderstanding!

include/seqan3/range/view/detail.hpp
h-2: I am trying to say: impl() calls operator() of the proto-adaptor since the proto-adaptor has to have an operator() for handling the range anyway.
Is that not clear when looking at the code? :thinking:

include/seqan3/range/view/take_until.hpp
h-2: not resolved?

include/seqan3/range/view/take_until.hpp
h-2: not resolved?

include/seqan3/range/view/detail.hpp
h-2: What formulation would you prefer?

include/seqan3/range/view/detail.hpp
h-2: (yes proto-adaptor as a term was dropped from the standard, there is no special term for them anymore, but we can still introduce it in our documentation)

include/seqan3/range/view/detail.hpp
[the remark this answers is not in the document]
h-2: :expressionless:

include/seqan3/range/view/take_until.hpp
smehringer: > This does not change the existing take_until

I know, but it is called similar but behaves differently, which is unintuitive.

What about my other concerns/questions?

include/seqan3/range/view/detail.hpp
smehringer: Ah ok, this just made it clear. Yes it is obvious from the implementation but I suspected there is more to it when I read the comment. 
 
```suggestion
 * This template delegates back to the proto adaptor for handling range input (impl() calls operator() of the proto-adaptor). 
```
no more. :)

include/seqan3/range/view/detail.hpp
smehringer: ok fine. but then add a definition in the tutorial and the documentation.

include/seqan3/range/view/detail.hpp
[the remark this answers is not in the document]
smehringer: :grin:

include/seqan3/range/view/take_until.hpp
h-2: > I know, but it is called similar but behaves differently, which is unintuitive.

How would you call it? The returned range is identical to that of take_until...

> What about my other concerns/questions?

> there might be formats where empty (e.g. tab separated) fields are allowed (luckily it is not in SAM), what about BLAST?

How is this related to my changes? I am not changing either of the formats...

> we automatically allow wrongly formatted files which include an extra newline

For the last line we currently accept newline or no-newline in any case. Accepting two newlines is not a breaking change since we agreed that we don't validate formats by default anyway, we just try to extract the information we need.

smehringer: > How would you call it? The returned range is identical to that of take_until...

I would call it the same but change the behaviour to that it does not consume all character matching `fun` but just the first. The consuming of `\r\n` should be part of `take_line` not `take_until_and_comsume`.

> How is this related to my changes? I am not changing either of the formats...

well I thought you added this functionality specifically for the use in our formats

> For the last line we currently accept newline or no-newline in any case. Accepting two newlines is not a breaking change since we agreed that we don't validate formats by default anyway, we just try to extract the information we need.

we will also allow

```
> seq1
ACTACTAG


ACTAGC
AC
```

But ok you are right that we do not validate formats anyway.

h-2: > I would call it the same but change the behaviour to that it does not consume all character matching fun but just the first.

But I added this functionality explicitly so that we can use it in I/O and don't have to call detail::consume so often. It's useful for skipping multiple whitespaces among other things.

> well I thought you added this functionality specifically for the use in our formats

Yes, and it works for those situations, or not? Skipping a single character is easy to do, you can just cal `++it;` , but skipping multiple ones involves another `detail::consume(foo | view::take_until(...))` which was what I was trying to avoid.

> we will also allow

That's already valid, because we take_until(is_char<'>'>) and filter whitespaces...

h-2:
And can you check if your final two points are met? Concerning the general direction of consuming views we can have a discussion at the strategy meeting and I promise to change the behaviour again if you want to, but right now this is blocking a lot of other small stuff.

rrahn:
Why was this merged? I didn't give a review yet.

PR #865, opened 2019-03-30
[FEATURE] cereal

eseiler:
Adds cereal support (and equality comparison) for SDSL data structures.

* Adds a test header for cereal tests
* Enables tests for containers using cereal (bitcompressed_vector)
* Enables serialisation via cereal for FM-Index
* Removes `load`/`store` function from FM-Index and FM-Index concept

**TODO**
* Would be nice to have a cereal tutorial or dedicated page on the documentation (other PR)

### DIFF
--- a/doc/tutorial/read_mapper/index.md
+++ b/doc/tutorial/read_mapper/index.md
@@ -69,7 +69,7 @@ Your `main` may look like this:
 \endsolution
 
 ## Step 2 - Reading the input
-As a next step, we want to use the parsed file name to read in our reference data. This will be done 
+As a next step, we want to use the parsed file name to read in our reference data. This will be done
 using seqan3::sequence_file_input class. As a guide, you can take a look at the
 \ref tutorial_sequence_file "Sequence I/O Tutorial".
 
@@ -112,7 +112,7 @@ Now that we have the necessary sequence information, we can create an index and 
 We want to create a new function `create_index`:
 * It takes `index_path` and `storage` as parameters
 * Creates a bi_fm_index
-* Stores the bi_fm_index (you may want to `throw` if this fails)
+* Stores the bi_fm_index
 
 We also need to change:
 * `run_program` to now also call `create_index`
@@ -127,7 +127,7 @@ This is the signature of `create_index`:
 \snippet doc/tutorial/read_mapper/read_mapper_indexer_step3.cpp solution
 Here is the complete program:
 \hint
-\include doc/tutorial/read_mapper/read_mapper_indexer_step3.cpp
+\snippet doc/tutorial/read_mapper/read_mapper_indexer_step3.cpp complete
 \endhint
 \endsolution
 
@@ -191,7 +191,7 @@ This is the signature of `map_reads`:
 \snippet doc/tutorial/read_mapper/read_mapper_step2.cpp solution
 Here is the complete program:
 \hint
-\include doc/tutorial/read_mapper/read_mapper_step2.cpp
+\snippet doc/tutorial/read_mapper/read_mapper_step2.cpp complete
 \endhint
 \endsolution
 
@@ -213,7 +213,7 @@ This is the alignment config:
 \snippet doc/tutorial/read_mapper/read_mapper_step3.cpp solution
 Here is the complete program:
 \hint
-\include doc/tutorial/read_mapper/read_mapper_step3.cpp
+\snippet doc/tutorial/read_mapper/read_mapper_step3.cpp complete
 \endhint
 \endsolution
 
@@ -236,6 +236,6 @@ This is the alignment_file_output construction:
 \snippet doc/tutorial/read_mapper/read_mapper_step4.cpp solution
 Here is the complete program:
 \hint
-\include doc/tutorial/read_mapper/read_mapper_step4.cpp
+\snippet doc/tutorial/read_mapper/read_mapper_step4.cpp complete
 \endhint
 \endsolution
--- a/include/seqan3/search/fm_index/detail/csa_alphabet_strategy.hpp
+++ b/include/seqan3/search/fm_index/detail/csa_alphabet_strategy.hpp
@@ -162,6 +162,30 @@ namespace sdsl
             m_C.load(in);
             read_member(m_sigma, in);
         }
+
+        template <typename archive_t>
+        void CEREAL_SAVE_FUNCTION_NAME(archive_t & ar) const
+        {
+            ar(CEREAL_NVP(m_C));
+            ar(CEREAL_NVP(m_sigma));
+        }
+
+        template <typename archive_t>
+        void CEREAL_LOAD_FUNCTION_NAME(archive_t & ar)
+        {
+            ar(CEREAL_NVP(m_C));
+            ar(CEREAL_NVP(m_sigma));
+        }
+
+        bool operator==(plain_byte_alphabet const & other) const noexcept
+        {
+            return (m_C == other.m_C) && (m_sigma == other.m_sigma);
+        }
+
+        bool operator!=(plain_byte_alphabet const & other) const noexcept
+        {
+            return !(*this == other);
+        }
         //!\endcond
     };
 
--- a/include/seqan3/search/fm_index/fm_index.hpp
+++ b/include/seqan3/search/fm_index/fm_index.hpp
@@ -388,17 +388,38 @@ public:
         return size() == 0;
     }
 
-    // operator== not implemented by sdsl indices yet
-    // bool operator==(fm_index const & rhs) const noexcept
-    // {
-    //     return index == rhs.index;
-    // }
+    /*!\brief Compares two indices.
+     * \returns `true` if the indices are equal, false otherwise.
+     *
+     * ### Complexity
+     *
+     * Linear.
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    bool operator==(fm_index const & rhs) const noexcept
+    {
+        // (void) rhs;
+        return (index == rhs.index) && (text_begin == rhs.text_begin);
+    }
 
-    // operator== not implemented by sdsl indices yet
-    // bool operator!=(fm_index const & rhs) const noexcept
-    // {
-    //     return !(*this == rhs);
-    // }
+    /*!\brief Compares two indices.
+     * \returns `true` if the indices are unequal, false otherwise.
+     *
+     * ### Complexity
+     *
+     * Linear.
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    bool operator!=(fm_index const & rhs) const noexcept
+    {
+        return !(*this == rhs);
+    }
 
     /*!\brief Returns a seqan3::fm_index_cursor on the index that can be used for searching.
      *        \if DEV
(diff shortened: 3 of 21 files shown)


.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "sdsl-lite"]
 	path = submodules/sdsl-lite
-	url = https://github.com/xxsds/sdsl-lite
+	url = https://github.com/eseiler/sdsl-lite
```
h-2: This was not supposed to be committed, right?

doc/tutorial/read_mapper/index.md
```diff
@@ -236,6 +236,6 @@ This is the alignment_file_output construction:
 \snippet doc/tutorial/read_mapper/read_mapper_step4.cpp solution
 Here is the complete program:
 \hint
-\include doc/tutorial/read_mapper/read_mapper_step4.cpp
+\snippet doc/tutorial/read_mapper/read_mapper_step4.cpp complete
```
h-2: Why these changes?

h-2: ok, got it

include/seqan3/search/fm_index/detail/csa_alphabet_strategy.hpp
```diff
@@ -162,6 +162,30 @@ namespace sdsl
             m_C.load(in);
             read_member(m_sigma, in);
         }
+
+        template <typename archive_t>
+        void CEREAL_SAVE_FUNCTION_NAME(archive_t & ar) const
+        {
+            ar(CEREAL_NVP(m_C));
+            ar(CEREAL_NVP(m_sigma));
```
h-2: doesn't NVP mean named-value-pair? Where is the name?

include/seqan3/search/fm_index/detail/csa_alphabet_strategy.hpp
```diff
@@ -162,6 +162,30 @@ namespace sdsl
...
+        {
+            return (m_C == other.m_C) && (m_sigma == other.m_sigma);
+        }
+
+        bool operator!=(plain_byte_alphabet const & other) const noexcept
+        {
+            return !(*this == other);
+        }
         //!\endcond
     };
```
h-2: (Why) is this file still necessary?

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -388,105 +388,76 @@ class fm_index
...
      * ### Exceptions
      *
      * No-throw guarantee.
      */
-    cursor_type begin() const noexcept
+    bool operator==(fm_index const & rhs) const noexcept
     {
-        return {*this};
+        return (index == rhs.index) && (text_begin == rhs.text_begin) && (text_begin_ss == rhs.text_begin_ss) &&
+               (text_begin_rs == rhs.text_begin_rs);
```
h-2: What is being compared? Why is it linear?

.gitmodules
eseiler: Nope, I just wanted to check if CI runs through with my SDSL PR

include/seqan3/search/fm_index/detail/csa_alphabet_strategy.hpp
eseiler: the name is the variable name.
The macro ist short hand for, e.g., `cereal::make_nvp("m_C", m_C);`

include/seqan3/search/fm_index/detail/csa_alphabet_strategy.hpp
eseiler: In general?
The SDSL alphabet strategies map the occurring symbols to ranks. This reduces the memory consumption if not every symbol occurs, but increases the runtime (additional lookup in the map).
This file uses identity (maps to the rank), hence faster.

Though I'd argue this could probably go into the SDSL.

include/seqan3/search/fm_index/fm_index.hpp
eseiler: We're comparing the underlying int_vector representation and the test begin positions.
Since comparing the vectors needs to compare the elements, the complexity is linear relative to its size, no?

include/seqan3/search/fm_index/detail/csa_alphabet_strategy.hpp
h-2: Yes, but I mean, do you need the names if they are as useless as they currently are? :wink:

include/seqan3/search/fm_index/detail/csa_alphabet_strategy.hpp
h-2: > Though I'd argue this could probably go into the SDSL.

That's what I meant. Can you open a card for this?

include/seqan3/search/fm_index/fm_index.hpp
h-2: True. But do you need to compare the actual index? Aren't two indexes of the same text identical?

test/include/seqan3/test/cereal.hpp
```diff
@@ -0,0 +1,59 @@
...
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides cereal functionality for tests.
+ * \author Enrico Seiler <enrico.seiler AT fu-berlin.de>
+ */
+
+#pragma once
+#if SEQAN3_WITH_CEREAL
```
h-2: You need to include seqan3/platform.hpp for this macro to be set

test/include/seqan3/test/cereal.hpp
```diff
@@ -0,0 +1,59 @@
...
+namespace test
+{
+
+template <typename in_archive_t, typename out_archive_t, typename TypeParam>
+void do_serialisation(TypeParam const l)
+{
+    // This makes sure the file is also deleted if an exception is thrown in one of the tests below.
+    // Generate unique file name.
+    tmp_filename filename{"cereal_test"};
+    {
```
h-2: newline before

test/include/seqan3/test/cereal.hpp
```diff
@@ -0,0 +1,59 @@
...
+        in_archive_t iarchive{is};
+        iarchive(in_l);
+        EXPECT_EQ(l, in_l);
+    }
+}
+
+} // namespace seqan3::test
+
+} //namespace seqan3
+#endif // SEQAN3_WITH_CEREAL
```
h-2: can you define these functions as NOPs if cereal isn't defined? Than we can save the checks in every test and just do it here.

include/seqan3/search/fm_index/fm_index.hpp
eseiler: Mhm, then I would have to compare the texts. Still linear?
And they could use different traits I guess. Then the indices would be different for the same texts.

test/include/seqan3/test/cereal.hpp
eseiler: 💡

test/include/seqan3/test/cereal.hpp
eseiler: Good idea, will do

include/seqan3/search/fm_index/detail/csa_alphabet_strategy.hpp
eseiler: Expressive uselessness? 🤷‍♂️

test/include/seqan3/test/cereal.hpp
eseiler: But that would mean testing every archive in general, but this should be fine, right?

include/seqan3/search/fm_index/fm_index.hpp
h-2: Well, my thought was, that 
  * if the indexes are guaranteed to be the same for the same text
  * and you store a pointer to the text
  * and the pointer value is identical
→ the index is identical and can be compared in `O(1)`. 

You could opt for a less strict definition of identity and compare the contents of the text instead. Or even say that the text is irrelevant and compare the contents of the index, but from my POV that would be unintuitive, because indexes over different alphabets could be equal if they have the same composition of ranks.

test/include/seqan3/test/cereal.hpp
h-2: You mean always testing textarchives and binaryarchives? Yes, that would be desirable even, I think.

include/seqan3/search/fm_index/fm_index.hpp
eseiler: Mhm, the problem is that 

1) You could have the same text twice and construct an index each and the indices would not be equal though they are.
2) The text is only a pointer, i.e., when I load an index, the text is a `nullptr` - so I can't really compare the text.

h-2: > You could have the same text twice and construct an index each and the indices would not be equal though they are.

Yes, if you want this behaviour you need to compare the texts.

> The text is only a pointer, i.e., when I load an index, the text is a nullptr - so I can't really compare the text.

Is the index in a valid state if the pointer is a nullptr?

eseiler: The index is.
The only thing that doesn't work is to ask the `fm_index_cursor` or such for the `query()` since it needs text access for it.
The search works as expected since it only returns text positions and doesn't access the indexed text.

eseiler: Btw

>because indexes over different alphabets could be equal if they have the same composition of ranks.

the `fm_index` is templatised over the text type. So we would have to have another `operator==` anyway to compare, e.g., `fm_index<dna4_vector>` and `fm_index<std::string>`?

h-2: > The index is.

Hm, we need to discuss this at the strategy meeting. I thought certain indexes also do in-text verification. That would not work, right?

> the fm_index is templatised over the text type. So we would have to have another operator== anyway to compare e.g., fm_index<dna4_vector> and fm_index<std::string>?

I was just listing different notions of equality above, I don't thing this should really be possible.

h-2: You cannot compare dna4_vector with std::string, why should you be able to compare indexes of either with each other?

eseiler: > Hm, we need to discuss this at the strategy meeting. I thought certain indexes also do in-text verification. That would not work, right?

Well, you still can, but you need to provide the text. The index doesn't own the text and hence it's not serialised with it - that's what we decided on when the fm_index was implemented.

> You cannot compare dna4_vector with std::string, why should you be able to compare indexes of either with each other?

You started the thing with having indices over different alphabets 😀

test/include/seqan3/test/cereal.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+/*!\brief Tests if an object is serialise for all cereal archive types.
+ * \tparam value_t The type to serialise.
+ * \param l The object to serialise.
+ *
+ * If cereal is **not** available, this function is a NOP.
+ * Otherwise it will call do_cerealisation() with cereal's `Binary`, `PortableBinary`, `JSON` and `XML` archives.
+ */
+template <typename value_t>
+void do_serialisation([[maybe_unused]] value_t const l)
```
h-2: this is just testing, but can you do no const and && in any case?

h-2: also above

eseiler: Ok, done

h-2:
fail

PR #867, opened 2019-03-30
Add --license option

### LINKED ISSUE #862: --license to print copyright
There are many tools printing the copyright/license information using `--license` instead of `--copyright`.
I think it would make sense to print the license information with either of both parameters (`--license` or `--copyright`).

tloka:
Print copyright information with `--license` in addition to `--copyright`.
Resolves #862

### DIFF
--- a/doc/tutorial/argument_parser/index.md
+++ b/doc/tutorial/argument_parser/index.md
@@ -59,7 +59,7 @@ The argument parser checks the following restrictions and throws a seqan3::parse
 * **Short identifiers**: must be unique and consist of only a single letter that is alphanumeric characters, `_` or `@`.
 * either the short or long id may be empty but not both at the same time.
 * Only the last positional option may be a list (see [lists](#section_list_positional_options)).
-* The flag identifiers `-h`, `--help`, `--advanced-help`, `--advanced-help`, `--export-help`, `--version`, `--copyright` are predefined and cannot be specified manually or used otherwise.
+* The flag identifiers `-h`, `--help`, `-hh`, `--advanced-help`, `--export-help`, `--version`, `--license` are predefined and cannot be specified manually or used otherwise.
 * The seqan3::argument_parser::parse function may only be called once (per parser).
 
 ## Input restrictions (seqan3::parser_invalid_argument)
@@ -86,7 +86,7 @@ Currently we support the following *special requests*:
 <tr> <td> `-hh/--advanced-help` </td> <td> Prints the advanced help page to the command line (`std::cout`) </td> </tr>
 <tr> <td> `--export-help` </td> <td> Exports the help page in a different format (`std::cout`) </td> </tr>
 <tr> <td> `--version`    </td> <td> Prints the version information to the command line (`std::cout`) </td> </tr>
-<tr> <td> `--copyright`  </td> <td> Prints the copyright information to the command line (`std::cout`) </td> </tr>
+<tr> <td> `--license`  </td> <td> Prints the license information to the command line (`std::cout`) </td> </tr>
 </table>
 
 \assignment{Assignment 1}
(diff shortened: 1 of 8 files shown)


tloka:
> Can you also add the flag in the `argument_parser::parse()` and `argument_parser::init()` documentation? grin
> 
> Otherwise LGMT +1

Done @smehringer

smehringer:
Jenkins issue is unrelated (some `java.lang.OutOfMemoryError: Java heap space` error...)
Code coverage seems also confused and nothing I see to improve.

doc/tutorial/argument_parser/index.md
```diff
@@ -59,7 +59,7 @@ The argument parser checks the following restrictions and throws a seqan3::parse
 * **Short identifiers**: must be unique and consist of only a single letter that is alphanumeric characters, `_` or `@`.
 * either the short or long id may be empty but not both at the same time.
 * Only the last positional option may be a list (see [lists](#section_list_positional_options)).
-* The flag identifiers `-h`, `--help`, `--advanced-help`, `--advanced-help`, `--export-help`, `--version`, `--copyright` are predefined and cannot be specified manually or used otherwise.
+* The flag identifiers `-h`, `--help`, `--advanced-help`, `--advanced-help`, `--export-help`, `--version`, `--copyright`, `--license` are predefined and cannot be specified manually or used otherwise.
```
h-2: why do we have "advanced-help" twice? :thinking:

smehringer: :eyes: because it is super important and I never make errors.. (as well as my reviewers)

@tloka Can you change the firsst `--advanced-help` to `-hh` ?

tloka:
> 
> 
> In general I am unsure about this change. One could reasonably argue that also `--author` or `--legal` should redirect here. Why shouldn't we just pick one of them?

In general I agree, but (maybe this is just subjective) my feeling is that `--license` is way more often used than `--copyright`. That's why I proposed to add it.

doc/tutorial/argument_parser/index.md
tloka: I will in case you finally decide to merge this PR.
Otherwise I can do it in a seperate PR.

h-2:
I don't care much either way, but I would prefer if we have a single name for it and not multiple ones. Do you have references for "--license" being used more often? I haven't seen either on many programs...

tloka:
> 
> 
> I don't care much either way, but I would prefer if we have a single name for it and not multiple ones. Do you have references for "--license" being used more often? I haven't seen either on many programs...

You're right, mostly it is just stored in a license file. However, whenever I saw it as a program parameter it was rather `--license`. I can't remember a single time seeing `--copyright` (but probably just because I was rather trying `--license`, that's what I meant by subjective).
Just looked through some argParse libraries, they all provide rather neither of both. However, also when using google I didn't find `--copyright` a single time whereas `--license` occured from time to time - But no, I don't have proper references for that.
However, I think it's not worse to spend to much time with discussing that and I am not going to fight for `--license`. If you think that `--copyright` is sufficient and will be found be the majority of users, there is no reason to change it. I can just state my feeling which was that I would never have the idea to try a `--copyright` parameter to display license information.

h-2:
Sure, I think the core team can just have a vote on it, I don't care for either term, I just want to pick one of them!

smehringer:
IMOP it's fine to have two names for the same thing but I don't care much so we\ll make a quick vote next Tuesday and I'll tell you the outcome.

smehringer:
Hi @tloka,

We decided to drop `--copyright` in favour of `--license`. Can you delete the copyright flag in your PR?

tloka:
@smehringer shall this change include the naming of fields and functions (e.g., `seqan3::detail::format_copyright()` >> `seqan3::detail::format_license()`) or is this supposed to be only a change at the user interface level?

smehringer:
@tloka good question. I think we should change fields and functions too since it would be confusing to have a `format_copyright` that is triggered by `--licence`

tloka:
@smehringer I recognized that the terms copyright and license were internally used as synonyms and therefore hard to distinguish for the developer.
Thus, I changed the internal fields in the following way: 
`short_copyright` -> `copyright`
`long_copyright` -> `license`
It should be clear now to put the copyright message into `copyright` and the actual license text into the `license` field.
The copyright is then printed e.g. in the help message while `--license` will print the full license text (if available).
I hope, this makes some sense :)

include/seqan3/argument_parser/auxiliary.hpp
```diff
@@ -80,11 +80,11 @@ struct argument_parser_meta_data // holds all meta information
     //!\brief A link to github, your website or a wiki page.
     std::string url;
     //!\brief Brief copyright (and/or license) information.
-    std::string short_copyright;
-    /*!\brief Detailed copyright information that will be displayed
-     *        when the user specifies "--copyright" on the command line.
+    std::string copyright;
```
smehringer:
```suggestion
    std::string licence;
```
:grin: ?

tloka: damn it ... -.-'

smehringer:
> I hope, this makes some sense :)

eehm now I am confused  :frowning: 

I though copyright and license **are synonyms **? 
So as a developer I would not now what to put in copyright and what to put in license.

include/seqan3/argument_parser/auxiliary.hpp
tloka: wait, no, this would be correct for the new copyright/license distinction.
The old short_copyright ist just copyright now.
But I messed up something in the tutorial ^^

tloka:
No they are not. Copyright declares who owns something, license declares the conditions under that someone else can use it.
e.g.,
Copyright: (c) Tobias Loka, 2015-2019
License: Licensed under BSD 3-clause

rrahn:
@tloka  thanks for this PR. We will have in the near future a reevaluation of the argument parser and we will rediscuss this topic. All information are listed on the referenced issue.

PR #871, opened 2019-04-04
[FIX] Uninitialised warning in SAM format

eseiler:
Fixes
```
/Users/enricoseiler/git/seqan3/include/seqan3/io/alignment_file/format_sam.hpp:668:54: error: '*((void*)&<anonymous> +4)' may be used uninitialized in this function [-Werror=maybe-uninitialized]
                 stream << (get<1>(mate).value() + 1) << separator; // SAM is 1 based
```

### DIFF
--- a/include/seqan3/io/alignment_file/format_sam.hpp
+++ b/include/seqan3/io/alignment_file/format_sam.hpp
@@ -618,10 +618,8 @@ public:
 
         stream << separator;
 
-        if (ref_offset.has_value())
-            stream << (ref_offset.value() + 1) << separator; // SAM is 1 based
-        else
-            stream << '0' << separator;                      // 0 indicates unmapped read if optional is not set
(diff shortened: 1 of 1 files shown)


include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -618,10 +618,8 @@ class alignment_file_format_sam
...
         stream << separator;
 
-        if (ref_offset.has_value())
-            stream << (ref_offset.value() + 1) << separator; // SAM is 1 based
-        else
-            stream << '0' << separator;                      // 0 indicates unmapped read if optional is not set
+        // SAM is 1 based, 0 indicates unmapped read if optional is not set
+        stream << (ref_offset.value_or(-1) + 1) << separator;
 
         stream << static_cast<unsigned>(mapq) << separator;
```
h-2: Now that I see this, why are you casting to unsigned, isn't the size fixed?

PR #872, opened 2019-04-04
[MISC] Doxygen warning construction fix alphabet I

MitraDarja:
Fixes for doxygen warning due to uncommented default construction in alphabet part I.
There are some empty constructors, I commented them with empty, hope that is okay.

### DIFF
--- a/include/seqan3/alphabet/aminoacid/aa20.hpp
+++ b/include/seqan3/alphabet/aminoacid/aa20.hpp
@@ -72,12 +72,12 @@ public:
     /*!\name Constructors, destructor and assignment
      * \{
      */
-    constexpr aa20() : base_t{} {}
-    constexpr aa20(aa20 const &) = default;
-    constexpr aa20(aa20 &&) = default;
-    constexpr aa20 & operator=(aa20 const &) = default;
-    constexpr aa20 & operator=(aa20 &&) = default;
-    ~aa20() = default;
+    constexpr aa20() : base_t{} {}                      //!< Defaulted.
+    constexpr aa20(aa20 const &) = default;             //!< Defaulted.
+    constexpr aa20(aa20 &&) = default;                  //!< Defaulted.
+    constexpr aa20 & operator=(aa20 const &) = default; //!< Defaulted.
+    constexpr aa20 & operator=(aa20 &&) = default;      //!< Defaulted.
+    ~aa20() = default;                                  //!< Defaulted.
 
     using base_t::base_t;
     //!\}
--- a/include/seqan3/alphabet/aminoacid/aa27.hpp
+++ b/include/seqan3/alphabet/aminoacid/aa27.hpp
@@ -54,12 +54,12 @@ public:
     /*!\name Constructors, destructor and assignment
      * \{
      */
-    constexpr aa27() : base_t{} {}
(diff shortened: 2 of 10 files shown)


include/seqan3/alphabet/aminoacid/aa20.hpp
```diff
@@ -72,12 +72,12 @@ class aa20 : public aminoacid_base<aa20, 20>
     /*!\name Constructors, destructor and assignment
      * \{
      */
-    constexpr aa20() : base_t{} {}
-    constexpr aa20(aa20 const &) = default;
-    constexpr aa20(aa20 &&) = default;
-    constexpr aa20 & operator=(aa20 const &) = default;
-    constexpr aa20 & operator=(aa20 &&) = default;
-    ~aa20() = default;
+    constexpr aa20() : base_t{} {}                      //!< Empty
```
h-2: This was a mistake originally also. Can you explicitly default the default constructor here and then document it as defaulted, too?

[If anything fails because of this, you can revert it again and I will take care of it. But it should work and if you are touching the lines anyway...]

h-2: Can you also add the `noexcept` keyword and align the keywords? And don't forget the period at the end :nail_care: 

e.g
```cpp
    constexpr aa20()                         noexcept = default; //!< Defaulted.
    constexpr aa20(aa20 const &)             noexcept = default; //!< Defaulted.
    constexpr aa20(aa20 &&)                  noexcept = default; //!< Defaulted.
    constexpr aa20 & operator=(aa20 const &) noexcept = default; //!< Defaulted.
    constexpr aa20 & operator=(aa20 &&)      noexcept = default; //!< Defaulted.
    ~aa20()                                  noexcept = default; //!< Defaulted.
```

Thanks!

MitraDarja: But in the other doxygen fixes, I didn't add a period in the end... Do you want me to add them also to the already merged options as well?

h-2: Hm, I am not sure, because I wasn't there when this was discussed in escala. @smehringer did you not discuss to add a `.` ?

In any case, you don't need to change things already changed, it would just apply to the new changes.

smehringer: I think discussing this takes too much time :sleepy: 
I really don't care about this dot. It makes this super important documentation one char longer... which maybe forces us to use the //\brief` version in probably 1% of the cases which is why I would allow "no dot".

h-2: Hm, well, I also think we shouldn't discuss too long. I propose to have the dot, because everything else has it, but not change code that was already changed.

rrahn: I agree with @h-2. All documented code (even single word sentences) should end with a dot. The only exceptions are `\name` expressions which are captions.

smehringer: fine

MitraDarja: I got errors, adding the noexcept, so I didn't.

PR #875, opened 2019-04-05
Argument parser fix

### LINKED ISSUE #854: Export man pages does not work with "--export-help=man"
Works only with "--export-help man"

### LINKED ISSUE #874: arg parser broken again on c++20
@smehringer With #840 you reverted my fixes from #708 and now the arg parser produces warnings again with c++20:
http://cdash.seqan.de/viewBuildError.php?onlydeltap&buildid=306570

smehringer:
Resolves #874 
Resolves #854 
and some other stuff (see single commit messages)

Best reviewed commit by commit


include/seqan3/argument_parser/detail/format_help.hpp
```diff
@@ -51,7 +51,7 @@ class format_help : public format_help_base<format_help>
      */
     format_help() = default;                                   //!< Defaulted.
     format_help(format_help const & pf) = default;             //!< Defaulted.
-    format_help & operator=(format_help const & pf) = default; //!< Defaulted.
+    format_help & operator=(format_help const &) = default; //!< Defaulted.
```
eseiler:
```suggestion
    format_help & operator=(format_help const &) = default;    //!< Defaulted.
```

h-2: indent

include/seqan3/argument_parser/detail/format_help.hpp
```diff
@@ -367,9 +367,7 @@ class format_help : public format_help_base<format_help>
      * \param[in] layout The command line parameters (e.g. padding) for correct printing.
      * \param[in] tab    The position offset (indentation) to start printing at.
      */
-    void print_text(std::string const & text,
-                    console_layout_struct const & layout,
-                    unsigned const tab)
+    void print_text(std::string const & text, unsigned const tab)
```
eseiler: Why no fixed length integer?

include/seqan3/argument_parser/detail/format_help.hpp
```diff
@@ -424,7 +422,7 @@ class format_help : public format_help_base<format_help>
     //!\brief Needed for correct formatting while calling different print functions.
     bool prev_was_paragraph{false};
     //!\brief Stores the relevant parameters of the documentation on the screen.
-    console_layout_struct layout;
+    static inline console_layout_struct const layout{};
```
h-2: ?

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -504,12 +504,12 @@ class argument_parser
                 std::string export_format{argv[i+1]};
 
                 if (export_format == "html")
-                    format = detail::format_html{};
+                    format = detail::format_html{true};
                 else if (export_format == "man")
-                    format = detail::format_man{};
+                    format = detail::format_man{true};
```
h-2: what is being set to true here?

include/seqan3/argument_parser/detail/format_help.hpp
eseiler: If it's static and const, it might as well be constexpr.

But this struct holds

```
            // Guess terminal screen width and set into layout.
            unsigned cols = get_terminal_width();
```

So the terminal width is set at compile time anyway?

smehringer: I had this super weird behaviour that when it was not const, then within print_text the values of `layout` suddenly changed :confused: Since it does not change anyway I made it const  which resolved the issue..

include/seqan3/argument_parser/argument_parser.hpp
smehringer: The `show advanced options`

include/seqan3/argument_parser/detail/format_help.hpp
smehringer: because I did not want to touch everything that I designed rather bad 2 years ago when I wrote this :grin:

include/seqan3/argument_parser/detail/format_help.hpp
smehringer: I guess I could make this constexpr but since the format is not I don't know if it is worth it

h-2: > then within print_text the values of layout suddenly changed

o_O

h-2: Did you initialise it before?

include/seqan3/argument_parser/argument_parser.hpp
h-2: Hm, that's kind of obscure. If they always show advanced, you shouldn't have to pass this parameter.

include/seqan3/argument_parser/detail/format_help.hpp
smehringer: I did, same result.

include/seqan3/argument_parser/argument_parser.hpp
smehringer: Well, they use the  `add_option` call from the base class which only adds a list item for advanced options if the spec is set. I would need to reimplement the 3 functions for html, man and ctd format then which I explicitly tried to avoid since they do the same thing. So this is no additional line of code versus +90 LOC.

h-2: You could just default-init it to true and not provide a constructor with the argument, except for format_help which is the only one which needs to edit it...

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -511,8 +526,9 @@ class argument_parser
                 // else if (export_format == "ctd")
                 //     format = detail::format_ctd{};
                 else
-                    throw validation_failed("Validation Failed. "
-                                            "Value of --export-help must be one of [html, man, ctd]");
+                    throw validation_failed{"Validation Failed for option --export-help: "
```
eseiler:
```suggestion
                    throw validation_failed{"Validation failed for option --export-help: "
```
?

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -352,10 +352,12 @@ class argument_parser
     /*!\brief Adds an help page text line to the seqan3::argument_parser.
      * \param[in] text The text to print.
      * \param[in] line_is_paragraph Whether to insert as paragraph
-     *            or just a line (only one line break if not a paragraph).
-     * \details This only affects the help page and other output formats.
+     *            or just a line (Default: false).
+     * \details
+     * If the line is not a paragraph (false), only one line break is appended otherwise two line breaks are appended.
```
eseiler:
```suggestion
     * If the line is not a paragraph (false), only one line break is appended, otherwise two line breaks are appended.
```

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -511,8 +526,9 @@ class argument_parser
                 // else if (export_format == "ctd")
                 //     format = detail::format_ctd{};
                 else
-                    throw validation_failed("Validation Failed. "
-                                            "Value of --export-help must be one of [html, man, ctd]");
+                    throw validation_failed{"Validation Failed for option --export-help: "
+                                            "Value of must be one of [html, man]"};
```
eseiler:
```suggestion
                                            "Value must be one of [html, man]"};
```

include/seqan3/core/detail/to_string.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+#include <seqan3/io/stream/debug_stream.hpp>
+
+namespace seqan3::detail
+{
+
+/*!\brief Streams all parameters and returns a concatenated string.
+ *
+ * \tparam value_type Must be streamable (stream << value).
+ * \param  values Variable number of parameters of any type that implement the stream operator.
```
eseiler: add `\returns`

include/seqan3/core/detail/to_string.hpp
```diff
@@ -0,0 +1,34 @@
...
+/*!\file
+ * \author Svenja Mehringer <svenja.mehringer AT fu-berlin.de>
+ * \brief Auxiliary for pretty printing of exception messages.
+ */
+
+#pragma once
+
+#include <seqan3/io/stream/debug_stream.hpp>
+
+namespace seqan3::detail
```
eseiler: I thought we don't do nested namespaces?

smehringer: o.0?

eseiler: I did it once and was asked to do
```
namespace seqan3
{

namespace detail
{

bla

} // namespace seqan3::detail

} // namespace seqan3
```

But apparently we do it in many places anyways.
I don't mind either way, so you can just resolve this.

h-2: No, it's exactly the other way around as can been seen almost everywhere in the library and in the wiki :point_up:

smehringer:
Code coverage reports two misses.... one `{` and one part of the code I did not touch and is definitely executed. I have no clue how I could improve on that.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -15,7 +15,7 @@
 #include <sstream>
 #include <regex>
 
-#include <seqan3/argument_parser/auxiliary.hpp>
+#include <seqan3/core/detail/to_string.hpp>
```
eseiler: order!

include/seqan3/argument_parser/validators.hpp
```diff
@@ -15,7 +15,7 @@
 #include <sstream>
 #include <regex>
```
eseiler: Also order

include/seqan3/core/detail/to_string.hpp
```diff
@@ -0,0 +1,35 @@
...
+{
+
+/*!\brief Streams all parameters via the seqan3::debug_stream and returns a concatenated string.
+ *
+ * \tparam    value_type Must be streamable (stream << value).
```
eseiler: Don't we want a `requires` for that?

smehringer: Yes but we would need to require that the type is streamable via the debug_stream_type which currently always resolves to true anyway because it has a generic, unconstrained overload :( I tried to changes this which causes a ICE on the rangesv3 1.0 beta branch (see #710)

eseiler: ok

PR #881, opened 2019-04-08
[TEST] Fastq and FM Index Benchmarks

Clemapfel:
Also tweaked the fasta benchmark (aswell as making it look a lot prettier) so the results are comparable with my fastq benchmark

Later on I might add comparisons with Seqan2 for every benchmark I've worked on


smehringer:
Hey, you need to rebase on the current master and resolve merge conflicts first. Tell me if you have any issues with the rebasing.

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,185 @@
...
+// @author: Clemens Cords <clemenscords@fu-berlin.de>
+
+#include <fstream>
+#include <cstring>
+#include <cmath>
+
+#include <benchmark/benchmark.h>
+
+#include <seqan3/io/sequence_file/format_fastq.hpp>
+#include <seqan3/alphabet/nucleotide/dna5.hpp>
```
smehringer: we order the includes alphabetically :nail_care:

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,185 @@
...
+    const std::string id{"@name"};
+    const std::string seq = DNA_SEQ | view::to_char;
+    const std::string qualid{"+"};
+    const std::string quality = QUALITY | view::to_char;
+
+    const unsigned int n_iterations = static_cast<unsigned int>(pow(10, power));
+
+    for (size_t i = 0; i < n_iterations; ++i)
+    {
+
```
smehringer: remove newline

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,185 @@
...
+        file += id + '\n'
```
smehringer: Does this fit into one line (180 char)? 
Also putting "+" in an extra string is a little too generic for the scope of a benchmark that is not exposed to users. you can directly insert the characters here.

EDIT: you seem to use them a lot in the write benchmarks too, so you might just add them once as a global variable and use them throughout the benchmarks.

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,185 @@
...
+// ============================================================================
+void fastq_write(benchmark::State & state)
+{
+    std::ostringstream ostream;
+
+    sequence_file_format_fastq format;
+    sequence_file_output_options options{};
+
+    const std::string id{"@name"};
+    const std::string seq = DNA_SEQ | view::to_char;
```
smehringer: you can write the seq directly with the format! no need to convert to char here.

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,185 @@
...
+// ============================================================================
+void fastq_read_with_quality(benchmark::State & state)
+{
+    std::istringstream istream{generate_dummy_fastq_file(POWER)};
+
+    sequence_file_format_fastq format;
+    const sequence_file_input_options<dna5, false> options{};
+
+    std::string id{};
+    std::string seq{};
```
smehringer:
```suggestion
    std::vector<dna5> seq{};
```
You usually want to read into a dna5 vector, not a string.
Same for all other write functions

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,185 @@
...
+    std::string qual{};
```
smehringer:
```suggestion
    std::vector<phred42> qual{};
```

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+    for (auto _ : state)
+    {
+        format.read(istream, options, seq, id, qual);
```
smehringer: Please `clear` the seq, id and qual every iteration otherwise the information is appended and the sequences grow very large and might skew the benchmark

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,185 @@
...
+    for (auto _ : state)
+    {
+        format.read(istream, options, seq, id, std::ignore);
+    }
+}
+
+BENCHMARK(fasta_read_comparison);
+ */
+
+BENCHMARK_MAIN();
```
smehringer: always put a new line at the end of the file :nail_care:

test/performance/search/index_construction_benchmark.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+#include <benchmark/benchmark.h>
+
+#include <seqan3/io/stream/debug_stream.hpp>
+#include <seqan3/search/fm_index/all.hpp>
+#include <seqan3/alphabet/all.hpp>
+
+using namespace seqan3;
+
+const std::size_t LENGTH = 50;
```
smehringer: We use east const :grin: :nail_care: 
https://mariusbancila.ro/blog/2018/11/23/join-the-east-const-revolution/

test/performance/search/index_construction_benchmark.cpp
```diff
@@ -0,0 +1,132 @@
...
+const unsigned int SEED = 1234;
+
+// ============================================================================
+// generate sequence deterministically based on fixed const seed
+// ============================================================================
+template<Alphabet alphabet_t>
+std::vector<alphabet_t> generate_test_sequence(const std::size_t length)
```
smehringer: Nice work on this function! I'm sorry though to tell you that we already have a function like this in `seqan3/test/include/seeqan3/test/performance/sequence_generator.hpp` which looks very much like yours but is already in the code base so it would be great if you use it.
It also just returns a vector over alphabet_t.

test/performance/search/index_construction_benchmark.cpp
```diff
@@ -0,0 +1,132 @@
...
+        text.push_back(s);
+    }
+
+    return text;
+}
+
+// ============================================================================
+// "overload" for chars
+// ============================================================================
+std::vector<char> generate_test_sequence_char(const std::size_t length)
```
smehringer: hm can you try if our function in the code base works for `char` ? 
(Exchange the member function `assign_rank` with the free function before you try it)
line 41 should be:
```cpp
        sequence.push_back(assign_rank(alphabet_t{}, dis_alpha(gen)));
```
I think assign rank is overloaded for `char`

test/performance/search/index_construction_benchmark.cpp
```diff
@@ -0,0 +1,132 @@
...
+    fm_index<std::vector<alphabet_t>> index{};
+
+    for (auto _ : state)
+    {
+        index = fm_index{sequence};
+    }
+}
+
+// ============================================================================
+// construct as many bi fm indexes as possible
```
smehringer:
```suggestion
// construct as many bi fm indices as possible
```

test/performance/search/index_construction_benchmark.cpp
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    return text;
+}
+
+// ============================================================================
+// construct as many fm indexes as possible
+// ============================================================================
+template<Alphabet alphabet_t>
+void construct_fm_index(benchmark::State & state)
```
smehringer: If you templatize your function over the fm index type you can remove the extra bi fm construction :)

you can then call the benchmark like this:
```
BENCHMARK_TEMPLATE(construct_index, fm_index<std::vector<dna4>>, dna4);
```

You can even access the alphabet type directly in from index via:
```
using alphabet_t = typename index_t:::char_type
```
Then you do not need the alphabet template argument.

test/performance/search/index_construction_benchmark.cpp
Clemapfel: I edited ``seqan3/test/include/seqan3/test/performance/sequence_generator.hpp`` and added it to this pull request since assign_rank is indeed overloaded for chars

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+// written for the purpose of the Bch Softwarepraktikum, early 2019 FU Berlin
+// @author: Clemens Cords <clemenscords@fu-berlin.de>
+
+#include <benchmark/benchmark.h>
+#include <cstring>
+#include <cmath>
+#include <chrono>
+#include <fstream>
+#include <seqan3/alphabet/nucleotide/dna5.hpp>
```
smehringer: new line between stl includes and seqan3 inlcudes. (Sorry for being so picky :D)

test/performance/io/format_fastq_benchmark.cpp
smehringer: not resolved yet

test/performance/io/format_fastq_benchmark.cpp
smehringer: not resolved yet

test/performance/io/format_fastq_benchmark.cpp
smehringer: Can you also Pause the state for clearing?

test/performance/search/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+#seqan3_benchmark(index_construction_benchmark.cpp)
```
smehringer: Found another missing newline :dark_sunglasses:

Clemapfel:
Closing this pull request to split both benchmarks into their own PR. I redid 90% of the code anyway so there's no point to keeping the commit history

PR #883, opened 2019-04-08
fixes some issues with view::take / view::drop / view::slice

h-2:
supersedes #852 

fixes:
  * view::drop and view::take only return sub-ranges on ForwardingRanges
  * our inherited_iterator_base can now inherit from iterators that are actually pointers (by wrapping them instead of inheriting from them)
  * switch more code to use our view::slice instead of ranges::view::slice

### DIFF
--- a/include/seqan3/range/detail/inherited_iterator_base.hpp
+++ b/include/seqan3/range/detail/inherited_iterator_base.hpp
@@ -21,6 +21,10 @@
 namespace seqan3::detail
 {
 
+//!\brief An empty class type used in meta programming.
+struct empty_type
+{};
+
 /*!\brief A CRTP base template for creating iterators that inherit from other iterators.
  * \tparam derived_t The CRTP specialisation.
  * \tparam base_t    The type to inherit from; must satisfy std::Iterator.
@@ -31,6 +35,8 @@ namespace seqan3::detail
  *
  * This template enables you to inherit from another iterator and just overload those functions
  * that you wish to change.
+ *
+ * Note that many of this class's members assume that the derived type is constructible from the base type.
  *
  * ### Example
  *
@@ -43,7 +49,7 @@ namespace seqan3::detail
  * \snippet test/unit/range/detail/inherited_iterator_base_test.cpp inherited_iterator_base def
  */
 template <typename derived_t, std::Iterator base_t>
-class inherited_iterator_base : public base_t
+class inherited_iterator_base : public std::conditional_t<std::is_pointer_v<base_t>, empty_type, base_t>
 {
 public:
     /*!\name Associated types
@@ -58,31 +64,54 @@ public:
     //!\}
 
     /*!\name Constructors, destructor and assignment
+     * \brief The exception specification is explicitly "inherited" to also work for pointers as base.
      * \{
      */
-    inherited_iterator_base()                                                          = default; //!< Defaulted.
-    constexpr inherited_iterator_base(inherited_iterator_base const & rhs)             = default; //!< Defaulted.
-    constexpr inherited_iterator_base(inherited_iterator_base && rhs)                  = default; //!< Defaulted.
-    constexpr inherited_iterator_base & operator=(inherited_iterator_base const & rhs) = default; //!< Defaulted.
-    constexpr inherited_iterator_base & operator=(inherited_iterator_base && rhs)      = default; //!< Defaulted.
-    ~inherited_iterator_base()                                                         = default; //!< Defaulted.
+    constexpr inherited_iterator_base()
+        noexcept(std::is_nothrow_default_constructible_v<base_t>)                       = default; //!< Defaulted.
+    constexpr inherited_iterator_base(inherited_iterator_base const & rhs)
+        noexcept(std::is_nothrow_copy_constructible_v<base_t>)                          = default; //!< Defaulted.
+    constexpr inherited_iterator_base(inherited_iterator_base && rhs)
+        noexcept(std::is_nothrow_move_constructible_v<base_t>)                          = default; //!< Defaulted.
+    constexpr inherited_iterator_base & operator=(inherited_iterator_base const & rhs)
+        noexcept(std::is_nothrow_copy_assignable_v<base_t>)                             = default; //!< Defaulted.
+    constexpr inherited_iterator_base & operator=(inherited_iterator_base && rhs)
+        noexcept(std::is_nothrow_move_assignable_v<base_t>)                             = default; //!< Defaulted.
+    ~inherited_iterator_base()
+        noexcept(std::is_nothrow_destructible_v<base_t>)                                = default; //!< Defaulted.
 
-    inherited_iterator_base(base_t it) : base_t{it} {}
+    //!\brief Delegate to base class if inheriting from non-pointer iterator.
+    constexpr inherited_iterator_base(base_t it) noexcept(std::is_nothrow_move_constructible_v<base_t>)
+    //!\cond
+        requires !std::is_pointer_v<base_t>
+    //!\endcond
+        : base_t{std::move(it)}
+    {}
+
+    //!\brief Initialise member if deriving from pointer.
+    constexpr inherited_iterator_base(base_t it) noexcept
+    //!\cond
+        requires std::is_pointer_v<base_t>
+    //!\endcond
+        : member{std::move(it)}
+    {}
     //!\}
 
     /*!\name Comparison operators
      * \brief Unless specialised in derived_type, all operators perform base_t's operator and cast to derived_t.
      * \{
      */
-    constexpr bool operator==(derived_t const & rhs) const noexcept(noexcept(base_t{} == base_t{}))
+    constexpr bool operator==(derived_t const & rhs) const
+        noexcept(noexcept(std::declval<base_t &>() == std::declval<base_t &>()))
     //!\cond
         requires std::EqualityComparable<base_t>
     //!\endcond
     {
-        return *this_to_base() == static_cast<base_t>(rhs);
+        return *this_to_base() == *rhs.this_to_base();
     }
 
-    constexpr bool operator!=(derived_t const & rhs) const noexcept(noexcept(base_t{} == base_t{}))
+    constexpr bool operator!=(derived_t const & rhs) const
+        noexcept(noexcept(std::declval<base_t &>() == std::declval<base_t &>()))
     //!\cond
         requires std::EqualityComparable<base_t>
     //!\endcond
@@ -90,23 +119,26 @@ public:
         return !(*this == rhs);
     }
 
-    constexpr bool operator<(derived_t const & rhs) const noexcept(noexcept(base_t{} < base_t{}))
+    constexpr bool operator<(derived_t const & rhs) const
+        noexcept(noexcept(std::declval<base_t &>() < std::declval<base_t &>()))
     //!\cond
         requires std::StrictTotallyOrdered<base_t>
     //!\endcond
     {
-        return *this_to_base() < static_cast<base_t>(rhs);
+        return *this_to_base() < *rhs.this_to_base();
     }
 
-    constexpr bool operator>(derived_t const & rhs) const noexcept(noexcept(base_t{} > base_t{}))
+    constexpr bool operator>(derived_t const & rhs) const
+        noexcept(noexcept(std::declval<base_t &>() > std::declval<base_t &>()))
     //!\cond
         requires std::StrictTotallyOrdered<base_t>
     //!\endcond
     {
-        return *this_to_base() > static_cast<base_t>(rhs);
+        return *this_to_base() > *rhs.this_to_base();
     }
 
-    constexpr bool operator<=(derived_t const & rhs) const noexcept(noexcept(base_t{} < base_t{}))
+    constexpr bool operator<=(derived_t const & rhs) const
+        noexcept(noexcept(std::declval<base_t &>() > std::declval<base_t &>()))
     //!\cond
         requires std::StrictTotallyOrdered<base_t>
     //!\endcond
@@ -114,7 +146,8 @@ public:
         return !(*this > rhs);
     }
 
-    constexpr bool operator>=(derived_t const & rhs) const noexcept(noexcept(base_t{} < base_t{}))
+    constexpr bool operator>=(derived_t const & rhs) const
+        noexcept(noexcept(std::declval<base_t &>() < std::declval<base_t &>()))
     //!\cond
         requires std::StrictTotallyOrdered<base_t>
     //!\endcond
@@ -128,7 +161,7 @@ public:
      * \{
     */
     //!\brief Pre-increment, return updated iterator.
-    constexpr derived_t & operator++() noexcept(noexcept(++base_t{}))
+    constexpr derived_t & operator++() noexcept(noexcept(++std::declval<base_t &>()))
     //!\cond
         requires std::InputIterator<base_t>
     //!\endcond
@@ -138,18 +171,19 @@ public:
     }
 
     //!\brief Post-increment, return previous iterator state.
-    constexpr derived_t operator++(int) noexcept(noexcept(++base_t{}))
+    constexpr derived_t operator++(int) noexcept(noexcept(++std::declval<derived_t &>()) &&
+                                                 noexcept(derived_t(std::declval<base_t &>())))
     //!\cond
         requires std::InputIterator<base_t>
     //!\endcond
     {
-        derived_t cpy{*this};
+        derived_t cpy{*this_to_base()};
         ++(*this_derived());
         return cpy;
     }
 
     //!\brief Pre-decrement, return updated iterator.
-    constexpr derived_t & operator--() noexcept(noexcept(--base_t{}))
+    constexpr derived_t & operator--() noexcept(noexcept(--std::declval<base_t &>()))
     //!\cond
         requires std::BidirectionalIterator<base_t>
     //!\endcond
@@ -159,18 +193,19 @@ public:
     }
 
     //!\brief Post-decrement, return previous iterator state.
-    constexpr derived_t operator--(int) noexcept(noexcept(--base_t{}))
+    constexpr derived_t operator--(int) noexcept(noexcept(--std::declval<derived_t &>()) &&
+                                                 noexcept(derived_t{std::declval<base_t &>()}))
     //!\cond
         requires std::BidirectionalIterator<base_t>
     //!\endcond
     {
-        derived_t cpy{*this};
+        derived_t cpy{*this_to_base()};
         --(*this_derived());
         return cpy;
     }
 
     //!\brief Move iterator to the right.
-    constexpr derived_t & operator+=(difference_type const skip) noexcept(noexcept(base_t{} += skip))
+    constexpr derived_t & operator+=(difference_type const skip) noexcept(noexcept(std::declval<base_t &>() += skip))
     //!\cond
         requires std::RandomAccessIterator<base_t>
     //!\endcond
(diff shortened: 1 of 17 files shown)


include/seqan3/range/view/take.hpp
```diff
@@ -437,90 +438,16 @@ struct take_fn
...
-        size_check(urange, target_size);
-        return {std::ranges::data(urange), target_size};
-    }
-
-    //!\overload
-    template <typename char_t, typename traits_t, typename alloc_t>
-    constexpr std::basic_string_view<char_t, traits_t>
-    operator()(std::basic_string<char_t, traits_t, alloc_t> const && urange, size_t target_size) = delete;
+        static_assert(std::ranges::ViewableRange<urng_t>,
+                      "The view::take adaptor can only be passed ViewableRanges, i.e. Views or & to non-View.");
```
smehringer: &-to-non-View

include/seqan3/range/view/drop.hpp
```diff
@@ -39,92 +39,61 @@ struct drop_fn
...
-    template <std::ranges::ViewableRange urng_t>
-    constexpr auto operator()(urng_t && urange, size_t drop_size) const noexcept
+    template <std::ranges::Range urng_t>
+    constexpr auto operator()(urng_t && urange, size_t drop_size) const
     {
-        auto b = std::ranges::begin(urange);
+        static_assert(std::ranges::ViewableRange<urng_t>,
+                      "The view::take adaptor can only be passed ViewableRanges, i.e. Views or & to non-View.");
 
+        [[maybe_unused]] size_t new_size = -1;
```
smehringer: below it's called `target_size` :nail_care:

test/unit/range/view/view_drop_test.cpp
```diff
@@ -93,8 +93,8 @@ void do_concepts(adaptor_t && adaptor)
     EXPECT_TRUE(std::ranges::View<decltype(v2)>);
     EXPECT_FALSE(std::ranges::SizedRange<decltype(v2)>);
     EXPECT_FALSE(std::ranges::CommonRange<decltype(v2)>);
-    EXPECT_TRUE(const_iterable_concept<decltype(v2)>); // because this is always subrange
-//     EXPECT_TRUE((std::ranges::OutputRange<decltype(v2), int>)); //TODO why fail?
+    EXPECT_FALSE(const_iterable_concept<decltype(v2)>);
+    EXPECT_TRUE((std::ranges::OutputRange<decltype(v2), int>)); //TODO why fail?
```
smehringer: remove "why fail" if it does not fail anymore

include/seqan3/range/view/drop.hpp
```diff
@@ -39,92 +39,61 @@ struct drop_fn
...
     }
 
-    /*!\brief       Call the view's constructor with the underlying view as argument.
-     * \returns     An instance of std::ranges::subrange.
+     /*!\brief Type erase if possible and forward to std::view::drop if not.
+     * \returns An instance of std::span, std::basic_string_view, std::ranges::subrange or std::ranges::drop_view.
      */
-    template <std::ranges::ViewableRange urng_t>
-    constexpr auto operator()(urng_t && urange, size_t drop_size) const noexcept
+    template <std::ranges::Range urng_t>
```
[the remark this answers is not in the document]
h-2: actual concepts checked inside now via static_assert, as we discussed.

include/seqan3/range/view/drop.hpp
```diff
@@ -39,92 +39,61 @@ struct drop_fn
...
+    constexpr auto operator()(urng_t && urange, size_t drop_size) const
     {
-        auto b = std::ranges::begin(urange);
+        static_assert(std::ranges::ViewableRange<urng_t>,
+                      "The view::take adaptor can only be passed ViewableRanges, i.e. Views or & to non-View.");
```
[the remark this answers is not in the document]
h-2: done

include/seqan3/range/view/take.hpp
h-2: done

h-2:
@rrahn I have two approvals and it's just fixes, but if you want to you can add another review :wink:

include/seqan3/range/detail/inherited_iterator_base.hpp
```diff
@@ -67,54 +71,74 @@ class inherited_iterator_base : public base_t
     constexpr inherited_iterator_base & operator=(inherited_iterator_base && rhs) = default;
     ~inherited_iterator_base() = default;
 
-    inherited_iterator_base(base_t it) : base_t{it} {}
+    //!\brief Delegate to base class if inheriting.
```
rrahn:
```suggestion
    //!\brief Delegate to base class if inheriting from non-pointer iterator.
```

include/seqan3/range/detail/inherited_iterator_base.hpp
```diff
@@ -67,54 +71,74 @@ class inherited_iterator_base : public base_t
...
+    //!\endcond
+        : base_t{std::move(it)}
+    {}
+
+    //!\brief Initialise member if deriving from pointer.
+    constexpr inherited_iterator_base(base_t it)
+    //!\cond
+        requires std::is_pointer_v<base_t>
+    //!\endcond
+    : member{std::move(it)}
```
rrahn: indentation

include/seqan3/range/detail/inherited_iterator_base.hpp
```diff
@@ -67,54 +71,74 @@ class inherited_iterator_base : public base_t
...
         requires std::StrictTotallyOrdered<base_t>
     //!\endcond
     {
-        return *this_to_base() > static_cast<base_t>(rhs);
+        return *this_to_base() > *rhs.this_to_base();
     }
 
-    constexpr bool operator<=(derived_t const & rhs) const noexcept(noexcept(base_t{} < base_t{}))
+    constexpr bool operator<=(derived_t const & rhs) const
+        noexcept(noexcept(std::declval<base_t &>() < std::declval<base_t &>()))
```
rrahn:
```suggestion
        noexcept(noexcept(std::declval<base_t &>() > std::declval<base_t &>()))
```

include/seqan3/range/detail/inherited_iterator_base.hpp
```diff
@@ -138,7 +162,7 @@ class inherited_iterator_base : public base_t
     }
 
     //!\brief Post-increment, return previous iterator state.
-    constexpr derived_t operator++(int) noexcept(noexcept(++base_t{}))
+    constexpr derived_t operator++(int) noexcept(noexcept(++std::declval<base_t &>()))
```
rrahn:
```suggestion
    constexpr derived_t operator++(int) noexcept(noexcept(std::declval<base_t &>()++))
```

include/seqan3/range/detail/inherited_iterator_base.hpp
```diff
@@ -159,7 +183,7 @@ class inherited_iterator_base : public base_t
     }
 
     //!\brief Post-decrement, return previous iterator state.
-    constexpr derived_t operator--(int) noexcept(noexcept(--base_t{}))
+    constexpr derived_t operator--(int) noexcept(noexcept(--std::declval<base_t &>()))
```
rrahn:
```suggestion
    constexpr derived_t operator--(int) noexcept(noexcept(std::declval<base_t &>()--))
```

include/seqan3/range/detail/inherited_iterator_base.hpp
rrahn: Note that post increment involves a copy of the iterator, which might throw. I think for this class it is enough to check if the post increment of the base class is noexcept because you store at most a pointer in this class. So copying this class will be noexcept if the base class is.

include/seqan3/range/detail/inherited_iterator_base.hpp
```diff
@@ -180,7 +204,7 @@ class inherited_iterator_base : public base_t
     }
 
     //!\brief Return a an iterator moved to the right.
-    constexpr derived_t operator+(difference_type const skip) const noexcept(noexcept(base_t{} += skip))
+    constexpr derived_t operator+(difference_type const skip) const noexcept(noexcept(std::declval<base_t &>() += skip))
```
rrahn:
```suggestion
    constexpr derived_t operator+(difference_type const skip) const noexcept(noexcept(std::declval<base_t &>() + skip))
```

include/seqan3/range/detail/inherited_iterator_base.hpp
```diff
@@ -190,7 +214,8 @@ class inherited_iterator_base : public base_t
     }
 
     //!\brief Non-member operator+ delegates to non-friend operator+.
-    constexpr friend derived_t operator+(difference_type const skip, derived_t const & it) noexcept(noexcept(base_t{} += skip))
+    constexpr friend derived_t operator+(difference_type const skip, derived_t const & it)
+        noexcept(noexcept(std::declval<base_t &>() += skip))
```
rrahn:
```suggestion
        noexcept(noexcept(std::declval<base_t &>() + skip))
```

include/seqan3/range/detail/inherited_iterator_base.hpp
```diff
@@ -208,7 +233,7 @@ class inherited_iterator_base : public base_t
     }
 
     //!\brief Return decremented copy of this iterator.
-    constexpr derived_t operator-(difference_type const skip) const noexcept(noexcept(base_t{} -= skip))
+    constexpr derived_t operator-(difference_type const skip) const noexcept(noexcept(std::declval<base_t &>() -= skip))
```
rrahn:
```suggestion
    constexpr derived_t operator-(difference_type const skip) const noexcept(noexcept(std::declval<base_t &>() - skip))
```

include/seqan3/range/detail/inherited_iterator_base.hpp
```diff
@@ -218,7 +243,8 @@ class inherited_iterator_base : public base_t
     }
 
     //!\brief Non-member operator- delegates to non-friend operator-.
-    constexpr friend derived_t operator-(difference_type const skip, derived_t const & it) noexcept(noexcept(base_t{} -= skip))
+    constexpr friend derived_t operator-(difference_type const skip, derived_t const & it)
+        noexcept(noexcept(std::declval<base_t &>() -= skip))
```
rrahn:
```suggestion
        noexcept(noexcept(std::declval<base_t &>() - skip))
```

include/seqan3/range/view/drop.hpp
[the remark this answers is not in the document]
rrahn: Did we discuss this? I can't remember this. I think it is related to some weird  compiler issues with the ranges library. This is fine with me, but the decision was made not made publicly available I guess?

include/seqan3/range/view/take.hpp
```diff
@@ -117,31 +117,31 @@ class view_take : public ranges::view_interface<view_take<urng_t, exactly, or_th
...
+        iterator_type & operator++() noexcept(noexcept(++std::declval<base_base_t &>()))
         {
-            base_base_t::operator++();
+            base_t::operator++();
             ++pos;
             return *this;
         }
 
-        iterator_type operator++(int) noexcept(noexcept(++base_base_t{}))
+        iterator_type operator++(int) noexcept(noexcept(++std::declval<base_base_t &>()))
```
```suggestion
        iterator_type operator++(int) noexcept(noexcept(++std::declval<base_base_t &>()++))
```

include/seqan3/range/view/take.hpp
```diff
@@ -117,31 +117,31 @@ class view_take : public ranges::view_interface<view_take<urng_t, exactly, or_th
...
         //!\endcond
         {
-            base_base_t::operator--();
+            base_t::operator--();
             --pos;
             return *this;
         }
 
-        iterator_type operator--(int) noexcept(noexcept(--base_base_t{}))
+        iterator_type operator--(int) noexcept(noexcept(--std::declval<base_base_t &>()))
```
```suggestion
        iterator_type operator--(int) noexcept(noexcept(--std::declval<base_base_t &>()--))
```

include/seqan3/range/view/take.hpp
```diff
@@ -536,6 +463,47 @@ struct take_fn
...
+
+        // string_view
+        if constexpr (is_type_specialisation_of_v<remove_cvref_t<urng_t>, std::basic_string_view>)
+        {
+            return urange.substr(0, target_size);
+        }
+        // string const
+        else if constexpr (is_type_specialisation_of_v<remove_cvref_t<urng_t>, std::basic_string> &&
+                           std::is_const_v<std::remove_reference_t<urng_t>> &&
+                           std::is_lvalue_reference_v<urng_t>)
```
rrahn: is_lvalue_reference does not make sense in this context, does it? The urng_t is a std::basic_string and cannot be a temporary by the viewable range concept? Or what I am missing?

include/seqan3/range/view/view_all.hpp
```diff
@@ -43,61 +44,51 @@ class all_fn : public adaptor_base<all_fn>
...
-    template <std::ranges::ViewableRange urng_t>
-    static auto impl(urng_t && urange)
-    {
-        return std::view::all(std::forward<urng_t>(urange));
+            return std::view::all(std::forward<urng_t>(urange));
+        }
+        // string const
+        else if constexpr (is_type_specialisation_of_v<remove_cvref_t<urng_t>, std::basic_string> &&
+                           std::is_const_v<std::remove_reference_t<urng_t>> &&
+                           std::is_lvalue_reference_v<urng_t>)
```
rrahn: same here. Do we really need std::is_lvalue_reference_v here?

include/seqan3/range/detail/inherited_iterator_base.hpp
rrahn: But, if the base_t is an input iterator Post increment might not copy and thus noexcept would still be wrong here if the copy constructor of this input iterator might throw :thinking: maybe you still need to check that also this is noexcept copyable

include/seqan3/range/detail/inherited_iterator_base.hpp
h-2: no, since it is implemented via pre-increment, the noexcept status depends also on pre-increment. However of the derived type.

include/seqan3/range/detail/inherited_iterator_base.hpp
h-2: see above.

include/seqan3/range/view/drop.hpp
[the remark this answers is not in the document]
h-2: I am pretty sure we discussed it quite a while ago. It's not related to different behaviour, it just makes errors more readable and reduces compile-time a bit.

include/seqan3/range/view/take.hpp
h-2: It's not required, but I thought it would help understand what's goinf on (that we really only have & types here). But I can remove it.

h-2: moved this to the documentation line above now

h-2:
@rrahn I have pushed changes and reworked some more stuff in the inherited iterator base. Please have a look again, but note that the noexcept specifier need not always refer to the respective noexcept operator on the base, but instead refer to what is actually going on (might refer to other operators of inherited_iterator_base or even the derived_t).

include/seqan3/range/view/take.hpp
rrahn: I think this was not resolved and need some similar strategy then the changes you made for the inherited_iterator_base class. So actually test the operations of the function and not what the base type would do. BTW I noticed that the constructor definitions are all marked noexcept explicitly. If I am not mistaken this depends on whether or not the base iterator is noexcept constructible since it is inheriting from it.

include/seqan3/range/view/take.hpp
rrahn: same here

include/seqan3/range/detail/inherited_iterator_base.hpp
rrahn: Right, but then the copyconstruction must be explicilty tested, which you did now  right?

include/seqan3/range/view/take.hpp
rrahn: Yeah, I am not sure it helps understanding because one immediately asks why this protection. Is there an overload I am not seeing? So I think it is better to remove it.

include/seqan3/range/detail/inherited_iterator_base.hpp
h-2: I changed this, please mark your comments as resolved, if they are resolved.

include/seqan3/range/detail/inherited_iterator_base.hpp
h-2: I am checking now if it is noexcept...copyable

include/seqan3/range/detail/inherited_iterator_base.hpp
h-2: yes, although it's not copy construction, it's construction-from-base_t

rrahn: Even better to test for it explicitly.

include/seqan3/range/detail/inherited_iterator_base.hpp
rrahn: this is so annoying that this is not automatically resolved 😒

include/seqan3/range/view/take.hpp
h-2: No, this only depends on the copy constructability of the iterator itself. I see no reason why the iterator should not be noexcept copy cosntructable, but I have added a check to make you happy.

rrahn: Because it implies that also all it's base classes need to be noexcept copy constructible. And while you enforce it for the derived class it is not enforced by the base class. And even worse, the base class inherits from another base class which is a template and we have no control over it. So copying the derived class will naturally call copy constructor of the base class which can possibly throw. So in the rare case that there is a potential throwing base iterator, we will get compiler errors with gcc9 wouldn't we? Or is there something that prevents this?

h-2:
I have now added a ridiculous amount of noexcept deduction. Can you please check if that's what you wanted and mark respective comments as resolved.

The tests still fail, I have no clue whats wrong.

include/seqan3/range/view/take.hpp
h-2: Inheriting constructors have an exception specification according to the stuff they are inheriting. If the base copy constructor is not noexcept, than the derived copy constructor won't be either:
https://stackoverflow.com/questions/18653726/is-the-default-move-constructor-defined-as-noexcept

rrahn: I think we are on totally different pages.
This is what I am referring to: https://godbolt.org/z/Ha-AEo
The iterator class in question has all it's constructors declared noexcept explicitly, while it's derived class (inherit_iterator_base) has not! This means, that we get a hard compiler error with gcc9 when the base_t iterator throws on construction (the one which comes from outside and we don't know about). So the explicit noexcept specification of this iterator class is wrong and should be just removed. This also means that there is no guarantee that post in/decrement if doing a copy of `this` is noexcept, unless we explicitly test for it. 

But anyway I will look into the code as soon as possible (like in the late evning).

include/seqan3/range/view/take.hpp
```diff
@@ -117,31 +117,33 @@ class view_take : public ranges::view_interface<view_take<urng_t, exactly, or_th
...
         {
-            base_base_t::operator++();
+            base_t::operator++();
             ++pos;
             return *this;
         }
 
-        iterator_type operator++(int) noexcept(noexcept(++base_base_t{}))
+        iterator_type operator++(int) noexcept(noexcept(++std::declval<iterator_type &>()) &&
+                                               noexcept(iterator_type{std::declval<iterator_type&>()}))
```
rrahn: Not so important but wouldn't `is_nothrow_copy_constructible` be also working here? 
In that case it would be more readable, but honestly I don't really care about this, so you can decide if you want to change this.

include/seqan3/range/view/take.hpp
h-2: You are absolutely right, I guess I was in the office for too long on Friday :wink:

include/seqan3/range/view/take.hpp
h-2: indeed

PR #884, opened 2019-04-08
[FIX] remove explicit and add const-qualification

h-2:
minor bug spotted when using this in lambda

### DIFF
--- a/include/seqan3/range/view/translation.hpp
+++ b/include/seqan3/range/view/translation.hpp
@@ -352,7 +352,7 @@ public:
 
     //!\brief Implicit conversion to container types.
     template <random_access_container_concept container_type>
-    explicit operator container_type()
+    operator container_type() const
     //!\cond
         requires std::is_same_v<aa27, value_type_t<container_type>>
     //!\endcond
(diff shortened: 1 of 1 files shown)


include/seqan3/range/view/translation.hpp
```diff
@@ -352,7 +352,7 @@ class view_translate_single
 
     //!\brief Implicit conversion to container types.
     template <random_access_container_concept container_type>
-    explicit operator container_type()
+    operator container_type() const
```
rrahn: Have we a made a decision about making these conversions implicit? The reason I am asking is, that this induces copying data if I put the view into a function that takes a suitable container as const & or &&. Not that this is bad, nor avoidable when dealing with legacy interfaces, but it happens unnoticed. 
Also there  is a proposal for explicit conversion using `ranges::to`, which is also now implemented in ranges library.

h-2: Yes, it's true that we need to discuss that, but I think it's something separate from this PR. All our views and currently also all range-v3 views are *implicitly* convertible and this view is also documented as being implicitly convertible while in reality it's marked as explicit.

rrahn: >Yes, it's true that we need to discuss that, but I think it's something separate from this PR. All our views and currently also all range-v3 views are implicitly convertible and this view is also documented as being implicitly convertible while in reality it's marked as explicit.

I made a discussion item for this. But it is ok for me to keep it consistent for now.

PR #901, opened 2019-04-12
[FIX] GCC8 bug workaround.

### LINKED ISSUE #893: UBSan failure in format_sam
http://cdash.seqan.de/viewBuildError.php?buildid=306756

smehringer:
Resolves #893 

A maybe-uninitialized error only occurs on GCC-8 in release with -fsanitize=undefined
The same code works fine with GCC7 (same flags) and with GCC8 in debug (also with -fsanitize=undefined)...

The following PR is a workaround that make the error vanish.

Do not merge yet, I'll file a bug report and add a link as a commment

### DIFF
--- a/include/seqan3/io/alignment_file/format_sam.hpp
+++ b/include/seqan3/io/alignment_file/format_sam.hpp
@@ -632,12 +632,16 @@ public:
         else if constexpr (detail::is_type_specialisation_of_v<std::remove_reference_t<decltype(get<0>(mate))>, std::optional>)
         {
             if (get<0>(mate).has_value())
-                write_range(stream_it, (header.ref_ids())[get<0>(mate).value()]);
+                // value_or(0) instead of value() (which is equivalent here) as a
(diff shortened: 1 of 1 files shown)


h-2:
Can you make the following out of this:
```cpp
#ifdef __GNUC__ 
// workaround ubsan false-positive in GCC: <link to tracker>
// remove when this is fixed

UGLY SOLUTION
#else

NICE SOLUTION
#endif
```

smehringer:
I have another "fix". One can replace value() by value_or() which does not trigger the warning and is in this case equivalent because `or` is never chosen as I have tested `has_value` before.... : 

```diff
- write_range(stream_it, header.ref_ids()[get<0>(mate).value()]);
+ write_range(stream_it, header.ref_ids()[get<0>(mate).value_or(0)]);
```
?

h-2:
Maybe deref instead of .value_or() works, too? In any case it would be a better workaround with value_or(), too. And I don't think you would need to add ifdefs then. Just a comment maybe.

smehringer:
> Maybe deref instead of .value_or() works, too?

No unfortunately not :( I tried that first

PR #903, opened 2019-04-12
[FIX] in DP alignment: keep old optimum if current cell score is equal

joergi-w:
I found an example where local alignments are extended too long, because the optimal score appeared again later and the algorithm chose the long version. This PR fixes this behaviour, such that always the shorter alignment is returned.


smehringer:
!rebuild

smehringer:
reopening because Jenkins is confused :roll_eyes:

smehringer:
I checked the code coverage which is confused as always

PR #857, opened 2019-03-28
Testing with MSys64-MinGW-gcc8.3

qPCR4vir:
MSys64-MinGW-gcc8.3 comes with a working filesystem. No more need of Boost (https://github.com/seqan/seqan3/pull/515).

Still some problems: 
```
template <typename T>
class istream : public ::testing::Test
{};

inline std::string const uncompressed{"The quick brown fox jumps over the lazy dog"};

TYPED_TEST_CASE_P(istream);

TYPED_TEST_P(istream, concept_check)
{
    EXPECT_TRUE((IStream<TypeParam, char>));
}
```
produce:
```
[ 44%] Built target gz_ostream_test
[ 44%] Building CXX object core/CMakeFiles/tuple_utility_test.dir/tuple_utility_test.cpp.obj
In file included from C:/Prog/ExtLib/cmake-seqan3-unit-build-release/vendor/googletest/googletest/include/gtest/gtest.h:67,
                 from C:\Prog\ExtLib\seqan3\test\unit\contrib\stream\gz_istream_test.cpp:8:
C:/Prog/ExtLib/cmake-seqan3-unit-build-release/vendor/googletest/googletest/include/gtest/gtest-typed-test.h:228:35: error: expected template-name before '<' token
   class TestName : public CaseName<gtest_TypeParam_> { \
                                   ^
C:/Prog/ExtLib/seqan3/test/unit/io/stream/istream_test_template.hpp:29:1: note: in expansion of macro 'TYPED_TEST_P'
 TYPED_TEST_P(IStream, concept_check)
 ^~~~~~~~~~~~
```
and with `aligned_alloc` in range: 
```
[100%] Building CXX object range/container/CMakeFiles/aligned_allocator_test.dir/aligned_allocator_test.cpp.obj
In file included from C:\Prog\ExtLib\seqan3\test\unit\range\container\aligned_allocator_test.cpp:10:
C:/Prog/ExtLib/seqan3/include/seqan3/range/container/aligned_allocator.hpp: In member function 'seqan3::aligned_allocator<value_t, alignment_v>::value_type* seqan3::aligned_allocator<value_t, alignment_v>::allocate(seqan3::aligned_allocator<value_t, alignment_v>::size_type)':
C:/Prog/ExtLib/seqan3/include/seqan3/range/container/aligned_allocator.hpp:122:48: error: 'aligned_alloc' is not a member of 'std'
         if (auto p = static_cast<pointer>(std::aligned_alloc(alignment, n * sizeof(value_type))))
[...]

PR #906, opened 2019-04-12
[FIX,TEST] Ignore OS flag in gzip compression

### LINKED ISSUE #723: Fix ostream test for macos
In the `gz_ostream_test.cpp` we write:

https://github.com/seqan/seqan3/blob/1264624bf6da2b7743587a23ee9eb7797ee13f02/test/unit/contrib/stream/gz_ostream_test.cpp#L20-L26

The first 10 bytes are flags:
```
'\x1f','\x8b','\x08','\x00','\x00','\x00','\x00','\x00','\x00','\x03'
  ID1    ID2    CM     FLG    MTIME  MTIME  MTIME  MTIME  XFL    OS
```

with the following meaning:

| Flag  | Meaning                          |
|-------|----------------------------------|
| ID1   | Magic byte, fixed to 1f          |
| ID2   | Magic byte, fixed to 8b          |
| CM    | Compression method, here DEFLATE |
| MTIME | 4 byte time stamp                |
| XFL   | extra flags                      |
| OS    | Operating System                 |

The operating system flag is set to 03 for unix.
When running it on MacOS it may be set to 03, 07 (for macos <10) or 13 (macos > 10).

So I think we shouldn't check the 'correctness' of the OS flag?

eseiler:
Resolves #723 
With this PR the unit tests do not fail anymore on my machine (mac).

I don't like checking the substrings in, e.g., `compression_by_filename_impl` for both gzip and bz2 compression since it is not necessary for bz2.
@h-2 Any preferred way to fix this?

### DIFF
--- a/test/unit/io/sequence_file/sequence_file_output_test.cpp
+++ b/test/unit/io/sequence_file/sequence_file_output_test.cpp
@@ -475,7 +475,9 @@ void compression_by_filename_impl(test::tmp_filename & filename, std::string_vie
         buffer = std::string{std::istreambuf_iterator<char>{fi}, std::istreambuf_iterator<char>{}};
     }
 
-    EXPECT_EQ(buffer, expected);
+    // Ignore the OS flag
+    EXPECT_EQ(buffer.substr(0,9), expected.substr(0,9));
+    EXPECT_EQ(buffer.substr(11), expected.substr(11));
 }
 
 template <typename comp_stream_t>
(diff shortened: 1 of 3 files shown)


smehringer:
What happens with the current test on windows?

test/unit/io/sequence_file/sequence_file_output_test.cpp
```diff
@@ -475,7 +475,9 @@ void compression_by_filename_impl(test::tmp_filename & filename, std::string_vie
         buffer = std::string{std::istreambuf_iterator<char>{fi}, std::istreambuf_iterator<char>{}};
     }
 
-    EXPECT_EQ(buffer, expected);
+    // Ignore the OS flag
+    EXPECT_EQ(buffer.substr(0,9), expected.substr(0,9));
```
smehringer: Addressing the issue @eseiler mentioned that the weird OS byte is only set by the gzip format but not bzip I would remove the EXPECT equalities here from the impl() function and instead test equality in the individual tests where we can test for exact equality for bzip and "exact without the 11th byte" for gzip

h-2:
> What happens with the current test on windows?

Same Problem on windows: #857

eseiler:
Superseded by #1014

Issue #89, opened 2017-07-20
Argument parser - export of file format (ctd) for KNIME engine 


smehringer:
Can be partly adapted by seqan2.

smehringer:
Will be reopened if planned for a sprint

SGSSGene:
We moved the argument parser into its own repo: [sharg-parser](https://github.com/seqan/sharg-parser)

This feature is being implemented by this PR here https://github.com/seqan/sharg-parser/pull/94

PR #907, opened 2019-04-13
Feature: input and output file and directory validator

rrahn:
These validators replace the file_ext and path_exists validator types using specific validator for input and output files and for input and output directories. 
These changes are necessary to determine for workflow systems, whether we have an input port or and output port that can be chained with other nodes in the workflow.

I first added the file and directory validators then adapted the tutorials and finally removed all instances of the path_exists_validator and file_ext_validator. 
I also removed for now the feature to switch between case sensitive and case insensitive comparison and always do case insensitive. This is due to possible other flags that might be added and to keep the interface as lenient as possible. I further have not found an example where the the file format depends on the case of the file extension. Which would be also a very doubtful feature I guess.

Partially addresses #89


include/seqan3/argument_parser/validators.hpp
```diff
@@ -235,69 +236,87 @@ value_list_validator(std::vector<const char *> const & v) -> value_list_validato
...
- *
- * \note The validator works on every type that can be implicitly cast to std::filesystem::path.
+ * This class provides a common interface for seqan3::input_file_validator and the seqan3::output_file_validator as
+ * well as the seqan3::input_directory_validator and seqan3::output_directory_validator.
+ * This class is a CRTP-base class and cannot be instantiated directly. Furthermore, this is an exposition-only class.
+ * Please, do not use this class to derive from it as its interface might change without notice.
  */
-class file_ext_validator
+template <typename derived_t>
+class path_validator_base
```
smehringer: Why not normal polymorphy? (provide a protected `check_extension` member function that can be called when defining the `operator()` in the derived classes, then you do not need impl). I think CRTP is a little over-engineering for validators.. you even need to document explicitily to not use it. I'm not sure that it saves so much lines of code. I would even only provide a `check_extension` detail function and make no inheritance.

any way the base class should be detail?

include/seqan3/argument_parser/validators.hpp
```diff
@@ -235,69 +236,87 @@ value_list_validator(std::vector<const char *> const & v) -> value_list_validato
...
+        static_cast<derived_t const &>(*this).validate_path_impl(path);
+
+        // If no valid extensions are given we can safely return here.
+        if (extensions.empty())
+            return;
+
+        // Check if extension is available.
+        if (!path.has_extension())
+            throw parser_invalid_argument(detail::to_string("The given filename ", path.string(),
+                                                            " has no extension."));
```
```suggestion
                                                            " has no extension but should have one of ... foo foo foo."));
```

include/seqan3/argument_parser/validators.hpp
```diff
@@ -235,69 +236,87 @@ value_list_validator(std::vector<const char *> const & v) -> value_list_validato
...
+
+        // Drop the dot.
+        std::string tmp_str = path.extension().string();
+        auto drop_less_ext = tmp_str | view::drop(1);
+
+        // Compares the extensions in lower case.
+        auto cmp_lambda = [&] (std::string const & cmp)
+        {
+            return std::ranges::equal(cmp | view::to_lower, drop_less_ext | view::to_lower);
```
```suggestion
            return std::ranges::equal(cmp | view::to_lower, tmp_str | view::drop(1) | view::to_lower);
```
and remove the extra view copy in line 300.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -235,69 +236,87 @@ value_list_validator(std::vector<const char *> const & v) -> value_list_validato
...
+        };
+
+        // Check if requested extension is present.
+        if (auto res = std::find_if(extensions.begin(), extensions.end(), cmp_lambda); res == extensions.end())
+        {
+            throw parser_invalid_argument(detail::to_string("Expected on of the following valid extensions: ",
+                                                             std::view::all(extensions), ". Got ",
```
```suggestion
                                                             extensions, ". Got ",
```
This should be handled by to-string automatically

include/seqan3/argument_parser/validators.hpp
```diff
@@ -235,69 +236,87 @@ value_list_validator(std::vector<const char *> const & v) -> value_list_validato
...
+                                                             std::string{std::ranges::begin(drop_less_ext),
```
```suggestion
                                                             drop_less_ext,
```
This should be handles by to_string automatically

include/seqan3/argument_parser/validators.hpp
```diff
@@ -309,68 +328,260 @@ class file_ext_validator
...
+    std::vector<std::string> extensions{};
+};
+
+/*!\brief A validator that checks if a given path is a valid input file.
+ * \ingroup argument_parser
+ * \implements seqan3::validator_concept
+ *
+ * \details
+ *
+ * On construction, the validator must receive a list (std::vector over std::string) of valid file extensions.
```
```suggestion
 * On construction, the validator may receive a list (std::vector over std::string) of valid file extensions.
```

You document that it **must** receive a list, but you provide a default constructor.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -309,68 +328,260 @@ class file_ext_validator
...
+     */
+    input_file_validator(std::vector<std::string> extensions) : base_t{std::move(extensions)}
+    {}
+    //!\}
+
     //!\brief Returns a message that can be appended to the (positional) options help page info.
     std::string get_help_page_message() const
     {
-        return detail::to_string("File name extension must be one of ", std::view::all(extensions), ".");
+        return detail::to_string("Input file formats: ", base_t::extensions | std::view::join(std::string{", "}), ".");
```
```suggestion
        return detail::to_string("Valid input file formats: ", base_t::extensions | std::view::join(std::string{", "}), ". The file is checked for existance.");
```

include/seqan3/argument_parser/validators.hpp
```diff
@@ -309,68 +328,260 @@ class file_ext_validator
...
 
-    //!\brief True if file extension is case sensitive
-    bool case_sensitive;
+    /*!\brief Tests whether file exists.
+     * \param path The input value to check.
+     * \throws parser_invalid_argument
+     */
+    void validate_path_impl(std::filesystem::path const & path) const
+    {
+        if (!std::filesystem::exists(path))
```
smehringer: Shouldn't we also check that it is readable?

include/seqan3/argument_parser/validators.hpp
```diff
@@ -235,69 +236,87 @@ value_list_validator(std::vector<const char *> const & v) -> value_list_validato
...
+        // Compares the extensions in lower case.
+        auto cmp_lambda = [&] (std::string const & cmp)
+        {
+            return std::ranges::equal(cmp | view::to_lower, drop_less_ext | view::to_lower);
+        };
+
+        // Check if requested extension is present.
+        if (auto res = std::find_if(extensions.begin(), extensions.end(), cmp_lambda); res == extensions.end())
+        {
+            throw parser_invalid_argument(detail::to_string("Expected on of the following valid extensions: ",
```
```suggestion
            throw parser_invalid_argument(detail::to_string("Expected one of the following valid extensions: ",
```

include/seqan3/argument_parser/validators.hpp
```diff
@@ -235,69 +236,87 @@ value_list_validator(std::vector<const char *> const & v) -> value_list_validato
...
+        auto drop_less_ext = tmp_str | view::drop(1);
+
+        // Compares the extensions in lower case.
+        auto cmp_lambda = [&] (std::string const & cmp)
+        {
+            return std::ranges::equal(cmp | view::to_lower, drop_less_ext | view::to_lower);
+        };
+
+        // Check if requested extension is present.
+        if (auto res = std::find_if(extensions.begin(), extensions.end(), cmp_lambda); res == extensions.end())
```
```suggestion
        if (std::find_if(extensions.begin(), extensions.end(), cmp_lambda) == extensions.end())
```
o.0 ?

include/seqan3/argument_parser/validators.hpp
```diff
@@ -309,68 +328,260 @@ class file_ext_validator
...
+private:
+
+    //!\brief The type of the base class.
+    using base_t = path_validator_base<input_directory_validator>;
+
+    //!\brief Befriends the base class.
+    friend base_t;
+
+    //!\brief Inherit the base constructor.
+    using base_t::base_t;
```
smehringer: why ?

include/seqan3/argument_parser/validators.hpp
```diff
@@ -309,68 +328,260 @@ class file_ext_validator
...
+
+private:
+
+    /*!\brief Checks if directory is valid.
+     * \param path The directory name to check.
+     * \throws parser_invalid_argument
+     */
+    void validate_path_impl(std::filesystem::path const & path) const
+    {
+        if (path.has_extension())
```
smehringer: Shouldn't an output directory also exist?

include/seqan3/argument_parser/validators.hpp
rrahn: As discussed, I changed it now to an abstract base class with virtual function dispatching.

include/seqan3/argument_parser/validators.hpp
rrahn: I reuse the drop_less_ext later in the error message.

include/seqan3/argument_parser/validators.hpp
rrahn: I had some issues, but now it is working.

include/seqan3/argument_parser/validators.hpp
rrahn: Copy'n'paste issue

include/seqan3/argument_parser/validators.hpp
rrahn: Resolved by the new implementation.

include/seqan3/argument_parser/validators.hpp
rrahn: Well, now the output directory might or might not exists. It will be checked in both cases if it is writable.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -235,57 +237,48 @@ value_list_validator(std::vector<const char *> const & v) -> value_list_validato
...
-
-    /*!\brief Tests whether the filepath \p path ends with a valid extension.
-     * \param path The input value to check.
-     * \throws parser_invalid_argument
+    file_validator_base() = default;                                        //!< Defaulted.
+    file_validator_base(file_validator_base const &) = default;             //!< Defaulted.
+    file_validator_base(file_validator_base &&) = default;                  //!< Defaulted.
+    file_validator_base & operator=(file_validator_base const &) = default; //!< Defaulted.
+    file_validator_base & operator=(file_validator_base &&) = default;      //!< Defaulted.
+    virtual ~file_validator_base() = default;                               //!< Virtual destructor.
```
smehringer: why?

doc/tutorial/argument_parser/index.md
```diff
@@ -382,24 +382,46 @@ Add a seqan3::value_list_validator to the `-a/--aggregate-by` option that sets t
...
+### The file validator
 
-On construction, the validator receives a list (vector) of valid file extensions.
-The validator throws a seqan3::parser_invalid_argument exception whenever a given filename (string) is not in the given extension list.
+SeqAn3 offers two file validator types: the seqan3::input_file_validator and the seqan3::output_file_validator.
+On construction, the validator receives a list (vector) of valid file extensions that are tested against the extension
+of the parsed option value.
+The validator throws a seqan3::parser_invalid_argument exception whenever a given filename's extension is not in the given list of valid extensions or if the specified path does not name a file but a directory. In addition, the
+seqan3::input_file_validator checks if the file already exists, while on the opposite the seqan3::output_file_validator
+checks if the file does not already exists in order to prevent overwriting an already existing file.
```
smehringer: Note: If you want to allow any extension just pass an empty list to the validator.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -297,68 +290,347 @@ class file_ext_validator
...
+class input_file_validator : public file_validator_base
+{
+public:
+    //!\brief Type of values that are tested by the validator.
+    using file_validator_base::value_type;
+
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    input_file_validator() = default;                                         //!< Defaulted.
```
smehringer: I thought in dynamic polymorphic constructors are inherited? (or at least importable by `using`)

include/seqan3/argument_parser/validators.hpp
```diff
@@ -297,68 +290,347 @@ class file_ext_validator
...
+
+        // Make sure the created dir is deleted after we are done.
+        std::unique_ptr<std::filesystem::path const, decltype(deleter)> raii{&dir, deleter};
+
+        std::error_code ec;
+        std::filesystem::create_directory(dir, ec); // does nothing and is not treated as error if path already exists.
+        // if error code was set or if dummy.txt could not be created within the output dir, throw an error.
+        if (static_cast<bool>(ec) || !path_is_writable(dir / "dummy.txt"))
+        {
+            throw parser_invalid_argument(detail::to_string("Cannot create directory: ", dir, "!"));
```
```suggestion
            throw parser_invalid_argument(detail::to_string("Directory: ", dir, " is not writable!"));
```

include/seqan3/argument_parser/validators.hpp
```diff
@@ -297,68 +290,347 @@ class file_ext_validator
...
+        }
     }
 
     //!\brief Returns a message that can be appended to the (positional) options help page info.
     std::string get_help_page_message() const
     {
-        return detail::to_string("The file or directory is checked for existence.");
+        return detail::to_string("An output directory.");
```
```suggestion
        return detail::to_string("The output directory must be writable.");
```

test/snippet/argument_parser/validators_chaining.cpp
```diff
@@ -9,10 +9,10 @@ int main(int argc, const char ** argv)
     std::string file_name;
 
     seqan3::regex_validator absolute_path_validator{"(/[^/]+)+/.*\\.[^/\\.]+$"};
-    seqan3::file_ext_validator my_file_ext_validator{{"sa", "so"}};
+    seqan3::input_file_validator my_file_ext_validator{{"sa", "so"}};
 
-    myparser.add_option(file_name, 'f', "file","Give me a file name/path.",
-                        seqan3::option_spec::DEFAULT, absolute_path_validator | my_file_ext_validator);
+    myparser.add_option(file_name, 'f', "file","Give me a file name with absolute path.",
+                        seqan3::option_spec::DEFAULT, absolute_path_validator | file_validator);
```
smehringer: file_validator?

test/snippet/argument_parser/validators_input_directory.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+int main(int argc, const char ** argv)
+{
+    seqan3::argument_parser myparser("Test", argc, argv); // initialize
+
+    //! [validator_call]
+    std::filesystem::path myfile;
+
+    myparser.add_option(myfile,'d',"dir","The directory containing the input files.",
+                        seqan3::option_spec::DEFAULT, seqan3::input_directory_validator());
```
smehringer:
```suggestion
                        seqan3::option_spec::DEFAULT, seqan3::input_directory_validator{});
```

test/snippet/argument_parser/validators_input_directory.cpp
```diff
@@ -0,0 +1,31 @@
...
+    {
+        myparser.parse();
+    }
+    catch (seqan3::parser_invalid_argument const & ext) // the user did something wrong
+    {
+        std::cerr << "[PARSER ERROR] " << ext.what() << "\n"; // customize your error message
+        return -1;
+    }
+
+    seqan3::debug_stream << "filename given by user passed validation: " << myfile << "\n";
```
smehringer:
```suggestion
    seqan3::debug_stream << "directory given by user passed validation: " << myfile << "\n";
```

test/snippet/argument_parser/validators_input_file.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+    //! [validator_call]
+    std::filesystem::path myfile;
+
+    myparser.add_option(myfile,'f',"file","The input file containing the sequences.",
+                        seqan3::option_spec::DEFAULT, seqan3::input_file_validator({"fa","fasta"}));
+    //! [validator_call]
+
+    // an exception will be thrown if the user specifies a filename
+    // that does not have one of the extensions ["fa","fasta"]
```
smehringer:
```suggestion
    // that does not have one of the extensions ["fa","fasta"] or if the file does not exist/is not readable.
```

test/snippet/argument_parser/validators_output_directory.cpp
```diff
@@ -0,0 +1,31 @@
...
+    {
+        myparser.parse();
+    }
+    catch (seqan3::parser_invalid_argument const & ext) // the user did something wrong
+    {
+        std::cerr << "[PARSER ERROR] " << ext.what() << "\n"; // customize your error message
+        return -1;
+    }
+
+    seqan3::debug_stream << "filename given by user passed validation: " << myfile << "\n";
```
smehringer:
```suggestion
    seqan3::debug_stream << "directory given by user passed validation: " << myfile << "\n";
```

test/snippet/argument_parser/validators_output_directory.cpp
```diff
@@ -0,0 +1,31 @@
...
+    seqan3::argument_parser myparser("Test", argc, argv); // initialize
+
+    //! [validator_call]
+    std::filesystem::path myfile;
+
+    myparser.add_option(myfile,'d',"dir","The output directory for storing the files.",
+                        seqan3::option_spec::DEFAULT, seqan3::output_directory_validator());
+    //! [validator_call]
+
+    // an exception will be thrown if the user specifies a directory name without trailing directory separator.
```
smehringer: you sure? I thought the is_directory checks for an extension?
also add a comment that is has to be writable

test/snippet/argument_parser/validators_output_file.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+    //! [validator_call]
+    std::filesystem::path myfile;
+
+    myparser.add_option(myfile,'f',"file","Output file containing the processed sequences.",
+                        seqan3::option_spec::DEFAULT, seqan3::output_file_validator({"fa","fasta"}));
+    //! [validator_call]
+
+    // an exception will be thrown if the user specifies a filename
+    // that does not have one of the extensions ["fa","fasta"]
```
smehringer:
```suggestion
    // that does not have one of the extensions ["fa","fasta"] or does already exist
```

test/unit/argument_parser/format_parse_validators_test.cpp
```diff
@@ -47,226 +47,399 @@ TEST(validator_test, fullfill_concept)
...
+
+    { // single file
+
+        { // empty list of file.
+            input_file_validator my_validator{};
+            EXPECT_NO_THROW(my_validator(tmp_name.get_path()));
+        }
+
+        { // file already exists.
+            std::filesystem::path does_not_exists{tmp_name.get_path()};
```
smehringer:
```suggestion
            std::filesystem::path does_not_exist{tmp_name.get_path()};
```
:nail_care:

test/unit/argument_parser/format_parse_validators_test.cpp
```diff
@@ -47,226 +47,399 @@ TEST(validator_test, fullfill_concept)
...
+            does_not_exists.replace_extension(".bam");
+            input_file_validator my_validator{formats};
+            EXPECT_THROW(my_validator(does_not_exists), parser_invalid_argument);
+        }
+
+        { // file has wrong format.
+            input_file_validator my_validator{std::vector{std::string{"sam"}}};
+                EXPECT_THROW(my_validator(tmp_name.get_path()), parser_invalid_argument);
```
smehringer:
```suggestion
            EXPECT_THROW(my_validator(tmp_name.get_path()), parser_invalid_argument);
```

doc/tutorial/argument_parser/small_snippets.cpp
```diff
@@ -72,10 +72,10 @@ parser.add_option(required_variable, 'n', "name", "I really need a name.", optio
 
 {
 argument_parser parser("Example Parser", argc, argv);
-//![path_existence_validator]
+//![input_file_validator]
 parser.add_positional_option(args.file_path, "Please provide a tab separated data file.",
-                             path_existence_validator{});
-//![path_existence_validator]
+                             input_file_validator{{"tsv"}});
```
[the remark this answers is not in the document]
rrahn: valid point. It should be convertible to path without.

doc/tutorial/argument_parser/index.md
```diff
@@ -382,24 +382,46 @@ Add a seqan3::value_list_validator to the `-a/--aggregate-by` option that sets t
...
-\snippet test/snippet/argument_parser/validators_path_existence.cpp validator_call
+\snippet test/snippet/argument_parser/validators_output_file.cpp validator_call
+
+### The directory validator
+
+In addition to the file validator types, SeqAn3 offers directory validator types. These are useful if one needs
+to provide an input directory (using the seqan3::input_directory_validator) or output directory
+(using the seqan3::output_directory_validator) where multiple files need to be read from or written too.
+The validators throw a seqan3::parser_invalid_argument exception whenever the parsed path contains a filename with an 
+extension. The seqan3::input_directory_validator also tests if the directory already exists and throws if this is not
```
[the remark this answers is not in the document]
rrahn: In general I would say no. If you specify an output directory the program might as well create one for you. Like the output prefix we had before, just that the prefix now is the directory. The parent path however needs to exist. I think it should then be the responsibility of the programmer to take care of creating the output directory.

include/seqan3/argument_parser/validators.hpp
rrahn: Well, that's the world of runtime polymorphism. 

> Deleting a derived class object using a pointer to a base class that has a non-virtual destructor results in undefined behavior. To correct this situation, the base class should be defined with a virtual destructor.

include/seqan3/argument_parser/validators.hpp
rrahn: That is true, but I am merely following our style guide rules here. Since it is not an aggregate all constructors need to be explicitly defined. I am importing the base class constructor anyhow. (See below).

include/seqan3/argument_parser/validators.hpp
rrahn: Well, I had this before, but I think it is more correct to say that it cannot be created. Not writable suggests that there are not enough permissions. But if the parent path does not exist create directory will fail. Hence, I used the more generic error message

include/seqan3/argument_parser/validators.hpp
rrahn: What about: `A valid path for the output directory.` 
I feel this fits better to describe the limit of this option. I mean output directory implies that it is writable.

test/snippet/argument_parser/validators_output_directory.cpp
rrahn: No it needs to be adapted. Good catch

doc/tutorial/argument_parser/index.md
```diff
@@ -382,24 +382,46 @@ Add a seqan3::value_list_validator to the `-a/--aggregate-by` option that sets t
...
 
-\snippet test/snippet/argument_parser/validators_path_existence.cpp validator_call
+\snippet test/snippet/argument_parser/validators_output_file.cpp validator_call
+
+### The directory validator
+
+In addition to the file validator types, SeqAn3 offers directory validator types. These are useful if one needs
+to provide an input directory (using the seqan3::input_directory_validator) or output directory
+(using the seqan3::output_directory_validator) where multiple files need to be read from or written too.
+The validators throw a seqan3::parser_invalid_argument exception whenever the parsed path contains a filename with an
```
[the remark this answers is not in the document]
rrahn: I changed it and also made it more fitting from the description to the current implementation

doc/tutorial/argument_parser/small_snippets.cpp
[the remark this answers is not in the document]
rrahn: Oh, I thought this is a different thing. No I either explicitly give a std::vector or I use the initializer list constructor like above.

include/seqan3/argument_parser/validators.hpp
smehringer: I thought our rule was: if you define one, define all. But you can include them all with one `using` statement because you do not define a new one..

rrahn: No this is not our rule. Our rule is if it is not an aggregate class, then specify all. 
Note that importing the base constructors doesn't change the fact that the compiler will generate the constructors of the derived class implicitly. So whether I add them or not. They will be created, but won't pop up in doxygen.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -235,57 +237,48 @@ value_list_validator(std::vector<const char *> const & v) -> value_list_validato
...
+     * \param path The path to validate.
+     *
+     * \details
+     *
+     * This is a pure virtual function and must be overloaded by the derived class.
+     */
+    virtual void operator()(std::filesystem::path const & path) const = 0;
 
-    /*!\brief Tests whether every value of v lies inside extensions.
+    /*!\brief Tests whether every path in list \p v exists.
```
```suggestion
    /*!\brief Tests whether every path in list \p v passes validation. See operator()(value_type const & value) for further information.
```

include/seqan3/argument_parser/validators.hpp
```diff
@@ -297,68 +290,347 @@ class file_ext_validator
...
+         std::for_each(v.begin(), v.end(), [&] (auto cmp) { this->operator()(cmp); });
+    }
+
+protected:
+
+    /*!\brief Validates the given filename path based on the specified extensions.
+     * \param path The filename path.
+     * \throws parser_invalid_argument if the specified extensions don't match the given path.
+     */
+    void validate_path_impl(std::filesystem::path const & path) const
```
smehringer: I know this is picky, but for me a path is either a directory or a filename. In this function you except a filename so maybe it should be called validate_filename_impl?

include/seqan3/argument_parser/validators.hpp
```diff
@@ -297,68 +290,347 @@ class file_ext_validator
...
+        if (std::filesystem::is_directory(path))
+        {
+            std::error_code ec{};
+            std::filesystem::directory_iterator{path, ec};  // if directory iterator cannot be created, ec will be set.
+            return !static_cast<bool>(ec); // test the error code.
+        }
+
+        // Must be a regular file.
+        if (!std::filesystem::is_regular_file(path))
+            throw std::invalid_argument{detail::to_string("Expected a regular file!", path, "!")};
```
smehringer: hm this error message is not helpful because one does not know what a "regular" file is. maybe reference std::filesystem?

include/seqan3/argument_parser/validators.hpp
```diff
@@ -297,68 +290,347 @@ class file_ext_validator
...
+
+        std::ifstream file{path};
+        return file.is_open() && file.good();
+    }
+
+    /*!\brief Checks if the given path is writable.
+     * \param path The path to check.
+     * \returns `true` if writable, otherwise `false`.
+     */
+    bool path_is_writable(std::filesystem::path const & path) const
```
smehringer: path_is_readable can be used for input file and directory but path_is_writable can only be used for an output file, correct? Then this should also be called filename_is_writable?

include/seqan3/argument_parser/validators.hpp
```diff
@@ -297,68 +290,347 @@ class file_ext_validator
...
+     * \{
+     */
+    input_file_validator() = default;                                         //!< Defaulted.
+    input_file_validator(input_file_validator const &) = default;             //!< Defaulted.
+    input_file_validator(input_file_validator &&) = default;                  //!< Defaulted.
+    input_file_validator & operator=(input_file_validator const &) = default; //!< Defaulted.
+    input_file_validator & operator=(input_file_validator &&) = default;      //!< Defaulted.
+    ~input_file_validator() = default;                                        //!< Defaulted.
+
+    //!\brief Import constructor for setting valid extensions.
```
smehringer: I think using declarations must not to be documented because the documentation is copied and an extra brief might confuse doxygen. (at least it was like that with member typedefs when I last checked)

include/seqan3/argument_parser/validators.hpp
```diff
@@ -297,68 +290,347 @@ class file_ext_validator
...
+        // Check extension.
+        file_validator_base::validate_path_impl(file);
     }
 
     //!\brief Returns a message that can be appended to the (positional) options help page info.
     std::string get_help_page_message() const
     {
-        return detail::to_string("File name extension must be one of ", std::view::all(extensions), ".");
+        return detail::to_string("Valid input file formats: ",
+                                 file_validator_base::extensions | std::view::join(std::string{", "}),
```
smehringer: If do not use `join`, the list is displayed as `Valid input file formats: [fasta, fa]` which is sufficient isn't it?

include/seqan3/argument_parser/validators.hpp
```diff
@@ -297,68 +290,347 @@ class file_ext_validator
...
+ *
+ * \note The validator works on every type that can be implicitly converted to std::filesystem::path.
  */
-class path_existence_validator
+class input_directory_validator : public file_validator_base
 {
 public:
-    //!\brief Type of values that are tested by validator
-    using value_type = std::string;
+    //!\brief Type of values that are tested by the validator.
```
smehringer: remove brief as it will be copied from base by doxygen

include/seqan3/argument_parser/validators.hpp
```diff
@@ -297,68 +290,347 @@ class file_ext_validator
...
+    //!\brief Inherit base constructor.
+    using file_validator_base::file_validator_base;
+    //!\}
+
+    // Import the base::operator()
+    using file_validator_base::operator();
+
+    /*!\brief Tests whether path is an existing directory and is readable.
+     * \param dir The input value to check.
+     * \throws parser_invalid_argument if the validation process failed or std::filesystem::filesystem_error if some
```
```suggestion
     * \throws seqan3::parser_invalid_argument if the validation process failed or std::filesystem::filesystem_error if some
```

include/seqan3/argument_parser/validators.hpp
```diff
@@ -297,68 +290,347 @@ class file_ext_validator
...
 
-    /*!\brief Tests whether every path (file or directory) in list \p v exists.
-     * \tparam range_type The type of range to check; must model std::ranges::ForwardRange and the value type must
-     *                    be convertible to std::filesystem::path.
-     * \param  v          The input range to iterate over and check every element.
-     * \throws parser_invalid_argument
+    //!\brief Returns a message that can be appended to the (positional) options help page info.
+    std::string get_help_page_message() const
+    {
+        return detail::to_string("An existing path for the input directory.");
```
```suggestion
        return detail::to_string("An existing, readable path for the input directory.");
```

include/seqan3/argument_parser/validators.hpp
```diff
@@ -297,68 +290,347 @@ class file_ext_validator
...
+     * \param dir The input value to check.
+     * \throws parser_invalid_argument if the validation process failed or std::filesystem::filesystem_error if some
+     *         unexpected errors during the filesystem operations occur.
+     */
+    void operator()(std::filesystem::path const & dir) const override
     {
-         std::for_each(v.begin(), v.end(), [&] (auto cmp) { (*this)(cmp); });
+        std::function<void(std::filesystem::path const *)> deleter;
+        if (std::filesystem::exists(dir))
+            deleter = [] (std::filesystem::path const *) { /* no-op */ };
```
smehringer: isn't this a bit over the top? 
Couldn't you just do
```
bool existed = std::filesystem::exists(dir);

//... create dir and dummy file

if (!existed)
    std::filesystem::remove(*p, ec);
```
?

test/snippet/argument_parser/validators_chaining.cpp
```diff
@@ -9,9 +9,9 @@ int main(int argc, const char ** argv)
     std::string file_name;
 
     seqan3::regex_validator absolute_path_validator{"(/[^/]+)+/.*\\.[^/\\.]+$"};
-    seqan3::file_ext_validator my_file_ext_validator{{"sa", "so"}};
+    seqan3::input_file_validator my_file_ext_validator{{"sa", "so"}};
 
-    myparser.add_option(file_name, 'f', "file","Give me a file name/path.",
+    myparser.add_option(file_name, 'f', "file","Give me a file name with absolute path.",
```
smehringer:
```suggestion
    myparser.add_option(file_name, 'f', "file","Give me a file name with an absolute path.",
```

test/snippet/argument_parser/validators_input_directory.cpp
```diff
@@ -0,0 +1,32 @@
...
+#include <seqan3/argument_parser/all.hpp>
+#include <seqan3/io/stream/debug_stream.hpp>
+#include <seqan3/std/filesystem>
+
+int main(int argc, const char ** argv)
+{
+    seqan3::argument_parser myparser("Test", argc, argv); // initialize
+
+    //! [validator_call]
+    std::filesystem::path myfile;
```
smehringer:
```suggestion
    std::filesystem::path mydir{};
```
always initialize values for the parser! otherwise the default values printed on the help page are weird.

test/snippet/argument_parser/validators_input_file.cpp
```diff
@@ -5,16 +5,17 @@
 
 int main(int argc, const char ** argv)
 {
-    seqan3::argument_parser myparser{"Test", argc, argv}; // initialize
+    seqan3::argument_parser myparser("Test", argc, argv); // initialize
 
-    //![validator_call]
+    //! [validator_call]
     std::filesystem::path myfile;
```
smehringer:
```suggestion
    std::filesystem::path myfile{};
```

test/snippet/argument_parser/validators_output_directory.cpp
```diff
@@ -0,0 +1,32 @@
...
+#include <seqan3/argument_parser/all.hpp>
+#include <seqan3/io/stream/debug_stream.hpp>
+#include <seqan3/std/filesystem>
+
+int main(int argc, const char ** argv)
+{
+    seqan3::argument_parser myparser("Test", argc, argv); // initialize
+
+    //! [validator_call]
+    std::filesystem::path myfile;
```
smehringer:
```suggestion
    std::filesystem::path mydir{};
```

test/snippet/argument_parser/validators_output_file.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include <seqan3/argument_parser/all.hpp>
+#include <seqan3/io/stream/debug_stream.hpp>
+#include <seqan3/std/filesystem>
+
+int main(int argc, const char ** argv)
+{
+    seqan3::argument_parser myparser("Test", argc, argv); // initialize
+
+    //! [validator_call]
+    std::filesystem::path myfile;
```
smehringer:
```suggestion
    std::filesystem::path myfile{};
```

doc/tutorial/argument_parser/index.md
```diff
@@ -382,24 +382,51 @@ Add a seqan3::value_list_validator to the `-a/--aggregate-by` option that sets t
...
+### The directory validator
+
+In addition to the file validator types, SeqAn3 offers directory validator types. These are useful if one needs
+to provide an input directory (using the seqan3::input_directory_validator) or output directory
+(using the seqan3::output_directory_validator) where multiple files need to be read from or written to.
+The seqan3::input_directory_validator checks whether the specified path is a directory and is readable.
+Similar, the seqan3::output_directory_validator checks whether the specified directory is writable and can be created,
+if it does not already exists.
+If the tests fail, a seqan3::parser_invalid_argument exception will be thrown. Also, if something unexpected with the
+filesystem happens, a std::filesystem_error will be thrown.
```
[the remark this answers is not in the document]
rrahn: anything that is beyond the control of the validator. For example, if createDirectory throws, or if remove(path) of a temporary file throws.

include/seqan3/argument_parser/validators.hpp
rrahn: Mhmm, AFAIK doxygen ignores this completely. But you never know what the next doxygen version brings, so I documented it anyway. I will at least keep it as a comment so less experienced people can understand what is happening.

include/seqan3/argument_parser/validators.hpp
rrahn: Why is this not helpful?

> A regular file is one type of file that may be stored in a file system. It is called "regular" primarily to distinguish it from other special types of files.

This has nothing to do with c++ but is a general term used in filesystems. It is debatable if also symlinks should be allowed but then I cannot distinguish if the linked filesystem object is a directory or a file. So I kept it like this for now.

include/seqan3/argument_parser/validators.hpp
rrahn: Yes, this is picky 🤣 . But honestly, I really don't see the point of changing this, as it is not even a public interface but a protected used internally.

include/seqan3/argument_parser/validators.hpp
rrahn: In fact it is not over the top. All filesystem operations can possibly throw and this is a resource that needs to be cleaned up in case something went wrong. Accordingly, I wrap it in a RAII object to avoid unexpected results. Also it would be necessary to put the exists check in at least two different places.

include/seqan3/argument_parser/validators.hpp
smehringer: ok if this is "common" its fine. I just did not know

doc/tutorial/argument_parser/index.md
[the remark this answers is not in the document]
rrahn: Yes, like a created temporary file cannot be removed or anything else that is beyond the validators control.

include/seqan3/argument_parser/validators.hpp
smehringer: yes it ignores this but I had trouble with typedefs where an extra brief actually introduced spurious documentation somewhere else. so yes a normal comment would be good

include/seqan3/argument_parser/validators.hpp
smehringer: it is still inconsistent with path_is_readable which acceppts both, file and dir.. :(
How picky am I allowed to be? (I just think it is confusing for the person who looks at it a ywar from now)

include/seqan3/argument_parser/validators.hpp
smehringer: But you are using the filesystem overloads that are noexcept :thinking: ?

include/seqan3/argument_parser/validators.hpp
rrahn: > How picky am I allowed to be?

Well, no one forbids you to be picky and it is perfectly fine to review this as critical as possible. But that doesn't mean that I have to agree with it 😇. And I still don't see the point, though I have changed it already.

include/seqan3/argument_parser/validators.hpp
rrahn: I check it again. I am pretty sure that I added it because the output was strangely without the space.

include/seqan3/argument_parser/validators.hpp
rrahn: not all. Only if the validator should capture this as an error. At least path_is_writable, ... oops I meant filename_is_writable 😈 , can throw std::filesystem_error. But even if not, to protect a resource you have to wrap it into RAII according to the CPP core guidelines.

rrahn:
@smehringer can you please rereview? Many thanks!

smehringer:
your tests fail (not only the coverage test). I'll review when all is fixed

include/seqan3/argument_parser/validators.hpp
smehringer: hm okay, I think making filename_is_writable noexcept would be good anyway. 

In your current solution, no error is thrown if the directory could not be deleted right? should this be the desired behaviour? Because then the directory might have been created but not removed... which might be weird behaviour for a user if the program fails afterward because the directory already exists (an the developer assumed it is not)?

rrahn:
Oh, could you please review first. The broken tests have nothing to do with the changes I applied. Merely some rebase on #951  that broke stuff.

include/seqan3/argument_parser/validators.hpp
rrahn: Well, something is really really broken if I can create the directory but not delete it. So this case will very likely never happen? Like when another process deletes the directory in the meantime.  

Within the deleter I cannot throw because it would happen during the destruction of the raii object. But I added an assertion.

test/snippet/argument_parser/validators_input_file.cpp
```diff
@@ -5,16 +5,17 @@
 
 int main(int argc, const char ** argv)
 {
-    seqan3::argument_parser myparser{"Test", argc, argv}; // initialize
+    seqan3::argument_parser myparser("Test", argc, argv); // initialize
```
smehringer: construction should use braces not parentheses?

EDIT: this should be the case for all the constructors in the snippets of argument parser and validators?

test/snippet/argument_parser/validators_input_file.cpp
```diff
@@ -5,16 +5,17 @@
...
-    //![validator_call]
-    std::filesystem::path myfile;
+    //! [validator_call]
+    std::filesystem::path myfile{};
 
-    myparser.add_option(myfile,'f',"file","Give me a filename.",
-                        seqan3::option_spec::DEFAULT, seqan3::path_existence_validator{});
-    //![validator_call]
+    myparser.add_option(myfile,'f',"file","The input file containing the sequences.",
+                        seqan3::option_spec::DEFAULT, seqan3::input_file_validator({"fa","fasta"}));
```
smehringer:
```suggestion
                        seqan3::option_spec::DEFAULT, seqan3::input_file_validator{{"fa","fasta"}});
```

test/snippet/argument_parser/validators_input_file.cpp
rrahn: Why making an exception for the argument parser class? 😕

smehringer: huh? no exception. I am asking why you are chaning braces to parentheses? I thought we always construct using braces?

rrahn: Yeah ok I kind of misinterpreted the comment as a suggestion from you. 🤕 
I have another commit changing all instances of the parentheses in the argument parser doc and tests.

doc/tutorial/argument_parser/index.md
```diff
@@ -382,24 +382,51 @@ Add a seqan3::value_list_validator to the `-a/--aggregate-by` option that sets t
...
 
-On construction, the validator receives a list (vector) of valid file extensions.
-The validator throws a seqan3::parser_invalid_argument exception whenever a given filename (string) is not in the given extension list.
+SeqAn3 offers two file validator types: the seqan3::input_file_validator and the seqan3::output_file_validator.
+On construction, the validator receives a list (vector) of valid file extensions that are tested against the extension
+of the parsed option value.
+The validator throws a seqan3::parser_invalid_argument exception whenever a given filename's extension is not in the
+given list of valid extensions. In addition, the seqan3::input_file_validator checks if the file already exists
+or if the specified path does not name a file but a directory. The seqan3::output_file_validator checks if the file does
+not already exists in order to prevent overwriting an already existing file.
```
h-2:
```suggestion
not already exist in order to prevent overwriting an already existing file.
```

In general maybe a little more readable:
```
In addition, the seqan3::input_file_validator checks if the file exists, is a regular file and is readable.
The seqan3::output_file_validator on the other hand ensures that the output does not already exist (in order to prevent overwriting an already existing file) and that it can be created.
```

:nail_care:

include/seqan3/argument_parser/validators.hpp
```diff
@@ -297,68 +285,350 @@ class file_ext_validator
...
+            return !static_cast<bool>(ec); // test the error code.
+        }
+
+        // Must be a regular file.
+        if (!std::filesystem::is_regular_file(path))
+            throw std::invalid_argument{detail::to_string("Expected a regular file!", path, "!")};
+
+        std::ifstream file{path};
+        return file.is_open() && file.good();
+    }
```
h-2: Hm, I don't think it's a good idea that we have different return paths from this function, i.e. that we return `false` for some files that are not readable and throw an exception for others.

I would either call it `void validate_readability` and always throw when something is wrong OR call it `bool is readable` and always return true/false and never throw. It makes it a lot easier to reason about.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -297,68 +285,350 @@ class file_ext_validator
...
+ * On construction, the validator can receive a list (std::vector over std::string) of valid file extensions.
+ * The struct acts as a functor that throws a seqan3::parser_invalid_argument exception whenever a given filename's
+ * extension (sts::filesystem::path) is not in the given list of valid file extensions, if the file does not exist, or
+ * if the file does not have the proper read permissions.
+ *
+ * \snippet test/snippet/argument_parser/validators_input_file.cpp usage
+ *
+ * \note The validator works on every type that can be implicitly converted to std::filesystem::path.
+ */
+class input_file_validator : public file_validator_base
```
h-2: Shall it be possible that other validators derive further from this class?

* If yes, please mark the same functions as virtual as in the base class (destructor, operator()).
* If no, please mark this class and/or the respective virtual functions as `final`.

include/seqan3/argument_parser/validators.hpp
h-2: Since everything here seems to operate via exceptions, one might as well use the mechanism everywhere.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -297,68 +285,350 @@ class file_ext_validator
...
+         std::for_each(v.begin(), v.end(), [&] (auto cmp) { this->operator()(cmp); });
+    }
+
+protected:
+
+    /*!\brief Validates the given filename path based on the specified extensions.
+     * \param path The filename path.
+     * \throws parser_invalid_argument if the specified extensions don't match the given path.
+     */
+    void validate_filename_impl(std::filesystem::path const & path) const
```
h-2: why is this called `_impl` and the others not?

include/seqan3/argument_parser/validators.hpp
```diff
@@ -297,68 +285,350 @@ class file_ext_validator
...
+    {
+        if (!std::filesystem::exists(file))
+            throw parser_invalid_argument(detail::to_string("The file ", file, " does not exist!"));
+
+        // Check if file has any read permissions.
+        if (!path_is_readable(file))
+            throw parser_invalid_argument(detail::to_string("Cannot read ", file, "!"));
+
+        // Check extension.
+        file_validator_base::validate_filename_impl(file);
```
h-2: Do you need the `file_validator_base::`? It's simple inheritance or not?

include/seqan3/argument_parser/validators.hpp
```diff
@@ -297,68 +285,350 @@ class file_ext_validator
...
+        if (std::filesystem::exists(dir))
+            deleter = [] (std::filesystem::path const *) { /* no-op */ };
+        else
+            deleter = [] (std::filesystem::path const * p)
+            {
+                [[maybe_unused]] std::error_code ec;
+                std::filesystem::remove(*p, ec);
+
+                assert(!static_cast<bool>(ec));
+            };
```
h-2: braces around block

include/seqan3/argument_parser/validators.hpp
h-2: One could of course also argue that something is severely broken if you can create the directory but create a file inside (which is the only thing that would lead to abnormal termination and lacking clean-up). But of course it is good style to do it with RAII.

In general I wonder whether we shouldn't factor this out somewhere and/or whether you can re-use the temporary-file code we have in our tests.

test/snippet/argument_parser/validators_output_file.cpp
```diff
@@ -0,0 +1,33 @@
...
+    //! [validator_call]
+
+    // an exception will be thrown if the user specifies a filename
+    // that does not have one of the extensions ["fa","fasta"],
+    // if the file already exists, or if the file is not writable.
+    try
+    {
+        myparser.parse();
+    }
+    catch (seqan3::parser_invalid_argument const & ext) // the user did something wrong
```
h-2: are all possible exceptions thrown derived from seqan3::parser_invalid_argument? I thought not...

include/seqan3/argument_parser/validators.hpp
rrahn: Not sure, but I find it good style, since I see, where I need to look for the function definition/declaration.

include/seqan3/argument_parser/validators.hpp
rrahn: good point

include/seqan3/argument_parser/validators.hpp
rrahn: > In general I wonder whether we shouldn't factor this out somewhere and/or whether you can re-use the temporary-file code we have in our tests.

I looked into the temp file and it was not usable like it is there. But indeed it could have some advantages to make it a RAII class. Since the destructor cannot throw we can offer an explicit remove member function that throws. So normal code can remove the path but if anything else went south the destructor will silently cleanup.

test/snippet/argument_parser/validators_output_file.cpp
rrahn: yes that needs to be adapted.

include/seqan3/argument_parser/validators.hpp
h-2: Hm, now it is called `is_...` but it is void which is not really nice. I mean it's not super-important but I think in general the functions should be verbs and if it is actually called `is_foo` than it should return bool.
Why not `void validate_readability` ?

rrahn: I change this as well. No hard feelings.

include/seqan3/io/detail/safe_filesystem_entry.hpp
```diff
@@ -0,0 +1,97 @@
...
+    {
+        std::error_code ec;
+        std::filesystem::remove_all(entry, ec);
+
+        assert(!static_cast<bool>(ec));
+    }
+    //!\}
+
+    /*!\brief Removes a file or empty directory.
+     * \returns `true` if the  underlying path could be removed, otherwise `false`.
```
h-2: This is not correct. According to the documentation it only returns false if the file didn't exist, because you are using the throwing-interface.

This is another example of were we (unfortunately also the standard to some degree) mixes return values and exceptions for error-handling. I would suggest uses the error_code interface and just ignoring the error code.

include/seqan3/io/detail/safe_filesystem_entry.hpp
```diff
@@ -0,0 +1,97 @@
...
+
+    /*!\brief Removes a file or directory and all its contents, recursively.
+    * \returns `true` if the  underlying path could be removed, otherwise `false`.
+    * \throws std::filesystem::filesystem_error on underlying OS API errors.
+    *
+    * \details
+    *
+    * Internally calls std::filesystem::remove_all on the stored std::filesystem::path.
+    */
+    bool remove_all()
```
h-2: Also remove_all() in the current fashion behaves differently from the destructor, because the destructor does not throw, but remove_all() does...

test/snippet/argument_parser/validators_input_directory.cpp
```diff
@@ -0,0 +1,37 @@
...
+    catch (seqan3::parser_invalid_argument const & ext) // the user did something wrong
+    {
+        std::cerr << "[PARSER ERROR] " << ext.what() << "\n"; // customize your error message
+        return -1;
+    }
+    catch (std::filesystem::filesystem_error const & ext)
+    {
+        std::cerr << "[IO ERROR] " << ext.what() << "\n"; // customize your error message
+        return -2;
+    }
```
h-2: Hm, this is not really nice. I think all of these should be parser errors, it is not intuitive that one should catch IO-errors on argument parsing, right?

In general I would suggest:
  * make all internal interfaces just return values
  * have a single point (e.g. `operator()` of the validator) where you translate return values / error codes into exceptions.
  * always throw parser exceptions
  * only catch parser exceptions in the tutorials/snippets/...

If you are absolutely against this and want to have different exception types I would argue that we should change the tutorials/snippets/... to catch `std::exception` because having multiple blocks is ugly, difficult to explain and easy to get wrong.

h-2: sorry, that I didn't make this clear earlier :see_no_evil: 
I didn't think about the implications of "what are the different exception types, let's deal with all of them".

include/seqan3/argument_parser/validators.hpp
```diff
@@ -297,68 +286,354 @@ class file_ext_validator
...
+        std::filesystem::create_directory(dir, ec); // does nothing and is not treated as error if path already exists.
+        // if error code was set or if dummy.txt could not be created within the output dir, throw an error.
+        if (static_cast<bool>(ec))
+            throw parser_invalid_argument(detail::to_string("Cannot create directory: ", dir, "!"));
+
+        if (!dir_exists)
+        {
+            detail::safe_filesystem_entry dir_guard{dir};
+            filename_is_writable(dir / "dummy.txt");
+            dir_guard.remove_all();
```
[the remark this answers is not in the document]
rrahn: yes, it is, but in this case if it fails (which is very unlikely) it will throw. In the destructor it will trigger an assertion but only in debug mode. So in case something went wrong on that IO level it will  have a more understandable error.

include/seqan3/io/detail/safe_filesystem_entry.hpp
rrahn: Ok, I see that the docs could be updated but I disagree on the second part. I strongly advice not to ignore the warnings. If an exception is thrown, then because something severe happened. Finding the cause of this is very hard without an exception. We don't do embedded coding, which is the actually the main reason for also offering the error code versions.

include/seqan3/io/detail/safe_filesystem_entry.hpp
rrahn: Well the destructor should not throw. And that's why I added this member function to also have a throwable interface, because it can come handy in some places.

test/snippet/argument_parser/validators_input_directory.cpp
rrahn: Ok, I agree that this is not beautiful. 

But I am also strongly against not forwarding errors from code that is not part of the validator checks. Like some low level IO errors because the filesystem was corrupted or whatever. 
To solve the issue that only one error type should be catched we  can throw_with_nested. In this case the upper exception is always a parser error. The user could then recursively print the nested exceptions if he wants to.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -297,68 +286,354 @@ class file_ext_validator
...
+     */
+    void path_is_readable(std::filesystem::path const & path) const
+    {
+        // Check if input directory is readable.
+        if (std::filesystem::is_directory(path))
+        {
+            std::error_code ec{};
+            std::filesystem::directory_iterator{path, ec};  // if directory iterator cannot be created, ec will be set.
+            if (static_cast<bool>(ec))
+                throw std::invalid_argument{detail::to_string("Cannot read the directory ", path ,"!")};
```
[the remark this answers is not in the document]
rrahn: Ohm I missed this. This should be parser_invalid_argument. Thx for spotting.

include/seqan3/io/detail/safe_filesystem_entry.hpp
h-2: But the thing that is incorrect in the docs is exactly the thing everyone will get wrong when they use the function. Because there is `std::remove_all` that throws, `std::remove_all` that gets error_code out-parameter and now there is seqan remove_all that returns bool, but also sometimes throws. It's a little messy...

rrahn: I did have a wrong return type for remove_all. That needs to be fixed yes. I can also add a second overload with error_code. But otherwise they just wrap the filesystem call nothing more.

test/snippet/argument_parser/validators_input_directory.cpp
h-2: I still think it would be nicer if all the interfaces return bool (or error_code) except the `operator()` and then you can throw parser exceptions with different texts depending on the function and the error_code (or you can even throw the error_code as part of the exception).

Catching and rethrowing is of course an option, too, it would at least guarantee that the top-level exception is always a parser exception. But it involves adding `catch` which we voted it against a while ago. But it's still better than the current solution.

I have a clear preference for the first, but I can live with the second. Let's not spend another week on this :wink:  But if you do do the second one, can you change the interfaces of the file resource to void functions that throw an error and not return some errors as `false` and some as exceptions? :pray: 

P.S: There is a proposal to do exceptions by values and we could actually implement that already by throwing error_codes and not types based on std::exceptions. We should discuss this a strategy meeting, but maybe after the release.

rrahn: >I still think it would be nicer if all the interfaces return bool (or error_code) except the operator() and then you can throw parser exceptions with different texts depending on the function and the error_code (or you can even throw the error_code as part of the exception).

Well, in general I am in favour of using both in combination (exception and error_codes). 
But this means to refactor the parser exceptions to define its own error_category and error codes. Then in case some other error happens its error code could be passed to the parser exception. The user can then easily examine the error code to see what went wrong.
But I am not refactoring this now.

> Catching and rethrowing is of course an option, too, it would at least guarantee that the top-level exception is always a parser exception. But it involves adding catch which we voted it against a while ago. But it's still better than the current solution.

In my opinion this would be the next best solution, which also keeps all information alive.

>I have a clear preference for the first, but I can live with the second. Let's not spend another week on this 😉 But if you do do the second one, can you change the interfaces of the file resource to void functions that throw an error and not return some errors as false and some as exceptions? 

They don't. They only throw on errors. Removing a non-existing file/directory is not an error. Hence they have the bool return value (or the max_int_t which I need to change).

rrahn:
So in general I think the entire argument parser exception could be refactored to use error_codes instead. This would also allow us to be compliant with possible future directives of C++. For now, I use nested exception to handle unexpected filesystem errors.

PR #912, opened 2019-04-15
[fix] make data structures defined in seqan3::detail always pretty printable (in test cases)

marehr:
I had the problem that defined types in `seqan3::detail` did not use the correct debug_stream overload if they appeared on the right hand side in 

```c++
EXPECT_EQ(seqan3::detail::my_struct{}, seqan3::detail::my_struct{});
```

The left hand side of `EXPECT_EQ` printed fine.

Putting `PrintTo` into the detail namespace solved the problem for me.

### DIFF
--- a/test/include/seqan3/test/pretty_printing.hpp
+++ b/test/include/seqan3/test/pretty_printing.hpp
@@ -29,3 +29,18 @@ void PrintTo (t const & v, std::ostream * out)
 //!\endcond
 
 } // namespace seqan3
+
+namespace seqan3::detail
+{
+
+//!\cond DEV
+//!\brief Overload for the googletest PrintTo function that always delegates to our debug_stream.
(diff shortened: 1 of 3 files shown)


eseiler:
googletest looks for a user defined overload in the namespace the class exists.
If it's in `seqan3` it looks for `seqan3::PrintTo`, if it's in `seqan3::detail` it looks for `seqan3::detail::PrintTo`.

I didn't test it, but I would suspect, pretty printing a `seqan3` class (like `dna4`) now doesn't work anymore?

So, instead of transferring our `PrintTo` to `seqan3::detail` it should exist in both namespaces.

marehr:
okay I investigated this a bit more and came to the same conclusion as @eseiler. I changed the test cases to make the behaviour of google test more clear.

PR #913, opened 2019-04-15
[feature] add bit_scan_reverse and max_bits

marehr:


include/seqan3/core/bit_manipulation.hpp
```diff
@@ -59,4 +68,31 @@ constexpr size_t next_power_of_two(size_t n)
     return n + 1;
 }
 
+/*!\brief Returns the position of the most significant bit (counting from right to left).
+ * \ingroup core
+ *
+ * \param[in] n A unsigned number.
```
eseiler:
```suggestion
 * \param[in] n An unsigned integer.
```

eseiler: `tparam` is missing?

include/seqan3/core/bit_manipulation.hpp
```diff
@@ -59,4 +68,31 @@ constexpr size_t next_power_of_two(size_t n)
...
+ *
+ * \attention *n = 0* is a special case and is undefined behaviour.
+ *
+ * \returns The position of the most significant bit.
+ */
+template <std::UnsignedIntegral unsigned_t>
+constexpr uint8_t bit_scan_reverse(unsigned_t n)
```
eseiler: I'm in strong favour of making the position zero-based.
For `00001000` I'd expect `3` as answer, currently it is `4`.

include/seqan3/core/bit_manipulation.hpp
```diff
@@ -17,11 +17,20 @@
...
 namespace seqan3::detail
 {
 
+/*!\brief How many bits has a type?
+ * \ingroup core
+ *
+ * \tparam type_t The type to determine the number of bits..
+ */
+template <typename type_t>
+constexpr auto max_bits = min_viable_uint_v<8u * sizeof(type_t)>;
```
eseiler: Maybe we can call it something like `bit_size`?

`max_bits` kinda implies dynamic size to me.

include/seqan3/core/bit_manipulation.hpp
marehr: What happens if `10000000`, currently we get `0` and we can distinguish that. The current implementation gives you the (zero-based) position of the first 1.

include/seqan3/core/bit_manipulation.hpp
marehr: I changed the name to `sizeof_bits` I hope this makes it more clear

include/seqan3/core/bit_manipulation.hpp
marehr: changed it

PR #917, opened 2019-04-15
Gap decorator anchor block

mariehoffmann:


rrahn:
@mariehoffmann many thanks for your contribution and your work on that. At the moment we pursue a more module based feature approach and we will put our focus on the search right now. I would at the moment rather close this PR and come to it later when we are putting more focus on the alignment module again.

PR #918, opened 2019-04-16
[FEATURE] Constexpr vector

smehringer:
Blocked by ~#860~ and Repeat_n view because `ranges::repeat_n is not constexpr :roll_eyes: 

The out-commented test will work as soon as repeat_n is implemented in seqan3 so you can ignore those for now. The container test shows that everything works in non-constexpr environment.


include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,867 @@
...
+    //!\cond
+        requires std::Constructible<value_type, /*ranges::iter_reference_t*/reference_t<begin_iterator_type>>
+    //!\endcond
+    constexpr constexpr_vector(begin_iterator_type begin_it, end_iterator_type end_it) : constexpr_vector{}
+    {
+        this->insert(cbegin(), begin_it, end_it);
+        sz = std::distance(begin_it, end_it);
+    }
+
+    template <std::ranges::InputRange other_range_t>
```
eseiler: Docs?

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,867 @@
...
+     *
+     * ### Complexity
+     *
+     * Linear in the size of `ilist`.
+     *
+     * ### Exceptions
+     *
+     * Strong exception guarantee (no data is modified in case an exception is thrown).
+     */
+    constexpr constexpr_vector & operator=(std::initializer_list<value_type_> ilist)
```
eseiler: Sometimes you use `value_type_` and sometimes `value_type`

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,867 @@
...
+    {
+        assign(std::begin(ilist), std::end(ilist));
+        return *this;
+    }
+
+    /*!\brief Assign from a different range.
+     * \tparam other_range_t The type of range to be inserted; must satisfy std::ranges::InputRange and
+     *                       std::CommonReference<value_type_t<other_range_t>, value_type>.
```
eseiler: but you require `std::Constructible` ?

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,867 @@
...
+        return &data[0];
+    }
+
+    //!\copydoc constexpr_vector::begin().
+    constexpr const_iterator cbegin() const noexcept
+    {
+        return &data[0];
+    }
+
+    //!\brief Returns iterator pass the end of the string.
```
eseiler:
```suggestion
    //!\brief Returns iterator past the end of the vector.
```

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,867 @@
...
+    }
+    //!\}
+
+    /*!\name Element access
+     * \{
+     */
+    /*!\brief Return the i-th element.
+     * \param[in]              i Index of the element to retrieve.
+     * \throws std::out_of_range If you access an element behind the last.
+     * \returns                  Either a writable proxy to the element or a copy (if called in const context).
```
eseiler: We don't have a proxy here?
So it's just a reference to the i-th element

eseiler: Maybe you could also add a \attention that this is not constexpr?

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,867 @@
...
+        if (i >= size()) // [[unlikely]]
+        {
+            throw std::out_of_range{"Trying to access element behind the last in constexpr_vector."};
+        }
+        return (*this)[i];
+    }
+
+    /*!\brief Return the i-th element.
+     * \param i The element to retrieve.
+     * \returns Either a writable proxy to the element or a copy (if called in const context).
```
eseiler: Same as for `at()`

eseiler: Also for all the other `\returns`

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,867 @@
...
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    constexpr size_type size() const noexcept
+    {
+        return sz;
+    }
+
+    /*!\brief Returns the maximum number of elements the container is able to hold due to system or library
```
eseiler: Technically the docs are correct, but you have  a special case since you fix the max_size at compile time

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,867 @@
...
+    {
+        return N;
+    }
+
+    /*!\brief Returns the number of elements that the container has currently allocated space for.
+     * \returns The capacity of the currently allocated storage.
+     *
+     * \attention
+     *
+     * This does not operate on underlying concat container, see concat_capacity().
```
eseiler: 🤔

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,867 @@
...
+    {
+        auto const pos_as_num = std::ranges::distance(cbegin(), pos);
+        auto const length = std::ranges::distance(begin_it, end_it);
+
+        if (length == 0)
+            return begin(); // nothing to insert
+
+        for (size_type i = sz + length - 1; i > pos_as_num + length - 1; --i)
+            data[i] = data[i - length];
+        // ranges::copy_backward(&data[pos_as_num], &data[sz], &data[sz + length]);
```
eseiler: why not?

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,867 @@
...
+    /*!\brief Inserts elements from initializer list before position in the container.
+     * \param   pos Iterator before which the content will be inserted. `pos` may be the end() iterator.
+     * \param ilist Initializer list with values to insert.
+     * \returns     Iterator pointing to the first element inserted, or `pos` if `ilist` is empty.
+     *
+     * Given the size `n` of `ilist`, this function results in undefined behaviour if `size()` + `n` > `capacity()`.
+     *
+     * ### Complexity
+     *
+     * Worst-case linear in concat_size(). This is a drawback over e.g. `std::vector<std::vector<alphabet>>`.
```
eseiler: 👀

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,867 @@
...
+
+    /*!\brief Removes specified elements from the container.
+     * \param begin_it Begin of range to erase.
+     * \param   end_it Behind the end of range to erase.
+     * \returns        Iterator following the last element removed. If the iterator `pos` refers to the last element,
+     *                 the end() iterator is returned.
+     *
+     * Invalidates iterators and references at or after the point of the erase, including the end() iterator.
+     *
+     * The iterator first does not need to be dereferenceable if first==end_it: erasing an empty range is a no-op.
```
eseiler: first?

test/unit/range/container/constexpr_vector_test.cpp
```diff
@@ -0,0 +1,365 @@
...
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
+// -----------------------------------------------------------------------------------------------------
+
+#include <gtest/gtest.h>
+
+#include <seqan3/range/container/constexpr_vector.hpp>
+#include <seqan3/range/container/constexpr_string.hpp>
+#include <seqan3/range/container/concept.hpp>
```
eseiler: order

test/unit/range/container/constexpr_vector_test.cpp
```diff
@@ -0,0 +1,365 @@
...
+#include <seqan3/std/ranges>
+
+#include <utility>
+#include <string>
```
eseiler: order, also before seqan3 includes?

test/unit/range/container/constexpr_vector_test.cpp
```diff
@@ -0,0 +1,365 @@
...
+    EXPECT_TRUE((std::is_trivially_copy_assignable_v<constexpr_vector<char, 4>>));
+    EXPECT_TRUE((std::is_nothrow_copy_assignable_v<constexpr_vector<char, 4>>));
+    EXPECT_TRUE((std::is_move_assignable_v<constexpr_vector<char, 4>>));
+    EXPECT_TRUE((std::is_trivially_move_assignable_v<constexpr_vector<char, 4>>));
+    EXPECT_TRUE((std::is_nothrow_move_assignable_v<constexpr_vector<char, 4>>));
+}
+
+TEST(constexpr_vector, concepts)
+{
+    EXPECT_TRUE((container_concept<constexpr_vector<char, 4>>));
```
eseiler: what's missing for `reservable_container_concept`? (stuff like `reserve` exists, although a no-op)

test/unit/range/container/constexpr_vector_test.cpp
```diff
@@ -0,0 +1,365 @@
...
+    int arr[3] = {1, 2, 3};
+    EXPECT_TRUE((std::Same<decltype(constexpr_vector{arr}), constexpr_vector<int, 3>>));
+
+    // Deduce value type and N
+    EXPECT_TRUE((std::Same<decltype(constexpr_vector<int, 5>{arr}), constexpr_vector<int, 5>>));
+
+    // char const *
+    EXPECT_TRUE((std::Same<decltype(constexpr_vector{"hi"}), constexpr_vector<char, 3>>));
+
+    // parameter pack a.k.a initilizer_lists are shitty
```
eseiler:
```suggestion
    // parameter pack a.k.a initialiser_lists are dissatisfactory
```

test/unit/range/container/container_test.cpp
```diff
@@ -56,6 +58,20 @@ TYPED_TEST(container, construction)
     EXPECT_EQ(t3, t7);
 }
 
+TYPED_TEST(container, swap)
```
eseiler: Why move the test up? 😁

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,867 @@
...
+    using difference_type = detail::min_viable_uint_t<N + 1>; //!< The difference_type type.
+    using size_type       = detail::min_viable_uint_t<N>;     //!< The size_type type.
+    //!\}
+
+    //!\cond
+    // this signals to range-v3 that something is a container :|
+    using allocator_type    = void;
+    //!\endcond
+
+    /*!\name Constructors, destructor and assignment
```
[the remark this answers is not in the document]
smehringer: if they are defaulted I guess so? I cannot declare constructors as noexpect can I?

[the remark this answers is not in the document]
rrahn: If they are defaulted or implicitly defined they will be noexcept if they can. If you require it to be noexcept (which you can) then constructing this class means that no member or base class are allowed to potentially throw on construction and you will get a compiler error. So leaving it as is is ok for the majority of the cases.

include/seqan3/range/container/constexpr_vector.hpp
smehringer: because although documented as such, it is not constexpr :roll_eyes: 
I'll remove the line.

test/unit/range/container/constexpr_vector_test.cpp
smehringer: ops :see_no_evil: , a left over

test/unit/range/container/container_test.cpp
smehringer: because the constructors often already use swap :point_up: so it is better to test this before as it might resolve errors in the constructors on the fly

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,919 @@
...
+    constexpr constexpr_vector & operator=(constexpr_vector const &) = default; //!< Defaulted.
+    constexpr constexpr_vector & operator=(constexpr_vector &&) = default;      //!< Defaulted.
+    ~constexpr_vector() = default;                                              //!< Defaulted.
+
+    /*!\brief Construct from an (smaller or equally sized) array over the same value type.
+     * \param[in] array The array to copy the data from.
+     *
+     * ### Complexity
+     *
+     * Constant.
```
eseiler: Are you sure 🤔

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,919 @@
...
+     * Strong exception guarantee (no data is modified in case an exception is thrown).
+     */
+    constexpr constexpr_vector(size_type n, value_type value) : constexpr_vector{}
+    {
+        this->insert(cbegin(), n, value);
+        sz = n;
+    }
+
+    /*!\brief Construct from two iterators.
+     * \tparam begin_iterator_type Must satisfy std::ForwardIterator and `value_type` must be constructible from
```
eseiler:
```suggestion
     * \tparam begin_iterator_type Must model std::ForwardIterator and `value_type` must be constructible from
```

also in other places

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,919 @@
...
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    constexpr size_type size() const noexcept
+    {
+        return sz;
+    }
+
+    /*!\brief Returns the maximum number of elements the container is able to hold which resolves to `N`.
```
eseiler:
```suggestion
    /*!\brief Returns the maximum number of elements the container is able to hold and resolves to `N`.
```

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,919 @@
...
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    constexpr size_type max_size() const noexcept
+    {
+        return N;
+    }
+
+    /*!\brief Returns the number of elements that the container is able to hold which resolves to `N`.
```
eseiler:
```suggestion
    /*!\brief Returns the number of elements that the container is able to hold and resolves to `N`.
```

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,919 @@
...
+ */
+
+#pragma once
+
+#include <array>
+#include <type_traits>
+
+#include <range/v3/view/repeat_n.hpp>
+
+#if SEQAN3_WITH_CEREAL
```
eseiler: Did we decide on how we want to handle the cereal includes?

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,919 @@
...
+    constexpr constexpr_vector & operator=(constexpr_vector const &) = default; //!< Defaulted.
+    constexpr constexpr_vector & operator=(constexpr_vector &&) = default;      //!< Defaulted.
+    ~constexpr_vector() = default;                                              //!< Defaulted.
+
+    /*!\brief Construct from an (smaller or equally sized) array over the same value type.
+     * \param[in] array The array to copy the data from.
+     *
+     * ### Complexity
+     *
+     * Liner in the size of `array`.
```
eseiler:
```suggestion
     * Linear in the size of `array`.
```

test/unit/range/container/constexpr_vector_test.cpp
eseiler: Not resolved 😡 
s < u 😁

test/unit/range/container/constexpr_vector_test.cpp
```diff
@@ -0,0 +1,364 @@
...
+#include <seqan3/range/container/constexpr_vector.hpp>
+#include <seqan3/std/ranges>
+
+using namespace seqan3;
+
+// standard construction.
+TEST(constexpr_vector, standard_construction)
+{
+    EXPECT_TRUE((std::is_default_constructible_v<constexpr_vector<char, 4>>));
+    // EXPECT_TRUE((std::is_trivially_default_constructible_v<constexpr_vector<char, 4>>));
```
eseiler: It's not?

include/seqan3/range/container/constexpr_vector.hpp
smehringer: they will stay in the header where they are needed (like it is now).
We will discuss at the next strategy meeting if we should make cereal a hard dependency. Then we can remove all this HAS_CEREAL stuff

test/unit/range/container/constexpr_vector_test.cpp
smehringer: I don't know :see_no_evil:

eseiler: A trivially default constructible class is a class (defined with class, struct or union) that:
* uses the implicitly defined default constructor.
* has no virtual members.
* **has no non-static data members with brace- or equal- initializers.**
* its base class and non-static data members (if any) are themselves also trivially default constructible types.

If you remove the brace initialisation from `data` and `sz`, it will work. But of course this breaks the other tests and is probably not what we want?

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,918 @@
...
+#include <seqan3/core/detail/int_types.hpp>
+#include <seqan3/core/metafunction/template_inspection.hpp>
+#include <seqan3/range/view/repeat_n.hpp>
+#include <seqan3/range/view/take.hpp>
+#include <seqan3/std/ranges>
+
+namespace seqan3
+{
+
+/*!\brief A constexpr vector implementation with dynamic size at compile time.
```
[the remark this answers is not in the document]
h-2: And we will use it outside of constexpr often, too. Like in the cigar_alphabet. It's just a vector with a fixed capacity, really.

test/unit/range/container/constexpr_vector_test.cpp
smehringer: Thanks for the explanations :)

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,918 @@
...
+ * \implements seqan3::reservable_container_concept
+ * \implements seqan3::Cerealisable
+ * \ingroup container
+ * \tparam value_type_ The underlying value type stored in the vector.
+ * \tparam N           The capacity of the constexpr vector.
+ */
+template <typename value_type_, size_t N>
```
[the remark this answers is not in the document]
smehringer: capacity might clash with the member function capacity... I think N is more meaningful than `cap` as it is usually used to denote the size of something?

include/seqan3/range/container/constexpr_vector.hpp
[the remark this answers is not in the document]
smehringer: hmm I agree that constexpr is not a name that fits the purpose... but `small_vector` isn't either (e.g. I would not assume its constexpr nor fixed in size)? `fixed_size_vector`?

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,918 @@
...
+    //!\cond
+    template <size_t N_other>
+    //!\cond
+        requires (N_other <= N)
+    //!\endcond
+    constexpr constexpr_vector(const value_type (&array)[N_other]) :
+        sz{N_other}
+    {
+        std::ranges::copy(array, data.begin());
+    }
```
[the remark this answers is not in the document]
smehringer: then I need a deduction guide but I guess that's to be preferred against two constructors?

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,918 @@
...
+     * \tparam other_value_type A parameter pack where each type is equal to value_type.
+     * \param[in] args The values to construct from.
+     */
+    template <typename ...other_value_type>
+    //!\cond
+        requires (std::Same<value_type, other_value_type> && ...)
+    //!\endcond
+    constexpr constexpr_vector(other_value_type... args) :
+        data{args...}, sz{sizeof...(other_value_type)}
+    {}
```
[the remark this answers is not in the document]
smehringer: Question: is the static_assert within the constructor body evaluated before member initialization happens? Because then `data{args...}` might fail before the static_assert catches.

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,918 @@
...
+     * Constant.
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    constexpr void resize(size_type const count)
+    {
+        sz = count;
+    }
```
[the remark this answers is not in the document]
smehringer: Why throw an exception? Doesn't an assert suffice?
std::vector does not throw if new size exceeds max size.

include/seqan3/range/container/constexpr_vector.hpp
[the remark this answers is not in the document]
h-2: IIRC you were one of the people insisting that we not use single-letter variable names :point_up: 

> I think N is more meaningful than cap as it is usually used to denote the size of something

Exactly, and it's not about the *size*.

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,918 @@
...
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    template <size_t N_other>
+    constexpr void swap(constexpr_vector<value_type, N_other> & rhs) noexcept
+    {
+        // cassert(sz <= rhs.capacity(), "this->size exceeds the capacity of to-be-swapped constexpr_vector.");
+        // cassert(rhs.size() <= capacity(), "Size of to-be-swapped constexpr_vector exceeds the capacity of this.");
```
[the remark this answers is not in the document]
smehringer: ok

include/seqan3/range/container/constexpr_vector.hpp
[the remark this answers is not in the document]
h-2: > fixed_size_vector

It's not fixed size! That's very important to denote in name and interface, this is exactly what people will be confused about...

Maybe there are better names than `small_vector`, but in the context of strings "small strings" is used for strings stored in static storage ("small string optimisation"). And it clearly suggests that you should only use this data structure when the number of elements will be small.

include/seqan3/range/container/constexpr_vector.hpp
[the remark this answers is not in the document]
h-2: Hm, good point. Well it will be probably be the same amount of code, but I have a small preference for one constructor and one guide... up to you!

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,918 @@
...
+     *
+     * \attention These functions are never called directly, see \ref serialisation for more details.
+     */
+    template <CerealArchive archive_t>
+    void CEREAL_SERIALIZE_FUNCTION_NAME(archive_t & archive)
+    {
+        archive(data);
+        archive(sz);
+    }
+    //!\endcond
```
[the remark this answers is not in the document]
smehringer: yes in the container_test

include/seqan3/range/container/constexpr_vector.hpp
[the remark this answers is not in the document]
h-2: > std::vector does not throw if new size exceeds max size.

Yes, but max_size on a vector is also never reached in practice. I suspect it will happen a lot more often for this data structure, because it's the central "limitation".  Since resize is not called that often and regular container semantics imply that it's not fast anyway, I think it wouldn't hurt to throw an exception here.

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,918 @@
...
+};
+
+} // namespace seqan3
```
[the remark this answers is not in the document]
smehringer: hm ContigousRange is true although I do not have the data() member function :thinking:

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,918 @@
...
+
+    /*!\name Member types
+     * \{
+     */
+    using value_type      = value_type_;                      //!< The value_type type.
+    using reference       = value_type &;                     //!< The reference type.
+    using const_reference = const value_type &;               //!< The const_reference type.
+    using iterator        = value_type *;                     //!< The iterator type.
+    using const_iterator  = value_type const *;               //!< The const_iterator type.
+    using difference_type = detail::min_viable_uint_t<N + 1>; //!< The difference_type type.
```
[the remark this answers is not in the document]
smehringer: changed it to `std::make_signed_t<size_type>`. that fine?

include/seqan3/range/container/constexpr_vector.hpp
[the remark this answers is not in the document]
h-2: Yes, because you have pointers as iterators, but you should still provide the member!

include/seqan3/range/container/constexpr_vector.hpp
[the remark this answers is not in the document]
smehringer: You did not answer by question :(

include/seqan3/range/container/constexpr_vector.hpp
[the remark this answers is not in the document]
smehringer: except that this function cannot be constexpr anymore then

include/seqan3/range/container/constexpr_vector.hpp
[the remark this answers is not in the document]
smehringer: ok

include/seqan3/range/container/constexpr_vector.hpp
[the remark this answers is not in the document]
smehringer: ok fine. I'll name it `capacity_`

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,918 @@
...
+     * In \f$O(count)\f$.
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    constexpr void assign(size_type const count, value_type const value) noexcept
+    {
+        clear();
+        this->insert(cbegin(), count, value);
```
h-2: why `this->` explicitly?

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,918 @@
...
+     *
+     * ### Complexity
+     *
+     * Linear in the size of `ilist`.
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    constexpr constexpr_vector & operator=(std::initializer_list<value_type> ilist) noexcept
```
h-2: Regarding adding `noexcept` everywhere. I was maybe a little quick. This depends, of course, on the value_type being no no_throw_copy_constructible / no_throw_move_constructible. I would maybe add a static member variable `is_noexcept = std::no_throw_copy_constructible_v<value_type>` and then change the `noexcept` on the constructors, assignment, insert... to tbe `noexcept(is_noexcept)`. The documentation would also need to be adapted.

include/seqan3/range/container/constexpr_vector.hpp
```diff
@@ -0,0 +1,918 @@
...
+     *
+     * No-throw guarantee.
+     */
+    constexpr size_type capacity() const noexcept
+    {
+        return capacity_;
+    }
+
+    //!\brief Since the capacity is fixed on compile time, this is a no-op.
+    constexpr void reserve(size_type const SEQAN3_DOXYGEN_ONLY(new_cap)) noexcept
```
h-2: why do you need the DOXYGEN_ONLY ? why not omit it?

include/seqan3/range/container/constexpr_vector.hpp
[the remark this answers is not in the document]
h-2: that's right :cry: 
Ok, let's only assert for now.

include/seqan3/range/container/constexpr_vector.hpp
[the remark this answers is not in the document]
h-2: > Question: is the static_assert within the constructor body evaluated before member initialization happens?

I don't know.

> Because then data{args...} might fail before the static_assert catches.

Yes, that's possible.

include/seqan3/range/container/constexpr_vector.hpp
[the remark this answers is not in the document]
h-2: No. If you have a capacity of 200, your size_type will be uint8_t, but your difference_type will be int8_t which cannot hold all the possible differences. It would probably be correct to `make_signed_t<min_viable_uint_t<(capacity_ + 1) * 2>>`, but maybe you can also just use `ptrdiff_t` here since (in contrast to size) no variable of that type is being saved.

include/seqan3/range/container/constexpr_vector.hpp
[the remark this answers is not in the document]
h-2: But why o_O ?

include/seqan3/range/container/constexpr_vector.hpp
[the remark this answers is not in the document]
smehringer: I tried it. The static assert catches before the member initialization is done

include/seqan3/range/container/constexpr_vector.hpp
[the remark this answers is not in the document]
smehringer: I'll just rename it to small vector as I cannot think of another name.

include/seqan3/range/container/constexpr_vector.hpp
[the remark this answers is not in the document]
smehringer: I dunno... but if it works, it works?
It is also private in the bitcompressed vector.

include/seqan3/range/container/constexpr_vector.hpp
smehringer: check after my next push if i have all the places correct

include/seqan3/range/container/small_vector.hpp
```diff
@@ -0,0 +1,931 @@
...
+     * ### Complexity
+     *
+     * Linear in `capacity_`.
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee if value_type is std::is_nothrow_copy_constructible.
+     */
+    template <size_t capacity2>
+    explicit constexpr small_vector(const value_type (&array)[capacity2]) noexcept(is_noexcept) :
```
h-2: you didn't change it to east-const

include/seqan3/range/container/small_vector.hpp
```diff
@@ -0,0 +1,931 @@
...
+    {
+        lhs.swap(rhs);
+    }
+
+    //!\overload
+    template <size_t capacity2>
+    friend constexpr void swap(small_vector && lhs, small_vector && rhs) noexcept(is_noexcept)
+    {
+        lhs.swap(rhs);
+    }
```
h-2: Concerning the swaps, the `std::swap` free function should be auto-generated since you have move-cosntructors so you don't need the friends. In fact you could probably have the members functions delegate to the free function, too. (this will not perform the micro-optimisation of saving one copy for temporaries, though.

include/seqan3/range/container/small_vector.hpp
smehringer: ops

include/seqan3/range/container/small_vector.hpp
smehringer: > Concerning the swaps, the std::swap free function should be auto-generated since you have move-cosntructors so you don't need the friends.

those friends are not in the std namespace but in the seqan3 namespace? Don;t we have our own swap?

> In fact you could probably have the members functions delegate to the free function, too.

No I cannot because neither `std::swap` nor `ranges::swap` are constexpr (prior to c++20)

include/seqan3/range/container/small_vector.hpp
```diff
@@ -0,0 +1,931 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
```
eseiler:
```suggestion
// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
```

test/unit/range/container/small_vector_test.cpp
```diff
@@ -0,0 +1,405 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
```
eseiler:
```suggestion
// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
```

PR #920, opened 2019-04-16
[FEATURE] binning_directory

### LINKED ISSUE #1562: [FEATURE] Add the interleaved bloom filter
Implement a binning directory that allows to answer set member ships for k-mers, such that the bins are identified which cover the k-mer.

Rationale: In a distributed search or other application the reference sequences are distributed over a set of bins. To avoid looking into all bins for every query, a mechanism is needed that will select only relevant bins that actually cover the query. This has a tremendous effect on reducing the runtime, while at the same time index structures over the bins can be dynamically updated. 

some design decisions:
Design decisions for the seqan3::interleaved_bloom_filter were done by making it similar to the `std::unordered_multimap`.

* The IBF is constructed with 
  ```cpp
  seqan3::interleaved_bloom_filter{seqan3::bin_count, seqan3::bin_size, seqan3::hash_number_count}
  ``` 
  We decided for `count` (instead of number or size) for the amount of bins because the std::unordered_multiset calls it `bucket_count`. We still call it `bin` instead of `bucket` because we want to call the abstract solution `binning_directory` and not `bucketing_directory`.
* A value (e.g. kmer or minimizer) will be inserted into the IBF with  
  ```cpp
  ibf.emplace(value, seqan3::bin_index{0})
  ``` 
  analogously to the std::unordered_multiset. `insert` will not be provided since we don't need it.
* To retrieve the bitvector which is set to `1` if a bin contains the (kmer-)value or `0` if not, we will use  
  ```cpp
  sdsl::bit_vector result = ibf.bulk_contains(value)
  ```
  This was chosen because the std::unordered_multiset uses `contains` to indicate whether a key is contained in any of the buckets(/bins) but we directly output a bulk result with a boolean (bit) for every bin.
* we will provide the getter: `ibf.bin_count` which returns the number of bins and `ibf.size()` which returns the `bin_size`.

eseiler:
Blocked by ~#933~

Resolves #1562

TODO:

- [x] Snippets
- [x] Docs (grouping)
- [x] make parameters strong types
- [x] ~~Debug streaming~~
- [x] Refactor tests

### DIFF
--- a/include/seqan3/search/dream_index/all.hpp
+++ b/include/seqan3/search/dream_index/all.hpp
@@ -1,0 +1,19 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2020, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \author Enrico Seiler <enrico.seiler AT fu-berlin.de>
+ * \brief Meta-header for the DREAM index module.
+ *
+ * \defgroup submodule_dream_index DREAM Index
+ * \brief Provides seqan3:interleaved_bloom_filter.
+ * \ingroup search
+ */
+
+ #pragma once
+
+ #include <seqan3/search/dream_index/interleaved_bloom_filter.hpp>
--- a/include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
+++ b/include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
@@ -184,6 +184,7 @@ private:
     /*!\brief Perturbs a value and fits it into the vector.
      * \param h The value to process.
      * \param i The seed to use.
+     * \returns A hashed value representing a position within the bounds of `data`.
      * \sa https://probablydance.com/2018/06/16/
      * \sa https://lemire.me/blog/2016/06/27
      */
--- a/include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
+++ b/include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
@@ -14,10 +14,10 @@
 
 #include <sdsl/bit_vectors.hpp>
 
+#include <seqan3/core/bit_manipulation.hpp>
 #include <seqan3/core/concept/cereal.hpp>
 #include <seqan3/core/detail/strong_type.hpp>
 #include <seqan3/std/algorithm>
-#include <seqan3/std/ranges>
 
 namespace seqan3
 {
@@ -164,6 +164,8 @@ private:
     size_t technical_bins{};
     //!\brief The size of each bin in bits.
     size_t bin_size_{};
+    //!\brief How many bits do we need to represent `bin_size_`?
+    size_t bin_size_in_bits{};
     //!\brief The number of 64-bit integers needed to store `bins` many bits (e.g. `bins = 50` -> `bin_words = 1`).
     size_t bin_words{};
     //!\brief The number of hash functions.
@@ -171,7 +173,11 @@ private:
     //!\brief The bitvector.
     data_type data{};
     //!\brief Precalculated seeds for hashing.
-    std::vector<size_t> hash_seeds{};
+    static constexpr std::array<size_t, 5> hash_seeds{13572355802537770549ULL, // 2**64 / (e/2)
+                                                      13043817825332782213ULL, // 2**64 / sqrt(2)
+                                                      10650232656628343401ULL, // 2**64 / sqrt(3)
+                                                      16499269484942379435ULL, // 2**64 / (sqrt(5)/2)
+                                                      4893150838803335377ULL}; // 2**64 / (3*pi/5)
     //!\brief The result buffer for a `bulk_contains()` query.
     mutable binning_bitvector result_buffer{};
 
@@ -182,6 +188,8 @@ private:
      */
     constexpr void hash_and_fit(size_t & h) const
     {
+        assert(bin_size_in_bits != 64);
+        h ^= h >> (64 - bin_size_in_bits); // Shift higher bits into lower bits
         h *= 11400714819323198485ULL; // = 2^64 / golden_ration, to expand h to 64 bit range
         // Use fastrange (integer modulo without division) if possible.
 #ifdef __SIZEOF_INT128__
@@ -209,7 +217,7 @@ public:
     /*!\brief Construct an uncompressed Interleaved Bloom Filter.
      * \param bins_ The number of bins.
      * \param size The bitvector size.
-     * \param funs The number of hash functions. Default 2.
+     * \param funs The number of hash functions. Default 2. At least 1, at most 5.
      *
      * \attention This constructor can only be used to construct **uncompressed** Interleaved Bloom Filters.
      *
@@ -232,20 +240,17 @@ public:
 
         if (bins == 0)
             throw std::logic_error{"The number of bins must be > 0."};
-        if (hash_funs == 0)
-            throw std::logic_error{"The number of hash functions must be > 0."};
+        if (hash_funs == 0 || hash_funs > 5)
+            throw std::logic_error{"The number of hash functions must be > 0 and <= 5."};
         if (bin_size_ == 0)
             throw std::logic_error{"The size of a bin must be > 0."};
 
+        bin_size_in_bits = 64 - detail::count_leading_zeros(bin_size_);
         bin_words = (bins + 63) >> 6; // = ceil(bins/64)
         technical_bins  = bin_words << 6; // = bin_words * 64
         data = sdsl::bit_vector(technical_bins * bin_size_);
 
         result_buffer.resize(bins);
-        hash_seeds.resize(hash_funs);
-
-        for (size_t i = 0; i < hash_funs; ++i)
-            hash_seeds[i] = i ^ 0x90b45d39fb6da1fa;
     }
 
     /*!\brief Construct a compressed Interleaved Bloom Filter.
@@ -264,8 +269,8 @@ public:
         requires data_layout_mode == data_layout::compressed
     //!\endcond
     {
-        std::tie(bins, technical_bins, bin_size_, bin_words, hash_funs, hash_seeds) =
-            std::tie(ibf.bins, ibf.technical_bins, ibf.bin_size_, ibf.bin_words, ibf.hash_funs, ibf.hash_seeds);
+        std::tie(bins, technical_bins, bin_size_, bin_words, hash_funs) =
+            std::tie(ibf.bins, ibf.technical_bins, ibf.bin_size_, ibf.bin_words, ibf.hash_funs);
 
         data = sdsl::sd_vector<>{ibf.data};
         result_buffer.resize(bins);
@@ -293,12 +298,12 @@ public:
     //!\endcond
     {
         assert(bin.get() < bins);
-        for (auto const & seed : hash_seeds)
+        for (size_t i = 0; i < hash_funs; ++i)
         {
-            size_t idx = seed * value;
+            size_t idx = hash_seeds[i] * value;
             hash_and_fit(idx);
             idx += bin.get();
-            assert (idx < data.size());
+            assert(idx < data.size());
             data[idx] = 1;
         };
     }
@@ -386,9 +391,14 @@ public:
     {
         assert(result_buffer.size() == bin_count());
 
-        std::vector<size_t> idx = hash_seeds;
+        std::array<size_t, 5> idx;
+        std::memcpy(&idx, &hash_seeds, sizeof(size_t) * hash_funs);
 
-        std::ranges::for_each(idx, [this, value] (size_t & i) { i *= value; hash_and_fit(i); });
+        for (size_t i = 0; i < hash_funs; ++i)
+        {
+            idx[i] *= value;
+            hash_and_fit(idx[i]);
+        }
 
         for (size_t batch = 0; batch < bin_words; ++batch)
         {
@@ -491,7 +501,6 @@ public:
         archive(bin_words);
         archive(hash_funs);
         archive(data);
-        archive(hash_seeds);
     }
     //!\endcond
 };
--- a/include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
+++ b/include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
@@ -183,7 +183,7 @@ private:
 
     /*!\brief Perturbs a value and fits it into the vector.
      * \param h The value to process.
-     * \param i The seed to use.
+     * \param seed The seed to use.
      * \returns A hashed value representing a position within the bounds of `data`.
      * \sa https://probablydance.com/2018/06/16/
      * \sa https://lemire.me/blog/2016/06/27
@@ -191,8 +191,8 @@ private:
     inline constexpr size_t hash_and_fit(size_t h, size_t const seed) const
     {
         h *= seed;
-        assert(hash_shift > 0);
-        h ^= h >> hash_shift; // Shift higher bits into lower bits
+        assert(hash_shift < 64);
+        h ^= h >> hash_shift; // XOR and shift higher bits into lower bits
         h *= 11400714819323198485ULL; // = 2^64 / golden_ration, to expand h to 64 bit range
         // Use fastrange (integer modulo without division) if possible.
 #ifdef __SIZEOF_INT128__
@@ -273,8 +273,8 @@ public:
         requires data_layout_mode == data_layout::compressed
     //!\endcond
     {
-        std::tie(bins, technical_bins, bin_size_, bin_words, hash_funs) =
-            std::tie(ibf.bins, ibf.technical_bins, ibf.bin_size_, ibf.bin_words, ibf.hash_funs);
+        std::tie(bins, technical_bins, bin_size_, hash_shift, bin_words, hash_funs) =
+            std::tie(ibf.bins, ibf.technical_bins, ibf.bin_size_, ibf.hash_shift, ibf.bin_words, ibf.hash_funs);
 
         data = sdsl::sd_vector<>{ibf.data};
         result_buffer.resize(bins);
--- a/include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
+++ b/include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
@@ -1,0 +1,501 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2020, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \author Enrico Seiler <enrico.seiler AT fu-berlin.de>
+ * \brief Provides seqan3::interleaved_bloom_filter.
+ */
+
+#pragma once
+
+#include <sdsl/bit_vectors.hpp>
+
+#include <seqan3/core/concept/cereal.hpp>
+#include <seqan3/core/detail/strong_type.hpp>
+#include <seqan3/std/algorithm>
+#include <seqan3/std/ranges>
+
+namespace seqan3
+{
+
+/*!\addtogroup submodule_dream_index
+ * \{
+ */
+//!\brief Determines if the Interleaved Bloom Filter is compressed.
+enum data_layout : bool
+{
+    uncompressed, //!< The Interleaved Bloom Filter is uncompressed.
+    compressed    //!< The Interleaved Bloom Filter is compressed.
+};
+
+//!\brief A strong type that represents the number of bins for the seqan3::interleaved_bloom_filter.
+struct bin_count : public detail::strong_type<size_t, bin_count, detail::strong_type_skill::convert>
+{
+    using detail::strong_type<size_t, bin_count, detail::strong_type_skill::convert>::strong_type;
+};
+
+//!\brief A strong type that represents the number of bits for each bin in the seqan3::interleaved_bloom_filter.
+struct bin_size : public detail::strong_type<size_t, bin_size, detail::strong_type_skill::convert>
+{
+    using detail::strong_type<size_t, bin_size, detail::strong_type_skill::convert>::strong_type;
+};
+
+//!\brief A strong type that represents the number of hash functions for the seqan3::interleaved_bloom_filter.
+struct hash_function_count : public detail::strong_type<size_t, hash_function_count, detail::strong_type_skill::convert>
+{
+    using detail::strong_type<size_t, hash_function_count, detail::strong_type_skill::convert>::strong_type;
+};
+
+//!\brief A strong type that represents the bin index for the seqan3::interleaved_bloom_filter.
+struct bin_index : public detail::strong_type<size_t, bin_index, detail::strong_type_skill::convert>
+{
+    using detail::strong_type<size_t, bin_index, detail::strong_type_skill::convert>::strong_type;
+};
+
+/*!\brief The IBF binning directory. A data structure that efficiently answers set-membership queries for multiple bins.
+ * \tparam data_layout_mode_ Indicates whether the underlying data type is compressed. See seqan3::data_layout.
+ *
+ * \details
+ *
+ * ### Binning Directory
+ *
+ * A binning directory is a data structure that can be used to determine set membership for elements.
+ * For example, a common use case is dividing a database into a fixed number (e.g. 1024) bins by some means
+ * of clustering (e.g. taxonomic binning or k-mer similarity clustering for genomic sequences).
+ * For a query, the binning directory can now answer in which bins the query (probably) occurs.
+ * In SeqAn we provide the Interleaved Bloom Filter (IBF) that can answer these queries efficiently.
+ *
+ * ### Interleaved Bloom Filter (IBF)
+ *
+ * The Interleaved Bloom Filter is a probabilistic data structure that extends the
+ * [Bloom Filter](https://en.wikipedia.org/wiki/Bloom_filter).
+ * A Bloom Filter can be thought of as a bitvector of length `n` and `h` hash functions and is used to determine set
+ * membership. To insert data, the data is hashed by the `h` hash functions (returning values in `[0, n)`) and the
+ * corresponding `h` positions in the bitvector are set to `1`. To query data, i.e. to determine whether the query
+ * belongs to the set the Bloom Filter was built for, the query is hashed by the same `h` hash functions and the
+ * corresponding positions are checked. If all `h` positions contain a `1`, the query is (probably) in the data set.
+ * Since the Bloom Filter has variable length, the hashing is not bijective, i.e. it may return true for a set
+ * membership query even though the query was never inserted into the Bloom Filter. Note that the Bloom Filter
+ * will always return `true` if the query was inserted, i.e. there may be false positives, but no false negatives.
+ *
+ * The Interleaved Bloom Filter now applies the concept of a Bloom Filter to multiple sets and provides a *global*
+ * data structure to determine set membership of a query in `b` data sets/bins.
+ * Conceptually, a Bloom Filter is created for each bin using the same fixed length and fixed hash functions for each
+ * filter. The resulting `b` Bloom Filters are then interleaved such that the `i`'th bit if each Bloom Filter are
+ * adjacent to each other:
+ * ```
+ * Bloom Filter 0       Bloom Filter 1      Bloom Filter 2      Bloom Filter 3
+ * |0.0|0.1|0.2|0.3|    |1.0|1.1|1.2|1.3|   |2.0|2.1|2.2|2.3|   |3.0|3.1|3.2|3.3|
+ * ```
+ * Where `x.y` denotes the `y`'th bit of the `x`'th Bloom Filter.
+ * ```
+ * Interleaved Bloom Filter
+ * |0.0|1.0|2.0|3.0|0.1|1.1|2.1|3.1|0.2|1.2|2.2|3.2|0.3|1.3|2.3|3.3|
+ * ```
+ * A query can now be searched in all `b` bins by computing the `h` hash functions, retrieving the `h` sub-bitvectors of
+ * length `b` starting at the positions indicated by the hash functions. The bitwise AND of these sub-bitvectors yields
+ * the binningvector, a bitvector of length `b` where the `i`'th bit indicates set membership in the `i`'th bin.
+ *
+ * ### Compression
+ *
+ * The Interleaved Bloom Filter can be compressed by passing `data_layout::compressed` as template argument.
+ * The compressed `seqan3::interleaved_bloom_filter<seqan3::data_layout::compressed>` can only be constructed from a
+ * `seqan3::interleaved_bloom_filter`, in which case the underlying bitvector is compressed.
+ * The compressed Interleaved Bloom Filter is immutable, i.e. only querying is supported.
+ *
+ * ### Thread safety
+ *
+ * The Interleaved Bloom Filter promises the basic thread-safety by the STL that all
+ * calls to `const` member functions are safe from multiple threads (as long as no thread calls
+ * a non-`const` member function at the same time).
+ *
+ * Additionally, concurrent calls to `set` are safe iff each thread handles a multiple of wordsize (=64) many bins.
+ * For example, calls to `set` from multiple threads are safe if `thread_1` accesses bins 0-63, `thread_2` bins 64-127,
+ * and so on.
+ */
+template <data_layout data_layout_mode_ = data_layout::uncompressed>
+class interleaved_bloom_filter
+{
+public:
+    //!\brief A bitvector representing the result of a call to `bulk_contains` of the seqan3::interleaved_bloom_filter.
+    class binning_bitvector : private sdsl::bit_vector
+    {
+    public:
+        /*!\name Constructors, destructor and assignment
+         * \{
+         */
+        binning_bitvector() = default; //!< Defaulted.
+        binning_bitvector(binning_bitvector const &) = default; //!< Defaulted.
+        binning_bitvector & operator=(binning_bitvector const &) = default; //!< Defaulted.
+        binning_bitvector(binning_bitvector &&) = default; //!< Defaulted.
+        binning_bitvector & operator=(binning_bitvector &&) = default; //!< Defaulted.
+        ~binning_bitvector() = default; //!< Defaulted.
+        //!\}
+
+        using sdsl::bit_vector::begin;
+        using sdsl::bit_vector::end;
+        using sdsl::bit_vector::operator==;
+        using sdsl::bit_vector::operator[];
+
+    private:
+        friend class interleaved_bloom_filter;
+        using sdsl::bit_vector::resize;
+        using sdsl::bit_vector::set_int;
+    };
+
+private:
+    //!\cond
+    template <data_layout data_layout_mode>
+    friend class interleaved_bloom_filter;
+    //!\endcond
+
+    //!\brief The underlying datatype to use.
+    using data_type = std::conditional_t<data_layout_mode_ == data_layout::uncompressed,
+                                         sdsl::bit_vector,
+                                         sdsl::sd_vector<>>;
+
+    //!\brief The number of bins specified by the user.
+    size_t bins{};
+    //!\brief The number of bins stored in the IBF (next multiple of 64 of `bins`).
+    size_t technical_bins{};
+    //!\brief The size of each bin in bits.
+    size_t bin_size_{};
+    //!\brief The number of 64-bit integers needed to store `bins` many bits (e.g. `bins = 50` -> `bin_words = 1`).
+    size_t bin_words{};
+    //!\brief The number of hash functions.
+    size_t hash_funs{};
+    //!\brief The bitvector.
+    data_type data{};
+    //!\brief Precalculated seeds for hashing.
+    std::vector<size_t> hash_seeds{};
+    //!\brief The result buffer for a `bulk_contains()` query.
+    mutable binning_bitvector result_buffer{};
+
+    /*!\brief Perturbs a value and fits it into the vector.
+     * \param h The value to process.
+     * \sa https://probablydance.com/2018/06/16/
+     * \sa https://lemire.me/blog/2016/06/27
+     */
+    constexpr void hash_and_fit(size_t & h) const
+    {
+        h *= 11400714819323198485ULL; // = 2^64 / golden_ration, to expand h to 64 bit range
+        // Use fastrange (integer modulo without division) if possible.
+#ifdef __SIZEOF_INT128__
+        h = static_cast<uint64_t>((static_cast<__uint128_t>(h) * static_cast<__uint128_t>(bin_size_)) >> 64);
+#else
+        h %= bin_size_;
+#endif
+        h *= technical_bins;
+    }
+
+public:
+    //!\brief Indicates whether the Interleaved Bloom Filter is compressed.
+    static constexpr data_layout data_layout_mode = data_layout_mode_;
+
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    interleaved_bloom_filter() = default; //!< Defaulted.
+    interleaved_bloom_filter(interleaved_bloom_filter const &) = default; //!< Defaulted.
+    interleaved_bloom_filter & operator=(interleaved_bloom_filter const &) = default; //!< Defaulted.
+    interleaved_bloom_filter(interleaved_bloom_filter &&) = default; //!< Defaulted.
+    interleaved_bloom_filter & operator=(interleaved_bloom_filter &&) = default; //!< Defaulted.
+    ~interleaved_bloom_filter() = default; //!< Defaulted.
+
+    /*!\brief Construct an uncompressed Interleaved Bloom Filter.
+     * \param bins_ The number of bins.
+     * \param size The bitvector size.
+     * \param funs The number of hash functions. Default 2.
+     *
+     * \attention This constructor can only be used to construct **uncompressed** Interleaved Bloom Filters.
+     *
+     * \details
+     *
+     * ### Example
+     *
+     * \include test/snippet/search/dream_index/interleaved_bloom_filter_constructor.cpp
+     */
+    interleaved_bloom_filter(seqan3::bin_count bins_,
+                             seqan3::bin_size size,
+                             seqan3::hash_function_count funs = seqan3::hash_function_count{2u})
+    //!\cond
+        requires data_layout_mode == data_layout::uncompressed
+    //!\endcond
+    {
+        bins = bins_.get();
+        bin_size_ = size.get();
+        hash_funs = funs.get();
+
+        if (bins == 0)
+            throw std::logic_error{"The number of bins must be > 0."};
+        if (hash_funs == 0)
+            throw std::logic_error{"The number of hash functions must be > 0."};
+        if (bin_size_ == 0)
+            throw std::logic_error{"The size of a bin must be > 0."};
+
+        bin_words = (bins + 63) >> 6; // = ceil(bins/64)
+        technical_bins  = bin_words << 6; // = bin_words * 64
+        data = sdsl::bit_vector(technical_bins * bin_size_);
+
+        result_buffer.resize(bins);
+        hash_seeds.resize(hash_funs);
+
+        for (size_t i = 0; i < hash_funs; ++i)
+            hash_seeds[i] = i ^ 0x90b45d39fb6da1fa;
+    }
+
+    /*!\brief Construct a compressed Interleaved Bloom Filter.
+     * \param[in] ibf The uncompressed seqan3::interleaved_bloom_filter.
+     *
+     * \attention This constructor can only be used to construct **compressed** Interleaved Bloom Filters.
+     *
+     * \details
+     *
+     * ### Example
+     *
+     * \include test/snippet/search/dream_index/interleaved_bloom_filter_constructor_compressed.cpp
+     */
+    interleaved_bloom_filter(interleaved_bloom_filter<data_layout::uncompressed> const & ibf)
+    //!\cond
+        requires data_layout_mode == data_layout::compressed
+    //!\endcond
+    {
+        std::tie(bins, technical_bins, bin_size_, bin_words, hash_funs, hash_seeds) =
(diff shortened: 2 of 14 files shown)


eseiler:
I'm getting timeouts for the coverage test 😬

include/seqan3/search/dream_index/concept.hpp
```diff
@@ -0,0 +1,41 @@
...
+/*!\file
+ * \author Enrico Seiler <enrico.seiler AT fu-berlin.de>
+ * \brief Provides concepts for the seqan3::dream_index module.
+ */
+
+#pragma once
+
+#include <seqan3/core/metafunction/template_inspection.hpp>
+#include <seqan3/range/container/concept.hpp>
+#include <seqan3/search/dream_index/binning_directory.hpp>
```
smehringer: why do you include the binning_directory here?
also why the container/concept?

include/seqan3/search/dream_index/detail/binning_directory_base.hpp
```diff
@@ -0,0 +1,201 @@
...
+
+        std::ranges::for_each(idx, [this, h](size_t & i) { i *= h; hash_and_fit(i); });
+
+        // SIMD?!! std::span?
+        for (size_t batch = 0; batch < bin_width; ++batch)
+        {
+           size_t tmp = data.get_int(idx[0]);
+           idx[0] += 64;
+
+           for(size_t i = 1; i < num_hash; ++i)
```
smehringer:
```suggestion
           for (size_t i = 1; i < num_hash; ++i)
```

include/seqan3/search/dream_index/detail/binning_directory_base.hpp
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    //!\overload
+    [[nodiscard]] sdsl::bit_vector get(size_t const h) const
+    {
+        sdsl::bit_vector result(bins);
+
+        std::vector<size_t> idx = pre_hash;
+
+        std::ranges::for_each(idx, [this, h](size_t & i) { i *= h; hash_and_fit(i); });
```
smehringer:
```suggestion
        std::ranges::for_each(idx, [this, h] (size_t & i) { i *= h; hash_and_fit(i); });
```

include/seqan3/search/dream_index/detail/binning_directory_base.hpp
```diff
@@ -0,0 +1,201 @@
...
+    inline size_t size() const noexcept
+    {
+        return data.size();
+    }
+
+    /*!\brief Test for equality.
+     * \param rhs binning_directory to compare to.
+     * \returns `true` if equal, `false` otherwise.
+     */
+    bool operator==(binning_directory_base const & rhs) const noexcept
```
smehringer: I think we always implement binary operators as friends

include/seqan3/search/dream_index/detail/binning_directory_base.hpp
```diff
@@ -0,0 +1,201 @@
...
+        archive(block_size);
+        archive(block_count);
+        archive(num_hash);
+        archive(data);
+        archive(pre_hash);
+    }
+    //!\endcond
+};
+
+}
```
smehringer: // namespace

include/seqan3/search/dream_index/detail/bitvector.hpp
```diff
@@ -0,0 +1,1194 @@
...
+/*!\brief A wrapper for the sdsl::bit_vector.
+ * \implements std::ReservableContainer
+ *
+ * This class additionally provides get_int(), set_int() and a custom
+ * \ref resize(size_type bits_, size_type block_size, size_type block_size_) "resize" function.
+ */
+class bitvector_uncompressed
+{
+private:
+    // friend bitvector_compressed;
```
smehringer: ?

include/seqan3/search/dream_index/detail/bitvector.hpp
```diff
@@ -0,0 +1,1194 @@
...
+    //!\brief The underlying sdsl vector implementation.
+    using sdsl_vector_type = sdsl::bit_vector;
+    //!\brief The underlying data structure.
+    sdsl_vector_type data_value;
+
+public:
+    /*!\name Member types
```
smehringer:
```suggestion
    /*!\name Associated types
```

include/seqan3/search/dream_index/detail/bitvector.hpp
```diff
@@ -0,0 +1,1194 @@
...
+
+    //!\cond
+    // this signals to range-v3 that something is a container :|
+    using allocator_type = void;
+    //!\endcond
+
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    bitvector_uncompressed() = default;                              //!< Default constructor.
```
smehringer: just `Defaulted.` and align all

include/seqan3/search/dream_index/detail/bitvector.hpp
```diff
@@ -0,0 +1,1194 @@
...
+     *
+     * In \f$O(count*value)\f$.
+     *
+     * \par Exceptions
+     *
+     * Strong exception guarantee (no data is modified in case an exception is thrown).
+     */
+    bitvector_uncompressed(size_type const count, value_type value = 0)
+    {
+        data_value = sdsl::bit_vector(count, value);
```
smehringer: () -> {}

include/seqan3/search/dream_index/detail/bitvector.hpp
```diff
@@ -0,0 +1,1194 @@
...
+    }
+
+    /*!\brief Construct/assign from pair of iterators.
+     * \tparam begin_iterator_type Must satisfy std::ForwardIterator.
```
smehringer: satisfy -> model

include/seqan3/search/dream_index/detail/bitvector.hpp
```diff
@@ -0,0 +1,1194 @@
...
+     * \param index The position to assign to.
+     * \param value The value to assign.
+     *
+     * \attention This method will set as many bits as there are in `size_type`, i.e. usually 64 bits.
+     *
+     * \par Complexity
+     *
+     * Constant.
+     */
+    void set_int(size_type const index, size_type const value)
```
smehringer: three questions:  
1. what value are you assigning? it is a bitvector so its value_type are bits?
2. is the set_int of data_value private or protected? 
3. what was your motivation of wrapping the sdsl vector instead of inheriting from it?

include/seqan3/search/dream_index/binning_directory.hpp
```diff
@@ -0,0 +1,257 @@
...
+//!\cond
+template <typename strategy, typename bitvector_t>
+class binning_directory {};
+//!\endcond
+
+/*!\brief The IBF binning directory.
+ * \tparam bitvector_t The underlying bitvector layout.
+ */
+template <typename bitvector_t_>
+class binning_directory<ibf, bitvector_t_>
```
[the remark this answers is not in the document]
eseiler: This is from an old review? The current implementation doesn't use tags

include/seqan3/search/dream_index/binning_directory.hpp
```diff
@@ -0,0 +1,257 @@
...
+        block_size  = bin_width << 6;
+        block_count = bits_ / block_size;
+
+        pre_hash.resize(num_hash);
+        for(size_t i = 0; i < num_hash; ++i)
+            pre_hash[i] = i ^ (15 * seed);
+    }
+    //!\}
+
+    /*!\brief Perturbs a value and fits it into the vector.
```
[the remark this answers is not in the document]
eseiler: Currently this is private. But I can be a bit more verbose in the `\details`

include/seqan3/search/dream_index/concept.hpp
eseiler: sdsl/bit_vector is probably enough
The concept was needed before

include/seqan3/search/dream_index/detail/bitvector.hpp
eseiler: I'll change it. But basically we need to befriend the compressed one because it needs access (compressed is only constructible through a uncompressed one).

include/seqan3/search/dream_index/detail/bitvector.hpp
eseiler: but his is no initializer list?

include/seqan3/search/dream_index/detail/bitvector.hpp
eseiler: 1. we set 64 bits at once
2. public
3. Mhm, the SDSL bitvector models ReservableContainer, but I wanted to add some asserts. But I could probably just inherit and be done with it?

smehringer: well, then this is a lot of code that just delegates to the underlying data member which would not be neccessary if you inherit (e.g. size(), max_size()), so if there is no reason against it I would say inheritance is better in this case and you overload only those functions where asserts are needed/useful. What do you think?

include/seqan3/search/dream_index/detail/binning_directory_base.hpp
```diff
@@ -0,0 +1,209 @@
...
+    {
+        return num_hash;
+    }
+
+    /*!\brief Determines set membership of a given value.
+     * \param[in,out] result The sdsl::bit_vector to store the result into.
+     * \param[in]     h      The raw hash value to process.
+     * \returns A sdsl::bit_vector of size bins where each position indicates the bin membership of the hash value.
+     */
+    [[nodiscard]] sdsl::bit_vector get(size_t const h, sdsl::bit_vector && result) const
```
smehringer: can't you default the result and save a function overload?
```suggestion
    [[nodiscard]] sdsl::bit_vector get(size_t const h, sdsl::bit_vector && result = sdsl::bit_vector(bins)) const
```
If not the overload below should at least use this overload instead of the code duplication
```cpp
    [[nodiscard]] sdsl::bit_vector get(size_t const h) const
    {
        sdsl::bit_vector result(bins);
        get(h, result);
    }
```

test/unit/search/dream_index/binning_directory_test.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+template <typename T>
+class binning_directory_compressed_test : public ::testing::Test
+{};
+
+using binning_directory_types = ::testing::Types<binning_directory>;
+using binning_directory_compressed_types = ::testing::Types<binning_directory_compressed>;
+
+TYPED_TEST_CASE(binning_directory_test, binning_directory_types);
+TYPED_TEST_CASE(binning_directory_compressed_test, binning_directory_compressed_types);
```
smehringer: what are you doing here? o.O
A typed test should be used with multiple types, otherwise, there is no point for a typed test. 
But you copy and pasted all tests for two typed test each testing only one type...?

include/seqan3/search/dream_index/binning_directory.hpp
```diff
@@ -0,0 +1,218 @@
...
+     */
+    binning_directory(size_t bins_, size_t bits_, size_t num_hash_ = 2)
+    {
+        assert(bins_ > 0);
+        assert(bits_ > 0);
+        assert(num_hash_ > 0);
+        if (bits_ < bins_)
+            throw std::logic_error{"There must be at least one bit per bin. Increase the current size of " +
+                                   std::to_string(bits_) + " to at least " + std::to_string(bins_) + "."};
+        bins = bins_;
```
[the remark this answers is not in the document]
eseiler: I can do it, but then I'll never catch the divide by `0` error in block_count, since it is done before I can do any asserts...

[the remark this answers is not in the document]
eseiler: Never mind, it's CRTP, so I cannot member initialise the parent's stuff. I could add a custom ctor for the parent and then use it in the member initialisation
```
binning_directory(...) : binning_directory_base{bins, num_hash, ...)
```
But I don't think it adds much to readability

include/seqan3/search/dream_index/detail/binning_directory_base.hpp
eseiler: I offer
```cpp
    [[nodiscard]] sdsl::bit_vector get(size_t const h) const
    {
        return get(h, sdsl::bit_vector(bins));
    }
```

I can't default it since the default depends on a non-static member:
https://godbolt.org/z/zanlhF

test/unit/search/dream_index/binning_directory_test.cpp
eseiler: It's from earlier and not refactored. I'll just do fixtures

include/seqan3/search/dream_index/binning_directory.hpp
```diff
@@ -0,0 +1,218 @@
...
+     */
+    inline void resize(size_t const new_bins)
+    {
+        if (new_bins < bins)
+        {
+            throw std::invalid_argument{"The new number of new bins must be greater"
+                                        " or equal to the current number of bins."};
+        }
+
+        size_t bin_width_ = (new_bins + 63) >> 6;
```
smehringer: Can you add a comment what happens here?

EDIT: Also `bin_width_` is not very expressive. Maybe `new_bin_width`

include/seqan3/search/dream_index/binning_directory.hpp
```diff
@@ -0,0 +1,218 @@
...
+     */
+    binning_directory_compressed() = default;                                                 //!< Defaulted.
+    binning_directory_compressed(binning_directory_compressed const &) = default;             //!< Defaulted.
+    binning_directory_compressed & operator=(binning_directory_compressed const &) = default; //!< Defaulted.
+    binning_directory_compressed(binning_directory_compressed &&) = default;                  //!< Defaulted.
+    binning_directory_compressed & operator=(binning_directory_compressed &&) = default;      //!< Defaulted.
+    ~binning_directory_compressed() = default;                                                //!< Defaulted.
+
+    /*!\brief Construct from seqan3::binning_directory.
+     * \param bd     The seqan3::binning_directory.
```
smehringer:
```suggestion
     * \param[in] bd     The seqan3::binning_directory.
```

include/seqan3/search/dream_index/binning_directory.hpp
```diff
@@ -0,0 +1,218 @@
...
+     */
+    binning_directory_compressed(binning_directory const & bd)
+    {
+        std::tie(bins, bin_width, block_size, block_count, num_hash, pre_hash) =
+            std::tie(bd.bins, bd.bin_width, bd.block_size, bd.block_count, bd.num_hash, bd.pre_hash);
+
+        data = sdsl::sd_vector<>{bd.data};
```
smehringer: This probably involves copying or is there a constructor that takes a temporary and is more efficient? Then I would consider that the constructor only takes temporaries

include/seqan3/search/dream_index/detail/binning_directory_base.hpp
smehringer: :+1:

include/seqan3/search/dream_index/detail/binning_directory_base.hpp
```diff
@@ -0,0 +1,209 @@
...
+
+    /*!\brief Returns the size of the underlying bitvector.
+     * \returns The size in bits of the underlying bitvector.
+     */
+    inline size_t size() const noexcept
+    {
+        return data.size();
+    }
+
+    /*!\brief Test for equality.
```
smehringer: Make those into a group `comparison operators`?

include/seqan3/search/dream_index/detail/binning_directory_base.hpp
```diff
@@ -0,0 +1,209 @@
...
+    {
+        return !(lhs == rhs);
+    }
+
+    /*!\brief Stream operator.
+     * \param s   The stream.
+     * \param rhs binning_directory to stream.
+     * \returns the stream.
+     */
+    friend auto& operator<<(auto& s, binning_directory_base const & rhs) // TODO proper overload for printing
```
smehringer: open TODO

test/performance/search/dream_index/binning_directory_benchmark.cpp
```diff
@@ -0,0 +1,144 @@
...
+        {
+            for (int32_t hash_num = 2; hash_num < 4; ++hash_num)
+            {
+                b->Args({bins, bits, hash_num, 100'000});
+            }
+        }
+    }
+}
+
+inline std::vector<size_t> generate_numbers(size_t const len = 500,
```
smehringer: You can use the `generate_sequence` function in `test/include/...`. It also works with numbers

test/performance/search/dream_index/binning_directory_benchmark.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+    auto gen = [&dist, &engine]() { return dist(engine); };
+    std::vector<size_t> sequence(len);
+    std::generate(std::ranges::begin(sequence), std::ranges::end(sequence), gen);
+
+    return sequence;
+}
+
+
+template <typename t>
```
smehringer: Be more expressive than `t` :)

test/performance/search/dream_index/binning_directory_benchmark.cpp
```diff
@@ -0,0 +1,144 @@
...
+        {
+            for (auto hash : hash_values)
+                res = bd.get(hash);
+        }
+
+        state.counters["hashes/sec"] = hashes_per_second(std::ranges::size(hash_values));
+    }
+};
+
+BENCHMARK_TEMPLATE_DEFINE_F(bd_benchmark, set, binning_directory) (benchmark::State & state)
```
smehringer: This looks over complicated..

I think it can be replaced with:

```cpp

auto set_up(size_t bins, size_t bits, size_t num_hash, size_t sequence_length)
{
    bin_values  = generate_numbers(sequence_length, 0u, bins - 1);
    hash_values = generate_numbers(sequence_length);
    binning_directory tmp(bins, bits, num_hash);
    bd = t(std::move(tmp));
    res.resize(bins);
    
    return {bin_values, hash_values, bd};
}

template <typename binning_directory_t>
void set_benchmark(::benchmark::State & state)
{
    auto && [ bin_values, hash_values, bd ] = set_up(state.range(0), state.range(1), state.range(2), state.range(3));
    
    for (auto _ : state)
    {
        for (auto [hash, bin] : std::view::zip(hash_values, bin_values))
            bd.set(hash, bin);
    }

    state.counters["hashes/sec"] = hashes_per_second(std::ranges::size(hash_values));
}

template <typename binning_directory_t>
void get_benchmark(::benchmark::State & state)
{
    auto && [ bin_values, hash_values, bd ] = set_up(state.range(0), state.range(1), state.range(2), state.range(3));
    
    for (auto _ : state)
    {
        for (auto hash : hash_values)
            res = bd.get(hash, std::move(res));
    }

    state.counters["hashes/sec"] = hashes_per_second(std::ranges::size(hash_values));
}

template <typename binning_directory_t>
void get_copy_benchmark(::benchmark::State & state)
{
    auto && [ bin_values, hash_values, bd ] = set_up(state.range(0), state.range(1), state.range(2), state.range(3));
    
    for (auto _ : state)
    {
        for (auto hash : hash_values)
            res = bd.get(hash);
    }

    state.counters["hashes/sec"] = hashes_per_second(std::ranges::size(hash_values));
}

BENCHMARK_TEMPLATE(set_benchmark, binning_directory)->Apply(arguments);
BENCHMARK_TEMPLATE(set_benchmark, binning_directory_compressed)->Apply(arguments);

BENCHMARK_TEMPLATE(get_benchmark, binning_directory)->Apply(arguments);
BENCHMARK_TEMPLATE(get_benchmark, binning_directory_compressed)->Apply(arguments);

BENCHMARK_TEMPLATE(get_copy_benchmark, binning_directory)->Apply(arguments);
BENCHMARK_TEMPLATE(get_copy_benchmark, binning_directory_compressed)->Apply(arguments);

```

test/unit/search/dream_index/binning_directory_test.cpp
smehringer: still open

include/seqan3/search/dream_index/all.hpp
```diff
@@ -0,0 +1,15 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
```
smehringer: check year everywhere

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,415 @@
...
+        if (num_hash_ == 0)
+            throw std::logic_error{"The number of hash functions must be > 0."};
+        if (bits_ < bins_)
+            throw std::logic_error{"There must be at least one bit per bin. Increase the current size of " +
+                                   std::to_string(bits_) + " to at least " + std::to_string(bins_) + "."};
+        bins = bins_;
+        num_hash = num_hash_;
+        data = sdsl::bit_vector(bits_);
+        bin_width   = (bins + 63) >> 6;
+        block_size  = bin_width << 6;
```
MitraDarja: Maybe comment why these shifts are necessary?

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,415 @@
...
+
+        std::vector<size_t> idx = pre_hash;
+
+        std::ranges::for_each(idx, [this, h] (size_t & i) { i *= h; hash_and_fit(i); });
+
+        for (size_t batch = 0; batch < bin_width; ++batch)
+        {
+           assert(idx[0] < data.size());
+           size_t tmp = data.get_int(idx[0]);
+           idx[0] += 64;
```
MitraDarja: Why add 64?

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,415 @@
...
+        return data.size();
+    }
+    //!\}
+
+    /*!\brief Increases the number of bins stored in the Interleaved Bloom Filter.
+     * \param[in] new_bins The new number of bins.
+     * \throws std::invalid_argument If passed number of bins is smaller than current number of bins.
+     *
+     * \attention This function is only available for **uncompressed** Interleaved Bloom Filters.
+     * \attention The new number of new bins must be greater or equal to the current number of bins.
```
MitraDarja: The new number of bins, delete new before bins

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,415 @@
...
+     *
+     * \details
+     *
+     * The resulting `seqan3::interleaved_bloom_filter` has an increased size proportional to the increase in the
+     * `bin_width` (the number of 64-bit words needed to represent `bins` many bins), e.g.
+     * resizing a `seqan3::interleaved_bloom_filter` with 40 bins to 73 bins also increases the `bin_width` from 1 to
+     * 2 and hence the new `seqan3::interleaved_bloom_filter` will be twice the size.
+     * This increase in size is necessary to avoid invalidating all computed hash functions.
+     * If you want to add more bins while keeping the size constant, you need to rebuild the interleaved_bloom_filter.
```
MitraDarja: Why ibf with underscores? Either write `seqan3::interleaved_bloom_filter` or Interleaved Bloom Filter

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,415 @@
...
+        requires data_layout_mode == data_layout::uncompressed
+    //!\endcond
+    {
+        if (new_bins < bins)
+            throw std::invalid_argument{"The value of new_bins must be >= the current number of bins."};
+
+        // Equivalent to ceil(new_bins / 64)
+        size_t new_bin_width = (new_bins + 63) >> 6;
+
+        if (new_bin_width == bin_width)
```
MitraDarja: Add comment, IBF remains the same, when # new bins == # old bins

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,415 @@
...
+        bin_width = new_bin_width;
+        block_size = new_block_size;
+    }
+
+    /*!\name Comparison operators
+     * \{
+     */
+    /*!\brief Test for equality.
+     * \param[in] lhs A seqan3::interleaved_bloom_filter.
+     * \param[in] rhs interleaved_bloom_filter to compare to.
```
MitraDarja: A seqan3::interleaved_bloom_filter to compare to.

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,415 @@
...
+     */
+    friend bool operator==(interleaved_bloom_filter const & lhs, interleaved_bloom_filter const & rhs) noexcept
+    {
+        return std::tie(lhs.bins, lhs.bin_width, lhs.block_size, lhs.block_count, lhs.num_hash, lhs.data) ==
+               std::tie(rhs.bins, rhs.bin_width, rhs.block_size, rhs.block_count, rhs.num_hash, rhs.data);
+    }
+
+    /*!\brief Test for inequality.
+     * \param[in] lhs A seqan3::interleaved_bloom_filter.
+     * \param[in] rhs interleaved_bloom_filter to compare to.
```
MitraDarja: A seqan3::interleaved_bloom_filter to compare to.

test/performance/search/dream_index/interleaved_bloom_filter_benchmark.cpp
```diff
@@ -0,0 +1,90 @@
...
+// Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+#include <benchmark/benchmark.h>
+
+#include <seqan3/search/dream_index/interleaved_bloom_filter.hpp>
+#include <seqan3/test/performance/sequence_generator.hpp>
+#include <seqan3/range/views/zip.hpp>
```
MitraDarja: #include <seqan3/range/views/zip.hpp> before <seqan3/test/performance/sequence_generator.hpp>

test/snippet/search/dream_index/interleaved_bloom_filter_constructor_compressed.cpp
```diff
@@ -0,0 +1,12 @@
+#include <seqan3/search/dream_index/interleaved_bloom_filter.hpp>
+
+int main()
+{
+    // Construct an uncompressed Interleaved Bloom Filter.
+    seqan3::interleaved_bloom_filter ibf{43, 8192, 3};
+
+    // Fill `ibf` with data.
```
MitraDarja: Should IBF not be capitalized?

test/snippet/search/dream_index/interleaved_bloom_filter_get.cpp
```diff
@@ -0,0 +1,17 @@
+#include <seqan3/search/dream_index/interleaved_bloom_filter.hpp>
+#include <seqan3/core/debug_stream.hpp>
```
MitraDarja: #include <seqan3/core/debug_stream.hpp> for #include <seqan3/search/dream_index/interleaved_bloom_filter.hpp> ?

test/unit/search/dream_index/interleaved_bloom_filter_test.cpp
```diff
@@ -0,0 +1,152 @@
...
+    static ibf_type make_ibf(size_t bins, size_t bits, size_t num_hash)
+    {
+        return ibf_type{seqan3::interleaved_bloom_filter{bins, bits, num_hash}};
+    }
+};
+
+using ibf_types = ::testing::Types<seqan3::interleaved_bloom_filter<seqan3::data_layout::uncompressed>,
+                                   seqan3::interleaved_bloom_filter<seqan3::data_layout::compressed>>;
+
+TYPED_TEST_SUITE(interleaved_bloom_filter_test, ibf_types, );
```
MitraDarja: Why is there a coma after ibf_types and then nothing?

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,360 @@
...
+    static constexpr size_t shift{27};
+    //!\brief Seed for hashing.
+    static constexpr size_t seed{0x90b45d39fb6da1fa};
+
+    /*!\brief Perturbs a value and fits it into the vector.
+     * \param h The value to process.
+     */
+    inline constexpr void hash_and_fit(size_t & h) const
+    {
+        h ^= h >> shift; // Basically Fibonacci hashing
```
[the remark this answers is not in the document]
eseiler: Well, the line is Fibonacci hashing with some extra step (https://probablydance.com/2018/06/16/fibonacci-hashing-the-optimization-that-the-world-forgot-or-a-better-alternative-to-integer-modulo/).

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,415 @@
...
+ * This datastructure represents an immutable Interleaved Bloom Filter, i.e. only querying is supported.
+ *
+ * ### Thread safety
+ *
+ * The Interleaved Bloom Filter promises the basic thread-safety by the STL that all
+ * calls to `const` member functions are safe from multiple threads (as long as no thread calls
+ * a non-`const` member function at the same time).
+ *
+ * Additionally, concurrent calls to `set` are safe iff each thread handles a multiple of wordsize (=64) many bins.
+ * For example, calls to `set` from multiple threads are safe iff `thread_1` accesses bins 0-63, `thread_2` bins 64-127,
```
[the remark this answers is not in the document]
eseiler: I think `iff` is a quite well known abbreviation...

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
eseiler: Because we read the first 64 bins, so in the next iteration, if there is one, we need to read the next 64 bins

test/snippet/search/dream_index/interleaved_bloom_filter_constructor_compressed.cpp
eseiler: It's the variable name. It comment should indicate that you do something with the uncompressed IBF and then compress it.

test/unit/search/dream_index/interleaved_bloom_filter_test.cpp
eseiler: https://github.com/google/googletest/issues/2271
Basically it's a bug in googletest and you need to put that comma there to not create a warning (=error for us).

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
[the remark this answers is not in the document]
smehringer: Maybe add the link as a comment?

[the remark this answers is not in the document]
MitraDarja: I agree, it would be nice to have the link

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,415 @@
...
+        bin_width = new_bin_width;
+        block_size = new_block_size;
+    }
+
+    /*!\name Comparison operators
+     * \{
+     */
+    /*!\brief Test for equality.
+     * \param[in] lhs A `seqan3::interleaved_bloom_filter`.
+     * \param[in] rhs `seqan3::interleaved_bloom_filter` to compare to.
```
MitraDarja: If you have an A in the first line, shouldn't it also have an A in the second line? So, A `seqan3::interleaved_bloom_filter` to compare to?

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,415 @@
...
+     */
+    friend bool operator==(interleaved_bloom_filter const & lhs, interleaved_bloom_filter const & rhs) noexcept
+    {
+        return std::tie(lhs.bins, lhs.bin_width, lhs.block_size, lhs.block_count, lhs.num_hash, lhs.data) ==
+               std::tie(rhs.bins, rhs.bin_width, rhs.block_size, rhs.block_count, rhs.num_hash, rhs.data);
+    }
+
+    /*!\brief Test for inequality.
+     * \param[in] lhs A `seqan3::interleaved_bloom_filter`.
+     * \param[in] rhs `seqan3::interleaved_bloom_filter` to compare to.
```
MitraDarja: If you have an A in the first line, shouldn't it also have an A in the second line? So, A `seqan3::interleaved_bloom_filter` to compare to?

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,518 @@
...
+        result_buffer.resize(bins_);
+    }
+    //!\brief Resize `result_buffer` instead of copying.
+    interleaved_bloom_filter & operator=(interleaved_bloom_filter const & rhs)
+    {
+        std::tie(bins_, bin_size_, bin_width, block_size, block_count, data, pre_hash) =
+            std::tie(rhs.bins_, rhs.bin_size_, rhs.bin_width, rhs.block_size, rhs.block_count, rhs.data, rhs.pre_hash);
+        result_buffer.resize(bins_);
+        return *this;
+    }
```
eseiler: Since we delete the copy ctor/assignment for the `binning_bitvector`, we have to customise both here.

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,518 @@
...
+     *
+     * \attention The result of this function must always be bound via reference, e.g. `auto &`.
+     *
+     * \details
+     *
+     * ### Example
+     *
+     * \include test/snippet/search/dream_index/interleaved_bloom_filter_bulk_contains.cpp
+     */
+    [[nodiscard]] binning_bitvector const & bulk_contains(size_t const value) const & noexcept
```
eseiler: Since we deleted copy ctor/assignment for `binning_bitvector`, we cannot do `auto res = ibf. bulk_contains(23)`, which would copy the `binning_bitvector` into `res`.
We enforce `auto & ref = ...`

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,518 @@
...
+
+           result_buffer.set_int(batch<<6, tmp);
+        }
+
+        return result_buffer;
+    }
+
+    // `bulk_contains` cannot be called on a temporary, since the the object the returned reference points to
+    // is immediately destroyed.
+    [[nodiscard]] binning_bitvector const & bulk_contains(size_t const value) const && noexcept = delete;
```
eseiler: Not really needed, but can prevent misuse, e.g. `seqan3::interleaved_bloom_filter{/*ctor args*/}.bulk_contains(23);`

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,518 @@
...
+
+/*!\brief A strong type that represents the number of hash functions for the seqan3::interleaved_bloom_filter.
+ * \relates seqan3::interleaved_bloom_filter
+ */
+struct hash_function_count : public detail::strong_type<size_t, hash_function_count, detail::strong_type_skill::convert>
+{
+    using detail::strong_type<size_t, hash_function_count, detail::strong_type_skill::convert>::strong_type;
+};
+
+/*!\brief A strong type that represents the a bin index for the seqan3::interleaved_bloom_filter.
```
MitraDarja: Without the a after the

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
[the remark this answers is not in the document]
MitraDarja: So, we do not provide a link?

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,518 @@
...
+               idx[i] += 64;
+           }
+
+           result_buffer.set_int(batch<<6, tmp);
+        }
+
+        return result_buffer;
+    }
+
+    // `bulk_contains` cannot be called on a temporary, since the the object the returned reference points to
```
MitraDarja: two the, delete one of them.:)

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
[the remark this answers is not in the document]
eseiler: Now we do :)

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,521 @@
...
+public:
+    //!\brief A bitvector representing the result of a call to `bulk_contains` of the seqan3::interleaved_bloom_filter.
+    class binning_bitvector : sdsl::bit_vector
+    {
+    public:
+        /*!\name Constructors, destructor and assignment
+         * \{
+         */
+        binning_bitvector() = default; //!< Defaulted.
+        binning_bitvector(binning_bitvector const &) = delete; //!< Deleted to prevent copying.
```
smehringer: Do not delete this and adjust the IBF constructors accordingly

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,521 @@
...
+     */
+    interleaved_bloom_filter(seqan3::bin_count bin_,
+                             seqan3::bin_size bits_,
+                             seqan3::hash_function_count funs = seqan3::hash_function_count{2u})
+    //!\cond
+        requires data_layout_mode == data_layout::uncompressed
+    //!\endcond
+    {
+        bins = bin_.get();
+        size_t bits = bits_.get();
```
smehringer: Change the usage of `bits` since this is now the bin size and not the bit size!

test/include/seqan3/test/performance/sequence_generator.hpp
```diff
@@ -55,6 +56,23 @@ auto generate_sequence(size_t const len = 500,
     return sequence;
 }
 
+template <arithmetic number_type>
+auto generate_numeric_sequence(size_t const len = 500,
+                               number_type const min = std::numeric_limits<number_type>::lowest(),
+                               number_type const max = std::numeric_limits<number_type>::max(),
+                               size_t const seed = 0)
+{
+    std::mt19937 engine(seed);
```
eseiler:
```suggestion
    std::mt19937_64 engine(seed);
```

eseiler: + changelog
+ hash_and_fit

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,501 @@
...
+     * \attention This constructor can only be used to construct **uncompressed** Interleaved Bloom Filters.
+     *
+     * \details
+     *
+     * ### Example
+     *
+     * \include test/snippet/search/dream_index/interleaved_bloom_filter_constructor.cpp
+     */
+    interleaved_bloom_filter(seqan3::bin_count bins_,
+                             seqan3::bin_size bin_size__,
```
smehringer: two underscores :see_no_evil: Maybe just call it size here in this very small context since the strong type is already expressive enough?
```suggestion
                             seqan3::bin_size size,
```

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,501 @@
...
+        data = sdsl::sd_vector<>{ibf.data};
+        result_buffer.resize(bins);
+    }
+    //!\}
+
+    /*!\name Modifiers
+     * \{
+     */
+    /*!\brief Inserts a value into a specific bin.
+     * \param[in] h   The raw hash value to process.
```
smehringer: Since we do not require a hash value here but an arbitrary number, I think we decided to name it like that?
```suggestion
     * \param[in] value   The raw numeric value to process.
```
Same for all places except in hash_and_fit

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,501 @@
...
+     * \param[in] bin The bin to insert into.
```
smehringer:
```suggestion
     * \param[in] bin The bin index to insert into.
```

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,501 @@
...
+            size_t idx = val * h;
+            hash_and_fit(idx);
+            idx += bin.get();
+            assert (idx < data.size());
+            data[idx] = 1;
+        };
+    }
+
+    /*!\brief Increases the number of bins stored in the Interleaved Bloom Filter.
+     * \param[in] new_bins_ The new number of bins.
```
smehringer: Why an underscore?
```suggestion
     * \param[in] new_bins The new number of bins.
```

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,501 @@
...
+    }
+    //!\}
+
+    /*!\name Lookup
+     * \{
+     */
+    /*!\brief Determines set membership of a given value.
+     * \param[in] value The raw value to process.
+     *
+     * \attention The result of this function must always be bound via reference, e.g. `auto &`.
```
smehringer:
```suggestion
     * \attention The result of this function must always be bound via reference, e.g. `auto &` to prevent copying.
```

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
eseiler: Because it's the strong type

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -164,14 +164,20 @@ class interleaved_bloom_filter
...
     size_t bin_size_{};
+    //!\brief How many bits do we need to represent `bin_size_`?
+    size_t bin_size_in_bits{};
     //!\brief The number of 64-bit integers needed to store `bins` many bits (e.g. `bins = 50` -> `bin_words = 1`).
     size_t bin_words{};
     //!\brief The number of hash functions.
     size_t hash_funs{};
     //!\brief The bitvector.
     data_type data{};
     //!\brief Precalculated seeds for hashing.
```
smehringer:
```suggestion
    //!\brief Precalculated seeds for multiplicative hashing. We use large irrational numbers for a uniform hashing.
```

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -182,6 +188,8 @@ class interleaved_bloom_filter
      */
     constexpr void hash_and_fit(size_t & h) const
     {
+        assert(bin_size_in_bits != 64);
+        h ^= h >> (64 - bin_size_in_bits); // Shift higher bits into lower bits
```
smehringer:
```suggestion
        h ^= h >> (64 - bin_size_in_bits); // XOR higher bits into lower bits to avoid information loss
```

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -164,14 +164,20 @@ class interleaved_bloom_filter
     size_t technical_bins{};
     //!\brief The size of each bin in bits.
     size_t bin_size_{};
+    //!\brief How many bits do we need to represent `bin_size_`?
+    size_t bin_size_in_bits{};
```
smehringer:
```
    //!\brief The number of bits to shift the hash value before doing multiplicative hashing (detail::count_leading_zeros(bin_size_)).
    size_t hash_shift{};
```

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -293,12 +298,12 @@ class interleaved_bloom_filter
     //!\endcond
     {
         assert(bin.get() < bins);
-        for (auto const & seed : hash_seeds)
+        for (size_t i = 0; i < hash_funs; ++i)
         {
-            size_t idx = seed * value;
+            size_t idx = hash_seeds[i] * value;
```
smehringer: move this inside of hash_and_fit and add a seed parameter to hash and fit

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -386,9 +391,14 @@ class interleaved_bloom_filter
     {
         assert(result_buffer.size() == bin_count());
 
-        std::vector<size_t> idx = hash_seeds;
+        std::array<size_t, 5> idx;
```
smehringer:
```suggestion
        std::array<size_t, 5> bloom_filter_indices; // which bits to check in each individual bloom filter
```
? `idx` does not tell me anything

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,510 @@
...
+           assert(idx[0] < data.size());
+           size_t tmp = data.get_int(idx[0]);
+           idx[0] += 64;
+
+           for (size_t i = 1; i < hash_funs; ++i)
+           {
+               assert(idx[i] < data.size());
+               tmp &= data.get_int(idx[i]);
+               idx[i] += 64;
+           }
```
smehringer:
```cpp
           size_t tmp{0b1111111111111111111111111111111111111111111111111111111111111111};
           for (size_t i = 0; i < hash_funs; ++i)
           {
               assert(idx[i] < data.size());
               tmp &= data.get_int(idx[i]);
               idx[i] += 64;
           }
```

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -0,0 +1,510 @@
...
+
+           result_buffer.set_int(batch<<6, tmp);
```
smehringer:
```suggestion
           result_buffer.set_int(batch << 6, tmp);
```

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
eseiler: We could also do

```
        if constexpr(data_layout_mode == data_layout::uncompressed)
        {
            std::memcpy(result_buffer.data(), data.data() + (idx[0] >> 6), technical_bins >> 3);
            for (size_t i = 1; i < hash_funs; ++i)
            {
                std::memcpy(buf.data(), data.data() + (idx[i] >> 6), technical_bins >> 3);
                result_buffer &= buf;
            }
        }
```

This gets better with a bigger bin number, breaking even at around 256 bins

PR #922, opened 2019-04-17
[FEATURE] shape and shape_iterator

vinzenzmay:
please find some questions in the code comments

### DIFF
--- a/include/seqan3/search/kmer_index/shape_iterator.hpp
+++ b/include/seqan3/search/kmer_index/shape_iterator.hpp
@@ -1,0 +1,197 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \author Enrico Seiler <enrico.seiler AT fu-berlin.de>
+ * \author Vinzenz May <vinzenz.may AT fu-berlin.de>
+ * \brief Provides the seqan3::shape_iterator.
+ */
+
+#pragma once
+
+#include <seqan3/alphabet/concept.hpp>
+#include <seqan3/search/kmer_index/shape.hpp>
+#include <seqan3/std/ranges>
+
+namespace seqan3
+{
+
+/*!\brief Iterator for calculating hash values via a given seqan3::shape.
+ * \tparam it_t Type of iterator on text. Must model std::ForwardIterator. Value type must model seqan3::Semialphabet.
+*/
+template <std::ForwardIterator it_t>
+    //!\cond
+    requires Semialphabet<value_type_t<it_t>>
+    //!\endcond
+class shape_iterator
+{
+private:
+    //!\brief The alphabet type of the passed iterator.
+    using alphabet_t = value_type_t<it_t>;
+
+    //!\brief The hash value.
+    size_t hash_value{0};
+
+    //!\brief The seqan3::shape to use for hashing.
+    shape const s;
+
+    //!\brief The factor for the left most position of the hash value.
+    size_t roll_factor{0};
+
+    //!\brief True if `s` contains only 1 (ungapped), false otherwise.
+    bool ungapped_shape{false};
+
+    //!\brief Iterator pointing to begin of the underlying text.
+    it_t const text_start;
+
+    //!\brief Calculates a hash value by explicitly looking at each position.
+    void hash_full()
+    {
+        text_right = text_left;
+        hash_value = s[0] * to_rank(*(text_right));
+
+        for (size_t i{1}; i < std::ranges::size(s); ++i)
+        {
+            std::advance(text_right, 1);
+
+            hash_value *= alphabet_size_v<alphabet_t>;
+            hash_value += s[i] * to_rank(*(text_right));
+        }
+    }
+
+    //!\brief Calculates a hash value by using rolling hash.
+    void hash_roll()
+    {
+        hash_value -= to_rank(*(text_left)) * roll_factor;
+
+        std::advance(text_left,  1);
+        std::advance(text_right, 1);
+
+        hash_value *= alphabet_size_v<alphabet_t>;
+        hash_value += to_rank(*(text_right));
+    }
+
+public:
+    //!\brief Iterator to the leftmost position of the k-mer.
+    it_t text_left;
+    //!\brief Iterator to the rightmost position of the k-mer.
+    it_t text_right;
+
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    shape_iterator()                                   = default; //!< Defaulted.
+    shape_iterator(shape_iterator const &)             = default; //!< Defaulted.
+    shape_iterator(shape_iterator &&)                  = default; //!< Defaulted.
+    shape_iterator & operator=(shape_iterator const &) = default; //!< Defaulted.
+    shape_iterator & operator=(shape_iterator &&)      = default; //!< Defaulted.
+    ~shape_iterator()                                  = default; //!< Defaulted.
+
+    /*!\brief Construct from a given iterator on the text and a seqan3::shape.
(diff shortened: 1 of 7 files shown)


include/seqan3/search/kmer_index/shape_iterator.hpp
```diff
@@ -0,0 +1,86 @@
...
+    shape s;
+
+public:
+    it_t text_left, text_right;
+
+    shape_iterator(it_t it_start, shape s_) : s(s_),
+                                    text_left(it_start),
+                                    text_right(it_start)
+    {
+        if (s.size()) // delete for better performance? - generates seg fauls if s.size() == 0 !
```
eseiler: Just do an `assert(s.size() > 0);`.
We will document it as undefined behaviour if you pass an empty shape.

include/seqan3/search/kmer_index/shape_iterator.hpp
```diff
@@ -0,0 +1,86 @@
...
+    bool operator==(const it_t& it)           { return text_right == it; }
+    bool operator==(const shape_iterator& it) { return text_right == it.text_left; }
+
+    bool operator!=(const it_t& it)           { return text_right != it; }
+    bool operator!=(const shape_iterator& it) { return text_right != it.text_left; }
+
+    void operator ++()
+    {
+
+        hash_value -= to_rank(*(text_left)) * pow(alphabet_size_v<alphabet_t>, std::ranges::size(s)-1);
```
eseiler: You should store the `pow(alphabet_size_v<alphabet_t>, std::ranges::size(s)-1)` in a member variable since this is a costly operation

test/performance/search/shape_iterator_benchmark.cpp
```diff
@@ -0,0 +1,82 @@
...
+#include <seqan3/search/kmer_index/shape.hpp>
+#include <seqan3/search/kmer_index/shape_iterator.hpp>
+
+using namespace seqan3;
+
+
+// ===========================================================================
+//  taken from affine_alignment_benchmark.cpp
+template <typename alphabet_t>
+auto generate_sequence_seqan3(size_t const len = 500,
```
eseiler: This is now in `seqan3/text/performance/sequence_generator.hpp`

test/performance/search/shape_iterator_benchmark.cpp
```diff
@@ -0,0 +1,82 @@
...
+    for (auto _ : state)
+        shape s();
+}
+
+static void shape_construct_benchmark(benchmark::State& state)
+{
+    size_t size = state.range(0);
+    for (auto _ : state)
+        shape s(size);
+}
```
eseiler: I think we don't need to benchmark the shape construction

test/performance/search/shape_iterator_benchmark.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+// performance tests auf dna4 ausreichend?
```
eseiler: For now yes

test/performance/search/shape_iterator_benchmark.cpp
```diff
@@ -0,0 +1,82 @@
...
+static void shape_iterator_construct(benchmark::State& state)
```
eseiler: We also don't really need to benchmark this

test/unit/search/kmer_index/shape_iterator_tests.cpp
```diff
@@ -0,0 +1,124 @@
...
+    EXPECT_EQ(std::ranges::size(shape(1,0,1,0,1,1)), 6u);
+    EXPECT_EQ(std::ranges::size(shape(1,1,1,1,1,1)), 6u);
+}
+
+// size 3 ausreichend?
+TEST(ShapeTest, HandlesShapes)
+{
+    shape s0(0,0,0);
+    std::array<bool,3> a0{false,false,false};
+    for (size_t i(0); i < a0.size(); i++)
```
eseiler: `EXPECT_TRUE(std::ranges::equal(s0.get(), a0));`

test/unit/search/kmer_index/shape_iterator_tests.cpp
```diff
@@ -0,0 +1,124 @@
...
+{
+
+    std::array<size_t,3> expected_hashes{1,6,11};
+    dna4_vector dna4_sample{"ACGT"_dna4};
+
+    shape shape2(2);
+    shape_iterator it(dna4_sample.begin(), shape2);
+
+    EXPECT_EQ(*it,expected_hashes[0]);
+    ++it; // *(++it) funktioniert nicht. warum?
```
eseiler: `*it; ++it` != `*++it`
`*++it` == `++it; *it`
`*it; ++it` == `*it++`, but produces a copy

vinzenzmay: I just couldn't compile *++it. Postfix operator is not defined in shape_iterator.
`
EXPECT_EQ(*it,expected_hashes[1]);
    ++it;
`
could not be written as
`EXPECT_EQ(*++it,expected_hashes[1]);`
Compiler says:

```
error: void value not ignored as it ought to be
     EXPECT_EQ(*++it,expected_hashes[1]);
```
I guess it would be more neat, that's all.

test/performance/search/shape_iterator_benchmark.cpp
vinzenzmay: Despite the first hash value is calculated upon construction?

test/unit/search/kmer_index/shape_iterator_tests.cpp
vinzenzmay: `EXPECT_TRUE(std::ranges::equal(s0.get(), a0));`
returns false when it should be true.

test/performance/search/shape_iterator_benchmark.cpp
eseiler: Well it's only one hash value. The typical usecase will probably be iterating over a sequence and generating hashes.

eseiler: You can leave this one in if you want, though.

test/unit/search/kmer_index/shape_iterator_tests.cpp
eseiler:
```cpp
TEST(ShapeTest, HandlesShapes)
{
    shape s0(0, 0, 0);
    std::array<bool, 32> a0{false, false, false};
    EXPECT_EQ(s0.get(), a0);

    shape s1(1, 0, 1);
    std::array<bool, 32> a1{true, false, true};
    EXPECT_EQ(s1.get(), a1);

    shape s2(3);
    std::array<bool, 32> a2;
    a2.fill(true);
    EXPECT_EQ(s2.get(), a2);
}
```

test/unit/search/kmer_index/shape_iterator_tests.cpp
eseiler: I mean the main point here is, that comparing `*++it` is semantically wrong since you first want to compare the value and then increment `it`. `*++it` would first increment `it` *and then* dereference it. Hence you'd be comparing the wrong values.

You could also implement a postincrement, but this would also have to create a copy.

Anyway, having the `++` on the next line is more readable, so I'd leave it like that.

vinzenzmay: I was aware of that. I wanted the increment in the last two calls, not in the first one. Well, in that case, I'll leave it as it is.

eseiler: Ah 💡 

Your `operator++` should look like this: `shape_iterator & operator++()` and have a `return *this;` at the end.
Then it works.

vinzenzmay: yes, that makes sense, thank you!

test/unit/search/kmer_index/shape_iterator_tests.cpp
vinzenzmay: so this doesn't work because the strong types get()-method returns an array of type <type, 32>. They can never be equal as long as I check against an array with length 3.
Because of that, I'd suggest to leave it as it is.

eseiler: That's why I wrote  `std::array<bool, 32> a0{false, false, false};` with a 32 and not a 3 😁

vinzenzmay: Ok, that's convenient :)

eseiler:
superseded by #946

PR #923, opened 2019-04-18
[FEATURE] view::repeat_n 

smehringer:


include/seqan3/range/view/repeat_n.hpp
```diff
@@ -0,0 +1,92 @@
...
+} // namespace seqan3::detail
+
+namespace seqan3::view
+{
+
+/*!\name General purpose views
+ * \{
+ */
+/*!\brief A view factory that repeats a given value infinitely.
+ * \tparam    value_t  The type of value to repeat wrapped in a std::view::single; must model std::CopyConstructible.
```
joergi-w: The term "must model..." is established for concepts, right? Then please note that `repeat_n` uses `static_assert` instead of a concept requirement.

test/unit/range/view/view_repeat_n_test.cpp
```diff
@@ -0,0 +1,109 @@
...
+        EXPECT_EQ(v.size(), 2u);
+        EXPECT_EQ(*v.begin(), str);
+        EXPECT_EQ(v[0], str);
+    }
+
+    // view
+    {
+        auto view = std::string{"foobar"} | view::persist | std::view::take(3);
+        auto v = view::repeat_n(view, 5);
+        EXPECT_TRUE(std::ranges::equal(*v.begin(), view));
```
joergi-w: It does not become clear what the expected outcome is: 
view = "foo".
v = "foofoofoofoofoo" or "foofo"? (implementation says `take_exactly(5)`)
*v.begin() = "f"?

Please prepare a snippet for `repeat_n` and write comments or more explicit documentation.

test/unit/range/view/view_repeat_n_test.cpp
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    // combinability
+    {
+        std::string str{"foobar"};
+        auto v = view::repeat_n(str, 5) | view::take_exactly(3);
+        EXPECT_EQ(*v.begin(), str);
+        EXPECT_EQ(std::ranges::size(v), 3u);
```
joergi-w: Also here some comments about the view contents would be helpful.

marehr:
some of your tests are failing due to a broken master; please rebase after #926 was merged.

include/seqan3/range/view/repeat_n.hpp
```diff
@@ -0,0 +1,92 @@
...
+{
+    /*!\brief todo
+     * \tparam irng_t The type of the range being processed. See seqan3::view::trim for requirements.
+     * \param irange The range being processed.
+     * \param threshold The minimum quality as a phred score [integral type].
+     */
+    template <typename value_t>
+    constexpr auto operator()(value_t && value, size_t const count) const
+    {
+        static_assert(std::CopyConstructible<value_t>, "The value passed to repeat_n must be copy constructible.");
```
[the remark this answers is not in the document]
smehringer: I think this is included somewhere in repeat.hpp or take_exaclty.hpp because the compiler does not complain :)

include/seqan3/range/view/repeat_n.hpp
smehringer: well but I static_assert the concept, so I thought it still `models` it? I would rather not document how we enforce the concept (requires or static_assert) for the user. They only have to know **that it is** enforced. I chose static_assert because of better error messages :) and we decided on a strategy meeting that whenever the concept is not required for overload resolution, a static_assert is more user-friendly. I think I forgot to write this down after the meeting for everyone. Sorry :(

test/unit/range/view/view_repeat_n_test.cpp
smehringer: You are right the documentation is not nice so far! `v = [foo, foo, foo, foo, foo]` :grin: 
I will add a snippet

test/unit/range/view/view_repeat_n_test.cpp
smehringer: I hope the test is more expressive now :)

include/seqan3/range/view/repeat_n.hpp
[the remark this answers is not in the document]
joergi-w: I have in mind that we do not want these chain-dependencies: If you change code and for some reason want to remove the header that itself includes <concepts>, it is not visible that this header is still required. Additionally, if we rely on compiler complaints, we create situations, where the order of include statements matters. Isn't the purpose of our header tests to prevent that?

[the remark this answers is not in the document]
smehringer: I included the header.

PR #925, opened 2019-04-18
[FIX] clean up alignment tests

joergi-w:
Alignment Tests: This PR moves the actual test definitions to *.cpp files and the common functions to *_test_template.hpp files. The directory structure analogous to the `include` folder is recovered.


marehr:
some of your tests are failing due to a broken master; please rebase after #926 was merged.

smehringer:
> And may I ask when we decided some of the changes? Why did you delete the fixtures and put them directly into the test files? I for example reuse fixtures in other fixtures and there are some advantages to keep them "global". Furthermore you put my max_errors fixture in the normal fixtures, but they are semantically different and I think they should be in separate files.

I will ask back: when did we decide to make global fixtures :grin:? Because as far as I remember we decided to keep the test structure just like the module/submodule structure (which I was surprised was not the case in alignment because of an extra fixture directory but I probably did not review the PR)

I agree that reusable code should be "global" but the way it is now (in master not in this PR), I did not see resuse of code, except of alignment_fixture.hpp (which I agree to keep). I only saw that the test.cpp's were duplicated code (-> test template) and that all relevant code was not in the test.cpp but in the fixture.hpp which I think is unintuitive? (and not done anywhere else)

I would propose to only keep stuff that is reused right now in a "global" space but everything that is not resused right now should go into the corresponding test.cpp. If you reuse some fixture at some point we can out-source those code parts again. Otherwise I think it is confusing why some parts are global and some are not. In any way I vote against an extra fixture directory :)

> Can you please make the changes so that git also recognise that you first renamed some files and afterwards removed content?

I guess this makes changes better trackable :+1:

rrahn:
In general, I agree with most of the points. And the organisation into test templates is anyway an outstanding task. So better now then later. But regarding the test fixtures, I really want to have them not removed or changed in this PR. There is a lot of ongoing code for the alignment stuff and it is very likely that a lot of stuff will be changed/adapted/touched in the next couple of weeks. Refactoring and cleanup stuff in the test directory has absolutely no priority right now and for the release. The same goes for the benchmarks. The point is that testing the alignment code with all it's configurations is quite cumbersome and will have a complexity on it's own. Whether or not er keep the test fixtures or where they should live is not a decision that we should make now but when we have most of the code done (including parallelisation/vectorisation parts).

marehr:
I think this PR is outdated. I think a complete new PR is needed to make those changes.

PR #931, opened 2019-04-24
Change license link

h-2:
a small fix to the license file and changing the link everywhere.


h-2:
This touches every file so it would be nice to get in soonish...

PR #935, opened 2019-04-26
[MISC] constexpr_string -> small_string

smehringer:


include/seqan3/range/container/small_vector.hpp
```diff
@@ -315,13 +315,13 @@ class small_vector
         return &data_[0];
     }
 
-    //!\copydoc small_vector::begin().
+    //!\overload
```
eseiler: Why not both copydoc?

test/unit/range/container/small_string_test.cpp
```diff
@@ -0,0 +1,301 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+#include <gtest/gtest.h>
+
+#include <seqan3/test/pretty_printing.hpp>
```
eseiler: Lexicographical order (all includes)

include/seqan3/range/container/small_vector.hpp
smehringer: well because one of them is an overload :D and `//!\overload` is safer to use because the copydoc link generation changes (one of the errors in the newest doxygen version).

include/seqan3/range/container/small_string.hpp
```diff
@@ -0,0 +1,260 @@
...
+/*!\brief Implements a small string that can be used for compile time computations.
+ * \ingroup container
+ * \tparam N The size of the small string.
+ *
+ * This class provides a string type that can be constructed, evaluated and operated on at compile time.
+ * The stored string can be accessed as either a std::string or a c-style string through the respective member functions.
+ * All operations are constexpr except the conversion functions.
+ *
+ * Internally the string stores an array of size `N+1` containing the `0`-byte to allow conversion to
+ * c-style strings. This also means that when creating an instance of this string from a string literal containing
```
eseiler:
```suggestion
 * c-style strings. This also means that when creating an instance of `seqan3::small_string` from a string literal containing
```

And maybe a small example

include/seqan3/range/container/small_vector.hpp
```diff
@@ -333,13 +333,13 @@ class small_vector
...
 
-    //!\copydoc small_vector::end().
+    //!\copydoc seqan3::small_vector::end()
     constexpr const_iterator end() const noexcept
     {
         return &data_[sz];
     }
 
-    //!\copydoc small_vector::end().
+    //!\copydoc seqan3::small_vector::end()
```
eseiler:
```suggestion
    //!\copydoc seqan3::small_vector::cend()
```

include/seqan3/range/container/small_vector.hpp
```diff
@@ -333,13 +333,13 @@ class small_vector
...
```
eseiler:
```suggestion
    //!\overload
```

?

include/seqan3/range/container/small_vector.hpp
smehringer: no because this is in the small_vector.hpp ! :D I just noticed myself. it has yo be end()

include/seqan3/range/container/small_vector.hpp
smehringer: I changed it to always use copydoc because the documentation looks better

doc/howto/write_a_view/solution_view.cpp
```diff
@@ -175,7 +175,7 @@ int main()
 
 //![main_adaptor]
     /* try the adaptor */
-    auto v2 = vec | std::view::reverse | view::my;
+    auto v2 = vec | std::view::reverse | ::view::my;
```
h-2: ?

include/seqan3/core/detail/reflection.hpp
```diff
@@ -112,20 +112,18 @@ struct get_display_name
         for (; *name_ptr == ' '; ++name_ptr)
         {}
 
-        for (unsigned i = 0; i < tmp.size(); ++i, ++name_ptr)
-        {
-            tmp[i] = *name_ptr;
-        }
+        tmp.insert(tmp.begin(), name_ptr, name_ptr + tmp.capacity());
```
h-2: not .size()?

include/seqan3/range/container/small_string.hpp
```diff
@@ -0,0 +1,265 @@
...
+    using base_t::base_t;
+    //!\brief Inherit the parent class member function assign.
+    using base_t::assign;
+
+    //!\brief The default constructor initializes the underlying data member with '\0's.
+    constexpr small_string() noexcept : base_t{}
+    {
+        for (size_t i = 0; i < capacity_ + 1; ++i)
+            this->data_[i] = '\0';
+    }
```
h-2: why doesn't `base_t{}` do the zeroing?

include/seqan3/range/container/small_string.hpp
```diff
@@ -0,0 +1,265 @@
...
+#pragma once
+
+#include <seqan3/range/container/small_vector.hpp>
+
+namespace seqan3
+{
+
+/*!\brief Implements a small string that can be used for compile time computations.
+ * \ingroup container
+ * \tparam N The size of the small string.
```
h-2: add `* \implements seqan3::ReservableContainer`

h-2: The tparam is not `N` and it's not the size!! Why was there no warning for this?

include/seqan3/range/container/small_string.hpp
```diff
@@ -0,0 +1,265 @@
...
+ *
+ * This class provides a string type that can be constructed, evaluated and operated on at compile time.
+ * The stored string can be accessed as either a std::string or a c-style string through the respective member functions.
+ * All operations are constexpr except the conversion functions.
```
h-2: What do you think about:

This class provides a string type for small strings and compile-time contexts. It has fixed capacity, but variable size within the capacity. It is always allocated on the stack and most of it's members are `constexpr`-qualified. The underlying data can be exposed as a null-terminated c-style string (without copying) and conversion operators to std::string are provided (this involves copying).

include/seqan3/range/container/small_string.hpp
```diff
@@ -0,0 +1,265 @@
...
+ *
+ * Internally the string stores an array of size `N+1` containing the `0`-byte to allow conversion to
+ * c-style strings. This also means that when creating an instance of seqan3::small_string from a string literal
+ * containing a `0`-byte the size() of this small_string instance is one less than that of the literal.
```
h-2: and instead of this (the second half is confusing because you are talking about size again, although it should be about capacity):


### Implementation notes

Internally the string stores a null-terminated array of size `capacity_ + 1` and the size of the string as a member. The smallest possible type is used for storage of the size so `small_string<30>` uses 32bytes of memory (one byte extra for the null-terminator and one byte to save the size).

include/seqan3/range/container/small_string.hpp
```diff
@@ -0,0 +1,265 @@
...
+ *
+ * Example:
+ *
+ * \include test/snippet/range/container/small_string.cpp
+ *
+ */
+template <size_t capacity_>
+class small_string : public small_vector<char, capacity_ + 1>
+{
+    //!\brief The parent class.
```
h-2: private:

include/seqan3/range/container/small_string.hpp
```diff
@@ -0,0 +1,265 @@
...
+    constexpr small_string & operator=(char const (&_lit)[N]) noexcept
+    {
+        static_assert(N <= capacity_ + 1, "Length of string literal exceeds capacity of small_string.");
+        assign(_lit);
+        return *this;
+    }
+
+    /*!\brief Assign from literal.
+     * \param _lit The literal to assign the string from.
+     *
```
h-2: Maybe add a sentence here that char arrays are expected to be null-terminated and that otherwise the last character is lost?

include/seqan3/range/container/small_string.hpp
```diff
@@ -0,0 +1,265 @@
...
+     * Linear in the size of the string.
+     */
+    std::string string() const
+    {
+        return std::string{this->cbegin(), this->cend()};
+    }
+
+    /*!\brief Returns the content represented as 0-terminated c-style string.
+     *
+     * \returns `const char *` The stored string.
```
h-2:
```suggestion
     * \returns `char const *` The stored string.
```

include/seqan3/range/container/small_string.hpp
```diff
@@ -0,0 +1,265 @@
...
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     *
+     * ### Complexity
+     *
+     * Constant.
+     */
+    constexpr const char * c_str() const noexcept
```
h-2:
```suggestion
    constexpr char const * c_str() const noexcept
```

include/seqan3/range/container/small_string.hpp
```diff
@@ -0,0 +1,265 @@
...
+    }
+    //!\}
+
+    /*!\name Modifiers
+     * \{
+     */
+    //!\copydoc seqan3::small_vector::clear
+    constexpr void clear() noexcept
+    {
+        *this = small_string{};
```
h-2: Probably not a big thing, but shouldn't it be enough to set size to 0 and [0]=0 ? That's constant the current call is linear...

include/seqan3/range/container/small_string.hpp
```diff
@@ -0,0 +1,265 @@
...
+    {
+        assert(count <= capacity_);
+
+        for (size_t i = this->sz; i <= count; ++i)
+            this->data_[i] = '\0';
+
+        this->sz = count;
+        this->data_[this->sz] = '\0';
+    }
+    //!\}
```
h-2: you are only providing one of the resize() overloads, what about the other one? How does that one ensure that the string end null-terminated? What about `insert()` at the end?

include/seqan3/range/container/small_string.hpp
```diff
@@ -0,0 +1,265 @@
...
+    //!\copydoc seqan3::small_vector::resize(size_type const)
+    constexpr void resize(size_type const count) noexcept
+    {
+        assert(count <= capacity_);
+
+        for (size_t i = this->sz; i <= count; ++i)
+            this->data_[i] = '\0';
+
+        this->sz = count;
+        this->data_[this->sz] = '\0';
```
h-2: you can make the data members visible so you don't need `this->` everywhere....

include/seqan3/range/container/small_string.hpp
h-2: What about `push_back()`?

include/seqan3/range/container/small_string.hpp
```diff
@@ -0,0 +1,265 @@
...
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     *
+     * ### Complexity
+     *
+     * Linear in the size of the strings.
+     */
+    template <size_t capacity2>
+    constexpr small_string<capacity_ + capacity2> operator+(small_string<capacity2> const & rhs) const noexcept
```
h-2: symmetric operators should be friends!

include/seqan3/range/container/small_string.hpp
```diff
@@ -0,0 +1,265 @@
...
+    //!\}
+};
+
+/*!\name Deduction guides
+ * \{
+ */
+//!\brief Deduces small_string from string literals.
+//!\relates small_string
+template <size_t N>
+small_string(const char (&)[N]) -> small_string<N - 1>;
```
h-2:
```suggestion
small_string(char const (&)[N]) -> small_string<N - 1>;
```

You really have difficulties with east const recently :thinking:

include/seqan3/range/container/small_vector.hpp
```diff
@@ -60,6 +60,7 @@ class small_vector
     using const_iterator  = value_type const *;                   //!< The const_iterator type.
     using difference_type = ptrdiff_t;                            //!< The difference_type type.
     using size_type       = detail::min_viable_uint_t<capacity_>; //!< The size_type type.
+
```
h-2: ?

include/seqan3/range/container/small_vector.hpp
h-2: Maybe that's something we should discuss? :man_shrugging:

include/seqan3/range/container/small_vector.hpp
```diff
@@ -863,48 +864,77 @@ class small_vector
     //!\{
 
     //!\brief Performs element-wise comparison.
-    friend constexpr bool operator==(small_vector const & lhs, small_vector const & rhs) noexcept
+    template <size_t cap2>
+    //!\cond
+        requires cap2 <= capacity_
```
h-2: maybe document in one line once why you have this constraint

test/snippet/range/container/small_string.cpp
```diff
@@ -0,0 +1,18 @@
...
+    constexpr seqan3::small_string sm{"hello"};     // construct from string literal at compile time!
+
+    static_assert(sm[0] == 'h');                    // This way I can also test it at compile-time
+
+    seqan3::debug_stream << sm.size() << std::endl; // prints 5! (the null character is only stored internally)
+
+    // conversion to a normal string:
+    std::string sm_string{sm.string()};
+    // conversion to a cstring:
+    const char * sm_cstr{sm.c_str()};
```
h-2: This is not a conversion strictly speaking...

include/seqan3/range/container/small_string.hpp
```diff
@@ -0,0 +1,265 @@
...
+     *
+     * ### Exceptions
+     *
+     * Strong exception guarantee. No data is modified.
+     *
+     * ### Complexity
+     *
+     * Linear in the size of the string.
+     */
+    std::string string() const
```
h-2: Actually I would just call the member `.str()` and not `.string()`. The latter is really weird and the STL uses `.str()`...

test/snippet/range/container/small_string.cpp
h-2: Also **please take care of east const**

include/seqan3/range/container/small_string.hpp
h-2: Also add an explicit conversion operator!

include/seqan3/range/container/small_vector.hpp
smehringer: doxygen :roll_eyes: 
If there is no space here the size_type of the derived child has no documentation for size_type...

doc/howto/write_a_view/solution_view.cpp
smehringer: I honestly don't know why this test failed in my PR all of a sudden. but it did with `namespace view is ambigous`.

include/seqan3/core/detail/reflection.hpp
smehringer: my default constructed `small_string` has size `0` so this would not work then. (Renes default constructed constexpr_string had size = capacity.). The `get_display_name_size_v<type>` with whom I templetize `small_string` sets the capacity exaclty to the length of `name_ptr`.

include/seqan3/range/container/small_vector.hpp
smehringer: right now it's done using copydoc everywhere. I think the documentation looks nicer that way. but yes I can add it to the next strategy meeting. doesn't hurt

include/seqan3/range/container/small_string.hpp
smehringer: why should it? the `small_vector` does never access the element one past the end and does not have the `.c_str()` functionality. It does brace initialization of the underlying array on default construction... I am not sure if this default initializes the elements too.

include/seqan3/range/container/small_string.hpp
smehringer: that's implicit because it is a class. do we always do it explicitly?

include/seqan3/range/container/small_string.hpp
smehringer: that's how it is done in `small_vector` but here we need to ensure that all positions that are presently not filled are the `\0` character such that when I push back a value the string is automatically correctly null terminated

include/seqan3/range/container/small_string.hpp
smehringer: > what about the other one? 

I thought since you provide a value with which to fill the remaining entries, this is usually done when resizing to a bigger size (which would not be a problem because the array contains `\0` at every position after its `.size()`). But you are right, in case I resize to a smaller string this is not correct.

> What about insert() at the end? What about push_back()?

Since the array contains `\0` at every position after its `.size()`, it will be automatically zero terminated.
That's why clear() sets all the `\0` characters again.

include/seqan3/range/container/small_string.hpp
smehringer: that was copy and pasted from renes code (as a tiny defence).

test/snippet/range/container/small_string.cpp
smehringer: changed it to `// access data directly with a pointer to the underlying zero-terminated array:`

include/seqan3/range/container/small_string.hpp
smehringer: it is called `str()` for streams but `string()` for filesystem paths.. but I'll change it to str

smehringer: > Also add an explicit conversion operator!

If we make it implicit, we can avoid the `.string()` in 
`get_display_name_v<me_type>.string()` :grin:

doc/howto/write_a_view/solution_view.cpp
h-2: :thinking:

include/seqan3/core/detail/reflection.hpp
h-2: > my default constructed small_string has size 0 so this would not work then.

Why? If it's size 0, you don't need to copy...

> The get_display_name_size_v<type> with whom I templetize small_string sets the capacity exaclty to the length of the string pointed to by name_ptr

Yes, but if there is nothing to copy there is nothing to copy and default initialisation is ok.

include/seqan3/range/container/small_string.hpp
h-2: > On second thought, It does brace initialization of the underlying array on default construction... I am not sure if this default initializes the elements too.

It does, that's what I meant!

include/seqan3/range/container/small_string.hpp
h-2: yep

include/seqan3/range/container/small_string.hpp
h-2: Hm, but this function is linear, while correcting on push_back or insert would be constant...

Even if you do stick with this design, use `assign` instead, because the above costs 2n (the RHS is constructed, 0ed and then copied; there is no move!).

include/seqan3/range/container/small_string.hpp
h-2: > (which would not be a problem because the array contains \0 at every position after its .size())

This is the core question. Guaranteeing that makes implementation simpler, but is more expensive...

include/seqan3/range/container/small_vector.hpp
h-2: image

include/seqan3/range/container/small_string.hpp
```diff
@@ -0,0 +1,301 @@
...
+        small_string<capacity_ + capacity2> tmp{lhs};
+        tmp.insert(tmp.end(), rhs.begin(), rhs.end());
+        return tmp;
+    }
+    //!\}
+
+    /*!\name Explicit conversion
+     * \{
+     */
+    explicit operator std::string() const
```
h-2: implicit as discussed

include/seqan3/core/detail/reflection.hpp
smehringer: I refactored the code further. It should be clear now what happens

include/seqan3/range/container/small_string.hpp
smehringer: done

include/seqan3/range/container/small_string.hpp
smehringer: I removed the constructor

include/seqan3/range/container/small_string.hpp
smehringer: I refactored the functions to be constant

include/seqan3/range/container/small_string.hpp
```diff
@@ -0,0 +1,328 @@
...
+    {
+        return capacity_;
+    }
+
+    //!\brief Returns the maximal capacity.
+    constexpr size_type capacity() const noexcept
+    {
+        return capacity_;
+    }
+    //!\}
```
h-2: you can make the previous two static

include/seqan3/range/container/small_string.hpp
```diff
@@ -0,0 +1,328 @@
...
+        resize(count, '\0');
+    }
+
+    //!\copydoc seqan3::small_vector::resize(size_type const, value_type const value)
+    constexpr void resize(size_type const count, char const value) noexcept
+    {
+        assert(count <= capacity_);
+
+        for (size_t i = count; i <= sz; ++i) // sz >= count; discard values in [count, sz]
+            data_[i] = '\0';
```
h-2: why these two lines?

include/seqan3/range/container/small_string.hpp
h-2: not 100% convinced yet :D

include/seqan3/range/container/small_string.hpp
```diff
@@ -0,0 +1,328 @@
...
+    //!\}
+
+    /*!\name Implicit conversion to std::string
+     * \{
+     */
+    operator std::string() const
+    {
+        return str();
+    }
+    //!\}
```
h-2: You documented this as a group, but not the member itself whihc will give a warning with newer doxygen. Also I would group it with `.str()` and `.c_str()`.

PR #937, opened 2019-04-28
[FIX] edit_distance for small patterns behaves the same as large patt…

marehr:
…ern when using max_errors

### DIFF
--- a/include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
+++ b/include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
@@ -492,12 +492,15 @@ bool pairwise_alignment_edit_distance_unbanded<database_t, query_t, align_config
         advance_score(hp, hn, score_mask);
 
         if constexpr(use_max_errors)
-            if (_score <= max_errors && on_hit())
+        {
+            // updating the last active cell
+            if (update_last_active_cell())
             {
                 add_state();
                 ++database_it;
                 return true;
             }
+        }
 
         add_state();
         ++database_it;


include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
```diff
@@ -492,12 +492,15 @@ bool pairwise_alignment_edit_distance_unbanded<database_t, query_t, align_config
         advance_score(hp, hn, score_mask);
 
         if constexpr(use_max_errors)
-            if (_score <= max_errors && on_hit())
+        {
+            // updating the last active cell
+            if(update_last_active_cell())
```
eseiler: I looked at the code for `update_last_active_cell()` and I'm a bit confused. Isn't this function a contradiction? It either calls `void` functions, or returns `on_hit()` which always returns `false` or it returns `false` at the very end of `update_last_active_cell()`.

But given that `on_hit()` is a contradiction itself, the `if` that was replaced (with a contradiction) was a contradiction anyways - so effectively nothing changed?

eseiler:
```suggestion
            if (update_last_active_cell())
```

marehr: > I looked at the code for `update_last_active_cell()` and I'm a bit confused. Isn't this function a contradiction? It either calls `void` functions, or returns `on_hit()` which always returns `false` or it returns `false` at the very end of `update_last_active_cell()`.
> 
> But given that `on_hit()` is a contradiction itself, the `if` that was replaced (with a contradiction) was a contradiction anyways - so effectively nothing changed?
> 
> Edit: Code coverage also says that the inside of the if is never reached.

You are right. This will be changed in a later PR (already in pipeline).

marehr:
@eseiler added white space after `if`

include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
smehringer: But how does this PR fix anything then?

marehr: it changes the behaviour of the code and output.

the point of this PR is not the `if` but what happens within the condition of the `if`.

PR #939, opened 2019-04-28
[FIX] fixes a bug when going before first row in edit_distance

marehr:


include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
```diff
@@ -319,8 +319,21 @@ class pairwise_alignment_edit_distance_unbanded
...
             return true;
 
-        if (is_global && last_block == 0)
-            return false;
+        if constexpr (is_global)
+        {
+            if (last_block == 0)  // [[unlikely]]
+            {
+                // Note: This is a special case where we hit before the first query character
+                // (i.e. score_mask == 0 && last_block == 0), which means that there is NO score in the entire column that is
```
eseiler: Line length

marehr:
@eseiler done

h-2:
You have introduced new branches, but not new tests which is why the coverage test is correctly saying that this PR reduces coverage. Can you add tests?

marehr:
@h-2 This will follow in a (bigger) PR; I cherry-picked this commit, because it is kinda standalone to make it smaller. And that PR introduces all kinds of changes to make this testable.

On a side-note we do not introduce a new code-branch here. We just make it some instructions longer compared to before, because the `if` was also not hit/tested before.

h-2:
Hm, ok.

Issue #869, opened 2019-04-02
Final refactoring of alphabets


h-2:
These are the discussed changes and some smaller additions by me not yet discussed.
@seqan/core please have a look and give me objections ASAP since I want to start working on this very soon.

`seqan3::SemialphabetConcept`:
  * refines `std::StrictTotallyOrdered` (order required to respond to ranks)
  * refines `std::Copyable` and requires `std::is_trivially_copyable`
  * requirement for `std::Regular` is dropped, but recommended for all type definitions
  - [x] `to_rank` customisation point object
    * looks for member function
    * looks for free function via ADL
    * not  picked up via ADL itself
  * [x] `alphabet_size_v`
    * looks for static member variable `::alphabet_size` (now it's `::value_size` – a weird name leftover from SeqAn2)
    * looks for constexpr free function via ADL `alphabet_size()` which makes it possible to define your own alphabet without overriding the metafunction in SeqAn

`seqan3::WritableSemialphabetConcept`:
  * refines `seqan3::SemialphabetConcept`
  * [x] `assign_rank_to(rank, alph)` customisation point object
    * was `assign_rank(alph, rank)` in seqan2-(target, source)-style; new style is more intuitive and allows for `assign_rank_to<dna5>(rank)` instead of explicit construction `assign_rank(dna5{}, rank)` [also necessary because we don't require regularity anymore]
    * looks for `assign_rank(rank)` member function
    * looks for `assign_rank_to(rank, alph)` free function via ADL
    * not  picked up via ADL itself
    * the CPO defines a generic overload for `assign_rank_to(rank, alph &&)` so this does not have to be customised (only the overload for lvalue-references has to be customised).

`seqan3::AlphabetConcept`
  * refines `seqan3::SemialphabetConcept`
  * [x] `to_char` customisation point analogous to above
 
`seqan3::WritableAlphabetConcept`
  * refines `seqan3::AlphabetConcept` and `seqan3::WritableSemialphabetConcept`
  * [x]  `assign_char_to(char, alph)` customisation point object analogous to `assign_rank_to(rank, alph)` 

PLUS `Constexpr` versions of all of the above in `seqan3::detail::`.

  - [x] Note that `underlying_rank_t` and `underlying_char_t` will be renamed to `alphabet_rank_t` and `alphabet_char_t` and will not be customisation points, they always resolve to the return type of `to_char()` and `to_rank().

  - [x] rename `alphabet_size_v` -> `alphabet_size`
[...]

PR #943, opened 2019-04-30
second phase of alphabet restructuring

h-2:
I have switched to a local fork to workaround USCiLab/cereal#565 for now.

Changes the alphabet related functions to be function objects. See also #869 

I recommend reviewing on a commit-by-commit basis.


marehr:
:construction: **WIP** :construction:  :construction_worker_man: :construction_worker_woman: :building_construction: :laughing:

h-2:
@eseiler Can you review this? Thanks!

include/seqan3/alphabet/concept.hpp
```diff
@@ -16,22 +16,936 @@
...
+ * \ingroup alphabet
+ *
+ * \details
+ *
+ * This is a function object. Invoke it with the parameters specified above.
+ *
+ * ### Example
+ *
+ * ```cpp
+ * char d = seqan3::to_char('A'_dna5);                  // d == 'A'
```
eseiler: snippet?

include/seqan3/alphabet/concept.hpp
```diff
@@ -16,22 +16,936 @@
...
+ * \ingroup alphabet
+ *
+ * \details
+ *
+ * This is a function object. Invoke it with the parameters specified above.
+ *
+ * ### Example
+ *
+ * ```cpp
+ * seqan3::dna5 d = seqan3::assign_char_to('A', seqan3::dna5{});   // d == 'A'_dna5
```
eseiler: snippet

include/seqan3/alphabet/concept.hpp
```diff
@@ -16,22 +16,936 @@
...
+ *
+ * This is a function object. Invoke it with the parameters specified above.
+ *
+ * Note that this is not a customisation point, it simply invokes seqan3::char_is_valid_for and seqan3::assign_char_to.
+ *
+ * ### Example
+ *
+ * ```cpp
+ * seqan3::dna5 d = seqan3::assign_char_strictly_to('A', seqan3::dna5{});   // d == 'A'_dna5
+ * assign_char_strictly_to('C', d);                                         // d == 'C'_dna5
```
eseiler: snippet

include/seqan3/alphabet/concept.hpp
```diff
@@ -16,22 +16,936 @@
...
+
+namespace seqan3
+{
+
+/*!\name Function objects
+ * \{
+ */
+
+/*!\brief Return the rank representation of an (Semi-)Alphabet object.
+ * \param alph The target object; its type must model seqan3::Alphabet.
```
smehringer: must model (Semi-)Alphabet
Is this enforced somewhere? It is rather must have either a to_rank member, etc...?

include/seqan3/alphabet/concept.hpp
```diff
@@ -16,22 +16,936 @@
...
+ * \returns The rank representation; an integral type.
+ * \ingroup alphabet
+ * \details
+ *
+ * This is a function object. Invoke it with the parameters specified above.
+ *
+ * ### Example
+ *
+ * ```cpp
+ * char d = seqan3::to_char('A'_dna5);                  // d == 'A'
```
smehringer: should be a to_rank example?

include/seqan3/alphabet/concept.hpp
```diff
@@ -16,22 +16,936 @@
...
+
+/*!\brief A type trait that holds the size of any given seqan3::Semialphabet.
+ * \ingroup alphabet
+ *
+ * \details
+ *
+ * ### Example
+ *
+ * ```cpp
+ * std::cout << seqan3::alphabet_size_v<seqan3::dna4>; // prints '4'
```
eseiler: snippet

include/seqan3/alphabet/concept.hpp
```diff
@@ -16,22 +16,936 @@
...
+// ============================================================================
+
+namespace seqan3::detail::adl::only
+{
+
+//!\brief Call `assign_rank_to()` on the argument, but only consider ADL.
+//!\ingroup alphabet
+template <typename alph_t>
+//!\cond
+    requires requires { assign_rank_to(std::declval<alphabet_rank_t<alph_t>>(), std::declval<alph_t &>()); }
```
smehringer: std::declval<alphabet_rank_t<alph_t> `const`>()

test/unit/alphabet/custom_alphabet_test.cpp
```diff
@@ -0,0 +1,69 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+#include <seqan3/core/platform.hpp>
+#include <cereal/types/common.hpp>
```
eseiler: order

include/seqan3/alphabet/concept.hpp
```diff
@@ -16,22 +16,936 @@
...
+
+namespace seqan3
+{
+
+/*!\name Function objects
+ * \{
+ */
+
+/*!\brief Assign a rank to an alphabet object.
+ * \param chr  The rank being assigned; must be of the seqan3::alphabet_rank_t of the target object.
```
smehringer: \param `rank`?

include/seqan3/alphabet/concept.hpp
```diff
@@ -16,22 +16,936 @@
...
+        }
+    }
+
+    //!\brief Operator definition for rvalues.
+    template <typename alph_t>
+    //!\cond
+        requires (has_assign_rank_to_adl_overload<alph_t>        ||
+                  has_assign_rank_to_adaptation_overload<alph_t> ||
+                  has_assign_rank_member<alph_t>) &&
+                  (!std::is_lvalue_reference_v<alph_t>)
```
smehringer: -> `std::is_rvalue_reference<alph_t>` then you don't the negation nor the paranthesis?

include/seqan3/alphabet/concept.hpp
```diff
@@ -16,22 +16,936 @@
...
+SEQAN3_CONCEPT has_char_is_valid_member = requires (alphabet_char_t<alph_t> chr)
+{
+    { alph_t::char_is_valid(chr) } -> bool;
+};
+
+template <typename alph_t>
+SEQAN3_CONCEPT has_minimal_alph_interface = requires (alphabet_char_t<alph_t> chr)
+{
+    { to_char(assign_char_to(chr, alph_t{})) }
+        -> alphabet_char_t<alph_t>;
```
smehringer: put into line above ?

include/seqan3/alphabet/concept.hpp
```diff
@@ -1,9 +1,9 @@
-// -----------------------------------------------------------------------------------------------------
+// ============================================================================-----------------------------------
```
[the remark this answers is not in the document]
h-2: that was a search and replace gone wrong...

include/seqan3/alphabet/concept.hpp
```diff
@@ -16,22 +16,936 @@
...
 #include <seqan3/core/concept/core_language.hpp>
+#include <seqan3/core/detail/reflection.hpp>
+#include <seqan3/core/metafunction/basic.hpp>
 #include <seqan3/core/metafunction/function.hpp>
-#include <seqan3/alphabet/adaptation/pre.hpp>
+#include <seqan3/alphabet/adaptation/char.hpp>
+#include <seqan3/alphabet/adaptation/uint.hpp>
 #include <seqan3/alphabet/concept_pre.hpp>
 #include <seqan3/alphabet/detail/member_exposure.hpp>
+#include <seqan3/alphabet/exception.hpp>
```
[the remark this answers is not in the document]
h-2: `c < d < e` or not?

include/seqan3/alphabet/concept.hpp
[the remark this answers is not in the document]
h-2: I hate snippets :expressionless:

include/seqan3/io/stream/parse_condition.hpp
```diff
@@ -346,7 +345,7 @@ struct parse_asserter
      *
      * Thread-safe.
      */
-    template <char_adaptation_concept char_type>
+    template <std::Integral char_type>
```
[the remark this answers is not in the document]
h-2: Counter-question: what happens when you provide `char32_t` and it is then cast to `char` upon interval checking?

include/seqan3/alphabet/concept.hpp
h-2: > It is rather must have either a to_rank member, etc...?

Yes. This is a little tricky, because when designing an alphabet, you need to provide an implementation → it is not the functor that requires the concept, but the other way around. However for 99% of the users it is "a functor that works on all types that model the concept".

I can change the documentation to be strictly correct, but it will be slightly less useful to our users.

include/seqan3/alphabet/concept.hpp
h-2: That won't work. If you pass a temporary of `T`, the type of `alph_t` will be `T`, not `T &&`. Only lvalue references are "part of the type" because they need to be for reference collapsing.

h-2:
I have added long(er) snippets to everything and addressed most of your points I think. I have also restructured the documentation for the customisation points a little so that it is hopefully more helpful. I have also added a second custom alphabet test that tests the "not-default-constructible" overloads (the nice adl-hack through associated namespaces of template args).

Also requested @rrahn for review.

If cereal does not react until Monday I will fork the repo and we need to discuss what to do before the release.

include/seqan3/alphabet/concept.hpp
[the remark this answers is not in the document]
eseiler: Yes, but shouldn't the `alphabet` includes precede the `core` includes?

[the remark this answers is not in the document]
h-2: Haha, yes. I only saw the three lines above in the comment :see_no_evil:

include/seqan3/alphabet/concept.hpp
smehringer: Hm.. The "strictly correct" documentation might be too verbose. But "must model alphabet" is not correct as this function is not requiring it but, as you said, helps types to model Alphabet. So this is confusing with how we document most other stuff. :(

include/seqan3/alphabet/concept.hpp
h-2: But, I have changed that already, or not?

include/seqan3/alphabet/concept.hpp
```diff
@@ -1,9 +1,9 @@
-// -----------------------------------------------------------------------------------------------------
+// ---------------------------------------------------------------------------------------------------------------
```
smehringer: still wrong

test/performance/alphabet/alphabet_benchmark.cpp
```diff
@@ -89,52 +89,52 @@ static void to_rank(benchmark::State& state)
...
+BENCHMARK_TEMPLATE(to_rank_, dna4);
+BENCHMARK_TEMPLATE(to_rank_, dna5);
+BENCHMARK_TEMPLATE(to_rank_, dna15);
+BENCHMARK_TEMPLATE(to_rank_, rna15);
+BENCHMARK_TEMPLATE(to_rank_, rna4);
+BENCHMARK_TEMPLATE(to_rank_, rna5);
+BENCHMARK_TEMPLATE(to_rank_, char);
+BENCHMARK_TEMPLATE(to_rank_, gapped<dna4>);
+BENCHMARK_TEMPLATE(to_rank_, alphabet_variant<gap,dna4,dna5,dna15,rna15,rna4,rna5>);
+BENCHMARK_TEMPLATE(to_rank_, alphabet_variant<char, dna4, dna5, dna15>);
```
smehringer: So how are the benchmarks? same/faster/slower?

include/seqan3/alphabet/concept.hpp
h-2: But I changed it, or not?

test/performance/alphabet/alphabet_benchmark.cpp
h-2: I checked, it's the same. I mean, everything else would be strange, it just calls the same functions in the end.

h-2:
I have DRYed a lot of the CPO-definitions so the number of added lines is reduces by about 400; there are no more "auxiliary concepts" and also no more free functions inside detail::adl::only. This required adding a small macro, though. Return types are now properly enforced and I fixed some smaller bugs, too.

Note that the CPOs are not in anonymous namespaces, because doxygen handles this **really** badly. It's not necessary though, unless we want to add friends to our own classes within namespace seqan3 that have the same names as the CPOs [we can just use members for this and friends of external types are handled correctly].

please give this a review soon everyone, thanks!

h-2:
The design in this file is based loosely on:
  * https://quuxplusone.github.io/blog/2018/03/19/customization-points-for-functions/
  * http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4381.html

(the anonymous namespace and static hackery in the proposal are not necessary, because we have inline variables now)

include/seqan3/alphabet/adaptation/char.hpp
```diff
@@ -27,117 +27,53 @@
...
- */
 /*!\brief Converting char to char is no-op (it will just return the value you pass in).
  * \tparam char_type One of `char`, `char16_t`, `char32_t` or `wchar_t`.
  * \param chr The alphabet letter that you wish to convert to char.
  * \returns The letter's value in the alphabet's rank type (usually `char`).
  */
 template <typename char_type>
-constexpr alphabet_char_t<char_type> to_char(char_type const chr) noexcept
-    requires detail::is_char_adaptation_v<char_type>
+constexpr char_type to_char(char_type const chr) noexcept
```
eseiler: This confused me for a bit.
The `alphabet_char_t` is gone because this is now only for `char` types and as such the `to_char` is a no-op, right?

Then it should also be documented as no-op?
(the 
```
 * \param chr The alphabet letter that you wish to convert to char.
 * \returns The letter's value in the alphabet's rank type (usually `char`).
```
Is not exactly correct then?
(If accepted, this also applies to the `uint` adaptations.)

include/seqan3/alphabet/adaptation/uint.hpp
```diff
@@ -27,114 +27,58 @@
...
+ * \param intgr The parameter's actual value is ignored.
+ * \returns The respective size (e.g. 256 for `uint8_t`).
  */
 template <typename uint_type>
-//!\cond
-    requires detail::is_uint_adaptation_v<uint_type>
-//!\endcond
-struct underlying_rank<uint_type>
+constexpr auto alphabet_size(uint_type const & SEQAN3_DOXYGEN_ONLY(intgr)) noexcept
+    requires ::seqan3::detail::is_uint_adaptation_v<uint_type>
```
eseiler: Doxygen?
Also the other `requires`

include/seqan3/alphabet/concept.hpp
```diff
@@ -182,7 +823,8 @@ template <CerealInputArchive archive_t, typename wrapped_alphabet_t>
 void CEREAL_LOAD_MINIMAL_FUNCTION_NAME(archive_t const &,
                                        wrapped_alphabet_t && l,
                                        alphabet_rank_t<detail::strip_cereal_wrapper_t<wrapped_alphabet_t>> const & r)
-    requires Semialphabet<detail::strip_cereal_wrapper_t<wrapped_alphabet_t>>
+    requires !std::is_const_v<std::remove_reference_t<wrapped_alphabet_t>> &&
```
eseiler: Doxygen

include/seqan3/alphabet/adaptation/char.hpp
```diff
@@ -27,117 +27,53 @@
...
+ * \param chr The parameters actual value is ignored.
+ * \returns The respective sizes (e.g. 256 for `char`).
  */
 template <typename char_type>
-//!\cond
-    requires detail::is_char_adaptation_v<char_type>
-//!\endcond
-struct underlying_rank<char_type>
+constexpr auto alphabet_size(char_type const & SEQAN3_DOXYGEN_ONLY(chr)) noexcept
+    requires ::seqan3::detail::is_char_adaptation_v<char_type>
```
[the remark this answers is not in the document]
rrahn: Sometimes doxygen displays this correctly if you put it after the function declarator.  
But it seems to be confusing what rules should be applied. Like, when to surround it with cond or not. I would generally say, we need to wait for concepts support in doxygen (really, don't know, if this will ever happen) and should rather put for all versions the stuff into a cond block

include/seqan3/alphabet/concept.hpp
```diff
@@ -14,24 +14,668 @@
...
+namespace seqan3::detail::adl::only
+{
+
+//!\brief Functor definition for seqan3::to_rank.
+struct to_rank_fn
+{
+private:
+    SEQAN3_CPO_IMPL(2, to_rank(v)                       )    // ADL
+    SEQAN3_CPO_IMPL(1, seqan3::adaptation::to_rank(v)   )    // customisation namespace
+    SEQAN3_CPO_IMPL(0, v.to_rank()                      )    // member
```
rrahn: Ok, I feel I am missing something here. So just to get it right. In order to avoid ambigious function call you add the priority_tag, as a parameter to the functions. Therefor you get unambigious overloads. The impl functions are unconstrained (neither per concept nor per SFINAE), but down there you call the impl function with `priority_tag<2>`. In this case you want to achieve, that it first tries the impl function with priority 2, if that fails then apply implicit conversion to priority_tag<1> etc. 
But if my type only supports .to_rank() and has no global to_rank in its namespace, would the compiler not emit an error in the first place?

include/seqan3/alphabet/concept.hpp
```diff
@@ -14,24 +14,668 @@
...
+    //!\brief Operator definition.
+    template <typename alph_t>
+    constexpr auto operator()(alph_t const a) const noexcept
+    //!\cond
+        -> decltype(impl(priority_tag<2>{}, a))
+    //!\endcond
+    {
+        static_assert(noexcept(impl(priority_tag<2>{}, a)),
+            "Only overloads that are marked noexcept are picked up by seqan3::to_rank().");
+        static_assert(std::Constructible<size_t, decltype(impl(priority_tag<2>{}, a))>,
```
rrahn: ConvertibleTo?

include/seqan3/alphabet/concept.hpp
```diff
@@ -14,24 +14,668 @@
...
+    template <typename alph_t>
+    //!\cond
+        requires (!std::is_lvalue_reference_v<alph_t>)
+    //!\endcond
+    constexpr auto operator()(seqan3::alphabet_rank_t<alph_t> const r, alph_t && a) const noexcept
+    //!\cond
+        -> std::remove_reference_t<decltype(impl(priority_tag<2>{}, a, r))>
+    //!\endcond
+    {
+        return (*this)(r, a); // call above function but return by value
```
rrahn: I would prefer the syntax: this->operator()(r, a); as it is easier to read. But I have no strong feelings.

include/seqan3/alphabet/concept.hpp
rrahn: But it is not resolved.

include/seqan3/core/detail/customisation_point.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+// ============================================================================
+// SEQAN3_CPO_IMPL
+// ============================================================================
+
+//!\brief A macro that helps defining the overload set of a customisation point.
+#define SEQAN3_CPO_IMPL(PRIO, TERM)                                                                                  \
+/*!\brief A customisation point overload.*/                                                                          \
+template <typename t, typename ...arg_ts>                                                                            \
+static constexpr auto impl(seqan3::detail::priority_tag<PRIO> const &, t && v, [[maybe_unused]] arg_ts && ... args)  \
```
rrahn: should you not require TERM in the function declarator?

include/seqan3/alphabet/concept.hpp
```diff
@@ -14,24 +14,668 @@
...
+ */
+inline constexpr auto assign_char_strictly_to = detail::adl::only::assign_char_strictly_to_fn{};
+//!\}
+} // namespace seqan3
+
+// ============================================================================
+// alphabet_size_v
+// ============================================================================
+
+namespace seqan3::detail::adl::only
```
rrahn: why the additional ::only subnamespace?

include/seqan3/alphabet/concept.hpp
```diff
@@ -14,24 +14,668 @@
...
+{
+/*!\brief Functor definition used indirectly by for seqan3::detail::alphabet_size_v.
+ * \tparam alph_t   The type being queried.
+ * \tparam s_alph_t `alph_t` with cvref removed and possibly wrapped in std::type_identity; never user-provide this!
+ * \ingroup alphabet
+ */
+template <typename alph_t,
+          typename s_alph_t = std::conditional_t<std::is_nothrow_default_constructible_v<remove_cvref_t<alph_t>> &&
```
rrahn: what happens if it is wrapped in type_identity. What is the use case for this? Only the value_size static member can then be used right? But is it then not a valid alphabet if it only supports alphabet_size() even if it is no nothrow constructible and not constexpr constructible?

include/seqan3/alphabet/concept.hpp
```diff
@@ -14,24 +14,668 @@
...
+                                                 seqan3::is_constexpr_default_constructible_v<remove_cvref_t<alph_t>>,
+                                                 remove_cvref_t<alph_t>,
+                                                 std::type_identity<alph_t>>>
+struct alphabet_size_fn
+{
+private:
+    SEQAN3_CPO_IMPL(2, (alphabet_size(v)                           ))    // ADL
+    SEQAN3_CPO_IMPL(1, (seqan3::adaptation::alphabet_size(v)       ))    // customisation namespace
+    SEQAN3_CPO_IMPL(0, (deferred_type_t<remove_cvref_t<alph_t>, decltype(v)>::value_size )) // member
```
rrahn: So the deferred type get's an additional parameter in order to avoid unused variable warning?

include/seqan3/alphabet/concept.hpp
```diff
@@ -14,24 +14,668 @@
...
+ * ### Customisation point
+ *
+ * This is a customisation point (TODO link to manual). To specify the behaviour for your own alphabet type,
+ * simply provide one of the three functions specified above.
+ */
+template <typename alph_t>
+//!\cond
+    requires requires { { detail::adl::only::alphabet_size_fn<alph_t>{}() }; }
+//!\endcond
+inline constexpr auto alphabet_size_v = detail::adl::only::alphabet_size_fn<alph_t>{}();
```
rrahn: This does not cause trouble when calling the templatized operator() as non-template?

include/seqan3/alphabet/concept.hpp
```diff
@@ -14,24 +14,668 @@
...
+struct alphabet_size_fn
+{
+private:
+    SEQAN3_CPO_IMPL(2, (alphabet_size(v)                           ))    // ADL
+    SEQAN3_CPO_IMPL(1, (seqan3::adaptation::alphabet_size(v)       ))    // customisation namespace
+    SEQAN3_CPO_IMPL(0, (deferred_type_t<remove_cvref_t<alph_t>, decltype(v)>::value_size )) // member
+
+public:
+    //!\brief Operator definition.
+    template <typename dummy = int> // need to make this a template to enforce deferred instantiation
```
rrahn: So without this the impl functions would be directly instantiated but it could happen that the constexpr functions are not known yet. So making it a template will force two-phase template instantiation. Or is there any other magic involved?

include/seqan3/alphabet/concept.hpp
rrahn: If that's so, why not just marking the v variable as [[maybe_unused]]?

include/seqan3/io/stream/parse_condition_detail.hpp
```diff
@@ -101,7 +102,7 @@ class parse_condition_base;
  *
  * \details
  *
- * The must be invocable with an seqan3::char_adaptation_concept type and supply a static constexpr `msg` member of type
+ * The must be invocable with an std::Integral type and supply a static constexpr `msg` member of type
```
rrahn: The what?

test/snippet/alphabet/assign_rank_to.cpp
```diff
@@ -0,0 +1,18 @@
...
+    seqan3::assign_rank_to(66, c);     // calls seqan3::adaptation::assign_rank_to(66, c); == 'B'
+
+    seqan3::dna5 d{};
+    seqan3::assign_rank_to(2, d);     // calls .assign_rank(2) member; == 'G'_dna5
+
+    // also works for temporaries:
+    seqan3::dna5 d2 = seqan3::assign_rank_to(2, seqan3::dna5{});
+
+    // too-large ranks are undefined behaviour:
+    seqan3::dna5 d3 = seqan3::assign_rank_to(50, seqan3::dna5{}); // 💣💥
```
rrahn: Please do not use 💣 💥 in comments.

test/unit/alphabet/custom_alphabet2_test.cpp
```diff
@@ -0,0 +1,91 @@
...
+constexpr bool char_is_valid_for(char const c, std::type_identity<my_alph> const &) noexcept
+{
+    switch (c)
+    {
+        case '0': case 'F': case 'f': case '1': case 'T': case 't': return true;
+        default: return false;
+    }
+}
+
+}
```
rrahn: //  my_namespace

test/unit/alphabet/custom_alphabet_test.cpp
```diff
@@ -0,0 +1,75 @@
...
+{
+    switch (c)
+    {
+        case '0': a = my_alph::ZERO; return a;
+        case '1': a = my_alph::ONE;  return a;
+        default:  a = my_alph::TWO;  return a;
+    }
+}
+
+}
```
rrahn: // my_namespace

include/seqan3/alphabet/all.hpp
```diff
@@ -51,7 +51,7 @@
  * ### Alphabet size
  *
  * All alphabets in SeqAn have a fixed size. It
- * can be queried via the seqan3::alphabet_size metafunction and *optionally* also the `value_size` static
+ * can be queried via the seqan3::alphabet_size metafunction and *optionally* also the `alphabet_size` static
```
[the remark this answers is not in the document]
rrahn: ok

include/seqan3/alphabet/adaptation/char.hpp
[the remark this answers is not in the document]
h-2: AFAICT parens and braces in the term break doxygen's handling of the trailing requires statement, but I can still add the \cond.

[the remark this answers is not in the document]
h-2: I'll move them to the top then as well

include/seqan3/alphabet/concept.hpp
h-2: that change was an unnecessary leftover anyway, I removed it.

include/seqan3/alphabet/concept.hpp
h-2: There is now an explicit requires clause.

h-2:
> But it is not resolved.

I can't reply to that inline :|  I don't understand what you mean. Because I say "alphabet object"? I have already removed the requirement for the concept... What else would you call it? This is as "correct" as is understandable for the general use-case I think.

The only 100% formally correct description would be "Assign an object of type alphabet_rank_t<your_type> to your type where your_type is a type that provides one of three overloads described below".
That would make very poor UX I think. 99.99% of the time people will look at this documentation item they want to call `assign_rankt_to` on an existing alphabet.

include/seqan3/alphabet/concept.hpp
h-2: Hm, I wanted to do `operator()(r, a)` but that does self-invocation.  The above solution was the shortest and it is even commented?

h-2:
I have updated the PR. There is some weird failure in an entirely unrelated alignment snippet, I really don't know what's going on. Apparantly it's trying to invoke to_char() with an `operator&` defined on ios 
:man_shrugging: 

any ideas someone?

include/seqan3/alphabet/concept.hpp
rrahn: I would have guessed, since you have a template operator that you need to call it with template in a template context.

```cpp
detail::adl::only::alphabet_size_fn<alph_t>{}.template operator()<>();
```
I mean it seems to be valid like it is. Just not sure if all compiler would allow this, but we will figure it out at some point I guess 😄

include/seqan3/alphabet/concept.hpp
rrahn: The member function template is generated anyway. But the compiler would ignore it if it knows, that it is never used? So also if you would pass it into a lambda, that does nothing with the variable you cannot get deffered instantiation?

include/seqan3/alphabet/concept.hpp
rrahn: > ADL looks also in the associated namespaces of the template parameter(!) so the user can provide overloads almost exactly like before.

😲

include/seqan3/alphabet/concept.hpp
h-2: It works when the types are deducted from arguments so I assumed defaulted template parameters would work as well...

include/seqan3/core/detail/customisation_point.hpp
h-2: But it leads to some more difficult to understand errors and is not coherent with our other style. I will change it to requires clauses.

h-2:
Another question that we can also discuss later is whether we want a general customisation namespace, e.g. `seqan3::customisation` which we use for customisation point implementations for types outside of seqan (built-in, std:: maybe) and as "upload" namespace ¹ in which people can provide specialisations for types outside of their namespaces.

This would mean renaming seqan3::adaptation to seqan3::customisation or seqan3::custom. I am in favour of this, but we can easily perform this change later, it will just change a few lines here.

¹ https://quuxplusone.github.io/blog/2018/03/19/customization-points-for-functions/

include/seqan3/alphabet/concept.hpp
rrahn: It still looks weird, but as I sad, no strong feelings.

include/seqan3/core/metafunction/basic.hpp
```diff
@@ -117,8 +117,60 @@ struct remove_rvalue_reference
...
+// deferred_type
+// ----------------------------------------------------------------------------
+
+/*!\brief Return the type identity; further arguments are ignored, but can make this type dependent if they are
+ *        [TransformationTrait].
+ * \tparam t The type to operate on.
+ * \tparam dependent_ts Any provided types are ignored.
+ */
+template <typename t, typename ...dependent_ts>
+struct deferred_type
```
rrahn: I don't think that this should be public API. The use case is very special and would require at least a snippet to get the idea. Instead I would put it into detail and maybe even in the adl stuff/customisation point stuff.

test/snippet/alphabet/assign_rank_to.cpp
```diff
@@ -0,0 +1,18 @@
...
+    seqan3::assign_rank_to(66, c);     // calls seqan3::adaptation::assign_rank_to(66, c); == 'B'
+
+    seqan3::dna5 d{};
+    seqan3::assign_rank_to(2, d);     // calls .assign_rank(2) member; == 'G'_dna5
+
+    // also works for temporaries:
+    seqan3::dna5 d2 = seqan3::assign_rank_to(2, seqan3::dna5{});
+
+    // too-large ranks are undefined behaviour:
+//    seqan3::dna5 d3 = seqan3::assign_rank_to(50, seqan3::dna5{}); // 💣💥
```
rrahn: Can you please remove 💣 💥 from the comment?

h-2:
> There is some weird failure in an entirely unrelated alignment snippet, I really don't know what's going on.

@rrahn did you have a look at this? Do you have a clue whats going on?

include/seqan3/io/stream/parse_condition_detail.hpp
h-2: not resolved?

include/seqan3/core/detail/customisation_point.hpp
h-2: not resolved?

test/snippet/alphabet/assign_rank_to.cpp
h-2: hm, ok :crying_cat_face:

test/unit/alphabet/custom_alphabet2_test.cpp
h-2: not resolved?

test/unit/alphabet/custom_alphabet_test.cpp
h-2: not resolved?

include/seqan3/alphabet/concept.hpp
```diff
@@ -14,24 +14,667 @@
...
+ * ### Example
+ *
+ * \include test/snippet/alphabet/assign_rank_to.cpp
+ *
+ * For an example of a full alphabet definition with free function implementations (solution 1. above),
+ * see seqan3::Alphabet.
+ *
+ * ### Customisation point
+ *
+ * This is a customisation point (TODO link to manual). To specify the behaviour for your own alphabet type,
```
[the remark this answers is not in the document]
h-2: well I am just writing up this stuff for my PhD so I will have some text to copy'n'pate for a respective documentation page soon. Depends also on discussion point above (I'll add to the next strategy meeting).

include/seqan3/alphabet/detail/alphabet_proxy.hpp
```diff
@@ -95,6 +93,37 @@ class alphabet_proxy : public alphabet_type
 
 #if 1// this is the one that works for most things, but not all
 
+namespace seqan3::detail
+{
+
+//!\brief Auxiliary transformation trait that is void for types by default.
+//!\ingroup alphabet
+template <typename t>
+struct alphabet_char_type_or_void
```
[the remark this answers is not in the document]
h-2: We don't have the `underlying_char` without `_t` anymore, that's why the change was necessary. The previous design suggested that it was a customisation point, because you could specialise it (which it was before). Now that it's not any longer, I wanted to remove the possibility (the template alias cannot be specialised by definition).

include/seqan3/core/detail/customisation_point.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+//!\brief A macro that helps defining the overload set of a customisation point.
+#define SEQAN3_CPO_IMPL(PRIO, TERM)                                                                                  \
+/*!\brief A customisation point overload.*/                                                                          \
+template <typename t, typename ...arg_ts>                                                                            \
+static constexpr decltype(auto) impl(seqan3::detail::priority_tag<PRIO> const &,                                     \
+                                     t && v,                                                                         \
+                                     [[maybe_unused]] arg_ts && ... args)                                            \
+    noexcept(noexcept(TERM))                                                                                         \
+    requires requires (seqan3::detail::priority_tag<PRIO> const &, t && v, [[maybe_unused]] arg_ts && ... args)      \
```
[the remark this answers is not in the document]
h-2: good point

include/seqan3/core/metafunction/basic.hpp
h-2: I can think of other places where this could be useful, but I can also move it to detail, if you prefer.

test/snippet/alphabet/assign_rank_to.cpp
h-2: So sad.

h-2:
> @rrahn did you have a look at this? Do you have a clue whats going on?

I have a workaround that can hopefully be reverted once the alphabet concepts are split with the next PR.

include/seqan3/core/detail/customisation_point.hpp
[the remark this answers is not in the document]
h-2: Wow, I have just reverted removing this, because it's the only way doxygen will accept the above code. Note that due to the macro I cannot put `//!\cond` blocks around the requires and `/*\cond*/` does not work. Apparently the `::` confuses Doxygen in that line so it stops parsing the rest of the line which makes everything work. If I remove the tag it somehow thinks the arguments to `requires requires` are variable definitions of type `requires` :man_shrugging: 

In any case I have reverted this and documented the hack.

h-2:
> I have a workaround that can hopefully be reverted once the alphabet concepts are split with the next PR.

I have found a more permanent fix to the problem that makes the workaround obsolete.

PR #933, opened 2019-04-25
[INFRA] Enable header tests for search module

eseiler:
<s>**Blocked by** https://github.com/xxsds/sdsl-lite/pull/66</s>

Now that the SDSL passes the header tests, we can re-enable the header tests for the search module.

PR #946, opened 2019-05-03
[FEATURE] shape, shape_iterator and view::kmer_hash

eseiler:
Supersedes #922 

- [x] Await judgement by codecov


include/seqan3/search/kmer_index/shape_iterator.hpp
```diff
@@ -0,0 +1,473 @@
...
+    //!\brief The size type of this iterator.
+    using position_type = std::make_unsigned_t<difference_type>;
+    //!\brief The alphabet type of the passed iterator.
+    using alphabet_t = value_type_t<it_t>;
+
+    //!\brief The hash value.
+    size_t hash_value{0};
+
+    //!\brief The seqan3::shape to use for hashing.
+    shape s;
```
smehringer: I would prefer longer variable names. grep and replace is tedious otherwise and it is not expressive.. maybe `shape_` or `kmer_shape`?

include/seqan3/search/kmer_index/shape_iterator.hpp
```diff
@@ -0,0 +1,473 @@
...
+        }
+    }
+
+    /*!\brief Decrements iterator by `skip`.
+     * \param skip Amount to decrement.
+     * \attention This function is only avaible if it_t models std::BidirectionalIterator.
+     */
+    void hash_reverse(difference_type const skip)
+    {
+        if (ungapped_shape && static_cast<size_t>(skip) < std::ranges::size(s))
```
smehringer: why do you use ranges::size on a shape? (s is the shape I guess) it is not a range is it? and if you know the type and it has a member function you can use it I think.

include/seqan3/search/kmer_index/shape_iterator.hpp
```diff
@@ -0,0 +1,473 @@
...
+        std::advance(text_left,  -1);
+        std::advance(text_right, -1);
+
+        hash_value /= alphabet_size_v<alphabet_t>;
+        hash_value += to_rank(*(text_left)) * roll_factor;
+    }
+
+public:
+    //!\brief Iterator to the leftmost position of the k-mer.
+    it_t text_left;
```
smehringer: why are those public?

include/seqan3/search/kmer_index/shape_iterator.hpp
eseiler: it is a `std::array`.. I'm not so sure about using the member function

include/seqan3/search/kmer_index/shape.hpp
```diff
@@ -0,0 +1,99 @@
...
+    *
+    * ### Exceptions
+    *
+    * No-throw guarantee.
+    *
+    * \attention The size of the pack must be in the interval [1, 32].
+    */
+    template <typename... t>
+    //!\cond
+        requires sizeof...(t) >= 2 && sizeof...(t) <= 32 && (std::Integral<t> && ...)
```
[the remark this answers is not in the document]
smehringer: ah I see.. well then it is fine I guess

eseiler:
|      Benchmark (name/sequence_length/k)        |     Throughput[Bp/s]        |    Time    |    CPU    | Iterations |
|------------------------------------------------|-----------------------------|------------|-----------|------------|
| seqan2_ungapped/1000/8                         |          338.809M/s         |    3044 ns |   2952 ns |   228884   |
| shape_iterator_ungapped/1000/8                 |          361.079M/s         |    2781 ns |   2769 ns |   238861   |
| seqan2_ungapped/1000/32                        |          344.308M/s         |    2989 ns |   2904 ns |   235807   |
| shape_iterator_ungapped/1000/32                |          364.317M/s         |    2758 ns |   2745 ns |   243567   |
| seqan2_ungapped/50000/8                        |          332.346M/s         |  169627 ns | 150446 ns |     4601   |
| shape_iterator_ungapped/50000/8                |          357.357M/s         |  140485 ns | 139916 ns |     4645   |
| seqan2_ungapped/50000/32                       |          353.45M/s          |  142501 ns | 141463 ns |     4428   |
| shape_iterator_ungapped/50000/32               |          354.364M/s         |  142111 ns | 141098 ns |     4805   |
| seqan2_gapped/1000/8                           |          338.198M/s         |    2967 ns |   2957 ns |   219057   |
| shape_iterator_gapped/1000/8                   |          134.621M/s         |    7476 ns |   7428 ns |    92921   |
| seqan2_gapped/1000/32                          |          115.799M/s         |    8682 ns |   8636 ns |    69363   |
| shape_iterator_gapped/1000/32                  |          45.703M/s          |   21987 ns |  21880 ns |    30026   |
| seqan2_gapped/50000/8                          |          335.232M/s         |  149748 ns | 149150 ns |     4448   |
| shape_iterator_gapped/50000/8                  |          131.69M/s          |  383638 ns | 379679 ns |     1797   |
| seqan2_gapped/50000/32                         |          111.697M/s         |  450866 ns | 447640 ns |     1582   |
| shape_iterator_gapped/50000/32                 |          45.3863M/s         | 1105704 ns |1101654 ns |      615   |
| shape_iterator_ungapped_random_access/1000/8   |          164.162M/s         |   61225 ns |  60915 ns |    11910   |
| shape_iterator_ungapped_random_access/1000/32  |          46.9264M/s         |  214028 ns | 213100 ns |     3143   |
| shape_iterator_ungapped_random_access/50000/8  |          158.698M/s         |   63233 ns |  63013 ns |    10814   |
| shape_iterator_ungapped_random_access/50000/32 |          47.0508M/s         |  213453 ns | 212536 ns |     2809   |
| shape_iterator_gapped_random_access/1000/8     |          169.638M/s         |   59281 ns |  58949 ns |    12044   |
| shape_iterator_gapped_random_access/1000/32    |          53.3727M/s         |  188106 ns | 187362 ns |     3185   |
| shape_iterator_gapped_random_access/50000/8    |          165.014M/s         |   60950 ns |  60601 ns |    10914   |
| shape_iterator_gapped_random_access/50000/32   |          51.8345M/s         |  193616 ns | 192922 ns |     3386   |

include/seqan3/search/kmer_index/shape.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+class shape : public std::array<bool, 32>
+{
+private:
+    //!\brief The size of the shape.
+    size_t k;
+
+public:
+
+    using std::array<bool, 32>::array;
```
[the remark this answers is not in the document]
eseiler: It's a `using`

include/seqan3/search/kmer_index/shape_iterator.hpp
```diff
@@ -0,0 +1,473 @@
...
+    //!\brief The alphabet type of the passed iterator.
+    using alphabet_t = value_type_t<it_t>;
+
+    //!\brief The hash value.
+    size_t hash_value{0};
+
+    //!\brief The seqan3::shape to use for hashing.
+    shape s;
+
+    //!\brief The factor for the left most position of the hash value.
```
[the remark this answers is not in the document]
eseiler: Mhm it's the factor when you calculate the rolling hash, if you want to substract the left most positions, the numerical value is `roll_factor * rank(leftmost_character)`. 

I have to explain rolling hash anyway in the docs I guess.

include/seqan3/search/kmer_index/shape_iterator.hpp
```diff
@@ -0,0 +1,497 @@
...
+    {
+        std::advance(text_left, -skip);
+        hash_full();
+    }
+
+    //!\brief Calculates a hash value by explicitly looking at each position.
+    void hash_full()
+    {
+        text_right = text_left;
+        hash_value = shape_[0] * to_rank(*text_right);
```
smehringer: what if the first position is "not inlcuded"? (it doesn't make much sense since you can just make your shape smaller by one... but we are also not enforcing/asserting this "studpidity")

include/seqan3/search/kmer_index/shape_iterator.hpp
```diff
@@ -0,0 +1,497 @@
...
+                                     std::next(std::ranges::begin(shape_), std::ranges::size(shape_)),
+                                     [](bool b) { return b; } );
+
+        roll_factor = std::pow(sigma, std::ranges::size(shape_) - 1);
+
+        hash_full();
+    };
+    //!\}
+
+    //!\anchor shape_iterator_comparison
```
smehringer: What's anchor doing :eyes: ?

include/seqan3/search/kmer_index/shape_iterator.hpp
```diff
@@ -0,0 +1,497 @@
...
+
+    /*!\brief Construct from a given iterator on the text and a seqan3::shape.
+    * /param[in] it_start Iterator pointing to the first position of the text.
+    * /param[in] s_ seqan3::shape
+    *
+    * ### Complexity
+    *
+    * Linear in size of shape.
+    */
+    shape_iterator(it_t it_start, shape s_) : shape_(s_),
```
smehringer: I think the style guide is 
```
myclass(...) :
    mem1{0}, mem2{0}, ...
```

include/seqan3/search/kmer_index/shape_iterator.hpp
eseiler: Mhm, I think it doesn't matter for the computation...it's still correct, just less efficient - in the worst case you have something like `0111` where you could just start computing the hashes one position later and use the `111` shape.
I guess we could add an assert to the constructor? AFAIK we assert in seqan2 that a shape ends and starts with a 1.

include/seqan3/search/kmer_index/shape_iterator.hpp
eseiler: I can do a `\ref` to an `\anchor`. But not to a `\name`.

include/seqan3/search/kmer_index/shape_iterator.hpp
smehringer: Your right it's correct anyway, I didn't see that. I think asserts don't hurt :)

include/seqan3/search/kmer_index/all.hpp
```diff
@@ -0,0 +1,16 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+#pragma once
```
smehringer: pragma once below author ? :nail_care:

include/seqan3/search/kmer_index/shape.hpp
```diff
@@ -0,0 +1,106 @@
...
+*/
+
+class shape : public std::array<bool, 32>
+{
+private:
+    //!\brief The size of the shape.
+    size_t k;
+
+public:
+
```
smehringer: no newline after `public:` :nail_care:

include/seqan3/search/kmer_index/shape.hpp
```diff
@@ -0,0 +1,106 @@
...
+     * \{
+     */
+    shape()                                    noexcept = default; //!< Defaulted.
+    constexpr shape(shape const &)             noexcept = default; //!< Defaulted.
+    constexpr shape(shape &&)                  noexcept = default; //!< Defaulted.
+    constexpr shape & operator=(shape const &) noexcept = default; //!< Defaulted.
+    constexpr shape & operator=(shape &&)      noexcept = default; //!< Defaulted.
+    ~shape()                                   noexcept = default; //!< Defaulted.
+
+    /*!\brief Construct from a given size.
```
smehringer:
```suggestion
    /*!\brief Construct an ungapped shape from a given size.
```
Or add in the brief section that this will fill the shape with "every position is important" or somehting like that.

include/seqan3/search/kmer_index/shape.hpp
```diff
@@ -0,0 +1,106 @@
...
+    * ### Complexity
+    *
+    * Constant.
```
smehringer: you sure? `fill()` is in *O(k)* ?

include/seqan3/search/kmer_index/shape.hpp
```diff
@@ -0,0 +1,106 @@
...
+    {
+        assert(k > 0 && k < 33);
+        fill(false);
+        std::fill_n(begin(), k_, true);
+    }
+
+    /*!\brief Construct from a parameter pack.
+    * ### Complexity
+    *
+    * Constant
```
smehringer: Also not correct, it is linear in the number of arguments.

include/seqan3/search/kmer_index/shape_iterator.hpp
```diff
@@ -0,0 +1,497 @@
...
+    //!\endcond
+class shape_iterator
+{
+public:
+    //!\brief Type for distances between iterators.
+    using difference_type = typename it_t::difference_type;
+    //!\brief Value type of this iterator.
+    using value_type = size_t;
+    //!\brief The reference type of this iterator.
+    using reference = shape_iterator &;
```
smehringer: the reference type should be `value_type & `??

include/seqan3/search/kmer_index/shape_iterator.hpp
smehringer: not resolved

include/seqan3/search/kmer_index/shape_iterator.hpp
eseiler: Anyway, the check is removed.

include/seqan3/search/kmer_index/shape.hpp
eseiler: mhm, `fill()` is in `size(array)`, but since our array is fixed size, it's constant? I.e. size is always 32

smehringer: Yes I would say linear in `k`. Otherwise access time `[]` is super constant as it does not even depend on `k` :grin:

eseiler:
I'll have a look at the `inherited_iterator_base`

include/seqan3/search/kmer_index/shape_iterator.hpp
```diff
@@ -0,0 +1,494 @@
...
+    //!\brief The hash value.
+    size_t hash_value{0};
+
+    //!\brief The seqan3::shape to use for hashing.
+    shape shape_;
+
+    //!\brief The factor for the left most position of the hash value.
+    size_t roll_factor{0};
+
+    //!\brief True if `shape_` contains only 1 (ungapped), false otherwise.
```
smehringer:
```suggestion
    //!\brief True if `shape_` contains only 1's at [0, k-1] (ungapped), false otherwise.
```

test/snippet/search/kmer_index/shape_iterator.cpp
```diff
@@ -0,0 +1,20 @@
...
+#include <seqan3/search/kmer_index/shape_iterator.hpp>
+#include <seqan3/io/stream/debug_stream.hpp>
+
+using seqan3::operator""_dna4;
+
+int main()
+{
+//![gapped]
+    auto text{"ACGTAGCT"_dna4}; // The text to hash
+    seqan3::shape s(1, 0, 1);   // An gapped 3-mer
```
smehringer:
```suggestion
    seqan3::shape s{1, 0, 1};   // An gapped 3-mer
```
:point_up:

test/unit/search/kmer_index/shape_test.cpp
```diff
@@ -0,0 +1,55 @@
...
+    EXPECT_TRUE((std::is_trivially_move_constructible_v<shape>));
+    EXPECT_TRUE((std::is_nothrow_move_constructible_v<shape>));
+    EXPECT_TRUE((std::is_copy_assignable_v<shape>));
+    EXPECT_TRUE((std::is_trivially_copy_assignable_v<shape>));
+    EXPECT_TRUE((std::is_nothrow_copy_assignable_v<shape>));
+    EXPECT_TRUE((std::is_move_assignable_v<shape>));
+    EXPECT_TRUE((std::is_trivially_move_assignable_v<shape>));
+    EXPECT_TRUE((std::is_nothrow_move_assignable_v<shape>));
+
+    shape s0(0, 0, 0);
```
smehringer:
```suggestion
    shape s0{0, 0, 0};
```
also everywhere else

eseiler:
Benchmark[seq_len/k]                 |Throughput[bp/s]  |     Time     |        CPU   |Iterations 
--------------------------|------------------|-------------|--------------|-----------
seqan_kmer_hash/1000/8    |348.089M/s        |    2867 ns     |    2853 ns   |    243795 
seqan_kmer_hash/1000/32   |325.167M/s        |    2993 ns     |    2980 ns   |    238183 
seqan_kmer_hash/50000/8   |355.52M/s         | 141154 ns     |  140619 ns   |      4845 
seqan_kmer_hash/50000/32  |353.035M/s        |  142342 ns     |  141541 ns   |      4687 
std_kmer_hash/1000/8      |241.554M/s        |    4127 ns     |    4111 ns   |    163673 
std_kmer_hash/1000/32     |63.5777M/s        |   15299 ns     |   15241 ns   |     44485 
std_kmer_hash/50000/8     |232.499M/s        |  216047 ns     |  215025 ns   |      3253 
std_kmer_hash/50000/32    |63.3054M/s        |  792137 ns     |  789332 ns   |       870

include/seqan3/search/kmer_index/detail/shape_iterator.hpp
```diff
@@ -0,0 +1,450 @@
...
+    * ### Complexity
+    *
+    * Linear in size of shape.
+    */
+    shape_iterator(it_t it_start, shape s_) : 
+        shape_{s_}, text_start{it_start}, text_left{it_start}, text_right{it_start}
+    {
+        assert(std::ranges::size(shape_) > 0);
+        ungapped_shape = std::all_of(std::ranges::begin(shape_),
+                                     std::next(std::ranges::begin(shape_), std::ranges::size(shape_) - 1),
```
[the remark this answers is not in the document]
eseiler: Yes, I didn't change it when I switched to the constexpr vector

include/seqan3/search/kmer_index/detail/shape_iterator.hpp
```diff
@@ -0,0 +1,450 @@
...
+        requires std::RandomAccessIterator<it_t>
+        //!\endcond
+    {
+        return static_cast<difference_type>(text_right - lhs.text_right);
+    }
+
+    /*!\brief Move the iterator to a given position and return the corresponding hash value.
+     * \attention This function is only avaible if it_t models std::RandomAccessIterator.
+     */
+    value_type operator[](std::make_unsigned_t<difference_type> const n)
```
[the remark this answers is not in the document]
eseiler: Probably yes, but it depends on the passed iterator type; so I'd leave it as is?

include/seqan3/search/kmer_index/detail/shape_iterator.hpp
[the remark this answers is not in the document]
eseiler: `std::ranges::all_of` is even better 😁

include/seqan3/search/kmer_index/detail/shape_iterator.hpp
```diff
@@ -0,0 +1,461 @@
...
+        if (lhs.shape_ != rhs.shape_)
+            return false;
+        return lhs.text_right < rhs.text_right;
+    }
+
+    //\!brief Compare to another shape_iterator.
+    friend bool operator>(shape_iterator const & lhs, shape_iterator const & rhs) noexcept
+    {
+        if (lhs.shape_ != rhs.shape_)
+            return false;
```
smehringer: your implementation of < and > is a little weird I think because usually, you would expect: if !(a < b) then a >= b.. but your operators always return false if the shapes are unequal. I don't know if this kind of comparison is useful if there is the possibility that the shape is different (we cannot check this at compile time)...

eseiler:
> I think you have some commits that dont belong into this PR?

Yes, I rebased on #933 to run the header test.

include/seqan3/range/view/kmer_hash.hpp
```diff
@@ -12,47 +12,163 @@
...
+
+//!\brief A deduction guide for the view class template.
+template <std::ranges::ViewableRange orng_t>
+kmer_hash_view(orng_t &&, shape const shape_) -> kmer_hash_view<std::ranges::all_view<orng_t>>;
+
+// ---------------------------------------------------------------------------------------------------------------------
+// kmer_hash_fn (adaptor definition)
+// ---------------------------------------------------------------------------------------------------------------------
 
 //![adaptor_def]
```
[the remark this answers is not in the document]
eseiler: ?

The `//![adaptor_def]` ? Some documentation relies on this as an example.

include/seqan3/range/view/kmer_hash.hpp
```diff
@@ -61,42 +177,51 @@ struct kmer_hash_fn
...
-     * | seqan3::reference_t             | seqan3::Semialphabet                  | std::size_t                                        |
-     *
-     * See the \link view view submodule documentation \endlink for detailed descriptions of the view properties.
-     *
-     * ### Example
-     * \snippet test/snippet/range/view/kmer_hash.cpp usage
-     * \hideinitializer
-     */
-    inline auto constexpr kmer_hash = detail::kmer_hash_fn{};
+/*!\name Alphabet related views
```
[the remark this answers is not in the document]
eseiler: This is from the old kmer_hash_view documentation ..  and is it kinda alphabet related

include/seqan3/search/kmer_index/detail/shape_iterator.hpp
eseiler: True, I should probably at least compare the shape sizes.

include/seqan3/range/view/kmer_hash.hpp
```diff
@@ -12,47 +12,163 @@
...
+// ---------------------------------------------------------------------------------------------------------------------
 
 //![adaptor_def]
 //!\brief view::kmer_hash's range adaptor object type (non-closure).
 struct kmer_hash_fn
 {
-    //!\brief Store the argument and return a range adaptor closure object.
-    constexpr auto operator()(size_t const k) const noexcept
+    //!\brief Returns an instance of seqan3::detail::kmer_hash_view constructed with \p urange and \p shape_.
+    template <std::ranges::ForwardRange urng_t>
```
h-2: only check Range here

include/seqan3/range/view/kmer_hash.hpp
```diff
@@ -12,47 +12,163 @@
...
+    auto operator()(urng_t && urange, shape const shape_) const
     {
-        return detail::adaptor_from_functor{*this, k};
```
h-2: static assert on ViewableRange and ForwardRange and Semialphabet inside. See for example view::translate

include/seqan3/range/view/kmer_hash.hpp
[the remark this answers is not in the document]
h-2: Please check whether that documentation is still accurately making use of this example now that you changed it completely :wink:

include/seqan3/range/view/kmer_hash.hpp
h-2: If Semialphabet is required?

include/seqan3/range/view/kmer_hash.hpp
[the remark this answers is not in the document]
h-2: Does it depend on the value type being an Alphabet or does it work on anything hashable?

include/seqan3/range/view/kmer_hash.hpp
```diff
@@ -61,42 +177,51 @@ struct kmer_hash_fn
...
+ * | std::ranges::ContiguousRange    |                                    | *lost*                           |
+ * |                                 |                                    |                                  |
+ * | std::ranges::ViewableRange      | *required*                         | *guaranteed*                     |
+ * | std::ranges::View               |                                    | *guaranteed*                     |
+ * | std::ranges::SizedRange         |                                    | *preserved*                      |
+ * | std::ranges::CommonRange        |                                    | *preserved*                      |
+ * | std::ranges::OutputRange        |                                    | *lost*                           |
+ * | seqan3::ConstIterableRange      |                                    | *preserved*                      |
+ * |                                 |                                    |                                  |
+ * | seqan3::reference_t             | seqan3::Semialphabet               | std::size_t                      |
```
h-2: Is it? See above.

include/seqan3/search/kmer_index/detail/shape_iterator.hpp
```diff
@@ -0,0 +1,461 @@
...
+    //!\brief Type for distances between iterators.
+    using difference_type = typename it_t::difference_type;
+    //!\brief Value type of this iterator.
+    using value_type = size_t;
+    //!\brief Pointer to `value_type`.
+    using pointer = value_type *;
+    //!\brief Reference to `value_type`.
+    using reference = value_type &;
+    //!\brief Reference to `value_type const`.
+    using const_reference = value_type const &;
```
h-2: Please double check these, I don't think they are correct, e.g. the reference type can't be `value_type &` as you are never returning references! Also the `pointer` type implies that you have `operator->` which you don't so I would remove it...

include/seqan3/search/kmer_index/detail/shape_iterator.hpp
```diff
@@ -0,0 +1,461 @@
...
+    //!\brief Tag this class depending on which concept it_t models.
+    using iterator_category = std::conditional_t<std::RandomAccessIterator<it_t>,
```
h-2: A little shorter:
```suggestion
    using iterator_category = std::conditional_t<std::ContiguousIterator<it_t>,
                                                 typename std::random_access_iterator_tag,
                                                 iterator_tag<it_t>;
```

include/seqan3/search/kmer_index/detail/shape_iterator.hpp
```diff
@@ -0,0 +1,461 @@
...
+ * ### Implementation detail
+ *
+ * To avoid dereferencing the sentinel when iterating, e.g. when using a for loop:
+ * \snippet test/snippet/search/kmer_index/shape_iterator.cpp detail
+ * the shape_iterator computes the hash value up until the second to last position and performs the addition of the 
+ * last position upon access (\ref operator* and \ref operator[]).
+ */
+template <std::ranges::ForwardRange urng_t>
+    //!\cond
+    requires Semialphabet<delete_const_t<reference_t<urng_t>>>
```
h-2: pending the changes in #979 the delete_const won't be necessary

include/seqan3/search/kmer_index/detail/shape_iterator.hpp
```diff
@@ -0,0 +1,461 @@
...
+
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    shape_iterator()                                   = default; //!< Defaulted.
+    shape_iterator(shape_iterator const &)             = default; //!< Defaulted.
+    shape_iterator(shape_iterator &&)                  = default; //!< Defaulted.
+    shape_iterator & operator=(shape_iterator const &) = default; //!< Defaulted.
+    shape_iterator & operator=(shape_iterator &&)      = default; //!< Defaulted.
+    ~shape_iterator()                                  = default; //!< Defaulted.
```
h-2: Which of these could be constexpr`

include/seqan3/search/kmer_index/detail/shape_iterator.hpp
```diff
@@ -0,0 +1,461 @@
...
+    bool ungapped_shape{false};
+
+    //!\brief Iterator pointing to begin of the underlying text.
+    it_t text_start;
+
+    //!\brief Iterator to the leftmost position of the k-mer.
+    it_t text_left;
+
+    //!\brief Iterator to the rightmost position of the k-mer.
+    it_t text_right;
```
h-2: This iterator is quite big :open_mouth: 

1. Please evaluate if you can reduce the size_t variables, e.g. what is the possibly largest value of `roll_factor`?
1. Can't you encode `ungapped_shape` within the shape?
1. Using the small_vector as base class is convenient, but inefficient, it is 33 bytes, although it could be done <=5 bytes. We need a constexpr bitset :expressionless: 
1. Do you really need text_start? Keep in mind that every iterator is 8bytes
1. We don't need the shape inside the iterator and inside the view. Ideally we would have a bitset, and than it would be more efficient to have it in the iterator, but if we don't have that, maybe we should have it jus tin the the view for now and use a pointer to the view.

test/include/seqan3/test/performance/std_kmer_hash.hpp
```diff
@@ -0,0 +1,102 @@
...
+        });
+    }
+};
+//![adaptor_def]
+
+} // namespace seqan3::detail
+
+namespace seqan3::view
+{
+    /*!\name Alphabet related views
```
h-2: left align

test/performance/range/view/view_kmer_hash_benchmark.cpp
```diff
@@ -0,0 +1,77 @@
...
+        for (auto h : seq | view::std_kmer_hash(k))
+            benchmark::DoNotOptimize(sum += h);
+    }
+
+    state.counters["Throughput[bp/s]"] = bp_per_second(sequence_length - k + 1);
+}
+
+BENCHMARK(seqan_kmer_hash)->Apply(arguments);
+BENCHMARK(std_kmer_hash)->Apply(arguments);
+
```
h-2: Hm, are "seqan_kmer_hash" and "std_kmer_hash" good names? Both are seqan, only one of them is rolling, or not?

test/unit/range/view/view_kmer_hash_test.cpp
```diff
@@ -70,3 +71,26 @@ TEST(kmer_hash, const_view)
...
+        std::vector<size_t> expected{9,50,44,27,6};
+        EXPECT_EQ(expected, hashes);
+    }
+    // {
+    //     auto remove_a = std::view::filter([](dna4 const x) { return x != 'A'_dna4; });
+    //     // std::vector<dna4> tmp = text | remove_a;
+    //     std::vector<size_t> hashes = text | remove_a | view::kmer_hash(3);
+    //     std::vector<size_t> expected{27, 46, 57};
+    //     EXPECT_EQ(expected, hashes);
+    // }
```
h-2: ?

include/seqan3/range/view/kmer_hash.hpp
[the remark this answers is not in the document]
eseiler: Actually on anything hashable since I only do arithmetic coding for which i just need a representation as unsigned integer.
Do we want a `Hashable` concept or should I just check that `std::hash`'s `operator()` can be called on the type?

include/seqan3/search/kmer_index/shape.hpp
```diff
@@ -0,0 +1,94 @@
...
+     *
+     * Linear in k.
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     *
+     * \attention The size must be in the interval [1, 32]. In Debug mode, an assertion will check this constraint.
+     */
+    constexpr shape(size_t k_) noexcept : small_vector<bool, 32>(k_, true), k(k_) {}
```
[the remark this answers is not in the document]
eseiler: But does this work?
I want to use the `count, value` constructor and I though `{}` would use the initialiser list constructor?

test/performance/range/view/view_kmer_hash_benchmark.cpp
eseiler: Not in general, just for ungapped shapes.

include/seqan3/search/kmer_index/detail/shape_iterator.hpp
eseiler: > This iterator is quite big 😮
> 
> 1. Please evaluate if you can reduce the size_t variables, e.g. what is the possibly largest value of `roll_factor`?
dna4 32-mer = `2^64`, so `size_t` seems fine
> 2. Can't you encode `ungapped_shape` within the shape?
Do you mean within the small_vector or within the class?
> 3. Using the small_vector as base class is convenient, but inefficient, it is 33 bytes, although it could be done <=5 bytes. We need a constexpr bitset 😑
I had an array before, but since we want to replace it anyway, it doesn't matter too much?
> 4. Do you really need text_start? Keep in mind that every iterator is 8bytes
If we want to model random access ... what would `operator[]` do if you only have your current position?
> 5. We don't need the shape inside the iterator and inside the view. Ideally we would have a bitset, and than it would be more efficient to have it in the iterator, but if we don't have that, maybe we should have it jus tin the the view for now and use a pointer to the view.
Mhm, I see the problem, but if we want to change it anyways once the bitset is there, we may as well leave it for now....

include/seqan3/search/kmer_index/detail/shape_iterator.hpp
eseiler: I remember that something didn't work when I didn't have `pointer`, but now it works without.
It also seems that I do not need to define `reference` and `const_reference` - so should I remove it?

include/seqan3/range/view/kmer_hash.hpp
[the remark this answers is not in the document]
eseiler: **But** I need `alphabet_size`. So it's probably Semialphabet after all?

test/performance/range/view/view_kmer_hash_benchmark.cpp
smehringer: > Hm, are "seqan_kmer_hash" and "std_kmer_hash" good names? Both are seqan, only one of them is rolling, or not?

well the view is both seqan but the hashing is done with std::hash for std_kmer_hash

include/seqan3/search/kmer_index/detail/shape_iterator.hpp
smehringer: > > Do you really need text_start? Keep in mind that every iterator is 8bytes
> If we want to model random access ... what would operator[] do if you only have your current position?

Hm I really wonder if random access makes sense for hashing a sequence since the usual use case is to hash from start to end..

h-2:
Regarding the underlying data structure: was there a reason we didn't use an sdsl-bitvector instead of the small_vector of bool? That would be smaller and we don't have anything constexpr, yet, anyway, right?

In general a small_bitvector would be easy to do, but probably not this week.

I would recommend the following now:
  * make the shape a small_vector of uint8_t instead of bool and encode gapped/ungapped in a higher order bit of the first element (it's guaranteed to be non-zero anyway)  
  * move the shape into the view
  * remove the start iterator
  * holds pointer to view in the iterator to get access to first element and shape.

include/seqan3/search/kmer_index/detail/shape_iterator.hpp
eseiler: I need the `pointer`:

```
/Users/enricoseiler/git/seqan3/include/seqan3/core/metafunction/iterator.hpp: In instantiation of 'struct seqan3::reference<seqan3::detail::shape_iterator<ranges::ref_view<std::vector<seqan3::dna4> > > >':
/Users/enricoseiler/git/seqan3/include/seqan3/core/metafunction/pre.hpp:71:7:   required by substitution of 'template<class t> using reference_t = typename seqan3::reference::type [with t = seqan3::detail::shape_iterator<ranges::ref_view<std::vector<seqan3::dna4> > >]'
/Users/enricoseiler/git/seqan3/include/seqan3/core/metafunction/range.hpp:78:11:   required from 'struct seqan3::reference<seqan3::detail::kmer_hash_view<ranges::ref_view<std::vector<seqan3::dna4> > > >'
/Users/enricoseiler/git/seqan3/include/seqan3/core/debug_stream.hpp:370:20:   required from here
/Users/enricoseiler/git/seqan3/include/seqan3/core/metafunction/iterator.hpp:56:11: error: no type named 'reference' in 'struct std::iterator_traits<seqan3::detail::shape_iterator<ranges::ref_view<std::vector<seqan3::dna4> > > >'
   56 |     using type = typename std::iterator_traits<std::remove_reference_t<it_t>>::reference;
      |           ^~~~
```

include/seqan3/search/kmer_index/shape.hpp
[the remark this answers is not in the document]
h-2: Ah, correct.

include/seqan3/range/view/kmer_hash.hpp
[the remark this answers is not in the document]
h-2: Well, it's a design decision you have to make. I guess we can restrict it to semi-alphabets for now and still change it later. Please document this requirement clearly in the view interface!

include/seqan3/range/view/kmer_hash.hpp
```diff
@@ -12,47 +12,616 @@
...
+        size_t hash_value{0};
+        
+        //!\brief Pointer to underlying view.
+        kmer_hash_view const * pview{nullptr};
+
+        //!\brief The factor for the left most position of the hash value.
+        size_t roll_factor{0};
+
+        //!\brief True if `shape_` contains only 1's at [0, k-1] (ungapped), false otherwise.
+        bool ungapped_shape{false};
```
h-2: move this into the shape as discussed

include/seqan3/range/view/kmer_hash.hpp
```diff
@@ -12,47 +12,616 @@
...
+            assert(std::ranges::size(pview->shape_) > 0);
+            ungapped_shape = std::ranges::all_of(pview->shape_, [](bool b) { return b; } );
+
+            roll_factor = std::pow(sigma, std::ranges::size(pview->shape_) - 1);
+
+            hash_full();
+        }
+        //!\}
+
+        //!\anchor shape_iterator_comparison
```
h-2: What does anchor do?

include/seqan3/range/view/kmer_hash.hpp
```diff
@@ -61,47 +630,51 @@ struct kmer_hash_fn
...
-    inline auto constexpr kmer_hash = detail::kmer_hash_fn{};
+/*!\name Alphabet related views
+ * \{
+ */
+
+/*!\brief               Computes hash values for each position of a range via a given shape.
+ * \tparam urng_t       The type of the range being processed. See below for requirements. [template parameter is
+ *                      omitted in pipe notation]
+ * \param[in] urange    The range being processed. [parameter is omitted in pipe notation]
+ * \param[in] shape_    The seqan3::shape that determines how to compute the hash value.
```
h-2:
```suggestion
 * \param[in] shape     The seqan3::shape that determines how to compute the hash value.
```

This doesn't refer to the actual parameter anyway so one might as well use a nicer name,

include/seqan3/range/view/kmer_hash.hpp
```diff
@@ -61,47 +630,51 @@ struct kmer_hash_fn
...
+ * | std::ranges::BidirectionalRange |                                    | *preserved*                      |
+ * | std::ranges::RandomAccessRange  |                                    | *preserved*                      |
+ * | std::ranges::ContiguousRange    |                                    | *lost*                           |
+ * |                                 |                                    |                                  |
+ * | std::ranges::ViewableRange      | *required*                         | *guaranteed*                     |
+ * | std::ranges::View               |                                    | *guaranteed*                     |
+ * | std::ranges::SizedRange         |                                    | *preserved*                      |
+ * | std::ranges::CommonRange        |                                    | *preserved*                      |
+ * | std::ranges::OutputRange        |                                    | *lost*                           |
+ConstIterableRangeRange      |                                    | *preserved*                      |
```
h-2: something broken hare

include/seqan3/search/kmer_index/detail/shape_iterator.hpp
h-2: You need reference, but you don't need pointer and const_reference. If something breaks without pointer being defined, set it to void.

include/seqan3/search/kmer_index/detail/shape_iterator.hpp
h-2: Not implemented yet, still has more conditionals.

include/seqan3/range/view/kmer_hash.hpp
```diff
@@ -12,47 +12,616 @@
...
+ * \implements std::ranges::View
+ * \implements std::ranges::RandomAccessRange
+ * \implements std::ranges::SizedRange
+ * \ingroup view
+ *
+ * \details
+ *
+ * Note that most members of this class are generated by ranges::view_interface which is not yet documented here.
+ */
+template <std::ranges::View urng_t>
```
h-2: check for Semialphabet here?

include/seqan3/range/view/kmer_hash.hpp
```diff
@@ -12,47 +12,616 @@
...
+        
+        /*!\brief Construct from a given iterator on the text and a seqan3::shape.
+        * /param[in] it_start Iterator pointing to the first position of the text.
+        * /param[in] s_ The seqan3::shape that determines which positions participate in hashing.
+        *
+        * ### Complexity
+        *
+        * Linear in size of shape.
+        */
+        shape_iterator(it_t it_start, kmer_hash_view const * v_) :
```
h-2: I would take by reference and take the address of inside, but I guess that's a matter of taste

include/seqan3/search/kmer_index/detail/shape_iterator.hpp
h-2: We should optimise the use-case for ForwardRange access, but it doesn't hurt to have RandomAccess.

test/performance/range/view/view_kmer_hash_benchmark.cpp
h-2: I still think it's confusing. `std_kmer_hash` looks like its was referrring to `std::kmer::hash` or something else in `std::`....

include/seqan3/range/view/kmer_hash.hpp
```diff
@@ -61,47 +630,51 @@ struct kmer_hash_fn
...
+ *
+ * See the \link view view submodule documentation \endlink for detailed descriptions of the view properties.
+ *
+ * ### Example
+ *
+ * \include test/snippet/range/view/kmer_hash.cpp
+ *
+ * \hideinitializer
+ */
+constexpr inline detail::kmer_hash_fn kmer_hash{};
```
h-2:
```suggestion
inline constexpr  auto kmer_hash = detail::kmer_hash_fn{};
```

Just to be inline with the other ones...

include/seqan3/search/kmer_index/shape.hpp
```diff
@@ -0,0 +1,91 @@
...
+* value.
+*
+* ### Example
+*
+* \include test/snippet/search/kmer_index/shape.cpp
+*
+* \attention 0 < size <= 32
+*/
+
+class shape : public small_vector<bool, 32>
```
h-2: I would maybe make this `small_vector<bool, 31>` and then the whole thing will be 32 bytes big and you can use the 31st bool to indicate whether its gapped or ungapped. But then you only have 30 bools for the shape.

include/seqan3/search/kmer_index/shape.hpp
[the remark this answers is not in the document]
h-2: yeah, the binary literal thing will conflict with the length constructor....

include/seqan3/search/kmer_index/shape.hpp
```diff
@@ -0,0 +1,91 @@
...
+     *
+     * \attention The size of the pack must be in the interval [1, 32].
+     */
+    template <typename... t>
+    //!\cond
+        requires sizeof...(t) >= 2 && sizeof...(t) <= 32 && (std::Integral<t> && ...)
+    //!\endcond
+    constexpr shape(t... args) noexcept
+    {
+        assert([] (auto... a) constexpr { return ((a == 0 || a == 1) && ...); }(args...)); // Only 0 and 1 allowed
```
h-2: Can't you do 
```cpp
assert( ((args == 0) || (args == 1) && ...) );
```

test/snippet/range/view/kmer_hash.cpp
```diff
@@ -1,14 +1,17 @@
...
-    std::vector<size_t> hashes = text | view::kmer_hash(3);
-    debug_stream << hashes << '\n'; // [6,27,44,50,9]
+    std::vector<seqan3::dna4> text{"ACGTAGC"_dna4};
+    
+    std::vector<size_t> hashes = text | seqan3::view::kmer_hash(seqan3::shape(3));
+    seqan3::debug_stream << hashes << '\n'; // [6,27,44,50,9]
+    
+    seqan3::debug_stream << (text | seqan3::view::kmer_hash(3)) << '\n'; // [6,27,44,50,9]
+    
+    seqan3::debug_stream << (text | seqan3::view::kmer_hash(seqan3::shape(1,0,1))) << '\n'; // [2,7,8,14,1]
```
h-2:
```suggestion
    seqan3::debug_stream << (text | seqan3::view::kmer_hash(seqan3::shape{1,0,1})) << '\n'; // [2,7,8,14,1]
```

test/snippet/range/view/kmer_hash.cpp
```diff
@@ -1,14 +1,17 @@
...
+using seqan3::operator""_dna4;
 
 int main()
 {
-    std::vector<dna4> text{"ACGTAGC"_dna4};
-    std::vector<size_t> hashes = text | view::kmer_hash(3);
-    debug_stream << hashes << '\n'; // [6,27,44,50,9]
+    std::vector<seqan3::dna4> text{"ACGTAGC"_dna4};
+    
+    std::vector<size_t> hashes = text | seqan3::view::kmer_hash(seqan3::shape(3));
```
h-2:
```suggestion
    std::vector<size_t> hashes = text | seqan3::view::kmer_hash(seqan3::shape(3));
```
Hm, it's really not good to mix a constructor that takes length and a constructor that takes a pack of the same type. Can you make a strong_type of the length parameter (maybe called it `ungapped`?).

include/seqan3/range/view/kmer_hash.hpp
eseiler: I can link to it in the documentation

include/seqan3/search/kmer_index/shape.hpp
eseiler:
```
shape.hpp:87:30: error: binary expression in operand of fold-expression
   87 |         assert( ((args == 0) || (args == 1) && ...) );
```

🤷‍♂

test/performance/range/view/view_kmer_hash_benchmark.cpp
eseiler: Renamed to `naive_kmer_hash`

eseiler:
seq_length | k  | SeqAn2     | SeqAn3     | naive
-----------|----|------------|------------| ----------
1000       | 8  | 361.766M/s | 356.063M/s | 248.596M/s
1000       | 30 | 361.14M/s  | 352.636M/s | 69.5414M/s
50000      | 8  | 359.345M/s | 362.559M/s | 241.435M/s
50000      | 30 | 361.306M/s | 361.882M/s | 69.3594M/s

test/snippet/range/view/kmer_hash.cpp
eseiler: Now that I have a strong_type for this, I could add the literal constructor for `0b011010101` and stuff like that...

include/seqan3/search/kmer_index/shape.hpp
h-2: This will work:
```cpp
assert( ((args == 0 || args == 1) && ...) );
```
(the same syntax you use inside the lambda :wink: )

include/seqan3/range/view/kmer_hash.hpp
```diff
@@ -61,47 +653,51 @@ struct kmer_hash_fn
...
+ * | std::ranges::BidirectionalRange |                                    | *preserved*                      |
+ * | std::ranges::RandomAccessRange  |                                    | *preserved*                      |
+ * | std::ranges::ContiguousRange    |                                    | *lost*                           |
+ * |                                 |                                    |                                  |
+ * | std::ranges::ViewableRange      | *required*                         | *guaranteed*                     |
+ * | std::ranges::View               |                                    | *guaranteed*                     |
+ * | std::ranges::SizedRange         |                                    | *preserved*                      |
+ * | std::ranges::CommonRange        |                                    | *lost*                           |
+ * | std::ranges::OutputRange        |                                    | *lost*                           |
+ * | ConstIterableRangeRange         |                                    | *preserved*                      |
```
h-2:
```suggestion
 * | seqan3::ConstIterableRangeRange |                                    | *preserved*                      |
```

include/seqan3/range/view/kmer_hash.hpp
```diff
@@ -12,47 +12,639 @@
...
+            std::ranges::advance(text_right, -1);
+            
+            hash_value /= sigma;
+            hash_value -= to_rank(*(text_right));
+            hash_value += to_rank(*(text_left)) * roll_factor;
+        }
+    };
+    
+public:
+    static_assert(std::ranges::ForwardRange<urng_t const>, "The kmer_hash_view only works on ForwardRanges");
```
h-2: I would put this as first thing into the class so that it doesn't get lost and is seen together with the requirements.

include/seqan3/search/kmer_index/shape.hpp
```diff
@@ -0,0 +1,141 @@
...
+    using detail::strong_type<uint8_t, ungapped>::strong_type;
+    
+    //!\brief Construct from an unsigned integral.
+    template<std::UnsignedIntegral t>
+    //!\cond
+        requires !std::Same<t, uint8_t>
+    //!\endcond
+    constexpr ungapped(t value) noexcept: ungapped(static_cast<uint8_t>(value))
+    {
+        assert(value <= 255);
```
h-2: But actually only 30 is allowed, right?

include/seqan3/search/kmer_index/shape.hpp
```diff
@@ -0,0 +1,141 @@
...
+    }
+};
+
+/*!\brief A class that defines what positions of a pattern to hash.
+*
```
h-2: \ingroup is missing; also for the strong type

test/snippet/range/view/kmer_hash.cpp
h-2: Yes, I would make two strong types
  * `ungapped`
  * `pattern` or `bin_literal` (bin_literal might be usable in other contexts as well)

And then I would remove the parameter pack constructor, but that's up to you.

include/seqan3/range/view/kmer_hash.hpp
```diff
@@ -12,47 +12,639 @@
...
+    //!\brief Store the shape and return a range adaptor closure object.
+    constexpr auto operator()(shape const shape_) const
+    {
+        return adaptor_from_functor{*this, shape_};
+    }
+    
+    //!\brief Construct a ungapped shape from the argument and return a range adaptor closure object.
+    constexpr auto operator()(ungapped const k) const
+    {
+        return adaptor_from_functor{*this, shape{k}};
```
h-2: If you have implicit constructors (like you do right now), you can just remove this overload and a shape will automatically be created when an ungapped or bin_literal value is passed.

h-2: Having implicit single-argument constructors here is ok from my POV, because the strong types are created explicitly for this purpose.

test/snippet/range/view/kmer_hash.cpp
h-2: In fact, you could even add a user defined numeric literal so one can do the following `0b11001101_shape` and it will return a shape. Pretty cool. But has room for confusion, because `5_shape` would also work and resolve to `0b101`...

include/seqan3/search/kmer_index/shape.hpp
```diff
@@ -0,0 +1,141 @@
...
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    constexpr size_type size() const noexcept
+    {
+        return sz - 1;
+    }
+};
```
h-2: Hm, maybe I was to eager with integrating the bool here. Because now one can flip a bit via [] and the ungapped bit will be wrong :(

h-2: You would need to overload the operator[], as well to prevent that.

include/seqan3/range/view/kmer_hash.hpp
```diff
@@ -12,47 +12,639 @@
...
+     * Constant.
+     *
+     */
+    auto size() const noexcept
+    //!\cond
+        requires std::SizedSentinel<std::ranges::sentinel_t<urng_t>, std::ranges::iterator_t<urng_t>>
+    //!\endcond
+    {
+        return end() - begin();
+    }
```
h-2: This is provided by the view_interface already

include/seqan3/search/kmer_index/shape.hpp
eseiler: I think I would have to overload quite a bit if I want to do it fail safe... `back()`, `insert()`, `operator[]`, if it's default constructed the last position is not yet the ungapped indicator....
So maybe just keep a bool in the iterator or view for now?

include/seqan3/search/kmer_index/shape.hpp
eseiler: Generally yes, but I think the strong_type should just check if the input is valid

include/seqan3/range/view/kmer_hash.hpp
eseiler: For some reason the concept check fails if I don't provide this overload.

include/seqan3/search/kmer_index/shape.hpp
h-2: Well, it all affects the interface of shape which part of our API, but I guess there is nothing to be done about it now. I would move it to the view.

include/seqan3/range/view/kmer_hash.hpp
h-2: I meant that you should ensure that the iterator and the sentinel of your view should model SizedSentinel. (not checking that on the underlying range)

h-2:
Hmmm.. since we won't have kmer-indexing and kmer-based search in the release anyway, we might as well move this to after the release; and instead focus on getting the fm_index interface straight.

eseiler:
@h-2 I switched to the `dynamic_bitset`. Can you have another look>

include/seqan3/search/kmer_index/shape.hpp
```diff
@@ -0,0 +1,124 @@
...
+ * ### Example
+ *
+ * \include test/snippet/search/kmer_index/shape.cpp
+ *
+ * \attention 0 < size <= 58
+ *
+ * \if DEV
+ * ### Implementation details
+ *
+ * The last position indicates whether the shape is ungapped, i.e. if `back()` returns `true` the shape is ungapped,
```
eseiler: This needs to be removed.

include/seqan3/range/view/kmer_hash.hpp
```diff
@@ -12,47 +12,645 @@
...
+            "The range parameter to view::kmer_hash must model std::ranges::ForwardRange.");
+        static_assert(Semialphabet<reference_t<urng_t>>,
+            "The range parameter to view::kmer_hash must be over elements of seqan3::Semialphabet.");
+
+        if (std::ceil(std::log2(alphabet_size<reference_t<urng_t>>)) * shape_.size() > 64u)
         {
-            std::hash<decltype(in)> h{};
-            return h(in);
-        });
+            throw std::invalid_argument{"The chosen shape/alphabet combination is not valid. "
```
eseiler: I'm not so sure if this is the correct place, but I need to do it somewhere .. or at least `assert()`.

include/seqan3/range/view/kmer_hash.hpp
```diff
@@ -12,47 +12,645 @@
...
+        //!\brief Type for distances between iterators.
+        using difference_type = typename it_t::difference_type;
+        //!\brief Value type of this iterator.
+        using value_type = size_t;
+        //!\brief The pointer type.
+        using pointer = void;
+        //!\brief Reference to `value_type`.
+        using reference = value_type;
+        //!\brief Tag this class depending on which concept it_t models.
+        using iterator_category = std::conditional_t<std::ContiguousIterator<it_t>,
```
h-2:
```suggestion
        using iterator_category = std::input_iterator_tag;
        using iterator_concept = std::conditional_t<std::ContiguousIterator<it_t>,
```

Because it returns values. We need to fix this for a few views.

include/seqan3/range/view/kmer_hash.hpp
```diff
@@ -12,47 +12,645 @@
...
+        using alphabet_t = value_type_t<it_t>;
+
+        //!\brief The alphabet size.
+        static constexpr auto const sigma{alphabet_size<alphabet_t>};
+
+        //!\brief The hash value.
+        size_t hash_value{0};
+
+        //!\brief Pointer to underlying view.
+        kmer_hash_view const * pview{nullptr};
```
h-2: If you store the shape here, you can get rid of the pointer, right? Would be the same size and less indirection.

include/seqan3/range/view/kmer_hash.hpp
```diff
@@ -12,47 +12,645 @@
...
+
+public:
+    /*!\name Associated types
+     * \{
+     */
+    //!\brief The iterator type of this view.
+    using iterator          = shape_iterator<urng_t>;
+    //!\brief The const_iterator type of this view.
+    using const_iterator    = shape_iterator<urng_t const>;
+    //!\}
```
h-2: These don't need to be visible since shape_iterator is private anyway.

include/seqan3/range/view/kmer_hash.hpp
```diff
@@ -12,47 +12,645 @@
...
+    kmer_hash_view & operator=(kmer_hash_view && rhs)      = default; //!< Defaulted.
+    ~kmer_hash_view()                                      = default; //!< Defaulted.
+
+    //!\brief Construct from a view and a given shape.
+    kmer_hash_view(urng_t urange_, shape const & s_) : urange{std::move(urange_)}, shape_{s_}
+    {}
+
+    //!\brief Construct from a non-view that can be view-wrapped and a given shape.
+    template <std::ranges::ViewableRange orng_t>
+    kmer_hash_view(orng_t && urange_, shape const & s_) :
```
h-2: Add some extra checks here that you can actually construct from the view. See other views for reference.

include/seqan3/range/view/kmer_hash.hpp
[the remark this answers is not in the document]
h-2: resolved?

include/seqan3/range/view/kmer_hash.hpp
```diff
@@ -12,47 +12,645 @@
...
-        return std::forward<urng_t>(urange) | ranges::view::sliding(k) | std::view::transform(
-        [] (auto const in)
+        static_assert(std::ranges::ViewableRange<urng_t>,
+            "The range parameter to view::kmer_hash cannot be a temporary of a non-view range.");
+        static_assert(std::ranges::ForwardRange<urng_t>,
+            "The range parameter to view::kmer_hash must model std::ranges::ForwardRange.");
+        static_assert(Semialphabet<reference_t<urng_t>>,
+            "The range parameter to view::kmer_hash must be over elements of seqan3::Semialphabet.");
+
+        if (std::ceil(std::log2(alphabet_size<reference_t<urng_t>>)) * shape_.size() > 64u)
```
h-2: Is 64 the right number?

include/seqan3/range/view/kmer_hash.hpp
h-2: It's ok to throw, the check only happens once. But I would maybe move it into the constructor so guarantee that it always happens.

include/seqan3/range/view/kmer_hash.hpp
```diff
@@ -61,47 +659,56 @@ struct kmer_hash_fn
...
+ *                      See below for the properties of the returned range.
+ * \ingroup view
+ *
+ * \details
+ *
+ * ### View properties
+ *
+ * | range concepts and reference_t  | `urng_t` (underlying range type)   | `rrng_t` (returned range type)   |
+ * |---------------------------------|:----------------------------------:|:--------------------------------:|
+ * | std::ranges::InputRange         | *required*                         | *preserved*                      |
```
h-2: Please add this line at the top:
```
 *  | std::Semiregular                |                                       | *preserved*                            |
 *  |                                 |                                       |                                      |
```

Since not all views are guaranteed to be Semiregular nowadays, we should indicate which are. I will change this later for the existing views.

include/seqan3/range/view/kmer_hash.hpp
eseiler: Yes, the hashes are calculated via arithmetic coding. So I need to ensure that the maximum hash value fits into a `uint64_t`.

**But** actually the correct way is to do `shape_.size() <= (64 / std::log2(alphabet_size<reference_t<urng_t>>))`.

include/seqan3/range/view/kmer_hash.hpp
eseiler: Mhm, I need it for random access:
`text_left = std::ranges::next(std::ranges::begin(*pview), n);`
When I jump with the `shape_iterator` I need to set this one text iterator to the corresponding position...

include/seqan3/range/view/kmer_hash.hpp
```diff
@@ -12,47 +12,647 @@
...
+    };
+
+/*!\name Associated types
+* \{
+*/
+//!\brief The iterator type of this view.
+using iterator          = shape_iterator<urng_t>;
+//!\brief The const_iterator type of this view.
+using const_iterator    = shape_iterator<urng_t const>;
+//!\}
```
h-2: wrong indention

include/seqan3/range/view/kmer_hash.hpp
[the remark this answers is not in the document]
h-2: The place where this is included, reads :

> This is the full proto-adaptor, first look at the second member function: it handles range and argument input and returns a range (in other cases this would delegate to the view's constructor if you have your own view type declared).

This needs to be adapted, or not?

include/seqan3/range/view/kmer_hash.hpp
h-2: Do you use these at all? You have return type auto below...

eseiler:
```cpp
auto begin()
{
  return iterator{...};
}
```

I changed it to
```cpp
iterator begin()
{
  return {...};
}
```

include/seqan3/range/view/kmer_hash.hpp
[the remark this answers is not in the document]
h-2: change to:

> This is the full proto-adaptor, first look at the second member function: it handles range and argument input and delegates to the view's constructor. In other, simpler cases you could invoke other adaptors here.

PR #947, opened 2019-05-03
[FEATURE] Adds ranges::move and ranges::move_backward.

rrahn:


h-2:
The algorithms, also those in `std::ranges::` are part of `<algorithms>` and not `<ranges>`.

rrahn:
That is true. So is copy, for_each and all other algorithms we have placed there before. Also iter_move is there, which is part of iterator and not ranges as well. I will clean it up. But in a second PR, because it will touch some files I guess.

h-2:
> That is true. So is copy, for_each and all other algorithms we have placed there before. Also iter_move is there, which is part of iterator and not ranges as well. I will clean it up. But in a second PR, because it will touch some files I guess.

I agree, but we can move the new stuff there already, or not?

rrahn:
Can we please just merge this now 🙏 
The refactoring can be done in a second PR and contains logical things.

h-2:
Well it's not about refactoring anything, it's just about committing the changes to a different file :man_shrugging:

But whatever, let's not discuss it too long. To what you think is best.

rrahn:
Then I'll merge this now and make the other changes tomorrow. Since no one is using these algorithms yet anyway, the merge conflicts will come from different use cases.

PR #953, opened 2019-05-07
[MISC] Some cleanup of std/ranges and std/iterator header

rrahn:
Moves std::ranges::iter_move to iterator header and also adds some documentation for other iterator access functions. 
Adds std::ranges::iter_swap to iterator header.


rrahn:
@h-2 can you have a quick look as well. It should not take long.

PR #959, opened 2019-05-09
[FIX] Doxygen warnings in Range Module

joergi-w:
This PR improves the SeqAn documentation in the range module and removes all 134 warnings from the doxygen build.

### DIFF
--- a/include/seqan3/range/container/bitcompressed_vector.hpp
+++ b/include/seqan3/range/container/bitcompressed_vector.hpp
@@ -971,33 +971,41 @@ public:
     }
     //!\}
 
-    //!\name Comparison operators
-    //!\{
+    /*!\name Comparison operators
+     * \{
+     */
+
+    //!\brief Checks whether `*this` is equal to `rhs`.
     constexpr bool operator==(bitcompressed_vector const & rhs) const noexcept
     {
         return data == rhs.data;
     }
 
+    //!\brief Checks whether `*this` is not equal to `rhs`.
     constexpr bool operator!=(bitcompressed_vector const & rhs) const noexcept
     {
         return data != rhs.data;
     }
 
+    //!\brief Checks whether `*this` is less than `rhs`.
     constexpr bool operator<(bitcompressed_vector const & rhs) const noexcept
     {
         return data < rhs.data;
     }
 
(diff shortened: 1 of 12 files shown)


include/seqan3/range/decorator/gap_decorator_anchor_set.hpp
```diff
@@ -650,7 +659,10 @@ class gap_decorator_anchor_set
      * \param[in] rhs The right-hand side gap decorator to compare.
      * \returns A boolean flag indicating (in)equality of the aligned sequences.
      *
+     * \details
```
eseiler: I thought everything after a new line is implicitly details?

joergi-w: I thought so as well, but a look at the html output shows that the title line below is not parsed correctly if we omit the \details keyword.

joergi-w:
As discussed I reverted the namespace nesting. This PR is ready for a final review.

include/seqan3/range/container/bitcompressed_vector.hpp
```diff
@@ -971,33 +971,41 @@ class bitcompressed_vector
     }
     //!\}
 
-    //!\name Comparison operators
-    //!\{
+    /*!\name Comparison operators
+     * \{
```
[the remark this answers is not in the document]
joergi-w: Doxygen does not parse this correctly and marks the following as undocumented. Apparently, we need to use block comments for subsequent lines that belong together.

include/seqan3/range/view/take.hpp
```diff
@@ -208,17 +224,20 @@ class view_take : public ranges::view_interface<view_take<urng_t, exactly, or_th
             }
         }
 
+        //!\brief Checks whether `lhs` is equal to `rhs`.
         constexpr friend bool operator==(sentinel_type const & lhs, iterator_type const & rhs) noexcept(noexcept(rhs == lhs))
         {
             return rhs == lhs;
         }
 
+        //!\brief Checks whether `*this` is equal to `rhs`.
```
```suggestion
        //!\brief Checks whether `*this` is not equal to `rhs`.
```

joergi-w: Oops – thanks for spotting that!

PR #961, opened 2019-05-10
[TEST] Fastq Benchmark

Clemapfel:
Benchmark format_file_fastq. Includes seqan2 comparison as well as reading from both stream and disk

### DIFF
--- a/test/performance/io/format_fastq_benchmark.cpp
+++ b/test/performance/io/format_fastq_benchmark.cpp
@@ -32,10 +32,10 @@ unsigned int const SEED = 1234;
 // ============================================================================
 // generate fastq file
 // ============================================================================
-static constexpr std::size_t SEQUENCE_LENGTH = 300; // length of nucleotide and quality sequence
+static constexpr std::size_t SEQUENCE_LENGTH = 50; // length of nucleotide and quality sequence
 static constexpr std::size_t N_ENTRIES_IN_FILE = 4069; // number of 3-line entries
 
-// run-time workaround because test::generate_sequence does not work at compile time
+// run-time execution workaround because test::generate_sequence does not work at compile time
 std::string fastq_file;
 bool has_been_init = false;
 
@@ -121,10 +121,6 @@ void fastq_read_from_stream_seqan2(benchmark::State & state)
 {
     using namespace seqan;
 
-    String<char> id{};
-    String<Dna5> seq{};
-    String<char> qual{};
-
     std::istringstream istream{};
 
     auto restart_iterator = [&istream]()    // c.f. format_fasta_benchmark
@@ -135,13 +131,17 @@ void fastq_read_from_stream_seqan2(benchmark::State & state)
         return directionIterator(comp, Input());
     };
 
+    String<char> id{};
+    String<Dna5> seq{};
+    String<char> qual{};
+
     for (auto _ : state)
     {
         auto it = restart_iterator();
         // should be same overhead as istream(file) in equivalent seqan3 benchmark
         // thus skews benchmark but still makes for a valid comparison between seqan2 and 3
 
-        readRecord(id, seq, it, seqan::Fastq{});
+        readRecord(id, seq, qual, it, seqan::Fastq{});
         clear(id);
         clear(seq);
         clear(qual);
@@ -175,6 +175,7 @@ void fastq_read_from_disc(benchmark::State & state)
     ostream << get_file();
     ostream.close();
 
+    // benchmark
     std::string id{};
     std::vector<dna5> seq{};
     std::vector<phred42> qual{};
@@ -183,16 +184,10 @@ void fastq_read_from_disc(benchmark::State & state)
     {
         sequence_file_input fin{file_name};
 
-        for (auto & [seq_f, id_f, qual_f] : fin)
-        {
-            id = id_f;
-            seq = seq_f;
-            qual = qual_f;
-        }
-
-        id.clear();
-        seq.clear();
-        qual.clear();
+        // read all records and store in internal buffer
+        auto it = fin.begin();
+        while (it != fin.end())
+            ++it;
     }
 
     // remove temporary file
@@ -229,18 +224,19 @@ void fastq_read_from_disc_seqan2(benchmark::State & state)
     ostream << get_file();
     ostream.close();
 
-    String<char> id{};
-    String<Dna5> seq{};
-    String<char> qual{};
+    // benchmark
+    StringSet<String<char>> ids{};
+    StringSet<String<Dna5>> seqs{};
+    StringSet<String<char>> quals{};
 
     for (auto _ : state)
     {
         SeqFileIn seqFileIn(file_name.c_str());
-        readRecord(id, seq, qual, seqFileIn);
+        readRecords(ids, seqs, quals, seqFileIn);
 
-        clear(id);
-        clear(seq);
-        clear(qual);
+        clear(ids);
+        clear(seqs);
+        clear(quals);
     }
 
     // delete
@@ -258,36 +254,6 @@ BENCHMARK(fastq_read_from_disc_seqan2);
 #endif
 
 /*
-// ============================================================================
-// read dummy fastq file ignoring only quality
-// ============================================================================
-void fastq_read_without_quality(benchmark::State & state)
-{
-    auto const fastq_file = generate_dummy_fastq_file(POWER);
-
-    sequence_file_format_fastq format;
-    sequence_file_input_options<dna5, false> const options{};
-
-    std::string id{};
-    std::vector<dna5> seq{};
-
-    for (auto _ : state)
-    {
-        state.PauseTiming();
-        std::istringstream istream{fastq_file};
-        state.ResumeTiming();
-
-        for (size_t i = 0; i < iterations_per_run; ++i)
-        {
-            format.read(istream, options, seq, id, std::ignore);
-            id.clear();
-            seq.clear();
-        }
-    }
-}
-
-BENCHMARK(fastq_read_without_quality);
-
 // ============================================================================
 // read dummy fastq file ignoring all parameters
 // ============================================================================
--- a/test/performance/io/format_fastq_benchmark.cpp
+++ b/test/performance/io/format_fastq_benchmark.cpp
@@ -93,8 +93,7 @@ void fastq_read_from_stream(benchmark::State & state)
     for (auto _ : state)
     {
         std::istringstream istream{file};
-        // refilling streams introduces the same constant to
-        // both seqan3 and 2 benchmark runtime so they are still comparable
+        // same constant as seqan2 benchmark
 
         for (size_t i = 0; i < n_entries_in_file; ++i)
             format.read(istream, options, seq, id, qual);
--- a/test/performance/io/format_fastq_benchmark.cpp
+++ b/test/performance/io/format_fastq_benchmark.cpp
@@ -211,7 +211,6 @@ void fastq_read_from_disk_seqan2(benchmark::State & state)
 
 #endif
 
-// weirdly formatted code but makes console output easier to read
 BENCHMARK(fastq_write_to_stream);
 
 BENCHMARK(fastq_read_from_stream)->Arg(100);
@@ -239,3 +238,5 @@ BENCHMARK(fastq_read_from_disk_seqan2)->Arg(100000);
 #endif
 
 BENCHMARK_MAIN();
+
+
--- a/test/performance/io/format_fastq_benchmark.cpp
+++ b/test/performance/io/format_fastq_benchmark.cpp
@@ -153,14 +153,14 @@ void fastq_read_from_stream_seqan2(benchmark::State & state)
 #endif
 
 // ============================================================================
-// read dummy fastq file from temporary file on disc
+// read dummy fastq file from temporary file on disk
 // ============================================================================
 template<std::size_t n_entries_in_file = N_ENTRIES_IN_FILE>
-void fastq_read_from_disc(benchmark::State & state)
+void fastq_read_from_disk(benchmark::State & state)
 {
     sequence_file_format_fastq format;
 
-    // create temporary file on disc
+    // create temporary file on disk
     std::filesystem::path tmp_dir = std::filesystem::temp_directory_path();
     std::string const file_name = tmp_dir/"tmp.fastq";
 
@@ -169,7 +169,7 @@ void fastq_read_from_disc(benchmark::State & state)
 
     if (!open_success)
     {
-        std::perror("Error creating temporary file \"tmp.fastq\" for fastq_read_from_disc.");
+        std::perror("Error creating temporary file \"tmp.fastq\" for fastq_read_from_disk.");
         std::cout << "aborting..." << std::endl;
         exit(1);
     }
@@ -196,7 +196,7 @@ void fastq_read_from_disc(benchmark::State & state)
     const int remove_success = std::remove(file_name.c_str());
     if (remove_success != 0) // sic
     {
-        std::perror("Error removing temporary file \"tmp.fastq\" for fastq_read_from_disc.");
+        std::perror("Error removing temporary file \"tmp.fastq\" for fastq_read_from_disk.");
         std::cout << "aborting..." << std::endl;
         exit(1);
     }
@@ -208,7 +208,7 @@ void fastq_read_from_disc(benchmark::State & state)
 #if SEQAN2_HAS_SEQAN3
 
 template<std::size_t n_entries_in_file = N_ENTRIES_IN_FILE>
-void fastq_read_from_disc_seqan2(benchmark::State & state)
+void fastq_read_from_disk_seqan2(benchmark::State & state)
 {
     using namespace seqan;
 
@@ -221,7 +221,7 @@ void fastq_read_from_disc_seqan2(benchmark::State & state)
 
     if (!open_success)
     {
-        std::perror("Error creating temporary file \"tmp.fastq\" for fastq_read_from_disc_seqan2.");
+        std::perror("Error creating temporary file \"tmp.fastq\" for fastq_read_from_disk_seqan2.");
         std::cout << "aborting..." << std::endl;
         exit(1);
     }
@@ -248,7 +248,7 @@ void fastq_read_from_disc_seqan2(benchmark::State & state)
     const int remove_success = std::remove(file_name.c_str());
     if (remove_success != 0)
     {
-        std::perror("Error removing temporary file \"tmp.fastq\" for fastq_read_from_disc_seqan2.");
+        std::perror("Error removing temporary file \"tmp.fastq\" for fastq_read_from_disk_seqan2.");
         std::cout << "aborting..." << std::endl;
         exit(1);
     }
@@ -262,23 +262,23 @@ BENCHMARK(fastq_write);
 
     BENCHMARK_TEMPLATE(fastq_read_from_stream, 1000);
     BENCHMARK_TEMPLATE(fastq_read_from_stream_seqan2, 1000);
-    BENCHMARK_TEMPLATE(fastq_read_from_disc, 1000);
-    BENCHMARK_TEMPLATE(fastq_read_from_disc_seqan2, 1000);
+    BENCHMARK_TEMPLATE(fastq_read_from_disk, 1000);
+    BENCHMARK_TEMPLATE(fastq_read_from_disk_seqan2, 1000);
 
     BENCHMARK_TEMPLATE(fastq_read_from_stream, 5000);
     BENCHMARK_TEMPLATE(fastq_read_from_stream_seqan2, 5000);
-    BENCHMARK_TEMPLATE(fastq_read_from_disc, 5000);
-    BENCHMARK_TEMPLATE(fastq_read_from_disc_seqan2, 5000);
+    BENCHMARK_TEMPLATE(fastq_read_from_disk, 5000);
+    BENCHMARK_TEMPLATE(fastq_read_from_disk_seqan2, 5000);
 
     BENCHMARK_TEMPLATE(fastq_read_from_stream, 10000);
     BENCHMARK_TEMPLATE(fastq_read_from_stream_seqan2, 10000);
-    BENCHMARK_TEMPLATE(fastq_read_from_disc, 10000);
-    BENCHMARK_TEMPLATE(fastq_read_from_disc_seqan2, 10000);
+    BENCHMARK_TEMPLATE(fastq_read_from_disk, 10000);
+    BENCHMARK_TEMPLATE(fastq_read_from_disk_seqan2, 10000);
 
 #elif
 
     BENCHMARK_TEMPLATE(fastq_read_from_stream);
-    BENCHMARK_TEMAPLTE(fastq_read_from_disc);
+    BENCHMARK_TEMAPLTE(fastq_read_from_disk);
 
 #endif
 
--- a/test/performance/io/format_fastq_benchmark.cpp
+++ b/test/performance/io/format_fastq_benchmark.cpp
@@ -1,0 +1,160 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
+// -----------------------------------------------------------------------------------------------------
+
+// written for the purpose of the Bch Softwarepraktikum, early 2019 FU Berlin
+// @author: Clemens Cords <clemenscords@fu-berlin.de>
+
+#include <benchmark/benchmark.h>
+#include <cstring>
+#include <cmath>
+#include <fstream>
+#include <seqan3/alphabet/nucleotide/dna5.hpp>
+#include <seqan3/alphabet/quality/phred42.hpp>
+#include <seqan3/io/sequence_file/format_fastq.hpp>
+#include <seqan3/io/sequence_file/input.hpp>
+#include <seqan3/range/view/to_char.hpp>
+
+using namespace seqan3;
+
+std::size_t POWER = 5; //files will have |10^POWER| lines
+
+auto const DNA_SEQ = "AGCTAGCAGCGATCGCGATCGATCAGCGATCGAGGAATATAT"_dna5;
+auto const QUALITY = "IIIIIHIIIIIIIIIIIIIIIIIIIIIIHHGIIIIHHGIIIH"_phred42;
+
+std::size_t const iterations_per_run = 4096;  // arbitrary number directly taken from fasta benchmark
+
+// ============================================================================
+// generate dummy fastq file with 10^power identical 4-line entries
+// ============================================================================
+std::string generate_dummy_fastq_file(std::size_t power)
+{
+    std::string file{};
+
+    std::string const id{"@name"};
+    std::string const seq = DNA_SEQ | view::to_char;
+    std::string const quality = QUALITY | view::to_char;
+
+    auto const n_iterations = static_cast<unsigned int>(pow(10, power));
+
+    for (size_t i = 0; i < n_iterations; ++i)
+    {
+        file += id + '\n' + seq + '\n' + '+' + '\n' + quality + '\n';
+    }
+
+    return file;
+}
+
+// ============================================================================
+// try to write 4-line entry to stream as often as possible
+// ============================================================================
+void fastq_write(benchmark::State & state)
+{
+    std::ostringstream ostream;
+
+    sequence_file_format_fastq format;
+    sequence_file_output_options options{};
+
+    std::string const id{"@name"};
+
+    for (auto _ : state)
+    {
+        for (size_t i = 0; i < iterations_per_run; ++i)
+            format.write(ostream, options, DNA_SEQ, id, QUALITY);
+    }
+}
+
+BENCHMARK(fastq_write);
+
+// ============================================================================
+// read dummy fastq file with all parameters
+// ============================================================================
+void fastq_read_with_quality(benchmark::State & state)
+{
+    auto const fastq_file = generate_dummy_fastq_file(POWER);
+
+    sequence_file_format_fastq format;
+    sequence_file_input_options<dna5, false> const options{};
+
+    std::string id{};
+    std::vector<dna5> seq{};
+    std::vector<phred42> qual{};
+
+    for (auto _ : state)
+    {
+        state.PauseTiming();
+        std::istringstream istream{fastq_file}; // costly so pause/resume is necessary
+        state.ResumeTiming();
+
+        for (size_t i = 0; i < iterations_per_run; ++i)
+        {
+            format.read(istream, options, seq, id, qual);
+            id.clear();
+            seq.clear();
+            qual.clear();
+        }
+    }
+}
+
+BENCHMARK(fastq_read_with_quality);
+
+// ============================================================================
+// read dummy fastq file ignoring only quality
+// ============================================================================
+void fastq_read_without_quality(benchmark::State & state)
+{
+    auto const fastq_file = generate_dummy_fastq_file(POWER);
+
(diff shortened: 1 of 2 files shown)


test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,285 @@
...
+        file_dict[n_entries] = file;
+    }
+
+    return file_dict[n_entries];
+}
+
+// ============================================================================
+// write 3-line entry to stream as often as possible
+// ============================================================================
+void fastq_write(benchmark::State & state)
```
smehringer:
```suggestion
void fastq_write_to_stream(benchmark::State & state)
```

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,285 @@
...
+    std::vector<dna5> seq{};
+    std::vector<phred42> qual{};
+
+    for (auto _ : state)
+    {
+        std::istringstream istream{get_file(n_entries_in_file)};
+        // refilling streams introduces the same constant to
+        // both seqan3 and 2 benchmark runtime so they are still comparable
+
+        format.read(istream, options, seq, id, qual);
```
smehringer: you initialize a whole file but then you read only one record here. we should probably read all the records (for example with a while loop over the stream or by using a sequence_file_input instead like you did for reading from disk.)

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,285 @@
...
+// ============================================================================
+// read dummy fastq file from temporary file on disk
+// ============================================================================
+template<std::size_t n_entries_in_file = N_ENTRIES_IN_FILE>
+void fastq_read_from_disk(benchmark::State & state)
+{
+    sequence_file_format_fastq format;
+
+    // create temporary file on disk
+    std::filesystem::path tmp_dir = std::filesystem::temp_directory_path();
```
smehringer: in `#include <seqan3/test/tmp_filename.hpp>` we have a `tmp_filename filename{"cereal_test"};` that you can use which creates a file in the tempdirectory for you and even removes it afterwards.

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,285 @@
...
+    BENCHMARK_TEMPLATE(fastq_read_from_disk_seqan2, 5000);
+
+    BENCHMARK_TEMPLATE(fastq_read_from_stream, 10000);
+    BENCHMARK_TEMPLATE(fastq_read_from_stream_seqan2, 10000);
+    BENCHMARK_TEMPLATE(fastq_read_from_disk, 10000);
+    BENCHMARK_TEMPLATE(fastq_read_from_disk_seqan2, 10000);
+
+#else
+
+    BENCHMARK_TEMPLATE(fastq_read_from_stream);
```
marehr: Travis performance test:

```
/home/travis/build/seqan/seqan3/test/performance/io/format_fastq_benchmark.cpp:280:46: error: ISO C++11 requires at least one argument for the "..." in a variadic macro [-Werror]
     BENCHMARK_TEMPLATE(fastq_read_from_stream);
                                              ^
```

Can you extract all `seqan3`-only  performance tests and move them out of the `ifdef`

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,285 @@
...
+// @author: Clemens Cords <clemens.cords@fu-berlin.de>
+
+#include <benchmark/benchmark.h>
+#include <cstring>
+#include <fstream>
+
+#define SEQAN2_HAS_SEQAN3 __has_include(<seqan/seq_io.h>)
+#if SEQAN2_HAS_SEQAN3
+    #include <seqan/seq_io.h>
+#endif
```
marehr: We have now a header to check for `seqan2`

```c++
#include <seqan3/test/seqan2.hpp>

// defines macro SEQAN3_HAS_SEQAN2
```

Can you include this header and replace `SEQAN2_HAS_SEQAN3` with `SEQAN3_HAS_SEQAN2`

seqan3/test/seqan2.hpp:
https://github.com/seqan/seqan3/blob/063af063f535098adc9a1bcf75810dd3726c8bb5/test/include/seqan3/test/seqan2.hpp

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,285 @@
...
+#include <seqan3/alphabet/quality/phred42.hpp>
+#include <seqan3/io/sequence_file/format_fastq.hpp>
+#include <seqan3/io/sequence_file/input.hpp>
+#include <seqan3/range/view/to_char.hpp>
+#include <seqan3/test/performance/sequence_generator.hpp>
+#include <seqan3/std/filesystem>
+
+using namespace seqan3;
+
+unsigned int const SEED = 1234;
```
marehr:
```suggestion
unsigned const seed = 1234;
```

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,285 @@
...
+
+// ============================================================================
+// generate fastq file
+// ============================================================================
+constexpr std::size_t SEQUENCE_LENGTH = 50; //length of nucleotide and quality sequence
```
marehr: Please, replace all occurrences of `std::size_t` with `size_t` (without namespace)

We don't declare variables in all upper-case, please adjust it to lower case.

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,285 @@
...
+    for (auto _ : state)
+    {
+        format.write(ostream, options, seq, id, quality);
+    }
+}
+
+// ============================================================================
+// read dummy fastq file from a stream
+// ============================================================================
+template<std::size_t n_entries_in_file = N_ENTRIES_IN_FILE>
```
marehr: Why are you defaulting the template parameter with `N_ENTRIES_IN_FILE` (= 4096) if you call it below with a different number (= 1000).

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,285 @@
...
+    sequence_file_format_fastq format;
+    sequence_file_input_options<dna5, false> const options{};
+
+    std::string id{};
+    std::vector<dna5> seq{};
+    std::vector<phred42> qual{};
+
+    for (auto _ : state)
+    {
+        std::istringstream istream{get_file(n_entries_in_file)};
```
marehr: I would move `get_file(n_entries_in_file)` out of the for-loop. AFAIK you try to benchmark reading from a stream.

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,285 @@
...
+// any functions later on that needs the exact same file can access it
+std::map<std::size_t, std::string> file_dict;
+
+std::string get_file(std::size_t n_entries)
+{
+    bool has_been_init = file_dict.find(n_entries) != file_dict.end();
+
+    if (!has_been_init)
+    {
+        std::cout << std::endl; // sic, helps formatting print output
```
marehr: I don't understand this.

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,285 @@
...
+// generate fastq file
+// ============================================================================
+constexpr std::size_t SEQUENCE_LENGTH = 50; //length of nucleotide and quality sequence
+constexpr std::size_t N_ENTRIES_IN_FILE = 4069; // default number of 3-line entries
+
+// if a file of length m is requested once, that file is stored so
+// any functions later on that needs the exact same file can access it
+std::map<std::size_t, std::string> file_dict;
+
+std::string get_file(std::size_t n_entries)
```
marehr: I"m not sure if this is the way to go. I understand your intention to ensure that every benchmark has the same input, so that the benchmark results are [more] comparable to each other AND to be efficient and don't recompute the same. But, I think in this case it would suffice to write a `generate_fasta_file` function without any memorization/result-caching.

Can you put everything within the `if (!has_been_init)` into a own function `generate_fasta_file`?

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,285 @@
...
+void fastq_write(benchmark::State & state)
+{
+    std::ostringstream ostream;
+
+    sequence_file_format_fastq format;
+    sequence_file_output_options options{};
+
+    std::string const id{"@name"};
+    auto seq = test::generate_sequence<dna5>(SEQUENCE_LENGTH, 0, SEED);
+    auto quality = test::generate_sequence<phred42>(SEQUENCE_LENGTH, 0, SEED);
```
marehr: Below you simply used `qual`; Can you stick to one way of naming?

test/performance/io/format_fastq_benchmark.cpp
Clemapfel: this creates a newline in between benchmark calls with different template parameters like so:
```
fastq_write_to_stream                     16099 ns        16098 ns        44227

fastq_read_from_stream<1000>              90514 ns        90516 ns         7443
fastq_read_from_disk<1000>             84545863 ns     84548582 ns            8
fastq_read_from_stream_seqan2<1000>       15878 ns        15879 ns        43251
fastq_read_from_disk_seqan2<1000>       6828959 ns      6829089 ns           98

fastq_read_from_stream<5000>             124365 ns       124369 ns         5683
fastq_read_from_disk<5000>            387559999 ns    387567259 ns            2
fastq_read_from_stream_seqan2<5000>       51202 ns        51203 ns        13720
fastq_read_from_disk_seqan2<5000>      34893085 ns     34894137 ns           20
```
It is obviously not needed but I find it makes the output nicer to read and compare and does not impact performance or functionality in any way

marehr: Please don't use that.

Furthermore, what happens if you change the output format `--benchmark_format=json`?

test/performance/io/format_fastq_benchmark.cpp
Clemapfel: fixed

test/performance/io/format_fastq_benchmark.cpp
Clemapfel: fixed

(idk if I'm allowed to resolve the conversation or if I'm supposed to not say anything and just push the changes :pensive:)

test/performance/io/format_fastq_benchmark.cpp
Clemapfel: fixed

test/performance/io/format_fastq_benchmark.cpp
Clemapfel: fixed

test/performance/io/format_fastq_benchmark.cpp
Clemapfel: fixed

test/performance/io/format_fastq_benchmark.cpp
Clemapfel: fixed

test/performance/io/format_fastq_benchmark.cpp
Clemapfel: fixed

test/performance/io/format_fastq_benchmark.cpp
Clemapfel: I kept 4069 to make it consistent with format_fasta_benchmark (line 35) but you're right and I removed both that line and default values for the template parameter

test/performance/io/format_fastq_benchmark.cpp
Clemapfel: fixed

test/performance/io/format_fastq_benchmark.cpp
Clemapfel: Do the contents of the file (what permutation of nucleotides / phred-scores) have absolutely no impact on runtime? If so, I'll edit it to just generate a new file every time since in that case you're absolutely right and as long as the files have the same number of chars the benchmarks are comparable.

My rational behind caching it was that I figured at some point during the parsing of the file there will be a moment where the current char c is compared against each introducing a constant depending on the char c like so:
```cpp
// pseudocode
nucleotide c = (...)
switch (c)
{
    case A:
        (...)
        break;

    case C:
        (...)
        break;

    case G:
        (...)
        break;

    case T:
        (...)
        break;
} 
```
Then if ``c == A`` it will introduce a constant of 1 (1 if-comparison), while ``c == T`` will introduce a constant of 4 (4 if-comparisons). For bigger alphabets like phred42 this means the variance can be up to 41 operations per char which would mean for a file with 10^6 lines this could mean at worst a difference of 41 *10^6 operations.

I know being this stingy about runtime in something like a benchmark that isn't really important (and has crazy variance just based on the cpu workload during execution) isn't necessary but I don't see a reason to actively make something less exact.

Please let me know, this is more about me learning how it works and out of principle than how the benchmark turns out. I already put everything in it's own function but kept the memory caching for now

test/performance/io/format_fastq_benchmark.cpp
marehr: Resolving a conversation should only be done by the one who started it.

You can signal that you think that you addressed a point by using the reaction-emoticons. I personally use :heart:, others use :+1:

test/performance/io/format_fastq_benchmark.cpp
h-2: I really would like spacing when printing to the command line actually. Is there another way of doing this?

test/performance/io/format_fastq_benchmark.cpp
marehr: If the value of the characters have an influence on the run time, we should notice some jitter in the results.
----

In this case the compiler is smarter than you :)

`A,C,G,T` are basically a fancy way of writing the numbers `0, 1, 2, 3`.

See  https://godbolt.org/z/M26FVb for an example what the compiler does in those cases.

Spoiler: this is the complete generated assembly; independent on how many cases you have. Tip: try to add the more cases, the assembly will stay at 5 instructions.
```
        mov     eax, DWORD PTR [rsp+12]
        xor     r8d, r8d
        cmp     eax, 3
        ja      .L1
        mov     r8d, DWORD PTR CSWTCH.3[0+rax*4]
```

test/performance/io/format_fastq_benchmark.cpp
marehr: I don't know. For me it sounds like a feature request.

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,256 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
```
marehr:
```suggestion
// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
```

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,256 @@
...
+// -----------------------------------------------------------------------------------------------------
+
+// @author: Clemens Cords <clemens.cords@fu-berlin.de>
```
marehr: We normally don't attribute authorship within test cases, but if you want you should use `\` doxygen style.

```suggestion
//!\author Clemens Cords <clemens.cords@fu-berlin.de>
```

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,256 @@
...
+#include <seqan3/io/sequence_file/format_fastq.hpp>
+#include <seqan3/io/sequence_file/input.hpp>
+#include <seqan3/range/view/to_char.hpp>
+#include <seqan3/std/filesystem>
+#include <seqan3/test/performance/sequence_generator.hpp>
+#include <seqan3/test/tmp_filename.hpp>
+
+using namespace seqan3;
+
+unsigned int const seed = 1234;
```
marehr:
```suggestion
constexpr unsigned seed = 1234u;
```

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,256 @@
...
+#endif
+
+// weirdly formatted code but makes console output easier to read
+BENCHMARK(fastq_write_to_stream);
+
+BENCHMARK_TEMPLATE(fastq_read_from_stream, 100);
+BENCHMARK_TEMPLATE(fastq_read_from_disk, 100);
+
+#if SEQAN3_HAS_SEQAN2
+    BENCHMARK_TEMPLATE(fastq_read_from_stream_seqan2, 100);
```
marehr: We normally don't add a level of indentation within `#if ... #endif` guards.

```suggestion
BENCHMARK_TEMPLATE(fastq_read_from_stream_seqan2, 100);
```

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+// @author: Clemens Cords <clemens.cords@fu-berlin.de>
+
+#include <benchmark/benchmark.h>
+#include <cstring>
+#include <fstream>
+
+#include <seqan3/test/seqan2.hpp>
+#if SEQAN3_HAS_SEQAN2
+    #include <seqan/seq_io.h>
```
marehr: I know other benchmarks indent here, but we normally don't do that. And I think the other benchmarks do it wrong for includes.

```suggestion
#include <seqan/seq_io.h>
```

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,256 @@
...
+        clear(quals);
+    }
+}
+
+#endif
+
+// weirdly formatted code but makes console output easier to read
+BENCHMARK(fastq_write_to_stream);
+
+BENCHMARK_TEMPLATE(fastq_read_from_stream, 100);
```
marehr: Okay so you still use `BENCHMARK_TEMPLATE`.

I think in this case it would suffice to use 
```c++
BENCHMARK(fastq_read_from_stream)->Arg(1000);

// retrieve 1000 within the benchmark with `state.range(0)`
```

See https://github.com/google/benchmark#passing-arguments for more details.

I'm not sure if you can define benchmarks multiple times with different :

```c++
// this works:
BENCHMARK(fastq_read_from_stream)->Arg(1000)->Arg(10000);

// not sure if this works
BENCHMARK(fastq_read_from_stream)->Arg(1000);
// other benchmarks
BENCHMARK(fastq_read_from_stream)->Arg(10000); // not sure if this works
```

test/performance/io/format_fastq_benchmark.cpp
marehr: Did you forget to commit it?; Or do I misunderstand your statement? :thinking: 

Or did you only remove `template<std::size_t n_entries_in_file = N_ENTRIES_IN_FILE>` from this benchmark.

test/performance/io/format_fastq_benchmark.cpp
Clemapfel: I removed the caching. This caused runtime to quadruple (since a new file has to be generated from scratch each template call)

test/performance/io/format_fastq_benchmark.cpp
Clemapfel: I removed all templatization in favor of this method

test/performance/io/format_fastq_benchmark.cpp
Clemapfel: I don't think there isn't another way, to guarantee that the newline is printed in the right place one would need a function like:
```cpp
void newline()
{
    // print newline
    return;
}
#define NEWLINE BENCHMARK(newline);

BENCHMARK(A1)
BENCHMARK(A2)
NEWLINE
BENCHMARK(B1)
BENCHMARK_MAIN()
```
But this isn't possible since the ``BENCHMARK`` macro will only accepts functions with the signature ``void (benchmark::state&)`` afaik. So there isn't a way to do this without modifying google benchmark I think.

Clemapfel: Even this wouldn't work because you can't control how often ``benchmark(...)`` would be invoked

```cpp
// some header
void newline()
{
    // print newline
    return;
}
#define NEWLINE newline();

// benchmark.cpp
void benchmark(benchmark::State & state) 
{
    (...)
    for (auto _ : state)
        (...)
        
    NEWLINE;
}

BENCHMARK(benchmark);
BENCHMARK(other_benchmark_function);
BENCHMARK_MAIN();
```

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,242 @@
...
+#if SEQAN3_HAS_SEQAN2
+void fastq_read_from_stream_seqan2(benchmark::State & state)
+{
+    using namespace seqan;
+
+    std::istringstream istream{};
+    size_t n_entries_in_file = state.range(0);
+    auto file = get_file(n_entries_in_file);
+
+    auto restart_iterator = [&istream, &file]()    // c.f. format_fasta_benchmark
```
marehr: An unsusal abbreviation :) A quick search always resulted in this form `cf.`

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,242 @@
...
+    std::vector<phred42> qual{};
+
+    size_t n_entries_in_file = state.range(0);
+    auto file = get_file(n_entries_in_file);
+
+    for (auto _ : state)
+    {
+        std::istringstream istream{file};
+        // refilling streams introduces the same constant to
+        // both seqan3 and 2 benchmark runtime so they are still comparable
```
marehr: Not sure how I feel about this comment. It it helpful or not? It shows that you thought hard about this, but it seems to be more like a mental note for yourself than a note for everyone.

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+    // create temporary file, automatically removed on destruction
+    test::tmp_filename file_name{"tmp.fastq"};
+    auto tmp_path = file_name.get_path();
+
+    std::ofstream ostream{tmp_path};
+    ostream << get_file(state.range(0));
+    ostream.close();
+
+    // benchmark
```
marehr: Is this comment necessary?

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,242 @@
...
+// ============================================================================
+void fastq_read_from_disk(benchmark::State & state)
+{
+    sequence_file_format_fastq format;
+
+    // create temporary file, automatically removed on destruction
+    test::tmp_filename file_name{"tmp.fastq"};
+    auto tmp_path = file_name.get_path();
+
+    std::ofstream ostream{tmp_path};
```
marehr:
```suggestion
    // fill temporary file with a fastq file
    std::ofstream ostream{tmp_path};
```

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,242 @@
...
+    ostream << get_file(state.range(0));
+    ostream.close();
+
+    // benchmark
+    std::string id{};
+    std::vector<dna5> seq{};
+    std::vector<phred42> qual{};
```
marehr: Hmm aren't these variables unused?

We should get a warnings because of unused variables. :thinking:

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+    std::ofstream ostream{tmp_path};
+    size_t n_entries_in_file = state.range(0);
+    ostream << get_file(n_entries_in_file);
+    ostream.close();
+
+    // benchmark
+    StringSet<String<char>> ids{};
+    StringSet<String<Dna5>> seqs{};
+    StringSet<String<char>> quals{};
```
marehr: unused?

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,242 @@
...
+{
+    using namespace seqan;
+
+    // temporary file
+    test::tmp_filename file_name{"tmp.fastq"};
+    auto tmp_path = file_name.get_path();
+
+    std::ofstream ostream{tmp_path};
+    size_t n_entries_in_file = state.range(0);
+    ostream << get_file(n_entries_in_file);
```
marehr:
```suggestion
    ostream << get_file(state.range(0));
```

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,242 @@
...
+void fastq_read_from_disk_seqan2(benchmark::State & state)
+{
+    using namespace seqan;
+
+    // temporary file
+    test::tmp_filename file_name{"tmp.fastq"};
+    auto tmp_path = file_name.get_path();
+
+    std::ofstream ostream{tmp_path};
+    size_t n_entries_in_file = state.range(0);
```
marehr:
```suggestion

```

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,242 @@
...
+BENCHMARK(fastq_read_from_disk)->Arg(10000);
+
+#if SEQAN3_HAS_SEQAN2
+BENCHMARK(fastq_read_from_stream_seqan2)->Arg(100000);
+BENCHMARK(fastq_read_from_disk_seqan2)->Arg(100000);
+#endif
+
+BENCHMARK_MAIN();
+
+
```
marehr: double newline at the end

```suggestion

```

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,242 @@
...
+{
+    sequence_file_format_fastq format;
+
+    // create temporary file, automatically removed on destruction
+    test::tmp_filename file_name{"tmp.fastq"};
+    auto tmp_path = file_name.get_path();
+
+    std::ofstream ostream{tmp_path};
+    ostream << get_file(state.range(0));
+    ostream.close();
```
marehr: We could also extract this portion into a new function:

```c++

auto create_fastq_file(size_t const entries)
{
    // create temporary file, automatically removed on destruction
    test::tmp_filename fastq_file{"tmp.fastq"};
    auto fastq_file_path = file_name.get_path();

    // fill temporary file with a fastq file
    std::ofstream ostream{fastq_file_path};
    ostream << generate_fastq_file(entries);
    ostream.close();
    return fastq_file;
}

auto fastq_file = create_fastq_file(1000);
fastq_file.get_path(); // this should give the same path as within the function, right?
```

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+std::string get_file(size_t n_entries)
+{
+    std::string file{};
+    std::string const id{"@name"};
+
+    for (size_t i = 0; i < n_entries; ++i)
+    {
+        auto seq = test::generate_sequence<dna5>(default_sequence_length, 0, seed);
+        std::string seq_string = seq | view::to_char;
```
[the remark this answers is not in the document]
smehringer: Yes but we want that?! so our benchmarks are comparable across builds?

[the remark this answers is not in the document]
marehr: If you have a fixed seed at the start of the function and change the seed in a deterministic fashion the whole process will still be deterministic.

```c++
auto seq = test::generate_sequence<dna5>(default_sequence_length, 0, i); // this is deterministic
// but every sequence will be different
```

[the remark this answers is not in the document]
Clemapfel: I thought that for each call of ``generate_sequence`` function the MT rng engine would advance it's state but since it's constructed newly for each call all the sequences are literally the same. I fixed it now

test/performance/io/format_fastq_benchmark.cpp
Clemapfel: I shortened it but I think it's important because it's not immediately obvious to the reader why the creation of the file should be inside the benchmark for loop instead of outside since it introduces a significant constant to runtime

test/performance/io/format_fastq_benchmark.cpp
Clemapfel: Since it's only ever called twice and it would shorten the 5 lines inside the benchmark to 2 I don't feel like it's necessary for now

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+    std::string const id{"@name"};
+    auto seq = test::generate_sequence<dna5>(default_sequence_length, 0, default_seed);
+    auto qual = test::generate_sequence<phred42>(default_sequence_length, 0, default_seed);
+
+    for (auto _ : state)
+    {
+        format.write(ostream, options, seq, id, qual);
+    }
+}
```
marehr: We had some API change and the new calling syntax is as follows:

```c++
    sequence_file_output fout{ostream, format_fastq{}};

    std::string const id{"@name"};
    auto seq = test::generate_sequence<dna5>(default_sequence_length, 0, default_seed);
    auto qual = test::generate_sequence<phred42>(default_sequence_length, 0, default_seed);

    for (auto _ : state)
        fout.emplace_back(seq, id, qual);
}
```

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,234 @@
...
+{
+    sequence_file_format_fastq format;
+    sequence_file_input_options<dna5, false> const options{};
+
+    std::string id{};
+    std::vector<dna5> seq{};
+    std::vector<phred42> qual{};
+
+    size_t entries_size = state.range(0);
+    auto file = generate_fastq_file(entries_size);
```
marehr:
```suggestion
    std::string file = generate_fastq_file(entries_size);
```

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+    for (auto _ : state)
```
marehr: We had some API change and the new calling syntax is as follows:

```c++
    std::istringstream istream{file};
    sequence_file_input fin{istream, format_fastq{}};

    for (auto[seq, id, qual] : fin)
    {}
}
```

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,234 @@
...
+    ostream.close();
+
+    for (auto _ : state)
+    {
+        sequence_file_input fin{tmp_path};
+
+        // read all records and store in internal buffer
+        auto it = fin.begin();
+        while (it != fin.end())
+            ++it;
```
marehr: Please change it here, too.

test/performance/io/format_fastq_benchmark.cpp
marehr: This has nothing to do with shorter code. This has something to do with readability and others that will copy this file as a blueprint that will make later changes a nightmare.

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+    for (size_t i = 0; i < entries_size; ++i, ++seed)
+    {
+        auto seq = test::generate_sequence<dna5>(default_sequence_length, 0, seed);
+        auto seq_as_chars = seq | views::to_char;
+
+        auto qual = test::generate_sequence<phred42>(default_sequence_length, 0, seed);
+        auto qual_as_chars = qual | views::to_char;
+
+        file << id << '\n' << std::string(seq_as_chars.begin(), seq_as_chars.end()) << '\n' << '+' << '\n' << std::string(qual_as_chars.begin(), qual_as_chars.end()) << '\n';
```
Clemapfel: views::to_char threw that implicit conversion from views to containers is now deprecated, I couldn't think of a better way to append and convert the ranges to std::string

PR #962, opened 2019-05-10
[FEATURE?] Only one build per branch concurrently

jpfeuffer:
Hopefully fixes daily crashes of the Docker server.

Alternative is to cancel the old running job:
https://stackoverflow.com/questions/46041579/jenkins-pipeline-milestone-not-cancelling-previous-ongoing-build/48956042#48956042

### DIFF
--- a/.jenkinsfile
+++ b/.jenkinsfile
@@ -4,9 +4,11 @@ import hudson.triggers.TimerTrigger;
 // Global properties
 // ============================================================================
 
+// Only one build at a time for each branch/PR
 // Start build every night between 1:00-4:59
 // Keep only last 10 builds
 properties([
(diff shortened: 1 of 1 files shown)


rrahn:
But this makes the parallel stage obsolete now?

jpfeuffer:
I don't know. I would hope that this only restricts the "uber" job. But I never used pipelines.

jpfeuffer:
But I think you need something like this. The workspace is "per branch" right (it uses JOB_NAME)? What if multiple job numbers access the same workspace?

rrahn:
Mhmm, I thought he would wait anyway for the previous builds to be finished before he starts the new build. But probably not.

PR #965, opened 2019-05-13
[FEATURE] add forwards for edit_distance

marehr:
Only a small PR that extracts some declarations / definitions out of a file. This is needed, because a next PR will introduce a score and trace matrix class, which need a forward declaration of  the edit distance algorithm.

### DIFF
--- a/include/seqan3/alignment/pairwise/edit_distance_fwd.hpp
+++ b/include/seqan3/alignment/pairwise/edit_distance_fwd.hpp
@@ -1,0 +1,68 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Forwards for seqan3::edit_distance_unbanded related types.
+ * \author Marcel Ehrhardt <marcel.ehrhardt AT fu-berlin.de>
+*/
+
+#pragma once
+
+#include <seqan3/core/platform.hpp>
+#include <seqan3/std/concepts>
+#include <seqan3/std/ranges>
+
+namespace seqan3::detail
+{
+/*!\todo Document me
+ * \ingroup pairwise_alignment
+ */
+template <typename traits_type>
+SEQAN3_CONCEPT EditDistanceTrait = requires
+{
+    typename std::remove_reference_t<traits_type>::word_type;
+    typename std::remove_reference_t<traits_type>::is_semi_global_type;
+
+    // Must be a boolean integral constant.
+    requires std::Same<typename std::remove_reference_t<traits_type>::is_semi_global_type::value_type, bool>;
+};
+
+/*!\brief The default traits type for the edit distance algorithm.
+ * \ingroup pairwise_alignment
+ */
+struct default_edit_distance_trait_type
+{
+    //!\brief The default word type.
+    using word_type = uint64_t;
+    //!\brief Semi global alignment is disabled by default.
+    using is_semi_global_type = std::false_type;
+};
+
+//!\brief Store no state for state_t.
+template <typename state_t>
+struct empty_state
+{};
+
+//!\brief If enabled is true state_t will be added to state_type.
+template <bool enabled, typename state_t>
+using enable_state_t = std::conditional_t<enabled, state_t, empty_state<state_t>>;
+
+//!\cond
+template <std::ranges::ViewableRange database_t,
+          std::ranges::ViewableRange query_t,
+          typename align_config_t,
+          EditDistanceTrait traits_t = default_edit_distance_trait_type>
+class pairwise_alignment_edit_distance_unbanded; //forward declaration
+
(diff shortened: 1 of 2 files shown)


marehr:
I renamed it to  `..._fwd` to follow the naming of the standard `iosfwd` and `range_fwd`.

include/seqan3/alignment/pairwise/edit_distance_fwd.hpp
```diff
@@ -0,0 +1,66 @@
...
+using enable_state_t = std::conditional_t<enabled, state_t, empty_state<state_t>>;
+
+template <std::ranges::ViewableRange database_t,
+          std::ranges::ViewableRange query_t,
+          typename align_config_t,
+          EditDistanceTrait traits_t = default_edit_distance_trait_type>
+class pairwise_alignment_edit_distance_unbanded; //forward declaration
+
+template <typename word_t, typename score_t, bool is_semi_global, bool use_max_errors>
+class edit_distance_score_matrix_full; //forward declaration
```
rrahn: Is this ok with doxygen 15? Maybe put them in a cond block to avoid possible future warnings?

include/seqan3/alignment/pairwise/edit_distance_fwd.hpp
```diff
@@ -0,0 +1,66 @@
...
+ */
+struct default_edit_distance_trait_type
+{
+    //!\brief The default word type.
+    using word_type = uint64_t;
+    //!\brief Semi global alignment is disabled by default.
+    using is_semi_global_type = std::false_type;
+};
+
+//!\brief Store no state for state_t.
```
rrahn: so state_t is itself a state but empty_state depends on it?

rrahn: Sorry it is just that I can't see the usage and I am wondering if you could also in this case just add an alias:
```cpp
template <typename state_t>
using empty_state = seqan3::detail::deferred_type<state_t>;
```

marehr: You can use it like this:

```c++
struct state1
{
  int var1;
  // a lot of member variables // costly to copy
};
struct state2
{
  int var2;
  // a lot of member variables // costly to copy
};

template <bool enable1, bool enable2>
class my_state : detail::enable_state_t<enable1, state1>, detail::enable_state_t<enable1, state1>
{
};

my_state<false, false>; // completely empty ; sizeof(my_state<false, false>) == 1
my_state<true, false>; // inherits state1; sizeof(my_state<false, false>) == sizeof(state1)
my_state<false, true>; // inherits state2; sizeof(my_state<false, false>) == sizeof(state2)
my_state<true, true>; // inherits state1 and state2 ; sizeof(my_state<false, false>) == sizeof(state1) + sizeof(state2)
```
So the main idea is to conditionally include states into a big state.

> Sorry it is just that I can't see the usage and I am wondering if you could also in this case just add an alias:
> 
> ```c++
> template <typename state_t>
> using empty_state = seqan3::detail::deferred_type<state_t>;
> ```

Yeah, I could also use `std::identity` to remove empty_state completely.

include/seqan3/alignment/pairwise/edit_distance_fwd.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+#pragma once
+
+#include <seqan3/core/platform.hpp>
+#include <seqan3/std/concepts>
+#include <seqan3/std/ranges>
+
+namespace seqan3::detail
+{
+/*!\todo Document me
```
smehringer: there is still a todo :eyes:

eseiler: Yes, I also saw this, but it's copy/paste ... so I don't think it's necessarily within the scope of this PR (though it would be nice)

include/seqan3/alignment/pairwise/edit_distance_fwd.hpp
rrahn: No, in general I am not against an empty base class that states a clear meaning, but from your example I don't understand why empty_state needs to have a template parameter? Why are you depending on the state that you explicitly want to not use in the empty_state?

include/seqan3/alignment/pairwise/edit_distance_fwd.hpp
rrahn: In general, if this is going to be part of the release we need to have it documented, although it is developer API, but I wish to have a release with no todo in the documentation. If you can promise this in a related PR to this, which will follow soon, then we can leave it.

marehr: @rrahn This todo is from you :laughing:

include/seqan3/alignment/pairwise/edit_distance_fwd.hpp
marehr: Simply, because of multiple inheritance.

It is not allowed to inherit the same class twice.
```c++
class empty_state{};

class state : empty_state, empty_state // not allowed
{};

template <typename>
class empty_state2{};

class state2 : empty_state2<long>, empty_state2<int> // allowed, different base classes
{};
```

include/seqan3/alignment/pairwise/edit_distance_fwd.hpp
rrahn: pah, everyone could claim this 😏

include/seqan3/alignment/pairwise/edit_distance_fwd.hpp
rrahn: ok

PR #966, opened 2019-05-13
[FEATURE] make trace_directions debug_streamable

marehr:


include/seqan3/alignment/matrix/trace_directions.hpp
```diff
@@ -47,3 +47,21 @@ namespace seqan3
 template <>
 constexpr bool add_enum_bitwise_operators<seqan3::detail::trace_directions> = true;
 } // namespace seqan3
+
+#include <seqan3/io/stream/debug_stream.hpp>
```
smehringer: move include to the top?

include/seqan3/alignment/matrix/trace_directions.hpp
```diff
@@ -47,3 +47,21 @@ namespace seqan3
...
+
+inline debug_stream_type & operator<<(debug_stream_type & s, detail::trace_directions trace)
+{
+    static char const * unicode[8]{"↺","↖","↑","↖↑","←","↖←","↑←","↖↑←"};
+    static char const * csv[8]{"N","D","U","DU","L","DL","UL","DUL"};
+
+    bool is_unicode = (s.flags2() & fmtflags2::utf8) == fmtflags2::utf8;
+    char const ** trace_dir = is_unicode ? unicode : csv;
+
+    s << trace_dir[(size_t)(trace) % 8u];
```
smehringer:
```suggestion
    s << trace_dir[static_cast<size_t>(trace) % 8u];
```

test/unit/alignment/matrix/trace_directions_test.cpp
```diff
@@ -0,0 +1,40 @@
...
+static constexpr trace_directions U = trace_directions::up;
+static constexpr trace_directions L = trace_directions::left;
+
+TEST(debug_stream, ascii)
+{
+    std::stringstream s{};
+    debug_stream_type stream{s};
+    stream << N << ";" << D << ";" << U << ";" << L << ";" << (D|U) << ";" << (D|L) << ";" << (U|L) << ";" << (D|U|L);
+
+    EXPECT_EQ(s.str(), "N;D;U;L;DU;DL;UL;DUL");
```
smehringer: don't you need to flush the stream before using str()? I know we do that in other places to ensure that it works correctly I think

include/seqan3/alignment/matrix/trace_directions.hpp
marehr: This is currently not working, because of cyclic inclusion.

`trace_direction` includes `debug_stream` includes `alignment_matrix_formatter` includes `trace_direction`

A follow up PR, where I remove `alignment_matrix_formatter` allows me to fix this problem.

test/unit/alignment/matrix/trace_directions_test.cpp
marehr: Why should I need to flush? I tried to read up on this, I found a couple of "unofficial"  statements http://www.cplusplus.com/forum/general/84962/, http://forums.codeguru.com/showthread.php?292719-How-to-clear-out-a-stringstream&p=938707#post938707 and so on.

I mean it makes sense, you don't have to commit you intermediate data to some device. Which flushing is intended for.

include/seqan3/alignment/matrix/trace_directions.hpp
smehringer: I feel like there are on average two follow ups for each PR you do :see_no_evil: I know small PRs are easier to review but this way it is hard to keep track of what you want to achieve on the long run.

include/seqan3/alignment/matrix/trace_directions.hpp
```diff
@@ -47,3 +47,26 @@ namespace seqan3
...
+#include <seqan3/io/stream/debug_stream.hpp>
+namespace seqan3
+{
+
+/*!\brief All trace_directions can be printed as ascii or as utf8 to the seqan3::debug_stream.
+ * \param s The seqan3::debug_stream.
+ * \param trace The trace direction.
+ * \relates seqan3::debug_stream_type
+ */
+inline debug_stream_type & operator<<(debug_stream_type & s, detail::trace_directions trace)
```
rrahn:
```suggestion
inline debug_stream_type & operator<<(debug_stream_type & s, detail::trace_directions const trace)
```

include/seqan3/alignment/matrix/trace_directions.hpp
```diff
@@ -47,3 +47,26 @@ namespace seqan3
...
+{
+    static char const * unicode[8]{"↺","↖","↑","↖↑","←","↖←","↑←","↖↑←"};
```
rrahn: With c++20 char8_t is used for utf-8 characters. Would that not better here as well?

include/seqan3/alignment/matrix/trace_directions.hpp
rrahn:
```suggestion
inline debug_stream_type & operator<<(debug_stream_type & s, detail::trace_directions const trace)
```

include/seqan3/alignment/matrix/trace_directions.hpp
marehr: Do we have this somewhere defined? I didn't know.

rrahn: this knowledge is provided by your always up-to-date online documentation system cppreference: https://en.cppreference.com/w/cpp/language/character_literal

h-2:
Guys, you killed over 30% of our nightlies on C++20 with this PR. I specifically removed all the `char8_t` in January as a fix, now your are introducing it back in :man_facepalming: 

Can you please do the following:

1. change the `char8_t` to `char` again
2. find out why this change breaks stuff like the dna4_test which should be unrelated
3. fix the alignment related stuff in the c++20 nightlies so I can add gcc9+std2a to travis

Especially the last thing is important so we don't start introducing more breakage like yesterday in the next two weeks causing more work for everyone in the end.

PR #971, opened 2019-05-14
[TEST] Restructure sam test to a test template.

smehringer:
This will help reduce code once I add the BAM tests


h-2:
What's up with the coverage again?

smehringer:
I think it is not updating itself... when I click on it the link, it does not show any misses.

smehringer:
@h-2 The code coverage only shows the miss because i did not update on the most recent master with the fix. Can this be merged then?

PR #972, opened 2019-05-14
[FIX] make view::take consume also at the very beginning.

### LINKED ISSUE #950: [BUG] in view_take_until_and_consume
If you add the following test to `take_line` it fails:
```cpp
TEST(view_take_line, eol_at_first_position)
{
    using sbt = std::istreambuf_iterator<char>;
    std::istringstream vec{"\nfoo"};
    auto stream_view = std::ranges::subrange<decltype(sbt{vec}), decltype(sbt{})> {sbt{vec}, sbt{}};
    EXPECT_EQ("", std::string(stream_view | view::take_line));
    EXPECT_EQ("foo", std::string(stream_view | view::take_line));
}

```

The problem is, that the iterator does not consume upon comparison, only upon incrementing which lets an input stream with take_line run into an infinite loop when the very first character is a character that evaluates `fun` to `true`.

I swa no elegant solution at first glance.

smehringer:
Resolves #950

### DIFF
--- a/include/seqan3/range/view/take_until.hpp
+++ b/include/seqan3/range/view/take_until.hpp
@@ -235,17 +235,18 @@ private:
         //!\brief Defaulted.
         ~iterator_type_consume_input()                                                             = default;
 
-        //!\brief Constructor that delegates to the CRTP layer.
-        iterator_type_consume_input(base_base_t it) noexcept(noexcept(base_t{it})) :
-            base_t{std::move(it)}
-        {}
-
         //!\brief Constructor that delegates to the CRTP layer and initialises the callable.
         iterator_type_consume_input(base_base_t it,
                                     fun_ref_t _fun,
                                     sentinel_type sen) noexcept(noexcept(base_t{it})) :
             base_t{std::move(it)}, fun{_fun}, stored_end{std::move(sen)}
-        {}
+        {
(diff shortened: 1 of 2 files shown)


include/seqan3/range/view/take_until.hpp
```diff
@@ -235,17 +235,18 @@ class view_take_until : public ranges::view_interface<view_take_until<urng_t, fu
...
                                     sentinel_type sen) noexcept(noexcept(base_t{it})) :
             base_t{std::move(it)}, fun{_fun}, stored_end{std::move(sen)}
-        {}
+        {
+            while (fun(**this))
+            {
+                at_end_gracefully = true;
+                base_t::operator++();
+            }
+        }
```
h-2: This invokes UB on multiple ocasions, because you never check at end. I would recommend doing
```cpp
if ((*this->this_to_base() != stored_end) && fun(**this))
    ++(*this);
```
The increment operator will take care of the rest.

smehringer: this is not completely correct because you skip the first character that evaluated to true and `at_end_gracefully` was not set :o

smehringer: any preference?

h-2: either is fine

PR #973, opened 2019-05-14
[FEATURE] Add host to little endian conversion.

rrahn:
Adds endian conversion helper.

### DIFF
--- a/include/seqan3/core/bit_manipulation.hpp
+++ b/include/seqan3/core/bit_manipulation.hpp
@@ -18,6 +18,15 @@
 #include <climits>
 #include <utility>
 
+// Find correct header for byte-order conversion functions.
+#if __has_include(<endian.h>) // unix GLIBC
+    #include <endian.h>
+#elif __has_include(<libkern/OSByteOrder.h>)  // APPLE
+    #include <libkern/OSByteOrder.h>
+#elif __has_include(<sys/endian.h>)  // *BSD
+    #include <sys/endian.h>
+#endif // __has_include(endian.h)
+
 #include <seqan3/core/detail/int_types.hpp>
 #include <seqan3/std/concepts>
 
@@ -94,6 +103,50 @@ constexpr uint8_t bit_scan_reverse(unsigned_t n)
     for (; n != 0; n >>= 1, ++i);
     return i - 1;
 #endif
+}
+
+/*!\brief Convert the byte encoding of integer values to little-endian byte order.
+ * \ingroup core
+ * \tparam type The type of the value to convert; must model std::Integral.
+ * \param  in   The input value to convert.
+ * \returns the converted value in little-endian byte-order.
+ *
+ * \details
+ *
+ * This function swaps the bytes if the host system uses big endian and if the size of the integral types is not
+ * bigger than 8 byte. Otherwise, the function does nothing and returns the unchanged input value.
+ */
+template <std::Integral type>
+inline constexpr type enforce_little_endian(type const in) noexcept
+{
+    if constexpr (sizeof(type) == 2)      // 16 bit integral type
+    {
+        #if __has_include(<libkern/OSByteOrder.h>)  // APPLE
+            return OSSwapHostToLittleInt16(in);
+        #else
+            return htole16(in);
+        #endif // __has_include(<libkern/OSByteOrder.h>)
+    }
+    else if constexpr (sizeof(type) == 4) // 32 bit integral type
+    {
+        #if __has_include(<libkern/OSByteOrder.h>)  // APPLE
+            return OSSwapHostToLittleInt32(in);
+        #else
+            return htole32(in);
+        #endif // __has_include(<libkern/OSByteOrder.h>)
+    }
+    else if constexpr (sizeof(type) == 8) // 64 bit integral type
+    {
+        #if __has_include(<libkern/OSByteOrder.h>)  // APPLE
+            return OSSwapHostToLittleInt64(in);
+        #else
+            return htole64(in);
+        #endif // __has_include(<libkern/OSByteOrder.h>)
+    }
+    else
+    {
+        return in;
+    }
 }
 
 } // namespace seqan3::detail
--- a/include/seqan3/core/bit_manipulation.hpp
+++ b/include/seqan3/core/bit_manipulation.hpp
@@ -21,12 +21,11 @@
 // Find correct header for byte-order conversion functions.
 #if __has_include(<endian.h>) // unix GLIBC
     #include <endian.h>
-#elif __has_include(<libkern/OSByteOrder.h>)  // APPLE
-    #include <libkern/OSByteOrder.h>
 #elif __has_include(<sys/endian.h>)  // *BSD
     #include <sys/endian.h>
 #endif // __has_include(endian.h)
 
+#include <seqan3/core/detail/endian.hpp>
 #include <seqan3/core/detail/int_types.hpp>
 #include <seqan3/std/concepts>
 
@@ -113,39 +112,37 @@ constexpr uint8_t bit_scan_reverse(unsigned_t n)
  *
  * \details
  *
- * This function swaps the bytes if the host system uses big endian and if the size of the integral types is not
- * bigger than 8 byte. Otherwise, the function does nothing and returns the unchanged input value.
+ * This function swaps the bytes if the host system uses big endian. In this case only 1, 2, 4, or 8 byte big
+ * integral types are allowed as input. On host systems with little endian this function is a no-op and returns the
+ * unchanged input value. Other systems with mixed endianness are not supported.
  */
 template <std::Integral type>
-inline constexpr type enforce_little_endian(type const in) noexcept
+constexpr type to_little_endian(type const in) noexcept
 {
-    if constexpr (sizeof(type) == 2)      // 16 bit integral type
+    if constexpr (endian::native == endian::little)
     {
-        #if __has_include(<libkern/OSByteOrder.h>)  // APPLE
-            return OSSwapHostToLittleInt16(in);
-        #else
+        return in;
+    }
+    else if constexpr (endian::native == endian::big)
+    {
+        static_assert(sizeof(type) <= 8,
+                      "Can only convert the byte encoding for integral numbers with a size of up to 8 bytes.");
+        static_assert(is_power_of_two(sizeof(type)),
+                      "Can only convert the byte encoding for integral numbers whose byte size is a power of two.");
+
+        if constexpr (sizeof(type) == 2)
             return htole16(in);
-        #endif // __has_include(<libkern/OSByteOrder.h>)
-    }
-    else if constexpr (sizeof(type) == 4) // 32 bit integral type
-    {
-        #if __has_include(<libkern/OSByteOrder.h>)  // APPLE
-            return OSSwapHostToLittleInt32(in);
-        #else
+        else if constexpr (sizeof(type) == 4)
             return htole32(in);
-        #endif // __has_include(<libkern/OSByteOrder.h>)
-    }
-    else if constexpr (sizeof(type) == 8) // 64 bit integral type
-    {
-        #if __has_include(<libkern/OSByteOrder.h>)  // APPLE
-            return OSSwapHostToLittleInt64(in);
-        #else
+        else if constexpr (sizeof(type) == 8)
             return htole64(in);
-        #endif // __has_include(<libkern/OSByteOrder.h>)
+        else
+            return in;  // single byte.
     }
     else
     {
-        return in;
+        static_assert(endian::native == endian::little || endian::native == endian::big,
+                      "Expected a little-endian or big-endian platform.");
     }
 }
 
--- a/test/unit/core/bit_manipulation_test.cpp
+++ b/test/unit/core/bit_manipulation_test.cpp
@@ -108,27 +108,27 @@ TYPED_TEST(unsigned_operations, bit_scan_reverse)
     }
 }
 
-TEST(enforce_little_endian, byte)
+TEST(to_little_endian, byte)
 {
     uint8_t val = 0x01;
 
-    EXPECT_EQ(enforce_little_endian(val), 0x01);
+    EXPECT_EQ(to_little_endian(val), 0x01);
 }
 
-TEST(enforce_little_endian, word)
+TEST(to_little_endian, word)
 {
     uint16_t val = 0x0102;  // 258
-    uint16_t res = enforce_little_endian(val);
+    uint16_t res = to_little_endian(val);
     char * res_p = reinterpret_cast<char *>(&res);
 
     EXPECT_EQ(*(res_p + 0), '\x02'); // LSB
     EXPECT_EQ(*(res_p + 1), '\x01'); // MSB
 }
 
-TEST(enforce_little_endian, double_word)
+TEST(to_little_endian, double_word)
 {
     uint32_t val = 0x01020304; // 16.909.060
-    uint32_t res = enforce_little_endian(val);
+    uint32_t res = to_little_endian(val);
     char * res_p = reinterpret_cast<char *>(&res);
 
     EXPECT_EQ(*(res_p + 0), '\x04');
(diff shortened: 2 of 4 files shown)


include/seqan3/core/bit_manipulation.hpp
```diff
@@ -96,4 +109,28 @@ constexpr uint8_t bit_scan_reverse(unsigned_t n)
...
+ * \details
+ *
+ * This function swaps the bytes if the host system uses big endian and if the size of the integral types is not
+ * bigger than 8 byte. Otherwise, the function does nothing and returns the unchanged input value.
+ */
+template <std::Integral type>
+inline constexpr type enforce_little_endian(type const in) noexcept
+{
+    if constexpr (sizeof(type) == 2)      // 16 bit integral type
+        return htole16(in);
```
marehr:
```suggestion
#ifdef defined(__APPLE__)
        return OSSwapHostToLittleInt16(x)
#else
        return htole16(in);
#endif
```

include/seqan3/core/bit_manipulation.hpp
```diff
@@ -18,6 +18,19 @@
...
 
+// BYTE-ORDER DETECTION (default is little-endian)
+#ifdef __GLIBC__
+    #include <endian.h>
+#elif defined(__APPLE__)
+    #include <libkern/OSByteOrder.h>
+    // Provide alias for macos system.
+    #define htole16(x) OSSwapHostToLittleInt16(x)
+    #define htole32(x) OSSwapHostToLittleInt32(x)
+    #define htole64(x) OSSwapHostToLittleInt64(x)
```
marehr: please don't expose those

test/unit/core/bit_manipulation_test.cpp
```diff
@@ -107,3 +107,48 @@ TYPED_TEST(unsigned_operations, bit_scan_reverse)
         }
     }
 }
+
+TEST(byte_order_transformation, host_to_little_byte)
```
marehr:
```suggestion
TEST(host_to_little, byte)
```

test/unit/core/bit_manipulation_test.cpp
```diff
@@ -107,3 +107,48 @@ TYPED_TEST(unsigned_operations, bit_scan_reverse)
...
+{
+    uint8_t val = 0x01;
+
+    EXPECT_EQ(enforce_little_endian(val), 0x01);
+}
+
+TEST(byte_order_transformation, host_to_little_word)
```
marehr:
```suggestion
TEST(host_to_little, word)
```

test/unit/core/bit_manipulation_test.cpp
```diff
@@ -107,3 +107,48 @@ TYPED_TEST(unsigned_operations, bit_scan_reverse)
...
+{
+    uint16_t val = 0x0102;  // 258
+    uint16_t res = enforce_little_endian(val);
+    char * res_p = reinterpret_cast<char *>(&res);
+
+    EXPECT_EQ(*(res_p + 0), '\x02'); // LSB
+    EXPECT_EQ(*(res_p + 1), '\x01'); // MSB
+}
+
+TEST(byte_order_transformation, host_to_little_double_word)
```
marehr:
```suggestion
TEST(host_to_little, double_word)
```

test/unit/core/bit_manipulation_test.cpp
```diff
@@ -107,3 +107,48 @@ TYPED_TEST(unsigned_operations, bit_scan_reverse)
...
+    uint32_t res = enforce_little_endian(val);
+    char * res_p = reinterpret_cast<char *>(&res);
+
+    EXPECT_EQ(*(res_p + 0), '\x04');
+    EXPECT_EQ(*(res_p + 1), '\x03');
+    EXPECT_EQ(*(res_p + 2), '\x02');
+    EXPECT_EQ(*(res_p + 3), '\x01');
+}
+
+TEST(byte_order_transformation, host_to_little_quad_word)
```
marehr:
```suggestion
TEST(host_to_little, quad_word)
```

include/seqan3/core/bit_manipulation.hpp
```diff
@@ -96,4 +105,48 @@ constexpr uint8_t bit_scan_reverse(unsigned_t n)
...
+/*!\brief Convert the byte encoding of integer values to little-endian byte order.
+ * \ingroup core
+ * \tparam type The type of the value to convert; must model std::Integral.
+ * \param  in   The input value to convert.
+ * \returns the converted value in little-endian byte-order.
+ *
+ * \details
+ *
+ * This function swaps the bytes if the host system uses big endian and if the size of the integral types is not
+ * bigger than 8 byte. Otherwise, the function does nothing and returns the unchanged input value.
```
smehringer: What about 1 byte integers? 
If this function cannot work on 1 byte or >8 byte integers than maybe it should be constrained to those instead of doing nothing?

And what about signed integers?

rrahn: Well, a 1 byte does not have a different byte-order on little endian vs big endian 😏. 
Perhaps "not bigger than 8 byte" is misleading, since the function is already constrained on integral types anyway. I'll remove the comment

rrahn:
> Did you test this manually on a big endian machine?

No unfortunately we don't have those. Do you have one?
But I confirmed that the functions on linux as well as on mac detect the host endianness correctly and therefore did nothing. We will see if it works, when Debian runs SeqAn3 tests on their test infrastructure. I think I remember that they also had a sparc and powerpc in their test infrastructure. For all other main target platforms nothing happens anyway.

include/seqan3/core/bit_manipulation.hpp
smehringer: > Well, a 1 byte does not have a different byte-order on little endian vs big endian

Oh, I though little/big endian is about the **bit** order not the byte order.. 

> Perhaps "not bigger than 8 byte" is misleading, since the function is already constrained on integral types anyway.

does an GCC's __int128 not model std::Integral?

And what about signed integers?

rrahn: > does an GCC's __int128 not model std::Integral?

Ok, I did more reading. In general, there is more to it. Integral does not protect against possibly larger types for example `implementation-defined extended integer types`. In addition, the standard does only guarantee a minimum width of the standard integer types. Thus, there could be systems with ints large than 8 byte or 12 and 48 bit width integers (older processors). 
But, it is very unlikely that we will ever run on such systems.
I will constraint it to is_integer which only holds for a predefined set of types.

> And what about signed integers?

It works for signed and unsigned. The system merely interprets the same number in a different byte-order, which does has no affect on the value of the represented bits.

h-2:
> No unfortunately we don't have those. Do you have one?

It can be emulated with e.g. QEMU on Linux: https://create.stephan-brumme.com/big-endian/

But I don't know if we want to spend time on that now.

rrahn:
> It can be emulated with e.g. QEMU on Linux: https://create.stephan-brumme.com/big-endian/

yes, I know, but I won't do it now unless someone has it already there.

> Regarding macOS: doesn't it have machine/endian.h with the same commands as the other unix platforms? Also there are no more big-endian Macs except iPhones

It does, which uses sys/endian.h which defines only the byte-order functions (ntohl etc.) but not the non standard functions we use here. 

>Also there are no more big-endian Macs except iPhones

True so far, but who said, we won't target iPhones/iPads 😏. Ok, joke aside. If I don't provide an overload for apple architecture, obviously the code fails. I did try to use the std::endian enum to check whether or not the host is little_endian but in agreement with marcel, we left this decision better made by the corresponding functions/compiler since that's what they for. Now I could do it like in SeqAn2 to offer an identity function if we detect a little endian system and otherwise have the implementation as above. But in my opinion the code wasn't beautiful before and neither is it now, but I believe it is a little safer as it can now deal with such rare cases. (Who knows if someone plans to port macos on PPC again). But yeah, if you think this is to much and it should be like before, I don't really care. 

>I wrote this code in SeqAn2 and I chose to not use value semantics, but pass-by-reference to make sure that the call is definitely optimised away on little endian platforms:
https://github.com/seqan/seqan/blob/master/include/seqan/platform.h#L500
Why did you decide not to follow this approach?

🤣  Right, because it is always a good idea to outsmart the compiler. It follows in general our coding guide line to prefer value semantics. In many places the code get's easier because you can omit creating a temporary before calling the function and since the htoleXX resolve to __uintXX_identity on little endian, the entire call is optimised away anyway. If you can prove me wrong with solid benchmarks I am happy to change this.

include/seqan3/core/bit_manipulation.hpp
```diff
@@ -96,4 +105,48 @@ constexpr uint8_t bit_scan_reverse(unsigned_t n)
...
+ */
+template <std::Integral type>
+inline constexpr type enforce_little_endian(type const in) noexcept
```
h-2: you don't need inline here

h-2:
> If you can prove me wrong with solid benchmarks I am happy to change this.

Normally the one who introduces a different approach from the one used before is required to convince the others with a benchmark :point_up: 

> Ok, joke aside. If I don't provide an overload for apple architecture, obviously the code fails. 

Yep, and we will need to add one for Windows, too. That's why I originally thought it would be best to have one generic fallback and only overload the rest for platforms where we need to change and where we have respective functions.

> But in my opinion the code wasn't beautiful before and neither is it now,

We agree on this.

> Now I could do it like in SeqAn2 to offer an identity function if we detect a little endian system and otherwise have the implementation as above. 
> Right, because it is always a good idea to outsmart the compiler. 

Have a NOP-function vs an identity function is not "out-smarting" the compiler, because it's `constexpr void {}` which is pretty much guaranteed to resolve to no operation. In fact we could even make it `consteval` on C++20 with a macro which guarantees that it resolves to NOP.

Always calling htole is not guaranteed to be NOP, it's true that I would expect it, but for example it isn't constexpr (but maybe that's not crucial):
```
<source>:6:12: error: call to non-'constexpr' function '__uint32_t __uint32_identity(__uint32_t)'

    6 |     return htole32(v);
```

If I would implement this now, I would not a rely on macros, I would just make a small custom constexpr check for big-endianness (via union trick or something like that) and then sth like this: 

```cpp
constexpr type make_little_endian(type const in) noexcept
{
    if constexpr (!is_big_endian) // already little endian
    {
         return in;
    }
    else if constexpr (__has_include(<endian.h>) ||  __has_include(<sys/endian.h>) // Linux / *BSD
    {
        static_assert(sizeof(type) == 2 ||  sizeof(type) == 4 || sizeof(type) == 8,
                      "Converting endianess for the type of integer not supported.");
        if constexpr (sizeof(type) == 2)        // 16 bit integral type
            return htole16(in);
        else if constexpr (sizeof(type) == 4)   // 32 bit integral type
            return htole32(in);
        else                                    // 64 bit integral type
            return htole64(in);
    }
    else
    {
        static_assert(!is_big_endian, "Big-Endianness is not supported on your platform");
    }
}
```
(this is by value, by reference/NOP would be very similiar)

This implementation guarantees that nothing happens on little endian and gives clear error messages if we do have big endian but no correct handling (your implementation returns identity in cases where it should produce an error). Handling of other platforms *can* be added by adding another if block and doesn't involve multiple line edits of different cases.

Anyway, up to you! The only thing that I think is quite important:
  * if you take by copy and return a value, please call the function `make_`... and not `enforce_` (the first creates something new, the second changes something existing).

include/seqan3/core/bit_manipulation.hpp
rrahn: Well, I know, but I wasn't sure, what we decided on? If we always put inline or not (even if the function is declared constexpr).

rrahn:
>Normally the one who introduces a different approach from the one used before is required to convince the others with a benchmark ☝️

And then you said you did it on purpose to optimise it out. Have you made benchmarks for this? 
But it is nuts to discuss this now. 

Anyway, I agree, that windows needs its own check. I did not research on that. It seems windows only supports the byteorder functions but not the unix ones we use so that would get ugly with windows support. I first had a solution that used std::endian to avoid the call if host is little endian. But @marehr argued against it, since the functions we call ensure this already. I agreed on this advice but now see that at least for windows some custom exception needs to be made. Since I also do not believe that someone is trying to use SeqAn on its old macintosh PPC I would bring back the std::endian version. 
But before I do it note, that there is no test feature macro and the enum is not a template class. So I can check for the cplusplus version but it is possible that it needs some protection in some rare cases, when the cplusplus version does not guarantee that std::endian is already present or not. But I think this won't happen to often.
Agreed?

h-2:
>  But it is nuts to discuss this now.

Yep, like I said: I don't insist on a pass-by-reference version.

> So I can check for the cplusplus version but it is possible that it needs some protection in some rare cases, when the cplusplus version does not guarantee that std::endian is already present or not. But I think this won't happen to often.

I am pretty sure that it will break immediately. My experience with new library features is that they may not be escaped at all or that they are escaped by being marked as  `> 201703` (which should be >= 2020XX but then it wouldn't work on current `-std=c++2a` which currently likely defines something like 201901). → There is no way for us to know based on which __cplusplus value it will be provided by a standard library.

rrahn:
Then we can only offer our own version in seqan3 namespace, and say that it is deprecated for >= c++20.

h-2:
> Then we can only offer our own version in seqan3 namespace, and say that it is deprecated for >= c++20.

Or just put it in detail? I don't think it will actually ever be used often...

rrahn:
I added it already in detail.

rrahn:
I added it already in detail.

include/seqan3/core/bit_manipulation.hpp
```diff
@@ -96,4 +104,47 @@ constexpr uint8_t bit_scan_reverse(unsigned_t n)
...
+ * unchanged input value. Other systems with mixed endianness are not supported.
+ */
+template <std::Integral type>
+constexpr type to_little_endian(type const in) noexcept
+{
+    if constexpr (seqan3::detail::endian::native == seqan3::detail::endian::little)
+    {
+        return in;
+    }
+    if constexpr (seqan3::detail::endian::native == seqan3::detail::endian::big)
```
h-2:
```suggestion
    else if constexpr (seqan3::detail::endian::native == seqan3::detail::endian::big)
```

include/seqan3/core/bit_manipulation.hpp
```diff
@@ -96,4 +104,47 @@ constexpr uint8_t bit_scan_reverse(unsigned_t n)
...
+    {
+        static_assert(sizeof(type) <= 8,
+                      "Can only convert the byte encoding for integral numbers with a size of up to 8 bytes.");
+        static_assert(is_power_of_two(sizeof(type)),
+                      "Can only convert the byte encoding for integral numbers whose byte size is a power of two.");
+
```
h-2: I would add the following to make life easier for those people who do try to port this to iPhones :wink: 
```cpp
static_assert(__has_include(<endian.h>) || __has_include(<sys/endian.h>),
              "No byte-order conversion available on your platform.")
```

include/seqan3/core/bit_manipulation.hpp
```diff
@@ -96,4 +104,47 @@ constexpr uint8_t bit_scan_reverse(unsigned_t n)
...
+        else if constexpr (sizeof(type) == 8)
+            return htole64(in);
+        else
+            return in;  // single byte.
+    }
+    else
+    {
+        static_assert(seqan3::detail::endian::native == seqan3::detail::endian::little ||
+                      seqan3::detail::endian::native == seqan3::detail::endian::big,
+                     "Expected a little-endian or big-endian platform.");
```
h-2:
```suggestion
                      "Expected a little-endian or big-endian platform.");
```

include/seqan3/core/bit_manipulation.hpp
rrahn: Dosn't work

rrahn: And I could hack it around, but this is not worth the time.

include/seqan3/core/bit_manipulation.hpp
h-2: no, no inline for constexpr functions (we do put it in front of global variables, but it makes a difference there).

rrahn:
@h-2 I merge this then?

include/seqan3/core/bit_manipulation.hpp
```diff
@@ -96,4 +104,47 @@ constexpr uint8_t bit_scan_reverse(unsigned_t n)
...
+            return htole32(in);
+        else if constexpr (sizeof(type) == 8)
+            return htole64(in);
+        else
+            return in;  // single byte.
+    }
+    else
+    {
+        static_assert(seqan3::detail::endian::native == seqan3::detail::endian::little ||
+                      seqan3::detail::endian::native == seqan3::detail::endian::big,
```
h-2: can't you omit the `seqan3::detail::` since this is in detail?

Sorry, just saw this. You can merge after this is fixed.

PR #975, opened 2019-05-14
[FEATURE] Adds magic_header to disambiguate compression streams.

rrahn:
This adds magic header variables for our compression streams. The main reason why we need this is that we need to access the bgzf header at multiple places. Hence, it is safer to bring it into a global variable. Since it would be weird to have this only for bgzf_compression but not the others, I decided to add it for all compression formats.

### DIFF
--- a/include/seqan3/io/detail/magic_header.hpp
+++ b/include/seqan3/io/detail/magic_header.hpp
@@ -82,7 +82,7 @@ struct bgzf_compression
  * Specialises seqan3::detail::magic_header for seqan3::detail::bgzf_compression.
  */
 template <>
-inline constexpr std::array<int8_t, 18 > magic_header<bgzf_compression>
+inline constexpr std::array<int8_t, 18> magic_header<bgzf_compression>
 {
 //  ID1                              ID2                              CM
     magic_header<gz_compression>[0], magic_header<gz_compression>[1], magic_header<gz_compression>[2],
--- a/include/seqan3/io/detail/magic_header.hpp
+++ b/include/seqan3/io/detail/magic_header.hpp
@@ -1,0 +1,95 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides seqan3::detail::magic_header.
+ * \author Rene Rahn <rene.rahn AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <array>
+
+#include <seqan3/core/platform.hpp>
+
+namespace seqan3::detail
+{
+
+//!\brief Defines a magic byte sequence to disambiguate different compression formats. Default is empty.
+//!\ingroup io
+template <typename header_tag_t>
+inline constexpr std::array<int8_t, 0> magic_header{};
+
+//!\brief A tag signifying a gz compressed file.
+//!\ingroup io
+struct gz_compression
+{};
+
+/*!\brief The magic byte sequence to disambiguate gz compressed files.
+ * \ingroup io
+ *
+ * \details
+ *
+ * Specialises seqan3::detail::magic_header for seqan3::detail::gz_compression.
+ */
+template <>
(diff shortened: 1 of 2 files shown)


include/seqan3/io/detail/magic_header.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+/*!\brief The magic byte sequence to disambiguate bgzf compressed files.
+ * \ingroup io
+ *
+ * \details
+ *
+ * Specialises seqan3::detail::magic_header for seqan3::detail::bgzf_compression.
+ */
+template <>
+inline constexpr std::array<int8_t, 18 > magic_header<bgzf_compression>
```
marehr:
```suggestion
inline constexpr std::array<int8_t, 18> magic_header<bgzf_compression>
```

include/seqan3/io/detail/magic_header.hpp
```diff
@@ -0,0 +1,95 @@
...
+{
+//  ID1                              ID2                              CM
+    magic_header<gz_compression>[0], magic_header<gz_compression>[1], magic_header<gz_compression>[2],
+//  FLG     [MTIME                       ]  XFL     OS      [XLEN        ]
+    '\x04', '\x00', '\x00', '\x00', '\x00', '\x00', '\xff', '\x06', '\x00',
+//  B       C       [SLEN        ]  [BSIZE       ]
+    '\x42', '\x43', '\x02', '\x00', '\x00', '\x00'
```
marehr: Why are the optional `[]` not all zero?

For example `[SLEN        ]` is `'\x02', '\x00'`; Why not `'\x00', '\x00'`?

rrahn: Because they are fixed by the format specification.

include/seqan3/io/detail/misc_input.hpp
```diff
@@ -97,7 +98,7 @@ inline auto make_secondary_istream(std::basic_istream<char_t> & primary_stream,
     std::string const extension = filename.extension().string();
 
     // set return value appropriately
-    if (starts_with(magic_number, std::array{'\x1f', '\x8b', '\x08'})) // GZIP
```
h-2: A few lines above the header is extracted, but only 4 bytes. You need to decide and hardcode how much you want to extract, but you need to make sure that you put everything back and also take into account that being at end before you read to the end might be valid, e.g. you may be at end, before you read the end of a bgzf magic header, but does that mean it is invalid or not compressed? What would
```
> A
C
```
(a valid fasta file) look compressed in Gzip? Is that smaller than the bgzf magic header?

include/seqan3/io/detail/magic_header.hpp
```diff
@@ -0,0 +1,95 @@
...
+{
+
+//!\brief Defines a magic byte sequence to disambiguate different compression formats. Default is empty.
+//!\ingroup io
+template <typename header_tag_t>
+inline constexpr std::array<int8_t, 0> magic_header{};
+
+//!\brief A tag signifying a gz compressed file.
+//!\ingroup io
+struct gz_compression
```
h-2: Is this the best place for the tags?

include/seqan3/io/detail/misc_input.hpp
rrahn: I know. But I haven‘t changed anything in the code except replacing the hardcoded byte arrays with the proper magic_header. BGZF is not added here yet. 
Regarding the end of file, the Bgzf header is always written no matter how small the input data. It then follows the compressed block in this case the fasta file followed by a Bgzf block footer and finally (after the last Bgzf block) the end of file marker. I guess in a valid gzip or bzip compressed file you will always have a valid header. Of course for small input sizes the compressed file could be bigger than the plain text files, but I guess this is of no concern.

include/seqan3/io/detail/magic_header.hpp
rrahn: Do you have a better idea? I don‘t care much. But given that we use it right now only in the combination with the magic_header this seems to be a valid first place. If it turns out to be useful in other contexts as well we can move it to some other place.

include/seqan3/io/detail/misc_input.hpp
h-2: > I know. But I haven‘t changed anything in the code except replacing the hardcoded byte arrays with the proper magic_header. BGZF is not added here yet.

Ah, ok. Well it's something to easily forget, because it might work, even if you don't change it. Please make a big TODO (in code or in your mind :wink: ) to not forget this!

> Regarding the end of file, the Bgzf header is always written no matter how small the input data. 

I wasn't referring to this. Anyway, let's discuss it when the code is actually changed.

rrahn: Ah, I now understand what you mean. To test between gzip or bgzf you might be at the end before the magic header ends. Yes, this is important and taken care of.

PR #976, opened 2019-05-14
[FIX] update score in edit distance correctly

marehr:


marehr:
> LGTM, but is there a test that tests the behaviour?
> If there was something wrong before, there should be a test case that tests it.

True and those test cases will come in a later PR. (Before that I need the forwards PR, score and trace matrix PR and maybe a couple of other things )

I try to make the PR's as small as possible to reduce the review time.

smehringer:
> I try to make the PR's as small as possible to reduce the review time.

I really like this :) but I also think that PR's should be self-contained and I agree with @eseiler that a fix should include a test for the corrected behaviour. Otherwise it is hard to track all the follow up issues.

rrahn:
> I really like this :) but I also think that PR's should be self-contained and I agree with @eseiler that a fix should include a test for the corrected behaviour. Otherwise it is hard to track all the follow up issues.

@smehringer are you explicitly requiring it to be added before this PR is merged? Sorry that I have to ask, but I can't fully understand your intention. I understand both sides, but I want a clear communication here.
On the other hand, @marehr would it be to complicated to have a test for this added here, or are the tests part of larger changes and this is just a small separate change? I know that you have been refactoring a lot for the edit distance and just want to make sure, that this won't over complicate things. 

In general for the future we should keep those things together. Also if the PR touches multiple files, but if they are split into multiple commits, one can still see the history of changes.

smehringer:
>  are you explicitly requiring it to be added before this PR is merged? 

well, I talked to @marehr and he said it is not easy to add tests at this point so I guess it is fine.

PR #977, opened 2019-05-14
[FIX] prev_last_active_cell does not need special attention on score_mask

marehr:
It turns out that we don't need this case.

### DIFF
--- a/include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
+++ b/include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
@@ -320,20 +320,8 @@ private:
             return true;
 
         if constexpr (is_global)
-        {
-            if (last_block == 0)  // [[unlikely]]
-            {
-                // Note: This is a special case where we hit before the first query character (i.e. score_mask == 0 &&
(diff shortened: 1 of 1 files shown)


marehr:
> LGTM, but is this a consequence of another PR?
> When we introduced this change, it actually fixed something IIRC.

Yes and no. I figured out that the change fixed a bug, but then later discovered that the bug was in a completely different part. The `advance_score` call turned out to be wrong, because in some situations in/decreased the actual score to a value which was wrong.

smehringer:
> Yes and no. I figured out that the change fixed a bug, but then later discovered that the bug was in a completely different part. The advance_score call turned out to be wrong, because in some situations in/decreased the actual score to a value which was wrong.

Can you maybe reference this fix you mention within this PR's commit message?

rrahn:
>Can you maybe reference this fix you mention within this PR's commit message?

I guess, this is more like an internal thing not reported as a bug or related to any ticket except some general implement edit distance algorithm. In general I am also in favor of having an issue for every PR but in the current state of the project I find it not very practical.

rrahn:
But of course @marehr if you have a ticket for this please add it but I guess this can be merged anyway.

PR #980, opened 2019-05-15
Feature: Adds latches and reader writer synchonisation for concurrent queues.

rrahn:
Implements a latch based on the last [c++ proposal](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0666r2.pdf) and also adds a reader_writer synchronisation point for concurrent data structures that are closable. 
These data structures help to synchronise concurrent code with concurrent data structures such as our buffer_queue in a safe way. Before (SeqAn2), it was all handled internally by the queue. 
These data structures are used for the bgzf_istream parallelisation. 

I put the code at the moment into contrib but I already started a verbose documentation of the classes. We should a) wait for the next c++ meeting to see if latches/barriers will make it into the c++ standard or not. Based on this we can either put some parts in the std module or we need to think about how much we want to expose these data structures as user interfaces, which I personally would prefer in the long term.

### DIFF
--- a/include/seqan3/contrib/parallel/buffer_queue.hpp
+++ b/include/seqan3/contrib/parallel/buffer_queue.hpp
@@ -63,23 +63,11 @@ public:
     // Default constructor sets capacity to 1 (still empty)
     buffer_queue() : buffer_queue{0u}
     {}
-    buffer_queue(buffer_queue const &) = delete;
-    buffer_queue(buffer_queue &&) = default;
+    buffer_queue(buffer_queue const &)             = delete;
+    buffer_queue(buffer_queue &&)                  = default;
     buffer_queue & operator=(buffer_queue const &) = delete;
-    buffer_queue & operator=(buffer_queue &&) = default;
-    ~buffer_queue()
-    {
-        // barrier
-        close();  // close the queue.
-
-        // wait until queue is empty (all readers are finished).
-        spin_delay delay{};
-        while (!is_empty())
-            delay.wait();
-
-        assert(tailPos == tailWritePos);
-        assert(headPos == headReadPos);
-    }
+    buffer_queue & operator=(buffer_queue &&)      = default;
+    ~buffer_queue()                                = default;
 
     // you can set the initial capacity here
     explicit buffer_queue(size_type const init_capacity)
@@ -99,7 +87,7 @@ public:
      * \{
      */
     template <typename value2_t>
-        requires std::Same<std::remove_reference_t<value2_t>, value_t>
+        requires std::ConvertibleTo<value2_t, value_t>
     void push(value2_t && value)
     {
         spin_delay delay{};
@@ -119,7 +107,7 @@ public:
     } // throws if closed
 
     template <typename value2_t>
-        requires std::Same<std::remove_reference_t<value2_t>, value_t>
+        requires std::ConvertibleTo<value2_t, value_t>
     queue_op_status wait_push(value2_t && value)
     {
         spin_delay delay{};
@@ -181,7 +169,7 @@ public:
      * \{
      */
     template <typename value2_t>
-        requires std::Same<std::remove_reference_t<value2_t>, value_t>
+        requires std::ConvertibleTo<value2_t, value_t>
     queue_op_status try_push(value2_t &&);
 
     queue_op_status try_pop(value_t &);
@@ -246,17 +234,20 @@ private:
     }
 
     template <typename value2_t>
-        requires (std::Same<std::remove_reference_t<value2_t>, value_type>) &&
+        requires (std::ConvertibleTo<value2_t, value_t>) &&
                  (buffer_policy == buffer_queue_policy::fixed)
-    bool overflow(value2_t &&);
+    bool overflow(value2_t &&)
+    {
+        return false;
+    }
 
     template <typename value2_t>
-        requires (std::Same<std::remove_reference_t<value2_t>, value_type>) &&
+        requires (std::ConvertibleTo<value2_t, value_t>) &&
                  (buffer_policy == buffer_queue_policy::dynamic)
     bool overflow(value2_t && value);
 
+    //!\brief The ring buffer.
     buffer_t data;
-
     alignas(std::hardware_destructive_interference_size) std::shared_mutex mutable mutex;
     alignas(std::hardware_destructive_interference_size) std::atomic<size_type>    headPos{0};
     alignas(std::hardware_destructive_interference_size) std::atomic<size_type>    headReadPos{0};
@@ -282,20 +273,9 @@ using dynamic_buffer_queue = buffer_queue<value_t, buffer_t, buffer_queue_policy
 // Functions
 // ============================================================================
 
-template <std::Semiregular value_t, SequenceContainer buffer_t, buffer_queue_policy buffer_policy>
+template <typename value_t, typename buffer_t, buffer_queue_policy buffer_policy>
 template <typename value2_t>
-    requires (std::Same<std::remove_reference_t<value2_t>,
-                        typename buffer_queue<value_t, buffer_t, buffer_policy>::value_type>) &&
-             (buffer_policy == buffer_queue_policy::fixed)
-inline bool buffer_queue<value_t, buffer_t, buffer_policy>::overflow(value2_t &&)
-{
-    return false;
-}
-
-template <std::Semiregular value_t, SequenceContainer buffer_t, buffer_queue_policy buffer_policy>
-template <typename value2_t>
-    requires (std::Same<std::remove_reference_t<value2_t>,
-                        typename buffer_queue<value_t, buffer_t, buffer_policy>::value_type>) &&
+    requires (std::ConvertibleTo<value2_t, value_t>) &&
              (buffer_policy == buffer_queue_policy::dynamic)
 inline bool buffer_queue<value_t, buffer_t, buffer_policy>::overflow(value2_t && value)
 {
@@ -380,7 +360,7 @@ inline bool buffer_queue<value_t, buffer_t, buffer_policy>::overflow(value2_t &&
 // currently filled    [tail, tailWrite)
 // currently removed   [head, headRead)
 
-template <std::Semiregular value_t, SequenceContainer buffer_t, buffer_queue_policy buffer_policy>
+template <typename value_t, typename buffer_t, buffer_queue_policy buffer_policy>
 inline queue_op_status buffer_queue<value_t, buffer_t, buffer_policy>::try_pop(value_t & result)
 {
     // try to extract a value
@@ -410,7 +390,7 @@ inline queue_op_status buffer_queue<value_t, buffer_t, buffer_policy>::try_pop(v
 
         newHeadReadPos = cyclic_increment(headReadPos, cap, roundSize);
 
-        if (this->headReadPos.compare_exchange_weak(headReadPos, newHeadReadPos))
+        if (this->headReadPos.compare_exchange_weak(headReadPos, newHeadReadPos, std::memory_order_acq_rel))
             break;
 
         spinDelay.wait();
@@ -423,7 +403,7 @@ inline queue_op_status buffer_queue<value_t, buffer_t, buffer_policy>::try_pop(v
     {
         spin_delay delay{};
         size_type old = headReadPos;
-        while (!this->headPos.compare_exchange_weak(old, newHeadReadPos))
+        while (!this->headPos.compare_exchange_weak(old, newHeadReadPos, std::memory_order_acq_rel))
         {
             old = headReadPos;
             delay.wait();  // add adapting delay in case of high contention.
@@ -457,9 +437,9 @@ inline queue_op_status buffer_queue<value_t, buffer_t, buffer_policy>::try_pop(v
  *                            Default is @link ParallelismTags#Parallel @endlink.
  */
 //
-template <std::Semiregular value_t, SequenceContainer buffer_t, buffer_queue_policy buffer_policy>
+template <typename value_t, typename buffer_t, buffer_queue_policy buffer_policy>
 template <typename value2_t>
-    requires std::Same<std::remove_reference_t<value2_t>, value_t>
+    requires std::ConvertibleTo<value2_t, value_t>
 inline queue_op_status buffer_queue<value_t, buffer_t, buffer_policy>::try_push(value2_t && value)
 {
     // try to push the value
@@ -486,9 +466,7 @@ inline queue_op_status buffer_queue<value_t, buffer_t, buffer_policy>::try_push(
             if (newTailWritePos >= headPos + roundSize)
                 break;
 
-            if (this->tailWritePos.compare_exchange_weak(tailWritePos, newTailWritePos,
-                                                         std::memory_order_release,
-                                                         std::memory_order_relaxed))
+            if (this->tailWritePos.compare_exchange_weak(tailWritePos, newTailWritePos, std::memory_order_acq_rel))
             {
                 auto it = std::ranges::begin(data) + (tailWritePos & (roundSize - 1));
                 *it = std::forward<value2_t>(value);
@@ -498,7 +476,7 @@ inline queue_op_status buffer_queue<value_t, buffer_t, buffer_policy>::try_push(
                 {
                     spin_delay delay{};
                     size_type old = tailWritePos;
-                    while (!this->tailPos.compare_exchange_weak(old, newTailWritePos))
+                    while (!this->tailPos.compare_exchange_weak(old, newTailWritePos, std::memory_order_acq_rel))
                     {
                         old = tailWritePos;
                         delay.wait();  // add adapting delay in case of high contention.
--- a/include/seqan3/contrib/parallel/buffer_queue.hpp
+++ b/include/seqan3/contrib/parallel/buffer_queue.hpp
@@ -20,7 +20,7 @@
 #include <type_traits>
 #include <vector>
 
-#include <seqan3/contrib/parallel/spin_delay.hpp>
+#include <seqan3/core/parallel/detail/spin_delay.hpp>
 #include <seqan3/core/metafunction/range.hpp>
 #include <seqan3/range/container/concept.hpp>
 #include <seqan3/std/algorithm>
@@ -90,7 +90,7 @@ public:
         requires std::ConvertibleTo<value2_t, value_t>
     void push(value2_t && value)
     {
-        spin_delay delay{};
+        detail::spin_delay delay{};
 
         for (;;)
         {
@@ -110,7 +110,7 @@ public:
         requires std::ConvertibleTo<value2_t, value_t>
     queue_op_status wait_push(value2_t && value)
     {
-        spin_delay delay{};
+        detail::spin_delay delay{};
 
         for (;;)
         {
@@ -127,7 +127,7 @@ public:
 
     value_type value_pop() // throws if closed
     {
-        spin_delay delay{};
+        detail::spin_delay delay{};
 
         value_type value;
         for (;;)
@@ -147,7 +147,7 @@ public:
 
     queue_op_status wait_pop(value_type & value)
     {
-        spin_delay delay{};
+        detail::spin_delay delay{};
 
         queue_op_status status;
         for (;;)
@@ -370,7 +370,7 @@ inline queue_op_status buffer_queue<value_t, buffer_t, buffer_policy>::try_pop(v
     size_type roundSize = this->roundSize;
     size_type headReadPos;
     size_type newHeadReadPos;
-    spin_delay spinDelay;
+    detail::spin_delay spinDelay;
 
     // wait for queue to become filled
     while (true)
@@ -401,7 +401,7 @@ inline queue_op_status buffer_queue<value_t, buffer_t, buffer_policy>::try_pop(v
 
     // wait for pending previous reads and synchronize headPos to headReadPos
     {
-        spin_delay delay{};
+        detail::spin_delay delay{};
         size_type old = headReadPos;
         while (!this->headPos.compare_exchange_weak(old, newHeadReadPos, std::memory_order_acq_rel))
         {
@@ -444,7 +444,7 @@ inline queue_op_status buffer_queue<value_t, buffer_t, buffer_policy>::try_push(
 {
     // try to push the value
     {
-        spin_delay delay{};
+        detail::spin_delay delay{};
 
         std::shared_lock read_lock(mutex);
 
@@ -474,7 +474,7 @@ inline queue_op_status buffer_queue<value_t, buffer_t, buffer_policy>::try_push(
 
                 // wait for pending previous writes and synchronise tailPos to tailWritePos
                 {
-                    spin_delay delay{};
+                    detail::spin_delay delay{};
                     size_type old = tailWritePos;
                     while (!this->tailPos.compare_exchange_weak(old, newTailWritePos, std::memory_order_acq_rel))
                     {
--- a/include/seqan3/contrib/parallel/buffer_queue.hpp
+++ b/include/seqan3/contrib/parallel/buffer_queue.hpp
@@ -248,7 +248,7 @@ private:
 
     //!\brief The ring buffer.
     buffer_t data;
-    alignas(std::hardware_destructive_interference_size) std::shared_mutex mutable mutex;
+    alignas(std::hardware_destructive_interference_size) std::shared_mutex mutable mutex{};
     alignas(std::hardware_destructive_interference_size) std::atomic<size_type>    headPos{0};
     alignas(std::hardware_destructive_interference_size) std::atomic<size_type>    headReadPos{0};
     alignas(std::hardware_destructive_interference_size) std::atomic<size_type>    tailPos{0};
@@ -390,7 +390,7 @@ inline queue_op_status buffer_queue<value_t, buffer_t, buffer_policy>::try_pop(v
 
         newHeadReadPos = cyclic_increment(headReadPos, cap, roundSize);
 
-        if (this->headReadPos.compare_exchange_weak(headReadPos, newHeadReadPos, std::memory_order_acq_rel))
+        if (this->headReadPos.compare_exchange_weak(headReadPos, newHeadReadPos))
             break;
 
         spinDelay.wait();
@@ -403,7 +403,7 @@ inline queue_op_status buffer_queue<value_t, buffer_t, buffer_policy>::try_pop(v
     {
         detail::spin_delay delay{};
         size_type old = headReadPos;
-        while (!this->headPos.compare_exchange_weak(old, newHeadReadPos, std::memory_order_acq_rel))
+        while (!this->headPos.compare_exchange_weak(old, newHeadReadPos))
         {
             old = headReadPos;
             delay.wait();  // add adapting delay in case of high contention.
@@ -466,7 +466,7 @@ inline queue_op_status buffer_queue<value_t, buffer_t, buffer_policy>::try_push(
             if (newTailWritePos >= headPos + roundSize)
                 break;
 
-            if (this->tailWritePos.compare_exchange_weak(tailWritePos, newTailWritePos, std::memory_order_acq_rel))
+            if (this->tailWritePos.compare_exchange_weak(tailWritePos, newTailWritePos))
             {
                 auto it = std::ranges::begin(data) + (tailWritePos & (roundSize - 1));
                 *it = std::forward<value2_t>(value);
@@ -476,7 +476,7 @@ inline queue_op_status buffer_queue<value_t, buffer_t, buffer_policy>::try_push(
                 {
                     detail::spin_delay delay{};
                     size_type old = tailWritePos;
-                    while (!this->tailPos.compare_exchange_weak(old, newTailWritePos, std::memory_order_acq_rel))
+                    while (!this->tailPos.compare_exchange_weak(old, newTailWritePos))
                     {
                         old = tailWritePos;
                         delay.wait();  // add adapting delay in case of high contention.
--- a/include/seqan3/contrib/parallel/latch.hpp
+++ b/include/seqan3/contrib/parallel/latch.hpp
@@ -1,0 +1,164 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides seqan3::contrib::latch.
+ * \author Rene Rahn <rene.rahn AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <atomic>
+#include <cassert>
+
+#include <seqan3/contrib/parallel/spin_delay.hpp>
+#include <seqan3/std/new>
+
+namespace seqan3::contrib
+{
+//!\cond
+/*!\brief A single-use synchronisation point to coordinate concurrent threads.
+ *
+ * \details
+ *
+ * A latch is a thread coordination mechanism that allows any number of threads to block until an expected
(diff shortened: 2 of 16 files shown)


marehr:
Why the assignment? Do you want a review?

rrahn:
yes of course, I just added on the wrong section

include/seqan3/contrib/parallel/buffer_queue.hpp
```diff
@@ -410,7 +390,7 @@ inline queue_op_status buffer_queue<value_t, buffer_t, buffer_policy>::try_pop(v
 
         newHeadReadPos = cyclic_increment(headReadPos, cap, roundSize);
 
-        if (this->headReadPos.compare_exchange_weak(headReadPos, newHeadReadPos))
+        if (this->headReadPos.compare_exchange_weak(headReadPos, newHeadReadPos, std::memory_order_acq_rel))
```
marehr: Why are you changing this from the default `std::memory_order_seq_cst`?

include/seqan3/contrib/parallel/latch.hpp
```diff
@@ -0,0 +1,164 @@
...
+     *
+     * no-throw
+     *
+     * ### Thread safety
+     *
+     * thread-safe
+     */
+    void arrive_and_wait(ptrdiff_t n = 1) noexcept
+    {
+        ++num_waiting; // ensure that destructor is not finished in-between the arrive and wait call.
```
marehr: > // ensure that destructor is not finished in-between the arrive and wait call.

Is this a todo or a gurantee?

include/seqan3/contrib/parallel/latch.hpp
```diff
@@ -0,0 +1,164 @@
...
+    latch(latch &&)                   = default; //!> Defaulted.
+    latch & operator=(latch const &)  = delete;  //!> Deleted.
+    latch & operator=(latch &&)       = default; //!> Defaulted.
+
+    //!\brief Destructs the latch and waits for all participating threads to arrive.
+    ~latch()
+    {
+        spin_delay delay{};
+        while (num_waiting.load(std::memory_order_acquire) > 0)
+            delay.wait();
```
marehr: What happens if multiple threads call the same destructor? Okay I see a latch is move-only, so only one thread is the owner.

include/seqan3/contrib/parallel/latch.hpp
```diff
@@ -0,0 +1,164 @@
...
+     * ### Thread safety
+     *
+     * thread-safe
+     */
+    void arrive_and_wait(ptrdiff_t n = 1) noexcept
+    {
+        ++num_waiting; // ensure that destructor is not finished in-between the arrive and wait call.
+        arrive(n);
+        wait();
+        --num_waiting;
```
marehr: is it guaranteed, that --num_waiting will not be reordered before the wait?

include/seqan3/contrib/parallel/latch.hpp
```diff
@@ -0,0 +1,164 @@
...
+            delay.wait();
+        --num_waiting;
+    }
+
+private:
+
+    //!\brief The number of participating threads.
+    alignas(std::hardware_destructive_interference_size) std::atomic<std::ptrdiff_t>         counter;
+    //!\brief The number of waiting threads.
+    alignas(std::hardware_destructive_interference_size) mutable std::atomic<std::ptrdiff_t> num_waiting;
```
marehr: yeah, that's how you do const correctness :yellow_heart:

include/seqan3/contrib/parallel/latch.hpp
```diff
@@ -0,0 +1,164 @@
...
+/*!\brief A single-use synchronisation point to coordinate concurrent threads.
+ *
+ * \details
+ *
+ * A latch is a thread coordination mechanism that allows any number of threads to block until an expected
+ * count is summed (exactly) by threads that arrived at the latch. The expected count is set when the latch is
+ * constructed. An individual latch is a single-use object; once the count has been reached, the latch cannot be reused.
+ * This implementation uses a lock-free mechanism if the atomic operations on the respective platform are lock-free.
+ *
+ * \note The adapts the proposal for the c++ standard
```
marehr: This adapts?

include/seqan3/contrib/parallel/reader_writer_latch.hpp
```diff
@@ -0,0 +1,323 @@
...
+struct reader_count : public detail::strong_type<size_t, reader_count>
+{
+    //!\brief An alias for the base type.
+    using base_t = detail::strong_type<size_t, reader_count>;
+
+    //!\brief Import the base constructor.
+    using base_t::base_t;
+};
+
+/*!\brief A single-use synchronisation point for closable concurrent data structures.
```
marehr: What are `closable concurrent data structures`?

include/seqan3/contrib/parallel/reader_writer_latch.hpp
```diff
@@ -0,0 +1,323 @@
...
+ *
+ * \details
+ *
+ * A reader writer latch is a thread coordination mechanism specifically designed to synchronise with a concurrent
+ * data structure such as the seqan3::contrib::buffer_queue. In particular, the latch is constructed with a
+ * seqan3::contrib::reader_count and a seqan3::contrib::writer_count and offers special functions for readers that
+ * arrive and writers that arrive at the synchronisation point. If all writers arrived at the synchronisation point
```
marehr: `for readers that arrive and writers that arrive` redundant `that arrive`

include/seqan3/contrib/parallel/reader_writer_latch.hpp
```diff
@@ -0,0 +1,323 @@
...
+ * the associated concurrent data structure is closed in a completion phase by one of the participating producer
+ * threads. Thus, the concurrent data structure will be automatically signalled that there is no more data coming.
+ * If all participating readers arrived at the synchronisation point no completion phase is invoked, but the
```
marehr: `no completion phase is invoked`? Is that right?

include/seqan3/contrib/parallel/reader_writer_latch.hpp
```diff
@@ -0,0 +1,323 @@
...
+ * destructor of the reader_writer_latch will wait until all readers, respectively writers arrived at the
+ * synchronisation point. This allows to protect a concurrent data structure against an early termination.
```
marehr: Can you reread the section. I'm not sure if everything is correct what you wrote.

include/seqan3/contrib/parallel/reader_writer_latch.hpp
```diff
@@ -0,0 +1,323 @@
...
+ * One can use the seqan3::contrib::scoped_reader_guard and seqan3::contrib::scoped_writer_guard to associate a thread
+ * with the reader_writer_latch as either a reaader or writer.
```
marehr: reaaaaaaaader :+1:

include/seqan3/contrib/parallel/reader_writer_latch.hpp
```diff
@@ -0,0 +1,323 @@
...
+ *
+ * The reader_writer_latch class is non-copyable and not default constructible.
+ */
+class reader_writer_latch
```
marehr: Is this class also part of the standard ts?

include/seqan3/contrib/parallel/reader_writer_latch.hpp
```diff
@@ -0,0 +1,323 @@
...
+     * \param ds     The concurrent data structure.
+     *
+     * \details
+     *
+     * Initialises a std::contrib::latch for the consumer threads and a std::contrib::latch for the producer threads.
+     * Initialises the completion function which calls `ds.close()` on completion. Note, only a reference to the
+     * concurrent data structure is stored. The caller must ensure that `ds` is not destructed before this.
+     */
+    template <typename concurrent_t>
+        requires requires { std::declval<concurrent_t>().close(); }  // requires closable queue.
```
marehr: closable queue or closable concurrent data structure?

include/seqan3/contrib/parallel/reader_writer_latch.hpp
```diff
@@ -0,0 +1,323 @@
...
+     * \param wcount The expected number of producer threads.
+     * \param ds     The concurrent data structure.
+     *
+     * \details
+     *
+     * Delegates to reader_writer_latch::reader_writer_latch(reader_count, writer_count, concurrent_t) with reader
+     * count set to 0.
+     */
+    template <typename concurrent_t>
+        requires requires { std::declval<concurrent_t>().close(); }  // requires closable queue.
```
marehr: closable queue or closable concurrent data structure?

include/seqan3/contrib/parallel/reader_writer_latch.hpp
```diff
@@ -0,0 +1,323 @@
...
+    reader_writer_latch(writer_count const wcount, concurrent_t & queue) :
```
marehr: above you named `queue` `ds`

include/seqan3/contrib/parallel/reader_writer_latch.hpp
```diff
@@ -0,0 +1,323 @@
...
+     * \param ds     The concurrent data structure.
+     *
+     * \details
+     *
+     * Delegates to reader_writer_latch::reader_writer_latch(reader_count, writer_count, concurrent_t) with writer
+     * count set to 0.
+     */
+    template <typename concurrent_t>
+        requires requires { std::declval<concurrent_t>().close(); }  // requires closable queue.
+    reader_writer_latch(reader_count const rcount, concurrent_t & queue) :
```
marehr: above you named `queue` `ds`

include/seqan3/contrib/parallel/reader_writer_latch.hpp
```diff
@@ -0,0 +1,323 @@
...
+     * \param rcount The expected number of consumer threads.
+     * \param ds     The concurrent data structure.
+     *
+     * \details
+     *
+     * Delegates to reader_writer_latch::reader_writer_latch(reader_count, writer_count, concurrent_t) with writer
+     * count set to 0.
+     */
+    template <typename concurrent_t>
+        requires requires { std::declval<concurrent_t>().close(); }  // requires closable queue.
```
marehr: closable queue or closable concurrent data structure?

include/seqan3/contrib/parallel/reader_writer_latch.hpp
```diff
@@ -0,0 +1,323 @@
...
+     *
+     * ### Thread safety
+     *
+     * thread-safe
+     */
+    void writer_arrive_and_wait() noexcept
+    {
+        writer_latch.arrive_and_wait();
+
+        std::call_once(flag, completion_fn);
```
marehr: didn't know that this exists :)

include/seqan3/contrib/parallel/reader_writer_latch.hpp
```diff
@@ -0,0 +1,323 @@
...
+     * ### Thread safety
+     *
+     * thread-safe
+     */
+    void writer_arrive() noexcept
+    {
+        writer_latch.arrive();
+
+        if (writer_latch.try_wait())
+            std::call_once(flag, completion_fn);
```
marehr: what happens if one thread invokes `writer_arrive_and_wait` and the other `writer_arrive`.

Is it possible that one thread executes `std::call_once(flag, completion_fn);` within `writer_arrive` and another thread did not execute `writer_arrive_and_wait`, yet?

`completion_fn` aka `ds.close` blocking? That means `writer_arrive` may be a `wait` function and has potential for death locks?

WARNING: This may be a stupid question.

include/seqan3/contrib/parallel/reader_writer_latch.hpp
```diff
@@ -0,0 +1,323 @@
...
+     *
+     * no-throw
+     *
+     * ### Thread safety
+     *
+     * thread-safe
+     */
+    void reader_arrive_and_wait() noexcept
+    {
+        reader_latch.arrive_and_wait();
```
marehr: How are the readers synchronised with the writers?

So if `writer_arrive_and_wait` was successful and closes the data structure, but there are still readers calling `arrive_and_wait`.

WARNING: This may be a stupid question.

include/seqan3/contrib/parallel/reader_writer_latch.hpp
```diff
@@ -0,0 +1,323 @@
...
+ * controlling participating producer threads of seqan3::contrib::reader_writer_latch for the duration of a scoped
+ * block. When a scoped_writer_guard object is created, it becomes a participating producer thread of the
+ * seqan3::contrib::reader_writer_latch it is given. When control leaves the scope in which the scoped_writer_guard
+ * object was created, the participating thread arrives at the synchronisation point in a non-blocking fashion and
+ * the scoped_writer_guard is destructed.
+ *
+ * The scoped_writer_guard class is non-copyable and not default constructible. The caller must ensure that the
+ * seqan3::contrib::reader_writer_latch is not destructed before the guard.
+ */
+class scoped_writer_guard
```
marehr: Is this part of the standard?

include/seqan3/contrib/parallel/reader_writer_latch.hpp
```diff
@@ -0,0 +1,323 @@
...
+};
+
+/*!\brief A strictly scope-based seqan3::contrib::reader_writer_latch wrapper for producer threads.
+ *
+ * \details
+ *
+ * The class scoped_writer_guard is a reader_writer_latch wrapper that provides a convenient RAII-style mechanism for
+ * controlling participating producer threads of seqan3::contrib::reader_writer_latch for the duration of a scoped
+ * block. When a scoped_writer_guard object is created, it becomes a participating producer thread of the
+ * seqan3::contrib::reader_writer_latch it is given. When control leaves the scope in which the scoped_writer_guard
```
marehr: control?

include/seqan3/contrib/parallel/reader_writer_latch.hpp
```diff
@@ -0,0 +1,323 @@
...
+ * object was created, the participating thread arrives at the synchronisation point in a non-blocking fashion and
```
marehr: I don't understand why it will not wait at the synchronisation point (aka in a non-blocking fashion). What do you mean with blocking.

include/seqan3/contrib/parallel/reader_writer_latch.hpp
```diff
@@ -0,0 +1,323 @@
...
+/*!\brief A strictly scope-based seqan3::contrib::reader_writer_latch wrapper for consumer threads.
+ *
+ * \details
+ *
+ * The class scoped_reader_guard is a reader_writer_latch wrapper that provides a convenient RAII-style mechanism for
+ * controlling participating consumer threads of seqan3::contrib::reader_writer_latch for the duration of a scoped
+ * block. When a scoped_reader_guard object is created, it becomes a participating consumer thread of the
+ * seqan3::contrib::reader_writer_latch it is given. When control leaves the scope in which the scoped_reader_guard
+ * object was created, the participating thread arrives at the synchronisation point in a non-blocking fashion and
+ * the scoped_reader_guard is destructed.
```
marehr: Same question as above.

include/seqan3/contrib/parallel/buffer_queue.hpp
rrahn: Because, it should be safe to acquire and release on the atomic, while it is more efficient on x86 architectures. `memory_order_seq_cst` will force a memory fence and thus can hit performance.

include/seqan3/contrib/parallel/latch.hpp
rrahn: Yep

include/seqan3/contrib/parallel/latch.hpp
rrahn: Well, it is a little confusing. The proposal says that the destructor should wait for all threads to leave wait. Of course this can't be guaranteed, since a thread could be stalled at any time by  a preemptive thread scheduler, thus after decrementing num_waiting the thread could lose control and the destructor could finish it's job. What it actually tries to prevent is that a thread is still waiting in the loop while the destructor is being called. I think there is no solution to that except to protect it from outside, but I took a glance into the reference implementation and they did like so. So I thought I take this as well.

include/seqan3/contrib/parallel/latch.hpp
rrahn: I assume yes, since wait uses memory_order_acquire, which means no read or write in the current thread can be reordered before this load on counter.

include/seqan3/contrib/parallel/latch.hpp
rrahn: I am merely following the proposal interface 😄

include/seqan3/contrib/parallel/reader_writer_latch.hpp
rrahn: Yes, we don't have a concept for this yet. But sooner or later we are going to bring this from contrib into seqan3 I think. The proposal for concurrent queues introduces the concept of closing a queue. If a queue is closed it won't allow pushes anymore. It also allows us to check when a queue is done reading from. Before in SeqAn this was managed internally by the queue itself by knowing the number of participating readers/writers. These latches/guards are used to achieve the same thing, but as external data structures.

include/seqan3/contrib/parallel/reader_writer_latch.hpp
rrahn: Yes, for readers we don't have a completion phase.

include/seqan3/contrib/parallel/reader_writer_latch.hpp
rrahn: Yes it is correct, but maybe I need to rephrase?

include/seqan3/contrib/parallel/reader_writer_latch.hpp
rrahn: nop, that is my doing. Do you have suggestions for a better name?

include/seqan3/contrib/parallel/reader_writer_latch.hpp
rrahn: It should be generalised, because I also need to have a concurrent stack interface for the bgzf_outstream

include/seqan3/contrib/parallel/reader_writer_latch.hpp
rrahn: nice, right

include/seqan3/contrib/parallel/reader_writer_latch.hpp
rrahn: Good question: Assume t1 enters `writer_arrive_and_wait` while t2 enters at the same time`writer_arrive`. Both are participating threads so the latch counter must be at least 2. If t1 has not executed yet t2 will decrement the counter and try_wait returns false. If t1 has executed in the meantime (until t2 asks try_wait) t2 will call the completion, but t1 is already finished. If t1 has executed but t2 not, then t1 will wait until t2 executes. If t2 executes then again either t2 or t1 will call the termination. The call_once is thread_safe and guarantees that only one thread is getting the privilege to call the completion phase. I think this is fairly safe.

include/seqan3/contrib/parallel/reader_writer_latch.hpp
rrahn: The readers are participating in a distinct synchronisation point. So if the writers are finished, the ds is closed, such that the readers can finish until the ds is empty or some other event removes them from the computation. Thus with this latch it is possible to avoid destructing a queue, simply by putting the latch after the queue in the member list. Thus on destruction of the wrapping data structure first the latch and then the queue is destructed, but only if all threads have finished there work.

include/seqan3/contrib/parallel/reader_writer_latch.hpp
rrahn: I took this from the documentation of https://en.cppreference.com/w/cpp/thread/lock_guard.

include/seqan3/contrib/parallel/reader_writer_latch.hpp
rrahn: because it calls arrive and not arrive and wait. The former will not wait, which means the thread resource can be used for other tasks.

include/seqan3/contrib/parallel/reader_writer_latch.hpp
rrahn: No, do you have suggestions to improve the naming?

include/seqan3/contrib/parallel/reader_writer_latch.hpp
rrahn: Same answer as above

include/seqan3/contrib/parallel/reader_writer_latch.hpp
marehr: Yeah it is hard to understand.

include/seqan3/contrib/parallel/reader_writer_latch.hpp
marehr: I meant the comment. I think the comment should be `closable concurrent data structure`

h-2:
Why not add this code to `core/parallel/` (or `core/parallel/detail/` if you want to be on the safe side)? This is all code written by yourself, it's documented and it looks like it conforms to our style-guide. 

`seqan3::contrib` is for external that fulfills neither of the above criteria....

include/seqan3/contrib/parallel/reader_writer_latch.hpp
marehr: Okay so control in the sense of control flow.

include/seqan3/contrib/parallel/reader_writer_latch.hpp
marehr: I'm not sure if guard is the right word, because you don't guard anything. You want multiple readers and/or multiple writers in that code block at the same time. The data structure (queue) handles then the concurrent accesses. This class tries to model the bookkeeping how many active reader and writers work on the queue.

I think `reader_process` would be a better name.

```c++
{
  reader_process{reader_writer_task};
  concurrent_reader{concurrent_reader_writer_task}; // or?
  // do what readers do 
} // signal that the reader process is finished.
``` 

Maybe we name the `reader_writer_latch` like `std::once_flag`, which contains the state of  `std::once`. So that the class is only considered as a state that we need to perform reader/writer processes.

test/unit/contrib/parallel/reader_writer_latch_test.cpp
```diff
@@ -0,0 +1,98 @@
...
+    // start the producer of source/consumer of target.
+    std::thread t1{[&] ()
+    {
+        {
+            scoped_writer_guard writer_guard{source_latch};
+
+            // Initialise source_queue.
+            for (size_t i = 0; i < job_size; ++i)
+            {
+                queue_op_status status = source_queue.try_push(1);
```
marehr: I would here sum up `i`, because the same value might hide side-effects within tested functions (for example reads the same value concurrently; even if already consumed).

test/unit/contrib/parallel/reader_writer_latch_test.cpp
```diff
@@ -0,0 +1,98 @@
...
+                return;
+
+            counter += val;
+        }
+    }};
+
+    // Start the consumer of source/producer of target.
+    std::vector<std::thread> pool;
+    for (size_t i = 0; i < threads; ++i)
+        pool.emplace_back(work);
```
marehr: Is there a reason why the lambda of work is not defined here?

test/unit/contrib/parallel/reader_writer_latch_test.cpp
```diff
@@ -0,0 +1,98 @@
...
+#include <seqan3/contrib/parallel/reader_writer_latch.hpp>
+
+using namespace seqan3::contrib;
+
+TEST(reader_writer_latch, parallel)
+{
+    size_t threads = std::thread::hardware_concurrency();
+
+    if (threads > 4)
+        threads = 4;
```
marehr: why do we limit this?

include/seqan3/contrib/parallel/buffer_queue.hpp
marehr: I don't have any expertise on memory order. Do you have a nice little tutorial that describes the difference?

> it should be safe to acquire and release on the atomic

I'm not sure `memory_order_acq_rel` has anything to do with being "safe" to acquire or release an atomic. I think it has something to do with the ability of the compiler to reorder the code positions of the assembly instructions.

rrahn: Well, I haven't had a very good tutorial. Some blogs etc. but in general you are right. I should at first make it as restricive as possible and guarantee sequential ordering. One needs to invest more time to completely reason about the correct behaviour. I change it back

include/seqan3/contrib/parallel/reader_writer_latch.hpp
rrahn: yes

rrahn:
> Why not add this code to core/parallel/ (or core/parallel/detail/ if you want to be on the safe side)? This is all code written by yourself, it's documented and it looks like it conforms to our style-guide.
> seqan3::contrib is for external code that fulfills neither of the above criteria....

In general I totally agree with you. But for the moment we only use it in combination with contrib stuff anyway. Since I wasn't completely sure yet about how to design the parallel dir I wanted to put it somewhere replacable. Also the latch implementation might get into c++20 and then I could have a drop and replace of it. I can do this all as well in a detail header but was at the moment going for the contrib. 
But I talk to @marehr of how we want to design the parallel module to avoid refactoring stuff in the end.

test/unit/contrib/parallel/reader_writer_latch_test.cpp
rrahn: In general, in SeqAn we had issues with travis.

include/seqan3/contrib/parallel/reader_writer_latch.hpp
```diff
@@ -0,0 +1,323 @@
...
+     * \param ds     The concurrent data structure.
+     *
+     * \details
+     *
+     * Delegates to reader_writer_latch::reader_writer_latch(reader_count, writer_count, concurrent_t) with reader
+     * count set to 0.
+     */
+    template <typename concurrent_t>
+        requires requires { std::declval<concurrent_t>().close(); }  // requires closable concurrent data structure.
+    reader_writer_latch(writer_count const wcount, concurrent_t & ds) :
```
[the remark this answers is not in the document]
rrahn: That's true, it becomes a normal latch, except that a normal latch (by proposal) does not have a completion phase and I didn't want to also implement the barrier, because it is more complex. I am still hoping that this will get into c++20. But in general you are right, the interface is wonky and should only allow writers and readers as the class name suggests.

include/seqan3/contrib/parallel/reader_writer_latch.hpp
rrahn: How about, 

reader_writer_latch -> agent_handler
scoped_reader_guard -> read_agent
scoped_writer_guard -> write_agent

test/unit/contrib/parallel/reader_writer_latch_test.cpp
rrahn: I find it easier to read.

include/seqan3/core/parallel/detail/reader_writer_manager.hpp
```diff
@@ -0,0 +1,315 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+     * \brief Provides seqan3::detail::reader_writer_manager and utility objects.
```
h-2: :nail_care:

include/seqan3/core/parallel/detail/reader_writer_manager.hpp
```diff
@@ -0,0 +1,315 @@
...
+ * done when their destructor is called.
+ *
+ * The reader_writer_manager class is move-only and cannot be default constructed.
+ */
+class reader_writer_manager
+{
+private:
+
+    //!\brief A strictly scope-based seqan3::detail::reader_writer_manager wrapper for producer threads.
+    class [[nodiscard]] scoped_writer_type
```
h-2: this works? :open_mouth:

include/seqan3/core/parallel/detail/reader_writer_manager.hpp
```diff
@@ -0,0 +1,315 @@
...
+     * Initialises a std::contrib::latch for the consumer threads and a std::contrib::latch for the producer threads.
+     * Initialises the completion function which calls `ds.close()` on completion. Note, only a reference to the
+     * concurrent data structure is stored. The caller must ensure that `ds` is not destructed before this.
+     *
+     * ### Exception
+     *
+     * Throws std::invalid_argument if reader count or writer count is less than 1.
+     */
+    template <typename concurrent_t>
+        requires requires { std::declval<concurrent_t>().close(); }  // requires closable concurrent data structure.
```
h-2: doxygen

include/seqan3/core/parallel/detail/reader_writer_manager.hpp
```diff
@@ -0,0 +1,315 @@
...
+     * data structure. On destruction of the returned RAII wrapper class
+     * seqan3::detail::reader_writer_manager::writer_arrive is called automatically.
+     *
+     * ### Exception
+     *
+     * no-throw
+     *
+     * ### Thread safety
+     *
+     * thread-safe
```
h-2: :nail_care: 

I think we write full sentences for this or at least "Sentence case."
```
     * ### Exception
     *
     *  Guaranteed not to throw.
     *
     * ### Thread safety
     *
     *  Thread-safe.
```

I think we should actually make doxygen macros for this, but that can happen after the release.

include/seqan3/core/parallel/detail/reader_writer_manager.hpp
rrahn: nice, right!

include/seqan3/contrib/parallel/buffer_queue.hpp
```diff
@@ -63,23 +63,11 @@ class buffer_queue
...
-        close();  // close the queue.
-
-        // wait until queue is empty (all readers are finished).
-        spin_delay delay{};
-        while (!is_empty())
-            delay.wait();
-
-        assert(tailPos == tailWritePos);
-        assert(headPos == headReadPos);
-    }
```
marehr: Just being curious, why is defaulted destructor now okay?

include/seqan3/core/parallel/detail/latch.hpp
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /*!\brief Constructs the latch with the expected number of threads.
+     * \param expected The number of threads participating in this synchronisation point.
+     */
+    explicit latch(ptrdiff_t const expected) :
+        counter{expected}
+    {
+        assert(expected >= 0);
+        num_waiting.store(0, std::memory_order_relaxed);
```
marehr: You use `std::memory_order_relaxed` instead of the default? This was from the proposal, right?

marehr:
travis is failing?! Ah it's gcc-8.

```c++
CMakeFiles/gap_scheme_test.dir/gap_scheme_test.cpp.o: In function `seqan3::test::tmp_filename::tmp_filename(char const*)':
gap_scheme_test.cpp:(.text._ZN6seqan34test12tmp_filenameC2EPKc[_ZN6seqan34test12tmp_filenameC5EPKc]+0x5e7): undefined reference to `std::filesystem::__cxx11::filesystem_error::_M_gen_what()'
gap_scheme_test.cpp:(.text._ZN6seqan34test12tmp_filenameC2EPKc[_ZN6seqan34test12tmp_filenameC5EPKc]+0x69d): undefined reference to `std::filesystem::__cxx11::filesystem_error::_M_gen_what()'
```

include/seqan3/core/parallel/detail/reader_writer_manager.hpp
h-2: pretty cool

PR #981, opened 2019-05-16
Add semialphabet any

sarahet:


include/seqan3/alphabet/composite/semialphabet_any.hpp
```diff
@@ -0,0 +1,81 @@
...
+ * \implements seqan3::StandardLayout
+ *
+ * \details
+ * This alphabet provides a generic representation for different alphabets of the same size by erasing
+ * the type of the alphabet it is constructed from. This enables the usage of a single type although
+ * assigning different alphabets. The semialphabet_any can also be converted to any other (semi-)alphabet
+ * of the same size.
+ * It is therefore possible to convert the semialphabet_any into an alphabet type that is not the original
+ * alphabet type. However, this should either be avoided or used with care as no warnings are given when attempting
+ * to convert the semialphabet_any into a type that is not comparable to the original alphabet type.
```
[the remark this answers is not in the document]
h-2: I would maybe add one more sentence:

```

The main advantage of using this data structure is to reduce instantiation of templates when using multiple alphabets of the same size and either their character representation is not important or they are reified at a later point in the program.

File: include/seqan3/alphabet/composite/semialphabet_any.hpp [hunk 2]
```diff
@@ -0,0 +1,81 @@
...
+ *
```
h-2: Add before the snippet
```
### Example
```

include/seqan3/alphabet/composite/semialphabet_any.hpp
```diff
@@ -0,0 +1,81 @@
...
+ * \snippet test/snippet/alphabet/composite/semialphabet_any.cpp example
+ */
```
h-2: Maybe add:
```
\see https://en.wikipedia.org/wiki/Type_erasure
\see https://en.wikipedia.org/wiki/Reification_(computer_science)
```

include/seqan3/alphabet/composite/semialphabet_any.hpp
```diff
@@ -0,0 +1,81 @@
...
+    constexpr semialphabet_any()                                     noexcept = default; //!< Defaulted.
+    constexpr semialphabet_any(semialphabet_any const &)             noexcept = default; //!< Defaulted.
+    constexpr semialphabet_any(semialphabet_any &&)                  noexcept = default; //!< Defaulted.
+    constexpr semialphabet_any & operator=(semialphabet_any const &) noexcept = default; //!< Defaulted.
+    constexpr semialphabet_any & operator=(semialphabet_any &&)      noexcept = default; //!< Defaulted.
+    ~semialphabet_any()                                              noexcept = default; //!< Defaulted.
+
+    //!\brief Construct semialphabet_any from alphabet of the same size
+    template <Semialphabet other_alph_t>
+        requires alphabet_size_v<other_alph_t> == size
```
h-2:
```suggestion
//!\cond
        requires (alphabet_size_v<other_alph_t> == size)
//!\endcond
```

We need to hide the requires clause from doxygen and we need to also put () around most terms now (because clang requires this).

include/seqan3/alphabet/composite/semialphabet_any.hpp
```diff
@@ -0,0 +1,81 @@
...
+    semialphabet_any(other_alph_t const other)
```
h-2: Please make both the constructor and the conversion operator `explicit`. This will at least prevent the conversions from happening automatically and force people to use `static_cast` (or direct initialisation).

include/seqan3/alphabet/composite/semialphabet_any.hpp
```diff
@@ -0,0 +1,81 @@
...
+    {
+        using seqan3::to_rank;
```
h-2: this weird stuff is not required anymore since #943

include/seqan3/alphabet/composite/semialphabet_any.hpp
h-2: Also for the conversion operator

include/seqan3/alphabet/composite/semialphabet_any.hpp
```diff
@@ -0,0 +1,81 @@
...
+        requires alphabet_size_v<other_alph_t> == size
+    operator other_alph_t()
+    {
+        other_alph_t other{};
+        assign_rank_to(to_rank(), other);
+        return other;
+    }
+};
+
+}
```
h-2:
```suggestion
} // namespace seqan3
```

:nail_care:

test/snippet/alphabet/composite/semialphabet_any.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+    // Conversion back to original alphabet
+    aa10li_vector v3 = v2;
+    debug_stream << v3 << "\n"; // AIKAAAKCB
+
+    // Conversion to another alphabet
+    // This is possible however should be used with care since there will be no warning given when attempting to convert
+    // a semialphabet_any into an alphabet that is not the original alphabet. This may lead to errors when transforming
+    // the semialphabet_any into an alphabet type that is not comparable to the original type.
+    aa10murphy_vector v4 = v2;
```
h-2: Hm, but this is what we want people not to do!

I would maybe make the example like this (roughly sketched and not tested, bad grammar)
```cpp
// print is a template and gets instantiated two times,
// because the behaviour is different for both types
template <typename rng_t>
void print(rng_t && r)
{
    seqan3::debug_stream << r << '\n';
}

// algorithm is not a template, only one instance is generated by the compiler;
// type information is encoded via a run-time parameter
void algorithm(std::vector<semialphabet_any<10>> & r, bool is_murphy)
{
    // a silly algorithm example that replaces rank 0 with rank 1:
    for (auto & v : r)
        if (seqan3::to_rank(v) == 0)
            seqan3::assign_rank_to(1, v);

    // here we reify the type for printing
    if (is_murphy)
        print(v | seqan3::view::convert<aa10murphy>);
    else
        print(v | seqan3::view::convert<aa10li>);
}

// two instances of algo_pre exist; they type erase the different arguments to the same type
// and encode the type information as a run-time parameter
void algo_pre(aa10li_vector const & v)
{
    seqan3::semialphabet_any<10> tmp = v | seqan3::view::convert<semialphabet_any<10>>;
    algo(tmp, false);
}
void algo_pre(aa10murphy_vector const & v)
{
    seqan3::semialphabet_any<10> tmp = v | seqan3::view::convert<semialphabet_any<10>>;
    algo(tmp, true);
}

int main()
{
    seqan33::aa10li_vector v1 {...};
    algo_pre(v1);

    seqan3::aa10murphy_vector v2 {...};
    algo_pre(v2);
}
```

And explanation:
```
The example illustrates how we can reduce the usage of templates
(or the amount of different instantiations) via type erasure. Having only
one function generated for `algo()` is the only benefit of  using `semialphabet_any`
here. Of course this only makes sense for your application if the part of the program
that is agnostic of the character representation (your equivalent of `algo()`) is
substantially larger than the specific parts – and if compile-time and/or 
size of the exectuble are a concern.
```

h-2: BTW: we don't use `using namespace seqan3;` in the snippets anymore...

include/seqan3/alphabet/composite/semialphabet_any.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+#pragma once
+
+#include <seqan3/alphabet/detail/alphabet_base.hpp>
+
+/*!\brief A semi-alphabet that type erases all other semi-alphabets of the same size.
+ * \ingroup composite
+ * \implements seqan3::TriviallyCopyable
+ * \implements seqan3::StandardLayout
+ *
```
h-2: \implements seqan3::Semialphabet

include/seqan3/alphabet/composite/semialphabet_any.hpp
sarahet: Not required because it needs to be replaced by something else? Or not required and should just work without?

Because I get an error when I remove it and this PR is so huge I can not immediately find an answer :smile:

h-2: Write `seqan3::` before the calls in the line below instead. This is only necessary if you have member functions of the same name. Otherwise you can omit the `seqan3::` entirely.

include/seqan3/alphabet/composite/semialphabet_any.hpp
```diff
@@ -0,0 +1,81 @@
...
+    {
+        using seqan3::to_rank;
+        assign_rank(to_rank(other));
+    }
+    //!\}
+
+    //!\brief Enable conversion of semialphabet_any into other (semi-)alphabet of the same size
+    template <Semialphabet other_alph_t>
+        requires alphabet_size_v<other_alph_t> == size
+    operator other_alph_t()
```
h-2: your conversion operator needs to be `const`-qualified, otherwise you imply that the function will change the object and it won't be possible to call it in a const-context, like in `view::convert`.

include/seqan3/alphabet/composite/semialphabet_any.hpp
h-2: > Write seqan3:: before the calls in the line below instead. 

Just before `to_rank()`

PR #983, opened 2019-05-16
[INFRA] use two jobs on travis

h-2:


h-2:
I have monitored RAM usage locally and it seems we never exceed ~4-5GB so two jobs should work fine on travis everywhere.

Issue #730, opened 2019-02-12
GCC9 support


rrahn:
Open reports:

- [ ] https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90003 (duplicate of https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89953)
- [ ] https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89953

Fixed GCC9 bugs:
- #708 
- #719 
- #730
- #902 
- #986 
- #989 
- #990

rrahn:
@h-2 is there more to do?

h-2:
In C++17 mode we only have one error left, I am not 100% sure this is a side-affect of other (reported) issues. I'll investigate that.

In C++20 mode there are 11 tests failing, AFAICT all due to a reported and fixed issue. These will disappear with the next package upgrade likely in a week or two.

rrahn:
Let us keep this alive until we know for sure then.

h-2:
We still have some open issues, many related to alignment code. @rrahn Can you have a look? I think GCC9 is scheduled for late april so we should really get these reported ASAP.

h-2:
An alignment snippet is also ICEing on GCC8-nightlies:
http://cdash.seqan.de/viewBuildError.php?buildid=303901

rrahn:
Yes, I'll check this.

h-2:
Can you link to all open GCC bug reports here?

rrahn:
- [ ] https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90003
- [ ] https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89953

h-2:
We now have workarounds for everything.

PR #986, opened 2019-05-19
[FIX] alignment_configurator for gcc9

marehr:
@h-2 You are first to review; @rrahn is second to review!

With this alignment compiles with gcc9 with and without `-std=c++2a`

### DIFF
--- a/include/seqan3/alignment/pairwise/alignment_configurator.hpp
+++ b/include/seqan3/alignment/pairwise/alignment_configurator.hpp
@@ -386,7 +386,12 @@ private:
             }
             else if constexpr (align_ends_cfg_t::template is_static<1>())
             {
+#if defined(__GNUC__) && __GNUC__ >= 9
+                constexpr bool free_ends_trailing = align_ends_cfg_t::template get_static<1>();
+                return configure_edit_traits(std::integral_constant<bool, free_ends_trailing>{});
+#else // ^^^ workaround / no workaround vvv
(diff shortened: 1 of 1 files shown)


include/seqan3/alignment/pairwise/alignment_configurator.hpp
```diff
@@ -386,7 +386,12 @@ struct alignment_configurator
             }
             else if constexpr (align_ends_cfg_t::template is_static<1>())
             {
+#if defined(__GNUC__) && __GNUC__ >= 9
+                constexpr bool free_ends_trailing = align_ends_cfg_t::template get_static<1>();
+                return configure_edit_traits(std::integral_constant<bool, free_ends_trailing>{});
```
rrahn: We could also just omit the macro hackery here and keep this version, which is probably valid for all compilers. 
Also why is only this part affected? There is at least a second function with the same semantics. And this breaks nothing?

rrahn:
Can you link please to #730

include/seqan3/alignment/pairwise/alignment_configurator.hpp
marehr: We discussed, that we want to keep track of such bugs.

h-2:
@rrahn Is it sufficient that plan on adding the `_WORKAROUND` infrastructure after the release or do you wnat changes to this code?

PR #990, opened 2019-05-20
[FIX] alignment_file/output on gcc9

marehr:
The main problem is that std::string_view requires that the char type
fulfills std::is_trivial_v via static_assert.

### DIFF
--- a/test/unit/alignment/aligned_sequence_test_template.hpp
+++ b/test/unit/alignment/aligned_sequence_test_template.hpp
@@ -27,7 +27,8 @@ TYPED_TEST_CASE_P(aligned_sequence);
 TYPED_TEST_P(aligned_sequence, fulfills_concept)
 {
     EXPECT_TRUE((AlignedSequence<TypeParam>));
-    EXPECT_FALSE((AlignedSequence<std::vector<dna4>>));
+    EXPECT_FALSE((AlignedSequence<std::vector<dna4>>)); // has no gap character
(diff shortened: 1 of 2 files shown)


test/unit/alignment/aligned_sequence_test_template.hpp
```diff
@@ -28,6 +28,7 @@ TYPED_TEST_P(aligned_sequence, fulfills_concept)
 {
     EXPECT_TRUE((AlignedSequence<TypeParam>));
     EXPECT_FALSE((AlignedSequence<std::vector<dna4>>));
+    EXPECT_TRUE((AlignedSequence<std::array<gapped<dna4>, 0>>));
```
marehr: @smehringer As it turns out, it seems that this is not a aligned sequence. I'm not sure which condition of the concept fails.

smehringer: yes this is because I require to be able to use insert and erase on an aligned sequence.
But it does not matter because I do not require the AlignedSequence concept for writing an alignment (I need no insert functionality). I just require that the value_type  is comparable to gap. so it should be fine? What does the output file tests say?

marehr: I added a strategy meeting card.

marehr:
@h-2 Next time you could write something that this is superseded by #1012 .

h-2:
Sorry, I wasn't requested on this PR and didn't see it.

PR #993, opened 2019-05-20
[FEATURE] add score_matrix and trace_matrix for edit_distance

marehr:

PR #994, opened 2019-05-20
[FEATURE] add score_matrix for edit_distance

marehr:


marehr:
Closed in favour for #993

PR #997, opened 2019-05-20
Alphabet last PR from me

h-2:
Just renaming and changing comparison operators to compare by value. But would be good to get in soon, because touches lots of files.

### DIFF
--- a/include/seqan3/alphabet/all.hpp
+++ b/include/seqan3/alphabet/all.hpp
@@ -123,7 +123,7 @@
  *
  * |                                  | seqan3::Semialphabet | seqan3::WritableSemialphabet | seqan3::Alphabet | seqan3::WritableAlphabet | Aux |
  * |----------------------------------|:--------------------:|:----------------------------:|:----------------:|:------------------------:|:---:|
- * | seqan3::alphabet_size_v          | ✅                    | ✅                            | ✅                | ✅                        |     |
+ * | seqan3::alphabet_size            | ✅                    | ✅                            | ✅                | ✅                        |     |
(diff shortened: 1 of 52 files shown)


include/seqan3/alphabet/all.hpp
```diff
@@ -123,7 +123,7 @@
  *
  * |                                  | seqan3::Semialphabet | seqan3::WritableSemialphabet | seqan3::Alphabet | seqan3::WritableAlphabet | Aux |
  * |----------------------------------|:--------------------:|:----------------------------:|:----------------:|:------------------------:|:---:|
- * | seqan3::alphabet_size_v          | ✅                    | ✅                            | ✅                | ✅                        |     |
+ * | seqan3::alphabet_size          | ✅                    | ✅                            | ✅                | ✅                        |     |
```
smehringer: indentation :nail_care:

test/unit/alphabet/composite/alphabet_variant_test.cpp
```diff
@@ -250,19 +250,19 @@ TEST(alphabet_variant_test, rank_type)
     EXPECT_TRUE((std::is_same_v<alphabet3_t::rank_type, uint16_t>));
 }
 
-TEST(alphabet_variant_test, value_size)
+TEST(alphabet_variant_test, alphabet_size)
```
smehringer:
```suggestion
TEST(alphabet_variant_test, alphabet_size_)
```
?

h-2: ?

h-2: I changed although I don't think it was necessary.

PR #998, opened 2019-05-20
[MISC] Move ostream iterator out of std/ranges header.

### LINKED ISSUE #988: We have std::ranges::ostream_iterator etc imported into std/ranges. But I couldn't find any reference that they will be added with c++20.
In this case we need to remove them and implement them in SeqAn.

Port the ostream iterator to seqan3/io/stream. Alias the std::ostream if ranges feature test macro is set otherwise forward to ranges::ostream_iterator.

rrahn:
Fixes #988

### DIFF
--- a/include/seqan3/io/stream/iterator.hpp
+++ b/include/seqan3/io/stream/iterator.hpp
@@ -1,0 +1,60 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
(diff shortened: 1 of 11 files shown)


include/seqan3/io/stream/iterator.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+/*!\file
+ * \brief Provides seqan3::ostream and seqan3::ostreambuf iterator.
+ * \author Rene Rahn <rene.rahn AT fu-berlin.de>
+ */
+
+#pragma once
+
+#ifdef __cpp_lib_ranges
+#include <iterator>
```
h-2: you need to include the header before checking the macro, otherwise it may not be defined.

rrahn: Ok, I included <version> which will provide all feature test macros.

h-2: version is c++20, it broke your builds... You can just include iterator undconditionally...

rrahn: Yeah, I just realized that. Haven’t read the notes. Could have been tagged for c++20 more clearly on cppreference though 😒

rrahn:
travis linker error on filsystem. otherwise good to go

PR #1001, opened 2019-05-20
[MISC] switch compute_* on/off depending on if we need the computation

marehr:


include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
```diff
@@ -97,12 +91,31 @@ class pairwise_alignment_edit_distance_unbanded
...
     //!\brief When true the computation will use the ukkonen trick with the last active cell and bounds the error to config.max_errors.
-    static constexpr bool use_max_errors = detail::MaxErrors<align_config_t>;
+    static constexpr bool use_max_errors = align_config_type::template exists<align_cfg::max_error>();
     //!\brief Whether the alignment is a semi-global alignment or not.
     static constexpr bool is_semi_global = traits_t::is_semi_global_type::value;
     //!\brief Whether the alignment is a global alignment or not.
     static constexpr bool is_global = !is_semi_global;
 
+    //!\brief Whether the alignment configuration indicates to compute and/or store the score.
+    static constexpr bool compute_score = align_config_type::template exists<align_cfg::result<with_score_type>>() ||
```
rrahn: I think this is superfluous. The minimum requirement is to compute the score. So you can be sure that it is enabled.

include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
```diff
@@ -391,57 +404,70 @@ class pairwise_alignment_edit_distance_unbanded
...
+            res_vt.alignment = alignment_trace(database, query, trace_matrix(), res_vt.back_coordinate);
         }
         return alignment_result<result_value_type>{std::move(res_vt)};
     }
 
     //!\brief Return the score of the alignment.
     score_type score() const noexcept
     {
+        static_assert(compute_score, "score() can only be computed if you specify the result type within your alignment "
+                                     "config, e.g. `align_cfg::edit | align_cfg::result{with_score}`");
```
rrahn: Also it is helpful to give an example in the error message, but I would not do it. In case we change something before 3.1 in the API we might not detect this here.

marehr: I could add a decltype here to check the syntax xD

include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
marehr: There might be configurations there I don't need that

include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
marehr: I removed the .e.g. part

PR #1002, opened 2019-05-21
Fix/small string and take

rrahn:
Yesterday while building all tests of the library for mac I spotted an error in the range iterator test. It caused an error when printing the small_string to std::cerr. Since the small_string should mimic the string interface I added overloads for the formatted input and output. Thereby, I saw that string has an additional erase interface which I added as well. 
Then while investigating the error, I updated the size type of the take view to use the size type of the underlying range if applicable. Still the test was wrong, because the size type in the test changed no matter what you would do. .

### DIFF
--- a/include/seqan3/range/container/small_string.hpp
+++ b/include/seqan3/range/container/small_string.hpp
@@ -12,6 +12,7 @@
 
 #pragma once
 
+#include <algorithm>
 #include <seqan3/range/container/small_vector.hpp>
 
 namespace seqan3
@@ -334,6 +335,65 @@ public:
     operator std::string() const
     {
         return str();
+    }
+    //!\}
+
+    /*!\name Input/output
+     * \{
+     */
+
+    /*!\brief Formatted output for the seqan3::small_string.
+     * \param[in,out] os  The std::basic_ostream to write to.
+     * \param[in]     str The seqan3::small_string to read from.
+     * \returns `os`.
+     *
+     * \details
+     *
+     * Internally calls `os << str.str()`.
+     */
+    friend std::ostream & operator<<(std::ostream & os, small_string const & str)
+    {
+        os << str.str();
+        return os;
+    }
+
+    /*!\brief Formatted input for the seqan3::small_string.
+     * \param[in,out] is  The std::basic_istream to read from.
+     * \param[out]    str The seqan3::small_string to write to.
+     * \returns `is`.
+     *
+     * \details
+     *
+     * Reads at most seqan3::small_string::max_size characters from the stream.
+     * If a stream error occurred or no characters could be extracted the std::ios_base::failbit is set.
+     * This may throw an exception.
+     */
+    friend std::istream & operator>>(std::istream & is, small_string & str)
+    {
+        // Check if stream is ok and skip leading whitespaces.
+        std::istream::sentry s(is);
+        if (s)
+        {
+            str.erase(); // clear the string
+            std::streamsize num_char = (is.width() > 0)
+                ? std::min<std::streamsize>(is.width(), str.max_size())
+                : str.max_size();
+            assert(num_char > 0);
+            for (std::streamsize n = num_char; n > 0 && !std::isspace(static_cast<char>(is.peek()), is.getloc()); --n)
+            {
+                char c = is.get();
+                if (is.eof())
+                    break;
+                str.push_back(c);
+            }
+
+            if (str.size() == 0) // nothing extracted so we set the fail bit.
+                is.setstate(std::ios_base::failbit);
+
+            is.width(0); // cancel the effects of std::setw, if any.
+        }
+
+        return is;
     }
     //!\}
 };
--- a/include/seqan3/range/container/small_string.hpp
+++ b/include/seqan3/range/container/small_string.hpp
@@ -228,6 +228,33 @@ public:
         sz = count;
         data_[sz] = '\0';
     }
+
+    /*!\brief Removes specified elements from the container.
+     * \param   index Remove the elements starting at `index`. Defaults to `0`.
+     * \param   count The number of elements to remove. Defaults to `max_size()`.
+     * \returns `*this`
+     *
+     * Invalidates iterators and references at or after the point of the erase, including the end() iterator.
+     *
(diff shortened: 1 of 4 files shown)


test/unit/range/container/small_string_test.cpp
```diff
@@ -111,6 +111,21 @@ TEST(small_string, implicit_conversion)
     EXPECT_EQ(str, "hello"s);  // explicit
 }
 
+TEST(small_string, erase)
+{
+    small_string em{"hello"};
+    em.erase();
+    EXPECT_EQ(em.str(), ""s);
```
smehringer: Can you please add a constexpr version of the test?

include/seqan3/range/container/small_string.hpp
```diff
@@ -334,6 +335,39 @@ class small_string : public small_vector<char, capacity_ + 1>
...
+    /*!\name Input/output
+     * \{
+     */
+
+    /*!\brief Formatted output for the seqan3::small_string.
+     * \param[in,out] os  The std::basic_ostream to write to.
+     * \param[in]     str The seqan3::small_string to read from.
+     * \returns `os`.
+     */
+    friend std::ostream & operator<<(std::ostream & os, small_string const & str)
```
smehringer: why is it not enough to be printable with our debug_stream?

test/unit/core/metafunction/range_iterator_test.cpp
```diff
@@ -58,7 +58,7 @@ void expect_same_types()
     constexpr bool val = std::is_same_v<meta::at_c<list1, pos>, meta::at_c<list2, pos>>;
     if constexpr (!val)
     {
-        std::cerr << "\'" << detail::get_display_name_v<meta::at_c<list1, pos>> << "\' not equal to \'"
+        std::cerr << "pos: " << pos << " \'" << detail::get_display_name_v<meta::at_c<list1, pos>>.str() << "\' not equal to \'"
```
smehringer: if you we keep your small_string overload for ostream introduced in the former commit you should not need the `.str()` here.

test/unit/range/container/small_string_test.cpp
rrahn: sure

include/seqan3/range/container/small_string.hpp
rrahn: Because it mimics all the interfaces of std::string except input/output. I think this is not super important but simply more complete and thus less confusing.

include/seqan3/range/container/small_string.hpp
```diff
@@ -228,6 +229,31 @@ class small_string : public small_vector<char, capacity_ + 1>
...
+     * Linear in size().
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    constexpr small_string & erase(size_type index = 0, size_type count = max_size()) noexcept
+    {
+        iterator it = this->begin() + index;
+        base_t::erase(it, it + std::min<size_type>(count, this->size() - index));
```
smehringer: maybe assert that `index  < this->size()`?

include/seqan3/range/container/small_string.hpp
```diff
@@ -309,6 +335,39 @@ class small_string : public small_vector<char, capacity_ + 1>
...
+     * \returns `is`.
+     */
+    friend std::istream & operator>>(std::istream & is, small_string & str)
+    {
+        str.erase();
+        std::string tmp;
+        is >> tmp;
+
+        auto min_size = std::min<typename std::string::size_type>(tmp.size(), str.max_size());
+        std::ranges::move(tmp.begin(), tmp.begin() + min_size, std::back_inserter(str));
```
smehringer: hm is this the expected behaviour, that the input is silently truncated?

rrahn: man 🤦‍♂ I was first adding the complete ugly implementation then thought tonight, ah just put it into a string and copy from there. But you are right, this might consume to much from the input.  So I need to get the ugly version back.

smehringer: we should assert though that the capacity is exceeded? 
I think trying to assign a string from stream that is larger than the capacity is UB so I don't know if this needs to be ugly?

rrahn: No, the stream has special functionality, for example only reading until a whitespace occurs or reading at most is.width  characters from the stream if width is set, etc. See here the details for the function https://en.cppreference.com/w/cpp/string/basic_string/operator_ltltgtgt. But if I parse it first into a string I cannot control the capacity there, so even if we only read as much characters as the capacity of small_string the basic string before might have consumed already more input. That's why I think that there isn't a more elegant strategy.

rrahn:
Only the filesystem linker error on travis.

include/seqan3/range/container/small_string.hpp
```diff
@@ -309,6 +337,53 @@ class small_string : public small_vector<char, capacity_ + 1>
...
+     * \returns `is`.
+     */
+    friend std::istream & operator>>(std::istream & is, small_string & str)
+    {
+        // Check if stream is ok and skip leading whitespaces.
+        std::istream::sentry s(is);
+        if (s)
+        {
+            str.erase(); // clear the string
+            std::streamsize num_char = (is.width() > 0) ? is.width() : str.max_size();
```
smehringer: take the maximum of width and str.max_size so you don't exceed the capacity and document the behaviour (only comsume at most str.capacity characters from stream).

include/seqan3/range/container/small_string.hpp
```diff
@@ -309,6 +337,65 @@ class small_string : public small_vector<char, capacity_ + 1>
...
+    }
+
+    /*!\brief Formatted input for the seqan3::small_string.
+     * \param[in,out] is  The std::basic_istream to read from.
+     * \param[out]    str The seqan3::small_string to write to.
+     * \returns `is`.
+     *
+     * \details
+     *
+     * Reads at most seqan3::small_string::small_size characters from the stream.
```
smehringer:
```suggestion
     * Reads at most seqan3::small_string::max_size characters from the stream.
```

rrahn: What? There is no `small_size` member function 😱

Issue #895, opened 2019-04-11
views need extra documentation entry for header


h-2:
Since adaptors don't get their own documentation, the views don't have an entry for which header file provides them. This confused people on the retreat.

rrahn:
Add `\headerfile` in documentation of view definitions, as they are not shown correctly by doxygen. 
@MitraDarja Can you work on this. You can come to me if you have questions for this.

rrahn:
fixed by #1004

PR #1004, opened 2019-05-21
[MISC] Include header file in documentation.

MitraDarja:
Adding includes to view functions
#895

### DIFF
--- a/include/seqan3/range/view/take_until.hpp
+++ b/include/seqan3/range/view/take_until.hpp
@@ -572,6 +572,11 @@ namespace seqan3::view
  *
  * \details
  *
+ * **Header**
+ * ```cpp
+ *      #include <seqan3/range/view/take_until.hpp>
+ * ```
+ *
  * ### View properties
  *
  * | range concepts and reference_t  | `urng_t` (underlying range type)      | `rrng_t` (returned range type)                     |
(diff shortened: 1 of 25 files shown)


include/seqan3/range/view/take_until.hpp
```diff
@@ -572,6 +572,11 @@ namespace seqan3::view
  *
  * \details
  *
+ * **Header**
+ * ```cpp
+ *      #include <seqan3/range/view/take_until.hpp>
+ * ```
+ *
```
h-2: There are multiple view adaptor definitions in this file, I think you forgot the others? Please also double-check the other files.

MitraDarja: No, I checked all Headers are in the definition, for some files there seems to be a general documentation part because it is the same for all functions.

h-2: What about view::take_exactly_or_throw? What about view::take_until_and_consume? ....

MitraDarja: Checked all view functions, all had a header.

h-2: Ah, then that's resolved because of the \copydetails

PR #1012, opened 2019-05-22
final GCC9 fixes

h-2:


include/seqan3/range/container/small_string.hpp
```diff
@@ -80,7 +80,7 @@ class small_string : public small_vector<char, capacity_ + 1>
      * No-throw guarantee.
      */
     template <size_t N>
-    explicit constexpr small_string(char const (&_lit)[N]) noexcept : small_string{}
+    constexpr small_string(char const (&_lit)[N]) noexcept : small_string{}
```
h-2: we need implicit construction from string literals to use this in template non-type argument contexts.

include/seqan3/contrib/parallel/buffer_queue.hpp
```diff
@@ -141,7 +141,7 @@ class buffer_queue
     {
         spin_delay delay{};
 
-        value_type value;
+        value_type value{};
```
h-2: This was actually a "may be used uninitialised"... maybe that was our problem on FreeBSD, too?

rrahn: unlikely, but let's see

PR #1013, opened 2019-05-22
Feature/ibgzf stream

rrahn:
Adds ibgzf_stream from SeqAn2. This is rebased on the latch PR. Only the last two commits are relevant.

Blocked by #980 

Resolves first part for #996

### DIFF
--- a/include/seqan3/contrib/stream/bgzf_istream.hpp
+++ b/include/seqan3/contrib/stream/bgzf_istream.hpp
@@ -1,0 +1,622 @@
+// zipstream Library License:
+// --------------------------
+//
+// The zlib/libpng License Copyright (c) 2003 Jonathan de Halleux.
+//
+// This software is provided 'as-is', without any express or implied warranty. In no event will the authors be held liable for any damages arising from the use of this software.
+//
+// Permission is granted to anyone to use this software for any purpose, including commercial applications, and to alter it and redistribute it freely, subject to the following restrictions:
+//
+// 1. The origin of this software must not be misrepresented; you must not claim that you wrote the original software. If you use this software in a product, an acknowledgment in the product documentation would be appreciated but is not required.
+//
+// 2. Altered source versions must be plainly marked as such, and must not be misrepresented as being the original software.
+//
+// 3. This notice may not be removed or altered from any source distribution
+//
+//
+// Author: Jonathan de Halleux, dehalleux@pelikhan.com, 2003   (original zlib stream)
+// Author: David Weese, dave.weese@gmail.com, 2014             (extension to parallel block-wise compression in bgzf format)
+// Author: Rene Rahn, rene.rahn AT fu-berlin.de, 2019          (adaption to SeqAn3)
+
+#pragma once
+
+#include <cstring>
+#include <condition_variable>
+#include <mutex>
+
+#include <seqan3/contrib/parallel/buffer_queue.hpp>
+#include <seqan3/contrib/stream/bgzf_stream_util.hpp>
+#include <seqan3/core/parallel/detail/reader_writer_manager.hpp>
+#include <seqan3/io/exception.hpp>
+
+namespace seqan3::contrib
+{
+
+// ===========================================================================
+// Classes
+// ===========================================================================
+
+// --------------------------------------------------------------------------
+// Class basic_bgzf_istreambuf
+// --------------------------------------------------------------------------
+
+template<
+    typename Elem,
+    typename Tr = std::char_traits<Elem>,
+    typename ElemA = std::allocator<Elem>,
+    typename ByteT = char,
+    typename ByteAT = std::allocator<ByteT>
+>
+class basic_bgzf_istreambuf : public std::basic_streambuf<Elem, Tr>
+{
+public:
+
+    typedef Tr                              traits_type;
+    typedef typename traits_type::char_type char_type;
+    typedef typename traits_type::int_type  int_type;
+    typedef typename traits_type::off_type  off_type;
+    typedef typename traits_type::pos_type  pos_type;
+
+private:
+
+    typedef std::basic_istream<Elem, Tr>& istream_reference;
+
+    typedef ElemA                   char_allocator_type;
+    typedef ByteT                   byte_type;
+    typedef ByteAT                  byte_allocator_type;
+    typedef byte_type*              byte_buffer_type;
+
+    typedef std::vector<char_type, char_allocator_type> TBuffer;
+    typedef fixed_buffer_queue<int32_t>                 TJobQueue;
+
+    static const size_t MAX_PUTBACK = 4;
+
+    // Allows serialized access to the underlying buffer.
+    struct Serializer
+    {
+        istream_reference   istream;
+        std::mutex          lock;
+        io_error            *error;
+        off_type            fileOfs;
+
+        Serializer(istream_reference istream) :
+            istream(istream),
+            error(NULL),
+            fileOfs(0u)
+        {}
+
+        ~Serializer()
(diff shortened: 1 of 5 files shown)


include/seqan3/io/exception.hpp
```diff
@@ -47,6 +51,19 @@ struct parse_error : std::runtime_error
...
+//!\brief Thrown if there is an io error in low level io operations such as in std::basic_streambuf operations.
+struct io_error : std::ios_base::failure
+{
+    //!\brief Constructor that forwards the exception string.
+    io_error(std::string const & s,
+             std::error_code const & ec) : std::ios_base::failure{s, ec}
+    {}
+
+    //!\brief Constructor that forwards the exception string.
+    explicit io_error(std::string const & s) : io_error{s, std::make_error_code(std::io_errc::stream)}
```
h-2: How is this error_code generated? Is there a global shared state of the streams? :scream:

include/seqan3/contrib/stream/bgzf_istream.hpp
```diff
@@ -0,0 +1,635 @@
...
+// --------------------------------------------------------------------------
+
+template<
+    typename Elem,
+    typename Tr = std::char_traits<Elem>,
+    typename ElemA = std::allocator<Elem>,
+    typename ByteT = char,
+    typename ByteAT = std::allocator<ByteT>
+>
+class basic_decompression_bgzf_streambuf : public std::basic_streambuf<Elem, Tr>
```
h-2: I would recommend to make the naming similar to the other streams so it is easier to interface from SeqAn, i.e. call this class `basic_bgzf_istreambuf` (for gz it's called `basic_gz_istreambuf`  and for bz2 it's called `basic_bz2_istreambuf`).

include/seqan3/contrib/stream/bgzf_istream.hpp
```diff
@@ -0,0 +1,635 @@
...
+// Author: Jonathan de Halleux, dehalleux@pelikhan.com, 2003   (original zlib stream)
+// Author: David Weese, dave.weese@gmail.com, 2014             (extension to parallel block-wise compression in bgzf format)
+// Author: Rene Rahn, rene.rahn AT fu-berlin.de, 2019          (adaption to SeqAn3)
+
+#pragma once
+
+// TODO(rrahn): replace me with constructor argument
+#ifndef SEQAN_BGZF_NUM_THREADS
+    #define SEQAN_BGZF_NUM_THREADS 16
+#endif
```
h-2: Can you replace this with an `inline static const` global variable that is initialised with `std::thread::hardware_concurrency()`? I think that would make a lot of sense now that we have that feature or not?

include/seqan3/contrib/stream/bgzf_stream_util.hpp
```diff
@@ -0,0 +1,328 @@
...
+
+#pragma once
+
+#include <algorithm>    // copy
+#include <array>
+
+#if SEQAN3_HAS_ZLIB
+// Zlib headers
+#include <zlib.h>
+#endif
```
h-2: I would suggest:
```cpp
#else
#error "This file cannot be used when building without GZip-support."
#endif
```

include/seqan3/contrib/stream/bgzf_stream_util.hpp
```diff
@@ -0,0 +1,328 @@
...
+                                                                  '\x00', '\x00', '\x00', '\x00',
+                                                                  '\x00', '\x00', '\x00', '\x00'}};
+
+template <typename TAlgTag>
+struct CompressionContext {};
+
+template <typename TAlgTag>
+struct DefaultPageSize;
+
+#if SEQAN3_HAS_ZLIB
```
h-2: this can then be removed

include/seqan3/contrib/stream/bgzf_stream_util.hpp
```diff
@@ -0,0 +1,328 @@
...
+{
+    static const unsigned MAX_BLOCK_SIZE = 64 * 1024;
+    static const unsigned BLOCK_FOOTER_LENGTH = 8;
+    // 5 bytes block overhead (see 3.2.4. at http://www.gzip.org/zlib/rfc-deflate.html)
+    static const unsigned ZLIB_BLOCK_OVERHEAD = 5;
+
+    // Reduce the maximal input size, such that the compressed data
+    // always fits in one block even for level Z_NO_COMPRESSION.
+    enum { BLOCK_HEADER_LENGTH = CompressionContext<detail::bgzf_compression>::BLOCK_HEADER_LENGTH };
+    static const unsigned VALUE = MAX_BLOCK_SIZE - BLOCK_HEADER_LENGTH - BLOCK_FOOTER_LENGTH - ZLIB_BLOCK_OVERHEAD;
```
h-2: Are these variables the same values as the MACRO constants in the other file? If yes, can you replace usage in the other file with using the static values instead?
[if this is a lot of work, it's not so important, I just thought that if we can prevent polluting the global namespace with names that would be nice]

include/seqan3/contrib/stream/bgzf_stream_util.hpp
h-2: (the other compression headers behave this way, too)

include/seqan3/io/exception.hpp
rrahn: I am not sure I understand what you mean? The function returns a new error_code (local variable) initialised with the enum value std::io_errc::stream. But I should change it to `std::error_code{std::io_errc::stream}`, which is effectively the same. But there is no shared state or something.

include/seqan3/contrib/stream/bgzf_istream.hpp
rrahn: Well, I left everything as it was in SeqAn2. I first thought about extending the constructors of the stream to pass the thread number from outside, and actually even further to provide an executor interface later. But that would make the interface for the alignment file also more complicated because you would have to specify the thread num/executor there as well. Until we have a proper concept for supporting parallelism throughout all our files I think the solution with the `inline static` variable is very good. But I would not make it const, so people can override it in their program by assigning another value to it.

include/seqan3/contrib/stream/bgzf_stream_util.hpp
rrahn: sure, I have a look.

rrahn:
> Just some minor stuff. Can you also add this to the stream_input_benchmark? Should just be a few lines and will immediately give you a comparison to SeqAn2!

I wanted to look at this when bgzf_ostream is in in another PR to keep it easier to review. But if everything else is ok, I can add it now.

include/seqan3/contrib/stream/bgzf_istream.hpp
h-2: > I first thought about extending the constructors of the stream to pass the thread number from outside, 

It's already a constructor arg of the streambuffer! One would only need to pass it through, but it's not important for the release.

> But that would make the interface for the alignment file also more complicated because you would have to specify the thread num/executor there as well.

Yes, I already thought of it. My plan was to simply add a `threads` member variable to the options of the files.  But this is also not relevant for the release and I am sure you have more ideas for this :wink: 

> Until we have a proper concept for supporting parallelism throughout all our files I think the solution with the inline static variable is very good. But I would not make it const, so people can override it in their program by assigning another value to it.

It's a good intermediate solution! We should definitely change it in the long run, though.

include/seqan3/io/exception.hpp
h-2: Well it's not important.

Issue #889, opened 2019-04-10
Add vecotisation for simple scores to the alignment module


rrahn:
Implement a vectorised alignment kernel with the inter-simd scheme.

PR #1017, opened 2019-05-24
Feature/simd unpack

rrahn:
Adds unpack operations for different target architectures.
This is needed for the to_simd_chunk view, which when using the matrix transpose functionality needs to unpack the vector types back to their original packing width.

I tested it with sde on sse4, avx2, and avx512

One part of #889


marehr:
As discussed with @rrahn: it seems that this does not perform on types other than `char`.

It seems that auto-vectorization yields the same performance.

In general we need more real-world benchmarks to implement this.

PR #1019, opened 2019-05-24
Ctd exporter parisi

emanueleparisi:
WIP version of the CTD exporter

### DIFF
--- a/include/seqan3/argument_parser/detail/format_ctd.hpp
+++ b/include/seqan3/argument_parser/detail/format_ctd.hpp
@@ -42,7 +42,7 @@ namespace
  * is completely initialized. Instead, every call is stored and evaluated only
  * when format_ctd::parse() is called.
  */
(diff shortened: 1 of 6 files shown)


include/seqan3/argument_parser/detail/format_ctd.hpp
```diff
@@ -0,0 +1,612 @@
...
+
+    template <typename option_type>
+    bool 
+    value_is_container(option_type const & /* value */) 
+    {
+        return SequenceContainer<option_type> && !std::is_same_v<option_type,
+                                                                 std::string>;
+    }
+
+    // TODO (emanueleparisi) This function is copied from format_parse.hpp, as it is
```
smehringer: yes, there is the format_base.hpp that you can (and should) inherit from. Then you can move this function into the base class and use it in both.

PR #1022, opened 2019-05-27
Feature/enable bgzf stream

### LINKED ISSUE #996: Add bgzf istream and ostream

rrahn:
fixes #996

### DIFF
--- a/include/seqan3/io/detail/misc_input.hpp
+++ b/include/seqan3/io/detail/misc_input.hpp
@@ -21,6 +21,8 @@
     #include <seqan3/contrib/stream/bz2_istream.hpp>
 #endif
 #ifdef SEQAN3_HAS_ZLIB
+    #include <seqan3/contrib/stream/bgzf_istream.hpp>
+    #include <seqan3/contrib/stream/bgzf_stream_util.hpp>
     #include <seqan3/contrib/stream/gz_istream.hpp>
 #endif
 #include <seqan3/io/detail/magic_header.hpp>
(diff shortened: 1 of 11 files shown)


include/seqan3/io/detail/misc_input.hpp
```diff
@@ -73,32 +75,50 @@ template <char_concept char_t>
...
-    // restore original stream content
-    for (size_t i = 0; i < magic_header_size; ++i)
-        primary_stream.putback(magic_number[magic_header_size - 1 - i]);
+    // unget all read chars.
+    for (size_t i = 0 ; i < read_chars; ++i)
+        primary_stream.unget();
+
+    // file is smaller than 4 bytes -> not compressed, could be legal
+    if (read_chars < 4)
+        return {&primary_stream, stream_deleter_noop};
```
h-2: Why 4 ? This was originally the header read, but now you read more. The problem is more complex however. A file might be smaller that the magic header of bgzf (18 bytes) and still be valid uncompressed  OR compressed. E.g. `> \n` compressed in zstd is 16 bytes.

h-2: Maybe removing this check will just be ok :thinking:

rrahn: I think everything is still correct. We get at most `read_chars` chars. If this is less than 4 we assume an uncompressed file. If `read_char` is exactly 18 (number of bytes we read at most) and it is a bgzf header we read bgzf. Otherwise we check if it is gz, bz2, or zsd. In this case we have read at least 4 bytes (size of magic header of zstd) and `read_chars` <= 18 bytes. If we don't check for at least 4 bytes, then I need to check again if `starts_with` is correct. Then of course I can remove the check of 4 bytes and return the uncompressed primary stream at the end. I simply moved it out since before we were not ungetting the read chars from the stream if it was less than 4 bytes.

test/unit/io/structure_file/structure_file_output_test.cpp
```diff
@@ -509,11 +551,12 @@ std::string expected_bz2
...
-{
-    test::tmp_filename filename{"structure_file_output_test.dbn.bz2"};
-    compression_by_filename_impl(filename, expected_bz2);
-}
+// TEST_F(structure_file_output_compression, by_filename_bz2)
+// {
+//     test::tmp_filename filename{"structure_file_output_test.dbn.bz2"};
+//     std::string buffer = compression_by_filename_impl(filename);
+//     EXPECT_EQ(buffer, expected_bz2);
+// }
```
h-2: Why was this commented out?

rrahn: Forgot to reenable.  thanks

PR #1023, opened 2019-05-27
[TEST] sam benchmark

Clemapfel:
Benchmark reading SAM files both from disk and stream, compare with seqan2
Also contains small fix for parsing cigar strings containing 'P'

### DIFF
--- a/test/performance/io/format_sam_benchmark.cpp
+++ b/test/performance/io/format_sam_benchmark.cpp
@@ -1,0 +1,320 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+// !\author Clemens Cords (clemens.cords@fu-berlin.de)
+
+#include <cctype>
+#include <cstring>
+#include <cmath>
+#include <fstream>
+#include <sstream>
+#include <type_traits>
+#include <streambuf>
+
+#include <benchmark/benchmark.h>
+
+#include <seqan3/alignment/pairwise/alignment_result.hpp>
+#include <seqan3/alignment/pairwise/alignment_configurator.hpp>
+#include <seqan3/alignment/pairwise/align_pairwise.hpp>
+#include <seqan3/alphabet/nucleotide/dna15.hpp>
+#include <seqan3/alphabet/quality/phred42.hpp>
+#include <seqan3/io/alignment_file/output.hpp>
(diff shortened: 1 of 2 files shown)


test/performance/io/format_sam_benchmark.cpp
```diff
@@ -0,0 +1,331 @@
...
+
+                case Mutation_type::SUBSTITUTION:
+                    mutated_sequence.push_back(alphabet_t{}.assign_rank(mutation_distribution(rng_engine)));
+                    break;
+            }
+        }
+    }
+
+    return mutated_sequence;
+}
```
Clemapfel: I wanted to ask whether I should move this function into it's own header close to ``test/performance/sequence_generator.hpp`` since it's technically generic enough to be reused for test cases later on

Clemapfel: I wanted to ask whether I should move this function into it's own header close to test/performance/sequence_generator.hpp since it's technically generic enough to be reused for test cases later on

test/performance/io/format_sam_benchmark.cpp
```diff
@@ -0,0 +1,331 @@
...
+        file_dict[n_queries] = file;
+    }
+
+    return file_dict[n_queries];
+}
+
+// ============================================================================
+// read as many sam files as possible
+// ============================================================================
+void sam_read_from_stream (benchmark::State & state)
```
smehringer: please exchange the format reading by directly reading from the file. this is easier and will model the reality better :)

```cpp
void sam_file_read_from_stream (benchmark::State & state)
{
    size_t n_queries = state.range(0);
    std::string sam_file = get_file(n_queries);

    std::istringstream istream{sam_file};

    for (auto _ : state)
    {
        // refill stream (same constant for seqan2 benchmark)
        istream.clear();
        istream.seekg(0, std::ios::beg);

        alignment_file_input fin{istream, format_sam{}};

        // read all records and store in internal buffer
        auto it = fin.begin();
        while (it != fin.end())
            ++it;
    }
}
```

test/performance/io/format_sam_benchmark.cpp
```diff
@@ -0,0 +1,331 @@
...
+    test::tmp_filename file_name{"tmp.sam"};
+    auto tmp_path = file_name.get_path();
+
+    size_t n_queries = state.range(0);
+
+    std::ofstream ostream{tmp_path};
+    ostream << get_file(n_queries);
+    ostream.close();
+
+    alignment_file_input sam_in {tmp_path.c_str(), fields<field::SEQ,
```
smehringer: you construct a file here that you do not use ;)

smehringer:
You can remove the changes in the sam files now and rebase onto the current master :)

smehringer:
Hm your benchmarks throws an error on travis 
```
terminate called after throwing an instance of 'std::logic_error'
  what():  The aligned sequences must have the same length.
```

Since the alignments are random, I don't if you can reproduce the error but this error should never happen. The pair of aligned sequences must always have the same length. Can you investigate?

smehringer:
Superseeded by #1920

PR #1025, opened 2019-05-28
[fix] update translate view

sarahet:
Removing shared pointer, adding non-const iterators and member functions.

This does not work yet! 
Some help needed finding the reason why the output range is not a view :pray:


h-2:
Sorry, I didn't make it today, but I will have a look at it first thing tomorrow.

h-2:
Blocked by https://github.com/ericniebler/range-v3/issues/1209

rrahn:
I am not sure if it helps, but when I implemented the  single pass view I had similar issues. The reason was that the constructor for viewable range ends up in a cyclic  dependency when requiring the ViewableRange concept on the template argument. I had to move the requirement below the self check (the rng is not the type of the view itself) and then it worked. But as I said, I am not sure that this is related to this, or whether you tried this already.

h-2:
> I am not sure if it helps, but when I implemented the single pass view I had similar issues. The reason was that the constructor for viewable range ends up in a cyclic dependency when requiring the ViewableRange concept on the template argument. I had to move the requirement below the self check (the rng is not the type of the view itself) and then it worked. But as I said, I am not sure that this is related to this, or whether you tried this already.

image

So, of course there were already gaurds against self. And I also tried removing the ViewableRange check in the constructor altogether which didn't help. What did make the difference is checking for:
```
std::Constructible<urng_t, ranges::ref_view<std::remove_reference_t<rng_t>>>
```
instead of
```
std::Constructible<urng_t, std::ranges::all_view<rng_t>>
```

For reasons I do not understand at all, invoking the unmodified constructor **did work** and produces an working object. The object just didn't model the concept for some reason.

include/seqan3/range/view/translate.hpp
```diff
@@ -321,28 +372,58 @@ class view_translate_single
...
+             return translate_triplet((urange)[n * 3 + 2], (urange)[n * 3 + 3], (urange)[n * 3 + 4]);
+             break;
+         case translation_frames::REV_FRAME_2:
+             return translate_triplet(complement((urange)[(urange).size() - n * 3 - 3]), complement((urange)[(urange).size() - n * 3 - 4]), complement((urange)[(urange).size() - n * 3 - 5]));
+             break;
+         default:
+             throw std::invalid_argument(multiple_frame_error.c_str());
+             break;
+     }
+    }
```
h-2: somehow the indention got broken here

include/seqan3/range/view/translate.hpp
```diff
@@ -353,6 +434,19 @@ class view_translate_single
...
+        ret.resize(size());
+        std::copy(cbegin(), cend(), ret.begin());
+        return ret;
+    }
+
+    //!\overload
+    template <RandomAccessContainer container_type>
     operator container_type() const
     //!\cond
         requires std::is_same_v<aa27, value_type_t<container_type>>
```
h-2: Please remove both of the conversion operators, they are not part of View interfaces anymore. We have #1033 now.

include/seqan3/range/view/translate.hpp
```diff
@@ -636,15 +743,36 @@ class view_translate
...
+    {
+        container_type ret;
+        ret.resize(size());
+        for (size_type i = 0; i < size(); i++)
+            ret[i] = static_cast<value_type_t<container_type>>(operator[](i));
+        return ret;
+    }
+
+    //!\overload
+    template <RandomAccessContainer container_type>
```
h-2: here as well

h-2:
Can you also have a look at the coverage:
https://codecov.io/gh/seqan/seqan3/pull/1025/diff?src=pr&el=tree#diff-aW5jbHVkZS9zZXFhbjMvcmFuZ2Uvdmlldy90cmFuc2xhdGUuaHBw

It seems like quite a few code-paths are not tested...

sarahet:
Tests are passing now. Does the code coverage gets re-computed again at some point? I fixed some things I found but to me it was not too obvious that there where so many things missing, actually. Would be good to see the report again now.

sarahet:
Ah well, I just realized if you just click on it again it shows the new results :see_no_evil: 
Looks much higher now!

h-2:
Thanks for updating/fixing this!

PR #1030, opened 2019-05-28
"fix" for the sequence file performance

h-2:
1. add our own view::istream that is faster than std::istreambuf_iterators
2. conditionally remove usage of most views from sequence file fasta
  * not all views seem to be bad in all situations, but view::filter and view::join seems to really kill the performance currently

This also increased the size of the benchmarked FastA-file a little, because the previous one was too small. That's why the numbers for SeqAn2 changed, too.

|          | SeqAn3 before | SeqAn3 after | SeqAn2 before | SeqAn2 after |
|--------|----------------|-------------|---------------|--------------|
| read  GCC7 |  139.474M/s | 454.196M/s | 383.446M/s | 495.321M/s |
| read GCC9 | 119.112M/s  | 510.165M/s | 446.884M/s | 553.42M/s |
| write GCC7 | 164.803M/s | 331.439M/s | 109.172M/s | 109.186M/s |
| write GCC9 | 180.341M/s | 302.799M/s | 113.784M/s | 109.434M/s |


include/seqan3/range/view/istreambuf.hpp
```diff
@@ -0,0 +1,251 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \author Hannes Hauswedell <hannes.hauswedell AT fu-berlin.de>
+ * \brief Provides seqan3::view::istream.
```
smehringer: istream_buf

include/seqan3/range/view/istreambuf.hpp
```diff
@@ -0,0 +1,251 @@
...
+private:
+    //!\brief Down-cast pointer to the stream-buffer.
+    stream_buffer_exposer<char_t, traits_t> * stream_buf = nullptr;
+
+public:
+    /*!\name Associated types
+     * \brief All are derived from the base_base_t.
+     * \{
+     */
+    using difference_type   = ptrdiff_t;               //!\< Defaults to ptrdiff_t.
```
smehringer: without the backslash!

include/seqan3/range/view/istreambuf.hpp
```diff
@@ -0,0 +1,251 @@
...
+template <typename char_t, typename traits_t = std::char_traits<char_t>>
+class fast_istreambuf_iterator
+{
+private:
+    //!\brief Down-cast pointer to the stream-buffer.
+    stream_buffer_exposer<char_t, traits_t> * stream_buf = nullptr;
+
+public:
+    /*!\name Associated types
+     * \brief All are derived from the base_base_t.
```
smehringer: who is this mysterious base_base_t :thinking:

include/seqan3/range/view/istreambuf.hpp
```diff
@@ -0,0 +1,251 @@
...
+
+// ============================================================================
+//  istreambuf_fn (adaptor definition)
+// ============================================================================
+
+//!\brief View adaptor/factory definition for view::istream.
+struct istreambuf_fn
+{
+    /*!\brief Return the view object.
+     * \param s Reference to the stream buffer.
```
smehringer: param[in, out]
also we usually put param below tparam

test/unit/range/view/view_istreambuf_test.cpp
```diff
@@ -0,0 +1,67 @@
...
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+#include <iostream>
+
+#include <gtest/gtest.h>
+
+#include <seqan3/alphabet/nucleotide/dna5.hpp>
+#include <seqan3/range/view/istreambuf.hpp>
+#include <seqan3/range/view/char_to.hpp>
```
smehringer: order :nail_care:

test/unit/range/view/view_istreambuf_test.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include <seqan3/range/view/complement.hpp>
+#include <seqan3/std/ranges>
+
+#include <seqan3/core/debug_stream.hpp>
```
smehringer: rather pretty printing?

test/unit/range/view/view_istreambuf_test.cpp
```diff
@@ -0,0 +1,67 @@
...
+    auto v2 = view::istreambuf(*is.rdbuf());
+    i = 0;
+    for (auto c : v2)
+        EXPECT_EQ(c, data[i++]);
+    EXPECT_EQ(i, data.size());
+
+    // combinability
+    is.clear();
+    is.seekg(0, std::ios::beg);
+    auto v3 = view::istreambuf(is) | view::char_to<dna5> | view::complement;
```
smehringer: can you test a take_until (instead) since it is the usual stuff we need in the io module?

include/seqan3/io/sequence_file/format_fasta.hpp
```diff
@@ -36,6 +29,8 @@
 #include <seqan3/range/shortcuts.hpp>
 #include <seqan3/range/detail/misc.hpp>
 #include <seqan3/range/view/char_to.hpp>
+#include <seqan3/range/view/istreambuf.hpp>
+#include <seqan3/range/view/interleave.hpp>
```
smehringer: you don't need this now as you do not use it because it is slower. I would delete it for now?

include/seqan3/io/sequence_file/format_fasta.hpp
```diff
@@ -191,20 +183,71 @@ class sequence_file_format_fasta
...
+                }
+
+                if (!at_delimiter)
+                    throw unexpected_end_of_input{"FastA ID line did not end in newline."};
+
+                for (; (it != e) && ((!is_char<'\n'>)(*it)); ++it)
+                {}
+
+            #else // ^^^ WORKAROUND | ORIGINAL ↓↓↓
+
```
smehringer: remove newline

include/seqan3/io/sequence_file/format_fasta.hpp
```diff
@@ -277,30 +342,35 @@ class sequence_file_format_fasta
...
+        #else // ^^^ WORKAROUND | ORIGINAL ↓↓↓
+
+            //TODO: combining chunk and join is substantially faster than view::interleave (2.5x), why?
             std::ranges::copy(seq | view::to_char
-                             | ranges::view::chunk(options.fasta_letters_per_line)
-                             | std::view::join(options.add_carriage_return
+                                  | ranges::view::chunk(options.fasta_letters_per_line)
+                                  | std::view::join(options.add_carriage_return
                                                     ? std::string_view{"\r\n"}
```
smehringer: String_view on a temporary?

include/seqan3/io/sequence_file/format_fasta.hpp
```diff
@@ -191,20 +182,71 @@ class sequence_file_format_fasta
...
+                    id.push_back(assign_char_to(*it, value_type_t<id_type>{}));
+                }
+
+                if (!at_delimiter)
+                    throw unexpected_end_of_input{"FastA ID line did not end in newline."};
+
+                for (; (it != e) && ((!is_char<'\n'>)(*it)); ++it)
+                {}
+
+            #else // ^^^ WORKAROUND | ORIGINAL ↓↓↓
```
rrahn: Shouldn't we not stick to ASCII characters whenever possible?

include/seqan3/io/sequence_file/format_fasta.hpp
```diff
@@ -191,20 +182,71 @@ class sequence_file_format_fasta
...
+
+                std::ranges::copy(stream_view | std::view::drop_while(is_id || is_blank)        // skip leading >
                                               | view::take_until_or_throw(is_cntrl || is_blank) // read ID until delimiter…
                                               | view::char_to<value_type_t<id_type>>,
                                   std::back_inserter(id));                                      // … ^A is old delimiter
 
                 // consume rest of line
                 detail::consume(stream_view | view::take_line_or_throw);
+            #endif
```
rrahn:
```suggestion
            #endif // SEQAN3_WORKAROUND_VIEW_PERFORMANCE
```
And other places

include/seqan3/io/sequence_file/format_fasta.hpp
```diff
@@ -191,20 +182,71 @@ class sequence_file_format_fasta
...
+                    throw unexpected_end_of_input{"FastA ID line did not end in newline."};
+
+            #else // ^^^ WORKAROUND | ORIGINAL ↓↓↓
+
+                std::ranges::copy(stream_view | view::take_line_or_throw                    // read line
+                                              | std::view::drop_while(is_id || is_blank)    // skip leading >
                                               | view::char_to<value_type_t<id_type>>,
                                   std::back_inserter(id));
+            #endif
+
```
rrahn: empty line

include/seqan3/range/view/istreambuf.hpp
```diff
@@ -0,0 +1,248 @@
...
+    fast_istreambuf_iterator()                                              noexcept = default; //!< Defaulted.
+    fast_istreambuf_iterator(fast_istreambuf_iterator const &)              noexcept = default; //!< Defaulted.
+    fast_istreambuf_iterator(fast_istreambuf_iterator &&)                   noexcept = default; //!< Defaulted.
+    fast_istreambuf_iterator & operator=(fast_istreambuf_iterator const &)  noexcept = default; //!< Defaulted.
+    fast_istreambuf_iterator & operator=(fast_istreambuf_iterator &&)       noexcept = default; //!< Defaulted.
+    ~fast_istreambuf_iterator()                                             noexcept = default; //!< Defaulted.
+
+    //!\brief Construct from a stream buffer.
+    explicit fast_istreambuf_iterator(std::basic_streambuf<char_t, traits_t> & ibuf) :
+        stream_buf{reinterpret_cast<stream_buffer_exposer<char_t, traits_t> *>(&ibuf)}
```
rrahn: not sure this valid. Have you verified that this does not trigger warnings with sanitizers enabled.
You trying to tell the compiler that this class is just another way of accessing the bit representation, but the members are all protected and could be overloaded and interleaved with other members in the derived class of basic_streambuf you are getting here. That's why I had to wrap it in SeqAn2 in the first place.

include/seqan3/range/view/istreambuf.hpp
```diff
@@ -0,0 +1,248 @@
...
+    {
+        assert(stream_buf != nullptr);
+        return *stream_buf->gptr();
+    }
+
+    /*!\name Comparison operators
+     * \brief We define comparison only against the sentinel.
+     * \{
+     */
+    //!\brief True if the read buffer is not empty; involves not vtable lookup.
```
rrahn:
```suggestion
    //!\brief True if the read buffer is not empty; involves no vtable lookup.
```

rrahn: and the other places

include/seqan3/range/view/istreambuf.hpp
```diff
@@ -0,0 +1,248 @@
...
+     * \returns A std::ranges::subrange over a detail::fast_istreambuf_iterator and std::ranges::default_sentinel_t.
+     */
+    template <typename stream_char_t, typename stream_traits_t>
+    constexpr auto operator()(std::basic_streambuf<stream_char_t, stream_traits_t> & s) const
+    {
+        return std::ranges::subrange<detail::fast_istreambuf_iterator<stream_char_t, stream_traits_t>,
+                                     std::ranges::default_sentinel_t>
+        {
+            detail::fast_istreambuf_iterator<stream_char_t, stream_traits_t>{s},
+            std::ranges::default_sentinel_t{}
```
rrahn:
```suggestion
            std::ranges::default_sentinel
```

include/seqan3/io/sequence_file/format_fasta.hpp
h-2: Well, we defined the following rule in the wiki ages ago:
```
 All files are expected to be UTF-8 encoded. Special characters are allowed in documentation and string literals, but not in regular code (function names etc) and file names.
```

I use UTF8 in a lot of places...

include/seqan3/range/view/istreambuf.hpp
h-2: Well, we are just storing a pointer, or not?

h-2: The following example at least does not produce sanitzier warnings:
```cpp

struct S
{
protected:
   void foo() {};
};

struct T : S
{
    using S::foo;
};

int main()
{
    S s;
    T * t = reinterpret_cast<T*>(&s);
}
```

smehringer:
Travis is green although it does not update here

PR #1031, opened 2019-05-29
[FIX] Fixes the unpack/pack operations for bgzf streams.

rrahn:


include/seqan3/contrib/stream/bgzf_stream_util.hpp
```diff
@@ -139,13 +140,17 @@ compressInit(CompressionContext<detail::bgzf_compression> & ctx)
 inline uint16_t
 _bgzfUnpack16(char const * buffer)
 {
-    return detail::to_little_endian(*reinterpret_cast<uint16_t const *>(buffer));
+    uint16_t tmp;
+    std::uninitialized_copy(buffer, buffer + sizeof(uint16_t), reinterpret_cast<char *>(&tmp));
```
smehringer: Why not memcpy?

h-2:
Why are these changes necessary?

marehr:
> Why are these changes necessary?

Because of valid sanitizer issues.

The problem is that 
```c++
*reinterpret_cast<uint16_t const *>(buffer)
```

of a

```c++
char const * buffer
```
is unsafe, because uint16_t has a different data alignment then char. So the reinterpret_cast is unsafe / undefined behaviour.

The fix first copies the corresponding bytes into a proper memory region and then access it.

h-2:
Ah, I didn't think about alignment issues.

PR #1033, opened 2019-05-30
[FEATURE] introduce ranges::to; remove implicit conversions

### LINKED ISSUE #915: move form implicit conversion to ranges::to

h-2:
fixes #915 

Since the views from the range library are still implicitly convertible, I might have missed a case or two, but we can fix that when upstream changes or we have a C++20 implementation to test against.

None of our views are implicitly convertible anymore.

### DIFF
--- a/test/unit/range/view/view_to_rank_test.cpp
+++ b/test/unit/range/view/view_to_rank_test.cpp
@@ -19,19 +19,19 @@ using namespace seqan3;
 TEST(view_to_rank, basic)
 {
     dna5_vector vec{"ACTTTGATA"_dna5};
-    std::vector<unsigned> cmp{0,1,4,4,4,2,0,4,0};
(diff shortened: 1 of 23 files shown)


test/unit/range/view/view_to_rank_test.cpp
```diff
@@ -19,19 +19,19 @@ using namespace seqan3;
 TEST(view_to_rank, basic)
 {
     dna5_vector vec{"ACTTTGATA"_dna5};
-    std::vector<unsigned> cmp{0,1,4,4,4,2,0,4,0};
+    std::vector<uint8_t> cmp{0,1,4,4,4,2,0,4,0};
```
smehringer:
```suggestion
    std::vector<dna5::rank_type> cmp{0,1,4,4,4,2,0,4,0}; // rank_type == uint8_t for dna5
```

PR #1039, opened 2019-06-03
cigar alphabet

h-2:


include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,144 @@
...
+     */
+    //!\brief Return the character representation.
+    small_string<11> to_char() const noexcept
+    {
+        small_string<11> ret{};
+        ret.resize(11);
+
+        auto [ ptr, errc ] = std::to_chars(ret.data(), ret.data() + 10, seqan3::to_rank(get<0>(*this)));
+
+        (void*)errc;
```
joergi-w: use `get<0>` above to avoid this?

h-2: It's not a pair...

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,145 @@
...
+    /*!\name Read functions
+     * \{
+     */
+    //!\brief Return the character representation.
+    small_string<11> to_char() const noexcept
+    {
+        small_string<11> ret{};
+        ret.resize(11);
+
+        auto [ ptr, errc ] = std::to_chars(ret.data(), ret.data() + 10, seqan3::to_rank(get<0>(*this)));
```
smehringer: why to_rank? if I have a tuple (5, 'M') I want it to print "5M", so I can directly take the count value without the rank?

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,145 @@
...
+    using base_t::operator<;
+    using base_t::operator>;
+
+    /*!\name Read functions
+     * \{
+     */
+    //!\brief Return the character representation.
+    small_string<11> to_char() const noexcept
+    {
+        small_string<11> ret{};
```
smehringer:
```suggestion
        small_string<11> ret{}; // maximum number of digits for uint32_t + 1 char for the cigar_op
```

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -0,0 +1,145 @@
...
+    //!\}
+
+    /*!\name Write functions
+     * \{
+     */
+    //!\brief Assign from the character representation.
+    cigar & assign_char(small_string<11> const s) noexcept
+    {
+        uint32_t num{};
+        auto [ ptr, errc ] = std::from_chars(s.data(), s.data() + 10, num);
```
smehringer: I think you can directly pass `get<0>(*this)` because std::from_chars does not modify the value on error

smehringer: Not resolved?

h-2: `get<0>` does not return a `uint32_t &`, but a proxy. std::from_chars cannot handle the proxy

h-2:
> LGTM but the code coverage says you are not testing one of your new overloads for 

As I siad, I had to re-add it, because it actually does.

PR #1041, opened 2019-06-03
fix stream_view

h-2:


h-2:
>  LGTM except that the fix is missing :D

I must have forgotten that 😉

PR #1045, opened 2019-06-04
[MISC] un-expose rank_type and char_type; not part of alphabet anymore

h-2:
also remove GCC BUG workaround on GCC >=8 because it was fixed.

In before the release because we don't want to expose stuff as API that won't be later on.


h-2:
Good point, I forgot about those.

PR #1046, opened 2019-06-04
Feature/parallel alignment interface

rrahn:
This PR adds the parallel interface to the align pairwise function and all its dependencies.

resolves first part of #889

##### TODOs that are still open
- [x] docu

##### TODOs that can be done later:
- [x] benchmark
- [ ] more tests for collections


include/seqan3/alignment/pairwise/alignment_configurator.hpp
```diff
@@ -273,12 +273,17 @@ struct alignment_configurator
             using first_seq_t = std::tuple_element_t<0, value_type_t<std::remove_reference_t<sequences_t>>>;
             using second_seq_t = std::tuple_element_t<1, value_type_t<std::remove_reference_t<sequences_t>>>;
```
h-2:
```suggestion
             using second_seq_t = std::tuple_element_t<1, value_type_t<sequences_t>>;
```

include/seqan3/alignment/pairwise/alignment_configurator.hpp
```diff
@@ -273,12 +273,17 @@ struct alignment_configurator
...
 
+            using wrapped_first_t  = decltype(std::declval<first_seq_t &>()  | view::all);
+            using wrapped_second_t = decltype(std::declval<second_seq_t &>() | view::all);
```
h-2:
```suggestion
            using wrapped_second_t = all_view<second_seq_t &>;
```

include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
```diff
@@ -176,14 +177,19 @@ class alignment_executor_two_way
...
         {
             auto && [tpl, idx] = *resource_iter;
             auto && [first_seq, second_seq] = tpl;
-            exec_handler.execute(kernel, idx, first_seq, second_seq, [this](auto && res){ *gptr = std::move(res); });
+            buffer_pointer write_to = gptr;
+            exec_handler.execute(kernel,
+                                 idx,
+                                 first_seq | view::all,
+                                 second_seq | view::all,
+                                 [write_to](auto && res){ *write_to = std::move(res); });
```
h-2:
```suggestion
                                 [write_to] (auto && res) { *write_to = std::move(res); });
```

:nail_care:

test/unit/alignment/pairwise/execution/alignment_executor_two_way_test.cpp
```diff
@@ -35,11 +35,11 @@ struct dummy_alignment
 };
 
 // Some globally defined test types
+using seq_type = decltype(std::declval<std::string &>() | view::all);
```
h-2:
```suggestion
using seq_type = all_view<std::string &>;
```

include/seqan3/alignment/pairwise/execution/execution_handler_parallel.hpp
```diff
@@ -0,0 +1,169 @@
...
+
+    /*!\name Execution
+     * \{
+     */
+    /*!\brief Invokes the passed alignment instance in a non-blocking manner.
+     * \copydetails execution_handler_sequential::execute
+     */
+
+     // * \tparam fn_type           The callable that needs to be invoked; must model std::Invocable with first_range_type
+     // *                           and second_range_type.
```
h-2: what happened to the documentation here?

include/seqan3/alignment/pairwise/execution/execution_handler_parallel.hpp
```diff
@@ -0,0 +1,169 @@
...
+                 std::Invocable<delegate_type, std::invoke_result_t<fn_type,
+                                                                    size_t const,
+                                                                    first_range_type,
+                                                                    second_range_type>>
+    //!\endcond
+    void execute(fn_type && func,
+                 size_t const idx,
+                 first_range_type && first_range,
+                 second_range_type && second_range,
+                 delegate_type && delegate)
```
h-2: are the ranges here views now too?

include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
```diff
@@ -44,7 +46,8 @@ template <std::ranges::ViewableRange resource_t,
           typename alignment_algorithm_t,
           typename execution_handler_t = execution_handler_sequential>
 //!\cond
-    requires std::ranges::InputRange<resource_t> &&
+    requires ((std::ranges::InputRange<resource_t> && std::ranges::SizedRange<resource_t>) ||
+             std::ranges::ForwardRange<resource_t>) &&
```
h-2: Are there pure input ranges that are sized? Why was it necessary to change this?

test/unit/alignment/pairwise/pairwise_alignment_collection_test_template.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+TYPED_TEST_P(pairwise_alignment_collection_test, back_coordinate)
+{
+    auto const & fixture = this->fixture();
+    configuration align_cfg = fixture.config | align_cfg::result{with_back_coordinate};
+    auto [database, query] = fixture.get_sequences();
+    auto res_vec = align_pairwise(typename TestFixture::policy_t{}, std::view::zip(database, query), align_cfg)
+                 | std::ranges::to<std::vector>;
+
+    EXPECT_TRUE((std::ranges::equal(res_vec | view::all
```
h-2: you usually don't need to call view::all explicitly or do you want our view::all to be called instead of std::view::all here?

include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
rrahn: Well I had once the idea to also allow a generating views or something like this to get the sequences. but for the parallel interface I do need the Size information before. Otherwise I would consume an InputRange by calling std::ranges::distance. This would ensure that it works as well. But I also agree that we are talking about very rare use cases and from that point it would be simply safer to require ForwardRange. I will change this.

test/unit/alignment/pairwise/pairwise_alignment_collection_test_template.hpp
rrahn: in general yes. I assumed to always prefer this as we get type erasure. 
For the test I had to materialize the alignment result view. Hence, I can not put it into another view before forwarding it as ref view.

include/seqan3/alignment/pairwise/execution/execution_handler_parallel.hpp
rrahn: yes. I wanted to check this if this works well in doxygen. Haven't tested it yet.

include/seqan3/alignment/pairwise/execution/execution_handler_parallel.hpp
rrahn: They are potentially always views. I could copy them instead getting them by forwarding reference.

test/unit/alignment/pairwise/pairwise_alignment_collection_test_template.hpp
rrahn: Forgot about it!!! I was completely confused!!! I need more sleep 😱 !

include/seqan3/alignment/pairwise/execution/execution_handler_parallel.hpp
h-2: Then again, views might be move-only in the future...

rrahn: yes, but I always take as non-reference so the caller needs to move into this, which is by the current design ensured. I do assert that the types are indeed views.

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -23,50 +23,60 @@
...
- * This function computes the pairwise alignment of a sequence pair or a range over sequence pairs.
- * The pairs can be any class template that models the seqan3::TupleLike.
- * In case of calling the alignment with a single tuple, the value is passed as a std::view::single to the range
- * aware interface. In this case the passed class must be std::CopyConstructible and both element types of the tuple
- * must model std::ranges::ViewableRange because it is copied into the std::view::single. The same
- * is true, if the input range is a temporary non-view range.
+ * This function computes the pairwise alignment for the given sequences. During the setup phase the most efficient
+ * implementation is selected depending on the configurations stored in the given seqan3::configuration object.
+ * In addition, the alignments can be computed in parallel if std::execution::par is given as the execution policy.
+ * On default the algorithm uses the sequenced execution (std::execution::seq).
```
h-2: adjust this. Also I would speak of parallel and "non-parallel" or "serial", not "sequenced", that's a lot less known.

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -23,50 +23,60 @@
...
+ *
+ * ### Compute a single alignment
+ *
+ * In some cases if only a single alignment needs to be computed, the two sequences can be passed as a pair.
```
h-2:
```suggestion
 * In cases where only a single alignment is to be computed, the two sequences can be passed as a pair.
```

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -23,50 +23,60 @@
...
+ * The pairs can be any class template that models the seqan3::TupleLike concept. The tuple elements must model
```
h-2:
```suggestion
 * The pair can be any class template that models the seqan3::TupleLike concept. The tuple elements must model
```

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -23,50 +23,60 @@
...
+ * std::ranges::ViewableRange and std::CopyConstructible as the tuple is converted into a range using std::view::single
```
h-2:
```suggestion
 * std::ranges::ViewableRange and std::CopyConstructible. 
```

The rest is detail and it's not necessary for users to know why I would say.

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -23,50 +23,60 @@
...
+ * (by copying/moving the sequence pair) and forwarded to the range aware interface.
  * Accordingly, the following example wouldn't compile:
  * ```cpp
  * std::pair p{"ACGTAGC", "AGTACGACG"};
  * auto rng = align_pairwise(p, config);
```
h-2: Does this actually work with string literals? Can you double-check?

h-2: (make a snippet out of this)

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -23,50 +23,60 @@
...
  * ```
  * You can use std::tie instead as shown in the example below.
  *
+ * ### Compute multiple alignments
+ *
+ * In many cases one needs to compute multiple pairwise alignments. Accordingly, the align_pairwise interface allows
+ * to pass a range over sequence pairs. The alignment algorithm will be configured only once for all submitted
+ * alignments and then computes the alignments sequenced or in parallel depending on the given execution policy.
+ *
```
h-2: Add:;
```
Since there is always a certain amount of initial setup required, it is advisable to pass many sequences / sequence-pairs to this algorithm instead of repeatedly calling it with a single pair.
```

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -23,50 +23,60 @@
...
  * ### Accessing the alignment results
  *
  * For each sequence pair one or more \ref seqan3::alignment_result "seqan3::alignment_result"s can be computed.
  * The seqan3::align_pairwise function returns an seqan3::alignment_range which can be used to iterate over the
- * alignments. The alignments are then computed on-demand when iterating over the results.
+ * alignments. If the sequenced execution policy is used the alignments are computed on-demand when iterating over the
+ * results. In case of a parallel execution all alignments are computed at once in parallel when calling `begin`
+ * on the associated seqan3::alignment_range.
```
h-2: Really? I thought this was handled by the queue.

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -134,23 +168,34 @@ constexpr auto align_pairwise(sequence_t && seq, alignment_config_t const & conf
                   std::ranges::ViewableRange<std::tuple_element_t<1, std::remove_reference_t<sequence_t>>>,
                   "Alignment configuration error: The tuple elements must model std::ranges::ViewableRange.");
 
-    return align_pairwise(std::view::single(std::forward<sequence_t>(seq)), config);
+    return align_pairwise(seqan3::seq, std::forward<sequence_t>(seq), config);
```
h-2: no view-wrapping anymore?

include/seqan3/alignment/pairwise/execution/execution_handler_parallel.hpp
```diff
@@ -0,0 +1,167 @@
...
+ *
+ * \note Instances of this class are not copyable.
+ *
+ * \attention This class cannot be reused for multiple calls. For this to work, it requires barriers and a queue that
+ *            can be reopened.
+ */
+class execution_handler_parallel
+{
+private:
+
```
h-2:
```suggestion
```

:nail_care:

include/seqan3/alignment/pairwise/execution/execution_handler_parallel.hpp
```diff
@@ -0,0 +1,167 @@
...
+    //!\brief The type erased task type.
+    using task_type = std::function<void()>;
+
+public:
+
```
h-2:
```suggestion
```

include/seqan3/alignment/pairwise/align_pairwise.hpp
rrahn: This is actually a counter example. This won't compile as stated above. So I cannot make it a snippet.

include/seqan3/alignment/pairwise/align_pairwise.hpp
rrahn: This function delegates to the exec aware interface which does now the view wrapping

include/seqan3/alignment/pairwise/align_pairwise.hpp
rrahn: This is due to some limitations we have right now with the queue and the synchronisation points. Either we always create a new execution handler and spawn the threads again on every invocation of underflow on the respective result buffer or what we actually need is a queue that can be reopened together with a barrier implementation. So as I said, in the near future the internal structures should resolve to a ring buffer which can invoke the parallel execution multiple times. But for now, this must do it.

include/seqan3/alignment/pairwise/align_pairwise.hpp
h-2: Wow, that's unhelpful. I thought we agreed that we didn't want to have "This won't work: " things...

include/seqan3/alignment/pairwise/execution/execution_handler_parallel.hpp
h-2: And now you just deleted it?

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -23,50 +23,61 @@
...
- * This function computes the pairwise alignment of a sequence pair or a range over sequence pairs.
- * The pairs can be any class template that models the seqan3::TupleLike.
- * In case of calling the alignment with a single tuple, the value is passed as a std::view::single to the range
- * aware interface. In this case the passed class must be std::CopyConstructible and both element types of the tuple
- * must model std::ranges::ViewableRange because it is copied into the std::view::single. The same
- * is true, if the input range is a temporary non-view range.
+ * This function computes the pairwise alignment for the given sequences. During the setup phase the most efficient
+ * implementation is selected depending on the configurations stored in the given seqan3::configuration object.
+ * In addition, the alignments can be computed in parallel if seqan3::par is given as the execution policy.
+ * On default the algorithm uses sequential execution (seqan3::seq).
```
h-2:
```suggestion
 * By default the algorithm uses sequential execution (seqan3::seq).
```

PR #1048, opened 2019-06-05
Fixing doc with wrong concept names

MitraDarja:
Fixing documentation with old concept names.


marehr:
we still have `std::input_Iterator` -> `std::InputIterator` and `std::output_Iterator` -> `std::OutputIterator`

PR #1051, opened 2019-06-05
[MISC] Require random access and sized range for pairwise alignment kernel.

rrahn:


marehr:
@rrahn a rebase is needed

PR #1060, opened 2019-06-05
[FIX] avoid -fsanitizer=undefined false-positive and correct compatibility 2 test.

smehringer:


test/unit/range/view/view_istreambuf_test.cpp
```diff
@@ -47,17 +47,16 @@ TEST(view_istreambuf, basic)
...
+    EXPECT_TRUE(std::ranges::equal(v3, comp));
 
     // combinability 2
     is.clear();
     is.seekg(0, std::ios::beg);
     auto v4 = view::istreambuf(is) | view::take_until(is_space);
     std::string out2 = v4;
     std::string comp2 = "ACGTATATATAT";
-    EXPECT_TRUE(std::ranges::equal(out, comp));
+    EXPECT_TRUE(std::ranges::equal(out2, comp2));
```
h-2: haha :clinking_glasses:

marehr: Is this a different fix?

test/unit/range/view/view_istreambuf_test.cpp
```diff
@@ -47,17 +47,16 @@ TEST(view_istreambuf, basic)
     is.clear();
     is.seekg(0, std::ios::beg);
     auto v3 = view::istreambuf(is) | view::char_to<dna5> | view::complement;
-    std::vector<dna5> out = v3;
     std::vector<dna5> comp{"TGCATATATATANTATATANAATNNNTATATT"_dna5};
-    EXPECT_TRUE(std::ranges::equal(out, comp));
+    EXPECT_TRUE(std::ranges::equal(v3, comp));
```
marehr: This helped?!

test/unit/range/view/view_istreambuf_test.cpp
h-2: seems like it

smehringer: Yes :laughing: 
I was really surprised when I made the first change and got an error that out is not known ^^

test/unit/range/view/view_istreambuf_test.cpp
smehringer: Yup... Because now there is no explicit conversion to `std::vector<dna5>` again.

PR #1066, opened 2019-06-06
[DOC] Some documentation fixes.

smehringer:


include/seqan3/io/record.hpp
```diff
@@ -31,7 +31,34 @@ namespace seqan3
  *
  * Some of the fields are shared between formats.
  *
- * \todo Table which files take which fields
+ * The following table shows the usage of fields in the respective files
+ * (Note that each valid format for a file must handle all of its fields):
+ *
+ * | Field          | Sequence IO | Alignment IO | Structure IO |
+ * | -------------- | ----------- | ------------ | ------------ |
+ * | SEQ            |      X      |      X       |       X      |
```
rrahn: We used to use ✅ sign in other places as well. I suggest to keep it consistent. 💅

smehringer: Would you make suggestions I can add? I will be in the office around 1pm.

smehringer: fine I did it myself

PR #1083, opened 2019-06-07
streamline alphabet tests

marehr:
This PR makes some changes that are inconsistent within the alphabet test cases.

* Add test case `type_properties`
* Correct `WritableQualityAlphabet` for references of quality alphabets
* Fix some small naming and include issues within mask alphabets
* Use `concept_check` as test name for concepts that the type models
* Cigar now uses `alphabet_test_template`
* Give a upper limit of iterations within `assign_rank`, `assign_char`
* And some other stuff


test/unit/alphabet/aminoacid/aminoacid_test_template.hpp
```diff
@@ -16,15 +16,24 @@
...
+    EXPECT_TRUE(AminoacidAlphabet<TypeParam const>);
+    EXPECT_TRUE(AminoacidAlphabet<TypeParam const &>);
+}
+
+TYPED_TEST_P(aminoacid, type_properties)
+{
+    EXPECT_TRUE((std::is_trivially_copyable_v<TypeParam>));
+    EXPECT_FALSE((std::is_trivially_default_constructible_v<TypeParam>));
+    EXPECT_FALSE((std::is_trivial_v<TypeParam>));
+    EXPECT_TRUE((std::is_standard_layout_v<TypeParam>));
```
h-2: Hm... I think we could just add the following to the template:
```cpp
    EXPECT_TRUE((std::is_trivially_copyable_v<TypeParam>));
    EXPECT_TRUE((std::is_standard_layout_v<TypeParam>));
```

Why should we enforce that they are not trivially-default-constructible?

marehr: I don't want to enforce that, but I want to know what types have what properties and most are not trivial. If some type changes to trivial, it would be better.

h-2: >  I want to know what types have what properties

But which properties do you mean? In the documentation we state that:

```
It is highly recommended that non-reference types that model this concept, also model:
    std::Regular
    std::is_trivially_copyable
    seqan3::StandardLayout
```

So it might make sense to check those concepts. Regular and StandardLayout can be checked in the alphabet_test_template. trivially_copyable also except that alphabet_proxy needs an exception (I am not sure if there even is a test for alphabet_proxy inside test_template so maybe it doesn't even need an exception). 

But why do you want to check trivial? No code depends on this, it doesn't influence anything.

> If some type changes to trivial,

How would a type "change"? Why does it matter? Also I can promise that with the current design, anything that uses aphabet_base cannot be trivial ever.

h-2:
Regarding the extra concept checks, see above. The other changes look good :+1:  

But let us please discuss current project priorities and TODOs for the next release together at a meeting and do less random refactoring.

smehringer:
@marehr @h-2 Shall I review this or is there something to discuss at the next DEV meeting?

marehr:
@smehringer This can now be reviewed.

marehr:
> Masked and Cigar are still missing from the tuple tests, otherwise LGTM!

This is more work than expected. I would postpone this.

smehringer:
>>Masked and Cigar are still missing from the tuple tests, otherwise LGTM!

>This is more work than expected. I would postpone this.

Did you add issues to do that? We need to test them :o

test/unit/alphabet/adaptation/uint_test.cpp
```diff
@@ -85,7 +91,7 @@ TYPED_TEST(uint_adaptation, assign_char)
     EXPECT_EQ((assign_char_to('C', l)), TypeParam{67});
 }
 
-TYPED_TEST(uint_adaptation, assign_char_strict)
+TYPED_TEST(uint_adaptation, assign_char_strictly_to)
```
smehringer: assign_rank and assign_char -> assign_rank_to and assign_char_to
:nail_care: since you changed this for the other test too

test/unit/alphabet/alphabet_test_template.hpp
```diff
@@ -194,33 +234,7 @@ TYPED_TEST_P(alphabet, comparison_operators)
...
-    EXPECT_TRUE(Alphabet<TypeParam const &>);
-
-    EXPECT_FALSE(WritableAlphabet<TypeParam const>);
-    EXPECT_FALSE(WritableAlphabet<TypeParam const &>);
-}
-
-REGISTER_TYPED_TEST_CASE_P(alphabet, alphabet_size_, default_value_constructor, global_assign_rank, global_to_rank,
-    copy_constructor, move_constructor, copy_assignment, move_assignment, swap, global_assign_char, global_to_char,
-    global_char_is_valid_for, global_assign_char_strict, comparison_operators, concept_check);
+REGISTER_TYPED_TEST_CASE_P(alphabet, concept_check, type_properties, alphabet_size_, default_value_constructor,
```
smehringer: just a wish not a review blocker. if each registered test is in one line, the diff would easier to spot next time :D

test/unit/alphabet/cigar/cigar_test.cpp
```diff
@@ -9,10 +9,10 @@
 
 #include <seqan3/alphabet/cigar/cigar.hpp>
 
-// does not include alphabet_test_template because that makes assumptions not covered by the concepts
-//TODO add alphabet_tuple tests when there
+#include "../alphabet_test_template.hpp"
 
-using namespace seqan3;
+INSTANTIATE_TYPED_TEST_CASE_P(cigar, alphabet, cigar);
+// note: cigar is not usable in constexpr context.
```
smehringer: actually, why not?

marehr:
Can you add one please? Thank you! 

Am 25. Juni 2019 08:05:20 MESZ schrieb smehringer <notifications@github.com>:

smehringer:
:roll_eyes:

PR #1085, opened 2019-06-10
[INFRA] make every seqan3 header self-contained; no exceptions anymore

marehr:


test/header/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@
 # shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
 # -----------------------------------------------------------------------------------------------------
 
-cmake_minimum_required (VERSION 3.2)
+cmake_minimum_required (VERSION 3.6)
```
rrahn: Have you checked that this version is compliant with all platforms that also support at least gcc-7?

test/header/CMakeLists.txt
```diff
@@ -19,11 +19,11 @@ option (SEQAN3_FULL_HEADER_TEST "Test seqan3 headers as well as the headers of e
 # header guards are working by including the same header twice.
 #
 # example invocation:
-#     seqan3_header_test (seqan3 "<path>/include/seqan3")
+#     seqan3_header_test (seqan3 "<path>/include/seqan3" "/exclude_regex|more_exclude/")
```
rrahn: I don't understand the example. 
Should more_exclude mean I can add more exclude patterns?
Then I suggest:
```suggestion
#     seqan3_header_test (seqan3 "<path>/include/seqan3" "")  - Nothing will be excluded
#     seqan3_header_test (seqan3 "<path>/include/seqan3" "/exclude_regex/[|more_excludes...]")  - separate multiple exludes with |
```
or can I only specify one more exlude?

test/header/CMakeLists.txt
marehr: I upped the version, which we already implicitly required. Because we already use the following feature which is available since 3.6

```cmake
list (FILTER header_files EXCLUDE REGEX "${exclude_regex}")
```

Since all builds go through since forever, this should have no impact

test/header/CMakeLists.txt
marehr: no you can give any regex

include/seqan3/std/charconv_detail.hpp
```diff
@@ -66,6 +66,41 @@
 #include <seqan3/core/concept/core_language.hpp>
 #include <seqan3/std/concepts>
 
+// -----------------------------------------------------------------------------
+// structs: to_chars_result, from_chars_result and chars_format
+// -----------------------------------------------------------------------------
+
+namespace std
```
smehringer: Now non-detail stuff is in a detail folder o.O 
I deliberately separated the "public API stuff I copied" from the "detail stuff to make public API work that I messed with"

test/header/CMakeLists.txt
smehringer: did you update the documentation?

marehr: No, I didn't update the documentation. Or, do we want to set the minimal required version of cmake to 3.6? I'm happy to do that, but that would be out the scope of the PR.

I added a card for the dev meeting to discuss a general increase of cmake version.

include/seqan3/std/charconv_detail.hpp
marehr: That's mainly why I assigned you.

The problem is if we don't do it this way, we get weird include order issues. Because, `charconv_detail.hpp` needs stuff from `charconv.hpp`, but on the other hand `charconv.hpp` needs `charconv_detail.hpp`.

If you have an idea how to keep the current structure with the separation of the public and private API, but without this dependency loop. I'm happy to change this. The documentation will still be public even if it defined within the `detail` file.

smehringer: Where do we get these weird issues? I thought the include order is clearly defined
```
// some code that is needed in detail.hpp
#include <detail.hpp>
 // some code that is dependent on detail.hpp
```
I thought this is the way includes work? The only thing is that detail.hpp cannot be independently included somewhere else, but it should not need to be as it is a detail header?

marehr: You are right, but like the title says no exceptions anymore, even our detail code should be includeable in any order.

smehringer: :( but why? when did we decide on that?

test/header/CMakeLists.txt
marehr: See core-developers chat.

include/seqan3/std/charconv_detail.hpp
marehr: We resolved this today at a quick strategy meeting, right?

test/header/CMakeLists.txt
marehr: I added more info, can you recheck?

test/header/CMakeLists.txt
marehr: I moved the fix of the cmake versions to PR #1094 , so we can discuss this over there

include/seqan3/std/charconv_detail.hpp
rrahn: Yes by merging the detail header into the non-detail header, right 🤔

marehr: done

marehr:
@rrahn can you have a second look and if everything is fine merge?

PR #1090, opened 2019-06-13
[MISC] Add 10s timeout to server calls.

### LINKED ISSUE #1091: version check server request timeout to long
The version checker uses default timeouts for wget curl or ftp, which leads to test timeouts if the server is not reachable. One could adapt the test to set a time limit on the future to wait for the result but in general I think the regular version check should not take longer than 10s. This also prevents the rare case that there is a thread consuming processor resources for 15 minutes or longer.

rrahn:
Currently, the version server is offline. This leads to timeouts of the version_debug|release_test.
In general I think the server request should not take longer than 10s. So I set a timeout on the request call.

fixes #1091

### DIFF
--- a/include/seqan3/argument_parser/detail/version_check.hpp
+++ b/include/seqan3/argument_parser/detail/version_check.hpp
@@ -429,15 +429,18 @@ private:
         return "powershell.exe -NoLogo -NonInteractive -Command \"& {Invoke-WebRequest -erroraction 'silentlycontinue' -OutFile";
 #else  // Unix based platforms.
         if (!system("wget --version > /dev/null 2>&1"))
-            return "wget -q -O";
+            return "wget --timeout=10 --tries=1 -q -O";
         else if (!system("curl --version > /dev/null 2>&1"))
-            return "curl -o";
-    #ifndef __linux  // ftp call does not work on linux
-        else if (!system("which ftp > /dev/null 2>&1"))
-            return "ftp -Vo";
-    #endif // __linux
-        else
+            return "curl --connect-timeout 10 -o";
+    // In case neither wget nor curl is available try ftp if system is OpenBSD or FreeBSD.
+    // Note, both systems have ftp command installed by default so we do not guard against it.
+    #if defined(__OpenBSD__) // In case
+            return "ftp -w10 -Vo";
+    #elseif defined(__FreeBSD__)
+            return "fetch --timeout=10 -o";
+    #else
             return "";
+    #endif // __OpenBSD__
 #endif  // defined(_WIN32)
     }
 
--- a/include/seqan3/argument_parser/detail/version_check.hpp
+++ b/include/seqan3/argument_parser/detail/version_check.hpp
@@ -102,7 +102,8 @@ public:
         // build up command for server call
         if (!program.empty())
         {
-            command = program + " " + cookie_path.string() + "/" + name + ".version " + server_url;
+            std::filesystem::path out_file = cookie_path / (name + ".version");
+            command = program + " " + out_file.string() + " " + server_url;
 #if defined(_WIN32)
             command = command + "; exit  [int] -not $?}\" > nul 2>&1";
 #else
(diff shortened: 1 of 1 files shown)


smehringer:
This should be done in one line by altering line 74/75 in version_check_test.hpp instead:

```cpp
        if (parser.version_check_future.valid())
            app_call_succeeded = parser.version_check_future.get();
```
to
```cpp
        if (parser.version_check_future.valid())
            if (parser.version_check_future.wait_for(std::chrono::seconds(10)) == std::future_status::ready)
                app_call_succeeded = parser.version_check_future.get();
```

Because that way the tests won't fail if there was a timeout since app_call_succeeded will be false (and the tests are aware of this). I designed it that way but forgot that future.get() will wait forever.

rrahn:
No, this only fixes the test. I did this on purpose to avoid spawning a thread for 15 minutes (default for wget) that does basically nothing except waiting. Even though one could argue that most thread schedulers should take care of this, it is simply not necessary to wait this long for a response. I think if the server is reachable 10s is sufficient.

include/seqan3/argument_parser/detail/version_check.hpp
```diff
@@ -98,12 +98,14 @@ class version_checker
         // build up command for server call
         if (!program.empty())
         {
-            command = program + " " + cookie_path.string() + "/" + name + ".version " + server_url;
+            command = program + " " + cookie_path.string() +
```
smehringer:
```suggestion
            command = program + " " + (cookie_path/name/ ".version ").string() + server_url;
```
?

rrahn:
@h-2 can I look at it and merge if ok. The builds are green although Travis is not updating the or status.

include/seqan3/argument_parser/detail/version_check.hpp
```diff
@@ -424,12 +425,12 @@ class version_checker
...
         if (!system("wget --version > /dev/null 2>&1"))
-            return "wget -q -O";
+            return "wget --timeout=10 --tries=1 -q -O";
         else if (!system("curl --version > /dev/null 2>&1"))
-            return "curl -o";
+            return "curl --connect-timeout 10 -o";
     #ifndef __linux  // ftp call does not work on linux
         else if (!system("which ftp > /dev/null 2>&1"))
-            return "ftp -Vo";
+            return "ftp -timeout 10 -Vo";
```
h-2: the ftp command doesn't have `-timeout` on OpenBSD and it cannot be used to make HTTP request so this definitely does not work. 
I would recommend the following:

```cpp
#elseif defined(__OpenBSD__)
    return "ftp -w10 -Vo";
#elseif defined(__FreeBSD__)
    return "fetch --timeout=10 -o";
#else
    return "";
#endif
```

h-2: (you don't need to check whether the ftp/fetch commands are available, they are guaranteed to be present on those systems.

h-2: The `ifndef linux` block needs to be removed now.

rrahn: I can assume ftp is available on those systems? 
I will add a comment though to make it clear that the code is fine as is.

include/seqan3/argument_parser/detail/version_check.hpp
```diff
@@ -429,15 +430,18 @@ class version_checker
...
+            return "wget --timeout=10 --tries=1 -q -O";
         else if (!system("curl --version > /dev/null 2>&1"))
-            return "curl -o";
-    #ifndef __linux  // ftp call does not work on linux
-        else if (!system("which ftp > /dev/null 2>&1"))
-            return "ftp -Vo";
-    #endif // __linux
-        else
+            return "curl --connect-timeout 10 -o";
+    // In case neither wget nor curl is available try ftp if system is OpenBSD or FreeBSD.
```
h-2:
```suggestion
    // In case neither wget nor curl is available try ftp/fetch on OpenBSD/FreeBSD.
    // which are always included with the respective base system.
```

include/seqan3/argument_parser/detail/version_check.hpp
```diff
@@ -429,15 +430,18 @@ class version_checker
...
+    // Note, both systems have ftp command installed by default so we do not guard against it.
+    #if defined(__OpenBSD__) // In case
+            return "ftp -w10 -Vo";
+    #elseif defined(__FreeBSD__)
+            return "fetch --timeout=10 -o";
```
h-2: indent the return statements less

include/seqan3/argument_parser/detail/version_check.hpp
```diff
@@ -429,15 +430,18 @@ class version_checker
...
-            return "curl -o";
-    #ifndef __linux  // ftp call does not work on linux
-        else if (!system("which ftp > /dev/null 2>&1"))
-            return "ftp -Vo";
-    #endif // __linux
-        else
+            return "curl --connect-timeout 10 -o";
+    // In case neither wget nor curl is available try ftp/fetch if system is OpenBSD/FreeBSD.
+    // Note, both systems have ftp command installed by default so we do not guard against it.
+    #if defined(__OpenBSD__) // In case
```
h-2: `// In case` ...?

PR #1093, opened 2019-06-14
[FIX] edit distance returns correct back coordinate

marehr:
@rrahn I already talked this over with you, that the general pairwise algorithm returns non-inclusive `back_coordinate`s and that the edit distance returns inclusive `back_coordinate`s.

It is desirable to have non-inclusive back coordinates and inclusive front coordinates, because with that one can express empty intervals. This PR makes `back_coordinate`s non-inclusive and resolves a discrepancy between the general and edit alignment algorithm.

As an example for non-inclusive:
https://github.com/seqan/seqan3/blob/81a1cb9f7ec99d1f87b88213bc959bb5b81f416a/test/unit/alignment/pairwise/fixture/global_affine_banded.hpp#L48-L58

And an example for inclusive:
https://github.com/seqan/seqan3/blob/81a1cb9f7ec99d1f87b88213bc959bb5b81f416a/test/unit/alignment/pairwise/fixture/global_edit_distance_unbanded.hpp#L26-L40

### DIFF
--- a/include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
+++ b/include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
@@ -273,11 +273,11 @@ class edit_distance_unbanded_global_policy :
         _best_score = self->_score;
     }
 
-    //!\copydoc edit_distance_unbanded_semi_global_policy::_best_score_col
-    size_t best_score_column() const noexcept
+    //!\brief Returns the first component of the #back_coordinate.
+    size_t back_coordinate_first() const noexcept
     {
         derived_t const * self = static_cast<derived_t const *>(this);
-        return std::ranges::size(self->database) - 1u;
+        return std::ranges::size(self->database);
     }
     //!\}
 
@@ -309,8 +309,9 @@ public:
         if (!self->is_valid())
             return self->invalid_coordinate();
 
-        size_t col = self->best_score_column();
-        return {column_index_type{col}, row_index_type{std::ranges::size(self->query) - 1u}};
+        column_index_type const first{self->back_coordinate_first()};
+        row_index_type const second{std::ranges::size(self->query)};
+        return {first, second};
     }
     //!\}
 };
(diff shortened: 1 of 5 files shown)


include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
```diff
@@ -406,7 +406,7 @@ class edit_distance_unbanded_semi_global_policy :
     size_t best_score_column() const noexcept
     {
         derived_t const * self = static_cast<derived_t const *>(this);
-        return std::ranges::distance(std::ranges::begin(self->database), _best_score_col);
+        return std::ranges::distance(std::ranges::begin(self->database), _best_score_col) + 1u;
```
joergi-w: That means `_best_score_col` is still the column of the inclusive back coordinate, right?
But you define `best_score_column()` as the column of the exclusive back coordinate.

I suggest you either let `_best_score_col` contain the exclusive coordinate value as well or rename it to distinguish the different meanings.

include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
```diff
@@ -277,7 +277,7 @@ class edit_distance_unbanded_global_policy :
     size_t best_score_column() const noexcept
     {
         derived_t const * self = static_cast<derived_t const *>(this);
-        return std::ranges::size(self->database) - 1u;
+        return std::ranges::size(self->database);
```
joergi-w: The documentation to this function (line 276) has become wrong now!
`//!\copydoc edit_distance_unbanded_semi_global_policy::_best_score_col` documents the inclusive coordinates. 

This mistake is copydoc'ed back to semi_global's `best_score_column()` function... Aside from the intransparent chain of copydoc here: In general I think it's legitimate that an inherited class copies documentation from its base, but the opposite direction should be avoided!

rrahn:
@marehr we need to update the naming as well. Because we choose back and front to refer to the inclusive versions. But now we should rather have begin and end to make clear that they behave like begin/end of ranges. This can be done later, but can you track this as an issue please. Especially, since this is an API breaking change. While we are add it, we should discuss the use cases for these coordinates and if the current design meets those requirements.

include/seqan3/alignment/pairwise/edit_distance_unbanded.hpp
marehr: thank you, that's a good idea. I'll name the function `back_coordinate_column`

marehr:
@joergi-w I changed the name from `best_score_column` to `back_coordinate_first` to reflect the semantic change. `back_coordinate_first` has now it's own documentation.

joergi-w:
http://docs.seqan.de/seqan/3-master-dev/structseqan3_1_1detail_1_1with__back__coordinate__type.html
could also need some information that the coordinate is exclusive.

marehr:
@joergi-w Opened #1112 as followup. To fix the general documentation.

Issue #424, opened 2018-09-12
newer cmake version give warnings on our cmake.config


h-2:
```
--   C++ Filesystem library:     via -lstdc++fs
CMake Warning (dev) at /usr/local/share/cmake/Modules/CheckIncludeFileCXX.cmake:69 (message):
  Policy CMP0075 is not set: Include file check macros honor
  CMAKE_REQUIRED_LIBRARIES.  Run "cmake --help-policy CMP0075" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  CMAKE_REQUIRED_LIBRARIES is set to:

    stdc++fs

  For compatibility with CMake 3.11 and below this check is ignoring it.
Call Stack (most recent call first):
  /home/mi/h4nn3s/devel/seqan3/build_system/seqan3-config.cmake:375 (check_include_file_cxx)
  /home/mi/h4nn3s/devel/seqan3/test/seqan3-test.cmake:41 (find_package)
```
would be nice to have this fixed before the workshop. @marehr can you have look?

PR #1094, opened 2019-06-14
updates cmake_minimum_required

### LINKED ISSUE #1095: settle minimal required cmake version for our tests
I just found out that we need to require
* `seqan3-config.cmake` min  cmake 3.4, otherwise cmake errors with `try_compile() works only for enabled languages.  Currently these are:    CXX`
* test/coverage needs min 3.4
* test/documentation needs min 3.7
  * 3.5:
    ```
    CMake Error at /usr/share/cmake-3.5/Modules/ExternalProject.cmake:1888 (message):
    Could not extract tarball filename from url:

    https://github.com/PeterFeicht/cppreference-doc/releases/download/v20190602/html-book-20190602.tar.xz
    ```
  * 3.6: 
    ```
    [ 50%] Performing configure step for 'download-cppreference-doxygen-web-tag'
    xzcat: html-book.tar.xz: No such file or directory
    tar: This does not look like a tar archive
    tar: cppreference-doxygen-web.tag.xml: Not found in archive
    tar: Exiting with failure status due to previous errors
    ```
* test/header needs min. 3.5.1 (because of test/performance and a different reason); if SEQAN3_FULL_HEADER_TEST=YES we need 3.6
* test/performance needs 3.5.1, because google benchmark requires it
* test/snippets needs 3.4
* test/unit needs 3.4

marehr:
* This fixes header tests that could not be build below cmake 3.12
* States all minimum required cmake versions and tests them in travis
* The minimum required cmake version, like the README.md says, is now 3.4.

Fixes #1095


h-2:
We discussed this again today. The consensus was:
  * add a small test that just includes the seqan3-config and uses cmake 3.4
  * change all other tests to use cmake-3.7

marehr:
> Do we have a test for the library config testing cmake 3,4?

Not yet. I discussed this with @h-2 and we decided to postpone this, because I had no good idea to do this.

I opened a follow-up issue #1134.

build_system/seqan3-config.cmake
```diff
@@ -71,7 +71,7 @@
 #
 # ============================================================================
 
-cmake_minimum_required (VERSION 3.2...3.12)
+cmake_minimum_required (VERSION 3.4...3.12)
```
h-2: is this a maximum version?

eseiler: Basically means: "We are aware of the policy changes introduced in cmake up to version 3.12 and accept the new behaviour of those policies. Please do not throw warnings for new policies."

Otherwise you'd get warnings about policies when compiling with cmake 3.12.

marehr: Yes, but no in the sense that you might thing. This still works with the most recent cmake version;.

Normally cmake warns you with policy warnings that some default behaviour changed and sometimes you need to set a policy to acknowledge that.

By using this syntax, we say that we tested our script for this version range and I think, that part I'm not to sure, that we acknowledge every policy change in that version range.

This also sets policies to the behaviour of 3.12, s.t. newer cmake versions behave like 3.12.

We introduced that to fix #424.

See https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html?highlight=minimum#command:cmake_minimum_required

h-2: that's cool :+1:

h-2:
Jenkins fails (this time correctly) because the cmake verison on jenkins is too old. I wonder how it managed to build stuff before. In any case, can you update cmake in the jenkins-script, too?

marehr:
cmake 3.5.1 was fine, because we did not build every test suite.

h-2:
> cmake 3.5.1 was fine, because we did not build every test suite.

Ah, ok. I thought Jenkins tested more stuff than travis.

.jenkinsfile
```diff
@@ -79,6 +79,13 @@ def run_unit_tests(slave_name, compiler, build_type, model, with_cereal)
     def workspace
     def platform
 
+    stage ('versions')
+    {
+        sh '$compiler --version'
+        sh 'cmake --version'
+        sh 'doxygen--version'
```
smehringer:
```suggestion
        sh 'doxygen --version'
```

PR #1105, opened 2019-06-18
Feature/alignment matrix

### LINKED ISSUE #1104: Refactor alignment matrix policy to use a unified interface.
Currently the alignment matrix policies directly expose the storage interface to the algorithm. Accordingly the alignment algorithm had to implement special verisons of the gap kernel to deal with the different matrix layouts. This should be encapsulated in a corresponding alignment matrix, which provides an unified interface to the algorithm. This step is necessary to reduce implement the vectorised kernel without copying the the several policies for the different matrix layouts.

rrahn:
Adds alignment matrices that give a unified access to the cells required by the standard alignment algorithm. I only had some testing of the basic functionality. The tests will follow soon.

Fixes #1104

### Todo 
- [x] tests

### Note

I will not update the alignment policies yet. I need first the traceback iterator interface. When this works, then I can adapt and reduce the policies.

### DIFF
--- a/include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column.hpp
+++ b/include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column.hpp
@@ -1,0 +1,163 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides seqan3::detail::alignment_score_matrix_one_column.
+ * \author Rene Rahn <rene.rahn AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <seqan3/alignment/matrix/detail/alignment_matrix_column_major_range_base.hpp>
+#include <seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_base.hpp>
+#include <seqan3/alignment/matrix/detail/alignment_score_matrix_proxy.hpp>
+#include <seqan3/core/concept/core_language.hpp>
+#include <seqan3/core/simd/concept.hpp>
+#include <seqan3/std/iterator>
+#include <seqan3/std/ranges>
+#include <seqan3/std/span>
+
+namespace seqan3::detail
+{
+
+/*!\brief An alignment score matrix storing only a single column for the computation.
+ * \tparam score_t The type of the score; must model either seqan3::Arithmetic or seqan3::detail::Simd.
+ * \ingroup alignment_matrix
+ *
+ * \details
+ *
+ * This implementation allocates only a single column for the alignment score matrix.
+ * The matrix allows access to the underlying values through a range based interface. An iterator over the score matrix
+ * iterates in column-major-order. Dereferencing an iterator returns a view over the current matrix column.
+ * The value type is seqan3::detail::alignment_score_matrix_proxy, which gives an unified access to the respective
+ * matrix cells as needed by the standard alignment algorithm. The matrix is modelled as std::ranges::InputRange since
+ * the alignment algorithm iterates only once over the complete matrix to calculate the values.
+ */
+template <typename score_t>
+class alignment_score_matrix_one_column :
+    protected alignment_score_matrix_one_column_base<score_t>,
+    public alignment_matrix_column_major_range_base<alignment_score_matrix_one_column<score_t>>
+{
+private:
+    static_assert(Arithmetic<score_t> || Simd<score_t>,
+                  "The score type must be either an arithmetic type or a simd vector type.");
+    //!\brief The base class for data storage.
+    using matrix_base_t = alignment_score_matrix_one_column_base<score_t>;
+    //!\brief The base class for iterating over the matrix.
+    using range_base_t = alignment_matrix_column_major_range_base<alignment_score_matrix_one_column<score_t>>;
+
+    //!\brief Befriend the range base class.
+    friend range_base_t;
+
+protected:
+    using typename matrix_base_t::element_type;
+    using typename range_base_t::alignment_column_type;
+    //!\copydoc seqan3::detail::alignment_matrix_column_major_range_base::column_data_view_type
+    using column_data_view_type = std::span<element_type>;
+
+    using matrix_base_t::num_cols;
+
+public:
+    /*!\name Associated types
+     * \{
+     */
+    //!\copydoc seqan3::detail::alignment_matrix_column_major_range_base::value_type
+    using value_type = alignment_score_matrix_proxy<score_t>;
+    //!\brief Same as value type.
+    using reference = value_type;
+    //!\copydoc seqan3::detail::alignment_matrix_column_major_range_base::iterator
+    using iterator = typename range_base_t::iterator;
+    //!\copydoc seqan3::detail::alignment_matrix_column_major_range_base::sentinel
+    using sentinel = typename range_base_t::sentinel;
+    using typename matrix_base_t::size_type;
+    //!\}
+
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    //!\brief Defaulted.
+    constexpr alignment_score_matrix_one_column() = default;
+    //!\brief Defaulted.
+    constexpr alignment_score_matrix_one_column(alignment_score_matrix_one_column const &) = default;
+    //!\brief Defaulted.
+    constexpr alignment_score_matrix_one_column(alignment_score_matrix_one_column &&) = default;
+    //!\brief Defaulted.
+    constexpr alignment_score_matrix_one_column & operator=(alignment_score_matrix_one_column const &) = default;
+    //!\brief Defaulted.
+    constexpr alignment_score_matrix_one_column & operator=(alignment_score_matrix_one_column &&) = default;
+    //!\brief Defaulted.
+    ~alignment_score_matrix_one_column() = default;
+
+    /*!\brief Construction from two ranges.
+     * \tparam first_sequence_t  The first range type; must model std::ranges::ForwardRange.
+     * \tparam second_sequence_t The second range type; must model std::ranges::ForwardRange.
+     *
+     * \param[in] first         The first range.
+     * \param[in] second        The second range.
+     * \param[in] initial_value The value to initialise the matrix with. Default initialised if not specified.
+     *
+     * \details
+     *
+     * Obtains only the sizes of the passed ranges in order to allocate the score matrix. Only one column
+     * is allocated.
+     */
+    template <std::ranges::ForwardRange first_sequence_t, std::ranges::ForwardRange second_sequence_t>
+    constexpr alignment_score_matrix_one_column(first_sequence_t && first,
+                                                second_sequence_t && second,
+                                                score_t const initial_value = score_t{})
+    {
+        matrix_base_t::num_cols = static_cast<size_type>(std::ranges::distance(first) + 1);
+        matrix_base_t::num_rows = static_cast<size_type>(std::ranges::distance(second) + 1);
+        matrix_base_t::pool.resize(matrix_base_t::num_rows + 1, element_type{initial_value, initial_value});
+    }
+    //!\}
+
+private:
+    //!\copydoc seqan3::detail::alignment_matrix_column_major_range_base::initialise_column
+    constexpr alignment_column_type initialise_column(size_type const SEQAN3_DOXYGEN_ONLY(column_index)) noexcept
+    {
+        return alignment_column_type{*this, column_data_view_type{std::addressof(matrix_base_t::pool[0]),
+                                                                  matrix_base_t::num_rows}};
+    }
+
+    //!\copydoc seqan3::detail::alignment_matrix_column_major_range_base::make_proxy
+    template <std::RandomAccessIterator iter_t>
+    constexpr value_type make_proxy(iter_t host_iter) noexcept
+    {
+        return {std::get<0>(*host_iter),            // current
+                std::get<0>(matrix_base_t::cache),  // last diagonal
+                std::get<1>(*host_iter),            // last left (read)
+                std::get<1>(*host_iter),            // next left (write)
+                std::get<2>(matrix_base_t::cache)}; // last up
+    }
+
+    //!\copydoc seqan3::detail::alignment_matrix_column_major_range_base::on_column_iterator_creation
+    template <std::RandomAccessIterator iter_t>
+    constexpr void on_column_iterator_creation(iter_t host_iter) noexcept
+    {
+        // Cache the next diagonal value.
+        std::get<1>(matrix_base_t::cache) = std::get<0>(*host_iter);
+    }
+
+    //!\copydoc seqan3::detail::alignment_matrix_column_major_range_base::before_column_iterator_increment
+    template <std::RandomAccessIterator iter_t>
+    constexpr void before_column_iterator_increment(iter_t SEQAN3_DOXYGEN_ONLY(host_iter)) noexcept
+    {
+        // Update the last diagonal value.
+        std::get<0>(matrix_base_t::cache) = std::move(std::get<1>(matrix_base_t::cache));
+    }
+
+    //!\copydoc seqan3::detail::alignment_matrix_column_major_range_base::after_column_iterator_increment
+    template <std::RandomAccessIterator iter_t>
+    constexpr void after_column_iterator_increment(iter_t host_iter) noexcept
+    {
+        // Cache the next diagonal value.
+        std::get<1>(matrix_base_t::cache) = std::move(std::get<0>(*host_iter));
+    }
+};
+
+} // namespace seqan3::detail
--- a/include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_banded.hpp
+++ b/include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_banded.hpp
@@ -1,0 +1,187 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides seqan3::detail::alignment_score_matrix_one_column_banded.
+ * \author Rene Rahn <rene.rahn AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <seqan3/alignment/band/static_band.hpp>
+#include <seqan3/alignment/matrix/detail/alignment_matrix_column_major_range_base.hpp>
+#include <seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_base.hpp>
+#include <seqan3/alignment/matrix/detail/alignment_score_matrix_proxy.hpp>
+#include <seqan3/std/iterator>
+#include <seqan3/std/ranges>
+#include <seqan3/std/span>
+
+namespace seqan3::detail
+{
+
+/*!\brief A banded alignment score matrix storing only a single banded column for the computation.
+ * \tparam score_t The type of the score; must model either seqan3::Arithmetic or seqan3::detail::Simd.
+ * \ingroup alignment_matrix
+ *
+ * \details
+ *
+ * This implementation allocates only a single banded column for the alignment score matrix.
+ * The matrix allows access to the underlying values through a range based interface. An iterator over the score matrix
+ * iterates in column-major-order. Dereferencing an iterator returns a view over the current matrix column.
+ * The value type is seqan3::detail::alignment_score_matrix_proxy, which gives an unified access to the respective
+ * matrix cells as needed by the standard alignment algorithm. The matrix is modelled as std::ranges::InputRange since
+ * the alignment algorithm iterates only once over the complete matrix to calculate the values.
+ */
+template <typename score_t>
+class alignment_score_matrix_one_column_banded :
+    protected alignment_score_matrix_one_column_base<score_t>,
+    public alignment_matrix_column_major_range_base<alignment_score_matrix_one_column_banded<score_t>>
+{
+private:
+    //!\brief The base class for data storage.
+    using matrix_base_t = alignment_score_matrix_one_column_base<score_t>;
+    //!\brief The base class for iterating over the matrix.
+    using range_base_t = alignment_matrix_column_major_range_base<alignment_score_matrix_one_column_banded<score_t>>;
+
+    //!\brief Befriend the range base class.
+    friend range_base_t;
+
+protected:
+    using typename matrix_base_t::element_type;
+    using typename range_base_t::alignment_column_type;
+    //!\copydoc seqan3::detail::alignment_matrix_column_major_range_base::column_data_view_type
+    using column_data_view_type = std::span<element_type>;
+
+    using matrix_base_t::num_cols;
+
+public:
+    /*!\name Associated types
+     * \{
+     */
+    //!\copydoc seqan3::detail::alignment_matrix_column_major_range_base::value_type
+    using value_type = alignment_score_matrix_proxy<score_t>;
+    //!\brief Same as value type.
+    using reference = value_type;
+    //!\copydoc seqan3::detail::alignment_matrix_column_major_range_base::iterator
+    using iterator = typename range_base_t::iterator;
+    //!\copydoc seqan3::detail::alignment_matrix_column_major_range_base::sentinel
+    using sentinel = typename range_base_t::sentinel;
+    using typename matrix_base_t::size_type;
+    //!\}
+
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    //!\brief Defaulted.
+    constexpr alignment_score_matrix_one_column_banded() = default;
+    //!\brief Defaulted.
+    constexpr alignment_score_matrix_one_column_banded(alignment_score_matrix_one_column_banded const &) = default;
+    //!\brief Defaulted.
+    constexpr alignment_score_matrix_one_column_banded(alignment_score_matrix_one_column_banded &&) = default;
+    //!\brief Defaulted.
+    constexpr alignment_score_matrix_one_column_banded &
+    operator=(alignment_score_matrix_one_column_banded const &) = default;
+    //!\brief Defaulted.
+    constexpr alignment_score_matrix_one_column_banded &
+    operator=(alignment_score_matrix_one_column_banded &&) = default;
+    //!\brief Defaulted.
+    ~alignment_score_matrix_one_column_banded() = default;
+
+    /*!\brief Construction from two ranges and a band.
+     * \tparam first_sequence_t  The first range type; must model std::ranges::ForwardRange.
+     * \tparam second_sequence_t The second range type; must model std::ranges::ForwardRange.
+     *
+     * \param[in] first          The first range.
+     * \param[in] second         The second range.
+     * \param[in] band           The seqan3::static_band in which to calculate the alignment.
+     * \param[in] initial_value  The value to initialise the matrix with. Default initialised if not specified.
+     *
+     * \details
+     *
+     * Does only obtain the sizes of the passed ranges in order to allocate the score matrix. Only allocates
+     * one column of the size of the band.
+     */
+    template <std::ranges::ForwardRange first_sequence_t,
+              std::ranges::ForwardRange second_sequence_t>
+    constexpr alignment_score_matrix_one_column_banded(first_sequence_t && first,
+                                                       second_sequence_t && second,
+                                                       static_band const & band,
+                                                       score_t const initial_value = score_t{})
+    {
+        matrix_base_t::num_cols = static_cast<size_type>(std::ranges::distance(first) + 1);
+        matrix_base_t::num_rows = static_cast<size_type>(std::ranges::distance(second) + 1);
+
+        band_col_index = std::min<int32_t>(std::max<int32_t>(band.upper_bound, 0), matrix_base_t::num_cols - 1);
+        band_row_index = std::min<int32_t>(std::abs(std::min<int32_t>(band.lower_bound, 0)),
+                                           matrix_base_t::num_rows - 1);
+        band_size = band_col_index + band_row_index + 1;
+
+        // Reserve one more cell to deal with last cell in the banded column which needs only the diagonal and up cell.
+        matrix_base_t::pool.resize(band_size + 1, element_type{initial_value, initial_value});
+    }
+    //!\}
+
+private:
+    //!\brief The column index where the upper bound of the band passes through.
+    int32_t band_col_index{};
+    //!\brief The row index where the lower bound of the band passes through.
+    int32_t band_row_index{};
+    //!\brief The size of the band.
+    int32_t band_size{};
+
+    //!\copydoc seqan3::detail::alignment_matrix_column_major_range_base::initialise_column
+    constexpr alignment_column_type initialise_column(size_type const column_index) noexcept
+    {
+        int32_t slice_begin = std::max<int32_t>(0, band_col_index - column_index);
+        int32_t row_end_index = column_index - band_col_index + band_size;
+        int32_t slice_end = band_size - std::max<int32_t>(row_end_index - matrix_base_t::num_rows, 0);
+
+        assert(row_end_index >= 0);
+        assert(slice_begin >= 0);
+        assert(slice_end > 0);
+        assert(slice_begin < slice_end);
+
+        return alignment_column_type{*this, column_data_view_type{std::addressof(matrix_base_t::pool[slice_begin]),
+                                                                  std::addressof(matrix_base_t::pool[slice_end])}};
+    }
+
+    //!\copydoc seqan3::detail::alignment_matrix_column_major_range_base::make_proxy
+    template <std::RandomAccessIterator iter_t>
+    constexpr value_type make_proxy(iter_t host_iter) noexcept
+    {
+        return {std::get<0>(*host_iter),             // current
+                std::get<0>(matrix_base_t::cache),   // last diagonal
+                std::get<1>(*(host_iter + 1)),       // last left (read)
+                std::get<1>(*(host_iter)),           // next left (write)
+                std::get<1>(matrix_base_t::cache)};  // last up
+    }
+
+    //!\copydoc seqan3::detail::alignment_matrix_column_major_range_base::on_column_iterator_creation
+    template <std::RandomAccessIterator iter_t>
+    constexpr void on_column_iterator_creation(iter_t host_iter) noexcept
+    {
+        // Cache the last diagonal value.
+        std::get<0>(matrix_base_t::cache) = std::get<0>(*host_iter);
+    }
+
+    //!\copydoc seqan3::detail::alignment_matrix_column_major_range_base::before_column_iterator_increment
+    template <std::RandomAccessIterator iter_t>
+    constexpr void before_column_iterator_increment(iter_t SEQAN3_DOXYGEN_ONLY(host_iter)) noexcept
+    {
+        // noop.
+    }
+
+    //!\copydoc seqan3::detail::alignment_matrix_column_major_range_base::after_column_iterator_increment
+    template <std::RandomAccessIterator iter_t>
+    constexpr void after_column_iterator_increment(iter_t host_iter) noexcept
+    {
+        // Cache the last diagonal value.
+        std::get<0>(matrix_base_t::cache) = std::get<0>(*host_iter);
+    }
+};
+
+} // namespace seqan3::detail
--- a/include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_base.hpp
+++ b/include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_base.hpp
@@ -1,0 +1,61 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides seqan3::detail::alignment_score_matrix_one_column_base.
+ * \author Rene Rahn <rene.rahn AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <array>
+#include <utility>
+#include <vector>
+
+#include <seqan3/core/concept/core_language.hpp>
+#include <seqan3/core/simd/concept.hpp>
+#include <seqan3/range/container/aligned_allocator.hpp>
+
+namespace seqan3::detail
+{
+
+/*!\brief A base class for alignment score matrices using only one column to compute the matrix.
+ * \tparam score_t The type of the scores.
+ * \ingroup alignment_matrix
+ *
+ * \details
+ *
+ * Manages the actual storage as a std::vector. How much memory is allocated is handled by the derived type.
+ * The `score_t` must either model the seqan3::Arithmetic or seqan3::detail::Simd vector concept.
+ */
+template <typename score_t>
+struct alignment_score_matrix_one_column_base
+{
+protected:
+    static_assert(Arithmetic<score_t> || Simd<score_t>,
+                  "Score type must either be either an arithmetic type or a simd vector type.");
+
+    //!\brief The actual element type.
+    using element_type = std::tuple<score_t, score_t>;
+    //!\brief The allocator type.
+    using allocator_type = aligned_allocator<element_type, sizeof(element_type)>;
+    //!\brief The type of the underlying storage.
+    using pool_type = std::vector<element_type, allocator_type>;
+    //!\brief The size type.
+    using size_type = size_t;
+public:
+    //!\brief The linearised memory pool storing only one column of the matrix.
+    pool_type pool{};
+    //!\brief Internal cache for the last diagonal and vertical value during the alignment computation.
+    std::array<score_t, 3> cache{};  // Third argument is used to cache next diagonal value in non-banded case.
+    //!\brief The number of columns.
+    size_type num_cols{};
+    //!\brief The number of num_rows.
+    size_type num_rows{};
+};
+
+} // namespace seqan3::detail
--- a/include/seqan3/alignment/matrix/detail/alignment_score_matrix_proxy.hpp
+++ b/include/seqan3/alignment/matrix/detail/alignment_score_matrix_proxy.hpp
@@ -1,0 +1,46 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides seqan3::detail::alignment_score_matrix_proxy.
+ * \author Rene Rahn <rene.rahn AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <seqan3/core/concept/core_language.hpp>
+#include <seqan3/core/simd/concept.hpp>
+
+namespace seqan3::detail
+{
+
+/*!\brief A proxy type for a unified access to the score matrix during alignment computation.
+ * \tparam score_type The type wrapped by this proxy.
+ * \ingroup alignment_matrix
+ *
+ * \details
+ *
+ * Provides named accessors to the respective values of the score matrix during the alignment computation.
+ * The `score_type` must either model the seqan3::Arithmetic or seqan3::detail::Simd vector concept.
+ * In some cases the last value to the left is stored in a different location then the next left value that
+ * will be used in the subsequent column (e.g. in the banded alignment matrix). Accordingly the left value is
+ * split in a reference for reading, referred to as `r_left` (for reading), and a reference to the next left value
+ * on the same row, referred to as `w_left` (for writing).
+ */
+template <typename score_type>
+struct alignment_score_matrix_proxy
+{
+    static_assert(Arithmetic<score_type> || Simd<score_type>,
+                  "Value type must either be an arithmetic type or a simd vector type.");
+
+    score_type & current; //!< The current value.
+    score_type & diagonal; //!< The last diagonal value.
+    score_type & r_left; //!< The last value to the left.
+    score_type & w_left; //!< The next value to the left.
+    score_type & up; //!< The last value above.
+};
+} // namespace seqan3::detail
--- a/include/seqan3/alignment/matrix/detail/alignment_trace_matrix_full.hpp
+++ b/include/seqan3/alignment/matrix/detail/alignment_trace_matrix_full.hpp
@@ -1,0 +1,173 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides seqan3::detail::alignment_trace_matrix_full.
+ * \author Rene Rahn <rene.rahn AT fu-berlin.de>
+ */
+
+#pragma once
+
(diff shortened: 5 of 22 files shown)


h-2:
Since you are working on a a lot stuff that is detail and you are writing lots of tests, maybe we should discuss again what our policy on testing is! I know that @smehringer and @marehr already had a disagreement about this.

My position is:

  * All private or `::detail` entities need to have some public, non-detail interface that tests them (otherwise they are not need for anything user-facing).
  * If you have 100% code coverage through such (indirect) tests, you do not need to add extra tests.

This follows the paradigm that you should test **observable behaviour** and not implementation detail. It can reduce the amount of tests that need writing significantly.
A side effect is that if you remove public code that uses the detail code, it will reduce coverage and alert you about "dead" detail code that can be removed.

I'll open a card for the next strategy meeting.

rrahn:
As far as I am concerned. I do unit testing. So what the name suggests I test interfaces decoupled from the rest of the interfaces. If that means testing a detail crtp base class by means of implementing a little helper that inherits from this and allows access to some interfaces. Yes, that should be done. Especially if you test algorithms with multiple dependencies it can be quite difficult to spot a bug if not the single components are tested.

rrahn:
I mean I totally agree that only the relevant parts of the interfaces should be tested. I am just literally confused about what you want me to take out from your comment? Maybe we discuss this offline.

include/seqan3/alignment/matrix/detail/alignment_matrix_column_major_range.hpp
```diff
@@ -0,0 +1,404 @@
...
+        }
+        //!\}
+
+    private:
+
+        //!\brief Pointer to the derived type.
+        derived_t * me_ptr;
+        //!\brief The current column index.
+        size_t column_index{};
+    };
```
joergi-w:
```suggestion
    }; // class iterator_type
```

include/seqan3/alignment/matrix/detail/alignment_matrix_column_major_range.hpp
```diff
@@ -0,0 +1,404 @@
...
+
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    constexpr alignment_matrix_column_major_range() = default; //!< Defaulted.
+    constexpr alignment_matrix_column_major_range(alignment_matrix_column_major_range const &) = default; //!< Defaulted.
+    constexpr alignment_matrix_column_major_range(alignment_matrix_column_major_range &&) = default; //!< Defaulted.
+    constexpr alignment_matrix_column_major_range & operator=(alignment_matrix_column_major_range const &) = default; //!< Defaulted.
+    constexpr alignment_matrix_column_major_range & operator=(alignment_matrix_column_major_range &&) = default; //!< Defaulted.
+    ~alignment_matrix_column_major_range() = default; //!< Defaulted.
```
joergi-w: align the `=` operators, please

joergi-w: What about `noexcept`?

include/seqan3/alignment/matrix/detail/alignment_matrix_column_major_range.hpp
```diff
@@ -0,0 +1,404 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides seqan3::detail::alignment_matrix_column_major_range.hpp.
```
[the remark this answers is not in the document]
joergi-w: Maybe make clear in the name that it denotes the dp matrix and not a score matrix.

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column.hpp
```diff
@@ -0,0 +1,192 @@
...
+#include <seqan3/core/concept/core_language.hpp>
+#include <seqan3/core/simd/concept.hpp>
+#include <seqan3/std/iterator>
+#include <seqan3/std/ranges>
+#include <seqan3/std/span>
+
+namespace seqan3::detail
+{
+
+/*!\brief An alignment score matrix storing only a single column for the computation.
```
joergi-w: Can it be called matrix then? 

And logically: how can a score matrix be reduced to less than _alphabet_size * (alphabet_size + 1) / 2_ entries, if you have to store the scores for all combinations of characters??

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column.hpp
```diff
@@ -0,0 +1,192 @@
...
+    using matrix_base_t = alignment_score_matrix_one_column_base<score_t>;
+    //!\brief The base class for iterating over the matrix.
+    using range_base_t = alignment_matrix_column_major_range<alignment_score_matrix_one_column<score_t>>;
+
+    //!\brief The actual element type.
+    using element_type = typename matrix_base_t::element_type;
+    //!\copydoc alignment_matrix_column_major_range::alignment_column_type
+    using alignment_column_type = typename range_base_t::alignment_column_type;
+    //!\brief The view type used to alias an alignment column
+    //!       [customises the behaviour of seqan3::detail::alignment_matrix_column_major_range].
```
joergi-w: you should use `/*! ... */` for consecutive comment lines.

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column.hpp
```diff
@@ -0,0 +1,192 @@
...
+    using matrix_base_t::cols;
+
+public:
+    /*!\name Associated types
+     * \{
+     */
+    //!\brief The proxy type.
+    using value_type = alignment_score_matrix_proxy_type<score_t>;
+    //!\brief Same as value type.
+    using reference = value_type;
```
joergi-w: use `reference_type`

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column.hpp
```diff
@@ -0,0 +1,192 @@
...
+    //!\brief The size type.
+    using size_type = typename matrix_base_t::size_type;
+    //!\copydoc alignment_matrix_column_major_range::iterator
+    using iterator = typename range_base_t::iterator;
```
joergi-w: same here, use `_type` (search also other occurrences, I stop marking it here)

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column.hpp
```diff
@@ -0,0 +1,192 @@
...
+    /*!\brief Construction from two ranges.
+     * \tparam first_sequence_t  The first range type; must model std::ranges::ForwardRange.
+     * \tparam second_sequence_t The second range type; must model std::ranges::ForwardRange.
+     * \param[in] first  The first range.
+     * \param[in] second The second range.
+     * \param[in] initial_value The value to initialise the matrix with. Default initialised if not specified.
+     *
+     * \details
+     *
+     * Does only obtains the sizes of the passed ranges in order to allocate the score matrix. Only one column
```
joergi-w:
```suggestion
     * Obtains only the sizes of the passed ranges in order to allocate the score matrix. Only one column
```

joergi-w: Why don't we just pass the sizes themselves?

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column.hpp
```diff
@@ -0,0 +1,192 @@
...
+
+private:
+
+    /*!\brief Returns the current alignment matrix column depending on the current column index.
+     * \param[in] column_index The current column index.
+     * \returns the current alignment matrix column.
+     *
+     * \details
+     *
+     * Creates a new alignment column based on the current column index.
```
joergi-w: The column index is not used. Why do we need to pass it?

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column.hpp
```diff
@@ -0,0 +1,192 @@
...
+     */
+    constexpr alignment_column_type initialise_column(size_t const SEQAN3_DOXYGEN_ONLY(column_index)) noexcept
+    {
+        return alignment_column_type{*this, column_data_view_type{std::addressof(matrix_base_t::pool[0]),
+                                                                  matrix_base_t::rows}};
+    }
+
+    /*!\brief Generates the proxy type referring to the corresponding cells within the underlying storage.
+     * \tparam    iter_t An iterator type over the current column view; Must model std::RandomAccessIterator.
```
joergi-w:
```suggestion
     * \tparam    iter_t An iterator type over the current column view; must model std::RandomAccessIterator.
```

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column.hpp
```diff
@@ -0,0 +1,192 @@
...
+    template <std::RandomAccessIterator iter_t>
+    constexpr void on_column_iterator_creation(iter_t iter) noexcept
+    {
+        // Cache the next diagonal value.
+        std::get<1>(matrix_base_t::cache) = std::get<0>(*iter);
+    }
+
+    /*!\brief Update the last diagonal used for the current computation.
+     * \tparam iter_t The type of the internal column iterator.
+     * \param  iter   The internal iterator pointing to the cell before incrementing.
```
joergi-w: Document that it is not used...

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_banded.hpp
```diff
@@ -0,0 +1,200 @@
...
+#include <seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_base.hpp>
+#include <seqan3/alignment/matrix/detail/alignment_score_matrix_proxy_type.hpp>
+#include <seqan3/std/iterator>
+#include <seqan3/std/ranges>
+#include <seqan3/std/span>
+
+namespace seqan3::detail
+{
+
+/*!\brief A banded alignment score matrix storing only a single banded column for the computation.
```
joergi-w: Wait... you are misusing the term **score matrix** here. 
As a score matrix we understand something like BLOSUM62 or a position-dependent score matrix. You mean the dp matrix that contains the prefix scores up to the current position. Please think about renaming the class and file name here and in related files.

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_banded.hpp
```diff
@@ -0,0 +1,200 @@
...
+        matrix_base_t::pool.resize(band_size + 1, element_type{initial_value, initial_value});
+    }
+    //!\}
+
+private:
+
+    //!\brief The column index where the upper bound of the band passes through.
+    int32_t band_col_index{};
+    //!\brief The row index where the lower bound of the band passes through.
+    int32_t band_row_index{};
```
joergi-w: The description is not so good as one can define the band in a way that the lower bound doesn't pass though any row because it is larger than the main diagonal... you probably expect a negative row index then.

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_banded.hpp
```diff
@@ -0,0 +1,200 @@
...
+
+    //!\copydoc alignment_score_matrix_one_column::initialise_column
+    constexpr alignment_column_type initialise_column(size_t const column_index) noexcept
+    {
+        int32_t slice_begin = std::max<int32_t>(0, band_col_index - column_index);
+        int32_t row_end_index = column_index - band_col_index + band_size;
+        int32_t slice_end = band_size - std::max<int32_t>(row_end_index - matrix_base_t::rows, 0);
+
+        assert(row_end_index >= 0);
+        assert(slice_begin >= 0);
```
joergi-w: always true, because you use std::max(0, ...)

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_banded.hpp
```diff
@@ -0,0 +1,200 @@
...
+     * \param  iter   The internal iterator pointing to the begin of the current column.
+     */
+    template <std::RandomAccessIterator iter_t>
+    constexpr void on_column_iterator_creation(iter_t iter) noexcept
+    {
+        // Cache the last diagonal value.
+        std::get<0>(matrix_base_t::cache) = std::get<0>(*iter);
+    }
+
+    //!\brief No-op. No steps necessary.
```
joergi-w: It may produce a doxygen warning for the undocumented `\tparam iter_t`.

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_base.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+/*!\brief A base class for alignment score matrices using only one column to compute the matrix.
+ * \tparam score_t The type of the scores.
+ * \ingroup alignment_matrix
+ *
+ * \details
+ *
+ * Manages the actual storage as a std::vector. How much memory is allocated is handled by the derived type.
+ * The `score_t` must either model the seqan3::Arithmetic or seqan3::detail::Simd vector concept.
+ * This class provides
```
joergi-w: ... (missing text)

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_base.hpp
```diff
@@ -0,0 +1,65 @@
...
+ */
+template <typename score_t>
+struct alignment_score_matrix_one_column_base
+{
+    static_assert(Arithmetic<score_t> || Simd<score_t>,
+                  "Value type must either be a trace_directions object or a Simd vector.");
```
joergi-w: It's score type, not value type. 
You assert for Arithmetic, not trace_directions.

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_base.hpp
```diff
@@ -0,0 +1,65 @@
...
+     * \{
+     */
+    //!\brief The actual element type.
+    using element_type = std::tuple<score_t, score_t>;
+    //!\brief The allocator type.
+    using allocator_type = aligned_allocator<element_type, sizeof(element_type)>;
+    //!\brief The type of the underlying storage.
+    using pool_type = std::vector<element_type, allocator_type>;
+    //!\brief The size type.
+    using size_type = size_t;
```
joergi-w: Why do we need this? `size_t` as return type below is also fine and saves a lookup for the actual `size_type`.

include/seqan3/alignment/matrix/detail/alignment_score_matrix_proxy_type.hpp
```diff
@@ -0,0 +1,49 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides seqan3::detail::alignment_matrix_score_base.
```
joergi-w: no

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_banded.hpp
```diff
@@ -0,0 +1,200 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides seqan3::detail::alignment_score_matrix_single_column_banded.
```
joergi-w:
```suggestion
 * \brief Provides seqan3::detail::alignment_score_matrix_one_column_banded.
```

include/seqan3/alignment/matrix/detail/alignment_score_matrix_proxy_type.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+/*!\brief A proxy type for a unified access to the score matrix during alignment computation.
+ * \tparam value_type The type wrapped by this proxy.
+ * \ingroup alignment_matrix
+ *
+ * \details
+ *
+ * Provides named accessors to the respective values of the score matrix during the alignment computation.
+ * The `value_type` must either model the seqan3::Arithmetic or seqan3::detail::Simd vector concept.
+ * Note for the banded case the left cell for reading points to a different location then the left cell to store the
```
joergi-w:
```suggestion
 * Note that in the banded case the left cell for reading points to a different location compared to the left cell to store the
```

joergi-w: confusing sentence... try to reformulate

include/seqan3/alignment/matrix/detail/alignment_score_matrix_proxy_type.hpp
```diff
@@ -0,0 +1,49 @@
...
+ * next element. Thus, there is a `r_left` and `w_left` element for reading and writing respectively.
+ */
+template <typename value_type>
+struct alignment_score_matrix_proxy_type
+{
+    static_assert(Arithmetic<value_type> || Simd<value_type>,
+                  "Value type must either be a trace_directions object or a Simd vector.");
```
joergi-w: how is trace_directions asserted?

include/seqan3/alignment/matrix/detail/alignment_matrix_column_major_range.hpp
```diff
@@ -0,0 +1,404 @@
...
+
+        /*!\brief The iterator over an alignment column.
+         *
+         * \details
+         *
+         * The iterator represents the current cell within an alignment column. When dereferenced it
+         * calls the function `make_proxy` of the derived class to obtain a proxy value over the current matrix cell.
+         * This proxy type depends on the derived type.
+         */
+        class iterator_type
```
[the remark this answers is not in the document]
rrahn: No, because they are strongly coupled. Making them external classes will produce another set of issues since the correct types and data must be available. Since friends cannot be transferred it will get messy to provide the correct information.

include/seqan3/alignment/matrix/detail/alignment_matrix_column_major_range.hpp
```diff
@@ -0,0 +1,404 @@
...
+        {
+        public:
+
+            /*!\name Associated types
+             * \{
+             */
+            //!\brief A proxy type dependent on the derived type.
+            using value_type = typename deferred_type<typename derived_t::value_type>::type;
+            //!\brief The same as value type since it is a proxy.
+            using reference = value_type;
```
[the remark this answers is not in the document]
rrahn: You have to ask the standard about this.

test/unit/alignment/matrix/detail/alignment_trace_matrix_full_test.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include <seqan3/alignment/matrix/detail/alignment_trace_matrix_full.hpp>
+#include <seqan3/alignment/matrix/trace_directions.hpp>
+
+#include "alignment_matrix_base_test_template.hpp"
+
+using namespace seqan3;
+using namespace seqan3::detail;
+
+using testing_types = ::testing::Types<
+    std::pair<detail::alignment_trace_matrix_full<trace_directions>, std::false_type>>;
```
joergi-w: add includes

test/unit/alignment/matrix/detail/simulated_alignment_test_template.hpp
```diff
@@ -0,0 +1,140 @@
...
+    // ----------------------------------------------------------------------------
+    // Initialise the first column
+    // ----------------------------------------------------------------------------
+
+    auto mat_it = this->matrix().begin();
+    auto col = *mat_it;
+    auto col_it = col.begin();;
+    (*col_it).current = 0;
+    (*col_it).up = this->gap;
+    (*col_it).w_left = this->gap;
```
joergi-w: Why don't you use `->`operator?

test/unit/alignment/matrix/detail/simulated_alignment_test_template.hpp
```diff
@@ -0,0 +1,140 @@
...
+    ++col_it;
+    size_t step = 0;
+
+    for (; col_it != col.end(); ++col_it, ++step)
+    {
+        (*col_it).current = (*col_it).up;
+        (*col_it).up += this->gap;
+        (*col_it).w_left = (*col_it).current + this->gap;
```
joergi-w: also here

include/seqan3/alignment/matrix/detail/alignment_trace_matrix_full_banded.hpp
```diff
@@ -0,0 +1,232 @@
...
+ * as needed by the standard alignment algorithm. The matrix is modelled as std::ranges::InputRange since the
+ * alignment algorithm iterates only once over the complete matrix to calculate the values.
+ *
+ * ### Only computing the coordinates
+ *
+ * Sometimes it is desired to only get access to the alignment coordinates. This can be achieved by setting
+ * `coordinate_only = true`. In this case no memory will be allocated and only an internal state is maintained to
+ * generate the alignment coordinates. The coordinates are relative to the banded matrix, i.e. they do not represent the
+ * corresponding row indices of the actual (unbanded) matrix.
+ */
```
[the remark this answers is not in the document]
rrahn: But it is part of the trace matrix interface and thus should be here.

include/seqan3/alignment/matrix/detail/alignment_trace_matrix_full_banded.hpp
```diff
@@ -0,0 +1,232 @@
...
+
+    //!\copydoc alignment_score_matrix_one_column::initialise_column
+    constexpr alignment_column_type initialise_column(size_t const column_index) noexcept
+    {
+        int32_t slice_begin = std::max<int32_t>(0, band_col_index - column_index);
+        int32_t row_end_index = column_index - band_col_index + band_size;
+        int32_t slice_end = band_size - std::max<int32_t>(row_end_index - matrix_base_t::rows, 0);
+
+        assert(row_end_index >= 0);
+        assert(slice_begin >= 0);
```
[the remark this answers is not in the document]
rrahn: as above, I'd like to keep the assert to explicitly state expected behavior.

test/unit/alignment/matrix/detail/simulated_alignment_test_template.hpp
rrahn: because it can't. The ->operator does not work for the iterator since it returns a temporary. It cannot return a pointer to a temporary then.

test/unit/alignment/matrix/detail/simulated_alignment_test_template.hpp
rrahn: like above

include/seqan3/alignment/matrix/detail/alignment_matrix_column_major_range.hpp
```diff
@@ -0,0 +1,404 @@
...
+ *
+ * This base class requires following function and types defined within the derived type in order to customise the
+ * returned type over the underlying matrix.
+ *
+ *  * `value_type`: The derived type must define a `value_type` type which determines the reference type of the
+ *                  inner range.
+ *  * `column_data_view_type`: A type that models std::ranges::RandomAccessRange and std::ranges::View and is the
+ *                            internal type used to iterate over a matrix column.
+ *  * `make_proxy`: This function is called by the dereferencing an iterator over the inner column range and must
+ *                  return a type that is convertible to the reference type of the column (see `value_type`).
```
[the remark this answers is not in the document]
joergi-w: Thank you, this is really helpful now 👍

include/seqan3/alignment/matrix/detail/alignment_matrix_column_major_range.hpp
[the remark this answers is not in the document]
joergi-w: Okay, I see the point. It already helps a lot that you marked the end of the classes with respective comments.

include/seqan3/alignment/matrix/detail/alignment_matrix_column_major_range.hpp
joergi-w: Regarding `noexcept`: Even if it is automatically noexcept, it would be helpful to see the keyword in the code and the documentation. 

Regarding the alignment: This is a PR and the name is therefore not likely to change again. I just wanted to point out that you start introducing a second code style into the library, because I usually see the rule-of-six aligned.

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column.hpp
joergi-w: okay 👍

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column.hpp
joergi-w: Okay, then leave it if you prefer...

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_banded.hpp
joergi-w: Sure, I agree!

include/seqan3/alignment/matrix/detail/alignment_trace_matrix_full.hpp
```diff
@@ -0,0 +1,199 @@
...
+
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    constexpr alignment_trace_matrix_full() = default; //!< Defaulted.
+    constexpr alignment_trace_matrix_full(alignment_trace_matrix_full const &) = default; //!< Defaulted.
+    constexpr alignment_trace_matrix_full(alignment_trace_matrix_full &&) = default; //!< Defaulted.
+    constexpr alignment_trace_matrix_full & operator=(alignment_trace_matrix_full const &) = default; //!< Defaulted.
+    constexpr alignment_trace_matrix_full & operator=(alignment_trace_matrix_full &&) = default; //!< Defaulted.
+    ~alignment_trace_matrix_full() = default; //!< Defaulted.
```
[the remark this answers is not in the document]
joergi-w: ok!

include/seqan3/alignment/matrix/detail/alignment_trace_matrix_full_banded.hpp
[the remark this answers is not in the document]
joergi-w: I agree!

include/seqan3/std/algorithm
```diff
@@ -93,5 +94,10 @@ using SEQAN3_DOXYGEN_ONLY(move =) ::ranges::move;
 */
 using SEQAN3_DOXYGEN_ONLY(sort =) ::ranges::sort;
 
+/*!\typedef std::ranges::transform
+* \brief Alias for ranges::transform. Sorts a range.
```
joergi-w: I don't understand the part `Sorts a range.`:
That's not what we expect from the transform function...

include/seqan3/alignment/matrix/detail/alignment_trace_matrix_full.hpp
```diff
@@ -0,0 +1,199 @@
...
+    /*!\brief Construction from two ranges.
+     * \tparam first_sequence_t  The first range type; must model std::ranges::ForwardRange.
+     * \tparam second_sequence_t The second range type; must model std::ranges::ForwardRange.
+     * \param[in] first  The first range.
+     * \param[in] second The second range.
+     * \param[in] initial_value The value to initialise the matrix with. Default initialised if not specified.
+     *
+     * \details
+     *
+     * Does only obtains the sizes of the passed ranges in order to allocate the trace back matrix.
```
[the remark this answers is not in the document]
joergi-w: The obtain part has not been changed...

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_banded.hpp
rrahn: No, the point is that except for very specific reasons, e.g. when doing bit operations, one should always prefer the signed types. This is however in many cases already messed up because the size type for all containers etc are unsigned. But in this specific case I do a lot of arithmetic operations and it helps to have them as signed integers, especially since I am getting negative values on purpose for some computations.

include/seqan3/alignment/matrix/detail/alignment_trace_matrix_full.hpp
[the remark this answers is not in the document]
rrahn: OK, we discussed it and dropped the alignment requirement on constructors.

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_banded.hpp
joergi-w: Good, then better leave it signed.

include/seqan3/std/algorithm
rrahn: Oh I missed that

rrahn:
@joergi-w Already did.

test/unit/alignment/matrix/detail/alignment_score_matrix_one_column_banded_test.cpp
```diff
@@ -0,0 +1,107 @@
...
+                              test_type);
+
+//-----------------------------------------------------------------------------
+// Test outer iterator
+//-----------------------------------------------------------------------------
+struct outer_iterator
+{};
+
+template <>
+struct iterator_fixture<outer_iterator> : alignment_matrix_base_test<test_type>
```
smehringer: why do you create an extra tag if you can just use `alignment_matrix_base_test<test_type>` as the type specialisation?

include/seqan3/alignment/matrix/detail/alignment_matrix_column_major_range.hpp
```diff
@@ -0,0 +1,474 @@
...
+
+    /*!\brief A column iterator over the alignment matrix.
+     *
+     * \details
+     *
+     * This iterator enables the iteration over the underlying alignment matrix in column-major-order.
+     * When dereferenced the iterator returns a seqan3::detail::alignment_matrix_column_major_range::column_type.
+     * The initialisation of the alignment-column depends on the derived type.
+     */
+    class iterator_type
```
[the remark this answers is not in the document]
rrahn: there are many reasons. The most significant one is that in the algorithm you need to know when you start the next column since you have to process the cells at the top of the column differently.

include/seqan3/alignment/matrix/detail/alignment_matrix_column_major_range.hpp
```diff
@@ -0,0 +1,474 @@
...
+        constexpr void operator++(int) noexcept
+        {
+            ++(*this);
+        }
+
+        /*!\name Comparison operators
+         * \{
+         */
+        //!\brief Returns `true` if the behind-the-end column was reached, `false` otherwise.
+        constexpr bool operator==(std::ranges::default_sentinel_t const &) const noexcept
```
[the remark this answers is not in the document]
rrahn: Only if you want to make sure that there is a conversion sequence from another type to the iterator type which then can use the friend declaration. But here we really don't want that.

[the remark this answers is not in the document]
rrahn: which reminds me to check if constructors are explicit

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column.hpp
```diff
@@ -0,0 +1,172 @@
...
+    using matrix_base_t = alignment_score_matrix_one_column_base<score_t>;
+    //!\brief The base class for iterating over the matrix.
+    using range_base_t = alignment_matrix_column_major_range<alignment_score_matrix_one_column<score_t>>;
+
+    //!\brief The actual element type.
+    using element_type = typename matrix_base_t::element_type;
+    //!\copydoc seqan3::detail::alignment_matrix_column_major_range::alignment_column_type
+    using alignment_column_type = typename range_base_t::alignment_column_type;
+    //!\copydoc seqan3::detail::alignment_matrix_column_major_range::column_data_view_type
+    using column_data_view_type = std::span<element_type>;
```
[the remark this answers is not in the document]
rrahn: Well, in this context it is not necessary, since the underlying container is not a generic user input but a fixed type that is part of the classes invariant.

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column.hpp
```diff
@@ -0,0 +1,172 @@
...
+    {
+        matrix_base_t::cols = static_cast<size_t>(std::ranges::distance(first) + 1);
+        matrix_base_t::rows = static_cast<size_t>(std::ranges::distance(second) + 1);
+        matrix_base_t::pool.resize(matrix_base_t::rows + 1, element_type{initial_value, initial_value});
+    }
+    //!\}
+
+private:
+
+    //!\copydoc seqan3::detail::alignment_matrix_column_major_range::initialise_column
```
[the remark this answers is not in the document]
rrahn: probably it does. Maybe I remove the note from the basic doc but note, that this is also never intended for public API

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_banded.hpp
```diff
@@ -0,0 +1,194 @@
...
+    /*!\name Auxiliary types
+     * \{
+     */
+    //!\brief The base class for data storage.
+    using matrix_base_t = alignment_score_matrix_one_column_base<score_t>;
+    //!\brief The base class for iterating over the matrix.
+    using range_base_t = alignment_matrix_column_major_range<alignment_score_matrix_one_column_banded<score_t>>;
+
+    //!\brief The actual element type.
+    using element_type = typename matrix_base_t::element_type;
```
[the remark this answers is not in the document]
rrahn: I remember that I had issues with that at some point and just resolved it like this and I never looked at it again. I will revisit it.

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_banded.hpp
```diff
@@ -0,0 +1,194 @@
...
+        return {std::get<0>(*host_iter),             // current
+                std::get<0>(matrix_base_t::cache),   // last diagonal
+                std::get<1>(*(host_iter + 1)),       // last left (read)
+                std::get<1>(*(host_iter)),           // next left (write)
+                std::get<1>(matrix_base_t::cache)};  // last up
+    }
+
+    //!\copydoc seqan3::detail::alignment_matrix_column_major_range::on_column_iterator_creation
+    template <std::RandomAccessIterator iter_t>
+    constexpr void on_column_iterator_creation(iter_t host_iter) noexcept
```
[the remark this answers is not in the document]
rrahn: That is not the iterator over the column but basically an iterator type over the view that is generated in `initialise_column`. Therefor it would be ok to get to the iterator type in the derived class but this way it is much more practical. Again there is no generic use case for this but a clear dependency between the derived and the base class

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_base.hpp
```diff
@@ -0,0 +1,64 @@
...
+    //!\}
+
+    //!\brief The linearised memory pool storing only one column of the matrix.
+    pool_type pool{};
+    //!\brief Internal cache for the last diagonal and vertical value during the alignment computation.
```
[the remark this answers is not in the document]
rrahn: in the banded case I need to cache the diagonal in between. that's where I use the third parameter

include/seqan3/alignment/matrix/detail/alignment_score_matrix_proxy_type.hpp
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * Provides named accessors to the respective values of the score matrix during the alignment computation.
+ * The `value_type` must either model the seqan3::Arithmetic or seqan3::detail::Simd vector concept.
+ * In some cases the last value to the left is stored in a different location then the next left value that
+ * will be used in the subsequent column (e.g. in the banded alignment matrix). Accordingly the left value is
+ * split in a reference for reading, referred to as `r_left` (for reading), and a reference to the next left value
+ * on the same row, referred to as `w_left` (for writing).
+ */
+template <typename value_type>
+struct alignment_score_matrix_proxy_type
```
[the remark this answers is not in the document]
rrahn: > We did not do that anywhere else

Muhhahahahaha, you need to discuss this with @h-2. He used it in several places before and I was like why? Now it falls back on me 🤣. Anyway, I gladly change this back.

include/seqan3/alignment/matrix/detail/alignment_trace_matrix_full.hpp
```diff
@@ -0,0 +1,202 @@
...
+ */
+template <typename trace_t, bool coordinate_only = false>
+class alignment_trace_matrix_full :
+    public alignment_trace_matrix_base<trace_t>,
+    public alignment_matrix_column_major_range<alignment_trace_matrix_full<trace_t, coordinate_only>>
+{
+private:
+
+    static_assert(std::Same<trace_t, trace_directions> || Simd<trace_t>,
+                  "Value type must either be a trace_directions object or a Simd vector.");
```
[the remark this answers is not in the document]
rrahn: That won't be possible since the vector type can only store regular integer types. It would be possible only to check at runtime if the value does no exceed the values of trace_directions, but of course you don't want to do that. Also in an assert you don't.

include/seqan3/alignment/matrix/detail/alignment_trace_matrix_full_banded.hpp
```diff
@@ -0,0 +1,238 @@
...
+ *
+ * ### Only computing the coordinates
+ *
+ * Sometimes it is desired to only get access to the alignment coordinates. This can be achieved by setting
+ * `coordinate_only = true`. In this case no memory will be allocated and only an internal state is maintained to
+ * generate the alignment coordinates. The coordinates are relative to the banded matrix, i.e. they do not represent the
+ * corresponding row indices of the actual (unbanded) matrix.
+ */
+template <typename trace_t, bool coordinate_only = false>
+class alignment_trace_matrix_full_banded :
```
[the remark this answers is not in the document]
rrahn: I really don't think that this will help making the design more clear. A lot of stuff is already outsourced. I'll see that I can remove the definition of the noop functions by putting defaults into the base class. But the most important point is that I don't want to define variables (especially in SIMD mode) if the algorithm don't use them.

test/unit/alignment/matrix/detail/alignment_score_matrix_one_column_banded_test.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+    using iterator_tag = std::input_iterator_tag;
+    static constexpr bool const_iterable = false;
+
+    using base_t::test_range;
+    std::vector<int> expected_range{0, 1, 2, 3, 4};
+
+    template <typename lhs_t, typename rhs_t>
+    static void expect_eq(lhs_t const &, rhs_t const &)
+    {}
```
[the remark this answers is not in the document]
rrahn: I think about it again. Maybe I can define one way to basic test this. I really just care more about the code coverage. The semantics are tested at another place already.

test/unit/alignment/matrix/detail/alignment_score_matrix_one_column_banded_test.cpp
rrahn: Because the current design does not allow me to define another iterator_fixture with the same test type (see below), because I would define the same type twice. One of the points I wasn't happy with in the first place.

include/seqan3/alignment/matrix/detail/alignment_matrix_column_major_range.hpp
```diff
@@ -0,0 +1,474 @@
...
+         * \{
+         */
+        //!\brief Returns an iterator to the begin of the column.
+        constexpr auto begin() noexcept
+        {
+            assert(me_ptr != nullptr);
+            return iterator_type{*this};
+        }
+
+        //!\brief Deleted begin for const-qualified alignment-columns.
```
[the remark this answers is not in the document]
rrahn: because there is currently no need for this and allowing cons-iterability does add a lot more complexity to the iterators and the other classes. This range interface implemented here is strongly coupled with the algorithm and it doesn't fit to add something that the algorithm does not need. I don't see another place where this might be used. 
I do add a comment though.

include/seqan3/alignment/matrix/detail/alignment_matrix_column_major_range.hpp
```diff
@@ -0,0 +1,474 @@
...
+#include <seqan3/std/ranges>
+#include <seqan3/std/span>
+
+namespace seqan3::detail
+{
+
+/*!\brief Provides a range interface for alignment matrices.
+ * \implements std::ranges::InputRange
+ * \ingroup alignment_matrix
+ * \tparam derived_t The derived type implementing the data model of the matrix.
```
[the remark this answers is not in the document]
rrahn: Ok, I will add a snippet for a use case.

include/seqan3/alignment/matrix/all.hpp
```diff
@@ -21,5 +21,6 @@
 #include <seqan3/alignment/matrix/alignment_coordinate.hpp>
 #include <seqan3/alignment/matrix/alignment_optimum.hpp>
 #include <seqan3/alignment/matrix/debug_matrix.hpp>
+#include <seqan3/alignment/matrix/detail/all.hpp>
```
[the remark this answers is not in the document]
rrahn: Well, except that everything but one class is detail in the matrix header at the moment. At some point this needs to be cleansed as well.

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column.hpp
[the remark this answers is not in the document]
rrahn: I could add three methods as noop in the base class and also removed the note since it is docu ented in the class's description

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_banded.hpp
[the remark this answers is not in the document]
rrahn: Ok, I was playing around with it. Also regarding how are things displayed in the doc. I think I slimmed down now to the minimum and also such that the doc looks nice.

include/seqan3/alignment/matrix/detail/alignment_matrix_column_major_range.hpp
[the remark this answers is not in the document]
smehringer: I did not ask you to change it ;) I was just curious

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column.hpp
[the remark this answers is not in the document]
smehringer: I have no idea what you are trying to say :laughing:

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_base.hpp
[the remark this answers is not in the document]
smehringer: pls add that in the comment

include/seqan3/alignment/matrix/detail/alignment_score_matrix_proxy_type.hpp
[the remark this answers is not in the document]
smehringer: I don't remember a discussion about this? Is there a reasoning behind the type?

test/unit/alignment/matrix/detail/alignment_score_matrix_one_column_banded_test.cpp
smehringer: Well because you should templatize the fixture over the iterator type that you are testing, not the range type. And your iterator types are not the same. 
In the first case it is `decltype(alignment_matrix_base_test<test_type>::test_range.begin())`  (which I am sure can be written out more nicely.). That way the test directly displays the actual type. 
and in the second case it is `decltype((*alignment_matrix_base_test<test_type>::test_range.begin()).begin())` 
But I don't care you think this artificial struct is nicer.

rrahn:
@smehringer I applied most of the suggested changes and also made the iterator tests testing something 😏

include/seqan3/alignment/matrix/detail/alignment_matrix_column_major_range.hpp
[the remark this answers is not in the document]
rrahn: I know, I was merely describing my motivation why I didn't add it. I am sorry if you felt offended. That was not my intention. 👼

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column.hpp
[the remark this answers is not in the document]
rrahn: Just, they are now not needed to be provided by the derived class. So they vanished and the class looks a lot more less copy'n'paste 😉

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_base.hpp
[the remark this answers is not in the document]
rrahn: I did already 😉

include/seqan3/alignment/matrix/detail/alignment_score_matrix_proxy_type.hpp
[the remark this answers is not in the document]
rrahn: Well, I can't remember completely but something about this being a type. So I usually only apply it if you would rather call your variable like that and it would otherwise clash with the type definition then. Another example is strong_type. But I can't remember having a rule or something.

test/unit/alignment/matrix/detail/alignment_score_matrix_one_column_banded_test.cpp
rrahn: > Well because you should templatize the fixture over the iterator type that you are testing, not the range type :D. And your iterator types are not the same.

But that's exactly the problem that I was referring to. I usually got the iterator from the range not the other way around. So I would need to define the types globally then get the iterators from them and then implement some fixture base classes that kind of infer back from the iterator type to the range type to provide the actual data. And now, since I have two different flavors each of the them with two different iterator types this all makes it a little confusing about how you apply your types. So it feels like the spezialisation of the iterator_fixture is kind of weird since I only have to spezialise it to tell gtest that this is the correct fixture to use. Anyway, other people might see it differently and I found a way to get what I want (with probably the same loc as in my suggested version) so all good from my side.

include/seqan3/alignment/matrix/detail/alignment_score_matrix_proxy_type.hpp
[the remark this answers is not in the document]
smehringer: Hm but every struct is basically a type?... and it is not called `vector_type`. 
I guess `iterator_type` is to not confuse it with the `iterator` type member but who knows. Thanks for changing it I think this holds data just like the vector and should not be called type. Especially if the concept names soon end with `_type` :rofl:

rrahn:
@smehringer ping

test/unit/alignment/matrix/detail/alignment_score_matrix_one_column_banded_test.cpp
```diff
@@ -0,0 +1,109 @@
...
+template <>
+struct iterator_fixture<outer_iterator> : alignment_matrix_base_test<test_type>
+{
+    using base_t = alignment_matrix_base_test<test_type>;
+
+    using iterator_tag = std::input_iterator_tag;
+    static constexpr bool const_iterable = false;
+
+    using base_t::test_range;
+    std::vector<int> expected_range{0, 0, 0, 0, 0};
```
smehringer: Doesn't the outer iterator return views on access? o.O

include/seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_base.hpp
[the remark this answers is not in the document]
rrahn: I don't understand. If you want to know what is the variable for you can look into the docs. If you want to touch the file and change some of it's implementation you have to look into the code anyway. So for the docs I think it is sufficient to know that it is some cached variable. To understand it's usage one needs to look into the code. And there it will be even further described.

include/seqan3/alignment/matrix/detail/alignment_score_matrix_proxy.hpp
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * Provides named accessors to the respective values of the score matrix during the alignment computation.
+ * The `score_type` must either model the seqan3::Arithmetic or seqan3::detail::Simd vector concept.
+ * In some cases the last value to the left is stored in a different location then the next left value that
+ * will be used in the subsequent column (e.g. in the banded alignment matrix). Accordingly the left value is
+ * split in a reference for reading, referred to as `r_left` (for reading), and a reference to the next left value
+ * on the same row, referred to as `w_left` (for writing).
+ */
+template <typename score_type>
+struct alignment_score_matrix_proxy
```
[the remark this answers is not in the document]
rrahn: No this is not possible, since I have different requirements on the value type. alignment_trace requires the type to be a trace direction. At the end there would be 8 loc saved, 4 if I do inline documentations, which I will do 😄 . Not sure that this is worth it then.

test/unit/alignment/matrix/detail/alignment_score_matrix_one_column_banded_test.cpp
rrahn: yeah, but I don't provide the range but only access the first element. As said before semantics are tested already in a simulation of the alignment algorithm.

PR #1042, opened 2019-06-03
[MISC] Search module redesign

eseiler:
Resolves #898 

- [x] FMIndex
- [x] BiFMIndex
- [x] FMIndexCursor
- [x] BiFMIndexCursor
- [x] Search
- [x] Docs
- [x] Snippets
- [x] Tutorial

PR #1107, opened 2019-06-19
[MISC] Change bool to strong enum in index template parameter

### LINKED ISSUE #1102: Change naming of index template argument
We changed the fm index interface to contain a bool template parameter to differentiate between text collection or single texts. I would discuss if we should not change this to a strong enum to make it more readable. Because bi_fm_index<true> (also type deduction should usually handle these cases without specifying the bool parameter) is not easy to understand. bi_fm_index<text_layout::collection> would be more expressive. In general, as long as public APIs contain a non-type template parameter they should be strong (except maybe numbers)

------

Resolution:

- text_layout::single
- text_layout::collection

- index_text::single
- index_text::collection

Enrico decides about the name
??

eseiler:
Resolves #1102

- [x] Wait for deprecated macro (#1118)
- [x] Test deprecated implementation

### DIFF
--- a/include/seqan3/search/fm_index/bi_fm_index.hpp
+++ b/include/seqan3/search/fm_index/bi_fm_index.hpp
@@ -24,13 +24,22 @@
 namespace seqan3
 {
 
+//!\cond
+SEQAN3_DEPRECATED_310
+void bi_fm_index_deprecation(bool);
+
+template <typename t>
+void bi_fm_index_deprecation(t);
+//!\endcond
+
 /*!\addtogroup submodule_fm_index
  * \{
  */
 
 /*!\brief The SeqAn Bidirectional FM Index
  * \implements seqan3::BiFmIndex
- * \tparam is_collection    Indicates whether this index works on a text collection (`true`) or a single text (`false`).
+ * \tparam is_collection    Indicates whether this index works on a text collection or a single text.
+ *                          See seqan3::text_layout.
  * \tparam sdsl_index_type_ The type of the underlying SDSL index, must model seqan3::SdslIndex.
  * \details
  *
@@ -55,8 +64,10 @@ namespace seqan3
  *
  * \attention When building an index for a **text collection** over any alphabet, the symbols with rank 254 and 255
  *            are reserved and may not be used in the text.
+ *
+ * \deprecated Use seqan3::text_layout to indicate single texts and text collections. The use of bool is deprecated.
  */
-template <bool is_collection = false, detail::SdslIndex sdsl_index_type_ = default_sdsl_index_type>
+template <auto is_collection = text_layout::single, detail::SdslIndex sdsl_index_type_ = default_sdsl_index_type>
 class bi_fm_index
 {
 protected:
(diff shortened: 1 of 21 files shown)


include/seqan3/search/fm_index/concept.hpp
```diff
@@ -92,7 +92,7 @@ template <typename t>
 SEQAN3_CONCEPT FmIndex = std::Semiregular<t> && requires (t index)
 {
     typename t::size_type;
-    // typename t::cursor_type;
+    typename t::cursor_type;
```
smehringer: what does this change has to do with this PR?

test/unit/search/fm_index/bi_fm_index_dna4_test.cpp
```diff
@@ -5,11 +5,16 @@
 // shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
 // -----------------------------------------------------------------------------------------------------
 
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
```
smehringer: hmm. should we define this locally in tests or maybe globally?

include/seqan3/search/fm_index/concept.hpp
eseiler: In #1042 I commented it out, but it shouldn't be.
I fixed it here by accident.

I guess we can leave this one-liner in the PR?

test/unit/search/fm_index/bi_fm_index_dna4_test.cpp
eseiler: I'd do it locally since we usually want deprecation warnings - except when we test the deprecated stuff.
If I put the pragma one line of the test, it throws warnings.

include/seqan3/search/fm_index/concept.hpp
smehringer: yes but I propose in different commits

test/unit/search/fm_index/bi_fm_index_dna4_test.cpp
smehringer: alright I guess we just need to keep track of those when we remove the deprecated stuff for the next release, that we remove the macro too

include/seqan3/search/fm_index/bi_fm_index.hpp
```diff
@@ -55,15 +64,17 @@ namespace seqan3
...
-template <bool is_collection = false, detail::SdslIndex sdsl_index_type_ = default_sdsl_index_type>
+template <auto is_collection = text_layout::single, detail::SdslIndex sdsl_index_type_ = default_sdsl_index_type>
 class bi_fm_index
 {
 protected:
     //!\brief The alphabet size of the text.
     size_t sigma{0};
     //!\brief Indicates whether index is built over a collection.
-    static constexpr bool is_collection_{is_collection};
+    static constexpr text_layout is_collection_{is_collection};
```
h-2: This can stay the same so you don't need all the casts. The strong_type is just for protecting the interface...

PR #1110, opened 2019-06-20
[FIX] 3 minor changes in alignment IO

smehringer:
1. BAM needs to write `*\0` for empty read ids (not `\0` like right now, which did not impede our files since they are equivalent but for samtools those are not equivalent)

2. SAM should not have requirements on types it does not use. (This way you could also pass std::ignore to the file)

3. The read id alphabet should not be modifiable by the traits object since there is no use case other than `char` (container is modifiable) and char is required by BAM. (Is this an API break?)

### DIFF
--- a/include/seqan3/alphabet/detail/convert.hpp
+++ b/include/seqan3/alphabet/detail/convert.hpp
@@ -39,7 +39,8 @@ constexpr std::array<out_t, alphabet_size<in_t>> convert_through_char_representa
     [] () constexpr
     {
         std::array<out_t, alphabet_size<in_t>> ret{};
-        for (typename in_t::rank_type i = 0; i < alphabet_size<in_t>; ++i)
+        // for (decltype(alphabet_size<in_t>) i = 0; ...) causes indefinite compilation :(
+        for (auto i = decltype(alphabet_size<in_t>){0}; i < alphabet_size<in_t>; ++i)
             assign_char_to(to_char(assign_rank_to(i, in_t{})), ret[i]);
         return ret;
     }()
--- a/include/seqan3/io/alignment_file/format_bam.hpp
+++ b/include/seqan3/io/alignment_file/format_bam.hpp
@@ -836,7 +836,7 @@ public:
                 /* block_size  */ 0,  // will be initialised right after
                 /* refID       */ -1, // will be initialised right after
                 /* pos         */ ref_offset.value_or(-1),
-                /* l_read_name */ std::max<uint8_t>(std::ranges::distance(id) + 1, 2) /* empty id is repr. by '*\0' */,
+                /* l_read_name */ std::max<uint8_t>(std::min<size_t>(std::ranges::distance(id) + 1, 255), 2),
                 /* mapq        */ mapq,
                 /* bin         */ reg2bin(ref_offset.value_or(-1), std::ranges::distance(get<1>(align))),
                 /* n_cigar_op  */ static_cast<uint16_t>(cigar.size()),
@@ -918,7 +918,7 @@ public:
             if (std::ranges::distance(id) == 0) // empty id is represented as * for backward compatibility
                 stream_it = '*';
             else
-                std::ranges::copy_n(std::ranges::begin(id), std::ranges::distance(id), stream_it); // write read id
+                std::ranges::copy_n(std::ranges::begin(id), core.l_read_name - 1, stream_it); // write read id
             stream_it = '\0';
 
             // write cigar
--- a/include/seqan3/io/alignment_file/format_bam.hpp
+++ b/include/seqan3/io/alignment_file/format_bam.hpp
@@ -866,7 +866,25 @@ public:
                     {
                         if (!std::ranges::empty(id_source)) // otherwise default will remain (-1)
                         {
-                            auto id_it = header.ref_dict.find(id_source);
+                            auto id_it = header.ref_dict.end();
+
+                            if constexpr (std::ranges::ContiguousRange<decltype(id_source)> &&
+                                          std::ranges::SizedRange<decltype(id_source)> &&
+                                          ForwardingRange<decltype(id_source)>)
+                            {
+                                id_it = header.ref_dict.find(std::span{std::ranges::data(id_source),
+                                                                       std::ranges::size(id_source)});
+                            }
+                            else
+                            {
+                                using header_ref_id_type = std::remove_reference_t<decltype(header.ref_ids()[0])>;
+
+                                static_assert(ImplicitlyConvertibleTo<decltype(id_source), header_ref_id_type>,
+                                  "The ref_id type is not convertible to the reference id information stored in the "
+                                  "reference dictionary of the header object.");
+
+                                id_it = header.ref_dict.find(id_source);
+                            }
 
                             if (id_it == header.ref_dict.end())
                             {
@@ -874,6 +892,7 @@ public:
                                                                      "not be found in BAM header ref_dict: ",
                                                                      header.ref_dict, ".")};
                             }
+
                             id_target = id_it->second;
                         }
                     }
--- a/include/seqan3/io/alignment_file/format_bam.hpp
+++ b/include/seqan3/io/alignment_file/format_bam.hpp
@@ -836,7 +836,7 @@ public:
                 /* block_size  */ 0,  // will be initialised right after
                 /* refID       */ -1, // will be initialised right after
                 /* pos         */ ref_offset.value_or(-1),
-                /* l_read_name */ static_cast<uint8_t>(std::ranges::distance(id) + 1),
+                /* l_read_name */ std::max<uint8_t>(std::ranges::distance(id) + 1, 2) /* empty id is repr. by '*\0' */,
                 /* mapq        */ mapq,
                 /* bin         */ reg2bin(ref_offset.value_or(-1), std::ranges::distance(get<1>(align))),
                 /* n_cigar_op  */ static_cast<uint16_t>(cigar.size()),
@@ -915,7 +915,10 @@ public:
 
             std::ranges::copy_n(reinterpret_cast<char *>(&core), sizeof(core), stream_it);  // write core
 
-            std::ranges::copy_n(std::ranges::begin(id), std::ranges::distance(id), stream_it); // write read id
+            if (std::ranges::distance(id) == 0) // empty id is represented as * for backward compatibility
+                stream_it = '*';
+            else
+                std::ranges::copy_n(std::ranges::begin(id), std::ranges::distance(id), stream_it); // write read id
             stream_it = '\0';
 
             // write cigar
--- a/include/seqan3/io/alignment_file/format_sam.hpp
+++ b/include/seqan3/io/alignment_file/format_sam.hpp
@@ -838,7 +838,7 @@ protected:
             if (is_char<'S'>(*std::ranges::begin(stream_view)))              // SQ (sequence dictionary) tag
             {
                 ref_info_present_in_header = true;
-                std::string id;
+                value_type_t<decltype(hdr.ref_ids())> id;
                 std::tuple<int32_t, std::string> info{};
 
                 parse_tag_value(id);                                         // parse required SN (sequence name) tag
@@ -1100,15 +1100,31 @@ public:
                       !std::Integral<std::remove_reference_t<ref_id_type>> &&
                       !detail::is_type_specialisation_of_v<std::remove_reference_t<ref_id_type>, std::optional>)
         {
-            static_assert(ImplicitlyConvertibleTo<ref_id_type &, typename decltype(header.ref_dict)::key_type>,
-                      "The ref_id type is not convertible to the reference id information stored in the "
-                      "reference dictionary of the header object.");
 
             if (options.sam_require_header && !std::ranges::empty(ref_id))
             {
-                if ((header.ref_dict).count(ref_id) == 0) // no reference id matched
-                    throw format_error{std::string("The ref_id '") + std::string(ref_id) +
-                                       "' was not in the list of references"};
+                auto id_it = header.ref_dict.end();
+
+                if constexpr (std::ranges::ContiguousRange<decltype(ref_id)> &&
+                              std::ranges::SizedRange<decltype(ref_id)> &&
+                              ForwardingRange<decltype(ref_id)>)
+                {
+                    id_it = header.ref_dict.find(std::span{std::ranges::data(ref_id), std::ranges::size(ref_id)});
+                }
+                else
+                {
+                    using header_ref_id_type = std::remove_reference_t<decltype(header.ref_ids()[0])>;
+
+                    static_assert(ImplicitlyConvertibleTo<ref_id_type, header_ref_id_type>,
+                                  "The ref_id type is not convertible to the reference id information stored in the "
+                                  "reference dictionary of the header object.");
+
+                    id_it = header.ref_dict.find(ref_id);
+                }
+
+                if (id_it == header.ref_dict.end()) // no reference id matched
+                    throw format_error{detail::to_string("The ref_id '", ref_id, "' was not in the list of references:",
+                                                         header.ref_ids())};
             }
         }
 
--- a/include/seqan3/io/alignment_file/format_sam.hpp
+++ b/include/seqan3/io/alignment_file/format_sam.hpp
@@ -988,7 +988,9 @@ public:
               typename align_type,
               typename qual_type,
               typename mate_type,
-              typename tag_dict_type>
+              typename tag_dict_type,
+              typename e_value_type,
+              typename bit_score_type>
     void write(stream_type                            &  stream,
                alignment_file_output_options const    &  options,
                header_type                            && header,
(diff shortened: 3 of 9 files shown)


include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -1033,11 +1035,6 @@ class alignment_file_output_format<format_sam>
                       "The id object must be a std::ranges::ForwardRange over "
                       "letters that model seqan3::Alphabet.");
 
-        static_assert((std::ranges::ForwardRange<ref_seq_type>    &&
```
eseiler: Why is the assert gone?
If it can be `std::ignore`, then you could assert that either `decays_to_ignore_v< ref_seq_type>` or the requirements are fulfilled.

smehringer: This type is not used anywhere so it should not be restricted I think

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -836,7 +836,7 @@ class alignment_file_output_format<format_bam> : alignment_file_output_format<fo
                 /* block_size  */ 0,  // will be initialised right after
                 /* refID       */ -1, // will be initialised right after
                 /* pos         */ ref_offset.value_or(-1),
-                /* l_read_name */ static_cast<uint8_t>(std::ranges::distance(id) + 1),
+                /* l_read_name */ std::max<uint8_t>(std::ranges::distance(id) + 1, 2) /* empty id is repr. by '*\0' */,
```
h-2: length is limited to 256 chars?

h-2:
> The read id alphabet should not be modifiable by the traits object since there is no use case other than char (container is modifiable) and char is required by BAM. (Is this an API break?)

Why would anything `ConvertibleTo<char>` not be ok? Currently `char8_t` would work for example, it will just be converted. Strictly speaking you should check `ConvertibleTo` Stream's char_type (this has implicitly been the case right now so changing to this behaviour is not breaking) and maybe restrict that further if that actually can't be `wchar_t`.

> SAM should not have requirements on types it does not use. (This way you could also pass std::ignore to the file)

In general all the static assert could/should be moved to the file, or not? They are the same for all formats anyway, because all formats are instantiated with all parameters. The dynamic asserts via exceptions are format specific and should be with the format...

include/seqan3/io/alignment_file/format_bam.hpp
smehringer: yup. funny right?
https://samtools.github.io/hts-specs/SAMv1.pdf

smehringer: seems like they do not expect protein database alignment

smehringer:
> In general all the static assert could/should be moved to the file, or not? They are the same for all formats anyway, because all formats are instantiated with all parameters. The dynamic asserts via exceptions are format specific and should be with the format...

Yes you are right. Shall I do this in this PR or another?

h-2:
> Yes you are right. Shall I do this in this PR or another?

However you like!

include/seqan3/io/alignment_file/format_bam.hpp
h-2: just curious: Do you assert on the length of the ID before writing it or will this cause overflow?

include/seqan3/alphabet/detail/convert.hpp
```diff
@@ -38,8 +39,10 @@ constexpr std::array<out_t, alphabet_size<in_t>> convert_through_char_representa
 {
     [] () constexpr
     {
+        static_assert(alphabet_size<in_t> <= std::numeric_limits<size_t>::max());
```
eseiler: `
A type whose size cannot be represented by std::size_t is ill-formed (since C++14)
`
(https://en.cppreference.com/w/cpp/types/size_t)

So we do not need this

include/seqan3/alphabet/detail/convert.hpp
```diff
@@ -38,8 +39,10 @@ constexpr std::array<out_t, alphabet_size<in_t>> convert_through_char_representa
...
+
         std::array<out_t, alphabet_size<in_t>> ret{};
-        for (typename in_t::rank_type i = 0; i < alphabet_size<in_t>; ++i)
+        for (size_t i = 0; i < alphabet_size<in_t>; ++i)
```
eseiler: Wouldn't `auto` work?
Otherwise, I would propose `detail::min_viable_uint<alphabet_size<in_t>>`.

include/seqan3/io/alignment_file/format_bam.hpp
h-2: ping

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -1100,14 +1097,19 @@ class alignment_file_output_format<format_sam>
...
+                using header_ref_id_type = std::remove_reference_t<decltype(header.ref_ids()[0])>;
+
+                static_assert(ImplicitlyConvertibleTo<value_type_t<ref_id_type>, value_type_t<header_ref_id_type>>,
+                              "The ref_id type is not convertible to the reference id information stored in the "
+                              "reference dictionary of the header object.");
+
+                auto key = ref_id | std::ranges::to<header_ref_id_type>;
+
+                if ((header.ref_dict).count(key) == 0) // no reference id matched
+                    throw format_error{std::string("The ref_id '") + std::string(key) +
```
h-2:
```suggestion
                    throw format_error{std::string_view("The ref_id '") + std::string_view(key) +
```

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -1100,14 +1097,19 @@ class alignment_file_output_format<format_sam>
...
-                      "The ref_id type is not convertible to the reference id information stored in the "
-                      "reference dictionary of the header object.");
 
             if (options.sam_require_header && !std::ranges::empty(ref_id))
             {
-                if ((header.ref_dict).count(ref_id) == 0) // no reference id matched
-                    throw format_error{std::string("The ref_id '") + std::string(ref_id) +
+                using header_ref_id_type = std::remove_reference_t<decltype(header.ref_ids()[0])>;
+
+                static_assert(ImplicitlyConvertibleTo<value_type_t<ref_id_type>, value_type_t<header_ref_id_type>>,
```
h-2: here you check convertibleTo

include/seqan3/io/alignment_file/input.hpp
```diff
@@ -122,8 +119,7 @@ SEQAN3_CONCEPT AlignmentFileInputTraits = requires (t v)
     requires SequenceContainer<typename t::template sequence_container<typename t::sequence_alphabet>>;
 
     // field::ID
-    requires WritableAlphabet<typename t::id_alphabet>;
-    requires SequenceContainer<typename t::template id_container<typename t::id_alphabet>>;
+    requires SequenceContainer<typename t::template id_container<char>>;
```
h-2: but here you fix it to char

include/seqan3/io/alignment_file/format_bam.hpp
smehringer: ops sorry. This will cause an overflow. Shall I throw or assert or truncate?

include/seqan3/io/alignment_file/format_sam.hpp
smehringer: Well I can not concatenate string views. I changed the whole by using the `detail::to_string` function

include/seqan3/io/alignment_file/format_sam.hpp
smehringer: this is the *reference* id type

include/seqan3/io/alignment_file/input.hpp
smehringer: this is the *read* id type

include/seqan3/io/alignment_file/format_bam.hpp
smehringer: @h-2 ping

smehringer: Since you ignore me I decided for truncating the id, because this ensures a correct BAM file without interrupting a pipeline. (e.g. if seqan3 is used for sam bam conversion and one single line in the middle of the SAM file has a too long read name it would be frustrating for the user to manually truncate it.)

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -866,7 +866,9 @@ class alignment_file_output_format<format_bam> : alignment_file_output_format<fo
                     {
                         if (!std::ranges::empty(id_source)) // otherwise default will remain (-1)
                         {
-                            auto id_it = header.ref_dict.find(id_source);
+                            using ref_id_t = std::remove_reference_t<decltype(header.ref_ids()[0])>;
+                            auto key = id_source | std::ranges::to<ref_id_t>;
+                            auto id_it = header.ref_dict.find(key);
```
h-2: change the key-type of the map to be ranges::subrange and decay contiguous iterators to pointers. 

Add a static_assert that the respective types are convertible to the key_type.

include/seqan3/alphabet/detail/convert.hpp
```diff
@@ -39,7 +39,7 @@ constexpr std::array<out_t, alphabet_size<in_t>> convert_through_char_representa
     [] () constexpr
     {
         std::array<out_t, alphabet_size<in_t>> ret{};
-        for (typename in_t::rank_type i = 0; i < alphabet_size<in_t>; ++i)
+        for (detail::min_viable_uint_t<alphabet_size<in_t>> i = 0; i < alphabet_size<in_t>; ++i)
```
h-2:
```suggestion
        for (decltype(alphabet_size<in_t>) i = 0; i < alphabet_size<in_t>; ++i)
```

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -866,14 +866,31 @@ class alignment_file_output_format<format_bam> : alignment_file_output_format<fo
...
+                                id_it = header.ref_dict.find(std::span{std::ranges::data(id_source),
+                                                                       std::ranges::size(id_source)});
+                            }
+                            else
+                            {
+                                using header_ref_id_type = std::remove_reference_t<decltype(header.ref_ids()[0])>;
+
+                                static_assert(ImplicitlyConvertibleTo<decltype(id_source), header_ref_id_type>,
+                                  "The ref_id type is not convertible to the reference id information stored in the "
+                                  "reference dictionary of the header object.");
```
h-2: Can you move this with the other static asserts / constraints?

include/seqan3/io/alignment_file/header.hpp
```diff
@@ -88,7 +90,9 @@ class alignment_file_header
     //!\brief Stream deleter with default behaviour (ownership assumed).
     static void ref_ids_deleter_default(ref_ids_type * ptr) { delete ptr; }
     //!\brief The key's type of ref_dict.
-    using key_type = std::ranges::all_view<reference_t<ref_ids_type>>;
+    using key_type = std::conditional_t<std::ranges::ContiguousRange<reference_t<ref_ids_type>>,
+                        std::span<const std::remove_reference_t<reference_t<reference_t<ref_ids_type>>>>,
```
h-2:
```suggestion
                        std::span<innermost_value_type_t<ref_ids_type>>,
```

include/seqan3/io/alignment_file/header.hpp
```diff
@@ -88,7 +90,9 @@ class alignment_file_header
...
+                        std::ranges::all_view<reference_t<ref_ids_type>>>;
```
h-2:
```suggestion
                        all_view<reference_t<ref_ids_type>>>;
```

h-2: Use seqan' view::all

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -866,14 +866,31 @@ class alignment_file_output_format<format_bam> : alignment_file_output_format<fo
                     {
                         if (!std::ranges::empty(id_source)) // otherwise default will remain (-1)
                         {
-                            auto id_it = header.ref_dict.find(id_source);
+                            auto id_it = header.ref_dict.end();
+
+                            if constexpr (std::ranges::ContiguousRange<decltype(id_source)>)
```
h-2: You forgot to check SizedRange here

include/seqan3/io/alignment_file/input.hpp
```diff
@@ -913,16 +905,26 @@ class alignment_file_input
...
         // initialise reference map and ref_dict if ref_ids are non-empty
-        for (size_t idx = 0; idx < ref_ids.size(); ++idx)
+        for (int32_t idx = 0; idx < std::ranges::distance(ref_ids); ++idx)
         {
-            header_ptr->ref_id_info.emplace_back(std::ranges::size(ref_sequences[idx]), "");
-            header_ptr->ref_dict[header_ptr->ref_ids()[idx]] = idx;
+            header_ptr->ref_id_info.emplace_back(std::ranges::distance(ref_sequences[idx]), "");
+
+            if constexpr (std::ranges::ContiguousRange<reference_t<typename traits_type::ref_ids>> &&
+                          std::ranges::SizedRange<reference_t<typename traits_type::ref_ids>>)
```
h-2: for all the cases you also need to check `seqan3::ForwardingRange` otherwise you might also take this branch for ranges that return temporary objects of contiguousrange that then go out of scope, but you are holding pointers to themm inside your span.

include/seqan3/alphabet/detail/convert.hpp
smehringer: I thought these are eqivalent

include/seqan3/io/alignment_file/format_bam.hpp
smehringer: then I need to add two static asserts for ref_id and mate_ref_id instead of one in this lambda that processes both?

include/seqan3/alphabet/detail/convert.hpp
smehringer: If I make that change, the bam test compiles forever :scream:

smehringer: `for (auto i = decltype(alphabet_size<in_t>){0}; i < alphabet_size<in_t>; ++i)` funktioniert wieder ^^

h-2: > If I make that change, the bam test compiles forever scream

It should actually be faster. Keep in mind that `detail::min_viable_uint_t<` instantiates a type, whereas decltype does not.

> for (auto i = decltype(alphabet_size<in_t>){0}; i < alphabet_size<in_t>; ++i) 

whut?

include/seqan3/io/alignment_file/format_bam.hpp
h-2: Well we want to move the static asserts to the file anyway, right?

smehringer: some of them.. there are some that are format specific. e.g. that BAM requires the header member to be present

smehringer: ah and this assert cannot go into the file wither because how do I know on file level if the ref?id might be ignored? Because if it is ignored this check should not be performed!

h-2: > some of them.. there are some that are format specific. e.g. that BAM requires the header member to be present

I thought we already discussed this and came to the conclusion that all the static_asserts are file by definition, because dynamic dispatch will lead to all formats being instantiated anyway.

include/seqan3/alphabet/detail/convert.hpp
smehringer: :woman_shrugging: I dunno :woman_shrugging:

include/seqan3/io/alignment_file/format_bam.hpp
smehringer: > I thought we already discussed this and came to the conclusion that all the static_asserts are file by definition, because dynamic dispatch will lead to all formats being instantiated anyway.

Yes I thought so but although they are instantiated it depends on how the read() function is called and thos requirements are different.. I noticed when I moved around some of the asserts in a temporary branch

h-2: Hm, I am not quite sure I get it. In any case, please create a card for this.

include/seqan3/alphabet/detail/convert.hpp
h-2: This is still weird and it is not guaranteed that the min_viable_uint_type is the correct type. Just because we decide to do that doesn't mean other alphabets don't use `size_t` so this will result in conversion. I don't understand why decltype should cause problems here. If it does, please document it in the code.

include/seqan3/io/alignment_file/format_bam.hpp
h-2: If we have a card for this, you can mark this as resolved.

include/seqan3/alphabet/detail/convert.hpp
smehringer: hm if decltype is more correct, i can use the second version?
```
for (auto i = decltype(alphabet_size<in_t>){0}; i < alphabet_size<in_t>; ++i)
```

h-2: yes

smehringer: alright. I changed it and made a comment that it does not work the other way

smehringer:
Jenkins says 100% passed but still shows a failure :woman_shrugging:

PR #1126, opened 2019-06-24
[MISC] restructure debug_stream

### LINKED ISSUE #905: Provide debug_stream for a custom type without pulling the complete library in.
We decided on the following design:

```
core/debug_stream.hpp
core/debug_stream_detail.hpp
```

debug_stream_detail.hpp:
* `include <iosfwd>`
* debug_stream definition, no overloads

foobar.hpp:
* `include <debug_stream_detail.hpp>`
* overload of `operator<<(debug_stream_type, foobar const &)`

debug_stream.hpp
* `include <iostream>`
* `include <debug_stream_detail.hpp>`
* overloads of stuff that doesn't have own files, like built-in types, ranges

https://github.com/seqan/seqan3/issues/905#issuecomment-494002847

-------
~~I would like to discuss this at a strategy meeting.~~
~~relates to #510 ~~

marehr:
Fixed #905.


test/include/seqan3/test/pretty_printing.hpp
```diff
@@ -12,7 +12,9 @@
 
 #pragma once
 
-#include<seqan3/core/debug_stream.hpp>
+#include <ios>
+
+#include <seqan3/core/debug_stream.hpp>
```
smehringer: hmm... so `#include <seqan3/core/debug_stream.hpp>` still pulls in the "whole library" right? Then pretty printing does, which we use throughout the tests, so we did not gain the desired independence of unit tests do we? :see_no_evil:

marehr: Yes, but this PR already improves it quite a lot. I would like to make a new PR to fix it for pretty_printing.

test/unit/range/view/view_drop_test.cpp
```diff
@@ -15,7 +15,6 @@
 
 #include <range/v3/view/unique.hpp>
 
-#include <seqan3/core/debug_stream.hpp>
```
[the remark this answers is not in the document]
marehr: It is not needed, because it is included from a different include. Do you want to keep it?

marehr:
@h-2 Can you rereview? Thank you :heart:

PR #1128, opened 2019-06-24
[FIX] add constructors

h-2:
Who let me get away with making classes without constructors?

### DIFF
--- a/include/seqan3/range/view/deep.hpp
+++ b/include/seqan3/range/view/deep.hpp
@@ -111,6 +111,13 @@ public:
     /*!\name Constructors, destructor and assignment
      * \{
      */
+    constexpr deep()                         noexcept = default; //!< Defaulted.
+    constexpr deep(deep const &)             noexcept = default; //!< Defaulted.
+    constexpr deep(deep &&)                  noexcept = default; //!< Defaulted.
+    constexpr deep & operator=(deep const &) noexcept = default; //!< Defaulted.
+    constexpr deep & operator=(deep &&)      noexcept = default; //!< Defaulted.
+    ~deep()                                  noexcept = default; //!< Defaulted.
+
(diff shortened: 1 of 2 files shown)


include/seqan3/range/view/deep.hpp
```diff
@@ -111,6 +111,13 @@ class deep : public detail::adaptor_base<deep<underlying_adaptor_t>, underlying_
     /*!\name Constructors, destructor and assignment
      * \{
      */
+    constexpr deep()                         noexcept = default; //!< Defaulted.
```
smehringer: aren't those inherited by `using base_type::base_type` ?

h-2: Inherited copy constructors are not considered for copy construction of the derived type. Only for construction of the derived type from the base type. For the default constructors it is not as important, but we have rule-of-six so I added all of them.

h-2:
BTW: jenkins as issues again (as predicted :zipper_mouth_face: )

include/seqan3/range/view/detail.hpp
```diff
@@ -239,8 +240,19 @@ class combined_adaptor :
     }
 
 public:
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    constexpr combined_adaptor()                                              = default; //!< Defaulted.
```
smehringer: why is the default constructor not noexcept (also for the others)?

h-2: Because they aren't always `noexcept`. It depends on the base class.

h-2: And `noexcept` it not like `constexpr` in that it is ok if possible and if not also ok; it hard fails if the base type's constructor is not noexcept

PR #1130, opened 2019-06-25
[FIX] Pressing return should default the version check.

### LINKED ISSUE #1108: version checker bug
Just pressing enter at the interactive screen does not work.

smehringer:
Resolve #1108


smehringer:
Code coverage of course fails because I cannot test the interactive code :disappointed:

PR #1138, opened 2019-06-26
[DOC] How to write a custom alphabet

joergi-w:
This is the HowTo tutorial about how to implement a custom alphabet that models the concept requirements.


doc/howto/write_an_alphabet/index.md
```diff
@@ -0,0 +1,150 @@
...
+static_assert(seqan3::Semialphabet<dna2>);      // compiler error
+static_assert(seqan3::Alphabet<dna2>);          // compiler error
+~~~
+
+You should see that your type models only the std::CopyConstructible concept.
+Let's have a look at the documentation for std::StrictTotallyOrdered. Can you guess what it describes?
+
+It describes the requirements for types that are comparable via `<`, `<=`, `>` and `>=`.
+This is useful so that you can sort a text over the alphabet, for example.
+Additionally, in order to model std::StrictTotallyOrdered we need to make our alphabet
```
smehringer:
```suggestion
Additionally, std::StrictTotallyOrdered is again a refinement of [std::EqualityComparable](\ref std::EqualityComparable), which requires `==` and `!=`, so let's first implement those with free member functions:
```

doc/howto/write_an_alphabet/index.md
```diff
@@ -0,0 +1,150 @@
...
+\endassignment
+\solution
+\snippet alphabet_custom_steps.cpp inequality
+\endsolution
+
+We see that our type now models std::EqualityComparable, which is a prerequisite of std::StrictTotallyOrdered.
+\snippet alphabet_custom_steps.cpp equality_comparable
+
+\assignment{Excercise}
+Implement the four compare operators *as free functions* and verify that your type models std::StrictTotallyOrdered.
```
smehringer:
```suggestion
Implement the four comparison operators *as free member functions* and verify that your type models std::StrictTotallyOrdered.
```

doc/howto/write_an_alphabet/index.md
```diff
@@ -0,0 +1,150 @@
...
+\endassignment
+\solution
+\snippet alphabet_custom_steps.cpp comparison
+\endsolution
+
+## Semialphabet
+
+Let's move on to the more interesting concepts. seqan3::Semialphabet requires the *rank* interface
```
smehringer:
```suggestion
Let's move on to the more interesting concepts. seqan3::Semialphabet requires the *rank interface*
```

doc/howto/write_an_alphabet/index.md
```diff
@@ -0,0 +1,150 @@
...
+that we have introduced in the [alphabet tutorial](\ref tutorial_alphabets). If we define the member functions
+`to_rank`, `assign_rank` and `alphabet_size` for our type, SeqAn automatically exposes them to the respective global
+functions `seqan3::to_rank`, `seqan3::assign_rank_to` and `seqan3::alphabet_size`. 
+Note that `assign_rank` is only required for seqan3::WritableSemialphabet.
+\snippet alphabet_custom_steps.cpp writable_semialphabet
+
```
smehringer: Add a static assert in the code snippet

`As you can see from the static_assert, our `dna2` alphabet now models seqan3::SemiAlphabet.`

and we can automatically use some free functions on our alphabet (look up `customization points wiki??` if you are interested in how this works):

`snippet with seqan3::to_rank // ADL magic`

doc/howto/write_an_alphabet/index.md
```diff
@@ -0,0 +1,150 @@
...
+\endsolution
+
+We see that our type now models std::EqualityComparable, which is a prerequisite of std::StrictTotallyOrdered.
+\snippet alphabet_custom_steps.cpp equality_comparable
+
+\assignment{Excercise}
+Implement the four compare operators *as free functions* and verify that your type models std::StrictTotallyOrdered.
+\endassignment
+\solution
+\snippet alphabet_custom_steps.cpp comparison
```
smehringer: please include the full solution. It is not too long yet and it should be copy-and-pastable.

For example, you have not shown any example with those functions *inside* the dna2 struct. I know this should be clear to most cpp developers but you never know

doc/howto/write_an_alphabet/index.md
```diff
@@ -0,0 +1,150 @@
...
+\snippet alphabet_custom_steps.cpp writable_semialphabet
+
+## Alphabet
+
+Now that you have a feeling for concepts, have a look at seqan3::Alphabet and make your type 
+also model the seqan3::Alphabet concept. If you keep the suggested namings, SeqAn will automatically expose
+the members to the respective global functions that model the concept.
+
+\assignment{Excercise}
+Please implement the member function `to_char(void)` which returns the char 'S' or 'W'
```
smehringer:
```suggestion
Please implement the member function `to_char()` which returns the char 'S' or 'W'
```

doc/howto/write_an_alphabet/index.md
```diff
@@ -0,0 +1,150 @@
...
+depending on the rank value.
+\endassignment
+\solution
+\snippet alphabet_custom_steps.cpp alphabet
+\endsolution
+
+\assignment{Excercise}
+Now model the full seqan3::WritableAlphabet concept.
+
+1. Implement the member function `dna2 & assign_char(char const)`.
```
smehringer: I would put the `assign char` task together with `to char` and then write one more sentence

```
seqan3::Alphabet additionally requires that the type has a member function `bool char_is_valid` which can be used to determine whether a given character is an actual representation of the alphabet. For `dna2`, `S` and `W`, as well as `s` and `w`,  should be valid but nothing else. 
```

before the write `char_is_valid` task

doc/howto/write_an_alphabet/index.md
```diff
@@ -0,0 +1,150 @@
...
+2. Implement the static member function `bool char_is_valid(char const)`.
+\endassignment
+\solution
+\snippet alphabet_custom_steps.cpp writable_alphabet
+\endsolution
+
+At this point the seqan3::WritableAlphabet concept should be modelled successfully:
```
smehringer:
```suggestion
At this point the seqan3::Alphabet concept should be modelled successfully and even seqan3::WritableAlphabet should be fine as we implemented `assign_char`
```

doc/howto/write_an_alphabet/index.md
```diff
@@ -0,0 +1,150 @@
...
+\snippet alphabet_custom_steps.cpp writable_alphabet_concept
+
+## Shortcut: Alphabet base template
+
+In reality, you do not need to define all the functions you have learned in this exercise manually.
+Instead, you can inherit your type from seqan3::alphabet_base and just define the char-rank conversion
+in both directions. Read the documentation of seqan3::alphabet_base for details and examples.
+
```
smehringer: I would add the example of how dna2 would look if we would inherit alphabet_base.

doc/howto/write_an_alphabet/index.md
joergi-w: I have put `(void)` for the reason that empty parentheses automatically link the function to the global to_char function which is misleading in this case.

joergi-w:
I addressed most of the comments already; open TODOs are the two additional snippets for customisation point and alphabet_base.

joergi-w:
Cool, then I pass the review to @h-2 now...

doc/howto/write_an_alphabet/index.md
```diff
@@ -0,0 +1,149 @@
...
+\snippet dna2_comparison_operators.cpp comparison
+\endsolution
+
+## Semialphabet
+
+Let's move on to the more interesting concepts. seqan3::Semialphabet requires the *rank interface*
+that we have introduced in the [alphabet tutorial](\ref tutorial_alphabets). If we define the member functions
+`to_rank`, `assign_rank` and `alphabet_size` for our type, SeqAn automatically exposes them to the respective global
+functions `seqan3::to_rank`, `seqan3::assign_rank_to` and `seqan3::alphabet_size`.
+Note that `assign_rank` is only required for seqan3::WritableSemialphabet.
```
h-2: This seems a little outdated and should address the CPOs more closely. Maybe something like this:

```
Let's move on to the more interesting concepts. seqan3::Semialphabet constitutes the *rank interface*
that we introduced in the [alphabet tutorial](\ref tutorial_alphabets). Have a look at the API reference again. Beyond the conceptional requirements, it also requires that seqan3::alphabet_size and seqan3::to_rank can be called on your alphabet.

There are different ways to satisfy seqan3::alphabet_size and seqan3::to_rank, have a look at the respective API reference and also [the documentation on customisation points](link).

In this case we choose to implement the functionality as member functions:
```

doc/howto/write_an_alphabet/index.md
```diff
@@ -0,0 +1,149 @@
...
+\snippet dna2_semialphabet.cpp semialphabet
+
+As you can see from the `static_assert` our dna2 alphabet now models seqan3::Semialphabet and
+seqan3::WritableSemialphabet:
+\snippet dna2_semialphabet.cpp writable_semialphabet_concept
+
+With this we can automatically use some free functions on our dna2 alphabet. If you are interested in how this works,
+read our section on [Customisation points](\ref about_customisation). Here is an example:
+\snippet dna2_semialphabet.cpp free_functions
```
h-2: I see that you cover it here, but I think it's better to argue the other way around:
Because the concept requires seqan3::to_char, we have to implement a to_char() member (or other implementation).

doc/howto/write_an_alphabet/index.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Alphabet
+
+Now that you have a feeling for concepts, have a look at seqan3::Alphabet and seqan3::WritableAlphabet and make
+your type also model these concepts. If you keep the suggested namings, SeqAn will automatically
+expose the members to the respective global functions that model the concept.
```
h-2: Functions don't "model a concept".

doc/howto/write_an_alphabet/index.md
```diff
@@ -0,0 +1,149 @@
...
+
+\assignment{Excercise}
+1. Implement the member function `char to_char(void)` which returns the char 'S' or 'W' depending on the rank value.
+2. Implement the member function `dna2 & assign_char(char const)`.
+3. seqan3::WritableAlphabet additionally requires that the type has a member function `bool char_is_valid`
+   which can be used to determine whether a given character is an actual representation of the alphabet.
+   For the `dna2` alphabet 'S' and 'W' as well as 's' and 'w' should be valid, but nothing else.
+   <br>
+   Implement the static member function `bool char_is_valid(char const)`.
```
h-2: No, this is not required, it is optional. I would maybe skip it in the first example. Or explicitly state that it is optional.

doc/howto/write_an_alphabet/index.md
```diff
@@ -0,0 +1,149 @@
...
+
+At this point the seqan3::Alphabet concept should be modelled successfully and even seqan3::WritableAlphabet 
+is fine because we implemented `assign_char` and `char_is_valid`.
+\snippet dna2_alphabet.cpp writable_alphabet_concept
+
+## Shortcut: Alphabet base template
+
+In reality, you do not need to define all the functions you have learned in this exercise manually.
+Instead, you can inherit your type from seqan3::alphabet_base and just define the char-rank conversion
+in both directions. Read the documentation of seqan3::alphabet_base for details.
```
h-2: ```Often it is not required to implement the entire class yourself, you can instead derive from seqan3::alphabet_base which defines most things for you if you provide certain conversion tables. Read the documentation of seqan3::alphabet_base for details.```

doc/howto/write_an_alphabet/index.md
```diff
@@ -0,0 +1,149 @@
...
+This is an example of a minimal custom alphabet that provides implementations for all necessary customisation
+points. As an enum class it already models std::StrictTotallyOrdered.
+\snippet test/unit/alphabet/custom_alphabet_test.cpp my_alph
+
+## Implementation of a non-default-constructible class
+
+This is an example of a custom alphabet that is not default-constructible and that has a non-default overload for
+seqan3::char_is_valid_for.
+\snippet test/unit/alphabet/custom_alphabet2_test.cpp my_alph
+
```
h-2: Describe the few things that are different and why!

doc/howto/write_an_alphabet/dna2_inherit_base.cpp
```diff
@@ -0,0 +1,39 @@
...
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
+// -----------------------------------------------------------------------------------------------------
+
+//! [dna2]
+#include <array>                                         // std::array
+#include <seqan3/alphabet/alphabet_base.hpp>             // alphabet_base
+#include <seqan3/alphabet/concept.hpp>                   // alphabet concept checks
+
+// inherit from alphabet_base
```
h-2:
```suggestion
// derive from alphabet_base
```

doc/howto/write_an_alphabet/dna2_alphabet.cpp
```diff
@@ -0,0 +1,98 @@
...
+        char const rank_to_char[2] {'S', 'W'};
+        return rank_to_char[rank];
+    }
+
+    dna2 & assign_char(char const ch) noexcept
+    {
+        switch (ch)
+        {
+            case 'W': rank = 1;
+            case 'w': rank = 1; break;                   // allow assignment from uppercase and lowercase
```
h-2: The first line is superfluous, plus this might trigger a warning. 

Instead put it in one line:

```suggestion
            case 'W': case 'w': rank = 1; break;             // allow assignment from uppercase and lowercase
```

h-2:
> Either add this everywhere or remove it everwhere 

I have to correct myself, it is required, don't remove it. Instead add it to the comparison as well.

doc/howto/write_an_alphabet/dna2_only_rank.cpp
```diff
@@ -0,0 +1,26 @@
...
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
+// -----------------------------------------------------------------------------------------------------
+
+//! [struct]
+#include <seqan3/alphabet/concept.hpp>                   // alphabet concept checks
+
+struct dna2
+{
+    uint8_t rank{};
```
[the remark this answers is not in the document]
joergi-w: We are speaking about a numerical representation and although bool would be the most memory-efficient type it might be somehow unintuitive for the user...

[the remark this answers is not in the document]
h-2: I am not saying it has to be bool, but there should be some guide for users which type to chose. bool is in fact not more memory efficient than `uint8_t`.

doc/howto/write_an_alphabet/index.md
```diff
@@ -35,9 +37,13 @@ As a consequence, we denote Guanine and Cytosine as strong (S) and Adenine and T
 In this section we want to implement a seqan3::Alphabet that consists of the characters `S` and `W` to represent
 strong and weak nucleobases.
 
-Let's start with a simple struct that only holds the alphabet's numerical representation, namely the **rank** value:
+Let's start with a simple struct that only holds the alphabet's numerical representation, namely the **rank** value.
 \snippet dna2_only_rank.cpp struct
```
h-2: "It is not specified how the value of an alphabet is stored internally, however alphabets typically store the rank in as a member variable."

doc/howto/write_an_alphabet/index.md
```diff
@@ -0,0 +1,165 @@
...
+which pair in a DNA molecule by building 3 hydrogen bonds. Adenine and Thymine pair with only 2 hydrogen bonds.
+As a consequence, we denote Guanine and Cytosine as strong (S) and Adenine and Thymine as weak (W) nucleobases.
+In this section we want to implement a seqan3::Alphabet that consists of the characters `S` and `W` to represent
+strong and weak nucleobases.
+
+Let's start with a simple struct that only holds the alphabet's numerical representation, namely the **rank** value.
+\snippet dna2_only_rank.cpp struct
+
+\note
+As our alphabet has just two letters we could model it as well with `bool` instead of `uint8_t`.
```
h-2:
```suggestion
The type of the member variable is typically the smallest unsigned integer type that can hold the maximum rank. In our case we have only two values so `bool` would be sufficient. However `bool` is not actually smaller than `uint8_t` and does not always behave like an unsigned integral type so we use `uint8_t` here. You only need a different type if your alphabet has an alphabet size >=255.
```

doc/howto/write_an_alphabet/index.md
```diff
@@ -0,0 +1,165 @@
...
+reference and also the [documentation on customisation points](\ref about_customisation).
+
+In this case we choose to implement the functionality as member functions:
+\snippet dna2_semialphabet.cpp semialphabet
+
+As you can see from the `static_assert` our dna2 alphabet now models seqan3::Semialphabet and
+seqan3::WritableSemialphabet:
+\snippet dna2_semialphabet.cpp writable_semialphabet_concept
+
+With this we can automatically use some free functions on our dna2 alphabet. Here is an example:
```
h-2:
```suggestion
You can also try out the customisation points directly (they appear like free functions). Here is an example:
```

doc/howto/write_an_alphabet/index.md
```diff
@@ -0,0 +1,165 @@
...
+\snippet dna2_semialphabet.cpp free_functions
+
+## Alphabet
+
+Now that you have a feeling for concepts, have a look at seqan3::Alphabet and seqan3::WritableAlphabet and make
+your type also model these concepts. If you keep the suggested namings, SeqAn will automatically
+expose the members to the respective global functions as we have seen above.
```
h-2: delete the last sentence as it is a little confusing (what suggested naming, there no "global" functions, they are functions objects...)

doc/howto/write_an_alphabet/index.md
```diff
@@ -0,0 +1,165 @@
...
+
+\assignment{Excercise}
+1. Implement the member function `char to_char(void)` which returns the char 'S' or 'W' depending on the rank value.
```
h-2: `(void)` looks very weird (like C-code from the 90s). Why did you write that?

doc/howto/write_an_alphabet/index.md
```diff
@@ -0,0 +1,165 @@
...
+2. Implement the member function `dna2 & assign_char(char const)`.
+3. If a seqan3::WritableAlphabet has a member function `bool char_is_valid` it can be used to determine
+   whether a given character is an actual representation of the alphabet. For the `dna2` alphabet 'S' and 'W'
+   as well as 's' and 'w' should be valid, but nothing else. The member function is optional, because SeqAn provides
+   a fallback implementation for seqan3::char_is_valid_for (see its documentation for details).
+   <br>
```
h-2: it is not clear *why* people should implement this. I would write something like this instead of the previous few lines:
"There is a function object that tells us which characters are valid for a given alphabet: seqan3::char_is_valid_for.
By default, all characters are "valid" that are preserved when being assigned from and than be converted back. But in some cases you want to change the default behaviour, e.g. declaring lower-case letters to be valid, as well. "

doc/howto/write_an_alphabet/index.md
h-2: This is still open.

doc/howto/write_an_alphabet/index.md
joergi-w: If I omit the void, then doxygen automatically links the function in a bold blue font to the global to_char function, which is misleading here. The `void` was the only way to suppress this.

eseiler: Can you try
```
\ref to_char(void) "char to_char()"
```
you may have to play a bit around with the first part, e.g. `seqan3::to_char(void)` if the first version doesn't work.

joergi-w: > Then char %to_char() should avoid automatic linking (I'm not sure if it also works if you put it into the `)

Thank you! Yes, it works and looks pretty now :)

doc/howto/write_an_alphabet/index.md
joergi-w: oops... I had just applied it for the first snippet. 
I have provided some explanations above the last snippet now.

doc/howto/write_an_alphabet/index.md
```diff
@@ -0,0 +1,176 @@
...
+
+\assignment{Excercise}
+1. Implement the member function `char %to_char()` which returns the char 'S' or 'W' depending on the rank value.
+2. Implement the member function `dna2 & assign_char(char const)`.
+3. There is a function object that tells us which characters are valid for a given alphabet: seqan3::char_is_valid_for.
+   By default, all characters are "valid" that are preserved when being assigned from and then be converted back.
+   But in some cases you want to change the default behaviour, e.g. declaring lower-case letters to be valid, as well.
+   <br>
+   **Optional task:** Implement the static member function `bool char_is_valid(char const)` in order to allow
+   'S', 'W', 's' and 'w' as valid characters.
```
h-2:
```suggestion
   also 's' and 'w' as valid characters.
```
The upper case letters are already valid by defalt.

joergi-w:
Thank you, it's done!

PR #1139, opened 2019-06-27
[FEATURE] begin_ra() and end_ra() for the gap_decorator.

smehringer:
This will provide random access iterators that are *not guaranteed to operate in constant time*.

### DIFF
--- a/include/seqan3/range/decorator/gap_decorator.hpp
+++ b/include/seqan3/range/decorator/gap_decorator.hpp
@@ -93,7 +93,7 @@ private:
      */
     class gap_decorator_iterator
     {
-    private:
+    protected:
         //!\brief Pointer to the underlying container structure.
         typename std::add_pointer_t<gap_decorator const> host{nullptr};
         //!\brief Stores the virtual position index for the seqan3::gap_decorator.
@@ -339,11 +339,160 @@ private:
         //!\}
     };
 
+    /*!\brief The iterator that moves over the seqan3::gap_decorator with random access semantics.
+     *
+     * \details
+     *
+     * \attention This iterator models std::RandomAccessIterator but **does not guarantee constant time access**.
+     *            Random access is provided in \f$O(\log k)\f$ time (where `k` is the number of gaps).
+     *
+     * This iterator returns values when dereferenced, not references, i.e. it does not model the
+     * [Cpp17RandomAccessIterator](https://en.cppreference.com/w/cpp/named_req/RandomAccessIterator)
+     * requirements of the STL. It does model the C++20 std::RandomAccessIterator.
+     */
+    class gap_decorator_iterator_ra : public gap_decorator_iterator
+    {
+    private:
+        //!\brief The base type.
+        using base_t = gap_decorator_iterator;
+
+    public:
+        /*!\name Associated types
+         * \brief Make the parent's member types visible.
+         * \{
+         */
+        //!\brief Type for distances between iterators.
+        using difference_type = typename gap_decorator::difference_type;
+        //!\brief Value type of container elements.
+        using value_type = typename gap_decorator::value_type;
+        //!\brief Reference type defined by container (which equals the const reference type).
+        using reference = typename gap_decorator::const_reference;
+        //!\brief Pointer of container value type.
+        using pointer = value_type *;
+        //!\brief Signals that this is a random access iterator.
+        using iterator_category = std::random_access_iterator_tag;
+        //!\}
+
+        /*!\name Constructors, destructor and assignment
+         * \{
+         */
+        constexpr gap_decorator_iterator_ra()                                              = default; //!< Defaulted.
+        constexpr gap_decorator_iterator_ra(gap_decorator_iterator_ra const &)             = default; //!< Defaulted.
+        constexpr gap_decorator_iterator_ra & operator=(gap_decorator_iterator_ra const &) = default; //!< Defaulted.
+        constexpr gap_decorator_iterator_ra (gap_decorator_iterator_ra &&)                 = default; //!< Defaulted.
+        constexpr gap_decorator_iterator_ra & operator=(gap_decorator_iterator_ra &&)      = default; //!< Defaulted.
+        ~gap_decorator_iterator_ra()                                                       = default; //!< Defaulted.
+
+        //!\brief Construct from seqan3::gap_decorator and initialise members.
+        explicit constexpr gap_decorator_iterator_ra(gap_decorator const & host) noexcept : base_t{host}
+        {}
+
+        //!\brief Construct from seqan3::gap_decorator and explicit position.
+        constexpr gap_decorator_iterator_ra(gap_decorator const & host,
+                                            typename gap_decorator::size_type const pos) noexcept : base_t{host, pos}
+        {}
+        //!\}
+
+        /*!\name Arithmetic operators
+        * \{
+        */
+        //!\brief Pre-increment, returns updated iterator.
+        constexpr gap_decorator_iterator_ra & operator++() noexcept
+        {
+            base_t::operator++();
+            return *this;
+        }
+
+        //!\brief Pre-decrement, returns updated iterator.
+        constexpr gap_decorator_iterator_ra & operator--() noexcept
+        {
+            base_t::operator--();
+            return *this;
+        }
+
+        //!\brief Post-increment, returns previous iterator state (delegates to pre-increment).
+        constexpr gap_decorator_iterator_ra operator++(int) noexcept
+        {
+            gap_decorator_iterator_ra cpy{*this};
+            ++(*this);
+            return cpy;
+        }
+
+        //!\brief Post-decrement, returns previous iterator state (delegates to pre-decrement).
+        constexpr gap_decorator_iterator_ra operator--(int) noexcept
+        {
+            gap_decorator_iterator_ra cpy{*this};
+            --(*this);
+            return cpy;
+        }
+
+        //!\brief Forward this iterator.
+        constexpr gap_decorator_iterator_ra & operator+=(difference_type const skip) noexcept
+        {
+            this->jump(this->pos + skip);
+            return *this;
+        }
+
+        //!\brief Forward copy of this iterator.
+        constexpr gap_decorator_iterator_ra operator+(difference_type const skip) const noexcept
+        {
+            return gap_decorator_iterator_ra{*(this->host), this->pos + skip};
+        }
+
+        //!\brief Non-member operator+ delegates to non-friend operator+.
+        constexpr friend gap_decorator_iterator_ra operator+(difference_type const skip,
+                                                             gap_decorator_iterator_ra const & it) noexcept
+        {
+            return it + skip;
+        }
+
+        //!\brief Decrement iterator by skip.
+        constexpr gap_decorator_iterator_ra & operator-=(difference_type const skip) noexcept
+        {
+            this->jump(this->pos - skip);
+            return *this;
+        }
+
+        //!\brief Return decremented copy of this iterator.
+        constexpr gap_decorator_iterator_ra operator-(difference_type const skip) const noexcept
+        {
+            return gap_decorator_iterator_ra{*(this->host), this->pos - skip};
+        }
+
+        //!\brief Non-member operator- delegates to non-friend operator-.
+        constexpr friend gap_decorator_iterator_ra operator-(difference_type const skip,
+                                                             gap_decorator_iterator_ra const & it) noexcept
+        {
+            return it - skip;
+        }
+
+        //!\brief Returns the distance between two iterators.
+        constexpr difference_type operator-(gap_decorator_iterator_ra const lhs) const noexcept
+        {
+            return static_cast<difference_type>(this->pos - lhs.pos);
+        }
+        //!\}
+
+        /*!\name Element access
+        * \{
+        */
+        //!\brief Return underlying container value currently pointed at.
+        constexpr reference operator[](difference_type const n) const noexcept
+        {
+            return *(*this + n);
+        }
+        //!\}
+    };
+
     //!\brief The iterator type of this container (a bidirectional iterator).
     using iterator = gap_decorator_iterator;
     //!\brief The const_iterator equals the iterator type. Since no references are ever returned and thus the underlying
     //!        sequence cannot be modified through the iterator there is no need for const.
     using const_iterator = iterator;
+    //!\brief The pseudo random access iterator type (provides functionality but **not in constant time**).
+    using iterator_ra = gap_decorator_iterator_ra;
+    //!\brief The const_iterator_ra equals the iterator_ra type.
+    using const_iterator_ra = iterator_ra;
 
 public:
     /*!\name Range-associated member types
@@ -578,8 +727,36 @@ public:
         return const_iterator{*this};
     }
 
-    /*!\brief Returns an iterator to the element following the last element of the decorator.
-     * \returns Iterator to the behind last element.
+    /*!\brief Returns a pseudo random access iterator to the first element of the container.
+     * \returns Iterator to the first element.
+     *
+     * If the container is empty, the returned iterator will be equal to end().
+     *
+     * \attention The returned iterator models the std::ranges::RandomAccessIterator but
+     *            **the operations on it do not guarantee constant time**.
+     *            Random access is provided in \f$O(\log k)\f$ time (where `k` is the number of gaps).
+     *
+     * ### Complexity
+     *
+     * Constant.
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    const_iterator_ra begin_ra() const noexcept
+    {
+        return gap_decorator_iterator_ra{*this};
+    }
+
+    //!\copydoc begin_ra()
+    const_iterator_ra cbegin_ra() const noexcept
+    {
+        return gap_decorator_iterator_ra{*this};
+    }
+
+    /*!\brief Returns an iterator pointing behind the last element of the decorator.
+     * \returns Iterator pointing behind the last element.
      *
      * \attention This element acts as a placeholder; attempting to dereference it results in undefined behaviour.
      *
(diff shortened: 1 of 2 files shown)


include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -336,6 +336,152 @@ class gap_decorator
...
+        {
+            gap_decorator_iterator_ra cpy{*this};
+            --(*this);
+            return cpy;
+        }
+
+        //!\brief Forward this iterator.
+        constexpr gap_decorator_iterator_ra & operator+=(difference_type const skip) noexcept
+        {
+            *this = gap_decorator_iterator_ra{*(this->host), this->pos + skip};
```
joergi-w: Can you avoid reconstructing the whole object?
Like setting the new pos and trigger the gap counting?

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -356,6 +502,10 @@ class gap_decorator
     //!\brief The const_iterator equals the iterator type. Since no references are ever returned and thus the underlying
     //!        sequence cannot be modified through the iterator there is no need for const.
     using const_iterator = iterator;
+    //!\brief The pseudo random access iterator type (provides functionality but **not in constant time**).
+    using iterator_ra = gap_decorator_iterator_ra;
+    //!\brief The const_iterator_ra equals the iterator_ra type.
+    using const_iterator_ra = iterator_ra;
```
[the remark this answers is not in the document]
smehringer: Yes, because the decorator can never be modified over the iterators, so every iterator is a const iterator anyway. The reason: First a decorator only "decorates another sequence with gaps", so it should not be able to alter the underlying sequence. Thus `*it = 'A'_dna4` would never be allowed and for consistency `*it = gap{}` is also not allowed but the decorator mus be modified through "insert_gap" functions.

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -336,6 +336,152 @@ class gap_decorator
...
+        //!\brief Value type of container elements.
+        using value_type = typename gap_decorator::value_type;
+        //!\brief Const-reference type defined by container (which equals the reference type).
+        using reference = typename gap_decorator::const_reference; // = reference
+        //!\brief Equals reference type.
+        using const_reference = reference;
+        //!\brief Pointer of container value type.
+        using pointer = value_type *;
+        //!\brief Signals that this is a random access iterator.
+        using iterator_category = std::random_access_iterator_tag;
```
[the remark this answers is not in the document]
smehringer: We currently use the ranges concepts pre-C++20 which checks this tag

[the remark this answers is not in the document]
joergi-w: Okay 👍

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -574,8 +727,35 @@ class gap_decorator
...
+        return gap_decorator_iterator_ra{*this};
+    }
+
+    //!\copydoc begin_ra()
+    const_iterator_ra cbegin_ra() const noexcept
+    {
+        return gap_decorator_iterator_ra{*this};
+    }
+
+    /*!\brief Returns an iterator pointing behind last element of the decorator.
```
joergi-w:
```suggestion
    /*!\brief Returns an iterator pointing behind the last element of the decorator.
```

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -597,6 +777,33 @@ class gap_decorator
     {
         return const_iterator{*this, size()};
     }
+
+    /*!\brief Returns a random access iterator pointing behind last element of the decorator.
```
joergi-w:
```suggestion
    /*!\brief Returns a random access iterator pointing behind the last element of the decorator.
```

test/unit/range/decorator/gap_decorator_test.cpp
```diff
@@ -58,6 +57,76 @@ class aligned_sequence<gap_decorator<inner_type_>> : public ::testing::Test
...
+using const_random_access_iterator_type = decltype(std::declval<decorator_t &>().cbegin_ra());
+
+template <>
+struct iterator_fixture<iterator_type> : public ::testing::Test
+{
+    using iterator_tag = std::bidirectional_iterator_tag;
+    static constexpr bool const_iterable = true;
+
+    std::vector<dna4> const vec{"ACTGACTG"_dna4};
+    decorator_t dec = [this] ()
```
joergi-w: Why do we need to capture the `this` pointer (as copy)? 
I don't see that it is used, also in `range_to_compare` below.

smehringer: I use the member `vec` inside both of the lambdas and since `this` is just a pointer the copy is cheap.

h-2: Keep in mind that the rules for capturing this have change in c++20. Please build your code once with g++9 in std=c++2a mode to make sure it produces no warnings.

smehringer: How do I build with c++2 in our build system . Does simply setting the std flag override/adds it?

h-2: Yes, plain cmake.

joergi-w: Isn’t capturing this->vec the best solution then?

smehringer: ok then it works

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -336,6 +339,152 @@ class gap_decorator
...
+        {
+            gap_decorator_iterator_ra cpy{*this};
+            --(*this);
+            return cpy;
+        }
+
+        //!\brief Forward this iterator.
+        constexpr gap_decorator_iterator_ra & operator+=(difference_type const skip) noexcept
+        {
+            this->jump(this->pos + skip);
```
rrahn: this means, you always start from scratch looking for the new position right? So always a binary search. 
It might be more efficient to check first if you would stay in the same node and conduct only binary search if this is not the case (so for small jumps it would be as efficient as it can be). But I am not familiar with the data structures used and if you can access the size of the current node (but I guess it should)

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -574,8 +727,35 @@ class gap_decorator
...
 
-    /*!\brief Returns an iterator to the element following the last element of the decorator.
-     * \returns Iterator to the behind last element.
+    /*!\brief Returns a pseudo random access iterator to the first element of the container.
+     * \returns Iterator to the first element.
+     *
+     * If the container is empty, the returned iterator will be equal to end().
+     *
+     * \attention The returned iterator models the std::ranges::RandomAccessIterator but
+     *            **the operations on it do not guarantee constant time**.
```
rrahn: but what do they guarantee?

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -574,8 +727,35 @@ class gap_decorator
...
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    iterator_ra begin_ra() noexcept
+    {
+        return gap_decorator_iterator_ra{*this};
+    }
+
```
rrahn: `begin_ra() const` is missing

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -597,6 +777,33 @@ class gap_decorator
...
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    iterator_ra end_ra() const noexcept
+    {
+        return gap_decorator_iterator_ra{*this, size()};
+    }
+
```
rrahn: missing `end_ra()` non const version and the const version should return const_iterator. For the docu

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -574,8 +727,35 @@ class gap_decorator
...
+     *
+     * ### Complexity
+     *
+     * Constant.
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    iterator_ra begin_ra() noexcept
```
rrahn: ok here you are just missing the `const`

rrahn: but I guess it looks weird in the docu if there is a const and you return the non-const iterator. So maybe just add a separate const version that returns a const iterator. That would less irritate the people I guess.

test/unit/range/decorator/gap_decorator_test.cpp
```diff
@@ -58,6 +57,70 @@ class aligned_sequence<gap_decorator<inner_type_>> : public ::testing::Test
...
+        decorator_t::const_iterator_ra cbegin(){ return range.cbegin_ra(); }
+        decorator_t::const_iterator_ra cend()  { return range.cend_ra(); }
+    };
+
+    expose_random_access_iterator test_range{};
+};
+
+using test_type = ::testing::Types<decorator_t::iterator, decorator_t::iterator_ra>;
+
+INSTANTIATE_TYPED_TEST_CASE_P(iterator_fixture, iterator_fixture, test_type);
```
rrahn: can you please give it a more meaning full name like gap_decorator_iterator instead of iterator_fixture?

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -336,6 +339,152 @@ class gap_decorator
...
     };
 
+    /*!\brief The iterator that moves over the seqan3::gap_decorator.
+     *
+     * \details
+     *
+     * \attention This iterator models std::RandomAccessIterator but **does not guarantee constant time access**.
+     *
+     * This iterator returns values when dereferenced, not references, i.e. it does not satisfy the semantic
+     * requirements of [LegacyForwardIterator](https://en.cppreference.com/w/cpp/named_req/ForwardIterator). It does
```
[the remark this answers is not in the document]
smehringer: actually it is the BidirectionalIterator

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -336,6 +339,152 @@ class gap_decorator
...
+     * model the C++20 std::BidirectionalIterator (and std::ForwardIterator implicitly).
```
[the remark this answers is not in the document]
smehringer: copy and paste error

include/seqan3/range/decorator/gap_decorator.hpp
smehringer: yes you are right but this is not trivial and would require some thought. I will add this to the todos and probably tackle it when revising the insert_gap function (to be more efficient with the iterator hint) in a separate PR?

include/seqan3/range/decorator/gap_decorator.hpp
smehringer: They operate in `log(g)` where `g` is the number of gaps. I added this in the description

include/seqan3/range/decorator/gap_decorator.hpp
smehringer: Ok so the gap decorator iterator is always const, because I do not allow to change the data structure via the index. thats why the gap_decorator did not have a non-const begin in the first place. I can explicitely delete it? and the const version should then return the const iterator of course.

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -336,6 +339,155 @@ class gap_decorator
...
+        }
+
+        //!\brief Returns the distance between two iterators.
+        constexpr difference_type operator-(gap_decorator_iterator_ra const lhs) const noexcept
+        {
+            return static_cast<difference_type>(this->pos - lhs.pos);
+        }
+        //!\}
+
+        /*!\name Element access
```
rrahn: can you check if this looks fine in the docu? I had a lot of trouble getting named groups being merged if they are repeated in the derived class. For me it always added a second group with the same name. It might help to make a using declaration of the *-operator or a SEQAN3_DOXYGEN_ONLY expression😞

smehringer: with the SEQAN3_DOXYGEN_ONLY and  even without, doxygen does not put them in the same group but makes two group with the same name. I think this is a general bug. Dies the SEQAN3_DOXYGEN_ONLY hack work sometimes?

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -336,6 +339,152 @@ class gap_decorator
...
+        using reference = typename gap_decorator::const_reference; // = reference
+        //!\brief Equals reference type.
+        using const_reference = reference;
+        //!\brief Pointer of container value type.
+        using pointer = value_type *;
+        //!\brief Signals that this is a random access iterator.
+        using iterator_category = std::random_access_iterator_tag;
+        //!\}
+
+        /*!\name Constructors/Destructors
```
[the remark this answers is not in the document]
rrahn: not resolved

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -336,6 +339,152 @@ class gap_decorator
...
+        }
+
+        //!\brief Non-member operator- delegates to non-friend operator-.
+        constexpr friend gap_decorator_iterator_ra operator-(difference_type const skip,
+                                                             gap_decorator_iterator_ra const & it) noexcept
+        {
+            return it - skip;
+        }
+
+        //!\brief Return offset between this and remote iterator's position.
```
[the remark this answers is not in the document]
rrahn: not resolved

include/seqan3/range/decorator/gap_decorator.hpp
rrahn: Well it is not a bug. It is simply a missing feature of doxygen. Since you need the base class as a standalone class you won't get around this i guess. I thought maybe he sees that if the variable was overloaded from the base class it will not show the base class version. But it seems this is not the case then?

include/seqan3/range/decorator/gap_decorator.hpp
rrahn: so what we need is a `begin_ra() const` and a `cbegin_ra()`, right? Becase it is always a const iterator. But the begin_ra is still not marked const yet?

include/seqan3/range/decorator/gap_decorator.hpp
rrahn: But now end_ra should return a const_iterator because it is const. Actually why is there a const_iteator and a iterator type if they are the same. I suggest to just remove the const_iterator type and make clear in the documentation that there is no non-const version of it.

test/unit/range/decorator/gap_decorator_test.cpp
rrahn: Did you revert your last changes? I see a lot of stuff that is not resolved but I thought in a previous version you already did? Either something is messing with my eyes or something strange happened. So is the instantiated test still called iterator_fixture in your source?

include/seqan3/range/decorator/gap_decorator.hpp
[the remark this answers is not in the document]
smehringer: hm it seems my changes have not been pushed :thinking:

include/seqan3/range/decorator/gap_decorator.hpp
smehringer: hm but isn't the member type `const_iterator` part of the (although not enforced) interface of a container/decorator? I can remove it but I am not sure about the usability.

include/seqan3/range/decorator/gap_decorator.hpp
smehringer: yes sorry I wasn't careful with the changes.
we are missing the const.
and we can discuss if we should remove the `const_iterator`

test/unit/range/decorator/gap_decorator_test.cpp
smehringer: something is messing with my PR o.O
I'll revisit all the review comments. I am sure I changed this already.

include/seqan3/range/decorator/gap_decorator.hpp
smehringer: No :( I tried it with and without SEQAN3_DOXYGEN_ONLY

smehringer:
@rrahn I think there is only the open question left, namely whether we should remove the const_iterator type member. 
I found that in the c++20 draft a `const_iterator` type member is required for containers but not for ranges/views.
Since the decorator is between the two, I guess we can choose.

rrahn:
> I found that in the c++20 draft a const_iterator type member is required for containers but not for ranges/views.
Since the decorator is between the two, I guess we can choose.

Well, I think we need to discuss this. In general I think that the decorator is more a container than a view. Of course it decorates another container, but stores additional values on top of it. So from the semantics I would say it models Container. In this case we need both iterator and const_iterator. Still we only need one function the cbegin and begin const version, since both types would be the same anyway. But I mean the gap decorator is just a special form, since it could of course change the value if the gap would not be binary but in the case of other decorators we actually can erase/insert/modify the values, e.g. journal_decorator

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -336,6 +339,163 @@ class gap_decorator
...
+         */
+        constexpr gap_decorator_iterator_ra()                                              = default; //!< Defaulted.
+        constexpr gap_decorator_iterator_ra(gap_decorator_iterator_ra const &)             = default; //!< Defaulted.
+        constexpr gap_decorator_iterator_ra & operator=(gap_decorator_iterator_ra const &) = default; //!< Defaulted.
+        constexpr gap_decorator_iterator_ra (gap_decorator_iterator_ra &&)                 = default; //!< Defaulted.
+        constexpr gap_decorator_iterator_ra & operator=(gap_decorator_iterator_ra &&)      = default; //!< Defaulted.
+        ~gap_decorator_iterator_ra()                                                       = default; //!< Defaulted.
+
+        //!\brief Construct from seqan3::gap_decorator and initialise members.
+        explicit constexpr gap_decorator_iterator_ra(gap_decorator const & host_) noexcept : base_t{host_}
```
rrahn: can you please remove the trailing underscore?

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -336,6 +339,163 @@ class gap_decorator
...
+         * \{
+         */
+        //!\brief Type for distances between iterators.
+        using difference_type = typename gap_decorator::difference_type;
+        //!\brief Value type of container elements.
+        using value_type = typename gap_decorator::value_type;
+        //!\brief Const-reference type defined by container (which equals the reference type).
+        using reference = typename gap_decorator::const_reference; // = reference
+        //!\brief Equals reference type.
+        using const_reference = reference;
```
rrahn: why do you need to define this?

smehringer: This is required by a container.. since we do not model container I could remove this. I don't care?

smehringer: @rrahn ping

rrahn: But it is not required by the iterator. So please remove it then.

smehringer:
@rrahn. I rebased onto current master but Jenkins still fails?

PR #1140, opened 2019-06-27
Add flag -Wno-unused for snippet tests.

joshuak94:


h-2:
Did you check that `-Wno-unused` includes all the unused warnings?

joshuak94:
Not quite, but I did check that -Wunused includes all of the unused warnings. Also the GCC manual says that any flag can be prepended with a "no" to exclude it.

PR #1141, opened 2019-06-27
[FEATURE] Add lzcnt, tzcnt, popcount

eseiler:
~~The SDSL functions are not constexpr yet. Since we only support GCC they are currently never used.~~

[~~SDSL PR~~](https://github.com/xxsds/sdsl-lite/pull/68) - merged

### DIFF
--- a/include/seqan3/core/bit_manipulation.hpp
+++ b/include/seqan3/core/bit_manipulation.hpp
@@ -15,6 +15,8 @@
 
 #include <meta/meta.hpp>
 
+#include <sdsl/bits.hpp>
+
 #include <climits>
 #include <utility>
 
@@ -77,7 +79,127 @@ constexpr size_t next_power_of_two(size_t n)
     return n + 1;
 }
 
-/*!\brief Returns the position of the most significant bit (counting from right to left).
+/*!\brief Returns the number of 1-bits.
+* \ingroup core
+*
+* \param[in] n An unsigned integer.
+*
+* \returns The number of 1-bits.
+*
+* ### Example
+*
+* \include test/snippet/core/detail/popcount.cpp
+*
+* ### Exception
+*
+* No-throw guarantee.
+*
+* ### Thread-safety
+*
+* Thread safe.
+*
+* ### Complexity
+*
+* Constant.
+*/
+template <std::UnsignedIntegral unsigned_t>
+constexpr uint8_t popcount(unsigned_t const n) noexcept
+{
+#if defined(__GNUC__)
+    if constexpr (sizeof(unsigned_t) == sizeof(unsigned long long))
+        return __builtin_popcountll(n);
+    else if constexpr (sizeof(unsigned_t) == sizeof(unsigned long))
+        return __builtin_popcountl(n);
+    else
+        return __builtin_popcount(n);
+#else
+    return sdsl::bits::cnt(n);
+#endif
+}
+
+/*!\brief Returns the number of leading 0-bits, starting at the most significant bit position.
+* \ingroup core
+*
+* \param[in] n An unsigned integer.
+*
+* \attention *n = 0* is a special case and is undefined behaviour.
+*
+* \returns The number of leading 0-bits.
+*
+* ### Example
+*
+* \include test/snippet/core/detail/count_leading_zeros.cpp
+*
+* ### Exception
+*
+* No-throw guarantee.
+*
+* ### Thread-safety
+*
+* Thread safe.
+*
+* ### Complexity
+*
+* Constant.
+*/
+template <std::UnsignedIntegral unsigned_t>
+constexpr uint8_t count_leading_zeros(unsigned_t const n) noexcept
+{
+    assert(n > 0); // n == 0 might have undefined behaviour
+#if defined(__GNUC__)
+    if constexpr (sizeof(unsigned_t) == sizeof(unsigned long long))
+        return __builtin_clzll(n);
+    else if constexpr (sizeof(unsigned_t) == sizeof(unsigned long))
+        return __builtin_clzl(n);
+    else
+        return __builtin_clz(n) + sizeof_bits<unsigned_t> - sizeof_bits<unsigned int>;
+#else
+    return sizeof_bits<unsigned_t> - sdsl::bits::hi(n) - 1;
+#endif
+}
+
+/*!\brief Returns the number of trailing 0-bits, starting at the least significant bit position.
+ * \ingroup core
+ *
+ * \param[in] n An unsigned integer.
+ *
+ * \attention *n = 0* is a special case and is undefined behaviour.
+ *
+ * \returns The number of trailing 0-bits.
+ *
+ * ### Example
+ *
+ * \include test/snippet/core/detail/count_trailing_zeros.cpp
+ *
+ * ### Exception
+ *
+ * No-throw guarantee.
+ *
+ * ### Thread-safety
+ *
+ * Thread safe.
+ *
+ * ### Complexity
+ *
+ * Constant.
+ */
+template <std::UnsignedIntegral unsigned_t>
+constexpr uint8_t count_trailing_zeros(unsigned_t const n) noexcept
+{
+    assert(n > 0); // n == 0 might have undefined behaviour
+#if defined(__GNUC__)
+    if constexpr (sizeof(unsigned_t) == sizeof(unsigned long long))
+        return __builtin_ctzll(n);
+    else if constexpr (sizeof(unsigned_t) == sizeof(unsigned long))
+        return __builtin_ctzl(n);
+    else
+        return __builtin_ctz(n);
+#else
+    return sdsl::bits::lo(n);
+#endif
+}
+
+/*!\brief Returns the position (0-based) of the most significant bit.
  * \ingroup core
  *
  * \param[in] n An unsigned integer.
@@ -85,22 +207,31 @@ constexpr size_t next_power_of_two(size_t n)
  * \attention *n = 0* is a special case and is undefined behaviour.
  *
  * \returns The position of the most significant bit.
+ *
+ * ### Example
+ *
+ * \include test/snippet/core/detail/most_significant_bit_set.cpp
+ *
+ * ### Exception
+ *
+ * No-throw guarantee.
+ *
+ * ### Thread-safety
+ *
+ * Thread safe.
+ *
+ * ### Complexity
+ *
+ * Constant.
  */
 template <std::UnsignedIntegral unsigned_t>
-constexpr uint8_t bit_scan_reverse(unsigned_t n)
+constexpr uint8_t most_significant_bit_set(unsigned_t const n) noexcept
 {
     assert(n > 0); // n == 0 might have undefined behaviour
 #if defined(__GNUC__)
-    if constexpr (sizeof(unsigned_t) == sizeof(unsigned long long))
-        return sizeof_bits<unsigned long long> - __builtin_clzll(n) - 1;
-    else if constexpr (sizeof(unsigned_t) == sizeof(unsigned long))
-        return sizeof_bits<unsigned long> - __builtin_clzl(n) - 1;
-    else
-        return sizeof_bits<unsigned> - __builtin_clz(n) - 1;
+    return sizeof_bits<unsigned_t> - count_leading_zeros(n) - 1;
 #else
-    uint8_t i = 0;
-    for (; n != 0; n >>= 1, ++i);
-    return i - 1;
+    return sdsl::bits::hi(n);
 #endif
 }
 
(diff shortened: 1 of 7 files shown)


test/unit/core/bit_manipulation_test.cpp
```diff
@@ -108,6 +108,103 @@ TYPED_TEST(unsigned_operations, bit_scan_reverse)
...
+            EXPECT_EQ(count_trailing_zeros(n), cnt) << "n " << n << " should have " << cnt << " trailing zeros.";
+        }
+    }
+}
+
+// http://graphics.stanford.edu/~seander/bithacks.html#NextBitPermutation
+template <std::UnsignedIntegral unsigned_t>
+unsigned_t permute_bits(unsigned_t v)
+{
+    if (v & (unsigned_t{1u} << (sizeof_bits<unsigned_t> - 1)))
```
eseiler: As a note:
The function produces permutations with the same number of 1 bits, e.g.
```
0b0000'0001
0b0000'0010
0b0000'0100
...
```
For some reason, for `uint32_t` and `uint64_t`, permuting `0b1000'00..` results in setting all bits to 1, even though it probably should not (http://coliru.stacked-crooked.com/a/e445e8159c8e45ce). had this problem locally and on CI. So I added the if to catch this.

marehr: The compiler might changes it to:

http://coliru.stacked-crooked.com/a/9d3b79279787b6ce

Which produces 111111111111111111.

test/unit/core/bit_manipulation_test.cpp
```diff
@@ -108,6 +108,103 @@ TYPED_TEST(unsigned_operations, bit_scan_reverse)
...
+    EXPECT_EQ(t5, sizeof_bits<unsigned_t> - 3u);
+    EXPECT_EQ(t6, sizeof_bits<unsigned_t> - 8u);
+
+    for (uint8_t cnt = 0; cnt < sizeof_bits<unsigned_t>; ++cnt)
+    {
+        unsigned_t start = std::numeric_limits<unsigned_t>::max() >> cnt;
+        unsigned_t end = start >> 1u;
+        for (unsigned_t n = start, k = 0u; n < end && k < max_iterations; ++n, ++k)
+        {
+            EXPECT_EQ(count_leading_zeros(n), cnt) << "n " << n << " should have " << cnt << " leading zeros.";
```
marehr: can you add the sdsl versions. They might behave differently and we will never know, because we can't test them.

test/unit/core/bit_manipulation_test.cpp
marehr: The method of http://graphics.stanford.edu/~seander/bithacks.html#NextBitPermutation does not seem to work when the msb is set. So the fix seems fine.

include/seqan3/core/bit_manipulation.hpp
```diff
@@ -77,33 +81,94 @@ constexpr size_t next_power_of_two(size_t n)
...
+        return __builtin_ctz(n);
 #else
-    uint8_t i = 0;
-    for (; n != 0; n >>= 1, ++i);
-    return i - 1;
+    return sdsl::bits::lo(n);
 #endif
 }
 
+/*!\brief Returns the position of the most significant bit (counting from right to left).
```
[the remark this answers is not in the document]
eseiler: > 0 based position? (add this information)

Ok.
> 
> also: does this position depend on little /big endianness?

Builtins are defined as `starting at the most/least significant bit position.` So yes, it depends since you interpret the bits differently.
> 
> and: do we need bit_scan_reverse at all?
> isn't `sizeof_bits<unsigned_t> - count_leading_zeros(n) - 1` == `count_trailing_zeros(n)`?

```cpp
uint16_t x = 0b0000'0001'0000'0000
uint8_t pos = sizeof_bits<uint16_t> - count_leading_zeros(x) - 1; 
//    8 = 16 - 7 - 1
uint8_t ctz = count_trailing_zeros(x);
//    ctz = 8
```
But
```cpp
uint16_t x = 0b0000'0001'0000'0010
uint8_t pos = sizeof_bits<uint16_t> - count_leading_zeros(x) - 1; 
//    8 = 16 - 7 - 1
uint8_t ctz = count_trailing_zeros(x);
//    ctz = 1
```

include/seqan3/core/bit_manipulation.hpp
```diff
@@ -77,33 +81,94 @@ constexpr size_t next_power_of_two(size_t n)
...
+ * \ingroup core
+ *
+ * \param[in] n An unsigned integer.
+ *
+ * \attention *n = 0* is a special case and is undefined behaviour.
+ *
+ * \returns The position of the most significant bit.
+ */
+template <std::UnsignedIntegral unsigned_t>
+constexpr uint8_t bit_scan_reverse(unsigned_t n)
```
smehringer:
```suggestion
constexpr uint8_t most_significant_bit(unsigned_t n)
```

include/seqan3/core/bit_manipulation.hpp
```diff
@@ -77,33 +81,94 @@ constexpr size_t next_power_of_two(size_t n)
...
+{
+    assert(n > 0); // n == 0 might have undefined behaviour
+    return sizeof_bits<unsigned_t> - count_leading_zeros(n) - 1;
```
smehringer: this does too many additions if the build in is not available:
```cpp
#if defined(__GNUC__)
    return sizeof_bits<unsigned_t> - sdsl::bits::hi(n) - 1;
#else 
    return sdsl::bits::hi(n);
```

eseiler:
> Is the inclusion of the new submodule sdsl revision intended?

Yes, I marked the sdsl functions constexpr.

include/seqan3/core/bit_manipulation.hpp
```diff
@@ -15,6 +15,10 @@
 
 #include <meta/meta.hpp>
 
+#ifndef __GNUC__
```
rrahn: Why this guard?

rrahn: Ah you say if it is not defined and then you use a fallback that is implemented in the sdsl. I see.

test/unit/core/bit_manipulation_test.cpp
```diff
@@ -81,29 +83,135 @@ class unsigned_operations : public ::testing::Test
...
 
-    for (uint8_t position = 0; position < 8u * sizeof(unsigned_t); ++position)
+    for (uint8_t position = 0; position < sizeof_bits<unsigned_t>; ++position)
     {
         unsigned_t start = unsigned_t{1u} << position;
         unsigned_t end = start << 1u;
         for (unsigned_t n = start, k = 0u; n < end && k < max_iterations; ++n, ++k)
         {
-            EXPECT_EQ(bit_scan_reverse(n), position) << "The position of the msb of " << n << " should be " << position;
+            EXPECT_EQ(sdsl::bits::hi(n), position) << "[SDSL] The position of the msb of " << n << " should be "
```
rrahn: why do we test sdsl stuff in SeqAn code?

test/unit/core/bit_manipulation_test.cpp
```diff
@@ -81,29 +83,135 @@ class unsigned_operations : public ::testing::Test
...
+    EXPECT_EQ(t5, sizeof_bits<unsigned_t> - 3u);
+    EXPECT_EQ(t6, sizeof_bits<unsigned_t> - 8u);
+
+    for (uint8_t cnt = 0; cnt < sizeof_bits<unsigned_t>; ++cnt)
+    {
+        unsigned_t start = std::numeric_limits<unsigned_t>::max() >> cnt;
+        unsigned_t end = start >> 1u;
+        for (unsigned_t n = start, k = 0u; n < end && k < max_iterations; ++n, ++k)
+        {
+            EXPECT_EQ(sizeof_bits<unsigned_t> - sdsl::bits::hi(n) - 1, n) << "[SDSL] n " << n << " should have "
```
rrahn: same here

test/unit/core/bit_manipulation_test.cpp
```diff
@@ -9,6 +9,8 @@
 
 #include <gtest/gtest.h>
 
+#include <sdsl/bits.hpp>
```
rrahn: if you include it here, we might not spot issues, if the header was not correctly included in the bit_manipulation file or removed. Also the header test should catch them, but I don't see the necessity to add this here.

test/unit/core/bit_manipulation_test.cpp
```diff
@@ -81,29 +83,135 @@ class unsigned_operations : public ::testing::Test
...
+        {
+            EXPECT_EQ(sdsl::bits::lo(n), cnt) << "[SDSL] n " << n << " should have " << cnt << " trailing zeros.";
+            EXPECT_EQ(count_trailing_zeros(n), cnt) << "n " << n << " should have " << cnt << " trailing zeros.";
+        }
+    }
+}
+
+// http://graphics.stanford.edu/~seander/bithacks.html#NextBitPermutation
+template <std::UnsignedIntegral unsigned_t>
+unsigned_t permute_bits(unsigned_t v)
```
rrahn: and we need this only for the tests?

test/unit/core/bit_manipulation_test.cpp
eseiler: To make sure the SDSL fallback works.
Was requested by @marehr.

test/unit/core/bit_manipulation_test.cpp
eseiler: Ok, I just always include sdsl/bits now in the header

test/unit/core/bit_manipulation_test.cpp
eseiler: Yes, to follow the scheme of the other tests, I wanted to compute all integers with a given number of 1's in them.

include/seqan3/core/bit_manipulation.hpp
marehr: NO WHY?!

smehringer: (you talk about rude review comments again :wink:)
because `bit_scan_reverse` doesn't tell me what's happening? Because it is actually the position of the most significant bit.

include/seqan3/core/bit_manipulation.hpp
```diff
@@ -77,30 +79,159 @@ constexpr size_t next_power_of_two(size_t n)
...
+*
+* No-throw guarantee.
+*
+* ### Thread-safety
+*
+* Thread safe.
+*
+* ### Complexity
+*
+* Constant.
```
rrahn: sdsl bits count does some magic bits trickery to get constant time? Or is there somewhere a fall back to O(n) bits?

include/seqan3/core/bit_manipulation.hpp
```diff
@@ -77,30 +79,159 @@ constexpr size_t next_power_of_two(size_t n)
...
```
rrahn: same here, is it guaranteed that sdsl::hi is a constant operation?

include/seqan3/core/bit_manipulation.hpp
```diff
@@ -77,30 +79,159 @@ constexpr size_t next_power_of_two(size_t n)
...
+ *
+ * No-throw guarantee.
+ *
+ * ### Thread-safety
+ *
+ * Thread safe.
+ *
+ * ### Complexity
+ *
+ * Constant.
```
rrahn: and here?

include/seqan3/core/bit_manipulation.hpp
```diff
@@ -77,30 +79,159 @@ constexpr size_t next_power_of_two(size_t n)
...
+        return __builtin_ctzl(n);
     else
-        return sizeof_bits<unsigned> - __builtin_clz(n) - 1;
+        return __builtin_ctz(n);
+#else
+    return sdsl::bits::lo(n);
+#endif
+}
+
+/*!\brief Returns the position (0-based) of the most significant bit, starting at the least significant bit position.
```
rrahn: but it scans from msb to lsb, doesn't it?

eseiler: Mhm, it's not really "scanning", or what do you mean?
`starting at the least significant bit position` refers to the positions, i.e. we count from right to left.

include/seqan3/core/bit_manipulation.hpp
eseiler: I looked a bit more into it and from what I've seen and read, it should be constant with all the lookup tables.
To be really sure, I would have to look it up in Donald Knuth's *The Art of Computer Programming*.

I can definitely say that it's never scanning the bits from left to right or such.

include/seqan3/core/bit_manipulation.hpp
rrahn: Ok, you want to indicate that the position is the postion when reading from right to left. But I think it is confusing. I think the first part of the sentence is clear enough about what will be returned. Also there is an example so it is clear how it is used. Maybe just drop the second part then.

PR #1144, opened 2019-06-28
[FIX] Incorrect copy/move of fm_index

### LINKED ISSUE #1121: Processing returned ranges from search segfaults
I have the following code, compiled with GCC 7.2.0 (c++17) and SeqAn 3.0.0:
```
void seq_search(const dna5_vector& query, const std::vector<fm_index<true,default_sdsl_index_type>& seq_idx)
{
   for (auto ref_it = seq_idx.begin(); ref_it != seq_idx.end(); ref_it++)
   {
      auto results = search(query, *ref_it);
      debug_stream << "There are " << results.size() << " hits.\n";
      debug_stream << results << '\n';
      if (results.size() > 0)
      {
         std::cerr << "found" << std::endl;
      }
   }
}
```
If `query` cannot be found then it's ok, but as soon as there is a match I get a segfault, I think at `results.size() > 0`. The number of hits doesn't print, and the result print line appears as `[1]` (which is wrong).

Any ideas what is causing this?
Is this because my libgcc/libstdc++ doesn't have an implementation of ranges?

eseiler:
Resolves #1121

### DIFF
--- a/include/seqan3/search/fm_index/fm_index.hpp
+++ b/include/seqan3/search/fm_index/fm_index.hpp
@@ -178,12 +178,39 @@ public:
     /*!\name Constructors, destructor and assignment
      * \{
      */
-    fm_index() = default;                             //!< Default constructor.
-    fm_index(fm_index const &) = default;             //!< Copy constructor.
-    fm_index & operator=(fm_index const &) = default; //!< Copy assignment.
-    fm_index(fm_index &&) = default;                  //!< Move constructor.
-    fm_index & operator=(fm_index &&) = default;      //!< Move assignment.
-    ~fm_index() = default;                            //!< Destructor.
+    fm_index() = default;              //!< Defaulted.
+
+    fm_index(fm_index const & rhs) :   //!< When copy constructing, also update internal data structures.
+        sigma{rhs.sigma}, index{rhs.index}, text_begin{rhs.text_begin}, text_begin_ss{rhs.text_begin_ss},
+        text_begin_rs{rhs.text_begin_rs}
+    {
+        text_begin_ss.set_vector(&text_begin);
+        text_begin_rs.set_vector(&text_begin);
+    }
+
+    fm_index(fm_index && rhs) :        //!< When move constructing, also update internal data structures.
+        sigma{std::move(rhs.sigma)}, index{std::move(rhs.index)}, text_begin{std::move(rhs.text_begin)},
+        text_begin_ss{std::move(rhs.text_begin_ss)}, text_begin_rs{std::move(rhs.text_begin_rs)}
+    {
+        text_begin_ss.set_vector(&text_begin);
+        text_begin_rs.set_vector(&text_begin);
+    }
+
+    fm_index & operator=(fm_index rhs) //!< When copy/move assigning, also update internal data structures.
+    {
+        index = std::move(rhs.index);
+        sigma = std::move(rhs.sigma);
+        text_begin = std::move(rhs.text_begin);
+        text_begin_ss = std::move(rhs.text_begin_ss);
+        text_begin_rs = std::move(rhs.text_begin_rs);
+
+        text_begin_ss.set_vector(&text_begin);
+        text_begin_rs.set_vector(&text_begin);
+
+        return *this;
+    }
+
+    ~fm_index() = default;             //!< Defaulted.
 
     /*!\brief Constructor that immediately constructs the index given a range. The range cannot be empty.
      * \tparam text_t The type of range to construct from; must model std::ranges::BidirectionalRange.
--- a/include/seqan3/search/fm_index/bi_fm_index.hpp
+++ b/include/seqan3/search/fm_index/bi_fm_index.hpp
@@ -120,12 +120,12 @@ public:
     /*!\name Constructors, destructor and assignment
      * \{
      */
-    bi_fm_index() = default;                                //!< Default constructor.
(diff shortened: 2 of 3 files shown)


include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -178,12 +178,25 @@ class fm_index
     /*!\name Constructors, destructor and assignment
      * \{
      */
-    fm_index() = default;                             //!< Default constructor.
-    fm_index(fm_index const &) = default;             //!< Copy constructor.
-    fm_index & operator=(fm_index const &) = default; //!< Copy assignment.
-    fm_index(fm_index &&) = default;                  //!< Move constructor.
-    fm_index & operator=(fm_index &&) = default;      //!< Move assignment.
-    ~fm_index() = default;                            //!< Destructor.
+    fm_index() = default;              //!< Default constructor.
```
smehringer:
```suggestion
    fm_index() = default;              //!< Defaulted.
```

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -178,12 +178,25 @@ class fm_index
...
+    fm_index(fm_index const & rhs)     //!< Copy constructor.
```
smehringer: If the constructor is not defaulted write why into the brief.

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -438,6 +452,34 @@ class fm_index
...
+        swap(lhs.text_begin, rhs.text_begin);
+        swap(lhs.text_begin_ss, rhs.text_begin_ss);
+        swap(lhs.text_begin_rs, rhs.text_begin_rs);
+
+        lhs.text_begin_ss.set_vector(&lhs.text_begin);
+        lhs.text_begin_rs.set_vector(&lhs.text_begin);
+        rhs.text_begin_ss.set_vector(&rhs.text_begin);
+        rhs.text_begin_rs.set_vector(&rhs.text_begin);
+    }
+
```
h-2: Since C++11 Copy'n'swap is typically implemented like this:

  * copy constructor with copy implementation
  * move constructor with move implementation
  * assignment operator that takes by value and std::swaps
  * You very rarely need to define a `.swap()` member since C++11. Always use std::swap which uses the move constructor automatically.

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -178,12 +178,26 @@ class fm_index
...
      * \{
      */
-    fm_index() = default;                             //!< Default constructor.
-    fm_index(fm_index const &) = default;             //!< Copy constructor.
-    fm_index & operator=(fm_index const &) = default; //!< Copy assignment.
-    fm_index(fm_index &&) = default;                  //!< Move constructor.
-    fm_index & operator=(fm_index &&) = default;      //!< Move assignment.
-    ~fm_index() = default;                            //!< Destructor.
+    fm_index() = default;              //!< Defaulted.
+    fm_index(fm_index const & rhs)     //!< When copy constructing, also set the pointers of rank and select support.
```
h-2: Do users of the fm_index (need to) know what rank and select support are?

include/seqan3/search/fm_index/fm_index.hpp
h-2: I would write it like this:

```cpp
fm_index(fm_index const & rhs) :
    index{rhs.index}, sigma{rhs.sigma}, text_begin{rhs.text_begin}, text_begin_ss{rhs.text_begin_ss}
{
    text_begin_ss.set_vector(&text_begin);
    text_begin_rs.set_vector(&text_begin);
}

fm_index(fm_index && rhs) :
    index{std::move(rhs.index)}, sigma{std::move(rhs.sigma)}, text_begin{std::move(rhs.text_begin)}, text_begin_ss{std::move(rhs.text_begin_ss)}
{
    text_begin_ss.set_vector(&text_begin);
    text_begin_rs.set_vector(&text_begin);
}

fm_index & operator=(fm_index rhs) //!< Delegate to swap().
{
    std::swap(*this, rhs);
    return *this;
}
```

This clearly highlights that only thing non-standard that happens is calling `set_vector` for two members.

eseiler: Mhm, using `std::swap` segfaults. It doesn't seem to set the pointers correctly.

include/seqan3/search/fm_index/fm_index.hpp
eseiler: `When copy constructing, also update internal data structures.` ?

include/seqan3/search/fm_index/fm_index.hpp
h-2: Hm, that sounds weird. Then again, maybe std::swap uses assignment which uses std::swap ... :thinking: 

In any case, if it works now and `std::swap` works now as well, we could also just put move assignment of the member into the assignment operator and get rid of the swap member, or not? I really don't like having a swap member, it looks weird :innocent:

eseiler: > Hm, that sounds weird. Then again, maybe std::swap uses assignment which uses std::swap ... 🤔
> 

It kinda looked like it in the debugger.
> In any case, if it works now and `std::swap` works now as well, we could also just put move assignment of the members (no swap required) into the assignment operator and get rid of the swap member, or not? I really don't like having a swap member, it looks weird 😇
Ok, done!

PR #1149, opened 2019-07-01
[FIX] Adds possible gz extension detection when detecting bgzf stream.

### LINKED ISSUE #1129: Reading from fastq.gz crashes
Hi SeqAn-team :)

I'm trying to open a large fastq.gz file with reads [1] and access the first sequence. It works with SeqAn2 but crashes with SeqAn3. It works however on the unzipped fastq file. So there might be something wrong with the decompression. As @smehringer pointed out both library versions use the same wrapper and zlib? Unfortunately I couldn't create a smaller example that failed.

SeqAn2 (works):

```cpp
CharString id;
Dna5String seq;
SeqFileIn seqFileIn(toCString(path));
readRecord(id, seq, seqFileIn);
cout << seq << '\n';
```

SeqAn3 (crashes):

```cpp
sequence_file_input fin{path};
debug_stream << get<field::SEQ>(fin.front()) << '\n';
```

```
terminate called after throwing an instance of 'seqan3::unhandled_extension_error'
  what():  No valid format found for this extension.
```

I'm working on both master branches of SeqAn2 and SeqAn3 and use ZLIB-1.2.11.

[1] https://ftp-trace.ncbi.nlm.nih.gov/giab/ftp/data/AshkenazimTrio/HG002_NA24385_son/NIST_Illumina_2x250bps/reads/D1_S1_L001_R2_001.fastq.gz

rrahn:
fixes #1129

### DIFF
--- a/include/seqan3/io/detail/misc_input.hpp
+++ b/include/seqan3/io/detail/misc_input.hpp
@@ -105,14 +105,14 @@ inline auto make_secondary_istream(std::basic_istream<char_t> & primary_stream, 
     if (read_chars == magic_number.size() && contrib::_bgzfCheckHeader(magic_number.data())) // BGZF
     {
     #ifdef SEQAN3_HAS_ZLIB
-        if ((extension == ".bgzf"))
+        if ((extension == ".gz") || (extension == ".bgzf"))
             filename.replace_extension();
 
         return {new contrib::basic_bgzf_istream<char_t>{primary_stream},
                 stream_deleter_default};
-        #else
-            throw file_open_error{"Trying to read from a bgzf file, but no ZLIB available."};
-        #endif
+    #else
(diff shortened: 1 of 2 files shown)


include/seqan3/io/detail/misc_input.hpp
```diff
@@ -105,7 +105,7 @@ inline auto make_secondary_istream(std::basic_istream<char_t> & primary_stream,
     if (read_chars == magic_number.size() && contrib::_bgzfCheckHeader(magic_number.data())) // BGZF
     {
     #ifdef SEQAN3_HAS_ZLIB
```
joergi-w: Can you please indent the corresponding `#else` and `#endif` statements below to the same depth as this line?

rrahn: Done

PR #1151, opened 2019-07-02
[FIX] include all constraints in CPO definitions

h-2:
For usability reasons I chose to implement some contraints in the CPO definitions as static_asserts. This has turned out to be a mistake, because it means that checking `seqan3::Alphabet` will result in a hard error if one of the constraints are satisfied, but the static_assert fails.

I have also simplified some of the documentation.

@joergi-w Once this is merged, could you perform similar changes for quality and structure?

### DIFF
--- a/include/seqan3/alphabet/concept.hpp
+++ b/include/seqan3/alphabet/concept.hpp
@@ -44,15 +44,15 @@ public:
     //!\brief Operator definition.
     template <typename alph_t>
     //!\cond
-        requires requires (alph_t const a) { { impl(priority_tag<2>{}, a) }; }
+        requires requires (alph_t const a)
+        {
+            { impl(priority_tag<2>{}, a) };
+            requires noexcept(impl(priority_tag<2>{}, a));
+            requires std::Integral<decltype(impl(priority_tag<2>{}, a))>;
+        }
     //!\endcond
     constexpr auto operator()(alph_t const a) const noexcept
     {
-        static_assert(noexcept(impl(priority_tag<2>{}, a)),
-            "Only overloads that are marked noexcept are picked up by seqan3::to_rank().");
-        static_assert(std::Constructible<size_t, decltype(impl(priority_tag<2>{}, a))>,
-            "The return type of your to_rank() implementation must be convertible to size_t.");
-
         return impl(priority_tag<2>{}, a);
     }
 };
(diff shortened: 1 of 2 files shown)


include/seqan3/alphabet/concept.hpp
```diff
@@ -259,14 +246,12 @@ namespace seqan3
...
- *   2. A free function `to_char(your_type const a)` in
- *      `namespace seqan3::custom`. The same restrictions apply as above.
+ *   1. A free function `to_char(your_type const a)` in the namespace of your type (or as `friend`).
+ *   2. A free function `to_char(your_type const a)` in `namespace seqan3::custom`.
  *   3. A member function called `to_char()`.
- *      It must be marked `noexcept` (`constexpr` is not required, but recommended) and the return type be
- *      of the respective char representation.
+ *
+ * Functions are only considered for one of the above cases if they are marked `noexcept` (`constexpr` is not required,
+ * but recommended).
```
joergi-w: No restriction for the return type here?

include/seqan3/alphabet/concept.hpp
```diff
@@ -322,27 +307,15 @@ struct assign_char_to_fn
     template <typename alph_t>
```
joergi-w: documentation: Operator definition is not anymore for lvalues only.

include/seqan3/alphabet/concept.hpp
```diff
@@ -643,14 +619,15 @@ namespace seqan3
...
+ *      returns the size.
  *   2. A free function `alphabet_size(your_type const &)` in `namespace seqan3::custom` that returns
- *      the size as an integral value. The same restrictions apply as above.
- *   3. A `static constexpr` data member called `alphabet_size` that is the size. It must
- *      be implicitly convertible to `size_t`.
+ *      the size.
+ *   3. A `static constexpr` data member called `alphabet_size` that is the size.
+ *
+ * Functions are only considered for one of the above cases if they are marked `noexcept` **and** `constexpr` and
+ * and if the returned type models std::Integral. For 1. and 2. the value of the argument to the function shall be
```
joergi-w: remove duplicate `and`

include/seqan3/alphabet/concept.hpp
h-2: No, currently not. We are still discussing whether we require convertibility to `char`...

joergi-w: I vote for doing it! But you can discuss/implement that later – I'm approving for now.

include/seqan3/alphabet/concept.hpp
```diff
@@ -130,31 +128,19 @@ struct assign_rank_to_fn
...
-        return impl(priority_tag<2>{}, a, r);
-    }
-
-    //!\brief Operator definition for rvalues.
+    //!\brief Operator definition.
     template <typename alph_t>
     //!\cond
         requires requires (seqan3::alphabet_rank_t<alph_t> const r, alph_t & a)
-            { { impl(priority_tag<2>{}, a, r) }; } && (!std::is_lvalue_reference_v<alph_t>)
+            {
```
rrahn: why was this before separated into lvalue and rvalues? Did something change semantically?

h-2: I was too stupid to realise before that one overload would be enough :man_shrugging:

rrahn: ok

PR #1153, opened 2019-07-03
[FEATURE] dynamic_bitset

eseiler:
### Immediate TODO:
- [X] resize needs to set bits to 0 if shrinking
- [x] Check side effects of above change
- [X] rough docs
- [x] refine docs
- [x] snippets
- [x] Take a guess which functions are not called, because codecov isn't marking them
- [x] increase coverage
- [x] replace ctor + resize with string ctor in tests
- [x] `to_string()` - with custom `debug_stream` overload if this does not work automatically when `to_string()` is defined. Standard range debug_stream'ing needs to be reversed.
---
### Maybe TODO:
Lavish features such as:
- [X] `std::string to_string(CharT zero = CharT('0'), CharT one = CharT('1')) const;` which allows you to set specific characters for 0 and 1
- [X] `unsigned long to_ulong() const` and `unsigned long long to_ullong() const`, but these are one-liners
- [x] `std::hash` support
---
### Note:
1️⃣ `begin()` points to the least significant bit, i.e. right to left instead left to right.\
Example:
```cpp
seqan3::dynamic_bitset t1{0b1100'0001};
t1[0]; // 1
t1[1]; // 0
...
t1[6]; // 1
t1[7]; // 1

for (auto it = t1.begin(); it != it.end(); ++it)
    *it; // 10000011
```
---
### Future TODO:
- extend for arbitrary size
- For `operator>>` we have the line:
```cpp
  if (is.eof())
      break;
```
I copied from the `small_string` implementation.
It's used correctly in this context (we check `is.eof()` *after* we read EOF), but we only read up to `max(is.width(), arg.max_size())` many characters, which already ensures we never read EOF, so the branch is never reached. @smehringer


include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,760 @@
...
+    //!\copydoc end()
+    constexpr const_iterator cend() const noexcept
+    {
+        return const_iterator{*this, size()};
+    }
+    //!\}
+
+    constexpr bool all() const noexcept
+    {
+        return !size() == 0 && count() == size();
```
marehr: `!size() == 0` ? this seems wierd

marehr:
Just curious, why is 58 the upper limit?

eseiler:
> Just curious, why is 58 the upper limit?

We use the most significant 6 bit of a `uint64_t` to encode the size.

include/seqan3/range/container/dynamic_bitset.hpp
eseiler: `all()` has to return `false` for an empty `std::bitset`

marehr: why? normally it returns the neutral element and this is true for empty

eseiler: https://en.cppreference.com/w/cpp/utility/bitset/all_any_none

🤷‍♂

eseiler: Even though `std::ranges::all_of` and `std::all_of` return `true` for empty ranges.

marehr: > true if all bits are set to true, otherwise false

This is now interpretation, but I would have said that all (0-many) bits are set true :joy:

eseiler: Would be more consistent and intuitive, and easier to implement :)

marehr: https://godbolt.org/z/8IqQSj

Says that empty many bits are true for all.

marehr:
A normal bitset can also be greater than 64bit, this one can't. I imagine you don't need this for your implementation.

include/seqan3/range/container/dynamic_bitset.hpp
eseiler: You're right. It has to be `std::bitset<0>` for the std one 😅

eseiler:
> A normal bitset can also be greater than 64bit, this one can't. I imagine you don't need this for your implementation.

Not for now, main application will be the shapes, where 32 is fine for now.
But future plan is to then extend it for > 64

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1479 @@
...
+        uint64_t value{};
+        for (size_t i = 0; i != N - 1; ++i)
+        {
+            if (_lit[i] == '0')
+            {
+                value <<= 1;
+            }
+            else if (_lit[i] == '1')
+            {
+                if (i != 0)
```
marehr: is this really needed?

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1479 @@
...
+     * In \f$O(count)\f$.
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    constexpr void assign(size_type const count, value_type const value) noexcept
+    {
+        clear();
+        auto tmp = view::repeat_n(value, count);
```
marehr: I personally would prefer to not use `ranges` when possible, but if it makes it more easy keep it.

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1479 @@
...
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    template <std::ranges::InputRange other_range_t>
+    //!\cond
+        requires std::Constructible<value_type, /*ranges::range_reference_t*/reference_t<other_range_t>>
+    //!\endcond
+    constexpr void assign(other_range_t && range) noexcept
```
marehr: @h-2 You said that containers will not get a range constructor in the standard, so we also don't need it, right?

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1479 @@
...
+    //!\}
+
+    /*!\name Bit manipulation
+     * \{
+     */
+    //!\brief Sets the bits to the result of binary AND on corresponding pairs of bits of *this and rhs.
+    constexpr dynamic_bitset & operator&=(dynamic_bitset const & rhs) noexcept
+    {
+        assert(size() == rhs.size());
+        data_ &= rhs.data_;
```
marehr: ah okay both have the same capacity AND same size, so the first bits (for the size of bits) stay the same.

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1479 @@
...
+class dynamic_bitset
+{
+private:
+    //!\brief Stores the actual data.
+    uint64_t data_{};
+
+    //!\brief The size in bits of the underlying data structure. 64 for `uint64_t`.
+    static constexpr uint8_t internal_bits{detail::sizeof_bits<decltype(data_)>};
+    //!\brief The number of bits reserved for representing size information in data_. 6 for `uint64_t`.
+    static constexpr uint8_t size_bits{detail::most_significant_bit_set(detail::next_power_of_two(internal_bits))};
```
marehr: If you want to encode the `size` and `value` into one integer, I think it would be better to use [bitfields](https://en.cppreference.com/w/cpp/language/bit_field), because the compiler will do all the hard work for you and it is a standard feature!

https://godbolt.org/z/aNcMKl

eseiler: I didn't know about it, but I think it's preferable.

include/seqan3/range/container/dynamic_bitset.hpp
eseiler: Mhm...maybe I can restructure it, but if I'm at pos 0 and I read 1, I can just do |= 1.
If I'm not at pos 0, I first need to shift it.

include/seqan3/range/container/dynamic_bitset.hpp
eseiler: If so, I'd rather do another PR where this is changed in all the containers we provide.

include/seqan3/range/container/dynamic_bitset.hpp
marehr: but is the value at pos 0 before the shift not 0 anyway?

include/seqan3/range/container/dynamic_bitset.hpp
marehr: kay

include/seqan3/range/container/dynamic_bitset.hpp
eseiler: Makes it easier since I can put the actual logic into `insert`

include/seqan3/range/container/dynamic_bitset.hpp
eseiler: Hah, indeed 😅

include/seqan3/range/container/dynamic_bitset.hpp
eseiler: I also added a card, so we don't forget.

test/snippet/range/container/dynamic_bitset_subscript.cpp
```diff
@@ -0,0 +1,13 @@
+#include <seqan3/core/debug_stream.hpp>
+#include <seqan3/range/container/dynamic_bitset.hpp>
+
+int main()
+{
+    seqan3::dynamic_bitset t1{0b1011'1000'1111};
+
+    // Positions are 0-based and start at the rightmost bit.
+    for (size_t i = 0; i < t1.size(); ++i)
+        seqan3::debug_stream << t1[i]; // 111100011101
```
eseiler:
```suggestion
        seqan3::debug_stream << t1[i]; // 1111'0001'1101
```

test/snippet/range/container/dynamic_bitset_begin.cpp
```diff
@@ -0,0 +1,13 @@
+#include <seqan3/core/debug_stream.hpp>
+#include <seqan3/range/container/dynamic_bitset.hpp>
+
+int main()
+{
+    seqan3::dynamic_bitset t1{0b1011'1000'1111};
+
+    // begin() refers to the rightmost position.
+    for (auto it = t1.begin(); it != t1.end(); ++it)
+        seqan3::debug_stream << *it; // 111100011101
```
eseiler:
```suggestion
        seqan3::debug_stream << *it; // 1111'0001'1101
```

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1837 @@
...
+     *
+     * Constant.
+     */
+    constexpr dynamic_bitset & operator<<=(size_t const count) noexcept
+    {
+        assert(count > 0);
+        assert(count < size());
+        uint8_t offset = 58 - size() + count;
+        data_.bits <<= offset;
+        data_.bits >>= offset - count;
```
marehr: `(data_.bits << size)& (1 << size) - 1` would something like this work?

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1837 @@
...
+     * ### Complexity
+     *
+     * Constant.
+     */
+    constexpr dynamic_bitset operator<<(size_t const count) const noexcept
+    {
+        assert(count > 0);
+        assert(count < size());
+        dynamic_bitset tmp{*this};
+        tmp <<= count;
```
marehr: overflow check?

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1837 @@
...
+     *
+     * Not thread-safe.
+     *
+     * ### Complexity
+     *
+     * Constant.
+     */
+    constexpr dynamic_bitset & set() noexcept
+    {
+        data_.bits = ~0ULL >> 6;
```
marehr: depending on the capacity?

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1837 @@
...
+     * Not thread-safe.
+     *
+     * ### Complexity
+     *
+     * Constant.
+     */
+    constexpr dynamic_bitset & flip() noexcept
+    {
+        uint8_t offset = 58 - size();
+        data_.bits = ~data_.bits << offset >> offset;
```
marehr: :) `and size?`

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1837 @@
...
+    {
+        assert(i <= size());
+        return {data_, i};
+    }
+
+    //!\copydoc operator[]()
+    constexpr const_reference operator[](size_t i) const
+    {
+        assert(i <= size());
+        return data_.bits << (63u - i) >> 63u;
```
marehr: maybe here also

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1837 @@
...
+     * No-throw guarantee.
+     */
+    constexpr void resize(size_type const count, value_type const value = false) noexcept
+    {
+        assert(count <= capacity_);
+        // In case of enlarging, we need the original size after setting the size bits.
+        size_t tmp_size{size()};
+        // Shrinking.
+        for (size_t i = tmp_size; i > count; --i)
+            (*this)[i - 1] = false;
```
marehr: maybe you can use bit masks here which only take O(1) time

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1837 @@
...
+        // Set size bits.
+        data_.size = count;
+        // Enlarging, only needed when we have to set bits.
+        for (size_t i = tmp_size; value && i < count; ++i)
+            (*this)[i] = value;
```
marehr: same here ` | 1 << something`

include/seqan3/range/container/dynamic_bitset.hpp
eseiler: Mhm, do we want an assert or something?
Behaviour is:
If you do `<<=` greater than capacity, then the data is 0, but the size bits are not affected, i.e. same effect as reset.
If you shift more than 63, you get the standard complaint:
```
dynamic_bitset.hpp:680:20: error: right operand of shift expression '(4288 << 64)' is >= than the precision of the left operand [-fpermissive]
  680 |         data_.bits <<= count;
```

include/seqan3/range/container/dynamic_bitset.hpp
eseiler: Good catch!
I think I'll refactor the test such that I test default capacity and a smaller capacity.

include/seqan3/range/container/dynamic_bitset.hpp
eseiler: But, we do assert in debug mode:
```cpp
        assert(count > 0);
        assert(count < size());
```

eseiler:
@marehr ping

smehringer:
maybe clean the commit history?

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1832 @@
...
+     *
+     * No-throw guarantee if value_type is std::is_nothrow_copy_constructible.
+     */
+    friend constexpr void swap(dynamic_bitset & lhs, dynamic_bitset & rhs) noexcept
+    {
+        lhs.swap(rhs);
+    }
+
+    //!\overload
+    friend constexpr void swap(dynamic_bitset && lhs, dynamic_bitset && rhs) noexcept
```
rrahn: You don't need this interface. It doesn't make sense.

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1832 @@
...
+     */
+    constexpr void swap(dynamic_bitset & rhs) noexcept
+    {
+        bitfield tmp = std::move(data_);
+        data_ = std::move(rhs.data_);
+        rhs.data_ = std::move(tmp);
+    }
+
+    //!\overload
+    constexpr void swap(dynamic_bitset && rhs) noexcept
```
rrahn: don't need this

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1832 @@
...
+     *
+     * ### Complexity
+     *
+     * Linear in the size of both containers.
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee if value_type is std::is_nothrow_copy_constructible.
+     */
+    friend constexpr void swap(dynamic_bitset & lhs, dynamic_bitset & rhs) noexcept
```
rrahn: why exposing it as global function?

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1832 @@
...
+     *
+     * Linear in the size of both containers.
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee if value_type is std::is_nothrow_copy_constructible.
+     */
+    constexpr void swap(dynamic_bitset & rhs) noexcept
+    {
+        bitfield tmp = std::move(data_);
```
rrahn:
```suggestion
        std::swap(*this, rhs);
```

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1832 @@
...
+    /*!\name Conversions
+     * \{
+     */
+    /*!\brief Converts the dynamic_bitset to a std::string.
+     * \tparam char_t Character type; must model seqan3::Char.
+     * \param[in] zero Character of type char_t representing `false`. Default '0'.
+     * \param[in] one Character of type char_t representing `true`. Default '1'.
+     * \throws std::bad_alloc from the the std::string constructor.
+     * \returns A std::string representing the dynamic_bitset.
+     *
```
rrahn: please add the \details to all documentation blocks

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1832 @@
...
+     * Throws std::bad_alloc from the std::string constructor.
+     */
+    template <typename char_t = char>
+    std::string to_string(char_t zero = char_t{'0'}, char_t one = char_t{'1'}) const
+    {
+        std::string str{};
+        str.reserve(size());
+        for (bool const bit : std::view::reverse(*this))
+            bit ? str.push_back(one) : str.push_back(zero);
+        return str;
```
rrahn: new line

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1832 @@
...
+    using const_iterator  = detail::random_access_iterator<dynamic_bitset const>;
+    //!\brief A std::ptrdiff_t.
+    using difference_type = ptrdiff_t;
+    //!\brief An unsigned integer type (usually std::size_t)
+    using size_type       = detail::min_viable_uint_t<capacity_>;
+    //!\}
+
+    //!\cond
+    // this signals to range-v3 that something is a container :|
+    using allocator_type    = void;
```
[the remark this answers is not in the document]
eseiler: > you want this to be not a view, right? So the correct way per standard is to overload

I want this to be a container recognised by the range library. We added the `allocator_type` for each of our container; otherwise the container_concept of the range library fails. 
I think it doesn't have to do something with views?

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1832 @@
...
+     *
+     * Delegates to the seqan3::dynamic_bitset(uint64_t value) constructor.
+     *
+     * ### Complexity
+     *
+     * Constant.
+     *
+     * ### Exceptions
+     *
+     * Throws std::invalid_argument if uint64_t representation of value has a set bit past the 58 one,
```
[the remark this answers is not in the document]
eseiler: Well, the 58'th element has index 57. But I agree that it could be phrased better.

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1832 @@
...
+     * i.e. only bits in [0,58) may be set.
+     */
+    template <std::Integral t>
+    //!\cond
+        requires detail::sizeof_bits<t> <= 64
+    //!\endcond
+    constexpr dynamic_bitset(t value) : dynamic_bitset(static_cast<uint64_t>(value))
```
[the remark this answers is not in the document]
eseiler: You're right, this one is not necessary anymore. I had previously the case that passing an int wouldn't work; but apparently it works now.

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1832 @@
...
+                value <<= 1;
+                value |= 1u;
+            }
+            else
+            {
+                throw std::invalid_argument{"The string to construct a dynamic_bitset from may only contain 0 and 1."};
+            }
+        }
+
+        (*this) = value;
```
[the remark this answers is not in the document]
eseiler: (I don't know)

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1832 @@
...
+     *
+     * ### Complexity
+     *
+     * Linear in the size of `ilist`.
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    constexpr dynamic_bitset & operator=(std::initializer_list<value_type> ilist) noexcept
```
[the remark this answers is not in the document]
eseiler: The might of the ctor with a `initializer_list` overwhelms all other constructors regardless of constraints. Hence it is always chosen, but never works.
We also don't have this as ctor in other containers because of this reason. But maybe there is a way?

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1832 @@
...
+        return const_iterator{*this};
+    }
+
+    //!\copydoc begin()
+    constexpr const_iterator cbegin() const noexcept
+    {
+        return const_iterator{*this};
+    }
+
+    //!\brief Returns iterator past the end of the dynamic_bitset.
```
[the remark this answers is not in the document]
eseiler: I'll just add the begin snippet.

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1832 @@
...
+     *
+     * \include test/snippet/range/container/dynamic_bitset_bitwise_and_member.cpp
+     *
+     * ### Exception
+     *
+     * No-throw guarantee.
+     *
+     * ### Thread-safety
+     *
+     * Not thread-safe.
```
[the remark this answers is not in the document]
eseiler: But the individual doc should stay?

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1832 @@
...
+    constexpr const_reference back() const noexcept
+    {
+        assert(size() > 0);
+        return (*this)[size()-1];
+    }
+
+    //!\brief Direct access to the underlying bit field.
+    constexpr bitfield * data() noexcept
+    {
+        return data_;
```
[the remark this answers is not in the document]
eseiler: `binary_data` ?

include/seqan3/range/container/dynamic_bitset.hpp
eseiler: std::swap is not constexpr.

So I need to provide something.

include/seqan3/range/container/dynamic_bitset.hpp
eseiler: std::swap is not constexpr

include/seqan3/range/container/dynamic_bitset.hpp
eseiler: We have this in all the containers, so why not here?

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1897 @@
...
+#include <seqan3/range/view/repeat_n.hpp>
+
+namespace seqan3
+{
+
+/*!\brief A constexpr bitset implementation with dynamic size at compile time.
+ * \implements seqan3::ReservableContainer
+ * \implements seqan3::Cerealisable
+ * \ingroup container
+ * \tparam capacity_   The capacity of the dynamic bitset
```
rrahn:
```suggestion
 * \tparam capacity_ The capacity of the dynamic bitset
```

rrahn: Please do not name it `capacity_`. Call it `bitset_capacity`, maybe?

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1897 @@
...
+ * capacity is fixed.
+ *
+ * ### Example
+ *
+ * \include test/snippet/range/container/dynamic_bitset_usage.cpp
+ *
+ * ### Thread safety
+ *
+ * This container provides no thread-safety beyond the promise given also by the STL that all
+ * calls to `const` member function are safe from multiple threads (as long as no thread calls
```
rrahn:
```suggestion
 * calls to `const` member functions are safe from multiple threads (as long as no thread calls
```

include/seqan3/range/container/dynamic_bitset.hpp
eseiler: Mhm `small_string` and `small_vector` also use `capacity_`.
I changed it to `bit_capacity`.

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1897 @@
...
+        {
+            return static_cast<bool>(internal.bits & mask);
+        }
+
+        //!\brief Returns the inverted value of the referenced bit.
+        constexpr bool operator~() const noexcept
+        {
+            return !static_cast<bool>(internal.bits & mask);
+        }
+
```
rrahn: double newline

include/seqan3/range/container/dynamic_bitset.hpp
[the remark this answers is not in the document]
rrahn: As far as I can tell, the container concept of the ranges library does not check for a type `allocator_type`. Can you give me an example where the expected behavior does not work without this type?

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1832 @@
...
+    //!\endcond
+    constexpr dynamic_bitset(begin_it_type begin_it, end_it_type end_it) noexcept:
+        dynamic_bitset{}
+    {
+        assign(begin_it, end_it);
+    }
+
+    /*!\brief Construct from a different range.
+     * \tparam other_range_t The type of range to be inserted; must satisfy std::ranges::InputRange and `value_type`
+     *                       must be constructible from reference_t<other_range_t>.
```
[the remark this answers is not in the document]
rrahn: still open

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1897 @@
...
+     * \param[in]      range The sequence to construct/assign from.
```
rrahn: I mean, leave it as is right now but for the future please do not align different doxygen entities. That looks just awful.

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1897 @@
...
+        requires std::Constructible<value_type, reference_t<other_range_t>>
+    //!\endcond
+    constexpr void assign(other_range_t && range) noexcept
+    {
+        assign(std::ranges::begin(range), std::ranges::end(range));
+    }
+
+    /*!\brief Assign from pair of iterators.
+     * \tparam begin_it_type Must satisfy std::ForwardIterator and the `value_type` must be constructible from
+     *                       the reference type of begin_it_type.
```
rrahn: can you please check all comments again, i.e. that code expressions are wrapped in a `code`-block. So that it is consistent through the documentation

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -0,0 +1,1897 @@
...
+        assert(size() < bit_capacity);
+        resize(size() + 1);
+        (*this)[size() - 1] = value;
+    }
+
+    /*!\brief Removes the last element of the container.
+     *
+    * \details
+     *
+     * Calling pop_back() on an empty container is undefined. In debug mode an assertion will be thrown.
```
rrahn:
```suggestion
     * Calling `pop_back()` on an empty container is undefined. In debug mode an assertion will be thrown.
```

include/seqan3/range/container/dynamic_bitset.hpp
rrahn: what are you saving? The constructor and the assignment are defaulted anyway. Non one us using the swap internally. the copy-n-swap idiom is not necessary here. Hence, std::swap would work already for this class and should be the preferred way. 
Mhmm but I see know, that it is the stupid container requirement. I'll make this a discussion point in the next dev meeting

include/seqan3/range/container/dynamic_bitset.hpp
rrahn: ok

include/seqan3/range/container/dynamic_bitset.hpp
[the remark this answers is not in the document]
rrahn: no you don't need to mark them not-thread safe as it is clear by the top level documentation. In general if there is no explicit statement you always assume not thread-safe

include/seqan3/range/container/dynamic_bitset.hpp
eseiler: Anyway, std::swap is not constexpr...

include/seqan3/range/container/dynamic_bitset.hpp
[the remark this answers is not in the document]
eseiler: It's used in `range/v3/range/conversion.hpp` as a hack for converting ranges to containers or something?

Idk about an example, I'd rather do a card for the dev meeting since this affects all the containers...

eseiler:
@rrahn all the doc issues should be resolved now

PR #1154, opened 2019-07-04
[DOC] Add full solution at the end of the argument parser tutorial.

### LINKED ISSUE #1143: [DOC] Add complete example of argument parser into tutorial
As discussed with @smehringer it would be more convenient for the user if there was a complete code example of the entire argument parser example at the end of the tutorial.

At the moment you only find the intermediate solutions in nested solution boxes that you need to expand. I would suggest to add a section with the entire code at the end of the tutorial and include it into the TOC.

smehringer:
Resolves #1143

### DIFF
--- a/doc/tutorial/argument_parser/index.md
+++ b/doc/tutorial/argument_parser/index.md
@@ -457,6 +457,15 @@ Add a seqan3::regex_validator to the first positional option that expects the `f
 \snippet doc/tutorial/argument_parser/solution6.cpp file_validator
 \endsolution
 
+# Full solution
+
+The following solution shows the complete code including all the little assignments of this tutorial
+that can serve as a copy'n'paste source for your application.
+
+\solution
+\include doc/tutorial/argument_parser/solution6.cpp
+\endsolution
+
 # Update Notifications
 
 When you run a SeqAn-based application for the first time, you will likely be asked about "update notifications".


doc/tutorial/argument_parser/index.md
```diff
@@ -457,6 +457,15 @@ Add a seqan3::regex_validator to the first positional option that expects the `f
 \snippet doc/tutorial/argument_parser/solution6.cpp file_validator
 \endsolution
 
+# Full solution
+
+Here is the complete code including all the little assignments of this tutorial
+that can serve as a copy'n'paste source for your application.
+
+\htmlonly <div class=\"assignment\"> <details><summary>Click me to show the code</summary> \endhtmlonly
```
rrahn: Why not using `solution` block?

smehringer: Because it is not an actual solution to a specific assignment?

rrahn: But ist the solution block not usable as a standalone block. I mean it is also a solution. Just the full solution. And i thought it was collapsable as well?

smehringer: Yes it is collapsable but it displays
```
> Solution
````
Instead of customizable: "click here to see full code"

rrahn: yeah, but you an just write above `The following soltion shows the complete block`. I would rather not put such a html sutff into md file. I also think that it would be possible to improve the doxygen macro to allow custom messages. Anyway, just using the solution block and adding a sentence like above should suffice for the moment.

smehringer: done

PR #1156, opened 2019-07-04
Restructure stream handles from files to formats.

smehringer:
Blocked by ~#1160~ 
Blocked by ~#1184~ 

Moves the stream handle from the file to the format
and for alignment SAM removes all usage of `operator<<()` in favour of working directly on the ostreambuf_iterator.

### DIFF
--- a/include/seqan3/io/alignment_file/format_bam.hpp
+++ b/include/seqan3/io/alignment_file/format_bam.hpp
@@ -18,6 +18,8 @@
 
 #include <seqan3/alphabet/detail/convert.hpp>
 #include <seqan3/alphabet/nucleotide/sam_dna16.hpp>
+#include <seqan3/contrib/stream/bgzf_istream.hpp>
+#include <seqan3/contrib/stream/bgzf_ostream.hpp>
 #include <seqan3/core/bit_manipulation.hpp>
 #include <seqan3/core/char_operations/predicate.hpp>
 #include <seqan3/core/concept/core_language.hpp>
@@ -37,6 +39,7 @@
 #include <seqan3/io/detail/ignore_output_iterator.hpp>
 #include <seqan3/io/detail/misc.hpp>
 #include <seqan3/range/detail/misc.hpp>
+#include <seqan3/range/view/istreambuf.hpp>
 #include <seqan3/range/view/slice.hpp>
 #include <seqan3/range/view/take_exactly.hpp>
 #include <seqan3/range/view/take_until.hpp>
@@ -87,11 +90,18 @@ struct alignment_record_core
     int32_t tlen;           //!< The template length of the read and its mate.
 };
 
-//!\brief The seqan3::alignment_file_input_format specialisation that handles formatted BAM input.
-//!\ingroup alignment_file
-template <>
-class alignment_file_input_format<format_bam> : alignment_file_input_format<format_sam>
+/*!\brief The seqan3::alignment_file_input_format specialisation that handles formatted BAM input.
+ * \ingroup alignment_file
+ * \tparam stream_char_type The underlying character type of the stream (usually `char`).
+ */
+template <typename stream_char_type>
+class alignment_file_input_format<format_bam, stream_char_type> :
+    alignment_file_input_format<format_sam, stream_char_type>
 {
+private:
+    //!\brief The SAM format type that this class inherits from.
+    using sam_fmt = alignment_file_input_format<format_sam, stream_char_type>;
+
 public:
     //!\brief Exposes the format tag that this class is specialised with
     using format_tag = format_bam;
@@ -99,7 +109,7 @@ public:
     /*!\name Constructors, destructor and assignment
      * \{
      */
-    alignment_file_input_format()                                                noexcept = default; //!< Defaulted.
+    alignment_file_input_format()                                                         = default; //!< Defaulted.
     //!\brief Copy construction is explicitly deleted, because you can't have multiple access to the same file.
     alignment_file_input_format(alignment_file_input_format const &)                      = delete;
     //!\brief Copy assignment is explicitly deleted, because you can't have multiple access to the same file.
@@ -107,11 +117,14 @@ public:
     alignment_file_input_format(alignment_file_input_format &&)                  noexcept = default; //!< Defaulted.
     alignment_file_input_format & operator=(alignment_file_input_format &&)      noexcept = default; //!< Defaulted.
     ~alignment_file_input_format()                                               noexcept = default; //!< Defaulted.
+
+    alignment_file_input_format(std::basic_istream<stream_char_type> & stream) :
+        stream_ptr{new contrib::basic_bgzf_istream<stream_char_type>{stream}}
+    {}
     //!\}
 
     //!\copydoc AlignmentFileInputFormat::read
-    template <typename stream_type,     // constraints checked by file
-              typename seq_legal_alph_type,
+    template <typename seq_legal_alph_type,
               typename ref_seqs_type,
               typename ref_ids_type,
               typename seq_type,
@@ -128,8 +141,7 @@ public:
               typename tag_dict_type,
               typename e_value_type,
               typename bit_score_type>
-    void read(stream_type                                             & stream,
-              alignment_file_input_options<seq_legal_alph_type> const & SEQAN3_DOXYGEN_ONLY(options),
+    bool read(alignment_file_input_options<seq_legal_alph_type> const & SEQAN3_DOXYGEN_ONLY(options),
               ref_seqs_type                                           & ref_seqs,
               alignment_file_header<ref_ids_type>                     & header,
               seq_type                                                & seq,
@@ -157,9 +169,11 @@ public:
         static_assert(detail::decays_to_ignore_v<flag_type> || std::Same<flag_type, uint16_t>,
                       "The type of field::FLAG must be uint8_t.");
 
-        using stream_buf_t = std::istreambuf_iterator<typename stream_type::char_type>;
-        auto stream_view = std::ranges::subrange<decltype(stream_buf_t{stream}), decltype(stream_buf_t{})>
-                               {stream_buf_t{stream}, stream_buf_t{}};
+        assert(stream_ptr != nullptr);
+        auto stream_view = view::istreambuf(*stream_ptr);
+
+        if (std::ranges::begin(stream_view) == std::ranges::end(stream_view)) // no records follow
+            return true;
 
         // these variables need to be stored to compute the ALIGNMENT
         [[maybe_unused]] int32_t offset_tmp{};
@@ -179,10 +193,10 @@ public:
             read_field(stream_view, tmp32);
 
             if (tmp32 > 0) // header text is present
-                read_header(stream_view | view::take_exactly_or_throw(tmp32)
-                                        | view::take_until_and_consume(is_char<'\0'>),
-                            header,
-                            ref_seqs);
+                sam_fmt::read_header(stream_view | view::take_exactly_or_throw(tmp32)
+                                                 | view::take_until_and_consume(is_char<'\0'>),
+                                     header,
+                                     ref_seqs);
 
             int32_t n_ref;
             read_field(stream_view, n_ref);
@@ -220,8 +234,8 @@ public:
 
             header_was_read = true;
 
-            if (stream_buf_t{stream} == stream_buf_t{}) // no records follow
-                return;
+            if (std::ranges::begin(stream_view) == std::ranges::end(stream_view)) // no records follow
+                return true;
         }
 
         // read alignment record into buffer
@@ -419,12 +433,16 @@ public:
             }
 
             // Alignment object construction
-            construct_alignment(align, cigar, core.refID, ref_seqs, core.pos, ref_length); // inherited from SAM format
+            sam_fmt::construct_alignment(align, cigar, core.refID, ref_seqs, core.pos, ref_length);
         }
+
+        return false;
     }
 
 protected:
     //!\privatesection
+    //!\brief A pointer to the bgzf uncompressed stream.
+    std::unique_ptr<std::basic_istream<stream_char_type>> stream_ptr{nullptr};
 
     //!\brief A variable that tracks whether the content of header has been read or not.
     bool header_was_read{false};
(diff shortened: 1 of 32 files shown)


include/seqan3/io/detail/misc_input.hpp
```diff
@@ -102,7 +101,7 @@ inline auto make_secondary_istream(std::basic_istream<char_t> & primary_stream,
     std::string const extension = filename.extension().string();
 
     // set return value appropriately
-    if (read_chars == magic_number.size() && contrib::_bgzfCheckHeader(magic_number.data())) // BGZF
+    if (read_chars == magic_number.size() && contrib::_bgzfCheckHeader(magic_number.data()) && extension != ".bam") // BGZF
```
eseiler: length

Mhm, I don't know whats supposedly faster, but I would put the fastest condition at the front.

test/unit/io/alignment_file/alignment_file_format_test_template.hpp
```diff
@@ -264,7 +264,7 @@ TYPED_TEST_P(alignment_file_read, read_in_all_but_empty_data)
     std::tuple<std::optional<int32_t>, std::optional<int32_t>, int32_t> mate;
     sam_tag_dictionary tag_dict;
 
-    ASSERT_NO_THROW(format.read(istream, input_options, this->ref_sequences, this->header, seq, qual, id, offset, std::ignore,
+    ASSERT_NO_THROW(format.read(input_options, this->ref_sequences, this->header, seq, qual, id, offset, std::ignore,
                                 ref_id_in, ref_offset, alignment, flag, mapq, mate, tag_dict, std::ignore, std::ignore));
```
eseiler: The `;` doesn't fit on this line! 😁

test/unit/io/alignment_file/alignment_file_format_test_template.hpp
```diff
@@ -486,34 +487,35 @@ TYPED_TEST_P(alignment_file_write, write_empty_members)
...
-    header.ref_id_info.push_back({this->ref_seq.size(), ""});
-    header.ref_dict[this->ref_id] = 0;
+    {
+        detail::alignment_file_output_format<TypeParam> format{ostream};
 
-    this->tag_dicts[0]["NM"_tag] = 7;
-    this->tag_dicts[0]["AS"_tag] = 2;
-    this->tag_dicts[1]["xy"_tag] = std::vector<uint16_t>{3,4,5};
+        alignment_file_header header{std::vector<std::string>{this->ref_id}};
+        header.ref_id_info.push_back({this->ref_seq.size(), ""});
```
eseiler: `emplace_back` should also work (does it?) and save you move. Also check for the other tests

test/unit/io/alignment_file/alignment_file_output_test.cpp
```diff
@@ -519,20 +519,20 @@ read1	41	ref	1	61	1S1M1D1M1I	ref	10	300	ACGT	!##$	AS:i:2	NM:i:7
...
-
-    EXPECT_EQ(reinterpret_cast<std::ostringstream &>(fout.get_stream()).str(), comp);
+    os.flush();
+    EXPECT_EQ(os.str(), comp);
 }
 
 TEST(rows, write_bam_file)
 {
-    test::tmp_filename const filename{"in_out.bam"};
+    test::tmp_filename const filename{"/tmp/in_out.bam"};
```
eseiler: Don't hardcode the tmp path
Probably a change for debugging?

test/unit/io/alignment_file/format_bam_test.cpp
```diff
@@ -5,6 +5,7 @@
 // shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
 // -----------------------------------------------------------------------------------------------------
 
+#include <fstream> // remove me
```
eseiler: I agree

include/seqan3/std/charconv
```diff
@@ -859,6 +859,23 @@ inline std::to_chars_result to_chars(char * first, char * last, value_type value
     return seqan3::detail::to_chars_itoa(first, last, value, is_signed<value_type>());
 }
 
+//!\brief std::to_chars overload for floating point via a stringstream.
+//!\ingroup charconv
+template <seqan3::FloatingPoint floating_point_type>
+inline std::to_chars_result to_chars(char * first, char * last, floating_point_type value, int = 10) noexcept
```
eseiler: `value` can be const
and what is the `int = 10` for?

test/unit/std/charconv_test.cpp
```diff
@@ -483,3 +504,19 @@ TYPED_TEST(from_char_real_test, non_valid_strings)
...
+TYPED_TEST(from_char_real_test, to_chars)
+{
+    TypeParam val{120.3};
+    char buffer[10];
+
+    auto res = std::to_chars(&buffer[0], &buffer[0] + sizeof(buffer), val);
+
+    EXPECT_EQ(res.ptr, &buffer[5]);
+    EXPECT_EQ(res.ec, std::errc{});
+    EXPECT_EQ((std::string_view{buffer, 5}), std::string_view{"120.3"});
```
eseiler: you don't test the case where the buffer is to small?

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -89,9 +92,12 @@ struct alignment_record_core
 
 //!\brief The seqan3::alignment_file_input_format specialisation that handles formatted BAM input.
 //!\ingroup alignment_file
-template <>
-class alignment_file_input_format<format_bam> : alignment_file_input_format<format_sam>
+template <typename stream_char_type>
```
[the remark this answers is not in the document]
smehringer: it is instantiated in the file object where the Char concept is checked

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -764,6 +779,8 @@ class alignment_file_output_format<format_bam> : alignment_file_output_format<fo
         }
         else
         {
+            auto & stream = *stream_ptr;
```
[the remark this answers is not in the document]
smehringer: this is removed again in the third commit I think?
Yes it is save as long as I only use the iterator. if I use the reference in between the iterator is invalidated.

If you pass a nullptr you cannot read/write. I should probably assert for this.

test/unit/io/alignment_file/alignment_file_output_test.cpp
smehringer: good catch

test/unit/std/charconv_test.cpp
smehringer: well it is the same function so I thought it is enough to test it once :)

include/seqan3/io/alignment_file/format_bam.hpp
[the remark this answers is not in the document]
eseiler: ok!

include/seqan3/io/detail/misc_input.hpp
smehringer: I think the size check at the front is already the fastest? The others compare characters linearly.

If you were aiming at "which condition is most likely false", since this is terminating the other comparisons it is also the size I guess. the extension != ".bam" is the most likely one as this function is also used for other formats.

test/unit/io/alignment_file/alignment_file_format_test_template.hpp
smehringer: it is just the tests. but fine

include/seqan3/std/charconv
smehringer: well, it is a copy anyway but I guess const does not hurt.

The int = 10 is for compatibility of the function signature. to_chars is documented to have a parameter `base`, but since I wrote a dirty, inefficient version of this function I do not use this parameter. To avoid unused parameter warnings it does not have a name, and in order to be callable without the base parameter I default the non-existent parameter... didn't know this works until I tried it ^^

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -87,31 +90,41 @@ struct alignment_record_core
     int32_t tlen;           //!< The template length of the read and its mate.
 };
 
-//!\brief The seqan3::alignment_file_input_format specialisation that handles formatted BAM input.
-//!\ingroup alignment_file
-template <>
-class alignment_file_input_format<format_bam> : alignment_file_input_format<format_sam>
+/*!\brief The seqan3::alignment_file_input_format specialisation that handles formatted BAM input.
+ * \ingroup alignment_file
+ * \tparam stream_char_type The underlying character of the stream (usually `char`).
```
eseiler:
```suggestion
 * \tparam stream_char_type The underlying character type of the stream (usually `char`).
```

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -87,31 +90,41 @@ struct alignment_record_core
...
-    alignment_file_input_format()                                                noexcept = default; //!< Defaulted.
+    alignment_file_input_format()                                                         = default; //!< Defaulted.
     //!\brief Copy construction is explicitly deleted, because you can't have multiple access to the same file.
     alignment_file_input_format(alignment_file_input_format const &)                      = delete;
     //!\brief Copy assignment is explicitly deleted, because you can't have multiple access to the same file.
     alignment_file_input_format & operator=(alignment_file_input_format const &)          = delete;
-    alignment_file_input_format(alignment_file_input_format &&)                  noexcept = default; //!< Defaulted.
-    alignment_file_input_format & operator=(alignment_file_input_format &&)      noexcept = default; //!< Defaulted.
-    ~alignment_file_input_format()                                               noexcept = default; //!< Defaulted.
+    alignment_file_input_format(alignment_file_input_format &&)                           = default; //!< Defaulted.
```
eseiler: Can the `noexcept`s make a comeback?

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -643,31 +660,42 @@ class alignment_file_input_format<format_bam> : alignment_file_input_format<form
...
-    alignment_file_output_format()                                                 noexcept = default; //!< Defaulted.
+    alignment_file_output_format()                                                          = default; //!< Defaulted.
     //!\brief Copy construction is explicitly deleted, because you can't have multiple access to the same file.
     alignment_file_output_format(alignment_file_output_format const &)                      = delete;
     //!\brief Copy assignment is explicitly deleted, because you can't have multiple access to the same file.
     alignment_file_output_format & operator=(alignment_file_output_format const &)          = delete;
-    alignment_file_output_format(alignment_file_output_format &&)                  noexcept = default; //!< Defaulted.
-    alignment_file_output_format & operator=(alignment_file_output_format &&)      noexcept = default; //!< Defaulted.
-    ~alignment_file_output_format()                                                noexcept = default; //!< Defaulted.
+    alignment_file_output_format(alignment_file_output_format &&)                           = default; //!< Defaulted.
```
eseiler: and here?

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -959,10 +987,14 @@ class alignment_file_output_format<format_bam> : alignment_file_output_format<fo
...
 
             // write optional fields
-            stream << tag_dict_binary_str;
+            std::ranges::copy_n(tag_dict_binary_str.begin(), tag_dict_binary_str.size(), stream_it);
         } // if constexpr (!detail::decays_to_ignore_v<header_type>)
     }
 
+private:
+    //!\brief A pointer to the bgzf uncompressed stream.
+    std::unique_ptr<std::basic_ostream<stream_char_type>> stream_ptr{nullptr};
```
eseiler: Can you do an `assert(stream_ptr != nullptr);` at a suitable location?

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -204,7 +208,9 @@ class alignment_file_input_format<format_sam>
                       detail::is_type_specialisation_of_v<ref_offset_type, std::optional>,
                       "The ref_offset must be a specialisation of std::optional.");
 
-        auto stream_view = view::istreambuf(stream);
+        if (std::ranges::begin(stream_view) == std::ranges::end(stream_view)) // file has no records
```
eseiler:
```suggestion
        if (std::ranges::empty(stream_view)) // file has no records
```
Should also work (http://eel.is/c++draft/range.prim.empty). 
But you can leave it as is if you want.

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -98,29 +100,34 @@ class sequence_file_input_format<format_embl>
     /*!\name Constructors, destructor and assignment
      * \{
      */
-    sequence_file_input_format()                                               noexcept = default; //!< Defaulted.
```
eseiler: why not anymore?

smehringer: Because the compiler complains ... I think to istreambuf_view decltype expression has no noexpect default construction. at least thats the only thing that changed.

include/seqan3/io/alignment_file/format_sam.hpp
smehringer: I would like to keep this. begin and end are in the iterator header while empty is in the ranges header. as a follow up we will try to reduce the usage of the ranges header so this would need to be changed back then (if it is possible to be independent)

eseiler:
> > But, theoretically, read and write are public (e.g. here). So there should be some deprecation? zipper_mouth_face
> 
> no they are not! we do not use the formats but the files for reading :)

Yes I know, but my point is that you can find it in the user docs ... i.e. if someone wanted to use it, they could and it would be a public api.
(Not that I want it to be deprecated.)

smehringer:
> Yes I know, but my point is that you can find it in the user docs.

ahh. true. Darn, I am not sure if this should stay public..
@h-2 the read functions of the detail format is public through the public input/output format concept :(

eseiler:
> > Yes I know, but my point is that you can find it in the user docs.
> 
> ahh. true. Darn, I am not sure if this should stay public..
> @h-2 the read functions of the detail format is public through the public input/output format concept :(

We can mark them with `\noapi` or whatever it is called.
I don't think it should be public since we want them to use the files.

h-2:
Remember that we do want people to be able to add their own formats so there need to be public concepts that define the interface of such formats.

smehringer:
But currently you cannot add your own format unless in the seqan3 namespace :see

h-2:
Yes, but we explicitly said before the release that this was something we want to fix before 3.1

smehringer:
Yes, was there a solution already?

h-2:
So we delay this for now?

smehringer:
> So we delay this for now?

Yes.

marehr:
blocked by other restructurings, close for now

PR #1158, opened 2019-07-04
Feature/alignment trace iterator

rrahn:
This is a first PR for adding a traceback iterator and a trace builder.

This PR does:

 * Add a two dimensional matrix with either column or row major order layout
 * Adds a customisable 2d iterator to iterate through the 2d matrix
 * Updates the Matrix concept to model standard container properties more closely.
 * Adds a traceback iterator for affine gaps and also a traceback path as a range based interface to iterate over the traceback path.

### Depends on 
~#1179~


include/seqan3/alignment/matrix/detail/traceback_path.hpp
```diff
@@ -0,0 +1,263 @@
...
+        if (static_cast<bool>(*matrix_iter & trace_directions::diagonal))
+            dir = trace_directions::diagonal;
+        else if (static_cast<bool>(*matrix_iter & trace_directions::up) ||
+                 static_cast<bool>(*matrix_iter & trace_directions::up_open))
+            dir = trace_directions::up;
+        else if (static_cast<bool>(*matrix_iter & trace_directions::left) ||
+                 static_cast<bool>(*matrix_iter & trace_directions::left_open))
+            dir = trace_directions::left;
+        else
+            dir = trace_directions::none;
```
marehr: ?

include/seqan3/alignment/matrix/detail/two_dimensional_matrix_iterator_base.hpp
```diff
@@ -0,0 +1,378 @@
...
+    //!\brief Returns a reference to the pointed-to-element after advancing the iterator by the given offset.
+    constexpr reference operator[](matrix_offset const & offset) const noexcept
+    {
+        return *(*this + offset);
+    }
+
+    //!\brief Returns a pointer to the pointed-to-element.
+    constexpr pointer operator->() const noexcept
+    {
+        &(*host_iter);
```
marehr: ???

include/seqan3/alignment/matrix/detail/traceback_path.hpp
```diff
@@ -0,0 +1,263 @@
...
+ * end of the traceback.
+ * Note since a traceback follows the path from sink to root in the alignment matrix the resulting path is
+ * also in reversed order. Thus the first value in the path corresponds to the last direction where the alignment
+ * ends.
+ */
+template <TwoDimensionalMatrixIterator matrix_iter_t>
+//!\cond
+    requires std::Same<value_type_t<matrix_iter_t>, trace_directions>
+//!\endcond
+class traceback_path : public std::ranges::view_interface<traceback_path<matrix_iter_t>>
```
marehr: do a subrange

include/seqan3/alignment/matrix/detail/traceback_path.hpp
rrahn: Yes this is cur_dir. I've read about issues with private member functions, namely that it is hard to understand from the interface which variables are going to be modified. Accordingly the proposed solution is to mark private member functions const and pass the members that are modified as parameters to the function.

rrahn:
@marehr ping

rrahn:
@marehr ping

include/seqan3/alignment/matrix/detail/trace_iterator.hpp
```diff
@@ -0,0 +1,234 @@
...
+    //!\cond
+        requires std::Constructible<matrix_iter_t, other_matrix_iter_t> ||
+                 std::Constructible<other_matrix_iter_t, matrix_iter_t>
+    //!\endcond
+    constexpr bool operator==(trace_iterator<other_matrix_iter_t> const & rhs) const noexcept
+    {
+        return *matrix_iter == *rhs.matrix_iter;
+    }
+
+    //!\brief Returns `true` if the point-to-element is seqan3::detail::trace_directions::none.
```
marehr: point-to-element? What does that mean?

include/seqan3/alignment/matrix/detail/trace_iterator.hpp
```diff
@@ -0,0 +1,234 @@
...
+        trace_directions old_dir = *matrix_iter;
+
+        assert(old_dir != trace_directions::none);
+
+        if (cur_dir == trace_directions::up)
+        {
+            matrix_iter -= matrix_offset{row_index_type{1}, column_index_type{0}};
+            // Set new traceback direction if last was up_open.
+            if (static_cast<bool>(old_dir & trace_directions::up_open))
+                set_trace_direction(); // Set the current trace_direction
```
marehr:
```suggestion
                set_current_trace_direction(*matrix_iter);
```

include/seqan3/alignment/matrix/detail/two_dimensional_matrix.hpp
```diff
@@ -0,0 +1,376 @@
...
+    class iterator_type :
+        public two_dimensional_matrix_iterator_base<iterator_type<matrix_t>, order>
+    {
+    private:
+        //!\brief The base class type.
+        using base_t = two_dimensional_matrix_iterator_base<iterator_type<matrix_t>, order>;
+
+        //!\brief Befriend the base crtp class.
+        template <typename derived_t, matrix_major_order other_order>
+            requires is_type_specialisation_of_v<derived_t, iterator_type> && other_order == order
```
marehr: I'm not sure, but I think requires in friend declaration is not allowed.

include/seqan3/alignment/matrix/detail/two_dimensional_matrix.hpp
```diff
@@ -0,0 +1,376 @@
...
+    /*!\brief A two-dimensional matrix iterator.
+     * \tparam matrix_t The underlying seqan3::detail::two_dimensional_matrix.
+     *
+     * \details
+     *
+     * Offers a two-dimensional iterator interface over the seqan3::detail::two_dimensional_matrix, which stores the
+     * data in a flattened one-dimensional vector.
+     */
+    template <typename matrix_t>
+    class iterator_type :
```
marehr: out of class?

include/seqan3/alignment/matrix/debug_matrix.hpp
```diff
@@ -465,4 +483,14 @@ inline debug_stream_type<char_t> & operator<<(debug_stream_type<char_t> & s, ali
     return s;
 }
 
+//!\overload
+template <std::ranges::InputRange alignment_matrix_t, typename char_t>
+//!\cond
+    requires detail::DebugStreamableRange<alignment_matrix_t> && detail::Matrix<alignment_matrix_t>
+//!\endcond
+inline debug_stream_type<char_t> & operator<<(debug_stream_type<char_t> & s, alignment_matrix_t && matrix)
```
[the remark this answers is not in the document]
rrahn: Well, it is a long time ago and I thought it was part of another PR. But obviously not. This overload is required since my changes will add a range interface to the matrix. This however, competes with the overload for ranges of our debug stream. Accordingly a matrix, that is a range must refine the corresponding range overload. In the future, we can discuss if not the entire matrix concept should be adapted to refine the range concept as a two-dimensional range.  But I wasn't sure yet if we want this. And b) the required changes are beyond this PR.

include/seqan3/alignment/matrix/detail/trace_iterator.hpp
rrahn: I think current does not help either. So I changed it as asked but left the interface name as is.

include/seqan3/alignment/matrix/detail/trace_iterator.hpp
rrahn: ok sit should be pointer-to-element. And it means what it says 🤓 . The element pointed to by the iterator

include/seqan3/alignment/matrix/detail/trace_iterator.hpp
```diff
@@ -0,0 +1,234 @@
...
+            cur_dir = trace_directions::up;
+        else if (static_cast<bool>(*matrix_iter & trace_directions::left) ||
+                 static_cast<bool>(*matrix_iter & trace_directions::left_open))
+            cur_dir = trace_directions::left;
+        else
+            cur_dir = trace_directions::none;
+    }
+
+    matrix_iter_t matrix_iter{}; //!< The underlying matrix iterator.
+    trace_directions cur_dir{};  //!< The current tracenback direction.
```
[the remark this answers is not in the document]
rrahn: why? cur and dir are viable known abbreviations

include/seqan3/alignment/matrix/detail/two_dimensional_matrix.hpp
rrahn: Mhmm, why should it not. It constraints the template arguments so a friend declaration shouldn't make an exception

rrahn:
@marehr I addressed all your issues

include/seqan3/core/detail/debug_stream_range.hpp
```diff
@@ -17,6 +17,52 @@
...
+ *
+ * \note If you need to further refine this concept for your type ensure that this concept comes before any
+ *       additional refinement, otherwise your concept definition/requires clause is ill-formed according to the
+ *       subsumption rules for concepts.
+ */
+template <typename rng_t>
+SEQAN3_CONCEPT DebugStreamableRange =
+    !std::Same<remove_cvref_t<reference_t<rng_t>>, remove_cvref_t<rng_t>> && // prevent recursive instantiation
+    // exclude null-terminated strings:
+    !(std::is_pointer_v<std::decay_t<rng_t>> && std::Same<remove_cvref_t<reference_t<rng_t>>, char>);
```
marehr: You removed `requires (reference_t<rng_t> l) { { s << l }; }` right? I imagine that that caused problems.

Because there is no debug_stream included in DebugStreamableRange and thus makes the name confusing.

include/seqan3/core/detail/debug_stream_range.hpp
```diff
@@ -17,6 +17,52 @@
...
+ * This concept refines the std::ranges::InputRange concept to allow streaming the range object to the debug stream,
+ * with the following requirements:
+ *
+ * * `rng_t` is not the same type as `reference_t<rng_t>`,
+ * * `rng_t` is not a pointer or c-style array,
+ * * `reference_t<rng_t>` is not `char`.
+ *
+ * \note If you need to further refine this concept for your type ensure that this concept comes before any
+ *       additional refinement, otherwise your concept definition/requires clause is ill-formed according to the
+ *       subsumption rules for concepts.
```
marehr: can you give a code example what you mean with this?

include/seqan3/core/detail/debug_stream_range.hpp
```diff
@@ -17,6 +17,52 @@
...
+ * Evaluates to `true` if the following expression is valid: `debug_stream << *rng.begin();`, where rng is of  type
+ * rng_t. Otherwise false.
+ */
+template <std::ranges::Range rng_t, typename char_t>
+constexpr bool reference_type_is_streamable_v = false;
+
+//!\cond
+template <std::ranges::Range rng_t, typename char_t>
+    requires requires (reference_t<rng_t> l, debug_stream_type<char_t> s) { { s << l }; }
+constexpr bool reference_type_is_streamable_v<rng_t, char_t> = true;
```
marehr: could we not include this into DebugStreamableRange again?

include/seqan3/core/type_traits/deferred_crtp_base.hpp
```diff
@@ -47,6 +48,39 @@ struct deferred_crtp_base
...
+ *
+ * The following snippet demonstrates the use of the deferred crtp base class instantiation.
+ *
+ * \include test/snippet/core/type_traits/deferred_crtp_base.cpp
+ *
+ * \see seqan3::detail::invoke_deferred_crtp_base
+ * \see seqan3::detail::deferred_crtp_base
+ */
+template <template <typename, auto ...> typename crtp_base, auto ...args>
+struct deferred_crtp_base_vargs
```
marehr: I haven't found this applied in this PR, so I think it is technically unrelated to this PR, but it looks fine.

include/seqan3/alignment/matrix/debug_matrix.hpp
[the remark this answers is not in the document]
rrahn: > This solution will not work on a standard compliant concepts implementation, because subsumption will only work on actual concept definitions (this changed in the standard at some point of time and is now wrong behaviour from gcc). That means that this overload is not more specialized than the Ranges one. We would need a new concept, e.g. RangableMatrix, RangeMatrix, or so, that refines Range.

So you saying:
```c++
template <typename t>
    requires ConceptA<t>
void foo();

template <typename t>
    requires ConceptA<t> && ConceptB<t>
void foo();
```

does not work but:


```c++
template <typename t>
    requires ConceptA<t>
void foo();

template <typename t>
concept ConceptC = ConceptA<t> && ConceptB<t>;

template <typename t>
    requires ConceptC<t>
void foo();
```
does work? What am I missing here?
Note that in the above code everything is a concept defintion and that the ranges overload uses the first two concepts in the exact same order.

```cpp
template <std::ranges::InputRange t, typename char_t>
    requires detail::DebugStreamableRange<t>
inline debug_stream_type<char_t> & operator<<(...)
{...}
```
So just aliasing the expression as with another concept name should not make a difference, should it?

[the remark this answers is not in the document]
rrahn: >I imagine that the range of the matrix is over all elements? Or is it a range over the columns? Or is it a range over the row? Because of this ambiguity, I would not make the matrix a range itself, but add functions like columns(), rows(), entries() to make it clear (those give then ranges back).

It implements a simple range interface providing a two_dimensional iterator. the layout of the matrix is independent of that. You always iterate along the major column order of the matrix. There is no ambiguity.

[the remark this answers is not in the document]
rrahn: >In general this use-case might give us headaches in other areas too. If one type satisfies two distinct concepts, but there are two overloads for the two concepts, they will fail on you. I have an idea how to find this problems with CI.
We should have a new test suite (like /test/unit, test/coverage) where every test header includes the complete library (seqan3/all.hpp) (normally a test only includes a subset and might miss this problem). By doing this we will get compiler errors for this use case and can fix them.

Yes, we already discussed separating unit tests and coverage tests. If you have an idea to improve it please add a card on the test module board. But for now this has absolutely no priority!!!

include/seqan3/core/detail/debug_stream_range.hpp
rrahn: that was something you want to statically assert for but not make overload resolution. if the type is not streamable to the stream, what would be the correct overload then? So I don't see the point of making this a constraint.

>Because there is no debug_stream included in DebugStreamableRange and thus makes the name confusing.

I don't understand what you mean?!

include/seqan3/core/detail/debug_stream_range.hpp
rrahn: 😭 Ok

include/seqan3/core/detail/debug_stream_range.hpp
rrahn: Again, this is not something you want to select another overload for.

include/seqan3/core/type_traits/deferred_crtp_base.hpp
rrahn: It was related at some point, but the code changed and I didn't need it anymore, but figured it might be still useful in other places. So we have it already.

include/seqan3/alignment/matrix/debug_matrix.hpp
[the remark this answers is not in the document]
marehr: > > This solution will not work on a standard compliant concepts implementation, because subsumption will only work on actual concept definitions (this changed in the standard at some point of time and is now wrong behaviour from gcc). That means that this overload is not more specialized than the Ranges one. We would need a new concept, e.g. RangableMatrix, RangeMatrix, or so, that refines Range.
> 
> So you saying:
> 
> ```c++
> template <typename t>
>     requires ConceptA<t>
> void foo();
> 
> template <typename t>
>     requires ConceptA<t> && ConceptB<t>
> void foo();
> ```
> 
> does not work but:
> 
> ```c++
> template <typename t>
>     requires ConceptA<t>
> void foo();
> 
> template <typename t>
> concept ConceptC = ConceptA<t> && ConceptB<t>;
> 
> template <typename t>
>     requires ConceptC<t>
> void foo();
> ```
> 
> does work? What am I missing here?

Yes this is working! That's exactly what I wanted to express. 

See https://en.cppreference.com/w/cpp/language/constraints#Atomic_constraints they give an example:

```c++
template<class T> constexpr bool is_meowable = true;
template<class T> constexpr bool is_cat = true;
 
template<class T>
concept Meowable = is_meowable<T>;
 
template<class T>
concept BadMeowableCat = is_meowable<T> && is_cat<T>;
 
template<class T>
concept GoodMeowableCat = Meowable<T> && is_cat<T>;
 
template<Meowable T>
void f1(T); // #1
 
template<BadMeowableCat T>
void f1(T); // #2
 
template<Meowable T>
void f2(T); // #3
 
template<GoodMeowableCat T>
void f2(T); // #4
 
void g(){
    f1(0); // error, ambiguous:
           // the is_meowable<T> in Meowable and BadMeowableCat forms distinct
           // atomic constraints that are not identical (and so do not subsume each other)
 
    f2(0); // OK, calls #4, more constrained than #3
           // GoodMeowableCat got its is_meowable<T> from Meowable
}
```

> Note that in the above code everything is a concept defintion and that the ranges overload uses the first two concepts in the exact same order.
> 
> ```c++
> template <std::ranges::InputRange t, typename char_t>
>     requires detail::DebugStreamableRange<t>
> inline debug_stream_type<char_t> & operator<<(...)
> {...}
> ```
> 
> So just aliasing the expression as with another concept name should not make a difference, should it?

It does.

include/seqan3/alignment/matrix/detail/two_dimensional_matrix.hpp
marehr: I'm not completely sure, but I guess that there is no wording in the standard for this and practically is a forward declaration. And also haven't seen any rules for forward declarations.

include/seqan3/core/detail/debug_stream_range.hpp
marehr: The name is DebugStreamableRange but there are no checks for debug_streamable

include/seqan3/alignment/matrix/debug_matrix.hpp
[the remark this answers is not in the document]
rrahn: Mhmm, I still have issues with this. According to partial ordering of constraints: `If declarations D1 and D2 are constrained and D1's associated constraints subsume D2's associated constraints (or if D2 is unconstrained), then D1 is said to be at least as constrained as D2. If D1 is at least as constrained as D2 and D2 is not at least as constrained as D1, then D1 is more constrained than D2. `

So I cannot read from this, that a conjunction of atomic constraints subsumes only another conjunction of constraints if it was first aliased into another constrained. 
Why should `A & B & C` not apply to these rules and `D = A & B & C` does? Can we discuss this offline?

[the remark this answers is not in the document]
rrahn: Ok, I tested it on Godbolt and I have to say that I am right 🏆 🤣 
No, joke aside: https://godbolt.org/z/rMGnpn 
Please help me understand what I am not seeing here. The example as given does work with the current clang concepts branch and is a simplified form of the version above.

include/seqan3/alignment/matrix/detail/two_dimensional_matrix.hpp
rrahn: Same with me. I couldn't find any documentation against it. So I assume as a template expression I can apply a requires clause.

include/seqan3/core/detail/debug_stream_range.hpp
rrahn: Ok, the name is weird. I can change this. I didn't want to expose these restrictions into the public user interface but they are necessary since you want to call another overload if it does not fulfill the range requirements for our debug_stream overload (e.g. filesystems directory_iterator). But that might be a problem as well if we allow users to add their own types to be printable to the debug_stream and they happen to be a range as well.

include/seqan3/core/detail/debug_stream_range.hpp
rrahn: I did remove it because it is not valid. The order of the constraints is not relevant as long as they are atomically the same (and refined of course)

PR #1160, opened 2019-07-05
[FEATURE] Add to_chars overload for floating points.

smehringer:


include/seqan3/std/charconv
```diff
@@ -859,6 +859,23 @@ inline std::to_chars_result to_chars(char * first, char * last, value_type value
...
+//!\brief std::to_chars overload for floating point via a stringstream.
+//!\ingroup charconv
+template <seqan3::FloatingPoint floating_point_type>
+inline std::to_chars_result to_chars(char * first, char * last, floating_point_type value, int = 10) noexcept
+{
+    std::ostringstream ss;
+    ss << value;
+    auto str = ss.str();
+
+    if (last - first < static_cast<int16_t>(str.size()))
```
joergi-w: Why `int16_t` instead of `std::ptrdiff_t`?

include/seqan3/std/charconv
```diff
@@ -859,6 +859,23 @@ inline std::to_chars_result to_chars(char * first, char * last, value_type value
     return seqan3::detail::to_chars_itoa(first, last, value, is_signed<value_type>());
 }
 
+//!\brief std::to_chars overload for floating point via a stringstream.
+//!\ingroup charconv
+template <seqan3::FloatingPoint floating_point_type>
+inline std::to_chars_result to_chars(char * first, char * last, floating_point_type value, int = 10) noexcept
```
joergi-w: Parameter 4 has no name, but a default value? Why?

smehringer: I thought it has to be consistent with the integral overload, but actually I do not support a diffferent base then 10 so I removed the parameter.

include/seqan3/std/charconv
smehringer: I used ptrdiff_t instead. Although I know that a proper arithmetic value will never be longer than 316 chars, this might avoid overflows.

smehringer:
@rrahn kind review ping :) this is a tiny PR that I would like to get in ASAP as another one depends on it.

test/unit/std/charconv_test.cpp
```diff
@@ -220,14 +224,31 @@ TYPED_TEST(integral_from_char_test, hexadicimal_number)
...
 {
     TypeParam val{120};
     char buffer[10];
 
-        // buffer.clear();
-    [[maybe_unused]] auto res = std::to_chars(&buffer[0], &buffer[0] + sizeof(buffer), val);
+    auto res = std::to_chars(&buffer[0], &buffer[0] + sizeof(buffer), val);
 
+    EXPECT_EQ(res.ptr, &buffer[3]);
+    EXPECT_EQ(res.ec, std::errc{});
```
rrahn: Error codes have an implicit bool conversion that I would use instead.

test/unit/std/charconv_test.cpp
```diff
@@ -220,14 +224,31 @@ TYPED_TEST(integral_from_char_test, hexadicimal_number)
...
+// =============================================================================
+
+TYPED_TEST(integral_from_char_test, to_chars)
 {
     TypeParam val{120};
     char buffer[10];
 
-        // buffer.clear();
-    [[maybe_unused]] auto res = std::to_chars(&buffer[0], &buffer[0] + sizeof(buffer), val);
+    auto res = std::to_chars(&buffer[0], &buffer[0] + sizeof(buffer), val);
```
rrahn: I recommend using std::array instead since it would be the c++ way.

test/unit/std/charconv_test.cpp
```diff
@@ -483,3 +504,19 @@ TYPED_TEST(from_char_real_test, non_valid_strings)
...
+TYPED_TEST(from_char_real_test, to_chars)
+{
+    TypeParam val{120.3};
+    char buffer[10];
+
+    auto res = std::to_chars(&buffer[0], &buffer[0] + sizeof(buffer), val);
+
+    EXPECT_EQ(res.ptr, &buffer[5]);
+    EXPECT_EQ(res.ec, std::errc{});
+    EXPECT_EQ((std::string_view{buffer, 5}), std::string_view{"120.3"});
```
rrahn: Again error code bool conversion as well as std::array over c-array

test/unit/std/charconv_test.cpp
smehringer: google test macro wants to apply the `operator!` first, since I expect the errc to be zero, and fails because the operand type is std::errc. I wouldn't want to static_cast into bool, I think the EQ is then more expressive?

include/seqan3/std/charconv
```diff
@@ -859,6 +859,23 @@ inline std::to_chars_result to_chars(char * first, char * last, value_type value
     return seqan3::detail::to_chars_itoa(first, last, value, is_signed<value_type>());
 }
 
+//!\brief std::to_chars overload for floating point via a std::stringstream for default base = 10.
+//!\ingroup charconv
+template <seqan3::FloatingPoint floating_point_type>
+inline std::to_chars_result to_chars(char * first, char * last, floating_point_type value) noexcept
+{
+    std::ostringstream ss;
```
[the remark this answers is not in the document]
rrahn: Maybe in 2 separate asserts so you can see by the line number which one failed.

test/unit/std/charconv_test.cpp
rrahn: Well then it is ok. If it get‘s to complicated don‘t spent time on that

test/unit/std/charconv_test.cpp
```diff
@@ -220,14 +224,31 @@ TYPED_TEST(integral_from_char_test, hexadicimal_number)
...
+// =============================================================================
+// std::to_chars for integral types
+// =============================================================================
+
+TYPED_TEST(integral_from_char_test, to_chars)
+{
+    TypeParam val{120};
+    std::array<char, 10> buffer{};
+
+    auto res = std::to_chars(buffer.data(), buffer.data() + sizeof(buffer), val);
```
rrahn:
```suggestion
    auto res = std::to_chars(buffer.data(), buffer.data() + buffer.size(), val);
```

test/unit/std/charconv_test.cpp
```diff
@@ -220,14 +224,31 @@ TYPED_TEST(integral_from_char_test, hexadicimal_number)
...
+
+TYPED_TEST(integral_from_char_test, to_chars_error)
 {
     TypeParam val{120};
-    char buffer[10];
+    std::array<char, 1> buffer{};
 
-        // buffer.clear();
-    [[maybe_unused]] auto res = std::to_chars(&buffer[0], &buffer[0] + sizeof(buffer), val);
+    auto res = std::to_chars(buffer.data(), buffer.data() + sizeof(buffer), val);
```
rrahn:
```suggestion
    auto res = std::to_chars(buffer.data(), buffer.data() + buffer.size(), val);
```

test/unit/std/charconv_test.cpp
```diff
@@ -483,3 +504,19 @@ TYPED_TEST(from_char_real_test, non_valid_strings)
...
+// =============================================================================
+// std::to_chars for float, double and long double
+// =============================================================================
+
+TYPED_TEST(from_char_real_test, to_chars)
+{
+    TypeParam val{120.3};
+    std::array<char, 10> buffer{};
+
+    auto res = std::to_chars(buffer.data(), buffer.data() + sizeof(buffer), val);
```
rrahn:
```suggestion
    auto res = std::to_chars(buffer.data(), buffer.data() + buffer.size(), val);
```

include/seqan3/std/charconv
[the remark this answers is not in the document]
smehringer: fine :roll_eyes:

PR #1161, opened 2019-07-05
[FIX] there is no #elseif

h-2:
fix a line introduced by #1090


h-2:
Jenkins failure unrelated (again...).

rrahn:
>Jenkins failure unrelated (again...).

It has nothing to do with jenkins this time. 
Somehow ctest erros when submitting the  test results as it seems to stall to submit to cdash. The reason for this is not clear yet. 
I will change the script to not fail when ctest is not able to submit the data.
Btw. travis is failing also sometimes due to some unrelated things.

PR #1171, opened 2019-07-09
[FEATURE] view_translate_join

sarahet:
This PR introduces the ```view_translate_join```

* Enables translation of two-dimensional input ranges into two-dimensional output ranges where ranges of frames per input sequence are joined
* Offers potential for better performance in later applications if this is desired in comparison to standard ```view_translate```


include/seqan3/range/view/translate_join.hpp
```diff
@@ -0,0 +1,405 @@
...
+     *
+     * No-throw guarantee.
+     */
+    iterator begin() noexcept
+    {
+        return {*this, 0};
+    }
+
+    //!\overload
+    const_iterator begin() const noexcept
```
h-2:
```suggestion
    const_iterator begin() const noexcept
        requires ConstIterableRange<urng_t>
```

Also for the other const-qualified member functions.

include/seqan3/range/view/translate_join.hpp
```diff
@@ -0,0 +1,405 @@
...
+     *
+     * ### Complexity
+     *
+     * Constant.
+     */
+    reference operator[](size_type const n)
+    {
+        assert(n < size());
+        unsigned index_frame = n % selected_frames.size();
+        unsigned index_urange = (n - index_frame) / selected_frames.size();
```
h-2:
```suggestion
        size_type index_urange = (n - index_frame) / selected_frames.size();
```
Also in the line above.

include/seqan3/range/view/translate_join.hpp
```diff
@@ -0,0 +1,405 @@
...
+        return urange[index_urange] | view::translate_single(selected_frames[index_frame]);
+    }
+
+    //!\overload
+    const_reference operator[](size_type const n) const
+    {
+        assert(n < size());
+        unsigned index_frame = n % selected_frames.size();
+        unsigned index_urange = (n - index_frame) / selected_frames.size();
```
h-2: here as well

include/seqan3/range/view/translate_join.hpp
```diff
@@ -0,0 +1,405 @@
...
+    }
+    //!\}
+};
+
+//!\brief Class template argument deduction for view_translate_join.
+template <typename urng_t>
+//!\cond
+    requires std::ranges::SizedRange<urng_t> &&
+             std::ranges::RandomAccessRange<urng_t> &&
+             NucleotideAlphabet<innermost_value_type_t<urng_t>>
```
h-2: You are not checking the same constraints as on the class.

include/seqan3/range/view/translate_join.hpp
```diff
@@ -0,0 +1,405 @@
...
+//!\endcond
+view_translate_join(urng_t &&, translation_frames const = translation_frames{}) -> view_translate_join<std::ranges::all_view<urng_t>>;
```
h-2:
```suggestion
view_translate_join(urng_t &&, translation_frames const = translation_frames{}) -> view_translate_join<all_view<urng_t>>;
```

include/seqan3/range/view/translate_join.hpp
```diff
@@ -0,0 +1,405 @@
...
+ * | std::ranges::View               |                                       | *guaranteed*                                       |
+ * | std::ranges::SizedRange         | *required*                            | *preserved*                                        |
+ * | std::ranges::CommonRange        |                                       | *guaranteed*                                       |
+ * | std::ranges::OutputRange        |                                       | *lost*                                             |
+ * | seqan3::ConstIterableRange      | *required*                            | *preserved*                                        |
+ * |                                 |                                       |                                                    |
+ * | seqan3::reference_t             | seqan3::NucleotideAlphabet            | std::ranges::View && std::ranges::RandomAccessRange && std::ranges::SizedRange |
+ *
+ * * `urng_t` is the type of the range modified by this view (input).
+ * * `rrng_type` is the type of the range returned by this view.
```
h-2:
```suggestion
 * * `rrng_t` is the type of the range returned by this view.
```

include/seqan3/range/view/translate_join.hpp
```diff
@@ -0,0 +1,405 @@
...
+ * \details
+ *
+ * This view can be used to translate nucleotide sequences into aminoacid sequences (see translation_frames for possible combination of frames).
+ * This view only operates on two-dimensional input (range of ranges) and outputs a range of ranges no matter the number of input sequences or
+ * the number of translation frames given. Therefore, it has the same capabilities as the standard view_translate but concatenates the different
+ * frames of the different input sequences rather than having a separate range for each input sequence containing the translated frames. In the output,
+ * frames are ordered in a way, that all requested frames are listed per sequence directly after each other in the order of input sequences.
+ * improved and efficient downstream post-processing if needed. However, the index of a frame for a specific sequence needs to be calculated via
+ * modulo operations in this case. The i-th frame of the j-th sequence can be calculated by n = (i * s) + j, where s is the number of frames used
+ * for translation (index starting at zero).
```
h-2: I would add a paragraph:

In short, this views behaves the same as:
```cpp
std::vector<std::vector<dna5>> vec = /* ... */;
auto v = vec | view::translate | std::view::join;
```
Except that the performance is better and the returned range still models std::ranges::RandomAccessRange and std::ranges::SizedRange.

include/seqan3/range/view/translate_join.hpp
```diff
@@ -0,0 +1,405 @@
...
+ * NOTE: If improved performance is not needed, the standard view_translate is recommended as it does not require further calculations when
+ * accessing a specific frame/sequence.
```
h-2: Well the main argument is that one want the "join" so I would say people should always use this when that is the desired behaviour. But maybe we can just omit this sentence? People will look for `view::translate` befor they look for `view::translate_join` anyways, I think...

test/snippet/range/view/translate_join.cpp
```diff
@@ -0,0 +1,39 @@
...
+int main()
+{
+    std::string const in1{"ACGTACGTACGTA"};
+    std::string const in2{"TCGAGAGCTTTAGC"};
+
+    // Input range needs to be two-dimensional
+    std::vector<std::vector<seqan3::dna4> > vec;
+    vec.resize(2);
+    vec[0] = in1 | seqan3::view::char_to<seqan3::dna4>;
+    vec[1] = in2 | seqan3::view::char_to<seqan3::dna4>;
```
h-2: why not directly initialise vec like this:
```cpp
std::vector<std::vector<seqan3::dna4> > vec{"ACGTACGTACGTA"_dna4, "TCGAGAGCTTTAGC"_dna4};
```
It's a little shorter 😉

include/seqan3/range/view/translate_join.hpp
```diff
@@ -0,0 +1,405 @@
...
+#include <seqan3/core/add_enum_bitwise_operators.hpp>
+#include <seqan3/core/type_traits/range.hpp>
+#include <seqan3/range/container/small_string.hpp>
+#include <seqan3/range/detail/random_access_iterator.hpp>
+#include <seqan3/range/view/deep.hpp>
+#include <seqan3/range/view/detail.hpp>
+#include <seqan3/range/view/translate.hpp>
+#include <seqan3/std/concepts>
+#include <seqan3/std/ranges>
+#include <seqan3/range/container/concept.hpp>
```
h-2: Which of these headers do you really need? I think container, dna5, aa27 are all not needed...

It seems like the following would do:
```cpp
include <seqan3/core/type_traits/range.hpp>
#include <seqan3/range/container/small_string.hpp>
#include <seqan3/range/detail/random_access_iterator.hpp>
#include <seqan3/range/view/translate.hpp>
#include <seqan3/std/ranges>
```
(just guessing)

test/unit/range/view/view_translate_join_test.cpp
```diff
@@ -0,0 +1,158 @@
...
+    EXPECT_TRUE(std::ranges::RandomAccessRange<decltype(v1)>);
+    EXPECT_TRUE(std::ranges::SizedRange<decltype(v1)>);
+    EXPECT_TRUE(std::ranges::View<decltype(v1)>);
+    EXPECT_TRUE(std::ranges::RandomAccessRange<value_type_t<decltype(v1)>>);
+    EXPECT_TRUE(std::ranges::SizedRange<value_type_t<decltype(v1)>>);
+    EXPECT_TRUE(std::ranges::View<value_type_t<decltype(v1)>>);
+    EXPECT_TRUE(std::ranges::RandomAccessRange<reference_t<decltype(v1)>>);
+    EXPECT_TRUE(std::ranges::SizedRange<reference_t<decltype(v1)>>);
+    EXPECT_TRUE(std::ranges::View<reference_t<decltype(v1)>>);
+}
```
h-2: We now have a test_template for the iterators of ranges. Can you ask @smehringer how to apply this for your view? It's likely only a few lines...

test/unit/range/view/view_translate_test.cpp
```diff
@@ -93,7 +93,7 @@ TYPED_TEST(nucleotide, view_translate)
 
     // default parameter translation_frames
     auto v2 = vec | view::translate();
-    // == [[T,Y,V,R]]
+    // == [[T,Y,V,R],[R,T,Y,V],[V,R,T],[Y,V,R,T],[T,Y,V,R],[R,T,Y]]
```
h-2: good catch!

test/unit/range/view/view_translate_test.cpp
```diff
@@ -156,7 +156,7 @@ TYPED_TEST(nucleotide, view_translate)
 
     // combinability and function syntax
     auto v11 = detail::view_translate(view::complement(vec), translation_frames::FWD_REV_0);
-    // == [[T,Y,V,R],[Y,V,R,T]]
+    // == [[C,M,H,A],[M,H,A,C]]
```
h-2: really? 🙈

include/seqan3/range/view/translate_join.hpp
```diff
@@ -0,0 +1,405 @@
...
+        return detail::view_translate_join{std::forward<urng_t>(urange), tf};
+    }
+
+    //!\brief This adaptor is usable without setting the frames parameter in which case the default is chosen.
+    template <std::ranges::Range urng_t>
+    constexpr friend auto operator|(urng_t && urange, translate_join_fn const & me)
+    {
+        return me(std::forward<urng_t>(urange));
+    }
+};
```
h-2: I think you can move this below the class and get rid of the forward...

include/seqan3/range/view/translate_join.hpp
```diff
@@ -0,0 +1,405 @@
...
+     */
+    //!\cond
+    // unfortunately we cannot specialise the variable template so we have to add an auxiliary here
+    template <typename t>
+        requires (dimension_v<t> == dimension_v<value_type> + 1) &&
+                 std::is_same_v<remove_cvref_t<innermost_value_type_t<value_type>>,
+                                remove_cvref_t<innermost_value_type_t<t>>>
+    static constexpr bool is_compatible_this_aux = true;
+    //!\endcond
+    //!\}
```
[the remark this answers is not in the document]
sarahet: I must admit I wasn't sure what it does so I left it from the ```view_translate``` :see_no_evil: 
Could you explain this?

include/seqan3/range/view/translate_join.hpp
sarahet: For what would you need ```<seqan3/core/type_traits/range.hpp>```?
But you are right few were not needed. However, you still need ```view_complement``` and ```view_char_to``` as they are used in the test.

test/unit/range/view/view_translate_test.cpp
sarahet: Yes, it is the same as ```v9``` as first the sequence complement is computed and then translated. Or am I missing something?

include/seqan3/range/view/translate_join.hpp
sarahet: What do you mean? The whole adaptor?

include/seqan3/range/view/translate_join.hpp
```diff
@@ -0,0 +1,405 @@
...
+     * \param[in] _tf The frames that should be used for translation.
+     */
+    template <typename rng_t>
+    //!\cond
+        requires !std::Same<remove_cvref_t<rng_t>, view_translate_join> &&
+                 std::ranges::ViewableRange<rng_t> &&
+                 std::Constructible<urng_t, ranges::ref_view<std::remove_reference_t<rng_t>>>
+    //!\endcond
+    view_translate_join(rng_t && _urange, translation_frames const _tf = translation_frames::SIX_FRAME)
+     : view_translate_join{std::view::all(std::forward<rng_t>(_urange)), _tf}
```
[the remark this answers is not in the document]
sarahet: When changing this (also holds for the above comment), I run into some problems. The view computes the correct results and the standard tests work but the output cannot be combined with another range anymore, cannot passed to ```debug_stream``` and concept check ```std::ranges::View<decltype(v1)>``` and all checks concerning ```value_type_t<decltype(v1)>``` cannot be built (all errors already arise when compiling). Is there something specific for the ```seqan3::view::all``` that I need to consider?

include/seqan3/range/view/translate_join.hpp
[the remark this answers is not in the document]
h-2: I think it's not required in the view_translate anymore either. It was a "helper concept" (not a real concept, just a variable) used to determine if you can conconstruct from something or convert to it...

include/seqan3/range/view/translate_join.hpp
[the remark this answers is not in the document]
h-2: Hm, it should work ™️ 

If it doesn't just keep std::view::all and I will have a look at it another time.

include/seqan3/range/view/translate_join.hpp
h-2: > For what would you need <seqan3/core/type_traits/range.hpp>?

`reference_t`

> However, you still need view_complement and view_char_to as they are used in the test.

If you just need them in the test, just include them in the test!

test/unit/range/view/view_translate_test.cpp
h-2: Could well be, I didnt double check. It reminds me that we need to add infrstructure to test the output of snippets.

include/seqan3/range/view/translate_join.hpp
h-2: yep! Usually the adaptor is always below. It just changed places in the translate.hpp because there are multiple views defined inside and one needs the other...

test/unit/range/view/view_translate_test.cpp
sarahet: It is at least compared to the vector that consists of these exact frames so since the test runs through it seems to be correct. Also this is not a snippet but a test case, right?

h-2: My whole comment was just meant as "In general, maybe there should be something that checks the actual values"!
It was not meant as criticism of the change! I totally believe you that you checked the output.

sarahet: Then I agree :wink:

include/seqan3/range/view/translate_join.hpp
sarahet: Putting the include into the test case does not work .. Is that possible for typed tests? Also the includes of ```view_char_to``` for example are always on top also for other view tests. Is that something that is not desired?

h-2: > Putting the include into the test case does not work .. Is that possible for typed tests? 

I don't understand what you mean. What doesn't work and why? In general every header should include what it uses and not more. Headers required by tests should only be included by the tests...

> Also the includes of view_char_to for example are always on top also for other view tests.

On top of what? 🤔 😆

sarahet: Oooops, I did not read properly .. Thought those are the includes from the test and was extremely confused. Thought you meant putting the include into the test case itself :laughing: 
Sorry for that .. :smile:

sarahet: That's also why I asked for what you would need the type traits ......... :see_no_evil:

sarahet:
@h-2 Should we also have an iterator test for all nucleotide alphabets?

include/seqan3/range/view/translate_join.hpp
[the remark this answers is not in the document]
sarahet: Ok, let's do that

include/seqan3/range/view/translate_join.hpp
```diff
@@ -0,0 +1,405 @@
...
+
+    /*!\brief Construct from another range.
+     * \param[in] _urange The underlying range (of ranges).
+     * \param[in] _tf The frames that should be used for translation.
+     */
+    template <typename rng_t>
+    //!\cond
+        requires !std::Same<remove_cvref_t<rng_t>, view_translate_join> &&
+                 std::ranges::ViewableRange<rng_t> &&
+                 std::Constructible<urng_t, ranges::ref_view<std::remove_reference_t<rng_t>>>
```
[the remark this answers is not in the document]
h-2:
```suggestion
                 std::Constructible<urng_t, std::ranges::all_view<rng_t>>
```

include/seqan3/range/view/translate_join.hpp
```diff
@@ -0,0 +1,386 @@
...
+     * No-throw guarantee.
+     */
+    size_type size() noexcept
+    {
+        return (size_type) seqan3::size(urange) * selected_frames.size();
+    }
+
+    //!\overload
+    size_type size() const noexcept
+    {
```
h-2: requires ConstIterable...

include/seqan3/range/view/translate_join.hpp
```diff
@@ -0,0 +1,386 @@
...
+    reference operator[](size_type const n)
+    {
+        assert(n < size());
+        size_type index_frame = n % selected_frames.size();
+        size_type index_urange = (n - index_frame) / selected_frames.size();
+        return urange[index_urange] | view::translate_single(selected_frames[index_frame]);
+    }
+
+    //!\overload
+    const_reference operator[](size_type const n) const
```
h-2: requires ConstIterable...

include/seqan3/range/view/translate_join.hpp
```diff
@@ -0,0 +1,386 @@
...
+ * frames of the different input sequences rather than having a separate range for each input sequence containing the translated frames. In the output,
+ * frames are ordered in a way, that all requested frames are listed per sequence directly after each other in the order of input sequences.
+ * improved and efficient downstream post-processing if needed. However, the index of a frame for a specific sequence needs to be calculated via
+ * modulo operations in this case. The i-th frame of the j-th sequence can be calculated by n = (i * s) + j, where s is the number of frames used
+ * for translation (index starting at zero).
+ *
+ * In short, this views behaves the same as:
+ * std::vector<std::vector<dna5>> vec {...};
+ * auto v = vec | view::translate | std::view::join;
+ * Except that the performance is better and the returned range still models std::ranges::RandomAccessRange and std::ranges::SizedRange.
```
h-2: You forgot the ` ``` `

h-2:
> @h-2 Should we also have an iterator test for all nucleotide alphabets?

We shouldn't need to add this here, but the simple view::translate_single should eventually get those for all nucleotides...

sarahet:
I cannot comment under the suggestion for some reason .. 
The suggested change ```std::Constructible<urng_t, std::ranges::all_view<rng_t>>``` does not work (as the other suggestion from above) .. this is why I did not change it in the first place when you mentioned it (not ignoring you here).

include/seqan3/range/view/translate_join.hpp
sarahet: oops ..

include/seqan3/range/view/translate_join.hpp
```diff
@@ -0,0 +1,390 @@
...
+ * \implements std::ranges::View
+ * \implements std::ranges::SizedRange
+ * \implements std::ranges::RandomAccessRange
+ * \tparam urng_t The type of the range being translated.
+ * \param[in] tf Translation frames to be used.
+ * \ingroup view
+ */
+template <std::ranges::View urng_t>
+//!\cond
+    requires std::ranges::SizedRange<urng_t> &&
```
smehringer: I guess those requirements are not needed for overload resolution but just for "type cheking"? If so we decided to use static asserts rather than concet. The error messages are more readable :)

test/unit/range/view/view_translate_join_test.cpp
```diff
@@ -0,0 +1,180 @@
...
+    using iterator_tag = std::random_access_iterator_tag;
+    static constexpr bool const_iterable = true;
+
+    std::vector<std::vector<seqan3::dna4> > vec{"ACGTACGTACGTA"_dna4, "TCGAGAGCTTTAGC"_dna4};
+    std::vector<std::vector<aa27> > expected_range{{"TYVR"_aa27}, {"RTYV"_aa27}, {"VRT"_aa27}, {"YVRT"_aa27}, {"TYVR"_aa27}, {"RTY"_aa27},
+                                                   {"SRAL"_aa27}, {"REL*"_aa27}, {"ESFS"_aa27}, {"AKAL"_aa27}, {"LKLS"_aa27}, {"*SSR"_aa27}};
+    decltype(view::translate_join(vec)) test_range = view::translate_join(vec);
+
+    template <typename A, typename B>
+    static void expect_eq(A && begin_iterator_value, B && expected_range_value)
```
smehringer:
```suggestion
    static void expect_eq(A && test_range_value, B && expected_range_value)
```
:nail_care: but not neccessary

include/seqan3/range/view/translate_join.hpp
sarahet: Isn't it enough that the concepts are checked via static_assert in the adaptor? This gets anyways called when constructing the range, right?

include/seqan3/range/view/translate_join.hpp
```diff
@@ -0,0 +1,390 @@
...
+ * This view only operates on two-dimensional input (range of ranges) and outputs a range of ranges no matter the number of input sequences or
+ * the number of translation frames given. Therefore, it has the same capabilities as the standard view_translate but concatenates the different
+ * frames of the different input sequences rather than having a separate range for each input sequence containing the translated frames. In the output,
+ * frames are ordered in a way, that all requested frames are listed per sequence directly after each other in the order of input sequences.
+ * improved and efficient downstream post-processing if needed. However, the index of a frame for a specific sequence needs to be calculated via
+ * modulo operations in this case. The i-th frame of the j-th sequence can be calculated by n = (i * s) + j, where s is the number of frames used
+ * for translation (index starting at zero).
+ *
+ * In short, this views behaves the same as:
+ * ```
```
h-2:
```suggestion
 * ```cpp
```
💅

include/seqan3/range/view/translate_join.hpp
```diff
@@ -0,0 +1,390 @@
...
+            "The range parameter to view::translate_join cannot be a temporary of a non-view range.");
+        static_assert(std::ranges::ViewableRange<urng_t>,
+            "The inner range of the range parameter to view::translate_join cannot be a temporary of a non-view range.");
+        static_assert(std::ranges::SizedRange<urng_t>,
+            "The range parameter to view::translate_join must model std::ranges::SizedRange.");
+        static_assert(std::ranges::SizedRange<urng_t>,
+            "The inner range of the range parameter to view::translate_join must model std::ranges::SizedRange.");
+        static_assert(std::ranges::RandomAccessRange<urng_t>,
+            "The range parameter to view::translate_join must model std::ranges::RandomAccessRange.");
+        static_assert(std::ranges::RandomAccessRange<urng_t>,
```
h-2: You forgot `reference_t` for each of the copy'n'pasted concept checks!

include/seqan3/range/view/translate_join.hpp
h-2: There is a general rule in SeqAn about having static_asserts instead of requirements if they are not required for overload resolution. For the regular translate views I would argue that we might actually add other overloads and in that case the constraints for this `class` would be the same, but we would change the constraints on the adaptor (which would then allow e.g. BidirectionalRange but would forward those to `view::translate | view::join`).

TLDR: It doesn't make much of a difference. Since I don't think this particular class will get another specialisation, you can also just copy the static assert from the adaptor into the class and remove the concept constraints.

include/seqan3/range/view/translate_join.hpp
sarahet: Classic! Sorry, I fixed it ..

h-2:
@smehringer Do you want a final review or can I merge this after CI passes?

PR #1174, opened 2019-07-10
[MISC] make the iterator test_template test a range not single iterators.

smehringer:


test/unit/range/detail/random_access_iterator_test.cpp
```diff
@@ -21,20 +21,26 @@
 using iterator_type = seqan3::detail::random_access_iterator<std::vector<int>>;
 using const_iterator_type = seqan3::detail::random_access_iterator<std::vector<int> const>;
 
+struct expose_iterator
+{
+    std::vector<int> * host{nullptr};
+
+    expose_iterator(std::vector<int> * host_) : host{host_} {}
```
rrahn: I would simply copy the vector to avoid possible side effects.

test/unit/range/iterator_test_template.hpp
```diff
@@ -25,35 +25,27 @@ struct iterator_fixture : public ::testing::Test
                                                                // std::bidirectional_iterator_tag
                                                                // std::random_access_iterator_tag
 
-    static constexpr bool const_iterable = true/false;         // Also test const_iterability.
+    static constexpr bool const_iterable = true/false;         // Also test const_iterability. (cbgein/cend required)
```
eseiler:
```suggestion
    static constexpr bool const_iterable = true/false;         // Also test const_iterability. (cbegin/cend required)
```

test/unit/range/iterator_test_template.hpp
```diff
@@ -25,35 +25,27 @@ struct iterator_fixture : public ::testing::Test
...
-
-    T begin_iterator;                                          // Begin of the iterator (range) to test.
-    T end_iterator;                                            // End of the iterator (range) to test.
-
-    --------------------------------------------------------------------------------------------------------------------
-    // Only need to be specified if const_iterable == true.
-    --------------------------------------------------------------------------------------------------------------------
-    const_T begin_const_iterator;                               // Begin of the const iterator (range) to test.
-    const_T end_const_iterator;                                 // end of the const iterator (range) to test.
+    t1 test_range;                                             // The range to test the iterators (bgein/end required).
```
eseiler:
```suggestion
    t1 test_range;                                             // The range to test the iterators (begin/end required).
```

rrahn:
@smehringer can you fix the comments as suggested by eseiler. Then we can merge it

PR #1178, opened 2019-07-11
5 small argument parser fixes

smehringer:
1. Windows command was not terminated by a semicolon :see_no_evil: 

2. The following is invalid should throw directly when calling `add_positional_option` not when calling `parse()`.
```cpp
    std::vector<int> listA;
    whatever_type optionB;

    myparser.add_positional_option(listA, ...);
    myparser.add_positional_option(optionB, ...); // only the last positional option may be a list (ambigiousness)

    myparser.parse() // current implementation throws an design_error here, not when adding the option
```

3. Options that are set as `option_spec::REQUIRED` should not print a default value on the help page. Since positional_options are always required, they should never have a default. Except the very last positional option that might be a list and can be empty (thus is not required).
@h-2 I think we talked about this?

4. unit test `parse_called_twice` was failing on travis because stderr is checked and was non-empty because of the version check. Explicitely turn off the version check for this test.

5. Exclude the interactive code path of asking the user about update notifications from the code coverage.

### DIFF
--- a/include/seqan3/argument_parser/argument_parser.hpp
+++ b/include/seqan3/argument_parser/argument_parser.hpp
@@ -287,6 +287,13 @@ public:
                                std::string const & desc,
                                validator_type validator = validator_type{}) // copy to bind rvalues
     {
+        if (has_positional_list_option)
+            throw parser_design_error{"You added a positional option with a list value before so you cannot add "
+                                      "any other positional options."};
+
+        if constexpr (SequenceContainer<option_type> && !std::Same<option_type, std::string>)
+            has_positional_list_option = true; // keep track of a list option because there must be only one!
+
         // copy variables into the lambda because the calls are pushed to a stack
         // and the references would go out of scope.
         std::visit([=, &value] (auto & f) { f.add_positional_option(value, desc, validator); }, format);
@@ -491,6 +498,9 @@ public:
 private:
     //!\brief Keeps track of whether the parse function has been called already.
     bool parse_was_called{false};
+
+    //!\brief Keeps track of whether the user has added a positional list option to check if this was the very last.
+    bool has_positional_list_option{false};
 
     //!\brief Set on construction and indicates whether the developer deactivates the version check calls completely.
     bool version_check_dev_decision{};
--- a/include/seqan3/argument_parser/detail/format_parse.hpp
+++ b/include/seqan3/argument_parser/detail/format_parse.hpp
@@ -683,8 +683,7 @@ private:
 
         if (SequenceContainer<option_type> && !std::is_same_v<option_type, std::string>) // vector/list will be filled with all remaining arguments
         {
-            if (positional_option_count != (positional_option_calls.size()))
-                throw parser_design_error("Lists are only allowed as the last positional option!");
+            assert(positional_option_count == positional_option_calls.size()); // checked on set up.
 
             while (it != argv.end())
             {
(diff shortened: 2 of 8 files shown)


include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -492,6 +499,9 @@ class argument_parser
     //!\brief Keeps track of whether the parse function has been called already.
     bool parse_was_called{false};
 
+    //!\brief Keeps track of whether theuser added a list position option to check if this was the very last.
+    bool positional_option_list_was_added{false};
```
joergi-w: an easier name would be `has_positional_list_option`

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -492,6 +499,9 @@ class argument_parser
     //!\brief Keeps track of whether the parse function has been called already.
     bool parse_was_called{false};
 
+    //!\brief Keeps track of whether theuser added a list position option to check if this was the very last.
```
joergi-w:
```suggestion
    //!\brief Keeps track of whether the user has added a positional list option to check if this was the very last.
```

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -287,6 +287,13 @@ class argument_parser
                                std::string const & desc,
                                validator_type validator = validator_type{}) // copy to bind rvalues
     {
+        if (positional_option_list_was_added)
```
joergi-w: As the variable was set in an `if constexpr` statement the variable itself could be compile time constant and then also this if expression can become constexpr?

smehringer: No, because I cannot track the order of functions calls at compile time and if I model the `positional_option_list_was_added` as a constexpr variable, I set its state once and than I cannot change it :/

include/seqan3/argument_parser/detail/version_check.hpp
```diff
@@ -403,7 +403,7 @@ class version_checker
 )";
             return true; // default: check version if you cannot ask the user
         }
-    }
+    } // LCOV_EXCL_STOP
```
joergi-w: You place it here, because the current line is not part of the excluded section. However, it is a slightly confusing layout because the corresponding bracket for the excluded if/else block is one line above. My idea: Insert the comment in its own line between the closing brackets and use the indentation of the if/else block.

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -683,8 +683,7 @@ class format_parse : public format_base
 
         if (SequenceContainer<option_type> && !std::is_same_v<option_type, std::string>) // vector/list will be filled with all remaining arguments
         {
-            if (positional_option_count != (positional_option_calls.size()))
-                throw parser_design_error("Lists are only allowed as the last positional option!");
+            assert(positional_option_count == (positional_option_calls.size())); // checked on set up.
```
rrahn:
```suggestion
            assert(positional_option_count == positional_option_calls.size()); // checked on set up.
```

test/unit/argument_parser/detail/format_help_test.cpp
```diff
@@ -238,7 +247,8 @@ TEST(help_page_printing, full_information)
                "example\n"
                "example2" +
                basic_version_str;
-    EXPECT_TRUE(ranges::equal((std_cout | std::view::filter(!is_space)), expected | std::view::filter(!is_space)));
+    EXPECT_TRUE(ranges::equal((std_cout | std::view::filter(!is_space)), expected | std::view::filter(!is_space))) <<
+      "\nexpected:\n----\n" << expected << "\n\ncurrent:\n----\n" << std_cout;
```
rrahn: indentation?

smehringer: I'll just remove this again. It was just for debugging.

PR #1179, opened 2019-07-11
[MISC] Update iterator test template

rrahn:
Uses the iterator type returned from the test range instead of
TypeParam. It cannot always be ensured that the TypeParam is the
iterator type to test.


rrahn:
only coverage test failed which is unrelated.

PR #1180, opened 2019-07-11
[FIX] Fixes the check for expect_eq member.

rrahn:


rrahn:
@smehringer this fixes an issue introduced with the previous update of the test template.

PR #1185, opened 2019-07-12
[FIX, FEATURE] Fix initialising the version_check_option. Add multi-level argument parsing.

### LINKED ISSUE #1177: Segmentation fault produced by version check on some machines.
Hi,

I am using the argument parser. On one machine the binary runs, on the other I get a segfault.

```
pocki@salz4 /ccb/salz4-2/pocki/HAPLO_RES % gdb ./kmc_unique                                                                                                                                                 :(
GNU gdb (GDB) Red Hat Enterprise Linux 7.6.1-94.el7
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-redhat-linux-gnu".
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>...
Reading symbols from /ccb/salz4-2/pocki/HAPLO_RES/kmc_unique...done.
(gdb) run
Starting program: /ccb/salz4-2/pocki/HAPLO_RES/./kmc_unique 
Detaching after fork from child process 8658.

Program received signal SIGSEGV, Segmentation fault.
0x0000000000000000 in ?? ()
(gdb) bt
#0  0x0000000000000000 in ?? ()
#1  0x00000000004d3c61 in std::thread::detach() ()
#2  0x000000000042b8da in seqan3::detail::version_checker::operator()(std::promise<bool>) ()
#3  0x0000000000435e14 in seqan3::argument_parser::parse() ()
#4  0x0000000000405f8c in main ()
```

I can fix it by turning version-check off. Interestingly the first one works, the second call produces the same segmentation fault (it seems the order of arguments matters ... shouldn't all arguments be parsed first before sending data over the internet? :thinking: ):

```
% ./kmc_unique --version-check 0 --help
***HELP PAGE***
% ./kmc_unique --help --version-check 0
[1]    9131 segmentation fault (core dumped)  ./kmc_unique --help --version-check 0
```

I include omp.h but I don't do anything fancy in my program. If I share the entire code I guess it will be hard to reproduce as it only happens on one of my machines. Do you have any idea that I could try fixing it? Where are the information stored when you first open the program and it asks you whether you want to run version checks every time? Maybe this file got messed up on that machine.

smehringer:
Resolves #1177

### DIFF
--- a/doc/tutorial/argument_parser/index.md
+++ b/doc/tutorial/argument_parser/index.md
@@ -466,6 +466,13 @@ that can serve as a copy'n'paste source for your application.
 \include doc/tutorial/argument_parser/solution6.cpp
 \endsolution
 
+# Subcommand argument parsing
+
+Many applications provide several sub programs, e.g. `git` comes with many functionalities like `git push`,
+`git pull`, `git checkout`, etc. each having their own help page.
+If you are interested in how this subcommand parsing can be done with the seqan3::argument_parser,
+take a look at our \link subcommand_arg_parse HowTo\endlink.
+
 # Update Notifications
 
 When you run a SeqAn-based application for the first time, you will likely be asked about "update notifications".
--- a/include/seqan3/argument_parser/argument_parser.hpp
+++ b/include/seqan3/argument_parser/argument_parser.hpp
@@ -553,6 +553,8 @@ private:
             return;
         }
 
+        bool special_format_was_set{false};
+
         for(int i = 1, argv_len = argc; i < argv_len; ++i) // start at 1 to skip binary name
         {
             std::string arg{argv[i]};
@@ -561,18 +563,18 @@ private:
             {
                 format = detail::format_help{false};
                 init_standard_options();
-                return;
+                special_format_was_set = true;
             }
             else if (arg == "-hh" || arg == "--advanced-help")
             {
                 format = detail::format_help{true};
                 init_standard_options();
-                return;
+                special_format_was_set = true;
             }
             else if (arg == "--version")
             {
                 format = detail::format_version{};
-                return;
+                special_format_was_set = true;
             }
             else if (arg.substr(0, 13) == "--export-help") // --export-help=man is also allowed
             {
@@ -600,12 +602,12 @@ private:
                     throw validation_failed{"Validation failed for option --export-help: "
                                             "Value must be one of [html, man]"};
                 init_standard_options();
-                return;
+                special_format_was_set = true;
             }
             else if (arg == "--copyright")
             {
                 format = detail::format_copyright{};
-                return;
+                special_format_was_set = true;
             }
             else if (arg == "--version-check")
             {
@@ -629,7 +631,8 @@ private:
             }
         }
 
-        format = detail::format_parse(argc, std::move(argv_new));
+        if (!special_format_was_set)
+            format = detail::format_parse(argc, std::move(argv_new));
     }
 
     //!\brief Adds standard options to the help page.
--- a/include/seqan3/argument_parser/argument_parser.hpp
+++ b/include/seqan3/argument_parser/argument_parser.hpp
@@ -29,6 +29,7 @@
 #include <seqan3/argument_parser/detail/version_check.hpp>
 #include <seqan3/core/char_operations/predicate.hpp>
 #include <seqan3/core/detail/terminal.hpp>
+#include <seqan3/core/detail/to_string.hpp>
 #include <seqan3/io/stream/concept.hpp>
 
 namespace seqan3
@@ -166,6 +167,7 @@ public:
      * \param[in] argc           The number of command line arguments.
      * \param[in] argv           The command line arguments to parse.
      * \param[in] version_check  Notify users about app version updates (default true).
+     * \param[in] subcommands    A list of subcommands (see \link subcommand_arg_parse subcommand parsing \endlink).
      *
      * \throws seqan3::parser_design_error if the application name contains illegal characters.
      *
@@ -178,14 +180,19 @@ public:
     argument_parser(std::string const app_name,
                     int const argc,
                     char const * const * const  argv,
-                    bool version_check = true) :
+                    bool version_check = true,
+                    std::vector<std::string> subcommands = {}) :
         version_check_dev_decision{version_check}
     {
         if (!std::regex_match(app_name, app_name_regex))
             throw parser_design_error{"The application name must only contain alpha-numeric characters "
                                                "or '_' and '-' (regex: \"^[a-zA-Z0-9_-]+$\")."};
+        for (auto & sub : subcommands)
+            if (!std::regex_match(sub, std::regex{"^[a-zA-Z0-9_]+$"}))
+                throw parser_design_error{"The subcommand name must only contain alpha-numeric characters or '_'."};
+
         info.app_name = std::move(app_name);
-        init(argc, argv);
+        init(argc, argv, std::move(subcommands));
     }
 
     //!\brief The destructor.
@@ -217,6 +224,8 @@ public:
      * \param[in]  desc      The description of the option to be shown in the help page.
      * \param[in]  spec      Advanced option specification, see seqan3::option_spec.
      * \param[in]  validator The validator applied to the value after parsing (callable).
+     *
+     * \throws seqan3::parser_design_error
      */
     template <typename option_type, validator validator_type = detail::default_validator<option_type>>
     //!\cond
@@ -231,6 +240,9 @@ public:
                     option_spec const & spec = option_spec::DEFAULT,
                     validator_type validator = validator_type{}) // copy to bind rvalues
     {
+        if (sub_parser != nullptr)
+            throw parser_design_error{"You may only specify flags for the top-level parser."};
+
         verify_identifiers(short_id, long_id);
         // copy variables into the lambda because the calls are pushed to a stack
         // and the references would go out of scope.
@@ -287,6 +299,9 @@ public:
                                std::string const & desc,
                                validator_type validator = validator_type{}) // copy to bind rvalues
     {
+        if (sub_parser != nullptr)
+            throw parser_design_error{"You may only specify flags for the top-level parser."};
+
         if (has_positional_list_option)
             throw parser_design_error{"You added a positional option with a list value before so you cannot add "
                                       "any other positional options."};
@@ -387,6 +402,19 @@ public:
 
         std::visit([this] (auto & f) { f.parse(info); }, format);
         parse_was_called = true;
+    }
+
+    //!\brief Returns a reference to the sub-parser instance if
+    //!       \link subcommand_arg_parse subcommand parsing \endlink was enabled.
+    argument_parser & get_sub_parser()
+    {
+        if (sub_parser == nullptr)
+        {
+            throw parser_design_error("You did not enable subcommand parsing on construction "
+                                      "so you cannot access the sub-parser!");
+        }
+
+        return *sub_parser;
     }
 
     //!\name Structuring the Help Page
@@ -511,13 +539,17 @@ private:
     //!\brief The future object that keeps track of the detached version check call thread.
     std::future<bool> version_check_future;
 
-    //!\brief Validates the application name to ensure an escaped server call
+    //!\brief Validates the application name to ensure an escaped server call.
     std::regex app_name_regex{"^[a-zA-Z0-9_-]+$"};
+
+    //!\brief Stores the sub-parser in case \link subcommand_arg_parse subcommand parsing \endlink is enabled.
+    std::unique_ptr<argument_parser> sub_parser{nullptr};
 
     /*!\brief Initializes the seqan3::argument_parser class on construction.
      *
-     * \param[in] argc     The number of command line arguments.
-     * \param[in] argv     The command line arguments.
+     * \param[in] argc        The number of command line arguments.
+     * \param[in] argv        The command line arguments.
+     * \param[in] subcommands The subcommand key words to split command line arguments into top-level and sub-parser.
      *
      * \throws seqan3::parser_invalid_argument
      *
@@ -542,7 +574,7 @@ private:
      * If `--export-help` is specified with a value other than html/man or ctd
      * a parser_invalid_argument is thrown.
      */
-    void init(int argc, char const * const * const argv)
+    void init(int argc, char const * const * const argv, std::vector<std::string> const & subcommands)
     {
         // cash command line input, in case --version-check is specified but shall not be passed to format_parse()
         std::vector<std::string> argv_new{};
@@ -559,15 +591,20 @@ private:
         {
             std::string arg{argv[i]};
 
+            if (std::ranges::find(subcommands, arg) != subcommands.end())
+            {
+                sub_parser = std::make_unique<argument_parser>(info.app_name + "-" + arg, argc - i, argv + i, false);
+                break;
+            }
             if (arg == "-h" || arg == "--help")
             {
-                format = detail::format_help{false};
+                format = detail::format_help{subcommands, false};
                 init_standard_options();
                 special_format_was_set = true;
             }
             else if (arg == "-hh" || arg == "--advanced-help")
             {
-                format = detail::format_help{true};
+                format = detail::format_help{subcommands, true};
                 init_standard_options();
                 special_format_was_set = true;
             }
@@ -592,9 +629,9 @@ private:
                 }
 
                 if (export_format == "html")
-                    format = detail::format_html{};
+                    format = detail::format_html{subcommands};
                 else if (export_format == "man")
-                    format = detail::format_man{};
+                    format = detail::format_man{subcommands};
                 // TODO (smehringer) use when CTD support is available
                 // else if (export_format == "ctd")
                 //     format = detail::format_ctd{};
@@ -632,7 +669,15 @@ private:
         }
 
         if (!special_format_was_set)
+        {
+            if (!subcommands.empty() && sub_parser == nullptr)
+            {
+                throw parser_invalid_argument{detail::to_string("Please specify which sub program you want to use ",
+                                              "(one of ", subcommands, "). Use -h/--help for more information.")};
+            }
+
             format = detail::format_parse(argc, std::move(argv_new));
+        }
     }
 
     //!\brief Adds standard options to the help page.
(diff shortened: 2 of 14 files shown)


include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -551,18 +553,18 @@ class argument_parser
             {
                 format = detail::format_help{false};
                 init_standard_options();
-                return;
+                special_format_was_set = true;
```
cpockrandt: It seems that if you pass multiple special formats (e.g., h and hh), it will overwrite it in the next loop iteration. I think we should either print an error message or always take the first one. Currently it would choose the last argument.

h-2:
How does that work with nested argument parsers? For nested arg parser the version check argument should refer to the nested inner arg parser if it comes after the keyowrd. I suspect these changes will break that.

We really need test-cases for nested arg parsers as it is a commonly used design.

rrahn:
Sorry to jump into this but what is a nested argument parser? Do we have a snippet or demo/tutorial for this? I can only think of a tool with different configurations who have their own options, e.g.samtools.

h-2:
> Sorry to jump into this but what is a nested argument parser? 

A program with different sub-programs, like `git` or `lambda` (equally famous tools ☝️ ). Each sub-program should have their own help/man-page et cetera...

> Do we have a snippet or demo/tutorial for this?

@smehringer said she would write a HowTo at some point (or at least we agreed that we want such a HowTo). Currently we have nothing that tests this, that's why I said new changes threaten to break it.

rrahn:
Ok, this is what I had in mind already.

doc/howto/multi_level_argument_parser/multi_level_arg_parse.cpp
```diff
@@ -0,0 +1,97 @@
...
+    return 0;
+}
+
+int main(int argc, char const ** argv)
+{
+    //![construction]
+    argument_parser top_level_parser{"mygit", argc, argv, true, {"push", "pull"}};
+    //![construction]
+
+    // Add information and options to your top level parser just as you would do with a normal one.
```
eseiler:
```suggestion
    // Add information and options to your top-level parser just as you would do with a normal one.
```

doc/tutorial/argument_parser/index.md
```diff
@@ -466,6 +466,13 @@ that can serve as a copy'n'paste source for your application.
 \include doc/tutorial/argument_parser/solution6.cpp
 \endsolution
 
+# Multi-level argument parsing
+
+Many applications provide several sub programs, e.g. `git` comes with many functionalities like `git push`,
+`git pull`, `git checkout`, etc. each having their own help page.
+If you are interested in how this multi-level parsing can be done with the seqan3::argument_parser,
+take a look at our \link multi_level_arg_parse How-to\endlink.
```
eseiler:
```suggestion
take a look at our \link multi_level_arg_parse HowTo\endlink.
```

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -389,6 +392,16 @@ class argument_parser
         parse_was_called = true;
     }
 
+    //!\brief Returns a reference to the sub-parser instance if
+    //!       \link multi_level_arg_parse multi-level parsing \endlink was enabled.
+    argument_parser & get_sub_parser()
+    {
+        if (sub_parser == nullptr)
+            throw parser_design_error("You did not enable multi-level parsing on construction "
```
eseiler: Braces?

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -511,13 +524,17 @@ class argument_parser
     //!\brief The future object that keeps track of the detached version check call thread.
     std::future<bool> version_check_future;
 
-    //!\brief Validates the application name to ensure an escaped server call
+    //!\brief Validates the application name to ensure an escaped server call.
     std::regex app_name_regex{"^[a-zA-Z0-9_-]+$"};
 
+    //!\brief Stores the sub-parser in case multi level parsing is enabled on construction.
```
eseiler:
```suggestion
    //!\brief Stores the sub-parser in case \link multi_level_arg_parse multi-level parsing \endlink is enabled on construction.
```

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -553,26 +570,38 @@ class argument_parser
...
         }
 
+        bool special_format_was_set{false};
+        bool sub_parser_was_set{false};
+
         for(int i = 1, argv_len = argc; i < argv_len; ++i) // start at 1 to skip binary name
         {
             std::string arg{argv[i]};
 
+            if (std::find(subcommands.begin(), subcommands.end(), arg) != subcommands.end())
```
eseiler:
```suggestion
            if (std::ranges::find(subcommands, arg) != subcommands.end())
```

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -553,26 +570,38 @@ class argument_parser
...
+            {
+                sub_parser = std::unique_ptr<argument_parser>{new argument_parser{info.app_name + "-" + arg,
```
eseiler: Why not

```cpp
sub_parser = std::make_unique<argument_parser>(info.app_name + "-" + arg, argc - i, argv + i, false);
```

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -629,7 +658,14 @@ class argument_parser
             }
         }
 
-        format = detail::format_parse(argc, std::move(argv_new));
+        if (!special_format_was_set)
+        {
+            if (!subcommands.empty() && !sub_parser_was_set)
+                throw parser_invalid_argument{detail::to_string("Please specify which sub program you want to use ",
```
eseiler: braces?

include/seqan3/argument_parser/detail/format_base.hpp
```diff
@@ -325,6 +326,19 @@ class format_help_base : public format_base
                 print_line(desc);
         }
 
+        if (!command_names.empty())
+        {
+            derived_t().print_section("Sub Commands");
```
eseiler:
```suggestion
            derived_t().print_section("Subcommands");
```

include/seqan3/argument_parser/detail/format_base.hpp
```diff
@@ -325,6 +326,19 @@ class format_help_base : public format_base
...
+            derived_t().print_line("This program must be invoked with one of the following subcommands:", false);
+            for (std::string & name : command_names)
```
eseiler:
```suggestion
            for (std::string const & name : command_names)
```

test/unit/argument_parser/format_parse_test.cpp
```diff
@@ -804,3 +804,85 @@ TEST(parse_test, version_check_option_error)
...
+
+        EXPECT_EQ(sub_parser.info.app_name, "top_level-sub1");
+
+        sub_parser.add_positional_option(option_value2, "foo bar");
+
+        EXPECT_NO_THROW(sub_parser.parse());
+        EXPECT_EQ("foo", option_value2);
+    }
+
+    // top level help page
```
eseiler:
```suggestion
    // top-level help page
```

test/unit/argument_parser/format_parse_test.cpp
```diff
@@ -804,3 +804,85 @@ TEST(parse_test, version_check_option_error)
...
+        const char * argv[]{"./top_level", "-h", "2", "sub1", "foo"};
+        argument_parser top_level_parser{"top_level", 5, argv, false, {"sub1", "sub2"}};
+        top_level_parser.add_option(option_value, 'f', "foo", "foo bar");
+
+        testing::internal::CaptureStdout();
+        EXPECT_EXIT(top_level_parser.parse(), ::testing::ExitedWithCode(EXIT_SUCCESS), "");
+        EXPECT_FALSE(std::string{testing::internal::GetCapturedStdout()}.empty());
+    }
+
+    // sub parser help page
```
eseiler:
```suggestion
    // sub-parser help page
```

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -511,13 +527,17 @@ class argument_parser
     //!\brief The future object that keeps track of the detached version check call thread.
     std::future<bool> version_check_future;
 
-    //!\brief Validates the application name to ensure an escaped server call
+    //!\brief Validates the application name to ensure an escaped server call.
     std::regex app_name_regex{"^[a-zA-Z0-9_-]+$"};
 
+    //!\brief Stores the sub-parser in case \link multi_level_arg_parse multi-level parsing \endlink is enabled on construction.
```
eseiler: Line break

smehringer:
Code coverage complains that format man is not tested :roll_eyes:

doc/howto/multi_level_argument_parser/multi_level_arg_parse.cpp
```diff
@@ -0,0 +1,97 @@
...
+    }
+    catch (parser_invalid_argument const & ext)
+    {
+        debug_stream << "[Error] " << ext.what() << "\n";
+        return -1;
+    }
+
+    return 0;
+}
+
```
h-2:
```suggestion
// =====================================================================================================================
// main
// =====================================================================================================================
```

If you use these in the code, add one before main

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -178,14 +180,15 @@ class argument_parser
...
+                    bool version_check = true,
+                    std::vector<std::string> subcommands = {}) :
         version_check_dev_decision{version_check}
     {
         if (!std::regex_match(app_name, app_name_regex))
             throw parser_design_error{"The application name must only contain alpha-numeric characters "
                                                "or '_' and '-' (regex: \"^[a-zA-Z0-9_-]+$\")."};
         info.app_name = std::move(app_name);
-        init(argc, argv);
+        init(argc, argv, subcommands);
```
h-2:
```suggestion
        init(argc, argv, std::move(subcommands));
```

:nail_care:

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -553,26 +573,35 @@ class argument_parser
             return;
         }
 
+        bool special_format_was_set{false};
+        bool sub_parser_was_set{false};
```
h-2: Why is this needed? Can't you check if  `unique_ptr != nullptr` ?

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -553,26 +573,35 @@ class argument_parser
...
+
         for(int i = 1, argv_len = argc; i < argv_len; ++i) // start at 1 to skip binary name
         {
             std::string arg{argv[i]};
 
+            if (std::ranges::find(subcommands, arg) != subcommands.end())
+            {
+                sub_parser = std::make_unique<argument_parser>(info.app_name + "-" + arg, argc - i, argv + i, false);
+                sub_parser_was_set = true;
```
h-2: I think this won't work in the following case:

```
git --string-option pull --flag bar push --foo
```
In this case it will identify the subcommand as pull, although it is push, or not?

h-2: Why don't you call parse on the the sub_parser here and return after that? Then you don't need `special_format_was_set` and the extra handling at the end.

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -705,7 +743,7 @@ class argument_parser
                  detail::format_html,
                  detail::format_man,
                  detail::format_copyright/*,
-                 detail::format_ctd*/> format{detail::format_help(0)};
+                 detail::format_ctd*/> format{detail::format_help{{}, false}};
```
h-2: Does this need to be initialised?

include/seqan3/argument_parser/argument_parser.hpp
smehringer: left over from previous solutions :see_no_evil:

include/seqan3/argument_parser/argument_parser.hpp
smehringer: > I think this won't work in the following case:

Yes you are right but I do not see how I can avoid that?

> Why don't you call parse on the the sub_parser here and return after that?

Because the sub parser is not set up yet by the user with all its `add_option` calls yet?

include/seqan3/argument_parser/argument_parser.hpp
smehringer: When the constructor is called the format would be default initialised which is not possible because format_parse has no default constructor..?

include/seqan3/argument_parser/argument_parser.hpp
h-2: * forbid options and positionals in the top-level parser if it has subcommends
* forbid subcommands that start with `-` or consist only of numbers

smehringer: If we do not allow any options anyway we don't need to restrict the command to "may not have only numbers"?

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -553,26 +573,33 @@ class argument_parser
             return;
         }
 
+        bool special_format_was_set{false};
```
h-2: Still don't understand why you need this. Just return as before.

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -553,26 +573,33 @@ class argument_parser
...
+
         for(int i = 1, argv_len = argc; i < argv_len; ++i) // start at 1 to skip binary name
         {
             std::string arg{argv[i]};
 
+            if (std::ranges::find(subcommands, arg) != subcommands.end())
+            {
+                sub_parser = std::make_unique<argument_parser>(info.app_name + "-" + arg, argc - i, argv + i, false);
+                break;
```
h-2: return here.

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -629,7 +656,16 @@ class argument_parser
             }
         }
 
-        format = detail::format_parse(argc, std::move(argv_new));
+        if (!special_format_was_set)
+        {
+            if (!subcommands.empty() && sub_parser == nullptr)
```
h-2:
```suggestion
            if (!subcommands.empty())
```
And remove the check above.

Then this implies that you only reach this point if there was no special format and no subcommand was set.

include/seqan3/argument_parser/argument_parser.hpp
h-2: Well, it's weird because it suggests that it has some kind of special meaning. 

Why not use std::monostate which was created for this purpose?
https://en.cppreference.com/w/cpp/utility/variant/monostate

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -166,6 +167,7 @@ class argument_parser
      * \param[in] argc           The number of command line arguments.
      * \param[in] argv           The command line arguments to parse.
      * \param[in] version_check  Notify users about app version updates (default true).
+     * \param[in] subcommands    A list of sub-programs (see \link subcommand_arg_parse subcommand parsing \endlink).
```
h-2: Why do you use the term "sub-program" here when you use subcommand everywhere else?

include/seqan3/argument_parser/detail/format_base.hpp
```diff
@@ -214,10 +214,11 @@ class format_help_base : public format_base
     ~format_help_base() = default;                                       //!< Defaulted.
 
     /*!\brief Initializes a format_help_base object.
+     * \param[in] names    A list of sub-program names for subcommand argument parsing.
```
h-2: here as well

include/seqan3/argument_parser/argument_parser.hpp
smehringer: I need to scan the whole argv in order to strip the version check option from the argv if present at any position. I cannot just return at any place.

include/seqan3/argument_parser/argument_parser.hpp
smehringer: If I return here I need to copy and paste the same if check as comes after the for loop into the current if scope. It is more dry like this.

include/seqan3/argument_parser/argument_parser.hpp
h-2: Hm, it's ugly, but oh well.

include/seqan3/argument_parser/argument_parser.hpp
smehringer: uh I did not know of this :+1:

smehringer: too bad... If I use std::monostate the std::visit call fails because monostate obviously has no `add_flag` member function :/

include/seqan3/argument_parser/argument_parser.hpp
h-2: I don't core.

include/seqan3/argument_parser/argument_parser.hpp
h-2: Hm, than maybe at least write something like "The state will be overwritten in any case." as comment?

h-2: Still open.

smehringer:
why did you remove your review? :scream_cat: 
(I already sais the code coverage fails because of the untested man format, if this is the case)

include/seqan3/argument_parser/argument_parser.hpp
smehringer: ok fine

h-2:
I remove my review request not because I am pissed, just so that I know which of the items in the list is actually waiting for me and not me waiting on the other person :wink:

smehringer:
I know, I just thought it is because of the code coverage that bites me any time I do something on the argument parser :innocent:

smehringer:
[issue](https://trello.com/c/zZTGUtVr/4-issue)

PR #1186, opened 2019-07-12
[FEATURE] Adds aligned sequence builder

rrahn:
This PR depends on #1158.

Adds an aligned sequence builder to create the actual alignment from a given trace path.
For this PR only the last commit is relevant.


test/unit/alignment/matrix/detail/aligned_sequence_builder_test.cpp
```diff
@@ -0,0 +1,300 @@
...
+TYPED_TEST(aligned_sequence_builder_fixture, build_from_2_3)
+{
+    auto p = this->path(matrix_offset{row_index_type{2}, column_index_type{3}});
+    auto [begin, end, alignment] = this->builder.build_from(p);
+
+    EXPECT_EQ(begin.row, 0u);
+    EXPECT_EQ(begin.col, 0u);
+    EXPECT_EQ(end.row, 2u);
+    EXPECT_EQ(end.col, 3u);
+    EXPECT_EQ(alignment.first  | view::to_char | std::ranges::to<std::string>, std::string{"--ACG"});
```
eseiler: Mhm, I thought you'd get `L,D,D` from the trace back, and the result would be
```
ACG
AG-
```

test/unit/alignment/matrix/detail/aligned_sequence_builder_test.cpp
```diff
@@ -0,0 +1,300 @@
...
+TYPED_TEST(aligned_sequence_builder_fixture, build_from_1_2)
+{
+    auto p = this->path(matrix_offset{row_index_type{1}, column_index_type{2}});
+    auto [begin, end, alignment] = this->builder.build_from(p);
+
+    EXPECT_EQ(begin.row, 0u);
+    EXPECT_EQ(begin.col, 0u);
+    EXPECT_EQ(end.row, 1u);
+    EXPECT_EQ(end.col, 2u);
+    EXPECT_EQ(alignment.first  | view::to_char | std::ranges::to<std::string>, std::string{"-AC"});
```
eseiler: And here `L,D`
```
AC
A-
```

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
```diff
@@ -0,0 +1,249 @@
...
+        matrix_coordinate path_end = trace_it.coordinate();
+
+        std::vector<std::pair<trace_directions, size_t>> traces;
+
+        while (trace_it != std::ranges::end(trace_path))
+        {
+            trace_directions last_dir = *trace_it;
+            size_t span = 0;
+            for (; trace_it != std::ranges::end(trace_path) && *trace_it == last_dir; ++trace_it, ++span)
+                ;
```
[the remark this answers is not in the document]
rrahn: You need to put a semicolon at the end of the for-loop. Some compilers may warn if this is not on a new line. At least I read about it. The purpose is for the reader to understand that this for loop has no body on purpose.

test/unit/alignment/matrix/detail/aligned_sequence_builder_test.cpp
```diff
@@ -0,0 +1,300 @@
...
+    auto p = this->path(matrix_offset{row_index_type{0}, column_index_type{3}});
+    auto [begin, end, alignment] = builder.build_from(p);
+
+    EXPECT_EQ(begin.row, 0u);
+    EXPECT_EQ(begin.col, 0u);
+    EXPECT_EQ(end.row, 0u);
+    EXPECT_EQ(end.col, 3u);
+    EXPECT_EQ(alignment.first  | view::to_char | std::ranges::to<std::string>, std::string{"ACG"});
+    EXPECT_EQ(alignment.second | view::to_char | std::ranges::to<std::string>, std::string{"---"});
+}
```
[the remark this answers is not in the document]
rrahn: I'll check

test/unit/alignment/matrix/detail/aligned_sequence_builder_test.cpp
rrahn: Nop, it models affine gaps. That means you need to follow a gap until you find the corresponding gap_open signal. With that respect linear gaps are a specialisation of affine gaps in which every gap is a gap open. Thus we can use the same trace algorithm for affine and linear gaps.

test/unit/alignment/matrix/detail/aligned_sequence_builder_test.cpp
rrahn: Like above

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
```diff
@@ -0,0 +1,249 @@
...
+ *
+ * \details
+ *
+ * This class builds the alignment from a given trace path over the specified sequences. Use the interface
+ * seqan3::detail::aligned_sequence_builder::build_from to get the alignment. The returned seqan3::AlignedSequence type
+ * is determined by the `traits_t` class which needs two member types that represent the seqan::AlignedSequence type.
+ * The specified types must be constructible from the original type refined with a seqan3::view::slice. So the following
+ * expressions must evaluate to `true` in order use the sequences with the given aligned sequence types:
+ * ```cpp
+ * std::Constructible<typename traits_t::fst_aligned_t, decltype(std::declval<fst_rng_t>() | view::slice(0,0))>
```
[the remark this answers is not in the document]
rrahn: Yes, it doesn't matter since it is only evaluated at compile time  but `(0, 1)` sounds better. I'll change this.

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
```diff
@@ -0,0 +1,249 @@
...
+    //!\endcond
+    static auto aligned_type(rng_t *) { return gap_decorator<rng_t>{}; }
+
+    //!\overload
+    template <typename rng_t>
+    static auto aligned_type(rng_t *) { return std::vector<gapped<value_type_t<rng_t>>>{}; }
+
+public:
+    //!\brief The aligned sequence type for the first sequence.
+    using fst_aligned_t = decltype(aligned_type(static_cast<fst_slice_t *>(nullptr)));
```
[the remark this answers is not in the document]
rrahn: Well, it I used a similar approach as used in ranges library. declval would also be ok since it is in an unevaluated context but I guess I revert it back to just pass the types as template arguments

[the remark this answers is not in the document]
eseiler: You can, but you don't need to.

test/unit/alignment/matrix/detail/aligned_sequence_builder_test.cpp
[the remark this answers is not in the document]
rrahn: thanks for spotting. I changed the editor and it was not configured to add final new line automatically.

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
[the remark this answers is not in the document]
h-2: 💅 🙏 :  Please just put `{}` there instead, it's already used for empty class bodies, empty function bodies ... and looks a lot more intentional.

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
```diff
@@ -0,0 +1,248 @@
...
+#include <seqan3/range/view/view_all.hpp>
+#include <seqan3/std/concepts>
+#include <seqan3/std/ranges>
+
+namespace seqan3::detail
+{
+
+/*!\brief Transforms the given range types into their aligned sequence format.
+ * \ingroup alignment_matrix
+ * \tparam fst_rng_t The type of the first range to be transformed to a aligned sequence type.
```
eseiler: Add `; must model std::ranges::ViewableRange.` ?
(2 times)

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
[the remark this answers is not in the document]
rrahn: 😥

rrahn:
@smehringer only the last commit is relevant

rrahn:
@smehringer ping (and remember only the last commit)

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
```diff
@@ -0,0 +1,250 @@
...
+ * \ingroup alignment_matrix
+ * \tparam fst_rng_t The type of the first range to be transformed to a aligned sequence type;
+ *                   must model std::ranges::ViewableRange.
+ * \tparam sec_rng_t The type of the second range to be transformed to a aligned sequence type;
+ *                   must model std::ranges::ViewableRange.
+ *
+ * \details
+ *
+ * First this trait refines the original range types with seqan3::view::slice. Then it transforms the given range type
+ * into a seqan3::gap_decorator if the expression is not ill-formed, otherwise the transformed aligned sequence type
```
smehringer: in which cases is the gap decorator ill-formed?

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
```diff
@@ -0,0 +1,250 @@
...
+ * expressions must evaluate to `true` in order use the sequences with the given aligned sequence types:
+ * ```cpp
+ * std::Constructible<typename traits_t::fst_aligned_t, decltype(std::declval<fst_rng_t>() | view::slice(0, 1))>
+ * std::Constructible<typename traits_t::sec_aligned_t, decltype(std::declval<sec_rng_t>() | view::slice(0, 1))>
+ * ```
+ */
+template <std::ranges::ViewableRange fst_rng_t,
+          std::ranges::ViewableRange sec_rng_t,
+          typename traits_t = aligned_sequence_builder_default_traits<fst_rng_t, sec_rng_t>>
+class aligned_sequence_builder
```
smehringer: What is the advantage of a class instead of a single free function `build_alignment(seq1, seq2, trace)`?

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
```diff
@@ -0,0 +1,250 @@
...
+    constexpr aligned_sequence_builder & operator=(aligned_sequence_builder const &) = default; //!< Defaulted.
+    constexpr aligned_sequence_builder & operator=(aligned_sequence_builder &&) = default; //!< Defaulted.
+    ~aligned_sequence_builder() = default; //!< Defaulted.
+
+    /*\brief Construction from the underlying sequences
+     * \param[in] fst_rng The first range to build the aligned sequence for.
+     * \param[in] sec_rng The second range to build the aligned sequence for.
+     */
+    constexpr aligned_sequence_builder(fst_rng_t fst_rng, sec_rng_t sec_rng) :
+        fst_rng{view::all(std::forward<fst_rng_t>(fst_rng))},
```
smehringer: why do you store views to the ranges, instead of passing them to `build_from`?

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
```diff
@@ -0,0 +1,250 @@
...
+        sec_rng{view::all(std::forward<sec_rng_t>(sec_rng))}
+    {}
+    //!\}
+
+    /*!\brief Builds the aligned sequences from the given trace path
```
smehringer:
```suggestion
    /*!\brief Builds the aligned sequences from the given trace path.
```

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
```diff
@@ -0,0 +1,250 @@
...
+    template <std::ranges::InputRange trace_path_t>
+    result_type build_from(trace_path_t && trace_path)
+    {
+        static_assert(std::Same<value_type_t<trace_path_t>, trace_directions>,
+                      "The value type of the trace path must be seqan3::detail::trace_directions");
+
+        auto trace_it = std::ranges::begin(trace_path);
+        matrix_coordinate path_end = trace_it.coordinate();
+
+        std::vector<std::pair<trace_directions, size_t>> traces;
```
smehringer: why do you fill this vector instead of directly going over the path and adding gaps? I am not familiar with the trace path but I imagine it to be something like `{up, diag, left, left, diag}`.  So why iterating over it twice and filling a vector instead of going along the trace until it changes and inserting the respective gaps directly? (It should still be done in the separate function of course.)

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
rrahn: Whenever you have non sized or non random access range. That clearly depends on the gap dectorator

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
rrahn: This is an interface strongly coupled with the algorithms and the other classes I added before so there is no need for providing an algorithm like interface. And there will be versions in the future that will maintain a stack so that you can actually parse all possible tracebacks and not just one of the optimal paths as it is right now. So the algorithm is not stateless and therefor the class design is much more flexible.

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
rrahn: So I can build over the same sequences different paths, e.g. when searching for sub optimal alignments or reporting all paths with the same score.

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
rrahn: because you need to parse the trace from right to left (from sink to source in matrix view). 
Therefor, whenever you add something in that order you need to update all the virtual positions to the right. Depeding on the underlying DS, this can be quadratic in runtime, whereas this implementation is linear by first getting all trace segments and then applying them from right to left.

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
smehringer: ugh we have to change that :unamused:

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
smehringer: >  And there will be versions in the future that will maintain a stack so that you can actually parse all possible tracebacks and not just one of the optimal paths as it is right now.

Hm I see the use case fo suboptimal paths, but don't you need to stack the trace paths, not the aligned_sequence_builder instances?

> So the algorithm is not stateless

What is the state? because the sequences are passed along anyway within the alignment algorithm (as far as I recall from previous reviews)

> the class design is much more flexible.
 
Can you give me an example where a class design storing the sequences is more flexible than a free function in detail? I am not able to see this as clearly as you do.

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
smehringer: but whats better in this 
```cpp
sequence_builder sb{seq1, seq2};
for (auto & trace : all_traces_of_interest)
    sb.build_from(trace);
```
compared to this
```cpp
for (auto & trace : all_traces_of_interest)
    build_alignment(seq1, seq2, trace);
```

The only thing I can see is that you would need to pass one variable (sequence_builder) to subsequent functions, instead of two (the two sequences.)?

I mean If you say all of this fits much better in your alignment design I'm fine with this, but I try to start keeping track of general design decisions to learn and apply them in the future. And I would have designed this as a free detail function and I am not convinced yet about the advantages :grin: help me out.

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
smehringer: But the current DS is the gap_decorator in most cases correct (Because why should you use a std::list insead of a std::vector)? And for the gap_decorator, it does not matter performance-wise if you add gaps at the front or back (red-black-tree). So now you add extra overhead that is not needed right now and makes the code slower because you expect that in the future this might be different? What happened to incremental refactoring? And maybe we should always use the most efficient way possible given a certain DS

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
rrahn: Well, the sort answer would be try to never use free functions if there is not a real good reason for it (like interface functions that model a concept for types that might not be able to have member functions, or algorithms that are not coupled to the data). Free functions is the way we did things in SeqAn2 out of a design decision. Now we want to have more encapsulation. Something that classes can provide not free functions and will give much more flexibility. In this case the data is strongly coupled with the methods acting on it.  

> Hm I see the use case fo suboptimal paths, but don't you need to stack the trace paths, not the aligned_sequence_builder instances?

Right! And how would you do it? You parse the trace in a DFS order. Now, you can either provide all paths at once because you have an algorithm (since we do not use coroutines yet) or you can do it on demand because you can store the state (which is btw. our general design for SeqAn3 - doing things lazy). Now, I wasn't sure yet if I need a range interface for this but this can be simply added in the future if required. But again, I don't see how this is a function exposed to the public.

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
rrahn: Ok, I just looked into the implementation and it seems that insert is always linear. Also note, that the documentation is wrong about this. So no matter which version it is always quadratic 😭.
So once the insert is implemented efficiently to find the location in the set the current version is as fast as if I insert from right to left. But note that if the requirements for a gap decorator are not fulfilled a std::vector<gapped> is used. In this case inserting from left to right is also better than inserting from right to left.

rrahn: **EDIT**: I just got confused about your comments. In general the current version is always at least as good as the insertion from left to right into the underlying DS. 
a: Either we store the Virtual positions directly on the elements (as it is right now). In this case you always need to scan all nodes right of it to update the VP. So this is always quadratic, since I have n inserts and n updates per insert, while in my case the algorithm is linear.
b: If you would store only an offset on the inner nodes to be added for the right subtree you need n*log(n) time for n elements to insert. The same for my version. linear scan and then log(n) per node to insert.

PR #1187, opened 2019-07-12
[DOCS] Search docs

### LINKED ISSUE #1142: Search module issues
## Documentation

The only thing regarding result_type I could find was:

```
An object modelling std::ranges::Range containing the hits (the type depends on the specification in cfg), or void if an on_hit delegate has been specified.
```

Well, what is a hit? And what does it mean that it depends on a config? Neither the documentation for module or sub-module cover this, nor the documentation of search() or search_cfg. Also the API documentation for search stuff in general is very scarce.

eseiler:
Resolves #1142 except the search result struct part.

The `\returns` docs will change again after we introduce the result struct.

`search()` now looks like this:
Screenshot 2019-07-13 at 01 11 16


include/seqan3/search/algorithm/search.hpp
```diff
@@ -28,14 +28,41 @@ namespace seqan3
...
  * \param[in] index   String index to be searched.
  * \param[in] cfg     A configuration object specifying the search parameters (e.g. number of errors, error types,
  *                    output format, etc.).
- * \returns An object modelling std::ranges::Range containing the hits (the type depends on the specification
-            in `cfg`), or `void` if an on_hit delegate has been specified.
+ * \returns
+ * One object of the following type if one query was passed and a std::vector containing objects of the following type
+ * for each query if a range of queries was passed:
+ * <ul>
+ *     <li>If seqan3::search_cfg::index_cursor was configured:
```
smehringer:
```suggestion
 *     <li>If seqan3::search_cfg::index_cursor was selected in the configuration:
```

include/seqan3/search/algorithm/search.hpp
```diff
@@ -28,14 +28,41 @@ namespace seqan3
...
+ *     <ul>
+ *         <li>A `std::vector<typename index_t::cursor_type>` containing index_cursors at the text positions were the
+ *             search was successful.</li>
+ *     </ul>
+ *     </li>
+ *     <li>If seqan3::search_cfg::text_position was configured:
```
smehringer:
```suggestion
 *     <li>If seqan3::search_cfg::text_position was selected in the configuration:
```

include/seqan3/search/algorithm/search.hpp
```diff
@@ -28,14 +28,41 @@ namespace seqan3
...
+ *     <ul>
+ *         <li>For **single** text indices: A `std::vector<size_t>` representing text positions.</li>
```
smehringer:
```suggestion
 *         <li>For **single** text indices: A `std::vector<size_t>` representing text positions where the search was successful.</li>
```

include/seqan3/search/algorithm/search.hpp
```diff
@@ -28,14 +28,41 @@ namespace seqan3
...
+ *         <li>For indices over text **collections**: A `std::vector<std::vector<std::pair<size_t, size_t>>>` were the
+ *             first element specifies the text index in the collection and the second element contains the position
```
smehringer: why a vector of a vector? not just a vector? And the sentence after that is confusing. 
So is there a vector for each index with sequence id and position id?

include/seqan3/search/algorithm/search.hpp
```diff
@@ -55,6 +82,10 @@ template <FmIndex index_t, typename queries_t, typename configuration_t>
 //!\endcond
 inline auto search(queries_t && queries, index_t const & index, configuration_t const & cfg)
 {
+    static_assert(std::ranges::SizedRange<queries_t>, "The query sequence/range must model SizedRange.");
+    static_assert(dimension_v<queries_t> == 1u || std::ranges::SizedRange<value_type_t<queries_t>>,
+                  "The elements of the range of queries must model SizedRange.");
+
```
smehringer: If the stuff required in the requires clause is also just for security and not for overload resolution you might want to directly change this if you are touching this code anyway

include/seqan3/search/configuration/max_error_common.hpp
```diff
@@ -20,8 +20,8 @@ namespace seqan3::search_cfg
 {
 
 /*!\brief A strong type of underlying type `uint8_t` or `double` that represents the number or rate of total errors.
- * \tparam value_t The underlying type
  * \ingroup search_configuration
+ * \tparam value_t The underlying type
```
smehringer: add a dot after the descrption :nail_care:

include/seqan3/search/algorithm/search.hpp
```diff
@@ -28,14 +28,42 @@ namespace seqan3
...
+ *     <li>If seqan3::search_cfg::text_position was selected in the configuration:
+ *     <ul>
+ *         <li>For **single** text indices: A `std::vector<size_t>` representing text positions where the search
+ *             was successful.</li>
+ *         <li>For indices over text **collections**: A `std::vector<std::pair<size_t, size_t>>` where the
+ *             first element of the `std::pair` specifies the text index in the collection and the second element
+ *             contains the position in that text of the collection.</li>
+ *     </ul>
+ *     </li>
+ *     <li>If an on_hit delegate has been configured:
```
rrahn: In general it does look nice, but I am wondering if it would not be more readable or easier to understand if you would provide a table with two columns: input and output and 4 rows for the 4 different cases. 
I also suggest to introduce a simple aggregate type to replace the pair. So we have a strong type with named accessors.
Something like: 
```cpp
struct text_position
{
    size_t collection_index;  //!< The index pointing to the respective text in the collection.
    size_t text_index //!< The index within the respective text.
};

But the naming can be adapted if their are better namings.

Of course I am still in favor to also return this type in case you submit only a single query, which is just a special case of the collection interface. So the user only needs to worry between index_curser and text_poisition. Or maybe call it text_cursor as well? 
```

include/seqan3/search/algorithm/search.hpp
```diff
@@ -46,15 +74,29 @@ namespace seqan3
...
+    {
+        static_assert(std::ranges::RandomAccessRange<queries_t>, "The query sequence must model RandomAccessRange.");
+        static_assert(std::ranges::SizedRange<queries_t>, "The query sequence must model SizedRange.");
+    }
+    else
+    {
+        static_assert(std::ranges::ForwardRange<queries_t>, "The query range must model ForwardRange.");
+        static_assert(std::ranges::SizedRange<queries_t>, "The query range must model SizedRange.");
+        static_assert(std::ranges::RandomAccessRange<value_type_t<queries_t>>,
+                      "Elements of the query range must model RandomAccessRange.");
```
rrahn: call it maybe collection instead of range as the query itself is also a range and it could be misleading 💅

include/seqan3/search/configuration/mode.hpp
```diff
@@ -39,17 +39,20 @@ namespace seqan3::search_cfg
...
 //!\brief Configuration element to receive one best hit (with the lowest number of errors).
+//!\ingroup search_configuration
 inline detail::search_mode_best constexpr best;
 
 /*!\brief Configuration element to receive all hits with the best number of errors plus the strata value.
  *        A strong type of underlying type `uint8_t` that represents the number or errors for strata.
  *        All hits are found with the fewest numbererrors plus 'value'.
- * \tparam value_t The underlying type
  * \ingroup search_configuration
+ * \tparam value_t The underlying type
```
rrahn:
```suggestion
 * \tparam value_t The underlying type.
```

include/seqan3/search/configuration/default_configuration.hpp
```diff
@@ -0,0 +1,36 @@
...
+#include <seqan3/search/configuration/output.hpp>
+
+namespace seqan3::search_cfg
+{
+
+/*!\brief The default configuration.
+ * \ingroup search_configuration
+ */
+inline constexpr configuration default_configuration =
+ max_error{total{0},
```
rrahn: can you put the max_error configs on one line?

include/seqan3/search/algorithm/search.hpp
eseiler: | text_layout | search_cfg::output | result of search()                                                                                                                                                              |
|-------------|--------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| single      | text_position      | A std::vector representing text positions where the search was successful.                                                                                                      |
| single      | index_cursor       | A std::vector containing index_cursors at the text positions where the search was successful.                                                                                   |
| collection  | text_position      | A std::vector> where the first element of the std::pair specifies the text index in the collection and the second element contains the position in that text of the collection. |
| collection  | index_cursor       | A std::vector containing index_cursors at the text positions where the search was successful.                                                                                   |

Something like this?

eseiler: Regarding replacing the `pair`: We want to do a search result struct anyway, so no need to do API  breaking changes in this PR?

rrahn: Yeah something like the table above.

> Regarding replacing the pair: We want to do a search result struct anyway, so no need to do API breaking changes in this PR?

No not in this PR. Sorry, I meant it for future consideration, but it was not so clear.

include/seqan3/search/algorithm/search.hpp
eseiler: All four asserts or only the last two?

rrahn: Like whenver you mean the collection and not the single query

include/seqan3/search/algorithm/search.hpp
eseiler: Ok, I'll try the table....but I'd do it in html. Otherwise a line will be 500 long since we can't do linebreaks in markdown tables.

rrahn: Oh no really 😱

rrahn: https://stackoverflow.com/questions/11700487/newline-in-markdown-table suggests to just use `<br>` to force the line break. I think it is still better than the entire table in html?

eseiler: Screenshot 2019-07-18 at 14 15 06

include/seqan3/search/algorithm/search.hpp
```diff
@@ -28,14 +28,53 @@ namespace seqan3
...
+ *     <td>A `std::vector<std::pair<size_t, size_t>>` where the first element of the
+ *         `std::pair` specifies the text index in the collection and the second element contains the position in that
+ *         text of the collection.</td>
+ *   </tr>
+ *   <tr>
+ *     <td style="text-align:center">\ref seqan3::text_layout "collection"</td>
+ *     <td style="text-align:center">\ref seqan3::search_cfg::index_cursor "index_cursor"</td>
+ *     <td>A `std::vector<typename index_t::cursor_type>` containing index_cursors at the text positions where the
+ *         search was successful.</td>
+ *   </tr>
```
rrahn: And the markdown in combination with `<br>` did not work?

include/seqan3/search/algorithm/search.hpp
eseiler: > https://stackoverflow.com/questions/11700487/newline-in-markdown-table suggests to just use `<br>` to force the line break. I think it is still better than the entire table in html?

But I don't want to insert a line break in the cell content, I want to linebreak the row.
Markdown: 1 line =  1 row

include/seqan3/search/algorithm/search.hpp
eseiler: `<br>` Only inserts a linebreak in the cell content, I still have to put the whole row on one line.

include/seqan3/search/algorithm/search.hpp
rrahn: But wouldn’t the line break not automatically render the entire row larger?

eseiler: What I mean to say is that if you put a `<br>` into a markdown table, it will only enforce a line break in the row (if html is supported).

| header 1 | header 2 |
|-----------|-----------|
| content 1| very long content that is very long and would go on forever because there wont be any line break inserted unless I say so or the renderer thinks it's time for a line break|

vs

| header 1 | header 2 |
|-----------|-----------|
| content 1| very long content that is very long<br> and would go on forever because there<br> wont be any line break inserted<br> unless I say so or the renderer thinks<br> it's time for a line break|


But the row still has to be a single line, e.g. no

| header 1 | header 2 |
|-----------|-----------|
| content 1| very long content that is very long
and would go on forever because there
 wont be any line break inserted
unless I say so or the renderer thinks
 it's time for a line break|

In general. Github markdown is a bit special as that it still displays it as a table, but doxygen wouldn't.

Since I don't want lines with a length of 400 (markdown), the markdown tables are translated to html anyways and html tables currently look nicer (more padding than markdown tables and a link to something in the header doesn't have the same color as the header background), I'd like to leave it like this since we will change it once the result struct is implemented.

rrahn: ok, sorry, for bugging. Just so bad that easy things are never easy. But let's keep it then as html table. 👍

eseiler:
@rrahn ping

PR #1189, opened 2019-07-15
[INFRA] attempt to fix coverage failings on travis

marehr:
This stackoverflow post suggest that ubuntu16.04 has problems with non-absolute paths https://stackoverflow.com/questions/37675961/error-while-code-coverage-report-using-lcov

This error was fixed upstreams https://github.com/linux-test-project/lcov/commit/632c25a0d1f5e4d2f4fd5b28ce7c8b86d388c91f.
But it seem to be fixed between version 1.13 and 1.14. Ubuntu 16.04 still has 1.12 so it might fix our sporadic problem.


marehr:
yeah it does not break anything; so let's try it.

PR #1192, opened 2019-07-23
Add CIGAR string support to alignment IO

smehringer:
~~Blocked by #1194~~ see last commit, the alternative way to get the help page description does not cause the ICE, so I am not dependent on the ranges-v3 update.  

Review commit by commit and note the following:

* **regarding commit 1**: When the alingment IO was written, no cigar alphabet was present. This commit introduces the alphabet and replaces `pair<char, uint32_t>`.

* **regarding commit 2**: This commit belongs to the refactoring of the first, but it did ONLY move the get_cigar_vector code above the get_cigar_string code to make use of it. Git does a horrible job in detecting this change (where actually there is none...). So I kept it separate to see the actual differences in the former commit.

* **regarding commit 3**: We ignored hard clipping and stored soft clipping in a separate variable before because it was not needed. Now the CIGAR string should represent the exact input, with hard clipping, so the functions were adapted to return the complete cigar vector.

* **regarding commit 4**: The parse_cigar/parse_binary_cigar functions were format specific and should not be free functions but member functions.

* **regarding commit 5**: Support reading of CIGAR fields.

* **regarding commit 6**: Support writing of CIGAR fields. Note that for BAM, more elaborate changes were needed to compute the ref_length from the CIGAR string (formerly obtained by simply querying the alignment length) because BAM writes out `bin` information for the BAM index that needs the ref_length information.


include/seqan3/io/alignment_file/detail.hpp
```diff
@@ -574,37 +579,37 @@ template <TupleLike alignment_type>
     requires std::tuple_size_v<remove_cvref_t<alignment_type>> == 2 &&
              detail::all_satisfy_aligned_seq<detail::tuple_type_list_t<alignment_type>>
 //!\endcond
-void alignment_from_cigar(alignment_type & alignment, std::vector<std::pair<char, size_t>> const & cigar)
+void alignment_from_cigar(alignment_type & alignment, std::vector<cigar> const & cigar_vector)
```
joergi-w: Please adapt doxygen's `\param[in]` statement to the renamed cigar_vector parameter.

include/seqan3/io/alignment_file/detail.hpp
```diff
@@ -574,37 +579,37 @@ template <TupleLike alignment_type>
...
-    for (auto [cigar_op, cigar_count] : cigar)
+    for (auto [cigar_count, op] : cigar_vector)
     {
-        assert(is_char<'M'>(cigar_op) || is_char<'='>(cigar_op) || is_char<'X'>(cigar_op) || is_char<'D'>(cigar_op) ||
-               is_char<'N'>(cigar_op) || is_char<'I'>(cigar_op) || is_char<'P'>(cigar_op)); // checked during IO
+        assert(('M'_cigar_op == op) || ('='_cigar_op == op) || ('X'_cigar_op == op) || ('D'_cigar_op == op) ||
+               ('N'_cigar_op == op) || ('I'_cigar_op == op) || ('P'_cigar_op == op)); // checked during IO
 
-        if (is_char<'M'>(cigar_op) || is_char<'='>(cigar_op) || is_char<'X'>(cigar_op))
+        if (('M'_cigar_op == op) || ('='_cigar_op == op) || ('X'_cigar_op == op))
```
joergi-w: This can be replaced with a `switch(op)` statement.

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -488,14 +488,14 @@ class alignment_file_input_format<format_sam>
      */
     template <typename align_type, typename ref_seqs_type>
     void construct_alignment(align_type                           & align,
-                             std::vector<std::pair<char, size_t>> & cigar,
+                             std::vector<cigar>                   & cigar_vector,
```
joergi-w: Please adapt doxygen's `\param[in]` statement to the renamed cigar_vector parameter.

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -476,6 +477,21 @@ class alignment_file_input_format<format_sam>
         }
     }
 
+    void transfer_soft_clipping_to(std::vector<cigar> const & cigar_vector, int32_t & sc_begin, int32_t & sc_end)
```
joergi-w: documentation is missing...

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -435,8 +438,39 @@ class alignment_file_input_format<format_bam> : alignment_file_input_format<form
     //!\brief Local buffer to read into while avoiding reallocation.
     std::string string_buffer{};
 
+    //!\copydoc std::tuple<std::vector<cigar>, size_t, size_t> parse_cigar(cigar_input_type && cigar_input)
```
joergi-w: remove return type for copydoc

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -477,6 +477,80 @@ class alignment_file_input_format<format_sam>
         }
     }
 
+
+    /*!\brief Updates the sequence lengths by `cigar_count` depending on the cigar operation `op`.
+     * \param[in, out]  ref_length  The reference sequence' length.
```
joergi-w: The reference sequence's length. (also below for query)

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -477,6 +477,80 @@ class alignment_file_input_format<format_sam>
...
+     * \param[in, out]  seq_length  The query sequence' length.
+     * \param[in]       op          The cigar operation.
+     * \param[in]       cigar_count The cigar count value to add to the length depending.
```
joergi-w: depending on...?

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -477,6 +477,80 @@ class alignment_file_input_format<format_sam>
...
+    template <std::ranges::InputRange cigar_input_type>
+    std::tuple<std::vector<cigar>, int32_t, int32_t> parse_cigar(cigar_input_type && cigar_input)
+    {
+        std::vector<cigar> operations{};
+        std::array<char, 20> buffer{}; // buffer to parse numbers with from_chars. Biggest number should fit in uint64_t
+        char op{'\0'};
+        uint32_t cigar_count{};
+        int32_t ref_length{}, seq_length{}; // length of aligned part for ref and query
+
+        // transform input into a signle input view if it isn't already
```
joergi-w: single

include/seqan3/io/alignment_file/detail.hpp
```diff
@@ -266,73 +267,76 @@ std::string get_cigar_string(alignment_type && alignment,
...
-    std::vector<std::pair<char, size_t>> result{};
+    std::vector<cigar> result{};
 
     if (!ref_seq.size())
         return result; // return empty string if sequences are empty
 
     // Add (S)oft-clipping at the start of the read
     if (query_start_pos)
-        result.emplace_back('S', query_start_pos);
+        result.push_back(cigar{query_start_pos, 'S'_cigar_op});
```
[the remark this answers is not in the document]
smehringer: good point

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -849,24 +851,37 @@ class alignment_file_output_format<format_bam> : alignment_file_output_format<fo
...
+                ref_length = std::ranges::distance(get<1>(align));
 
-            for (auto chr : get<1>(align))
-                if (chr == gap{})
-                    ++off_end;
+                // compute possible distance from alignment end to sequence end
+                // which indicates soft clipping at the end.
+                // This should be replaced by a free count_gaps function for
+                // aligned sequences which is more efficient if possible.
+                auto off_end{std::ranges::distance(seq) - offset};
```
joergi-w: Is this `auto` a signed type? If not you should have an assertion below that `off_end -= ref_length;`cannot become negative. 

Anyway I would not use auto here, as the resulting type from the difference cannot be easily seen.

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -849,24 +851,37 @@ class alignment_file_output_format<format_bam> : alignment_file_output_format<fo
...
+                cigar_vector = detail::get_cigar_vector(align, offset, off_end);
+            }
+            else
+            {
+                int32_t d{}; // dummy seq length
+                for (auto & [co, op] : cigar_vector)
+                    alignment_file_input_format<format_sam>::update_alignment_lengths(ref_length, d, op.to_char(), co);
+            }
 
             if (cigar_vector.size() > 65535) // cannot be represented with 16bits, must be written into the sam tag CG
```
joergi-w: use `size >= (1 << 16)` to make clear

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -1292,6 +1293,11 @@ class alignment_file_output_format<format_sam>
 
             write_range(stream_it, detail::get_cigar_string(std::forward<align_type>(align), offset, off_end));
         }
+        else if (!cigar_vector.empty()) // both fields are never non-empty (fields asserted on file level)
```
joergi-w: never non-empty => always empty

test/unit/io/alignment_file/alignment_file_format_test_template.hpp
```diff
@@ -495,6 +495,47 @@ TYPED_TEST_P(alignment_file_write, with_header)
...
+                                                                      field::REF_OFFSET,
+                                                                      field::MAPQ,
+                                                                      field::CIGAR, // cigar instead of alignment
+                                                                      field::OFFSET,
+                                                                      field::MATE,
+                                                                      field::SEQ,
+                                                                      field::QUAL,
+                                                                      field::TAGS>{}};
+
+        for (size_t i = 0; i < 3; ++i)
```
joergi-w: Use `0ul` and `3ul` literals because you compare unsigned long values.
(Doesn't it trigger a warning?)

include/seqan3/io/alignment_file/output_format_concept.hpp
```diff
@@ -111,6 +114,7 @@ SEQAN3_CONCEPT AlignmentFileOutputFormat =
                   ref_id_type                            && ref_id,
                   ref_offset_type                        && ref_offset,
                   align_type                             && align,
```
joergi-w: Why are they documented as `&&` but in the code they are `&`?

doc/tutorial/alignment_file/alignment_file_read_cigar.cpp
```diff
@@ -0,0 +1,44 @@
...
+@HD	VN:1.6	SO:coordinate
+@SQ	SN:ref	LN:45
+r001	99	ref	7	30	8M2I4M1D3M	=	37	39	TTAGATAAAGGATACTG	*
+r003	0	ref	9	30	5S6M	*	0	0	GCCTAAGCTAA	*	SA:Z:ref,29,-,6H5M,17,0;
+r004	0	ref	16	30	6M14N5M	*	0	0	ATAGCTTCAGC	*
+r003	2064	ref	29	17	5M	*	0	0	TAGGC	*	SA:Z:ref,9,+,5S6M,30,1;
+r001	147	ref	37	30	9M	=	7	-39	CAGCGGCAT	*	NM:i:1
+)//![sam_file]";
+
+        std::ofstream file{std::filesystem::temp_directory_path()/"my.sam"};
```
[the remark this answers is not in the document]
smehringer: nop the filesystem path library provides `operator/()` which will add `/` or `\` depending on you operating system

test/unit/io/alignment_file/alignment_file_format_test_template.hpp
```diff
@@ -332,6 +332,24 @@ TYPED_TEST_P(alignment_file_read, read_mate_but_not_ref_id_without_ref)
...
+        {{1, 'H'_cigar_op}, {7, 'M'_cigar_op}, {1, 'D'_cigar_op}, {1, 'M'_cigar_op}, {1, 'S'_cigar_op}},
+        {{1, 'S'_cigar_op}, {1, 'M'_cigar_op}, {1, 'P'_cigar_op}, {1, 'M'_cigar_op}, {1, 'I'_cigar_op},
+         {1, 'M'_cigar_op}, {1, 'I'_cigar_op}, {1, 'D'_cigar_op}, {1, 'M'_cigar_op}, {1, 'S'_cigar_op}}
+    };
+
+    typename TestFixture::stream_type istream{this->simple_three_reads_input};
+    alignment_file_input fin{istream, TypeParam{}, fields<field::CIGAR>{}};
+
+    size_t i{0};
+    for (auto & [cigar_v] : fin)
```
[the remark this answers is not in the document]
smehringer: Yes, I need this because it is a 1-element-tuple that is decomposed to its element. Without the `[]` I would get the tuple :)

include/seqan3/io/alignment_file/output_format_concept.hpp
smehringer: well && on a template accepts both lvalues and rvalues, The concept only checks lvalues. 
Rvalues should probably be required too... But htis should be part of separate PR. I'll make a card

test/unit/io/alignment_file/alignment_file_format_test_template.hpp
smehringer: nope ^^. I changed it

include/seqan3/io/alignment_file/detail.hpp
smehringer: I did replace it in when the code was moved in to the sam format

test/unit/io/alignment_file/alignment_file_format_test_template.hpp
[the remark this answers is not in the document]
joergi-w: Okay, I didn't expect a 1-element tuple ;) Then it's fine!

doc/tutorial/alignment_file/index.md
```diff
@@ -87,7 +87,7 @@ To make things clearer, here is the table of SAM columns connected to the corres
 | 3  | RNAME         | seqan3::field::REF_ID                             |
 | 4  | POS           | seqan3::field::REF_OFFSET                         |
 | 5  | MAPQ          | seqan3::field::MAPQ                               |
-| 6  | CIGAR         | implicitly stored in seqan3::field::ALIGNMENT     |
+| 6  | CIGAR         | implicitly stored in seqan3::field::ALIGNMENT or directly in seqan3::CIGAR    |
```
joergi-w: use `seqan3::field::CIGAR`

include/seqan3/io/alignment_file/detail.hpp
```diff
@@ -87,107 +87,138 @@ template <typename reference_char_type, typename query_char_type>
...
- * \returns An std::string representing the alignment as a CIGAR string.
+ * \returns An std::vector<seqan3::cigar> representing the alignment.
  *
- * ### Theoretical Example:
+ * \details
  *
- * The following alignment reference sequence on top and the query sequence at
- * the bottom.
+ * \attention Note that CIGAR elements (respectively by their
+ *            CIGAR operation) are always related to on one of the two sequences
```
joergi-w: ~on~

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -806,27 +851,41 @@ class alignment_file_output_format<format_bam> : alignment_file_output_format<fo
...
+                ref_length = std::ranges::distance(get<1>(align));
 
-            for (auto chr : get<1>(align))
-                if (chr == gap{})
-                    ++off_end;
+                // compute possible distance from alignment end to sequence end
+                // which indicates soft clipping at the end.
+                // This should be replaced by a free count_gaps function for
+                // aligned sequences which is more efficient if possible.
+                assert(std::ranges::distance(seq) >= offset);
```
joergi-w: for a signed type you do not need this assertion...

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -806,27 +851,41 @@ class alignment_file_output_format<format_bam> : alignment_file_output_format<fo
...
+                int32_t off_end{static_cast<int32_t>(std::ranges::distance(seq)) - offset};
 
-            off_end -= std::ranges::distance(get<1>(align));
+                for (auto chr : get<1>(align))
+                    if (chr == gap{})
+                        ++off_end;
 
-            std::vector<std::pair<char, size_t>> cigar = detail::get_cigar_vector(align, offset, off_end);
+                off_end -= ref_length;
```
joergi-w: in case you change to unsigned, you need the assertion for this

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -476,26 +483,107 @@ class alignment_file_input_format<format_sam>
...
+        }
+    };
+
+    /*!\brief Parses a cigar string into a vector of operation-count pairs (e.g. (M, 3)).
+     * \ingroup alignment_file
+     * \tparam cigar_input_type The type of a single pass input view over the cigar string; must model
+     *                          std::ranges::InputRange.
+     * \param[in]  cigar_input  The single pass input view over the cigar string to parse.
+     *
+     * \returns A tuple of size five containing (1) std::vector over seqan3::cigar, that describes
```
joergi-w: ~five~ three

include/seqan3/io/alignment_file/input_format_concept.hpp
```diff
@@ -157,6 +162,7 @@ SEQAN3_CONCEPT AlignmentFileInputFormat =
  * \param[out]    ref_id      The buffer for seqan3::field::REF_ID input.
  * \param[out]    ref_offset  The buffer for seqan3::field::REF_OFFSET input.
  * \param[out]    align       The buffer for seqan3::field::ALIGNMENT input.
+ * \param[out]    cigar       The buffer for seqan3::field::CIGAR input.
```
joergi-w: add `cigar_type & cigar` also to the parameter list of `\fn void read` above (lines 128-133)

include/seqan3/io/alignment_file/format_bam.hpp
smehringer: neither reflength nor off_end are unsigned

include/seqan3/io/alignment_file/output_format_concept.hpp
joergi-w: Thanks!

ppericard:
Hi,
Any news about integrating this pull request?
I am very eager to be able to get the CIGAR string from alignment files to continue my developments.
Thanks for all the work ^_^
Pierre

smehringer:
Hi @ppericard,

great to hear that this module is in use :)

There is an internal compiler error in the range library that has been fixed in the new range-v3 release so I am just wating for some PRs that make our library compatible to the that release. I think it will be a matter of 2-3 weeks. 

I will try raise the priority of this!

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -851,24 +853,37 @@ class alignment_file_output_format_REMOVEME<format_bam> : alignment_file_output_
...
-                    ++off_end;
+                // compute possible distance from alignment end to sequence end
+                // which indicates soft clipping at the end.
+                // This should be replaced by a free count_gaps function for
+                // aligned sequences which is more efficient if possible.
+                int32_t off_end{static_cast<int32_t>(std::ranges::distance(seq)) - offset};
 
-            off_end -= std::ranges::distance(get<1>(align));
+                for (auto chr : get<1>(align))
+                    if (chr == gap{})
```
rrahn: `off_end += static_cast<uint8_t>(chr == gap{});`
Actually, I checked on compiler explorer and both versions seem to result in the same code. Still it would be good to be explicit

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -851,24 +853,37 @@ class alignment_file_output_format_REMOVEME<format_bam> : alignment_file_output_
...
+                        ++off_end;
 
-            std::vector<cigar> cigar_vector = detail::get_cigar_vector(align, offset, off_end);
+                off_end -= ref_length;
+                cigar_vector = detail::get_cigar_vector(align, offset, off_end);
+            }
+            else
+            {
+                int32_t d{}; // dummy seq length
+                for (auto & [co, op] : cigar_vector)
```
rrahn: co and op? maybe better names?

include/seqan3/io/alignment_file/output_format_concept.hpp
```diff
@@ -145,6 +149,7 @@ SEQAN3_CONCEPT alignment_file_output_format =
  * \param[in]     ref_id     The data for seqan3::field::REF_ID, e.g. the reference id..
  * \param[in]     ref_offset The data for seqan3::field::REF_OFFSET, i.e. the start position of the alignment in \p ref_seq.
  * \param[in]     align      The data for seqan3::field::ALIGN, e.g. the alignment between query and ref.
+ * \param[in]     cigar_vector The data for seqan3::field::CIGAR, e.g. representing the alignment between query and ref.
```
rrahn: you might want to align all parameters. Or just remove it completely for exact this reason 😄

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -154,7 +154,8 @@ class sequence_file_output_format_REMOVEME<format_sam>
 
         align_format.write(stream, output_options, std::ignore,
                            default_or(sequence), default_or(qualities), default_or(id),
-                           0, std::string_view{}, std::string_view{}, -1, default_align_t{}, 0, 0,
+                           0, std::string_view{}, std::string_view{}, -1, default_align_t{},
+                           std::vector<cigar>{}, 0, 0,
```
rrahn: fits not on the line above?

include/seqan3/io/alignment_file/detail.hpp
```diff
@@ -87,21 +87,21 @@ template <typename reference_char_type, typename query_char_type>
...
-                            bool const extended_cigar)
+cigar_op compare_aligned_values(reference_char_type const reference_char,
+                                query_char_type const query_char,
+                                bool const extended_cigar)
 {
     return (reference_char == gap{})
                 ? (query_char == gap{})
-                    ? 'P'
-                    : 'I'
+                    ? 'P'_cigar_op
```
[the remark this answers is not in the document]
smehringer: This will probably be faster depending on the compiler optimization for ternary expressions but I do not find this more readable. Especially when you are not accustomed to bit operations you really have to think about what's happening here whereas in if-else (ternary) expressions you just read it.

I would say we wait for the benchmark to see if this makes a performance difference but since I do not want to stall this PR I'll just include your changes

include/seqan3/io/alignment_file/detail.hpp
```diff
@@ -352,38 +355,38 @@ std::vector<std::pair<char, size_t>> get_cigar_vector(alignment_type && alignmen
...
     size_t sc_end_count{};   // number of soft clipped bases at the end
     std::array<char, 20> buffer{}; // buffer to parse numbers with from_chars. Biggest number should fit in uint64_t
-    char cigar_op{'\0'};
-    size_t cigar_count{}, ref_length{}, seq_length{}; // length of aligned part for ref and query
+    char op{'\0'};
+    uint32_t cigar_count{};
+    size_t ref_length{}, seq_length{}; // length of aligned part for ref and query
 
-    auto update_lengths_fn = [&ref_length, &seq_length, &cigar_op, &cigar_count] ()
+    auto update_lengths_fn = [&ref_length, &seq_length, &op, &cigar_count] ()
```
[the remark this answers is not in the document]
smehringer: `cigar_value` suggests that this would be a value of type `cigar`, which is not true. I changed it to `cigar_operation` I hope that is fine for you.

include/seqan3/io/alignment_file/detail.hpp
```diff
@@ -104,65 +104,77 @@ cigar_op compare_aligned_values(reference_char_type const reference_char,
...
  * \param  extended_cigar  Whether to print the extended cigar alphabet or not. See CIGAR operation.
- * \returns An std::string representing the alignment as a CIGAR string.
+ * \returns An std::vector<seqan3::cigar> representing the alignment.
  *
- * ### Theoretical Example:
+ * \details
  *
- * The following alignment reference sequence on top and the query sequence at
- * the bottom.
+ * \attention Note that CIGAR elements (respectively by their
```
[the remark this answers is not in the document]
smehringer: Because I like it more compact and think 120 line length is too long.. and since documentation is :nail_care: I thought it is not a problem but personal flavour.

include/seqan3/io/alignment_file/detail.hpp
```diff
@@ -238,106 +251,65 @@ std::string get_cigar_string(alignment_type && alignment,
                              uint32_t const query_end_pos = 0,
                              bool const extended_cigar = false)
 {
-    return get_cigar_string(get<0>(alignment), get<1>(alignment),
-                            query_start_pos, query_end_pos, extended_cigar);
+    std::stringstream ss;
+    debug_stream_type ds{ss};
```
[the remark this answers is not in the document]
smehringer: I agree that these detail functions are not at all a nice design, this was part of the alignment IO PR that needed to be merged for a release on time pressure (not that this is a real excuse but just saying).
I'm more in favour to provide views that do the conversion on a public level. I already have a card for this in the discussions board but we didn't have enough time for it last monday.

include/seqan3/io/alignment_file/detail.hpp
```diff
@@ -238,106 +251,65 @@ std::string get_cigar_string(alignment_type && alignment,
...
  *
  * \attention Note that CIGAR elements (respectively by their
- *            CIGAR operation) are always related to on one of the two sequences
+ *            CIGAR operation) are always related to one of the two sequences
  *            in a pairwise alignment. In this case, the resulting cigar_vector
  *            is based on sequence at the second position of the \p alignment pair,
  *            namely the query sequence.
  *
- * ### Example:
+ * ### Theoretical Example:
```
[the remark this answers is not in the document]
smehringer: Because there is no code snippet?

include/seqan3/io/alignment_file/detail.hpp
```diff
@@ -238,106 +251,65 @@ std::string get_cigar_string(alignment_type && alignment,
...
 //!\cond
-    requires std::tuple_size_v<remove_cvref_t<alignment_type>> == 2
+    requires std::detail::weakly_equality_comparable_with<gap, reference_t<ref_seq_type>> &&
+             std::detail::weakly_equality_comparable_with<gap, reference_t<query_seq_type>>
 //!\endcond
-std::vector<cigar> get_cigar_vector(alignment_type && alignment,
-                                    uint32_t const query_start_pos = 0,
-                                    uint32_t const query_end_pos = 0,
-                                    bool const extended_cigar = false)
+std::string get_cigar_string(ref_seq_type && ref_seq,
```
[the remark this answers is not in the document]
smehringer: I know you are not in favour of this whole design but I would like to avoid too many naming changes in this already big PR. I can refactor all the namings in a later PR, if that is fine?

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -476,6 +477,27 @@ class alignment_file_input_format_REMOVEME<format_sam>
...
 
+
+    /*!\brief Transfer soft clipping information from the \p cigar_vector to \p sc_begin and \p sc_end.
+     * \param[in] cigar_vector The cigar information to parse for soft-clipping.
+     * \param[in,out] sc_begin The soft clipping at the beginning of the alignment to set.
+     * \param[in,out] sc_end   The soft clipping at the end of the alignment to set.
+     */
+    void transfer_soft_clipping_to(std::vector<cigar> const & cigar_vector, int32_t & sc_begin, int32_t & sc_end)
+    {
+        // check for soft clipping at the first two positions
```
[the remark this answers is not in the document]
smehringer: Because soft and hard clipping can only occur at the beginning or end of an aligned read.
It can be only hard or soft clipping (first position to check) but it can also occur both (first position hard clipping, second soft clipping).

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -435,8 +438,41 @@ class alignment_file_input_format_REMOVEME<format_bam> : alignment_file_input_fo
...
+    //!\param[in] n_cigar_op The number of cigar elements to read from the cigar_input.
+    //!\copydetails parse_cigar(cigar_input_type && cigar_input)
+    template <std::ranges::input_range cigar_input_type>
+    auto parse_cigar(cigar_input_type && cigar_input, uint16_t n_cigar_op)
+    {
+        std::vector<cigar> operations{};
+        char op{'\0'};
+        uint32_t cigar_count{};
+        int32_t ref_length{}, seq_length{};
+        uint32_t o_and_c{};
```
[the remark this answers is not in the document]
smehringer: operation_and_count :D

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -435,8 +438,41 @@ class alignment_file_input_format_REMOVEME<format_bam> : alignment_file_input_fo
     //!\brief Local buffer to read into while avoiding reallocation.
     std::string string_buffer{};
 
+    //!\copybrief parse_cigar(cigar_input_type && cigar_input)
+    //!\param[in] n_cigar_op The number of cigar elements to read from the cigar_input.
+    //!\copydetails parse_cigar(cigar_input_type && cigar_input)
+    template <std::ranges::input_range cigar_input_type>
+    auto parse_cigar(cigar_input_type && cigar_input, uint16_t n_cigar_op)
```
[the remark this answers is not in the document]
smehringer: > class method should be const?

Well the streams are not within the format but they will be soon, and then parsing the input means changing the stream iterator member. So now it should be const but it won't be anymore soon. Shall I add it for now?

include/seqan3/io/alignment_file/detail.hpp
```diff
@@ -104,65 +104,77 @@ cigar_op compare_aligned_values(reference_char_type const reference_char,
...
+/*!\brief Creates a CIGAR string (SAM format) given an alignment represented by two aligned sequences.
  * \ingroup alignment_file
  *
- * \tparam ref_seq_type    Must model std::ranges::forward_range. The value_type must
- *                         be equality comparable to seqan3::gap.
- * \tparam query_seq_type  Must model std::ranges::forward_range. The value_type must
- *                         be equality comparable to seqan3::gap.
- * \param  ref_seq         The reference sequence to compare against the query sequence.
- * \param  query_seq       The query sequence to build the CIGAR string for.
+ * \tparam alignment_type  Must model the seqan3::tuple_like and must
```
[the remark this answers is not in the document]
smehringer: well right now the concept seqan3::aligned_sequence is to restrictive as it requires "insert_gap" although I only need comparibility to gap so I would rather not require the concept here.

doc/tutorial/alignment_file/alignment_file_read_cigar.cpp
```diff
@@ -0,0 +1,44 @@
+#include <fstream>
+
+#include <seqan3/std/filesystem>
+
+struct write_file_dummy_struct
+{
+    write_file_dummy_struct()
+    {
+
+auto file_raw = R"//![sam_file](
```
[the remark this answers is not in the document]
smehringer: I explicitly want the doxygen command in there so I can use this in a snippet saying, `let's take the following sam file`

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -91,7 +91,7 @@ class sequence_file_input_format_REMOVEME<format_sam>
         else
         {
             align_format.read(stream, align_options, std::ignore, default_header, sequence, qualities, id,
-                              std::ignore, std::ignore, std::ignore, std::ignore, std::ignore,
+                              std::ignore, std::ignore, std::ignore, std::ignore, std::ignore, std::ignore,
```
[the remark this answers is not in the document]
smehringer: ?

doc/tutorial/alignment_file/alignment_file_read_cigar.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+        std::ofstream file{std::filesystem::temp_directory_path()/"my.sam"};
+        std::string str{file_raw};
+        file << str.substr(1); // skip first newline
+    }
+};
+
+write_file_dummy_struct go{};
+
+//![code]
```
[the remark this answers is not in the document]
smehringer: well this code is completely copy and pastable, it just excludes the file creation above so I guess a snippet is still fine?

doc/tutorial/alignment_file/alignment_file_read_cigar.cpp
```diff
@@ -0,0 +1,44 @@
...
+r001	147	ref	37	30	9M	=	7	-39	CAGCGGCAT	*	NM:i:1
+)//![sam_file]";
+
+        std::ofstream file{std::filesystem::temp_directory_path()/"my.sam"};
+        std::string str{file_raw};
+        file << str.substr(1); // skip first newline
+    }
+};
+
+write_file_dummy_struct go{};
```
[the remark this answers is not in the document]
smehringer: In a beginners tutorial I think the use case of reading a file is more appropriate that a stream?
I also state in the tutorial that this example is used on this file (by adding the file in a snippet).

include/seqan3/io/alignment_file/input.hpp
```diff
@@ -448,6 +450,8 @@ class alignment_file_input
                                          typename traits_type::quality_alphabet>;
     //!\brief The type of field::FLAG is fixed to uint16_t.
     using flag_type                = uint16_t;
+    //!\brief The type of field::FLAG is fixed to uint16_t.
+    using cigar_type                = std::vector<cigar>;
```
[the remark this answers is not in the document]
smehringer: :roll_eyes:

test/unit/io/alignment_file/alignment_file_format_test_template.hpp
```diff
@@ -332,6 +332,24 @@ TYPED_TEST_P(alignment_file_read, read_mate_but_not_ref_id_without_ref)
...
+        EXPECT_EQ(cigar_v, expected[i++]);
```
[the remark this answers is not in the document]
smehringer: There is a test "read_in_all_but_empty_data" which checks if the parsing works although every entry in the record is empty

include/seqan3/io/alignment_file/format_bam.hpp
smehringer: isn't the if check more explicit?

include/seqan3/io/alignment_file/detail.hpp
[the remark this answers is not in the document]
rrahn: In my opinion code style, counting in comments and API documentation, should follow a style guide and not personal preferences to avoid humble mumble in the code. I know that this was never documented appropriately but this seems to me like a relatively new flavor that was never done before. So I would recommend to stick to the common style and use the 120 line lengths. Before introducing new styles I would prefer discussing it. So to avoid confusions as we had when a subgroup of people decided putting complex nested classes out of its enclosing class body, which was not communicated properly 😅 .

include/seqan3/io/alignment_file/detail.hpp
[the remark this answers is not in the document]
rrahn: yeah I know. That was just the feeling of why we need to handle things more agile or different then before and the idea came into my mind and maybe there was already discussion for this. But cool, let's revisit it later.

include/seqan3/io/alignment_file/detail.hpp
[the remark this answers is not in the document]
rrahn: I agree, there is a lot of things that are going on. A lot of things we need to revisit for file design anyway. So rather have a dedicated discussion and plan for this in the future. Let's not change it now then, because I feel more will be happening

include/seqan3/io/alignment_file/format_sam.hpp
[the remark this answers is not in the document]
rrahn: So there can be hard clipping followed by soft clipping. Is that it?

include/seqan3/io/alignment_file/format_bam.hpp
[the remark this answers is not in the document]
rrahn: I guess, you can decide.

include/seqan3/io/alignment_file/format_bam.hpp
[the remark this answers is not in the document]
rrahn: maybe call it then `operation_and_count` 😉

doc/tutorial/alignment_file/alignment_file_read_cigar.cpp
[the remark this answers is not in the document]
rrahn: oh ok.

doc/tutorial/alignment_file/alignment_file_read_cigar.cpp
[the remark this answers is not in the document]
rrahn: Well, the reason is that the snippet when executed will create a file called with a generic name in a generic location maybe used by some other snippets as well. So it would be possible that if they get executed they overwrite something another process is using. That is when for example we start to explicitly check for the right output of the files in tests. But also it writes something into the tmp directory and not removing it anymore. So we decided to make all snippets work with streams if they are not explicitly demonstrating some file interface. I guess that is the case here. All I/O snippets were adapted to this design.

include/seqan3/io/sequence_file/format_sam.hpp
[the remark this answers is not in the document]
rrahn: this interface. It is killing me! 🔪

include/seqan3/io/alignment_file/format_bam.hpp
rrahn: not explicit for the human eye I mean, but for the compiler 🤖

doc/tutorial/alignment_file/alignment_file_read_cigar.cpp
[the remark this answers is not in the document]
rrahn: I guess we need to move the temporary file from the tests into the library. Right now you creating my.sam in the tmp directory. I wonder how often that happens in our code snippets. That seems like something will go wrong along the way.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -595,9 +595,9 @@ class output_file_validator : public file_validator_base
     //!\brief Returns a message that can be appended to the (positional) options help page info.
     std::string get_help_page_message() const
     {
-        return detail::to_string("Valid output file formats: [",
-                                 extensions | std::views::join(std::string{", "}),
-                                 "]");
+        using namespace std::literals::string_literals;
```
rrahn: what happend with to_string?

include/seqan3/io/alignment_file/detail.hpp
```diff
@@ -104,65 +104,77 @@ cigar_op compare_aligned_values(reference_char_type const reference_char,
...
-                             query_seq_type && query_seq,
-                             uint32_t const query_start_pos = 0,
-                             uint32_t const query_end_pos = 0,
-                             bool const extended_cigar = false)
+std::vector<cigar> get_cigar_vector(alignment_type && alignment,
+                                    uint32_t const query_start_pos = 0,
+                                    uint32_t const query_end_pos = 0,
+                                    bool const extended_cigar = false)
 {
+    auto & ref_seq = get<0>(alignment);
```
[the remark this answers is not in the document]
rrahn: Actually I have to correct my self. std::get is not a customisation point. So you would not be allowed to overload get in std namespace. I think we did it in some places but it must be removed. This also means, that you don't need `using std::get` because you would not look for an overload in std namespace for a third party type.

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -410,7 +416,10 @@ class alignment_file_input_format_REMOVEME<format_bam> : alignment_file_input_fo
                                        "record.")};
 
                     auto cigar_view = std::views::all(std::get<std::string>(it->second));
-                    std::tie(cigar, ref_length, seq_length, offset_tmp, soft_clipping_end) = detail::parse_cigar(cigar_view);
+                    std::tie(tmp_cigar_vector, ref_length, seq_length) = sam_fmt::parse_cigar(cigar_view);
+                    offset_tmp = 0, soft_clipping_end = 0;
```
rrahn:
```suggestion
                    offset_tmp = soft_clipping_end = 0;
```

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -432,8 +444,41 @@ class alignment_file_input_format_REMOVEME<format_bam> : alignment_file_input_fo
...
     std::string string_buffer{};
 
+    //!\copybrief parse_cigar(cigar_input_type && cigar_input)
+    //!\param[in] n_cigar_op The number of cigar elements to read from the cigar_input.
+    //!\copydetails parse_cigar(cigar_input_type && cigar_input)
+    template <typename cigar_input_type>
+    [[nodiscard]] auto parse_cigar(cigar_input_type && cigar_input, uint16_t n_cigar_op) const
+    {
+        std::vector<cigar> operations{};
+        char op{'\0'};
```
rrahn:
```suggestion
        char operation{};
```

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -432,8 +444,41 @@ class alignment_file_input_format_REMOVEME<format_bam> : alignment_file_input_fo
...
+        uint32_t cigar_count{};
+        int32_t ref_length{}, seq_length{};
+        uint32_t o_and_c{};
+        constexpr char const * CIGAR_MAPPING = "MIDNSHP=X*******";
```
rrahn: we actually do not use capital letters anymore for constants, do we?

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -806,27 +853,43 @@ class alignment_file_output_format_REMOVEME<format_bam> : alignment_file_output_
...
+                    if (chr == gap{})
+                        ++off_end;
 
-            std::vector<std::pair<char, size_t>> cigar = detail::get_cigar_vector(align, offset, off_end);
+                off_end -= ref_length;
+                cigar_vector = detail::get_cigar_vector(align, offset, off_end);
+            }
+            else
+            {
+                int32_t d{}; // dummy seq length
```
rrahn:
```suggestion
                int32_t dummy_seq_length{}; 
```

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -476,26 +483,110 @@ class alignment_file_input_format_REMOVEME<format_sam>
...
 
+    /*!\brief Updates the sequence lengths by `cigar_count` depending on the cigar operation `op`.
+     * \param[in, out]  ref_length  The reference sequence's length.
+     * \param[in, out]  seq_length  The query sequence's length.
+     * \param[in]       op          The cigar operation.
+     * \param[in]       cigar_count The cigar count value to add to the length depending on the cigar operation.
+     */
+    static void update_alignment_lengths(int32_t & ref_length,
+                                         int32_t & seq_length,
+                                         char cigar_operation,
```
rrahn:
```suggestion
                                         char const cigar_operation,
```

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -476,26 +483,110 @@ class alignment_file_input_format_REMOVEME<format_sam>
...
+                throw format_error{"Corrupted cigar string encountered"};
+
+            update_alignment_lengths(ref_length, seq_length, cigar_operation, cigar_count);
+            operations.emplace_back(cigar_count, cigar_op{}.assign_char(cigar_operation));
+        }
+
+        return {operations, ref_length, seq_length};
+    }
+
+    /*!\brief Transfer soft clipping information from the \p cigar_vector to \p sc_begin and \p sc_end.
```
rrahn:
```suggestion
    /*!\brief Transfer soft clipping information from the `cigar_vector` to `sc_begin` and `sc_end`.
```

include/seqan3/io/alignment_file/output.hpp
```diff
@@ -228,6 +230,13 @@ class alignment_file_output
                   "please refer to the documentation of "
                   "seqan3::alignment_file_output::field_ids for the accepted values.");
 
+    static_assert([] () constexpr
+                  {
+                      return !(selected_field_ids::contains(field::ALIGNMENT) &&
+                               selected_field_ids::contains(field::CIGAR));
+                  }(),
+                  "You may not select field::ALIGNMENT and field::CIGAR at the same time.");
```
rrahn: actually why not? Both information are already present. And how do I specify to  get the cigar field instead of the alignment when I want otherwise have the default behavior? Is there a recommended way of doing it?

doc/tutorial/alignment_file/alignment_file_read_cigar.cpp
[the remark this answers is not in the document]
smehringer: we could use the test::tmp_filename

include/seqan3/io/alignment_file/output.hpp
smehringer: hm.. I guess we could.

include/seqan3/io/alignment_file/format_sam.hpp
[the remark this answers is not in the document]
smehringer: yes

doc/tutorial/alignment_file/alignment_file_read_cigar.cpp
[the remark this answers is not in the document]
smehringer: again, can't we use the tmp_filename ?

include/seqan3/io/alignment_file/format_sam.hpp
smehringer: I think the \p command does this already?

include/seqan3/argument_parser/validators.hpp
smehringer: It causes an ICE because of a ranges-v3 but that is fixed if we update to the new ranges release, but since this is stalled I wanted to get this PR ready to merge anyway. I hope the ranges update comes soon but since this PR is of interest to users we should discuss to merge it anyway if we cannot update soonish.

include/seqan3/io/alignment_file/detail.hpp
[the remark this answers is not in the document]
smehringer: fine

include/seqan3/io/alignment_file/detail.hpp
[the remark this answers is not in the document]
smehringer: But I need it if I want to allow types that allow either `std::get` access or `seqan3::get` access.

include/seqan3/io/alignment_file/output.hpp
smehringer: ah no! I thought this is input.
But this is the output file. So of course we cannot provide both, the alingment and a cigar string because which should be written as they refer to the same thing.

include/seqan3/io/alignment_file/format_bam.hpp
smehringer: If it is the same code I vote for more explicit to the eye :eye:

include/seqan3/io/alignment_file/output.hpp
rrahn: Ah is output! 👍

include/seqan3/io/alignment_file/detail.hpp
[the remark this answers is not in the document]
rrahn: yes, you are right. Confusing topic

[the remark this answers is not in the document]
rrahn: But is it added now? The block does not seem updated?

include/seqan3/io/alignment_file/detail.hpp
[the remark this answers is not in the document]
rrahn: Can you remove please the theoretical. I think it is misleading. It is still an example isn't it?

doc/tutorial/alignment_file/alignment_file_read_cigar.cpp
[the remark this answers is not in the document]
rrahn: yes we can, but we have to port it to core or something.

doc/tutorial/alignment_file/alignment_file_read_cigar.cpp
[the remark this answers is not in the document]
rrahn: no, because in a deployed library test is not part of the package

include/seqan3/io/alignment_file/format_bam.hpp
rrahn: ok, leave it,

include/seqan3/argument_parser/validators.hpp
rrahn: Pheww, I know what you mean, but this will then stay in this forever. Unless, we can employ something that I just read about recently: https://www.fluentcpp.com/2019/01/01/todo_before-clean-codebase-2019/
The idea is to have a time based static assert, that triggers an assertion after a certain date has passed. Thus we could say ok in 2 weeks I reevaluate it again. But that's just gadget-me talking.

include/seqan3/io/alignment_file/detail.hpp
```diff
@@ -46,15 +47,12 @@ struct view_equality_fn
...
+ * \returns A seqan3::cigar_op representing the alignment operation between the two values.
  *
  * \details
  *
- * \attention Note that CIGAR elements (respectively by their
- *            CIGAR operation) are always related to on one of the two sequences
- *            in a pairwise alignment. In this case, the resulting
- *            operation is related to the \p query_char.
+ * \attention Note that CIGAR elements (respectively by their CIGAR operation) are always related to one of the two
+ *            sequences in a pairwise alignment. In this case, the resulting operation is related to the \p query_char.
```
rrahn: you might want to grep for \p and then replace it with tick marks as the desired behavior seems not to work.

include/seqan3/io/alignment_file/detail.hpp
```diff
@@ -209,11 +229,9 @@ std::string get_cigar_string(ref_seq_type && ref_seq,
  *
  * \details
  *
- * \attention Note that CIGAR elements (respectively by their
- *            CIGAR operation) are always related to on one of the two sequences
- *            in a pairwise alignment. In this case, the resulting cigar_vector
- *            is based on sequence at the second position of the \p alignment pair,
- *            namely the query sequence.
+ * \attention Note that CIGAR elements (respectively by their CIGAR operation) are always related to one of the
+ *            two sequences in a pairwise alignment. In this case, the resulting cigar_vector is based on sequence
```
rrahn: I don't understand this. What information do you like to express?

include/seqan3/io/alignment_file/format_sam.hpp
rrahn: as discussed.

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -476,26 +483,110 @@ class alignment_file_input_format_REMOVEME<format_sam>
...
+     *
+     * For example, the view over the cigar string "1H4M1D2M2S" will return
+     * `{[(H,1), (M,4), (D,1), (M,2), (S,2)], 7, 6}`.
+     */
+    template <typename cigar_input_type>
+    [[nodiscard]] std::tuple<std::vector<cigar>, int32_t, int32_t> parse_cigar(cigar_input_type && cigar_input) const
+    {
+        std::vector<cigar> operations{};
+        std::array<char, 20> buffer{}; // buffer to parse numbers with from_chars. Biggest number should fit in uint64_t
+        char cigar_operation{'\0'};
```
rrahn:
```suggestion
        char cigar_operation{};
```
should be the same

include/seqan3/argument_parser/validators.hpp
smehringer: Pheww this is hacky :laughing: 
I kind of hoped then we pull the ranges update before this review is done. But the change is not too bad is it? I mean it is just another way of solving this, not a real workaround..

smehringer:
@rrahn  Please mark the others as resolved :pray:

doc/tutorial/alignment_file/alignment_file_read_cigar.cpp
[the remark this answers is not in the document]
smehringer: Can I leave it like this and then port the tmp file ?

include/seqan3/io/alignment_file/detail.hpp
smehringer: That the cigar string is a relative thing... If it says "deletion of 2 bases", it means a deletion in the query not the reference, so you need to aware which is the reference sequence and which is the query sequence. 

In the case of `map_aligned_values_to_cigar_op` it is query_char and reference_char.

include/seqan3/argument_parser/validators.hpp
smehringer: seems like #1267 can soon be merged

include/seqan3/io/alignment_file/detail.hpp
```diff
@@ -75,128 +73,150 @@ struct view_equality_fn
...
  * \param  extended_cigar  Whether to print the extended cigar alphabet or not. See CIGAR operation.
- * \returns An std::string representing the alignment as a CIGAR string.
+ * \returns An std::vector<seqan3::cigar> representing the alignment.
  *
- * ### Theoretical Example:
+ * \details
  *
- * The following alignment reference sequence on top and the query sequence at
- * the bottom.
+ * \attention Note that CIGAR elements (respectively by their CIGAR operation) are always related to one of the two
```
rrahn:
```suggestion
The first sequence is always considered the reference sequence while the second one is considered the query sequence. The cigar operations regarding insertions and deletions are set accordingly.
```

Is it this, what you want to express?
If yes, could make the same/similar note above for the other function?

include/seqan3/io/alignment_file/detail.hpp
rrahn: I don't know. I find the description a bit complicated. I proposed some alternative description in a comment above. What do you think?

include/seqan3/io/alignment_file/detail.hpp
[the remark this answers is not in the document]
rrahn: unresolved?

include/seqan3/io/alignment_file/detail.hpp
```diff
@@ -230,41 +248,39 @@ std::string get_cigar_string(ref_seq_type && ref_seq,
...
+ * \returns An std::string representing the alignment as a CIGAR string.
  *
  * \details
  *
- * \attention Note that CIGAR elements (respectively by their
- *            CIGAR operation) are always related to on one of the two sequences
- *            in a pairwise alignment. In this case, the resulting cigar_vector
- *            is based on sequence at the second position of the \p alignment pair,
- *            namely the query sequence.
+ * \attention Note that CIGAR elements (respectively by their CIGAR operation) are always related to one of the two
```
rrahn: that would be also a candidate not to miss out if you agree on changing the description as proposed.

include/seqan3/io/alignment_file/detail.hpp
```diff
@@ -76,128 +73,149 @@ struct view_equality_fn
...
- * \tparam ref_seq_type    Must model std::ranges::forward_range. The value_type must
- *                         be equality comparable to seqan3::gap.
- * \tparam query_seq_type  Must model std::ranges::forward_range. The value_type must
- *                         be equality comparable to seqan3::gap.
- * \param  ref_seq         The reference sequence to compare against the query sequence.
- * \param  query_seq       The query sequence to build the CIGAR string for.
+ * \tparam alignment_type  Must model the seqan3::tuple_like and must have std::tuple_size 2.
+ *                         Each tuple element must be a range over values comparable to seqan3::gap.
+ * \param  alignment       The alignment, represented by a pair of aligned sequences,
+ *                         to be transformed into CIGAR_vector based on the
```
rrahn:
```
CIGAR_vector
cigar_vector
cigar vector
```
From these three I would prefer the latter since it does not name a variable really. Also in the note block

smehringer:
@rrahn I also updated the changelog!

smehringer:
@rrahn I think this can be merged now. I also rebased on current master.

smehringer:
@rrahn  ping

Issue #1199, opened 2019-07-30
Error compiling the argument parser test program


ppericard:
Hi,

I'm encountering a pb testing seqan3 and trying to go through the tutorial programs.

I compiled the first two tutorial programs without any error, but I have a fatal compiling error when trying to compile the minimal argument parser tutorial.
I tried with gcc-7 and gcc-9 and got 2 different errors.

Thanks in advance for your help

Cheers
Pierre

## Platform

  - SeqAn version:  latest as of today (a60194e)
  - Operating system: `Linux ppericard-matam-dev 4.15.0-54-generic #58-Ubuntu SMP Mon Jun 24 10:55:24 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux`
  - Compiler: `g++-7 (Ubuntu 7.4.0-1ubuntu1~18.04.1) 7.4.0` and `g++-9 (Ubuntu 9.1.0-2ubuntu2~18.04) 9.1.0 `

## Description

The argument parser tutorial program won't compile. The hello_world and debug stream tutorial programs are compiling and working without any pb (http://docs.seqan.de/seqan/3-master-user/tutorial_first_example.html)

### How to repeat the problem

[main.cpp.txt](https://github.com/seqan/seqan3/files/3448586/main.cpp.txt)

```
(base) ubuntu@ppericard-matam-dev:~/data/ovgraphbuild2/source$ g++-7 -O3 -DNDEBUG -Wall -Wextra -std=c++17 -fconcepts -I ../seqan3/include -isystem ../seqan3/submodules/range-v3/include -isystem ../seqan3/submodules/sdsl-lite/include -isystem ../seqan3/submodules/cereal/include -DSEQAN3_HAS_ZLIB=1 -DSEQAN3_HAS_BZIP2=1 main.cpp
```

```
(base) ubuntu@ppericard-matam-dev:~/data/ovgraphbuild2/source$ g++-9 -O3 -DNDEBUG -Wall -Wextra -std=c++17 -fconcepts -I ../seqan3/include -isystem ../seqan3/submodules/range-v3/include -isystem ../seqan3/submodules/sdsl-lite/include -isystem ../seqan3/submodules/cereal/include -DSEQAN3_HAS_ZLIB=1 -DSEQAN3_HAS_BZIP2=1 main.cpp
```

I also tried using cmake and end up with the same error

### Expected behaviour

The program should compile without error

### Actual behaviour

```
[...]

PR #1200, opened 2019-08-02
[FIX] fix detection of pthread on some linuxes

### LINKED ISSUE #1193: Do I need to add -lpthread to linker flags?
## Platform

  - SeqAn version: 3.0.0
  - Operating system: Linux weissweiler.molgen.mpg.de 4.19.19.mx64.244 #1 SMP Tue Feb 5 13:01:13 CET 2019 x86_64 GNU/Linux
  - Compiler: gcc (GCC) 7.3.0

### How to repeat the problem

I'm writing an application based on seqan3. This is how my project directory looks like:
```
submodules/seqan3
src/
- CMakeLists.txt
- main.cpp
```

CMakeLists.txt:
```
cmake_minimum_required (VERSION 3.4)
project (seqan3_tutorial CXX)
set(SeqAn3_DIR "${CMAKE_SOURCE_DIR}/../submodules/seqan3/build_system")
find_package (SeqAn3 REQUIRED)
add_executable (main main.cpp)
target_link_libraries (main seqan3::seqan3)
```

main.cpp:
```
#include <seqan3/argument_parser/all.hpp>   // includes all necessary headers
#include <seqan3/core/debug_stream.hpp>     // our custom output stream
#include <seqan3/std/filesystem>            // use std::filesystem::path
#include <seqan3/io/sequence_file/all.hpp>  // FASTA support
#include <seqan3/io/alignment_file/all.hpp> // SAM/BAM support

using namespace seqan3;

struct cmd_arguments
{
    std::filesystem::path alignment_file_path{};
    std::filesystem::path reference_file_path{};
};

void initialize_argument_parser(argument_parser & parser, cmd_arguments & args)
{
    parser.info.author = "foo";
    parser.info.short_description = "foo";
    parser.info.version = "0.0.1";
    parser.add_positional_option(args.alignment_file_path, "Read alignments in SAM or BAM format.",
                                 input_file_validator{{"sam", "bam"}} );
    parser.add_positional_option(args.reference_file_path, "Reference genome in FASTA format",
                                 input_file_validator{{"fa", "fasta"}} );
}

{
    argument_parser myparser{"detectBreakends", argc, argv};        // initialise myparser
    cmd_arguments args{};
    initialize_argument_parser(myparser, args);
    try
    {
        myparser.parse();                                          // trigger command line parsing
    }
    catch (parser_invalid_argument const & ext)                     // catch user errors
    {
        debug_stream << "[Error] " << ext.what() << "\n"; // customise your error message
        return -1;
    }
    // run_program(args.alignment_file_path, args.reference_file_path);
    return 0;
}

```

### Expected behaviour

I would expect the program to compile successfully with:
```
mkdir build
cd build
cmake ../src
make
```

### Actual behaviour

```
>make
Scanning dependencies of target main
[ 50%] Building CXX object CMakeFiles/main.dir/main.cpp.o
[100%] Linking CXX executable main
/usr/lib/gcc/x86_64-pc-linux-gnu/7.3.0/../../../../x86_64-pc-linux-gnu/bin/ld: CMakeFiles/main.dir/main.cpp.o: undefined reference to symbol 'pthread_create@@GLIBC_2.2.5'
/usr/lib/gcc/x86_64-pc-linux-gnu/7.3.0/../../../../x86_64-pc-linux-gnu/bin/ld: /lib/libpthread.so.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/main.dir/build.make:86: main] Error 1
make[1]: *** [CMakeFiles/Makefile2:73: CMakeFiles/main.dir/all] Error 2
make: *** [Makefile:84: all] Error 2
```

I found that add -lpthread to the linker flags solves this issue. But I wondered why it is necessary to do this manually if I only use the seqan argument parser and not much more. Any idea?

Cheers
David

h-2:
fixes #1193 #1199

It's strange that we didn't see this earlier. I know that for macOS we don't need to explicitly add it. Apparently FreeBSD also doesn't need it, but we should have seen this on Debian, or not? Maybe not all Linuxes are affected. In any case, the proposed fix is the recommended way of doing it with CMake nowadays.

### DIFF
--- a/build_system/seqan3-config.cmake
+++ b/build_system/seqan3-config.cmake
@@ -363,6 +363,24 @@ else ()
 endif ()
 
 # ----------------------------------------------------------------------------
+# thread support (pthread, windows threads)
+# ----------------------------------------------------------------------------
+
+set (THREADS_PREFER_PTHREAD_FLAG TRUE)
+find_package (Threads QUIET)
+
+if (Threads_FOUND)
+    set (SEQAN3_LIBRARIES ${SEQAN3_LIBRARIES} Threads::Threads)
+    if ("${CMAKE_THREAD_LIBS_INIT}" STREQUAL "")
+        seqan3_config_print ("Thread support:             builtin.")
+    else ()
+        seqan3_config_print ("Thread support:             via ${CMAKE_THREAD_LIBS_INIT}")
+    endif ()
+else ()
+    seqan3_config_print ("Thread support:             not found.")
+endif ()
+
+# ----------------------------------------------------------------------------
 # Require Ranges and SDSL
 # ----------------------------------------------------------------------------
 


h-2:
Note that apparently the FindThreads.cmake has a bug where it prints stuff, even if told to be quiet. I have reported it:
https://gitlab.kitware.com/cmake/cmake/issues/19550

(purely cosmetic)

PR #1204, opened 2019-08-07
[FEATURE] our own meta library

h-2:
This is our own meta library with almost identical interfaces for dealing with seqan3::type_list, but also parameter packs (it's usually faster to work on the latter).

Currently available:

* at (linear-time)
* front
* back
* concat
* drop (linear-time)
* drop_front (constant)
* drop_last (linear-time)
* take (linear-time)
* take_last (linear-time)
* take (linear-time)
* split_after (linear-time)

Notably absent:
* drop_back because it's not possible to do in constant time without integer_sequence and a whole lot instantiaten. Actually, back was already tricky! @rrahn: You could have a look at this after we merge this, or you could just use drop_front on your example and move the `void` to the top and the commas to the front.

### DIFF
--- a/include/seqan3/core/type_list.hpp
+++ b/include/seqan3/core/type_list.hpp
@@ -7,7 +7,7 @@
 
 /*!\file
  * \author Hannes Hauswedell <hannes.hauswedell AT fu-berlin.de>
- * \brief Provides seqan3::type_list and auxiliary type traits.
+ * \brief Provides seqan3::type_list.
  */
 
 #pragma once
@@ -18,6 +18,10 @@
 
 namespace seqan3
 {
+
+// ----------------------------------------------------------------------------
+// type_list class
+// ----------------------------------------------------------------------------
 
 /*!\brief Type that contains multiple types, an alias for
  * [meta::list](https://ericniebler.github.io/range-v3/structmeta_1_1list.html).
@@ -30,6 +34,10 @@ using type_list = meta::list<types...>;
 
 namespace seqan3::detail
 {
+
+// ----------------------------------------------------------------------------
+// TypeList concept
+// ----------------------------------------------------------------------------
 
 /*!\brief Auxiliary concept that checks whether a type is a specialisation of seqan3::type_list.
  * \ingroup core
--- a/include/seqan3/core/type_list/traits.hpp
+++ b/include/seqan3/core/type_list/traits.hpp
@@ -1,0 +1,552 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \author Hannes Hauswedell <hannes.hauswedell AT fu-berlin.de>
+ * \brief Provides traits for seqan3::type_list.
+ */
+
+#pragma once
+
+#include <seqan3/core/type_list.hpp>
+
+// ----------------------------------------------------------------------------
+// seqan3::pack_traits::detail
+// ----------------------------------------------------------------------------
+
+namespace seqan3::pack_traits::detail
+{
+
+/*!\brief Implementation for seqan3::pack_traits::at.
+ * \tparam idx      The index.
+ * \tparam head_t   Currently viewed pack_t element.
+ * \tparam tail_t   Rest of the type pack.
+ * \ingroup type_list
+ */
+template <ptrdiff_t idx, typename head_t, typename ...tail_t>
+auto at()
+{
+    if constexpr (idx == 0)
+        return std::type_identity<head_t>{};
+    else if constexpr (idx > 0)
+#if __clang__
+        return std::type_identity<__type_pack_element<idx - 1, tail_t...>>;
+#else
+        return at<idx - 1, tail_t...>();
+#endif // __clang__
+    else
+        return at<sizeof...(tail_t) + 1 + idx, head_t, tail_t...>();
+}
+
+/*!\brief Implementation for seqan3::pack_traits::front.
+ * \tparam head_t   Currently viewed pack_t element.
+ * \tparam tail_t   Rest of the type pack.
+ * \ingroup type_list
+ */
+template <typename head_t, typename ...tail_t>
+std::type_identity<head_t> front();
+
+/*!\brief Implementation for seqan3::pack_traits::drop_front.
+ * \tparam head_t   Currently viewed pack_t element.
+ * \tparam tail_t   Rest of the type pack.
+ * \ingroup type_list
+ */
+template <typename head_t, typename ...tail_t>
+type_list<tail_t...> drop_front();
+
+} // namespace seqan3::pack_traits::detail
+
+// ----------------------------------------------------------------------------
+// seqan3::pack_traits
+// ----------------------------------------------------------------------------
+
+namespace seqan3::pack_traits
+{
+
+/*!\name Type pack traits (return a value)
+ * \{
+ */
+
+/*!\brief The size of a type pack.
+ * \tparam pack_t The type pack.
+ * \returns `sizeof...(pack_t)`
+ * \ingroup type_list
+ *
+ * \details
+ *
+ * ### (Compile-time) Complexity
+ *
+ * Constant
+ */
+template <typename ...pack_t>
+inline constexpr size_t size = sizeof...(pack_t);
+
+//!\}
+
+/*!\name Type pack traits (return a single type)
+ * \{
+ */
+
+/*!\brief Return the type at given index from the type pack.
+ * \tparam idx    The index; must be smaller than the size of the pack.
+ * \tparam pack_t The type pack.
+ * \ingroup type_list
+ *
+ * \details
+ *
+ * Negative indexes are supported (e.g. `at<-1, int, double, bool &>` is `bool &`).
+ *
+ * ### (Compile-time) Complexity
+ *
+ * Equal to `idx` (linear).
+ */
+template <ptrdiff_t idx, typename ...pack_t>
+//!\cond
+    requires (idx >= 0 && idx < sizeof...(pack_t)) ||
+             (-idx <= sizeof...(pack_t))
+//!\endcond
+using at = typename decltype(detail::at<idx, pack_t...>())::type;
+
+/*!\brief Return the first type from the type pack.
+ * \tparam pack_t The type pack.
+ * \ingroup type_list
+ *
+ * \details
+ *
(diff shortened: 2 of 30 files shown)


include/seqan3/core/type_list.hpp
```diff
@@ -31,10 +35,485 @@ using type_list = meta::list<types...>;
...
+ *
+ * ### (Compile-time) Complexity
+ *
+ * Equal to `size<list_t> - n` (linear).
+ */
+template <ptrdiff_t n, seqan3::detail::TypeList list_t>
+//!\cond
+    requires (n >= 0 && n <= size<list_t>)
+//!\endcond
+using drop_last = typename decltype(detail::split_after_impl<size<list_t> - n>(type_list<>{}, list_t{}))::first_type;
```
rrahn: accordingly we can model this with the negative index and with take as well. I think it is more readable.

include/seqan3/core/type_list.hpp
```diff
@@ -31,10 +35,485 @@ using type_list = meta::list<types...>;
...
+ * \ingroup type_traits
+ *
+ * \details
+ *
+ * ### (Compile-time) Complexity
+ *
+ * Constant.
+ */
+template <typename ...pack_t>
+using concat = type_list<pack_t...>;
```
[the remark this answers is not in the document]
rrahn: Honestly, I don't think we need this at all. I mean that would be simply `type_list<t1, t2, t3, t4>`. Why bother to have a separate traits for this, which is basically just an alias?

include/seqan3/core/type_list.hpp
```diff
@@ -31,10 +35,485 @@ using type_list = meta::list<types...>;
...
+ *
+ * ### (Compile-time) Complexity
+ *
+ * Equal to `size<list_t> - n` (linear).
+ */
+template <ptrdiff_t n, seqan3::detail::TypeList list_t>
+//!\cond
+    requires (n >= 0 && n <= size<list_t>)
+//!\endcond
+using take_last = typename decltype(detail::split_after_impl<size<list_t> - n>(type_list<>{}, list_t{}))::second_type;
```
rrahn: First of all this would be the same as:
```suggestion
using take_last = drop<size<list_t> - n>;
```
right?
And I am wondering if this should not be modeled with negative indices`take<-n, list_t>`? 
That would it make consistent with `at`
So:

```c++
template <ptrdiff_t n, seqan3::detail::TypeList list_t>
//!\cond
    requires (n < 0 && -n <= size<list_t>)
//!\endcond
using take = drop<size<list_t> + n>;
```

test/unit/core/type_list_test.cpp
```diff
@@ -19,3 +20,167 @@ TEST(type_list, basic)
...
     EXPECT_TRUE((std::is_same_v<meta::at_c<t, 1>, char>));
 }
+
+// ----------------------------------------------------------------------------
+// pack tests
+// ----------------------------------------------------------------------------
+
+TEST(pack_traits, size)
+{
+    EXPECT_EQ(( pack_traits::size<int, bool &, double const> ),
```
rrahn:
```suggestion
    EXPECT_EQ((pack_traits::size<int, bool &, double const>),
```

test/unit/core/type_list_test.cpp
```diff
@@ -19,3 +20,167 @@ TEST(type_list, basic)
...
+
+TEST(pack_traits, back)
+{
+    EXPECT_TRUE((std::is_same_v<pack_traits::back<int, bool &, double const, long, float>,
+                                float>));
+}
+
+TEST(pack_traits, concat)
+{
+    EXPECT_TRUE((std::is_same_v<pack_traits::concat<int, bool &, double const, long, float>,
```
rrahn: I think we really don't need this. There is no added functionality by this.

include/seqan3/core/type_list.hpp
```diff
@@ -19,6 +19,10 @@
 namespace seqan3
 {
 
+// ----------------------------------------------------------------------------
+// type_list class
+// ----------------------------------------------------------------------------
```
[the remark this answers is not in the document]
h-2: Yes, I know.

include/seqan3/core/type_list.hpp
```diff
@@ -31,10 +35,485 @@ using type_list = meta::list<types...>;
...
+{
+
+/*!\brief Implementation for seqan3::pack_traits::at.
+ * \tparam idx      The index.
+ * \tparam head_t   Currently viewed pack_t element.
+ * \tparam tail_t   Rest of the parameter pack.
+ * \ingroup type_traits
+ */
+template <ptrdiff_t idx, typename head_t, typename ...tail_t>
+auto at_impl()
```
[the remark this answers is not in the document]
h-2: well they are different things (type vs functions...) so I wasn't sure. But I don't care much about it.

include/seqan3/core/type_list.hpp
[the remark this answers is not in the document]
h-2: Well I wanted to keep the interfaces for list VS types very similar. 

You might have
```cpp
template <typename list1, typename list2>
```
and want to go to `concat<list1, list2>`.

Similarly:
```cpp
template <typename ...pack1, typename ...pack2>
```
and go to `concat<pack1..., pack2...>` .

Of course you could just write `type_list` for the packs, but I think there is value in consistency...

include/seqan3/core/type_list.hpp
h-2: I had it like that before, but it's horrible :ghost: 

The problem is that it gets very confusing regarding inclusive/exclusive indexing.

take and drop are specifically not modeled as "take until third element", but as "take three elements". What would `take<-3>` imply? "Take until the third-last"? Before or after? What does it imply regarding the size of the list?

This way it is very simple: `take<n>` and `take_last <n>` always return something of size n. `drop<n>` and `drop_last <n>`  always return something of the original size - n.

range-v3 is also getting a `view::take_last` so we would be in-sync regarding naming.

include/seqan3/core/type_list.hpp
h-2: I agree, it can be written shorter. See above for why I am opposed to negative idnexing there. 

(Believe me, I wanted this, I designed it that way and it confused me so much I got convinced it wasn't a good idea)

test/unit/core/type_list_test.cpp
h-2: It's just an alias :man_shrugging:

include/seqan3/core/type_list.hpp
```diff
@@ -31,10 +35,485 @@ using type_list = meta::list<types...>;
...
+namespace seqan3::list_traits::detail
+{
+
+/*!\brief Implementation for seqan3::list_traits::at.
+ * \tparam idx      The index.
+ * \tparam pack_t   Types in the type list.
+ * \ingroup type_traits
+ */
+template <ptrdiff_t idx, typename ...pack_t>
+std::type_identity<seqan3::pack_traits::at<idx, pack_t...>> at_impl(type_list<pack_t...>);
```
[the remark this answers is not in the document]
h-2: I can change it if you want to.

include/seqan3/core/type_list.hpp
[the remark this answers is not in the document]
rrahn: But, there is no consistency with this, because 
```
template <typename ...pack1, typename ...pack2>
```
won't be possible if they are not already wrapped in some class. Which does not happen to often outside of the context of type_lists anyway. Also since they are in different namespaces you have to explicitly choose the traits anyway. Hence, there is no need for consistency. Either I am in the context of parameter packs or I am in the context of type_lists. Because then you could also argue that you could allow 
```
template <typename ...pack1, typename ...pack2, typename ...pack3>
...

concat<pack1..., pack2..., pack3....>
```
But this does not work with type_lists and would be more confusing then because of inconsistencies. Of course we could blow up the concat for type_lists but I think that it is rarely necessary (ok, let's see about that). So I'd like to keep these things separate and would argue that for parameter packs a concat is simply not necessary.

include/seqan3/core/type_list.hpp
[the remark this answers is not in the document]
rrahn: Yes, because otherwise it would deviate from the rest of the code base and kind of counteracts the meaning of the detail namespace.

include/seqan3/core/type_list.hpp
rrahn: Well, the negative index would perfectly align with take because take<-3> takes exactly the last three elements. That's how it is implemented anyway (`take_last<n> <=> drop<N-n>`). Thus you would have `take<N> == take<-N>` and `take<1> <=> at<0>` and `take<-1> <=> at<N-1> <=> at<-1>`. 
But if the ranges library get's a special take_last then why not having it here as well. But honestly, I don't like that  `at` has then a different semantics. But oh boy,  `at_last` would be the wrong way to go as well 🙈 .

test/unit/core/type_list_test.cpp
rrahn: but as I said above an alias that does allow more functionality than it's type list counterpart

include/seqan3/core/type_list.hpp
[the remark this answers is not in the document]
rrahn: but all impl stuff are now functions, aren't they?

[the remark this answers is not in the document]
h-2: Yes, but the public entity isn't. Anyway, I can change it.

include/seqan3/core/type_list.hpp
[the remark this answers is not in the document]
h-2: Not convinced fully, but I will remove, because it's really not that important.

include/seqan3/core/type_list.hpp
h-2: > Well, the negative index would perfectly align with take because take<-3> takes exactly the last three elements.

Oh, wow, ok, you thought it would work that way.

> But if the ranges library get's a special take_last then why not having it here as well. 

Yes, I really think it makes it clearer as we are not working with indexes at all, but with counts. And negative counts are weird.

h-2:
I think I have adressed all of your issues. I have also moved the stuff around and added a changelog.

include/seqan3/core/type_list/traits.hpp
```diff
@@ -0,0 +1,552 @@
...
+auto at()
+{
+    if constexpr (idx == 0)
+        return std::type_identity<head_t>{};
+    else if constexpr (idx > 0)
+#if __has_builtin(__type_pack_element)
+        return std::type_identity<__type_pack_element<idx - 1, tail_t...>>;
+#else
+        return at<idx - 1, tail_t...>();
+#endif
```
rrahn: can you mark with a comment that this is builtin ends here.

include/seqan3/core/type_list.hpp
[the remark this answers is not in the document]
h-2: added a card to Core project

include/seqan3/core/type_list.hpp
rrahn: > Oh, wow, ok, you thought it would work that way.

It should work that way if `at` works that way!

PR #1205, opened 2019-08-08
[FEATURE] view::async_buffer

h-2:
Enables async I/O even with multiple consumer threads.

Does not yet model View, because it is not Semiregular. I don't mind merging this now and adapting the tests when range-v3 is brought up to date.

TODO:

  - [x] add changelog


h-2:
failed macOS tests are due to #1206

include/seqan3/range/view/async_buffer.hpp
```diff
@@ -0,0 +1,513 @@
...
+ * Thread: 0x80116bf00     Seq:    seq12
+ * Thread: 0x80116ba00     Seq:    seq10
+ * ```
+ * This shows that indeed elements from the underlying range are processed non-sequentially, that there are two threads
+ * and that work is "balanced" between them (one thread processed more element than the other, because its "work"
+ * per item happened to be smaller).
+ *
+ * Note that you might encounter jumbled output if by chance two threads write to the stream at the exact same time.
+ *
+ * If you remove the line with `auto f1 = ...` you will get sequential processing:
```
eseiler:
```suggestion
 * If you remove the line starting with `auto f1 = ...` you will get sequential processing:
```

test/unit/range/view/view_async_buffer_test.cpp
```diff
@@ -0,0 +1,105 @@
...
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
+
+    } // thread sync at destruction of v1; tests working destruction with full buffer
+
+    // check that no over-consumption:
+    EXPECT_TRUE(std::ranges::equal(v0, "TATCGAGAGCTTTAGC"_dna4)); // total of 10 chars consumed
+    //TODO: actually 11 because queue buffers one more element that requested :o
+}
+
+//TODO: add combinability tests once the class models View.
```
eseiler: Can we also do a card for this?

test/unit/range/view/view_async_buffer_test.cpp
```diff
@@ -0,0 +1,105 @@
...
+        ++b; ++b; ++b; ++b;
+
+        // give time to rebuffer next five elements
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
+
+    } // thread sync at destruction of v1; tests working destruction with full buffer
+
+    // check that no over-consumption:
+    EXPECT_TRUE(std::ranges::equal(v0, "TATCGAGAGCTTTAGC"_dna4)); // total of 10 chars consumed
+    //TODO: actually 11 because queue buffers one more element that requested :o
```
eseiler: (Somehow didn't process my comment in the review)

😱 

Probably because of `auto tmp = std::move(*it);` in `produce()`?

eseiler: So we might need to check if the queue is not full and then consume+push; instead of consume and wait until push is successful.

include/seqan3/range/view/async_buffer.hpp
```diff
@@ -0,0 +1,513 @@
...
+        //!\brief Reference type.
+        using reference         = value_type;
+        //!\brief Iterator category.
+        using iterator_category = std::input_iterator_tag;
+        //!\}
+
+        /*!\name Construction, destruction and assignment
+        * \{
+        */
+        //!\brief Default constructor caches the firs value.
```
eseiler:
```suggestion
        //!\brief Default constructor caches the first value.
```

:)

include/seqan3/range/view/async_buffer.hpp
```diff
@@ -0,0 +1,513 @@
...
+
+    //!\brief Abort thread before it reaches the end.
+    std::atomic<bool> stop_producer{false};
+
+    /*!\brief An input_iterator over the associated range.
+    * \implements std::InputIterator
+    * \ingroup view
+    * \tparam view_type The type of the associated type.
+    *
+    * This iterator reduces every iterator type of the associated view to an single pass input iterator.
```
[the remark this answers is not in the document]
h-2: The text is a copy'n'paste survivor that needs to be changed anyway!

include/seqan3/range/view/async_buffer.hpp
```diff
@@ -0,0 +1,513 @@
...
+        using iterator_category = std::input_iterator_tag;
+        //!\}
+
+        /*!\name Construction, destruction and assignment
+        * \{
+        */
+        //!\brief Default constructor caches the firs value.
+        async_buffer_iterator()                                              noexcept = default;
+        //!\brief Copy construction.
+        async_buffer_iterator(async_buffer_iterator const & rhs)             noexcept = default; //TODO delete
```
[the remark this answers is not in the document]
h-2: ?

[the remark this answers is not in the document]
h-2: Cannot yet be deleted, because the constraints on InputIterator are not yet relaxed.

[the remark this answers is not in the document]
eseiler: Ah, ok

include/seqan3/range/view/async_buffer.hpp
```diff
@@ -0,0 +1,513 @@
...
+
+public:
+    /*!\name Constructor, destructor, and assignment.
+     * \{
+     * \brief All standard functions are explicitly defaulted.
+     */
+    //!\brief Default default-constructor.
+    async_buffer_view() = default;
+    //!\brief Default copy-constructor.
+    async_buffer_view(async_buffer_view const &) = default; //TODO delete
```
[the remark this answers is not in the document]
h-2: This can be removed, because the class currently doesnt model View anyway...

include/seqan3/range/view/async_buffer.hpp
```diff
@@ -0,0 +1,513 @@
...
+
+    //!\brief Iterates over the underlying range and pushes into the buffer until at end or stopped.
+    void produce()
+    {
+        auto it = std::ranges::begin(urange);
+        auto e  = std::ranges::end(urange);
+
+        detail::spin_delay delay{};
+
+        while ((it != e) && (!stop_producer.load()))// TODO likely
```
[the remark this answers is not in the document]
h-2: > So we stop reading if either we've read everything, or the view tells us to stop the producer thread, right?

Yes!

> Latter is needed if we do something lazily with the output of the async I/O and we want to stop before having read everything, correct?

It's needed when destructing the view, before all elements are moved into/through the buffer. Because the background thread is still running and we need to stop it.

It's possible to use wait_push instead which blocks when the queue is full. But the problem is that we get a dead-lock if 
  * the queue is full and produce is stuck in wait
  * nothing pops elements off
  * we destruct the view

include/seqan3/range/view/async_buffer.hpp
```diff
@@ -0,0 +1,513 @@
...
+// → async_buffer_view.buffer
+// → async_buffer_iterator.cached_value
+// → user
+//-----------------------------------------------------------------------------
+
+namespace seqan3::detail
+{
+
+/*!\brief Adds async_buffer behavior to the underlying range.
+ * \tparam urng_t The underlying range type.
```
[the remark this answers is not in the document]
h-2: It's asserted by the adaptor, but I can do it here, too...

test/unit/range/view/view_async_buffer_test.cpp
h-2: No, it's simply because of:
https://github.com/seqan/seqan3/blob/master/include/seqan3/contrib/parallel/buffer_queue.hpp#L75

@rrahn tried to explain to me why this is, but I didn't fully get it...

h-2: We have it all figured out!

test/unit/range/view/view_async_buffer_test.cpp
h-2: done

include/seqan3/range/view/async_buffer.hpp
```diff
@@ -0,0 +1,538 @@
...
+        //!\brief Iterator category.
+        using iterator_category = std::input_iterator_tag;
+        //!\}
+
+        /*!\name Construction, destruction and assignment
+        * \{
+        */
+        async_buffer_iterator()                                              noexcept = default; //!< Defaulted.
+        //TODO: delete:
+        async_buffer_iterator(async_buffer_iterator const & rhs)             noexcept = default; //!< Deleted.
```
eseiler: Mhm shouldn't it be `//!< Defaulted.` then until it is actually deleted?
Same with copy assignment.

include/seqan3/range/view/async_buffer.hpp
```diff
@@ -0,0 +1,538 @@
...
+        }
+        //!\}
+    };
+
+public:
+    /*!\name Constructor, destructor, and assignment.
+     * \{
+     */
+    async_buffer_view()                                         = default; //!< Defaulted.
+    async_buffer_view(async_buffer_view const &)                = delete;  //!< Deleted.
```
eseiler: Usually we wanted to write a short explanation as to why it is deleted instead of `Deleted.`.
E.g., `//!< This is a move-only type.` or `//!< The queue cannot be shared.` or something like this.

include/seqan3/range/view/async_buffer.hpp
```diff
@@ -0,0 +1,538 @@
...
+    async_buffer_view(async_buffer_view &&)                     = default; //!< Defaulted.
+    async_buffer_view & operator=(async_buffer_view const &)    = delete;  //!< Deleted.
+    async_buffer_view & operator=(async_buffer_view &&)         = default; //!< Defaulted.
+
+    /*!\brief Tell thread to stop and wait for it.
+     *
+     * \details
+     *
+     * Destruction of the view clears the buffer and temporary storage. Up to `n + 1` elements may be deleted
```
eseiler: Is deleted the correct word? 
When I have a ForwardRange and I read *deleted* I would kinda expect that it is really deleted - e.g. if I have a container, the size changes. But actually it's just *moved*, so the positions where the elements where are now still accessible and contain valid, but undefined stuff, right?

include/seqan3/range/view/async_buffer.hpp
```diff
@@ -0,0 +1,538 @@
...
+     * iterators that are each valid individually. It is thread-safe to operate on different iterators
+     * from different threads (however it is not thread-safe to operate on a single iterator from different
+     * threads).
+     */
+    async_buffer_iterator begin()
+    {
+        return {buffer};
+    }
+
+    //!\brief Const version of begin is deleted, since the underlying view_state must be mutable.
```
eseiler:
```suggestion
    //!\brief Const version of begin is deleted since the underlying view_state must be mutable.
```
(also for end).

Does begin link correctly?

include/seqan3/range/view/async_buffer.hpp
```diff
@@ -0,0 +1,538 @@
...
+/*!\name General purpose views
+ * \{
+ */
+
+/*!\brief A view adapter that returns a concurrent-queue-like view over the underlying range.
+ * \tparam urng_t         The type of the range being processed. See below for requirements.
+ * \param[in,out] urange  The range being processed.
+ * \param[in] buffer_size Size of the buffer. Choose the size depending on the expected work per element.
+ * \returns A view that pre-fetches elements from the underlying range and provides a thread-safe interface.
+ * See below for the properties of the returned range.
```
eseiler: I'd indent this

include/seqan3/range/view/async_buffer.hpp
h-2: Hm, yeah, I don't care!

include/seqan3/range/view/async_buffer.hpp
h-2: I will introduce a policy later that says that all single-passt input iterators and single-pass views are non-copyable. But I can also put it in a comment.

include/seqan3/range/view/async_buffer.hpp
h-2: Well, yes,  "moved-from objects shall be placed in a valid but unspecified state."

The elements are destructed, though. I can write that instead of deleted.

include/seqan3/range/view/async_buffer.hpp
h-2: Most definitely not. In my defence: I copy'n'pasted this from @rrahn :wink:

include/seqan3/range/view/async_buffer.hpp
```diff
@@ -0,0 +1,538 @@
...
+     * \details
+     *
+     * Destruction of the view clears the buffer and temporary storage. Up to `n + 1` elements may be destructed
+     * together with the view. [This may be important if the underlying range is also single-pass and you
+     * are trying to estimate how many elements will be consumed – in case you do not plan on consuming the
+     * entire range.]
+     */
+    ~async_buffer_view()
+    {
+        stop_producer.store(true);
```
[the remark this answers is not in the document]
rrahn: You also need to wait until the queue is empty after it was closed, otherwise you get a data race with destruction of the queue and consumer threads. If not you would need to add some inter-thread signaling, which is kind of complex in this model.

include/seqan3/range/view/async_buffer.hpp
```diff
@@ -0,0 +1,538 @@
...
+ * ### Range consumption
+ *
+ * This view always moves elements from the underlying range into its buffer which means that the elements in
+ * the underlying range will be invalidated! For underlying ranges that are single-pass, this makes no difference, but
+ * it might be unexpected for multi-pass ranges (std::ranges::ForwardRange).
+ *
+ * Typically this adaptor is used when you want to consume the entire underlying range; there is no way to stop
+ * this view from buffering elements before the end of the underlying range is reached other than destructing it.
+ * If it is destructed before the end of the underlying range is reached, it will stop consuming elements, however,
+ * those elements that are currently in it's buffer are destructed with it.
```
rrahn: As my comment in the destructor suggests that would  be a data race. So after fixing this you might want to update the comment here as well.

include/seqan3/range/view/async_buffer.hpp
```diff
@@ -0,0 +1,538 @@
...
+ *
+ * Note that in addition to the buffer of the view, every iterator has its own one-element-buffer. Dereferencing
+ * the iterator moves the element out of that buffer and incrementing it refills the buffer from the queue inside the
```
rrahn: Like I said, not sure if the * is allowed to move because of invalidating equality-preservation.

test/snippet/range/view/async_buffer.cpp
```diff
@@ -0,0 +1,67 @@
+#include <cstdlib>                              // std::rand
+#include <future>                               // std::async
+#include <string>                               // std::string
+#include <seqan3/core/debug_stream.hpp>         // seqan3::debug_stream
```
rrahn: empty line before seqan3 header

test/snippet/range/view/async_buffer.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include <seqan3/io/sequence_file/input.hpp>    // seqan3::sequence_file_input
+#include <seqan3/range/view/async_buffer.hpp>   // seqan3::view::async_buffer
```
rrahn: Please remove the comments. At least from the very obvious includes

test/snippet/range/view/async_buffer.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+    // create an input file from the string above
+    seqan3::sequence_file_input fin{std::istringstream{fasta_file}, seqan3::format_fasta{}};
+
+    // create the async buffer around the input file
+    // spawns a background thread that tries to keep four records in the buffer
+    auto v = fin | seqan3::view::async_buffer(4);
+
+    // create a lambda function that iterates over the async buffer when called
+    // (the buffer gets dynamically refilled as soon as empty)
```
rrahn: the buffer get's refilled constantly?!

test/snippet/range/view/async_buffer.cpp
```diff
@@ -0,0 +1,67 @@
...
+    auto l = [&v] ()
```
rrahn: `l` is very bad name. Please give it a more meaningful name like `print_sequence` or something.

test/snippet/range/view/async_buffer.cpp
```diff
@@ -0,0 +1,67 @@
...
+{
+    // initialise random number generator, only needed for demonstration purposes
+    std::srand(std::time(nullptr));
+
+    // create an input file from the string above
+    seqan3::sequence_file_input fin{std::istringstream{fasta_file}, seqan3::format_fasta{}};
+
+    // create the async buffer around the input file
+    // spawns a background thread that tries to keep four records in the buffer
+    auto v = fin | seqan3::view::async_buffer(4);
```
rrahn:
```suggestion
    auto async_fin_view = fin | seqan3::view::async_buffer(4);
```

test/snippet/range/view/async_buffer.cpp
```diff
@@ -0,0 +1,67 @@
...
+    // create a lambda function that iterates over the async buffer when called
+    // (the buffer gets dynamically refilled as soon as empty)
+    auto l = [&v] ()
+    {
+        for (auto r : v)
+        {
+            // pretend we are doing some work
+            std::this_thread::sleep_for(std::chrono::milliseconds(std::rand() % 1000));
+            // print current thread and sequence ID
+            seqan3::debug_stream << "Thread: " << std::this_thread::get_id() << '\t'
```
rrahn: why don't you align the last `<<` operator 😏

test/unit/range/view/view_async_buffer_test.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+    std::vector<seqan3::dna4> vec{"ACGTACGTACGTATCGAGAGCTTTAGC"_dna4};
+    std::vector<seqan3::dna4> expected_range{"ACGTACGTACGTATCGAGAGCTTTAGC"_dna4};
+    decltype(view::async_buffer(vec, 3)) test_range = view::async_buffer(vec, 3);
+};
+
+using test_type = ::testing::Types<iterator_type>;
+INSTANTIATE_TYPED_TEST_CASE_P(iterator_fixture, iterator_fixture, test_type);
+
+
```
rrahn: double new line

test/unit/range/view/view_async_buffer_test.cpp
```diff
@@ -0,0 +1,106 @@
...
+    std::vector<seqan3::dna4> vec{"ACGTACGTACGTATCGAGAGCTTTAGC"_dna4};
+
+    auto v = vec | view::async_buffer(3);
+
+    EXPECT_TRUE(std::ranges::equal(vec, v));
+}
+
+TEST(async_buffer, buffer_size_zero)
+{
+    std::vector<seqan3::dna4> vec{"ACGTACGTACGTATCGAGAGCTTTAGC"_dna4};
```
rrahn: please outsource into a test fixture.

test/unit/range/view/view_async_buffer_test.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+    {
+        auto v1 = v0 | view::async_buffer(5);
+
+        // consume five elements (construction already consumes one)
+        auto b = std::ranges::begin(v1);
+        ++b; ++b; ++b; ++b;
+
+        // give time to rebuffer next five elements
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
rrahn: this is a bad test, since you have no guarantee that the producer thread will rebuffer in this time. It could be yielded and come alive first after the 100 milliseconds or within. I am afraid this can't be solved easily.

test/unit/range/view/view_async_buffer_test.cpp
```diff
@@ -0,0 +1,106 @@
...
+    auto v1 = vec | view::async_buffer(1);
+
+    EXPECT_TRUE(std::ranges::InputRange<decltype(v1)>);
+    EXPECT_FALSE(std::ranges::ForwardRange<decltype(v1)>);
+    EXPECT_FALSE(std::ranges::RandomAccessRange<decltype(v1)>);
+    EXPECT_FALSE(std::ranges::SizedRange<decltype(v1)>);
+    EXPECT_FALSE(ConstIterableRange<decltype(v1)>);
+
+    EXPECT_FALSE(std::Semiregular<decltype(v1)>);
+    EXPECT_FALSE(std::ranges::View<decltype(v1)>);          // ← should become true once View is relaxed
```
rrahn: 🙈

test/unit/range/view/view_async_buffer_test.cpp
h-2: A test fixture for a single line? Is that work the added complexity?

test/unit/range/view/view_async_buffer_test.cpp
h-2: I need to test exactly that behaviour. I know behaviour is not guaranteed, but I think it's fairly reasonable to expect this to work in 99.9999999% of the cases.

I don't see another way of testing this, short of adding undocumented public accessors to the view that expose the internal "fill-state". That would be **very ugly**.

test/unit/range/view/view_async_buffer_test.cpp
h-2: :tada:

include/seqan3/range/view/async_buffer.hpp
```diff
@@ -0,0 +1,538 @@
...
+    }
+
+    //!\brief Thread that runs #produce in the background.
+    std::thread producer;
+
+    //!\brief Abort thread before it reaches the end.
+    std::atomic<bool> stop_producer{false};
+
+    //!\brief The iterator of the view.
+    class async_buffer_iterator
```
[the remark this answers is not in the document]
rrahn: Well, yes, I know. So did I. But especially when the iterator get's bloated and or complex it is nicer to separate it from the wrapping class. It helps to focus on the things relevant for the wrapping class. We should discuss this.

include/seqan3/range/view/async_buffer.hpp
```diff
@@ -0,0 +1,538 @@
...
+
+    //!\brief Construction from std::ranges::ViewableRange.
+    template <typename other_urng_t>
+    //!\cond
+    requires !std::Same<remove_cvref_t<other_urng_t>, async_buffer_view> && // prevent recursive instantiation
+             std::ranges::ViewableRange<other_urng_t> &&
+             std::Constructible<urng_t, ranges::ref_view<std::remove_reference_t<other_urng_t>>>
+    //!\endcond
+    async_buffer_view(other_urng_t && _urng, size_t const buffer_size) :
+        async_buffer_view{std::view::all(_urng), buffer_size}
```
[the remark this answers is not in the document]
rrahn: yes?

include/seqan3/range/view/async_buffer.hpp
rrahn: 😱 No this is a very bad design. If you want to allow that consumers are not consuming everything you need a different signaling mechanism, i.e. consumers signal that they are done reading. The simplest way to do it is by using a barrier then. In that case you need to specify in the construction of the view how many consumers you are expecting and the destructor needs to wait for the consumers to signal they are done (arrived at the barrier managed by the view). Then you can safely destruct the queue even if it is not empy after closing since you are sure that no one is reading from it anymore. I need to think about how you can make it work nicely with the iterator.

test/snippet/range/view/async_buffer.cpp
rrahn: why????

test/snippet/range/view/async_buffer.cpp
rrahn: Well, yes we started this ugliness but `auto r : v` is really bad to read. But I won't insist on this now.

test/unit/range/view/view_async_buffer_test.cpp
rrahn: well don't do it if you think it is not worth it. I would change it though.

test/unit/range/view/view_async_buffer_test.cpp
rrahn: Well, I'd like you to remember the weird scenarios we had with parallel tests we assumed they were correct and still caused a lot of trouble. Now you arguing, a test that is wrong by design is ok since it presumably happens not very often? I don't understand the reasoning. As said before you need to handle it differently. In this case I guess what you need is a barrier and on destruction of the iterator you arrive on the barrier within the buffer. If all iterators arrived (destructed) you can safely destruct the buffer after closing it and waiting for the producer thread. That is the only way to prevent a data race.

include/seqan3/range/view/async_buffer.hpp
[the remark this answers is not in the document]
h-2: The current solution is confusing and using our view all will cause problems with some constructors (those looking for ref_view).
I will rename our seqan3::view::all to seqan3::view::simplifiy_type. We will not use it automatically, but recommend that people use it at the beginning of complex pipes.

h-2:
I have adapted everything according to our discussion.

> most of the minor comments where annotated but actually not fixed. Please check this again.

Please review your comments and mark resolved issues as such so that I know which ones you are referring to. (Some things have been changed outside the immediate context so it may appear as there was no change when there was).

include/seqan3/range/view/async_buffer.hpp
```diff
@@ -0,0 +1,538 @@
...
+    std::thread producer;
+
+    //!\brief Abort thread before it reaches the end.
+    std::atomic<bool> stop_producer{false};
+
+    //!\brief The iterator of the view.
+    class async_buffer_iterator
+    {
+        //!\brief The sentinel type to compare to.
+        using sentinel_type      = std::ranges::default_sentinel_t;
```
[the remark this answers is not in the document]
rrahn: still open

include/seqan3/range/view/async_buffer.hpp
```diff
@@ -0,0 +1,538 @@
...
+        */
+        //!\brief Pre-increment.
+        async_buffer_iterator & operator++() noexcept
+        {
+            if (at_end) // TODO unlikely
+                return *this;
+
+            detail::spin_delay delay{};
+
+            assert (buffer_ptr != nullptr);
```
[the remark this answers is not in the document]
rrahn: still open

include/seqan3/range/view/async_buffer.hpp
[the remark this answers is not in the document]
rrahn: > The current solution is confusing 

I agree, but it is nothing we discussed yet. So please make a point for the discussion board

> those looking for ref_view). 

😱

include/seqan3/range/view/async_buffer.hpp
```diff
@@ -0,0 +1,481 @@
...
+ * the file contains genome-sized sequences, it would be better to buffer only a single sequence (buffering 100
+ * sequences would result in the entire file being preloaded and likely consuming significant memory).
+ *
+ * ### Range consumption
+ *
+ * This view always moves elements from the underlying range into its buffer which means that the elements in
+ * the underlying range will be invalidated! For underlying ranges that are single-pass, this makes no difference, but
+ * it might be unexpected for multi-pass ranges (std::ranges::ForwardRange).
+ *
+ * Typically this adaptor is used when you want to consume the entire underlying range. While destructing
```
rrahn:
```suggestion
 * Typically this adaptor is used when you want to consume the entire underlying range. Destructing
```

include/seqan3/range/view/async_buffer.hpp
```diff
@@ -0,0 +1,481 @@
...
+ * this view before all elements have been read will also stop the thread that moves object from the underlying
+ * range, it is difficult to assess which elements in the underlying range might still be valid.
```
rrahn:
```suggestion
 * range.
```

include/seqan3/range/view/async_buffer.hpp
```diff
@@ -0,0 +1,481 @@
...
+ * **In general, assume that it is not safe to access the underlying range separately once it has been passed
```
rrahn:
```suggestion
 * **In general, safe access to the underlying range in any other context is not guaranteed and once it has been passed
```

include/seqan3/range/view/async_buffer.hpp
```diff
@@ -0,0 +1,481 @@
...
+ * to seqan3::view::async_buffer.**
+ *
+ * Note that in addition to the buffer of the view, every iterator has its own one-element-buffer. Dereferencing
+ * the iterator returns a reference to the element in the buffer, usually you will want to move this element out
+ * of the buffer with std::move. Incrementing the iterator refills the buffer from the queue inside the
```
rrahn:
```suggestion
 * of the buffer with std::move or std::iter_move. Incrementing the iterator refills the buffer from the queue inside the
```

test/snippet/range/view/async_buffer.cpp
rrahn: Still why??????????

test/unit/range/view/view_async_buffer_test.cpp
rrahn: still open

test/unit/range/view/view_async_buffer_test.cpp
```diff
@@ -0,0 +1,104 @@
...
+        // consume five elements (construction already consumes one)
+        auto b = std::ranges::begin(v1);
+        ++b; ++b; ++b; ++b;
+
+        // give time to rebuffer next five elements, not a problem if not successful
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
+
+    } // thread sync at destruction of v1; tests working destruction with full buffer
+
+    EXPECT_GE(std::ranges::distance(v0), 17); // total of at most 10 chars consumed
```
rrahn: So this means we could basically remove the sleep command right

h-2: I want to check that construction does not dead-lock when the buffer is not empty. The current code has a very high chance of testing this important behaviour. Removing the sleep means we have a very high change of *not* testing this behaviour.

h-2:
Combinability is broken, don't merge this yet.

h-2:
I have addressed the last issues, I think. I even moved the iterator definition out :man_cartwheeling:

h-2:
Fixed the test failure.

test/snippet/range/view/async_buffer.cpp
rrahn: then do_work would be good?

include/seqan3/range/detail/inherited_iterator_base.hpp
```diff
@@ -183,11 +188,17 @@ class inherited_iterator_base : public std::conditional_t<std::is_pointer_v<base
         return *this_derived();
     }
 
+    //!\brief Post-increment of non-copyable iterators returns void.
+    constexpr void operator++(int) noexcept(noexcept(++std::declval<derived_t &>()))
+    {
+        ++(*this_derived());
```
rrahn: Well, this is only true for input iterator right. So since this is a base class to inherit from input iterators might never fulfill the Cpp17IteratorConcept. Not sure, that we want this for the generic base class. So Cpp17Iterator returns a copy of the dereferenced element  before the increment. Have you considered this?

include/seqan3/range/detail/inherited_iterator_base.hpp
```diff
@@ -294,16 +305,34 @@ class inherited_iterator_base : public std::conditional_t<std::is_pointer_v<base
...
     }
 
     //!\brief Return pointer to this iterator.
-    constexpr pointer operator->() const noexcept(noexcept(*std::declval<base_t &>()))
+    constexpr pointer operator->() noexcept(noexcept(*std::declval<base_t &>()))
+    //!\cond
+        requires std::InputIterator<base_t>
+    //!\endcond
+    {
+        return &*this_to_base();
```
rrahn: don't get me wrong, but shouldn't it be `&**this_to_base()` 🤔
I mean you want to return a pointer to the referenced element not to the iterator itself. Why is the compiler not complaining? My head is going 💥

include/seqan3/range/detail/inherited_iterator_base.hpp
```diff
@@ -312,19 +341,29 @@ class inherited_iterator_base : public std::conditional_t<std::is_pointer_v<base
     }
 
     //!\brief Return underlying container value currently pointed at.
-    constexpr decltype(auto) operator[](std::make_signed_t<difference_type> const n) const
+    constexpr decltype(auto) operator[](std::make_signed_t<difference_type> const n)
```
rrahn: why is the difference type unsigned? That would not be valid per concept defintion as it has to be signed_integer_like

include/seqan3/range/view/async_buffer.hpp
```diff
@@ -0,0 +1,499 @@
...
+        if (buffer_ptr->wait_pop(cached_value) == contrib::queue_op_status::closed)
+            at_end = true;
+
+        return *this;
+    }
+
+    //!\brief Post-increment.
+    void operator++(int) noexcept
+    {
+        ++(*this);
```
rrahn: so this iterator becomes no Cpp17Iterator. In case this matters.

include/seqan3/range/detail/inherited_iterator_base.hpp
h-2: The version you are talking about is immediately below it. It is more refined than this overload because it requires Semiregular. Since all ForwardIterators are semirgeular, it will use the other overload. But it will even use the other one if the iterator is input-iterator and semiregular which is what we want!

include/seqan3/range/detail/inherited_iterator_base.hpp
h-2: the arrow operator induces itself recursively :sunglasses: 

> If a user-defined operator-> is provided, the operator-> is called again on the value that it returns, recursively, until an operator-> is reached that returns a plain pointer. After that, built-in semantics are applied to that pointer.

include/seqan3/range/detail/inherited_iterator_base.hpp
h-2: > why is the difference type unsigned?

Is it?

include/seqan3/range/view/async_buffer.hpp
h-2: It's not required: https://en.cppreference.com/w/cpp/named_req/InputIterator

include/seqan3/range/detail/inherited_iterator_base.hpp
rrahn: yes

rrahn: I don't know, but why is it wrapped in make_signed?

include/seqan3/range/view/async_buffer.hpp
rrahn: > value_type x = *i;
++i;
return x;

> *i++ -> convertible to value_type

rrahn: And further down:
>   typename std::common_reference_t<decltype(*i++)&&, typename std::readable_traits<I>::value_type&>;

include/seqan3/range/detail/inherited_iterator_base.hpp
rrahn: wow, I didn't know

include/seqan3/range/detail/inherited_iterator_base.hpp
rrahn: ok

include/seqan3/range/detail/inherited_iterator_base.hpp
h-2: It resolves to the underlying difference_type. Apparently we inherited something somewhere that was unsigned. I don't recall entirely. It's not new code so I would pref to leave it like this. I will revisit this file when we have move-only iterators after the next range upgrade.

include/seqan3/range/view/async_buffer.hpp
h-2: You are correct, but we wouldn't be able to satisfy "equivalent expression" in any case, right?

include/seqan3/range/detail/inherited_iterator_base.hpp
rrahn: ok

include/seqan3/range/view/async_buffer.hpp
rrahn: so we leave it as no CPP17Iterator. You might want to add then a sentence to the view description that the returned iterator does not model Cpp17InputIterator. After this we can merge it.

test/unit/range/view/view_async_input_buffer_test.cpp
```diff
@@ -0,0 +1,116 @@
...
+        auto b = std::ranges::begin(v1);
+        ++b; ++b; ++b; ++b;
+
+        /* Give time to rebuffer next five elements so the queue will not be empty.
+         * This is not required for this test to be successful, but it is the only
+         * way destruction with non-empty buffer is at least likely to happen.
+         * And we want it to happen to make sure we don't dead-lock on it.
+         */
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
+
```
rrahn: empty line

rrahn:
CI is broken

test/unit/range/view/view_async_input_buffer_test.cpp
```diff
@@ -0,0 +1,115 @@
...
+    auto v1 = vec | view::async_input_buffer(1);
+
+    EXPECT_TRUE(std::ranges::InputRange<decltype(v1)>);
+    EXPECT_FALSE(std::ranges::ForwardRange<decltype(v1)>);
+    EXPECT_FALSE(std::ranges::RandomAccessRange<decltype(v1)>);
+    EXPECT_FALSE(std::ranges::SizedRange<decltype(v1)>);
+    EXPECT_FALSE(ConstIterableRange<decltype(v1)>);
+    EXPECT_TRUE(std::ranges::View<decltype(v1)>);
+}
```
rrahn: maybe add one more test with an empty range.

h-2:
Jenkins failure unrelated.

PR #1208, opened 2019-08-12
[FIX] Add data access for containers

eseiler:
* Adds `raw_data()` for `bitcompressed_vector` and allows access of the underlying SDSL vector.
* Renames `data()` to `raw_data()` in `concatenated_sequences` since `data()` implies contiguous memory locations.
* `small_vector` and `small_string` have `data()` and are contiguous, ergo no changes here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,11 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 * **The `type_list` header has moved:**
   If you included `<seqan3/core/type_list.hpp>` you need to change the path to `<seqan3/core/type_list/type_list.hpp>`.
 
+#### Range
+
+* **The `seqan3::concatenated_sequences::data()` function has been deprecated:**
+  Use `seqan3::concatenated_sequences::raw_data()` instead.
+
 ## Notable Bug-fixes
 
 # 3.0.0 ("Escala")
--- a/include/seqan3/range/container/bitcompressed_vector.hpp
+++ b/include/seqan3/range/container/bitcompressed_vector.hpp
@@ -518,6 +518,27 @@ public:
         return (*this)[size()-1];
     }
 
+    /*!\brief Provides direct, unsafe access to underlying data structures.
+     * \returns A reference to an SDSL bitvector.
+     *
+     * \details
+     *
+     * \noapi
+     *
+     * The exact representation of the data is implementation defined. Do not rely on it for API stability.
+     */
+    constexpr data_type & raw_data() noexcept
+    {
+        return data;
+    }
+
+    //!\copydoc raw_data()
+    constexpr data_type const & raw_data() const noexcept
+    {
+        return data;
+    }
+    //!\}
+
     /*!\name Capacity
      * \{
      */
@@ -1008,7 +1029,7 @@ public:
     template <CerealArchive archive_t>
     void CEREAL_SERIALIZE_FUNCTION_NAME(archive_t & archive)
     {
-        archive(data); //TODO: data not yet serialisable
+        archive(data);
     }
     //!\endcond
 };
--- a/include/seqan3/range/container/concatenated_sequences.hpp
+++ b/include/seqan3/range/container/concatenated_sequences.hpp
@@ -53,7 +53,7 @@ namespace seqan3
  * * Better cache locality when parsing the sequences linearly (and often also on random access).
  * * Constant time access to the concatenation of the sequences via concat().
  * * This access is also writable so that certain transformations can be done globally, instead of element-wise.
- * * Also direct access to the delimiters via data() [this is used by some algorithms].
+ * * Also direct access to the delimiters via raw_data() [this is used by some algorithms].
  *
  * The disadvantages are:
  *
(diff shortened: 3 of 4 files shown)


svnbgnk:
The operator compare all the bitcompressed data from one vector to another, right?
https://github.com/seqan/seqan3/blob/25d06a0aeb0e775a9c8c44ca8c93848aef59a127/include/seqan3/range/container/bitcompressed_vector.hpp#L978

test/unit/range/container/container_of_container_test.cpp
```diff
@@ -152,10 +152,10 @@ TYPED_TEST(container_of_container, element_access)
         EXPECT_EQ(dna4_vector(t2.concat()), "ACGTACGTGAGGA"_dna4);
 
         // data
-        EXPECT_EQ(std::get<0>(t1.data()), "ACGTACGTGAGGA"_dna4);
-        EXPECT_EQ(std::get<0>(t2.data()), "ACGTACGTGAGGA"_dna4);
-        EXPECT_EQ(std::get<1>(t1.data()), (std::vector<size_type>{0, 4, 8, 13}));
-        EXPECT_EQ(std::get<1>(t2.data()), (std::vector<size_type>{0, 4, 8, 13}));
+        EXPECT_EQ(std::get<0>(t1.raw_data()), "ACGTACGTGAGGA"_dna4);
```
svnbgnk: t1.raw_data() return a pair. I did not know that you can also use std::get(). IMO this is a bit confusing.

eseiler:
> The operators compare all the bitcompressed data from one vector to another, right?
> https://github.com/seqan/seqan3/blob/25d06a0aeb0e775a9c8c44ca8c93848aef59a127/include/seqan3/range/container/bitcompressed_vector.hpp#L978-L981

We delegate to the compairson operator of the underlying data type (sdsl::int_vector), which then compares the content.

eseiler:
> One thing I noticed that did not data with raw_data() swap:
> https://github.com/seqan/seqan3/blob/25d06a0aeb0e775a9c8c44ca8c93848aef59a127/include/seqan3/range/container/bitcompressed_vector.hpp#L937-L941
> 
> .
> Is there any reason for that?

No reason, just didn't see it. I can access the private `data` member within the class, but I'll change it for consistency.

test/unit/range/container/container_of_container_test.cpp
eseiler: `std::get` is usually how you access data of pair/tuple.
Also see https://en.cppreference.com/w/cpp/utility/pair

include/seqan3/range/container/bitcompressed_vector.hpp
```diff
@@ -518,6 +518,19 @@ class bitcompressed_vector
         return (*this)[size()-1];
     }
 
+    //!\brief Direct access to the underlying SDSL vector.
```
h-2: Maybe you should describe more clearly what the type will be as this is exposed to the API. Note that if we expose this to the API, we cannot change the internal type anymore without breakage.

include/seqan3/range/container/bitcompressed_vector.hpp
```diff
@@ -964,37 +977,37 @@ class bitcompressed_vector
     //!\brief Checks whether `*this` is equal to `rhs`.
     constexpr bool operator==(bitcompressed_vector const & rhs) const noexcept
     {
-        return data == rhs.data;
+        return raw_data() == rhs.raw_data();
```
h-2: Are these changes necessary? I don't care much ....

include/seqan3/range/container/concatenated_sequences.hpp
```diff
@@ -602,16 +602,30 @@ class concatenated_sequences
...
+    {
+        return raw_data();
+    }
+
+    //!\copydoc raw_data()
+    //!\deprecated Use raw_data() instead.
+    SEQAN3_DEPRECATED_310 std::pair<decltype(data_values) const &, decltype(data_delimiters) const &> data() const
+    {
+        return raw_data();
+    }
```
h-2: :+1:

include/seqan3/range/container/bitcompressed_vector.hpp
eseiler: I changed it a bit. I'd like to link to a documentation of the bitvector. But there are no online docs for the sdsl3 bitvector. And the sdsl2 one looks like [this](https://algo2.iti.kit.edu/gog/docs/html/namespacesdsl.html#a93df8f263fced9ae9a8b0f9f37e444eb).

Anyway, I noticed that in `concatenated_sequences` we added this sentence:
`This exact representation of the data is implementation defined. Do not rely on it for API stability.`
So it's basically not API and we could add `\noapi` for doxygen to this function?

Then we could discuss if all our container `data`/`raw_data` functions are no api.

include/seqan3/range/container/bitcompressed_vector.hpp
eseiler: Mhm, I was afraid of the code coverage since I didn't want to test the `data`/`raw_data` functions.

I'll revert it and see what happens.

include/seqan3/range/container/bitcompressed_vector.hpp
h-2: > So it's basically not API and we could add \noapi for doxygen to this function?

:+1:

> Then we could discuss if all our container data/raw_data functions are no api.

:+1:

CHANGELOG.md
```diff
@@ -40,6 +40,10 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 * **The `type_list` header has moved:**
   If you included `<seqan3/core/type_list.hpp>` you need to change the path to `<seqan3/core/type_list/type_list.hpp>`.
 
+### The `data()` function of `seqan3::concatenated_sequences` has been deprecated
+
+Use `raw_data()` instead.
```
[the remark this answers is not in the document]
eseiler: If we mark the functions as noapi we don't need to include it in the changelog?

PR #1210, opened 2019-08-14
[FIX] minor fix to view::take that makes it common and const-iterable…

h-2:
… more often

### DIFF
--- a/include/seqan3/range/view/take.hpp
+++ b/include/seqan3/range/view/take.hpp
@@ -281,7 +281,6 @@ private:
         //!\}
     }; // class iterator_type
 
-public:
     /*!\name Associated types
      * \{
      */
@@ -303,6 +302,7 @@ public:
     using const_iterator    = detail::transformation_trait_or_t<std::type_identity<iterator_type<urng_t const>>, void>;
     //!\}
 
+public:
     /*!\name Constructors, destructor and assignment
      * \{
      */
@@ -321,12 +321,22 @@ public:
     constexpr view_take(urng_t _urange, size_t const _size)
         : urange{std::move(_urange)}, target_size{_size}
     {
-        if constexpr (exactly && or_throw && std::ranges::SizedRange<urng_t>)
+        if constexpr (std::ranges::SizedRange<urng_t>)
         {
-            if (seqan3::size(_urange) < _size)
+            if (std::ranges::size(urange) < target_size)
             {
-                throw std::invalid_argument{
-                    "You are trying to construct a view::take_exactly_or_throw from a range that is strictly smaller."};
+                if constexpr (exactly && or_throw)
+                {
+                    throw std::invalid_argument
+                    {
+                        "You are trying to construct a view::take_exactly_or_throw from a range that is strictly "
+                        "smaller."
+                    };
+                }
+                else
+                {
+                    target_size = std::ranges::size(urange);
+                }
             }
         }
     }
@@ -362,23 +372,32 @@ public:
      *
      * No-throw guarantee.
      */
-    constexpr iterator begin() noexcept
+    constexpr auto begin() noexcept
     {
-        return {seqan3::begin(urange), 0, target_size, &(*this)};
+        if constexpr (std::ranges::RandomAccessRange<urng_t> && std::ranges::SizedRange<urng_t>)
+            return std::ranges::begin(urange);
+        else
+            return iterator{std::ranges::begin(urange), 0, target_size, this};
     }
 
     //!\copydoc begin()
-    constexpr const_iterator begin() const noexcept
+    constexpr auto begin() const noexcept
         requires ConstIterableRange<urng_t>
     {
-        return {seqan3::cbegin(urange), 0, target_size};
+        if constexpr (std::ranges::RandomAccessRange<urng_t> && std::ranges::SizedRange<urng_t>)
+            return std::ranges::cbegin(urange);
+        else
+            return const_iterator{std::ranges::cbegin(urange), 0, target_size};
     }
 
     //!\copydoc begin()
-    constexpr const_iterator cbegin() const noexcept
+    constexpr auto cbegin() const noexcept
         requires ConstIterableRange<urng_t>
     {
-        return {seqan3::cbegin(urange), 0, target_size};
+        if constexpr (std::ranges::RandomAccessRange<urng_t> && std::ranges::SizedRange<urng_t>)
+            return std::ranges::cbegin(urange);
+        else
+            return const_iterator{std::ranges::cbegin(urange), 0, target_size};
     }
 
     /*!\brief Returns an iterator to the element following the last element of the range.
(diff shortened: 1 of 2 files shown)


include/seqan3/range/view/take.hpp
```diff
@@ -321,12 +321,19 @@ class view_take : public ranges::view_interface<view_take<urng_t, exactly, or_th
...
         {
-            if (seqan3::size(_urange) < _size)
+            if (std::ranges::size(_urange) < _size)
             {
-                throw std::invalid_argument{
-                    "You are trying to construct a view::take_exactly_or_throw from a range that is strictly smaller."};
+                if constexpr (exactly && or_throw)
+                {
+                    throw std::invalid_argument{
+                        "You are trying to construct a view::take_exactly_or_throw from a range that is strictly smaller."};
```
eseiler: Line length

include/seqan3/range/view/take.hpp
```diff
@@ -321,12 +321,22 @@ class view_take : public ranges::view_interface<view_take<urng_t, exactly, or_th
     constexpr view_take(urng_t _urange, size_t const _size)
         : urange{std::move(_urange)}, target_size{_size}
     {
-        if constexpr (exactly && or_throw && std::ranges::SizedRange<urng_t>)
+        if constexpr (std::ranges::SizedRange<urng_t>)
         {
-            if (seqan3::size(_urange) < _size)
+            if (std::ranges::size(_urange) < _size)
```
rrahn: you can't use `_urange` here after you moved it into `urange`.

rrahn: please use in general the classes variables instead of the passed arguments.

include/seqan3/range/view/take.hpp
```diff
@@ -321,12 +321,22 @@ class view_take : public ranges::view_interface<view_take<urng_t, exactly, or_th
     constexpr view_take(urng_t _urange, size_t const _size)
         : urange{std::move(_urange)}, target_size{_size}
```
rrahn: btw, you can give the arguments the same name as the classes variables. Only make sure to prefix the member in the constructor's body with `this->`

include/seqan3/range/view/take.hpp
```diff
@@ -362,23 +372,32 @@ class view_take : public ranges::view_interface<view_take<urng_t, exactly, or_th
...
      * No-throw guarantee.
      */
-    constexpr iterator begin() noexcept
+    constexpr auto begin() noexcept
     {
-        return {seqan3::begin(urange), 0, target_size, &(*this)};
+        if constexpr (std::ranges::RandomAccessRange<urng_t> && std::ranges::SizedRange<urng_t>)
+            return std::ranges::begin(urange);
+        else
+            return iterator{seqan3::begin(urange), 0, target_size, &(*this)};
```
rrahn: is `&(*this)` not the same as `this`? 🤔

rrahn: don't use seqan3::begin

include/seqan3/range/view/take.hpp
```diff
@@ -362,23 +372,32 @@ class view_take : public ranges::view_interface<view_take<urng_t, exactly, or_th
...
     //!\copydoc begin()
-    constexpr const_iterator begin() const noexcept
+    constexpr auto begin() const noexcept
         requires ConstIterableRange<urng_t>
     {
-        return {seqan3::cbegin(urange), 0, target_size};
+        if constexpr (std::ranges::RandomAccessRange<urng_t> && std::ranges::SizedRange<urng_t>)
+            return std::ranges::cbegin(urange);
+        else
+            return const_iterator{seqan3::cbegin(urange), 0, target_size};
```
rrahn: seqan3::cbegin

include/seqan3/range/view/take.hpp
```diff
@@ -362,23 +372,32 @@ class view_take : public ranges::view_interface<view_take<urng_t, exactly, or_th
...
     //!\copydoc begin()
-    constexpr const_iterator cbegin() const noexcept
+    constexpr auto cbegin() const noexcept
         requires ConstIterableRange<urng_t>
     {
-        return {seqan3::cbegin(urange), 0, target_size};
+        if constexpr (std::ranges::RandomAccessRange<urng_t> && std::ranges::SizedRange<urng_t>)
+            return std::ranges::cbegin(urange);
+        else
+            return const_iterator{seqan3::cbegin(urange), 0, target_size};
```
rrahn: seqan3::cbegin

include/seqan3/range/view/take.hpp
```diff
@@ -396,21 +415,30 @@ class view_take : public ranges::view_interface<view_take<urng_t, exactly, or_th
      */
     constexpr auto end() noexcept
     {
-        return seqan3::end(urange);
+        if constexpr (std::ranges::RandomAccessRange<urng_t> && std::ranges::SizedRange<urng_t>)
+            return std::ranges::begin(urange) + target_size;
+        else
+            return seqan3::end(urange);
```
rrahn: seqan3::end

include/seqan3/range/view/take.hpp
```diff
@@ -396,21 +415,30 @@ class view_take : public ranges::view_interface<view_take<urng_t, exactly, or_th
...
 
     //!\copydoc end()
     constexpr auto end() const noexcept
         requires ConstIterableRange<urng_t>
     {
-        return seqan3::cend(urange);
+        if constexpr (std::ranges::RandomAccessRange<urng_t> && std::ranges::SizedRange<urng_t>)
+            return std::ranges::cbegin(urange) + target_size;
+        else
+            return seqan3::cend(urange);
```
rrahn: seqan3::cend

include/seqan3/range/view/take.hpp
```diff
@@ -396,21 +415,30 @@ class view_take : public ranges::view_interface<view_take<urng_t, exactly, or_th
...
 
     //!\copydoc end()
     constexpr auto cend() const noexcept
         requires ConstIterableRange<urng_t>
     {
-        return seqan3::cend(urange);
+        if constexpr (std::ranges::RandomAccessRange<urng_t> && std::ranges::SizedRange<urng_t>)
+            return std::ranges::cbegin(urange) + target_size;
+        else
+            return seqan3::cend(urange);
```
rrahn: seqan3::cend

include/seqan3/range/view/take.hpp
h-2: I hate `this->`

include/seqan3/range/view/take.hpp
h-2: > is &(*this) not the same as this? thinking

:thinking:  I copy'n'pasted it from before.

include/seqan3/range/view/take.hpp
rrahn: I hate the ugly `_` prefix on the constructor arguments.

include/seqan3/range/view/take.hpp
```diff
@@ -362,23 +372,32 @@ class view_take : public ranges::view_interface<view_take<urng_t, exactly, or_th
...
     //!\copydoc begin()
-    constexpr const_iterator begin() const noexcept
+    constexpr auto begin() const noexcept
         requires ConstIterableRange<urng_t>
     {
-        return {seqan3::cbegin(urange), 0, target_size};
+        if constexpr (std::ranges::RandomAccessRange<urng_t> && std::ranges::SizedRange<urng_t>)
+            return std::ranges::cbegin(urange);
+        else
+            return const_iterator{std::ranges::cbegin(urange), 0, target_size};
```
rrahn: why does this iterator does not need this? this is on purpose?

include/seqan3/range/view/take.hpp
```diff
@@ -362,23 +372,32 @@ class view_take : public ranges::view_interface<view_take<urng_t, exactly, or_th
...
     //!\copydoc begin()
-    constexpr const_iterator cbegin() const noexcept
+    constexpr auto cbegin() const noexcept
         requires ConstIterableRange<urng_t>
     {
-        return {seqan3::cbegin(urange), 0, target_size};
+        if constexpr (std::ranges::RandomAccessRange<urng_t> && std::ranges::SizedRange<urng_t>)
+            return std::ranges::cbegin(urange);
+        else
+            return const_iterator{std::ranges::cbegin(urange), 0, target_size};
```
rrahn: same here

include/seqan3/range/view/take.hpp
h-2: Only single-pass iterators need to pass a pointer/reference to self, because in that case iterating consumes the range and the size of the take becomes smaller with every step.

The const version of the iterator doesn't exist for pure input ranges so it's not neccessary.

(For ForwardRanges the parameter is ignore)

rrahn: Ok

PR #1212, opened 2019-08-14
[FEATURE] lazy evaluation for type traits

h-2:


include/seqan3/core/type_traits/lazy.hpp
```diff
@@ -0,0 +1,122 @@
...
+ * \implements seqan3::TransformationTrait
+ */
+template <template <typename ...> typename template_t, typename ...spec_t>
+struct instantiate<lazy<template_t, spec_t...>>
+{
+    //!\brief Return type of the trait [instantiates the template arguments].
+    using type = template_t<spec_t...>;
+};
+
+/*!\brief A transformation trait that instantiates seqan3::lazy types. Transformation trait shortcut.
```
eseiler:
```suggestion
/*!\brief A transformation trait that instantiates seqan3::lazy types (TransformationTrait shortcut).
```
We do it like this in the other `core/type_traits/*` headers, but I don't care too much.

include/seqan3/core/type_traits/lazy.hpp
```diff
@@ -0,0 +1,122 @@
...
+};
+
+template <typename on_true_t, typename on_false_t>
+struct lazy_conditional<false, on_true_t, on_false_t>
+{
+    using type = instantiate_t<on_false_t>;
+};
+//!\endcond
+
+/*!\brief Behaves like std::conditional_t, but instantiates types wrapped in seqan3::lazy. Transformation trait shortcut.
```
eseiler: Same as above. Also line length

test/snippet/core/type_traits/lazy.cpp
```diff
@@ -0,0 +1,34 @@
+#include <vector>                               // std::vector
+#include <forward_list>                         // std::forward_list
```
eseiler: Order lexicographically and put a new line between stl and seqan3 includes.

test/unit/core/type_traits/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 seqan3_test(basic_test.cpp)
 seqan3_test(deferred_crtp_base_test.cpp)
+seqan3_test(lazy_test.cpp)
```
eseiler: You might as well sort this.

test/unit/core/type_traits/lazy_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+    EXPECT_TRUE((std::is_same_v<lazy_conditional_t<true,
+                                                   std::true_type,
+                                                   std::false_type>,
+                                std::true_type>));
+    EXPECT_TRUE((std::is_same_v<lazy_conditional_t<false,
+                                                   std::true_type,
+                                                   std::false_type>,
+                                std::false_type>));
+
+    // lazy behaviour, safe
```
eseiler: What do safe/important mean in this context?

test/snippet/core/type_traits/lazy.cpp
h-2: We put the space there also in snippets?

eseiler: I skimmed through some snippets and we do if we have stl includes. But I didn't check all.

test/unit/core/type_traits/lazy_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+    EXPECT_TRUE((std::is_same_v<lazy_conditional_t<true,
+                                                   lazy<std::vector, int>,
+                                                   lazy<std::list, int>>,
+                                std::vector<int>>));
+    EXPECT_TRUE((std::is_same_v<lazy_conditional_t<false,
+                                                   lazy<std::vector, int>,
+                                                   lazy<std::list, int>>,
+                                std::list<int>>));
+
+    // lazy behaviour, necessary to compile
```
eseiler: I don't think this test case shows the necessity to use `lazy` and `lazy_conditional_t` instead of `std::conditional_t` in order to compile.
```cpp
    EXPECT_TRUE((std::is_same_v<std::conditional_t<true,
                                                   integral_identity_t<int>,
                                                   void>,
                                int>));
```
and
```cpp
    EXPECT_TRUE((std::is_same_v<lazy_conditional_t<true,
                                                   lazy<integral_identity_t, int>,
                                                   void>,
                                int>));
```
Are equivalent. 
BUT
```cpp
    EXPECT_TRUE((std::is_same_v<std::conditional_t<false,
                                                   integral_identity_t<double>,
                                                   int>,
                                int>));
```
will not compile and
```cpp
    EXPECT_TRUE((std::is_same_v<lazy_conditional_t<false,
                                                   lazy<integral_identity_t, double>,
                                                   int>,
                                int>));
```
will compile.

test/snippet/core/type_traits/lazy.cpp
h-2: (I changed it)

PR #1213, opened 2019-08-14
[DOC] Update changelog

eseiler:


eseiler:
Mhm, I didn't find a way to link inside a codeblock. (I want to link to seqan3::text_layout since a member of the enum can't be automatically linked).

CHANGELOG.md
```diff
@@ -22,15 +22,27 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 ## New features
 
 * Traits for "metaprogramming" with `seqan3::type_list` and type packs have been added.
+* Support for PowerPC has been added.
```
h-2: I think we should make an extra section for `## Platform support`. 

Also please specify which kind of PowerPC. This is little endian ppc64 or did we also test big endian?

I am reluctant about adding offical platform support that we have no CI for, it could break with any commit.

CHANGELOG.md
```diff
@@ -22,15 +22,27 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
...
 ## API changes
 
 ### The `type_list` header has moved
 
-If you included `<seqan3/core/type_list.hpp>` you need to change the path to `<seqan3/core/type_list/type_list.hpp>`.
+If you included `<seqan3/core/type_list.hpp>`, you need to change the path to `<seqan3/core/type_list/type_list.hpp>`.
+
+### The `seqan3::fm_index` and `seqan3::bi_fm_index` now use strong typed template parameters
+
+Instead of `false` and `true`, [seqan3::text_layout::single](@ref seqan3::text_layout) and
```
h-2:
```suggestion
Instead of `false` and `true`, [seqan3::text_layout::single](\ref seqan3::text_layout) and
```

How do these links look when people open the MD file from the directory? Isn't a link for generated automatically if you just write `seqan3::text_layout::single`?

CHANGELOG.md
```diff
@@ -22,15 +22,27 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
...
+[seqan3::text_layout::collection](@ref seqan3::text_layout) are now used to indicate single texts and text
+collections, respectively.<br>
+If you used `fm_index<false>` to create an index over a single text, use `fm_index<text_layout::single>`.<br>
+If you used `fm_index<true>` to create an index over a collection, use `fm_index<text_layout::collection>`.<br>
+The changes in `seqan3::bi_fm_index` are analoguous.
 
 ## Notable Bug-fixes
 
+* Copying and moving the `seqan3::fm_index` and `seqan3::bi_fm_index` now work properly.
```
h-2: We should have some rules for which things are notable :thinking: 

I leaning towards only listing things that were reported by users, but I am not sure about it. Let's discuss at the next strategy meeting!

CHANGELOG.md
eseiler: Agreed. Let's postpone it until we have CI.

CHANGELOG.md
eseiler: Either I do `\ref ...` or `[text](@ref link)` (sic).
It can't link to `::single` or `::collection`; so I have to manually link to the enum instead.

CHANGELOG.md
eseiler: It was reported...so it'd fit. 
We could argue that moving or even copying an index is not the most common use case and the fix is therefore not notable. 
So let's discuss on Monday and postpone this PR until then.

eseiler:
@h-2 If you think the bugfix should be included, you can review again/merge. If you don't want it in the changelog, you can close this PR.

PR #1214, opened 2019-08-15
[FEATURE] Adds all_of traits for type list and parameter packs.

rrahn:
Allows to reduce a type pack or type list by checking a predicate on the type identiry of each type.


eseiler:
Without looking too much into it, it seems similar to [std::conjunction](https://en.cppreference.com/w/cpp/types/conjunction) but more flexible.

include/seqan3/core/type_list/traits.hpp
```diff
@@ -234,6 +235,30 @@ auto split_after(type_list<pack1_t...>, type_list<head_t, pack2_t...>)
         return split_after<idx - 1>(type_list<pack1_t..., head_t>{}, type_list<pack2_t...>{});
 }
 
+//!\cond
+// Fall back that returns false if the unary function declaration is invalid for any given type.
+constexpr auto all_of(...)
+{
+    return false;
+}
+//!\endcond
```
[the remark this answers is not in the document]
rrahn: Not sure, what you mean but if I have something that does not SFINAE on an instantiation failure and is still ill-formed it will fail. It's only that you can use in addition SFINAE and do not provide a default case for every trait to test.

include/seqan3/core/type_list/traits.hpp
```diff
@@ -447,6 +472,32 @@ template <ptrdiff_t n, seqan3::detail::TypeList list_t>
...
+ * ### (Compile-time) Complexity
+ *
+ * Equal to `size<list_t>` (linear).
+ */
+template <seqan3::detail::TypeList list_t, typename unary_function_t>
+constexpr bool all_of(unary_function_t func)
+{
+    return detail::all_of(list_t{}, func);
+}
+
```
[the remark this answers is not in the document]
rrahn: >Currently all our publicly facing list/pack-traits are types or values. But this is a function which complicates the whole design 

Who said, that it only has to be types or values? There is nothing wrong with a function. I also still think this goes into a different header called algorithms. But I really don't care at the moment because these are structural issues not code issues.

>Also in the current design the trait specific arguments come first, and the type_list/pack last. This is necessary for packs and enables us to implement the type_list version in terms of the type pack version for most traits (instead of the other way around which causes more template instantiation).

Well the order is really the smallest concern. I did it like this in the first place because it is done so in the algorithms as well. And it would be even possible to implement `all_of(v1, v2, v3, fn)` using `take` and the fact that `(args, ...)` would return the last argument. Yet, that might be some additional instantiation but it is all not nested not recursive so I think performance will be a small issue on this.

rrahn:
>In general I had planned:
>
>pack_traits::apply (takes a transformation trait, applies to every element) [in #1215 ]
>pack_traits::apply_v (takes a type trait, applies to every element) [naming not clear yet]
>
>Other names for these could be pack_traits::transform. Or one could use that for something else.

Besides that apply means to invoke a function on all arguments and not element wise, we will likely add more algorithms on types because they are useful in many places. At least that is what I had planned.

>What your proposal does is in fact running an arbitrary function on some values, with the exception that those values are previously constructed from types. With the purpose of later passing a lambda that also actually doesn't to anything other than evaluating a type property.
This is switching between type-space and value-space quite often which is not easy to understand and likely also not the best thing for compiler performance (although I am not sure about it).

Well, I can't be sure either but it looks like that the compiler will not not have problems optimizing this. Anyway, if performance is your biggest concern we need to use a third library from people that have spent a good time optimising this. 

In general, this is so much easier than the pure type based approach and closer to modern c++. You don't need to pollute the namespace with helper structs as you can directly call it in the compile time expression with a lambda. Once we have template names it will become even more easier with lambdas because you don't need the decltype anymore. It is a generic solution that can be used in type and value land. 

>I see that we might want the ability to run lambdas on a type list, but I would prefer if we can discuss the general design before blowing it up.

Absolutely. Never meant to be the absolute and final design. It is a first proposal that in my opinion is close to the use of standard algorithms and therefor has a major advantage over designs where I need to provide template member types in structs that evaluate to bool_constant types based on SFINAE or constraints. 

>
>Can you do one of the simpler temporary solutions for #863 like defining a concept and immediately expanding on that?

Sure.

rrahn:
So I moved the functionality to core/detail/pack_algorithm and also moved the already existing for_each functionality to this file.

rrahn:
@smehringer ping

smehringer:
Travis was failing so I did not review this before

rrahn:
you can ignore jenkins for this PR.

include/seqan3/core/type_list/traits.hpp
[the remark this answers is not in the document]
smehringer: is this resolved?

test/unit/core/type_list_test.cpp
```diff
@@ -162,6 +162,39 @@ TEST(pack_traits, split_after)
                                 type_list<>>));
 }
 
+TEST(pack_traits, all_of)
+{
+    constexpr bool res1 = pack_traits::all_of<int8_t, int16_t, uint32_t>(
+                          [](auto type_id)
+                          {
+                              return std::is_integral_v<typename decltype(type_id)::type>;
```
smehringer: Can you also use a concept here?

test/snippet/core/detail/all_of_pack.cpp
```diff
@@ -0,0 +1,36 @@
+
```
smehringer: remove newline :nail_care:

test/snippet/core/detail/all_of_type_list.cpp
```diff
@@ -0,0 +1,46 @@
+
```
smehringer: remove newline :nail_care:

test/snippet/core/detail/all_of_type_list_2.cpp
```diff
@@ -0,0 +1,46 @@
+
```
smehringer: remove newline. (check all your snippets)
what is the difference between this file and all_of_type_list.cpp?

test/snippet/core/detail/for_each_type_list.cpp
```diff
@@ -30,27 +30,27 @@ int main()
         static_assert(std::is_same_v<id_t, std::type_identity<type>>, "id is of type std::type_identity<type>");
 
         if constexpr(std::is_same_v<type, bool>)
-            debug_stream << "bool";
+            seqan3::debug_stream << "bool";
```
smehringer: remove `using namespace seqan3;` in line 8

test/snippet/core/detail/for_each_type_list_2.cpp
```diff
@@ -0,0 +1,54 @@
+#include <iostream>
```
smehringer: you are using debug stream, do you need this include?

rrahn:
> You introduce code in commit 1 & 2 that you delete again in commit 3. I guess you plan to squash this or ? It if confusing to review commit by commit because it is hard to track if there are other changes than those of altering your feature again. It would be nice if you split these changes from the feature.

yes. I'll squash this. 

> I did not review type traits that often so I am a littel confused about the naming.
When do we call something _type? (all_of_type)

Well, we need to somehow distinguish all_of for type lists and parameter packs since we do not use different namespaces as it is done in the type_list traits at the moment. Accordingly, the `_type` version denotes the overload for type_lists.

smehringer:
> Well, we need to somehow distinguish all_of for type lists and parameter packs since we do not use different namespaces as it is done in the type_list traits at the moment. Accordingly, the _type version denotes the overload for type_lists.

Ugh... But we usually put `_type` behind names that denote a type `using iterator_type = ...` . It is confusing that a type trait is now called `_type` :( 
Is this a naming we use throughout the library already or did we recently intrduce/dicuss this and I missed it?

test/snippet/core/detail/all_of_type_list_2.cpp
rrahn: The call to the function. In version one you can pass a type_list instance as argument in 2 you give an explicit template argument

rrahn:
> Ugh... But we usually put _type behind names that denote a type using iterator_type = ... . It is confusing that a type trait is now called _type :(
Is this a naming we use throughout the library already or did we recently intrduce/dicuss this and I missed it?

This was already introduced for `for_each_type`, which was a name used to iterate over types in a type list or template parameter pack. We also had this for parameter packs called for_each_value. I removed the `value` suffix for the parameter packs, but we need a different name for the type versions otherwise there is an ambiguous overload. Note, that these functions are not purely type_traits and as such are not added to type_list/traits, where we explicitly introduced  namespaces. Since this is pure detail and only for developers I guess the naming is not so relevant. In general, I am not a fan of using the suffix type in a class name as well, since all classes are types. We had something like `for_each_in_type_list` but that was to verbose for the use cases we considered. So if you like you can add a discussion item for the next strategy meeting but my feeling is that it is not worth it to have a lengthy discussion about it.

smehringer:
I agree that if it is strictly detail it's not worth a lengthy discussion.
One more suggestion though, what about `type_list_for_each`/`type_list_all_of`? It would mimic the namespace you mentioned for type traits (that I assume would be `type_list::all_of``).
If you disagree, I will just review the PR again.

rrahn:
@smehringer ping

smehringer:
I was waiting for an answer bit I guess this means you disagree

smehringer:
since I am second reviewer now maybe someone else should review this first

rrahn:
> I was waiting for an answer but I guess this means you disagree?

Haven't seen the comment.

rrahn:
> One more suggestion though, what about type_list_for_each/type_list_all_of? It would mimic the namespace you mentioned for type traits (that I assume would be `type_list::all_of``).
If you disagree, I will just review the PR again.

It is actually `list_traits` and `pack_traits`. So another idea would be namespaces like `list_algorithm` and `pack_algorithm`. That I am more in favor of then prefixing or suffixing the function. But then again this is just detail...

test/snippet/core/detail/all_of_type_list.cpp
rrahn: But we always use a new line to separate seqan3 headers from other?

test/snippet/core/detail/all_of_pack.cpp
rrahn: we always use an empty line between seqan3 and other library headers

rrahn:
@marehr ping

marehr:
> @marehr ping

Sorry for the late response, I have the feeling that this PR would be best discussed in person.

rrahn:
> Sorry for the late response, I have the feeling that this PR would be best discussed in person.

Ok, then please put it on the card for next Monday discussion. I do agree that we need to be certain about the design and that here are multiple cornerstones here.

marehr:
Yep. I added a card https://github.com/orgs/seqan/projects/4#card-26492576

joshuak94:
@rrahn What's the status here?

include/seqan3/core/type_list/traits.hpp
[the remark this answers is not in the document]
rrahn: i think after the discussion in the strategy meeting all of it is resolved.

rrahn:
@joshuak94 I am waiting on the review, From POV this is done.

include/seqan3/core/detail/pack_algorithm.hpp
```diff
@@ -0,0 +1,220 @@
...
+ * ### Example
+ *
+ * \include test/snippet/core/detail/for_each_in_pack.cpp
+ *
+ * ### Complexity
+ *
+ * Linear in the number of elements in the pack.
+ *
+ * \attention Opposed to the std::for_each the argument order is changed, such that the first argument is the unary
+ * function to invoke on each argument followed by the arguments.
```
marehr:
```suggestion
 * function to invoke on each argument followed by the arguments. This is due to a constraint in the c++ language regarding parameter packs.
```

test/snippet/core/detail/all_of_in_pack.cpp
```diff
@@ -0,0 +1,37 @@
+#include <type_traits>
+
+#include <seqan3/core/detail/pack_algorithm.hpp>
+#include <seqan3/core/type_list/type_list.hpp>
+#include <seqan3/core/debug_stream.hpp>
+
+int main()
+{
+    // With c++20 you could also write it like this
+    // auto fn = []<typename type>(std::type_identity<type>)
```
marehr:
```suggestion
    // auto fn = []<typename type>(type &&)
```

test/snippet/core/detail/all_of_in_type_list.cpp
```diff
@@ -24,26 +27,21 @@ int main()
...
-            seqan3::debug_stream << "float";
+            return true;
         else if constexpr(std::is_same_v<type, incomplete::type>)
-            seqan3::debug_stream << "incomplete::type";
-
-        seqan3::debug_stream << ", ";
+            return false;
     };
 
     // prints each type name, i.e. "int, float, bool, incomplete::type, \n"
```
marehr:
```suggestion
```

test/snippet/core/detail/all_of_in_type_list.cpp
```diff
@@ -24,26 +27,21 @@ int main()
...
-    seqan3::detail::for_each_type<int, float, bool, incomplete::type>(fn);
+    seqan3::debug_stream << std::boolalpha;
+    seqan3::debug_stream << seqan3::detail::all_of<seqan3::type_list<int, float, bool, incomplete::type>>(fn); // prints `false`.
```
marehr: I think in this case `static_assert` would be much clearer. It made more since printing something for `for_each`

test/snippet/core/detail/all_of_in_pack.cpp
```diff
@@ -0,0 +1,37 @@
...
+            return true;
+        else if constexpr(std::is_same_v<id_t, double>)
+            return true;
+        else
+            return false;
+    };
+
+    // prints each type name, i.e. "int, float, bool, incomplete::type, \n"
+    seqan3::debug_stream << std::boolalpha;
+    seqan3::debug_stream << seqan3::detail::all_of(fn, 3, 1.2, false, "something else"); // prints `false`.
```
marehr: make this static_asserts

test/snippet/core/detail/all_of_in_pack.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+        if constexpr(std::is_same_v<id_t, bool>)
+            return true;
+        else if constexpr(std::is_same_v<id_t, int>)
+            return true;
+        else if constexpr(std::is_same_v<id_t, double>)
+            return true;
+        else
+            return false;
+    };
```
marehr: I think this snippet could also have a snippet where you ask for certain values.

```c++
        if constexpr(std::is_same_v<id_t, bool>)
            return id == false;
        else if constexpr(std::is_same_v<id_t, int>)
            return id == 3;
        else if constexpr(std::is_same_v<id_t, double>)
            return std::abs(id - 1.2) < 0.00001;
        else
            return false;
```

test/unit/core/detail/pack_algorithm_test.cpp
```diff
@@ -7,13 +7,58 @@
...
 #include <seqan3/core/type_list/type_list.hpp>
 
 using namespace seqan3;
 
-TEST(parameter_pack, for_each_value)
+//-----------------------------------------------------------------------------
+// all_of
+//-----------------------------------------------------------------------------
+
+struct test_integral_meta
```
marehr:
```suggestion
struct is_integral_fn
```

I'm not sure with `_fn`, but it is our naming scheme for view functors.

test/unit/core/detail/pack_algorithm_test.cpp
```diff
@@ -7,13 +7,58 @@
...
+{
+    bool operator()(...) { return false; }
+
+    template <typename identity_t>
+        requires std::integral<typename identity_t::type>
+    bool operator()(identity_t) { return true; }
+};
+
+auto test_integral_value = [] (auto value)
```
marehr:
```suggestion
auto is_integral_value = [] (auto value)
```

test/unit/core/detail/pack_algorithm_test.cpp
```diff
@@ -7,13 +7,58 @@
...
+{
+    return std::is_integral_v<decltype(value)>;
+};
+
+TEST(pack_algorithm, all_of_in_type_list)
+{
+    EXPECT_TRUE(detail::all_of<type_list<>>(test_integral_meta{}));
+    EXPECT_TRUE((detail::all_of<type_list<int8_t, int16_t, uint32_t>>(test_integral_meta{})));
```
marehr:
```suggestion
    EXPECT_TRUE((detail::all_of<type_list<int8_t, int16_t, uint32_t>>(is_integral_fn{})));
```

I think it is easier to read if you write test as if you would actually use the API. The test here is not for `test_integral_meta`, but for `all_of`. So the name should not contain test.

include/seqan3/core/detail/pack_algorithm.hpp
```diff
@@ -0,0 +1,220 @@
...
+ *
+ * Linear in the number of elements in the pack.
+ *
+ * \attention Opposed to the std::all_of the argument order is changed, such that the first argument is the unary
+ * predicate to invoke on each argument followed by the arguments.
+ *
+ * \sa https://en.cppreference.com/w/cpp/language/parameter_pack
+ */
+template <typename unary_predicate_t, typename ...pack_t>
+[[nodiscard]] constexpr bool all_of(unary_predicate_t && fn, pack_t && ...args)
```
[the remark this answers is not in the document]
rrahn: Well, there are many opinions out there that actually are in favor of nodiscard being the default. 
Also from the proposal: 

> You almost never want to discard the return value.  (Using the return value is almost always an essential part of the interface of the given function.) 

I think this pretty much brings it to the point. Not using the return value renders the function useless. So why did the programmer use it then? Did he implied to do something different. Did he not understand the interface? Maybe for_each was the thing he was looking then for. So that is why I am in strong favor to always use nodiscard, except in those interfaces where the return value might give you a "debug" information you do not really need in the program logic. As suggested by the proposal as well.

include/seqan3/core/detail/pack_algorithm.hpp
```diff
@@ -0,0 +1,220 @@
...
+     *
+     * \details
+     *
+     * Invokes the wrapper by passing the types wrapped in std::type_identity.
+     */
+    template <typename wrapper_t>
+    //!\cond
+        requires std::invocable<wrapper_t, std::type_identity<args_t>...>
+    //!\endcond
+    static constexpr std::invoke_result_t<wrapper_t, std::type_identity<args_t>...> invoke(wrapper_t && wrapper)
```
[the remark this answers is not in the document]
rrahn: well, I called it wrapper because it wraps the function that should be actually invoked. Since it is a template function I cannot pass the function by reference but had to use a lambda that then calls the actual function. Thus I figured fn_t is a little unspecific.

include/seqan3/core/detail/pack_algorithm.hpp
```diff
@@ -0,0 +1,220 @@
...
+ * \tparam pack_t The template parameter pack to apply the target function on.
+ *
+ * \details
+ *
+ * The meta algorithms provide a pack version and a type based version. In the type based version the algorithm is
+ * called on the types contained in the enclosing list type after they have been wrapped in std::type_identity.
+ * This requires to expand the types of the list type in a helper function since it cannot be done inside of the
+ * function template declaration. To avoid adding an additional impl function for every algorithm, which takes care of
+ * the unpacking of the parameters, this helper struct provides an invocation method to pass the wrapped parameters
+ * to the desired parameter pack version of the algorithm. Since the target function is iteself a function template
```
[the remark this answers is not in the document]
rrahn: I'll revisit it

include/seqan3/core/detail/pack_algorithm.hpp
```diff
@@ -0,0 +1,220 @@
...
+ * \include test/snippet/core/detail/all_of_in_type_list.cpp
+ *
+ * ### Complexity
+ *
+ * Linear in the number of types in the list. Requires linear number of template instantiations.
+ */
+template <typename list_t, typename unary_predicate_t>
+[[nodiscard]] constexpr bool all_of(unary_predicate_t && fn)
+//!\cond
+    requires type_list_specialisation<list_t> || tuple_like<list_t>
```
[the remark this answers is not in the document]
rrahn: I am happy to remove the tuple based version and it is true, that if one wants to pass a tuple one can use the transformation trait to make it a type_list. This makes the interface explicit and also the expectations for this function.

include/seqan3/core/detail/pack_algorithm.hpp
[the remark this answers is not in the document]
rrahn: I did rename it, because I realised that also with the naming of wrapper_t I was actually pursuing the specific case of the functions we had so far, but that does not necessarily need to be. So I adapted the naming and the documentation accordingly.

include/seqan3/core/detail/pack_algorithm.hpp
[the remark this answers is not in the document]
rrahn: > I think type_list_specialisation is a weird / clunky name and we should consider a rename. I think before it was called TypeList that makes this a tough one to rename.

I know. @h-2  and I had a discussion about it. And while we both weren't really satisfied with the name we couldn't find a better solution. So if you have better suggestion that is always welcome. Put it on the discussion board maybe.

> if we also want to allow tuple_like, we should introduce a new concept that extracts the requires clause.

No I don't think we want this. As you said there is already a transformation available for tuple to type_list and that is expressive enough. The tuple should not be used to just store the types. It would be possible to use std::apply on the tuple and call the algorithm on the expanded arguments there.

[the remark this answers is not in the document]
marehr: okay good!

include/seqan3/core/detail/pack_algorithm.hpp
```diff
@@ -0,0 +1,241 @@
...
+ *
+ * This function behaves like std::all_of but on parameter packs. The invocation(s) will be done without any loop.
+ *
+ * ### Example
+ *
+ * \include test/snippet/core/detail/all_of_in_pack.cpp
+ *
+ * ### Complexity
+ *
+ * Linear in the number of elements in the pack.
```
marehr:
```suggestion
 * Linear in the number of types in the pack.
```

include/seqan3/core/detail/pack_algorithm.hpp
```diff
@@ -0,0 +1,241 @@
...
+ *
+ * ### Example
+ *
+ * \include test/snippet/core/detail/all_of_in_type_list.cpp
+ *
+ * ### Complexity
+ *
+ * Linear in the number of types in the seqan3::type_list.
+ *
+ * [Compile-time complexity: Linear number of template instantiations.]
```
marehr: Here it is in square brackets.

include/seqan3/core/detail/pack_algorithm.hpp
```diff
@@ -0,0 +1,241 @@
...
+ * Instantiating a type might not work because they might not be std::default_constructible.
+ * In addition, it is possible to invoke the unary function on incomplete types.
+ *
+ * ### Example
+ *
+ * \include test/snippet/core/detail/for_each_in_type_list.cpp
+ *
+ * ### Complexity
+ *
+ * Linear in the number of types in the type list. Requires linear number of template instantiations.
```
marehr: Here not in square brackets

include/seqan3/core/detail/pack_algorithm.hpp
rrahn: No, this is about expanding the parameter pack, so it is about elements and not types. Or am I missing something here?

marehr: right sorry

include/seqan3/core/detail/pack_algorithm.hpp
```diff
@@ -0,0 +1,244 @@
...
+ *
+ * \attention Opposed to the std::for_each the argument order is changed, such that the first argument is the unary
+ * function to invoke on each argument followed by the arguments.
+ * This is due to a constraint in the c++ language regarding parameter packs.
+ *
+ * \sa https://en.cppreference.com/w/cpp/language/parameter_pack
+ */
+template <typename unary_function_t, typename ...pack_t>
+//!\cond
+    requires (std::invocable<unary_function_t, pack_t> && ... && true)
```
smehringer: isn't this sufficient:
```suggestion
    requires (std::invocable<unary_function_t, pack_t> && ...)
```

include/seqan3/core/detail/pack_algorithm.hpp
```diff
@@ -0,0 +1,244 @@
...
+//!\endcond
+constexpr void for_each(unary_function_t && fn, pack_t && ...args)
+{
+    [[maybe_unused]] int r = (fn(std::forward<pack_t>(args)), ..., 0);
```
smehringer: Why do you need to store this inside an unused variable?
This should work also:
```suggestion
    (fn(std::forward<pack_t>(args)), ...);
```

include/seqan3/core/detail/pack_algorithm.hpp
rrahn: But it results in the exactly same result. But of course this can be expressed so as well.

include/seqan3/core/detail/pack_algorithm.hpp
rrahn: that works equally likely, yes

include/seqan3/core/detail/pack_algorithm.hpp
```diff
@@ -0,0 +1,244 @@
...
+ *
+ * Concrete this helper does the following:
+ *  * expand the types within the type list
+ *  * wrap each type in std::identity and instantiate it (i.e. transform a type into a value)
+ *  * call a target function with the instances of std::identity (as a pack).
+ *
+ * This is a technical trick to make a type representable as a value. Instantiating a type might not always work
+ * because not every type provides a default constructor. In addition it is possible to use incomplete types as well.
+ */
+template <template <typename ...> typename type_list_t, typename ...args_t>
```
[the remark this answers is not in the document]
rrahn: Well, this is an internal data structure that is used by the more or less public interfaces that check the concepts already. So in order to keep compile time to a minimum and since I believe that this structure does not benefit in the regular use case from this concept check, I left it away.

include/seqan3/core/detail/pack_algorithm.hpp
```diff
@@ -0,0 +1,244 @@
...
+    {
+        return fn(std::type_identity<args_t>{}...);
+    }
+};
+
+//-----------------------------------------------------------------------------
+// all_of
+//-----------------------------------------------------------------------------
+
+/*!\brief Tests whether a given predicate evaluates to `true` for each element in a parameter pack.
```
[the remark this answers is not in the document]
rrahn: > A function parameter pack is a function parameter that accepts zero or more function arguments.

From cppreference. I adapted the brief to use exactly this expression `function parameter pack`. Would that be clearer? At least it follows the definition more closely?

rrahn:
Failing build is unrelated.

include/seqan3/core/detail/pack_algorithm.hpp
smehringer: um yes, thats why I was wandering about the unneccessary `&& true`, does it have any purpose then?

PR #1219, opened 2019-08-19
[INFRA] Make steps view visible again in blue ocean.

rrahn:
Enables the step view per node on blue ocean view in jenkins.


rrahn:
> so the main-problem was that jenkins can't handle sub-tasks.

Yes, apparently it worked once but didn't work anymore. There is an open issue that I tracked. I personally find the subtasks quite handy to group by platform or other things.

PR #1221, opened 2019-08-19
Fix issues with missing zlib

### LINKED ISSUE #1106: [BUG] Including file io does not work without zlib
Because in line 105 of `misc_input.hpp`

```cpp
```

the function `_bgzfCheckHeader` is used although declared in the bgzf_stream header which is **not included** if zlib is not available.

rrahn:
Fixes #1106 

Blocked by ~~#1224~~

### DIFF
--- a/include/seqan3/io/detail/magic_header.hpp
+++ b/include/seqan3/io/detail/magic_header.hpp
@@ -13,12 +13,15 @@
 #pragma once
 
 #include <array>
+#include <string>
+#include <vector>
 
+#include <seqan3/core/bit_manipulation.hpp>
 #include <seqan3/core/type_traits/template_inspection.hpp>
-#include <seqan3/core/platform.hpp>
 #include <seqan3/core/type_list/type_list.hpp>
 #include <seqan3/core/type_list/traits.hpp>
 #include <seqan3/std/type_traits>
+#include <seqan3/std/span>
 
 namespace seqan3::detail
 {
@@ -85,6 +88,26 @@ struct bgzf_compression
     //  B       C       [SLEN        ]  [BSIZE       ]
         '\x42', '\x43', '\x02', '\x00', '\x00', '\x00'
     };
+
+    /*!\brief Checks if the given header is a bgzf header.
+     * \param[in] header The header to validate.
+     * \returns `true` if it is a bgzf header, otherwise `false`.
+     */
+    template <typename char_t, ptrdiff_t extend>
+    static bool validate_header(std::span<char_t, extend> header)
+    {
+        static_assert(std::detail::WeaklyEqualityComparableWith<char_t, char>,
+                      "The given char type of the span must be comparable with char.");
+
+        return (header[0] == magic_header[0] &&                                                     // GZ_ID1
+                header[1] == magic_header[1] &&                                                     // GZ_ID2
+                header[2] == magic_header[2] &&                                                     // GZ_CM
+                (header[3] & magic_header[3]) != 0 &&                                                // FLG_FEXTRA
+                to_little_endian(*reinterpret_cast<uint16_t *>(&header[10])) == magic_header[10] &&  // BGZF_ID1
+                header[12] == magic_header[12] &&                                                    // BGZF_ID2
+                header[13] == magic_header[13] &&                                                    // BGZF_SLEN
+                to_little_endian(*reinterpret_cast<uint16_t *>(&header[14])) == magic_header[14]);   // BGZF_XLEN
+    }
 };
 
 /*!\brief A seqan3::type_list containing the available compression formats.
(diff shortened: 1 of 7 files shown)


include/seqan3/io/detail/magic_header.hpp
```diff
@@ -13,9 +13,11 @@
 #pragma once
 
 #include <array>
+#include <string>
+#include <vector>
```
smehringer: in the node you added, you do not use `vector` or `string` ?

include/seqan3/io/detail/misc_input.hpp
```diff
@@ -104,14 +104,14 @@ inline auto make_secondary_istream(std::basic_istream<char_t> & primary_stream,
         extension = filename.extension().string().substr(1);
 
     // tests whether the given extension matches with one of the given compression tags.
-    auto contains_extension = [] (auto compression_tag, auto const & extension)
+    [[maybe_unused]] auto contains_extension = [] (auto compression_tag, auto const & extension)
```
smehringer: why don't you move this inside of the first if scope to avoid the maybe_unused macro?

rrahn: That does not work. Since it might be that zlib is not available but zstd is.

include/seqan3/io/detail/magic_header.hpp
rrahn: Yes, but I cleaned up the includes and string and vector are used inside of the file at other places.

include/seqan3/io/detail/magic_header.hpp
```diff
@@ -85,6 +87,29 @@ struct bgzf_compression
...
+    static bool validate_header(char const * header)
+    {
+        auto unpack = [] (char const * buffer)
+        {
+            uint16_t tmp;
+            std::uninitialized_copy(buffer, buffer + sizeof(uint16_t), reinterpret_cast<char *>(&tmp));
+            return to_little_endian(tmp);
+        };
+
+        return (header[0] == static_cast<char>(magic_header[0]) &&   // GZ_ID1
```
rrahn: remove static_cast

rrahn:
@h-2 ping

include/seqan3/io/detail/magic_header.hpp
```diff
@@ -85,6 +87,29 @@ struct bgzf_compression
     //  B       C       [SLEN        ]  [BSIZE       ]
         '\x42', '\x43', '\x02', '\x00', '\x00', '\x00'
     };
+
+    /*!\brief Checks if the given header is a bgzf header.
+     * \param[in] header The header to validate.
+     * \returns `true` if it is a bgzf header, otherwise `false`.
+     */
+    static bool validate_header(char const * header)
```
h-2: Can we use `std::string_view` here? I think we should adopt that as policy everywhere instead of `char const *`...

include/seqan3/io/detail/misc_input.hpp
h-2:
```suggestion
    [[maybe_unused]] constexpr auto contains_extension = [] (auto compression_tag, auto const & extension)
```

include/seqan3/io/detail/magic_header.hpp
rrahn: Pheww, I am not sure if this is actually what you want in this context. I would rather pass it as the original std::array data.

include/seqan3/io/detail/misc_input.hpp
rrahn: Well, it this is never evaluated at compile time but I can make this change

include/seqan3/std/span
```diff
@@ -384,14 +366,23 @@ public:
...
+constexpr auto begin(span<span_tp, extend> s) noexcept
+{
+    return s.begin();
+}
+
+template <typename span_tp, ptrdiff_t extend>
+constexpr auto end(span<span_tp, extend> s) noexcept
+{
+    return s.end();
+}
```
h-2: these shouldnt be nessary, there are friends.

include/seqan3/std/span
```diff
@@ -17,16 +17,19 @@
 //!\cond
 #pragma once
 
+#if __has_include(<span>)
```
h-2: This will likely not be enough, but we can fix it later.

include/seqan3/std/span
rrahn: They are replacing the friends on purpose. With the friends everything breaks because of infinite recursive instantiation when evaluating std::ranges::begin :unamused:

include/seqan3/std/span
rrahn: agreed

h-2:
Can you rebase now that the span is merged?

rrahn:
@h-2 done

PR #1224, opened 2019-08-22
[MISC] Add range based constructor to span

rrahn:
Adapts the proposal [p1394r3](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1394r3.pdf) to add  range based constructor.

### DIFF
--- a/include/seqan3/std/iterator
+++ b/include/seqan3/std/iterator
@@ -34,6 +34,10 @@ namespace std
 //!\addtogroup iterator
 //!\{
 
+// ------------------------------------------------------------------
+// iterator type traits
+// ------------------------------------------------------------------
+
 /*!\brief A type trait class that provides uniform interface to the properties of types that model the
  *        WeaklyIncrementable concept.
  * \tparam t The type to query.
@@ -104,6 +108,11 @@ struct incrementable_traits_or_iterator_traits<t>
  */
 template <typename t>
 using iter_difference_t = typename detail::incrementable_traits_or_iterator_traits<t>::difference_type;
+
+/*!\typedef std::iter_reference_t
+ * \brief Alias for ranges::iter_reference_t. Returns the reference type of the iterator.
+ */
+using SEQAN3_DOXYGEN_ONLY(iter_reference_t = ) ::ranges::iter_reference_t;
 
 // ------------------------------------------------------------------
 // auxiliary functions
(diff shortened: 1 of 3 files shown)


marehr:
Was this voted in? I thought they were NOT adding range based constructors.

rrahn:
According to @h-2 this will definitely go in.

marehr:
Thank you for clarifying. That makes sense.

h-2:
> Thank you for clarifying. That makes sense.

The reasoning was that for containers in could lead to unexpected copying while for views it doesnt!

include/seqan3/std/span
```diff
@@ -407,8 +398,13 @@ template <class span_tp, ptrdiff_t span_extent>
     constexpr void swap(span<span_tp, span_extent> &lhs, span<span_tp, span_extent> &rhs) noexcept
     { lhs.swap(rhs); }
 
-
 //  Deduction guides
+template <ContiguousIterator iterator_t>
```
eseiler: Is this one needed? It's not (explicitly) in the proposal.

include/seqan3/std/span
```diff
@@ -418,11 +414,8 @@ template<class span_tp, size_t span_sz>
 template<class span_tp, size_t span_sz>
     span(const array<span_tp, span_sz>&) -> span<const span_tp, span_sz>;
 
-template<class container>
-    span(container&) -> span<typename container::value_type>;
-
-template<class container>
-    span(const container&) -> span<const typename container::value_type>;
+template<ranges::ContiguousRange range_t>
```
eseiler:
```suggestion
template <ranges::ContiguousRange range_t>
```

I guess you could do a search/replace for this since it's in a few locations

include/seqan3/std/iterator
```diff
@@ -301,6 +301,11 @@ SEQAN3_CONCEPT ContiguousIterator = RandomAccessIterator<i> &&
                                     ::ranges::ContiguousIterator<i>;
 //!\endcond
 
+/*!\typedef std::iter_reference_t
```
eseiler: It may fit better next to line 106?  (`iter_difference_t`)
But not so important.

include/seqan3/std/span
rrahn: Well, I am wondering why it is missing since it is a template argument and how would he deduce the type from the iterator otherwise? At least without it would not compile

eseiler: It makes sense that it wouldn't compile without it; just wanted to make sure.
I just compared your changes to the changes in the proposal, and this one was the only difference.

PR #1227, opened 2019-08-22
[INFRA] Delete workspace at end of build.

rrahn:
In order to avoid eating up the memory of the docker host.


.jenkinsfile
```diff
@@ -139,6 +137,12 @@ def run_unit_tests(slave_name, compiler, build_type, model, with_cereal)
                                      description: 'Running ctest failed.')
         throw ex
     }
+    finally
```
marehr: ah this means whatever happens, exception or not, clean the workspace?

rrahn: yes

PR #1225, opened 2019-08-22
[MISC] various fixes related to alphabet CPOs

h-2:
Major change:

  * We don't have `namespace seqan3::custom` anymore, because using that only worked if additions to it were included before the definition of the CPO.
  * We have `template <typename t> struct seqan3::custom` instead. It is specialised over the type that you are creating customisations for. Customisations are added as `static` member functions instead of free functions (the struct acts like "namespace template").
  * I have adapted all existing code to follow this and added an extra test. Including alphabet/concept doesn't automatically include the adaptation alphabets anymore, because it's not necessary.

Other changes:
  * Added poison pill overloads to prevent non-ADL forms of unqualified lookup.
  * Changed order of CPO checks. custom is now checked first, because it is the most explicit form of customisation, it should be able to overwrite the other ones.
  * various minor fixes and cleanup

Open design question:
We now have one type that can be specialised for all customisation points. This means if someone wants T to model alphabet, and they also want T to model ArgParserValuable (or however the thing is going to be called) then they need to provide the customisations for both concepts in the same specialisation "instance" of seqan3::custom.

We could fix this by having `seqan3::custom::alphabet<T>` for the alphabet customisation points and then other sub-types in `seqan3::custom::` for other customisation points, but I think this is a little overkill. 90% of our users will never write their own alphabets, 90% of the remaining users will program their types with members or ADL. 1% might want to adapt a third party type, but targetting a third party type that is already overloaded by a fourth-party for a different customisation point seems very unlikely. Even then they could do it with tricks like `requires true`.

BUT I wanted to say it upfront.

PR #1222, opened 2019-08-20
[MISC] Add alphabet type template to fm_index

eseiler:
Partially reverts #1042 

- [x] How to deprecate?
- [x] Add tests for new errors in cerealisation.

PR #1228, opened 2019-08-26
[FIX] visibility errors on gcc

marehr:
This PR fixes a variant of a visibility bug of gcc.

Apparently, if you use the short-notation for concepts in templates, like this,

```c++
template <typename Iterator, Sentinel<Iterator>>
struct foo;
```

all private/protected members of classes are accessible.

~Currently blocked by #1222 and #1225~

### DIFF
--- a/include/seqan3/alphabet/concept.hpp
+++ b/include/seqan3/alphabet/concept.hpp
@@ -78,7 +78,7 @@ void to_rank(args_t ...) = delete;
 //!\brief Functor definition for seqan3::to_rank.
 struct to_rank_fn
 {
-private:
+public:
     SEQAN3_CPO_IMPL(2, seqan3::custom::alphabet<decltype(v)>::to_rank(v))    // explicit customisation
     SEQAN3_CPO_IMPL(1, to_rank(v)                                       )    // ADL
     SEQAN3_CPO_IMPL(0, v.to_rank()                                      )    // member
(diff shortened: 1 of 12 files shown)


marehr:
@smehringer Can you have a look? Thank you <3

include/seqan3/alphabet/concept.hpp
```diff
@@ -78,7 +78,7 @@ void to_rank(args_t ...) = delete;
 //!\brief Functor definition for seqan3::to_rank.
 struct to_rank_fn
 {
-private:
+public:
```
smehringer: why do the impl function need to be public? As I understand it, they are only used by the `operator()`?

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -495,7 +495,7 @@ class argument_parser
      */
     argument_parser_meta_data info;
 
-private:
+protected:
```
smehringer: I thought we want to befriend a test fixture? 
(any way you forgot the `//!\privatesection`)

include/seqan3/range/view/translate_join.hpp
```diff
@@ -49,6 +49,10 @@ class view_translate_join : public ranges::view_base
     small_vector<translation_frames, 6> selected_frames{};
 
 protected:
+    //!\brief Befriend the following class s.t. iterator and const_iterator can be defined for this type.
+    template <typename, typename>
+    friend class detail::random_access_iterator_base;
```
smehringer: I don't understand this fix :/

include/seqan3/alphabet/concept.hpp
marehr: Because the require statement in gcc only allows public statements; clang on the other hand allows this code.

Not sure if this is a bug/workaround, but @h-2 said that I should not bother, because it is `detail` anyway.

PR #1229, opened 2019-08-27
Feature/alignment debug mode

rrahn:
Adds support for getting the alignment matrix and the trace matrix as debug configuration from the alignment result. 
Note, the actual behavior of dumping the matrix in the debug versions will be added by another PR, since I am currently refactoring a lot inside the kernel and need this functionality for debugging purposes.

### DIFF
--- a/include/seqan3/alignment/configuration/align_config_debug.hpp
+++ b/include/seqan3/alignment/configuration/align_config_debug.hpp
@@ -1,0 +1,38 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides seqan3::align_cfg::debug.
+ * \author Rene Rahn <rene.rahn AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <seqan3/alignment/configuration/detail.hpp>
+#include <seqan3/core/algorithm/algorithm_debugging.hpp>
+#include <seqan3/core/algorithm/pipeable_config_element.hpp>
+
+namespace seqan3::align_cfg
+{
+//!\cond DEV
+/*!\brief Configuration element for debugging the alignment algorithm.
+ * \ingroup alignment_configuration
+ *
+ * \details
+ *
+ * Using this configuration allows to output the alignment matrices from the DP algorithm using the
+ * returned seqan3::alignment_result.
+ * The score matrix is always accessible, while the trace matrix can only be computed if an alignment was
+ * requested via the seqan3::align_cfg::result configuration.
+ *
+ * \note This configuration is only useful for debugging purposes as it can have a significant impact on the
+ *       performance.
+ */
+inline constexpr detail::algorithm_debugging<std::integral_constant<detail::align_config_id,
+                                                                    detail::align_config_id::debug>> debug{};
+//!\endcond
+}  // namespace seqan3::align_cfg
--- a/include/seqan3/alignment/pairwise/align_result_selector.hpp
+++ b/include/seqan3/alignment/pairwise/align_result_selector.hpp
@@ -14,13 +14,18 @@
 
 #include <type_traits>
 
+#include <seqan3/alignment/configuration/align_config_debug.hpp>
 #include <seqan3/alignment/configuration/align_config_result.hpp>
 #include <seqan3/alignment/matrix/alignment_coordinate.hpp>
+#include <seqan3/alignment/matrix/detail/two_dimensional_matrix.hpp>
+#include <seqan3/alignment/matrix/trace_directions.hpp>
 #include <seqan3/alphabet/gap/gapped.hpp>
 #include <seqan3/core/algorithm/configuration.hpp>
 #include <seqan3/core/type_traits/basic.hpp>
 #include <seqan3/core/type_traits/range.hpp>
+#include <seqan3/core/type_traits/template_inspection.hpp>
 #include <seqan3/core/type_traits/transformation_trait_or.hpp>
+#include <seqan3/core/type_list/traits.hpp>
 #include <seqan3/core/type_list/type_list.hpp>
 #include <seqan3/range/decorator/gap_decorator.hpp>
 #include <seqan3/range/view/view_all.hpp>
@@ -64,30 +69,29 @@ template <std::ranges::forward_range first_range_t,
 //!\endcond
 struct align_result_selector
 {
+private:
+
     //!\brief Helper function to determine the actual result type.
-    static constexpr auto _determine()
+    static constexpr auto select()
     {
-        using score_type            = int32_t;
+        using score_type = int32_t;
 
         if constexpr (std::remove_reference_t<configuration_t>::template exists<align_cfg::result>())
         {
-            if constexpr (std::same_as<remove_cvref_t<decltype(get<align_cfg::result>(configuration_t{}).value)>,
-                                    with_back_coordinate_type>)
+            if constexpr (configuration_t::template exists<align_cfg::result<with_back_coordinate_type>>())
             {
                 return alignment_result_value_type<uint32_t,
                                                    score_type,
                                                    alignment_coordinate>{};
             }
-            else if constexpr (std::same_as<remove_cvref_t<decltype(get<align_cfg::result>(configuration_t{}).value)>,
-                                         with_front_coordinate_type>)
+            else if constexpr (configuration_t::template exists<align_cfg::result<with_front_coordinate_type>>())
             {
                 return alignment_result_value_type<uint32_t,
                                                    score_type,
                                                    alignment_coordinate,
                                                    alignment_coordinate>{};
             }
-            else if constexpr (std::same_as<remove_cvref_t<decltype(get<align_cfg::result>(configuration_t{}).value)>,
-                                         with_alignment_type>)
+            else if constexpr (configuration_t::template exists<align_cfg::result<with_alignment_type>>())
             {
                 // Due to an error with gcc8 we define these types beforehand.
                 using first_gapped_seq_type = gapped<value_type_t<first_range_t>>;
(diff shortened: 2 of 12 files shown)


include/seqan3/core/type_list/traits.hpp
```diff
@@ -329,6 +329,22 @@ auto split_after(type_list<pack1_t...>, type_list<head_t, pack2_t...>)
 template <template <typename> typename trait_t, typename ...pack_t>
 type_list<trait_t<pack_t>...> transform(type_list<pack_t...>);
 
+/*!\brief Implementation for replace_at.
+ * \tparam replace_t The type replacing the old one.
+ * \tparam pack1_t   Types in the first type list.
+ * \tparam head_t    Head of the second type list.
```
joergi-w: Is this the type to be replaced? Please document so, if that's the case.

include/seqan3/core/type_list/traits.hpp
```diff
@@ -329,6 +329,22 @@ auto split_after(type_list<pack1_t...>, type_list<head_t, pack2_t...>)
...
+ * \tparam pack2_t   Rest of the second type list.
```
joergi-w: pack1 and pack2 are collections of types before and after the element subject to replacement, right? Can you make it clear in the documentation, please?

include/seqan3/alignment/pairwise/align_result_selector.hpp
```diff
@@ -116,8 +119,39 @@ struct align_result_selector
...
+        if constexpr (!configuration_t::template exists<align_cfg::debug>())
+        {
+            return alignment_result_value_t{};
+        }
+        else
+        {
+            using as_type_list = transfer_template_args_onto_t<alignment_result_value_t, type_list>;
+            using score_matrix_t = two_dimensional_matrix<int32_t, std::allocator<int32_t>, matrix_major_order::column>;
+            using trace_matrix_t = two_dimensional_matrix<trace_directions,
+                                                        std::allocator<trace_directions>,
```
joergi-w: missing include: 
```cpp
```

include/seqan3/alignment/pairwise/align_result_selector.hpp
```diff
@@ -116,8 +119,39 @@ struct align_result_selector
...
+
+            if constexpr (configuration_t::template exists<align_cfg::result<with_alignment_type>>())
+            {
+                using with_score_t = list_traits::replace_at<score_matrix_t, 5, as_type_list>;
+                return transfer_template_args_onto_t<list_traits::replace_at<trace_matrix_t, 6, with_score_t>,
+                                                     alignment_result_value_type>{};
+            }
+            else
+            {
+                return transfer_template_args_onto_t<list_traits::replace_at<score_matrix_t, 5, as_type_list>,
```
joergi-w: When I set the debug config, I receive always the score matrix, and only if the alignment was requested I additionally receive the trace matrix. Please describe this behaviour in the documentation.

include/seqan3/alignment/pairwise/alignment_result.hpp
```diff
@@ -200,6 +209,44 @@ class alignment_result
...
+    //!\cond DEV
+    /*!\brief Returns the score matrix used to compute the alignment.
+     * \return The score matrix.
+     *
+     * \details
+     *
+     * This function is only used for debugging such that performance can be affected significantly when enabling
+     * seqan3::align_cfg::debug.
+     *
+     * \attention This function with fail the compilation, if the alignment was not requested in the alignment
```
joergi-w: Inconsistency: In the selector in `augment_if_debug` you make a distinction about whether alignment has been selected only for the trace matrix (and not for the score matrix). And here is stated that alignment is necessary for the score matrix, which is not true for your implementation!

include/seqan3/alignment/pairwise/alignment_result.hpp
```diff
@@ -200,6 +209,44 @@ class alignment_result
...
+
+    /*!\brief Returns the trace matrix used to compute the alignment.
+     * \return The score matrix.
+     *
+     * \details
+     *
+     * This function is only used for debugging such that performance can be affected significantly when enabling
+     * seqan3::align_cfg::debug.
+     *
+     * \attention This function with fail the compilation, if the alignment was not requested in the alignment
```
joergi-w: Shouldn't you add that also the _debug_ config is required?

include/seqan3/alignment/pairwise/align_result_selector.hpp
rrahn: It is?
> Using this configuration allows to output the alignment score matrix and if requested the alignment trace matrix.\

I rephrase it to make it more clear.

include/seqan3/alignment/pairwise/alignment_result.hpp
rrahn: Well, probably yes. It is only visible to developers though.

include/seqan3/alignment/configuration/align_config_debug.hpp
```diff
@@ -0,0 +1,46 @@
...
+ * returned seqan3::alignment_result.
+ * The score matrix is always accessible, while the trace matrix can only be computed if an alignment was
+ * requested via the seqan3::align_cfg::result configuration.
+ *
+ * \note This configuration is only useful for debugging purposes as it can have a significant impact on the
+ *       performance.
+ */
+inline constexpr debug_type debug{};
+//!\endcond
+}  // namespace seqan3::align_cfg
```
h-2: why not ::detail?

include/seqan3/alignment/pairwise/align_result_selector.hpp
```diff
@@ -116,8 +120,39 @@ struct align_result_selector
         }
     }
 
-    //!\brief The determined result type.
-    using type = decltype(_determine());
+    //!\brief Augments the result type with the debug output for score and trace matrix.
+    template <typename alignment_result_value_t>
+    static constexpr auto augment_if_debug(alignment_result_value_t)
+    {
+        if constexpr (!configuration_t::template exists<align_cfg::debug_type>())
```
h-2: In general, it is better style to not negate the condition and instead just switch the code blocks, but that's :nail_care:

include/seqan3/alignment/pairwise/alignment_result.hpp
```diff
@@ -19,17 +19,21 @@ namespace seqan3::detail
...
  */
 template <typename id_t,
           typename score_t,
           typename back_coord_t = std::nullopt_t *,
           typename front_coord_t = std::nullopt_t *,
-          typename alignment_t = std::nullopt_t *>
+          typename alignment_t = std::nullopt_t *,
+          typename score_debug_matrix_t = std::nullopt_t *,
+          typename trace_debug_matrix_t = std::nullopt_t *>
 struct alignment_result_value_type
```
h-2: Why is the alignment_result type in detail? Aren't people to supposed to know how to access it?

include/seqan3/alignment/pairwise/alignment_result.hpp
```diff
@@ -200,6 +209,43 @@ class alignment_result
         return data.alignment;
     }
     //!\}
+
+    //!\cond DEV
```
h-2: Why is this wrapped in DEV when the class is part of detail anyway?

include/seqan3/core/detail/empty_type.hpp
```diff
@@ -0,0 +1,25 @@
...
+#pragma once
+
+#include <seqan3/core/platform.hpp>
+
+namespace seqan3::detail
+{
+
+//!\brief An empty class type used in meta programming.
+//!\ingroup core
+struct empty_type
```
h-2: I think we have this on other places, too, maybe move to core?

h-2: Oh, it is in code already. Ok. One place that you need to change is inherited_iterator_base, because that exactly also defines this type (with the same name).

include/seqan3/alignment/configuration/align_config_debug.hpp
rrahn: Mhmm, yes, I could put the config in detail.

rrahn: For now it is only visible for developer but I found it nicer to have the chain use `align_cfg` as well. But this is not so important and detail would be fine as well

include/seqan3/alignment/pairwise/alignment_result.hpp
rrahn: That is a internal value type used to store all the members and to selectively disable some of the values in the template argument. This way the user does not need to bother with the template arguments and only sees a simple class that has member functions documented properly. Of course the entire type signature is therefor implementation defined and this is not ideal but for the moment the best compromise I could find.

include/seqan3/core/detail/empty_type.hpp
rrahn: It was moved from there

include/seqan3/alignment/pairwise/alignment_result.hpp
rrahn: No the class itself is not in detail. Only the internal value type.

rrahn:
> On a general design level: Do you expect more debug stuff? Would it make sense to have a extra debug member of the return type and put the other stuff inside so all debug things are wrapped together? I am sure the index could use something similar...

I do not expect more debug stuff for now. But something we can keep in mind for later point. The entire debug stuff is not visible to the user anyway

include/seqan3/alignment/configuration/align_config_debug.hpp
rrahn: So I changed it to some global config thing. But note that the actual debug thing need to be in align_cfg to disambiguate it from other definitions in other namespaces. Like for the search you need your own config element since it uses internal enum ids to validate the combinability. Thus leaving the actual variable in align_cfg makes sense. But I put the global type into detail and to some generic place usable for other algorithms

h-2: ok

include/seqan3/alignment/pairwise/alignment_result.hpp
h-2: Ok, I don't fully know how it works, I just wanted to bring your attention to it.

include/seqan3/core/detail/empty_type.hpp
h-2: Hm, I didn't see it being removed there, but maybe it was.

include/seqan3/alignment/pairwise/alignment_result.hpp
```diff
@@ -190,7 +199,7 @@ class alignment_result
 
     /*!\brief Returns the actual alignment, i.e. the base pair matching.
      * \return At least two aligned sequences, which represent the alignment.
-     * \attention This function with fail the compilation, if the alignment was not requested in the alignment
+     * \attention Compiling this function will fail, if the alignment was not requested in the alignment
```
h-2: As mentioned in the other PR, I don't fully get this, but we can discuss it after it has been merged.

PR #1233, opened 2019-09-02
[FIX] Buffer the first record when calling begin instead of at construction.

### LINKED ISSUE #1232: File Input options are set after the first record is read
## Description

Big program in file i/o. If you want to set options on the file, this can only be done AFTER an instance of sequence_file_input is created. This means that **the first record is always read with default options** :(

### How to repeat the problem

```cpp
    std::istringstream istream{std::string
    {
        "> ID1 lala\n"
        "ACGTTTTTTTTTTTTTTT\n"
        "> ID2\n"
        "ACGTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTT\n"
        "> ID3 lala\n"
        "ACGTTTA\n"
    }};

    sequence_file_input fin{istream, format_fasta{}};

    fin.options.truncate_ids = true;

    auto it = fin.begin();
    std::cout << get<field::ID>(*it) << std::endl; // ID1 lala
   ++it;
    std::cout << get<field::ID>(*it) << std::endl; // ID2
    ++it;
    std::cout << get<field::ID>(*it) << std::endl; // ID3
```

:sob: 

This was never caught by our test because we only do detail tests of the formats but never with sequence files. I started to change this because the read function of the format will soon not be accessible anymore.

smehringer:
Resolves #1232

I included a test in sequence file input. 
Alignment file cannot be tested since there are no input options yet.
Structure file already has a test for truncate_ids which would fail if the public API is tested instead of the detail test with format. This will be changed in the next PR so I would refrain from already changing it now.

### DIFF
--- a/include/seqan3/io/alignment_file/input.hpp
+++ b/include/seqan3/io/alignment_file/input.hpp
@@ -732,6 +732,7 @@ public:
      */
     /*!\brief Returns an iterator to current position in the file.
      * \returns An iterator pointing to the current position in the file.
+     * \throws seqan3::format_error
      *
      * Equals end() if the file is at end.
      *
@@ -743,10 +744,17 @@ public:
      *
      * ### Exceptions
      *
-     * No-throw guarantee.
+     * Throws seqan3::format_error if the first record could not be read into the buffer.
      */
-    iterator begin() noexcept
+    iterator begin()
     {
+        // buffer first record
+        if (!first_record_was_read)
+        {
+            read_next_record();
+            first_record_was_read = true;
+        }
+
         return {*this};
     }
 
@@ -793,7 +801,7 @@ public:
      */
     reference front() noexcept
     {
-        return record_buffer;
+        return *begin();
     }
     //!\}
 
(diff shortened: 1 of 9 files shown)


h-2:
still broken

include/seqan3/io/alignment_file/input.hpp
```diff
@@ -732,21 +732,29 @@ class alignment_file_input
...
+     * \throws seqan3::format_error
      *
      * Equals end() if the file is at end.
      *
      * \include test/snippet/io/alignment_file/alignment_file_input_begin_and_front.cpp
      *
      * ### Complexity
      *
-     * Constant.
+     * Linear in the length of the first record in the file.
```
h-2:
```suggestion
     * Constant.
```

include/seqan3/io/alignment_file/input.hpp
```diff
@@ -732,21 +732,29 @@ class alignment_file_input
...
      *
      * ### Exceptions
      *
-     * No-throw guarantee.
+     * Throws a seqan3::format_error if the input is corrupted.
```
h-2:
```suggestion
     * Throws seqan3::format_error if a the first record could not be read into the buffer.
```

PR #1237, opened 2019-09-03
Misc/refactor aligned sequence builder

rrahn:
**NOTE** This PR is for discussing the current situation with the begin_ra/end_ra and the to_random_access view concept.

In General I am not very happy with doing all this additional random access semantics stuff. 
Some of my general concerns:

~1. The view only wraps the original type and forwards its begin_ra/end_ra members as begin/end. But the type is lost and thus it is not directly usable for using it as the original data structure. But often this is what you want. For example, the gap_decorator fulfills the aligned sequence concept and the bidirectional range concept. You can raise it to random access range as proposed originally but then you loose the aligned sequence concept. However, in case of the gap_decorator you want to be able to use the interface also with the random access iterator (which is also fixed in this PR). Then you can use iterator from a different type as regular iterators to a different interface.
2. I don't see many other places except our decorators where you would model random access even if it  might not be constant time access. Thus, we are effectively preventing the efficient use case in the default version (it was fixable in the gap_decorator, but not sure if I can generalize this), which is worse then expecting that access should be constant even if it is not. I think if documented it should be ok to only stick to one begin/end and always give the weak_random_access_iterator in this case. It would be in general less confusing to use (see first point).~

The view random access range is moved out into another PR together with some changes related to the gap_decorator. This PR is now rebased on this.


include/seqan3/range/views/pseudo_random_access.hpp
```diff
@@ -0,0 +1,199 @@
...
+    requires std::random_access_iterator<decltype(rng.begin_ra())>;
+    requires std::sentinel_for<decltype(rng.begin_ra()), decltype(rng.end_ra())>;
+};
+//!\endcond
+
+// ============================================================================
+//  pseudo_random_access_fn (adaptor definition)
+// ============================================================================
+
+//!\brief View adaptor definition for views::random_access.
```
eseiler:
```suggestion
//!\brief View adaptor definition for views::pseudo_random_access.
``

File: include/seqan3/range/views/pseudo_random_access.hpp [hunk 2]
```diff
@@ -0,0 +1,199 @@
...
+
+    /*!\brief Call the view's constructor with the underlying view as argument.
+     * \returns An instance of the adapted range.
+     */
+    template <std::ranges::viewable_range urng_t>
+    static constexpr auto impl(urng_t && urange)
+    {
+        static_assert(std::ranges::random_access_range<urng_t> || pseudo_random_access_range<urng_t>,
+                      "The adapted range must either model std::ranges::random_access_range or must be "
+                      "a specific seqan range type that supports pseudo random access.");
```
eseiler:
```suggestion
                      "a specific SeqAn range type that supports pseudo random access.");
```

include/seqan3/range/views/pseudo_random_access.hpp
```diff
@@ -0,0 +1,199 @@
...
+ * \ingroup views
+ *
+ * \details
+ *
+ * **Header**
+ * ```cpp
+ *      #include <seqan3/range/views/pseudo_random_access.hpp>
+ * ```
+ *
+ * A pseudo random access range is a range, whose iterator typically defines all the interfaces necessary to allow
```
eseiler:
```suggestion
 * A pseudo random access range is a range whose iterator typically defines all the interfaces necessary to allow
```

include/seqan3/range/views/pseudo_random_access.hpp
```diff
@@ -0,0 +1,199 @@
...
+ * iterator tag.
+ *
+ * ### View properties
+ *
+ * | range concepts and reference_t   | `urng_t` (underlying range type)  | `rrng_t` (returned range type)             |
+ * |----------------------------------|:---------------------------------:|:------------------------------------------:|
+ * | std::ranges::input_range         | *required*                        | *preserved*                                |
+ * | std::ranges::forward_range       |                                   | *preserved*                                |
+ * | std::ranges::bidirectional_range |                                   | *preserved*                                |
+ * | std::ranges::random_access_range |                                   | *preserved*                                |
```
eseiler:
```suggestion
 * | std::ranges::random_access_range |                                   | *guaranteed*                              |
```

And also the weaker concepts

test/snippet/range/view/pseudo_random_access.cpp
```diff
@@ -0,0 +1,27 @@
...
+int main()
+{
+    using seqan3::operator""_dna4;
+
+    // A gap decorator is a pseudo random access range using logarithmic time complexity internally.
+    auto seq = "ACGTACGACT"_dna4;
+    seqan3::gap_decorator dec{seq};
+
+    // The default interface models at most std::bidirectional_range.
+    auto it = std::ranges::begin(dec);          // Returned iterator models std::bidirectional_iterator
```
eseiler:
```suggestion
    auto it = std::ranges::begin(dec);          // Returned iterator models std::bidirectional_iterator.
```

test/snippet/range/view/pseudo_random_access.cpp
```diff
@@ -0,0 +1,27 @@
...
+    std::ranges::advance(it, 3);                // Advancing the iterator takes linear time.
+    seqan3::debug_stream << *it << '\n';        // "T"
+
+    // After adapting it with pseudo_random_access the returned range models std::random_access_range.
```
eseiler:
```suggestion
    // After adapting it with pseudo_random_access, the returned range models std::random_access_range.
```

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -576,12 +179,16 @@ ame Arithmetic operators
...
+    iterator_t insert_gap(iterator_t const it, size_type const count = 1)
     {
+        static_assert(std::same_as<iterator_t, const_iterator> || std::same_as<iterator_t, const_iterator_ra>,
+                      "Expects class iterator or iterator_ra type.");
+
         if (!count) // [[unlikely]]
             return it;
 
-        size_type const pos = std::distance(begin(), it);
+        size_type const pos = it - begin();
```
eseiler: Wouldn't this require random access iterators?
Why not just leave the `std::distance`

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -650,10 +260,14 @@ ame Arithmetic operators
...
-    iterator erase_gap(iterator const first, iterator const last)
+    template <typename iterator_t>
+    iterator_t erase_gap(iterator_t const first, iterator_t const last)
     {
-        size_type const pos1 = std::distance(begin(), first);
-        size_type const pos2 = std::distance(begin(), last);
+        static_assert(std::same_as<iterator_t, const_iterator> || std::same_as<iterator_t, const_iterator_ra>,
+                      "Expects class iterator or iterator_ra type.");
+
+        size_type const pos1 = first - begin();
```
eseiler: Same here

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -1038,6 +656,389 @@ template <std::ranges::view urng_t>
 gap_decorator(urng_t range) -> gap_decorator<urng_t>;
 //!\}
 
+/*!\brief A special iterator that enforces std::random_access_iterator albeit it's complexity is logarithmic.
```
eseiler:
```suggestion
/*!\brief A special iterator that enforces std::random_access_iterator albeit its complexity is logarithmic.
```

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -1038,6 +656,389 @@ template <std::ranges::view urng_t>
...
+ *
+ * \attention This iterator models std::random_access_iterator but **does not guarantee constant time access**.
+ *            Random access is provided in \f$O(\log k)\f$ time (where `k` is the number of gaps).
+ *
+ * This iterator returns values when dereferenced, not references, i.e. it does not model the
+ * [Cpp17random_access_iterator](https://en.cppreference.com/w/cpp/named_req/random_access_iterator)
+ * requirements of the STL. It does model the C++20 std::random_access_iterator.
+ *
+ * You can get a view over the gap_decorator with this random access iterator by using
+ * seqan3::views::pseudo_random_access. But note, that after adapting the gap_decorator the non-range interfaces
```
eseiler:
```suggestion
 * seqan3::views::pseudo_random_access. But note that after adapting the gap_decorator the non-range interfaces
```

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -1038,6 +656,389 @@ template <std::ranges::view urng_t>
...
+ * are lost (e.g. insert_gap or erase_gap). This adaptor allows to use the gap_decorator as a range in algorithms
+ * like std::ranges::distance and get the more efficient logarithmic implementation. The regular iterator only
+ * supports std::bidirectional_iterator and thus linear time would be needed to calculated the distance.
```
eseiler:
```suggestion
 * supports std::bidirectional_iterator and thus linear time would be needed to calculate the distance.
```

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -1038,6 +656,389 @@ template <std::ranges::view urng_t>
...
+class gap_decorator<inner_type>::gap_decorator_iterator_ra
+{
+protected:
+    //!\brief Pointer to the underlying container structure.
+    typename std::add_pointer_t<gap_decorator const> host{nullptr};
+    //!\brief Stores the virtual position index for the seqan3::gap_decorator.
+    typename gap_decorator::size_type pos{0u};
+    //!\brief Stores the physical position in the ungapped/underlying view.
+    int64_t ungapped_view_pos{0}; // must be signed because we need this value to be -1 in case of leading gaps.
+    //!\brief Stores the position (inkl. gaps) where the last (consecutive) gap that is still before the current
```
eseiler:
```suggestion
    //!\brief Stores the position (incl. gaps) where the last (consecutive) gap that is still before the current
```

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -1038,6 +656,389 @@ template <std::ranges::view urng_t>
...
+ * supports std::bidirectional_iterator and thus linear time would be needed to calculated the distance.
+ *
+ * \see seqan3::gap_decorator::gap_decorator_iterator
+ */
+template <std::ranges::viewable_range inner_type>
+//!\cond
+    requires std::ranges::random_access_range<inner_type> && std::ranges::sized_range<inner_type> &&
+             (std::is_const_v<std::remove_reference_t<inner_type>> || std::ranges::view<inner_type>)
+//!\endcond
+class gap_decorator<inner_type>::gap_decorator_iterator_ra
```
eseiler: Any reason to move this outside the class or just preference?

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -1038,6 +656,389 @@ template <std::ranges::view urng_t>
...
+    //!\}
+};
+
+/*!\brief The iterator type over a seqan3::gap_decorator.
+ *
+ * \details
+ *
+ * This iterator returns values when dereferenced, not references, i.e. it does not satisfy the semantic
+ * requirements of [Cpp17BidirectionalIterator](https://en.cppreference.com/w/cpp/named_req/bidirectional_iterator).
+ * It does model the C++20 std::bidirectional_iterator. In addition it offers all interfaces of a standard
```
eseiler:
```suggestion
 * It does model the C++20 std::bidirectional_iterator. In addition, it offers all interfaces of a standard
```

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -1038,6 +656,389 @@ template <std::ranges::view urng_t>
...
+ * std::random_access_iterator except the iterator category which is std::bidirectional_iterator_tag, because the
+ * complexity of the iterator is logarithmic and not constant. However, all interfaces inside the seqan3::gap_decorator
+ * make use of the more efficient logarithmic implementation. Only if you want to use the seqan3::gap_decorator in a
+ * generic algorithm, e.g. std::ranges::distance, the slower linear version will be picked due to the constraints of
+ * the iterator category. To achieve optimal performance in a generic context you can use
+ * seqan3::views::pseudo_random_access to get a subrange over seqan3::gap_decorator::gap_decorator_iterator_ra which
+ * models std::random_access_iterator albeit it's non conforming runtime complexity.
```
eseiler:
```suggestion
 * models std::random_access_iterator albeit its non conforming runtime complexity.
```

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -1038,6 +656,389 @@ template <std::ranges::view urng_t>
...
+ *
+ * ### Implementation details
+ *
+ * This iterator inherits from seqan3::detail::inherited_iterator_base all interfaces from its random access iterator
```
eseiler:
```suggestion
 * This iterator inherits through the seqan3::detail::inherited_iterator_base all interfaces from its random access iterator
```

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -1038,6 +656,389 @@ template <std::ranges::view urng_t>
...
+ * counter part seqan3::gap_decorator::gap_decorator_iterator_ra, but overwrites the iterator category to be
```
eseiler:
```suggestion
 * counterpart seqan3::gap_decorator::gap_decorator_iterator_ra, but overwrites the iterator category to be
```

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
```diff
@@ -204,39 +235,28 @@ class aligned_sequence_builder
...
-                    assert(false); // This should never happen.
-                }
+            if (dir == trace_directions::up)
+                fst_it = insert_gap(fst_aligned, fst_it, span);
+
+            if (dir == trace_directions::left)
+                sec_it = insert_gap(sec_aligned, sec_it, span);
+
+            if constexpr (iterator_is_advanceable<decltype(fst_it)> && iterator_is_advanceable<decltype(sec_it)>)
+            { // If iterator supports +=-operator irrespective of it's category use this (e.g. gap_decorator iterator)
```
eseiler:
```suggestion
            { // If iterator supports +=-operator irrespective of its category use this (e.g. gap_decorator iterator).
```

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
```diff
@@ -204,39 +235,28 @@ class aligned_sequence_builder
...
+                fst_it += span;
+                sec_it += span;
+            }
+            else
+            { // ohterwise fall back to whatever the best way is to advance the iterator.
```
eseiler:
```suggestion
            { // Otherwise fall back to whatever the best way is to advance the iterator.
```

include/seqan3/range/decorator/gap_decorator.hpp
rrahn: Begin resolves to the concept and the gap decorator only provides in the default way a "bidirectional iterator" also it supports all interfaces of a random access iterator. So internally I can use the faster implementation because I know the type. Calling distance would otherwise always fallback to linear runtime.

include/seqan3/range/decorator/gap_decorator.hpp
rrahn: as above

include/seqan3/range/decorator/gap_decorator.hpp
rrahn: in general the class itself becomes really hard to read with large nested classes inside of it. So we started to outsource the implementation in other places as well.

include/seqan3/range/views/pseudo_random_access.hpp
```diff
@@ -0,0 +1,199 @@
...
+ *
+ * \attention This is a concept requirement, not an actual function (however types
+ *            modelling this concept will provide an implementation).
+ */
+/*!\fn pseudo_random_access_range::end_ra()
+ * \brief Returns a pseudo random access sentinel.
+ *
+ * \details
+ *
+ * The returned sentinel must model std::sentinel_for with the corresponding iterator obtained by begin_ra().
```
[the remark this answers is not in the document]
rrahn: Yes, shortly a random access range must not be necessarily a sized range and we only care about random access here.

include/seqan3/range/decorator/gap_decorator.hpp
eseiler: Ah ok, makes sense 👍

include/seqan3/range/views/pseudo_random_access.hpp
```diff
@@ -0,0 +1,199 @@
...
+ * iterator tag.
+ *
+ * ### View properties
+ *
+ * | range concepts and reference_t   | `urng_t` (underlying range type)  | `rrng_t` (returned range type)             |
+ * |----------------------------------|:---------------------------------:|:------------------------------------------:|
+ * | std::ranges::input_range         | *required*                        | *preserved*                                |
+ * | std::ranges::forward_range       |                                   | *preserved*                                |
+ * | std::ranges::bidirectional_range |                                   | *preserved*                                |
+ * | std::ranges::random_access_range |                                   | *guaranteed*                               |
```
eseiler: bidirectional_range, forward_range and input_range are also guaranteed

rrahn: true

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -1038,6 +656,389 @@ template <std::ranges::view urng_t>
...
+ * \attention This iterator models std::random_access_iterator but **does not guarantee constant time access**.
+ *            Random access is provided in \f$O(\log k)\f$ time (where `k` is the number of gaps).
+ *
+ * This iterator returns values when dereferenced, not references, i.e. it does not model the
+ * [Cpp17random_access_iterator](https://en.cppreference.com/w/cpp/named_req/random_access_iterator)
+ * requirements of the STL. It does model the C++20 std::random_access_iterator.
+ *
+ * You can get a view over the gap_decorator with this random access iterator by using
+ * seqan3::views::pseudo_random_access. But note that after adapting the gap_decorator the non-range interfaces
+ * are lost (e.g. insert_gap or erase_gap). This adaptor allows to use the gap_decorator as a range in algorithms
```
smehringer:
```suggestion
 * are lost (e.g. insert_gap or erase_gap). This adaptor allows using the gap_decorator as a random access range in algorithms
```

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -1038,6 +656,389 @@ template <std::ranges::view urng_t>
...
+ * like std::ranges::distance and get the more efficient logarithmic implementation. The regular iterator only
+ * supports std::bidirectional_iterator and thus linear time would be needed to calculate the distance.
```
smehringer:
```suggestion
 * is only identified std::bidirectional_iterator and thus linear time would be needed to calculate the distance.
```
since it does support random access but is not identiefied as asuch via the tag..?

include/seqan3/range/decorator/gap_decorator.hpp
smehringer: If we do this now we should make this official since others need to know. 
I added a discussion card. We should be consistent with our code style.

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -1038,6 +656,389 @@ template <std::ranges::view urng_t>
...
+/*!\brief The iterator type over a seqan3::gap_decorator.
+ *
+ * \details
+ *
+ * This iterator returns values when dereferenced, not references, i.e. it does not satisfy the semantic
+ * requirements of [Cpp17BidirectionalIterator](https://en.cppreference.com/w/cpp/named_req/bidirectional_iterator).
+ * It does model the C++20 std::bidirectional_iterator. In addition, it offers all interfaces of a standard
+ * std::random_access_iterator except the iterator category which is std::bidirectional_iterator_tag, because the
+ * complexity of the iterator is logarithmic and not constant. However, all interfaces inside the seqan3::gap_decorator
+ * make use of the more efficient logarithmic implementation. Only if you want to use the seqan3::gap_decorator in a
```
smehringer:
```suggestion
 * make use of the more efficient logarithmic implementation. Be aware, that if you use the seqan3::gap_decorator in a
```

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -1038,6 +656,389 @@ template <std::ranges::view urng_t>
...
+ * generic algorithm, e.g. std::ranges::distance, the slower linear version will be picked due to the constraints of
+ * the iterator category. To achieve optimal performance in a generic context you can use
+ * seqan3::views::pseudo_random_access to get a subrange over seqan3::gap_decorator::gap_decorator_iterator_ra which
```
smehringer:
```suggestion
 * seqan3::views::pseudo_random_access to get a range over seqan3::gap_decorator::gap_decorator_iterator_ra which
```
I know it is of type subrange, but this sounds as if I can additionally crop the range.

include/seqan3/range/views/pseudo_random_access.hpp
```diff
@@ -0,0 +1,199 @@
...
+ */
+//!\cond
+template <typename rng_t>
+SEQAN3_CONCEPT pseudo_random_access_range = std::ranges::range<rng_t> && requires (rng_t rng)
+{
+    rng.begin_ra();
+    rng.end_ra();
+
+    requires std::random_access_iterator<decltype(rng.begin_ra())>;
+    requires std::sentinel_for<decltype(rng.begin_ra()), decltype(rng.end_ra())>;
```
smehringer: shouldn't this be sized_sentinal since I want it to be distractable from the begin iterator?

include/seqan3/range/views/pseudo_random_access.hpp
```diff
@@ -0,0 +1,199 @@
...
+namespace seqan3::views
+{
+
+/*!\name General purpose views
+ * \{
+ */
+
+/*!\brief            A view adaptor that converts a pseudo random access range to a std::random_access_range.
+ * \tparam urng_t    The type of the range being processed. See below for requirements. [template parameter is
+ *                   omitted in pipe notation]
```
smehringer:
```suggestion
 *                   omitted in pipe notation].
```

include/seqan3/range/views/pseudo_random_access.hpp
```diff
@@ -0,0 +1,199 @@
...
+ * \param[in] urange The range being processed. [parameter is omitted in pipe notation]
```
smehringer:
```suggestion
 * \param[in] urange The range being processed. [parameter is omitted in pipe notation].
```

include/seqan3/range/views/pseudo_random_access.hpp
```diff
@@ -0,0 +1,199 @@
...
+ *
+ * ### Return type
+ *
+ * | `urng_t` (underlying range type)       | `rrng_t` (returned range type)                       |
+ * |:--------------------------------------:|:----------------------------------------------------:|
+ * | `std::ranges::random_access_range`     | `std::ranges::ref_view<urng_t>`                      |
+ * | `seqan3::gap_decorator`                | `std::ranges::subrange<ra_iterator_t, ra_iterator_t>`|
+ *
+ * The adaptor returns exactly the type specified above. In the second case `ra_iterator_t` refers to
+ * seqan3::gap_decorator::gap_decorator_iterator_ra.
```
smehringer: should we link this if this is hidden from the documentation?

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
```diff
@@ -204,39 +235,28 @@ class aligned_sequence_builder
...
+                sec_it = insert_gap(sec_aligned, sec_it, span);
+
+            if constexpr (iterator_is_advanceable<decltype(fst_it)> && iterator_is_advanceable<decltype(sec_it)>)
+            { // If iterator supports +=-operator irrespective of its category use this (e.g. gap_decorator iterator)
+                fst_it += span;
+                sec_it += span;
+            }
+            else
+            { // Otherwise fall back to whatever the best way is to advance the iterator.
+                std::ranges::advance(fst_it, span);
```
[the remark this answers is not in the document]
rrahn: Well, yes I was under the wrong impression, that insert gap returns the iterator behind the inserted gap, but It returns an iterator to the beginning of the inserted element which is correct. So in this case I always have to advance both iterators.

include/seqan3/range/views/pseudo_random_access.hpp
rrahn: No, that is a little tricky. We only want to know if the returned iterator is an random_access_iterator (so you can compute the distance between two iterators of the same type) and that the returned sentinel is a sentinel for this iterator. That does not necessarily mean, that `sentinel - iterator` is also valid. Simply put, there could be a random access range that is not sized, e.g. if it is an infinite range like repeat. So size'dness of a range is orthogonal to whether or not it is random access.

include/seqan3/range/views/pseudo_random_access.hpp
rrahn: Sure, that is like copy'n'paste and then needs to be fixed everywhere.

include/seqan3/range/views/pseudo_random_access.hpp
rrahn: like above

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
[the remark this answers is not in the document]
smehringer: ah ok. Did this fix/change some test? It seems a major change in building the sequence?

include/seqan3/range/views/pseudo_random_access.hpp
smehringer: Phew, true :+1:

include/seqan3/range/views/pseudo_random_access.hpp
rrahn: it is not hidden in the documentation. I was not sure whether or not this is a bug of doxygen, but albeit that the nested iterator classes are declared in a private section they are both visible in the user documentation. So in order to be clear about the type I decided to then also link to it.

include/seqan3/range/views/pseudo_random_access.hpp
smehringer: maybe Add this to the documentation to do issue :)

include/seqan3/range/views/pseudo_random_access.hpp
smehringer: Can't you wrap this again into `//!\privatesection`?
Otherwise you could move this documentation into the class and wrap this code into `//!\cond DEV`

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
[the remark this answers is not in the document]
rrahn: Let's say the concrete tests are ingeniously designed to not catch this 😄 
But it is now (later if everything goes through) tested through the alignment tests thouroghly.

include/seqan3/range/views/pseudo_random_access.hpp
rrahn: Well, I guess that was intentional without the dot. See it is everywhere in the view documentation. The dot is before and this is kind of an additional note for the reader.

include/seqan3/range/decorator/gap_decorator.hpp
rrahn: I adapted your version a little bit

include/seqan3/range/views/pseudo_random_access.hpp
smehringer: ok

include/seqan3/range/views/pseudo_random_access.hpp
rrahn: But honestly, I think it is a good thing to have this iterator public documented. Now most of our concerns are gone. We do not expose an additional begin/end interface but the user can see exactly the type he is getting when using it with the view. It is also documented in the iterator itself of how to get it.

smehringer: fine, I don't care much.

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
```diff
@@ -204,39 +235,28 @@ class aligned_sequence_builder
...
-                {
-                    assert(false); // This should never happen.
-                }
+            if (dir == trace_directions::up)
+                fst_it = insert_gap(fst_aligned, fst_it, span);
+
+            if (dir == trace_directions::left)
+                sec_it = insert_gap(sec_aligned, sec_it, span);
+
+            if constexpr (iterator_is_advanceable<decltype(fst_it)> && iterator_is_advanceable<decltype(sec_it)>)
```
smehringer: Why aren't you just querying the `random_access_iterator`? We should not implement an iterator that is advancable but nothing else? Then we avoid an extra concept

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -1038,6 +655,390 @@ template <std::ranges::view urng_t>
 gap_decorator(urng_t range) -> gap_decorator<urng_t>;
 //!\}
 
+/*!\brief A special iterator that enforces std::random_access_iterator albeit its complexity is logarithmic.
```
smehringer:
```suggestion
/*!\brief A special iterator that models std::random_access_iterator albeit its complexity is logarithmic.
```

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -1038,6 +655,390 @@ template <std::ranges::view urng_t>
...
+ * are lost (e.g. seqan3::gap_decorator::insert_gap or seqan3::gap_decorator::erase_gap).
+ * This adaptor allows using the gap_decorator as a random access range in algorithms like std::ranges::distance and get
+ * the more efficient logarithmic implementation. The regular iterator only models std::bidirectional_iterator
+ * and thus linear time would be needed to calculate the distance.
+ *
+ * \see seqan3::gap_decorator::gap_decorator_iterator
+ */
+template <std::ranges::viewable_range inner_type>
+//!\cond
+    requires std::ranges::random_access_range<inner_type> && std::ranges::sized_range<inner_type> &&
```
smehringer: do we really need all the requires here again? They are probably already checked when instantiating the class

test/unit/range/views/view_pseudo_random_access_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include <seqan3/alphabet/nucleotide/dna4.hpp>
+#include <seqan3/range/decorator/gap_decorator.hpp>
+#include <seqan3/range/views/pseudo_random_access.hpp>
+#include <seqan3/range/views/to_char.hpp>
+#include <seqan3/std/ranges>
+
+using namespace seqan3;
+
+template <typename t>
+class pseudo_random_access_test : public ::testing::Test
```
smehringer: why don't you use the iterator test_template :D

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
rrahn: But that’s the whole point of the mess with the gap decorator iterator. It is not a random access but I can use the += on it anyhow. The concept explicitly checks the iterator category.
 so I would not get the desired behavior.

include/seqan3/range/decorator/gap_decorator.hpp
rrahn: Yes it is required so that Compiler knows that it is the same class.

test/unit/range/views/view_pseudo_random_access_test.cpp
rrahn: Since the class is not defining its own iterator but forwarding it, it seemed not so important. But i don’t mind adding it.

include/seqan3/range/decorator/gap_decorator.hpp
rrahn: I think using the word enforce makes it more clear to the user, that something unusual is happening. Concrete that it models only syntactically the RA iterator. At least that was my intention. Do you disagree?

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
smehringer: ah sorry. Can't you check random_access or psuedo_random_access?

include/seqan3/range/decorator/gap_decorator.hpp
smehringer: Maybe then it's better to move these requirements into the class using static asserts. They need to be checked only once and I think we wanted to do this anyway if requirements do not participate in overload resolution.

include/seqan3/range/decorator/gap_decorator.hpp
smehringer: ok fine

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
rrahn: you mean random_access_range? That wouldn't work, since it refines the random_access_iterator concept. The pseudo_random_access would not be helpful either, since I don't use the begin_ra interface but the normal iterator interface and the psudo_random_access_range doesn't guarantee anything about the standard iterator. But in the latest #1267 there were some problems detected that I missed. I will need to refactor the class a little bit to use a policy based design. I think it also to improve it's usage. I will make this before this PR and rebase it then onto this, so this one is blocked for the time being.

smehringer: ok

smehringer:
>  I think it also to improve it's usage. I will make this before this PR and rebase it then onto this, so this one is blocked for the time being.

Is this not blocked anymore?

rrahn:
> Is this not blocked anymore?

No it is not anymore. I applied the changes on top of this PR. some thing I messed up. don't wanna talk about it :rage4:

include/seqan3/range/decorator/gap_decorator.hpp
smehringer: not resolved

test/snippet/range/views/test_back_inserter_zip.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+#include <seqan3/core/debug_stream.hpp>
+#include <seqan3/std/concepts>
+#include <seqan3/std/ranges>
+
+#include <range/v3/view/zip_with.hpp>
+
+
+int main()
+{
```
smehringer: what does this snippet have to do with all of this?

include/seqan3/core/type_traits/lazy.hpp
```diff
@@ -95,13 +95,50 @@ struct lazy_conditional : instantiate<std::conditional_t<decision, on_true_t, on
...
+ *        parameters.
+ * \implements seqan3::unary_type_trait
+ * \tparam query_t The type of the template class to test.
+ * \tparam args_t  The template parameter pack to instantiate the template class with.
+ *
+ * \details
+ *
+ * Note, this unary type trait can be used in a seqan3::detail::lazy_conditional expression to check if instantiating
+ * a template class with specific template arguments would result in a valid template definition. Thus, the template
+ * parameters of the checked class must be constraint accordingly.
```
smehringer:
```suggestion
 * parameters of the checked class must be constrained accordingly.
```

test/snippet/core/type_traits/is_instantiable_with.cpp
```diff
@@ -0,0 +1,28 @@
...
+    using cond_t = seqan3::detail::lazy_conditional_t<seqan3::detail::is_instantiable_with_v<foo, t>,
+                                                      seqan3::detail::lazy<foo, t>,
+                                                      t>;
+    return cond_t{v};
+}
+
+int main()
+{
+    auto a = bar(10);
+    seqan3::debug_stream << "a: " << a.value << "\n";
```
smehringer:
```suggestion
    seqan3::debug_stream << "a: " << a.value << "\n"; // prints ??
```

include/seqan3/range/decorator/gap_decorator.hpp
rrahn: No, we can't do it. See I added the handy unary type trait is_instantiable_with to check if I can instantiate the class template with a  specific type. But for this to work I need to constraint the template parameter. static asserts in the body won't work this way.

test/snippet/range/views/test_back_inserter_zip.cpp
rrahn: boah. Why is this added? :rage3:

include/seqan3/range/decorator/gap_decorator.hpp
smehringer: uff... this will be a problem for every type now then that you'd like to test with `is_instantiable_with` because so far we modeled all the requirements via static asserts :/

test/snippet/core/type_traits/is_instantiable_with.cpp
```diff
@@ -0,0 +1,28 @@
...
+{
+    using cond_t = seqan3::detail::lazy_conditional_t<seqan3::detail::is_instantiable_with_v<foo, t>,
+                                                      seqan3::detail::lazy<foo, t>,
+                                                      t>;
+    return cond_t{v};
+}
+
+int main()
+{
+    auto a = bar(10);
```
smehringer:
```suggestion
    foo<int> a = bar(10);
```
This is more expressive about what happens

test/snippet/core/type_traits/is_instantiable_with.cpp
```diff
@@ -0,0 +1,28 @@
...
+    seqan3::debug_stream << "a: " << a.value << "\n"; // print 10
+    auto b = bar(0.4f);
```
smehringer:
```suggestion
    float b = bar(0.4f);
```

include/seqan3/range/decorator/gap_decorator.hpp
rrahn: yes. Because we had a wrong assumption on what we can do with it. Having the static_assertion in addition shouldn't matter, though. However, this doesn't need to be that strict. This is a detail function used in a few places only. But handy for such cases.

rrahn: Yes. That's true but unfortunately kind of limiting for fun like this.

rrahn:
Ok, I assigned it to @marehr.

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -576,12 +178,16 @@ ame Arithmetic operators
      * Average and worst case (insertion before last gap): \f$O(k)\f$,
      * Best case (back insertion): \f$O(\log k)\f$.
      */
-    iterator insert_gap(iterator const it, size_type const count = 1)
+    template <typename iterator_t>
+    iterator_t insert_gap(iterator_t const it, size_type const count = 1)
```
marehr: I think we should be as explicit in public api as possible. This signature suggest that we accept ANY iterator.

I asked this further down, but why did we inherit the bidrectional from the random_access one again? I forgot which benefit it had.

I personally would prefer a signature like this. Because this will also generate nice documentation.

```c++
template <typename iterator_category>
gap_decorator_iterator<iterator_category> insert_gap(gap_decorator_iterator<iterator_category> const it, size_type const count = 1)
```

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -1038,6 +655,390 @@ template <std::ranges::view urng_t>
...
+ * iterator counterpart seqan3::gap_decorator::gap_decorator_iterator_ra, but overwrites the iterator category to be
+ * std::bidirectional_iterator_tag.
+ */
+template <std::ranges::viewable_range inner_type>
+//!\cond
+    requires std::ranges::random_access_range<inner_type> && std::ranges::sized_range<inner_type> &&
+             (std::is_const_v<std::remove_reference_t<inner_type>> || std::ranges::view<inner_type>)
+//!\endcond
+class gap_decorator<inner_type>::gap_decorator_iterator :
+    public detail::inherited_iterator_base<gap_decorator_iterator, gap_decorator_iterator_ra>
```
marehr: What was the rational again why we inherited this? I forgot.

marehr:
CI is failing; https://travis-ci.org/seqan/seqan3/jobs/586123431#L5487

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
```diff
@@ -27,94 +30,49 @@
...
-                                     decltype(std::declval<sec_rng_t>() | views::slice(0, 1))>,
-                  "The given sequence type cannot be transformed to a aligned_sequence. It needs to be a views::slice "
-                  "over the original type.");
-
-    using fst_aligned_t = typename traits_t::fst_aligned_t; //!< The aligned sequence type for the first range.
-    using sec_aligned_t = typename traits_t::sec_aligned_t; //!< The aligned sequence type for the second range.
+    //!\brief The aligned sequence type for the first range.
+    using fst_aligned_t =
+        lazy_conditional_t<
+            is_instantiable_with_v<gap_decorator, decltype(std::declval<fst_rng_t>() | views::slice(0, 1))>,
```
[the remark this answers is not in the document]
rrahn: Well, at the moment we only support RA ranges. This should work. However, the gap_decorator imposes some requirements on the type like being ra, and sized and const_iteratble etc. So there might be a scenario where your input sequences does not fulfill some of these requirements, e.g. list. This means that the data structure cannot be gap decorated and we need a copy of it instead.

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
```diff
@@ -172,15 +147,21 @@ class aligned_sequence_builder
...
-        matrix_coordinate path_begin = trace_it.coordinate();
+        res.begin = trace_it.coordinate();
 
-        fst_aligned_t fst_aligned{fst_rng | views::slice(path_begin.col, path_end.col)};
-        sec_aligned_t sec_aligned{sec_rng | views::slice(path_begin.row, path_end.row)};
+        if (band_col_offset.has_value())
+        {
+            res.begin.row += res.begin.col - band_col_offset.value();
+            res.end.row += res.end.col - band_col_offset.value();
+        }
```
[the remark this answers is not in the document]
rrahn: yes, I agree.  When adapting it for the banded case I realised that this is needs to be changed. Back then I thought, I do it later. Just get finished with the PR. 
But now I really reconsider it and will change it before. 

Just as a note, this means to augment the two_dimensional matrix for banded cases. Because this is the only place that knows the mapping between real matrix and the underlying buffer.

include/seqan3/range/views/pseudo_random_access.hpp
```diff
@@ -0,0 +1,199 @@
...
+ * | seqan3::const_iterable_range     |                                   | *preserved*                                |
+ * |                                  |                                   |                                            |
+ * | seqan3::reference_t              |                                   | seqan3::reference_t<urng_t>                |
+ *
+ * See the \link views views submodule documentation \endlink for detailed descriptions of the view properties.
+ *
+ * This adaptor requires that the underlying range models either std::ranges::random_access_range or is one of the
+ * following range types:
+ *
+ *  * seqan3::gap_decorator.
```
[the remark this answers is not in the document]
rrahn: Well, that won't be a long list in the first place I guess. 
If I make it as part of the concept, I have to expose the begin_ra/end_ra interface. This whole design looks really awful. 
In my humble opinion what we need to make it actually independent of the begin_ra/end_ra interface is to define a pseudo_random_access_iterator concept. That would require all interfaces of a random access but expects something that is weaker than random_access_iterator_tag. The range then must provide an adaptor iterator that overwrites the iterator category to be random access then. I think this is a more logic design for this interface. This involves writing a view and updating the inherited_iterator_base at some places. This way we get the proper docmentation at least.

include/seqan3/range/views/pseudo_random_access.hpp
```diff
@@ -0,0 +1,199 @@
...
+ * of the adapted range improves to constant time, but merely all syntactical requirements are fulfilled including the
+ * iterator tag.
+ *
+ * ### View properties
+ *
+ * | range concepts and reference_t   | `urng_t` (underlying range type)  | `rrng_t` (returned range type)             |
+ * |----------------------------------|:---------------------------------:|:------------------------------------------:|
+ * | std::ranges::input_range         | *required*                        | *guaranteed*                               |
+ * | std::ranges::forward_range       |                                   | *guaranteed*                               |
+ * | std::ranges::bidirectional_range |                                   | *guaranteed*                               |
```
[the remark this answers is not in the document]
rrahn: No, this is not true. It might be arguable that forward_range is required, because allowing an interface that does something enhanced on an input range seems weird. But for the same reason a bidirectional_range can be pseudo_random_access a forward range can be as well.

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
```diff
@@ -137,6 +95,21 @@ class aligned_sequence_builder
     {}
     //!\}
 
+    /*!\brief Sets the column offset corresponding to the upper diagonal of the band within the matrix.
+     * \param[in] offset The offset to set.
+     *
+     * \details
+     *
+     * The offset corresponds to the cell where the upper diagonal intersects with the first row of the matrix after
+     * trimming the sequences. This option is only necessary if the matrix was computed with a banded alignment.
```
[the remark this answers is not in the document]
rrahn: No, this must be handled here. We don't know the end of the traceback before hand. Unless we want to parse the trace two times. But we don't want to do that. For structural and performance reasons. So since the trace can be only covering a slice of the original sequences the actual alignment is sliced (trimmed might be really the wrong term here).

include/seqan3/range/decorator/gap_decorator.hpp
rrahn: > I asked this further down, but why did we inherit the bidrectional from the random_access one again? I forgot which benefit it had.

Because we want the normal iterator do all the things a random access iterator can without being a random access iterator. So that the default is that we get the slow performance in generic algorithms because people might expect something faster if they require random access but they won't get it with constant time access. This is not my opinion especially since the decorator is trying to emulate a sequence and jumping in it seems just natural. But it was argued that the std::set also does not provide random access with logarithmic performance.

include/seqan3/range/decorator/gap_decorator.hpp
rrahn: see above.

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
```diff
@@ -27,94 +30,49 @@
...
 
 public:
 
     //!\brief The result type when building the aligned sequences.
     struct result_type
     {
-        matrix_coordinate begin; //!< The coordinate where the trimmed first and second sequence begin.
-        matrix_coordinate end; //!< The coordinate where the trimmed first and second sequence end.
-        std::pair<fst_aligned_t, sec_aligned_t> alignment; //!< The alignment over the trimmed sequences.
+        matrix_coordinate begin{}; //!< The coordinate where the trimmed first and second sequence begin.
```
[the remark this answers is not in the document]
rrahn: I did changed the behavior and the naming. Also I wasn't happy with begin and end since it might indicate some itertor or something like that. I hope the renaming reflects the logic of this more.

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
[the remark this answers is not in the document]
rrahn: Ok, I fixed it in another PR. This one will be rebased on that.

test/snippet/range/view/pseudo_random_access.cpp
```diff
@@ -0,0 +1,27 @@
...
+    using seqan3::operator""_dna4;
+
+    // A gap decorator is a pseudo random access range using logarithmic time complexity internally.
+    auto seq = "ACGTACGACT"_dna4;
+    seqan3::gap_decorator dec{seq};
+
+    // The default interface models at most std::bidirectional_range.
+    auto it = std::ranges::begin(dec);          // Returned iterator models std::bidirectional_iterator.
+    std::ranges::advance(it, 3);                // Advancing the iterator takes linear time.
+    seqan3::debug_stream << *it << '\n';        // "T"
```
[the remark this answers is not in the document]
rrahn: I enhanced the documentation

test/unit/alignment/matrix/detail/aligned_sequence_builder_test.cpp
```diff
@@ -247,6 +249,20 @@ TYPED_TEST(aligned_sequence_builder_fixture, build_from_0_0)
...
+    auto p = this->path(matrix_offset{row_index_type{2}, column_index_type{3}});
+    this->builder.set_band_column_offset(0);
+    auto [begin, end, alignment] = this->builder(p);
+
+    EXPECT_EQ(begin.row, 0u);
+    EXPECT_EQ(begin.col, 0u);
+    EXPECT_EQ(end.row, 5u);
+    EXPECT_EQ(end.col, 3u);
+    EXPECT_EQ(alignment.first  | views::to_char | std::ranges::to<std::string>, std::string{"--ACG"});
+    EXPECT_EQ(alignment.second | views::to_char | std::ranges::to<std::string>, std::string{"AG---"});
```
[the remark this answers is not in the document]
rrahn: was removed due to removing the evil interface.

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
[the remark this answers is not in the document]
rrahn: all done from my point now.

include/seqan3/range/decorator/gap_decorator.hpp
rrahn: @smehringer I mark this as resolved. Please unresolve if you think otherwise.

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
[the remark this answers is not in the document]
rrahn: Note, that the random access related stuff is now part of another PR. See the description for more information

rrahn:
@marehr I have moved the random_access things out of this PR into another PR. So this is now about the changes to the aligned sequence builder only. Everything starting at commit 
#569bb10  is relevant for this PR. The other changes are from #1283.

marehr:
Okay, thank you I have a look at it. Just to make sure: 

`[FEATURE] Add is_instantiable_with unary type trait.`

is the first commit that I should look at.

include/seqan3/range/views/pseudo_random_access.hpp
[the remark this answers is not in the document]
marehr: How? A forward range can only go forward. If you don't have any functionality to go backwards then a `pseudo_random_access` can't work.

On some more thoughts. Okay it's doable, because you can always make a linear scan from the beginning to find the predecessor of an iterator position. 

Boah that's nasty, but you can theoretically implement a `pseudo_random_access` range only via `forward_iterators` and `begin` and `end` :boom: . But you need the `forward_range`, because you must be able to reiterate the same content.

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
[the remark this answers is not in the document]
marehr: The trimming part yes, but my initial comment was directed at the offset only used in the banded case. And that - you said - was removed, so that should have made this function obsolete.

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
```diff
@@ -159,10 +127,17 @@ class aligned_sequence_builder
...
+        // helper function to transform the coordinate into a pair
+        auto to_pair = [] (matrix_coordinate const coordinate) constexpr
+        {
+            return std::pair{coordinate.col, coordinate.row};
+        };
+
+        result_type res{};
         auto trace_it = std::ranges::begin(trace_path);
-        matrix_coordinate path_end = trace_it.coordinate();
+        std::tie(res.cut_interval_first.second, res.cut_interval_second.second) = to_pair(trace_it.coordinate());
```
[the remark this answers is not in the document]
rrahn: Instead I would propose offering a tuple_conversion inside of the matrix coordinate.
First, I don't like the tuple interface  because`get<0/1>` does not tell me anything. Second, I don't want to mix things in one class. So either it is a tuple or it is something with it's own functionality.

include/seqan3/core/type_traits/lazy.hpp
```diff
@@ -95,13 +95,50 @@ struct lazy_conditional : instantiate<std::conditional_t<decision, on_true_t, on
...
+template <template <typename ...> typename query_t, typename ...args_t>
+struct is_instantiable_with :
+//!\cond
+    public std::false_type
+//!\endcond
+{};
+
+//!\cond
+template <template <typename ...> typename query_t, typename ...args_t>
+    requires requires { typename std::type_identity<query_t<args_t...>>::type; }
```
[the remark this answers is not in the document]
rrahn: Huh, I am a little confused. Whether a type can be instantiated with the given template arguments or whether the resulting type is incomplete is a different pair of shoes. This unary type trait is not supposed to tell me if I can form an instance of this class template. This becomes far more complex (if possible at all) since I don‘t know how I can form an instance of the class without introspection. I have to check if a SFINAE expression using declval can help. In addition, this is a detail helper function for our code base in which we wouldn‘t allow the baz use case anyhow. There are probably reasons when you want to have the type unconstrained like in some meta code, but I don‘t see that this would be a use case for this unary type trait. Let‘s discuss this in the office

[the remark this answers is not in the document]
rrahn: Ok as discussed I rename it to `is_class_template_declarable_with`

include/seqan3/alignment/matrix/detail/trace_iterator_banded.hpp
```diff
@@ -54,8 +54,14 @@ class trace_iterator_banded : public trace_iterator_base<trace_iterator_banded<m
 
     /*!\brief Constructs from the underlying trace matrix iterator indicating the start of the trace path.
      * \param[in] matrix_iter The underlying matrix iterator.
+     * \param[in] pivot_column The last column index which is still inside of the band in the first row of the
+     *                         banded matrix.
      */
-    constexpr trace_iterator_banded(matrix_iter_t const matrix_iter) noexcept : base_t{matrix_iter}
+    template <typename index_t>
+    constexpr trace_iterator_banded(matrix_iter_t const matrix_iter, column_index_type<index_t> const & pivot_column)
```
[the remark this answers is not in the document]
rrahn: I need to rethink it but mainly because it avoids some additional book keeping.

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
```diff
@@ -214,10 +170,8 @@ class aligned_sequence_builder
      * \param[in,out] fst_aligned The first aligned sequence to insert gaps into.
      * \param[in,out] sec_aligned The second aligned sequence to insert gaps into.
      */
-    template <typename reverse_traces_t>
-    void fill_aligned_sequence(reverse_traces_t && rev_traces,
-                               fst_aligned_t & fst_aligned,
-                               sec_aligned_t & sec_aligned) const
+    template <typename reverse_traces_t, typename fst_t, typename sec_t>
```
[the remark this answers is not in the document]
rrahn: that's true

test/unit/alignment/matrix/detail/aligned_sequence_builder_test.cpp
```diff
@@ -95,155 +97,131 @@ TYPED_TEST(aligned_sequence_builder_fixture, construction)
 TYPED_TEST(aligned_sequence_builder_fixture, build_from_2_3)
 {
     auto p = this->path(matrix_offset{row_index_type{2}, column_index_type{3}});
-    auto [begin, end, alignment] = this->builder(p);
+    auto [cut_interval_first, cut_interval_second, alignment] = this->builder(p);
```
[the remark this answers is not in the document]
rrahn: As above subrange would indicate that it is in fact a subrange. But it is not. It is just the interval of the slice/subrange. I also kept the naming with the first/second. At least this matches the pair member notation in the stl.

include/seqan3/alignment/matrix/detail/aligned_sequence_builder.hpp
```diff
@@ -30,93 +33,80 @@
...
 public:
 
     //!\brief The result type when building the aligned sequences.
-    struct result_type
+    struct [[nodiscard]] result_type
     {
-        matrix_coordinate begin; //!< The coordinate where the trimmed first and second sequence begin.
-        matrix_coordinate end; //!< The coordinate where the trimmed first and second sequence end.
-        std::pair<fst_aligned_t, sec_aligned_t> alignment; //!< The alignment over the trimmed sequences.
+        std::pair<size_t, size_t> slice_interval_first{}; //!< The slice positions of the first sequence.
```
marehr: :)

test/unit/alignment/matrix/detail/aligned_sequence_builder_test.cpp
```diff
@@ -95,155 +97,131 @@ TYPED_TEST(aligned_sequence_builder_fixture, construction)
 TYPED_TEST(aligned_sequence_builder_fixture, build_from_2_3)
 {
     auto p = this->path(matrix_offset{row_index_type{2}, column_index_type{3}});
-    auto [begin, end, alignment] = this->builder(p);
+    auto [slice_interval_first, slice_interval_second, alignment] = this->builder(p);
```
marehr: :)

marehr:
@rrahn You messed your rebases up and introduced new files that should have been deleted: 

test/unit/range/views/view_pseudo_random_access_test.cpp
include/seqan3/range/views/pseudo_random_access.hpp
test/snippet/range/views/pseudo_random_access.cpp

PR #1238, opened 2019-09-03
Feature/add banded trace iterator

rrahn:
Refactors the trace iterator to use crtp base class.
adds implementation of unbanded trace iterator.
adds implementation of banded trace iterator.


eseiler:
@rrahn only `[FEATURE] Add trace iterator for banded trace matrices.` is relevant? 
Also the `[FEATURE] Adds trace iterator` commit only adds the test to the cmake file, just in case you wanted to split the features and not squash it later.

rrahn:
> only [FEATURE] Add trace iterator for banded trace matrices. is relevant?

yes only the last one. The other commit was somehow messed up. I removed it.

include/seqan3/alignment/matrix/detail/trace_iterator_banded.hpp
```diff
@@ -0,0 +1,99 @@
...
+    constexpr trace_iterator_banded(trace_iterator_banded<other_matrix_iter_t> const other) noexcept : base_t{other}
+    {}
+    //!\}
+
+private:
+
+    //!\copydoc seqan3::detail::trace_iterator_base::go_left
+    constexpr void go_left(matrix_iter_t & iter) const noexcept
+    {
+        iter -= matrix_offset{row_index_type{-1}, column_index_type{1}};
```
eseiler: Why always `-=`? 
Should be equivalent to `iter += matrix_offset{row_index_type{1}, column_index_type{-1}};`.

test/unit/alignment/matrix/detail/trace_iterator_banded_test.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+TEST_F(trace_iterator_banded_test, concepts)
+{
+    EXPECT_TRUE(std::ranges::view<path_type>);
+    EXPECT_TRUE(std::ranges::input_range<path_type>);
+    EXPECT_TRUE(std::ranges::forward_range<path_type>);
+    EXPECT_FALSE(std::ranges::bidirectional_range<path_type>);
+}
+
+TEST_F(trace_iterator_banded_test, trace_path_2_6)
```
eseiler:
```suggestion
TEST_F(trace_iterator_banded_test, trace_path_2_5)
```

include/seqan3/alignment/matrix/detail/trace_iterator.hpp
```diff
@@ -12,16 +12,12 @@
...
-#include <seqan3/std/ranges>
+#include <seqan3/alignment/matrix/detail/trace_iterator_base.hpp>
 
 namespace seqan3::detail
 {
 
-/*!\brief A trace iterator for affine gaps.
+/*!\brief A trace iterator an unbanded trace matrix.
  * \ingroup alignment_matrix
  * \extends std::forward_iterator
```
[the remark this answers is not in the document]
rrahn: Ok, I add it to the banded iterator. But note, that this is not the requirement of the iterator but that this iterator extends the forward_iterator. But probably I can leave this away. I have another look into the docu

include/seqan3/alignment/matrix/detail/trace_iterator_banded.hpp
rrahn: It is, but is it easier to read? From the compiler point of view this shouldn't make a difference. But another coder will see that something is decremented so it is going back. That  was the intention to use subtraction.

include/seqan3/alignment/matrix/detail/trace_iterator_base.hpp
```diff
@@ -0,0 +1,284 @@
...
+#include <seqan3/alignment/matrix/detail/two_dimensional_matrix_iterator_concept.hpp>
+#include <seqan3/alignment/matrix/detail/two_dimensional_matrix_iterator_base.hpp>
+#include <seqan3/alignment/matrix/trace_directions.hpp>
+#include <seqan3/std/concepts>
+#include <seqan3/std/ranges>
+
+namespace seqan3::detail
+{
+
+/*!\brief A crtp base class for trace iterator implentations for the alignment algortihms.
```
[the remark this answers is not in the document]
rrahn: Not sure why the editor didn't mark those?

include/seqan3/alignment/matrix/detail/trace_iterator_banded.hpp
eseiler: For me this line was confusing when doing the test by hand.
I first forgot, that those two functions are overloaded;
when I then checked here, it was kinda confusing that there is a `-1` but you actually have to go one down...so `+1`. For me everything was inverted.

I guess if you're used to `-=`, it comes more naturally. 
And I think it's mainly this line because of the double negation; the other functions are more straight forward; so we can leave it as it is,

include/seqan3/alignment/matrix/detail/trace_iterator_base.hpp
```diff
@@ -0,0 +1,284 @@
...
+     * \param[in] matrix_iter The underlying matrix iterator.
+     */
+    constexpr trace_iterator_base(matrix_iter_t const matrix_iter) noexcept : matrix_iter{matrix_iter}
+    {
+        set_trace_direction(*matrix_iter);
+    }
+
+    /*!\brief Constructs from the underlying trace matrix iterator indicating the start of the trace path.
+     * \tparam other_matrix_iter_t The underlying matrix iterator type of `other`; the condition
+     *                             `std::constructible_from<matrix_iter_t, other_matrix_iter_t>` must evaluate to `true`.
```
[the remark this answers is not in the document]
rrahn: Only the `.` ☝️

rrahn:
jenkins test is unrelated spurious fail of concurrent queue

include/seqan3/alignment/matrix/detail/trace_iterator_banded.hpp
rrahn: yes it is only the overload in the banded case that makes this more difficult to read since the actual column in the band is shifted one cell up. Maybe I leave a comment though

include/seqan3/alignment/matrix/detail/trace_iterator_banded.hpp
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * \details
+ *
+ * This iterator follows a given trace in a banded trace matrix.
+ */
+template <two_dimensional_matrix_iterator matrix_iter_t>
+class trace_iterator_banded : public trace_iterator_base<trace_iterator_banded<matrix_iter_t>, matrix_iter_t>
+{
+private:
+
```
h-2: We don't usually have empty lines after private/public, just before, but that's :nail_care:

include/seqan3/alignment/matrix/detail/trace_iterator_base.hpp
```diff
@@ -0,0 +1,295 @@
...
+    }
+    //!\}
+
+    /*!\name Comparison operators
+     * \{
+     */
+    //!\brief Returns `true` if both iterators are equal, `false` otherwise.
+    constexpr friend bool operator==(derived_t const & lhs, derived_t const & rhs) noexcept
+    //!\cond
+        requires std::equality_comparable<matrix_iter_t>
```
h-2: friend requirements are not enforced...

include/seqan3/alignment/matrix/detail/trace_iterator_base.hpp
```diff
@@ -0,0 +1,295 @@
...
+
+public:
+    /*!\name Associated types
+     * \{
+     */
+    using value_type = trace_directions; //!< The value type.
+    using reference = value_type; //!< The reference type.
+    using pointer = value_type *; //!< The pointer type.
+    using difference_type = std::ptrdiff_t; //!< The difference type.
+    using iterator_category = std::forward_iterator_tag; //!< Forward iterator tag.
```
h-2: If the reference == value_type the category can at most be input. But the iterator_concept can be more. (I know this is also not correct in other places).

include/seqan3/alignment/matrix/detail/trace_iterator_base.hpp
```diff
@@ -0,0 +1,295 @@
...
+    void set_trace_direction(trace_directions const dir) noexcept
+    {
+        if (static_cast<bool>(dir & trace_directions::diagonal))
+            current_direction = trace_directions::diagonal;
+        else if (static_cast<bool>(dir & trace_directions::up) ||
+                 static_cast<bool>(dir & trace_directions::up_open))
+            current_direction = trace_directions::up;
+        else if (static_cast<bool>(dir & trace_directions::left) ||
+                 static_cast<bool>(dir & trace_directions::left_open))
+            current_direction = trace_directions::left;
```
h-2: add {} to improve readability :nail_care:

PR #1242, opened 2019-09-03
[FIX] Always use https if available

### LINKED ISSUE #1239: Use HTTPS URL in GitHub description
Please use https://www.seqan.de/.

eseiler:
This PR also fixes dead links.

No HTTPS available:

bzip.org
cmbi.ru.nl
doxygen.nl
ericniebler.com
open-std.org
seqan-update.informatik.uni-tuebingen.de
stroustrup.com
valgrind.org

We also have some XML with the xml-namespace set to the default stuff.
Since it doesn't really matter what is written there, we could just use `https` there (valid URL) or just put whatever namespace there.
But this isn't technically an URL, but an URI (?) and never accessed anyway.

### DIFF
--- a/doc/howto/write_a_view/index.md
+++ b/doc/howto/write_a_view/index.md
@@ -9,7 +9,7 @@ This HowTo documents how to write a view using the standard library and some hel
 \note
 Some of the links from this HowTo only resolve in the developer documentation because
 they refer to entities from the seqan3::detail namespace.
-We recommend you open this tutorial from [the developer documentation](http://docs.seqan.de/seqan/3.0.0-master-dev/).
+We recommend you open this tutorial from [the developer documentation](https://docs.seqan.de/seqan/3-master-dev/).
 
 # Motivation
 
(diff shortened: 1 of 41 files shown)


include/seqan3/io/sequence_file/format_genbank.hpp
```diff
@@ -52,8 +52,8 @@ namespace seqan3
  *
  * ### Introduction
  *
- * genbank is the format used in GenBank sequence database. See
- * (http://quma.cdb.riken.jp/help/gbHelp.html) for a an in-depth description of the format.
+ * genbank is the format used in the GenBank sequence database. See [this example record at NCBI]
+ * (https://www.ncbi.nlm.nih.gov/Sitemap/samplerecord.html) for an in-depth description of the format.
```
smehringer:
```suggestion
 * (https://www.ncbi.nlm.nih.gov/Sitemap/samplerecord.html) for more details about the format.
```
it is not an in-depth description if it is just a sample record?

doc/howto/write_a_view/index.md
```diff
@@ -9,7 +9,7 @@ This HowTo documents how to write a view using the standard library and some hel
 \note
 Some of the links from this HowTo only resolve in the developer documentation because
 they refer to entities from the seqan3::detail namespace.
-We recommend you open this tutorial from [the developer documentation](http://docs.seqan.de/seqan/3.0.0-master-dev/).
+We recommend you open this tutorial from [the developer documentation](https://docs.seqan.de/seqan/3.0.0-master-dev/).
```
smehringer: I think it is https://docs.seqan.de/seqan/3-master-dev/

include/seqan3/io/sequence_file/format_genbank.hpp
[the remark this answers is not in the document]
eseiler: They also explain it.

 But I like your suggestion more

doc/howto/write_a_view/index.md
eseiler: Good catch. I'll check if there are more occurrences

PR #1243, opened 2019-09-03
Feature/add alignment cache

rrahn:
Adds little helper class for the alignment cache instead of using a tuple which is hard to read since we can't use expressive namings for the variables.

~~**NOTE: Only the last commit is relevant.**~~


rrahn:
> It is weird to merge something without a use case but I guess you will refactor this later on.

Well, no one want's to review a >2k LOC change either so at some level I have to make a compromise 🤷‍♂

rrahn:
> The expressive names could be used in the unit test here: https://github.com/seqan/seqan3/blob/5913290cf1b38e9e488a42a571d5a6c48873641b/test/unit/alignment/pairwise/policy/affine_gap_banded_init_policy_test.cpp#L98-99.

A lot of these parts will be removed when I am done with the whole refactoring.

PR #1244, opened 2019-09-06
[FIX, FEATURE] Fixes some formats and restructures test: detail -> public

smehringer:
The restructuring "detail -> public" and a test template **halved the LOC** !

---

* **commit 1: Fix truncate_id option for genbank format.**
The genebank format did not handle the truncate_ids option correctly (It always truncated ids after the first space).

* **commit 2: Fix structured_seq handling to also throw if sequence and structure differ in length..**
The vienna format did only check for equal length in sequence and structure if the seq_type did not decay to std::ignore. It is legal though to read in a structured_seq object (alphabet tuple) in which case seq_type decays to ignore but the same length should still be checked.

* **commit 3: sequence file sam format now truncates ids if specified..**
The sequence file sam format did not handle the truncate_ids option correctly (It never truncated ids after the first space).

* **commit 4: Restructure file IO tests to test public interface.**
The restructuring is necessary before the formats will be changed to be customisable because then, read and write look differently and are protected. If I touch the tests anyway, it is better to test the publich API directly. Also a test template for sequence file formats was introduced.

Some comments:

* In commit 4 there is a name change in format_sam from `format_error` to `parse_error`. Apperrently in sequence file formats we use parse_error everywhere and in alignment files `format_error`. I needed to make this change for the test template but in general we should make the exception names consistent.

* in the sequence file format sam test I deleted a test: wrong_qual3. Wrong quality characters in FASTQ are implicitely casted when reading so this should be the same in both (and it is with the public design)

### DIFF
--- a/include/seqan3/io/sequence_file/format_genbank.hpp
+++ b/include/seqan3/io/sequence_file/format_genbank.hpp
@@ -143,10 +143,19 @@ public:
             else
             {
                 detail::consume(stream_view | views::take_until(!is_blank));
-                // read id
-                std::ranges::copy(stream_view | views::take_until_or_throw(is_cntrl || is_blank)
-                                              | views::char_to<value_type_t<id_type>>,
-                                                std::back_inserter(id));
+
+                auto read_id_until = [&stream_view, &id] (auto predicate)
+                {
+                    std::ranges::copy(stream_view | views::take_until_or_throw(predicate)
+                                                  | views::char_to<value_type_t<id_type>>,
+                                                    std::back_inserter(id));
+                };
+
+                if (options.truncate_ids)
+                    read_id_until(is_space);
+                else
+                    read_id_until(is_cntrl);
+
                 detail::consume(stream_view | views::take_line_or_throw);
             }
         }
--- a/include/seqan3/io/sequence_file/format_sam.hpp
+++ b/include/seqan3/io/sequence_file/format_sam.hpp
@@ -19,6 +19,7 @@
 #include <vector>
 
 #include <seqan3/io/alignment_file/format_sam.hpp>
+#include <seqan3/io/exception.hpp>
 #include <seqan3/io/sequence_file/input_format_concept.hpp>
 #include <seqan3/io/sequence_file/input_options.hpp>
 #include <seqan3/io/sequence_file/output_format_concept.hpp>
@@ -99,10 +100,10 @@ public:
 
         if constexpr (!detail::decays_to_ignore_v<seq_type>)
             if (std::ranges::distance(sequence) == 0)
-                throw format_error{"The sequence information must not be empty."};
+                throw parse_error{"The sequence information must not be empty."};
         if constexpr (!detail::decays_to_ignore_v<id_type>)
             if (std::ranges::distance(id) == 0)
-                throw format_error{"The sequence information must not be empty."};
+                throw parse_error{"The id information must not be empty."};
 
         if (options.truncate_ids)
             id = id | views::take_until_and_consume(is_space) | std::ranges::to<id_type>;
--- a/include/seqan3/io/sequence_file/format_sam.hpp
+++ b/include/seqan3/io/sequence_file/format_sam.hpp
@@ -23,6 +23,8 @@
 #include <seqan3/io/sequence_file/input_options.hpp>
 #include <seqan3/io/sequence_file/output_format_concept.hpp>
 #include <seqan3/io/sequence_file/output_options.hpp>
+#include <seqan3/range/views/take_until.hpp>
+#include <seqan3/std/ranges>
 
 namespace seqan3::detail
 {
@@ -71,7 +73,7 @@ public:
               typename id_type,
               typename qual_type>
     void read(stream_type                                                               & stream,
-              sequence_file_input_options<seq_legal_alph_type, seq_qual_combined> const & SEQAN3_DOXYGEN_ONLY(options),
+              sequence_file_input_options<seq_legal_alph_type, seq_qual_combined> const & options,
               seq_type                                                                  & sequence,
               id_type                                                                   & id,
               qual_type                                                                 & qualities)
@@ -96,11 +98,14 @@ public:
         }
 
         if constexpr (!detail::decays_to_ignore_v<seq_type>)
-            if (std::distance(std::ranges::begin(sequence), std::ranges::end(sequence)) == 0)
+            if (std::ranges::distance(sequence) == 0)
                 throw format_error{"The sequence information must not be empty."};
         if constexpr (!detail::decays_to_ignore_v<id_type>)
-            if (std::distance(std::ranges::begin(id), std::ranges::end(id)) == 0)
+            if (std::ranges::distance(id) == 0)
                 throw format_error{"The sequence information must not be empty."};
+
+        if (options.truncate_ids)
+            id = id | views::take_until_and_consume(is_space) | std::ranges::to<id_type>;
     }
 
 private:
--- a/include/seqan3/io/structure_file/format_vienna.hpp
+++ b/include/seqan3/io/structure_file/format_vienna.hpp
@@ -216,13 +216,17 @@ public:
         }
 
         // READ STRUCTURE (if present)
+        [[maybe_unused]] int64_t structure_length{};
         if constexpr (!detail::decays_to_ignore_v<structure_type>)
         {
             if constexpr (structured_seq_combined)
             {
                 assert(std::addressof(seq) == std::addressof(structure));
                 using alph_type = typename value_type_t<structure_type>::structure_alphabet_type;
-                std::ranges::copy(read_structure<alph_type>(stream_view), begin(structure));
+                // We need the structure_length parameter to count the length of the structure while reading
+                // because we cannot infer it from the (already resized) structure_seq object.
+                auto res = std::ranges::copy(read_structure<alph_type>(stream_view), std::ranges::begin(structure));
+                structure_length = std::ranges::distance(std::ranges::begin(structure), res.out);
 
                 if constexpr (!detail::decays_to_ignore_v<bpp_type>)
                     detail::bpp_from_rna_structure<alph_type>(bpp, structure);
@@ -231,25 +235,27 @@ public:
             {
                 using alph_type = value_type_t<structure_type>;
                 std::ranges::copy(read_structure<alph_type>(stream_view), std::back_inserter(structure));
+                structure_length = std::ranges::distance(structure);
 
                 if constexpr (!detail::decays_to_ignore_v<bpp_type>)
                     detail::bpp_from_rna_structure<alph_type>(bpp, structure);
             }
-            if constexpr (!detail::decays_to_ignore_v<seq_type>)
-                if (size(seq) != size(structure))
-                    throw parse_error{"Found sequence and associated structure of different length."};
         }
         else if constexpr (!detail::decays_to_ignore_v<bpp_type>)
         {
             detail::bpp_from_rna_structure<wuss51>(bpp, read_structure<wuss51>(stream_view));
-
-            if constexpr (!detail::decays_to_ignore_v<seq_type>)
-                if (size(seq) != size(bpp))
-                    throw parse_error{"Found sequence and associated structure of different length."};
+            structure_length = std::ranges::distance(bpp);
         }
         else
         {
             detail::consume(stream_view | views::take_until(is_space)); // until whitespace
+        }
+
+        if constexpr (!detail::decays_to_ignore_v<seq_type> &&
+                      !(detail::decays_to_ignore_v<structure_type> && detail::decays_to_ignore_v<bpp_type>))
+        {
+            if (std::ranges::distance(seq) != structure_length)
+                throw parse_error{"Found sequence and associated structure of different length."};
         }
 
         // READ ENERGY (if present)
--- a/test/unit/io/sequence_file/sequence_file_format_embl_test.cpp
+++ b/test/unit/io/sequence_file/sequence_file_format_embl_test.cpp
@@ -18,38 +18,78 @@
 #include <seqan3/std/algorithm>
 #include <seqan3/test/pretty_printing.hpp>
 
+#include "sequence_file_format_test_template.hpp"
+
 using namespace seqan3;
 
-// ----------------------------------------------------------------------------
-// general
-// ----------------------------------------------------------------------------
-
-TEST(general, concepts)
+template <>
+struct sequence_file_read<format_embl> : public sequence_file_data
 {
-    EXPECT_TRUE((sequence_file_input_format<format_embl>));
-    EXPECT_TRUE((sequence_file_output_format<format_embl>));
-}
+    std::string standard_input
+    {
+R"(ID ID1;	 stuff
+SQ Sequence 1859 BP; 609 A; 314 C; 355 G; 581 T; 0 other;
+  ACGTTTTTTT TTTTTTTT        18
+//
+ID ID2;
+SQ Sequence 1859 BP; 609 A; 314 C; 355 G; 581 T; 0 other;
+  ACGTTTTTTT TTTTTTTTTT TTTTTTTTTT TTTTTTTTTT TTTTTTTTTT TTTTTTTTTT 60
+TTTTTTTTTT TTTTTTTTTT TT        82
+//
+ID ID3 lala;
+SQ Sequence 1859 BP; 609 A; 314 C; 355 G; 581 T; 0 other;
+  ACGTTTA        7
+//)"
+    };
+
+    std::string illegal_alphabet_character_input
+    {
+R"(ID ID1;	 stuff
+SQ Sequence 1859 BP; 609 A; 314 C; 355 G; 581 T; 0 other;
+  AXGTTTTTTT TTTTTTTT        18
+//)"
+    };
+
+    std::string standard_output
+    {
+R"(ID ID1; 18 BP.
+SQ Sequence 18 BP;
+ACGTTTTTTT TTTTTTTT                                               18
+//
+ID ID2; 82 BP.
+SQ Sequence 82 BP;
+ACGTTTTTTT TTTTTTTTTT TTTTTTTTTT TTTTTTTTTT TTTTTTTTTT TTTTTTTTTT 60
+TTTTTTTTTT TTTTTTTTTT TT                                          82
+//
+ID ID3 lala; 7 BP.
+SQ Sequence 7 BP;
+ACGTTTA                                                           7
+//
+)"
+    };
+
+    std::string no_or_ill_formatted_id_input
+    {
+R"(IK ID1;   stuff
+SQ Sequence 1859 BP; 609 A; 314 C; 355 G; 581 T; 0 other;
+  ACGTTTTTTT TTTTTTTT        18
+//)"
+    };
+};
+
+// ---------------------------------------------------------------------------------------------------------------------
+// parametrized tests
+// ---------------------------------------------------------------------------------------------------------------------
+
+INSTANTIATE_TYPED_TEST_CASE_P(embl, sequence_file_read, format_embl);
+INSTANTIATE_TYPED_TEST_CASE_P(embl, sequence_file_write, format_embl);
 
 // ----------------------------------------------------------------------------
 // reading
 // ----------------------------------------------------------------------------
 
-struct read : public ::testing::Test
+struct read : public sequence_file_data
 {
-    std::vector<std::string> expected_ids
-    {
-        { "ID1" },
-        { "ID2" },
-        { "ID3 lala" },
-    };
-
-    std::vector<dna5_vector> expected_seqs
-    {
-        { "ACGTTTTTTTTTTTTTTT"_dna5 },
-        { "ACGTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTT"_dna5 },
-        { "ACGTTTA"_dna5 },
-    };
-
     std::string input
     {
 R"(ID ID1;	stuff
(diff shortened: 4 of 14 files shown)


include/seqan3/io/structure_file/format_vienna.hpp
```diff
@@ -215,42 +215,43 @@ class structure_file_input_format_REMOVEME<format_vienna>
...
         // READ STRUCTURE (if present)
+        [[maybe_unused]] size_t structure_length{};
         if constexpr (!detail::decays_to_ignore_v<structure_type>)
         {
             if constexpr (structured_seq_combined)
             {
                 assert(std::addressof(seq) == std::addressof(structure));
                 using alph_type = typename value_type_t<structure_type>::structure_alphabet_type;
-                std::ranges::copy(read_structure<alph_type>(stream_view), begin(structure));
+                std::ranges::copy(read_structure<alph_type>(stream_view, structure_length), begin(structure));
```
joergi-w: Please write a comment that we need this length counter for the structured sequence field, because we cannot ensure the length equality through `size(seq)` and `size(structure)`.

test/unit/io/sequence_file/sequence_file_format_embl_test.cpp
```diff
@@ -191,83 +174,8 @@ ID ID3 lala;
...
-    {
-        EXPECT_TRUE((std::ranges::equal(seq,  expected_seqs[counter])));
-        EXPECT_TRUE((std::ranges::equal(id,  expected_ids[counter])));
-
-        counter++;
-    }
-
-    EXPECT_EQ(counter, 3u);
+    }};
+    //TODO whrere did this code go?
```
joergi-w: open TODO?

test/unit/io/sequence_file/sequence_file_format_test_template.hpp
```diff
@@ -0,0 +1,245 @@
...
+
+#include <sstream>
+
+#include <gtest/gtest.h>
+
+#include <seqan3/alphabet/quality/all.hpp>
+#include <seqan3/io/sequence_file/input_format_concept.hpp>
+#include <seqan3/io/sequence_file/output_format_concept.hpp>
+#include <seqan3/io/sequence_file/input.hpp>
+#include <seqan3/io/sequence_file/output.hpp>
```
joergi-w: Maybe remove the 4 includes and use
```cpp
#include <seqan3/io/sequence_file/all.hpp>
```

joergi-w: Additionally, you have to add includes for vector, string and std::logic_error.

test/unit/io/structure_file/format_vienna_test.cpp
```diff
@@ -55,16 +58,16 @@ struct read : public ::testing::Test
...
+        { "UUGGAGUACACAACCUGUACACUCUUUC"_rna5 }
     };
 
-    std::vector<std::vector<dot_bracket3>> const expected_structure
+    std::vector<std::vector<wuss<51>>> const expected_structure
     {
-        { "(((((((..((((........)))).((((.........)))).....(((((.......))))))))))))."_db3 },
-        { "..(((((..(((...)))..)))))..."_db3 }
+        { "(((((((..((((........)))).((((.........)))).....(((((.......))))))))))))."_wuss51 },
+        { "..(((((..(((...)))..)))))..."_wuss51 }
```
joergi-w: Why not rna4 and db3?

test/unit/io/structure_file/format_vienna_test.cpp
```diff
@@ -83,25 +86,14 @@ struct read : public ::testing::Test
         }
     };
 
-    detail::structure_file_input_format_REMOVEME<format_vienna> format;
-
-    structure_file_input_options<rna4, false> options;
-
-    static constexpr auto ig = std::ignore; // shortcut
+    structure_file_input_options<rna15, false> options;
```
joergi-w: include rna15

test/unit/io/structure_file/format_vienna_test.cpp
```diff
@@ -12,14 +12,17 @@
...
-#include <seqan3/alphabet/nucleotide/rna4.hpp>
-#include <seqan3/alphabet/structure/dot_bracket3.hpp>
+#include <seqan3/alphabet/nucleotide/rna5.hpp>
+#include <seqan3/alphabet/structure/wuss.hpp>
 #include <seqan3/alphabet/structure/structured_rna.hpp>
 #include <seqan3/io/structure_file/format_vienna.hpp>
 #include <seqan3/io/structure_file/input_format_concept.hpp>
 #include <seqan3/io/structure_file/output_format_concept.hpp>
+#include <seqan3/io/structure_file/input.hpp>
+#include <seqan3/io/structure_file/output.hpp>
```
joergi-w: include structure_file/all instead

test/unit/io/structure_file/format_vienna_test.cpp
```diff
@@ -83,25 +86,14 @@ struct read : public ::testing::Test
...
 
-    rna4_vector seq;
-    std::string id;
-    std::vector<std::set<std::pair<float, uint8_t>>> bpp;
-    std::vector<dot_bracket3> structure;
-    std::vector<structured_rna<rna4, dot_bracket3>> structured_seq;
-    std::optional<double> energy;
-
-    void bpp_test(std::vector<uint8_t> const & bpp_comp)
+    void bpp_test(std::vector<std::set<std::pair<double, size_t>>> & bpp, std::vector<uint8_t> const & bpp_comp)
```
joergi-w: include vector, set, tuple

test/unit/io/structure_file/format_vienna_test.cpp
```diff
@@ -118,22 +110,20 @@ struct read : public ::testing::Test
...
-            EXPECT_NO_THROW(( format.read(istream, options, seq, id, bpp, structure, energy, ig, ig, ig, ig) ));
-
-            EXPECT_EQ(check_energy, energy.has_value());
-            if (check_seq)       { EXPECT_TRUE(( std::ranges::equal(seq, expected_seq[idx]) )); }
-            if (check_id)        { EXPECT_TRUE(( std::ranges::equal(id, expected_id[idx]) )); }
-            if (check_structure) { EXPECT_TRUE(( std::ranges::equal(structure, expected_structure[idx]) )); }
-            if (check_structure) { bpp_test(expected_interactions[idx]); }
-            if (check_energy)    { EXPECT_DOUBLE_EQ(*energy, expected_energy[idx]); }
+            EXPECT_EQ(check_energy, get<field::ENERGY>(*it).has_value());
+            if (check_seq)       { EXPECT_TRUE((std::ranges::equal(get<field::SEQ>(*it), expected_seq[idx]))); }
```
joergi-w: missing include for std::ranges::equal?

test/unit/io/structure_file/format_vienna_test.cpp
```diff
@@ -236,75 +226,68 @@ struct read_fields : public read {};
...
-
-        EXPECT_TRUE(std::ranges::equal(seq, expected_seq[idx]));
-        seq.clear();
+        EXPECT_TRUE(std::ranges::equal(get<field::SEQ>(*it), expected_seq[idx]));
     }
 }
 
 TEST_F(read_fields, only_id)
 {
     std::stringstream istream{input};
```
joergi-w: include sstream

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -96,11 +97,14 @@ class sequence_file_input_format_REMOVEME<format_sam>
         }
 
         if constexpr (!detail::decays_to_ignore_v<seq_type>)
-            if (std::distance(std::ranges::begin(sequence), std::ranges::end(sequence)) == 0)
+            if (std::ranges::distance(sequence) == 0)
```
joergi-w: include for distance?

include/seqan3/io/sequence_file/format_genbank.hpp
```diff
@@ -144,9 +144,18 @@ class sequence_file_input_format_REMOVEME<format_genbank>
...
+                    std::ranges::copy(stream_view | view::take_until_or_throw(is_cntrl || is_blank)
+                                                  | view::char_to<value_type_t<id_type>>,
+                                                    std::back_inserter(id));
+                }
+                else
+                {
+                    std::ranges::copy(stream_view | view::take_until_or_throw(is_cntrl)
+                                                  | view::char_to<value_type_t<id_type>>,
+                                                    std::back_inserter(id));
+                }
```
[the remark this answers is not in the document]
joergi-w: or the same as in sam format:
```cpp
if (options.truncate_ids)
    id = id | view::take_until_and_consume(is_space) | std::ranges::to<id_type>;
```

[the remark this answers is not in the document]
smehringer: I like the first suggestion. The second would require to copy the id again which I would try to avoid

test/unit/io/structure_file/format_vienna_test.cpp
smehringer: Simply because the others are the default and I would need to add an extra traits object to the test and then always specify all template arguments of the sequence_file_input object. I think the test is just as fine now?
We should think about adding tests for different alphabets though.

test/unit/io/structure_file/format_vienna_test.cpp
smehringer: it is defined in `seqan3/std/algorithm` which is included :)

test/unit/io/structure_file/format_vienna_test.cpp
smehringer: it is included :)

test/unit/io/structure_file/format_vienna_test.cpp
joergi-w: Sure, it is fine like this! I was just curious why it was changed, as the structure uses only a tiny subset of the wuss alphabet now. But I see the benefit of omitting the template arguments :)

test/unit/io/structure_file/format_vienna_test.cpp
joergi-w: Oh true, in this file the includes were visually hidden in Github's folding mechanism, so I didn't spot them...

include/seqan3/io/structure_file/format_vienna.hpp
```diff
@@ -216,42 +216,46 @@ class structure_file_input_format_REMOVEME<format_vienna>
...
         if constexpr (!detail::decays_to_ignore_v<structure_type>)
         {
             if constexpr (structured_seq_combined)
             {
                 assert(std::addressof(seq) == std::addressof(structure));
                 using alph_type = typename value_type_t<structure_type>::structure_alphabet_type;
-                std::ranges::copy(read_structure<alph_type>(stream_view), begin(structure));
+                // We need the structure_length parameter to count the length of the structure while reading
+                // because we cannot infer it from the (already resized) structure_seq object.
+                std::ranges::copy(read_structure<alph_type>(stream_view, structure_length), begin(structure));
```
rrahn: std::ranges::begin

include/seqan3/io/structure_file/format_vienna.hpp
```diff
@@ -216,42 +216,46 @@ class structure_file_input_format_REMOVEME<format_vienna>
...
         }
         else
         {
             detail::consume(stream_view | views::take_until(is_space)); // until whitespace
         }
 
+        if constexpr (!detail::decays_to_ignore_v<seq_type> &&
+                      !(detail::decays_to_ignore_v<structure_type> && detail::decays_to_ignore_v<bpp_type>))
+        {
+            if (size(seq) != structure_length)
```
rrahn: std::ranges::size? And is sized range required for seq?

include/seqan3/io/structure_file/format_vienna.hpp
```diff
@@ -216,42 +216,46 @@ class structure_file_input_format_REMOVEME<format_vienna>
...
+                std::ranges::copy(read_structure<alph_type>(stream_view, structure_length), begin(structure));
 
                 if constexpr (!detail::decays_to_ignore_v<bpp_type>)
                     detail::bpp_from_rna_structure<alph_type>(bpp, structure);
             }
             else
             {
                 using alph_type = value_type_t<structure_type>;
-                std::ranges::copy(read_structure<alph_type>(stream_view), std::back_inserter(structure));
+                std::ranges::copy(read_structure<alph_type>(stream_view, structure_length),
```
rrahn: Just wondering. What do we know about the type of `structure`?  I see size was used on it, so we assume sized_range? It is also likely that we parse the input always in a vector or another object with probably having a sized_sentinel? 
If all of this is true then we can get the length of the structure quite easily. 
Either it is std::ranges::distance/size whatever on the structure that is stand alone or 
`std::ranges::distance(std::ranges::begin(structure), copy_result.second)` if it is part of the sequence.
Not exactly sure what the second member of copy_result is called.

include/seqan3/io/sequence_file/format_sam.hpp
```diff
@@ -96,11 +97,14 @@ class sequence_file_input_format_REMOVEME<format_sam>
...
                 throw format_error{"The sequence information must not be empty."};
         if constexpr (!detail::decays_to_ignore_v<id_type>)
-            if (std::distance(std::ranges::begin(id), std::ranges::end(id)) == 0)
+            if (std::ranges::distance(id) == 0)
```
rrahn: you night want to include `seqan3/std/ranges` then as well

test/unit/io/sequence_file/sequence_file_format_genbank_test.cpp
```diff
@@ -86,40 +58,81 @@ ORIGIN
...
+INSTANTIATE_TYPED_TEST_CASE_P(genbank, sequence_file_read, format_genbank);
+INSTANTIATE_TYPED_TEST_CASE_P(genbank, sequence_file_write, format_genbank);
+
+// ----------------------------------------------------------------------------
+// reading
+// ----------------------------------------------------------------------------
+
+struct read : public sequence_file_read<format_genbank>
+{
+    sequence_file_input_options<dna15, false> options;
```
rrahn: options are not initialised

test/unit/io/sequence_file/sequence_file_format_genbank_test.cpp
```diff
@@ -86,40 +58,81 @@ ORIGIN
...
 
-            EXPECT_NO_THROW(( format.read(istream, options, seq, id, std::ignore) ));
-            EXPECT_EQ(id, expected_ids[i]);
-            EXPECT_EQ(seq, expected_seqs[i]);
-            EXPECT_TRUE((ranges::equal(seq, expected_seqs[i])));
-            EXPECT_TRUE((ranges::equal(id, expected_ids[i])));
+        auto it = fin.begin();
+        for (unsigned i = 0; i < 3; ++i, ++it)
+        {
+            EXPECT_EQ(get<field::ID>(*it), ids[i]);
```
rrahn: ok, but what happens with no_throw? This is not relevant anymore?

test/unit/io/sequence_file/sequence_file_format_genbank_test.cpp
```diff
@@ -134,239 +147,42 @@ REFERENCE   1  (bases 1 to 5028)
...
-TEST_F(read, only_id)
-{
-    std::stringstream istream{input};
-
-    for (unsigned i = 0; i < 3; ++i)
-    {
-        id.clear();
-        seq.clear();
-
-        EXPECT_NO_THROW( (format.read(istream, options, std::ignore, id, std::ignore) ));
```
rrahn: so all this has moved into the test template. But are the semantics of the test identical. I noticed that some things are not checking throw or no_throw expressions?

test/unit/io/sequence_file/sequence_file_format_genbank_test.cpp
```diff
@@ -134,239 +147,42 @@ REFERENCE   1  (bases 1 to 5028)
...
-    detail::sequence_file_output_format_REMOVEME<format_genbank> format;
-
     sequence_file_output_options options;
 
     std::ostringstream ostream;
 
     void do_write_test()
     {
+        sequence_file_output fout{ostream, format_genbank{}, fields<field::SEQ, field::ID>{}};
+        fout.options = options;
```
rrahn: also options are not initialized yet. Is that a problem?

test/unit/io/sequence_file/sequence_file_format_sam_test.cpp
```diff
@@ -363,65 +161,42 @@ struct write : public ::testing::Test
...
     std::vector<std::vector<phred42>> quals
     {
         { "!##$"_phred42 },
         { "!##$&'()*+,-./+)*+,-)*+,-)*+,-)*+,BDEBDEBDEBDEBDEBDEBDEBDEBDEBDEBDEBDEBDEBDEBDBDDEBDBEEBEBE"_phred42 },
         { "!!*+,-./+*+,-./+!!FF!!"_phred42 },
     };
-    detail::sequence_file_output_format_REMOVEME<format_sam> format;
+
     sequence_file_output_options options;
+
```
rrahn: remove new line

test/unit/io/sequence_file/sequence_file_format_test_template.hpp
```diff
@@ -0,0 +1,246 @@
...
+    std::stringstream istream{this->standard_input};
+    sequence_file_input fin{istream, TypeParam{}};
+
+    auto it = fin.begin();
+    for (unsigned i = 0; i < 3; ++i, ++it)
+    {
+        EXPECT_TRUE((std::ranges::equal(get<field::SEQ>(*it), this->seqs[i])));
+        EXPECT_EQ(get<field::ID>(*it), this->ids[i]);
+        if constexpr (std::same_as<TypeParam, format_fastq> ||
+        	          std::same_as<TypeParam, format_sam>)
```
rrahn: fits not one one line?

test/unit/io/sequence_file/sequence_file_format_test_template.hpp
```diff
@@ -0,0 +1,246 @@
...
+    std::vector<std::string> ids
+    {
+        { "ID1" },
+        { "ID2" },
+        { "ID3 lala" },
+    };
+
+    std::vector<dna5_vector> seqs
+    {
+        { "ACGTTTTTTTTTTTTTTT"_dna5 },
```
rrahn: so every testable format has to match now exactly these inputs from the test data?
So I would propose making them constants and have the formats as well as this test data fixture use these constants. That makes adding a test much easier for the future.

test/unit/io/structure_file/format_vienna_test.cpp
```diff
@@ -55,16 +55,16 @@ struct read : public ::testing::Test
...
+    std::vector<rna5_vector> const expected_seq
     {
-        { "GCGGAUUUAGCUCAGUUGGGAGAGCGCCAGACUGAAGAUUUGGAGGUCCUGUGUUCGAUCCACAGAAUUCGCA"_rna4 },
-        { "UUGGAGUACACAACCUGUACACUCUUUC"_rna4 }
+        { "GCGGAUUUAGCUCAGUUGGGAGAGCGCCAGACUGAAGAUUUGGAGGUCCUGUGUUCGAUCCACAGAAUUCGCA"_rna5 },
+        { "UUGGAGUACACAACCUGUACACUCUUUC"_rna5 }
     };
 
-    std::vector<std::vector<dot_bracket3>> const expected_structure
+    std::vector<std::vector<wuss<51>>> const expected_structure
```
rrahn: Oo such a fundamental change? That was copy'n'paste error before?

include/seqan3/io/structure_file/format_vienna.hpp
smehringer: Currently we do not enforce the range to be a sized range (The default is std::vector though).
So std::distance would be linear again in case it is not sized. I don't know if we want to be restricted to that because we usually just want to fill the container with values?

include/seqan3/io/structure_file/format_vienna.hpp
smehringer: no. I changed it to distance

test/unit/io/sequence_file/sequence_file_format_genbank_test.cpp
smehringer: well the reading is done while the iterator `it` is incremented. 
I could move the incrementation from inside the for loop definition `()` to inside the loop scop and wrap it in no throw again but the test will fail if the operator throws anyway AND I think the error message without the expect is much more readable... (Besides we have a lot of functions throughout the library that may throw which we did not wrap in EXPECT_NO_THROW).

test/unit/io/sequence_file/sequence_file_format_genbank_test.cpp
smehringer: doesn't seem so but I initialised all of them

test/unit/io/sequence_file/sequence_file_format_genbank_test.cpp
smehringer: No there are not identical. They now test the same semantics but with the public interface.

test/unit/io/structure_file/format_vienna_test.cpp
smehringer: `wuss<51>` is the default structure alphabet if you use the public interface. The test would be much more bloated if I would add a traits object that changes the default alphabet to `dot_bracket` because automatic type deduction would not kick in if I specialise the file with its first template parameter. So I thought, since we do not test several alphabets yet and it should (and does) work with either alphabet, I might change it for simplicity.

test/unit/io/sequence_file/sequence_file_format_test_template.hpp
smehringer: No, since a test fixture does not need to inherit from `sequence_file_data` but may specify these variables itself. So this is flexible but still reduces code for the ones that can be tested with this because they do not need to defines the same thing again

include/seqan3/io/structure_file/format_vienna.hpp
rrahn: But that's the whole point. If we argue that we don't care if you use a non performant data type, so be it. But the default is always sized. Then we can as well use the information we already get returned from the copy algorithm instead of injecting it into another linear one. that means, if std::distance is already slow on the user provided type and we use it anyway, it doesn't matter if it is called one or two times. As a matter of fact, right now, it will always do some additional linear operation. So we didn't gain anything.

test/unit/io/sequence_file/sequence_file_format_test_template.hpp
rrahn: I still don't get it. 
>So this is flexible but still reduces code for the ones that can be tested with this because they do not need to defines the same thing again

A format has to be created to match these ids and sequences which is then read from in the public file based interface, right? So there need to be "file" definitions for exactly these values. And they have to be the same. So changing something here. For example, by realising that there was a bug within the qualities or something means I have to go through all the files, figure out which one is using these values and then adapt them by hand there as well. I suggest that this can be circumvented by having global values for these, which can be reused by the tests that want to use them. Of course if they overwrite their behavior then it is specific to the test. But I believe this is a critical aspect here.

test/unit/io/sequence_file/sequence_file_format_genbank_test.cpp
rrahn: Yes, but the relevant parts that test the corresponding throw statments are covered?

smehringer: yes

PR #1248, opened 2019-09-09
Massively reduce include graph within alphabet module.

marehr:
Right now nearly every small change effects the whole library, because everything includes everything. This PR reduces the number of includes enormously.

### Before:

Screenshot_2019-09-09 SeqAn3 seqan3 alphabet nucleotide dna15 hpp File Reference(1)

### After:

Screenshot_2019-09-09 SeqAn3 seqan3 alphabet nucleotide dna15 hpp File Reference

### DIFF
--- a/include/seqan3/alphabet/concept.hpp
+++ b/include/seqan3/alphabet/concept.hpp
@@ -16,7 +16,7 @@
 #include <seqan3/core/concept/cereal.hpp>
 #include <seqan3/core/concept/core_language.hpp>
 #include <seqan3/core/detail/customisation_point.hpp>
-#include <seqan3/core/detail/reflection.hpp>
+#include <seqan3/std/type_traits>
 
 // ============================================================================
 // forwards
@@ -544,7 +544,7 @@ struct assign_char_strictly_to_fn
     decltype(auto) operator()(seqan3::alphabet_char_t<alph_t> const r, alph_t & a) const
     {
         if (!seqan3::char_is_valid_for<alph_t>(r))
-            throw seqan3::invalid_char_assignment{seqan3::detail::get_display_name_v<alph_t>, r};
+            throw seqan3::invalid_char_assignment{"Alphabet", r};
 
         return seqan3::assign_char_to(r, a);
     }
--- a/include/seqan3/alphabet/concept.hpp
+++ b/include/seqan3/alphabet/concept.hpp
@@ -1044,5 +1044,3 @@ SEQAN3_CONCEPT writable_constexpr_alphabet =
 //!\endcond
 
 } // namespace seqan3::detail
-
-#include <seqan3/alphabet/detail/hash.hpp>
--- a/include/seqan3/alphabet/concept.hpp
+++ b/include/seqan3/alphabet/concept.hpp
@@ -17,9 +17,6 @@
 #include <seqan3/core/concept/core_language.hpp>
 #include <seqan3/core/detail/customisation_point.hpp>
 #include <seqan3/core/detail/reflection.hpp>
-#include <seqan3/core/type_traits/basic.hpp>
-#include <seqan3/core/type_traits/function.hpp>
-#include <seqan3/std/concepts>
 
 // ============================================================================
 // forwards
--- a/include/seqan3/alphabet/hash.hpp
+++ b/include/seqan3/alphabet/hash.hpp
@@ -1,0 +1,41 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \author Enrico Seiler <enrico.seiler AT fu-berlin.de>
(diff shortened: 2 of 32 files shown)


include/seqan3/alphabet/concept.hpp
```diff
@@ -14,12 +14,29 @@
...
-#include <seqan3/core/concept/core_language.hpp>
 #include <seqan3/core/detail/customisation_point.hpp>
-#include <seqan3/core/detail/reflection.hpp>
-#include <seqan3/core/type_traits/basic.hpp>
-#include <seqan3/core/type_traits/function.hpp>
-#include <seqan3/std/concepts>
+#include <seqan3/std/type_traits>
+
+// ============================================================================
+// obscure gcc 7, 8, 9 bug to get ADL working
```
eseiler: Is there a bug report for this?
And why would including small_vector fix this?

include/seqan3/alphabet/concept.hpp
```diff
@@ -546,7 +563,7 @@ struct assign_char_strictly_to_fn
     decltype(auto) operator()(seqan3::alphabet_char_t<alph_t> const r, alph_t & a) const
     {
         if (!seqan3::char_is_valid_for<alph_t>(r))
-            throw seqan3::invalid_char_assignment{seqan3::detail::get_display_name_v<alph_t>, r};
+            throw seqan3::invalid_char_assignment{"Alphabet", r};
```
eseiler: Mhm, this saves one include, but the error message will now be less meaningful?

include/seqan3/alphabet/hash.hpp
```diff
@@ -0,0 +1,42 @@
...
+#include <seqan3/alphabet/concept.hpp>
+
+namespace std
+{
+/*!\brief Struct for hashing a character.
+ * \ingroup alphabet
+ * \tparam alphabet_t The type of character to hash; Must model seqan3::semialphabet.
+ */
+template <typename alphabet_t>
+    //!\cond
```
eseiler:
```suggestion
//!\cond
```

include/seqan3/alphabet/hash.hpp
```diff
@@ -0,0 +1,42 @@
...
+    requires seqan3::semialphabet<alphabet_t>
+    //!\endcond
```
eseiler:
```suggestion
//!\endcond
```

include/seqan3/alphabet/quality/aliases.hpp
```diff
@@ -13,8 +13,8 @@
 #pragma once
 
 #include <seqan3/alphabet/concept.hpp>
-#include <seqan3/alphabet/quality/qualified.hpp>
 #include <seqan3/alphabet/quality/concept.hpp>
+#include <seqan3/alphabet/quality/qualified.hpp>
```
eseiler: phred42 < qualified

include/seqan3/alphabet/hash.hpp
marehr: Fixing your stuff :)

https://github.com/seqan/seqan3/commit/a6af6d2f49e947871d0a51db06d556620ca8d7ed

include/seqan3/alphabet/concept.hpp
marehr: I feel so stupid. This actually is what will be fixed by https://github.com/seqan/seqan3/pull/1228

include/seqan3/alphabet/concept.hpp
eseiler: Still open, but I don't have a strong feeling about this.

marehr: We need to do this, because `seqan3::detail::get_display_name_v<alph_t>` includes small string which in turn pulls in the complete range library. We want to avoid this.

eseiler:
There's a typo in one commit:
[FIX] way to**o** many includes in alphabet

include/seqan3/alphabet/concept.hpp
smehringer: Sorry to unresolve this but I am unsure in how important the include graph is in contrast to readable error messages? Can we discuss this?

side note: Did you chack the error output now? I think in your case it will say "Assigning H to an alphabet of type Alphabet would incur information loss." ... which is confusing I think

smehringer: Sorry to unresolve this but I am unsure in how important the include graph is in contrast to readable error messages? Can we discuss this?

side note: Did you chack the error output now? I think in your case it will say "Assigning H to an alphabet of type Alphabet would incur information loss." ... which is confusing I think

marehr:
I'm stuck, not sure how to progress this PR.

Not excluding 
`seqan3::detail::get_display_name_v<alph_t>` will result in increasing the include graph 
from


Screenshot from 2019-11-14 13-19-09

to

Screenshot from 2019-11-14 13-06-48

I would strongly suggest to take the changes as they are right now and later add the diagnostic version back by trimming out all the ranges dependencies.

rrahn:
I will merge this for now because it touches a lot of stuff. The error message issue is a valid point. Maybe we just change the error message to something more generic or provide a CPO for giving an alphabet a name but make it not part of the concept. So like in default if nothing is specified we can write something like: 
`Assigning H to the underlying alphabet would incur information loss.`
while with the CPO implemented we could do: 
`Assigning H to the dna alphabet would incur information loss.`
Where the CPO for dna resolves to `"dna"`. 
Or we simply make a generic error message like 
`Invalid character for the alphabet: <char>`
It would be then good to also output the stack trace but IIRC there is no standardised way to do so (depends on the compiler and such)

I make a discussion ticket and a card in the respective module.

PR #1254, opened 2019-09-10
[FEATURE] Allow convertible queries in search

### LINKED ISSUE #1246: Make index work with alphabets that are explicitly convertible to the index alph type
----

This is a follow up from #471.

eseiler:
Resolves #1246

### DIFF
--- a/include/seqan3/search/fm_index/bi_fm_index_cursor.hpp
+++ b/include/seqan3/search/fm_index/bi_fm_index_cursor.hpp
@@ -82,6 +82,8 @@ private:
     using sdsl_char_type = typename index_type::sdsl_char_type;
     //!\brief Type of the alphabet size in the underlying SDSL index.
     using sdsl_sigma_type = typename index_type::sdsl_sigma_type;
+    //!\brief Alphabet type of the index.
+    using index_char_type = typename index_t::char_type;
 
     //!\brief Type of the underlying FM index.
     index_type const * index;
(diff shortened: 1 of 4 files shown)


eseiler:
> So the only change is that you static_cast the value to `index_char_type` before you call rank or else on it correct?

Yes.

include/seqan3/search/fm_index/bi_fm_index_cursor.hpp
```diff
@@ -82,6 +82,8 @@ class bi_fm_index_cursor
     using sdsl_char_type = typename index_type::sdsl_char_type;
     //!\brief Type of the alphabet size in the underlying SDSL index.
     using sdsl_sigma_type = typename index_type::sdsl_sigma_type;
+    //!\brief Alphabet type of the index.
+    using index_char_type = typename index_t::char_type;
```
rrahn: Why is not called `index_alphabet_type` then? The naming is misleading.

eseiler: Indeed, but then I would propose to rename it from char to alphabet everywhere in the search / fm_index module where it refers to an actual alphabet and not some kind of char/wchar , etc.

smehringer: Then I would leave it like this to be consistent with the current index design and reevaluate changing it everywhere in another PR

eseiler:
@rrahn @smehringer Do you agree to do the renaming as a follow-up issue? I might combine it with fixing Doxygen warnings :)

smehringer:
I agree

eseiler:
@smehringer ready to be merged.
There is a card for the renaming task.

PR #1257, opened 2019-09-11
[DOC] Update howto porting with differences in the version check feature.

### LINKED ISSUE #1226: -DSEQAN_DISABLE_VERSION_CHECK missing from Seqan3
For Debian we'd like a cmake-level directive to turn off version checking, thanks!

smehringer:
Resolves #1226

### DIFF
--- a/doc/howto/porting/index.md
+++ b/doc/howto/porting/index.md
@@ -71,6 +71,23 @@ Some components that are rather independent and good first targets to start port
   * The argument parser.
   * Input and Output.
 
+## The argument parser version check feature
+
+The major change is that we do not provide a cmake-level directive to turn off version checking.
+Instead, the current implementation lets you disable the version check in the following ways:
+
+  1. The developer can disable the feature permanently on construction of the seqan3::argument_parser.
+  2. The administrator or user can deactivate the feature globally on their system via the environment variable
+     `SEQAN3_NO_VERSION_CHECK`.
+  3. The user can disable the feature temporarily for one day via a command-line option when calling the application.
+
+Otherwise, if non of the above happened, the user will be prompted at the first call of the application to select
+a configuration (Never/Always/Yes/No). However, if the application is executed in a non-terminal process,
+e.g within a cron job or a workflow, no prompt will be issued but instead the version check will be triggered in
+the background. This will happen no more than once per day.
+
+If you are interested in the new argument parser design, take a look at our tutorial \ref tutorial_argument_parser.
+
 ## SeqAn2 features missing from SeqAn3
 
 This is a non-exhaustive list of features that are currently missing, but **will be ported**:


doc/howto/porting/index.md
```diff
@@ -71,6 +71,23 @@ Some components that are rather independent and good first targets to start port
...
+Instead, the current implementation lets you disable the version check in the following ways:
+
+  1. The developer can disable the feature permanently on construction of the seqan3::argument_parser.
+  2. The administrator or user can deactivate the feature globally on their system via the environment variable
+     `SEQAN3_NO_VERSION_CHECK`.
+  3. The user can disable the feature temporarily for one day via a command-line option when calling the application.
+
+Otherwise, if non of the above happened, the user will be prompted at the first call of the application to select a
+configuration (Never/Always/Yes/No). However, if the application is executed in a non-terminal process,
+e.g within a cron job or a workflow, no prompt will be issued and a version check will be performed in the background
```
mr-c: This sentence could imply that there is some process that keeps running in the background and will issue a version check daily. I think you mean to convey that the version check happens no more than once a day in the non-terminal situation.

smehringer: Hm is this better?:
` However, if the application is executed in a non-terminal process,
e.g within a cron job or a workflow, no prompt will be issued but instead the version check will be triggered in the background once per day.`

rrahn: Maybe:
```
However, if the application is executed in a non-terminal process, e.g within a cron job or a workflow, no prompt will be issued but instead the version check will be triggered in the background. This solely happens once per day.
```

smehringer: @mr-c What do you think?

mr-c: One small tweak:

> This solely happens once per day.

to

> This will happen no more than once per day.

doc/howto/porting/index.md
```diff
@@ -71,6 +71,25 @@ Some components that are rather independent and good first targets to start port
...
+Instead, the current implementation lets you disable the version check in the following ways:
+
+  1. The developer can disable the feature permanently on construction of the seqan3::argument_parser.
+  2. The administrator or user can deactivate the feature globally on their system via the environment variable
+     `SEQAN3_NO_VERSION_CHECK`.
+  3. The user can disable the feature temporarily for one day via a command-line option when calling the application.
+
+Otherwise, if non of the above happened,
+the user will be prompted at the first call of the application to select a configuration (Never/Always/Yes/No).
+However, if the application is executed in a non-terminal process,
```
rrahn: why are you wrapping the end of line so early?

smehringer:
> maybe if you could use the full line length limit? That would be awesome nail_care

I usually break sentences by structure instead of pure line length. But I don't care..

PR #1264, opened 2019-09-13
[WIP] zip

eseiler:


h-2:
Wow, nice speed-up!

eseiler:
I reworked the benchmark to be more generic.
I wonder why zipping 2 vectors is slower than zipping 3 vectors :thinking:

h-2:
In the benchmark, can you `benchmark::DoNotOptimize( )` instead of the dummy?

eseiler:
@marehr Ready for review!

h-2:
Performance seems to have regressed significantly. Is this due to the common_tuple? Can you double-check this? 
If this makes such a difference, can you make the reference type be a regular tuple when none of the underlying ranges are output ranges?

If tuple types make a big difference, I would also be interested in seeing how the pod_tuple fares. If it turns out to be much better, we can also change it to allow non-pod types and possibly const-assignability.

test/snippet/range/views/zip.cpp
```diff
@@ -0,0 +1,19 @@
...
+{
+    std::vector<size_t> vec1{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10};
+    std::vector<std::string> vec2{"hello", "goodbye", "hello", "goodbye"};
+    auto rep = seqan3::views::repeat('L');
+
+    auto v = seqan3::views::zip(vec1, vec2, rep);
+    seqan3::debug_stream << v << '\n';  // Prints [(0,hello,L),(1,goodbye,L),(2,hello,L),(3,goodbye,L)]
+
+    *v.begin() = std::tuple{9, "MOO", 'P'};
+    seqan3::debug_stream << v << '\n';  // Prints [(9,MOO,P),(1,goodbye,P),(2,hello,P),(3,goodbye,P)]
```
h-2: Repeat is an output_range? o_O

test/unit/range/views/view_zip_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+    using tuple1_t = std::tuple<int, std::string, char>;
+    using tuple2_t = std::tuple<int, std::string>;
+    using result1_t = std::vector<tuple1_t>;
+    using result2_t = std::vector<tuple2_t>;
+
+    using rng_not_common_t = decltype(views::zip(vi, vs, vr));
+    using rng_common_t = decltype(views::zip(vi, vs));
+    using rng_const_t = decltype(views::zip(vi, vc));
+    using view_const_t = decltype(views::zip(vi, vc)) const;
+    using rng_view_const_t = decltype(views::zip(vi, vc)) const;
```
h-2: Where is the difference between the two?

test/snippet/range/views/zip.cpp
smehringer: yes. we talked about this and there was no significant reason to prohibit it.

h-2:
Any news on this? The alphabet changes have landed now.

You can request me for second review if you want!

eseiler:
I'm having trouble with this use case (alignment code):

```cpp
    seqan3::dna4_vector a{"AAAAA"_dna4};
    seqan3::dna4_vector b{"TTTTT"_dna4};
    std::tuple<seqan3::dna4_vector, seqan3::dna4_vector> d{a, b};
    auto w = ranges::views::single(d);
    auto v = views::zip(w, std::views::iota(0));
    v.end(); // fails
```

Using `std::views::iota(0, 5)` works, so it's some problem with being sized or not.
The actual use case is `views::zip(w, std::views::iota(0)) | views::chunk(1)` where it then complains that `operator|` doesn't exist.

h-2:
> I'm having trouble with this use case

I will try to have a look at it this week, otherwise we can have a look at it together next monday.

eseiler:
Superseded by https://github.com/seqan/seqan3/pull/2971

PR #1265, opened 2019-09-13
Fix/pairwise combine const

rrahn:
Adds common_tuple and common_pair alias for ranges versions.
Makes pairwise combine view also an output range and fixes an issue when the underlying range is const.

### DIFF
--- a/include/seqan3/range/views/pairwise_combine.hpp
+++ b/include/seqan3/range/views/pairwise_combine.hpp
@@ -14,6 +14,7 @@
 
 #include <cmath>
 
+#include <seqan3/core/common_tuple.hpp>
 #include <seqan3/range/concept.hpp>
 #include <seqan3/range/views/detail.hpp>
 #include <seqan3/std/ranges>
@@ -72,8 +73,8 @@ private:
         using underlying_val_t = typename std::iterator_traits<underlying_iterator_type>::value_type;
         //!\brief Alias for the reference type of the underlying iterator type.
         using underlying_ref_t = typename std::iterator_traits<underlying_iterator_type>::reference;
-    public:
 
+    public:
         /*!\name Associated types
          * \{
          */
@@ -83,7 +84,7 @@ private:
         //!\brief The value type.
         using value_type        = std::tuple<underlying_val_t, underlying_val_t>;
         //!\brief The reference type.
-        using reference         = std::tuple<underlying_ref_t, underlying_ref_t>;
+        using reference         = common_tuple<underlying_ref_t, underlying_ref_t>;
         //!\brief The pointer type.
         using pointer           = void;
         //!\brief The iterator category tag.
@@ -143,7 +144,7 @@ private:
         constexpr reference operator*() const
             noexcept(noexcept(*std::declval<underlying_iterator_type>()))
         {
-            return {*first_it, *second_it};
+            return reference{*first_it, *second_it};
         }
 
         /*!\brief Access the element at the given index
@@ -160,8 +161,7 @@ private:
         }
         //!\}
 
-        /*!
-ame Arithmetic operators
+        /*!\name Arithmetic operators
          * \{
          */
         //!\brief Pre-increment operator.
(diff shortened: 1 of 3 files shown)


test/unit/range/views/view_pairwise_combine_test.cpp
```diff
@@ -145,25 +145,25 @@ TYPED_TEST(pairwise_combine_iterator_test, associated_types)
     { // non-const view over non-const range
         using u_ref_t = typename std::iterator_traits<u_iter_t>::reference;
         EXPECT_TRUE((std::is_same_v<typename std::iterator_traits<v_iter_t>::reference,
-                                    std::tuple<u_ref_t, u_ref_t>>));
+                                    ranges::common_tuple<u_ref_t, u_ref_t>>));
```
eseiler:
```suggestion
                                    seqan3::common_tuple<u_ref_t, u_ref_t>>));
```

include/seqan3/range/views/pairwise_combine.hpp
```diff
@@ -143,7 +144,7 @@ class pairwise_combine_view : public ranges::view_interface<pairwise_combine_vie
         constexpr reference operator*() const
             noexcept(noexcept(*std::declval<underlying_iterator_type>()))
         {
-            return {*first_it, *second_it};
+            return reference{*first_it, *second_it};
```
eseiler: Doesn't work without the `reference`?

rrahn: Unfortunately not.

test/unit/range/views/view_pairwise_combine_test.cpp
```diff
@@ -353,11 +353,21 @@ TYPED_TEST(pairwise_combine_test, view_concept)
...
+    EXPECT_TRUE((std::ranges::output_range<typename TestFixture::view_t, std::tuple<char &, char &>>));
     EXPECT_EQ(std::ranges::bidirectional_range<TypeParam>,
               std::ranges::bidirectional_range<typename TestFixture::view_t>);
     EXPECT_EQ(std::ranges::sized_range<TypeParam>, std::ranges::sized_range<typename TestFixture::view_t>);
     EXPECT_EQ(std::ranges::random_access_range<TypeParam>, std::ranges::random_access_range<typename TestFixture::view_t>);
+
+    EXPECT_TRUE(std::ranges::input_range<typename TestFixture::const_view_t>);
+    EXPECT_TRUE(std::ranges::forward_range<typename TestFixture::const_view_t>);
+    EXPECT_TRUE(std::ranges::view<typename TestFixture::const_view_t>);
+    //TODO: Investigate, because this should be false. It cannot be used as a output range!
```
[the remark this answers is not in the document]
rrahn: That is the magic of the common_tuple/common_pair 🐰 🎩 . With a regular std::tuple this does not work, because of the writable concept which requires that `const proxy_t && = value_t` works.
> writable has the awkward const_­cast expressions to reject iterators with prvalue non-proxy reference types that permit rvalue assignment but do not also permit const rvalue assignment.
Consequently, an iterator type I that returns std​::​string by value does not model writable<I, std​::​string>.


And you want it to be an output range like e.g. the zip view, so you can do something like:
```cpp
auto zipped = views::zip(rng1, rng2, ...);
for (auto && elem : zipped)
   elem = std::tuple{val1, val2, ...};
```
So the zip view should preserve output_range if all of the contained ranges are output_ranges. To solve the todo, I would first wait for the ranges update and then it is probably a little nasty to find the reasons for this. I was looking into implementing the common_pair/tuple for our lib (we need anyway). And see if that helps. But that is beyond of this PR.

test/unit/range/views/view_pairwise_combine_test.cpp
```diff
@@ -478,6 +488,22 @@ TEST(pairwise_combine_fn_test, filter_input)
...
+
+    *v.begin() = std::tuple{'x', 'y'};
+
+    auto it = v.begin();
+    EXPECT_EQ(*it,   (std::tuple{'x', 'y'}));
+    EXPECT_EQ(*++it, (std::tuple{'x', 'c'}));
+    EXPECT_EQ(*++it, (std::tuple{'x', 'd'}));
+    EXPECT_EQ(*++it, (std::tuple{'y', 'c'}));
+    EXPECT_EQ(*++it, (std::tuple{'y', 'd'}));
+    EXPECT_EQ(*++it, (std::tuple{'c', 'd'}));
```
smehringer: Maybe add 
```cpp
EXPECT_EQ(orig, std::vector{'x', 'y', 'c', 'd'});
```

PR #1266, opened 2019-09-13
[DOC] Rename all instances of SeqAn3 to SeqAn.

rrahn:
Fixes SeqAn3 naming in the documentation where appropriate.

### DIFF
--- a/include/seqan3/argument_parser/detail/version_check.hpp
+++ b/include/seqan3/argument_parser/detail/version_check.hpp
@@ -409,7 +409,7 @@ public:
     static constexpr std::string_view unregistered_app = "UNREGISTERED_APP";
     //!\brief The message directed to the developer of the app if a new seqan3 version is available.
     static constexpr std::string_view message_seqan3_update =
-        "[SEQAN3 INFO] :: A new SeqAn3 version is available online.\n"
+        "[SEQAN3 INFO] :: A new SeqAn version is available online.\n"
         "[SEQAN3 INFO] :: Please visit www.github.com/seqan/seqan3.git for an update\n"
         "[SEQAN3 INFO] :: or inform the developer of this app.\n"
         "[SEQAN3 INFO] :: If you don't wish to receive further notifications, set --version-check OFF.\n\n";
--- a/doc/howto/subcommand_argument_parser/index.md
+++ b/doc/howto/subcommand_argument_parser/index.md
@@ -2,7 +2,7 @@
 
 [TOC]
 
-This HowTo shows you how to write an argument parser with subcommand like `git push` using SeqAn3.
+This HowTo shows you how to write an argument parser with subcommand like `git push` using SeqAn.
 
 \tutorial_head{Easy, 15 min, \ref tutorial_argument_parser, }
 
--- a/doc/howto/write_a_view/index.md
+++ b/doc/howto/write_a_view/index.md
@@ -2,7 +2,7 @@
 
 [TOC]
 
-This HowTo documents how to write a view using the standard library and some helpers from SeqAn3.
+This HowTo documents how to write a view using the standard library and some helpers from SeqAn.
 
 \tutorial_head{Difficult, 120 min, \ref tutorial_concepts\,\ref tutorial_ranges, }
 
--- a/doc/howto/write_an_alphabet/index.md
+++ b/doc/howto/write_an_alphabet/index.md
@@ -2,7 +2,7 @@
 
 [TOC]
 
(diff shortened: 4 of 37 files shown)


joshuak94:
Why are we renaming all of the documentation to SeqAn when the namespace will still be seqan3?

rrahn:
> Why are we renaming all of the documentation to SeqAn when the namespace will still be seqan3?

seqan3 is a name specific to a version. The next major version won't be a complete rewrite of the library (at least it should not, since that would mean we have done something terribly wrong). So in these places we meant to refer to SeqAn as a library not pinned to a specific version. Because if we move to seqan4 no one will see this and suddenly we refer to seqan3 in a seqan4 documentation. I left some namings to seqan3, e.g. SEQAN3_INFO in the argument parser. Since upgrading to  seqan4 would destroy backwards compatibility there must be a differentiation between there is a new version and there is a new version that breaks your code. So my intention here is to keep the SEQAN3_INFO as to make it visible for the user which version his application is building on.

include/seqan3/argument_parser/detail/version_check.hpp
```diff
@@ -192,7 +192,7 @@ class version_checker
                               " " +
                               out_file.string() +
                               " " +
-                              std::string{"http://seqan-update.informatik.uni-tuebingen.de/check/SeqAn3_"} +
+                              std::string{"http://seqan-update.informatik.uni-tuebingen.de/check/SeqAn_"} +
```
eseiler: Sure ?

smehringer: No !

rrahn: no that's over eagerly replaced.

rrahn: but why did no test catch this? Should there not be at least one failing test? Something, like a dummy app registered on the server?

smehringer: I cannot add a test because if the server does not respond because of some other issue (other than a wrong link) we will have a broken CI that has nothing to do with our library and we wanted to avoid this.

rrahn: Ok, let's say our general assumption is that the server is always online. Would it then not be of help to see that the test is failing. Otherwise, we might not even realize that the server is not responding for whatever reason.

smehringer: What you mean is tested, since I check if given that the server is online and I ping him for the seqan version then I expect a correct result. 
Actually, the change you introduced will NEVER raise a failure because `SeqAn` is just as valid in the URL as `SeqAn3` since we also track the old library version :wink:

rrahn: Well, not if there was a specific app we are expecting to see, right? Like, foo_bar which obviously should not also exist as SeqAn_foo_bar.  But IIRC there were this developer setting. Like if you are a developer and you are not registered there is a response like: Hey don't you want to register and benefit from update notifications? I would assume there is a test that checks for an existing app and would instead get the developer notification, since the server address is not invalid but not as expected?

smehringer: > I would assume there is a test that checks for an existing app and would instead get the developer notification, since the server address is not invalid but not as expected?

I am not sure if I understand you correctly but no, for the developer notification "`please register your app...`" the server call has to be successfull and return `UNKNOWN_APP`. So there will be no message if the server is not available or the call is wrong.

rrahn: > "please register your app..." the server call has to be successfull and return UNKNOWN_APP

But that's what I am referring to. If I change the server address in such a way that it is still valid so the call is successful but there was no app registered with this name because like above SeqAn_foo_bar and SeqAn3_foo_bar are two distinct names. So I assume there is a test that expects the answer of a successful call with a different outcome. Such as there is a new version. So this test would expect to see "NEW_VERSION" but actually gets, "UNKNOWN_APP", so it fails.

smehringer: lets discuss this in person

rrahn: We discussed this offline. so I mark this resolved.

smehringer:
Did you check the Jenkins failure?

rrahn:
> Did you check the Jenkins failure?

@smehringer it is unrelated due to the spurious buffer queue dead lock

smehringer:
ok. let's wait for the ranges PR. even though I think there is no conflict I don't want to risk hannes rage :rage2: :grin:

PR #1270, opened 2019-09-17
Some alphabet test restructuring, adding templates to mask test 

### LINKED ISSUE #1131: [follow-up#1083] add cigar and mask to tuple test
After refactoring the alphabet_tuple_test into a template, cigar and mask should be added/instantiated. This involves some work since there seems to be some problems @marehr ?
- [x] cigar (fixed by #1083)
- [ ] mask (soon fixed by #1270 )

smehringer:
Resolves #1131

### DIFF
--- a/test/unit/alphabet/adaptation/char_test.cpp
+++ b/test/unit/alphabet/adaptation/char_test.cpp
@@ -11,21 +11,22 @@
 
 #include <seqan3/alphabet/adaptation/all.hpp>
 
-#include "../alphabet_test_template.hpp"
 #include "../alphabet_constexpr_test_template.hpp"
+#include "../alphabet_test_template.hpp"
+#include "../semi_alphabet_constexpr_test_template.hpp"
+#include "../semi_alphabet_test_template.hpp"
 
 using namespace seqan3;
 
-// char32_t and wchar_t, too slow
-using fast_char_types = ::testing::Types<char, char16_t, char32_t, wchar_t>;
+using char_types = ::testing::Types<char, char16_t, char32_t, wchar_t>;
 
-INSTANTIATE_TYPED_TEST_CASE_P(char_adaptation, alphabet_, fast_char_types);
-INSTANTIATE_TYPED_TEST_CASE_P(char_adaptation, alphabet_constexpr, fast_char_types);
+INSTANTIATE_TYPED_TEST_CASE_P(char_adaptation, alphabet_, char_types);
+INSTANTIATE_TYPED_TEST_CASE_P(char_adaptation, semi_alphabet_test, char_types);
+INSTANTIATE_TYPED_TEST_CASE_P(char_adaptation, alphabet_constexpr, char_types);
+INSTANTIATE_TYPED_TEST_CASE_P(char_adaptation, semi_alphabet_constexpr, char_types);
 
 template <typename T>
 using char_adaptation = ::testing::Test;
-
-using char_types = ::testing::Types<char, char16_t, char32_t, wchar_t>;
 
 TYPED_TEST_CASE(char_adaptation, char_types);
 
--- a/test/unit/alphabet/nucleotide/rna4_test.cpp
+++ b/test/unit/alphabet/nucleotide/rna4_test.cpp
@@ -5,12 +5,16 @@
 // shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
 // -----------------------------------------------------------------------------------------------------
 
-#include "../alphabet_test_template.hpp"
 #include "../alphabet_constexpr_test_template.hpp"
+#include "../alphabet_test_template.hpp"
+#include "../semi_alphabet_constexpr_test_template.hpp"
+#include "../semi_alphabet_test_template.hpp"
 #include "nucleotide_test_template.hpp"
 
 INSTANTIATE_TYPED_TEST_CASE_P(rna4, alphabet_, rna4);
+INSTANTIATE_TYPED_TEST_CASE_P(rna4, semi_alphabet_test, rna4);
 INSTANTIATE_TYPED_TEST_CASE_P(rna4, alphabet_constexpr, rna4);
+INSTANTIATE_TYPED_TEST_CASE_P(rna4, semi_alphabet_constexpr, rna4);
 INSTANTIATE_TYPED_TEST_CASE_P(rna4, nucleotide, rna4);
 
 TEST(rna4, to_char_assign_char)
--- a/test/unit/alphabet/semi_alphabet_constexpr_test_template.hpp
+++ b/test/unit/alphabet/semi_alphabet_constexpr_test_template.hpp
@@ -1,0 +1,159 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+#include <gtest/gtest.h>
+
+#include <seqan3/alphabet/concept.hpp>
+
+using namespace seqan3;
+
+template <typename T>
+using semi_alphabet_constexpr = ::testing::Test;
+
+TYPED_TEST_CASE_P(semi_alphabet_constexpr);
+
+TYPED_TEST_P(semi_alphabet_constexpr, concept_check)
+{
+    EXPECT_TRUE(detail::constexpr_semialphabet<TypeParam>);
+    EXPECT_TRUE(detail::constexpr_semialphabet<TypeParam &>);
+
+    EXPECT_TRUE(detail::constexpr_semialphabet<TypeParam const>);
+    EXPECT_TRUE(detail::constexpr_semialphabet<TypeParam const &>);
+
+    EXPECT_TRUE(detail::writable_constexpr_semialphabet<TypeParam>);
+    EXPECT_TRUE(detail::writable_constexpr_semialphabet<TypeParam &>);
+
+    EXPECT_FALSE(detail::writable_constexpr_semialphabet<TypeParam const>);
+    EXPECT_FALSE(detail::writable_constexpr_semialphabet<TypeParam const &>);
+}
+
+TYPED_TEST_P(semi_alphabet_constexpr, default_value_constructor)
+{
+    [[maybe_unused]] constexpr TypeParam t0{};
+}
+
+TYPED_TEST_P(semi_alphabet_constexpr, global_assign_rank)
+{
+    constexpr size_t rank = 1 % alphabet_size<TypeParam>;
+    [[maybe_unused]] constexpr TypeParam t0{seqan3::assign_rank_to(rank, TypeParam{})};
+}
+
+TYPED_TEST_P(semi_alphabet_constexpr, global_to_rank)
+{
+    constexpr size_t rank = 1 % alphabet_size<TypeParam>;
+    constexpr TypeParam t0{seqan3::assign_rank_to(rank, TypeParam{})};
+    constexpr bool b = (to_rank(t0) == rank);
+    EXPECT_TRUE(b);
+}
+
+TYPED_TEST_P(semi_alphabet_constexpr, copy_constructor)
+{
+    constexpr alphabet_rank_t<TypeParam> rank = 1 % alphabet_size<TypeParam>;
+    constexpr TypeParam t1{seqan3::assign_rank_to(rank, TypeParam{})};
+
+    constexpr TypeParam t2{t1};
+    constexpr TypeParam t3(t1);
+    EXPECT_EQ(t1, t2);
+    EXPECT_EQ(t2, t3);
+}
+
+TYPED_TEST_P(semi_alphabet_constexpr, move_constructor)
+{
+    constexpr alphabet_rank_t<TypeParam> rank = 1 % alphabet_size<TypeParam>;
+    constexpr TypeParam t0{seqan3::assign_rank_to(rank, TypeParam{})};
+    constexpr TypeParam t1{t0};
+
+    constexpr TypeParam t2{std::move(t1)};
+    constexpr TypeParam t3(std::move(t2));
+    EXPECT_EQ(t2, t0);
+    EXPECT_EQ(t3, t0);
+}
+
+TYPED_TEST_P(semi_alphabet_constexpr, copy_assignment)
+{
+    constexpr size_t rank = 1 % alphabet_size<TypeParam>;
+    constexpr TypeParam t0{seqan3::assign_rank_to(rank, TypeParam{})};
+    // constexpr context:
+    constexpr TypeParam t3 = [&] () constexpr
+    {
+        TypeParam t1{seqan3::assign_rank_to(rank, TypeParam{})};
+        TypeParam t2{};
+        t2 = t1;
+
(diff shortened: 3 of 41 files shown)


include/seqan3/alphabet/adaptation/char.hpp
```diff
@@ -94,10 +94,10 @@ struct alphabet<char_type>
      * \param[in,out] chr The alphabet letter that you wish to assign to.
      * \returns A reference to the alphabet letter you passed in.
      */
-    static constexpr char_type & assign_rank_to(decltype(alphabet::to_rank(char_type{})) const rank,
-                                                char_type & chr) noexcept
+    static constexpr char_type & assign_rank_to(size_t const rank, char_type & chr) noexcept
```
eseiler: Did something break when not doing this?

test/unit/alphabet/semi_alphabet_constexpr_test_template.hpp
```diff
@@ -0,0 +1,168 @@
...
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+#include <gtest/gtest.h>
+
+#include <seqan3/alphabet/concept.hpp>
+
+#if SEQAN3_WITH_CEREAL
```
eseiler: I don't see cereal tested anywhere?
Also, this can be replaced with `#include <seqan3/test/cereal.hpp>`

test/unit/alphabet/semi_alphabet_test_template.hpp
```diff
@@ -0,0 +1,187 @@
...
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+#include <gtest/gtest.h>
+
+#include <seqan3/alphabet/concept.hpp>
+#include <seqan3/alphabet/exception.hpp>
+#include <seqan3/core/concept/core_language.hpp>
+// #include <seqan3/test/pretty_printing.hpp>
```
eseiler: artefact?

include/seqan3/alphabet/adaptation/char.hpp
smehringer: yes, but not visually in a test: when I assign a rank that is larger than can be modelled by rank_type the value is implicitly cast and the assert would not catch the misuse.

test/unit/alphabet/semi_alphabet_constexpr_test_template.hpp
smehringer: good point... I cannot see it either. I copied this over from the alphabet_test. I'll check if I can remove them

test/unit/alphabet/semi_alphabet_test_template.hpp
smehringer: ops. I think this caused problems with the mask alphabet because it is no debug_streamable. I might need to change that

include/seqan3/alphabet/adaptation/char.hpp
h-2: This is not generic, rank types can be signed as well. Also we do not specify in the concept that the validity of the rank is asserted so we cannot enforce this on every alphabet here. Some alphabet may handle out-of-bounds ranks differently.

smehringer: I asked you this on Gitter :(
You actually had an out commented test in the alphabet test template that said `todo assert death when rank is > alphabet size` 
That's why I came across this issue. How should we handle this then?

h-2: > I asked you this on Gitter :(

Well, I also said I am not available 24h/day.

> todo assert death when rank is > alphabet size

That would only work in debug anyway. It's not something sensible to test. At most it should be tested directly on the member function and not the free function.

> That's why I came across this issue. How should we handle this then?

I am not yet sure it is an issue. Assigning a value of 16543 to a dna4 letter is undefined behaviour. If cast to `uint8_t` makes this resolve to a value that is accepted that is not a problem per se. If we want to catch that we need to change all the assignment functions to handle the largest possible type, assert on that and then always downcast. That costs performance even in NDEBUG.

smehringer: alright, then I'll remove the test and the changes that correspond to this.

test/unit/core/debug_stream_test.cpp
```diff
@@ -11,6 +11,7 @@
 #include <string>
 
 #include <seqan3/alphabet/nucleotide/all.hpp>
+#include <seqan3/alphabet/mask/mask.hpp>
```
eseiler: Order

test/unit/alphabet/adaptation/char_test.cpp
```diff
@@ -13,14 +13,18 @@
 
 #include "../alphabet_test_template.hpp"
 #include "../alphabet_constexpr_test_template.hpp"
+#include "../semi_alphabet_test_template.hpp"
+#include "../semi_alphabet_constexpr_test_template.hpp"
 
 using namespace seqan3;
 
 // char32_t and wchar_t, too slow
```
rrahn: this comment suggest something which is not reflected in the code?

test/unit/alphabet/composite/semialphabet_any_test.cpp
```diff
@@ -9,6 +9,8 @@
 
 #include "../alphabet_test_template.hpp"
 #include "../alphabet_constexpr_test_template.hpp"
+#include "../semi_alphabet_test_template.hpp"
+#include "../semi_alphabet_constexpr_test_template.hpp"
```
rrahn: You added these test templates but nothing else in the code? Wrong includes, separated includes? what happened?

test/unit/alphabet/nucleotide/rna4_test.cpp
```diff
@@ -7,10 +7,14 @@
 
 #include "../alphabet_test_template.hpp"
 #include "../alphabet_constexpr_test_template.hpp"
+#include "../semi_alphabet_test_template.hpp"
+#include "../semi_alphabet_constexpr_test_template.hpp"
```
rrahn: just seeing i here but it concerns all files. The ALPHABETICAL ORDER 😱 💥  🤣

test/unit/alphabet/semi_alphabet_constexpr_test_template.hpp
```diff
@@ -0,0 +1,156 @@
...
+using namespace seqan3;
+
+template <typename T>
+using semi_alphabet_constexpr = ::testing::Test;
+
+TYPED_TEST_CASE_P(semi_alphabet_constexpr);
+
+TYPED_TEST_P(semi_alphabet_constexpr, concept_check)
+{
+    EXPECT_TRUE(detail::constexpr_semialphabet<TypeParam   >);
```
rrahn: So I can now read it better because the bracket closes with the one below 🤦‍♂ (sorry this irony had to be 😁 )
However, I would be in strong favor of removing these empty spaces at the end. 💅

test/unit/alphabet/semi_alphabet_constexpr_test_template.hpp
```diff
@@ -0,0 +1,156 @@
...
+}
+
+TYPED_TEST_P(semi_alphabet_constexpr, default_value_constructor)
+{
+    [[maybe_unused]] constexpr TypeParam t0{};
+}
+
+TYPED_TEST_P(semi_alphabet_constexpr, copy_constructor)
+{
+    constexpr TypeParam t1{};
```
rrahn: Maybe with an assigned value? I think the test has then more value

test/unit/alphabet/semi_alphabet_constexpr_test_template.hpp
```diff
@@ -0,0 +1,156 @@
...
+    constexpr TypeParam t2{std::move(t1)};
+    EXPECT_EQ(t2, t0);
+    constexpr TypeParam t3(std::move(t2));
+    EXPECT_EQ(t3, t0);
+}
+
+TYPED_TEST_P(semi_alphabet_constexpr, global_assign_rank)
+{
+    constexpr size_t rank = 1 % alphabet_size<TypeParam>;
+    [[maybe_unused]] constexpr TypeParam t0{assign_rank_to(rank, TypeParam{})};
```
rrahn: you should always make a qualified call to the CPO. seqan3::assign_rank_to. The same for all other places.

test/unit/alphabet/semi_alphabet_test_template.hpp
```diff
@@ -0,0 +1,184 @@
...
+
+TYPED_TEST_P(semi_alphabet_test, alphabet_size_)
+{
+    EXPECT_GT(alphabet_size<TypeParam>, 0u);
+}
+
+TYPED_TEST_P(semi_alphabet_test, default_value_constructor)
+{
+    [[maybe_unused]] TypeParam t1;
+    [[maybe_unused]] TypeParam t2{};
```
rrahn: fun test. Maybe check the default value of it. also if this means that you might test some other behaviour as well. But since both tests would fail it is still easy to track what might fail and what not.

test/unit/alphabet/semi_alphabet_test_template.hpp
```diff
@@ -0,0 +1,184 @@
...
+}
+
+TYPED_TEST_P(semi_alphabet_test, global_assign_rank_to)
+{
+    // this double checks the value initialisation
+    EXPECT_EQ((assign_rank_to(0, TypeParam{})), TypeParam{});
```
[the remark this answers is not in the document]
rrahn: and all other places

test/unit/alphabet/semi_alphabet_constexpr_test_template.hpp
smehringer: 1. This is copy and pasted from hannes tests. It was just moved into a different file.
2. Considering that you and marcel were the ones complaining about too much irony in review comments, you seem to use a lot of irony in your last 5 reviews of my code. :wink:

test/unit/alphabet/adaptation/char_test.cpp
smehringer: I don't understand your issue?
As the comment says char32_t and wchar_t are too slow to test in the test template because iterating over every possible rank value takes for ever.

test/unit/alphabet/composite/semialphabet_any_test.cpp
smehringer: This is funny because the other test templates were also not used. I did this via a big regex so I did not catch this

test/unit/alphabet/nucleotide/rna4_test.cpp
smehringer: thanks for spotting. good thing editors have regex support ;)

test/unit/alphabet/semi_alphabet_constexpr_test_template.hpp
smehringer: These tests are copy and pasted from the alphabet_test_template. Changing them is not in the scope of this PR I think

rrahn: But what is the value of this test. If I have written a shitty copy constructor that in fact just calls he default constructor I won't be able to see anything. I think that should be covered. And also corrected from the source of this template if it is wrong there as well.

test/unit/alphabet/semi_alphabet_test_template.hpp
rrahn: If it is not even used the compiler will very likely remove it anyway. So if it is just for testing that there is a default constructor you can do `std::is_default_constructible_v`. At least you can test for it. And instead of a compilation error you will get a failing test.

test/unit/alphabet/adaptation/char_test.cpp
smehringer: ah true. then the comment can go :+1:

test/unit/alphabet/semi_alphabet_constexpr_test_template.hpp
smehringer: > But what is the value of this test. If I have written a shitty copy constructor that in fact just calls he default constructor I won't be able to see anything. I think that should be covered.

So you are suggesting that the variable should hold a value != the default? 
I see your point but it is also a little against the unit test design because in order to (generically) assign a different value than the default I would need to call `assign_rank_to`. So I would not know whether the copy construction failed or the assign_rank_to function? 

> And also corrected from the source of this template if it is wrong there as well.

Sorry, I meant "this was moved here from the alphabet_test_template". So this is the only place.

smehringer: I changed them.

> And also corrected from the source of this template if it is wrong there as well.

Sorry, I meant "there were **moved** here from the former alphabet_test_template", so there is only one place to change them.

test/unit/alphabet/semi_alphabet_constexpr_test_template.hpp
```diff
@@ -0,0 +1,156 @@
...
+    EXPECT_TRUE(detail::writable_constexpr_semialphabet<TypeParam   >);
+    EXPECT_TRUE(detail::writable_constexpr_semialphabet<TypeParam & >);
+
+    EXPECT_FALSE(detail::writable_constexpr_semialphabet<TypeParam const   >);
+    EXPECT_FALSE(detail::writable_constexpr_semialphabet<TypeParam const & >);
+}
+
+TYPED_TEST_P(semi_alphabet_constexpr, default_value_constructor)
+{
+    [[maybe_unused]] constexpr TypeParam t0{};
```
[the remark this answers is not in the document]
rrahn: Ok, I just checked we have: `is_constexpr_default_constructible_v`
I think the test should then be:
```suggestion
    EXPECT_TRUE(is_constexpr_default_constructible_v<TypeParam>);
```

test/unit/alphabet/semi_alphabet_constexpr_test_template.hpp
```diff
@@ -0,0 +1,159 @@
...
+}
+
+TYPED_TEST_P(semi_alphabet_constexpr, global_assign_rank)
+{
+    constexpr size_t rank = 1 % alphabet_size<TypeParam>;
+    [[maybe_unused]] constexpr TypeParam t0{seqan3::assign_rank_to(rank, TypeParam{})};
```
rrahn:
```suggestion
    constexpr TypeParam t0{seqan3::assign_rank_to(rank, TypeParam{})};
    EXPECT_EQ(t0, TypeParam{rank});
```

test/unit/alphabet/semi_alphabet_test_template.hpp
```diff
@@ -0,0 +1,184 @@
...
+}
+
+TYPED_TEST_P(semi_alphabet_test, alphabet_size_)
+{
+    EXPECT_GT(alphabet_size<TypeParam>, 0u);
+}
+
+TYPED_TEST_P(semi_alphabet_test, default_value_constructor)
+{
+    EXPECT_TRUE((std::is_default_constructible_v<TypeParam>));
```
rrahn: I think this is redundant, since the second test checks already if it is default constructible, but also if it doesn't throw.

test/unit/alphabet/semi_alphabet_constexpr_test_template.hpp
smehringer: You cannot directly construct an alphabet instance from its rank. There is no way we can test the value except with `to_rank`

PR #1271, opened 2019-09-17
[FIX,DOC] Doxygen 1.8.16 warnings

### LINKED ISSUE #754: doxygen 1.8.16 throws some warnings

eseiler:
Resolves #754

The more esoteric stuff:
 * After opening a group, there usually needs to be a new line unless the documentation comment style changes from line (`//!`) to block (`/*!`) or vice versa. A member also cannot be documented in the same documentation block as the group was opened in.
* If an entire function is excluded via `//!\cond`, we cannot `//!\cond` the requires clause since we did not close the last `//!\cond` block that uses the same condition (in this case no condition).
* In `alphabet/aminoacid/translation.hpp` I needed to change the template parameter name to help doxygen resolve the function signature. Otherwise there would be two functions with the same signature (parameter name does not count, parameter template does count) and it would only take the last overload and complain about multiple or missing parameter documentation.
* For travis, we need to change from `texlive-latex-base` to `texlive-latex-extra`, because the LaTex package `newunicodechar` is now required. Usually it would be enough to install base and run `tlmgr install newunicodechar` ... but this is again broken and would need workarounds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,7 +89,7 @@ matrix:
     addons:
       apt:
         # adds epstopdf, ghostscript, latex
-        packages: ['texlive-font-utils', 'ghostscript', 'texlive-latex-base']
+        packages: ['texlive-font-utils', 'ghostscript', 'texlive-latex-extra']
     env:
       - BUILD=documentation
       - CMAKE_VERSION=3.7.2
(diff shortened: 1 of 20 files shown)


smehringer:
Nice, does this fix all of the warnings?

smehringer:
> For travis, we need to change from texlive-latex-base to texlive-latex-extra, because the LaTex package newunicodechar is now required. Usually it would be enough to install base and run tlmgr install newunicodechar ... but this is again broken and would need workarounds.

why are we building in latex anyway? isn't html enough?

eseiler:
We need to render the formulas

PR #1286, opened 2019-10-06
Fix/parallel queue dead lock

### LINKED ISSUE #1081: Investigate dead lock of dynamic buffer queue on linux
An empty dynamic buffer queue seems to dead lock if too many readers are acquiring the shared lock. From some first investigations it seems that the producer thread acquiring the unique lock for the shared mutex is never locking because there are always consumer threads in the shared lock.

Things to further investigate:
 * Is the shared mutex in the standard implementation starvation free?
 * Are all consumers leaving the shared lock or is there at least one thread always locking the shared lock?
 * Is a non shared implementation with a normal mutex and uniqe locks in the consumer section also dead locking?

rrahn:
Fixes #1081 

Note I couldn't reproduce this particular scenario described in #1081 anymore but found an issue with the test of the buffer queue which resulted in the spurious dead locks. This had nothing to do with the buffer queue implementation so far. But I did refactor the buffer queue in order to make the whole code more readable. So this PR will close this issue since all tests should now work reliable (fingers crossed). When working with the distribution of the alignment tasks more closely and this problem should reappear we need to further investigate.

### DIFF
--- a/include/seqan3/contrib/parallel/buffer_queue.hpp
+++ b/include/seqan3/contrib/parallel/buffer_queue.hpp
@@ -20,6 +20,7 @@
 #include <type_traits>
 #include <vector>
 
+#include <seqan3/core/bit_manipulation.hpp>
 #include <seqan3/core/parallel/detail/spin_delay.hpp>
 #include <seqan3/core/type_traits/range.hpp>
 #include <seqan3/range/container/concept.hpp>
@@ -47,6 +48,42 @@ enum struct buffer_queue_policy : uint8_t
     dynamic
 };
 
+// Ringbuffer implementation:
+// The underlying buffer has size (number of actual elements + 1). This is a trick to easily check if the queue is empty
+// or full. Furthermore, the ring buffer uses 4 pointers. The actual push_back and pop_front position as well as
+// the pending push_back and pop_front position. The latter indicates the position that have been advanced concurrently
+// by multiple threads from either end.
+//
+// head: position to read/extract from the queue (first inserted elment) => pop_front_position
+// tail: position where to write to/push new elements to the queue => push_back_position
+// head_read: The actual position after x threads concurrently popped from the queue. => pending_pop_front_position
+// tail_write: The actual position after x threads concurrently pushed to the queue. => pending_push_back_position
+//  [  ?  ]  [  4  ]  [  3  ]  [  8  ]  [  0  ]  [  x  ]  [  ?  ]
+//                       |                          ^
+//                       v                          |
+//             head            headRead   tail  tailWrite
+//
+// valid buffer between  [headRead, tail)
+// currently filled      [tail, tailWrite)
+// currently removed     [head, headRead)
+//
+// State: empty = (head == tail)
+//  [  ?  ]  [  ?  ]  [  ?  ]  [  ?  ]  [  ?  ]  [  ?  ]  [  ?  ]
+//                                        tail
+//                                        head
+// The head is on the same position as tail.
+// This means that currently no element is in the buffer.
+
+// State: full = (tail + 1 == head)
+//  [  2  ]  [  4  ]  [  3  ]  [  ?  ]  [  8  ]  [  0  ]  [  7  ]
+//                               tail
+//                                        head
+// The tail is one position before the head.
+// This means that currently no element can be added to the buffer since it is full.
+// Strategies are to either wait until some elements have been popped or to expand the capacity of the
+// queue by one, inserting the element at the current tail position and moving all elements starting from head one
+// position to the right.
+
 template <std::semiregular value_t,
           sequence_container buffer_t = std::vector<value_t>,
           buffer_queue_policy buffer_policy = buffer_queue_policy::dynamic>
@@ -72,15 +109,15 @@ public:
     // you can set the initial capacity here
     explicit buffer_queue(size_type const init_capacity)
     {
-        data.resize(init_capacity + 1);
-        roundSize = static_cast<size_type>(1) << static_cast<size_type>(std::log2(data.size() - 1) + 1);
+        buffer.resize(init_capacity + 1);
+        ring_buffer_capacity = seqan3::detail::next_power_of_two(buffer.size());
     }
 
     template <std::ranges::input_range range_type>
         requires std::convertible_to<value_type_t<range_type>, value_type>
     buffer_queue(size_type const init_capacity, range_type && r) : buffer_queue{init_capacity}
     {
-        std::ranges::copy(r, std::ranges::begin(data));
+        std::ranges::copy(r, std::ranges::begin(buffer));
     }
 
     /*!\name Waiting operations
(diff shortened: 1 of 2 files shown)


include/seqan3/contrib/parallel/buffer_queue.hpp
```diff
@@ -191,46 +228,88 @@ class buffer_queue
...
         std::unique_lock write_lock(mutex);
-        return headPos == tailPos;
+        return pop_front_position == push_back_position;
     }
 
     bool is_full() const noexcept
     {
         std::unique_lock write_lock(mutex);
-        return size_impl() == data.max_size();
+        return !has_free_slots(pop_front_position, push_back_position);
```
joergi-w: The private member function `has_free_slots` is a simple inequality. Wouldn't it be easier to insert the inequality straight here (and avoid the function call and copying the two variables)?

include/seqan3/contrib/parallel/buffer_queue.hpp
```diff
@@ -191,46 +228,88 @@ class buffer_queue
...
+    /*!\brief Checks if enough slots are available for the `push_back_position` to write to the ring buffer.
+     * \param[in] pop_front_position The thread local position to read from.
+     * \param[in] push_back_position The thread local position to write to the buffer.
+     *
+     * \details
+     *
+     * The seqan3::contrib::buffer_queue::cyclic_increment ensures that the `push_back_position` is at least
+     * `ring_buffer_capacity` many slots ahead of `pop_front_position` if the queue is full.
+     */
+    constexpr bool has_free_slots(size_type const pop_front_position, size_type const push_back_position) const
```
joergi-w: 1. The variables `pop_front_position` and `push_back_position` are member variables, so you do not need to pass them as arguments to `has_free_slots` as they are already known...
2. This function needs a write lock, I assume.

include/seqan3/contrib/parallel/buffer_queue.hpp
```diff
@@ -281,47 +360,53 @@ inline bool buffer_queue<value_t, buffer_t, buffer_policy>::overflow(value2_t &&
...
+    // Expects no pending pushes or pops in unique lock.
+    assert(push_back_position == this->pending_push_back_position);
+    assert(pop_front_position == this->pending_pop_front_position);
 
     bool valueWasAppended = false;
 
     // did we reach the capacity limit (another thread could have done the upgrade already)?
-    if (cyclic_increment(tailPos, cap, roundSize) >= headPos + roundSize)
+    // buffer is full if tail_pos + 1 == head_pos
+    if (!has_free_slots(pop_front_position, cyclic_increment(push_back_position)))
```
joergi-w: use the `is_full` member function?

test/unit/contrib/parallel/buffer_queue_parallel_test.cpp
```diff
@@ -189,9 +190,10 @@ void test_buffer_queue_wait_throw(size_t initialCapacity)
                         push_status = ex;
                     }
                 }
-               // printf("stop writer #%ld %lu\n", tid, offset_end - offset);
+                // printf("stop writer #%ld %lu\n", tid, offset_end - offset);
                 // Last writer! No more values will come, so we close the queue.
-                if (registered_writer.fetch_sub(1) == 1)
+                ++registered_writer;
```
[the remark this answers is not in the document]
rrahn: Yes, this is necessary and the actual root of the dead lock. Otherwise I had to introduce a separate counting variable. The reason is, that the reader threads wait for the writers to register and the other way around. In order to guarantee at least some overlap between concurrent write and read. So the reader wait for all writers to register. The same do the writers. But, there might be the risk that between the registered_reader being incremented and checking if all writers are registered one writer might have finished it's job and already decremented the writer count. Thus the reader will wait forever. By inverting the decrement to an increment here we avoid this dead lock.

include/seqan3/contrib/parallel/buffer_queue.hpp
rrahn: Well, I think reading `push_back_position == pop_front_position + ring_buffer_capacity` is really hard to understand for others. It requires to understand the complete logic of the implementation of the ring buffer which is really hard, especially when trying to reason about certain semantics of the code. So this check is refactored on purpose to make it more understandable what happens here. Also there is now a documentation behind the rational of this implementation.

include/seqan3/contrib/parallel/buffer_queue.hpp
rrahn: Well, you need to be careful here. The member functions are atomics and we use specifically the thread local versions inside of the functions here. This is important since the state of the atomic might have been changed since a particular thread read a state of the atomic until it uses it. Thus, the thread local variables are not the same as the global atomic ones.

include/seqan3/contrib/parallel/buffer_queue.hpp
rrahn: Well, that wouldn't be possible since a) is_full also acquires a unique lock and you can;t do that b) the question is if it is full after the push_back_position was incremented by one. That is not the semantic of the is_full member function.

rrahn:
@joergi-w I answered on all you comments. I feel, that still a lot of misunderstanding exists in the implementation of the queue. Should we discuss this in person? Maybe the documentation should be improved or maybe there are just some misunderstandings about the usage?

rrahn:
@joergi-w  I  updated a little the documentation and variable naming.

include/seqan3/contrib/parallel/buffer_queue.hpp
```diff
@@ -47,6 +48,42 @@ enum struct buffer_queue_policy : uint8_t
...
+// State: full = (tail + 1 == head)
+//  [  2  ]  [  4  ]  [  3  ]  [  ?  ]  [  8  ]  [  0  ]  [  7  ]
+//                               tail
+//                                        head
+// The tail is one position before the head.
+// This means that currently no element can be added to the buffer since it is full.
+// Strategies are to either wait until some elements have been popped or to expand the capacity of the
+// queue by one, inserting the element at the current tail position and moving all elements starting from head one
+// position to the right.
+
```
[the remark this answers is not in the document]
rrahn: Well, yes this was backported from SeqAn2 into the contrib directory. We said everything in contrib does not need to be documented. We excluded it in the doxygen builds. But at some point I will incorporate it into SeqAn3 when the time permits. But all documentation I did already in doxygen mode so it is easier to incorporate it anyway

include/seqan3/contrib/parallel/buffer_queue.hpp
```diff
@@ -436,7 +510,6 @@ inline queue_op_status buffer_queue<value_t, buffer_t, buffer_policy>::try_pop(v
  *                            @endlink tag can only be used if one thread calls <tt>appendValue</tt> at a time.
  *                            Default is @link ParallelismTags#Parallel @endlink.
  */
-//
 template <typename value_t, typename buffer_t, buffer_queue_policy buffer_policy>
 template <typename value2_t>
```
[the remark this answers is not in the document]
rrahn: no, the first one is always the template argument of the class the second one is the template argument of the function

include/seqan3/contrib/parallel/buffer_queue.hpp
```diff
@@ -191,46 +228,90 @@ class buffer_queue
...
         std::unique_lock write_lock(mutex);
-        return size_impl() == data.max_size();
+        return is_ring_buffer_exhausted(pop_front_position, push_back_position);
     }
 
     size_type size() const noexcept
     {
         std::unique_lock write_lock(mutex);
-        return size_impl();
+        size_type number_of_elements = push_back_position - pop_front_position;
```
smehringer: are you sure that this never turns negative and cause an underflow?
Maybe assert LEQ beforehand

rrahn: Boah, I need to think. But I guess you are right. In general we assume that pop_front position is always before the push_back position, but since both positions always increasing it could happen that suddenly the push_back_position overlfows starting from 0 again. this is ok from the algotithm but would here result in a wrong size. In general I do not see many useful things you can do with the size on the concurrent queue, but it should be correct anyway.

smehringer:
[Pull requests](https://trello.com/c/9syzPMX5/8-pull-requests)

PR #1289, opened 2019-10-09
Misc/refactor alignment algorithm final 3

### LINKED ISSUE #1259: Add debug mode and output score and trace matrix
For better debugging it would be great to have the trace matrix and the score matrix outputted as an alignment result.
This means to add these options to the alignment config or make a  new debugging config.

### LINKED ISSUE #1260: Improve code readability in alignment algorithm
Clean up the alignment algorithm implementation. At the moment all internal types are modeled as tuples. But the more complexity is added to the algorithm the harder it gets to read the code. Therefore it would be good introduce strong types so we have named accessors to the different values.In addition it might be necessary to formulate concrete concepts on the types so that the code can be used with different implementations generically.

### LINKED ISSUE #1261: refactor traceback parser
At the moment we use a traceback parsing algorithm for banded and unbanded, which are very alike. And in the case of simd alignment and block based processing this should be refactored into it's own class. This needs to be synchronised with the edit-distance algorithm as well.

### LINKED ISSUE #1263: Some improvements to align result config
Design the alignment result options in such a way, that they are ordered and include the previous options. This is to reduce this code:

```cpp
if constexpr(config_t::template exists<align_cfg::result<with_score_type>>())
{
res.score = score;
}
else if constexpr (config_t::template exists<align_cfg::result<with_end_position_type>>())
{
res.score = score;
res.end_coordinate = end;
}
...
``` 

to 
```cpp
constexpr auto res_key = decltype(get<align_cfg::result>(config).value)::option_key;
if constexpr(res_key >= align_cfg::result::score_key)
{
res.score = score;
}
if constexpr (res_key >= align_cfg::result::end_coordinate_key)
{
// res.score = score; not needed as already included by the previous if statement
res.end_coordinate = end;
}
...
```

rrahn:
Supersedes #1277 

fixes #1259 
fixes #1260 
fixes #1261 
fixes #1263


include/seqan3/alignment/pairwise/alignment_configurator.hpp
```diff
@@ -141,45 +145,38 @@ struct alignment_configurator
...
+                return alignment_trace_matrix_full_banded<trace_t>{};
+            else
+                return alignment_trace_matrix_full<trace_t>{};
         }
 
     public:
         //!\brief The matrix policy based on the configurations given by `config_type`.
-        using type = decltype(select<config_type>());
+        using type = deferred_crtp_base<alignment_matrix_policy,
+                                        decltype(select_score_matrix()),
```
smehringer: is abusing static function return types really nicer than a `std::conditional<>`?
Are you using the selection somewhere else?

include/seqan3/alignment/pairwise/alignment_configurator.hpp
```diff
@@ -141,45 +145,38 @@ struct alignment_configurator
 
     /*!\brief Transformation trait that chooses the correct matrix policy.
      * \tparam config_type The configuration for which to select the correct matrix policy.
-     * \tparam trait_types A template parameter pack with additional traits to augment the selected policy.
+     * \tparam score_t The type used for the score matrix.
+     * \tparam trace_t The type used for the trace matrix.
      */
-    template <typename config_type, typename score_allocator_t, typename trace_allocator_t>
+    template <typename config_type, typename score_t, typename trace_t>
     struct select_matrix_policy
```
smehringer: hm.. we usually do not use verbs in the naming of type traits. Why not `matrix_policy`?

include/seqan3/alignment/pairwise/policy/affine_gap_init_policy.hpp
```diff
@@ -64,109 +64,103 @@ class affine_gap_init_policy
      * \tparam        cell_t       The underlying cell type.
      * \tparam        cache_t      The type of the cache.
      * \param[in,out] current_cell The current cell in the dynamic programming matrix.
-     * \param[in,out] cache        The cache storing hot helper variables.
+     * \param[in]     cache        The cache storing hot helper variables.
      */
     template <typename cell_t, typename cache_t>
     constexpr auto init_origin_cell(cell_t && current_cell, cache_t & cache) const noexcept
```
smehringer: if `cache` is an `in` paramter why is it not const?

include/seqan3/alignment/pairwise/policy/affine_gap_init_policy.hpp
```diff
@@ -64,109 +64,103 @@ class affine_gap_init_policy
...
      * \tparam        cache_t      The type of the cache.
      * \param[in,out] current_cell The current cell in the dynamic programming matrix.
-     * \param[in,out] cache        The cache storing hot helper variables.
+     * \param[in]     cache        The cache storing hot helper variables.
      */
     template <typename cell_t, typename cache_t>
     constexpr auto init_column_cell(cell_t && current_cell, cache_t & cache) const noexcept
     {
-        using std::get;
+        auto [score_cell, trace_cell] = current_cell;
```
smehringer: Don't you take a copy here instead of taking a reference? Do you want this?

include/seqan3/alignment/pairwise/policy/alignment_matrix_policy.hpp
```diff
@@ -0,0 +1,146 @@
...
+     */
+    template <std::ranges::forward_range fst_range_t, std::ranges::forward_range snd_range_t, typename score_t>
+    constexpr void allocate_matrix(fst_range_t && fst_range,
+                                   snd_range_t && snd_range,
+                                   static_band const & band,
+                                   alignment_algorithm_cache<score_t> const & cache)
+    {
+        assert(cache.gap_extension_score <= 0); // We expect it to be always non-positive.
+
+        score_t inf = std::numeric_limits<score_t>::min() - cache.gap_extension_score;
```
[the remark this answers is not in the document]
rrahn: No I want this to be -inf if possible. The problem is that integers can't handle -inf, so this will use as much as possible from the score range for the alignment.

include/seqan3/alignment/pairwise/policy/alignment_matrix_policy.hpp
```diff
@@ -0,0 +1,146 @@
...
+#include <seqan3/range/views/slice.hpp>
+
+namespace seqan3::detail
+{
+
+/*!\brief Manages the alignment and score matrix.
+ *
+ */
+template <typename alignment_algorithm_t, typename score_matrix_type, typename trace_matrix_type>
+class alignment_matrix_policy
```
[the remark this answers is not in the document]
rrahn: Then it would be called alignment_matrix_policy_base suggesting that this is a base for another policy, which it is not. We wanted to have a subfolder depth of at most 2, which means the policy folder needs to be dropped and instead all policies go into the detail directory of pairwise. Therefore, keeping policy in the naming is essential in my POV.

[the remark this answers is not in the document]
rrahn: `alignment_matrix_base_policy` is not much better I guess. It then suggests that it is a policy for the base class, which it is not.

include/seqan3/alignment/configuration/align_config_result.hpp
```diff
@@ -23,26 +23,42 @@ namespace seqan3::detail
  * \ingroup alignment_configuration
  */
 struct with_score_type
-{};
+{
+    //!\privatesection
+    //!\brief The internal rank for computing a score.
```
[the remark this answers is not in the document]
rrahn: Maybe the desctiption is a little odd. I will reformulate it. But just that you can better understand the rational. Before this change, these types were distinct, i.e. I couldn't subsume the configurations. So I had to check, do you want to compute the score? Then only compute the score. Do you want to compute the alignment? Ok, then compute the score, the coordinates and the alignment. Now I can say? Do you want to compute the score? Ok, compute it. DO you want to also compute the alignment? Good then also compute the alignment but note, that the score is already computed because computing the alignment means computing the score.

include/seqan3/alignment/pairwise/policy/alignment_matrix_policy.hpp
```diff
@@ -0,0 +1,146 @@
...
+     * \param[in] fst_range The first range (corresponds to horizontal dimension).
+     * \param[in] snd_range The second range (corresponds to vertical dimension).
+     *
+     * \details
+     *
+     * Initialises the underlying score and trace matrices and sets the respective matrix iterators to the begin of the
+     * corresponding matrix.
+     */
+    template <typename fst_range_t, typename snd_range_t>
+    constexpr void allocate_matrix(fst_range_t && fst_range, snd_range_t && snd_range)
```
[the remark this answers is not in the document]
rrahn: This has nothing to do with RAII. The class itself (the algorithm) does not know anything about the ranges it is invoked on. Thus the data storage is first allocated when the algorithm is invoked with the sequences. These of course follow the RAII principle, so that no memory will be leaked in case something goes wrong.

include/seqan3/alignment/pairwise/policy/alignment_matrix_policy.hpp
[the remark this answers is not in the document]
smehringer: > Then it would be called alignment_matrix_policy_base suggesting that this is a base for another policy

It's not? :eyes: 
Then what is this CRTP for?

include/seqan3/alignment/pairwise/policy/alignment_matrix_policy.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+    //!\brief Initialises the score and trace matrix iterator after allocating the matrices.
+    constexpr void initialise_matrix_iterator() noexcept
+    {
+        score_matrix_iter = score_matrix.begin();
+        trace_matrix_iter = trace_matrix.begin();
+    }
+
+    /*!\brief Trims the sequences to the band parameters.
+    * \tparam fst_range_t The type of the first sequence (or packed sequences).
```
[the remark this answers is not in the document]
rrahn: Well, I think I remove the concept checks from all the internal interfaces. These things should have been captured already before.

include/seqan3/alignment/pairwise/policy/alignment_matrix_policy.hpp
```diff
@@ -0,0 +1,146 @@
...
+    * \param[in] fst_range The first sequence (or packed sequences).
+    * \param[in] snd_range The first sequence (or packed sequences).
+    * \param[in] band The band.
+    *
+    * \details
+    *
+    * If the band does not intersect with the origin or the sink of the matrix the
+    * sequences are trimmed, such that the band starts in the origin and ends in the sink.
+    */
+    template <std::ranges::forward_range fst_range_t, std::ranges::forward_range snd_range_t, typename band_t>
```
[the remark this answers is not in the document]
rrahn: as above, internally I would presumably not use any concepts in the future. These things are anyway strongly coupled with the algorithm and should be handled before already.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -295,51 +282,20 @@ class alignment_algorithm :
...
-            res.score = get<3>(cache).score;
-            res.back_coordinate = this->map_banded_coordinate_to_range_position(get<3>(cache).coordinate);
-            std::tie(res.front_coordinate, res.alignment) = compute_traceback(first_range,
-                                                                              second_range,
-                                                                              get<3>(cache).coordinate);
-        }
-        return alignment_result{res};
+    //!\brief Initialises the debug matrices for the given sequences.
+    template <typename first_range_t, typename second_range_t>
+    constexpr void initialise_debug_matrices(first_range_t & first_range, second_range_t & second_range)
```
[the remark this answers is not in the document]
rrahn: `&` because I do not need to forward rvalues inside of the alignment algorithm. Not `const &` because I don't care if the sequence is `const_iterable` or not.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -387,38 +363,66 @@ class alignment_algorithm :
...
         auto const & score_scheme = get<align_cfg::scoring>(*cfg_ptr).value;
-        std::ranges::for_each(first_range, [&, this](auto seq1_value)
-        {
-            // Move internal matrix to next column.
-            this->go_next_column();
 
-            auto col = this->current_column();
-            this->init_row_cell(*std::ranges::begin(col), cache);
+        column_iter_t column_iter{};
+        ranges::for_each(first_range, [&, this] (auto const seq1_value)
```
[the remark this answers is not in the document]
rrahn: std::for_each requires common range, which is not required. I could do everything with a normal for loop, but does it make it more clear? It certainly does not reduce code. 🤷‍♂

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -436,147 +440,215 @@ class alignment_algorithm :
...
-            };
+        res.id = index;
 
-            // In banded case we need to refine the back coordinate to map to the correct position within the
-            // second range.
+        using result_config_type = std::remove_reference_t<decltype(
+                seqan3::get<align_cfg::result>(std::declval<config_t>()).value)>;
+
+        // Choose what needs to be computed.
+        if constexpr (result_config_type::rank >= 0)  // compute score
```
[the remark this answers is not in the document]
rrahn: Because the information given needs to be static anyway, since it determines the result output which allows us to define a std::function object with certain dynamic configurations in place. Since the interface should be similar to the other config interfaces these are encoded as different types. To do the things you can do with enums I added the static constexpr variable as seena bove. but using an enum for this would be overkill I guess

include/seqan3/alignment/pairwise/alignment_configurator.hpp
rrahn: Since when ??? 
This is common for most of the type traits in the STL. remove_reference, add_const, enable_if, etc.

include/seqan3/alignment/pairwise/alignment_configurator.hpp
rrahn: Yes, old stuff.

include/seqan3/alignment/pairwise/policy/affine_gap_init_policy.hpp
rrahn: These are proxies anyway. So RVO should avoid any copying. But I've learned that RVO might not work with structured bindings. That is something that I need to re-investigate. But that would be a rather silent refactorisation in the future.

rrahn: Like, before doing further micro level optomisations first the things has to be thoroughly benchmarked. But that is something after vectorisation has added so we know the end state of it all.

include/seqan3/alignment/pairwise/policy/affine_gap_init_policy.hpp
rrahn: I'll guess that can be const then

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -348,28 +304,48 @@ class alignment_algorithm :
...
-
-        std::ranges::for_each(col | ranges::view::drop_exactly(1), [&cache, this](auto && cell)
+        // Get the initial column.
+        auto zip_column = views::zip(*this->score_matrix_iter, *this->trace_matrix_iter);
+        assert(!zip_column.empty()); // Must contain at least one element.
+
+        // Initialise first cell.
+        this->init_origin_cell(zip_column.front(), cache);
+        // Take everything but last cell. Note the zip view does not preserve sized range at the moment.
+        auto column = zip_column | views::take_exactly(std::ranges::size(*this->score_matrix_iter) - 1);
```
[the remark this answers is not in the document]
rrahn: I do not enforce it but it is also nothing that is configurable from outside. So the invariants of the used classes guarantee it.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -387,38 +363,66 @@ class alignment_algorithm :
...
+        ranges::for_each(first_range, [&, this] (auto const seq1_value)
+        {
+            // Move to the next column in the alignment matrix and initialise first cell.
+            zip_column_t zip_column = views::zip(*++this->score_matrix_iter, *++this->trace_matrix_iter);
+            assert(!zip_column.empty());
 
-            auto second_range_it = std::ranges::begin(second_range);
-            std::ranges::for_each(col | ranges::view::drop_exactly(1), [&, this] (auto && cell)
+            // If the column is empty (the second range is empty)
+            if (is_snd_range_empty) // TODO [[unlikely]]
```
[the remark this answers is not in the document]
rrahn: You cannot say much about efficiency here. It would be probably very inefficient if the condition randomly changes and thus a jump is performed including flushing the instruction pipeline. But that condition is set outside of the loop and never changes inside. There is potentially a lot the compiler can do to optimise it. If we want to be sure it needs to be measured. But the performance optimisation is probably somethign I'll do in the background once we know the complete kernel including vectorisation.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -436,147 +440,215 @@ class alignment_algorithm :
...
+    /*!\brief Dumps the current alignment matrix in the debug score and if requested trace matrix.
+     * \param[in] column The current column.
+     *
+     * \details
+     *
+     * Copies the current score and if configured the curent trace column into the local debug score and trace matrix.
+     * In the banded case the full matrix will be allocated with std::optional values and only the respective parts
+     * of the matrix that are present in the band are filled.
+     */
+    template <typename zipped_column_t>
```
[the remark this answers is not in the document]
rrahn: Again, this is an internal interface not used from outside. So I left certain concept checks away on purpose.

include/seqan3/alignment/pairwise/policy/affine_gap_init_policy.hpp
rrahn: Nop, that is in fact an in,out parameter

include/seqan3/alignment/pairwise/policy/alignment_matrix_policy.hpp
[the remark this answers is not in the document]
rrahn: > It's not? 👀

No it is not! It is a policy for the alignment algorithm class. Not a base class for a policy for the alignment algorithm class. Hence, policy_base suggests something different. It is a policy and it was implemented by means of CRTP design pattern.

include/seqan3/alignment/pairwise/policy/alignment_matrix_policy.hpp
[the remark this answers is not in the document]
rrahn: I don't get it. What is wrong with lowest?

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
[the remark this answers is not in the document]
smehringer: from [fluent C++](https://www.fluentcpp.com/2018/03/30/is-stdfor_each-obsolete/)

> for_each is a tool for raising the level of abstraction of a range-based for loop. And showing the inside of the lambda within the call to for_each kills this purpose.

So if you save the lambda somewhere because you want to reuse it, ir its a aisngle line of code, I am in favour of std::ranges::for_each.
But if you just put the lambda inside the for_each body I think a range based for loop is more clear.

(Also I am not sure if introducing a lambda that captures many variables by reference may have any overhead?)

include/seqan3/alignment/pairwise/policy/alignment_matrix_policy.hpp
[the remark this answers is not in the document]
smehringer: Ok, I thought RAII is not only about avoiding the leak of memory but also to structure your data in a way that initialising it already allocates all necessary space (if possible).

include/seqan3/alignment/pairwise/policy/alignment_matrix_policy.hpp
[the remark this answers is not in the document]
smehringer: Hm... I see that we need to reduce compile time but restricting a type with a concept also helps the reader of the code to better understand what's happening. I first thought here that `band` just is a size and then it suddenly had lower and upper bounds. It is also hard to know if something is checked upstream if you have not developed the whole alignment design :( so should we really sacrifice expressiveness for compile times?

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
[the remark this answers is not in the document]
smehringer: I hate this non const_iterablity of views :roll_eyes:.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -436,147 +440,215 @@ class alignment_algorithm :
...
+            res.score_debug_matrix = std::move(score_debug_matrix);
+            if constexpr (result_config_type::rank == 3) // compute alignment
+                res.trace_debug_matrix = std::move(trace_debug_matrix);
         }
-        else
+
+        return res;
+    }
+
+    /*!\brief Dumps the current alignment matrix in the debug score and if requested trace matrix.
```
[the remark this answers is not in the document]
smehringer: In english it doesn't :see

include/seqan3/alignment/pairwise/policy/affine_gap_init_policy.hpp
smehringer: I don't see the `return` in the function? ... am I blind?

include/seqan3/alignment/pairwise/policy/alignment_matrix_policy.hpp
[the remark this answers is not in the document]
rrahn: If a user needs to wait 5 or minutes for the program to compile then this becomes a bottleneck. Since there is only one band object at the moment I changed it to the explicit class

include/seqan3/alignment/pairwise/policy/affine_gap_init_policy.hpp
rrahn: The return type should be void. The score_cell is a proxy type that internally stores reference members to the corresponding cells within the matrix. So everything is lightweight anyway. The general hope is that the compiler can see this and directly uses the reference members eliminating all unnecessary copies. But of course I keep it in mind and benchmark it thoroughly. I will annotate what is going on.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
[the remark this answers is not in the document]
rrahn: > So if you save the lambda somewhere because you want to reuse it, ir its a aisngle line of code, I am in favour of std::ranges::for_each.
But if you just put the lambda inside the for_each body I think a range based for loop is more clear.

I wouldn’t say that in general. At least the code above is expressive since it does tell you in the first line what are the variables and in the following body how it is used.. But my concern is that there are no 0-cost abstractions. So that could be a reason to make it a regular for loop. Maybe that is also expresses a clearer message, of hey, this is a hot kernel. Don’t mess with it. It is done on purpose. On the other hand there might be different kernels called depending on vectorisation or not. We’ll see 👨‍💻

include/seqan3/alignment/pairwise/policy/affine_gap_init_policy.hpp
smehringer: ah the tuple elements are references. 
If so it does not hurt to still make it `auto &` instead of `auto` for clarity. Godbolt doesn't show a difference. And I don't exactly know which copies you mean to avoid, but maybe you can explain this to me in person.

include/seqan3/alignment/pairwise/policy/alignment_matrix_policy.hpp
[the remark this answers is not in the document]
smehringer: But then we are back to the seqan2 style where everything is just `typename` in every detail function and no one, except the designer of the module, knows which functions get which type of values :(

And I would strongly hope that the compilers will at some point cach concept checks so additional checks do not add overhead.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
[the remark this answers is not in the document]
smehringer: I don't understand exactly which of your answers supports or is against my proposal.

> Maybe that is also expresses a clearer message, of hey, this is a hot kernel

which one? either way I disagree because the lambda is defined within the body of std::for_each so **it is** a range based for loop with an extra capture.

> On the other hand there might be different kernels called depending on vectorisation or not

Well then maybe the hot kernel should be outsourced into it's own function and **then** you can call std::ranges::for_each with this function.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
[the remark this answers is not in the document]
smehringer: I really don't see the design gain here.

> Because the information given needs to be static anyway

An enum can be static?

> Since the interface should be similar to the other config interfaces

But those are much more complex and sometimes require extra parameters again.
You just need flags. Enums model flags. 

Now you have 4 structs with 4 member variables and 4 constexpr variables.
Instead of 1 scoped enum with 4 values.

include/seqan3/alignment/pairwise/policy/alignment_matrix_policy.hpp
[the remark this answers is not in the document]
rrahn: > Ok, I thought RAII is not only about avoiding the leak of memory but also to structure your data in a way that initialising it already allocates all necessary space (if possible).

It does initialises everything it needs when constructing the class. The sequences are not part of it. They are first given by the function call operator. Then the memory is initialised. Everything is RAII. Everything is as it ought to be.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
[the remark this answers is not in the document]
rrahn: > which one? either way I disagree because the lambda is defined within the body of std::for_each so it is a range based for loop with an extra capture. 
Well then maybe the hot kernel should be outsourced into it's own function and then you can call std::ranges::for_each with this function.

The point is using std::for_each adds a level of abstraction. That will imply costs. Thus, the only argument against using it would be to use less abstractions as possible. Outsourcing the lambda makes only sense if it is used multiple times or if there are conditional differences calling the same for_each with different functions. None of it applies to the current PR. I am reluctant to make structural changes just because of some personal preferences. DOing some micro optimisation doesn't make a lot of sense since we do not know how the code looks like once the vectorisation is incorporated. If the final design is known, we can start optimising kernels by switching between regular for loops and for_each algorithms. Until then it is a waste of time to do so.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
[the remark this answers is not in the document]
rrahn: There is a larger design goal behind all this, that makes the use of the config arguments of algorithms for the users much easier.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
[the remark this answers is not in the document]
rrahn: Ok, I sketch the implied changes here:
```cpp
if (is_snd_range_empty)
    std::for_each(rng, kernel_to_invoke(all parameters that are necessary));
else
    std::for_each(rng, other_kernel_to_invoke(all parameters that are necessary));
```

That means adding function call objects - maybe global, but preferably as nested structs of this class. That means repeating some logic inside of these or further refactoring that into other private functions. All of this needs to be guaranteed to be inlined, which might be very well the case. But all of it might change again once vectorisation is there and thoroughly benchmarked, since keeping all variables close to the executing kernel brings great performance benefits. And there were a lot of these unexpected things happening already in SeqAn. So I can or rather will change it if there are clear indications for performance boosts, otherwise I might going to change it if it clearly improves readability by keeping the performance stable. But all of it only happens when I am looking at the benchmarks.

include/seqan3/alignment/pairwise/policy/affine_gap_init_policy.hpp
rrahn: I made it `const &`, which is semantically similar to taking it per copy.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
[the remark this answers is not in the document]
rrahn: Ok, after much debating and clearly missing all the points. The short story is I am changing this here to range based for loops. Makes total sense in this case anyway

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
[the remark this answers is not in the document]
rrahn: I know that what you are saying is completely right. In general, I completely agree with you. But I still would like to wait for final performance meassurements to come to the right conclusions and by adding hopefully the right means of abstractions to make the code still composable and efficient. Is that ok for the current state?

rrahn:
I'll check the code coverage.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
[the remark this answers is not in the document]
rrahn: I'd love to explain this to you, but it is better done on offline.

include/seqan3/alignment/pairwise/policy/affine_gap_init_policy.hpp
```diff
@@ -146,27 +153,30 @@ class affine_gap_init_policy
     template <typename cell_t, typename cache_t>
     constexpr auto init_row_cell(cell_t && current_cell, cache_t & cache) const noexcept
     {
-        using std::get;
+        // score_cell = seqan3::detail::alignment_score_matrix_proxy
+        // trace_cell = seqan3::detail::alignment_trace_matrix_proxy
+        // The proxies store pointers to the actual memory locations of the underlying matrices. Declaring the proxies
+        // const forbids us to redirect the pointers, but it is still possible to assign new values to the referenced
+        // memory locations.
```
smehringer: add notion of shallow/deep constness

include/seqan3/alignment/pairwise/policy/affine_gap_policy.hpp
```diff
@@ -58,73 +52,96 @@ class affine_gap_policy
     //!\}
 
     /*!\brief Computes the score of the current cell.
-     * \tparam  matrix_entry_type The type of the current cell.
-     * \tparam  cache_type        The type of the cache.
+     * \tparam  cell_t     The type of the current cell.
```
smehringer: Can you document what the `cell_t` looks like? I know you avoid concepts here but for clarity it would be good to document that this is a tuple (?) of references to the corresponding matrices. 

Copy and paste this for all `typename cell_t`

include/seqan3/alignment/pairwise/policy/affine_gap_policy.hpp
```diff
@@ -58,73 +52,96 @@ class affine_gap_policy
...
+        // memory locations.
+        auto const & [score_cell, trace_cell] = current_cell;
+        constexpr bool with_trace = !decays_to_ignore_v<std::remove_reference_t<decltype(trace_cell.current)>>;
         // Precompute the diagonal score.
-        score_t tmp = prev_score + score;
-
-        // Compute where the max comes from.
-        if constexpr (decays_to_ignore_v<decltype(trace_value)>) // Don't compute a traceback
+        score_t tmp = score_cell.diagonal + score;
+        // TODO: Differentiate between trace and no trace.
```
smehringer: open Todo?

include/seqan3/alignment/pairwise/policy/affine_gap_policy.hpp
```diff
@@ -58,73 +52,96 @@ class affine_gap_policy
...
 
-        if constexpr (decays_to_ignore_v<decltype(trace_value)>)
-        {
-            vt_score = (vt_score < tmp) ? tmp : vt_score;
-            hz_score = (hz_score < tmp) ? tmp : hz_score;
-        }
-        else
+    /*!\brief Computes the score of the first cell within the band.
+     * \tparam cell_t  The type of the current cell.
+     * \tparam cache_t The type of the cache.
```
smehringer: Again. This documentation does not help, because what is this cache type?

include/seqan3/alignment/pairwise/policy/alignment_matrix_policy.hpp
```diff
@@ -0,0 +1,157 @@
...
+ * \tparam trace_matrix_t The type of the used trace matrix.
+ *
+ * \details
+ *
+ * This policy is used to manage the score and trace matrix of the alignment algorithm. On invocation of an alignment
+ * instance the necessary memory is allocated and the corresponding matrix iterators are initialised. These
+ * iterators are used as a global state within this particular alignment instance and are accessed from the alignment
+ * algorithm.
+ */
+template <typename alignment_algorithm_t, typename score_matrix_t, typename trace_matrix_t>
```
smehringer: In the documentation you say that alignment_algorithm_t is a specialisation of `seqan3::alignment_algorithm`, why is this not reflected in the code? I thought we avoid concept checks in private member function but not in just any detail struct/function?
(and the template arguments of a class should be checked via static_asserts instead of concepts here then)

Same goes for score_type and trace_type, are they arithmetic?

include/seqan3/alignment/pairwise/policy/alignment_matrix_policy.hpp
```diff
@@ -0,0 +1,157 @@
...
+class alignment_matrix_policy
+{
+private:
+
```
smehringer: remove newline :nail_care:

include/seqan3/alignment/pairwise/policy/alignment_matrix_policy.hpp
```diff
@@ -0,0 +1,157 @@
...
+    constexpr void allocate_matrix(fst_range_t && fst_range, snd_range_t && snd_range)
+    {
+        score_matrix = score_matrix_t{fst_range, snd_range};
+        trace_matrix = trace_matrix_t{fst_range, snd_range};
+
+        initialise_matrix_iterator();
+    }
+
+    /*!\brief Allocates the memory of the underlying matrices.
+     * \param[in] fst_range The first range (corresponds to horizontal dimension).
```
smehringer: missing `\tparam`

include/seqan3/alignment/pairwise/policy/alignment_matrix_policy.hpp
```diff
@@ -0,0 +1,157 @@
...
+    constexpr alignment_matrix_policy() = default; //!< Defaulted.
+    constexpr alignment_matrix_policy(alignment_matrix_policy const &) = default; //!< Defaulted.
+    constexpr alignment_matrix_policy(alignment_matrix_policy &&) = default; //!< Defaulted.
+    constexpr alignment_matrix_policy & operator=(alignment_matrix_policy const &) = default; //!< Defaulted.
+    constexpr alignment_matrix_policy & operator=(alignment_matrix_policy &&) = default; //!< Defaulted.
+    ~alignment_matrix_policy() = default; //!< Defaulted.
+    //!}
+
+    /*!\brief Allocates the memory of the underlying matrices.
+     * \param[in] fst_range The first range (corresponds to horizontal dimension).
```
smehringer: missing tparam

include/seqan3/alignment/pairwise/policy/alignment_matrix_policy.hpp
rrahn: There is no concept for score_matrix_t or trace_matrix_t, yet.

rrahn: algorithm_t cannot be constraint. This is the derived class of this CRTP policy and thus is not fully defined at this point. I will make it clear in the documentation that this is the derived type.

test/unit/alignment/pairwise/policy/affine_gap_policy_test.cpp
```diff
@@ -0,0 +1,141 @@
...
+{
+public:
+    using base_t = seqan3::detail::affine_gap_policy<affine_gap_policy_mock<score_t>, score_t>;
+    using init_base_t = seqan3::detail::affine_gap_init_policy<affine_gap_policy_mock<score_t>>;
+
+    using base_t::base_t;
+    using init_base_t::init_base_t;
+
+    using base_t::compute_cell;
+    //TODO: remove me! using base_t::compute_first_band_cell;
```
marehr: TODO?

include/seqan3/alignment/pairwise/policy/affine_gap_policy.hpp
```diff
@@ -58,73 +65,102 @@ class affine_gap_policy
...
-        auto & [prev_score, vt_score, vt_trace]       = prev_cell;
-
-        //TODO For the local alignment we might be able to use GCC overflow builtin arithmetics, which
-        // allows us to check if overflow/underflow would happen. Not sure, if this helps with the performance though.
-        // (see https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html)
-
+        // score_cell = seqan3::detail::alignment_score_matrix_proxy
+        // trace_cell = seqan3::detail::alignment_trace_matrix_proxy
+        auto & [score_cell, trace_cell] = current_cell;
+        constexpr bool with_trace = !decays_to_ignore_v<std::remove_reference_t<decltype(trace_cell.current)>>;
```
marehr: Not sure, but if it "decays to ignore" shouldn't the reference be removed by decay?

include/seqan3/alignment/pairwise/policy/affine_gap_policy.hpp
```diff
@@ -58,73 +65,102 @@ class affine_gap_policy
...
+     * `cell_t` is the result type of dereferencing the zipped iterator over the respective alignment score matrix and
+     * the alignment trace matrix used inside of the seqan3::detail::alignment_matrix_policy. The first parameter
+     * stored in the zipped tuple is the seqan3::detail::alignment_score_matrix_proxy and the second value is the
+     * seqan3::detail::alignment_trace_matrix_proxy.
      */
-    template <typename matrix_entry_type, typename cache_t>
-    constexpr void compute_cell(matrix_entry_type && current_cell,
-                                cache_t & cache,
+    template <typename cell_t>
+    constexpr void compute_cell(cell_t && current_cell,
```
marehr: Why is it not `cell_t & current_cell`?

test/unit/alignment/pairwise/policy/affine_gap_init_policy_test.cpp
```diff
@@ -0,0 +1,175 @@
...
+public:
+    using score_matrix_t = std::tuple_element_t<0, test_types>;
+    using trace_matrix_t = std::tuple_element_t<1, test_types>;
+
+    using score_matrix_iter_t = typename score_matrix_t::iterator;
+    using trace_matrix_iter_t = typename trace_matrix_t::iterator;
+
+    static constexpr bool with_trace =
+        !decays_to_ignore_v<
+            std::remove_reference_t<decltype(std::declval<typename trace_matrix_t::value_type>().current)>>;
```
marehr: why the remove reference? I thought decay does that.

test/unit/alignment/pairwise/policy/affine_gap_policy_test.cpp
```diff
@@ -0,0 +1,141 @@
...
+    std::tuple<alignment_score_matrix_one_column_banded<int>,
+               alignment_trace_matrix_full_banded<trace_directions, true>,
+               std::true_type>
+    >;
+
+TYPED_TEST_CASE(affine_gap_fixture, testing_types);
+
+TYPED_TEST(affine_gap_fixture, compute_cell)
+{
+    auto zip_column = this->column();
```
marehr: zip here

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -348,28 +310,48 @@ class alignment_algorithm :
...
     void initialise_matrix(cache_t & cache)
     {
-        // Get the current dynamic programming matrix.
-        auto col = this->current_column();
-
-        this->init_origin_cell(*std::ranges::begin(col), cache);
-
-        std::ranges::for_each(col | ranges::view::drop_exactly(1), [&cache, this](auto && cell)
+        // Get the initial column.
+        auto zip_column = views::zip(*this->score_matrix_iter, *this->trace_matrix_iter);
```
marehr: zip here

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -387,38 +369,67 @@ class alignment_algorithm :
...
-            this->go_next_column();
 
-            auto col = this->current_column();
-            this->init_row_cell(*std::ranges::begin(col), cache);
+        column_iter_t column_iter{};
+
+        for (auto const seq1_value : first_range)
+        {
+            // Move to the next column in the alignment matrix and initialise first cell.
+            zip_column_t zip_column = views::zip(*++this->score_matrix_iter, *++this->trace_matrix_iter);
```
marehr: zip here

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -387,38 +369,67 @@ class alignment_algorithm :
...
+                dump_alignment_column(zip_column);
+        }
+
+        assert(this->score_matrix_iter != this->score_matrix.end());
+        assert(this->trace_matrix_iter != this->trace_matrix.end());
+
+        if (is_snd_range_empty) // TODO [[unlikely]]
+        {
+            this->check_score_last_column_or_cell(
+                views::zip(*this->score_matrix_iter, *this->trace_matrix_iter).front(), cache);
```
marehr: zip here

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -436,147 +447,214 @@ class alignment_algorithm :
...
 
-        std::ranges::for_each(first_range | views::take_exactly(this->band_column_index), [&, this](auto first_range_value)
+        column_iter_t column_iter{};
+        for (auto const first_range_value : first_range | views::take_exactly(this->score_matrix.band_col_index))
         {
-            this->go_next_column(); // Move to the next column.
-            auto col = this->current_column();
-            this->init_row_cell(*std::ranges::begin(col), cache); // initialise first row of dp matrix.
+            // Prepare next column
+            zip_column_t zip_column = views::zip(*++this->score_matrix_iter, *++this->trace_matrix_iter);
```
marehr: zip here

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -436,147 +447,214 @@ class alignment_algorithm :
...
-            {
-                this->compute_cell(std::forward<decltype(cell)>(cell),
-                                   cache,
-                                   score_scheme.score(first_range_value, *second_range_it));
-                ++second_range_it;
-            });
-
-            if (this->band_touches_last_row()) // TODO [[unlikely]]
+            // Prepare next column.
+            zip_column_t zip_column = views::zip(*++this->score_matrix_iter, *++this->trace_matrix_iter);
```
marehr: zip here

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -436,147 +447,214 @@ class alignment_algorithm :
...
+            ++begin_row;  // Move down the band and start at next row.
+
+            if constexpr (is_debug_mode)
+                dump_alignment_column(zip_column);
+        }
+
+        if (last_column_has_one_cell) // TODO [[unlikely]]
+        {
+            this->check_score_last_column_or_cell(
+                    views::zip(*this->score_matrix_iter, *this->trace_matrix_iter).front(), cache);
```
marehr: zip here

rrahn:
I've finished the prototype for the vectorised alignment and would based on that knowledge do some bigger restructurings. Both of you @marehr and @smehringer are right about certain code perils and I agree that most of the stuff can be even further simplified. I will tackle this in another PR, as this would mean another major change which should be separated from behavioral changes. Also before I commit the vectorised kernel.

rrahn:
> You are using zip(score_column, trace_column) so often that I would make it a first class entity. alignment_column_range{.score = score_matrix_column, .trace = trace_matrix_column}. The purpose of this PR was to get rid of std::get, but now you reintroduced get<0>(current_cell).current, because of this.

That is true, but now you see where the get is coming from. Before you had to know the internal layout of banded and unbanded matrix implementation to call the right get in a different context. Thus, I believe that this get is still ok, because it is in the same context.

rrahn:
@marehr So I am good to go with this? I just want to reassure that this is ok based on our agreement before.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -436,147 +447,214 @@ class alignment_algorithm :
...
+     * \details
+     *
+     * Copies the current score and if configured the current trace column into the local debug score and trace matrix.
+     * In the banded case the full matrix will be allocated with std::optional values and only the respective parts
+     * of the matrix that are present in the band are filled.
+     */
+    template <typename zipped_column_t>
+    void dump_alignment_column(zipped_column_t column)
+    {
+        using std::get;
```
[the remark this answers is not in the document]
rrahn: I don't understand

include/seqan3/alignment/pairwise/policy/alignment_matrix_policy.hpp
```diff
@@ -0,0 +1,167 @@
...
+     * \param[in] fst_range The first sequence to align.
+     * \param[in] snd_range The second sequence to align.
+     *
+     * \details
+     *
+     * Initialises the underlying score and trace matrices and sets the respective matrix iterators to the begin of the
+     * corresponding matrix.
+     */
+    template <typename fst_range_t, typename snd_range_t>
+    constexpr void allocate_matrix(fst_range_t && fst_range, snd_range_t && snd_range)
```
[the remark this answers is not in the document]
rrahn: what the ranges? If someone want's to use a filtered sequence with all Ns removed so be it. Why would I limit that. It should also not be this performance critical since the operations are done on the sequence level and not within the algorithm. Of course some things like getting the size are more expensive because they are now linear instead of constant but I wouldn't limit that in the algorithm

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
rrahn: I don't understand what you mean?

include/seqan3/alignment/pairwise/policy/affine_gap_policy.hpp
rrahn: I look into it

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
rrahn: Ah, you mean to refactor it.

PR #1293, opened 2019-10-11
File IO - formats now have protected read/write_[file]_record member functions.

smehringer:
As discussed. Although we have the PR stop, I think this is necessary because it removes the `_REMOVEME` file structures that should not remain in the library


include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -135,36 +135,81 @@ struct format_sam
 
     //!\brief The format version string.
     static constexpr char format_version[4] = "1.6";
-};
 
-} // namespace seqan3
+protected:
+    //!\privatesection
+    // Note that some member variables and functions are protected, because format_bam inherits from format_sam.
```
h-2: Since these files are extra-ordinarily long, I would only function signatures into the class and define the functions below. Just so that someone looking at the class definition can more easily assess what the API is.

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -135,36 +135,81 @@ struct format_sam
...
-//!\brief The seqan3::alignment_file_input_format_REMOVEME specialisation that handles formatted SAM input.
-//!\ingroup alignment_file
-template <>
-class alignment_file_input_format_REMOVEME<format_sam>
-{
-public:
-    //!\brief Exposes the format tag that this class is specialised with
-    using format_tag = format_sam;
+    //!\brief A variable that tracks whether the content of header has been written or not.
+    bool written_header{false};
```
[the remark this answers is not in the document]
h-2: There are a bunch of functions in here that are also not part of the API. Since for this class the protected area is part of API that should be clearly separated. I would make a shared base class for sam and bam.

include/seqan3/io/sequence_file/output.hpp
```diff
@@ -38,7 +38,7 @@
 #include <seqan3/io/sequence_file/format_fasta.hpp>
 #include <seqan3/io/sequence_file/format_fastq.hpp>
 #include <seqan3/io/sequence_file/format_genbank.hpp>
-#include <seqan3/io/sequence_file/format_sam.hpp>
+#include <seqan3/io/alignment_file/format_sam.hpp>
```
h-2: This is not pretty, but it can't be helped I guess...

include/seqan3/io/alignment_file/format_sam.hpp
h-2: (just a suggestion)

smehringer: done

include/seqan3/io/sequence_file/output.hpp
smehringer: no. unless I declare the functions in the `alignment_file/format_sam.hpp` and then define them in a `sequence_file/format_sam.hpp`. But either way, the alignment format is included and an extra header might be confusing...?

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -56,63 +56,15 @@ namespace seqan3
  *
  * \copydetails seqan3::format_sam
  */
-struct format_bam
+struct format_bam : private detail::format_sam_base
```
h-2: private ineheritance o_O ?

Also I would make this a class now that it has stuff like inheritance.

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -132,1033 +84,1120 @@ class alignment_file_input_format_REMOVEME<format_bam> : alignment_file_input_fo
...
-                string_buffer.resize(tmp32 - 1);
-                std::ranges::copy_n(std::ranges::begin(stream_view), tmp32 - 1, string_buffer.data()); // copy without \0 character
-                std::ranges::next(std::ranges::begin(stream_view)); // skip \0 character
+    using format_sam_base::read_field; // inherit read_field functions from format_base explicitly
 
-                read_field(stream_view, tmp32); // l_ref (length of reference sequence)
+    /*!\brief Reads a arithmetic field from binary stream by directly reinterpreting the bits.
+     * \tparam stream_view_type  The type of the stream as a view.
+     * \param[in, out] stream_view  The stream view to read from.
+     * \param[in, out] target       An arithmetic value to store the parsed value in.
```
h-2: Really also in?

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -132,1033 +84,1120 @@ class alignment_file_input_format_REMOVEME<format_bam> : alignment_file_input_fo
...
-                {
-                    throw format_error{"Provided reference has unequal length as specified in the header."};
-                }
-            }
+    /*!\brief Delegate parsing of std::optional types to parsing of the inner value type.
+     * \tparam stream_view_type     The type of the stream as a view.
+     * \tparam optional_value_type  The inner type of a the std::optional type of \p target.
+     *
+     * \param[in, out] stream_view  The stream view to iterate over.
+     * \param[in, out] target       The std::optional object to store the parsed value.
```
h-2: same here?

include/seqan3/io/alignment_file/format_base.hpp
```diff
@@ -0,0 +1,724 @@
...
+/*!\brief       The alignment base format.
+ * \implements  AlignmentFileFormat
+ * \ingroup     alignment_file
+ *
+ * \details
+ *
+ * Since the SAM and BAM format share a lot of functionality, this abstract base class defines common member variables
+ * and functions that are used in both formats.
+ */
+struct format_sam_base
```
h-2:
```suggestion
class format_sam_base
```
And add defaulted constructors... It's more than plain storage!

include/seqan3/io/alignment_file/format_base.hpp
```diff
@@ -0,0 +1,724 @@
...
+public:
+    //!\brief The format version string.
+    static constexpr char format_version[4] = "1.6";
+
+protected:
+    //!\brief A buffer used when parsing arithmetic values with std::from_chars.
+    std::array<char, 316> buffer{}; // Doubles can be up to 316 characters
+
+    //!\brief A variable that tracks whether the content of header has been written or not.
+    bool written_header{false};
```
h-2:
```suggestion
    bool header_was_written{false};
```
?

include/seqan3/io/alignment_file/format_base.hpp
```diff
@@ -0,0 +1,724 @@
...
+ */
+struct format_sam_base
+{
+public:
+    //!\brief The format version string.
+    static constexpr char format_version[4] = "1.6";
+
+protected:
+    //!\brief A buffer used when parsing arithmetic values with std::from_chars.
+    std::array<char, 316> buffer{}; // Doubles can be up to 316 characters
```
h-2:
```suggestion
    std::array<char, 316> arithmetic_buffer{}; // Doubles can be up to 316 characters
```

Since you have other buffers, as well?

include/seqan3/io/alignment_file/format_base.hpp
```diff
@@ -0,0 +1,724 @@
...
+
+/*!\brief       The alignment base format.
+ * \implements  AlignmentFileFormat
+ * \ingroup     alignment_file
+ *
+ * \details
+ *
+ * Since the SAM and BAM format share a lot of functionality, this abstract base class defines common member variables
+ * and functions that are used in both formats.
+ */
```
h-2: Maybe add the `\noapi` tag?

include/seqan3/io/alignment_file/format_base.hpp
```diff
@@ -0,0 +1,724 @@
...
+ *          the alignment, (2) the aligned reference length, (3) the aligned query sequence length.
+ *
+ * \details
+ *
+ * For example, the view over the cigar string "1H4M1D2M2S" will return
+ * `{[(H,1), (M,4), (D,1), (M,2), (S,2)], 7, 6}`.
+ */
+template <typename cigar_input_type>
+[[nodiscard]]
+inline std::tuple<std::vector<cigar>, int32_t, int32_t> format_sam_base::parse_cigar(cigar_input_type && cigar_input) const
```
h-2: don't use nodiscard as you take arguments by `&&`

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -125,46 +121,36 @@ namespace seqan3
  * The SAM header (if present) is read/written once in the beginning before the
  * first record is read/written.
  */
-struct format_sam
+struct format_sam : private detail::format_sam_base
```
h-2: OK, I understand the private inheritance now and it's the right thing to do :+1: 

But still make classes of the types, please!

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -184,1400 +170,959 @@ class alignment_file_input_format_REMOVEME<format_sam>
...
-        read_field(field_view, ref_id_tmp);
-        check_and_assign_ref_id(ref_id, ref_id_tmp, header, ref_seqs);
+    template <typename stream_view_type>
+    void read_field(stream_view_type && stream_view, sam_tag_dictionary & target);
 
-        read_field(field_view, ref_offset_tmp);
-        --ref_offset_tmp; // SAM format is 1-based but SeqAn operates 0-based
+    template <typename stream_it_t, typename field_type>
+    //!\cond
+        requires std::ranges::forward_range<field_type>
```
h-2: use intermediate concept syntax

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -184,1400 +170,959 @@ class alignment_file_input_format_REMOVEME<format_sam>
...
-        // The rest of the header lines
-        // -------------------------------------------------------------------------------------------------------------
-        while (is_char<'@'>(*std::ranges::begin(stream_view)))
+        if (options.sam_require_header && !std::ranges::empty(ref_id))
         {
-            std::ranges::next(std::ranges::begin(stream_view));              // skip @
+            auto id_it = header.ref_dict.end();
 
-            if (is_char<'S'>(*std::ranges::begin(stream_view)))              // SQ (sequence dictionary) tag
+            if constexpr (std::ra
```
h-2: The braces in this switch statement are redundant.

h-2: intermediate syntax

include/seqan3/io/alignment_file/input_format_concept.hpp
```diff
@@ -30,13 +30,16 @@
...
+template <typename format_type>
+struct alignment_file_input_format_exposer : format_type
+{
+public:
+    using format_type::read_alignment_record;
+};
 
 } // namespace seqan3::detail
 
+
```
h-2: remove double newline

include/seqan3/io/detail/misc.hpp
```diff
@@ -78,14 +78,11 @@ void set_format(format_variant_type & format,
...
-            using fmt_type = remove_cvref_t<decltype(fmt)>;
-            using fmt_tag  = typename fmt_type::format_tag;
-
-            for (auto const & ext : fmt_tag::file_extensions)
+            for (auto const & ext : fmt.file_extensions)
             {
                 if (std::ranges::equal(ext, extension))
                 {
-                    format = fmt_type{};
+                    format = std::move(fmt);
```
h-2: Why this change?

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -210,46 +182,23 @@ class sequence_file_input_format_REMOVEME<format_embl>
...
+                               seq_type                             && sequence,
+                               id_type                              && id,
+                               qual_type                            && SEQAN3_DOXYGEN_ONLY(qualities))
     {
         seqan3::ostreambuf_iterator stream_it{stream};
         [[maybe_unused]] size_t sequence_size = 0;
         [[maybe_unused]] char buffer[50];
         if constexpr (!detail::decays_to_ignore_v<seq_type>)
-            sequence_size = ranges::size(sequence);
+            sequence_size = std::ranges::distance(sequence);
```
h-2: Are non sized sequences allowed?

include/seqan3/io/alignment_file/format_base.hpp
```diff
@@ -0,0 +1,724 @@
+// -----------------------------------------------------------------------------------------------------
```
h-2: Call this file format_sam_base maybe?

include/seqan3/io/alignment_file/format_bam.hpp
smehringer: should the other file formats become classes to? They have a lof of member function but not custom constructors nor inheritance

include/seqan3/io/alignment_file/format_bam.hpp
smehringer: yep. 
1. I need the exact type of the arithmetic value to trigger to correct std::from_chars overload (it does e.g. range/overflow checks). 
2. I get the arithmetic value by reference in the outer `read()` function anyway so I just pass this on instead of copying the result.

include/seqan3/io/alignment_file/format_bam.hpp
smehringer: see above

include/seqan3/io/alignment_file/format_base.hpp
smehringer: I thought all or nothing and nothing is fine here?

include/seqan3/io/alignment_file/format_base.hpp
smehringer: It's in the detail scope anyway?

include/seqan3/io/alignment_file/format_sam.hpp
smehringer: you mean the short syntax within the template syntax?

include/seqan3/io/alignment_file/format_sam.hpp
smehringer: Did this comment really correspond to line 727? I don't understand

include/seqan3/io/alignment_file/format_bam.hpp
h-2: Hm, an edge case I would say. Up to you.

include/seqan3/io/alignment_file/format_base.hpp
h-2: Our rules for aggregate POD types state that they may not have private/protected, see https://github.com/seqan/seqan3/wiki/Classes-and-Structs

include/seqan3/io/alignment_file/format_base.hpp
h-2: Ah really? Ok.

include/seqan3/io/alignment_file/format_sam.hpp
h-2: Don't use "short sytax" for this. Short syntax is `std::ranges::forward_range auto var_name`.

include/seqan3/io/alignment_file/format_sam.hpp
h-2: Some problem of Github. If you look at the actual file, my comment annotates line 1019. Also the snippet above isn't actually in the file.

include/seqan3/io/alignment_file/format_bam.hpp
smehringer: ah you are right. it's just an out parameter :see

include/seqan3/io/alignment_file/format_base.hpp
smehringer: ok. but then all of the formats are now classes

include/seqan3/io/alignment_file/format_sam.hpp
smehringer: hm so we never put braces in case statements?

include/seqan3/io/detail/misc.hpp
smehringer: Since the `meta::for_each` now loops directly over the formats (and not over the `_REMOVEME` structs anymore), I can set the `format` function parameter by directly assigning `fmt`.

include/seqan3/io/sequence_file/format_embl.hpp
smehringer: yes I never require sized ranges. It might be inconsistent in the code because I did not have this in the back of my mind when writing this code mnonths ago.

include/seqan3/io/alignment_file/format_base.hpp
h-2: Yep.

include/seqan3/io/alignment_file/format_sam.hpp
h-2: We put them there if we need to, e.g. when declaring local variables.

Regular C++ rules for switch-case :wink:

include/seqan3/io/detail/misc.hpp
h-2: But there is no state that you are preserving, right? To me the change implies that there is something in `fmt` that you want to store in `format` now. Or that it is prohibitively expensive to default-construct `fm_type{}` at that point...

h-2: I wouldn't do it in that case, but I don't think we have strict rules on this or that you have to follow my advice on it.

include/seqan3/io/sequence_file/format_embl.hpp
h-2: Hm, it seems like we should definitely document clearly what the requirements are...

include/seqan3/io/alignment_file/format_bam.hpp
h-2: not fixed

include/seqan3/io/alignment_file/format_bam.hpp
h-2: see above :wink:

include/seqan3/io/alignment_file/format_sam_base.hpp
```diff
@@ -0,0 +1,735 @@
...
+ *          the alignment, (2) the aligned reference length, (3) the aligned query sequence length.
+ *
+ * \details
+ *
+ * For example, the view over the cigar string "1H4M1D2M2S" will return
+ * `{[(H,1), (M,4), (D,1), (M,2), (S,2)], 7, 6}`.
+ */
+template <typename cigar_input_type>
+[[nodiscard]]
+inline std::tuple<std::vector<cigar>, int32_t, int32_t> format_sam_base::parse_cigar(cigar_input_type && cigar_input) const
```
h-2: not resolved yet

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -56,63 +56,26 @@ namespace seqan3
  *
  * \copydetails seqan3::format_sam
  */
-struct format_bam
+class format_bam : private detail::format_sam_base
 {
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    format_bam() = default; //!\brief Defaulted.
```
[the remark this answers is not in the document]
h-2: Also add `noexcpt`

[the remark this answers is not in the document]
smehringer: I thought explicitily defaulted constructor are by default noexcept?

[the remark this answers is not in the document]
smehringer: btw. `format_sam` cannot be noexcept default constructible  because it has a string buffer which is not `noexcept` default constructible..

marehr:
```
Test project /home/travis/build/seqan/seqan3-build

    Start 1: doc_usr_test

1/2 Test #1: doc_usr_test .....................   Passed   10.04 sec

    Start 2: doc_dev_test

2/2 Test #2: doc_dev_test .....................***Failed   13.54 sec

/home/travis/build/seqan/seqan3/include/seqan3/io/alignment_file/format_sam_base.hpp:69: warning: Member format_sam_base() noexcept=default (function) of class seqan3::detail::format_sam_base is not documented.
```

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -132,1033 +95,1107 @@ class alignment_file_input_format_REMOVEME<format_bam> : alignment_file_input_fo
...
-                std::ranges::copy_n(std::ranges::begin(stream_view), tmp32 - 1, string_buffer.data()); // copy without \0 character
-                std::ranges::next(std::ranges::begin(stream_view)); // skip \0 character
+    using format_sam_base::read_field; // inherit read_field functions from format_base explicitly
 
-                read_field(stream_view, tmp32); // l_ref (length of reference sequence)
+    /*!\brief Reads a arithmetic field from binary stream by directly reinterpreting the bits.
+     * \tparam stream_view_type  The type of the stream as a view.
+     * \tparam number_type       The type of number to parse; must model std::integral.
+     * \param[in, out] stream_view  The stream view to read from.
+     * \param[in]      target       An integral value to store the parsed value in.
```
h-2: Now you have turned it from `in, out` to just `in` although it's just `out` :expressionless:

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -132,1033 +95,1107 @@ class alignment_file_input_format_REMOVEME<format_bam> : alignment_file_input_fo
...
+    void read_field(stream_view_type && stream_view, number_type & target)
+    {
+        std::ranges::copy_n(std::ranges::begin(stream_view), sizeof(target), reinterpret_cast<char *>(&target));
+    }
 
-                auto id_it = header.ref_dict.find(string_buffer);
+    /*!\brief Reads a float field from binary stream by directly reinterpreting the bits.
+     * \tparam stream_view_type  The type of the stream as a view.
+     * \param[in, out] stream_view  The stream view to read from.
+     * \param[out]     target       An float value to store the parsed value in.
```
h-2: it's correct here

include/seqan3/io/alignment_file/format_bam.hpp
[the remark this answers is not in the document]
h-2: > I thought explicitily defaulted constructor are by default noexcept?

Yes, but writing noexcept generates a diagnostic if it cannot be made noexcept. It's a good way to check.

> btw. format_sam cannot be noexcept default constructible because it has a string buffer which is not noexcept default constructible..

Does the string buffer need to be dynamic? Then that's OK! 
If you know it will always be less than e.g. 1000, I would use a seqan3::small_string instead. Although that would make objects expensive to move :thinking: 
Anyway, up to you. Maybe add a source comment why it's not noexcept.

include/seqan3/io/alignment_file/format_bam.hpp
smehringer: sorry!

include/seqan3/io/alignment_file/format_bam.hpp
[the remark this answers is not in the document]
smehringer: > Yes, but writing noexcept generates a diagnostic if it cannot be made noexcept. It's a good way to check.

True, but I think you are the only one watching out for this. I see a lot of places were noexcept was not explicitly used... I always feel like we should be consistent in our reviews, thats why I tend to put these types of "unimportant" discussion points to our list for the strategy meeting.. but it is probably not worth the discussion

> Does the string buffer need to be dynamic? Then that's OK!

yes it is for the quality string. Usually quality strings in SAM are small but you never know :woman_shrugging: 

I'll add a source comment

include/seqan3/io/alignment_file/format_sam_base.hpp
```diff
@@ -0,0 +1,747 @@
...
+    format_sam_base() noexcept = default; //!< Defaulted.
+    format_sam_base(format_sam_base const &) noexcept = default; //!< Defaulted.
+    format_sam_base & operator=(format_sam_base const &) noexcept = default; //!< Defaulted.
+    format_sam_base(format_sam_base &&) noexcept = default; //!< Defaulted.
+    format_sam_base & operator=(format_sam_base &&) noexcept = default; //!< Defaulted.
+    ~format_sam_base() noexcept = default; //!< Defaulted.
+    //!\}
+
+    //!\brief The format version string.
+    static constexpr char format_version[4] = "1.6";
```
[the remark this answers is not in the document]
h-2: I would actually use a `small_string` here, but it can be done in a follow-up if it's the only thing.

PR #1295, opened 2019-10-11
[MISC] Add nodiscard to every function that has no side effects.

rrahn:
@seqan/core 

This adds the nodiscard to all functions without side effects (at least most of them).
There are still some open questions:
Where to place the nodiscard?
1. on a separate line before the function expression
2. directly before the return type

I have used both versions so one can get a feeling for it.

Should all iterator functions that have no site effects also declared nodiscard?
e.g. operator+ and operator- etc. 
Should the range closure objects also have all operator| declared nodiscard?
Should the seqan3::get interface also be declared nodiscard? 

There are still some inconsistencies but first these three main questions need to be clarified.

We should gather the functions that we want to have declared nodiscard here and when we have a final voting I can wrap up the PR.
Note, that I added some code fixes as well. I will put them into different commits once we have an agreement. Also note that I did not compile it yet. So there will be a lot of errors. But it makes first sense to fix them when we know what to annotate with nodiscard and what not.


smehringer:
many of the attributes you added are ignored by the compiler. we should be sure how and where we can add the attribute before we decide no?

eseiler:
> many of the attributes you added are ignored by the compiler. we should be sure how and where we can add the attribute before we decide no?

>Also note that I did not compile it yet. So there will be a lot of errors. But it makes first sense to fix them when we know what to annotate with nodiscard and what not.

This is (right now) just a look-and-feel PR to discuss how to do it and address open questions.

eseiler:
This are basically the places we can put `[[nodiscard]]`
https://godbolt.org/z/1FJT0b

rrahn:
> But yes, seems like constexpr [[nodiscard]] or inline [[nodiscard]] are invalid anyways.

Seems like it! Which why I am in favor of the following syntax:

```cpp
template <types>
[[nodiscard]]
[constexpr|inline] return_type function_name(arguments)
    requires
```

I think it wasn't to much of the functions that would suffer from a trailing requires expression, especially when we decide to not do it for generic functions like get

h-2:
This highlights that in all likelihood, nodiscard also prevents optional inlining (which makes sense if you think about it). I am pretty sure that is not something we want to do in most places, right?

h-2:
Strike that, I misunderstood the comment above to.indicate that any combination of nodiscard and inline/constexpr is invalid.

marehr:
I think this can be closed for now, right?

Or did the PR followed our resolution?

> [[nodiscard]]: We add it to all functions that are guaranteed not to have side effects and return non-void:
> *   member functions that are const-qualified and have no parameters or only const-parameters
> *   free functions that only have const-qualified parameters
> *   functions with forwarding references are explicitly not part of this rule.

rrahn:
I close this for now.

Issue #1220, opened 2019-08-19
gcc 9.2.1: internal compiler error: in equal, at cp/constexpr.c:1088


mr-c:
## Platform

  - SeqAn version: 3.0.0
  - Operating system: `Linux 370b2ea81811 5.0.0-20-generic #21~18.04.1-Ubuntu SMP Thu Jun 27 04:04:37 UTC 2019 x86_64 GNU/Linux` (Debian "unstable" Docker software container running on Ubuntu)
  - Compiler: `c++ (Debian 9.2.1-1) 9.2.1 20190813`

### How to repeat the problem

Use the following `Dockerfile`
```
FROM debian:unstable

RUN apt-get update && apt-get upgrade -y && apt-get install -y --no-install-recommends \
  debhelper \
  cmake \
  zlib1g-dev \
  libboost-dev \
  libgtest-dev \
  liblemon-dev \
  help2man \
  python \
  libbz2-dev \
  build-essential \
  git \
  ca-certificates \
  devscripts

WORKDIR /build
RUN git clone --depth=1 https://salsa.debian.org/med-team/seqan3.git
WORKDIR /build/seqan3
RUN uscan --force && dpkg-buildpackage -uc -us

# or
WORKDIR /build
RUN git clone --recursive https://github.com/seqan/seqan3.git && mkdir seqan3/debug
WORKDIR /build/seqan3/debug
RUN cmake ../test/unit -DCMAKE_BUILD_TYPE=debug -DCMAKE_CXX_COMPILER=g++-9
```

### Expected behaviour

Compilation suceeds

### Actual behaviour

> [ 23%] Building CXX object alphabet/CMakeFiles/custom_alphabet_test.dir/custom_alphabet_test.cpp.o
[...]

PR #1296, opened 2019-10-11
[INFRA] Enable catalina g++-9 builds.

rrahn:
Blocked by #1362

### DIFF
--- a/.jenkinsfile
+++ b/.jenkinsfile
@@ -197,13 +197,27 @@ for (int h = 0; h < axis_agent.size(); ++h)
                 {
                     if (agent == "osx") // build on macos
                     {
-                        node("$agent && $axis_compiler_value && !catalina") // disable catalina builds for now.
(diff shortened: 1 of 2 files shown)


eseiler:
Either you downgrade to 9.1 or we re-discuss https://github.com/seqan/seqan3/issues/1220

rrahn:
> Either you downgrade to 9.1 or we re-discuss #1220

Well, non of this will work with Xcode 11 due to the include error that I reported earlier. So for now we have to stick then to one mac machine which is ok for now I guess.

rrahn:
@eseiler mit dem FIT für gcc9, this should now run through 🤞

eseiler:
the `aligned_allocator_test` fails (throws `bad_alloc` sometimes)

rrahn:
right... that was also breaking 🤦‍♂

PR #1301, opened 2019-10-21
[FIX] visibility errors on gcc - core

marehr:
Superseeds #1228

fixes partially #1317

### DIFF
--- a/include/seqan3/core/detail/pack_algorithm.hpp
+++ b/include/seqan3/core/detail/pack_algorithm.hpp
@@ -104,10 +104,10 @@ struct type_list_expander<type_list_t<args_t...>>
  * \sa https://en.cppreference.com/w/cpp/language/parameter_pack
  */
 template <typename unary_predicate_t, typename ...pack_t>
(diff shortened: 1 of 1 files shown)


include/seqan3/core/detail/pack_algorithm.hpp
```diff
@@ -104,10 +104,10 @@ struct type_list_expander<type_list_t<args_t...>>
  * \sa https://en.cppreference.com/w/cpp/language/parameter_pack
  */
 template <typename unary_predicate_t, typename ...pack_t>
-[[nodiscard]] constexpr bool all_of(unary_predicate_t && fn, pack_t && ...args)
 //!\cond
     requires (std::predicate<unary_predicate_t, pack_t> && ...)
 //!\endcond
+constexpr bool all_of [[nodiscard]] (unary_predicate_t && fn, pack_t && ...args)
```
rrahn: should we then not completely remove the nodiscard? I mean that is something really ugly and we might just add nodiscard once it can be used in a stable fashion and not before. I mean it does not break the API when we add it at a later time

PR #1302, opened 2019-10-21
[FIX] visibility errors on gcc - io

marehr:
Superseeds #1228

fixes partially #1317

### DIFF
--- a/include/seqan3/io/alignment_file/format_bam.hpp
+++ b/include/seqan3/io/alignment_file/format_bam.hpp
@@ -490,7 +490,7 @@ protected:
         return std::tuple{operations, ref_length, seq_length};
     }
 
-    // inherit read_field function from format_sam
+    //!\brief inherit read_field function from format_sam
     using sam_fmt::read_field;
 
     /*!\brief Reads a arithmetic field from binary stream by directly reinterpreting the bits.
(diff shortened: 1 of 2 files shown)


smehringer:
no clue :woman_shrugging:

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -490,7 +490,7 @@ class alignment_file_input_format_REMOVEME<format_bam> : alignment_file_input_fo
         return std::tuple{operations, ref_length, seq_length};
     }
 
-    // inherit read_field function from format_sam
+    //!\brief inherit read_field function from format_sam
```
rrahn: using declarations importing names are usually ignored  by doxygen. Did this change?

marehr: I'm not sure. I even don't know why changing something from protected to public causes the using to be interpreted differently. :confused:

smehringer: This will all be changed in PR #1293 anyway and then I'' make sure all documentation is as desired. But I would like to keep the changes here low so I don't have to many rebasing issues.

marehr:
@rrahn :ping_pong:

PR #1303, opened 2019-10-21
[FIX] visibility errors on gcc - range

marehr:
Superseeds #1228

fixes partially #1317

### DIFF
--- a/include/seqan3/range/container/dynamic_bitset.hpp
+++ b/include/seqan3/range/container/dynamic_bitset.hpp
@@ -1799,7 +1799,6 @@ public:
     }
     //!\}
 
-private:
     //!\cond DEV
     /*!\brief Serialisation support function.
      * \tparam archive_t Type of `archive`; must satisfy seqan3::cereal_archive.
--- a/include/seqan3/range/container/dynamic_bitset.hpp
+++ b/include/seqan3/range/container/dynamic_bitset.hpp
@@ -49,6 +49,10 @@ template <size_t bit_capacity = 58>
 class dynamic_bitset
 {
 private:
+    //!\brief Befriend other template instantiations of dynamic_bitset.
+    template <size_t>
+    friend class dynamic_bitset;
+
     //!\brief A bit field representing size and bit information stored in one `uint64_t`.
     struct bitfield
     {
@@ -1082,13 +1086,13 @@ public:
     //!\brief Direct access to the underlying bit field.
     constexpr bitfield * raw_data() noexcept
     {
-        return data;
+        return &data;
     }
 
     //!\copydoc raw_data()
     constexpr bitfield const * raw_data() const noexcept
     {
-        return data;
+        return &data;
     }
     //!\}
 
@@ -1580,19 +1584,13 @@ public:
      */
     //!\brief Performs element-wise comparison.
     template <size_t cap>
-    //!\cond
-        requires cap <= bit_capacity /* resolves ambiguousness when comparing two dynamic_bitsets of unequal capacity */
-    //!\endcond
     friend constexpr bool operator==(dynamic_bitset const & lhs, dynamic_bitset<cap> const & rhs) noexcept
     {
-        return lhs.data.size == rhs.data.size && lhs.data.bits == rhs.data.bits;
+        return lhs.data.size == rhs.raw_data()->size && lhs.data.bits == rhs.raw_data()->bits;
     }
 
     //!\brief Performs element-wise comparison.
     template <size_t cap>
-    //!\cond
-        requires cap <= bit_capacity
-    //!\endcond
     friend constexpr bool operator!=(dynamic_bitset const & lhs, dynamic_bitset<cap> const & rhs) noexcept
     {
         return !(lhs == rhs);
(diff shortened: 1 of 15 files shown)


include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -1580,62 +1587,44 @@ class dynamic_bitset
      */
     //!\brief Performs element-wise comparison.
     template <size_t cap>
-    //!\cond
-        requires cap <= bit_capacity /* resolves ambiguousness when comparing two dynamic_bitsets of unequal capacity */
-    //!\endcond
-    friend constexpr bool operator==(dynamic_bitset const & lhs, dynamic_bitset<cap> const & rhs) noexcept
+    constexpr bool operator==(dynamic_bitset<cap> const & rhs) const noexcept
```
smehringer: why this change? I thought we always declare comprison operators as friends?

include/seqan3/range/container/small_string.hpp
```diff
@@ -158,11 +158,12 @@ class small_string : public small_vector<char, capacity_ + 1>
      *
      * No-throw guarantee if value_type is std::is_nothrow_copy_constructible.
      */
-    template <std::forward_iterator begin_it_type, std::sentinel_for<begin_it_type> end_it_type>
-    constexpr void assign(begin_it_type begin_it, end_it_type end_it) noexcept
+    template <std::forward_iterator begin_it_type, typename end_it_type>
     //!\cond
-        requires std::constructible_from<value_type, /*ranges::iter_reference_t*/reference_t<begin_it_type>>
+        requires std::sentinel_for<end_it_type, begin_it_type> &&
+                 std::constructible_from<value_type, /*ranges::iter_reference_t*/reference_t<begin_it_type>>
```
smehringer:
```suggestion
                 std::constructible_from<value_type, std::ranges::iter_reference_t<begin_it_type>>
```
I think this should work now?

include/seqan3/range/views/pairwise_combine.hpp
```diff
@@ -130,8 +130,9 @@ class pairwise_combine_view : public std::ranges::view_interface<pairwise_combin
          * Allows construction of a const iterator (operating on a const range) from a non-const iterator.
          * This special constructor is needed as it is not covered by the standard copy and move constructors.
          */
-        template <std::convertible_to<range_type &> other_range_type>
-            requires std::same_as<std::remove_const_t<other_range_type>, std::remove_const_t<range_type>>
+        template <typename other_range_type>
+            requires std::convertible_to<other_range_type, range_type &> &&
```
smehringer: doxugen `//!\cond`

include/seqan3/range/container/small_string.hpp
marehr: thank you for the catch, I look into it.

include/seqan3/range/container/dynamic_bitset.hpp
marehr: I try to remember, but there was a problem. Enrico wrote: resolves ambiguousness when comparing two dynamic_bitsets of unequal capacity. I think this is a different issue.

I think the problem was that

```c++
friend constexpr bool operator==(dynamic_bitset const & lhs, dynamic_bitset<cap> const & rhs) noexcept;
```

means: my current class instance `dynamic_bitset` befriends `operator==` where first argument is my current class instance and the second argument is any other class instance of dynamic_bitset. (NOTE: they don't befriend this function overload)

This means that `operator==` is only allowed to access `lhs`, but not `rhs`, because `rhs` is not necessary the same type as the befriending `lhs`.

TLDR; Only lhs type befriends that class, rhs type does not.

----

I'm not to sure if we should always use `friend` declarations. 

I read https://blogs.accu.org/?p=1974 and then tried to play around with https://godbolt.org/z/ZmIqxA.

AFAIU we befriend to have the `hidden friend` property that means that a friend declared function WILL not be found by ADL of other types. And it allows to implicitly converts the left hand side in your type (if you have an implicit conversion constructor).

That means we lose in this case the implicit conversion of the left hand side. If it's important we need to figure out something else.

If we want this, we need to declare
```c++
    template <size_t cap1, size_t cap2>
    friend inline constexpr bool operator==(dynamic_bitset<cap1> const & lhs, dynamic_bitset<cap2> const & rhs) noexcept;
```
and define the friend function outside once, but then any implicit conversation from any other type to bitset would use that comparison operator. 

Making the private members protected would work.

marehr: Maybe the more killer argument:

```c++
#include <bitset>

int main()
{
    std::bitset<5> a{};
    a == 5; // works
    // 5 == a; // does not works (because operator== is a member)
    return 0;
}
```

smehringer: Phew I am not sure if I understood everything :see_no_evil: 
let's talk about this in person

marehr: yeah, I talked with @rrahn and he was in favour for the member instead of friend. If I understood @eseiler correctly, than he doesn't mind. Maybe we should put it as a discussion for next monday.

include/seqan3/range/container/small_string.hpp
marehr: I just saw that I didn't comment this out, so this is not in the scope of my PR.

include/seqan3/range/container/dynamic_bitset.hpp
marehr: We know use the public api.

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -1082,13 +1088,13 @@ class dynamic_bitset
     //!\brief Direct access to the underlying bit field.
     constexpr bitfield * raw_data() noexcept
     {
-        return data;
+        return &data;
```
smehringer: how was this not caught before? ...

include/seqan3/range/container/dynamic_bitset.hpp
```diff
@@ -1580,59 +1587,41 @@ class dynamic_bitset
      */
     //!\brief Performs element-wise comparison.
     template <size_t cap>
-    //!\cond
-        requires cap <= bit_capacity /* resolves ambiguousness when comparing two dynamic_bitsets of unequal capacity */
```
smehringer: (btw. you say the outcommenting std::ranges::iter_reference_t in line 1289 is not in the scope of this PR but here you also change something unrelated?)

marehr: Yep, because this is self-contained, but this change would be needed across multiple files.

marehr: If you want, I can change the requires back.

include/seqan3/range/container/dynamic_bitset.hpp
marehr: Because it is a constexpr (implicitly inline) function and thus will not be "instantiated" The coverage says also 100% even though it isn't, because all those constexpr functions are hidden.

smehringer: How to trick the coverage :smile:

include/seqan3/range/container/dynamic_bitset.hpp
smehringer: no it's fine. just fix travis and I'll merge

marehr:
@smehringer Travis is failing because this PR is depending on #1307. I'll ping here once the other one was merged.

marehr:
@smehringer Merge me :o)

PR #1304, opened 2019-10-21
[FIX] visibility errors on gcc - argument_parser

marehr:
Superseeds #1228

fixes partially #1317

### DIFF
--- a/include/seqan3/argument_parser/argument_parser.hpp
+++ b/include/seqan3/argument_parser/argument_parser.hpp
@@ -29,6 +29,7 @@
 #include <seqan3/argument_parser/detail/version_check.hpp>
 #include <seqan3/core/char_operations/predicate.hpp>
 #include <seqan3/core/detail/terminal.hpp>
+#include <seqan3/core/detail/test_accessor.hpp>
 #include <seqan3/core/detail/to_string.hpp>
 #include <seqan3/io/stream/concept.hpp>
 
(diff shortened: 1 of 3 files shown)


include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -29,6 +29,7 @@
 #include <seqan3/argument_parser/detail/version_check.hpp>
 #include <seqan3/core/char_operations/predicate.hpp>
 #include <seqan3/core/detail/terminal.hpp>
+#include <seqan3/core/detail/test_accessor.hpp>
```
eseiler: This does not exist yet?

marehr: I feel stupid! Forgot to add it :frowning:

marehr:
travis fails because of missing patch from #1310. Should I rebase onto new master and restart the builds?

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -536,6 +537,9 @@ class argument_parser
     //!\brief Whether the **user** specified to perform the version check (true) or not (false), default unset.
     std::optional<bool> version_check_user_decision;
 
+    //!\brief Befriend seqan3::detail::test_accessor to grant access to version_check_future.
+    friend struct ::seqan3::detail::test_accessor;
```
smehringer: :nail_care: suggestion:
I would propose to put this at the very end of the class. I know we group variables and functions, but this is something very irrelevant to the actual code and should probably not be grouped with "normal" stuff and need not to be seen immediately.

marehr: I put it near the thing that gets accessed. So that this friend accessor might be seen if we change the something on version_check_future. I propose to put the friend local to the reason why it is there. If a friend has multiple reasons I would put it at the back of the class.

marehr: Lets put it as a discussion point.

marehr:
@smehringer I think this can be merged

include/seqan3/argument_parser/argument_parser.hpp
marehr: We decided that we put the friend `near` the variable.

PR #1305, opened 2019-10-21
[FIX] visibility errors on gcc - std

marehr:
Superseeds #1228

### DIFF
--- a/include/seqan3/std/span
+++ b/include/seqan3/std/span
@@ -86,8 +86,9 @@ public:
     constexpr span(iterator_t f, index_type count) noexcept : data_{addressof(*f)}
         { (void) count; assert(span_extent == count); } // "size mismatch in span's constructor (f, count)"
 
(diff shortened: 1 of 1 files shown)


include/seqan3/std/span
```diff
@@ -86,8 +86,9 @@ public:
     constexpr span(iterator_t f, index_type count) noexcept : data_{addressof(*f)}
         { (void) count; assert(span_extent == count); } // "size mismatch in span's constructor (f, count)"
 
-    template <contiguous_iterator iterator_t, sized_sentinel_for<iterator_t> sentinel_t>
-        requires is_convertible_v<remove_reference_t<iter_reference_t<iterator_t>>(*)[], element_type(*)[]> &&
+    template <contiguous_iterator iterator_t, typename sentinel_t>
+        requires sized_sentinel_for<sentinel_t, iterator_t> &&
```
smehringer: doxugen `\\!cond` missing for all of them

marehr: can't do this, because the complete file is `//!cond`; see line17 to 425

PR #1297, opened 2019-10-14
More alphabet cleanup

h-2:
* This PR cleans up the mess in the composite alphabets that wrongly used free function operators which led to spurious compile failures because they were considered in entirely unrelated situations.
  * The composites now used friends that are constrained via SFINAE, because friends can't be constrained with concepts in GCC Concepts TS.
  * Many unused overloads were removed and also a lot of useless meta-machinery.
  * Fixes a design bug: a `alphabet_variant<a>` should only be implicitly constructible from `b`, if `a` is implicitly constructible from `b`. Previously it was possible to implicitly construct the variant, even if `a` can only be explicitly constructed from `b`. I added `explicit` to the respective constructor and fixed one affected test-case.
  * I have added tests, because of a lot of behaviour was previously untested. Only a single test was modified, because the behaviour it tested should never have worked.
  * Fixed a bunch of small things a long the way (in separate commits).

The main work is the fourth commit. I recommend to review by commit.

PR #1306, opened 2019-10-21
[FIX] visibility errors on gcc - alphabet

marehr:
Superseeds #1228

fixes partially #1317

### DIFF
--- a/include/seqan3/alphabet/composite/alphabet_tuple_base.hpp
+++ b/include/seqan3/alphabet/composite/alphabet_tuple_base.hpp
@@ -614,7 +614,7 @@ template <typename indirect_component_type, typename derived_type, typename ...c
              !detail::weakly_equality_comparable_by_members_with<indirect_component_type, derived_type>
 //!\endcond
 constexpr bool operator==(indirect_component_type const lhs,
-                          alphabet_tuple_base<derived_type, component_types...> const rhs) noexcept
+                          alphabet_tuple_base<derived_type, component_types...> const & rhs) noexcept
 {
     return rhs == lhs;
 }
@@ -626,7 +626,7 @@ template <typename indirect_component_type, typename derived_type, typename ...i
              !detail::weakly_equality_comparable_by_members_with<indirect_component_type, derived_type>
 //!\endcond
 constexpr bool operator!=(indirect_component_type const lhs,
-                          alphabet_tuple_base<derived_type, indirect_component_types...> const rhs) noexcept
+                          alphabet_tuple_base<derived_type, indirect_component_types...> const & rhs) noexcept
 {
     return rhs != lhs;
 }
@@ -638,7 +638,7 @@ template <typename indirect_component_type, typename derived_type, typename ...i
              !detail::weakly_ordered_by_members_with<indirect_component_type, derived_type>
 //!\endcond
 constexpr bool operator<(indirect_component_type const lhs,
-                         alphabet_tuple_base<derived_type, indirect_component_types...> const rhs) noexcept
+                         alphabet_tuple_base<derived_type, indirect_component_types...> const & rhs) noexcept
 {
     return rhs > lhs;
 }
@@ -650,7 +650,7 @@ template <typename indirect_component_type, typename derived_type, typename ...i
              !detail::weakly_ordered_by_members_with<indirect_component_type, derived_type>
 //!\endcond
 constexpr bool operator>(indirect_component_type const lhs,
-                         alphabet_tuple_base<derived_type, indirect_component_types...> const rhs) noexcept
+                         alphabet_tuple_base<derived_type, indirect_component_types...> const & rhs) noexcept
 {
     return rhs < lhs;
 }
@@ -662,7 +662,7 @@ template <typename indirect_component_type, typename derived_type, typename ...i
              !detail::weakly_ordered_by_members_with<indirect_component_type, derived_type>
 //!\endcond
 constexpr bool operator<=(indirect_component_type const lhs,
-                          alphabet_tuple_base<derived_type, indirect_component_types...> const rhs) noexcept
+                          alphabet_tuple_base<derived_type, indirect_component_types...> const & rhs) noexcept
 {
     return rhs >= lhs;
 }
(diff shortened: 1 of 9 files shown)


include/seqan3/alphabet/composite/alphabet_tuple_base.hpp
```diff
@@ -674,7 +674,7 @@ template <typename indirect_component_type, typename derived_type, typename ...i
              !detail::weakly_ordered_by_members_with<indirect_component_type, derived_type>
 //!\endcond
 constexpr bool operator>=(indirect_component_type const lhs,
-                          alphabet_tuple_base<derived_type, indirect_component_types...> const rhs) noexcept
+                          alphabet_tuple_base<derived_type, indirect_component_types...> const & rhs) noexcept
 {
```
h-2: These won't be necessary if #1297 gets merged...

include/seqan3/alphabet/concept.hpp
```diff
@@ -618,7 +618,7 @@ void alphabet_size(args_t ...) = delete;
 template <typename alph_t>
 struct alphabet_size_fn
 {
-private:
+public:
```
h-2: These are weird, but oh well.

include/seqan3/alphabet/detail/alphabet_proxy.hpp
```diff
@@ -263,38 +263,41 @@ class alphabet_proxy : public
      *        e.g. seqan3::alphabet_variant.
      * \{
      */
+private:
+    //!\brief work around a gcc bug that disables short-circuiting of operator&& in an enable_if_t of a friend function
+    template <typename t>
+    static constexpr bool is_alphabet_comparable_with = !std::is_same_v<derived_type, t> &&
+                                                        detail::weakly_equality_comparable_with<alphabet_type, t>;
+
```
h-2: This makes a difference? Interesting.

marehr: Yes it does. Maybe this fixes your problems where short circuiting didn't work.

h-2: I moved all the checks out into external variables anyway so I implicitly did the same thing, I think.

include/seqan3/alphabet/concept.hpp
rrahn: But do we know why this is necessary? I mean I don't see right now why the private CPO members are not accessible from within the class itself?

test/unit/alphabet/nucleotide/sam_dna16_test.cpp
```diff
@@ -25,6 +25,11 @@ INSTANTIATE_TYPED_TEST_CASE_P(sam_dna16, semi_alphabet_test, sam_dna16);
 INSTANTIATE_TYPED_TEST_CASE_P(sam_dna16, alphabet_constexpr, sam_dna16);
 INSTANTIATE_TYPED_TEST_CASE_P(sam_dna16, semi_alphabet_constexpr, sam_dna16);
 
+struct sam_dna16_accessor : public sam_dna16
+{
+    using sam_dna16::rank_to_char;
```
rrahn: so this is protected in sam_dna_16, but we have a test that tests this conversion of the sam?
Should that test not use the public interface to check the correctness of the conversion? I mean I am asking myself why it is protected and not private?

include/seqan3/alphabet/concept.hpp
marehr: clang allows this, gcc not. Not sure what the standard says, I would expect it to work, but it doesn't.

marehr: I asked on the concepts [slack](https://cpplang.slack.com/archives/C6K47F8TT/p15674338610107000), but none answered.

h-2: It works with GCC10 with c++20 concepts and concepts-ts. I would consider it a bug of GCC<=9, but one that likely won't be fixed.
Nothing to be done about it.

h-2: (You can WORKAROUND declare it for GCC<=9 but that arguably makes the code worse that just making it public, I think)

marehr: okay, good to know

rrahn: 💥

test/unit/alphabet/detail/alphabet_proxy_test.cpp
```diff
@@ -120,6 +128,10 @@ class alphabet_proxy_example2 : public alphabet_proxy<alphabet_proxy_example2, m
     constexpr alphabet_proxy_example2 & operator=(alphabet_proxy_example2 const &) = default;
     constexpr alphabet_proxy_example2 & operator=(alphabet_proxy_example2 &&) = default;
     ~alphabet_proxy_example2() = default;
+
+    constexpr alphabet_proxy_example2(alphabet_type const a) noexcept
+        : base_t{a}
```
rrahn: Why is this not on one line? I see no need to wrap this? Also when wrapping I think we agreed on putting the operator on the previous line and then continue the remaining expression on the next except for the pipe-notation for the ranges.

test/unit/alphabet/detail/alphabet_proxy_test.cpp
```diff
@@ -31,6 +32,12 @@ class alphabet_proxy_example : public alphabet_proxy<alphabet_proxy_example, dna
     constexpr alphabet_proxy_example & operator=(alphabet_proxy_example const &) = default;
     constexpr alphabet_proxy_example & operator=(alphabet_proxy_example &&) = default;
     ~alphabet_proxy_example() = default;
+
+    constexpr alphabet_proxy_example(alphabet_type const a) noexcept
+        : base_t{a}
```
rrahn: on the previous line?

test/unit/alphabet/detail/alphabet_proxy_test.cpp
marehr: can you give an example?

PR #1310, opened 2019-10-22
[FIX] ICE in constexpr alphabet tests

eseiler:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=91607


smehringer:
@seqan/core Can I merge this so the CI passes again? (Jenkins fail is unrelated)

smehringer:
Since marehr and rene (on gitter approved), I'll merge this now and we can discuss anything else later on.

PR #1311, opened 2019-10-22
[FIX] visibility errors on gcc - search

marehr:
Superseeds #1228


marehr:
travis fails because of missing patch from #1310. Should I rebase onto new master and restart the builds?

smehringer:
please rebase

PR #1318, opened 2019-10-23
[MISC] Change the naming of the  structured binding variables

rrahn:
If the surrounding context of this macro adds a variable called left or right an internal compiler error will be triggered. So we rename the variables to reduce the chance that this occurs.


rrahn:
@marehr sorry you already approved. I accidently rerequested you.

PR #1322, opened 2019-10-25
Vectorise configuration: implementation, doc and test

### LINKED ISSUE #1321: Vectorise configuration
**TASK: Add vectorise configuration**

- [x] Depends on [add config id's](https://github.com/seqan/seqan3/projects/10#card-27980956)
- [x] Create a new file `align_config_vectorise.hpp` in [include/seqan3/alignment/configuration](https://github.com/seqan/seqan3/tree/8993bb63efa8f844f23bd2207e9bac1061c94bd1/include/seqan3/alignment/configuration)
- [x] Use [align_config_max_error](https://github.com/seqan/seqan3/blob/8993bb63efa8f844f23bd2207e9bac1061c94bd1/include/seqan3/alignment/configuration/align_config_max_error.hpp#L36-L41) as a starting point
- [x] Add `align_cfg::vectorise` to [align_config_common_test.cpp](https://github.com/seqan/seqan3/blob/8993bb63efa8f844f23bd2207e9bac1061c94bd1/test/unit/alignment/configuration/align_config_common_test.cpp)
- [x] Add a new test in [test/unit/alignment/configuration](https://github.com/seqan/seqan3/tree/8993bb63efa8f844f23bd2207e9bac1061c94bd1/test/unit/alignment/configuration)
- [x] Add snippet and doc.
- [x] Adapt documentation in `enum align_config_id`

joergi-w:
Resolves #1321


Irallia:
In test/unit/alignment/configuration/align_config_common_test.cpp remove the
#include <seqan3/alignment/configuration/align_config_vectorise.hpp> line,
because it should be included with /Users/lydia/seqan3/include/seqan3/alignment/configuration/all.hpp, which needs to get updated.

include/seqan3/alignment/configuration/align_config_vectorise.hpp
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+#pragma once
+
+#include <seqan3/alignment/configuration/detail.hpp>
+#include <seqan3/core/algorithm/pipeable_config_element.hpp>
+
+namespace seqan3::align_cfg
+{
+/*!\brief Sets a flag whether the alignment computation should be performed with SIMD vectors.
```
svnbgnk: I know they are synonyms but I like "if" more than whether in programming context.

include/seqan3/alignment/configuration/detail.hpp
```diff
@@ -32,7 +32,7 @@ enum struct align_config_id : uint8_t
     parallel,     //!< ID for the seqan3::align_cfg::parallel option.
     result,       //!< ID for the \ref seqan3::align_cfg::result "result" option.
     scoring,      //!< ID for the \ref seqan3::align_cfg::scoring "scoring" option.
-    vectorise,    //!< ID for the seqan3::align_cfg::vectorise option.
+    vectorise,    //!< ID for the \ref seqan3::align_cfg::vectorise "vectorise" option.
```
svnbgnk: Thematically, Vectorise should be written directly under parallel. Additionally it makes sense to scoring/result configuration at the end. 
The configuration matrix below this snippet was changed beforehand? The matrix has to be adjusted accordingly.

joergi-w: The members had been ordered alphabetically, and we just kept this style. I don't see any advantage of grouping, as this interface is not public to users. 

Yes, the matrix has been updated in PR #1313 last week.

include/seqan3/alignment/configuration/align_config_vectorise.hpp
joergi-w: Okay, I adapt this :)

test/unit/alignment/configuration/align_config_common_test.cpp
```diff
@@ -26,7 +26,8 @@ using test_types = ::testing::Types<align_cfg::aligned_ends<std::remove_const_t<
                                     align_cfg::mode<detail::global_alignment_type>,
                                     align_cfg::mode<detail::local_alignment_type>,
                                     align_cfg::result<>,
-                                    align_cfg::scoring<nucleotide_scoring_scheme<int8_t>>>;
+                                    align_cfg::scoring<nucleotide_scoring_scheme<int8_t>>,
```
svnbgnk: The place of vectorise would also change here.

joergi-w: As stated above, this determines just the order of the tests being executed and has been established to an alphabetical order...

include/seqan3/alignment/configuration/detail.hpp
svnbgnk: Ok I did not see it ordered alphabetically.

include/seqan3/alignment/configuration/align_config_vectorise.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+/*!\brief Enables that the alignment computation is performed with SIMD vectors.
+ * \ingroup alignment_configuration
+ *
+ * \details
+ *
+ * SIMD vectors allow the computation of several pairwise alignments simultaneously in one process.
+ * Depending on your processor architecture you can gain a significant speed-up.
+ *
+ * Default: false
```
marehr: This is now unnecessary.

joergi-w: True!

joergi-w:
I squashed the commits into one for merging.

include/seqan3/alignment/configuration/align_config_vectorise.hpp
```diff
@@ -0,0 +1,53 @@
...
+#pragma once
+
+#include <seqan3/alignment/configuration/detail.hpp>
+#include <seqan3/core/algorithm/pipeable_config_element.hpp>
+#include <seqan3/core/detail/empty_type.hpp>
+
+namespace seqan3::detail
+{
+
+/*!\brief A global configuration type used to enable vectorisation of the alignment algorithm.
```
rrahn:
```suggestion
/*!\brief A tag to select the vectorised alignment algorithm.
```

include/seqan3/alignment/configuration/align_config_vectorise.hpp
```diff
@@ -0,0 +1,53 @@
...
+    //!\brief Internal id to check for consistent configuration settings.
+    static constexpr detail::align_config_id id{detail::align_config_id::vectorise};
+};
+
+} // namespace seqan3::detail
+
+namespace seqan3::align_cfg
+{
+
+/*!\brief Enables that the alignment computation is performed with SIMD vectors.
```
rrahn:
```suggestion
/*!\brief Enables the vectorised alignment computation if possible for the current configuration.
```

include/seqan3/alignment/configuration/align_config_vectorise.hpp
```diff
@@ -0,0 +1,53 @@
...
+ * \ingroup alignment_configuration
+ *
+ * \details
+ *
+ * SIMD vectors allow the computation of several pairwise alignments simultaneously in one process.
```
rrahn:
```suggestion
 * In the vectorised alignment computation several pairwise sequence alignments are processed simultaneously in one invocation. To do so, we pack the alignments in so called extended SIMD registers which allow to compute a single instruction on multiple data at the same time. Depending on your processor architecture you can gain a significant speed-up, e.g. by running up to 64 alignments in parallel on the latest intel CPUs. In our mode we vectorise multiple alignments and not a single alignment. This means that you should provide many sequences to compute as one batch rather than computing them separately as there won't be performance gains.
```

include/seqan3/alignment/configuration/align_config_vectorise.hpp
```diff
@@ -0,0 +1,57 @@
...
+namespace seqan3::align_cfg
+{
+
+/*!\brief Enables the vectorised alignment computation if possible for the current configuration.
+ * \ingroup alignment_configuration
+ *
+ * \details
+ *
+ * In the vectorised alignment computation several pairwise sequence alignments are processed simultaneously in one
+ * invocation. To do so, we pack the alignments in so called extended SIMD registers which allow to compute a single
```
rrahn:
```suggestion
 * invocation. To do so, the alignments are packed in so called extended SIMD registers which allows to compute a single
```

include/seqan3/alignment/configuration/align_config_vectorise.hpp
```diff
@@ -0,0 +1,57 @@
...
+ * instruction on multiple data at the same time. Depending on your processor architecture you can gain a significant
+ * speed-up, e.g. by running up to 64 alignments in parallel on the latest intel CPUs. In our mode we vectorise
```
rrahn:
```suggestion
 * speed-up, e.g. by running 8, 16, or 32 and more alignments in parallel. The vectorisation implementation uses an inter-sequence scheme. This means that multiple pairwise alignments are computed in parallel instead of parallelising a single pairwise alignment computation. 
```

include/seqan3/alignment/configuration/align_config_vectorise.hpp
```diff
@@ -0,0 +1,57 @@
...
+ * multiple alignments and not a single alignment. This means that you should provide many sequences to compute as
```
rrahn:
```suggestion
 * multiple alignments and not a single alignment. This means that many sequences should be provided that are computed as a batch rather than computing them ...
```

rrahn:
Thanks a lot. I did changed my sentences again a little. So that it is more in passive and not talking about us. I hope you don't mind changing them again. After that I think we are good to go! 👍

PR #1335, opened 2019-10-29
[FEATURE] Adds range aware operator to edit distance algorithm.

### LINKED ISSUE #1334: ### Range awareness - level1
**TASK: add new call operator to edit_algorithm**

- [x] Add new call operator; see [edit_distance_algorithm.hpp](https://github.com/seqan/seqan3/blob/8993bb63efa8f844f23bd2207e9bac1061c94bd1/include/seqan3/alignment/pairwise/edit_distance_algorithm.hpp#L70-L79)
- [x] that takes a range, calls the two sequences call operator on each sequence pair and writes the results into a vector, return the vector
- [x] no tests needed, add new documentation

rrahn:
Fixes #1334

### DIFF
--- a/include/seqan3/alignment/pairwise/edit_distance_algorithm.hpp
+++ b/include/seqan3/alignment/pairwise/edit_distance_algorithm.hpp
@@ -12,6 +12,9 @@
 
 #pragma once
 
+#include <tuple>
+
+#include <seqan3/alignment/pairwise/detail/concept.hpp>
 #include <seqan3/alignment/pairwise/edit_distance_unbanded.hpp>
 
 namespace seqan3::detail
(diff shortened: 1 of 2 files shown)


rrahn:
@svnbgnk I had to update the PR because I missed the point that the range type is over the indexed sequence pairs. The new code models that, but I also made this clear in you PRs already. So just that you know before you update something to have another look.

include/seqan3/alignment/pairwise/edit_distance_algorithm.hpp
```diff
@@ -60,9 +81,42 @@ class edit_distance_algorithm
...
+     *
+     * \returns A std::vector over seqan3::alignment_result.
+     *
+     * \details
+     *
+     * Computes for each contained sequence pair the respective alignment and returns all alignment results in a
+     * vector.
+     */
+    template <indexed_sequence_pair_range indexed_sequence_pairs_t>
+    constexpr auto operator()(indexed_sequence_pairs_t && indexed_sequence_pairs)
```
marehr: In this sprint I thought we only add `start_idx, range` and then do

```c++
result_vector.push_back((*this)(++start_idx, get<0>(sequence_pair), get<1>(sequence_pair)));
```

The interface change that we give the id with the sequence should be discussed with the team.

rrahn: But the current interface of the two way executor supports this interface already, since it already provides indexed sequence pairs. Hence the code is easier to adapt. So from the current state of the implementation this would make more sense wouldn't it?

marehr: Oh interesting. I wasn't aware that we already zipped the sequence pair and the index internally like 7 months ago. I thought this was a new idea when we discussed that. When it makes absolutely sense to just pass the complete range directly to the algorithm.

@smehringer As you can see [here](https://github.com/seqan/seqan3/blame/e6406b468e0d98befa1b6a142da4bfe3a5d4efd6/include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp#L223) this is quite old.

marehr: We should then update the tasks, so that we just pass that range through.

include/seqan3/alignment/pairwise/edit_distance_algorithm.hpp
```diff
@@ -13,10 +13,31 @@
...
+ *
+ * \details
+ *
+ * The following requirements are expected on a type that models indexed sequence pairs:
+ * * the type models std::ranges::forward_range,
+ * * the value type of this range models seqan3::tuple_like with exactly 2 elements,
+ * * the first type of the value type models seqan3::tuple_like with exactly 2 elements.
+ */
+template <typename t>
+SEQAN3_CONCEPT indexed_sequence_pair_range = requires (std::ranges::range_value_t<t> value)
```
marehr: I guess that we later move this concept to a different file, because it is needed at some other place to.

rrahn: Yes, that is the plan. But we might as well put it already in some designated file.

PR #1341, opened 2019-10-31
[FIXES-1339] avoid underflow for input of size 0 and 1

### LINKED ISSUE #1339: views::translate overflows buffer
I have the feeling that views:translates has an off-by-one error leading to a buffer overflow.

## Platform

  - SeqAn version: 3.0.0-440-g40d8d895
  - Operating system: Linux 5.3.7-arch1-1-ARCH
  - Compiler: g++ (GCC) 9.2.0

### How to repeat the problem

Try the [ranges/views tutorial](https://docs.seqan.de/seqan/3-master-user/tutorial_ranges.html) with tiny inputs.

```C++
#include <seqan3/range/views/all.hpp>            // include all of SeqAn's views
#include <seqan3/std/ranges>                    // include all of the standard library's views

int main(int argc, char** argv)
{
    if (argc == 1)
        return 0;
    std::string s{argv[1]};
    auto s_as_dna = s | seqan3::views::char_to<seqan3::dna5>;

    seqan3::debug_stream << "Original: " << s_as_dna << '\n';
    seqan3::debug_stream << "RevComp:  " << (s_as_dna | std::views::reverse | seqan3::views::complement) << '\n';
    seqan3::debug_stream << "Frames:   " << (s_as_dna | seqan3::views::translate) << '\n';
}
```

### Actual behaviour

The program segfaults if the input sequence is zero or one character long.

### Expected behaviour

Six empty frames.

h-2:
fixes #1339

(does this need double review?)

### DIFF
--- a/include/seqan3/range/views/translate.hpp
+++ b/include/seqan3/range/views/translate.hpp
@@ -319,12 +319,12 @@ public:
             case translation_frames::FWD_FRAME_1:
                 [[fallthrough]];
             case translation_frames::REV_FRAME_1:
-                return (seqan3::size(urange) - 1) / 3;
+                return (std::max<size_type>(seqan3::size(urange), 1) - 1) / 3;
                 break;
             case translation_frames::FWD_FRAME_2:
                 [[fallthrough]];
             case translation_frames::REV_FRAME_2:
-                return (seqan3::size(urange) - 2) / 3;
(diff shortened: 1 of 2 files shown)


kloetzl:
I can confirm this fixes the issue. :+1:

PR #1346, opened 2019-11-03
[MISC] Minor refactorisation of transfer_soft_clipping_to to make it more readable.

rrahn:
@smehringer I made you first and last reviewer 😉  . I don't think someone else needs to review this first.

### DIFF
--- a/include/seqan3/io/alignment_file/format_sam.hpp
+++ b/include/seqan3/io/alignment_file/format_sam.hpp
@@ -558,19 +558,31 @@ protected:
      */
     void transfer_soft_clipping_to(std::vector<cigar> const & cigar_vector, int32_t & sc_begin, int32_t & sc_end) const
     {
-        auto v_size = cigar_vector.size();
+        // Checks if the given index in the cigar vector is a soft clip.
+        auto soft_clipping_at = [&] (size_t const index) { return cigar_vector[index] == 'S'_cigar_op; };
+        // Checks if the given index in the cigar vector is a hard clip.
+        auto hard_clipping_at = [&] (size_t const index) { return cigar_vector[index] == 'H'_cigar_op; };
+        // Checks if the given cigar vector as at least min_size many elements.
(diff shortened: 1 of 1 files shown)


include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -558,19 +558,29 @@ class alignment_file_input_format_REMOVEME<format_sam>
...
+            sc_begin = cigar_count_at(0);
+        else if (vector_size_at_least(2) && hard_clipping_at(0) && soft_clipping_at(1))
+            sc_begin = cigar_count_at(1);
 
         // check for soft clipping at the last two positions
-        if (v_size > 1 && 'S'_cigar_op == cigar_vector[v_size - 1])
-            sc_end = get<0>(cigar_vector[v_size - 1]);
-        else if (v_size > 2 && 'H'_cigar_op == cigar_vector[v_size - 1] && 'S'_cigar_op == cigar_vector[v_size - 2])
-            sc_end = get<0>(cigar_vector[v_size - 2]);
+        auto last_index = cigar_vector.size() - 1;
```
smehringer: Hm, this underflows if the size is 0. It does not matter since you check the vector size before using it to access the vector but it might be still error prone if we ever change something in this function?

rrahn: Yes you are right and I also didn't put much thought into it since it is in general protected by the if clause. I wouldn't say that it is less safer but as an alternative we could do
```
auto last_index = vector_size_at_least(1) ? cigar_vector.size() - 1 : 0;
auto second_last_index = vector_size_at_least(2) ? last_index - 1 : 0;
```

That would still cause some errors if the if conditions are changed. As an alternative one could put these expressions back into the if-clauses but that was less readable in my opinion. 
What do you think?

smehringer: Ok then lets just leave a comment

PR #1332, opened 2019-10-29
Feature/add vectorised scoring scheme

rrahn:
Resolves #1319, 
Resolves #1333

Adds vectorised scoring scheme for simple scores differentiating between match and mismatch scores only. The vectorised version lives in detail as it is not exposed to the user.

PR #1359, opened 2019-11-05
[FEATURE] Add a simd vectorised scoring scheme which uses scoring matricies

### LINKED ISSUE #1352: ### Vectorise Scoring Schemes - level2
**TASK: Use gather in vectorised_scoring_scheme.score()**

- [ ] Depends on [write class that converts a scoring_scheme into a vectorised scoring_scheme](https://github.com/seqan/seqan3/projects/10#card-27981564)
- [ ] preprocess in the constructor, s.t. we can use a gather in score()

joshuak94:
Along with appropriate test file.

Follows similar structure to #1332.

Does not yet fully satisfy #1352. Still need to use gather in score and other optimizations.

Resolves #1352

### DIFF
--- a/include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
+++ b/include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
@@ -1,0 +1,186 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides seqan3::detail::simd_matrix_scoring_scheme.
+ * \author Joshua Kim <joshua.kim AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <seqan3/alignment/configuration/align_config_mode.hpp>
+#include <seqan3/alignment/scoring/scoring_scheme_concept.hpp>
+#include <seqan3/alphabet/concept.hpp>
+#include <seqan3/core/concept/cereal.hpp>
+#include <seqan3/core/simd/concept.hpp>
+#include <seqan3/core/simd/simd_algorithm.hpp>
+#include <seqan3/std/concepts>
+
+namespace seqan3::detail
+{
+
+/*!\brief A vectorised scoring scheme to handle scoring matrices.
+ * \ingroup scoring
+ * \tparam simd_score_t The type of the simd vector; must model seqan3::simd::simd_concept.
+ * \tparam alphabet_t The type of the alphabet over which to define the scoring scheme; must model seqan3::semialphabet.
+ * \tparam alignment_t The type of the alignment to compute; must be either seqan3::detail::global_alignment_type or
+ *                     seqan3::detail::local_alignment_type.
+ * \tparam scoring_scheme_t The type of the scoring scheme which will be used. Must model seqan3::scoring_scheme with
+ *                          the given `alphabet_t`.
+ *
+ * \details
+ *
+ * When scoring two seqan3::detail::simd vectors, this performs element-wise lookups of the compared simd vectors
+ * using the underlying scoring matrix and returns the result in another seqan3::detail::simd vector.
+ *
+ * \note Note that the alphabet type information is lost during the conversion to the simd vectors and
+ * only the ranks of the alphabet are used.
+ */
+template <simd_concept simd_score_t, semialphabet alphabet_t, typename alignment_t, typename scoring_scheme_t>
+//!\cond
+    requires scoring_scheme<scoring_scheme_t, alphabet_t> &&
+             (std::same_as<alignment_t, detail::local_alignment_type> ||
+              std::same_as<alignment_t, detail::global_alignment_type>)
+//!\endcond
+class simd_matrix_scoring_scheme
+{
+public:
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    constexpr simd_matrix_scoring_scheme() = default; //!< Defaulted.
+    constexpr simd_matrix_scoring_scheme(simd_matrix_scoring_scheme const &) = default; //!< Defaulted.
+    constexpr simd_matrix_scoring_scheme(simd_matrix_scoring_scheme &&) = default; //!< Defaulted.
+    constexpr simd_matrix_scoring_scheme & operator=(simd_matrix_scoring_scheme const &) = default; //!< Defaulted.
+    constexpr simd_matrix_scoring_scheme & operator=(simd_matrix_scoring_scheme &&) = default; //!< Defaulted.
+    ~simd_matrix_scoring_scheme() = default; //!< Defaulted.
+
+    //!\copydoc seqan3::detail::simd_matrix_scoring_scheme::initialise_from_scalar_scoring_scheme
+    constexpr explicit simd_matrix_scoring_scheme(scoring_scheme_t const & scoring_scheme)
+    {
+        initialise_from_scalar_scoring_scheme(scoring_scheme);
+    }
+
+    //!\copydoc seqan3::detail::simd_matrix_scoring_scheme::initialise_from_scalar_scoring_scheme
+    constexpr simd_matrix_scoring_scheme & operator=(scoring_scheme_t const & scoring_scheme)
+    {
+        initialise_from_scalar_scoring_scheme(scoring_scheme);
+        return *this;
+    }
+    //!\}
+
+    /*!\name Score computation
+     *\{
+     */
+    /*!\brief Given two simd vectors, compute an element-wise score and return another simd vector.
+     * \param[in] lhs The left operand to compare.
+     * \param[in] rhs The right operand to compare.
+     * \returns A simd vector of the same type as the input, with scores computed between both inputs.
+     *
+     * ### Exception
+     *
+     * No-throw guarantee.
+     *
+     * ### Complexity
+     *
+     * Linear in the length of one input vector (`lhs` and `rhs` are equally sized).
+     *
+     * ### Thread safety
+     *
+     * Thread-safe.
+     */
+    constexpr simd_score_t score(simd_score_t const & lhs, simd_score_t const & rhs) const noexcept
+    {
+        simd_score_t result{};
+
+        for (size_t i = 0; i < simd_traits<simd_score_t>::length; ++i)
+        {
+            if (is_padded(lhs[i]) || is_padded(rhs[i]))
+            {
+                if constexpr (std::same_as<alignment_t, detail::global_alignment_type>)
+                    result[i] = 1;
+                else
+                    result[i] = -1;
+            }
+            else
+            {
+                result[i] = internal_scoring_scheme.score(assign_rank_to(lhs[i], alphabet_t{}),
+                                                          assign_rank_to(rhs[i], alphabet_t{}));
+            }
+        }
+
+        return result;
+    }
+    //!\}
+
+private:
+    //!\brief Internally stores the given scalar scoring scheme matrix.
+    scoring_scheme_t internal_scoring_scheme;
+
+    /*!\brief Store the given scoring scheme matrix into a private member variable.
+     * \param[in] scoring_scheme The scoring scheme to initialise the vectorised match and mismatch score with.
+     */
+    constexpr void initialise_from_scalar_scoring_scheme(scoring_scheme_t const & scoring_scheme)
+    {
+        using score_t = decltype(std::declval<scoring_scheme_t const &>().score(alphabet_t{}, alphabet_t{}));
+        using simd_scalar_t = typename simd_traits<simd_score_t>::scalar_type;
+
+        // Check if the scoring scheme match and mismatch scores do not overflow with the respective scalar type.
+        if constexpr (sizeof(simd_scalar_t) < sizeof(score_t))
+        {
+            if (min_or_max_exceeded<score_t, simd_scalar_t>(scoring_scheme))
+                throw std::invalid_argument{"The selected scoring scheme score overflows "
+                                            "for the selected scalar type of the simd type."};
+        }
+
+        internal_scoring_scheme = scoring_scheme;
+    }
+
+    /*!\brief Check if any score in the scoring scheme matrix exceeds the min or max value allowed by the simd vector.
+     * \param[in] scoring_scheme The scoring scheme to check for min or max violations.
+     * \tparam score_t The type of the scores in the scoring scheme matrix.
+     * \tparam simd_scalar_t The type of the simd vector.
+     * \returns Returns `true` if it encounters a value larger than the maximum or lower than the minimum allowed by
+     *          the simd vector, and `false` if otherwise.
+     */
+    template <typename score_t, typename simd_scalar_t>
+    constexpr bool min_or_max_exceeded(scoring_scheme_t const & scoring_scheme)
+    {
+        score_t max_val = static_cast<score_t>(std::numeric_limits<simd_scalar_t>::max());
+        score_t min_val = static_cast<score_t>(std::numeric_limits<simd_scalar_t>::lowest());
+
+        for (uint8_t i = 0; i < alphabet_size<alphabet_t>; ++i)
+        {
+            for (uint8_t j = 0; j < alphabet_size<alphabet_t>; ++j)
+            {
+                score_t tmp_score = scoring_scheme.score(assign_rank_to(i, alphabet_t{}),
+                                                         assign_rank_to(j, alphabet_t{}));
+
+                if (tmp_score > max_val || tmp_score < min_val)
+                    return true;
+            }
+        }
+
+        return false;
+    }
+
+    /*!\brief Checks whether the given value is a padded value or not.
+     * \param[in] value The value to check if it is padded.
+     * \returns Returns `true` if the value is padded and `false` otherwise.
+     *
+     * \details This check is done by casting the scalar type of the value into an unsigned value, and then checking
+     *          to see if this is larger than the alphabet size, which would imply that it is a padded value.
+     */
+    constexpr bool is_padded(typename simd_traits<simd_score_t>::scalar_type const & value) const
+    {
+        using unsigned_scalar_t = typename std::make_unsigned<typename simd_traits<simd_score_t>::scalar_type>::type;
+
+        return static_cast<unsigned_scalar_t>(value) >= alphabet_size<alphabet_t>;
+    }
+};
+
+} // namespace seqan3::detail
--- a/test/unit/alignment/scoring/simd_matrix_scoring_scheme_test.cpp
+++ b/test/unit/alignment/scoring/simd_matrix_scoring_scheme_test.cpp
@@ -1,0 +1,223 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE
+// -----------------------------------------------------------------------------------------------------
+
+#include <gtest/gtest.h>
+
+#include <seqan3/alignment/scoring/aminoacid_scoring_scheme.hpp>
+#include <seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp>
+#include <seqan3/alphabet/aminoacid/aa27.hpp>
+#include <seqan3/test/pretty_printing.hpp>
+#include <seqan3/test/simd_utility.hpp>
+
+using namespace seqan3;
+using namespace seqan3::detail;
+
+template <typename simd_t>
+struct simd_matrix_scoring_scheme_test : public ::testing::Test
+{
+    using scalar_t = typename simd_traits<simd_t>::scalar_type;
+
+    scalar_t padded_value1 = std::numeric_limits<scalar_t>::lowest();      // sets the most significant bit.
+    scalar_t padded_value2 = std::numeric_limits<scalar_t>::lowest() >> 1; // sets the bit before most significant bit.
+};
+
+using simd_test_types = ::testing::Types<//simd::simd_type_t<int8_t>,
+                                         //simd::simd_type_t<int16_t>,
+                                         simd::simd_type_t<int32_t>>;
+
+TYPED_TEST_CASE(simd_matrix_scoring_scheme_test, simd_test_types);
+
+TYPED_TEST(simd_matrix_scoring_scheme_test, basic_construction)
+{
+    using scheme_t = simd_matrix_scoring_scheme<TypeParam,
+                                                aa27,
+                                                detail::global_alignment_type,
+                                                aminoacid_scoring_scheme<>>;
+
+    EXPECT_TRUE(std::is_nothrow_default_constructible_v<scheme_t>);
+    EXPECT_TRUE(std::is_nothrow_copy_constructible_v<scheme_t>);
+    EXPECT_TRUE(std::is_nothrow_move_constructible_v<scheme_t>);
+    EXPECT_TRUE(std::is_nothrow_copy_assignable_v<scheme_t>);
+    EXPECT_TRUE(std::is_nothrow_move_assignable_v<scheme_t>);
+    EXPECT_TRUE(std::is_nothrow_destructible_v<scheme_t>);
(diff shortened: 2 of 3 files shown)


include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,150 @@
...
+#include <seqan3/alphabet/concept.hpp>
+#include <seqan3/core/concept/cereal.hpp>
+#include <seqan3/core/simd/concept.hpp>
+#include <seqan3/core/simd/simd_algorithm.hpp>
+#include <seqan3/std/concepts>
+
+namespace seqan3::detail
+{
+
+/*!\brief A vectorised scoring scheme to handle scoring matricies.
```
eseiler:
```suggestion
/*!\brief A vectorised scoring scheme to handle scoring matrices.
```

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,150 @@
...
+ *                     seqan3::detail::local_alignment_type.
+ * \tparam scoring_scheme_t The type of the scoring scheme which will be used. Must model seqan3::scoring_scheme with
+ *                          the given alphabet_t.
+ *
+ * \details
+ *
+ * Wraps a regular scoring scheme matrix by storing the matrix in a private member variable. Then, when scoring two
+ * seqan3::detail::simd vectors, it performs element-wise lookups into the scoring matrix and returns the result in
+ * another seqan3::detail::simd vector.
+ * Note during the conversion to the simd vectors the alphabet type information is lost and
```
eseiler: Either

```suggestion
 * Note that, during the conversion to the simd vectors, the alphabet type information is lost and
```

eseiler: Or 

```suggestion
 * Note that the alphabet type information is lost during the conversion to the simd vectors and
```

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,150 @@
...
+             (std::same_as<alignment_t, detail::local_alignment_type> ||
+              std::same_as<alignment_t, detail::global_alignment_type>)
+//!\endcond
+class simd_matrix_scoring_scheme
+{
+public:
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    //!\brief Default.
```
eseiler:
```suggestion
    //!\brief Defaulted.
```

x6


You can probably also use the inline notation
```
//!< Defaulted.
```

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,150 @@
...
+    }
+    //!\}
+
+    /*!\name Score computation
+     *\{
+     */
+    /*!\brief Given two simd vectors, compute an element-wise score and return in another simd vector.
+     * \param[in] lhs The left operand to compare.
+     * \param[in] rhs The right operand to compare.
+     * \returns Returns a simd vector of same type as input, with scores computed between both inputs.
```
eseiler:
```suggestion
     * \returns A simd vector of same type as input, with scores computed between both inputs.
```

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,150 @@
...
+     *
+     * ### Exception
+     *
+     * No-throw guarantee.
+     *
+     * ### Complexity
+     *
+     * The complexity is linear over the length of the input vectors, which are the same length.
```
eseiler:
```suggestion
     * Linear in the length of one input vector. (`lhs` and `rhs` are equally sized)
```

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+        for (int i = 0; i < simd_traits<simd_score_t>::length; i++)
+        {
+            if (is_padded(lhs[i]) || is_padded(rhs[i]))
+            {
+                result[i] = std::same_as<alignment_t, detail::global_alignment_type> ? 1 : -1;
+            }
+            else
+            {
+                result[i] = internal_scoring_scheme.score(assign_rank_to(lhs[i], alphabet_t{}), assign_rank_to(rhs[i], alphabet_t{}));
```
eseiler: linelength

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,150 @@
...
+            }
+        }
+
+        return result;
+    }
+    //!\}
```
eseiler: new line

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,150 @@
...
+private:
+    /*!\brief Internally stores the given scalar scoring scheme matrix.
```
eseiler: Fits on one line

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,150 @@
...
+     */
+    scoring_scheme_t internal_scoring_scheme;
+
+    /*!\brief Store the given scoring scheme matrix into a private member variable.
+     * \param[in] scoring_scheme The scoring scheme to initialise the vectorised match and mismatch score from.
```
eseiler:
```suggestion
     * \param[in] scoring_scheme The scoring scheme to initialise the vectorised match and mismatch score with.
```

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,150 @@
...
+     */
+    constexpr void initialise_from_scalar_scoring_scheme(scoring_scheme_t const & scoring_scheme)
+    {
+        internal_scoring_scheme = scoring_scheme;
+    }
+
+    /*!\brief Check whether the given value is a padded value or not. This check is done by casting the scalar
```
eseiler:
```suggestion
    /*!\brief Checks whether the given value is a padded value or not. This check is done by casting the scalar
```

test/unit/alignment/scoring/simd_matrix_scoring_scheme_test.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+using simd_test_types = ::testing::Types<//simd::simd_type_t<int8_t>,
+                                         //simd::simd_type_t<int16_t>,
+                                         simd::simd_type_t<int32_t>>;
+
+TYPED_TEST_CASE(simd_matrix_scoring_scheme_test, simd_test_types);
+
+TYPED_TEST(simd_matrix_scoring_scheme_test, basic_construction)
+{
+    using scheme_t = simd_matrix_scoring_scheme<TypeParam, aa27, detail::global_alignment_type, aminoacid_scoring_scheme<>>;
```
eseiler: linelength 
x6

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,151 @@
...
+              std::same_as<alignment_t, detail::global_alignment_type>)
+//!\endcond
+class simd_matrix_scoring_scheme
+{
+public:
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    //!< Defaulted.
+    constexpr simd_matrix_scoring_scheme() = default;
```
eseiler: The syntax is
```suggestion
    constexpr simd_matrix_scoring_scheme() = default; //!< Defaulted.
```

joshuak94: Makes sense, I was confused why it was failing...

joshuak94:
@eseiler Yeah I spoke with @rrahn and he thought it would be better to first create the class in a "trivial" way, as a proof of concept/framework, and then after it has tests and everything I could implement the rest. That's why I didn't put "resolves" in my PR description, because there is still some stuff missing.

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,145 @@
...
+    {
+        initialise_from_scalar_scoring_scheme(scoring_scheme);
+        return *this;
+    }
+    //!\}
+
+    /*!\name Score computation
+     *\{
+     */
+    /*!\brief Given two simd vectors, compute an element-wise score and return in another simd vector.
```
smehringer:
```suggestion
    /*!\brief Given two simd vectors, compute an element-wise score and return another simd vector.
```
?

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,145 @@
...
+     * \param[in] lhs The left operand to compare.
+     * \param[in] rhs The right operand to compare.
+     * \returns A simd vector of same type as input, with scores computed between both inputs.
```
smehringer:
```suggestion
     * \returns A simd vector of the same type as input, with scores computed between both inputs.
```

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,145 @@
...
+    //!\}
+
+private:
+    //!\brief Internally stores the given scalar scoring scheme matrix.
+    scoring_scheme_t internal_scoring_scheme;
+
+    /*!\brief Store the given scoring scheme matrix into a private member variable.
+     * \param[in] scoring_scheme The scoring scheme to initialise the vectorised match and mismatch score with.
+     */
+    constexpr void initialise_from_scalar_scoring_scheme(scoring_scheme_t const & scoring_scheme)
```
smehringer: why do we have this private member function if this is only doing an assignment? I don't see the benefit of abstraction here.

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,145 @@
...
+    {
+        internal_scoring_scheme = scoring_scheme;
+    }
+
+    /*!\brief Checks whether the given value is a padded value or not. This check is done by casting the scalar
```
smehringer: Everything after the first sentence is not brief anymore but should go into \details of this documentation

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,145 @@
...
+     *        type of the value into an unsigned value, and then checking to see if this is larger than the alphabet
+     *        size, which would imply that it is a padded value.
+     * \param[in] value The value to check if it is padded.
+     * \returns Returns true if the value is padded and false otherwise.
```
smehringer:
```suggestion
     * \returns Returns `true` if the value is padded and `false` otherwise.
```

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,145 @@
...
+     */
+    constexpr bool is_padded(typename simd_traits<simd_score_t>::scalar_type const & value) const
+    {
+        typedef typename std::make_unsigned<typename simd_traits<simd_score_t>::scalar_type>::type unsigned_type;
```
smehringer: we use `using`.
```suggestion
        using unsigned_type = typename std::make_unsigned<typename simd_traits<simd_score_t>::scalar_type>::type;
```

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
joshuak94: I copied it over from @rrahn 's pull request, and kept it in because I thought we might want the format to be consistent... Does it not matter?

joshuak94: @rrahn Should I mirror the format that you had in your pull request, with `initialize_from_scalar_scoring_scheme`? Or should I just put the `internal_scoring_scheme = scoring_scheme` assignment in the constructor/assignment operator?

rrahn: No, if the scoring scheme is the only member that is initialised you can actually use the direct member initialiser in the constructor. For the assignment operator you assign it of course in the body.

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
rrahn: I would suggest `unsigned_scalar_t`

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,146 @@
...
+     *
+     * Linear in the length of one input vector. (`lhs` and `rhs` are equally sized)
+     *
+     * ### Thread safety
+     *
+     * Thread-safe.
+     */
+    constexpr simd_score_t score(simd_score_t const & lhs, simd_score_t const & rhs) const noexcept
+    {
+        simd_score_t result;
```
rrahn:
```suggestion
        simd_score_t result{};
```

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+        for (int i = 0; i < simd_traits<simd_score_t>::length; i++)
```
rrahn:
```suggestion
        for (int i = 0; i < simd_traits<simd_score_t>::length; ++i)
```

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
rrahn: But hold on. You actually need to test if the score stored in the scoring scheme does not exceed the largest value that is representable with the scalar type of the simd vector. So whenever the score type of the underlying scoring scheme is smaller or equal in the byte size this is not a problem. But if you have a byte packing in the simd vector and you use int16_t as a score type for the scoring scheme an overflow is possible. Thus, this should be tested during the construction. This would then again validate to move this function out.

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,146 @@
...
+        {
+            if (is_padded(lhs[i]) || is_padded(rhs[i]))
+            {
+                result[i] = std::same_as<alignment_t, detail::global_alignment_type> ? 1 : -1;
```
rrahn:
```suggestion
                if constexpr (std::same_as<alignment_t, detail::global_alignment_type>)
                    result[i] = 1;
                 else 
                    result[i] = -1;
```

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
rrahn:
```suggestion
        for (size_t i = 0; i < simd_traits<simd_score_t>::length; i++)
```

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
joshuak94: How would that look? Maybe something like this...?
```cpp
        static_assert(scoring_scheme_t::score_type <= simd_traits<simd_score_t>::scalar_type);

```cpp

But I don't know how to do this correctly... Do I get the max possible value of `score_type` and `scalar_type` and compare them?

Comment (joshuak94) [reply include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp]: Ah wait I see you did it in yours.

**Comment by joshuak94** [comment]
@rrahn Alright, I tried implementing the check for this. I couldn't think of a more clever way than just looking through every value in the matrix and comparing it.

Are the matrices guaranteed to be symmetric along the main diagonal? Maybe then I could make it a bit faster.

File: include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp [hunk 22]
```diff
@@ -0,0 +1,190 @@
...
+
+#include <seqan3/alignment/configuration/align_config_mode.hpp>
+#include <seqan3/alignment/scoring/scoring_scheme_concept.hpp>
+#include <seqan3/alphabet/concept.hpp>
+#include <seqan3/core/concept/cereal.hpp>
+#include <seqan3/core/simd/concept.hpp>
+#include <seqan3/core/simd/simd_algorithm.hpp>
+#include <seqan3/std/concepts>
+
+#include <seqan3/core/debug_stream.hpp>
```
rrahn: That should be removed?

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,190 @@
...
+
+namespace seqan3::detail
+{
+
+/*!\brief A vectorised scoring scheme to handle scoring matrices.
+ * \ingroup scoring
+ * \tparam simd_score_t The type of the simd vector; must model seqan3::detail::simd_concept.
```
rrahn:
```suggestion
 * \tparam simd_score_t The type of the simd vector; must model seqan3::simd::simd_concept.
```

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,190 @@
...
+ * \tparam alphabet_t The type of the alphabet over which to define the scoring scheme; must model seqan3::semialphabet
+ *                    and must have an alphabet size of at least 2.
```
rrahn: Not that it would be a relevant use case but in this general case you can also have single character alphabet. So I would remove the restriction here.

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,190 @@
...
+ * \tparam alignment_t The type of the alignment to compute; must be either seqan3::detail::global_alignment_type or
+ *                     seqan3::detail::local_alignment_type.
+ * \tparam scoring_scheme_t The type of the scoring scheme which will be used. Must model seqan3::scoring_scheme with
+ *                          the given alphabet_t.
```
rrahn: please always annotate parameter names or built in types such as `bool` with inline code

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,190 @@
...
+ *
+ * \details
+ *
+ * Wraps a regular scoring scheme matrix by storing the matrix in a private member variable. Then, when scoring two
```
rrahn: I like the detail of this. I would advice to not to much talk about implementation details such as storing it as a private member. If this changes for any particular reason, than this easily overseen and thus does not correspond anymore to the documentation. Maybe something like, `uses the underlying scoring scheme to perform an element-wise lookup of the compared simd vectors.`

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,190 @@
...
+ * seqan3::detail::simd vectors, it performs element-wise lookups into the scoring matrix and returns the result in
+ * another seqan3::detail::simd vector.
+ * Note that the alphabet type information is lost during the conversion to the simd vectors and
```
rrahn: \note

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,190 @@
...
+ * only the ranks of the alphabet are used.
+ */
+template <simd_concept simd_score_t, semialphabet alphabet_t, typename alignment_t, typename scoring_scheme_t>
+//!\cond
+    requires seqan3::alphabet_size<alphabet_t> > 1 &&
```
rrahn: can be removed

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,190 @@
...
+     * \param[in] rhs The right operand to compare.
+     * \returns A simd vector of the same type as the input, with scores computed between both inputs.
+     *
+     * ### Exception
+     *
+     * No-throw guarantee.
+     *
+     * ### Complexity
+     *
+     * Linear in the length of one input vector. (`lhs` and `rhs` are equally sized)
```
rrahn:
```suggestion
     * Linear in the length of one input vector (`lhs` and `rhs` are equally sized).
```

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        internal_scoring_scheme = scoring_scheme;
+    }
+
+    /*!\brief Check if any score in the scoring scheme matrix exceeds the min or max value allowed by the simd vector.
+     * \param[in] scoring_scheme The scoring scheme to check for min or max violations.
+     * \tparam score_t The type of the scores in the scoring scheme matrix.
+     * \tparam simd_scalar_t The type of the simd vector.
+     * \returns Returns true if it encounters a value larger than the maximum or lower than the minimum allowed by
```
rrahn:
```suggestion
     * \returns Returns `true` if it encounters a value larger than the maximum or lower than the minimum allowed by
```

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,190 @@
...
+     *          the simd vector, and false if otherwise.
```
rrahn:
```suggestion
     *          the simd vector, and `false` if otherwise.
```

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
rrahn: not resolved?

include/seqan3/alignment/scoring/detail/simd_matrix_scoring_scheme.hpp
```diff
@@ -0,0 +1,190 @@
...
+    /*!\brief Checks whether the given value is a padded value or not.
+     * \param[in] value The value to check if it is padded.
+     * \returns Returns true if the value is padded and false otherwise.
+     *
+     * \details This check is done by casting the scalar type of the value into an unsigned value, and then checking
+     *          to see if this is larger than the alphabet size, which would imply that it is a padded value.
+     */
+    constexpr bool is_padded(typename simd_traits<simd_score_t>::scalar_type const & value) const
+    {
+        typedef typename std::make_unsigned<typename simd_traits<simd_score_t>::scalar_type>::type unsigned_scalar_t;
```
rrahn: please use the `using
```suggestion
        using unsigned_scalar_t = typename std::make_unsigned<typename simd_traits<simd_score_t>::scalar_type>::type;
```

joshuak94: Can I ask what the difference is between typedef and using? Is it just for consistency's sake?

rrahn: `typedef` was used before C++11 and `using` is now the more appropriate way. It is in fact an assignment of a type and thus it is easier to read. Other from that there is not much of a difference.

rrahn:
@joshuak94 Something broke one mac. The tests are not going through. See: http://cdash.seqan.de/testDetails.php?test=62383151&build=347660

joshuak94:
@rrahn All tests pass now! Had an uninitialized matrix which apparently was OK for me but not for MacOS...

rrahn:
@joshuak94 can you pleas update the commit messages to follow our guidelines. Like using a tag to indicate the change that is applied. I can squash commit the entire PR if you want to but just in case you want to keep things separately.

joshuak94:
@rrahn Done!

PR #1354, opened 2019-11-05
Feature/add chunk param to alignment executor

rrahn:
Depends on #1350 

Fixes #1353

PR #1360, opened 2019-11-05
Feature/chunk resource range in alignment executor

### LINKED ISSUE #1355: ### Range awareness - level3
**TASK: Chunk range of sequence pair according to the given parameter in alignment_executor_two_way**

- [ ] Depends on [add new parameter to alignment_executor_two_way constructor for chunking](https://github.com/seqan/seqan3/projects/10#card-27977538) (#1354)
- [ ] Chunk the range accordingly.
- [ ] [OPTION1] Call execution handler on each element in the chunk. See [alignment_executor_two_way](https://github.com/seqan/seqan3/blob/8993bb63efa8f844f23bd2207e9bac1061c94bd1/include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp#L220-L231)
- [ ] [OPTION2] This task can be merged with the task [Chunk range of sequence pair according to the given parameter in alignment_executor_two_way](https://github.com/seqan/seqan3/projects/10#card-27979762) if the dependencies in that task are met

rrahn:
Depends on #1354

Fixes #1355

Only last two commits are relevant

### DIFF
--- a/include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
+++ b/include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
@@ -22,6 +22,7 @@
 #include <seqan3/core/parallel/execution.hpp>
 #include <seqan3/core/type_traits/range.hpp>
 #include <seqan3/range/shortcuts.hpp>
+#include <seqan3/range/views/chunk.hpp>
 #include <seqan3/range/views/view_all.hpp>
 #include <seqan3/range/views/zip.hpp>
 #include <seqan3/std/ranges>
@@ -56,7 +57,7 @@ private:
      * \{
      */
     //!\brief The underlying resource type.
-    using resource_type = decltype(views::zip(std::declval<resource_t>(), std::views::iota(0)));
+    using resource_type = decltype(views::zip(std::declval<resource_t>(), std::views::iota(0)) | views::chunk(1));
     //!\brief The iterator over the underlying resource.
     using resource_iterator_t = std::ranges::iterator_t<resource_type>;
     //!\brief The value type of the resource.
@@ -157,8 +158,6 @@ public:
                                alignment_algorithm_t fn,
                                size_t chunk_size = 1u,
                                exec_policy_t const & SEQAN3_DOXYGEN_ONLY(exec) = seq) :
-        resource{views::zip(std::forward<resource_t>(resrc), std::views::iota(0))},
-        resource_it{resource.begin()},
         kernel{std::move(fn)},
         _chunk_size{chunk_size}
     {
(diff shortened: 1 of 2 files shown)


include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
```diff
@@ -252,19 +254,27 @@ class alignment_executor_two_way
...
-            auto && [first_seq, second_seq] = tpl;
-            buffer_pointer write_to = gptr;
-            exec_handler.execute(kernel,
-                                 idx,
-                                 first_seq | views::all,
-                                 second_seq | views::all,
-                                 [write_to] (auto && res) { *write_to = std::move(res); });
+            for (auto && [sequence_pair, idx] : *resource_it)
+            {
+                buffer_pointer write_to = gptr;
```
eseiler: Why not 
```cpp
auto && [first_seq, second_seq] = sequence_pair;
```

Then you do not need the `std::get` and it gets more readable (at least for me).

eseiler: Why do you copy the pointer?
As far as I can see, the lambda copies it anyway?

rrahn: There is no special reason. The point is that the bind variables are not really needed more than once, so not unpacking them has at least a  message: "Don't spoil the scope with something you don't need." 
But once all of the PRs dealing with the adaption of ranges have been resolved the entire inner loop will vanish. So I guess it is not urgent to change that. Would that be ok for you?

rrahn: Sure thing. It was just from the old kernel. I can change that but will be modified soon by another PR anyway (see above). Would it be ok for you to leave this then as is?

eseiler: ok

eseiler: ok

include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
```diff
@@ -169,10 +168,13 @@ class alignment_executor_two_way
         if (chunk_size == 0u)
             throw std::invalid_argument{"The chunk size must be greater than 0."};
 
+        resource = views::zip(std::forward<resource_t>(resrc), std::views::iota(0)) | views::chunk(chunk_size);
```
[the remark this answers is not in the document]
rrahn: Well, not segfaulting but it should throw. I guess I should test it but anyway. I believe it is a wrong input and therefor I make it after I am sure that the input is valid.

PR #1362, opened 2019-11-07
[FIX] Fixes aligned allocator issue for non posix implementation of aligned_alloc.

rrahn:
* Uses operator new and delete instead of aligned_alloc.

### DIFF
--- a/include/seqan3/range/container/aligned_allocator.hpp
+++ b/include/seqan3/range/container/aligned_allocator.hpp
@@ -28,6 +28,12 @@ namespace seqan3
  *
  * \details
  *
+ * This class allocates memory at the given alignment boundary. If the specified `alignment` is not supported by the
+ * used allocation method a std::bad_alloc exception will be thrown. For requested alignments larger than
+ * `__STDCPP_DEFAULT_NEW_ALIGNMENT__`, also called new-extended alignments, the storage will have the alignment
+ * specified by the value `alignment`. Otherwise, the storage is aligned for any object that does not have new-extended
+ * alignment, e.g. int or double, and is of the requested size.
+ *
  * \include test/snippet/range/container/aligned_allocator.cpp
  *
  * Will output something like:
@@ -59,7 +65,7 @@ template <typename value_t, size_t alignment_v = __STDCPP_DEFAULT_NEW_ALIGNMENT_
 class aligned_allocator
 {
 public:
-    //!\brief The memory-alignment of the allocation
+    //!\brief The memory-alignment of the allocation. Defaults to `__STDCPP_DEFAULT_NEW_ALIGNMENT__`.
     static constexpr size_t alignment = alignment_v;
 
     //!\brief The value type of the allocation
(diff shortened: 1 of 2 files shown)


include/seqan3/range/container/aligned_allocator.hpp
```diff
@@ -84,72 +90,108 @@ class aligned_allocator
...
+     * \details
+     *
+     * If the alignment of the new type exceeds the alignment of the current allocator, the larger alignment will
+     * be used.
      */
     template <typename new_value_type>
     struct rebind
     {
+        //!\brief The alignment for the rebound allocator.
+        static constexpr size_t other_alignment = (alignof(new_value_type) > alignment)
```
marehr: std::max ?

include/seqan3/range/container/aligned_allocator.hpp
```diff
@@ -28,6 +28,12 @@ namespace seqan3
  *
  * \details
  *
+ * This class allocates memory at the given alignment boundary. If the specified `alignment` is not supported by the
+ * used allocation method a std::bad_alloc exception will be thrown. For requested alignments larger than
+ * `__STDCPP_DEFAULT_NEW_ALIGNMENT__` the storage will have the alignment specified by the value `alignment`.
+ * Otherwise, the storage is aligned for any object that does not have new-extended alignment and is of the requested
```
marehr: What is new-extended? Can you link it to some cppreference page?

marehr: And is the other one not of the requested size?

marehr: Can you give some examples which "objects" might be effected by the else case?

include/seqan3/range/container/aligned_allocator.hpp
```diff
@@ -28,6 +28,12 @@ namespace seqan3
...
+ * size.
+ *
  * \include test/snippet/range/container/aligned_allocator.cpp
  *
  * Will output something like:
```
marehr: Does the below example still apply?

include/seqan3/range/container/aligned_allocator.hpp
```diff
@@ -84,72 +90,108 @@ class aligned_allocator
...
+     *         operator new function throws or the requested memory exceeds the maximal number of elements to allocate.
+     *
+     * \details
+     *
+     * Allocates `n * sizeof(value_type)` bytes of uninitialized storage by calling
+     * [operator new](https://en.cppreference.com/w/cpp/memory/new/operator_new).
+     * If the given `alignment` is bigger than
+     * [__STDCPP_DEFAULT_NEW_ALIGNMENT__](https://en.cppreference.com/w/cpp/memory/new/align_val_t) the alignment
+     * aware operator new that takes as second argument the alignment as std::align_val_t. The allocation methods
+     * used by the global operator new functions is unspecified, e.g. std::malloc or std::aligned_alloc. Accordingly,
```
marehr:
```suggestion
     * The allocation methods, e.g. std::malloc or std::aligned_alloc, used by the global operator new functions is unspecified. Accordingly,
```

include/seqan3/range/container/aligned_allocator.hpp
```diff
@@ -84,72 +90,108 @@ class aligned_allocator
...
+     * the requirements on the alignment and the size of allocated bytes is implementation defined.
```
marehr: Does this has any impact on the intended behaviour of this class?

include/seqan3/range/container/aligned_allocator.hpp
```diff
@@ -84,72 +90,108 @@ class aligned_allocator
...
-        // causes the function to fail and return a null pointer (C11, as
-        // published, specified undefined behavior in this case, this was
-        // corrected by DR 460).
-        // https://en.cppreference.com/w/cpp/memory/c/aligned_alloc
-        if (auto p = static_cast<pointer>(std::aligned_alloc(alignment, n * sizeof(value_type))))
-            return p;
-#endif
-
-        throw std::bad_alloc();
+            throw std::bad_alloc{};
```
marehr: codecov is complaining that this is not tested.

Can you test it with a ridiculous high allocation request? Like, let's say `std::numeric_limits<size_type>::max()` for value_type `uint64_t`.

include/seqan3/range/container/aligned_allocator.hpp
```diff
@@ -84,72 +90,108 @@ class aligned_allocator
...
+
+        size_t bytes_to_allocate = n * sizeof(value_type);
+        if constexpr (alignment <= __STDCPP_DEFAULT_NEW_ALIGNMENT__)
+            return static_cast<pointer>(::operator new(bytes_to_allocate));
+        else // Use alignment aware allocator function.
+            return static_cast<pointer>(::operator new(bytes_to_allocate, static_cast<std::align_val_t>(alignment)));
```
marehr: before this change we also throwed a bad-alloc on a returned null-pointer, can we have this behaviour again?

include/seqan3/range/container/aligned_allocator.hpp
rrahn: yes we should test it.

test/unit/range/container/aligned_allocator_test.cpp
```diff
@@ -80,6 +80,68 @@ TEST(aligned_allocator, memory_alignment)
     alloc.deallocate(begin, size);
 }
 
+TEST(aligned_allocator, memory_alignment_new_extended)
```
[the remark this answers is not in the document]
rrahn: this is language from the c++ draft. It means that it's alignment is bigger than the value of `__STDCPP_DEFAULT_NEW_ALIGNMENT__`

test/unit/range/container/aligned_allocator_test.cpp
```diff
@@ -80,6 +80,68 @@ TEST(aligned_allocator, memory_alignment)
...
+    large_alignment * end   = begin + size;
+
+    constexpr size_t alignment = alignof(large_alignment);
+    EXPECT_EQ(sizeof(large_alignment), 64u);
+    EXPECT_EQ(alignof(large_alignment), 64u);
+
+    EXPECT_EQ(memory_alignment(begin, alignment), 0u);
+    EXPECT_EQ(memory_alignment(end,   alignment), 0u);
+
+    EXPECT_EQ(memory_alignment(begin + 1,  alignment), 0u);
```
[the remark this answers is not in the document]
rrahn: because it does not use a larger alignment than the alignment of the type that is allocated for (which is 64 bytes). So every value starts at 64 byte boundary. Also that's why I called this test default to show the behavior of a new extended type where we use the default alignment of the type and not a larger one.

include/seqan3/range/container/aligned_allocator.hpp
rrahn: this is not necessary anymore because I use the throwing versions of operator new. These are defined in the standard to always return non null-pointer values or else throw. So we don't need to handle the output anymore.

include/seqan3/range/container/aligned_allocator.hpp
rrahn: I don't understand you question. Could you be more specific? I mean this is the same problem when to use std::aligned_alloc. Depending on how operator new is implemented and depending on the platform different things are possible. So for example it might still be a problem in the future, that on an architecture that uses posix mem_align arbitrary sizes of bytes to allocate can be used, while on others this might be restricted to sizes that are a multiple of the alignment. But that is implementation defined and beyond the scope of this class. It only makes sure that it calls aligned memory, where it uses whatever `::operator new(n)` uses to align up to alignments of `__STDCPP_DEFAULT_NEW_ALIGNMENT__` and beyond that  aligns the memory depending on whatever `::operator new(n, alignment)` uses.

include/seqan3/range/container/aligned_allocator.hpp
rrahn: like everything up to `__STDCPP_DEFAULT_NEW_ALIGNMENT__` bytes. IIRC that is 16 byte. But I look into this and extend the example with some comments. I guess it shows this already.

include/seqan3/range/container/aligned_allocator.hpp
rrahn: yes. I made sure of it.

include/seqan3/range/container/aligned_allocator.hpp
marehr: Note: We call the new operator with the semantic requirements that the c++ standard specifies/demands, but be aware that users can overload any (global) new operator that might not adhere to the standard and might cause std::bad_alloc or unaligned pointer.

include/seqan3/range/container/aligned_allocator.hpp
```diff
@@ -28,6 +28,12 @@ namespace seqan3
  *
  * \details
  *
+ * This class allocates memory at the given alignment boundary. If the specified `alignment` is not supported by the
```
smehringer: can you always use the term `memory-alignment` instead of just `alignment` like in the documentation below? It helps to know up front that this has nothing to do with our pairwise alignment

include/seqan3/range/container/aligned_allocator.hpp
```diff
@@ -59,7 +65,7 @@ template <typename value_t, size_t alignment_v = __STDCPP_DEFAULT_NEW_ALIGNMENT_
 class aligned_allocator
 {
 public:
-    //!\brief The memory-alignment of the allocation
+    //!\brief The memory-alignment of the allocation. Defaults to `__STDCPP_DEFAULT_NEW_ALIGNMENT__`.
     static constexpr size_t alignment = alignment_v;
```
smehringer: There is no defualt?

include/seqan3/range/container/aligned_allocator.hpp
```diff
@@ -28,6 +28,12 @@ namespace seqan3
  *
  * \details
  *
+ * This class allocates memory at the given alignment boundary. If the specified `alignment` is not supported by the
+ * used allocation method a std::bad_alloc exception will be thrown. For requested alignments larger than
+ * `__STDCPP_DEFAULT_NEW_ALIGNMENT__`, also called new-extended alignments, the storage will have the alignment
+ * specified by the value `alignment`. Otherwise, the storage is aligned for any object that does not have new-extended
+ * alignment, e.g. int or double, and is of the requested size.
```
smehringer: From this text, I do not understand what a `memory alignment` is.
If you think this is common knowledge, can you give just me an explanation? If not, one or few sentences would be great since this is a public interface.  

Side note: I see that the number is always a multiple of the word size. Is this required and if so do you static_assert this?

include/seqan3/range/container/aligned_allocator.hpp
```diff
@@ -84,72 +90,108 @@ class aligned_allocator
...
+     * \throws Throws std::bad_alloc if allocation fails, i.e. either the call to the throwing version of
+     *         operator new function throws or the requested memory exceeds the maximal number of elements to allocate.
+     *
+     * \details
+     *
+     * Allocates `n * sizeof(value_type)` bytes of uninitialized storage by calling
+     * [operator new](https://en.cppreference.com/w/cpp/memory/new/operator_new).
+     * If the given `alignment` is bigger than
+     * [__STDCPP_DEFAULT_NEW_ALIGNMENT__](https://en.cppreference.com/w/cpp/memory/new/align_val_t) the alignment
+     * aware operator new that takes as second argument the alignment as std::align_val_t.
```
smehringer: I think this sentence is mal formatted... at least I cannot make sense of it

include/seqan3/range/container/aligned_allocator.hpp
```diff
@@ -84,72 +90,108 @@ class aligned_allocator
...
+     * to seqan3::aligned_allocator::allocate.
+     *
+     * \param[in] p The pointer to the memory to be deallocated.
+     * \param[in] n The number of elements to be deallocated.
+     *
      * \details
      *
-     * The argument n must be equal to the first argument of the call to allocate() that originally produced p;
-     * otherwise, the behavior is undefined.
+     * The argument `n` must be equal to the first argument of the call to seqan3::aligned_allocator::allocate that
```
smehringer: This seems error prone? Can't we cache `n` and remove the parameter from this function?

test/unit/range/container/aligned_allocator_test.cpp
```diff
@@ -47,6 +48,12 @@ TEST(aligned_allocator, conversion_constructor)
     [[maybe_unused]] aligned_allocator<float, 16> float_alloc{int_alloc};
 }
 
+TEST(aligned_allocator, request_to_much_memory)
```
smehringer:
```suggestion
TEST(aligned_allocator, request_too_much_memory)
```
:nail_care:

include/seqan3/range/container/aligned_allocator.hpp
rrahn: should be put above

include/seqan3/range/container/aligned_allocator.hpp
rrahn: This is defined and documented according to the allocator interface requirements: https://en.cppreference.com/w/cpp/named_req/Allocator, and I am afraid we need to keep this interface.

include/seqan3/range/container/aligned_allocator.hpp
rrahn: I will add a section about what memory alignment is.

include/seqan3/range/container/aligned_allocator.hpp
rrahn: > Side note: I see that the number is always a multiple of the word size. Is this required and if so do you static_assert this?

I am not sure I understand what do you mean?

PR #1369, opened 2019-11-21
[FEATURE] Add type_name function to get human-readable name of a dynamic type.

rrahn:
Adds a function to use typeid operator to get the name of the type at runtime.
This name is human-readable and not mangled as it is the standard for gcc and clang.

This is a follow up of #1367 which also handles mangled names.


h-2:
Can we make the machinery for "typename2string" and "typename2string_at_compile_time" more alike?

rrahn:
> Can we make the machinery for "typename2string" and "typename2string_at_compile_time" more alike?

Yes, I thought about this. That's why I put it into reflection for now. Also I don't think it should be called reflection as it suggests more type inspection then just getting the name. Maybe more like type_info.hpp or something. But I didn't want to change this in this PR. Actually, the changes here are supposed to be part of the already merged PR but now it is just a followup PR because I wasn't fast enough. Anyhow, I put this on the discussion for next strategy meeting to make it more alike.

rrahn:
Ahhhrrg, if we make this change then again the ranges is included because of the include of reflection which includes small_string! What a bummer! So, there need to be another solution for this. Otherwise the entire changes to typeid would have been useless in the first place.

test/unit/core/detail/type_inspection_test.cpp
```diff
@@ -0,0 +1,54 @@
...
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+#include <gtest/gtest.h>
+
+#include <seqan3/core/type_list/traits.hpp>
+#include <seqan3/core/type_list/type_list.hpp>
+#include <seqan3/core/detail/type_inspection.hpp>
+
+using namespace seqan3;
```
eseiler: We do not want to do this anymore

test/unit/core/detail/type_inspection_test.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+// Some test namespace to check if namespace information are preserved within the naming.
+namespace foo
+{
+template <typename ...type>
+struct bar
+{};
+} // namespace foo
+
+// Some types to test if type inspetion works as expected.
```
eseiler:
```suggestion
// Some types to test if type inspection works as expected.
```

test/unit/core/detail/type_inspection_test.cpp
```diff
@@ -0,0 +1,54 @@
...
+// Note, that the returned name might differ between compiler vendors and thus must be adapted accordingly,
```
eseiler:
```suggestion
// Note that the returned name might differ between compiler vendors and thus must be adapted accordingly
```

test/unit/core/detail/type_inspection_test.cpp
rrahn: since when, and what would be the rational for this?

h-2: Please read up on the meeting notes if you haven't been there :wink:

rrahn: i did but i must have missed this.

test/unit/core/detail/type_inspection_test.cpp
```diff
@@ -0,0 +1,52 @@
...
+                                          foo::bar<char>, foo::bar<foo::bar<char, double>>>;
+
+// Helper type list to use some traits functions on type lists.
+using as_type_list_t = seqan3::detail::transfer_template_args_onto_t<reflection_types, seqan3::type_list>;
+
+template <typename param_type>
+class type_inspection : public ::testing::Test
+{
+    // The corresponding list of names that should be generated. Must have the same order as `reflection_types`.
+    inline static const std::vector names{"char", "char16_t", "char32_t", "short", "double",
```
smehringer: Maybe add an alphabet of ours to test?

rrahn: but, that's why I used the special namespace variant, to show that it does work with types within a namespace. What would be different for our types in seqan3 namespace?

smehringer: Yes fine. I just thought it would be easier then creating an extra type withing a namespace. but I don't care :) I already approved

PR #1373, opened 2019-11-22
Feature/pass through chunk to alignment call

### LINKED ISSUE #1372: ### Range awareness - level4
**TASK: Pass a chunk of the range directly to the execution handler**

- [x] Depends on [Chunk range of sequence pair according to the given parameter in alignment_executor_two_way](https://github.com/seqan/seqan3/projects/10#card-27977713)
- [x] Depends on [execution_handler_sequential should operate on a range of sequence pairs](https://github.com/seqan/seqan3/projects/10#card-27977085)
- [x] Depends on [execution_handler_parallel should operate on a range of sequence pairs](https://github.com/seqan/seqan3/projects/10#card-27977114)
- [x] Pass range of sequence pairs (chunk) directly to the execution_handler
- [x] Use the new `exec_handler.execute(kernel, start_id, chunk, callback)` syntax.

rrahn:
Fixes #1372

The alignment executor will now call the execution handler directly with passing the chunk of indexed sequence pairs. This also means that now the return type is a vector which needs to be moved to the result buffer.

### DIFF
--- a/include/seqan3/alignment/pairwise/align_pairwise.hpp
+++ b/include/seqan3/alignment/pairwise/align_pairwise.hpp
@@ -25,6 +25,8 @@
 #include <seqan3/alignment/pairwise/execution/all.hpp>
 #include <seqan3/core/algorithm/all.hpp>
 #include <seqan3/core/parallel/execution.hpp>
+#include <seqan3/core/simd/simd_traits.hpp>
+#include <seqan3/core/simd/simd.hpp>
 #include <seqan3/core/type_traits/basic.hpp>
 #include <seqan3/range/views/persist.hpp>
 #include <seqan3/std/concepts>
@@ -162,36 +164,29 @@ constexpr auto align_pairwise(sequence_t && sequences,
     // Pipe with views::persist to allow rvalue non-view ranges.
     auto seq_view = std::forward<sequence_t>(sequences) | views::persist;
     // Configure the alignment algorithm.
-    auto kernel = detail::alignment_configurator::configure<decltype(seq_view)>(config);
+    auto && [algorithm, adapted_config] = detail::alignment_configurator::configure<decltype(seq_view)>(config);
 
     //!brief Lambda function to translate specified parallel and or vectorised configurations into their execution rules.
     constexpr auto get_execution_rule = [] ()
     {
-        using exec_rule = std::remove_reference_t<alignment_config_t>;
-
-        if constexpr (exec_rule::template exists<align_cfg::parallel>())
-        {
-            if constexpr (exec_rule::template exists<detail::vectorise_tag>())
-            {
-                return seqan3::par_unseq;
-            }
-            else
-            {
+        if constexpr (alignment_config_t::template exists<align_cfg::parallel>())
                 return seqan3::par;
-            }
-        }
-        else if constexpr (exec_rule::template exists<detail::vectorise_tag>())
-        {
-            return seqan3::unseq;
-        }
         else
-        {
             return seqan3::seq;
-        }
     };
 
+    size_t chunk_size = 1;
+    if constexpr (alignment_config_t::template exists<detail::vectorise_tag>())
+    {
+        using score_t = typename detail::align_config_result_score<decltype(adapted_config)>::type;
+        chunk_size = simd_traits<simd_type_t<score_t>>::length;
+    }
+
     // Create a two-way executor for the alignment.
-    detail::alignment_executor_two_way executor{std::move(seq_view), kernel, 1, get_execution_rule()};
+    detail::alignment_executor_two_way executor{std::move(seq_view),
+                                                std::move(algorithm),
+                                                chunk_size,
+                                                get_execution_rule()};
     // Return the range over the alignments.
     return alignment_range{std::move(executor)};
 }
--- a/include/seqan3/alignment/pairwise/align_pairwise.hpp
+++ b/include/seqan3/alignment/pairwise/align_pairwise.hpp
@@ -153,8 +153,8 @@ template <typename sequence_t, typename alignment_config_t>
 constexpr auto align_pairwise(sequence_t && sequences,
                               alignment_config_t const & config)
 {
-    using first_seq_t  = std::tuple_element_t<0, value_type_t<sequence_t>>;
-    using second_seq_t = std::tuple_element_t<1, value_type_t<sequence_t>>;
+    using first_seq_t  = std::tuple_element_t<0, std::ranges::range_value_t<sequence_t>>;
+    using second_seq_t = std::tuple_element_t<1, std::ranges::range_value_t<sequence_t>>;
 
     static_assert(std::ranges::random_access_range<first_seq_t> && std::ranges::sized_range<first_seq_t>,
                   "Alignment configuration error: The sequence must model random_access_range and sized_range.");
--- a/include/seqan3/alignment/pairwise/alignment_configurator.hpp
+++ b/include/seqan3/alignment/pairwise/alignment_configurator.hpp
@@ -56,6 +56,29 @@ SEQAN3_CONCEPT align_pairwise_single_input =
     std::ranges::viewable_range<std::tuple_element_t<0, value_t>> &&
     std::ranges::viewable_range<std::tuple_element_t<1, value_t>>;
 
+/*!\brief A transformation trait to extract the score type used within the seqan3::align_cfg::result object.
+ * \implements seqan3::transformation_trait
+ * \ingroup pairwise_alignment
+ *
+ * \tparam config_t The alignment configuration type; must be of type seqan3::configuration and must contain
+ *                  a seqan3::align_cfg::result object.
+ */
+template <typename config_t>
+//!\cond
+    requires is_type_specialisation_of_v<config_t, configuration> && config_t::template exists<align_cfg::result>()
+//!\endcond
+struct align_config_result_score
+{
+private:
+
+    //!\brief Helper type definition to store the type of the result config.
+    using result_config_t = std::remove_reference_t<decltype(seqan3::get<align_cfg::result>(std::declval<config_t>()))>;
+public:
+
+    //!\brief The score type used for the alignment result configuration.
+    using type = typename result_config_t::score_type;
+};
+
 /*!\brief A helper concept to test for correct range input in seqan3::align_pairwise.
  * \ingroup pairwise_alignment
  *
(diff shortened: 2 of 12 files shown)


include/seqan3/alignment/pairwise/detail/concept.hpp
```diff
@@ -14,35 +14,78 @@
...
+    requires std::tuple_size_v<t> == 2;
+    requires std::ranges::forward_range<std::tuple_element_t<0, t>>;
+    requires std::ranges::forward_range<std::tuple_element_t<1, t>>;
+    requires semialphabet<std::ranges::range_value_t<std::tuple_element_t<0, t>>>;
+    requires semialphabet<std::ranges::range_value_t<std::tuple_element_t<1, t>>>;
+};
+//!\endcond
+
+/*!\interface seqan3::detail::sequence_pair_range <>
+ * \brief A helper concept to check if a type is a range over seqan3::detail::sequence_pair's.
```
joergi-w: ...over `seqan3::detail::sequence_pair`.

include/seqan3/alignment/pairwise/detail/concept.hpp
```diff
@@ -14,35 +14,78 @@
...
+ * \ingroup pairwise_alignment
+ *
+ * \tparam t The type to check.
+ *
+ * \details
+ *
+ * This concept checks if the given type models a std::ranges::forward_range over and that the value type of the
```
joergi-w: ~over~

include/seqan3/alignment/pairwise/detail/concept.hpp
```diff
@@ -14,35 +14,78 @@
...
-SEQAN3_CONCEPT indexed_sequence_pair_range = requires (std::ranges::range_value_t<t> value)
+SEQAN3_CONCEPT indexed_sequence_pair_range = std::ranges::forward_range<t> &&
+                                             requires (std::ranges::range_value_t<t> value)
 {
-    requires std::ranges::forward_range<t>;
     requires tuple_like<decltype(value)>;
     requires std::tuple_size_v<decltype(value)> == 2;
-    requires tuple_like<std::tuple_element_t<0, decltype(value)>>;
-    requires std::tuple_size_v<std::tuple_element_t<0, decltype(value)>> == 2;
+    requires sequence_pair<std::tuple_element_t<0, decltype(value)>>;
```
joergi-w: Idea: Should we require some numeric/integral type or at least EqualityComparable for the index? Otherwise users are able to set weird types, which may cause side effects at some point.

include/seqan3/alignment/pairwise/detail/utility_traits.hpp
```diff
@@ -0,0 +1,45 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides utility traits for the configuration and execution of the alignment algorithm.
```
joergi-w: The name `utility traits` does not give any insight what is in this file.
Consider renaming the file to chunk_traits.hpp or similar (even if you plan to extend this file later).

include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
```diff
@@ -302,11 +300,13 @@ class alignment_executor_two_way
     void move_initialise(alignment_executor_two_way && other) noexcept
     {
         kernel = std::move(other.kernel);
+        _chunk_size = std::move(other._chunk_size);
```
joergi-w: Could you remove the leading underscore from the member variable `chunk_size`? I don't see any reason for this and all other members are also without leading `_`.

include/seqan3/alignment/pairwise/detail/concept.hpp
rrahn: The thing is I am not yet completey sure what the requested index type should be. Is it always just a numeric value? It certainly will be at some point an pair over numeric values. Can it be something more complex? As a matter of fact, we also don't care. We just propagate the index as is to the alignment result. So we don't do any operations on that. The only thing we can require is copyability. That is something I can indeed test.

include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
rrahn: Actually, I'd love to but there is a function called chunk_size as well and that doesn't work. I do not want to invent a new name for the member and rather stick in these rare cases to what most people do extending the name with an underscore.

include/seqan3/alignment/pairwise/detail/utility_traits.hpp
rrahn: I renamed it to type_traits. And yes it is already extended in another open PR.

rrahn:
@joergi-w I addressed all your comments. I have one more question though. Could you next time please use the suggestion feature for small changes like documentation issues? Then I can directly submit them as a commit and don't have to find the places in code that need to be adapted. That would be of great help and speedup applying the changes. Thank you very much.

include/seqan3/alignment/pairwise/detail/concept.hpp
joergi-w: Okay, that's a good solution.

joergi-w:
> @joergi-w I addressed all your comments. I have one more question though. Could you next time please use the suggestion feature for small changes like documentation issues? Then I can directly submit them as a commit and don't have to find the places in code that need to be adapted. That would be of great help and speedup applying the changes. Thank you very much.

I actually would love to use the suggestion tool whenever applicable. Unfortunately, Github doesn't offer it when I display the changes commit-wise... (Error message: You are viewing a subset of changes.) 
Does it work for you or do you know a solution?

rrahn:
No not a perfect solution. I would annotate stuff and later make them suggestions. But also agree that this is cumbersome for the reviewer.

smehringer:
> I actually would love to use the suggestion tool whenever applicable. Unfortunately, Github doesn't offer it when I display the changes commit-wise... (Error message: You are viewing a subset of changes.)
Does it work for you or do you know a solution?

You can just manually type \`\`\`suggestion ...```!

include/seqan3/alignment/pairwise/detail/type_traits.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+/*!\brief A transformation trait to retrieve the chunked range over indexed sequence pairs.
+ * \ingroup pairwise_alignment
+ * \implements seqan3::transformation_trait
+ *
+ * \tparam sequence_pairs_t The type of the sequences to be transformed; must model seqan3::detail::sequence_pair_range.
+ *
+ * \details
+ *
+ * This transformation trait transforms a range over sequence pairs into an indexed range over sequence pairs.
```
smehringer:
```suggestion
 * This transformation trait transforms a range over sequence pairs into a range over indexed sequence pairs.
```
?

include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
```diff
@@ -56,19 +57,19 @@ class alignment_executor_two_way
...
      */
-    //!\brief The underlying resource type.
-    using resource_type = decltype(views::zip(std::declval<resource_t>(), std::views::iota(0)) | views::chunk(1));
+    //!\brief The underlying resource type augmented with an index and a chunked view.
+    using chunked_resource_type = typename chunked_indexed_sequence_pairs<resource_t>::type;
     //!\brief The iterator over the underlying resource.
-    using resource_iterator_t = std::ranges::iterator_t<resource_type>;
-    //!\brief The value type of the resource.
-    using resource_value_type = std::ranges::range_value_t<resource_type>;
+    using chunked_chunked_resource_iterator_t = std::ranges::iterator_t<chunked_resource_type>;
```
smehringer: sometimes you abbreviate with `_t` and sometimes you call it `_type` (`chunked_resource_type` vs `chunked_chunked_resource_iterator_t`)
Maybe change one of them to match the others? (wwithout chenging the whole file but only the lines you touch anyway)

include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
```diff
@@ -257,27 +257,25 @@ class alignment_executor_two_way
         // Apply the alignment execution.
         size_t count = 0;
         size_t buffer_limit = in_avail();
-        assert(buffer_limit >= chunk_size());
 
-        while (count < buffer_limit && !is_eof())
+        size_t current_chunk_size = std::ranges::distance(*chunked_resource_it);
```
smehringer: maybe assert that the `chunked_resource_it` is a forward_iterator? might save some debugging time if we ever use a wrong view on top

smehringer: and you can set this variable inside of the for loop signature.

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -162,36 +164,29 @@ constexpr auto align_pairwise(sequence_t && sequences,
...
     //!brief Lambda function to translate specified parallel and or vectorised configurations into their execution rules.
     constexpr auto get_execution_rule = [] ()
     {
-        using exec_rule = std::remove_reference_t<alignment_config_t>;
-
-        if constexpr (exec_rule::template exists<align_cfg::parallel>())
-        {
-            if constexpr (exec_rule::template exists<detail::vectorise_tag>())
-            {
-                return seqan3::par_unseq;
```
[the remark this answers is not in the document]
rrahn: Because we don't have a special excecution handler for vectorised code. We simply do the chunking as a high level functionality in the alignment executor. Only the algorithm that receives the chunks does the vectorisation or not.

include/seqan3/alignment/pairwise/alignment_configurator.hpp
```diff
@@ -194,13 +218,19 @@ struct alignment_configurator
...
      * \details
      *
      * This function reads the seqan3::configuration object and generates the corresponding alignment algorithm type.
-     * During this process some runtime configurations are converted to static configurations if required. The return
-     * type is a std::function which is generated in the following way:
+     * During this process some runtime configurations are converted to static configurations if required.
+     * In case of a missing configuration that has a default, e.g. the seqan3::align_cfg::result option, the
+     * default version of this configuration element is added to the passed configuration object.
+     * The return type is a pair over a std::function object and the configuration object. Thus, the calling function
+     * has access to the possibly modified configuration object. This way it can examine the same configurations as
```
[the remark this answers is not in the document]
rrahn: I think the previous sentence was clear enough. I removed the second one.

include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
rrahn: right. I kept the naming close to stl. So iterator without any suffix and everything else with _type suffix.

include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
rrahn: > maybe assert that the chunked_resource_it is a forward_iterator? might save some debugging time if we ever use a wrong view on top

I don't see this as a necessary step as the underlying range is required to be forward range. I don't see why chunk view should ever degrade a forward range to an input range.

smehringer:
only serparate review commits into the commit structure

PR #1378, opened 2019-11-23
[FEATURE] Make the final chunk only as long as the longest sequence in the collection.

rrahn:
Before this change the last chunk always had a size of simd_traits<simd_t>::length, which caused some issues when dealing with the last chunk, since the last elements in the transformed sequence might not have carried any information anymore if the size of the longest sequence was not a multiple of the simd size. Instead, the new code tracks the end of the last column in the final chunk and returns a span that only covers the sequences.

### DIFF
--- a/include/seqan3/core/simd/view_to_simd.hpp
+++ b/include/seqan3/core/simd/view_to_simd.hpp
@@ -86,7 +86,7 @@ private:
     //!\brief Check if fast load is enabled.
     static constexpr bool fast_load = std::ranges::contiguous_range<inner_range_type> &&
                                       std::sized_sentinel_for<std::ranges::iterator_t<inner_range_type>,
-                                                         std::ranges::sentinel_t<inner_range_type>> &&
+                                                              std::ranges::sentinel_t<inner_range_type>> &&
                                       sizeof(alphabet_rank_t<value_type_t<inner_range_type>>) == 1 &&
                                       sizeof(scalar_type) <= 2; // micro benchmark suggest using int8_t or int16_t has best performance.
 
@@ -200,8 +200,8 @@ private:
  *
  * \details
  *
- * This iterator models std::input_iterator and std::Cpp17input_iterator.
- * When dereferencing it returns a reference to a range over simd vectors of type `simd_t`.
+ * This iterator models std::input_iterator.
+ * When dereferencing it returns a span over a range with value type `simd_t`.
  */
 template <std::ranges::view urng_t, simd::simd_concept simd_t>
 class view_to_simd<urng_t, simd_t>::iterator_type
@@ -210,7 +210,7 @@ public:
     /*!\name Associated types
      * \{
      */
-    using reference = chunk_type &; //!< The reference type.
+    using reference = std::span<std::ranges::range_value_t<chunk_type>>; //!< The reference type.
     using value_type = chunk_type; //!< The value type.
     using pointer = chunk_type *; //!< The pointer type.
     using difference_type = ptrdiff_t; //!< The difference type.
@@ -253,6 +253,8 @@ public:
             cached_sentinel[seq_id] = std::ranges::end(this_view.empty_inner_range);
         }
 
+        // initialises the chunk sizes for the respective chunks.
+        std::ranges::fill(current_chunk_sizes, chunk_size);
         // Check if this is the final chunk already.
         final_chunk = all_iterators_reached_sentinel();
 
@@ -268,7 +270,8 @@ public:
     constexpr reference operator*() const noexcept
     {
         assert(this_view != nullptr);
-        return this_view->cached_simd_chunks[current_chunk_pos];
+        return std::span{this_view->cached_simd_chunks[current_chunk_pos].begin(),
+                         current_chunk_sizes[current_chunk_pos]};
     }
 
     //!\brief Returns a pointer to the current chunk of simd vectors.
@@ -287,7 +290,7 @@ public:
         if constexpr (fast_load)
         { // Check if cached chunks have been already consumed and we need to fetch the next chunks.
             if (current_chunk_pos == final_chunk_pos)
-        {
+            {
                 underflow();
                 current_chunk_pos = 0;
             }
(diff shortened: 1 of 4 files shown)


marehr:
@rrahn I looked at the [code cov](https://codecov.io/gh/seqan/seqan3/compare/65f6ad001bc7976dbd56649038bc85ab5e4bf98c...483c96db7a7dbd68d1e147115b4fdf568ad98391/diff) and there are some untested places. Can you verify if it is intended that those places are not tested yet?

include/seqan3/core/simd/view_to_simd.hpp
```diff
@@ -532,9 +539,16 @@ class view_to_simd<urng_t, simd_t>::iterator_type
             // Subtract one from last_chunk_pos to get the correct 0-based index.
             final_chunk_pos = std::max<int8_t>(final_chunk_pos, --last_chunk_pos);
         }
+
         final_chunk = all_iterators_reached_sentinel();
-        simd::transpose(matrix);
 
+        if (final_chunk)
+        { // Store the size of the last chunk.
+            size_t size_of_last_chunk = max_size_of_last_chunk % chunk_size;
```
marehr: `max_size_of_last_chunk` is misleading, because it can be bigger than chunk_size. Or are you referring to a different chunk? If think `max_remaining_elements` would be better.

include/seqan3/core/simd/view_to_simd.hpp
```diff
@@ -503,27 +508,29 @@ class view_to_simd<urng_t, simd_t>::iterator_type
...
         {  // Iterate over each block depending on the packing of the target simd vector.
             uint8_t last_chunk_pos = 0;
-            for (uint8_t j = 0; j < num_chunks; ++j)
+            for (uint8_t chunk_pos = 0; chunk_pos < num_chunks; ++chunk_pos)
             {
-                uint8_t pos = j * chunk_size + i; // matrix entry to fill
-                if (cached_sentinel[i] - cached_iter[i] >= max_size) // not in final block
+                uint8_t pos = chunk_pos * chunk_size + sequence_pos; // matrix entry to fill
+                size_t current_chunk_size = cached_sentinel[sequence_pos] - cached_iter[sequence_pos];
+                max_size_of_last_chunk = std::max(max_size_of_last_chunk, current_chunk_size);
```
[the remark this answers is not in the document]
rrahn: actually, it feels that the entire loop could be improved. 
But that was not my primary goal here. So I didn't spent much time on making it real tight. Just wanted to get it working for now. When benchmarking suggests that there would be a measurable gain in doing this more optimal one can do it at a later point. But if you see immediate improvements we can discuss this and improve it.

include/seqan3/core/simd/view_to_simd.hpp
```diff
@@ -472,6 +467,33 @@ class view_to_simd<urng_t, simd_t>::iterator_type
...
+     * Sets the index of the final chunk (`final_chunk_pos`) and updates the end position of the final chunk such that
+     * the view ends at the last character of the longest sequences contained in the set of sequences to be transformed.
+     */
+    template <typename array_t>
+    constexpr void update_final_chunk_position(array_t const & iterators_before_update) noexcept
+    {
+        size_t max_distance = 0;
+        for (auto && [it, sent] : views::zip(iterators_before_update, cached_sentinel))
+            if (size_t distance = std::ranges::distance(it, sent); distance > max_distance)
+                max_distance = distance;
```
marehr: max_distance = std::max(std::ranges::distance(it, sent), max_distance);

include/seqan3/core/simd/view_to_simd.hpp
```diff
@@ -578,10 +586,10 @@ class view_to_simd<urng_t, simd_t>::iterator_type
     std::array<std::ranges::sentinel_t<inner_range_type>, chunk_size> cached_sentinel{};
     //!\brief Pointer to the associated range.
     view_to_simd * this_view{nullptr};
-    //!\brief An array that stores the end position for every sequence within the current chunk.
-    std::array<uint8_t, total_chunks> current_chunk_sizes{};
+    //!\brief Stores the size of all chunks at index 0 and the size of the final chunk at index 1.
+    std::array<uint8_t, 2> current_chunk_sizes{chunk_size, chunk_size};
```
marehr: as we talked about, this should be made into two variables `chunk_size` and `final_chunk_size`. (since `current_chunk_sizes[0]` stays always the same and is effectively `chunk_size`, we only need `final_chunk_size`.)

include/seqan3/core/simd/view_to_simd.hpp
```diff
@@ -479,16 +473,16 @@ class view_to_simd<urng_t, simd_t>::iterator_type
...
-            if (size_t distance = std::ranges::distance(it, sent); distance > max_distance)
-                max_distance = distance;
+            max_distance = std::max<size_t>(std::ranges::distance(it, sent), max_distance);
+
 
         assert(max_distance > 0);
         assert(max_distance <= (total_chunks * chunk_size));
 
         --max_distance;
         final_chunk_pos = max_distance  / chunk_size;
```
marehr: this is called `final_chunk_pos` ...

include/seqan3/core/simd/view_to_simd.hpp
```diff
@@ -479,16 +473,16 @@ class view_to_simd<urng_t, simd_t>::iterator_type
...
         // first we should be able to check the chunk position.
-        current_chunk_sizes[1] = (max_distance % chunk_size) + 1;
+        size_of_final_chunk = (max_distance % chunk_size) + 1;
```
marehr: ... and this `size_of_final_chunk`.

I think `final_chunk_size` is okay as name.

include/seqan3/core/simd/view_to_simd.hpp
```diff
@@ -211,7 +211,7 @@ class view_to_simd<urng_t, simd_t>::iterator_type
      */
     using reference = std::span<std::ranges::range_value_t<chunk_type>>; //!< The reference type.
     using value_type = chunk_type; //!< The value type.
-    using pointer = chunk_type *; //!< The pointer type.
+    using pointer = void; //!< The pointer type.
```
marehr: This has no implications? I didn't know that setting the pointer type as void is valid :)

rrahn: Yes. The pointer type is only required if you can use the operator->. Since, we return a proxy we can only make this safe by returning a unique_ptr. But the question is if that is something you want to have. Right now it merely means that it cannot be a CPP17Iterator. But I can live with that.

rrahn:
@marehr polite ping

include/seqan3/core/simd/view_to_simd.hpp
```diff
@@ -428,43 +421,68 @@ class view_to_simd<urng_t, simd_t>::iterator_type
...
+     * Sets the index of the final chunk (`final_chunk_pos`) and updates the end position of the final chunk such that
+     * the view ends at the last character of the longest sequences contained in the set of sequences to be transformed.
+     */
+    template <typename array_t>
+    constexpr void update_final_chunk_position(array_t const & iterators_before_update) noexcept
+    {
+        size_t max_distance = 0;
+        for (auto && [it, sent] : views::zip(iterators_before_update, cached_sentinel))
+            max_distance = std::max<size_t>(std::ranges::distance(it, sent), max_distance);
+
```
smehringer: remove extra newline :nail_care:

Issue #1381, opened 2019-11-26
Build fails on Raspberry PI architectures (armel, armhf, and arm64)


mr-c:
Since Raspberry PI is popular in classrooms and with technology enthusiasts, it would be nice if Seqan3 worked there as well.

- [x] non-x86 systems do not have `xmmintrin.h` which is a mandatory include in`/include/seqan3/core/parallel/detail/spin_delay.hpp` Needs to be made conditional and a non-SSE spin lock added for armel armhf arm64. Fix at https://github.com/seqan/seqan3/pull/1382

- [x]  `error: narrowing conversion of ‘'\213'’ from ‘char’ to ‘signed char’ ` and similar
(already fixed since the 3.0.0 release) Backported patch at https://salsa.debian.org/med-team/seqan3/blob/2f1ba053284b6464fe691a137c1008de6a4bf652/debian/patches/magic_int_to_char

- [ ] ` error: ‘cur.seqan3::fm_index_cursor<seqan3::fm_index<false> >::parent_lb’ may be used uninitialized in this function` and similar 
```
include/seqan3/search/fm_index/fm_index_cursor.hpp: In function ‘auto seqan3::search(queries_t&&, const index_t&, const configuration_t&) [with index_t = seqan3::fm_index<false>; queries_t = std::vector<seqan3::dna4>; configuration_t = seqan3::configuration<seqan3::search_cfg::max_error_rate<seqan3::search_cfg::total<double>, seqan3::search_cfg::deletion<double> > >]’:
/build/seqan3-3.0.0+ds2/include/seqan3/search/fm_index/fm_index_cursor.hpp:314:19: error: ‘cur.seqan3::fm_index_cursor<seqan3::fm_index<false> >::parent_lb’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
  314 |         size_type new_parent_lb = parent_lb, new_parent_rb = parent_rb;
      |                   ^~~~~~~~~~~~~
```
Full log for the above is at https://gist.github.com/mr-c/5e062fa2e9f8dc1e68674480e87ad13f

- [x] And some failed tests after adding `-Wno-narrowing -Wno-maybe-unitialized` to get past the above:
```
[ RUN      ] alignment_file_input_f.decompression_by_filename_gz
unknown file: Failure
C++ exception with description "No valid format found for this extension." thrown in the test body.
[  FAILED  ] alignment_file_input_f.decompression_by_filename_gz (2 ms)
[...]

PR #1382, opened 2019-11-26
[FIX] support spins on armel armhf arm64

mr-c:
Partially addresses https://github.com/seqan/seqan3/issues/1381

### DIFF
--- a/include/seqan3/core/parallel/detail/spin_delay.hpp
+++ b/include/seqan3/core/parallel/detail/spin_delay.hpp
@@ -13,7 +13,10 @@
 #pragma once
 
 #include <thread>
-#include <xmmintrin.h>
+
+#if defined(__SSE2__)  // AMD and Intel
+#include <xmmintrin.h>  // _mm_pause()
+#endif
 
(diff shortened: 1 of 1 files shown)


include/seqan3/core/parallel/detail/spin_delay.hpp
```diff
@@ -72,6 +75,10 @@ class spin_delay
     {
         #if defined(__SSE2__)  // AMD and Intel
             _mm_pause();
+	#elif defined(__armel__) || defined(__ARMEL__) // armel, but broken
```
rrahn: it seems the alignment to the other statements is off by 2 spaces

rrahn: Why is this not covered by the final else block? Or does __armel__ also define __arm__ but not the other way around?

mr-c: Arggh.. spaces vs. tabs :-P

mr-c: Dropped

PR #1394, opened 2019-12-02
[FIX] don't hard-fail on concept check

h-2:
You may not do the things you are doing in the signature of the requires expression, because their validity depends on the result of the requires expression.

This leads to hard-failure for types that would normally soft-fail. Since the concept is checked on globally defined `operator|` this can lead to all sorts of view-resolution failures to land here and hard-fail.

@sarahet Please check if this removes the first error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,8 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 * **Changed class signature of input/output file validators:**
   Most user code will be unaffected; to fix possible compiler errors you need to add an empty template parameter list to
   the respective instances (e.g. change `input_file_validator` to `input_file_validator<>`).
+* The member type that denotes which arguments a `validator` can validate has been renamed from `value_type` to
+  `option_value_type`.
 
 #### Core
 
--- a/include/seqan3/argument_parser/argument_parser.hpp
+++ b/include/seqan3/argument_parser/argument_parser.hpp
@@ -231,7 +231,7 @@ public:
     template <typename option_type, validator validator_type = detail::default_validator<option_type>>
     //!\cond
         requires (argument_parser_compatible_option<option_type> ||
-                  argument_parser_compatible_option<typename option_type::value_type>) &&
+                  argument_parser_compatible_option<std::ranges::range_value_t<option_type>>) &&
                   std::invocable<validator_type, option_type>
     //!\endcond
     void add_option(option_type & value,
@@ -293,7 +293,7 @@ public:
     template <typename option_type, validator validator_type = detail::default_validator<option_type>>
     //!\cond
         requires (argument_parser_compatible_option<option_type> ||
-                  argument_parser_compatible_option<typename option_type::value_type>) &&
+                  argument_parser_compatible_option<std::ranges::range_value_t<option_type>>) &&
                   std::invocable<validator_type, option_type>
     //!\endcond
     void add_positional_option(option_type & value,
--- a/include/seqan3/argument_parser/validators.hpp
+++ b/include/seqan3/argument_parser/validators.hpp
@@ -51,7 +51,7 @@ namespace seqan3
  * \brief You can expect these (meta-)functions on all types that implement seqan3::validator.
  * \{
  */
-/*!\typedef     using value_type
+/*!\typedef     using option_value_type
  * \brief       The type of value on which the validator is called on.
  * \relates     seqan3::validator
  *
@@ -59,9 +59,9 @@ namespace seqan3
  * \attention This is a concept requirement, not an actual typedef (however types satisfying this concept
  * will provide an implementation).
  */
-/*!\fn              void operator()(value_type const & cmp) const
+/*!\fn              void operator()(option_value_type const & cmp) const
  * \brief           Validates the value 'cmp' and throws a seqan3::validation_failed on failure.
- * \tparam          value_type The type of the value to be validated.
+ * \tparam          option_value_type The type of the value to be validated.
  * \param[in,out]   cmp The value to be validated.
  * \relates         seqan3::validator
  * \throws          seqan3::validation_failed if value 'cmp' does not pass validation.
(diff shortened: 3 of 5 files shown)


include/seqan3/argument_parser/validators.hpp
```diff
@@ -83,12 +83,11 @@ namespace seqan3
...
 template <typename validator_type>
 SEQAN3_CONCEPT validator = std::copyable<remove_cvref_t<validator_type>> &&
-                           requires(validator_type validator,
-                                    typename std::remove_reference_t<validator_type>::value_type value)
+                           requires(validator_type validator)
 {
     typename std::remove_reference_t<validator_type>::value_type;
 
-    { validator(value) } -> void;
+    { validator(typename std::remove_reference_t<validator_type>::value_type{}) } -> void;
```
smehringer: I would use decval in case the value type is not default constructible?

h-2: Yes, but it doesn't wark because GCC7 and GCC8 at least trigger ODR-use of that statement (although they shouldn't) which results in the compiler telling you that you cannot use declval in an evaluated context. :man_shrugging:

h-2:
(do not merge yet, still WIP)

h-2:
Ok, a different "fix" was necessary, because the problem is that the concept check of GCC doesn't cope with something existing, but being private/protected. It will result in hard errors instead of failing the concept.

But since `::value_type` is taken as an indicator for iterators and ranges it's not the best choice anyway. I have just renamed it here which solves all the problems. Is that ok with you, @smehringer ?

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -293,7 +293,7 @@ class argument_parser
     template <typename option_type, validator validator_type = detail::default_validator<option_type>>
     //!\cond
         requires (argument_parser_compatible_option<option_type> ||
-                  argument_parser_compatible_option<typename option_type::value_type>) &&
+                  argument_parser_compatible_option<std::ranges::range_value_t<option_type>>) &&
```
h-2: this is just a drive-by fix

include/seqan3/argument_parser/validators.hpp
smehringer: :expressionless: really...

smehringer:
fine with me. Good thing we are not stable yet...

smehringer:
CI is failing though

CHANGELOG.md
```diff
@@ -48,6 +48,8 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 * **Changed class signature of input/output file validators:**
   Most user code will be unaffected; to fix possible compiler errors you need to add an empty template parameter list to
   the respective instances (e.g. change `input_file_validator` to `input_file_validator<>`).
+* The member type that denotes which arguments a `validator` can validate has been renamed from `value_type` to
```
rrahn: a) why and b) what has this to do with the hard fails on concept checks?

marehr:
https://travis-ci.org/seqan/seqan3/jobs/620336210#L5884-L5962

CHANGELOG.md
h-2: See above

> Ok, a different "fix" was necessary, because the problem is that the concept check of GCC doesn't cope with something existing, but being private/protected. It will result in hard errors instead of failing the concept.

h-2: (our ranges often have protected or private `::value_type`)

h-2:
> https://travis-ci.org/seqan/seqan3/jobs/620336210#L5884-L5962

I know, I hope to have time for it tomorrow.

include/seqan3/argument_parser/validators.hpp
rrahn: yeah I've seen this as well, which is annoying.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -97,7 +97,7 @@ SEQAN3_CONCEPT validator = std::copyable<remove_cvref_t<validator_type>> &&
  * \ingroup argument_parser
  * \implements seqan3::validator
  *
- * \tparam option_value_type Must be a (container of) arithmetic type(s).
+ * \tparam option_input_type Must be a (container of) arithmetic type(s).
```
rrahn: Sorry, but this class has no template parameter

include/seqan3/argument_parser/validators.hpp
```diff
@@ -176,12 +176,12 @@ class arithmetic_range_validator
  *
  * \include test/snippet/argument_parser/validators_2.cpp
  */
-template <typename option_value_type>
+template <typename option_input_type>
```
rrahn: here tparam is not documented 🤕

CHANGELOG.md
```diff
@@ -48,6 +48,8 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
...
+  `input_type`.
```
rrahn: I guess calling it `option_value_type` would be more fitting. At the end a validator validates the value of an option. Accordingly, we want to know the value type an option represents.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -896,14 +896,14 @@ namespace detail
  * The default validator is needed to make the validator parameter of
  * argument_parser::add_option and argument_parser::add_option optional.
  */
-template <typename option_value_type>
+template <typename option_input_type>
```
rrahn: tparam not documented

include/seqan3/argument_parser/validators.hpp
rrahn: This would then be `option_value_t` to not clash with the typedef inside.

include/seqan3/argument_parser/validators.hpp
rrahn: it would be then as well `option_value_t`

PR #1395, opened 2019-12-03
Alphabet stuff episode 100

h-2:
* Splits the alpahbet benchmarks and add more alphabets.
* Optimises some composite alphabets significantly.
* For unclear reasons, compiler workarounds for alphabet_tuple_base can now be removed again :man_shrugging: (still required for alphabet_variant, I double-checked)

review by commit recommended.


include/seqan3/alphabet/composite/semialphabet_any.hpp
```diff
@@ -79,7 +79,7 @@ class semialphabet_any : public alphabet_base<semialphabet_any<size>, size, void
     //!\brief Enable conversion of semialphabet_any into other (semi-)alphabet of the same size
     template <semialphabet other_alph_t>
     //!\cond
-        requires (alphabet_size<other_alph_t> == size)
+        requires ((alphabet_size<other_alph_t> == size) && std::regular<other_alph_t>)
```
eseiler: Do you need only default-constructibility, or also equality comparison and copiability (=regular)? 
So either the requirement or the commit message is not correct :)

test/performance/CMakeLists.txt
```diff
@@ -15,8 +15,14 @@ macro (seqan3_benchmark benchmark_cpp)
     seqan3_test_component (target "${benchmark}" TARGET_NAME)
     seqan3_test_component (test_name "${benchmark}" TEST_NAME)
 
+    # force alignment of benchmarked loops so that numbers are reliable:
```
eseiler: We don't have `[BENCH]`, only `[TEST]` in the commit message

eseiler: Maybe you can add another sentence as to when to disable it?
I remember we discussed that this may not help/ is disruptive when the loops are big?
This could go here, or even just that when benchmarks are weird this could be a option to try to disable?

test/performance/alphabet/alphabet_assign_char_benchmark.cpp
```diff
@@ -0,0 +1,64 @@
...
+// -----------------------------------------------------------------------------------------------------
+
+#include <algorithm>
+#include <cstring>
+
+#include <benchmark/benchmark.h>
+
+#include <seqan3/alphabet/all.hpp>
+
+using namespace seqan3;
```
eseiler: Mhm, don't we want to get rid of it in _every_ test?

include/seqan3/alphabet/composite/semialphabet_any.hpp
h-2: `semialphabet` already includes everything from `std::regular` other than default-constructibility.

test/performance/alphabet/alphabet_assign_char_benchmark.cpp
h-2: Well, I don't know. My assumption was:
* unit test ensure that our interfaces can be used in the advertised way and that we have code coverage. Since we advertise using all things qualified, that's what we should test.
* performance tests highlight performance issues and regressions. I don't see how calling qualified/unqualified has an impact on this, so I left it this way.

test/performance/CMakeLists.txt
h-2: :+1:

test/performance/alphabet/alphabet_assign_char_benchmark.cpp
eseiler: So everywhere **except** in performance tests.
Then we should write this down more clearly somewhere.

h-2: Yes. I mean I don't really care. We can also say that we do it everywhere for consistency.

test/performance/CMakeLists.txt
h-2: I addressed this.

h-2: I addressed this.

test/performance/alphabet/alphabet_assign_char_benchmark.cpp
h-2: Ok, I have made everything qualified.

include/seqan3/alphabet/mask/masked.hpp
```diff
@@ -100,39 +92,48 @@ class masked : public alphabet_tuple_base<masked<sequence_alphabet_t>, sequence_
...
-        return masked{}.assign_char(c).to_char() == c;
-    }
+        [] ()
+        {
+            std::array<char_type, alphabet_size> ret{};
+
+            for (size_t i = 0; i < alphabet_size; ++i)
+            {
+                ret[i] = (i < alphabet_size / 2)
+                       ? seqan3::to_char(seqan3::assign_rank_to(i, sequence_alphabet_type{}))
```
smehringer: 1. We should require that `sequence_alphabet_t` is default constructible then (in the template argument)
2. Why is the name `sequence_alphabet_t` and not just `alphabet_t`? Any special reason or just legacy code?

test/performance/alphabet/alphabet_assign_char_benchmark.cpp
```diff
@@ -0,0 +1,63 @@
...
+        r = i++;
+
+    alphabet_t a{};
+    for (auto _ : state)
+    {
+        for (char_t c : chars)
+        {
+            benchmark::DoNotOptimize
+            (
+                seqan3::assign_char_to(c, a)
```
smehringer: put into one line? :nail_care:

test/performance/alphabet/alphabet_assign_char_benchmark.cpp
```diff
@@ -0,0 +1,63 @@
...
+BENCHMARK_TEMPLATE(assign_char_, seqan3::aa27);
+BENCHMARK_TEMPLATE(assign_char_, seqan3::phred42);
+BENCHMARK_TEMPLATE(assign_char_, seqan3::phred63);
+/* adaptations */
+BENCHMARK_TEMPLATE(assign_char_, char);
+BENCHMARK_TEMPLATE(assign_char_, char32_t);
+/* alphabet variant */
+BENCHMARK_TEMPLATE(assign_char_, seqan3::gapped<seqan3::dna4>);
+BENCHMARK_TEMPLATE(assign_char_, seqan3::alphabet_variant<seqan3::gap, seqan3::dna4, seqan3::dna5, seqan3::dna15,
+                                                      seqan3::rna15, seqan3::rna4, seqan3::rna5>);
```
smehringer:
```suggestion
                                                          seqan3::rna15, seqan3::rna4, seqan3::rna5>);
```

include/seqan3/alphabet/mask/masked.hpp
h-2: > We should require that sequence_alphabet_t is default constructible then (in the template argument)

alphabet_tuple_base enforces this kind of (! reference, but it should be "semiregular").

h-2: > Why is the name sequence_alphabet_t and not just alphabet_t? Any special reason or just legacy code?

There are two alphabets. The mask alphabet and the sequence_alphabet.

PR #1398, opened 2019-12-04
[I/O] Remove column interface for structure files

### LINKED ISSUE #1397: [Structure IO] Remove file column interface
**Level 1 - Easy**

We expose a broken file column interface for our files. Read the documentation of `structure_file_input` to get to know the interface.

- [x] Remove all tests for the column interface (usually have `column` in their test name)
- [x] Remove the `Tuple interface` from the `sequence_file_input` object (all get functions)
- [x] Remove the `file_as_tuple_type columns_buffer;`
- [x] Remove the member function `read_columns`
- [x] Remove the std-overloads for the tuple like interface
- [x] Remove all `Field column types and tuple type` (see group of types in the beginning)
- [x] Remove the column based reading from the documentation of structure_file_input. Check if it is mentioned anywhere else in the documentation!
- [x] Remove all container_container concepts from the `structure_file_input_traits` and remove the corresponding documentation
- [x] Remove the container_container types from structure_file_input_default_traits_rna/aa

joergi-w:
Resolves #1397.


joergi-w:
Thank you!

joergi-w:
> With the code you removed you know cannot do fout = fin anymore can you? Is this tested? (It is for sequence files)

Yes, `fout = fin` is still possible and tested here: https://github.com/seqan/seqan3/blob/cfcc96ee77847e043ac2432d9cdeaf5d7a15935f/test/unit/io/structure_file/structure_file_output_test.cpp#L378-L380
https://github.com/seqan/seqan3/blob/cfcc96ee77847e043ac2432d9cdeaf5d7a15935f/test/unit/io/structure_file/structure_file_output_test.cpp#L324-L331

rrahn:
>I think the output must not be changed! Only the get interface of input files because it is broken. @seqan/core am I right? 

Well, the output files don't have a get interface so that shouldn't be a problem. 
However, I am not completely convinced that the additional tuple/record based interface adds a lot of value. I am certainly not convinced that we have a huge performance gain here, because we save one pointer indirection when writing the records. So I wouldn't mind removing it as well?

I mean it would essentially boil down to the decision of whether 1 is so much better than 2. 
```cpp
/*1*/ fout = std::tie(c1, c2, c3));
...
/*2*/ std::ranges::move(views::zip(c1, c2, c3), std::back_inserter(fout));
```
Note, that 1 might not be supported by all file types. Can  be done very wrong by doing:
```
fout = std::make_tuple(c1, c2, c3));
```
The difference is subtle and when doing wrong would result in copying the entire data into the tuple before writing it to the file. But also I haven't used either interface in a use case so there might be other valid points I am not considering.

> With the code you removed you know cannot do fout = fin anymore can you? Is this tested? (It is for sequence files)

Why not? But I am very certain that there is at least one snippet that should fail if this is not working anymore?!

h-2:
I agree the functions should be removed. You can still use assignment by range interface and do

```cpp
fout = views::zip(c1, c2, c3);
```

No moving and back-inserters required :)

smehringer:
I agree, we should remove the interface then. And we should add a test for the example @h-2 gave :)

CHANGELOG.md
```diff
@@ -58,6 +58,11 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 * **The `type_list` header has moved:**
   If you included `<seqan3/core/type_list.hpp>` you need to change the path to `<seqan3/core/type_list/type_list.hpp>`.
 
+#### Input/Output
```
smehringer:
```suggestion
#### I/O
```
:nail_care: Since this is the actual directory name ?

joergi-w: There was already a section `Input/Output` in the file above so I wanted to keep it consistent. I changed both to `I/O` now.

joergi-w:
> And can you merge the changelog commit into the actual commit that introduced the change? such that those appear together :)

I made two commits: 
1. removes the input interface and adds the changelog entry
2. removes the output interface as well as documents and tests the zip iterator version

rrahn:
> No moving and back-inserters required :)

Even better.

PR #1400, opened 2019-12-05
Misc/remove char type from file

### LINKED ISSUE #1399: [IO] Remove the char_type template parameter
**Level 1 - Intermediate**

Files are not properly tested with other stream_char_types than `char`. Formats currently assume that the stream_char_type is at least convertible to/from `char`, but some things are definitely broken when the type is not `char`.

Remove the ability to specify a stream char type other than char.

- [ ] Adapt tests
- [ ] Remove char_type template arguments from all input and output interfaces and hard code it to `char`

rrahn:
Fixes #1399 

Removes the char type from the files.

It also fixes the file exposer which caused hard fails in the format concept checks due to the using statement to import the read/write interfaces for the different file formats. The problem here is that for types that are not formats the using declaration causes compiler errors, which are different from not finding an viable overload. This was first visible when the char type was removed. Also I am not quite sure why?
Strangely, the exposer were not documented nor excluded from documentation but doxygen didn't trigger a warning. Not sure why? But I added the documentation for them as well.

A changelog will be added soon as well.
I also need to check at which places the documentation explicitly refers to the char type. There were three snippets that used it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,9 +62,16 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 
 #### I/O
 
-* The field-based in- and output interface for structure files through std::get and std::tie has been removed.
+* **Removed the field-based in- and output interface for structure files through std::get and std::tie:**
   Output can instead be achieved with seqan3::views:zip(), for input we will implement unzip() in the future.
 * The `field::FLAG` of SAM/BAM input and output is now an **enum** instead of a simple integer (see seqan3::sam_flag).
+
+* **Removed the char type from the input and output files:**
+  Most user code will be unaffected; however, if you have fully specified all templates of any of the input or output
+  files in your code, you need to remove the template parameter to select the char type of the stream
+  (e.g. change `seqan3::sequence_file_input<traits_t, fields_t, formats_t, char>` to
+  `seqan3::sequence_file_input<traits_t, fields_t, formats_t>`). Before this change, setting the char type gave the
+  impression that also streams over wide characters are supported which is not the case yet.
 
 #### Range
 
--- a/include/seqan3/io/alignment_file/input_format_concept.hpp
+++ b/include/seqan3/io/alignment_file/input_format_concept.hpp
@@ -31,11 +31,28 @@
 namespace seqan3::detail
 {
 
+/*!\brief Internal class used to expose the actual format interface to read alignment records from the file.
+ * \ingroup alignment_file
+ *
+ * \tparam format_type The type of the format to be exposed.
+ *
+ * \details
+ *
+ * Exposes the protected member function `read_alignment_record` from the given `format_type`, such that the file can
+ * call the proper function for the selected format.
+ */
 template <typename format_type>
 struct alignment_file_input_format_exposer : public format_type
 {
 public:
-    using format_type::read_alignment_record;
+    // Can't use `using format_type::read_alignment_record` as it produces a hard failure in the format concept check
+    // for types that do not model the format concept, i.e. don't offer the proper read_alignment_record interface.
+    //!\brief Forwards to the seqan3::alignment_file_input_format::read_alignment_record interface.
+    template <typename ...ts>
+    void read_alignment_record(ts && ...args)
+    {
+        format_type::read_alignment_record(std::forward<ts>(args)...);
+    }
 };
 
 } // namespace seqan3::detail
--- a/include/seqan3/io/alignment_file/output.hpp
+++ b/include/seqan3/io/alignment_file/output.hpp
@@ -49,7 +49,6 @@ namespace seqan3
  *                              fields IDs; only relevant if these can't be deduced.
  * \tparam valid_formats        A seqan3::type_list of the selectable formats (each
  *                              must model seqan3::alignment_file_output_format).
- * \tparam stream_char_type     The type of character of the underlying stream, must model seqan3::builtin_character.
  *
  * \details
  *
@@ -183,7 +182,6 @@ template <detail::fields_specialisation selected_field_ids_ =
                      field::BIT_SCORE,
                      field::HEADER_PTR>,
           detail::type_list_of_alignment_file_output_formats valid_formats_ = type_list<format_sam, format_bam>,
-          builtin_character stream_char_type_ = char,
           typename ref_ids_type = ref_info_not_given>
 class alignment_file_output
 {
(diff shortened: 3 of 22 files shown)


include/seqan3/io/sequence_file/input_format_concept.hpp
```diff
@@ -27,10 +27,36 @@ namespace seqan3::detail
...
+ *
+ * Exposes the protected member function `read_sequence_record` from the given `format_type`, such that the file can
+ * call the proper function for the selected format.
+ */
 struct sequence_file_input_format_exposer : public format_type
 {
 public:
-    using format_type::read_sequence_record;
+
+    // Can't use `using format_type::read_alignment_record` as it produces a hard failure in the format concept check
```
joergi-w:
```suggestion
    // Can't use `using format_type::read_sequence_record` as it produces a hard failure in the format concept check
```

include/seqan3/io/sequence_file/input_format_concept.hpp
```diff
@@ -27,10 +27,36 @@ namespace seqan3::detail
...
+    // for types that do not model the format concept, i.e. don't offer the proper read_alignment_record interface.
```
joergi-w:
```suggestion
    // for types that do not model the format concept, i.e. don't offer the proper read_sequence_record interface.
```

include/seqan3/io/sequence_file/output_format_concept.hpp
```diff
@@ -25,11 +25,40 @@
 namespace seqan3::detail
 {
 
+/*!\brief Internal class used to expose the actual format interface to write sequence records within the file.
```
joergi-w: within the file => to/into the file? 
(also in alignment and structure file output concepts)

joergi-w: Nice! You also changed the input concepts :)

CHANGELOG.md
```diff
@@ -60,9 +60,16 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
...
-* The field-based in- and output interface for structure files through std::get and std::tie has been removed.
+* **Removed the field-based in- and output interface for structure files through std::get and std::tie:**
   Output can instead be achieved with seqan3::views:zip(), for input we will implement unzip() in the future.
 
+* **Removed the char type from the input and output files:**
+  Most user code will be unaffected; however, if you have fully specified all templates of any of the input or output
+  files in your code, you need to remove the template parameter to select the char type of the stream
+  (e.g. change `seqan3::sequence_file_input<traits_t, fields_t, formats_t, char>` to
+  `seqan3::sequence_file_input<traits_t, fields_t, formats_t>`). Before this change, setting the char type gave the
+  impression that also streams over wide characters are supported which is not the case yet.
```
smehringer:
```suggestion
  impression that streams over wide characters are supported which is not the case yet.
```

include/seqan3/io/alignment_file/input_format_concept.hpp
```diff
@@ -30,11 +30,82 @@
...
+ *
+ * Exposes the protected member function `read_alignment_record` from the given `format_type`, such that the file can
+ * call the proper function for the selected format.
+ */
 template <typename format_type>
 struct alignment_file_input_format_exposer : public format_type
 {
 public:
-    using format_type::read_alignment_record;
+    // Can't use `using format_type::read_alignment_record` as it produces a hard failure in the format concept check
```
smehringer: phew, this is ugly.

Can you try to revert this and instead change the concept below to the following:

```diff
SEQAN3_CONCEPT alignment_file_input_format =
-   requires (detail::alignment_file_input_format_exposer<t>                      & v,
-             std::ifstream                                                       & stream,
+   requires (std::ifstream                                                       & stream,
              alignment_file_input_options<dna5>                                  & options,
              std::vector<dna5_vector>                                            & ref_sequences,
              alignment_file_header<>                                             & header,
              dna5_vector                                                         & seq,
              std::vector<phred42>                                                & qual,
              std::string                                                         & id,
              int32_t                                                             & offset,
              dna5_vector                                                         & ref_seq,
              std::optional<int32_t>                                              & ref_id,
              std::optional<int32_t>                                              & ref_offset,
              std::pair<std::vector<gapped<dna4>>, std::vector<gapped<dna4>>>     & align,
              std::vector<cigar>                                                  & cigar,
              uint16_t                                                            & flag,
              uint8_t                                                             & mapq,
              std::tuple<std::optional<int32_t>, std::optional<int32_t>, int32_t> & mate,
              sam_tag_dictionary                                                  & tag_dict,
              double                                                              & e_value,
              double                                                              & bit_score)
{
    t::file_extensions;
+   { detail::alignment_file_input_format_exposer<t>; }
    // std::same_as<decltype(t::file_extensions), std::vector<std::string>>;
-   { v.read_alignment_record(stream,
+   {detail::alignment_file_input_format_exposer<t>{}.read_alignment_record(stream,
                              options,
                              ref_sequences,

```

rrahn: That won't work either. As I described in the comment the concept check hard fails if the using declaration fails. That is not an instantiation error as you would get with a function template because the using declaration is not used as a SFINAE expression. It doesn't matter were you put it into the concept. So, yes it is very ugly but I see no easy solution for it. A simpler interface would make it more bearable but that is another story. I could try to use some fancy SFINAE stuff inside of the exposer, but that won't make it more readable nor am I sure that it will work. The problem here is that the member function is protected and can't be used in the top-level but only within the exposer type.

include/seqan3/io/alignment_file/input_format_concept.hpp
```diff
@@ -31,11 +31,82 @@
...
+              typename align_type,
+              typename cigar_type,
+              typename flag_type,
+              typename mapq_type,
+              typename qual_type,
+              typename mate_type,
+              typename tag_dict_type,
+              typename e_value_type,
+              typename bit_score_type>
+    void read_alignment_record(stream_type & stream,
```
smehringer: Can you replace this at least with

```cpp
template <typename arg_types...>
void read_alignment_record(arg_types ...args)
{ // ...
}
``` 
? We do enough concept checks to enforce the proper parameters?

include/seqan3/io/alignment_file/output.hpp
```diff
@@ -829,61 +833,56 @@ class alignment_file_output
 /*!\brief Deduces selected_field_ids from input and sets alignment_file_output::ref_ids_type to
  * seqan3::detail::ref_info_not_given. Valid formats and stream_char_type are defaulted.
```
smehringer: You forgot to adapt the documentation in this file

rrahn: good catch

smehringer:
travis still fails

rrahn:
> travis still fails

Documentation was broken due to the last changes. Is not fixed. I'll merge when it run through

PR #1402, opened 2019-12-06
[DOC] Add more tests for global alignment.

rrahn:
* Adds more tests for the global alignment. These tests will be reused to test collections with parallel and vectorised execution.
* Adds tests for corner cases, e.g. aligning against empty sequences.
* Renamed the test fixtures to better compare them.
* Removed the global using declaration for seqan3 namespaces.

### DIFF
--- a/test/unit/alignment/pairwise/fixture/global_affine_unbanded.hpp
+++ b/test/unit/alignment/pairwise/fixture/global_affine_unbanded.hpp
@@ -24,15 +24,22 @@
 namespace seqan3::test::alignment::fixture::global::affine::unbanded
 {
 
-using namespace seqan3;
-using namespace seqan3::detail;
+inline constexpr auto align_config = seqan3::align_cfg::mode{seqan3::global_alignment}
+                                   | seqan3::align_cfg::gap{seqan3::gap_scheme{seqan3::gap_score{-1},
+                                                            seqan3::gap_open_score{-10}}};
+inline constexpr auto align_config_dna_score = align_config
+                                             | seqan3::align_cfg::scoring{
+                                                    seqan3::nucleotide_scoring_scheme{seqan3::match_score{4},
+                                                                                      seqan3::mismatch_score{-5}}};
 
-inline constexpr auto align_config = align_cfg::mode{global_alignment} |
-                                     align_cfg::gap{gap_scheme{gap_score{-1}, gap_open_score{-10}}};
-
-static auto dna4_01 = []()
+// Naming is used to disambiguate configs for global alignment with different scoring parameters.
+// `part_XX` indicates that the config is the same but the sequences differ in their lengths.
+// This is helpful to collect different test fixtures to test vectorised alignments where we can
+// only apply alignments with the same config.
+static auto dna4_match_4_mismatch_5_gap_1_open_10_part_01 = []()
 {
-    return alignment_fixture
+    using seqan3::operator""_dna4;
+    return seqan3::test::alignment::fixture::alignment_fixture
     {
         // score: 8 (7 insertions, 1 substitutions)
         // alignment:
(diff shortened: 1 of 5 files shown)


test/unit/alignment/pairwise/fixture/global_affine_unbanded.hpp
```diff
@@ -135,4 +135,255 @@ static auto dna4_02 = []()
...
+        /*A*/  u  ,DuL,
+        /*G*/  u  ,uL ,
+        /*C*/  u  ,uL ,
+        /*G*/  u  ,uL
+        }
+    };
+}();
+
+
+static auto dna4_match_4_mismatch_5_gap_1_open_10_seq1_empty_part_06 = []()
```
eseiler: Do you want to have the `part_06` part here (and the next two)? The `seq_empty` makes it already different from the others before.

test/unit/alignment/pairwise/pairwise_alignment_single_test_template.hpp
```diff
@@ -15,14 +15,12 @@
 
 #include "fixture/alignment_fixture.hpp"
 
-using namespace seqan3;
-using namespace seqan3::detail;
-using namespace seqan3::test::alignment::fixture;
+// using namespace seqan3::test::alignment::fixture;
```
eseiler: remove

test/unit/alignment/pairwise/pairwise_alignment_single_test_template.hpp
```diff
@@ -36,6 +34,8 @@ TYPED_TEST_CASE_P(pairwise_alignment_test);
 
 TYPED_TEST_P(pairwise_alignment_test, score)
 {
+    using namespace seqan3;
```
eseiler: remove and do qualified calls

test/unit/alignment/pairwise/fixture/global_affine_unbanded.hpp
rrahn: probably not. I did wanted to know that they are special because of empty sequences.

test/unit/alignment/pairwise/pairwise_alignment_single_test_template.hpp
rrahn: 😭

test/unit/alignment/pairwise/fixture/global_affine_unbanded.hpp
eseiler: Do as you like

test/unit/alignment/pairwise/pairwise_alignment_single_test_template.hpp
eseiler: 😈

test/unit/alignment/pairwise/fixture/global_affine_unbanded.hpp
```diff
@@ -24,15 +24,18 @@
...
+inline constexpr auto align_config_dna_score = align_config
+                                             | seqan3::align_cfg::scoring{
+                                                    seqan3::nucleotide_scoring_scheme{seqan3::match_score{4},
+                                                                                      seqan3::mismatch_score{-5}}};
 
-inline constexpr auto align_config = align_cfg::mode{global_alignment} |
-                                     align_cfg::gap{gap_scheme{gap_score{-1}, gap_open_score{-10}}};
-
-static auto dna4_01 = []()
+static auto dna4_match_4_mismatch_5_gap_1_open_10_part_01 = []()
```
smehringer: Question: why is it neccessary to include the paramters in the name? It feels like this could lead to inconsistencies in the future if you change some parameters (especially since they are at the top not inside of the lambda)

EDIT: also if you are this expressive it is weird that part 01-05 are not expressive in a sense that I do not now what is testing what? I see that all tests here have the same scoring scheme anyway. It feels more helpful if the names would be `dna4_default_scoring_scheme_score_only` and then rename your `align_config_dna_score` to `align_config_dna_default_score` or something similar.

test/unit/alignment/pairwise/fixture/global_affine_unbanded.hpp
```diff
@@ -135,4 +139,261 @@ static auto dna4_02 = []()
...
+        /*C*/  u  ,uL ,
+        /*T*/  u  ,uL ,
+        /*A*/  u  ,DuL,
+        /*G*/  u  ,uL ,
+        /*C*/  u  ,uL ,
+        /*G*/  u  ,uL
+        }
+    };
+}();
+
```
smehringer: remove extra newline :nail_care:

test/unit/alignment/pairwise/fixture/global_affine_unbanded.hpp
rrahn: Because I want to identify tests with the same configurations, which is necessary for testing vectorised alignments. 
So there might be added tests with different scores that cannot be put into the same vectorisation code - part01 - 05 indicates that they have the same configuration but different sequences. Not sure why this would be an issue?

PR #1404, opened 2019-12-06
[MISC] make simd_backend a template template in simd_type defintion.

marehr:
It makes more sense to give the backend without any template parameters, because they are independent of them.

This PR affects no tests, because we don't test the backend feature, yet.

### DIFF
--- a/include/seqan3/core/simd/simd.hpp
+++ b/include/seqan3/core/simd/simd.hpp
@@ -44,18 +44,18 @@ inline namespace simd
  */
 template <typename scalar_t,
           size_t length = detail::default_simd_length<scalar_t, detail::default_simd_backend>,
-          typename simd_backend = detail::default_simd_backend<scalar_t, length>>
-struct simd_type : simd_backend
+          template <typename scalar_t_, size_t length_> typename simd_backend = detail::default_simd_backend>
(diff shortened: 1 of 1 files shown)


include/seqan3/core/simd/simd.hpp
```diff
@@ -44,18 +44,18 @@ inline namespace simd
  */
 template <typename scalar_t,
           size_t length = detail::default_simd_length<scalar_t, detail::default_simd_backend>,
-          typename simd_backend = detail::default_simd_backend<scalar_t, length>>
-struct simd_type : simd_backend
+          template <typename scalar_t_, size_t length_> typename simd_backend = detail::default_simd_backend>
```
joergi-w: The template types `scalar_t_` and `length_` are not used, cannot you just omit them? 
```cpp
template <typename, size_t>
```
Or did you forget to use them as parameters for `simd_backend`?

marehr: you are right they are completely unused, but @rrahn said to me in a different PR that I should always name them :man_shrugging:

joergi-w: Why do we need a template then, if its types are not used?
I'm sorry but I don't understand the benefit of this change.

PR #1406, opened 2019-12-06
[TEST] fix some issues with simd tests on other platforms

marehr:
fix some issues with simd tests on other platforms (see #1381) and I also removed the namespace seqan3.

### DIFF
--- a/test/unit/core/simd/simd_test.cpp
+++ b/test/unit/core/simd/simd_test.cpp
@@ -9,44 +9,44 @@
 
 #include <seqan3/core/simd/all.hpp>
 
-using namespace seqan3;
-
-using int16x_t = seqan3::simd::simd_type<int16_t>::type;
-
 TEST(simd, auto_length)
 {
-    if constexpr(simd_traits<int16x_t>::max_length == 64u)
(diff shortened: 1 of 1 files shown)


test/unit/core/simd/simd_test.cpp
```diff
@@ -9,44 +9,44 @@
...
 #include <seqan3/core/simd/all.hpp>
 
-using namespace seqan3;
-
-using int16x_t = seqan3::simd::simd_type<int16_t>::type;
-
 TEST(simd, auto_length)
 {
-    if constexpr(simd_traits<int16x_t>::max_length == 64)
+    using int16x_t = seqan3::simd::simd_type<int16_t>::type;
```
rrahn: can you not use using seqan3::simd within the test body? It spoils the entire test. Or is this a general problem?

marehr: you mean `using namespace seqan3::simd`? sure I can xD

marehr: We fully specify everything in tests now :)

rrahn: 😭

PR #1407, opened 2019-12-06
[MISC] Do not use x86 intrinsics if not available.

marehr:
This fixes the platform issues when x86 intrinsics (#1381) are not available, like on arm or powerpc.

The main idea is to always forward declare each simd_algorithm for each instruction set, e.g. load_sse4, load_avx2, load_avx512.

If the instruction set is available we define the forward declared algorithm, e.g. load_avx2.

Note: We forward declare for example load_avx2 s.t. we can use the name load_avx2 inside of the simd::load function regardless of whether avx2 intrinsics are available or not. We guarantee in simd::load that load_avx2 will only be invoked if the instruction set provides avx2 or higher (via constexpr, but each constexpr branch needs to contain valid c++ code).

### DIFF
--- a/include/seqan3/core/simd/detail/builtin_simd_intrinsics.hpp
+++ b/include/seqan3/core/simd/detail/builtin_simd_intrinsics.hpp
@@ -1,0 +1,29 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides intrinsics include for builtin simd.
+ * \author Marcel Ehrhardt <marcel.ehrhardt AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <seqan3/core/platform.hpp>
+
+#if __has_include(<x86intrin.h>)
+#include <x86intrin.h> // x86 intrinsics (linux)
+#endif
+
+#if __has_include(<intrin.h>)
+#include <intrin.h> // x86 intrinsics (windows)
+#endif
+
+// MSVC doesn't define SSE4 macros, even if the instruction set is available (e.g. when AVX is defined)
+#if defined(_MSC_VER) && defined(__AVX__) && !defined(__SSE4_1__) && !defined(__SSE4_2__)
+#define __SSE4_1__ 1
+#define __SSE4_2__ 1
+#endif
--- a/include/seqan3/core/simd/detail/simd_algorithm_avx2.hpp
+++ b/include/seqan3/core/simd/detail/simd_algorithm_avx2.hpp
@@ -1,0 +1,156 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides specific algorithm implementations for AVX2 instruction set.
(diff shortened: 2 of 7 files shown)


include/seqan3/core/simd/detail/simd_algorithm_avx2.hpp
```diff
@@ -0,0 +1,152 @@
...
+template <simd::simd_concept target_simd_t, simd::simd_concept source_simd_t>
+constexpr target_simd_t upcast_signed_avx2(source_simd_t const & src);
+
+/*!\copydoc seqan3::detail::upcast_unsigned
+ * \attention This is the implementation for AVX2 intrinsics.
+ */
+template <simd::simd_concept target_simd_t, simd::simd_concept source_simd_t>
+constexpr target_simd_t upcast_unsigned_avx2(source_simd_t const & src);
+
+/*!\copydoc seqan3::detail::extract_halve
```
eseiler:
```suggestion
/*!\copydoc seqan3::detail::extract_half
```

At some point it should be renamed...

marehr: Next PR after this will rename it.

include/seqan3/core/simd/detail/builtin_simd_intrinsics.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+#if __has_include(<x86intrin.h>)
+#include <x86intrin.h> // x86 intrinsics (linux)
+#endif
+
+#if __has_include(<intrin.h>)
+#include <intrin.h> // x86 intrinsics (windows)
+#endif
+
+#if defined(_MSC_VER) && defined(__AVX__) && !defined(__SSE4_1__) && !defined(__SSE4_2__)
```
rrahn: Please add a comment that windows might not define SSE if avx is available and that we need this to allow using sse also if avx is supported.

include/seqan3/core/simd/detail/simd_algorithm_avx2.hpp
```diff
@@ -0,0 +1,152 @@
...
+{
+
+template <simd::simd_concept simd_t>
+constexpr simd_t load_avx2(void const * mem_addr)
+{
+    return reinterpret_cast<simd_t>(_mm256_loadu_si256(reinterpret_cast<__m256i const *>(mem_addr)));
+}
+
+template <simd::simd_concept simd_t>
+inline void transpose_matrix_avx2(std::array<simd_t, simd_traits<simd_t>::length> & matrix);
```
rrahn: So you declare it here again? Why is that?

include/seqan3/core/simd/detail/simd_algorithm_avx2.hpp
```diff
@@ -0,0 +1,152 @@
...
+    }
+    else // cast from epi32 to epi64
+    {
+        static_assert(simd_traits<source_simd_t>::length == 8, "Expected 32 bit scalar type.");
+        return reinterpret_cast<target_simd_t>(_mm256_cvtepu32_epi64(tmp));
+    }
+}
+
+template <uint8_t index, simd::simd_concept simd_t>
+constexpr simd_t extract_halve_avx2(simd_t const & src);
```
rrahn: same here. Only declared but not defined?

include/seqan3/core/simd/detail/simd_algorithm_avx512.hpp
```diff
@@ -0,0 +1,152 @@
...
+#include <seqan3/core/simd/simd_traits.hpp>
+
+//-----------------------------------------------------------------------------
+// forward declare avx512 simd algorithms that use avx512 intrinsics
+//-----------------------------------------------------------------------------
+
+namespace seqan3::detail
+{
+/*!\copydoc seqan3::simd::load
+ * \attention This is the implementation for AVX512 intrinsics.
```
rrahn:
```suggestion
 * \remark This is the implementation for AVX512 intrinsics.
```
and all other places in all other files as well, pleae

include/seqan3/core/simd/detail/simd_algorithm_avx2.hpp
rrahn: can you make a comment that this needs to be implemented here?

include/seqan3/core/simd/detail/simd_algorithm_avx512.hpp
marehr: I forgot what we said here, but this should be `\attention`.

rrahn:
@marehr looks good! You can squash now the last two commits.

mr-c:
Personally, I think this is the last PR on my wish list for a January release of Seqan3 3.0.1 :-)

marehr:
@rrahn done please merge

PR #1408, opened 2019-12-06
Misc/remove reflection

rrahn:
Removes the old mechanism to extract the name of a type by a more stable situation using typeid.


include/seqan3/argument_parser/detail/format_help.hpp
```diff
@@ -17,11 +17,10 @@
 #include <cassert>
 #include <iostream>
 
-#include <range/v3/view/repeat_n.hpp>
-
 #include <seqan3/argument_parser/detail/format_base.hpp>
 #include <seqan3/core/detail/debug_stream_type.hpp>
 #include <seqan3/core/detail/terminal.hpp>
+#include <seqan3/range/views/repeat_n.hpp>
```
eseiler: Kinda unrelated :D

rrahn: well, not really. Because it some how broke due to removing other stuff. So there must have been a weird include graph. I only fixed stuff that was broken due to the changes.

include/seqan3/argument_parser/detail/format_help.hpp
```diff
@@ -498,7 +497,7 @@ class format_copyright : public format_help
     void parse(argument_parser_meta_data const & parser_meta)
     {
         meta = parser_meta;
-        debug_stream_type strem{std::cout};
```
smehringer: :see_no_evil:

include/seqan3/core/char_operations/predicate_detail.hpp
```diff
@@ -236,7 +236,7 @@ template <char_predicate... condition_ts>
 struct char_predicate_combiner : public char_predicate_base<char_predicate_combiner<condition_ts...>>
 {
     //!\brief The message representing the disjunction of the associated conditions.
-    static constexpr auto msg = detail::condition_message_v<'|', condition_ts...>;
+    inline static const std::string msg = detail::condition_message_v<'|', condition_ts...>;
```
smehringer: I wonder if using a `std::string_view` as the member is any better than `std::string`?

rrahn: That should not work, since we need the strings to use the append feature. They are not constant and thus the string view would have dangling pointers as being initialised with temporaries.

include/seqan3/core/char_operations/predicate_detail.hpp
```diff
@@ -101,7 +101,7 @@ class char_predicate_base;
  * \details
  *
  * An object of the type must be invocable with a std::integral type and supply a static constexpr `msg` member of type
- * seqan3::small_string.
+ * seqan3::string.
```
smehringer:
```suggestion
 * std::string.
```
!

smehringer: you might want to grab for this in your changes

rrahn: I did this alredy. Sorry, I rebased because I thought this can be then merged already.

smehringer: huh, But it is still in the code?

rrahn: the diff shows that it is not in the code anymore. I also don't see it in my checkout 🤷‍♂

rrahn:
@smehringer polite ping

include/seqan3/core/char_operations/predicate_detail.hpp
rrahn: 🤦‍♂ didn't see the actual issue.

PR #1425, opened 2019-12-15
[INFRA] Support seqan3-config-version.cmake

marehr:
This PR adds support for requesting different versions of seqan3:

* `find_package(SeqAn3 3.1)` will search for SeqAn3 with a version of at least 3.1.0 and at most 4.0.0

To support this we need a new cmake file called `seqan3-config-version.cmake`

See https://cmake.org/cmake/help/v3.16/command/find_package.html#version-selection for more information.


build_system/seqan3-config-version.cmake
```diff
@@ -0,0 +1,48 @@
...
+    set(PACKAGE_VERSION_COMPATIBLE FALSE)
+  endif()
+
+  if(PACKAGE_FIND_VERSION STREQUAL PACKAGE_VERSION)
+      set(PACKAGE_VERSION_EXACT TRUE)
+  endif()
+endif()
+
+# if the installed or the using project don't have CMAKE_SIZEOF_VOID_P set, ignore it:
+if("${CMAKE_SIZEOF_VOID_P}" STREQUAL "" OR "8" STREQUAL "")
```
joergi-w: You compare "8" with "" which is always false:

```suggestion
if("${CMAKE_SIZEOF_VOID_P}" STREQUAL "" OR FALSE)
```

build_system/seqan3-config-version.cmake
```diff
@@ -0,0 +1,48 @@
...
+  return()
+endif()
+
+# check that the installed version has the same 32/64bit-ness as the one which is currently searching:
+if(NOT CMAKE_SIZEOF_VOID_P STREQUAL "8")
```
joergi-w: Actually you could add something like `AND NOT "${CMAKE_SIZEOF_VOID_P}" STREQUAL ""` to this condition and omit the if clause above?! 

Here you do not wrap the variable in `"${ }"`, does it make a difference?

build_system/seqan3-config.cmake
```diff
@@ -139,7 +139,9 @@ endmacro ()
 # Detect if we are a clone of repository and if yes auto-add submodules
 # ----------------------------------------------------------------------------
 # Note that seqan3-config.cmake can be standalone and thus SEQAN3_CLONE_DIR might be empty.
-find_path (SEQAN3_CLONE_DIR NAMES build_system/seqan3-config.cmake HINTS "${CMAKE_CURRENT_LIST_DIR}/..")
+# * `SEQAN3_CLONE_DIR` was already found in seqan3-config-version.cmake
+# * `SEQAN3_INCLUDE_DIR` was already found in seqan3-config-version.cmake
+find_path (SEQAN3_SUBMODULES_DIR NAMES submodules/range-v3 HINTS "${SEQAN3_CLONE_DIR}" "${SEQAN3_INCLUDE_DIR}/seqan3")
```
joergi-w: Again:
```suggestion
find_path (SEQAN3_SUBMODULES_DIR NAMES submodules/sdsl-lite HINTS "${SEQAN3_CLONE_DIR}" "${SEQAN3_INCLUDE_DIR}/seqan3")
```

build_system/seqan3-config-version.cmake
marehr: Yeah true, I just used a script to first auto-generated this file and afterwards fill it with stuff.

build_system/seqan3-config-version.cmake
marehr: Good question :) I'm not sure, I guess wrapping it in "" would have the same effect.

joergi-w: Then make it consistent with above, please :)

marehr: > Actually you could add something like `AND NOT "${CMAKE_SIZEOF_VOID_P}" STREQUAL ""` to this condition and omit the if clause above?! 

true, but I don't know how boolean operator precedences work in cmake, so I keep it split

joergi-w: And parentheses don't work?

marehr: I think it becomes unreadable

build_system/seqan3-config-version.cmake
```diff
@@ -0,0 +1,48 @@
...
+# extract version from seqan3/version.hpp header
+file(STRINGS "${SEQAN3_INCLUDE_DIR}/seqan3/version.hpp" SEQAN3_VERSION_HPP REGEX "#define SEQAN3_VERSION_(MAJOR|MINOR|PATCH)")
+string(REGEX REPLACE "#define SEQAN3_VERSION_(MAJOR|MINOR|PATCH) " "" PACKAGE_VERSION "${SEQAN3_VERSION_HPP}")
+string(REGEX REPLACE ";" "." PACKAGE_VERSION "${PACKAGE_VERSION}")
+
+if(PACKAGE_VERSION VERSION_LESS PACKAGE_FIND_VERSION)
+  set(PACKAGE_VERSION_COMPATIBLE FALSE)
+else()
+
+  if(PACKAGE_VERSION MATCHES "^([0-9]+)\\.")
```
rrahn: please use the same spacing as used for regular source code

build_system/seqan3-config-version.cmake
```diff
@@ -0,0 +1,48 @@
...
+    set(CVF_VERSION_MAJOR "${CMAKE_MATCH_1}")
```
rrahn: can we make this a proper name or document it.

build_system/seqan3-config-version.cmake
```diff
@@ -0,0 +1,48 @@
+# -----------------------------------------------------------------------------------------------------
+# Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+# Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+# This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+# shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+# -----------------------------------------------------------------------------------------------------
+
+# This file adds version support for `find_package(SEQAN3 3.1)`.
+
```
rrahn: can you please document that this was auto generated by which script in which version and a link to it?

CHANGELOG.md
```diff
@@ -41,6 +41,12 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
   You are only affected if you added types to `seqan3::custom::`.
   Please see [About Customisation](http://docs.seqan.de/seqan/3-master-user/about_customisation.html).
 
+### CMake
+
+* [find_package](https://cmake.org/cmake/help/latest/command/find_package.html#version-selection) now accepts minimum
+  versions (e.g. `find_package(SEQAN3 3.1.0)` requires at least seqan3 with a version of `>= 3.1.0` and `< 4.0.0`).
+* `SEQAN3_VERSION_STRING` was renamed to `SEQAN3_VERSION` (`find_package(SEQAN3)`).
```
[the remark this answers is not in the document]
marehr: This variable will be set by `find_package(SEQAN3)`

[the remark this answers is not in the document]
rrahn: Can you make it a full sentence then please. Like
```
The variable `SEQAN3_VERSION_STRING` defined by find_package(SEQAN3) was renamed to `SEQAN3_VERSION`.
```

CHANGELOG.md
```diff
@@ -55,6 +49,12 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
   Most user code will be unaffected; to fix possible compiler errors you need to add an empty template parameter list to
   the respective instances (e.g. change `input_file_validator` to `input_file_validator<>`).
 
+#### CMake
+
+* [find_package](https://cmake.org/cmake/help/latest/command/find_package.html#version-selection) now accepts minimum
+versions (e.g. `find_package(SEQAN3 3.1.0)` requires at least seqan3 with a version of `>= 3.1.0` and `< 4.0.0`).
```
rrahn: We should actually not use 3.1.0 here, even if it is just an example, because we don't have a 3.1.0 release yet. Better use 3.0.1, which would at least be correct when we release it. What do you think?

rrahn:
@marehr you can already rebase and squash all the review commits when you applied the last change. Many thanks

marehr:
@rrahn done :)

PR #1426, opened 2019-12-15
[INFRA] some cmake maintenance 

marehr:
Handle `CMAKE_REQUIRED_QUIET` like most other `FindMyModule.cmake` do and follow naming scheme used in cmake.

### DIFF
--- a/build_system/seqan3-config.cmake
+++ b/build_system/seqan3-config.cmake
@@ -214,7 +214,7 @@ option (SEQAN3_NO_BZIP2 "Don't use BZip2, even if present." OFF)
 # Require C++17
 # ----------------------------------------------------------------------------
 
-set (CMAKE_REQUIRED_FLAGS_ORIGINAL ${CMAKE_REQUIRED_FLAGS})
+set (CMAKE_REQUIRED_FLAGS_SAVE ${CMAKE_REQUIRED_FLAGS})
 
 set (CXXSTD_TEST_SOURCE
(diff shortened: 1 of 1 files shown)


build_system/seqan3-config.cmake
```diff
@@ -238,7 +241,7 @@ option (SEQAN3_NO_BZIP2 "Don't use BZip2, even if present." OFF)
 # Require C++17
 # ----------------------------------------------------------------------------
 
-set (CMAKE_REQUIRED_FLAGS_ORIGINAL ${CMAKE_REQUIRED_FLAGS})
+set (CMAKE_REQUIRED_FLAGS_SAVE ${CMAKE_REQUIRED_FLAGS})
```
rrahn: Why this change? I am afraid, I can't completely follow your intent here?!

marehr: > Handle `CMAKE_REQUIRED_QUIET` like most other `FindMyModule.cmake` do and ***follow naming scheme used in cmake***.

If you look into cmake modules, you see that the naming uses `_SAVE` if you save the previous content of a variable to restore it later.

See for example https://github.com/Kitware/CMake/blob/512ab500f06d6c645985cc8014c5e6291b9a059f/Modules/CMakeBackwardCompatibilityCXX.cmake#L40

I thought maybe of using the same terminology but I can revert it if it is not understandable

rrahn: ok. I don't mind

PR #1445, opened 2019-12-19
[FEATURE] Search result range

### LINKED ISSUE #1435: [Search] `search()` should lazily generate results
**Level 1 - Easy**

Make `search()` return a range where:
* `operator++` may fetch new results
* `operator*` returns the cached result
* `begin()` fetches result if there is none

See alignment code: https://github.com/seqan/seqan3/blob/master/include/seqan3/alignment/pairwise/alignment_range.hpp
Depends on `search_result` task

- [ ]  Implement the range that models input_range_concept
- [ ] add tests for the range with mocked search results
- [ ] add documentation

rrahn:
fixes #1435 

This implements already the lazy range for the search results. 
The tutorials were only fixed for snippet tests. The entire documentation must be updated after the result type and this range are merged. 
Also the tests must be touched and merged to work with the alignment result.


test/unit/search/search_collection_test.cpp
```diff
@@ -18,20 +18,26 @@ using namespace seqan3;
 using namespace seqan3::search_cfg;
 using namespace std::string_literals;
 
-template <typename T>
+template <typename index_t>
 class search_test : public ::testing::Test
 {
 public:
+    using collection_res_t = std::pair<typename index_t::size_type, typename index_t::size_type>;
```
marehr:
```suggestion
    using collection_result_t = std::pair<typename index_t::size_type, typename index_t::size_type>;
```

test/unit/search/search_collection_test.cpp
```diff
@@ -18,20 +18,26 @@ using namespace seqan3;
...
-    T index{text};
+    index_t index{text};
 };
 
-template <typename T>
+template <typename index_t>
 class search_string_test : public ::testing::Test
 {
 public:
+    using collection_res_t = std::pair<typename index_t::size_type, typename index_t::size_type>;
```
marehr:
```suggestion
    using collection_result_t = std::pair<typename index_t::size_type, typename index_t::size_type>;
```

test/unit/search/search_collection_test.cpp
```diff
@@ -44,149 +50,141 @@ TYPED_TEST_CASE(search_string_test, fm_index_string_types);
 
 TYPED_TEST(search_test, error_free)
 {
-    using result_t = std::pair<typename TypeParam::size_type, typename TypeParam::size_type>;
-    using hits_result_t = std::vector<result_t>;
+    using hits_result_t = typename TestFixture::hits_result_t;
 
+    hits_result_t hits{{0, {0, 0}}, {0, {0, 4}}, {0, {0, 8}},
```
marehr:
```suggestion
    hits_result_t expect_hits{{0, {0, 0}}, {0, {0, 4}}, {0, {0, 8}},
```

test/unit/search/search_collection_test.cpp
```diff
@@ -44,149 +50,141 @@ TYPED_TEST_CASE(search_string_test, fm_index_string_types);
...
-    using hits_result_t = std::vector<result_t>;
+    using hits_result_t = typename TestFixture::hits_result_t;
+
     std::vector<qualified<dna4, phred42>> query{{'A'_dna4, '!'_phred42},
                                                 {'C'_dna4, '!'_phred42},
                                                 {'G'_dna4, '!'_phred42},
                                                 {'T'_dna4, '!'_phred42}};
 
-    EXPECT_EQ(uniquify(search(query, this->index)), (hits_result_t{{0, 0}, {0, 4}, {0, 8}, {1, 0}, {1, 4}, {1, 8}}));
+    hits_result_t hits{{0, {0, 0}}, {0, {0, 4}}, {0, {0, 8}},
```
marehr:
```suggestion
    hits_result_t expect_hits{{0, {0, 0}}, {0, {0, 4}}, {0, {0, 8}},
```

test/unit/search/search_collection_test.cpp
```diff
@@ -44,149 +50,141 @@ TYPED_TEST_CASE(search_string_test, fm_index_string_types);
...
 
 TYPED_TEST(search_test, multiple_queries)
 {
-    using result_t = std::vector<std::pair<typename TypeParam::size_type, typename TypeParam::size_type>>;
-    using hits_result_t = std::vector<result_t>;
+    using hits_result_t = typename TestFixture::hits_result_t;
+
     std::vector<std::vector<dna4>> const queries{{"GG"_dna4, "ACGTACGTACGT"_dna4, "ACGTA"_dna4}};
 
+    hits_result_t hits{{1, {0, 0}}, {1, {1, 0}},
```
marehr: expect_hits

test/unit/search/search_collection_test.cpp
```diff
@@ -44,149 +50,141 @@ TYPED_TEST_CASE(search_string_test, fm_index_string_types);
...
 }
 
 TYPED_TEST(search_string_test, error_free_string)
 {
-    using result_t = std::pair<typename TypeParam::size_type, typename TypeParam::size_type>;
-    using hits_result_t = std::vector<result_t>;
+    using hits_result_t = typename TestFixture::hits_result_t;
 
     {
+        hits_result_t hits{{0, {0, 14}}, {0, {0, 18}}, {0, {1, 17}}};
```
marehr: expect_hits

test/unit/search/search_collection_test.cpp
```diff
@@ -44,149 +50,141 @@ TYPED_TEST_CASE(search_string_test, fm_index_string_types);
...
 }
 
 TYPED_TEST(search_string_test, error_free_raw)
 {
-    using result_t = std::pair<typename TypeParam::size_type, typename TypeParam::size_type>;
-    using hits_result_t = std::vector<result_t>;
+    using hits_result_t = typename TestFixture::hits_result_t;
 
     {
+        hits_result_t hits{{0, {0, 14}}, {0, {0, 18}}, {0, {1, 17}}};
```
marehr: expect_hits

test/unit/search/search_collection_test.cpp
```diff
@@ -44,149 +50,141 @@ TYPED_TEST_CASE(search_string_test, fm_index_string_types);
...
 {
-    using result_t = std::vector<std::pair<typename TypeParam::size_type, typename TypeParam::size_type>>;
-    using hits_result_t = std::vector<result_t>;
+    using hits_result_t = typename TestFixture::hits_result_t;
 
     std::vector<std::string> const queries{"at", "Jon"};
 
-    EXPECT_EQ(uniquify(search(queries, this->index)), (hits_result_t{{{0, 14}, {0, 18}, {1, 17}},
-                                                                     {}})); // 3 and 0 hits
+    hits_result_t hits{{0, {0, 14}}, {0, {0, 18}}, {0, {1, 17}}};
```
marehr: expect_hits

test/unit/search/search_collection_test.cpp
```diff
@@ -44,149 +50,141 @@ TYPED_TEST_CASE(search_string_test, fm_index_string_types);
...
 
 TYPED_TEST(search_string_test, multiple_queries_raw)
 {
-    using result_t = std::vector<std::pair<typename TypeParam::size_type, typename TypeParam::size_type>>;
-    using hits_result_t = std::vector<result_t>;
+    using hits_result_t = typename TestFixture::hits_result_t;
 
-    EXPECT_EQ(uniquify(search({"at", "Jon"}, this->index)), (hits_result_t{{{0, 14}, {0, 18}, {1, 17}},
-                                                                           {}})); // 3 and 0 hits
+    hits_result_t hits{{0, {0, 14}}, {0, {0, 18}}, {0, {1, 17}}};
```
marehr: expect_hits

test/unit/search/search_test.cpp
```diff
@@ -224,237 +229,247 @@ TYPED_TEST(search_test, error_insertion)
...
-    using hits_result_t = std::vector<typename TypeParam::size_type>;
+    using hits_result_t = typename TestFixture::hits_result_t;
 
-    hits_result_t possible_hits{0, 4, 8}; // any of 0, 4, 8 ... 1, 5, 9 are not best hits
+    hits_result_t possible_hits{{0, 0}, {0, 4}, {0, 8}}; // any of 0, 4, 8 ... 1, 5, 9 are not best hits
     {
         configuration const cfg = max_error{total{1}} | mode{best};
 
-        hits_result_t result = search("ACGT"_dna4, this->index, cfg);
+        hits_result_t result = transform_to_vector(search("ACGT"_dna4, this->index, cfg));
```
marehr: ah you are using it here,

test/unit/search/search_test.cpp
```diff
@@ -224,237 +229,247 @@ TYPED_TEST(search_test, error_insertion)
...
+        EXPECT_EQ(transform_to_vector(search("AAAA"_dna4, this->index, cfg)), (hits_result_t{})); // no hit
     }
 
     // {
     //     // best hit ACGT with 1 error, i.e. 1+1
     //     configuration const cfg = max_total_error{1} | strategy_strata{1};
-    //     EXPECT_EQ(uniquify(search(this->index, "CCGT"_dna4, cfg)), (hits_result_t{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10}));
+    //     EXPECT_EQ(uniquify(search(this->index, "CCGT"_dna4, cfg)), (hits_result_t{{0, 0}, {0, 1}, {0, 2}, {0, 3},
+    //                                                                               {0, 4}, {0, 5}, {0, 6}, {0, 7},
+    //                                                                               {0, 8}, {0, 9}, {0, 1}{0, 0}}));
```
marehr: why is that commented out?

test/unit/search/search_test.cpp
```diff
@@ -224,237 +229,247 @@ TYPED_TEST(search_test, error_insertion)
...
     // }
 
     // {
     //     // best hit ACGT with 1 error, i.e. 1+1
     //     configuration const cfg = max_total_error{1} | strategy_strata{1};
-    //     EXPECT_EQ(uniquify(search(this->index, "CCGT"_dna4, cfg)), (hits_result_t{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10}));
+    //     EXPECT_EQ(uniquify(search(this->index, "CCGT"_dna4, cfg)), (hits_result_t{{0, 0}, {0, 1}, {0, 2}, {0, 3},
+    //                                                                               {0, 4}, {0, 5}, {0, 6}, {0, 7},
+    //                                                                               {0, 8}, {0, 9}, {0, 1}{0, 0}}));
     // }
```
marehr: why is that commented out?

include/seqan3/search/algorithm/search_result_range.hpp
```diff
@@ -0,0 +1,331 @@
...
+ */
+template <typename search_algorithm_t, std::ranges::view query_range_t>
+//!\cond
+    requires std::ranges::forward_range<query_range_t> &&
+             std::invocable<search_algorithm_t, std::ranges::range_reference_t<query_range_t>>
+//!\endcond
+class search_result_range
+{
+private:
+    class search_result_range_iterator;
```
marehr: do you need this?

include/seqan3/search/algorithm/search_result_range.hpp
```diff
@@ -0,0 +1,331 @@
...
+
+    //!\brief Befriend the iterator type.
+    friend class search_result_range_iterator;
```
marehr: isn't this enough?

include/seqan3/search/algorithm/search_result_range.hpp
```diff
@@ -0,0 +1,331 @@
...
+     *
+     * \returns `true` if hits could be found for the current query, otherwise `false`.
+     *
+     * \details
+     *
+     * Invokes the search on the current query and increments the query iterator.
+     */
+    bool next()
+    {
+        using std::swap;
```
marehr: you don't use this :)

include/seqan3/search/algorithm/search_result_range.hpp
```diff
@@ -0,0 +1,331 @@
...
+    {
+        if (bool has_next_results = range_ptr->next(); has_next_results)
+        {
+            current_buffer_it = std::ranges::begin(range_ptr->result_buffer);
+            last_buffer_it = std::ranges::end(range_ptr->result_buffer);
+        }
+        else if (!range_ptr->at_end())
+        {
+            fetch_next_query_results();
+        }
```
marehr: Isn't this a loop?

```
 // call next until it returns true or we reached the end
while(!(range_ptr->next() || range_ptr->at_end()));
 
// these are guaranteed to be valid
// even if next always returned false in that case is the result_buffer empty
current_buffer_it = std::ranges::begin(range_ptr->result_buffer); 
last_buffer_it = std::ranges::end(range_ptr->result_buffer);
```

include/seqan3/search/algorithm/search_result_range.hpp
```diff
@@ -0,0 +1,331 @@
...
+    search_result_range & operator=(search_result_range const &) = delete; //!< This is a move-only type.
+    search_result_range & operator=(search_result_range &&) = default; //!< Defaulted.
+    ~search_result_range() = default; //!< Defaulted.
+
+    /*!\brief Constructs a search result range.
+     *
+     * \param[in] search_algorithm The search algorithm to invoke on a single query.
+     * \param[in] query_range The range over the queries to be searched.
+     */
+    search_result_range(search_algorithm_t search_algorithm, query_range_t query_range) :
```
marehr: What happens if `query_range_t` is a lvalue reference? (because of the move?)

include/seqan3/search/algorithm/search.hpp
```diff
@@ -15,10 +15,63 @@
...
+     * Checks if the given thresholds for the configured errors are valid. Otherwise throws std::invalid_argument.
+     */
+    template <typename configuration_t>
+    static void validate_error_configuration(configuration_t const & cfg)
+    {
+        using search_traits_t = detail::search_traits<configuration_t>;
+
+        if constexpr (search_traits_t::search_with_max_error)
+        {
+            auto & [total, subs, ins, del] = get<search_cfg::max_error>(cfg).value;
```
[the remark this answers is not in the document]
rrahn: Well, the values are always lvalues so && would always resolve to &.

include/seqan3/search/algorithm/search.hpp
```diff
@@ -15,10 +15,63 @@
...
+                throw std::invalid_argument{"The substitution error threshold is higher than the total error "
+                                            "threshold."};
+            if (ins > total)
+                throw std::invalid_argument{"The insertion error threshold is higher than the total error threshold."};
+            if (del > total)
+                throw std::invalid_argument{"The deletion error threshold is higher than the total error threshold."};
+        }
+        else if constexpr (search_traits_t::search_with_max_error_rate)
+        {
+            auto & [total, subs, ins, del] = get<search_cfg::max_error_rate>(cfg).value;
```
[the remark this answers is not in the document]
rrahn: See above

include/seqan3/search/algorithm/search.hpp
```diff
@@ -87,63 +140,47 @@ inline auto search(queries_t && queries,
...
+                static_assert(std::ranges::forward_range<queries_t>, "The query collection must model forward_range.");
+                static_assert(std::ranges::sized_range<queries_t>, "The query collection must model sized_range.");
+                static_assert(std::ranges::random_access_range<value_type_t<queries_t>>,
+                            "Elements of the query collection must model random_access_range.");
+                static_assert(std::ranges::sized_range<value_type_t<queries_t>>,
+                            "Elements of the query collection must model sized_range.");
+            }
 
-    using cfg_t = remove_cvref_t<configuration_t>;
+            static_assert(detail::is_type_specialisation_of_v<remove_cvref_t<configuration_t>, configuration>,
```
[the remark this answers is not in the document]
rrahn: No they don't but albeit I did some little refactoring. I think the entire configuration step should be refactored. It would be great to have a separate configuration step similar to the alignment code and then invoke the configured search function. This makes it also easier to add parallel execution and possibly reusing the existing execution handler for this.

include/seqan3/search/algorithm/search.hpp
```diff
@@ -87,63 +140,47 @@ inline auto search(queries_t && queries,
...
-    static_assert(detail::is_type_specialisation_of_v<remove_cvref_t<configuration_t>, configuration>,
-                  "cfg must be a specialisation of seqan3::configuration.");
+    {   // If no output configuration was, default to returning text positions.
+        if constexpr (!search_traits_t::has_output_configuration)
+        {
+            return search(std::forward<queries_t>(queries), index, cfg | search_cfg::output{search_cfg::text_position});
+        }
+        else
+        {
+            if constexpr(dimension_v<queries_t> == 1u)
```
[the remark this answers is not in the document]
rrahn: yes but it is old code that I didn't touch. My refactoring only involves not repeatedly calling the same function again.

doc/tutorial/read_mapper/read_mapper_step4.cpp
```diff
@@ -66,19 +66,18 @@ void map_reads(std::filesystem::path const & query_path,
...
     for (auto & [query, id, qual] : query_in)
     {
-        auto positions = search(query, index, search_config);
-        for (auto & [idx, pos] : positions)
+        for (auto & [query_idx, text_pos] : search(query, index, search_config))
         {
-            size_t start = pos ? pos - 1 : 0;
-            std::span text_view{std::data(storage.seqs[idx]) + start, query.size() + 1};
+            size_t start = text_pos.second ? text_pos.second - 1 : 0;
+            std::span text_view{std::data(storage.seqs[text_pos.first]) + start, query.size() + 1};
```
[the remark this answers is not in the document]
rrahn: Because it is not a search result yet. It will be changed to a search result. So it was before already a pair. Second, the first is the index of the sequence and the second is the position of the hit within the sequence. Anyway, that will be changed to a search result.

doc/tutorial/search/search_solution2.cpp
```diff
@@ -23,11 +20,8 @@ void run_text_collection()
                                           "GGCCAGACAACCCGGCGCTAATGCACTCA"_dna4};
     seqan3::fm_index index{text};
 
-    auto results = search("GCT"_dna4, index);
-    std::ranges::sort(results);
     seqan3::debug_stream << "===== Running on a text collection =====\n";
-    seqan3::debug_stream << "There are " << results.size() << " hits.\n";
```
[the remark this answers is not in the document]
rrahn: I did, yes. Not guaranteeing that I missed something.

include/seqan3/search/algorithm/detail/search.hpp
```diff
@@ -12,11 +12,12 @@
 
 #pragma once
 
-#include <seqan3/core/type_traits/pre.hpp>
+// #include <seqan3/core/type_traits/pre.hpp>
```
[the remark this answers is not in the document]
rrahn: I remove it

include/seqan3/search/algorithm/detail/search.hpp
```diff
@@ -181,36 +182,32 @@ inline auto search_single(index_t const & index, query_t & query, configuration_
...
     {
-        // TODO: if constexpr (contains<search_cfg::id::on_hit>(cfg))
-        std::vector<std::vector<hit_t>> hits;
-        hits.reserve(std::distance(queries.begin(), queries.end()));
-        for (auto const query : queries)
-        {
-            hits.push_back(search_single(index, query, cfg));
-        }
-        return hits;
+        return search_all(index, std::views::single(std::forward<queries_t>(queries)), cfg);
```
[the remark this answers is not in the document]
rrahn: what do you mean?

include/seqan3/search/algorithm/search.hpp
```diff
@@ -163,7 +200,7 @@ inline auto search(std::initializer_list<char const * const> const & queries,
     std::vector<std::string_view> query;
     query.reserve(std::ranges::size(queries));
     std::ranges::for_each(queries, [&query] (char const * const q) { query.push_back(std::string_view{q}); });
```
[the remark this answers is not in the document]
rrahn: 🤷‍♂

include/seqan3/search/algorithm/search.hpp
```diff
@@ -163,7 +200,7 @@ inline auto search(std::initializer_list<char const * const> const & queries,
...
-    return search(query, index, cfg);
+    return search(std::move(query) | views::persist, index, cfg);
```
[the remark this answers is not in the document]
rrahn: In general I am in favor of removing this interface altogether. For now, this is a quick fix to propagate the temporary as a view to the result range.

include/seqan3/search/algorithm/search_result_range.hpp
rrahn: I need to declare the iterator, or do I not?

include/seqan3/search/algorithm/search_result_range.hpp
rrahn: are these not two different expressions?

include/seqan3/search/algorithm/search_result_range.hpp
rrahn: it never is since it is required to be a view.

include/seqan3/search/algorithm/search_result_range.hpp
rrahn: not anymore that's right

test/unit/search/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ seqan3_test (sdsl_index_test.cpp)
 
 seqan3_test (search_collection_test.cpp)
 seqan3_test (search_configuration_test.cpp)
+seqan3_test (search_result_range_test.cpp)
```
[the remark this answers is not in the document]
rrahn: i look into it

include/seqan3/search/algorithm/search_result_range.hpp
rrahn: if you say it like this 😏

test/unit/search/helper.hpp
```diff
@@ -11,38 +11,31 @@
...
 }
 
-template <typename T1, typename T2>
-std::vector<std::vector<std::pair<T1, T2>>> uniquify(std::vector<std::vector<std::pair<T1, T2>>> v)
+template <typename search_result_range_t>
+std::vector<std::ranges::range_value_t<search_result_range_t>> uniquify(search_result_range_t && rng)
 {
-    std::for_each(v.begin(), v.end(), [] (auto & hits) { uniquify(hits); } );
-    return v;
+    auto unique_res = transform_to_vector(std::forward<search_result_range_t>(rng));
```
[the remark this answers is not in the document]
rrahn: The first part does not work, since the passed range is not a common range. For the reason of the second part, I don't know. That was before I left it because it is not part of this PR anyway. If you feel that is worth discussing you could put it on the strategy meeting.

test/unit/search/search_result_range_test.cpp
```diff
@@ -0,0 +1,102 @@
...
+{
+    seqan3::search_result_range rng{dummy_search_algorithm{}, query_range | seqan3::views::all};
+    EXPECT_TRUE((std::is_same_v<decltype(rng),
+                            seqan3::search_result_range<dummy_search_algorithm,
+                                                        seqan3::all_view<std::vector<std::string>&>>>));
+}
+
+TEST_F(search_result_range_test, empty_query_range)
+{
+    query_range.clear();
```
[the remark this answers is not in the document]
rrahn: No, the test fixtures are initialised for every test instance. So modifications are not visible outside of the current test.

test/unit/search/search_test.cpp
rrahn: dunno. Was so before

test/unit/search/search_test.cpp
rrahn: 🤷‍♂

test/unit/search/helper.hpp
```diff
@@ -11,38 +11,31 @@
...
 
-template <typename T1, typename T2>
-std::vector<std::pair<T1, T2>> uniquify(std::vector<std::pair<T1, T2>> v)
-{
-    std::sort(v.begin(), v.end());
-    v.erase(std::unique(v.begin(), v.end()), v.end());
-    return v;
+    std::vector<hit_t> results{};
+    for (auto it = rng.begin(); it != rng.end(); ++it)
+        results.push_back(std::ranges::iter_move(it));
```
rrahn: views::to

test/unit/search/search_test.cpp
rrahn: , missing

rrahn:
@marehr I hope I did catch everything

rrahn:
@marehr btw, the commit order is correct. Only GitHub displays it in a wrong order: https://help.github.com/en/github/committing-changes-to-your-project/why-are-my-commits-in-the-wrong-order
Another reason why not to use rebase to often during the review process.

include/seqan3/search/algorithm/search_result_range.hpp
```diff
@@ -0,0 +1,323 @@
...
+namespace seqan3
+{
+
+/*!\brief An input range over the search results generated by the given search algorithm.
+ * \ingroup submodule_search_algorithm
+ * \implements std::ranges::input_range
+ *
+ * \tparam search_algorithm_t The type of the search algorithm; must model std::invocable with the value type of
+ *                            the given range query over queries.
+ * \tparam query_range_t The type of the range over the queries; must model std::ranges::forward_range.
```
marehr:
```suggestion
 * \tparam query_range_t The type of the range over the queries; must model std::ranges::view and std::ranges::forward_range.
```

rrahn:
@marehr I fixed the last outstanding issue and did already a rebase.

marehr:
codecov shows a drop in coverage:

https://codecov.io/gh/seqan/seqan3/compare/9cb88241b70e9c50715359eb37ecd48919344295...f74b9547b645723fb571b33f95565550e309529d/diff#D1-54...72

include/seqan3/search/algorithm/detail/search.hpp
```diff
@@ -181,36 +181,32 @@ inline auto search_single(index_t const & index, query_t & query, configuration_
...
-    //                  delegate for each hit separately at once.
-    using text_pos_t = std::conditional_t<index_t::text_layout_mode == text_layout::collection,
-                                          std::pair<typename index_t::size_type, typename index_t::size_type>,
-                                          typename index_t::size_type>;
-    using hit_t = std::conditional_t<cfg_t::template exists<search_cfg::output<detail::search_output_index_cursor>>(),
-                                     typename index_t::cursor_type,
-                                     text_pos_t>;
-
-    if constexpr (std::ranges::forward_range<queries_t> && std::ranges::random_access_range<value_type_t<queries_t>>)
+    if constexpr (!std::ranges::forward_range<std::ranges::range_value_t<queries_t>>)
```
smehringer: What about a collection of input ranges? wouldn't this cause an infinite loop here?

doc/tutorial/read_mapper/read_mapper_step3.cpp
```diff
@@ -58,11 +58,10 @@ void map_reads(std::filesystem::path const & query_path,
 
     for (auto & [query, id, qual] : query_in | seqan3::views::take(20))
     {
-        auto positions = search(query, index, search_config);
-        for (auto & [idx, pos] : positions)
+        for (auto & [query_idx, text_pos] : search(query, index, search_config))
         {
-            size_t start = pos ? pos - 1 : 0;
-            std::span text_view{std::data(storage.seqs[idx]) + start, query.size() + 1};
+            size_t start = text_pos.second ? text_pos.second - 1 : 0;
```
smehringer: Huh, so the result now is `{id, {id, pos}}`? I thought the result range is flattened.

include/seqan3/search/algorithm/search.hpp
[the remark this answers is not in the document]
rrahn: the result range that is returned holds the queries. It requires the queries to be a view. Hence the initialiser list which is converted into a std::vector must propagate the queries on the heap to the result range. Just using view::all would cause a dangling reference.

include/seqan3/search/algorithm/detail/search.hpp
rrahn: This interface is called internally and is a detail interface. The public interface that calls this does all necessary checks. Input ranges are not allowed in general. The only question is if this is a collection of collection or not. And if so it calls itself again by making it a collection of collection.

include/seqan3/search/algorithm/search.hpp
```diff
@@ -87,63 +168,29 @@ inline auto search(queries_t && queries,
...
-            return detail::search_all(index, queries, cfg2);
+    {   // If no output configuration was, default to returning text positions.
+        if constexpr (!search_traits_t::has_output_configuration)
+        {
+            return search(std::forward<queries_t>(queries), index, cfg | search_cfg::output{search_cfg::text_position});
+        }
         else
-            return detail::search_all(index, queries, cfg2 | search_cfg::output{search_cfg::text_position});
+        {
+            detail::search_configuration_validator::validate_query_type<queries_t>();
```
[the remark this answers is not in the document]
rrahn: > You only validate the query type if a mode and output config was set but shouldn't this be always checked? 

No matter what you specify in the config it will always pass through this lines of code. If some options are not present a default configuration for them will be added and the function is called again with the modified config.
So I think that should be fine?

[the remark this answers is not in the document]
smehringer: Ah I did not see that, that's not so easy to spot. I would prefer a lambda that just alters the config :grin:  but yes it's fine like this.

include/seqan3/search/algorithm/detail/search.hpp
smehringer: Maybe add an assert anyway?

include/seqan3/search/algorithm/search_result_range.hpp
```diff
@@ -0,0 +1,324 @@
...
+     * \details
+     *
+     * Initialises the iterator with the associated search result range. If `first_invocation_of_begin` is `true`,
+     * the search algorithm is invoked on the first query.
+     */
+    constexpr search_result_range_iterator(search_result_range & range, bool const & first_invocation_of_begin) :
+        range_ptr{& range},
+        at_end(range.at_end())
+    {
+        if (!first_invocation_of_begin)
```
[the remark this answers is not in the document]
rrahn: what is misleading? the first time begin is called the first query is searched during the construction of the iterator. Afterwards, a search is only triggered if the ++operator is called and the current search buffer has been consumed.

include/seqan3/search/algorithm/search_result_range.hpp
```diff
@@ -0,0 +1,324 @@
...
+    }
+
+    //!\brief The current iterator over the buffered search results.
+    result_buffer_iterator current_buffer_it{};
+    //!\brief The end iterator over the buffered search results.
+    result_buffer_iterator last_buffer_it{};
+    //!\brief Pointer to the underlying range.
+    search_result_range * range_ptr{};
+    //!\brief Indicates the end of the underlying range.
+    bool at_end{true};
```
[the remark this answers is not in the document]
rrahn: No easily, since the comparison operator is a friend and has no access to the range the iterator is pointing to. Making the operator a friend of the range class does not seem worth the change. Besides, this is an input iterator, which will be move-only anyway. So size does not really matter.

include/seqan3/search/algorithm/search_result_range.hpp
```diff
@@ -0,0 +1,324 @@
...
+protected:
+    /*!\brief Receives the search results from the next query.
+     *
+     * \returns `true` if hits could be found for the current query, otherwise `false`.
+     *
+     * \details
+     *
+     * Invokes the search on the current query and increments the query iterator.
+     */
+    bool next()
```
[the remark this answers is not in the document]
rrahn: Personal preference. There is no advantage in making it private either. It would allow you to inherit from this and use the functionality to do something. But I agree that this might not be ever the case.

doc/tutorial/read_mapper/read_mapper_step3.cpp
rrahn: The result range is flattened. But example does process the queries query by query. In this case we get the id of the query and the position within the reference index which is a collection. Hence, the return type as a pair. This will be modified once this is merged with the search result

include/seqan3/search/algorithm/detail/search.hpp
rrahn: The point is in my opinion the entire design should be and probably needs to be refactored for the parallel implementation of the search interface. Then we can put this altogether in a configuration class similar to the alignment interface instead of having functions calling functions calling functions where we need to add all redundant checks again and again. So for now, I would simply leave it as is.

include/seqan3/search/algorithm/search_result_range.hpp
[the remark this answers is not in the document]
rrahn: I've updated. It was indeed misleading. Good spot.

rrahn:
@smehringer I rebased this on current master and fixed the open requests.

test/unit/search/search_collection_test.cpp
```diff
@@ -45,162 +51,153 @@ TYPED_TEST_SUITE(search_string_test, fm_index_string_types, );
...
+        EXPECT_EQ(uniquify(search("ACGG"_dna4, this->index, cfg)), hits_result_t{});
     }
 
     {
         // successful and unsuccesful exact search using empty max_total_error_rate
-        seqan3::configuration const cfg = seqan3::search_cfg::max_error_rate{};
-        EXPECT_EQ(seqan3::uniquify(search("ACGT"_dna4, this->index, cfg)), (hits_result_t{{0, 0}, {0, 4}, {0, 8},
-                                                                                          {1, 0}, {1, 4}, {1, 8}}));
-        EXPECT_EQ(seqan3::uniquify(search("ACGG"_dna4, this->index, cfg)), (hits_result_t{}));
+        seqan3::configuration const cfg =seqan3::search_cfg::max_error_rate{};
```
smehringer:
```suggestion
        seqan3::configuration const cfg = seqan3::search_cfg::max_error_rate{};
```
:nail_care: 
Also in a few cases below
Edit: you removed the space almost everywhere. Was this intended? But we do put a space after `=` don't we?

test/unit/search/search_test.cpp
```diff
@@ -104,55 +107,54 @@ TYPED_TEST(search_test, error_free)
...
                                                                          seqan3::search_cfg::deletion{.0}};
-    EXPECT_EQ(seqan3::uniquify(search(queries, this->index, cfg)), (hits_result_t{{}, {0}, {0, 4}})); // 0, 1 and 2 hits
+    EXPECT_EQ(uniquify(search(queries, this->index, cfg)), (hits_result_t{{1, 0}, {2, 0}, {2, 4}})); // 0, 1 and 2 hits
 }
 
 TYPED_TEST(search_test, invalid_error_configuration)
 {
     seqan3::configuration const cfg = seqan3::search_cfg::max_error{seqan3::search_cfg::total{0},
-                                                                    seqan3::search_cfg::substitution{1}};
+                                                                            seqan3::search_cfg::substitution{1}};
```
smehringer:
```suggestion
                                                                    seqan3::search_cfg::substitution{1}};
```

include/seqan3/search/algorithm/detail/search.hpp
```diff
@@ -180,36 +181,32 @@ inline auto search_single(index_t const & index, query_t & query, configuration_
...
+        using text_pos_t = std::conditional_t<index_t::text_layout_mode == text_layout::collection,
+                                              std::pair<typename index_t::size_type, typename index_t::size_type>,
+                                              typename index_t::size_type>;
+        using hit_t = std::conditional_t<search_traits_t::search_return_index_cursor,
+                                         typename index_t::cursor_type,
+                                         text_pos_t>;
+        using query_t = std::ranges::range_value_t<queries_t>;
+        using search_fn_t = std::function<std::vector<hit_t>(query_t &)>;
+
+        search_fn_t search_fn = [&, cfg] (query_t & query) { return search_single(index, query, cfg); };
```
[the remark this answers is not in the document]
rrahn: not sure. Was from 3 months back. Can't remember. Will remove.

include/seqan3/search/algorithm/search_result_range.hpp
```diff
@@ -0,0 +1,320 @@
...
+    //!\}
+
+    /*!\name Comparison operators
+     * \{
+     */
+    //!\brief Checks whether `lhs` is equal to the sentinel.
+    friend constexpr bool operator==(search_result_range_iterator const & lhs,
+                                     std::ranges::default_sentinel_t const &) noexcept
+    {
+        return lhs.at_end && (lhs.current_buffer_it == lhs.last_buffer_it);
```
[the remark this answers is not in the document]
rrahn: It does work, once I befriend `operator==()` in the enclosing range class. Note, that friendship is not transitive. This means just because `friend operator==()` can access the private member of my_range::iterator it cannot access the private member of the enclosing class my_range. Accordingly, the `operator==` cannot access private members of my_range through a pointer stored in my_range::iterator. Here is an example: https://godbolt.org/z/ZZuRfk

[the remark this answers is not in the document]
rrahn: But it is not worht the effort to add complicated friend declarations just to safe a bool in a move-only iterator.

include/seqan3/search/algorithm/search_result_range.hpp
```diff
@@ -0,0 +1,320 @@
...
+     * \{
+     */
+
+    /*!\brief Access the pointed-to element.
+     * \returns A reference to the current element.
+     */
+    //!\brief Returns the current search result.
+    reference operator*() const noexcept
+    {
+        return *current_buffer_it;
```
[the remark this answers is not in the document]
rrahn: The iterator is an input iterator and does need to be copyable since C++20 anymore as long as it does not model CPP17Iterator, which is not required in this case. In fact, it is discouraged to copy input iterators in general because of the unclear behaviour. The stream iterators indeed store a copy of the element inside of the iterator. But that was not ideal before, when we had to think about possibly copying this iterator. I am not sure about the range library and if they updated the concepts already. In this case it would be fine to mark this iterator move-only and the iterator could store all information making the range a borrowed range. But that wasn't like so when I wrote the version and thus it was better to keep the data in the range and iterator just pointing to it in case someone was copying it for what ever reason. I believe a lot of stuff breaks once we enforce this change. And I think there is no general benefit of changing now everything. So, once we know that current input iterator concepts are not required to be copyable anymore we can make all input iterators move-only and the code above should be fine.

include/seqan3/search/algorithm/detail/search.hpp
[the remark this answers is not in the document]
rrahn: Ah I remember why. The config might have been changed within the call to search. But search returns another object which captures an invocable. This lambda. When I return that range the config created within the call of search goes out of scope and finally we have a dangling reference. So this must be copied here to make sure it can be used safely.

include/seqan3/search/algorithm/search_result_range.hpp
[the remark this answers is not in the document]
smehringer: > But it is not worht the effort to add complicated friend declarations just to safe a bool in a move-only iterator.

Yes I agree. I just didn't know that befriending is not transitive, thanks for the explanation.

include/seqan3/search/algorithm/search_result_range.hpp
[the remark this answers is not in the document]
smehringer: Ok, thanks for the explanation. I will make a card for this so we don't forget to check the concepts.

PR #1446, opened 2019-12-20
[FEATURE] Search dynamic mode config

### LINKED ISSUE #1436: Make `mode()` a runtime decision
(user feedback)

Make it a hybrid configuration, i.e. it can be either set at runtime or at compile time if known already. 
See https://github.com/seqan/seqan3/blob/master/include/seqan3/alignment/configuration/align_config_aligned_ends.hpp as a reference to implement both using almost the same interface.
Make the configuration options enums and then in addition define integral_constant values from the enums. Then the config can be called from the enum and the enum can be changed at runtime, or if the configuration is already known one can use the variable with the encoded enum.

```cpp
enum struct find : uint8_t
{
    best
};
inline constexpr std::integral_constant<find, find::best> find_best{};
...
search_cfg::mode(find::best)  // uses enum
search_cfg::mode(find_best)  // uses static type info
```

- [ ] think about a better name for find
- [ ] adapt the configuration to use enum
- [ ] tests
- [ ] adapt algorithm config
- [ ] adapt docu (API, tutorial)

rrahn:
Resolves #1436

### DIFF
--- a/include/seqan3/search/configuration/mode.hpp
+++ b/include/seqan3/search/configuration/mode.hpp
@@ -34,7 +34,6 @@ struct search_mode_all_best {};
 //!\brief Type for the "best" value for the configuration element "mode".
 //!\ingroup search_configuration
 struct search_mode_best {};
-
 } // namespace seqan3::detail
 
 namespace seqan3::search_cfg
@@ -61,8 +60,20 @@ struct strata : detail::strong_type<uint8_t, strata, detail::strong_type_skill::
     using detail::strong_type<uint8_t, strata, detail::strong_type_skill::convert>::strong_type;
 };
 
+//!\brief A helper variable template to check if the given mode type is a valid search mode.
+//!\ingroup search_configuration
+//!\hideinitializer
+template <typename search_mode_t>
+inline constexpr bool is_valid_search_mode_v = std::same_as<search_mode_t, detail::search_mode_all> ||
+                                               std::same_as<search_mode_t, detail::search_mode_all_best> ||
+                                               std::same_as<search_mode_t, detail::search_mode_best> ||
+                                               std::same_as<search_mode_t, strata>;
+
 /*!\brief Configuration element to determine the search mode.
  * \ingroup search_configuration
+ *
+ * \tparam search_mode_t The type of the search mode to use; seqan3::search_cfg::is_valid_mode_type must evaluate to
+ *                       `true` or must be of type seqan3::dynamic_state.
  *
  * \details
  *
@@ -92,11 +103,7 @@ struct strata : detail::strong_type<uint8_t, strata, detail::strong_type_skill::
  */
 template <typename search_mode_t>
 //!\cond
-    requires std::same_as<search_mode_t, detail::search_mode_all> ||
-             std::same_as<search_mode_t, detail::search_mode_all_best> ||
-             std::same_as<search_mode_t, detail::search_mode_best> ||
-             std::same_as<search_mode_t, strata> ||
-             std::same_as<search_mode_t, dynamic_state>
+    requires is_valid_search_mode_v<search_mode_t> || std::same_as<search_mode_t, dynamic_state>
 //!\endcond
 class mode : public pipeable_config_element<mode<search_mode_t>, search_mode_t>
 {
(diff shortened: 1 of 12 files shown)


test/unit/search/search_test.cpp
```diff
@@ -300,6 +300,13 @@ TYPED_TEST(search_test, search_strategy_all)
         configuration const cfg = max_error{total{1}} | mode{all};
         EXPECT_EQ(uniquify(search("ACGT"_dna4, this->index, cfg)), (hits_result_t{0, 1, 4, 5, 8, 9}));
     }
+
+    { // Test with dynamic mode.
+        mode dynamic_mode{};
```
joergi-w: If I remember correctly, in tests we now have to specify `seqan3::` namespace before all types and functions of seqan3?

include/seqan3/search/configuration/mode.hpp
```diff
@@ -75,28 +78,166 @@ struct strata : detail::strong_type<uint8_t, strata, detail::strong_type_skill::
...
+ * seqan3::search_cfg::mode configuration. Since, the same configuration element cannot be added multiple times a
+ * distinct type is used to forward the actual information. This means that implementors need to use this configuration
+ * instead of seqan3::search_cfg::mode when implementing the algorithm.
+ */
+template <typename search_mode_t>
+//!\cond
+    requires std::same_as<search_mode_t, search_mode_all> ||
+             std::same_as<search_mode_t, search_mode_all_best> ||
+             std::same_as<search_mode_t, search_mode_best> ||
+             std::same_as<search_mode_t, search_cfg::strata>
```
joergi-w: This is already the third occurrence of a check for any valid search mode in this file. Maybe it already makes sense to have a local shortcut like
```cpp
template <typename search_mode_t> 
bool is_search_mode = ...
```

test/unit/search/search_test.cpp
rrahn: yes, but I won't have this refactoring done for this PR, since it is about a feature. The other thing must be done separately.

rrahn:
@joergi-w I had to make a force push after I rebased onto master. So I also incorporated the changes already into the respective commits. I am sorry if this causes some inconvenience.

test/snippet/search/configuration_modes_dynamic.cpp
```diff
@@ -0,0 +1,19 @@
+#include <seqan3/search/configuration/all.hpp>
+
+int main()
+{
+    seqan3::search_cfg::mode my_mode{}; // Enables the dynamic configuration, defaults to seqan3::search_cfg::all.
+
+    if (true)
+        my_mode = seqan3::search_cfg::best; // The best mode is selected.
+    else
+        my_mode = seqan3::search_cfg::strata{2}; // The strata mode is selected.
```
joergi-w: The `else` branch is never executed...?

rrahn: yes, that is right. but that does not matter. It is a demo to show how to use it. And usually it would be by some condition you would do the one or the other thing. that is also more consistent with the text that describes this snippet.

joergi-w: Okay, fine then 👍

rrahn:
@smehringer polite ping

smehringer:
I did not manage to review this last week and since this is not an urgent issue and we need to redesign the config, doesn't it make more since to refactor first and then apply this change? Otherwise we need to refactor this code again?

rrahn:
> I did not manage to review this last week and since this is not an urgent issue and we need to redesign the config, doesn't it make more since to refactor first and then apply this change? Otherwise we need to refactor this code again?

Since it is not clear when we are going to refactor it, this should be already added. The other point is that the mechanics will be the same, just different class names are used. So I don't see that this can't be incrementally adapted later?

smehringer:
I thought we decided to postpone this altogether?

rrahn:
> I thought we decided to postpone this altogether?

I think this is a very independent PR. The search result range should wait until after the release because it involves more stuff. But anyway, we could also put that in after the release. I don't care. In general, I was trying to make a point why we should not postpone it until we start refactoring the things. I would like to do it first, when we arrive at an actual use case. But that is independent of the mechanics of this PR.

rrahn:
Will be tackled as part of the next iteration with the refined configuration naming.

PR #1447, opened 2019-12-22
[TEST] add seqan2 char predicate benchmarks

h-2:


test/performance/core/char_operations/char_predicate_benchmark.cpp
```diff
@@ -26,41 +42,63 @@ static void simple(benchmark::State & state)
...
+//  combined
+// ============================================================================
+
+template <tag id>
 static void combined(benchmark::State & state)
 {
     size_t sum = 0;
     size_t i = 0;
 
-    [[maybe_unused]] auto constexpr mycheck = is_punct || is_upper || is_digit;
```
smehringer: are you sure that storing a variable has no performance difference compared to using it directly?

h-2: Yep!

h-2:
> Can you post the result of the benchmark? :)

It's in master now so you can surprise yourself :)

PR #1448, opened 2019-12-22
[TEST,FIX] Several more benchmarks, minor optimisations

h-2:


h-2:
Jenkins Mac failures are unrelated.

joshuak94:
@h-2 I'm not sure what all of the seqan2 stuff is for... But I'm sure it makes sense!

test/include/seqan3/test/performance/sequence_generator.hpp
```diff
@@ -78,15 +78,21 @@ auto generate_sequence_seqan2(size_t const len = 500,
                               size_t const variance = 0,
                               size_t const seed = 0)
 {
+    size_t max_val;
+    if constexpr (std::unsigned_integral<alphabet_t>)
+        max_val = std::numeric_limits<size_t>::max();
```
rrahn: I can't follow this change. When alphabet_t is an integral then we always use size_t? But that would exceed the max value if alphabet_t is uint32_t? What am I a missing. Also in SeqAn `seqan::ValueSize<uint64_t>::VALUE` returns 0, which should be -1  which should be std::numeric_limits<uint64_t>::max(), or not? Could you elaborate?

test/include/seqan3/test/performance/sequence_generator.hpp
```diff
@@ -78,15 +78,21 @@ auto generate_sequence_seqan2(size_t const len = 500,
                               size_t const variance = 0,
                               size_t const seed = 0)
 {
+    size_t max_val;
```
rrahn:
```suggestion
    size_t max_val{};
```

test/performance/range/container_push_back_benchmark.cpp
```diff
@@ -92,13 +94,78 @@ BENCHMARK_TEMPLATE(push_back, seqan3::bitcompressed_vector, seqan3::aa27);
...
+BENCHMARK_TEMPLATE(push_back2, seqan::String, seqan::Packed<>, seqan::AminoAcid);
+// BENCHMARK_TEMPLATE(push_back2, seqan::String, seqan::Packed<>, seqan::Dna5Q); // broken in SeqAn2
+
+BENCHMARK_TEMPLATE(push_back2, seqan::String, seqan::Array<10'000>, seqan::Dna);
+BENCHMARK_TEMPLATE(push_back2, seqan::String, seqan::Array<10'000>, seqan::Dna5);
+BENCHMARK_TEMPLATE(push_back2, seqan::String, seqan::Array<10'000>, seqan::Iupac);
+BENCHMARK_TEMPLATE(push_back2, seqan::String, seqan::Array<10'000>, seqan::AminoAcid);
+BENCHMARK_TEMPLATE(push_back2, seqan::String, seqan::Array<10'000>, seqan::Dna5Q);
+
+#endif
```
rrahn:
```suggestion
#endif // SEQAN3_HAS_SEQAN2
```

test/performance/range/container_push_back_benchmark.cpp
```diff
@@ -92,13 +94,78 @@ BENCHMARK_TEMPLATE(push_back, seqan3::bitcompressed_vector, seqan3::aa27);
...
+void push_back2(benchmark::State & state)
+{
+    alphabet_t a{};
+
+    for (auto _ : state)
+    {
+        container_t<alphabet_t, spec_t> c;
+        for (size_t i = 0; i < 10'000; ++i)
+            seqan::appendValue(c, a);
+        a = seqan::back(c);
```
rrahn: new line before assignment

test/performance/range/container_push_back_benchmark.cpp
```diff
@@ -92,13 +94,78 @@ BENCHMARK_TEMPLATE(push_back, seqan3::bitcompressed_vector, seqan3::aa27);
...
+// ============================================================================
+
+#if SEQAN3_HAS_SEQAN2
+
+#include <seqan/sequence.h>
+
+template <template <typename...> typename container_t, typename spec_t, typename alphabet_t>
+void push_back2(benchmark::State & state)
+{
+    alphabet_t a{};
```
rrahn:
```suggestion
    alphabet_t alpha{};
```

test/performance/range/container_push_back_benchmark.cpp
```diff
@@ -92,13 +94,78 @@ BENCHMARK_TEMPLATE(push_back, seqan3::bitcompressed_vector, seqan3::aa27);
...
+
+    for (auto _ : state)
+    {
+        container_t<alphabet_t, spec_t> c;
```
rrahn:
```suggestion
        container_t<alphabet_t, spec_t> sequence;
```

test/performance/range/container_seq_read_benchmark.cpp
```diff
@@ -121,24 +128,107 @@ BENCHMARK_TEMPLATE(sequential_read, std::vector, seqan3::dna4, true);
...
+
+#if SEQAN3_HAS_SEQAN2
+
+#include <seqan/sequence.h>
+
+template <template <typename...> typename container_t, typename spec_t, typename alphabet_t, bool const_ = false>
+void sequential_read2(benchmark::State & state)
+{
+    auto cont_rando = seqan3::test::generate_sequence_seqan2<alphabet_t>(10'000, 0, 0);
+    container_t<alphabet_t, spec_t> source = cont_rando;
```
rrahn:
```suggestion
    container_t<alphabet_t, spec_t> sequence{seqan3::test::generate_sequence_seqan2<alphabet_t>(10'000, 0, 0)};
```

test/performance/range/container_seq_read_benchmark.cpp
```diff
@@ -121,24 +128,107 @@ BENCHMARK_TEMPLATE(sequential_read, std::vector, seqan3::dna4, true);
...
+
+    using source_ref_t = std::conditional_t<const_,
+                                            container_t<alphabet_t, spec_t> const &,
+                                            container_t<alphabet_t, spec_t> &>;
+
+    source_ref_t source_ref{source};
```
rrahn:
```suggestion
    source_ref_t sequence_ref{source};
```

test/performance/range/container_seq_read_benchmark.cpp
rrahn: See others as well.

test/performance/range/container_seq_write_benchmark.cpp
```diff
@@ -95,13 +95,64 @@ BENCHMARK_TEMPLATE(sequential_write, seqan3::bitcompressed_vector, seqan3::aa27)
...
+
+#if SEQAN3_HAS_SEQAN2
+
+#include <seqan/sequence.h>
+
+template <template <typename...> typename container_t, typename spec_t, typename alphabet_t>
+void sequential_write2(benchmark::State & state)
+{
+    auto cont_rando = seqan3::test::generate_sequence_seqan2<alphabet_t>(10'000, 0, 0);
+    container_t<alphabet_t, spec_t> source = cont_rando;
```
rrahn: could be simplified here as well.

test/performance/range/container_seq_write_benchmark.cpp
```diff
@@ -95,13 +95,64 @@ BENCHMARK_TEMPLATE(sequential_write, seqan3::bitcompressed_vector, seqan3::aa27)
...
+
+    alphabet_t a{};
```
rrahn: naming

test/performance/range/container_seq_write_benchmark.cpp
```diff
@@ -95,13 +95,64 @@ BENCHMARK_TEMPLATE(sequential_write, seqan3::bitcompressed_vector, seqan3::aa27)
...
+    for (auto _ : state)
+    {
+        for (auto && c : source)
+            c = a;
```
rrahn: DO_NOT_OPTIMIZE?

test/performance/range/container_seq_write_benchmark.cpp
```diff
@@ -95,13 +95,64 @@ BENCHMARK_TEMPLATE(sequential_write, seqan3::bitcompressed_vector, seqan3::aa27)
...
+template <template <typename...> typename container_t, typename spec_t, typename alphabet_t>
+void sequential_write2(benchmark::State & state)
+{
+    auto cont_rando = seqan3::test::generate_sequence_seqan2<alphabet_t>(10'000, 0, 0);
+    container_t<alphabet_t, spec_t> source = cont_rando;
+
+    alphabet_t a{};
+    for (auto _ : state)
+    {
+        for (auto && c : source)
```
rrahn:
```suggestion
        for (auto & c : source)
```
Otherwise the assignment might not be valid if indeed a temporary is returned. Which is never the case but the compiler error would point to the actual issue

test/performance/range/container_seq_write_benchmark.cpp
rrahn: It probably not necessary, but if it has side affects it will be the same as for the other benchmarks? Just a thought.

test/performance/range/views/view_comparison_seq_read_benchmark.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+// ============================================================================
+//  SeqAn3
+// ============================================================================
+
+template <typename rng_t, typename counter_t = size_t>
+void sequential_read_impl(benchmark::State & state, rng_t && range, counter_t && counter = 0)
+{
+    using alphabet_t = std::ranges::range_value_t<rng_t>;
+    alphabet_t a;
```
rrahn:
```suggestion
    alphabet_t alpha{};
```

test/performance/range/container_seq_write_benchmark.cpp
rrahn:
```suggestion
        for (auto && range_value : source)
```

test/performance/range/views/view_comparison_seq_read_benchmark.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+    for (auto _ : state)
+    {
+        for (auto && c : range)
```
rrahn:
```suggestion
        for (auto && range_value : range)
```

test/performance/range/views/view_comparison_seq_read_benchmark.cpp
```diff
@@ -0,0 +1,237 @@
...
+            benchmark::DoNotOptimize(a = c);
+
+        counter = 0;
+    }
+}
+
+template <template <typename> typename container_t, tag id>
+void sequential_read(benchmark::State & state)
+{
+    auto rando = seqan3::test::generate_sequence<seqan3::dna4>(10000, 0, 0);
```
rrahn:
```suggestion
    auto sequence = seqan3::test::generate_sequence<seqan3::dna4>(10000, 0, 0);
```
rando = 🐩 ?

test/performance/range/views/view_comparison_seq_read_benchmark.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+    container_t<seqan3::dna4> seq(rando.begin(), rando.end());
+
+    if constexpr (id == baseline)
+    {
+        sequential_read_impl(state, seq);
+    }
+    else if constexpr (id == slice)
+    {
+        auto long_rando = seqan3::test::generate_sequence<seqan3::dna4>(30000, 0, 0);
```
rrahn: long_rando = 🐕 ?

test/performance/range/views/view_filter_benchmark.cpp
```diff
@@ -0,0 +1,96 @@
...
+}
+
+template <typename fun_t>
+void loop_if(benchmark::State & state, fun_t fun)
+{
+    std::vector<seqan3::dna4> input = seqan3::test::generate_sequence<seqan3::dna4>(10000, 0, 0);
+
+    for (auto _ : state)
+    {
+        std::vector<seqan3::dna4> output;
```
rrahn:
```suggestion
        std::vector<seqan3::dna4> output{};
```

test/performance/range/views/view_filter_benchmark.cpp
```diff
@@ -0,0 +1,96 @@
...
+        for (seqan3::dna4 const c : input)
+            if (fun(c))
+                output.push_back(c);
+        output.clear();
```
rrahn: empty line before clear

test/performance/range/views/view_filter_benchmark.cpp
```diff
@@ -0,0 +1,96 @@
...
+void loop_view(benchmark::State & state, fun_t fun)
+{
+    std::vector<seqan3::dna4> input = seqan3::test::generate_sequence<seqan3::dna4>(10000, 0, 0);
+
+    for (auto _ : state)
+    {
+        std::vector<seqan3::dna4> output;
+        for (seqan3::dna4 const c : input | std::views::filter(fun))
+            output.push_back(c);
+        output.clear();
```
rrahn: empty line before clear

test/performance/range/views/view_filter_benchmark.cpp
```diff
@@ -0,0 +1,96 @@
...
+}
+
+template <typename fun_t>
+void loop_view(benchmark::State & state, fun_t fun)
+{
+    std::vector<seqan3::dna4> input = seqan3::test::generate_sequence<seqan3::dna4>(10000, 0, 0);
+
+    for (auto _ : state)
+    {
+        std::vector<seqan3::dna4> output;
```
rrahn:
```suggestion
        std::vector<seqan3::dna4> output{};
```

test/performance/range/views/view_filter_benchmark.cpp
```diff
@@ -0,0 +1,96 @@
...
+}
+
+template <typename fun_t>
+void copy_view(benchmark::State & state, fun_t fun)
+{
+    std::vector<seqan3::dna4> input = seqan3::test::generate_sequence<seqan3::dna4>(10000, 0, 0);
+
+    for (auto _ : state)
+    {
+        std::vector<seqan3::dna4> output;
```
rrahn:
```suggestion
        std::vector<seqan3::dna4> output{};
```

test/performance/range/views/view_filter_benchmark.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+#include <benchmark/benchmark.h>
+
+#include <seqan3/alphabet/all.hpp>
+#include <seqan3/std/ranges>
+#include <seqan3/test/performance/sequence_generator.hpp>
+
+/** This benchmarks evaluates views::filter against other forms of filtering **/
+
+auto const happy  = [] (seqan3::dna4 const c) { return c.to_rank() < 42; }; // always true
```
rrahn:
```suggestion
auto const always_true  = [] (seqan3::dna4 const c) { return c.to_rank() < 42; }; // always true
```

test/performance/range/views/view_filter_benchmark.cpp
```diff
@@ -0,0 +1,96 @@
...
+auto const rando  = [] (seqan3::dna4 const c) { return c.to_rank() > 1;  }; // true in 50% of the cases
+auto const sad    = [] (seqan3::dna4 const c) { return c.to_rank() > 42; }; // never true
```
rrahn:
```suggestion
auto const always_false    = [] (seqan3::dna4 const c) { return c.to_rank() > 42; }; // never true
```

test/performance/range/views/view_filter_benchmark.cpp
```diff
@@ -0,0 +1,96 @@
...
+#include <benchmark/benchmark.h>
+
+#include <seqan3/alphabet/all.hpp>
+#include <seqan3/std/ranges>
+#include <seqan3/test/performance/sequence_generator.hpp>
+
+/** This benchmarks evaluates views::filter against other forms of filtering **/
+
+auto const happy  = [] (seqan3::dna4 const c) { return c.to_rank() < 42; }; // always true
+auto const rando  = [] (seqan3::dna4 const c) { return c.to_rank() > 1;  }; // true in 50% of the cases
```
rrahn:
```suggestion
auto const half_true  = [] (seqan3::dna4 const c) { return c.to_rank() > 1;  }; // true in 50% of the cases
```

test/performance/range/views/view_comparison_seq_read_benchmark.cpp
```diff
@@ -0,0 +1,237 @@
...
+}
+
+BENCHMARK_TEMPLATE(sequential_read2, seqan::String, tag::baseline);
+BENCHMARK_TEMPLATE(sequential_read2, seqan::String, tag::slice);
+BENCHMARK_TEMPLATE(sequential_read2, seqan::String, tag::comp);
+BENCHMARK_TEMPLATE(sequential_read2, seqan::String, tag::rev);
+BENCHMARK_TEMPLATE(sequential_read2, seqan::String, tag::rev_comp);
+BENCHMARK_TEMPLATE(sequential_read2, seqan::String, tag::to_char);
+BENCHMARK_TEMPLATE(sequential_read2, seqan::String, tag::to_upper);
+#endif
```
rrahn:
```suggestion
#endif // SEQAN3_HAS_SEQAN2
```

test/performance/range/views/view_comparison_seq_read_benchmark.cpp
```diff
@@ -0,0 +1,237 @@
...
+#include <seqan3/alphabet/all.hpp>
+#include <seqan3/range/views/all.hpp>
+#include <seqan3/test/performance/sequence_generator.hpp>
+#include <seqan3/test/seqan2.hpp>
+
+enum tag
+{
+    baseline,
+    slice,
+    comp,
```
rrahn:
```suggestion
    complement,
```

test/performance/range/views/view_comparison_seq_read_benchmark.cpp
```diff
@@ -0,0 +1,237 @@
...
+    rev,
```
rrahn:
```suggestion
    reverse,
```

test/performance/range/views/view_comparison_seq_read_benchmark.cpp
```diff
@@ -0,0 +1,237 @@
...
+    rev_comp,
```
rrahn:
```suggestion
    reverse_complement,
```

test/performance/range/container_seq_write_benchmark.cpp
h-2: ~~I had a much longer answer that I deleted again.~~

Short and polite: "range_value" adds nothing in readability over "c". The line of code clearly states that "c" is a range element. It's like writing 
```c++
// This is a range-based-for-loop that loops over all elements of the range
```
before the line. Or writing:
```c++
for (int integer_counter = 0; integer_counter < 10; ++integer_counter)
```
For me these constructs make it harder to parse the code, because long names take inherently longer to read, they imply a certain complexity (complicated names are given to complicated things) or a meaning that is more global (it could even be read as a function parameter or a name from a namespace).

I agree that all names in the API should be well-reasoned and explanatory, but I think regulating naming down to this degree is -- politely -- a waste of everyone's time.

If there is strong consensus in the team that this is necessary, then I will change it and not spend more time discussing it, but since everyone always complains about the huge amounts of time that reviews take: maybe this is a good place to reduce "regulation".

test/performance/range/views/view_comparison_seq_read_benchmark.cpp
h-2: To illustrate my point from above: we have two entities in SeqAn3 that are called "complement": `seqan3::complement` and `seqan3::views::complement`. Calling the local entitiy by the same name *reduces* readability of the code.

test/performance/range/views/view_comparison_seq_read_benchmark.cpp
h-2: 🤣 

I agree 100% that this particular name is badly chosen, and I will definitely change it.

test/performance/range/views/view_filter_benchmark.cpp
h-2: "Happy" vs "sad" is a well-established term:
https://en.wikipedia.org/wiki/Happy_path

But I have no strong feelings about this.

h-2 (commented):
I am catching up with my stale PRs. Sorry if I have missed some of the policy changes and if I sound annoying for bringing up past discussions.

I will give my 2¢ for all things that I find strange in case other people share my sentiment, but I don't want all these things rediscussed per se. Whenever there is strong consensus on a policy, just say so, and I will change my code.

test/performance/range/container_seq_write_benchmark.cpp
marehr: Not agreeing here, I think the basic question in naming anything is always "What is the thing I have here." and we should explicitly write it out. Not making any exceptions. I can flip your argument, if you named your stuff better, we wouldn't need to  decipher what you mean and could drastically decrease the review time.

In this case, with only reading that single line, I don't know what is happening here. `for (auto && c : source)` is telling me nothing.

We have two cases for naming:

1) If we don't know anything about the domain, as in fully generic code like `for_each(range, f)`. `range_value` is more expressive than a single letter. For example `c` might be short for a character and if I read it, I don't know whether you didn't know your domain (because it is fully generic) or if you knew it and used an abbreviation. We could give it a rule to allow a single letter which has no meaning, like for example `x`, but we rarely write functions like that.

2) We know the domain. For example the range is a (biological) sequence and hence the range value type is a character / alphabet. In this case we had two naming schemes `for (auto && chr : sequence)` or `for (auto && alphabet : sequence)` (I'm fine with both of them with slight preference for the first.) I encountered some examples in the library where we used `for (auto && c : s)` and these don't age well, because you don't immediately know what is going on.

Summing it up `range_value` would suggest me that you either knew the domain and were lazy to think of a proper name, or it was generic and you gave it a generic name.

My personal goal is that we should always try to reduce communication errors as much as possible and in this case we can  reduce 4 possible error classes

* domain unknown / non-descriptive name, (e.g. `for (auto && c : range)`)
* domain unknown / descriptive name, (e.g. `for (auto && range_value : range)`, `for (auto && value : range)`)
* domain known / non-descriptive name, (e.g. `for (auto && c : s)`)
* domain known / descriptive name (e.g. `for (auto && chr : sequence)`)

to 2 possible error classes

* domain unknown / descriptive name, (e.g. `for (auto && range_value : range)`, `for (auto && value : range)`)
* domain known / descriptive name (e.g. `for (auto && chr : sequence)`)

Just by being always descriptive.

test/performance/range/views/view_filter_benchmark.cpp
marehr: never heard of it.

test/performance/range/container_seq_write_benchmark.cpp
h-2: > Otherwise the assignment might not be valid if indeed a temporary is returned. Which is never the case but the compiler error would point to the actual issue

Nono, we need this because some contaiers (e.g. bitcompressed) return proxies. The temporaries are materialised, that's the point of the proxy.

h-2: > My personal goal is that we should always try to reduce communication errors as much as possible 

Yes, I think we all want this. Just because someone disagrees on the "how" does not mean they are not interested in good communication or usability. But let's not drag this on for too long. I will change the names, but I seriously think this will make getting contributions more difficult as well as life harder for reviewers. Also, please add a guideline for this, so that contributors can look this up somewhere and it is not something that depends on the particular reviewer.

For the concrete names:
* `for (auto && range_value : range)` "range_value" is not typically a value but a reference. I will use `range_element` in the general case *if that is ok*.
* `for (auto && alphabet : sequence)` "alphabet" in SeqAn always refers to a type or gourp of types, and never to objects or references. -> there is no "sequence of alphabets" (unless you are iterating over type_list). If domain vocabulary is to be used (which I think is not bad in general), I will use `base : sequence`  *if that is ok*.

marehr: > I will change the names, but I seriously think this will make getting contributions more difficult as well as life harder for reviewers. Also, please add a guideline for this, so that contributors can look this up somewhere and it is not something that depends on the particular reviewer.

Contributors are in most cases just okay with what someone suggests, and they don't argue against them.

> "range_value" is not typically a value but a reference. I will use range_element in the general case if that is ok.

True, but every reference should be convertible to a "range_value". `range_element` is a completely new word that we never used until now, but I like it. I'm arguing with myself if just `element` would be good enough.

> I will use `base : sequence` if that is ok

Again, we never used that term before. And a `base` is, at least for me, a base class. I thought this benchmark is for every alphabet instance, and we used "chr" throughout the library to name it.

PR #1450, opened 2019-12-22
[FIX] type printing for references

h-2:
It was broken for const-ness and references which both weren't tested...

### DIFF
--- a/include/seqan3/core/detail/type_inspection.hpp
+++ b/include/seqan3/core/detail/type_inspection.hpp
@@ -43,7 +43,7 @@ namespace seqan3::detail
 template <typename type>
 inline std::string const type_name_as_string = [] ()
 {
-    std::string demangled_name;
+    std::string demangled_name{};
 #if defined(__GNUC__) || defined(__clang__) // clang and gcc only return a mangled name.
     using safe_ptr_t = std::unique_ptr<char, std::function<void(char *)>>;
 
--- a/include/seqan3/core/detail/type_inspection.hpp
+++ b/include/seqan3/core/detail/type_inspection.hpp
@@ -32,7 +32,7 @@ namespace seqan3::detail
  *
  * \tparam type The type to get the human-readable name for.
  *
- * \detail
+ * \details
  *
  * On gcc and clang std::type_info only returns a mangled name.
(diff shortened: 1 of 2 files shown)


include/seqan3/core/detail/type_inspection.hpp
```diff
@@ -41,18 +41,28 @@ namespace seqan3::detail
  * \note The returned name is implementation defined and might change between different tool chains.
  */
 template <typename type>
-inline static const std::string type_name_as_string = [] ()
+inline std::string const type_name_as_string = [] ()
 {
+    std::string ret;
```
rrahn:
```suggestion
    std::string demangled_name{};
```

rrahn:
[L33](https://github.com/seqan/seqan3/pull/1450/files#diff-89986d8c742bee39cd1e3524ecbf3cd5L33) uses the wrong doxygen command. This should be `\details` I am wondering why this wasn't detected before 🤔

include/seqan3/core/detail/type_inspection.hpp
h-2: I used `ret` throughout the library for the return value of lambda expressions, especially in the context of immediately evaluated lambdas used in the initialisation of variables.
But I can change it here, it's not important.

h-2:
> L35 uses the wrong doxygen command. This should be \details I am wondering why this wasn't detected before thinking

Fixed it while I was there.

include/seqan3/core/detail/type_inspection.hpp
```diff
@@ -41,18 +41,28 @@ namespace seqan3::detail
  * \note The returned name is implementation defined and might change between different tool chains.
  */
 template <typename type>
-inline static const std::string type_name_as_string = [] ()
+inline std::string const type_name_as_string = [] ()
 {
+    std::string demangled_name;
```
rrahn:
```suggestion
    std::string demangled_name{};
```

h-2: Our styleguide says:

> initialise all variables upon declaration, unless you really know what you are doing

I know what I am doing :laughing:

h-2:
Please squash this then.

PR #1451, opened 2019-12-22
[FIX] remove explicit noexcept from rule-of-5

h-2:
noexceptness depends on the value_type, because a value of the value_type is held as member. This also means that only value_types are supported that are default-constructible, maybe it would be better to wrap the member into a semiregular-optional-thingy.

I have made a card for this.

### DIFF
--- a/include/seqan3/range/views/async_input_buffer.hpp
+++ b/include/seqan3/range/views/async_input_buffer.hpp
@@ -203,16 +203,17 @@ public:
     //!\}
 
     /*!\name Construction, destruction and assignment
+     * \brief Not explicitly `noexcept` because this depends on construction/copy/... of value_type.
      * \{
      */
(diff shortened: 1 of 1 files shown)


marehr:
@sarahet Do you have time for this?

include/seqan3/range/views/async_input_buffer.hpp
```diff
@@ -203,16 +203,17 @@ class async_input_buffer_view<urng_t>::async_input_buffer_iterator
     //!\}
 
     /*!\name Construction, destruction and assignment
+     * \brief Not explicitly `noexcept` because this depends on construction/copy/... of value_type.
```
rrahn: Not sure, that comment helps. But it won't hurt either way I guess.

PR #1453, opened 2019-12-26
FastQ performance

h-2:
Based on an external application that converts FastQ to FastA I have benchmarked and optimised the affected code. Depending on System this yields a 2x speed-up but the application is still about 2x slower than SeqAn2.

I have spent a significant time trying to debug this with perf and Intel vtunes. My current understanding is that SeqAn2 is faster, because it doesn't use std::iostreams directly. Apparently it has its own layer in between that has larger streambuffers or it even does it's own low-level I/O, I am not sure. In any case what they are doing is faster and I don't think we can get there performance-wise with standard I/O streams.

Switching to the non-view implementation of reading FastQ brings some performance but is a lot more complicated, this is a real trade-off.
The team should discuss this and come up with a long-term strategy that answers the following questions:
  * How much slowdown over SeqAn2 is acceptable?
  * How much code overhead is acceptable to get better performance?
  * Should optimisations happen primarily on code that uses I/O (i.e. now views in the format) or should one try to get better low-level utilities that might allow using views on a higher level (i.e. larger streambuffers, OS-specific paging, MMAP-backed streams)?


h-2:
Oh, and I really suggest that *any new code needs extensive benchmarks*. 

Some things that might be helpful, maybe someone should write a wiki article:

  *  There seems to be huge difference in SeqAn3 between `-O3 -DNDEBUG` and `-O2 -g`. In my tests the difference in SeqAn2 was < 10% and in SeqAn3 is sometimes 100%. This makes profiling hard.
  * I see big differences between hardware, even without `-march=native`, e.g. the changes to chunking behaviour in FastA-output make almost no difference on my modern AMD machine, but reduce total program runtime by 25% on Redwood.

h-2:
I have removed the LIKELY /UNLIKELY macros again, because they decreased performance in all benchmarks on my system. Even when just putting SEQAN3_UNLIKELY around a condition *never* evaluates to true in the whole program runtime (i.e. those conditions that result in throwing exceptions in FastQ).
I don't know why, but it shows once again that it is always important to measure.

include/seqan3/io/record.hpp
```diff
@@ -246,16 +246,33 @@ struct record : detail::transfer_template_args_onto_t<field_types, std::tuple>
...
+    void clear() noexcept
     {
         clear_impl(*this, std::make_integer_sequence<size_t, field_types::size()>{});
     }
 private:
-    //!\brief Auxiliary function for clear().
+    //!\brief Auxiliary functions for clear().
     template <size_t ...Is>
-    constexpr void clear_impl(base_type & tup, std::integer_sequence<size_t, Is...> const &)
+    static constexpr void clear_impl(base_type & tup, std::integer_sequence<size_t, Is...> const &) noexcept
```
smehringer: is this really always noexcept? Since there might be aome type that has no clear but does memory allocation when being default constructed

include/seqan3/io/sequence_file/format_fasta.hpp
```diff
@@ -339,23 +343,55 @@ class format_fasta
               typename seq_type>
     void write_seq(stream_it_t & stream_it,
                    sequence_file_output_options const & options,
-                   seq_type && seq)
+                   seq_type & seq)
```
smehringer: why don't we take a temporary here anymore?

h-2:
> If you are fine with it I could cherry pick your first two commits, s.t. they can be merged already, or you do a separate PR?

Sure, go ahead!

include/seqan3/io/record.hpp
h-2: Good point. It's unlikely, but not impossible. noexcept should be conditional.

include/seqan3/io/sequence_file/format_fasta.hpp
h-2: Hm, it's debatable what the correct thing here is. My motivation in changing it was that these functions are not being forwarded to, so they never receive temporaries. But the general policy was that we always use `type &&` for out-parameters with type deduction so we might as well keep it.

h-2:
I removed the code that was already merged (or discussed and discarded). The current state builds again against master, but I think @marehr was also working on this so this is currently just a draft open to discussion.

SGSSGene:
@smehringer  will overtake this PR and check if this is still an performance improvement.

eseiler:
I noticed some improvement:

```
Benchmark                             Before[M/s]  After[M/s]
-------------------------------------------------------------
fastq_read_from_stream_seqan3/100     415.09       456.501
fastq_read_from_stream_seqan3/1000    409.358      473.763
fastq_read_from_stream_seqan3/10000   410.244      464.949
fastq_read_from_disk_seqan3/100       191.171      203.07
fastq_read_from_disk_seqan3/1000      263.259      316.307
fastq_read_from_disk_seqan3/10000     282.857      334.441
```

I did compile with `-O3 -march=native -DNDEBUG`

smehringer:
hm.. you think we should merge it? I have no strong feelings. I just had a 3% speedup but its more than nothing of course.

eseiler:
Depending on the use case, it's almost 20%, no?

PR #1460, opened 2020-01-06
[FIX] Make @HD line in SAM format optional.

### LINKED ISSUE #1457: Exception on opening example bam file
This file:
https://figshare.com/articles/Example_BAM_file/1460736

triggers:
```
terminate called after throwing an instance of 'seqan3::format_error'
  what():  Illegal SAM header tag in @HG starting with:L
```

According to samtools view, the entire header is:
```
@SQ     SN:1    LN:30427671
@SQ     SN:2    LN:19698289
@SQ     SN:3    LN:23459830
@SQ     SN:4    LN:18585056
@SQ     SN:5    LN:26975502
@SQ     SN:M    LN:366924
@SQ     SN:C    LN:154478
```

And according to the SAM spec this is very valid, it gives the following example itself: `@SQ SN:ref LN:45`...

smehringer:
Resolves #1457

To the diff: I reallt just transolcated the parsing of the header inside the while loop.. but translocations and git don't go together nicely :woman_shrugging:

### DIFF
--- a/include/seqan3/io/alignment_file/format_sam_base.hpp
+++ b/include/seqan3/io/alignment_file/format_sam_base.hpp
@@ -453,45 +453,43 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
         read_field(stream_view | views::take_until_or_throw(is_char<'\t'> || is_char<'\n'>), value);
     };
 
-    // @HQ line
-    // -------------------------------------------------------------------------------------------------------------
-    parse_tag_value(hdr.format_version); // parse required VN (version) tag
-
-    // The SO, SS and GO tag are optional and can appear in any order
-    while (is_char<'\t'>(*std::ranges::begin(stream_view)))
-    {
-        std::ranges::next(std::ranges::begin(stream_view));              // skip tab
-        std::string * who = std::addressof(hdr.grouping);
-
-        if (is_char<'S'>(*std::ranges::begin(stream_view)))
-        {
-            std::ranges::next(std::ranges::begin(stream_view));          // skip S
-
-            if (is_char<'O'>(*std::ranges::begin(stream_view)))          // SO (sorting) tag
-                who = std::addressof(hdr.sorting);
-            else if (is_char<'S'>(*std::ranges::begin(stream_view)))     // SS (sub-order) tag
-                who = std::addressof(hdr.subsorting);
-            else
-                throw format_error{std::string{"Illegal SAM header tag: S"} +
-                                   std::string{static_cast<char>(*std::ranges::begin(stream_view))}};
-        }
-        else if (!is_char<'G'>(*std::ranges::begin(stream_view)))        // GO (grouping) tag
-        {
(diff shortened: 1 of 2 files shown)


joergi-w:
> To the diff: I reallt just transolcated the parsing of the header inside the while loop.. but translocations and git don't go together nicely 🤷‍♀

I want to point out a trick for that: The git diff is so bad, because you had to indent the loop and all lines actually differ. Via the  ⚙️ button you can hide whitespace changes in the split view and then it is very easy to spot the actual code changes!

smehringer:
> I want to point out a trick for that: The git diff is so bad, because you had to indent the loop and all lines actually differ. Via the :gear: button you can hide whitespace changes in the split view and then it is very easy to spot the actual code changes!

Great thanks!

include/seqan3/io/alignment_file/format_sam_base.hpp
```diff
@@ -453,45 +453,43 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
-                who = std::addressof(hdr.sorting);
-            else if (is_char<'S'>(*std::ranges::begin(stream_view)))     // SS (sub-order) tag
-                who = std::addressof(hdr.subsorting);
-            else
-                throw format_error{std::string{"Illegal SAM header tag: S"} +
-                                   std::string{static_cast<char>(*std::ranges::begin(stream_view))}};
-        }
-        else if (!is_char<'G'>(*std::ranges::begin(stream_view)))        // GO (grouping) tag
+        if (is_char<'H'>(*std::ranges::begin(stream_view))) // HD (header) tag
         {
```
rrahn: where do you consume the D of the HD header tag?

include/seqan3/io/alignment_file/format_sam_base.hpp
```diff
@@ -453,45 +453,43 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
         read_field(stream_view | views::take_until_or_throw(is_char<'\t'> || is_char<'\n'>), value);
     };
 
-    // @HQ line
```
rrahn: why removing this? Of course after fixing it to HD.

include/seqan3/io/alignment_file/format_sam_base.hpp
smehringer: It is done in the lambda `parse_tag_value` that always skips everythin until `:` which in case of the first tag in a header line is `@H[D    VN]:` and for every following tag just the tag name.

include/seqan3/io/alignment_file/format_sam_base.hpp
smehringer: 1) Because I do not now if the header comes first 2) It is now documented in line 460

smehringer:
> One general item. Have you checked, that all top-level header tags are optional? As far as I can see only some of the inner fields are required, but only if the respective header information is present.

Yes, I checked again and all of the header lines are optional and no order is presumed between individual header lines (at least I did not find anything in the specs). Wiithin a header line there are required fields an a required order.

PR #1461, opened 2020-01-06
[FIX] use insert_gap correctly

marehr:
use `insert_gap` in alignment_trace_algorithms correctly since it didn't work for `gap_decorators`

This may be an example for a `[[nodiscard]]` attribute :)

### DIFF
--- a/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
+++ b/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
@@ -423,8 +423,8 @@ std::ranges::iterator_t<range_type> insert_gap(range_type & rng,
  * \tparam range_type   Type of the range to modify; must have an erase_gap(it) member function.
  * \param[in,out] rng   The range to modify.
  * \param[in] pos_it    The iterator pointing to the position where to erase one gap.
- * \returns       An iterator following the removed element. If the iterator `pos_it` refers to the last element, the
- *                std::ranges::end() iterator is returned.
+ * \returns An iterator following the removed element. If the iterator `pos_it` refers to the last element, the
+ *          std::ranges::end() iterator is returned.
  *
  * \details
  *
--- a/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
+++ b/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
@@ -233,11 +233,14 @@ SEQAN3_CONCEPT aligned_sequence =
  *                                The value type must be a seqan3::gapped alphabet.
  * \param[in,out] aligned_seq     The aligned container to modify.
  * \param[in]     pos_it          The iterator pointing to the position where to insert a gap.
+ * \returns       std::ranges::iterator_t<aligned_seq_t> An iterator pointing to the inserted gap.
  *
  * \details
  *
  * This function delegates to the member function `insert(iterator, value)` of
  * the container.
+ *
+ * \note This may cause reallocations and thus invalidate all iterators and references. Use the returned iterator.
  */
 template <sequence_container aligned_seq_t>
 //!\cond
@@ -256,11 +259,15 @@ inline typename aligned_seq_t::iterator insert_gap(aligned_seq_t & aligned_seq,
  * \param[in,out] aligned_seq     The aligned container to modify.
  * \param[in]     pos_it          The iterator pointing to the position where to insert gaps.
  * \param[in]     size            The number of gap symbols to insert (will result in a gap of length `size`).
+ * \returns       std::ranges::iterator_t<aligned_seq_t> An iterator pointing to the first inserted gap or `pos_it` if
+ *                `size == 0`.
  *
  * \details
  *
  * This function delegates to the member function `insert(iterator, `size`, `value`)`
  * of the container.
+ *
+ * \note This may cause reallocations and thus invalidate all iterators and references. Use the returned iterator.
  */
 template <sequence_container aligned_seq_t>
 //!\cond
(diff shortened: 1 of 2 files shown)


smehringer:
> This may be an example for a `[[nodiscard]]` attribute :)

But you may want to insert a gap and that's it. That way you don't need the return value to continue iterating over the decorator. So I think this is not necessarily `[[nodiscard]]`.

include/seqan3/alignment/matrix/alignment_trace_algorithms.hpp
```diff
@@ -136,12 +136,12 @@ inline alignment_t alignment_trace(database_t && database,
         {
             coordinate.col = std::max<size_t>(coordinate.col, 1) - 1;
             --end_aligned_db;
-            insert_gap(std::get<1>(aligned_seq), end_aligned_qy);
+            end_aligned_qy = insert_gap(std::get<1>(aligned_seq), end_aligned_qy);
```
joergi-w: The return value of `insert_gap` is unfortunately not documented... can it possibly return a different value than the given `end_aligned_qy` parameter? 
If no, then this assignment is a no-op which is rather confusing than beneficial. 
If yes, why do we need to touch both db and qy variables while going left in the traceback? Also, then you might need to check if the returned value is still in the valid range.

joergi-w: After @smehringer explained to me the meaning of the return value, I'm sure your change is good! Please add the `\returns` documentation for `insert_gap` instead.

marehr:
@joergi-w I added documentation to the interfaces.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -113,56 +113,71 @@ namespace seqan3
...
-/*!\fn      inline aligned_seq_t::iterator erase_gap(aligned_seq_t & aligned_seq, typename aligned_seq_t::const_iterator
- *          pos_it)
+/*!\fn      inline std::ranges::iterator_t<aligned_seq_t> erase_gap(aligned_seq_t & aligned_seq,
+ *          typename aligned_seq_t::const_iterator pos_it)
  * \brief   Erase a seqan3::gap from an aligned sequence.
  *
  * \tparam        aligned_seq_t   Type of the range to modify; must model seqan3::aligned_sequence.
  * \param[in,out] aligned_seq     The aligned sequence to modify.
  * \param[in]     pos_it          The iterator pointing to the position where to erase a gap.
+ * \returns       An Iterator following the removed element. If the iterator `pos_it` refers to the last element, the
```
smehringer:
```suggestion
 * \returns       An iterator following the removed element. If the iterator `pos_it` refers to the last element, the
```

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -113,56 +113,71 @@ namespace seqan3
...
+ *                std::ranges::end() iterator is returned.
  *
  * \throws seqan3::gap_erase_failure if there is no seqan3::gap at \p pos_it.
  *
  * \details
+ * \note      This may cause reallocations and thus invalidate all iterators and references. Use the returned iterator.
```
smehringer:
```suggestion
 * \note      This may cause reallocations and thus invalidates all iterators and references. Use the returned iterator.
```

grep for this everywhere else too!

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -379,32 +411,36 @@ template <typename range_type>
...
 }
 
 /*!\brief An implementation of seqan3::aligned_sequence::erase_gap for ranges with the corresponding
  *        member function erase_gap(it).
  * \ingroup aligned_sequence
  * \tparam range_type   Type of the range to modify; must have an erase_gap(it) member function.
  * \param[in,out] rng   The range to modify.
- * \param[in] it        The iterator pointing to the position where to erase one gap.
+ * \param[in] pos_it    The iterator pointing to the position where to erase one gap.
+ * \returns       An Iterator following the removed element. If the iterator `pos_it` refers to the last element, the
```
smehringer: :nail_care:  The spacing is weird. It is neither aligned to the others nor removed?

marehr: What would you propose?

smehringer: I would actually remove the indentation. I was in favor of aligning everything for quite some time but it always leads to more line changes when refactoring because the indentation needs to be adapted too.

marehr: I just changed this one instance. If you think more indentations are weird can you annotate them?

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -414,12 +450,16 @@ std::ranges::iterator_t<range_type> erase_gap(range_type & rng,
  * \param[in,out] rng   The range to modify.
  * \param[in] first     The iterator pointing to the position where to start erasing gaps.
  * \param[in] last      The iterator pointing to the position where to stop erasing gaps.
+ * \returns       An iterator following the last removed element. If the iterator `last` refers to the last element, the
```
smehringer: Another weird indentation :D

marehr:
the third time is the charm :laughing:

smehringer:
ping me if I the CI is done

marehr:
@smehringer ping

PR #1467, opened 2020-01-07
Makes the exception names of the argument parser consistent

### LINKED ISSUE #481: [Argument Parser] make the exception names consistent
There should be one parser_exception that all the others inherit from and none has `parse_` in its name

smehringer:
Resolves #481

I changed the names according to our resolution in the strategy meeting ([card](https://github.com/orgs/seqan/projects/4#card-26600178))
Additionally, the first commit get's rid of the internal exception in the function retrive_value. The success is indicated by a return value instead and the exception is thrown outside. This also made the exceptions std::type_conversion-error and std::overflow_errow obsolete which lowers the number of exceptions in general.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -152,6 +152,13 @@ Note that 3.1.0 will be the first API stable release and interfaces in this rele
 * The member type that denotes which arguments a `validator` can validate has been renamed from `value_type` to
   `option_value_type`
   ([\#1394](https://github.com/seqan/seqan3/pull/1394)).
+* Some exception names were altered and some removed ([\#1394](https://github.com/seqan/seqan3/pull/1467)):
+  * The exception seqan3::parser_invalid_argument was renamed to seqan3::argument_parser_error.
+  * The exception seqan3::validation_failed was renamed to seqan3::validation_error.
+  * The exception seqan3::parser_design_error was renamed to seqan3::design_error and also inherits from
+    seqan3::argument_parser_error.
+  * The exception seqan3::type_conversion_failed was deprecated, you can catch seqan3::user_input_error instead.
+  * The exception seqan3::overflow_error_on_conversion was deprecated, you can catch seqan3::user_input_error instead.
 
 #### Build system
 
--- a/include/seqan3/argument_parser/argument_parser.hpp
+++ b/include/seqan3/argument_parser/argument_parser.hpp
@@ -114,10 +114,10 @@ namespace seqan3
  * Developer errors are those that violate the seqan3::argument_parser design
  * (e.g. calling the seqan3::argument_parser::parse function twice or specifying
  * two different options with the same identifier.)
- * In this case, a  seqan3::parser_design_error is thrown.
+ * In this case, a seqan3::design_error is thrown.
  *
  * The second kind are user errors, due to invalid command line calls. In this
- * case a seqan3::parser_invalid_argument is thrown.
+ * case a seqan3::argument_parser_error is thrown.
  *
  * For example:
  *
@@ -162,7 +162,7 @@ public:
     argument_parser(argument_parser &&) = default;                  //!< Defaulted.
     argument_parser & operator=(argument_parser &&) = default;      //!< Defaulted.
 
-    /*!\brief Initializes an argument_parser object from the command line arguments.
+    /*!\brief Initializes an seqan3::argument_parser object from the command line arguments.
      *
      * \param[in] app_name       The name of the app that is displayed on the help page.
      * \param[in] argc           The number of command line arguments.
@@ -170,7 +170,7 @@ public:
      * \param[in] version_check  Notify users about app version updates (default true).
      * \param[in] subcommands    A list of subcommands (see \link subcommand_arg_parse subcommand parsing \endlink).
      *
-     * \throws seqan3::parser_design_error if the application name contains illegal characters.
+     * \throws seqan3::design_error if the application name contains illegal characters.
      *
      * The application name must only contain alpha-numeric characters, '_' or '-',
      * i.e. the following regex must evaluate to true: `\"^[a-zA-Z0-9_-]+$\"`.
@@ -186,11 +186,11 @@ public:
         version_check_dev_decision{version_check}
     {
         if (!std::regex_match(app_name, app_name_regex))
-            throw parser_design_error{"The application name must only contain alpha-numeric characters "
+            throw design_error{"The application name must only contain alpha-numeric characters "
                                                "or '_' and '-' (regex: \"^[a-zA-Z0-9_-]+$\")."};
         for (auto & sub : subcommands)
             if (!std::regex_match(sub, std::regex{"^[a-zA-Z0-9_]+$"}))
-                throw parser_design_error{"The subcommand name must only contain alpha-numeric characters or '_'."};
+                throw design_error{"The subcommand name must only contain alpha-numeric characters or '_'."};
 
         info.app_name = std::move(app_name);
         init(argc, argv, std::move(subcommands));
@@ -226,7 +226,7 @@ public:
      * \param[in]  spec      Advanced option specification, see seqan3::option_spec.
      * \param[in]  validator The validator applied to the value after parsing (callable).
      *
-     * \throws seqan3::parser_design_error
+     * \throws seqan3::design_error
      */
     template <typename option_type, validator validator_type = detail::default_validator<option_type>>
     //!\cond
@@ -242,7 +242,7 @@ public:
                     validator_type validator = validator_type{}) // copy to bind rvalues
     {
         if (sub_parser != nullptr)
-            throw parser_design_error{"You may only specify flags for the top-level parser."};
+            throw design_error{"You may only specify flags for the top-level parser."};
 
         verify_identifiers(short_id, long_id);
         // copy variables into the lambda because the calls are pushed to a stack
@@ -274,7 +274,7 @@ public:
      *
      * \tparam option_type Must have a formatted input function (stream >> value).
      *                     If option_type is a container, its value type must have the
-     *                     formateted input function (exception: std::string is not
+     *                     formatted input function (exception: std::string is not
      *                     regarded as a container).
      *                     See <a href="https://en.cppreference.com/w/cpp/concept/FormattedInputFunction"> FormattedInputFunction </a>.
      * \tparam validator_type The type of validator to be applied to the option
@@ -284,7 +284,7 @@ public:
      * \param[in]  desc      The description of the positional option to be shown in the help page.
      * \param[in]  validator The validator applied to the value after parsing (callable).
      *
-     * \throws seqan3::parser_design_error
+     * \throws seqan3::design_error
      *
      * \details
      *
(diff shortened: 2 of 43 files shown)


include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -285,84 +292,80 @@ class format_parse : public format_base
...
-     * \param[out] value Stores the casted value.
-     * \param[in]  in    The input argument to be casted.
-     *
-     * \throws seqan3::type_conversion_failed
-     * \throws seqan3::overflow_error_on_conversion
+     * \tparam option_t The optiona value type; must model seqan3::arithmetic.
+     * \param[out] value Stores the cast value.
+     * \param[in] in The input argument to be cast.
+     * \returns seqan3::option_cast_result::type_conversion_error if \p in could not be converted to an arithmetic type
+     *          via std::from_chars, seqan3::option_cast_result::overflow_error if \p in could be converted but the
```
MitraDarja: What does \p stand for?
Do we always shorten \param[in] with \p in?

include/seqan3/argument_parser/exceptions.hpp
```diff
@@ -19,6 +19,17 @@
 namespace seqan3
 {
 
+//!\brief This class is eprecated.
```
MitraDarja: deprecated

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -285,84 +292,80 @@ class format_parse : public format_base
 
     /*!\brief Tries to cast an input string into a value.
      * \tparam option_t Must model seqan3::input_stream_over.
-     * \param[out] value Stores the casted value.
-     * \param[in]  in    The input argument to be casted.
-     *
-     * \throws seqan3::type_conversion_failed
+     * \param[out] value Stores the cast value.
+     * \param[in] in The input argument to be cast.
+     * \returns seqan3::option_cast_result::type_conversion_error if \p in could not be converted via the stream
```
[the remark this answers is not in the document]
smehringer: With \p doxygen just formats the next word in code style. I will change it to back ticks ` which isused more frequently

smehringer:
sorry I did not see that you approved already. You can or cannot check the differences and approve again. In general if you find typos, you can request changes as they are just as valid as code changing requests.

include/seqan3/argument_parser/exceptions.hpp
```diff
@@ -133,13 +134,13 @@ class validation_failed : public parser_invalid_argument
  * - Reuse of a short or long identifier (must be unique)
  * - Both identifiers must not be empty (one is ok)
  */
-class parser_design_error : public std::logic_error
+class design_error : public std::logic_error
```
rrahn: in the Changelog you mentioned, that this inherits from argument_parser_error, which I would prefer. I also would not inherit the argument_parser_error from invalid_argument but make it it's own exception.

include/seqan3/argument_parser/exceptions.hpp
```diff
@@ -34,93 +45,83 @@ namespace seqan3
...
     /*!\brief The constructor.
      * \param[in] s The error message.
      */
-    overflow_error_on_conversion(std::string const & s) : parser_invalid_argument(s) {}
+    option_declared_multiple_times(std::string const & s) : argument_parser_error(s) {}
 };
 
 //!\brief Argument parser exception thrown when an argument could not be casted to the according type.
-class validation_failed : public parser_invalid_argument
+class validation_error : public argument_parser_error
```
rrahn: So if I get this right. When people use one of this exceptions they will get an compiler error because it is no longer present. When they used parser_invalid_argument they will get a deprecation message but also thrown exceptions are never catched and the program terminates. So which means we have two options. Either deprecate everything or let the compiler always fail when using old exceptions. I am tending for the former solution. Or am I missing something here?

CHANGELOG.md
```diff
@@ -54,6 +54,13 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
   the respective instances (e.g. change `input_file_validator` to `input_file_validator<>`).
 * The member type that denotes which arguments a `validator` can validate has been renamed from `value_type` to
   `option_value_type`.
+* **Some exception names were altered and some removed**
+  * The exception seqan3::parser_invalid_argument was renamed to seqan3::argument_parser_error.
+  * The exception seqan3::validation_failed was renamed to seqan3::validation_error.
+  * The exception seqan3::parser_design_error was renamed to seqan3::design_error and also inherits from
+    seqan3::argument_parser_error.
+  * The exception seqan3::type_conversion_failed was removed, you can catch seqan3::user_input_error instead.
```
rrahn: should then be deprecated.

CHANGELOG.md
```diff
@@ -54,6 +54,13 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
...
+  * The exception seqan3::overflow_error was removed, you can catch seqan3::user_input_error instead.
```
rrahn: should then be deprecated.

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -285,84 +292,80 @@ class format_parse : public format_base
...
         requires input_stream_over<std::istringstream, typename container_option_t::value_type>
     //!\cond
-    void retrieve_value(container_option_t & value, std::string const & in)
+    option_cast_result cast_option_value(container_option_t & value, std::string const & in)
     {
-        typename container_option_t::value_type tmp;
+        typename container_option_t::value_type tmp{};
 
-        retrieve_value(tmp, in); // throws on failure
+        auto res = cast_option_value(tmp, in); // throws on failure
```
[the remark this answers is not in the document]
smehringer: Hm.. So when there is a conversion error the tmp variable will not be altered so I am appending the default constructed value. So this is not an issue but probably it is better to not append anything at all.

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -285,84 +292,80 @@ class format_parse : public format_base
...
     //!\endcond
 
-    /*!\brief Appends a casted value to its container.
-     *
-     * \tparam container_option_t Must satisfy the seqan3::sequence_container and
-     *                            its value_type must satisfy the seqan3::input_stream_over
-     *
-     * \param[out] value container that stores the casted value.
-     * \param[in]  in    The input argument to be casted.
+    /*!\brief Appends a cast value to its container.
```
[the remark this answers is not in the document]
smehringer: I talked briefly with @marehr about this and decided to switch from `cast` (which I agree is not appropiate) to `parse` everywhere in the documentation.

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -406,22 +408,58 @@ class format_parse : public format_base
...
+    template <typename option_type>
+    void throw_on_input_error(option_cast_result const res,
+                              std::string const & option_name,
+                              std::string const & input_value)
+    {
+        std::string msg{"Value cast failed for " + option_name + ": "};
+
+        if (res == option_cast_result::conversion_error)
+        {
+            throw user_input_error{msg + "Argument " + input_value + " could not be cast to type " +
```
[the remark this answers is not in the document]
smehringer: see above -> `parsed`

include/seqan3/argument_parser/exceptions.hpp
smehringer: > in the Changelog you mentioned, that this inherits from argument_parser_error, which I would prefer.

Good catch

> I also would not inherit the argument_parser_error from invalid_argument but make it it's own exception.

You mean it should inherit from std::exception instead of std::invalid_argument?

include/seqan3/argument_parser/exceptions.hpp
smehringer: Alright, I will deprecate all the old names and we can remove them with 3.1

smehringer:
Sorry, I had to rebase and add a different commit for the renaming in the documentation to keep the history clean. Doing that later would have been too compilcated.

include/seqan3/argument_parser/exceptions.hpp
rrahn: Yes, because design_error is not an invalid_argument is it? I would think it is an own error for itself.

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -162,15 +162,15 @@ class argument_parser
     argument_parser(argument_parser &&) = default;                  //!< Defaulted.
     argument_parser & operator=(argument_parser &&) = default;      //!< Defaulted.
 
-    /*!\brief Initializes an argument_parser object from the command line arguments.
+    /*!\brief Initializes seqan3::an argument_parser object from the command line arguments.
```
rrahn:
```suggestion
    /*!\brief Initializes an seqan3::argument_parser object from the command line arguments.
```

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -406,22 +413,58 @@ class format_parse : public format_base
...
+    template <typename option_type>
+    void throw_on_input_error(option_parse_result const res,
+                              std::string const & option_name,
+                              std::string const & input_value)
+    {
+        std::string msg{"Value parse failed for " + option_name + ": "};
+
+        if (res == option_parse_result::error)
+        {
+            throw user_input_error{msg + "Argument " + input_value + " could not be cast to type " +
```
rrahn:
```suggestion
            throw user_input_error{msg + "Argument " + input_value + " could not be parsed as type " +
```

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -283,86 +290,87 @@ class format_parse : public format_base
...
     //!\cond
-    void retrieve_value(container_option_t & value, std::string const & in)
+    option_parse_result parse_option_value(container_option_t & value, std::string const & in)
     {
-        typename container_option_t::value_type tmp;
+        typename container_option_t::value_type tmp{};
 
-        retrieve_value(tmp, in); // throws on failure
-        value.push_back(tmp);
+        auto res = parse_option_value(tmp, in); // throws on failure
```
rrahn:
```suggestion
        auto res = parse_option_value(tmp, in);
```

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -406,22 +413,58 @@ class format_parse : public format_base
...
             value = false;
         else
-            throw type_conversion_failed("Argument '" + in + "' could not be casted to boolean.");
+            return option_parse_result::error;
+
+        return option_parse_result::success;
+    }
+
+    /*!\brief Tries to parse an input string into boolean value.
+     * \param[in] res A result value of parseing an input string to the respective option value type.
```
rrahn:
```suggestion
     * \param[in] res A result value of parsing an input string to the respective option value type.
```

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -406,22 +413,58 @@ class format_parse : public format_base
...
+     * \param[in] option_name The name of the option who's input was parsed.
```
rrahn:
```suggestion
     * \param[in] option_name The name of the option whose input was parsed.
```
?

rrahn:
Have you verified the code coverage issue?

smehringer:
It has nothing to do with me. I'll rebase to current master to see if this fixes it.

smehringer:
@rrahn Code coverage only fails because of the filesystem errors that in the file validators were renamed but are not tested. So I think this can be merged.

smehringer:
@rrahn I updated the Changelog entry to link this PR.
The code coverage fail is unrelated to this PR, since it records 4 misses that are due to the untested error handling of the `input_directory_validator`s where I changed the exception names so I trigger the misses again. Fixes the code coverage should be part of another PR, I'll make a card.

PR #1468, opened 2020-01-08
[Alignment IO] Also test reading and writing hard clipping at the end of the cigar string.

### LINKED ISSUE #1116: Alignment File cannot handle Hard clipping at the end
Just so I don't forget

smehringer:
Resolves #1116 

I created this issue a while ago but I can not reproduce the error. I added tests to show that hard clipping at the end of the cigar string is not an issue for SAM or BAM. This was either fixed along the way or was never a problem..


test/unit/io/alignment_file/format_sam_test.cpp
```diff
@@ -33,8 +33,8 @@ R"(@HD	VN:1.6	SO:coordinate	SS:coordinate:queryname	GO:none
 R"(@HD	VN:1.6
 @SQ	SN:ref	LN:34
 read1	41	ref	1	61	1S1M1D1M1I	ref	10	300	ACGT	!##$	AS:i:2	NM:i:7
-read2	42	ref	2	62	1H7M1D1M1S	=	10	300	AGGCTGNAG	!##$&'()*	xy:B:S,3,4,5
-read3	43	ref	3	63	1S1M1P1M1I1M1I1D1M1S	=	10	300	GGAGTATA	!!*+,-./
+read2	42	ref	2	62	1H7M1D1M1S2H	ref	10	300	AGGCTGNAG	!##$&'()*	xy:B:S,3,4,5
```
smehringer: To the reviewer: Note that `=` in SAM format means "equal to the reference name given at the front of the record". So my change has no semantic effect! I needed to change it to be able to use this string as a comparison when writing the cigar string.
Handling `=` is also tested in a different part of the tests so no code coverage is dropped.

test/unit/io/alignment_file/alignment_file_format_test_template.hpp
```diff
@@ -545,7 +547,7 @@ TYPED_TEST_P(alignment_file_write, cigar_vector)
     }
 
     this->ostream.flush();
-    EXPECT_EQ(this->ostream.str(), this->simple_three_reads_output);
+    EXPECT_EQ(this->ostream.str(), this->simple_three_reads_input);
```
joergi-w: Why this change to input? You want to write the alignment file, so it should be output...

test/unit/io/alignment_file/format_sam_test.cpp
joergi-w: Thanks for the explanation :)

test/unit/io/alignment_file/alignment_file_format_test_template.hpp
smehringer: Because in the case that I write an alignment directly, or I read directly write, hard clipping is not preserved in the cigar string (because it cannot be represented in a pair of aligned sequences). In the case of this test, hard clipping is preserved when writing because I write a cigar vector (not an alignment), so I need to compare to the original input instead of the output string.

I will drop a note in the test :)

joergi-w: Thanks 👍

PR #1471, opened 2020-01-08
Update range library to 0.10.0

### LINKED ISSUE #1479: build fails: latest seqan3 dev + rangv3 0.10 on Debian dev
## Platform

  - SeqAn version:
Using https://github.com/seqan/seqan3/commit/b42671eac39ba0b47c054c4e9767268840dbbdca and range-v3 0.10.0-1
  - Operating system: `Linux mrc-tux 5.3.0-26-generic #28-Ubuntu SMP Wed Dec 18 05:37:46 UTC 2019 x86_64 GNU/Linux`

  - Compiler: `c++ (Debian 9.2.1-22) 9.2.1 20200104`

### How to repeat the problem

I can provide a Dockerfile, if needed

### Expected behaviour

Successful build followed by tests passing

### Actual behaviour

Build fails:

`/build/seqan3-3.0.0.0.git.20200110.b42671eac39/include/seqan3/range/decorator/gap_decorator.hpp:297:13: error: no match for ‘operator=’ (operand types are ‘seqan3::gap_decorator<std::span<seqan3::dna4, -1> >’ and ‘seqan3::detail::view_take<std::span<seqan3::dna4, -1>, false, false>’)`

https://gist.github.com/mr-c/12d1031685d347b88cc2bae4140d1dfd

marehr:
Fixes #1479

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,8 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 
 ## API changes
 
+* **The required version of the ranges-v3 library has increased:** We now support the versions >= 0.10.0 and < 0.11.0,
+  increasing the previous requirement of >= 0.5.0 and < 0.6.0.
 * **Customising for third party types has changes slightly:**
   You are only affected if you added types to `seqan3::custom::`.
   Please see [About Customisation](http://docs.seqan.de/seqan/3-master-user/about_customisation.html).
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Please see the [online documentation](https://docs.seqan.de/seqan/3-master-user/
 |**compiler**       | [GCC](https://gcc.gnu.org)                           | ≥ 7      | no other compiler is currently supported!   |
 |**build system**   | [CMake](https://cmake.org)                           | ≥ 3.4    | optional, but recommended                   |
 |**required libs**  | [SDSL](https://github.com/xxsds/sdsl-lite)           | ≥ 3      |                                             |
-|                   | [Range-V3](https://github.com/ericniebler/range-v3)  | ≥ 1.0    |                                             |
+|                   | [Range-V3](https://github.com/ericniebler/range-v3)  | ≥ 0.10.0 |                                             |
 |**optional libs**  | [cereal](https://github.com/USCiLab/cereal)          | ≥ 1.2.3  | required for serialisation and CTD support  |
 |                   | [zlib](https://github.com/madler/zlib)               | ≥ 1.2    | required for `*.gz` and `.bam` file support |
 |                   | [bzip2](http://www.bzip.org)                         | ≥ 1.0    | required for `*.bz2` file support           |
--- a/include/seqan3/std/span
+++ b/include/seqan3/std/span
@@ -54,6 +54,25 @@ struct is_std_array_impl<array<span_tp, span_sz>> : public true_type {};
 template <class span_tp>
 struct is_std_array : public is_std_array_impl<remove_cv_t<span_tp>> {};
 
+} // namespace std
+
+#if RANGE_V3_VERSION >= 901
+namespace ranges
+{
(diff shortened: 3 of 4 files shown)


mr-c:
This is appreciated, as Debian has already upgraded to Range 0.10.0

include/seqan3/std/span
```diff
@@ -54,6 +54,25 @@ struct is_std_array_impl<array<span_tp, span_sz>> : public true_type {};
 template <class span_tp>
 struct is_std_array : public is_std_array_impl<remove_cv_t<span_tp>> {};
 
+} // namespace std
+
+#if RANGE_V3_VERSION >= 901
```
eseiler: We require `>= 9.0.0`.  I guess you mean `>= 900`?
But then we wouldn't need this `#if` since `platform.hpp` already checks it

marehr: no this was introduced between 9.0.1 and 10.0.0

CHANGELOG.md
```diff
@@ -45,6 +45,8 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 
 ## API changes
 
+* The required version of the ranges-v3 library has increased. We now support the versions >= 0.9.0 and < 0.11.0,
```
rrahn:
```suggestion
* **The required version of the ranges-v3 library has increased:** We now support the versions >= 0.9.0 and < 0.11.0,
```

rrahn:
@marehr you can squash the last commit and we can merge.

h-2:
Just as a note: If you advertise that you support both 0.9.x and 0.10.x you need to test both branches individually, too. (there is no stable API right now)

I would suggest just moving to the new version and as soon as a stable version is released, use that.

rrahn:
So you suggest dropping support for 0.9. altogether?

h-2:
Yep! The interfaces between range-v3 versions are unstable anyway (until 1.0). As long as SeqAn3 requires range-v3 (no std::ranges yet), I don't see why it needs to provide backwards compatibility. And like I said, supporting multiple versions means having to test basically all tests with both.

marehr:
@rrahn done

smehringer:
I would also only advertise the version we are testing.

marehr:
@smehringer @h-2 @rrahn I upped the minimum requirement

CHANGELOG.md
```diff
@@ -45,6 +45,8 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 
 ## API changes
 
+* **The required version of the ranges-v3 library has increased:** We now support the versions >= 0.9.0 and < 0.11.0,
```
smehringer: The documentation needs also an update :)

README.md
```diff
@@ -35,7 +35,7 @@ Please see the [online documentation](https://docs.seqan.de/seqan/3-master-user/
 |**compiler**       | [GCC](https://gcc.gnu.org)                           | ≥ 7      | no other compiler is currently supported!   |
 |**build system**   | [CMake](https://cmake.org)                           | ≥ 3.4    | optional, but recommended                   |
 |**required libs**  | [SDSL](https://github.com/xxsds/sdsl-lite)           | ≥ 3      |                                             |
-|                   | [Range-V3](https://github.com/ericniebler/range-v3)  | ≥ 1.0    |                                             |
+|                   | [Range-V3](https://github.com/ericniebler/range-v3)  | ≥ 0.9.0  |                                             |
```
smehringer: here too

smehringer:
> @smehringer @h-2 @rrahn I upped the minimum requirement

Thanks, although I don't know what @rrahn votes for.  But if we want we can still add nightlies for the other version and change the requirements again I think.

rrahn:
> Thanks, although I don't know what @rrahn votes for. But if we want we can still add nightlies for the other version and change the requirements again I think.

No, please not. I don't know what you think,  but I am in total favor of having as least dependencies as possible.

PR #1472, opened 2020-01-09
[FEATURE] fast ostream buf iterator

### LINKED ISSUE #1570: Custom ostreambuf iterator
Implement a custom ostreambuf iterator that uses less virtual function calls then the one in the STL to improve runtime when writing to the underlying stream buffer.

smehringer:
Analoguesly to the fast_istreambuf_iterator, this ostreambuf iterator does less virtual function calls.

Still WIP!

Resolves #1570

### DIFF
--- a/include/seqan3/io/sequence_file/format_embl.hpp
+++ b/include/seqan3/io/sequence_file/format_embl.hpp
@@ -17,8 +17,6 @@
 #include <string_view>
 #include <vector>
 
-#include <range/v3/view/chunk.hpp>
-
 #include <seqan3/alphabet/nucleotide/dna5.hpp>
 #include <seqan3/core/char_operations/predicate.hpp>
 #include <seqan3/core/type_traits/range.hpp>
@@ -31,15 +29,11 @@
 #include <seqan3/range/detail/misc.hpp>
 #include <seqan3/range/views/char_to.hpp>
 #include <seqan3/range/views/istreambuf.hpp>
-#include <seqan3/range/views/join.hpp>
 #include <seqan3/range/views/repeat_n.hpp>
 #include <seqan3/range/views/to_char.hpp>
-#include <seqan3/range/views/take_exactly.hpp>
 #include <seqan3/range/views/take_line.hpp>
 #include <seqan3/range/views/take_until.hpp>
-#include <seqan3/range/views/to.hpp>
 #include <seqan3/std/algorithm>
-#include <seqan3/std/charconv>
 #include <seqan3/std/ranges>
 
 namespace seqan3
@@ -206,9 +200,10 @@ protected:
                                id_type                              && id,
                                qual_type                            && SEQAN3_DOXYGEN_ONLY(qualities))
     {
-        seqan3::ostreambuf_iterator stream_it{stream};
+        seqan3::detail::fast_ostreambuf_iterator stream_it{*stream.rdbuf()};
+
         [[maybe_unused]] size_t sequence_size = 0;
-        [[maybe_unused]] char buffer[50];
+
         if constexpr (!detail::decays_to_ignore_v<seq_type>)
             sequence_size = std::ranges::distance(sequence);
 
@@ -224,18 +219,16 @@ protected:
 
             if (options.embl_genbank_complete_header)
             {
-                std::ranges::copy(id, stream_it);
+                stream_it.write_range(id);
             }
             else
             {
-                std::ranges::copy(std::string_view{"ID "}, stream_it);
-                std::ranges::copy(id, stream_it);
-                std::ranges::copy(std::string_view{"; "}, stream_it);
-                auto res = std::to_chars(&buffer[0], &buffer[0] + sizeof(buffer), sequence_size);
-                std::copy(&buffer[0], res.ptr, stream_it);
-                std::ranges::copy(std::string_view{" BP.\n"}, stream_it);
+                stream_it.write_range(std::string_view{"ID "});
+                stream_it.write_range(id);
+                stream_it.write_range(std::string_view{"; "});
+                stream_it.write_number(sequence_size);
+                stream_it.write_range(std::string_view{" BP.\n"});
             }
-
         }
 
         // Sequence
@@ -245,32 +238,47 @@ protected:
         }
         else
         {
-            if (ranges::empty(sequence)) //[[unlikely]]
+            if (std::ranges::empty(sequence)) //[[unlikely]]
                 throw std::runtime_error{"The SEQ field may not be empty when writing embl files."};
 
-            std::ranges::copy(std::string_view{"SQ Sequence "}, stream_it);
-            auto res = std::to_chars(&buffer[0], &buffer[0] + sizeof(buffer), sequence_size);
-            std::copy(&buffer[0], res.ptr, stream_it);
-            std::ranges::copy(std::string_view{" BP;\n"}, stream_it);
-            auto seqChunk = sequence | ranges::view::chunk(60);
-            unsigned int i = 0;
-            size_t bp = 0;
-            for (auto chunk : seqChunk)
+            // write beginning of sequence record
+            stream_it.write_range(std::string_view{"SQ Sequence "});
+            stream_it.write_number(sequence_size);
+            stream_it.write_range(std::string_view{" BP;\n"});
+
+            // write sequence in chunks of 60 bp's with a space after 10 bp's
+            auto char_sequence = sequence | views::to_char;
+            auto it = std::ranges::begin(char_sequence);
+            size_t written_chars{0};
+            uint8_t chunk_size{10u};
+
+            while (it != std::ranges::end(char_sequence))
             {
-                std::ranges::copy(chunk | views::to_char
-                                        | ranges::view::chunk(10)
-                                        | views::join(' '), stream_it);
-                ++i;
+                auto current_end = it;
+                size_t steps = std::ranges::advance(current_end, chunk_size, std::ranges::end(char_sequence));
+
+                using subrange_t = std::ranges::subrange<decltype(it), decltype(it), std::ranges::subrange_kind::sized>;
+                it = stream_it.write_range(subrange_t{it, current_end, chunk_size - steps});
                 stream_it = ' ';
-                bp = std::min(sequence_size, bp + 60);
-                uint8_t num_blanks = 60 * i - bp;  // for sequence characters
-                num_blanks += num_blanks / 10;     // additional chunk separators
-                std::ranges::copy(views::repeat_n(' ', num_blanks), stream_it);
-                std::ranges::copy(std::to_string(bp), stream_it);
-                stream_it = '\n';
+                written_chars += chunk_size;
+
+                if (written_chars % 60 == 0)
+                {
+                    stream_it.write_number(written_chars);
+                    stream_it.write_end_of_line(options.add_carriage_return);
+                }
             }
-            std::ranges::copy(std::string_view{"//"}, stream_it);
-            stream_it = '\n';
+
+            // fill last line
+            auto characters_in_last_line = sequence_size % 60;
+            auto number_of_padding_needed = 65 - characters_in_last_line - characters_in_last_line / chunk_size;
+            stream_it.write_range(views::repeat_n(' ', number_of_padding_needed));
+            stream_it.write_number(sequence_size);
+            stream_it.write_end_of_line(options.add_carriage_return);
+
+            // write end-of-record-symbol
+            stream_it.write_range(std::string_view{"//"});
+            stream_it.write_end_of_line(options.add_carriage_return);
         }
     }
 };
--- a/include/seqan3/io/sequence_file/format_fasta.hpp
+++ b/include/seqan3/io/sequence_file/format_fasta.hpp
@@ -135,7 +135,7 @@ protected:
                                id_type && id,
                                qual_type && SEQAN3_DOXYGEN_ONLY(qualities))
     {
-        seqan3::ostreambuf_iterator stream_it{stream};
+        seqan3::detail::fast_ostreambuf_iterator stream_it{*stream.rdbuf()};
 
         // ID
         if constexpr (detail::decays_to_ignore_v<id_type>)
@@ -315,11 +315,8 @@ private:
     }
 
     //!\brief Implementation of writing the ID.
-    template <typename stream_it_t,
-              typename id_type>
-    void write_id(stream_it_t & stream_it,
-                   sequence_file_output_options const & options,
-                   id_type && id)
+    template <typename stream_it_t, typename id_type>
+    void write_id(stream_it_t & stream_it, sequence_file_output_options const & options, id_type && id)
     {
         if (options.fasta_legacy_id_marker)
             stream_it = ';';
@@ -329,48 +326,36 @@ private:
         if (options.fasta_blank_before_id)
             stream_it = ' ';
 
-        std::ranges::copy(id, stream_it);
-
-        detail::write_eol(stream_it, options.add_carriage_return);
+        stream_it.write_range(id);
+        stream_it.write_end_of_line(options.add_carriage_return);
     }
 
     //!\brief Implementation of writing the sequence.
-    template <typename stream_it_t,
-              typename seq_type>
-    void write_seq(stream_it_t & stream_it,
-                   sequence_file_output_options const & options,
-                   seq_type && seq)
+    template <typename stream_it_t, typename seq_type>
+    void write_seq(stream_it_t & stream_it, sequence_file_output_options const & options, seq_type && seq)
     {
+        auto char_sequence = seq | views::to_char;
+
         if (options.fasta_letters_per_line > 0)
         {
-        #if SEQAN3_WORKAROUND_VIEW_PERFORMANCE
-            size_t count = 0;
-            for (auto c : seq)
+            /* Using `views::interleave` is probably the way to go but that needs performance-tuning.*/
+            auto it = std::ranges::begin(char_sequence);
+            auto end = std::ranges::end(char_sequence);
+
+            while (it != end)
             {
-                stream_it = to_char(c);
-                if (++count % options.fasta_letters_per_line == 0)
-                    detail::write_eol(stream_it, options.add_carriage_return);
+                /* Note: This solution is slightly suboptimal for sized but non-random-access ranges.*/
+                auto current_end = it;
+                size_t steps = std::ranges::advance(current_end, options.fasta_letters_per_line, end);
+                using subrange_t = std::ranges::subrange<decltype(it), decltype(it), std::ranges::subrange_kind::sized>;
+                it = stream_it.write_range(subrange_t{it, current_end, (options.fasta_letters_per_line - steps)});
+                stream_it.write_end_of_line(options.add_carriage_return);
             }
-            if (count % options.fasta_letters_per_line != 0)
-                detail::write_eol(stream_it, options.add_carriage_return);
-
-        #else // ↑↑↑ WORKAROUND | ORIGINAL ↓↓↓
-
-            //TODO: combining chunk and join is substantially faster than views::interleave (2.5x), why?
-            std::ranges::copy(seq | views::to_char
-                                  | ranges::view::chunk(options.fasta_letters_per_line)
-                                  | views::join(options.add_carriage_return
-                                                    ? std::string_view{"\r\n"}
-                                                    : std::string_view{"\n"}),
-                              stream_it);
-            detail::write_eol(stream_it, options.add_carriage_return);
-        #endif // SEQAN3_WORKAROUND_VIEW_PERFORMANCE
         }
         else
         {
-            // No performance workaround here, because transform views alone are fast
-            std::ranges::copy(seq | views::to_char, stream_it);
-            detail::write_eol(stream_it, options.add_carriage_return);
+            stream_it.write_range(char_sequence);
+            stream_it.write_end_of_line(options.add_carriage_return);
         }
     }
 };
--- a/include/seqan3/io/stream/iterator.hpp
+++ b/include/seqan3/io/stream/iterator.hpp
@@ -6,8 +6,10 @@
 // -----------------------------------------------------------------------------------------------------
 
 /*!\file
- * \brief Provides seqan3::ostream and seqan3::ostreambuf iterator.
+ * \brief Provides seqan3::fast_istreambuf_iterator and seqan3::fast_ostreambuf_iterator, as well as, seqan3::ostream
+ *        and seqan3::ostreambuf iterator.
  * \author Rene Rahn <rene.rahn AT fu-berlin.de>
+ * \author Svenja Mehringer <svenja.mehringer AT fu-berlin.de>
  */
 
 #pragma once
@@ -19,6 +21,10 @@
 #endif // __cpp_lib_ranges
 
 #include <seqan3/core/platform.hpp>
+#include <seqan3/range/views/take.hpp>
+#include <seqan3/range/views/drop.hpp>
+#include <seqan3/std/algorithm>
+#include <seqan3/std/charconv>
 #include <seqan3/std/ranges>
 
 namespace seqan3::detail
@@ -166,6 +172,211 @@ public:
         return !(rhs == std::ranges::default_sentinel);
     }
     //!\}
+};
+
+/*!\brief Functionally the same as std::ostreambuf_iterator, but offers writing a range more efficiently.
+ * \ingroup stream
+ * \tparam char_t       The stream's character type.
+ * \tparam traits_t     The stream's traits type.
+ *
+ * \details
+ *
+ * The functions seqan3::fast_ostreambuf_iterator::write_range and seqan3::fast_ostreambuf_iterator::write_n allow
+ * more efficient writing of ranges by writing in chunks that avoiding overflow checks.
+ *
+ * \include test/snippet/io/detail/iterator_write_range.cpp
+ */
+template <typename char_t, typename traits_t = std::char_traits<char_t>>
+class fast_ostreambuf_iterator
+{
+private:
+    //!\brief Down-cast pointer to the stream-buffer.
+    stream_buffer_exposer<char_t, traits_t> * stream_buf = nullptr;
+
+public:
+    /*!\name Associated types
+     * \{
+     */
+    using difference_type   = ptrdiff_t;                //!< Defaults to ptrdiff_t.
+    using value_type        = char_t;                   //!< The char type of the stream.
+    using reference         = char_t;                   //!< The char type of the stream.
+    using pointer           = void;                     //!< Has no pointer type.
+    using iterator_category = std::output_iterator_tag; //!< Pure output iterator.
+    //!\}
+
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    fast_ostreambuf_iterator() noexcept = default; //!< Defaulted.
+    fast_ostreambuf_iterator(fast_ostreambuf_iterator const &) noexcept = default; //!< Defaulted.
+    fast_ostreambuf_iterator(fast_ostreambuf_iterator &&) noexcept = default; //!< Defaulted.
+    fast_ostreambuf_iterator & operator=(fast_ostreambuf_iterator const &) noexcept = default; //!< Defaulted.
+    fast_ostreambuf_iterator & operator=(fast_ostreambuf_iterator &&) noexcept = default; //!< Defaulted.
+    ~fast_ostreambuf_iterator() noexcept = default; //!< Defaulted.
+
+    //!\brief Construct from a stream buffer.
+    explicit fast_ostreambuf_iterator(std::basic_streambuf<char_t, traits_t> & ibuf) :
+        stream_buf{reinterpret_cast<stream_buffer_exposer<char_t, traits_t> *>(&ibuf)}
+    {
+        assert(stream_buf != nullptr);
+        if (stream_buf->pptr() == stream_buf->epptr())
+            stream_buf->overflow(); // ensures that put area has space available
+    }
+    //!\}
+
+    /*!\name Arithmetic operators
+     * \{
+     */
+    //!\brief no op.
+    fast_ostreambuf_iterator & operator++()
+    {
+        return *this;
+    }
+    //!\overload
+    fast_ostreambuf_iterator & operator++(int)
+    {
+        return *this;
+    }
+    //!\}
+
+    //!\brief no op.
+    fast_ostreambuf_iterator & operator*()
+    {
+        return *this;
+    }
+
+    //!\brief Writes a character to the associated output stream.
+    fast_ostreambuf_iterator & operator=(char_t const c)
+    {
+        assert(stream_buf != nullptr);
+        if (stream_buf->pptr() == stream_buf->epptr())
+        {
+            if (stream_buf->sputc(c) == traits_t::eof()) // overflow() [virtual], then write character
+            {
+                // LCOV_EXCL_START
+                throw std::ios_base::failure{"Cannot write to output stream (reached traits::eof() condition)."};
+                // LCOV_EXCL_STOP
+            }
+        }
+        else
+        {
+            *stream_buf->pptr() = c;
+            stream_buf->pbump(1); // advance pptr() in put area without any checks
+        }
+        return *this;
+    }
+
+    //!\brief Returns `true if this iterator has encountered the end-of-file condition on output, `false` otherwise.
+    bool failed() const noexcept
+    {
+        return stream_buf->overflow() == traits_t::eof();
+    }
+
+    /*!\brief Writes a range to the associated output.
+     * \tparam range_type The type of range to write; Must model std::ranges::forward_range.
+     * \param[in] rng The range to write.
+     * \returns If `range_type` models `std::ranges::forwarding_range` returns an iterator pointing to end of the range
+     *          (rng) else returns `void`.
+     *
+     * This function avoids the buffer-at-end check by writing the range in chunks, where a chunks has the size of
+     * the remaining space in the put area of the buffer.
+     * If the range type models `std::ranges::sized_range` the chunks are written using `std::ranges::copy_n`, which
+     * may use memcpy if applicable. Otherwise, a simple for loop iterates over the chunk.
+     *
+     * \attention You can only use the return value (end iterator) if your range type models
+     *            `std::ranges::forwarding_range`.
+     *
+     * Example:
+     *
+     * \include test/snippet/io/detail/iterator_write_range.cpp
+     */
+    template <std::ranges::forward_range range_type>
+    //!\cond
+        requires forwarding_range<range_type>
+    //!\endcond
+    auto write_range(range_type && rng)
+    {
+        using sen_t = std::ranges::sentinel_t<range_type>;
+        using it_t = std::ranges::iterator_t<range_type>;
+
+        it_t it = std::ranges::begin(rng);
+        sen_t end = std::ranges::end(rng);
+
+        while (it != end)
+        {
+            size_t const buffer_space = stream_buf->epptr() - stream_buf->pptr();
+
+            if constexpr (std::ranges::sized_range<range_type>)
+            {
+                size_t const characters_to_write = std::min<size_t>(std::ranges::distance(it, end), buffer_space);
+                auto copy_res = std::ranges::copy_n(it, characters_to_write, stream_buf->pptr());
+                it = copy_res.in;
+                stream_buf->pbump(characters_to_write);
+            }
+            else
+            {
+                size_t i = 0;
+                for (; it != end && i < buffer_space; ++it, ++i)
+                    *stream_buf->pptr() = *it;
+                stream_buf->pbump(i);
+            }
+
+            if (it == end) // no more characters to write
+                return it;
+
+            // Push one more character and flush
+            if (stream_buf->overflow(*it) == traits_t::eof())
+            {
+                // LCOV_EXCL_START
+                throw std::ios_base::failure{"Cannot write to output stream (reached traits::eof() condition)."};
+                // LCOV_EXCL_STOP
+            }
+
+            ++it; // drop 1 character that has been written in overflow()
+        }
+
+        return it;
+    }
+
+    //!\cond
+    // overload for non-forwarding_range types that return void
+    template <std::ranges::forward_range range_type>
+    void write_range(range_type && rng)
+    {
+        write_range(rng); // lvalue is always a safe range. return value is ignored because iterator would be dangling
+    }
+    //!\endcond
+
+    /*!\brief Writes a number to the underlying stream buffer using std::to_chars.
+     * \tparam number_type The type of number; must model seqan3::arithmetic.
+     * \param[in] num The number to write.
+     */
+    template <arithmetic number_type>
+    auto write_number(number_type num)
+    {
+        if (stream_buf->epptr() - stream_buf->pptr() > 300) // enough space for any number, should be likely
+        {
+            auto res = std::to_chars(stream_buf->pptr(), stream_buf->epptr(), num);
+            stream_buf->pbump(res.ptr - stream_buf->pptr()); // advance pptr
+        }
+        else
+        {
+            std::array<char, 300> arithmetic_buffer{};
+            auto res = std::to_chars(&arithmetic_buffer[0], &arithmetic_buffer[0] + sizeof(arithmetic_buffer), num);
+            write_range(std::ranges::subrange<char *, char *>(&arithmetic_buffer[0], res.ptr));
+        }
+    }
+
+    /*!\brief Write `"\n"` or `"\r\n"` to the stream buffer, depending on arguments.
+     * \param add_cr Whether to add carriage return, too.
+     * \ingroup io
+     */
+    void write_end_of_line(bool const add_cr)
+    {
+        if (add_cr)
+            *this = '\r';
+        *this = '\n';
+    }
 };
 
 } // namespace seqan3::detail
--- a/test/performance/io/CMakeLists.txt
+++ b/test/performance/io/CMakeLists.txt
@@ -1,5 +1,6 @@
 seqan3_benchmark(format_fasta_benchmark.cpp)
 seqan3_benchmark(format_vienna_benchmark.cpp)
 seqan3_benchmark(lowlevel_stream_input_benchmark.cpp)
+seqan3_benchmark(lowlevel_stream_output_benchmark.cpp)
 seqan3_benchmark(stream_input_benchmark.cpp)
 seqan3_benchmark(stream_output_benchmark.cpp)
--- a/test/unit/io/stream/fast_streambuf_iterator_test.cpp
+++ b/test/unit/io/stream/fast_streambuf_iterator_test.cpp
@@ -10,6 +10,10 @@
 #include <seqan3/io/stream/iterator.hpp>
 #include <seqan3/std/iterator>
 
+// -----------------------------------------------------------------------------
+// fast_istreambuf_iterator
+// -----------------------------------------------------------------------------
+
 TEST(fast_istreambuf_iterator, concept)
 {
     EXPECT_TRUE(std::input_iterator<seqan3::detail::fast_istreambuf_iterator<char>>);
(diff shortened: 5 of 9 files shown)


smehringer:
@h-2 
If you have time, do you want to take a look and maybe repeat your benchmark? 
This may improve performance. (If I did not screw up, ... I have no clue if I understood it correctly)

include/seqan3/io/sequence_file/format_fasta.hpp
```diff
@@ -135,7 +136,8 @@ class format_fasta
                                id_type && id,
                                qual_type && SEQAN3_DOXYGEN_ONLY(qualities))
     {
-        seqan3::ostreambuf_iterator stream_it{stream};
+        auto stream_view = seqan3::views::ostreambuf(stream);
+        auto stream_it = stream_view.begin();
```
h-2: I don't think we need `views::ostreambuf` at all since there are no other "pure output views" currently and we never compare with end.

Just use the iterator directly like before!

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +166,119 @@ class fast_istreambuf_iterator
...
+        }
+        return *this;
+    }
+
+    bool failed() const noexcept
+    {
+        return eof;
+    }
+
+    /*!\name Comparison operators
```
h-2: pure output iterators don't need these at all.

include/seqan3/io/sequence_file/format_fasta.hpp
```diff
@@ -135,7 +135,7 @@ class format_fasta
                                id_type && id,
```
h-2: Note that #1453 (waiting for review by you :wink) also touches these files!

include/seqan3/io/sequence_file/format_fasta.hpp
```diff
@@ -329,47 +326,39 @@ class format_fasta
...
+
+            for (distance_type c = 0; c < num_lines; ++c)
             {
-                stream_it = to_char(c);
-                if (++count % options.fasta_letters_per_line == 0)
-                    detail::write_eol(stream_it, options.add_carriage_return);
-            }
-            if (count % options.fasta_letters_per_line != 0)
+                stream_it.write_n(it, options.fasta_letters_per_line);
                 detail::write_eol(stream_it, options.add_carriage_return);
```
h-2: Are you not writing the first line of the sequence again and again here? Were do you increment `it`?
Is splitting the sequence and writing it line-wise better than chunk | join and writing as a whole? Could be either...

include/seqan3/io/stream/iterator.hpp
```diff
@@ -19,6 +21,7 @@
 #endif // __cpp_lib_ranges
 
 #include <seqan3/core/platform.hpp>
+#include <seqan3/range/views/drop.hpp>
```
h-2: Where is this used?

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +169,138 @@ class fast_istreambuf_iterator
...
+
+public:
+    /*!\name Associated types
+     * \{
+     */
+    using difference_type   = ptrdiff_t;               //!< Defaults to ptrdiff_t.
+    using value_type        = char_t;                  //!< The char type of the stream.
+    using reference         = char_t;                  //!< The char type of the stream.
+    using pointer           = void;                    //!< Has no pointer type.
+    using iterator_category = std::input_iterator_tag; //!< Pure input iterator.
```
h-2:
```suggestion
    using iterator_category = std::output_iterator_tag; //!< Pure output iterator.
```

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +169,138 @@ class fast_istreambuf_iterator
...
+        else
+        {
+            it = std::ranges::copy_n(it, buffer_space, stream_buf->pptr()).in;
+            stream_buf->pbump(buffer_space);
+
+            if (stream_buf->overflow(*it) == traits_t::eof()) // update put area by trying to write the next char
+                throw std::ios_base::failure{"Could not write to output stream."};
+
+            ++it;
+            ++buffer_space;
```
h-2: why not just call overflow?

h-2: or `*this = *it;` ?

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +169,138 @@ class fast_istreambuf_iterator
...
+     * \param[in] rng The range to write.
+     *
+     * \note This function calls std::ranges::distance(rng) and might be inefficient for ranges that aren't sized.
+     */
+    template <std::ranges::forward_range range_type>
+    void write_range(range_type && rng)
+    {
+        auto it = rng.begin();
+        write_n(it, std::ranges::distance(rng));
+    }
```
h-2: Why do we need two functions (write_n and write_range)?

include/seqan3/io/sequence_file/format_fasta.hpp
smehringer: My touch is way lighter :smirk_cat: Can you benchmark after my chagne again and see if we still need the extensive changes you introduce in #1453 ?

include/seqan3/io/sequence_file/format_fasta.hpp
smehringer: > Are you not writing the first line of the sequence again and again here? Were do you increment it?

`write_n` increments `it`.. I already thought that this might be a design flaw. I can also return the updated iterator instead?

> Is splitting the sequence and writing it line-wise better than chunk | join and writing as a whole? Could be either...

Well, your code explicitly stated that a for loop is faster than view chunk join and changing the ostreambuf iterator won't change that? So I would suggest to only keep the most efficient code even though the range version is more pretty.
I do not have time for extensive benchmarks on this :/

include/seqan3/io/stream/iterator.hpp
smehringer: It was used in your version of write_range. I forgot to remove it again.

include/seqan3/io/stream/iterator.hpp
smehringer: ops

include/seqan3/io/stream/iterator.hpp
smehringer: > why not just call overflow?

I do call overflow in the if clause above this? I just increment it here because overflow has written the character already (and I cannot call `overflow()` without a character because it won't actually overflow then). I could not increment buffer_space here but instead do `write_n(it, n - buffer_space + 1);` in the line below.

> or *this = *it; ?

Because this would do another if check within `operator=()` but I already know that I am at the end of my buffer :woman_shrugging:

include/seqan3/io/stream/iterator.hpp
smehringer: Convenience?
(`write_n` is more generic and good for chunked output like fasta-line-wrapping without using `view::take` all the time. But write_range is more convinient in general)

smehringer:
Lowlevel benchmarks

```
------------------------------------------------------------------------------
Benchmark                                    Time             CPU   Iterations
------------------------------------------------------------------------------
read_all<tag::std_stream_it>           7122309 ns      7121778 ns           97
read_all<tag::std_streambuf_it>        1522230 ns      1522175 ns          461
read_all<tag::seqan3_streambuf_it>      531766 ns       531724 ns         1291
read_all<tag::seqan2_stream_it>         378025 ns       378010 ns         1842
write_all<tag::std_stream_it>           191470 ns       180745 ns         3890
write_all<tag::std_streambuf_it>         58340 ns        45578 ns        15905
write_all<tag::seqan3_streambuf_it>      57148 ns        46310 ns        16013
write_all<tag::seqan2_stream_it>         74148 ns        45971 ns        15493

```

smehringer:
@h-2 Why is my iterator not an `std::ostream_iterator` :cry:

include/seqan3/io/sequence_file/format_fasta.hpp
```diff
@@ -329,47 +326,39 @@ class format_fasta
...
+
+        auto v = seq | views::to_char;
+
         if (options.fasta_letters_per_line > 0)
         {
-        #if SEQAN3_WORKAROUND_VIEW_PERFORMANCE
-            size_t count = 0;
-            for (auto c : seq)
+            auto it = v.begin();
+            distance_type num_lines = std::ranges::distance(seq) / options.fasta_letters_per_line;
```
h-2: Calling distance might be really expensive if the range is not sized! Fasta files can hold genomes...

include/seqan3/io/sequence_file/format_fasta.hpp
h-2: > write_n increments it.. I already thought that this might be a design flaw. I can also return the updated iterator instead?

Yes, that would be better! Iterators are usually passed by copy or move, not by reference.

> Well, your code explicitly stated that a for loop is faster than view chunk join and changing the ostreambuf iterator won't change that? So I would suggest to only keep the most efficient code even though the range version is more pretty. I do not have time for extensive benchmarks on this :/

Well, the point about my code was that you *can* easily benchmark it by turning a macro on/off :nerd_face: 

If I am to benchmark this again, it would really be a lot easier, if you only replace the lines that have `copy()` with `write_range()`...

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +169,138 @@ class fast_istreambuf_iterator
...
+    void write_n(iterator_type & it, size_t n)
+    {
+        if (size_t buffer_space = stream_buf->epptr() - stream_buf->pptr(); n <= buffer_space)
+        {
+            it = std::ranges::copy_n(it, n, stream_buf->pptr()).in; // copy_n returns incremented iterators {in, out}
+            stream_buf->pbump(n);
+        }
+        else
+        {
+            it = std::ranges::copy_n(it, buffer_space, stream_buf->pptr()).in;
```
h-2:
```suggestion
            /* Fill up remaining buffer */
            it = std::ranges::copy_n(it, buffer_space, stream_buf->pptr()).in;
```

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +169,138 @@ class fast_istreambuf_iterator
...
+            stream_buf->pbump(buffer_space);
+
+            if (stream_buf->overflow(*it) == traits_t::eof()) // update put area by trying to write the next char
```
h-2:
```suggestion
            /* Push one more character and flush */
            if (stream_buf->overflow(*it) == traits_t::eof()) // update put area by trying to write the next char
```

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +169,138 @@ class fast_istreambuf_iterator
...
+                throw std::ios_base::failure{"Could not write to output stream."};
+
+            ++it;
+            ++buffer_space;
+
+            write_n(it, n - buffer_space);
```
h-2:
```suggestion
            /* Write remaining input data recursively */
            write_n(it, n - buffer_space);
```

include/seqan3/io/stream/iterator.hpp
h-2: > I cannot call overflow() without a character because it won't actually overflow then

Ah, ok :+1: 

I thought it was strange that the `== eof() and  throw`  had disappeared from the other code blocks but reappeared here now. But, up to you!

include/seqan3/io/stream/iterator.hpp
h-2: Some documentation would help understanding the semantics here!

include/seqan3/io/stream/iterator.hpp
h-2: I think, I would have designed it as a single function since view_take is free for random access ranges and even single iterators can be turned into such a range via `views::counted`, but that's up to you!

h-2:
> Lowlevel benchmarks

It's slower for single character writes than the std::ostreambuf_iterator in microbenchmarks? :thinking: 
Did you have a look at LLVM's oder GNU's ostreambuf implementation by chance?

> EDIT: with write range

That looks good!

include/seqan3/io/stream/iterator.hpp
smehringer: Well but as you said earlier (that distance() might be expensive for non sized views) we do not always have random access ranges, so I thought this is more efficient ?

smehringer:
> It's slower for single character writes than the std::ostreambuf_iterator in microbenchmarks? thinking
Did you have a look at LLVM's oder GNU's ostreambuf implementation by chance?

no it's not ? It's roughly equal

include/seqan3/io/sequence_file/format_fasta.hpp
smehringer: Well but without knowing the size, I cannot efficiently use the buffer.. I could differentiate between both cases..

h-2:
> no it's not ? It's roughly equal

It's even 5% faster, I don't know what confused me there :man_facepalming:

include/seqan3/io/stream/iterator.hpp
h-2: It's hard to say without benchmarking to be honest.

include/seqan3/io/sequence_file/format_fasta.hpp
smehringer: I removed the dependance on std::distance

include/seqan3/io/stream/iterator.hpp
smehringer: So I did a benchmark and the time was the same so I removed write_n now

include/seqan3/io/stream/iterator.hpp
smehringer: it is used now

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -245,34 +255,51 @@ class format_embl
...
+
+            // write sequence in chunks of 60 bp's with a space after 10 bp's
+
+            auto char_sequence = sequence | views::to_char;
+            size_t written_characters{0};
+
+            auto it = stream_it.write_range(char_sequence | views::take(10)); // write first 10 characters
+            stream_it = ' ';
+
+            while (it != char_sequence.end())
```
h-2: Do not rely on member functions when you are not sure about the type. We make no promises regarding what `views::to_char` returns so it's better to use `std::ranges::begin|end|...`.

h-2: This affects quite a few places!

include/seqan3/io/sequence_file/format_fasta.hpp
```diff
@@ -329,47 +326,35 @@ class format_fasta
...
     //!\brief Implementation of writing the sequence.
-    template <typename stream_it_t,
-              typename seq_type>
-    void write_seq(stream_it_t & stream_it,
-                   sequence_file_output_options const & options,
-                   seq_type && seq)
+    template <typename stream_it_t, typename seq_type>
+    void write_seq(stream_it_t & stream_it, sequence_file_output_options const & options, seq_type && seq)
     {
+        // using distance_type = decltype(std::ranges::distance(seq));
```
h-2:
```suggestion
```

include/seqan3/io/sequence_file/format_fasta.hpp
```diff
@@ -329,47 +326,35 @@ class format_fasta
...
-            std::ranges::copy(seq | views::to_char
-                                  | ranges::view::chunk(options.fasta_letters_per_line)
-                                  | views::join(options.add_carriage_return
-                                                    ? std::string_view{"\r\n"}
-                                                    : std::string_view{"\n"}),
-                              stream_it);
-            detail::write_eol(stream_it, options.add_carriage_return);
-        #endif // SEQAN3_WORKAROUND_VIEW_PERFORMANCE
+                written_characters += options.fasta_letters_per_line;
+            }
```
h-2: Ok, I know I said we don't care about non-sized or non-random-access ranges. BUT :rofl: 

views::drop is has linear construction on non-random-access ranges in regard to its parameter. And you are going over the whole sequence again every iteration, first the first 70 chars, next time 140 chars... Although you in fact have the iterator where the sequence starts every time! I propose the following loop which is almost identical:

```cpp
            while (it != v.end())
            {
                 auto e = it;
                 std::ranges::advance(e, options.fasta_letters_per_line, std::ranges::end(v));
                 auto r = std::ranges::subrange{it, e};
                 it = stream_it.write_range(r);
            }
```

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +171,124 @@ class fast_istreambuf_iterator
...
+     */
+    template <std::ranges::forward_range range_type>
+    auto write_range(range_type && rng)
+    {
+        auto buffer_space = stream_buf->epptr() - stream_buf->pptr();
+        auto copy_res = std::ranges::copy(rng | views::take(buffer_space), stream_buf->pptr());
+        auto written_chars = copy_res.out - stream_buf->pptr();
+        stream_buf->pbump(written_chars);
+
+        auto remaining_range = rng | views::drop(written_chars);
```
h-2:
```suggestion
```

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +171,124 @@ class fast_istreambuf_iterator
...
+
+        if (remaining_range.begin() == remaining_range.end()) // no more characters to write
```
h-2:
```suggestion
        if (copy_res.in == std::ranges::end(rng)) // no more characters to write
```

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +171,124 @@ class fast_istreambuf_iterator
...
+            return copy_res.in;
+
+        // Push one more character and flush
+        if (stream_buf->overflow(*copy_res.in) == traits_t::eof())
+            throw std::ios_base::failure{"Cannot write to output stream (reached traits::eof() condition)."};
+
+        // drop 1 character that has been written in overflow() and write remaining range recursively
+        return write_range(remaining_range | views::drop(1));
```
h-2:
```suggestion
        return write_range(remaining_range | views::drop(written_chars + 1));
```

The previous small changes prevent creating two drop-views in case no flattening happens.

include/seqan3/io/stream/iterator.hpp
h-2: :+1:

test/performance/io/lowlevel_stream_benchmark.cpp
```diff
@@ -98,11 +99,97 @@ void read_all(benchmark::State & state)
...
+    std::vector<char> cont_rando = seqan3::test::generate_sequence<char>(10'000, 0, 0);
+
+    for (auto _ : state)
+    {
+        std::ofstream os{filename.get_path(), std::ios::binary};
+        seqan3::detail::fast_ostreambuf_iterator<char> it{*os.rdbuf()};
+
+        it.write_range(cont_rando);
+    }
+}
```
h-2: why not integrate this with the rest? Just add an extra tag `tag::seqan3_streambuf_range`!

include/seqan3/io/stream/iterator.hpp
h-2: You removed the documentation again in the latest version :roll_eyes:

include/seqan3/io/stream/iterator.hpp
smehringer: This might not work because the input to `std::ranges::copy` is `rng | take(..)` so the iretors are not comparable :(

h-2: Ah, correct! But You can do :
```cpp
auto tview = rng | views::take(buffer_space);
auto copy_res = std::ranges::copy(tview, stream_buf->pptr());
// ...
if (copy_res.in == std::ranges::end(tview)) // no more characters to write
```

Or, not?

h-2: Oh, you are right of course! However, this means that you are also not returning an iterator of the original range when calling `return copy_res.in;` and outside this function this will be dangling (it refers to the take_view) and you are assigning it back to the original iterator :boom: (This all just works because our iterators inherit from each other and casting actually does the correct thing by chance)

I would recommend just doing what I proposed on the outside, as well: 

```cpp
                 auto it = std::ranges::begin(rng);
                 auto e = it;
                 std::ranges::advance(e, buffer_space, std::ranges::end(v));
                 auto copy_res = std::ranges::copy(it, e, stream_buf->pptr());
              
                 // ...

                 if (copy_res.in == std::ranges::end(rng)) // no more characters to write
```

Now `copy_res.in` is guaranteed to be of the original iterator type. These changes should have zero overhead for sized random access ranges (advance is in O(1) for them).

h-2: However, this functions should also be documented that if an rvalue of a range is passed to it and that range is not a safe_range than any returned iterator will still be dangling.

include/seqan3/io/stream/iterator.hpp
smehringer: well because I removed write_n?

test/performance/io/lowlevel_stream_benchmark.cpp
smehringer: Because it is not a direct comparison to std and seqan2 and also not available for the input streambuf iterator?

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -245,34 +255,51 @@ class format_embl
...
-                ++i;
+                written_characters += 10;
+
+                if (written_characters % 60 == 0)
+                {
+                    write_number(stream_it, written_characters);
+                    detail::write_eol(stream_it, options.add_carriage_return);
+                }
+
+                it = stream_it.write_range(char_sequence | views::drop(written_characters) | views::take(10));
```
h-2: Is this the same problem as below? Non-random-access ranges are parsed by increasing amount of written_characters ever iteration?

include/seqan3/io/sequence_file/format_fasta.hpp
```diff
@@ -329,47 +326,31 @@ class format_fasta
...
+            auto it = std::ranges::begin(char_sequence);
+
+            while (it != std::ranges::end(char_sequence))
             {
-                stream_it = to_char(c);
-                if (++count % options.fasta_letters_per_line == 0)
-                    detail::write_eol(stream_it, options.add_carriage_return);
-            }
-            if (count % options.fasta_letters_per_line != 0)
+                auto end_it = std::ranges::next(it, options.fasta_letters_per_line, std::ranges::end(char_sequence));
```
h-2: Here at least, you only need to move by `options.fasta_letters_per_line`, above the amount increases on every iteration.

include/seqan3/io/sequence_file/format_fasta.hpp
```diff
@@ -329,47 +326,31 @@ class format_fasta
...
+                it = stream_it.write_range(std::ranges::subrange<decltype(it), decltype(end_it)>{it, end_it});
                 detail::write_eol(stream_it, options.add_carriage_return);
```
h-2: Can you add a note inside the code or as part of the functions documentation that 
a) The performance might be suboptimal for non-random-access-ranges 
b) `views::interleave` is probably the way to go but that needs performance-tuning itself

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +171,145 @@ class fast_istreambuf_iterator
...
+    //!\brief Returns `true if this iterator has encountered the end-of-file condition on output, `false` otherwise.
+    bool failed() const noexcept
+    {
+        return stream_buf->overflow() == traits_t::eof();
+    }
+
+    /*!\brief Writes a range to the associated output.
+     * \tparam range_type The type of range to write; Must model std::ranges::forward_range.
+     * \param[in] rng The range to write.
+     */
```
h-2: Can you document the design decisions and the general mechanics of this function "for future generations"? :wink:

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +171,145 @@ class fast_istreambuf_iterator
...
+            for (size_t i = 0; rng_it != rng_end && i < buffer_space; ++rng_it, ++i)
+            {
+                *stream_buf->pptr() = *rng_it;
+                stream_buf->pbump(1);
+            }
+        }
+
+        if (rng_it == rng_end) // no more characters to write
+        {
+            if constexpr (std::ranges::viewable_range<range_type>)
```
h-2:
```suggestion
            if constexpr (std::ranges::safe_range<range_type>)
```

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +171,145 @@ class fast_istreambuf_iterator
...
+     * \param[in] rng The range to write.
+     */
+    template <std::ranges::forward_range range_type>
+    auto write_range(range_type && rng)
+    {
+        size_t buffer_space = stream_buf->epptr() - stream_buf->pptr();
+
+        auto rng_it = std::ranges::begin(rng);
+        auto rng_end = std::ranges::end(rng);
+
```
h-2:
```suggestion
size_t size_left = 0;

```

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +171,145 @@ class fast_istreambuf_iterator
...
+        if constexpr (std::ranges::sized_range<range_type>)
+        {
+            auto characters_to_write = std::min<size_t>(std::ranges::size(rng), buffer_space);
+            auto copy_res = std::ranges::copy_n(rng_it, characters_to_write, stream_buf->pptr());
+            rng_it = copy_res.in;
+            stream_buf->pbump(characters_to_write);
```
h-2:
```suggestion
            stream_buf->pbump(characters_to_write);
            size_left -= characters_to_write;
```

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +171,145 @@ class fast_istreambuf_iterator
...
+        }
+
+        // Push one more character and flush
+        if (stream_buf->overflow(*rng_it) == traits_t::eof())
+            throw std::ios_base::failure{"Cannot write to output stream (reached traits::eof() condition)."};
+
+        ++rng_it; // drop 1 character that has been written in overflow()
+
+        // write remaining range recursively
+        return write_range(std::ranges::subrange<decltype(rng_it), decltype(rng_end)>{rng_it, rng_end});
```
h-2:
```suggestion
        if constexpr (std::ranges::sized_range<range_type> &&
                      !std::sized_sentinel_for<std::ranges::sentinel_t<range_type>, std::ranges::iterator_t<range_type>>)
            return write_range(std::ranges::subrange<decltype(rng_it), decltype(rng_end), size_t>{rng_it, rng_end, size_left - 1});
        else
            return write_range(std::ranges::subrange<decltype(rng_it), decltype(rng_end)>{rng_it, rng_end});
```

include/seqan3/io/sequence_file/format_embl.hpp
smehringer: I wanted to make a card for embl that the format should be improved for non-random access ranges.. because I thought fasta is the more critical one. I can delete the whole embl transformation altogether if this takes too much time.

include/seqan3/io/sequence_file/format_fasta.hpp
smehringer: Well yes, there is no way I can circumcent this, except if I write char by char or if I introduce a `write_n(it, n, sentinal)` function to the iterator.. I thought you decided against both?

include/seqan3/io/stream/iterator.hpp
smehringer: `error: ‘safe_range’ is not a member of ‘std::ranges’` ...

include/seqan3/io/stream/iterator.hpp
smehringer: How can a range be sized but the sentinal type not model `sized_sentinal_for` for the iterator?

smehringer: Oh for view take.. :sleepy:

smehringer:
I've got the following problem with the sized subrange issue that might be a bug


The following snippet does not compile :( 

```cpp
#include <seqan3/range/views/repeat.hpp>
#include <seqan3/std/ranges>

int main()
{
    auto v = seqan3::views::repeat('A');
    auto t = v | seqan3::views::take(10);
    auto sr = std::ranges::subrange<decltype(t.begin()), decltype(t.end())>{t.begin(), t.end(), 10};
    seqan3::debug_stream << t;
}

```

The first ctor is rejected because it and sen do not model `sized_sentinel_for` (which is expected)
But the next ctor complains that the required `store_size_` evaluated to false but `store_size_` only checks `K == subrange_kind::sized && !(bool)sized_sentinel_for<S, I>`....
:woman_shrugging: 

```
/srv/public/smehringer/seqan3/submodules/range-v3/include/range/v3/view/subrange.hpp:268:19: note: candidate: ‘constexpr ranges::subrange<I, S, K>::subrange(I2&&, S, ranges::subrange<I, S, K>::size_type) requires  convertible_to_not_slicing_<I2, I> and sized_sentinel_for<S, I> [with I2 = seqan3::detail::view_take<seqan3::detail::repeat_view<char>, false, false>::iterator_type<seqan3::detail::repeat_view<char> >; I = seqan3::detail::view_take<seqan3::detail::repeat_view<char>, false, false>::iterator_type<seqan3::detail::repeat_view<char> >; S = ranges::default_sentinel_t; ranges::subrange_kind K = ranges::subrange_kind::unsized; ranges::subrange<I, S, K>::size_type = long unsigned int]’
  268 |         constexpr CPP_ctor(subrange)(I2 && i, S s, size_type n)( //
      |                   ^~~~~~~~
/srv/public/smehringer/seqan3/submodules/range-v3/include/range/v3/view/subrange.hpp:268:19: note:   constraints not satisfied
In file included from /srv/public/smehringer/seqan3/include/seqan3/std/iterator:25,
                 from /srv/public/smehringer/seqan3/include/seqan3/core/type_traits/iterator.hpp:20,
                 from /srv/public/smehringer/seqan3/include/seqan3/core/type_traits/range.hpp:20,
                 from /srv/public/smehringer/seqan3/include/seqan3/core/detail/debug_stream_range.hpp:18,
                 from /srv/public/smehringer/seqan3/include/seqan3/core/debug_stream.hpp:17,
                 from /srv/public/smehringer/seqan3/test/snippet/range/views/repeat.cpp:1:
/srv/public/smehringer/seqan3/submodules/range-v3/include/range/v3/iterator/concepts.hpp:237:22: note: within ‘template<class S, class I> concept const bool ranges::sized_sentinel_for<S, I> [with S = ranges::default_sentinel_t; I = seqan3::detail::view_take<seqan3::detail::repeat_view<char>, false, false>::iterator_type<seqan3::detail::repeat_view<char> >]’
  237 |     CPP_concept_bool sized_sentinel_for =
      |                      ^~~~~~~~~~~~~~~~~~
/srv/public/smehringer/seqan3/submodules/range-v3/include/range/v3/iterator/concepts.hpp:237:22: note:     with ‘const ranges::default_sentinel_t& s’
/srv/public/smehringer/seqan3/submodules/range-v3/include/range/v3/iterator/concepts.hpp:237:22: note:     with ‘const seqan3::detail::view_take<seqan3::detail::repeat_view<char>, false, false>::iterator_type<seqan3::detail::repeat_view<char> >& i’
/srv/public/smehringer/seqan3/submodules/range-v3/include/range/v3/iterator/concepts.hpp:237:22: note: the required expression ‘((((s - i), (i - s)), requires_<same_as<ranges::iter_difference_t<decltype (i)>, decltype ((s - i))> >), requires_<same_as<ranges::iter_difference_t<decltype (i)>, decltype ((i - s))> >)’ would be ill-formed
In file included from /srv/public/smehringer/seqan3/include/seqan3/std/concepts:25,
                 from /srv/public/smehringer/seqan3/include/seqan3/alphabet/exception.hpp:19,
                 from /srv/public/smehringer/seqan3/include/seqan3/alphabet/concept.hpp:15,
                 from /srv/public/smehringer/seqan3/include/seqan3/core/detail/debug_stream_alphabet.hpp:15,
                 from /srv/public/smehringer/seqan3/include/seqan3/core/debug_stream.hpp:15,
                 from /srv/public/smehringer/seqan3/test/snippet/range/views/repeat.cpp:1:
/srv/public/smehringer/seqan3/submodules/range-v3/include/range/v3/view/subrange.hpp:255:19: note: candidate: ‘constexpr ranges::subrange<I, S, K>::subrange(I2&&, S, ranges::subrange<I, S, K>::size_type) requires  convertible_to_not_slicing_<I2, I> and (store_size_<K, S, I>)() [with I2 = seqan3::detail::view_take<seqan3::detail::repeat_view<char>, false, false>::iterator_type<seqan3::detail::repeat_view<char> >; I = seqan3::detail::view_take<seqan3::detail::repeat_view<char>, false, false>::iterator_type<seqan3::detail::repeat_view<char> >; S = ranges::default_sentinel_t; ranges::subrange_kind K = ranges::subrange_kind::unsized; ranges::subrange<I, S, K>::size_type = long unsigned int]’
  255 |         constexpr CPP_ctor(subrange)(I2 && i, S s, size_type n)( //
      |                   ^~~~~~~~
/srv/public/smehringer/seqan3/submodules/range-v3/include/range/v3/view/subrange.hpp:255:19: note:   constraints not satisfied
/srv/public/smehringer/seqan3/submodules/range-v3/include/range/v3/view/subrange.hpp:255:19: note: ‘(store_size_<K, S, I>)()’ evaluated to false
```

include/seqan3/io/sequence_file/format_embl.hpp
h-2: > I wanted to make a card for embl that the format should be improved for non-random access ranges.. 

That's ok, too of course!

include/seqan3/io/sequence_file/format_fasta.hpp
h-2: This was just explanation for the previous comment, this if fine.

> I introduce a write_n(it, n, sentinal) function to the iterator.. I thought you decided against both?

As I said before, the `write_n` function won't help us. It makes this potentially faster but the case where the size is not known potentially slower (because of requiring `distance`).

h-2: [except that this should be replaces with the `couning_next` that I proposed.

include/seqan3/io/stream/iterator.hpp
h-2: [especially return value...]

include/seqan3/io/stream/iterator.hpp
h-2: Hm, it's in range-v3-master but not yet 0.10.0. Then using `seqan3::forwarding_range` for now.

h-2:
> But the next ctor complains that the required store_size_ evaluated to false but store_size_ only checks K == subrange_kind::sized && !(bool)sized_sentinel_for<S, I>....

Well, that's a very clear indication that you should set the third template parameter to `subrange_kind::sized`, or not? :laughing: 

I thought the third parameter was the size_type, but this mechanic is better. I have also updated the suggested code snippet above.

smehringer:
> Well, that's a very clear indication that you should set the third template parameter to subrange_kind::sized, or not? :laughing:

Of course I tried that too :roll_eyes: same result

h-2:
The problem is fixed by ericniebler/range-v3#1415

Just use `ranges::subrange` for now.

include/seqan3/io/sequence_file/format_fasta.hpp
```diff
@@ -329,47 +326,33 @@ class format_fasta
...
+            auto end = std::ranges::end(char_sequence);
+            while (it != end)
             {
-                stream_it = to_char(c);
-                if (++count % options.fasta_letters_per_line == 0)
-                    detail::write_eol(stream_it, options.add_carriage_return);
-            }
-            if (count % options.fasta_letters_per_line != 0)
+                using subrange_t = ranges::subrange<decltype(it), decltype(it), ranges::subrange_kind::sized>;
+                /* Note: Calling detail::counting_next is suboptimal for non-random-access ranges.*/
```
h-2: I think the current design is only suboptimal for sized non-random-access ranges. For all non-sized non-random-access ranges parsing outside eliminates the parsing inside (because the subrange being passed in is always sized now). So the solution is really good for 98% of the cases, I think.

include/seqan3/io/sequence_file/format_fasta.hpp
h-2: I still think adding a note somewhere about views::interleave wouldn't be too bad. Since that was created exactly for this purpose and actually preserves size and random-access...

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +171,182 @@ class fast_istreambuf_iterator
...
+    bool failed() const noexcept
+    {
+        return stream_buf->overflow() == traits_t::eof();
+    }
+
+    /*!\brief Writes a range to the associated output.
+     * \tparam range_type The type of range to write; Must model std::ranges::forward_range.
+     * \param[in] rng The range to write.
+     * \returns If `range_type` models `std::ranges::forwarding_range` returns an iterator pointing to end of the range
+     *          (rng) else returns void.
```
h-2:
```suggestion
     *          (rng) else returns `void`.
```

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +171,182 @@ class fast_istreambuf_iterator
...
+            return write_range(std::ranges::subrange<it_t, sen_t>{rng_it, rng_end});
+        }
+    }
+
+    //!\cond
+    // overload for non-forwarding_range types that return void
+    template <std::ranges::forward_range range_type>
+    void write_range(range_type && rng)
+    {
+        write_range(rng);
```
h-2:
```suggestion
        write_range(rng); // return value is ignored because iterator would be dangling
```

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +171,182 @@ class fast_istreambuf_iterator
...
+    //!\endcond
+    auto write_range(range_type && rng)
+    {
+        using sen_t = std::ranges::sentinel_t<range_type>;
+        using it_t = std::ranges::iterator_t<range_type>;
+
+        it_t rng_it = std::ranges::begin(rng);
+        sen_t rng_end = std::ranges::end(rng);
+
+        size_t buffer_space = stream_buf->epptr() - stream_buf->pptr();
```
h-2:
```suggestion
        size_t const buffer_space = stream_buf->epptr() - stream_buf->pptr();
```

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +171,182 @@ class fast_istreambuf_iterator
...
+
+        if constexpr (std::ranges::sized_range<range_type>)
+        {
+            auto characters_to_write = std::min<size_t>(std::ranges::size(rng), buffer_space);
```
h-2:
```suggestion
            size_t const characters_to_write = std::min<size_t>(std::ranges::size(rng), buffer_space);
```

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +171,182 @@ class fast_istreambuf_iterator
...
+        // Push one more character and flush
+        if (stream_buf->overflow(*rng_it) == traits_t::eof())
+            throw std::ios_base::failure{"Cannot write to output stream (reached traits::eof() condition)."};
+
+        ++rng_it; // drop 1 character that has been written in overflow()
+
+        // Write remaining range recursively. For sized ranges, ensure that the subrange is also sized.
+        if constexpr (std::ranges::sized_range<range_type> && !std::ranges::sized_sentinel_for<sen_t, it_t>)
+        {
+            using subrange_t = ranges::subrange<it_t, sen_t, ranges::subrange_kind::sized>;
```
h-2:
```suggestion
            // TODO: change to std::ranges::subrange... when updating to range-v3 > 0.10.0
            using subrange_t = ranges::subrange<it_t, sen_t, ranges::subrange_kind::sized>;
```

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +171,182 @@ class fast_istreambuf_iterator
...
+            static_assert(forwarding_range<range_type> == forwarding_range<subrange_t>, "not the same");
```
h-2: what's dis?

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +171,182 @@ class fast_istreambuf_iterator
...
+
+            size_t remaining_size = std::ranges::size(rng) - buffer_space - 1;
```
h-2:
```suggestion
            size_t const remaining_size = std::ranges::size(rng) - buffer_space - 1;
```

include/seqan3/range/detail/misc.hpp
```diff
@@ -40,4 +40,34 @@ template <std::ranges::forward_range rng_t>
...
+ * \ingroup range
+ * \tparam iterator_t Type of the iterator; must model std::ranges::forward_iterator.
+ * \tparam size_type Type of the size; must model arithmetic.
+ * \tparam sentinal_t Type of the range; must model std::ranges::sentinel_for<sentinal_t, iterator_t>.
+ * \param[in] it The iterator to increment.
+ * \param[in] s The number of steps to increment.
+ * \param[in] sen The sentinel used as a bound for incrementing `it`.
+ * \returns a std::pair over the incremented iterator and how often `it` was incremented.
+ */
+template <std::ranges::forward_iterator iterator_t, arithmetic size_type, typename sentinal_t>
```
h-2:
```suggestion
template <std::ranges::forward_iterator iterator_t, std::integral size_type, typename sentinal_t>
```

You cannot increment by `0.4` I think :wink:

include/seqan3/range/detail/misc.hpp
```diff
@@ -40,4 +40,34 @@ template <std::ranges::forward_range rng_t>
...
+//!\cond
+    requires std::ranges::sentinel_for<sentinal_t, iterator_t>
+//!\endcond
+std::pair<iterator_t, size_type> counting_next(iterator_t it, size_type s, sentinal_t sen)
```
h-2:
```suggestion
std::pair<iterator_t, size_type> counting_next(iterator_t it, size_type const s, sentinal_t sen)
```

:nail_care:

test/performance/io/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 seqan3_benchmark(format_fasta_benchmark.cpp)
 seqan3_benchmark(format_vienna_benchmark.cpp)
-seqan3_benchmark(lowlevel_stream_input_benchmark.cpp)
+seqan3_benchmark(lowlevel_stream_benchmark.cpp)
```
h-2: I would prefer having seperate tests for input and output, because shorter benchmarks means you get to see the desired results quicker!

Input and output are separate for a lot of other things, too.

include/seqan3/io/stream/iterator.hpp
h-2: Already fixed upstream!

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +171,182 @@ class fast_istreambuf_iterator
...
+            return rng_it;
+
+        // Push one more character and flush
+        if (stream_buf->overflow(*rng_it) == traits_t::eof())
+            throw std::ios_base::failure{"Cannot write to output stream (reached traits::eof() condition)."};
+
+        ++rng_it; // drop 1 character that has been written in overflow()
+
+        // Write remaining range recursively. For sized ranges, ensure that the subrange is also sized.
+        if constexpr (std::ranges::sized_range<range_type> && !std::ranges::sized_sentinel_for<sen_t, it_t>)
```
h-2:
```suggestion
        if constexpr (std::ranges::sized_range<range_type>)
```

I think this is enough actually, because the subrange ignores the third parameter if the first two can be substracted from each other.

include/seqan3/io/sequence_file/format_fasta.hpp
smehringer: I did add this comment :) but further up before the whole while loop

include/seqan3/io/stream/iterator.hpp
smehringer: ops :D left over debugging

test/performance/io/CMakeLists.txt
smehringer: Fine..

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -195,6 +193,19 @@ class format_embl
...
     }
 
+    /*!\brief Writes a number to the stream iterator using std::to_chars.
+     * \tparam stream_iterator_type The type of ostream iterator.
+     * \tparam number_type The type of number; must model seqan3::arithmetic.
+     * \param[in,out] stream_it The ostream iterator to write to.
+     * \param[in] num The number to write.
+     */
+    template <typename stream_iterator_type, arithmetic number_type>
+    auto write_number(stream_iterator_type & stream_it, number_type num)
```
marehr: missing test cases

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -195,6 +193,19 @@ class format_embl
...
+    {
+        auto res = std::to_chars(&arithmetic_buffer[0], &arithmetic_buffer[0] + sizeof(arithmetic_buffer), num);
+        stream_it.write_range(std::ranges::subrange<char *, char *>(&arithmetic_buffer[0], res.ptr));
+    }
```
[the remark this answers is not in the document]
h-2: For historic reasons! Prior to this PR, we did not have our own stream_it type. I agree that now that we have it, `write_number()` and `write_eol` could become members of it. It could happen as part of this PR, but it could also happen separately.

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +171,182 @@ class fast_istreambuf_iterator
...
+     * \attention You can only use the return value (end iterator) if your range type models
+     *            `std::ranges::forwarding_range`.
+     *
+     * Example:
+     *
+     * \include test/snippet/io/detail/iterator_write_range.cpp
+     */
+    template <std::ranges::forward_range range_type>
+    //!\cond
+        requires forwarding_range<range_type>
```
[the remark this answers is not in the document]
smehringer: Do you mean `safe_range`? Does not compile with gcc-7 :(

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +171,182 @@ class fast_istreambuf_iterator
...
+     * \returns If `range_type` models `std::ranges::forwarding_range` returns an iterator pointing to end of the range
+     *          (rng) else returns `void`.
+     *
+     * This function avoids the buffer-at-end check by writing the range in chunks, where a chunks has the size of
+     * the remaining space in the put area of the buffer.
+     * If the range type models `std::ranges::sized_range` the chunks are written using `std::ranges::copy_n`, which
+     * may use memcpy if applicable. Otherwise, a simple for loop iterates over the chunk.
+     *
+     * \attention You can only use the return value (end iterator) if your range type models
+     *            `std::ranges::forwarding_range`.
```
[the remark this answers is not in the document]
smehringer: I excluded the below overload for non-forwarding ranges from the documentation and adapted this documentation to fit both overloads.. I thought this was less confusing?

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +171,182 @@ class fast_istreambuf_iterator
...
+            return write_range(std::ranges::subrange<it_t, sen_t>{rng_it, rng_end});
+        }
+    }
+
+    //!\cond
+    // overload for non-forwarding_range types that return void
+    template <std::ranges::forward_range range_type>
+    void write_range(range_type && rng)
+    {
+        write_range(rng); // return value is ignored because iterator would be dangling
```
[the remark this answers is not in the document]
smehringer: I added a test

include/seqan3/range/detail/misc.hpp
```diff
@@ -40,4 +40,34 @@ template <std::ranges::forward_range rng_t>
...
+template <std::ranges::forward_iterator iterator_t, std::integral size_type, typename sentinal_t>
+//!\cond
+    requires std::ranges::sentinel_for<sentinal_t, iterator_t>
+//!\endcond
+std::pair<iterator_t, size_type> counting_next(iterator_t it, size_type const s, sentinal_t sen)
+{
+    size_type count = 0;
+    if constexpr (std::sized_sentinel_for<sentinal_t, iterator_t>)
+    {
+        count = std::min<size_type>(sen - it, s);
```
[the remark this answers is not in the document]
smehringer: thank you for pointing this out. this actually makes `counting_next` obsolete

include/seqan3/io/sequence_file/format_embl.hpp
smehringer: now done in iterator test since `write_number` is now a member function of the iterator

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +171,182 @@ class fast_istreambuf_iterator
...
+            throw std::ios_base::failure{"Cannot write to output stream (reached traits::eof() condition)."};
+
+        ++rng_it; // drop 1 character that has been written in overflow()
+
+        // Write remaining range recursively. For sized ranges, ensure that the subrange is also sized.
+        if constexpr (std::ranges::sized_range<range_type>)
+        {
+            // Note: Using ranges here because of bug https://github.com/ericniebler/range-v3/pull/1415 (fixed)
+            // TODO: change to std::ranges::subrange, when updating to range-v3 > 0.10.0
+            using subrange_t = ranges::subrange<it_t, sen_t, ranges::subrange_kind::sized>;
```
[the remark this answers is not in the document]
marehr: Still open

include/seqan3/io/stream/iterator.hpp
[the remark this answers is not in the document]
marehr: Not good :(

test/unit/io/stream/fast_streambuf_iterator_test.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+        EXPECT_EQ(rng_it, (rng.begin() + 5));
+        EXPECT_EQ(ostr.str(), std::string_view{"test\n"});
+
+        rng_it = it.write_range(std::ranges::subrange<str_it_t, str_it_t>(rng.begin() + 5, rng.end()));
+
+        EXPECT_EQ(rng_it, rng.end());
+        EXPECT_EQ(ostr.str(), std::string_view{"test\ntestest"});
+    }
+
```
marehr: a test that tests the overflow behaviour is missing

test/unit/io/stream/fast_streambuf_iterator_test.cpp
```diff
@@ -0,0 +1,170 @@
...
+TEST(fast_ostreambuf_iterator, write_number)
+{
+    std::ostringstream ostr{};
+    seqan3::detail::fast_ostreambuf_iterator<char> it{*ostr.rdbuf()};
+
+    uint64_t num{54389234};
+
+    it.write_number(num);
+
+    EXPECT_EQ(ostr.str().substr(0, 12), std::string_view{"54389234"});
```
marehr: a test that tests if buffer < 300

smehringer:
@marehr sad ping :(

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -245,34 +241,53 @@ class format_embl
...
+
+            // write end-of-record-symbol
+            stream_it.write_range(std::string_view{"//"});
+            stream_it.write_end_of_line(options.add_carriage_return);
         }
     }
+
+private:
+    //!\brief A buffer to convert numbers into characters with std::to_chars.
+    std::array<char, 50> arithmetic_buffer{};
```
marehr: unused?

include/seqan3/io/sequence_file/format_embl.hpp
```diff
@@ -37,9 +35,8 @@
 #include <seqan3/range/views/take_exactly.hpp>
 #include <seqan3/range/views/take_line.hpp>
 #include <seqan3/range/views/take_until.hpp>
-#include <seqan3/range/views/to.hpp>
+#include <seqan3/range/views/chunk.hpp>
```
marehr: is this include even used?

what about the other includes? there seems to be a lot redundancy.

include/seqan3/io/stream/iterator.hpp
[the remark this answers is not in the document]
marehr: Still open

include/seqan3/io/stream/iterator.hpp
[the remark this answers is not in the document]
marehr: Sorry for the noise, but I can't find the test (edit: Found it). And I still think that this construct is not possible.

edit: My guess is an infinite loop :+1:

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +172,213 @@ class fast_istreambuf_iterator
...
+            size_t const remaining_size = std::ranges::size(rng) - buffer_space - 1;
+            return write_range(subrange_t{rng_it, rng_end, remaining_size});
```
marehr: I just saw this, but why do you recurse (i.e. build up a call-stack) instead of a loop?

test/unit/io/stream/fast_streambuf_iterator_test.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+        EXPECT_EQ(rng_it, rng.end());
+        EXPECT_EQ(ostr.str(), std::string_view{"test\ntestest"});
+    }
+
+    { // write_range on a non-forwarding range returns void
+        std::ostringstream ostr{};
+        seqan3::detail::fast_ostreambuf_iterator<char> it{*ostr.rdbuf()};
+
+        EXPECT_TRUE((std::same_as<void, decltype(it.write_range(std::string{"foo"}))>));
```
marehr: Please write tests that executes something and see what happens, not just syntax checking.

My guess is an infinite loop :)

include/seqan3/io/sequence_file/format_embl.hpp
smehringer: good catch

include/seqan3/io/sequence_file/format_embl.hpp
smehringer: it is not used anymore

include/seqan3/io/stream/iterator.hpp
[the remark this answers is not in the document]
smehringer: Well 
1. since there is a test you can see that it works?
2. An lvalue is always a safe range so why should this ever cause an infinite loop as `rng` whithin the function scope is now an lvalue

test/unit/io/stream/fast_streambuf_iterator_test.cpp
smehringer: There is a test case here:
https://github.com/seqan/seqan3/pull/1472/files#diff-39898526460e7bc8f04d2d99e96e67feR109

smehringer: no infinite loop

smehringer: I added an test. no infinite loop

marehr: Where?

include/seqan3/io/stream/iterator.hpp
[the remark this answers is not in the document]
marehr: Boah this is not intuitive at all; can you add a comment that rng is a lvalue and hence the `safe range` overload is taken

include/seqan3/io/stream/iterator.hpp
smehringer: Yes you are right, I changed it into a for loop

test/unit/io/stream/fast_streambuf_iterator_test.cpp
smehringer: wait 'til I push

include/seqan3/io/stream/iterator.hpp
smehringer: This also makes the subrange workaround obsolete

include/seqan3/io/stream/iterator.hpp
```diff
@@ -166,6 +172,204 @@ class fast_istreambuf_iterator
...
+                it = copy_res.in;
+                stream_buf->pbump(characters_to_write);
+            }
+            else
+            {
+                for (size_t i = 0; it != end && i < buffer_space; ++it, ++i)
+                {
+                    *stream_buf->pptr() = *it;
+                    stream_buf->pbump(1);
+                }
```
h-2:
```cpp
                size_t i = 0;
                for (; it != end && i < buffer_space; ++it, ++i)
                    *stream_buf->pptr() = *it;

                stream_buf->pbump(i);
```

Only one virtual function call?

smehringer: pbump is not a virtual function call!

but it might be faster nonetheless

smehringer:
@marehr I think I have addressed all of your suggestions now.

We need to keep track of `safe_range` which has been renamed to `borrowed_range`, I'll make a card for this in the range module.

If you approve, do not merge this yet because I want to clean the commit history

PR #1478, opened 2020-01-10
[MISC] prepare span header for gcc10

marehr:


include/seqan3/std/span
```diff
@@ -19,7 +19,9 @@
 
 #if __has_include(<span>)
 #include <span>
+#endif
 
+#if defined(__cpp_lib_span)
 #include <seqan3/core/platform.hpp>
 #else // __has_include(<span>)
```
joergi-w: now you should adapt this comment... 💅

marehr: true xD

include/seqan3/std/span
```diff
@@ -19,9 +19,11 @@
 
 #if __has_include(<span>)
 #include <span>
+#endif
 
+#if defined(__cpp_lib_span)
 #include <seqan3/core/platform.hpp>
-#else // __has_include(<span>)
+#else // defined(__cpp_lib_span)
```
rrahn: I guess there must be a comment for the closing macro that has still __has_include?

include/seqan3/std/span
```diff
@@ -19,9 +19,11 @@
 
 #if __has_include(<span>)
 #include <span>
+#endif
```
rrahn: what if __has_include(<span>) is true but cpp_lib_span isn't? Can we guarantee that this never happens?

marehr: The header span is always available >= gcc10. cpp_lib_span will only be defined if you build in c++2a mode.

PR #1480, opened 2020-01-11
[FEATURE] Vectorised alignment with traceback

### LINKED ISSUE #1572: Add vectorised alignment computation of the traceback
(preparation for computing traceback in vectorised alignment)

implement two dimensional iterator wrapper that provides the two_dimensional_matrix_iterator but gets an additional parameter in order to select the simd position to give a filtered view over the vectorised traceback matrix that can be used to compute the traceback for one alignment only.

- [ ] write test for two dimensional iterator wrapper
- [ ] implement two dimensional iterator
- [ ] document two dimensional iterator

rrahn:
Enables computation of the front coordinate and alignment in simd mode.

Resolves #1572


rrahn:
Will be revisited after the more efficient kernel is implemented.

PR #1470, opened 2020-01-08
[FEATURE] Add read/write sequence record to BAM format

rrahn:
Adds bam format to sequence input/output file.

fixes #1418 

### Changes

* introduces a new crtp base class for sam and bam format to add the interface necessary for sequence file input/output
* fixes some issues detected within the bam format
* adds bam format to default formats of sequence input/output file
* refactors test structure of sequence file formats to reduce some code repetition
* adds test for sequence file bam format

PR #1494, opened 2020-01-16
[FIX] Throw exception instead of assertion

rrahn:
When the quality string has not the same size as the sequence a format_error should be thrown instead of an assertion.

Cherry-picked from #1470

### DIFF
--- a/include/seqan3/io/alignment_file/format_bam.hpp
+++ b/include/seqan3/io/alignment_file/format_bam.hpp
@@ -888,7 +888,11 @@ inline void format_bam::write_alignment_record([[maybe_unused]] stream_type &  s
         }
         else
         {
-            assert(static_cast<int32_t>(std::ranges::distance(qual)) == core.l_seq);
+            if (std::ranges::distance(qual) != core.l_seq)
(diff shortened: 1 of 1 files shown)


include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -888,6 +888,11 @@ inline void format_bam::write_alignment_record([[maybe_unused]] stream_type &  s
         }
         else
         {
+            if (std::ranges::distance(qual) != core.l_seq)
+                throw format_error{detail::to_string("Expected quality of same length as sequence with size ",
+                                                     core.l_seq, ". Got quality with size ",
+                                                     std::ranges::distance(qual), " instead.")};
+
             assert(static_cast<int32_t>(std::ranges::distance(qual)) == core.l_seq);
```
smehringer: then remove the assert

rrahn: Well, I left it because it doesn't hurt. But I don't mind removing it.

smehringer: But it will never be evaluated because the exception is thrown before. So it is unreachable anyway

PR #1496, opened 2020-01-16
[FIX] Fixes call of sse4 extract interface.

rrahn:


include/seqan3/core/simd/simd_algorithm.hpp
```diff
@@ -248,7 +248,7 @@ constexpr simd_t extract_eighth(simd_t const & src)
     if constexpr (simd_traits<simd_t>::length < 8) // In case there are less elements available return unchanged value.
         return src;
     else // if constexpr (simd_traits<simd_t>::max_length == 16) // SSE4
-        return extract_eighth_sse4(src);
+        return detail::extract_eighth_sse4<index>(src);
```
marehr: why are you adding `detail::` here but not on the other occurrences?

PR #1497, opened 2020-01-17
[FIX] Fix member difference_type for views::kmer_hash to use a type trait.

smehringer:
Appeared when using kmer hash on a span

### DIFF
--- a/include/seqan3/range/views/kmer_hash.hpp
+++ b/include/seqan3/range/views/kmer_hash.hpp
@@ -90,7 +90,7 @@ private:
          * \{
          */
         //!\brief Type for distances between iterators.
-        using difference_type = typename it_t::difference_type;
+        using difference_type = typename std::iter_difference_t<it_t>;
         //!\brief Value type of this iterator.
         using value_type = size_t;
         //!\brief The pointer type.


include/seqan3/range/views/kmer_hash.hpp
```diff
@@ -90,7 +90,7 @@ class kmer_hash_view : public std::ranges::view_interface<kmer_hash_view<urng_t>
          * \{
          */
         //!\brief Type for distances between iterators.
-        using difference_type = typename it_t::difference_type;
+        using difference_type = typename std::iterator_traits<it_t>::difference_type;
```
h-2:
```suggestion
        using difference_type = std::iter_difference_t<it_t>;
```

C++20-style traits should be used for iterators and ranges everywhere now.

smehringer: We should write that down in the wiki

rrahn: that would be great if you could do that.

smehringer: I added a card in the documentation project since the wiki is outdated.

PR #1501, opened 2020-01-17
[FIX] alignment pretty printing

h-2:
Seems like we don't have any tests for debug printing. Printing the alignment at least was broken.

I don't have time to add tests, but its maybe something to keep track of.

### DIFF
--- a/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
+++ b/include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
@@ -552,14 +552,17 @@ inline bool constexpr all_satisfy_aligned_seq<type_list<elems...>> = (aligned_se
  * \param alignment The alignment that shall be formatted. All sequences must be equally long.
  * \return          The given stream to which the alignment representation is appended.
  */
-template <tuple_like tuple_t, typename char_t>
+template <typename tuple_t, typename char_t>
 //!\cond
-    requires detail::all_satisfy_aligned_seq<detail::tuple_type_list_t<tuple_t>>
+    requires !std::ranges::input_range<tuple_t> &&
+             !alphabet<tuple_t> && // exclude alphabet_tuple_base
+             tuple_like<remove_cvref_t<tuple_t>> &&
+             detail::all_satisfy_aligned_seq<detail::tuple_type_list_t<remove_cvref_t<tuple_t>>>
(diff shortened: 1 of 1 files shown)


include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -552,14 +552,17 @@ inline bool constexpr all_satisfy_aligned_seq<type_list<elems...>> = (aligned_se
  * \param alignment The alignment that shall be formatted. All sequences must be equally long.
  * \return          The given stream to which the alignment representation is appended.
  */
-template <tuple_like tuple_t, typename char_t>
+template <typename tuple_t, typename char_t>
 //!\cond
-    requires detail::all_satisfy_aligned_seq<detail::tuple_type_list_t<tuple_t>>
+    requires !std::ranges::input_range<tuple_t> &&
+             !alphabet<tuple_t> && // exclude alphabet_tuple_base
```
h-2: this stuff is necessary, because the overall we are competing with has it.

rrahn:
could you add a card for this? That would be much appreciated.

include/seqan3/alignment/aligned_sequence/aligned_sequence_concept.hpp
```diff
@@ -552,14 +552,17 @@ inline bool constexpr all_satisfy_aligned_seq<type_list<elems...>> = (aligned_se
...
+             tuple_like<remove_cvref_t<tuple_t>> &&
+             detail::all_satisfy_aligned_seq<detail::tuple_type_list_t<remove_cvref_t<tuple_t>>>
 //!\endcond
-inline debug_stream_type<char_t> & operator<<(debug_stream_type<char_t> & stream, tuple_t const & alignment)
+inline debug_stream_type<char_t> & operator<<(debug_stream_type<char_t> & stream, tuple_t && alignment)
```
rrahn: Has this change something to do with the overload or simply to not promote the const-ness to the stored sequence types as we don't know anything about their const-iterability?

h-2: It's important for the overload, because `&&` is preferred over `const &` for temporaries and the other signature also has `&&`. 

In any case, it's also the right thing to do :tm: for generic interfaces.

PR #1504, opened 2020-01-20
[NOP] Version bump 3.0.1

smehringer:
Version bump that should have followed the 3.0.0 release

blocked by #1507

### DIFF
--- a/include/seqan3/version.hpp
+++ b/include/seqan3/version.hpp
@@ -21,7 +21,7 @@
 //!\brief The minor version as MACRO.
 #define SEQAN3_VERSION_MINOR 0
 //!\brief The patch version as MACRO.
-#define SEQAN3_VERSION_PATCH 0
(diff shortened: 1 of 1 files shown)


marehr:
:cry:

marehr:
We have to make the version in the documentation visible.

The documentation is not distinguishable from
* https://docs.seqan.de/seqan/3.0.0/ and
* https://docs.seqan.de/seqan/3.0.1/

Should I open an issue for this or just make a Card?

smehringer:
How do you want to do this?

rrahn:
We would need a mechanism to switch between versions like we had for SeqAn. Is that something doxygen might support?

h-2:
The SeqAn2 mechanism is just a small javascript thing that JongKyu wrote. It can probably be added easily to Doxygen, too.

PR #1507, opened 2020-01-20
[MISC] Exchange hard coded seqan version in html test with seqan3::seqan3_version.

smehringer:
This needs to be merged before bumping the version s.t. no test breaks


rrahn:
yeah, this is what I expected. That somewhere there is a reference to the hard coded version.

smehringer:
:grin:

PR #1526, opened 2020-01-24
[DOC] Add version selector to doxygen

eseiler:
Try it live https://docs.seqan.de/seqan/dev_script/

This would require a rebuild of the versioned documentations (3.0.0 and 3.0.1).

Screenshot 2020-01-24 at 18 59 21

### DIFF
--- a/test/documentation/seqan3_footer.html
+++ b/test/documentation/seqan3_footer.html
@@ -4,7 +4,7 @@
 <div id="nav-path" class="navpath"><!-- id is needed for treeview function! -->
   <ul>
     <div id="list_bottom_right" style="text-align:right; float:left; color:black; padding: 0px 0px 0px 10px"></div>
-    <script type="text/javascript" charset="utf-8" src="../../js/version.js"></script>
+    <script type="text/javascript" charset="utf-8" src="version.js"></script>
     $navpath
     <li class="footer">$generatedby
     <a href="http://www.doxygen.org/index.html">
--- a/test/documentation/seqan3_footer.html
+++ b/test/documentation/seqan3_footer.html
@@ -1,0 +1,23 @@
+<!-- HTML footer for doxygen 1.8.16-->
+<!-- start footer part -->
+<!--BEGIN GENERATE_TREEVIEW-->
+<div id="nav-path" class="navpath"><!-- id is needed for treeview function! -->
+  <ul>
+    <div id="list_bottom_right" style="text-align:right; float:left; color:black; padding: 0px 0px 0px 10px"></div>
+    <script type="text/javascript" charset="utf-8" src="../../js/version.js"></script>
+    $navpath
+    <li class="footer">$generatedby
+    <a href="http://www.doxygen.org/index.html">
+    <img class="footer" src="$relpath^doxygen.png" alt="doxygen"/></a> $doxygenversion </li>
+  </ul>
+</div>
+<!--END GENERATE_TREEVIEW-->
+<!--BEGIN !GENERATE_TREEVIEW-->
+<hr class="footer"/><address class="footer"><small>
+$generatedby &#160;<a href="http://www.doxygen.org/index.html">
+<img class="footer" src="$relpath^doxygen.png" alt="doxygen"/>
+</a> $doxygenversion
+</small></address>
+<!--END !GENERATE_TREEVIEW-->
+</body>
+</html>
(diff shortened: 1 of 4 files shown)


test/documentation/seqan3_footer.html
```diff
@@ -0,0 +1,23 @@
+<!-- HTML footer for doxygen 1.8.16-->
+<!-- start footer part -->
+<!--BEGIN GENERATE_TREEVIEW-->
+<div id="nav-path" class="navpath"><!-- id is needed for treeview function! -->
+  <ul>
+    <div id="list_bottom_right" style="text-align:right; float:left; color:black; padding: 0px 0px 0px 10px"></div>
+    <script type="text/javascript" charset="utf-8" src="../../js/version.js"></script>
```
eseiler: This is the actual non-standard doxygen stuff

marehr: Where is version.js? This should be within the repository.

eseiler: Well, currently the setup (pre-existing) is like that you have the `version.js` which fills the dropdown menu and switches pages when you select something.
The possible choices are parsed by `version.php` which crawls trough the directory, collects the folder names and returns a json.

It would probably work if you ship both with the library if that's what you want.

eseiler: It's maybe kinda hard to tell the script where to look for the subfolders of the different versions...

marehr: Yeah, please check all necessary files in :)

eseiler: Ok, done

test/documentation/version.js
```diff
@@ -0,0 +1,56 @@
...
+    version_select.addEventListener("change", function(){changeVersion(this.id);}, false);
+
+    // current selection is..
+    cur_sel = window.location.pathname.split("/")[2];
+
+    for(i=0; i < arr.length; ++i)
+    {
+        var op = document.createElement("option");
+        op.value = arr[i];
+        op.text = arr[i];
```
MitraDarja: Why is op.value the same as op.text?

eseiler: value is the internal "value" that is used when the option is selected in the dropdown menu. text is the label you can see in the dropdown menu.
In this case, both are the same since we want to show, e.g. "3.0.0" and also switch the URL to "3.0.0" once we select it.

test/documentation/version.php
```diff
@@ -0,0 +1,29 @@
...
+$list = array();
+foreach( $files as $file  )
+{
+    if( strpos($file, "master_") !== FALSE )
+        continue;
+    if( $file == "master")
+        continue;
+    if( strpos($file, "develop") !== FALSE )
+        continue;
+    if( $file[0] == "1")
```
smehringer: This is done to skip seqan versions 1 and 2 right? Add a comment either way.
Do we really want to also skip learning_resources?

test/documentation/version.php
```diff
@@ -0,0 +1,29 @@
+<?php
+/* Jongkyu Kim(j.kim@fu-berlin.de), 2016.01.12 */
```
smehringer: Maybe add "Adapted/Adjusted by Encrio Seiler" if you did adapt the script. Or "minor adaptions".

test/documentation/version.php
eseiler: Mhm, I don't know if we really need it in the API docs. the learning resources don't necessarily have something to do with SeqAn....

Also, it's rather easy to change afterwards by just replacing version.php on the server.

eseiler: Also, what is the order of the enries we would like?

Currently
```
3.0.1
3.0.0
3-master-user
3-master-dev
```

rrahn: I agree, that we don't want to display the learning resources in the SeqAn API documentation. The labels are based on the directories right? Some general thoughts: Since we have only one branch called master we don't need to call the documentation master. It should be rather `head` which reflects more that it is based on the latest commit. Second, the choices for user and dev bad I think. I propose to change the general url to:
```
http://docs.seqan.de/seqan3/library_user_head/
http://docs.seqan.de/seqan3/library_developer_head/
```
And accordingly change the old seqan link to seqan2 in the top level. That must be done however separately and maybe before we tackle the 2.4.1 release.

eseiler: I agree and would do another PR for the directory changes after this one is merged?

rrahn: yes, in a separate PR would be great.

PR #1528, opened 2020-01-25
search improvements

h-2:
NOT FOR MERGE

@eseiler some inspiration for performance tuning of the search. This PR improves `unidirectional_search_all/highErrorReadsSearch0` by 20% (impact on complex searches is expectedly smaller).

Left to do in this domain:
  * also make the vector-of-cursors an out-parameter (in the internal interface)
  * add a public search_buffers config element so that all of the buffers can be reused and no dynamic allocation happens in the search at all.

I expect the latter to have an even stronger influence in parallel contexts where allocations are notoriously slow.


SGSSGene:
Outdated, will be reevaluatet when we work on the search module.

PR #1529, opened 2020-01-26
[FIX] break up cyclic include of std/range and std/iterator

marehr:


include/seqan3/std/iterator
```diff
@@ -30,7 +30,6 @@
 #include <range/v3/iterator/traits.hpp>
 
 #include <seqan3/std/concepts>
-#include <seqan3/std/ranges>
```
rrahn: how are the above changes related to this dependency?

include/seqan3/core/type_traits/iterator.hpp
```diff
@@ -72,7 +72,7 @@ template <std::input_iterator it_t>
 struct rvalue_reference<it_t>
 {
     //!\brief Return the member type as return type.
-    using type = decltype(std::ranges::iter_move(std::declval<it_t &>()));
+    using type = std::iter_rvalue_reference_t<it_t>;
```
marehr: `<ranges>` pulls always `<iterator>` in.

But now, using `std::ranges::iter_move` in `<iterator>` has the downside that `iter_move` is only defined in `<ranges>`. This makes it a cyclic dependency (`<iterator>` pulls in `<ranges>` pulls in `<iterator>`). Using `std::iter_rvalue_reference_t` instead of `iter_move` has the improvement, that they are only pulling  `<iterator>` in, even though on cppreference it is declared as 

```c++
template< /*dereferenceable*/ T>
    requires /* see below */
using iter_rvalue_reference_t = decltype(ranges::iter_move(std::declval<T&>()));
```
https://en.cppreference.com/w/cpp/iterator/iter_t

The other changes were made to make all of those use the correct `std::` version.

A second step after this PR is to create a task to replace `rvalue_reference` and so on with the `std::` versions.

rrahn: Ah, sorry I overlooked the ranges::iter_move. Too early for me 😴  

>A second step after this PR is to create a task to replace rvalue_reference and so on with the std:: versions.

Yes, I wanted to get rid of these type trait wrappers and always use the std versions. This could be a good first issue for someone new to the library.

PR #1531, opened 2020-01-26
[TEST] Don't static_assert concepts

marehr:


test/unit/range/container/container_of_container_test.cpp
```diff
@@ -27,12 +27,9 @@ using container_of_container_types = ::testing::Types<std::vector<std::vector<dn
 
 TYPED_TEST_SUITE(container_of_container, container_of_container_types, );
 
-using In = std::ranges::iterator_t<seqan3::concatenated_sequences<seqan3::bitcompressed_vector<seqan3::dna4> > >;
-
-static_assert(std::readable<In>);
-
 TYPED_TEST(container_of_container, concepts)
 {
+    EXPECT_TRUE(std::readable<std::ranges::iterator_t<TypeParam>>);
```
MitraDarja: Probably, a stupid question, but why can you replace all the stuff ou had before in the std::ranges::iterator_t just with TypeParam?

marehr: I don't fully understand the question, but TypeParam will be any of these types (see container_of_container_types):
* `std::vector<std::vector<dna4>>`
* `concatenated_sequences<std::vector<dna4>>`
* `concatenated_sequences<bitcompressed_vector<dna4>>`

So the concept was checked only for the third instantiation, but this concept should be true for any of the given containers.

MitraDarja: Ah okay, thanks for the explainantion. :)

test/unit/range/container/container_of_container_test.cpp
```diff
@@ -27,12 +27,9 @@ using container_of_container_types = ::testing::Types<std::vector<std::vector<dn
 
 TYPED_TEST_SUITE(container_of_container, container_of_container_types, );
 
-using In = std::ranges::iterator_t<seqan3::concatenated_sequences<seqan3::bitcompressed_vector<seqan3::dna4> > >;
-
-static_assert(std::readable<In>);
```
smehringer: Do you know why this was done in the first place? Is this maybe a debugging left over? 

Also: I don't know if the container concept requires that the iterator on the container is readable? So this might just need to be removed without a replacement of an `EXPECT_TRUE``

marehr: Interestingly, `std::readable` was renamed to `std::indirectly_readable`

https://github.com/cplusplus/draft/commit/333b6eb8c147fce1a9ddcd905ee3bb1a09b91e30#diff-f762baff089b9a748fa0f3216c173577R83-R86

The definition of that concept is:

> Types that are indirectly readable by applying operator* model the indirectly_readable concept, including pointers, smart pointers, and iterators.

So this is always true for iterators.

> Do you know why this was done in the first place? Is this maybe a debugging left over?

Yeah, it slipped in through in this commit:

https://github.com/seqan/seqan3/commit/d8c1eacb0513e72780e175d41e684d16035c6861#diff-af991e93906e7b126071faaa91ac01b2R69-R72

which was part of pr #446.

I don't have hard feelings to keep this.

smehringer: @h-2 Can you maybe comment on this? I think it can be removed.

h-2: Yep, looks like debugging leftover!

PR #1536, opened 2020-01-26
[MISC] remove restraints on comparison operators of small string

marehr:
Fixes

```
/seqan3/test/unit/range/container/small_string_test.cpp: In member function ‘virtual void small_string_equality_Test::TestBody()’:
/seqan3/test/unit/range/container/small_string_test.cpp:249:49: error: no match for ‘operator==’ (operand types are ‘seqan3::small_string<4>’ and ‘seqan3::small_string<5>’)
  249 |     constexpr bool cmp3 = small_string{"hell"}  == small_string{"hello"};
      |                           ~~~~~~~~~~~~~~~~~~~~  ^~ ~~~~~~~~~~~~~~~~~~~~~
      |                           |                        |
      |                           small_string<4>          small_string<5>
In file included from /seqan3/include/seqan3/range/container/small_string.hpp:15,
                 from /seqan3/test/unit/range/container/small_string_test.cpp:14:
/seqan3/include/seqan3/range/container/small_vector.hpp:875:27: note: candidate: ‘template<long unsigned int cap2>  requires (‘integer_type’ not supported by expression) <= 5 constexpr bool seqan3::operator==(const seqan3::small_vector<char, 6>&, const seqan3::small_vector<char, cap2>&)’
  875 |     friend constexpr bool operator==(small_vector const & lhs, small_vector<value_type, cap2> const & rhs) noexcept
      |                           ^~~~~~~~
```

### DIFF
--- a/include/seqan3/range/container/small_vector.hpp
+++ b/include/seqan3/range/container/small_vector.hpp
@@ -869,9 +869,6 @@ public:
 
     //!\brief Performs element-wise comparison.
     template <size_t cap2>
-    //!\cond
-        requires cap2 <= capacity_ /* resolves ambiguousness when comparing two small_vectors of unequal capacity */
-    //!\endcond
     friend constexpr bool operator==(small_vector const & lhs, small_vector<value_type, cap2> const & rhs) noexcept
     {
         return std::ranges::equal(lhs, rhs);
@@ -879,9 +876,6 @@ public:
 
     //!\brief Performs element-wise comparison.
     template <size_t cap2>
-    //!\cond
-        requires cap2 <= capacity_
-    //!\endcond
     friend constexpr bool operator!=(small_vector const & lhs, small_vector<value_type, cap2> const & rhs) noexcept
     {
         return !(lhs == rhs);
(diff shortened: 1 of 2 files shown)


marehr:
Apparently:

https://github.com/seqan/seqan3/blob/979df093f202e14253df48fd489d70e3162362a5/test/unit/range/container/small_string_test.cpp#L245-L321

include/seqan3/range/container/small_vector.hpp
```diff
@@ -869,29 +869,20 @@ class small_vector
 
     //!\brief Performs element-wise comparison.
     template <size_t cap2>
-    //!\cond
-        requires cap2 <= capacity_ /* resolves ambiguousness when comparing two small_vectors of unequal capacity */
```
h-2: I am pretty sure these used to be necessary. How does the compiler resolve the ambiguity now? Is the operator of the LHS or the RHS taken?

I mean it's great if we can remove it, just making sure.

h-2:
> Apparently:

Can you change those tests to not `using namespace seqan3;` before removing this. It might behave differently because hidden friends et cetera.

include/seqan3/range/container/small_vector.hpp
marehr: I don't think so, because requires on friends never worked up until gcc 9.

`small_vector` is a complete type in this case and there is no other type specialisation or overload for small_vector.

marehr:
I removed the namespace, but like I thought it has no effect.

include/seqan3/range/container/small_vector.hpp
h-2: > I don't think so, because requires on friends never worked up until gcc 10.

You are right of course!

PR #1475, opened 2020-01-10
[INFRA] Add top level CMakeLists.txt

marehr:
Add a top-level CMakeLists.txt which exposes seqan3 targets (by calling `find_package(SeqAn3)`), defines `make install` of the seqan3 header-only library and defines how to package seqan3.

This resolves #1387.

PR #1543, opened 2020-01-28
[MISC] scaffold 3.0.2 changelog

marehr:


CHANGELOG.md
```diff
@@ -24,6 +24,48 @@ The following API changes should be documented as such:
 If possible, provide tooling that performs the changes, e.g. a shell-script.
 -->
 
+# 3.0.2
+
+Note that 3.1.0 will be the first API stable release and interfaces in this release might still change.
+
+## New features
+
+#### Alphabet
```
smehringer: I wouldn't already add all of the modules since we might not have any changes in one of them when we do more frequent releases.

marehr: I would give a full pre-fill, because merge conflicts will be less likely and before release we can remove not used headlines.

smehringer: This merge conflict, if it will appear, is really easy to fix so for me this reason is not strong enough to add stuff that we remove anyway. Since we have a tie I will reassign to @rrahn and he can decide.

rrahn: I agree. We should not add modules that are not likely to appear in the release. Especially, since I want to focus only one major topic for the next releases.

marehr:
Is this change okay?

rrahn:
superseded by #1475

PR #1547, opened 2020-01-28
[FIX] add sstream to charconv

marehr:
`charconv` uses `std::ostringstream`, but does not include it's header, see https://en.cppreference.com/w/cpp/io/basic_ostringstream.

https://github.com/seqan/seqan3/blob/979df093f202e14253df48fd489d70e3162362a5/include/seqan3/std/charconv#L875-L877

### DIFF
--- a/include/seqan3/std/charconv
+++ b/include/seqan3/std/charconv
@@ -31,6 +31,7 @@
 #include <cstring>
 #include <limits>
 #include <type_traits>
+#include <sstream>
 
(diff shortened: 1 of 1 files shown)


include/seqan3/std/charconv
```diff
@@ -31,6 +31,7 @@
 #include <cstring>
 #include <limits>
 #include <type_traits>
+#include <sstream>
```
joergi-w: Please sort alphabetically.

PR #1553, opened 2020-02-04
[FEATURE] Increase I/O performance by using a larger buffer when reading from file.

smehringer:
This makes it also visible that we might want a CRTP base class for files since I had to copy and paste quite some code..

### DIFF
--- a/include/seqan3/io/alignment_file/input.hpp
+++ b/include/seqan3/io/alignment_file/input.hpp
@@ -582,7 +582,7 @@ public:
      */
     alignment_file_input(std::filesystem::path filename,
                          selected_field_ids const & SEQAN3_DOXYGEN_ONLY(fields_tag) = selected_field_ids{}) :
-        primary_stream{new std::ifstream{filename, std::ios_base::in | std::ios::binary}, stream_deleter_default}
+        primary_stream{new std::ifstream{}, stream_deleter_default}
     {
         init(filename);
     }
@@ -658,7 +658,7 @@ public:
                          typename traits_type::ref_ids & ref_ids,
                          typename traits_type::ref_sequences & ref_sequences,
                          selected_field_ids const & SEQAN3_DOXYGEN_ONLY(fields_tag) = selected_field_ids{}) :
-        primary_stream{new std::ifstream{filename, std::ios_base::in | std::ios::binary}, stream_deleter_default}
+        primary_stream{new std::ifstream{}, stream_deleter_default}
     {
         // initialize reference information
         set_references(ref_ids, ref_sequences);
@@ -848,6 +848,9 @@ protected:
     //!/brief Initialisation based on a filename.
     void init(std::filesystem::path & filename)
     {
+        primary_stream->rdbuf()->pubsetbuf(stream_buffer.data(), stream_buffer.size());
+        static_cast<std::basic_ifstream<char> *>(primary_stream.get())->open(filename,
+                                                                             std::ios_base::in | std::ios::binary);
         // open stream
         if (!primary_stream->good())
             throw file_open_error{"Could not open file " + filename.string() + " for reading."};
@@ -875,6 +878,8 @@ protected:
      */
     //!\brief Buffer for a single record.
     record_type record_buffer;
+    //!\brief A larger (compared to stl default) stream buffer to use when reading from a file.
+    std::vector<char> stream_buffer{std::vector<char>(1'000'000)};
     //!\}
 
     /*!\name Stream / file access
--- a/include/seqan3/io/alignment_file/input.hpp
+++ b/include/seqan3/io/alignment_file/input.hpp
@@ -584,7 +584,7 @@ public:
                          selected_field_ids const & SEQAN3_DOXYGEN_ONLY(fields_tag) = selected_field_ids{}) :
         primary_stream{new std::ifstream{}, stream_deleter_default}
     {
-        init(filename);
+        init_by_filename(std::move(filename));
     }
 
     /*!\brief Construct from an existing stream and with specified format.
@@ -615,7 +615,7 @@ public:
                          selected_field_ids const & SEQAN3_DOXYGEN_ONLY(fields_tag) = selected_field_ids{}) :
         primary_stream{&stream, stream_deleter_noop}
     {
-        init(file_format{});
+        init_by_format<file_format>();
     }
 
     //!\overload
@@ -628,7 +628,7 @@ public:
                          selected_field_ids const & SEQAN3_DOXYGEN_ONLY(fields_tag) = selected_field_ids{}) :
         primary_stream{new stream_t{std::move(stream)}, stream_deleter_default}
     {
-        init(file_format{});
+        init_by_format<file_format>();
     }
 
     /*!\brief Construct from filename and given additional reference information.
@@ -663,7 +663,7 @@ public:
         // initialize reference information
         set_references(ref_ids, ref_sequences);
 
-        init(filename);
+        init_by_filename(std::move(filename));
     }
 
     /*!\brief Construct from an existing stream and with specified format.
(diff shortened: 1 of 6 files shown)


include/seqan3/io/sequence_file/input.hpp
```diff
@@ -357,6 +357,8 @@ class sequence_file_input
                   }(),
                   "You may not select field::seq_qual and either of field::seq and field::qual at the same time.");
 
+    // static_assert(detail::type_list_of_sequence_file_input_formats<valid_formats>, "");
```
marehr: ?

include/seqan3/io/alignment_file/input.hpp
```diff
@@ -875,6 +878,8 @@ class alignment_file_input
      */
     //!\brief Buffer for a single record.
     record_type record_buffer;
+    //!\brief A larger (compared to stl default) stream buffer to use when reading from a file.
+    char stream_buffer[1'000'000];
```
marehr: Not sure if we want a stack buffer of that size. Did you double check how seqan2 did this?

include/seqan3/io/alignment_file/input.hpp
```diff
@@ -848,6 +848,9 @@ class alignment_file_input
     //!/brief Initialisation based on a filename.
     void init(std::filesystem::path & filename)
```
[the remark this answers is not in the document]
smehringer: > this should have a different name init_by_filename and the other one should be called init_by_format

yes that would be better. Naming is not my best skill :see_no_evil: 

> Why is this not directly done in the constructor?

Because I have 5 or 6 constructors that need 2 types of initialisation.

include/seqan3/io/sequence_file/input.hpp
smehringer: ops. left over debugging

include/seqan3/io/alignment_file/input.hpp
smehringer: @rrahn said this is what seqan2 is doing (1 MB buffer) :woman_shrugging: but I did not confirm again. I also wanted to have this PR so we can talk about it but I thought it is good if somebody already has a close look at it because I was unsure if that's the way to do it.

include/seqan3/io/alignment_file/input.hpp
```diff
@@ -658,7 +658,7 @@ class alignment_file_input
                          typename traits_type::ref_ids & ref_ids,
                          typename traits_type::ref_sequences & ref_sequences,
                          selected_field_ids const & SEQAN3_DOXYGEN_ONLY(fields_tag) = selected_field_ids{}) :
-        primary_stream{new std::ifstream{filename, std::ios_base::in | std::ios::binary}, stream_deleter_default}
+        primary_stream{new std::ifstream{}, stream_deleter_default}
```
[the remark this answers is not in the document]
smehringer: I don't understand the suggestion. You remove the `std::function` from the type, but we need this to be a function object so we can manipulate whether the pointer is owning or not?

include/seqan3/io/alignment_file/input.hpp
h-2: Yes, that's also what I suggested. Stack buffer is completely fine, but I would still use `std::array` because that's always preferable.

h-2:
Do you have numbers how this affects performance?

smehringer:
> Do you have numbers how this affects performance?

Actually not as much as I thought in my single (!) macro benchmark: 
fasta input 1,3GB `4.25s` to `4.15s` ... 
Micro benchmarks are only tested on streams. I didn't have time yet to do more extensive benchmarks but the profiler showed a 20% less amount of time spend in the if condition of the `operator++` in the istream_buf_iterator. Main performance bottleneck now is puch_back on the sequence (which is improved by your clear function in the fastq PR that I cherry picked). We are now at the same speed fas seqan2 for 1,3 GB fasta data input (10x human chr10) and even a little faster on 1 GB fasta input with small 500bp reads.
But as I said, more extensive benchmarking would be needed.

include/seqan3/io/alignment_file/input.hpp
smehringer: I changed it to a heap buffer since the performance did not drop by the change.

rrahn:
> This makes it also visible that we might want a CRTP base class for files since I had to copy and paste quite some code..

This doesn't need to be a CRTP. It seems that all files have the streams so storing them and intialising them can be all done by regular non-virutal base class i guess.

smehringer:
I am reassigning the review since @marehr is still sick

smehringer:
@rrahn polite ping

include/seqan3/io/alignment_file/input.hpp
rrahn: So I looked more closely into the SeqAn implementation and it seems that the default buffer size changes quite a lot for different stream implementations. But nonethless I believe using heap or stack is not important here. But I am in favour of heap memory at this point.

include/seqan3/io/alignment_file/input.hpp
```diff
@@ -875,6 +878,8 @@ class alignment_file_input
      */
     //!\brief Buffer for a single record.
     record_type record_buffer;
+    //!\brief A larger (compared to stl default) stream buffer to use when reading from a file.
+    std::vector<char> stream_buffer{std::vector<char>(1'000'000)};
```
rrahn: Why not directly?
```suggestion
    std::vector<char> stream_buffer(1'000'000);
```

include/seqan3/io/alignment_file/input.hpp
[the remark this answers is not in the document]
rrahn: But it was not renamed. I think it can be renamed in this PR. That is a small refactorisation that makes sense in this context.

include/seqan3/io/alignment_file/input.hpp
smehringer: @rrahn This syntax is not allowed by the compiler because it can be [ambigous](https://stackoverflow.com/questions/24836526/why-c11-in-class-initializer-cannot-use-parentheses) in some situations.

rrahn: right. that thing. Only this would work but does not make it better:
```cpp
std::vector<char> stream_buffer{1'000'000, std::allocator<char>{}};
```
So leave it as is.

include/seqan3/io/alignment_file/input.hpp
```diff
@@ -846,7 +846,7 @@ class alignment_file_input
     //!\privatesection
 
     //!/brief Initialisation based on a filename.
-    void init(std::filesystem::path & filename)
+    void init_by_filename(std::filesystem::path & filename)
```
rrahn: why is the path captured by reference? 
The constructor doesn't do anything with this right?
So should it not be `const &`?

smehringer: oh I actually do something. within `make_secondary_stream` the filename might be stripped from its compression extension!

rrahn: but then it should be a copy. You only want to strip it internally I guess?

smehringer: It is already copied by the constructor so it's safe to capture it by reference in the `init` function?

rrahn: But then someone modifies the constructor and does not know that the variable was changed. So this function should make clear, that it does not change the path or it documents that it might.

smehringer: Alright. Any preference?

rrahn: By copy and then move it in from the calling function would be my preference.

smehringer: ok

smehringer: done

smehringer:
@marehr Can you please approve the PR? Otherwise merging is blocked. I had to reassign the review due to your sick leave.

smehringer:
@marehr ping

Issue #1556, opened 2020-02-05
clang-10 support


marehr:
Bugfixes:

Reported Compiler Bugs:
* [ ] [trunk] Assertion `!isValueDependent() && "Expression evaluator can't be called on a dependent expression."' failed. (https://github.com/saarraz/clang-concepts-monorepo/issues/54)

smehringer:
@seqan/seqan this is a test

PR #1558, opened 2020-02-05
[FIX] clang10: struct / class should be the same in all declarations

marehr:
```
seqan3/include/seqan3/core/char_operations/predicate_detail.hpp:162:1: error: 'char_predicate_base' defined as a struct template here but previously declared as a class template; this is valid, but may result in linker errors under the Microsoft C++ ABI [-Werror,-Wmismatched-tags]
struct char_predicate_base
^
seqan3/include/seqan3/core/char_operations/predicate_detail.hpp:94:1: note: did you mean struct here?
class char_predicate_base;
^~~~~
struct
```

Part of `clang10-support` #1556

### DIFF
--- a/include/seqan3/core/char_operations/predicate_detail.hpp
+++ b/include/seqan3/core/char_operations/predicate_detail.hpp
@@ -91,7 +91,7 @@ inline const std::string condition_message_v
 
 //!\cond
 template <typename condition_t>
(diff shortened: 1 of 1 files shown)


smehringer:
Travis is fine, it just did not update this PR, so I am merging this anyway

PR #1563, opened 2020-02-06
[MISC] Add definition of done on feature request.

rrahn:
@seqan/core 

Please have a look on the DoD addons to feature requests.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -19,3 +19,15 @@ One of the following will happen to feature requests:
     see the [contributor guide](https://github.com/seqan/seqan3/pull/CONTRIBUTING.md).
   * The feature request is invalid and will be closed.
 -->
+
+### Done
+
+- [ ] Implementation and design approved
+- [ ] Unit tests pass
+- [ ] Test coverage > 97%
+- [ ] Micro benchmarks implemented
+- [ ] API documentation added
+- [ ] Tutorial/teaching material added
+- [ ] Macro benchmark with <= 10 % performance difference to SeqAn2 (if applicable)
+- [ ] Tests compile in less than 30 seconds
+- [ ] Changelog entry added


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -19,3 +19,15 @@ One of the following will happen to feature requests:
     see the [contributor guide](https://github.com/seqan/seqan3/pull/CONTRIBUTING.md).
   * The feature request is invalid and will be closed.
 -->
+
+### Done
+
+- [ ] Implementation and design approved
+- [ ] Unit tests pass
+- [ ] Test coverage > 97%
```
eseiler: Why only 97?
Since we can exclude "unreachable" code (e.g. malloc throws), there isn't really a reason we shouldn't strive for 100%?

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -19,3 +19,15 @@ One of the following will happen to feature requests:
...
+- [ ] Micro benchmarks implemented
```
eseiler:
```suggestion
- [ ] Microbenchmarks implemented
```

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -19,3 +19,15 @@ One of the following will happen to feature requests:
...
+- [ ] API documentation added
+- [ ] Tutorial/teaching material added
+- [ ] Macro benchmark with <= 10 % performance difference to SeqAn2 (if applicable)
```
eseiler:
```suggestion
- [ ] Macrobenchmark with <= 10 % performance difference to SeqAn2 (if applicable)
```

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -19,3 +19,15 @@ One of the following will happen to feature requests:
...
+- [ ] Tests compile in less than 30 seconds
```
eseiler: On what machine? :) Also the Debug tests or only in Release?

smehringer:
In general, I would rather have this as a PR template. The issue templates are for external users that know nothing about our task structure and might be confused about this list (e.g. they might think they need to edit this list to tell us when they think the request is done for them personally). Also not every external feature request is directly worked on so the list might be outdated once we actually work on  it.

.github/ISSUE_TEMPLATE/feature_request.md
rrahn: Well, ok we can specify a mode Debug/Release. The question would be what does @seqan/core think about 30 seconds. Is it too much or too low?

.github/ISSUE_TEMPLATE/feature_request.md
rrahn: Yes, we could have 100%, but not in all cases. There are special cases with the SIMD tests and SIMD alignment tests which cannot reach 100% as long as vectorised test infrastructure is not incorporated. Thus, marking the code unreachable is wrong and error prone since it is easy to forget that the things need to be removed once the infrastructure is added. As an alternative we could make 100% and only in the exceptional case where we know we can reach 100% but not yet, we allow a lower threshold. I just want to avoid as many exceptions as possible since it is important to adhere to the DoD.

.github/ISSUE_TEMPLATE/feature_request.md
rrahn: > On what machine? :) 

Your desktop/laptop computer for a single test case. I guess the differences between the CPUs aren't that big between the developers in the team. But if required we might find a minimum CPU spec and use that. But these things should not become to complicated, otherwise they will be useless if the team starts to ignore them.

rrahn:
> In general, I would rather have this as a PR template. The issue templates are for external users that know nothing about our task structure and might be confused about this list (e.g. they might think they need to edit this list to tell us when they think the request is done for them personally). Also not every external feature request is directly worked on so the list might be outdated once we actually work on it.

Some points on this:
1. Making it a as template for a PR is the wrong place. The flow of work is based on a story (in our case a GitHub issue). This story contains all important information that helps the developers to implement the feature including the acceptance criteria. A very important concept in this is the DoD. This is something the team is held accountable for so that the feature is really shippable and satisfies the requirements. Thus, and as can be seen by the list, in most cases a PR  solves only parts of an issue and therefore should not contain the DoD. 
2. The feature template is not only for external user but also thought for internal use. I will also refine the structure and the content a little to reflect more the required information for a feature request/user story.
I do agree, that the current format does not clearly determine what the responsibilities for the issue creator are. A solution would be to have it in a collapsable block that indicates for dev only. Another could be an automated commenting feature, that will add this once the issue was created. However, that can be done later and I don't see at the moment no better way to handle this automatically using GitHub as the ticket management system.
3. The DoD is seen as one of the most important things within the development process. Because a feature is only done, when it is DONE. The current list suggests, albeit we made a lot of great progress on the modules, that we haven't reached that DoD (as defined above) for many things yet. This has many implications especially regarding our deliverable of a stable 3.1 release. I am against putting it only somewhere into the wiki as it should be visible for the team, especially with the current lack of experience and proficiency regarding agile development.
As an alternative I could prepare offline versions that can be placed in each office. But I feel, that this is only a sub-optimal solution at the moment.
4. Regarding the outdated list. Yes, the list might be outdated at some point if the issue is stalled and the DoD did change in the meantime. However, we also agreed on converting feature requests to cards if it doesn't fit into the short-term planning. Thus, by reopening, we might as well get the list up-to-date and then work on this shortly after. Anyhow, with the current system it will always require some hand-made adaptions I am afraid. 

Another alternative could be to manually add the DoD list to the story when it is added to the backlog (tackling it in the next iterations). This would be the time were additional requirements are added (acceptance criteria). But I believe for the reasons I mentioned before, that in the current state an automated solution would be better. And honestly, how many external feature request did we get in the past? 

Having said that, neither the DoD nor the place of it are ever fixed. As part of a regular reflection on the teams progress also the DoD should be updated to meet the requirements accordingly. This is more like an initial version which will evolve over time. Also a different issue types should have different DoDs. For example a lot of the things are for some bug fixes probably not so relevant.

If you ask me for the optimal solution, then a separate block indicating the DoD rules should be showed on the ticket which is linked to a template. If the template is changed all referencing lists will update or notify about the change. The issue can't be closed as long as items on the list are not checked (This might not be necessary in case it hinders the development process, if things can't be automatically closed if points are not checked before - but again there is the step of reflecting about these things in order to adapt the workflow).
If someone knows about a good solution to get this with GitHub I would be more than happy.
Or maybe some other mechanism we could use?
Anyway, since this got quite long now I propose to discuss this in the office together on Monday.

.github/ISSUE_TEMPLATE/feature_request.md
eseiler: Well, then let's do 100% and allow a lower threshold for SIMD...

rrahn:
I am closing this in favor of the new product backlog repository. I added the information for now on the wiki.

PR #1578, opened 2020-02-07
[FEATURE] Improve performance by calling clear() on containers in a record.

### LINKED ISSUE #1579: Use clear function when clearing elements of the file record if possible.
Check the performance if clear is called on the container types stored within file record instead of intialising them in each step.

smehringer:
This increases performance since we avoid new memory allocation when default constructing a new sequence container. Additionally, if the sequence length for each record does not change, push_back is more efficient because it never has to resize the container.

Resolves #1579

### DIFF
--- a/include/seqan3/io/record.hpp
+++ b/include/seqan3/io/record.hpp
@@ -224,6 +224,27 @@ struct fields
 template <typename field_types, typename field_ids>
 struct record : detail::transfer_template_args_onto_t<field_types, std::tuple>
 {
+private:
+    //!\brief Auxiliary functions for clear().
+    template <typename t>
+    //!\cond
+        requires requires (t & v) { v.clear(); }
+    //!\endcond
+    static constexpr void clear_element(t & v) noexcept(noexcept(v.clear()))
+    {
+        v.clear();
+    }
+
+    //!\overload
+    template <typename t>
+    static constexpr void clear_element(t & v) noexcept(noexcept(std::declval<t &>() = t{}))
+    {
+        v = {};
+    }
+
+    //!\brief A lambda function that expands a pack and calls `clear_element` on every argument in the pack.
+    static constexpr auto expander = [] (auto & ...args) { (clear_element(args), ...); };
+
 public:
     //!\brief A specialisation of std::tuple.
     using base_type = detail::transfer_template_args_onto_t<field_types, std::tuple>;
@@ -231,12 +252,12 @@ public:
     /*!\name Constructors, destructor and assignment
      * \{
      */
-    record() = default;                           //!< Defaulted
-    record(record const &) = default;             //!< Defaulted
-    record & operator=(record const &) = default; //!< Defaulted
-    record(record &&) = default;                  //!< Defaulted
-    record & operator=(record &&) = default;      //!< Defaulted
-    ~record() = default;                          //!< Defaulted
+    record() = default;                           //!< Defaulted.
+    record(record const &) = default;             //!< Defaulted.
+    record & operator=(record const &) = default; //!< Defaulted.
+    record(record &&) = default;                  //!< Defaulted.
+    record & operator=(record &&) = default;      //!< Defaulted.
+    ~record() = default;                          //!< Defaulted.
 
     //!\brief Inherit std::tuple's constructors.
     using base_type::base_type;
@@ -245,17 +266,10 @@ public:
     static_assert(field_types::size() == field_ids::as_array.size(),
                   "You must give as many IDs as types to seqan3::record.");
 
-    //!\brief (Re-)Initialise all tuple elements with `{}`.
-    void clear()
+    //!\brief Clears containers that provide `.clear()` and (re-)initialises all other elements with `= {}`.
+    void clear() noexcept(noexcept(std::apply(expander, std::declval<record &>())))
     {
-        clear_impl(*this, std::make_integer_sequence<size_t, field_types::size()>{});
-    }
-private:
-    //!\brief Auxiliary function for clear().
-    template <size_t ...Is>
-    constexpr void clear_impl(base_type & tup, std::integer_sequence<size_t, Is...> const &)
-    {
-        ((std::get<Is>(tup) = {}), ...);
+        std::apply(expander, *this);
     }
 };
 


include/seqan3/io/record.hpp
```diff
@@ -246,16 +246,34 @@ struct record : detail::transfer_template_args_onto_t<field_types, std::tuple>
...
-        ((std::get<Is>(tup) = {}), ...);
+        (clear_impl2(std::get<Is>(tup)), ...);
+    }
+
+    //!\brief Auxiliary functions for clear().
+    template <typename t>
+    //!\cond
+        requires requires (t & v) { v.clear(); }
+    //!\endcond
+    static constexpr void clear_impl2(t & v) noexcept
```
h-2: needs to be conditional noexcept, too. We know nothing about the function, other than what its called.

include/seqan3/io/record.hpp
```diff
@@ -246,16 +246,34 @@ struct record : detail::transfer_template_args_onto_t<field_types, std::tuple>
...
+    {
+        v.clear();
+    }
+
+    //!\brief Auxiliary functions for clear().
+    template <typename t>
+    static constexpr void clear_impl2(t & v) noexcept(noexcept(t{}))
```
h-2:
```suggestion
    static constexpr void clear_impl2(t & v) noexcept(noexcept(std::declval(t &) = t{}))
```
Even if the default constructor is noexcept, the assignment needn't be.

smehringer:
@h-2 If now everything is conditional, wouldn't it be fine to just not declare them noexcept?

h-2:
> @h-2 If now everything is conditional, wouldn't it be fine to just not declare them noexcept?

I don't know, it might affect optimisation possibilities.

include/seqan3/io/record.hpp
```diff
@@ -246,16 +246,34 @@ struct record : detail::transfer_template_args_onto_t<field_types, std::tuple>
                   "You must give as many IDs as types to seqan3::record.");
 
     //!\brief (Re-)Initialise all tuple elements with `{}`.
```
rrahn: documentation is not correct anymore.

include/seqan3/io/record.hpp
```diff
@@ -246,16 +246,34 @@ struct record : detail::transfer_template_args_onto_t<field_types, std::tuple>
...
-    void clear()
+    void clear() noexcept(noexcept(clear_impl(*this, std::make_integer_sequence<size_t, field_types::size()>{})))
     {
         clear_impl(*this, std::make_integer_sequence<size_t, field_types::size()>{});
```
rrahn:
```suggestion
        std::apply([this] (auto & ...args) { (clear_element(args),...); }, *this);
```
Then with custom detail concept:
```cpp
template <typename t>
SEQAN3_CONCEPT has_clear_member_function = requires (t & e) { {e.clear() } -> void; };
```

and proper private member function
```cpp
template <typename element_t>
void clear_element(element_t & element)
{
    if constexpr (has_clear_member_function<element_t>)
        element.clear();
    else
        element = element_t{};
}
```

rrahn:
@smehringer I will make a separate issue for this. You linked it to an epic which will collect all the separate improvements.
Can you then please provide on the respective issue the numbers that show the improvement?

smehringer:
> @smehringer I will make a separate issue for this. You linked it to an epic which will collect all the separate improvements.

Sorry, I was not aware that we do Epics again.

> Can you then please provide on the respective issue the numbers that show the improvement?

This would only be possible if we would have a macro-test infrastructure s.t. I can link against reproducible tests. Right now I could only tell you that in increased performance by x% (I would need to check) in my single macro test.
I would rather collect all improvements and then check before the next release how much the speed improved and then state the overall benefit instead of adding every small improvement?

include/seqan3/io/record.hpp
smehringer: How am I suppose to adjust the conditional noexcept in this version of a combined `clear_element` ?

h-2: Why do we need an extra concept if this is only used on one place? I thought the policy was to only externalise these if they are used in multiple places.

I also think that it is not possible to easily define `noexcept` with the `if constexpr`. I like the `std::apply`, though. But in general, maybe we should spend too much time discussed this mini-PR?

include/seqan3/io/record.hpp
```diff
@@ -224,19 +224,40 @@ struct fields
...
+    template <typename t>
+    //!\cond
+        requires requires (t & v) { v.clear(); }
+    //!\endcond
+    static constexpr void clear_element(t & v) noexcept(noexcept(v.clear()))
+    {
+        v.clear();
+    }
+
+    //!\brief Auxiliary functions for clear().
```
eseiler: Why not 

```suggestion
    //!\overload
```

include/seqan3/io/record.hpp
```diff
@@ -224,19 +224,40 @@ struct fields
...
+    template <typename t>
+    static constexpr void clear_element(t & v) noexcept(noexcept(std::declval<t &>() = t{}))
+    {
+        v = {};
+    }
+
+    //!\brief A lambda function that expands a pack, calling clear_element on every argument in the pack.
```
eseiler:
```suggestion
    //!\brief A lambda function that expands a pack and calls `clear_element` on every argument in the pack.
```

include/seqan3/io/record.hpp
```diff
@@ -224,19 +224,40 @@ struct fields
...
+    static constexpr auto expander = [] (auto & ...args) { (clear_element(args),...); };
```
eseiler:
```suggestion
    static constexpr auto expander = [] (auto & ...args) { (clear_element(args), ...); };
```

include/seqan3/io/record.hpp
eseiler: Still open?

test/snippet/alphabet/to_char.cpp
```diff
@@ -1,11 +1,129 @@
-#include <seqan3/alphabet/concept.hpp>
-#include <seqan3/alphabet/adaptation/char.hpp>
-#include <seqan3/alphabet/nucleotide/dna5.hpp>
+#include <bitset>
```
eseiler: 🙃

include/seqan3/io/record.hpp
rrahn: Well in C++20 we don't need an extra concept as this:
```cpp
    if constexpr (requires (element_t & e) { {e.clear()} -> std::same_as<void>;})
        element.clear();
    else
        element = element_t{};
```
would be valid code. But, well, that is future.

PR #1590, opened 2020-02-13
[INFRA] add .gitpod.yml to checkout and build seqan3 in an online editor

### LINKED ISSUE #1589: Provide .gitpod.yml to try out seqan3 in an online editor
See https://gitpod.io.

Screenshot from 2020-02-13 15-05-44

marehr:
Fixes #1589

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,3 @@
+tasks:
+  - init: git submodule init && git submodule update
+    command: git submodule update


smehringer:
Can you shortly explain the two lines?

.gitpod.yml
```diff
@@ -0,0 +1,3 @@
+tasks:
```
marehr: https://www.gitpod.io/docs/config-start-tasks/

.gitpod.yml
```diff
@@ -0,0 +1,3 @@
...
+  - init: git submodule init && git submodule update
```
marehr: https://www.gitpod.io/docs/config-start-tasks/#code-classlanguage-textinitcode-command

.gitpod.yml
```diff
@@ -0,0 +1,3 @@
...
+    command: git submodule update
```
marehr: https://www.gitpod.io/docs/config-start-tasks/#defining-commands

smehringer: hm, why would you always want to call `git submodule update` every time? I see that the init command is necessary but the submodule won't change unless a new version is checked out?

marehr: I followed https://www.gitpod.io/docs/cpp-in-gitpod/

which suggests

https://github.com/Circuito-io/ComponentEditor/blob/master/.gitpod.yml

I don't see why this hurts, because this will only be executed if you startup the workspace.

smehringer: alright

PR #1599, opened 2020-02-18
[FEATURE] Added *.fas file extension to FASTA files

cpockrandt:


h-2:
Out of curiosity: is this common?

You know that you can simply do the following somewhere in your program:

```cpp
seqan3::fasta_format::file_extensions.push_back("fas");
```

And it will work for you?

rrahn:
What I found on the internet:
>What is a FAS file?
Compiled script written in AutoLISP, a variation of the LISP programming language used specifically for the extension of AutoCAD software; often used for creating macros that automate common processes within AutoCAD; can be used for distributing utilities that add new features to the software.

🤣 
Anyhow, there is rather spare information on this format extension. I found one thread in SeqAnswers saying it is a common format?! @cpockrandt can you name the databases or other useful information of where this extension is used regularly? So when adding it to the vaild formats we can at least give user documentation accordingly.

smehringer:
First thing when I looked for FASTA extension `.fas` : https://filext.com/file-extension/FAS
Other sources mentioning the extension or use it in a FASTA context (seems common to me as nobody is like "huh, fas?"): 
* http://justsolve.archiveteam.org/wiki/FASTA_and_FASTQ
* http://seqanswers.com/forums/showthread.php?t=21264
* https://www.researchgate.net/post/How_do_I_merge_several_multisequence-fasta_files_to_create_one_tree_for_subsequent_Unifrac_analysis

rrahn:
But since @cpockrandt seemes to have such a format I was just curious of where it is coming from. Is it a database or another tool etc.. I mean It could be of course just because of personal preferences of some single person so nothing relevant. I mean I have no objections of adding this extension.

cpockrandt:
A user of genmap made me aware of it and it seems to be a common file ending (maybe not widely used).

I don't see any reason why we shouldn't add it to SeqAn3. I don't think it should be the app developer's responsibility to add less common but still used file endings to each of his/her/their apps. 

... assuming the argument parser won't print a huge list of possible file endings and make the help page less readable.

rrahn:
@cpockrandt thanks for the info. As I said I don't see any reasons not to add this. I just wanted to know if there is a greater community behind that file ending because of regular use of databases etc. Anyhow, thanks for contributing :D

Issue #1571, opened 2020-02-07
I/O Performance


rrahn:
Evaluate different possibilities to improve the overall I/O performance when parsing files.

PR #1628, opened 2020-03-05
[FEATURE] Increase reading FASTA performance by reducing an if-check.

smehringer:
This improves reading a FASTA file by roughly 2-3% (1GB macro benchmark).
Not a lot but we might as well include it.

Belongs to #1571

### DIFF
--- a/include/seqan3/io/sequence_file/format_fasta.hpp
+++ b/include/seqan3/io/sequence_file/format_fasta.hpp
@@ -276,10 +276,11 @@ private:
             auto e = stream_view.end();
             for (; (it != e) && ((!is_id)(*it)); ++it)
             {
-                if ((is_space || is_digit)(*it))
-                    continue;
-                else if (not_in_alph(*it))
+                if (not_in_alph(*it))
                 {
+                    if ((is_space || is_digit)(*it))
+                        continue;
+
                     throw parse_error{std::string{"Encountered an unexpected letter: "} +
                                         not_in_alph.msg +
                                         " evaluated to true on " +


smehringer:
> So, if I were to read `char` from a FASTA file (you could, right?) , the behaviour would change?
> 
> Because, before, we would ignore space and digits in the sequence, and now (since they are in the alphabet) we would write them?
> Could this case happen anyway? (Even if... ignoring is probably more wrong than just writing them).

Phew, good catch!
Actually, if the alphabet is `char` we do not want to skip digits and spaces since these are legal characters in alphabet char aren't they? I am not sure...

rrahn:
I would in general argue, that if the legal alphabet is char everything is allowed, since it might be an extended sequence format. However, I would forbid to read in non-printable characters as they should not be part of the format.

smehringer:
On another thought, we ignore digits and spaces to support the following (I think? I cannot find specs for it):
```
>seq1
10 AGATCGAGCT
```
 reading it as char or dna4 should yield the same sequence no?

h-2:
> reading it as char or dna4 should yield the same sequence no?

Yes, absolutely. It's up to the user whether they want their data in dna4, char or their own alphabet, but it's part of the format logic that numbers and spaces are skipped. 

OTOH we should add optimisations for if the alphabet is char because we can omit going through the converting view that throws. I have these in a branch somewhere and will open a PR when the other IO-stuff is sorted out.

Altogether 2-3% is also not much, I suspect there are lower hanging fruit with a higher impact in SeqAn3 right now.

rrahn:
> Yes, absolutely. It's up to the user whether they want their data in dna4, char or their own alphabet, but it's part of the format logic that numbers and spaces are skipped.
 
I am curious though. Can you point me/us to any tool and/or description that allows this? As far as I understand NCBI and EMBL tools (online services) won't allow for this kind of formatted fasta input. 

And yes if we bound this to the format then also a legal alphabet of type char should filter it. 

> OTOH we should add optimisations for if the alphabet is char because we can omit going through the converting view that throws. I have these in a branch somewhere and will open a PR when the other IO-stuff is sorted out.

If optimisation is the goal, then saying use char instead of dna alphabets is kind of counter-intutive and the question is why we don't do this as a default. Question is if we should rather have a pure fasta format skipping these tests and a more flexible fasta format for users that really need these. Maybe there is a domain where this is used heavily?!

> Altogether 2-3% is also not much, I suspect there are lower hanging fruit with a higher impact in SeqAn3 right now.

:+1: :+1: :+1: :+1:

h-2:
> I am curious though. Can you point me/us to any tool and/or description that allows this? As far as I understand NCBI and EMBL tools (online services) won't allow for this kind of formatted fasta input.

I had an issue about this in SeqAn2 or Lambda2, I don't remember, but people used these kind of files. A quick search returns a few hits that document this as valid, e.g. this beginner's book:
https://books.google.de/books?id=Guj1AgAAQBAJ&pg=PA79&lpg=PA79&dq=fasta+with+spaces
(top left)

> If optimisation is the goal, then saying use char instead of dna alphabets is kind of counter-intutive and the question is why we don't do this as a default.

Because we provide high-level data structures and type safety by default. For most use cases this is desired, it's the reason why we use alphabets at all and not plain char. But if you e.g. write a tool that filters sequences shorter than X in fasta-files and you do not care about amino acids or nucleotides, you might as well use `char` as alphabet types. It still means numbers and spaces shouldn't be part of it, though.

> Question is if we should rather have a pure fasta format skipping these tests and a more flexible fasta format for users that really need these. 

Yes, that is something I discussed with @smehringer in the beginning as well, also for versions that are slower, but enforce conventions more strictly (opposite use-case). The problem with FastA is just that it's a mess and it would be difficult to decide which of the 1000 flavours gets its own format and which doesn't. If you do something like that I would propose making single-line fasta (where there are also no linebreaks) the "fast-fasta" format, but altogether I think there are many things right now that are much more important -- especially considering the low performance gains.

rrahn:
> I had an issue about this in SeqAn2 or Lambda2, I don't remember, but people used these kind of files. A quick search returns a few hits that document this as valid, e.g. this beginner's book:
https://books.google.de/books?id=Guj1AgAAQBAJ&pg=PA79&lpg=PA79&dq=fasta+with+spaces
(top left)

Well, it is more the numbers that I find odd though. But the point is, it must have been generated somehow. So what are the tools that generate this kind of format? Do they want to write this kind of format with SeqAn? But I am not a data expert so if no one really knows for sure, we need to find the people that actually know about it and ask them. 

> Because we provide high-level data structures and type safety by default. For most use cases this is desired, it's the reason why we use alphabets at all and not plain char. But if you e.g. write a tool that filters sequences shorter than X in fasta-files and you do not care about amino acids or nucleotides, you might as well use char as alphabet types. It still means numbers and spaces shouldn't be part of it, though.

That is the point. If performance really matters, I guess that leaving the conversion out does not get a huge benefit. So we need other technologies like SIMD parsing that acutally make a difference. Even for simple conversions or filterings I would argue no to lift the validation as it is always important and something a lot to get wrong with. 

> Yes, that is something I discussed with @smehringer in the beginning as well, also for versions that are slower, but enforce conventions more strictly (opposite use-case). The problem with FastA is just that it's a mess and it would be difficult to decide which of the 1000 flavours gets its own format and which doesn't. If you do something like that I would propose making single-line fasta (where there are also no linebreaks) the "fast-fasta" format, 

Well that is really harsh. I mean there is a clear documentation that almost everyone is using when referring to fasta which is the wikipedia article. So assuming that this is the `true` Fasta format which seems to be required by all important services then having two formats (The true one and the extended one (including spaces and numbers)) we would cover 99% of all cases. For the rest it should be simple and nicely documented to add your own format. 

> but altogether I think there are many things right now that are much more important -- especially considering the low performance gains.

That for sure. Something I am going to take care of right now to get a proper work plan for the next weeks/months. But I was simply struggling with where the information about these format specifications come from. I had hoped you point me to some famous tools or blogs that describe/use these formats.

h-2:
> I had hoped you point me to some famous tools or blogs that describe/use these formats.

I wish I had those. :cry: 

>  mean there is a clear documentation that almost everyone is using when referring to fasta which is the wikipedia article. [...] we would cover 99% of all cases.

Yes, but that also states at least for whitespaces:

> Anything other than a valid character would be ignored (including spaces, tabulators, asterisks, etc...).

smehringer:
Ok so I'll close this PR and we leave it as is for now. The performance gain was really low anyway, so it is nothing we need to discuss at length for now.

PR #1631, opened 2020-03-06
[MISC] Replace template parameter csa-type

### LINKED ISSUE #1630: Replace template parameter csa-type from `fm_index_cursor::backward_search` with class template parameter
Why does `fm_index_cursor::backward_search` have a template parameter? Isn't the csa-type already encoded in the class's template parameter?

----
### TODO
- [x] Add private alias `using sdsl_index_type = typename index_t::sdsl_index_type;`
- [x] Remove template parameter and replace `csa_t` in `backward_search` with `sdsl_index_type` (2x, signature+if constexpr)
* Note, `bidirectional_search` in `bi_fm_index_cursor` needs to  have a template parameter, because the function is called both on fwd and rev index.

Irallia:
Resolves #1630

### DIFF
--- a/include/seqan3/search/fm_index/fm_index_cursor.hpp
+++ b/include/seqan3/search/fm_index/fm_index_cursor.hpp
@@ -78,6 +78,8 @@ private:
     using node_type = detail::fm_index_cursor_node<index_t>;
     //!\brief Type of the representation of characters in the underlying SDSL index.
     using sdsl_char_type = typename index_type::sdsl_char_type;
+    //!\brief Type of the SDSL index.
+    using sdsl_index_type = typename index_t::sdsl_index_type;
     //!\brief Type of the alphabet size in the underlying SDSL index.
     using sdsl_sigma_type = typename index_type::sdsl_sigma_type;
     //!\brief Alphabet type of the index.
@@ -106,15 +108,17 @@ private:
     }
 
     //!\brief Optimized backward search without alphabet mapping
-    template <detail::sdsl_index csa_t>
-    bool backward_search(csa_t const & csa, sdsl_char_type const c, size_type & l, size_type & r) const noexcept
+    bool backward_search(sdsl_index_type const & csa,
+                         sdsl_char_type const c,
+                         size_type & l,
+                         size_type & r) const noexcept
     {
         assert(l <= r && r < csa.size());
 
         size_type _l, _r;
 
         size_type cc = c;
-        if constexpr(!std::same_as<typename csa_t::alphabet_type, sdsl::plain_byte_alphabet>)
+        if constexpr(!std::same_as<index_alphabet_type, sdsl::plain_byte_alphabet>)
         {
             cc = csa.char2comp[c];
             if (cc == 0 && c > 0) // [[unlikely]]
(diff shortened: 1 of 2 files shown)


include/seqan3/search/fm_index/fm_index_cursor.hpp
```diff
@@ -106,15 +108,18 @@ class fm_index_cursor
...
+                         size_type & l,
+                         size_type & r) const noexcept
     {
         assert(l <= r && r < csa.size());
 
         size_type _l, _r;
 
         size_type cc = c;
-        if constexpr(!std::same_as<typename csa_t::alphabet_type, sdsl::plain_byte_alphabet>)
+        if constexpr(!std::same_as<typename sdsl_index_type::alphabet_type, sdsl::plain_byte_alphabet>)
```
smehringer:
```suggestion
        if constexpr(!std::same_as<index_alphabet_type, sdsl::plain_byte_alphabet>)
```
we already store this too

include/seqan3/search/fm_index/fm_index_cursor.hpp
```diff
@@ -106,15 +108,18 @@ class fm_index_cursor
     }
 
     //!\brief Optimized backward search without alphabet mapping
-    template <detail::sdsl_index csa_t>
-    bool backward_search(csa_t const & csa, sdsl_char_type const c, size_type & l, size_type & r) const noexcept
+    template <detail::sdsl_index sdsl_index_type>
```
smehringer: when you use the class member type within the function, you do not need the template at all :) you can remove this line

Irallia: Yes, that did the trick. Thx!

Irallia:
> Looks good!
> Can you do the same in the bi_fm_index_cursor (in a separate commit)
> It has the function (line 146)
> 
> ```c++
> template <detail::sdsl_index csa_t>
>     bool bidirectional_search(csa_t const & csa, sdsl_char_type const c,
>                               size_type & l_fwd, size_type & r_fwd,
>                               size_type & l_bwd, size_type & r_bwd) const noexcept
> ```
> 
> And it lools like this needs the exact same restructering :)
> And here (line 200)
> 
> ```c++
>     template <detail::sdsl_index csa_t>
>     bool bidirectional_search_cycle(csa_t const & csa, sdsl_char_type const c,
>                                     size_type const l_parent, size_type const r_parent,
>                                     size_type & l_fwd, size_type & r_fwd,
>                                     size_type & l_bwd, size_type & r_bwd) const noexcept
> ```

https://github.com/seqan/seqan3/issues/1630: Note, bidirectional_search in bi_fm_index_cursor needs to have a template parameter, because the function is called both on fwd and rev index.

smehringer:
> #1630: Note, bidirectional_search in bi_fm_index_cursor needs to have a template parameter, because the function is called both on fwd and rev index.

Oh sorry, I did not read the issue carefully. 
But the two FM indices of the Bi-FM index have the same sdsl_index_type, so the underlying sdsl data structures used in backward_search should also be the same, or not? Did you try?

PR #1632, opened 2020-03-10
Remove bi/fm_index concepts

smehringer:
As discussed in the strategy meeting I tried to remove the concepts.
Here is a WIP branch but I have the following issue:

Since the bi/fm_index classes have type AND value template arguments, I can neither use `is_type_specialisation_of` nor `is_value_specialisation_of` to check whether I am working with fm_index or bi_fm_index.. or am I doing something wrong?

The only solution I see would be the very lengthy one I show in the `search_algo` function or to keep the concepts but move them to detail. 

@seqan/core what do you think?

### DIFF
--- a/include/seqan3/search/algorithm/detail/search.hpp
+++ b/include/seqan3/search/algorithm/detail/search.hpp
@@ -26,6 +26,19 @@ namespace seqan3::detail
  * \{
  */
 
+/*!\brief Searches a query sequence in an index.
+ *
+ * \copydetails search_algo_bi
+ */
+template <bool abort_on_hit, typename index_t, typename query_t, typename delegate_t>
+inline void search_algo(index_t const & index, query_t & query, search_param const error_left, delegate_t && delegate)
+{
+    if constexpr (bi_fm_index_specialisation<index_t>)
+        search_algo_bi<abort_on_hit>(index, query, error_left, delegate);
+    else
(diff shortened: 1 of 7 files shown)


smehringer:
Closing this to be re-evaluated later because
1. The concept cannot easily be replaced by a specialisation check
2. This needs a discussion about what the search interface should accept (currently in accepts `fm_index_specialisations` ) once we plan the `seqan3::dream_index`

h-2:
Why don't you just accepts concrete type templates? It's not meant to be customisable anyway.

smehringer:
> Why don't you just accepts concrete type templates? It's not meant to be customisable anyway.

I'm not sure what you mean.. If you mean that there should be two function overloads
```cpp
template <typename ... >
search_algo(fm_index<...>)


template <typename ... >
search_algo(bi_fm_index<...>)
```

then this would involve more refactoring / renaming since this is quite weird in the search code right now.
* **unidirectional search** :`search_all(index_t)` -> `search_single(index_t)` -> `search_algo(index_t)` -> 
`search_algo_uni(index_t)` -> `search_trivial(index_t)`

* **bidirectional search** :`search_all(index_t)` -> `search_single(index_t)` -> `search_algo(index_t)` -> 
`search_algo_bi(index_t)`

PR #1651, opened 2020-03-16
[FIX] Fixes the constructor of the combined range closur adaptor.

rrahn:
Fixes the constructor of the combined range adaptor closure object. Before it could not be called in a constexpr context, even though it should. Also the forward declarations where useless  because the tempalte parameters came from the class defintion and not the constructor so if the type was not explicitly set to an reference type (which I don't see how it should happen) it was always a copy of the closure objects. Which is also fine in this context I believe.

### DIFF
--- a/include/seqan3/range/views/detail.hpp
+++ b/include/seqan3/range/views/detail.hpp
@@ -252,12 +252,12 @@ public:
 
     //!\brief Inherit the base type's constructors.
     using base_type::base_type;
-    //!\}
 
     //!\brief Store both arguments in the adaptor.
-    combined_adaptor(left_adaptor_t l, right_adaptor_t r) :
+    constexpr combined_adaptor(left_adaptor_t l, right_adaptor_t r) :
         base_type{std::forward<left_adaptor_t>(l), std::forward<right_adaptor_t>(r)}
     {}
+    //!\}
 };
 
 // ============================================================================
--- a/test/unit/range/views/adaptor_base_test.cpp
+++ b/test/unit/range/views/adaptor_base_test.cpp
@@ -10,6 +10,7 @@
 
 #include <gtest/gtest.h>
 
+#include <seqan3/core/type_traits/function.hpp>
 #include <seqan3/range/views/detail.hpp>
 
 using namespace seqan3;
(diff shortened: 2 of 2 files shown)


include/seqan3/range/views/detail.hpp
```diff
@@ -252,12 +252,11 @@ class combined_adaptor :
 
     //!\brief Inherit the base type's constructors.
     using base_type::base_type;
-    //!\}
 
     //!\brief Store both arguments in the adaptor.
-    combined_adaptor(left_adaptor_t l, right_adaptor_t r) :
-        base_type{std::forward<left_adaptor_t>(l), std::forward<right_adaptor_t>(r)}
+    constexpr combined_adaptor(left_adaptor_t l, right_adaptor_t r) : base_type{std::move(l), std::move(r)}
```
marehr: I don't fully understand this.

Before:
* If `left/right_adaptor_t` is lvalue this would be passed as is.
* If `left/right_adaptor_t` is rvalue this would move out the content into a new temporary.

Now:
* If `left/right_adaptor_t` is lvalue this would move out the content into a new temporary.
* If `left/right_adaptor_t` is rvalue this would move out the content into a new temporary.

I don't understand why this avoids copying?

```suggestion
    constexpr combined_adaptor(left_adaptor_t & l, right_adaptor_t & r) : base_type{std::move(l), std::move(r)}
```
?

rrahn: `left_adaptor_t` and `right_adaptor_t` are template parameters of the class and not of the constructor! Accordingly, these were never forwarding references. So only if you would have defined a type `combined_adaptor<type &, type &>` it would have been forwarded as l-value reference. In all other cases this constructor always took the arguments by value rendering the forward declaration useless and also potentially dangerous. So l-values will be copied and r-values will be moved. While before all elements were copied unless the class templates were references. 

If you add `&` to the arguments than r-values can never be used in this expression.

h-2: I have vague memories that the original behaviour was actually intended. Are you sure that the template parameters may not already include a `&` in some cases? If so, you should definitely not move here.

marehr: @h-2 Thank you for clarifying. I talked with rrahn two days ago and we decided to just add the `constexpr`, because we were not sure if that would have an impact and if the code was intended as is.

h-2: > After all the range closure objects are state-less objects (at least so far).

But you could easily create closure objects that have state, so I think it's a valid use-case...

[just procrastinating right now :innocent: ]

h-2: > @h-2 Thank you for clarifying. I talked with rrahn two days ago and we decided to just add the constexpr, because we were not sure if that would have an impact and if the code was intended as is.

:+1:

I can't gaurantee that it was, but I have a hunch. It's not often that you see `std::forward` on something that is not `T &&` but in especially when `T` is a template parameter from somewhere else, it can make a lot of sense!

rrahn: > But you could easily create closure objects that have state, so I think it's a valid use-case...

But none of ours do, don't they? Are there examples that do this? Just curious.
The thing, is without a deduction guide the ctad would never capture lvalue references with the current constructor. And nowhere in our code we actually define the templates explicitly such that I was wondering, what was intended here. But I agreed on @marehr's opinion, that this change would not do the same as before and since it is burried in detail it should not be important anyway. Unless there is some specification that closure objects must be semiregular? But I am not aware of such a restriction.

rrahn:
@marehr can you have another look

Issue #1610, opened 2020-02-25
Remove #pragma GCC diagnostic ignored "-Wdeprecated-declarations"


marehr:
https://github.com/seqan/seqan3/blob/2763c900318c548ae824ca352547618ee5e323d4/include/seqan3/io/structure_file/input.hpp#L25-L27

https://github.com/seqan/seqan3/blob/2763c900318c548ae824ca352547618ee5e323d4/include/seqan3/io/structure_file/output.hpp#L23-L25

https://github.com/seqan/seqan3/blob/2763c900318c548ae824ca352547618ee5e323d4/include/seqan3/io/sequence_file/input.hpp#L21-L23

https://github.com/seqan/seqan3/blob/2763c900318c548ae824ca352547618ee5e323d4/include/seqan3/io/sequence_file/output.hpp#L21-L23

https://github.com/seqan/seqan3/blob/e9a04e6a6d5e7d17bc9d6d3fc40352f098827d4c/test/unit/range/container/container_of_container_test.cpp#L248-L250

@smehringer Can you coordinate this? Not sure if this is a good first issue.

smehringer:
Did you check if no warnings are triggered when removing them?

marehr:
No, this is part of this issue.

smehringer:
@simonsasse Can you check if removing any of these lines still triggers a compiler warning? (Note that there is always a `#pragma pop` somewhere at the end that you have to remove to)

simonsasse:
I did not edit:
https://github.com/seqan/seqan3/blob/e9a04e6a6d5e7d17bc9d6d3fc40352f098827d4c/test/unit/range/container/container_of_container_test.cpp#L248-L250

This provoked issues compiling with GCC-7. The container_of_container_test failed.

PR #1653, opened 2020-03-17
[I/O,MISC] Remove #pragma GCC diagnostics

simonsasse:
These changes resolve issue #[1610](https://github.com/seqan/seqan3/issues/1610).
I removed all 
- #pragma GCC diagnostic push
- #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
- #pragma GCC diagnostic pop

In the file seqan3/test/unit/range/container/container_of_container_test.cpp, I removed the whole paragraph although I'm not 100% sure if that was right.
All tests ran properly.

### DIFF
--- a/include/seqan3/io/alignment_file/all.hpp
+++ b/include/seqan3/io/alignment_file/all.hpp
@@ -17,9 +17,6 @@
  * \brief Provides files and formats for handling alignment data.
  */
 
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
-
 #include <seqan3/io/alignment_file/format_bam.hpp>
 #include <seqan3/io/alignment_file/format_sam.hpp>
 #include <seqan3/io/alignment_file/header.hpp>
@@ -30,5 +27,3 @@
 #include <seqan3/io/alignment_file/output_format_concept.hpp>
 #include <seqan3/io/alignment_file/output_options.hpp>
 #include <seqan3/io/alignment_file/sam_tag_dictionary.hpp>
-
-#pragma GCC diagnostic pop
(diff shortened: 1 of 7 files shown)


smehringer:
> General question: Why don't we remove these pragmas in .../seqan3/include/seqan3/io/alignment_file/all.hpp and others, do they still avoid errors?

@simonsasse Can you check those too?

simonsasse:
So I removed all occurrences of these lines in the library except in `submodules and in `sam_tag_dictionary.hpp` where the removal caused an error.

marehr:
> So I removed all occurrences of these lines in the library except in `submodules` and in `sam_tag_dictionary.hpp` where the removal caused an error.

Can you post the error message? Ty

simonsasse:
@marehr I think I missunderstood something:
In `sam_tag_dictionary.hpp` the code looks like this:
`#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wpedantic"
//...
#pragma GCC diagnostic pop`
Am I supposed to delete these lines as well? I just noticed that it ignores another flag `-Wpedantic` and not `-Wdeprecated-declarations`

simonsasse:
If I remove those lines in `sam_tag_dictionary.hpp` this error shows while building the `unit/io` tests:
` ISO C++ did not adopt string literal operator templates taking an argument pack of characters [-Werror=pedantic]`
`   71 | constexpr uint16_t operator""_tag()`
`      |                    ^~~~~~~~`

simonsasse:
Since the `container_of_container_test` failed with travis and Jenkins I would undo the changes to this file: `test/unit/range/container/container_of_container_test.cpp`? the problems did not occur on my laptop with gcc-9.

smehringer:
@simonsasse For now, undo the changes wherever they cause build failures but document those in the issue (and check the ones you fixed). Then we can come back to them later

simonsasse:
I removed the changes in `container_of_container_test` because they caused errors while building with gcc-7. I noted this in the issue.

PR #1654, opened 2020-03-18
Feature/minimizer view

MitraDarja:
Implements the minimiser view without looking at the reverse strand (will be a seperate PR).

blocked by #1756


include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,550 @@
...
+    minimiser & operator=(minimiser && rhs)      = default; //!< Defaulted.
+    ~minimiser()                                 = default; //!< Defaulted.
+
+    //!\brief Construct from a view and a given number of elements in one window (w_elems).
+    minimiser(urng_t urange_, uint32_t const & w_) : urange{std::move(urange_)}, w_elems{w_}{}
+
+    //!\brief Construct from a non-view that can be view-wrapped and a given number of elements in one window (w_elems).
+    template <typename rng_t>
+    //!\cond
+     requires !std::same_as<seqan3::remove_cvref_t<rng_t>, minimiser> &&
```
joergi-w, line 69: Do we exclude minimiser for technical reasons or because there is no application yet?
Technically a minimiser returns a range of numbers and I could use a minimiser on this range again...

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,550 @@
...
+     * \{
+     */
+    constexpr window_iterator()                                    = default; //!< Defaulted.
+    constexpr window_iterator(window_iterator const &)             = default; //!< Defaulted.
+    constexpr window_iterator(window_iterator &&)                  = default; //!< Defaulted.
+    constexpr window_iterator & operator=(window_iterator const &) = default; //!< Defaulted.
+    constexpr window_iterator & operator=(window_iterator &&)      = default; //!< Defaulted.
+    ~window_iterator()                                             = default; //!< Defaulted.
+
+    /*!\brief Construct from a given iterator of std::size_t, a second iterator of std::size_t and w_elems.
```
joergi-w, line 196: Again, the description says w_elems but the parameter is named `w`. Please keep consistent naming.

joergi-w:
```suggestion
    /*!\brief Construct from begin and end iterators of a given range over std::size_t elements, and the number elements per window.
```

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,550 @@
...
+    it_t window_left;
+
+    //!\brief Iterator to the rightmost element of one window.
+    it_t window_right;
+
+    //!\brief The number of elements in one window.
+    uint32_t w_elems;
+
+    //!\brief Stored values per window.
+    std::deque<uint64_t> windowValues;
```
joergi-w, line 367: Please use snake case according to our code style rules: `window_values`.

test/unit/range/views/view_minimiser_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+    EXPECT_FALSE((std::ranges::output_range<decltype(v2), size_t>));
+
+    auto v3 = flist_text2 | kmer_view | minimiser_view;
+    EXPECT_TRUE(std::ranges::input_range<decltype(v3)>);
+    EXPECT_TRUE(std::ranges::forward_range<decltype(v3)>);
+    EXPECT_TRUE(std::ranges::view<decltype(v3)>);
+    EXPECT_FALSE(std::ranges::common_range<decltype(v3)>);
+    EXPECT_TRUE(seqan3::const_iterable_range<decltype(v3)>);
+    EXPECT_FALSE((std::ranges::output_range<decltype(v3), size_t>));
+}
```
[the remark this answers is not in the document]
MitraDarja: Maybe I misunderstand views, but isn't the type of a view of a std::vector<seqan3::dna4> different to a view of a seqan3::bitcompressed_vector<seqan3::dna4>? So, the reason why I didn't do it with a for-loop is that I would need a container, which holds these different types to loop over them. Plus, that is how we implemented it in the kmer_hash tests.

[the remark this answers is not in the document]
joergi-w: You are right that they are different types and you cannot use a normal for-loop. But there are typed tests which you can call on a seqan3::type_list of different types for instance, which makes sense if you also want to add the const types or if there are further types in the future.

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,550 @@
...
+class minimiser : public std::ranges::view_interface<minimiser<urng_t>>
+{
+private:
+    static_assert(std::ranges::forward_range<urng_t const>, "The minimiser only works on forward_ranges");
+
+    //!\brief The underlying range.
+    urng_t urange;
+
+    //!\brief The number of elements in one window.
+    uint32_t w_elems;
```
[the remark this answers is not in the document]
MitraDarja: I think the terms window span/size/length would be misleading for someone who reads about minimizer in papers, because there these terms refer to the window on the text not on the hash values, so I think another term makes it less confusing, but I will change it to num_w_elems. Thanks! :)

test/unit/range/views/view_minimiser_test.cpp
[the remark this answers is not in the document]
MitraDarja: Thanks for the explanation! :)

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,550 @@
...
+#include <seqan3/range/hash.hpp>
+
+namespace seqan3::detail
+{
+// ---------------------------------------------------------------------------------------------------------------------
+// minimiser class
+// ---------------------------------------------------------------------------------------------------------------------
+
+/*!\brief The type returned by seqan3::views::minimiser.
+ * \tparam urng_t The type of the underlying ranges, must model std::forward_range, the reference type must model
```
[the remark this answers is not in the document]
MitraDarja: I added a test for integral types, I think that all integral types should work... :thinking:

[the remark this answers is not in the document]
joergi-w: Do minimisers restrict on ranges of type size_t [1] or can they be applied for any (unsigned?) integer type like unsigned int, short, int etc. [2]?

If [1]: You can use `std::is_same_v<T, size_t>` to test whether `T` has `size_t` type. 
If [2]: Keep the assertion for std::Integral as implemented (and extend by `!is_signed` if needed). Change all your descriptions where you say something about `size_t` to (unsigned) integral type. Add one test case that creates one range per integral type and compute a minimiser to verify that it works.

include/seqan3/range/views/minimiser.hpp
joergi-w: Make it simple (details are in the param description already): 
Construct from two iterators and the number of elements per window.

include/seqan3/range/views/minimiser.hpp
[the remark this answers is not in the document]
MitraDarja: I think, I have never seen a usage with negative values, but in theory there is no reason why minimizer should not work on these values as well, so I will change it to integral. Thanks for the explanation! :smile:

include/seqan3/range/views/minimiser.hpp
```diff
@@ -24,7 +24,7 @@ namespace seqan3::detail
 
 /*!\brief The type returned by seqan3::views::minimiser.
  * \tparam urng_t The type of the underlying ranges, must model std::forward_range, the reference type must be
- *                std::size_t. The typical use case is that the reference type is the result of seqan3::kmer_hash.
+ *                std::integral. The typical use case is that the reference type is the result of seqan3::kmer_hash.
```
joergi-w, line 27: the reference type must model std::integral.

include/seqan3/range/views/minimiser.hpp
```diff
@@ -39,8 +39,8 @@ class minimiser : public std::ranges::view_interface<minimiser<urng_t>>
 {
 private:
     static_assert(std::ranges::forward_range<urng_t const>, "The minimiser only works on forward_ranges.");
-    static_assert(std::is_integral<reference_t<urng_t>>::value, "The reference type of the underlying range must model "
-                                                     "std::size_t.");
+    static_assert(std::is_integral<reference_t<urng_t>>::value, "The reference type of the underlying range must be of "
+                                                                "type std::integral.");
```
joergi-w, line 43: The reference type of the underlying range must model std::integral.

include/seqan3/range/views/minimiser.hpp
```diff
@@ -503,13 +503,13 @@ namespace seqan3::views
  * \{
  */
 
-/*!\brief                   Computes minimisers for a range of type std::size_t. A minimiser is the smallest value in a
+/*!\brief                   Computes minimisers for a range of type std::integral. A minimiser is the smallest value in a
```
joergi-w, line 506: ... for a range of integral numbers. (std::integral is not a type but a concept of a group of types with certain properties)

joergi-w:
You have doxygen warnings for doc_dev_test:
```
range/views/minimiser.hpp:51: warning: Compound seqan3::detail::minimiser::window_iterator is not documented.
range/views/minimiser.hpp:159: warning: Compound seqan3::detail::minimiser< urng_t >::window_iterator is not documented.
```

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,553 @@
...
+// ---------------------------------------------------------------------------------------------------------------------
+
+//![adaptor_def]
+//!\brief views::minimiser's range adaptor object type (non-closure).
+struct minimiser_fn
+{
+    //!\brief Store the number of elements in one window and return a range adaptor closure object.
+    constexpr auto operator()(uint32_t const & num_w_elems) const
+    {
+        return seqan3::detail::adaptor_from_functor{*this, num_w_elems};
```
joergi-w, line 474: `seqan3::detail::adaptor_from_functor`: this needs an include (see header test)

include/seqan3/range/views/minimiser.hpp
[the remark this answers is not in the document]
joergi-w: Good! 👍  
And please remember to also add the suggested test case for other types. Currently only size_t is tested.

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,555 @@
...
+    //!\name Comparison operators
+    //!\{
+
+    //!\brief Compare to iterator on underlying range.
+    friend bool operator==(window_iterator const & lhs, sentinel_t const & rhs) noexcept
+    {
+        return lhs.window_right == rhs;
+    }
+
+    //!\brief Compare to iterator on underlying range.
```
smehringer, line 228: You are not comparing to an iterator of the underlying range here?

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,555 @@
...
+     * \param skip Amount to increment.
+     * \attention This function is only avaible if `it_t` models std::random_access_iterator.
+     */
+    void get_minimiser(difference_type const skip)
+    //!\cond
+        requires std::random_access_iterator<it_t>
+    //!\endcond
+    {
+        std::ranges::advance(window_left, skip);
+        window_first();
```
smehringer, line 406: I don't think you can model a random-access range. Sorry if I was not clear about this but I thought we agreed on a forward range being fine. 
The reason is that you now jump to the `skip`-th position in the underlying view but not the `skip`-th minimizer.
And you cannot know how many minimizers there are beforehand.

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,555 @@
...
+    //!\endcond
+    {
+        window_left += n;
+        get_minimiser();
+        return operator*();
+    }
+
+private:
+    //!brief Iterator to last element in range.
+    sentinel_t last_elem;
```
smehringer, line 354: Since the name `last_elem` could also refer to the last elem in the window
```suggestion
    sentinel_t urange_end;
```

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,555 @@
...
+    //!\brief Iterator to the leftmost element of one window.
+    it_t window_left;
+
+    //!\brief Iterator to the rightmost element of one window.
+    it_t window_right;
+
+    //!\brief The number of elements in one window.
+    uint32_t num_w_elems;
+
+    //!\brief Stored values per window.
```
smehringer, line 371: Add a small explanation of why we need to store the values

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,555 @@
...
+    std::deque<uint64_t> window_values;
+
+    //!\brief Increments iterator by 1.
+    void get_minimiser()
+    {
+        if (window_values.size() == 0)
+            window_first();
```
smehringer, line 378: if the `else` block has braces, the if block gets braces too :nail_care:

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,555 @@
...
+        for (uint32_t i = 0; (i < num_w_elems - 1) ; i++)
+        {
+            window_values.push_back(*window_right);
+            std::ranges::advance(window_right,  1);
+        }
+        window_values.push_back(*window_right);
+        minimiser_value = *(std::min_element(std::begin(window_values), std::end(window_values)));
+    }
+
+    //!\brief Calculates the next minimiser value via rolling hash.
```
smehringer, line 423: rolling hash? copy and paste error?

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,555 @@
...
+        }
+        else
+        {
+            window_values.pop_front();
+        }
+
+        uint64_t newElem = *window_right;
+        window_values.push_back(newElem);
+
+        if ((window_values.back() < minimiser_value))
```
smehringer, line 448:
```suggestion
        if (new_element < minimiser_value)
```

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,555 @@
...
+                minimiser_changed = true;
+            }
+
+        }
+        else
+        {
+            window_values.pop_front();
+        }
+
+        uint64_t newElem = *window_right;
```
smehringer, line 445:
```suggestion
        uint64_t new_element = *window_right;
```

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,555 @@
...
+        window_values.push_back(newElem);
+
+        if ((window_values.back() < minimiser_value))
+        {
+            minimiser_value = *(std::end(window_values) - 1);
```
smehringer, line 450:
```suggestion
            minimiser_value = new_element;
```

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,555 @@
...
+
+    //!\brief Increments iterator by 1.
+    void get_minimiser()
+    {
+        if (window_values.size() == 0)
+            window_first();
+        else
+        {
+            // Call next_minimiser until minimiser value changed or end of the underlying range is reached.
+            while((!minimiser_changed) && (std::ranges::next(window_right) != last_elem))
```
smehringer, line 382: you do not need to store the variable `minimiser_changed` (and this makes the iterator more light weigth) if you make `next_minimizer` return a boolean. The boolean indicates whether a new minimizer has been found or the end has been reached (`true`)
```suggestion
            while(!next_minimiser())
```
Then inside `next_minimiser`
add after advancing window_left and window_right
```cpp
if (window_right == las_elem)
    return true;
```
This also saves the call to `std::ranges::next`

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,555 @@
...
+            {
+                next_minimiser();
+            }
+            if (!minimiser_changed)
```
smehringer, line 386: I think with the changes above mentioned this can also go because you are advancing wondow_left and window_right within next_minimizer before you return true

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,555 @@
...
+ *
+ *
+ * ### View properties
+ *
+ * | Concepts and traits              | `urng_t` (underlying range type)   | `rrng_t` (returned range type)   |
+ * |----------------------------------|:----------------------------------:|:--------------------------------:|
+ * | std::ranges::input_range         | *required*                         | *preserved*                      |
+ * | std::ranges::forward_range       | *required*                         | *preserved*                      |
+ * | std::ranges::bidirectional_range |                                    | *lost*                           |
+ * | std::ranges::random_access_range |                                    | *lost*                           |
```
smehringer, line 531: here you even document that random access range is lost but above you tried to implement it?

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,555 @@
...
+    //!\brief Construct from a view and a given number of elements in one window.
+    minimiser(urng_t urange_, uint32_t const & w_) : urange{std::move(urange_)}, num_w_elems{w_}{}
+
+    //!\brief Construct from a non-view that can be view-wrapped and a given number of elements in one window.
+    template <typename rng_t>
+    //!\cond
+     requires std::ranges::viewable_range<rng_t> &&
+              std::constructible_from<urng_t, ranges::ref_view<std::remove_reference_t<rng_t>>>
+    //!\endcond
+    minimiser(rng_t && urange_, uint32_t const & k_, uint32_t const & w_) :
```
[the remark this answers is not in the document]
eseiler: Also you should capture by const value instead of const ref, no?

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,555 @@
...
+{
+private:
+    static_assert(std::ranges::forward_range<urng_t const>, "The minimiser only works on forward_ranges.");
+    static_assert(std::is_integral<reference_t<urng_t>>::value, "The reference type of the underlying range must model "
+                                                                "std::integral.");
+    //!\brief The underlying range.
+    urng_t urange;
+
+    //!\brief The number of elements in one window.
+    uint32_t num_w_elems;
```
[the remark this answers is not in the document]
MitraDarja: It is not the window_size anymore bur the number of elements in a window, so I think naming it window_size would be confusing.

include/seqan3/range/views/minimiser.hpp
MitraDarja: Is that not what the sentinel is? In kmer_hash this is commented as "Compare to iterator on text."

include/seqan3/range/views/minimiser.hpp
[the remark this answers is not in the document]
smehringer: I thought the two are synonyms.
Doesn't a window size of 10 mean 10 elements in the window?

include/seqan3/range/views/minimiser.hpp
smehringer: Ah I see the confusion. The end of the minimizer range is also the end of the underlying range, thats why the sentinel type is the same. This comment suggests that I can compare to any iterator of the underlying range, but I can only compare to the sentinel.
How about `Compare to the sentinel of the underlying range.`

include/seqan3/range/views/minimiser.hpp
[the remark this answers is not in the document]
MitraDarja: No, a window_size of 10 means w - k + 1 elements that are completely in the window, we can also say we want to consider all k-mers starting in a window (which is done in some papers), but in all our example figures we do the former... But this view should be capable of handling both, it depends on what an user defines for num_w_elems, so I still would say not naming it window_size is a good idea.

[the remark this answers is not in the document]
smehringer: ah ok. thanks for the explanation. I would still try to avoid abbreviations as it makes it easier to read the code without jumping to the documentation of the variable.
What about `elements_in_window` or `window_values_size` (if I remember the name of your std::deque variable that stores the window values/elements). You also switch between the name "elements" or "values", it would be easier to understand if you'd only use one of them for consistency?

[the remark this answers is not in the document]
MitraDarja: Ok, I changed the variable name to `window_values_size` and removed all mentions of the name element in the context of the minimiser.

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,481 @@
...
+     */
+    minimiser_view()                                       = default; //!< Defaulted.
+    minimiser_view(minimiser_view const & rhs)             = default; //!< Defaulted.
+    minimiser_view(minimiser_view && rhs)                  = default; //!< Defaulted.
+    minimiser_view & operator=(minimiser_view const & rhs) = default; //!< Defaulted.
+    minimiser_view & operator=(minimiser_view && rhs)      = default; //!< Defaulted.
+    ~minimiser_view()                                      = default; //!< Defaulted.
+
+    //!\brief Construct from a view and a given number of values in one window.
+    minimiser_view(urng_t urange_, uint32_t const & w_) : urange{std::move(urange_)}, window_values_size{w_}{}
```
smehringer, line 67:
```suggestion
    minimiser_view(urng_t urange_, uint32_t const w) : urange{std::move(urange_)}, window_values_size{w_}{}
```
And the documentation `\param` is missing
Same for the contructor below

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,481 @@
...
+// ---------------------------------------------------------------------------------------------------------------------
+// minimiser_view class
+// ---------------------------------------------------------------------------------------------------------------------
+
+/*!\brief The type returned by seqan3::views::minimiser.
+ * \tparam urng_t The type of the underlying ranges, must model std::forward_range, the reference type must model
+ *                std::totally_ordered. The typical use case is that the reference type is the result of
+ *                seqan3::kmer_hash.
+ * \implements std::ranges::view
+ * \implements std::ranges::random_access_range
```
smehringer, line 31: remove this as the view does not implement a random access range. Same below with sized range.

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,481 @@
...
+    //!\brief The pointer type.
+    using pointer = void;
+    //!\brief Reference to `value_type`.
+    using reference = value_type;
+    //!\brief Tag this class as forward iterator.
+    using iterator_category = std::forward_iterator_tag;
+    //!\brief Tag this class depending on which concept `it_t` models.
+    using iterator_concept = std::conditional_t<std::contiguous_iterator<it_t>,
+                                                typename std::forward_iterator_tag,
+                                                seqan3::iterator_tag_t<it_t>>;
```
smehringer, line 186: Right now it says "If the range is a continues_ranges this view models a forward_range, otherwise it models whatever the input range models". Which is wrong, since we can only model at most a forward range. So I think this one is correct:

```suggestion
    using iterator_concept = std::conditional_t<std::input_iterator<it_t>,
                                                std::input_iterator_tag,
                                                std::forward_iterator_tag>;
```
I also wonder why this differs from the iterator_category. I know you copied it from the kmer_hash. @eseiler do you remember the rational behind it when you wrote the kmer_hash?
Otherwise, I would make them the same.

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,481 @@
...
+    //!\brief The number of values in one window.
+    uint32_t window_values_size;
+
+    //!\brief Stored values per window. It is necessary to store them, because a shift can remove the current minimiser.
+    std::deque<uint64_t> window_values;
+
+    //!\brief Increments iterator by 1.
+    void get_minimiser()
+    {
+        if (window_values.size() == 0)
```
smehringer, line 326: If clauses are very expensive in "hot-loops". I'm not sure, but this if-statement is only true at the construction of the window_iterator, right? Then Maybe this if clause can be removed then if you call `window_first` directly in the constructor. 
EDIT:
And on second thought, can you check if we need to store `window_value_size` at all then? When I grep for it in the iterator it is only used in the constructor, `window_first` and the comparison operators. 
In the comparator operators you can use `window_values.size()` directly

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,481 @@
...
+            window_first();
+        else // Call next_minimiser until minimiser value changed or end of the underlying range is reached.
+            while(!next_minimiser()){}
```
smehringer, line 329:
```suggestion
            while (!next_minimiser()) {}
```
:nail_care:

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,481 @@
...
+        if (minimiser_value == *(std::begin(window_values)))
+        {
+            window_values.pop_front();
+            if (!window_values.empty())
+            {
+                window_values.push_back(new_value);
+                minimiser_value = *(std::min_element(std::begin(window_values), std::end(window_values)));
+                return true;
+            }
+
```
smehringer, line 363: Remove empty newline :nail_care:

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,481 @@
...
+        }
+
+        window_values.push_back(new_value);
+
+        if (new_value < minimiser_value)
+        {
+            minimiser_value = new_value;
+            return true;
+        }
+        else if (window_values.size() == 1)
```
smehringer, line 377: Hm, do we allow a `window_value_size` of 1? It doesn't make much sense as it just returns the original range and it adds an extra if-clause just for that.

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,481 @@
...
+                return true;
+            }
+
+        }
+        else
+        {
+            window_values.pop_front();
+        }
+
+        window_values.push_back(new_value);
```
smehringer, line 370: Since you `return` in the if-clause to jump out of the function and otherwise (else) continue will all of the code below, the else statement is unnecessary.
```suggestion

        window_values.pop_front();
        window_values.push_back(new_value);
```

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,481 @@
...
+    }
+};
+//![adaptor_def]
+
+} // namespace seqan3::detail
+
+namespace seqan3::views
+{
+
+/*!\name Alphabet related views
```
smehringer, line 430: Is this an alphabet related view?

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,481 @@
...
+ * \{
+ */
+
+/*!\brief                           Computes minimisers for a range of integral numbers. A minimiser is the smallest
```
smehringer, line 434:
```suggestion
/*!\brief                           Computes minimisers for a range of comparable values. A minimiser is the smallest
```

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,481 @@
...
+ * \param[in] urange                The range being processed. [parameter is omitted in pipe notation]
+ * \param[in] window_values_size    The number of values in one window.
+ * \returns                         A range of std::totally_ordered where each value is the minimal value for one window.
+ *                                  See below for the properties of the returned range.
+ * \ingroup views
+ *
+ * \details
+ *
+ * A minimiser is the smallest value in a window. For example for the following list of hash values
+ * [28, 100, 9, 23, 4, 1, 72, 37, 8] and 4 as window_values_size, the minimiser values are [9,4,1].
```
smehringer, line 447:
```suggestion
 * [28, 100, 9, 23, 4, 1, 72, 37, 8] and 4 as `window_values_size`, the minimiser values are [9,4,1].
```

include/seqan3/range/views/minimiser.hpp
eseiler: kmer_hash view:
```cpp
    using iterator_concept = std::conditional_t<std::contiguous_iterator<it_t>,
                                                 typename std::random_access_iterator_tag,
                                                 iterator_tag_t<it_t>>;
```
If the underlying range is contiguous, we can just say we are random access, because we know the memory location of each element.
If not, we need to check the iterator regarding the concept it models / the tag it provides.

So if you model at most forward, just check if the underlying range is forward. Otherwise you are input.

include/seqan3/range/views/minimiser.hpp
MitraDarja: When I put window_first in the constructor, the test for the forward_range fails, but everything else passes. Are we okay with losing the foward_range property?

include/seqan3/range/views/minimiser.hpp
MitraDarja: True, should I then throw when the user enters a window_value_size of 1?

include/seqan3/range/views/minimiser.hpp
MitraDarja: No, what would be better? Comparable views?

include/seqan3/range/views/minimiser.hpp
smehringer: Yes, that would be good

include/seqan3/range/views/minimiser.hpp
smehringer: Hmm this is weird. Can you check why we loose that property? We only change something internally, it should not have an influence on the public API

include/seqan3/range/views/minimiser.hpp
smehringer: Yes, but why is the `iterator_category` different than the  `iterator_concept`?

include/seqan3/range/views/minimiser.hpp
MitraDarja: The reason is, that it then uses the input_iterator_tag instead of the forward one. Removing the lines about the iterator_concept solved this issue, so I would just do that?

include/seqan3/range/views/minimiser.hpp
MitraDarja: When I remove the `iterator_concept` all test pass as well (and I do not have the issue described below), so I would just remove it?

include/seqan3/range/views/minimiser.hpp
smehringer: I just checked. We only have two categories so far. I guess it's `General purpose views` then

include/seqan3/range/views/minimiser.hpp
smehringer: No, I don't think we should remove the iterator category. I don't remember the exact details (@rrahn can you help) but I think this is needed to be correctly identifier as the respective range in c++20.
We need to check why this happens. If you push the latest changes and are done with the rest I take a look if you like.

MitraDarja: That would be nice, thanks. :) 
I have pushed everything else now.

include/seqan3/range/views/minimiser.hpp
eseiler: `iterator_category ` is pre-c++20 and is a tag, `iterator_concept` is c++20 and is a concept.
The concepts and tags have different requirements, hence they may be different - maybe not in this view though.

smehringer: Thank you! Do you know if for the respective concept (LegacyInputIterator, input_iterator) there is an actual c++ concept?
I wonder if 
`std::input_iterator` tests for LegacyInputIterator and 
`std::ranges::input_iterator` tests for input_iterator?

Then I would propose
```cpp
    //!\brief Tag this class as forward iterator.
    using iterator_category = std::conditional_t<std::forward_iterator<it_t>,
                                                 std::forward_iterator_tag,
                                                 std::input_iterator_tag>;
    //!\brief Tag this class depending on which concept `it_t` models.
    using iterator_concept = std::conditional_t<std::ranges::forward_iterator<it_t>,
                                                std::forward_iterator_tag,
                                                std::input_iterator_tag>;
```

eseiler: AFAIK the category is something the developer promises, there is just an exposition-only concept (since c++20) for this.

IIRC, in c++20 you do not really need the category because you have concepts.
However (legacy) algorithms may check the category to determine what they can do.
That means you still have to provide the category.
Additionally you can define the concept, which directly refers to the c++20 concepts. This is probably gonna be used somewhere, otherwise it wouldn't be in the standard - maybe for non-c++20 stuff that is aware of c++concepts...

However, both are different - something with proxy iterators....

Couldn't you just set the category to input iff  `seqan3::iterator_tag_t` says input otherwise it's forward?
And the concept would be the same as the category I think

smehringer: > Couldn't you just set the category to input iff seqan3::iterator_tag_t says input otherwise it's forward?

Isn't this equivalent to what I did? If the underlying range is more then input we model forward otherwise input.

> And the concept would be the same as the category I think

After you explanation (thanks) yes, I agree.

test/unit/range/views/view_minimiser_test.cpp
```diff
@@ -0,0 +1,116 @@
...
+    std::vector<seqan3::dna4> text1{"AAAAAAAAAA"_dna4};
+    std::vector<seqan3::dna4> text1_short{"AAAAAA"_dna4};
+    result_t result1{0, 0, 0}; // Same result for ungapped and gapped
+    result_t result1a{0};      // windows_size == text_size, same result for ungapped and gapped
+
+    std::vector<seqan3::dna4> text2{"AC"_dna4};
+    result_t result2{}; // Same result for ungapped and gapped
+
+    seqan3::bitcompressed_vector<seqan3::dna4> text3{"ACGGCGACGTTTAG"_dna4};
+    result_t ungapped_no_rev3{26, 97, 27};    // ACGG, CGAC, ACGT
```
smehringer, line 56: What does `no_rev` mean? You call it `result1` for `text1` but `ungapped_no_rev3` for `text3` please by consistent with the naming scheme one way or the other (result vs no_rev).

Also, sometimes you store a text in here, but then you store some texts within the individual tests.
In general, you would store the text here, if you re-use it in multiple tests.

test/unit/range/views/view_minimiser_test.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+TYPED_TEST_SUITE(minimiser_view_properties_test, underlying_range_types, );
+
+class minimiser_test : public ::testing::Test
+{
+protected:
+    std::vector<seqan3::dna4> text1{"AAAAAAAAAA"_dna4};
+    std::vector<seqan3::dna4> text1_short{"AAAAAA"_dna4};
+    result_t result1{0, 0, 0}; // Same result for ungapped and gapped
+    result_t result1a{0};      // windows_size == text_size, same result for ungapped and gapped
```
smehringer, line 50:
```suggestion
    result_t result1_short{0};      // windows_size == text_size, same result for ungapped and gapped
```
or what does `a` mean?

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,476 @@
...
+    {
+        return lhs == rhs.window_right;
+    }
+
+    //!\brief Compare to the sentinel of the underlying range.
+    friend bool operator==(window_iterator const & lhs, window_iterator const & rhs) noexcept
+    {
+        return std::tuple(lhs.window_right, lhs.window_values.size()) ==
+               std::tuple(rhs.window_right, rhs.window_values.size());
+    }
```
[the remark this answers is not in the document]
MitraDarja: The reason why I changed it to tuple was that it got rid of the error "cannot bind non-const lvalue reference of type ‘long unsigned int&’ to an rvalue of type ‘std::deque<long unsigned int>::size_type’ {aka ‘long unsigned int’}"

test/unit/range/views/view_minimiser_test.cpp
MitraDarja: no_rev stands for no reverse strand. I would rename the results for text3 then result3_ungapped_no_rev and result3_gapped_no_rev, to be more consistent? (For result1 and result2 no specification is necessary, because the results are the same for ungapped, gapped and reverse and no reverse).
I only have texts in the typed test, because the text there is of type TypeParam and as far as I understood it, TypeParam is only declared inside of a typed test. (?)

include/seqan3/range/views/minimiser.hpp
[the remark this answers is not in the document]
smehringer: ah yes, that because .size() returns a temporary and cannot be bound by std::tie.
Then maybe just do it like in the operators below
```suggestion
    friend bool operator==(window_iterator const & lhs, window_iterator const & rhs) noexcept
    {
        return lhs.window_right == rhs.window_right &&
               lhs.window_values.size() == rhs.window_values.size();
    }
```

test/unit/range/views/view_minimiser_test.cpp
smehringer: Alright with the naming

> I only have texts in the typed test, because the text there is of type TypeParam and as far as I understood it, TypeParam is only declared inside of a typed test. (?)

The `TypeParam` yes, but it is the same as the template paramter of the test fixture.
Anyway, I don't want to spend a lot of time making tests pretty so just leave it as it is :)

include/seqan3/range/views/minimiser.hpp
```diff
@@ -0,0 +1,480 @@
...
+    using sentinel_t = std::ranges::sentinel_t<rng_t>;
+
+public:
+    /*!\name Associated types
+     * \{
+     */
+    //!\brief Type for distances between iterators.
+    using difference_type = typename std::iter_difference_t<it_t>;
+    //!\brief Value type of this iterator.
+    using value_type = reference_t<urng_t>;
```
smehringer, line 182:
```suggestion
    using value_type = std::ranges::range_value_t<urng_t>;
```
we don't want the value type to have a reference and the use the ranges function not our own (Wiep is changing that)

test/unit/range/views/view_minimiser_test.cpp
```diff
@@ -0,0 +1,137 @@
...
+    result_t gapped_no_rev{3, 5, 3};                     // A--T, C--C, A--T - "-" for gap
+    EXPECT_EQ(ungapped_no_rev, text | kmer_view | minimiser_view | seqan3::views::to<result_t>);
+    EXPECT_EQ(gapped_no_rev, text | gapped_kmer_view | minimiser_view | seqan3::views::to<result_t>);
+}
+
+TEST_F(minimiser_test, ungapped_kmer_hash)
+{
+    EXPECT_EQ(result1, text1 | kmer_view | minimiser_view | seqan3::views::to<result_t>);
+    EXPECT_THROW(text1_short | kmer_view | minimiser_view2, std::invalid_argument);
+    EXPECT_EQ(result2, text2 | kmer_view | seqan3::views::to<result_t>);
```
smehringer, line 113: 1. why is there no minimizer view here?

2. please change `text2` to `too_short_text`
 
3. change the following
```suggestion
    auto empty_view = too_short_text | kmer_view;
    EXPECT_TRUE(std::ranges::empty(empty_view));
```
It will get rid of the error on jenkins.

Change the same thing in the gapped test

CHANGELOG.md
```diff
@@ -64,6 +64,9 @@ Note that 3.1.0 will be the first API stable release and interfaces in this rele
 * The `seqan3::begin()`, `seqan3::end()`, `seqan3::cbegin()`, `seqan3::cend()`, `seqan3::size()`, `seqan3::empty()`
   functions have been deprecated:
   Use `std::ranges::{begin|end|cbegin|cend|size|empty}()` instead ([\#1663](https://github.com/seqan/seqan3/pull/1663)).
+* Added `seqan3::views::minimiser_view`, a view that minimisers for a range of comparable values given the
+  window_values_size
```
smehringer, line 68:
```suggestion
* Added `seqan3::views::minimiser`, a view that computes the minimum in a window shifted over a range of comparable values.
```

MitraDarja:
@smehringer Can you merge this? 😃

PR #1656, opened 2020-03-19
[TEST] add seqan2 chunked output to benchmark

h-2:
SeqAn2 could do chunked output already, but SeqAn3's implementation is ~5% faster.

### DIFF
--- a/test/performance/io/lowlevel_stream_output_benchmark.cpp
+++ b/test/performance/io/lowlevel_stream_output_benchmark.cpp
@@ -75,10 +75,12 @@ void write_all(benchmark::State & state)
         {
             it.write_range(cont_rando);
         }
+        #ifdef SEQAN3_HAS_SEQAN2
         else if constexpr (id == tag::seqan2_stream_it_write_range)
         {
             seqan::write(it, cont_rando2);
         }
+        #endif
         else
         {
             for (auto chr : cont_rando)
--- a/test/performance/io/lowlevel_stream_output_benchmark.cpp
+++ b/test/performance/io/lowlevel_stream_output_benchmark.cpp
@@ -27,7 +27,8 @@ enum class tag
     std_streambuf_it,
     seqan3_streambuf_it,
     seqan3_streambuf_it_write_range,
-    seqan2_stream_it
+    seqan2_stream_it,
+    seqan2_stream_it_write_range
 };
 
 template <tag id>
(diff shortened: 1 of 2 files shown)


test/performance/io/lowlevel_stream_output_benchmark.cpp
```diff
@@ -40,6 +41,9 @@ void write_all(benchmark::State & state)
     // sequence to write:
     std::vector<char> cont_rando = seqan3::test::generate_sequence<char>(10'000, 0, 0);
 
+#ifdef SEQAN3_HAS_SEQAN2
+    auto cont_rando2 = seqan3::test::generate_sequence_seqan2<char>(10'000, 0, 0);
```
smehringer: The comparison should be fair even though the sequence `cont_rando` and `cont_rando2` will differ right?

h-2: Yes! There is no difference. It's only important that it's not all zeros.

rrahn: Maybe reconsider better names for `cont_rando`. This sounds like a bad name for a 🐶

h-2: It *is* the name of my :dog2: . Don't mess with it.

rrahn: that pure little thing. 😄

rrahn, line 45:
```suggestion
    auto seqan2_sequence = seqan3::test::generate_sequence_seqan2<char>(10'000, 0, 0);
```

test/performance/io/lowlevel_stream_output_benchmark.cpp
```diff
@@ -40,6 +41,9 @@ void write_all(benchmark::State & state)
     // sequence to write:
     std::vector<char> cont_rando = seqan3::test::generate_sequence<char>(10'000, 0, 0);
```
rrahn, line 42:
```suggestion
    std::vector<char> sequence = seqan3::test::generate_sequence<char>(10'000, 0, 0);
```

test/performance/io/lowlevel_stream_output_benchmark.cpp
```diff
@@ -71,6 +75,12 @@ void write_all(benchmark::State & state)
         {
             it.write_range(cont_rando);
```
rrahn, line 76:
```suggestion
            it.write_range(sequence);
```

test/performance/io/lowlevel_stream_output_benchmark.cpp
```diff
@@ -71,6 +75,12 @@ void write_all(benchmark::State & state)
...
         }
+        #ifdef SEQAN3_HAS_SEQAN2
+        else if constexpr (id == tag::seqan2_stream_it_write_range)
+        {
+            seqan::write(it, cont_rando2);
```
rrahn, line 81:
```suggestion
            seqan::write(it, seqan2_sequence);
```

PR #1663, opened 2020-03-20
[FIX] Deprecate seqan3::{begin, end, size}

### LINKED ISSUE #1662: Use std::ranges::{begin, end, size} instead of seqan3::{begin, end, size}
I had troubles to find where we said that we wanted to change this, but #498 was one that did this partially. And we talked about this in the core team before (probably in meeting https://github.com/seqan/seqan3/wiki/Core-Meeting-notes#2019-12-09-rrahn).

* [x] Deprecate `seqan3::{begin, end, size}`
* [x] Remove usage in seqan3.
* [x] Don't forget Changelog

marehr:
Resolves #1662

* [x] Changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,12 @@ Note that 3.1.0 will be the first API stable release and interfaces in this rele
 
 ## API changes
 
+#### Range
+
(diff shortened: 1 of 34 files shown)


rrahn:
Looks good so far. Can you please udpate the changelog and resolve the last two commit messages.

marehr:
@rrahn I added the changelog

PR #1666, opened 2020-03-20
[TEST] Don't filter out whitespaces in strings 

marehr:


smehringer:
This change will lead to spurious test failures because of the possibility of line wraps when terminal sizes differ. You can reproduce this easily by narrowing your terminal and running the test again.
That's the reason why we remove white spaces.

marehr:
@smehringer Thank you, I accounted this by making it possible to set the terminal at a fixed width within the tests.

smehringer:
@marehr That's a good idea. Now we also have a diff when things fail don't we?

include/seqan3/argument_parser/detail/format_help.hpp
```diff
@@ -87,20 +87,23 @@ class format_help : public format_help_base<format_help>
         unsigned rightColumnTab;
 
         //!\brief The constructor.
-        console_layout_struct() :
+        //!\param[in] terminal_width The width of the terminal.
+        console_layout_struct(unsigned terminal_width) :
```
rrahn:
```suggestion
        console_layout_struct(uint32_t const terminal_width) :
```

marehr:
> @marehr That's a good idea. Now we also have a diff when things fail don't we?

Yes :+1:

include/seqan3/argument_parser/detail/format_help.hpp
marehr: The underlying type is `unsigned`, so either change all of the fields to `uint32_t` or use `unsigned` in the constructor.

rrahn: Yes change all of the fields. That's what I am implying 😏

test/unit/argument_parser/detail/format_help_test.cpp
```diff
@@ -32,55 +32,78 @@ const char * argv1[] = {"./help_add_test --version-check 0", "-h"};
...
+                                      "    -hh, --advanced-help\n"
+                                      "          Prints the help page including advanced options.\n"
+                                      "    --version\n"
+                                      "          Prints the version information.\n"
+                                      "    --copyright\n"
+                                      "          Prints the copyright/license information.\n"
+                                      "    --export-help (std::string)\n"
+                                      "          Export the help page information. Value must be one of [html, man].\n"
+                                      "    --version-check (bool)\n"
+                                      "          Whether to to check for the newest app version. Default: 1.\n";
```
joergi-w: repetitive `to`: should be fixed in another PR probably

test/unit/argument_parser/format_parse_validators_test.cpp
```diff
@@ -605,25 +659,31 @@ TEST(validator_test, arithmetic_range_validator_success)
...
-    std::string expected = std::string("test_parser"
-                           "==========="
-                           "POSITIONAL ARGUMENTS"
-                           "    ARGUMENT-1 (List of signed 32 bit integer's)"
-                           "          desc Default: []. Value must be in range [-20,20]." +
+    std::string expected = std::string("test_parser\n"
+                           "===========\n"
+                           "\n"
+                           "POSITIONAL ARGUMENTS\n"
+                           "    ARGUMENT-1 (List of signed 32 bit integer's)\n"
```
joergi-w: No apostrophe in `integer's`: This is plural form, not genitive.

test/unit/argument_parser/format_parse_validators_test.cpp
```diff
@@ -919,21 +1003,26 @@ TEST(validator_test, regex_validator_success)
...
-                           "===========" +
+    std::string expected = std::string("test_parser\n"
+                           "===========\n"
+                           "\n" +
                            basic_options_str +
-                           "    -s, --string-option (List of std::string's)"
-                           "          desc Default: []. Value must match the pattern '[a-zA-Z]+@[a-zA-Z]+\\.com'." +
+                           "\n"
+                           "  \n"
+                           "    -s, --string-option (List of std::string's)\n"
```
joergi-w: the same

marehr: I added a Card https://github.com/seqan/seqan3/projects/7#card-35506628 for these little wording mistakes.

test/unit/argument_parser/detail/format_help_test.cpp
```diff
@@ -58,35 +59,57 @@ std::string license_text()
...
     EXPECT_EXIT(parser1.parse(), ::testing::ExitedWithCode(EXIT_SUCCESS), "");
     std_cout = testing::internal::GetCapturedStdout();
-    expected = "test_parser"
-               "===========" +
+    expected = "test_parser\n"
+               "===========\n"
+               "\n" +
                basic_options_str +
+               "\n"
+               "  \n"
```
smehringer, line 109: Seems like this is part of the `basic_option_str` no? Then you don't have to add it every time manually?

marehr: Since the tests pass, I would say that is not part of `basic_option_str` :)

smehringer: Of course the current tests pass because you added this after every single occurrence of `basic_options_str` no matter what follows afterwards? :) 

It actually is part of it, see [here](https://github.com/seqan/seqan3/blob/master/include/seqan3/argument_parser/argument_parser.hpp#L708) (Although I would not like to point out this open todo in the code.. :speak_no_evil: ).

PR #1669, opened 2020-03-20
[MISC] argument_parser: use predicate instead of seqan3::views::to_lower

marehr:


include/seqan3/argument_parser/validators.hpp
```diff
@@ -375,13 +374,16 @@ class file_validator_base
         auto drop_less_ext = tmp_str | views::drop(1);
 
         // Compares the extensions in lower case.
-        auto cmp_lambda = [&] (std::string const & cmp)
+        auto case_insensitive_cmp = [&] (std::string const & ext)
         {
-            return std::ranges::equal(cmp | views::to_lower, drop_less_ext | views::to_lower);
+            return std::ranges::equal(ext, drop_less_ext, [](auto && chr1, auto && chr2)
```
smehringer: We expect these to be characters since we are comparing strings right?
then I think we should capture by value not by reference.

:nail_care: If you are being more expressive with the name, you might also want to change it to case_insensitive`_comparator`? :)

:nail_care: space. 
```suggestion
            return std::ranges::equal(ext, drop_less_ext, [] (auto && chr1, auto && chr2)
```

include/seqan3/argument_parser/validators.hpp
```diff
@@ -375,13 +374,16 @@ class file_validator_base
         auto drop_less_ext = tmp_str | views::drop(1);
 
         // Compares the extensions in lower case.
-        auto cmp_lambda = [&] (std::string const & cmp)
+        auto case_insensitive_equal_to = [&] (std::string const & ext)
         {
-            return std::ranges::equal(cmp | views::to_lower, drop_less_ext | views::to_lower);
+            return std::ranges::equal(ext, drop_less_ext, [] (char const chr1, char const chr2)
+                   {
+                       return std::tolower(chr1) == std::tolower(chr2);
```
rrahn: not that performance would be an issue here but we do have our own seqan3::to_lower predicates in core char_operations. It also does not pull the ranges...

marehr: I know, but this is one step in decoupling argument parser from seqan3.

rrahn: yajk I forgot.

PR #1672, opened 2020-03-22
[FIX] Add missing front and back member function to span

### LINKED ISSUE #1676: Our span does not offer front/back interface.
Add front back interface to our std/span implementation.

The standard says: https://eel.is/c++draft/span.elem

```c++
constexpr reference front() const;
// Preconditions: empty() is false.
// Effects: Equivalent to: return *data();

constexpr reference back() const;
// Preconditions: empty() is false.
// Effects: Equivalent to: return *(data() + (size() - 1));
```

rrahn:
Fixes #1676

### DIFF
--- a/include/seqan3/std/span
+++ b/include/seqan3/std/span
@@ -185,6 +185,16 @@ public:
     constexpr index_type size_bytes() const noexcept { return span_extent * sizeof(element_type); }
     constexpr bool empty()            const noexcept { return span_extent == 0; }
 
+    constexpr reference front() const noexcept
+    {
+        return *begin();
+    }
+
+    constexpr reference back() const noexcept
+    {
+        return *(end() - 1);
+    }
+
     constexpr reference operator[](index_type idx) const noexcept
     {
         assert(idx >= 0 && idx < size()); // "span<T,N>[] index out of bounds"
@@ -347,6 +357,16 @@ public:
     constexpr index_type size()       const noexcept { return size_; }
     constexpr index_type size_bytes() const noexcept { return size_ * sizeof(element_type); }
     constexpr bool empty()            const noexcept { return size_ == 0; }
+
+    constexpr reference front() const noexcept
+    {
+        return *begin();
+    }
+
+    constexpr reference back() const noexcept
+    {
+        return *(end() - 1);
+    }
 
     constexpr reference operator[](index_type idx) const noexcept
     {


include/seqan3/std/span
```diff
@@ -348,6 +358,16 @@ public:
     constexpr index_type size_bytes() const noexcept { return size_ * sizeof(element_type); }
     constexpr bool empty()            const noexcept { return size_ == 0; }
 
+    constexpr reference front() const noexcept
+    {
+        return this->operator[](0);
```
marehr: wouldn't this be more intuitive?

```suggestion
        return this->operator[0];
```

rrahn: but this does not work. Only this would work `(*this)[0]` AFAIK but I think it is not more readable.

marehr: Ah I thought that you tried this.

We have different styles right now:

https://github.com/seqan/seqan3/blob/a7278bbb55800ae9216d27c2212984aa92b57022/include/seqan3/core/char_operations/predicate_detail.hpp#L49-L55

https://github.com/seqan/seqan3/blob/a7278bbb55800ae9216d27c2212984aa92b57022/include/seqan3/range/views/take.hpp#L273-L280

(I only found these occurrences)

Far more common in our code base is:

https://github.com/seqan/seqan3/blob/a7278bbb55800ae9216d27c2212984aa92b57022/include/seqan3/io/alignment_file/sam_tag_dictionary.hpp#L350-L360

https://github.com/seqan/seqan3/blob/a7278bbb55800ae9216d27c2212984aa92b57022/include/seqan3/range/container/bitcompressed_vector.hpp#L425-L432

https://github.com/seqan/seqan3/blob/a7278bbb55800ae9216d27c2212984aa92b57022/include/seqan3/range/container/concatenated_sequences.hpp#L526-L532

https://github.com/seqan/seqan3/blob/a7278bbb55800ae9216d27c2212984aa92b57022/include/seqan3/range/container/dynamic_bitset.hpp#L974-L980

https://github.com/seqan/seqan3/blob/a7278bbb55800ae9216d27c2212984aa92b57022/include/seqan3/range/container/small_vector.hpp#L369-L376

https://github.com/seqan/seqan3/blob/a7278bbb55800ae9216d27c2212984aa92b57022/include/seqan3/range/decorator/gap_decorator.hpp#L368-L373

rrahn: So what should I do now? I think calling out the operator helps to understand what is meant.  But I don't really care. Also keep in mind that this is just inside the std module which is kind of external and not necessarily applying to all our rules.

marehr: Let's keep your way, I just wanted to say that we use the other style more often and that it is not less readable.

marehr:
Changelog entry?

include/seqan3/std/span
```diff
@@ -185,6 +185,16 @@ public:
     constexpr index_type size_bytes() const noexcept { return span_extent * sizeof(element_type); }
     constexpr bool empty()            const noexcept { return span_extent == 0; }
 
+    constexpr reference front() const noexcept
+    {
+        return this->operator[](0);
```
smehringer: This is real picky, so feel free to ignore it, but on cppreference it is documented as returning `*this->begin()`. Won't make a difference ever right? (I checked because I did not know that span only has a const version of `front()` but of course that makes sense :woman_facepalming:)

Funnily, `back()` is documented as returning `*(this->end()-1)`.. I thought one always avoids this.

rrahn:
I'd like to not overwhelm the changelog with mini fixes that are only reported internally and obviously did not break anything before. This falls under the category several minor fixes. We should however document things if they are reported from external site.

include/seqan3/std/span
marehr: https://eel.is/c++draft/span.elem says:

```c++
constexpr reference operator[](size_type idx) const;
// Preconditions: idx < size() is true.
// Effects: Equivalent to: return *(data() + idx);

constexpr reference front() const;
// Preconditions: empty() is false.
// Effects: Equivalent to: return *data();

constexpr reference back() const;
// Preconditions: empty() is false.
// Effects: Equivalent to: return *(data() + (size() - 1));

constexpr pointer data() const noexcept;
// Effects: Equivalent to: return data_­;
```

smehringer:
I agree that this is notable because although it does not break stuff it directly affects public API.

If we don't want to overwhelm the Changelog we just have to do more frequent releases :D

rrahn:
> I would say that this should be added. It is not a notable change, but it is still a change that is more than a mini-fix.

Is it a fix because we didn't have a standard conform implementation? Is it a feature because no one ever said that this is standard conform and front and back are not used generically on ranges yet. So if a closure object returns a span we can't use back/front on that in a generic context. So only if you work with a span. Where the next point comes into play

> it directly affects public API

Technically not, since it isn't even documented in the public API. No one even knows that we have a span implementation unless he/she looks into the code.

So I can merely add a documentation entry and add in the changelog that we know have span impelemented as defined by C++20 if that makes you happy.

PR #1675, opened 2020-03-22
[MISC] io: move include from io/detail/misc to io/alignment_file/format_sam

marehr:
Remove unnecessary includes


MitraDarja:
What is the advantage of moving the include? Plus travis does not work... Can I/Should I approve independent of that?

marehr:
> What is the advantage of moving the include? Plus travis does not work... Can I/Should I approve independent of that?

I think I resolved the build failure now. As you can see, this include at the wrong place did implicitly include (more) library code than needed.

PR #1677, opened 2020-03-22
[FIX] don't use seqan3::is_char in std/charconv

marehr:
Our `seqan3/std/*` headers (shim/polyfill headers) should not include anything
from seqan3 to avoid build failures like missing includes if the standard
headers are available.

Fixes partially #1678

### DIFF
--- a/include/seqan3/argument_parser/detail/format_parse.hpp
+++ b/include/seqan3/argument_parser/detail/format_parse.hpp
@@ -17,6 +17,7 @@
 #include <vector>
 
 #include <seqan3/argument_parser/detail/format_base.hpp>
+#include <seqan3/core/char_operations/predicate.hpp>
 #include <seqan3/core/detail/type_inspection.hpp>
 #include <seqan3/std/concepts>
 #include <seqan3/std/charconv>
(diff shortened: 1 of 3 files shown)


include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -17,6 +17,7 @@
 #include <vector>
 
 #include <seqan3/argument_parser/detail/format_base.hpp>
+#include <seqan3/core/char_operations/predicate.hpp>
```
rrahn: But this is still necessary for the argument parser?

marehr: the argument parser uses right now some functionality of `char_operations/predicate.hpp`, but it forgot to add the include, because it was indirectly included by `seqan3/std/charconv`, which the argument parser also includes.

rrahn:
something broke in the unit tests

marehr:
@rrahn Was fixed, can I merge? Travis did run through and passed, but here it shows it as pending?!

marehr:
I retriggered the build.

Issue #1678, opened 2020-03-22
Remove dependencies of seqan3 in `seqan3/std/*` headers


marehr:
Our `seqan3/std/*` headers (shim/polyfill headers) should not include anything
from seqan3 to avoid build failures like missing includes if the standard
headers are available.

PR #1680, opened 2020-03-22
[FIX] simplify std::ranges::enable_view overloads

marehr:
A change in https://github.com/ericniebler/range-v3/pull/1455 simplified
the heuristics for std::ranges::enable_view. So the gap_decorator is by
default now no view anymore.

This PR is also in line with #1678 to remove dependencies of our library
in the std headers.

---

old description

I know that this introduces a macro (SEQAN3_STD_RANGES_NAMESPACE), but
this reduces code bloat for a feature that is only used for a single
data structure right now.

This PR is also in line with #1678 to remove dependencies of our library
in the std headers.


h-2:
I managed to get the weird heuristics out of the standard so none of this code is required anymore. Changes to range-v3 were committed a couple of days ago so you could remove the code after upgrading: 
https://github.com/ericniebler/range-v3/pull/1455

marehr:
@h-2 Okay, does that change any impact on our other views? Or do they all inherit from `std::ranges::view_interface`?

h-2:
All our views inherit `std::ranges::view_interface` or `std::ranges::view_base` if I am not mistaken.

rrahn:
> I managed to get the weird heuristics out of the standard so none of this code is required anymore. 

Ah that is good!

PR #1683, opened 2020-03-22
[DOC] add closing comment namespace seqan3

marehr:


marehr:
> Just a quick understanding question:
> What does this `//namespace seqan3` tell the reader ?

```c++
namespace seqan3
{
//...
} // namespace seqan3
```

 without the comment we don't know that this closing brace is for the namespace

PR #1685, opened 2020-03-23
[FIX] sort all hits and remove duplicates once

svnbgnk:
Instead of sorting and "uniquifying" the added locations from each cursor. All result are uniquifying at once.

### DIFF
--- a/include/seqan3/search/algorithm/detail/search.hpp
+++ b/include/seqan3/search/algorithm/detail/search.hpp
@@ -152,9 +152,9 @@ inline auto search_single(index_t const & index, query_t & query, configuration_
             {
                 for (auto const & text_pos : cur.locate())
                     hits.push_back(text_pos);
-                std::sort(hits.begin(), hits.end());
-                hits.erase(std::unique(hits.begin(), hits.end()), hits.end());
             }
+            std::sort(hits.begin(), hits.end());
+            hits.erase(std::unique(hits.begin(), hits.end()), hits.end());
         }
         return hits;
     }


smehringer:
@svnbgnk I think it is possible to merge/uniqify the cursor ranges before calling `locate()` on each position, since this would be even more efficient we should maybe tackle it in one pr. @eseiler already tried this but I don't know if there is a branch somewhere..

marehr:
> @svnbgnk I think it is possible to merge/uniqify the cursor ranges before calling `locate()` on each position, since this would be even more efficient we should maybe tackle it in one pr. @eseiler already tried this but I don't know if there is a branch somewhere..

This seems like an immediate and easy fix (even to rebase). What you mentioned sounds like a bigger restructuring. From my POV I think this PR is okay in the meantime until this happened.

smehringer:
@marehr
If the result contains millions of queries (not unlikely), I'm afraid that this performance improvement might cause a major memory regression :( Of course, I cannot be sure but this would mean it has to be tested beforehand, but then maybe the effort should be put into the even better solution right away? If you think testing it is worth it, it's alright with me, no strong feelings. It was just that I had the same idea and then Enrico was concerned about the memory footprint and we came up with the merging which seemed to increase performance significantly.

rrahn:
I am not sure I understand what you mean with millions of queries? These are only the results of a single query. If one query results in 1 million hits then we are talking about 8 or 16 Mb memory. Since we changed to the lazy result range we do not buffer the results of all queries anymore but only of a single query.

smehringer:
Sorry, I meant a single query with lots of hits (database search)

h-2:
See also https://github.com/seqan/seqan3/pull/1528

For many search schemes, we don't need to sort/make_unique at all, e.g. search schemes already return unique results. This should be encoded in that type of search and avoided.

svnbgnk:
@h-2 Foe Search Schemes results are only unique for hamming distance. 
Most of the time we sort a mostly sorted vector. The performance for that probably varies from each compiler (https://solarianprogrammer.com/2012/10/24/cpp-11-sort-benchmark/).

h-2:
> @h-2 Foe Search Schemes results are only unique for hamming distance.

Sure, but this is likely the use-case 90% of people use :smile: And this function is called millions of times! 

If it is known which parameter combinations lead to a unique result-list, why should we sort this again?

rrahn:
> Sure, but this is likely the use-case 90% of people use

I wouldn't support that in general. However, it seems anyway like a good approach to not do any unncessary operations if we can avoid it and if the information is already statically available. So I would take the presented ideas and cover them in the project board for the future. But this PR seems rather like a no-brainer to me that can be merged with no harm.

PR #1688, opened 2020-03-24
[FIX] fm_index_cursor::path_label for 2D texts should not return views::join | views::slice

### LINKED ISSUE #1646: fm_index_cursor::path_label for 2D texts should not return views::join | views::slice
`fm_index_cursor::path_label` for two-dimensional texts that are random_access_ranges over random_access_ranges should not return `views::join | views::slice` because `views::join` is slow and breaks random access. Instead return a slice on the i-th element of the text.

(Added by h-2)

Irallia:
Resolves #1646
I made the change for fm_index_cursor and bi_fm_index_cursor. 
I also fixed a snippet and corrected some snippet comments.

### DIFF
--- a/include/seqan3/search/fm_index/bi_fm_index_cursor.hpp
+++ b/include/seqan3/search/fm_index/bi_fm_index_cursor.hpp
@@ -908,14 +908,21 @@ public:
                       "The alphabet types of the given text and index differ.");
         assert(index != nullptr);
 
-        size_type const loc = offset() - index->fwd_fm.index[fwd_lb];   // Position of query in concatenated texts.
-        size_type const text_idx = index->fwd_fm.text_begin_rs.rank(loc + 1) - 1; // Position of subtext in text vector,
-                                                                        // where the query was found
-                                                                        // = rank of Bitvector of the concatenated texts
-        size_type const query_begin = loc - index->fwd_fm.text_begin_ss.select(text_idx + 1);
-                                                                        // Substract lengths of previous sequences
-        return text[text_idx] | views::slice(query_begin, query_begin + query_length());
-                                                                        // Take subtext, slice query out of it    }
+        // Position of query in concatenated text.
+        size_type const location = offset() - index->fwd_fm.index[fwd_lb];
+        // The rank represents the number of delimiters before position `location + 1`.
+        // Furthermore, this number represents the i-th text this (text) position originates from, since each delimiter
+        // corresponds to one text in the whole concatenated text from left-to-right.
+        size_type const rank = index->fwd_fm.text_begin_rs.rank(location + 1);
+        // The sum of all text lengths prior to the i-th text is the start location of the i-th text in the concatenated
+        // sequence.
+        size_type const start_location = index->fwd_fm.text_begin_ss.select(rank);
+        assert(rank > 0);
+        size_type const text_id = rank - 1;
+        // Substract lengths of previous sequences.
+        size_type const query_begin = location - start_location;
+        // Take subtext, slice query out of it <-my
+        return text[text_id] | views::slice(query_begin, query_begin + query_length());
     }
 
     /*!\brief Counts the number of occurrences of the searched query in the text.
--- a/include/seqan3/search/fm_index/bi_fm_index_cursor.hpp
+++ b/include/seqan3/search/fm_index/bi_fm_index_cursor.hpp
@@ -19,7 +19,6 @@
 #include <seqan3/alphabet/adaptation/char.hpp>
 #include <seqan3/alphabet/adaptation/uint.hpp>
 #include <seqan3/core/type_traits/range.hpp>
-#include <seqan3/range/views/join.hpp>
 #include <seqan3/range/views/slice.hpp>
 #include <seqan3/search/fm_index/bi_fm_index.hpp>
 #include <seqan3/std/ranges>
@@ -911,20 +910,18 @@ public:
         // Position of query in concatenated text.
         size_type const location = offset() - index->fwd_fm.index[fwd_lb];
 
-        // The rank represents the number of delimiters before position `location + 1`.
-        // Furthermore, this number represents the i-th text this (text) position originates from, since each delimiter
-        // corresponds to one text in the whole concatenated text from left-to-right.
+        // The rank represents the number of start positions of the individual sequences/texts in the collection
+        // before position `location + 1` and thereby also the number of delimiters.
         size_type const rank = index->fwd_fm.text_begin_rs.rank(location + 1);
         assert(rank > 0);
         size_type const text_id = rank - 1;
 
-        // The sum of all text lengths prior to the i-th text is the start location of the i-th text in the concatenated
-        // sequence.
+        // The start location of the `text_id`-th text in the sequence (position of the `rank`-th 1 in the bitvector).
         size_type const start_location = index->fwd_fm.text_begin_ss.select(rank);
         // Substract lengths of previous sequences.
         size_type const query_begin = location - start_location;
 
-        // Take subtext, slice query out of it <-my
+        // Take subtext, slice query out of it
         return text[text_id] | views::slice(query_begin, query_begin + query_length());
     }
 
--- a/include/seqan3/search/fm_index/bi_fm_index_cursor.hpp
+++ b/include/seqan3/search/fm_index/bi_fm_index_cursor.hpp
@@ -908,9 +908,14 @@ public:
                       "The alphabet types of the given text and index differ.");
         assert(index != nullptr);
 
-        size_type const loc = offset() - index->fwd_fm.index[fwd_lb];
-        size_type const query_begin = loc - index->fwd_fm.text_begin_rs.rank(loc + 1) + 1; // Substract delimiters
-        return text | views::join | views::slice(query_begin, query_begin + query_length());
+        size_type const loc = offset() - index->fwd_fm.index[fwd_lb];   // Position of query in concatenated texts.
+        size_type const text_idx = index->fwd_fm.text_begin_rs.rank(loc + 1) - 1; // Position of subtext in text vector,
+                                                                        // where the query was found
+                                                                        // = rank of Bitvector of the concatenated texts
+        size_type const query_begin = loc - index->fwd_fm.text_begin_ss.select(text_idx + 1);
+                                                                        // Substract lengths of previous sequences
+        return text[text_idx] | views::slice(query_begin, query_begin + query_length());
+                                                                        // Take subtext, slice query out of it    }
     }
 
     /*!\brief Counts the number of occurrences of the searched query in the text.
--- a/include/seqan3/search/fm_index/bi_fm_index_cursor.hpp
+++ b/include/seqan3/search/fm_index/bi_fm_index_cursor.hpp
@@ -910,17 +910,20 @@ public:
 
         // Position of query in concatenated text.
         size_type const location = offset() - index->fwd_fm.index[fwd_lb];
+
         // The rank represents the number of delimiters before position `location + 1`.
         // Furthermore, this number represents the i-th text this (text) position originates from, since each delimiter
         // corresponds to one text in the whole concatenated text from left-to-right.
         size_type const rank = index->fwd_fm.text_begin_rs.rank(location + 1);
+        assert(rank > 0);
+        size_type const text_id = rank - 1;
+
         // The sum of all text lengths prior to the i-th text is the start location of the i-th text in the concatenated
         // sequence.
         size_type const start_location = index->fwd_fm.text_begin_ss.select(rank);
-        assert(rank > 0);
-        size_type const text_id = rank - 1;
         // Substract lengths of previous sequences.
         size_type const query_begin = location - start_location;
+
         // Take subtext, slice query out of it <-my
         return text[text_id] | views::slice(query_begin, query_begin + query_length());
     }
--- a/include/seqan3/search/fm_index/fm_index_cursor.hpp
+++ b/include/seqan3/search/fm_index/fm_index_cursor.hpp
@@ -482,9 +482,14 @@ public:
                       "The alphabet types of the given text and index differ.");
         assert(index != nullptr);
 
-        size_type const loc = offset() - index->index[node.lb];
-        size_type const query_begin = loc - index->text_begin_rs.rank(loc + 1) + 1; // Substract delimiters
-        return text | views::join | views::slice(query_begin, query_begin + query_length());
+        size_type const loc = offset() - index->index[node.lb];         // Position of query in concatenated texts.
+        size_type const text_idx = index->text_begin_rs.rank(loc + 1) - 1; // Position of subtext in text vector,
+                                                                        // where the query was found
+                                                                        // = rank of Bitvector of the concatenated texts
+        size_type const query_begin = loc - index->text_begin_ss.select(text_idx + 1);
+                                                                        // Substract lengths of previous sequences
+        return text[text_idx] | views::slice(query_begin, query_begin + query_length());
+                                                                        // Take subtext, slice query out of it
     }
 
     /*!\brief Counts the number of occurrences of the searched query in the text.
--- a/include/seqan3/search/fm_index/fm_index_cursor.hpp
+++ b/include/seqan3/search/fm_index/fm_index_cursor.hpp
@@ -484,18 +484,21 @@ public:
 
         // Position of query in concatenated text.
         size_type const location = offset() - index->index[node.lb];
+
         // The rank represents the number of delimiters before position `location + 1`.
         // Furthermore, this number represents the i-th text this (text) position originates from, since each delimiter
         // corresponds to one text in the whole concatenated text from left-to-right.
         size_type const rank = index->text_begin_rs.rank(location + 1);
+        assert(rank > 0);
+        size_type const text_id = rank - 1;
+
         // The sum of all text lengths prior to the i-th text is the start location of the i-th text in the concatenated
         // sequence.
         size_type const start_location = index->text_begin_ss.select(rank);
-        assert(rank > 0);
-        size_type const text_id = rank - 1;
         assert(text_id + 1 != 0);
         // Substract lengths of previous sequences.
         size_type const query_begin = location - start_location;
+
         // Take subtext, slice query out of it <-my
         return text[text_id] | views::slice(query_begin, query_begin + query_length());
     }
--- a/include/seqan3/search/fm_index/fm_index_cursor.hpp
+++ b/include/seqan3/search/fm_index/fm_index_cursor.hpp
@@ -21,7 +21,6 @@
 
 #include <seqan3/alphabet/concept.hpp>
 #include <seqan3/core/type_traits/range.hpp>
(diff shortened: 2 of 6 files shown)


marehr:
@Irallia do you know since when the snippet is broken? was it broken from the start?

Irallia:
> @Irallia do you know since when the snippet is broken? was it broken from the start?

I looked at the log history and yes it was broken from the beginning. (Commit: 8c902e25)
I just went through the snippets to help understand how ```path_label()``` works and I came across the fact that the output from test/snippet/search/bi_fm_index_cursor_to_rev_cursor.cpp made no sense to me.

include/seqan3/search/fm_index/fm_index_cursor.hpp
```diff
@@ -482,9 +482,14 @@ class fm_index_cursor
...
-        size_type const query_begin = loc - index->text_begin_rs.rank(loc + 1) + 1; // Substract delimiters
-        return text | views::join | views::slice(query_begin, query_begin + query_length());
+        size_type const loc = offset() - index->index[node.lb];         // Position of query in concatenated texts.
+        size_type const text_idx = index->text_begin_rs.rank(loc + 1) - 1; // Position of subtext in text vector,
+                                                                        // where the query was found
+                                                                        // = rank of Bitvector of the concatenated texts
+        size_type const query_begin = loc - index->text_begin_ss.select(text_idx + 1);
+                                                                        // Substract lengths of previous sequences
+        return text[text_idx] | views::slice(query_begin, query_begin + query_length());
+                                                                        // Take subtext, slice query out of it
```
marehr, line 492: Not sure why you subtract above and add again later. Can an underflow happen, i.e. `text_idx = -1`? Can `index->text_begin_rs.rank(loc + 1)` be `0`? Can you add `assert(text_idx + 1 != 0)`?

Is the data layout like this: "$FIRSTSTRING#SECONDSTRING%THIRDSTRING"? So that `rank` returns always  > 0, i.e. the number of delimiters? Do you know what happens when `loc + 1` is a delimiter? For example if it is a at position `%` in my example? Do you know why we take `loc + 1`? Does this avoid to be on the wrong delimiter?

Can you rearrange the code like this:

```suggestion
        // Position of query in concatenated text.
        size_type const location = offset() - index->index[node.lb];

        // The rank represents the number of delimiters before position `location + 1`.
        // Furthermore, this number represents the i-th text this text position originated from, since each delimiter corresponds to one text in the whole concatenated text from left-to-right.
        size_type const rank = index->text_begin_rs.rank(loc + 1);
        // The sum of all text lengths prior to the i-th text is the start location of the i-th text in the concatenated sequence.
        size_type const start_location = index->text_begin_ss.select(rank);

        assert(rank > 0);
        size_type const text_id = rank - 1;
        size_type const query_begin = location - start_location;
        return text[text_id] | views::slice(query_begin, query_begin + query_length());
```

test/snippet/search/bi_fm_index_cursor_to_rev_cursor.cpp
```diff
@@ -10,21 +10,22 @@ int main()
...
-    cur.extend_right("AAC"_dna4);                              // search the sequence "AAC"
-    seqan3::debug_stream << cur.path_label(genome) << '\n';    // outputs "AAC"
-    auto uni_it = cur.to_rev_cursor();                         // unidirectional cursor on the text "CAAGCAATTAAG"
-    seqan3::debug_stream << uni_it.path_label(genome) << '\n'; // outputs "CAA"
+    auto cur = index.cursor();                                      // create a cursor
+    cur.extend_right("AAC"_dna4);                                   // search the sequence "AAC"
+    seqan3::debug_stream << cur.path_label(genome) << '\n';         // outputs "AAC"
+    auto uni_it = cur.to_rev_cursor();                              // unidirectional cursor on the text "CAAGCAATTAAG"
+    auto genome_rev = genome | std::views::reverse;                 // create reverse text
+    seqan3::debug_stream << uni_it.path_label(genome_rev) << '\n';  // outputs "CAA"
```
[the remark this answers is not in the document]
Irallia: When I copy and paste the code, it prints:
```
Example to_rev_cursor()
AAC
CAA
GAAT
2
```
And that is, what it should do right?
If you remove the ```genome_rev```, than it prints:
```
Example to_rev_cursor()
AAC
GAA
GAAT
2
```
Because the reverse cursor is finding CAA at the end of the reversed text, but path_label was printing the end of the not-reversed text. (Path label used the wrong input.)

include/seqan3/search/fm_index/fm_index_cursor.hpp
Irallia: > Not sure why you subtract above and add again later. Can an underflow happen, i.e. `text_idx = -1`? Can `index->text_begin_rs.rank(loc + 1)` be `0`? Can you add `assert(text_idx + 1 != 0)`?
> Is the data layout of the concatenated text like this?:
> 
> > "$FIRSTSTRING#SECONDSTRING%THIRDSTRING"
> 
> So that `rank` returns always > 0, i.e. the number of delimiters is strictly positive? Do you know what happens when `loc + 1` is a delimiter? In other words what would happen if `location + 1` is `%` in my example? Do you know why we take `loc + 1`? Does this avoid being on the wrong delimiter and therefore giving back the wrong index of the current text?

Yes I can add `assert(text_idx + 1 != 0)`. The deliminiters should be the same, but yes I think the text looks like this.

I have made several attempts to omit the +1 and -1 or postpone it to make it clearer, but then tests always fail because we are running above or below an index.
The point is that ```index->text_begin_rs.rank(location);``` and ```index->text_begin_rs.rank(location + 1);``` can be equal or different.

If you have a look at these two tests, I changed rank to ```index->text_begin_rs.rank(location);```:
```shell
~/build/seqan3/unit_debug$ search/fm_index_cursor/fm_index_cursor_collection_test 
...
[ RUN      ] default_traits/fm_index_cursor_collection_test/0.query
-------------------
[ACGACG,TGCGATCGA]
location 0
old rank 1
rank 0
start_location 0
text_id 0
query_begin 0
result ACG
[       OK ] default_traits/fm_index_cursor_collection_test/0.query (109 ms)
...
[ RUN      ] default_traits/fm_index_cursor_collection_test/0.incomplete_alphabet
-------------------
[ATATAT,ATATAT]
location 1
old rank 1
rank 1
start_location 7
text_id 1
query_begin 18446744073709551610
result
/seqan3/test/unit/search/fm_index_cursor/fm_index_cursor_collection_test_template.hpp:319: Failure
Value of: std::ranges::equal(it.path_label(text), "T"_dna4)
  Actual: false
Expected: true
[  FAILED  ] ...
```
I have to admit, I still don't fully understand how rank works here, but everything works with (loc + 1). Maybe @smehringer  would like to add something?

test/snippet/search/bi_fm_index_cursor_to_rev_cursor.cpp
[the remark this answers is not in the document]
marehr: Ah I just executed the original snippet; but you changed it. And I think you are right :)

test/snippet/search/bi_fm_index_cursor_to_rev_cursor.cpp
```diff
@@ -10,21 +10,22 @@ int main()
...
     // cursor was to the right:
     // cur.cycle_back();
     // seqan3::debug_stream << cur.last_rank() << '\n';
 
-    uni_it.extend_right('G'_dna4);                             // search the sequence "CAAG"
-    seqan3::debug_stream << uni_it.path_label(genome) << '\n'; // outputs "CAAG"
-    seqan3::debug_stream << uni_it.last_rank() << '\n';        // outputs 2
-    uni_it.cycle_back();                                       // search the sequence "CAAT"
+    uni_it.extend_right('G'_dna4);                                  // search the sequence "CAAG"
+    seqan3::debug_stream << uni_it.path_label(genome) << '\n';      // outputs "GAAT"
```
marehr, line 28: I don't understand this outpur, shouldn't it be `GAAT` on the forward iterator?

include/seqan3/search/fm_index/fm_index_cursor.hpp
```diff
@@ -482,9 +482,22 @@ class fm_index_cursor
...
-        size_type const loc = offset() - index->index[node.lb];
-        size_type const query_begin = loc - index->text_begin_rs.rank(loc + 1) + 1; // Substract delimiters
-        return text | views::join | views::slice(query_begin, query_begin + query_length());
+        // Position of query in concatenated text.
+        size_type const location = offset() - index->index[node.lb];
+        // The rank represents the number of delimiters before position `location + 1`.
+        // Furthermore, this number represents the i-th text this (text) position originates from, since each delimiter
+        // corresponds to one text in the whole concatenated text from left-to-right.
+        size_type const rank = index->text_begin_rs.rank(location + 1);
+        // The sum of all text lengths prior to the i-th text is the start location of the i-th text in the concatenated
```
marehr, line 491:
```suggestion

        // The sum of all text lengths prior to the i-th text is the start location of the i-th text in the concatenated
```

include/seqan3/search/fm_index/fm_index_cursor.hpp
```diff
@@ -482,9 +482,22 @@ class fm_index_cursor
...
+        // sequence.
+        size_type const start_location = index->text_begin_ss.select(rank);
+        assert(rank > 0);
+        size_type const text_id = rank - 1;
+        assert(text_id + 1 != 0);
+        // Substract lengths of previous sequences.
```
marehr, line 497:
```suggestion

        // Substract lengths of previous sequences.
```

include/seqan3/search/fm_index/fm_index_cursor.hpp
marehr: That's fine if it works for +1. We should maybe add the `+1` directly in the location variable.

I think that both the `offset` and the `index` are zero-based and that we double subtract the zero-position in the difference, so we need to add one to get the real location.

include/seqan3/search/fm_index/bi_fm_index_cursor.hpp
```diff
@@ -907,9 +907,21 @@ class bi_fm_index_cursor
...
+        size_type const rank = index->fwd_fm.text_begin_rs.rank(location + 1);
+        // The sum of all text lengths prior to the i-th text is the start location of the i-th text in the concatenated
+        // sequence.
+        size_type const start_location = index->fwd_fm.text_begin_ss.select(rank);
+        assert(rank > 0);
+        size_type const text_id = rank - 1;
+        // Substract lengths of previous sequences.
+        size_type const query_begin = location - start_location;
+        // Take subtext, slice query out of it <-my
+        return text[text_id] | views::slice(query_begin, query_begin + query_length());
```
marehr, line 924: Please put some spaces between the operations that belong together.

```suggestion
        // Position of query in concatenated text.
        size_type const location = offset() - index->fwd_fm.index[fwd_lb] + 1;
        // The rank represents the number of delimiters before position `location`.
        // Furthermore, this number represents the i-th text this (text) position originates from, since each delimiter
        // corresponds to one text in the whole concatenated text from left-to-right.
        size_type const rank = index->fwd_fm.text_begin_rs.rank(location);

        assert(rank > 0);

        // The sum of all text lengths prior to the i-th text is the start location of the i-th text in the concatenated
        // sequence.
        size_type const start_location = index->fwd_fm.text_begin_ss.select(rank);
        size_type const text_id = rank - 1;

        // Substract lengths of previous sequences.
        size_type const query_begin = location - start_location;

        // Take subtext, slice query out of it <-my
        return text[text_id] | views::slice(query_begin, query_begin + query_length());
```

include/seqan3/search/fm_index/fm_index_cursor.hpp
```diff
@@ -482,9 +482,22 @@ class fm_index_cursor
...
+        size_type const query_begin = location - start_location;
+        // Take subtext, slice query out of it <-my
+        return text[text_id] | views::slice(query_begin, query_begin + query_length());
```
marehr, line 500:
```suggestion
        // Position of query in concatenated text.
        size_type const location = offset() - index->index[node.lb] + 1;
        // The rank represents the number of delimiters before position `location`.
        // Furthermore, this number represents the i-th text this (text) position originates from, since each delimiter
        // corresponds to one text in the whole concatenated text from left-to-right.
        size_type const rank = index->text_begin_rs.rank(location);

        assert(rank > 0);

        // The sum of all text lengths prior to the i-th text is the start location of the i-th text in the concatenated
        // sequence.
        size_type const start_location = index->text_begin_ss.select(rank);
        size_type const text_id = rank - 1;

        // Substract lengths of previous sequences.
        size_type const query_begin = location - start_location;

        // Take subtext, slice query out of it <-my
        return text[text_id] | views::slice(query_begin, query_begin + query_length());
```

include/seqan3/search/fm_index/fm_index_cursor.hpp
smehringer: Ok some clarification (that I hope I understood correctly from the code):

@Irallia You comments are a bit misleading (as are the non-expressive variable names in the search code :woman_facepalming:). The variable `text_begin_ss` is the select-support-data structures of the variable `text_begin` which is a bitvector that **stores the begin positions of every text in a collection`. I think that's what confused you @marehr because you thought that the bitvector marks the position of the delimiters. Thus, The rank is always greater than 0 since the very first bit will always be a 1.  

Now to the mystery of the `+1` you talked about. You must know that the `rank at position i` in the SDSL (at least I am pretty sure) is defined as `the number of 1's up until, but not including, position i`. Which means that if there is a `1` at position `i` it is missed. Thus, the `location` is correctly computed but the rank query must inquire the rank at position one after `i` in order to resolve the correct text.

> I think that both the offset and the index are zero-based and that we double subtract the zero-position in the difference, so we need to add one to get the real location.

I am not exactly sure what you meant by this sentence. Maybe you can rephrase it in case you think there is still an issue.

include/seqan3/search/fm_index/bi_fm_index_cursor.hpp
marehr: (Memo to myself or others: this example was wrong, because 
`size_type const location = offset() - index->fwd_fm.index[fwd_lb] + 1;` would need a `size_type const query_begin = location - 1 - start_location;` Just the spacing was left from this suggestion.)

test/snippet/search/bi_fm_index_cursor_to_rev_cursor.cpp
marehr: This is still open.

smehringer:
@marehr Can you resolve your comments? (I assume they are resolved since you approved)

include/seqan3/search/fm_index/bi_fm_index_cursor.hpp
```diff
@@ -908,9 +908,24 @@ class bi_fm_index_cursor
...
 
-        size_type const loc = offset() - index->fwd_fm.index[fwd_lb];
-        size_type const query_begin = loc - index->fwd_fm.text_begin_rs.rank(loc + 1) + 1; // Substract delimiters
-        return text | views::join | views::slice(query_begin, query_begin + query_length());
+        // Position of query in concatenated text.
+        size_type const location = offset() - index->fwd_fm.index[fwd_lb];
+
+        // The rank represents the number of delimiters before position `location + 1`.
+        // Furthermore, this number represents the i-th text this (text) position originates from, since each delimiter
+        // corresponds to one text in the whole concatenated text from left-to-right.
```
smehringer, line 916: Your comment makes it seem like `index->fwd_fm.text_begin_rs.` stores the position of the delimiters.
But it stores the start position of the texts.
I think this makes it clearer: (Feel free to rephrase it further)
```suggestion
        // The rank represents the number of start positions of individual sequences/texts in the collection 
        // before position `location + 1` and thereby also the number of delimiters.
```

include/seqan3/search/fm_index/bi_fm_index_cursor.hpp
```diff
@@ -908,9 +908,24 @@ class bi_fm_index_cursor
...
+        size_type const rank = index->fwd_fm.text_begin_rs.rank(location + 1);
+        assert(rank > 0);
+        size_type const text_id = rank - 1;
+
+        // The sum of all text lengths prior to the i-th text is the start location of the i-th text in the concatenated
+        // sequence.
```
smehringer, line 922: Always state the important part, the result, first and then the cause, for a better understanding :)
And technically we are not computing the sum of lengths but `select` just asks "where is the position of the `rank`-th 1".
```suggestion
        // The start location of the `text_id`-th text in the sequence (position of the `rank`-th 1 in the bitvector).
```

include/seqan3/search/fm_index/fm_index_cursor.hpp
```diff
@@ -482,9 +482,25 @@ class fm_index_cursor
...
+        // Furthermore, this number represents the i-th text this (text) position originates from, since each delimiter
+        // corresponds to one text in the whole concatenated text from left-to-right.
+        size_type const rank = index->text_begin_rs.rank(location + 1);
+        assert(rank > 0);
+        size_type const text_id = rank - 1;
+
+        // The sum of all text lengths prior to the i-th text is the start location of the i-th text in the concatenated
+        // sequence.
+        size_type const start_location = index->text_begin_ss.select(rank);
+        assert(text_id + 1 != 0);
```
smehringer, line 498: Does this assert make sense if you are already testing `rank > 0`?

include/seqan3/search/fm_index/bi_fm_index_cursor.hpp
```diff
@@ -908,9 +908,24 @@ class bi_fm_index_cursor
...
+        size_type const start_location = index->fwd_fm.text_begin_ss.select(rank);
+        // Substract lengths of previous sequences.
+        size_type const query_begin = location - start_location;
+
+        // Take subtext, slice query out of it <-my
```
smehringer, line 927: what is `<-my` supposed to mean?

include/seqan3/search/fm_index/fm_index_cursor.hpp
marehr: yeah in can be omitted

include/seqan3/search/fm_index/bi_fm_index_cursor.hpp
marehr: I think this is worse. Sometime you have to paraphrase what the intend or effect is and not what it is technically doing.

smehringer: I think it is always easier to understand if the comment starts with "what is this" and then "why is it computed like this".
And for me it is actually confusing why the intend should be to compute the "sum of all text lengths"? when I actually just want to know "Where is this 1 that I computed in the rank query beforehand" which is my start position of the text.
Do you know rank and select queries? This is a very typical use case.

smehringer:
Ok since comments can be discussed for ages before everyone is happy with the wording, You can just leave it or change it or rewrite it yourself. Whatever you feel like :) 
Just additionally address the remaining assert and making two commits out of the four, then I'll merge it.

PR #1690, opened 2020-03-25
[DOC] io: DRY doc of shared content between alignment input and output.

smehringer:
These were hard to keep consistent

### DIFF
--- a/include/seqan3/io/alignment_file/all.hpp
+++ b/include/seqan3/io/alignment_file/all.hpp
@@ -22,7 +22,7 @@
  * many additional information. Well-known formats include the SAM/BAM format used to store read mapping data or the
  * BLAST format that stores the results of a query search against a data base.
  *
- * \note Take a look at our tutorial \ref tutorial_alignment_file for a walk through of how to read alignment files.
+ * \note For a step-by-step guide take a look at our tutorial: \ref tutorial_alignment_file.
  *
  * \copydetails seqan3::alignment_file_input::field_ids
  *
--- a/include/seqan3/io/alignment_file/all.hpp
+++ b/include/seqan3/io/alignment_file/all.hpp
@@ -15,6 +15,20 @@
 /*!\defgroup alignment_file Alignment File
  * \ingroup io
  * \brief Provides files and formats for handling alignment data.
+ *
+ * ### Introduction
+ *
+ * Alignment files are primarily used to store pairwise alignments of two biological sequences and often come with
+ * many additional information. Well-known formats include the SAM/BAM format used to store read mapping data or the
+ * BLAST format that stores the results of a query search against a data base.
+ *
+ * \note Take a look at our tutorial \ref tutorial_alignment_file for a walk through of how to read alignment files.
+ *
+ * \copydetails seqan3::alignment_file_input::field_ids
+ *
+ * All of these fields are retrieved by default (and in that order).
+ * Note that some of the fields are specific to the SAM format (e.g. seqan3::field::flag) while others are specific to
+ * BLAST format (e.g. seqan3::field::bit_score). Please see the corresponding formats for more details.
  */
 
 #pragma GCC diagnostic push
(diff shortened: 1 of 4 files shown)


include/seqan3/io/alignment_file/all.hpp
```diff
@@ -15,6 +15,20 @@
...
  * \ingroup io
  * \brief Provides files and formats for handling alignment data.
+ *
+ * ### Introduction
+ *
+ * Alignment files are primarily used to store pairwise alignments of two biological sequences and often come with
+ * many additional information. Well-known formats include the SAM/BAM format used to store read mapping data or the
+ * BLAST format that stores the results of a query search against a data base.
+ *
+ * \note Take a look at our tutorial \ref tutorial_alignment_file for a walk through of how to read alignment files.
```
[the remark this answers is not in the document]
smehringer: or `on`? let's rewrite this
```suggestion
 * \note For a step-by-step guide take a look at our tutorial: \ref tutorial_alignment_file.
```

include/seqan3/io/alignment_file/all.hpp
```diff
@@ -15,6 +15,20 @@
...
+ * many additional information. Well-known formats include the SAM/BAM format used to store read mapping data or the
+ * BLAST format that stores the results of a query search against a data base.
+ *
+ * \note For a step-by-step guide take a look at our tutorial: \ref tutorial_alignment_file.
+ *
+ * \copydetails seqan3::alignment_file_input::field_ids
+ *
+ * All of these fields are retrieved by default (and in that order).
+ * Note that some of the fields are specific to the SAM format (e.g. seqan3::field::flag) while others are specific to
+ * BLAST format (e.g. seqan3::field::bit_score). Please see the corresponding formats for more details.
```
rrahn: The weird thing about this is that we don't have blast yet. And I don't see this to happen very soon. So we have spoiled the records with something that does not give any benefit yet. So I am wondering if we should not remove those for now. And then when it becomes relevant add them, as to avoid to much confusion.

rrahn: Oh, unless of course someone wrote a structured binding expression for all 15 fields already? Oo

smehringer: hm it might be a little but it also shows that we have a plan for this feature :woman_shrugging: 
I don't think it hurts much, so I would avoid an API break, but if you like we can decide about his in the next DEV meeting, I don't care much.

include/seqan3/io/alignment_file/input.hpp
```diff
@@ -490,8 +459,29 @@ class alignment_file_input
...
+     *   8. seqan3::field::cigar
+     *   9. seqan3::field::mapq
+     *   10. seqan3::field::qual
+     *   11. seqan3::field::flag
+     *   12. seqan3::field::mate
+     *   13. seqan3::field::tags
+     *   14. seqan3::field::evalue
+     *   15. seqan3::field::bit_score
+     *
+     * There exists one more field for alignment files, the seqan3::field::header_ptr, but this field is mostly used
```
rrahn: why does this need to be a field in the record?

include/seqan3/io/alignment_file/all.hpp
rrahn: Well, I am rather not for an API break, but also would avoid documentation that points to presumably existing formats when they don't. As a user I would now go looking for the BLAST format and would find nothing. That is disappointing. So I would make the documentation here less generic and only talk about SAM/BAM for now.

include/seqan3/io/alignment_file/input.hpp
smehringer: Because if I do `fin | fout` (range interface) I need to pass the header information somehow

smehringer:
@rrahn Since this PR only reduces redundant code by using `\copydetails` can this PR be merged, even though we will discuss all your open questions regarding the general alignment file design?

Issue #1591, opened 2020-02-13
Sequence file input validator does not accept gzipped files


sarahet:
When defining an input file with a validator for sequence input files the following way:

```
parser.add_option(options.queryFile, 'q', "query", "Query sequences.",
seqan3::option_spec::REQUIRED,
seqan3::input_file_validator<seqan3::sequence_file_input<>>{});
```
it does not accept ```.gz``` files but only ```[embl,fasta,fa,fna,ffn,faa,frn,fastq,fq,genbank,gb,gbk,sam]```

Shouldn't this be possible as most sequence input files are actually compressed?

rrahn:
Hey @sarahet thanks for the report.

This is currently indeed not supported and would be a feature. It does need some thoughts though to do it correctly. We will discuss this and see how we can put this into out road map.

sarahet:
Great, thanks. A related issue would be that input validators generally do not have the possibility to allow for file endings hat include multiple dots, like ```.fastq.gz```. Could that be included as well or is that rather not planned? Related https://github.com/seqan/lambda/issues/161

rrahn:
yes, I guess that is somewhat related. is the multiple dot version restricted to 2 dots assuming one valid extension and one compression extension? Or are more compex cases possible?

sarahet:
So far I can only think of one extension and a compression extension, but of course there could be other use cases, that I currently don't have in mind ..

smehringer:
Is it a use-case to compress you file multiple times, e.g. `fasta.gz.bz2`? (I don't even know if this makes sense, just wondering)

rrahn:
I think `.fasta.tar.bz` etc. could be. Not that we would support that at the moment but maybe interesting for the future

PR #1691, opened 2020-03-25
[MISC] argument parser: Move get_help_message to file_validator_base.

smehringer:
Just DRY out some code

### DIFF
--- a/include/seqan3/argument_parser/validators.hpp
+++ b/include/seqan3/argument_parser/validators.hpp
@@ -444,10 +444,13 @@ protected:
         file_guard.remove();
     }
 
-    //!\brief Returns the information of valid file formats.
+    //!\brief Returns the information of valid file extensions.
     std::string valid_extensions_help_page_message() const
     {
-        return detail::to_string("Valid file formats are: [", extensions | views::join(std::string{", "}), "].");
+        if (extensions.empty())
+            return "";
+        else
+            return detail::to_string("Valid file extensions are: [", extensions | views::join(std::string{", "}), "].");
     }
 
     //!\brief Stores the extensions.
--- a/include/seqan3/argument_parser/validators.hpp
+++ b/include/seqan3/argument_parser/validators.hpp
@@ -24,7 +24,6 @@
 #include <seqan3/io/detail/misc.hpp>
 #include <seqan3/io/detail/safe_filesystem_entry.hpp>
 #include <seqan3/range/container/concept.hpp>
-#include <seqan3/range/views/join.hpp>
 #include <seqan3/std/algorithm>
 #include <seqan3/std/concepts>
 #include <seqan3/std/filesystem>
@@ -448,9 +447,17 @@ protected:
     std::string valid_extensions_help_page_message() const
     {
         if (extensions.empty())
+        {
             return "";
+        }
         else
-            return detail::to_string("Valid file extensions are: [", extensions | views::join(std::string{", "}), "].");
+        {
+            std::string msg{" Valid file extensions are: [" + extensions[0]};
+            for (auto it = extensions.begin() + 1; it != extensions.end(); ++it)
+                msg += (", " + *it);
+            msg += "].";
+            return msg;
+        }
     }
 
     //!\brief Stores the extensions.
@@ -569,7 +576,7 @@ public:
     //!\brief Returns a message that can be appended to the (positional) options help page info.
     std::string get_help_page_message() const
     {
-        return "The input file must exist and read permissions must be granted. " +
+        return "The input file must exist and read permissions must be granted." +
                valid_extensions_help_page_message();
     }
 };
@@ -671,7 +678,7 @@ public:
     //!\brief Returns a message that can be appended to the (positional) options help page info.
     std::string get_help_page_message() const
     {
-        return "The output file must not exist already and write permissions must be granted. " +
+        return "The output file must not exist already and write permissions must be granted." +
                valid_extensions_help_page_message();
     }
 };
--- a/include/seqan3/argument_parser/validators.hpp
+++ b/include/seqan3/argument_parser/validators.hpp
@@ -24,6 +24,7 @@
 #include <seqan3/io/detail/misc.hpp>
 #include <seqan3/io/detail/safe_filesystem_entry.hpp>
 #include <seqan3/range/container/concept.hpp>
+#include <seqan3/range/views/join.hpp>
 #include <seqan3/std/algorithm>
 #include <seqan3/std/concepts>
 #include <seqan3/std/filesystem>
@@ -447,17 +448,9 @@ protected:
     std::string valid_extensions_help_page_message() const
     {
         if (extensions.empty())
-        {
             return "";
-        }
         else
-        {
-            std::string msg{" Valid file extensions are: [" + extensions[0]};
-            for (auto it = extensions.begin() + 1; it != extensions.end(); ++it)
-                msg += (", " + *it);
-            msg += "].";
-            return msg;
-        }
+            return detail::to_string(" Valid file extensions are: [", extensions | views::join(std::string{", "}), "].");
     }
 
     //!\brief Stores the extensions.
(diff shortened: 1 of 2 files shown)


smehringer:
@marehr polite ping

include/seqan3/argument_parser/validators.hpp
```diff
@@ -357,8 +357,15 @@ class file_validator_base
          std::for_each(v.begin(), v.end(), [&] (auto cmp) { this->operator()(cmp); });
     }
 
-protected:
+    //!\brief Returns a message that can be appended to the (positional) options help page info.
+    std::string get_help_page_message() const
```
marehr: why not protected?

include/seqan3/argument_parser/validators.hpp
```diff
@@ -557,14 +564,6 @@ class input_file_validator : public file_validator_base
             std::rethrow_exception(std::current_exception());
         }
     }
-
-    //!\brief Returns a message that can be appended to the (positional) options help page info.
-    std::string get_help_page_message() const
-    {
-        return detail::to_string("Valid input file formats: [",
```
marehr: This message is now different to the one above

test/unit/argument_parser/format_parse_validators_test.cpp
```diff
@@ -149,7 +149,7 @@ TEST(validator_test, input_file)
                                "==========="
                                "POSITIONAL ARGUMENTS"
                                "    ARGUMENT-1 (std::filesystem::path)"
-                               "          desc Valid input file formats: [fa, sam, fasta]"} +
```
marehr: is this now an input or an output file? How can you tell as a user?

smehringer:
> I'm not sure if omitting text is better for the user. I liked the output before, what are the benefits for the user?

You are right, I didn't notice that there is no mention of the option being an input or output file. I thought there was some more, also regarding the writability stuff. The benefit is, that it poses less of a chance for inconsistencies. I worked on fixing #1591 and adding the compression formats to the help page to each of the two (almost identical...) help page messages called for a DRY version to me.

> I'm okay with putting a common function in the base to concat the extensions as a function and use it in the get_help_message.

I'll change it accordingly, and also take a look at the writability

smehringer:
:smile: any suggestion? Usually, I would ask @eseiler...

include/seqan3/argument_parser/validators.hpp
```diff
@@ -561,9 +566,8 @@ class input_file_validator : public file_validator_base
     //!\brief Returns a message that can be appended to the (positional) options help page info.
     std::string get_help_page_message() const
     {
-        return detail::to_string("Valid input file formats: [",
-                                 extensions | views::join(std::string{", "}),
-                                 "]");
+        return "The input file must exist and read permissions must be granted. " +
```
rrahn, line 569: I am not sure. Is that really necessary with the permissions? That kind of clutters the output. It is clear that you need to have permissions. If not you will get an exception saying that you don't have permissions. 
I think if the description reads: `An input file with a valid extension: [ext1, ext2, ...]`
The other thing is `An output file with a valid extension: [ext1, ext2, ...]`
Then the code is 
```cpp
return valid_extensions_help_page_message("input file");
```
And the help page message assembles the string.
Then also it checks if valid extensions is empty and omits the part with the valid extensions. Also note, that we should talk about extensions and not formats.

smehringer: > I am not sure. Is that really necessary with the permissions? That kind of clutters the output. It is clear that you need to have permissions.

Well, although "it is obvious that you need them", it is a feature we provide to check for this and I thought this documentation helps to indicate this?

And whatever we decide about the permissions, I think it is important to note that the output file must not exist. This is not the usual behaviour by default as @joergi-w mentioned in his recent issue.

smehringer:
I just rebased because there were merge conflicts

include/seqan3/argument_parser/validators.hpp
rrahn: > And whatever we decide about the permissions, I think it is important to note that the output file must not exist. This is not the usual behaviour by default as @joergi-w mentioned in his recent issue.

Mhmm, I think it is not necessary but I have no strong feelings. But the other points should be adresed. `valid formats` ➡️ `valid extension` and if the extension set is empty this string should not be displayed at all.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -444,10 +444,13 @@ class file_validator_base
...
 
-    //!\brief Returns the information of valid file formats.
+    //!\brief Returns the information of valid file extensions.
     std::string valid_extensions_help_page_message() const
     {
-        return detail::to_string("Valid file formats are: [", extensions | views::join(std::string{", "}), "].");
+        if (extensions.empty())
+            return "";
+        else
+            return detail::to_string("Valid file extensions are: [", extensions | views::join(std::string{", "}), "].");
```
rrahn, line 453:
```suggestion
            return detail::to_string("Valid file extensions are: [", extensions | std::views::join(std::string{", "}), "].");
```

And check that there is no seqan3 views join include.
Regarding `detail::to_string`. Isn't this part of the dependencies @marehr tried to remove from the argument parser? This could be written as `str1 + str2 + str3`

test/unit/argument_parser/format_parse_validators_test.cpp
```diff
@@ -191,9 +191,11 @@ TEST(validator_test, input_file_ext_from_file)
 {
     // Give as a template argument the seqan3 file type to get all valid extensions for this file.
     seqan3::input_file_validator<dummy_file> validator{};
-
     EXPECT_EQ(validator.get_help_page_message(), "The input file must exist and read permissions must be granted. "
-                                                 "Valid file formats are: [fa, fasta, sam, bam].");
+                                                 "Valid file extensions are: [fa, fasta, sam, bam].");
+
+    seqan3::input_file_validator validator2{};
+    EXPECT_EQ(validator2.get_help_page_message(), "The input file must exist and read permissions must be granted. ");
```
rrahn, line 198:
```suggestion
    EXPECT_EQ(validator2.get_help_page_message(), "The input file must exist and read permissions must be granted.");
```
?

test/unit/argument_parser/format_parse_validators_test.cpp
```diff
@@ -297,9 +299,12 @@ TEST(validator_test, output_file_ext_from_file)
...
     // Give as a template argument the seqan3 file type to get all valid extensions for this file.
     seqan3::output_file_validator<dummy_file> validator{};
+    EXPECT_EQ(validator.get_help_page_message(), "The output file must not exist already and write permissions must "
+                                                 "be granted. Valid file extensions are: [fa, fasta, sam, bam].");
 
-    EXPECT_EQ(validator.get_help_page_message(), "The output file must not exist already and write permissions must be granted. "
-                                                 "Valid file formats are: [fa, fasta, sam, bam].");
+    seqan3::output_file_validator validator2{};
+    EXPECT_EQ(validator2.get_help_page_message(), "The output file must not exist already and write permissions must "
+                                                  "be granted. ");
```
rrahn, line 307:
```suggestion
                                                  "be granted.");
```
?

include/seqan3/argument_parser/validators.hpp
smehringer: The `detail::to_string` is still present in many parts of the code. He sais this one is an open ToDo and he might reimplement the function so using it here won't make a difference?

marehr: This one is tough, because the `seqan3::views::join` isn't the same as the `std::view::join`.

I had to warkaround it, but it is okay for me to use `seqan3::views::join` here. Also keep the `detail::to_string`.

In this case it is fine as it is.

smehringer:
@rrahn I still have no rights to restart jobs on Jenkins. The error is unrelated. Maybe you can restart Jenkins.

Issue #1692, opened 2020-03-25
Call alignment algorithm with a callback function instead of returning the results


rrahn:
This enables easier control when the alignment supports calling a user defined callback instead of returning a lazy range. It also removes the dependencies of the algorithms to call the meta function to configure the alignment result at different places causing potential hard to read compiler issues if changes are not propagated to all invocations.

### Edit 09.04.2020

Resolution after discussion with @marehr about this changes.
The problem with infering the type from the callback in the alignment algorithm is in general a little complicated. Still it is a good design to have the alignment algorithm decide when to call the user defined callback and possibly react on a given return type. Like terminating the algorithm. The alignment result type should, however, be accessible through the alignment configuration traits. To do this, the algorithm configurator will add a detail config element during the configuration process containing the alignment result type. This can be then accessed via the alignment configuration traits.
The execturo two way class is explicitly instantiated with the result type and does not need to infer the result type from the algorithm stored inside of the std::function object in a complicated template expression. This makes the two_executor also more generic to be reused for other algorithms like the search.

Tasks:

- [x] add a detail config element for the alignment result type
- [x] update the alignment configuration traits class to add a type for the result which will resolve to empty_type if it is not yet known (just in case the config was not added yet).
- [x] update the alignment_executor_two_way to get the buffer value type explicitly as template argument
- [x] Let the algorithm function operator get an additional parameter for the callback and call this instead of returning the results

PR #1693, opened 2020-03-25
[MISC] Move is_constexpr into type_traits basic.

rrahn:
There are some cyclic dependencies when function pulls in some other headers from core type traits.
Since the macro works in general for all expressions it is better suited in the basic header of type_traits.

Part of #1692

### DIFF
--- a/test/unit/core/type_traits/function_test.cpp
+++ b/test/unit/core/type_traits/function_test.cpp
@@ -7,7 +7,7 @@
 
 #include <gtest/gtest.h>
 
-#include <seqan3/core/type_traits/function.hpp>
+#include <seqan3/core/type_traits/basic.hpp>
 
 constexpr
 int    constexpr_nonvoid_free_fun(int i) { return i; }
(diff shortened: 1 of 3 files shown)


rrahn:
@wvandertoorn thansk for the review. Can you please also approve the PR, because we need 2 approvals in order to merge it. Thank you

test/unit/core/type_traits/function_test.cpp
```diff
@@ -7,7 +7,7 @@
 
 #include <gtest/gtest.h>
 
-#include <seqan3/core/type_traits/function.hpp>
+#include <seqan3/core/type_traits/basic.hpp>
```
smehringer: It's weird that the test called `function_test` is not including `seqan3/core/type_traits/function.hpp`
It currently just tests the `SEQAN3_IS_CONSTEXPR` macro? So maybe the test should be renamed?

rrahn: This is fixed in upcoming PRs. Otherwise I would have deleted this header to just reintroduce it in the next PR. That seemd kind of worthless to me

smehringer: ok

PR #1697, opened 2020-03-27
[FIX] Fix constexpr and noexcept usage in the gap_decorator

smehringer:
`constexpr`:
 Since the gap_decorator contains a set which cannot be constexpr
    constructed, the gap_decorator itslef can also not be constexpr
    constructed. If it cannot be constructed, none of the member functions
    can be constexpr. And since the gap_decorator_iterator stores a pointer
    to the non-constexpr gap_decorator it cannot be constexpr itself. 

`noexcept`:
```
gap_decorator::size() is only noexcept iff the underlying view is noexcept.
gap_decorator::operator[]() is only noexcept iff dereferencing the gap_decorator_iterator is noexcept.
gap_decorator::operator==() isn't noexcept because gap_decorator_iterator::operator++() isn't.
gap_decorator::operator!=() isn't noexcept because gap_decorator::operator==() isn't.
gap_decorator::operator<() isn't noexcept because gap_decorator_iterator::operator++() isn't.
gap_decorator::operator<=() isn't noexcept because gap_decorator_iterator::operator++() isn't.
gap_decorator::operator>=() isn't noexcept because gap_decorator::operator<() isn't.
gap_decorator::operator>() isn't noexcept because gap_decorator::operator<=() isn't.

gap_decorator_iterator::jump() isn't noexcept because std::set::lower_bound isn't noexcept.
Constructing the gap_decorator_iterator from host and position isn't noexcept because gap_decorator_iterator::jump() isn't.
gap_decorator_iterator::operator++() isn't noexcept because std::set::iterator::operator++() isn\t documented noexcept so we cannot be sure.
gap_decorator_iterator::operator++(int) isn't noexcept because gap_decorator_iterator::operator++() isn't.
gap_decorator_iterator::operator*() is only noexcept iff operator[]() of the underlying view is noexcept.
...and so on
```

@seqan/core Am I wrong with something?

### DIFF
--- a/include/seqan3/range/decorator/gap_decorator.hpp
+++ b/include/seqan3/range/decorator/gap_decorator.hpp
@@ -122,16 +122,11 @@ public:
     //!\brief Default constructor. Attention: all operations on a solely default constructed decorator,
     //!       except assigning a new range, are UB.
     gap_decorator() = default;
-    //!\brief Copy constructor.
-    gap_decorator(gap_decorator const &) = default;
-    //!\brief Copy construction via assignment.
-    gap_decorator & operator=(gap_decorator const &) = default;
-    //!\brief Move constructor.
-    gap_decorator(gap_decorator && rhs) = default;
-    //!\brief Move assignment.
-    gap_decorator & operator=(gap_decorator && rhs) = default;
-    //!\brief Use default deconstructor.
-    ~gap_decorator() = default;
+    gap_decorator(gap_decorator const &) = default; //!< Defaulted.
+    gap_decorator & operator=(gap_decorator const &) = default; //!< Defaulted.
+    gap_decorator(gap_decorator && rhs) = default; //!< Defaulted.
+    gap_decorator & operator=(gap_decorator && rhs) = default; //!< Defaulted.
+    ~gap_decorator() = default; //!< Defaulted.
 
     //!\brief Construct with the ungapped range type.
     template <typename other_range_t>
@@ -153,9 +148,9 @@ public:
      *
      * ### Exceptions
      *
-     * No-throw guarantee.
+     * Strong exception guarantee.
      */
-    size_type size() const noexcept
+    size_type size() const
     {
         if (anchors.size())
             return anchors.rbegin()->second + ungapped_view.size();
@@ -394,7 +389,7 @@ public:
      * \f$O(\log k)\f$ where \f$k\f$ is the number of gaps.
      *
      */
-    reference operator[](size_type const i) const noexcept
+    reference operator[](size_type const i) const
     {
         return *iterator{*this, i};
     }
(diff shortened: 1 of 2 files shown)


rrahn:
I summarised my take from the noexcept discussion under seqan/product_backlog#45.

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -152,7 +150,7 @@ class gap_decorator
      *
      * No-throw guarantee.
      */
-    size_type size() const noexcept
+    size_type size() const noexcept(noexcept(std::declval<ungapped_view_type>().size()))
```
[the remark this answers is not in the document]
smehringer: I removed it as discussed

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -391,7 +389,7 @@ class gap_decorator
      * \f$O(\log k)\f$ where \f$k\f$ is the number of gaps.
      *
      */
-    constexpr reference operator[](size_type const i) const noexcept
+    reference operator[](size_type const i) const
```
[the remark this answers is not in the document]
smehringer: As discussed, gap_decorator cannot be `constexpr` in the foreseeable future because of `std::set` so I left the code as-is.

include/seqan3/range/decorator/gap_decorator.hpp
```diff
@@ -507,7 +505,7 @@ class gap_decorator
      * ### Exceptions
      * No-throw guarantee.
      */
-    constexpr size_type gap_length(set_iterator_type it) const noexcept
+    size_type gap_length(set_iterator_type it) const noexcept
```
rrahn, line 508: and this is now guaranteed nothrow?

smehringer: I thought so... the code is
```cpp
return (it == anchors.begin()) ? it->second : it->second - (*std::prev(it)).second;
```
`anchors.begin()` is documented as `noexcept` and `std::prev` is constexpr so it is also noecept because constexpr function cannot throw. 
The iterator of the set is comparing and dereferencing a pair of two integers, so since working with PODs this should also be safe?

marehr: constexpr modifier does not imply noexcept when the execution is done at runtime

https://godbolt.org/z/o1FBYD

smehringer: Oh, very interesting. Then I will remove the noexcept here too!

smehringer:
@rrahn polite ping

rrahn:
lgtm now. Please cleanup the commit history or I can squash commit it. Whatever you prefer, just let me know.

Issue #1549, opened 2020-01-29
Deprecate `type_traits/iterator.hpp`


marehr:
- [ ] Remove tests if they only test standard behaviour.
- [ ] These occurrences need to be deprecated:

https://github.com/seqan/seqan3/blob/d2cbbf039d3c97e07cbf81bfb47e7b732a6e5210/include/seqan3/core/type_traits/iterator.hpp#L37-L42

https://github.com/seqan/seqan3/blob/d2cbbf039d3c97e07cbf81bfb47e7b732a6e5210/include/seqan3/core/type_traits/iterator.hpp#L54-L59

https://github.com/seqan/seqan3/blob/d2cbbf039d3c97e07cbf81bfb47e7b732a6e5210/include/seqan3/core/type_traits/iterator.hpp#L71-L76

https://github.com/seqan/seqan3/blob/d2cbbf039d3c97e07cbf81bfb47e7b732a6e5210/include/seqan3/core/type_traits/iterator.hpp#L94-L99

https://github.com/seqan/seqan3/blob/d2cbbf039d3c97e07cbf81bfb47e7b732a6e5210/include/seqan3/core/type_traits/iterator.hpp#L111-L116

---

Do the following for each struct `value_type`, `reference`, `rvalue_reference`, `difference_type` (each of those changes should be a different commit and maybe also a different PR):

* deprecate it, for example here: https://github.com/seqan/seqan3/blob/7df9e8389a435a61fd903697ec352dc9c025123a/include/seqan3/core/type_traits/pre.hpp#L41

  add the `SEQAN3_DEPRECATED_310` :

   ```c++
  template <typename t>
  struct SEQAN3_DEPRECATED_310 value_type;
  ```
* build all test cases (unit tests, performance tests, snippets, and so on) and see which test cases throw a deprecation warning.
* replace all occurrences of e.g. `seqan3::value_type<...>` with `std::iter_value_t<...>`
* repeat until everything builds and passes. 
* Be aware that `seqan3::value_type<...>` is also defined on ranges, so a simple search and replace will not work. But, you can search for all occurrences and figure out if it is used in a range context or a iterator context. Replace it for the `iterator context`.
* You are most likely nearly done after this.
* As a final step try to remove the complete struct (e.g. `seqan3::value_type<...>`) and see if everything builds (don't commit this, we will remove that later, because we can search for SEQAN3_DEPRECATED_310)
* if so that means that you successfully replaced all occurrences.

This will also fix #776.

[...]

PR #1698, opened 2020-03-31
Deprecate type_traits/iterator difference_type

wvandertoorn:
Resolves part of #1549 

Replaces occurrences of ```seqan3::difference_type<...>``` with ```std::iter_difference_t ``` or ```std::ranges::range_difference_t```.

### DIFF
--- a/include/seqan3/core/type_traits/iterator.hpp
+++ b/include/seqan3/core/type_traits/iterator.hpp
@@ -112,7 +112,7 @@ template <std::weakly_incrementable it_t>
 struct size_type<it_t>
 {
     //!\brief Return the member type as return type.
-    using type = std::make_unsigned_t<difference_type_t<it_t>>;
+    using type = std::make_unsigned_t<std::iter_difference_t<it_t>>;
 };
 
 // see specialisation for ranges in core/type_traits/range.hpp
--- a/include/seqan3/core/type_traits/iterator.hpp
+++ b/include/seqan3/core/type_traits/iterator.hpp
@@ -92,7 +92,7 @@ struct rvalue_reference<it_t>
  * \tparam it_t The type you wish to query; must model std::weakly_incrementable.
  */
 template <std::weakly_incrementable it_t>
-struct SEQAN3_DEPRECATED_310 difference_type<it_t>
+struct difference_type<it_t>
 {
     //!\brief Return the member type as return type.
     using type = std::iter_difference_t<it_t>;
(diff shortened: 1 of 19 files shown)


rrahn:
If this only partially fixes #1549 then you should remove the `resolves` identifier or change it to `resolves part of #1549`, otherwise the issue will be closed when the pr gets merged. But I guess this is not your intention?

wvandertoorn:
Force-pushed with cleaned up git history

marehr:
@wvandertoorn Just to give you a heads up, we are currently discussing what the best strategy is to deprecate in this case.

marehr:
In the meantime, I would suggest you to delete the deprecation in this PR and do the actual deprecation later in a new PR. 

So that we don't stall you :)

include/seqan3/core/type_traits/pre.hpp
```diff
@@ -156,7 +156,7 @@ using const_reference_t = typename const_reference<t>::type;
  * There is a shortcut for this transformation trait: seqan3::difference_type_t
  */
 template <typename t>
-struct difference_type;
+struct SEQAN3_DEPRECATED_310 difference_type;
```
marehr, line 159:
```suggestion
struct difference_type;
```

include/seqan3/core/type_traits/range.hpp
```diff
@@ -127,7 +127,7 @@ template <std::ranges::range rng_t>
 //!\cond
     requires !std::input_or_output_iterator<rng_t>
 //!\endcond
-struct difference_type<rng_t>
+struct SEQAN3_DEPRECATED_310 difference_type<rng_t>
```
marehr, line 130:
```suggestion
struct difference_type<rng_t>
```

include/seqan3/core/type_traits/pre.hpp
marehr: let's do this later. :)

include/seqan3/core/type_traits/iterator.hpp
```diff
@@ -92,7 +92,7 @@ struct rvalue_reference<it_t>
  * \tparam it_t The type you wish to query; must model std::weakly_incrementable.
  */
 template <std::weakly_incrementable it_t>
-struct difference_type<it_t>
+struct SEQAN3_DEPRECATED_310 difference_type<it_t>
```
marehr, line 95: @smehringer  This has absolutely no effect. Good catch!

@wvandertoorn  can you remove this? Thank you!

smehringer: But we want to deprecate this? Why remove it?

marehr: because
* this PR makes only the changes and the deprecation comes in a later PR (we discussed that in the strategy meeting)
* this deprecation is only on the partial specialisation and it has no effect what-so-ever (attributes are ignored on partial specialisations). The deprecation must happen at the **main** struct if we want to deprecate.

smehringer: ah right. thanks!

PR #1699, opened 2020-03-31
[FIX] kmer_hash works for too small ranges

### LINKED ISSUE #1643: view_kmer_hash does not handle too small ranges properly
## Platform

  - SeqAn version: 9a465083bfcc12dc35bf395a690294c4db85ccad
  - Operating system: `Linux <pc-name> 4.4.0-18362-Microsoft #476-Microsoft Fri Nov 01 16:53:00 PST 2019 x86_64 x86_64 x86_64 GNU/Linux` (WSL)
  - Compiler: `g++-9 (Ubuntu 9.2.1-17ubuntu1~18.04.1) 9.2.1 20191102`

### How to repeat the problem

```cpp
#include <seqan3/range/views/kmer_hash.hpp>

int main()
{
    std::vector<seqan3::dna4> sequence{"ACGATCGATCGTAGCTACTGAGC"_dna4};

    seqan3::debug_stream << (sequence | seqan3::views::kmer_hash(seqan3::ungapped{23u})) << '\n'; // OK [6829917194121]
    seqan3::debug_stream << (sequence | seqan3::views::kmer_hash(seqan3::ungapped{24u})) << '\n'; // OK []
    seqan3::debug_stream << (sequence | seqan3::views::kmer_hash(seqan3::ungapped{25u})) << '\n'; // SegFault [0,0,...]
}
```

### Expected behaviour

Do not SegFault and print `[]`

### Actual behaviour

Does SegFault and does not print `[]`

MitraDarja:
Fixes #1643

### DIFF
--- a/include/seqan3/range/views/kmer_hash.hpp
+++ b/include/seqan3/range/views/kmer_hash.hpp
@@ -115,7 +115,7 @@ public:
      */
     auto begin() noexcept
     {
-        return shape_iterator<urng_t>{std::ranges::begin(urange), shape_};
+        return shape_iterator<urng_t>{std::ranges::begin(urange), std::ranges::end(urange), shape_};
     }
 
     //!\copydoc begin()
@@ -124,7 +124,7 @@ public:
         requires const_iterable_range<urng_t>
     //!\endcond
     {
-        return shape_iterator<urng_t const>{std::ranges::begin(urange), shape_};
+        return shape_iterator<urng_t const>{std::ranges::begin(urange), std::ranges::end(urange), shape_};
     }
 
     //!\copydoc begin()
@@ -251,14 +251,17 @@ public:
     *
     * Linear in size of shape.
     */
-    shape_iterator(it_t it_start, shape s_) :
-        shape_{s_}, text_left{it_start}, text_right{it_start}
+    shape_iterator(it_t it_start, sentinel_t it_end, shape s_) :
+        shape_{s_}, text_left{it_start}, text_right{std::ranges::next(it_start, s_.size(), it_end)}
     {
         assert(std::ranges::size(shape_) > 0);
 
-        roll_factor = pow(sigma, static_cast<size_t>(std::ranges::size(shape_) - 1));
+        if (shape_.size() <= std::ranges::distance(text_left, text_right))
+        {
+            roll_factor = pow(sigma, static_cast<size_t>(std::ranges::size(shape_) - 1));
 
-        hash_full();
+            hash_full();
+        }
     }
     //!\}
 
--- a/test/unit/range/views/view_kmer_hash_test.cpp
+++ b/test/unit/range/views/view_kmer_hash_test.cpp
@@ -30,8 +30,6 @@ protected:
 
     static constexpr auto ungapped_view = seqan3::views::kmer_hash(seqan3::ungapped{3});
     static constexpr auto gapped_view = seqan3::views::kmer_hash(0b101_shape);
-    static constexpr auto ungapped_view2 = seqan3::views::kmer_hash(seqan3::ungapped{4});
-    static constexpr auto gapped_view2 = seqan3::views::kmer_hash(0b1001_shape);
 
     std::vector<seqan3::dna4> text1{"AAAAA"_dna4};
     std::vector<seqan3::dna4> const ctext1{"AAAAA"_dna4};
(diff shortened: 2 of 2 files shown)


rrahn:
Can you please also add an explicit test for the issue, which for example checks that the size is 0 or the view is empty when the shape is bigger than the underlying text. Right now it is hard to follow how the offending behaviour is actually tested and thus fixed.

MitraDarja:
@rrahn 

I added tests lines like this one:
```
EXPECT_EQ(ungapped3, ctext3 | ungapped_view2 | seqan3::views::to<result_t>);
```
Or do you prefer an extra TEST_F() for this?

include/seqan3/range/views/kmer_hash.hpp
```diff
@@ -115,7 +115,7 @@ class kmer_hash_view : public std::ranges::view_interface<kmer_hash_view<urng_t>
      */
     auto begin() noexcept
     {
-        return shape_iterator<urng_t>{std::ranges::begin(urange), shape_};
+        return shape_iterator<urng_t>{std::ranges::begin(urange), std::ranges::end(urange), shape_};
```
smehringer: I think it would require fewer changes if you just check `std::distance` of `std::ranges::begin(urange), std::ranges::end(urange)` here and `return std::ranges::end(urange)` if the requirement is not met instead of passing an extra argument to the constructor?

rrahn: No, because that would be a different type than the shape_iterator, which first is wrong and second not possible due to different types in the return expression deduced at runtime. The iterator must know whether it is at the end or not, so that `size()`  works correctly. For this it has to set the `text_right` iterator correctly (in this case to the end of the underlying range). So, you can't do it without an extra argument to the constructor.

rrahn:
> I added tests lines like this one:

Yes, but it is better to also add a particular test case with the issue number as name and a link to the ticket so that we can track that this was an issue. Sometimes, this can make tracking regresssion bugs much easier.
I also looked into the test and it was really difficult for me to see in this structure what the expected beghavior really is. Something like `EXPECT_EQ(view.size(), 0u)`, or `EXPECT_TRUE(view.empty())` are really beneficial to understand what should happen and also what breaks when it breaks.

include/seqan3/range/views/kmer_hash.hpp
```diff
@@ -251,14 +251,17 @@ class kmer_hash_view<urng_t>::shape_iterator
...
     */
-    shape_iterator(it_t it_start, shape s_) :
-        shape_{s_}, text_left{it_start}, text_right{it_start}
+    shape_iterator(it_t it_start,  sentinel_t it_end, shape s_) :
+        shape_{s_}, text_left{it_start}, text_right{std::ranges::next(it_start, s_.size(), it_end)}
     {
         assert(std::ranges::size(shape_) > 0);
 
-        roll_factor = pow(sigma, static_cast<size_t>(std::ranges::size(shape_) - 1));
+        if (shape_.size() <= std::distance(text_left, text_right))
```
rrahn, line 259:
```suggestion
        if (shape_.size() <= std::ranges::distance(text_left, text_right))
```
Beacuse it is the constrained version.

include/seqan3/range/views/kmer_hash.hpp
```diff
@@ -251,14 +251,17 @@ class kmer_hash_view<urng_t>::shape_iterator
     *
     * Linear in size of shape.
     */
-    shape_iterator(it_t it_start, shape s_) :
-        shape_{s_}, text_left{it_start}, text_right{it_start}
+    shape_iterator(it_t it_start,  sentinel_t it_end, shape s_) :
```
rrahn, line 254:
```suggestion
    shape_iterator(it_t it_start, sentinel_t it_end, shape s_) :
```

test/unit/range/views/view_kmer_hash_test.cpp
```diff
@@ -187,3 +192,12 @@ TEST_F(kmer_hash_test, issue1614)
     EXPECT_EQ(sequence | seqan3::views::kmer_hash(seqan3::ungapped{25}) | seqan3::views::to<std::vector<size_t>>,
               seqan3::views::repeat_n(298023223876953124, 26) | seqan3::views::to<std::vector<size_t>>);
 }
+
+// https://github.com/seqan/seqan3/issues/1643
+TEST_F(kmer_hash_test, issue1643)
+{
+    EXPECT_EQ(ungapped3, text3 | ungapped_view2 | seqan3::views::to<result_t>);
```
rrahn, line 199: It is really hard to understand what the test is about. So it becomes not clear that ungapped3 is empty and that the result should be that the resulting range is empty. Also it is not shown how the offending code from the referenced issue is fixed, so the reported use case should be used.
```cpp
    seqan3::dna4_vector text_23_elements{"ACGATCGATCGTAGCTACTGAGC"_dna4};
    
    auto k_mer_size_23_view = text_23_elements | seqan3::views::kmer_hash(seqan3::ungapped{23u};
    EXPECT_EQ(k_mer_size_23_view.size(), 1u);
    EXPECT_EQ(k_mer_size_23_view[0], 6829917194121u);

    auto k_mer_size_24_view = text_23_elements | seqan3::views::kmer_hash(seqan3::ungapped{24u};
    EXPECT_TRUE(k_mer_size_24_view.empty());

    auto k_mer_size_25_view = text_23_elements | seqan3::views::kmer_hash(seqan3::ungapped{25u};
    EXPECT_TRUE(k_mer_size_25_view.empty());
```

PR #1700, opened 2020-03-31
Deprecate seqan3::rvalue_reference

wvandertoorn:
Resolves part of #1549

Replaces occurrences of ```seqan3::rvalue_reference_t``` with ```std::iter_rvalue_reference_t``` or ```std::ranges::range_rvalue_reference_t```

### DIFF
--- a/test/unit/core/type_traits/range_iterator_test.cpp
+++ b/test/unit/core/type_traits/range_iterator_test.cpp
@@ -120,17 +120,15 @@ TEST(range_and_iterator, rvalue_reference_)
     using iterator_of_int_vector = std::ranges::iterator_t<std::vector<int>>;
     using foreign_iterator = seqan3::detail::random_access_iterator<std::vector<int>>;
     auto v = std::views::iota(1);
-    using type_list_example = seqan3::type_list<seqan3::rvalue_reference_t<std::vector<int>>, // short
-                                                typename seqan3::rvalue_reference<std::vector<int>>::type, // long
+    using type_list_example = seqan3::type_list<std::ranges::range_rvalue_reference_t<std::vector<int>>, // short
 // No types have member,yet:
 //                                              typename std::vector<int>::rvalue_reference, // member type
-                                                seqan3::rvalue_reference_t<std::vector<int> const>, // const container
-                                                seqan3::rvalue_reference_t<iterator_of_int_vector>, // iterator
-                                                seqan3::rvalue_reference_t<foreign_iterator>, // iterator2
-                                                seqan3::rvalue_reference_t<decltype(v)>>; // range, no member
+                                                std::ranges::range_rvalue_reference_t<std::vector<int> const>, // const container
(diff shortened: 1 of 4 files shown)


test/unit/core/type_traits/range_iterator_test.cpp
```diff
@@ -120,17 +120,15 @@ TEST(range_and_iterator, rvalue_reference_)
     using iterator_of_int_vector = std::ranges::iterator_t<std::vector<int>>;
     using foreign_iterator = seqan3::detail::random_access_iterator<std::vector<int>>;
     auto v = std::views::iota(1);
-    using type_list_example = seqan3::type_list<seqan3::rvalue_reference_t<std::vector<int>>, // short
-                                                typename seqan3::rvalue_reference<std::vector<int>>::type, // long
```
MitraDarja: Why did you delete these two lines?

wvandertoorn: These lines explicitly test the behaviour of the seqan3::rvalue_reference struct that is being replaced. But maybe the complete delete isn't the way to go, @marehr ?

wvandertoorn:
@seqan/core I see test are not running through because the deprecation warnings are being treated as errors, how does one go about this?

rrahn:
@wvandertoorn But that means that there are still parts of the code, that uses these deprecated traits doesn't it? So they should be fixed.

test/unit/core/type_traits/range_iterator_test.cpp
marehr: For this expression

```c++
typename seqan3::rvalue_reference<std::vector<int>>::type, // long
```

there is no equivalent in the standard, because they only support the short notation `seqan3::rvalue_reference_t`.

See https://en.cppreference.com/w/cpp/ranges/iterator_t

So we just remove this test case.

MitraDarja: Ok, thanks for the clearification. :)

wvandertoorn:
@rrahn I double checked it once more locally and everything builds and passes. I discussed it with @marehr, and the cause of the problem is that 
``` 
template <typename t>
using rvalue_reference_t = typename rvalue_reference<t>::type; 
``` 
in ```core/type_traits/pre.hpp```  ([here](https://github.com/seqan/seqan3/blob/2d2a45227d464527c53925fa3fe78c0ae191f52f/include/seqan3/core/type_traits/pre.hpp#L106)) uses the deprecated ```rvalue_reference```. @marehr said he would think about how to deal with this.

wvandertoorn:
The actual deprecation of `seqan3::rvalue_reference` will be part of a later, separate PR. See discussion https://github.com/seqan/seqan3/pull/1698#issuecomment-608473000 for details.

marehr:
@rrahn I was kinda involved in this PR, can you give the final okay?

PR #1701, opened 2020-03-31
[MISC] Remove unnecessary template arguments from search_cfg::max_error.

smehringer:
This also gets rid of dependencies on ranges.

### DIFF
--- a/include/seqan3/search/configuration/max_error.hpp
+++ b/include/seqan3/search/configuration/max_error.hpp
@@ -13,15 +13,12 @@
 
 #pragma once
 
-#include <range/v3/numeric/accumulate.hpp>
-
 #include <seqan3/core/algorithm/configuration.hpp>
 #include <seqan3/core/algorithm/pipeable_config_element.hpp>
 #include <seqan3/core/detail/pack_algorithm.hpp>
-#include <seqan3/range/views/slice.hpp>
 #include <seqan3/search/configuration/detail.hpp>
 #include <seqan3/search/configuration/max_error_common.hpp>
-#include <seqan3/std/algorithm>
+#include <seqan3/search/detail/search_common.hpp>
 
 namespace seqan3::search_cfg
 {
@@ -33,18 +30,10 @@ namespace seqan3::search_cfg
  *          a deletion corresponds to a base deleted from the query sequence that does occur in the indexed text.
  *          Deletions at the beginning and at the end of the sequence are not considered during a search.
  */
-template <typename ...errors_t>
-//!\cond
-    requires sizeof...(errors_t) <= 4 &&
-            ((detail::is_type_specialisation_of_v<std::remove_reference_t<errors_t>, total> ||
-              detail::is_type_specialisation_of_v<std::remove_reference_t<errors_t>, substitution> ||
-              detail::is_type_specialisation_of_v<std::remove_reference_t<errors_t>, deletion>  ||
-              detail::is_type_specialisation_of_v<std::remove_reference_t<errors_t>, insertion>) && ...)
-//!\endcond
-class max_error : public pipeable_config_element<max_error<errors_t...>, std::array<uint8_t, 4>>
+class max_error : public pipeable_config_element<max_error, detail::search_param>
 {
     //!\brief An alias type for the base class.
-    using base_t = pipeable_config_element<max_error<errors_t...>, std::array<uint8_t, 4>>;
+    using base_t = pipeable_config_element<max_error, detail::search_param>;
 
     //!\brief Helper function to check valid max error configuration.
     template <typename ..._errors_t>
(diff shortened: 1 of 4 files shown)


include/seqan3/search/configuration/max_error.hpp
```diff
@@ -67,26 +56,21 @@ class max_error : public pipeable_config_element<max_error<errors_t...>, std::ar
...
-    constexpr max_error(max_error &&)                  noexcept = default; //!< Move constructor.
-    constexpr max_error & operator=(max_error const &) noexcept = default; //!< Copy assignment.
-    constexpr max_error & operator=(max_error &&)      noexcept = default; //!< Move assignment.
-    ~max_error()                                       noexcept = default; //!< Destructor.
+    constexpr max_error() = default; //!< Defaulted.
+    constexpr max_error(max_error const &) = default; //!< Defaulted.
+    constexpr max_error(max_error &&) = default; //!< Defaulted.
+    constexpr max_error & operator=(max_error const &) = default; //!< Defaulted.
+    constexpr max_error & operator=(max_error &&) = default; //!< Defaulted.
+    ~max_error() = default; //!< Defaulted
```
Irallia, line 73: Can you explain to me, why you changed the descriptions?
Why not say 'Defaulted move constructor'?

smehringer: We agreed on this a long while ago, because this documentation is really only for avoiding doxygen warnings (since doxygen shows the `= default` in the docu) so we will just consistently use ``//!< Defaulted.` everywhere when constructors/assignment are explicitily defaulted :)

smehringer:
@rrahn I don't know why the code coverage fails. It shows missed lines on the switch case but I am very sure that insertions and deletions are configured at some point in the search test..

include/seqan3/search/configuration/max_error.hpp
rrahn, line 73:
```suggestion
    ~max_error() = default; //!< Defaulted.
```

rrahn: can you please add `noexcept` for move construction and move assignment as well as the destructor.

include/seqan3/search/configuration/max_error.hpp
```diff
@@ -110,42 +94,40 @@ class max_error : public pipeable_config_element<max_error<errors_t...>, std::ar
...
         if constexpr (((std::remove_reference_t<errors_t>::_id() == 0) || ...) && sizeof...(errors) == 1)
         {
-            std::ranges::fill(base_t::value | views::slice(1, 4), base_t::value[0]);
+            value.substitution = value.insertion = value.deletion = value.total;
         } // otherwise if total is not set but any other field is set than use total as the sum of all set errors.
         else if constexpr (!((std::remove_reference_t<errors_t>::_id() == 0) || ...) && sizeof...(errors) > 0)
         {
-            base_t::value[0] = std::min(static_cast<uint8_t>(255), ranges::accumulate(base_t::value | views::slice(1, 4),
-                                                                              static_cast<uint8_t>(0)));
+            value.total = std::min(255u, static_cast<unsigned>(value.substitution) + value.insertion + value.deletion);
```
rrahn, line 127:
```suggestion
            value.total = std::min<uint32_t>(255, value.substitution + value.insertion + value.deletion);
```

smehringer: I deliberately did this. I see now in CompilerExplorer that it works, but why?!
In your suggestion, the sum (of ins + del+ sub) will be cast into `uint32_t`. Doesn't the cast happen **after** the computation of the sum, but the sum might have overflowed already?

include/seqan3/search/configuration/max_error.hpp
smehringer: It would be weird to state some of them but not all, indicating that the other might actually not be noexcept. So either we put noexcept everywhere explicitly or let it be done implicitly by the compiler.

include/seqan3/search/configuration/max_error.hpp
rrahn: No, beacuse there is no `+-operator` for int8_t types defined. So they get promoted to int arithmetics which will cause no overflow in that case.

smehringer: ah, thanks for clarifying

include/seqan3/search/configuration/max_error.hpp
rrahn: not resolved! 💅

include/seqan3/search/detail/search_common.hpp
```diff
@@ -23,13 +23,26 @@ namespace seqan3::detail
...
+    uint8_t substitution{};
     //!\brief Total number of insertion errors.
-    uint8_t insertion;
+    uint8_t insertion{};
     //!\brief Total number of deletion errors.
-    uint8_t deletion;
+    uint8_t deletion{};
+
+    //!\brief Returns `true` if all member variables of `lhs` and `rhs` are equal, `false` otherwise.
+    constexpr friend bool operator==(search_param const & lhs, search_param const & rhs) noexcept
```
rrahn, line 35: test?

smehringer: it is tested in search_configuration_test where I compare the max_error configuration that holds a now `search_param` struct

PR #1702, opened 2020-03-31
[MISC] Update range library

marehr:
In https://github.com/ericniebler/range-v3/pull/1455 the range-v3 library
harmonises range-v3 with C++20's `std::ranges`. One of these changes is
the behaviour of `std::ranges::enable_view`.

This change is needed for #1680 that will simplify code regarding the
`seqan3::gap_decorator`.

### DIFF
--- a/test/unit/alignment/debug_stream_aligned_sequence_test.cpp
+++ b/test/unit/alignment/debug_stream_aligned_sequence_test.cpp
@@ -11,8 +11,8 @@
 #include <seqan3/alphabet/gap/gapped.hpp>
 #include <seqan3/alphabet/nucleotide/dna4.hpp>
 #include <seqan3/alphabet/nucleotide/rna5.hpp>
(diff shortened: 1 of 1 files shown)


marehr:
Current build failure was introduced in https://github.com/ericniebler/range-v3/commit/6103268542c27210dbc3f99f1fa0c1e348b52184.

include/seqan3/std/span
```diff
@@ -70,12 +70,23 @@ inline constexpr bool enable_safe_range;
 template<class span_tp>
     requires std::is_span<span_tp>::value
 inline constexpr bool enable_safe_range<span_tp> = true;
+
+// ranges library defines it as, there as we allow const and reference versions:
```
joergi-w, line 74: There is something wrong in this sentence with `it as, there as`

marehr: This was fixed in #1712

PR #1703, opened 2020-04-01
[TEST] Add max_error(_rate) configuration test and remove redundant search tests

smehringer:
The search test was often just testing if two configurations which are semantically the same also result in the same `search()` result. This is a lot of overhead in the search test. 
Instead, I added configuration tests and removed the redundant search tests.

### DIFF
--- a/test/unit/search/search_configuration_test.cpp
+++ b/test/unit/search/search_configuration_test.cpp
@@ -52,6 +52,114 @@ TYPED_TEST(search_configuration_test, configuration_exists)
     EXPECT_TRUE(decltype(cfg)::template exists<TypeParam>());
 }
 
+TEST(search_configuration_test, max_error_defaults)
+{
+    // empty config defaults to 0 for every error
+    EXPECT_EQ((seqan3::search_cfg::max_error{}.value),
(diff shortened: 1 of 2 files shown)


test/unit/search/search_configuration_test.cpp
```diff
@@ -52,6 +52,114 @@ TYPED_TEST(search_configuration_test, configuration_exists)
     EXPECT_TRUE(decltype(cfg)::template exists<TypeParam>());
 }
 
+TEST(search_configuration_test, max_error_defaults)
+{
+    // empty config defaults to 0 for every error
+    EXPECT_EQ((seqan3::search_cfg::max_error{}.value),
```
MitraDarja, line 58: E capitalized? (If so, then also for all following comments) Plus a dot in the end?
If not capitalized, than I would write the ifs also with an uncapitalized i, to be consistent.

smehringer: I made the other comments start with a lower case i since they are not proper sentences (so no dot)

smehringer:
> But, I don't understand why the coverage test fails.

me neither .. I rebase to current master in the hope that it might have been a spurious bug.

PR #1704, opened 2020-04-01
Pass result to alignment/add function traits

rrahn:
Adds a traits class to evaluate properties of std::function objects such as the return type or the type of the arguments. 
This is a part of #1692, which will in the end remove the dependency to call make_result_type at various different places with potentially different types and thereby causing unforseen issues.

### DIFF
--- a/include/seqan3/core/type_traits/function.hpp
+++ b/include/seqan3/core/type_traits/function.hpp
@@ -12,7 +12,51 @@
 
 #pragma once
 
-#include <seqan3/core/platform.hpp>
+#include <functional>
+
+#include <seqan3/core/type_list/traits.hpp>
+
+namespace seqan3
+{
+// ----------------------------------------------------------------------------
+// function_traits
+// ----------------------------------------------------------------------------
+
+//!\cond
+template <typename function_t>
+struct function_traits;
+//!\endcond
+
+/*!\brief A traits class to provide a uniform interface to the properties of a std::function type.
+ * \ingroup type_traits
+ *
+ * seqan3::function_traits is the type trait class that provides a uniform interface to the properties of
+ * a std::function type.
+ * This makes it possible to access the return type and the argument types of the stored target function.
+ *
+ * ### Example
+ *
+ * \include snippet/core/type_traits/function_traits.cpp
+ */
+template <typename return_t, typename ...args_t>
+struct function_traits<std::function<return_t(args_t...)>>
+{
+    //!\brief The number of arguments passed to the std::function target.
+    static constexpr size_t argument_count = sizeof...(args_t);
+
+    //!\brief The return type of the function target.
+    using result_type = return_t;
+
+    /*!\brief The argument type at the given `index`.
+     * \tparam index The position of the argument to get the type for; must be smaller than `argument_count`.
+     */
+    template <size_t index>
+    //!\cond
+        requires index < argument_count
+    //!\endcond
+    using argument_type_at = pack_traits::at<index, args_t...>;
+};
+} // namespace seqan3
 
 namespace seqan3::detail
 {
--- a/test/snippet/core/type_traits/function_traits.cpp
+++ b/test/snippet/core/type_traits/function_traits.cpp
@@ -1,0 +1,18 @@
+#include <functional>
+#include <string>
+
+#include <seqan3/core/type_traits/function.hpp>
+#include <seqan3/std/concepts>
+
+std::function my_caller = [] (size_t position, std::string & sequence)
+{
+    assert(position < sequence.size());
+    return sequence[position];
+};
+
+using my_function_t = decltype(my_caller);
+
+static_assert(std::same_as<seqan3::function_traits<my_function_t>::result_type, char>);
+static_assert(seqan3::function_traits<my_function_t>::argument_count == 2);
+static_assert(std::same_as<seqan3::function_traits<my_function_t>::argument_type_at<0>, size_t>);
+static_assert(std::same_as<seqan3::function_traits<my_function_t>::argument_type_at<1>, std::string &>);
--- a/test/unit/core/type_traits/basic_test.cpp
+++ b/test/unit/core/type_traits/basic_test.cpp
@@ -18,6 +18,10 @@
 #include <seqan3/core/type_traits/basic.hpp>
 #include <seqan3/range/detail/random_access_iterator.hpp>
 
+//------------------------------------------------------------------------------
+// Tests for remove_cvref transformation trait
+//------------------------------------------------------------------------------
+
 TEST(type_trait, remove_cvref_t)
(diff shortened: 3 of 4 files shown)


include/seqan3/core/type_traits/function.hpp
```diff
@@ -12,7 +12,50 @@
...
+// ----------------------------------------------------------------------------
+// function_traits
+// ----------------------------------------------------------------------------
+
+//!\cond
+template <typename function_t>
+struct function_traits;
+//!\endcond
+
+/*!\brief A traits class to provide uniform interface to the properties of a std::function type.
```
smehringer, line 30:
```suggestion
/*!\brief A traits class to provide a uniform interface to the properties of an `std::function` type.
```

include/seqan3/core/type_traits/function.hpp
```diff
@@ -12,7 +12,50 @@
...
+ * \ingroup type_traits
+ *
+ * function_traits is the type trait class that provides uniform interface to the properties of std::function types.
```
smehringer, line 33:
```suggestion
 * `seqan3::function_traits` is the type trait class that provides a uniform interface to the properties of `std::function` types.
```

include/seqan3/core/type_traits/function.hpp
```diff
@@ -12,7 +12,50 @@
 
 #pragma once
 
-#include <seqan3/core/platform.hpp>
+#include <functional>
+
+#include <seqan3/core/type_list/traits.hpp>
+
+namespace seqan3
```
smehringer, line 19: Why is this not detail like most of the core utilities?

include/seqan3/core/type_traits/function.hpp
```diff
@@ -12,7 +12,50 @@
...
+    using result_type = return_t;
+
+    /*!\brief The argument type at the given `index`.
+     * \tparam index The position of the argument to get the type for; must be smaller than `argument_count`.
+     */
+    template <size_t index>
+    //!\cond
+        requires index < argument_count
+    //!\endcond
+    using argument_type = pack_traits::at<index, args_t...>;
```
smehringer, line 56:
```suggestion
    using argument_type_at = pack_traits::at<index, args_t...>;
```

test/snippet/core/type_traits/function_traits.cpp
```diff
@@ -0,0 +1,18 @@
+#include <functional>
+#include <string>
+
+#include <seqan3/core/type_traits/function.hpp>
+#include <seqan3/std/concepts>
+
+std::function my_caller = [] (size_t arg1, std::string & arg2)
```
smehringer, line 7: :nail_care: but it's easier to understand the function at first glance.
```suggestion
std::function my_caller = [] (size_t position, std::string & sequence)
```

test/unit/core/type_traits/basic_test.cpp
```diff
@@ -34,3 +38,73 @@ TEST(type_trait, remove_cvref_t)
     EXPECT_FALSE((std::is_same_v<int*, seqan3::remove_cvref_t<int[3]>>));   // type stays same
     // the last example would be true for std::decay_t
 }
+
+//------------------------------------------------------------------------------
+// Tests for is_constexpr
+//------------------------------------------------------------------------------
+
+constexpr
+int    constexpr_nonvoid_free_fun(int i) { return i; }
```
smehringer, line 47: Since when are you in favour of aligning names with weird spaces? ^^
I would really like the `constexpr` on the same line as the function declaration since having them on top of the function is no in synch with our usual style (and it looks wierd).

test/unit/core/type_traits/basic_test.cpp
```diff
@@ -34,3 +38,73 @@ TEST(type_trait, remove_cvref_t)
...
+int nonconstexpr_nonvoid_free_fun_const_ref(int const & i) { return i; }
+
+constexpr
+void    constexpr_void_free_fun(int) { return; }
+void nonconstexpr_void_free_fun(int) { return; }
+
+struct constexpr_nonvoid_member_t
+{
+    int constexpr get_i(int i)
+    {
```
smehringer, line 61: since you were putting a small function scope on the same line as the function declaration above, I would also do it here to be consistent (or change the above)

test/unit/core/type_traits/basic_test.cpp
```diff
@@ -34,3 +38,73 @@ TEST(type_trait, remove_cvref_t)
...
+
+struct constexpr_void_member_t
+{
+    void constexpr get_i(int)
+    {}
+};
+
+struct nonconstexpr_nonvoid_member_t
+{
+    int get_i(int i)
```
smehringer, line 74: can't this function also live in the `constexpr_nonvoid_member_t` struct?

include/seqan3/core/type_traits/function.hpp
rrahn: why `an`?

include/seqan3/core/type_traits/function.hpp
rrahn: > std::iterator_traits is the type trait class that provides uniform interface to the properties of LegacyIterator types. This makes it possible to implement algorithms only in terms of iterators.

🤷‍♂

include/seqan3/core/type_traits/function.hpp
rrahn: Because it can be very useful in many other contexts as well and it works on an entity of the STL and not on seqan3 types.

test/unit/core/type_traits/basic_test.cpp
rrahn: Dunno. Seems to work. Won't try.

test/unit/core/type_traits/basic_test.cpp
rrahn: this is just copied over. I am not the author of this code. Anyway I see to fix this

include/seqan3/core/type_traits/function.hpp
rrahn: Also note, that std::function is a link to the cppreference doc.

test/unit/core/type_traits/basic_test.cpp
rrahn: Ok I looked a the code. And I decided not to change either of it. The class defintion look more like the riht style guide and that is good. The free function definitons above would really bloat the test and do not help for understanding what is happening. Since it is a test and we don't enforce strict rules on tests this should stay as is.

rrahn:
@smehringer polite ping

smehringer:
Code coverage seems unreliable these days..

smehringer:
@rrahn Do you want to clean the commit history or can I squash?

PR #1707, opened 2020-04-01
Pass result to alignment/add transformation trait to get alignment result value type

rrahn:
Adds a transformation trait to access the result value type of the alingment result class.
This change is part of seqan3 #1692.

In addition it fixes the documentation of the alignment result class.

### DIFF
--- a/include/seqan3/alignment/pairwise/alignment_result.hpp
+++ b/include/seqan3/alignment/pairwise/alignment_result.hpp
@@ -82,6 +82,10 @@ alignment_result_value_type(id_t, score_t, back_coord_t, front_coord_t, alignmen
     -> alignment_result_value_type<id_t, score_t, back_coord_t, front_coord_t, alignment_t>;
 //!\}
 
+//!\cond
+template <typename result_t>
+struct alignment_result_value_type_accessor;
+//!\endcond
 } // namespace seqan3::detail
 
 namespace seqan3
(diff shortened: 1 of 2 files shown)


include/seqan3/alignment/pairwise/alignment_result.hpp
```diff
@@ -130,17 +146,20 @@ class alignment_result
      * \{
      */
 
-    /*!\brief Constructs a seqan3::alignment_result from an `alignment_result_traits` object.
+    //!\privatesection
```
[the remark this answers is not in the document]
rrahn: No, this is intended. The better design would be to have a factory class that is a friend of this class and can invoke this constructor which should be in fact private. This comes later. For now it is ok to just document it private so that users won't see this in the API docs.

rrahn:
> Can you please explain to me the benefit of the accessor struct over exposing a public alignment_result_value_type in the result class?

The reason is that this is implementation detail and not part of the public interface of this class. The user should merely get the result object and access the computed values but not tinker with the implementation. But my changes (the ones that follow) allow me to pass the result in as an argument into the alignment function and then use this information there to obtain the result type. But since I need to pass the alignment_result object I want to have a simple way to extract the value type from it. But again this is detail and not part of the user interface.

include/seqan3/alignment/pairwise/alignment_result.hpp
```diff
@@ -254,3 +272,24 @@ class alignment_result
...
+ * \ingroup pairwise_alignment
+ *
+ * \tparam result_t The type of the alignment result.
+ *
+ * \details
+ *
+ * The template parameter must be a specialisation of seqan3::alignment_result otherwise this class is incomplete.
+ */
+template <typename result_value_t>
+struct alignment_result_value_type_accessor<alignment_result<result_value_t>>
```
smehringer, line 289: I thought you wanted to befriend the algorithm anyway, can't you do that and then access a private member type of the class?

rrahn: After todays meeting is this now resolved?

test/unit/alignment/pairwise/alignment_result_test.cpp
```diff
@@ -268,3 +268,12 @@ TEST(alignment_result_test, empty_type)
     seqan3::alignment_result tmp(tr);
     // members must not be accessed
 }
+
+TEST(alignment_result_test, access_result_value_type)
+{
+    seqan3::detail::alignment_result_value_type result_value{2u, 5};
+    seqan3::alignment_result result(result_value);
+
+    using result_value_t = typename seqan3::detail::alignment_result_value_type_accessor<decltype(result)>::type;
```
smehringer, line 277: Sorry, one more question.
Isn't this equivalent without the need of the accessor struct:
```suggestion
    using result_value_t = seqan3::list_traits::front<seqan3::detail::transfer_template_args_onto_t<decltype(result)>, seqan3::type_list>;
```

rrahn: Sure, it does the same thing.
But how does this help the reader of the code to understand what happens? 
In addition, the combiler must instantiate first a type list to and then also instantiate on that a front type which is clearly more work and then it isn't documented.

smehringer: But why do we have those meta type traits if we shouldn't make use of them because they are more overhead?

rrahn: No, they are useful when they are useful. For example when I work with tuples or type lists. But now I don't work wtih tuples or type lists. I don't need a generic solution for this because the trait is bound to the alignment resutlt.

PR #1708, opened 2020-04-01
[MISC] Use more appropriate suffix _type for type aliases defined in alignment traits.

rrahn:
Some minor change to be more consistent with naming style.

### DIFF
--- a/include/seqan3/alignment/pairwise/alignment_algorithm.hpp
+++ b/include/seqan3/alignment/pairwise/alignment_algorithm.hpp
@@ -91,8 +91,8 @@ private:
     //!\brief The type of the score debug matrix.
     using score_debug_matrix_t =
         std::conditional_t<traits_t::is_debug,
-                           two_dimensional_matrix<std::optional<typename traits_t::original_score_t>,
-                                                  std::allocator<std::optional<typename traits_t::original_score_t>>,
+                           two_dimensional_matrix<std::optional<typename traits_t::original_score_type>,
+                                                  std::allocator<std::optional<typename traits_t::original_score_type>>,
                                                   matrix_major_order::column>,
                            empty_type>;
     //!\brief The type of the trace debug matrix.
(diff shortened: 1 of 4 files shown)


include/seqan3/alignment/pairwise/detail/type_traits.hpp
```diff
@@ -77,33 +77,33 @@ struct alignment_configuration_traits
     static constexpr bool is_debug = config_t::template exists<detail::debug_mode>();
```
marehr, line 77: config_t? :)

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -247,7 +247,7 @@ class alignment_algorithm :
     {
         assert(static_cast<size_t>(std::ranges::distance(sequences)) <= traits_t::alignments_per_vector);
 
-        using simd_score_t = typename traits_t::score_t;
+        using simd_score_t = typename traits_t::score_type;
```
marehr, line 250:
```suggestion
        using simd_score_type = typename traits_t::score_type;
```
? But I don't care.

include/seqan3/alignment/pairwise/detail/type_traits.hpp
rrahn: yes, I should adapt that.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
rrahn: Well, good question. We have not defined any particular rules about this. So far I use it very inconsistent but try to establish the following rule: use `_t` if it is a template parameter of a class or function or only used in local scope of a function. Use `_type` when defined as member types of a class/struct.

smehringer: Since I wouldn't do rules for every little thing, I suggest a rule for using `_type` for public member types of a class struct (since consistency in the public API is nice), and otherwise, leave it to the author of the code.

PR #853, opened 2019-03-27
[FEATURE] iterator_tag_t

eseiler:
Resolves #791 

This adds a `iterator_tag_t` metafunction that returns the correct iterator_tag depending on the modelled concept.

This can be used instead of `std::iterator_traits<it_t>::iterator_category` and allows us to use our views (which expect a `iterator_category` and now can automatically deduce it with `iterator_tag_t`) in combination with views from the range library that do not set the `iterator_tag` iff concept support is detected (this is a C++20 behaviour).

PR #1710, opened 2020-04-02
[FIX] Iterator Tag in kmer_hash

MitraDarja:
For the minimiser view (#1654) I had problems preserving the forward_range properties, changing the tag in the kmer hash from input range to iterator_tag solved this problem.

### DIFF
--- a/include/seqan3/range/views/kmer_hash.hpp
+++ b/include/seqan3/range/views/kmer_hash.hpp
@@ -224,7 +224,7 @@ public:
     //!\brief Reference to `value_type`.
     using reference = value_type;
     //!\brief Tag this class as input iterator.
-    using iterator_category = std::input_iterator_tag;
+    using iterator_category = typename std::iterator_traits<it_t>::iterator_category;
     //!\brief Tag this class depending on which concept `it_t` models.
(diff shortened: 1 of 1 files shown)


MitraDarja:
@smehringer What other tests do you want to add? There are already concept tests.

include/seqan3/range/views/kmer_hash.hpp
```diff
@@ -224,7 +224,7 @@ class kmer_hash_view<urng_t>::shape_iterator
     //!\brief Reference to `value_type`.
     using reference = value_type;
     //!\brief Tag this class as input iterator.
-    using iterator_category = std::input_iterator_tag;
+    using iterator_category = iterator_tag_t<it_t>;
```
rrahn, line 227:
```suggestion
    using iterator_category = typename std::iterator_traits<it_t>::iterator_category;
 ```

eseiler: I think `seqan3::iterator_tag_t` was more correct.
See https://github.com/seqan/seqan3/pull/853
IIRC iterators may not provide the category, at least we encountered it sometimes...
Or we somewhere down the line expect that it is there...

PR #1712, opened 2020-04-05
[MISC] Use correct definition of std::span<T, E> (extent should be std::size_t)

marehr:
Extracted from PR #1702 to make this change better trackable.

Do we need a changelog entry for this?

### DIFF
--- a/include/seqan3/std/span
+++ b/include/seqan3/std/span
@@ -35,13 +35,13 @@
 
 namespace std
 {
-inline constexpr ptrdiff_t dynamic_extent = -1;
-template <typename span_tp, ptrdiff_t span_extent = dynamic_extent> class span;
+inline constexpr size_t dynamic_extent = -1;
+template <typename span_tp, size_t span_extent = dynamic_extent> class span;
 
 template <class span_tp>
 struct is_span_impl : public false_type {};
 
-template <class span_tp, ptrdiff_t span_extent>
+template <class span_tp, size_t span_extent>
 struct is_span_impl<span<span_tp, span_extent>> : public true_type {};
 
 template <class span_tp>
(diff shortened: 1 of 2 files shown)


joergi-w:
> Extracted from PR #1702 to make this change better trackable.
> 
> Do we need a changelog entry for this?

I suppose yes, because the changes can break application code...

marehr:
> Looks good! Are no tests affected?

No, because this is no breaking change. This just clashed with the ranges library, because they forward declared `template <typename span_tp, size_t span_extent> class std::span;`, which caused an compiler error of a mismatch between the forward declaration and the actual definition (this was literally the only way this could cause an error). 

You usage is the same as before. You could call `std::span<char, std::ptrdiff_t{5}>` and this would compile without any error or warning.

include/seqan3/std/span
```diff
@@ -70,12 +70,26 @@ inline constexpr bool enable_safe_range;
...
+// whereas we allow const and reference versions.
+//
+// the standard says:
+//   Pursuant to [namespace.std], users may specialize enable_­view to true for cv-unqualified program-defined types
+//   which model view, and false for types which do not. Such specializations shall be usable in constant expressions
+//   ([expr.const]) and have type const bool.
+// https://eel.is/c++draft/range.view#5
+template<class span_tp>
+    requires std::is_span<span_tp>::value
+inline constexpr bool enable_view<span_tp> = true;
```
rrahn, line 86: This is still valid? 

> #if RANGE_V3_VERSION >= 901

marehr: yeah this is still valid. the above `#if RANGE_V3_VERSION < 1000` could be taken out, because we require at least `>= 1000`. 

https://github.com/seqan/seqan3/blob/93de9c772d9cf96966439a525e4907eea56414c5/include/seqan3/core/platform.hpp#L69-L87

rrahn: Right. But it also means we could change the entire condition to only allow `RANGE_V3_VERSION >= 1000` or not?

marehr: done

PR #1713, opened 2020-04-06
[MISC] Some small refactoring that removes the public function search_all.

smehringer:
Best reviewed commit-by-commit!

### DIFF
--- a/include/seqan3/search/detail/search.hpp
+++ b/include/seqan3/search/detail/search.hpp
@@ -180,24 +180,17 @@ inline auto search_single(index_t const & index, query_t & query, configuration_
 template <typename index_t, typename queries_t, typename configuration_t>
 inline auto search_all(index_t const & index, queries_t && queries, configuration_t const & cfg)
 {
-    if constexpr (!std::ranges::forward_range<std::ranges::range_value_t<queries_t>>)
-    {
-        return search_all(index, std::views::single(std::forward<queries_t>(queries)), cfg);
-    }
-    else
-    {
-        // return type: for each query: a vector of text_positions (or cursors)
-        // delegate params: text_position (or cursor). we will withhold all hits of one query anyway to filter
-        //                  duplicates. more efficient to call delegate once with one vector instead of calling
-        //                  delegate for each hit separately at once.
-        using query_t = std::ranges::range_value_t<queries_t>;
-        using single_query_result_t = decltype(search_single(index, std::declval<query_t &>(), cfg));
-        using search_fn_t = std::function<single_query_result_t(query_t &)>;
+    // return type: for each query: a vector of text_positions (or cursors)
+    // delegate params: text_position (or cursor). we will withhold all hits of one query anyway to filter
+    //                  duplicates. more efficient to call delegate once with one vector instead of calling
+    //                  delegate for each hit separately at once.
+    using query_t = std::ranges::range_value_t<queries_t>;
+    using single_query_result_t = decltype(search_single(index, std::declval<query_t &>(), cfg));
+    using search_fn_t = std::function<single_query_result_t(query_t &)>;
 
-        search_fn_t search_fn = [&, cfg] (query_t & query) { return search_single(index, query, cfg); };
+    search_fn_t search_fn = [&, cfg] (query_t & query) { return search_single(index, query, cfg); };
 
-        return search_result_range{std::move(search_fn), std::forward<queries_t>(queries) | views::type_reduce};
-    }
+    return search_result_range{std::move(search_fn), std::forward<queries_t>(queries) | views::type_reduce};
 }
 //!\}
 
--- a/include/seqan3/search/detail/search.hpp
+++ b/include/seqan3/search/detail/search.hpp
@@ -159,39 +159,6 @@ inline auto search_single(index_t const & index, query_t & query, configuration_
     }
 }
 
-/*!\brief Search a query or a range of queries in an index.
- * \tparam index_t    Must model seqan3::fm_index_specialisation.
- * \tparam queries_t  Must model std::ranges::random_access_range over the index's alphabet.
- *                    a range of queries must additionally model std::ranges::forward_range.
- * \param[in] index   String index to be searched.
- * \param[in] queries A single query or a range of queries.
- * \param[in] cfg     A configuration object specifying the search parameters.
- * \returns `True` if and only if `abort_on_hit` is `true` and a hit has been found.
- *
- * ### Complexity
- *
- * Each query takes \f$O(|query|^e)\f$ where \f$e\f$ is the maximum number of errors.
- *
- * ### Exceptions
- *
- * Strong exception guarantee if iterating the query does not change its state and if invoking a possible delegate
- * specified in `cfg` also has a strong exception guarantee; basic exception guarantee otherwise.
- */
-template <typename index_t, typename queries_t, typename configuration_t>
-inline auto search_all(index_t const & index, queries_t && queries, configuration_t const & cfg)
-{
-    // return type: for each query: a vector of text_positions (or cursors)
-    // delegate params: text_position (or cursor). we will withhold all hits of one query anyway to filter
-    //                  duplicates. more efficient to call delegate once with one vector instead of calling
-    //                  delegate for each hit separately at once.
-    using query_t = std::ranges::range_value_t<queries_t>;
-    using single_query_result_t = decltype(search_single(index, std::declval<query_t &>(), cfg));
-    using search_fn_t = std::function<single_query_result_t(query_t &)>;
-
-    search_fn_t search_fn = [&, cfg] (query_t & query) { return search_single(index, query, cfg); };
-
-    return search_result_range{std::move(search_fn), std::forward<queries_t>(queries) | views::type_reduce};
-}
 //!\}
 
 } // namespace seqan3::detail
--- a/include/seqan3/search/search.hpp
+++ b/include/seqan3/search/search.hpp
@@ -73,6 +73,20 @@ struct search_configuration_validator
         }
     }
 
+    template <typename configuration_t>
+    static auto add_defaults(configuration_t const & cfg)
+    {
+        static_assert(detail::is_type_specialisation_of_v<configuration_t, configuration>,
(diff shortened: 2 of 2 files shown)


rrahn:
As part of which issue are you solving this?

smehringer:
Well, it is not really related to a specific issue since it only removes an unnecessary free function.
I guess it fits best to the general restructuring of the search algorithm.
Do we have an issue about the restructuring of the search algorithm into an algorithm functor?

rrahn:
Not yet, but I wanted to refine the board today. So probably I will have an issue then that can be used.

include/seqan3/search/search.hpp
```diff
@@ -73,6 +73,20 @@ struct search_configuration_validator
         }
     }
 
+    template <typename configuration_t>
+    static auto add_defaults(configuration_t const & cfg)
```
rrahn, line 89: Ok, two things:

First this would not be the place to add this. The class is intended for validation and not handling defaults.
Second, you replace are recursive call with another recursive call? Why the change at all then?

If you want to get rid of the recursive call you would rather have two separate functions that indicate with their name what they do and can be called in sequence. 
I propose to directly add a configuration class that offers for now only the two functions. This would be the first part of having later a separate configuration step and a invoking/result_range step.

include/seqan3/search/detail/search.hpp
```diff
@@ -180,24 +180,17 @@ inline auto search_single(index_t const & index, query_t & query, configuration_
...
-        using query_t = std::ranges::range_value_t<queries_t>;
-        using single_query_result_t = decltype(search_single(index, std::declval<query_t &>(), cfg));
-        using search_fn_t = std::function<single_query_result_t(query_t &)>;
+    // return type: for each query: a vector of text_positions (or cursors)
+    // delegate params: text_position (or cursor). we will withhold all hits of one query anyway to filter
+    //                  duplicates. more efficient to call delegate once with one vector instead of calling
+    //                  delegate for each hit separately at once.
+    using query_t = std::ranges::range_value_t<queries_t>;
+    using single_query_result_t = decltype(search_single(index, std::declval<query_t &>(), cfg));
+    using search_fn_t = std::function<single_query_result_t(query_t &)>;
```
rrahn:
```suggestion
    using query_t = std::ranges::range_reference_t<queries_t>;
```

include/seqan3/search/detail/search.hpp
```diff
@@ -180,24 +180,17 @@ inline auto search_single(index_t const & index, query_t & query, configuration_
...
 
-        search_fn_t search_fn = [&, cfg] (query_t & query) { return search_single(index, query, cfg); };
+    search_fn_t search_fn = [&, cfg] (query_t & query) { return search_single(index, query, cfg); };
```
rrahn:
```suggestion
    std::algorithm search_fn = [&, cfg] (query_t query) { return search_single(index, std::forward<query_t>(query), cfg); };
```

rrahn: From the feeling this shoud be how you would invoke it. So somewhere we iterate over the queries and call `search_single(*queries_it)`. That means this should capture the reference type of the queries range and forward it to the `search_single`. In case this is an l-value reference we forward it as such otherwise we move it and have perfect forwarding. But please look into the search_range implementation to verify that it is called like this.

include/seqan3/search/search.hpp
```diff
@@ -197,7 +209,20 @@ inline auto search(queries_t && queries,
...
 
 //!\cond DEV
-//! \overload
+// Overload for a single query (not a collection of queries)
+//!\overload
+template <fm_index_specialisation index_t,
+          typename query_t,
+          typename configuration_t = decltype(search_cfg::default_configuration)>
+    requires !std::ranges::forward_range<std::ranges::range_value_t<query_t>>
+inline auto search(query_t && query,
```
rrahn, line 228:
```suggestion
inline auto search(query_t query,
```

include/seqan3/search/search.hpp
```diff
@@ -197,7 +209,20 @@ inline auto search(queries_t && queries,
...
+                   index_t const & index,
+                   configuration_t const & cfg = search_cfg::default_configuration)
+{
+    return search(std::views::single(std::forward<query_t>(query)), index, cfg);
```
rrahn, line 232:
```suggestion
    return search(std::views::single(std::move(query)), index, cfg);
```

include/seqan3/search/search.hpp
```diff
@@ -197,7 +209,20 @@ inline auto search(queries_t && queries,
 }
 
 //!\cond DEV
-//! \overload
+// Overload for a single query (not a collection of queries)
+//!\overload
+template <fm_index_specialisation index_t,
+          typename query_t,
+          typename configuration_t = decltype(search_cfg::default_configuration)>
+    requires !std::ranges::forward_range<std::ranges::range_value_t<query_t>>
```
rrahn, line 227: Ok, this is currently only working because the above function is not constrained. Which is also not right. As soon as the above function requires forward range on the queries it breaks, because of ambigious overloads. 
I propose the following change to also reflect the actual requirements on the types:
```cpp
template <std::ranges::forward_range queries_t>
    requires std::ranges::random_access_range<std::ranges::range_value_t<queries_t>>
void foo(queries_t && qry)
{
    std::cout << "Take range of ranges!\n";
}

template <std::ranges::random_access_range query_t>
    requires (!std::ranges::range<std::ranges::range_value_t<query_t>>)
void foo(query_t && qry)
{
    std::cout << "Take single range!\n";
}
```

include/seqan3/search/search.hpp
rrahn: views::single always takes per copy so we can take per value and always move here.

include/seqan3/search/search.hpp
smehringer: > Second, you replace are recursive call with another recursive call? Why the change at all then?

I meant recursive calls of `search` in favour of delegating this task to a class.

It is true that this class is checking the configuration right now given it's name but I thought this still fits because it is all part of handling the configuration.

> If you want to get rid of the recursive call you would rather have two separate functions that indicate with their name what they do and can be called in sequence.

I also had this in mind but I don't if it's preferable over this solution?

> I propose to directly add a configuration class that offers for now only the two functions. This would be the first part of having later a separate configuration step and a invoking/result_range step.

I agree about separating those two, but why have a separate validation-class and (default-)configuration-class? Why not have one class that handles everything related to the configuration, defaults+validation+whatever? Or did I understand you wrong?

rrahn: Because it becomes quite quickly very complex and it helps for the reader if these kind of things are separated. Moreover, you have more control and can better document the intention of what it does.

rrahn: > I also had this in mind but I don't if it's preferable over this solution?

Yes it is. I also didn't do it in the alignment but would like to change that at some point as it is simply easier to understand and can be better documented.

include/seqan3/search/detail/search.hpp
smehringer: I don't know `std::algorithm`. Can you point me to the documentation?

include/seqan3/search/search.hpp
smehringer: I would not include the requirements of the random access range here. We decided to only require stuff that is needed for overload resolution and static assert the other requirements (as is done in `validate_query`)

include/seqan3/search/search.hpp
smehringer: > Yes it is. I also didn't do it in the alignment but would like to change that at some point as it is simply easier to understand and can be better documented.

See the update in commit 18b87913165776e5622fc07e3395752570b033e8
I don't think this makes it easier as it is way less compact. But I also don't care much.

include/seqan3/search/detail/search.hpp
rrahn: ah that should have been std::function of course

include/seqan3/search/search.hpp
rrahn: > I would not include the requirements of the random access range here. We decided to only require stuff that is needed for overload resolution and static assert the other requirements (as is done in validate_query)

0. Obviously, that is needed for overload resoultion.
1. That was a wrong decision and is outdated or should be if not done so yet. 
2. Validating the query is not top-level and in detail but it should cause the error on the top-level. It would be different if there were a configuration dependent assert, so that a certain property is only available for a certain configuration. Here you would throw an exception, but since you can infer it at compile time you would prefer the static_assert. But here we are talking about primary types independent of the configuration, so I think it does not apply to this.
3. What happened with your opinion that it is better to see the type requirements immediately? Seemed to be relevant in a private member function of a detail class. 😝 
4. Because with gcc-10 you get really good diagnostics anyway. Don't we have a range_of_ranges concept also? If not might be worth to discuss.

smehringer: 0. Why? the dimension is needed fot overload resolution not the random access restriction?
1. I disagree because you can add a rational in the static_assert comment.
2.  I don't understand what you are trying to tell me. `validate_query_type()` has only static_asserts, no throw statements and is evaluated top-level in the generic function `search()` not within the algorithm. I consider this top-level, even if it passes through a view::single wrapper beforehand.
3. out of context
4. Having great stuff in GCC10 does not mean we should abandon nice error messages in the former versions.

smehringer:
@rrahn Can you please resolve the comments you deem resolved :pray:

include/seqan3/search/search.hpp
```diff
@@ -24,6 +24,73 @@
...
+     *
+     * The following defaults are added if the respective configuration is not set by the user:
+     *
+     * * If seqan3::search_cfg::mode was not set, it defaults to seqan3::search_cfg::all.
+     * * If seqan3::search_cfg::output was not set, it defaults to seqan3::search_cfg::text_position.
+     */
+    template <typename configuration_t>
+    static auto add_defaults(configuration_t const & cfg)
+    {
+        static_assert(detail::is_type_specialisation_of_v<configuration_t, configuration>,
```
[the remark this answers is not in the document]
smehringer: As discussed, this will be changed in a different PR after the ext DEV meeting

PR #1715, opened 2020-04-06
[MISC] Simplify redundant code.

smehringer:
The following code change is completely equivalent to the one before but with redundant code removed.

### DIFF
--- a/include/seqan3/search/detail/search.hpp
+++ b/include/seqan3/search/detail/search.hpp
@@ -88,14 +88,12 @@ inline auto search_single(index_t const & index, query_t & query, configuration_
         max_error2.total = 0;
         while (internal_hits.empty() && max_error2.total <= max_error.total)
         {
-            /*
-            * If you only want the best hit (search_traits_t::search_best_hits), you stop after finding the first hit,
-              the hit with the least errors (`abort_on_hit` is true).
-            * If you are in strata mode (search_traits_t::search_strata_hits), you do the same as with best hits,
-              but then do the extra step afterwards (`abort_on_hit` is true).
-            * If you want all best hits (search_traits_t::search_all_best_hits), you do not stop after the first hit
-              but continue the current search algorithm/max_error2 pattern (`abort_on_hit` is true).
-            */
+            // * If you only want the best hit (search_traits_t::search_best_hits), you stop after finding the first hit,
+            //   the hit with the least errors (`abort_on_hit` is true).
+            // * If you are in strata mode (search_traits_t::search_strata_hits), you do the same as with best hits,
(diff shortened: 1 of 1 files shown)


include/seqan3/search/detail/search.hpp
```diff
@@ -80,41 +80,30 @@ inline auto search_single(index_t const & index, query_t & query, configuration_
...
+    if constexpr (search_traits_t::search_best_hits ||
+                  search_traits_t::search_all_best_hits ||
+                  search_traits_t::search_strata_hits)
     {
         detail::search_param max_error2{max_error};
         max_error2.total = 0;
         while (internal_hits.empty() && max_error2.total <= max_error.total)
         {
-            detail::search_algo<true>(index, query, max_error2, internal_delegate);
+            // search_traits_t::search_best_hits:     abort_on_hit = true
```
rrahn, line 91: ?

smehringer: This is pure documentation on why I set `abort_on_hit` to `!search_traits_t::search_all_best_hits`
I can remove it?

rrahn: It is ok to document the intention but I would prefer an actual sentence explaining the rational.

smehringer: I rephrased this into sentences so it does not look like outcommented code.

eseiler: The rational is that when you want to find *all* hits, you cannot stop once you found any hit.
Since this will end in recursive function calls, you can think of it as traversing a search tree, and you have to find all paths in the tree that satisfy the hit condition.
If you only want the *best* hit, you stop after finding the first hit (the hit with the least errors).
If you are in *strata* mode, you do the same as with best hits, but then do the extra stuff afterwards.

smehringer: Thanks. I changed it accordingly.

include/seqan3/search/detail/search.hpp
```diff
@@ -80,45 +80,41 @@ inline auto search_single(index_t const & index, query_t & query, configuration_
...
+                  search_traits_t::search_all_best_hits ||
+                  search_traits_t::search_strata_hits)
     {
         detail::search_param max_error2{max_error};
         max_error2.total = 0;
         while (internal_hits.empty() && max_error2.total <= max_error.total)
         {
-            detail::search_algo<true>(index, query, max_error2, internal_delegate);
+            /*
+            * If you only want the best hit (search_traits_t::search_best_hits), you stop after finding the first hit,
```
rrahn, line 92: Could you make it a line comment? 💅 
It looks like a doxygen entry gone wrong. 🙈

PR #1718, opened 2020-04-07
[MISC] Disable auto deduction of traits type in unbanded edit distance algorithm

rrahn:
So it works to pass in the callback function.
Part of #1692

This change has no further implications as only an internal implementation is adapted which is wrapped by a specific calling interface.

### DIFF
--- a/include/seqan3/alignment/pairwise/align_result_selector.hpp
+++ b/include/seqan3/alignment/pairwise/align_result_selector.hpp
@@ -52,7 +52,8 @@ struct align_result_selector
 {
 private:
     //!\brief The user configured score type.
-    using score_type = typename alignment_configuration_traits<configuration_t>::original_score_type;
+    using score_type = typename alignment_configuration_traits<
+                                    std::remove_reference_t<configuration_t>>::original_score_type;
 
     //!\brief Helper function to determine the actual result type.
     static constexpr auto select()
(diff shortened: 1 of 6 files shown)


include/seqan3/alignment/pairwise/align_result_selector.hpp
```diff
@@ -52,7 +52,7 @@ struct align_result_selector
 {
 private:
     //!\brief The user configured score type.
-    using score_type = typename alignment_configuration_traits<configuration_t>::original_score_type;
+    using score_type = typename alignment_configuration_traits<remove_cvref_t<configuration_t>>::original_score_type;
```
marehr, line 55: Why `cvref`? `std::remove_reference_t` should be enough, if you just want to allow `alignment_configuration_traits` to access `typename configuration_t::my_own_type`.

marehr: I run into problems when someone removed the `const` from the type and I could access a `non-const member` in a concept, but got a hard-compiler error when I actual couldn't access the member.

I would prefer to ALWAYS only remove the reference when you try to access the type.

See
https://godbolt.org/z/CpFqpw

rrahn: Yes, that is true. I saw it like enable_view were I am only allowed to overload it for cv-unqualified types. But in case of a traits class I might want to preserve constness to define the correct types. That is of course not the case with the alignment traits right now but who knows.

PR #1720, opened 2020-04-07
Pass result to alignment/add alignment function traits

rrahn:
Part of #1692 
Allows to extract the respective parameters for the returned function object wrapping the alignment algorithm. This allows later to determine the actual result type from the passed callback and decouples the configuration from the implementation of the algorithm.

### DIFF
--- a/include/seqan3/alignment/pairwise/detail/type_traits.hpp
+++ b/include/seqan3/alignment/pairwise/detail/type_traits.hpp
@@ -26,6 +26,7 @@
 #include <seqan3/core/bit_manipulation.hpp>
 #include <seqan3/core/simd/simd_traits.hpp>
 #include <seqan3/core/simd/simd.hpp>
+#include <seqan3/core/type_traits/function.hpp>
 #include <seqan3/core/type_traits/template_inspection.hpp>
 #include <seqan3/range/views/chunk.hpp>
 #include <seqan3/range/views/zip.hpp>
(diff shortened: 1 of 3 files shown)


include/seqan3/alignment/pairwise/detail/type_traits.hpp
```diff
@@ -117,4 +126,24 @@ struct alignment_configuration_traits
...
+//------------------------------------------------------------------------------
+// alignment_function_traits
+//------------------------------------------------------------------------------
+
+/*!\brief A traits class to provide a uniform access to the properties of the wrapped alignment algorithm.
+ * \ingroup pairwise_alignment
+ *
+ * \tparam function_t The type of the std::function object that stores the alignment algorithm as the target.
+ */
+template <typename function_t>
```
smehringer, line 138: Template requirements are missing.
This seems like a "generic" usable for any aligment algoirhtm functor so readable error messages would be nice?

rrahn: Yes and no. It is not completely trivial and the STL does not provide a suitable type trait to check if something is a function object. So I wanted to skip this 😁  But I make another PR that adds is_function_or_is_function_object to core/utility

rrahn: Not sure if that is so easy. Anyway. I would make it another PR and add this if this already merged.

PR #1725, opened 2020-04-08
Utility/general is function unary trait

rrahn:
This adds a unary type trait which extends the std::is_function trait to also recognise valid lambda types, std::function types and funtion pointer types. 
Part of #1692 to ensure that certain trait classes are well defined for generic types.

### Followup

- [x] #1720 add requires clause for alignment function traits. See #1726 
- [ ] Add changelog.


include/seqan3/core/type_traits/function.hpp
```diff
@@ -16,8 +16,115 @@
...
+ *
+ * \details
+ *
+ * Provides the member constant `value` which is set to `true` if one of the following expressions evaluates to `true`:
+ * * Evaluating the type in std::is_function returns `true`
+ * * Creating a std::function object from an instance of `t` is not ill-formed and causes no substitution error.
+ * Otherwise, `value` is `false`.
+ */
+template <typename t>
+struct is_function
```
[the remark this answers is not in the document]
marehr: We talked about this and this should be a concept with the name `seqan3::function_like`.

include/seqan3/core/type_traits/function.hpp
```diff
@@ -16,8 +16,115 @@
...
+template <typename ...invocable_ts>
+struct multi_invocable : invocable_ts...
+{
+    //!\brief Inherit the function call operators.
+    using invocable_ts::operator()...;
+};
+
+//!\brief Deduction guides for seqan3::detail::multi_invocable.
+template <typename ...invocable_ts>
+multi_invocable(invocable_ts...) -> multi_invocable<invocable_ts...>;
```
marehr, line 38: This change is noise and I would have preferred if old things stay at old places (extremely minimizes merge conflict potential). If you want to make this change, we can do it in a different PR.

include/seqan3/core/type_traits/function.hpp
```diff
@@ -16,8 +16,115 @@
...
+struct is_function : std::bool_constant<detail::is_function<t>::value>
+{};
+
+/*!\brief Helper variable template for seqan3::is_function unary transformation trait.
+ * \ingroup type_traits
+ *
+ * \tparam t The type to check if it is a valid function, lambda function, std::function or function pointer type.
+ */
+template <typename t>
+inline constexpr bool is_function_v = is_function<t>::value;
```
marehr, line 126: I'm not completely sure if this needs to be public api.

test/unit/core/type_traits/function_test.cpp
```diff
@@ -35,3 +35,245 @@ TEST(function_traits, argument_type_at)
...
+        return true;
+    }
+};
+
+TEST(is_function, false_functions)
+{
+    auto generic_fn = [] (auto x) { return x; };
+    EXPECT_FALSE(seqan3::is_function<int>::value);
+    EXPECT_FALSE(seqan3::is_function<int const *>::value);
+    EXPECT_FALSE(seqan3::is_function<fn_object_rvalue_ref>::value);
```
marehr, line 269: why is this no function?

test/unit/core/type_traits/function_test.cpp
```diff
@@ -35,3 +35,245 @@ TEST(function_traits, argument_type_at)
...
+{
+    auto operator()(int) &&
+    {
+        return true;
+    }
+};
+
+TEST(is_function, false_functions)
+{
+    auto generic_fn = [] (auto x) { return x; };
```
marehr, line 266: Can you add a function object with a template argument.

test/unit/core/type_traits/function_test.cpp
```diff
@@ -35,3 +35,245 @@ TEST(function_traits, argument_type_at)
...
+    EXPECT_FALSE(seqan3::is_function<int>::value);
+    EXPECT_FALSE(seqan3::is_function<int const *>::value);
+    EXPECT_FALSE(seqan3::is_function<fn_object_rvalue_ref>::value);
+    EXPECT_FALSE(seqan3::is_function<decltype(generic_fn)>::value);
```
marehr, line 270: Hmm this is bad that means this concepts can only says yes, if the given function-ptr/obj is fully specialised and has linkage. We normally write fully generic code, so under which circumstances is this function detection useful?

test/unit/core/type_traits/function_test.cpp
rrahn: a class template or a generic call operator?

rrahn:
Ok, after some discussion with @marehr and evaluating the current use cases we decided to drop this feature in favour of another solution.

PR #1730, opened 2020-04-09
[FEATURE] Add internal alignment configuration to capture alignment result from algorithm configuration

rrahn:
Part of #1692 

### Description

This change allows to capture the configured alignment result in the alignment configuration object that is passed to the alignment result. Thus, the algorithm can ensure that it uses the same alignment algorithm as it was originally configured and avoids possible incompatibilities that are hard to track.

### DIFF
--- a/include/seqan3/alignment/configuration/align_config_alignment_result_capture.hpp
+++ b/include/seqan3/alignment/configuration/align_config_alignment_result_capture.hpp
@@ -1,0 +1,78 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2020, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides seqan3::align_cfg::alignment_result_capture.
+ * \author Rene Rahn <rene.rahn AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <seqan3/alignment/configuration/detail.hpp>
+#include <seqan3/alignment/pairwise/alignment_result.hpp>
+#include <seqan3/core/algorithm/pipeable_config_element.hpp>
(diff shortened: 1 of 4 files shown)


include/seqan3/alignment/configuration/align_config_alignment_result_capture.hpp
```diff
@@ -0,0 +1,78 @@
...
+ * \brief Configuration element capturing the configured seqan3::alignment_result for the alignment algorithm.
+ * \ingroup alignment_configuration
+ * \tparam alignment_result_t The alignment result type to capture; must be a type specialisation of
+ *                            seqan3::alignment_result.
+ *
+ * \details
+ *
+ * This configuration element allows to capture the concrete seqan3::alignment_result type after configuring the
+ * alignment algorithm with the seqan3::detail::alignment_configurator. The actual result type is wrapped in
+ * std::type_identity to preserve the trivial type properties of the configuration element. Thus, on access the
```
smehringer, line 61: > The actual result type is wrapped std::type_identity to preserve the trivial type properties

Some other config elements don't preserve the trivial type properties either I think. Is this really necessary?

include/seqan3/alignment/configuration/align_config_alignment_result_capture.hpp
```diff
@@ -0,0 +1,78 @@
...
+{
+    //!\brief Internal id to check for consistent configuration settings.
+    static constexpr detail::align_config_id id{detail::align_config_id::alignment_result_capture};
+};
+
+} // namespace seqan3::detail
+
+namespace seqan3::align_cfg
+{
+/*!\if DEV
```
smehringer, line 51: is \if the same as \cond?

include/seqan3/alignment/configuration/align_config_alignment_result_capture.hpp
rrahn: yes, otherwise the code won't compile in some expressions.

include/seqan3/alignment/configuration/align_config_alignment_result_capture.hpp
rrahn: They are somewhat similar. The problem is that the function itself has a `\cond \endcond` expression around the reuires expression. But I had some weird issues with the nested `cond` expression (due to the requires block) so that was an elegant solution to let them only appear in dev docu.

PR #1731, opened 2020-04-09
[MISC] use std::(ranges::range|iter)_(value|reference)_t

marehr:
This fixes part of #1549. I'm not sure why this version works, but I
tested it for gcc 7, 8 and 9 and it worked.

The main problem I had was to allow the syntax:

```c++
// this should call the constructor which takes two iterators
concatenated_sequences seqs{iterator, sentinel};
```

https://github.com/seqan/seqan3/blob/75f934f1886ee225fb1bad01b9c9f474caf4c508/include/seqan3/range/container/concatenated_sequences.hpp#L299-L307

In many versions I tried, this call instead tried

https://github.com/seqan/seqan3/blob/75f934f1886ee225fb1bad01b9c9f474caf4c508/include/seqan3/range/container/concatenated_sequences.hpp#L321-L328

This caused some issues, because the template also tried to instantiate
those constructors with not-yet defined class types (like for example the
iterator type within the class)

All those complicated overloads are only available in this class, to
allow constructions of sequences of "compatible" type (i.e. range value
types that are convertible to the value type of the class)
`std::vector<rna4>` and `std::vector<dna4>`.

For the future (and maybe gcc10 compatibility) we should remove the
ability to plug any possible way to add somehow compatible types. This
should be limited.

For now the proposed fix keeps the old behaviour.

### DIFF
--- a/include/seqan3/range/container/concatenated_sequences.hpp
+++ b/include/seqan3/range/container/concatenated_sequences.hpp
@@ -223,7 +223,9 @@ protected:
     //!\hideinitializer
     // cannot use the concept, because this class is not yet fully defined
     template <typename t>
+    //!\cond
         requires std::ranges::range<std::iter_reference_t<t>>
+    //!\endcond
     static constexpr bool iter_value_t_is_compatible_with_value_type =
         !std::is_same_v<remove_cvref_t<t>, concatenated_sequences> &&
         (
@@ -236,7 +238,9 @@ protected:
     //!\hideinitializer
     // cannot use the concept, because this class is not yet fully defined
     template <std::ranges::range t>
+    //!\cond
         requires std::ranges::range<std::ranges::range_reference_t<t>>
+    //!\endcond
     static constexpr bool range_value_t_is_compatible_with_value_type =
         !std::is_same_v<remove_cvref_t<t>, iterator> &&
         !std::is_same_v<remove_cvref_t<t>, const_iterator> &&
@@ -264,7 +268,8 @@ public:
     ~concatenated_sequences() = default;
 
     /*!\brief Construct/assign from a different range.
-     * \tparam rng_of_rng_type The type of range to be inserted; must satisfy \ref is_compatible_this.
+     * \tparam rng_of_rng_type The type of range to be inserted; must satisfy
+     *         \ref range_value_t_is_compatible_with_value_type.
      * \param rng_of_rng The sequences to construct/assign from.
      *
      * ###Complexity
@@ -292,7 +297,7 @@ public:
     }
 
     /*!\brief Construct/assign with `count` times `value`.
-     * \tparam rng_type The type of range to be inserted; must satisfy \ref is_compatible_value.
+     * \tparam rng_type The type of range to be inserted; must satisfy \ref is_compatible_with_value_type.
      * \param count Number of elements.
      * \param value The initial value to be assigned.
      *
@@ -319,7 +324,7 @@ public:
 
     /*!\brief Construct/assign from pair of iterators.
      * \tparam begin_iterator_type Must satisfy std::forward_iterator and must satisfy
-     * \ref is_compatible_value.
+     *         \ref iter_value_t_is_compatible_with_value_type.
      * \tparam end_iterator_type Must satisfy std::sized_sentinel_for.
      * \param begin_it begin of range to construct/assign from.
      * \param end_it end of range to construct/assign from.
@@ -343,7 +348,7 @@ public:
     }
 
     /*!\brief Construct/assign from `std::initializer_list`.
-     * \tparam rng_type The type of range to be inserted; must satisfy \ref is_compatible_value.
+     * \tparam rng_type The type of range to be inserted; must satisfy \ref is_compatible_with_value_type.
      * \param ilist an `std::initializer_list` of `rng_type`.
      *
      * ###Complexity
@@ -355,14 +360,16 @@ public:
      * Strong exception guarantee (no data is modified in case an exception is thrown).
      */
     template <std::ranges::forward_range value_type_t = value_type>
+    //!\cond
         requires is_compatible_with_value_type<value_type_t>
+    //!\endcond
     concatenated_sequences(std::initializer_list<value_type_t> ilist)
     {
         assign(std::begin(ilist), std::end(ilist));
     }
 
     /*!\brief Construct/assign from `std::initializer_list`.
-     * \tparam rng_type The type of range to be inserted; must satisfy \ref is_compatible_value.
+     * \tparam rng_type The type of range to be inserted; must satisfy \ref is_compatible_with_value_type.
      * \param ilist an `std::initializer_list` of `rng_type`.
      *
      * ###Complexity
@@ -384,7 +391,8 @@ public:
     }
 
     /*!\brief Construct/assign from a different range.
-     * \tparam rng_of_rng_type The type of range to be inserted; must satisfy \ref is_compatible_this.
+     * \tparam rng_of_rng_type The type of range to be inserted; must satisfy
+     *         \ref range_value_t_is_compatible_with_value_type.
      * \param rng_of_rng The sequences to construct/assign from.
      *
      * ###Complexity
@@ -406,7 +414,7 @@ public:
     }
 
     /*!\brief Construct/assign with `count` times `value`.
-     * \tparam rng_type The type of range to be inserted; must satisfy \ref is_compatible_value.
+     * \tparam rng_type The type of range to be inserted; must satisfy \ref is_compatible_with_value_type.
      * \param count Number of elements.
      * \param value The initial value to be assigned.
      *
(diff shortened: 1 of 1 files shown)


include/seqan3/range/container/concatenated_sequences.hpp
```diff
@@ -182,33 +182,73 @@ class concatenated_sequences
...
-                                                (dimension_v<t> == dimension_v<value_type> &&
-                                                std::convertible_to<reference_t<t>, value_type_t<value_type>>);
-    //!\}
+    template <std::ranges::range t>
+    static constexpr bool is_compatible_with_value_type =
+        std::is_same_v<remove_cvref_t<t>, value_type> ||
+        std::is_same_v<remove_cvref_t<t>, reference> ||
+        std::is_same_v<remove_cvref_t<t>, const_reference> ||
+        (
+            !std::same_as<remove_cvref_t<t>, iterator> &&
```
rrahn, line 201: is this not contradicting the requirement of t in the top-level expression? In other words, can it ever be a iterator here?

include/seqan3/range/container/concatenated_sequences.hpp
```diff
@@ -182,33 +182,73 @@ class concatenated_sequences
...
+        );
 
     //!\cond
     // unfortunately we cannot specialise the variable template so we have to add an auxiliary here
     template <typename t>
-        requires (dimension_v<t> == dimension_v<value_type> + 1) &&
-                  is_compatible_value<reference_t<t>>
-    static constexpr bool is_compatible_this_aux = true;
+        requires std::ranges::range<std::iter_reference_t<t>> &&
+                 (dimension_v<t> == dimension_v<value_type> + 1) &&
```
rrahn, line 211: `t` is an iterator and can have a dimension?

include/seqan3/range/container/concatenated_sequences.hpp
```diff
@@ -342,15 +383,16 @@ class concatenated_sequences
     template <std::ranges::forward_range rng_type>
     concatenated_sequences & operator=(std::initializer_list<rng_type> ilist)
```
rrahn, line 384: maybe adapt the template name here as well, since you changed already above

include/seqan3/range/container/concatenated_sequences.hpp
marehr: that's the problem here, if you are within an evaluation during the definition of this class.

That means the C++ parser started 

```c++
template <typename inner_type,
          typename data_delimiters_type = std::vector<typename inner_type::size_type>>
//!\cond
    requires reservible_container<std::remove_reference_t<inner_type>> &&
             reservible_container<std::remove_reference_t<data_delimiters_type>> &&
             std::is_same_v<size_type_t<inner_type>, value_type_t<data_delimiters_type>>
//!\endcond
class concatenated_sequences
{
```
and it did not reach

```c++
} // closing brace of concatenated_sequences
```

concatenated_sequences and all it's member are in a weird limbo state there not everything might be completely defined yet, e.g. iterator might be an incomplete type. The standard behaviour of gcc < 10 is to just assume that the concept check on such a type is true and tries to instantiate all the expressions in this static bool (you could comment that `std::ranges::range t` thing out and it would have the same behaviour). The static bool forces the expression to be able to short-circut and `std::same_as` is able to compare with incomplete types.

Like I wrote, I'm not completely sure why this works, but I have seen that the requires has no effect.

include/seqan3/range/container/concatenated_sequences.hpp
marehr: yes, right now the dimension is defined as:

https://github.com/seqan/seqan3/blob/35f3efb613c9cb039f227f38aba207aa49ecceb4/include/seqan3/core/type_traits/range.hpp#L196-L216

This behaviour is the same as before and yeah, I know, that this is utterly confusing.

This has to be fixed somehow. 

Maybe by `dimension_v<t> == dimension_v<std::iter_reference_t<t>> + 1 == dimension_v<value_type> + 1`

include/seqan3/range/container/concatenated_sequences.hpp
```diff
@@ -283,7 +324,7 @@ class concatenated_sequences
 
     /*!\brief Construct/assign from pair of iterators.
      * \tparam begin_iterator_type Must satisfy std::forward_iterator and must satisfy
-     * \ref is_compatible_value.
+     *         \ref iter_value_t_is_compatible_with_value_type.
```
marehr, line 327: buhu

include/seqan3/range/container/concatenated_sequences.hpp
```diff
@@ -228,7 +268,8 @@ class concatenated_sequences
     ~concatenated_sequences() = default;
 
     /*!\brief Construct/assign from a different range.
-     * \tparam rng_of_rng_type The type of range to be inserted; must satisfy \ref is_compatible_this.
+     * \tparam rng_of_rng_type The type of range to be inserted; must satisfy
+     *         \ref range_value_t_is_compatible_with_value_type.
```
marehr, line 272: buhu

include/seqan3/range/container/concatenated_sequences.hpp
```diff
@@ -342,15 +383,16 @@ class concatenated_sequences
...
     //!\endcond
     {
         assign(std::begin(ilist), std::end(ilist));
         return *this;
     }
 
     /*!\brief Construct/assign from a different range.
-     * \tparam rng_of_rng_type The type of range to be inserted; must satisfy \ref is_compatible_this.
+     * \tparam rng_of_rng_type The type of range to be inserted; must satisfy
+     *         \ref range_value_t_is_compatible_with_value_type.
```
marehr, line 395: buhu

include/seqan3/range/container/concatenated_sequences.hpp
marehr: Sorry can't find a rename, in what should I rename this?

include/seqan3/range/container/concatenated_sequences.hpp
```diff
@@ -384,18 +426,19 @@ class concatenated_sequences
...
     //!\endcond
     {
         concatenated_sequences rhs{count, value};
         swap(rhs);
     }
 
     /*!\brief Construct/assign from pair of iterators.
-     * \tparam begin_iterator_type Must satisfy std::forward_iterator and satisfy \ref is_compatible_value.
+     * \tparam begin_iterator_type Must satisfy std::forward_iterator and satisfy
+     *         \ref iter_value_t_is_compatible_with_value_type.
```
marehr, line 441: buhu

include/seqan3/range/container/concatenated_sequences.hpp
```diff
@@ -998,7 +1044,8 @@ class concatenated_sequences
     }
 
     /*!\brief Inserts elements from range `[first, last)` before position in the container.
-     * \tparam begin_iterator_type Must satisfy std::forward_iterator and \ref is_compatible_value.
+     * \tparam begin_iterator_type Must satisfy std::forward_iterator and
+     *         \ref iter_value_t_is_compatible_with_value_type.
```
marehr, line 1048: buhu

include/seqan3/range/container/concatenated_sequences.hpp
rrahn: That sounds definitely like a to complicated design, where the benefits seem to be mariginal.

include/seqan3/range/container/concatenated_sequences.hpp
rrahn: >  I know, that this is utterly confusing.

If just that. :) I think it was a really good decion fromt the standard to seperate the type traits between iterators and ranges and we should adhere to that. But ok, it was like so before, so we can keep it.

include/seqan3/range/container/concatenated_sequences.hpp
rrahn: Like in the initialiser list constructor? There you changed `rng_type` to `value_type_t`. And now in the assignment you don't, but I guess what you like to express is. Well, it is a generic type but please make note, that it should be `value_type`. Also I think as you said in the comments in the future this should only be `std::initializer_list<value_type>` and not some generic compatible type. That will always be troublesome through weird conversion sequences.

include/seqan3/range/container/concatenated_sequences.hpp
marehr: Yeah :(

include/seqan3/range/container/concatenated_sequences.hpp
```diff
@@ -318,17 +359,17 @@ class concatenated_sequences
...
      * Strong exception guarantee (no data is modified in case an exception is thrown).
      */
-    template <std::ranges::forward_range rng_type = value_type>
-    concatenated_sequences(std::initializer_list<rng_type> ilist)
+    template <std::ranges::forward_range value_type_t = value_type>
     //!\cond
-        requires is_compatible_value<rng_type>
+        requires is_compatible_with_value_type<value_type_t>
     //!\endcond
+    concatenated_sequences(std::initializer_list<value_type_t> ilist)
```
marehr, line 366:
```suggestion
    concatenated_sequences(std::initializer_list<value_type_t> ilist)
```

Ah now I found it, you meant this one.

include/seqan3/range/container/concatenated_sequences.hpp
marehr: Not completely sure what I should do, I assume you want `rng_type` renamed to `value_type_t`.

marehr:
@rrahn Can you have a final look? Thank you :)

PR #1732, opened 2020-04-09
[MISC] use std::ranges::range_reference_t

marehr:
This one is quite hard to understand:

Before it slowly build up the required expression

```c++
alphabet<std::ranges::range_reference_t<std::ranges::range_reference_t<typename t::ref_sequences>>>
// the given ref_sequences are ranges of sequences (a sequence is a range over an alphabet)
```

and had fallbacks if one of those steps failed in that expression, such that the whole line
is ALWAYS a valid expression what-so-ever happens.

I moved the expression into a requires(), because this has the property
if anything fails or is not defined the whole requires() expression will
evaluate to false. Exactly the same semantics as before, but a bit
better readable.

Fixes part of #1549


eseiler:
As a followup question: Now we don't check for forward range anymore? Is this important to check?
I.e. for ref_ids (the next few lines) we have a very similar check and we check for forward range.

PR #1733, opened 2020-04-09
[MISC] use std::ranges::range_reference_t

marehr:
We can't use `detail::transformation_trait_or_t` any more, because
`std::ranges::range_reference_t` is not a transformation trait.

We can't just use the expression

```c++
std::ranges::range_reference_t<typename traits_type::ref_sequences const>;
```

because this expression can fail if `ref_sequences` is not a range and
triggers a hard compiler error.

We use `std::lazy` to only evaluate the above expression, if
`ref_sequence` is really a range, otherwise it falls back to `dummy_ref_type`.
By doing this the type `ref_sequence_unsliced_type` will ALWAYS be
defined.

I could have also only written that in one statement, like before, but I
think this is now more readable.

Fixes part of #1549

### DIFF
--- a/include/seqan3/io/alignment_file/input.hpp
+++ b/include/seqan3/io/alignment_file/input.hpp
@@ -370,6 +370,16 @@ private:
     //!\brief The dummy ref sequence type if no reference information were given.
     using dummy_ref_type = decltype(views::repeat_n(typename traits_type::sequence_alphabet{}, size_t{}) |
                                     std::views::transform(detail::access_restrictor_fn{}));
+
+    //!\brief The unsliced ref sequence type if reference information were given.
+    using ref_sequence_unsliced_type =
+        detail::lazy_conditional_t<std::ranges::range<typename traits_type::ref_sequences const>,
+                                  detail::lazy<std::ranges::range_reference_t,
+                                               typename traits_type::ref_sequences const>,
+                                  dummy_ref_type>;
+
+    //!\brief The ref sequence type if reference information were given.
+    using ref_sequence_sliced_type = decltype(std::declval<ref_sequence_unsliced_type>() | views::slice(0, 0));
 public:
     /*!\name Field types and record type
      * \brief These types are relevant for record/row-based reading; they may be manipulated via the \ref traits_type
@@ -391,11 +401,7 @@ public:
      */
     using ref_sequence_type = std::conditional_t<std::same_as<typename traits_type::ref_sequences, ref_info_not_given>,
                                                  dummy_ref_type,
-                                                 decltype(std::declval<
-                                                     detail::transformation_trait_or_t<
-                                                        seqan3::reference<typename traits_type::ref_sequences const>,
-                                                        dummy_ref_type> /* does not matter as type is not chosen */
-                                                     >() | views::slice(0, 0))>;
+                                                 ref_sequence_sliced_type>;
     /*!\brief The type of field::ref_id is fixed to std::optional<int32_t>.
      *
      * To be consistent with the BAM format, the field::ref_id will hold the index to the actual reference


include/seqan3/io/alignment_file/input.hpp
```diff
@@ -370,6 +370,16 @@ class alignment_file_input
     //!\brief The dummy ref sequence type if no reference information were given.
     using dummy_ref_type = decltype(views::repeat_n(typename traits_type::sequence_alphabet{}, size_t{}) |
                                     std::views::transform(detail::access_restrictor_fn{}));
+
+    //!\brief The unsliced ref sequence type if reference information were given.
+    using ref_sequence_unsliced_type =
+        detail::lazy_conditional_t<std::ranges::range<typename traits_type::ref_sequences const>,
+                                  detail::lazy<std::ranges::range_reference_t,
+                                               typename traits_type::ref_sequences const>,
+                                  dummy_ref_type>;
```
smehringer, line 379: Ok so it has been a while since I wrote this code so I might remember something wrong, but I think your change is not wuite equivalent.
AFAIK: The reference sequence is either a sliced range or a dummy_ref (which has the exact size needed to it is not sliced)

Your condition is std::ranges::range<typename traits_type::ref_sequences const>
(which is the same as in line 402, it is confusing that it differs here). So the "else" type here is never chosen in `ref_sequence_type` since the second std::conditional in line 402 chooses the `dummy_ref_type` anyway. So you now have 2 conditional expressions instead of the 1 that is needed? That doesn't make it much easier to understand as one might think both are necessary?

marehr: Transformation step by step

```c++
// new:
ref_sequence_sliced_type

// old expression:
decltype(std::declval<
    detail::transformation_trait_or_t<
       seqan3::reference<typename traits_type::ref_sequences const>,
       dummy_ref_type> /* does not matter as type is not chosen */
    >() | views::slice(0, 0))>;
```

---

We need to show that `ref_sequence_sliced_type` is semantically the same as before.

```c++
// new
using ref_sequence_sliced_type = decltype(std::declval<ref_sequence_unsliced_type>() | views::slice(0, 0));

// old
decltype(std::declval<
//
    detail::transformation_trait_or_t<
       seqan3::reference<typename traits_type::ref_sequences const>,
       dummy_ref_type> /* does not matter as type is not chosen */
//
    >() | views::slice(0, 0))>;
```

---

We need to show that `ref_sequence_unsliced_type` is semantically the same as before.

```c++
// new 
using ref_sequence_unsliced_type =
    detail::lazy_conditional_t<std::ranges::range<typename traits_type::ref_sequences const>,
                              detail::lazy<std::ranges::range_reference_t,
                                           typename traits_type::ref_sequences const>,
                              dummy_ref_type>;

// old

detail::transformation_trait_or_t<
   seqan3::reference<typename traits_type::ref_sequences const>,
   dummy_ref_type> /* does not matter as type is not chosen */
```

`seqan3::reference::type` is defined and thus the same as `std::ranges::range_reference_t` iff `std::ranges::range` is true.

That means the old expression would return `seqan3::reference_t = std::ranges::range_reference_t` iff `typename traits_type::ref_sequences const` is a range.

```c++
detail::lazy_conditional_t<std::ranges::range<typename traits_type::ref_sequences const>,
                          detail::lazy<std::ranges::range_reference_t,
                                       typename traits_type::ref_sequences const>,
                          dummy_ref_type>;

// is semantically the same as

std::conditional_t<std::ranges::range<typename traits_type::ref_sequences const>,
                   std::ranges::range_reference_t<typename traits_type::ref_sequences const>,
                   dummy_ref_type>;
```

That means the new expression returns `std::ranges::range_reference_t<typename traits_type::ref_sequences const>` iff `typename traits_type::ref_sequences const` is a range.

---

Concluding that each sub-expression is semantically equivalent and thus the new and old `ref_sequence_type` expresses the same.

marehr: > Your condition is std::ranges::range
> (which is the same as in line 402, it is confusing that it differs here). So the "else" type here is never chosen in `ref_sequence_type` since the second std::conditional in line 402 chooses the `dummy_ref_type` anyway. So you now have 2 conditional expressions instead of the 1 that is needed? That doesn't make it much easier to understand as one might think both are necessary?

The (new) line 402 says for me

```c++
   using ref_sequence_type = std::conditional_t<std::same_as<typename traits_type::ref_sequences, ref_info_not_given>,
```

I don't see there any condition `std::ranges::range`?! That makes it a bit hard to understand your argument.

The old and the new expression have both 2 conditional expression.

In the old expression it was more hidden, because the expression `detail::transformation_trait_or_t` is defined as:

https://github.com/seqan/seqan3/blob/70b4e57575ae1f92be2f0f5e70e0ba7d090e492d/include/seqan3/core/type_traits/transformation_trait_or.hpp#L42-L45

smehringer: Ok, I see now that you are right that your transformation is equivalent. 

> I don't see there any condition std::ranges::range?! That makes it a bit hard to understand your argument.

Yes but the tag `ref_info_not_given` is set when the type is not a range. So the two are still testing the same thing.
Anyway, it's not so important. I just remembered that we might get rid of this weird dummy sequence and then this will also simplify.

PR #1735, opened 2020-04-10
requires clauses must be primary expressions

marehr:
This fix is part of seqan/product_backlog#53.

### DIFF
--- a/include/seqan3/core/pod_tuple.hpp
+++ b/include/seqan3/core/pod_tuple.hpp
@@ -178,7 +178,7 @@ pod_tuple(types && ...) -> pod_tuple<types...>;
 template <std::size_t i, typename ...types>
 constexpr auto & get(seqan3::pod_tuple<types...> & t) noexcept
 //!\cond
-    requires i < sizeof...(types)
+    requires (i < sizeof...(types))
 //!\endcond
 {
     if constexpr (i == 0)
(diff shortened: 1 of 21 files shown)


marehr:
> Just to be sure the requires here are something different?
> 
> https://github.com/seqan/seqan3/blob/332f4b3749f7baeee1efb6e1e2cd310df0f2fdc0/include/seqan3/range/container/concept.hpp#L165
> .

https://github.com/seqan/seqan3/blob/332f4b3749f7baeee1efb6e1e2cd310df0f2fdc0/include/seqan3/range/container/concept.hpp#L165-L169

The first requires means the boolean expression must be true, otherwise the complete concept is false.
The second requires "creates" a new scope and is true if everything within that scope is true or false if one of the statements fail.

include/seqan3/core/pod_tuple.hpp
```diff
@@ -353,10 +353,13 @@ constexpr auto const && get(seqan3::pod_tuple<types...> const && t) noexcept
  * \see [std::tuple_element](https://en.cppreference.com/w/cpp/utility/tuple/tuple_element)
  */
 template <std::size_t i, template <typename ...> typename t, typename ...types>
-    requires i < sizeof...(types) &&
+//!\cond
+    requires (i < sizeof...(types)) &&
             std::is_base_of_v<seqan3::pod_tuple<types...>, t<types...>>
```
smehringer, line 358:
```suggestion
             std::is_base_of_v<seqan3::pod_tuple<types...>, t<types...>>
```
:grin: :nail_care:

marehr:
> So `foo_concept<...>` as well as `foo_concept<...> && moo_concept` is a primary expression but `!foo_concept<...>` isn't?

Yes

See https://en.cppreference.com/w/cpp/language/constraints#Requires_clauses too

Issue #1608, opened 2020-02-25
Improve the tutorial


wvandertoorn:
Improvements to the tutorial, per section.

#### Quick Setup (using CMake)

-   [x] 0.0.1 Directory Structure

    The directory should look like this:

            tutorial
            ├── source
            ├── build
            └── seqan3
                ├── build_system
                ├── doc
                ├── include
                ├── submodules
                ├── test
                ├── CHANGELOG.md
                ├── CMakeLists.txt
                ├── CODE_OF_CONDUCT.md
                ├── CONTRIBUTING.md
                ├── LICENSE.md
                └── README.md

### Tutorial

-   [x] 1.0.1 Name/Number all the exercises and assignments in the tutorial in a consistent way.

    In the current state it's hard to keep well-arranged notes on where exactly that usefull information is explained / snippet can be found.
-   [ ] 1.0.2 Section with common usecases code.

    I think it would be very useful to include a section of more basic snippets to the How-To. 
    A listing of common use cases helps to quickly see some of the things SeqAn3 can do. The structure of the tutorial makes that is it hard to obtain such an overview (different pages, lot's of scrolling).

    Remarks such as "For this, have a look at this section of the tutorial/at this submodule" and hyperlinks to the tutorial should be included. This improves using the tutorial as documentation for users with a moderate knowledge of the library (not currently doing the tutorial, but having a question).

-   [ ] 1.0.3 Include pointers/tips/information on which headers to use or take a look at.

    E.g. as in "Exercise: Reading a FASTQ file" from Sequence file I/O, Reading a sequence file, Reading records:

    "Note: you include the seqan3::debug_stream with the following header
    `#include <seqan3/core/debug_stream.hpp>` "

    Having such remarks in the task helps with learning where everything is defined.

#### First steps

- [x] 1.1.0 Fix Note on "this is a code snippet"

##### Align two sequences

-   [x] 1.1.1 Be more explicit about header includes.

    In the section above, the user has been introduced to Modules in SeqAn3, i.e. how to include the  needed headers. I feel that knowing which headers to include for your code is a big part of learning to work with SeqAn3 as well.  `seqan3::align_cfg::edit` and `seqan3::align_cfg::result` are used in the snippet, but I was confused by how that could work without the proper includes. I would suggest to be more explicit on the includes here. Either by:

    -   Teaching to include the header files per function
        ```cpp
          #include <seqan3/alignment/configuration/align_config_edit.hpp>
[...]

PR #1737, opened 2020-04-15
Improvements to the Ranges section of the tutorial

wvandertoorn:
Resolves part of #1608 (1.5)

Kept as separate commits for readability.

### DIFF
--- a/doc/tutorial/ranges/index.md
+++ b/doc/tutorial/ranges/index.md
@@ -47,21 +47,21 @@ refined concepts (i.e. types that model a stronger concept, always also model th
 
 | Concept                           | Description                                                 |
 |-----------------------------------|-------------------------------------------------------------|
-| std::ranges::input_range           | can be iterated from beginning to end **at least once**     |
-| std::ranges::forward_range         | can be iterated from beginning to end **multiple times**    |
-| std::ranges::bidirectional_range   | iterator can also move backwards with `--`                  |
-| std::ranges::random_access_range    | you can jump to elements **in constant-time** `[]`          |
-| std::ranges::contiguous_range      | elements are always stored consecutively in memory          |
+| std::ranges::input_range          | can be iterated from beginning to end **at least once**     |
+| std::ranges::forward_range        | can be iterated from beginning to end **multiple times**    |
+| std::ranges::bidirectional_range  | iterator can also move backwards with `--`                  |
+| std::ranges::random_access_range  | you can jump to elements **in constant-time** `[]`          |
+| std::ranges::contiguous_range     | elements are always stored consecutively in memory          |
 
 For the well-known containers from the standard library this matrix shows which concepts they model:
 
-|                                   | std::forward_list | std::list | std::deque | std::array | std::vector |
-|-----------------------------------|:-----------------:|:---------:|:----------:|:----------:|:-----------:|
-| std::ranges::input_range           | ✅                 | ✅         | ✅          | ✅          | ✅           |
-| std::ranges::forward_range         | ✅                 | ✅         | ✅          | ✅          | ✅           |
-| std::ranges::bidirectional_range   |                   | ✅         | ✅          | ✅          | ✅           |
-| std::ranges::random_access_range    |                   |           | ✅          | ✅          | ✅           |
-| std::ranges::contiguous_range      |                   |           |            | ✅          | ✅           |
+|                                    | std::forward_list | std::list | std::deque | std::array | std::vector |
+|------------------------------------|:-----------------:|:---------:|:----------:|:----------:|:-----------:|
+| std::ranges::input_range           | ✅                | ✅        | ✅         | ✅         | ✅          |
+| std::ranges::forward_range         | ✅                | ✅        | ✅         | ✅         | ✅          |
+| std::ranges::bidirectional_range   |                   | ✅        | ✅         | ✅         | ✅          |
+| std::ranges::random_access_range   |                   |           | ✅         | ✅         | ✅          |
+| std::ranges::contiguous_range      |                   |           |            | ✅         | ✅          |
 
 There are also range concepts that are independent of input or output or one of the above concept, e.g.
 std::ranges::sized_range which requires that the size of a range can be computed and in constant time.
@@ -163,7 +163,7 @@ std::cout << *v.begin() << '\n'; // should print 4
 ```
 \endassignment
 \solution
-\snippet doc/tutorial/ranges/range_snippets.cpp solution1
+\include doc/tutorial/ranges/range_solution1.cpp
 \endsolution
 
 ## View concepts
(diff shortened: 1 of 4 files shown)


doc/tutorial/ranges/index.md
```diff
@@ -47,21 +47,21 @@ refined concepts (i.e. types that model a stronger concept, always also model th
...
-| std::ranges::bidirectional_range   |                   | ✅         | ✅          | ✅          | ✅           |
-| std::ranges::random_access_range    |                   |           | ✅          | ✅          | ✅           |
-| std::ranges::contiguous_range      |                   |           |            | ✅          | ✅           |
+|                                    | std::forward_list | std::list | std::deque | std::array | std::vector |
+|------------------------------------|:-----------------:|:---------:|:----------:|:----------:|:-----------:|
+| std::ranges::input_range           | ✅                | ✅        | ✅         | ✅         | ✅          |
+| std::ranges::forward_range         | ✅                | ✅        | ✅         | ✅         | ✅          |
+| std::ranges::bidirectional_range   |                   | ✅        | ✅         | ✅         | ✅          |
+| std::ranges::random_access_range   |                   |           | ✅         | ✅         | ✅          |
+| std::ranges::contiguous_range      |                   |           |            | ✅         | ✅          |
```
Irallia, line 64: There is a problem with this sign everywhere in our code. This ✅ is wider than one normal charakter.
I think we decided to leave it as it is. Or what do you say @seqan/core?
But there is still at least one space you can leave. :)

doc/tutorial/ranges/index.md
```diff
@@ -208,21 +208,22 @@ Which of the following concepts do you think `v` models?
...
-| std::ranges::contiguous_range    |        |
-|                                  |        |
+| std::ranges::random_access_range |         |
+| std::ranges::contiguous_range    |         |
+|                                  |         |
 | std::ranges::view                |   ✅    |
-| std::ranges::sized_range         |        |
-| std::ranges::output_range        |        |
+| std::ranges::sized_range         |         |
+| std::ranges::output_range        |         |
```
Irallia, line 216: same here

doc/tutorial/ranges/index.md
wvandertoorn: @seqan/core polite ping, what is the consensus on how to deal with the check mark character?

rrahn: yes you can leave it like this.

doc/tutorial/ranges/index.md
```diff
@@ -208,21 +208,22 @@ Which of the following concepts do you think `v` models?
...
 
-The filter does not preserve RandomAccess and therefore not Contiguous, because it doesn't "know" which element
-of the underlying range is the i-th one in constant time.
-This also means we don't know the size.
+The filter does not preserve the RandomAccess concept because it doesn't "know" which element of the underlying range
```
rrahn, line 218: Here is what I would write:
```
Suprised? Let's have a closer look at the std::views::filter view. The filter view only returns the value of the underlying range for which the given predicate evaluates to `true`. To know which value is element of the filter view the view has to look at each of them. Thus, it must scan the underlying range value-by-value and cannot jump to an arbitrary location in constant time since it cannot know how many elements it had to skip without looking at them. Accordingly, the std::views::filter preserves only std::ranges::bidirectional_range, because it can scan the text in reverse order as well. Since the view cannot guarantee that the values lie in contigous memory it can also not preserve std::ranges::contiguous_range. Similarly, the view cannot model std::ranges::sized_range as it cannot determine the number of values not filtered out in constant time.
```

doc/tutorial/ranges/index.md
```diff
@@ -208,21 +208,22 @@ Which of the following concepts do you think `v` models?
...
+is the i-th one in constant time, therefore `v` does not model `std::ranges::random_access_range`. From this, it
+follows that `v` does not model `std::ranges::contiguous_range` nor `std::ranges::sized_range`, because both require
+RandomAccess.
 
 The transform on the other hand produces a new element on every access (the result of the multiplication), therefore
 `v` is not an output range, you cannot assign values to its elements.
```
rrahn, line 224:
```suggestion
`v` is not an std::ranges::output_range, you cannot assign values to its elements.
```

doc/tutorial/ranges/index.md
```diff
@@ -208,21 +208,22 @@ Which of the following concepts do you think `v` models?
...
-This would have prevented modelling the contiguous_range as well – if it hadn't been already by the filter – because
-values are created on-demand and are not stored in memory at all.
+Note that this prevents modelling the contiguous_range as well because values are created on-demand and are not stored
```
rrahn, line 225:
```suggestion
Note that this prevents modelling the std::ranges::contiguous_range as well because values are created on-demand and are not stored
```

doc/tutorial/ranges/range_solution1.cpp
```diff
@@ -0,0 +1,12 @@
+#include <iostream>
+#include <vector>
+#include <seqan3/std/ranges>                    // include all of the standard library's views
+
+int main()
+{
+    std::vector vec{1, 2, 3, 4, 5, 6};
+    auto v = vec | std::views::filter(   [] (auto const i) { return i % 2 == 0; })
```
rrahn, line 8:
```suggestion
    auto v = vec | std::views::filter([] (auto const i) { return i % 2 == 0; })
```

doc/tutorial/ranges/range_solution3.cpp
```diff
@@ -1,13 +1,30 @@
...
 #include <seqan3/std/ranges>                    // include all of the standard library's views
 
+
 int main(int argc, char** argv)
 {
     if (argc == 1)
         return 0;
 
     std::string s{argv[1]};
+    // Optional, use the argument_parser instead:
```
rrahn, line 13: Mhmm, I don't understand why we should propose the alternative with the argument parser but only add it as a comment? I would say we always use the argument parser. Otherwise the learners end up with a solution which we would actually not promote.

wvandertoorn: Good point. I changed the solution to use the argument parser.

doc/tutorial/ranges/index.md
wvandertoorn: I used your piece, it's really nice, thanks!

doc/tutorial/ranges/index.md
```diff
@@ -208,21 +208,19 @@ Which of the following concepts do you think `v` models?
...
+| std::ranges::sized_range         |         |
+| std::ranges::output_range        |         |
 
-The filter does not preserve RandomAccess and therefore not Contiguous, because it doesn't "know" which element
-of the underlying range is the i-th one in constant time.
-This also means we don't know the size.
+Suprised? Let's have a closer look at the std::views::filter view. The filter view only returns the value of the underlying range for which the given predicate evaluates to `true`. To know which value is element of the filter view the view has to look at each of them. Thus, it must scan the underlying range value-by-value and cannot jump to an arbitrary location in constant time since it cannot know how many elements it had to skip without looking at them. Accordingly, the std::views::filter preserves only std::ranges::bidirectional_range, because it can scan the text in reverse order as well. Since the view cannot guarantee that the values lie in contiguous memory it can also not preserve std::ranges::contiguous_range. Similarly, the view cannot model std::ranges::sized_range as it cannot determine the number of values not filtered out in constant time.
```
eseiler, line 218:
```suggestion
Surprised? Let's have a closer look at the std::views::filter view. The filter view only returns the value of the 
underlying range for which the given predicate evaluates to `true`. To know which value is an element of the filter
view, the view has to look at each of them. Thus, it must scan the underlying range value-by-value and cannot jump to an 
arbitrary location in constant time since it cannot know how many elements it had to skip without looking at them. 
Accordingly, the std::views::filter preserves only std::ranges::bidirectional_range, because it can scan the text in 
reverse order as well. Since the view cannot guarantee that the values lie in contiguous memory, it can also not 
preserve std::ranges::contiguous_range. Similarly, the view cannot model std::ranges::sized_range as it cannot determine
the number of values not filtered out in constant time.
```

line length + some typos

doc/tutorial/ranges/index.md
```diff
@@ -208,21 +208,19 @@ Which of the following concepts do you think `v` models?
...
 
 The transform on the other hand produces a new element on every access (the result of the multiplication), therefore
-`v` is not an output range, you cannot assign values to its elements.
-This would have prevented modelling the contiguous_range as well – if it hadn't been already by the filter – because
-values are created on-demand and are not stored in memory at all.
+`v` is not a std::ranges::output_range, you cannot assign values to its elements.
+Note that this prevents modelling the std::ranges::contiguous_range as well because values are created on-demand and are not stored
+in memory at all.
```
eseiler, line 223:
```suggestion
Note that this prevents modelling the std::ranges::contiguous_range as well because values are created on-demand and
are not stored in memory at all.
```

rrahn:
@wvandertoorn this looks great now. The only thing left todo would be to fix the commit messages. We have this style guide about commit messages, which you can find here: http://docs.seqan.de/seqan/3-master-user/about_contributing.html
So the commits you want to preserve should have the prefix `[DOC]`.

PR #1740, opened 2020-04-15
[FEATURE] Adapt the alignment algorithm to pass in the alignment result type via callback instead of redefining it in all places.

rrahn:
This allows us to define the alignment result type only once at the top-level when the algorithm is configured.
Accoridingly, the type must not be redefined within the algorithms and possible incompatibilities between the calls are avoided.
In addition, the algorithm can now decide when to call the user callback which will allow in the future to signal back to the algorithm what should happen. For example interrupt the execution.

part of #1692 

Note, there is a follow up PR that will add the feature to get the result type from the alignment config (see #1730). This then simplifies the definitions of the callback constraints in the algorithms.


test/performance/alignment/edit_distance_unbanded_benchmark.cpp
```diff
@@ -61,15 +66,21 @@ void seqan3_edit_distance_dna4(benchmark::State & state)
...
                                                                            decltype(edit_distance_cfg),
+                                                                           alignment_result_t,
                                                                            std::false_type>;
 
+
     for (auto _ : state)
     {
         seqan3::detail::edit_distance_unbanded edit_distance{seq1, seq2, edit_distance_cfg, edit_traits_t{}};
-        edit_distance(0u);
+        edit_distance(0u, [] (auto) {/*noop*/});
```
wvandertoorn, line 83: I think the "do nothing", as you used in `test/unit/alignment/pairwise/edit_distance_unbanded_test_template.hpp` is a bit nicer than "noop"

test/performance/alignment/edit_distance_unbanded_benchmark.cpp
```diff
@@ -143,17 +154,22 @@ void seqan3_edit_distance_dna4_collection(benchmark::State & state)
...
+                                                                           alignment_result_t,
                                                                            std::false_type>;
 
     for (auto _ : state)
     {
         for (auto && [seq1, seq2] : vec)
         {
             seqan3::detail::edit_distance_unbanded edit_distance{seq1, seq2, edit_distance_cfg, edit_traits_t{}};
-            edit_distance(0u);
+            edit_distance(0u, [] (auto) {/*noop*/});
```
wvandertoorn, line 172: here too

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -173,31 +175,31 @@ class alignment_algorithm :
      * |space (begin positions) |\f$ O(n*m) \f$    |\f$ O(n*k) \f$     |
      * |space (alignment)       |\f$ O(n*m) \f$    |\f$ O(n*k) \f$     |
      */
-    template <indexed_sequence_pair_range indexed_sequence_pairs_t>
+    template <indexed_sequence_pair_range indexed_sequence_pairs_t,
+              typename callback_t>
```
smehringer, line 179: put on the line above :nail_care:

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -173,31 +175,31 @@ class alignment_algorithm :
...
-        return results;
+            compute_single_pair(idx, get<0>(sequence_pair), get<1>(sequence_pair), callback);
     }
 
     //!\overload
-    template <indexed_sequence_pair_range indexed_sequence_pairs_t>
-    auto operator()(indexed_sequence_pairs_t && indexed_sequence_pairs)
+    template <indexed_sequence_pair_range indexed_sequence_pairs_t, typename callback_t>
+    void operator()(indexed_sequence_pairs_t && indexed_sequence_pairs,
+                    callback_t && callback)
```
smehringer, line 197: put on line above

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -173,31 +175,31 @@ class alignment_algorithm :
      * |space (begin positions) |\f$ O(n*m) \f$    |\f$ O(n*k) \f$     |
      * |space (alignment)       |\f$ O(n*m) \f$    |\f$ O(n*k) \f$     |
      */
-    template <indexed_sequence_pair_range indexed_sequence_pairs_t>
+    template <indexed_sequence_pair_range indexed_sequence_pairs_t,
+              typename callback_t>
     //!\cond
-        requires !traits_t::is_vectorised
+        requires !traits_t::is_vectorised &&
+                 is_type_specialisation_of_v<
```
smehringer, line 182: shouldn't this (also) require "invokable on result type"?

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -665,22 +676,18 @@ class alignment_algorithm :
...
-        using indexed_sequence_pair_t = std::ranges::range_value_t<indexed_sequence_pair_range_t>;
-        using sequence_pair_t = std::tuple_element_t<0, indexed_sequence_pair_t>;
-        using sequence1_t = std::tuple_element_t<0, sequence_pair_t>;
-        using sequence2_t = std::tuple_element_t<1, sequence_pair_t>;
-
-        using result_value_t = typename align_result_selector<sequence1_t, sequence2_t, config_t>::type;
-
-        std::vector<alignment_result<result_value_t>> results{};
-        results.reserve(std::ranges::distance(index_sequence_pairs));
+        using alignment_result_t = typename function_traits<callback_t>::template argument_type_at<0>;
```
smehringer, line 689: it is a little confusing that the `alignment_result_t` is not equal to the `result_value_t` as they sound like synonyms, but of course, this does not relate to the PR.

include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
```diff
@@ -262,9 +260,10 @@ class alignment_executor_two_way
             current_chunk_size = std::ranges::distance(current_chunk);
             assert(in_avail() >= current_chunk_size);
 
-            exec_handler.execute(kernel, std::move(current_chunk), [write_to = gptr] (auto && alignment_results)
+            exec_handler.execute(kernel, std::move(current_chunk), [write_to = gptr] (auto && alignment_result) mutable
```
smehringer, line 263: Why does this need to be mutable now? (or maybe why didn't is need to be mutable before)

include/seqan3/alignment/pairwise/execution/execution_handler_parallel.hpp
```diff
@@ -117,7 +117,7 @@ class execution_handler_parallel
         // Asynchronously pushes the alignment job as a task to the queue.
         task_type task = [=, indexed_sequence_pairs = std::move(indexed_sequence_pairs)] ()
         {
-            delegate(algorithm(std::move(indexed_sequence_pairs)));
+            algorithm(std::move(indexed_sequence_pairs), delegate);
```
smehringer, line 120: now it's called delegate and not callback? Are they synonyms or is there a difference?

test/performance/alignment/edit_distance_unbanded_benchmark.cpp
```diff
@@ -61,15 +66,21 @@ void seqan3_edit_distance_dna4(benchmark::State & state)
     auto seq2 = generate_sequence<seqan3::dna4>(sequence_length, 0, 1);
     int score = 0;
 
-    using edit_traits_t = seqan3::detail::default_edit_distance_trait_type<std::add_lvalue_reference_t<decltype(seq1)>,
-                                                                           std::add_lvalue_reference_t<decltype(seq2)>,
+    using seq1_ref_t = std::add_lvalue_reference_t<decltype(seq1)>;
```
smehringer, line 69: uh I didn't know this type trait exists. good to know. Does this just add a `&` to the type?

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
rrahn: that will be anyway changed in a follow up PR since we pass now the result type as part of the alignment configuration and then can actually use the regular std::invocable concept.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
rrahn: but it will be anyway resolve once we use the builder pattern for the alignment as well

test/performance/alignment/edit_distance_unbanded_benchmark.cpp
rrahn: that and more. It also adheres to the reference collapsing rules.

include/seqan3/alignment/pairwise/execution/execution_handler_parallel.hpp
rrahn: It was called like this before and yes they have the same meaning.

include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
rrahn: All captured arguments are const by default unless you declare the lambda mutable. Before I had a const pointer but didn't care because I dereferenced it and since const on pointers does not propgate I could write to the memory that write_to was pointing to. The current change implies that the algorithm calls the delgate once for every computed alignment result instead of collecting them in a vector and returning the vector. So I know have to make sure that the lambda updates the `write_to` position, which is only possible if it is not `const`.

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
rrahn: can I do this in the upcoming PRs since I am anyway touching these lines?

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
rrahn: Like above. I'd like to do it in the next PRs since I am changing this anyway.

rrahn:
@smehringer thanks for the review. I'd like to make some of the 💅  comments in the next PRs that follow since I am touching these lines anyway due to the adaption of the concept.

rrahn:
@smehringer because of merge conflicts I had to rebase anyway. So I also changed the three lines you commented on.

rrahn:
The jenkins error is unrelated. While I was playing around with jenkins I used this branch to figure out what is going on. I might have changed something in the local buffer for this particular branch on catalina. So this is not affecting other PRs.

smehringer:
ops.
Travis passes but doesn't update so I'll merge

PR #1742, opened 2020-04-16
[MISC] Remove unnecessary template arguments from max_error_rate class.

smehringer:
Analog to #1701


smehringer:
Closed for now after the search config discussion is through

PR #1749, opened 2020-04-20
[MISC] Remove 'using namespace seqan3' from test/unit/alignment/*.

### LINKED ISSUE #1748: [MISC] Remove 'using namespace seqan3' from test/unit/alignment/* (left overs)
Remove 'using namespace seqan3' from
- [ ] test/unit/alignment/*
(without /configuration, /matrix, /pairwise)

Irallia:
Resolves #1748

### DIFF
--- a/test/unit/alignment/aligned_sequence_test.cpp
+++ b/test/unit/alignment/aligned_sequence_test.cpp
@@ -20,7 +20,7 @@ class aligned_sequence_<container_type> : public ::testing::Test
 {
 public:
     // Initializer function is needed for the typed test because the gapped_decorator
-    // will be initialized differently than the naive vector<seqan3::gapped<dna>>.
+    // will be initialized differently than the naive std::vector<seqan3::gapped<dna>>.
     void initialise_typed_test_container(container_type & container_, seqan3::dna4_vector const & target)
     {
         container_.clear();
--- a/test/unit/alignment/aligned_sequence_test.cpp
+++ b/test/unit/alignment/aligned_sequence_test.cpp
@@ -14,26 +14,24 @@
 
 #include "../alignment/aligned_sequence_test_template.hpp"
 
-using namespace seqan3;
-
 template <typename container_type>
-    requires aligned_sequence<container_type>
+    requires seqan3::aligned_sequence<container_type>
 class aligned_sequence_<container_type> : public ::testing::Test
 {
 public:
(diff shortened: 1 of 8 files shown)


test/unit/alignment/aligned_sequence_test.cpp
```diff
@@ -14,26 +14,24 @@
...
 template <typename container_type>
-    requires aligned_sequence<container_type>
+    requires seqan3::aligned_sequence<container_type>
 class aligned_sequence_<container_type> : public ::testing::Test
 {
 public:
     // Initializer function is needed for the typed test because the gapped_decorator
-    // will be initialized differently than the naive vector<gapped<dna>>.
-    void initialise_typed_test_container(container_type & container_, dna4_vector const & target)
+    // will be initialized differently than the naive vector<seqan3::gapped<dna>>.
```
marehr, line 23:
```suggestion
    // will be initialized differently than the naive std::vector<seqan3::gapped<dna>>.
```

Irallia:
> Thank you :)
> 
> BTW was this the last one?

Nearly. 2 to go ;-)

test/unit/alignment/aligned_sequence_test_template.hpp
```diff
@@ -15,12 +16,9 @@
 #include <seqan3/core/detail/debug_stream_alphabet.hpp>
 #include <seqan3/core/detail/debug_stream_range.hpp>
 #include <seqan3/io/alignment_file/detail.hpp>
-#include <seqan3/std/iterator>
 
 using seqan3::operator""_dna4;
 
-using namespace seqan3; // necessary right now because assign_unaligned is falsely a hidden friend
```
smehringer, line 22: this is not needed anymore?

Irallia: This line was commented by @h-2, but all tests were all running through without this line. ¯\_(ツ)_/¯

h-2: It's true that thiscan be removed, but it is not possible to write `seqan3::assign_unaligned()` instead of just `assign_unaligned()` which would be the correct way.

eseiler: Time for git bisect :D

h-2: I've sent the core team my thought on how this should be designed with CPOs...

PR #1750, opened 2020-04-20
[FIX] Remove const from static assert

### LINKED ISSUE #1754: take_until does not work with reverse
Combining the view take_until with reverse and then using kmer_hash did lead to two separate problems.

First, the static assert in kmer_hash fails, because kmer_hash expects a const-iterability view, which is not true for `seqan3::views::take_until | std::views::reverse`.
Removing the const from the static assert in kmer_hash should solve this problem.

```cpp
auto stop_at_t = seqan3::views::take_until([] (seqan3::dna4 const x) { return x == 'T'_dna4; });
std::vector<seqan3::dna4> textt{"ACGTAGC"_dna4};
seqan3::debug_stream << (textt | stop_at_t | seqan3::views::complement | std::views::reverse
                                                    | seqan3::views::kmer_hash(seqan3::ungapped{4})
                                                    | std::views::reverse) << "\n";
// Leads to a static assert failure
```

Second, using only take_until with reverse did not lead to the correct result:

```cpp
auto stop_at_t = seqan3::views::take_until([] (seqan3::dna4 const x) { return x == 'T'_dna4; });
std::vector<seqan3::dna4> textt{"ACGTAGC"_dna4};
seqan3::debug_stream << (textt | stop_at_t | seqan3::views::complement | std::views::reverse
                                                    | seqan3::views::kmer_hash(seqan3::ungapped{4})
                                                    | std::views::reverse) << "\n";
seqan3::debug_stream << (textt | stop_at_t  | std::views::reverse ) << "\n"; // prints: CGATGCA, not as expected GCA
```

The reason for this behaviour is that take_until does not model `sized_sentinel_for`. A solution for this should be returning `std::ranges::default_sentinel` in the end function of take until.

MitraDarja:
Resolves #1754

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -187,6 +187,8 @@ Note that 3.1.0 will be the first API stable release and interfaces in this rele
   ([\#1410](https://github.com/seqan/seqan3/pull/1410)).
 * Renamed `seqan3::views::all` to `seqan3::views::type_reduce`
   ([\#1410](https://github.com/seqan/seqan3/pull/1410)).
+* Added size() function to `seqan3::views::kmer_hash`
+  ([\#1722](https://github.com/seqan/seqan3/pull/1722)).
 
 #### Search
 
--- a/include/seqan3/range/detail/inherited_iterator_base.hpp
+++ b/include/seqan3/range/detail/inherited_iterator_base.hpp
@@ -401,6 +401,7 @@ private:
         return static_cast<derived_t const *>(this);
     }
 
+protected:
     //!\brief Cast this to base type.
     constexpr base_t * this_to_base()
     {
--- a/include/seqan3/range/views/kmer_hash.hpp
+++ b/include/seqan3/range/views/kmer_hash.hpp
@@ -39,21 +39,9 @@ template <std::ranges::view urng_t>
 class kmer_hash_view : public std::ranges::view_interface<kmer_hash_view<urng_t>>
 {
 private:
-<<<<<<< HEAD
     static_assert(std::ranges::forward_range<urng_t>, "The kmer_hash_view only works on forward_ranges");
     static_assert(semialphabet<reference_t<urng_t>>, "The reference type of the underlying range must model "
                   "seqan3::semialphabet.");
-=======
-<<<<<<< HEAD
-    static_assert(std::ranges::forward_range<urng_t const>, "The kmer_hash_view only works on forward_ranges");
-    static_assert(semialphabet<std::ranges::range_reference_t<urng_t>>,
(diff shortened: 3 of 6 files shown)


MitraDarja:
> 
> 
> Thanks for the PR!
> 
> One question: Do we need a test for this?

Maybe, I will just add my example to the kmer_hash tests. :)

marehr:
@MitraDarja Can you first clean-up your commit history, this is confusing? Thank you!

MitraDarja:
@marehr Sure, I can, but I am not completely sure what you mean with clean-up. You mean using the fixup-function to merge some commits or make the commit messages more clear?

marehr:
I talked with Rene about this PR, because I found the solution a bit "hacky". We talked about this a bit and found a better solution.

Just as a reminder: The general technic will be to introduce a new `sentinel_type`, which will store the underlying sentinel. The `iterator_type` and the `sentinel_type` need to be made equality comparable. This will have the advantage that the iterator does not need to store it's end in every instance and we follow the design of iterator and sentinels.

@rrahn I assigned you, because you wanted to help Mitra with this.

marehr:
> @marehr Sure, I can, but I am not completely sure what you mean with clean-up. You mean using the fixup-function to merge some commits or make the commit messages more clear?

yeah merging commits, each commit should tell a clear story and should be build-able.

looking at 

* https://github.com/seqan/seqan3/pull/1750/commits/eceb6c14add67a9cc0ae75af8e624acd8a49f879 comments out some `noexcept` and some tests, but why?
* https://github.com/seqan/seqan3/pull/1750/commits/a7dcc9cd9114f7b89413581c407af522af9aecd4 contains merge-conflict markers and has commented out tests?
* https://github.com/seqan/seqan3/pull/1750/commits/7db7ddf99bbcd1f0befd187490747799469f3794 removes the marker again; changes some tests?
* https://github.com/seqan/seqan3/pull/1750/commits/3a915c7b6ffce241398d4c976c34ea415b479121 comments something in again
* https://github.com/seqan/seqan3/pull/1750/commits/73d624b3d1da8a4fd3a3b0cbb4b2fabf6e6352c0 removes something which was added before?

It doesn't looks like a PR with a clean history. I can't even judge if the commit messages are even accurate, because they do multiple things and do unexpected things.

We said that we shouldn't rebase **during** PR, but I would expect a clean patch series **before** creating it. This also lessens the Review time, because you will find in that final clean-up some stuff yourself (like the commented out parts).

MitraDarja:
Solved partly by #1802,  reverse compatibility follows in an later PR.

PR #1756, opened 2020-04-22
[FIX] Fix const compatibility for the shape_iterator of the kmer_hash view.

smehringer:
When introducing the iterator test template to the kmer_view, the following does not compile:

1. iterator urng_t const vs non-const comparison
     ```cpp
     EXPECT_TRUE(std::ranges::begin(this->test_range) == std::ranges::cbegin(this->test_range));
     ```
2. iterator const urng_t from iterator urng_t construction 
    ```cpp
    const_iterator_type it{std::ranges::begin(this->test_range)};
    ```

This PR introduces a custom constructor from the non-const version that fixes both problems.
.

### DIFF
--- a/include/seqan3/range/views/kmer_hash.hpp
+++ b/include/seqan3/range/views/kmer_hash.hpp
@@ -223,6 +223,9 @@ private:
     //!\brief The sentinel type of the underlying range.
     using sentinel_t = std::ranges::sentinel_t<rng_t>;
 
+    template <typename urng2_t>
+    friend class shape_iterator;
+
 public:
     /*!\name Associated types
      * \{
@@ -252,6 +255,19 @@ public:
     constexpr shape_iterator & operator=(shape_iterator const &) = default; //!< Defaulted.
     constexpr shape_iterator & operator=(shape_iterator &&)      = default; //!< Defaulted.
     ~shape_iterator()                                            = default; //!< Defaulted.
+
+    //!\brief Allow iterator on a const range to be constructible from an iterator over a non-const range.
+    template <typename urng2_t>
+    //!\cond
+        requires std::same_as<std::remove_const_t<urng_t>, urng2_t>
+    //!\endcond
+    shape_iterator(shape_iterator<urng2_t> it) :
+        hash_value{std::move(it.hash_value)},
+        roll_factor{std::move(it.roll_factor)},
+        shape_{std::move(it.shape_)},
+        text_left{std::move(it.text_left)},
+        text_right{std::move(it.text_right)}
+    {}
 
     /*!\brief Construct from a given iterator on the text and a seqan3::shape.
     * /param[in] it_start Iterator pointing to the first position of the text.
--- a/include/seqan3/range/views/kmer_hash.hpp
+++ b/include/seqan3/range/views/kmer_hash.hpp
@@ -507,14 +507,12 @@ public:
     /*!\brief Move the iterator by a given offset and return the corresponding hash value.
      * \attention This function is only avaible if `it_t` models std::random_access_iterator.
      */
-    value_type operator[](difference_type const n)
+    reference operator[](difference_type const n) const
     //!\cond
         requires std::random_access_iterator<it_t>
     //!\endcond
     {
-        text_left += n;
-        hash_full();
-        return operator*();
+        return *(*this + n);
     }
 
     //!\brief Return the hash value.
--- a/test/unit/range/views/view_kmer_hash_test.cpp
+++ b/test/unit/range/views/view_kmer_hash_test.cpp
@@ -17,6 +17,8 @@
 #include <seqan3/range/views/take_until.hpp>
 #include <seqan3/test/expect_range_eq.hpp>
 
+#include "../iterator_test_template.hpp"
+
 #include <gtest/gtest.h>
 
 using seqan3::operator""_dna4;
@@ -28,6 +30,24 @@ static constexpr auto ungapped_view = seqan3::views::kmer_hash(seqan3::ungapped{
 static constexpr auto gapped_view = seqan3::views::kmer_hash(0b101_shape);
 static constexpr auto prefix_until_first_thymine = seqan3::views::take_until([] (seqan3::dna4 x)
                                                    { return x == 'T'_dna4; });
+
+using iterator_type = std::ranges::iterator_t<decltype(std::declval<seqan3::dna4_vector&>() | gapped_view)>;
+
+template <>
+struct iterator_fixture<iterator_type> : public ::testing::Test
+{
+    using iterator_tag = std::random_access_iterator_tag;
+    static constexpr bool const_iterable = true;
+
+    seqan3::dna4_vector text{"ACGTAGC"_dna4};
+
+    decltype(text | gapped_view) test_range = text | gapped_view;
+
+    std::vector<size_t> expected_range{2, 7, 8, 14, 1};
+};
+
+using test_type = ::testing::Types<iterator_type>;
+INSTANTIATE_TYPED_TEST_SUITE_P(iterator_fixture, iterator_fixture, test_type, );
 
 template <typename T>
 class kmer_hash_ungapped_test: public ::testing::Test {};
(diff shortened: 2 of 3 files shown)


rrahn:
It would be great to have a bug report for this, which also explains the rationale why this is missing etc. Also since you are now working on that but as part of which issue and which iteration?

smehringer:
Well, I only discovered this while investigating the coverage miss in #1654 (which was caused by not properly using the iterator_test_template). The minimiser view has the same bug but cannot be fixed before kmer_hash is fixed because it depends on the kmer_hash. 
So should I link this to the minimiser view?

test/unit/range/views/view_kmer_hash_test.cpp
```diff
@@ -17,12 +19,31 @@
...
 #include <seqan3/range/views/to.hpp>
 
-#include <gtest/gtest.h>
+#include "../iterator_test_template.hpp"
 
 using seqan3::operator""_dna4;
 using seqan3::operator""_dna5;
 using seqan3::operator""_shape;
 
+using iterator_type = std::ranges::iterator_t< decltype(std::declval<seqan3::dna4_vector&>()
```
joergi-w, line 28: remove this whitespace? 💅 
```suggestion
using iterator_type = std::ranges::iterator_t<decltype(std::declval<seqan3::dna4_vector&>()
```

include/seqan3/range/views/kmer_hash.hpp
```diff
@@ -211,6 +211,9 @@ class kmer_hash_view<urng_t>::shape_iterator
     //!\brief The sentinel type of the underlying range.
     using sentinel_t = std::ranges::sentinel_t<rng_t>;
 
+    template <typename urng2_t>
+    friend class shape_iterator;
```
joergi-w, line 215: I have 2 questions here:
1. Why do we need the friend declaration, although `shape_iterator` is public?
2. Neither in the modified test nor in the iterator test template the `shape_iterator` is used. How do you actually test the new functionality?

smehringer:
> > This PR introduces a template parameter to the comparison operators that fixes this.
> 
> I couldn't find the code for this... did you forget to commit something?

Sorry! I did not update the description! I had the templates there but it did not work because `friend`ship is not transitive. Then Rene pointed out that I do not need to alter the comparison operators if I have the constructor because the constructor allows for implicit conversion.
So now it works with just the constructor.
 
> And please check the test again. I have tried hard to understand the iterator test template and probably I have missed it, but I couldn't find where you test the shape iterator and I was also expecting to see these lines somewhere that you claimed to fix in this PR:

The iterator test template has a member that is `test_range` which I set to the `views::kmer_hash`. Now in the iterator test template I always take the begin and end of the `kmer_hash` view which returns a shape iterator. 
So everything in the tests is operating on the shape_iterator. Does this make it clearer?

include/seqan3/range/views/kmer_hash.hpp
smehringer: Good question :)
So below, I introduced a new constructor from a shape_iterator which is specialised over a different `urng_t`. This implies that the two shape_iterators do not have the same type (differnet underlying type).
In order to copy over the members, I need to access them on `rhs`, but they are declared private. That's why I have to befriend every shape_iterator regardless of its underlying `urng_t` type in order to be able to access the private members.

Regarding the tests I answered below. If anything is still not clear please ask again :)

joergi-w: Ah thank you, I think I understand now: A class with different template arguments is considered a different class and therefore private and protected member access is prohibited.

smehringer: Yes exactly

include/seqan3/range/views/kmer_hash.hpp
```diff
@@ -241,6 +244,19 @@ class kmer_hash_view<urng_t>::shape_iterator
     constexpr shape_iterator & operator=(shape_iterator &&)      = default; //!< Defaulted.
     ~shape_iterator()                                            = default; //!< Defaulted.
 
+    //!\brief Allow iterator on a const range to be constructible from an iterator over a non-const range.
+    template <typename urng2_t>
+    //!\cond
+        requires std::same_as<std::remove_const_t<urng_t>, urng2_t>
+    //!\endcond
+    shape_iterator(shape_iterator<urng2_t> const & it) :
```
rrahn, line 252:
```suggestion
    shape_iterator(shape_iterator<urng2_t> it) :
```

include/seqan3/range/views/kmer_hash.hpp
```diff
@@ -241,6 +244,19 @@ class kmer_hash_view<urng_t>::shape_iterator
...
+        hash_value{it.hash_value},
```
rrahn, line 253:
```suggestion
        hash_value{std::move(it.hash_value)},
```

rrahn: and the others

test/unit/range/views/view_kmer_hash_test.cpp
```diff
@@ -17,12 +19,31 @@
...
 
-#include <gtest/gtest.h>
+#include "../iterator_test_template.hpp"
 
 using seqan3::operator""_dna4;
 using seqan3::operator""_dna5;
 using seqan3::operator""_shape;
 
+using iterator_type = std::ranges::iterator_t<decltype(std::declval<seqan3::dna4_vector&>()
+                                                        | seqan3::views::kmer_hash(0b1001_shape))>;
```
rrahn, line 29:
```suggestion
                                                     | seqan3::views::kmer_hash(0b1001_shape))>;
```

test/unit/range/views/view_kmer_hash_test.cpp
```diff
@@ -17,12 +19,31 @@
...
+
+template <>
+struct iterator_fixture<iterator_type> : public ::testing::Test
+{
+    using iterator_tag = std::forward_iterator_tag;
```
rrahn, line 34: why is this only forward? Should it not be random access?

test/unit/range/views/view_kmer_hash_test.cpp
```diff
@@ -17,12 +19,31 @@
...
+    static constexpr bool const_iterable = true;
+
+    seqan3::dna4_vector text{"ACGTAGC"_dna4};
+
+    decltype(seqan3::views::kmer_hash(text, 0b101_shape)) test_range = seqan3::views::kmer_hash(text, 0b101_shape);
```
rrahn, line 39: you are aware that here the shape is `0b101` and in the iterator defintion `0b1001`?

test/unit/range/views/view_kmer_hash_test.cpp
smehringer: good catch, I forgot to change it back

test/unit/range/views/view_kmer_hash_test.cpp
smehringer: nope. Copy and paste error. It will not affect the type but it should be consistent anyway

test/unit/range/views/view_kmer_hash_test.cpp
smehringer: It actually revealed another bug of the kmer_hash in the `operator[]()` of the iterator. I fixed it in a new commit and altered the Changelog, since I think this semantic change is notable.

include/seqan3/range/views/kmer_hash.hpp
```diff
@@ -491,14 +507,12 @@ class kmer_hash_view<urng_t>::shape_iterator
     /*!\brief Move the iterator by a given offset and return the corresponding hash value.
      * \attention This function is only avaible if `it_t` models std::random_access_iterator.
      */
-    value_type operator[](difference_type const n)
+    value_type operator[](difference_type const n) const
```
rrahn, line 510:
```suggestion
    reference operator[](difference_type const n) const
```

smehringer:
I rebased such that the individual commits can be merged as is afterwards.

PR #1757, opened 2020-04-22
Pass result to alignment/enable alignment result type in alignment configurations

rrahn:
Part of #1692 

### Description

Adds a type definition for the alignment result type inside of the alignment configuration traits. This allows to get the result type in a centralised way that is available to all algorithms and the result type needs to be defined only in one place during the configuration phase.

~### Note for the reviewer:~

~This PR is based on #1740 and only the last commit is relevant for this change.~

### DIFF
--- a/include/seqan3/alignment/pairwise/alignment_configurator.hpp
+++ b/include/seqan3/alignment/pairwise/alignment_configurator.hpp
@@ -17,6 +17,7 @@
 #include <utility>
 #include <vector>
 
+#include <seqan3/alignment/configuration/align_config_alignment_result_capture.hpp>
 #include <seqan3/alignment/matrix/detail/alignment_score_matrix_one_column.hpp>
 #include <seqan3/alignment/matrix/detail/alignment_score_matrix_one_column_banded.hpp>
 #include <seqan3/alignment/matrix/detail/alignment_trace_matrix_full.hpp>
@@ -265,9 +266,13 @@ public:
             using alignment_result_value_t = typename align_result_selector<std::remove_reference_t<wrapped_first_t>,
                                                                             std::remove_reference_t<wrapped_second_t>,
                                                                             config_t>::type;
-            using callback_on_result_t = std::function<void(alignment_result<alignment_result_value_t>)>;
+            using alignment_result_t = alignment_result<alignment_result_value_t>;
+            using callback_on_result_t = std::function<void(alignment_result_t)>;
             // Define the function wrapper type.
             using function_wrapper_t = std::function<void(indexed_sequence_pair_chunk_t, callback_on_result_t)>;
+
+            // Capture the alignment result type.
+            auto captured_cfg = cfg | align_cfg::alignment_result_capture<alignment_result_t>;
 
             // ----------------------------------------------------------------------------
             // Test some basic preconditions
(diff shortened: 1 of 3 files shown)


rrahn:
travis is actually green but somehonw the status is not updated

include/seqan3/alignment/pairwise/detail/type_traits.hpp
```diff
@@ -24,6 +25,7 @@
 #include <seqan3/alignment/configuration/align_config_vectorise.hpp>
 #include <seqan3/alignment/matrix/trace_directions.hpp>
 #include <seqan3/alignment/pairwise/detail/concept.hpp>
+#include <seqan3/core/detail/empty_type.hpp>
 #include <seqan3/core/algorithm/configuration.hpp>
 #include <seqan3/core/bit_manipulation.hpp>
```
MitraDarja, line 30: Alphabetical order 💅

test/unit/alignment/pairwise/alignment_configurator_test.cpp
```diff
@@ -9,6 +9,7 @@
 
 #include <range/v3/view/single.hpp>
 
+#include <seqan3/alignment/pairwise/detail/type_traits.hpp>
 #include <seqan3/alignment/pairwise/alignment_configurator.hpp>
```
MitraDarja, line 13: Alphabetical order

include/seqan3/alignment/pairwise/detail/type_traits.hpp
```diff
@@ -80,6 +82,23 @@ template <typename configuration_t>
 //!\endcond
 struct alignment_configuration_traits
 {
+private:
+    //!\brief Helper function to determine the alignment result type.
+    static constexpr auto determine_alignment_result_type() noexcept
```
MitraDarja, line 87: Probably a stupid question, but why are we certain this is noexcept?

rrahn: because it is anyway just evaluated at compile time and within a constexpr context a function cannot throw. So this serves as an additional documentation, that we never expect something odd here.

include/seqan3/alignment/pairwise/alignment_configurator.hpp
```diff
@@ -265,10 +266,14 @@ struct alignment_configurator
...
                                                                             std::remove_reference_t<wrapped_second_t>,
                                                                             config_t>::type;
-            using callback_on_result_t = std::function<void(alignment_result<alignment_result_value_t>)>;
+            using alignment_result_t = alignment_result<alignment_result_value_t>;
+            using callback_on_result_t = std::function<void(alignment_result_t)>;
             // Define the function wrapper type.
             using function_wrapper_t = std::function<void(indexed_sequence_pair_chunk_t, callback_on_result_t)>;
 
+            // Capture the alignment result type.
+            auto captured_cfg = cfg | align_cfg::alignment_result_capture<alignment_result_t>;
```
marehr, line 275: I would suggest to rename this to `cfg_with_result_type`, (I saw this in the following PR)

because `align_cfg::alignment_result_capture` would be `align_cfg::result_type` with our current naming scheme

rrahn: I did rename the variable in #1760 and the config will be renamed somewhen later.

rrahn:
Mhmm, code coverage is strange. It suggests that the defaulted copy constructor of the banded score and trace matrix are not called anymore, but literally I didn't change anything in the configuration process that could indicate this. I assume a false positive here

PR #1758, opened 2020-04-23
Pass result to alignment/pass in alignment result type in two executor

rrahn:
Part of #1692

Makes the two executor more flexible by providing the actual algorithm result type from outside.
Later we could merge this with some restructering for the search execution model as well.

### Notes for the reviewer

Only the last commit is relevant

### DIFF
--- a/include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
+++ b/include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
@@ -156,7 +156,7 @@ public:
     //!\endcond
     alignment_executor_two_way(resource_t resrc,
                                alignment_algorithm_t fn,
-                               value_t buffer_value = value_t{},
+                               value_t const buffer_value = value_t{},
                                size_t chunk_size = 1u,
                                exec_policy_t const & SEQAN3_DOXYGEN_ONLY(exec) = seq) :
         kernel{std::move(fn)},
@@ -170,9 +170,9 @@ public:
         chunked_resource_it = chunked_resource.begin();
 
         if constexpr (std::same_as<execution_handler_t, execution_handler_parallel>)
-            init_buffer(std::ranges::distance(resrc), buffer_value);
+            init_buffer(std::ranges::distance(resrc), std::move(buffer_value));
         else
-            init_buffer(_chunk_size, buffer_value);
+            init_buffer(_chunk_size, std::move(buffer_value));
     }
 
     //!}
@@ -288,9 +288,9 @@ private:
      * \param[in] size The initial size of the buffer.
      * \param[in] init_value The value used to initialise the buffer.
      */
-    void init_buffer(size_t const size, value_t const & init_value)
+    void init_buffer(size_t const size, value_t const init_value)
     {
-        buffer.resize(size, init_value);
+        buffer.resize(size, std::move(init_value));
         setg(std::ranges::end(buffer), std::ranges::end(buffer));
     }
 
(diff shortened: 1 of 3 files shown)


include/seqan3/alignment/pairwise/detail/type_traits.hpp
```diff
@@ -24,6 +25,7 @@
 #include <seqan3/alignment/configuration/align_config_vectorise.hpp>
 #include <seqan3/alignment/matrix/trace_directions.hpp>
 #include <seqan3/alignment/pairwise/detail/concept.hpp>
+#include <seqan3/core/detail/empty_type.hpp>
```
svnbgnk, line 28: This include should be under core/algorithm.

test/unit/alignment/pairwise/alignment_configurator_test.cpp
```diff
@@ -9,6 +9,7 @@
 
 #include <range/v3/view/single.hpp>
 
+#include <seqan3/alignment/pairwise/detail/type_traits.hpp>
```
svnbgnk, line 12: Not needed (duplicate).

include/seqan3/alignment/pairwise/detail/type_traits.hpp
rrahn: I am not sure I understand what you mean. It is a generic type that can be used in many different contexts.

svnbgnk: Sorry for the confusion I meant it in the context of lexicographical order. So first core/algorithm... then core/detail

include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
```diff
@@ -166,9 +170,9 @@ class alignment_executor_two_way
         chunked_resource_it = chunked_resource.begin();
 
         if constexpr (std::same_as<execution_handler_t, execution_handler_parallel>)
-            init_buffer(std::ranges::distance(resrc));
+            init_buffer(std::ranges::distance(resrc), buffer_value);
```
smehringer, line 173: You take the parameter by copy so shouldn't you move here, and take the value by copy in the `init_buffer` too?

include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
```diff
@@ -153,6 +156,7 @@ class alignment_executor_two_way
     //!\endcond
     alignment_executor_two_way(resource_t resrc,
                                alignment_algorithm_t fn,
+                               value_t buffer_value = value_t{},
```
smehringer, line 159:
```suggestion
                               value_t const buffer_value = value_t{},
```
?

rrahn:
> Why is it more flexible to pass the result (type) into the executor as an extra parameter if the alignmnt_result_type is already passed on by the config in the algorithm?

Because it requires to know the layout of the underlying algorithm. But to decouple this from the alignment and use it generically for other algorithms this is the correct way. Also simpler from the interface as it is clear what happens

include/seqan3/alignment/pairwise/execution/alignment_executor_two_way.hpp
rrahn: Basically, it would be the same. So I would change it at this place. But in a more recent PR (not yet published) I'll remove this entirely. So I would like to keept it as is.

smehringer: It is really hard to review something and understand the rationale if "everything is changed in the next PR anyway". I know we asked you to split your PRs in smaller ones a while ago but I really have a hard time to keep track of what you are trying to do and that way I cannot really review the changes?

marehr: In a week you can progress a lot with a PR which is a building block for more changes. This happens and I think a "I found another solution in the meantime, but I would like to get the changes that were made in here merged, s.t. the review work was not completely in vain" a plausible reason to defer this. I think rene is trust worthy enough if he promises them to do later.

PR #1759, opened 2020-04-23
Pass result to alignment/enable alignment result detection through algorithm traits

rrahn:
Part of #1692 

Updates the alignment algorithms to use the alignment result type from the alignment configuration traits and avoid differences in the redefintion of the same template parameter.

### Note for the reviewer

Only the last commit is relevant.


marehr:
@rrahn  Why did you close this one?

PR #1764, opened 2020-04-25
 [FEATURE] new test macro EXPECT_RANGE_EQ

marehr:
Part of seqan/product_backlog#46

Add a new test macro `EXPECT_RANGE_EQ`.

It works for all ranges because we are copying them into std::vectors;
We might want to optimize this for large ranges, because the copy is not necessary if the given range is a forward range.

### DIFF
--- a/test/include/seqan3/test/expect_range_eq.hpp
+++ b/test/include/seqan3/test/expect_range_eq.hpp
@@ -1,0 +1,54 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2020, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides test utilities for std::ranges::range types.
+ * \author Marcel Ehrhardt <marcel.ehrhardt AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <gtest/gtest.h>
+
+#include <seqan3/std/algorithm>
+#include <seqan3/std/ranges>
+
+#include <seqan3/core/detail/debug_stream_range.hpp>
+#include <seqan3/core/detail/to_string.hpp>
+#include <seqan3/test/pretty_printing.hpp>
+
+namespace seqan3::test
+{
+#define EXPECT_RANGE_EQ(val1, val2) \
+    EXPECT_PRED_FORMAT2(::seqan3::test::expect_range_eq{}, val1, val2);
+
+struct expect_range_eq
+{
+    template <typename rng_t>
+    auto copy_range(rng_t && rng)
+    {
+        using value_t = std::ranges::range_value_t<rng_t>;
+        std::vector<value_t> rng_copy{};
+        std::ranges::copy(rng, std::ranges::back_inserter(rng_copy));
+        return rng_copy;
+    }
+
+    template <std::ranges::range lhs_t, std::ranges::range rhs_t>
+    ::testing::AssertionResult operator()(char const * lhs_expression, char const * rhs_expression,
+                                          lhs_t && lhs, rhs_t && rhs)
+    {
+        std::vector lhs_copy = copy_range(lhs);
+        std::vector rhs_copy = copy_range(rhs);
+
+        if (std::ranges::equal(lhs_copy, rhs_copy))
+            return ::testing::AssertionSuccess();
+
+        return ::testing::internal::CmpHelperEQFailure(lhs_expression, rhs_expression, lhs_copy, rhs_copy);
+    }
+};
+
+} // namespace seqan3::test
--- a/test/include/seqan3/test/pretty_printing.hpp
+++ b/test/include/seqan3/test/pretty_printing.hpp
@@ -22,7 +22,8 @@ namespace seqan3
 //!\cond DEV
 //!\brief Overload for the googletest PrintTo function that always delegates to our debug_stream.
 template <typename t>
-    requires true // tricks the compiler to consider this as more specialized than googletests generic PrintTo
+    requires (!std::input_or_output_iterator<t>)
+    // tricks the compiler to consider this as more specialized than googletests generic PrintTo
 void PrintTo (t const & v, std::ostream * out)
 {
     debug_stream_type my_stream{*out};
--- a/test/unit/io/record_test.cpp
+++ b/test/unit/io/record_test.cpp
@@ -15,10 +15,22 @@
 #include <seqan3/io/detail/record.hpp>
 #include <seqan3/core/concept/tuple.hpp>
 #include <seqan3/std/algorithm>
+#include <seqan3/test/expect_range_eq.hpp>
 
 using seqan3::operator""_dna4;
 
 using default_fields = seqan3::fields<seqan3::field::seq, seqan3::field::id, seqan3::field::qual>;
+
+// This is needed for EXPECT_RANGE_EQ:
+namespace seqan3
+{
+template <typename char_t>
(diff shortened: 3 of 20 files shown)


test/unit/test/expect_range_eq_test.cpp
```diff
@@ -0,0 +1,152 @@
...
+// Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+#include <gtest/gtest.h>
+#include <gtest/gtest-spi.h> // provides test utility to test google test itself
+
+#include <string_view>
+#include <seqan3/std/span>
```
eseiler, line 12:
```suggestion
#include <seqan3/std/span>
#include <string_view>
```

test/unit/io/record_test.cpp
```diff
@@ -15,11 +15,23 @@
 #include <seqan3/io/detail/record.hpp>
 #include <seqan3/core/concept/tuple.hpp>
 #include <seqan3/std/algorithm>
```
eseiler, line 17: Can you sort the `seqan3/std` includes in the files you touch?

test/include/seqan3/test/expect_range_eq.hpp
```diff
@@ -0,0 +1,54 @@
...
+    template <typename rng_t>
+    auto copy_range(rng_t && rng)
+    {
+        using value_t = std::ranges::range_value_t<rng_t>;
+        std::vector<value_t> rng_copy{};
+        std::ranges::copy(rng, std::ranges::back_inserter(rng_copy));
+        return rng_copy;
+    }
+
+    template <std::ranges::range lhs_t, std::ranges::range rhs_t>
```
eseiler, line 40: You wrote it in the PR message, but I don't think it would be much work to add another `operator()` that is constrained to `input` or `forward_ranges`?

The standard actually says that `ranges::equal` needs `input_range`s, so why would you require `forward_range`?

marehr: > You wrote it in the PR message, but I don't think it would be much work to add another `operator()` that is constrained to `input` or `forward_ranges`?

yeah it wouldn't

> The standard actually says that `ranges::equal` needs `input_range`s, so why would you require `forward_range`?

To build the message. I can call an operation on an `input_range` only once.

1) ` if (std::ranges::equal(lhs_copy, rhs_copy))`
2) `::testing::internal::CmpHelperEQFailure(lhs_expression, rhs_expression, lhs_copy, rhs_copy);`

marehr: I can omit the temporary copy if I can reiterate the same range.

test/unit/io/record_test.cpp
marehr: :place sad puppy eyes here:

Thank you for not needing to do this here :)

test/include/seqan3/test/expect_range_eq.hpp
eseiler: Hah right, you need to iterate twice.

Can you omit the copy for ranges where you can do this?

marehr: I could; Wasn't sure if we wanted this :)

eseiler: Agreed to do later

test/include/seqan3/test/expect_range_eq.hpp
```diff
@@ -0,0 +1,54 @@
...
+    ::testing::AssertionResult operator()(char const * lhs_expression, char const * rhs_expression,
+                                          lhs_t && lhs, rhs_t && rhs)
+    {
+        std::vector lhs_copy = copy_range(lhs);
+        std::vector rhs_copy = copy_range(rhs);
+
+        if (std::ranges::equal(lhs_copy, rhs_copy))
```
smehringer, line 47: I guess you cannot use the usual macros here? Otherwise `EXPECT_EQ(lhs_copy, rhs_copy)` should do the same as the three lines here no?

test/unit/io/record_test.cpp
```diff
@@ -15,11 +15,23 @@
...
 
 using seqan3::operator""_dna4;
 
 using default_fields = seqan3::fields<seqan3::field::seq, seqan3::field::id, seqan3::field::qual>;
 
+// This is needed for EXPECT_RANGE_EQ:
+namespace seqan3
+{
+template <typename char_t>
+inline debug_stream_type<char_t> & operator<<(debug_stream_type<char_t> & stream, field f)
```
smehringer, line 28: I think we claim that all our types are "debug streamable" so this should be part of the API. But it can be done in another PR (to answer your next question). Could you please add a card (or how you would prefer to track this)?

test/include/seqan3/test/pretty_printing.hpp
```diff
@@ -22,7 +22,8 @@ namespace seqan3
 //!\cond DEV
 //!\brief Overload for the googletest PrintTo function that always delegates to our debug_stream.
 template <typename t>
-    requires true // tricks the compiler to consider this as more specialized than googletests generic PrintTo
+    requires (!std::input_or_output_iterator<t>)
```
smehringer, line 25: why do you need this?

test/include/seqan3/test/expect_range_eq.hpp
smehringer, line 40: what is the compiler message if the range value types are not comparable? Should this be caught here?

test/include/seqan3/test/expect_range_eq.hpp
marehr: Yes this is copied from the function that EXPECT_EQ boils down. Using `EXPECT_EQ(lhs_copy, rhs_copy)` in here would loose the ability to see which expressions where involved, who called this and at which position this failed.

test/include/seqan3/test/pretty_printing.hpp
marehr: I wouldn't add this if we don't need this.

See https://github.com/seqan/seqan3/pull/1764/commits/ea2d5d29788959fa6340d3e20adbeebef3e6e218 for an explanation.

test/unit/io/record_test.cpp
marehr: I'm not sure, but I think not every type needs to be "debug streamable", e.g. the detail::types. I don't know if we had a resolution of what types need this. I think we only had a design goal to make every (public) seqan3 type serialisable.

However you like; Could you please add this card? This is io related and I don't know what you would expect here as an output. I just did some hack here and casted the enum as its value representation. Maybe strategy meeting?

test/include/seqan3/test/expect_range_eq.hpp
marehr: I don't know and I'm unfortunately short on time to test this right now, I think you would get some compiler error that 

```c++
if (std::ranges::equal(lhs_copy, rhs_copy))
```

can't be compiled. I expect that you will get a nice stack trace with the involved types and you will see that `"if (std::ranges::equal(lhs_copy, rhs_copy))"` failed for your given ranges. My guess is that this will be less cryptic than putting that requirement on a random `operator()`.

test/unit/io/record_test.cpp
smehringer: Yes, only the public API should be debug-streamable but fields are public. 
We can add this to the DEV meeting although it has low priority I think.

test/include/seqan3/test/pretty_printing.hpp
smehringer: I didn't suspect that you added this without a reason, I was just asking why because I didn't see where you would like to print iterators. Maybe we should change this to "requires debug-streamable" instead?

marehr: We are not able to determine in the current design whether a type is debug-streamable or not.

debug_stream says always "I can", but fails when instantiating. This and some other issues with debug-stream lead to this (incomplete) proposal / issue list https://github.com/seqan/product_backlog/issues/63.

PR #1767, opened 2020-04-27
[MISC] Remove debug requirement for score, end_position and begin_position test in pairwise alignment test template.

rrahn:
In the upcoming PRs I will step-by-step the implementation of the alignment algorithm which will run in the expected performance we'd like to see. This will enable the test for this new branch. Also it is redundant to test the trace and score matrix as they are tested once in the traceback test.


rrahn:
Yes CI fails because it needs to be rebased with the last open Commit about changing the way how this alignment result type is passed into the alignment algorithm. I can change the commit Tag to test.

PR #1761, opened 2020-04-23
Optimise alignment/part1 native global alignment

rrahn:
First part of refactoring the alignment to a more optimised version:

Old microbenchmark of global affine alignment without vectorisation:
```
----------------------------------------------------------------------------------------------
Benchmark                                    Time             CPU   Iterations UserCounters...
----------------------------------------------------------------------------------------------
seqan3_affine_dna4                      912108 ns       901837 ns          724 CUPS=278.322M/s cells=251.001k
seqan2_affine_dna4                      507343 ns       501442 ns         1293 CUPS=500.558M/s cells=251.001k

seqan3_affine_dna4_collection          3780352 ns      3737470 ns          185 CUPS=272.939M/s cells=1020.1k
seqan2_affine_dna4_collection          2544802 ns      2516513 ns          273 CUPS=405.363M/s cells=1020.1k
```

New microbenchmark of global affine alignment without vectorisation:

```
----------------------------------------------------------------------------------------------
Benchmark                                    Time             CPU   Iterations UserCounters...
----------------------------------------------------------------------------------------------
seqan3_affine_dna4                      508742 ns       501681 ns         1315 CUPS=500.32M/s cells=251.001k
seqan2_affine_dna4                      506287 ns       500708 ns         1280 CUPS=501.292M/s cells=251.001k

seqan3_affine_dna4_collection          2098579 ns      2070642 ns          332 CUPS=492.649M/s cells=1020.1k
seqan2_affine_dna4_collection          2559029 ns      2508209 ns          273 CUPS=406.705M/s cells=1020.1k
```

### TODO: Place here the macobenchmark results.

PR #1768, opened 2020-04-27
[FEATURE] Add new optimised dynamic alignment algorithm.

rrahn:
The optimised alignment implements the naive global alignment with affine gap costs without encapsulating any functionality into policies.
This is the most performant implementation that could be measured for this kind of alignment.
It only computes global alignment and is not vectorised yet.

It also will already start with the new algorithm implementation which will be step-by-step refactored and put into the correct module structure.

### Note at the reviewer:

This is a baseline implementation and the core function as it is placed there won't be there for long. But it allows to follow the refactoring process and always go back to a working version. So if not all variables have a proper name maybe that can be handled not too strict. 😄

### DIFF
--- a/include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
+++ b/include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
@@ -1,0 +1,224 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2020, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides seqan3::detail::pairwise_alignment_algorithm.
+ * \author Rene Rahn <rene.rahn AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <seqan3/std/concepts>
+#include <seqan3/std/ranges>
+
+#include <seqan3/alignment/configuration/align_config_gap.hpp>
+#include <seqan3/alignment/configuration/align_config_scoring.hpp>
+#include <seqan3/alignment/pairwise/detail/type_traits.hpp>
+#include <seqan3/alignment/scoring/gap_scheme.hpp>
+#include <seqan3/core/detail/empty_type.hpp>
+#include <seqan3/core/detail/type_inspection.hpp>
+#include <seqan3/range/views/drop.hpp>
+#include <seqan3/range/views/zip.hpp>
+
+namespace seqan3::detail
+{
+
+/*!\brief The alignment algorithm type to compute standard pairwise alignment using dynamic programming.
+ * \implements std::invocable
+ * \ingroup pairwise_alignment
+ *
+ * \tparam alignment_configuration_t The configuration type; must be of type seqan3::configuration.
+ *
+ * \details
+ *
+ * ### Configuration
+ *
+ * The first template argument is the type of the alignment configuration. The alignment configuration was used to
+ * configure the `alignment algorithm type` within the seqan3::detail::alignment_configurator.
+ * The algorithm computes a column based dynamic programming matrix given two sequences.
+ * After the computation a user defined callback function is invoked with the computed seqan3::alignment_result.
+ */
+template <typename alignment_configuration_t>
+//!\cond
+    requires is_type_specialisation_of_v<alignment_configuration_t, configuration>
+//!\endcond
+class pairwise_alignment_algorithm
+{
+private:
+    //!\brief The alignment configuration traits type with auxiliary information extracted from the configuration type.
+    using traits_type = alignment_configuration_traits<alignment_configuration_t>;
+    //!\brief The type of the scoring scheme.
+    using scoring_scheme_type =  typename traits_type::scoring_scheme_type;
+    //!\brief The configured alignment result type.
+    using alignment_result_type = typename traits_type::alignment_result_type;
+
+    static_assert(!std::same_as<alignment_result_type, empty_type>, "Alignment result type was not configured.");
+
+    //!\brief The configured scoring scheme.
+    scoring_scheme_type m_scoring_scheme{};
+    //!\brief The configured gap scheme.
+    gap_scheme<int8_t> m_gap_scheme{};
+
+public:
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    pairwise_alignment_algorithm() = default; //!< Defaulted.
+    pairwise_alignment_algorithm(pairwise_alignment_algorithm const &) = default; //!< Defaulted.
+    pairwise_alignment_algorithm(pairwise_alignment_algorithm &&) = default; //!< Defaulted.
+    pairwise_alignment_algorithm & operator=(pairwise_alignment_algorithm const &) = default; //!< Defaulted.
+    pairwise_alignment_algorithm & operator=(pairwise_alignment_algorithm &&) = default; //!< Defaulted.
+    ~pairwise_alignment_algorithm() = default; //!< Defaulted.
+
+    /*!\brief Constructs and initialises the algorithm using the alignment configuration.
+     * \param config The configuration passed into the algorithm.
+     *
+     * \details
+     *
+     * Initialises the algorithm given the user settings from the alignment configuration object.
+     */
+    pairwise_alignment_algorithm(alignment_configuration_t const & config)
+    {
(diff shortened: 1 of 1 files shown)


include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
```diff
@@ -0,0 +1,226 @@
...
+        // Initialise matrix.
+        optimal_column.clear();
+        horizontal_column.clear();
+        optimal_column.resize(sequence2.size() + 1, 0);
+        horizontal_column.resize(sequence2.size() + 1, 0);
+        int32_t diagonal{};
+        int32_t vertical{gap_open};
+
+        // Initialise the first column.
+        for (auto && [opt, hor] : views::zip(optimal_column, horizontal_column) | views::drop(1))
```
MitraDarja, line 181: I do not understand what the drop does here, are we then ignoring the first elements in optimal_column and horizontal_column, and if so why?

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
```diff
@@ -0,0 +1,226 @@
...
+    {
+        // Use gap costs locally.
+        int32_t gap_extension{m_gap_scheme.get_gap_score()};
+        int32_t gap_open{m_gap_scheme.get_gap_open_score() + gap_extension};
+
+        std::vector<int32_t> optimal_column{};
+        std::vector<int32_t> horizontal_column{};
+
+        // Initialise matrix.
+        optimal_column.clear();
```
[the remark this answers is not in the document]
rrahn: Well, not necessary here. But I also tried caching the matrix as a state of the algorithm and this seems to be a left over. I change this.

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
rrahn: Yes, I initiaist the first column. The first cell always is initialised with 0. So I just start with the next cell in the column, which is the first one that is initialised with gap costs.

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
```diff
@@ -0,0 +1,226 @@
...
+    }
+    //!\}
+
+    /*!\name Invocation
+     * \{
+     */
+    /*!\brief Computes the pairwise sequence alignment for the given range over indexed sequence pairs.
+     * \tparam indexed_sequence_pairs_t The type of indexed_sequence_pairs; must model
+     *                                  seqan3::detail::indexed_sequence_pair_range.
+     * \tparam callback_t The type of the callback function that is called with the alignment result; must model
```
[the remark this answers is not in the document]
rrahn: we do this with the semicolon in all places. At least I thought 🤔

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
```diff
@@ -0,0 +1,226 @@
...
+        {
+            result_value_t res{};
+            res.id = idx;
+            res.score = compute_matrix(get<0>(sequence_pair), get<1>(sequence_pair));
+            callback(alignment_result_type{res});
+        }
+    }
+
+protected:
+    /*!\brief Compute the actual alignment by given the two sequences.
```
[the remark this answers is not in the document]
rrahn: Is it really from? I think it should be for, but in general the sentence was odd. So I'll change it adequately.

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
```diff
@@ -0,0 +1,226 @@
...
+     * \tparam sequence1_t The type of the first sequence; must model std::ranges::forward_range.
+     * \tparam sequence2_t The type of the second sequence; std::ranges::forward_range.
```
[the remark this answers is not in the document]
rrahn: AFAIK

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
rrahn: I added a comment.

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
```diff
@@ -0,0 +1,224 @@
...
+            res.id = idx;
+            res.score = compute_matrix(get<0>(sequence_pair), get<1>(sequence_pair));
+            callback(alignment_result_type{res});
+        }
+    }
+
+protected:
+    /*!\brief Compute the actual alignment.
+     * \tparam sequence1_t The type of the first sequence; must model std::ranges::forward_range.
+     * \tparam sequence2_t The type of the second sequence; std::ranges::forward_range.
```
MitraDarja, line 157:
```suggestion
     * \tparam sequence2_t The type of the second sequence; must model std::ranges::forward_range.
```

rrahn:
Ah yes, in #1761 I opened a reference PR and split the commits into separate PRs since they are all blocked by the remaining open PRs for refactoring the alignment algorithm. The next PR will enable this branch and then it is tested as well.

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
```diff
@@ -0,0 +1,224 @@
...
+    pairwise_alignment_algorithm & operator=(pairwise_alignment_algorithm const &) = default; //!< Defaulted.
+    pairwise_alignment_algorithm & operator=(pairwise_alignment_algorithm &&) = default; //!< Defaulted.
+    ~pairwise_alignment_algorithm() = default; //!< Defaulted.
+
+    /*!\brief Constructs and initialises the algorithm using the alignment configuration.
+     * \param config The configuration passed into the algorithm.
+     *
+     * \details
+     *
+     * Initialises the algorithm given the user settings from the alignment configuration object.
```
smehringer, line 82: I would document the default gap_scheme introduced in the constructor.

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
```diff
@@ -0,0 +1,224 @@
...
+     * |space (alignment)       |\f$ O(n*m) \f$    |\f$ O(n*k) \f$     |
+     */
+    template <indexed_sequence_pair_range indexed_sequence_pairs_t, typename callback_t>
+    //!\cond
+        requires std::invocable<callback_t, alignment_result_type>
+    //!\endcond
+    void operator()(indexed_sequence_pairs_t && indexed_sequence_pairs, callback_t && callback)
+    {
+        using result_value_t = typename alignment_result_value_type_accessor<alignment_result_type>::type;
+        using std::get;
```
smehringer, line 143: Is the sequence pair something else than a std pair/tuples sometimes?

rrahn:
> As Mitra pointed out, there is no test that covers this code. Even if it serves refactoring purposes only at least a small test (and ideally a benchmark) should be added I think. I wonder why the code coverage does not cry out. Does it not cover headers that are not included anywhere? Or did I miss a test somewhere?

Yes, like I described in the last comment, this is a separate commit extracted from a bigger series of changes. The draft for this can be found here:  #1761. The changes related to this draft will enable proper tests as well as benchmarks as is visible in #1761

> Question: So the abstraction with policies was causing a runtime regression? Does this only concern the "hot loops" or the design in general? (Sorry if you told me but I don't remember)  We should keep this in mind for the search algorithm refactorisation.

Well, the point was to figure this out. The upcoming commit series will do what should have done right from the beginning. Namely, it shows code that runs and has benchmarks that show that it is fast. Then the code is refactored into policies step by step. And not making it complex with all features before testing it thoroughly.

So the result is that the abstraction with policies was and still is awesome. However in the initial design I tried to solve some issues via CRTP, which turned out to be only working because of the concepts bug of gcc. Now I try to remove this unnecessary dependency while I incrementally introduce the new abstraction to unify the alignment algorithms, by overloading existing policies. All these steps will be tested and benchmarked in micro and macro benchmarks along the way. the experiences I made are already included in the search refactoring design with policies. For example I did provide a godbolt example with this design when we had the result factory discussion for the search.

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
rrahn: I can't know. It is generic code. It could be seqan3::pod_tuple for example or any other third-library tuple implementation. So I have to call get unqualified. However, since get is a function template with an explicit template argument it must be first introduced into the scope otherwise the compiler does not know that he should actually look for a function called get and tries to evaluate the expression `get < 0 > (...)` like `get smaller than 0 greater than (expression)`, which will fail to compile.

PR #1766, opened 2020-04-27
[MISC] Remove 'using namespace seqan3' from test/unit/*.

Irallia:
Last cleanup for test/unit/*.
Resolves #1576

PR #1769, opened 2020-04-28
[FIX] Missing seqan3 namespace identifier in gap_decorator_test.cpp

wvandertoorn:
Something appears to have gone wrong in the removal of  `using namespace seqan3` from `test/unit/range/gap_decorator_test.cpp`. The seqan3 namespace identifier were missing. This PR adds them.

Edit (@eseiler):
The `gap_decorator_test` has `#include "../../alignment/aligned_sequence_test_template.hpp"`, this file had an `using namespace seqan3;`. 
So in the `unit/range` PR (#1766) , the gap_decorator` still worked because it had an `using namespace` in an include. 
With the `unit/alignment` PR (#1749) this was removed, but because the latter PR (#1766) was not rebased on master, the error did not occur in the PR but the following master branch build

### DIFF
--- a/test/unit/range/decorator/gap_decorator_test.cpp
+++ b/test/unit/range/decorator/gap_decorator_test.cpp
@@ -21,12 +21,13 @@
 #include "../iterator_test_template.hpp"
 #include "../../alignment/aligned_sequence_test_template.hpp"
 
-using decorator_t = gap_decorator<std::vector<dna4> const &>;
+using decorator_t = seqan3::gap_decorator<std::vector<seqan3::dna4> const &>;
 
-std::vector<dna4> const dummy_obj{}; // dummy lvalue for type declaration of views
-using decorator_t2 = gap_decorator<
+std::vector<seqan3::dna4> const dummy_obj{}; // dummy lvalue for type declaration of views
+using decorator_t2 = seqan3::gap_decorator<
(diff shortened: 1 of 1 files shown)


test/unit/range/decorator/gap_decorator_test.cpp
```diff
@@ -21,12 +21,13 @@
...
+using decorator_t = seqan3::gap_decorator<std::vector<seqan3::dna4> const &>;
 
-std::vector<dna4> const dummy_obj{}; // dummy lvalue for type declaration of views
-using decorator_t2 = gap_decorator<
+std::vector<seqan3::dna4> const dummy_obj{}; // dummy lvalue for type declaration of views
+using decorator_t2 = seqan3::gap_decorator<
                          decltype(std::ranges::subrange<decltype(dummy_obj.begin()),
-                                                 decltype(dummy_obj.begin())>{dummy_obj.begin(), dummy_obj.end()})>;
+                                                        decltype(dummy_obj.begin())>{dummy_obj.begin(),
+                                                                                     dummy_obj.end()})>;
```
joergi-w, line 30: Here it seems beneficial to define a type
```cpp
using iter_t = decltype(dummy_obj.begin());
```
before to shorten this. However, this is just an aesthetical suggestion. 💅

eseiler:
Coverage is probably an artifact from failing master-branch reference builds

PR #1770, opened 2020-04-28
[MISC] Remove specific alignment constraints from execution handler.

rrahn:
As one step towards an uniform execution design for different algorithms this commit removes the constraints on the sequence types that need to be passed to the alignment, because other algorithms might have different requirements on this type.

### DIFF
--- a/include/seqan3/alignment/pairwise/execution/execution_handler_parallel.hpp
+++ b/include/seqan3/alignment/pairwise/execution/execution_handler_parallel.hpp
@@ -99,15 +99,14 @@ public:
 
     /*!\brief Takes underlying range of sequence pairs and invokes an alignment on each instance.
      * \tparam algorithm_t              The type of the alignment algorithm.
-     * \tparam indexed_sequence_pairs_t The type of underlying sequence pairs annotated with an index;
-     *                                  must model seqan3::detail::indexed_sequence_pair_range.
+     * \tparam indexed_sequence_pairs_t The type of underlying sequence pairs annotated with an index.
      * \tparam delegate_type            The type of the callable invoked on the std::invoke_result of `algorithm_t`.
      *
      * \param[in] algorithm              The alignment algorithm to invoke.
      * \param[in] indexed_sequence_pairs The range of underlying annotated sequence pairs to be aligned.
      * \param[in] delegate               A callable which will be invoked on each result of the computed alignments.
      */
-    template <typename algorithm_t, indexed_sequence_pair_range indexed_sequence_pairs_t, typename delegate_type>
+    template <typename algorithm_t, typename indexed_sequence_pairs_t, typename delegate_type>
     void execute(algorithm_t && algorithm,
                  indexed_sequence_pairs_t indexed_sequence_pairs,
                  delegate_type && delegate)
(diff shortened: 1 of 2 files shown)


include/seqan3/alignment/pairwise/execution/execution_handler_parallel.hpp
```diff
@@ -99,15 +99,14 @@ class execution_handler_parallel
 
     /*!\brief Takes underlying range of sequence pairs and invokes an alignment on each instance.
      * \tparam algorithm_t              The type of the alignment algorithm.
```
marehr, line 101: Comment like this will be changed later?

include/seqan3/alignment/pairwise/execution/execution_handler_parallel.hpp
```diff
@@ -99,15 +99,14 @@ class execution_handler_parallel
...
-     *                                  must model seqan3::detail::indexed_sequence_pair_range.
+     * \tparam indexed_sequence_pairs_t The type of underlying sequence pairs annotated with an index.
      * \tparam delegate_type            The type of the callable invoked on the std::invoke_result of `algorithm_t`.
      *
      * \param[in] algorithm              The alignment algorithm to invoke.
      * \param[in] indexed_sequence_pairs The range of underlying annotated sequence pairs to be aligned.
      * \param[in] delegate               A callable which will be invoked on each result of the computed alignments.
      */
-    template <typename algorithm_t, indexed_sequence_pair_range indexed_sequence_pairs_t, typename delegate_type>
+    template <typename algorithm_t, typename indexed_sequence_pairs_t, typename delegate_type>
```
marehr, line 109: Shouldn't the name change? Like `data_t`, `work_data_t`, `algorithm_data_t`?

Is the assumption that you move in the data still valid?

include/seqan3/alignment/pairwise/execution/execution_handler_parallel.hpp
rrahn: yes this will be changed. The plan would be to move it to core as well, since it is used by other algorithms. When moving I will also update the documentation. This is just a separate commit for the first part that generalises the two way executor.

include/seqan3/alignment/pairwise/execution/execution_handler_parallel.hpp
rrahn: Honestly, all these questions can be answered when we know the sructure and the data of the search algorithm. Also then it is the question if we can define a common requirement on the type. Or if this has to be a variadic template etc.

rrahn:
> So if I understand this correctly, the algorithms like pairwise_align and search are still enforcing their input, but the execution handler are agnostic on the input data for the algorithms?

I am not sure I understand you correctly. But I try to explain the rational in more detail. So far the execution handler were only coupled with the alignment so I thought it is a good idea to make it strict to the input. But of course, this is a more generic encapsulation and should basically know nothing about the internal types required by the algorithm. The next PR will then also remove the chunking within the two way executor and will just work on the given range. This is in my opiniton the better design as it also fixes multiple issues that would occur sooner or later. It then allows to deal with algorithm calls that do not produce a result (e.g. because the min score was not reached or the there is no hit for the query). So whether or not the input is chunked will then be handled from outside the executor and the executor does not care about the structure of the input. This generalises the use for the search and the alignment. Does this answer your question?

include/seqan3/alignment/pairwise/execution/execution_handler_parallel.hpp
rrahn: But for now it still exists within the context of alignments. But I wanted to separate the structural changes from the documentation/location changes.

Issue #1502, opened 2020-01-17
gcc-10 -std=c++17 support


marehr:
Keeps track of all changes for gcc10

Bugfixes:
* [x] [MISC] prepare span header for gcc10 (#1478)
* [x] [MISC] scoring scheme: concept template paramater can't be constrained (#1539)
* [x] [FIX] trace_iterator_base: use same constraints (#1538)
* [x] [FIX] alignment matrix tests (#1537)
* [x] [MISC] remove restraints on comparison operators of small string (#1536)
* [x] [FIX] enums shouldn't be used with same_as (#1535)
* [x] [FIX] add ; after requires expression (#1534)
* [x] [MISC] gap_scheme and static_band needs stdexcept (#1533)
* [x] [FIX] forward declare must use the same declaration (#1532)
* [x] [TEST] Don't static_assert concepts (#1531)
* [x] [TEST] two_dimensional_matrix_test use original fix again (#1530)
* [x] [FIX] break up cyclic include of std/range and std/iterator (#1529 )

Error Classes:

* [x] BUG_ALIGNED_SEQUENCE_TEST_TEMPLATE
* [x] BUG_ALPHABET_ANY
* [x] BUG_ALPHABET_TUPLE_BASE
* [x] BUG_CONCATENATED_SEQUENCES
* [x] BUG_CONCATENATED_SEQUENCES_BITCOMPRESSED_CONTAINER
* [x] BUG_RANGE_EMPLACE
* [x] BUG_RANGE_META_IS_BASE_OF
* [x] BUG_TUPLE_LIKE seems to be also caused by BUG_ALPHABET_TUPLE_BASE

<details><summary>See overview of affected test cases</summary>

**BUG_ALIGNED_SEQUENCE_TEST_TEMPLATE**:
  * seqan3_test(aligned_sequence_test.cpp)

**BUG_ALPHABET_ANY**:
  * seqan3_test(semialphabet_any_test.cpp)

**BUG_ALPHABET_TUPLE_BASE**:
  * seqan3_test(alphabet_tuple_base_test.cpp)
  * seqan3_test(composite_integration_test.cpp)
  * seqan3_test(qualified_test.cpp)
  * seqan3_test(structured_aa_test.cpp)
  * seqan3_test(structured_rna_test.cpp)
  * seqan3_test(debug_stream_alphabet_test.cpp)

**BUG_CONCATENATED_SEQUENCES**:
  * seqan3_test(debug_stream_container_of_container_test.cpp)

[...]

PR #1771, opened 2020-04-28
[FIX] transfer_template_args_onto and transfer_template_vargs_onto should not instantiate every type!

marehr:
Part of #1502

### DIFF
--- a/include/seqan3/core/type_traits/template_inspection.hpp
+++ b/include/seqan3/core/type_traits/template_inspection.hpp
@@ -59,7 +59,7 @@ template <template <typename ...> typename source_template,
 //!\cond
     requires requires ()
     {
-        std::declval<target_template<source_arg_types...>>();
+        typename target_template<source_arg_types...>;
     }
 //!\endcond
 struct transfer_template_args_onto<source_template<source_arg_types...>, target_template>
@@ -108,7 +108,7 @@ template <template <auto ...> typename source_template,
 //!\cond
     requires requires ()
     {
-        std::declval<target_template<source_varg_types...>>();
+        typename target_template<source_varg_types...>;
     }
 //!\endcond
 struct transfer_template_vargs_onto<source_template<source_varg_types...>, target_template>
--- a/test/unit/core/type_traits/template_inspection_test.cpp
+++ b/test/unit/core/type_traits/template_inspection_test.cpp
@@ -13,6 +13,12 @@
 #include <seqan3/core/type_traits/template_inspection.hpp>
 #include <seqan3/core/type_list/type_list.hpp>
 
+template <std::integral t>
+struct constraint_bar
+{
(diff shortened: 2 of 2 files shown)


include/seqan3/core/type_traits/template_inspection.hpp
```diff
@@ -59,7 +59,7 @@ template <template <typename ...> typename source_template,
 //!\cond
     requires requires ()
     {
-        std::declval<target_template<source_arg_types...>>();
+        typename target_template<source_arg_types...>;
```
smehringer, line 62: why is the `typename` needed?

test/unit/core/type_traits/template_inspection_test.cpp
```diff
@@ -34,6 +40,29 @@ TEST(template_inspect, transfer_template_args_onto_t)
     EXPECT_TRUE((std::is_same_v<t, std::tuple<int, char, double>>));
 }
 
+TEST(template_inspect, transfer_template_args_onto_with_constraint)
+{
+    // This test ensures that transfer_template_args_onto uses internally only
+    // type declarations and not type instantiations
+
+    using bar_char = constraint_bar<char>; // this is fine, even though it contains a static_assert
```
smehringer, line 48: how? Is char cast to int or does the static_assert only hit if you instantiate bar_char?

marehr: yeah that's the point :)

* `constraint_bar<char>{}` would static_assert.
* `std::declval<constraint_bar<char>>()` would static_assert.
* `using bar_char = constraint_bar<char>;` does not static_assert
* `typename constraint_bar<char>;` in concepts does not static_assert

 the expression `using bar_char = constraint_bar<char>;` just (forward) declares and does not evaluate the class body, so the `static_assert` will not be evaluated.

include/seqan3/core/type_traits/template_inspection.hpp
rrahn: Yes good question. If I see it correctly, then we just wanted to know before whether the type `target_template` substituted with the given `source_arg_types`  is actual a valid type definition. But in the concept defintion this can be expressed explicitly as a type requirement. Here is the excerpt of [cppreference](https://en.cppreference.com/w/cpp/language/constraints):

>A type requirement is the keyword typename followed by a type name, optionally qualified. The requirement is that the named type is valid: this can be used to verify that a certain named nested type exists, or that a class template specialization names a type, or that an alias template specialization names a type. A type requirement naming a class template specialization does not require the type to be complete.

test/unit/core/type_traits/template_inspection_test.cpp
rrahn: Ah, here it is explained as well. 👍

PR #1774, opened 2020-04-28
[FIX] a better fix for concatenated sequences?!

marehr:
This could have been the fix for #1731, but I just recently saw this
technique in the standard library. This solution should be a lot less
janky than the previous one. I tested it with gcc7, 8, 9 and 10 and it
worked without a problem.

I wanted to finish #1549 and after changing the definition of
`dimension_v` the previous construction of #1731 broke, so had to
rethink a solution for this.

### DIFF
--- a/include/seqan3/range/container/concatenated_sequences.hpp
+++ b/include/seqan3/range/container/concatenated_sequences.hpp
@@ -127,7 +127,7 @@ template <typename inner_type,
 //!\cond
     requires reservible_container<std::remove_reference_t<inner_type>> &&
              reservible_container<std::remove_reference_t<data_delimiters_type>> &&
-             std::is_same_v<std::ranges::range_size_t<inner_type>, value_type_t<data_delimiters_type>>
+             std::is_same_v<std::ranges::range_size_t<inner_type>, std::ranges::range_value_t<data_delimiters_type>>
 //!\endcond
 class concatenated_sequences
 {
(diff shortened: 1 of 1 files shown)


marehr:
You are right instead of having `true/false` this now is `true/undefined`. To be more concrete, undefined is a substitution failure and concepts can handle them. There is paradigm in C++ that is called SFINAE ([Substitution failure is not an error](https://en.wikipedia.org/wiki/Substitution_failure_is_not_an_error)) and we use that in this situation.

include/seqan3/range/container/concatenated_sequences.hpp
```diff
@@ -185,69 +185,41 @@ class concatenated_sequences
     //!\cond
     // unfortunately we cannot specialise the variable template so we have to add an auxiliary here
     template <std::ranges::range t>
-        requires std::convertible_to<std::ranges::range_reference_t<t>, std::ranges::range_value_t<value_type>>
-    static constexpr bool is_compatible_with_value_type_aux = dimension_v<t> == dimension_v<value_type>;
+    static constexpr bool is_compatible_with_value_type_aux(std::type_identity<t>)
+    {
+        return dimension_v<t> == dimension_v<value_type>&&
```
rrahn, line 190:
```suggestion
        return dimension_v<t> == dimension_v<value_type> &&
```

marehr:
I had to rebase to resolve a merge conflict.

PR #1775, opened 2020-04-29
[TEST] remove test of seqan3::size_type on iterators

marehr:
We defined seqan3::size_type on iterators, but it actually makes no sense.
The standard defines `std::ranges::range_size_t` if the range is sized.

We don't use it internally except for tests, so removing the test makes
it possible to deprecate it (which is the task #1549).


rrahn:
restarted jenkins. Will merge afterwards

PR #1779, opened 2020-04-30
[MISC] Add a proxy type for an affine matrix cell.

rrahn:
Adds a wrapper type for the affine cell type in the new alignment algorithm. The simplified score matrix will return a transformed zip view over the different score matrices needed to compute the affine gap cost (optimal score, horizontal score and vertical score). The new matrix does not add any perfromance penalties but the tuple interface is hard to use inside of the algorithm. Therefor, the proxy just inherits from the actual tuple type and adds named accessor functions that can be used inside of the alignment. 

You can see the usage in the reference implementation in the draft PR #1761.

### DIFF
--- a/include/seqan3/alignment/matrix/detail/affine_cell_proxy.hpp
+++ b/include/seqan3/alignment/matrix/detail/affine_cell_proxy.hpp
@@ -1,0 +1,117 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2020, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides seqan3::detail::affine_cell_proxy.
+ * \author Rene Rahn <rene.rahn AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <tuple>
+
+#include <seqan3/core/tuple_utility.hpp>
+
+namespace seqan3::detail
+{
+
+/*!\brief A wrapper for an affine score matrix cell.
+ * \implements seqan3::tuple_like
+ * \ingroup pairwise_alignment
+ *
+ * \details
+ *
+ * This wrapper provides a uniform access to the different elements of the cell within an affine score matrix. This
+ * includes the optimal score, the horizontal gap score and the vertical gap score.
+ */
+template <tuple_like tuple_t>
+//!\cond
+    requires std::tuple_size_v<tuple_t> == 3
+//!\endcond
+class affine_cell_proxy : public tuple_t
+{
+public:
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    affine_cell_proxy()= default; //!< Defaulted.
+    affine_cell_proxy(affine_cell_proxy const &) = default; //!< Defaulted.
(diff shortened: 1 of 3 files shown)


include/seqan3/alignment/matrix/detail/affine_cell_proxy.hpp
```diff
@@ -0,0 +1,102 @@
...
+namespace seqan3::detail
+{
+
+/*!\brief A wrapper for an affine score matrix cell.
+ * \implements seqan3::tuple_like
+ * \ingroup pairwise_alignment
+ *
+ * \details
+ *
+ * This wrapper provides a uniform access to the different elements of the cell within an affine score matrix. This includes
```
Irallia, line 28: This line got too wide.

smehringer: still open?

include/seqan3/alignment/matrix/detail/affine_cell_proxy.hpp
```diff
@@ -0,0 +1,108 @@
...
+ */
+template <tuple_like tuple_t>
+class affine_cell_proxy : public tuple_t
+{
+public:
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    //!\brief Defaulted.
+    affine_cell_proxy() noexcept(std::is_nothrow_default_constructible_v<tuple_t>)= default;
```
smehringer, line 39: I think the noexcept is added by the compiler by default if possible (which will include the tuple_t check I guess)?

include/seqan3/alignment/matrix/detail/affine_cell_proxy.hpp
```diff
@@ -0,0 +1,108 @@
...
+/*!\brief A wrapper for an affine score matrix cell.
+ * \implements seqan3::tuple_like
+ * \ingroup pairwise_alignment
+ *
+ * \details
+ *
+ * This wrapper provides a uniform access to the different elements of the cell within an affine score matrix. This includes
+ * the optimal score, the horizontal gap score and the vertical gap score.
+ */
+template <tuple_like tuple_t>
```
smehringer, line 31: Can you explain the rationale to me why this is not a fixed tuple type?

test/unit/alignment/matrix/detail/affine_cell_proxy_test.cpp
```diff
@@ -0,0 +1,92 @@
...
+    using cell_t = seqan3::detail::affine_cell_proxy<base_t>;
+
+    int optimal{4};
+    int horizontal{-1};
+    int vertical{10};
+
+    cell_t affine_cell{optimal, horizontal, vertical};
+    cell_t const const_affine_cell{affine_cell};
+};
+
```
smehringer, line 24: I think the concept check `tuple_like` is missing?

include/seqan3/alignment/matrix/detail/affine_cell_proxy.hpp
rrahn: basically you are right. I just changed it yesterday after I was reading cppreference code where they explicitly document it like this. So never change to late stuff that is working :)

test/unit/alignment/matrix/detail/affine_cell_proxy_test.cpp
rrahn: well I am not explicitly needing it, but I can add the check anyhow

include/seqan3/alignment/matrix/detail/affine_cell_proxy.hpp
rrahn: Because it will be a proxy around the type that is returned from the zip view or the algorithm, which are different tuple types. So it just behaves like the type it inherits from also capturing the correct tuple element types but offers in this case additional interfaces to make reading the code and accessing the values more understandable.

PR #1780, opened 2020-04-30
[FEATURE] Add simplified single column score matrix for affine gap costs.

rrahn:
Adds a simplified implementation of the single column score matrix for affine gap costs.
It removes this complicated nested CRTP design of the first version, and seems to be optimised quite well by the compiler. The matrix allows an input range interface over the columns of the matrix using a simple counted iterator mechanism. It retuns a zipped view over the three columns for the optimal score, the horizontal sore and the vertical score. To allow a better access, they are tranformed to a affine_cell_proxy_type (#1779) which gives better names for accessing the stored tuple values.

You can see the usage in the reference implementation in the draft PR #1761.


include/seqan3/alignment/matrix/detail/score_matrix_single_column.hpp
```diff
@@ -0,0 +1,246 @@
...
+/*!\brief Score matrix for the pairwise alignment using only a single column.
+ * \ingroup alignment_matrix
+ * \implements std::ranges::input_range
+ *
+ * \tparam score_t The type of the score.
+ *
+ * \details
+ *
+ * In many cases it is sufficient to store only a single score column to compute the alignment between two sequences.
+ * Since the alignment is computed iteratively column by column the same memory can be reused for the next score.
```
eseiler, line 34:
```suggestion
 * Since the alignment is computed iteratively column by column, the same memory can be reused for the next score.
```

include/seqan3/alignment/matrix/detail/score_matrix_single_column.hpp
```diff
@@ -0,0 +1,246 @@
...
+ * This score matrix scores a column for the optimal score and the horizontal score and uses a virtual column that
+ * stores only one value of type `score_t` for the vertical column. Hence, this matrix can only be used for a column
```
eseiler, line 36:
```suggestion
 * This score matrix stores the complete column for both the optimal and horizontal score, but only stores a single 
 * value for the vertical column. Hence, this matrix can only be used for a column
```

I think the sentence is a bit convoluted and hence hard to understand. The concept of a `virtual column` and the type of the value stored are not too important here?

include/seqan3/alignment/matrix/detail/score_matrix_single_column.hpp
```diff
@@ -0,0 +1,246 @@
...
+     *
+     * \param[in] sequence1 The first sequence.
+     * \param[in] sequence2 The second sequence.
+     */
+    template <std::ranges::forward_range sequence1_t, std::ranges::forward_range sequence2_t>
+    void reset_matrix(sequence1_t && sequence1, sequence2_t && sequence2)
+    {
+        size_t number_of_rows = std::ranges::distance(sequence2) + 1;
+        m_optimal_column.resize(number_of_rows, 0);
+        m_horizontal_column.resize(number_of_rows, 0);
```
eseiler, line 91: Either the function has a bad name and/or brief, or the `resize` here is wrong, or assumes something that is not written in the docs.

If optimal/horizontal column contains values, resizing them won't reset the vector.
```
std::vector<int> vec{1,2,3,4};
vec.resize(6, 0); // 1,2,3,4,0,0
```

And if I am pedantic: This function may or may not reallocate the memory depending on the capacity?

include/seqan3/alignment/matrix/detail/score_matrix_single_column.hpp
```diff
@@ -0,0 +1,246 @@
...
+    std::ranges::default_sentinel_t end() const = delete;
+    //!\}
+};
+
+/*!\brief Score matrix for the pairwise alignment using only a single column.
+ * \implements std::input_iterator
+ *
+ * \details
+ *
+ * Implements a counted iterator to simulate the iteration over the actual matrix. When dereferencing the
```
eseiler, line 123:
```suggestion
 * Implements a counted iterator to simulate the iteration over the actual matrix. When dereferenced, the
```

include/seqan3/alignment/matrix/detail/score_matrix_single_column.hpp
```diff
@@ -0,0 +1,246 @@
...
+ * iterator returns a view over the allocated memory for the respective columns. The returned view is zips
```
eseiler, line 124:
```suggestion
 * iterator returns a view over the allocated memory of the respective columns. The returned view zips
```

include/seqan3/alignment/matrix/detail/score_matrix_single_column.hpp
```diff
@@ -0,0 +1,246 @@
...
+ * the three columns into a single range and transforms the returned tuple with the references to a
```
eseiler, line 125:
```suggestion
 * the three columns into a single range and transforms the returned tuple to a
```

I know you want to describe `physical_column_t &`, etc. but I am not sure this is really important here - It's also hard to understand unless you see the code a few lines further down.

include/seqan3/alignment/matrix/detail/score_matrix_single_column.hpp
```diff
@@ -0,0 +1,246 @@
...
+    //!\brief The transform adaptor to convert the tuple from the zip view into a seqan3::detail::affine_cell_type.
+    static constexpr auto transform_to_affine_cell = std::views::transform([] (auto && tpl)
+    {
+        using fwd_tuple_t = decltype(tpl);
+        return affine_cell_proxy<remove_cvref_t<fwd_tuple_t>>{std::forward<fwd_tuple_t>(tpl)};
+    });
+
+    //!\brief The pointer to the underlying matrix.
+    score_matrix_single_column * m_host_ptr{nullptr};
+    //!\brief The current column stored as a position.
```
eseiler, line 148:
```suggestion
    //!\brief The current column index.
```

?

include/seqan3/alignment/matrix/detail/score_matrix_single_column.hpp
```diff
@@ -0,0 +1,246 @@
...
+
+    /*!\name Arithmetic operators
+     * \{
+     */
+    //!\brief Move `this` to the next column.
+    matrix_iterator & operator++()
+    {
+        ++m_current_column;
+        return *this;
+    }
```
eseiler, line 208: Wouldn't you need to also increment `m_host_ptr`?
Otherwise you always return the same column?

include/seqan3/alignment/matrix/detail/score_matrix_single_column.hpp
rrahn: ok, yes it might be a little confusing about what should happen. I updated the documentation accordingly.

include/seqan3/alignment/matrix/detail/score_matrix_single_column.hpp
rrahn: That's the semantics of this implementation. It only fakes a real matrix but in reality it only stores a single column that will be used as the next column. So the algorithm is basically overwriting the old column, since we don't need access to it anymore. At least the algorithm knows about it and this is strongly coupled with the algorithm

include/seqan3/alignment/matrix/detail/score_matrix_single_column.hpp
```diff
@@ -76,21 +76,29 @@ class score_matrix_single_column
     ~score_matrix_single_column() = default; //!< Defaulted.
     //!\}
 
-    /*!\brief Resets the matrix and reallocates the memory of the underlying matrix.
+    /*!\brief Resets the matrix dimensions given by two the sequences.
```
eseiler, line 79:
```suggestion
    /*!\brief Resets the matrix dimensions given by two sequences.
```

include/seqan3/alignment/matrix/detail/score_matrix_single_column.hpp
```diff
@@ -76,21 +76,29 @@ class score_matrix_single_column
...
      * \param[in] sequence1 The first sequence.
      * \param[in] sequence2 The second sequence.
+     *
+     * \details
+     *
+     * Resizes the optimal and the horizontal score column to the new dimensions given by the sizes of the
+     * two sequences. The number of columns is set to the size of sequence1 plus 1 for the initialisation and
+     * the number of rows is set to the size of sequence2 plus 1 for the initialisation as well.
+     * Reallocation happens only if the new column size exceeds the current capacity of the optimal and horizontal
+     * score column. The underlying vectors won't be cleared during the reset.
```
eseiler, line 92:
```suggestion
     * score column. The underlying vectors will not be cleared during the reset.
```

include/seqan3/alignment/matrix/detail/score_matrix_single_column.hpp
smehringer: still open?

include/seqan3/alignment/matrix/detail/score_matrix_single_column.hpp
```diff
@@ -0,0 +1,254 @@
...
+ *
+ * ### Range interface
+ *
+ * The matrix offers a input range interface over the columns of the matrix. Dereferencing the iterator will return
+ * another range which represents the actual score column in memory. The returned range is a
+ * transformed seqan3::views::zip view over the optimal, horizontal and vertical column. The reference type of this
+ * view is the seqan3::detail::affine_cell_proxy, which offers a practical interface to access the value of the
+ * optimal, horizontal and vertical value of the underlying matrices.
+ */
+template <typename score_t>
```
smehringer, line 47: should be arithmetic I guess?

include/seqan3/alignment/matrix/detail/score_matrix_single_column.hpp
```diff
@@ -0,0 +1,254 @@
...
+    {
+        return std::ranges::default_sentinel;
+    }
+
+    //!\brief This score matrix is not const-iterable.
+    std::ranges::default_sentinel_t end() const = delete;
+    //!\}
+};
+
+/*!\brief Score matrix for the pairwise alignment using only a single column.
```
smehringer, line 126:
```suggestion
/*!\brief Score matrix iterator for the pairwise alignment using only a single column.
```

include/seqan3/alignment/matrix/detail/score_matrix_single_column.hpp
```diff
@@ -0,0 +1,254 @@
...
+    //!\}
+
+    /*!\name Access operators
+     * \{
+     */
+    //!\brief Returns the range over the current column.
+    reference operator*() const
+    {
+        return views::zip(m_host_ptr->m_optimal_column, m_host_ptr->m_horizontal_column, m_host_ptr->m_vertical_column)
+             | transform_to_affine_cell;
```
smehringer, line 204:
```suggestion
               | transform_to_affine_cell;
```

include/seqan3/alignment/matrix/detail/score_matrix_single_column.hpp
```diff
@@ -0,0 +1,254 @@
...
+    static constexpr auto transform_to_affine_cell = std::views::transform([] (auto && tpl)
+    {
+        using fwd_tuple_t = decltype(tpl);
+        return affine_cell_proxy<remove_cvref_t<fwd_tuple_t>>{std::forward<fwd_tuple_t>(tpl)};
+    });
+
+    //!\brief The pointer to the underlying matrix.
+    score_matrix_single_column * m_host_ptr{nullptr};
+    //!\brief The current column index.
+    size_t m_current_column{};
```
smehringer, line 157:
```suggestion
    size_t current_column{};
```
Or will the naming change somwhat later?
(same for the other members, for me the `m_` prefix everywhere does not help much)

test/unit/alignment/matrix/detail/score_matrix_single_column_test.cpp
```diff
@@ -0,0 +1,61 @@
...
+            EXPECT_EQ(actual_cell.optimal_score(), std::get<0>(expected_cell));
+            EXPECT_EQ(actual_cell.horizontal_score(), std::get<1>(expected_cell));
+            EXPECT_EQ(actual_cell.vertical_score(), std::get<2>(expected_cell));
+        }
+    }
+};
+
+INSTANTIATE_TYPED_TEST_SUITE_P(score_matrix_single_column_test,
+                               iterator_fixture,
+                               matrix_iterator_t, );
```
smehringer, line 61:
```suggestion
INSTANTIATE_TYPED_TEST_SUITE_P(score_matrix_single_column_test, iterator_fixture, matrix_iterator_t, );
```

include/seqan3/alignment/matrix/detail/score_matrix_single_column.hpp
rrahn: I think it is aligned with the guide about piping views such that all views start on the same column.

include/seqan3/alignment/matrix/detail/score_matrix_single_column.hpp
rrahn: Yes, it helps a lot in many cases. And there is no rule about it. So we need a rule. That's my proposal. :)

test/unit/alignment/matrix/detail/score_matrix_single_column_test.cpp
rrahn: I feel the wrapping helps in this expression to see what is what? 🤷 But I don't care much

include/seqan3/alignment/matrix/detail/score_matrix_single_column.hpp
rrahn: Mhmm strange. I thought i added it to the batch 🤷 I do it now.

PR #1785, opened 2020-05-04
[MISC] Cleanup for the refactoring task: Remove 'using namespace seqan3'

Irallia:
This should be the last PR for this topic.
Found and removed:
- One comment in the `include` directory,
- some last `using namespace seqan3::test` in `test/performance`,
- And one `using namespace seqan3::detail` in `test/snippet`.

(Resolves ... Is there any issue for the whole task?)

### DIFF
--- a/include/seqan3/core/concept/cereal.hpp
+++ b/include/seqan3/core/concept/cereal.hpp
@@ -117,7 +117,6 @@ SEQAN3_CONCEPT cereal_text_archive = false;
  *
  * ```
  * #include <seqan3/core/concept/cereal.hpp>
- * using namespace seqan3;
  *
  * // fundamental types are serialisable
  * static_assert(cerealisable<int>);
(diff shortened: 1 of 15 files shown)


include/seqan3/core/concept/cereal.hpp
```diff
@@ -117,7 +117,6 @@ SEQAN3_CONCEPT cereal_text_archive = false;
  *
  * ```
  * #include <seqan3/core/concept/cereal.hpp>
- * using namespace seqan3;
```
smehringer, line 120: If you remove this here you also need to add `seqan3::` to the crealisable concept.
This should actually be a real snippet. 
Could maybe remove this change and factor the whole thing out in a follow up PR?

Irallia: Yes, you are absolutely right. I wasn't really looking. I have now adjusted the snippet. Since this changes only 3 more lines, I would leave it in this PR.

PR #1791, opened 2020-05-06
Optimise alignment/part2 optimum tracker policy

rrahn:
In the new alignment implementation we let the alignment algorithm decide when to check for the new alignment optimum. This removes the dependency between the recursion policy and the tracker policy as we used it before. This in turn allows us to remove the CRTP design from the alignment algorithm. The performance is still very competitive as the microbenchmark shows below. My macrobenchmarks on this design confirmed this as well.

This includes only the pre-work and does only work for the global alignment so far. My goal is to get as quickly as possible all minimal requirements into the new alignment design to provide a working vectorised version which has the optimal runtime of SeqAn2. For this I need to provide a special tracker that overloads the current one in order to check the end points for sequences that have not the same size in the batch. This is a general problem with our inter-sequence vectorisation design and my initial benchmarks show for the new version a 1.5x speedup over the implementation in SeqAn2.

Old microbenchmark of global affine alignment without vectorisation:
```
----------------------------------------------------------------------------------------------
Benchmark                                    Time             CPU   Iterations UserCounters...
----------------------------------------------------------------------------------------------
seqan3_affine_dna4                      912108 ns       901837 ns          724 CUPS=278.322M/s cells=251.001k
seqan2_affine_dna4                      507343 ns       501442 ns         1293 CUPS=500.558M/s cells=251.001k

seqan3_affine_dna4_collection          3780352 ns      3737470 ns          185 CUPS=272.939M/s cells=1020.1k
seqan2_affine_dna4_collection          2544802 ns      2516513 ns          273 CUPS=405.363M/s cells=1020.1k
```

New microbenchmark of global affine alignment without vectorisation:

```
----------------------------------------------------------------------------------------------
Benchmark                                    Time             CPU   Iterations UserCounters...
----------------------------------------------------------------------------------------------
seqan3_affine_dna4                      508742 ns       501681 ns         1315 CUPS=500.32M/s cells=251.001k
seqan2_affine_dna4                      506287 ns       500708 ns         1280 CUPS=501.292M/s cells=251.001k

seqan3_affine_dna4_collection          2098579 ns      2070642 ns          332 CUPS=492.649M/s cells=1020.1k
seqan2_affine_dna4_collection          2559029 ns      2508209 ns          273 CUPS=406.705M/s cells=1020.1k
```

### DIFF
--- a/include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
+++ b/include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
@@ -165,54 +165,137 @@ protected:
         // Initialisation phase: allocate memory and initialise first column.
         // ---------------------------------------------------------------------
 
+        this->reset_optimum(); // Reset the tracker for the new alignment computation.
+
         thread_local score_matrix_single_column<int32_t> local_matrix{};
         local_matrix.reset_matrix(std::forward<sequence1_t>(sequence1), std::forward<sequence2_t>(sequence2));
         auto matrix_iter = local_matrix.begin();
-        auto alignment_column = *matrix_iter;
 
-        // Initialise the first column.
-        *alignment_column.begin() = this->initialise_origin_cell();
-        for (auto cell : alignment_column | seqan3::views::drop(1))
-            cell = this->initialise_first_column_cell(cell);
+        initialise_column(*matrix_iter, sequence2);
 
         // ---------------------------------------------------------------------
-        // Recursion phase: compute column-wise the alignment matrix.
+        // Iteration phase: compute column-wise the alignment matrix.
         // ---------------------------------------------------------------------
 
-        // Compute remaining matrix
-        ++matrix_iter; // Move to next column.
-        for (auto sequence1_iter = sequence1.begin();
-             sequence1_iter != sequence1.end();
-             ++sequence1_iter, ++matrix_iter)
+        for (auto sequence1_value : sequence1)
         {
-            alignment_column = *matrix_iter;
-            auto alignment_column_iter = alignment_column.begin();
-
-            // Initialise first cell of current column.
-            auto cell = *alignment_column_iter;
-            typename traits_type::score_type diagonal = cell.optimal_score();
-            *alignment_column_iter = this->initialise_first_row_cell(cell);
-
-            ++alignment_column_iter;  // Move to next cell in column.
-            for (auto sequence2_iter = sequence2.begin();
-                 sequence2_iter != sequence2.end();
-                 ++sequence2_iter, ++alignment_column_iter)
-            {
-                auto cell = *alignment_column_iter;
-                typename traits_type::score_type next_diagonal = cell.optimal_score();
-                *alignment_column_iter = this->compute_inner_cell(diagonal,
-                                                                  cell,
-                                                                  m_scoring_scheme.score(*sequence1_iter,
-                                                                                         *sequence2_iter));
-                diagonal = next_diagonal;
-            }
+            ++matrix_iter;
+            compute_column(*matrix_iter, sequence1_value, sequence2);
         }
 
         // ---------------------------------------------------------------------
-        // Wrap up phase: track score of last column
+        // Final phase: track score of last column
         // ---------------------------------------------------------------------
 
-        return alignment_column[std::ranges::distance(sequence2)].optimal_score();
+        auto alignment_column = *matrix_iter;
+        auto column_iterator = alignment_column.begin();
+        this->track_last_column_cell(*column_iterator);
+
+        for ([[maybe_unused]] auto && unused : sequence2)
+            this->track_last_column_cell(*++column_iterator);
+
+        this->track_final_cell(*column_iterator);
+
+        return this->tracked_optimum();
+    }
+
+    /*!\brief Initialise the first column of the alignment matrix.
+     * \tparam alignment_matrix_column_t The type of the alignment matrix; must model std::ranges::input_range.
+     * \tparam sequence2_t The type of the second sequence; must model std::ranges::input_range.
+     *
+     * \param[in] first_column The first column of the alignment matrix to initialise.
+     * \param[in] sequence2 The second sequence used to determine the size of the column.
+     *
+     * \details
+     *
+     * The first column of the alignment matrix does not require any character comparisons of the sequences that
+     * shall be aligned. The second sequence is thus only needed to determine the size of the column.
+     * The computation of the column is split into three phases: the initialisation phase, the iteration phase, and
+     * the final phase. In the initialisation phase the first cell of the column is computed and in the iteration
+     * phase all remaining cells are computed. In the final phase the last cell is possibly evaluated for a new
+     * alignment optimum.
+     */
+    template <std::ranges::input_range alignment_matrix_column_t, std::ranges::input_range sequence2_t>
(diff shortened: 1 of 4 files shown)


include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
```diff
@@ -0,0 +1,173 @@
...
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides seqan3::detail::policy_optimum_tracker.
+ * \author Rene Rahn <rene.rahn AT fu-berlin.de>
+ */
+
+#pragma once
```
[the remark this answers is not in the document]
rrahn: well, I guess this is one of the controversial features discussed in the community. 
The first thing is, that it is not a standard feature but so far a well supported one by the major compiler vendors. At least the one we promise to work with.
Another issues that arised in the past is that the pragma did had issues with large code bases where it can happen that the same header base name (`a/foo.hpp` and `b/foo.hpp`) was used in different locations. However, that is also a problem for include guards (but of course can be fixed manuualy) but it was fixed already quite some time ago. I am not aware of other issues that might happen though.
This means as far as I know there is no disadvantage of `#pragma once` over the include guards. At least we didn't encounter one yet. Not sure if it comiles faster, which is indicated by some threads. However, the general advantage over include guards is spelling safety:
```
#indef some_intermediate_header_with_a_long_name
#define some_intermidiate_header_with_a_long_name
...
#endif // some_intermediate_header_with_a_long_name
```
So as you can see, not only is it much easier to type but also prevents some very hard to track errors which could literally blow the compile time and the binary size if the header is included many many times or end in some weird not so obvious to fix error messages because of redefinitions etc.. Similarly to have a non matching pair of parenthesis in the code which gives you suddenly errors about some namespaces not defined in some completely random files ;)
So our policy is to go with `#pragma once` as long as it does not raise any troubles, which seems to be very unlikely and hope to live happily ever after. 😄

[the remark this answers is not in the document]
wvandertoorn: Thank you for the elaborate explanation!

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
```diff
@@ -0,0 +1,173 @@
...
+/*!\brief Implements the tracker to store the global optimum for a particular alignment computation.
+ * \ingroup pairwise_alignment
+ *
+ * \tparam alignment_configuration_t The type of the alignment configuration; must be a type specialisation of
+ *                                   seqan3::configuration.
+ *
+ * \details
+ *
+ * Implements the interface to track the alignment optimum. It updates the currently stored optimum if the
+ * optimal score of the given cell compares greater or equal to the sotred optimum.
```
smehringer, line 33:
```suggestion
 * optimal score of the given cell compares greater or equal to the stored optimum.
```

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
```diff
@@ -0,0 +1,173 @@
...
+ * Special methods are offered to track any cell (for example when computing the local alignment), the last cell of a
+ * column or a row (for example when using free-end gaps), or the final cell of the entire matrix (for example in the
+ * standard global alignment).
+ * Depending on the configuration they might or might not evaluate the stored optimum score of that cell to compare
+ * it with the currently stored maximum.
```
smehringer, line 38:
```suggestion
 * it with the currently stored optimum.
```
for consistency with text above

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
```diff
@@ -0,0 +1,173 @@
...
+     *
+     * A call to this function only tracks the optimal score of the given cell if the configuration of the alignment
+     * algorithm requires it, for example when a global alignment shall be computed.
+     */
+    template <typename cell_t>
+    decltype(auto) track_final_cell(cell_t && cell) noexcept
+    {
+        m_optimal_score = (cell.optimal_score() >= m_optimal_score) ? cell.optimal_score() : m_optimal_score;
+        return std::forward<cell_t>(cell);
+    }
```
smehringer, line 156: Some things I don't understand yet:

- The other `track_...` function don't do anything yet because the new alignment code only works for global alignments so far, but you added the functions here already, correct?
- The documentation for `track_final_cell` indicates that the optimal score is only tracked if configured, but I don't see a dependency on the configuration here? 
- Can you give an example on how the other functions will differ from the `track_final_cell` in the future?

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
```diff
@@ -165,54 +165,137 @@ class pairwise_alignment_algorithm : protected policies_t...
...
+        // Iteration phase: iterate over column and compute each cell
+        // ---------------------------------------------------------------------
+
+        for ([[maybe_unused]] auto && unused : sequence2)
         {
-            alignment_column = *matrix_iter;
-            auto alignment_column_iter = alignment_column.begin();
+            ++first_column_iter;
+            *first_column_iter = this->track_cell(this->initialise_first_column_cell(*first_column_iter));
+        }
```
smehringer, line 236: This looks like the column is already resized to the size of sequence2, since you do not push_back but assign. Can you not query the size of an alignment column, then you would not need sequence2 for the size?

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
rrahn: >* The other track_... function don't do anything yet because the new alignment code only works for global alignments so far, but you added the functions here already, correct?

correct, to see that the interface of the algorithm already works with this and the vectorised global alignment needs some of them already.

>The documentation for track_final_cell indicates that the optimal score is only tracked if configured, but I don't see a dependency on the configuration here?

Yep, because later we evaluate the configuration object and check what parameters have been set for free-end gaps etc and then conditionally enable/disable the compare and exchange part in the respective functions.

> Can you give an example on how the other functions will differ from the track_final_cell in the future?

For example:

```cpp
auto track_last_row_cell(cell)
{
   if (m_track_last_row_enabled)
      compare_and_exchange(cell);

   return ...;
}
```
Where compare_and_exchange will be the function implementing the actual comparison and update of the global optimum, which will be resused by all functions, or overwritten by specialisations of this policy.

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
rrahn: I might not always have a sized range. In general, this is not a generic part since we control the matrix implementation but with combining them with views and modifying them it can happen that a column is not sized for any particular reason. 
I also don't want to go to the end of the column since I want to do something special with the last cell of the column. Since a column expects at most an input range (actually also an output range but that will come later), I cannot simply set the iterator to the one before the end. So in either case I have to introduce some additional variable to iterate over the column. A valid point would be to just iterate over the sequence and not dereference the value. But I couldn't find any implications when looking at the assembly. So far the results indicate no change. In the banded case I might safe some head-ach since the sequence slice will determine the actual size of the column.

PR #1792, opened 2020-05-07
[FIX] Prohibit long-option value pairs without a separation.

### LINKED ISSUE #1544: Argument Parser throws wrong error
if you have a valid option `--foo` and a non-existant option `--foo-bar` and give both a la `--foo X --foo-bar Y` the arg parser will print:
```
Option -f/--foo is no list/container but specified multiple times
```
Instead of just saying that `--foo-bar` does not exist.

smehringer:
For example, given the long option name "--opt", then "--optValue" was
possible. This caused ambiguities if one long option name is the prefix
of another. We now only allow "--opt Value" and "--opt=value".

Fixes #1544

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,15 @@ Note that 3.1.0 will be the first API stable release and interfaces in this rele
 
 ## Notable Bug-fixes
 
+### Argument Parser
+
+* Long option identifiers and their value must be separated by a space or equal sign `=`.
+  Handling this restriction resolves the ambiguity if one long option identifier is the prefix of 
+  another ([\#1792](https://github.com/seqan/seqan3/pull/1792)).
+
+  Valid short id value pairs: `-iValue`, `-i=Value`, `-i Value`
+  Valid long id value pairs: `--id=Value`, `--id Value` (prohibited now: `--idValue`)
+
 ### I/O
 
 * The `seqan3::field::cigar` was added to the default fields for reading and writing alignment files
--- a/include/seqan3/argument_parser/detail/format_parse.hpp
+++ b/include/seqan3/argument_parser/detail/format_parse.hpp
@@ -202,7 +202,8 @@ private:
     }
 
     /*!\brief Finds the position of a short/long identifier in format_parse::argv.
-     *
+     * \tparam id_type The identifier type; must be either of type `char` if it denotes a short identifier or
+     *                 std::string if it denotes a long identifier.
      * \param[in] begin_it The iterator where to start the search of the identifier.
      *                     Note that the end iterator is kept as a member variable.
      * \param[in] id       The identifier to search for (must not contain dashes).
@@ -210,8 +211,16 @@ private:
      *          the list pointed to by begin_t. If the list does not contain the
      *          identifier `id`, the member variable `end_of_options_it` is returned.
      *
-     * Note: The `id` is compared to the prefix of each value in the list, such
-     *       that "-idValue" arguments are correctly identified.
+     * \details
+     *
+     * **Valid short-id value pairs are: `-iValue`, `-i=Value`, or `-i Value`**
+     * If the `id` passed to this function is of type `char`, it is assumed to be a short identifier.
+     * The `id` is found by comparing the prefix of every argument in argv to the `id` prepended with a single `-`.
+     *
+     * **Valid long id value pairs are: `--id=Value`, `--id Value`**.
+     * If the `id` passed to this function is of type `std::string`, it is assumed to be a long identifier.
+     * The `id` is found by comparing every argument in argv to `id` prepended with two dashes (`--`)
+     * or a prefix of such followed by the equal sign `=`.
      */
     template <typename id_type>
     std::vector<std::string>::iterator find_option_id(std::vector<std::string>::iterator const begin_it, id_type const & id)
@@ -220,13 +229,22 @@ private:
             return end_of_options_it;
 
         return (std::find_if(begin_it, end_of_options_it,
-            [&] (const std::string & v)
+            [&] (std::string const & current_arg)
             {
-                size_t id_size{(prepend_dash(id)).size()};
-                if (v.size() < id_size)
-                    return false; // cannot be the correct identifier
+                std::string full_id = prepend_dash(id);
 
-                return v.substr(0, id_size) == prepend_dash(id); // check if prefix of v is the same
+                if constexpr (std::same_as<id_type, char>) // short id
+                {
+                    // check if current_arg starts with "-o", i.e. it correctly identifies all short notations:
+                    // "-ovalue", "-o=value", and "-o value".
+                    return current_arg.substr(0, full_id.size()) == full_id;
+                }
+                else
+                {
+                    // only "--opt Value" or "--opt=Value" are valid
+                    return current_arg.substr(0, full_id.size()) == full_id && // prefix is the same
+                           (current_arg.size() == full_id.size() || current_arg[full_id.size()] == '='); // space or `=`
+                }
             }));
     }
 
--- a/test/unit/argument_parser/format_parse_test.cpp
+++ b/test/unit/argument_parser/format_parse_test.cpp
@@ -60,16 +60,6 @@ TEST(parse_type_test, add_option_long_id)
     EXPECT_TRUE((testing::internal::GetCapturedStderr()).empty());
     EXPECT_EQ(option_value, "option_string");
 
-    // add with no space
-    const char * argv2[] = {"./argument_parser_test", "--string-optionoption_string"};
-    seqan3::argument_parser parser2{"test_parser", 2, argv2, false};
-    parser2.add_option(option_value, 'S', "string-option", "this is a string option.");
-
-    testing::internal::CaptureStderr();
-    EXPECT_NO_THROW(parser2.parse());
-    EXPECT_TRUE((testing::internal::GetCapturedStderr()).empty());
-    EXPECT_EQ(option_value, "option_string");
-
     const char * argv3[] = {"./argument_parser_test", "--string-option=option_string"};
     seqan3::argument_parser parser3{"test_parser", 2, argv3, false};
     parser3.add_option(option_value, 's', "string-option", "this is a string option.");
@@ -934,5 +924,48 @@ TEST(parse_test, subcommand_argument_parser_success)
         const char * argv[]{"./top_level", "-f", "2", "subiddysub", "foo"};
         EXPECT_THROW((seqan3::argument_parser{"top_level", 5, argv, false, {"sub1", "sub2"}}),
                      seqan3::argument_parser_error);
+    }
+}
+
+TEST(parse_test, issue1544)
+{
+    {   // wrong separation of long value:
+        std::string option_value;
+        const char * argv[] = {"./argument_parser_test", "--foohallo"};
+        seqan3::argument_parser parser{"test_parser", 2, argv, false};
+        parser.add_option(option_value, 'f', "foo", "this is a string option.");
+
+        EXPECT_THROW(parser.parse(), seqan3::unknown_option);
+    }
+
+    {   // unknown option (`--foo-bar`) that has a prefix of a known option (`--foo`)
+        std::string option_value;
+        const char * argv[] = {"./argument_parser_test", "--foo", "hallo", "--foo-bar", "ballo"};
+        seqan3::argument_parser parser{"test_parser", 5, argv, false};
+        parser.add_option(option_value, 'f', "foo", "this is a string option.");
+
+        EXPECT_THROW(parser.parse(), seqan3::unknown_option);
+    }
+
+    {   // known option (`--foo-bar`) that has a prefix of a unknown option (`--foo`)
+        std::string option_value;
+        const char * argv[] = {"./argument_parser_test", "--foo", "hallo", "--foo-bar", "ballo"};
+        seqan3::argument_parser parser{"test_parser", 5, argv, false};
+        parser.add_option(option_value, 'f', "foo-bar", "this is a string option.");
+
+        EXPECT_THROW(parser.parse(), seqan3::unknown_option);
+    }
+
+    {   // known option (`--foo`) is a prefix of another known option (`--foo-bar`)
+        std::string foo_option_value;
+        std::string foobar_option_value;
+        const char * argv[] = {"./argument_parser_test", "--foo", "hallo", "--foo-bar", "ballo"};
+        seqan3::argument_parser parser{"test_parser", 5, argv, false};
+        parser.add_option(foo_option_value, 'f', "foo", "this is a prefix of foobar.");
+        parser.add_option(foobar_option_value, 'b', "foo-bar", "this has prefix foo.");
+
+        EXPECT_NO_THROW(parser.parse());
+        EXPECT_EQ(foo_option_value, "hallo");
+        EXPECT_EQ(foobar_option_value, "ballo");
     }
 }


include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -222,11 +222,30 @@ class format_parse : public format_base
         return (std::find_if(begin_it, end_of_options_it,
             [&] (const std::string & v)
             {
-                size_t id_size{(prepend_dash(id)).size()};
-                if (v.size() < id_size)
-                    return false; // cannot be the correct identifier
+                std::string full_id = prepend_dash(id);
 
-                return v.substr(0, id_size) == prepend_dash(id); // check if prefix of v is the same
+                if constexpr (std::same_as<id_type, char>) // short id
```
eseiler, line 227: Applies to both `prepend_dash` and the `if constexpr`

This also depends on how this function is used, but this won't work for short id's if the input is either a CStyle String or a std::string (Instead of `o` you pass `"o" or a `std::string`).

If this is intended, you could maybe add a sentence to the documentation that it is a strict requirement that the `id` must be passed as `char` to be identified as short id?

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -222,11 +222,30 @@ class format_parse : public format_base
         return (std::find_if(begin_it, end_of_options_it,
```
eseiler, line 222: A few lines above in the documentation:

```
     * Note: The `id` is compared to the prefix of each value in the list, such
     *       that "-idValue" arguments are correctly identified.	     *       that "-idValue" arguments are correctly identified.
```

`"-idValue"` won't work anymore.
Also, maybe you can use `\note`? Or at least add an explicit `\details`

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -222,11 +222,30 @@ class format_parse : public format_base
...
+                    return v.substr(0, full_id.size()) == full_id;
+                }
+                else
+                {
+                    // only "--opt Value" or "--opt=Value" are valid
+                    if (v.substr(0, full_id.size()) == full_id)
+                    {
+                        if (v.size() == full_id.size())
+                            return true; // value was separated by space and is next in argv
+                        else if (v[full_id.size()] == '=')
```
eseiler, line 239: Is this save? 

What if my input in the arg list is `--validArg validValue --NoValue`. Would it access out of bounds?

eseiler:
And also please add a changelog entry

include/seqan3/argument_parser/detail/format_parse.hpp
smehringer: Yes, it is a silent requirement, as I handled it that way all over the argument parser (the add option call takes char and std::string for them).
I think enforcing this somehow would require some refactoring, so will just document it for now.

include/seqan3/argument_parser/detail/format_parse.hpp
smehringer: I need to adjust the documentation! Thanks for spotting this.

in you example there is one twirk. we do not allow multi character short ids so "-idValue" will never occur.
Valid short id value pairs: `-iValue`, `-i=Value`, `-i Value`
Valid long id value pairs (now) `--id=Value`, `--id Value`

smehringer: ops that was my example not yours... anyway I'll adjust it

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -202,16 +202,25 @@ class format_parse : public format_base
...
      * \returns An iterator pointing to the first occurrence of the identifier in
      *          the list pointed to by begin_t. If the list does not contain the
      *          identifier `id`, the member variable `end_of_options_it` is returned.
      *
-     * Note: The `id` is compared to the prefix of each value in the list, such
-     *       that "-idValue" arguments are correctly identified.
+     * \details
+     *
+     * **Valid short-id value pairs are: `-iValue`, `-i=Value`, or `-i Value`**
+     * If the `id` passed to this function is of type `char` it is assumed to be a short identifier.
```
eseiler, line 217:
```suggestion
     * If the `id` passed to this function is of type `char`, it is assumed to be a short identifier.
```

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -202,16 +202,25 @@ class format_parse : public format_base
...
+     * The `id` is found by comparing the prefix of every argument in argv to the `id` prepended with a single `-`.
+     *
+     * **Valid long id value pairs are: `--id=Value`, `--id Value`**.
+     * If the `id` passed to this function is of type `char` it is assumed to be a short identifier.
```
eseiler, line 221:
```suggestion
     * If the `id` passed to this function is of type `std::string`, it is assumed to be a long identifier.
```

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -202,16 +202,25 @@ class format_parse : public format_base
...
+     * The `id` is found by comparing every argument in argv whether it is `id` prepended with a two dashes `--`
```
eseiler, line 222:
```suggestion
     * The `id` is found by comparing every argument in argv to `id` prepended with two dashes (`--`)
```

include/seqan3/argument_parser/detail/format_parse.hpp
eseiler: Yes!

But `if (v.substr(0, full_id.size()) == full_id)` should throw if `v.size() < full_id.size()`, because `substr` has a bounds check.
Can we construct an example where this might happen?
Is there a case where we might look up a very long ID in a argv that contains only short stuff?
It should throw if we look up `IAmAVeryLongID` in a argv of `["-not", "so", "--long"]`?

CHANGELOG.md
```diff
@@ -71,6 +71,14 @@ Note that 3.1.0 will be the first API stable release and interfaces in this rele
 
 ## Notable Bug-fixes
 
+### Argument Parser
+
+* Long option identifiers and their value must be separated by a space or equal sign `=`.
+  Handling this restriction resolves the ambiguity if one long option identifier is the prefix of another ([\#1792](https://github.com/seqan/seqan3/pull/1792)).
```
eseiler, line 77:
```suggestion
  Handling this restriction resolves the ambiguity if one long option identifier is the prefix of 
  another ([\#1792](https://github.com/seqan/seqan3/pull/1792)).
```

include/seqan3/argument_parser/detail/format_parse.hpp
smehringer: You can also do multi-line suggestions :)

include/seqan3/argument_parser/detail/format_parse.hpp
smehringer: I'll add a test!

smehringer: I added a test with a valid option `foo-bar` and a command line of `--foo value --foo-bar value`, which should trigger this case. It works as expected.
I looked up the documentation of `str.subtr(pos1, pos2)` again and if `pos2 > str.end()` then `str.subtr(pos1, str.end())` is returned and no exception is thrown.

eseiler: Yeah, I read it wrong, it throws if you do `str.substr(pos, count)`  and `pos` is out of bounds, if `pos+count` is out of bounds, it just returns the suffix starting at `pos` if `str`

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -202,16 +202,25 @@ class format_parse : public format_base
     }
 
     /*!\brief Finds the position of a short/long identifier in format_parse::argv.
-     *
+     * \tparam id_type The identifier type. If it is `char`, the parameter `id` is assumed to be a short identifier,
```
rrahn, line 205:
```suggestion
     * \tparam id_type The identifier type; must be either of type `char` if it denotes a short identifier or std::string if it denotes a long identifier.
```

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -222,11 +231,30 @@ class format_parse : public format_base
...
             {
-                size_t id_size{(prepend_dash(id)).size()};
-                if (v.size() < id_size)
-                    return false; // cannot be the correct identifier
+                std::string full_id = prepend_dash(id);
 
-                return v.substr(0, id_size) == prepend_dash(id); // check if prefix of v is the same
+                if constexpr (std::same_as<id_type, char>) // short id
+                {
+                    // check if prefix of v is the same since "-oValue" "-o=Value" "-o value" are all valid
```
rrahn, line 238:
```suggestion
                    // check if v starts with "-o", i.e. it correctly identifies all short notations: "-ovalue", "-o=value", and "-o value".
```

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -222,11 +231,30 @@ class format_parse : public format_base
         return (std::find_if(begin_it, end_of_options_it,
             [&] (const std::string & v)
```
rrahn, line 232:
```suggestion
            [&] (std::string const & option_identifier)
```

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -222,11 +231,30 @@ class format_parse : public format_base
...
+                        else if (v[full_id.size()] == '=')
+                            return true; // value is separated by '='
+                        else
+                            return false;
+                    }
+                    else
+                    {
+                        return false;
+                    }
+                }
```
rrahn, line 257:
```suggestion
                    return v.substr(0, full_id.size()) == full_id && (v.size() == full_id.size() || v[full_id.size()] == '=');
```
?

test/unit/argument_parser/format_parse_test.cpp
```diff
@@ -936,3 +926,37 @@ TEST(parse_test, subcommand_argument_parser_success)
                      seqan3::argument_parser_error);
     }
 }
+
+TEST(parse_test, issue1544)
+{
+    {   // unknown option that has a prefix of a known option
```
rrahn, line 932:
```suggestion
    {   // unknown option (`--foo-bar`) that has a prefix of a known option (`--foo`)
```

test/unit/argument_parser/format_parse_test.cpp
```diff
@@ -936,3 +926,37 @@ TEST(parse_test, subcommand_argument_parser_success)
...
+        std::string option_value;
+        const char * argv[] = {"./argument_parser_test", "--foo", "hallo", "--foo-bar", "ballo"};
+        seqan3::argument_parser parser{"test_parser", 5, argv, false};
+        parser.add_option(option_value, 'f', "foo", "this is a string option.");
+
+        EXPECT_THROW(parser.parse(), seqan3::unknown_option);
+    }
+
+    {   // known option that has a prefix of a unknown option
```
rrahn, line 941:
```suggestion
    {   // known option (`--foo-bar`) that has a prefix of a unknown option (`--foo`)
```

test/unit/argument_parser/format_parse_test.cpp
```diff
@@ -936,3 +926,37 @@ TEST(parse_test, subcommand_argument_parser_success)
...
+        std::string option_value;
+        const char * argv[] = {"./argument_parser_test", "--foo", "hallo", "--foo-bar", "ballo"};
+        seqan3::argument_parser parser{"test_parser", 5, argv, false};
+        parser.add_option(option_value, 'f', "foo-bar", "this is a string option.");
+
+        EXPECT_THROW(parser.parse(), seqan3::unknown_option);
+    }
+
+    {   // known option is a prefix of another known option
```
rrahn, line 950:
```suggestion
    {   // known option (`--foo`) is a prefix of another known option (`--foo-bar`)
```

test/unit/argument_parser/format_parse_test.cpp
```diff
@@ -936,3 +926,37 @@ TEST(parse_test, subcommand_argument_parser_success)
...
+        std::string foobar_option_value;
+        const char * argv[] = {"./argument_parser_test", "--foo", "hallo", "--foo-bar", "ballo"};
+        seqan3::argument_parser parser{"test_parser", 5, argv, false};
+        parser.add_option(foo_option_value, 'f', "foo", "this is a prefix of foobar.");
+        parser.add_option(foobar_option_value, 'b', "foo-bar", "this has prefix foo.");
+
+        EXPECT_NO_THROW(parser.parse());
+        EXPECT_EQ(foo_option_value, "hallo");
+        EXPECT_EQ(foobar_option_value, "ballo");
+    }
```
rrahn, line 961: Does this also work when mixing short and long identifier. Basically, when I call  `-f` and `--foo-bar`?

test/unit/argument_parser/format_parse_test.cpp
```diff
@@ -936,3 +926,37 @@ TEST(parse_test, subcommand_argument_parser_success)
...
+}
```
rrahn, line 962: Is there a test missing for calling the argument parser with: `--foohallo`. So without the equal sign or separated with a space? This should than throw something, right?

smehringer: I think it is tested by option `--foo` with value `--foo-bar` with the value being `-bar`? I can remove the dash between `foo-bar` if this makes it clearer?

test/unit/argument_parser/format_parse_test.cpp
smehringer: No because I explicitly test for the single `-` or double dash `--` so those are never amigious.

include/seqan3/argument_parser/detail/format_parse.hpp
smehringer: Nice thanks.

smehringer: Although the line is too long so I will store a variable and return that.

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -202,8 +202,9 @@ class format_parse : public format_base
     }
 
     /*!\brief Finds the position of a short/long identifier in format_parse::argv.
-     * \tparam id_type The identifier type. If it is `char`, the parameter `id` is assumed to be a short identifier,
-                       on the other hand, if it is `std::string`, the parameter `id` is assumed to be a long identifier.
+     * \tparam id_type The identifier type; must be either of type `char` if it denotes a short identifier or
+     *                 std::string if it denotes a long identifier. On the other hand, if it is `std::string`, the
+     *                 parameter `id` is assumed to be a long identifier.
```
rrahn, line 207:
```suggestion
     *                 std::string if it denotes a long identifier.
```

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -229,31 +230,22 @@ class format_parse : public format_base
...
-                        else if (v[full_id.size()] == '=')
-                            return true; // value is separated by '='
-                        else
-                            return false;
-                    }
-                    else
-                    {
-                        return false;
-                    }
+                    bool valid_long_id = current_arg.substr(0, full_id.size()) == full_id && // prefix is the same
```
rrahn, line 246: why don't you return it directly?

test/unit/argument_parser/format_parse_test.cpp
rrahn: amigious 🤠

test/unit/argument_parser/format_parse_test.cpp
rrahn: Ah ok, from the test design I thought that this was a different thing to test. But it does test the same that is right. I guess the issue is not the dash inbetween but that the tests checks if the one is a prefix of another. So from the test I don't actually see that long-option value pairs without a separation are not allowed. 
```cpp
    {   // wrong separation of long value:
        std::string option_value;
        const char * argv[] = {"./argument_parser_test", "--foohallo"};
        seqan3::argument_parser parser{"test_parser", 2, argv, false};
        parser.add_option(option_value, 'f', "foo", "this is a string option.");

        EXPECT_THROW(parser.parse(), seqan3::unknown_option);
    }
```

include/seqan3/argument_parser/detail/format_parse.hpp
smehringer: oh sorry

include/seqan3/argument_parser/detail/format_parse.hpp
smehringer: Because multiline return statements are ugly? :D

test/unit/argument_parser/format_parse_test.cpp
smehringer: Alright. I added the test next to the existing ones. Doesn't hurt to be expressive although they internally test the same thing

include/seqan3/argument_parser/detail/format_parse.hpp
rrahn: Oh my, can you please directly return it? We usually don’t create extra variables if it is not part of another composition or used multiple times. I also think it does not make the code easier to understand since the expression is the same and one is wondering why there was an extra variable. I was wondering that it was meant for something else. That’s why I asked :)

PR #1793, opened 2020-05-07
gcc-10 --std=c++2a iterators

marehr:
* gcc10-std2a add readable = indirectly_readable to std namespace (for now)
* gcc10-std2a add writable = indirectly_writable to std namespace (for now)

The standard has renamed `std::readable` to `std::indirectly_readable` and
`std::writable` to `std::indirectly_writable`, but we internally use the old names.

To make `-std=c++2a` work, we import the names into the standard namespace.

In a later PR we should do the renaming.

----

We falsely use

* `std::ranges::back_inserter` that is actually in `std::back_inserter`
* `std::ranges::default_sentinel` that is actually in `std::default_sentinel`
* `std::ranges::default_sentinel_t` that is actually in `std::default_sentinel_t`

To make `-std=c++2a` work, we import the names into the standard ranges namespace.

In a later PR we should put them in the namespace they belong.

### DIFF
--- a/include/seqan3/std/iterator
+++ b/include/seqan3/std/iterator
@@ -37,6 +37,22 @@ concept writable = ::std::indirectly_writable<Out, T>;
 } // namespace std::deprecated
 } // namespace std
 
+// we wrongly assume that these are in std::ranges namespace, even though they are in std namespace
+namespace std::ranges
+{
+inline namespace deprecated
+{
+// https://eel.is/c++draft/iterator.synopsis#lib:default_sentinel
+// // Header <iterator> synopsis
+// // [default.sentinels], default sentinels
+// struct default_sentinel_t;
+// inline constexpr default_sentinel_t default_sentinel{};
+using ::std::default_sentinel_t;
+using ::std::default_sentinel;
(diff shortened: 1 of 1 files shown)


include/seqan3/std/iterator
```diff
@@ -69,7 +99,7 @@ namespace
     using ::ranges::permutable;
     using ::ranges::projected;
     using ::ranges::random_access_iterator;
-    using ::ranges::readable;
+    using ::ranges::readable; // deprecated or renamed? -> indirectly_readable?
```
MitraDarja, line 102: Maybe add a TODO in the comment, so it is easier to find with a search? I don't know how we usually track things that are supposed to be changed in the future.

marehr: This is a forgotten left-over; I track all open changes regarding C++20 in 

https://github.com/seqan/product_backlog/issues/49

marehr:
@MitraDarja  I removed that comment.

marehr:
> The problem is that the ranges library still defines them in the wrong namespace or with the wrong name?

Yes

PR #1794, opened 2020-05-07
gcc-10 --std=c++2a concepts

marehr:
There is no `std::detail::weakly_equality_comparable_with` concept.
The standard introduces a exposition-only concept named `weakly-equality-comparable-with`
(https://eel.is/c++draft/concept.equalitycomparable#1).

This PR adds this concept into the std namespace, but a later PR should
remove all usage of this exposition-only concept or we should move that
concept into the seqan namespace.

---

There is no `std::boolean` concept.
The standard introduces a exposition-only concept named `boolean-testable`
(https://eel.is/c++draft/concepts#concept.booleantestable-1).

This PR adds this concept into the std namespace, but a later PR should
remove all usage of this exposition-only concept or we should move that
concept into the seqan namespace.

---

The standard has renamed `std::default_constructible` to `std::default_initializable`, but we internally use `std::default_constructible`.

To make `-std=c++2a` work, we import the name into the standard namespace.

In a later PR we should do the renaming.

### DIFF
--- a/include/seqan3/std/concepts
+++ b/include/seqan3/std/concepts
@@ -17,6 +17,18 @@
 
 #ifdef __cpp_lib_concepts        // C++20 ranges available
 #include <concepts>
+
+namespace std
+{
+inline namespace deprecated
+{
+/*!\brief This should be named default_initializable.
+ * \deprecated Use std::default_initializable instead.
+ */
+template<class T>
+concept default_constructible = default_initializable<T>;
+} // inline namespace std::deprecated
+} // namespace std
 #else                         // use range-v3 to emulate
 
 #include <functional>
--- a/include/seqan3/std/concepts
+++ b/include/seqan3/std/concepts
@@ -54,6 +54,29 @@ concept boolean =
     };
 } // inline namespace std::deprecated
 } // namespace std
+
+namespace std::detail
+{
+/*!\interface   std::detail::weakly_equality_comparable_with <>
+ * \tparam t1   The first type to compare.
+ * \tparam t2   The second type to compare.
+ * \brief       Requires the two operands to be comparable with `==` and `!=` in both directions.
+ * \sa          https://en.cppreference.com/w/cpp/concepts/equality_comparable
+ */
+//!\cond
+template <class T, class U>
+concept weakly_equality_comparable_with =
+    requires(std::remove_reference_t<T> const & t,
+             std::remove_reference_t<U> const & u)
+    {
+        t == u; requires boolean<decltype(t == u)>;
+        t != u; requires boolean<decltype(t != u)>;
+        u == t; requires boolean<decltype(u == t)>;
+        u != t; requires boolean<decltype(u != t)>;
+    };
+//!\endcond
+} // namespace std::detail
+
 #else                         // use range-v3 to emulate
 
 #include <functional>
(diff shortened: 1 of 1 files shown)


include/seqan3/std/concepts
```diff
@@ -17,6 +17,66 @@
...
+        b1 || b2; requires same_as<decltype(b1 || b2), bool>;
+        a || b2;  requires same_as<decltype(a || b2), bool>;
+        b1 == b2; requires convertible_to<decltype(b1 == b2), bool>;
+        b1 == a;  requires convertible_to<decltype(b1 == a), bool>;
+        a == b2;  requires convertible_to<decltype(a == b2), bool>;
+        b1 != b2; requires convertible_to<decltype(b1 != b2), bool>;
+        b1 != a;  requires convertible_to<decltype(b1 != a), bool>;
+        a != b2;  requires convertible_to<decltype(a != b2), bool>;
+    };
+} // namespace std::deprecated
```
joergi-w, line 55:
```suggestion
} // inline namespace std::deprecated
```

include/seqan3/std/concepts
```diff
@@ -17,6 +17,66 @@
...
+{
+/*!\interface   std::detail::weakly_equality_comparable_with <>
+ * \tparam t1   The first type to compare.
+ * \tparam t2   The second type to compare.
+ * \brief       Requires the two operands to be comparable with `==` and `!=` in both directions.
+ * \sa          https://en.cppreference.com/w/cpp/concepts/weakly_equality_comparable_with
+ */
+//!\cond
+template <class T, class U>
+concept weakly_equality_comparable_with =
```
joergi-w, line 68: Why the `weakly`? Cannot we use just `equality_comparable_with`, which is more intuitive? Otherwise, please say in the documentation what weakly means.

marehr: This is the point; we use it internally even though we might use the non-exposition only version.

marehr: But this is not the scope of this PR; this PR just enables to able to compile it with `-std=c++2a`.

joergi-w: ok!

include/seqan3/std/concepts
```diff
@@ -17,6 +17,66 @@
 
 #ifdef __cpp_lib_concepts        // C++20 ranges available
 #include <concepts>
+
```
smehringer, line 20: Here you haven't made the ifdef of including `<concepts>` separate from `__cpp_lib_concepts` as in the last PR I reviewed, I thought it is favourable?

include/seqan3/std/concepts
```diff
@@ -17,6 +17,66 @@
...
+ * \deprecated Use std::default_initializable instead.
+ */
+template<class T>
+concept default_constructible = default_initializable<T>;
+
+/*!\brief This should be removed.
+ * \deprecated No viable alternative.
+ */
+template <class B>
+concept boolean =
```
smehringer, line 35: Can the concept boolean in line 133 be removed now since we define it here?

include/seqan3/std/concepts
marehr: You are right and I'm also confused why this works;

EDIT:// boah `platform.hpp` includes `<version>` which defines `__cpp_lib_concepts` :facepalm:

include/seqan3/std/concepts
marehr: It should also be removed in the future;

but we can't do it right now, because neither ranges-v3 (= non `-std=c++2a` mode) nor std::ranges (= `-std=c++2a` mode) defines it, but we use it in the library.

include/seqan3/std/concepts
marehr: I'll change this to the new pattern.

PR #1795, opened 2020-05-07
gcc-10 --std=c++2a ranges

marehr:
This one was a bit unexpected, but ranges and std::ranges define there own `view_base` class and they both say, if my own version of that class is a base class then it is a view.

This fix makes views of these two worlds interoperable.

There is also an open issue in range-v3: ericniebler/range-v3#1470

---

There is no `std::ranges::all_view`

* which should have been in `std::views`
* and was renamed to `std::views::all_t` (https://eel.is/c++draft/ranges).

This PR adds this view into the std ranges namespace, but a later PR should
remove all usage of this.

### DIFF
--- a/include/seqan3/std/ranges
+++ b/include/seqan3/std/ranges
@@ -13,8 +13,35 @@
 
 #pragma once
 
-#if __cpp_lib_ranges // C++20 ranges available
+#if __has_include(<ranges>)
 #include <ranges>
+#endif // __has_include(<ranges>)
+
+#if defined(__cpp_lib_ranges) // C++20 ranges available
+#include <seqan3/std/iterator>
+
+#include <range/v3/range/concepts.hpp>
+
+namespace ranges
+{
+//!\brief std::ranges::views are valid range-v3 views
+template<class T>
+//!\cond
+    requires ::std::derived_from<T, ::std::ranges::view_base>
+//!\endcond
+inline constexpr bool enable_view<T> = true;
+} // namespace ranges
+
+namespace std::ranges
+{
+//!\brief range-v3 views are valid std::ranges::views
+template<class T>
+//!\cond
+    requires ::std::derived_from<T, ::ranges::view_base>
+//!\endcond
+inline constexpr bool enable_view<T> = true;
+} // namespace std::ranges
+
 #else // implement via range-v3
 
 //!\cond
--- a/include/seqan3/std/ranges
+++ b/include/seqan3/std/ranges
@@ -42,6 +42,18 @@ template<class T>
 inline constexpr bool enable_view<T> = true;
 } // namespace std::ranges
 
+namespace std::ranges
+{
+inline namespace deprecated
+{
+/*!\brief this should be in std::ranges::views and not std::ranges
+ * \deprecated Use std::views::all_t instead.
+ */
+template <typename range_t>
+using all_view = std::views::all_t<range_t>;
+} // namespace std::ranges::deprecated
+} // namespace std::ranges
+
 #else // implement via range-v3
 
 //!\cond


include/seqan3/std/ranges
```diff
@@ -13,8 +13,45 @@
 
 #pragma once
 
-#if __cpp_lib_ranges // C++20 ranges available
+#if __has_include(<ranges>)
 #include <ranges>
+#endif // __has_include(<ranges>)
+
+#if defined(__cpp_lib_ranges) // C++20 ranges available
```
smehringer, line 20: Does `__has_include(<ranges>)` imply `defined(__cpp_lib_ranges)`? 

If yes, why did you move the `<ranges>` include outside of this if condition? (This looks like the `<ranges>` include is independent of all the following code but before the else belonged to the `<ranges>` include if clause.)
If no, there is a possibility that we include the ranges header **and** the still include the `range/v3/...` headers (else clause), if I understand it right? Couldn't this potentially cause errors?

include/seqan3/std/ranges
```diff
@@ -13,8 +13,45 @@
...
+#include <seqan3/std/iterator>
```
smehringer, line 21: if c++20 is available, shouldn't `<iterator>` be available?

include/seqan3/std/ranges
```diff
@@ -13,8 +13,45 @@
...
+
+#include <range/v3/range/concepts.hpp>
+
+namespace ranges
+{
+// std::ranges::views are valid range-v3 views
+template<class T>
+//!\cond
```
smehringer, line 29: This is not documented so you don't need the `cond`s here?

marehr: Oh you are right, the think is doxygen does not evaluate the else branch in these constructions:

```c++
#if 1
...
#else
...
#endif
```

include/seqan3/std/ranges
marehr: Like for `<ranges>` we also need to declare things in `<iterator>` which are not standard conform and were our mistakes when writing `<iterator>`.

Since `<ranges>` always pulls in `<iterator>` and our library expects some of the wrongly defined iterator stuff, I pull it explicitly at this point.

include/seqan3/std/ranges
marehr: In C++17 mode the header exists, but does not contain anything.

We could also include `<version>` and would get the macro `__cpp_lib_ranges`.

Since we use the same construct in `std/span` I kept the old approach to first include the header (which might be empty) and after that use the feature macro to make sure it exists.

marehr:
> Does the removal depend on the range fix or on c++20? Did you note this down somewhere?

On a fix from our side and a fix in the ranges library. The standard renamed some names and the ranges library hasn't adopted them.

include/seqan3/std/ranges
smehringer: > In C++17 mode the header exists, but does not contain anything.

why? o.O

I still think it is confusing that way, I also don't like it in span. Can it not happen that <ranges> has some content but not everything to have `__cpp_lib_ranges` set? 
If not, I guess just leave it.

include/seqan3/std/ranges
smehringer: So your point is to leave them anyway?

smehringer:
> > Does the removal depend on the range fix or on c++20? Did you note this down somewhere?
> 
> On a fix from our side and a fix in the ranges library. The standard renamed some names and the ranges library hasn't adopted them.

 Did you note this down somewhere? :)

include/seqan3/std/ranges
marehr: I added a doxygen note.

include/seqan3/std/ranges
marehr: > I still think it is confusing that way, I also don't like it in span. Can it not happen that has some content but not everything to have `__cpp_lib_ranges` set?

This is something we can never make sure. If someone only half-implements something and falsely advertises via `__cpp_lib_ranges`, which is the same as saying "Hi I'm able to support ranges completely", we are always screwed.



> > In C++17 mode the header exists, but does not contain anything.
> 
> why? o.O

Think about it: The includes are always stored in one location. The only difference is that gcc will enable / disable certain behaviours in the header depending on the C++ mode. If you are requesting C++17 you don't want to pull in any concepts stuff.

That means a existing header is not the same as "it is also available and usable".

marehr:
> > > Does the removal depend on the range fix or on c++20? Did you note this down somewhere?
> > 
> > 
> > On a fix from our side and a fix in the ranges library. The standard renamed some names and the ranges library hasn't adopted them.
> 
> Did you note this down somewhere? :)

yes in https://github.com/seqan/product_backlog/issues/49

PR #1802, opened 2020-05-14
[FIX] Remove const from static assert

MitraDarja:
Kmer_hash view now accepts not const_iterable ranges like the combination of take_until with reverse
Part of #1754

### DIFF
--- a/test/unit/range/views/view_kmer_hash_test.cpp
+++ b/test/unit/range/views/view_kmer_hash_test.cpp
@@ -206,3 +206,22 @@ TYPED_TEST(kmer_hash_ungapped_test, issue1719)
         EXPECT_EQ(4u, v3.size());
     }
 }
+
+//https://github.com/seqan/seqan3/issues/1754
+TYPED_TEST(kmer_hash_ungapped_test, issue1754)
+{
+    TypeParam text1{'A'_dna4, 'C'_dna4, 'G'_dna4, 'T'_dna4, 'A'_dna4, 'G'_dna4, 'C'_dna4}; // ACGTAGC
+    if constexpr (std::ranges::bidirectional_range<TypeParam>) // excludes forward_list
+    {
+        EXPECT_NO_THROW(text1 | prefix_until_first_thymine | std::views::reverse | ungapped_view);
+    }
+}
+
+TYPED_TEST(kmer_hash_gapped_test, issue1754)
+{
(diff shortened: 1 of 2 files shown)


test/unit/range/views/view_kmer_hash_test.cpp
```diff
@@ -206,3 +206,22 @@ TYPED_TEST(kmer_hash_ungapped_test, issue1719)
         EXPECT_EQ(4u, v3.size());
     }
 }
+
+//https://github.com/seqan/seqan3/issues/1754
+TYPED_TEST(kmer_hash_ungapped_test, issue1754)
+{
+    TypeParam text1{'A'_dna4, 'C'_dna4, 'G'_dna4, 'T'_dna4, 'A'_dna4, 'G'_dna4, 'C'_dna4}; // ACGTAGC
```
rrahn, line 213: Does `"ACGTAGC"_dna4` not suffice here?

test/unit/range/views/view_kmer_hash_test.cpp
```diff
@@ -206,3 +206,22 @@ TYPED_TEST(kmer_hash_ungapped_test, issue1719)
...
+    if constexpr (std::ranges::bidirectional_range<TypeParam>) // excludes forward_list
+    {
+        EXPECT_NO_THROW(text1 | prefix_until_first_thymine | std::views::reverse | ungapped_view);
+    }
+}
+
+TYPED_TEST(kmer_hash_gapped_test, issue1754)
+{
+    TypeParam text1{'A'_dna4, 'C'_dna4, 'G'_dna4, 'T'_dna4, 'A'_dna4, 'G'_dna4, 'C'_dna4}; // ACGTAGC
```
rrahn, line 222: same here?

test/unit/range/views/view_kmer_hash_test.cpp
MitraDarja: No, because in the list of the TypeParam is also std::list and that requires this not so nice writing of the text. We could test only on a dna4_vector and write it with `"ACGTAGC"_dna4`, but then we test less...

marehr:
This branch was directly pushed to the `seqan/seqan3` repository. Next time please push on your fork :) Thank you!

MitraDarja:
Wrong branch

PR #1803, opened 2020-05-14
Parallelise search/part2 move executor to core

rrahn:
Second part of seqan/product_backlog#27 which moves the executor to core and removes all documentation references to concrete alignment instances.


include/seqan3/core/algorithm/detail/blocking_algorithm_executor.hpp
```diff
@@ -74,37 +76,25 @@ class blocking_algorithm_executor
...
-    using bucket_iterator_type = std::ranges::iterator_t<bucket_type>;
     //!\}
 
-    //!\brief Return status for seqan3::detail::blocking_algorithm_executor::underflow.
-    enum underflow_status
+    //!\brief Return status for seqan3::detail::blocking_algorithm_executor::fill_buffer.
+    enum fill_status
     {
         non_empty_buffer, //!< The buffer is not fully consumed yet and contains at least one element.
         empty_buffer, //!< The buffer is empty after calling underflow.
```
simonsasse, line 93: should this be: `The buffer is empty after calling *fill_status*.` ?

include/seqan3/core/algorithm/detail/blocking_algorithm_executor.hpp
```diff
@@ -150,115 +140,102 @@ class blocking_algorithm_executor
...
     {
         if (!is_buffer_empty())  // Not everything consumed yet.
-            return underflow_status::non_empty_buffer;
+            return fill_status::non_empty_buffer;
 
         if (is_eof())  // Case: reached end of resource.
-            return underflow_status::end_of_resource;
+            return fill_status::end_of_resource;
 
         // Reset the buckets and the bucket iterator
```
simonsasse, line 230: Doesn't this now resets the `buckets` and the `buffer_iterator`? (line  268ff)

include/seqan3/core/algorithm/detail/blocking_algorithm_executor.hpp
```diff
@@ -293,14 +265,14 @@ class blocking_algorithm_executor
      * Clears all buckets and sets the bucket iterator to the first iterator, such that the allocated memory for each
```
simonsasse, line 265: probably same as above: doesn't this set the `buffer_it` and not the `bucket_iterator` to the first bucket in the buffer and clears all buckets ?

include/seqan3/core/algorithm/detail/blocking_algorithm_executor.hpp
```diff
@@ -293,14 +265,14 @@ class blocking_algorithm_executor
...
-        for (auto & bucket : bucket_vector)
+        for (auto & bucket : buffer)
             bucket.clear();
 
         // Reset the iterator over the buckets.
-        bucket_iterator = bucket_vector.begin();
+        buffer_it = buffer.begin();
     }
 
     /*!\brief Finds the first non-empty bucket starting from the current bucket iterator.
```
simonsasse, line 278: should be changed to:
Finds the first non-empty bucket starting from the current `buffer_it`. ?

include/seqan3/core/algorithm/detail/execution_handler_sequential.hpp
```diff
@@ -12,40 +12,36 @@
...
+        requires std::invocable<algorithm_t, algorithm_input_t, callback_t>
+    //!\endcond
+    void execute(algorithm_t && algorithm, algorithm_input_t && input, callback_t && callback)
     {
-        algorithm(std::forward<indexed_sequence_pairs_t>(indexed_sequence_pairs),
-                  std::forward<delegate_type>(delegate));
+        algorithm(std::forward<algorithm_input_t>(input), std::forward<callback_t>(callback));
     }
 
     //!\brief Waits for the submitted alignments jobs to finish. (Noop).
```
simonsasse, line 47: here is still the `alignment` keyword. Should be changed to something like:
waits for the sequentially executed algorithm to finish.

rrahn:
@simonsasse as promised I separated the commits so that the various changes can be followed more easily. I had to rebase for this but adapted all issues you found. Many thanks for this.

test/unit/core/algorithm/detail/blocking_algorithm_executor_test.cpp
```diff
@@ -85,14 +83,14 @@ TYPED_TEST(blocking_algorithm_executor_test, is_eof)
                                                    size_t,
                                                    TypeParam>;
 
-    alignment_executor_t exec{this->sequence_pairs, algorithm_t{dummy_alignment{}}};
+    alignment_executor_t exec{this->sequence_pairs, algorithm_t{dummy_algorithm{}}};
```
simonsasse, line 86: I don't know if thats necessary: Should the `alignment_executor_t` template typename be renamed to something like `algorithm_executor_t` ?

rrahn:
> In the algorithm_executor_test, the alignment_executor_t is just a possible example for the blocking_algorithm-executor, right ?

Well, it should be renamed as well. Thanks for spotting this.

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -181,7 +182,7 @@ constexpr auto align_pairwise(sequence_t && sequences,
                                      | views::chunk(traits_t::alignments_per_vector);
 
     // Create a two-way executor for the alignment.
-    detail::alignment_executor_two_way executor{indexed_sequence_chunk_view,
+    detail::blocking_algorithm_executor executor{indexed_sequence_chunk_view,
```
smehringer, line 186: I liked the prefix-based name before. What about `algorithm_executor_blocking`?

Side question: Can the executor only work with algorithms? (or is everything that takes an executor an algorithm?) Just wondering if the `algorithm` in the name is necessary.

include/seqan3/alignment/pairwise/align_pairwise.hpp
```diff
@@ -19,8 +19,9 @@
 
 #include <meta/meta.hpp>
 
-#include <seqan3/alignment/pairwise/alignment_result.hpp>
 #include <seqan3/alignment/pairwise/alignment_configurator.hpp>
+#include <seqan3/alignment/pairwise/alignment_range.hpp>
```
smehringer, line 23: You introduce this include here but you don't change anything in the code? Is this necessary or is this part of a different commit?

include/seqan3/alignment/pairwise/alignment_range.hpp
```diff
@@ -73,19 +65,18 @@ class alignment_range
     ~alignment_range() = default; //!< Defaulted.
 
     //!\brief Explicit deletion to forbid copy construction of the underlying executor.
-    explicit alignment_range(alignment_executor_type const & _alignment_executor) = delete;
+    explicit alignment_range(alignment_executor_type const & alignment_executor) = delete;
 
     /*!\brief Constructs a new alignment range by taking ownership over the passed alignment buffer.
-     * \tparam _alignment_executor_type The buffer type. Must be the same type as `alignment_executor_type` when
-     *                                  references and cv-qualifiers are removed.
```
smehringer, line 80: why remove the type description? I think it does not copy from the class template type description automatically does it?

include/seqan3/core/algorithm/detail/blocking_algorithm_executor.hpp
```diff
@@ -34,7 +34,7 @@ namespace seqan3::detail
  * \ingroup execution
  * \tparam resource_t            The underlying range of sequence pairs to be computed; must model
  *                               std::ranges::viewable_range and std::ranges::input_range.
- * \tparam alignment_algorithm_t The alignment algorithm to be invoked on each sequence pair.
+ * \tparam algorithm_t The algorithm to be invoked on the elements of the given resource; must model std::semiregular.
  * \tparam value_t               The value type to buffer results generated from the algorithm; must model
```
smehringer:
```suggestion
 * \tparam algorithm_t           The algorithm to be invoked on the elements of the given resource; must model std::semiregular.
 * \tparam value_t               The value type to buffer results generated from the algorithm; must model
```

include/seqan3/core/algorithm/detail/blocking_algorithm_executor.hpp
```diff
@@ -35,8 +35,7 @@ namespace seqan3::detail
  * \tparam resource_t            The underlying range of sequence pairs to be computed; must model
  *                               std::ranges::viewable_range and std::ranges::input_range.
  * \tparam algorithm_t The algorithm to be invoked on the elements of the given resource; must model std::semiregular.
- * \tparam value_t               The value type to buffer results generated from the algorithm; must model
- *                               std::semiregular.
+ * \tparam algorithm_result_t The result type generated by the algorithm; must model std::semiregular.
  * \tparam execution_handler_t   The execution handler managing the execution of the alignments.
```
smehringer:
```suggestion
 * \tparam algorithm_result_t    The result type generated by the algorithm; must model std::semiregular.
 * \tparam execution_handler_t   The execution handler managing the execution of the alignments.
```

include/seqan3/core/algorithm/detail/blocking_algorithm_executor.hpp
```diff
@@ -174,7 +161,7 @@ class blocking_algorithm_executor
     //!\endcond
     blocking_algorithm_executor(resource_t resource,
                                algorithm_t algorithm,
-                               value_t const SEQAN3_DOXYGEN_ONLY(buffer_value) = value_t{},
+                                algorithm_result_t const SEQAN3_DOXYGEN_ONLY(result) = algorithm_result_t{},
                                exec_policy_t const & SEQAN3_DOXYGEN_ONLY(exec) = seq) :
```
smehringer:
```suggestion
                               algorithm_result_t const SEQAN3_DOXYGEN_ONLY(result) = algorithm_result_t{},
                               exec_policy_t const & SEQAN3_DOXYGEN_ONLY(exec) = seq) :
```

include/seqan3/alignment/pairwise/align_pairwise.hpp
rrahn: Well the semantics are clear, since it calls something with an input. That is in a very generic case an algorithm. We could call it function executor but does it make it better? Also the context right now is for algorithms so why should it be named differently? 
I rename it to algorithm_executor_blocking though, to be more conistent with other namings.

include/seqan3/alignment/pairwise/align_pairwise.hpp
rrahn: Before it was pulled in indirectly.

include/seqan3/alignment/pairwise/alignment_range.hpp
rrahn: Because the constructor is not a template function. It get's an explicit type, which is documented in the class documentation

include/seqan3/core/algorithm/detail/blocking_algorithm_executor.hpp
rrahn: well, it is the otherway around. Thanks for spotting.

rrahn: this is already fixed in some commit.

smehringer:
Is the commit history like you want it to be when merged?
I guess the renaming of the executor (two way -> blocking_..., blocking_... -> -..._blocking ) could be merged

PR #1804, opened 2020-05-15
Optimised vectorised global alignment.

rrahn:
Enables the vectorised global alignment.

Micro benchmark with AVX2:

### Seqan3 new
```console
Run on (4 X 3600 MHz CPU s)
CPU Caches:
  L1 Data 32K (x4)
  L1 Instruction 32K (x4)
  L2 Unified 262K (x4)
  L3 Unified 6291K (x1)
Load Average: 3.50, 3.48, 3.09
--------------------------------------------------------------------------------------------------------------------------------------
Benchmark                                                                            Time             CPU   Iterations UserCounters...
--------------------------------------------------------------------------------------------------------------------------------------
seqan3_affine_dna4_accelerated/simd_with_score/0/real_time                     2795783 ns      2786265 ns          253 CUPS=8.35123G/s cells=23.3482M total=-18.004M
seqan3_affine_dna4_accelerated/simd_with_score/8/real_time                     2987575 ns      2980251 ns          235 CUPS=7.79106G/s cells=23.2764M total=-16.8248M
seqan3_affine_dna4_accelerated/simd_with_score/16/real_time                    3239686 ns      3229834 ns          211 CUPS=7.16485G/s cells=23.2119M total=-15.1144M
seqan3_affine_dna4_accelerated/simd_with_score/24/real_time                    3511635 ns      3498046 ns          197 CUPS=6.59069G/s cells=23.1441M total=-14.2201M
seqan3_affine_dna4_accelerated/simd_with_score/32/real_time                    3779830 ns      3771973 ns          183 CUPS=6.10703G/s cells=23.0835M total=-13.3112M
seqan3_affine_dna4_accelerated/simd_with_score/40/real_time                    4113648 ns      4079439 ns          171 CUPS=5.59517G/s cells=23.0166M total=-12.6039M
seqan3_affine_dna4_accelerated/simd_with_score/48/real_time                    4402465 ns      4390226 ns          159 CUPS=5.2142G/s cells=22.9553M total=-11.8838M
seqan3_affine_dna4_accelerated/simd_with_score/56/real_time                    4735617 ns      4722425 ns          134 CUPS=4.83419G/s cells=22.8929M total=-10.1945M
seqan3_affine_dna4_accelerated/simd_with_score/64/real_time                    5010244 ns      5000563 ns          119 CUPS=4.55656G/s cells=22.8295M total=-9.26189M
```

### Seqan3 old
```console
--------------------------------------------------------------------------------------------------------------------------------------
[...]

PR #1813, opened 2020-05-18
[FIX] Fixes handling of non full batches in the to_simd view

rrahn:
In the old design we assumed that calling begin and end on the underlying range is always valid.
But this assumption does not necessarily hold in all cases.
This change ensures that empty slots in the batch are initialised correctly.

part of #1804

### DIFF
--- a/include/seqan3/core/simd/view_to_simd.hpp
+++ b/include/seqan3/core/simd/view_to_simd.hpp
@@ -188,7 +188,6 @@ public:
 private:
 
     urng_t urng{}; //!< The underlying range.
-    std::ranges::range_value_t<urng_t> empty_inner_range{}; //!< A defaulted inner range if fewer sequences are present.
     std::array<chunk_type, total_chunks> cached_simd_chunks{}; //!< The cached chunks of transformed simd vectors.
     simd_t padding_simd_vector{}; //!< A cached simd vector with the padding symbol.
     scalar_type padding_value{}; //!< The padding value used to fill the corresponding simd vector element.
@@ -244,12 +243,20 @@ public:
             cached_sentinel[seq_id] = std::ranges::end(*it);
         }
 
-        // If the batch is not full, i.e. less than chunk_size many sequences, then fill them up with dummy sequences.
+        // The batch is empty and by default the constructed iterator is pointing to the end.
+        if (seq_id == 0)
+            return;
+
+        // The batch is not empty but might not be full either.
+        // If a slot is supposed to be empty, it will be initialised with the iterator of the first sequence set to the
+        // end emulating an empty sequence.
+        auto sentinel_it = std::ranges::next(cached_iter[0], cached_sentinel[0]);
         for (; seq_id < chunk_size; ++seq_id)
         {
-            cached_iter[seq_id] = std::ranges::begin(this_view.empty_inner_range);
-            cached_sentinel[seq_id] = std::ranges::end(this_view.empty_inner_range);
+            cached_iter[seq_id] = sentinel_it;
+            cached_sentinel[seq_id] = cached_sentinel[0];
         }
+
         // Check if this is the final chunk already.
         final_chunk = all_iterators_reached_sentinel();
 
@@ -582,9 +589,9 @@ private:
     //!\brief The current chunk position.
     uint8_t current_chunk_pos{0};
     //!\brief Flag indicating that final chunk was reached.
-    bool final_chunk{false};
+    bool final_chunk{true};
     //!\brief Flag indicating that iterator is at end.
-    bool at_end{false};
+    bool at_end{true};
 };
 
 // ============================================================================
--- a/test/unit/core/simd/view_to_simd_test.cpp
+++ b/test/unit/core/simd/view_to_simd_test.cpp
@@ -317,3 +317,35 @@ TYPED_TEST(view_to_simd_test, ends_not_on_chunk_boundary)
     auto simd_view = this->sequences | seqan3::views::to_simd<simd_t>;
     this->compare(simd_view, this->transformed_simd_vec);
 }
+
+// See https://github.com/seqan/seqan3/pull/1813
+// Before this fix we created a default constructed instance of the underlying range to use it as an empty range.
+// But a default constructed view is not guaranteed to be comparable, i.e. it results in UB.
+// This test checks now that if the underlying range is a view the conversion to simd sequence still works.
+TYPED_TEST(view_to_simd_test, issue_1813)
+{
+    using simd_t = typename TestFixture::simd_t;
+    using container_t = typename TestFixture::container_t;
+    using take_view_t = decltype(std::declval<container_t &>() | std::views::take(5));
+
+    std::vector<take_view_t> take_sequences{};
+
+    for (auto & sequence : this->sequences)
+        take_sequences.push_back(sequence | std::views::take(10));
+
+    // Remove one sequence
+    take_sequences.pop_back();
+
+    // If simd size is only 1, the compare range needs to be cleared.
+    if (take_sequences.empty())
+        this->transformed_simd_vec.clear();
+
+    // delete the last sequence in the set.
+    for (simd_t & vec : this->transformed_simd_vec)
+        vec[seqan3::simd::simd_traits<simd_t>::length - 1] = TestFixture::padding_value_dna4;
+
+    { // without padding
+        auto v = seqan3::views::to_simd<simd_t>(take_sequences);
+        this->compare(v, this->transformed_simd_vec | std::views::take(10));
+    }
+}


include/seqan3/core/simd/view_to_simd.hpp
```diff
@@ -244,14 +243,22 @@ class view_to_simd<urng_t, simd_t>::iterator_type
...
-        // If the batch is not full, i.e. less than chunk_size many sequences, then fill them up with dummy sequences.
-        for (; seq_id < chunk_size; ++seq_id)
+        // The batch is not empty but might not be full either.
+        // If a slot is supposed to be empty, it will be initialised with the iterator of the first sequence set to the
+        // end emulating an empty sequence.
+        if (seq_id > 0)
         {
-            cached_iter[seq_id] = std::ranges::begin(this_view.empty_inner_range);
-            cached_sentinel[seq_id] = std::ranges::end(this_view.empty_inner_range);
+            auto it = std::ranges::next(cached_iter[0], cached_sentinel[0]); // Move the iterator to the end.
```
marehr, line 251:
```suggestion
            // find the end iterator of the current iterator (`sentinel_it == cached_sentinel`)
            auto sentinel_it = std::ranges::next(cached_iter[0], cached_sentinel[0]); 
```

include/seqan3/core/simd/view_to_simd.hpp
```diff
@@ -244,14 +243,22 @@ class view_to_simd<urng_t, simd_t>::iterator_type
...
+            // If the batch is not full, i.e. less than chunk_size many sequences, then fill them up with dummy sequences.
```
marehr, line 252:
```suggestion
            // If the batch is not full, i.e. less than chunk_size many sequences, then fill them up with empty sequences.
```

rrahn:
> Can you add a Test Case?

> Do you have an example which underlying ranges can't be accessed by std::ranges::begin?
Can you add this as test case?

This is not an issue of begin not accessible. But there is no guarantee that a default constructed view is comparable. Something like this is:
```cpp
using take_t = decltype(std::declval<std::vector<int> &>() | std::views::take(10));
take_t my_take{};
if (my_take.begin() == my_take.end()) { ... }
```
is UB. So testing is a little difficult.

> I found the earlier solution easier to understand.

Well, but did not work :)

> I guess that you ensure forward_iterators/ranges in the view itself right?

Yes

include/seqan3/core/simd/view_to_simd.hpp
```diff
@@ -244,14 +243,24 @@ class view_to_simd<urng_t, simd_t>::iterator_type
             cached_sentinel[seq_id] = std::ranges::end(*it);
         }
```
marehr, line 245:
```suggestion
if (seq_id == 0u)
    return;
    
// following code knows that we have at least one sequence
```

marehr:
Okay you send me https://godbolt.org/z/59DHrr

I guess it makes no sense to add a test case for this, because the previous implementation was semantically wrong.

test/unit/core/simd/view_to_simd_test.cpp
```diff
@@ -89,22 +89,36 @@ class view_to_simd_test : public ::testing::Test
...
-                                    std::tuple<std::vector<seqan3::dna4>, seqan3::simd::simd_type_t<uint32_t>>,
-                                    std::tuple<std::vector<seqan3::dna4>, seqan3::simd::simd_type_t<uint64_t>>,
-                                    std::tuple<std::deque<seqan3::dna4>, seqan3::simd::simd_type_t<int8_t>>,
-                                    std::tuple<std::deque<seqan3::dna4>, seqan3::simd::simd_type_t<int16_t>>,
-                                    std::tuple<std::deque<seqan3::dna4>, seqan3::simd::simd_type_t<int32_t>>,
-                                    std::tuple<std::deque<seqan3::dna4>, seqan3::simd::simd_type_t<int64_t>>,
-                                    std::tuple<std::deque<seqan3::dna4>, seqan3::simd::simd_type_t<uint8_t>>,
-                                    std::tuple<std::deque<seqan3::dna4>, seqan3::simd::simd_type_t<uint16_t>>,
-                                    std::tuple<std::deque<seqan3::dna4>, seqan3::simd::simd_type_t<uint32_t>>,
-                                    std::tuple<std::deque<seqan3::dna4>, seqan3::simd::simd_type_t<uint64_t>>
```
marehr, line 107: Can we keep this what you removed?

I would prefer if we keep `seqan3::simd::simd_type_t<>` with it's auto max-length.

Or is it really necessary to assume at least 2 elements?

test/unit/core/simd/view_to_simd_test.cpp
```diff
@@ -317,3 +331,31 @@ TYPED_TEST(view_to_simd_test, ends_not_on_chunk_boundary)
...
+        take_sequences.push_back(sequence | std::views::take(10));
+
+    // Remove one sequence
+    take_sequences.pop_back();
+
+    // delete the last sequence in the set.
+    for (simd_t & vec : this->transformed_simd_vec)
+        vec[seqan3::simd::simd_traits<simd_t>::length - 1] = TestFixture::padding_value_dna4;
+
+    { // without padding
```
marehr, line 357:
```suggestion
    { // without padding
```
? But isn't this with padding?

test/unit/core/simd/view_to_simd_test.cpp
```diff
@@ -317,3 +331,31 @@ TYPED_TEST(view_to_simd_test, ends_not_on_chunk_boundary)
     auto simd_view = this->sequences | seqan3::views::to_simd<simd_t>;
     this->compare(simd_view, this->transformed_simd_vec);
 }
+
+// See https://github.com/seqan/seqan3/pull/1813
+// Before this fix we created a default constructed instance of the underlying range to use it as an empty range.
+// But a default constructed view is not guaranteed to be comparable, i.e. it results in UB.
+// This test checks now that if the underlying range is a view the conversion still works.
+TYPED_TEST(view_to_simd_test, default_constructed_empty_range_bug)
```
marehr, line 339:
```suggestion
TYPED_TEST(view_to_simd_test, issue1813)
```

test/unit/core/simd/view_to_simd_test.cpp
rrahn: Well, if it is not 2 then the one for loop in the iterator constructor is never executed and the coverage test cries.

test/unit/core/simd/view_to_simd_test.cpp
rrahn: Well, it is always with padding. But what I mean is with custom padding or not. So if not padded it will use some default value, which is the alphabet size I believe.

test/unit/core/simd/view_to_simd_test.cpp
marehr: I think we should fix those issues with a infrastructure change that builds for SSE4, AVX2 and so on.

I expect it right now that coverage tests and other thing related to simd are not reliable.

test/unit/core/simd/view_to_simd_test.cpp
marehr: What is custom padding? Ah I see this is terminology that you already introduced in the test. Okay!

test/unit/core/simd/view_to_simd_test.cpp
rrahn: as you wish

test/unit/core/simd/view_to_simd_test.cpp
```diff
@@ -89,22 +89,36 @@ class view_to_simd_test : public ::testing::Test
...
+                                    std::tuple<std::vector<seqan3::dna4>, simd_type_t<uint32_t>>,
+                                    std::tuple<std::vector<seqan3::dna4>, simd_type_t<uint64_t>>,
+                                    std::tuple<std::deque<seqan3::dna4>, simd_type_t<int8_t>>,
+                                    std::tuple<std::deque<seqan3::dna4>, simd_type_t<int16_t>>,
+                                    std::tuple<std::deque<seqan3::dna4>, simd_type_t<int32_t>>,
+                                    std::tuple<std::deque<seqan3::dna4>, simd_type_t<int64_t>>,
+                                    std::tuple<std::deque<seqan3::dna4>, simd_type_t<uint8_t>>,
+                                    std::tuple<std::deque<seqan3::dna4>, simd_type_t<uint16_t>>,
+                                    std::tuple<std::deque<seqan3::dna4>, simd_type_t<uint32_t>>,
+                                    std::tuple<std::deque<seqan3::dna4>, simd_type_t<uint64_t>>
```
marehr, line 121: This is still open

rrahn: forgot to push

PR #1814, opened 2020-05-18
[FIX] minimiser view should tell you on which value_t it can operate

marehr:


include/seqan3/range/views/minimiser.hpp
```diff
@@ -420,6 +420,9 @@ struct minimiser_fn
      * \returns                         A range of converted values.
      */
     template <std::ranges::range urng_t>
+    //!\cond
+        requires std::convertible_to<std::ranges::range_value_t<urng_t>, uint64_t>
```
smehringer, line 424: This is not true for the minimizer view but only for the minimizer_hash.
The minimizer view should only require that the type is equality comparable (and copy constructible I think).

smehringer: This is static asserted in the view class.
But I see now that the std::deque actually stores `uint64_t` instead of value_type as it should. I missed that in the review. This needs to be fixed instead of this requires.

marehr: Okay thank you, I only saw that the deque takes this type. Will fix this accordingly :)

smehringer: Thank you! if you don't have time I can do it as well, just tell me.

marehr: I think this would be the best idea.

marehr:
Closing because @smehringer wanted to fix the real issue.

PR #1816, opened 2020-05-18
[INFRA] doc: limit execution time of rsync to 60s

marehr:


eseiler:
I'd propose to also change the name of the workflow while we are touching the file.

(line 1) `name: Workflow to deploy documentation for SeqAn3`

When I write the workflow, I did not know how It would look like in the GitHub UI and I think it's rather too verbose and long.
Maybe `name: Deploy Documentation` conveys the same meaning while not trying to tell its life story

marehr:
> I'd propose to also change the name of the workflow while we are touching the file.
> 
> (line 1) `name: Workflow to deploy documentation for SeqAn3`
> 
> When I write the workflow, I did not know how It would look like in the GitHub UI and I think it's rather too verbose and long.
> Maybe `name: Deploy Documentation` conveys the same meaning while not trying to tell its life story

Done

.github/workflows/documentation.yaml
```diff
@@ -1,4 +1,4 @@
-name: Workflow to deploy documentation for SeqAn3
+name: Deploy https://docs.seqan.de/
```
rrahn, line 1:
```suggestion
name: Deploy SeqAn3 documentation to https://docs.seqan.de/
```

marehr: See https://github.com/seqan/seqan3/pull/1816#issuecomment-630074757

enrico asked me to shorten the name

rrahn: But I don't deploy the server? Then rather have it `Deploy documentation (https://docs.seqan.de/)`

marehr: afaik; we deploy the sources to the server

PR #1815, opened 2020-05-18
[NOP] minimiser view: fixup code-style

marehr:

PR #1818, opened 2020-05-18
[FIX] cxx20: use std::default_sentinel_t on minimiser view

marehr:
The expression `std::ranges::begin(this->test_range) == std::ranges::cend(this->test_range)`
for `this->test_range = seqan3::detail::minimiser_view` did not compile.

I tried to find out, but ultimately could only figure out that
the left hand side was

* `lhs_iterator = std::ranges::begin(this->test_range)`
* `lhs_iterator == minimiser_view<rng>::window_iterator<rng>`
  * with `rng = non-const std::ranges::transform_view` internally defined as
  * `lhs_iterator::sentinel == std::ranges::end(rng)` and
  * `std::ranges::end(rng) == std::ranges::transform_view::sentinel<false>`

Whereas the right side was

* `rhs_sentinel = std::ranges::cend(this->test_range)`
* `rhs_sentinel == std::ranges::cend(rng)`
    * with `rng = non-const std::ranges::transform_view`
* `rhs_sentinel == std::ranges::transform_view::sentinel<true>`

The `==` in `lhs_iterator` is defined as
* `operator==(lhs_iterator const &, lhs_iterator::sentinel const &)` which is the same as
* `operator==(minimiser_view<rng>::window_iterator<rng> const &, std::ranges::transform_view::sentinel<false> const &)`

The requested `==` of the expression
* `std::ranges::begin(this->test_range) == std::ranges::cend(this->test_range)`
* needs the following signature
* `operator==(minimiser_view<rng>::window_iterator<rng> const &, std::ranges::transform_view::sentinel<true> const &)`

As you can see, the signature does not match completely.

Normally you can convert a `std::ranges::transform_view::sentinel<false>` into a `std::ranges::transform_view::sentinel<true>` and from what I can see, the compiler is normally able to do this conversion sequence.

See [std::ranges::transform_view::iterator](https://eel.is/c++draft/range.transform.iterator) and [std::ranges::transform_view::sentinel](https://eel.is/c++draft/range.transform.sentinel).

They both don't have a special case for `sentinel<false>` or `sentinel<true>`.
(Note that they have less comparison operators, this is because since std20 you automatically define all reverse operators for `==` e.g. `a == b` also defines `b == a` and you get `!=` for free if `==` was defined.)

minimiser view has the same comparison operators defined and has a constructor for non-const to const conversion, but fails to find the conversion sequence.

My guess is that the conversion sequence does not work, because lhs and rhs are in different namespaces.
(Haven't confirmed that yet)

---

The fix in this PR is to make the signature of the `==` operator the same for const and non-const version.

This is done by using `std::default_sentinel_t`.

### DIFF
--- a/include/seqan3/range/views/minimiser.hpp
+++ b/include/seqan3/range/views/minimiser.hpp
@@ -136,25 +136,13 @@ public:
      *
      * No-throw guarantee.
      */
-    auto end()
-    {
-        return std::ranges::end(urange);
-    }
-
-    //!\copydoc end()
     auto end() const
-    //!\cond
-        requires seqan3::const_iterable_range<urng_t>
-    //!\endcond
     {
-        return std::ranges::end(urange);
+        return std::ranges::default_sentinel;
     }
 
     //!\copydoc end()
     auto cend() const
-    //!\cond
-        requires seqan3::const_iterable_range<urng_t>
-    //!\endcond
     {
         return end();
     }
@@ -171,6 +159,8 @@ private:
     using it_t = std::ranges::iterator_t<rng_t>;
     //!\brief The sentinel type of the underlying range.
     using sentinel_t = std::ranges::sentinel_t<rng_t>;
+    //!\brief The sentinel type of the minimiser_view.
+    using sentinel_type = std::ranges::default_sentinel_t;
 
     template <typename urng2_t>
     friend class window_iterator;
@@ -248,15 +238,15 @@ public:
     //!\{
 
     //!\brief Compare to the sentinel of the underlying range.
-    friend bool operator==(window_iterator const & lhs, sentinel_t const & rhs)
+    friend bool operator==(window_iterator const & lhs, sentinel_type const &)
     {
-        return lhs.window_right == rhs;
+        return lhs.window_right == lhs.urange_end;
     }
 
     //!\brief Compare to the sentinel of the underlying range.
-    friend bool operator==(sentinel_t const & lhs, window_iterator const & rhs)
+    friend bool operator==(sentinel_type const & lhs, window_iterator const & rhs)
     {
-        return lhs == rhs.window_right;
+        return rhs == lhs;
     }
 
     //!\brief Compare to another window_iterator.
@@ -267,13 +257,13 @@ public:
     }
 
     //!\brief Compare to the sentinel of the underlying range.
-    friend bool operator!=(window_iterator const & lhs, sentinel_t const & rhs)
+    friend bool operator!=(window_iterator const & lhs, sentinel_type const & rhs)
     {
         return !(lhs == rhs);
     }
 
     //!\brief Compare to the sentinel of the underlying range.
-    friend bool operator!=(sentinel_t const & lhs, window_iterator const & rhs)
+    friend bool operator!=(sentinel_type const & lhs, window_iterator const & rhs)
     {
         return !(lhs == rhs);
     }


marehr:
@rrahn I assigned you, because I need a second opinion here.

rrahn:
Well, the problem here is that the conversion sequence `sentinel<true>` to `sentinel<false>` ist not allowed. I cannot construct a non-const iterator/sentinel from a const iterator/sentinel since this removes the const qualification of the underlying type.
```cpp
    constexpr sentinel(sentinel<!Const> i)
      requires Const && convertible_to<sentinel_t<V>, sentinel_t<Base>>;
```
In fact the transform view does also not allow all possible comparisons: https://godbolt.org/z/PAXQAK

It is the question whether this is a defect already in the standard or not. AFAIK there is no concept requiring this. Basically, the fix in our implementation could be to define the sentinel type over the const urng if it is a const-iterable otherwise use the normal sentinel type. This way we always promote to the const-qualified version, which should always work.

include/seqan3/range/views/minimiser.hpp
```diff
@@ -171,6 +162,8 @@ class minimiser_view<urng_t>::window_iterator
     using it_t = std::ranges::iterator_t<rng_t>;
     //!\brief The sentinel type of the underlying range.
     using sentinel_t = std::ranges::sentinel_t<rng_t>;
+    //!\brief The sentinel type of the minimiser_view.
+    using sentinel_type = std::ranges::default_sentinel_t;
```
rrahn, line 166: One thing. I find it a little confusing now that there is a sentinel_t and a sentinel_type. Can we replace all instances with std::ranges::default_sentinel_t and omit this type definition here? 
It would make it less confusing.

smehringer: I agree, could you change this please? Or rename `sentinal_t` to `underlying_rng_sentinel_t`?

include/seqan3/range/views/minimiser.hpp
```diff
@@ -136,18 +136,9 @@ class minimiser_view : public std::ranges::view_interface<minimiser_view<urng_t>
      *
      * No-throw guarantee.
      */
-    auto end()
```
smehringer, line 139: I am confused. Why did you remove the non-const overload completely?

include/seqan3/range/views/minimiser.hpp
```diff
@@ -136,18 +136,9 @@ class minimiser_view : public std::ranges::view_interface<minimiser_view<urng_t>
...
-    {
-        return std::ranges::end(urange);
-    }
-
-    //!\copydoc end()
     auto end() const
-    //!\cond
-        requires seqan3::const_iterable_range<urng_t>
```
smehringer, line 147: The `end const` version now has no requires clause, but the `cend()` function does?

include/seqan3/range/views/minimiser.hpp
marehr: I would prefer `urng_sentinel_t`

marehr:
It turned out that the underlying problem was a defect in the standard:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95322

Which results in a nice LWG issue: :smile: :tada: 
https://cplusplus.github.io/LWG/issue3448

include/seqan3/range/views/minimiser.hpp
marehr: You only need  a `const` version if you have a different return type.

include/seqan3/range/views/minimiser.hpp
marehr: I removed it

include/seqan3/range/views/minimiser.hpp
marehr: I have a PR #1815 I'll do it in that PR.

PR #1819, opened 2020-05-18
Optimise alignment/part2 simd matrix coordinate

rrahn:
part of #1804 

Allows to use matrix_coordinate with simd types such that we can have a simd coordinate matrix. Updates the alignment configuration traits to provide correct simd coordinate types and index types. Note, that the index type is the unsigned version of the original score type but needs some extra handling of floating-point types.

### DIFF
--- a/include/seqan3/alignment/matrix/detail/matrix_coordinate.hpp
+++ b/include/seqan3/alignment/matrix/detail/matrix_coordinate.hpp
@@ -16,15 +16,20 @@
 
 #include <seqan3/core/concept/core_language.hpp>
 #include <seqan3/core/detail/strong_type.hpp>
+#include <seqan3/core/simd/concept.hpp>
+#include <seqan3/core/simd/simd_algorithm.hpp>
 #include <seqan3/std/concepts>
 
 namespace seqan3::detail
 {
 /*!\brief A strong type for designated initialisation of the column index of a marix.
  * \ingroup alignment_matrix
- * \tparam index_type The type of the index to store; must model seqan3::arithmetic.
+ * \tparam index_type The type of the index to store; must model seqan3::arithmetic or seqan3::simd::simd_concept.
  */
-template <std::integral index_type>
+template <typename index_type>
+//!\cond
+    requires (std::integral<index_type> || simd_concept<index_type>)
+//!\endcond
 struct column_index_type : detail::strong_type<index_type, column_index_type<index_type>>
 {
     //!!\brief Import base class constructor.
@@ -42,13 +47,20 @@ column_index_type(index_type) -> column_index_type<std::ptrdiff_t>;
 //!\brief Deduces an unsigned integral type to size_t.
 template <std::unsigned_integral index_type>
 column_index_type(index_type) -> column_index_type<size_t>;
+
+//!\brief Deduces the template argument from a simd vector index type.
+template <simd_concept index_type>
+column_index_type(index_type) -> column_index_type<index_type>;
 //!\}
 
 /*!\brief A strong type for designated initialisation of the row index of a marix.
  * \ingroup alignment_matrix
- * \tparam index_type The type of the index to store; must model seqan3::arithmetic.
+ * \tparam index_type The type of the index to store; must model seqan3::arithmetic or seqan3::simd::simd_concept
  */
-template <std::integral index_type>
+template <typename index_type>
+//!\cond
+    requires (std::integral<index_type> || simd_concept<index_type>)
+//!\endcond
 struct row_index_type : detail::strong_type<index_type, row_index_type<index_type>>
 {
     //!!\brief Import base class constructor.
@@ -66,11 +78,20 @@ row_index_type(index_type) -> row_index_type<std::ptrdiff_t>;
 //!\brief Deduces an unsigned integral type to size_t.
 template <std::unsigned_integral index_type>
 row_index_type(index_type) -> row_index_type<size_t>;
+
+//!\brief Deduces the template argument from a simd vector index type.
+template <simd_concept index_type>
+row_index_type(index_type) -> row_index_type<index_type>;
 //!\}
 
-//!\brief A representation of a location or offset within a two-dimensional matrix.
-//!\ingroup alignment_matrix
-template <std::integral index_t>
+/*!\brief A representation of a location or offset within a two-dimensional matrix.
+ * \ingroup alignment_matrix
+ * \tparam index_t The underlying index type; must model seqan3::arithmetic or seqan3::simd::simd_concept.
+ */
+template <typename index_t>
+//!\cond
+    requires (std::integral<index_t> || simd_concept<index_t>)
+//!\endcond
 struct matrix_index
 {
     /*!\name Constructors, destructor and assignment
@@ -87,9 +108,40 @@ struct matrix_index
      * \param row_idx The row index to set.
      * \param col_idx The column index to set.
      */
-    constexpr matrix_index(row_index_type<index_t> const row_idx,
-                           column_index_type<index_t> const col_idx) noexcept
-        : row{row_idx.get()}, col{col_idx.get()}
+    constexpr matrix_index(row_index_type<index_t> const row_idx, column_index_type<index_t> const col_idx) noexcept :
+        row{row_idx.get()},
+        col{col_idx.get()}
+    {}
+
+    /*!\brief Construction from strongly typed row index and column index over a scalar type when the index is a simd
+     *        vector.
+     *
+     * \tparam scalar_index_t The type of the scalar index type; must model seqan3::arithmetic and must model
+     *                        std::convertible_to the scalar type of the simd vector `index_t`.
+     *
+     * \param row_idx The row index to set.
+     * \param col_idx The column index to set.
+     *
+     * \details
+     *
+     * This constructor initialises the row and col index which is represented as simd vectors. The vectors are
+     * initialised with the scalar types of the given strong types over the scalar value. This constructor is only
+     * available if `index_t` is a simd vector type and the `scalar_index_t` is convertible to the scalar type of
+     * the simd index.
+     */
+    template <seqan3::arithmetic scalar_index_t>
+    constexpr matrix_index(row_index_type<scalar_index_t> const row_idx,
+                           column_index_type<scalar_index_t> const col_idx) noexcept
+    //!\cond
+        requires (simd_concept<index_t> &&
+                  requires ()
+                  {
+                     requires std::convertible_to<scalar_index_t, typename simd_traits<scalar_index_t>::scalar_type>;
+                  })
+    //!\endcond
+    // Note the explicit type conversion is necessary since the scalar type might be of smaller bit size.
+        : row{simd::fill<index_t>(static_cast<typename simd_traits<index_t>::scalar_type>(row_idx.get()))},
+          col{simd::fill<index_t>(static_cast<typename simd_traits<index_t>::scalar_type>(col_idx.get()))}
     {}
 
     /*!\brief Construction from other matrix_index with different integral type.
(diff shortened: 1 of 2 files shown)


include/seqan3/alignment/matrix/detail/matrix_coordinate.hpp
```diff
@@ -87,9 +108,33 @@ struct matrix_index
...
+     * available if `index_t` is a simd vector type.
+     */
+    template <seqan3::arithmetic scalar_index_t>
+    constexpr matrix_index(row_index_type<scalar_index_t> const row_idx,
+                           column_index_type<scalar_index_t> const col_idx) noexcept
+    //!\cond
+        requires simd_concept<index_t>
+    //!\endcond
+        : row{simd::fill<index_t>(static_cast<typename simd_traits<index_t>::scalar_type>(row_idx.get()))},
+          col{simd::fill<index_t>(static_cast<typename simd_traits<index_t>::scalar_type>(col_idx.get()))}
```
marehr, line 137:
```suggestion
        : row{simd::fill<index_t>(typename simd_traits<index_t>::scalar_type{row_idx.get()})},
          col{simd::fill<index_t>(typename simd_traits<index_t>::scalar_type{col_idx.get()})}
```

include/seqan3/alignment/pairwise/detail/type_traits.hpp
```diff
@@ -83,6 +85,12 @@ template <typename configuration_t>
 struct alignment_configuration_traits
 {
 private:
+    //!\brief Transforms integral types to unsigned types and chooses the minimal viable uint for floating point types.
```
marehr, line 88: float?

include/seqan3/alignment/pairwise/detail/type_traits.hpp
```diff
@@ -83,6 +85,12 @@ template <typename configuration_t>
...
+    template <arithmetic score_t>
+    using select_scalar_index_t = lazy_conditional_t<std::integral<score_t>,
+                                                     lazy<std::make_unsigned_t, score_t>,
+                                                     min_viable_uint_t<1ull << (sizeof_bits<score_t> - 1)>>;
```
marehr, line 92: This does not seem like a float. Okay you somehow transform a float into a uint of the same bit size.

marehr: Why? This seems extremely unrelated?

include/seqan3/alignment/matrix/detail/matrix_coordinate.hpp
```diff
@@ -130,12 +175,22 @@ template <std::integral row_index_t, std::integral col_index_t>
...
+/*!\brief A coordinate type to access an element inside of a two-dimensional simd vector matrix.
+ * \ingroup alignment_matrix
+ * \tparam index_t The underlying index type; must model seqan3::simd::simd_concept.
+ */
+template <simd_concept index_t>
+using simd_matrix_coordinate = matrix_index<index_t>;
+
 //!\brief An offset type to move a matrix iterator in two-dimensional space.
 //!\ingroup alignment_matrix
 using matrix_offset = matrix_index<std::ptrdiff_t>;
```
marehr, line 196: This does not need a simd variant?

include/seqan3/alignment/pairwise/detail/type_traits.hpp
rrahn: yes, why not? The algorithm can also work with floating point types.

include/seqan3/alignment/matrix/detail/matrix_coordinate.hpp
rrahn: Not yet

include/seqan3/alignment/pairwise/detail/type_traits.hpp
rrahn: Because I want the index type to be an unsigned type. And if the score type is a floating point I need to know the corresponding uint type that has the same byte size as float or double if I put it into a simd vector.

include/seqan3/alignment/matrix/detail/matrix_coordinate.hpp
rrahn: No this won't work if the scalar type of the simd type is smaller than the type stored in row_index or column_index

include/seqan3/alignment/pairwise/detail/type_traits.hpp
marehr, line 88:
```suggestion
    //!\brief An index type (i.e. unsigned integral) for a score_type which has the same bit size.
```

include/seqan3/alignment/pairwise/detail/type_traits.hpp
marehr: Now that I understand the purpose, I think this should be a property of the `simd_traits`.

We already have

```c++
using int32x_t = seqan3::simd_type_t<int>;
typename seqan3::simd_traits<int32x_t>::mask_type; // same as int32x_t{} == int32x_t{};
typename seqan3::simd_traits<int32x_t>::swizzle_type; // same as seqan3::simd_type_t<uint8_t>;
```

We should add a

```c++
typename simd_traits<int32x_t>::rebind<new_scalar_t> //  same length, same max_length, but with new scalar_t; gives nice compiler error if new scalar t has not same bit size as 
```

and should also add a `index` structure that is compatible to the simd vector.

```c++
typename seqan3::simd_traits<int32x_t>::index_type; // simd_traits<int_simd_t>::rebind<unsigned_of_same_bit_size>
```

We keep your changes here, and add this as an "easy" task for a next sprint

include/seqan3/alignment/matrix/detail/matrix_coordinate.hpp
marehr: I see. You are right. This is a bummer. can you add a comment 

```suggestion
        : row{simd::fill<index_t>(static_cast<typename simd_traits<index_t>::scalar_type>(row_idx.get()))}, // row idx is size_t, but index_t might be of smaller bit size
          col{simd::fill<index_t>(static_cast<typename simd_traits<index_t>::scalar_type>(col_idx.get()))}
```

include/seqan3/alignment/pairwise/detail/type_traits.hpp
rrahn: Obviously, I would have not addded something like this if it wasn't needed. So there is a test already that performs this with floating point types. That broke. This fixes it. Note this is independent of SIMD stuff.

include/seqan3/alignment/pairwise/detail/type_traits.hpp
rrahn: No the index_type has nothing todo with the simd traits type. I would not add this to the simd traits class. I like the idea of the rebind though. For the latter I suggest if needed somewhere else we have an explicit transformation trait for this. But to cite you here: `it does not follow our policy of adding stuff only if it is needed.` That however does not resolve the problem of identifying the right index type based on the input type. So suggestion would be to have an explicit tranformation trait:

```cpp
template <typename t>
struct index_type_for;

template <std::integral t>
struct index_type_for<t> 
{ 
    using type = std::make_unsigned_t<t>; 
} ;

template <std::floating_point t>
struct index_type_for<t> 
{ 
    using type = min_viable_uint_t<1ull << (sizeof_bits<score_t> - 1)>; 
} ;

template <arithmetic t>
using index_type_for_t = typename index_type_for<t>::type;
```

Of course we could also put the two overloads together but maybe it can be better described of what happens here. If we are putting it out anyway I would prefer the more verbose version over the shorter version.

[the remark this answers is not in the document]
rrahn: Why? This is really not making it better inside of the class. Also it would be:

```
    template <arithmetic score_t>
    using select_scalar_index_t = lazy_conditional_t<std::integral<score_t>,
                                                     lazy<unsigned_scalar_index_of_integral, score_t>,
                                                     lazy<unsigned_scalar_index_of_floating_point, score_t>;
```
If you like I can add a more descriptive comment or detail body to the type.

include/seqan3/alignment/matrix/detail/matrix_coordinate.hpp
rrahn: I added the comment above and also added an explicit concept check to require that the scalar_index_t is convertible to the scalar type in the first place.

rrahn:
@marehr polite ping. And could you add a ticket for the simd rebind option and put it into the column planned?

include/seqan3/alignment/pairwise/detail/type_traits.hpp
marehr: > No the index_type has nothing todo with the simd traits type. I would not add this to the simd traits class. I like the idea of the rebind though. For the latter I suggest if needed somewhere else we have an explicit transformation trait for this.

I disagree here, because we already have `mask_type` and `swizzle_type`, which are transformation traits specific to a simd vector. It would make sense to add a way to give you a `index_type`. But let's discuss this later.

include/seqan3/alignment/pairwise/detail/type_traits.hpp
```diff
@@ -83,6 +85,21 @@ template <typename configuration_t>
...
 private:
+    /*!\brief An index type (i.e. unsigned integral) for a score_type which has the same bit size.
+     * \tparam score_t The score type for which the corresponding index type shall be returned; must mode
+     *                 seqan3::arithmetic
+     *
+     * \details
+     *
+     * If the score type models std::integral it will simply be converted to its unsigned counterpart using the
+     * transformation trait std::make_unsigned. For floating point types the bit size is determined and the
+     * corresponding minimal viable unsigned integral type (seqan3::detail::min_viable_uint_t) is returned.
```
marehr, line 96:
```suggestion
     * If the score type models std::integral it will simply be converted to its unsigned counterpart.
     * For floating point types the bit size is determined and the corresponding minimal viable
     * unsigned integral type (seqan3::detail::min_viable_uint_t) is returned.
```

include/seqan3/alignment/pairwise/detail/type_traits.hpp
```diff
@@ -83,6 +85,21 @@ template <typename configuration_t>
...
+     */
+    template <arithmetic score_t>
+    using select_scalar_index_t = lazy_conditional_t<std::integral<score_t>,
+                                                     lazy<std::make_unsigned_t, score_t>,
+                                                     min_viable_uint_t<1ull << (sizeof_bits<score_t> - 1)>>;
```
marehr, line 101:
```suggestion
    using select_scalar_index_t = min_viable_uint_t<1ull << (sizeof_bits<score_t> - 1)>;
```

marehr:
> @marehr polite ping. And could you add a ticket for the simd rebind option and put it into the column planned?

I added https://github.com/seqan/product_backlog/issues/108

include/seqan3/alignment/pairwise/detail/type_traits.hpp
rrahn: But, swizzle type and mask type are entities that are specific to the intrinsics. For some instructions like AVX512 one could wrongly think that the index type is something you provide for the shuffle operation which is then a special non-simd register type. But we want to express inside of one algorithm, that we want the corresponding unsigned type of the score type (and in case of the float/double the corresponding unsigned type). This can be achieved already with the rebind but is not a property of the simd vector itself.

include/seqan3/alignment/matrix/detail/matrix_coordinate.hpp
```diff
@@ -87,9 +108,40 @@ struct matrix_index
...
+     * the simd index.
+     */
+    template <seqan3::arithmetic scalar_index_t>
+    constexpr matrix_index(row_index_type<scalar_index_t> const row_idx,
+                           column_index_type<scalar_index_t> const col_idx) noexcept
+    //!\cond
+        requires (simd_concept<index_t> &&
+                  requires (simd_traits<scalar_index_t> traits)
+                  {
+                     requires std::convertible_to<scalar_index_t, typename decltype(traits)::scalar_type>;
```
marehr, line 139:
```suggestion
                  requires ()
                  {
                     requires std::convertible_to<scalar_index_t, typename simd_traits<scalar_index_t>::scalar_type>;
```

rrahn: But what is the advantage of this?

PR #1821, opened 2020-05-18
[FIX] ppc64le: warning with alphabets [-W=type-limits]

marehr:
```c++
/scratch/local/seiler/nightly-builds/src/include/seqan3/alphabet/concept.hpp:759:39: error: comparison is always true due to limited range of data type [-Werror=type-limits]
  759 |     requires seqan3::alphabet_size<t> >= 0;
      |              ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~
cc1plus: all warnings being treated as errors
```

with `gcc10  -std=c++17 -fconcepts`

### DIFF
--- a/include/seqan3/alphabet/concept.hpp
+++ b/include/seqan3/alphabet/concept.hpp
@@ -756,8 +756,7 @@ SEQAN3_CONCEPT semialphabet =
     std::is_nothrow_copy_constructible_v<t> &&
     requires (t v)
 {
-    requires seqan3::alphabet_size<t> >= 0;
-
+    { seqan3::alphabet_size<t> };
     { seqan3::to_rank(v) };
 };
 //!\endcond


eseiler:
I can confirm that this fixes the errors; however there are 2 more locations (both asserts):

https://github.com/seqan/seqan3/blob/89f6b61d36989d2daa00ef823595930c1af8c54f/include/seqan3/alignment/matrix/detail/two_dimensional_matrix.hpp#L193-L194

https://github.com/seqan/seqan3/blob/89f6b61d36989d2daa00ef823595930c1af8c54f/include/seqan3/alignment/matrix/detail/two_dimensional_matrix.hpp#L205-L206

Here is the build log for building only the alphabet fix:
[log_1.txt](https://github.com/seqan/seqan3/files/4643575/log_1.txt)
`94% tests passed, 166 tests failed out of 2720` (Note that there are 6676 tests)

Here is the build log for building with both assertions removed:
[log_2.txt](https://github.com/seqan/seqan3/files/4643578/log_2.txt)
`99% tests passed, 21 tests failed out of 6250` (Note that there are 6676 tests)

marehr:
@eseiler  As we discussed; I don't have access to the machine, could you overtake the remaining issues with ppc64le?

Thank you!

PR #1823, opened 2020-05-18
[FEATURE] Adds simd optimum tracker policy.

rrahn:
Renames member variables of policy_optimum_tracker.
Uses binary update operation to compare and update the optimum.
Updates the configurator to choose the correct policy.
Updates the pairwise alignment algorithm to reflect the changes.

extracted from #1804

### DIFF
--- a/include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
+++ b/include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
@@ -166,7 +166,7 @@ protected:
     decltype(auto) track_cell(cell_t && cell, matrix_coordinate_type coordinate) noexcept
     {
         if (test_every_cell)
-            compare_and_set_optimum(optimal_score, optimal_coordinate, cell.best_score(), std::move(coordinate));
+            invoke_comparator(cell, std::move(coordinate));
 
         return std::forward<cell_t>(cell);
     }
@@ -189,11 +189,8 @@ protected:
     decltype(auto) track_last_row_cell(cell_t && cell, matrix_coordinate_type coordinate) noexcept
     {
         if (test_last_row_cell && !test_every_cell)
-            compare_and_set_optimum(optimal_score, optimal_coordinate, cell.best_score(), std::move(coordinate));
-
-        return std::forward<cell_t>(cell);
+            invoke_comparator(cell, std::move(coordinate));
     }
-
     /*!\brief Tracks the last cell of a column within the alignment matrix.
      *
      * \tparam cell_t The cell type of the alignment matrix; must have a member function `best_score()`.
@@ -212,7 +209,7 @@ protected:
     decltype(auto) track_last_column_cell(cell_t && cell, matrix_coordinate_type coordinate) noexcept
     {
         if (test_last_column_cell && !test_every_cell)
-            compare_and_set_optimum(optimal_score, optimal_coordinate, cell.best_score(), std::move(coordinate));
+            invoke_comparator(cell, std::move(coordinate));
 
         return std::forward<cell_t>(cell);
     }
@@ -235,16 +232,30 @@ protected:
     decltype(auto) track_final_cell(cell_t && cell, matrix_coordinate_type coordinate) noexcept
     {
         if (!(test_every_cell || test_last_row_cell || test_last_column_cell))
-            compare_and_set_optimum(optimal_score, optimal_coordinate, cell.best_score(), std::move(coordinate));
-
-        return std::forward<cell_t>(cell);
+            invoke_comparator(cell, std::move(coordinate));
     }
-
     //!\brief Resets the optimum such that a new alignment can be computed.
     void reset_optimum() noexcept
     {
         optimal_score = std::numeric_limits<score_type>::lowest();
         optimal_coordinate = {};
+    }
+
+    /*!\brief Handles the invocation of the optimum comparator and updater.
+     * \tparam cell_t The cell type of the alignment matrix; must have a member function `best_score()`.
+     *
+     * \param[in] cell The current cell to be tracked.
+     * \param[in] coordinate The matrix coordinate of the current cell.
+     *
+     * \details
+     *
+     * Forwards the score and coordinate pair as a tuple and invokes the compare and set operation with the
+     * so far best score/coordinate pair and the current score/coordinate pair.
+     */
+    template <typename cell_t>
+    void invoke_comparator(cell_t && cell, matrix_coordinate_type coordinate) noexcept
+    {
+        compare_and_set_optimum(optimal_score, optimal_coordinate, cell.best_score(), std::move(coordinate));
     }
 };
 } // namespace seqan3::detail
--- a/include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
+++ b/include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
@@ -15,35 +15,84 @@
 
 #include <limits>
 
+#include <seqan3/alignment/matrix/detail/coordinate_matrix.hpp>
 #include <seqan3/alignment/matrix/detail/matrix_coordinate.hpp>
 #include <seqan3/alignment/pairwise/detail/type_traits.hpp>
 #include <seqan3/core/algorithm/configuration.hpp>
+#include <seqan3/core/concept/tuple.hpp>
 #include <seqan3/core/type_traits/template_inspection.hpp>
 
 namespace seqan3::detail
 {
+
+/*!\brief A function object that compares and possibly updates the alignment optimum with the current cell.
+ * \ingroup pairwise_alignment
+ *
+ * \details
+ *
+ * Updates the current alignment optimum with the new score and the respective coordinate if the new score
+ * compares greater or equal to the score of the current optimum.
+ */
+struct max_score_updater
+{
+    /*!\brief Compares and updates the optimal score-coordinate pair.
+     * \tparam score_t The type of the score to track; must model std::totally_ordered and
+     *                 std::assignable_from `const & score_t`.
+     * \tparam coordinate_t The type of the coordinate to track; must model std::assignable_from `const & coordinate_t`.
+     *
+     * \param[in,out] optimal_score The optimal score to update.
+     * \param[in,out] optimal_coordinate The optimal coordinate to update.
+     * \param[in] current_score The score of the current cell.
+     * \param[in] current_coordinate The coordinate of the current cell.
+     *
+     * \details
+     *
+     * Compares the current_score with the optimal score and updates the optimal score and coordinate if the current
+     * one is the new optimum. Otherwise, keeps the old optimum.
+     */
+    template <typename score_t, typename coordinate_t>
+    //!\cond
+        requires (std::totally_ordered<score_t> && std::assignable_from<score_t &, score_t const &> &&
+                  std::assignable_from<coordinate_t &, coordinate_t const &>)
+    //!\endcond
+    void operator()(score_t & optimal_score,
+                    coordinate_t & optimal_coordinate,
+                    score_t current_score,
+                    coordinate_t current_coordinate) const noexcept
+    {
+        bool const is_better_score = current_score >= optimal_score;
+        optimal_score = (is_better_score) ? std::move(current_score) : optimal_score;
+        optimal_coordinate = (is_better_score) ? std::move(current_coordinate) : optimal_coordinate;
+    }
+};
 
 /*!\brief Implements the tracker to store the global optimum for a particular alignment computation.
  * \ingroup pairwise_alignment
  *
  * \tparam alignment_configuration_t The type of the alignment configuration; must be a type specialisation of
  *                                   seqan3::configuration.
- *
+ * \tparam optimum_updater_t The type of the optimum update operation, which compares and updates the alignment optimum
+ *                           with the current cell; must model std::semiregular.
  * \details
  *
- * Implements the interface to track the alignment optimum. It updates the currently stored optimum if the
- * optimal score of the given cell compares greater or equal to the stored optimum.
+ * Implements the interface to track the alignment optimum. It updates the currently stored optimum using the
+ * optimum update operation. The optimum update operation is stored inside of the class and can have a state.
+ * The optimum updater must be invokable with a reference to the optimal score and coordinate and the
+ * score and coordinate of the current cell.
  * Special methods are offered to track any cell (for example when computing the local alignment), the last cell of a
  * column or a row (for example when using free-end gaps), or the final cell of the entire matrix (for example in the
  * standard global alignment).
- * Depending on the configuration they might or might not evaluate the stored optimum score of that cell to compare
- * it with the currently stored optimum.
  * The optimum needs to be reset in between alignment computations in order to ensure that the correct result is
  * tracked.
  */
-template <typename alignment_configuration_t>
+template <typename alignment_configuration_t, std::semiregular optimum_updater_t>
 //!\cond
-    requires is_type_specialisation_of_v<alignment_configuration_t, configuration>
+    requires is_type_specialisation_of_v<alignment_configuration_t, configuration> &&
+             std::invocable<optimum_updater_t,
+                            typename alignment_configuration_traits<alignment_configuration_t>::score_type &,
+                            typename alignment_configuration_traits<alignment_configuration_t>::matrix_coordinate_type &,
+                            typename alignment_configuration_traits<alignment_configuration_t>::score_type,
+                            typename alignment_configuration_traits<alignment_configuration_t>::matrix_coordinate_type>
 //!\endcond
 class policy_optimum_tracker
 {
@@ -52,38 +101,22 @@ protected:
     using traits_type = alignment_configuration_traits<alignment_configuration_t>;
     //!\brief The configured score type.
     using score_type = typename traits_type::score_type;
+    //!\brief The matrix coordinate type that is used to locate a cell inside of the alignment matrix.
+    using matrix_coordinate_type = typename traits_type::matrix_coordinate_type;
 
     //!\brief The tracked score of the global optimum.
     score_type optimal_score{};
     //!\brief The matrix coordinate of the tracked optimum.
-    matrix_coordinate optimal_coordinate{};
+    matrix_coordinate_type optimal_coordinate{};
+    //!\brief The function object to compare and exchange the optimum.
+    optimum_updater_t compare_and_set_optimum{};
 
-    //!\brief Flag to handle the free end gaps in the last row of the alignment.
-    bool last_row_is_free{};
-    //!\brief Flag to handle the free end gaps in the last column of the alignment.
-    bool last_column_is_free{};
-
-    /*!\brief Find the new maximum for a given cell of the alignment matrix.
-     *
-     * \tparam cell_t The cell type of the alignment matrix; must have a member function `best_score()`.
-     *
-     * \param[in] cell The current cell to be tracked.
-     * \param[in] coordinate The matrix coordinate of the current cell.
-     *
-     * \returns The forwarded cell.
-     *
-     * \details
-     *
-     * A call to this function only tracks the optimal score of the given cell if the configuration of the alignment
-     * algorithm requires it.
-     */
-    template <typename cell_t>
-    void find_new_optimum(cell_t && cell, matrix_coordinate coordinate) noexcept
-    {
-        bool const is_better_score = cell.best_score() >= optimal_score;
-        optimal_score = is_better_score ? cell.best_score() : optimal_score;
-        optimal_coordinate = is_better_score ? std::move(coordinate) : optimal_coordinate;
-    }
+    //!\brief Whether every cell of the alignment matrix shall be tracked.
+    bool test_every_cell{false};
+    //!\brief Whether cells of the last row shall be tracked.
+    bool test_last_row_cell{false};
+    //!\brief Whether cells of the last column shall be tracked.
+    bool test_last_column_cell{false};
 
     /*!\name Constructors, destructor and assignment
      * \{
@@ -108,8 +141,8 @@ protected:
         {
             // front_end_first, back_end_first, front_end_second, back_end_second
             auto align_ends_config = get<align_cfg::aligned_ends>(config).value;
-            last_row_is_free = align_ends_config[1];
-            last_column_is_free = align_ends_config[3];
+            test_last_row_cell = align_ends_config[1];
+            test_last_column_cell = align_ends_config[3];
         }
         reset_optimum();
     }
@@ -130,8 +163,11 @@ protected:
      * algorithm requires it, for example when a local alignment shall be computed.
      */
     template <typename cell_t>
-    decltype(auto) track_cell(cell_t && cell, matrix_coordinate SEQAN3_DOXYGEN_ONLY(coordinate)) noexcept
+    decltype(auto) track_cell(cell_t && cell, matrix_coordinate_type coordinate) noexcept
     {
+        if (test_every_cell)
+            compare_and_set_optimum(optimal_score, optimal_coordinate, cell.best_score(), std::move(coordinate));
+
         return std::forward<cell_t>(cell);
     }
 
@@ -150,10 +186,10 @@ protected:
      * algorithm requires it, for example when a semi-global alignment shall be computed.
      */
     template <typename cell_t>
-    decltype(auto) track_last_row_cell(cell_t && cell, matrix_coordinate coordinate) noexcept
+    decltype(auto) track_last_row_cell(cell_t && cell, matrix_coordinate_type coordinate) noexcept
     {
-        if (last_row_is_free)
-            find_new_optimum(cell, std::move(coordinate));
+        if (test_last_row_cell && !test_every_cell)
+            compare_and_set_optimum(optimal_score, optimal_coordinate, cell.best_score(), std::move(coordinate));
 
         return std::forward<cell_t>(cell);
     }
@@ -173,10 +209,10 @@ protected:
      * algorithm requires it, for example when a semi-global alignment shall be computed.
      */
     template <typename cell_t>
-    decltype(auto) track_last_column_cell(cell_t && cell, matrix_coordinate coordinate) noexcept
+    decltype(auto) track_last_column_cell(cell_t && cell, matrix_coordinate_type coordinate) noexcept
     {
-        if (last_column_is_free)
-            find_new_optimum(cell, std::move(coordinate));
+        if (test_last_column_cell && !test_every_cell)
+            compare_and_set_optimum(optimal_score, optimal_coordinate, cell.best_score(), std::move(coordinate));
 
         return std::forward<cell_t>(cell);
     }
@@ -196,10 +232,10 @@ protected:
      * algorithm requires it, for example when a global alignment shall be computed.
      */
     template <typename cell_t>
-    decltype(auto) track_final_cell(cell_t && cell, matrix_coordinate coordinate) noexcept
+    decltype(auto) track_final_cell(cell_t && cell, matrix_coordinate_type coordinate) noexcept
     {
-        if (!(last_column_is_free || last_row_is_free))
-            find_new_optimum(cell, std::move(coordinate));
+        if (!(test_every_cell || test_last_row_cell || test_last_column_cell))
+            compare_and_set_optimum(optimal_score, optimal_coordinate, cell.best_score(), std::move(coordinate));
 
         return std::forward<cell_t>(cell);
     }
--- a/include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
+++ b/include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
@@ -204,15 +204,6 @@ protected:
         return std::forward<cell_t>(cell);
     }
 
-    /*!\brief Returns the stored optimum.
-     *
-     * \returns The stored optimum.
-     */
-    score_type tracked_optimum() const noexcept
-    {
-        return optimal_score;
-    }
-
     //!\brief Resets the optimum such that a new alignment can be computed.
     void reset_optimum() noexcept
     {
--- a/include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
+++ b/include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
@@ -133,7 +133,8 @@ protected:
      *
      * \details
      *
-     * Resets the optimum on construction.
+     * Reads the state of seqan3::align_cfg::aligned_ends and enables the tracking of the last row or column if
+     * requested. Otherwise, only the last cell will be tracked.
      */
     policy_optimum_tracker(alignment_configuration_t const & config)
     {
@@ -144,7 +145,6 @@ protected:
             test_last_row_cell = align_ends_config[1];
             test_last_column_cell = align_ends_config[3];
         }
-        reset_optimum();
     }
     //!\}
 
--- a/include/seqan3/alignment/pairwise/detail/policy_optimum_tracker_simd.hpp
+++ b/include/seqan3/alignment/pairwise/detail/policy_optimum_tracker_simd.hpp
@@ -1,0 +1,240 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2020, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides seqan3::detail::policy_optimum_tracker_simd.
+ * \author Rene Rahn <rene.rahn AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <limits>
+#include <seqan3/std/ranges>
+
+#include <seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp>
+#include <seqan3/core/simd/simd.hpp>
+#include <seqan3/core/simd/simd_algorithm.hpp>
+#include <seqan3/core/simd/simd_traits.hpp>
+#include <seqan3/core/type_traits/lazy.hpp>
+#include <seqan3/range/views/zip.hpp>
+
+namespace seqan3::detail
+{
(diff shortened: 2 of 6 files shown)


rrahn:
I know, but I didn’t have time yet to do it. I ping you when I am done.

rrahn:
@marehr now it is ready for review ;)

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
```diff
@@ -14,50 +14,112 @@
...
     score_type optimal_score{};
     //!\brief The matrix coordinate of the tracked optimum.
-    matrix_coordinate optimal_coordinate{};
+    matrix_coordinate_type optimal_coordinate{};
+    //!\brief The function object to compare and exchange the optimum.
+    binary_updater_t compare_and_set_optimum{};
+    //!\}
+/*!\name Switch parameters
+     * \{
+     */
```
marehr, line 114:
```suggestion
    /*!\name Switch parameters
     * \{
     */
```

What are "Switch" parameters?

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
```diff
@@ -116,8 +182,12 @@ class policy_optimum_tracker
      * algorithm requires it, for example when a semi-global alignment shall be computed.
      */
     template <typename cell_t>
-    decltype(auto) track_last_row_cell(cell_t && cell, matrix_coordinate SEQAN3_DOXYGEN_ONLY(coordinate)) noexcept
+    decltype(auto) track_last_row_cell(cell_t && cell, matrix_coordinate_type coordinate) noexcept
     {
+        if (test_last_row_cell && !test_every_cell)
+            compare_and_set_optimum(std::forward_as_tuple(optimal_score, optimal_coordinate),
+                                 std::forward_as_tuple(cell.optimal_score(), std::move(coordinate)));
```
marehr, line 189:
```suggestion
        if (test_last_row_cell && !test_every_cell)
            compare_and_set_optimum(std::forward_as_tuple(optimal_score, optimal_coordinate),
                                    std::forward_as_tuple(cell.optimal_score(), std::move(coordinate)));
```

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
```diff
@@ -14,50 +14,112 @@
...
+    //!\}
+/*!\name Switch parameters
+     * \{
+     */
+    //!\brief Whether every cell of the alignment matrix shall be tracked.
+    bool test_every_cell{false};
+    //!\brief Whether cells of the last row shall be tracked.
+    bool test_last_row_cell{false};
+    //!\brief Whether cells of the last column shall be tracked.
+    bool test_last_column_cell{false};
```
marehr, line 120: This is okay? That they are not constexpr?

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
```diff
@@ -96,8 +158,12 @@ class policy_optimum_tracker
      * algorithm requires it, for example when a local alignment shall be computed.
      */
     template <typename cell_t>
-    decltype(auto) track_cell(cell_t && cell, matrix_coordinate SEQAN3_DOXYGEN_ONLY(coordinate)) noexcept
+    decltype(auto) track_cell(cell_t && cell, matrix_coordinate_type coordinate) noexcept
     {
+        if (test_every_cell)
+            compare_and_set_optimum(std::forward_as_tuple(optimal_score, optimal_coordinate),
+                                    std::forward_as_tuple(cell.optimal_score(), std::move(coordinate)));
```
marehr, line 165: This expression is written 3 times, is there no way to dry this?

rrahn:
> It is good that they are in a single PR because this seems to be one logical change, but it would make reviewing a whole easier, if I could go by commit and commit.
 
I will try to put some stuff out, but that is quite connected anyhow. And making several commits of the same file is often hard to rebase correctly.

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
rrahn: Sure, you can always add one more function to it. But is that really a problem here?

rrahn:
One codecoverage issue will be fixed later. The other one is due to non simd support in CI.

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker_simd.hpp
```diff
@@ -0,0 +1,201 @@
...
+     * \param[in] config The alignment configuration (not used in this context).
+     *
+     * \details
+     *
+     * Resets the optimum on construction. Sets track_last_row_cell and track_last_column_cell to true in order
+     * to check if this is a cell with a potential optimum.
+     */
+    policy_optimum_tracker_simd(alignment_configuration_t const & SEQAN3_DOXYGEN_ONLY(config))
+    {
+        base_policy_t::test_last_row_cell = true;
```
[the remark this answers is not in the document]
rrahn: > The constructor of base_policy_t has if constexpr (traits_type::with_free_end_gaps) why does this one not have this?

Because this has nothing to do with free end gaps yet. This is the semantic needed for how we implement global alignments with simd. We always look into the last row and last column and test via the coordinates whether or not this cell should be tracked. Since in the global alignment we only have one optimum whose position is fixed.

> I still think having a simd class deriving from a scalar class is wrong, because simd is more general than scalar

So you are basically saying that OOP means that the derived class has always less features than the base class? I can do everything with simd that I can do with scalar, except that comparison does not result in a bool (it is overloaded) and that I have a subscript operator to access a single element of the simd vector (scalar is extended). So in my OOP world, that would mean I extend the interface of the scalar type and thus simd is more specialised than scalar. Your example would rather be `scalar=point` and `simd=ellipse`.  -> `class ellipse : point`

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker_simd.hpp
```diff
@@ -0,0 +1,201 @@
...
+
+/*!\brief Implements the tracker to store the global optimum for a particular alignment computation.
+ * \ingroup pairwise_alignment
+ * \copydetails seqan3::detail::policy_optimum_tracker
+ */
+template <typename alignment_configuration_t, std::semiregular binary_update_operation_t>
+//!\cond
+    requires is_type_specialisation_of_v<alignment_configuration_t, configuration>
+//!\endcond
+class policy_optimum_tracker_simd :
```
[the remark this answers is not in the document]
rrahn: No, I won't do this. I need to handle various different simd operations based on different alignment modes and I am not putting this all into one base class. I need more members etc than in the scalar type, I need different initialisations and later different overloads for the respective functions and this is a correct abstraction to make the class more specialised for simd. So I guess, that is the correct abstraction to do. I don't understand why you are so eager to remove this subclass?

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker_simd.hpp
```diff
@@ -0,0 +1,201 @@
...
+ *
+ * \tparam simd_t The simd index type; must model seqan3::simd::simd_concept.
+ *
+ * \details
+ *
+ * Updates the current alignment optimum with the new score and the respective coordinate only if the coordinate
+ * matches the stored column and row index for the respective alignment. The respective coordinates need to be
+ * pre-computed.
+ */
+struct max_score_updater_simd_global
```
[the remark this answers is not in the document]
rrahn: It is not about the simd functionality. It is about the structure which is quite different to the scalar version. The entire computation is completely the same. But simd means that we process multiple sequences of possibly different length in a single operation. And that needs some specific structural changes. These changes are reflected by living in a dedicated class, so that any developer can follow what happens in which context.

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
```diff
@@ -15,35 +15,78 @@
...
+                 std::assignable_from<std::tuple_element_t<1, lhs_t>, std::tuple_element_t<1, rhs_t>>)
+    //!\endcond
+    void operator()(lhs_t && optimal_score_coordinate_pair, rhs_t && current_score_coordinate_pair) const
+    {
+        auto && [optimal_score, optimal_coordinate] = optimal_score_coordinate_pair;
+        auto && [current_score, current_coordinate] = current_score_coordinate_pair;
+
+        bool const is_better_score = current_score >= optimal_score;
+        optimal_score = (is_better_score) ? current_score : optimal_score;
+        optimal_coordinate = (is_better_score) ? std::move(current_coordinate) : optimal_coordinate;
```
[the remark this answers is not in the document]
rrahn: > Why are the names different?

No specific reason. I can rename them.

> Why is the method different?

They have a different semantic. 
Can you remember, that I developed a scoring function which propagates the optimal score from the valid submatrix to the last row/column of the encompassing matrix. then I have to look into the correct coordinates to track the score. So in the global simd alignment, I track every cell  of the last row/column and store the score if the precomputed coordinate is the same as the current one. In the scalar mode I have only one optimum and it is sufficient to update the best score, whenever I call the update operation. In scalar mode, this is always the same for all modes: semi-global and local etc. but in simd mode I need to handle these modes differently.

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
```diff
@@ -150,10 +181,11 @@ class policy_optimum_tracker
      * algorithm requires it, for example when a semi-global alignment shall be computed.
      */
     template <typename cell_t>
-    decltype(auto) track_last_row_cell(cell_t && cell, matrix_coordinate coordinate) noexcept
+    decltype(auto) track_last_row_cell(cell_t && cell, matrix_coordinate_type coordinate) noexcept
     {
-        if (last_row_is_free)
-            find_new_optimum(cell, std::move(coordinate));
```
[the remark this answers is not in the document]
rrahn: I will introduce an extra function to forward the correct structure to the comparison function

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
```diff
@@ -108,8 +135,8 @@ class policy_optimum_tracker
         {
             // front_end_first, back_end_first, front_end_second, back_end_second
             auto align_ends_config = get<align_cfg::aligned_ends>(config).value;
-            last_row_is_free = align_ends_config[1];
-            last_column_is_free = align_ends_config[3];
+            test_last_row_cell = align_ends_config[1];
+            test_last_column_cell = align_ends_config[3];
```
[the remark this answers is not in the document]
rrahn: No, it is not needed for the global simd case. So these combinations of local, semi-global etc. will come later. This is only about simd global alignment.

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
rrahn: The reason was, that I basically want to tell that score and coordinate are associated and that the resulting operation is a binary operation which is more natural than passing 4 arguments.

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker_simd.hpp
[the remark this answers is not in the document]
marehr: Okay when make those classes completely independent of each other, I don't like the forced inheritance if they really share so little with each other.

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker_simd.hpp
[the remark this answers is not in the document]
marehr: I think this is exactly the problem, they are both not complete sub-types of each other, so a normal "inheritance" does not fit. We should go by their skills and if you don't use `simd[]` and are ignoring that `simd == simd = mask != bool` (you don't need that property most of the time) and use special CPOs (like)

* `seqan3::numeric::is_mask_v<bool / simd_mask> = true`
* `seqan3::numeric::is_signed_v<int_simd_or_scalar> = true`
* `seqan3::numeric::is_unsigned_v<unsigned_simd_or_scalar> = true`
* `seqan3::numeric::make_signed_t<unsigned_simd_or_scalar>`
* `seqan3::numeric::make_unsigned_t<signed_simd_or_scalar>`
* `seqan3::numeric::assign_scalar<simd_or_scalar>(scalar) -> simd_or_scalar`,
* `seqan3::numeric::masked_assign(simd_or_scalar_mask, simd_or_scalar) -> simd_or_scalar` 
* `seqan3::numeric::ternary(simd_or_scalar_mask, simd_or_scalar, simd_or_scalar) -> simd_or_scalar` 
* `seqan3::numeric::zero<simd_or_scalar[_mask]>() -> simd_or_scalar[_mask]`
* `seqan3::numeric::one<simd_or_scalar[_mask]>() -> simd_or_scalar[_mask]`
* `seqan3::numeric::iota<simd_or_scalar>() -> simd_or_scalar`
* `seqan3::numeric::min(simd_or_scalar, simd_or_scalar) -> simd_or_scalar_mask`
* `seqan3::numeric::max(simd_or_scalar, simd_or_scalar) -> simd_or_scalar_mask`
* `seqan3::numeric::all_of(simd_or_scalar_mask) -> bool`
* `seqan3::numeric::any_of(simd_or_scalar_mask) -> bool`

I think we can express most kernels exactly the same. Of course do you need more pre- / post-processing code with simd, but I think these can live in the same class with some constexpr if's or `requires seqan3::simd::simd_concept<>` overloads.

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
```diff
@@ -14,50 +14,112 @@
...
- * Depending on the configuration they might or might not evaluate the stored optimum score of that cell to compare
- * it with the currently stored optimum.
  * The optimum needs to be reset in between alignment computations in order to ensure that the correct result is
  * tracked.
  */
-template <typename alignment_configuration_t>
+template <typename alignment_configuration_t, std::semiregular binary_updater_t>
 //!\cond
-    requires is_type_specialisation_of_v<alignment_configuration_t, configuration>
+requires is_type_specialisation_of_v<alignment_configuration_t, configuration>
```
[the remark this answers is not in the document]
marehr: Still open

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
marehr: I kind of understand the "more" natural thing, because you logically group them.

But I think it obfuscates which variables are lvalue and which ones are not and makes it in general harder to understand the semantics of the function. 

Since we bundled it into one call (`invoke_comparator`), it is not really harder to understand the semantics any more. On the other hand this also takes the point of needing it to be more natural, since we only have a single invocation, and we don't need a "natural" feeling for a bypassing piece of code.

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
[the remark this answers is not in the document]
marehr: Maybe call this class `policy_optimum_tracker_global`?

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker_simd.hpp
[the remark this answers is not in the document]
rrahn: I really don‘t see your argument here. When putting shared functionality in a base class I have one class that adds nothing to it. So it would boil down to an alias nothing more. If I merge the classes, understanding the code and cases becomes ridiculously complicated. In a separate class I have a separate and clear documentation of what happens. And why should policies not inherit from each other? That doesn‘t make sense. Especially, since the one is a specialisation of the other.

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker_simd.hpp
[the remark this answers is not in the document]
rrahn: Well the problem is, that simd is a difficult type. It merges the concepts of arithmetic types and array-like types. The only safe abstraction would be as you suggest to handle scalar types as a special alias for simd[0]. You could have a dedicated simd_scalar_wrapper that fulfills the requirements for simd, write one kernel and cross the fingers that the compiler can optimize this away. We even talked about it sometimes in the past. But we don‘t have it. And yes you can spent another two months to make it „right“. But I don‘t have the time to do so and so has no one else. It is not planed and it does not generate sufficient value to make it a priority. If we have the feature at some point in the future, then I’ll gladly spent one afternoon to adapt the code. Right now, it is good as it is.

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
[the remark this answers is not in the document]
marehr: Still open

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker_simd.hpp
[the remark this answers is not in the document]
marehr: Look at the constructor and if you can't use the inherited constructor to initialise the base class that means you can't use the base class. This violates OOP for me, because the base class is left in a (possible) invalid state.

That's why you should use either a CRTP class, because you can require that your scalar derived class defines `reset_optimum` which can be used within the CRTP classes' constructor (would be my compromise). Or you merge them, in that case `reset_optimum` has the right meaning, because of a `constexpr if` (my preferred way). Or you make them complete separate (my least preferred), because how much code does the simd case reuse? Not much!

I don't buy your argument that it would make the code more difficult to understand, because you can always have well-defined functions that enable/disable depending on `requires is_vectorised` and this would minimise the number of constexpr. This makes it even easier to compare different methods, because they are in the same class and are local to each other. Having two classes, assumes that you need to know that a second class exists and this is so seqan2 style of scattering all closely coupled things over multiple locations. This is insane.

I had a really difficult time to review this, because I couldn't verify easily that you do the same or something similar in the scalar case and I think this sucks that we have such a disconnect here.

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker_simd.hpp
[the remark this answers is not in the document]
marehr: I never said that this should block anything, I try to find a compromise here and I get the PR back to review just as it is, without any constructive message. The message that I received is that you don't want to change anything at all, because it is inconvenient or not "planned", and I don't know how to feel about this.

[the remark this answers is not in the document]
rrahn: No, it's not that I won't change it because I don't want to or because it is inconvenient. I don't change it because I haven't seen the argument yet, that choosing another way to express the same thing is beneficial.

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker_simd.hpp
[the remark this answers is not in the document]
rrahn: > Look at the constructor and if you can't use the inherited constructor to initialise the base class that means you can't use the base class. This violates OOP for me, because the base class is left in a (possible) invalid state. 

Done. I call the base constructor. 😄 

> I don't buy your argument that it would make the code more difficult to understand, because you can always have well-defined functions that enable/disable depending on requires is_vectorised and this would minimise the number of constexpr. This makes it even easier to compare different methods, because they are in the same class and are local to each other. Having two classes, assumes that you need to know that a second class exists and this is so seqan2 style of scattering all closely coupled things over multiple locations. This is insane.

Sorry, but this opinion changes regularly within the team. I have been through enough examples where it doesn't matter what you code. It is always the other way that is more correct and usually due to personal preferences. I'm not taking me out of this. Now again, you want me to choose between three alternatives, which all have their pros and cons. None of them covers all the things that might come. So any alternative is likely to change in the future when we know more about the dependencies.
I am not saying that this is the final best cool design whatsoever. I am saying, there is quite a lot of things that have to be considered, added, modified etc. until all the functionality with simd and scalar and so on works with the wanted performance. If all these components are known and coded, then we can sit down and - hell yeah - optimise the last bit of code out of it. I'd love that. And I value your expertise in this. And you can point your finger at me if it ends up to be a solution of yours. But please let us first get a complete picture. My focus is on the performance now, so I can make the benchmarks etc.

rrahn:
@marehr I haven't rebased yet, but are waiting for your go to rebase. Then I will also look into the errors.

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker_simd.hpp
```diff
@@ -32,9 +32,16 @@ namespace seqan3::detail
...
+ * This operation is specific to the global alignment in vectorised mode. The tracking of the last cells of the
+ * different alignment matrices that are computed simultaneously in one vector unit depends on how the scoring of the
+ * global alignment works. Any alignment matrix that is smaller than the largest matrix defined by the longest sequence
+ * in the collection of the first sequences and in the collection of the second sequences, will have its last cell
+ * in the middle of the encompassing matrix. In order to track this cell without checking every cell of the alignment
+ * matrix for the correct coordinate, the last cell of every contained matrix is projected along its diagonal to either
+ * the last row or the last column of the encompassing matrix. Within the algorithm only the cells of the last row,
+ * respectively column are tracked. Since the cells to track are fixed, the respective coordinates for every
+ * contained matrix can be precomputed. Subsequently, within this update operation only score is updated if the
+ * coordinate of the current cell compares equal to the precomputed coordinate of any of the contained matrices.
```
marehr, line 44:
```suggestion
 * contained matrix can be precomputed. Subsequently, within this update operation the score is only updated if the
 * coordinate of the current cell compares equal to the precomputed coordinate of any of the contained matrices.
```

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker_simd.hpp
```diff
@@ -149,6 +156,37 @@ class policy_optimum_tracker_simd :
...
      * Finally, the added offset is removed from the score to obtain the true value.
+     *
+     * ### Example
+     *
+     * Consider the following collections:
+     * collection 1: ["aaa", "aa",   "a"]
+     * collection 2: [  "a", "aa", "aaa"]
+     *
+     * Based on the length of the sequences the encompassing alignment matrix has the dimensions 4x4.
+     * The following graphic depicts this matrix. A `x` marks the last cell for the contained matrix given the
```
marehr, line 167: There are no `x`'s right?

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker_simd.hpp
```diff
@@ -149,6 +156,37 @@ class policy_optimum_tracker_simd :
...
+     *
+     * As can be seen, the end of each matrix must not necessarily be the end of the encompassing matrix. To avoid
+     * tracking every cell in the matrix the end point candidates will be projected along the diagonal to the
+     * last row or column of the encompassing matrix. In the example above, 1 and 3 are projected with an offset of `0`,
+     * while 2 is projected to the coordinate (3,3) with an projection offset of `1`. During the computation of the
+     * alignment the simd scoring scheme ensures that outside of the original matrix a fixed cost is added to every
+     * cell. The original score is shifted by this cost multiplied with the computed projection offset.
+     * In the global alignment it is suffcient to only track the optimal score in the last row and column of the
+     * encompassing matrix and only at the precomputed coordinate projections. Eventually, the score offset is
+     * subtracted to obtain the original score.
```
marehr, line 189: Thank you so much, this is a great explanation! :)

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker_simd.hpp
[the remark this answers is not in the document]
marehr: For the record. We talked this in person, and we both agreed that both approaches would have worked. We decided that we go with the static-runtime polymorphism for now, because it would be too much trouble to change it now. But we both agreed that we might change this in the future, once we have all the use cases implemented.

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
marehr: This is still open, I personally would prefer the explicit argument list, but let's sneakingly wait if the 2nd reviewer writes "WTH?!"

rrahn: ah ok. I didn't see this. No I can change this. No hard feelings to call it with four parameters.

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker_simd.hpp
rrahn: it is supposed to be a wild card for the number.

rrahn: Maybe I call it the number, instead of A `x`

rrahn:
> I have to double check my other open issues, but I think this PR is ready for second review.

Yes, I didn't know either what was still open 😆

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker_simd.hpp
rrahn, line 167:
```suggestion
     * The following graphic depicts this matrix. The number `i` marks the last cell for the contained matrix given the
```

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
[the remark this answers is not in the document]
smehringer: still open? Or maybe rebased somewhere else..

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
```diff
@@ -15,35 +15,78 @@
...
+
 /*!\brief Implements the tracker to store the global optimum for a particular alignment computation.
  * \ingroup pairwise_alignment
  *
  * \tparam alignment_configuration_t The type of the alignment configuration; must be a type specialisation of
  *                                   seqan3::configuration.
- *
+ * \tparam binary_updater_t The type of a binary update operation to update the alignment optimum; must model
+ *                          std::semiregular and must be invocable with two tuples each containing the score and
+ *                          the coordinate.
```
smehringer, line 76: Isn't your `max_score_updater` one example for this type? If so `max_score_updater` does not get two tuples?

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
```diff
@@ -86,7 +86,7 @@ struct max_score_updater
  */
 template <typename alignment_configuration_t, std::semiregular binary_updater_t>
 //!\cond
-requires is_type_specialisation_of_v<alignment_configuration_t, configuration>
+    requires is_type_specialisation_of_v<alignment_configuration_t, configuration>
```
smehringer, line 89: Well you can just squash this commit ([NOP] Fix indentation in policy optimum tracker. ) into the former ([MISC] Introduce binary matrix updater..)

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
```diff
@@ -229,16 +226,30 @@ class policy_optimum_tracker
...
-
     //!\brief Resets the optimum such that a new alignment can be computed.
     void reset_optimum() noexcept
     {
         optimal_score = std::numeric_limits<score_type>::lowest();
         optimal_coordinate = {};
     }
+
+    /*!\brief Handles the invocation of the optimum comparator and updater.
+     * \tparam cell_t The cell type of the alignment matrix; must have a member function `optimal_score()`.
```
smehringer, line 238:
```suggestion
     * \tparam cell_t The cell type of the alignment matrix; must have a member function `best_score()`.
```

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker_simd.hpp
```diff
@@ -0,0 +1,235 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2020, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides seqan3::detail::policy_optimum_tracker.
```
smehringer, line 9:
```suggestion
 * \brief Provides seqan3::detail::policy_optimum_tracker_simd.
```

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker_simd.hpp
```diff
@@ -0,0 +1,235 @@
...
+struct max_score_updater_simd_global
+{
+
+    /*!\brief Compares and updates the optimal score-coordinate pair.
+     * \tparam score_t The type of the score to track; must model std::assignable_from `const & score_t`.
+     * \tparam coordinate_t The type of the coordinate to track; must be a seqan3::matrix_index type with members that
+     *                      model seqan3::simd::simd_concept.
+     *
+     * \param[in,out] optimal_score The optimal score to update.
+     * \param[in,out] optimal_coordinate The optimal to update.
```
smehringer, line 53:
```suggestion
     * \param[in,out] optimal_coordinate The optimal coordinate to update.
```

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker_simd.hpp
```diff
@@ -0,0 +1,235 @@
...
+    template <typename score_t, typename coordinate_t>
+    //!\cond
+        requires (std::assignable_from<score_t &, score_t const &> &&
+                  requires (coordinate_t coordinate)
+                  {
+                      requires simd_concept<decltype(coordinate.col)>;
+                      requires simd_concept<decltype(coordinate.row)>;
+                  })
+    //!\endcond
+    void operator()(score_t & optimal_score,
```
smehringer, line 72: side question, but why is this function not static? This way you would need to store an instance. The updater doesn't seem to have a state

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker_simd.hpp
```diff
@@ -0,0 +1,235 @@
...
+    policy_optimum_tracker_simd & operator=(policy_optimum_tracker_simd const &) = default; //!< Defaulted.
+    policy_optimum_tracker_simd & operator=(policy_optimum_tracker_simd &&) = default; //!< Defaulted.
+    ~policy_optimum_tracker_simd() = default; //!< Defaulted.
+
+    /*!\brief Construction and initialisation using the alignment configuration.
+     * \param[in] config The alignment configuration (not used in this context).
+     *
+     * \details
+     *
+     * Resets the optimum on construction. Sets track_last_row_cell and track_last_column_cell to true in order
```
smehringer, line 131:
```suggestion
     * Resets the optimum at construction. Sets track_last_row_cell and track_last_column_cell to `true` in order
```
?

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
rrahn: Why? It has nothing to do with the former commit.

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
rrahn: Ah I need to update this documentation. Thanks for pointing this out.

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
```diff
@@ -15,33 +15,76 @@
...
+     */
+    template <typename score_t, typename coordinate_t>
+    //!\cond
+        requires (std::totally_ordered<score_t> && std::assignable_from<score_t &, score_t const &> &&
+                  std::assignable_from<coordinate_t &, coordinate_t const &>)
+    //!\endcond
+    void operator()(score_t & optimal_score,
+                    coordinate_t & optimal_coordinate,
+                    score_t const & current_score,
+                    coordinate_t const & current_coordinate) const noexcept
```
rrahn, line 61:
```suggestion
                    coordinate_t current_coordinate) const noexcept
```

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker_simd.hpp
rrahn: That is the regular implementation of a function object. You don't need to have something static here. Without a state this is even an empty class that occupies at most one byte. I think that is not the problem. I could rewrite it as a named lambda object, but then the requires clause becomes unreadable because of `remove_cvref_t<decltype(auto)>` for every variable I refer to. Hence it is a function object. Second, in the semi-global alignment I need to restrict the area for which the alignment score can be tracked based on the sizes of the contained matrices. This will be done, by storing the outermost column and row coordinate for each submatrix.

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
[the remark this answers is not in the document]
rrahn: shouldn't be open anymore

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
```diff
@@ -15,35 +15,78 @@
...
+    {
+        bool const is_better_score = current_score >= optimal_score;
+        optimal_score = (is_better_score) ? current_score : optimal_score;
+        optimal_coordinate = (is_better_score) ? std::move(current_coordinate) : optimal_coordinate;
```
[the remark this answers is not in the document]
smehringer: heah?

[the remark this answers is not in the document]
rrahn: The code was different before by passing in tuples etc. Now the interface changed but I did not update the function body. But now I did. :)

smehringer:
you probably want to rebase

PR #1826, opened 2020-05-19
[FIX] enable bidirectional tests for random_access iterators.

smehringer:
Before we always enabled a test if the iterator tag was "equal to" the bidirectional tag, but the test needs to be enabled also if it is the random access tag. Instead of testing for equality, we can use `std::derived_from`.

### DIFF
--- a/test/unit/range/iterator_test_template.hpp
+++ b/test/unit/range/iterator_test_template.hpp
@@ -42,10 +42,10 @@ struct iterator_fixture : public ::testing::Test
 };
 
 // Helper concept to check whether the test fixture has a member function expect_eq.
-template <typename T>
-SEQAN3_CONCEPT HasExpectEqualMemberFunction = requires(T & a)
+template <typename t>
+SEQAN3_CONCEPT has_expect_equal_member_function = requires(t & a)
 {
-    requires std::same_as<decltype(T::expect_eq(*std::ranges::begin(a.test_range),
+    requires std::same_as<decltype(t::expect_eq(*std::ranges::begin(a.test_range),
                                                 *std::ranges::begin(a.expected_range))), void>;
 };
 
@@ -53,7 +53,7 @@ SEQAN3_CONCEPT HasExpectEqualMemberFunction = requires(T & a)
 template <typename T, typename A, typename B>
 void expext_eq(A && a, B && b)
 {
-    if constexpr (HasExpectEqualMemberFunction<iterator_fixture<T>>)
+    if constexpr (has_expect_equal_member_function<iterator_fixture<T>>)
         iterator_fixture<std::remove_reference_t<T>>::expect_eq(a, b);
     else
         EXPECT_EQ(a, b);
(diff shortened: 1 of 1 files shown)


test/unit/range/iterator_test_template.hpp
```diff
@@ -65,58 +76,61 @@ TYPED_TEST_P(iterator_fixture, concept_check)
...
+
+    if constexpr (at_least_forward_iterator_tag<typename TestFixture::iterator_tag>)
     {
-        static_assert(std::forward_iterator<decltype(std::ranges::begin(this->expected_range))>,
+        static_assert(std::ranges::forward_iterator<decltype(std::ranges::begin(this->expected_range))>,
                       "expected_range must have a begin member function and "
-                      "the returned iterator must model std::forward_iterator.");
-        EXPECT_TRUE(std::forward_iterator<iterator_type>);
+                      "the returned iterator must model std::ranges::forward_iterator.");
+        EXPECT_TRUE(std::ranges::forward_iterator<iterator_type>);
```
marehr, line 98:
```suggestion
        EXPECT_TRUE(std::forward_iterator<iterator_type>);
```

This is OUR fuck up that `std::ranges` also contains `std::forward_iterator`,  `std::default_sentinel` and so on.

The rule is: Iterator related things are in `std::`, ranges related things are in `std::ranges`

To remove those things from `std::ranges` is part of https://github.com/seqan/product_backlog/issues/49

test/unit/range/iterator_test_template.hpp
```diff
@@ -42,18 +42,29 @@ struct iterator_fixture : public ::testing::Test
...
-    requires std::same_as<decltype(T::expect_eq(*std::ranges::begin(a.test_range),
+    requires std::same_as<decltype(t::expect_eq(*std::ranges::begin(a.test_range),
                                                 *std::ranges::begin(a.expected_range))), void>;
 };
 
+// Helper concept to check whether the iterator tag is at least a forwar_iterator.
+template <typename t>
+SEQAN3_CONCEPT at_least_forward_iterator_tag = std::same_as<t, std::forward_iterator_tag> ||
+                                               std::same_as<t, std::bidirectional_iterator_tag> ||
+                                               std::same_as<t, std::random_access_iterator_tag>;
```
marehr, line 56: They inherit from each other; so a simple `derived_from` would suffice.

See https://en.cppreference.com/w/cpp/iterator/iterator_tags

smehringer: thanks

test/unit/range/iterator_test_template.hpp
```diff
@@ -116,10 +117,11 @@ TYPED_TEST_P(iterator_fixture, concept_check)
...
         }
     }
-    else
+
+    if (!std::derived_from<typename TestFixture::iterator_tag, std::input_iterator_tag>)
     {
-        FAIL() << "The iterator tag member type must be one of std::forward_iterator_tag,"
-               << "std::bidirectional_iterator_tag, std::random_access_iterator_tag.";
+        FAIL() << "The iterator tag member type must be one of std::input_iterator_tag,"
+               << "std::forward_iterator_tag, std::bidirectional_iterator_tag, or std::random_access_iterator_tag.";
```
marehr, line 124: or [contiguous_iterator_tag](https://en.cppreference.com/w/cpp/iterator/iterator_tags) xD

smehringer:
Here it is https://github.com/seqan/product_backlog/issues/94

PR #1827, opened 2020-05-19
[FIX] ppc64le: remove tautology

eseiler:
Fixes

```console
/dev/shm/tmp/seqan3/include/seqan3/alignment/matrix/detail/two_dimensional_matrix.hpp:193:31: error: comparison of unsigned expression in ‘>= 0’ is always true [-Werror=type-limits]
  193 |         assert(coordinate.col >= 0u);
```

### DIFF
--- a/include/seqan3/alignment/matrix/detail/two_dimensional_matrix.hpp
+++ b/include/seqan3/alignment/matrix/detail/two_dimensional_matrix.hpp
@@ -200,8 +200,8 @@ public:
     //!\copydoc operator[]
     constexpr const_reference operator[](matrix_coordinate const & coordinate) const noexcept
     {
-        assert(static_cast<size_t>(coordinate.col) < cols());
-        assert(static_cast<size_t>(coordinate.row) < rows());
+        assert(coordinate.col < cols());
+        assert(coordinate.row < rows());
 
         return *(begin() + matrix_offset{row_index_type{static_cast<std::ptrdiff_t>(coordinate.row)},
                                          column_index_type{static_cast<std::ptrdiff_t>(coordinate.col)}});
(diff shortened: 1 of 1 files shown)


eseiler:
travis finished but the status does not update

include/seqan3/alignment/matrix/detail/two_dimensional_matrix.hpp
```diff
@@ -190,8 +190,6 @@ class two_dimensional_matrix
      */
     constexpr reference operator[](matrix_coordinate const & coordinate) noexcept
     {
-        assert(coordinate.col >= 0u);
-        assert(coordinate.row >= 0u);
```
smehringer, line 194: Hm but are we sure that this is always an unsigned type?

eseiler: https://github.com/seqan/seqan3/blob/f54cb5e6048eb2c205e0dc913fb44166c94357cf/include/seqan3/alignment/matrix/detail/matrix_coordinate.hpp#L135-L137

yes

smehringer: @rrahn looks fine from my side but maybe you take a second look because I don't know the reason why `matrix_index<size_t>` is templatized and it might be changed/used differently?

eseiler: I asked him when I removed it, and he said it's always unsigned and defined as size_t

smehringer: alright

smehringer:
merging because travis is green but not updating

PR #1828, opened 2020-05-19
 [FIX] explicitly import cxx20 things into std namespace

marehr:
With this PR the std headers have the same entities defined for `-std=c++17` as well as `-std=c++2a`.

We once decided that we will "import" things from the ranges library on demand, this PR comments in ranges stuff that we actually use.

I prefixed all names with `::ranges::cpp20::`, but be aware that this does not mean that each of those names really exist in the ranges library.

This is PART of seqan/product_backlog#49

### DIFF
--- a/include/seqan3/range/views/minimiser.hpp
+++ b/include/seqan3/range/views/minimiser.hpp
@@ -190,7 +190,7 @@ public:
     /*!\brief Tag this class as a forward iterator if `urng_t` models std::ranges::forward range, otherwise as an
      *        input iterator.
      */
-    using iterator_category = std::conditional_t<std::ranges::forward_iterator<it_t>,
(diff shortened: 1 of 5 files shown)


wvandertoorn:
There is a minimiser commit in here, was that intentional?

marehr:
> There is a minimiser commit in here, was that intentional?

Yes, the only (wrong) usage of `std::ranges::forward_iterator` was in the minimiser view, so I fixed it in here. The correct place would be `std::forward_iterator`. I decided to use `std::ranges::forward_range`, because I think in a view context it makes more sense to directly operate on the underlying range.

marehr:
> One more thing. the borrowed range is still dependent on the ranges library, right?

Yes, we use `ranges::enable_safe_range` and `ranges::enable_view` directly, because we need to add them in the namespace they are defined.

The rename of those need to be done in the ranges library.

PR #1829, opened 2020-05-19
[INFRA] Update travis.yml

eseiler:
Some stuff we are doing in the `travis.yml` has no effect anymore or is deprecated.

See https://github.com/seqan/app-template/blob/master/.travis.yml for an up to date `travis.yml`

Since we build everything except coverage in Release mode, I didn't put Release/Debug in the name. 
If you click a job on the overview page, you will still see the flags used for this job.

If we are not running the snippet tests, we can also run `ctest` with 2 threads. Since the snippets create and delete files and some separate tests create/delete the same files, they are not safe to run in parallel. Anyway, the snippets only take a few seconds to run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
-sudo: required
+version: ~> 1.0
+os: linux
 dist: xenial
 language: cpp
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -9,109 +10,90 @@ cache:
...
+        - CC=gcc-7
+        - BUILD=performance
+        - BUILD_TYPE=Release
+    - name: "Header gcc7"
+      env:
+        - CXX=g++-7
+        - CC=gcc-7
+        - BUILD=header
+        - BUILD_TYPE=Release
+        - CXXFLAGS="-Wno-deprecated-declarations"
```
marehr, line 63: Can you try to remove this flag? We shouldn't suppress some warnings globally only because we had some issues at one point in time.

eseiler: Works 👍

marehr: Nice :) I feared that we hid some errors

PR #1831, opened 2020-05-20
WIP on dynamic search strategy

smehringer:


include/seqan3/search/search.hpp
```diff
@@ -297,6 +298,22 @@ inline auto search(queries_t && queries,
...
+//!\endcond
+inline auto search(queries_t && queries, index_t const & index, configuration_t const & cfg)
+{
+    auto hit_strategy_variant = get<search_cfg::hit>(cfg).selection;
+    auto cfg_wihtout_strategy = cfg.template remove<search_cfg::hit>();
+
+    return std::visit([&] (auto strategy)
+                      {
+                          return search(std::forward<queries_t>(queries), index, cfg_wihtout_strategy | strategy);
+                      }, hit_strategy_variant);
```
marehr, line 314: The problem is here right?

rrahn: In general what you want to do is something like this:
```cpp
auto [search_algo, adapted_config] = search_configurator::configure<quries_t>(index, cfg);
return search_range{std::move(search_algo), queries};
```

And in the configuration you need to call the visit and all things necessary to configure the algorithm. That must be returned as a std::function in order to allow type erasue on the configured algorithm. Only the input and the return type must be fix at compile time in order to define the std::function object.

smehringer: yes. You can also check the error message. I think it says somehting a long the lines of  "search_result_range<   .... search_cfg::hit_all  ... >" is not convertible to "search_result_range<   .... search_cfg::hit  ... >".

smehringer: > In general what you want to do is something like this:
...

Just read this now. I don't know how. Can you either explain this to me, or take on the issue yourself?

rrahn:
Superseded by #1853 
I close this as it seems to be done? Please reopen if there was anything else you needed this PR for.

PR #1833, opened 2020-05-20
[TEST] rename alphabet_ test suite in test template

marehr:
@simonsasse asked me why we have a underscore in

```c++
TEST(range_and_iterator, innermost_value_type_) // ...
```

I wanted to give him the rational that we have a name clash, but then it
came to me that we actually don't have any name clashes any more due to
https://github.com/seqan/product_backlog/issues/89.

This means we can remove the underscore or `global_` constructs from our
test cases which avoided ambiguous names.

This PR should be used as an example how to rename other test suites.

Part of https://github.com/seqan/product_backlog/issues/90


marehr:
@svnbgnk This is ready to review :)

PR #1835, opened 2020-05-20
[INFRA] don't override CMAKE_CXX_COMPILER_LAUNCHER

marehr:
We override CMAKE_CXX_COMPILER_LAUNCHER in our tests. This is bad if you want to use different `COMPILER_LAUNCHER` such as [distcc](https://github.com/distcc/distcc).

### DIFF
--- a/test/seqan3-test.cmake
+++ b/test/seqan3-test.cmake
@@ -158,7 +158,12 @@ macro (seqan3_require_ccache)
         find_package_message (CCACHE_PROGRAM "Finding program ccache - Success" "[${CCACHE_PROGRAM}]")
         # New option since cmake >= 3.4:
         # https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_LAUNCHER.html
-        set (CMAKE_CXX_COMPILER_LAUNCHER "${CCACHE_PROGRAM}")
+        if (NOT CMAKE_VERSION VERSION_LESS 3.15) # cmake >= 3.15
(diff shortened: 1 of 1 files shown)


test/seqan3-test.cmake
```diff
@@ -158,7 +158,16 @@ macro (seqan3_require_ccache)
         find_package_message (CCACHE_PROGRAM "Finding program ccache - Success" "[${CCACHE_PROGRAM}]")
         # New option since cmake >= 3.4:
         # https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_LAUNCHER.html
-        set (CMAKE_CXX_COMPILER_LAUNCHER "${CCACHE_PROGRAM}")
+        if (NOT CMAKE_VERSION VERSION_LESS 3.15) # cmake >= 3.15
+            list (PREPEND CMAKE_CXX_COMPILER_LAUNCHER "${CCACHE_PROGRAM}")
```
smehringer, line 162: Does this automatically not prepend a non-set variable? I could find a note about this in the [documentation](https://cmake.org/cmake/help/v3.15/command/list.html)

marehr: If you look up, you will see that we guarantee that `CCACHE_PROGRAM` has a value / is set.

But yes, you are in general right, if the variable would be empty `PREPEND("a;b;c;d;e;f", "") = ";a;b;c;d;e;f"`.

But thanks to you, I saw the `list(INSERT <list> <index> [<element>...])` syntax; I overlooked it before, I will change the pre `cmake 3.15` version.

Issue #1595, opened 2020-02-17
Current LLVM copy of to_chars is wrong for integers with more than 9 digits


smehringer:
#### Platform

  - SeqAn version: tested with 3.0.1
  - Operating system: Linux
  - Compiler: tested with GCC-7/GCC-9

#### Description

Converting a  number with 9 digits or more fails with the current implementation in the `std/charconv` header. If I include the stl `<charconv>` with GCC-9 directly the error does not occur.

There is also the possibility that we introduced the error while copying..
We should update the code and may want to consider copying the MSVC version that is supposed to be fast on floats.

### How to repeat the problem

The following test fails with the current implementation

```cpp
TEST(to_chars_test, large_number)
{
    uint64_t val{123456789};
    std::array<char, 100> buffer{};

    auto res = std::to_chars(buffer.data(), buffer.data() + buffer.size(), val);

    EXPECT_EQ(res.ptr, &buffer[9]);
    EXPECT_EQ(res.ec, std::errc{});
    EXPECT_EQ((std::string_view{buffer.data(), 9}), std::string_view{"123456789"});
}
```

marehr:
Core-Meeting:

@marehr has free reign to do this; main idea is to put the llvm implementation into `contrib` and add the custom bits, like the `float` overloads, in `std/charconv`.

marehr:
2020-06-30 Core Meeting

Discussed license change in the LLVM project from 

> This file is dual licensed under the MIT and the University of Illinois Open Source Licenses. 
> https://github.com/llvm/llvm-project/blob/d27489645bb8829a952f79e3b211202844238391/libcxx/src/charconv.cpp

to

> SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
> https://github.com/llvm/llvm-project/blob/master/libcxx/src/charconv.cpp

PR #1837, opened 2020-05-22
put charconv into contrib

marehr:
This PR 
* does the first task of putting the llvm version of charconv into the `contrib` folder. 
See https://github.com/seqan/seqan3/issues/1595#issuecomment-626630112
* splits the charconv test into two tests; one for integral types which would test the llvm implementation and one for floating types which would test our own implementation

I didn't upgrade the llvm version yet, because they had a license change to `apache2 WITH llvm-exception`

See
* https://github.com/llvm/llvm-project/blob/d27489645bb8829a952f79e3b211202844238391/libcxx/include/charconv
* https://github.com/llvm/llvm-project/blob/d27489645bb8829a952f79e3b211202844238391/libcxx/src/charconv.cpp

My second reason was to test the upgrade-path if we ever need to add new version of this.
And my third reason: Like always, a small change can lead to a couple of unintended CI breakage; I wanted to see if everything works with the new structure which uses the old code.

-----

# PLEASE DON'T SQUASH-MERGE THIS PR


marehr:
(Documentation is currently failing, but you don't see it in CI, because of #1838)

marehr:
@simonsasse This can be reviewed now, documentation error was fixed

marehr:
> Looks good, thanks again for the explanation ! :)
> The "body" of namespaces does not get indented right ?

yes

include/seqan3/contrib/charconv/charconv.cpp
```diff
@@ -0,0 +1,235 @@
+//===------------------------- charconv.cpp -------------------------------===//
```
smehringer, line 1: What's this file for?

include/seqan3/contrib/charconv/charconv_pre.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+#pragma once
+
+#include <cassert>
+#include <system_error>
+#include <limits>
+
+#include <seqan3/core/platform.hpp>
+
+#define SEQAN3_CONTRIB_CHARCONV_LIBCPP_BEGIN_NAMESPACE_STD namespace seqan3::contrib::charconv {
```
smehringer, line 16: It's a little confusing that the macro name states "begin std namespace" but you are opening `seqan3::contrib::charconv`?

include/seqan3/contrib/charconv/charconv_pre.hpp
```diff
@@ -0,0 +1,57 @@
...
+#define SEQAN3_CONTRIB_CHARCONV_LIBCPP_END_NAMESPACE_STD }
+// #define SEQAN3_CONTRIB_CHARCONV_LIBCPP_HAS_NO_PRAGMA_SYSTEM_HEADER 1 // needs to be set if compiling with MSVC
```
smehringer, line 18: Why don't you wrap this in 
```
#if defined (_MSC_VER)
// ...
#endif
```
? (not sure if this macro is the right one)

include/seqan3/contrib/charconv/charconv_pre.hpp
```diff
@@ -0,0 +1,57 @@
...
+// #define SEQAN3_CONTRIB_CHARCONV_LIBCPP_PUSH_MACROS
+// #define SEQAN3_CONTRIB_CHARCONV_LIBCPP_POP_MACROS
+// #define SEQAN3_CONTRIB_CHARCONV_LIBCPP_AVAILABILITY_TO_CHARS
+#define SEQAN3_CONTRIB_CHARCONV_LIBCPP_FUNC_VIS
+#define SEQAN3_CONTRIB_CHARCONV_LIBCPP_ENUM_VIS
+#define SEQAN3_CONTRIB_CHARCONV_LIBCPP_TYPE_VIS
+// #define SEQAN3_CONTRIB_CHARCONV_LIBCPP_CONSTEXPR constexpr
+#define SEQAN3_CONTRIB_CHARCONV_LIBCPP_HIDDEN
+#define SEQAN3_CONTRIB_CHARCONV_LIBCPP_INLINE_VISIBILITY
+#define SEQAN3_CONTRIB_CHARCONV_LIBCPP_ASSERT(condition, message) assert(condition)
```
smehringer, line 30: Can you shortly explain the neccessity of these macros? (I am not doubting them, just not understanding them)

include/seqan3/contrib/charconv/charconv_pre.hpp
```diff
@@ -0,0 +1,57 @@
...
+// import things from std namespace which charconv uses
+namespace seqan3::contrib::charconv
+{
+using std::errc;
+using std::declval;
+using std::numeric_limits;
+using std::is_integral;
+using std::is_unsigned;
+using std::is_signed;
+// using std::make_unsigned;
```
smehringer, line 41: can you drop a comment why this is out commented?
Same for `using std::enable_if;`

test/unit/std/charconv_int_test.cpp
```diff
@@ -0,0 +1,253 @@
...
+TYPED_TEST(integral_from_char_test, to_chars_error)
+{
+    TypeParam val{120};
+    std::array<char, 1> buffer{};
+
+    auto res = std::to_chars(buffer.data(), buffer.data() + buffer.size(), val);
+
+    EXPECT_EQ(res.ptr, buffer.data() + buffer.size());
+    EXPECT_EQ(res.ec, std::errc::value_too_large);
+}
```
smehringer, line 253: Can you add the test mentioned in the issue https://github.com/seqan/seqan3/issues/1595 that checks whether the error still occurs?
Sorry if it is there but I overlooked it.

include/seqan3/contrib/charconv/charconv.cpp
marehr: https://github.com/llvm/llvm-project/blob/d27489645bb8829a952f79e3b211202844238391/libcxx/src/charconv.cpp

We import the file as is and try to do as little modification as possible.

include/seqan3/contrib/charconv/charconv_pre.hpp
marehr: true, but I want to keep the files as close to their original as possible.

include/seqan3/contrib/charconv/charconv_pre.hpp
marehr: We don't support MSVC right now, so I think I don't need to do this. Fun fact: MSVC has full charconv support, so this header should not be included.

include/seqan3/contrib/charconv/charconv_pre.hpp
marehr: These are all prefixed MACROS that the original implementation https://github.com/llvm/llvm-project/blob/d27489645bb8829a952f79e3b211202844238391/libcxx/include/charconv uses.

Our previous approach of renaming and seqan3-ifing external things was wrong, because an update is REALLY hard to do. The new approach is to keep the original sources as close as they are and only do changes that are deemed necessary.

include/seqan3/contrib/charconv/charconv_pre.hpp
marehr: I can remove them :)

test/unit/std/charconv_int_test.cpp
marehr: No, like the description of this PR said

> I didn't upgrade the llvm version yet, because they had a license change to `apache2 WITH llvm-exception`

and 

> My second reason was to test the upgrade-path if we ever need to add new version of this.
And my third reason: Like always, a small change can lead to a couple of unintended CI breakage; I wanted to see if everything works with the new structure which uses the old code.

This means that I just copied the original file of charconv that we already used.
The next PR will update the charconv implementation and adds a test case.

The reason for this:
* Use a known working version of charconv so that the infra structure change is known to work.
* The new code has a new license which I wanted to discuss with you.
* Test the update-path. That means is a upgrade of charconv as painless as possible.

include/seqan3/contrib/charconv/charconv_pre.hpp
marehr: I removed them, because I don't need to define them.

include/seqan3/contrib/charconv/charconv_pre.hpp
marehr: I removed them :)

smehringer:
Marcel wanted to clean the commit history before merging

PR #1838, opened 2020-05-22
Fix ci doxygen

marehr:
The main problem was that the test command

```c++
    test -z \"\$output\";
    ! find . -not -name \"*_source.html\" -name \"*.html\" -print0 | xargs -0 grep \"requires\" | grep \"memname\"
```

ignored the return value of the first command. (whether doxygen gave back warnings or not)


marehr:
> oh. At least it seems that we did not introduce any warnings in the meantime, right?

yeah xD

PR #1849, opened 2020-05-26
[FIX] cxx20: import std::cpp20::back_inserter

marehr:
This is a left-over of #1828

I forgot that `gcc-10 -std=c++2a` needs to define the std 20 version of `std::back_inserter` in our `std::cpp20` namespace.

### DIFF
--- a/include/seqan3/std/iterator
+++ b/include/seqan3/std/iterator
@@ -15,6 +15,20 @@
 
 #include <iterator>
 
+// import API that changed behaviour from C++-17 to C++20 into std::cpp20
+#ifdef __cpp_lib_ranges
+namespace std
+{
+inline namespace cpp20
+{
+// see https://github.com/orgs/seqan/projects/4#card-37029069
+using ::std::back_inserter;
+// using ::std::front_inserter;
+// using ::std::insert_iterator;
+} // inline namespace cpp20
+} // namespace std
+#endif // __cpp_lib_ranges
+
 #ifndef __cpp_lib_ranges // implement C++20 iterators via range-v3
 
 #ifndef RANGES_DEEP_STL_INTEGRATION


include/seqan3/std/iterator
```diff
@@ -15,6 +15,20 @@
...
+// import API that changed behaviour from C++-17 to C++20 into std::cpp20
+#ifdef __cpp_lib_ranges
+namespace std
+{
+inline namespace cpp20
+{
+// see https://github.com/orgs/seqan/projects/4#card-37029069
+using ::std::back_inserter;
+// using ::std::front_inserter;
+// using ::std::insert_iterator;
```
smehringer, line 27: what's up with these two?

marehr: We thought we have the policy only to import C++20 stuff that we actually use.

smehringer:
```suggestion
```
Alright, but why leave them here out commented?

marehr: It's the general style of this file, see

https://github.com/seqan/seqan3/blob/8adb0b8aa5e095439c5ca1cd74bd630d369036e6/include/seqan3/std/iterator#L168-L175

We want this, to make it easier for someone to comment it in if they need it.

smehringer: > We want this, to make it easier for someone to comment it in if they need it.

alright.

smehringer:
can you check jenkins?

marehr:
Upload to cdash failed:

> Size: 16573642
   Error when uploading file: /Users/builder/jenkins/ws/seqan3_build_pipeline_PR-1849@2/build/Testing/20200526-0059/Test.xml
   Error message was: Operation too slow. Less than 1 bytes/sec transferred the last 120 seconds
   Problems when submitting via HTTP
script returned exit code 255

PR #1850, opened 2020-05-26
[FEATURE] Add comparability skill to strong type.

smehringer:
This comes in very handy in a lot of cases.
E.g. consider 
```cpp
    using t = std::variant<my_strong_type1, my_strong_type2>;
    t v1{my_strong_type1{2}};
    t v2{my_strong_type1{3}};
```
you can neither do 
```cpp
    v1 == v2 // error, no comparison operator for my_strong_type1
```
nor
```cpp
    v1.get() == v2.get() // error, variant has no member get
```
You would need some nested std::visit for the two variants which is ugly.

### DIFF
--- a/include/seqan3/core/detail/strong_type.hpp
+++ b/include/seqan3/core/detail/strong_type.hpp
@@ -391,7 +391,7 @@ public:
     }
     //!\}
 
-    /*!\name Comparison operators.
+    /*!\name Comparison operators
      * \brief Only available if the corresponding skill from seqan3::detail::strong_type_skill is added.
      *
      * \if DEV
@@ -399,22 +399,22 @@ public:
      * \endif
      * \{
      */
-    //!\brief Return whether this instance is equal to `other`.
+    //!\brief Return whether this instance is equal to `rhs`.
     constexpr bool operator==(strong_type const & other) const
     //!\cond
         requires ((skills & strong_type_skill::comparable) != strong_type_skill::none)
     //!\endcond
     {
-        return get() == other.get();
+        return get() == rhs.get();
     }
 
-    //!\brief Return whether this instance is not equal to `other`.
-    constexpr bool operator!=(strong_type const & other) const
+    //!\brief Return whether this instance is not equal to `rhs`.
+    constexpr bool operator!=(strong_type const & rhs) const
     //!\cond
         requires ((skills & strong_type_skill::comparable) != strong_type_skill::none)
     //!\endcond
     {
-        return !(*this == other);
+        return !(*this == rhs);
     }
     //!\}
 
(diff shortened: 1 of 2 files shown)


marehr:
I reassigned, because I have a lot of open reviews.

include/seqan3/core/detail/strong_type.hpp
```diff
@@ -390,6 +391,33 @@ class strong_type
     }
     //!\}
 
+    /*!\name Comparison operators.
```
rrahn, line 394:
```suggestion
    /*!\name Comparison operators
```
No dots at the end of name sections.

include/seqan3/core/detail/strong_type.hpp
```diff
@@ -390,6 +391,33 @@ class strong_type
...
+     * \brief Only available if the corresponding skill from seqan3::detail::strong_type_skill is added.
+     *
+     * \if DEV
+     * Implemented as member functions because requires does not work on friends.
+     * \endif
+     * \{
+     */
+    //!\brief Return whether this instance is equal to `other`.
```
rrahn, line 402:
```suggestion
    //!\brief Return whether this instance is equal to `rhs`.
```
Very often we use the term lhs and rhs for the comparison operations. I would like to keep this as much as possible.

include/seqan3/core/detail/strong_type.hpp
```diff
@@ -390,6 +391,33 @@ class strong_type
...
+    }
+
+    //!\brief Return whether this instance is not equal to `other`.
+    constexpr bool operator!=(strong_type const & other) const
+    //!\cond
+        requires ((skills & strong_type_skill::comparable) != strong_type_skill::none)
+    //!\endcond
+    {
+        return !(*this == other);
+    }
```
rrahn, line 418:
```suggestion
        requires ((skills & strong_type_skill::comparable) != strong_type_skill::none)
    //!\endcond
    {
        return get() == rhs.get();
    }

    //!\brief Return whether this instance is not equal to `rhs`.
    constexpr bool operator!=(strong_type const & rhs) const
    //!\cond
        requires ((skills & strong_type_skill::comparable) != strong_type_skill::none)
    //!\endcond
    {
        return !(*this == rhs);
    }
```

include/seqan3/core/detail/strong_type.hpp
```diff
@@ -390,6 +391,33 @@ class strong_type
...
+    /*!\name Comparison operators
+     * \brief Only available if the corresponding skill from seqan3::detail::strong_type_skill is added.
+     *
+     * \if DEV
+     * Implemented as member functions because requires does not work on friends.
+     * \endif
+     * \{
+     */
+    //!\brief Return whether this instance is equal to `rhs`.
+    constexpr bool operator==(strong_type const & other) const
```
smehringer, line 403:
```suggestion
    constexpr bool operator==(strong_type const & rhs) const
```

PR #1706, opened 2020-04-01
[FEATURE] search result struct

smehringer:
supersedes #1441 
~blocked by #1799~

For the reviewer: 
I split the changes into for logical commits because I think it makes it easier to review. Since the commits do not compiler without each other though, I would squash them in the end again.

To commit 1 (`Unify cursor.locate() return value`): In order to easily adapt the `make_result` function in the policy_result_builder, it was convenient to change the return value of `cursor.locate()` and I had in mind that we want to unify them both anyway. I only noticed the tail of tests I needed to adapt after everything was done. I'm sorry that this made the commit unnecessary heavy but we need this anyway.

Fixes https://github.com/seqan/seqan3/issues/1434
Part of https://github.com/seqan/product_backlog/issues/28

PR #1853, opened 2020-05-27
Search config/main dynamic hit config

rrahn:
### Description

Sometimes a program needs to support different hit strategies based on some user input. Since these are mostly runtime decisisons the code can become quite cumbersome to handle the static hit configurations. Instead, one can use the dynamic hit configuration element seqan3::search::cfg::hit. This configuration element allows to set one of the above mentioned hit configurations at runtime. Later during the configuration phase of the search algorithm the selected search configuration is used for the final search algorithm. If the dynamic hit configuration is default constructed it does not hold any hit configuration. If you call search with the dynamic configuration in this state an exception will be thrown.
Also note that using the dynamic configuration might have implications on the compile time, so we recommend to use the static configurations if only a single hit strategy is supported.

### Changes

The follwoing things are handled in this PR:

* Add dynamic hit configuration including test and documentation
* Move the search configurator in it's own class
* Let the search configurator return a type erased std::function object
* Adapts the search configurator to handle dynamic hit configuration
* Add result type configuration element to store the configured result type for the search
* Adapts the search traits type
* Renames policy_result_builder to policy_search_result_builder and makes it a template class with the configuration template.
* Refactors a little the search_trivial test to not depend on the internal function of the unidirectional search algorithm anymore.

Also note that this PR kind of depends on #1706. So some parts are commented but in general this already works.

### Fixes

Fixes seqan/product_backlog#79
Fixes seqan/product_backlog#31
Fixes seqan/seqan3#1436

### DIFF
--- a/include/seqan3/search/configuration/hit.hpp
+++ b/include/seqan3/search/configuration/hit.hpp
@@ -15,8 +15,11 @@
 
 #pragma once
 
+#include <variant>
+
 #include <seqan3/core/algorithm/pipeable_config_element.hpp>
 #include <seqan3/core/detail/empty_type.hpp>
+#include <seqan3/core/type_list/traits.hpp>
 #include <seqan3/search/configuration/detail.hpp>
 
 namespace seqan3::detail
@@ -85,6 +88,77 @@ inline constexpr detail::hit_single_best_tag hit_single_best{};
  */
 struct hit_strata : public pipeable_config_element<hit_strata, uint8_t>
 {
+    //!\privatesection
+    //!\brief Internal id to check for consistent configuration settings.
+    static constexpr detail::search_config_id id{detail::search_config_id::hit};
+};
+
+/*!\brief A dynamic configuration element to configure the hit strategy at runtime.
+ * \ingroup search_configuration
+ * \sa \ref search_configuration_subsection_hit_strategy "Section on Hit Strategy"
+ */
+class hit : public pipeable_config_element<hit, seqan3::detail::empty_type>
+{
+public:
+    /*!\brief The type of the std::variant holding the hit configuration element alternatives
+     *
+     * The additional detail::empty_type marks the seqan3::search_cfg::hit as default constructed, with no selected hit
+     * configuration which can be checked within the search algorithm.
+     */
+    using hit_variant_type = std::variant<detail::empty_type,
+                                          detail::hit_all_tag,
+                                          detail::hit_all_best_tag,
+                                          detail::hit_single_best_tag,
+                                          hit_strata>;
+
+    /*!\name Constructors, assignment and destructor
+     * \{
+     */
+    hit() = default; //!< Defaulted.
+    hit(hit const &) = default; //!< Defaulted.
+    hit(hit &&) = default; //!< Defaulted.
+    hit & operator=(hit const &) = default; //!< Defaulted.
+    hit & operator=(hit &&) = default; //!< Defaulted.
+    ~hit() = default; //!< Defaulted.
+
+    /*!\brief Sets the given configuration element to the dynamic hit configuration element.
+     *
+     * \tparam hit_config_t The type of the hit configuration element to set.
+     * \param[in] hit_config The hit configuration element to set.
+     *
+     * \details
+     *
+     * Only one of the static hit configuration elements are valid: seqan3::search_cfg::hit_all,
+     * seqan3::search_cfg::hit_all_best, seqan3::search_cfg::hit_single_best and seqan3::search_cfg::hit_strata.
+     */
+    template <typename hit_config_t>
+    //!\cond
+        requires pack_traits::contains<hit_config_t, detail::hit_all_tag,
+                                                     detail::hit_all_best_tag,
+                                                     detail::hit_single_best_tag,
+                                                     hit_strata>
+    //!\endcond
+    explicit hit(hit_config_t hit_config) noexcept : hit_variant{std::move(hit_config)}
+    {}
+
+    //!\copydoc seqan3::search_cfg::hit::hit(hit_config_t hit_config)
+    template <typename hit_config_t>
+    //!\cond
+        requires pack_traits::contains<hit_config_t, detail::hit_all_tag,
+                                                     detail::hit_all_best_tag,
+                                                     detail::hit_single_best_tag,
+                                                     hit_strata>
+    //!\endcond
+    hit & operator=(hit_config_t hit_config) noexcept
+    {
+        hit_variant = std::move(hit_config);
+        return *this;
+    }
+    //!\}
+
+    //!\brief A std::variant over the valid hit configuration elements.
+    hit_variant_type hit_variant{};
+
     //!\privatesection
     //!\brief Internal id to check for consistent configuration settings.
     static constexpr detail::search_config_id id{detail::search_config_id::hit};
--- a/include/seqan3/search/configuration/max_error_rate.hpp
+++ b/include/seqan3/search/configuration/max_error_rate.hpp
@@ -18,6 +18,7 @@
 #include <seqan3/core/algorithm/configuration.hpp>
 #include <seqan3/core/algorithm/pipeable_config_element.hpp>
 #include <seqan3/core/detail/pack_algorithm.hpp>
+#include <seqan3/core/type_traits/template_inspection.hpp>
 #include <seqan3/range/views/slice.hpp>
 #include <seqan3/search/configuration/detail.hpp>
 #include <seqan3/search/configuration/max_error_common.hpp>
@@ -38,10 +39,10 @@ namespace seqan3::search_cfg
 template <typename ...errors_t>
 //!\cond
     requires (sizeof...(errors_t) <= 4) &&
-             ((detail::is_type_specialisation_of_v<std::remove_reference_t<errors_t>, total> ||
-               detail::is_type_specialisation_of_v<std::remove_reference_t<errors_t>, substitution> ||
-               detail::is_type_specialisation_of_v<std::remove_reference_t<errors_t>, deletion>  ||
-               detail::is_type_specialisation_of_v<std::remove_reference_t<errors_t>, insertion>) && ...)
+             ((seqan3::detail::is_type_specialisation_of_v<std::remove_reference_t<errors_t>, total> ||
+               seqan3::detail::is_type_specialisation_of_v<std::remove_reference_t<errors_t>, substitution> ||
+               seqan3::detail::is_type_specialisation_of_v<std::remove_reference_t<errors_t>, deletion>  ||
+               seqan3::detail::is_type_specialisation_of_v<std::remove_reference_t<errors_t>, insertion>) && ...)
 //!\endcond
 class max_error_rate : public pipeable_config_element<max_error_rate<errors_t...>, std::array<double, 4>>
 {
@@ -77,7 +78,7 @@ public:
 
     //!\privatesection
     //!\brief Internal id to check for consistent configuration settings.
-    static constexpr detail::search_config_id id{detail::search_config_id::max_error_rate};
+    static constexpr seqan3::detail::search_config_id id{seqan3::detail::search_config_id::max_error_rate};
 
     //!\publicsection
     /*!\name Constructor, destructor and assignment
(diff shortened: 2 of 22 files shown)


include/seqan3/search/detail/search_configurator.hpp
```diff
@@ -0,0 +1,125 @@
...
+     * \tparam index_t The type of the index.
+     * \param[in] cfg The search configuration object that is passed to the algorithm.
+     * \param[in] index The index that is passed to the algorithm.
+     * \returns A search algorithm.
+     *
+     * \details
+     *
+     * If the `index_t` models seqan3::bi_fm_index_specialisation, then the
+     * seqan3::detail::search_scheme_algorithm is chosen. Otherwise, the
+     * detail::unidirectional_search_algorithm is chosen.
```
simonsasse, line 150: Very picky but shouldn't this also be **seqan3**::detail_unidirectional_search_algorithm ?

rrahn, line 150:
```suggestion
     * seqan3::detail::unidirectional_search_algorithm is chosen.
```

smehringer:
Can you resolve the conflicts before I review? Thank you :)

rrahn:
@smehringer done

include/seqan3/search/configuration/result_type.hpp
```diff
@@ -0,0 +1,72 @@
...
+ * \see seqan3::search_cfg::result_type
+ */
+template <typename search_result_t>
+//!\cond
+    requires seqan3::detail::is_type_specialisation_of_v<search_result_t, search_result>
+//!\endcond
+struct result_type_tag :
+    public pipeable_config_element<result_type_tag<search_result_t>,
+                                   std::type_identity<search_result_t>>
+{
```
smehringer, line 44:
```suggestion
struct result_type_tag : public pipeable_config_element<result_type_tag<search_result_t>, empty_type>
{
    //!\brief The result type to propagate.
    using search_result_type = search_result_t;
    
```
Since the `value` member will go away sooner or later anyway, wouldn't it be more convenient to just make this empty_type and use a type member variable instead? Then you can use e.g. `search_cfg::result_type::search_result_type` instead of decltyping value and then removing type identity.

If you use `using type = ...` you might even use this as kind of a type trait later on

include/seqan3/search/detail/search_configurator.hpp
```diff
@@ -25,6 +33,46 @@ namespace seqan3::detail
...
+     * \tparam index_t The type of the index.
+     * \tparam policies_t A template parameter pack over the policies to specify the behavior of the algorithm.
+     */
+    template <typename configuration_t, typename index_t, typename ...policies_t>
+    struct select_search_algorithm
+    {
+        //!\brief The selected algorithm type based on the index.
+        using type =
+            lazy_conditional_t<bi_fm_index_specialisation<index_t>,
+                               lazy<instantiate_t,
```
smehringer, line 70: what or who is `instantiate_t`?

include/seqan3/search/detail/search_configurator.hpp
```diff
@@ -100,26 +150,87 @@ class search_configurator
...
+
+        return selected_algorithm_t{new_cfg, index};
+    };
+
+    // Check if dynamic config present, otherwise continue.
+    if constexpr (configuration_t::template exists<search_cfg::hit>())
+    {
+        auto hit_variant = get<search_cfg::hit>(cfg).value;
+
+        if (std::holds_alternative<empty_type>(hit_variant))
```
smehringer, line 209: Hm I wonder if this is better than just setting letting the variant default to `hit_all` on initialisation as this is also the default when no hit configuration is given?

include/seqan3/search/detail/search_configurator.hpp
```diff
@@ -100,26 +150,87 @@ class search_configurator
...
+ */
+template <typename algorithm_t, typename configuration_t, typename index_t>
+algorithm_t search_configurator::configure_hit_strategy(configuration_t const & cfg, index_t const & index)
+{
+    // Select the actual search algorithm and return an instance constructed with the adapted config and the index.
+    auto generate_algorithm = [&] (auto new_cfg) -> algorithm_t
+    {
+        using new_configuration_t = decltype(new_cfg);
+        using selected_algorithm_t =
+            typename select_search_algorithm<new_configuration_t,
```
smehringer, line 196: It is a little unintuitive that the function `configure_hit_strategy` also selects the search algorithm. Can this be done outside of this function?

include/seqan3/search/configuration/max_error_rate.hpp
```diff
@@ -38,10 +39,10 @@ namespace seqan3::search_cfg
 template <typename ...errors_t>
 //!\cond
     requires (sizeof...(errors_t) <= 4) &&
-             ((detail::is_type_specialisation_of_v<std::remove_reference_t<errors_t>, total> ||
-               detail::is_type_specialisation_of_v<std::remove_reference_t<errors_t>, substitution> ||
-               detail::is_type_specialisation_of_v<std::remove_reference_t<errors_t>, deletion>  ||
-               detail::is_type_specialisation_of_v<std::remove_reference_t<errors_t>, insertion>) && ...)
+             ((seqan3::detail::is_type_specialisation_of_v<std::remove_reference_t<errors_t>, total> ||
```
smehringer, line 42: Can you maybe merge this commit with the one that introduces the error? s.t. the commit history is mostly "stable" meaning that if I do git bisect I have commits that pass CI?

include/seqan3/search/configuration/result_type.hpp
```diff
@@ -0,0 +1,72 @@
...
+#pragma once
+
+#include <seqan3/std/type_traits>
+
+#include <seqan3/core/algorithm/pipeable_config_element.hpp>
+#include <seqan3/core/type_traits/template_inspection.hpp>
+#include <seqan3/search/configuration/detail.hpp>
+#include <seqan3/search/search_result.hpp>
+
+namespace seqan3::search_cfg::detail
```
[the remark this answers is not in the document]
rrahn: yes we decided this. and it needs to be adjusted for the other elements as well

include/seqan3/search/detail/search_configurator.hpp
rrahn: It is a deferred instantiation template of the given type and it's arguments. It is defined in `core/type_traits/lazy.hpp`

include/seqan3/search/detail/search_configurator.hpp
rrahn: I could make an explicit function to just handle the algorithm selection. The idea is to have  a continuation interface: like configure a -> configure b -> configure c -> .... But I also often received the feedback that for 2 lines of code an additional function bloat is not seen. BUt I also agree, that this way we can document the behavior much easier and are flexible with the configuration chain. Then again it keeps the changes minimal and we don't add a feature that we don't neeed yet. I really don't care. If I see for both points pros and cons. Just let me know what you'd prefer.

include/seqan3/search/configuration/max_error_rate.hpp
rrahn: I think all commits are compilable and runnable. At least I spent a lot of time to verify this. But I'll check again.

include/seqan3/search/detail/search_configurator.hpp
rrahn: For the configuartion we decided to not infer a default on the construction but let the algorithm handle the default. But even then I wasn't sure what is better in this situation. So currently we infer some default if the user provides nothing. But the default would be the slowest configuration finding all hits. As far as I know this is also something that you don't want in many scenarios. The question here is the same with the alignment. Whether global or local should be default is very dependent on the use case and nothing we should default but require from the user to be set. Now the user explicitly wants to set the dynamic hit strategy but made a mistake in the code and actually never sets anything but doesn't know it (for example he would like to never use all_hits strategy). In this case we would assume a default which might be something the user really don't want and he/she either get's wrong results or is wondering about the performance or memory consumption etc. So from that perspective I think it would be wiser to not assume a default configuration here but let the user know that the search was misconfigured. Defaults are only good if they do the right thing for the user, but I don't see this here.

include/seqan3/search/configuration/max_error_rate.hpp
rrahn: Ah but this is only for the header test. Boah, no one is bisecting the header test, right?! That would take hours because you have to compile everything.

include/seqan3/search/detail/search_configurator.hpp
rrahn: Well, I kind of made it more complex then necessary
I removed the additional instantiate_t. Thanks for spotting 👍

include/seqan3/search/detail/search_configurator.hpp
rrahn: I separated it now in the last fixup commit

include/seqan3/search/configuration/max_error_rate.hpp
smehringer: true, I just read test not header test. leave it as it is.

include/seqan3/search/configuration/hit.hpp
```diff
@@ -90,4 +93,71 @@ struct hit_strata : public pipeable_config_element<hit_strata, uint8_t>
...
 
+/*!\brief A dynamic configuration element to configure the hit strategy at runtime.
+ * \ingroup search_configuration
+ * \sa \ref search_configuration_subsection_hit_strategy "Section on Hit Strategy"
+ */
+class hit : public pipeable_config_element<hit, seqan3::detail::empty_type>
+{
+public:
+    //!\brief The type of the std::variant holding the hit configuration element alternatives.
+    using hit_variant_type = std::variant<detail::empty_type,
```
smehringer, line 104:
```suggestion
    /*!\brief The type of the std::variant holding the hit configuration element alternatives
     *
     * The additional detail::empty_type marks the seqan3::search_cfg::hit as default constructed, with no selected hit
     * configuration which can be checked within the search algorithm.
     */
    using hit_variant_type = std::variant<detail::empty_type,
```
Or something along these lines. because it is public and the user might wonder what empty_type stands for.

include/seqan3/search/configuration/result_type.hpp
```diff
@@ -0,0 +1,73 @@
...
+/*!\brief Configuration element storing the configured seqan3::search_result for the search algorithm.
+ * \ingroup search_configuration
+ * \tparam search_result_t The search result type to store; must be a type specialisation of seqan3::search_result.
+ *
+ * \details
+ *
+ * This configuration element stores the seqan3::search_result type after configuring the
+ * search algorithm with the seqan3::detail::search_configurator. The actual result type is wrapped in
+ * std::type_identity to preserve the trivial type properties of the configuration element. Thus, on access the
+ * actual type needs to be unwrapped using the member typedef `type` before it can be used.
```
smehringer, line 59:
```suggestion
 * search algorithm with the seqan3::detail::search_configurator.
```

include/seqan3/search/configuration/result_type.hpp
```diff
@@ -0,0 +1,73 @@
...
+ * The result type can be accessed via the seqan3::detail::search_traits over the corresponding
+ * search configuration type.
+ * If the stored search result was not added yet to the search configuration the corresponding
+ * result type member will deduce to seqan3::detail::empty_type.
```
smehringer, line 63: Is this still the case?

test/snippet/search/dynamic_hit_configuration_example.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+int main()
+{
+    // Default constructed: Has no hit strategy selected.
+    seqan3::search_cfg::hit dynamic_hit{};
+
+    // Select hit_all
+    dynamic_hit = seqan3::search_cfg::hit_all;
+
+    // If condition is true chose strata strategy, otherwise find the single best hit.
```
smehringer, line 11:
```suggestion
    // If condition is true choose strata strategy, otherwise find the single best hit.
```

include/seqan3/search/configuration/hit.hpp
rrahn: I already have a ticket to make empty type public since we are using it now quite a lot in a public API.

PR #1855, opened 2020-05-27
Parallelise search/part3 enabel executor in search algorithm

rrahn:
Updates the two search algorithms be invoked with a second callback argument.
This callback is invoked with the generated search results.
Also the search algorithm now associates an index
with every query and propagates this indexed query pair to the algorithms.
Updates the search result range to work with the algorithm executor.

part of seqan/product_backlog#27


rrahn:
Superseded by #1880

PR #1856, opened 2020-05-28
Clemapfel fastq benchmark

rrahn:
Supersedes #961 

Final celanup and finishing of the initial fastq benchmark:

```
Running ./io/format_fastq_benchmark
Run on (4 X 3600 MHz CPU s)
CPU Caches:
  L1 Data 32K (x4)
  L1 Instruction 32K (x4)
  L2 Unified 262K (x4)
  L3 Unified 6291K (x1)
Load Average: 3.18, 3.32, 3.34
----------------------------------------------------------------------------------------------
Benchmark                                    Time             CPU   Iterations UserCounters...
----------------------------------------------------------------------------------------------
fastq_write_to_stream_seqan3/100         25380 ns        25067 ns        30960 bytes_per_run=12k bytes_per_second=456.546M/s iterations_per_run=100
fastq_write_to_stream_seqan3/1000       231688 ns       226090 ns         3629 bytes_per_run=120k bytes_per_second=506.175M/s iterations_per_run=1000
fastq_write_to_stream_seqan3/10000     2239495 ns      2221013 ns          388 bytes_per_run=1.2M bytes_per_second=515.265M/s iterations_per_run=10k
fastq_read_from_stream_seqan3/100        59234 ns        58034 ns        11928 bytes_per_run=12k bytes_per_second=197.196M/s iterations_per_run=100
fastq_read_from_stream_seqan3/1000      579207 ns       573936 ns         1166 bytes_per_run=120k bytes_per_second=199.397M/s iterations_per_run=1000
fastq_read_from_stream_seqan3/10000    5998109 ns      5768174 ns          121 bytes_per_run=1.2M bytes_per_second=198.401M/s iterations_per_run=10k
fastq_read_from_disk_seqan3/100         131451 ns       130473 ns         4998 bytes_per_run=12k bytes_per_second=87.7123M/s iterations_per_run=100
fastq_read_from_disk_seqan3/1000        730356 ns       720919 ns          857 bytes_per_run=120k bytes_per_second=158.743M/s iterations_per_run=1000
fastq_read_from_disk_seqan3/10000      6452693 ns      6257824 ns          108 bytes_per_run=1.2M bytes_per_second=182.877M/s iterations_per_run=10k
fastq_read_from_stream_seqan2/100        21331 ns        21224 ns        32040 bytes_per_run=12k bytes_per_second=539.202M/s iterations_per_run=100
fastq_read_from_stream_seqan2/1000      212555 ns       211392 ns         3200 bytes_per_run=120k bytes_per_second=541.367M/s iterations_per_run=1000
fastq_read_from_stream_seqan2/10000    2232872 ns      2156875 ns          328 bytes_per_run=1.2M bytes_per_second=530.587M/s iterations_per_run=10k
fastq_read_from_disk_seqan2/100          67768 ns        67455 ns         7616 bytes_per_run=12k bytes_per_second=169.656M/s iterations_per_run=100
fastq_read_from_disk_seqan2/1000        391385 ns       376115 ns         1879 bytes_per_run=120k bytes_per_second=304.271M/s iterations_per_run=1000
fastq_read_from_disk_seqan2/10000      3406304 ns      3365443 ns          210 bytes_per_run=1.2M bytes_per_second=340.047M/s iterations_per_run=10k
```


test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,242 @@
...
+    seqan3::debug_stream_type format_stream{stream_buffer};
+
+    std::string const id{"@name"};
+
+    auto seed = default_seed;
+
+    for (size_t i = 0; i < entries_size; ++i, ++seed)
+    {
+        auto random_sequence = seqan3::test::generate_sequence<seqan3::dna5>(default_sequence_length, 0, seed);
+        auto random_qualties = seqan3::test::generate_sequence<seqan3::phred42>(default_sequence_length, 0, seed);
```
joergi-w, line 48: As you already use `default_sequence_length` I suggest deleting line 43 and use
```suggestion
        auto random_sequence = seqan3::test::generate_sequence<seqan3::dna5>(default_sequence_length, 0, default_seed);
        auto random_qualities = seqan3::test::generate_sequence<seqan3::phred42>(default_sequence_length, 0, default_seed);
```

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,242 @@
...
+        format_stream << id << '\n'  // write id
+                      << random_sequence << '\n' // write sequence
+                      << '+' << '\n' // write qualities separator
+                      << random_qualties << '\n'; // write qualities
```
joergi-w, line 52: Here is a typo:
```suggestion
                      << random_qualities << '\n'; // write qualities
```

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    return stream_buffer.str();
+}
+
+// ============================================================================
+// save file on disc temporarily
+// ============================================================================
+
+auto create_fastq_file(size_t const entries)
```
joergi-w, line 62: For consistency with the function parameter above (adapt also line 70):
```suggestion
auto create_fastq_file(size_t const entries_size)
```

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,242 @@
...
+    std::string file = generate_fastq_file(state.range(0));
+
+    seqan3::dna5_vector sequence{};
+    std::string id{};
+    std::vector<seqan3::phred42> quality{};
+
+    for (auto _ : state)
+    {
+        state.PauseTiming();
+        std::istringstream istream{file};
```
joergi-w, line 111: Why don't we move this line before the loop, after `std::string file` is created? 
`file` does not change in this function and I do not see why we should create new istreams all the time.

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+void fastq_read_from_stream_seqan2(benchmark::State & state)
+{
+    std::istringstream istream{};
+    size_t entries_size = state.range(0);
+    auto file = generate_fastq_file(entries_size);
+
+    auto restart_iterator = [&istream, &file]()    // cf. format_fasta_benchmark
+    {
+        istream = std::istringstream{file};
```
joergi-w, line 176: the same here

test/performance/io/format_fastq_benchmark.cpp
rrahn: because it will be consumed after the first execution of the benchmark since it is an input stream. Hence calling it again without resetting it before will give a wrong benchmark. 
I could try to set the underlying buffer, which might be a non-invasive operation, but if handling it inside of a paused time block has no time effects on the benchmark the more complicated code seems not worth the effort.

rrahn:
> Thank you for working on this! I have a few suggestions and I'm wondering why you have not implemented the write_to_disk and write_seqan2 benchmarks as well. Maybe they are a future task or not needed at all?

I just cleaned up the initial benchmark since it was almost done but then never worked on so we get these old open PRs merged. I did not intend to make it complete. We will do that when we actually work on the IO.

test/performance/io/format_fastq_benchmark.cpp
joergi-w: Okay, then it's fine!

test/performance/io/format_fastq_benchmark.cpp
rrahn: Ah the seed seems to be varied for every generated sequence.

joergi-w: true ;)

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+#include <benchmark/benchmark.h>
+
+#include <cstring>
+#include <seqan3/std/filesystem>
+#include <fstream>
+
+#include <seqan3/alphabet/nucleotide/dna5.hpp>
+#include <seqan3/alphabet/quality/phred42.hpp>
+#include <seqan3/core/detail/debug_stream_type.hpp>
```
smehringer, line 16:
```suggestion
```

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,242 @@
...
+    auto seed = default_seed;
+
+    for (size_t i = 0; i < entries_size; ++i, ++seed)
+    {
+        auto random_sequence = seqan3::test::generate_sequence<seqan3::dna5>(default_sequence_length, 0, seed);
+        auto random_qualities = seqan3::test::generate_sequence<seqan3::phred42>(default_sequence_length, 0, seed);
+        format_stream << id << '\n'  // write id
+                      << random_sequence << '\n' // write sequence
+                      << '+' << '\n' // write qualities separator
+                      << random_qualities << '\n'; // write qualities
```
smehringer, line 52:
```suggestion
        format_stream.emplace_back(random_sequence, id, random_qualities);
```
No need to do it manually? Will also work on Windows (line breaks)

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,242 @@
...
+    std::string const id{"@name"};
+    auto seq = seqan3::test::generate_sequence<seqan3::dna5>(default_sequence_length, 0, default_seed);
+    auto qual = seqan3::test::generate_sequence<seqan3::phred42>(default_sequence_length, 0, default_seed);
+
+    for (auto _ : state)
+        fout.emplace_back(seq, id, qual);
+}
+
+// ============================================================================
+// seqan3 benchmark
```
smehringer, line 93:
```suggestion
// seqan3 fastq input benchmark
```

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+    // fill temporary file with a fastq file
+    std::ofstream ostream{fastq_file_path};
+    ostream << generate_fastq_file(entries_size);
+    ostream.close();
+    return fastq_file;
+}
+
+// ============================================================================
+// write 3-line entry to stream as often as possible
```
smehringer, line 76:
```suggestion
// seqan3 fastq output benchmark - write 3-line entry to stream as often as possible
```

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,242 @@
...
+        for (auto && [read_sequence, read_id, read_quality] : fin)
+        {
+            sequence = std::move(read_sequence);
+            id = std::move(read_id);
+            quality = std::move(read_quality);
+        }
+
+        sequence.clear();
+        id.clear();
+        quality.clear();
```
smehringer, line 124:
```suggestion
        seqan3::sequence_file_input fin{istream, seqan3::format_fastq{}};
        auto it = fin.begin();
        for (size_t i = 0; i < iterations_per_run; ++i)
            it++;
```
This is how the fasta benchmark does it and I think the comparison to seqan2 is fairer this way because you are doing an additional move and clear, although the record already contains the data (and is cleared). 
In a real use case you would not store the data separately but work directly on the record no?

test/performance/io/format_fastq_benchmark.cpp
```diff
@@ -0,0 +1,242 @@
...
+    ostream.close();
+
+    seqan::StringSet<seqan::String<char>> ids{};
+    seqan::StringSet<seqan::String<seqan::Dna5>> seqs{};
+    seqan::StringSet<seqan::String<char>> quals{};
+
+    for (auto _ : state)
+    {
+        seqan::SeqFileIn seqFileIn(tmp_path.c_str());
+        seqan::readRecords(ids, seqs, quals, seqFileIn);
```
smehringer, line 223: Why does the reading-from-stream-benhcmark is done with single `readRecord` calls and reading-from-disc is done with chunk-reading  `readRecords`? Is this a fair comparison compared to reading line by line with seqan3?

test/performance/io/format_fastq_benchmark.cpp
rrahn: Looks good. I was merely trying to fix the original benchmark. Though the fasta does it the same way. But this version should be more stable wirh regard to the format etc.

test/performance/io/format_fastq_benchmark.cpp
rrahn: Well, I don't know why this was added but not for the others. but the missing write tests can be added later when we work on I/O again

test/performance/io/format_fastq_benchmark.cpp
rrahn: No I think in a real scenario you are storing the sequences in a container. Which means the important comparison would be to store the data in a container and clearly you would then move it here. So this would be closer to a real scenario. But maybe we are not interested in the stuff around (can be part of a macro benchmark) and only try to measure the actual operation necessary. So I make a final commit to adapt the benchmark to the fasta benchmark.

test/performance/io/format_fastq_benchmark.cpp
rrahn: Dunno

rrahn: The messaage was: `the PR is good as is just needs some more love`. But now everything has changed anyway. So I adapt this as well.

smehringer:
Do you want to clean your commit history so we don't loose @Clemapfel 's commit?

PR #1857, opened 2020-05-28
[INFRA,FIX] Increase ccache size in travis

eseiler:
This allows the coverage build to be cached
ccache

The main time sinks are now running the tests in debug mode (coverage) and running the benchmarks (performance).

Going from xenial (16.04) to bionic (18.04) also upgrades the ccache package, which is quite an old version on xenial.
The default cache size for ccache is 5G on newer versions, the default for the old version is 500M.
The coverage build needs around 5.1G (~50G HDD available on travis).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 version: ~> 1.0
 os: linux
-dist: xenial
+dist: bionic
 language: cpp
 
 git:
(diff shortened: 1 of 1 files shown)


eseiler:
Coverage seems unrelated

rrahn:
Haha that means we need enable ccach on jenkins as well. Otherwise this becomes the bottleneck of the CI. Since we have some other jenkins related issues, I would plan this for the next iteration.

PR #1859, opened 2020-05-28
[FEATURE] Argument Parser: new member function that checks if an option was set.

### LINKED ISSUE #1741: [Argument Parser] Make list of options available that were set by the user
### Situation
I follow a concept that makes program settings based on several sources, including:

- Command line arguments
- Config file
- Other input files (in my case, run info files of Illumina)

### Problem
As far as I understand, there is no easy access to check whether a parameter was set via the command line or not. The only was to find out is to compare to the default value after parsing.
However, this approach would not work if the user should be allowed to reset the default value that would be changed from an external resource (e.g., the Illumina run info file).

### Example
The file doesn't really look like this. However, I hope you get what I mean.

Default read length: 150bp

```
RunInfo.xml
<read_length>250bp</read_length>
```

The user wants to reset the value to 150bp via the command line.  
There is currently no (simple) way to detect whether the internal value of 150 for the respective field comes from the command line or is simply the default - the program therefore can't decide whether to use the value from the RunInfo.xml or not.

### Solution
It would be nice to have a function like `bool argument_parser::option_is_set(std::string long_id)` (and the same for `short_id`) that returns `true` if the option was set by the user (can only be called after parsing).  
Further, it would be convenient if the user could explicitly reset the default without knowing about its value with a `default` keyword, regardless of the option type (`--read-length default`).

smehringer:
Fixes #1741 

Current look:

```cpp
{
    int i{5}; // default
    seqan3::argument_parser parser{"test_parser", argc, argv};
    // parser.add_option(i, 'f', "foo", "description");
    // parser.parse(); 
    // return parser.option_overwritten('f');
    // return parser.default_overwritten('f');
    // return parser.default_overwridden('f');
    // return parser.overwritten_by_user('f');
    // return parser.overwritten_on_cmd('f');
    // return parser.option_is_set('f'); // tloka's fav
    return parser.is_option_set('f');
    // return parser.option_is_set_by_user('f');
    // return parser.option_is_set_on_cmd('f');
    // return parser.set_by_user('f');
}
```

```cpp
{
    std::optional<int> i{std::nullopt}; // default

    seqan3::argument_parser parser{"test_parser", argc, argv};
    parser.add_option(i, 'f', "foo", "Default is computed.");
    parser.parse(); 

    return i.has_value();
}
```

would return 1 if called with `./test_parser -f 1`
would return 0 if called with `./test_parser -g 1`

There are open design questions:
1. Naming: `option_was_set`, `option_is_set`, `is_set`, `was_set`, `option_is_overwritten` ... suggestions?
2. By design, this function **could** be called without the necessity of
  (a) a respective option was added (`parser.add_option('f', ...)`)
  (b) `parser.parse()` was called and all options values were parsed
  Do we want to allow this? We can easily introduce these semantic requirements. 

**Resolution**: `parser.is_option_set`

The function will return an enum that can be implicitly cast to bool.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,9 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 
 * We expanded the `seqan3::output_file_validator`, with a parameter `seqan3::output_file_open_options` to allow overwriting
   output files ([\#2009](https://github.com/seqan/seqan3/pull/2009)).
+* The `seqan3::argument_parser` has a new member function `seqan3::argument_parser::is_option_set` that
+  checks whether an option, identified by its long or short name, was set on the command line by the user
+  ([\#1859](https://github.com/seqan/seqan3/pull/1859)).
 
 #### Search
 
--- a/include/seqan3/argument_parser/argument_parser.hpp
+++ b/include/seqan3/argument_parser/argument_parser.hpp
@@ -566,6 +566,25 @@ private:
     //!\brief Stores the sub-parser names in case \link subcommand_arg_parse subcommand parsing \endlink is enabled.
     std::vector<std::string> subcommands{};
 
+    /*!\brief The format of the argument parser that decides the behavior when
+     *        calling the seqan3::argument_parser::parse function.
+     *
+     * \details
+     *
+     * The format is set in the function argument_parser::init.
+     */
+    std::variant<detail::format_parse,
+                 detail::format_help,
+                 detail::format_short_help,
+                 detail::format_version,
+                 detail::format_html,
+                 detail::format_man,
+                 detail::format_copyright/*,
+                 detail::format_ctd*/> format{detail::format_help{{}, false}}; // Will be overwritten in any case.
+
+    //!\brief List of option/flag identifiers that are already used.
+    std::set<std::string> used_option_ids{"h", "hh", "help", "advanced-help", "export-help", "version", "copyright"};
+
     /*!\brief Initializes the seqan3::argument_parser class on construction.
      *
      * \param[in] argc        The number of command line arguments.
@@ -764,22 +783,6 @@ private:
         if (detail::format_parse::is_empty_id(short_id) && detail::format_parse::is_empty_id(long_id))
             throw design_error("Option Identifiers cannot both be empty.");
     }
-
-    /*!\brief The format of the argument parser that decides the behavior when
-     * calling the seqan3::argument_parser::parse function.
-     * \details The format is set in the function argument_parser::init.
-     */
-    std::variant<detail::format_parse,
-                 detail::format_help,
-                 detail::format_short_help,
-                 detail::format_version,
-                 detail::format_html,
-                 detail::format_man,
-                 detail::format_copyright/*,
-                 detail::format_ctd*/> format{detail::format_help{{}, false}}; // Will be overwritten in any case.
-
-    //!\brief List of option/flag identifiers that are already used.
-    std::set<std::string> used_option_ids{"h", "hh", "help", "advanced-help", "export-help", "version", "copyright"};
 };
 
 } // namespace seqan3
--- a/include/seqan3/argument_parser/argument_parser.hpp
+++ b/include/seqan3/argument_parser/argument_parser.hpp
@@ -585,6 +585,9 @@ private:
     //!\brief List of option/flag identifiers that are already used.
     std::set<std::string> used_option_ids{"h", "hh", "help", "advanced-help", "export-help", "version", "copyright"};
 
+    //!\brief The command line arguments.
+    std::vector<std::string> cmd_arguments{};
+
     /*!\brief Initializes the seqan3::argument_parser class on construction.
      *
      * \param[in] argc        The number of command line arguments.
@@ -598,7 +601,7 @@ private:
      *
      * \details
      *
-     * This function adds all command line parameters to a std::vector<std::string>
+     * This function adds all command line parameters to the cmd_arguments member variable
      * to take advantage of the vector functionality later on. Additionally,
      * the format member variable is set, depending on which parameters are given
      * by the user:
@@ -619,9 +622,6 @@ private:
      */
     void init(int argc, char const * const * const argv)
     {
-        // cash command line input, in case --version-check is specified but shall not be passed to format_parse()
-        std::vector<std::string> argv_new{};
-
         if (argc <= 1) // no arguments provided
         {
             format = detail::format_short_help{};
@@ -708,18 +708,17 @@ private:
                 else
                     throw validation_error{"Value for option --version-check must be 1 or 0."};
 
+                // in case --version-check is specified it shall not be passed to format_parse()
                 argc -= 2;
             }
             else
             {
-                argv_new.push_back(std::move(arg));
+                cmd_arguments.push_back(std::move(arg));
             }
         }
 
         if (!special_format_was_set)
-        {
-            format = detail::format_parse(argc, std::move(argv_new));
-        }
+            format = detail::format_parse(argc, cmd_arguments);
     }
 
     //!\brief Adds standard options to the help page.
--- a/include/seqan3/argument_parser/argument_parser.hpp
+++ b/include/seqan3/argument_parser/argument_parser.hpp
@@ -425,6 +425,64 @@ public:
         return *sub_parser;
     }
 
+    /*!\brief Checks whether the option identifier (`id`) was set on the command line by the user.
+     * \tparam id_type Either type `char` or a type that a `std::string` is constructible from.
+     * \param[in] id The short (`char`) or long (`std::string`) option identifier to search for.
+     * \returns `true` if option identifier `id` was set on the command line by the user.
+     * \throws seqan3::design_error if the function is used incorrectly (see details below).
+     *
+     * \details
+     *
+     * You can only ask for option identifiers that were added to the parser beforehand via
+     * `seqan3::argument_parser::add_option`.
+     * As in the `seqan3::argument_parser::add_option` call, pass short identifiers as a `char` and long identifiers
+     * as a `std::string` or a type that a `std::string` is constructible from (e.g. a `const char *`).
+     *
+     * ### Example
+     *
+     * \include test/snippet/argument_parser/is_option_set.cpp
+     *
+     * ### Exceptions
+     *
+     * This function throws a seqan3::design_error if
+     * * `seqan3::argument_parser::parse()` was not called before.
+     * * a long identifier was passed (e.g. a `std::string`) that only consists of a single character. If you mean to
+     *   pass a short identifier, please pass it as a `char` not a `std::string`.
+     * * the option identifier cannot be found in the list of valid option identifiers that were added to the parser
+     *   via `seqan3::argument_parser::add_option()` calls beforehand.
+     */
+    template <typename id_type>
+    //!\cond
+        requires std::same_as<id_type, char> || std::constructible_from<std::string, id_type>
+    //!\endcond
+    bool is_option_set(id_type const & id) const
+    {
+        if (!parse_was_called)
+            throw design_error{"You can only ask which options have been set after calling the function `parse()`."};
+
+        // the detail::format_parse::find_option_id call in the end expects either a char or std::string
+        using char_or_string_t = std::conditional_t<std::same_as<id_type, char>, char, std::string>;
+        char_or_string_t short_or_long_id = {id}; // e.g. convert char * to string here if necessary
+
+        if constexpr (!std::same_as<id_type, char>) // long id was given
+        {
+            if (short_or_long_id.size() == 1)
+            {
+                throw design_error{"Long option identifiers must be longer than one character! If " + short_or_long_id +
+                                   "' was meant to be a short identifier, please pass it as a char ('') not a string"
+                                   " (\"\")!"};
+            }
+        }
+
+        if (std::find(used_option_ids.begin(), used_option_ids.end(), std::string{id}) == used_option_ids.end())
+            throw design_error{"You can only ask for option identifiers that you added with add_option() before."};
+
+        // we only need to search for an option before the `end_of_options_indentifier` (`--`)
+        auto end_of_options = std::find(cmd_arguments.begin(), cmd_arguments.end(), end_of_options_indentifier);
+        auto option_it = detail::format_parse::find_option_id(cmd_arguments.begin(), end_of_options, short_or_long_id);
+        return option_it != end_of_options;
+    }
+
     //!\name Structuring the Help Page
     //!\{
 
@@ -559,6 +617,9 @@ private:
 
     //!\brief Validates the application name to ensure an escaped server call.
     std::regex app_name_regex{"^[a-zA-Z0-9_-]+$"};
+
+    //!\brief Signals the argument parser that no options follow this string but only positional arguments.
+    static constexpr std::string_view const end_of_options_indentifier{"--"};
 
     //!\brief Stores the sub-parser in case \link subcommand_arg_parse subcommand parsing \endlink is enabled.
     std::unique_ptr<argument_parser> sub_parser{nullptr};
--- a/include/seqan3/argument_parser/detail/format_parse.hpp
+++ b/include/seqan3/argument_parser/detail/format_parse.hpp
@@ -160,14 +160,14 @@ public:
     }
 
     /*!\brief Finds the position of a short/long identifier in format_parse::argv.
+     * \tparam iterator_type The type of iterator that defines the range to search in.
      * \tparam id_type The identifier type; must be either of type `char` if it denotes a short identifier or
      *                 std::string if it denotes a long identifier.
      * \param[in] begin_it The iterator where to start the search of the identifier.
-     *                     Note that the end iterator is kept as a member variable.
-     * \param[in] id       The identifier to search for (must not contain dashes).
-     * \returns An iterator pointing to the first occurrence of the identifier in
-     *          the list pointed to by begin_t. If the list does not contain the
-     *          identifier `id`, the member variable `end_of_options_it` is returned.
+     * \param[in] end_it The iterator one past the end of where to search the identifier.
+     * \param[in] id The identifier to search for (must not contain dashes).
+     * \returns An iterator pointing to the first occurrence of `id` in the list pointed to by `begin_it`
+     *          or `end_it` if it is not contained.
      *
      * \details
      *
@@ -180,13 +180,13 @@ public:
      * The `id` is found by comparing every argument in argv to `id` prepended with two dashes (`--`)
      * or a prefix of such followed by the equal sign `=`.
      */
-    template <typename id_type>
-    std::vector<std::string>::iterator find_option_id(std::vector<std::string>::iterator const begin_it, id_type const & id)
+    template <typename iterator_type, typename id_type>
+    static iterator_type find_option_id(iterator_type begin_it, iterator_type end_it, id_type const & id)
     {
         if (is_empty_id(id))
-            return end_of_options_it;
+            return end_it;
 
-        return (std::find_if(begin_it, end_of_options_it,
+        return (std::find_if(begin_it, end_it,
             [&] (std::string const & current_arg)
             {
                 std::string full_id = prepend_dash(id);
@@ -528,12 +528,12 @@ private:
     template <typename option_type, typename id_type>
     bool get_option_by_id(option_type & value, id_type const & id)
     {
-        auto it = find_option_id(argv.begin(), id);
+        auto it = find_option_id(argv.begin(), end_of_options_it, id);
 
         if (it != end_of_options_it)
             identify_and_retrieve_option_value(value, it, id);
 
-        if (find_option_id(it, id) != end_of_options_it) // should not be found again
+        if (find_option_id(it, end_of_options_it, id) != end_of_options_it) // should not be found again
            throw option_declared_multiple_times("Option " + prepend_dash(id) +
                                                 " is no list/container but declared multiple times.");
 
@@ -557,13 +557,13 @@ private:
     //!\endcond
     bool get_option_by_id(option_type & value, id_type const & id)
     {
-        auto it = find_option_id(argv.begin(), id);
+        auto it = find_option_id(argv.begin(), end_of_options_it, id);
         bool seen_at_least_once{it != end_of_options_it};
 
         while (it != end_of_options_it)
         {
             identify_and_retrieve_option_value(value, it, id);
-            it = find_option_id(it, id);
+            it = find_option_id(it, end_of_options_it, id);
         }
 
         return seen_at_least_once;
--- a/include/seqan3/argument_parser/detail/format_parse.hpp
+++ b/include/seqan3/argument_parser/detail/format_parse.hpp
@@ -219,7 +219,7 @@ private:
     * \param[in] long_id The name of the long identifier.
     * \returns The input long name prepended with a double dash.
     */
-    std::string prepend_dash(std::string const & long_id)
+    static std::string prepend_dash(std::string const & long_id)
     {
         return ("--" + long_id);
     }
@@ -228,7 +228,7 @@ private:
     * \param[in] short_id The name of the short identifier.
     * \returns The input short name prepended with a single dash.
     */
-    std::string prepend_dash(char const short_id)
+    static std::string prepend_dash(char const short_id)
     {
         return ("-" + std::string(1, short_id));
     }
--- a/include/seqan3/argument_parser/detail/format_parse.hpp
+++ b/include/seqan3/argument_parser/detail/format_parse.hpp
@@ -66,7 +66,7 @@ public:
      * \param[in] argc_ The number of command line arguments.
      * \param[in] argv_ The command line arguments to parse.
      */
-    format_parse(int const argc_, std::vector<std::string> && argv_) :
+    format_parse(int const argc_, std::vector<std::string> argv_) :
         argc{argc_ - 1}, argv{std::move(argv_)}
     {}
     //!\}
--- a/include/seqan3/argument_parser/detail/format_parse.hpp
+++ b/include/seqan3/argument_parser/detail/format_parse.hpp
@@ -221,7 +221,7 @@ private:
     */
     static std::string prepend_dash(std::string const & long_id)
     {
-        return ("--" + long_id);
+        return {"--" + long_id};
     }
 
     /*!\brief Appends a double dash to a short identifier and returns it.
@@ -230,7 +230,7 @@ private:
     */
     static std::string prepend_dash(char const short_id)
     {
-        return ("-" + std::string(1, short_id));
+        return {"-" + std::string{short_id}};
     }
 
     /*!\brief Returns "-[short_id]/--[long_id]" if both are non-empty or just one of them if the other is empty.
--- a/include/seqan3/argument_parser/detail/format_parse.hpp
+++ b/include/seqan3/argument_parser/detail/format_parse.hpp
@@ -159,48 +159,6 @@ public:
             return is_char<'\0'>(id);
     }
 
-private:
-    //!\brief Describes the result of parsing the user input string given the respective option value type.
-    enum class option_parse_result
-    {
-        success, //!< Parsing of user input was successful.
-        error, //!< There was some error while trying to parse the user input.
-        overflow_error //!< Parsing was successful but the arithmetic value would cause an overflow.
-    };
-
-    /*!\brief Appends a double dash to a long identifier and returns it.
-    * \param[in] long_id The name of the long identifier.
-    * \returns The input long name prepended with a double dash.
-    */
-    std::string prepend_dash(std::string const & long_id)
-    {
-        return ("--" + long_id);
-    }
-
-    /*!\brief Appends a double dash to a short identifier and returns it.
-    * \param[in] short_id The name of the short identifier.
-    * \returns The input short name prepended with a single dash.
-    */
-    std::string prepend_dash(char const short_id)
-    {
-        return ("-" + std::string(1, short_id));
-    }
-
-    /*!\brief Returns "-[short_id]/--[long_id]" if both are non-empty or just one of them if the other is empty.
-    * \param[in] short_id The name of the short identifier.
-    * \param[in] long_id  The name of the long identifier.
-    * \returns The short_id prepended with a single dash and the long_id prepended with a double dash, separated by '/'.
-    */
-    std::string combine_option_names(char const short_id, std::string const & long_id)
-    {
-        if (short_id == '\0')
-            return prepend_dash(long_id);
-        else if (long_id.empty())
-            return prepend_dash(short_id);
-        else // both are set (note: both cannot be empty, this is caught before)
-            return prepend_dash(short_id) + "/" + prepend_dash(long_id);
-    }
-
     /*!\brief Finds the position of a short/long identifier in format_parse::argv.
      * \tparam id_type The identifier type; must be either of type `char` if it denotes a short identifier or
      *                 std::string if it denotes a long identifier.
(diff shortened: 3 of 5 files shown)


include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -415,6 +415,24 @@ class argument_parser
...
+        return detail::format_parse::find_option_id(cmd_arguments.begin(), end, id) != end;
+    }
+
+    // allows all types that are implicitly convertible to std::string
+    //!\overload
+    bool option_was_set(std::string const & id)
+    {
+        auto end = std::find(cmd_arguments.begin(), cmd_arguments.end(), "--");
+        return detail::format_parse::find_option_id(cmd_arguments.begin(), end, id) != end;
+    }
```
eseiler, line 434:
```suggestion
    bool option_was_set(char const id)
    {
        auto end = std::find_end(cmd_arguments.begin(), cmd_arguments.end(), "-");
        return detail::format_parse::find_option_id(cmd_arguments.begin(), end, id) != end;
    }

    // allows all types that are implicitly convertible to std::string
    //!\overload
    bool option_was_set(std::string const & id)
    {
        auto end = std::find_end(cmd_arguments.begin(), cmd_arguments.end(), "--");
        return detail::format_parse::find_option_id(cmd_arguments.begin(), end, id) != end;
    }
```

I think in your test case `{"./argument_parser_test", "-l", "hallo", "--foobar", "ballo"}` this works, because it always ends at --foobar and encloses all args.

rrahn:
### Resolution 24.06.2020

We have a preference for the member function design.
The OP of the issue should be asked for feedback on this as well, to come to a final solution.

include/seqan3/argument_parser/argument_parser.hpp
smehringer: No this is something different.
I actually do want to search for the first occurrence of "--". Note: Not as a prefix (or other -fix) of another word but ONLY `"--"` because it signals the end of options and marks the beginning of arguments for positional options. It is special in its meaning. And since I am searching for option names I must only search before the arguments for positional options.
Was that explained in an understandable way? Otherwise I'll try again.

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -163,11 +163,10 @@ class format_parse : public format_base
...
      * \param[in] begin_it The iterator where to start the search of the identifier.
-     *                     Note that the end iterator is kept as a member variable.
-     * \param[in] id       The identifier to search for (must not contain dashes).
-     * \returns An iterator pointing to the first occurrence of the identifier in
-     *          the list pointed to by begin_t. If the list does not contain the
-     *          identifier `id`, the member variable `end_of_options_it` is returned.
+     * \param[in] end_it The iterator where to end the search of the identifier.
+     * \param[in] id The identifier to search for (must not contain dashes).
+     * \returns An iterator pointing to the first occurrence of `id` in the list pointed to by `begin_it`
+     *          or `end` if it is not contained.
```
eseiler, line 169:
```suggestion
     *          or `end_it` if it is not contained.
```

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -163,11 +163,10 @@ class format_parse : public format_base
      * \tparam id_type The identifier type; must be either of type `char` if it denotes a short identifier or
      *                 std::string if it denotes a long identifier.
      * \param[in] begin_it The iterator where to start the search of the identifier.
-     *                     Note that the end iterator is kept as a member variable.
-     * \param[in] id       The identifier to search for (must not contain dashes).
-     * \returns An iterator pointing to the first occurrence of the identifier in
-     *          the list pointed to by begin_t. If the list does not contain the
-     *          identifier `id`, the member variable `end_of_options_it` is returned.
+     * \param[in] end_it The iterator where to end the search of the identifier.
```
eseiler, line 166: Isn't this "one past where you want the search to end"

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -66,7 +66,7 @@ class format_parse : public format_base
      * \param[in] argc_ The number of command line arguments.
      * \param[in] argv_ The command line arguments to parse.
      */
-    format_parse(int const argc_, std::vector<std::string> && argv_) :
+    format_parse(int const argc_, std::vector<std::string> argv_) :
```
eseiler, line 69: Why copy?

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -416,6 +416,24 @@ class argument_parser
...
     }
 
+    /*!\brief Checks whether the option distinguished by `id` was set on the command line by the user.
+     * \param[in] id The option identifier to search for (must not contain dashes!).
+     * \returns `true` if option identifier `id` was set on the command line by the user.
+     */
+    bool is_option_set(char const id)
+    {
+        auto end = std::find(cmd_arguments.begin(), cmd_arguments.end(), "--");
+        return detail::format_parse::find_option_id(cmd_arguments.begin(), end, id) != end;
```
eseiler, line 426: I'm having trouble understanding what you are doing.

1. Why search for `--` when you are looking at short ids?
2. Wouldn't you want to use `std::find_end` since you surely don't want `end` to point to the first occurrence of `--`?
3. Why don't you just pass begin and end of cmd_arguments?

test/unit/argument_parser/format_parse_test.cpp
```diff
@@ -969,3 +969,16 @@ TEST(parse_test, issue1544)
...
     }
 }
+
+TEST(parse_test, is_option_set)
+{
+    const char * argv[] = {"./argument_parser_test", "-l", "hallo", "--foobar", "ballo", "--", "--loo"};
+    seqan3::argument_parser parser{"test_parser", 5, argv, false};
+
+    EXPECT_TRUE(parser.is_option_set('l'));
+    EXPECT_TRUE(parser.is_option_set("foobar"));
```
eseiler, line 979: Related to my comments for `is_option_set`, I think there are a few cases missing:

1. What if I add another short option that is the last option, or in between?
2. Same for long options.

test/unit/argument_parser/format_parse_test.cpp
```diff
@@ -969,3 +969,16 @@ TEST(parse_test, issue1544)
...
+
+    EXPECT_FALSE(parser.is_option_set("foo"));
+    EXPECT_FALSE(parser.is_option_set('f'));
+    EXPECT_FALSE(parser.is_option_set("loo"));
```
eseiler, line 983: I would say that `loo` is set?
It's in the `argv` after all...

eseiler (changes requested):

include/seqan3/argument_parser/argument_parser.hpp
eseiler: I just noticed that you answered yesterday under my month old comment, maybe you can add the reasoning for the `--` as a comment, because it is really confusing.
Also, where does the `--` come from?

smehringer: > No this is something different.
I actually do want to search for the first occurrence of "--". Note: Not as a prefix (or other -fix) of another word but ONLY "--" because it signals the end of options and marks the beginning of arguments for positional options. It is special in its meaning. And since I am searching for option names I must only search before the arguments for positional options.
Was that explained in an understandable way? Otherwise I'll try again.

`--` is a POSIX convention to singal that there are no more option names after this. That way the user can use dash in arguments without confusing the argument parser. This done like this in most argument parsers.

See [POSIX conventions](http://www.iitk.ac.in/esc101/05Aug/tutorial/essential/attributes/_posix.html)

test/unit/argument_parser/format_parse_test.cpp
smehringer: See explanation above.

include/seqan3/argument_parser/detail/format_parse.hpp
smehringer:
```suggestion
     * \param[in] end_it The iterator one past the end of where to search the identifier.
```

test/unit/argument_parser/format_parse_test.cpp
smehringer: I don't exactly know what you mean.. can you give a command line example?

include/seqan3/argument_parser/argument_parser.hpp
eseiler: Do you add the `--` somewhere or is it magically there?

test/unit/argument_parser/format_parse_test.cpp
eseiler: It's ok now after explaining the `--` thing

include/seqan3/argument_parser/argument_parser.hpp
smehringer: It is user input on the command line that may be there

include/seqan3/argument_parser/detail/format_parse.hpp
smehringer: this change, in commit fe5ec6628f7cb34f19aa87a24809a26c08d3de71, is part of storing the command line arguments in the argument parser (before I did not need them and forwarded them to `format_parse` but now I need them to enable searching if an option was set).
Since I store them in a variable, I don't pass a forwarded rvalue to `format_parse` anymore (see [argument_parser.hpp:709](https://github.com/seqan/seqan3/pull/1859/commits/fe5ec6628f7cb34f19aa87a24809a26c08d3de71#diff-18eb4ceee3f51232eaef3efff94f6f15R709)) but instead I pass an lvalue. Since format_parse is moving `argv_` into its member `argv` in the member initialisation below, I need a copy here.

smehringer:
I hope I have answered all your questions. If I missed a requested change please tell me again :)

include/seqan3/argument_parser/argument_parser.hpp
smehringer: Phew maybe in the tutorial. I thought this is common knowledge somehow. 
I will make a follow up issue to make sure the parsing conventions are mentioned somewhere. It oculd just be I said somewhere "we follow official POSIX conventions".

CHANGELOG.md
```diff
@@ -52,6 +52,8 @@ Note that 3.1.0 will be the first API stable release and interfaces in this rele
   * `seqan3::argument_parser::add_line`
   * `seqan3::argument_parser::add_list_item`
   Note that other `seqan3::argument_parser::option_spec`s like `REQUIRED` are ignored.
+* The `seqan3::argument_parser` has a new member function `seqan3::argument_parser::is_option_set` that
```
eseiler, line 55: This goes into master and hence should 3.0.3

eseiler (approved):
One small thing for the changelog, otherwise lgtm

include/seqan3/argument_parser/argument_parser.hpp
smehringer: Here is the issue https://github.com/seqan/product_backlog/issues/201

include/seqan3/argument_parser/detail/format_parse.hpp
rrahn: Here an explanation why you should prefer copy and move over const-ref and copy. The advantage is that rvalues will never be copied but only moved (not if you capture them by const-ref and then copy inside the function). lvalues need to be copied at least once anyway and from the performance perspective it doesn't matter if it happens inside of the function or in the function parameter. One additional advantage of copying in the function parameter is that it is easier to write a strong exception safe function since the body of the function is not executed if something throws during that copy.

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -416,6 +416,24 @@ class argument_parser
         return *sub_parser;
     }
 
+    /*!\brief Checks whether the option distinguished by `id` was set on the command line by the user.
+     * \param[in] id The option identifier to search for (must not contain dashes!).
+     * \returns `true` if option identifier `id` was set on the command line by the user.
+     */
+    bool is_option_set(char const id)
+    {
+        auto end = std::find(cmd_arguments.begin(), cmd_arguments.end(), "--");
```
rrahn, line 425:
```suggestion
        auto end_of_options = std::find(cmd_arguments.begin(), cmd_arguments.end(), "--");
```

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -416,6 +416,24 @@ class argument_parser
...
+        return detail::format_parse::find_option_id(cmd_arguments.begin(), end, id) != end;
+    }
+
+    // allows all types that are implicitly convertible to std::string
+    //!\overload
+    bool is_option_set(std::string const & id)
+    {
+        auto end = std::find(cmd_arguments.begin(), cmd_arguments.end(), "--");
```
rrahn, line 433:
```suggestion
        auto end_of_options = std::find(cmd_arguments.begin(), cmd_arguments.end(), "--");
```

include/seqan3/argument_parser/argument_parser.hpp
rrahn: And while looking at it. You should make `--` a static constexpr variable somewhere in the class's scope such that it also gets a proper name: `end_of_options_marker` or something like that.

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -554,6 +572,25 @@ class argument_parser
     //!\brief Stores the sub-parser in case \link subcommand_arg_parse subcommand parsing \endlink is enabled.
     std::unique_ptr<argument_parser> sub_parser{nullptr};
 
+    /*!\brief The format of the argument parser that decides the behavior when
+     * calling the seqan3::argument_parser::parse function.
```
rrahn, line 576:
```suggestion
    /*!\brief The format of the argument parser that decides the behavior when calling the seqan3::argument_parser::parse function.
```
or
```suggestion
    /*!\brief The format of the argument parser that decides the behavior when
     *        calling the seqan3::argument_parser::parse function.
```

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -554,6 +572,25 @@ class argument_parser
...
+     * \details The format is set in the function argument_parser::init.
```
rrahn, line 577:
```suggestion
     * \details 
     *
     * The format is set in the function argument_parser::init.
```

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -554,6 +572,25 @@ class argument_parser
...
+     */
+    std::variant<detail::format_parse,
```
rrahn, line 579: Is this the only place where the variant type is defined?

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -554,6 +572,25 @@ class argument_parser
...
+                 detail::format_help,
+                 detail::format_short_help,
+                 detail::format_version,
+                 detail::format_html,
+                 detail::format_man,
+                 detail::format_copyright/*,
+                 detail::format_ctd*/> format{detail::format_help{{}, false}}; // Will be overwritten in any case.
+
+    //!\brief List of option/flag identifiers that are already used.
+    std::set<std::string> used_option_ids{"h", "hh", "help", "advanced-help", "export-help", "version", "copyright"};
```
rrahn, line 589: also this is the only place where these entities are set/used?

include/seqan3/argument_parser/argument_parser.hpp
rrahn: some more thoughts about that. If I understand this correctly, then we have to call parse before we can ask whether an option was set by the user. So should it not throw in the case parse wasn't called before? I think we have a variable that caches this?

rrahn: The other thing I'd like to reconsider from the implementation is the duplication of this code. 
First, I observed that an extra copy is made to store the parsed arguments in the argument parser. 
Second, the arguments are only necessary if format_parse was called so also this option can only be available if format_parse was called.

So the code should be refactored that I can ask for the set option in the format_parse class. Calling it for any other parsing option can then throw, right? So since all formats inherit from a base class this can be achieved easily. Then only the format_parse class overloads this and inside of the argument_parser you call std::visit on the selected format.

rrahn (changes requested):
Ok, I have some general considerations about the implementation of this feature. Please find my comments below.

CHANGELOG.md
```diff
@@ -24,6 +24,10 @@ The following API changes should be documented as such:
 If possible, provide tooling that performs the changes, e.g. a shell-script.
 -->
 
+# 3.0.3
+
+Note that 3.1.0 will be the first API stable release and interfaces in this release might still change.
```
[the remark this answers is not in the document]
smehringer: Whops. So the argument parser feature was supposed to be under this line directly. got mixed up.
But the line itself was added to 3.0.1 and 3.0.2 so I added it to 3.0.3 too.

include/seqan3/argument_parser/argument_parser.hpp
smehringer: Yes

include/seqan3/argument_parser/argument_parser.hpp
smehringer: yes

include/seqan3/argument_parser/argument_parser.hpp
smehringer: > some more thoughts about that. If I understand this correctly, then we have to call parse before we can ask whether an option was set by the user. So should it not throw in the case parse wasn't called before? I think we have a variable that caches this?

No, we do not **have to** call parse before checking this. It is independent. (see test case)
This is the case because we have the list `argv` already on construction will all the information we need.
We can of course enforce this restriction anyway if we think it is confusing for the user or more intuitive :shrug:.

> The other thing I'd like to reconsider from the implementation is the duplication of this code.

Which duplication?

> First, I observed that an extra copy is made to store the parsed arguments in the argument parser.

Yes. We now have to copies of `argv`. One in the `argument_parser` and one in `format_parse`.
But this is needed either way, even if I move this list into `format_parse` because format_parse is stripping it's copy of `argv` of arguments while parsing the information s.t. it can afterwards detect "too many" or "unknown option identifiers". So in order to enable `is_option_set` I need to store the original `argv` in some place anyway.

> Second, the arguments are only necessary if format_parse was called so also this option can only be available if format_parse was called.

Technically no, see above.

> So the code should be refactored that I can ask for the set option in the format_parse class. Calling it for any other parsing option can then throw, right? So since all formats inherit from a base class this can be achieved easily. Then only the format_parse class overloads this and inside of the argument_parser you call std::visit on the selected format.

IMHO I think that `is_option_set` is not strongly attached to `format_parse`. It's a feature of the argument parser that doesn't involve actual parsing of argument values but only checking if an id can be found. Is this such a strong design decision, whether we call a static function of format_parse or use std::visit?

eseiler:
Can you rebase?

smehringer:
Since there are several open discussion points I will wait with the work on this PR.

include/seqan3/argument_parser/argument_parser.hpp
rrahn: > No, we do not have to call parse before checking this. It is independent. (see test case)
> IMHO I think that is_option_set is not strongly attached to format_parse. It's a feature of the argument parser that doesn't involve actual parsing of argument values but only checking if an id can be found. Is this such a strong design decision, whether we call a static function of format_parse or use std::visit?

But the design doesn't add up in my opinion. If I understand it correctly, then right now the API requires to call parse in order to validate and set the parameters with the options from the command line or to print the help page. So the design deliberately forces us to call parse after we added the options to the parser. Everything is deferred. Accordingly, in the proper design the check if an option was set should be put after the parse method was called. The reason is that the interface should only work for valid options and throw if you try to ask for an option that wasn't added to the parser. But this can only be known after the option was added, and parse functions here as the terminator. So that's why I think that the current solution does not reflect the design of the argument parser.

rrahn (changes requested):
I am still a little bit unhappy with the design. This is merely because I believe that the argument parser design might not be reflected properly. I tried to explain my concerns in the comments. Please feel free to put it into core discussion if you have a different opinion or if I did not clear things up. Thank you!

test/unit/argument_parser/format_parse_test.cpp
```diff
@@ -969,3 +969,16 @@ TEST(parse_test, issue1544)
...
+
+TEST(parse_test, is_option_set)
+{
+    const char * argv[] = {"./argument_parser_test", "-l", "hallo", "--foobar", "ballo", "--", "--loo"};
+    seqan3::argument_parser parser{"test_parser", 5, argv, false};
+
+    EXPECT_TRUE(parser.is_option_set('l'));
+    EXPECT_TRUE(parser.is_option_set("foobar"));
+
+    EXPECT_FALSE(parser.is_option_set("foo"));
```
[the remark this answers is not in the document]
smehringer: see below

smehringer:
@rrahn polite ping

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -416,6 +416,49 @@ class argument_parser
         return *sub_parser;
     }
 
+    /*!\brief Checks whether the option distinguished by `id` was set on the command line by the user.
+     * \param[in] id The option identifier to search for (must not contain dashes!).
+     * \returns `true` if option identifier `id` was set on the command line by the user.
+     */
```
rrahn, line 422: The documentation is part of the public API and should be a little bit more verbose. For example it is not clear that the function might throw. Second the documentation says that id should not have dashes. But the overloaded function would get the same information, while it is not true in this case.

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -416,6 +416,49 @@ class argument_parser
...
+        std::for_each(id.begin(), id.end(), [&allowed] (char c)
+        {
+            if (!allowed(c))
+                throw design_error("Long identifiers may only contain alphanumeric characters, '_', '-', or '@'.");
+        });
+
+        // we only need to search for an option before the `end_of_options_indentifier` (`--`)
+        auto end_of_options = std::find(cmd_arguments.begin(), cmd_arguments.end(), end_of_options_indentifier);
+        return detail::format_parse::find_option_id(cmd_arguments.begin(), end_of_options, id) != end_of_options;
+    }
```
rrahn, line 460: Second, what happens if I call is_option_set("a")? is that really a design error? It would be a user error, right? But it should also just work. And is is_option_set('_') allowed as an option?
In addition, I see a lot of potential to DRY this function. For example, you could write a detail function that get's the option identifier always as a string and in addition the set of valid characters. Then you can only need to handle the really different things in the tpo-level functions.

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -551,9 +594,33 @@ class argument_parser
...
 
+    //!\brief Signals the argument parser that no options follow this string but only positional arguments.
+    static constexpr std::string_view const end_of_options_indentifier{"--"};
+
     //!\brief Stores the sub-parser in case \link subcommand_arg_parse subcommand parsing \endlink is enabled.
     std::unique_ptr<argument_parser> sub_parser{nullptr};
 
+    /*!\brief The format of the argument parser that decides the behavior when
+     *        calling the seqan3::argument_parser::parse function.
+     * \details
```
rrahn, line 605:
```suggestion
    /*!\brief The format of the argument parser that decides the behavior when
     *        calling the seqan3::argument_parser::parse function.
     *
     * \details
```

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -416,6 +416,49 @@ class argument_parser
         return *sub_parser;
     }
 
+    /*!\brief Checks whether the option distinguished by `id` was set on the command line by the user.
+     * \param[in] id The option identifier to search for (must not contain dashes!).
+     * \returns `true` if option identifier `id` was set on the command line by the user.
+     */
+    bool is_option_set(char const id)
```
rrahn, line 423:
```suggestion
    bool is_option_set(char const id) const
```

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -416,6 +416,49 @@ class argument_parser
...
+            throw design_error("Short option identifiers may only contain alphanumeric characters, '_', or '@'.");
+
+        // we only need to search for an option before the `end_of_options_indentifier` (`--`)
+        auto end_of_options = std::find(cmd_arguments.begin(), cmd_arguments.end(), end_of_options_indentifier);
+        return detail::format_parse::find_option_id(cmd_arguments.begin(), end_of_options, id) != end_of_options;
+    }
+
+    // allows all types that are implicitly convertible to std::string
+    //!\overload
+    bool is_option_set(std::string const & id)
```
rrahn, line 440:
```suggestion
    bool is_option_set(std::string const & id) const
```

test/unit/argument_parser/format_parse_test.cpp
```diff
@@ -969,3 +969,32 @@ TEST(parse_test, issue1544)
...
+    EXPECT_FALSE(parser.is_option_set("loo"));
+
+    // errors:
+    EXPECT_THROW(parser.is_option_set(""), seqan3::design_error); // empty option
+    EXPECT_THROW(parser.is_option_set("--foo"), seqan3::design_error); // no dash at the beginning
+    EXPECT_THROW(parser.is_option_set("f"), seqan3::design_error); // long id must be greater than 1 character
+    EXPECT_THROW(parser.is_option_set("f!o"), seqan3::design_error); // no weird character
+
+    EXPECT_THROW(parser.is_option_set('!'), seqan3::design_error); // no weird character
+    EXPECT_THROW(parser.is_option_set('-'), seqan3::design_error); // no weird character
```
rrahn, line 999:
```suggestion
    EXPECT_THROW(parser.is_option_set('-'), seqan3::design_error); // no weird character
    EXPECT_THROW(parser.is_option_set('\0'), seqan3::design_error); // no control character
```

rrahn (changes requested):
Nicely done. Some suggestions to make the code more concise.

include/seqan3/argument_parser/argument_parser.hpp
smehringer: I thought about this again and I propose the following: Since we do enforce that `parse()` is called before this function can be used, I can now just check if the requested option identifier is in the list of `used_option_identifiers`. Those have already been tested in the `add_option` calls beforehand so I can omit all the testing completely?

rrahn: But you specify the option by a string, which could contain invalid characters, aren't you? So you would need to have some checks but you could resue what is already there to test them properly, which is probably the same code as done when adding the options?

smehringer:
Sorry I had to rebase because of conflicts so I rebased all the commits correctly right away

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -425,6 +425,62 @@ class argument_parser
...
+     * seqan3::argument_parser::add_option.
+     *
+     * ### Example:
+     *
+     * \include test/snippet/argument_parser/is_option_set.cpp
+     */
+    bool is_option_set(char const id) const
+    {
+        if (!parse_was_called)
+            throw design_error("You can only asked which options have been set after calling the function `parse()`.");
```
rrahn, line 444:
```suggestion
            throw design_error("You can only ask which options have been set after calling the function `parse()`.");
```

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -425,6 +425,62 @@ class argument_parser
...
+            throw design_error{"Long option identifiers must be longer than one character! If " + id + "' was meant to "
+                               "be a short identifier, please pass it as a char ('') not a string (\"\")!"};
+
+        if (std::find(used_option_ids.begin(), used_option_ids.end(), id) == used_option_ids.end())
+            throw design_error{"You can only ask for option identifiers that you added with add_option() before."};
+
+        // we only need to search for an option before the `end_of_options_indentifier` (`--`)
+        auto end_of_options = std::find(cmd_arguments.begin(), cmd_arguments.end(), end_of_options_indentifier);
+        return detail::format_parse::find_option_id(cmd_arguments.begin(), end_of_options, id) != end_of_options;
+    }
```
rrahn, line 482:
```suggestion
    bool is_option_set(char const id) const
    {
        return is_option_set_impl(std::string{id}, false);
    }
   
    bool is_option_set(std::string const & id) const
    {
         return is_option_set_impl(id, true);
    }

    bool is_option_set_impl(std::string const & id, bool const is_long_option_id) const
    {
        if (!parse_was_called)
            throw design_error{"You can only ask which options have been set after calling the function `parse()`."};

        if (is_long_option_id && id.size() == 1)
            throw design_error{"Long option identifiers must be longer than one character! If " + id + "' was meant to "
                               "be a short identifier, please pass it as a char ('') not a string (\"\")!"};

        if (std::find(used_option_ids.begin(), used_option_ids.end(), id) == used_option_ids.end())
            throw design_error{"You can only ask for option identifiers that you added with add_option() before."};

        // we only need to search for an option before the `end_of_options_indentifier` (`--`)
        auto end_of_options = std::find(cmd_arguments.begin(), cmd_arguments.end(), end_of_options_indentifier);
        return detail::format_parse::find_option_id(cmd_arguments.begin(), end_of_options, id) != end_of_options;
    }
```

rrahn: This also also reduces the typo to only one occurrence :)

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -425,6 +425,62 @@ class argument_parser
...
     }
 
+    /*!\brief Checks whether the **short** option identifier (`id`) was set on the command line by the user.
+     * \param[in] id The short option identifier to search for.
+     * \returns `true` if option identifier `id` was set on the command line by the user.
+     *
+     * \details
+     *
+     * You can only ask for option identifiers that were added to the parser beforehand via
+     * seqan3::argument_parser::add_option.
```
rrahn, line 435: The API documentation is missing the throw statement and a description when and what this function throws.

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -425,6 +425,62 @@ class argument_parser
...
+
+    /*!\brief Checks whether the **long** option identifier (`id`) was set on the command line by the user.
+     * \param[in] id The long option identifier to search for (must not contain dashes!).
+     * \returns `true` if option identifier `id` was set on the command line by the user.
+     *
+     * \details
+     *
+     * You can only ask for option identifiers that were added to the parser beforehand via
+     * seqan3::argument_parser::add_option.
+     *     *
```
rrahn, line 462:
```suggestion
     *  
```

And also the exception specification is missing.

rrahn (changes requested):

include/seqan3/argument_parser/argument_parser.hpp
smehringer: This does not work because 
```
is_option_set_impl(std::string{id}, false);
```
will trigger the exception of long ids, that they must be longer than one char.

smehringer:
@rrahn polite ping :)

include/seqan3/argument_parser/argument_parser.hpp
rrahn: Wait what? I did care of this by passing this additional bool parameter to the internal interface. See, the check whether the long_id has only one element is only evaluate if `is_long_option_id` is set to true. But it is not in the case of a short id. Other than that the code was exactly the same.

rrahn: Oh, no. This goes on in the parse operations as well. I see.

include/seqan3/argument_parser/argument_parser.hpp
```diff
@@ -425,60 +425,58 @@ class argument_parser
...
         if (!parse_was_called)
-            throw design_error{"You can only asked which options have been set after calling the function `parse()`."};
+            throw design_error{"You can only ask which options have been set after calling the function `parse()`."};
 
-        if (id.size() == 1)
-            throw design_error{"Long option identifiers must be longer than one character! If " + id + "' was meant to "
-                               "be a short identifier, please pass it as a char ('') not a string (\"\")!"};
+        // the detail::format_parse::find_option_id call in the end expects either a char or std::string
+        using char_or_string_t = std::conditional_t<std::same_as<id_type, char>, char, std::string>;
+        char_or_string_t id_ = {id}; // e.g. convert char * to string here if necessary
```
rrahn, line 465:
```suggestion
        char_or_string_t short_or_long_id{id}; // e.g. convert char * to string here if necessary
```

rrahn (changes requested):
lgtm! only renaming of the id_ variable. You can already rebase. Thank you!

rrahn (approved):

PR #1866, opened 2020-05-31
[FIX] cxx20: return-type-requirement is not a type-constraint

marehr:
This PR fixes https://github.com/seqan/product_backlog/issues/88, please
see further explanations in that issue.

This is one of the most important PRs to get `-std=c++2a` support.

With regex: "\{[ ]*(.+?)\s*\}\s*->\s*(.+);" and replace "requires std::same_as<decltype($1), $2>;"

### DIFF
--- a/include/seqan3/alphabet/concept.hpp
+++ b/include/seqan3/alphabet/concept.hpp
@@ -455,7 +455,7 @@ public:
         {
             { impl(priority_tag<3>{}, a, dummy{}) };
             requires noexcept(impl(priority_tag<3>{}, a));
-            requires std::convertible_to<decltype(impl(priority_tag<3>{}, a)), bool>;
+            SEQAN3_RETURN_TYPE_CONTRAINT(impl(priority_tag<3>{}, a), std::convertible_to, bool);
         }
     //!\endcond
     constexpr bool operator()(alphabet_char_t<alph_t> const a) const noexcept
@@ -539,8 +539,8 @@ struct assign_char_strictly_to_fn
     //!\cond
         requires requires (alph_t a, seqan3::alphabet_char_t<alph_t> r)
         {
-            { seqan3::assign_char_to(r, a) } -> alph_t;
-            { seqan3::char_is_valid_for<alph_t>(r) } -> bool;
+            SEQAN3_RETURN_TYPE_CONTRAINT(seqan3::assign_char_to(r, a), std::convertible_to, alph_t);
+            SEQAN3_RETURN_TYPE_CONTRAINT(seqan3::char_is_valid_for<alph_t>(r), std::same_as, bool);
         }
     //!\endcond
     decltype(auto) operator()(seqan3::alphabet_char_t<alph_t> const r, alph_t & a) const
@@ -556,8 +556,8 @@ struct assign_char_strictly_to_fn
     //!\cond
         requires requires (alph_t a, seqan3::alphabet_char_t<alph_t> r)
         {
-            { seqan3::assign_char_to(r, a) } -> alph_t;
-            { seqan3::char_is_valid_for<alph_t>(r) } -> bool;
+            SEQAN3_RETURN_TYPE_CONTRAINT(seqan3::assign_char_to(r, a), std::convertible_to, alph_t);
+            SEQAN3_RETURN_TYPE_CONTRAINT(seqan3::char_is_valid_for<alph_t>(r), std::same_as, bool);
         }
     //!\endcond
     auto operator()(seqan3::alphabet_char_t<alph_t> const r, alph_t && a) const
--- a/include/seqan3/core/concept/tuple.hpp
+++ b/include/seqan3/core/concept/tuple.hpp
@@ -33,7 +33,7 @@ namespace seqan3::detail
 template <typename tuple_t>
 SEQAN3_CONCEPT tuple_size = requires (tuple_t v)
 {
-    { std::tuple_size<tuple_t>::value } -> size_t;
+    SEQAN3_RETURN_TYPE_CONTRAINT(std::tuple_size<tuple_t>::value, std::convertible_to, size_t);
 };
 //!\endcond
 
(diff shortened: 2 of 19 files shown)


include/seqan3/alphabet/concept.hpp
```diff
@@ -538,8 +538,8 @@ struct assign_char_strictly_to_fn
     //!\cond
         requires requires (alph_t a, seqan3::alphabet_char_t<alph_t> r)
         {
-            { seqan3::assign_char_to(r, a) } -> alph_t;
-            { seqan3::char_is_valid_for<alph_t>(r) } -> bool;
+            requires std::convertible_to<decltype(seqan3::assign_char_to(r, a)), alph_t>;
```
smehringer, line 541: why convertible? isn't it `alph_t &`?

include/seqan3/io/alignment_file/output_format_concept.hpp
```diff
@@ -95,25 +95,9 @@ SEQAN3_CONCEPT alignment_file_output_format =
...
-                               flag,
-                               mapq,
-                               mate,
-                               tag_dict,
-                               e_value,
-                               bit_score
-                               ) } -> void;
+    requires std::same_as<decltype(v.write_alignment_record(stream, options, header, seq, qual, id, offset, ref_seq,
+                                                            ref_id, ref_offset, align, cigar, flag, mapq, mate,
+                                                            tag_dict, e_value, bit_score)), void>;
```
smehringer, line 100: I prefer this "multiple parameters on one line " here too but when I wrote it, Rene said this is not our style guide.

include/seqan3/io/stream/concept.hpp
```diff
@@ -16,6 +16,7 @@
 #include <type_traits>
 
 #include <seqan3/core/platform.hpp>
+#include <seqan3/std/concepts>
```
smehringer, line 19: Group this with the std headers (it won't let me do suggestions :( )

include/seqan3/range/container/concept.hpp
```diff
@@ -102,27 +103,27 @@ SEQAN3_CONCEPT container = requires (type val, type val2, type const cval, typen
     >;
 */
     // methods and operator
-    { type{}          } -> type;   // default constructor
-    { type{type{}}    } -> type;   // copy/move constructor
-    { val = val2      } -> type &; // assignment
+    requires std::same_as<decltype(type{}), type>;   // default constructor
+    requires std::same_as<decltype(type{type{}}), type>;   // copy/move constructor
+    requires std::same_as<decltype(val = val2), type &>; // assignment
```
smehringer, line 108:
```suggestion
    requires std::same_as<decltype(type{}), type>;         // default constructor
    requires std::same_as<decltype(type{type{}}), type>;   // copy/move constructor
    requires std::same_as<decltype(val = val2), type &>;   // assignment
```

include/seqan3/io/alignment_file/output_format_concept.hpp
rrahn: Our style guide also says no more than 5 parameters 😏

marehr: :vomiting_face:

include/seqan3/alphabet/concept.hpp
marehr: Nop, this can return proxies.

include/seqan3/range/container/concept.hpp
marehr: Never will I realign things, I will remove alignments :sunglasses:

marehr:
@smehringer I now use macros to enable/disable return type requirements.

BTW gcc 9 does not support `{expression} -> std::same_as<type>` fully either.

include/seqan3/core/platform.hpp
```diff
@@ -49,6 +49,14 @@
 #   error "SeqAn3 requires C++ Concepts, either vie the TS (flag: -fconcepts) or via C++20 (flag: -std=c++2a / -std=c++20)."
 #endif
 
+#if _GNUC_ < 10
```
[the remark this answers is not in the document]
marehr: > I thought for < GCC 10 we use `{expression} -> concept_name<__VA_ARGS__>` since this is the current pre GCC10 style confused

No, pre gcc-10 supported

```c++
{expression} -> bool;
```

>= gcc-10 supports ONLY

```c++
{expression} -> std::same_as<bool>;
```

So in a return type needs to be a `concept` name. Before gcc-10 this `concept` was always `std::convertible_to`. You are able since gcc9 to write the following expression:

```c++
{expression} -> std::same_as<bool>;
```

But it actually has some problems that `std::same_as` can't deal with incomplete types. 

This means we know use

```c++
// for >= gcc10:
{expression} -> concept_name<types...>;
// for < gcc10:
{expression}; requires concept_name<decltype(expression), types...>;
```

> And either way, you said GCC9 is more ismilar to GCC 10?

As I wrote in https://github.com/seqan/seqan3/pull/1866#issuecomment-640303597

> BTW gcc 9 does not support `{expression} -> std::same_as<type>` fully either.

include/seqan3/core/concept/tuple.hpp
```diff
@@ -49,14 +49,19 @@ SEQAN3_CONCEPT tuple_get = requires (tuple_t & v, tuple_t const & v_c)
...
-    { get<0>(std::move(v))   } -> typename std::tuple_element<0, tuple_t>::type;
+    // unfortunately std::assignable_from requires lvalue-reference, but we want to accept xvalues too (returned
+    // proxies)
+    SEQAN3_RETURN_TYPE_CONTRAINT(get<0>(v_c), std::convertible_to, typename std::tuple_element<0, tuple_t>::type);
+    SEQAN3_RETURN_TYPE_CONTRAINT(get<0>(std::move(v)),
+                                 std::convertible_to, typename std::tuple_element<0, tuple_t>::type);
     // TODO: The return type for std::tuple is wrong until gcc-8.0, for gcc > 8.0 this is fixed.
     { get<0>(std::move(v_c)) };// -> typename std::tuple_element<0, tuple_t>::type const &&;
+    // SEQAN3_RETURN_TYPE_CONTRAINT(get<0>(std::move(v_c)),
+    //                              std::convertible_to, typename std::tuple_element<0, tuple_t>::type const &&);
```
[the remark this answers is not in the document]
marehr: :question: before it was

```c++
{ get<0>(std::move(v_c)) };// -> typename std::tuple_element<0, tuple_t>::type const &&;
```

Now I added a version with `SEQAN3_RETURN_TYPE_CONTRAINT`, which is still commented out, because we would need to wrap that in `if gcc > 8.0` (assuming the comment is correct; there is no reference to a gcc ticket.).

I can remove that again if you want.

include/seqan3/core/platform.hpp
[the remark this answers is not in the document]
smehringer: thanks for clarifying!

PR #1871, opened 2020-06-03
[API] Create a version of pipeable_config_element without value

joergi-w:
In this way we can introduce alternative config variables and simply remove the `true` template argument when it's done. 

Discussed in the meeting on 2nd June.


CHANGELOG.md
```diff
@@ -82,6 +82,12 @@ Note that 3.1.0 will be the first API stable release and interfaces in this rele
   * The configuration element `seqan3::search_cfg::mode` does not exist anymore.
     You can replace it by directly using one of the above mentioned "hit strategy" configuration elements
     ([\#1639](https://github.com/seqan/seqan3/pull/1639)).
+    
+#### Core
+
+* Configuration refactoring:
+  * The `seqan3::pipeable_config_element` has an additional template value that specifies whether a `value` member is 
+    present and in use. If a config element uses the `value` member, you need to set the 3rd template argument `true`.
```
eseiler, line 90: I think the configuration implementation is rather detail and is only for during the refactoring, so it shouldn't be in the changelog

include/seqan3/core/algorithm/concept.hpp
```diff
@@ -24,7 +24,7 @@ namespace seqan3
 {
 //!\cond
 // Forward declarations
-template <typename derived_t, typename value_t>
+template <typename derived_t, typename value_t, bool needs_value = false>
```
eseiler, line 27: We said we wanted to add `has_value_member` which defaults to `true` and that specifying `true` is the new behaviour; but this can be of course done one way or the other :)

The actual point I want to make is, that if you default it here to `true`, you don't need to touch every configuration object (they will just default to `true` and have the member).
You would only need to change it for the new ones - which I think is preferable, because there are already some PRs that refactor configs and they would all get merge conflicts.

include/seqan3/core/algorithm/concept.hpp
```diff
@@ -57,19 +57,15 @@ class config_element_base;
...
 //!\cond
 template <typename config_t>
 SEQAN3_CONCEPT config_element = std::semiregular<std::remove_reference_t<config_t>> &&
 requires (config_t c)
 {
     { std::remove_reference_t<config_t>::id };
-    { c.value };
     // Must inherit from the pipeable_config_element class.
-    requires std::is_base_of_v<pipeable_config_element<config_t, decltype(c.value)>, config_t>;
+    // requires std::is_base_of_v<pipeable_config_element<config_t, decltype(c.value)>, config_t>;
```
eseiler, line 68: Mhm, I'm not entirely sure how to handle this; we probably need to expose the value_t of the configuration?

include/seqan3/core/algorithm/concept.hpp
joergi-w: I thought the other way round: If now we add the additional parameter, then each PR will remove them gradually, and we reach the final design without further API breaks. In the end, we can simply remove the version with 3 template parameters. 

If we default to `true` now, then we have to remove all the additional parameters in another big API change that touches every config object. Are you sure this is the way we want to go? If yes, then I'll revert my changes tomorrow and apply your suggestion.

include/seqan3/core/algorithm/concept.hpp
joergi-w: Yes, I was thinking the same here, but that almost feels like the old design... Is `value_t` actually used in the new design or is its purpose only the concept check? My hope was that after refactoring we could just include this line again without the decltype part.

CHANGELOG.md
joergi-w: To me it looks like a public API change... However, I'll delete it.

include/seqan3/core/algorithm/concept.hpp
marehr: My original intent was to not touch existing code, because everyone who touches will use the `false` one. If everyone uses `false` we can remove the extra.

Your approach is exactly the same, but in reverse. I don't mind this works equally as good. And I think it is the same amount of work.

----

BTW

Oh I haven't seen that the second template parameter is not needed anymore after this.
```suggestion
template <typename derived_t, typename value_t = void, bool needs_value = false>
```
so we need to do this :)

```cpp
struct hit_strata : public pipeable_config_element<hit_strata, uint8_t, true>

// will become after adaptation

struct hit_strata : public pipeable_config_element<hit_strata>
```

include/seqan3/core/algorithm/concept.hpp
marehr: No, just remove it.

CHANGELOG.md
marehr: This is internal and is thus not a API change :)

include/seqan3/core/algorithm/configuration.hpp
```diff
@@ -32,30 +32,34 @@ namespace seqan3
...
 
-template <typename lhs_derived_t, typename lhs_value_t, typename rhs_derived_t, typename rhs_value_t>
-constexpr auto operator|(pipeable_config_element<lhs_derived_t, lhs_value_t> && lhs,
-                         pipeable_config_element<rhs_derived_t, rhs_value_t> && rhs)
+template <typename lhs_derived_t, typename lhs_value_t, bool lhs_needs_value,
+          typename rhs_derived_t, typename rhs_value_t, bool rhs_needs_value>
+constexpr auto operator|(pipeable_config_element<lhs_derived_t, lhs_value_t, lhs_needs_value> && lhs,
+                         pipeable_config_element<rhs_derived_t, rhs_value_t, rhs_needs_value> && rhs)
 {
     return configuration{static_cast<lhs_derived_t &&>(lhs)}.push_back(static_cast<rhs_derived_t &&>(rhs));
```
marehr, line 40: This will be `undefined behaviour` if the given config `lhs` is not of type `lhs_derived_t`.

marehr:
I think must go back to the drawing board, because it will not work as intended:

https://godbolt.org/z/YjFyZc

---

We can't keep this an aggregate type if we want to keep the base class approach.

* We could make them aggregate, but we would need remove the base class and thus change the piping syntax (over concepts).
* We could keep the base class, making them non-aggregate, but we would need to add constructors.  (note: the different issue with the piping; the base class shouldn't be allowed to be constructed)

CHANGELOG.md
joergi-w: It's removed :)

marehr:
@rrahn used in #1876 the constructor variant, see 

https://github.com/seqan/seqan3/pull/1876/commits/8cec061295bb381f4372d8dd52fb0d662997ace2#diff-f9b7ceceaf010f2731caabb17553d00aR72-R73

joergi-w:
Are you discussing this today in the core meeting? Please let me know if there is a decision and then I can implement it accordingly.

marehr:
Core-Meeting 2020-06-08:

Configs will add constructors for their members

include/seqan3/core/algorithm/pipeable_config_element.hpp
```diff
@@ -27,7 +28,15 @@ namespace seqan3
...
-struct pipeable_config_element
+struct pipeable_config_element<derived_t, value_t, false>
+{
+    static_assert(std::is_same_v<value_t, void>,
+                  "Please remove the value_t template parameter if needs_value == false.");
+};
+
+//! \overload
+template <typename derived_t, typename value_t>
+struct pipeable_config_element<derived_t, value_t, true>
```
marehr, line 39:
```suggestion
struct pipeable_config_element<derived_t, value_t>;

template <typename derived_t>
struct pipeable_config_element<derived_t, void>;
{};

//! \overload
template <typename derived_t, typename value_t>
struct pipeable_config_element<derived_t, value_t>
```

include/seqan3/core/algorithm/pipeable_config_element.hpp
```diff
@@ -27,7 +28,15 @@ namespace seqan3
  * \tparam value_t   The type of the wrapped value.
  */
 template <typename derived_t, typename value_t>
```
marehr, line 30:
```suggestion
template <typename derived_t, typename value_t = void>
```

include/seqan3/core/algorithm/configuration.hpp
joergi-w: Is there a solution for this that I should include in this PR?
Maybe we can assert or require the type dependency?

joergi-w:
Rebased on master.

marehr:
Travis says:

seqan3/include/seqan3/core/type_list/traits.hpp:40:

```c++
/*!\brief Implementation for seqan3::pack_traits::find_if.
 * \tparam pred_t   The predicate that is being evaluated.
 * \tparam pack_t   The type pack.
 * \returns The position of the first type `t` in `pack_t` for whom `pred_t<t>::value` is true.
 * \ingroup type_list
 */
```

It seems that `pred_t<t>::value` is associated with `configuration::value` xD

Let's look in the docs: https://docs.seqan.de/seqan/3-master-dev/group__type__list.html#ga5246fc497be7a3b38e42b804f35bcd7e

Indeed :laughing: 

Not sure how to escape this class name with doxygen.

eseiler:
> ```c++
> `pred_t<t>::value`
> ```

`%pred_t<t>::value` should work. `%` disables link generation. 
But you should look at the generated docs to make sure it's correct

joergi-w:
> `%pred_t<t>::value` should work. `%` disables link generation.
> But you should look at the generated docs to make sure it's correct

Thank you pointing this out! I'm going to fix it.

joergi-w:
I could fix the documentation with the `%` sign and double backticks. The output website looks good. Now the Travis CI is green, but Jenkins reports red, although the tests were successful.

eseiler:
Thanks for checking the docs :+1:

Jenkins fail is unrelated.

Issue #1877, opened 2020-06-06
gcc10: Compile failure on search snippet


SGSSGene:
## Platform

  - SeqAn version: cf2505c (6. June)
  - Operating system: `Linux schenker 5.6.15-arch1-1 #1 SMP PREEMPT Wed, 27 May 2020 23:42:26 +0000 x86_64 GNU/Linux`
  - Compiler: `gcc (GCC) 10.1.0 `

## Description
The snippet `test/snippet/search/search.cpp` does not compiling.

### How to repeat the problem
I used seqan3/app-template and updated seqan3 and replaced `main.cpp` with the content of `test/snippet/search/search.cpp` (Is there a easier way to test compile snippets?)

### Expected behaviour
Expected to compile.

### Actual behaviour
Compile error:
```
In file included from /home/gene/app-template/lib/seqan3/include/seqan3/search/configuration/result_type.hpp:20,
                 from /home/gene/app-template/lib/seqan3/include/seqan3/search/detail/search_configurator.hpp:21,
                 from /home/gene/app-template/lib/seqan3/include/seqan3/search/search.hpp:22,
                 from /home/gene/app-template/src/main.cpp:5:
/home/gene/app-template/lib/seqan3/include/seqan3/search/search_result.hpp: In instantiation of ‘class seqan3::search_result<long unsigned int, seqan3::detail::empty_type, long unsigned int, long unsigned int>’:
/home/gene/app-template/lib/seqan3/include/seqan3/search/detail/search_configurator.hpp:158:68:   required from ‘static auto seqan3::detail::search_configurator::configure_algorithm(const configuration_t&, const index_t&) [with query_t = std::vector<seqan3::dna4>&; configuration_t = seqan3::configuration<seqan3::search_cfg::max_error, seqan3::search_cfg::output<seqan3::detail::search_output_text_position>, seqan3::detail::hit_all_tag>; index_t = seqan3::fm_index<seqan3::dna4, seqan3::collection, sdsl::csa_wt<sdsl::wt_pc<sdsl::balanced_shape, sdsl::int_vector<1>, sdsl::rank_support_v<1, 1>, sdsl::select_support_scan<>, sdsl::select_support_scan<0>, sdsl::byte_tree<> >, 16, 10000000, sdsl::sa_order_sa_sampling<>, sdsl::isa_sampling<>, sdsl::plain_byte_alphabet> >]’
[...]

PR #1878, opened 2020-06-06
[FIX] gcc-10.1: workaround for ICE

marehr:
```c++
core/algorithm/configuration_test.cpp:209:72: internal compiler error: Segmentation fault
  209 |         EXPECT_TRUE((std::same_as<decltype(cfg.template remove<foobar>()), seqan3::configuration<foo, bar>>));
      |                                                                        ^
```

Fixes partially #1877


rrahn:
We might get rid of it altogether when we use the free function interfaces instead? Should we wait for this?

PR #1880, opened 2020-06-06
[FEATURE] Adjust the search range and search configuration to use algorithm executor.

rrahn:
### Changes:

Updates the two search algorithms be invoked with a second callback argument.
This callback is invoked with the generated search results.
Also the search algorithm now associates an index
with every query and propagates this indexed query pair to the algorithms.
Updates the search result range to work with the algorithm executor.

Supersedes #1855


include/seqan3/search/search_result_range.hpp
```diff
@@ -70,31 +67,21 @@ class search_result_range
...
      */
 
-    /*!\brief Returns an iterator to the first element of the search result range.
-     * \return An iterator to the first element.
-     *
-     * \details
-     *
-     * The fist invocation of this function will invoke the search for the first query in the underlying range.
-     */
+    //!\brief Returns an iterator to the first element of the search result range.
```
wvandertoorn, line 81: I agree that the `\brief` is already quite descriptive, but I think keeping the `\return` would be nice for the documentation. That way the documentation is consistent for all functions in this file (see  `end()` and `next()` for example).

test/unit/search/search_result_range_test.cpp
```diff
@@ -21,71 +20,56 @@
...
-using search_result_range_t = seqan3::search_result_range<dummy_search_algorithm,
-                                                          seqan3::type_reduce_view<std::vector<std::string>& >>;
-using search_range_iterator = std::ranges::iterator_t<search_result_range_t>;
-
 struct search_result_range_test : ::testing::Test
 {
-    std::vector<std::string> query_range{std::string{"query1"}, std::string{"query1"}, std::string{"query3"}};
+    using indexed_queries_t = std::vector<std::pair<size_t, std::string>>;
+    using algorithm_result_t = std::pair<size_t, size_t>;
+
```
wvandertoorn, line 48:
```suggestion
```

include/seqan3/search/detail/policy_search_result_builder.hpp
```diff
@@ -39,73 +39,81 @@ struct policy_search_result_builder
...
+     * \param[in] callback The callback to invoke for every hit.
      *
      * \details
      *
      * The result is independent from the search modus (all, single_best, all_best, strata).
      */
-    template <typename index_cursor_t>
-    auto make_results(std::vector<index_cursor_t> internal_hits)
+    template <typename index_cursor_t, typename query_index_t, typename callback_t>
+    auto make_results(std::vector<index_cursor_t> internal_hits, query_index_t idx, callback_t && callback)
```
smehringer, line 56:
```suggestion
    void make_results(std::vector<index_cursor_t> internal_hits, query_index_t idx, callback_t && callback)
```
more expressive now?

also for the others if you agree

PR #1882, opened 2020-06-08
[FIX] cxx20: make small_string a proper structural type

marehr:
‘seqan3::small_string<2>’ is not a valid type for a template non-type parameter because it is not structural

Definition of structural type: https://wg21.link/p1907

```c++
In file included from seqan3/test/unit/io/alignment_file/sam_tag_dictionary_test.cpp:12:
seqan3/include/seqan3/io/alignment_file/sam_tag_dictionary.hpp:64:27: error: ‘seqan3::small_string<2>’ is not a valid type for a template non-type parameter because it is not structural
   64 | template <small_string<2> str> // TODO: better handling if too large string is provided?
      |                           ^~~
In file included from seqan3/include/seqan3/io/alignment_file/sam_tag_dictionary.hpp:21,
                 from seqan3/test/unit/io/alignment_file/sam_tag_dictionary_test.cpp:12:
seqan3/include/seqan3/range/container/small_string.hpp:42:7: note: ‘seqan3::small_string<2>::<anonymous>’ has a non-structural type
   42 | class small_string : public small_vector<char, capacity_ + 1>
      |       ^~~~~~~~~~~~
In file included from seqan3/include/seqan3/range/container/small_string.hpp:15,
                 from seqan3/include/seqan3/io/alignment_file/sam_tag_dictionary.hpp:21,
                 from seqan3/test/unit/io/alignment_file/sam_tag_dictionary_test.cpp:12:
seqan3/include/seqan3/range/container/small_vector.hpp:927:39: note: ‘seqan3::small_vector<char, 3>::data_’ is not public
  927 |     std::array<value_type, capacity_> data_{};
      |                                       ^~~~~
```

### DIFF
--- a/include/seqan3/range/container/small_vector.hpp
+++ b/include/seqan3/range/container/small_vector.hpp
@@ -920,7 +920,7 @@ public:
     }
     //!\}
 
-protected:
+public:
     //!\privatesection
 
     //!\brief Stores the actual data_.


include/seqan3/range/container/small_vector.hpp
```diff
@@ -920,7 +920,7 @@ class small_vector
     }
     //!\}
 
-protected:
+public:
     //!\privatesection
```
joergi-w, line 924: For the change to public interface I suggest a few further actions:
1. Is `\privatesection` still valid? To me it looks like it should be removed. 
2. Rename `sz` and `data_` to `size` and `data`, as they are public now.
3. Remove the `public:` specifier below, it is duplicated.

marehr: we already have public members `data` and `size`.

`\privatesection` is still valid.

joergi-w: I don't understand why it is still valid, but if you know what you're doing I'm fine ;)

marehr: `sz` and `data_` are still private, because our recommended way to access that data is `size()` and `data()`. So we should not advertise the "private" members in our documentation.

joergi-w: Okay, you can fix only point 3 and then I approve :)

marehr: The `public:` below re-enables public documentation. If we merge them I would need to add a `//!\publicsection` comment. I would prefer to keep `public:` instead.

joergi-w: okay, then we can keep it...

smehringer: I agree with point 3. of joerg but I don't care much

PR #1883, opened 2020-06-08
[FIX] workaround gcc10: view + std::filesystem:path constructor = bad

marehr:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=93983 has the consequences
that a view with a custom constructor of std::filesystem::path produces
an infinite loop.

### DIFF
--- a/doc/tutorial/alignment_file/alignment_file_solution2.cpp
+++ b/doc/tutorial/alignment_file/alignment_file_solution2.cpp
@@ -91,9 +91,15 @@ int main()
 
     seqan3::alignment_file_input mapping_file{tmp_dir/"mapping.sam", ref_ids, ref_seqs, field_type{}};
 
+#if !SEQAN3_WORKAROUND_GCC_93983
     auto mapq_filter = std::views::filter([] (auto & rec) { return seqan3::get<seqan3::field::mapq>(rec) >= 30; });
+#endif // !SEQAN3_WORKAROUND_GCC_93983
 
+#if SEQAN3_WORKAROUND_GCC_93983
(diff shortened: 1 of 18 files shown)


eseiler:
Can you check if `SEQAN3_WORKAROUND_GCC_93983` shows up in the affected snippets? If yes, adapt our doxygen config:

https://github.com/seqan/seqan3/blob/cf2505cb51715f89f1bc0c658f681f36fac89dfa/test/documentation/seqan3_doxygen_cfg.in#L55
should then be
```console
EXPAND_AS_DEFINED      = SEQAN3_CPO_IMPL SEQAN3_DEPRECATED_310 SEQAN3_WORKAROUND_GCC_93983
```

marehr:
> Can you check if `SEQAN3_WORKAROUND_GCC_93983` shows up in the affected snippets? If yes, adapt our doxygen config:
> 
> https://github.com/seqan/seqan3/blob/cf2505cb51715f89f1bc0c658f681f36fac89dfa/test/documentation/seqan3_doxygen_cfg.in#L55
> 
> should then be
> 
> ```
> EXPAND_AS_DEFINED      = SEQAN3_CPO_IMPL SEQAN3_DEPRECATED_310 SEQAN3_WORKAROUND_GCC_93983
> ```

It has a default meaning so, no this does not need to be added.

marehr:
> Rebase+Addressing my earlier comment:
> [#1883 (comment)](https://github.com/seqan/seqan3/pull/1883#issuecomment-640492066)

I tried it out and it didn't work :(

I think this is a "feature" request that macros are evaluated in snippets.

PR #1885, opened 2020-06-08
[FIX] cxx20: repeat_view uses single_view wrong?!

marehr:
```c++
/seqan3/include/seqan3/range/views/repeat.hpp:178:100: error: no matching function for call to ‘std::ranges::single_view<std::_Swallow_assign>::single_view(<brace-enclosed initializer list>)’
  178 |     constexpr explicit repeat_view(value_t const & value) : single_value{ranges::single_view{value}}
      |
```

### DIFF
--- a/include/seqan3/range/views/repeat.hpp
+++ b/include/seqan3/range/views/repeat.hpp
@@ -50,6 +50,9 @@ private:
     //!\brief The sentinel type is set to std::default_sentinel_t.
     using sentinel_type = std::default_sentinel_t;
 
+    //!\brief The view which wraps the single value to repeat.
+    using single_value_t = decltype(std::views::single(std::declval<value_t>()));
+
     /*!\name Associated types
      * These associated types are needed in seqan3::detail::random_access_iterator.
      * \{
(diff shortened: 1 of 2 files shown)


include/seqan3/range/views/repeat.hpp
```diff
@@ -50,6 +50,9 @@ class repeat_view : public std::ranges::view_interface<repeat_view<value_t>>
     //!\brief The sentinel type is set to std::default_sentinel_t.
     using sentinel_type = std::default_sentinel_t;
 
+    //!\brief The sentinel type is set to std::ranges::default_sentinel_t.
```
simonsasse, line 53: I think this Line should contain something else. Maybe `The single value type is set to XXX`

include/seqan3/range/views/repeat.hpp
```diff
@@ -284,7 +287,7 @@ class repeat_view : public std::ranges::view_interface<repeat_view<value_t>>
 
 private:
     //!\brief A std::views::single over the input.
```
simonsasse, line 289: what is a view with a single element good for ? (just an understanding question)

marehr: Sometimes you want to wrap a single element and make it a range.

It is kind of the same as making the number `5` a set with a single element `5` i.e. `{5}`. To allow set operations on that single element collection.

include/seqan3/range/views/repeat.hpp
marehr: You are absolutely right :)

PR #1886, opened 2020-06-08
[FIX] cxx20: view_to_simd add explicit empty()

marehr:
Add explicit empty() since the implicit definition from view_interface
requires std::ranges::forward_range, but view_to_simd is only an input range.

See https://eel.is/c++draft/view.interface#1

### DIFF
--- a/include/seqan3/core/simd/view_to_simd.hpp
+++ b/include/seqan3/core/simd/view_to_simd.hpp
@@ -166,7 +166,8 @@ public:
     constexpr void cend() const noexcept = delete;
     //!\}
 
-    constexpr size_t empty() const noexcept
+    //!\brief Checks whether the range is empty.
+    constexpr bool empty() const noexcept
     //!\cond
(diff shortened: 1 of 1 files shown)


include/seqan3/core/simd/view_to_simd.hpp
```diff
@@ -166,6 +166,17 @@ class view_to_simd : public std::ranges::view_interface<view_to_simd<urng_t, sim
     constexpr void cend() const noexcept = delete;
     //!\}
 
+    constexpr size_t empty() const noexcept
```
marehr, line 169:
```suggestion
    //!\brief Checks whether the range is empty.
    constexpr bool empty() const noexcept
```

marehr:
> Is this already tested?

Yes, see

https://github.com/seqan/seqan3/blob/32c1b2e960ce2083c3d89e9e6687d7129a612490/test/unit/core/simd/view_to_simd_test.cpp#L214-L225

PR #1887, opened 2020-06-08
[TEST] cxx20: add regression test for zip

marehr:
The ranges library had a problem with C++20 and their views::zip implementation
(see https://github.com/ericniebler/range-v3/issues/1480).

This PR adds a regression test.


test/unit/range/views/view_zip_test.cpp
```diff
@@ -0,0 +1,32 @@
...
+#include <seqan3/std/ranges>
+
+// https://github.com/ericniebler/range-v3/issues/1480
+TEST(zip_view, gcc10bug_rangev3_1480)
+{
+    std::vector<char> const first_sequence{};
+    std::vector<char> const second_sequence{};
+
+    auto zip_view = seqan3::views::zip(first_sequence, second_sequence);
+    auto fn = [&] ([[maybe_unused]] auto && ch)
```
wvandertoorn, line 27: I think it would be good to make the names a bit more descriptive.

rrahn: not resolved?

rrahn, line 27:
```suggestion
    auto fn = [&] ([[maybe_unused]] auto && ch)
```
Why is `ch` maybe_unused? Is that part of the test?

test/unit/range/views/view_zip_test.cpp
```diff
@@ -0,0 +1,32 @@
...
+    {
+        EXPECT_TRUE((std::same_as<decltype(ch), const char>));
+    };
+    std::ranges::for_each(zip_view, fn);
```
rrahn, line 31:
```suggestion
    std::ranges::for_each(zip_view, test_element);
```

test/unit/range/views/view_zip_test.cpp
```diff
@@ -0,0 +1,32 @@
...
+// https://github.com/ericniebler/range-v3/issues/1480
+TEST(zip_view, gcc10bug_rangev3_1480)
+{
+    std::vector<char> const first_sequence{};
+    std::vector<char> const second_sequence{};
+
+    auto zip_view = seqan3::views::zip(first_sequence, second_sequence);
+    auto fn = [&] ([[maybe_unused]] auto && ch)
+    {
+        EXPECT_TRUE((std::same_as<decltype(ch), const char>));
```
rrahn, line 29: Should `ch` not a be a `std::pair<char const &, char const &>`?

marehr: You are right, but since it is `ranges::common_tuple<char const &, char const &>`. I removed this EXPECT_TRUE.

rrahn: `ranges::common_pair`, right?

test/unit/range/views/view_zip_test.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+#include <gtest/gtest.h>
+
+#include <seqan3/range/views/zip.hpp>
+#include <seqan3/std/algorithm>
+#include <seqan3/std/ranges>
+
+// https://github.com/ericniebler/range-v3/issues/1480
+TEST(zip_view, gcc10bug_rangev3_1480)
+{
```
rrahn, line 22:
```suggestion
{ // This regression test only checks if the respective code compiles.
```

test/unit/range/views/view_zip_test.cpp
marehr: No `common_tuple`

test/unit/range/views/view_zip_test.cpp
```diff
@@ -0,0 +1,28 @@
...
+#include <gtest/gtest.h>
+
+#include <seqan3/range/views/zip.hpp>
+#include <seqan3/std/algorithm>
+#include <seqan3/std/ranges>
+
+// https://github.com/ericniebler/range-v3/issues/1480
+TEST(zip_view, gcc10bug_rangev3_1480)
+{ // This regression test only checks if the respective code compiles.
+    std::vector<char> const first_sequence{};
```
marehr, line 23:
```suggestion
{
    // This regression test only checks if the respective code compiles.
    std::vector<char> const first_sequence{};
```

Issue #1542, opened 2020-01-28
Search on bi_fm_index finds too many results


SGSSGene:
## Platform

  - SeqAn version: master (b0b27968), also happens on older master (3937c594)
  - Operating system: `linux schenker 5.4.13-arch1-1 #1 SMP PREEMPT Fri, 17 Jan 2020 23:09:54 +0000 x86_64 GNU/Linux`
  - Compiler: gcc 9.2.0

## Description
Searches on an bi_fm_index can lead to results which have more errors than specified.

### How to repeat the problem
Code that I used

```c++
// swapping format.h and ranges will "fix" the issue
#include <fmt/chrono.h>
#include <fmt/color.h>
#include <fmt/format.h>


#include <seqan3/alphabet/all.hpp>
#include <seqan3/core/debug_stream.hpp>
#include <seqan3/io/sequence_file/all.hpp>
#include <seqan3/io/sequence_file/all.hpp>
#include <seqan3/search/algorithm/search.hpp>
#include <seqan3/search/fm_index/bi_fm_index.hpp>
#include <seqan3/std/filesystem>
#include <seqan3/std/ranges>
#include <seqan3/test/performance/sequence_generator.hpp>

using namespace seqan3;

int main(int argc, char * argv[]) {

	auto genome = std::vector<std::vector<seqan3::dna4>>{seqan3::test::generate_sequence<seqan3::dna4>(100'000'000, 0, 0)};

    auto index = seqan3::bi_fm_index{genome};
    // Note: instead of creating a new index, we reload the index
    //decltype(seqan3::bi_fm_index{genome}) index;
	//{
		//std::ifstream is{"index-random.file", std::ios::binary};
		//cereal::BinaryInputArchive iarchive{is};
		//iarchive(index);
	//}


	// create query - since it uses the same seed as an the index creation, we will have a result at pos 0
	auto queries = std::vector<std::vector<seqan3::dna4>>{};
	for (int i{0}; i < 1; ++i) {
		queries.push_back(seqan3::test::generate_sequence<seqan3::dna4>(20, 0, i));
	}

	{
		seqan3::configuration const cfg = seqan3::search_cfg::max_error{
		    seqan3::search_cfg::total{3},
		    seqan3::search_cfg::substitution{3},
		    seqan3::search_cfg::insertion{0},
		    seqan3::search_cfg::deletion{0}};

		auto all_hits = seqan3::search(queries, index, cfg);
		for (auto hits : all_hits) {
[...]

PR #1892, opened 2020-06-10
[MISC] Refactor fm_index

eseiler:
Resolves https://github.com/seqan/seqan3/issues/1542


marehr:
```c++
auto genome = std::vector<"CGTAGGTCGATGC"_dna4, "CAAAA"_dna4>;
seqan3::bi_fm_index index{genome};
auto it = index.cursor(); // bidirectional cursor
it.extend_right("CGT"_dna4); // search CGT
seqan3::debug_stream << it.locate() << '\n'; // (0,0)
auto rev_it = it.to_rev_cursor();
seqan3::debug_stream << rev_it.locate() << '\n'; // (?,?)
```

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -25,6 +25,59 @@
...
+            if (std::ranges::empty(text))
+                throw std::invalid_argument("The text to index cannot be empty.");
+        }
+        else
+        {
+            static_assert(std::ranges::bidirectional_range<text_t>,
+                          "The text collection must model bidirectional_range.");
+            static_assert(std::ranges::bidirectional_range<std::ranges::range_reference_t<text_t>>,
+                          "The elements of the text collection must model bidirectional_range.");
+            static_assert(alphabet_size<range_innermost_value_t<text_t>> <= 256, "The alphabet is too big.");
```
joergi-w, line 69: Maybe it's not important here, but I spotted that this can be moved out of the `if constexpr` block, because it is the same statement in both branches.

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -229,61 +284,46 @@ class fm_index
...
-        static_assert(alphabet_size<range_innermost_value_t<text_t>> <= 256, "The alphabet is too big.");
-        static_assert(std::convertible_to<range_innermost_value_t<text_t>, alphabet_t>,
-                     "The alphabet of the text collection must be convertible to the alphabet of the index.");
-        static_assert(range_dimension_v<text_t> == 2, "The input must be a text collection.");
+        detail::fm_index_validator::validate<alphabet_t, text_layout_mode_>(text);
 
-        // text collection must not be empty
-        if (std::ranges::begin(text) == std::ranges::end(text))
-            throw std::invalid_argument("The text that is indexed cannot be empty.");
+        std::vector<size_t> text_sizes;
```
joergi-w, line 291: Is the text size known (std::ranges::distance)? Then you could allocate the vector to its final length here...

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -304,12 +344,28 @@ class fm_index
...
-        // we need at least one delimiter
-        if (std::ranges::distance(text) == 1)
-            tmp_text.back() = delimiter;
-
-        std::ranges::reverse(tmp_text);
+            std::ranges::reverse(tmp_text);
+        }
+        else
+        {
+            // If only one text is in the text collection, we need at least one delimiter.
```
joergi-w, line 357: This comment seems contradictory: What do we need to delimit, if there is just one text? And _at least one_ means there can be even multiple, or did you mean _at most_?
Maybe you can give some more details to avoid confusion here.

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -525,3 +581,62 @@ fm_index(text_t &&) -> fm_index<range_innermost_value_t<text_t>, text_layout{ran
...
+ * \implements seqan3::cerealisable
+ *
+ * \details
+ *
+ *  This FM Index reverses the given text before constructing the seqan3::fm_index.
+ *  This type is used by the seqan3::bi_fm_index.
+ */
+template <semialphabet alphabet_t,
+        text_layout text_layout_mode_,
+        detail::sdsl_index sdsl_index_type_ = default_sdsl_index_type>
```
joergi-w, line 603: Why do the template parameters have the trailing underscore but `alphabet_t` has not? 
```suggestion
          text_layout text_layout_mode,
          detail::sdsl_index sdsl_index_type = default_sdsl_index_type>
```

joergi-w: These lines are not aligned with the line above.

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -525,3 +581,62 @@ fm_index(text_t &&) -> fm_index<range_innermost_value_t<text_t>, text_layout{ran
...
+
+        this->construct(reverse_text, true);
+    }
+
+    //!\copydoc fm_index::construct()
+    template <std::ranges::range text_t>
+    //!\cond
+        requires (text_layout_mode_ == text_layout::single)
+    //!\endcond
+    void construct_(text_t && text)
```
joergi-w, line 624: Can't these overloads be avoided with a simple `if constexpr` inside the function?

test/unit/search/search_collection_test.cpp
```diff
@@ -143,17 +143,18 @@ TYPED_TEST(search_test, convertible_query)
...
 
-    seqan3::configuration const cfg = seqan3::search_cfg::max_error_total{seqan3::search_cfg::error_count{1}} |
-                                      seqan3::search_cfg::max_error_substitution{seqan3::search_cfg::error_count{1}} |
+    seqan3::configuration const cfg = seqan3::search_cfg::max_error_total{seqan3::search_cfg::error_count{3}} |
+                                      seqan3::search_cfg::max_error_substitution{seqan3::search_cfg::error_count{3}} |
                                       seqan3::search_cfg::max_error_insertion{seqan3::search_cfg::error_count{0}} |
                                       seqan3::search_cfg::max_error_deletion{seqan3::search_cfg::error_count{0}};
     typename TestFixture::hits_result_t expected_hits{{0, 0}};
-    EXPECT_RANGE_EQ(search("ACGACACG"_dna4, index, cfg) | ref_id_and_position, expected_hits);
+    EXPECT_RANGE_EQ(search("GGTGTGTCG"_dna4, index, cfg) | ref_id_and_position, expected_hits);
```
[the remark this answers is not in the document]
eseiler: It's an example from https://github.com/seqan/seqan3/issues/1542 where there were wrong results; if it's an exact match or not doesn't matter too much because this is just a small test to cover the special case when there is only one text in the collection - so I'm basically just replacing the test with a better one.

include/seqan3/search/fm_index/bi_fm_index.hpp
```diff
@@ -120,30 +120,14 @@ class bi_fm_index
...
-    //!\endcond
     void construct(text_t && text)
     {
         detail::fm_index_validator::validate<alphabet_t, text_layout_mode_>(text);
 
         fwd_fm = fm_index_type{text};
         rev_fm = rev_fm_index_type{text};
     }
-
+
```
joergi-w, line 130: Here are spaces accidentally inserted 💅

include/seqan3/search/fm_index/fm_index.hpp
joergi-w: ok!

include/seqan3/search/fm_index/fm_index.hpp
joergi-w: Thanks!

include/seqan3/search/fm_index/bi_fm_index.hpp
```diff
@@ -72,7 +72,12 @@ class bi_fm_index
     using sdsl_index_type = sdsl_index_type_;
 
     //!\brief The type of the underlying SDSL index for the reversed text.
-    using rev_sdsl_index_type = sdsl_index_type_;
+    using rev_sdsl_index_type = sdsl::csa_wt<sdsl_wt_index_type::wavelet_tree_type, // Wavelet tree type
+                                             10'000'000, // Sampling rate of the suffix array
+                                             10'000'000, // Sampling rate of the inverse suffix array
```
smehringer, line 77: Just a side question. Why do we need an inverse suffix array?

include/seqan3/search/fm_index/bi_fm_index.hpp
```diff
@@ -72,7 +72,12 @@ class bi_fm_index
     using sdsl_index_type = sdsl_index_type_;
 
     //!\brief The type of the underlying SDSL index for the reversed text.
-    using rev_sdsl_index_type = sdsl_index_type_;
+    using rev_sdsl_index_type = sdsl::csa_wt<sdsl_wt_index_type::wavelet_tree_type, // Wavelet tree type
```
smehringer, line 75: Does this type need to be "compatible" with the forward index type? Meaning that some template parameters have to be the same? If so we could maybe static_assert the respective requirements.

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -339,12 +344,28 @@ class fm_index
...
+            // we need at least one delimiter
+            if (number_of_texts == 1)
+                tmp_text.back() = delimiter;
 
-        // we need at least one delimiter
-        if (std::ranges::distance(text) == 1)
-            tmp_text.back() = delimiter;
-
-        std::ranges::reverse(tmp_text);
+            std::ranges::reverse(tmp_text);
```
smehringer, line 352: why are we reversing the text here in the `if (!reverse)` branch?

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -560,3 +581,62 @@ fm_index(text_t &&) -> fm_index<range_innermost_value_t<text_t>, text_layout{ran
...
+    //!\cond
+        requires (text_layout_mode_ == text_layout::single)
+    //!\endcond
+    void construct_(text_t && text)
+    {
+        auto reverse_text = text | std::views::reverse;
+
+        this->construct(reverse_text);
+    }
+public:
```
smehringer:
```suggestion
    }
    
public:
```

include/seqan3/search/fm_index/fm_index.hpp
```diff
@@ -274,51 +284,46 @@ class fm_index
     //!\cond
         requires (text_layout_mode_ == text_layout::collection)
     //!\endcond
-    void construct(text_t && text)
+    void construct(text_t && text, bool reverse = false)
```
smehringer, line 286: why does this `construct` has a boolean for reversing in the normal `fm_index` when you have a whole class that can handle all the reversing?

include/seqan3/search/fm_index/bi_fm_index.hpp
eseiler: Not too sure, AFAIK it's sometimes used for construction stuff; but we don't need it - but SDSL want's a rate anyway.

include/seqan3/search/fm_index/bi_fm_index.hpp
eseiler: Mhm, one question would be, if we allow stuff other than `csa_wt` - I don't think so. We could double check this.
Other than that the template params are taken from the forward index - sampling rate is fixed for the reverse index and the sampling type is also fixed. it doesn't really matter, because we will never ask the reverse index for location, so the sampling does not matter. Also, the sampling types are policy classes and not exposed in the SDSL, so I cant just do `sdsl_wt_index_type::sa_sampling_t`.

include/seqan3/search/fm_index/fm_index.hpp
eseiler: the `reverse` is whether the original text was reversed (reverse_fm_index).
If it's a "normal" fm_index, you can just reverse the test; if the text comes from the `reverse_fm_index` you have to do special stuff if there is only one item in the collection

include/seqan3/search/fm_index/fm_index.hpp
eseiler: It just handles the reversing of the text; the alternative would be to copy the complete code of the construct method to the reverse_fm_index and add the tiny if I need...

include/seqan3/search/fm_index/bi_fm_index.hpp
smehringer: I hope that it's not build if it\s not needed :see_no_evil:

PR #1893, opened 2020-06-10
Algorithm configuration/add get or function

rrahn:
Part of seqan/product_backlog#92

Replaces the value_or with the get_or interface which has a different syntax and does not assume that there is a value member anymore. This is another step of adapting the configuration elements.


test/unit/core/algorithm/configuration_test.cpp
```diff
@@ -222,58 +223,139 @@ TEST(configuration, remove_by_type_template)
...
+        EXPECT_EQ(cfg.get_or(foo{"test"}).value, "test");
     }
 
     { // const l-value
-        seqan3::configuration<bax, bar> const cfg_c{cfg};
-        EXPECT_FLOAT_EQ(cfg_c.value_or<bax>(1.3), 2.2);
-        EXPECT_FLOAT_EQ(cfg_c.value_or<foo>(1.3), 1.3);
+        EXPECT_FLOAT_EQ(std::as_const(cfg).get_or(bax{1.3}).value, 2.2);
+        EXPECT_TRUE((std::same_as<decltype(std::as_const(cfg).get_or(bax{1.3})), bax const &>));
+        EXPECT_EQ(std::as_const(cfg).get_or(foo{"test"}).value, "test");
```
MitraDarja, line 239: We checked before also the value of 1.3, why are we not doing that anymore? (Same question for the other tests.)

test/unit/core/algorithm/configuration_test.cpp
```diff
@@ -222,58 +223,139 @@ TEST(configuration, remove_by_type_template)
...
+TEST(configuration, get_or_by_type_constexpr)
 {
-    seqan3::configuration cfg = bar{1} | foobar<>{std::vector<int>{0, 1, 2, 3}};
+    constexpr seqan3::configuration cfg = bax{2.2} | bar{1};
 
     { // l-value
-        EXPECT_TRUE(ranges::equal(cfg.value_or<foobar>(3.3), std::vector{0, 1, 2, 3}));
-        EXPECT_FLOAT_EQ(cfg.value_or<foo>(1.3), 1.3);
+        constexpr auto element = cfg.get_or(bax{1.3});
+        EXPECT_FLOAT_EQ(element.value, 2.2);
```
MitraDarja, line 265: Why is the type not tested here as done in the tests above? Could we use a typed test here?

test/unit/core/algorithm/configuration_test.cpp
rrahn: Because the semantics changed. Before I could give any alternative, but now the type to look for is deduced by the type of the alternative we pass in. So in this case we use `foo<std::string>` and want to see if any `foo` specialisation is present. If not we return `foo<std::string>`.

test/unit/core/algorithm/configuration_test.cpp
rrahn: Well, it better should be :)

MitraDarja:
Oh, I missed that there are merge conflicts. Should I review then again once there are resolved?

rrahn:
There should not be any major change involved. Usually a merge conflict does not mean that the PR is not ready. But since we are not rebasing until the reviewer gives green light, there might be potential merge conflicts. So I am fine if you are requesting the second reviewer. I will then rebase everything. Unless of course you want to 🙃.

include/seqan3/core/algorithm/configuration.hpp
```diff
@@ -561,6 +473,58 @@ class configuration : public std::tuple<configs_t...>
...
+     *
+     * \details
+     *
+     * Use the type `query_t` to check if such a configuration element is stored in `me`. If this is `true` then
+     * the stored configuration element is returned using perfect forwarding. If this evaluates to `false` the
+     * given alternative is returned. If `query_t` is a class template then it is checked if any
+     * specialisation of this class template is stored.
+     */
+    template <typename this_t, typename query_t, typename alternative_t>
+    static constexpr decltype(auto) get_or_impl(this_t && me,
```
smehringer, line 495: why do we need to pass *this as an argument?

rrahn: In order to safe 8 repetitions of the same interface to handle all possible const and reference qualifications and for template and template template arguments.

smehringer: hm ok, I still don't really understand how this works.

rrahn: From your question it was hard to tell, what you asked for. So I try to explain it a little bit more in detail. At first I was calling a member function of this object. This member function is overloaded for `&`, `const &`, `&&` and `const &&`. So all four combinations of const and reference. These 4 versions had to be repeated for the template template version. Now we would need 8 versions of the same function with identical body. So I was outsourcing them into detail functions. But then I had to repeat all 8 versions with the const and reference. But that didn't safe anything. So I made the functions static and perfectly forwarded `*this` to combine all 4 versions in one function call. Now `this` is either an lvalue reference or rvalue reference. In the first case everything is safe since the object lives outside of the function call. The latter case is also safe since the object is not valid any more on the caller site. Using a static function does not bind the calling method to any instance of the object.

PR #1897, opened 2020-06-12
[MISC] Replace throw statements with static_asserts for search_result.

smehringer:
Fixes seqan/product_backlog#119

Edit (@eseiler):
Note that this works because the member function body is only instantiated/evaluated if the function is actually called at some point in the code.
See https://godbolt.org/z/W25LV3

### DIFF
--- a/include/seqan3/search/search_result.hpp
+++ b/include/seqan3/search/search_result.hpp
@@ -58,9 +58,9 @@ namespace seqan3
  * * seqan3::search_result::reference_id()
  * * seqan3::search_result::reference_begin_pos()
  *
- * Note that the index cursor is not included in a hit by default. If you are trying to access that information,
- * an exception will be thrown. You con configure the result of the search with the output configuration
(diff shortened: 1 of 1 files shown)


include/seqan3/search/search_result.hpp
```diff
@@ -58,9 +58,9 @@ namespace seqan3
  * * seqan3::search_result::reference_id()
  * * seqan3::search_result::reference_begin_pos()
  *
- * Note that the index cursor is not included in a hit by default. If you are trying to access that information,
- * an exception will be thrown. You con configure the result of the search with the output configuration
- * (see seqan3::search_cfg::output).
+ * Note that the index cursor is not included in a hit by default. If you are trying to use the respective member
+ * function, a static_assert will prevent you from doing so. You con configure the result of the search with the output
```
wvandertoorn, line 62:
```suggestion
 * function, a static_assert will prevent you from doing so. You can configure the result of the search with the output
```

PR #1898, opened 2020-06-12
Range sequence concept

smehringer:
Part of seqan/product_backlog#60

### DIFF
--- a/include/seqan3/alphabet/all.hpp
+++ b/include/seqan3/alphabet/all.hpp
@@ -166,6 +166,7 @@
  *
  * We provide specialised containers with certain properties in the \ref range module.
  *
+ * A container over an seqan3::alphabet automatically models the seqan3::sequence concept.
  */
 
 #pragma once
--- a/include/seqan3/range/concept.hpp
+++ b/include/seqan3/range/concept.hpp
@@ -14,7 +14,7 @@
 
 #include <seqan3/std/ranges>
 
-#include <seqan3/core/platform.hpp>
+#include <seqan3/alphabet/concept.hpp>
 
 namespace seqan3
 {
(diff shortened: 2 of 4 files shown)


include/seqan3/range/concept.hpp
```diff
@@ -135,4 +135,23 @@ SEQAN3_CONCEPT pseudo_random_access_range =
     pseudo_random_access_iterator<std::ranges::iterator_t<rng_t>>;
 //!\endcond
 
+/*!\interface seqan3::sequence <>
+ * \brief The generic concept for a sequence.
+ * \ingroup range
+ * \extends std::ranges::input_range
+ *
+ * We define a range over an alphabet as a *sequence*.
+ * Thus, a type models the concept seqan3::sequence is it is at least an std::ranges::input_range
```
eseiler, line 144:
```suggestion
 * Thus, a type models the concept seqan3::sequence if it is at least an std::ranges::input_range
```

include/seqan3/alphabet/all.hpp
```diff
@@ -166,6 +166,7 @@
  *
  * We provide specialised containers with certain properties in the \ref range module.
  *
+ * A container over an alphabet automatically models the seqan3::sequence.
```
rrahn, line 169:
```suggestion
 * A container over an seqan3::alphabet automatically models the seqan3::sequence concept.
```

include/seqan3/range/concept.hpp
```diff
@@ -135,4 +135,23 @@ SEQAN3_CONCEPT pseudo_random_access_range =
     pseudo_random_access_iterator<std::ranges::iterator_t<rng_t>>;
 //!\endcond
 
+/*!\interface seqan3::sequence <>
+ * \brief The generic concept for a sequence.
+ * \ingroup range
+ * \extends std::ranges::input_range
+ *
+ * We define a range over an alphabet as a *sequence*.
```
rrahn, line 143:
```suggestion
 * We define a range over an seqan3::alphabet as a *sequence*.
```

include/seqan3/range/concept.hpp
```diff
@@ -135,4 +135,23 @@ SEQAN3_CONCEPT pseudo_random_access_range =
...
+ * Thus, a type models the concept seqan3::sequence if it is at least an std::ranges::input_range
```
rrahn, line 144:
```suggestion
 * A type models seqan3::sequence if it is at least an std::ranges::input_range
```
And then I would remove the sentence above because they are literally the same.

marehr:
2020-06-15 Core-Meeting

The Problem:

The `sequence` concept name has a different from the `sequence_container` concept. `sequence` does mean a biological sequence and `sequence_container` is a named requirement of the standard (see https://en.cppreference.com/w/cpp/named_req/SequenceContainer).

We decided some time ago that we want to get rid of all the `container` concepts that would solve this problem.

PR #1901, opened 2020-06-13
[FEATURE] add seqan3::detail::iter_pointer[_t]

marehr:
Similar to `std::iter_value_t` we need a `seqan3::detail::iter_pointer` type trait to get the pointer-type of an iterator.

The problem that `std::iterator_traits` will not be defined for some `std::*iterator`'s, because it is only defined (automatically) for legacy input iterators. This means the most generic way to access the value-type of an iterator is `std::iter_value_t`, because `std::iterator_traits::value_type` might not be defined for it.

Normally (C++20) views don't define `pointer`, because it is a thing of the past, but as long as we support older gcc versions we need a universal way to access the pointer type.

### DIFF
--- a/include/seqan3/core/type_traits/iterator.hpp
+++ b/include/seqan3/core/type_traits/iterator.hpp
@@ -167,10 +167,10 @@ struct size_type<it_t>
 template <typename it_t>
 struct iterator_tag
 {
-SEQAN3_DOXYGEN_ONLY(
+#if SEQAN3_DOXYGEN_ONLY(1)0
     //!\brief The [iterator_category](https://en.cppreference.com/w/cpp/iterator/iterator_tags).
     using type = iterator_category;
-)
+#endif
 };
 
 //!\cond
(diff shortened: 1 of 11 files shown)


include/seqan3/core/type_traits/iterator.hpp
```diff
@@ -231,3 +231,45 @@ using iterator_tag_t = typename iterator_tag<it_t>::type;
...
+ * \implements seqan3::transformation_trait
+ * \tparam it_t The type to operate on.
+ * \see seqan3::detail::iter_pointer_t
+ *
+ * \attention
+ * C++20 does not provide a `std::iter_pointer_t`, because the new C++20 iterators do not need to provide a pointer
+ * type.
+ */
+template <typename it_t>
+struct iter_pointer
```
Irallia, line 251: Why don't you call it `pointer_type`?

marehr: In C++20 they follow the following name scheme for member of [`iterator_traits`](https://en.cppreference.com/w/cpp/iterator/iterator_traits):

* `std::iterator_traits::difference_type` -> `std::iter_difference_t`
* `std::iterator_traits::value_type` -> `std::iter_value_t`
* `std::iterator_traits::pointer` -> `seqan3::detail::iter_pointer_t`
* `std::iterator_traits::reference` -> `std::iter_reference_t`
* `std::iterator_traits::iterator_category` -> ???

Generally in C++ a type `<name>_t` is a shortcut for `using <name>_t = typename <name>::type;`

Since the shortcut is named `seqan3::detail::iter_pointer_t`, the struct should be named `seqan3::detail::iter_pointer`.

Irallia: Thanks for the explanation!
I were just looking at `std::iter_value_t`, where it's called `value_type`.

include/seqan3/range/views/single_pass_input.hpp
```diff
@@ -285,7 +277,7 @@ class single_pass_input_iterator<single_pass_input_view<view_type>>
     //!\copydoc operator==
     friend constexpr bool
     operator==(sentinel_type const & s,
-               single_pass_input_iterator<single_pass_input_view<view_type>> const & rhs) noexcept
+               single_pass_input_iterator const & rhs) noexcept
```
smehringer, line 280: fit on line above now? (It wont let me do suggestions)
:nail_care:

smehringer:
Hm, my general comment was lost:

LGTM but your commit history may be a little mixed up: If I don't read it wrong, your "use detail::iter_pointer_t" comes before "add detail::iter_pointer_t" no?

marehr:
> Hm, my general comment was lost:
> 
> LGTM but your commit history may be a little mixed up: If I don't read it wrong, your "use detail::iter_pointer_t" comes before "add detail::iter_pointer_t" no?

No, github just displays it wrongly.

PR #1902, opened 2020-06-13
[FIX] cxx20: std::ranges::copy does not work on any range

marehr:
This is a gcc10.1 regression and is fixed for gcc10.2.

See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95578

### DIFF
--- a/include/seqan3/core/platform.hpp
+++ b/include/seqan3/core/platform.hpp
@@ -239,6 +239,15 @@
 #   endif
 #endif
 
+//!\brief https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95578
+#ifndef SEQAN3_WORKAROUND_GCC_95578 // regressed in gcc10, fixed for gcc10.2
+#   if defined(__GNUC__) && (__GNUC__ == 10 && __GNUC_MINOR__ < 2)
+#       define SEQAN3_WORKAROUND_GCC_95578 1
+#   else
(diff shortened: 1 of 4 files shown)


include/seqan3/core/platform.hpp
```diff
@@ -230,6 +230,15 @@
 #   endif
 #endif
 
+//!\brief https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95578
+#ifndef SEQAN3_WORKAROUND_GCC_95578
+#   if defined(__GNUC__) && (__GNUC__ == 10)
```
marehr, line 235:
```suggestion
#ifndef SEQAN3_WORKAROUND_GCC_95578 // regressed in gcc10, fixed for gcc10.2
#   if defined(__GNUC__) && (__GNUC__ == 10 && __GNUC_MINOR__ < 2)
```

test/include/seqan3/test/expect_range_eq.hpp
```diff
@@ -33,7 +33,12 @@ struct expect_range_eq
     {
         using value_t = std::ranges::range_value_t<rng_t>;
         std::vector<value_t> rng_copy{};
+#if SEQAN3_WORKAROUND_GCC_95578
+        for (auto && v: rng)
```
rrahn, line 37:
```suggestion
        for (auto && v : rng)
```

PR #1904, opened 2020-06-13
[NOP] views/rank_to: code clean-up

marehr:


smehringer:
Ignoring Jenkins because there are issues in Tuebingen

PR #1905, opened 2020-06-13
[FIX] use std::ranges::ref_view instead of ranges one

marehr:
Part of https://github.com/seqan/product_backlog/issues/124

### DIFF
--- a/include/seqan3/range/views/async_input_buffer.hpp
+++ b/include/seqan3/range/views/async_input_buffer.hpp
@@ -119,7 +119,7 @@ public:
     //!\cond
     requires (!std::same_as<remove_cvref_t<other_urng_t>, async_input_buffer_view>) && // prevent recursive instantiation
              std::ranges::viewable_range<other_urng_t> &&
-             std::constructible_from<urng_t, ranges::ref_view<std::remove_reference_t<other_urng_t>>>
+             std::constructible_from<urng_t, std::ranges::ref_view<std::remove_reference_t<other_urng_t>>>
     //!\endcond
     async_input_buffer_view(other_urng_t && _urng, size_t const buffer_size) :
         async_input_buffer_view{std::views::all(_urng), buffer_size}
--- a/include/seqan3/range/views/enforce_random_access.hpp
+++ b/include/seqan3/range/views/enforce_random_access.hpp
@@ -69,7 +69,7 @@ public:
     //!\cond
      requires (!std::same_as<remove_cvref_t<viewable_rng_t>, view_enforce_random_access>) &&
               std::ranges::viewable_range<viewable_rng_t> &&
-              std::constructible_from<urng_t, ranges::ref_view<std::remove_reference_t<viewable_rng_t>>>
+              std::constructible_from<urng_t, std::ranges::ref_view<std::remove_reference_t<viewable_rng_t>>>
     //!\endcond
     explicit constexpr view_enforce_random_access(viewable_rng_t range) :
         view_enforce_random_access{std::views::all(range)}
--- a/include/seqan3/range/views/kmer_hash.hpp
+++ b/include/seqan3/range/views/kmer_hash.hpp
@@ -84,7 +84,7 @@ public:
     //!\cond
      requires (!std::same_as<remove_cvref_t<rng_t>, kmer_hash_view>) &&
               std::ranges::viewable_range<rng_t> &&
-              std::constructible_from<urng_t, ranges::ref_view<std::remove_reference_t<rng_t>>>
+              std::constructible_from<urng_t, std::ranges::ref_view<std::remove_reference_t<rng_t>>>
     //!\endcond
     kmer_hash_view(rng_t && urange_, shape const & s_) :
         urange{std::views::all(std::forward<rng_t>(urange_))}, shape_{s_}
--- a/include/seqan3/range/views/minimiser.hpp
+++ b/include/seqan3/range/views/minimiser.hpp
@@ -77,7 +77,7 @@ public:
     template <typename rng_t>
     //!\cond
      requires std::ranges::viewable_range<rng_t> &&
-              std::constructible_from<urng_t, ranges::ref_view<std::remove_reference_t<rng_t>>>
+              std::constructible_from<urng_t, std::ranges::ref_view<std::remove_reference_t<rng_t>>>
     //!\endcond
     minimiser_view(rng_t && urange_, size_t const w_) :
     urange{std::views::all(std::forward<rng_t>(urange_))}, window_values_size{w_} {}
--- a/include/seqan3/range/views/pairwise_combine.hpp
+++ b/include/seqan3/range/views/pairwise_combine.hpp
@@ -535,7 +535,8 @@ public:
     //!\cond
         requires (!std::same_as<remove_cvref_t<other_range_t>, pairwise_combine_view>) &&
                  std::ranges::viewable_range<other_range_t> &&  // Must come after self type check to avoid conflicts with the move constructor.
-                 std::constructible_from<underlying_range_type, ranges::ref_view<std::remove_reference_t<other_range_t>>>
+                 std::constructible_from<underlying_range_type,
+                                         std::ranges::ref_view<std::remove_reference_t<other_range_t>>>
             //TODO: Investigate: the following expression is equivalent to the one above but raises a weird assertion in
             //      the ranges adaptor suggesting that the pairwise_combine_view is not a viewable_range.
             //      std::constructible_from<underlying_range_type, decltype(std::views::all(std::declval<other_range_t &&>()))>
--- a/include/seqan3/range/views/single_pass_input.hpp
+++ b/include/seqan3/range/views/single_pass_input.hpp
@@ -104,7 +104,7 @@ public:
     //!\cond
     requires (!std::same_as<remove_cvref_t<other_urng_t>, single_pass_input_view> &&
               std::ranges::viewable_range<other_urng_t> &&  // Must come after self type check to avoid conflicts with the move constructor.
-              std::constructible_from<urng_t, ranges::ref_view<std::remove_reference_t<other_urng_t>>>)
+              std::constructible_from<urng_t, std::ranges::ref_view<std::remove_reference_t<other_urng_t>>>)
     //!\endcond
     explicit single_pass_input_view(other_urng_t && _urng) :
         single_pass_input_view{std::views::all(_urng)}
--- a/include/seqan3/range/views/translate.hpp
+++ b/include/seqan3/range/views/translate.hpp
@@ -225,7 +225,7 @@ public:
     //!\cond
      requires (!std::same_as<remove_cvref_t<rng_t>, view_translate_single>) &&
               std::ranges::viewable_range<rng_t> &&
-              std::constructible_from<urng_t, ranges::ref_view<std::remove_reference_t<rng_t>>>
+              std::constructible_from<urng_t, std::ranges::ref_view<std::remove_reference_t<rng_t>>>
     //!\endcond
     view_translate_single(rng_t && _urange, translation_frames const _tf = translation_frames::FWD_FRAME_0)
      : view_translate_single{std::views::all(std::forward<rng_t>(_urange)), _tf}
@@ -612,7 +612,7 @@ public:
     //!\cond
         requires (!std::same_as<remove_cvref_t<rng_t>, view_translate>) &&
                  std::ranges::viewable_range<rng_t> &&
-                 std::constructible_from<urng_t, ranges::ref_view<std::remove_reference_t<rng_t>>>
+                 std::constructible_from<urng_t, std::ranges::ref_view<std::remove_reference_t<rng_t>>>
     //!\endcond
     view_translate(rng_t && _urange, translation_frames const _tf = translation_frames::SIX_FRAME)
      : view_translate{std::views::all(std::forward<rng_t>(_urange)), _tf}
--- a/include/seqan3/range/views/translate_join.hpp
+++ b/include/seqan3/range/views/translate_join.hpp
@@ -134,7 +134,7 @@ public:
     //!\cond
         requires (!std::same_as<remove_cvref_t<rng_t>, view_translate_join>) &&
                  std::ranges::viewable_range<rng_t> &&
-                 std::constructible_from<urng_t, ranges::ref_view<std::remove_reference_t<rng_t>>>
+                 std::constructible_from<urng_t, std::ranges::ref_view<std::remove_reference_t<rng_t>>>
     //!\endcond
     view_translate_join(rng_t && _urange, translation_frames const _tf = translation_frames::SIX_FRAME)
      : view_translate_join{std::views::all(std::forward<rng_t>(_urange)), _tf}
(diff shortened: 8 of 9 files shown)


include/seqan3/range/views/pairwise_combine.hpp
```diff
@@ -535,7 +535,7 @@ class pairwise_combine_view : public std::ranges::view_interface<pairwise_combin
     //!\cond
         requires (!std::same_as<remove_cvref_t<other_range_t>, pairwise_combine_view>) &&
                  std::ranges::viewable_range<other_range_t> &&  // Must come after self type check to avoid conflicts with the move constructor.
-                 std::constructible_from<underlying_range_type, ranges::ref_view<std::remove_reference_t<other_range_t>>>
+                 std::constructible_from<underlying_range_type, std::ranges::ref_view<std::remove_reference_t<other_range_t>>>
```
eseiler, line 538: Line length.
The line was a;ready too long before this PR, but maybe you can fix it now.

include/seqan3/std/ranges
```diff
@@ -172,7 +172,7 @@ using ::ranges::cpp20::views::all;
 template<viewable_range R>
 using all_t = ::ranges::cpp20::all_view<R>;
 } // namespace views
-// using ::ranges::cpp20::ref_view;
+using ::ranges::cpp20::ref_view;
```
eseiler, line 175: Does this also/Can we  disable the usage of `ranges::ref_view`?

marehr: Not sure what you mean?

You mean that we forcefully disallow the usage of `ranges::ref_view`?

I don't know how, maybe you have an idea.

include/seqan3/range/views/pairwise_combine.hpp
marehr, line 538:
```suggestion
                 std::constructible_from<underlying_range_type,
                                         std::ranges::ref_view<std::remove_reference_t<other_range_t>>>
```

include/seqan3/range/views/async_input_buffer.hpp
```diff
@@ -119,7 +119,7 @@ class async_input_buffer_view : public std::ranges::view_interface<async_input_b
     //!\cond
     requires (!std::same_as<remove_cvref_t<other_urng_t>, async_input_buffer_view>) && // prevent recursive instantiation
              std::ranges::viewable_range<other_urng_t> &&
-             std::constructible_from<urng_t, ranges::ref_view<std::remove_reference_t<other_urng_t>>>
+             std::constructible_from<urng_t, std::ranges::ref_view<std::remove_reference_t<other_urng_t>>>
```
rrahn, line 122:
```suggestion
             std::constructible_from<urng_t, std::views::all_t<other_urng_t>>
```

include/seqan3/range/views/enforce_random_access.hpp
```diff
@@ -69,7 +69,7 @@ class view_enforce_random_access : public std::ranges::view_interface<view_enfor
     //!\cond
      requires (!std::same_as<remove_cvref_t<viewable_rng_t>, view_enforce_random_access>) &&
               std::ranges::viewable_range<viewable_rng_t> &&
-              std::constructible_from<urng_t, ranges::ref_view<std::remove_reference_t<viewable_rng_t>>>
+              std::constructible_from<urng_t, std::ranges::ref_view<std::remove_reference_t<viewable_rng_t>>>
```
rrahn, line 72:
```suggestion
              std::constructible_from<urng_t, std::views::all_t<viewable_rng_t>>
```

include/seqan3/range/views/kmer_hash.hpp
```diff
@@ -84,7 +84,7 @@ class kmer_hash_view : public std::ranges::view_interface<kmer_hash_view<urng_t>
     //!\cond
      requires (!std::same_as<remove_cvref_t<rng_t>, kmer_hash_view>) &&
               std::ranges::viewable_range<rng_t> &&
-              std::constructible_from<urng_t, ranges::ref_view<std::remove_reference_t<rng_t>>>
+              std::constructible_from<urng_t, std::ranges::ref_view<std::remove_reference_t<rng_t>>>
```
rrahn, line 87:
```suggestion
              std::constructible_from<urng_t, std::views::all_t<rng_t>>
```

include/seqan3/range/views/minimiser.hpp
```diff
@@ -77,7 +77,7 @@ class minimiser_view : public std::ranges::view_interface<minimiser_view<urng_t>
     template <typename rng_t>
     //!\cond
      requires std::ranges::viewable_range<rng_t> &&
-              std::constructible_from<urng_t, ranges::ref_view<std::remove_reference_t<rng_t>>>
+              std::constructible_from<urng_t, std::ranges::ref_view<std::remove_reference_t<rng_t>>>
```
rrahn, line 80:
```suggestion
              std::constructible_from<urng_t, std::views::all_t<rng_t>>
```

include/seqan3/range/views/pairwise_combine.hpp
```diff
@@ -535,7 +535,8 @@ class pairwise_combine_view : public std::ranges::view_interface<pairwise_combin
     //!\cond
         requires (!std::same_as<remove_cvref_t<other_range_t>, pairwise_combine_view>) &&
                  std::ranges::viewable_range<other_range_t> &&  // Must come after self type check to avoid conflicts with the move constructor.
-                 std::constructible_from<underlying_range_type, ranges::ref_view<std::remove_reference_t<other_range_t>>>
+                 std::constructible_from<underlying_range_type,
+                                         std::ranges::ref_view<std::remove_reference_t<other_range_t>>>
```
rrahn, line 539:
```suggestion
                 std::constructible_from<underlying_range_type, std::views::all_t<other_range_t>>
```

include/seqan3/range/views/single_pass_input.hpp
```diff
@@ -104,7 +104,7 @@ class single_pass_input_view : public std::ranges::view_interface<single_pass_in
     //!\cond
     requires (!std::same_as<remove_cvref_t<other_urng_t>, single_pass_input_view> &&
               std::ranges::viewable_range<other_urng_t> &&  // Must come after self type check to avoid conflicts with the move constructor.
-              std::constructible_from<urng_t, ranges::ref_view<std::remove_reference_t<other_urng_t>>>)
+              std::constructible_from<urng_t, std::ranges::ref_view<std::remove_reference_t<other_urng_t>>>)
```
rrahn, line 107:
```suggestion
              std::constructible_from<urng_t, std::views::all_t<other_urng_t>>)
```

include/seqan3/range/views/translate.hpp
```diff
@@ -225,7 +225,7 @@ class view_translate_single : public ranges::view_base
     //!\cond
      requires (!std::same_as<remove_cvref_t<rng_t>, view_translate_single>) &&
               std::ranges::viewable_range<rng_t> &&
-              std::constructible_from<urng_t, ranges::ref_view<std::remove_reference_t<rng_t>>>
+              std::constructible_from<urng_t, std::ranges::ref_view<std::remove_reference_t<rng_t>>>
```
rrahn, line 228:
```suggestion
              std::constructible_from<urng_t, std::views::all_t<rng_t>>
```

include/seqan3/range/views/translate.hpp
```diff
@@ -612,7 +612,7 @@ class view_translate : public ranges::view_base
     //!\cond
         requires (!std::same_as<remove_cvref_t<rng_t>, view_translate>) &&
                  std::ranges::viewable_range<rng_t> &&
-                 std::constructible_from<urng_t, ranges::ref_view<std::remove_reference_t<rng_t>>>
+                 std::constructible_from<urng_t, std::ranges::ref_view<std::remove_reference_t<rng_t>>>
```
rrahn, line 615:
```suggestion
                 std::constructible_from<urng_t, std::views::all_t<rng_t>>
```

include/seqan3/range/views/translate_join.hpp
```diff
@@ -134,7 +134,7 @@ class view_translate_join : public ranges::view_base
     //!\cond
         requires (!std::same_as<remove_cvref_t<rng_t>, view_translate_join>) &&
                  std::ranges::viewable_range<rng_t> &&
-                 std::constructible_from<urng_t, ranges::ref_view<std::remove_reference_t<rng_t>>>
+                 std::constructible_from<urng_t, std::ranges::ref_view<std::remove_reference_t<rng_t>>>
```
rrahn, line 137:
```suggestion
                 std::constructible_from<urng_t, std::views::all_t<rng_t>>
```

marehr:
> I am not sure. We have this ref_view everywhere, and I feel it will break something when we use std::views::all_t here. But that would be the correct thing. Since the parameter can be a viewable_range it means it can be a view. If it is a view, then `ref_view<view_t>` is valid but not the same type as what `std::views::all` returns. I know, that your PR wants to fix something else, but maybe we can make this correct already, since it touches all our views. But only, if this just works immediately. if this causes some hard to track errors, then we need to reevaluate this later with another ticket.

No, I will figure out what the problem is :)

You are right that this is right now wrong and should be fixed.

marehr:
> But only, if this just works immediately. if this causes some hard to track errors, then we need to reevaluate this later with another ticket.

I would want to revert it to the original, because I ran into some infinite recursions with <= gcc-9. gcc-10 could build this.

---

After looking at it a bit, I think it was a "correct" way to create those views.

Take `single_pass_input_view` as an example:

https://github.com/seqan/seqan3/blob/e5fc01fd97acbd574c55f591d080f637b172d70f/include/seqan3/range/views/single_pass_input.hpp#L37-L38

Says it requires a **view**. The name `urng_t` is misleading in this case, it should be called `uview_t`.

https://github.com/seqan/seqan3/blob/e5fc01fd97acbd574c55f591d080f637b172d70f/include/seqan3/range/views/single_pass_input.hpp#L155-L157

The type deduction ensures that `single_pass_input_view` gets a view as `urng_t`.

`std::views::all_t` is either the original view `= urng_t` (if it was a view) or `std::ranges::ref_view<urng_t>` if it was only a `std::ranges::viewable_range`.

https://github.com/seqan/seqan3/blob/e5fc01fd97acbd574c55f591d080f637b172d70f/include/seqan3/range/views/single_pass_input.hpp#L98-L100

Ensures that the original view (if it was a view) can be used to construct `single_pass_input_view`. Or a `std::ranges::ref_view<urng_t>` instance.

We could remove this constructor if we made sure that the adaptor of `single_pass_input_view` uses `std::views::all(std::forward(urange))` when constructing `single_pass_input_view`:

https://github.com/seqan/seqan3/blob/e5fc01fd97acbd574c55f591d080f637b172d70f/include/seqan3/range/views/single_pass_input.hpp#L102-L111

But, we can also keep it to allow `single_pass_input_view{std::forward(urange)}`.

This constructor does the following:
* if `urange` is a view it will be disabled and allows to use the other constructor above
* if `urange` is not a view (the other constructor above will be disabled, because this one is more specialised) and it will wrap the `urange` in `std::ranges::ref_view<urng_t>`, which would also be the result of `std::views::all(urange)`.

This semantic is not obvious at all, because it heavily relies on the property that

```c++
std::constructible_from<urng_t, std::ranges::ref_view<other_urng_t>>
``` 

will only be true if `other_urng_t` = original `urng_t'` is **NOT** a view.

Reason:

1. If original `urng_t'` (from the type deduction) is a view then
   ```c++
   other_urng_t = urng_t' [cvref]

   // and 

   urng_t = std::views::all_t<urng_t' [cvref]> // from the type deduction
          = urng_t' [cvref] // because it is a view

   // and

   urng_t' [cvref] != std::ranges::ref_view<urng_t' [cvref]> // types mismatch


   // therefore

      std::constructible_from<urng_t, std::ranges::ref_view<other_urng_t>>
   => std::constructible_from<urng_t' [cvref], std::ranges::ref_view<urng_t' [cvref]>> // urng_t = urng_t' [cvref] = other_urng_t
   => false // types mismatch and ref_view has no conversion operator to urng_t' => not constructible
   ```
   (except `urng_t' [cvref]` would have such a constructor, but that seems illogical?!)

   Since the types mismatch, this constructor overload is not viable.
2. If original `urng_t'` (from the type deduction) is not a view, but viewable then
   ```c++
   other_urng_t = urng_t' [const] &
   
   // and
   
   urng_t = std::views::all_t<urng_t' [const] &> // from the type deduction
          = std::ranges::ref_view<urng_t' [const] &> // because it is NOT a view


   // thus
   
      std::constructible_from<urng_t, std::ranges::ref_view<other_urng_t>>
   => std::constructible_from<std::ranges::ref_view<urng_t' [const] &>, std::ranges::ref_view<urng_t' [const] &>>
   => true
   ```

   The first constructor is also a valid constructor, because

   ```c++
      explicit single_pass_input_view(urng_t _urng)
   => explicit single_pass_input_view(std::ranges::ref_view<urng_t' [const] &> _urng)

   // and

   std::ranges::ref_view<urng_t' [const] &>{urange} // is a valid expression
   ```

   But I think it is not considered in overload resolution, because `urng_t = std::ranges::ref_view<urng_t' [const] &>` is a direct match for the `explicit single_pass_input_view(other_urng_t && _urng)` constructor, whereas the first constructor would only be considered if the direct matches are not applicable and conversion sequences would be considered.
3. What would happen if we call `single_pass_input_view{std::views::all(urange)}`?

   If `urange` is a view, case (1) applies. Otherwise

   ```c++
   other_urng_t = std::ranges::ref_view<urng_t' [const] &> // because we pass std::views::all(urange) in

   // and
   
   urng_t = std::ranges::ref_view<urng_t' [const] &>

   
   // thus
      std::constructible_from<urng_t, std::ranges::ref_view<other_urng_t>>
   => std::constructible_from<std::ranges::ref_view<urng_t' [const] &>, std::ranges::ref_view<std::ranges::ref_view<urng_t' [const] &>>>
   => false
   ```

   Since `other_urng_t == urng_t`, the first constructor is active.

This is REALLY hard to understand and it bakes the question of how we could simplify that. This also relies on the fact that `std::constructible_from<urng_t' [cvref], std::ranges::ref_view<urng_t' [cvref]>>` is always a `false` expression.

I think it would be safe to replace `std::constructible_from` with `std::same_as`, but this would be still very hard to understand.

marehr:
https://eel.is/c++draft/range.transform

Does not define a second constructor at all, it relies on the fact that 

```c++
constexpr transform_view(V base, F fun);

and

template<class R, class F>
transform_view(R&&, F) -> transform_view<views::all_t<R>, F>;
```

Tries to implicitly convert a `R` which is a `viewable_range` and not a `view` to a `views::all_t<R> == views::ref_view<R>` which makes this REALLY easy to write (only one constructor needed) and semi-hard to understand, which would be fine, because we can say, the standard does this too.

rrahn:
> Tries to implicitly convert a R which is a viewable_range and not a view to a views::all_t<R> == views::ref_view<R> which makes this REALLY easy to write (only one constructor needed) and semi-hard to understand, which would be fine, because we can say: "the standard does this too".

I know. And I was always wondering why we needed this complicated construction thing. So the resolution also for me would be to change this for good by following the standard approach.

rrahn:
And many thanks for this detailed exploration of the problem. I remembered that there was a problem but I hoped it was due to ranges stuff and no longer present.

PR #1906, opened 2020-06-13
[FIX] use std::ranges::view_base instead of ranges one

marehr:
Part of https://github.com/seqan/product_backlog/issues/124


smehringer:
whats up with the code coverage?

marehr:
> whats up with the code coverage?

I don't quite understand your question :question: 

The code is still the same for all gcc version except for `gcc-10 -std=c++2a` (because this one uses the standard library).

PR #1907, opened 2020-06-13
Optimise alignment/part 4 result builder

rrahn:
Adds result builder policy to new alignment result implementation.
This will replace the result value mechanism as we used it before in the future.

part of seqan/product_backlog#74

### DIFF
--- a/include/seqan3/alignment/pairwise/alignment_result.hpp
+++ b/include/seqan3/alignment/pairwise/alignment_result.hpp
@@ -15,10 +15,20 @@
 
 #include <optional>
 
+#include <seqan3/core/algorithm/configuration.hpp>
 #include <seqan3/core/type_traits/template_inspection.hpp>
 
 namespace seqan3::detail
 {
+
+// forward declaration for friend declaraion in alignment_result.
+template <typename>
+#if !SEQAN3_WORKAROUND_GCC_93467
+//!\cond
+    requires is_type_specialisation_of_v<configuration_t, configuration>
+//!\endcond
+#endif // !SEQAN3_WORKAROUND_GCC_93467
+class policy_alignment_result_builder;
 
 /*!\brief A struct that contains the actual alignment result data.
  * \ingroup pairwise_alignment
@@ -114,6 +124,8 @@ namespace seqan3
  * \if DEV
  * To access the type of the passed alignment result value use the seqan3::detail::alignment_result_value_type_accessor
  * transformation trait.
+ * The seqan3::detail::policy_alignment_result_builder is used to set the respective
+ * values. This class is befriended with this class to allow access to the private data member.
  * \endif
  */
 template <typename alignment_result_value_t>
@@ -131,16 +143,25 @@ private:
      * \{
      */
     //! \brief The type for the alignment identifier.
-    using id_t          = decltype(data.id);
+    using id_t = decltype(data.id);
     //! \brief The type for the resulting score.
-    using score_t       = decltype(data.score);
+    using score_t = decltype(data.score);
     //! \brief The type for the back coordinate.
-    using back_coord_t   = decltype(data.back_coordinate);
+    using back_coord_t = decltype(data.back_coordinate);
     //! \brief The type for the front coordinate.
     using front_coord_t = decltype(data.front_coordinate);
     //! \brief The type for the alignment.
-    using alignment_t   = decltype(data.alignment);
+    using alignment_t = decltype(data.alignment);
     //!\}
+
+    //!\brief Befriend alignment result builder.
+    template <typename configuration_t>
+    #if !SEQAN3_WORKAROUND_GCC_93467
+    //!\cond
+        requires is_type_specialisation_of_v<configuration_t, configuration>
+    //!\endcond
+    #endif // !SEQAN3_WORKAROUND_GCC_93467
+    friend class detail::policy_alignment_result_builder;
 
 public:
     /*!\name Constructors, destructor and assignment
--- a/include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
+++ b/include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
@@ -139,15 +139,15 @@ public:
     //!\endcond
     void operator()(indexed_sequence_pairs_t && indexed_sequence_pairs, callback_t && callback)
     {
-        using result_value_t = typename alignment_result_value_type_accessor<alignment_result_type>::type;
         using std::get;
 
         for (auto && [sequence_pair, idx] : indexed_sequence_pairs)
         {
-            result_value_t res{};
-            res.id = idx;
-            res.score = compute_matrix(get<0>(sequence_pair), get<1>(sequence_pair));
-            callback(alignment_result_type{res});
+            compute_matrix(get<0>(sequence_pair), get<1>(sequence_pair));
+            this->make_result_and_invoke(std::forward<decltype(sequence_pair)>(sequence_pair),
+                                         std::move(idx),
+                                         this->tracked_optimum(),
+                                         callback);
         }
     }
 
@@ -160,7 +160,7 @@ protected:
      * \param[in] sequence2 The second sequence to compute the alignment for.
      */
     template <std::ranges::forward_range sequence1_t, std::ranges::forward_range sequence2_t>
-    int32_t compute_matrix(sequence1_t && sequence1, sequence2_t && sequence2)
+    void compute_matrix(sequence1_t && sequence1, sequence2_t && sequence2)
     {
         // ---------------------------------------------------------------------
         // Initialisation phase: allocate memory and initialise first column.
@@ -205,8 +205,6 @@ protected:
             this->track_last_column_cell(*++alignment_column_it, *++cell_index_column_it);
 
         this->track_final_cell(*alignment_column_it, *cell_index_column_it);
-
-        return this->tracked_optimum();
     }
 
     /*!\brief Initialise the first column of the alignment matrix.
--- a/include/seqan3/alignment/pairwise/detail/policy_alignment_result_builder.hpp
+++ b/include/seqan3/alignment/pairwise/detail/policy_alignment_result_builder.hpp
@@ -1,0 +1,116 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2020, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides seqan3::detail::policy_alignment_result_builder.
+ * \author Rene Rahn <rene.rahn AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <seqan3/alignment/pairwise/detail/type_traits.hpp>
+#include <seqan3/core/algorithm/configuration.hpp>
+#include <seqan3/core/detail/empty_type.hpp>
+#include <seqan3/core/type_traits/template_inspection.hpp>
+
+namespace seqan3::detail
+{
+
+/*!\brief Implements the alignment result builder.
+ * \ingroup pairwise_alignment
+ *
+ * \tparam alignment_configuration_t The type of the alignment configuration; must be a type specialisation of
+ *                                   seqan3::configuration.
+ *
+ * \details
+ *
+ * Implements the interfaces to build the alignment result based on the previously selected output configurations.
+ */
+template <typename alignment_configuration_t>
+#if !SEQAN3_WORKAROUND_GCC_93467
+//!\cond
+    requires is_type_specialisation_of_v<alignment_configuration_t, configuration>
+//!\endcond
+#endif // !SEQAN3_WORKAROUND_GCC_93467
+class policy_alignment_result_builder
+{
+protected:
+    //!\brief The configuration traits type.
+    using traits_type = alignment_configuration_traits<alignment_configuration_t>;
+    //!\brief The alignment result type.
+    using result_type = typename traits_type::alignment_result_type;
+
+    static_assert(!std::same_as<result_type, empty_type>, "The alignment result type was not configured.");
+
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    policy_alignment_result_builder() = default; //!< Defaulted.
+    policy_alignment_result_builder(policy_alignment_result_builder const &) = default; //!< Defaulted.
+    policy_alignment_result_builder(policy_alignment_result_builder &&) = default; //!< Defaulted.
+    policy_alignment_result_builder & operator=(policy_alignment_result_builder const &) = default; //!< Defaulted.
+    policy_alignment_result_builder & operator=(policy_alignment_result_builder &&) = default; //!< Defaulted.
+    ~policy_alignment_result_builder() = default; //!< Defaulted.
+
+    /*!\brief Construction and initialisation using the alignment configuration.
+     * \param[in] config The alignment configuration [not used in this context].
+     */
+    policy_alignment_result_builder(alignment_configuration_t const & SEQAN3_DOXYGEN_ONLY(config))
+    {}
+    //!\}
+
+    /*!\brief Builds the seqan3::alignment_result based on the given alignment result type and then invokes the
+     *        given callable with the result.
+     *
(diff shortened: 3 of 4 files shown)


include/seqan3/alignment/pairwise/alignment_result.hpp
```diff
@@ -114,6 +124,8 @@ namespace seqan3
  * \if DEV
  * To access the type of the passed alignment result value use the seqan3::detail::alignment_result_value_type_accessor
  * transformation trait.
+ * In the new implementation the seqan3::detail::policy_alignment_result_builder is used to set the respective
```
marehr, line 127: How will this comment age? How long will this be "new"?

include/seqan3/alignment/pairwise/alignment_result.hpp
```diff
@@ -15,11 +15,21 @@
...
 {
 
+// forward declaration for friend declaraion in alignment_result.
+template <typename>
+#if !SEQAN3_WORKAROUND_GCC_93467
+//!\cond
+    requires is_type_specialisation_of_v<configuration_t, configuration>
+//!\endcond
+#endif // !SEQAN3_WORKAROUND_GCC_93467
+class policy_alignment_result_builder;
```
marehr, line 31: EDIT: I think I misread your intention. You want to add this constraint, but gcc doesn't let you, right?

---

I'm not sure if this really a bug, because forward declaring something should have the same template constraints, but let's see what the consensus of clang, gcc ans msvc is:

https://godbolt.org/z/H4FfSC

I would expect that a forward declare must have the same constraints, otherwise you could get inconsistencies when using that forward declared struct.

The gcc [ticket 93467](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=93467) quotes the standard:

> "A constrained declaration may only be redeclared using the same syntactic form. No diagnostic is required."

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
```diff
@@ -139,15 +139,13 @@ class pairwise_alignment_algorithm : protected policies_t...
...
     void operator()(indexed_sequence_pairs_t && indexed_sequence_pairs, callback_t && callback)
     {
-        using result_value_t = typename alignment_result_value_type_accessor<alignment_result_type>::type;
         using std::get;
 
-        for (auto && [sequence_pair, idx] : indexed_sequence_pairs)
+        for (auto && indexed_sequence_pair : indexed_sequence_pairs)
         {
-            result_value_t res{};
-            res.id = idx;
```
marehr, line 148: What happens to the id?

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
```diff
@@ -139,15 +139,13 @@ class pairwise_alignment_algorithm : protected policies_t...
...
-            res.score = compute_matrix(get<0>(sequence_pair), get<1>(sequence_pair));
-            callback(alignment_result_type{res});
+            auto && sequence_pair = get<0>(indexed_sequence_pair);
```
marehr, line 146:
```suggestion
            auto & sequence_pair = get<0>(indexed_sequence_pair);
```

Isn't this guaranteed? Because we call it with lvalue `indexed_sequence_pair`?

include/seqan3/alignment/pairwise/detail/policy_alignment_result_builder.hpp
```diff
@@ -0,0 +1,111 @@
...
+    policy_alignment_result_builder(alignment_configuration_t const & SEQAN3_DOXYGEN_ONLY(config))
+    {}
+    //!\}
+
+    /*!\brief Builds the seqan3::alignment_result based on the given alignment result type and then invokes the
+     *        given callable with the result.
+     *
+     * \tparam indexed_sequence_pair_t The type of the indexed sequence pair.
+     * \tparam score_t The type of the score.
+     * \tparam callback_t The type of the callack to invoke.
```
marehr, line 71:
```suggestion
     * \tparam callback_t The type of the callback to invoke.
```

include/seqan3/alignment/pairwise/detail/policy_alignment_result_builder.hpp
```diff
@@ -0,0 +1,111 @@
...
+     *
+     * \param[in] indexed_sequence_pair The indexed sequence pair [maybe unused].
+     * \param[in] score The best alignment score [maybe unused].
```
marehr, line 74:
```suggestion
     * \param[in] indexed_sequence_pair The indexed sequence pair.
     * \param[in] score The best alignment score.
```

We don't use it anywhere. Why should a user care if we may not use that parameter, he has to provide that anyway to call it.

include/seqan3/alignment/pairwise/detail/policy_alignment_result_builder.hpp
```diff
@@ -0,0 +1,111 @@
...
+                               callback_t && callback)
+    {
+        using std::get;
+        using invalid_t = std::nullopt_t *;
+
+        result_type result{};
+
+        static_assert(!std::same_as<decltype(result.data.id), invalid_t>,
+                      "Invalid configuration. Expected result with id!");
+        result.data.id = std::move(get<1>(indexed_sequence_pair));
```
marehr, line 99: ? `indexed_sequence_pair` is used?

include/seqan3/alignment/pairwise/detail/policy_alignment_result_builder.hpp
```diff
@@ -0,0 +1,111 @@
...
+     * Generates a seqan3::alignment_result object with the results computed during the alignment. Depending on the
+     * seqan3::align_cfg::result configuration only the requested values are stored. In some cases some additional
+     * work is done to generate the requested result. For example computing the associated alignment from the traceback
+     * matrix.
+     */
+    template <typename indexed_sequence_pair_t, typename score_t, typename callback_t>
+    //!\cond
+        requires std::invocable<callback_t, result_type>
+    //!\endcond
+    void build_result_and_send([[maybe_unused]] indexed_sequence_pair_t && indexed_sequence_pair,
```
marehr, line 88:
```suggestion
    void make_result_and_invoke([[maybe_unused]] indexed_sequence_pair_t && indexed_sequence_pair,
```

?

include/seqan3/alignment/pairwise/alignment_result.hpp
rrahn: Until the old implementation can be removed completely.

include/seqan3/alignment/pairwise/alignment_result.hpp
rrahn: Ah, the edit came after your comment. Now it makes sense. Right, gcc doesn't let me although the standard requires it.

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
rrahn: it is set later in the builder policy

include/seqan3/alignment/pairwise/detail/policy_alignment_result_builder.hpp
marehr: Why is this a single function? Can't we just call this `make_result` that returns a `result_type` and the callback is called within the loop?

marehr: Or is the idea that you might have multiple results and this handles them?

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
rrahn: Not exactly: If std::tuple_element_t<0, pair_t> == something && then it would be binding a lvalue reference (after applying reference collapsing rules) to something that was an rvalue reference. Would that make sense?

include/seqan3/alignment/pairwise/detail/policy_alignment_result_builder.hpp
rrahn: 🤠

include/seqan3/alignment/pairwise/detail/policy_alignment_result_builder.hpp
rrahn: right, I was referring to the sequences and forgot that I extract the ID already.

include/seqan3/alignment/pairwise/detail/policy_alignment_result_builder.hpp
rrahn: Don't care. Had this also as an alternative.

include/seqan3/alignment/pairwise/detail/policy_alignment_result_builder.hpp
rrahn: Ah no, I was referring to the future here already since we said we want to have the user decide what to get and what not. So it might happen that the sequences and id are not used.

rrahn: I would keep it then, ok?

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
marehr: The problem I have with `&&` is that we might move data out, the following call gets the tuple again, so that means me pass a tuple that might have moved out data.

include/seqan3/alignment/pairwise/detail/policy_alignment_result_builder.hpp
rrahn: > Or is the idea that you might have multiple results and this handles them?

Yes, in the simd case I need to loop over the scores and possibly generate the alignment for every simd sequence. If I make it a single function then logic of building the result is in different parts of the code, which does not make sense.

marehr: I would prefer the `make_result`, because it is more close to the current implementation.

rrahn: the name?

include/seqan3/alignment/pairwise/detail/policy_alignment_result_builder.hpp
```diff
@@ -0,0 +1,111 @@
...
+     * Generates a seqan3::alignment_result object with the results computed during the alignment. Depending on the
+     * seqan3::align_cfg::result configuration only the requested values are stored. In some cases some additional
+     * work is done to generate the requested result. For example computing the associated alignment from the traceback
+     * matrix.
+     */
+    template <typename indexed_sequence_pair_t, typename score_t, typename callback_t>
+    //!\cond
+        requires std::invocable<callback_t, result_type>
+    //!\endcond
+    void build_result_and_invoke([[maybe_unused]] indexed_sequence_pair_t && indexed_sequence_pair,
```
marehr, line 88:
```suggestion
    void make_result_and_invoke([[maybe_unused]] indexed_sequence_pair_t && indexed_sequence_pair,
```

I have a strong preference for make.

rrahn: should the policy then not be called `maker` 🤣

marehr, line 88: Boah, I'm not sure about the semantic of this function. This function is not really building or making any results, it kind of pre-initialises the result and passes that to the callback. Where is the other data like front, back coordinate and trace back populated again? Is this additional step really needed why can't the called callback do this step?

rrahn: This is the new alingment implementation. Currently, it only computes the global alignment. the other features are currently implemented in parallel. This is absolutely the right place to put something like the trace builder and so on in. We discussed this design in the core meeting and decided for this design.

marehr: Ah okay! So to double check, in the future this function will populate the front, back coordinate and the other stuff?

include/seqan3/alignment/pairwise/alignment_result.hpp
marehr, line 127:
```suggestion
 * The seqan3::detail::policy_alignment_result_builder is used to set the respective
```

include/seqan3/alignment/pairwise/detail/policy_alignment_result_builder.hpp
marehr: No, it is the builder pattern, but we create = make results. BTW in `policy_search_result_builder` we called it `make_results`.

rrahn: absolutely.

rrahn: I mean, I introduced it because I am working on the traceback part right now.

include/seqan3/alignment/pairwise/detail/policy_alignment_result_builder.hpp
```diff
@@ -0,0 +1,111 @@
...
+                      "Invalid configuration. Expected result with id!");
+        result.data.id = std::move(get<1>(indexed_sequence_pair));
+
+        if constexpr (traits_type::compute_score)
+        {
+            static_assert(!std::same_as<decltype(result.data.score), invalid_t>,
+                          "Invalid configuration. Expected result with score!");
+            result.data.score = std::move(score);
+        }
+
```
marehr, line 107:
```suggestion

       // TODO: Add other result.data like sequence1_begin_position, sequence1_end_position, and so on.

```

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
marehr: How do we solve this? The sequence builder will need this data inside the `make_result_...`, or?

include/seqan3/alignment/pairwise/detail/policy_alignment_result_builder.hpp
rrahn: I don't think this is necessary in this PR. I mean I have the change in front of my eyes. This is nothing I forget until monday.

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
rrahn: That is a really good point. The only solution would be to pass the sequence pair and the id separatey into the call. I wanted to keep the parameter list small but maybe this is better in this case.

include/seqan3/alignment/pairwise/detail/policy_alignment_result_builder.hpp
marehr: This is better understandable for the reviewer. :shrug:

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
marehr: Still open

rrahn:
> Move thing is till open

I reverted the change of the for loop and are now passing the sequence and the id separately.

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
```diff
@@ -141,11 +141,10 @@ class pairwise_alignment_algorithm : protected policies_t...
     {
         using std::get;
 
-        for (auto && indexed_sequence_pair : indexed_sequence_pairs)
+        for (auto && [sequence_pair, idx] indexed_sequence_pair : indexed_sequence_pairs)
         {
-            auto && sequence_pair = get<0>(indexed_sequence_pair);
             compute_matrix(get<0>(sequence_pair), get<1>(sequence_pair));
-            this->make_result_and_invoke(indexed_sequence_pair, this->tracked_optimum(), callback);
+            this->make_result_and_invoke(sequence_pair, std::move(idx), this->tracked_optimum(), callback);
```
marehr, line 147:
```suggestion
            this->make_result_and_invoke(std::forward<decltype(sequence_pair)>(sequence_pair), std::move(idx), this->tracked_optimum(), callback);
```

Does this make sense? I'm not sure

rrahn: Huh, I need to think about it.

smehringer: Is this still open or is there a resolution?

Issue #124, opened 2018-02-20
Where to place `test/` utility functions?


marehr:
#120 introduced `/test/include/seqan3_test/tmp_filename.hpp`

I would like to discuss where files should life and what the namespace is.

We could do it in the following ways:
* put it into  `/include/seqan3/test/`
* put it into  `/include/seqan3_test/`
* put it into  `/test/include/seqan3/`
* put it into  `/test/include/seqan3_test/`

I think the namespace `seqan3::test` is already good.

h-2:
I would put them in `/test/include/seqan3/`

rrahn:
I am also now for changing it to `/test/include/seqan3/`
At the beginning, I wasn't sure, because of having the same include name for library and test code, but then again we have the special `test` namespace, which makes the include name obsolete, I think.
I will change this now for the current PR #120.

rrahn:
Ok, after a fruitful 🍌 discussion with @marehr, it seems still better to leave the original idea of using `/test/include/seqan3/test` 
```cpp
#include <seqan3/test/xyz.hpp>
#include <seqan3/range/view.hpp>
```
or `/test/include/seqan3_test` 
```cpp
#include <seqan3_test/xyz.hpp>
#include <seqan3/range/view.hpp>
```
Both versions are much better for a destination of test utility functionality and stuff from the main repository. 
The question remains of how to choose the namespace.
Either:
```cpp
namespace seqan3::test
{
 ...
}

// Usage:
using namespace seqan3;
do_something ...
test::tmp_file_name ...
```
or 
```cpp
namespace seqan3_test
{
 ...
}

// Usage:
using namespace seqan3;
do_something ...
::seqan3_test::tmp_file_name ...
```
So after considering this I suggest we use the first approach.

h-2:
> Ok, after a fruitful banana discussion with @marehr, it seems still better to leave the original idea of using /test/include/seqan3/test
> or /test/include/seqan3_test

OK, I have not strong feelings about it, but would prefer `/test/include/seqan3/test`

> The question remains of how to choose the namespace.

A sub-namespace is definitely preferable, because you can call stuff from the immediate parent namespace (in our case `seqan3`) without prefixing it, whereas with a separate namespace you would need lot's of `seqan3::` ...

Also this should be in line with the above decision. (if it's `seqan3/test` it should be `seqan3::test`, if it's `seqan3_test` it should also be `seqan3_test`),

marehr:
[...]

PR #1909, opened 2020-06-14
 [FIX] use std::cpp20::ostream[buf]_iterator instead of seqan3::ostream[buf]_iterator

marehr:
Part of #124

### DIFF
--- a/include/seqan3/io/alignment_file/format_bam.hpp
+++ b/include/seqan3/io/alignment_file/format_bam.hpp
@@ -694,7 +694,7 @@ inline void format_bam::write_alignment_record([[maybe_unused]] stream_type &  s
         if (ref_offset.has_value() && (ref_offset.value() + 1) < 0)
             throw format_error{detail::to_string("The ref_offset object must be >= -1 but is: ", ref_offset)};
 
-        seqan3::ostreambuf_iterator stream_it{stream};
+        std::cpp20::ostreambuf_iterator stream_it{stream};
 
         // ---------------------------------------------------------------------
         // Writing the BAM Header on first call
--- a/include/seqan3/io/stream/iterator.hpp
+++ b/include/seqan3/io/stream/iterator.hpp
@@ -14,19 +14,15 @@
 
 #pragma once
 
-#include <iterator>
-
-#ifndef __cpp_lib_ranges
-#include <range/v3/iterator/stream_iterators.hpp>
-#endif // __cpp_lib_ranges
+#include <seqan3/std/algorithm>
+#include <seqan3/std/charconv>
+#include <seqan3/std/iterator>
+#include <seqan3/std/ranges>
 
 #include <seqan3/core/platform.hpp>
 #include <seqan3/core/concept/core_language.hpp>
-#include <seqan3/range/views/take.hpp>
 #include <seqan3/range/views/drop.hpp>
-#include <seqan3/std/algorithm>
-#include <seqan3/std/charconv>
-#include <seqan3/std/ranges>
+#include <seqan3/range/views/take.hpp>
 
 namespace seqan3::detail
 {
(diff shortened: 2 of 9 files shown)


include/seqan3/io/stream/iterator.hpp
```diff
@@ -14,19 +14,15 @@
...
-#endif // __cpp_lib_ranges
+#include <seqan3/std/algorithm>
+#include <seqan3/std/charconv>
+#include <seqan3/std/iterator>
+#include <seqan3/std/ranges>
 
 #include <seqan3/core/platform.hpp>
 #include <seqan3/core/concept/core_language.hpp>
 #include <seqan3/range/views/take.hpp>
 #include <seqan3/range/views/drop.hpp>
```
simonsasse, line 25:
```suggestion
#include <seqan3/range/views/drop.hpp>
#include <seqan3/range/views/take.hpp>
```

include/seqan3/std/iterator
```diff
@@ -19,13 +19,17 @@
 #ifdef __cpp_lib_ranges
 namespace std
 {
-inline namespace cpp20
+namespace cpp20
```
rrahn, line 22: removing inline was necessary, because?

marehr: I normally write my reasons in the commit itself:

https://github.com/seqan/seqan3/pull/1909/commits/37dfe17eeda35476d2afb16c7824cb8e098c061f

> It turned out that inline namespace for `std::cpp20` was a bad idea
since it produced ambiguousness for `std::ostream_iterator` and
`std::cpp20::ostream_iterator`.

include/seqan3/io/stream/iterator.hpp
```diff
@@ -381,42 +377,3 @@ class fast_ostreambuf_iterator
...
- *        constructed.
- */
-using SEQAN3_DOXYGEN_ONLY(ostream_iterator =) ::ranges::ostream_iterator;
-
-/*!\typedef seqan3::ostreambuf_iterator
- * \brief Alias for ranges::ostreambuf_iterator. Writes successive characters onto the output stream from which it
- *        was constructed.
- */
-using SEQAN3_DOXYGEN_ONLY(ostreambuf_iterator =) ::ranges::ostreambuf_iterator;
-
```
marehr, line 419: > Just one question to be sure. The renamed iterator types have nothing to do with our faster stream iterator implementations?

As you can see, this has nothing to do with our faster stream iterators. This is from way back when we didn't have the `seqan3/std` folder for "standard" implementations.

That means `seqan3::ostreambuf_iterator` was either `ranges::ostreambuf_iterator` (pre C++20) or `std::ostreambuf_iterator` (C++20).

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -694,7 +694,7 @@ inline void format_bam::write_alignment_record([[maybe_unused]] stream_type &  s
         if (ref_offset.has_value() && (ref_offset.value() + 1) < 0)
             throw format_error{detail::to_string("The ref_offset object must be >= -1 but is: ", ref_offset)};
 
-        seqan3::ostreambuf_iterator stream_it{stream};
+        std::cpp20::ostreambuf_iterator stream_it{stream};
```
smehringer, line 697: Why did you change this? We explicitly introduced our own ostreambuf_iterator because we had a significant performance improvement when exposing the buffer pointers and writing directly to the buffer.

smehringer: Edit: Sorry, I just saw that you are right `seqan3::ostreambuf_iterator` indeed is either ranges::ostreambuf_iterator or std::ostreambuf_iterator. It is the `seqan::fast_ostreambuf_iterator` I was talking about. I think it should be changed to  `seqan::fast_ostreambuf_iterator` then.

marehr: good that I renamed it to avoid confusion :)

marehr: This is a performance improvement and I would not want to do it in this PR. I think we should do measurements if we swap the implementations. I think this can be done by a student :)

smehringer: I think we should change it right away, why wait for an improvement. But rene is the reviwer so I'll leave this PR to him

include/seqan3/std/iterator
rrahn: > I normally write my reasons in the commit itself:

I am usually not going through the commits.

include/seqan3/io/alignment_file/format_bam.hpp
rrahn: Yes and I/O is not part of this or the next quarter. So we can look at the performance when we work on that.

rrahn: I would really have the applications done with the other stuff that is on our plate.

include/seqan3/io/stream/iterator.hpp
rrahn: Ok, thanks for clarifying.

include/seqan3/io/alignment_file/format_bam.hpp
marehr: I'm not against it, because it is low-hanging fruit, but I think it is outside this PR.

PR #1910, opened 2020-06-14
[INFRA] modularise cmake scripts

marehr:
We want to add more features via cmake to our build infrastructure and want to re-use it in our application template that it makes sense to split out functionality of our cmake-infrastructure into an own dedicated folder.

@mr-c This might have implications for your debian packages.

### DIFF
--- a/test/cmake/add_subdirectories.cmake
+++ b/test/cmake/add_subdirectories.cmake
@@ -1,0 +1,39 @@
+# -----------------------------------------------------------------------------------------------------
+# Copyright (c) 2006-2020, Knut Reinert & Freie Universität Berlin
+# Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+# This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+# shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+# -----------------------------------------------------------------------------------------------------
+
+# Calls add_subdirectory on all (direct) subdirectories of the given directory if they contain a `CMakeLists.txt`
+#
+# Example:
+# If we have
+# * /some/path/directory/subdir1/CMakeLists.txt
+# * /some/path/directory/subdir2/CMakeLists.txt
+# * /some/path/directory/subdir3/CMakeLists.txt
+# * /some/path/directory/subdir4/has-no-CMakeLists.txt
+#
+# This macro calls
+# * `add_subdirectory ("/some/path/directory/subdir1")`,
+# * `add_subdirectory ("/some/path/directory/subdir2")`, and
+# * `add_subdirectory ("/some/path/directory/subdir3")`,
+# but not `add_subdirectory ("/some/path/directory/subdir4")`, because it does not contain a CMakeLists.txt
+macro (add_subdirectories_of directory)
+    file (GLOB ENTRIES
+          RELATIVE ${directory}
+          ${directory}/[!.]*)
+
+    foreach (ENTRY ${ENTRIES})
+        if (IS_DIRECTORY ${directory}/${ENTRY})
+            if (EXISTS ${directory}/${ENTRY}/CMakeLists.txt)
+                add_subdirectory (${directory}/${ENTRY} ${CMAKE_CURRENT_BINARY_DIR}/${ENTRY})
+            endif ()
+        endif ()
+    endforeach ()
+    unset (ENTRIES)
+endmacro ()
+
+macro (add_subdirectories)
+    add_subdirectories_of(${CMAKE_CURRENT_SOURCE_DIR})
+endmacro ()
--- a/test/cmake/seqan3_require_benchmark.cmake
+++ b/test/cmake/seqan3_require_benchmark.cmake
@@ -1,0 +1,50 @@
+# -----------------------------------------------------------------------------------------------------
+# Copyright (c) 2006-2020, Knut Reinert & Freie Universität Berlin
+# Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+# This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+# shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+# -----------------------------------------------------------------------------------------------------
+
+# Exposes the google-benchmark target `gbenchmark`.
+macro (seqan3_require_benchmark)
+    enable_testing ()
+
+    set (gbenchmark_project_args ${SEQAN3_EXTERNAL_PROJECT_CMAKE_ARGS})
+    list (APPEND gbenchmark_project_args "-DBENCHMARK_ENABLE_TESTING=false")
+    # google-benchmarks suggest to use LTO (link-time optimisation), but we don't really need that because we are a
+    # header only library. This option might be still interesting for external libraries benchmarks.
+    # see https://github.com/google/benchmark#debug-vs-release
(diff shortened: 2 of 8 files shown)


test/cmake/seqan3_require_benchmark.cmake
```diff
@@ -0,0 +1,47 @@
...
+# shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+# -----------------------------------------------------------------------------------------------------
+
+# Exposes the google-benchmark target `gbenchmark`.
+macro (seqan3_require_benchmark)
+    enable_testing ()
+
+    set (gbenchmark_project_args ${SEQAN3_EXTERNAL_PROJECT_CMAKE_ARGS})
+    list (APPEND gbenchmark_project_args "-DBENCHMARK_ENABLE_TESTING=false")
+    # list (APPEND gbenchmark_project_args "-DBENCHMARK_ENABLE_LTO=true")
```
MitraDarja, line 14: What is the purpose of this comment?

test/cmake/seqan3_test_component.cmake
```diff
@@ -0,0 +1,39 @@
...
+# Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+# This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+# shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+# -----------------------------------------------------------------------------------------------------
+
+# Get a specific component of a test file which follows the seqan3 naming scheme.
+# e.g. target_source_file = "range/views/take.cpp"
+# component:
+#  * TARGET_NAME - the target name (e.g. "take")
+#  * TARGET_UNIQUE_NAME - the target name which includes the target_path (e.g. range-view-take)
```
MitraDarja, line 12: What is the purpose of the target_unique_name?

test/cmake/seqan3_require_benchmark.cmake
marehr: It was there before.

The documentation for google benchmark suggest to use LTO, see https://github.com/google/benchmark#debug-vs-release.

We don't really need it, because we don't link any code, but it might be still interesting to use it. I think I had some problems using google benchmark with LTO, so I commented it out again, but kept it, because it might be handy.

But since I never touched this line ever again we might just remove it. :shrug:

test/cmake/seqan3_test_component.cmake
marehr: Like it said: the target name which includes the target_path (e.g. range-view-take)

```cmake
macro (seqan3_test_component TEST_TARGET_NAME "range/views/take.cpp" TARGET_NAME)
macro (seqan3_test_component TEST_TARGET_UNIQUE_NAME "range/views/take.cpp" TARGET_UNIQUE_NAME)
mesage(STATUS "TEST_TARGET_NAME: ${TEST_TARGET_NAME}") // take
mesage(STATUS "TARGET_UNIQUE_NAME: ${TARGET_UNIQUE_NAME}") // range-views-take
```

test/cmake/seqan3_require_benchmark.cmake
MitraDarja: I would say either remove it or add a comment with the link, so it is clearer why it is there?

marehr, line 14:
```suggestion
    # google-benchmarks suggest to use LTO (link-time optimisation), but we don't really need that because we are a
    # header only library. This option might be still interesting for external libraries benchmarks.
    # see https://github.com/google/benchmark#debug-vs-release
    # list (APPEND gbenchmark_project_args "-DBENCHMARK_ENABLE_LTO=true")
```

marehr: Explained it :)

test/cmake/add_subdirectories.cmake
```diff
@@ -0,0 +1,25 @@
+# -----------------------------------------------------------------------------------------------------
+# Copyright (c) 2006-2020, Knut Reinert & Freie Universität Berlin
+# Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+# This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+# shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+# -----------------------------------------------------------------------------------------------------
+
+macro (add_subdirectories_of directory)
```
rrahn, line 8: Can you document input and what happens inside of the function?
Whether or not variables are exposed to the parent directory or are only valid in the subdirectories and these kind of cmake stuff? I know it is cumbersome, but might help in the future.

test/cmake/seqan3_require_benchmark.cmake
```diff
@@ -0,0 +1,50 @@
+# -----------------------------------------------------------------------------------------------------
+# Copyright (c) 2006-2020, Knut Reinert & Freie Universität Berlin
+# Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+# This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+# shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+# -----------------------------------------------------------------------------------------------------
+
+# Exposes the google-benchmark target `gbenchmark`.
+macro (seqan3_require_benchmark)
+    enable_testing ()
```
rrahn, line 10: I just saw it in other places as well. Somehow this looks strange if the parenthesis is separated with a blank if you compare it to c++ code. Is this really the recommended style for cmake?

test/cmake/add_subdirectories.cmake
marehr: :unamused: A pull request that just moved existing code into own files, all right lets make this proper documented :shrug:

test/cmake/seqan3_require_benchmark.cmake
marehr: Ask Hannes, he was in favour for this. cmake advertises this spacing in https://cmake.org/examples/, but uses a different one in the documentation: https://cmake.org/cmake/help/v3.18/manual/cmake-compile-features.7.html#compile-feature-requirements.

I was more in favour for `macro(seqan3_require_benchmark)`, `enable_testing()`, `if(hello)`, `endif()` and so on

test/cmake/add_subdirectories.cmake
marehr, line 8:
```suggestion

# Calls add_subdirectory on all (direct) subdirectories of the given directory if they contain a `CMakeLists.txt`
#
# Example:
# If we have
# * /some/path/directory/subdir1/CMakeLists.txt
# * /some/path/directory/subdir2/CMakeLists.txt
# * /some/path/directory/subdir3/CMakeLists.txt
# * /some/path/directory/subdir4/has-no-CMakeLists.txt
#
# This macro calls
# * `add_subdirectory ("/some/path/directory/subdir1")`,
# * `add_subdirectory ("/some/path/directory/subdir2")`, and
# * `add_subdirectory ("/some/path/directory/subdir3")`,
# but not `add_subdirectory ("/some/path/directory/subdir4")`, because it does not contain a CMakeLists.txt
macro (add_subdirectories_of directory)
```

test/cmake/seqan3_require_benchmark.cmake
rrahn: The point is, it is not consistent and given the style in C++ it is always awkward to see it differently here. But not so important.

rrahn:
you can rebase and then we can merge it.

PR #1911, opened 2020-06-14
[FEATURE] Technical Binning Directory

eseiler:
Resolves https://github.com/seqan/product_backlog/issues/36

Open:
  * Introduce `view_adaptor_for<adaptor_t, range_t>`? (` requires(range_t range, adaptor_t adaptor) { {range | adaptor} -> std::ranges::view }` )
  * simplify template arguments for compressed version (own class?, solve via config?)

Follow-up (?):
  * Multithreading for the ctor. Each thread can handle wordsize many bins (otherwise there may be concurrent write access to a memory position).

### DIFF
--- a/include/seqan3/search/dream_index/technical_binning_directory.hpp
+++ b/include/seqan3/search/dream_index/technical_binning_directory.hpp
@@ -37,6 +37,36 @@ struct ibf_config
     uint8_t threads{1u}; //!< The number of threads to use functions.
 };
 
+enum class hash_variant : uint8_t
+{
+    kmer,
+    minimiser
+};
+
+template <auto t, typename hasher_t>
+struct hash_proxy;
+
+template <typename hasher_t>
+struct hash_proxy<hash_variant::kmer, hasher_t>
+{
+    size_t kmer_size;
+    hasher_t hasher(/*seqan3::views::kmer_hash(seqan3::ungapped{5u})*/); // view is not default constructible
+
+    hash_proxy(size_t kmer_size_) : kmer_size(kmer_size_)/*,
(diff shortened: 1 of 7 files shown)


include/seqan3/search/dream_index/technical_binning_directory.hpp
```diff
@@ -0,0 +1,66 @@
...
+    technical_binning_directory() = default; //!< Defaulted.
+    technical_binning_directory(technical_binning_directory const &) = default; //!< Defaulted.
+    technical_binning_directory & operator=(technical_binning_directory const &) = default; //!< Defaulted.
+    technical_binning_directory(technical_binning_directory &&) = default; //!< Defaulted.
+    technical_binning_directory & operator=(technical_binning_directory &&) = default; //!< Defaulted.
+    ~technical_binning_directory() = default; //!< Defaulted.
+
+    template <typename rng_t>
+    technical_binning_directory(std::vector<rng_t> const & technical_bins,
+                                hash_view_t && hash_view,
```
marehr, line 44:
```suggestion
                                hash_view_t hash_view,
```

Because `hash_view_t` is concrete type here.

eseiler:
### Core meeting 20-2-2021
* The IBF will get a function `count_values(query_kmers)` and the `counting_agent`
* The technical_binning_directory (TBD) will have the function `count_query(query)` that delegates to `count_kmers` in the IBF. 
* The TBD will not be part of the library just yet. It may remain just a cook book entry.

eseiler:
Since we will solve this differently, I'll close this PR for now.

PR #1915, opened 2020-06-16
[FIX] typo in the return type constraint macro

joergi-w:
This PR fixes the macro that has been introduced in PR #1866. 

I choose the same reviewers as in #1866.


smehringer:
Travis is done without errors but not updating so I am merging this

PR #1922, opened 2020-06-21
[FEATURE] Counting vector

eseiler:
Resolves https://github.com/seqan/product_backlog/issues/77

### DIFF
--- a/include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
+++ b/include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
@@ -603,6 +603,95 @@ private:
     friend class membership_agent;
     using sdsl::bit_vector::resize;
     using sdsl::bit_vector::set_int;
+    template <std::integral value_t>
+    friend class counting_vector;
+};
+
+/*!\brief A data structure that behaves like a std::vector and can be used to consolidate the results of multiple calls
+ *        to seqan3::interleaved_bloom_filter::membership_agent::bulk_contains.
+ * \tparam value_t The type of the count. Must model std::integral.
+ *
+ * \details
+ *
+ * When using the seqan3::interleaved_bloom_filter::membership_agent::bulk_contains operation, a common use case is to
+ * add up, for example, the results for all k-mers in a query. This yields, for each bin, the number of k-mers of a
+ * query that are in the respective bin. Such information can be used to apply further filtering or abundance estimation
+ * based on the k-mer counts.
+ *
+ * The seqan3::counting_vector offers an easy way to add up the individual
+ * seqan3::interleaved_bloom_filter::membership_agent::binning_bitvector by offering an `+=` operator.
+ *
+ * The `value_t` template parameter should be chosen in a way that no overflow occurs if all calls to `bulk_contains`
+ * return a hit for a specific bin. For example, `uint8_t` will suffice when processing short Illumina reads, whereas
+ * long reads will require at least `uint32_t`.
+ *
+ * ### Example
+ *
+ * \include test/snippet/search/dream_index/counting_vector.cpp
+ */
+template <std::integral value_t>
+class counting_vector : public std::vector<value_t>
+{
+private:
+    //!\brief The base type.
+    using base_t = std::vector<value_t>;
+public:
+    /*!\name Constructors, destructor and assignment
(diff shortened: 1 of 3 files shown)


include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -599,8 +599,62 @@ class interleaved_bloom_filter<data_layout_mode>::bulk_agent::binning_bitvector
...
+    template <typename t>
+    //!\cond
+        requires std::is_base_of<sdsl::bit_vector, t>::value
+    //!\endcond
+    counting_vector & operator+=(t const & rhs)
+    {
+        assert(size() >= rhs.size()); // The counting vector may be bigger than what we need.
+        size_t bin{0};
+        // Each iteration can handle 64 bits, so we need to iteratate `((rhs.size() + 63) >> 6` masny times
+        for (size_t batch = 0; batch < ((rhs.size() + 63) >> 6); ++batch)
```
eseiler, line 636:
```suggestion
        // Each iteration can handle 64 bits, so we need to iteratate `((rhs.size() + 63) >> 6` masny times
        for (size_t batch = 0, bin = 0; batch < ((rhs.size() + 63 + 63) >> 6); bin = 64 * batch++)
```

Also add testcase:
0^64 1^64

test/snippet/search/dream_index/counting_vector.cpp
```diff
@@ -0,0 +1,23 @@
...
+    ibf.emplace(126, seqan3::bin_index{9u});
+    ibf.emplace(712, seqan3::bin_index{3u});
+    ibf.emplace(237, seqan3::bin_index{9u});
+
+    // The counting_vector must be at least as big as there are bins.
+    seqan3::counting_vector<uint8_t> counts(12, 0);
+
+    auto agent = ibf.membership_agent();
+    counts += agent.bulk_contains(712);
+    seqan3::debug_stream << counts << '\n'; // prints [0,0,0,1,0,0,0,0,0,0,0,0]
```
Irallia, line 18: Maybe add some documentation:
```suggestion
    counts += agent.bulk_contains(712);
    // A `1` at position `i` indicates the (probable) presence of the query in bin `i`.
    // After adding bins a `2` (or higher) indicates the (probable) presence of 2 (or more) queries in bin `i`.
    seqan3::debug_stream << counts << '\n'; // prints [0,0,0,1,0,0,0,0,0,0,0,0]
```

eseiler:
> LGFM, just a bit more comments in the snippet, would be helpful.

I added some

> And does it make sense to insert the snippet in the same document twice?

I think it makes sense to have the snippet as example for the class and as documentation for the actual interesting part of the implementation ... it just so happens that there's nothing interesting to show in the class-documentation than what the `operator+=` does.

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -603,6 +603,95 @@ class interleaved_bloom_filter<data_layout_mode>::membership_agent::binning_bitv
...
+    counting_vector(counting_vector &&) = default; //!< Defaulted.
+    counting_vector & operator=(counting_vector &&) = default; //!< Defaulted.
+    ~counting_vector() = default; //!< Defaulted.
+
+    using base_t::base_t;
+    //!\}
+
+    /*!\brief Bin-wise adds the bits of a seqan3::interleaved_bloom_filter::membership_agent::binning_bitvector.
+     * \tparam rhs_t The type of the right-hand side.
+     *         Must be seqan3::interleaved_bloom_filter::membership_agent::binning_bitvector.
```
smehringer, line 654: This does not correspond to your requires? Change either fo them

eseiler: I cannot require the binning_bitvector because I cannot deduce the a template argument (data_layout)  of the right-hand side of a template member function.
But I can require that the member function is of a certain base class.

What I actually want is what I wrote, what I can do is what I wrote and we do not want to use sdsl in public.

eseiler:
> General question: Didn't we want to make an `counting_agent` instead of a individual counting vector?

This is part of the technical binning directory.

PR #1923, opened 2020-06-22
[FIX] seqan3::complement cannot require std::same_as on the input and return type.

### LINKED ISSUE #1743: Complement view is not working with bitcompressed dna vector
The seqan3::views::complement given a static_assert error, when using a seqan3::bitcompressed_vector. (Error mesage. "The innermost value type must satisfy the nucleotide_alphabet.") I would expect complement to work on bitcompressed vectors as well.

```cpp
#include <seqan3/range/container/bitcompressed_vector.hpp>
#include <seqan3/range/views/complement.hpp>

int main()
{
    seqan3::bitcompressed_vector<seqan3::dna4> text1{"ACGTAGC"_dna4};
    auto hashes2 = text1 | seqan3::views::complement;
    seqan3::debug_stream << hashes2 << '\n';
}
```

smehringer:
Fixes #1743 

Before this was not possible (now it is):
```cpp
    seqan3::bitcompressed_vector<seqan3::dna4> v{seqan3::dna4{}, seqan3::dna4{}};
    auto proxy = *v.begin();
    auto complement = seqan3::complement(proxy);

    // or working with views:
    auto complement = v | seqan3::views::complement; 
```
**Reason**: The `seqan3::complement` cpo had the requirement that the return type is the same as the type of the parameter. But the complement of the proxy returned a `seqan3::dna4` again.
I removed this requirement from the concept.

I also thought about making the complement of a proxy return a proxy again but then where would this proxy point to?
So I think changing the requirement of the concept was the right way.

mentioning @h-2, just in case you want to take a look.

### DIFF
--- a/include/seqan3/alphabet/nucleotide/concept.hpp
+++ b/include/seqan3/alphabet/nucleotide/concept.hpp
@@ -43,10 +43,12 @@ public:
         {
             { impl(priority_tag<2>{}, nucl) };
             requires noexcept(impl(priority_tag<2>{}, nucl));
-            requires std::same_as<nucleotide_t, decltype(impl(priority_tag<2>{}, nucl))>;
+            // requires std::common_with<decltype(impl(priority_tag<2>{}, nucl)), nucleotide_t>; // triggers an ICE
+            requires alphabet<decltype(impl(priority_tag<2>{}, nucl))>;
+            { impl(priority_tag<2>{}, impl(priority_tag<2>{}, nucl)) }; // you can take the complement again
         }
     //!\endcond
-    constexpr nucleotide_t operator()(nucleotide_t const nucl) const noexcept
+    constexpr auto operator()(nucleotide_t const nucl) const noexcept
     {
         return impl(priority_tag<2>{}, nucl);
     }
--- a/test/unit/range/container/bitcompressed_vector_test.cpp
+++ b/test/unit/range/container/bitcompressed_vector_test.cpp
@@ -1,0 +1,12 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2020, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+#include <gtest/gtest.h>
+
+#include "container_test_template.hpp"
+
+INSTANTIATE_TYPED_TEST_SUITE_P(bitcompressed, container_over_dna4_test, seqan3::bitcompressed_vector<seqan3::dna4>, );
--- a/test/unit/range/container/bitcompressed_vector_test.cpp
+++ b/test/unit/range/container/bitcompressed_vector_test.cpp
@@ -7,6 +7,33 @@
 
 #include <gtest/gtest.h>
 
+#include <seqan3/alphabet/nucleotide/concept.hpp>
+#include <seqan3/alphabet/nucleotide/dna4.hpp>
+#include <seqan3/range/views/complement.hpp>
+#include <seqan3/test/expect_range_eq.hpp>
+
 #include "container_test_template.hpp"
 
 INSTANTIATE_TYPED_TEST_SUITE_P(bitcompressed, container_over_dna4_test, seqan3::bitcompressed_vector<seqan3::dna4>, );
+
+using seqan3::operator""_dna4;
+
+TEST(bitcompressed_vector_test, issue1743_complement_on_proxy)
+{ // https://github.com/seqan/seqan3/issues/1743
+    seqan3::bitcompressed_vector<seqan3::dna4> v{'A'_dna4};
+
+    auto proxy = *v.begin();
+    auto complement = seqan3::complement(proxy);
+
+    EXPECT_TRUE((std::same_as<decltype(complement), seqan3::dna4>));
+    EXPECT_EQ(complement, 'T'_dna4);
+}
+
+TEST(bitcompressed_vector_test, issue1743_view_combinability)
+{ // https://github.com/seqan/seqan3/issues/1743
+    seqan3::bitcompressed_vector<seqan3::dna4> v{'A'_dna4, 'C'_dna4, 'G'_dna4, 'T'_dna4};
+    auto complement = v | seqan3::views::complement;
+
+    EXPECT_EQ(v.size(), complement.size());
+    EXPECT_RANGE_EQ(complement, (seqan3::dna4_vector{'T'_dna4, 'G'_dna4, 'C'_dna4, 'A'_dna4}));
+}
--- a/test/unit/range/container/container_test_template.hpp
+++ b/test/unit/range/container/container_test_template.hpp
@@ -20,21 +20,17 @@
 using seqan3::operator""_dna4;
 
 template <typename T>
-class container_ : public ::testing::Test
+class container_over_dna4_test : public ::testing::Test
 {};
 
-using container_types = ::testing::Types<std::vector<seqan3::dna4>,
-                                         seqan3::bitcompressed_vector<seqan3::dna4>,
-                                         seqan3::small_vector<seqan3::dna4, 1000>>;
+TYPED_TEST_SUITE_P(container_over_dna4_test);
 
-TYPED_TEST_SUITE(container_, container_types, );
-
-TYPED_TEST(container_, concepts)
+TYPED_TEST_P(container_over_dna4_test, concepts)
 {
     EXPECT_TRUE(seqan3::reservible_container<TypeParam>);
 }
 
-TYPED_TEST(container_, construction)
+TYPED_TEST_P(container_over_dna4_test, construction)
 {
     TypeParam t1;
     TypeParam t2{};
@@ -57,7 +53,7 @@ TYPED_TEST(container_, construction)
     EXPECT_EQ(t3, t7);
 }
 
-TYPED_TEST(container_, swap)
+TYPED_TEST_P(container_over_dna4_test, swapping)
 {
     TypeParam t0{};
     TypeParam t1{'A'_dna4, 'C'_dna4, 'C'_dna4, 'G'_dna4, 'T'_dna4};
@@ -71,7 +67,7 @@ TYPED_TEST(container_, swap)
     EXPECT_EQ(t1, (TypeParam{'A'_dna4, 'C'_dna4, 'C'_dna4, 'G'_dna4, 'T'_dna4}));
 }
 
-TYPED_TEST(container_, assign)
+TYPED_TEST_P(container_over_dna4_test, assign)
 {
     TypeParam t0{'C'_dna4, 'C'_dna4};
     TypeParam t1{'A'_dna4, 'C'_dna4, 'C'_dna4, 'G'_dna4, 'T'_dna4};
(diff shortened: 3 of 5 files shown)


h-2:
> The seqan3::complement cpo had the requirement that the return type is the same as the type of the parameter. But the complement of the proxy returned a seqan3::dna4 again.
> I removed this requirement from the concept.

It would probably be best to require that the type itself and the type returned by `complement()` satisfy `std::common_with`. And also that the type is a nucleotide itself (this can likely not be checked because of cyclic dependencies but should be specified in documention). Both requirements ensure that you can complement a complement and stay in the nucleotide domain and get back to something that is common with the original.

> I also thought about making the complement of a proxy return a proxy again but then where would this proxy point to?

No, this is not required. `complement()` always creates a new value anyway, so no reference/proxy semantics are required.

smehringer:
> It would probably be best to require that the type itself and the type returned by complement() satisfy std::common_with. And also that the type is a nucleotide itself (this can likely not be checked because of cyclic dependencies but should be specified in documention). Both requirements ensure that you can complement a complement and stay in the nucleotide domain and get back to something that is common with the original.

I already had the idea of `std::convertible_to` which resulted in an ICE. And `std::common_with` also triggers an ICE. :(

smehringer:
instead of requiring `nucleotide_alphabet` I can require `impl(priority_tag<2>{}, impl(priority_tag<2>{}, nucl))` which is basically the same.

test/unit/range/container/bitcompressed_vector_test.cpp
```diff
@@ -0,0 +1,27 @@
...
+// Copyright (c) 2006-2020, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+#include <gtest/gtest.h>
+
+#include <seqan3/alphabet/nucleotide/dna4.hpp>
+#include <seqan3/alphabet/nucleotide/concept.hpp>
```
joergi-w, line 11: sort alphabetically

test/unit/range/container/bitcompressed_vector_test.cpp
```diff
@@ -0,0 +1,27 @@
...
+{ // https://github.com/seqan/seqan3/issues/1743
+
+    seqan3::bitcompressed_vector<seqan3::dna4> v{seqan3::dna4{}, seqan3::dna4{}};
+
+    auto proxy = *v.begin();
+    auto complement = seqan3::complement(proxy);
+
+    EXPECT_TRUE((std::same_as<decltype(complement), seqan3::dna4>));
+    EXPECT_EQ(complement, seqan3::assign_rank_to(3, seqan3::dna4{}));
+}
```
joergi-w, line 27: Please add a test for `view_complement` as well.

test/unit/range/container/container_test_template.hpp
```diff
@@ -20,21 +20,17 @@
 using seqan3::operator""_dna4;
 
 template <typename T>
-class container_ : public ::testing::Test
+class container_over_dna4_test : public ::testing::Test
 {};
 
-using container_types = ::testing::Types<std::vector<seqan3::dna4>,
-                                         seqan3::bitcompressed_vector<seqan3::dna4>,
-                                         seqan3::small_vector<seqan3::dna4, 1000>>;
```
joergi-w, line 28: Here you remove the tests for 3 different types, but you only add the test for bitcompressed again. Shouldn't we keep tests also for small_vector and vector?

smehringer: I also added the test template in small_vector_test.cpp !
And I thought we do no need to test `std::vector` :woman_shrugging:

joergi-w: True... sorry I couldn't find it, but I see it's there.

smehringer:
@rrahn polite ping :)

test/unit/range/container/container_test_template.hpp
```diff
@@ -20,21 +20,17 @@
 using seqan3::operator""_dna4;
 
 template <typename T>
-class container_ : public ::testing::Test
+class container_over_dna4_test : public ::testing::Test
```
rrahn, line 23: It is extremely weird, why a template is constrained on the alphabet. The alphabet should be either given explicitly or, since we require it to be a container, extracted from the given test type.

include/seqan3/alphabet/nucleotide/concept.hpp
```diff
@@ -43,10 +43,10 @@ struct complement_fn
         {
             { impl(priority_tag<2>{}, nucl) };
             requires noexcept(impl(priority_tag<2>{}, nucl));
-            requires std::same_as<nucleotide_t, decltype(impl(priority_tag<2>{}, nucl))>;
+            { impl(priority_tag<2>{}, impl(priority_tag<2>{}, nucl)) }; // you can take the complement again
```
rrahn, line 46: We need a back log item to analyse why `requires std::common_with<decltype(impl(priority_tag<2>{}, nucl)), nucleotide_t>;` is not working correctly. This is rather due to some error on our side I believe.

rrahn: Also please add the following: `requires alphabet<decltype(impl(priority_tag<2>{}, nucl))>;` to at least gurantee, that I don't return some uncomplete alphabet type that just offers a complement function.

test/unit/range/container/bitcompressed_vector_test.cpp
```diff
@@ -7,6 +7,35 @@
...
+#include <seqan3/range/views/complement.hpp>
+
 #include "container_test_template.hpp"
 
 INSTANTIATE_TYPED_TEST_SUITE_P(bitcompressed, container_over_dna4_test, seqan3::bitcompressed_vector<seqan3::dna4>, );
+
+TEST(bitcompressed_vector_test, issue1743_complement_on_proxy)
+{ // https://github.com/seqan/seqan3/issues/1743
+
+    seqan3::bitcompressed_vector<seqan3::dna4> v{seqan3::dna4{}, seqan3::dna4{}};
```
rrahn, line 21:
```suggestion
    seqan3::bitcompressed_vector<seqan3::dna4> v(2);
```

test/unit/range/container/bitcompressed_vector_test.cpp
```diff
@@ -7,6 +7,35 @@
...
+    auto complement = seqan3::complement(proxy);
+
+    EXPECT_TRUE((std::same_as<decltype(complement), seqan3::dna4>));
+    EXPECT_EQ(complement, seqan3::assign_rank_to(3, seqan3::dna4{}));
+}
+
+TEST(bitcompressed_vector_test, issue1743_view_combinability)
+{ // https://github.com/seqan/seqan3/issues/1743
+
+    seqan3::bitcompressed_vector<seqan3::dna4> v{seqan3::dna4{}, seqan3::dna4{}, seqan3::dna4{}, seqan3::dna4{}};
```
rrahn, line 33:
```suggestion
    seqan3::bitcompressed_vector<seqan3::dna4> v(4);
```

test/unit/range/container/bitcompressed_vector_test.cpp
```diff
@@ -7,6 +7,35 @@
...
+
+    auto complement = v | seqan3::views::complement;
+
+    EXPECT_EQ(v.size(), complement.size());
+
+    for (auto chr : complement)
+        EXPECT_EQ(chr, seqan3::assign_rank_to(3, seqan3::dna4{}));
```
rrahn, line 40:
```suggestion
    RANGE_EXPECT_EQ(complement, seqan3::dna4_vector{'T'_dna4, 'G'_dna4, 'C'_dna4, 'A'_dna4});
```
And make the vector above:

```cpp
seqan3::dna4_vector{'A'_dna4, 'C'_dna4, 'G'_dna4, 'T'_dna4}
```

test/unit/range/container/container_test_template.hpp
smehringer: I explicitly named the test template `over_dna4` to raise attention to this "extremely weird" test but I think refactoring the whole container test should not be part of this PR.

smehringer:
> I have the feeling that common_with<dna4, bitcompressed_proxy> will always trigger this ICE or infinite compiler loop and should then be fixed.

The following works:
```
TEST(bitcompressed_vector_test, common_with_ice)
{
    EXPECT_TRUE((std::common_with<seqan3::bitcompressed_vector<seqan3::dna4>::reference, seqan3::dna4>));
    EXPECT_TRUE((std::common_with<seqan3::dna4, seqan3::bitcompressed_vector<seqan3::dna4>::reference>));
}
```

test/unit/range/container/container_test_template.hpp
rrahn: Please add at least a ticket to refactor the container test. We have other ideas to also make a view test template.

test/unit/range/container/bitcompressed_vector_test.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+TEST(bitcompressed_vector_test, issue1743_view_combinability)
+{ // https://github.com/seqan/seqan3/issues/1743
+
+    seqan3::bitcompressed_vector<seqan3::dna4> v{'A'_dna4, 'C'_dna4, 'G'_dna4, 'T'_dna4};
+
+    auto complement = v | seqan3::views::complement;
+
+    EXPECT_EQ(v.size(), complement.size());
+
```
rrahn, line 41:
```suggestion
```

test/unit/range/container/bitcompressed_vector_test.cpp
```diff
@@ -0,0 +1,43 @@
...
+    EXPECT_TRUE((std::same_as<decltype(complement), seqan3::dna4>));
+    EXPECT_EQ(complement, 'T'_dna4);
+}
+
+TEST(bitcompressed_vector_test, issue1743_view_combinability)
+{ // https://github.com/seqan/seqan3/issues/1743
+
+    seqan3::bitcompressed_vector<seqan3::dna4> v{'A'_dna4, 'C'_dna4, 'G'_dna4, 'T'_dna4};
+
```
rrahn, line 37:
```suggestion
```

test/unit/range/container/bitcompressed_vector_test.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include "container_test_template.hpp"
+
+INSTANTIATE_TYPED_TEST_SUITE_P(bitcompressed, container_over_dna4_test, seqan3::bitcompressed_vector<seqan3::dna4>, );
+
+using seqan3::operator""_dna4;
+
+TEST(bitcompressed_vector_test, issue1743_complement_on_proxy)
+{ // https://github.com/seqan/seqan3/issues/1743
+
```
rrahn, line 23:
```suggestion
```

test/unit/range/container/bitcompressed_vector_test.cpp
```diff
@@ -0,0 +1,43 @@
...
+    auto proxy = *v.begin();
+    auto complement = seqan3::complement(proxy);
+
+    EXPECT_TRUE((std::same_as<decltype(complement), seqan3::dna4>));
+    EXPECT_EQ(complement, 'T'_dna4);
+}
+
+TEST(bitcompressed_vector_test, issue1743_view_combinability)
+{ // https://github.com/seqan/seqan3/issues/1743
+
```
rrahn, line 35:
```suggestion
```

test/unit/range/container/container_test_template.hpp
smehringer: https://github.com/seqan/product_backlog/issues/150

PR #1927, opened 2020-06-23
[TEST] add regression test for https://github.com/ericniebler/range-v…

marehr:
…3/issues/1514

https://github.com/ericniebler/range-v3/issues/1514

### DIFF
--- a/include/seqan3/std/ranges
+++ b/include/seqan3/std/ranges
@@ -42,6 +42,21 @@ template<class T>
 inline constexpr bool enable_view<T> = true;
 } // namespace std::ranges
 
+namespace ranges::detail
+{
+// gcc10 uses for std::ranges::range_difference_t<std::ranges::iota_view<size_t, size_t>> == __int128
+#if __SIZEOF_INT128__
+__extension__ typedef __int128 int128_t;
+#if RANGES_CXX_INLINE_VARIABLES >= RANGES_CXX_INLINE_VARIABLES_17
+template<>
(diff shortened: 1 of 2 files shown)


include/seqan3/std/ranges
```diff
@@ -42,6 +42,21 @@ template<class T>
 inline constexpr bool enable_view<T> = true;
 } // namespace std::ranges
 
+namespace ranges::detail
+{
+// gcc10 uses for std::ranges::range_difference_t<std::ranges::iota_view<size_t, size_t>> == __int128
+#if __SIZEOF_INT128__
```
eseiler, line 48: Do we actually support anything that does not offer 128 bit integers?
AFAIK GCC and clang offer them, but MSVC does not

eseiler: Just asking because we did this:
https://github.com/seqan/seqan3/blob/5831b57a5bcbbf5a03d9ff725d91ca2783a8898b/include/seqan3/search/dream_index/interleaved_bloom_filter.hpp#L214-L228
Which is obviously wrong if you consider that the index might serialised with 128 bit support and then read somewhere without 128 bit support - both cases produce different results (modulo != fastrange).
But since this currently cannot happen (no MSVC support), it doesn't have high priority...

marehr: Normally gcc, if you enable strict mode, does not support this, because it is not standard c++. (I don't know how to enable strict mode, I forgot but I think it was called `_STRICT_ANSI_` or so.)

PR #1929, opened 2020-06-23
[MISC] Update range library

marehr:
A couple of PRs

* https://github.com/ericniebler/range-v3/pull/1500
* https://github.com/ericniebler/range-v3/pull/1494
* https://github.com/ericniebler/range-v3/pull/1493

got merged that are needed for C++20 support.


wvandertoorn:
What is the usual way to do this, should I review the code in Eric Niebler's repo or is this review request only a formality in light of the two-review-policy?

marehr:
These are external dependencies and as long as our builds are not breaking it is okay to green light.

Sometime we have to also change our code to make it compatible, but this time we could upgrade without changing anything, so this review is only a formality in light of the two-review-policy :)

PR #1908, opened 2020-06-13
wip traceback alignment

rrahn:
The draft PR for the traceback computation in the new alignment. 
solves seqan/product_backlog#74

PR #1934, opened 2020-06-25
[MISC] Use alignment builder policy in banded alignment algorithm.

rrahn:
part of #1908 

Rebases the banded alignment on the current master using the alignment result policy.

### DIFF
--- a/include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
+++ b/include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
@@ -99,17 +99,15 @@ public:
     //!\endcond
     void operator()(indexed_sequence_pairs_t && indexed_sequence_pairs, callback_t && callback)
     {
-        using result_value_t = typename alignment_result_value_type_accessor<alignment_result_type>::type;
         using std::get;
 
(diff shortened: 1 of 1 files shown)


include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
```diff
@@ -99,17 +99,15 @@ class pairwise_alignment_algorithm_banded :
...
 
         for (auto && [sequence_pair, idx] : indexed_sequence_pairs)
         {
-            result_value_t res{};
-            res.id = idx;
-
             compute_matrix(get<0>(sequence_pair), get<1>(sequence_pair));
-            res.score = this->tracked_optimum();
-            callback(alignment_result_type{res});
+            this->make_result_and_invoke(std::forward<decltype(sequence_pair)>(sequence_pair),
```
marehr, line 107: This means we have shared code between different alignment algorithms that we have to keep track of?

rrahn: Not really. Because at the end I can dry this out again. I just need to wait until I have all dependencies.

PR #1935, opened 2020-06-25
Optimise alignment/part4 introduce alignment matrix policy

rrahn:
part of #1908 

Introduces a policy to configure and acquire the alignment matrices. 
It also unifies the initialisation interface for the banded and unbanded alignment algorithm.
This prepares the code to allow different alignment matrix implementations later. Thus, if the alignment is required the alignment matrix will be augmented with the trace matrix to store the trace pointer. Later we want to have different trace pointer implementations, for example to only store the trace checkpoints in the wavefront execution mode.

### DIFF
--- a/include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
+++ b/include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
@@ -137,11 +137,21 @@ public:
     //!\endcond
     void operator()(indexed_sequence_pairs_t && indexed_sequence_pairs, callback_t && callback)
     {
+        using matrix_index_t = typename traits_type::matrix_index_type;
         using std::get;
+
+        thread_local score_matrix_single_column<score_type> alignment_matrix{};
+        coordinate_matrix<matrix_index_t> index_matrix{};
 
         for (auto && [sequence_pair, idx] : indexed_sequence_pairs)
         {
-            compute_matrix(get<0>(sequence_pair), get<1>(sequence_pair));
+            size_t number_of_columns = std::ranges::distance(get<0>(sequence_pair)) + 1;
+            size_t number_of_rows = std::ranges::distance(get<1>(sequence_pair)) + 1;
+
+            alignment_matrix.resize(column_index_type{number_of_columns}, row_index_type{number_of_rows});
+            index_matrix.resize(column_index_type{number_of_columns}, row_index_type{number_of_rows});
+
+            compute_matrix(get<0>(sequence_pair), get<1>(sequence_pair), alignment_matrix, index_matrix);
             this->make_result_and_invoke(std::forward<decltype(sequence_pair)>(sequence_pair),
                                          std::move(idx),
                                          this->optimal_score,
@@ -160,6 +170,7 @@ public:
     {
         using simd_collection_t = std::vector<score_type, aligned_allocator<score_type, alignof(score_type)>>;
         using original_score_t = typename traits_type::original_score_type;
+        using matrix_index_t = typename traits_type::matrix_index_type;
 
         // Extract the batch of sequences for the first and the second sequence.
         auto seq1_collection = indexed_sequence_pairs | views::get<0> | views::get<0>;
@@ -174,7 +185,16 @@ public:
         convert_batch_of_sequences_to_simd_vector(simd_seq1_collection, seq1_collection, traits_type::padding_symbol);
         convert_batch_of_sequences_to_simd_vector(simd_seq2_collection, seq2_collection, traits_type::padding_symbol);
 
-        compute_matrix(simd_seq1_collection, simd_seq2_collection);
+        thread_local score_matrix_single_column<score_type> alignment_matrix{};
+        coordinate_matrix<matrix_index_t> index_matrix{};
+
+        size_t const number_of_columns = std::ranges::distance(simd_seq1_collection) + 1;
+        size_t const number_of_rows = std::ranges::distance(simd_seq2_collection) + 1;
+
+        alignment_matrix.resize(column_index_type{number_of_columns}, row_index_type{number_of_rows});
+        index_matrix.resize(column_index_type{number_of_columns}, row_index_type{number_of_rows});
+
+        compute_matrix(simd_seq1_collection, simd_seq2_collection, alignment_matrix, index_matrix);
 
         size_t index = 0;
         for (auto && [sequence_pair, idx] : indexed_sequence_pairs)
@@ -232,31 +252,37 @@ protected:
     /*!\brief Compute the actual alignment.
      * \tparam sequence1_t The type of the first sequence; must model std::ranges::forward_range.
      * \tparam sequence2_t The type of the second sequence; must model std::ranges::forward_range.
+     * \tparam alignment_matrix_t The type of the alignment matrix; must model std::ranges::input_range and its
+     *                            std::ranges::range_reference_t type must model std::ranges::forward_range.
+     * \tparam index_matrix_t The type of the index matrix; must model std::ranges::input_range and its
+     *                            std::ranges::range_reference_t type must model std::ranges::forward_range.
      *
      * \param[in] sequence1 The first sequence to compute the alignment for.
      * \param[in] sequence2 The second sequence to compute the alignment for.
+     * \param[in] alignment_matrix The alignment matrix to compute.
+     * \param[in] index_matrix The index matrix corresponding to the alignment matrix.
      */
-    template <std::ranges::forward_range sequence1_t, std::ranges::forward_range sequence2_t>
-    void compute_matrix(sequence1_t && sequence1, sequence2_t && sequence2)
+    template <std::ranges::forward_range sequence1_t,
+              std::ranges::forward_range sequence2_t,
+              std::ranges::input_range alignment_matrix_t,
+              std::ranges::input_range index_matrix_t>
+    //!\cond
+        requires std::ranges::forward_range<std::ranges::range_reference_t<alignment_matrix_t>> &&
+                 std::ranges::forward_range<std::ranges::range_reference_t<index_matrix_t>>
+    //!\endcond
+    void compute_matrix(sequence1_t && sequence1,
+                        sequence2_t && sequence2,
+                        alignment_matrix_t && alignment_matrix,
+                        index_matrix_t && index_matrix)
     {
         // ---------------------------------------------------------------------
         // Initialisation phase: allocate memory and initialise first column.
         // ---------------------------------------------------------------------
-        using matrix_index_t = typename traits_type::matrix_index_type;
 
         this->reset_optimum(); // Reset the tracker for the new alignment computation.
 
-        thread_local score_matrix_single_column<score_type> local_score_matrix{};
-        coordinate_matrix<matrix_index_t> local_index_matrix{};
-
-        size_t number_of_columns = std::ranges::distance(sequence1) + 1;
-        size_t number_of_rows = std::ranges::distance(sequence2) + 1;
-
-        local_score_matrix.resize(column_index_type{number_of_columns}, row_index_type{number_of_rows});
-        local_index_matrix.resize(column_index_type{number_of_columns}, row_index_type{number_of_rows});
-
-        auto alignment_matrix_it = local_score_matrix.begin();
-        auto indexed_matrix_it = local_index_matrix.begin();
+        auto alignment_matrix_it = alignment_matrix.begin();
+        auto indexed_matrix_it = index_matrix.begin();
 
         initialise_column(*alignment_matrix_it, *indexed_matrix_it, sequence2);
 
--- a/include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
+++ b/include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
@@ -165,7 +165,6 @@ public:
     {
         using simd_collection_t = std::vector<score_type, aligned_allocator<score_type, alignof(score_type)>>;
         using original_score_t = typename traits_type::original_score_type;
-        using matrix_index_t = typename traits_type::matrix_index_type;
 
         // Extract the batch of sequences for the first and the second sequence.
         auto seq1_collection = indexed_sequence_pairs | views::get<0> | views::get<0>;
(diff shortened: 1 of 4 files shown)


include/seqan3/alignment/pairwise/detail/policy_alignment_matrix.hpp
```diff
@@ -149,5 +154,23 @@ class policy_alignment_matrix
...
+     * \param[in] sequence1_size The size of the first sequence.
+     * \param[in] sequence2_size The size of the second sequence.
+     *
+     * \throws seqan3::invalid_alignment_configuration if the band is invalid for the given sequence sizes and the
+     *         alignment configuration.
+     */
+    void check_valid_band_configuration(size_t const sequence1_size, size_t const sequence2_size) const
+    {
+        bool const upper_diagonal_ends_before_last_cell = (upper_diagonal + sequence2_size) < sequence1_size;
+        bool const lower_diagonal_ends_behind_last_cell = (-lower_diagonal + sequence1_size) < sequence2_size;
```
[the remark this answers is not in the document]
marehr: Imagine having 1mio sequences and the last one is the odd-one-out, i.e. the band does not fit, and you get an exception that you can't handle gracefully (Even if you handle that exception, you can't resume the computation). This would be horrible! I would prefer to give an "infinite" score back and an empty alignment, instead of some exception. I suggested over and over again modelling the `infinite` score either as `std::optional` or as `the largest int score`, because of this and now you hack something that does not account for this and it will be a harder and harder to get this semantics into the code.

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
```diff
@@ -71,22 +66,10 @@ class pairwise_alignment_algorithm_banded :
...
-        auto band = get<seqan3::align_cfg::band_fixed_size>(config);
-
-        lower_diagonal = band.lower_diagonal.get();
-        upper_diagonal = band.upper_diagonal.get();
-
-        if (upper_diagonal < lower_diagonal || upper_diagonal < 0 || lower_diagonal > 0)
-            throw invalid_alignment_configuration{"The selected band [" + std::to_string(lower_diagonal) + ":" +
-                                                  std::to_string(upper_diagonal) + "] is not supported because it "
-                                                  "does not completely encloses both sequences."};
-    }
```
marehr, line 89: Okay you removed that here again. Without much thought, is the sequence invariant check more general than this?\\

I think the `upper_diagonal < lower_diagonal` is the only invariant that is a user error, and we should warn him about this before starting the computation.

rrahn: No, the user error would also to provide a band which does not cover the first cell and a global alignment is requested. A local alignment or a semi-global alignment would be valid though.

include/seqan3/alignment/pairwise/detail/policy_alignment_matrix.hpp
```diff
@@ -60,15 +66,37 @@ class policy_alignment_matrix
...
+    policy_alignment_matrix(alignment_configuration_t const & config)
+    {
+        using seqan3::get;
+
+        auto band = config.get_or(seqan3::align_cfg::band_fixed_size{});
+
+        lower_diagonal = band.lower_diagonal.get();
+        upper_diagonal = band.upper_diagonal.get();
+
+        if (upper_diagonal < lower_diagonal || upper_diagonal < 0 || lower_diagonal > 0)
```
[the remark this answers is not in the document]
rrahn: No, this is not a valid band in the global alignment. And something else is yet not supported. This is merely one step towards the alignment computation. I have some orthogonal changes which I have to rebase then. Whatever comes first.

include/seqan3/alignment/pairwise/detail/policy_alignment_matrix.hpp
```diff
@@ -94,15 +123,29 @@ class policy_alignment_matrix
                           size_t sequence2_size,
                           score_type initial_score = score_type{}) const
     {
+        assert(sequence1_size < std::numeric_limits<int64_t>::max());
+        assert(sequence2_size < std::numeric_limits<int64_t>::max());
```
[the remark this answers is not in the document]
rrahn: I changed it to uint64_t because as far as I know size_t does not guarantee 64 bits but is platform dependent.

include/seqan3/alignment/pairwise/detail/policy_alignment_matrix.hpp
```diff
@@ -94,15 +123,29 @@ class policy_alignment_matrix
...
+
+            size_t const band_size = std::min<int64_t>(upper_diagonal - lower_diagonal + 1, sequence2_size);
+            alignment_matrix.resize(column_index_type{sequence1_size}, row_index_type{band_size}, initial_score);
+        }
+        else
+        {
+            alignment_matrix.resize(column_index_type{sequence1_size}, row_index_type{sequence2_size}, initial_score);
+        }
+
         alignment_matrix.resize(column_index_type{sequence1_size}, row_index_type{sequence2_size}, initial_score);
```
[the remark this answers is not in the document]
rrahn: I don't understand what you mean?

include/seqan3/alignment/pairwise/detail/policy_alignment_matrix.hpp
```diff
@@ -94,15 +123,29 @@ class policy_alignment_matrix
...
+        if constexpr (traits_t::is_banded)
+        {
+            assert(upper_diagonal - lower_diagonal + 1 > 0);
+
+            size_t const band_size = std::min<int64_t>(upper_diagonal - lower_diagonal + 1, sequence2_size);
+            alignment_matrix.resize(column_index_type{sequence1_size}, row_index_type{band_size}, initial_score);
+        }
+        else
+        {
+            alignment_matrix.resize(column_index_type{sequence1_size}, row_index_type{sequence2_size}, initial_score);
```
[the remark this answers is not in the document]
rrahn: Ok, I see the point. I will make a separate PR getting all of them straight.

include/seqan3/alignment/pairwise/detail/policy_alignment_matrix.hpp
```diff
@@ -149,5 +154,23 @@ class policy_alignment_matrix
...
+
+    /*!\brief Checks whether the band is valid for the given sequence sizes.
+     *
+     * \param[in] sequence1_size The size of the first sequence.
+     * \param[in] sequence2_size The size of the second sequence.
+     *
+     * \throws seqan3::invalid_alignment_configuration if the band is invalid for the given sequence sizes and the
+     *         alignment configuration.
+     */
+    void check_valid_band_configuration(size_t const sequence1_size, size_t const sequence2_size) const
```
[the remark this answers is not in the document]
rrahn: exactly. There are two potential states and the one can only be evaluated when the sequences are known.

include/seqan3/alignment/pairwise/detail/policy_alignment_matrix.hpp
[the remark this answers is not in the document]
rrahn: As I said in the other PR. The plan after the features are there is to not throw at all but propagate the error state. This has the ultimate advantage, that we don't throw in multi-threaded library code but the user has the full control over it.

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
rrahn: > I think the upper_diagonal < lower_diagonal is the only invariant that is a user error, and we should warn him about this before starting the computation.

But we can't do it inside of configuration object because the parameters can be set separately. So it throws during configuration.

include/seqan3/alignment/pairwise/detail/policy_alignment_matrix.hpp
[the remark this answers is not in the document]
rrahn: I am not sure what you want me to do. These two band checks are already done only in one class.

rrahn:
@marehr there were a lot of discussion and some changes. I hope I got everything adapted.

include/seqan3/alignment/pairwise/detail/policy_alignment_matrix.hpp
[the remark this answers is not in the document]
marehr: You are doing `alignment_matrix.resize` twice, I don't know if this is your intention

include/seqan3/alignment/pairwise/detail/policy_alignment_matrix.hpp
```diff
@@ -0,0 +1,172 @@
...
+    {
+        bool const upper_diagonal_ends_before_last_cell = (upper_diagonal + sequence2_size) < sequence1_size;
+        bool const lower_diagonal_ends_behind_last_cell = (-lower_diagonal + sequence1_size) < sequence2_size;
+
+        if (upper_diagonal_ends_before_last_cell || lower_diagonal_ends_behind_last_cell)
+            throw invalid_alignment_configuration{"The selected band [" + std::to_string(lower_diagonal) + ":" +
+                                                  std::to_string(upper_diagonal) + "] does not cover the last cell."};
```
marehr, line 169: The same here, only "validity checks" that concern global alignment, wrapped in `if constexpr global alignment`

include/seqan3/alignment/pairwise/detail/policy_alignment_matrix.hpp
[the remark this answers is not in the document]
marehr: This is still open :)

rrahn:
@marehr this done now. I will rebase when you give green light.

include/seqan3/alignment/pairwise/detail/policy_alignment_matrix.hpp
```diff
@@ -138,14 +142,15 @@ class policy_alignment_matrix
         size_t const column_count = sequence1_size + 1;
         size_t row_count = sequence2_size + 1;
 
+        alignment_matrix.resize(column_index_type{column_count}, row_index_type{row_count}, initial_score);
+
```
marehr, line 146:
```suggestion
```

Why do you need this twice? :joy:

include/seqan3/alignment/pairwise/detail/policy_alignment_matrix.hpp
```diff
@@ -0,0 +1,182 @@
...
+        if constexpr (traits_t::is_banded)
+        {
+            assert(upper_diagonal - lower_diagonal + 1 > 0);
+
+            row_count = std::min<int64_t>(upper_diagonal - lower_diagonal + 1, row_count);
+        }
+
+        alignment_matrix.resize(column_index_type{column_count}, row_index_type{row_count}, initial_score);
```
marehr, line 154: You are doing it already here.

include/seqan3/alignment/pairwise/detail/policy_alignment_matrix.hpp
rrahn: It was supposed to be the index matrix :)

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
```diff
@@ -101,9 +101,26 @@ class pairwise_alignment_algorithm_banded :
     {
         using std::get;
 
+        thread_local score_matrix_single_column<int32_t> alignment_matrix{};
+        coordinate_matrix<uint32_t> index_matrix{};
+        size_t const band_size = upper_diagonal - lower_diagonal + 1;
```
smehringer:
```suggestion
        assert(upper_diagonal >= lower_diagonal);
        size_t const band_size = upper_diagonal - lower_diagonal + 1;
```

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
```diff
@@ -141,9 +141,18 @@ class pairwise_alignment_algorithm : protected policies_t...
...
         using std::get;
 
+        thread_local score_matrix_single_column<int32_t> alignment_matrix{};
+        coordinate_matrix<uint32_t> index_matrix{};
+
         for (auto && [sequence_pair, idx] : indexed_sequence_pairs)
         {
-            compute_matrix(get<0>(sequence_pair), get<1>(sequence_pair));
+            size_t number_of_columns = std::ranges::distance(get<0>(sequence_pair)) + 1;
+            size_t number_of_rows = std::ranges::distance(get<1>(sequence_pair)) + 1;
```
smehringer:
```suggestion
            size_t const number_of_columns = std::ranges::distance(get<0>(sequence_pair)) + 1;
            size_t const number_of_rows = std::ranges::distance(get<1>(sequence_pair)) + 1;
```

include/seqan3/alignment/pairwise/detail/policy_alignment_matrix.hpp
```diff
@@ -92,10 +92,14 @@ class policy_alignment_matrix
...
-        if (upper_diagonal < lower_diagonal || upper_diagonal < 0 || lower_diagonal > 0)
+        bool invalid_band = upper_diagonal < lower_diagonal;
+        if constexpr (traits_t::is_global)
+            invalid_band |= (upper_diagonal < 0 || lower_diagonal > 0);
+
+        if (invalid_band)
             throw invalid_alignment_configuration{"The selected band [" + std::to_string(lower_diagonal) + ":" +
-                                                  std::to_string(upper_diagonal) + "] is not supported because it "
-                                                  "does not completely encloses both sequences."};
+                                                  std::to_string(upper_diagonal) + "] cannot be used with the current "
```
smehringer, line 101: Did I request this in another PR? please add a note why the band is invalid in the error message

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
```diff
@@ -71,22 +66,10 @@ class pairwise_alignment_algorithm_banded :
      *
      * Initialises the algorithm given the user settings from the alignment configuration object.
      *
-     * \throws seqan3::invalid_alignment_configuration if the given band settings are invalid.
+     * \throws seqan3::invalid_alignment_configuration.
```
smehringer, line 68:
```suggestion
```
?

rrahn: Well, it still throws, but it happens in the base policy constructor now.

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
rrahn: this is in the final version already fixed.

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
rrahn: this is in the final version already fixed.

rrahn:
@smehringer There is only the documentation of the error message. So I would already rebase since travis needs ages to get through.

smehringer:
sure!

PR #1937, opened 2020-06-25
Optimise alignment/part5 banded alignment with free end gaps

rrahn:
part of #1908 

Enables the free end gaps feature for banded alignments in new optimised alignment code.

### DIFF
--- a/include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
+++ b/include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
@@ -82,10 +82,15 @@ public:
         lower_diagonal = band.lower_diagonal.get();
         upper_diagonal = band.upper_diagonal.get();
 
-        if (upper_diagonal < lower_diagonal || upper_diagonal < 0 || lower_diagonal > 0)
+        // Band is invalid if ...
+        bool const invalid_band = upper_diagonal < lower_diagonal || // upper diagonal is smaller than lower diagonal,
+                                  (upper_diagonal < 0 && !this->first_column_is_free) || // band starts in first column but does not use free ends,
+                                  (lower_diagonal > 0 && !this->first_row_is_free); // band starts in first row but does not use free ends.
+
+        if (invalid_band)
             throw invalid_alignment_configuration{"The selected band [" + std::to_string(lower_diagonal) + ":" +
-                                                  std::to_string(upper_diagonal) + "] is not supported because it "
-                                                  "does not completely encloses both sequences."};
+                                                  std::to_string(upper_diagonal) + "] is not valid for the "
+                                                  "configured alignment because the optimum cannot be computed."};
     }
     //!\}
 
@@ -127,7 +132,10 @@ protected:
         bool const upper_diagonal_ends_before_last_cell = (upper_diagonal + sequence2_size) < sequence1_size;
         bool const lower_diagonal_ends_behind_last_cell = (-lower_diagonal + sequence1_size) < sequence2_size;
 
-        if (upper_diagonal_ends_before_last_cell || lower_diagonal_ends_behind_last_cell)
+        bool const invalid_band = (lower_diagonal_ends_behind_last_cell && !this->test_last_column_cell) || // band ends in last column but does not use free ends,
+                                  (upper_diagonal_ends_before_last_cell && !this->test_last_row_cell); // band ends in last row but does not use free ends.
+
+        if (invalid_band)
             throw invalid_alignment_configuration{"The selected band [" + std::to_string(lower_diagonal) + ":" +
                                                   std::to_string(upper_diagonal) + "] does not cover the last cell."};
     }
@@ -252,7 +260,7 @@ protected:
 
         this->track_last_column_cell(*alignment_column_it, *cell_index_column_it);
 
-        for (; first_row_index < sequence2_size; ++first_row_index)
+        for (size_t last_row = std::min(sequence2_size, row_size); first_row_index < last_row; ++first_row_index)
             this->track_last_column_cell(*++alignment_column_it, *++cell_index_column_it);
 
         this->track_final_cell(*alignment_column_it, *cell_index_column_it);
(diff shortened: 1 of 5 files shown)


include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
```diff
@@ -82,10 +82,15 @@ class pairwise_alignment_algorithm_banded :
...
+        bool const invalid_band = upper_diagonal < lower_diagonal || // upper diagonal is smaller than lower diagonal,
+                                  (upper_diagonal < 0 && !this->first_column_is_free) || // band starts in first column but does not use free ends,
+                                  (lower_diagonal > 0 && !this->first_row_is_free); // band starts in first row but does not use free ends.
+
+        if (invalid_band)
             throw invalid_alignment_configuration{"The selected band [" + std::to_string(lower_diagonal) + ":" +
-                                                  std::to_string(upper_diagonal) + "] is not supported because it "
-                                                  "does not completely encloses both sequences."};
+                                                  std::to_string(upper_diagonal) + "] is not valid for the "
+                                                  "configured alignment because the optimum cannot be computed."};
```
marehr, line 93: This conflicts with https://github.com/seqan/seqan3/pull/1935/files#diff-a374649bb6a2078f62e2bbc24c7014ffL77-L89, right?

What is your ultimate version?

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
```diff
@@ -128,7 +133,10 @@ class pairwise_alignment_algorithm_banded :
         bool const upper_diagonal_ends_before_last_cell = (upper_diagonal + sequence2_size) < sequence1_size;
         bool const lower_diagonal_ends_behind_last_cell = (-lower_diagonal + sequence1_size) < sequence2_size;
 
-        if (upper_diagonal_ends_before_last_cell || lower_diagonal_ends_behind_last_cell)
+        bool const invalid_band = (lower_diagonal_ends_behind_last_cell && !this->last_column_is_free) || // band ends in last column but does not use free ends,
+                                  (upper_diagonal_ends_before_last_cell && !this->last_row_is_free); // band ends in last row but does not use free ends.
+
+        if (invalid_band)
             throw invalid_alignment_configuration{"The selected band [" + std::to_string(lower_diagonal) + ":" +
                                                   std::to_string(upper_diagonal) + "] does not cover the last cell."};
```
marehr, line 141: Same here, this conflicts with https://github.com/seqan/seqan3/pull/1935/files#diff-a374649bb6a2078f62e2bbc24c7014ffL126-L134

What is your ultimate version?

include/seqan3/alignment/pairwise/detail/policy_affine_gap_recursion.hpp
```diff
@@ -97,13 +97,10 @@ class policy_affine_gap_recursion
...
-        {
-            // front_end_first, back_end_first, front_end_second, back_end_second
-            auto align_ends_config = get<align_cfg::aligned_ends>(config).value;
-            first_row_is_free = align_ends_config[0];
-            first_column_is_free = align_ends_config[2];
-        }
+
+        auto align_ends_config = config.get_or(align_cfg::aligned_ends{free_ends_none}).value;
+        first_row_is_free = align_ends_config[0];
+        first_column_is_free = align_ends_config[2];
```
marehr, line 103: This is nice :)

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
```diff
@@ -253,7 +261,7 @@ class pairwise_alignment_algorithm_banded :
 
         this->track_last_column_cell(*alignment_column_it, *cell_index_column_it);
 
-        for (; first_row_index < sequence2_size; ++first_row_index)
+        for (; first_row_index < std::min(sequence2_size, row_size); ++first_row_index)
```
marehr, line 264:
```suggestion
        for (size_t last_row = std::min(sequence2_size, row_size); first_row_index < last_row; ++first_row_index)
```

This min is "technically" called in every loop iteration

test/unit/alignment/pairwise/semi_global_affine_banded_test.cpp
```diff
@@ -22,3 +22,88 @@ using pairwise_semiglobal_affine_banded_testing_types = ::testing::Types<
...
+    auto config = base_config() |
+                  seqan3::align_cfg::band_fixed_size{seqan3::align_cfg::lower_diagonal{-8},
+                                                     seqan3::align_cfg::upper_diagonal{-1}} |
+                  seqan3::align_cfg::aligned_ends{seqan3::end_gaps{seqan3::front_end_first{std::true_type{}},
+                                                                   seqan3::back_end_first{std::true_type{}},
+                                                                   seqan3::front_end_second{std::false_type{}},
+                                                                   seqan3::back_end_second{std::true_type{}}}};
+
+    EXPECT_THROW((align_pairwise(std::tie(sequence1, sequence2), config)),
+                 seqan3::invalid_alignment_configuration);
```
marehr, line 68: I don't know if all of these throws are really user errors.

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
rrahn: The ultimate version is to not throw at all inside of the alignment algorithm but rather set an error state inside of the alignment result. Then the user can ask if an error state was set or in case he calls a member function while the error state is set it will throw. This allows us to not worry about correctly populating exceptions in threaded code, which is everything but trivial. Then the user has to make sure his code is handling the exceptions correctly. Until then, I want to put this into common structures to not repeat the code.

test/unit/alignment/pairwise/semi_global_affine_banded_test.cpp
rrahn: Well, the begin yes the end point maybe not so. But the mid-term plan is to wrap these kind of states into a error state of the result object.

rrahn: Is that ok, to postpone this to the end with all features available and then make the error state?

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
rrahn: I agree, but I will make it more verbose in #1935 because it has a more appropriate structure to do this.

PR #1939, opened 2020-06-27
[TEST] add new test case and make existing ones more fool-proof

marehr:
Part of https://github.com/seqan/product_backlog/issues/62

---

If ranges were not correctly implemented, the tests

* move_forward_pre_test
* move_forward_post_test
* move_backward_pre_test (split from move_backward_test)
* move_backward_post_test (split from move_backward_test)

could have an infinite loop, if the iterators didn't move to their end.

We now iterate over the expected range to make sure the test terminates.

This commit further adds the test case `difference_sentinel` which tests
`begin - end` and `end - begin` with all const/non-const combinations.

### DIFF
--- a/test/unit/range/iterator_test_template.hpp
+++ b/test/unit/range/iterator_test_template.hpp
@@ -200,7 +200,8 @@ inline void move_forward_pre_test(it_begin_t && it_begin, it_sentinel_t && it_en
 {
     // pre-increment
     auto rng_it = std::ranges::begin(rng);
-    for (auto it = it_begin; it != it_end; ++it, ++rng_it)
+    auto rng_it_end = std::ranges::end(rng);
+    for (auto it = it_begin; it != it_end && rng_it != rng_it_end; ++it, ++rng_it)
         expext_eq<test_type>(*it, *rng_it);
 }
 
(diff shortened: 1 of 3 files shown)


test/unit/range/iterator_test_template.hpp
```diff
@@ -209,7 +210,8 @@ inline void move_forward_post_test(it_begin_t && it_begin, it_sentinel_t && it_e
 {
     // post-increment
     auto rng_it = std::ranges::begin(rng);
-    for (auto it = it_begin; it != it_end; it++, ++rng_it)
+    auto rng_it_end = std::ranges::end(rng);
+    for (auto it = it_begin; it != it_end && rng_it !=rng_it_end; it++, ++rng_it)
```
MitraDarja, line 214:
```suggestion
    for (auto it = it_begin; it != it_end && rng_it != rng_it_end; it++, ++rng_it)
```

test/unit/range/iterator_test_template.hpp
```diff
@@ -334,7 +381,7 @@ inline void jump_backward_test(it_begin_t && it_begin, rng_t && rng)
 {
     size_t sz = std::ranges::distance(rng);
 
-    auto pre_end_it = it_begin + sz - 1;
+    auto pre_end_it = it_begin + (sz - 1);
```
MitraDarja, line 384: Why did you add the brackets?

marehr: Because it is inefficient :see_no_evil: and two different use cases.

`it_begin + sz - 1` is the same as `(it_begin + sz) - 1`, that means we call `it_begin.operator+(sz).operator-(1)` instead of `it_begin.operator+(sz - 1)`.

MitraDarja: Ah okay, thanks for explaining me this. :)

PR #1943, opened 2020-06-29
[INFRA] Add gcc-10 -std=c++2a support; fixes seqan/product_backlog#61

marehr:
Part of  seqan/product_backlog#61

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,12 @@ jobs:
         - BUILD=unit
         - BUILD_TYPE=Release
         - CXXFLAGS="-std=c++17 -fconcepts"
+    - name: "Unit gcc10 (c++20)"
+      env:
+        - CXX=g++-10
+        - CC=gcc-10
+        - BUILD=unit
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -44,6 +44,12 @@ jobs:
         - BUILD=unit
         - BUILD_TYPE=Release
         - CXXFLAGS="-std=c++17 -fconcepts"
+    - name: "Unit gcc10 (c++2a)"
```
eseiler, line 47:
```suggestion
    - name: "Unit gcc10 (c++20)"
```
?

marehr: It is technically only C++2a, because the standard wasn't finalised, yet.

eseiler: Ok, was wondering, because you can pass `-std=c++20` to GCC10

marehr: Okay, if it is like that we use -std=c++20 instead :)

marehr:
@eseiler Thank you for figuring out what caused the problem.

PR #1949, opened 2020-07-01
[FIX] Explicit return type in transform lambda to avoid hard compiler error.

rrahn:
Part of #1908 

When the return type for the lambda is not given the entire body must be evaluated in the overload resolution of the transform view.
There also the const qualified version is checked by the compiler. If inside the body of the lambda function is
an expression that results in a constraint error the compiler will omit an hard error because he cannot deduce the return type.
To fix this, the return type is given explicitly in order to allow SFINAE friendly substituion errors due to incomplete types in the return type.


smehringer:
The buffer queue seems to fail sporadically on jenkins...

smehringer:
You can run Jenkins again or just merge as you like

PR #1952, opened 2020-07-01
[FIX] gcc10: use std::variant::emplace instead of implicit assignment

marehr:
Our PPC CI uses 10.1.0 release from source and it triggered an error
which most distros seems to have hot-patched.

I encountered that problem in the past and used this workaround, but the
problem was fixed in the meantime so that I did not need to submit this
patch.

The patch is confirmed to work by @eseiler

### DIFF
--- a/include/seqan3/io/detail/misc.hpp
+++ b/include/seqan3/io/detail/misc.hpp
@@ -82,7 +82,7 @@ void set_format(format_variant_type & format,
             {
                 if (std::ranges::equal(ext, extension))
                 {
-                    format = fm_type{};
+                    format.template emplace<fm_type>();
                     format_found = true;
                     return;
                 }


smehringer:
Hm this time Jenkins actually says 3 tests out of 100 fail. 

```
The following tests FAILED:
	4568 - contrib/parallel/buffer_queue_parallel_test::buffer_queue.spmc_dynamicsize (Timeout)
	4569 - contrib/parallel/buffer_queue_parallel_test::buffer_queue.spmc_fixedsize (Timeout)
	4850 - core/parallel/detail/reader_writer_manager_test::reader_writer_manager.parallel (Timeout)
```

It is unrelated right?

eseiler:
> Hm this time Jenkins actually says 3 tests out of 100 fail.
> 
> ```
> The following tests FAILED:
> 	4568 - contrib/parallel/buffer_queue_parallel_test::buffer_queue.spmc_dynamicsize (Timeout)
> 	4569 - contrib/parallel/buffer_queue_parallel_test::buffer_queue.spmc_fixedsize (Timeout)
> 	4850 - core/parallel/detail/reader_writer_manager_test::reader_writer_manager.parallel (Timeout)
> ```
> 
> It is unrelated right?

I reran and yes, unrelated

PR #1954, opened 2020-07-02
[FIX] kmer_hash size works even with take_until | reverse

### LINKED ISSUE #1953: [BUG] kmer_hash size function does not work with take_until and reverse
After finding out that ` seqan3::views::take_until | std::views::reverse` can be a sized_range and that is expected behabiour (see #1946 ), I still encountered problems when using this with the minimiser_view, I believe the actual reason for that is the following error:

### Expected behaviour

```cpp
static constexpr auto ungapped_view = seqan3::views::kmer_hash(seqan3::ungapped{3});
static constexpr auto prefix_until_first_thymine = seqan3::views::take_until([] (seqan3::dna4 x)
                                                   { return x == 'T'_dna4; });
std::vector<seqan3::dna4> text1{'A'_dna4, 'C'_dna4, 'G'_dna4, 'T'_dna4, 'A'_dna4, 'G'_dna4, 'C'_dna4}; // ACGTAGC
auto v = text1 | prefix_until_first_thymine | std::views::reverse | ungapped_view;
v.size(); // should return 1
```

### Actual behaviour

It does not compile with the error message: 
```
error: no match for call to ‘(const ranges::_size_::fn) (const ranges::reverse_view<seqan3::detail::view_take_until<ranges::ref_view<std::vector<seqan3::dna4> >, <lambda(seqan3::dna4)>, false, false> >&)’
         using size_type = decltype(std::ranges::size(urange));

```

### Solution
I don't know why, but removing the const from kmer_hash's size function solves the problem.

MitraDarja:
resolves #1953

### DIFF
--- a/include/seqan3/range/views/kmer_hash.hpp
+++ b/include/seqan3/range/views/kmer_hash.hpp
@@ -181,9 +181,19 @@ public:
     auto size()
     //!\cond
         requires std::ranges::sized_range<urng_t>
+     //!\endcond
+    {
+        using size_type = std::ranges::range_size_t<urng_t>;
+        return std::max<size_type>(std::ranges::size(urange) + 1, shape_.size()) - shape_.size();
+    }
+
+    //!\copydoc size()
+    auto size() const
+    //!\cond
+        requires std::ranges::sized_range<urng_t const>
     //!\endcond
     {
-        using size_type = decltype(std::ranges::size(urange));
+        using size_type = std::ranges::range_size_t<urng_t const>;
         return std::max<size_type>(std::ranges::size(urange) + 1, shape_.size()) - shape_.size();
     }
 };
--- a/include/seqan3/range/views/kmer_hash.hpp
+++ b/include/seqan3/range/views/kmer_hash.hpp
@@ -178,7 +178,7 @@ public:
     /*!\brief Returns the size of the range, if the underlying range is a std::ranges::sized_range.
      * \returns Size of range.
      */
-    auto size() const
+    auto size()
     //!\cond
         requires std::ranges::sized_range<urng_t>
     //!\endcond
--- a/test/unit/range/views/view_kmer_hash_test.cpp
+++ b/test/unit/range/views/view_kmer_hash_test.cpp
@@ -230,3 +230,19 @@ TYPED_TEST(kmer_hash_gapped_test, issue1754)
         EXPECT_RANGE_EQ(result_t{8}, text1 | stop_at_t | std::views::reverse | gapped_view);
     }
 }
+
+// https://github.com/seqan/seqan3/issues/1953
+TEST(kmer_hash_ungapped_test, issue1953)
+{
+    std::vector<seqan3::dna4> text1{'A'_dna4, 'C'_dna4, 'G'_dna4, 'T'_dna4, 'A'_dna4, 'G'_dna4, 'C'_dna4}; // ACGTAGC
+    auto v = text1 | prefix_until_first_thymine | std::views::reverse | ungapped_view;
+    EXPECT_EQ(1u, v.size());
+}
+
+// https://github.com/seqan/seqan3/issues/1953
+TEST(kmer_hash_gapped_test, issue1953)
+{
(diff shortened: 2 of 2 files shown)


include/seqan3/range/views/kmer_hash.hpp
```diff
@@ -178,7 +178,7 @@ class kmer_hash_view : public std::ranges::view_interface<kmer_hash_view<urng_t>
     /*!\brief Returns the size of the range, if the underlying range is a std::ranges::sized_range.
      * \returns Size of range.
      */
-    auto size() const
+    auto size()
```
marehr, line 181:
```suggestion
    auto size()
    //!\cond
        requires std::ranges::sized_range<urng_t>
    //!\endcond
    {
        using size_type = decltype(std::ranges::size(urange));
        return std::max<size_type>(std::ranges::size(urange) + 1, shape_.size()) - shape_.size();
    }

    auto size() const
    //!\cond
        requires std::ranges::sized_range<urng_t const>
    //!\endcond
    {
        using size_type = decltype(std::ranges::size(urange));
        return std::max<size_type>(std::ranges::size(urange) + 1, shape_.size()) - shape_.size();
    }
```

We should propagate `const` if we can.

MitraDarja:
```c++
auto && v1 = text1;
auto && v2 = text1 | prefix_until_first_thymine;
auto && v3 = text1 | prefix_until_first_thymine | std::views::reverse;
auto && v4 = text1 | prefix_until_first_thymine | std::views::reverse | ungapped_view;

std::ranges::size(std::as_const(v1)); -> Compiles
std::ranges::size(std::as_const(v2)); -> Does not compile
std::ranges::size(std::as_const(v3)); -> Does not compile, compiles without std::as_const
std::ranges::size(std::as_const(v4)); -> Does not compile, compiles without std::as_const
```

MitraDarja:
@marehr *polite ping*

marehr:
And another new gcc issue, your use case seems to the bane for the standard library xD

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=96113

test/unit/range/views/view_kmer_hash_test.cpp
```diff
@@ -230,3 +230,19 @@ TYPED_TEST(kmer_hash_gapped_test, issue1754)
         EXPECT_RANGE_EQ(result_t{8}, text1 | stop_at_t | std::views::reverse | gapped_view);
     }
 }
+
+// https://github.com/seqan/seqan3/issues/1953
+TEST(kmer_hash_ungapped_test, issue1953)
+{
+    std::vector<seqan3::dna4> text1{'A'_dna4, 'C'_dna4, 'G'_dna4, 'T'_dna4, 'A'_dna4, 'G'_dna4, 'C'_dna4}; // ACGTAGC
+    auto v = text1 | prefix_until_first_thymine | std::views::reverse | ungapped_view;
```
smehringer, line 238:
```suggestion
    auto v = text1 | prefix_until_first_thymine | std::views::reverse | ungapped_view;
    auto const v_const = text1 | prefix_until_first_thymine | std::views::reverse | ungapped_view;
    EXPECT_EQ(1u, v_const.size());
```
Does this work?

MitraDarja: No, I get then the following error:

"error: passing ‘const seqan3::detail::kmer_hash_view<ranges::reverse_view<seqan3::detail::view_take_until<ranges::ref_view<std::vector<seqan3::dna4> >, <lambda(seqan3::dna4)>, false, false> > >’ as ‘this’ argument discards qualifiers [-fpermissive]
"

Is this unexpected?

marehr: I already asked that: It doesn't.

marehr: As it turns out it shouldn't work, because `std::views::reverse` drops const-iterability unless the view is a common range and this is not the case here. So it is expected that only the non-const version works.

smehringer: thanks for the error message & explanation you both :)

PR #1964, opened 2020-07-03
Cbegin cend

wvandertoorn:
Resolves https://github.com/seqan/product_backlog/issues/143 and  https://github.com/seqan/product_backlog/issues/142.

* `cbegin()` and `cend()` are removed from views, because standard views don't have them.
*  For all views `std::ranges::cbegin` and `std::ranges::cend` are used in the const overload of `begin()` and `end()`.

For `persist_view` a const overload of `begin()` and `end()` with return type `auto` were required to make this work.

### DIFF
--- a/include/seqan3/range/views/enforce_random_access.hpp
+++ b/include/seqan3/range/views/enforce_random_access.hpp
@@ -94,15 +94,6 @@ public:
         return enforced_random_access_iterator<decltype(std::ranges::cbegin(urng))>{std::ranges::cbegin(urng)};
     }
 
-    //!\copydoc seqan3::detail::view_enforce_random_access::begin
-    constexpr auto cbegin() const noexcept
-    //!\cond
-        requires const_iterable_range<urng_t>
-    //!\endcond
-    {
-        return begin();
-    }
-
     /*!\brief Returns the sentinel to the end of the range.
      *
      * \details
@@ -129,15 +120,6 @@ public:
             return enforced_random_access_iterator<decltype(std::ranges::cend(urng))>{std::ranges::cend(urng)};
         else
             return std::ranges::cend(urng);
-    }
-
-    //!\copydoc seqan3::detail::view_enforce_random_access::end
-    constexpr auto cend() const noexcept
-    //!\cond
-        requires const_iterable_range<urng_t>
-    //!\endcond
-    {
-        return end();
     }
     //!\}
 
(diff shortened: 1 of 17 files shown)


include/seqan3/range/views/persist.hpp
```diff
@@ -94,15 +85,18 @@ class view_persist : public std::ranges::view_interface<view_persist<urng_t>>
...
 
     //!\copydoc begin()
-    const_iterator cbegin() const noexcept
+    auto begin() const noexcept
+    //!\cond
+        requires const_iterable_range<urng_t>
+    //!\endcond
     {
-        return begin();
+        return std::ranges::cbegin(*urange);
```
marehr, line 99: @rrahn, @smehringer, @eseiler This was a nice side effect of enforcing this rule, because `urange` is a pointer in this case and the const overload only has the effect that `urange` is now const (i.e. the pointer), but it is only shallow const, because it points to a non-const object `*urange`. This means `std::views::persist` never propagated the constness of the underlying range :)

include/seqan3/range/views/enforce_random_access.hpp
```diff
@@ -114,7 +105,7 @@ class view_enforce_random_access : public std::ranges::view_interface<view_enfor
     constexpr auto end() noexcept
     {
         if constexpr (std::ranges::common_range<urng_t>)
-            return enforced_random_access_iterator<decltype(std::ranges::end(urng))>{std::ranges::end(urng)};
+            return enforced_random_access_iterator<decltype(std::ranges::cend(urng))>{std::ranges::cend(urng)};
```
eseiler, line 108: I think this is wrong, because this is not the const qualified `end()`, the next `end()` is const and alrady uses `std::ranges::cend`.

wvandertoorn: You are right. I'm sorry, fixed.

include/seqan3/range/views/persist.hpp
rrahn: But this was right, since the persist view has pointer semantic similar to the ref_view of the standard. It does not modify the underlying range it only places it on the heap so that it allows to pipe rvalue non-view ranges. So it feels like it should be closer to the ref_view than to other views. But that is only a quick thought on this. Maybe we should discuss this.

marehr: Core-Meeting:

we'll keep deep-constness (the current solution)

marehr:
Upps! We discussed this: https://github.com/orgs/seqan/projects/4#card-41487782, but forgot to come back.

We decided that the changes are fine as they are.

PR #1965, opened 2020-07-03
Fix gcc10 error

### LINKED ISSUE #1963: [BUG] In g++-10 seqan3::views::complement | seqan3::views::kmer_hash forward_range property is not preserved
### Description 
Using `seqan3::views::complement | seqan3::views::kmer_hash` loses the forward_range property, it should preserve it though and with g++-version below 10 it does.

### How to repeat the problem

Compile with g++-10:
```c++
seqan3::dna5_vector text{"ACGTA"_dna5};
static_assert(std::ranges::forward_range<decltype(text | seqan3::views::complement | seqan3::views::kmer_hash(seqan3::ungapped{2}))>); // leads to a static assertion, which is not what we would expect
```

MitraDarja:
Resolves #1963 

Not sure, why this solves this, but this enforces that a bidirectional range stays bidirectional (and also stays a forward range). I also changed the first condition from contigous to random_access, otherwise a bitcompressed_vector would not result in a random access range, alternatively one could add another std::conditional and check first contigous, than random_access and then bidirectional, I am not sure what behaviour we want.

I also changed the parameter type in test of 1754 because the type was wrong.

### DIFF
--- a/include/seqan3/range/views/kmer_hash.hpp
+++ b/include/seqan3/range/views/kmer_hash.hpp
@@ -241,9 +241,11 @@ public:
     //!\brief Tag this class as input iterator.
     using iterator_category = typename std::iterator_traits<it_t>::iterator_category;
     //!\brief Tag this class depending on which concept `it_t` models.
-    using iterator_concept = std::conditional_t<std::contiguous_iterator<it_t>,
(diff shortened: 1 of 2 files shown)


marehr:
Superseded by #1968

Issue #1962, opened 2020-07-02
Parallel search not working


ddurai:
## Platform

  - SeqAn version: seqAn3
  - Operating system: Linux ddurai 4.15.0-101-generic #102-Ubuntu SMP Mon May 11 10:07:26 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

  - Compiler:7.5.0

## Question

Dear SeqAn developers

I am trying to develop a correction algorithm for which I am using seqAn3. As the part of the algorithm, I have to search a string in an index. i am trying to parallelize my search using the following search config:

```cpp
#include <seqan3/core/algorithm/configuration.hpp>
#include <seqan3/search/configuration/max_error.hpp>
#include <seqan3/search/configuration/parallel.hpp>

seqan3::configuration const cfg = seqan3::search_cfg::max_error_total{seqan3::search_cfg::error_count{2}} |
                                  seqan3::search_cfg::max_error_substitution{seqan3::search_cfg::error_count{2}} |
                                  seqan3::search_cfg::parallel{8};
```

However, i encounter the following error:

```
/path_to_seqan/seqan3/include/seqan3/search/configuration/parallel.hpp:33:79: error: ‘search_config_id’ is not a member of ‘seqan3::search_cfg::detail’
 using parallel = seqan3::detail::parallel_mode<std::integral_constant<detail::search_config_id,

/path_to_seqan/seqan3/include/seqan3/search/configuration/parallel.hpp:34:79: error: ‘search_config_id’ is not a member of ‘seqan3::search_cfg::detail’
                                                                       detail::search_config_id::parallel>>;
```
Can you please help me with the issue

regards
Dilip

eseiler:
Hello Dilip,

the snippet you posted compiles with GCC 7.5.0 on the current master branch.
What version of SeqAn3 are you using?

ddurai:
Hi eseiler,
I am using seqan3 version 3.0.1 downloaded from github

eseiler:
Thanks!

If you're using 3.0.1, the correct syntax would be
```cpp
#include <seqan3/core/algorithm/configuration.hpp>
#include <seqan3/search/configuration/max_error.hpp>
#include <seqan3/search/configuration/parallel.hpp>

seqan3::configuration const cfg = seqan3::search_cfg::max_error{seqan3::search_cfg::total{2},
                                                                seqan3::search_cfg::substitution{2}} |
                                  seqan3::search_cfg::parallel{8};
```

[...]

PR #1917, opened 2020-06-17
[MISC] Config parallel with member thread_count

joergi-w:
This PR adds a member variable `thread_count` to the parallel config (it replaces the old value member). It influences both `seqan3::align_cfg::parallel` and `seqan3::search_cfg::parallel`.

Documentation and tests are included in the second commit.

Closes https://github.com/seqan/product_backlog/issues/68 and contributes to https://github.com/seqan/product_backlog/issues/131.

PR #1969, opened 2020-07-06
[FIX] Broken parallel search

eseiler:
Resolves https://github.com/seqan/seqan3/issues/1962

When the parallel search was added in #1880, no tests were added. Consequently, #1917 did not trigger errors for the wrong config.

### DIFF
--- a/test/unit/search/search_collection_test.cpp
+++ b/test/unit/search/search_collection_test.cpp
@@ -172,6 +172,41 @@ TYPED_TEST(search_test, multiple_queries)
     EXPECT_RANGE_EQ(search(queries, this->index, cfg) | query_id, expected_query_ids);
 }
 
+TYPED_TEST(search_test, parallel_queries)
+{
+    constexpr size_t num_queries{100u};
+    std::vector<std::vector<seqan3::dna4>> const queries{num_queries, {"ACGTACGTACGT"_dna4}};
+
+    seqan3::configuration const cfg = seqan3::search_cfg::max_error_total{seqan3::search_cfg::error_rate{.0}} |
+                                      seqan3::search_cfg::max_error_substitution{seqan3::search_cfg::error_rate{.0}} |
+                                      seqan3::search_cfg::max_error_insertion{seqan3::search_cfg::error_rate{.0}} |
+                                      seqan3::search_cfg::max_error_deletion{seqan3::search_cfg::error_rate{.0}} |
+                                      seqan3::search_cfg::parallel{
+                                          std::min<uint32_t>(2, std::thread::hardware_concurrency())};
+
+    typename TestFixture::hits_result_t expected_hits{}; // Expected hits are {0, 0} and {1, 0} for each of the queries.
+    std::vector<size_t> expected_query_ids{}; // Expected query ids are 0,0,1,1,...,num_queries-1,num_queries-1.
+    expected_hits.reserve(2 * num_queries);
+    expected_query_ids.reserve(2 * num_queries);
+    for (size_t i = 0; i < num_queries; ++i)
+    {
+        expected_hits.emplace_back(0, 0);
+        expected_hits.emplace_back(1, 0);
+        expected_query_ids.emplace_back(i);
+        expected_query_ids.emplace_back(i);
+    }
+
+    EXPECT_RANGE_EQ(search(queries, this->index, cfg) | ref_id_and_position, expected_hits);
+    EXPECT_RANGE_EQ(search(queries, this->index, cfg) | query_id, expected_query_ids);
+}
+
+TYPED_TEST(search_test, parallel_without_parameter)
+{
+    seqan3::configuration cfg = seqan3::search_cfg::parallel{};
+
+    EXPECT_THROW(search("AAAA"_dna4, this->index, cfg), std::runtime_error);
+}
+
 TYPED_TEST(search_string_test, error_free_string)
 {
     typename TestFixture::hits_result_t empty_result{};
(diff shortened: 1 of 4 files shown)


rrahn:
Great, many thanks. Can you also add a more robust parallel execution test inside of the search framework. Something, where we can test at least the execution of 100 queries or so in parallel mode?

test/unit/search/search_test.cpp
```diff
@@ -106,6 +106,20 @@ TYPED_TEST(search_test, multiple_queries)
...
+    std::vector<std::vector<seqan3::dna4>> const queries{100, {"ACGTACGTACGT"_dna4}};
+
+    seqan3::configuration const cfg = seqan3::search_cfg::max_error_total{seqan3::search_cfg::error_rate{.0}} |
+                                      seqan3::search_cfg::max_error_substitution{seqan3::search_cfg::error_rate{.0}} |
+                                      seqan3::search_cfg::max_error_insertion{seqan3::search_cfg::error_rate{.0}} |
+                                      seqan3::search_cfg::max_error_deletion{seqan3::search_cfg::error_rate{.0}} |
+                                      seqan3::search_cfg::parallel{2};
+
+    EXPECT_RANGE_EQ(search(queries, this->index, cfg) | query_id, std::views::iota(0, 100));
+    EXPECT_RANGE_EQ(search(queries, this->index, cfg) | position, std::vector(100, 0));
```
eseiler, line 120: Just as a note:
`std::vector(100, 0)` cannot be `std::vector{100, 0}`, because then it would use the initialiser list constructor (=[100,0]) instead of the "count, value" constructor (=[0,0,...,0]).

wvandertoorn:
I don't fully understand the "[FIX] No conversion from ranges::detail::diffmax_t" commit. What does it change?

eseiler:
> I don't fully understand the "[FIX] No conversion from ranges::detail::diffmax_t" commit. What does it change with regard to the parallel search?

The conversion here:
```cpp
if constexpr (std::same_as<execution_handler_t, execution_handler_parallel>)
        if constexpr (std::same_as<execution_handler_t, execution_handler_parallel>)
            buffer_size = std::ranges::distance(resource); // << this
```
failed.
Apparently there is no valid implicit conversion in this assignment to get from `ranges::detail::diffmax_t` to `size_t`; although the conversion works for the alignment parallel config. 
So I needed to explicitly cast the value.

wvandertoorn:
Alright, interesting. Thanks for the clarification

test/unit/search/search_collection_test.cpp
```diff
@@ -172,6 +172,40 @@ TYPED_TEST(search_test, multiple_queries)
...
+TYPED_TEST(search_test, parallel_queries)
+{
+    std::vector<std::vector<seqan3::dna4>> const queries{100, {"ACGTACGTACGT"_dna4}};
+
+    seqan3::configuration const cfg = seqan3::search_cfg::max_error_total{seqan3::search_cfg::error_rate{.0}} |
+                                      seqan3::search_cfg::max_error_substitution{seqan3::search_cfg::error_rate{.0}} |
+                                      seqan3::search_cfg::max_error_insertion{seqan3::search_cfg::error_rate{.0}} |
+                                      seqan3::search_cfg::max_error_deletion{seqan3::search_cfg::error_rate{.0}} |
+                                      seqan3::search_cfg::parallel{2};
+
```
rrahn, line 184:
```suggestion
```

test/unit/search/search_collection_test.cpp
```diff
@@ -172,6 +172,40 @@ TYPED_TEST(search_test, multiple_queries)
...
 
+TYPED_TEST(search_test, parallel_queries)
+{
+    std::vector<std::vector<seqan3::dna4>> const queries{100, {"ACGTACGTACGT"_dna4}};
+
+    seqan3::configuration const cfg = seqan3::search_cfg::max_error_total{seqan3::search_cfg::error_rate{.0}} |
+                                      seqan3::search_cfg::max_error_substitution{seqan3::search_cfg::error_rate{.0}} |
+                                      seqan3::search_cfg::max_error_insertion{seqan3::search_cfg::error_rate{.0}} |
+                                      seqan3::search_cfg::max_error_deletion{seqan3::search_cfg::error_rate{.0}} |
+                                      seqan3::search_cfg::parallel{2};
```
rrahn, line 183:
```suggestion
                                      seqan3::search_cfg::parallel{std::min<size_t>(2, std::thread::hardware_concurrency())};
```

test/unit/search/search_collection_test.cpp
```diff
@@ -172,6 +172,40 @@ TYPED_TEST(search_test, multiple_queries)
     EXPECT_RANGE_EQ(search(queries, this->index, cfg) | query_id, expected_query_ids);
 }
 
+TYPED_TEST(search_test, parallel_queries)
+{
+    std::vector<std::vector<seqan3::dna4>> const queries{100, {"ACGTACGTACGT"_dna4}};
```
rrahn, line 177: Please, make the size a constexpr variable and use it in the test, so that it only needs to be changed once.,

test/unit/search/search_collection_test.cpp
```diff
@@ -174,20 +174,21 @@ TYPED_TEST(search_test, multiple_queries)
 
 TYPED_TEST(search_test, parallel_queries)
 {
-    std::vector<std::vector<seqan3::dna4>> const queries{100, {"ACGTACGTACGT"_dna4}};
+    constexpr size_t const num_queries{100u};
```
rrahn, line 177:
```suggestion
    constexpr size_t num_queries{100u};
```
That is already a constant expression :)

test/unit/search/search_test.cpp
```diff
@@ -108,16 +108,18 @@ TYPED_TEST(search_test, multiple_queries)
 
 TYPED_TEST(search_test, parallel_queries)
 {
-    std::vector<std::vector<seqan3::dna4>> const queries{100, {"ACGTACGTACGT"_dna4}};
+    constexpr size_t const num_queries{100u};
```
rrahn, line 111:
```suggestion
    constexpr size_t num_queries{100u};
```

PR #1973, opened 2020-07-08
[FIX] Make char_is_valid member function public for alphabet_variant.

### LINKED ISSUE #1972: [Bug] seqan3::is_in_alphabet has wrong results with gapped alphabets
## Platform

  - SeqAn version: 3.0.2
  - Operating system: macOS 10.15.5
  - Compiler: gcc9

## Description

The implementation for `seqan3::is_in_alphabet` is not correct for gapped alphabets.
I would expect to see the correct results or a compiler error if this case is not implemented.

Hint for a possible solution: There is a function `bool char_is_valid(char_type const chr)` in `alphabet_variant.hpp:578` which is supposed to do the job, but it is actually never called.

### How to repeat the problem

Simply paste the lines below in a snippet like this:

```cpp
using namespace seqan3;
std::cout << std::boolalpha << (is_in_alphabet<gapped<rna5>>('a')) << std::endl;
```

### Expected behaviour

```cpp
is_in_alphabet<gapped<rna5>>('a')    // true
is_in_alphabet<gapped<rna5>>('c')    // true
is_in_alphabet<gapped<rna5>>('g')    // true
is_in_alphabet<gapped<rna5>>('u')    // true
is_in_alphabet<gapped<rna5>>('-')    // true
is_in_alphabet<gapped<rna5>>('n')    // true
is_in_alphabet<gapped<rna5>>('s')    // false
```

### Actual behaviour

```cpp
is_in_alphabet<gapped<rna5>>('a')    // false
is_in_alphabet<gapped<rna5>>('c')    // false
is_in_alphabet<gapped<rna5>>('g')    // false
is_in_alphabet<gapped<rna5>>('u')    // false
is_in_alphabet<gapped<rna5>>('-')    // true
is_in_alphabet<gapped<rna5>>('n')    // false
is_in_alphabet<gapped<rna5>>('s')    // false
```

smehringer:
Fixes #1972 

Note that `char_is_valid` is public in all other alphabets. I double checked.

### DIFF
--- a/include/seqan3/alphabet/composite/alphabet_variant.hpp
+++ b/include/seqan3/alphabet/composite/alphabet_variant.hpp
@@ -433,6 +433,20 @@ public:
     }
     //!\}
 
+    //!\brief Validate whether a character is valid in the by any of the combined alphabet.
+    static constexpr bool char_is_valid(char_type const chr) noexcept
+    {
+        bool is_valid{false};
+
+        meta::for_each(alternatives{}, [&] (auto && alt)
+        {
+            if (char_is_valid_for<remove_cvref_t<decltype(alt)>>(chr))
+                is_valid = true;
+        });
+
+        return is_valid;
+    }
+
 protected:
     //!\privatesection
 
@@ -573,20 +587,6 @@ protected:
 
         return char_to_rank;
     }();
-
-    //!\brief Validate whether a character is valid in the by any of the combined alphabet.
-    static constexpr bool char_is_valid(char_type const chr) noexcept
-    {
-        bool is_valid{false};
-
-        meta::for_each(alternatives{}, [&] (auto && alt)
-        {
-            if (char_is_valid_for<remove_cvref_t<decltype(alt)>>(chr))
-                is_valid = true;
-        });
-
-        return is_valid;
-    }
 };
 
 } // namespace seqan3
--- a/include/seqan3/alphabet/composite/alphabet_variant.hpp
+++ b/include/seqan3/alphabet/composite/alphabet_variant.hpp
@@ -433,7 +433,7 @@ public:
     }
     //!\}
 
-    //!\brief Validate whether a character is valid in the by any of the combined alphabet.
+    //!\brief Validate whether a character is valid in the combined alphabet.
     static constexpr bool char_is_valid(char_type const chr) noexcept
     {
         bool is_valid{false};
--- a/test/unit/alphabet/composite/alphabet_variant_test.cpp
+++ b/test/unit/alphabet/composite/alphabet_variant_test.cpp
@@ -335,7 +335,8 @@ TEST(alphabet_variant_test, two_different_variants)
 }
 
 TEST(alphabet_variant_test, char_is_valid_for)
-{ // see issue https://github.com/seqan/seqan3/issues/1972
+{
+    // see issue https://github.com/seqan/seqan3/issues/1972
     EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('A')); // valid seqan3::rna5 char
     EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('a')); // valid seqan3::rna5 char
     EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('-')); // valid seqan3::gap char
--- a/test/unit/alphabet/composite/alphabet_variant_test.cpp
+++ b/test/unit/alphabet/composite/alphabet_variant_test.cpp
@@ -336,38 +336,28 @@ TEST(alphabet_variant_test, two_different_variants)
 
 TEST(alphabet_variant_test, char_is_valid_for)
 {
-    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('A'));
-    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('C'));
-    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('G'));
-    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('U'));
-    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('T'));
-    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('N'));
-    EXPECT_FALSE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('S'));
-    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('a'));
-    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('c'));
-    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('g'));
-    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('u'));
-    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('t'));
-    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('-'));
-    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('n'));
-    EXPECT_FALSE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('s'));
+    using char_t = seqan3::alphabet_char_t<seqan3::gapped<seqan3::rna5>>;
+
+    char_t i = std::numeric_limits<char_t>::min();
+    char_t j = std::numeric_limits<char_t>::max();
+
+    for (; i < j; ++i)
+    {
+        EXPECT_EQ(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>(i),
+                  seqan3::char_is_valid_for<seqan3::gap>(i) || seqan3::char_is_valid_for<seqan3::rna5>(i));
+    }
 }
 
 TEST(alphabet_variant_test, is_in_alphabet)
 { // see issue https://github.com/seqan/seqan3/issues/1972
-    EXPECT_TRUE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('A'));
-    EXPECT_TRUE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('C'));
-    EXPECT_TRUE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('G'));
-    EXPECT_TRUE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('U'));
-    EXPECT_TRUE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('T'));
-    EXPECT_TRUE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('N'));
-    EXPECT_FALSE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('S'));
-    EXPECT_TRUE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('a'));
-    EXPECT_TRUE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('c'));
-    EXPECT_TRUE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('g'));
-    EXPECT_TRUE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('u'));
-    EXPECT_TRUE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('t'));
-    EXPECT_TRUE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('-'));
-    EXPECT_TRUE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('n'));
-    EXPECT_FALSE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('s'));
+    using char_t = seqan3::alphabet_char_t<seqan3::gapped<seqan3::rna5>>;
+
+    char_t i = std::numeric_limits<char_t>::min();
+    char_t j = std::numeric_limits<char_t>::max();
+
+    for (; i < j; ++i)
+    {
+        EXPECT_EQ(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>(i),
+                  seqan3::is_in_alphabet<seqan3::gap>(i) || seqan3::is_in_alphabet<seqan3::rna5>(i));
+    }
 }
--- a/test/unit/alphabet/composite/alphabet_variant_test.cpp
+++ b/test/unit/alphabet/composite/alphabet_variant_test.cpp
@@ -336,12 +336,18 @@ TEST(alphabet_variant_test, two_different_variants)
 
 TEST(alphabet_variant_test, char_is_valid_for)
 {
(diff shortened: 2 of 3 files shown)


include/seqan3/alphabet/composite/alphabet_variant.hpp
```diff
@@ -433,6 +433,20 @@ class alphabet_variant : public alphabet_base<alphabet_variant<alternative_types
     }
     //!\}
 
+    //!\brief Validate whether a character is valid in the by any of the combined alphabet.
```
joergi-w, line 436:
```suggestion
    //!\brief Validate whether a character is valid in the combined alphabet.
```

test/unit/alphabet/composite/alphabet_variant_test.cpp
```diff
@@ -13,6 +13,7 @@
 #include <seqan3/alphabet/composite/alphabet_variant.hpp>
 #include <seqan3/alphabet/gap/gap.hpp>
 #include <seqan3/alphabet/nucleotide/all.hpp>
+#include <seqan3/core/char_operations/predicate.hpp>
```
marehr, line 16: Why do we need this?

test/unit/alphabet/composite/alphabet_variant_test.cpp
```diff
@@ -332,3 +333,41 @@ TEST(alphabet_variant_test, two_different_variants)
...
+    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('N'));
+    EXPECT_FALSE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('S'));
+    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('a'));
+    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('c'));
+    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('g'));
+    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('u'));
+    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('t'));
+    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('-'));
+    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('n'));
+    EXPECT_FALSE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('s'));
```
marehr, line 353:
```suggestion
    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('A')); // valid seqan3::rna5 char 
    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('-')); // valid seqan3::gap char 
    
    EXPECT_FALSE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('S')); // neither seqan3::rna5 nor seqan3::gap
```

test/unit/alphabet/composite/alphabet_variant_test.cpp
```diff
@@ -332,3 +333,41 @@ TEST(alphabet_variant_test, two_different_variants)
...
+    EXPECT_TRUE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('N'));
+    EXPECT_FALSE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('S'));
+    EXPECT_TRUE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('a'));
+    EXPECT_TRUE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('c'));
+    EXPECT_TRUE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('g'));
+    EXPECT_TRUE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('u'));
+    EXPECT_TRUE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('t'));
+    EXPECT_TRUE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('-'));
+    EXPECT_TRUE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('n'));
+    EXPECT_FALSE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('s'));
```
marehr, line 372:
```suggestion
    EXPECT_TRUE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('A')); // valid seqan3::rna5 char
    EXPECT_TRUE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('-')); // valid seqan3::gap char
 
    EXPECT_FALSE(seqan3::is_in_alphabet<seqan3::gapped<seqan3::rna5>>('S')); // neither seqan3::rna5 nor seqan3::gap
```

joergi-w: We should have also lowercase characters in the test, because it seems to make a difference, as stated [here](https://github.com/seqan/seqan3/issues/1972#issuecomment-655297777).

test/unit/alphabet/composite/alphabet_variant_test.cpp
smehringer: Because `seqan3::is_in_alphabet` is defined in this header

test/unit/alphabet/composite/alphabet_variant_test.cpp
smehringer: I replaced them with a for loop over every possible char

test/unit/alphabet/composite/alphabet_variant_test.cpp
```diff
@@ -336,38 +336,28 @@ TEST(alphabet_variant_test, two_different_variants)
...
-    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('t'));
-    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('-'));
-    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('n'));
-    EXPECT_FALSE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('s'));
+    using char_t = seqan3::alphabet_char_t<seqan3::gapped<seqan3::rna5>>;
+
+    char_t i = std::numeric_limits<char_t>::min();
+    char_t j = std::numeric_limits<char_t>::max();
+
+    for (; i < j; ++i)
```
joergi-w, line 344: Nice idea! 
One remark: With `i < j` the value of `max()` is excluded from the loop.

test/unit/alphabet/composite/alphabet_variant_test.cpp
```diff
@@ -332,3 +333,31 @@ TEST(alphabet_variant_test, two_different_variants)
...
+    {
+        EXPECT_EQ(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>(i),
+                  seqan3::char_is_valid_for<seqan3::gap>(i) || seqan3::char_is_valid_for<seqan3::rna5>(i));
+    }
+}
+
+TEST(alphabet_variant_test, is_in_alphabet)
```
marehr, line 351: Move me into `char_predicate_test`.

Also have a second test that tests the specific equivalent classes.

test/unit/alphabet/composite/alphabet_variant_test.cpp
```diff
@@ -332,3 +333,31 @@ TEST(alphabet_variant_test, two_different_variants)
     seqan3::alphabet_variant<seqan3::rna4, seqan3::gap> r2{'A'_rna4};
     EXPECT_EQ(l, r2); // this works because rna4 and dna4 are implicitly convertible to each other
 }
+
+TEST(alphabet_variant_test, char_is_valid_for)
```
marehr, line 337: Can you re-add the specific tests of the equivalent classes? I would always prefer if we have a generic implementation and one that validates that for specific values. Thank you :)

```c++
    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('A')); // valid seqan3::rna5 char 
    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('-')); // valid seqan3::gap char 
    EXPECT_FALSE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('S')); // neither seqan3::rna5 nor seqan3::gap
```

marehr, line 337: It would be really awesome if we could use TYPED_TEST here, we would only need to somehow extract all bases.

EDIT://

Oh, we have that `::seqan3_required_types`, which is a `type_list<alphabets...>`

Maybe we should make this a public property :thinking:

test/unit/alphabet/composite/alphabet_variant_test.cpp
smehringer: I disagree :( Because this issue was not because `seqan3::is_in_alphabet` in predicate was wrong but because the alphabet_variant implementation had a flaw. So this is an issue of the alphabet_variant and should therefore be in its test.
Since `seqan3::is_in_alphabet` is only using `seqan3::char_is_valid_for` the test for the latter would suffice and we could remove the dependency on `predicate.hpp`. But we should nor make the predicate test depends on `alphabet_variant.hpp`, `dna4.hpp` and `gapped.hpp`.

test/unit/alphabet/composite/alphabet_variant_test.cpp
smehringer: this can be part of another PR?

test/unit/alphabet/composite/alphabet_variant_test.cpp
smehringer: The char_type is fixed here and its not 32 or 64 bits so I thought it's fine without the max iterations.

test/unit/alphabet/composite/alphabet_variant_test.cpp
smehringer: > Move me into char_predicate_test.

See https://github.com/seqan/seqan3/pull/1973#discussion_r452115051

> Also have a second test that tests the specific equivalent classes.

What do you mean by that?

marehr: This is a regression test for `seqan3::is_in_alphabet` and that lives in `test/unit/core/char_operations/char_predicate_test.cpp`

test/unit/alphabet/composite/alphabet_variant_test.cpp
```diff
@@ -332,3 +333,43 @@ TEST(alphabet_variant_test, two_different_variants)
...
+
+    char_t i = std::numeric_limits<char_t>::min();
+    char_t end = std::numeric_limits<char_t>::max();
+
+    // note that without the cast i <= end would result in an ininite loop
+    for (; static_cast<int64>(i) <= static_cast<int64>(end); ++i)
+    {
+        EXPECT_EQ(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>(i),
+                  seqan3::char_is_valid_for<seqan3::gap>(i) || seqan3::char_is_valid_for<seqan3::rna5>(i));
+    }
```
marehr, line 354:
```suggestion
    using gapped_alphabet_t = TypeParam;
    using gapped_alphabet_bases_t = typename gapped_alphabet_t::seqan3_required_types;
    using char_t = seqan3::alphabet_char_t<gapped_alphabet_t>;

    char_t i = std::numeric_limits<char_t>::min();
    char_t end = std::numeric_limits<char_t>::max();

    // note that without the cast i <= end would result in an ininite loop
    for (; static_cast<int64>(i) <= static_cast<int64>(end); ++i)
    {
        // is thar char valid for any of the bases?
        bool is_valid{};
        seqan3::detail::for_each<gapped_alphabet_bases_t>([&is_valid, i](auto id)
        {
             using type = typename decltype(id)::type;
             is_valid = is_valid || seqan3::char_is_valid_for<type>(i);
        });

        EXPECT_EQ(seqan3::char_is_valid_for<gapped_alphabet_t>(i), is_valid);
    }
```

Shouldn't this work? And we can make the test case a typed test?

test/unit/alphabet/composite/alphabet_variant_test.cpp
marehr: * What is that regression about? Oh, yeah about  `seqan3::is_in_alphabet` which belongs to the "test/unit/core/char_operations/char_predicate_test.cpp"
* Which test already includes *all* alphabets?
   hint:
   https://github.com/seqan/seqan3/blob/d06f9acae65e01375109f43f0bfa5202520a534f/test/unit/core/char_operations/char_predicate_test.cpp#L11-L14

The regression should be added where the regression belongs to.
Imagine someone who uses `seqan3::is_in_alphabet`, where will he look for tests? Sure, he will look in the `gapped` test, because it is obvious.

Please change this. Thank you!

test/unit/alphabet/composite/alphabet_variant_test.cpp
smehringer: I'll remind you of this conversation when I request a change and you insist that it is not the scope of the current PR

test/unit/alphabet/composite/alphabet_variant_test.cpp
```diff
@@ -332,3 +333,38 @@ TEST(alphabet_variant_test, two_different_variants)
     seqan3::alphabet_variant<seqan3::rna4, seqan3::gap> r2{'A'_rna4};
     EXPECT_EQ(l, r2); // this works because rna4 and dna4 are implicitly convertible to each other
 }
+
+TEST(alphabet_variant_test, char_is_valid_for)
+{ // see issue https://github.com/seqan/seqan3/issues/1972
+    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('A')); // valid seqan3::rna5 char
```
marehr, line 339:
```suggestion
TEST(alphabet_variant_test, char_is_valid_for)
{
    // see issue https://github.com/seqan/seqan3/issues/1972
    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('A')); // valid seqan3::rna5 char
```

test/unit/alphabet/composite/alphabet_variant_test.cpp
```diff
@@ -332,3 +333,38 @@ TEST(alphabet_variant_test, two_different_variants)
...
+    EXPECT_TRUE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('-')); // valid seqan3::gap char
+    EXPECT_FALSE(seqan3::char_is_valid_for<seqan3::gapped<seqan3::rna5>>('S')); // neither seqan3::rna5 nor seqan3::gap
+}
+
+template <typename T>
+using alphabet_variant_typed_test = ::testing::Test;
+
+TYPED_TEST_SUITE(alphabet_variant_typed_test, alphabet_variant_types, );
+
+TYPED_TEST(alphabet_variant_typed_test, char_is_valid_for)
```
marehr, line 350:
```suggestion
TYPED_TEST(alphabet_variant_test, char_is_valid_for)
```

test/unit/core/char_operations/char_predicate_test.cpp
```diff
@@ -439,3 +440,42 @@ TEST(char_predicate_, char_types)
         EXPECT_FALSE(seqan3::is_in_alphabet<seqan3::dna5>(char16_t{256}));
     }
 }
+
+TEST(char_predicate_, is_in_alphabet_alphabet_variant)
+{ // see issue https://github.com/seqan/seqan3/issues/1972
```
marehr, line 445:
```suggestion

// see issue https://github.com/seqan/seqan3/issues/1972
TEST(char_predicate_, issue1972)
{
```

test/unit/core/char_operations/char_predicate_test.cpp
```diff
@@ -439,3 +440,42 @@ TEST(char_predicate_, char_types)
...
+    {
+        bool is_valid{};
+        seqan3::detail::for_each<gapped_alphabet_bases_t>([&is_valid, i](auto id)
+        {
+             using type = typename decltype(id)::type;
+             is_valid = is_valid || seqan3::is_in_alphabet<type>(i);
+        });
+        EXPECT_EQ(seqan3::is_in_alphabet<gapped_alphabet_t>(i), is_valid);
+    }
+}
```
marehr, line 481:
```suggestion
```

PR #1974, opened 2020-07-08
[FEATURE] Argument parser file validator now allows compression extensions.

### LINKED ISSUE #1176: ArgParser: extension containing . accepted, but fail
In my attempt to support `.fasta.gz`, I added `fasta.gz` as an extension. But the argument parser tells me:
```
Validation failed for option -d/--database: Expected one of the following valid extensions: [fa,fq,fasta,fastq,fa.gz,fq.gz,fasta.gz,fastq.gz]! Got gz instead!
```

One could argue whether extensions containing the dot should be allowed, but I think they should. If not, there should be an error when setting up such an extension, not a silent failure.

In any case, we also need handling of compression formats.

smehringer:
Fixes #1176


smehringer:
We will do this differently

PR #1976, opened 2020-07-08
[FIX] to_chars for integer

### LINKED ISSUE #1595: Current LLVM copy of to_chars is wrong for integers with more than 9 digits

eseiler:
Resolves #1595


test/unit/std/charconv_int_test.cpp
```diff
@@ -231,14 +233,19 @@ TYPED_TEST(integral_from_char_test, hexadicimal_number)
...
-    EXPECT_EQ((std::string_view{buffer.data(), 3}), std::string_view{"120"});
+        auto res = std::to_chars(buffer.data(), buffer.data() + buffer.size(), val);
+
+        EXPECT_EQ(res.ptr, &buffer[num_digits]);
+        EXPECT_EQ(res.ec, std::errc{});
+        EXPECT_EQ((std::string_view{buffer.data(), num_digits}), std::string_view{std::to_string(val)});
+    }
 }
 
 TYPED_TEST(integral_from_char_test, to_chars_error)
```
marehr, line 251: Can you add the regression test?

```c++
TEST(to_chars_test, large_number)
{
    uint64_t val{123456789};
    std::array<char, 100> buffer{};

    auto res = std::to_chars(buffer.data(), buffer.data() + buffer.size(), val);

    EXPECT_EQ(res.ptr, &buffer[9]);
    EXPECT_EQ(res.ec, std::errc{});
    EXPECT_EQ((std::string_view{buffer.data(), 9}), std::string_view{"123456789"});
}
```
? Thank you :)

include/seqan3/contrib/charconv/charconv.cpp
```diff
@@ -176,25 +176,7 @@ __u64toa(uint64_t value, char* buffer)
...
-        else
-        {
-            if (v0 < 10000000)
-                buffer = append3(buffer, b0);
-            else
-                buffer = append4(buffer, b0);
-        }
-
-        buffer = append4(buffer, c0);
+        buffer = __u32toa(v0, buffer);
```
marehr, line 179: Isn't v0 a prefix for all numbers that have more digits than `100000000`? What happens for small numbers like `10`, where `v0` is zero and that prefix should be "" (empty)?

test/unit/std/charconv_int_test.cpp
```diff
@@ -231,14 +233,19 @@ TYPED_TEST(integral_from_char_test, hexadicimal_number)
...
 {
-    TypeParam val{120};
-    std::array<char, 10> buffer{};
+    uint8_t max_num_digits = static_cast<uint8_t>(std::log10(std::numeric_limits<TypeParam>::max())) + 1;
+    std::array<char, 20> buffer{};
 
-    auto res = std::to_chars(buffer.data(), buffer.data() + buffer.size(), val);
+    for (uint8_t num_digits = 1; num_digits <= max_num_digits; ++num_digits)
+    {
+        TypeParam val = seqan3::pow(10, num_digits - 1);
```
marehr, line 241:
```suggestion
        // 1, 10, 100, 1000, 10000, ..., 10^(ceil_log_10(max_number))
        TypeParam val = seqan3::pow(10, num_digits - 1);
```

marehr: I would change it to 1, 12, 123, 1234, 12345, 123456, ..., 123456789,  1234567890, ...

```c++
TypeParam value = 1;
value *= 10 + num_digits % 10;
```

test/unit/std/charconv_int_test.cpp
```diff
@@ -231,14 +233,19 @@ TYPED_TEST(integral_from_char_test, hexadicimal_number)
 
 TYPED_TEST(integral_from_char_test, to_chars)
```
marehr, line 234: Can we have a second test with an explicit small number?

This test is great, because is tests this in a generic way, but sometimes it is also valuable to have explicit instances to test.

Like some edge case: `0`, `42`, `1234`

include/seqan3/contrib/charconv/charconv.cpp
eseiler: This part of the function is only for values in `[100000000, 10000000000000000)`

test/unit/std/charconv_int_test.cpp
eseiler: Well, it's kinda enclosed in the generic test, because it breaks for every value in `[100000000, 10000000000000000)`, so `100000000` is also a regression test.

test/unit/std/charconv_int_test.cpp
eseiler: done

test/unit/std/charconv_int_test.cpp
eseiler: I added it, but there is no edge case for all the types that I could put there ..
the implementation's first branch checks if there are 4 (32 bit) or 8 (64 bit) zeros.
So basically every number I add there - and I'm limited by the 8 bit integer - will result in the same code path

include/seqan3/contrib/charconv/charconv.cpp
eseiler: The if body starting at 130 will be used if the value is smaller than `10000000000000000`

test/unit/std/charconv_int_test.cpp
marehr: Yeah, but a regression test is a test that is exactly like the bug report stated. We want to make sure that that particular instance didn't work, but now works. With this, we can always guarantee that that particular use case is working with our code.

But of course, we should also write generic tests that will cover those regressions.

include/seqan3/contrib/charconv/charconv.cpp
marehr: Ah, I haven't seen that. Thank you.

test/unit/std/charconv_int_test.cpp
marehr: Thank you :)

test/unit/std/charconv_int_test.cpp
```diff
@@ -238,7 +256,7 @@ TYPED_TEST(integral_from_char_test, to_chars)
 
     EXPECT_EQ(res.ptr, &buffer[3]);
     EXPECT_EQ(res.ec, std::errc{});
-    EXPECT_EQ((std::string_view{buffer.data(), 3}), std::string_view{"120"});
+    EXPECT_EQ((std::string_view{buffer.data(), 3}), std::string_view{std::to_string(val)});
```
marehr, line 259:
```suggestion
    EXPECT_EQ((std::string_view{buffer.data(), 3}), std::string_view{"120"});
```

please keep this

test/unit/std/charconv_int_test.cpp
```diff
@@ -251,3 +269,16 @@ TYPED_TEST(integral_from_char_test, to_chars_error)
     EXPECT_EQ(res.ptr, buffer.data() + buffer.size());
     EXPECT_EQ(res.ec, std::errc::value_too_large);
 }
+
+// https://github.com/seqan/seqan3/issues/1595
+TEST(to_chars_test, large_number)
```
marehr, line 274:
```suggestion
TEST(to_chars_test, issue_1595)
```

PR #1980, opened 2020-07-13
[FIX] performance test: missing include for cpp20 gcc10

eseiler:
This adds a missing include.
In cpp20 mode, we do not include the ranges header, but use the STL `<ranges>`, so we need to explicitly include the ranges header.

With this, unit, snippet and performance tests work in cpp20 mode.

### DIFF
--- a/test/performance/range/views/view_all_benchmark.cpp
+++ b/test/performance/range/views/view_all_benchmark.cpp
@@ -7,12 +7,13 @@
 
 #include <deque>
 #include <list>
+#include <seqan3/std/ranges>
 #include <string>
 #include <vector>
 
 #include <benchmark/benchmark.h>
 
-#include <seqan3/std/ranges>
+#include <range/v3/view/all.hpp>
 
 #include <seqan3/range/views/type_reduce.hpp>
 
--- a/test/performance/range/views/view_all_benchmark.cpp
+++ b/test/performance/range/views/view_all_benchmark.cpp
@@ -13,8 +13,6 @@
 
 #include <benchmark/benchmark.h>
 
-#include <range/v3/view/all.hpp>
-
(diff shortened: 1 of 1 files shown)


test/performance/range/views/view_all_benchmark.cpp
```diff
@@ -7,12 +7,13 @@
...
 #include <deque>
 #include <list>
+#include <seqan3/std/ranges>
 #include <string>
 #include <vector>
 
 #include <benchmark/benchmark.h>
 
-#include <seqan3/std/ranges>
+#include <range/v3/view/all.hpp>
```
joergi-w, line 16: Note: This include is already contained in `#include <seqan3/std/ranges>`.

rrahn, line 16: But what is used from the ranges library from all.hpp that is not provided by the standard ranges header?

eseiler: The `::ranges::view::all` that is benchmarked.
My first idea was to excluded the benchmark if there is no range-v3, but Marcel disagreed.

We also still have range-v3 includes in many files....

rrahn: Ok, I am confused now. Should it then not test the standard views::all? Why are we interested in the ranges views::all if we don't use it? I wouldn't disable it. I would just test the views::all that we get. In a life without range-v3 this becomes useless anyway.

eseiler: Yes, this was my other idea, too.
If cpp17 -> useless because they are the same (`std::views::all` = `::ranges::views:all`)
If cpp20 -> useless because why would you use the range-v3 version?

I can change it, I just didn't because Marcel had objections.

PR #1983, opened 2020-07-13
Parallel search/merge algorithm ranges

rrahn:
This PR merges the alignment range and search range returned by the algorithms into the algorithm_range.

resolves seqan/product_backlog#27


include/seqan3/core/algorithm/algorithm_range.hpp
```diff
@@ -18,93 +18,93 @@
...
  *
  * \if DEV
  * Note that the required type is not enforced in order to test this class without adding the entire machinery for
  * the seqan3::detail::algorithm_executor_blocking.
  * \endif
  */
-template <typename alignment_executor_type>
-class alignment_range
+template <typename algorithm_executor_type>
+class algorithm_range
```
smehringer, line 41:
```suggestion
class algorithm_result_range
```
The name `algorithm_range` rather suggests a range over several algorithms, no?

rrahn: Well, that's exactly the point. It is a range over algorithms meaning, that it has to execute the algorithms to get the result. The other name might suggest, that the algorithms have been already executed, and this is only a range over the results.

smehringer: Hmm I see your point, but `*alignment_range.begin()` will return an `alignment_result` or a `search_result` and not an algorithm? That the range is lazy is usually not reflected in the name.

rrahn: It does return a result that is true, but has to execute an algorithm for it. That is not trivial. I suggest to call it then `algorithm_result_generator_range` in order to reflect, that it generates something.

smehringer: Sorry for the late response.

If it is fine with you, I'd prefer `algorithm_result_generator_range` because it is more expressive. And since we usually do not explicitily state this name in snippets I think the rather long name is also fine. Thanks for discussing it :)

PR #1894, opened 2020-06-11
[FEATURE] Improve minimizer computation on repeat regions with large window size

smehringer:
Fixes seqan/product_backlog#59

Before: (benchmark id is abbreviated for better readability)

```
Benchmark                                                                               Time             CPU   Iterations UserCounters...
-----------------------------------------------------------------------------------------------------------------------------------------
poly_A<ungapped>/50000/8/13      694525 ns       694451 ns          991 Throughput[bp/s]=71.9892M/s
poly_A<ungapped>/50000/8/28     1304560 ns      1304527 ns          525 Throughput[bp/s]=38.3227M/s
poly_A<ungapped>/50000/30/35     686750 ns       686747 ns          986 Throughput[bp/s]=72.7648M/s
poly_A<ungapped>/50000/30/50    1304849 ns      1304859 ns          536 Throughput[bp/s]=38.2961M/s
poly_A<gapped>/50000/8/13       1287592 ns      1287535 ns          539 Throughput[bp/s]=38.8285M/s
poly_A<gapped>/50000/8/28       1912495 ns      1912509 ns          363 Throughput[bp/s]=26.14M/s
poly_A<gapped>/50000/30/35      3015136 ns      3015089 ns          234 Throughput[bp/s]=16.5736M/s
poly_A<gapped>/50000/30/50      3724698 ns      3724598 ns          188 Throughput[bp/s]=13.4165M/s
```

after

```
Benchmark                                                                               Time             CPU   Iterations UserCounters...
-----------------------------------------------------------------------------------------------------------------------------------------
poly_A<ungapped>/50000/8/13      501208 ns       501199 ns         1341 Throughput[bp/s]=99.7467M/s
poly_A<ungapped>/50000/8/28      499846 ns       499836 ns         1344 Throughput[bp/s]=100.019M/s
poly_A<ungapped>/50000/30/35     501712 ns       501384 ns         1343 Throughput[bp/s]=99.6662M/s
poly_A<ungapped>/50000/30/50     500758 ns       500737 ns         1321 Throughput[bp/s]=99.795M/s
poly_A<gapped>/50000/8/13       1119021 ns      1118997 ns          618 Throughput[bp/s]=44.6766M/s
poly_A<gapped>/50000/8/28       1129533 ns      1128814 ns          613 Throughput[bp/s]=44.2881M/s
poly_A<gapped>/50000/30/35      2861012 ns      2860958 ns          244 Throughput[bp/s]=17.4665M/s
[...]

PR #1987, opened 2020-07-17
[FEATURE] use reserve strand in minimiser hash view

MitraDarja:
Use reserve strand in minimiser hash view.

### DIFF
--- a/test/unit/range/views/view_minimiser_hash_test.cpp
+++ b/test/unit/range/views/view_minimiser_hash_test.cpp
@@ -39,10 +39,10 @@ template <>
 struct iterator_fixture<iterator_type> : public ::testing::Test
 {
     using iterator_tag = std::forward_iterator_tag;
-    static constexpr bool const_iterable = true;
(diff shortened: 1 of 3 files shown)


MitraDarja:
Note: Once #1894 is merged the result of the gapped example of the snippet will change.

test/unit/range/views/view_minimiser_hash_test.cpp
```diff
@@ -39,10 +39,10 @@ template <>
 struct iterator_fixture<iterator_type> : public ::testing::Test
 {
     using iterator_tag = std::forward_iterator_tag;
-    static constexpr bool const_iterable = true;
+    static constexpr bool const_iterable = false;
```
rrahn, line 42: Why does this change make the minimiser_hash non-const iterable?

eseiler: It is probably const_iterable once again after #1992

Issue #1988, opened 2020-07-17
[BUG] kmer_hash does not preserve const iterability with std::views::reverse


MitraDarja:
As mentioned in #1751 kmer_hash does not preserve const iterability with std::views::reverse and because we test the minimiser view based on kmer_hash the minimiser view does not preserve it either.

```cpp
std::vector<seqan3::dna4> text{"ACGGCGACGTTTAG"_dna4};
auto v = text | seqan3::views::kmer_hash(seqan3::ungapped{4}) | std::views::reverse;
EXPECT_TRUE(seqan3::const_iterable_range<decltype(v)>); // Is false
auto v2 = text | seqan3::views::kmer_hash(seqan3::ungapped{4});
EXPECT_TRUE(seqan3::const_iterable_range<decltype(v2)>); // Is true as expected.
```

MitraDarja:
I tested:
```cpp
EXPECT_TRUE(std::ranges::input_range<decltype(v) const>);
EXPECT_TRUE(std::ranges::forward_range<decltype(v)> == std::ranges::forward_range<decltype(v) const>);
EXPECT_TRUE(std::ranges::bidirectional_range<decltype(v)> == std::ranges::bidirectional_range<decltype(v) const>);
EXPECT_TRUE(std::ranges::random_access_range<decltype(v)> == std::ranges::random_access_range<decltype(v) const>);
```
All of these return false, so every part of the const iterability definiton fails.

eseiler:
Your observation is correct.
We should not drop the `common_range` property.

If you change the `shape_iterator::end()` implementation to this, it works:

```cpp
    auto end() noexcept
    {
        if constexpr (std::ranges::common_range<urng_t>)
            return shape_iterator<urng_t>{std::ranges::end(urange), std::ranges::end(urange), shape_};
        else
            return std::ranges::end(urange);
    }

    //!\copydoc end()
    auto end() const noexcept
    //!\cond
        requires const_iterable_range<urng_t>
    //!\endcond
    {
        if constexpr (std::ranges::common_range<urng_t const>)
            return shape_iterator<urng_t const>{std::ranges::end(urange), std::ranges::end(urange), shape_};
        else
            return std::ranges::end(urange);
    }

    //!\copydoc end()
    auto cend() const noexcept
    //!\cond
        requires const_iterable_range<urng_t>
    //!\endcond
    {
        if constexpr (std::ranges::common_range<urng_t const>)
[...]

PR #1992, opened 2020-07-20
[FIX] Preserve common_range in kmer_hash_view

eseiler:
Resolves https://github.com/seqan/seqan3/issues/1988
Closes https://github.com/seqan/product_backlog/issues/161

### DIFF
--- a/include/seqan3/range/views/kmer_hash.hpp
+++ b/include/seqan3/range/views/kmer_hash.hpp
@@ -127,15 +127,6 @@ public:
         return shape_iterator<urng_t const>{std::ranges::begin(urange), std::ranges::end(urange), shape_};
     }
 
-    //!\copydoc begin()
-    auto cbegin() const noexcept
-    //!\cond
-        requires const_iterable_range<urng_t>
-    //!\endcond
-    {
-        return begin();
-    }
-
     /*!\brief Returns an iterator to the element following the last element of the range.
      * \returns Iterator to the end.
      *
@@ -153,7 +144,11 @@ public:
      */
     auto end() noexcept
     {
-        return std::ranges::end(urange);
+        // Assigning the end iterator to the text_right iterator of the shape_iterator only works for common ranges.
+        if constexpr (std::ranges::common_range<urng_t>)
+            return shape_iterator<urng_t>{std::ranges::begin(urange), std::ranges::end(urange), shape_, true};
+        else
+            return std::ranges::end(urange);
     }
 
     //!\copydoc end()
@@ -162,16 +157,11 @@ public:
         requires const_iterable_range<urng_t>
     //!\endcond
     {
-        return std::ranges::end(urange);
-    }
-
-    //!\copydoc end()
-    auto cend() const noexcept
-    //!\cond
-        requires const_iterable_range<urng_t>
-    //!\endcond
-    {
-        return end();
+        // Assigning the end iterator to the text_right iterator of the shape_iterator only works for common ranges.
+        if constexpr (std::ranges::common_range<urng_t const>)
+            return shape_iterator<urng_t const>{std::ranges::begin(urange), std::ranges::end(urange), shape_, true};
+        else
+            return std::ranges::end(urange);
     }
     //!\}
 
@@ -281,6 +271,7 @@ public:
 
     /*!\brief Construct from a given iterator on the text and a seqan3::shape.
     * /param[in] it_start Iterator pointing to the first position of the text.
+    * /param[in] it_end   Sentinel pointing to the end of the text.
     * /param[in] s_       The seqan3::shape that determines which positions participate in hashing.
     *
     * \details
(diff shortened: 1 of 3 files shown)


include/seqan3/range/views/kmer_hash.hpp
```diff
@@ -153,7 +153,17 @@ class kmer_hash_view : public std::ranges::view_interface<kmer_hash_view<urng_t>
      */
     auto end() noexcept
     {
-        return std::ranges::end(urange);
+        if constexpr (std::ranges::common_range<urng_t>)
```
smehringer, line 156: Hm.. Why don't we always return a shape iterator then? Does the urange sentinel have any advantages?

include/seqan3/range/views/kmer_hash.hpp
```diff
@@ -171,7 +191,17 @@ class kmer_hash_view : public std::ranges::view_interface<kmer_hash_view<urng_t>
...
+        {
+            return shape_iterator<urng_t const>{std::next(std::ranges::begin(urange),
+                                                          std::ranges::distance(urange) - shape_.size() + 1),
+                                                std::ranges::end(urange),
+                                                shape_};
+        }
+        else
+        {
+            return end();
+        }
```
smehringer, line 204: This shouldn't be necessary since `end()` already has the if constexpr?

include/seqan3/range/views/kmer_hash.hpp
```diff
@@ -294,7 +324,7 @@ class kmer_hash_view<urng_t>::shape_iterator
     {
         assert(std::ranges::size(shape_) > 0);
 
-        if (shape_.size() <= std::ranges::distance(text_left, text_right))
+        if (shape_.size() <= std::ranges::distance(text_left, text_right) + 1)
```
smehringer, line 327: Why this change?

include/seqan3/range/views/kmer_hash.hpp
```diff
@@ -511,26 +511,15 @@ class kmer_hash_view<urng_t>::shape_iterator
     }
 
     /*!\anchor shape_iterator_operator-difference
-     * \brief Return offset between this and remote iterator's position.
-     * \attention This function is only available if `it_t` models std::random_access_iterator.
-     */
-    difference_type operator-(shape_iterator const & lhs) const noexcept
```
smehringer, line 487: Didn't you just introduce this operator in the first commit ?

include/seqan3/range/views/kmer_hash.hpp
eseiler: Yes, apparently we can just guarantee common_range; I did something before where it did not work, but now it works.

include/seqan3/range/views/kmer_hash.hpp
eseiler: I can already remote the cbegin and cend in this PR.

include/seqan3/range/views/kmer_hash.hpp
eseiler: Basically:

```
        // shape size = 3
        // Text:     1 2 3 4 5 6 7 8 9
        // it_start: ^
        // it_end:       ^
        // distance(it_start, it_end) = 2
```

But I'm still investigating other stuff.

include/seqan3/range/views/kmer_hash.hpp
eseiler: In the first commit I introduced the friend `operator-`, here I remove the member `operator-`

eseiler:
As it turns out, the iterator semantics were wrong, i.e. the end() was not pointing to the end (or rather, one past the end).
I still need to simplify the code and with forward_list there is still a ASAN error.

include/seqan3/range/views/kmer_hash.hpp
eseiler: No, we can't, because in order to retain the semantics of an end_iterator, we need to be able to assign the end of the underlying range to the `text_right` iterator of the `shape_iterator`. This only works if both have a common type, i.e. the range is a common range.

smehringer: Thanks for explaining!

include/seqan3/range/views/kmer_hash.hpp
```diff
@@ -290,16 +281,66 @@ class kmer_hash_view<urng_t>::shape_iterator
...
+        // text_left: ^
+        // text_right:    ^
+        // distance(text_left, text_right) = 2
+        if (shape_.size() <= std::ranges::distance(text_left, text_right) + 1)
+        {
+            roll_factor = pow(sigma, static_cast<size_t>(std::ranges::size(shape_) - 1));
             hash_full();
         }
+
+        text_right = it_end;
```
smehringer, line 343: Does `hash_full` change the text_right, or why are you doing the assignment here again?

eseiler: Yes, it changes the iterators and text_right would point to the last element of the text, instead to the end

include/seqan3/range/views/kmer_hash.hpp
```diff
@@ -290,16 +281,66 @@ class kmer_hash_view<urng_t>::shape_iterator
...
+    *
+    * Linear in size of shape.
+    */
+    shape_iterator(it_t it_start, sentinel_t it_end, shape s_, bool SEQAN3_DOXYGEN_ONLY(is_end)) : shape_{s_}
+    {
+        assert(std::ranges::size(shape_) > 0);
+
+        auto urange_size = std::ranges::distance(it_start, it_end);
+        auto step = (shape_.size() + 1 > urange_size) ? 0 : urange_size - shape_.size() + 1;
+        text_left = std::ranges::next(it_start, step, it_end);
```
rrahn, line 329: Mhmm, since we do this anyway, would it not be sufficient to use this in the end function and call the old constructor with text_it put to the respective position?

eseiler: If I got it correct, no, because then the right text iterator would point one before the end instead to the end

rrahn: Ok, so if I get it right, then the shape_iterator stores the begin and the end of the current infix over the text to compute the hash. That would mean text_lext points to the first valid character in the shape and text_right to the one behind the last one of the shape. Which is ok, because we ensure that after the last increment (when we reached the end) the text_right points to the regular end of the underlying range (in case it is a common range). In the other case we compare the text_right with the sentinel. So why does it point to one before the end?

eseiler: I have to look at it again, it was quite confusing to make it point to where it should.

PR #1994, opened 2020-07-27
Dockerfile for the Debian rolling release development distribution

mr-c:


marehr:
Thank you for sharing how to reproduce. I like the idea of providing a Dockerfile for easier reproduction, but I'm closing this for now, because we need to think how to incorporate this properly. This might fit into the github-actions sub-folder.

PR #1996, opened 2020-07-29
[TASK] support gcc10.2

### LINKED ISSUE #1993: support gcc 10.2.0
- SeqAn version: 3.0.1
  - Operating system: Debian's rolling development distribution "sid"
  - Compiler: `gcc (Debian 10.2.0-3) 10.2.0`

## Description

seqan 3.0.1 + gcc 10.2.0, tests either fail to compile or ICE with -std=c++17 -fconcepts

### How to repeat the problem

Build the Debian package

### Expected behaviour

No errors

### Actual behaviour

See https://gist.github.com/mr-c/97aecb94becc22aacfdafb89a66038b4

Thanks!

marehr:
Fixes #1993

A change to the standard [1] seems to disallow the following, at least gcc [2] interprets it this way:

```c++
template <typename T> int f(T);
template <typename T> requires true int f(T const&);

int i = f(42);
```

That means constraint functions are only more specialised if they have the same argument "structure".

If you changed the first signature to "int f(T const &);", it would compile again.

This PR unifies the argument "structure"
* for `operator<<(debug_stream, any_type)`,
* for `[bi_]fm_index_cursor::extend_(left|right)`, and
* for `seqan3::search`.

[1] - http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p2113r0.html
[2] - https://gcc.gnu.org/bugzilla/show_bug.cgi?id=96333

### DIFF
--- a/include/seqan3/alignment/matrix/alignment_coordinate.hpp
+++ b/include/seqan3/alignment/matrix/alignment_coordinate.hpp
@@ -396,7 +396,7 @@ public:
  *
  * Prints the alignment coordinate as a tuple.
  */
-template <typename coordinate_type, typename char_t>
+template <typename char_t, typename coordinate_type>
 //!\cond
     requires std::same_as<remove_cvref_t<coordinate_type>, alignment_coordinate> ||
              detail::is_value_specialisation_of_v<remove_cvref_t<coordinate_type>,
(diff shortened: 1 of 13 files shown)


mr-c:
Hey @marehr , thanks for this. Can you provide me with a backport of this patch to https://github.com/seqan/seqan3/releases/tag/3.0.1 ? I tried, but it seems too much has changed for me to figure it out.

marehr:
I think it is easier to just do the next release, because we did quite a lot of changes.

PR #1997, opened 2020-07-31
Remove all.hpp from tests

marehr:
This reduces re-compile time for CI, compile time and interference of unrelated files.

~~PLEASE SQUASH MERGE AT THE END~~ **I already did this**


rrahn:
@marehr unfortunately, there is a conflict, that needs to be resolved first. After that I can squash merge

marehr:
I'm merging, jenkins couldn't upload, but from what I can see it, all tests passed on jenkins.

PR #1999, opened 2020-07-31
avoid including complete debug_stream in tests

marehr:
Part of https://github.com/seqan/product_backlog/issues/167


test/include/seqan3/test/pretty_printing.hpp
```diff
@@ -14,6 +14,8 @@
 
 #include <ios>
 
+#include <seqan3/std/iterator>
```
eseiler, line 17:
```suggestion
#include <ios>
#include <seqan3/std/iterator>
```

PR #2003, opened 2020-08-01
[MISC] make inherited_iterator_base::base public (formerly named this_to_base)

marehr:
This is part of seqan/product_backlog#62

### DIFF
--- a/include/seqan3/range/detail/inherited_iterator_base.hpp
+++ b/include/seqan3/range/detail/inherited_iterator_base.hpp
@@ -107,22 +107,19 @@ public:
     {}
     //!\}
 
-    //!\brief Cast this to base type.
-    constexpr base_t & base()
+    //!\brief Get a copy of the base.
+    constexpr base_t base() const &
+    //!\cond
+        requires std::copyable<base_t>
+    //!\endcond
     {
-        if constexpr (wrap_base)
-            return member;
-        else
-            return *this;
+        return as_base();
     }
 
     //!\copydoc base
-    constexpr base_t const & base() const
+    constexpr base_t base() &&
     {
-        if constexpr (wrap_base)
-            return member;
-        else
-            return *this;
+        return std::move(as_base());
     }
 
     /*!\name Comparison operators
@@ -204,7 +201,7 @@ public:
         requires requires (base_t_ i) { ++i; }
     //!\endcond
     {
-        ++base();
+        ++as_base();
         return *this_derived();
     }
 
@@ -217,7 +214,7 @@ public:
         requires requires (base_t_ i) { i++; requires !std::same_as<decltype(i++), base_t_>; }
     //!\endcond
     {
-        return base()++;
+        return as_base()++;
     }
 
     //!\brief Post-increment, return previous iterator state.
@@ -231,7 +228,7 @@ public:
                  std::constructible_from<derived_t, base_t_>
     //!\endcond
     {
-        return derived_t{base()++};
+        return derived_t{as_base()++};
     }
 
     //!\brief Pre-decrement, return updated iterator.
@@ -243,7 +240,7 @@ public:
         requires requires (base_t_ i) { --i; }
     //!\endcond
     {
-        --base();
+        --as_base();
         return *this_derived();
     }
 
@@ -257,7 +254,7 @@ public:
         requires requires (base_t_ i) { i--; } && std::constructible_from<derived_t, base_t_>
     //!\endcond
     {
-        return derived_t{base()--};
+        return derived_t{as_base()--};
     }
 
     //!\brief Move iterator to the right.
@@ -269,7 +266,7 @@ public:
         requires requires (base_t_ i, difference_type const n) { i += n; }
     //!\endcond
     {
-        base() += skip;
+        as_base() += skip;
         return *this_derived();
     }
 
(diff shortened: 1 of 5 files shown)


include/seqan3/range/detail/inherited_iterator_base.hpp
```diff
@@ -107,6 +107,24 @@ class inherited_iterator_base : public std::conditional_t<std::is_pointer_v<base
...
+    constexpr base_t & base()
+    {
+        if constexpr (wrap_base)
+            return member;
+        else
+            return *this;
+    }
+
+    //!\copydoc base
+    constexpr base_t const & base() const
```
rrahn, line 120:
```suggestion
    constexpr base_t base() const noexcept(std::is_nothrow_copyconstructible_v<base_t>)
```

include/seqan3/range/detail/inherited_iterator_base.hpp
```diff
@@ -107,6 +107,24 @@ class inherited_iterator_base : public std::conditional_t<std::is_pointer_v<base
     {}
     //!\}
 
+    //!\brief Cast this to base type.
+    constexpr base_t & base()
```
rrahn, line 111:
```suggestion
    constexpr base_t & base() noexcept
```

include/seqan3/range/detail/inherited_iterator_base.hpp
marehr: Hmm, can you explain me why this would be better? I think I want to have the `const &` back, to avoid a copy, because I assume that this iterator is a bit bigger. Also, I kinda want to keep the original data as it is. I know we don't use any virtual here, so any potential data slicing isn't an issue here, but it somehow feels wrong to copy here.

The standard uses this similarly, but in the other direction:
https://eel.is/c++draft/range.utility#view.interface-1

```c++
    constexpr D& derived() noexcept {                   // exposition only
      return static_cast<D&>(*this);
    }
    constexpr const D& derived() const noexcept {       // exposition only
      return static_cast<const D&>(*this);
    }

```

rrahn: If I get it right, than base_t is the actual iterator that is wrapped. So after you call base() you only have base and nothing more, whether this is the part inherited from or the stored member. Since iterators are always supposed to be cheap to copy you would always copy them. As far as I know, this is what the base() function of all the STD iterators do. In fact they differentiate between `base () const &` and `base() &&`. 
So it seems you want to simply replace this `to_base` function with a public interface but then mix the semantics of these functions. I'd think you should then offer both. One internal interface which simply does the "static_cast" and a proper `base()` interface that would do a conversion. Because your example is only done internally and not exposed to the public interface. I would like to keep that here as well, as `++my_it.base()` can do a lot of unexpected things if it were actually a reference, since you are actually modifying the state of `my_it` without using the ++operator of `my_it`.

marehr: >  In fact they differentiate between `base () const &` and `base() &&`.

This makes totally sense, because an iterator might be move-only (some input iterators) and we shouldn't copy in this cases.

> One internal interface which simply does the "static_cast" and a proper `base()` interface that would do a conversion.

I have to look at it again, but I think I need that state changing property within the inherited iterator. So I guess we need an `as_base` protected function that can be accessed by the derived class and one which is public.

include/seqan3/range/detail/inherited_iterator_base.hpp
```diff
@@ -107,6 +107,21 @@ class inherited_iterator_base : public std::conditional_t<std::is_pointer_v<base
     {}
     //!\}
 
+    //!\brief Get a copy of the base.
```
marehr, line 110: @rrahn Maybe you have a better description in mind, because it does not completely fit for the `&&` overload

include/seqan3/range/detail/inherited_iterator_base.hpp
```diff
@@ -107,22 +107,19 @@ class inherited_iterator_base : public std::conditional_t<std::is_pointer_v<base
     {}
     //!\}
 
-    //!\brief Cast this to base type.
-    constexpr base_t & base()
+    //!\brief Get a copy of the base.
```
rrahn, line 110: Maybe?
```suggestion
    //!\brief Returns an [rvalue](https://en.cppreference.com/w/cpp/language/value_category) of the base.
```

include/seqan3/range/detail/inherited_iterator_base.hpp
```diff
@@ -107,6 +107,21 @@ class inherited_iterator_base : public std::conditional_t<std::is_pointer_v<base
...
+    constexpr base_t base() const &
+    //!\cond
+        requires std::copyable<base_t>
+    //!\endcond
+    {
+        return as_base();
+    }
+
+    //!\copydoc base
```
marehr, line 119:
```suggestion
    //!\brief Returns an [rvalue](https://en.cppreference.com/w/cpp/language/value_category) of the base.
```

include/seqan3/range/detail/inherited_iterator_base.hpp
marehr: ty <3

PR #2005, opened 2020-08-02
No base class construction pipeable_config_element

wvandertoorn:
Resolves https://github.com/seqan/product_backlog/issues/118

The only class which is allowed to construct `pipeable_config_element<...>{}` is the derived class of `pipeable_config_element<...>{}`. `pipeable_config_element<...>{}` befriends the derived class (this makes private members accessible for the derived class).

### DIFF
--- a/include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
+++ b/include/seqan3/alignment/configuration/align_config_aligned_ends.hpp
@@ -547,6 +547,20 @@ template <typename end_gaps_t>
 //!\endcond
 struct aligned_ends : public pipeable_config_element<aligned_ends<end_gaps_t>, end_gaps_t>
 {
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    aligned_ends() = default; //!< Defaulted.
+    aligned_ends(aligned_ends const &) = default; //!< Defaulted.
+    aligned_ends(aligned_ends &&) = default; //!< Defaulted.
+    aligned_ends & operator=(aligned_ends const &) = default; //!< Defaulted.
+    aligned_ends & operator=(aligned_ends &&) = default; //!< Defaulted.
+    ~aligned_ends() = default; //!< Defaulted.
+
+    //!\brief Construct from base type.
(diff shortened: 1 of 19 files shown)


Irallia:
You got an error with travis. I think your constructors should be public.

```bash
Scanning dependencies of target align_config_scoring_test

[  5%] Building CXX object alignment/configuration/CMakeFiles/align_config_scoring_test.dir/align_config_scoring_test.cpp.o

In file included from /home/travis/build/seqan/seqan3/test/unit/alignment/configuration/align_config_vectorise_test.cpp:13:

/home/travis/build/seqan/seqan3/include/seqan3/alignment/configuration/align_config_vectorise.hpp:55:50: error: ‘constexpr seqan3::pipeable_config_element<derived_t, void>::pipeable_config_element() [with derived_t = seqan3::detail::vectorise_tag]’ is private within this context

   55 | inline constexpr detail::vectorise_tag vectorise{};

      |                                                  ^
```

eseiler:
The constructors should be private; but the derived classes inherit the visibility of the base class constructor.
I.e, all the constructors of the config elements are now private.

I think you also need to add ctors to the derived classes - which are then again public and can internally (per =default) use the private ctor of the base class. Although I'm not absolutely sure, the ticket is rather sparse.

If you add `vectorise_tag() = default;` for the vectorise config element, it works again.
Also, gcc7 and 8 work because they have some defects regarding visibility...
gcc9 and 10 get it correct.

wvandertoorn:
gcc7 wasn't agreeing with the `constexpr`s in search/configuration/max_error.hpp. Without those, it should now work for all compilers.

eseiler:
> gcc7 wasn't agreeing with the `constexpr`s in search/configuration/max_error.hpp. Without those, it should now work for all compilers.

In general, if you use defaulted constructors, they will be `constexpr` and `noexcept` if they can - so we do not add these two keywords to `=default` (or `=delete`) constructors.

wvandertoorn:
@eseiler Ah, I see. I wasn't aware of that. I've force-pushed to update the PR (remove the constexpr'ness of the default constructors) and cleaned up the commit history.

PR #2007, opened 2020-08-05
Optimise alignment/part4 combined score and trace matrix

rrahn:
This adds a combined score and trace matrix which will be used later for the trace computation. The job of this class is to provide a unified interface to the alignment matrix wither with only the score or with the score and the trace information. In addition, the design of the class allows to combine different score matrix and trace matrix implementations, which is something we want for the future. For example to provide a sampled trace matrix that only stores specific columns, or a score matrix implementing storing the entire matrix etc.

part of #1908


include/seqan3/alignment/matrix/detail/affine_cell_proxy.hpp
```diff
@@ -123,11 +182,55 @@ class affine_cell_proxy : public tuple_t
...
+    }
+
+    //!\brief Access the horizontal score of the wrapped score matrix cell.
+    decltype(auto) horizontal_trace() & { return get_trace_impl<1>(*this); }
+    //!\overload
+    decltype(auto) horizontal_trace() const & { return get_trace_impl<1>(*this); }
+    //!\overload
+    decltype(auto) horizontal_trace() && { return get_trace_impl<1>(std::move(*this)); }
+    //!\overload
+     decltype(auto) horizontal_trace() const &&
```
simonsasse, line 214: There seems to be an unnecessary space here :)
```suggestion
    decltype(auto) horizontal_trace() const &&
```

include/seqan3/alignment/matrix/detail/combined_score_and_trace_matrix.hpp
```diff
@@ -0,0 +1,337 @@
...
+ */
+template <std::ranges::input_range score_matrix_t, std::ranges::input_range trace_matrix_t>
+//!\cond
+    requires (std::ranges::input_range<std::ranges::range_reference_t<score_matrix_t>> &&
+              std::ranges::input_range<std::ranges::range_reference_t<trace_matrix_t>>)
+//!\endcond
+class combined_score_and_trace_matrix
+{
+private:
+    //!\brief The score type extracted from the score matrix. (TODO: make this less complicate by exposing the score type).
```
simonsasse, line 52:
```suggestion
    //!\brief The score type extracted from the score matrix. (TODO: make this less complicated by exposing the score type).
```

include/seqan3/alignment/matrix/detail/combined_score_and_trace_matrix.hpp
```diff
@@ -0,0 +1,337 @@
...
+    {
+        return views::zip(*score_matrix_it, *trace_matrix_it) | transform_to_combined_matrix_cell;
+    }
+    //!\}
+
+    /*!\name Arithmetic operators
+     * \{
+     */
+    //!\brief Advance the iterator to the next alignment matrix column.
+    iterator & operator++()
```
simonsasse, line 241: Just a question: Whats the difference between these two ++ operator definitions and why does the first as a return value, but no input and the second one an input but no return value ? :)

rrahn: There are two increment operators: pre-increment (`++it`) and post-increment (`it++`). The latter is implemented as `iterator operator++(int)`. Now the semantic difference between those two is the following:

```cpp
auto it = rng.begin();
auto it2 = ++it; // it2 == it and both point to the element at position 1 of the rng.
auto it3 = it++; // Now it3 == it2 != it, i.e. it++ returns a copy of the iterator before the increment of it.
```
So in the last line it3 and it2 both point to the second element and it to the third element.
Since c++20 an input-iterator is not required to return a copy of the iterator pointing to the previous element, respectively a copy of the previously pointed to element. any more. This is great, since for input iterators making a copy can be difficult, because the advancing one copy will invalidate the other copy. That's why the post-increment operator is here defined as `void`.

simonsasse: Thanks for this explanation ! :)

include/seqan3/alignment/matrix/detail/combined_score_and_trace_matrix.hpp
```diff
@@ -0,0 +1,354 @@
...
+
+#include <seqan3/alignment/matrix/detail/affine_cell_proxy.hpp>
+#include <seqan3/alignment/matrix/detail/matrix_coordinate.hpp>
+#include <seqan3/core/concept/core_language.hpp>
+#include <seqan3/core/type_traits/range.hpp>
+#include <seqan3/range/views/zip.hpp>
+
+namespace seqan3::detail
+{
+/*!\brief An alignment matrix that combines a score matrix and a trace matrix to a common interface.
```
eseiler, line 25:
```suggestion
/*!\brief An alignment matrix that combines a score matrix with a trace matrix into a common interface.
```

include/seqan3/alignment/matrix/detail/combined_score_and_trace_matrix.hpp
```diff
@@ -0,0 +1,354 @@
...
+ * \implements std::ranges::input_range
+ *
+ * \tparam score_matrix_t The type of the score matrix; must model std::ranges::input_range (for more requirements see
+ *                        details section).
+ * \tparam trace_matrix_t The type of the trace matrix; must model std::ranges::input_range (for more requirements see
+ *                        details section).
+ *
+ * \details
+ *
+ * To compute the actual alignment an additional trace matrix is required, which stores the trace directions
```
eseiler, line 36:
```suggestion
 * To compute the actual, alignment an additional trace matrix is required. It stores the trace directions
```

include/seqan3/alignment/matrix/detail/combined_score_and_trace_matrix.hpp
```diff
@@ -0,0 +1,354 @@
...
+ * to the previous cells for any computed cell within the alignment matrix. This matrix combines a given score and
+ * trace matrix to a complete alignment matrix. The iterator increments over the columns of each underlying matrix and
+ * combines them into a zip view, i.e. the combined alignment matrix is a range over ranges with the inner range also
+ * modelling std::ranges::input_range.
+ * To provide a uniform interface to the alignment algorithm, the returned zip view is further transformed, such that
+ * its reference type is a seqan3::detail::affine_cell_proxy.
```
eseiler, line 42:
```suggestion
 * to the previous cells for any computed cell within the alignment matrix. The following matrix combines a given score 
 * with a trace matrix into a complete alignment matrix. The iterator increments over the columns of each underlying 
 * matrix and combines them into a zip view, i.e. the combined alignment matrix is a range over ranges with the inner 
 * range also modelling std::ranges::input_range.
 * To provide a uniform interface to the alignment algorithm, the returned zip view is further transformed such that
 * its reference type is a seqan3::detail::affine_cell_proxy.
```

include/seqan3/alignment/matrix/detail/combined_score_and_trace_matrix.hpp
```diff
@@ -0,0 +1,354 @@
...
+ */
+template <std::ranges::input_range score_matrix_t, std::ranges::input_range trace_matrix_t>
+//!\cond
+    requires (std::ranges::input_range<std::ranges::range_reference_t<score_matrix_t>> &&
+              std::ranges::input_range<std::ranges::range_reference_t<trace_matrix_t>>)
+//!\endcond
+class combined_score_and_trace_matrix
+{
+private:
+    //!\brief The score type extracted from the score matrix. (TODO: make this less complicated by exposing the score type).
```
eseiler, line 52: Line length.
Did you want to do this in this PR or did you want to create a backlog item?

include/seqan3/alignment/matrix/detail/combined_score_and_trace_matrix.hpp
```diff
@@ -0,0 +1,354 @@
...
+/*!\brief Combined score and trace matrix iterator for the pairwise sequence alignment.
+ * \implements std::input_iterator
+ *
+ * \details
+ *
+ * Implements a zipped iterator over the given score and trace matrix. It assumes that both matrices have the
+ * same number of columns. If this invariant does not hold using this iterator causes undefined behaviour.
+ * When dereferencing the iterator the current score matrix column and trace matrix column will be zipped using the
+ * seqan3::views::zip view and additionally transformed such that the reference type of the returned alignment column
+ * is a seqan3::detail::affine_cell_proxy holding the score and trace values of the respective alignment matrix cells.
```
eseiler, line 155:
```suggestion
 * Implements a zipped iterator over the given score and trace matrix. It assumes that both matrices have the
 * same number of columns. If this invariant does not hold, using this iterator is undefined behaviour.
 * When dereferencing the iterator, the current score matrix column and trace matrix column will be zipped using the
 * seqan3::views::zip view and additionally transformed such that the reference type of the returned alignment column
 * is a seqan3::detail::affine_cell_proxy holding the score and trace values of the respective alignment matrix cells.
```

include/seqan3/alignment/matrix/detail/combined_score_and_trace_matrix.hpp
```diff
@@ -0,0 +1,354 @@
...
+
+    /*!\name Constructor, assignment and destructor
+     * \{
+     */
+    iterator() noexcept = default; //!< Defaulted.
+    iterator(iterator const &) noexcept = default; //!< Defaulted.
+    iterator(iterator &&) noexcept = default; //!< Defaulted.
+    iterator & operator=(iterator const &) noexcept = default; //!< Defaulted.
+    iterator & operator=(iterator &&) noexcept = default; //!< Defaulted.
+    ~iterator() = default; //!< Defaulted.
```
eseiler, line 226: Won't defaulted ctors be noexcept if they can? So you want to make sure that they are noexcept here?

include/seqan3/alignment/matrix/detail/combined_score_and_trace_matrix.hpp
rrahn: Yes, I want to do it later, since it means changing some other files as well, that would not fit into this PR.

include/seqan3/alignment/matrix/detail/combined_score_and_trace_matrix.hpp
rrahn: It should be not enforced. Good point. Copy'n'paste error.

eseiler: (same for sentinel)

rrahn:
sorry @simonsasse I was to impatient and tried to request @eseiler two times because the browser was not refreshing, but then accidentally requested you again.

Issue #2011, opened 2020-08-06
With Doxygen v1.8.20 + seqan3 3.0.2, the documentation build fails due to warnings


mr-c:
Testing https://github.com/seqan/seqan3/commit/bf04354bab903dc7f8c12f3bb478c02cd5f2c553 with `CXXFLAGS='-O3 -std=c++17 -fconcepts'` (using the included range-v3, not the Debian package)

A workaround to skip these warning sensistive tests (or a fix to remove the warnings) would be very welcome.

## Platform

  - SeqAn version: almost v3.0.2
  - Operating system: Linux mrc-tux 5.7.0-1-amd64 #1 SMP Debian 5.7.6-1 (2020-06-24) x86_64 GNU/Linux
  - Compiler: gcc (Debian 10.2.0-3) 10.2.0

## Description
When using Doxygen v1.8.18 the documentation build fails due to warnings

### How to repeat the problem

1. download https://github.com/mr-c/seqan3/blob/docker_build_debian_sid/test/Dockerfile.Debian
2. change to a local checkout of the seqan3 code
3. from the root of the seqan3 repository run `docker build -f .../path/to/Dockerfile.Debian .`

### Expected behaviour
All tests pass

### Actual behaviour
```
The following tests FAILED:
	  1 - doc_usr_no_warnings_test (Failed)
	  3 - doc_dev_no_warnings_test (Failed)


1/4 Test #1: doc_usr_no_warnings_test ..........***Failed   12.69 sec
/build/seqan3-3.0.2~rc1+bf04354+ds/doc/tutorial/pairwise_alignment/index.md:24: warning: image file name align_transcript.png is ambiguous.
Possible candidates:
   /build/seqan3-3.0.2~rc1+bf04354+ds/build_documentation/doc_dev/html/align_transcript.png
   /build/seqan3-3.0.2~rc1+bf04354+ds/build_documentation/doc_usr/html/align_transcript.png
   /build/seqan3-3.0.2~rc1+bf04354+ds/doc/tutorial/pairwise_alignment/align_transcript.png

/build/seqan3-3.0.2~rc1+bf04354+ds/doc/tutorial/ranges/views/minimiser/index.md:34: warning: image file name minimiser.png is ambiguous.
Possible candidates:
   /build/seqan3-3.0.2~rc1+bf04354+ds/build_documentation/doc_dev/html/minimiser.png
[...]

PR #2015, opened 2020-08-09
[MISC] Use doxygen 1.8.19

eseiler:
The prebuilt doxygen we use now needs libclang and glib>=2.29 (i.e. ubuntu 20.04).
Technically, focal is only provided as preview for github actions...

Fixes #2015

### DIFF
--- a/.github/workflows/documentation.yaml
+++ b/.github/workflows/documentation.yaml
@@ -8,7 +8,7 @@ on:
 jobs:
   deploy_documentation:
     name: Deploy Documentation
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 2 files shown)


smehringer:
> Fixes #2015

You referenced this PR in your description? Did you mean https://github.com/seqan/seqan3/issues/2011 ?

smehringer:
Hm I wonder if this should go into release 3.0.2 instead of master?

eseiler:
> Hm I wonder if this should go into release 3.0.2 instead of master?

We only deploy docs via github workflows on the master branch...

marehr:
I'll merge, we can always "backport" this commit by cherry-picking it.

PR #2012, opened 2020-08-06
Search config/on result

rrahn:
This PR introduces the on_result configuration element for the search algorithm.

PR #2019, opened 2020-08-10
Search confguration/on result release

rrahn:
Supersedes #2012 
fixes seqan/product_backlog#169

### DIFF
--- a/include/seqan3/core/semiregular_box.hpp
+++ b/include/seqan3/core/semiregular_box.hpp
@@ -1,0 +1,37 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2019, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2019, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides seqan3::semiregular_box.
+ * \author Rene Rahn <rene.rahn AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <range/v3/utility/semiregular_box.hpp>
+
+#include <seqan3/core/platform.hpp>
+
+namespace seqan3
+{
+/*!\brief Utility wrapper that behaves like std::optional but makes the type conform with the std::semiregular concept.
+          Imported from ranges::semiregular_box.
+ * \ingroup core
+ *
+ * \see https://en.cppreference.com/w/cpp/ranges/semiregular_wrapper
+ */
+using SEQAN3_DOXYGEN_ONLY(semiregular_box =) ::ranges::semiregular_box;
+
+/*!\brief Utility transformation trait to get a wrapper type that models std::semiregular. Imported from
+          ranges::semiregular_box_t.
+ * \ingroup core
+ *
+ * \see https://en.cppreference.com/w/cpp/ranges/semiregular_wrapper
+ */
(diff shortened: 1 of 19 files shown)


include/seqan3/core/semiregular_box.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+#include <seqan3/core/platform.hpp>
+
+namespace seqan3
+{
+// Use semiregular box from ranges library.
+using ::ranges::semiregular_box;
+
+// Use semiregular box transformation trait from ranges library.
+using ::ranges::semiregular_box_t;
```
eseiler, line 25: doxygen may complain, see https://github.com/seqan/seqan3/blob/master/include/seqan3/core/common_tuple.hpp for example

rrahn:
> (you can change the branch to merge against via the edit button near the PR name)

I know, but I accidently rebased some stuff from master locally and so I decided to make a clean branch from the release branch.

include/seqan3/core/semiregular_box.hpp
rrahn: I could run the documentation without issues. Let's see what the CI says.

rrahn: Ok, I saw, that we also referenced it in the documentation so I decided to make give it a proper documentation entry.

include/seqan3/search/detail/search_traits.hpp
```diff
@@ -92,6 +93,9 @@ struct search_traits
                                                      output_reference_id |
                                                      output_reference_begin_position |
                                                      output_index_cursor;
+
+    //!\brief A flag indicating whether a user provided callback was given.
+    static constexpr bool is_one_way_execution = search_configuration_t::template exists<search_cfg::on_result>();
```
smehringer, line 98: I am confused with the naming. "Two-way execution" I first got to know when you introduced it in the context of executors/parallelization (or do I remember that wrongly?). Now you use one-way execution for a user defined callback? Can we maybe rather use something like `has_user_defined_callback`? 

EDIT: I see below that you use it to differentiate between executors. Can you change the brief then? `If the user provides a callback, we need to use a one_way_execution...`

include/seqan3/search/configuration/all.hpp
```diff
@@ -163,4 +163,20 @@
  * The seqan3::search_cfg::parallel configuration element can be combined with any other search configuration.
  *
  * \include test/snippet/search/configuration_parallel.cpp
+ *
+ * ### User callback
+ *
+ * In the default case, a call to seqan3::search returns a lazy range over the results of the search. This lazy range
+ * has the advantage that the results are always in a deterministic order even if the search is executed in parallel.
+ * Sometimes, however, it might be challenging to provide a user defined callback.
```
smehringer, line 171:
```suggestion
 * Sometimes, however, it might be preferred to provide a user defined callback.
```
? Why should it be challenging

rrahn: Ask @eseiler 😏  I had demanding.

eseiler: demanding in this context means `difficult`; but obviously you meant preferred, I just didn't catch your meaning :)

include/seqan3/search/detail/search_traits.hpp
rrahn: This is the point. In the executor context two-way means that it returns something to the user he can wait for and one-way means it does not return anything but just executes it and the user needs to make sure the code is handled properly. But I see that the term might be confusing and I will rename it.

include/seqan3/search/configuration/all.hpp
rrahn: Ok, what I actually meant was desirable

smehringer: :smile: desirable it is

PR #2023, opened 2020-08-11
Simplify seqan3/std/concepts

marehr:
Fixes https://github.com/seqan/product_backlog/issues/190

### DIFF
--- a/include/seqan3/std/concepts
+++ b/include/seqan3/std/concepts
@@ -24,6 +24,7 @@
 
 #include <concepts/concepts.hpp>
 #include <range/v3/functional/concepts.hpp>
+#include <range/v3/utility/common_type.hpp>
 
 #include <seqan3/core/type_traits/basic.hpp>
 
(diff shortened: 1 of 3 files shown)


include/seqan3/std/concepts
```diff
@@ -64,81 +64,100 @@ namespace std
...
 using ::concepts::unsigned_integral;
-
-/*!\interface std::floating_point <>
- * \brief The concept std::floating_point<T> is satisfied if and only if T is a floating-point type.
- * \sa https://en.cppreference.com/w/cpp/types/is_floating_point
- */
+// using ::concepts::floating_point;
 //!\cond
 template <typename t>
 SEQAN3_CONCEPT floating_point = std::is_floating_point_v<t>;
```
smehringer: why did you remove the documentation fthe floating_point concept although we still need to implement it?

marehr: Because std::floating_point should use the external dokumentation of cppreference.

PR #2031, opened 2020-08-13
[Alignment Configuration] Rename align_cfg::alignment_result_capture to align_cfg::detail::result_type

simonsasse:
Resolves seqan/product_backlog#183

This moves the alignment_result_capture to a new namespace seqan3::align_cfg::detail and then renames it to result_type.
I needed to specify the whole namespace path seqan3::detail::XX for most of the view files. This is due to the newly created seqan3::align_cfg::detail which had a name clash with seqan3::detail.

Merging this PR closes #2018 because it has the same intention.

Blocked by #2033

PR #2020, opened 2020-08-10
[FEATURE] Alignment Configuration: put align_cfg::debug to align_cfg::detail::debug

Irallia:
Resolves https://github.com/seqan/product_backlog/issues/181

Blocked by #2033

PR #2033, opened 2020-08-14
[FIX] add namespace seqan3::align_cfg::detail

marehr:
Blocks #2031 and #2020

### DIFF
--- a/include/seqan3/alignment/configuration/all.hpp
+++ b/include/seqan3/alignment/configuration/all.hpp
@@ -31,6 +31,12 @@
  * \brief A special sub namespace for the alignment configurations.
  */
 
+ /*!\if DEV
+  * \namespace seqan3::align_cfg::detail
+  * \copydoc seqan3::detail
+  * \endif
+  */
+
 /*!\defgroup alignment_configuration Configuration
  * \ingroup alignment
  * \brief Provides configuration elements for the pairwise alignment configuration.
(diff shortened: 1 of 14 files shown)


include/seqan3/alignment/configuration/all.hpp
```diff
@@ -31,6 +31,12 @@
  * \brief A special sub namespace for the alignment configurations.
  */
 
+ /*!\cond DEV
+  * \namespace seqan3::align_cfg::detail
+  * \copydoc seqan3::detail
+  * \endcond
+  */
```
eseiler, line 38:
```suggestion
 /*!\if DEV
  * \namespace seqan3::align_cfg::detail
  * \copydoc seqan3::detail
  * \endif
  */
```

We should use `if` for inside comment blocks and `cond` for exluding code / multiple comment blocks

marehr: Is this the same?

I just copied the code from

https://github.com/seqan/seqan3/blob/aa45291d7aafd88d2d58c2e4fcf1702b1fbf6220/include/seqan3/core/all.hpp#L48-L56

eseiler: if would be better style; but then you can just leave it as it is.

marehr: Changed it

PR #2038, opened 2020-08-14
[FIX] use std::ranges::borrowed_range instead of seqan3::forwarding_range

marehr:
Part of https://github.com/seqan/product_backlog/issues/49

### DIFF
--- a/include/seqan3/range/concept.hpp
+++ b/include/seqan3/range/concept.hpp
@@ -48,21 +48,6 @@ SEQAN3_CONCEPT const_iterable_range =
     (std::ranges::random_access_range<std::remove_const_t<type>> == std::ranges::random_access_range<type const>);
 //!\endcond
 
-/*!\interface seqan3::forwarding_range<>
- * \ingroup range
- * \extends std::ranges::range
- * \brief Specifies a range whose iterators may outlive the range and remain valid.
- * \see https://eel.is/c++draft/range.req
- */
-//!\cond
-template <typename type>
-SEQAN3_CONCEPT forwarding_range = std::ranges::range<type> && requires (type && val)
-{
-    std::ranges::begin(std::forward<type>(val));
-    std::ranges::end(std::forward<type>(val));
-};
-//!\endcond
-
 /*!\interface seqan3::pseudo_random_access_iterator <>
  * \ingroup range
  * \extends   std::forward_iterator
--- a/include/seqan3/std/ranges
+++ b/include/seqan3/std/ranges
@@ -30,6 +30,22 @@ template<class T>
     requires ::std::derived_from<T, ::std::ranges::view_base>
 //!\endcond
 inline constexpr bool enable_view<T> = true;
+
+// std::ranges::borrowed_range's are valid range-v3 borrowed_range's
+//!\cond
+template <::std::input_or_output_iterator I, ::std::sentinel_for<I> S, ::std::ranges::subrange_kind K>
+inline constexpr bool enable_borrowed_range<::std::ranges::subrange<I, S, K>> = true;
+
+template <class T>
+inline constexpr bool enable_borrowed_range<::std::ranges::empty_view<T>> = true;
+
+template <::std::weakly_incrementable W, ::std::semiregular Bound>
+inline constexpr bool enable_borrowed_range<::std::ranges::iota_view<W, Bound>> = true;
+
+template <class T>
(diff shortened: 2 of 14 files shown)


include/seqan3/range/concept.hpp
```diff
@@ -48,11 +48,14 @@ SEQAN3_CONCEPT const_iterable_range =
     (std::ranges::random_access_range<std::remove_const_t<type>> == std::ranges::random_access_range<type const>);
 //!\endcond
 
+inline namespace deprecated
```
smehringer, line 51: Why do we use a an deprecated inline namespace instead of the deprecation macro?

marehr: you can't deprecate a concept, they don't accept an attribute

smehringer: meh

marehr:
@smehringer I added another fix; maybe you can have a more thoroughly look at it again. :)

include/seqan3/range/concept.hpp
marehr: I removed the concept completely and added a CHANGELOG entry.

include/seqan3/std/ranges
```diff
@@ -30,6 +30,22 @@ template<class T>
     requires ::std::derived_from<T, ::std::ranges::view_base>
 //!\endcond
 inline constexpr bool enable_view<T> = true;
+
+// std::ranges::borrowed_range's are valid range-v3 borrowed_range's
+//!\cond
+template <::std::input_or_output_iterator I, ::std::sentinel_for<I> S, ::std::ranges::subrange_kind K>
```
smehringer, line 36: doesn't the `::std::sentinel_for<I>` trigger the visibility bug?

include/seqan3/std/ranges
```diff
@@ -40,6 +56,13 @@ template<class T>
     requires ::std::derived_from<T, ::ranges::view_base>
 //!\endcond
 inline constexpr bool enable_view<T> = true;
+
+//!\brief std::ranges::borrowed_range's are valid range-v3 borrowed_range's
+template<class T>
+//!\cond
+    requires ::ranges::enable_borrowed_range<T>
```
smehringer, line 63: So every `::ranges::enable_borrowed_range<T>` is a `std::ranges::borrowed_range`

But above only enable the `::ranges::enable_borrowed_range` for `std::ranges::iota, std::ranges::subrange, std::ranges::ref_view` and `std::ranges::empty_view`?

include/seqan3/std/ranges
marehr: no, because this path will only be used with gcc>=10.

include/seqan3/std/ranges
marehr: yeah, because of recursion. `std::ranges::borrowed_range` -> `::ranges::enable_borrowed_range<T>` -> `std::ranges::enable_borrowed_range<T>` -> `::ranges::enable_borrowed_range<T>` -> `std::ranges::enable_borrowed_range<T>` -> `::ranges::enable_borrowed_range<T>` -> `std::ranges::enable_borrowed_range<T>` -> `::ranges::enable_borrowed_range<T>` -> `std::ranges::enable_borrowed_range<T>` -> `::ranges::enable_borrowed_range<T>` -> `std::ranges::enable_borrowed_range<T>` -> `::ranges::enable_borrowed_range<T>` -> `std::ranges::enable_borrowed_range<T>` -> `::ranges::enable_borrowed_range<T>` -> `std::ranges::enable_borrowed_range<T>` -> `::ranges::enable_borrowed_range<T>` -> `std::ranges::enable_borrowed_range<T>` -> `::ranges::enable_borrowed_range<T>` -> `std::ranges::enable_borrowed_range<T>` -> `::ranges::enable_borrowed_range<T>` -> `std::ranges::enable_borrowed_range<T>` -> `::ranges::enable_borrowed_range<T>` -> `std::ranges::enable_borrowed_range<T>` -> `::ranges::enable_borrowed_range<T>` -> `std::ranges::enable_borrowed_range<T>`.

I've chosen `std::ranges::{iota, subrange, ref_view, empty_view}`, because those are the only borrowed_ranges defined in the standard. And `::ranges` might have a fluctuating amount of them.

smehringer: ok. Thanks for explaining :)

smehringer (approved):

smehringer:
I guess you want to rebase

PR #2053, opened 2020-08-18
[TEST] less dependencies in ostream_test_template

marehr:


Irallia (approved):
If there is no issue, a description in the PR would be good, but I think I understand, whats going on here.
LGFM, thank you.

smehringer (approved):

PR #2056, opened 2020-08-18
[FIX] clang: use constexpr version of std::ceil(std::log2(x))

marehr:


rrahn:
all these clang issues? Can they be linked to the clang ticket?

include/seqan3/range/container/bitcompressed_vector.hpp
```diff
@@ -65,7 +66,7 @@ class bitcompressed_vector
 {
 private:
     //!\brief The number of bits needed to represent a single letter of the alphabet_type.
-    static constexpr size_t bits_per_letter = std::ceil(std::log2(alphabet_size<alphabet_type>));
+    static constexpr size_t bits_per_letter = detail::next_power_of_two(alphabet_size<alphabet_type>);
```
eseiler, line 69:
```suggestion
    static constexpr size_t bits_per_letter = detail::next_power_of_two(1ULL << detail::most_significant_bit_set(alphabet_size<alphabet_type>));
```

Should be correct. Can you check the indentation?

eseiler, line 69:
```suggestion
    static constexpr size_t bits_per_letter = detail::most_significant_bit_set(detail::next_power_of_two(alphabet_size<alphabet_type>));
```

Should be the way. Can you check intendation?

marehr: Hmm `next_power_of_two` says in the documentation: Returns 2^{\lceil\log_2(n)\rceil}. Ah we want that without the 2^. 

Shouldn't be detail::most_significant_bit_set enough then?

Maybe we should add the description that it also gives the `ceil(log_2(n))`

```suggestion
    static constexpr size_t bits_per_letter = detail::most_significant_bit_set(alphabet_size<alphabet_type>);
```

eseiler: Not really

E.g.:

```
62
0011 1110
```

MSB is 5, we need 6

```
64
0100 0000
```

MSB is 6, but we also just need 6 bit

eseiler (approved):

include/seqan3/core/math.hpp
```diff
@@ -16,7 +16,80 @@
...
+ *
+ * ### Thread-safety
+ *
+ * Thread safe.
+ *
+ * ### Complexity
+ *
+ * Constant.
+ */
+template <std::unsigned_integral unsigned_t>
```
smehringer, line 51: so we only allow unsigneds. shouldn't we make a a fallback to `std` for other integer types?

smehringer (commented):
One question, after that I'll merge.

include/seqan3/core/math.hpp
eseiler: `std::log2` return NaN (a special double/float) if the input is smaller than 0 and sets some funny error handling makro or whatever https://en.cppreference.com/w/cpp/numeric/math/log2

So, if we wanted integers we would need error handling since the msb stuff would still work.
And currently, there's no need for it?

marehr: The problem is that std::log2 is defined on "real" types. So they have to handle negative numbers, the logarithm is "normally" not defined on negative numbers and 0, so limiting this to unsigned types is reasonable.

smehringer (approved):

PR #2057, opened 2020-08-18
[FIX] use feature-testing macro for std::type_identity in type_traits

marehr:


wvandertoorn:
travis is unhappy about a type that uses the anonymous namespace [-Werror=subobject-linkage] in the `instatiate` struct in core/type_traits/lazy.hpp:49

wvandertoorn (changes requested):
Travis is unhappy about a type that uses the anonymous namespace [-Werror=subobject-linkage] in the `instatiate` struct in core/type_traits/lazy.hpp:49

What is the gain for using an anonymous namespace here, what does it fix?

eseiler:
> What is the gain for using an anonymous namespace here, what does it fix?

Basically the same as `static`, but you can use it for everything. E.g. you can't to a `stattic` type.

include/seqan3/std/type_traits
```diff
@@ -44,5 +46,6 @@ struct type_identity<t>
 //!\ingroup std_type_traits
 template <typename t>
 using type_identity_t = typename type_identity<t>::type;
+}
```
rrahn, line 49:
```suggestion
} // namespace
```

maybe?

marehr:
I used the proper approach and fixed it via the feature testing macro; clang currently misses that macro, hence I created a bug ticket https://bugs.llvm.org/show_bug.cgi?id=47261

smehringer (approved):

PR #2058, opened 2020-08-18
[TEST] simplify overly complicated decltype in single_pass_input_test

marehr:
clang complained:

```c++
error: member access into incomplete type 'single_pass_input<std::__1::vector<char, std::__1::allocator<char>>>'
    decltype(std::declval<single_pass_input>().get_data()) data;
                                              ^
```

### DIFF
--- a/test/unit/range/views/view_single_pass_input_test.cpp
+++ b/test/unit/range/views/view_single_pass_input_test.cpp
@@ -28,7 +28,7 @@ class single_pass_input : public ::testing::Test
         cmp_data = get_data();
     }
 
-    auto get_data()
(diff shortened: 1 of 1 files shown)


MitraDarja (approved):
LGTM, but can you explain to me why the static solved the error? :sweat_smile:

smehringer (approved):
@MitraDarja 

Making a function `static`, means you can call that function *without* an instance of the respective class. 
Before, `std::declval<single_pass_input>()` was creating an instance of `single_pass_input` just to call `get_data` on it. Now we don't need this anymore because `get_data` is static.

PR #2059, opened 2020-08-18
[FIX] pairwise_combine: compute size only once

marehr:


include/seqan3/range/views/pairwise_combine.hpp
```diff
@@ -219,7 +219,8 @@ class pairwise_combine_view : public std::ranges::view_interface<pairwise_combin
         requires std::ranges::sized_range<underlying_range_type>
     //!\endcond
     {
-        return (std::ranges::size(u_range) * (std::ranges::size(u_range) - 1) / 2);
+        auto size = std::ranges::size(u_range);
```
rrahn, line 222: Why is this necessary? std::ranges::size requires (semantically) constant time. Is there anything more to it, than just reducing one additional size call?
```suggestion
        auto const size = std::ranges::size(u_range);
```

Irallia (approved):
LGFM, thx

include/seqan3/range/views/pairwise_combine.hpp
marehr: I can also drop the PR. I found it easier to read `(size * (size - 1)) / 2` and you are correct that size is constant and I at first thought it wasn't, but this change will remove that doubt.

include/seqan3/range/views/pairwise_combine.hpp
```diff
@@ -219,7 +219,8 @@ class pairwise_combine_view : public std::ranges::view_interface<pairwise_combin
         requires std::ranges::sized_range<underlying_range_type>
     //!\endcond
     {
-        return (std::ranges::size(u_range) * (std::ranges::size(u_range) - 1) / 2);
+        auto const size = std::ranges::size(u_range);
```
smehringer, line 222: Since `size()` is constant. Does this PR really improve the performance? (Just asking, I don't care for either version evenif it does not)

smehringer (approved):
Isn't `size()` constant anyway?

include/seqan3/range/views/pairwise_combine.hpp
marehr: Nop; it shouldn't improve performance since it is constant time. It's more like a style guide, [FIX] might be the wrong tag for this.

smehringer: Alright.

PR #2064, opened 2020-08-19
Align cfg output/sequence ids

rrahn:
resolves seqan/product_backlog#191

### DIFF
--- a/include/seqan3/alignment/pairwise/alignment_algorithm.hpp
+++ b/include/seqan3/alignment/pairwise/alignment_algorithm.hpp
@@ -595,9 +595,9 @@ private:
     //!\cond
         requires (!traits_t::is_vectorised)
     //!\endcond
-    constexpr void make_alignment_result(index_t const idx,
-                                         sequence1_t & sequence1,
-                                         sequence2_t & sequence2,
+    constexpr void make_alignment_result([[maybe_unused]] index_t const idx,
+                                         [[maybe_unused]] sequence1_t & sequence1,
+                                         [[maybe_unused]] sequence2_t & sequence2,
                                          callback_t & callback)
     {
         using result_value_t = typename alignment_result_value_type_accessor<alignment_result_t>::type;
@@ -611,7 +611,11 @@ private:
 
         result_value_t res{};
 
-        res.id = idx;
+        if constexpr (traits_t::output_sequence1_id)
+            res.sequence1_id = idx;
+
+        if constexpr (traits_t::output_sequence2_id)
+            res.sequence2_id = idx;
 
         // Choose what needs to be computed.
         if constexpr (traits_t::compute_score)
@@ -690,7 +694,12 @@ private:
         {
             (void) sequence_pairs;
             result_value_t res{};
-            res.id = alignment_index;
+
+            if constexpr (traits_t::output_sequence1_id)
+                res.sequence1_id = alignment_index;
+
+            if constexpr (traits_t::output_sequence2_id)
+                res.sequence2_id = alignment_index;
 
             if constexpr (traits_t::compute_score)
                 res.score = this->alignment_state.optimum.score[simd_index];  // Just take this
--- a/include/seqan3/alignment/pairwise/alignment_result.hpp
+++ b/include/seqan3/alignment/pairwise/alignment_result.hpp
@@ -34,15 +34,17 @@ class policy_alignment_result_builder;
 /*!\brief A struct that contains the actual alignment result data.
  * \ingroup pairwise_alignment
  *
- * \tparam id_t                  The type for the alignment identifier.
- * \tparam score_t               The type for the resulting score.
- * \tparam end_positions_t       The type for the end positions, can be omitted.
- * \tparam begin_positions_t     The type for the begin positions, can be omitted.
- * \tparam alignment_t           The type for the alignment, can be omitted.
- * \tparam score_debug_matrix_t  The type for the score matrix. Only present if seqan3::align_cfg::detail::debug is enabled.
- * \tparam trace_debug_matrix_t  The type for the trace matrix. Only present if seqan3::align_cfg::detail::debug is enabled.
+ * \tparam sequence1_id_t        The type of the alignment identifier for the first sequence.
+ * \tparam sequence2_id_t        The type of the alignment identifier for the second sequence.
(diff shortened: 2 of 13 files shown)


include/seqan3/alignment/pairwise/alignment_algorithm.hpp
```diff
@@ -611,7 +611,11 @@ class alignment_algorithm :
 
         result_value_t res{};
 
-        res.id = idx;
+        if constexpr (traits_t::output_sequence1_id)
+            res.sequence1_id = idx;
+
+        if constexpr (traits_t::output_sequence2_id)
+            res.sequence2_id = idx;
```
MitraDarja, line 618: Why is res.sequence1_id and res.sequence2_id set to the same value here? They are not the same, are they?

include/seqan3/alignment/pairwise/alignment_result.hpp
```diff
@@ -34,15 +34,17 @@ class policy_alignment_result_builder;
 /*!\brief A struct that contains the actual alignment result data.
  * \ingroup pairwise_alignment
  *
- * \tparam id_t                  The type for the alignment identifier.
+ * \tparam sequence1_id_t        The type for the alignment identifier for the first sequence.
+ * \tparam sequence2_id_t        The type for the alignment identifier for the second sequence.
```
MitraDarja, line 38: The "type of" instead of "type for"?

include/seqan3/alignment/pairwise/detail/policy_alignment_result_builder.hpp
```diff
@@ -104,9 +104,11 @@ class policy_alignment_result_builder
 
         result_type result{};
 
-        static_assert(!std::same_as<decltype(result.data.id), invalid_t>,
-                      "Invalid configuration. Expected result with id!");
```
MitraDarja, line 108: Why do we not need this static assert anymore?

test/unit/alignment/pairwise/alignment_result_debug_stream_test.cpp
```diff
@@ -36,44 +36,45 @@ TEST(alignment_result_test, debug_streamable)
     seqan3::debug_stream_type debug_stream{ostream};
 
     { // Print id and score
-        seqan3::detail::alignment_result_value_type result_value{id, score};
+        seqan3::detail::alignment_result_value_type result_value{id, id, score};
```
MitraDarja, line 39: I find it confusing that we have two times id here... Maybe explain it in a comment? Otherwise it looks like a mistake.

MitraDarja (changes requested):
LGTM, but the tests are not running through and I have some questions. :)

include/seqan3/alignment/pairwise/alignment_algorithm.hpp
rrahn: At the moment they are the same. This has to be changed later when we actually allow the user to give the ids from outside.

include/seqan3/alignment/pairwise/detail/policy_alignment_result_builder.hpp
rrahn: Because before it was required to be set. Now it is optional and depends on the user input.

test/unit/alignment/pairwise/alignment_result_debug_stream_test.cpp
rrahn: Well, this is an intermediate fix for the tests. As it turned out we don't need the type deduction and also later we actually don't need the indirection of having a separate value type to set for the alignment_result. We changed that already for the search result. So at some point in the near future the whole test will be changed.

include/seqan3/alignment/pairwise/alignment_result.hpp
rrahn, line 38:
```suggestion
 * \tparam sequence1_id_t        The type of the alignment identifier for the first sequence.
 * \tparam sequence2_id_t        The type of the alignment identifier for the second sequence.
```

include/seqan3/alignment/pairwise/alignment_result.hpp
```diff
@@ -34,15 +34,17 @@ class policy_alignment_result_builder;
...
  * \tparam score_t               The type for the resulting score.
  * \tparam end_positions_t       The type for the end positions, can be omitted.
  * \tparam begin_positions_t     The type for the begin positions, can be omitted.
  * \tparam alignment_t           The type for the alignment, can be omitted.
  * \tparam score_debug_matrix_t  The type for the score matrix. Only present if seqan3::align_cfg::detail::debug is enabled.
  * \tparam trace_debug_matrix_t  The type for the trace matrix. Only present if seqan3::align_cfg::detail::debug is enabled.
```
rrahn, line 44:
```suggestion
 * \tparam sequence1_id_t        The type of the alignment identifier for the first sequence.
 * \tparam sequence2_id_t        The type of the alignment identifier for the second sequence.
 * \tparam score_t               The type of the resulting score.
 * \tparam end_positions_t       The type of the end positions, can be omitted.
 * \tparam begin_positions_t     The type of the begin positions, can be omitted.
 * \tparam alignment_t           The type of the alignment, can be omitted.
 * \tparam score_debug_matrix_t  The type of the score matrix. Only present if seqan3::align_cfg::detail::debug is enabled.
 * \tparam trace_debug_matrix_t  The type of the trace matrix. Only present if seqan3::align_cfg::detail::debug is enabled.
```

MitraDarja (approved):
Thanks for the clearifications! :)

smehringer:
Otherwise LGTM!

include/seqan3/alignment/pairwise/alignment_result.hpp
```diff
@@ -190,14 +210,26 @@ class alignment_result
      * \{
      */
 
-    /*!\brief Returns the alignment identifier.
-     * \return The id field.
+    /*!\brief Returns the alignment identifier of the first sequence.
+     * \return The id of the first sequence.
+     */
+    constexpr sequence1_id_t sequence1_id() const noexcept
```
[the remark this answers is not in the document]
rrahn: Not quite sure. It should take not too long. But definitely not in the current release. I will add a remark in the module documentation.

PR #2065, opened 2020-08-19
[FEATURE] Enable borrowed range feature for iota_simd_view.

rrahn:
resolves seqan/product_backlog#93


include/seqan3/core/simd/view_iota_simd.hpp
```diff
@@ -274,3 +274,11 @@ template <simd_concept index_simd_t>
 inline constexpr detail::iota_simd_view_fn<index_simd_t> iota_simd{};
 
 } // namespace seqan3::views
+
+namespace std::ranges
```
marehr, line 278: Option1:

```suggestion
namespace ranges
```

This would be the minimal working overload for now, because https://github.com/seqan/seqan3/pull/2038 defines `::std::ranges::enable_borrowed_range = ::ranges::enable_borrowed_range`.
---

Option 2:

```suggestion
#ifdef __cpp_lib_ranges // C++20 ranges available
namespace std::ranges
#else
namespace ranges
#endif 
```

This would be the right one, once https://github.com/ericniebler/range-v3/pull/1472 would be merged (it is on hold for some time now)

The problem is that this would only be half-correct with our current version of the ranges library, because in C++20 

```c++
::std::ranges::enable_borrowed_range<iota_simd_view> = true;
::ranges::enable_borrowed_range<iota_simd_view> = false; // does not use the std one.
```

----

Option 3:

```suggestion
namespace ranges
{
//!\cond
template <seqan3::simd_concept index_simd_t>
inline constexpr bool enable_borrowed_range<seqan3::detail::iota_simd_view<index_simd_t>> = true;
//!\endcond
}

#ifdef __cpp_lib_ranges
namespace std::ranges
{
//!\cond
template <seqan3::simd_concept index_simd_t>
inline constexpr bool enable_borrowed_range<seqan3::detail::iota_simd_view<index_simd_t>> = true;
//!\endcond
}
#endif // __cpp_lib_ranges
```

would be the "best", because we define that property for `std::ranges` and `ranges`.

----

All options have the drawback that they will not last forever. :(

include/seqan3/core/simd/view_iota_simd.hpp
```diff
@@ -274,3 +274,11 @@ template <simd_concept index_simd_t>
...
+{
+//!\cond
+template <seqan3::simd_concept index_simd_t>
+inline constexpr bool enable_borrowed_range<seqan3::detail::iota_simd_view<index_simd_t>> = true;
+//!\endcond
+}
```
marehr, line 284:
```suggestion
} // namespace std::ranges
```

marehr (changes requested):
Pick one option :) This should be marked as blocked by https://github.com/seqan/seqan3/pull/2038

include/seqan3/core/simd/view_iota_simd.hpp
rrahn: Ah what a pitty. I thought somehow this was all good already. But I see now that we have to overload the ranges namespace if we don't run on c++20 😢

marehr (approved):

smehringer (approved):

PR #2068, opened 2020-08-20
Remove dependencies of seqan3 in `seqan3/std/*` headers

marehr:
Fixes seqan/product_backlog#135

### DIFF
--- a/include/seqan3/std/concepts
+++ b/include/seqan3/std/concepts
@@ -22,8 +22,6 @@
 #include <concepts/concepts.hpp>
 #include <range/v3/functional/concepts.hpp>
 
-#include <seqan3/core/platform.hpp>
-
 /*!\defgroup std std
  * \brief A subset of the C++20 standard library made available in pre-C++20 contexts.
  *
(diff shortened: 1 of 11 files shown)


wvandertoorn (approved):

include/seqan3/std/concepts
```diff
@@ -86,7 +84,7 @@ using ::concepts::unsigned_integral;
 // using ::concepts::floating_point;
 //!\cond
 template <typename t>
-SEQAN3_CONCEPT floating_point = std::is_floating_point_v<t>;
+META_CONCEPT floating_point = std::is_floating_point_v<t>;
```
smehringer, line 87: where does the META_CONCEPT come from ?

smehringer (commented):
So we don't need any of the stuff defined in platform in the std headers?

include/seqan3/std/concepts
eseiler: From the ranges library

marehr: > So we don't need any of the stuff defined in platform in the std headers?

In C++10 mode, we also don't import any seqan3 code, so we should make those headers agnostic of them.

smehringer (approved):

PR #2071, opened 2020-08-24
[INFRA] Add macOS to GitHub actions

eseiler:
Unfortunately, we can't really simplify the matrix.
If we do everything for both OS, we need to exclude the stuff we don't want - which is harder than adding the stuff we want.

You can write:
```yaml
matrix:
  os: [ubuntu-20.04, macos-10.15]
```

But not
```yaml
matrix:
  include:
  - ...
    os: [ubuntu-20.04, macos-10.15]
```

Even though both will be available as `matrix.os`. The latter will be interpreted as `ubuntu-20.04, macos-10.15` instead of `ubuntu-20.04` and `macos-10.15`.

### DIFF
--- a/test/unit/core/type_traits/range_iterator_test.cpp
+++ b/test/unit/core/type_traits/range_iterator_test.cpp
@@ -159,13 +159,19 @@ TEST(range_and_iterator, difference_type)
     //TODO(h-2): add something that actually has a different difference_type
     using iterator_of_int_vector = std::ranges::iterator_t<std::vector<int>>;
     using foreign_iterator = seqan3::detail::random_access_iterator<std::vector<int>>;
-    auto v = std::views::iota(1);
     using type_list_example = seqan3::type_list<std::ranges::range_difference_t<std::vector<int>>, // short
                                                 typename std::vector<int>::difference_type, // member type
                                                 std::ranges::range_difference_t<std::vector<int> const>, // const container
                                                 std::iter_difference_t<iterator_of_int_vector>, // iterator
-                                                std::iter_difference_t<foreign_iterator>, // iterator2
-                                                std::ranges::range_difference_t<decltype(v)>>; // range, no member
+                                                std::iter_difference_t<foreign_iterator>>; // range, no member
+
+    using comp_list = seqan3::type_list<std::ptrdiff_t,
+                                        std::ptrdiff_t,
+                                        std::ptrdiff_t,
+                                        std::ptrdiff_t,
+                                        std::ptrdiff_t>;
+
+    expect_same_types<type_list_example, comp_list>();
 
     // views::ints' difference_type is not std::ptrdiff_t, but depends on the size.
     // For infinite views, like in our case, it's std::int_fast64_t (or std::int_fast32_t on 32bit).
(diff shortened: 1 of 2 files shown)


eseiler:
MacOS gcc10 error:

```
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from range_and_iterator
[ RUN      ] range_and_iterator.difference_type
pos: 5 'long' not equal to 'long long' 
../../../seqan3/test/unit/core/type_traits/range_iterator_test.cpp:63: Failure
Value of: val
  Actual: false
Expected: true
[  FAILED  ] range_and_iterator.difference_type (0 ms)
[----------] 1 test from range_and_iterator (0 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (0 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] range_and_iterator.difference_type
```

test/unit/core/type_traits/range_iterator_test.cpp
```diff
@@ -53,11 +53,18 @@ TEST(range_and_iterator, sentinel)
...
+
     if constexpr (!val)
     {
-        std::cerr << "pos: " << pos << " \'" << seqan3::detail::type_name_as_string<seqan3::list_traits::at<pos, list1>>
-                  << "\' not equal to \'" << seqan3::detail::type_name_as_string<seqan3::list_traits::at<pos, list2>>
+        std::cerr << "pos: " << pos << " \'" << seqan3::detail::type_name_as_string<type_1>
+                  << "\' not equal to \'" << seqan3::detail::type_name_as_string<type_2>
                   << "\' \n";
     }
     EXPECT_TRUE(val);
```
marehr, line 70: I wouldn't want to change this here.

I would just change the offending test case

```patch
-------------- test/unit/core/type_traits/range_iterator_test.cpp --------------
index 008deca9c..417d6d69d 100644
@@ -158,15 +158,21 @@ TEST(range_and_iterator, difference_type)
 {
     //TODO(h-2): add something that actually has a different difference_type
     using iterator_of_int_vector = std::ranges::iterator_t<std::vector<int>>;
     using foreign_iterator = seqan3::detail::random_access_iterator<std::vector<int>>;
-    auto v = std::views::iota(1);
     using type_list_example = seqan3::type_list<std::ranges::range_difference_t<std::vector<int>>, // short
                                                 typename std::vector<int>::difference_type, // member type
                                                 std::ranges::range_difference_t<std::vector<int> const>, // const container
                                                 std::iter_difference_t<iterator_of_int_vector>, // iterator
-                                                std::iter_difference_t<foreign_iterator>, // iterator2
-                                                std::ranges::range_difference_t<decltype(v)>>; // range, no member
+                                                std::iter_difference_t<foreign_iterator>>; // range, no member
+
+    using comp_list = seqan3::type_list<std::ptrdiff_t,
+                                        std::ptrdiff_t,
+                                        std::ptrdiff_t,
+                                        std::ptrdiff_t,
+                                        std::ptrdiff_t>;
+
+    expect_same_types<type_list_example, comp_list>();
 
     // views::ints' difference_type is not std::ptrdiff_t, but depends on the size.
     // For infinite views, like in our case, it's std::int_fast64_t (or std::int_fast32_t on 32bit).
     // On most platforms this is the same as std::size_t (`long int` on Linux and FreeBSD;
@@ -176,16 +182,13 @@ TEST(range_and_iterator, difference_type)
     using view_int_diff_t = std::conditional_t<sizeof(std::ptrdiff_t) == 8,
                                                std::int_fast64_t,
                                                std::int_fast32_t>;
 
-    using comp_list = seqan3::type_list<std::ptrdiff_t,
-                                        std::ptrdiff_t,
-                                        std::ptrdiff_t,
-                                        std::ptrdiff_t,
-                                        std::ptrdiff_t,
-                                        view_int_diff_t>;
-
-    expect_same_types<type_list_example, comp_list>();
+    auto v = std::views::iota(1);
+    using iota_difference_t = std::ranges::range_difference_t<decltype(v)>;
+    EXPECT_TRUE((std::signed_integral<iota_difference_t>));
+    EXPECT_EQ(sizeof(iota_difference_t), sizeof(view_int_diff_t));
 }
 
 TEST(range_and_iterator, size_type)
 {
```

marehr (approved):

smehringer (approved):

PR #2075, opened 2020-08-25
[INFRA] Run CI with SeqAn2 included

eseiler:


joergi-w (approved):
Looks good! Thank you :)

doc/tutorial/pairwise_alignment/index.md
```diff
@@ -21,7 +21,7 @@ The goal of the pairwise alignment is to obtain an optimal transcript that descr
 to each other by means of substitutions, insertions, or deletions. The computed transcript describes then the operations
 necessary to translate the one sequence into the other, as can be seen in the following picture.
 
-\image html align_transcript.png width=800px
+\image html transcript_alignment.png width=800px
```
marehr, line 24: I understand this fix, but shouldn't the actual question be, why a different path is even considered and why this happens?

This seems like an awful thing, if doxygen just searches some random paths. This can be really annoying if we have other external sources that have images.

Would a relative path ./align_transcript.png fix this to?

eseiler: You are right, we can just change the IMAGE_PATH of doxygen to not scan the entire directory, but just `doc` and `test/documentation` since that is where we but the images

marehr: That would be great :)

marehr (approved):
thank you!

PR #2076, opened 2020-08-27
[FEATURE] Node access for fm_index_cursor

### LINKED ISSUE #2061: FM index cursor enable node access
Hey,

I have one more request concerning the FM index cursor. While with #2048  you provide cerealisation for FM index cursors (which is great and clearly resolves #2044 ), this unfortunately does not resolve a different issue I have with the current implementation of the cursor.

### Need
For my FM index algorithm, I need to compare the locations of FM index cursors. I use this for sorting, removing duplicates, detecting nodes that are "contained" in other nodes, etc.. Therefore, I need access to the cursor members to compare cursors with each other, either manually or by enabling the implementation of operators such as < to use stl functions.

### Solution
To be able to do this with SeqAn3, I would need to access at least the node of FM index cursors which is currently not possible (private member). Would it be possible to provide access to the node of FM index cursors? As these are not too big, I guess a getter returning a copy would in principle be fine in case you want to ensure that corruption of a cursor is not possible, though returning a reference or a getter to access each node member individually would probably be more efficient.

### Notes
I am aware that `==` and `!=` operators for FM index cursors are defined. Still, this is not sufficient for my purpose as I need very special types of comparisons which would probably not be appropriate for general purpose (e.g., in some of my comparisons the depth should not be included in the comparison, which it currently is for the == operator which generally also makes sense). This is why I would like to have access to the members and define my own comparisons instead of asking for operators directly.

eseiler:
Resolves #2061

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,14 @@ The following API changes should be documented as such:
 If possible, provide tooling that performs the changes, e.g. a shell-script.
 -->
 
+# 3.0.3
+
+## New features
+
+### Search
+
+* The `seqan3::fm_index_cursor` exposes its suffix array interval ([\#2076](https://github.com/seqan/seqan3/pull/2076)).
+
 # 3.0.2
 
 Note that 3.1.0 will be the first API stable release and interfaces in this release might still change.
--- a/include/seqan3/search/fm_index/fm_index_cursor.hpp
+++ b/include/seqan3/search/fm_index/fm_index_cursor.hpp
@@ -32,7 +32,7 @@ namespace seqan3
  * \{
  */
 
-//!\brief The underlying suffix array intervals.
+//!\brief The underlying suffix array interval.
 struct suffix_array_interval
 {
     //!\brief The begin position of the interval ("left boundary").
--- a/include/seqan3/search/fm_index/fm_index_cursor.hpp
+++ b/include/seqan3/search/fm_index/fm_index_cursor.hpp
@@ -32,6 +32,39 @@ namespace seqan3
  * \{
  */
 
+//!\brief The underlying suffix array intervals.
+struct suffix_array_interval
+{
+    //!\brief The begin position of the interval ("left boundary").
+    size_t begin_position;
+    //!\brief The exclusive end position of the interval ("right boundary").
+    size_t end_position;
+
+    /*!\name Comparison operators
+     * \{
+     */
+    /*!\brief Test for equality.
+     * \param[in] lhs A `seqan3::suffix_array_interval`.
+     * \param[in] rhs `seqan3::suffix_array_interval` to compare to.
+     * \returns `true` if equal, `false` otherwise.
+     */
+    friend bool operator==(suffix_array_interval const & lhs, suffix_array_interval const & rhs) noexcept
+    {
+        return lhs.begin_position == rhs.begin_position && lhs.end_position == rhs.end_position;
+    }
+
+    /*!\brief Test for inequality.
+     * \param[in] lhs A `seqan3::suffix_array_interval`.
+     * \param[in] rhs `seqan3::suffix_array_interval` to compare to.
+     * \returns `true` if unequal, `false` otherwise.
+     */
+    friend bool operator!=(suffix_array_interval const & lhs, suffix_array_interval const & rhs) noexcept
+    {
+        return !(lhs == rhs);
+    }
+    //!\}
+};
+
 /*!\brief The SeqAn FM Index Cursor.
  * \implements seqan3::fm_index_cursor_specialisation
  * \tparam index_t The type of the underlying index; must model seqan3::fm_index_specialisation.
@@ -168,7 +201,9 @@ public:
         index(&_index),
         node({0, _index.index.size() - 1, 0, 0}),
         sigma(_index.index.sigma - index_t::text_layout_mode)
-    {}
+    {
+        assert(_index.index.size() != 0);
+    }
     //\}
 
     /*!\brief Compares two cursors.
@@ -364,7 +399,7 @@ public:
      * \returns `true` if there exists a query in the text where the rightmost character of the query is
      *          lexicographically larger than the current rightmost character of the query.
      *
-     * Example:
+     * ### Example
      *
      * \include test/snippet/search/fm_index_cursor.cpp
      *
@@ -404,7 +439,7 @@ public:
     /*!\brief Outputs the rightmost rank.
      * \returns Rightmost rank.
      *
-     * Example:
+     * ### Example
      *
      * \include test/snippet/search/fm_index_cursor.cpp
      *
@@ -424,11 +459,37 @@ public:
         return index->index.comp2char[node.last_char] - 1; // text is not allowed to contain ranks of 0
     }
 
+    /*!\brief Returns the half-open suffix array interval.
+     * \returns A seqan3::suffix_array_interval contains the half-open interval.
+     *
+     * ### Example
+     *
+     * \include test/snippet/search/fm_index_cursor.cpp
+     *
+     * ### Complexity
+     *
+     * Constant.
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    seqan3::suffix_array_interval suffix_array_interval() const noexcept
+    {
+        assert(index != nullptr);
+
+        return {node.lb, node.rb + 1};
+    }
+
     /*!\brief Returns the length of the searched query.
      *        \if DEV
      *            Returns the depth of the cursor node in the implicit suffix tree.
      *        \endif
      * \returns Length of query.
+     *
+     * ### Example
+     *
+     * \include test/snippet/search/fm_index_cursor.cpp
      *
      * ### Complexity
      *
--- a/test/unit/search/fm_index_cursor/fm_index_cursor_collection_test_template.hpp
+++ b/test/unit/search/fm_index_cursor/fm_index_cursor_collection_test_template.hpp
@@ -6,15 +6,14 @@
 // -----------------------------------------------------------------------------------------------------
 
 #include <gtest/gtest.h>
+
+#include <seqan3/std/algorithm>
 #include <type_traits>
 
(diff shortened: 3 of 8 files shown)


marehr (commented):
I thought we wanted to expose the intervals; I'm not in favour for exposing the node. If we can, I would want to hide them. I'm marking this as WIP for now, because it is still not clear what the use case is.

CHANGELOG.md
```diff
@@ -24,6 +24,15 @@ The following API changes should be documented as such:
...
 -->
 
+# 3.0.3
+
+## New features
+
+### Search
+
+* The `seqan3::fm_index_cursor` exposes its left and right bound in the suffix array
+  ([\#2061](https://github.com/seqan/seqan3/pull/2061)).
```
eseiler, line 34:
```suggestion
  ([\#2076](https://github.com/seqan/seqan3/pull/2076)).
```

test/unit/search/fm_index_cursor/fm_index_cursor_collection_test_template.hpp
```diff
@@ -156,10 +185,20 @@ TYPED_TEST_P(fm_index_cursor_collection_test, extend_right_char)
...
+        EXPECT_EQ(it.right_bound(), 4u);
+    }
 
     EXPECT_TRUE(it.extend_right(seqan3::views::slice(this->text1, 1, 2)));   // "C"
     EXPECT_EQ(seqan3::uniquify(it.locate()), (std::vector<std::pair<uint64_t, uint64_t>>{{0,0}, {0,3}}));
     EXPECT_EQ(it.query_length(), 2u);
+    if constexpr (!seqan3::bi_fm_index_specialisation<typename TypeParam::index_type>)
+    {
+        EXPECT_EQ(it.left_bound(), 5u);
+        EXPECT_EQ(it.right_bound(), 6u);
```
marehr, line 200: If bidirectional would implement this, would it return the same values?

include/seqan3/search/fm_index/fm_index_cursor.hpp
```diff
@@ -423,6 +423,42 @@ class fm_index_cursor
...
+     *
+     * ### Complexity
+     *
+     * Constant.
+     *
+     * ### Exceptions
+     *
+     * No-throw guarantee.
+     */
+    size_type left_bound() const noexcept
```
marehr, line 437: A general question: would `std::pair suffix_array_bounds` be a better function?

marehr (commented):
Looks already pretty good, just one question

test/unit/search/fm_index_cursor/fm_index_cursor_collection_test_template.hpp
eseiler: I would delegate the `left_bound` on a `bi_fm_index` to the `left_bound` of the forward `fm_index`, so it should return the same values.

include/seqan3/search/fm_index/fm_index_cursor.hpp
eseiler: Good idea, may be more concise. I'll ask Tobias.

eseiler: I'm not sure, we can probably leave it as is?
If we change it, I would rather like to return some struct `suffix_array_bounds` that has a `left` and `right` member.

marehr:
Core Meeting 2020-09-08
-----

We want a `suffix_array_interval()` function that returns a `seqan3::suffix_array_interval` which is an aggregate struct and has the two members `left_bound` and `right_bound` (of size_t) and behaves like a `std::pair`. 

Document if half-open or not.

marehr (commented):
(Waiting on implementing the half-open interval accessor.)

eseiler:
@marehr can you have a quick look at `include/seqan3/search/fm_index/fm_index_cursor.hpp`

If it is fine, I can start adjusting all the tests and documentation

marehr:
> @marehr can you have a quick look at `include/seqan3/search/fm_index/fm_index_cursor.hpp`
> 
> If it is fine, I can start adjusting all the tests and documentation

looks fine :) Or do you think that something seems odd?

eseiler:
@marehr I also added a conversion operator to tuple ... it is nice for the use case where you want to reuse variables (see snippet)

```cpp
auto [left_bound, right_bound] = cur.suffix_array_interval();
/// .. do some stuff with cursor
std::tie(left_bound, right_bound) = cur.suffix_array_interval();
```

include/seqan3/search/fm_index/fm_index_cursor.hpp
```diff
@@ -32,6 +32,45 @@ namespace seqan3
...
+//!\brief The underlying suffix array intervals.
+struct suffix_array_interval
+{
+    //!\brief The begin position of the interval ("left boundary").
+    size_t begin_position;
+    //!\brief The exclusive end position of the interval ("right boundary").
+    size_t end_position;
+
+    //!\brief Convert to tuple.
+    operator std::tuple<size_t &, size_t &>()
```
marehr, line 44:
```suggestion
    operator std::tuple<size_t &, size_t &>() &
```

marehr: I don't know, you can also do this:

```c++

struct foo
{
  int a;
  int b;
};

int main()
{
    foo bar{1, 3};

    auto & [a, b] = bar;

    b = 5;

    return bar.b; // has value 5
}
```

so we don't need a std::tie.

include/seqan3/search/fm_index/fm_index_cursor.hpp
```diff
@@ -32,6 +32,45 @@ namespace seqan3
...
+     * \{
+     */
+    /*!\brief Test for equality.
+     * \param[in] lhs A `seqan3::suffix_array_interval`.
+     * \param[in] rhs `seqan3::suffix_array_interval` to compare to.
+     * \returns `true` if equal, `false` otherwise.
+     */
+    friend bool operator==(suffix_array_interval const & lhs, suffix_array_interval const & rhs) noexcept
+    {
+        return std::tie(lhs.begin_position, lhs.end_position) == std::tie(rhs.begin_position, rhs.end_position);
```
marehr, line 59:
```suggestion
        return lhs.begin_position == rhs.begin_position && lhs.end_position == rhs.end_position;
```

name me old fashioned, but why not like this?

marehr (commented):
Looks awesome :)

Do we really need the tuple thingy?

Did you find any instances where the interval can be empty?

include/seqan3/search/fm_index/fm_index_cursor.hpp
eseiler: But what I want is 
```cpp
int main()
{
    foo bar{1, 3};

    auto & [a, b] = bar;

    b = 5;

    foo bar2{2, 4};
    std::tie(a, b) = bar2;

    return bar.b; // has value 5
}
```

eseiler:
> Did you find any instances where the interval can be empty?

No, the only thing that could happen is that you construct the cursor on an empty index.
Internally, closed intervals are used, so `rb` would underflow. But this was already a problem before...
I added an assert when constructing the cursor, such that it will assert if constructed on an empty index.

The most elegant solution would be to use half-open intervals in the FM-index, because then the cursor is always in a valid state (The cursor on an empty index would then use `[0,0}` instead of `[0, size_t{-1}]`.)

include/seqan3/search/fm_index/fm_index_cursor.hpp
eseiler: I tried the `&`:

https://godbolt.org/z/4E6KcT

It doesn't work because .. ehm .. not so sure...
It would work if we the suffix_array_interval struct wasn't generated by another struct.

rrahn, line 44:
```suggestion
    operator std::tuple<size_t, size_t>() const noexcept
```
So this wouldn't work?

Because your version should require that this is an lvalue reference as well. Then the question remains, what is with the other versions `const &` should be also added. Move version are not required since we move/copy built-in types anyway.

eseiler: Nope https://godbolt.org/z/j9avh9

marehr (approved):

include/seqan3/search/fm_index/fm_index_cursor.hpp
```diff
@@ -32,6 +32,39 @@ namespace seqan3
  * \{
  */
 
+//!\brief The underlying suffix array intervals.
```
rrahn, line 35:
```suggestion
//!\brief The underlying suffix array interval.
```

test/unit/search/fm_index_cursor/fm_index_cursor_collection_test_template.hpp
```diff
@@ -44,6 +43,10 @@ TYPED_TEST_P(fm_index_cursor_collection_test, ctr)
     // custom constructor
     TypeParam it0{fm};
     EXPECT_EQ(it0.query_length(), 0u);
+    if constexpr (!seqan3::bi_fm_index_specialisation<typename TypeParam::index_type>)
```
rrahn, line 46: Could this not be part of the test fixture?

test/unit/search/fm_index_cursor/fm_index_cursor_test_template.hpp
```diff
@@ -47,6 +46,10 @@ TYPED_TEST_P(fm_index_cursor_test, ctr)
     // custom constructor
     TypeParam it0{fm};
     EXPECT_EQ(it0.query_length(), 0u);
+    if constexpr (!seqan3::bi_fm_index_specialisation<typename TypeParam::index_type>)
```
rrahn, line 49: same here.  So I would call: 

```cpp
if constexpr (!TestFixture::is_bi_fm_index)
{
...
}
```

rrahn (changes requested):
Looks good so far. Some minor change requrests.

rrahn (approved):
Thanks a lot 👍

PR #2078, opened 2020-08-28
[INFRA] Cache downloads

eseiler:


MitraDarja (approved):

.github/workflows/ci.yml
```diff
@@ -174,6 +174,12 @@ jobs:
           ref: develop
           path: seqan3/submodules/seqan
 
+      - name: Get cached CMake
+        uses: actions/cache@v2
+        with:
+          path: /tmp/cmake-download
+          key: ${{ runner.os }}-CMake_${{ env.CMAKE_VERSION }}
+
       - name: Setup CMake
```
smehringer, line 183: So this is just to use the download cache (of the program)?
I guess if the program in `tmp/cmake-download` does not exist it is downloaded?

smehringer (approved):

.github/workflows/ci.yml
eseiler: Yes.

In the next step, we do `wget --no-clobber ...` which means download if the file does not exist.

PR #2079, opened 2020-08-29
[FIX] Move remove_cvref_t into std namespace

wvandertoorn:
Fixes https://github.com/seqan/product_backlog/issues/134

### DIFF
--- a/include/seqan3/std/type_traits
+++ b/include/seqan3/std/type_traits
@@ -15,13 +15,25 @@
 
 #include <type_traits>
 
+//!\brief A workaround for __cpp_lib_remove_cvref for gcc version >=9.0 and <9.4 (in C++17 mode).
+//!       Those versions implemented std::remove_cvref_t, but did not define that feature detection macro.
+#ifndef SEQAN3_WORKAROUND_CPP_LIB_REMOVE_CVREF
+#   if defined(__cpp_lib_remove_cvref)
+#       define SEQAN3_WORKAROUND_CPP_LIB_REMOVE_CVREF 1
+#   elif defined(__GNUC__) && ((__GNUC__ == 9) && (__GNUC_MINOR__ < 4)) && __cplusplus > 201703L
+#       define SEQAN3_WORKAROUND_CPP_LIB_REMOVE_CVREF 1
+#   else
+#       define SEQAN3_WORKAROUND_CPP_LIB_REMOVE_CVREF 0
+#   endif
+#endif
+
 namespace std
 {
 
 // ----------------------------------------------------------------------------
 // remove_cvref_t
 // ----------------------------------------------------------------------------
-#ifndef __cpp_lib_remove_cvref
+#if !SEQAN3_WORKAROUND_CPP_LIB_REMOVE_CVREF
 /*!\brief Return the input type with `const`, `volatile` and references removed.
  * \tparam t The type to operate on.
  */
@@ -42,7 +54,7 @@ struct remove_cvref<t> // needed for gcc-9, it defines std::remove_cvref but doe
  */
 template <typename t>
 using remove_cvref_t = typename remove_cvref<t>::type;
-#endif // __cpp_lib_remove_cvref
+#endif // !SEQAN3_WORKAROUND_CPP_LIB_REMOVE_CVREF
 
 // ----------------------------------------------------------------------------
 // type_identity
--- a/include/seqan3/std/type_traits
+++ b/include/seqan3/std/type_traits
@@ -18,6 +18,35 @@
 namespace std
 {
 
+// ----------------------------------------------------------------------------
+// remove_cvref_t
+// ----------------------------------------------------------------------------
+#ifndef __cpp_lib_remove_cvref
+/*!\brief Return the input type with `const`, `volatile` and references removed.
+ * \tparam t The type to operate on.
+ */
+template <typename t>
+struct remove_cvref;
+
+//!\cond
+template <typename t>
+    requires true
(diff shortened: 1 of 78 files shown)


eseiler:
<details><summary>Hint for failing travis (click me):</summary>

`std::remove_cvref_t` is defined in cpp20 mode.
So you need to wrap the definition in a feature test macro, i.e.
```cpp
#if feature_test_macro
template <typename t>
struct remove_cvref
...
#endif
```

You can look up specific macros here: https://en.cppreference.com/w/cpp/utility/feature_test
You can either search for the header they are defined in (`type_traits`) or search for the definition you are looking for (`remove_cvref`).

I think you can do the same with the `type_identity` that is defined in the same file while you are at it. I should have the same problem, but apparently we are not using it?

</details>

include/seqan3/std/type_traits
```diff
@@ -23,6 +23,27 @@ namespace std
...
+struct remove_cvref
+{
+    using type = std::remove_cv_t<std::remove_reference_t<t>>;
+};
+
+/*!\brief Return the input type with `const`, `volatile` and references removed (transformation_trait shortcut).
+ * \tparam t The type to operate on.
+ */
+template <typename t>
+using remove_cvref_t = typename remove_cvref<t>::type;
```
marehr, line 42:
```suggestion
#ifndef __cpp_lib_remove_cvref
template <typename t>
struct remove_cvref
{
    using type = std::remove_cv_t<std::remove_reference_t<t>>;
};

/*!\brief Return the input type with `const`, `volatile` and references removed (transformation_trait shortcut).
 * \tparam t The type to operate on.
 */
template <typename t>
using remove_cvref_t = typename remove_cvref<t>::type;
#endif // __cpp_lib_remove_cvref
```

marehr:
This should go against release;

Can you split the single commit into three commits?

1) add `remove_cvref` to  `include/seqan3/std/type_traits`
2) change all occurrences to `std::remove_cvref`
3) deprecate `seqan3::remove_cvref` (not removing it)

include/seqan3/std/type_traits
```diff
@@ -23,6 +23,29 @@ namespace std
...
+// ----------------------------------------------------------------------------
+/*!\brief Return the input type with `const`, `volatile` and references removed.
+ * \tparam t The type to operate on.
+ */
+#ifndef __cpp_lib_remove_cvref
+template <typename t>
+struct remove_cvref
+{
+    using type = std::remove_cv_t<std::remove_reference_t<t>>;
+};
```
marehr, line 37:
```suggestion
#ifndef __cpp_lib_remove_cvref
/*!\brief Return the input type with `const`, `volatile` and references removed.
 * \tparam t The type to operate on.
 */
template <typename t>
struct remove_cvref;

//!\cond
template <typename t>
    requires true
struct remove_cvref<t> // needed for gcc-9, it defines std::remove_cvref but does not populate __cpp_lib_remove_cvref
{
    using type = std::remove_cv_t<std::remove_reference_t<t>>;
};
//!\endcond
```

wvandertoorn:
Thank a lot both for the suggestions and help!

include/seqan3/std/type_traits
```diff
@@ -24,6 +24,35 @@ namespace std
...
+//!\endcond
+
+/*!\brief Return the input type with `const`, `volatile` and references removed (transformation_trait shortcut).
+ * \tparam t The type to operate on.
+ */
+template <typename t>
+using remove_cvref_t = typename remove_cvref<t>::type;
```
eseiler, line 50:
```suggestion
template <typename t>
//!\cond
    requires true // needed for gcc-9, it defines std::remove_cvref but does not populate __cpp_lib_remove_cvref
//!\endcond
using remove_cvref_t = typename remove_cvref<t>::type;
```

Apparently this only happens for GCC9 on MacOS

eseiler: This should also fail on gcc9 linux...

Also, in line 21: `#ifndef __cpp_lib_type_identity`
should go down to line 56

include/seqan3/alignment/matrix/detail/combined_score_and_trace_matrix.hpp
```diff
@@ -0,0 +1,354 @@
+// -----------------------------------------------------------------------------------------------------
```
eseiler, line 1: Rebase errror?

include/seqan3/std/type_traits
eseiler: @marehr That means we need a workaround macro? `__cpp_lib_remove_cvref` not being defined will be fixed in 9.4

marehr:
@wvandertoorn Is this still a draft?

wvandertoorn:
@marehr Kinda, I was waiting on input to Erico's question on to how to handle the failing macro.

> @marehr That means we need a workaround macro? `__cpp_lib_remove_cvref` not being defined will be fixed in 9.4

include/seqan3/std/type_traits
eseiler: Not really, this is the patch http://gcc.1065356.n8.nabble.com/committed-0-8-libstdc-Add-update-fix-feature-test-macros-td1682441.html

This is where I found it (remove_cvref has a note that says since 9.4 https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html)

marehr:
@eseiler Can you have a look? I think I got it fixed :)

marehr (approved):

eseiler (approved):

Issue #1482, opened 2020-01-14
Infra: Enable arm64 CI builds


mr-c:
Now that arm64 is working, lets enable CI for it as well: https://docs.travis-ci.com/user/multi-cpu-architectures/

This can also be used for ppc64le & s390x

PR #2083, opened 2020-08-31
[INFRA] Run travis as nightly on multiple CPU archs

eseiler:
Resolves https://github.com/seqan/seqan3/issues/1482

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,107 +20,31 @@ addons:
       - g++-9
       - g++-10
       - lcov
+      - cmake
+arch:
+  - arm64
+  - s390x
+  - ppc64le
 
-# https://docs.travis-ci.com/user/languages/c
-jobs:
-  include:
-    - name: "Coverage gcc7"
-      env:
-        - CXX=g++-7
-        - CC=gcc-7
-        - BUILD=coverage
-        - BUILD_TYPE=Debug
-    - name: "Unit gcc9 (c++2a)"
-      env:
-        - CXX=g++-9
-        - CC=gcc-9
-        - BUILD=unit
-        - BUILD_TYPE=Release
-        - CXXFLAGS="-std=c++2a"
-    - name: "Unit gcc10 (c++17)"
-      env:
-        - CXX=g++-10
-        - CC=gcc-10
-        - BUILD=unit
-        - BUILD_TYPE=Release
-        - CXXFLAGS="-std=c++17 -fconcepts"
-    - name: "Unit gcc10 (c++20)"
-      env:
-        - CXX=g++-10
-        - CC=gcc-10
-        - BUILD=unit
-        - BUILD_TYPE=Release
-    - name: "Unit gcc8"
-      env:
-        - CXX=g++-8
-        - CC=gcc-8
-        - BUILD=unit
-        - BUILD_TYPE=Release
-    - name: "Unit gcc7"
-      env:
-        - CXX=g++-7
-        - CC=gcc-7
-        - BUILD=unit
-        - BUILD_TYPE=Release
-    - name: "Performance gcc7"
-      env:
-        - CXX=g++-7
-        - CC=gcc-7
-        - BUILD=performance
-        - BUILD_TYPE=Release
-    - name: "Header gcc7"
-      env:
-        - CXX=g++-7
-        - CC=gcc-7
-        - BUILD=header
-        - BUILD_TYPE=Release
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -20,107 +20,31 @@ addons:
...
+  - BUILD=header      CXX=g++-7  CC=gcc-7  BUILD_TYPE=Release
+  - BUILD=snippet     CXX=g++-7  CC=gcc-7  BUILD_TYPE=Release
 
 install:
-  - |
-    HOME_BIN_PATH=`realpath ~/bin`
-    mkdir -p ${HOME_BIN_PATH}
-  - |
-    # install cmake
-    CMAKE_VERSION="3.7.2"
```
marehr, line 109: If we don't fix the cmake version, the log will get to big, we need to fix our cmake script to use the less verbose version.

```cmake
    if (NOT CMAKE_VERSION VERSION_LESS 3.10) # cmake >= 3.10
        gtest_discover_tests(${target} TEST_PREFIX "${test_name}::" PROPERTIES TIMEOUT "30")
    else ()
        add_test (NAME "${test_name}" COMMAND ${target})
    endif ()
```

should be

```cmake
    option (SEQAN3_VERBOSE_TESTS ON)
    if (SEQAN3_TEST_VERBOSE_TESTS AND NOT CMAKE_VERSION VERSION_LESS 3.10) # cmake >= 3.10
        gtest_discover_tests(${target} TEST_PREFIX "${test_name}::" PROPERTIES TIMEOUT "30")
    else ()
        add_test (NAME "${test_name}" COMMAND ${target})
    endif ()
```

marehr (commented):
looks good, we need to make the test output shorter for travis

.travis.yml
eseiler: That means we should run the nightlies with `-DSEQAN3_VERBOSE_TESTS=OFF` ?

marehr: jop

eseiler:
I will do the adjustements to the cmake script in the release branch (CI should also use SEQAN3_VERBOSE_TESTS=OFF)

marehr (approved):
lgtm!

.travis.yml
```diff
@@ -20,107 +20,36 @@ addons:
...
       - g++-10
       - lcov
+      - cmake
+arch:
+  - arm64
+  - s390x
+  - ppc64le
+
+env:
+  - BUILD=coverage    CXX=g++-7  CC=gcc-7  BUILD_TYPE=Debug
```
rrahn, line 30: this is neat 🤓

.travis.yml
```diff
@@ -138,10 +67,14 @@ before_script:
 
 script:
   - export SEQAN3_NO_VERSION_CHECK=1
+  - export CCACHE_COMPRESS=true
+  - export CCACHE_COMPRESSLEVEL=6
+  - export CCACHE_MAXSIZE=5G
```
rrahn, line 72: this is now enough or not important because it is a nightly build?

.travis.yml
```diff
@@ -138,10 +67,14 @@ before_script:
...
   - make -k -j2
   - |
-    if test coverage != "${BUILD}"; then
-      if [[ "${BUILD}" =~ ^(snippet)$ ]]; then
+    if [[ "${{ matrix.build }}" =~ ^(coverage)$ ]]; then
+      : ; else
```
rrahn, line 76: what happens here?

rrahn (commented):
looks great. Just some minor questions.

.travis.yml
eseiler: Yes, but unfortunately you cannot name it - naming only works for `include`.

But when you use include, no matrix is automatically generated (job X architecture).

.travis.yml
eseiler: We don't need to run ctest if we are building coverage - running the tests is already part of building the coverage, so I just do nothing (no-op operator `:`).

eseiler: It was `if test coverage != "${BUILD}"; then` before, but I didn't understand what it did until I removed it; I think not mixing different syntaxes probably helps understanding

.travis.yml
eseiler: When we compress it, it gets considerably smaller, I think the GitHub actions ccaches are around 30-60MiB, so we should have always compressed them; 5G are now more than enough

rrahn (approved):
thank you

PR #2091, opened 2020-09-02
Alignment/make free ends runtime decision

rrahn:
Removes the static configuration for aligned ends from the old alignment algorithm.

### DIFF
--- a/include/seqan3/alignment/pairwise/policy/find_optimum_policy.hpp
+++ b/include/seqan3/alignment/pairwise/policy/find_optimum_policy.hpp
@@ -14,6 +14,7 @@
 
 #include <type_traits>
 
+#include <seqan3/alignment/configuration/align_config_method.hpp>
 #include <seqan3/alignment/matrix/alignment_optimum.hpp>
 #include <seqan3/alignment/pairwise/detail/alignment_algorithm_state.hpp>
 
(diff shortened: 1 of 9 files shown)


rrahn:
Please note that the coverage test fail is only temporarily, since all these code paths will be removed once we switched to the new implementation.

wvandertoorn (changes requested):
Cool! Just one minor rephrasing suggestion.

wvandertoorn (approved):

include/seqan3/alignment/pairwise/policy/find_optimum_policy.hpp
```diff
@@ -66,6 +74,18 @@ class find_optimum_policy
...
+    //!\brief Initialises the policy with the configuration.
+    template <typename configuration_t>
+    find_optimum_policy(configuration_t const & config)
+    {
+        if constexpr (configuration_t::template exists<method_local_tag>())
+            test_every_cell =  true;
+
+        auto method_global_config = config.get_or(align_cfg::method_global{});
+        test_last_row_cell = method_global_config.free_end_gaps_sequence1_trailing;
+        test_last_column_cell = method_global_config.free_end_gaps_sequence2_trailing;
```
smehringer, line 67: aren't `test_every_cell` and `test_last_row/column_cell` contradictory ?
So why set both and not just in an else branch ?

smehringer (approved):
Just a question.
Since I'm not deeply involved in the alignment algorithm I can only judge the changes by the passing CI so LGTM :)

include/seqan3/alignment/pairwise/policy/find_optimum_policy.hpp
rrahn: Not necessarily. But the functions to check for a new score are implemented sucht that if every cell is tracked the explicit functions for tracking the last row/column won't check the same cell a second time.

PR #2093, opened 2020-09-03
[INFRA] seqan3/std/* header files MUST NOT include any seqan3 header

marehr:
Fixes https://github.com/seqan/product_backlog/issues/135

### DIFF
--- a/test/header/CMakeLists.txt
+++ b/test/header/CMakeLists.txt
@@ -82,6 +82,15 @@ TEST(${header_test_name_safe}) {}")
 #endif")
                 endif ()
 
+                # seqan3/std/* must not include platform.hpp (and therefore any other seqan3 header)
+                # See https://github.com/seqan/product_backlog/issues/135
+                if (header MATCHES "seqan3/std/")
+                    file (APPEND "${header_target_source}" "
+#ifdef SEQAN3_DOXYGEN_ONLY
+#error \"The standard header '${header}' file MUST NOT include any other seqan3 header (except for seqan3/contrib)\"
+#endif")
+                endif ()
+
                 # test whether seqan3 has the visibility bug on lower gcc versions
                 # https://github.com/seqan/seqan3/issues/1317
                 file (APPEND "${header_target_source}" "


test/header/CMakeLists.txt
```diff
@@ -82,6 +82,15 @@ TEST(${header_test_name_safe}) {}")
 #endif")
                 endif ()
 
+                # seqan3/std/* must not include platform.hpp (and therefore any other seqan3 header)
+                # See https://github.com/seqan/product_backlog/issues/135
+                if (header MATCHES "seqan3/std/")
+                    file (APPEND "${header_target_source}" "
+#ifdef SEQAN3_DOXYGEN_ONLY
+#error \"The standard header '${header}' file MUST NOT include any other seqan3 header (except for seqan3/contrib)\"
+#endif")
```
MitraDarja, line 91: If this is commented out, should we not remove it or add a comment when this can be included again?

MitraDarja (changes requested):
Sorry, for the late review. LGTM, just one question.

test/header/CMakeLists.txt
marehr: This is not commented out :) Notice the "" surrounding and `#` is actually C++ Code.

MitraDarja: Oh. okay. :sweat_smile:

MitraDarja (approved):

test/header/CMakeLists.txt
```diff
@@ -82,6 +82,15 @@ TEST(${header_test_name_safe}) {}")
 #endif")
                 endif ()
 
+                # seqan3/std/* must not include platform.hpp (and therefore any other seqan3 header)
+                # See https://github.com/seqan/product_backlog/issues/135
+                if (header MATCHES "seqan3/std/")
+                    file (APPEND "${header_target_source}" "
+#ifdef SEQAN3_DOXYGEN_ONLY
```
smehringer, line 89: I don't exactly know how this works in detecting seqan3 headers in /std 
but directly above, when checking platform.hpp it is 
`#ifndef SEQAN3_DOXYGEN_ONLY` (if`n`def)
and here it is 
`#ifdef SEQAN3_DOXYGEN_ONLY` (ifdef)

I guess its right since tests are passing but just double checking :)

eseiler: `platform.hpp` defines `SEQAN3_DOXYGEN_ONLY` and is included by all seqan3 headers.

If we include a seqan3 header, we want `SEQAN3_DOXYGEN_ONLY` to be defined, hence we error if it is not (`ifndef`).

If we include something from `seqan3/std`, we don't want any seqan3 header, i.e. no `SEQAN3_DOXYGEN_ONLY` and hence we error if it is defined (`ifdef`).

smehringer: ah thanks for the explanation :)

smehringer (approved):

PR #2106, opened 2020-09-08
Optimise alignment/part6 enable banded end position

rrahn:
Enables the banded global alignment using end-positions. This is a first part of making the banded alignment vectorised.
See seqan/product_backlog#114

### DIFF
--- a/include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
+++ b/include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
@@ -88,6 +88,10 @@ public:
             size_t const sequence1_size = std::ranges::distance(get<0>(sequence_pair));
             size_t const sequence2_size = std::ranges::distance(get<1>(sequence_pair));
 
+            // Initialise the cell updater with the dimensions of the regular matrix.
+            this->compare_and_set_optimum.set_target_indices(row_index_type{sequence2_size},
+                                                             column_index_type{sequence1_size});
+
             auto && [alignment_matrix, index_matrix] = this->acquire_matrices(sequence1_size,
                                                                               sequence2_size,
                                                                               this->lowest_viable_score());
--- a/include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
+++ b/include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
@@ -228,7 +228,7 @@ protected:
     }
 
     /*!\brief Computes a column of the band that does not start in the first row of the alignment matrix.
-     * \tparam alignment_column_t The type of the alignment column; must model std::ranges::input_range.
+     * \tparam alignment_column_t The type of the alignment column; must model std::ranges::forward_range.
      * \tparam cell_index_column_t The type of the indexed column; must model std::ranges::input_range.
      * \tparam sequence1_value_t The value type of sequence1; must model seqan3::semialphabet.
      * \tparam sequence2_t The type of the second sequence; must model std::ranges::input_range.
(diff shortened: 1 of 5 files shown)


MitraDarja (changes requested):
LGTM, just some small things. :)

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
```diff
@@ -228,7 +231,7 @@ class pairwise_alignment_algorithm_banded :
     }
 
     /*!\brief Computes a column of the band that does not start in the first row of the alignment matrix.
-     * \tparam alignment_column_t The type of the alignment column; must model std::ranges::input_range.
+     * \tparam alignment_column_t The type of the alignment column; must model std::ranges::forward_range.
```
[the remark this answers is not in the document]
rrahn: because I need to copy the iterator and this only guaranteed for forward_iterators.

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
```diff
@@ -66,6 +66,70 @@ struct max_score_updater
...
 
+/*!\brief A function object that compares and possibly updates the alignment optimum with the current cell.
+ * \ingroup pairwise_alignment
+ *
+ * \details
+ *
+ * Updates the current alignment optimum with the new score and the respective coordinate if the new score
+ * compares greater or equal to the score of the current optimum and the cell is either in the target row or column.
+ * For the banded alignment this additional check is needed to only track the cells that are really in the last
+ * row of the global alignment matrix and not at the end of the column that does not reach the last alignment row yet.
```
rrahn, line 77:
```suggestion
 * For the banded alignment this additional check is needed because the band might not reach the last row of the matrix for all columns. 
 * Thus, the last cell of a column will only be checked if the corresponding coordinate points to the last row of the matrix.
```

MitraDarja (approved):
LGTM. Thanks for the clearification.:)

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
```diff
@@ -276,7 +281,8 @@ class pairwise_alignment_algorithm_banded :
 
         for (auto && sequence2_value : sequence2 | views::drop(1))
         {
-            auto cell = *alignment_column_it;
+            last_alignment_column_it = alignment_column_it;
+            auto cell = *last_alignment_column_it;
```
smehringer, line 285: Can you explain this change to me? What are you tracking here since it seems `last_alignment_column_it` is always overridden with the current column iterator, so they are always the same no?

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
```diff
@@ -66,6 +66,70 @@ struct max_score_updater
...
 
+/*!\brief A function object that compares and possibly updates the alignment optimum with the current cell.
+ * \ingroup pairwise_alignment
+ *
+ * \details
+ *
+ * Updates the current alignment optimum with the new score and the respective coordinate if the new score
+* is greater or equal to the score of the current optimum and the cell is either in the target row or column.
+ * For the banded alignment this additional check is needed because the band might not reach the last row of the matrix for all columns. 
+ * Thus, the last cell of a column will only be checked if the corresponding coordinate points to the last row of the matrix.
```
smehringer, line 77: Sorry I don't get it.. So from the code I understand that the score is only updated if the input cell *is in the target_row / target_column*. But the documentation here and the naming `max_score_banded_updater` suggests that this function is *in general* updating the score. But what about the inner matrix cells? They will not be updated here. Are there multiple "score_updaters"?

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
```diff
@@ -66,6 +66,70 @@ struct max_score_updater
...
+                                      (current_score >= optimal_score);
+        optimal_score = (is_better_score) ? std::move(current_score) : optimal_score;
+        optimal_coordinate = (is_better_score) ? std::move(current_coordinate) : optimal_coordinate;
+    }
+
+    /*!\brief Sets the target index for the last row and column of the matrix.
+     * \param row_index The target index of the last row.
+     * \param col_index The target index of the last column.
+     */
+    void set_target_indices(size_t row_index, size_t col_index) noexcept
```
smehringer, line 126: Don't you have strong types for columns and rows already that you can use here?

smehringer (commented):
Only some questions

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
rrahn: Not quite. I think I will add a small example in the documentation. 
The problem is based on the internal implementation using a single column score matrix. As long as the band touches the first row of the matrix the relative column index points to the same cell inside of this column as the column before. When the band starts to move downwards the relative index to the previous column changes. This is reflected with the special function for the banded alignment. But before I try to explain it here I try to make a picture of it in the documentation.

include/seqan3/alignment/pairwise/detail/policy_optimum_tracker.hpp
rrahn: Yes, there can be different updater functions. These are basically binary functions that compare and update the current score. So there is also another one that is used by regular global and the local alignment and the correct update function is set by the alignment configuration. I will also try to make a little picture to make it more clear what is meant.

rrahn: And later for the local alignment we can use the regular max updater because we look at every cell anyway.

rrahn:
@smehringer I tried to add more verbose documentation with graphical support. I hope this helps understanding the implementation. Please let me know if I should rephrase something or add something that is not clear. Thank you.

smehringer (approved):
Thanks a lot for the detailed documentation! It is much clearer now :)

PR #2107, opened 2020-09-09
[INFRA] Do not deploy docs if faulty

eseiler:
The ctest command has to build the documentation to check for warnings, i.e. we do not need to build it beforehand.


marehr (approved):

rrahn (approved):
thanks

PR #2109, opened 2020-09-10
[FIX] Timeout in debug nightlies

eseiler:
For example: https://cdash.seqan.de/viewTest.php?onlyfailed&buildid=50873


eseiler:
Same needs to be added for the recently introduced simd protein benchmark

simonsasse (approved):
👌

smehringer (approved):

PR #2111, opened 2020-09-10
[INFRA] Add codecov.yml

eseiler:
Resolves https://github.com/seqan/product_backlog/issues/205

My goal for now is:
* Complain in PRs about any coverage drop
* Do not complain in master if > 97%
* Make codecov independent of CI (do not fail just because CI failed, do not wait for CI to finish to post the coverage results)

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,13 @@
+# See https://docs.codecov.io/docs/codecovyml-reference
+codecov:
+  require_ci_to_pass: no
+  notify:
+    wait_for_ci: no
+
+coverage:
+  status:
+    project:
+      default:
+        if_ci_failed: success
+        target: auto
+        threshold: 0.01%


smehringer:
Now codecov/project/master is shown in the list of PRs. Has this been this way before?

eseiler:
> Now codecov/project/master is shown in the list of PRs. Has this been this way before?

No, but it does not seem possible to have different behaviour for different branches.
I asked at their support forum, but I didn't get an answer yet.

Maybe we just set the threshold to 0.01% to avoid failing coverage when removing x lines and hence getting x less hits.

eseiler:
I didn't get a reply yet, so I would just got for the 0.01 threshold and adapt it later, if the things we want are possible.

smehringer:
Sorry, yes I think that's fine I think

.codecov.yml
```diff
@@ -0,0 +1,12 @@
+codecov:
```
marehr, line 1:
```suggestion
# see documentation at ...
codecov:
```

marehr: it would be always nice to have some initial clue where to look :)

marehr (approved):

.codecov.yml
eseiler: done

smehringer (approved):

PR #2112, opened 2020-09-11
[TEST] Force compiler to actually execute the search algorithm.

smehringer:
On my personal laptop (Linux, Ubuntu 18, GCC 7) the search algorithm was not executed in the benchmarks.

### DIFF
--- a/test/performance/search/search_benchmark.cpp
+++ b/test/performance/search/search_benchmark.cpp
@@ -166,8 +166,9 @@ void unidirectional_search_all_collection(benchmark::State & state, options && o
     for (auto _ : state)
     {
         auto results = search(reads, index, cfg);
-        benchmark::DoNotOptimize(sum += std::ranges::distance(results));
+        sum += std::ranges::distance(results);
     }
+    benchmark::DoNotOptimize(sum);
 }
 
 //============================================================================
@@ -191,8 +192,9 @@ void unidirectional_search_all(benchmark::State & state, options && o)
     for (auto _ : state)
     {
         auto results = search(reads, index, cfg);
-        benchmark::DoNotOptimize(sum += std::ranges::distance(results));
+        sum += std::ranges::distance(results);
     }
+    benchmark::DoNotOptimize(sum);
 }
 
 //============================================================================
@@ -216,8 +218,9 @@ void bidirectional_search_all(benchmark::State & state, options && o)
     for (auto _ : state)
     {
         auto results = search(reads, index, cfg);
-        benchmark::DoNotOptimize(sum += std::ranges::distance(results));
+        sum += std::ranges::distance(results);
     }
+    benchmark::DoNotOptimize(sum);
 }
 
 //============================================================================
@@ -242,8 +245,9 @@ void unidirectional_search_stratified(benchmark::State & state, options && o)
     for (auto _ : state)
     {
         auto results = search(reads, index, cfg);
-        benchmark::DoNotOptimize(sum += std::ranges::distance(results));
+        sum += std::ranges::distance(results);
     }
+    benchmark::DoNotOptimize(sum);
 }
 
 //============================================================================
@@ -268,8 +272,9 @@ void bidirectional_search_stratified(benchmark::State & state, options && o)
     for (auto _ : state)
     {
         auto results = search(reads, index, cfg);
-        benchmark::DoNotOptimize(sum += std::ranges::distance(results));
+        sum += std::ranges::distance(results);
     }
+    benchmark::DoNotOptimize(sum);
 }
 
 BENCHMARK_CAPTURE(unidirectional_search_all_collection, highErrorReadsSearch0,
--- a/test/performance/search/search_benchmark.cpp
+++ b/test/performance/search/search_benchmark.cpp
@@ -144,6 +144,9 @@ std::vector<alphabet_t> generate_repeating_sequence(size_t const template_length
 //  undirectional; trivial_search, collection, dna4, all-mapping
 //============================================================================
 
+// Note: We force the actual computation of the search() by going through the lazy algorithm result range (input_range)
+//       using std::ranges::distance within the for loop.
+
 void unidirectional_search_all_collection(benchmark::State & state, options && o)
 {
     size_t set_size = 10;
--- a/test/performance/search/search_benchmark.cpp
+++ b/test/performance/search/search_benchmark.cpp
@@ -162,8 +162,12 @@ void unidirectional_search_all_collection(benchmark::State & state, options && o
     seqan3::fm_index index{collection};
     seqan3::configuration cfg = seqan3::search_cfg::max_error_total{seqan3::search_cfg::error_count{o.searched_errors}};
 
+    size_t sum{};
     for (auto _ : state)
+    {
         auto results = search(reads, index, cfg);
+        benchmark::DoNotOptimize(sum += std::ranges::distance(results));
+    }
 }
 
 //============================================================================
@@ -183,8 +187,12 @@ void unidirectional_search_all(benchmark::State & state, options && o)
                                                                   o.prob_deletion, o.stddev);
     seqan3::configuration cfg = seqan3::search_cfg::max_error_total{seqan3::search_cfg::error_count{o.searched_errors}};
 
+    size_t sum{};
     for (auto _ : state)
+    {
         auto results = search(reads, index, cfg);
+        benchmark::DoNotOptimize(sum += std::ranges::distance(results));
+    }
 }
 
 //============================================================================
@@ -204,8 +212,12 @@ void bidirectional_search_all(benchmark::State & state, options && o)
                                                                   o.prob_deletion, o.stddev);
     seqan3::configuration cfg = seqan3::search_cfg::max_error_total{seqan3::search_cfg::error_count{o.searched_errors}};
 
+    size_t sum{};
     for (auto _ : state)
+    {
         auto results = search(reads, index, cfg);
+        benchmark::DoNotOptimize(sum += std::ranges::distance(results));
+    }
 }
 
 //============================================================================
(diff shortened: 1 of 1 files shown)


test/performance/search/search_benchmark.cpp
```diff
@@ -162,8 +162,12 @@ void unidirectional_search_all_collection(benchmark::State & state, options && o
     seqan3::fm_index index{collection};
     seqan3::configuration cfg = seqan3::search_cfg::max_error_total{seqan3::search_cfg::error_count{o.searched_errors}};
 
+    size_t sum{};
     for (auto _ : state)
+    {
         auto results = search(reads, index, cfg);
+        benchmark::DoNotOptimize(sum += std::ranges::distance(results));
+    }
 }
```
eseiler, line 171: I think we should either do

```cpp
    size_t sum{};
    for (auto _ : state)
    {
        auto results = search(reads, index, cfg);
        sum += std::ranges::distance(results);
    }
    benchmark::DoNotOptimize(sum);
```

or 
```cpp
    for (auto _ : state)
    {
        benchmark::DoNotOptimize(search(reads, index, cfg));
    }
```

Do these make a difference for you?

joergi-w: I assume that the sum is only needed to use the results, so the compiler does not discard the whole search?
Please add a comment on this or use the second suggestion (if it works).
Maybe you can also write `benchmark::DoNotOptimize(results);`, which is clearer than using the sum?

joergi-w (changes requested):

test/performance/search/search_benchmark.cpp
eseiler: Yes, `benchmark::DoNotOptimize(results);` is also fine. I think the main/first question is whether we want `benchmark::DoNotOptimize` inside or outside the loop.

It's probably more clear if it is outside the loop and may cause less worrying about whether this is now measured currently or not.

Generally, we can kinda do whatever we want as long as we use the macro:

> `DoNotOptimize(<expr>)` forces the result of `<expr>` to be stored in either memory or a register. For GNU based compilers it acts as read/write barrier for global memory. More specifically, it forces the compiler to flush pending writes to memory and reload any other values as necessary.
> 
> Note that `DoNotOptimize(<expr>)` does not prevent optimizations on `<expr>` in any way. `<expr>` may even be removed entirely when the result is already known.

smehringer:
Phew, results look a little weird.. 

So I checked whether there is a difference between all suggestions (with a subset of the search benchmark of all categories):

1. **no using benchmark::DoNotOptimize at all** (current master)
```cpp
    for (auto _ : state)
        search(reads, index, cfg);
```
```
unidirectional_search_all_collection/highErrorReadsSearch0            87.5 ns         87.5 ns      8286790
unidirectional_search_all_collection/highErrorReadsSearch1            87.2 ns         87.2 ns      7770911
unidirectional_search_all/lowErrorReadsSearch3                        86.3 ns         86.3 ns      8180261
unidirectional_search_all/highErrorReadsSearch0                       90.9 ns         90.9 ns      7160684
bidirectional_search_all/lowErrorReadsSearch3                          114 ns          114 ns      6148005
bidirectional_search_all/highErrorReadsSearch0                         113 ns          113 ns      5986505
unidirectional_search_stratified/lowErrorReadsSearch3Strata0Rep       90.1 ns         90.1 ns      7899309
unidirectional_search_stratified/lowErrorReadsSearch3Strata1Rep       89.8 ns         89.7 ns      7770068
bidirectional_search_stratified/lowErrorReadsSearch3Strata0Rep         116 ns          116 ns      5964146
bidirectional_search_stratified/lowErrorReadsSearch3Strata1Rep         121 ns          121 ns      5973768
```

2. **Use benchmark::DoNotOptimize in loop for sum computation** (current state of PR)
```cpp
    size_t sum{};
    for (auto _ : state)
    {
        auto results = search(reads, index, cfg);
        benchmark::DoNotOptimize(sum += std::ranges::distance(results));
    }
```
```
unidirectional_search_all_collection/highErrorReadsSearch0          195334 ns       195321 ns         3539
unidirectional_search_all_collection/highErrorReadsSearch1         2894684 ns      2894647 ns          245
unidirectional_search_all/lowErrorReadsSearch3                   189621543 ns    189594268 ns            4
unidirectional_search_all/highErrorReadsSearch0                      81632 ns        81621 ns         8557
bidirectional_search_all/lowErrorReadsSearch3                     57839027 ns     57832315 ns           12
bidirectional_search_all/highErrorReadsSearch0                       85467 ns        85463 ns         8182
unidirectional_search_stratified/lowErrorReadsSearch3Strata0Rep    4461845 ns      4461657 ns          152
unidirectional_search_stratified/lowErrorReadsSearch3Strata1Rep   17408807 ns     17408447 ns           39
bidirectional_search_stratified/lowErrorReadsSearch3Strata0Rep     1294833 ns      1294763 ns          528
bidirectional_search_stratified/lowErrorReadsSearch3Strata1Rep     6359830 ns      6359122 ns          109
```
3. **Use benchmark::DoNotOptimize outside loop for sum** (Enrico suggestion 1)
```cpp
    size_t sum{};
    for (auto _ : state)
    {
        auto results = search(reads, index, cfg)
        sum += std::ranges::distance(results);
    }
    benchmark::DoNotOptimize(sum);
```
```
unidirectional_search_all_collection/highErrorReadsSearch0          198330 ns       198299 ns         3634
unidirectional_search_all_collection/highErrorReadsSearch1         2862708 ns      2862535 ns          233
unidirectional_search_all/lowErrorReadsSearch3                   191820975 ns    191787530 ns            4
unidirectional_search_all/highErrorReadsSearch0                      82796 ns        82785 ns         8727
bidirectional_search_all/lowErrorReadsSearch3                     59591145 ns     59585919 ns           12
bidirectional_search_all/highErrorReadsSearch0                       86762 ns        86703 ns         8160
unidirectional_search_stratified/lowErrorReadsSearch3Strata0Rep    4549947 ns      4549655 ns          154
unidirectional_search_stratified/lowErrorReadsSearch3Strata1Rep   17621813 ns     17609713 ns           39
bidirectional_search_stratified/lowErrorReadsSearch3Strata0Rep     1336325 ns      1336260 ns          536
bidirectional_search_stratified/lowErrorReadsSearch3Strata1Rep     6275867 ns      6271669 ns          111
```
4. **Use benchmark::DoNotOptimize inside loop for search** (Enrico suggestion 2)
```cpp
    for (auto _ : state)
    {
        benchmark::DoNotOptimize(search(reads, index, cfg));
    }
```
```
unidirectional_search_all_collection/highErrorReadsSearch0            91.0 ns         91.0 ns      7995766
unidirectional_search_all_collection/highErrorReadsSearch1            95.3 ns         95.3 ns      7555582
unidirectional_search_all/lowErrorReadsSearch3                        87.4 ns         87.4 ns      8030201
unidirectional_search_all/highErrorReadsSearch0                       90.8 ns         90.8 ns      7699700
bidirectional_search_all/lowErrorReadsSearch3                          105 ns          105 ns      6438030
bidirectional_search_all/highErrorReadsSearch0                         110 ns          110 ns      6571698
unidirectional_search_stratified/lowErrorReadsSearch3Strata0Rep       88.2 ns         88.2 ns      7603717
unidirectional_search_stratified/lowErrorReadsSearch3Strata1Rep       85.6 ns         85.6 ns      6629886
bidirectional_search_stratified/lowErrorReadsSearch3Strata0Rep         109 ns          109 ns      6381328
bidirectional_search_stratified/lowErrorReadsSearch3Strata1Rep         108 ns          108 ns      6332604
```

So it seems that 2. and 3. do not make a difference but 4. is weird..
@eseiler Do you have any idea why? 
I guess otherwise I would go with your first suggestion. And I thought the second suggestion should not be used anyway because "we want to avoid any possible impact on actual code optimisation".

eseiler:
> So it seems that 2. and 3. do not make a difference but 4. is weird..
> @eseiler Do you have any idea why?

We would have to capture the return value...
Then maybe just go with
```cpp
    size_t sum{};
    for (auto _ : state)
    {
        auto results = search(reads, index, cfg)
        sum += std::ranges::distance(results);
    }
    benchmark::DoNotOptimize(sum);
```

joergi-w (approved):
Okay, thanks!

eseiler:
Can you make a backlog ticket that we document somewhere how we want to do the `DoNotOptimize` stuff and at some point should adjust the benchmarks accordingly?

smehringer:
> Can you make a backlog ticket that we document somewhere how we want to do the DoNotOptimize stuff and at some point should adjust the benchmarks accordingly?

Done: https://github.com/seqan/product_backlog/issues/214

test/performance/search/search_benchmark.cpp
```diff
@@ -162,8 +162,13 @@ void unidirectional_search_all_collection(benchmark::State & state, options && o
     seqan3::fm_index index{collection};
     seqan3::configuration cfg = seqan3::search_cfg::max_error_total{seqan3::search_cfg::error_count{o.searched_errors}};
 
+    size_t sum{};
     for (auto _ : state)
+    {
         auto results = search(reads, index, cfg);
+        sum += std::ranges::distance(results);
```
rrahn, line 169:
```suggestion
        sum += std::ranges::distance(results); // Forces the actual computation of the search through the lazy algorithm result range.
```

rrahn: Ok, this is at every place. Could you just put that in the top of the benchmark file. So one can read up on why we do this distance computation? Then this comment must not be repeated for every benchmark

rrahn (changes requested):
In general looks good. I would just add a comment explaining why std::ranges::distance is used to measure the performance.

rrahn (approved):

rrahn:
I would wait for the release branch merge though. Just in case this destroys something again.

PR #2113, opened 2020-09-11
[MISC] Remove unnecessary second transform from lazy_locate.

smehringer:
I found another patch I did when investigating the search performance. I think it didn't have an significant impact but either way it is unnecessary to use two transform views if using one is equivalent.


eseiler (approved):

rrahn (approved):
yes, looks good.

Issue #2115, opened 2020-09-14
Different (wrong) results for FM index search with qualified sequences


tloka:
## Platform

  - SeqAn version: 3.0.2
  - Operating system: Ubuntu
  - Compiler: GCC 9.2.1

## Description
When searching a qualified sequence in an FM index using seqan3::search, I get wrong results.  
No idea if this behavior is expected as the ranks are different than with the non-qualified alphabet I create the index with. However, for me the behavior was unexpected, is highly non-intuitive and can quickly lead to wrong behavior when not being aware of this.
See below for a minimal example.

### How to repeat the problem

Code:
```cpp
#include <seqan3/alphabet/nucleotide/dna5.hpp>
#include <seqan3/alphabet/quality/phred42.hpp>
#include <seqan3/alphabet/quality/qualified.hpp>
#include <seqan3/core/debug_stream.hpp>
#include <seqan3/search/fm_index/fm_index.hpp>
#include <seqan3/search/search.hpp>

using seqan3::operator""_dna5;

int main()
{
   // Build Index
   std::vector<seqan3::dna5_vector> genomes{"ACAG"_dna5};
   seqan3::fm_index index{genomes};

   // Create qualified query
   std::vector<seqan3::qualified<seqan3::dna5, seqan3::phred42>> dnaq_query{{'A'_dna5, seqan3::phred42{0}},
                                                                            {'C'_dna5, seqan3::phred42{15}},
                                                                            {'G'_dna5, seqan3::phred42{30}},
                                                                            {'G'_dna5, seqan3::phred42{41}}};

   // Create non-qualified query
   std::vector<seqan3::dna5> dna5_query = "ACGG"_dna5;

   // Search config: allow 1 substitution
   seqan3::configuration cfg = seqan3::search_cfg::max_error_total{seqan3::search_cfg::error_count{1}} |
                               seqan3::search_cfg::max_error_substitution{seqan3::search_cfg::error_count{1}} |
                               seqan3::search_cfg::max_error_insertion{seqan3::search_cfg::error_count{0}} |
                               seqan3::search_cfg::max_error_deletion{seqan3::search_cfg::error_count{0}} |
                               seqan3::search_cfg::hit_all_best;

   // Search both queries in the index
   auto dnaq_search_hits = seqan3::search(dnaq_query, index, cfg);
   auto dna5_search_hits = seqan3::search(dna5_query, index, cfg);

   // Create output
[...]

PR #2116, opened 2020-09-15
[FIX] Wrong ranks in search algorithm

eseiler:
Resolves https://github.com/seqan/seqan3/issues/2115

The search algorithm uses the ranks of the given query. If the alphabet type of the query does not equal the alphabet type of the index, hits may be missed.
The index already converts the query, i.e. ,for the given example, `extend_right` on the cursor correctly returned true, but the search did not find the hits, because it used the ranks of the qualified query instead of the normal, unqualified query that the index uses.

The second commit adds a test for debug_streaming, as there was none before and coverage dropped because of this.

### DIFF
--- a/include/seqan3/search/search.hpp
+++ b/include/seqan3/search/search.hpp
@@ -15,8 +15,11 @@
 #include <seqan3/std/algorithm>
 #include <seqan3/std/ranges>
 
+#include <seqan3/core/algorithm/algorithm_result_generator_range.hpp>
 #include <seqan3/core/algorithm/configuration.hpp>
 #include <seqan3/core/algorithm/detail/algorithm_executor_blocking.hpp>
+#include <seqan3/range/views/convert.hpp>
+#include <seqan3/range/views/deep.hpp>
 #include <seqan3/range/views/persist.hpp>
 #include <seqan3/range/views/zip.hpp>
 #include <seqan3/search/configuration/default_configuration.hpp>
@@ -24,7 +27,6 @@
 #include <seqan3/search/configuration/parallel.hpp>
 #include <seqan3/search/detail/search_configurator.hpp>
 #include <seqan3/search/detail/search_traits.hpp>
-#include <seqan3/core/algorithm/algorithm_result_generator_range.hpp>
 
 namespace seqan3::detail
 {
@@ -100,7 +102,8 @@ template <fm_index_specialisation index_t,
           std::ranges::forward_range queries_t,
           typename configuration_t = decltype(search_cfg::default_configuration)>
 //!\cond
-    requires std::ranges::forward_range<std::ranges::range_reference_t<queries_t>>
+    requires std::ranges::forward_range<std::ranges::range_reference_t<queries_t>> &&
+             std::same_as<range_innermost_value_t<queries_t>, typename index_t::alphabet_type>
 //!\endcond
 inline auto search(queries_t && queries,
                    index_t const & index,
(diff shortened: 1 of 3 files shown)


include/seqan3/search/search.hpp
```diff
@@ -165,6 +168,26 @@ inline auto search(queries_t && queries,
...
+             (!std::same_as<range_innermost_value_t<queries_t>, typename index_t::alphabet_type>)
+inline auto search(queries_t && queries,
+                   index_t const & index,
+                   configuration_t const & cfg = search_cfg::default_configuration)
+{
+    static_assert(std::convertible_to<range_innermost_value_t<queries_t>, typename index_t::alphabet_type>,
+                  "The alphabet of the text collection must be convertible to the alphabet of the index.");
+
+    if constexpr (range_dimension_v<queries_t> == 2u)
+        return search(queries | views::deep{views::convert<typename index_t::alphabet_type>}, index, cfg);
```
MitraDarja, line 186: Why do we need the deep function here?

test/unit/search/search_test.cpp
```diff
@@ -483,6 +483,38 @@ TYPED_TEST(search_test, parallel_without_parameter)
...
+{
+    std::ostringstream oss;
+    seqan3::debug_stream_type stream{oss};
+    stream << search("TAC"_dna4, this->index);
+    EXPECT_EQ(oss.str(), "[<query_id:0, reference_id:0, reference_pos:3>"
+                         ",<query_id:0, reference_id:0, reference_pos:7>]");
+}
+
+// https://github.com/seqan/seqan3/issues/2115
+TYPED_TEST(search_test, issue_2115)
```
MitraDarja, line 496: Isn't that the exact same test as in search_collection_test? Why do we need the test twice?

MitraDarja (changes requested):
LGTM, but I have some questions to understand the code better. :sweat_smile:

include/seqan3/search/search.hpp
eseiler: Because we can pass multiple queries, i.e. a vector of queries: `{"ACGT"_dna4, "GGTA"_dna4}`
If it's a range of queries, we need to convert each query (deep).
If it's a single query, we just convert the query.

test/unit/search/search_test.cpp
eseiler: Yes, but it needs to work for both indices built over collections and single texts.

MitraDarja (approved):
Thanks for the clearification.:)

include/seqan3/search/search.hpp
```diff
@@ -165,6 +168,26 @@ inline auto search(queries_t && queries,
 }
 
 //!\cond DEV
+// Convert query sequence if it does not match the alphabet type of the index.
+//!\overload
+template <fm_index_specialisation index_t,
+          std::ranges::forward_range queries_t,
+          typename configuration_t = decltype(search_cfg::default_configuration)>
+    requires std::ranges::forward_range<std::ranges::range_reference_t<queries_t>> &&
+             (!std::same_as<range_innermost_value_t<queries_t>, typename index_t::alphabet_type>)
```
smehringer, line 177: not the same but `convertible to` should be required?

smehringer (changes requested):
only one thing, otherwise LGTM :+1:

include/seqan3/search/search.hpp
eseiler: No, it must not be the same, because that means we need to convert it such that it is the same.
There's a static assert for the convertible thingie. but it is not needed for the overload resultion.

smehringer: ah ok :+1:

smehringer (approved):

PR #2120, opened 2020-09-16
[FIX] address issue in banded semi global alignment.

rrahn:
Fixes seqan/product_backlog#213 ... the issue with accessing invalid memory in the banded semi global alignment.
The problem was that the upper diagonal of the band ended before the actual end of the first sequence. 
Within the current implementation this would then try to access elements from the second sequence that are behind the end. To fix this, the first sequence is trimmed, such that the last cell of the band also represents the last cell in the "trimmed" matrix and then reaches the end of the first sequence to stop the alignment.

### DIFF
--- a/include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
+++ b/include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
@@ -85,18 +85,24 @@ public:
 
         for (auto && [sequence_pair, idx] : indexed_sequence_pairs)
         {
-            size_t const sequence1_size = std::ranges::distance(get<0>(sequence_pair));
+            size_t sequence1_size = std::ranges::distance(get<0>(sequence_pair));
             size_t const sequence2_size = std::ranges::distance(get<1>(sequence_pair));
-
-            // Initialise the cell updater with the dimensions of the regular matrix.
-            this->compare_and_set_optimum.set_target_indices(row_index_type{sequence2_size},
-                                                             column_index_type{sequence1_size});
 
             auto && [alignment_matrix, index_matrix] = this->acquire_matrices(sequence1_size,
                                                                               sequence2_size,
                                                                               this->lowest_viable_score());
 
-            compute_matrix(get<0>(sequence_pair), get<1>(sequence_pair), alignment_matrix, index_matrix);
+            // Initialise the cell updater with the dimensions of the regular matrix.
+            this->compare_and_set_optimum.set_target_indices(row_index_type{sequence2_size},
+                                                             column_index_type{sequence1_size});
+
+            // Shrink the first sequence if the band ends before its actual end.
+            sequence1_size = std::min(sequence1_size, this->upper_diagonal + sequence2_size);
+
+            compute_matrix(get<0>(sequence_pair) | views::take(sequence1_size),
+                           get<1>(sequence_pair),
+                           alignment_matrix,
+                           index_matrix);
             this->make_result_and_invoke(std::forward<decltype(sequence_pair)>(sequence_pair),
                                          std::move(idx),
                                          this->optimal_score,


include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
```diff
@@ -249,38 +255,6 @@ class pairwise_alignment_algorithm_banded :
      * phase the first cell of the column is computed and in the iteration phase all remaining cells are computed.
      * In the final phase the last cell is possibly evaluated for a new alignment optimum.
      * Note that the length of `sequence2` determines the size of the column.
-     *
```
simonsasse, line 252: Why is this doc part deleted ? 🤔

simonsasse (commented):
Looks good, just wondering why you deleted the documentation? 😊

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
rrahn: Not sure. I applied an older stashed commit and this might have removed this. I revert this. Thanks for spotting.

simonsasse (approved):
Thanks 👌

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
```diff
@@ -85,18 +85,24 @@ class pairwise_alignment_algorithm_banded :
...
+
             // Initialise the cell updater with the dimensions of the regular matrix.
             this->compare_and_set_optimum.set_target_indices(row_index_type{sequence2_size},
                                                              column_index_type{sequence1_size});
 
-            auto && [alignment_matrix, index_matrix] = this->acquire_matrices(sequence1_size,
-                                                                              sequence2_size,
-                                                                              this->lowest_viable_score());
+            // Shrink the first sequence if the band ends before its actual end.
+            sequence1_size = std::min(sequence1_size, this->upper_diagonal + sequence2_size);
```
smehringer, line 100: Is it fine to first `set_target_indices` (line 96) of the score updater and **afterwards** changing the size of `sequence1`?

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
```diff
@@ -85,18 +85,24 @@ class pairwise_alignment_algorithm_banded :
...
 
-            compute_matrix(get<0>(sequence_pair), get<1>(sequence_pair), alignment_matrix, index_matrix);
+            compute_matrix(get<0>(sequence_pair) | views::take(sequence1_size),
+                           get<1>(sequence_pair),
```
smehringer, line 103: So this can only happen with the upper_diagonal and sequence1 and not with the lower_diagonal and sequence2?

smehringer (approved):
Just two question that I'm sure you considered but that just popped up while reviewing because I don't have a good overview of the whole algorithm. So just double checking :)

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
rrahn: yes, because in the acquire matrix algorithm we have to check if the band is valid, which we can only do on the original size. That means we might allocate a little to much memory, but that is usually not a problem since for the score we do not allocate the whole matrix and if we have the trace then I expect, that the difference is not by orders of magnitude.

include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
rrahn: yes. The point is that we have a column based approach such that the last column might be different in the band. The last row might be different as well, but then the column is limited by the band. This is already part of the algorithm but happens inside of the function compute_matrix

PR #2125, opened 2020-09-18
[MISC] Increase search coverage to 100%

eseiler:
We have some branches in the search algorithm that are never tested...

This adds two tests that cover the last few lines - after this, the search module has 100% coverage (and currently would be the only one to achieve this....).

One test just uses 4 errors instead of 3. This activates all the cases where a dynamic search scheme is used (up to and including 3 errors we know the optimal search scheme at compile time, i.e. we never use dynamic search schemes for errors <= 3).

Another one adds a test that covers two `if`s very deep inside the algorithm that handle a case where you do deletions at the end of the block after having done substitutions/insertions; I just enumerated over all error type counts from 1 to 5 and over all queries of length 5 and took one that hit the lines.


wvandertoorn (approved):
Cool! Neeaaaaarly at 100%

rrahn (approved):

PR #2175, opened 2020-09-28
[TEST] Add throw test for slice view to increase code coverage

MitraDarja:

### DIFF
--- a/test/unit/range/views/view_slice_test.cpp
+++ b/test/unit/range/views/view_slice_test.cpp
@@ -129,7 +129,7 @@ TEST(view_slice, underlying_is_shorter)
     EXPECT_EQ("oob", v);
 }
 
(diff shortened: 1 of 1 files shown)


Irallia (approved):
LGFM, thank you!

test/unit/range/views/view_slice_test.cpp
```diff
@@ -129,6 +129,12 @@ TEST(view_slice, underlying_is_shorter)
     EXPECT_EQ("oob", v);
 }
 
+TEST(view_slice, throw)
```
smehringer, line 132: `throw` isn't a very meaningful name. What are you testing here?

smehringer (changes requested):

smehringer (approved):

PR #2176, opened 2020-09-28
[TEST] Add throw test for translate single view to increase code cove…

MitraDarja:

I would have added tests for the throws in lines 318 and 391, but if I understand the code correctly they can never be reached because the throw in line 212 will always throw first. Correct?
(Line number from translate.hpp, see: https://codecov.io/gh/seqan/seqan3/src/master/include/seqan3/range/views/translate.hpp)

### DIFF
--- a/test/unit/range/views/view_translate_test.cpp
+++ b/test/unit/range/views/view_translate_test.cpp
@@ -256,6 +256,101 @@ TYPED_TEST(nucleotide, view_translate)
         EXPECT_RANGE_EQ(v11[i], cmp6[i]);
 }
 
+TYPED_TEST(nucleotide, view_translate_const)
+{
+    std::string const in{"ACGTACGTACGTA"};
+    std::vector<TypeParam> vec = in | seqan3::views::char_to<TypeParam> | seqan3::views::to<std::vector>;
+    std::vector<std::vector<seqan3::aa27> > cmp1{{"TYVR"_aa27}};
+    std::vector<std::vector<seqan3::aa27> > cmp2{{"TYVR"_aa27}, {"YVRT"_aa27}};
+    std::vector<std::vector<seqan3::aa27> > cmp3{{"TYVR"_aa27}, {"RTYV"_aa27}, {"VRT"_aa27}};
+    std::vector<std::vector<seqan3::aa27> > cmp4{{"TYVR"_aa27}, {"RTYV"_aa27}, {"VRT"_aa27}, {"YVRT"_aa27},
+                                                 {"TYVR"_aa27}, {"RTY"_aa27}};
+    std::vector<std::vector<seqan3::aa27> > cmp5{{"TYVR"_aa27}, {"VRT"_aa27}};
+    std::vector<std::vector<seqan3::aa27> > cmp6{{"CMHA"_aa27}, {"MHAC"_aa27}};
+    std::vector<std::vector<seqan3::aa27> > cmp7{{"CMHA"_aa27}};
+
+    // default parameter translation_frames
+    auto const v1 = vec | seqan3::views::translate;
+    // == [[T,Y,V,R],[R,T,Y,V],[V,R,T],[Y,V,R,T],[T,Y,V,R],[R,T,Y]]
+    EXPECT_EQ(v1.size(), cmp4.size());
+    for (unsigned i = 0; i < v1.size(); i++)
+        EXPECT_RANGE_EQ(v1[i], cmp4[i]);
+
+    // default parameter translation_frames
+    auto const v2 = vec | seqan3::views::translate();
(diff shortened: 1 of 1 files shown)


eseiler:
> I would have added tests for the throws in lines 318 and 391, but if I understand the code correctly they can never be reached because the throw in line 212 will always throw first. Correct?
> (Line number from translate.hpp, see: https://codecov.io/gh/seqan/seqan3/src/master/include/seqan3/range/views/translate.hpp)

Can you check what happens with a default constructed view? Then the enum would be default initialised to `0` and the switch cases should use the default .

MitraDarja:
> > I would have added tests for the throws in lines 318 and 391, but if I understand the code correctly they can never be reached because the throw in line 212 will always throw first. Correct?
> > (Line number from translate.hpp, see: https://codecov.io/gh/seqan/seqan3/src/master/include/seqan3/range/views/translate.hpp)
> 
> Can you check what happens with a default constructed view? Then the enum would be default initialised to `0` and the switch cases should use the default .

For the following code:
```cpp
auto v7 = seqan3::detail::view_translate_single{vec};
    EXPECT_THROW(v7.size(), std::invalid_argument);
    seqan3::debug_stream << v7 << "\n";
    std::cout << v7.size();
```

I get the results:
 Actual: it throws nothing.
TYVR
4

Looks like it is initialized to the `seqan3::translation_frames::FWD_FRAME_0`

eseiler:
> > > I would have added tests for the throws in lines 318 and 391, but if I understand the code correctly they can never be reached because the throw in line 212 will always throw first. Correct?
> > > (Line number from translate.hpp, see: https://codecov.io/gh/seqan/seqan3/src/master/include/seqan3/range/views/translate.hpp)
> > 
> > 
> > Can you check what happens with a default constructed view? Then the enum would be default initialised to `0` and the switch cases should use the default .
> 
> @eseiler
> For the following code:
> 
> ```c++
> auto v7 = seqan3::detail::view_translate_single{vec};
>     EXPECT_THROW(v7.size(), std::invalid_argument);
>     seqan3::debug_stream << v7 << "\n";
>     std::cout << v7.size();
> ```
> 
> I get the results:
> Actual: it throws nothing.
> TYVR
> 4
> 
> Looks like it is initialized to the `seqan3::translation_frames::FWD_FRAME_0`

```cpp
auto v7 = seqan3::detail::view_translate_single{};
```
would be the default ctor.

simonsasse (approved):
Looks good :)

eseiler:
You can use `seqan3::views::translate_single(vec, seqan3::translation_frames{})` to test the throws for `size` and `operator[]`

eseiler (changes requested):
Just 💅 

Can you rebase on the current master?
The coverage build doesn't see that `operator[]` is now covered....

smehringer:
@eseiler Looks good, feel free to do the second review and merge :)

test/unit/range/views/view_translate_test.cpp
```diff
@@ -87,6 +87,14 @@ TYPED_TEST(nucleotide, view_translate_single)
...
     EXPECT_RANGE_EQ(v6, cmp3);
+
+    // Construct with multiple frames
+    EXPECT_THROW(seqan3::views::translate_single(vec, seqan3::translation_frames::FWD), std::invalid_argument);
+
+    // Construct with default (empty) frame
+    auto v7 = seqan3::views::translate_single(vec, seqan3::translation_frames{});
+    EXPECT_THROW(v7.size(), std::invalid_argument);
+    EXPECT_THROW(v7[0], std::invalid_argument);
 }
```
eseiler, line 98:
```suggestion
    EXPECT_THROW(v7[0], std::invalid_argument);
    
    auto const v8 = seqan3::views::translate_single(vec, seqan3::translation_frames{});
    EXPECT_THROW(v8.size(), std::invalid_argument);
    EXPECT_THROW(v8[0], std::invalid_argument);
}
```

eseiler (changes requested):
I think we are missing the const versions?
Let's see what happens

eseiler:
Ok, https://codecov.io/gh/seqan/seqan3/pull/2176/src/include/seqan3/range/views/translate.hpp now shows what's missing, even though the throw in 421 is definitely executed and tested...

MitraDarja:
@eseiler Rebasing on the new master did not solve this problem either... What should we do about it?

eseiler:
> @eseiler Rebasing on the new master did not solve this problem either... What should we do about it?

Can you look at https://codecov.io/gh/seqan/seqan3/pull/2176/src/include/seqan3/range/views/translate.hpp and check if the missing cases are run (either breakpoints in debugger or just some debug printing).

eseiler (approved):

PR #2178, opened 2020-09-28
[FIX] Underflow in search

eseiler:
AddressSanitizer on macOS sees this, others not....
Can be easily confirmed via using `.at()` instead of `[]`.

### DIFF
--- a/include/seqan3/search/detail/search_scheme_algorithm.hpp
+++ b/include/seqan3/search/detail/search_scheme_algorithm.hpp
@@ -372,7 +372,7 @@ inline bool search_ss_deletion(cursor_t cur, query_t & query,
     if (min_error_left_in_block == 0)
     {
         uint8_t const block_id2 = std::min<uint8_t>(block_id + 1, search.blocks() - 1);
-        bool const go_right2 = search.pi[block_id2] > search.pi[block_id2 - 1];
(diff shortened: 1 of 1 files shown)


include/seqan3/search/detail/search_scheme_algorithm.hpp
```diff
@@ -462,7 +462,7 @@ inline bool search_ss_children(cursor_t cur, query_t & query,
                     else
                     {
                         uint8_t const block_id2 = std::min<uint8_t>(block_id + 1, search.blocks() - 1);
-                        bool const go_right2 = search.pi[block_id2] > search.pi[block_id2 - 1];
+                        bool const go_right2 = block_id2 == 0 ? true : search.pi[block_id2] > search.pi[block_id2 - 1];
```
marehr, line 465: the exact same code twice, this screams for putting it into a single function..

marehr (approved):

smehringer (changes requested):
This might be worth a changlog entry?

eseiler:
> This might be worth a changlog entry?

You're sure?
There is no meaningful way to describe the input that may lead the branches being evaluated in a way that causes the overflow + out of bounds access.
We can write `Fixed an issue where an invalid memory access may have happened in the search.` 
But it's not really helpful...

smehringer:
I see, alright then we won't make an entry :)

smehringer (approved):

PR #2180, opened 2020-09-29
[FIX] Make argument_parser-ctor boolean parameter into an proper enum.

smehringer:
This resolves an error I recently encountered, namely that

```cpp
seqan3::argument_parser parser{"test_parser", argc, argv, {"subparser1", "subparser2"}};
```

compiles because the list of subparser is implicitely converted to bool and taken as the version_check boolean parameter. :see_no_evil: 

This is fixed by making the boolean parameter into a proper enum (which should have been done anyway).

### DIFF
--- a/doc/tutorial/argument_parser/disable_version_check.cpp
+++ b/doc/tutorial/argument_parser/disable_version_check.cpp
@@ -2,6 +2,6 @@
 
 int main(int argc, char ** argv)
 {
-    seqan3::argument_parser myparser{"Game-of-Parsing", argc, argv, false};
-    // disable version checks permanently ----------------------------^
+    seqan3::argument_parser myparser{"Game-of-Parsing", argc, argv, seqan3::update_notifications::off};
+    // disable update notifications permanently ----------------------------^
 }
--- a/include/seqan3/argument_parser/auxiliary.hpp
+++ b/include/seqan3/argument_parser/auxiliary.hpp
@@ -248,6 +248,13 @@ enum option_spec
                    */
 };
 
+//!\brief Indicates whether application allows automatic update notifications by the seqan3::argument_parser.
+enum class update_notifications
+{
+    on, //!< Automatic update notifications should be enabled.
+    off //!< Automatic update notifications should be disabled.
+};
+
 /*!\brief Stores all parser related meta information of the seqan3::argument_parser.
  * \ingroup argument_parser
  *
--- a/include/seqan3/argument_parser/detail/version_check.hpp
+++ b/include/seqan3/argument_parser/detail/version_check.hpp
@@ -20,6 +20,7 @@
 #include <iostream>
 #include <regex>
 
+#include <seqan3/argument_parser/auxiliary.hpp>
 #include <seqan3/argument_parser/detail/terminal.hpp>
 #include <seqan3/io/detail/misc.hpp>
 #include <seqan3/io/detail/safe_filesystem_entry.hpp>
@@ -268,7 +269,8 @@ public:
     }
 
     /*!\brief The central decision whether to perform the version check or not.
-     * \param[in] developer_approval Whether the developer approved (true) or not (false).
+     * \param[in] developer_approval Whether the developer approved (update_notifications::on) or not
+     *                               (update_notifications::off).
      * \param[in] user_approval      Whether the user approved (true) or not (false) or did not decide (unset optional).
      *
      * The following rules apply:
@@ -291,9 +293,9 @@ public:
      * if possible (seqan3::detail::is_terminal()), what he wants to do, set the according cookie for the next time
      * and continue. If we cannot ask the user, the default kicks in (do the check).
      */
-    bool decide_if_check_is_performed(bool developer_approval, std::optional<bool> user_approval)
+    bool decide_if_check_is_performed(update_notifications developer_approval, std::optional<bool> user_approval)
     {
-        if (!developer_approval)
+        if (developer_approval == update_notifications::off)
             return false;
 
         if (std::getenv("SEQAN3_NO_VERSION_CHECK") != nullptr) // environment variable was set
--- a/test/unit/argument_parser/format_parse_test.cpp
+++ b/test/unit/argument_parser/format_parse_test.cpp
@@ -14,7 +14,7 @@ TEST(parse_type_test, add_option_short_id)
     std::string option_value;
 
     const char * argv[] = {"./argument_parser_test", "-s", "option_string"};
-    seqan3::argument_parser parser{"test_parser", 3, argv, false};
+    seqan3::argument_parser parser{"test_parser", 3, argv, seqan3::update_notifications::off};
     parser.add_section("My options");       // no-op for code coverage
     parser.add_subsection("My suboptions"); // no-op for code coverage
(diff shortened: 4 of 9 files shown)


include/seqan3/argument_parser/detail/version_check.hpp
```diff
@@ -26,6 +26,18 @@
 #include <seqan3/std/charconv>
 #include <seqan3/version.hpp>
 
+namespace seqan3
+{
+
+//!\brief Indicated whether application allow automatic update notifications by the seqan3::argument_parser.
```
joergi-w, line 32:
```suggestion
//!\brief Indicates whether application allows automatic update notifications by the seqan3::argument_parser.
```

include/seqan3/argument_parser/detail/version_check.hpp
```diff
@@ -26,6 +26,18 @@
...
+enum class update_notifications
```
joergi-w, line 33: A publicly visible (non-detail) data structure is defined in a detail header here... I suggest moving it to `include/seqan3/argument_parser/argument_parser.hpp`?

doc/tutorial/argument_parser/disable_version_check.cpp
```diff
@@ -2,6 +2,6 @@
 
 int main(int argc, char ** argv)
 {
-    seqan3::argument_parser myparser{"Game-of-Parsing", argc, argv, false};
+    seqan3::argument_parser myparser{"Game-of-Parsing", argc, argv, seqan3::update_notifications::disabled};
     // disable version checks permanently ----------------------------^
```
joergi-w, line 6:
```suggestion
    // disable update notifications permanently ----------------------------^
```

joergi-w (changes requested):
Just some details below... And maybe it would be useful to add a test for the bug you found?
```cpp
seqan3::argument_parser parser{"test_parser", argc, argv, {"subparser1", "subparser2"}};
```

include/seqan3/argument_parser/detail/version_check.hpp
smehringer: Good point, I will move it.

smehringer: Note that I could not move it to argument_parser.hpp because argument_parser.hpp depends on version_check.hpp and version_check.hpp needs the `update_notifications` struct so this would create dependency loop. I put it into auxiliary instead.

smehringer:
> Just some details below... And maybe it would be useful to add a test for the bug you found?
> 
> ```c++
> seqan3::argument_parser parser{"test_parser", argc, argv, {"subparser1", "subparser2"}};
> ```

I cannot add a test because it would not compile now since ` {"subparser1", "subparser2"}` cannot be cast to `update_notifications` now.

joergi-w (changes requested):
Of course - if it does not compile, we cannot have a test for the bug. 
I like the idea of putting the enum to the auxiliary file. However, I need to give a red light again, because it seems that the include directive appears in the wrong header file:

include/seqan3/argument_parser/validators.hpp
```diff
@@ -20,6 +20,7 @@
 #include <regex>
 #include <sstream>
 
+#include <seqan3/argument_parser/auxiliary.hpp>
```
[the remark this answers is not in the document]
smehringer: whops, sorry.

joergi-w (approved):
Thank you!

doc/tutorial/argument_parser/disable_version_check.cpp
```diff
@@ -2,6 +2,6 @@
 
 int main(int argc, char ** argv)
 {
-    seqan3::argument_parser myparser{"Game-of-Parsing", argc, argv, false};
-    // disable version checks permanently ----------------------------^
+    seqan3::argument_parser myparser{"Game-of-Parsing", argc, argv, seqan3::update_notifications::disabled};
+    // disable update notifications permanently ----------------------------^
```
rrahn, line 6: So permanentely means here that setting the update notification off is encoded in the binary and therefore never asks on the command line? Is that clear from the tutorial that this happens?

include/seqan3/argument_parser/auxiliary.hpp
```diff
@@ -248,6 +248,13 @@ enum option_spec
                    */
 };
 
+//!\brief Indicates whether application allows automatic update notifications by the seqan3::argument_parser.
+enum class update_notifications
+{
+    enabled, //!< Automatic update notifications should be enabled.
+    disabled //!< Automatic update notifications should be disabled.
+};
```
rrahn, line 256:
```suggestion
    on, //!< Automatic update notifications should be enabled.
    off //!< Automatic update notifications should be disabled.
};
```

Just wondering, as it is essentially the same but shorter.

test/unit/argument_parser/format_parse_test.cpp
```diff
@@ -14,7 +14,7 @@ TEST(parse_type_test, add_option_short_id)
     std::string option_value;
 
     const char * argv[] = {"./argument_parser_test", "-s", "option_string"};
-    seqan3::argument_parser parser{"test_parser", 3, argv, false};
+    seqan3::argument_parser parser{"test_parser", 3, argv, seqan3::update_notifications::disabled};
```
rrahn, line 17: Does these test have a test fixture? If so you could provide a functio; `make_parser_without_updates(args...)` which always creates a parser with the arguments but the update notification set to `off`.

test/unit/argument_parser/format_parse_validators_test.cpp
```diff
@@ -135,7 +135,7 @@ TEST(validator_test, input_file)
         // option
         std::string const & path = tmp_name.get_path().string();
         const char * argv[] = {"./argument_parser_test", "-i", path.c_str()};
-        seqan3::argument_parser parser{"test_parser", 3, argv, false};
+        seqan3::argument_parser parser{"test_parser", 3, argv, seqan3::update_notifications::disabled};
```
rrahn, line 138: Pretty much like this here. They could even share the test fixture instance. Then you don't have these thousand line replacements in the future anymore.

rrahn (changes requested):

doc/tutorial/argument_parser/disable_version_check.cpp
smehringer: > So permanentely means here that setting the update notification off is encoded in the binary and therefore never asks on the command line?

yes

> Is that clear from the tutorial that this happens?

In the tutorial I refer to the documentation of `seqan3::argument_parser` were I have an extra section 
image

So I think it is clear?

test/unit/argument_parser/format_parse_test.cpp
smehringer: No this test does not have a test fixture yet. I would avoid restructuring the whole test now but I can make a follow up issue?

rrahn: Well, yes. I mean the whole point of incremental refactoring is to do it, when you are woking on it. You can make a follow up issue but that shoud be attended right away anyway. It is ok to have this PR merged before though, but I suggest handling this right away would be the best choice.

rrahn (approved):
thank you ;)

PR #2181, opened 2020-10-02
[MISC] Delete unreachable if in fastq format

MitraDarja:

### DIFF
--- a/include/seqan3/io/sequence_file/format_fastq.hpp
+++ b/include/seqan3/io/sequence_file/format_fastq.hpp
@@ -178,12 +178,6 @@ protected:
             }
         }
 
-        /* 2nd ID line */
-        if (*stream_it != '+') // [[unlikely]]
-        {
-            throw parse_error{std::string{"Expected '+' on beginning of 2nd ID line, got: "} +
-                              detail::make_printable(*stream_it)};
-        }
         detail::consume(stream_view | views::take_line_or_throw);
 
         /* Qualities */


wvandertoorn:
> Merging [#2181](https://codecov.io/gh/seqan/seqan3/pull/2181?src=pr&el=desc) into [master](https://codecov.io/gh/seqan/seqan3/commit/5a910046bacaa6bcd6f3bd7282ac9181a7265f9d?el=desc) will **not change** coverage.

Should the changes not show up here?

MitraDarja:
You are right and it can't work because the uncovered line can never be reached, because

```c++
auto seq_view = stream_view \| views::take_until_or_throw(is_char<'+'>)    // until 2nd ID line
```

is requiring a plus. Question is, do we want to cover the unlikely case that a plus is missing in a fastq-file? Then I need to adapt the condition above or we leave it as it and delete the check, if there is a plus. (@seqan/core Could you make a decision?)

eseiler:
The line to cover: https://codecov.io/gh/seqan/seqan3/src/master/include/seqan3/io/sequence_file/format_fastq.hpp#L185
The view that may throw: https://codecov.io/gh/seqan/seqan3/src/master/include/seqan3/io/sequence_file/format_fastq.hpp#L150

Looks pretty unreachable if the fastq format input is correct. What happens if there is a `+` in the first line? Then the second line should be parsed incorrectly and the if should catch that the second line does not begin with `+`?

MitraDarja:
@eseiler 

> The line to cover: https://codecov.io/gh/seqan/seqan3/src/master/include/seqan3/io/sequence_file/format_fastq.hpp#L185
> The view that may throw: https://codecov.io/gh/seqan/seqan3/src/master/include/seqan3/io/sequence_file/format_fastq.hpp#L150
> 
> Looks pretty unreachable if the fastq format input is correct. What happens if there is a `+` in the first line? Then the second line should be parsed incorrectly and the if should catch that the second line does not begin with `+`?

Not sure, what you mean with first line:
```
"@ID1+\n"
"ACGTTTTTTTT\nTTTTTTT\n"
 "-\n"
"!##$\n%&'()*+,-./++-\n"
```

The plus here has no impact.

```
"@ID1\n"
"+ACGTTTTTTTT\nTTTTTTT\n"
 "-\n"
"!##$\n%&'()*+,-./++-\n"
```
This does not throw.

```
"@ID1\n"
"ACGTT+TTTTTT\nTTTTTTT\n"
 "-\n"
"!##$\n%&'()*+,-./++-\n"
```

This does not throw.

```
"@ID1\n"
"ACGTT+TTTTTT\nTTTTTTT\n+"
 "-\n"
"!##$\n%&'()*+,-./++-\n"
```
This does not throw.

It always parses beforehand until there is a +, so the if can only be reached if `*stream_it _= '+'`.

eseiler:
I think the if can go because of the `views::take_until_or_throw(is_char<'+'>)` part.

wvandertoorn (approved):

smehringer (approved):
I agree that this part can never be reached. The only alternative would be to change the parsing to keep the if which would probably increase the runtime so I'm also in favor of removing the if clause.

smehringer (changes requested):
One thing though: Can you change the commit message to what is actually happening now? :)

MitraDarja:
@smehringer 

> One thing though: Can you change the commit message to what is actually happening now? :)

I already did? I also changed the name of the PR here. :)

smehringer:
Oh sorry!

smehringer (approved):

PR #2182, opened 2020-10-02
[TEST] Add coverage to sequence file format fasta test

MitraDarja:

### DIFF
--- a/test/unit/io/sequence_file/sequence_file_format_fasta_test.cpp
+++ b/test/unit/io/sequence_file/sequence_file_format_fasta_test.cpp
@@ -180,6 +180,31 @@ TEST_F(read, mixed_issues)
     do_read_test(input);
 }
 
+TEST_F(read, fail_no_newline_after_id)
+{
+    std::string input{
+        ">ID1"
+        "ACGTTTTTTTTTTTTTTT"
+    };
+    std::stringstream istream{input};
+
+    seqan3::sequence_file_input fin{istream, seqan3::format_fasta{}};
+    EXPECT_THROW(fin.begin(), seqan3::unexpected_end_of_input);
+}
+
+TEST_F(read, fail_no_newline_after_truncate_id)
+{
+    std::string input{
+        ">ID1 to_be_truncated"
+        "ACGTTTTTTTTTTTTTTT"
+    };
+    std::stringstream istream{input};
+
+    seqan3::sequence_file_input fin{istream, seqan3::format_fasta{}};
+    fin.options.truncate_ids = true;
+    EXPECT_THROW(fin.begin(), seqan3::unexpected_end_of_input);
+}
+
 // ----------------------------------------------------------------------------
 // writing
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


test/unit/io/sequence_file/sequence_file_format_fasta_test.cpp
```diff
@@ -180,6 +180,33 @@ TEST_F(read, mixed_issues)
...
 }
 
+TEST_F(read, fail_no_newline_after_id)
+{
+    std::string input;
+    std::stringstream istream{input =
+    {
+        ">ID1"
+        "ACGTTTTTTTTTTTTTTT"
+    }};
```
eseiler, line 190:
```suggestion
    std::string input
    {
        ">ID1"
        "ACGTTTTTTTTTTTTTTT"
    };
    std::stringstream istream{input};
```

test/unit/io/sequence_file/sequence_file_format_fasta_test.cpp
```diff
@@ -180,6 +180,33 @@ TEST_F(read, mixed_issues)
...
+}
+
+TEST_F(read, fail_no_newline_after_truncate_id)
+{
+    std::string input;
+    std::stringstream istream{input =
+    {
+        ">ID1"
+        "ACGTTTTTTTTTTTTTTT"
+    }};
```
eseiler, line 203:
```suggestion
    std::string input
    {
        ">ID1"
        "ACGTTTTTTTTTTTTTTT"
    };
    std::stringstream istream{input};
```

eseiler (changes requested):

eseiler (approved):

test/unit/io/sequence_file/sequence_file_format_fasta_test.cpp
```diff
@@ -180,6 +180,31 @@ TEST_F(read, mixed_issues)
...
+    std::stringstream istream{input};
+
+    seqan3::sequence_file_input fin{istream, seqan3::format_fasta{}};
+    EXPECT_THROW(fin.begin(), seqan3::unexpected_end_of_input);
+}
+
+TEST_F(read, fail_no_newline_after_truncate_id)
+{
+    std::string input{
+        ">ID1"
```
smehringer, line 198:
```suggestion
        ">ID1 to_be_truncated"
```

smehringer (changes requested):
Just one small thing because truncating ids would best be tested with something to truncate I guess

smehringer:
well that's a bug then. Good thing we added a test. I sent you a patch on gitter that you can apply (I would do a separate [FIX] commit prior to the one adding the test)

smehringer (changes requested):
Looks good! 
Can you just clean the commit history? It should look something like this:
`[FIX] Check if sequence is present when reading.`
`[TEST] Add fasta format test for no-newline--after-id-error.`

The Fix has to be *before* the Test such that the history is always "buildable" :)
``

smehringer (approved):
Perfect thanks :)

PR #2184, opened 2020-10-02
[FIX] Allow '*' at the beginning of string fields in SAM format.

### LINKED ISSUE #2195: SAM/BAM IO: Quality string starting with '*' mess up the parsing
## Platform

  - SeqAn version: current master and release branch
  - Operating system: linux
  - Compiler: GCC7/8/9/10

## Description

Quality string starting with '*' mess up the parsing

### How to repeat the problem

The following test added to format_sam.hpp reproduces the error
```cpp
TEST_F(sam_format, quality_string_starts_with_asterics)
{
    {
        std::istringstream istream
        {
            "r1\t4\t1\t10\t0\t5M\t=\t136097\t-121\tACTGA\t*9>9;\tNM:i:1\tMQ:i:0\n"
        };
        seqan3::alignment_file_input fin{istream, seqan3::format_sam{}};

        EXPECT_EQ(seqan3::get<seqan3::field::qual>(*fin.begin()).size(), 5u);
    }
}
```

### Expected behaviour

The test should pass. Quality strings or any other strings that can be marked as empty with `*` should still be allowed to start with `*`, e.g. `*id_1`.

### Actual behaviour

Parsing is messed up because the string is only parsed for `*` and then continues on the rest of the string trying to parse another field (in the case of the quality string, it tries to parse the rest of the quality string as SAM tags throwing errors on the way)

smehringer:
An error that occurred for @h-2. The offending line was added as a test.

Fixes #2195

### DIFF
--- a/include/seqan3/io/alignment_file/format_sam_base.hpp
+++ b/include/seqan3/io/alignment_file/format_sam_base.hpp
@@ -371,11 +371,23 @@ inline void format_sam_base::read_field(stream_view_type && stream_view,
 template <typename stream_view_type, std::ranges::forward_range target_range_type>
 inline void format_sam_base::read_field(stream_view_type && stream_view, target_range_type & target)
 {
-    if (!is_char<'*'>(*std::ranges::begin(stream_view)))
-        std::ranges::copy(stream_view | views::char_to<std::ranges::range_value_t<target_range_type>>,
-                          std::cpp20::back_inserter(target));
-    else
-        std::ranges::next(std::ranges::begin(stream_view)); // skip '*'
+    using target_range_value_t = std::ranges::range_value_t<target_range_type>;
+    using begin_iterator_t = std::ranges::iterator_t<stream_view_type>;
+    using end_iterator_t = std::ranges::sentinel_t<stream_view_type>;
+
+    // Note that we need to cache the begin iterator since the stream_view is an input range that may be consuming
+    // and in that case might read `past-the-end` on a second call of std::ranges::begin.
+    if (auto it = std::ranges::begin(stream_view); it != std::ranges::end(stream_view))
+    {
+        // Write to target if field does not represent an empty string, denoted as single '*' character.
+        if (char c = *it; !(++it == std::ranges::end(stream_view) && c == '*'))
+        {
+            target.push_back(seqan3::assign_char_to(c, target_range_value_t{}));
+            std::ranges::copy(std::ranges::subrange<begin_iterator_t, end_iterator_t>{it, std::ranges::end(stream_view)}
+                              | views::char_to<target_range_value_t>,
+                              std::cpp20::back_inserter(target));
+        }
+    }
 }
 
 /*!\brief Reads arithmetic fields using std::from_chars.
--- a/test/unit/io/alignment_file/format_sam_test.cpp
+++ b/test/unit/io/alignment_file/format_sam_test.cpp
@@ -328,3 +328,34 @@ TEST_F(sam_format, write_different_header)
     EXPECT_EQ(ostream.str(),
               "@HD\tVN:1.6\tSO:coordinate\tSS:query\tGO:reference\n@SQ\tSN:ref\tLN:34\n*\t0\tref\t1\t0\t*\t*\t0\t0\t*\t*\n");
 }
+
+TEST_F(sam_format, issue2195)
+{ // see issue https://github.com/seqan/seqan3/issues/2195
+    {
+        std::istringstream istream
+        {
+            "*r1\t4\t1\t10\t0\t5M\t=\t136097\t-121\tACTGA\t*9<9;\tNM:i:1\tMQ:i:0\n"
+        };
+        seqan3::alignment_file_input fin{istream, seqan3::format_sam{}};
+
+        using seqan3::operator""_phred42;
+        std::vector<seqan3::phred42> expected_quality = "*9<9;"_phred42;
+
+        EXPECT_RANGE_EQ(seqan3::get<seqan3::field::id>(*fin.begin()), std::string{"*r1"});
+        EXPECT_RANGE_EQ(seqan3::get<seqan3::field::qual>(*fin.begin()), expected_quality);
+    }
+
+    {
+        std::istringstream istream
+        {
+            "*\t4\t1\t10\t0\t2M\t=\t136097\t-121\tAC\t*1\tNM:i:1\tMQ:i:0\n"
+        };
+        seqan3::alignment_file_input fin{istream, seqan3::format_sam{}};
+
+        using seqan3::operator""_phred42;
+        std::vector<seqan3::phred42> expected_quality = "*1"_phred42;
+
+        EXPECT_RANGE_EQ(seqan3::get<seqan3::field::id>(*fin.begin()), std::string{""});
+        EXPECT_RANGE_EQ(seqan3::get<seqan3::field::qual>(*fin.begin()), expected_quality);
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,13 @@ Note that 3.1.0 will be the first API stable release and interfaces in this rele
 * SeqAn 3.0.2 is known to compile with GCC 7.5, 8.4, 9.3 and 10.2. Future versions (e.g. GCC 10.3 and 11) might work,
   but weren’t yet available at the time of this release.
 
+## Notable Bug-fixes
+
+#### I/O
+
+* The SAM/BAM format reads the quality field (or any other text field) that starts with an asterisk (`*`) but is not
+  empty correctly now ([\#2184](https://github.com/seqan/seqan3/pull/2184)).
+
 ## New features
 
 #### Alignment


include/seqan3/io/alignment_file/format_sam_base.hpp
```diff
@@ -49,6 +49,8 @@
 #include <seqan3/std/concepts>
 #include <seqan3/std/ranges>
 
+#include <seqan3/core/debug_stream.hpp>
+
 namespace seqan3::detail
 {
```
eseiler, line 55:
```suggestion
#include <seqan3/std/ranges>

namespace seqan3::detail
{
```

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -407,7 +407,7 @@ inline void format_sam::read_alignment_record(stream_type & stream,
                   "The ref_offset must be a specialisation of std::optional.");
 
     auto stream_view = views::istreambuf(stream);
-    auto field_view = stream_view | views::take_until_or_throw_and_consume(is_char<'\t'>);
+    auto field_view = stream_view | views::take_until_or_throw(is_char<'\t'>);
```
eseiler, line 410: Why this change?

include/seqan3/io/alignment_file/format_sam_base.hpp
smehringer: whops thanks :)

include/seqan3/io/alignment_file/format_sam.hpp
smehringer: So the initial problem was that `*` indicates and empty SAM field and the function `read_field` only checks if `*` is present (before this PR). But I did not check if the field ONLY contains the `*`. 

So I need to check if there is a `\t` after the `*`but I cannot do this if the `\t` is consumed in general (the change you asked the reason for). Note that I cannot check if `field_view` is empty instead of checking if the next char is `\t` because on an input_range the `take..` view will always "regenerate".

```cpp
// input_range = "1\t2\t3\n"
auto v = input_range | take_until_and_consume('\t');
seqan3::debug_stream << v << std::endl; // prints 1
seqan3::debug_stream << v << std::endl;  // prints 2
```

Is that understandable?

eseiler: Yep; unfortunate that we can't keep the consume :(

h-2:
This slightly shorter patch works for me, too;

```cpp
template <typename stream_view_type, std::ranges::forward_range target_range_type>
inline void format_sam_base::read_field(stream_view_type && stream_view, target_range_type & target)
{
    auto it = std::ranges::begin(stream_view);
    if (it == std::ranges::end(stream_view))
        return;

    char c = *it;
    ++it;

    using val_t = std::ranges::range_value_t<target_range_type>;

    if (it != std::ranges::end(stream_view)) // This is a string with more than one character
    {
        // the first character could be a '*'
        target.push_back(assign_char_to(c, val_t{}));
        std::ranges::copy(stream_view | views::char_to<val_t>, std::cpp20::back_inserter(target));
    }
    else if (c != '*') // single '*' means "empty sequence"; other character is 1-element range
    {
        target.push_back(assign_char_to(c, val_t{}));
    }
}
```
Just replace read_field with this, no need to change anything else. 

format_sam_test and format_bam_test passed for me, also by problematic file.

include/seqan3/io/alignment_file/format_sam.hpp
h-2: This is only problematic if there are empty fields / double tabs. This is not allowed in SAM/BAM.

Note that in your example `v`is "at-end" in between the calls. The second (implicit) call to to `begin()` resets it. As long as you don't do that, there is no problem.

smehringer: > Note that in your example vis "at-end" in between the calls, i.e. the first begin iterator does compare true to the end. Otherwise it wouldn't stop printing after "1". Only when you create a new iterator (which implicitly happens in your example) does this new iterator start at the next field.

Ah ok thanks. That's the fact that I was missing when trying to fix this :)

eseiler (changes requested):
LGTM, but you deleted the test?

smehringer:
ah damnit

h-2:
In general quite a few places like https://github.com/seqan/seqan3/blob/master/include/seqan3/io/alignment_file/format_bam.hpp#L954  use very many repeated calls to `std::ranges::begin(stream_view)` where I think it would be enough to just call `auto it = std::ranges::begin(stream_view)` once and than work with `it`. This is 
1. safer, because it prevents the potential "reset" of the stream discussed here
2. easier to read (`++it` vs `std::ranges::next(std::ranges::begin(stream_view))` )
3. maybe even a little faster

smehringer:
> In general quite a few places like https://github.com/seqan/seqan3/blob/master/include/seqan3/io/alignment_file/format_bam.hpp#L954 use very many repeated calls to `std::ranges::begin(stream_view)` where I think it would be enough to just call `auto it = std::ranges::begin(stream_view)` once and than work with `it`. This is
> 
>     1. safer, because it prevents the potential "reset" of the stream discussed here
> 
>     2. easier to read (`++it` vs `std::ranges::next(std::ranges::begin(stream_view))` )
> 
>     3. maybe even a little faster

Thanks for the input :) I will add an issue in the backlog about this.

eseiler (approved):

rrahn:
> An error that occurred for @h-2. The offending line was added as a test.

issue?

test/unit/io/alignment_file/format_sam_test.cpp
```diff
@@ -328,3 +328,16 @@ TEST_F(sam_format, write_different_header)
     EXPECT_EQ(ostream.str(),
               "@HD\tVN:1.6\tSO:coordinate\tSS:query\tGO:reference\n@SQ\tSN:ref\tLN:34\n*\t0\tref\t1\t0\t*\t*\t0\t0\t*\t*\n");
 }
+
+TEST_F(sam_format, quality_string_starts_with_asterics)
```
rrahn, line 332: Would be good to have a tracking issue for this.

include/seqan3/io/alignment_file/format_sam_base.hpp
```diff
@@ -371,11 +371,23 @@ inline void format_sam_base::read_field(stream_view_type && stream_view,
...
+
+        if (it != std::ranges::end(stream_view)) // This is a string with more than one character
+        {
+            target.push_back(assign_char_to(c, val_t{}));
+            std::ranges::copy(stream_view | views::char_to<val_t>, std::cpp20::back_inserter(target));
+        }
+        else if (c != '*') // single '*' means "empty sequence"; other character is 1-element range
+        {
+            target.push_back(assign_char_to(c, val_t{}));
+        }
```
rrahn, line 389:
```suggestion
        // Write to target if field does not represent an empty string, denoted as single '*' character.
        if (char c = *it; !(c == '*' && ++it == std::ranges::end(view)))
        {
            target.push_back(seqan3::assign_char_to(c, val_t{}));
            std::ranges::copy(stream_view | views::char_to<val_t>, std::cpp20::back_inserter(target));
        }
```

include/seqan3/io/alignment_file/format_sam_base.hpp
```diff
@@ -371,11 +371,23 @@ inline void format_sam_base::read_field(stream_view_type && stream_view,
 template <typename stream_view_type, std::ranges::forward_range target_range_type>
 inline void format_sam_base::read_field(stream_view_type && stream_view, target_range_type & target)
 {
-    if (!is_char<'*'>(*std::ranges::begin(stream_view)))
-        std::ranges::copy(stream_view | views::char_to<std::ranges::range_value_t<target_range_type>>,
-                          std::cpp20::back_inserter(target));
-    else
-        std::ranges::next(std::ranges::begin(stream_view)); // skip '*'
+    using val_t = std::ranges::range_value_t<target_range_type>;
```
rrahn, line 374:
```suggestion
    using target_range_value_t = std::ranges::range_value_t<target_range_type>;
```

rrahn (changes requested):
Some simplifications. And I'd like to track this via issues. Because it took a bit to understand what the problem was  by just looking at the code.

include/seqan3/io/alignment_file/format_sam_base.hpp
h-2: This will not work for all cases, because -- as discussed above -- you are calling `std::ranges::begin()` a second time implicitly in line 382 (when iterating over `stream_view`). In the case of an underlying view that is consuming and it was at end, this will read "past-the-end".

Another issue with this code is, that in the case where `c != '*'` you are not incrementing the iterator (because of lazy-evaluation of `&&`) and thus writing `c` to the target twice.

I don't like have a line twice in the code, but I think it is the easiest (correct) solution.

rrahn: >This will not work for all cases, because -- as discussed above -- you are calling std::ranges::begin() a second time implicitly in line 382 (when iterating over stream_view). In the case of an underlying view that is consuming and it was at end, this will read "past-the-end".

Oh, that is actually true. Good point. But nothing that can be easily fixed by calling the right interface:
Since we already have the iterator we can just use it to form a subrange which does not consume anything
```cpp
std::ranges::copy(std::ranges::subrange<...>{it, std::ranges::end(view)} | views::char_to<val_t>, ...);
```
That should work right?

> Another issue with this code is, that in the case where c != '*' you are not incrementing the iterator (because of lazy-evaluation of &&) and thus writing c to the target twice.

Yeah, that's right. I had the order the other way around, which schould be guaranteed to work. In the end, I thought it is nicer to have the char check before but this would introduce a subtle bug indeed.

h-2: > That should work right?

Yes, the proposed solution should work and is more elegant! Although I think it is considerably more difficult to parse for humans while the slightly longer one is a lot more straight-forward (just thinking about how easy it was for you to make this mistake and that I needed to think hard on whether it works now 🤓 ).

But I will leave others to judge 😉 I have no strong preference about it.

rrahn: The problem is that when writing an algorithm with ranges you should pay attention to when you call begin. I don't think  that more verbose code makes the actual problem clearer, as it is quite subtle. At least not for someone who is reading the code for the first time. So I would support the concise version with a clearer comment at the beginning.

rrahn: Anyhow, the introduced issues make up some good test cases that should be added.

smehringer: > Anyhow, the introduced issues make up some good test cases that should be added.

I don't really see the good test cases. The function in question is already used with a consuming and a non-consuming input range, both empty (`*`) in some cases and non-empty in others. The case of a field starting with `*` but not being empty was also added (I added additionally added for the id field which is consuming in the test case - that one is new).

smehringer:
@rrahn polite ping

include/seqan3/io/alignment_file/format_sam_base.hpp
```diff
@@ -371,11 +371,24 @@ inline void format_sam_base::read_field(stream_view_type && stream_view,
...
+    {
+        char c = *it;
+        // Write to target if field does not represent an empty string, denoted as single '*' character.
+        if (!(++it == std::ranges::end(stream_view) && c == '*'))
+        {
+            target.push_back(seqan3::assign_char_to(c, target_range_value_t{}));
+            std::ranges::copy(std::ranges::subrange<begin_iterator_t, end_iterator_t>{it, std::ranges::end(stream_view)}
+                              | views::char_to<target_range_value_t>,
+                              std::cpp20::back_inserter(target));
+        }
```
rrahn, line 390:
```suggestion
        // Write to target if field does not represent an empty string, denoted as single '*' character.
        if (char c = *it; !(++it == std::ranges::end(stream_view) && c == '*'))
        {
            target.push_back(seqan3::assign_char_to(c, target_range_value_t{}));
            std::ranges::copy(std::ranges::subrange<begin_iterator_t, end_iterator_t>{it, std::ranges::end(stream_view)}
                              | views::char_to<target_range_value_t>,
                              std::cpp20::back_inserter(target));
        }
```

test/unit/io/alignment_file/format_sam_test.cpp
```diff
@@ -328,3 +328,26 @@ TEST_F(sam_format, write_different_header)
...
+        seqan3::alignment_file_input fin{istream, seqan3::format_sam{}};
+
+        std::vector<seqan3::phred42> expected_quality
+        {
+            seqan3::assign_char_to('*', seqan3::phred42{}),
+            seqan3::assign_char_to('9', seqan3::phred42{}),
+            seqan3::assign_char_to('<', seqan3::phred42{}),
+            seqan3::assign_char_to('9', seqan3::phred42{}),
+            seqan3::assign_char_to(';', seqan3::phred42{})
+        };
```
rrahn, line 348:
```suggestion
        using seqan3::operator""_phred42;
        std::vector<seqan3::phred42> expected_quality = "*9<9;"_phred42;
```

test/unit/io/alignment_file/format_sam_test.cpp
```diff
@@ -328,3 +328,26 @@ TEST_F(sam_format, write_different_header)
     EXPECT_EQ(ostream.str(),
               "@HD\tVN:1.6\tSO:coordinate\tSS:query\tGO:reference\n@SQ\tSN:ref\tLN:34\n*\t0\tref\t1\t0\t*\t*\t0\t0\t*\t*\n");
 }
+
+TEST_F(sam_format, issue2195)
+{ // see issue https://github.com/seqan/seqan3/issues/2195
+    {
+        std::istringstream istream
+        {
+            "*r1\t4\t1\t10\t0\t5M\t=\t136097\t-121\tACTGA\t*9<9;\tNM:i:1\tMQ:i:0\n"
```
rrahn, line 337: Add also an example with exactly one other character, e.g. `*1`

rrahn (changes requested):
looks in general fine. Some small suggestions. You can already rebase.

test/unit/io/alignment_file/format_sam_test.cpp
smehringer: done

rrahn (approved):

PR #2186, opened 2020-10-05
[TEST] add edit distance alignment benchmark template

marehr:
This adds benchmarks (not used yet) for the direct call to the edit distance algorithm.

### DIFF
--- a/test/performance/alignment/edit_distance_benchmark_template.hpp
+++ b/test/performance/alignment/edit_distance_benchmark_template.hpp
@@ -1,0 +1,99 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2020, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+#pragma once
+
+#include <benchmark/benchmark.h>
+
+#include <seqan3/test/alignment/align_pairwise_edit_distance.hpp>
+#include <seqan3/test/performance/sequence_generator.hpp>
+#include <seqan3/test/performance/units.hpp>
+#include <seqan3/test/seqan2.hpp>
+
+// ----------------------------------------------------------------------------
+//  seqan3 edit distance pairwise alignment benchmarks
+// ----------------------------------------------------------------------------
+
+// This directly benchmarks the alignment algorithm without going through the whole align_pairwise function
+template <typename sequence_pair_generator_t, typename align_configs_t>
+void seqan3_align_pairwise_edit_distance_benchmark(benchmark::State & state,
+                                                   sequence_pair_generator_t sequence_pair_generator,
+                                                   align_configs_t && edit_distance_cfg)
+{
+    using seqan3::test::edit_distance_algorithm;
+
+    auto sequence_pair_or_pairs = sequence_pair_generator(state);
+    constexpr bool collection_benchmark = sequence_pair_generator.is_collection;
+    using sequence_t = typename sequence_pair_generator_t::sequence_t;
+
+    int64_t total = 0;
+
+    auto algorithm = edit_distance_algorithm::select<sequence_t, sequence_t>(edit_distance_cfg);
+
+    if constexpr (collection_benchmark)
+    {
+        for (auto _ : state)
+            for (auto && [sequence1, sequence2] : sequence_pair_or_pairs)
+                total += algorithm(sequence1, sequence2, edit_distance_cfg).score();
+    }
+    else
+    {
+        for (auto _ : state)
+        {
+            auto & [sequence1, sequence2] = sequence_pair_or_pairs;
+            total += algorithm(sequence1, sequence2, edit_distance_cfg).score();
+        }
+    }
+
+    std::conditional_t<collection_benchmark, decltype(std::views::all), decltype(std::views::single)> view_adaptor{};
+    state.counters["cells"] = seqan3::test::pairwise_cell_updates(view_adaptor(sequence_pair_or_pairs),
+                                                                  edit_distance_cfg);
+    state.counters["CUPS"] = seqan3::test::cell_updates_per_second(state.counters["cells"]);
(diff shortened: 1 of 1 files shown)


test/performance/alignment/edit_distance_benchmark_template.hpp
```diff
@@ -0,0 +1,98 @@
...
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+#pragma once
+
+#include <benchmark/benchmark.h>
+
+#include <seqan3/test/alignment/align_pairwise_edit_distance.hpp>
+#include <seqan3/test/performance/units.hpp>
+#include <seqan3/test/performance/sequence_generator.hpp>
```
joergi-w, line 14: Bring in alphabetical order, please ;)

test/performance/alignment/edit_distance_benchmark_template.hpp
```diff
@@ -0,0 +1,98 @@
...
+        {
+            for (auto && [sequence1, sequence2] : sequence_pair_or_pairs)
+                total += algorithm(sequence1, sequence2, edit_distance_cfg).score();
+        }
+        else
+        {
+            auto & [sequence1, sequence2] = sequence_pair_or_pairs;
+            total += algorithm(sequence1, sequence2, edit_distance_cfg).score();
+        }
+    }
```
joergi-w, line 48: Please swap the `if` and the `for` statements, because `is_collection` is constant and does not need to be decided in each loop cycle.

test/performance/alignment/edit_distance_benchmark_template.hpp
```diff
@@ -0,0 +1,98 @@
...
+// ----------------------------------------------------------------------------
+
+template <typename sequence_pair_generator_t, typename align_cfg_t>
+void seqan2_align_pairwise_edit_distance_benchmark(benchmark::State & state,
+                                                   sequence_pair_generator_t sequence_pair_generator,
+                                                   align_cfg_t seqan3_align_cfg)
+{
+    using seqan3::test::edit_distance_algorithm_seqan2;
+
+    auto [sequences1, sequences2] = sequence_pair_generator(state);
```
joergi-w, line 70: This is different from the SeqAn3 test! Here the sequences are already in separate variables before the time is measured, although above the structured binding is inside the time-critical section.

test/performance/alignment/edit_distance_benchmark_template.hpp
```diff
@@ -0,0 +1,98 @@
...
+    constexpr bool is_collection = sequence_pair_generator.is_collection;
+    using sequence_t = typename sequence_pair_generator_t::sequence_t;
+
+    auto algorithm_seqan2 = edit_distance_algorithm_seqan2::select<sequence_t, sequence_t>(seqan3_align_cfg);
+
+    int64_t total = 0;
+    for (auto _ : state)
+    {
+        if constexpr (is_collection)
```
joergi-w, line 79: Again, please swap the logic here...
```cpp
if constexpr (is_collection)
{
    for (auto _ : state) ...
}
else
{
    for (auto _ : state) ...
}
```

test/performance/alignment/edit_distance_benchmark_template.hpp
```diff
@@ -0,0 +1,98 @@
...
+        {
+            for (size_t i = 0u; i < seqan::length(sequences1); ++i)
```
joergi-w, line 81: The literal for size_t is `ull`, no?
```suggestion
            for (size_t i = 0ull; i < seqan::length(sequences1); ++i)
```

joergi-w: Why don't we use a for-each loop, as in the SeqAn3 version?

joergi-w (changes requested):

rrahn:
can this PR be somehow connected to an open Issue?

test/performance/alignment/edit_distance_benchmark_template.hpp
marehr: It is like before and this has absolutely no performance impact in an optimized build.

test/performance/alignment/edit_distance_benchmark_template.hpp
marehr: Okay I can change it to have the semantics that this are different benchmarks

test/performance/alignment/edit_distance_benchmark_template.hpp
marehr: > The literal for size_t is `ull`, no?

You don't know what `size_t` is as a type. I just grep'ed the library and `ull` is only used in some place, we normally use `size_t i = 0u` to indicate that this is an unsigned value.

> Why don't we use a for-each loop, as in the SeqAn3 version?

This is how seqan2 recommends using the API. I guess for-each would work, but I know that begin and end is defined for literally any namespace within seqan2, so I would try to avoid using it.

joergi-w: Okay, then it's fine. Note that [cppreference](https://en.cppreference.com/w/cpp/types/size_t) initializes `size_t` simply with `= 0` in the example.

test/performance/alignment/edit_distance_benchmark_template.hpp
joergi-w: I trust you in this point...

joergi-w (approved):

test/performance/alignment/edit_distance_benchmark_template.hpp
```diff
@@ -0,0 +1,99 @@
...
+#include <seqan3/test/alignment/align_pairwise_edit_distance.hpp>
+#include <seqan3/test/performance/sequence_generator.hpp>
+#include <seqan3/test/performance/units.hpp>
+#include <seqan3/test/seqan2.hpp>
+
+// ----------------------------------------------------------------------------
+//  seqan3 edit distance pairwise alignment benchmarks
+// ----------------------------------------------------------------------------
+
+//!\brief This directly benchmarks the alignment algorithm without going through the whole align_pairwise function
```
smehringer, line 21:
```suggestion
// This directly benchmarks the alignment algorithm without going through the whole align_pairwise function
```
I hope this doesn't appear in the documentation anyway but just for future copy and paste I would avoid it

smehringer (approved):

PR #2187, opened 2020-10-05
[TEST] DRY direct edit distance algorithm call in benchmarks

marehr:
This change dries existing code to directly calling the edit distance algorithm (without going through align_pairwise and the associated cost of it). This change might also be usable in the unit tests, but I first have to look if I can use it and what I need to change to make it work :)


simonsasse (approved):
Couldn't find anything. Looks good 👍

rrahn:
Can this be onnected with an existing issue?

rrahn:
So this looks good. But it is just a refactoring of the methods used before, which then can be reusued by other tests, right? So you cannot tell if that compiles faster than going through the regular align_pairwise interface? Or do you have timings for that?

rrahn (approved):
Nice :)

PR #2194, opened 2020-10-07
[INFRA] add-path is deprecated

eseiler:
See https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

New way: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#adding-a-system-path


marehr (approved):
interesting :)

smehringer (approved):

smehringer:
Code coverage fails xD

PR #2206, opened 2020-10-13
[INFRA] build documentation only once

eseiler:
~~Blocked by https://github.com/seqan/seqan3/pull/2205~~ (Works as expected: Failed before #2205 was merged and succeeds now)

For the documentation we do `make` and then `ctest`. This obscures errors that may occur during the build via `make`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -289,7 +289,11 @@ jobs:
         run: |
           ccache -p || true
           cd seqan3-build
-          make -k -j2
+          if [[ "${{ matrix.build }}" =~ ^(documentation)$ ]]; then
+            make download-cppreference-doxygen-web-tag
+          else
+            make -k -j2
+          fi
           ccache -s || true
 
       - name: Run tests


.github/workflows/ci.yml
```diff
@@ -282,7 +282,10 @@ jobs:
         run: |
           ccache -p || true
           cd seqan3-build
-          make -k -j2
+          if [[ "${{ matrix.build }}" =~ ^(documentation)$ ]]; then
+            make download-cppreference-doxygen-web-tag; else
+            make -k -j2
+          fi
```
marehr, line 288:
```suggestion
          if [[ "${{ matrix.build }}" =~ ^(documentation)$ ]]; then
            make download-cppreference-doxygen-web-tag;
          else
            make -k -j2
          fi
```

I don't know bash enough, but is it possible to move the else to the next line?

eseiler: Should work, we can do else on a new line.

I'll make it a backlog item

eseiler: https://github.com/seqan/product_backlog/issues/230

marehr: Further above uses 

```
        run: |
          if [ "$RUNNER_OS" == "Linux" ]; then
            sudo apt-get install --yes ${{ matrix.cxx }}
          else
            brew install --force-bottle $(echo "${{ matrix.cxx }}" | sed "s/++-/cc@/g")
          fi
```

not sure why we can't use it here if we introduce this.

marehr (approved):

smehringer (approved):

PR #2208, opened 2020-10-13
[FIX] only accept push events on master.

rrahn:


.github/workflows/ci.yml
```diff
@@ -3,8 +3,8 @@ name: SeqAn3 CI
 on:
   push:
     branches:
-      # Push events to branches matching refs/heads/seqan/seqan3
-      - 'seqan/seqan3'
+      # Push events to branches matching refs/heads/master
+      - 'master'
       # Push events to branches matching refs/heads/releases/*
       - 'releases/**'
```
eseiler, line 9: We names our release branch `release-3.0.1`,if we keep this naming scheme, we should adapt the regex?

eseiler (approved):

.github/workflows/ci.yml
rrahn: I changed it to `release*`. This should reflect the release branch nameing. At least for now.

smehringer (approved):

Issue #2209, opened 2020-10-13
search_benchmark test failure


sagitter:
Hi all.

## Platform

  - SeqAn version: 3.0.2
  - Operating system: Fedora 34 (devel branch)
  - Compiler: GCC-10.2.1
  - Build log: https://kojipkgs.fedoraproject.org//work/tasks/4611/53394611/build.log

## Description

`search_benchmark` test is failing:

```
+ ctest --force-new-ctest-process -j1 -VV --output-on-failure --debug -R search_benchmark .
/builddir/build/BUILD/cmake-3.18.3/Source/cmCTest.cxx:417 Here: 417
/builddir/build/BUILD/cmake-3.18.3/Source/cmCTest.cxx:703 UpdateCTestConfiguration  from :/builddir/build/BUILD/seqan3-3.0.2-Source/build-performance/DartConfiguration.tcl
/builddir/build/BUILD/cmake-3.18.3/Source/cmCTest.cxx:429 Here: 429
/builddir/build/BUILD/cmake-3.18.3/Source/cmCTest.cxx:2566 * Read custom CTest configuration directory: /builddir/build/BUILD/seqan3-3.0.2-Source/build-performance
/builddir/build/BUILD/cmake-3.18.3/Source/cmCTest.cxx:2571 * Check for file: /builddir/build/BUILD/seqan3-3.0.2-Source/build-performance/CTestCustom.cmake
/builddir/build/BUILD/cmake-3.18.3/Source/cmCTest.cxx:2591 * Check for file: /builddir/build/BUILD/seqan3-3.0.2-Source/build-performance/CTestCustom.ctest
/builddir/build/BUILD/cmake-3.18.3/Source/cmCTest.cxx:703 UpdateCTestConfiguration  from :/builddir/build/BUILD/seqan3-3.0.2-Source/build-performance/DartConfiguration.tcl
/builddir/build/BUILD/cmake-3.18.3/Source/CTest/cmCTestTestHandler.cxx:391 Test project /builddir/build/BUILD/seqan3-3.0.2-Source/build-performance
/builddir/build/BUILD/cmake-3.18.3/Source/CTest/cmCTestTestHandler.cxx:1714 Constructing a list of tests
/builddir/build/BUILD/cmake-3.18.3/Source/CTest/cmCTestTestHandler.cxx:2357 Add test: alignment/global_affine_alignment_benchmark
[...]

PR #2211, opened 2020-10-14
[FIX] search_benchmark test failure (#2209)

marehr:
This PR fixes https://github.com/seqan/seqan3/issues/2209. 

The main problem was that constructing the following with `stddev = 0` results in undefined behaviour. 

```c++
std::normal_distribution<> dis_error_count{static_cast<double>(simulated_errors), stddev};
```

The code was written in a way that it never executed `dis_error_count` while `stddev = 0`, so we never encountered any real problems.

### DIFF
--- a/test/performance/search/search_benchmark.cpp
+++ b/test/performance/search/search_benchmark.cpp
@@ -64,7 +64,14 @@ std::vector<std::vector<alphabet_t>> generate_reads(std::vector<alphabet_t> cons
     std::vector<std::vector<alphabet_t>> reads;
     std::mt19937_64 gen{seed};
 
-    std::normal_distribution<> dis_error_count{static_cast<double>(simulated_errors), stddev};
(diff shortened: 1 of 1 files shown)


simonsasse (approved):
👍

test/performance/search/search_benchmark.cpp
```diff
@@ -64,7 +64,14 @@ std::vector<std::vector<alphabet_t>> generate_reads(std::vector<alphabet_t> cons
     std::vector<std::vector<alphabet_t>> reads;
     std::mt19937_64 gen{seed};
 
-    std::normal_distribution<> dis_error_count{static_cast<double>(simulated_errors), stddev};
+    auto error_count_distribution = [simulated_errors, stddev](auto & generator) -> uint8_t
```
rrahn, line 67:
```suggestion
    auto error_count_distribution = [simulated_errors, stddev] (auto & generator) -> uint8_t
```

rrahn (approved):
looks good. Only some minor style change.

PR #2215, opened 2020-10-14
[INFRA] Unify if/then/else syntax

eseiler:
Resolves https://github.com/seqan/product_backlog/issues/230

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -293,15 +293,14 @@ jobs:
           ccache -s || true
 
       - name: Run tests
+        if: matrix.build != 'coverage' # Already ran by the make call
         shell: bash -ex {0}
         run: |
           cd seqan3-build
-          if [[ "${{ matrix.build }}" =~ ^(coverage)$ ]]; then
-            : ; else
-            if [[ "${{ matrix.build }}" =~ ^(snippet)$ ]]; then
-              ctest . --output-on-failure]; else
-              ctest . -j2 --output-on-failure
-            fi
+          if [[ "${{ matrix.build }}" =~ ^(snippet)$ ]]; then
+            ctest . --output-on-failure
+          else
+            ctest . -j2 --output-on-failure
           fi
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -293,15 +293,14 @@ jobs:
...
-          if [[ "${{ matrix.build }}" =~ ^(coverage)$ ]]; then
-            : ; else
-            if [[ "${{ matrix.build }}" =~ ^(snippet)$ ]]; then
-              ctest . --output-on-failure]; else
-              ctest . -j2 --output-on-failure
-            fi
+          if [[ "${{ matrix.build }}" =~ ^(snippet)$ ]]; then
+            ctest . --output-on-failure
+          else
+            ctest . -j2 --output-on-failure
```
marehr, line 303: Is this the same? Because above the complete stuff will not be executed if `matrix.build = coverage`, or?

.travis.yml
```diff
@@ -96,9 +96,11 @@ script:
...
     if [[ "${BUILD}" =~ ^(coverage)$ ]]; then
-      : ; else
+      : # no-op
+    else
       if [[ "${BUILD}" =~ ^(snippet)$ ]]; then
-        ctest . --output-on-failure]; else
+        ctest . --output-on-failure
+      else
         ctest . -j2 --output-on-failure
       fi
```
marehr, line 105: here you kept the semantic?

.github/workflows/ci.yml
eseiler: exactly, I think it makes sense to just not execute the step if it is a no-op

.travis.yml
eseiler: Yes, travis is just a bash script, so there are no steps I can constraint by the build type (if that's what you meant)

marehr (approved):
nice; thank you!

smehringer (approved):

Issue #2214, opened 2020-10-14
Segmentation fault when executing solution6.cpp


eaasna:
## Platform

  - SeqAn version: 3.0.0
  - Operating system: Ubuntu 18.0.5 
  - Compiler: GCC 7.5.0

### How to repeat the problem

Problem caused by `regex_validator`

```
//![file_validator]
    parser.add_positional_option(args.file_path, "Please provide a tab separated seasons file.",
                                 seqan3::regex_validator{".*seasons\\..+$"} | seqan3::input_file_validator{{"tsv"}} );
//![file_validator]
```

### Expected behaviour

Running example code

### Actual behaviour

Segmentation fault (core dumped)

marehr:
Hi @eaasna congratulation on you first issue :tada: 

Can you give me a bit more detail? Did you try to compile `doc/tutorial/argument_parser/solution6.cpp`? How did you do this? Did you do it this way:

```c++
cmake <seqan3-path>/test/snippet

make
```

Can you install `valgrind` (`sudo apt install valgrind`) and run `valgrind ./solution6_snippet` and post the complete output?

eaasna:
Hi @marehr yes I tried the `solution6.cpp` and the error appeared at execution not compilation

Here's the output for `valgrind ./parsing -H -s 2 --aggregate-by median data.tsv`

```
==11173== Memcheck, a memory error detector
==11173== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==11173== Using Valgrind-3.13.0 and LibVEX; rerun with -h for copyright info
==11173== Command: ./parsing -H -s 2 --aggregate-by median data_seasons.tsv
==11173== 
[SEQAN3 INFO] :: Thank you for using SeqAn!
[SEQAN3 INFO] :: Do you wish to register your app for update notifications?
[SEQAN3 INFO] :: Just send an email to support@seqan.de with your app name and version number.
[SEQAN3 INFO] :: If you don't wish to receive further notifications, set --version-check OFF.

==11173== Stack overflow in thread #1: can't grow stack to 0x1ffe801000
==11173== 
==11173== Process terminating with default action of signal 11 (SIGSEGV)
==11173==  Access not within mapped region at address 0x1FFE801FF8
==11173== Stack overflow in thread #1: can't grow stack to 0x1ffe801000
[...]

PR #2216, opened 2020-10-14
[FIX] fix regex validator for std::filesystem::path

smehringer:
Resolves https://github.com/seqan/seqan3/issues/2214

### DIFF
--- a/include/seqan3/argument_parser/validators.hpp
+++ b/include/seqan3/argument_parser/validators.hpp
@@ -929,7 +929,8 @@ public:
     //!\endcond
     void operator()(range_type const & v) const
     {
-         std::for_each(v.begin(), v.end(), [&] (auto cmp) { (*this)(cmp); });
+        for (option_value_type const & file_name : v)
+            (*this)(file_name);
     }
 
     //!\brief Returns a message that can be appended to the (positional) options help page info.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,8 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 * Argument parsing with subcommands: If the user forgets or misspells the subcommand, the error is thrown when calling
   `seqan3::argument_parser::parse()` and not on construction of the `seqan3::argument_parser`
   ([\#2179](https://github.com/seqan/seqan3/pull/2179)).
+* The `seqan3::regex_validator` parses `std::filesystem::path`'s correctly now
+  ([\#2216](https://github.com/seqan/seqan3/pull/2216)).
 
 # 3.0.2
 
--- a/test/unit/argument_parser/format_parse_validators_test.cpp
+++ b/test/unit/argument_parser/format_parse_validators_test.cpp
@@ -1059,6 +1059,20 @@ TEST(validator_test, regex_validator_success)
         EXPECT_EQ(option_vector[1], "tina@rambo.com");
     }
 
+    { // option - std::filesystem::path
+        std::filesystem::path path_option;
+        const char * argv[] = {"./argument_parser_test", "-s", "rita@rambo.com"};
+        seqan3::argument_parser parser{"test_parser", 3, argv, seqan3::update_notifications::off};
+        test_accessor::set_terminal_width(parser, 80);
+        parser.add_option(path_option, 's', "string-option", "desc",
+                          seqan3::option_spec::DEFAULT, email_vector_validator);
+
+        testing::internal::CaptureStderr();
+        EXPECT_NO_THROW(parser.parse());
+        EXPECT_TRUE((testing::internal::GetCapturedStderr()).empty());
+        EXPECT_EQ(path_option, "rita@rambo.com");
+    }
+
     { // get help page message
         option_vector.clear();
         const char * argv[] = {"./argument_parser_test", "-h"};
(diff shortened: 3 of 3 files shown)


eseiler:
So the problem was that `std::filesystem::path` models `forward_range` and that makes it use the wrong overload because it is more constraint than the `option_value_type` one?

And then weird stuff happened because it validated by iterating over the path and checking each increment?

smehringer:
> So the problem was that `std::filesystem::path` models `forward_range` and that makes it use the wrong overload because it is more constraint than the `option_value_type` one?

It's fine that it models forward_range but unfortunately the requires clause is also true, namely that the value_type of the range is convertible to std::string. The range value type of a `std::filesystem::path` is again a `std::filesystem::path`  :roll_eyes: 

> And then weird stuff happened because it validated by iterating over the path and checking each increment?

Yes, somehow it creates an infinite loop :thinking:

eseiler:
> Yes, somehow it creates an infinite loop 🤔

Shouldn't the loop happen on compiletime, because that's where the types must be resolved.
I thought it will try to iterate, create some string, then something goes out of scope and then it segfaults...

smehringer:
> > Yes, somehow it creates an infinite loop thinking
> 
> Shouldn't the loop happen on compiletime, because that's where the types must be resolved.
> I thought it will try to iterate, create some string, then something goes out of scope and then it segfaults...

No it compiles fine but the infinite loop is created when iterating over the file path itself. I don't know what the specs say but I think this will just return the path itself for ever and ever

eseiler:
> No it compiles fine but the infinite loop is created when iterating over the file path itself. I don't know what the specs say but I think this will just return the path itself for ever and ever

Ah yes, it returns the path itself. I thought that you meant that the function call is recursive before.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -925,11 +925,12 @@ class regex_validator
...
     template <std::ranges::forward_range range_type>
     //!\cond
-        requires std::convertible_to<std::ranges::range_value_t<range_type>, option_value_type const &>
+        requires std::convertible_to<std::ranges::range_value_t<range_type>, option_value_type const &> &&
+                 (!std::same_as<range_type, std::filesystem::path>)
     //!\endcond
     void operator()(range_type const & v) const
     {
-         std::for_each(v.begin(), v.end(), [&] (auto cmp) { (*this)(cmp); });
+        std::for_each(v.begin(), v.end(), [&] (auto cmp) { (*this)(cmp); });
```
marehr, line 933: You require `std::convertible_to<std::ranges::range_value_t<range_type>, option_value_type const &>` above, but you don't use that property anywhere in here.

https://en.cppreference.com/w/cpp/concepts/convertible_to says:

```c++
template <class From, class To>

concept convertible_to =
  std::is_convertible_v<From, To> &&
  requires(std::add_rvalue_reference_t<From> (&f)()) {
    static_cast<To>(f());
  };
```

so shouldn't you call the function like this?

```suggestion
        std::for_each(v.begin(), v.end(), [&] (auto cmp)
        {
            (*this)(static_cast<option_value_type const &>(cmp));
        });
```

?

include/seqan3/argument_parser/validators.hpp
```diff
@@ -925,11 +925,12 @@ class regex_validator
      */
     template <std::ranges::forward_range range_type>
     //!\cond
-        requires std::convertible_to<std::ranges::range_value_t<range_type>, option_value_type const &>
+        requires std::convertible_to<std::ranges::range_value_t<range_type>, option_value_type const &> &&
+                 (!std::same_as<range_type, std::filesystem::path>)
```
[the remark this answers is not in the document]
eseiler: The general problem is:

```
    !std::same_as<std::remove_cvref_t<std::ranges::range_reference_t<rng_t>>,
                                std::remove_cvref_t<rng_t>> && // prevent recursive instantiation
```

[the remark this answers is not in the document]
marehr: Okay, I would prefer this version then :)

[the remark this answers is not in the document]
smehringer:
```suggestion
                 (!std::same_as<std::ranges::range_value_t<rng_t>, rng_t>) // prevent recursive instantiation
```

include/seqan3/argument_parser/validators.hpp
smehringer: Yes that's better :)

marehr (approved):
thank you :+1:

include/seqan3/argument_parser/validators.hpp
rrahn: Well, despite making this more readable, convetible_to requires implicit conversion. So before it was fine as well.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -929,7 +929,10 @@ class regex_validator
     //!\endcond
     void operator()(range_type const & v) const
     {
-         std::for_each(v.begin(), v.end(), [&] (auto cmp) { (*this)(cmp); });
+        std::for_each(v.begin(), v.end(), [&] (auto && cmp)
+        {
+            (*this)(static_cast<option_value_type const &>(cmp));
+        });
```
rrahn, line 935:
```suggestion
        for (option_value_type const & file_name : v)
            (*this)(file_name);
```
?

include/seqan3/argument_parser/validators.hpp
marehr: Not completely sure what you mean: here was the problem that this called recursively the same `operator()`, because `void operator()(range_type const & v) const` was more specific than the version of `operator()` that required the implicit cast.

include/seqan3/argument_parser/validators.hpp
marehr: This would revert the actual fix xD

marehr: This would hide the cast, but I'm fine with this version too.

rrahn: It says so in the docs 🤷

rrahn (approved):
thanks :shipit:

PR #2217, opened 2020-10-14
[INFRA] header test now tests self-includes

marehr:
Fixes https://github.com/seqan/product_backlog/issues/232

This PR implements a new `test/header/generate_header_source.cmake` which generates for each given header file (e.g. `include/seqan3/search/fm_index/bi_fm_index_cursor.hpp`) two `.cpp` files.

* `<build_dir>/seqan3_header_test/seqan3/search/fm_index/bi_fm_index_cursor.hpp-header-guard.cpp`
  which checks if the header implements a header guard.
  This will be checked by including the header twice, if the header guard is missing the second include will produce redeclaration errors, i.e.
  ```c++
  #include <seqan3/search/fm_index/bi_fm_index_cursor.hpp>
  #include <seqan3/search/fm_index/bi_fm_index_cursor.hpp>
  ```
* `<build_dir>/seqan3_header_test/seqan3/search/fm_index/bi_fm_index_cursor.hpp-no-self-include.cpp`
  which checks if the header will not be included itself (cyclic include). For this we copy the content of the header file into the source file and if any included header includes the header itself again, we would get redeclaration errors, i.e.
  ```c++
  // ....
  #pragma once

  #include <array>

  #include <sdsl/suffix_trees.hpp>

  #include <seqan3/alphabet/adaptation/char.hpp>
  #include <seqan3/alphabet/adaptation/uint.hpp>
  #include <seqan3/core/type_traits/range.hpp>
  #include <seqan3/search/fm_index/fm_index.hpp>
  #include <seqan3/search/fm_index/fm_index_cursor.hpp>
  #include <seqan3/std/ranges>

  namespace seqan3
  // ...
  ```

### DIFF
--- a/test/header/CMakeLists.txt
+++ b/test/header/CMakeLists.txt
@@ -12,11 +12,10 @@ include (../seqan3-test.cmake)
 
 option (SEQAN3_FULL_HEADER_TEST "Test seqan3 headers as well as the headers of external libraries" OFF)
 
-# We compile each header twice in separate compilation units. Each alone is
-# sufficient to test that the header is functional, but both are needed to check
-# for link errors, which can happen if the header accidentally defines a
-# variable, e.g. a global or class static member. Furthermore this tests that
-# header guards are working by including the same header twice.
+# We compile each header twice in separate compilation units, where the second compilation omits the header guard (check
+# for cyclic includes). Each alone is sufficient to test that the header is functional, but both are needed to check for
+# link errors, which can happen if the header accidentally defines a variable, e.g. a global or class static member.
+# Furthermore this tests that header guards are working by including the same header twice.
 #
 # example invocation:
 #     seqan3_header_test (seqan3 "<path>/include/seqan3" "test.hpp|/folder/|test2.hpp")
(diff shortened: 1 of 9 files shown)


joergi-w (changes requested):
The CMake changes look fine! CI reports an issue with a missing include: The FM index unit tests seem to use also the biFM index.

marehr:
@joergi-w I added a more detailed explanation, can you have a second look. I fixed the tests too. :)

test/header/CMakeLists.txt
```diff
@@ -28,6 +27,53 @@ option (SEQAN3_FULL_HEADER_TEST "Test seqan3 headers as well as the headers of e
...
+#   #include <seqan3/alphabet/concept.hpp>
+#   #include <seqan3/core/type_traits/range.hpp>
+#   #include <seqan3/range/views/slice.hpp>
+#   #include <seqan3/search/fm_index/fm_index.hpp> // redeclaration errors if this includes bi_fm_index_cursor.hpp
+#   #include <seqan3/search/fm_index/fm_index_cursor.hpp>
+#   #include <seqan3/std/ranges>
+#
+#   namespace seqan3
+#   // ...
+#   ```
```
joergi-w, line 73: This snippet could be much shorter to make your point clear. In fact, it is rather misleading if there are so many includes. And the `#pragma once` directive must be omitted in order to see the redeclaration errors?!

I think this would be enough:
```cpp
// This is file bi_fm_index_cursor.hpp

// ...
#include <seqan3/search/fm_index/fm_index.hpp> // redeclaration errors if fm_index.hpp includes bi_fm_index_cursor.hpp
```

joergi-w (approved):
Nice, thank you! I have one idea to make the snippet more comprehensive:

joergi-w:
Please avoid force-pushes between reviews, because now it was impossible to comment a line while showing the differences.

test/header/CMakeLists.txt
marehr: Changed :)

test/header/CMakeLists.txt
```diff
@@ -12,11 +12,10 @@ include (../seqan3-test.cmake)
 
 option (SEQAN3_FULL_HEADER_TEST "Test seqan3 headers as well as the headers of external libraries" OFF)
 
-# We compile each header twice in separate compilation units. Each alone is
-# sufficient to test that the header is functional, but both are needed to check
-# for link errors, which can happen if the header accidentally defines a
-# variable, e.g. a global or class static member. Furthermore this tests that
-# header guards are working by including the same header twice.
+# We compile each header twice in separate compilation units, where the second compilation omits the header guard (check
```
smehringer, line 15: This will make the header tests take even longer right? Not that I'm against this check but just  thinking of the long CI times.

smehringer (approved):
I am really not the best person to judge CMake Code but it looks good to me.

test/header/CMakeLists.txt
rrahn: Mhmm, but the header CI tests are among the first that finish because of ccache. At least I did not need to wait for header tests but rather the OS machines being spun up. Here it says it was successfull in 2 minutes, which isn't too bad.

PR #2219, opened 2020-10-14
[MISC] optimisations and clean-up

h-2:
This PR does the following:

* use our fast streambuf iterators instead of regular ones
* avoid repeated calls to `std::ranges::begin()` (instead creates iterator once and uses that)
* replaces some very long `if`-`if else`-chains with `switch`-`case`.

This improves performance of my real-world use-case by at least 25%.

See also https://github.com/seqan/product_backlog/issues/227

### DIFF
--- a/include/seqan3/io/alignment_file/format_bam.hpp
+++ b/include/seqan3/io/alignment_file/format_bam.hpp
@@ -38,6 +38,7 @@
 #include <seqan3/io/detail/ignore_output_iterator.hpp>
 #include <seqan3/io/detail/misc.hpp>
 #include <seqan3/range/detail/misc.hpp>
+#include <seqan3/range/views/istreambuf.hpp>
 #include <seqan3/range/views/slice.hpp>
 #include <seqan3/range/views/take_exactly.hpp>
 #include <seqan3/range/views/take_until.hpp>
@@ -313,9 +314,7 @@ inline void format_bam::read_alignment_record(stream_type & stream,
     static_assert(detail::decays_to_ignore_v<flag_type> || std::same_as<flag_type, sam_flag>,
                   "The type of field::flag must be seqan3::sam_flag.");
 
-    using stream_buf_t = std::istreambuf_iterator<typename stream_type::char_type>;
-    auto stream_view = std::ranges::subrange<decltype(stream_buf_t{stream}), decltype(stream_buf_t{})>
-                           {stream_buf_t{stream}, stream_buf_t{}};
+    auto stream_view = seqan3::views::istreambuf(stream);
 
     // these variables need to be stored to compute the ALIGNMENT
     [[maybe_unused]] int32_t offset_tmp{};
@@ -373,7 +372,7 @@ inline void format_bam::read_alignment_record(stream_type & stream,
 
         header_was_read = true;
 
-        if (stream_buf_t{stream} == stream_buf_t{}) // no records follow
+        if (std::ranges::begin(stream_view) == std::ranges::end(stream_view)) // no records follow
             return;
     }
 
@@ -695,7 +694,7 @@ inline void format_bam::write_alignment_record([[maybe_unused]] stream_type &  s
         if (ref_offset.has_value() && (ref_offset.value() + 1) < 0)
             throw format_error{detail::to_string("The ref_offset object must be >= -1 but is: ", ref_offset)};
 
-        std::cpp20::ostreambuf_iterator stream_it{stream};
+        detail::fast_ostreambuf_iterator stream_it{*stream.rdbuf()};
 
         // ---------------------------------------------------------------------
         // Writing the BAM Header on first call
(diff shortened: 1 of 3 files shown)


smehringer:
Since I am second reviewer I am delegating this to someone else beforehand.

marehr:
```c++
/home/runner/work/seqan3/seqan3/seqan3/include/seqan3/io/alignment_file/format_sam.hpp:1083: warning: argument 'stream' of command @param is not found in the argument list of seqan3::format_sam::write_tag_fields(stream_it_t &stream_it, sam_tag_dictionary const &tag_dict, char const separator)
/home/runner/work/seqan3/seqan3/seqan3/include/seqan3/io/alignment_file/format_sam.hpp:281: warning: The following parameter of seqan3::format_sam::write_tag_fields(stream_it_t &stream_it, sam_tag_dictionary const &tag_dict, char const separator) is not documented:
  parameter 'stream_it'
```

documentation fails

h-2:
> documentation fails

Yep, I saw that. It's just a minor thing. I hope to have time for it after work today.

marehr:
@h-2 It would be nice if you could split your PR into smaller chunks. Here are at least x different optimisations that you applied and each would be very happy to be in its own commit.

h-2:
>  documentation fails

I think I fixed it, but I locally get many more warnings on the documentation test, so it's hard to tell. I will look into that separately.

> It would be nice if you could split your PR into smaller chunks. Here are at least x different optimisations that you applied and each would be very happy to be in its own commit.

I promise to try and do this in the future, but it would take a lot of (my unpaid) time to pull these things into separate commits now since many of the same lines are touched by everything. Also, I think that the changes are still reasonably well-connected and the entire PR is not that huge -- but I know people disagree on these things. 

In general, I am also not at all offended if people push changes onto my branch before merging to satisfy the latest style 😉

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -373,7 +372,7 @@ inline void format_bam::read_alignment_record(stream_type & stream,
 
         header_was_read = true;
 
-        if (stream_buf_t{stream} == stream_buf_t{}) // no records follow
+        if (std::ranges::begin(stream_view) == std::ranges::end(stream_view)) // no records follow
```
eseiler, line 375:
```suggestion
        if (std::ranges::empty(stream_view)) // no records follow
```
?

include/seqan3/io/alignment_file/format_sam.hpp
```diff
@@ -793,49 +790,52 @@ inline void format_sam::write_alignment_record(stream_type & stream,
...
 
-    stream << static_cast<unsigned>(mapq) << separator;
+    stream_it.write_number(static_cast<unsigned>(mapq));
+    *stream_it = separator;
 
     if (!std::ranges::empty(cigar_vector))
     {
-        for (auto & c : cigar_vector)
-            stream << c.to_string(); // returns a small_vector instead of char so write_range doesn't work
+        for (auto & c : cigar_vector) //TODO THIS IS PROBABLY TERRIBLE PERFORMANCE_WISE
```
eseiler, line 837: Did you see/measure any performance impact or does it still need to be confirmed?
Or, do you already have any idea what could be done differently?

include/seqan3/io/alignment_file/format_sam_base.hpp
```diff
@@ -447,165 +447,175 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
+                    else if (!is_char<'G'>(*it))                        // GO (grouping) tag
+                    {
+                        throw format_error{std::string{"Illegal SAM header tag in @HG starting with: "} + *it};
+                    }
+
+                    parse_tag_value(*who);
                 }
-                else if (!is_char<'G'>(*std::ranges::begin(stream_view))) // GO (grouping) tag
+                ++it;                                                   // skip newline
+            } break;
```
eseiler, line 495: This `break` still belongs to the `case`?
If so, move the `break` into the scope that just ended, to make that clear?

(Same for other cases)

include/seqan3/io/alignment_file/format_sam_base.hpp
```diff
@@ -447,165 +447,175 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
-                hdr.ref_dict[(hdr.ref_ids())[(hdr.ref_ids()).size() - 1]] = (hdr.ref_ids()).size() - 1;
-            }
-        }
-        else if (is_char<'R'>(*std::ranges::begin(stream_view)))         // RG (read group) tag
-        {
-            std::pair<std::string, std::string> tmp{};
+                parse_tag_value(tmp.id);                                // read required ID tag
 
-            parse_tag_value(get<0>(tmp));                                // read required ID tag
+                // The PN, CL, PP, DS, VN are optional tags and can be given in any order.
```
eseiler, line 565: As a side note, what happens if I pass `VR` or something as tag?
Because it seems that we only check for `N` but not the second character?

include/seqan3/io/alignment_file/format_sam.hpp
h-2: > Did you see/measure any performance impact or does it still need to be confirmed?

It's just an intuition. The string returned is stack-allocated, so it's not terribly bad, but it still means we are unpacking every single element into one string and than copying that string into another.

> Or, do you already have any idea what could be done differently?

A private helper function could remedy this (basically doing what `.to_string()`does but writing directly into the target).

I have quite a few other things/ideas regarding the SAM format handling, I just didn't want us to forget this.

include/seqan3/io/alignment_file/format_sam_base.hpp
h-2: Hm, I always put it behind the block to indicate that the `break` belongs to the `case` (which is also outside the `{ ... } `). And having the break directly before the subsequent `case` clearly indicates that there is no fallthrough.

But it's really not that important to me :)

include/seqan3/io/alignment_file/format_sam_base.hpp
h-2: There are quite a few places where the second letter is not checked when it is assumed to be unambiguous. I was also a little surprised by this, and I guess it is something we could change.
But it is unrelated to the other changes in this PR.

include/seqan3/io/alignment_file/format_sam_base.hpp
eseiler: That's fair, I was just really confused why you end the scope and then put the `break`

eseiler (approved):

include/seqan3/io/alignment_file/format_sam_base.hpp
smehringer: I think I did not want to include to many `if` checks for the second character for performance reasons. We did not decide even yet whether we want to be very strict on format checking (e.g. samtools is not at all and swallows a lot).

smehringer (approved):

PR #2220, opened 2020-10-15
[FIX] Non boolean option in cmake

eseiler:
When not set, `SEQAN3_BENCHMARK_MIN_TIME ` will be evaluated to `ON` and then passed as `--benchmark_min_time=ON`


marehr (approved):
That's stupid that you can only set boolean defaults :(

smehringer (approved):

PR #2221, opened 2020-10-15
[INFRA] ci-skip

eseiler:


marehr:
I've seen that @smehringer requested this, but do we really need this?

eseiler:
> I've seen that @smehringer requested this, but do we really need this?

It's a rather standard feature of other ci provider and I for one almost never do `commit suggestions` because I don't want to manually cancel the CI

PR #2223, opened 2020-10-15
[MISC] Make config_element_specialisation more restrictive.

rrahn:
Now the config_element_specialisation checks if the class inherits from the pipeable_config_element base class.

First commit of #2198

### DIFF
--- a/include/seqan3/core/algorithm/concept.hpp
+++ b/include/seqan3/core/algorithm/concept.hpp
@@ -36,16 +36,45 @@ namespace seqan3::detail
 // Concept config_element_specialisation
 // ----------------------------------------------------------------------------
 
-//!\cond
-template <typename derived_t>
-class config_element_base;
-//!\endcond
-
-/*!\interface seqan3::detail::config_element_specialisation <>
- * \brief Concept for an algorithm configuration.
+#ifdef SEQAN3_WORKAROUND_GCC_PIPEABLE_CONFIG_CONCEPT
+/*!\brief A helper class to check if a type has a static member called `id`.
  * \ingroup algorithm
  *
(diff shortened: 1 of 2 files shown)


rrahn:
@marehr here you go.

marehr (approved):

include/seqan3/core/algorithm/concept.hpp
```diff
@@ -60,10 +89,15 @@ class config_element_base;
 //!\}
 //!\cond
 template <typename config_t>
-SEQAN3_CONCEPT config_element_specialisation = std::semiregular<std::remove_reference_t<config_t>> &&
-requires (config_t c)
+SEQAN3_CONCEPT config_element_specialisation = requires
 {
-    { std::remove_reference_t<config_t>::id };
+    requires std::is_base_of_v<seqan3::pipeable_config_element<config_t>, config_t>;
+    requires std::semiregular<config_t>;
```
smehringer, line 95: Why did you omit the `std::remove_reference_t` when testing semi regularity?

smehringer (changes requested):
Just one question, otherwise LGTM

include/seqan3/core/algorithm/concept.hpp
rrahn: This concept asks whether a given type is an instance of a pipebale config element. Since this requires semiregularity, it would be wrong to strip away references and cv-qualifiers, because otherwise the caller using a forwarding reference in a function template would think he could call `t{}` which he can't if it is an l-value reference. Hence, it is the responsibility of the caller to ensure that the references are stripped.

smehringer (approved):

PR #2224, opened 2020-10-16
[MISC] Move the  compatibility_table  declaration into the algorithm/concept header.

rrahn:
This declaration is needed later for the config_element_pipeable_with concept defintion.

second commit of #2198

### DIFF
--- a/include/seqan3/core/algorithm/concept.hpp
+++ b/include/seqan3/core/algorithm/concept.hpp
@@ -33,6 +33,18 @@ namespace seqan3::detail
 {
 
 // ----------------------------------------------------------------------------
+// compatibility_table
+// ----------------------------------------------------------------------------
+
+/*!\brief Declaration of algorithm specific compatibility table.
+ * \ingroup algorithm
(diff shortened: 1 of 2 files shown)


include/seqan3/core/algorithm/concept.hpp
```diff
@@ -32,6 +32,18 @@ struct pipeable_config_element;
...
+// compatibility_table
+// ----------------------------------------------------------------------------
+
+/*!\brief Declaration of algorithm specific compatibility table.
+ * \ingroup algorithm
+ * \tparam algorithm_id_type The type of the algorithm specific id. Algorithm configurations must maintain
+ *                           this table to allow validation checks.
+ */
+template <typename algorithm_id_type>
+inline constexpr std::array<std::array<void, 0>, 0> compatibility_table;
```
marehr, line 45: When working with clang it showed that `void` is an invalid type for array, because `reference` is defined as `value_type&` and that is an invalid type.

But we can fix that in a follow up PR.

marehr (approved):

smehringer (approved):

Issue #2225, opened 2020-10-16
format_sam_benchmark not terminating in Debug mode


eseiler:
## Platform

  - SeqAn version: master
  - Operating system: Ubuntu 20.04
  - Compiler: GCC 7.5/8.4/9.3/10.2

## Description

Infinite(?) loop(?).

Git bisect says #1985 is the first bad commit, though I do net yet see how...
Debugger points to this assert (the evaluation of the `std::distance` does not finish):
https://github.com/seqan/seqan3/blob/8c43f647f137fd7ca4aeee53f2403eb7b94b7a82/include/seqan3/io/alignment_file/detail.hpp#L360
However, removing the assert results in some failed assertion in the gap_decorator (and the program actually terminates).

If it helps:
![Capture](https://user-images.githubusercontent.com/12967715/96262374-fe086580-0fc1-11eb-9ed3-a9bddf72398e.PNG)

### How to repeat the problem

Compile and run `format_sam_benchmark` in Debug mode.

### Expected behaviour

Terminate after a few seconds.

### Actual behaviour

Terminate after a few aeons.

rrahn:
Can you reduce this to a specific sequence that causes this issue and then make a separate test for just this offending sequence?

eseiler:
I'll try to reduce it. The problem is definitely in the generate reference+query and align it section... `100M400I4294966896S` is not really a valid cigar for 100query vs 500 reference.

eseiler:
I reduced it somewhat
```cpp
#include <seqan3/alignment/pairwise/align_pairwise.hpp>
#include <seqan3/core/debug_stream.hpp>
#include <seqan3/io/alignment_file/input.hpp>
#include <seqan3/io/alignment_file/output.hpp>
#include <seqan3/test/performance/sequence_generator.hpp>
#include <seqan3/test/tmp_filename.hpp>

using seqan3::operator""_dna4;

int main()
{
    std::vector<seqan3::dna4> reference{"ATTATTTATTACTATAAAGATCTAGGAATTAGGGCCGGTATTTCCCAGTTGAGAACAATGGGCGATCTACCACTCAGA"
                                        "CGCGGCGTCACCCATTAATGCGC"_dna4};
[...]

PR #2226, opened 2020-10-16
[FIX] wrong usage of SAM interface

eseiler:
Resolves https://github.com/seqan/seqan3/issues/2225
(yes, it works in debug mode)

### DIFF
--- a/test/performance/io/format_sam_benchmark.cpp
+++ b/test/performance/io/format_sam_benchmark.cpp
@@ -63,15 +63,15 @@ static std::string create_sam_file_string(size_t const n_queries)
         {
             auto query = seqan3::test::generate_sequence<seqan3::dna4>(read_size, length_variance, seed + i);
             auto qualities = seqan3::test::generate_sequence<seqan3::phred42>(read_size, length_variance, seed + i);
-            auto align_result = *(seqan3::align_pairwise(std::tie(query, reference), config).begin());
+            auto align_result = *(seqan3::align_pairwise(std::tie(reference, query), config).begin());
             std::string const current_query_id = query_prefix + std::to_string(i);
 
             sam_out.emplace_back(query,                                   // field::seq
                                  current_query_id,                        // field::id
-                                 align_result.sequence1_begin_position(), // field::offset
+                                 align_result.sequence2_begin_position(), // field::offset
                                  reference,                               // field::ref_seq
                                  reference_id,                            // field::ref_id
-                                 align_result.sequence2_begin_position(), // field::ref_offset
+                                 align_result.sequence1_begin_position(), // field::ref_offset
                                  align_result.alignment(),                // field::alignment
                                  align_result.score(),                    // field::mapq
                                  qualities,                               // field::qual
--- a/include/seqan3/io/alignment_file/format_sam.hpp
+++ b/include/seqan3/io/alignment_file/format_sam.hpp
@@ -847,6 +847,9 @@ inline void format_sam::write_alignment_record(stream_type & stream,
         for (auto chr : get<1>(align))
             if (chr == gap{})
                 ++off_end;
+
+        // Might happen if get<1>(align) doesn't correspond to the reference.
+        assert(off_end >= std::ranges::size(get<1>(align)));
         off_end -= std::ranges::size(get<1>(align));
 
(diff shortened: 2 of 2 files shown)


simonsasse (approved):
looks good 👍

marehr:
@eseiler Is there a lesson learned, why and how this could have been circumvented and/or a user could have made the same error?

eseiler:
> @eseiler Is there a lesson learned, why and how this could have been circumvented and/or a user could have made the same error?

Maybe! I would need some feedback (@smehringer). If this is public and a valid case for how to write/generate a SAM file, then we should somehow ensure that either the input for the alignment is correct `(reference, query)` or that we handle the `(query, reference)` case internally (though, how would you detect which is which).

Anyway, this might be part of the I/O refactoring - the interface might just change in a way that this cannot happen any more.
I.e., we should think about this error, but I don't think it's blocking this PR.

test/performance/io/format_sam_benchmark.cpp
```diff
@@ -63,15 +63,15 @@ static std::string create_sam_file_string(size_t const n_queries)
...
             std::string const current_query_id = query_prefix + std::to_string(i);
 
             sam_out.emplace_back(query,                                   // field::seq
                                  current_query_id,                        // field::id
-                                 align_result.sequence1_begin_position(), // field::offset
+                                 align_result.sequence2_begin_position(), // field::offset
                                  reference,                               // field::ref_seq
                                  reference_id,                            // field::ref_id
-                                 align_result.sequence2_begin_position(), // field::ref_offset
+                                 align_result.sequence1_begin_position(), // field::ref_offset
```
smehringer, line 74: Ah now I see. So the actual fix is that the leading and trailing gaps were configured wrongly for `(query, reference)` but are now correct with `(reference, query)`. The resulting (maybe weird) alignment caused the error when writing SAM.
Do I understand this correctly? Because then this just circumvents the error and does not fix it?

eseiler: Well, I have to find the line of code, but the problem is that inferring the CIGAR from the alignment in the SAM output implicitly requires the first sequence to be the reference

https://github.com/seqan/seqan3/blob/master/include/seqan3/io/alignment_file/format_sam.hpp#L840-L853

`<1>` is accessed, and IIRC, `<1>` needs to be the query.

marehr:
> > @eseiler Is there a lesson learned, why and how this could have been circumvented and/or a user could have made the same error?
> 
> Maybe! I would need some feedback (@smehringer). If this is public and a valid case for how to write/generate a SAM file, then we should somehow ensure that either the input for the alignment is correct `(reference, query)` or that we handle the `(query, reference)` case internally (though, how would you detect which is which).

Thank you! I thought something along this line :)

> 
> Anyway, this might be part of the I/O refactoring - the interface might just change in a way that this cannot happen any more.
> I.e., we should think about this error, but I don't think it's blocking this PR.

Of course :)

test/performance/io/format_sam_benchmark.cpp
smehringer: Thanks for pointing out the code. 

Can you add the following assert before the size is subtracted from `off_end`:

```cpp
assert(off_end >= std::ranges::size(get<1>(align)); // Might happen if get<1>(align) doesn't correspond to the reference.
```

I even think that this might be a case were we should throw but another if branch would of course impact the performance. I would add the assert for now and re-evaluate this once we have a stable IO interface and start working on the performance.

smehringer:
> Maybe! I would need some feedback (@smehringer). If this is public and a valid case for how to write/generate a SAM file, then we should somehow ensure that either the input for the alignment is correct `(reference, query)` or that we handle the `(query, reference)` case internally (though, how would you detect which is which).

We cannot directly prevent the wrong ordering because both the reference and the query will in most cases have the same time. What we can do is add an extra check. A check whether the CIGAR string or alignment is "correctly" representing the two sequences. We would need to compute the sequence size from the given CIGAR vector/alignment though which has a penalty on the runtime. This is of course in the current design. 

> Anyway, this might be part of the I/O refactoring - the interface might just change in a way that this cannot happen any more.
> I.e., we should think about this error, but I don't think it's blocking this PR.

So let's wait what we decide and meanwhile add only the assert I mentioned above.

smehringer (approved):

PR #2228, opened 2020-10-26
[INFRA] CMake version number check to be the minimum version

### LINKED ISSUE #2227: Using cmake 3.14 options but cmake require only 3.7
## Platform

  - SeqAn version: 3.0.2
  - Operating system: Linux 4.14.193-149.317.amzn2.x86_64 #1 SMP Thu Sep 3 19:04:44 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
  - Compiler: cpp (GCC) 7.3.1 20180712 (Red Hat 7.3.1-9)

## Description
The cmake file
`build_system/seqan3-install.cmake`
has
`cmake_minimum_required (VERSION 3.7)`

Though it also is using version 3.14 Features
```
install (
    FILES
    ...
    TYPE DOC
)
```
If cmake 3.14 features are being used then the cmake_minimum_required should be updated

### How to repeat the problem
Have cmake version 3.13 or below installed

### Expected behaviour
Install to work

### Actual behaviour
Error as install has no Destination

lhovo:
Fix CMake version number check to be the minimum version for the features that are in the CMakeFiles.
`install (
    FILES
    ....
    TYPE DOC
)`
The use of TYPE was released in cmake 3.14, and will fail on any prior version

Edit (@smehringer): Resolves #2227

### DIFF
--- a/build_system/seqan3-install.cmake
+++ b/build_system/seqan3-install.cmake
@@ -7,7 +7,7 @@
 
 # This file describes where and which parts of SeqAn3 should be installed to.
 
-cmake_minimum_required (VERSION 3.7)
+cmake_minimum_required (VERSION 3.14)
 
 include (GNUInstallDirs)
 


marehr (approved):
Thank you for your contribution :)

Minor note, our commit message style would be:

"[INFRA] CMake version number check to be the minimum version"

smehringer (approved):

PR #2230, opened 2020-10-30
Add coverage to embl sequence file test

MitraDarja:

By changing stuff to Stuff, the do while loop in the embl code runs more than once so the do is covered.


joergi-w (approved):
I see, it's because of the condition `is_char<'S'>`. Good fix, thank you!

smehringer (approved):

PR #2233, opened 2020-11-02
[TEST] Increase coverage for tmp_filename

eseiler:
Just changing the env variable is not enough, because `std::filesystem::temp_directory_path()` already checks if the returned path exists and is valid. 
That means the `std::filesystem::filesystem_error` was always thrown inside the `temp_directory_path()` instead of inside our `tmp_filename` constructor.

### DIFF
--- a/include/seqan3/std/filesystem
+++ b/include/seqan3/std/filesystem
@@ -9,9 +9,11 @@
 /*!\file
  * \brief This header includes C++17 filesystem support and imports it into namespace std::filesystem (independent of
  *        whether it is marked as "experimental").
- * \author Vitor C. Piro <pirov AT zedat.fu-berlin.de >
+ * \author Vitor C. Piro <pirov AT zedat.fu-berlin.de>
+ * \author Enrico Seiler <enricoseiler AT fu-berlin.de>
  */
 
+//!\cond
 #pragma once
 
 #if __has_include(<filesystem>)
@@ -19,9 +21,101 @@
 #else
 #include <experimental/filesystem>
 
+namespace std::experimental::filesystem
+{
+
+enum class perm_options : uint32_t
+{
+    replace = 1u,
+    add = 1u<<1,
+    remove = 1u<<2,
+    nofollow = 1u<<3
+};
+
+constexpr perm_options operator& (perm_options lhs, perm_options rhs) noexcept
+{
+    using underlying_t = std::underlying_type_t<perm_options>;
+    return static_cast<perm_options>(static_cast<underlying_t>(lhs) & static_cast<underlying_t>(rhs));
+}
+
+constexpr perm_options operator| (perm_options lhs, perm_options rhs) noexcept
+{
+    using underlying_t = std::underlying_type_t<perm_options>;
+    return static_cast<perm_options>(static_cast<underlying_t>(lhs) | static_cast<underlying_t>(rhs));
+}
+
+constexpr perm_options operator^ (perm_options lhs, perm_options rhs) noexcept
+{
+    using underlying_t = std::underlying_type_t<perm_options>;
+    return static_cast<perm_options>(static_cast<underlying_t>(lhs) ^ static_cast<underlying_t>(rhs));
+}
+
+constexpr perm_options operator~ (perm_options lhs) noexcept
+{
+    using underlying_t = std::underlying_type_t<perm_options>;
+    return static_cast<perm_options>(~static_cast<underlying_t>(lhs));
+}
+
+constexpr perm_options & operator&= (perm_options & lhs, perm_options rhs) noexcept
+{
+    lhs = lhs & rhs;
+    return lhs;
+}
+
+constexpr perm_options & operator|= (perm_options & lhs, perm_options rhs) noexcept
+{
+    lhs = lhs | rhs;
+    return lhs;
+}
+
+constexpr perm_options & operator^= (perm_options & lhs, perm_options rhs) noexcept
+{
+    lhs = lhs ^ rhs;
+    return lhs;
+}
+
+inline void permissions(path const & p,
+                        perms prms,
+                        perm_options opts,
+                        error_code & ec)
+{
+    auto follow_symlinks = (static_cast<bool>(opts & perm_options::nofollow)) ? perms::symlink_nofollow : perms::none;
+
+    switch (opts)
+    {
+        case perm_options::replace :
+            permissions(p, perms::remove_perms | follow_symlinks | perms::all, ec);
+            permissions(p, perms::add_perms | follow_symlinks | prms, ec);
+            break;
+        case perm_options::add :
+            permissions(p, perms::add_perms | follow_symlinks | prms, ec);
+            break;
+        case perm_options::remove :
+            permissions(p, perms::remove_perms | follow_symlinks | prms, ec);
+            break;
+        default :
+            break; // Actually UB
+    }
+}
+
+inline void permissions(path const & p,
+                        perms prms,
+                        perm_options opts)
+{
+    error_code ec;
+    permissions(p, prms, opts, ec);
+    if (ec.value())
+        throw filesystem_error("Cannot set permissions", p, ec);
+}
+
+} // namespace std::experimental::filesystem
+
 namespace std
 {
+
 namespace filesystem = std::experimental::filesystem;
+
 } // namespace std
 
 #endif // __has_include(experimental/filesystem)
+//!\endcond
--- a/test/unit/test/tmp_filename_test.cpp
+++ b/test/unit/test/tmp_filename_test.cpp
@@ -67,13 +67,90 @@ TEST(tmp_filename_dtr, dtr)
     EXPECT_FALSE(fs::exists(path.parent_path()));
 }
 
-// throw if invalid TMPDIR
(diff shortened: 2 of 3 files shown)


test/unit/test/tmp_filename_test.cpp
```diff
@@ -67,13 +67,25 @@ TEST(tmp_filename_dtr, dtr)
     EXPECT_FALSE(fs::exists(path.parent_path()));
 }
 
-// throw if invalid TMPDIR
+// throw if TMPDIR not writeable
```
eseiler, line 70: I would need to wrap this inside
```cpp
#if __has_include(<filesystem>)
// Setting the permissions with perm_options is not available in the experimental/filesystem branch.
```

In this case, I would mark the corresponding tested throw inside the `tmp_filename` as unreachable until we use at least gcc8 for coverage?

marehr: First of all, I find the new version definitely better!

But how do we solve the gcc7 fiasco?

```c++
// pre C++17 with filesystem TS
// see https://en.cppreference.com/w/cpp/experimental/fs/permissions
fs::permissions("test.txt", fs::perms::add_perms | fs::perms::owner_all | fs::perms::group_all);

// real C++17 with final filesystem API
// see https://en.cppreference.com/w/cpp/filesystem/permissions
fs::permissions("test.txt", fs::perms::owner_all | fs::perms::group_all, fs::perm_options::add);
```

Since the parameter changed from 2 to 3 inputs, we could add a new function `std::filesystem::permission` and `std::filesystem::perm_options` that supports this functionality.

eseiler: I added the functionality

marehr (approved):
lgtm

include/seqan3/std/filesystem
```diff
@@ -9,19 +9,123 @@
...
+
+constexpr perm_options operator^ (perm_options lhs, perm_options rhs) noexcept
+{
+    return static_cast<perm_options>(static_cast<int>(lhs) ^ static_cast<int>(rhs));
+}
+
+constexpr perm_options operator~ (perm_options lhs) noexcept
+{
+    return static_cast<perm_options>(~static_cast<int>(lhs));
+}
```
rrahn, line 53:
```suggestion
enum class perm_options : uint32_t
{
    replace = 1,
    add = 1<<1,
    remove = 1<<2,
    nofollow = 1<<3
};

constexpr perm_options operator& (perm_options lhs, perm_options rhs) noexcept
{
    using underlying_t = std::underlying_type_t<perm_options>;
    return static_cast<perm_options>(static_cast<underlying_t>(lhs) & static_cast<underlying_t>(rhs));
}

constexpr perm_options operator| (perm_options lhs, perm_options rhs) noexcept
{
    using underlying_t = std::underlying_type_t<perm_options>;
    return static_cast<perm_options>(static_cast<underlying_t>(lhs) | static_cast<underlying_t>(rhs));
}

constexpr perm_options operator^ (perm_options lhs, perm_options rhs) noexcept
{
    using underlying_t = std::underlying_type_t<perm_options>;
    return static_cast<perm_options>(static_cast<underlying_t>(lhs) ^ static_cast<underlying_t>(rhs));
}

constexpr perm_options operator~ (perm_options lhs) noexcept
{
    using underlying_t = std::underlying_type_t<perm_options>;
    return static_cast<perm_options>(~static_cast<underlying_t>(lhs));
}
```

test/unit/test/tmp_filename_test.cpp
```diff
@@ -67,13 +67,25 @@ TEST(tmp_filename_dtr, dtr)
...
+    // Set TMPDIR. This is the first env var that is looked up for `temp_directory_path` inside the `tmp_filename`.
+    setenv("TMPDIR", test_directory.c_str(), 1); // name, value, overwrite
+
+    EXPECT_THROW(seqan3::test::tmp_filename t1{"throw"}, std::filesystem::filesystem_error);
+
+    // Add write permissions (needed for deletion) and delete the test directory.
+    std::filesystem::permissions(test_directory,
+                                 std::filesystem::perms::owner_write,
+                                 std::filesystem::perm_options::add);
+    std::filesystem::remove(test_directory);
```
rrahn, line 90: Ok, this is not a trivial code block by far.
Here are some issues I detected: 

The return type of create_directory is not evaluated (a clear library issue by not declaring the code as [[nodiscard]]). Hence, what ever is tested next is not valid anymore.
Second, setting the permissions can throw. In this case if the first one throws, the resource (directory) is not removed. That can happen btw anytime inside of the code block.
Also the `remove` return type is not evaluated.

I suggest to put the creation of the directory and the handling of the file permissions in an explicit RAII wrapper class for this test only. It should have an explicit release function that returns the state of remove and tests that the result is true.
The RAII wrapper class should attempt to free the resource on destruction using the no-throw version.
The entire block is put in a try-catch block and the catch block always FAILS() with a dedicated error message and the hint that there is probably a leaking resource in the directory with the offending path name.

rrahn: Maybe it would also be good to ensure, that the directory did not exist before. Otherwise, other processes might get interrupted when this test removes the write permissions.

rrahn: So basically the RAII wrapper class throws as well on construction and the error is handled by the catch block making it a failing test not a terminated test.

rrahn (changes requested):
Thank you for the work. I have some general questions and some specific changes to make the test more robust. Cheers

include/seqan3/std/filesystem
```diff
@@ -9,19 +9,123 @@
...
+    try
+    {
+        permissions(p, prms, opts);
+        ec.clear();
+    }
+    catch (std::system_error const & e)
+    {
+        ec = e.code();
+    }
+}
```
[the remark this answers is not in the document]
eseiler: If we want to also have the error_code interface, we would have to
* duplicate all the code with (once with ec, once without) or
* let the non-ec version use the ec-version and 'rethrow' a non-0 ec as exception (if this is possible)

test/unit/test/tmp_filename_test.cpp
```diff
@@ -67,13 +67,97 @@ TEST(tmp_filename_dtr, dtr)
...
+    tmp_directory(tmp_directory const &) = delete; //< Deleted.
+    tmp_directory(tmp_directory &&) = default; //< Defaulted.
+    tmp_directory & operator=(tmp_directory const &) = delete; //< Deleted.
+    tmp_directory & operator=(tmp_directory &&) = default; //< Defaulted.
+
+    /*!\brief Construct a directory from a given path.
+     * \throws if std::filesystem::create_directory throws.
+     * \throws std::filesystem::filesystem_error if the directory already exists.
+     */
+    explicit tmp_directory(std::filesystem::path const & path) : directory_path(path)
```
rrahn, line 84:
```suggestion
    explicit tmp_directory(std::filesystem::path path) : directory_path(std::move(path))
```

test/unit/test/tmp_filename_test.cpp
```diff
@@ -67,13 +67,97 @@ TEST(tmp_filename_dtr, dtr)
...
+
+    /*!\brief Remove owner write permissions.
+     * \throws if std::filesystem::permissions throws.
+     */
+    void remove_owner_write_permission()
+    {
+        std::filesystem::permissions(directory_path,
+                                     std::filesystem::perms::owner_write,
+                                     std::filesystem::perm_options::remove);
+    }
```
rrahn, line 114: I would put that directly into the constructor. It is really just for this test.

test/unit/test/tmp_filename_test.cpp
```diff
@@ -67,13 +67,97 @@ TEST(tmp_filename_dtr, dtr)
...
     {
-        char str[] = "TMPDIR=/invalid";
-        putenv(str);
+        [[maybe_unused]] std::error_code ec;
+        // Add write permissions needed for deletion.
+        std::filesystem::permissions(directory_path,
+                                     std::filesystem::perms::owner_write,
+                                     std::filesystem::perm_options::add,
+                                     ec);
+        std::filesystem::remove(directory_path, ec);
```
rrahn, line 103: I would put this into a private function and share the code with the release function. Both call it with the error code but in addition the release function uses the error code to rethrow the exception if it was set.

test/unit/test/tmp_filename_test.cpp
```diff
@@ -67,13 +67,97 @@ TEST(tmp_filename_dtr, dtr)
     EXPECT_FALSE(fs::exists(path.parent_path()));
 }
 
-// throw if invalid TMPDIR
-TEST(tmp_filename_throw, throw)
+//!\brief A wrapper class to handle construction, destruction and permission handling of a temporary directory.
+class tmp_directory
```
rrahn, line 71:
```suggestion
class read_only_directory
```

rrahn (changes requested):
That looks much more safer now. Thank you very much! I have only some DRY code suggestions and a renaming suggestion. 😅

rrahn (approved):
lgtm. Thank you

rrahn:
can you rebase? Then we can merge it.

PR #2238, opened 2020-11-08
[TEST] Fix test naming scheme for the view tests

marehr:
Our naming scheme for our tests normally follows the scheme:

* a header [/include/seqan3]/module/submodule/some_class.hpp has a 
* corresponding test [/test/unit]/module/submodule/some_class_test.cpp

There are some exception to this rule, for example if the `some_class_test` can't be a unique target name (this is a limitation of cmake), but in general we use that scheme.

This change renames all tests (e.g. `views/view_as_const_test.cpp` to `views/as_const_test.cpp`) since the corresponding header are named without a `view_` prefix (e.g. `views/as_const.hpp`).

Part of https://github.com/seqan/product_backlog/issues/257


simonsasse (approved):
Look good 👍

smehringer (approved):

PR #2247, opened 2020-11-09
[FEATURE] Add config_element_pipeable_with concept.

rrahn:
This concept is used generically to test if two given configurations are combinable with each other.
With this concept the pipe-operator can be implemented without converting the configuration elements to the base class.

third commit of #2198

### DIFF
--- a/include/seqan3/core/algorithm/concept.hpp
+++ b/include/seqan3/core/algorithm/concept.hpp
@@ -42,7 +42,7 @@ namespace seqan3::detail
  *                           this table to allow validation checks.
  */
 template <typename algorithm_id_type>
-inline constexpr std::array<std::array<void, 0>, 0> compatibility_table;
+inline constexpr std::array<std::array<void *, 0>, 0> compatibility_table;
 
 // ----------------------------------------------------------------------------
 // Concept config_element_specialisation
@@ -61,6 +61,10 @@ inline constexpr std::array<std::array<void, 0>, 0> compatibility_table;
 struct config_id_accessor
 {
 private:
+    //!\brief Helper variable template to convert the configuration enum identifier to an integer.
+    template <typename config_t>
+    static constexpr int32_t as_int = static_cast<int32_t>(std::remove_cvref_t<config_t>::id);
+
     //!\brief Helper function to check if static id member exists.
     template <typename config_t>
     static constexpr auto has_id_member(int) -> decltype((static_cast<void>(config_t::id), true))
@@ -76,6 +80,37 @@ private:
     }
 
 public:
+    //!\brief Type alias for the internal enumeration type that represents the id.
+    template <typename config_t>
+    using id_type = std::remove_cvref_t<decltype(config_t::id)>;
+
+    /*!\brief Checks if two configuration types are compatible.
+     *
+     * \tparam config1_t The type of the first configuration to check against the second.
+     * \tparam config2_t The type of the second configuration.
+     *
+     * \details
+     *
+     * Uses the seqan3::detail::compatibility_table of the corresponding configuration element domain to check
+     * if both configurations can be combined.
+     */
+    template <typename config1_t, typename config2_t>
+    static constexpr auto is_compatible()
+    {
+        if constexpr (has_id_member<config1_t>(0) && has_id_member<config2_t>(0)) // needed for gcc-9 -stc=c++2a
+        {
+            if constexpr (std::same_as<id_type<config1_t>, id_type<config2_t>>) // needed for gcc-7 and gcc-8
+                return std::bool_constant<compatibility_table<id_type<config1_t>>[as_int<config1_t>]
+                                                                                 [as_int<config2_t>]>{};
+            else
+                return std::false_type{};
+        }
+        else
+        {
+            return std::false_type{};
+        }
+    }
+
     //!\brief Variable template that evaluates to `true` if the type has a static id member, otherwise `false`.
     template <typename config_t>
     static constexpr bool has_id = has_id_member<config_t>(0);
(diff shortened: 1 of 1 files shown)


include/seqan3/core/algorithm/concept.hpp
```diff
@@ -76,6 +80,37 @@ struct config_id_accessor
...
+            if constexpr (std::same_as<id_type<config1_t>, id_type<config2_t>>) // needed for gcc-7 and gcc-8
+                return std::bool_constant<compatibility_table<id_type<config1_t>>[as_int<config1_t>]
+                                                                                 [as_int<config2_t>]>{};
+            else
+                return std::false_type{};
+        }
+        else
+        {
+            return std::false_type{};
+        }
```
marehr, line 111: Could we use the

```c++
#if SEQAN3_WORKAROUND_GCC_... // gcc 9
  // gcc 9 code to make this work
#elif SEQAN3_WORKAROUND_GCC_... // gcc 7 and gcc 8
  // gcc 7 and 8 code to make this work
#else // ^^^ workaround / no workaround vvv
  // code that should work as it is
#endif

**Review by marehr — CHANGES_REQUESTED** [review]
Thank you!

Small change and this is good to go.

File: include/seqan3/core/algorithm/concept.hpp (hunk 1 above)
Comment (rrahn) [reply include/seqan3/core/algorithm/concept.hpp:111]: Well, the whole thing should not be needed and work as it is. The concept checks already if both have  id_member and if they have the same type. Still for some reasons on some compilers I had to repeat the checks in the function as well. 
Are you sure, that cluttering it any further with workaround macros makes it more readable? I doubt that the compile time reduces much, since everything was instantiated at least once anyway.

Comment (marehr) [reply include/seqan3/core/algorithm/concept.hpp:111]: no, I purely think, what happens when we drop gcc 7, gcc 8 and so on. We should make the transition at that point as easy as possible by marking that we did a workaround, and we wanted to have a different solution.

Comment (rrahn) [reply include/seqan3/core/algorithm/concept.hpp:111]: This is what the code would look like:
```cpp
#if     SEQAN3_WORKAROUND_GCC_9 // gcc 9
        if constexpr (has_id_member<config1_t>(0) && has_id_member<config2_t>(0)) // needed for gcc-9 -stc=c++2a
        {
#if     SEQAN3_WORKAROUND_GCC_less_than_9 // gcc 7 and gcc 8    
            if constexpr (std::same_as<id_type<config1_t>, id_type<config2_t>>) // needed for gcc-7 and gcc-8
#endif  SEQAN3_WORKAROUND_GCC_less_than_9 // ^^^ workaround / no workaround vvv
                return std::bool_constant<compatibility_table<id_type<config1_t>>[as_int<config1_t>]
                                                                                 [as_int<config2_t>]>{};
#if     SEQAN3_WORKAROUND_GCC_less_than_9 // gcc 7 and gcc 8    
            else
                return std::false_type{};
#endif SEQAN3_WORKAROUND_GCC_less_than_9 // ^^^ workaround / no workaround vvv
        }
        else
        {
            
            return std::false_type{};
        }
#endif  SEQAN3_WORKAROUND_GCC_9 // gcc 9
```
Not the fan of this, since the whole thing was supposed to not exist for any of the compilers.

rrahn: I mean the workaround is already below inside of the concept. I think that is expressive enough.

include/seqan3/core/algorithm/concept.hpp
```diff
@@ -113,4 +148,34 @@ SEQAN3_CONCEPT config_element_specialisation = requires
...
+ *  * are defined within the same algorithm configuration domain,
+ *  * a seqan3::detail::compatibility_table is defined for the configuration elements and
+ *    returns `true` for both configurations.
+ */
+//!\cond
+template <typename config1_t, typename config2_t>
+SEQAN3_CONCEPT config_element_pipeable_with =
+    config_element_specialisation<config1_t> &&
+    config_element_specialisation<config2_t> &&
+#ifdef SEQAN3_WORKAROUND_GCC_PIPEABLE_CONFIG_CONCEPT
```
marehr, line 171:
```suggestion
#if SEQAN3_WORKAROUND_GCC_PIPEABLE_CONFIG_CONCEPT
```

??

include/seqan3/core/algorithm/concept.hpp
marehr: How about:

```c++
#if SEQAN3_WORKAROUND_GCC_PIPEABLE_CONFIG_CONCEPT
    template <typename config1_t, typename config2_t>
    static constexpr auto is_compatible()
    {
        if constexpr (!(has_id_member<config1_t>(0) && has_id_member<config2_t>(0))) // needed for gcc-9 -stc=c++2a
            return std::false_type{};
#if (__GNUC__ < 9)
        else if constexpr (!std::same_as<id_type<config1_t>, id_type<config2_t>) // needed for gcc-7 and gcc-8
            return std::false_type{};
#endif
        else
            return std::bool_constant<compatibility_table<id_type<config1_t>>[as_int<config1_t>][as_int<config2_t>]>{};
    }
#endif // SEQAN3_WORKAROUND_GCC_PIPEABLE_CONFIG_CONCEPT
```
?

marehr (commented):
.

include/seqan3/core/algorithm/concept.hpp
rrahn: let's see

include/seqan3/core/algorithm/concept.hpp
```diff
@@ -94,22 +94,20 @@ struct config_id_accessor
...
-                                                                                 [as_int<config2_t>]>{};
-            else
-                return std::false_type{};
-        }
-        else
-        {
+        if constexpr (!(has_id_member<config1_t>(0) && has_id_member<config2_t>(0))) // needed for gcc-9 -stc=c++2a
+            return std::false_type{};
+#if (__GNUC__ < 9)
+        else if constexpr (!std::same_as<id_type<config1_t>, id_type<config2_t>) // needed for gcc-7 and gcc-8
```
marehr, line 104: `>` for `same_as` is missing :see_no_evil:

include/seqan3/core/algorithm/concept.hpp
```diff
@@ -42,7 +42,7 @@ namespace seqan3::detail
  *                           this table to allow validation checks.
  */
 template <typename algorithm_id_type>
-inline constexpr std::array<std::array<void, 0>, 0> compatibility_table;
+inline constexpr std::array<std::array<void *, 0>, 0> compatibility_table;
```
marehr, line 45:
```suggestion
inline constexpr std::array<std::array<void *, 0>, 0> compatibility_table{};
```

include/seqan3/core/algorithm/concept.hpp
marehr: still open

include/seqan3/core/algorithm/concept.hpp
marehr: let's do your way... :)

marehr (approved):

include/seqan3/core/algorithm/concept.hpp
```diff
@@ -76,6 +80,37 @@ struct config_id_accessor
...
+    template <typename config1_t, typename config2_t>
+    static constexpr auto is_compatible()
+    {
+        if constexpr (has_id_member<config1_t>(0) && has_id_member<config2_t>(0)) // needed for gcc <= 9
+        {
+            if constexpr (std::same_as<id_type<config1_t>, id_type<config2_t>>)
+                return std::bool_constant<compatibility_table<id_type<config1_t>>[as_int<config1_t>]
+                                                                                 [as_int<config2_t>]>{};
+            else
+                return std::false_type{};
```
smehringer, line 106:
```suggestion
            if constexpr (std::same_as<id_type<config1_t>, id_type<config2_t>>)
            {
                return std::bool_constant<compatibility_table<id_type<config1_t>>[as_int<config1_t>]
                                                                                 [as_int<config2_t>]>{};
            }
            else
            {
                return std::false_type{};
            }
```

include/seqan3/core/algorithm/concept.hpp
```diff
@@ -42,13 +42,13 @@ namespace seqan3::detail
  *                           this table to allow validation checks.
  */
 template <typename algorithm_id_type>
-inline constexpr std::array<std::array<void, 0>, 0> compatibility_table;
+inline constexpr std::array<std::array<void *, 0>, 0> compatibility_table{};
```
smehringer, line 45: Why is it now void * instead of void?

smehringer (approved):

include/seqan3/core/algorithm/concept.hpp
marehr: void is an invalid value type for std::array since [std::array](https://en.cppreference.com/w/cpp/container/array) uses `reference` which is defined as `value_type&`. But, `void &` is no valid type.

smehringer: ah ok. thanks for the explanation :)

PR #2249, opened 2020-11-11
Core utility split/type traits

rrahn:
* Moves the core/type_list module to utility/type_list.
* Splits the type_list traits module into traits working on type_list and traits working on type packs.
* Adapts the documentation, tests and snippets to the new module split.
* Adds general all header and utility folder documentation.


eseiler (changes requested):
* include order is different for `core` and `utility` (fix now or later?)
* didn't you propose to have an `algorithm.hpp` for both `type_list` and `type_pack`?

rrahn:
> include order is different for core and utility (fix now or later?)

I was in the making of adding a commit fixing the include order 👍 

> didn't you propose to have an algorithm.hpp for both type_list and type_pack?

Yes, but it is not part of this PR, since the algorithm stuff was in a different sub-module. But we can now move them to the respective sub-module in the next step.

eseiler (approved):

eseiler:
Ah, and you need to adapt the snippet includes in the docs, they are still trying to include from cor.

rrahn:
Mhmm, the buffer queue test failed because the checksum was not the same. Seems not related.

include/seqan3/core/type_list/all.hpp
```diff
@@ -7,30 +7,13 @@
 
 /*!\file
  * \author Hannes Hauswedell <hannes.hauswedell AT fu-berlin.de>
- * \brief Provides seqan3::type_list and metaprogramming utilities.
+ * \brief [DEPRECATED] Provides seqan3::type_list and metaprogramming utilities.
+ * \deprecated This header will be removed in 3.1. Please use seqan3/utility/type_list/all.hpp instead.
```
smehringer, line 11:
```suggestion
 * \deprecated This header will be removed in 3.1. Please use seqan3/utility/type_list/all.hpp and 
 *                     seqan3/utility/type_pack/all.hpp instead.
```

Since only type_list traits may not be enough in certain cases?

include/seqan3/utility/all.hpp
```diff
@@ -0,0 +1,29 @@
...
+/*!\defgroup utility Utility
+ * \brief Provides additional utility functionality used by multiple modules.
+ *
+ * The utility module contains concepts, functions, traits and classes that
+ * are independent of the remaining modules in SeqAn. These implementations
+ * are considered external functionality, i.e. they could have been outsourced
+ * into their own libraries.
+ *
+ * The utility module has no dependency to any other module except the \ref core
+ * module.
```
smehringer, line 28: Did we conclude that utility is/is not API stable? I forgot :see_no_evil: I would drop a note about API stability either way.

include/seqan3/utility/type_pack/all.hpp
```diff
@@ -0,0 +1,27 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2020, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \author Rene Rahn <rene.rahn AT fu-berlin.de>
+ * \brief Provides seqan3::type_list and associated metaprogramming utilities.
```
smehringer, line 10:
```suggestion
 * \brief Provides metaprogramming utilities for parameter packs.
```

include/seqan3/utility/type_pack/all.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+#include <seqan3/utility/type_pack/traits.hpp>
+
+/*!\defgroup type_pack Type Pack
+ * \brief Provides metaprogramming utilities for working on type packs.
+ * \ingroup utility
+ *
+ * \details
+ *
+ * All traits on a type pack are defined in the header `<seqan3/utility/type_pack/traits.hpp>`.
```
smehringer, line 21: Hm I see why you named the folder "type_pack" to be in line with "type_list" but in the documentation I would prefer to call it parameter pack because type pack is not really the name for it?
Maybe also include a link to the parameter pack documentation of parameter packs: https://en.cppreference.com/w/cpp/language/parameter_pack

test/snippet/core/detail/all_of_in_pack.cpp
```diff
@@ -1,8 +1,8 @@
 #include <type_traits>
 
-#include <seqan3/core/detail/pack_algorithm.hpp>
-#include <seqan3/core/type_list/type_list.hpp>
 #include <seqan3/core/debug_stream.hpp>
+#include <seqan3/core/detail/pack_algorithm.hpp>
+#include <seqan3/utility/type_list/type_list.hpp>
```
smehringer, line 5: Do you really need the type_list include here? I don't see a type list being used

test/snippet/core/detail/all_of_in_type_list.cpp
```diff
@@ -1,8 +1,8 @@
 #include <type_traits>
 
-#include <seqan3/core/detail/pack_algorithm.hpp>
-#include <seqan3/core/type_list/type_list.hpp>
 #include <seqan3/core/debug_stream.hpp>
+#include <seqan3/core/detail/pack_algorithm.hpp>
```
smehringer, line 4: Now I am confused. This snippet is called `all_of_in_type_list` but you are including `pack_algorithm`? Does pack_algorithm also include algorithms on type lists?

include/seqan3/utility/type_pack/traits.hpp
```diff
@@ -0,0 +1,473 @@
...
+// ----------------------------------------------------------------------------
+
+namespace seqan3::pack_traits::detail
+{
+
+/*!\brief Implementation for seqan3::pack_traits::find.
+ * \tparam query_t  The type you are searching for.
+ * \tparam pack_t   The type pack.
+ * \returns The position of the first occurrence of `query_t` in `pack_t` or `-1` if it is not contained.
+ * \ingroup type_list
```
smehringer: Isn't there a type_pack group now?
Check all `\ingroup` commands of this file

smehringer (changes requested):
Phew this was a lot.

It was a little hard to review commit by commit, because documentation and deprecation updates were done after moving and then include order changed again. So I reviewed all at once, in order to not review stuff that changed anyway in a later commit. I hope I did not miss something important that way.

include/seqan3/core/type_list/all.hpp
```diff
@@ -7,30 +7,13 @@
...
  */
 
 #pragma once
 
-#include <seqan3/core/type_list/type_list.hpp>
-#include <seqan3/core/type_list/traits.hpp>
+#include <seqan3/utility/type_list/all.hpp>
```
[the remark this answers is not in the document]
rrahn: No, because the list traits use the pack traits. Everything works like before.

include/seqan3/core/type_list/traits.hpp
```diff
@@ -7,886 +7,13 @@
 
 /*!\file
  * \author Hannes Hauswedell <hannes.hauswedell AT fu-berlin.de>
- * \brief Provides traits for seqan3::type_list.
+ * \brief [DEPRECATED] Provides traits for seqan3::type_list.
+ * \deprecated This header will be removed in 3.1. Please use seqan3/utility/type_list/traits.hpp instead.
```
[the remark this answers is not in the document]
rrahn: No, is not. Same reason :).

test/snippet/core/detail/all_of_in_type_list.cpp
rrahn: pack algorithm is from a different header. So it was not part of this change.

include/seqan3/alphabet/composite/alphabet_tuple_base.hpp
```diff
@@ -26,6 +26,8 @@
 #include <seqan3/core/concept/tuple.hpp>
 #include <seqan3/core/type_traits/transformation_trait_or.hpp>
 #include <seqan3/utility/detail/integer_traits.hpp>
+#include <seqan3/utility/type_list/traits.hpp>
+#include <seqan3/utility/type_pack/traits.hpp>
```
smehringer, line 30: Why do you need to add those here when there were not included before?

smehringer (approved):

include/seqan3/alphabet/composite/alphabet_tuple_base.hpp
rrahn: Because the file is using both of them. Before it was getting it indirectly through some other include, which itself didn't used them. So by removing it, it appeared that this header is the one that actually needs it.

smehringer:
GCC8 on macOS is failing? Maybe a rebased fixes it. I remember Enrico working on something

eseiler:
> GCC8 on macOS is failing? Maybe a rebase fixes it. I remember Enrico working on something

Fail is fxed on master

PR #2252, opened 2020-11-11
[INFRA] cmake: use fetch content

marehr:
This goes one step into solving https://github.com/seqan/product_backlog/issues/223

FetchContent became now the proper way to include external CMake Project, alongside using a submodule. See https://cliutils.gitlab.io/modern-cmake/chapters/projects/fetch.html

FetchContent has the advantage that it does the download phase at cmake configuration time (this allows the direct usage of defined targets of that cmake project) and the build phase when you are building your binary.

### DIFF
--- a/test/cmake/seqan3_require_benchmark.cmake
+++ b/test/cmake/seqan3_require_benchmark.cmake
@@ -6,9 +6,7 @@
 # -----------------------------------------------------------------------------------------------------
 
 # Exposes the google-benchmark target `gbenchmark`.
-macro (seqan3_require_benchmark)
-    enable_testing ()
-
+macro (seqan3_require_benchmark_old gbenchmark_git_tag)
     set (gbenchmark_project_args ${SEQAN3_EXTERNAL_PROJECT_CMAKE_ARGS})
     list (APPEND gbenchmark_project_args "-DBENCHMARK_ENABLE_TESTING=false")
     # google-benchmarks suggest to use LTO (link-time optimisation), but we don't really need that because we are a
@@ -26,7 +24,7 @@ macro (seqan3_require_benchmark)
         gbenchmark_project
         PREFIX gbenchmark_project
         GIT_REPOSITORY "https://github.com/google/benchmark.git"
-        GIT_TAG "v1.5.0"
+        GIT_TAG "${gbenchmark_git_tag}"
         SOURCE_DIR "${SEQAN3_BENCHMARK_CLONE_DIR}"
         CMAKE_ARGS "${gbenchmark_project_args}"
         BUILD_BYPRODUCTS "${gbenchmark_path}"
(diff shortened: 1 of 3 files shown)


eseiler:
https://github.com/seqan/seqan3/blob/54ead509f721cdf07ef63f476fdbe25610e99fb7/test/seqan3-test.cmake#L12

Needs to be bumped to 3.11?

test/cmake/seqan3_require_benchmark.cmake
```diff
@@ -48,3 +46,29 @@ macro (seqan3_require_benchmark)
 
     unset (gbenchmark_path)
 endmacro ()
+
+macro (seqan3_require_benchmark)
+    enable_testing ()
+
+    set (gbenchmark_git_tag "v1.5.0")
+
+    if (NOT ${CMAKE_VERSION} VERSION_LESS 3.14)
```
marehr, line 55: cmake 3.7 is still support, but cmake 3.14 or higher will use the new method. This makes it easier as a transition for our members.

test/cmake/seqan3_require_test.cmake
```diff
@@ -31,10 +29,7 @@ macro (seqan3_require_test)
         gtest_project
         PREFIX gtest_project
         GIT_REPOSITORY "https://github.com/google/googletest.git"
-        # we currently have warnings that were introduced in
-        # 03867b5389516a0f185af52672cf5472fa0c159c, which are still available
-        # in "release-1.8.1", see https://github.com/google/googletest/issues/1419
```
Irallia, line 36: I suppose you deleted these lines because we no longer have these warnings?
I saw that the googletest issue was closed.

test/cmake/seqan3_require_benchmark.cmake
```diff
@@ -48,3 +46,29 @@ macro (seqan3_require_benchmark)
 
     unset (gbenchmark_path)
 endmacro ()
+
+macro (seqan3_require_benchmark)
+    enable_testing ()
+
+    set (gbenchmark_git_tag "v1.5.0")
+
+    if (NOT CMAKE_VERSION VERSION_LESS 3.14)
```
Irallia, line 55: Does this `if / else` exist because the fetch does not work under version 3.14?

Irallia (approved):
As far as I can understand, I approve. 
I just left a few questions for my understanding. Thank you :)

And note, @eseiler has left a comment.

test/cmake/seqan3_require_benchmark.cmake
eseiler: Jup, I skipped the line when reading the changes...

Still, macOS cmake has some problem with it...

marehr: The problem was that the ci script does the bechmark/googletest install step during configuration, but the target that we used to express that don't always exist any more. This also means that we use on mac the default cmake installation and not the one we download.

test/cmake/seqan3_require_test.cmake
marehr: Yes, this is left-over.

test/cmake/seqan3_require_benchmark.cmake
marehr: perfect score :100:!

eseiler (approved):

PR #2256, opened 2020-11-11
[INFRA] Move core/detail

eseiler:
Unchanged:

detail/all.hpp
detail/customisation_point.hpp
detail/empty_type.hpp
detail/test_accessor.hpp
detail/strong_type.hpp 

Changed:

detail/endian.hpp -> seqan3/std/bit
detail/int_types.hpp -> utility/detail/.  
detail/to_string.hpp -> core/debug_stream/detail/to_string.hpp
detail/type_inspection.hpp -> utility/.


include/seqan3/alphabet/alphabet_base.hpp
```diff
@@ -15,7 +15,7 @@
 #include <cassert>
 
 #include <seqan3/alphabet/concept.hpp>
-#include <seqan3/core/detail/int_types.hpp>
+#include <seqan3/utility/detail/int_types.hpp>
```
marehr, line 18: do we sort these? or do we do that afterwards (this makes merge conflict less likely)?

marehr (approved):
Not sure about the reordering;

include/seqan3/alphabet/alphabet_base.hpp
smehringer: I would sort them right away.

include/seqan3/utility/detail/int_types.hpp
```diff
@@ -0,0 +1,51 @@
...
+//!\brief Given a value, cast the value as the smallest unsigned integer that can hold it.
+//!\sa seqan3::min_viable_uint_t
+template <uint64_t value>
+constexpr auto min_viable_uint_v = static_cast<min_viable_uint_t<value>>(value);
+
+// ------------------------------------------------------------------
+// size_in_values_v
+// ------------------------------------------------------------------
+
+//!\brief Return the number of values an integral type can have, i.e. the different between min and max.
```
smehringer, line 43:
```suggestion
//!\brief Return the number of values an integral type can have, i.e. the difference between min and max.
```

test/unit/core/detail/int_types_test.cpp
```diff
@@ -6,7 +6,8 @@
 // -----------------------------------------------------------------------------------------------------
 
 #include <gtest/gtest.h>
-#include <seqan3/core/detail/int_types.hpp>
+
+#include <seqan3/utility/detail/int_types.hpp>
```
smehringer, line 10: In general, now that I looked at the file, `int_types.hpp` isn't very expressive. Can we give the file a better name, now that we rename it anyway? As I see it, there are no int_types defined but rather some traits of integers? What about `int_traits.hpp`? I'll post something on Gitter.

include/seqan3/utility/type_inspection.hpp
```diff
@@ -0,0 +1,68 @@
...
+#if defined(__GNUC__) || defined(__clang__)
+#include <cxxabi.h>
+#endif // defined(__GNUC__) || defined(__clang__)
+
+#include <functional>
+#include <memory>
+#include <string>
+#include <typeinfo>
+
+#include <seqan3/core/platform.hpp>
```
smehringer, line 24: Hm, since utility should be independent from seqan, do we still need to include the platform header?

include/seqan3/utility/type_inspection.hpp
```diff
@@ -0,0 +1,68 @@
...
+ * \details
+ *
+ * On gcc and clang std::type_info only returns a mangled name.
+ * The mangled name can be converted to human-readable form using implementation-specific API such as
+ * abi::__cxa_demangle. In other implementations the name returned is already human-readable.
+ *
+ * \note The returned name is implementation defined and might change between different tool chains.
+ */
+template <typename type>
+inline std::string const type_name_as_string = [] ()
```
smehringer, line 44: Since this file ONLY contains `type_name_as_string` can we also call it `type_name_as_string.hpp`? I think we agreed to do this unless not possible.

include/seqan3/utility/type_inspection.hpp
```diff
@@ -0,0 +1,68 @@
...
+#endif // defined(__GNUC__) || defined(__clang__)
+
+#include <functional>
+#include <memory>
+#include <string>
+#include <typeinfo>
+
+#include <seqan3/core/platform.hpp>
+
+namespace seqan3::detail
```
smehringer, line 26: Now the file is not in the detail folder anymore but the entity is still in the detail namespace?

include/seqan3/core/detail/pack_algorithm.hpp
```diff
@@ -6,241 +6,15 @@
...
+ * \brief [DEPRECATED] Provides algorithms for meta programming, parameter packs and seqan3::type_list.
  * \author Rene Rahn <rene.rahn AT fu-berlin.de>
+ * \deprecated This header is deprecated and will be removed in SeqAn-3.1.0;. Please \#include
+ *             <seqan3/utility/detail/pack_algorithm.hpp> instead.
  */
 
 #pragma once
 
-#include <utility>
+#include <seqan3/utility/detail/pack_algorithm.hpp>
```
smehringer, line 17: Didn't we want to think about if pack algorithms should go into the "type_pack" utility submodule? I wonder if we should do all the renaming work twice or maybe postpone this one until we decide where it finally goes?

(Note: These changes clash with Renes PR about the type trait splitting)

smehringer (changes requested):
Can you change your commit messages:
* [INFRA] -> [MISC] (or whatever tag we decide is appropriate on Gitter :D)
* Rename ... `to ...`

include/seqan3/utility/type_inspection.hpp
eseiler: Yes, they are not excluded from the header test.

include/seqan3/core/detail/pack_algorithm.hpp
eseiler: Then I'll just remove the commit for the pack_algorithm

include/seqan3/utility/type_inspection.hpp
smehringer: Yes I think for now that's best. We can still decide if we want to make it public later on. (I don't remember if we explicitly decided to make this public?)

smehringer (approved):
LGTM. I'll wait for Gitter reponses on the `integer_traits.hpp` name before merging. Feel free to request me again or ping me if we decide it and I forget

PR #2258, opened 2020-11-12
[FIX] use proper workaround syntax

marehr:
This PR adapts older code to how we handle workarounds now. At some point we started to bundle our workarounds in one place (include/seqan3/core/platform.hpp) and made them available to all header instead of just leaving a comment.

### DIFF
--- a/include/seqan3/alphabet/concept.hpp
+++ b/include/seqan3/alphabet/concept.hpp
@@ -649,12 +649,11 @@ public:
     }
 };
 
-//!\cond
-// required to prevent https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89953
+#if SEQAN3_WORKAROUND_GCC_89953
 template <typename alph_t>
     requires requires { { alphabet_size_fn<alph_t>{} }; }
 inline constexpr auto alphabet_size_obj = alphabet_size_fn<alph_t>{};
-//!\endcond
+#endif // SEQAN3_WORKAROUND_GCC_89953
 
 } // namespace seqan3::detail::adl_only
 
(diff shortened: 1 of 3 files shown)


include/seqan3/alphabet/concept.hpp
```diff
@@ -699,12 +698,20 @@ namespace seqan3
...
              requires { { detail::adl_only::alphabet_size_obj<alph_t>() }; } // ICE workarounds
 //!\endcond
 inline constexpr auto alphabet_size = detail::adl_only::alphabet_size_obj<alph_t>();
+#else // ^^^ workaround / no workaround vvv
+template <typename alph_t>
+//!\cond
+    requires requires { { detail::adl_only::alphabet_size_fn<alph_t>{}() }; }
+//!\endcond
+inline constexpr auto alphabet_size = detail::adl_only::alphabet_size_fn<alph_t>{}();
+#endif // SEQAN3_WORKAROUND_GCC_89953
```
eseiler, line 714: Wondering if you could save some duplication here by putting just the `&& requires`  into a newline.
Depends on if it works and helps with understanding the code (the difference is clearer) and if we want to violate code style for this....

eseiler (approved):

include/seqan3/alphabet/concept.hpp
marehr: Note that both lines

```cpp
alphabet_size = detail::adl_only::alphabet_size_obj<alph_t>();
```

and

```
alphabet_size = detail::adl_only::alphabet_size_fn<alph_t>{}();
```

are different

rrahn (approved):
thank you ;)

PR #2260, opened 2020-11-12
[MISC] move {core => utility}/simd

marehr:
Part of seqan/product_backlog#160


MitraDarja (changes requested):
Looks good, but one coverage tests does not run through...

marehr:
> Looks good, but one coverage tests does not run through...

It only says that the views have untested code paths which is a known issue.

MitraDarja (approved):

include/seqan3/alignment/matrix/alignment_optimum.hpp
```diff
@@ -17,7 +17,7 @@
 #include <seqan3/alignment/matrix/detail/matrix_coordinate.hpp>
 #include <seqan3/core/concept/core_language.hpp>
 #include <seqan3/core/type_traits/template_inspection.hpp>
-#include <seqan3/core/simd/concept.hpp>
+#include <seqan3/utility/simd/concept.hpp>
```
smehringer, line 20: The header is now at the wrong position. Even if we do other header rearrangements in separate commits, we should place the new header directly into the correct header-order-location.

include/seqan3/core/simd/concept.hpp
```diff
@@ -6,114 +6,14 @@
 // -----------------------------------------------------------------------------------------------------
 
 /*!\file
- * \brief Provides seqan3::simd::simd_concept.
+ * \brief [DEPRECATED] Provides seqan3::simd::simd_concept.
  * \author Marcel Ehrhardt <marcel.ehrhardt AT fu-berlin.de>
+ * \deprecated This header will be removed in 3.1. Please use seqan3/utility/simd/concept.hpp instead.
```
smehringer, line 11:
```suggestion
 * \deprecated This header will be removed in SeqAn-3.1.0. Please use seqan3/utility/simd/concept.hpp instead.
```

include/seqan3/core/simd/view_to_simd.hpp
```diff
@@ -15,7 +15,7 @@
 #include <seqan3/alphabet/concept.hpp>
 #include <seqan3/core/detail/pack_algorithm.hpp>
 #include <seqan3/utility/simd/concept.hpp>
-#include <seqan3/core/simd/simd_algorithm.hpp>
+#include <seqan3/utility/simd/simd_algorithm.hpp>
```
smehringer: Side question: why prefix the header with `simd_` if it is already in the simd submodule? why not just `algorithm.hpp`?

include/seqan3/utility/simd/simd_algorithm.hpp
```diff
@@ -0,0 +1,444 @@
...
+#include <array>
+#include <utility>
+
+#include <seqan3/utility/simd/concept.hpp>
+#include <seqan3/core/simd/detail/builtin_simd.hpp>
+#include <seqan3/core/simd/detail/simd_algorithm_sse4.hpp>
+#include <seqan3/core/simd/detail/simd_algorithm_avx2.hpp>
+#include <seqan3/core/simd/detail/simd_algorithm_avx512.hpp>
+#include <seqan3/core/simd/simd_traits.hpp>
+#include <seqan3/std/concepts>
```
smehringer:
```suggestion
#include <array>
#include <seqan3/std/concepts>
#include <utility>

#include <seqan3/core/simd/detail/builtin_simd.hpp>
#include <seqan3/core/simd/detail/simd_algorithm_sse4.hpp>
#include <seqan3/core/simd/detail/simd_algorithm_avx2.hpp>
#include <seqan3/core/simd/detail/simd_algorithm_avx512.hpp>
#include <seqan3/core/simd/simd_traits.hpp>
#include <seqan3/utility/simd/concept.hpp>
```

smehringer (changes requested):
And change the commit tags once we decided on MISC, FIX or INFRA :D

include/seqan3/core/simd/view_to_simd.hpp
marehr: not sure the list said:

* simd/simd_algorithm.hpp -> utility/simd/.
* simd/simd_traits.hpp -> utility/simd/.

so I kept it, I added this question to https://github.com/seqan/product_backlog/issues/160

smehringer (changes requested):
And you commit messages still need the `[MISC]`  tag :)

I would really like to do the renaming of 
`core/simd/simd_traits.hpp` -> `utility/simd/traits.hpp` 
`core/simd/simd_algorithm.hpp` -> `utility/simd/algorithm.hpp`
right away in this PR (if we decide for it).
Since if not, we would have to deprecate `utility/simd/simd_algorithm` again don't we :laughing: ?

include/seqan3/utility/simd/all.hpp
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+#pragma once
+
+//!\cond DEV
+
+/*!\defgroup simd Simd
+ * \brief The simd module contains a unified interface to provide simd types and functions used in seqan3.
+ * \see https://en.wikipedia.org/wiki/SIMD
+ * \ingroup core
```
smehringer, line 20: Thanks to @simonsasse who just corrected me in my PR, I checked yours again. 
This should be in group utility now!
```suggestion
 * \ingroup utility
```

check all other files :see_no_evil:

smehringer (changes requested):

include/seqan3/utility/simd/all.hpp
marehr: I only have one since the documentation is within the simd module

include/seqan3/core/simd/all.hpp
```diff
@@ -6,42 +6,16 @@
...
+SEQAN3_DEPRECATED_HEADER(
+   "This header is deprecated and will be removed in SeqAn-3.1.0; Please #include <seqan3/utility/simd/all.hpp> instead.")
 
-/*!\namespace seqan3::simd
- * \brief The SeqAn namespace for simd data types, algorithms and meta functions.
- * \sa https://en.wikipedia.org/wiki/SIMD What is SIMD conceptually?
- * \sa https://en.wikipedia.org/wiki/stream_REMOVEMEing_SIMD_Extensions Which SIMD architectures exist?
- */
-//!\endcond
+#include <seqan3/utility/simd/all.hpp>
```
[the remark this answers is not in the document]
marehr: I thought there might be an issue if I include `#include <seqan3/utility/simd/all.hpp>` first, that first some other deprecation warning appears that might mislead that you have to use `seqan3/utility/simd/all.hpp` instead of `seqan3/code/simd/all.hpp`. But on second thought this should not be possible, because of our tests.

marehr:
> And you commit messages still need the `[MISC]` tag :)
done

> 
> I would really like to do the renaming of
> `core/simd/simd_traits.hpp` -> `utility/simd/traits.hpp`
> `core/simd/simd_algorithm.hpp` -> `utility/simd/algorithm.hpp`
> right away in this PR (if we decide for it).
> Since if not, we would have to deprecate `utility/simd/simd_algorithm` again don't we laughing ?

this is kind of open; a later rename wouldn't be needed to be deprecated since in between releases we should be flexible to do this.

smehringer (approved):

include/seqan3/core/simd/view_to_simd.hpp
rrahn: I think it would be ok to rename them without the prefix. That would be more consistent with the rest. I think the configs in alignment need to do this as well.

smehringer: Resolution 2020-11-23:

We should rename them right away.

marehr: I'm not in favour for that; The class is called `simd_traits` and the header reflects that name.

smehringer: Ah ok. well if the main class in this actually called `simd_traits` I would also keep it as is.
What about `simd_algorithm?` is it the same case? Otherwise we decided to call it just `algorithm.hpp`

marehr:
> Ah ok. well if the main class in this actually called `simd_traits` I would also keep it as is.
> What about `simd_algorithm?` is it the same case? Otherwise we decided to call it just `algorithm.hpp`
https://github.com/seqan/seqan3/pull/2260#discussion_r528639082

I renamed the algorithm header, but kept `simd_traits.hpp`

PR #2269, opened 2020-11-18
[MISC] Enable computation of begin position in alignment result builder.

rrahn:
Enables the computation of the begin position inside of the refactored alignment.
Before, only the score and end position could be computed. Now it is also possible to compute the begin positions. 
In the upcoming PRs also the trace computation will be enabled including the logging mechanism.

### DIFF
--- a/include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
+++ b/include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp
@@ -144,6 +144,7 @@ public:
                                          std::move(idx),
                                          this->optimal_score,
                                          this->optimal_coordinate,
+                                         alignment_matrix,
                                          callback);
         }
     }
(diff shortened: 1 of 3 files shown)


marehr:
Something went wrong with the unit tests.

```
[ 19%] Building CXX object alignment/pairwise/CMakeFiles/global_affine_banded_collection_simd_test.dir/global_affine_banded_collection_simd_test.cpp.o
In file included from ../../../seqan3/include/seqan3/alignment/pairwise/alignment_configurator.hpp:29:0,
                 from ../../../seqan3/include/seqan3/alignment/pairwise/align_pairwise.hpp:22,
                 from /home/runner/work/seqan3/seqan3/seqan3/test/unit/alignment/pairwise/global_affine_banded_collection_simd_test.cpp:12:
```

marehr (approved):

marehr:
Question: This change does not include tests, does that means it will not be tested right now?

rrahn:
> Question: This change does not include tests, does that mean it will not be tested right now?

Well, this is a code path that is currently not evaluated since it is still disabled in the configuration. In the next steps we add the logging mechanisms and extend it to the computation of the alignment. By that we also enable this feature within the configurator and see that it is tested

smehringer (approved):

PR #2270, opened 2020-11-18
[FIX] MacOS gcc8 CI

eseiler:
Apparently the image has changed and now the globally installed version of 8.4.0 is used instead of the brew one.
The globally installed one has some errors....


marehr (approved):
ty

smehringer (approved):

PR #2271, opened 2020-11-18
[INFRA] Only run macOS CI if linux CI was successful

eseiler:
Blocked by https://github.com/seqan/seqan3/pull/2270


marehr:
Can we somehow deduplicate it?

eseiler:
> Can we somehow deduplicate it?

Looking at it, https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/creating-a-composite-run-steps-action seems promising.

eseiler:
This works as intended, but there (currently) is no way to avoid the code duplication

eseiler:
I'll reevaluate this once we can avoid the code duplication

PR #2275, opened 2020-11-20
[FIX] Sporadic failures in parallel buffer queue test

eseiler:
It may happen that the writer thread closes the queue before all elements have been read.
I added an atomic counter and only close the queue if all elements have been read by the reader threads.


eseiler:
Superseded by https://github.com/seqan/seqan3/pull/2276

Issue #2278, opened 2020-11-24
align_cfg::parallel does not have an effect


eseiler:
## Platform

  - SeqAn version: master
  - Operating system: Linux 5.8.0-rc5+ #1 SMP Fri Jul 17 19:07:45 CEST 2020 ppc64le ppc64le ppc64le GNU/Linux
  - Compiler: all

## Description

The `parallel` option is ignored. 
On our nightlies, the alignment tests/snippets/etc. use 160 threads instead of the configured 4.

### How to repeat the problem

Run any test/snippet/etc that uses the `parallel` config.

### Expected behaviour

Do not use all threads.

### Actual behaviour

Uses all threads.

PR #2279, opened 2020-11-24
[FIX] parallel config ending up ignored

eseiler:
Resolves https://github.com/seqan/seqan3/issues/2278

We need to fix some things such that `move` works.

* lhs => default constructed => 160 threads
* rhs => custom construced => 4 threads
* lhs = std::move(rhs) => lhs.reset(rhs.release())
* rhs.release() => return ptr to resource and set ptr = nullptr => destructor ist not called, but not a problem here
* lhs.reset(new_ptr) => old_ptr = lhs.ptr; lhs.ptr = new_ptr; lhs.deleter(old_ptr) => problem! The destructor destoys the internal state with the thread pool before alle the threads have been joined since `wait()` hasn't been called.

### DIFF
--- a/include/seqan3/core/algorithm/detail/algorithm_executor_blocking.hpp
+++ b/include/seqan3/core/algorithm/detail/algorithm_executor_blocking.hpp
@@ -313,6 +313,7 @@ private:
     {
         algorithm = std::move(other.algorithm);
         buffer_size = std::move(other.buffer_size);
+        exec_handler = std::move(other.exec_handler);
         // Get the old resource position.
         auto old_resource_position = std::ranges::distance(std::ranges::begin(other.resource),
                                                            other.resource_it);
--- a/include/seqan3/core/algorithm/detail/execution_handler_parallel.hpp
+++ b/include/seqan3/core/algorithm/detail/execution_handler_parallel.hpp
@@ -44,6 +44,9 @@ namespace seqan3::detail
  * submit new algorithm tasks.
  *
  * \note Instances of this class are not copyable.
+ * 
+ * \warning This class is only thread-safe in a single producer context. Multiple consumers are allowed.
+ *          Concurrent invocation of the interfaces are undefined behaviour.
  *
  * \attention This class cannot be reused for multiple calls. For this to work, it requires barriers and a queue that
  *            can be reopened.
@@ -94,13 +97,7 @@ public:
     execution_handler_parallel(execution_handler_parallel &&) = default; //!< Defaulted.
     execution_handler_parallel & operator=(execution_handler_parallel const &) = delete; //!< Deleted.
     execution_handler_parallel & operator=(execution_handler_parallel &&) = default; //!< Defaulted.
-
-    //!\brief Waits for threads to finish.
-    ~execution_handler_parallel()
-    {
-        if (state != nullptr)
-            wait();
-    }
+    ~execution_handler_parallel() = default; //!< Defaulted.
     //!\}
 
     /*!\brief Asynchronously schedules a new algorithm task with the given input and callback.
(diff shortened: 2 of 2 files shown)


marehr (approved):

include/seqan3/core/algorithm/detail/algorithm_executor_blocking.hpp
```diff
@@ -313,6 +313,7 @@ class algorithm_executor_blocking
     {
         algorithm = std::move(other.algorithm);
         buffer_size = std::move(other.buffer_size);
+        std::swap(exec_handler, other.exec_handler);
```
rrahn, line 316:
```suggestion
        using std::swap;
        swap(exec_handler, other.exec_handler);
```

Mhmm, in general the current default move implementation of the parallel execution handler is wrong. 
This should be fixable by adding a custom deleter to the unique ptr, which does what the destructor of the class does: Closing the queue and waiting on the threads to finish. Then the destructor doesn't need to do this.
And this didn't work: `exec_handler = std::move(other.exec_handler)`?

rrahn (changes requested):

include/seqan3/core/algorithm/detail/algorithm_executor_blocking.hpp
eseiler: Nope, you could do a member initialization here: https://github.com/seqan/seqan3/blob/d37ba83a333992a551ec835097e564d8b9a00604/include/seqan3/core/algorithm/detail/algorithm_executor_blocking.hpp#L124
( `: exec_handler{other.exec_handler}`).

Putting `exec_handler = std::move(other.exec_handler)` at any point doesn't work because the `exec_handler` was already default constructed; and apparently then the destructor doesn't work properly.

rrahn: Ok, I just wanted to make sure. A possible reason for this is that during the move assignment the old resource with the default initiliased thread pool is simply deleted (as if calling `lhs.reset(rhs.release())`). That has side effects since threads that are still joinable are simply destructed. Hence, the unique ptr needs a special deleter that takes care of this properly. The only weird thing is that swap wouldn't do anything different here, since it is done on the wrapper class which does not offer a swap member or swap overload. That's why I am confused that the proposed code should work.

include/seqan3/core/algorithm/detail/execution_handler_parallel.hpp
```diff
@@ -189,23 +183,47 @@ class execution_handler_parallel
...
-            state->queue.close();
+            stop_and_wait();
+        }
+        //!\}
 
-            for (auto & t : state->thread_pool)
+        //!\brief Waits until all threads have been joined.
+        void stop_and_wait()
+        {
+            if (!is_waiting)
```
rrahn, line 214: I am wondering if the current mechanism is actually thread safe. 
So it could happen that one thread calls `stop_and_wait` and sets waiting to true. 
If now another thread is calling the destructor for some reason it will see that someone is already waiting but doesn't care and destructs the object. We can also not concurrently join the threads as this results in a data-race. And it is also possible that `is_waiting` is set to `true` after another thread has already entered the if block.
So we need to synchronise this i.e. only one thread can ever join the threads and all other wait until threads have been joined. I guess right now we only have a single producer so it is not really a problem but that might change in the future and then it will get tricky to figure out what is going on here.
So we could documented as not thread safe to call stop and wait and also document that the entire class is only intended for a single producer model. Then we could also throw away the is_waiting flag, because it suggests some safety that does not exixt

rrahn (changes requested):

eseiler:
I removed the `is_waiting` flag and documented `stop_and_wait` as well as `internal_state` as not thread-safe.

include/seqan3/core/algorithm/detail/execution_handler_parallel.hpp
```diff
@@ -189,29 +183,61 @@ class execution_handler_parallel
...
+    }
+
+private:
+    /*!\brief An internal state stored on the heap to allow safe move construction/assignment of the class.
+     *
+     * \details
+     * 
+     * ### Thread safety
+     * 
+     * This class is only intended for use with a single producer model.
```
rrahn, line 196: I think we need to put that as a warning comment in the class documentation of the execution handler. Because there will be rather looking there instead of the internal class documentation. 
So the message should convey that this execution handler is only thread safe in a single producer context but allows multiple consumers. And that concurrent invocation of the interfaces will lead to undefined behaviour.

rrahn (changes requested):
I think this is good now. But I would make the documentation entry part of the enclosed class' documentation and also emphasise it with a warning.

include/seqan3/core/algorithm/detail/execution_handler_parallel.hpp
eseiler:
```
 * \warning This class is only thread-safe in a single producer context. Multiple consumers are allowed.
 *          Concurrent invocation of the interfaces are undefined behaviour.
```

? That''s what I'll add.

rrahn (approved):
ok, great. This thing should be done. 👍

PR #2284, opened 2020-11-25
[INFRA] Make codecov patch report non-failing

eseiler:
The patch report is often broken.
It is meant to just look at the changes of commit, but is not very reliant.

Usually we look at the project report which summarizes added/deleted lines and increases/decreases in hit code lines

See https://docs.codecov.io/docs/commit-status

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,13 +1,17 @@
 # See https://docs.codecov.io/docs/codecovyml-reference
 codecov:
-  require_ci_to_pass: no
+  require_ci_to_pass: no  # codecov reports its results independent of whether CI passed
   notify:
-    wait_for_ci: no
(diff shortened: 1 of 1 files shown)


joergi-w (approved):

.codecov.yml
```diff
@@ -11,3 +11,7 @@ coverage:
         if_ci_failed: success
         target: auto
         threshold: 0.01%
+    patch: 
+      default:
+        if_ci_failed: success
```
rrahn, line 16: A  comment of why we do this here would be not bad. So in the future maintainer understand why this was done.

rrahn (changes requested):

.codecov.yml
eseiler: I commented everything

eseiler:
@rrahn ping

rrahn (approved):

PR #2285, opened 2020-11-26
[MISC] Change seqan3::option_spec enum names to lower case.

smehringer:
I had some free time to tackle other stuff and this was something API relevant that needed to be done for a while.

Part of https://github.com/seqan/product_backlog/issues/246

The enum names of `seqan3::option_spec` were changed to lower case:
  * `seqan3::option_spec::DEFAULT` is replaced by `seqan3::option_spec::defaulted`.
  * `seqan3::option_spec::REQUIRED` is replaced by `seqan3::option_spec::required`.
  * `seqan3::option_spec::ADVANCED` is replaced by `seqan3::option_spec::advanced`.
  * `seqan3::option_spec::HIDDEN` is replaced by `seqan3::option_spec::hidden`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,7 +54,8 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 
 #### Argument Parser
 
-* The enum names of `seqan3::option_spec` were changed to lower case:
+* The enum names of `seqan3::option_spec` were changed to lower case
+  ([\#2285](https://github.com/seqan/seqan3/pull/2285)):
   * `seqan3::option_spec::DEFAULT` is replaced by `seqan3::option_spec::defaulted`.
   * `seqan3::option_spec::REQUIRED` is replaced by `seqan3::option_spec::required`.
   * `seqan3::option_spec::ADVANCED` is replaced by `seqan3::option_spec::advanced`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,16 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 * The `seqan3::regex_validator` parses `std::filesystem::path`'s correctly now
   ([\#2216](https://github.com/seqan/seqan3/pull/2216)).
 
+## API changes
+
+#### Argument Parser
+
+* The enum names of `seqan3::option_spec` were changed to lower case:
+  * `seqan3::option_spec::DEFAULT` is replaced by `seqan3::option_spec::defaulted`.
+  * `seqan3::option_spec::REQUIRED` is replaced by `seqan3::option_spec::required`.
+  * `seqan3::option_spec::ADVANCED` is replaced by `seqan3::option_spec::advanced`.
+  * `seqan3::option_spec::HIDDEN` is replaced by `seqan3::option_spec::hidden`.
+
 # 3.0.2
 
 Note that 3.1.0 will be the first API stable release and interfaces in this release might still change.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,7 +56,7 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 
 * The enum names of `seqan3::option_spec` were changed to lower case
   ([\#2285](https://github.com/seqan/seqan3/pull/2285)):
-  * `seqan3::option_spec::DEFAULT` is replaced by `seqan3::option_spec::defaulted`.
+  * `seqan3::option_spec::DEFAULT` is replaced by `seqan3::option_spec::standard`.
   * `seqan3::option_spec::REQUIRED` is replaced by `seqan3::option_spec::required`.
   * `seqan3::option_spec::ADVANCED` is replaced by `seqan3::option_spec::advanced`.
   * `seqan3::option_spec::HIDDEN` is replaced by `seqan3::option_spec::hidden`.
--- a/doc/tutorial/argument_parser/index.md
+++ b/doc/tutorial/argument_parser/index.md
@@ -340,7 +340,7 @@ Our applications often do not allow just any value to be passed as input argumen
 
 A *validator* is a [functor](https://stackoverflow.com/questions/356950/what-are-c-functors-and-their-uses) that is called within the argument parser after retrieving and converting a command line argument. We provide several validators, which we hope cover most of the use cases, but you can always create your own validator (see section [Create your own validator](#section_create_your_own_validator)).
 
-\attention You can pass a validator to the seqan3::argument_parser::add_option function only after passing the seqan3::option_spec parameter. Pass the seqan3::option_spec::defaulted tag, if there are no further restrictions on your option.
+\attention You can pass a validator to the seqan3::argument_parser::add_option function only after passing the seqan3::option_spec parameter. Pass the seqan3::option_spec::defaulted tag if there are no further restrictions on your option.
 
 ## SeqAn validators
 
--- a/doc/tutorial/argument_parser/index.md
+++ b/doc/tutorial/argument_parser/index.md
@@ -308,12 +308,12 @@ Set an option/flag to hidden, if you want to completely hide it from the user. I
 
 Summary:
 
-| Tag        | Description                                                    |
-|:-----------|:---------------------------------------------------------------|
-| defaulted  | The default tag with no special behaviour.                     |
-| required   | Required options will cause an error if not provided.          |
-| advanced   | Advanced options are only displayed wit `-hh/--advanced-help`. |
-| hidden     | Hidden options are never displayed when exported.              |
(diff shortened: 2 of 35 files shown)


simonsasse (approved):
Looks Good, just wondered if `seqan3::option_spec::default` might be more straight forward ? 🥳

include/seqan3/argument_parser/auxiliary.hpp
```diff
@@ -224,28 +224,33 @@ inline debug_stream_type<char_t> & operator<<(debug_stream_type<char_t> & s, opt
...
+    hidden = 4,/*!< Set an option/flag to hidden, if you want to completely hide it from
+                  * the user. It will never appear on the help page nor any export format.
+                  * For example, this can be useful for debugging reasons.
+                  * (e.g. "A tool for mapping reads to the genome").
+                  */
+
+    DEFAULT SEQAN3_DEPRECATED_310  = 8,  //!< \deprecated Use seqan3::option_spec::defaulted instead.
+    REQUIRED SEQAN3_DEPRECATED_310 = 16, //!< \deprecated Use seqan3::option_spec::required instead.
+    ADVANCED SEQAN3_DEPRECATED_310 = 32, //!< \deprecated Use seqan3::option_spec::advanced instead.
+    HIDDEN SEQAN3_DEPRECATED_310   = 64, //!< \deprecated Use seqan3::option_spec::hidden instead.
```
eseiler, line 253:
```suggestion
    DEFAULT SEQAN3_DEPRECATED_310  = defaulted,  //!< \deprecated Use seqan3::option_spec::defaulted instead.
    REQUIRED SEQAN3_DEPRECATED_310 = required, //!< \deprecated Use seqan3::option_spec::required instead.
    ADVANCED SEQAN3_DEPRECATED_310 = advanced, //!< \deprecated Use seqan3::option_spec::advanced instead.
    HIDDEN SEQAN3_DEPRECATED_310   = hidden  //!< \deprecated Use seqan3::option_spec::hidden instead.
```

Otherwise, the deprecated values have no functionality.
I also removed the comma after the last entry since it should not be necessary?

smehringer:
> Looks Good, just wondered if `seqan3::option_spec::default` might be more straight forward ? partying_face

`default` is a reserved name because it is a compiler keyword so I cannot use it.
E.g. in constructors we write `foo() = default;`

CHANGELOG.md
```diff
@@ -50,6 +50,16 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 * The `seqan3::regex_validator` parses `std::filesystem::path`'s correctly now
   ([\#2216](https://github.com/seqan/seqan3/pull/2216)).
 
+## API changes
+
+#### Argument Parser
+
+* The enum names of `seqan3::option_spec` were changed to lower case:
```
rrahn, line 57: can you add the link to this PR?

doc/tutorial/argument_parser/index.md
```diff
@@ -340,7 +340,7 @@ Our applications often do not allow just any value to be passed as input argumen
 
 A *validator* is a [functor](https://stackoverflow.com/questions/356950/what-are-c-functors-and-their-uses) that is called within the argument parser after retrieving and converting a command line argument. We provide several validators, which we hope cover most of the use cases, but you can always create your own validator (see section [Create your own validator](#section_create_your_own_validator)).
 
-\attention You can pass a validator to the seqan3::argument_parser::add_option function only after passing the seqan3::option_spec parameter. Pass the seqan3::option_spec::DEFAULT tag, if there are no further restrictions on your option.
+\attention You can pass a validator to the seqan3::argument_parser::add_option function only after passing the seqan3::option_spec parameter. Pass the seqan3::option_spec::defaulted tag, if there are no further restrictions on your option.
```
rrahn, line 343:
```suggestion
\attention You can pass a validator to the seqan3::argument_parser::add_option function only after passing the seqan3::option_spec parameter. Pass the seqan3::option_spec::defaulted tag if there are no further restrictions on your option.
```

Not sure so  100%

doc/tutorial/argument_parser/index.md
```diff
@@ -310,10 +310,10 @@ Summary:
 
 | Tag        | Description                                                    |
 |:-----------|:---------------------------------------------------------------|
-| DEFAULT    | The default tag with no special behaviour.                     |
-| REQUIRED   | Required options will cause an error if not provided.          |
-| ADVANCED   | Advanced options are only displayed wit `-hh/--advanced-help`. |
-| HIDDEN     | Hidden options are never displayed when exported.              |
+| defaulted  | The default tag with no special behaviour.                     |
```
rrahn, line 313: I think the meaning of defaulted as adjective is wrong here. So `default` was actually correct.
And I just realise that we use this misleading word for all the defaulted constructors 🤯.

rrahn (changes requested):
Thank you. The `defaulted` -> `default` naming should be addressed. Maybe you can independently of me research about the meaning `defaulted` as an adjective. If I am not mistaken, then we have this all over the library :cry:

doc/tutorial/argument_parser/index.md
smehringer: I **cannot** use default :( ! (in lower case letters) because it is a compiler keyword. So I thought defaulted is the next best alternative. If you deem defaulted wrong, how else should we name it?

rrahn: But the meaning is rather: `required option`, `hidden option` etc. As far as I remember, we use a verb for functions only (it would be good if I could point now to the style guide but we don't have it). Nonetheless, I never've seen an enum value as a verb so it would be confusing to start now. Another, question, why do we have a default tag? Would that not be something internal, if the user just does not specify anything? What is the default behaviour? Why do you need to say explicitly it is default, which is already the meaning if I don't specify the other? Maybe we can remove it?

smehringer: > Nonetheless, I never've seen an enum value as a verb so it would be confusing to start now.

hm.. I see. I have no strong feelings about this. 

> Another, question, why do we have a default tag?

Because you need to specify the option_spec if you want to specify a validator.

```
p.add_option(var, 'i', "integer", "desc", option_spec::default, arithmetic_validator{3, 5});
```

> What is the default behaviour?

Neither, required, advanced nor hidden. Just a  normal option.

Possible alternatives (meaning "nothing special"): `standard`, `normal`, `simple`, `orderly`, `regular`, `classic`, `usual`.

eseiler: The default is public, because you need to provide it if you use a validator:

```cpp
    parser.add_option(args.sam_path, 'o', "output", "The output SAM file path.",
                      seqan3::option_spec::DEFAULT,
                      seqan3::output_file_validator{seqan3::output_file_open_options::create_new, {"sam"}});
```

Maybe we could use something like `standard` (which is a synonym of default) or even `normal`

marehr: New idea: option_spec has "require" semantics.

The enum names of `seqan3::option_spec` were changed to lower case:
  * `seqan3::option_spec::DEFAULT` is replaced by `seqan3::option_spec::optional` (default-visible and non-required).
  * `seqan3::option_spec::REQUIRED` is replaced by `seqan3::option_spec::required` (default-visible and required).
  * `seqan3::option_spec::ADVANCED` is replaced by `seqan3::option_spec::advanced` (visible and non-required).
  * `seqan3::option_spec::HIDDEN` is replaced by `seqan3::option_spec::hidden` (non-visible and non-required).

smehringer:
Resolution 14-12-2020:

We will name it `seqan3::option_spec::standard` for the default case.

smehringer:
@rrahn I included the namechange and updated the documentation :)

doc/tutorial/argument_parser/index.md
```diff
@@ -308,12 +308,12 @@ Set an option/flag to hidden, if you want to completely hide it from the user. I
...
-|:-----------|:---------------------------------------------------------------|
-| defaulted  | The default tag with no special behaviour.                     |
-| required   | Required options will cause an error if not provided.          |
-| advanced   | Advanced options are only displayed wit `-hh/--advanced-help`. |
-| hidden     | Hidden options are never displayed when exported.              |
+| Tag        | Description                                                                                       |
+|:-----------|:--------------------------------------------------------------------------------------------------|
+| standard   | The default tag (non-required and always visible).                                                |
+| required   | Required options will cause an error if not provided (required and always visible).               |
+| advanced   | Advanced options are only displayed wit `-hh/--advanced-help`. (non-required and partly visible). |
```
rrahn, line 315:
```suggestion
| advanced   | Advanced options are only displayed wit `-hh/--advanced-help`. (non-required and visible on request). |
```

rrahn (approved):
just a minor suggestion for improvement

PR #2286, opened 2020-11-30
[TEST] Add contigous iterator to iterator_test_template 

smehringer:
fixes https://github.com/seqan/product_backlog/issues/94

Adds tests for contiguous iterators to the iterator_test_template.

### DIFF
--- a/include/seqan3/std/memory
+++ b/include/seqan3/std/memory
@@ -1,0 +1,68 @@
+// -*- C++ -*-
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2020, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \brief Provides C++20 additions to the \<memory\> header.
+ * \author Svenja Mehringer <svenja.mehringer AT fu-berlin.de>
+ */
+
+#pragma once
+
+#if __has_include(<version>)
+#include <version> // from C++20 all feature macros should be defined here
+#endif
+
+//!\brief A workaround for __cpp_lib_to_address for gcc version on macOS >=9.0 and < 9.4 (in C++20 mode).
+//!       Those versions implemented std::to_address, but did not define that feature detection macro.
+#ifndef SEQAN3_WORKAROUND_CPP_LIB_TO_ADDRESS_IS_PRESENT
+#   if defined(__cpp_lib_to_address)
+#       define SEQAN3_WORKAROUND_CPP_LIB_TO_ADDRESS_IS_PRESENT 1
+#   elif defined(__GNUC__) && defined(__APPLE__) && (__GNUC__ == 9) && __cplusplus > 201703L
+#       define SEQAN3_WORKAROUND_CPP_LIB_TO_ADDRESS_IS_PRESENT 1
+#   else
+#       define SEQAN3_WORKAROUND_CPP_LIB_TO_ADDRESS_IS_PRESENT 0
+#   endif
+#endif
+
+#include <memory>
+
+#if !SEQAN3_WORKAROUND_CPP_LIB_TO_ADDRESS_IS_PRESENT
+
+/*!\defgroup memory memory
+ * \ingroup std
+ * \brief The \<memory\> header from C++20's standard library.
+ */
+
+namespace std
+{
+
+/*!\brief Obtain the address represented by p without forming a reference to the object pointed to by p.
+ * \sa https://en.cppreference.com/w/cpp/memory/to_address
+ * \ingroup memory
+ */
+template <typename T>
+constexpr T * to_address(T * p) noexcept
+{
+    static_assert(!std::is_function_v<T>);
+    return p;
+}
+
+//!\overload
+template<typename T>
+constexpr auto to_address(const T & p) noexcept
+{
+    if constexpr (requires { std::pointer_traits<T>::to_address(p); })
+        return std::pointer_traits<T>::to_address(p);
+    else
+        return std::to_address(p.operator->());
+}
+
+} // namespace std
+
+#endif // SEQAN3_WORKAROUND_CPP_LIB_TO_ADDRESS_IS_PRESENT
+
--- a/test/unit/range/iterator_test_template.hpp
+++ b/test/unit/range/iterator_test_template.hpp
@@ -10,6 +10,7 @@
 #include <gtest/gtest.h>
 
 #include <seqan3/std/iterator>
+#include <seqan3/std/memory>
 #include <seqan3/std/ranges>
 
 #include <seqan3/core/platform.hpp>
@@ -25,6 +26,7 @@ struct iterator_fixture : public ::testing::Test
                                                                // std::forward_iterator_tag
                                                                // std::bidirectional_iterator_tag
                                                                // std::random_access_iterator_tag
+                                                               // std::contiguous_iterator_tag
 
     static constexpr bool const_iterable = true/false;         // Also test const_iterability. (const begin/end required)
 
@@ -97,6 +99,9 @@ TYPED_TEST_P(iterator_fixture, concept_check)
     EXPECT_EQ(std::random_access_iterator<iterator_type>,
               (std::derived_from<typename TestFixture::iterator_tag, std::random_access_iterator_tag>));
 
+    EXPECT_EQ(std::contiguous_iterator<iterator_type>,
+              (std::derived_from<typename TestFixture::iterator_tag, std::contiguous_iterator_tag>));
+
     if constexpr (TestFixture::const_iterable)
     {
         using const_iterator_type = decltype(std::ranges::cbegin(this->test_range));
@@ -110,12 +115,16 @@ TYPED_TEST_P(iterator_fixture, concept_check)
 
         EXPECT_EQ(std::random_access_iterator<const_iterator_type>,
                   (std::derived_from<typename TestFixture::iterator_tag, std::random_access_iterator_tag>));
+
+        EXPECT_EQ(std::contiguous_iterator<const_iterator_type>,
+                  (std::derived_from<typename TestFixture::iterator_tag, std::contiguous_iterator_tag>));
     }
 
     if (!std::derived_from<typename TestFixture::iterator_tag, std::input_iterator_tag>)
     {
-        FAIL() << "The iterator tag member type must be one of std::input_iterator_tag,"
-               << "std::forward_iterator_tag, std::bidirectional_iterator_tag, or std::random_access_iterator_tag.";
+        FAIL() << "The iterator tag member type must be one of std::input_iterator_tag, "
+               << "std::forward_iterator_tag, std::bidirectional_iterator_tag, std::random_access_iterator_tag, or "
+               << "std::contiguous_iterator_tag.";
     }
 }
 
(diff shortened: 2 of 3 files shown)


test/unit/range/iterator_test_template.hpp
```diff
@@ -642,6 +651,88 @@ TYPED_TEST_P(iterator_fixture, compare_geq)
...
+
+template<class T>
+constexpr auto to_address(const T& p) noexcept
+{
+    if constexpr (requires{ std::pointer_traits<T>::to_address(p); }) {
+        return std::pointer_traits<T>::to_address(p);
+    } else {
+        return to_address(p.operator->());
+    }
+}
```
marehr, line 675: move to `seqan3/std/memory`

marehr: Why is https://en.cppreference.com/w/cpp/memory/addressof not possible?

test/unit/range/iterator_test_template.hpp
```diff
@@ -642,6 +651,88 @@ TYPED_TEST_P(iterator_fixture, compare_geq)
...
+
+TYPED_TEST_P(iterator_fixture, address_test)
+{
+    if constexpr (std::derived_from<typename TestFixture::iterator_tag, std::contiguous_iterator_tag>)
+    {
+        auto begin_it = std::ranges::begin(this->test_range);
+        EXPECT_EQ(to_address(begin_it), std::addressof(*begin_it));
+    }
+
+    if constexpr (std::derived_from<typename TestFixture::iterator_tag, std::random_access_iterator_tag> &&
```
marehr, line 685:
```suggestion
    if constexpr (std::derived_from<typename TestFixture::iterator_tag, std::contiguous_iterator_tag> &&
```

?

test/unit/range/iterator_test_template.hpp
smehringer: > move to seqan3/std/memory

Yes! just testing if all compilers are fine with the implementation. This is rather a draft PR.
And actually it does not work :( I will investigate.

> Why is https://en.cppreference.com/w/cpp/memory/addressof not usable?

I just followed the standard proposal http://eel.is/c++draft/iterator.concept.contiguous

test/unit/range/iterator_test_template.hpp
smehringer: whoops. good catch, thanks.

smehringer: that was also why I got error :slap:

smehringer:
hmm.. @marehr Do you have an idea why MacOS gcc9 c++2a failes although the others pass?

I tried to guard the `std::to_address` feature with 

`#ifndef __cpp_lib_to_address`
or
`#if __cpp_lib_to_address < 201711L`
both with the same result of only macOS failing.. (not sure whats the correct approach. I saw both in other headers)

eseiler:
> hmm.. @marehr Do you have an idea why MacOS gcc9 c++2a failes although the others pass?
> 
> I tried to guard the `std::to_address` feature with
> 
> `#ifndef __cpp_lib_to_address`
> or
> `#if __cpp_lib_to_address < 201711L`
> both with the same result of only macOS failing.. (not sure whats the correct approach. I saw both in other headers)

The feature macro is missing.

The condition would be `__APPLE__ && defined(__GNUC__) && __GNUC__ == 9 && __cplusplus > 201703L`.
Can you please add this missing feature macro in the same style as the one in `include/seqan3/std/type_traits` , i.e. also with the gnuc_minor check.

include/seqan3/std/memory
```diff
@@ -0,0 +1,68 @@
...
+#if __has_include(<version>)
+#include <version> // from C++20 all feature macros should be defined here
+#endif
+
+//!\brief A workaround for __cpp_lib_to_address for gcc version on macOS >=9.0 and < 9.4 (in C++20 mode).
+//!       Those versions implemented std::to_address, but did not define that feature detection macro.
+#ifndef SEQAN3_WORKAROUND_CPP_LIB_TO_ADDRESS
+#   if defined(__cpp_lib_to_address)
+#       define SEQAN3_WORKAROUND_CPP_LIB_TO_ADDRESS 1
+#   elif defined(__GNUC__) && defined(__APPLE__) && ((__GNUC__ == 9) && (__GNUC_MINOR__ < 4)) && __cplusplus > 201703L
```
marehr: I wouldn't give an upper limit for the minor, because we know that the macos gcc 9 does not define that macro. 

Reason:
1. We know that macos gcc 9 is defining that function but forgot to add the version feature macro
2. If macos gcc 9 defines __cpp_lib_to_address in a later minor version it will use the first #2287 if case.

include/seqan3/std/memory
```diff
@@ -54,12 +54,18 @@ constexpr T * to_address(T * p) noexcept
...
-        return std::to_address(p.operator->());
+    return std::pointer_traits<T>::to_address(p);
+}
+
+//!\overload
+template<typename T>
+    requires requires (const T & p) { p.operator->(); }
+constexpr auto to_address(const T & p) noexcept
+{
+    return std::to_address(p.operator->());
```
marehr, line 68: Can you revert it to the original again?

test/unit/range/iterator_test_template.hpp
```diff
@@ -642,6 +651,69 @@ TYPED_TEST_P(iterator_fixture, compare_geq)
...
+    if constexpr (std::derived_from<typename TestFixture::iterator_tag, std::contiguous_iterator_tag> &&
+                  TestFixture::const_iterable)
+    {
+        auto cbegin_it = std::ranges::cbegin(this->test_range);
+        EXPECT_EQ(std::to_address(cbegin_it), std::addressof(*cbegin_it));
+    }
+}
+
+template <typename it_begin_t, typename it_sentinel_t>
+inline void address_difference_test(it_begin_t && it_begin, it_sentinel_t && it_end)
```
SGSSGene, line 675: This is more a general question: Do we add `inline` for template functions that are anyways `inline`?

SGSSGene (approved):
A fine test! :+1: 

A have added a question about `inline` template functions.

test/unit/range/iterator_test_template.hpp
rrahn: I'd say it is not always trivial. But in general we follow this here: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-inline

Even though funciton templates are usually inline: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-inline, we put it in front of free functions and exertnally defined member functions, since we don't worry about ABI. This relieves the programmer to not think about why or why not a function was declared inline.

include/seqan3/std/memory
```diff
@@ -0,0 +1,68 @@
...
+{
+    static_assert(!std::is_function_v<T>);
+    return p;
+}
+
+//!\overload
+template<typename T>
+constexpr auto to_address(const T & p) noexcept
+{
+    if constexpr (requires{ std::pointer_traits<T>::to_address(p); })
```
rrahn, line 59:
```suggestion
    if constexpr (requires { std::pointer_traits<T>::to_address(p); })
```

rrahn: I once had an issue with an older compiler using this expression. But maybe it is fixed now.

include/seqan3/std/memory
```diff
@@ -0,0 +1,68 @@
...
+#   elif defined(__GNUC__) && defined(__APPLE__) && (__GNUC__ == 9) && __cplusplus > 201703L
+#       define SEQAN3_WORKAROUND_CPP_LIB_TO_ADDRESS 1
+#   else
+#       define SEQAN3_WORKAROUND_CPP_LIB_TO_ADDRESS 0
+#   endif
+#endif
+
+#include <memory>
+
+#if !SEQAN3_WORKAROUND_CPP_LIB_TO_ADDRESS
```
rrahn, line 34: this macro defintion is confusing. So if the workaround is set, we do not do the workaround?

test/unit/range/iterator_test_template.hpp
rrahn, line 675:
```suggestion
inline void address_difference_test(it_begin_t it_begin, it_sentinel_t it_end)
```

test/unit/range/iterator_test_template.hpp
```diff
@@ -642,6 +651,69 @@ TYPED_TEST_P(iterator_fixture, compare_geq)
...
+
+    if constexpr (std::derived_from<typename TestFixture::iterator_tag, std::contiguous_iterator_tag> &&
+                  TestFixture::const_iterable)
+    {
+        using difference_t = std::iter_difference_t<decltype(std::ranges::cbegin(this->test_range))>;
+        auto it_cbegin = std::ranges::cbegin(this->test_range);
+        auto it_cend = std::ranges::cend(this->test_range);
+        EXPECT_EQ(std::to_address(it_cend), std::to_address(it_cbegin) + static_cast<difference_t>(it_cend - it_cbegin));
+    }
+}
```
rrahn, line 715: ins't this tested already before implicitly?

rrahn (changes requested):

test/unit/range/iterator_test_template.hpp
SGSSGene: Wouldn't this force a copy of the iterators? or do we assume that iterators are so lean and clean that copying is better?

rrahn: Yes, iterators are intended to be lightweight. In fact all STL algorithms take iterators by copy.

include/seqan3/std/memory
smehringer: I renamed it to `SEQAN3_WORKAROUND_CPP_LIB_TO_ADDRESS_IS_PRESENT`
as it is a seqan3 workaround and it defines whether to_address is present or not.

test/unit/range/iterator_test_template.hpp
smehringer: Oh, Yes you are right. I'll delete the test.

smehringer:
@marehr Can you second review instead of @rrahn?

test/unit/range/iterator_test_template.hpp
```diff
@@ -642,6 +651,69 @@ TYPED_TEST_P(iterator_fixture, compare_geq)
     }
 }
 
+// ---------------------------------------------------------------------------------------------------------------------
+// Contiguous Iterator
+// ---------------------------------------------------------------------------------------------------------------------
+
+TYPED_TEST_P(iterator_fixture, address_test)
```
marehr, line 658:
```suggestion
TYPED_TEST_P(iterator_fixture, address_of_begin)
```

test/unit/range/iterator_test_template.hpp
```diff
@@ -642,6 +651,69 @@ TYPED_TEST_P(iterator_fixture, compare_geq)
...
+    }
+
+    if constexpr (std::derived_from<typename TestFixture::iterator_tag, std::contiguous_iterator_tag> &&
+                  TestFixture::const_iterable)
+    {
+        address_difference_test(std::ranges::cbegin(this->test_range), std::ranges::cend(this->test_range));
+    }
+}
+
+TYPED_TEST_P(iterator_fixture, address_to_end)
```
marehr, line 697:
```suggestion
TYPED_TEST_P(iterator_fixture, address_of_end)
```

Can you move this test after `address_of_begin`?

test/unit/range/iterator_test_template.hpp
```diff
@@ -642,6 +651,69 @@ TYPED_TEST_P(iterator_fixture, compare_geq)
...
+    }
+}
+
+template <typename it_begin_t, typename it_sentinel_t>
+inline void address_difference_test(it_begin_t it_begin, it_sentinel_t it_end)
+{
+    using difference_t = std::iter_difference_t<it_begin_t>;
+
+    for (auto it = it_begin; it != it_end; ++it)
+        EXPECT_EQ(std::to_address(it), std::to_address(it_begin) + static_cast<difference_t>(it - it_begin));
```
marehr, line 680:
```suggestion

    for (auto it = it_begin; it != it_end; ++it)
    {
        EXPECT_EQ(std::to_address(it), std::to_address(it_begin) + static_cast<difference_t>(it - it_begin));
        EXPECT_EQ(std::to_address(it), std::to_address(it_end) - static_cast<difference_t>(it_end - it));
    }
```

I think we should also test the address from the sentinel, since it might not be of the same type as the begin iterator.

test/unit/range/iterator_test_template.hpp
marehr: I like this as it is now. Because we have in the general the following cases:

1. difference between iterator, iterator
2. difference between iterator, sentinel (and vice versa)
3. difference between sentinel, sentinel

So that means we should also have 3 test cases that check that the addresses are contiguous for all these cases.

1. is done by address_difference and address_test
2. is done by address_of_end and by my suggestion for address_difference
3. is done by address_to_end

marehr (changes requested):
Sorry for the extra work, looks really great.

test/unit/range/iterator_test_template.hpp
smehringer: This test is already after address_of_begin?

marehr: Currently we have this order:

```c++
TYPED_TEST_P(iterator_fixture, address_of_begin)
TYPED_TEST_P(iterator_fixture, address_difference)
TYPED_TEST_P(iterator_fixture, address_of_end)
```

I wanted this order:

```c++
TYPED_TEST_P(iterator_fixture, address_of_begin)
TYPED_TEST_P(iterator_fixture, address_of_end)
TYPED_TEST_P(iterator_fixture, address_difference)
```

smehringer: ah now I get it :D

marehr (approved):
Thank you :)

PR #2288, opened 2020-11-30
[MISC] Simplify the configuration constructor.

rrahn:
The constructor now checks via a concept if the config element can be
constructed from the given configuration element.
This is the first step to remove the crtp-pattern from the base
config element.

### DIFF
--- a/include/seqan3/core/configuration/configuration.hpp
+++ b/include/seqan3/core/configuration/configuration.hpp
@@ -13,9 +13,6 @@
 #pragma once
 
 #include <seqan3/std/concepts>
-#include <tuple>
-
-#include <meta/meta.hpp>
 
 #include <seqan3/core/algorithm/pipeable_config_element.hpp>
 #include <seqan3/core/configuration/detail/concept.hpp>
@@ -101,19 +98,17 @@ public:
     ~configuration()                                           = default; //!< Defaulted.
 
     /*!\brief Constructs a configuration from a single configuration element.
-     * \param elem The element to store.
+     * \tparam config_element_t The configuration element to add; must model
+     *                          seqan3::detail::config_element_specialisation.
+     * \param[in] config_element The configuration element to construct the configuration from.
      */
-    template <typename derived_t, typename value_t>
-    constexpr configuration(pipeable_config_element<derived_t, value_t> && elem) :
-        base_type{static_cast<derived_t &&>(std::move(elem))}
-    {}
-
-    /*!\brief Constructs a configuration from a single configuration element.
-     * \param elem The element to store.
-     */
-    template <typename derived_t, typename value_t>
-    constexpr configuration(pipeable_config_element<derived_t, value_t> const & elem) :
-        base_type{static_cast<derived_t const &>(elem)}
+    template <typename config_element_t>
+    //!\cond
+        requires (!std::same_as<std::remove_cvref_t<config_element_t>, configuration>) &&
+                 detail::config_element_specialisation<std::remove_cvref_t<config_element_t>>
+    //!\endcond
+    constexpr configuration(config_element_t && config_element) :
+        base_type{std::forward<config_element_t>(config_element)}
     {}
     //!\}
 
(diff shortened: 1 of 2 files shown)


include/seqan3/core/algorithm/configuration.hpp
```diff
@@ -572,14 +570,11 @@ class configuration : public std::tuple<configs_t...>
...
-/*!\brief Deduces the correct configuration element type from the passed seqan3::pipeable_config_element.
- * \relates seqan3::configuration
- */
-template <typename derived_t, typename value_t>
-configuration(pipeable_config_element<derived_t, value_t> const &) -> configuration<std::remove_cvref_t<derived_t>>;
+template <typename config_t>
+//!\cond
+    requires detail::config_element_specialisation<std::remove_cvref_t<config_t>>
+//!\endcond
+configuration(config_t &&) -> configuration<std::remove_cvref_t<config_t>>;
```
SGSSGene, line 577: This simplification is great.
I think this can pass like this.

One thing I would like to note, (I am not even sure it is legal). Maybe it is possible to remove the `std::remove_cvref_t` in the requires clause of the constructorstor since it is applied by the deduction rule anyways?
so:
```
template <typename config_element_t>
    requires (!std::same_as<config_element_t, configuration>) &&
             detail::config_element_specialisation<config_element_t>
constexpr configuration(config_element_t && config_element) :
    base_type{std::forward<config_element_t>(config_element)}
{}
```

But maybe this is to aggressive type of optimization of code length?

include/seqan3/core/algorithm/configuration.hpp
```diff
@@ -101,19 +101,17 @@ class configuration : public std::tuple<configs_t...>
...
-    /*!\brief Constructs a configuration from a single configuration element.
-     * \param elem The element to store.
-     */
-    template <typename derived_t, typename value_t>
-    constexpr configuration(pipeable_config_element<derived_t, value_t> const & elem) :
-        base_type{static_cast<derived_t const &>(elem)}
+    template <typename config_element_t>
+    //!\cond
+        requires (!std::same_as<std::remove_cvref_t<config_element_t>, configuration>) &&
+                 detail::config_element_specialisation<std::remove_cvref_t<config_element_t>>
```
SGSSGene, line 111:
```suggestion
        requires (!std::same_as<config_element_t, configuration>) &&
                 detail::config_element_specialisation<config_element_t>
```

SGSSGene (changes requested):
This simplification is great.
I think this can pass like this.

One thing I would like to note, (I am not even sure it is legal). Maybe it is possible to remove the `std::remove_cvref_t` in the requires clause of the constructorstor since it is applied by the deduction rule anyways?
so:
```
template <typename config_element_t>
    requires (!std::same_as<config_element_t, configuration>) &&
             detail::config_element_specialisation<config_element_t>
constexpr configuration(config_element_t && config_element) :
    base_type{std::forward<config_element_t>(config_element)}
{}
```

But maybe this is to aggressive type of optimization of code length?

include/seqan3/core/algorithm/configuration.hpp
rrahn: Thank you for the review 👍 
The remove cvref is actually necessary because the function parameter is captured as a forwarding reference. 
This means that it can eiter capture a temporary as an rvalue reference `obj &&` or an lvalue as regular reference `obj &`. Now the real fun starts with the reference collapsing rules. If `config_element_t` (or I use it here short `t`), is an lvalue reference we have following substitution `t &` `&& obj` which collapses to `t &` `obj` and the reference is part of the template parameter. If it is an rvalue reference we have `t &&` `&& obj` which collapses to `t` `&& obj` so the template parameter does not contain a reference anymore. Accordingly, the interface needs to remove the reference to compare it with the concrete type `configuration` which neither contains const nor references. Also the config_element_specialisation concept does not strip away the reference, since that has semantically different menaings. The most obvious one is that a reference is not an object, such that an object has member functions but a reference does not. So one requirement of that concept is that the type is semiregular which is not true for references. You could argue, that the concept could remove this, but then the caller cannot differentiate whether he has a pure object or a reference thereof. So if the caller would create an default constructed instance of that type, it would not work anymore, since a reference cannot be default constructed. So the concept would give the wrong impression.

SGSSGene: Yes! you are right.
I have mixed up class template with the template of the constructor. For what ever reason I thought the type deduction determines the constructor template, but it only decides the class template.

SGSSGene (approved):

include/seqan3/core/algorithm/configuration.hpp
smehringer, line 577: Do deduction guides need constraints? :thinking:

smehringer (approved):

include/seqan3/core/algorithm/configuration.hpp
rrahn: They are hypothetical function templates which also participate in overload resolution. There are some rules about when a deduction guide is preferred over a regular constructor, but this is not important here. But I can remove the clutter by just taking the arugment by value such that I don't have to explicitly decay the type.

smehringer: Interesting! Thanks for the explanation.

PR #2290, opened 2020-12-01
[FEATURE] Alphabet: Add the quality alphabet seqan3::phred94

Irallia:
Resolves https://github.com/seqan/product_backlog/issues/267

The task _'Check if the tutorials need an update.'_ did not require any work, because there is no tutorial yet. I will add a tutorial when solving https://github.com/seqan/product_backlog/issues/266.
For task _'Update the documentation on the quality landing page'_ I left three comments for the reviewer, which I will remove after discussion.

I splittet the PR in several commits and would recommend to review commit wise.

PR #2291, opened 2020-12-04
[DOC] Refactor phred snippets. The int cast is not needed.

Irallia:
This is a follow up of https://github.com/seqan/seqan3/pull/2290
and connected to https://github.com/seqan/product_backlog/issues/266.
The cast to integer is not needed for a better output. This is automaticly done with the debug_stream.


marehr (approved):
Thank you and very nice :))

smehringer (approved):
needs a rebase though :)

PR #2299, opened 2020-12-11
[FIX] too many columns in doxygen class list

eseiler:
@Irallia noticed that our doxygen class list is not well formatted:

https://docs.seqan.de/seqan/3-master-dev/classes.html

After this PR:

fix_doxy

See https://www.doxygen.nl/manual/config.html#cfg_cols_in_alpha_index

### DIFF
--- a/test/documentation/seqan3_doxygen_cfg.in
+++ b/test/documentation/seqan3_doxygen_cfg.in
@@ -63,6 +63,8 @@ EXCLUDE_SYMBOLS        = seqan3::contrib
 FORMULA_FONTSIZE       = 14
 HTML_DYNAMIC_SECTIONS  = YES
 
+COLS_IN_ALPHA_INDEX    = 1
+
 ## detect headers without extensions (in std module)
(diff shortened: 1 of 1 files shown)


Irallia (approved):
Thanks for the fast solution!

smehringer (commented):
Just a question. Doxygen documents this parameter with a note:

```
This tag requires that the tag ALPHABETICAL_INDEX is set to YES. 
```

We do not set this (as far as I can tell). 
Since your solution seems to work, it's not a big deal but can you maybe double check what happens?

eseiler:
> Just a question. Doxygen documents this parameter with a note:
> 
> ```
> This tag requires that the tag ALPHABETICAL_INDEX is set to YES. 
> ```
> 
> We do not set this (as far as I can tell).
> Since your solution seems to work, it's not a big deal but can you maybe double check what happens?

The default is `YES` and we only put the non-default stuff in our config :)

smehringer (approved):

Issue #2234, opened 2020-11-02
latest googletest (1.10.0.20201025-1) doesn't like the *test_template.hpp


mr-c:
## Platform

  - SeqAn version: 3.0.2+ds-3
  - Operating system: Debian Unstable
  - Compiler: GNU 10.2.0

## Description

Tests fail with a recent googletest git snapshot. Seems the are no longer doing releases?

This is affecting Debian's package of seqan3 :-( https://bugs.debian.org/973001

### How to repeat the problem

Build seqan3 tests with https://github.com/google/googletest/tree/a6dfd3aca7f2f91f95fc7ab650c95a48420d513d

### Expected behaviour

Tests pass

### Actual behaviour

```
3107/7128 Test #3107: alphabet/cigar/cigar_test::GoogleTestVerification.UninstantiatedTypeParameterizedTestSuite<alphabet> ..........................................................................................................................................................................................................................................................................................................***Failed    0.32 sec
Running main() from /usr/src/googletest/googletest/src/gtest_main.cc
Note: Google Test filter = GoogleTestVerification.UninstantiatedTypeParameterizedTestSuite<alphabet>
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from GoogleTestVerification
[ RUN      ] GoogleTestVerification.UninstantiatedTypeParameterizedTestSuite<alphabet>
/tmp/autopkgtest-lxc.7229b7jl/downtmp/autopkgtest_tmp/unit/alphabet/cigar/../alphabet_test_template.hpp:86: Failure
Type parameterized test suite alphabet is defined via REGISTER_TYPED_TEST_SUITE_P, but never instantiated via INSTANTIATE_TYPED_TEST_SUITE_P. None of the test cases will run.

Ideally, TYPED_TEST_P definitions should only ever be included as part of binaries that intend to use them. (As opposed to, for example, being placed in a library that may be linked in to get other utilities.)

To suppress this error for this test suite, insert the following line (in a non-header) in the namespace it is defined in:

GTEST_ALLOW_UNINSTANTIATED_PARAMETERIZED_TEST(alphabet);
[  FAILED  ] GoogleTestVerification.UninstantiatedTypeParameterizedTestSuite<alphabet> (0 ms)
[----------] 1 test from GoogleTestVerification (0 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (0 ms total)
[  PASSED  ] 0 tests.
[...]

PR #2301, opened 2020-12-13
[FIX] #2234: support latest googletest

marehr:
See https://github.com/seqan/seqan3/issues/2234#issuecomment-720476694

### DIFF
--- a/test/unit/test/pretty_printing_test.cpp
+++ b/test/unit/test/pretty_printing_test.cpp
@@ -45,7 +45,7 @@ TEST(pretty_printing, gtest_output)
     EXPECT_EQ(gtest_str(std::make_tuple<int, int>(42, -10)), "(42, -10)"s);
     EXPECT_EQ(debug_str(std::make_tuple<int, int>(42, -10)), "(42,-10)"s);
 
-#if GTEST_INTERNAL_HAS_VARIANT // googletest >1.10.0
+#if GTEST_INTERNAL_HAS_VARIANT // googletest at head (>=1.10.x)
     EXPECT_EQ(gtest_str(std::variant<int>{0}), "('int(index = 0)' with value 0)"s);
 #else // googletest <= 1.10.0
     EXPECT_EQ(gtest_str(std::variant<int>{0}), "0"s);
--- a/test/unit/test/pretty_printing_test.cpp
+++ b/test/unit/test/pretty_printing_test.cpp
@@ -52,7 +52,7 @@ TEST(pretty_printing, gtest_output)
 #endif
     EXPECT_EQ(debug_str(std::variant<int>{0}), "0"s);
 
-#if GTEST_INTERNAL_HAS_OPTIONAL // googletest >1.10.0
+#if GTEST_INTERNAL_HAS_OPTIONAL // googletest at head (>=1.10.x)
     EXPECT_EQ(gtest_str(std::optional<int>{}), "(nullopt)"s);
 #else // googletest <= 1.10.0
     EXPECT_EQ(gtest_str(std::optional<int>{}), "<VALUELESS_OPTIONAL>"s);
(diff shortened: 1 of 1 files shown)


test/unit/test/pretty_printing_test.cpp
```diff
@@ -44,18 +44,27 @@ TEST(pretty_printing, gtest_output)
...
+    EXPECT_EQ(gtest_str(std::optional<int>{}), "(nullopt)"s);
+#else // googletest <= 1.10.0
     EXPECT_EQ(gtest_str(std::optional<int>{}), "<VALUELESS_OPTIONAL>"s);
+#endif
     EXPECT_EQ(debug_str(std::optional<int>{}), "<VALUELESS_OPTIONAL>"s);
+}
+
+TEST(pretty_printing, std_output)
+{
+    // if any of these tests get supported by googletest move them to gtest_output
```
SGSSGene, line 65: Is there a common understanding of what it means that these test are supported by googletest?

To me it seems like it is already support. It seems like I am missing something.

test/unit/test/pretty_printing_test.cpp
```diff
@@ -44,18 +44,27 @@ TEST(pretty_printing, gtest_output)
...
     EXPECT_EQ(gtest_str(std::make_tuple<int, int>(42, -10)), "(42, -10)"s);
     EXPECT_EQ(debug_str(std::make_tuple<int, int>(42, -10)), "(42,-10)"s);
-}
-
-TEST(pretty_printing, std_output)
-{
-    EXPECT_EQ(gtest_str(std::vector<std::vector<int>>{{0,1}, {2,3}, {1,2}, {0}}), "[[0,1],[2,3],[1,2],[0]]"s);
-    EXPECT_EQ(debug_str(std::vector<std::vector<int>>{{0,1}, {2,3}, {1,2}, {0}}), "[[0,1],[2,3],[1,2],[0]]"s);
 
+#if GTEST_INTERNAL_HAS_VARIANT // googletest >1.10.0
```
SGSSGene, line 48: I am not sure of googletest version semantics.
`GTEST_INTERNAL_HAS_VARIANT` is in there `master` branch, but I can't find it in there `v1.10.x` branch.

If I had to guess, I would say it only appears in `googletest >= 1.11.0`
and it will not be in any `googletest < 1.11.0` version. So if for what ever reason there is a 1.10.1 version I would assume they will not include it?

SGSSGene (commented):
Nice PR!
I am not sure about the googletest versioning. I also couldn't find any information on it.
Maybe it is better to leave the `googletest > 1.10.0` away.

test/unit/test/pretty_printing_test.cpp
```diff
@@ -44,18 +44,27 @@ TEST(pretty_printing, gtest_output)
...
+    EXPECT_EQ(gtest_str(std::variant<int>{0}), "('int(index = 0)' with value 0)"s);
+#else // googletest <= 1.10.0
     EXPECT_EQ(gtest_str(std::variant<int>{0}), "0"s);
+#endif
     EXPECT_EQ(debug_str(std::variant<int>{0}), "0"s);
 
+#if GTEST_INTERNAL_HAS_OPTIONAL // googletest >1.10.0
```
SGSSGene, line 55:
```suggestion
#if GTEST_INTERNAL_HAS_OPTIONAL // googletest at head (>=1.10.x)
```

test/unit/test/pretty_printing_test.cpp
SGSSGene, line 48:
```suggestion
#if GTEST_INTERNAL_HAS_VARIANT // googletest at head (>=1.10.x)
```

SGSSGene (approved):
Let googles "at head" strategy not break our code!

test/unit/test/pretty_printing_test.cpp
smehringer: > To me it seems like it is already support. It seems like I am missing something.

There are indirectly supported through our debug_stream (hacky hack). That's why the output for debug stream and google test look alike. But it could be that gtest will handle the output on it's own at some point and then it whould be moved.

smehringer (approved):

PR #2302, opened 2020-12-14
[FIX] change `single_pass_input and `async_input_buffer` move-only

SGSSGene:
- Removes `single_pass_input` and `async_input_buffer` copy-constructor and assignment-operator.
- Adjust unit test to work with move-only views.

### DIFF
--- a/test/unit/range/views/async_input_buffer_test.cpp
+++ b/test/unit/range/views/async_input_buffer_test.cpp
@@ -82,7 +82,7 @@ TEST(async_input_buffer, destruct_with_full_buffer)
     auto v0 = vec | seqan3::views::single_pass_input;
 
     {
-        auto v1 = v0 | seqan3::views::async_input_buffer(5);
+        auto v1 = std::move(v0) | seqan3::views::async_input_buffer(5);
 
         // consume five elements (construction already consumes one)
         auto b = std::ranges::begin(v1);
@@ -96,7 +96,7 @@ TEST(async_input_buffer, destruct_with_full_buffer)
         std::this_thread::sleep_for(std::chrono::milliseconds(100));
     } // thread sync at destruction of v1; tests working destruction with full buffer
 
-    EXPECT_GE(std::ranges::distance(v0), 17); // total of at most 10 chars consumed
+    EXPECT_GE(std::ranges::distance(vec), 17); // total of at most 10 chars consumed
 }
 
 TEST(async_input_buffer, combinability)
--- a/test/unit/range/views/async_input_buffer_test.cpp
+++ b/test/unit/range/views/async_input_buffer_test.cpp
@@ -75,6 +75,7 @@ TEST(async_input_buffer, buffer_size_huge)
     EXPECT_RANGE_EQ(vec, v);
 }
 
+#if 0
 TEST(async_input_buffer, destruct_with_full_buffer)
 {
     seqan3::dna4_vector vec{"ACGTACGTACGTATCGAGAGCTTTAGC"_dna4};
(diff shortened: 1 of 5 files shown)


test/unit/range/views/single_pass_input_test.cpp
```diff
@@ -180,12 +180,15 @@ TYPED_TEST(single_pass_input, view_iterate)
...
         TypeParam tmp{this->cmp_data};
-        auto zipper = seqan3::views::zip(tmp, std::move(view));
-        for (auto it = zipper.begin(); it != zipper.end(); ++it)
+        auto tmp_it = tmp.begin();
+        for (auto && elem : view)
         {
-            EXPECT_EQ(std::get<0>(*it), std::get<1>(*it));
+            EXPECT_EQ(elem, *tmp_it);
+            ++tmp_it;
         }
```
eseiler, line 192: Can you check if `EXPECT_RANGE_EQ` works?
something like `EXPECT_RANGE_EQ(tmp, std::move(view))`

SGSSGene: Awesome! Works perfectly. I used it without calling `std::move`. Is there any pros/cons to this?

SGSSGene:
There is a test: where I am not sure of its meaning and if it shouldn't actually fail:

```
TYPED_TEST(single_pass_input, iterator_construction)
{
    using view_type = seqan3::detail::single_pass_input_view<
                        decltype(std::declval<TypeParam>() | seqan3::views::persist)>;
    using iterator_type = std::ranges::iterator_t<view_type>;
    EXPECT_TRUE(std::is_default_constructible_v<iterator_type>);
    EXPECT_TRUE(std::is_copy_constructible_v<iterator_type>);
    EXPECT_TRUE(std::is_move_constructible_v<iterator_type>);
    EXPECT_TRUE(std::is_copy_assignable_v<iterator_type>);
    EXPECT_TRUE(std::is_move_assignable_v<iterator_type>);
    EXPECT_TRUE(std::is_destructible_v<iterator_type>);
}
```

test/unit/range/views/single_pass_input_test.cpp
eseiler: No, just less clutter and it covers some different cases.
You can check the implementation if you want to: test/include/seqan3/test/expect_range_eq.hpp

test/unit/range/views/async_input_buffer_test.cpp
```diff
@@ -96,7 +96,7 @@ TEST(async_input_buffer, destruct_with_full_buffer)
         std::this_thread::sleep_for(std::chrono::milliseconds(100));
     } // thread sync at destruction of v1; tests working destruction with full buffer
 
-    EXPECT_GE(std::ranges::distance(v0), 17); // total of at most 10 chars consumed
+    EXPECT_GE(std::ranges::distance(vec), 17); // total of at most 10 chars consumed
```
marehr, line 99: @rrahn Can you give directions how to change this test?

eseiler:
### 21-12-2020
Only make move-only if we use the cpp20 implementation (and not the ranges lib).

test/unit/range/views/async_input_buffer_test.cpp
SGSSGene: I have worked a little more on this, but still can't get an handle on it.
my thoughts:
a) ez! 
b) where should the dummy variable sit? and what should it measure?
c) This doesn't work, since the elements them self don't get moved. There are only references pointed to them passed around.

marehr:
We decided to postpone this https://github.com/seqan/product_backlog/issues/263#issuecomment-766681268

PR #2304, opened 2020-12-15
[DOC] Replace the word 'defaulted' with better/correct alternative.

smehringer:
`defaulted` as an adjective means something like `fails to fulfil an obligation`. 
We sometimes do not use this word correctly (since we always mean `set to default`).

Only `to be defaulted` actually means `set to default`. To avoid this possible confusing, some correct sentences were also changed.

### DIFF
--- a/include/seqan3/alphabet/composite/alphabet_variant.hpp
+++ b/include/seqan3/alphabet/composite/alphabet_variant.hpp
@@ -268,7 +268,7 @@ public:
      *
      * \details
      *
-     * Most assignments happen through implicit conversion and the defaulted assignment operator. This is for the rest.
+     * Most assignments happen through implicit conversion and the default assignment operator. This is for the rest.
      */
     template <typename indirect_alternative_t>
     //!\cond
--- a/include/seqan3/alphabet/composite/alphabet_variant.hpp
+++ b/include/seqan3/alphabet/composite/alphabet_variant.hpp
@@ -268,7 +268,7 @@ public:
      *
      * \details
      *
-     * Most assignments happen through implicit conversion and the default assignment operator. This is for the rest.
+     * Most assignments happen through implicit conversion and the default assignment operator. This constructor is for the rest.
      */
     template <typename indirect_alternative_t>
     //!\cond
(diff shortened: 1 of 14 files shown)


include/seqan3/alphabet/composite/alphabet_variant.hpp
```diff
@@ -268,7 +268,7 @@ class alphabet_variant : public alphabet_base<alphabet_variant<alternative_types
      *
      * \details
      *
-     * Most assignments happen through implicit conversion and the defaulted assignment operator. This is for the rest.
+     * Most assignments happen through implicit conversion and the default assignment operator. This is for the rest.
```
Irallia, line 271:
```suggestion
     * Most assignments happen through implicit conversion and the default assignment operator. This alphabet_variant is for the rest.
```

Irallia (approved):
LGFM, just one word I would add, for a bit more meaning 💅

include/seqan3/io/detail/misc_input.hpp
```diff
@@ -71,7 +71,7 @@ inline bool starts_with(ref_t && reference, query_t && query)
 /*!\brief Depending on the magic bytes of the given stream, return a decompression stream or forward the primary stream.
  * \param[in] primary_stream The primary (device) stream for reading.
  * \param[in,out] filename  The associated filename; compression extensions will be stripped. [optional]
- * \returns A pointer to the secondary stream with defaulted or NOP'ed deleter.
+ * \returns A pointer to the secondary stream with a default deleter or a NOP'ed deleter.
```
rrahn, line 74:
```suggestion
 * \returns A pointer to the secondary stream with a default deleter or a nop-deleter.
```

include/seqan3/io/detail/misc_output.hpp
```diff
@@ -32,7 +32,7 @@ namespace seqan3::detail
 /*!\brief Depending on the given filename/extension, create a compression stream or just forward the primary stream.
  * \param[in] primary_stream The primary (uncompressed) stream for writing.
  * \param[in,out] filename  The associated filename; compression extensions will be stripped.
- * \returns A pointer to the secondary stream with defaulted or NOP'ed deleter.
+ * \returns A pointer to the secondary stream with a default deleter or a NOP'ed deleter.
```
rrahn, line 35:
```suggestion
 * \returns A pointer to the secondary stream with a default deleter or a nop-deleter.
```

rrahn (approved):

PR #2305, opened 2020-12-15
[FIX] too many threads used in unit tests

eseiler:
This sets the default number of threads for the `execution_handler_parallel` to `1`.
All the tests were the handler was expected to use multiple threads were adjusted to explicitly request not more than 4 threads.

The `execution_handler_parallel` is detail and hence not used in public code. It handles the thread pool when, e.g., using the alignment or search algorithms in parallel via the config. This config requires a value (no default), hence the number of threads is always set by the user.

When we use an algorithm in parallel, we also default construct a `execution_handler_parallel` along the way. If the default is set to use all threads, we have to generate the thread pool and a queue. However, this default constructed `execution_handler_parallel` is also moved away and practically immediately destructed, hence we need to close the queue.
Closing a queue when using many threads currently may lead to a live/deadlock.

### DIFF
--- a/include/seqan3/core/algorithm/detail/execution_handler_parallel.hpp
+++ b/include/seqan3/core/algorithm/detail/execution_handler_parallel.hpp
@@ -89,7 +89,20 @@ public:
         }
     }
 
-    //!\brief Constructs the execution handler spawning std::thread::hardware_concurrency many threads.
+    /*!\brief Constructs the execution handler spawning 1 thread.
+     *
+     * \details
+     * 
+     * ### Why only 1 thread?
+     * 
+     * This class is not public. It handles the thread pool when, e.g., using the alignment or search algorithms in 
+     * parallel via the config. This config requires a value (no default), hence the number of threads is always 
+     * set by the user.
+     * 
+     * When we use an algorithm in parallel, we also default construct a execution_handler_parallel along the way. If 
+     * the default is set to use all threads, we have to generate the thread pool and a queue. However, this default 
+     * constructed execution_handler_parallel is immediately moved away and destructed.
+     */
     execution_handler_parallel() : execution_handler_parallel{1u}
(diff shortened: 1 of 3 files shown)


include/seqan3/core/algorithm/detail/execution_handler_parallel.hpp
```diff
@@ -90,7 +90,7 @@ class execution_handler_parallel
     }
 
     //!\brief Constructs the execution handler spawning std::thread::hardware_concurrency many threads.
-    execution_handler_parallel() : execution_handler_parallel{std::thread::hardware_concurrency()}
+    execution_handler_parallel() : execution_handler_parallel{1u}
```
joergi-w, line 93: Please correct the `\brief` documentation and add (e.g. in \detail) the reasoning you wrote in this PR why we set the default to 1.

joergi-w (changes requested):

include/seqan3/core/algorithm/detail/execution_handler_parallel.hpp
eseiler: I added it. I didn't add the sentence about live/deadlock since it is a bug and should be fixed soon™

joergi-w (approved):
Thank you :)

smehringer:
I didn't follow the long discussions on this, so it might be better if @rrahn reviews this?

eseiler:
> I didn't follow the long discussions on this, so it might be better if @rrahn reviews this?

If you want to, yes. But the discussion on this was: "We should set it to 1 because it is always set by the user, and using all threads has unintended side effects". And the rest of the reasoning is in the description :D
The queue discussions, on the other hand, were quite long.

smehringer (approved):
Alright then, looks good to me :)

PR #2308, opened 2020-12-18
[FIX] chain_validator: use common_type to chain

marehr:
Part of https://github.com/seqan/product_backlog/issues/246 and https://github.com/seqan/product_backlog/issues/40

This reduces dependencies on seqan3 by removing `seqan3::detail::is_char_adaptation_v`.

Blocked by #2307

### DIFF
--- a/test/unit/argument_parser/format_parse_validators_test.cpp
+++ b/test/unit/argument_parser/format_parse_validators_test.cpp
@@ -1134,6 +1134,84 @@ TEST(validator_test, regex_validator_error)
     EXPECT_THROW(parser4.parse(), seqan3::validation_error);
 }
 
+TEST(validator_test, chaining_validators_common_type)
+{
+    // chaining integral options stay integral
+    {
+        int max_int = std::numeric_limits<int>::max();
+        std::vector v_int{1, 2, 3, max_int};
+        std::vector v_unsigned{4u, static_cast<unsigned>(max_int)};
+
+        EXPECT_TRUE((std::same_as<std::vector<int>, decltype(v_int)>));
+        EXPECT_TRUE((std::same_as<std::vector<unsigned>, decltype(v_unsigned)>));
+
+        seqan3::value_list_validator validator_int{v_int};
+        seqan3::value_list_validator validator_unsigned{v_unsigned};
+
+        EXPECT_TRUE((std::same_as<seqan3::value_list_validator<int>, decltype(validator_int)>));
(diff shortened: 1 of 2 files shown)


test/unit/argument_parser/format_parse_validators_test.cpp
```diff
@@ -825,16 +825,16 @@ TEST(validator_test, value_list_validator_success)
     // type deduction
     // --------------
     // all arithmetic types are deduced to double in order to easily allow chaining of arithmetic validators
-    EXPECT_TRUE((std::same_as<seqan3::value_list_validator<double>,
+    EXPECT_TRUE((std::same_as<seqan3::value_list_validator<int>,
```
SGSSGene, line 828: This seems to contradict the previous comment

test/unit/argument_parser/format_parse_validators_test.cpp
```diff
@@ -1134,6 +1134,84 @@ TEST(validator_test, regex_validator_error)
...
+        EXPECT_NO_THROW(validator_int(max_int));
+        EXPECT_NO_THROW(validator_unsigned(max_int));
+        EXPECT_NO_THROW(validator(max_int));
+    }
+
+    // chaining mixed arithmetic options will be highest common arithmetic type
+    {
+        // note: this number is not representable by double and multiple integer values represent the same double value
+        int64_t max_int64 = std::numeric_limits<int64_t>::max();
+        EXPECT_EQ(static_cast<double>(max_int64), static_cast<double>(max_int64 - 1));
```
SGSSGene, line 1172: I didn't know this! This is the weirdest line of code. But nice!

SGSSGene (changes requested):
I like common_type_t and common_with! havn't used these type traits before.
I have little trouble understanding what common_type_t does in detail. Maybe we can discuss this?

Only changes is a "double" -> "int" change, where the comment is wrong.

test/unit/argument_parser/format_parse_validators_test.cpp
marehr: I changed it to "all arithmetic types are deduced to their common type in order to easily allow chaining of arithmetic validators"

test/unit/argument_parser/format_parse_validators_test.cpp
marehr: It really is a weird line of code; but I hope the comment could help to understand it :)

SGSSGene (approved):
Nice! :-)

eseiler (approved):

PR #2309, opened 2020-12-18
[FIX] non-terminating tests

### LINKED ISSUE #2282: High thread counts lead to non-terminating tests
## Platform

  - SeqAn version: 
  - Operating system: 
  - Compiler: 

## Description

After https://github.com/seqan/seqan3/pull/2276, tests and snippets using the `parallel_queue` do not terminate if there are too many threads accessing the queue.

If there are 160 threads, there is one thread trying to close the queue, while there are 159 threads acquiring the same lock to check if the queue is closed.

In the parallel align_pairwise, you can track it to 
https://github.com/seqan/seqan3/blob/d37ba83a333992a551ec835097e564d8b9a00604/include/seqan3/core/algorithm/detail/algorithm_executor_blocking.hpp#L236
which then calls
https://github.com/seqan/seqan3/blob/d37ba83a333992a551ec835097e564d8b9a00604/include/seqan3/core/algorithm/detail/execution_handler_parallel.hpp#L195

eseiler:
Applies the patch from https://github.com/seqan/seqan3/issues/2282

### DIFF
--- a/include/seqan3/contrib/parallel/buffer_queue.hpp
+++ b/include/seqan3/contrib/parallel/buffer_queue.hpp
@@ -168,15 +168,18 @@ public:
         value_type value{};
         for (;;)
         {
-            auto status = try_pop(value);
+            if (!writer_waiting.load())
+            {
+                auto status = try_pop(value);
 
-            if (status == queue_op_status::closed)
-                throw queue_op_status::closed;
-            else if (status == queue_op_status::success)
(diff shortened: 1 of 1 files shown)


smehringer (approved):

marehr (approved):
As I said in the core chat, we should definitely figure out, if this is the real fix.

PR #2311, opened 2020-12-18
[FIX] type_name_as_string: handle nullptr construction if __cxa_demangle fails

marehr:
On gcc-9 the following type can't be demangled by type_name_as_string

```c++
inline constexpr auto on_result_caller = [] (auto &&) {};

seqan3::configuration cfg{seqan3::search_cfg::on_result<decltype(on_result_caller)>{}};

std::cout << seqan3::detail::type_name_as_string<decltype(cfg)> << std::endl;
// prints: N6seqan313configurationIJNS_10search_cfg9on_resultIKUlOT_E_EEEEE (abi::__cxa_demangle error status (-2): mangled_name is not a valid name under the C++ ABI mangling rules.)
```

Before this commit `abi::__cxa_demangle` returned a `nullptr` and that will fail to execute the binary, because `std::string` can't be constructed from a nullptr.

This commit will handle `abi::__cxa_demangle` errors and will produce a similar output to the above.

### DIFF
--- a/include/seqan3/utility/detail/type_name_as_string.hpp
+++ b/include/seqan3/utility/detail/type_name_as_string.hpp
@@ -47,9 +47,22 @@ inline std::string const type_name_as_string = [] ()
 #if defined(__GNUC__) || defined(__clang__) // clang and gcc only return a mangled name.
     using safe_ptr_t = std::unique_ptr<char, std::function<void(char *)>>;
 
+    // https://gcc.gnu.org/onlinedocs/libstdc++/libstdc++-html-USERS-4.3/a01696.html
     int status{};
     safe_ptr_t demangled_name_ptr{abi::__cxa_demangle(typeid(type).name(), 0, 0, &status),
                                   [] (char * name_ptr) { free(name_ptr); }};
+
+    // We exclude status != 0, because this code can't be reached normally, only if there is a defect in the compiler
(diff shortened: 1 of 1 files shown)


MitraDarja (approved):

include/seqan3/utility/detail/type_name_as_string.hpp
```diff
@@ -47,9 +47,22 @@ inline std::string const type_name_as_string = [] ()
 #if defined(__GNUC__) || defined(__clang__) // clang and gcc only return a mangled name.
     using safe_ptr_t = std::unique_ptr<char, std::function<void(char *)>>;
 
+    // https://gcc.gnu.org/onlinedocs/libstdc++/libstdc++-html-USERS-4.3/a01696.html
     int status{};
     safe_ptr_t demangled_name_ptr{abi::__cxa_demangle(typeid(type).name(), 0, 0, &status),
                                   [] (char * name_ptr) { free(name_ptr); }};
+
+    // We exclude status != 0, because this code can't be reached normally, only if there is a defect in the compiler
+    // itself, since the type is directly given by the compiler. see https://github.com/seqan/seqan3/pull/2311
```
smehringer, line 56:
```suggestion
    // itself, since the type is directly given by the compiler. See https://github.com/seqan/seqan3/pull/2311.
```

The PR can also be seen when using git blame. But I don't mind referencing it anyway

smehringer (approved):

PR #2312, opened 2020-12-18
[TEST] new EXPECT_SAME_TYPE test macro

marehr:
New EXPECT_SAME_TYPE test macro, see seqan/product_backlog#46

```c++
EXPECT_SAME_TYPE(int, float);
```

has the same semantic as writing

```c++
EXPECT_TRUE((std::same_as<int, float>))
```

but it will produce better diagnostics if the test failed:

```
  int
    Which is: "int"
  float
    Which is: "float"
```

### DIFF
--- a/test/include/seqan3/test/expect_same_type.hpp
+++ b/test/include/seqan3/test/expect_same_type.hpp
@@ -1,0 +1,62 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2020, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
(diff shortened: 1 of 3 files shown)


MitraDarja (approved):
LGTM. :+1:

test/include/seqan3/test/expect_same_type.hpp
```diff
@@ -0,0 +1,61 @@
...
+                                          std::type_identity<lhs_t>, std::type_identity<rhs_t>)
+    {
+        auto remove_wrap_type_identity = [](std::string str)
+        {
+            // EXPECT_SAME_TYPE_DEPAREN adds a space after the prefix
+            std::string prefix = "std::type_identity< ";
+            std::string suffix = ">{}";
+
+            size_t str_start = str.find(prefix) + prefix.size();
+            size_t str_end = str.rfind(suffix);
```
smehringer, line 46: Just wondering if you want to add an assert or some other security check in case prefix or suffix cannot be found in the string?

smehringer (approved):

test/include/seqan3/test/expect_same_type.hpp
marehr:
```suggestion
            size_t str_start = str.find(prefix) + prefix.size();
            size_t str_end = str.rfind(suffix);
            assert(str_end >= str_start);
```

PR #2318, opened 2020-12-23
[API] Deprecate is_in_alphabet

joergi-w:
I have replaced all usages (only I/O module) by `seqan3::char_is_valid_for` and fixed the error messages. Furthermore, I removed the relevant snippet and unit tests. 

A problem was that I could not use a deprecated `detail::is_in_alphabet_type` to preserve the functionality, so I removed the detail code and return always true. 

This PR closes https://github.com/seqan/product_backlog/issues/262.


include/seqan3/io/sequence_file/format_fasta.hpp
```diff
@@ -269,7 +271,7 @@ class format_fasta
 
         if constexpr (!detail::decays_to_ignore_v<seq_type>)
         {
-            auto constexpr not_in_alph = !is_in_alphabet<seq_legal_alph_type>;
+            auto constexpr is_legal_alph = char_is_valid_for<seq_legal_alph_type>;
```
MitraDarja, line 274: Just out of curiosity, why the change from the not? Is that the better practice?

include/seqan3/utility/char_operations/predicate.hpp
```diff
@@ -55,16 +54,13 @@ inline constexpr auto is_in_interval = detail::is_in_interval_type<interval_firs
...
- * \include test/snippet/utility/char_operations/char_predicate_is_in_alphabet.cpp
  * \hideinitializer
+ * \deprecated Use seqan3::char_is_valid_for instead.
  */
-template <alphabet alphabet_t>
-inline constexpr auto is_in_alphabet = detail::is_in_alphabet_type<alphabet_t>{};
+template <typename alphabet_t>
+SEQAN3_DEPRECATED_310 inline constexpr bool is_in_alphabet = true;
 
 /*!\brief Checks whether a given letter is the same as the template non-type argument.
```
MitraDarja, line 65: Should that not be deleted?

MitraDarja (changes requested):
Looks good to me, just one minor thing in the comments and one question. Plus, why did you not replace the tests with seqan3::char_is_valid_for?

include/seqan3/io/sequence_file/format_fasta.hpp
joergi-w: **Technical reason:** The `char_is_valid_for` holds a function (with return type bool) and cannot be negated, while the `is_in_alphabet` was inherited from `char_predicate_base` and has the [not-operator](http://docs.seqan.de/seqan/3-master-dev/structseqan3_1_1detail_1_1char__predicate__base.html#aaa9a323dabf6386ebd0b0eb14a51d854) implemented. 

**Practical reason:** All other formats implement the `is_legal_alph` version and for consistency it should be changed. Additionally, the error message has been wrong here: _"Encountered an unexpected letter: is_in_alphabet\<seqan3::rna4\> evaluated to true on 'N'."_

include/seqan3/utility/char_operations/predicate.hpp
joergi-w: In the issue it states **deprecate** seqan3::is_in_alphabet, so I did not delete it.

joergi-w:
> Plus, why did you not replace the tests with seqan3::char_is_valid_for?

We have already unit tests for `char_is_valid_for` for general alphabets in _test/unit/alphabet/alphabet_test_template.hpp_ and for gapped alphabets in _test/unit/alphabet/gap/gapped_test.cpp_ and _test/unit/alphabet/composite/alphabet_variant_test.cpp_. 
The snippets are in _test/snippet/alphabet/char_is_valid_for.cpp_.

Additional features like `msg` do not exist for `char_is_valid_for` and therefore I decided to delete the `is_in_alphabet` tests.

MitraDarja (approved):
Thanks for the clearification. :)

include/seqan3/utility/char_operations/predicate.hpp
```diff
@@ -55,16 +54,13 @@ inline constexpr auto is_in_interval = detail::is_in_interval_type<interval_firs
...
  *
- * ### Example
- * \include test/snippet/utility/char_operations/char_predicate_is_in_alphabet.cpp
  * \hideinitializer
+ * \deprecated Use seqan3::char_is_valid_for instead.
  */
-template <alphabet alphabet_t>
-inline constexpr auto is_in_alphabet = detail::is_in_alphabet_type<alphabet_t>{};
+template <typename alphabet_t>
+SEQAN3_DEPRECATED_310 inline constexpr bool is_in_alphabet = true;
```
smehringer, line 63: Oh this is tricky. I see why you did this but this will break the API (and possibly external code!).

There is another solution that is less breaking :)

Revert the changes here except the deprecation warning.
Do not delete detail::is_in_alphabet_type but instead wrap the whole thing in 
```cpp
#ifdef SEQAN3_DEPRECATED_310
 // struct is_in_alphabet_type ...
#endif // SEQAN3_DEPRECATED_310
```

That way, it will remain there without triggering a deprecating warning but we will be able to spot it as "This needs to be deleted once 3.1 is released".

We did that for `seqan3/core/detail/iterator_traits.hpp` if you wanna take a look.

Also do not delete the tests just that but wrap them in 

```cpp

TEST(... non_deprecated)
// ....

#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wdeprecated-declarations"

TEST(deprecated_stuff)
// ...

#pragma GCC diagnostic pop
```

smehringer (changes requested):
I'm very sorry for the late review!

test/unit/utility/char_operations/char_predicate_test.cpp
```diff
@@ -10,13 +10,17 @@
 #include <string>
 
 #include <seqan3/alphabet/aminoacid/aa27.hpp>
+#include <seqan3/alphabet/concept.hpp>
```
smehringer, line 13: Is this also needed if we remove `is_in_alphabet`?

smehringer (commented):
LGTM! Thanks :) 

Just one question before I merge

test/unit/utility/char_operations/char_predicate_test.cpp
joergi-w: Yes, in the very last test we use `char_is_valid_for` which is defined in `seqan3/alphabet/concept.hpp`.

joergi-w: Thinking about it again, I could also remove the last test and the concept include, as we have plenty of `char_is_valid_for` tests already. It was introduced because we had an issue with the gapped alphabets not being recognised correctly by `is_in_alphabet`.

smehringer: No it's fine like this I think. One test more does not hurt :)

smehringer (approved):

PR #2323, opened 2021-01-08
[DOC] Change doc for add_enum_bitwise_operators

MitraDarja:
Resolves https://github.com/seqan/product_backlog/issues/260

### DIFF
--- a/include/seqan3/alignment/matrix/trace_directions.hpp
+++ b/include/seqan3/alignment/matrix/trace_directions.hpp
@@ -20,7 +20,9 @@ namespace seqan3::detail
 
 /*!\brief The possible directions a trace can have. The values can be combined by the logical `|`-operator.
  * \ingroup alignment_matrix
- * \sa seqan3::add_enum_bitwise_operators <seqan3::detail::trace_directions> enables combining enum values.
+ * \implements seqan3::enum_bitwise_operators
+ * \implements seqan3::enum_bitwise_operators
+  * \sa seqan3::enum_bitwise_operators enables combining enum values.
  * \sa seqan3::detail::alignment_trace_matrix implementations use this enum as matrix entry type.
  */
 enum struct trace_directions : uint8_t
@@ -43,14 +45,13 @@ enum struct trace_directions : uint8_t
 
 namespace seqan3
 {
+//!\cond DEV
 //!\brief Enable bitwise operators for enum seqan3::detail::trace_directions.
 //!\ingroup alignment_matrix
+//!\sa seqan3::enum_bitwise_operators enables combining enum values.
 template <>
 constexpr bool add_enum_bitwise_operators<seqan3::detail::trace_directions> = true;
-} // namespace seqan3
-
-namespace seqan3
-{
+//!\endcond
 
 /*!\brief All trace_directions can be printed as ascii or as utf8 to the seqan3::debug_stream.
  * \param s The seqan3::debug_stream.
--- a/include/seqan3/alignment/matrix/trace_directions.hpp
+++ b/include/seqan3/alignment/matrix/trace_directions.hpp
@@ -22,7 +22,7 @@ namespace seqan3::detail
  * \ingroup alignment_matrix
  * \implements seqan3::enum_bitwise_operators
  * \implements seqan3::enum_bitwise_operators
-  * \sa seqan3::enum_bitwise_operators enables combining enum values.
+ * \sa seqan3::enum_bitwise_operators enables combining enum values.
  * \sa seqan3::detail::alignment_trace_matrix implementations use this enum as matrix entry type.
  */
 enum struct trace_directions : uint8_t
--- a/include/seqan3/core/add_enum_bitwise_operators.hpp
+++ b/include/seqan3/core/add_enum_bitwise_operators.hpp
@@ -18,7 +18,64 @@
 
 namespace seqan3
 {
+/*!\interface seqan3::enum_bitwise_operators
+ * \brief You can expect these functions on all types that overload seqan3::add_enum_bitwise_operators.
+ */
+/*!\name Requirements for seqan3::enum_bitwise_operators
+ * \relates seqan3::enum_bitwise_operators
+ * \brief You can expect these member functions.
+ * \{
+ * \fn operator&(lhs_t lhs, rhs_t rhs)
+ * \brief Returns the binary `&` operator of lhs and rhs.
+ * \param lhs First enum.
+ * \param rhs Second enum.
+ *
+ * \returns the binary conjunction of `lhs` and `rhs`.
+ *
+ * \fn operator|(lhs_t lhs, rhs_t rhs)
+ * \brief Returns the binary `|` operator of lhs and rhs.
+ * \param lhs First enum.
+ * \param rhs Second enum.
+ *
+ * \returns the binary disjunction of `lhs` and `rhs`.
+ *
+ * \fn operator^(lhs_t lhs, rhs_t rhs)
+ * \brief Returns the binary `^` operator of lhs and rhs.
+ * \param lhs First enum.
+ * \param rhs Second enum.
+ *
+ * \returns the binary XOR operation on `lhs` and `rhs`.
+ *
+ * \fn operator~(lhs_t lhs)
+ * \brief Returns the binary `~` operator of lhs.
+ * \param lhs First enum.
+ *
+ * \returns the binary NOT operation on `lhs`.
+ *
+ * \fn operator&=(lhs_t & lhs, rhs_t rhs)
+ * \brief Returns the binary `&=` operator of lhs and rhs.
+ * \param lhs First enum.
+ * \param rhs Second enum.
+ *
+ * \returns the binary AND assigment on `lhs`.
+ *
+ * \fn operator|=(lhs_t & lhs, rhs_t rhs)
+ * \brief Returns the binary `|=` operator of lhs and rhs.
+ * \param lhs First enum.
+ * \param rhs Second enum.
+ *
+ * \returns the binary OR assignment on `lhs`.
+ *
+ * \fn operator^=(lhs_t & lhs, rhs_t rhs)
+ * \brief Returns the binary `^=` operator of lhs and rhs.
+ * \param lhs First enum.
+ * \param rhs Second enum.
+ *
+ * \returns the binary XOR assignment on `lhs`.
+ * \}
+ */
 
+//!\cond DEV
 /*!\brief Set to true for a scoped enum to have binary operators overloaded.
  * \ingroup core
  *
@@ -38,6 +95,9 @@ constexpr bool add_enum_bitwise_operators = false;
  * \brief Perform binary operations like on ints or weak enums. These overloads are available if
  * seqan3::add_enum_bitwise_operators is defined for your type.
  * \ingroup core
+ *
+ * \details
+ *
  * \see seqan3::add_enum_bitwise_operators
  * \{
  */
@@ -93,5 +153,6 @@ constexpr t & operator^= (t & lhs, t rhs) noexcept
     return lhs;
 }
 //!\}
+//!\endcond
 
 } // namespace seqan3
--- a/include/seqan3/core/detail/debug_stream_type.hpp
+++ b/include/seqan3/core/detail/debug_stream_type.hpp
@@ -36,9 +36,9 @@ enum fmtflags2
     default_            = small_int_as_number
 };
 
+//!\cond DEV
 //!\brief Overload bitwise operators for seqan3::fmtflags2.
 //!\sa seqan3::enum_bitwise_operators enables combining enum values.
-//!\cond DEV
 template <>
 constexpr bool add_enum_bitwise_operators<fmtflags2> = true;
 //!\endcond
--- a/include/seqan3/core/detail/debug_stream_type.hpp
+++ b/include/seqan3/core/detail/debug_stream_type.hpp
@@ -25,6 +25,8 @@ namespace seqan3
 
 //!\brief Flags that change the behaviour of the seqan3::debug_stream.
 //!\ingroup stream
+//!\implements seqan3::enum_bitwise_operators
+//!\sa seqan3::enum_bitwise_operators enables combining enum values.
 enum fmtflags2
 {
     none                = 0,        //!< No flag is set.
@@ -35,8 +37,11 @@ enum fmtflags2
 };
 
 //!\brief Overload bitwise operators for seqan3::fmtflags2.
+//!\sa seqan3::enum_bitwise_operators enables combining enum values.
+//!\cond DEV
 template <>
 constexpr bool add_enum_bitwise_operators<fmtflags2> = true;
+//!\endcond
 
 // ------------------------------------------------------------------
 // seqan3::debug_stream_type
--- a/include/seqan3/core/detail/strong_type.hpp
+++ b/include/seqan3/core/detail/strong_type.hpp
@@ -27,6 +27,8 @@ namespace seqan3::detail
 
 /*!\brief Enum class for all supported operations that can be added to a seqan3::detail::strong_type.
  * \ingroup core
+ * \implements seqan3::enum_bitwise_operators
+ * \sa seqan3::enum_bitwise_operators enables combining enum values.
  */
 enum struct strong_type_skill
 {
@@ -59,7 +61,8 @@ enum struct strong_type_skill
 
 namespace seqan3
 {
-//!\cond
+//!\cond DEV
+//!\sa seqan3::enum_bitwise_operators enables combining enum values.
 template <>
 constexpr bool add_enum_bitwise_operators<seqan3::detail::strong_type_skill> = true;
 //!\endcond
--- a/include/seqan3/core/detail/strong_type.hpp
+++ b/include/seqan3/core/detail/strong_type.hpp
@@ -62,6 +62,7 @@ enum struct strong_type_skill
 namespace seqan3
 {
 //!\cond DEV
+//!\brief Overload bitwise operators for seqan3::detail::strong_type_skill.
 //!\sa seqan3::enum_bitwise_operators enables combining enum values.
 template <>
 constexpr bool add_enum_bitwise_operators<seqan3::detail::strong_type_skill> = true;
--- a/include/seqan3/io/alignment_file/misc.hpp
+++ b/include/seqan3/io/alignment_file/misc.hpp
@@ -24,6 +24,8 @@ struct ref_info_not_given
 
 /*!\brief An enum flag that describes the properties of an aligned read (given as a SAM record).
  * \ingroup alignment_file
+ * \implements seqan3::enum_bitwise_operators
+ * \sa seqan3::enum_bitwise_operators enables combining enum values.
  *
  * The SAM flag are bitwise flags, which means that each value corresponds to a specific bit that is set and that they
  * can be combined and tested using binary operations.
@@ -84,9 +86,12 @@ enum class sam_flag : uint16_t
    supplementary_alignment = 0x800  //!< This sequence is part of a split alignment and is not the primary alignment.
 };
 
+//!\cond DEV
 //!\brief Enables bitwise operations for seqan3::sam_flags.
+//!\sa seqan3::enum_bitwise_operators enables combining enum values.
 template <>
 constexpr bool add_enum_bitwise_operators<sam_flag> = true;
+//!\endcond
 
 /*!\brief Overload for the seqan3::sam_flags.
  * \tparam char_t Type char type of the debug_stream.
--- a/include/seqan3/range/views/translate.hpp
+++ b/include/seqan3/range/views/translate.hpp
@@ -61,6 +61,8 @@ namespace seqan3
 {
 
 //!\brief Specialisation values for single and multiple translation frames.
+//!\implements seqan3::enum_bitwise_operators
+//!\sa seqan3::enum_bitwise_operators enables combining enum values.
 enum class translation_frames : uint8_t
 {
     FWD_FRAME_0 = 1,                                    //!< The first forward frame starting at position 0
@@ -77,9 +79,12 @@ enum class translation_frames : uint8_t
     SIX_FRAME = FWD | REV                               //!< All frames
 };
 
+//!\cond DEV
 //!\brief Enable bitwise operators for enum translation_frames.
+//!\sa seqan3::enum_bitwise_operators enables combining enum values.
 template <>
 constexpr bool add_enum_bitwise_operators<translation_frames> = true;
+//!\endcond
 
 }
 


eaasna (commented):
Thank you for the PR! Could you please specify which part of [seqan/product_backlog#260](https://github.com/seqan/product_backlog/issues/260) still needs to be done after this?

MitraDarja:
@eaasna I think nothing, but I wasn't sure if I understood the core decision correctly, so I wanted to ask Rene or Enrico about it. :)

include/seqan3/core/add_enum_bitwise_operators.hpp
```diff
@@ -34,7 +34,8 @@ namespace seqan3
 template <typename t>
 constexpr bool add_enum_bitwise_operators = false;
 
-/*!\name Binary operators for scoped enums
+//!\cond DEV
+/*!\interface enum_bitwise_operators
```
eseiler, line 38: It could be that it works like this, but here is what I thought:

The `\interface`  should not be here, it should be more like in, e.g., the fm_index_concept https://github.com/seqan/seqan3/blob/971b3e41524892431fa686981c52975709e0b7f4/include/seqan3/search/fm_index/concept.hpp#L144

You do the `\interface` and then you can document stuff that belongs to the interface, you can also use `\fn` to write function signatures without actually defining cpp functions: https://www.doxygen.nl/manual/commands.html#cmdfn
I think we use `\fn` somewhere :)

Also, the interface documentation should then be more verbose, e.g. what does `&` mean?

MitraDarja: I am not sure, I know what you mean. 
So, I should move the `\interface` to line 22?
And then explain what every operator does? Should this then be part of DEV too or not?
Plus, does it really make sense to explain every operator, I would assume the reference to the typical binary operators is enough.

`seqan3::scoring_scheme_for` uses `\fn`, I assume you want every operator explained with an `\fn`?

eseiler: You need to write something like the concepts:
```
/!*\interface seqan3::bitwise_enum_operators

  \brief blablaba
  Does ....
  \fn operator&
*/
```

One sentence for each operator is enough, but all need to be documented since we want to tag the other classes with this interface.
The definition of the operators goes into DEV (just like you already did, but the interface is public).

include/seqan3/core/add_enum_bitwise_operators.hpp
```diff
@@ -19,7 +19,8 @@
 namespace seqan3
 {
 
-/*!\brief Set to true for a scoped enum to have binary operators overloaded.
+/*!\interface enum_bitwise_operators
+ * \brief Set to true for a scoped enum to have binary operators overloaded.
```
eseiler, line 23:
```suggestion
/*!\brief Set to true for a scoped enum to have binary operators overloaded.
```

eseiler:
I dismissed my review because there is no first review yet.

But please check whether you need both `\sa` and `\implements` to properly crosslink.

MitraDarja:
> I dismissed my review because there is no first review yet.
> 
> But please check whether you need both `\sa` and `\implements` to properly crosslink.

It also works without `\sa`, but I think it looks better with `\sa`, because it is better formatted. :)

include/seqan3/alignment/matrix/trace_directions.hpp
```diff
@@ -45,6 +44,8 @@ namespace seqan3
 {
 //!\brief Enable bitwise operators for enum seqan3::detail::trace_directions.
 //!\ingroup alignment_matrix
+//!\implements enum_bitwise_operators
+//!\sa seqan3::add_enum_bitwise_operators <seqan3::detail::trace_directions> enables combining enum values.
```
joergi-w, line 48: I think that we do not use space before template parentheses:
```suggestion
//!\sa seqan3::add_enum_bitwise_operators<seqan3::detail::trace_directions> enables combining enum values.
```

include/seqan3/alignment/matrix/trace_directions.hpp
```diff
@@ -45,6 +44,8 @@ namespace seqan3
...
 template <>
 constexpr bool add_enum_bitwise_operators<seqan3::detail::trace_directions> = true;
 } // namespace seqan3
```
joergi-w, line 51: Here, the seqan3 namespace is closed and immediately re-opened again. Let's merge them?

include/seqan3/core/add_enum_bitwise_operators.hpp
```diff
@@ -18,6 +18,78 @@
 
 namespace seqan3
 {
+/*!\interface seqan3::enum_bitwise_operators
+ * \brief You can expect these functions on all types that implement seqan3::add_enum_bitwise_operators.
+ */
+/*!\name Requirements for seqan3::enum_bitwise_operators
```
joergi-w, line 24: You can leave the comment open, however it's not wrong what you did... just pointing it out.
```suggestion
 *
 * \name Requirements for seqan3::enum_bitwise_operators
```

include/seqan3/core/add_enum_bitwise_operators.hpp
```diff
@@ -18,6 +18,78 @@
...
+ * \relates seqan3::enum_bitwise_operators
+ * \brief You can expect these member functions.
+ * \{
+ * \fn operator&(lhs_t lhs, rhs_t rhs)
+ * \brief Returns the binary `&` operator of lhs and rhs.
+ * \param lhs First enum.
+ * \param rhs Second enum.
+ *
```
joergi-w, line 32: Do we need something like
```suggestion
 * \returns the conjunction of `lhs` and `rhs`.
```
? If yes, then please add it for each operator below as well.

include/seqan3/core/add_enum_bitwise_operators.hpp
```diff
@@ -18,6 +18,78 @@
...
+ * will provide an implementation).
+ *
+ * \fn operator|(lhs_t lhs, rhs_t rhs)
+ * \brief Returns the binary `|` operator of lhs and rhs.
+ * \param lhs First enum.
+ * \param rhs Second enum.
+ *
+ * \details
+ * \attention This is a concept requirement, not an actual function (however types satisfying this concept
+ * will provide an implementation).
```
joergi-w, line 44: This is essentially copied for each function documentation here and it therefore seems crowded. Consider whether it appears better in the heading, i.e. move it below the \name tag (somewhere after line 24).

include/seqan3/core/detail/debug_stream_type.hpp
```diff
@@ -35,6 +35,8 @@ enum fmtflags2
 };
 
 //!\brief Overload bitwise operators for seqan3::fmtflags2.
+//!\implements enum_bitwise_operators
+//!\sa seqan3::add_enum_bitwise_operators <seqan3::fmtflags2> enables combining enum values.
```
joergi-w, line 39:
```suggestion
//!\sa seqan3::add_enum_bitwise_operators<seqan3::fmtflags2> enables combining enum values.
```

include/seqan3/core/detail/strong_type.hpp
```diff
@@ -59,6 +59,8 @@ enum struct strong_type_skill
 
 namespace seqan3
 {
+//!\implements enum_bitwise_operators
+//!\sa seqan3::add_enum_bitwise_operators <seqan3::detail::trace_directions> enables combining enum values.
```
joergi-w, line 63:
```suggestion
//!\sa seqan3::add_enum_bitwise_operators<seqan3::detail::trace_directions> enables combining enum values.
```

include/seqan3/io/alignment_file/misc.hpp
```diff
@@ -85,6 +85,8 @@ enum class sam_flag : uint16_t
 };
 
 //!\brief Enables bitwise operations for seqan3::sam_flags.
+//!\implements enum_bitwise_operators
+//!\sa seqan3::add_enum_bitwise_operators <seqan3::sam_flag> enables combining enum values.
```
joergi-w, line 89:
```suggestion
//!\sa seqan3::add_enum_bitwise_operators<seqan3::sam_flag> enables combining enum values.
```

include/seqan3/range/views/translate.hpp
```diff
@@ -78,6 +78,8 @@ enum class translation_frames : uint8_t
 };
 
 //!\brief Enable bitwise operators for enum translation_frames.
+//!\implements enum_bitwise_operators
+//!\sa seqan3::add_enum_bitwise_operators <seqan3::translation_frames> enables combining enum values.
```
joergi-w, line 82:
```suggestion
//!\sa seqan3::add_enum_bitwise_operators<seqan3::translation_frames> enables combining enum values.
```

joergi-w (changes requested):

include/seqan3/core/add_enum_bitwise_operators.hpp
MitraDarja: I got some warning messages then, don't know why, but that is why I used two comments. :)

include/seqan3/core/add_enum_bitwise_operators.hpp
MitraDarja: I think it is better to have it for every function, in case a user is just looking for the specifics of one operator the user might miss an attention that is somewhere above.

include/seqan3/core/add_enum_bitwise_operators.hpp
eseiler: The `interface` entity gets generated when the comment block is closed, or soemthing like this.
`interface` and `name` are two things and you need to close the block inbetween

joergi-w: Thanks, then this is resolved :)

include/seqan3/core/add_enum_bitwise_operators.hpp
joergi-w: Sure, fine for me!

include/seqan3/core/add_enum_bitwise_operators.hpp
```diff
@@ -18,6 +18,92 @@
...
+ *
+ *
+ * \fn operator^(lhs_t lhs, rhs_t rhs)
+ * \brief Returns the binary `^` operator of lhs and rhs.
+ * \param lhs First enum.
+ * \param rhs Second enum.
+ *
+ * \returns the binary `^` operator of `lhs` and `rhs`.
```
joergi-w, line 56:
```suggestion
 * \returns the binary XOR operaton on `lhs` and `rhs`.
```

include/seqan3/core/add_enum_bitwise_operators.hpp
```diff
@@ -18,6 +18,92 @@
...
+ *
+ * \details
+ * \attention This is a concept requirement, not an actual function (however types satisfying this concept
+ * will provide an implementation).
+ *
+ * \fn operator~(lhs_t lhs)
+ * \brief Returns the binary `~` operator of lhs.
+ * \param lhs First enum.
+ *
+ * \returns the binary `~` operator of `lhs`.
```
joergi-w, line 66:
```suggestion
 * \returns the binary NOT operaton on `lhs`.
```

joergi-w (changes requested):
There are still two unresolved issues and I would use the names of the operators to make it more comprehensible:

joergi-w (approved):
Thank you!

include/seqan3/alignment/matrix/trace_directions.hpp
```diff
@@ -45,12 +44,10 @@ namespace seqan3
 {
 //!\brief Enable bitwise operators for enum seqan3::detail::trace_directions.
 //!\ingroup alignment_matrix
+//!\implements enum_bitwise_operators
+//!\sa seqan3::add_enum_bitwise_operators<seqan3::detail::trace_directions> enables combining enum values.
```
eseiler, line 48: I have a feeling that this might be a better fit for the enum itself since we augment the enum with some operators. Can you double-check the look and feel?

include/seqan3/core/add_enum_bitwise_operators.hpp
```diff
@@ -18,6 +18,92 @@
...
+ * \fn operator&(lhs_t lhs, rhs_t rhs)
+ * \brief Returns the binary `&` operator of lhs and rhs.
+ * \param lhs First enum.
+ * \param rhs Second enum.
+ *
+ * \returns the binary conjunction of `lhs` and `rhs`.
+ *
+ * \details
+ * \attention This is a concept requirement, not an actual function (however types satisfying this concept
+ * will provide an implementation).
```
eseiler, line 37: I would remove this (for all the operators) since it is not a concept

include/seqan3/core/detail/strong_type.hpp
```diff
@@ -59,6 +59,8 @@ enum struct strong_type_skill
 
 namespace seqan3
 {
+//!\implements enum_bitwise_operators
+//!\sa seqan3::add_enum_bitwise_operators<seqan3::detail::trace_directions> enables combining enum values.
 //!\cond
 template <>
 constexpr bool add_enum_bitwise_operators<seqan3::detail::strong_type_skill> = true;
```
eseiler, line 66: Since there is `\cond` this documentation should have no effect.

eseiler (changes requested):
Can you try out if it works if you move the `\implements` and `\sa` to the actual enum and wrap the `add_enum_bitwise_operators` in a `\cond` (maybe `\cond DEV`?) `\endcond`. I don't necessarily see the need to document `add_enum_bitwise_operators` when all it does is augment the enum and instead just document it in the enum.

MitraDarja:
> Can you try out if it works if you move the `\implements` and `\sa` to the actual enum and wrap the `add_enum_bitwise_operators` in a `\cond` (maybe `\cond DEV`?) `\endcond`. I don't necessarily see the need to document `add_enum_bitwise_operators` when all it does is augment the enum and instead just document it in the enum.

If we do this (for both versions of `\cond` with or without DEV) the `add_enum_bitwise_operators< X >` is not listed extra and can only be found in the list Variables. Is that what we want?

include/seqan3/core/detail/strong_type.hpp
MitraDarja: That is true, so should I remove the `\cond`?

include/seqan3/alignment/matrix/trace_directions.hpp
MitraDarja: As described below, than it only appears in the Variable list.

eseiler:
> If we do this (for both versions of `\cond` with or without DEV) the `add_enum_bitwise_operators< X >` is not listed extra and can only be found in the list Variables. Is that what we want?

IMO it should not show up in the user documentation since the `add_enum_bitwise_operators` is not API. It is ok if it shows up in the developer documentation.

MitraDarja:
> > If we do this (for both versions of `\cond` with or without DEV) the `add_enum_bitwise_operators< X >` is not listed extra and can only be found in the list Variables. Is that what we want?
> 
> IMO it should not show up in the user documentation since the `add_enum_bitwise_operators` is not API. It is ok if it shows up in the developer documentation.

@eseiler But where in the dev documentation, do you want it to appear? Because if we move implements and sa to the actual enum, then it appears only in the Variable list not as an extra function.

include/seqan3/alignment/matrix/trace_directions.hpp
```diff
@@ -20,7 +20,6 @@ namespace seqan3::detail
 
 /*!\brief The possible directions a trace can have. The values can be combined by the logical `|`-operator.
  * \ingroup alignment_matrix
```
eseiler, line 22:
```suggestion
 * \ingroup alignment_matrix
 * \implements seqan3::enum_bitwise_operators
 * \sa seqan3::enum_bitwise_operators enables combining enum values.
```

`\implements` and `\sa` for all enums

include/seqan3/alignment/matrix/trace_directions.hpp
eseiler, line 48:
```suggestion
//!\sa seqan3::add_enum_bitwise_operators enables combining enum values.
```

include/seqan3/alignment/matrix/trace_directions.hpp
```diff
@@ -45,12 +44,10 @@ namespace seqan3
 {
 //!\brief Enable bitwise operators for enum seqan3::detail::trace_directions.
 //!\ingroup alignment_matrix
+//!\implements enum_bitwise_operators
+//!\sa seqan3::add_enum_bitwise_operators<seqan3::detail::trace_directions> enables combining enum values.
 template <>
 constexpr bool add_enum_bitwise_operators<seqan3::detail::trace_directions> = true;
```
eseiler, line 50: `add_enum_bitwise_operators` should be wrapped in `\cond DEV`

include/seqan3/core/add_enum_bitwise_operators.hpp
```diff
@@ -18,6 +18,62 @@
 
 namespace seqan3
 {
+/*!\interface seqan3::enum_bitwise_operators
+ * \brief You can expect these functions on all types that implement seqan3::add_enum_bitwise_operators.
```
eseiler, line 22:
```suggestion
 * \brief You can expect these functions on all types that overload seqan3::add_enum_bitwise_operators.
```

include/seqan3/core/add_enum_bitwise_operators.hpp
```diff
@@ -18,6 +18,62 @@
...
+ * \param lhs First enum.
+ *
+ * \returns the binary NOT operation on `lhs`.
+ *
+ * \fn operator&=(lhs_t & lhs, rhs_t rhs)
+ * \brief Returns the binary `&=` operator of lhs and rhs.
+ * \param lhs First enum.
+ * \param rhs Second enum.
+ *
+ * \returns lhs.
```
eseiler, line 60:
```suggestion
 * \returns the binary AND assigment on `lhs`.
```

include/seqan3/core/add_enum_bitwise_operators.hpp
```diff
@@ -18,6 +18,62 @@
...
+ *
+ * \fn operator|=(lhs_t & lhs, rhs_t rhs)
+ * \brief Returns the binary `|=` operator of lhs and rhs.
+ * \param lhs First enum.
+ * \param rhs Second enum.
+ *
+ * \returns lhs.
```
eseiler, line 67:
```suggestion
 * \returns the binary OR assignment on `lhs`.
```

include/seqan3/core/add_enum_bitwise_operators.hpp
```diff
@@ -18,6 +18,62 @@
...
+ *
+ * \fn operator^=(lhs_t & lhs, rhs_t rhs)
+ * \brief Returns the binary `^=` operator of lhs and rhs.
+ * \param lhs First enum.
+ * \param rhs Second enum.
+ *
+ * \returns lhs.
```
eseiler, line 74:
```suggestion
 * \returns the binary XOR assignment on `lhs`.
```

include/seqan3/core/add_enum_bitwise_operators.hpp
```diff
@@ -34,10 +90,14 @@ namespace seqan3
 template <typename t>
 constexpr bool add_enum_bitwise_operators = false;
 
+//!\cond DEV
```
eseiler, line 93: Move this cond to line 77 (such that add_enum_bitwise_operators  is also DEV only)

include/seqan3/core/detail/debug_stream_type.hpp
```diff
@@ -35,6 +35,8 @@ enum fmtflags2
 };
 
 //!\brief Overload bitwise operators for seqan3::fmtflags2.
+//!\implements enum_bitwise_operators
+//!\sa seqan3::add_enum_bitwise_operators<seqan3::fmtflags2> enables combining enum values.
 template <>
 constexpr bool add_enum_bitwise_operators<fmtflags2> = true;
```
eseiler, line 41:
```suggestion
//!\cond DEV
//!\brief Overload bitwise operators for seqan3::fmtflags2.
//!\implements enum_bitwise_operators
//!\sa seqan3::add_enum_bitwise_operators enables combining enum values.
template <>
constexpr bool add_enum_bitwise_operators<fmtflags2> = true;
//!cond DEV
```

eseiler (changes requested):
The enum should have both `\implements` and `\sa` (see example). The add_enum_bitwise_operators overload should only have `\sa` and wrapped in `\cond DEV` (also see example).

eseiler:
And you may need a rebase? At least I needed to rebase

MitraDarja:
> And you may need a rebase? At least I needed to rebase

Yes, I needed one. I squashed all commits with the rebase already, I hope that was ok.

include/seqan3/alignment/matrix/trace_directions.hpp
```diff
@@ -20,7 +20,9 @@ namespace seqan3::detail
 
 /*!\brief The possible directions a trace can have. The values can be combined by the logical `|`-operator.
  * \ingroup alignment_matrix
- * \sa seqan3::add_enum_bitwise_operators <seqan3::detail::trace_directions> enables combining enum values.
+ * \implements seqan3::enum_bitwise_operators
+ * \implements seqan3::enum_bitwise_operators
+  * \sa seqan3::enum_bitwise_operators enables combining enum values.
```
eseiler, line 25:
```suggestion
 * \implements seqan3::enum_bitwise_operators
 * \sa seqan3::enum_bitwise_operators enables combining enum values.
```

include/seqan3/core/detail/debug_stream_type.hpp
```diff
@@ -35,8 +37,11 @@ enum fmtflags2
 };
 
 //!\brief Overload bitwise operators for seqan3::fmtflags2.
+//!\sa seqan3::enum_bitwise_operators enables combining enum values.
+//!\cond DEV
```
eseiler, line 41:
```suggestion
//!\cond DEV
//!\brief Overload bitwise operators for seqan3::fmtflags2.
//!\sa seqan3::enum_bitwise_operators enables combining enum values.
```

include/seqan3/core/detail/strong_type.hpp
```diff
@@ -59,7 +61,8 @@ enum struct strong_type_skill
 
 namespace seqan3
 {
-//!\cond
+//!\cond DEV
+//!\sa seqan3::enum_bitwise_operators enables combining enum values.
```
eseiler, line 65:
```suggestion
//!\cond DEV
//!\brief Overload bitwise operators for seqan3::detail::strong_type_skill.
//!\sa seqan3::enum_bitwise_operators enables combining enum values.
```

eseiler (changes requested):
💅

eseiler (approved):

PR #2325, opened 2021-01-08
[FIX] Make class interleaved_bloom_filter::membership_agent public.

### LINKED ISSUE #2324: [IBF] membership_agent class of the interleaved bloom filter is private.
## Description

It is currently not possible to access the `membership_agent` type of the `seqan3::interleaved_bloom_filter`. 
I want to do this:

```cpp
void func(typename seqan3::interleaved_bloom_filter<>::membership_agend & agent)
{
    // do something
}
```

But the type is private.
It is used in public contexts though:

```
seqan3::interleaved_bloom_filter<> ibf{...};
auto & agent = ibf.membership_agent();
```

So the type should be also public.

smehringer:
The documentation can be left untouched since it was already public before.

fixes #2324

### DIFF
--- a/include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
+++ b/include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
@@ -129,8 +129,6 @@ private:
     //!\cond
     template <data_layout data_layout_mode>
     friend class interleaved_bloom_filter;
-
-    class membership_agent;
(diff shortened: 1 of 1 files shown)


SGSSGene (approved):
It was private, now it is public! Very elegant solution :-)

eseiler (approved):

PR #2319, opened 2021-01-05
Argument parser some refactoring

smehringer:

PR #2326, opened 2021-01-08
Factor out bold formatting of help page export into separate function.

smehringer:
Feel free to comment on the name of the function `in_bold`. I didn't think a lot about the naming.

See #2319 for every commit that is planned for some refactoring. It will make it easier to keep the help page export features in synch.

### DIFF
--- a/include/seqan3/argument_parser/detail/format_help.hpp
+++ b/include/seqan3/argument_parser/detail/format_help.hpp
@@ -138,9 +138,9 @@ protected:
     void print_subsection(std::string const & title)
     {
         std::ostream_iterator<char> out(std::cout);
-        std::cout << '\n' << to_text("\\fB");
+        std::cout << '\n';
         std::fill_n(out, layout.leftPadding / 2, ' ');
-        std::cout << title << to_text("\\fP") << '\n';
+        std::cout << in_bold(title) << '\n';
         prev_was_paragraph = false;
     }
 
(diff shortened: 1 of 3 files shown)


include/seqan3/argument_parser/detail/format_help.hpp
```diff
@@ -418,6 +415,15 @@ class format_help : public format_help_base<format_help>
             std::cout << '\n';
     }
 
+    /*!\brief Format string as in_bold.
```
MitraDarja, line 418: Shouldn't it be "Format string in bold"?

include/seqan3/argument_parser/detail/format_html.hpp
```diff
@@ -182,18 +182,16 @@ class format_html : public format_help_base<format_html>
         // Print legal stuff
         if ((!meta.short_copyright.empty()) || (!meta.long_copyright.empty()) || (!meta.citation.empty()))
         {
-            std::cout << "<h2>Legal</h2>\n<strong>";
+            std::cout << "<h2>Legal</h2>\n";
```
MitraDarja, line 185: Why are you deleting the strong here without replacing it with a in_bold?

MitraDarja (changes requested):
Looks good to me, but two minor things.
I am fine with the naming. 😄

include/seqan3/argument_parser/detail/format_html.hpp
smehringer: It is replaced but in the line below. 
Previously the person who wrote/copied this (probably me :see_no_evil:) tried to save some ... strings I guess.
E.g. 
```
std::cout << "foo<strong>"
          << var << "</strong>"; // var is bold
```
instead of 
```
std::cout << "foo"
          << "<strong>" << var << "</strong>"; // var is bold
```

include/seqan3/argument_parser/detail/format_help.hpp
smehringer: yes

smehringer:
```suggestion
    /*!\brief Format string in bold.
```

MitraDarja (approved):
Thanks for the clarification. 😄

marehr (approved):

PR #2330, opened 2021-01-11
Configuration/better config commit 5 7

rrahn:
This Pr reduces quite a lot of complexity within the configuration object. Most of it by reducing the different combinations of the pipe operator, which is now handled through one function using an additional append member function.

### DIFF
--- a/include/seqan3/core/configuration/configuration.hpp
+++ b/include/seqan3/core/configuration/configuration.hpp
@@ -24,40 +24,6 @@
 namespace seqan3
 {
 
-//!\cond
-// Forward declaration for friend declaration definitions below.
-template <detail::config_element ... configs_t>
-class configuration;
-
-template <typename lhs_derived_t, typename lhs_value_t, typename rhs_derived_t, typename rhs_value_t>
-constexpr auto operator|(pipeable_config_element<lhs_derived_t, lhs_value_t> && lhs,
-                         pipeable_config_element<rhs_derived_t, rhs_value_t> && rhs)
-{
-    return configuration{static_cast<lhs_derived_t &&>(lhs)}.push_back(static_cast<rhs_derived_t &&>(rhs));
-}
-
-template <typename lhs_derived_t, typename lhs_value_t, typename rhs_derived_t, typename rhs_value_t>
-constexpr auto operator|(pipeable_config_element<lhs_derived_t, lhs_value_t> && lhs,
-                         pipeable_config_element<rhs_derived_t, rhs_value_t> const & rhs)
-{
-    return configuration{static_cast<lhs_derived_t &&>(lhs)}.push_back(static_cast<rhs_derived_t const &>(rhs));
-}
-
-template <typename lhs_derived_t, typename lhs_value_t, typename rhs_derived_t, typename rhs_value_t>
-constexpr auto operator|(pipeable_config_element<lhs_derived_t, lhs_value_t> const & lhs,
-                         pipeable_config_element<rhs_derived_t, rhs_value_t> && rhs)
-{
-    return configuration{static_cast<lhs_derived_t const &>(lhs)}.push_back(static_cast<rhs_derived_t &&>(rhs));
-}
-
-template <typename lhs_derived_t, typename lhs_value_t, typename rhs_derived_t, typename rhs_value_t>
-constexpr auto operator|(pipeable_config_element<lhs_derived_t, lhs_value_t> const & lhs,
-                         pipeable_config_element<rhs_derived_t, rhs_value_t> const & rhs)
-{
-    return configuration{static_cast<lhs_derived_t const &>(lhs)}.push_back(static_cast<rhs_derived_t const &>(rhs));
-}
-//!\endcond
-
 // ----------------------------------------------------------------------------
 // configuration
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


include/seqan3/core/configuration/configuration.hpp
```diff
@@ -227,167 +245,6 @@ class configuration : public std::tuple<configs_t...>
...
-     * \tparam lhs_derived_t The derived type of the left hand side operand.
-     * \tparam lhs_value_t   The value type of the left hand side operand.
-     * \tparam rhs_derived_t The derived type of the right hand side operand.
-     * \tparam rhs_value_t   The value type of the right hand side operand.
-     * \param[in] lhs        The left hand operand.
-     * \param[in] rhs        The right hand operand.
-     * \returns A new seqan3::configuration containing `lhs` and `rhs`.
-     */
-    template <typename lhs_derived_t, typename lhs_value_t, typename rhs_derived_t, typename rhs_value_t>
-    friend constexpr auto operator|(pipeable_config_element<lhs_derived_t, lhs_value_t> && lhs,
```
eaasna, line 243: I noticed that all of these friend declarations have been removed. Could you please explain why the new implementation of combining configuration objects doesn't use friend declarations?

include/seqan3/core/configuration/configuration.hpp
```diff
@@ -571,6 +351,37 @@ template <detail::config_element config_t>
 configuration(config_t) -> configuration<config_t>;
 //!\}
 
+/*!\brief Combines configuration and configuration elements to create a new configuration object.
```
eaasna, line 354: Just a small detail. Are "configuration" and "configuration object" used interchangeably to mean a seqan3::configuration object? Found a terminology here [/seqan/product_backlog/issues/131](https://github.com/seqan/product_backlog/issues/131) that doesn't mention this

eaasna (commented):
Hi @rrahn thank you for the PR!

include/seqan3/core/configuration/configuration.hpp
rrahn: When this class and the pipe-operator was initially implemented, we used a base class to distinguish a configuration object from any other object. This was done via a base class, which allows us to define the operator interface via a reference to the base class and only classes derived from this base class can be used. However, we needed the actual type of the derived class and thus used a CRTP design patterh to get from the base class back to the derived class. Accordingly, we needed all combinations of the |-operator, which were quite a lot.
Instead, we now just use a concept that checks if a type is a base of `configuration_element_base` similar to how views are enabled by inheriting from a view_base class. These classes are empty and do nothing except providing a strong type that allows to distinguish certain types from other objects. Using a concept, however, we could reduce all but one definition of a |-operator. This operator is now a free-function because it uses the public append interface of the configuration class and does not need access to the private internals of the configuration class. That's why it is not a friend function anymore. This makes sense, since the configuration class can also be modified via member functions (which came in later btw). The pipe-operator is simply a neat trick to make the user interface a little more user-friendly.
So both things are now possible:

```cpp
configuration config1 = configuration{align_cfg::method_global{}}.append(align_cfg::band_fixed_size{...}).append(align_cfg::output_score{});
configuration config2 = align_cfg::method_global{} | align_cfg::band_fixed_size{...} | align_cfg::output_score{};
```

include/seqan3/core/configuration/configuration.hpp
rrahn: I see your point here. So yes, it does but it could be clearer here I guess. I am trying to update the PR today. Thank you.

rrahn: Ok, I changed only a little bit the documentation. Do you think it is clear now?

include/seqan3/core/configuration/configuration.hpp
eaasna: Thank you! I appreciate the explanation

eaasna (approved):

include/seqan3/core/configuration/configuration.hpp
eaasna: Where is this change visible?

include/seqan3/core/configuration/configuration.hpp
rrahn: You are welcome.

include/seqan3/core/configuration/configuration.hpp
rrahn: Oh, there was an error during the push so it wasn't published yet. Now the rebased version should contain the change.

eaasna: Ah yes, that clears it. Thanks!

include/seqan3/core/configuration/configuration.hpp
```diff
@@ -196,6 +162,58 @@ class configuration : public std::tuple<configs_t...>
...
+        {
+            return configuration<configs_t..., std::remove_cvref_t<other_configuration_t>>
+            {
+                std::tuple_cat(static_cast<base_type>(*this),
+                               std::tuple{std::forward<other_configuration_t>(other_config)})
+            };
+        }
+        else
+        {
+            using other_base_t = typename std::remove_cvref_t<other_configuration_t>::base_type;
```
marehr, line 197:
```suggestion
            // The case that other_configuration_t is a seqan3::configuration
            using other_base_t = typename std::remove_cvref_t<other_configuration_t>::base_type;
```

include/seqan3/core/configuration/configuration.hpp
```diff
@@ -196,6 +162,58 @@ class configuration : public std::tuple<configs_t...>
...
+            using other_base_maybe_const_t =
+                std::conditional_t<std::is_const_v<std::remove_reference_t<other_configuration_t>>,
+                                   std::add_const_t<other_base_t>,
+                                   other_base_t>;
+            using fwd_other_base_t = std::conditional_t<std::is_lvalue_reference_v<other_configuration_t>,
+                                                        std::add_lvalue_reference_t<other_base_maybe_const_t>,
+                                                        std::add_rvalue_reference_t<other_base_maybe_const_t>>;
```
marehr, line 204: This took me a while.

Maybe add a comment "We transfer the type qualifier, like &, &&, const, and any combination of them, from other_configuration_t to ::base_type."

include/seqan3/core/configuration/configuration.hpp
```diff
@@ -196,6 +162,58 @@ class configuration : public std::tuple<configs_t...>
...
+
+            using other_configs_list_t = detail::transfer_template_args_onto_t<other_base_t, type_list>;
```
marehr, line 206:
```suggestion
            // add config_elements from other_base_t to the config_elements of this class
            using other_configs_list_t = detail::transfer_template_args_onto_t<other_base_t, type_list>;
```

include/seqan3/core/configuration/configuration.hpp
```diff
@@ -227,167 +245,6 @@ class configuration : public std::tuple<configs_t...>
...
-                                    configuration<rhs_configs_t...> const & rhs)
-    {
-        using lhs_base_t = typename configuration::base_type;
-        using rhs_base_t = typename configuration<rhs_configs_t...>::base_type;
-
-        return make_configuration(std::tuple_cat(static_cast<lhs_base_t>(lhs),
-                                                 static_cast<rhs_base_t>(rhs)));
-    }
-    //!\}
-
```
marehr, line 390: Oh!! Nice!

test/unit/core/algorithm/pipeable_config_element_test.cpp
```diff
@@ -79,25 +79,25 @@ TEST(pipeable_config_element, element_with_configuration)
     // lvalue | lvalue
     {
         auto cfg = b2 | tmp;
-        EXPECT_TRUE((std::is_same_v<decltype(cfg), seqan3::configuration<bar, bax>>));
+        EXPECT_TRUE((std::is_same_v<decltype(cfg), seqan3::configuration<bax, bar>>));
```
marehr, line 82: Is there a reason why you changes the order?

test/unit/core/configuration/configuration_test.cpp
```diff
@@ -171,6 +171,72 @@ TEST(configuration, exists_by_type_template)
     EXPECT_FALSE(decltype(cfg)::exists<foo>());
 }
 
+TEST(configuration, append_configuration_element)
+{
+    {
+        seqan3::configuration<foo, bar> cfg{};
+        auto new_cfg = cfg.append(bax{});
```
marehr, line 178: Just for my self: rvalue

test/unit/core/configuration/configuration_test.cpp
```diff
@@ -171,6 +171,72 @@ TEST(configuration, exists_by_type_template)
...
+        EXPECT_TRUE((std::same_as<decltype(new_cfg), seqan3::configuration<foo, bar, bax>>));
+    }
+
+    {
+        seqan3::configuration<foo, bar> cfg{};
+        bax b{};
+        auto new_cfg = cfg.append(b);
```
marehr, line 185: Just for my self: lvalue

test/unit/core/configuration/configuration_test.cpp
```diff
@@ -171,6 +171,72 @@ TEST(configuration, exists_by_type_template)
...
+        EXPECT_TRUE((std::same_as<decltype(new_cfg), seqan3::configuration<foo, bar, bax>>));
+    }
+
+    {
+        seqan3::configuration<foo, bar> cfg{};
+        bax b{};
+        auto new_cfg = cfg.append(std::as_const(b));
```
marehr, line 192: Just for my self: lvalue const

test/unit/core/configuration/configuration_test.cpp
```diff
@@ -171,6 +171,72 @@ TEST(configuration, exists_by_type_template)
...
+        EXPECT_TRUE((std::same_as<decltype(new_cfg), seqan3::configuration<foo, bar, bax>>));
+    }
+
+    {
+        seqan3::configuration<> cfg{};
+        bax b{};
+        auto new_cfg = cfg.append(std::as_const(b));
```
marehr, line 199: Just for my self: lvalue const, but empty lhs

Hmm from the above scheme, I would have expected `rvalue const`

marehr: This case I would expect in `append_configuration_element_special_cases`

test/unit/core/configuration/configuration_test.cpp
```diff
@@ -171,6 +171,72 @@ TEST(configuration, exists_by_type_template)
...
+        EXPECT_TRUE((std::same_as<decltype(new_cfg), seqan3::configuration<bax>>));
+    }
+}
+
+TEST(configuration, append_configuration)
+{
+    {
+        seqan3::configuration<foo, bar> cfg{};
+        auto new_cfg = cfg.append(seqan3::configuration{bax{}});
```
marehr, line 208: rvalue

test/unit/core/configuration/configuration_test.cpp
```diff
@@ -171,6 +171,72 @@ TEST(configuration, exists_by_type_template)
...
+        EXPECT_TRUE((std::same_as<decltype(new_cfg), seqan3::configuration<foo, bar, bax>>));
+    }
+
+    {
+        seqan3::configuration<foo, bar> cfg{};
+        seqan3::configuration cfg2{bax{}};
+        auto new_cfg = cfg.append(cfg2);
```
marehr, line 215: lvalue

test/unit/core/configuration/configuration_test.cpp
```diff
@@ -171,6 +171,72 @@ TEST(configuration, exists_by_type_template)
...
+        EXPECT_TRUE((std::same_as<decltype(new_cfg), seqan3::configuration<foo, bar, bax>>));
+    }
+
+    {
+        seqan3::configuration<foo, bar> cfg{};
+        seqan3::configuration cfg2{bax{}};
+        auto new_cfg = cfg.append(std::as_const(cfg2));
```
marehr, line 222: lvalue const

test/unit/core/configuration/configuration_test.cpp
```diff
@@ -171,6 +171,72 @@ TEST(configuration, exists_by_type_template)
...
+        EXPECT_TRUE((std::same_as<decltype(new_cfg), seqan3::configuration<foo, bar, bax>>));
+    }
+
+    {
+        seqan3::configuration<> cfg{};
+        seqan3::configuration cfg2{bax{}};
+        auto new_cfg = cfg.append(std::as_const(cfg2));
```
marehr, line 229: lvalue const, but LHS empty

test/unit/core/configuration/configuration_test.cpp
```diff
@@ -171,6 +171,72 @@ TEST(configuration, exists_by_type_template)
...
+        EXPECT_TRUE((std::same_as<decltype(new_cfg), seqan3::configuration<bax>>));
+    }
+
+    {
+        seqan3::configuration<> cfg{};
+        auto new_cfg = cfg.append(seqan3::configuration<>{});
```
marehr, line 235: rvalue, but RHS empty

include/seqan3/core/configuration/detail/concept.hpp
```diff
@@ -183,3 +181,51 @@ SEQAN3_CONCEPT config_element_pipeable_with =
...
+ * seqan3::detail::config_element_pipeable_with `config2_t`. If `config2_t` is a seqan3::configuration, the check will
+ * be expanded to every configuration element contained in the configuration type. Only if `config1_t` is combineable
+ * with every element stored inside of the given configuration, this helper variable template evaluates to `true`,
+ * otherwise `false`.
+ * If `config1_t` is a seqan3::configuration the same applies in combination with the configuration element `config2_t`.
+ * If both `config1_t` and `config2_t` are seqan3::configuration types, then the cartesian product between the
+ * configuration elements of the first configuration and the second configuration are tested.
+ */
+template <typename config1_t, typename config2_t>
+inline constexpr bool is_config_element_combineable_v = detail::config_element_pipeable_with<config1_t, config2_t>;
```
marehr, line 212: Considering our stable / experimental discussions, does this need to be seqan3::is_config_element_combineable_v ?

marehr (commented):
Thank you so much, this reduces a lot of boiler code. A couple of minor remarks ❤️

include/seqan3/core/configuration/configuration.hpp
rrahn: ok, yes. That seems plausible :)

test/unit/core/algorithm/pipeable_config_element_test.cpp
rrahn: Well, before the order depended on which overload of the |-operator was invoked because for symmetry reasons we often had: `operator|(cfg_elem lhs, configuration rhs) { return rhs | lhs }`.
Now the order is always maintained, i.e. elements from lhs always come before elements of rhs.

test/unit/core/configuration/configuration_test.cpp
rrahn: Yeah, but I am not really sure if `const &&` is a necessary use case. Simply, because the configuration itself is not really generic since the types (configuration elements) that are added are fixed.

include/seqan3/core/configuration/detail/concept.hpp
rrahn: Well, configuration is at least right now supposed to be stable. That means that an error that uses entities that are not in user scope as well are hard to track for the user. So it seems reasonable to offer this in seqan3:: namespace. Though, it is  not something a user would necessarily use in it's own code base. Would that be something NOAPI would then be good for?

test/unit/core/configuration/configuration_test.cpp
marehr: Okay

include/seqan3/core/configuration/detail/concept.hpp
marehr: Interesting idea. Maybe NOAPI is in this case like exposition-only in the standard. We define the intent, but don't fully specify it. :)

marehr: I like that idea, can you add \noapi to this?

marehr (approved):
This seems good, just add \noapi to that helper and I think this is good to go :)

include/seqan3/core/configuration/detail/concept.hpp
```diff
@@ -183,3 +181,51 @@ SEQAN3_CONCEPT config_element_pipeable_with =
...
+ * \details
+ *
+ * This helper variable template checks if `config1_t` fulfills the concept requirements
+ * seqan3::detail::config_element_pipeable_with `config2_t`. If `config2_t` is a seqan3::configuration, the check will
+ * be expanded to every configuration element contained in the configuration type. Only if `config1_t` is combineable
+ * with every element stored inside of the given configuration, this helper variable template evaluates to `true`,
+ * otherwise `false`.
+ * If `config1_t` is a seqan3::configuration the same applies in combination with the configuration element `config2_t`.
+ * If both `config1_t` and `config2_t` are seqan3::configuration types, then the cartesian product between the
+ * configuration elements of the first configuration and the second configuration are tested.
```
marehr, line 209:
```suggestion
 * configuration elements of the first configuration and the second configuration are tested.
 *
 * \noapi{"Exposition only"}
```

include/seqan3/core/configuration/detail/concept.hpp
marehr: @rrahn I did that for you :)

marehr:
@rrahn I added the exposition only text and rewrote the history

image

I'll merge after CI finishes.

PR #2336, opened 2021-01-13
[MISC] Fix html print_line break symbol.

smehringer:
The next commit being part of some argument parser refactoring (all changes shown in #2319).

Small fix to a html break


joergi-w (approved):
The `<br />` tag seems not to be wrong (e.g. [link](https://www.webmaster-crashkurs.de/zeilenumbruch-br.htm)), so why change it? Maybe don't call it a fix in the commit message. I'm fine with both versions, so I approve.

eseiler:
See also https://stackoverflow.com/a/1946446, but I find the `<br>` tags easier to read and we don't do xhtml

smehringer:
You're right the tag is not incorrect but we should be consistent. I'll remove the fix from the commit message :)

marehr (approved):

PR #2340, opened 2021-01-15
[FEATURE] seqan3::sequence_record

marehr:
This implements the Resolution:

* [1. Step one record type with all member functions.](https://docs.google.com/document/d/1EuGAtT0X5tDbBNFdHeuRfU2RRVOne8CSOJt3lI0U5UY/edit#bookmark=id.c1bmao8g76c5)

This implements
* sequence_record
* adds member to sequence_record, names were decided here: https://docs.google.com/document/d/1EuGAtT0X5tDbBNFdHeuRfU2RRVOne8CSOJt3lI0U5UY/edit#heading=h.vha98e10523x
* and does some infrastructure


include/seqan3/io/record.hpp
```diff
@@ -268,6 +272,436 @@ struct record : detail::transfer_template_args_onto_t<field_types, std::tuple>
...
+    //!\copydoc seqan3::record::qualities
+    decltype(auto) qualities() &
+    {
+        return get_impl(field_constant<seqan3::field::qual>{}, static_cast<base_type &>(*this));
+    }
+    //!\copydoc seqan3::record::qualities
+    decltype(auto) qualities() const &
+    {
+        return get_impl(field_constant<seqan3::field::qual>{}, static_cast<base_type const &>(*this));
+    }
```
eseiler, line 337: Wouldn't `quality` be nore consistent?

include/seqan3/io/record.hpp
```diff
@@ -268,6 +272,436 @@ struct record : detail::transfer_template_args_onto_t<field_types, std::tuple>
...
+    decltype(auto) tags() &
+    {
+        return get_impl(field_constant<seqan3::field::tags>{}, static_cast<base_type &>(*this));
+    }
+    //!\copydoc seqan3::record::tags
+    decltype(auto) tags() const &
+    {
+        return get_impl(field_constant<seqan3::field::tags>{}, static_cast<base_type const &>(*this));
+    }
+
```
eseiler, line 695: What about `bit_score` and `evalue`?

They are currently unused, but will they be removed?

eseiler (changes requested):
Minor stuff

marehr (commented):
A coupleof copy paste errors

include/seqan3/io/record.hpp
```diff
@@ -268,6 +272,436 @@ struct record : detail::transfer_template_args_onto_t<field_types, std::tuple>
...
+    {
+        return get_impl(field_constant<seqan3::field::react>{}, static_cast<base_type &>(*this));
+    }
+    //!\copydoc seqan3::record::reactivity
+    decltype(auto) reactivity() const &
+    {
+        return get_impl(field_constant<seqan3::field::react>{}, static_cast<base_type const &>(*this));
+    }
+
+    //!\brief Reactivity error values given in a vector corresponding to REACT.
```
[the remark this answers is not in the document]
marehr: @joergi-w Can you help? 

Note to myself: Change description above.

include/seqan3/io/record.hpp
marehr: I didn't add them, because we don't use those values, yet.

include/seqan3/io/record.hpp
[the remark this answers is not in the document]
joergi-w: Yes, REACT means the reactivity. (I guess in the original implementation we had an enum that names all the fields in capital letters.) Each value of `seqan3::record::reactivity_errors` corresponds to a value in `seqan3::record::reactivity`, and denotes the possible error of the reactivity value.

include/seqan3/io/record.hpp
```diff
@@ -268,6 +272,436 @@ struct record : detail::transfer_template_args_onto_t<field_types, std::tuple>
...
+    {
+        return get_impl(field_constant<seqan3::field::offset>{}, static_cast<base_type &>(*this));
+    }
+    //!\copydoc seqan3::record::offset
+    decltype(auto) offset() const &
+    {
+        return get_impl(field_constant<seqan3::field::offset>{}, static_cast<base_type const &>(*this));
+    }
+
+    //!\brief Base pair probability matrix of interactions, usually a matrix of float numbers.
```
marehr, line 360: @joergi-w Do you know why it is a `matrix`? From what I can see it is a vector that matches the sequence, i.e. each letter in a sequence has one numerical value.

In the test you called it interactions, would `sequence_interactions` be a good name?

Furthermore, in the test the type of the `base_pair_probability` is `std::vector<std::vector<uint8_t>>` (at least for vienna) and not float.

I'm slightly confused.

joergi-w: > @joergi-w Do you know why it is a `matrix`? From what I can see it is a vector that matches the sequence, i.e. each letter in a sequence has one numerical value.

It is a matrix, because each letter in a sequence is paired with every other letter in the same sequence. The float values are the probability of each pairing.
 
> In the test you called it interactions, would `sequence_interactions` be a good name?

An interaction is a 1:1 pairing while the whole matrix describes many:many pairings. A compromise name is interaction probabilities.

> Furthermore, in the test the type of the `base_pair_probability` is `std::vector<uint8_t>` (at least for vienna) and not float.
> 
> I'm slightly confused.

Probably I multiply each value with some factor to work in int-space... I can have a deeper look this afternoon.

marehr: Thank you for the answer!

> Probably I multiply each value with some factor to work in int-space... I can have a deeper look this afternoon.

That would be awesome :)

> It is a matrix, because each letter in a sequence is paired with every other letter in the same sequence. The float values are the probability of each pairing.

This sound like, if the sequence has size n, then you have a mapping [n] x [n] ↦ float. That would mean n² values, but in the test data we only have n values.

One example is:

```
> example 2
UUGGAGUACACAACCUGUACACUCUUUC
..(((((..(((...)))..)))))... (-3.71)
```

Hmm okay, I have looked at the wrong place, bpp has the following type:

```
std::vector<std::set<std::pair<double, size_t>>> & bpp
```

Do you still know why a hash map wasn't enough as data-structure?

I see, so format_vienna does not only give you the sequence, pseudoknot structure and energy, but also computes that bpp-matrix. In spirit how the alignment is computed from the other values.

joergi-w: > Do you still know why a hash map wasn't enough as a data-structure?

Yes, I need access in the order of their probability (the double type value) and this is not a unique key for a hash map. 

> I see, so format_vienna does not only give you the sequence, pseudoknot structure and energy, but also computes that bpp-matrix. In spirit how the alignment is computed from the other values.

Exactly. It becomes clearer and more useful, as soon as we implement further structure formats. But in order to transform LaRA 2 to SeqAn3 I need to wait for pairwise alignment with position-specific score matrices (pssm), so there is low priority for the other formats, currently.

joergi-w: > I need access in the order of their probability (the double type value)

Actually with `set<pair>` I wanted to implement a _priority queue_, and this was the best way.

marehr:
@eseiler Can you have a second look? ty 💟

eseiler (approved):
LGTM

The sequence_record commit should probably be `[FEATURE]`

marehr:
Damn I skipped the first reviewer :( Sorry 😢

marehr:
@joergi-w I accidentally skipped the first review. Can you do it? 🙇🏽

joergi-w:
So I'm actually second reviewer 🎉 
Sure, I will do it until Friday (today I already reviewed 3 PRs and I need to take a break).

joergi-w (approved):
Looks great! I have one question, but I agree to merge as it is.

test/unit/io/sequence_file/sequence_file_format_embl_test.cpp
```diff
@@ -119,6 +119,9 @@ SQ Sequence 1859 BP; 609 A; 314 C; 355 G; 581 T; 0 other;
         {
             EXPECT_EQ(seqan3::get<seqan3::field::id>(*it), ids[i]);
             EXPECT_EQ(seqan3::get<seqan3::field::seq>(*it), seqs[i]);
+
+            EXPECT_EQ((*it).id(), ids[i]);
```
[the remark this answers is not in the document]
marehr: We discussed that just yesterday :)

It is not guaranteed that C++20 iterators have a pointer access operator (they don't require that). The reason is simple, you could not give back a temporary object, like a proxy, if you would require that. So the safest way, since C++20 is to write `(*it).member` instead of `it->member`.

PR #2341, opened 2021-01-16
__cpp_lib_* fixes

marehr:
This PR fixes the following:
* __cpp_lib_endian is not known for `g++-8 -std=c++2a`, but it still defines `std::endian`
* __cpp_lib_to_address is not known for `g++-8/9 -std=c++2a`, but it still defines `std::to_address`
* further this PR makes the workaround macro names consistent in the naming scheme and behaviour:
  * `SEQAN3_CPP_LIB_*` corresponds to the definition of `__cpp_lib_*`
  * It always has defined / undefined semantics like the original feature macro `__cpp_lib_*`
    * Hence `#ifdef SEQAN3_CPP_LIB_*` means that the functionality is already implemented by the standard library
      * Or was disabled by a library user that might have conflicts with our shim/polyfill implementation
    * and`#ifndef SEQAN3_CPP_LIB_*` means that the functionality is missing in the standard library
      * in this case we define a shim/polyfill implementation

### DIFF
--- a/include/seqan3/std/bit
+++ b/include/seqan3/std/bit
@@ -42,7 +42,17 @@
 #   endif
 #endif
 
-#ifndef __cpp_lib_endian
+//!\brief A workaround for __cpp_lib_endian for gcc version 8.x (in C++20 mode).
+//!       Those versions implemented std::endian, but did not define that feature detection macro.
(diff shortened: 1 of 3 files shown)


MitraDarja (approved):
LGTM

eseiler (approved):

PR #2342, opened 2021-01-16
[MISC] move llvm's std::span implementation into contrib

marehr:


include/seqan3/contrib/span/span
```diff
@@ -0,0 +1,436 @@
+// -*- C++ -*-
+//!\cond
+//===------------------------------ span ---------------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is dual licensed under the MIT and the University of Illinois Open
+// Source Licenses. See LICENSE.TXT for details.
+//
+//===---------------------------------------------------------------------===//
```
Irallia, line 10: I assume that this is intended and still applies?

include/seqan3/contrib/span/span
```diff
@@ -0,0 +1,436 @@
+// -*- C++ -*-
```
Irallia, line 1: Do we need this? What does it mean, just that it is C++ code?

include/seqan3/std/span
```diff
@@ -1,13 +1,10 @@
 // -*- C++ -*-
```
Irallia, line 1: And we keep this line?

Irallia (changes requested):
LGFM, just some questions from my side.

include/seqan3/contrib/span/span
marehr: Yep, we copied that span implementation from llvm

include/seqan3/contrib/span/span
marehr: Yes, this is an "editor" standard to set the file extension for extension less files.

include/seqan3/std/span
marehr: Yes!

Irallia (approved):
Okay thanks for the explanation! ❤️

marehr:
Can you @eseiler  take over for @smehringer ?

eseiler (approved):

PR #2344, opened 2021-01-18
[DOC] Put scoring scheme config into correct doxygen group.

rrahn:
The scoring scheme configuration documentation did not appear in the alignment/configuration module.


eaasna (approved):
Thanks!

marehr (approved):

PR #2272, opened 2020-11-18
[INFRA] Remove algorithm/bound.hpp & alignment/band/static_band.hpp

Irallia:
Resolves partly seqan/product_backlog#160

Removed because not used.

Details for this: https://gist.github.com/rrahn/f6b09ad67bebd14bd9c58f77c53e0450

PR #2345, opened 2021-01-18
[DOC] Deprecate seqan3::static_band documentation.

smehringer:
In #2272 only the header was marked as deprecated in doxygen but not the entity itself. This way you don't really see the deprecation in the API documentation.

### DIFF
--- a/include/seqan3/alignment/band/static_band.hpp
+++ b/include/seqan3/alignment/band/static_band.hpp
@@ -22,8 +22,9 @@
 namespace seqan3
 {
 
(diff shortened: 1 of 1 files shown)


SGSSGene (approved):
Short and to the point.
Does this also mean that including "static_band.hpp" is deprecated?

smehringer:
> Short and to the point.
> Does this also mean that including "static_band.hpp" is deprecated?

Yep

eseiler (approved):

PR #2347, opened 2021-01-18
[API] Move the list of file formats into the sequence file input.

rrahn:
This PR removes the variant used internally for the sequence file input. The main advantage is that the template list of the input file gets reduced by one parameter. This change does not effect any critical path of the sequence file input, since the mechanism for storing the selected format is completely internal and never exposed to the user. 
Reducing the template parameter also shows that the interface becomes easier to use, because the user does not need to remember to set the correct file format in the template parameter list of the class template, when she calls the interface with a specific format tag.
Note the current change keeps the internal usage of a variant in order to use the currently existing interfaces to extract the format type from the extension of a file path. Thus, the remaining classes are still valid and the internals can be adapted once everything is ported to the new internal format selection mechanism.

Another thing I could observe, is the misleading name of `valid_formats`. While this PR does still preserve this additional type definition, I think it should be removed. The reason is that the format is a generic part in the file I/O. Basically, any file that preserves the sequence information could be used inside of the sequence input file. The file however, cannot know the entire universe of formats and hence should not expose something that might indicate something like a restricted list of formats that are only allowed.

### DIFF
--- a/include/seqan3/io/sequence_file/input.hpp
+++ b/include/seqan3/io/sequence_file/input.hpp
@@ -450,7 +450,15 @@ public:
         secondary_stream = detail::make_secondary_istream(*primary_stream, filename);
 
         // initialise format handler or throw if format is not found
-        detail::set_format(format, filename);
+        using format_variant_t = std::variant<format_embl, format_fasta, format_fastq, format_genbank, format_sam>;
+        format_variant_t format_variant{};
+        detail::set_format(format_variant, filename);
+
+        std::visit([&] (auto && selected_format)
+        {
+            using format_t = std::remove_cvref_t<decltype(selected_format)>;
+            format = std::make_unique<selected_input_format<format_t>>();
+        }, format_variant);
     }
     /* NOTE(h-2): Curiously we do not need a user-defined deduction guide for the above constructor.
      * A combination of default template parameters and auto-deduction guides works as expected,
@@ -482,7 +490,7 @@ public:
                         file_format        const & SEQAN3_DOXYGEN_ONLY(format_tag),
                         selected_field_ids const & SEQAN3_DOXYGEN_ONLY(fields_tag) = selected_field_ids{}) :
         primary_stream{&stream, stream_deleter_noop},
-        format{detail::sequence_file_input_format_exposer<file_format>{}}
+        format{std::make_unique<selected_input_format<file_format>>()}
     {
         static_assert(list_traits::contains<file_format, valid_formats>,
                       "You selected a format that is not in the valid_formats of this file.");
@@ -501,7 +509,7 @@ public:
                         file_format        const & SEQAN3_DOXYGEN_ONLY(format_tag),
                         selected_field_ids const & SEQAN3_DOXYGEN_ONLY(fields_tag) = selected_field_ids{}) :
         primary_stream{new stream_t{std::move(stream)}, stream_deleter_default},
-        format{detail::sequence_file_input_format_exposer<file_format>{}}
+        format{std::make_unique<selected_input_format<file_format>>()}
     {
         static_assert(list_traits::contains<file_format, valid_formats>,
                       "You selected a format that is not in the valid_formats of this file.");
@@ -588,9 +596,12 @@ public:
     }
     //!\}
 
+    //!\brief The input file options type.
+    using sequence_file_input_options_type
+        = sequence_file_input_options<typename traits_type::sequence_legal_alphabet,
+                                      selected_field_ids::contains(field::_seq_qual_deprecated)>;
     //!\brief The options are public and its members can be set directly.
-    sequence_file_input_options<typename traits_type::sequence_legal_alphabet,
-                             selected_field_ids::contains(field::_seq_qual_deprecated)> options;
+    sequence_file_input_options_type options;
 
 protected:
     //!\privatesection
@@ -623,14 +634,9 @@ protected:
     bool first_record_was_read{false};
     //!\brief File is at position 1 behind the last record.
     bool at_end{false};
-
-    //!\brief Type of the format, an std::variant over the `valid_formats`.
-    using format_type = typename detail::variant_from_tags<valid_formats,
-                                                           detail::sequence_file_input_format_exposer>::type;
-    //!\brief The actual std::variant holding a pointer to the detected/selected format.
-    format_type format;
     //!\}
 
+private:
     //!\brief Tell the format to move to the next record and update the buffer.
     void read_next_record()
     {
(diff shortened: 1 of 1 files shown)


include/seqan3/io/sequence_file/input.hpp
```diff
@@ -325,7 +318,7 @@ class sequence_file_input
     //!\brief A seqan3::fields list with the fields selected for the record.
     using selected_field_ids    = selected_field_ids_;
     //!\brief A seqan3::type_list with the possible formats.
-    using valid_formats         = valid_formats_;
+    using valid_formats         = type_list<format_embl, format_fasta, format_fastq, format_genbank, format_sam>;
```
smehringer, line 321: But with this change you cannot restrict your sequence file to only accept e.g. FASTA and FASTQ files?

MitraDarja (changes requested):
Just a question and one small thing in the documentation.

include/seqan3/io/sequence_file/input.hpp
```diff
@@ -449,7 +442,15 @@ class sequence_file_input
...
-        detail::set_format(format, filename);
+        using format_variant_t = std::variant<format_embl, format_fasta, format_fastq, format_genbank, format_sam>;
+        format_variant_t format_variant{};
+        detail::set_format(format_variant, filename);
+
+        std::visit([&] (auto && selected_format)
+        {
+            using format_t = std::remove_cvref_t<decltype(selected_format)>;
+            format = std::make_unique<selected_input_format<format_t>>();
+        }, format_variant);
```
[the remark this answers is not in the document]
rrahn: Of course. We have currently functions in place that selects the format based on the file extension. This detail function sets the format in a variant it is called with. To keep the changes for this minimal and limited to only the sequence file the detail selection function was not touched. So in the first iteration I simply put the variant into the constructor (out of the file scope) and use this visit function to extract the selected format. This format is then stored type-erased in the format base pointer.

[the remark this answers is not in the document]
rrahn: Later, once everyone approves with this we can adapt the other files as well and then replace the detail function with some less boilerplate version which does not need the additional variant anymore.

include/seqan3/io/sequence_file/input.hpp
rrahn: This is according to the resolution that we do not yet give the opportunity to specify the set of valid formats. 
This can be added when we resolved how we want to handle formats. It would be already possible to have a dedicated constructor for this, but if not decided otherwise this will first come after 3.1. 
Also please see my notes above. I think valid formats should be removed entirely. First, it won't be needed, second the naming is misleading. Or looking at it in a different way: Is there anything that would make it worthwhile for a file to expose the formats a user can specify? After all, the user can specify any file as long as it fulfills the involved concepts.

MitraDarja (approved):
Thanks for the explaination. :)

include/seqan3/io/sequence_file/input.hpp
```diff
@@ -307,12 +305,7 @@ struct sequence_file_input_default_traits_aa : sequence_file_input_default_trait
 
 template <
     sequence_file_input_traits traits_type_ = sequence_file_input_default_traits_dna,
-    detail::fields_specialisation selected_field_ids_ = fields<field::seq, field::id, field::qual>,
-    detail::type_list_of_sequence_file_input_formats valid_formats_ = type_list<format_embl,
-                                                                                format_fasta,
-                                                                                format_fastq,
-                                                                                format_genbank,
-                                                                                format_sam>>
```
marehr, line 315: For an easier upgrade-path, I would want to keep this template parameter for now. But effectively shutting that template parameter down. I'm not sure how to emit deprecation notices on template parameters, but we can think of a creative way :)

I have this rough idea, how to do it:
https://godbolt.org/z/fPq4sG

include/seqan3/io/sequence_file/input.hpp
marehr: As I said in the above line, I would still support this for now, but only store it under-the-hood differently :)

include/seqan3/io/sequence_file/input.hpp
```diff
@@ -449,7 +442,15 @@ class sequence_file_input
         secondary_stream = detail::make_secondary_istream(*primary_stream, filename);
 
         // initialise format handler or throw if format is not found
-        detail::set_format(format, filename);
+        using format_variant_t = std::variant<format_embl, format_fasta, format_fastq, format_genbank, format_sam>;
```
marehr, line 445:
```suggestion
        using format_variant_t = typename detail::variant_from_tags<valid_formats,
                                                                    detail::sequence_file_input_format_exposer>::type;
```

include/seqan3/io/sequence_file/input.hpp
```diff
@@ -644,28 +642,100 @@ class sequence_file_input
...
+     * \tparam format_t The user specific format type to store.
+     *
+     * \details
+     *
+     * This class implements the format specific read operation based on the instantiated format type.
+     * The specfic format type is selected at runtime and stored through a pointer to the abstract
+     * seqan3::sequence_file_input::selected_input_format_base class. A virtual function call then ensures that the
+     * specific read record function of the selected format is invoked.
+     */
+    template <typename format_t>
```
marehr, line 686: Just for clarification: 

We could constrain `format_t` here that it only accepts valid sequence formats, right?

include/seqan3/io/sequence_file/input.hpp
```diff
@@ -644,28 +642,100 @@ class sequence_file_input
...
+     *
+     * \details
+     *
+     * This abstract base class is used to store the user given input format as a type-erased base class in a
+     * std::unique_ptr. There is exactly one derived type that implements this abstract base class called
+     * seqan3::sequence_file_input::selected_input_format which is a private subclass of
+     * seqan3::sequence_file_input. It is not exposed to the user and allows to hide the implementation detail of
+     * storing a specfic format instance which is first known at runtime.
+     */
+    struct selected_input_format_base
```
marehr, line 658:
```suggestion
    struct sequence_format_base
```

or

```suggestion
    struct selected_sequence_format
```
 Other suggestions?

marehr: I'm not sure, but I think it would be more clear if we would name it a bit more specific.

This base class can only be used within this class and only supports sequence file formats.

include/seqan3/io/sequence_file/input.hpp
```diff
@@ -644,28 +642,100 @@ class sequence_file_input
...
+     *
+     * \details
+     *
+     * This class implements the format specific read operation based on the instantiated format type.
+     * The specfic format type is selected at runtime and stored through a pointer to the abstract
+     * seqan3::sequence_file_input::selected_input_format_base class. A virtual function call then ensures that the
+     * specific read record function of the selected format is invoked.
+     */
+    template <typename format_t>
+    struct selected_input_format final : public selected_input_format_base
```
marehr, line 687:
```suggestion
    struct sequence_format final : public sequence_format_base
```

or

```suggestion
    struct sequence_format final : public selected_sequence_format
```
 Other suggestions?

include/seqan3/io/sequence_file/input.hpp
```diff
@@ -644,28 +642,100 @@ class sequence_file_input
...
+         *
+         * \param[in, out] instream The input stream to extract the next record from.
+         * \param[in, out] record_buffer The record buffer to fill.
+         * \param[in] options User specific format options set from outside.
+         *
+         * \details
+         *
+         * Invokes the actual read sequence record function for the selected format and fills the record accordingly.
+         */
+        void read_sequence_record(std::istream & instream,
```
marehr, line 710: Couldn't we add the documentation to the "interface" and just do here a `\override`?

marehr (commented):
Really great work! In general, I wouldn't just remove any template parameter without any chance to "warn" our users about this.

I like the changes a lot and I think they can co-exists with the current variant implementation, but they show the might to remove that template parameter entirely.

include/seqan3/io/sequence_file/input.hpp
rrahn: Well, the intention was to refactor this into a base class in the next steps to get rid of the duplicated code. 
In this case, you would have something more generic like I chose here. But the renaming can also be done later I suppose.

include/seqan3/io/sequence_file/input.hpp
rrahn: I think restricting it in the constructor would be the right place. As said, this would be more part of a base class, which then is not format agnostic.

include/seqan3/io/sequence_file/input.hpp
rrahn: Maybe?

include/seqan3/io/sequence_file/input.hpp
rrahn: Ok, then I will first apply the internal changes and keep the templates as are. Then this must be deprecated in a follow up PR.

marehr (approved):
Can this be squashed merged?

PR #2354, opened 2021-01-26
[FIX] seqan3::detail::get_or_ignore should accept all tuple-like types

marehr:
The old implementation of seqan3::detail::get_or_ignore seemed to work for
many tuple types, but failed for tuple types that have less template
parameters than tuple elements.

### DIFF
--- a/include/seqan3/io/detail/record.hpp
+++ b/include/seqan3/io/detail/record.hpp
@@ -137,20 +137,26 @@ auto const & get_or_ignore(record<field_types, field_ids> const & r)
 }
 
 //!\copydoc seqan3::detail::get_or_ignore
-template <size_t i, template <tuple_like ...types_> typename tuple_like_t, typename ...types>
-auto & get_or_ignore(tuple_like_t<types...> & t)
+template <size_t i, typename tuple_t>
+//!\cond
+    requires tuple_like<tuple_t>
+//!\endcond
+auto & get_or_ignore(tuple_t & t)
 {
-    if constexpr (i < sizeof...(types))
+    if constexpr (i < std::tuple_size_v<tuple_t>)
         return std::get<i>(t);
     else
         return std::ignore;
 }
 
 //!\copydoc seqan3::detail::get_or_ignore
-template <size_t i, template <tuple_like ...types_> typename tuple_like_t, typename ...types>
-auto const & get_or_ignore(tuple_like_t<types...> const & t)
+template <size_t i, typename tuple_t>
+//!\cond
+    requires tuple_like<tuple_t>
+//!\endcond
+auto const & get_or_ignore(tuple_t const & t)
 {
-    if constexpr (i < sizeof...(types))
+    if constexpr (i < std::tuple_size_v<tuple_t>)
         return std::get<i>(t);
     else
         return std::ignore;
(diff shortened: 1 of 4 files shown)


include/seqan3/io/detail/record.hpp
```diff
@@ -137,20 +137,26 @@ auto const & get_or_ignore(record<field_types, field_ids> const & r)
 }
 
 //!\copydoc seqan3::detail::get_or_ignore
-template <size_t i, template <tuple_like ...types_> typename tuple_like_t, typename ...types>
-auto & get_or_ignore(tuple_like_t<types...> & t)
+template <size_t i, typename tuple_t>
+//!\cond
+    requires tuple_like<tuple_t>
+//!\endcond
```
joergi-w, line 143: The following would be shorter and does not need the `cond` part – is there a reason to not write it this way?
```suggestion
template <size_t i, tuple_like tuple_t>
```

joergi-w (changes requested):
I would rename the `test/unit/io/detail/detail_record_test.cpp` test file to something that contains tuple, because it is basically about it. The repetition of detail is also not meaningful. How about `test/unit/io/detail/record_tuple_test.cpp`?

include/seqan3/io/detail/record.hpp
```diff
@@ -137,20 +137,26 @@ auto const & get_or_ignore(record<field_types, field_ids> const & r)
...
         return std::ignore;
 }
 
 //!\copydoc seqan3::detail::get_or_ignore
-template <size_t i, template <tuple_like ...types_> typename tuple_like_t, typename ...types>
-auto const & get_or_ignore(tuple_like_t<types...> const & t)
+template <size_t i, typename tuple_t>
+//!\cond
+    requires tuple_like<tuple_t>
+//!\endcond
```
marehr, line 156:
```suggestion
template <size_t i, tuple_like tuple_t>
```

joergi-w (approved):
Thank you!

eseiler (approved):

test/unit/io/detail/detail_record_test.cpp
```diff
@@ -0,0 +1,107 @@
...
+{};
+
+template <size_t index, typename id_t>
+struct tuple_element<index, custom_tuple<id_t>> : public tuple_element<index, typename custom_tuple<id_t>::base_t>
+{};
+} // namespace std
+
+TEST(get_or_ignore, custom_tuple)
+{
+    custom_tuple<std::string> tuple{};
```
eseiler, line 98: This is what did not work before

marehr:
> I would rename the `test/unit/io/detail/detail_record_test.cpp` test file to something that contains tuple, because it is basically about it. The repetition of detail is also not meaningful. How about `test/unit/io/detail/record_tuple_test.cpp`?

Sorry, I saw your comment and read it, but forgot to answer.  The main problem is that we already have a test that is called `record_test` and it is not possible to have multiple cmake targets with the same name. I have chosen `detail_record_test.hpp`, because the test is in `detail/record_test.hpp` and other test case that have name-clashes prefix one folder up.

I opened an issue https://github.com/seqan/product_backlog/issues/257 (section: Problem with unique test target names) that discusses possible "naming schemes" for tests with name clashes.

PR #2355, opened 2021-01-27
[MISC] add seqan3::detail::record_like

marehr:
This adds a better concept for record types. 

Before it just checked if the record type is `seqan3::record`, now it also allows derivatives of `seqan3::record`.


eaasna (approved):
Looks good, thanks!

eseiler (approved):

PR #2357, opened 2021-01-28
[FIX] gcc7: seqan3::get<seqan3::field>(record const &&)

marehr:
This PR introduces
* a workaround macro `SEQAN3_WORKAROUND_GCC_94967`
* uses it on known places
* adds a test to check that `seqan3::get<seqan3::field>(record const &&)` works

### DIFF
--- a/test/unit/alignment/matrix/detail/affine_cell_proxy_test.cpp
+++ b/test/unit/alignment/matrix/detail/affine_cell_proxy_test.cpp
@@ -7,10 +7,9 @@
 
 #include "gtest/gtest.h"
 
-#include <seqan3/test/expect_same_type.hpp>
-
 #include <seqan3/alignment/matrix/detail/affine_cell_proxy.hpp>
 #include <seqan3/core/tuple/common_tuple.hpp>
+#include <seqan3/test/expect_same_type.hpp>
 
 //------------------------------------------------------------------------------
 // score cell proxy
--- a/test/unit/alignment/matrix/detail/affine_cell_proxy_test.cpp
+++ b/test/unit/alignment/matrix/detail/affine_cell_proxy_test.cpp
@@ -7,6 +7,8 @@
 
 #include "gtest/gtest.h"
 
+#include <seqan3/test/expect_same_type.hpp>
+
 #include <seqan3/alignment/matrix/detail/affine_cell_proxy.hpp>
 #include <seqan3/core/tuple/common_tuple.hpp>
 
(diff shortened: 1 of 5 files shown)


test/unit/alignment/matrix/detail/affine_cell_proxy_test.cpp
```diff
@@ -7,6 +7,8 @@
 
 #include "gtest/gtest.h"
 
+#include <seqan3/test/expect_same_type.hpp>
```
MitraDarja, line 10: Alphabetical order and without a line in between?

MitraDarja (changes requested):
LGTM, just two questions:
Did we change the order of the includes, so the test includes are separate and come first?  (If yes, you can already request a second reviewer and I will approve, if not change the include order.;))

My second question is, what does this comment mean:
`// ^^^ workaround / no workaround vvv`
especially the ^^^ and the vvv?

test/unit/alignment/matrix/detail/affine_cell_proxy_test.cpp
marehr: To be honest, we don't have an explicit rule, yet. We have some tests that split test includes from library includes and some not. So it's a good question. I change it to alphabetically order.

test/unit/alignment/matrix/detail/affine_cell_proxy_test.cpp
```diff
@@ -7,6 +7,8 @@
...
+
 #include <seqan3/alignment/matrix/detail/affine_cell_proxy.hpp>
 #include <seqan3/core/tuple/common_tuple.hpp>
```
marehr, line 13:
```suggestion
#include <seqan3/alignment/matrix/detail/affine_cell_proxy.hpp>
#include <seqan3/core/tuple/common_tuple.hpp>
#include <seqan3/test/expect_same_type.hpp>
```

test/unit/io/record_test.cpp
```diff
@@ -10,10 +10,12 @@
 
 #include <gtest/gtest.h>
 
+#include <seqan3/test/expect_range_eq.hpp>
+#include <seqan3/test/expect_same_type.hpp>
+
```
marehr, line 15:
```suggestion
```

test/unit/io/record_test.cpp
```diff
@@ -10,10 +10,12 @@
...
 #include <seqan3/alphabet/nucleotide/dna4.hpp>
 #include <seqan3/core/detail/debug_stream_alphabet.hpp>
 #include <seqan3/io/record.hpp>
-#include <seqan3/test/expect_range_eq.hpp>
 #include <seqan3/utility/tuple/concept.hpp>
```
marehr, line 19:
```suggestion
#include <seqan3/test/expect_range_eq.hpp>
#include <seqan3/test/expect_same_type.hpp>
#include <seqan3/utility/tuple/concept.hpp>
```

marehr:
> My second question is, what does this comment mean:
> `// ^^^ workaround / no workaround vvv`
> especially the ^^^ and the vvv?

the code ABOVE (= ^^^) is the workaround, the code below (= vvv) is the original intended code.

MitraDarja (approved):

eseiler (approved):

Issue #2353, opened 2021-01-26
Extensions with dot don't work in argument parser


h-2:
## Platform

  - SeqAn version: master from around December
  - Operating system: RHEL 7
  - Compiler: gcc10.2

### How to repeat the problem

```cpp
parser.add_positional_option(p1, "FOO", seqan3::input_file_validator{{"foo.bar"}});
```

This code is accepted and nothing in the documentation suggests that this doesn't work. But when I pass a file called  `X.foo.bar` to the program, the argument parser throws and says 

```
... the following valid extensions: [foo.bar]. Got ".bar" instead!"
```

I know that we have special treatment for files now (which is nice but might be better served by a different validator), but having an extension with a dot is quite common, even independent of compression.

### Expected behaviour

This should work.

### Actual behaviour

It doesn't.

### Possible fix

This should be quite simple. Just check if the path ends_with the respective string instead of calling `.extension()`.

marehr:
How is this related to https://github.com/seqan/seqan3/issues/1176?

eseiler:
This would work:

```patch
diff --git a/include/seqan3/argument_parser/validators.hpp b/include/seqan3/argument_parser/validators.hpp
index b177467d..59192522 100644
--- a/include/seqan3/argument_parser/validators.hpp
+++ b/include/seqan3/argument_parser/validators.hpp
@@ -374,7 +374,8 @@ protected:
                                                      " one of the following valid extensions:", extensions, "!")};
 
         // Drop the dot.
-        std::string drop_less_ext = path.extension().string().substr(1);
+        std::string const filename = path.filename().string();
+        std::string drop_less_ext = filename.substr(filename.find(".") + 1);
 
         // Compares the extensions in lower case.
         auto case_insensitive_equal_to = [&] (std::string const & ext)

```

Question: 
* `foo.bar.bax.baz` would then also only validate if `bar.bax.bax` is given as valid extension, etc.

----

> How is this related to #1176?

Similar, we may allow for compression. Passing `fastq` as valid extension should allow `fastq.gz`, `fastq.bz2`, etc. without the user explicitly specifying the compression extensions.
Svenja worked on it, but then we decided to do it differently...

marehr:
### 2021-02-01 Core-Meeting

We match the suffix. We need an implementation for c++17 for `seqan3::detail::string_ends_with` which behaves like C++20 `std::string::ends_with`.

PR #2363, opened 2021-02-01
[FIX] Extension with dots not working in validator

eseiler:
Resolves https://github.com/seqan/seqan3/issues/2353

### DIFF
--- a/include/seqan3/argument_parser/validators.hpp
+++ b/include/seqan3/argument_parser/validators.hpp
@@ -373,23 +373,26 @@ protected:
             throw validation_error{detail::to_string("The given filename ", path.string(), " has no extension. Expected"
                                                      " one of the following valid extensions:", extensions, "!")};
 
-        // Drop the dot.
-        std::string drop_less_ext = path.extension().string().substr(1);
+        std::string file_path{path.filename().string()};
+
+        // Leading dot indicates a hidden file is not part of the extension.
+        if (file_path.front() == '.')
+            file_path.erase(0, 1);
+
+        // Store a string_view containing all extensions for a better error message.
+        std::string const all_extensions{file_path.substr(file_path.find(".") + 1)};
 
         // Compares the extensions in lower case.
-        auto case_insensitive_equal_to = [&] (std::string const & ext)
+        auto case_insensitive_ends_with = [&] (std::string const & ext)
         {
-            return std::ranges::equal(ext, drop_less_ext, [] (char const chr1, char const chr2)
-                   {
-                       return std::tolower(chr1) == std::tolower(chr2);
-                   });
+            return case_insensitive_string_ends_with(file_path, ext);
         };
 
         // Check if requested extension is present.
-        if (std::ranges::find_if(extensions, case_insensitive_equal_to) == extensions.end())
+        if (std::ranges::find_if(extensions, case_insensitive_ends_with) == extensions.end())
         {
             throw validation_error{detail::to_string("Expected one of the following valid extensions: ", extensions,
-                                                     "! Got ", drop_less_ext, " instead!")};
+                                                     "! Got ", all_extensions, " instead!")};
         }
     }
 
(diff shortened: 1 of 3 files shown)


include/seqan3/argument_parser/validators.hpp
```diff
@@ -373,23 +373,36 @@ class file_validator_base
...
-                   {
-                       return std::tolower(chr1) == std::tolower(chr2);
-                   });
+            return string_ends_with(file_path, ext);
         };
 
+        bool ext_found{false};
+
+        // Iterate over all possible extensions, starting with the longest. E.g., "foo.bar.baz", "bar.baz", "baz".
+        for (size_t pos = file_path.find("."); !ext_found && pos != std::string_view::npos; pos = file_path.find("."))
```
h-2, line 395: What happens here? Why not just iterate over the valid extensions and check each time if `file_path` ends with that?

include/seqan3/argument_parser/validators.hpp
```diff
@@ -451,6 +464,24 @@ class file_validator_base
             return detail::to_string(" Valid file extensions are: [", extensions | views::join(std::string{", "}), "].");
     }
 
+    /*!\brief Helper function that checks if a string is a suffix of another string.
+     * \param str The string to be searched.
+     * \param suffix The suffix to be searched for.
+     * \returns `true` if `suffix` is a suffix of `str`, otherwise `false`.
+     * \sa https://en.cppreference.com/w/cpp/string/basic_string_view/ends_with
```
h-2, line 471: This does not behave as described, because it ignores case 💅🏻

include/seqan3/argument_parser/validators.hpp
eseiler: Hah, yeah no idea. Probably the amalgamation of different ideas...

SGSSGene (approved):
Sieht gut aus ✔

include/seqan3/argument_parser/validators.hpp
```diff
@@ -373,23 +373,27 @@ class file_validator_base
             throw validation_error{detail::to_string("The given filename ", path.string(), " has no extension. Expected"
                                                      " one of the following valid extensions:", extensions, "!")};
 
-        // Drop the dot.
-        std::string drop_less_ext = path.extension().string().substr(1);
+        std::string const path_as_string{path.filename().string()};
+        std::string_view file_path{path_as_string};
```
marehr, line 377: Not sure about all of this string_views, do we really need them?

I think it would be fine to use strings here.

include/seqan3/argument_parser/validators.hpp
```diff
@@ -451,6 +455,23 @@ class file_validator_base
...
+    bool string_ends_with(std::string_view str, std::string_view suffix) const
+    {
+        size_t const suffix_length{suffix.size()};
+        size_t const str_length{str.size()};
+        return suffix_length > str_length ? 
+               false : 
+               std::ranges::equal(str.substr(str_length - suffix_length), suffix, [] (char const chr1, char const chr2)           
+               {
+                   return std::tolower(chr1) == std::tolower(chr2);
+               });
```
marehr, line 472: Can we use the definition from https://en.cppreference.com/w/cpp/string/basic_string_view/ends_with ?

```c++
size() >= sv.size() && compare(size() - sv.size(), npos, sv) == 0
```

One range less :)

include/seqan3/argument_parser/validators.hpp
```diff
@@ -451,6 +455,23 @@ class file_validator_base
             return detail::to_string(" Valid file extensions are: [", extensions | views::join(std::string{", "}), "].");
     }
 
+    /*!\brief Helper function that checks if a string is a suffix of another string. Case insensitive.
+     * \param str The string to be searched.
+     * \param suffix The suffix to be searched for.
+     * \returns `true` if `suffix` is a suffix of `str`, otherwise `false`.
+     */
+    bool string_ends_with(std::string_view str, std::string_view suffix) const
```
marehr, line 463:
```suggestion
    bool string_view_ends_with(std::string_view str, std::string_view suffix) const
```

marehr (commented):
Thank you, this looks excellent. Can we use the ends_with definition from cppreference.com?

include/seqan3/argument_parser/validators.hpp
marehr: We could also implement as new std::char_traits that make the string case-insensitive https://en.cppreference.com/w/cpp/string/char_traits

marehr (approved):

PR #2359, opened 2021-01-28
[MISC] introduce new concepts seqan3::detail::(writable_)pairwise_alignment

marehr:
This PR resolves https://github.com/seqan/product_backlog/issues/60 task
by implementing seqan3::detail::(writable_)pairwise_alignment.

PR #2365, opened 2021-02-02
introduce new seqan3::detail::transfer_type_modifier_onto

marehr:
This is needed to finish PR #2359 to workaround a nasty "bug" described in https://github.com/seqan/product_backlog/issues/286.


include/seqan3/core/detail/transfer_type_modifier_onto.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+namespace seqan3::detail
+{
+
+/*!\brief Transfers the type modifier `&`, `&&` and `const` (and any combination) to the target type.
+ * \implements seqan3::transformation_trait
+ * \tparam source_t The type you wish to transfer the type modifier from.
+ * \tparam target_t The type you wish to transfer the type modifier to.
+ * \ingroup core
+ * If the target_t already has a type modifier, lets say `const`, it will keep that type modifier.
```
joergi-w, line 27:
```suggestion
 * If the target_t already has a type modifier, let's say `const`, it will keep that type modifier.
```
Instead of "let's say" you can use a more formal "e.g.".

joergi-w (approved):

joergi-w:
In my opinion, this feature can become part of `template_inspection.hpp`, which contains already related functions (which I would love to see in stable (non-detail) API in the future in order to use them in my application).

include/seqan3/core/detail/transfer_type_modifier_onto.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+namespace seqan3::detail
+{
+
+/*!\brief Transfers the type modifier `&`, `&&` and `const` (and any combination) to the target type.
+ * \implements seqan3::transformation_trait
+ * \tparam source_t The type you wish to transfer the type modifier from.
+ * \tparam target_t The type you wish to transfer the type modifier to.
+ * \ingroup core
+ * If the target_t already has a type modifier, let's say `const`, it will keep that type modifier.
```
marehr, line 27:
```suggestion
 * If the target_t already has a type modifier, e.g. `const`, it will keep that type modifier.
```

marehr:
> In my opinion, this feature can become part of `template_inspection.hpp`, which contains already related functions (which I would love to see in stable (non-detail) API in the future in order to use them in my application).

`template_inspection.hpp` contains only utility that works on `template template` things. So it doesn't really fit. We decided to make core NOAPI that means you can use it, but we don't guarantee that it will never break.

include/seqan3/core/detail/transfer_type_modifier_onto.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+namespace seqan3::detail
+{
+
+/*!\brief Transfers the type modifier `&`, `&&` and `const` (and any combination) to the target type.
+ * \implements seqan3::transformation_trait
+ * \tparam source_t The type you wish to transfer the type modifier from.
+ * \tparam target_t The type you wish to transfer the type modifier to.
+ * \ingroup core
+ * If the target_t already has a type modifier, e.g. `const`, it will keep that type modifier.
```
eseiler, line 27:
```suggestion
 * If the `target_t` already has a type modifier, e.g. `const`, it will keep that type modifier.
```

include/seqan3/core/detail/transfer_type_modifier_onto.hpp
```diff
@@ -0,0 +1,70 @@
...
+                                                               std::is_lvalue_reference_v<target_t>,
+                                                               std::add_lvalue_reference_t<maybe_rvalue_reference_t>,
+                                                               maybe_rvalue_reference_t>;
+public:
+
+    //!\brief Transfers the type modifier `&`, `&&` and `const` (and any combination) to the target type.
+    using type = maybe_lvalue_reference_target_t;
+};
+
+/*!\brief Transfers the type modifier `&`, `&&` and `const` (and any combination) to the target type.
```
eseiler, line 60:
```suggestion
/*!\brief Transfers the type modifier `&`, `&&` and `const` (and any combination) to the target type
```

eseiler (changes requested):
💅

eseiler (approved):

PR #2370, opened 2021-02-05
[FEAT] adds sandboxed_path

SGSSGene:
This adds the sandboxed_path class which is a safer
std::filesystem::path variant.
It is going to be the base for a refactored tmp_directory class.


test/include/seqan3/test/sandboxed_path.hpp
```diff
@@ -0,0 +1,334 @@
...
+ * like it. In addition it receives a sandbox directory at construction time.
+ * Functions are overloaded and a check for the invariant is added.
+ *
+ *  Invariant:
+ *  - sandboxed_path is always converted to an absolute path
+ *  - sandboxed_path always points to a file inside a given sandbox directory.
+ *  - sandbox directory is unmutable during the life cycle of a sandboxed_path
+ *
+ * Caveat:
+ *  - relatives paths are not possible
```
simonsasse, line 33:
```suggestion
 *  - relative paths are not possible
```

test/include/seqan3/test/sandboxed_path.hpp
```diff
@@ -0,0 +1,334 @@
...
+ *  - sandboxed_path is not 100% tight
```
simonsasse, line 34: What is a 100% tight path vs. a not 100% tight path ? :D

test/include/seqan3/test/sandboxed_path.hpp
```diff
@@ -0,0 +1,334 @@
...
+{
+private:
+    std::filesystem::path const sandbox_path;
+
+public:
+    /*!\brief Construction of a sandboxed_path
+     * \param path must be an absolute path.
+     *
+     * A sandboxed_path initialized with this constructor will
+     * point to path and disallow extension that leave path.
```
simonsasse, line 47: maybe "disallow an extension that leaves the path" or "disallow extensions that leave the  path" is a bit clearer.

test/include/seqan3/test/sandboxed_path.hpp
```diff
@@ -0,0 +1,334 @@
...
+
+    sandboxed_path() = delete; //!< Default construction not possible
+    sandboxed_path(sandboxed_path const&) = default;    //!< Default copy constructor
+    sandboxed_path(sandboxed_path&&) noexcept = default; //!< Default move constructor
+
+
+private:
+    void normalize()
+    {
+        auto concated_path   = sandbox_path / std::filesystem::path{*this};
```
simonsasse, line 79:
```suggestion
        auto concatenated_path   = sandbox_path / std::filesystem::path{*this};
```

test/include/seqan3/test/sandboxed_path.hpp
```diff
@@ -0,0 +1,334 @@
...
+        auto normalized_path = concated_path.lexically_normal();
```
simonsasse, line 80: same here :)

test/include/seqan3/test/sandboxed_path.hpp
```diff
@@ -0,0 +1,334 @@
...
+    {
+        auto parent_path = std::filesystem::path::parent_path();
+        return sandboxed_path{sandbox_path, parent_path};
+    }
+
+
+    void swap(sandboxed_path &) noexcept = delete; //!< Swap operator is not possible
+
+private:
+    void clear() = delete; //!< Unlikly operator
```
simonsasse, line 321:
```suggestion
    void clear() = delete; //!< Unlikely operator
```

simonsasse (changes requested):
Just some typos. Everything else looks good to me :)

SGSSGene:
@simonsasse Thanks for your feedback.
I have corrected all the typos. I also clarified a few more than you requested. It would be great if you could take another look.
Also, I changed a "detail" on how the `sandboxed_path` works. Now it has to store absolute paths, otherwise it is way to easy
to break out of the sandbox.

The gcc7 is failing, I have no solution for it yet. I am still thinking about it. But I guess what I have to do is to implement
`weakly_canonical` and `lexically_relative` for gcc7.
Maybe it is possible to have a less strict version of `sandboxed_path` in gcc7, with less good implementations of the missing functions. (or leaving them out all together)

test/include/seqan3/test/sandboxed_path.hpp
```diff
@@ -0,0 +1,344 @@
...
+    sandboxed_path() = delete; //!< Default construction not possible
+    sandboxed_path(sandboxed_path const&) = default;    //!< Default copy constructor
+    sandboxed_path(sandboxed_path&&) noexcept = default; //!< Default move constructor
+
+
+private:
+    /*!\brief Normalizes the path.
+     *
+     * Normalization means that is converted to a absolute path and
+     * the path is lexically normalized as described in std::filesystem::path
```
simonsasse, line 80:
```suggestion
     * Normalization means that the path is converted to an absolute path and
     * lexically normalized as described in std::filesystem::path
```

simonsasse (changes requested):
@SGSSGene Just a tiny thing. :)
It looks like your test fails not only for gcc7 but also for the other versions now.

SGSSGene:
Resolution: don't check for invariant for gcc7

test/include/seqan3/test/sandboxed_path.hpp
```diff
@@ -0,0 +1,416 @@
...
+
+/** Free sandboxed_path append operator.
+ *  This work the same as std::filesystem::operator/(std::filesystem::path&)
+ *  and additionally checks the invariant.
+ */
+template <typename Rhs>
+sandboxed_path operator/(sandboxed_path lhs, Rhs const& rhs) {
+    lhs /= rhs;
+    return lhs;
+}
```
eseiler, line 414: You should be able to remove this.

std::filesystem::operator/(std::filesystem::path&) works via ADL and ADL also looks for base classes.

test/include/seqan3/test/sandboxed_path.hpp
```diff
@@ -0,0 +1,416 @@
...
+    {
+        auto parent_path = std::filesystem::path::parent_path();
+        return sandboxed_path{sandbox_path, parent_path};
+    }
+
+
+    void swap(sandboxed_path &) noexcept = delete; //!< Swap operator is not possible
+
+private:
+    void clear() = delete; //!< Unlikely to be usefull, not implemented.
```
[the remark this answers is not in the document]
SGSSGene: I changed it to public. The issue is that the invariant requires that the current path is an absolute path. calling `clear()` will always violate the invariant.

test/include/seqan3/test/sandboxed_path.hpp
```diff
@@ -0,0 +1,416 @@
...
+     * Basic exception guarantee
+     */
+    sandboxed_path parent_path() const
+    {
+        auto parent_path = std::filesystem::path::parent_path();
+        return sandboxed_path{sandbox_path, parent_path};
+    }
+
+
+    void swap(sandboxed_path &) noexcept = delete; //!< Swap operator is not possible
```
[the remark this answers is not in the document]
SGSSGene: true, should work :-)

simonsasse (approved):
Looks good 👍

test/include/seqan3/test/sandboxed_path.hpp
```diff
@@ -0,0 +1,461 @@
...
+     * Basic exception guarantee
+     */
+    template <typename source_t>
+    sandboxed_path & assign(source_t const & source)
+    {
+        std::filesystem::path::assign(source);
+        normalise();
+        checkInvariant();
+        return *this;
+    }
```
eseiler, line 245: Maybe you can think of another name than `source` and `source_t`?
It's from the standard, but it would be probably more clear if we used `path_t` and `new_path` or something like this?

test/include/seqan3/test/sandboxed_path.hpp
```diff
@@ -0,0 +1,461 @@
...
+        if (rel_path.string().find("..") == 0) {
+            throw std::filesystem::filesystem_error("Leaving temporary directory is not allowed!",
+                                                sandbox_path, *this,
+                                                std::make_error_code(std::errc::invalid_argument));
+        }
+#endif
+    }
+
+    /*!\brief Replaces the path with a new path.
+     * \tparam source_t The type of the source.
```
eseiler, line 229:
```suggestion
     * \tparam source_t The type of the new path.
```

eseiler: `\param` is missing

test/include/seqan3/test/sandboxed_path.hpp
```diff
@@ -0,0 +1,461 @@
...
+
+/** Free sandboxed_path append operator.
+ *  This work the same as std::filesystem::operator/(std::filesystem::path&)
+ *  and additionally checks the invariant.
+ */
+template <typename Rhs>
+sandboxed_path operator/(sandboxed_path lhs, Rhs const& rhs) {
+    lhs /= rhs;
+    return lhs;
+}
```
eseiler, line 459:
```suggestion
/*!\brief Append a path to a seqan3::test::sanboxed_path.
 * \tparam path_t The type of the path to append.
 * \param lhs The seqan3::test::sandboxed_path.
 * \param rhs The path to append.
 * \relates seqan3::test::sandboxed_path
 *
 *  This work the same as std::filesystem::operator/(std::filesystem::path&)
 *  and additionally checks the invariant.
 */
template <typename path_t>
sandboxed_path operator/(sandboxed_path lhs, path_t const & rhs)
{
    lhs /= rhs;
    return lhs;
}
```

eseiler (changes requested):
Our maximum line length is 120 (also in documentation), so feel free to reflow where you think it would help.

We do not build (yet?) documentation for `test/include`, so there are no checks for the documentation. If you still want to check (but this is just if you are curious), you could add the file to INPUT and INCLUDE_PATH: https://github.com/seqan/seqan3/blob/dfdccb460da99d1be8610fce98c28824c1f5b554/test/documentation/seqan3_doxygen_cfg.in#L9

test/include/seqan3/test/sandboxed_path.hpp
```diff
@@ -0,0 +1,461 @@
...
+     */
+    explicit sandboxed_path(std::filesystem::path sandbox_path, std::filesystem::path path)
+        : std::filesystem::path {path}
+        , sandbox_path             {std::move(sandbox_path)}
+    {
+        normalise();
+        checkInvariant();
+    }
+
+    sandboxed_path() = delete; //!< Deleted.
```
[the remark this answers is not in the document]
SGSSGene: Can't be done. The sandbox_directory path would be "", which is invalid since we need to have an absolute path.

test/include/seqan3/test/sandboxed_path.hpp
```diff
@@ -0,0 +1,461 @@
...
+                {
+                    path_parts.emplace_back(word);
+                }
+            }
+            current_pos = end_pos;
+        }
+        std::string normalised_path;
+        for (auto const& p : path_parts) {
+            normalised_path += "/" + p;
+        }
```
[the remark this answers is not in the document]
SGSSGene: I don't want a view as a result, but an actuall string. It could be done with std::accumulate. But that seems quit anoying:
```
auto normalised_path = std::accumulate(begin(path_parts), end(path_parts), std::string{""}, [](auto prev, auto const& next)
{
    return prev + '/' + next;
});
```
Yes it is one line shorter, but one of the lines is way longer ;-) Not sure if it is what we want.

test/include/seqan3/test/sandboxed_path.hpp
```diff
@@ -0,0 +1,461 @@
...
+            if (is_relative()) {
+                return (sandbox_path / *this).string();
+            } else {
+                return string();
+            }
+        }();
+        // Now we need to manually collapse any "." and ".."
+        size_t current_pos = path.find("/", 0); // find first "/" character
+        std::vector<std::string> path_parts;
+        while (current_pos < path.size()) {
```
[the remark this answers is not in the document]
SGSSGene: I am not sure how this works. I tried:
```
#include <seqan3/std/ranges>
....
for (auto word : path_string | std::views::spit('/')) {
// unclear what to do with word
}
```
`word` is of type `ranges::detail::split_outer_iterator<ranges::split_view<ranges::ref_view<const std::__cxx11::basic_string<char> >, ranges::single_view<char> >, true>`
The documentation is to spare for me to make anything useful out of it. Any hints?

test/include/seqan3/test/sandboxed_path.hpp
```diff
@@ -0,0 +1,461 @@
...
+ *
+ *  Invariant:
+ *  - sandboxed_path is always converted to an absolute path
+ *  - sandboxed_path always points to a file inside a given sandbox directory.
+ *  - sandbox directory is immutable during the life cycle of a sandboxed_path
+ *
+ * Caveat:
+ *  - relative paths are not possible
+ *  - some functions will leave the sandboxed environment.
+ *    - e.g.: calling relative_path() leaves the environment of sandboxed_path.
```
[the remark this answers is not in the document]
SGSSGene: `relative_path()` takes a `std::filesystem::path` because of slicing it will accept a `sandboxed_path` and returns a `std::filesystem::path`. This returned `std::filesystem::path` is of course outside of the handle of `sandboxed_path`. No gurantees we can make. We could offer our own `relative_path()` function, but it doesn't really make sense, since one of the constraints of `sandboxed_path` is that we operate on absolute paths (to avoid other issues).
All in all it would take a lot more work to get a 100% tight solution while being a drop in replacement for std::filesystem::path and easy to use.

test/include/seqan3/test/sandboxed_path.hpp
```diff
@@ -0,0 +1,562 @@
...
+        : std::filesystem::path {std::move(path)}
+        , sandbox_directory     {std::move(sandbox_directory)}
+    {
+        normalise();
+        check_invariant();
+    }
+
+    sandboxed_path() = delete;                           //!< Deleted.
+    sandboxed_path(sandboxed_path const&) = default;     //!< Defaulted.
+    sandboxed_path(sandboxed_path&&) noexcept = default; //!< Defaulted.
```
eseiler, line 77: Copy and move assignment missing

test/include/seqan3/test/sandboxed_path.hpp
```diff
@@ -0,0 +1,562 @@
...
+     */
+    explicit sandboxed_path(std::filesystem::path sandbox_directory, std::filesystem::path path)
+        : std::filesystem::path {std::move(path)}
+        , sandbox_directory     {std::move(sandbox_directory)}
+    {
+        normalise();
+        check_invariant();
+    }
+
+    sandboxed_path() = delete;                           //!< Deleted.
```
eseiler, line 75: Why deleted?
I would only delete the default constructor if constructed object is ill-formed, if it just doesn't make sense to default construct it, I would still leave it.

test/include/seqan3/test/sandboxed_path.hpp
```diff
@@ -0,0 +1,562 @@
...
+        while (current_pos < str.size())
+        {
+            auto end_pos = std::min(str.find(delim, current_pos + 1), str.size());
+            auto begin   = str.data() + current_pos + 1;
+            auto length  = end_pos - current_pos - 1;
+            result.emplace_back(begin, length);
+            current_pos = end_pos;
+        }
+        return result;
+    }
```
eseiler, line 152:
```suggestion
    static std::vector<std::string_view> split_string(std::string const & str, char const delim)
    {
        std::vector<std::string_view> result{};
        // If string is empty, nothing to split
        if (str.empty())
        {
            return result;
        }

        size_t current_pos = str.find(delim, 0);

        // If string doesn't contain any `delim` return view to the string
        if (current_pos == std::string::npos)
        {
            result.emplace_back(str);
            return result;
        }

        while (current_pos < str.size())
        {
            auto end_pos = std::min(str.find(delim, current_pos + 1), str.size());
            auto begin   = str.data() + current_pos + 1;
            auto length  = end_pos - current_pos - 1;
            result.emplace_back(begin, length);
            current_pos = end_pos;
        }
        return result;
    }
```

test/include/seqan3/test/sandboxed_path.hpp
```diff
@@ -0,0 +1,562 @@
...
+#if SEQAN3_WORKAROUND_GCC_INCOMPLETE_FILESYSTEM
+    /*!\brief Splits a string into string_views
+     * \param str String that is going to be split
+     * \param delim Delimiter at which point the string should be split.
+     */
+    static std::vector<std::string_view> split_string(std::string const& str, char delim) {
+        std::vector<std::string_view> result{};
+        // If string is empty, nothing to split
+        if (str.empty()) {
+            return result;
```
eseiler, line 132: coverage complains

test/include/seqan3/test/sandboxed_path.hpp
```diff
@@ -0,0 +1,562 @@
...
+            }
+            // If the path element is ".." we have to remove the last added
+            // path element.
+            else if (word == "..")
+            {
+                if (path_parts.empty())
+                {
+                    throw std::filesystem::filesystem_error("Path can not be normalised",
+                                                            *this,
+                                                            std::make_error_code(std::errc::invalid_argument));
```
eseiler, line 199: coverage complains

test/include/seqan3/test/sandboxed_path.hpp
```diff
@@ -0,0 +1,562 @@
...
+     *
+     * Basic exception guarantee
+     */
+    template <typename input_iter_t>
+    sandboxed_path & append(input_iter_t first, input_iter_t second)
+    {
+        std::filesystem::path::append(first, second);
+        normalise();
+        check_invariant();
+        return *this;
```
eseiler, line 397: coverage complains

eseiler (changes requested):
💅 
Can you check codecov?

test/include/seqan3/test/sandboxed_path.hpp
SGSSGene: It is ill-formed. It would always throw an exception inside the constructor.

test/include/seqan3/test/sandboxed_path.hpp
SGSSGene: They are directly below the constructors, maybe not obviously but they are covered by the template copy/move assignments.
```
    template <typename path_t>
    sandboxed_path & operator=(path_t const & new_path)
...
```

eseiler: We still add the non-template versions.

test/include/seqan3/test/sandboxed_path.hpp
eseiler: Where would it throw?

`sandbox_path() = default;` should default construct the `sandbox_directory` and this shouldn't throw.

It will be nonsense since you cannot change the directory, but default-constructability is a nice trait to have.

SGSSGene: I can't do `sandbox_path() = default`. I have to do at least:
```
sandbox_path() {
check_invariant();
}
```
and the `check_invariant()` would through.

I could of course say `sandbox_path() : sandbox_path('/') {}` and construct it to always be the root directory. But that seems rather suprising.

test/include/seqan3/test/sandboxed_path.hpp
SGSSGene: Ah, ok, than I will add them :-)

test/include/seqan3/test/sandboxed_path.hpp
```diff
@@ -0,0 +1,600 @@
...
+     * ### Exceptions
+     *
+     * Basic exception guarantee
+     */
+    sandboxed_path & operator=(sandboxed_path && new_path)
+    {
+        std::filesystem::path::operator=(std::move(new_path));
+        check_invariant();
+        return *this;
+    }
```
eseiler, line 112: Wouldn't `=default;` be ok here because the `new_path` must have had checked its invariant at some point?

eseiler (changes requested):
One last thing :)

test/include/seqan3/test/sandboxed_path.hpp
SGSSGene: This would only be possible if `this->sandbox_directory == new_path.sandbox_directory` was true. But `new_path` can have a path to a different sandbox. The way I designed this class is that `sandbox_directory` is unchangeable, even under a move.

eseiler: Ah, right :)

eseiler (approved):

PR #2372, opened 2021-02-09
alphabet: first things that are marked stable / experimental.

marehr:
This was decided in the Core-Meeting 2021-02-10

### DIFF
--- a/include/seqan3/alphabet/alphabet_base.hpp
+++ b/include/seqan3/alphabet/alphabet_base.hpp
@@ -48,6 +48,7 @@ namespace seqan3
  * This creates an alphabet called `ab` which has size two and the two letters 'A' and 'B':
  * \include test/snippet/alphabet/detail/alphabet_base.cpp
  *
+ * \stableapi{Since version 3.1.}
  */
 template <typename derived_type, size_t size, typename char_t = char>
 class alphabet_base
@@ -58,9 +59,18 @@ protected:
     /*!\name Member types
      * \{
      */
-    //!\brief The char representation; conditional needed to make semi alphabet definitions legal.
+    /*!\brief The char representation; conditional needed to make semi alphabet definitions legal.
+     * \details We need a return type for seqan3::alphabet_base::to_char and seqan3::alphabet_base::assign_char other
+     *          than void to make these in-class definitions valid when `char_t` is void.
+     *
+     * \stableapi{Since version 3.1.}
+     */
     using char_type = std::conditional_t<std::same_as<char_t, void>, char, char_t>;
-    //!\brief The type of the alphabet when represented as a number (e.g. via to_rank()).
+
+    /*!\brief The type of the alphabet when represented as a number (e.g. via to_rank()).
+     *
+     * \stableapi{Since version 3.1.}
+     */
     using rank_type = detail::min_viable_uint_t<size - 1>;
     //!\}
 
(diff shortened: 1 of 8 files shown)


include/seqan3/alphabet/alphabet_base.hpp
```diff
@@ -58,9 +59,15 @@ class alphabet_base
     /*!\name Member types
      * \{
      */
-    //!\brief The char representation; conditional needed to make semi alphabet definitions legal.
+    /*!\brief The char representation; conditional needed to make semi alphabet definitions legal.
```
joergi-w, line 62: `char_type` is for Alphabet concept, not Semialphabet...
```suggestion
    /*!\brief The char representation; conditional needed to make alphabet definitions legal.
```

joergi-w (changes requested):
There is one issue where I would like you to look at:

include/seqan3/alphabet/alphabet_base.hpp
eseiler: This is open

eseiler (approved):
Please address the point raised by @joergi-w 
Otherwise LGTM

include/seqan3/alphabet/alphabet_base.hpp
marehr: I added

```
      * \details We need a return type for seqan3::alphabet_base::to_char and seqan3::alphabet_base::assign_char other
      *          than void to make these in-class definitions valid when `char_t` is void.
 ```

I hope this makes it more clear :)

Comment (joergi-w) [reply include/seqan3/alphabet/alphabet_base.hpp:62]: I think I understand now: The char type is void for semi alphabets, which makes a conditional necessary. I agree and give green light. 
But what still confuses me: this is alphabet base, not semi alphabet base, so why is `void` allowed at all? And wouldn't it be better to `enable_if` the char type only if it is not void? (The same with assign_char and to_char...) Because we don't want that e.g. `semialphabet::to_char` exists and returns `char`...

**Review by joergi-w — APPROVED** [review]
I feel that my issue is out of scope for this PR, but I still find a little confusion here:

File: include/seqan3/alphabet/alphabet_base.hpp (hunk 1 above)
Comment (marehr) [reply include/seqan3/alphabet/alphabet_base.hpp:62]: We don't have a dedicated semi_alphabet_base, our alphabet_base does both cases. If char_t is void it is a semi-alphabet, otherwise it is an alphabet.

> And wouldn't it be better to `enable_if` the char type only if it is not void?

We do a requires on to_char and assign_char if the value is void or not. So to_char is not available if the alphabet_base is a semi-alphabet.

The Problem is more technically, because 

```c++
char_type foo() requires char_type != void
{
   return something;
}
```

has a return statement and if the return type (= char_type) is void, the function definition is invalid and will throw a compiler error, but we want a SFINAE error here, but for that the function definition must be syntactically valid. The crux here is that the return type could be any type, even `void *`, because `foo` will never be considered, because the requires will be false in these cases. 

We could also define it as 

```c++
auto foo() requires char_type != void
{
   return something;
}
```

but we would lose the documentation.

Or define it old-school like:

```c++
std::enable_if_t<char_type != void, char_t> foo() 
{
   return something;
}
```

Another way would be to split out current `alphabet_base` into a `alphabet_base` and a `semi_alphabet_base`, but we have some alphabets (primarily proxies) that can be either conditionally. So we need at some point a mechanism to be either conditionally.

I'm also un-happy that we have a char_type that is not always the same as char_t, but there is no easy way to conditionally not define a type.

joergi-w:
Thank you :)

PR #2379, opened 2021-02-14
[MISC] deprecate seqan3::field::seq_qual

SGSSGene:
- adds [DEPRECATED] in many places.
- erased mentioning of seq_qual everywhere I could find
- Wrap all usages of seqan3::field::seq_qual
- referenced cookbook to usage of field::seq and field::qual
- referenced cookbook in changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,10 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
   to add convenient functions that compute information based on the record itself and to provide better documentation.
   ([\#2340](https://github.com/seqan/seqan3/pull/2340), [\#2380](https://github.com/seqan/seqan3/pull/2380),
   [\#2389](https://github.com/seqan/seqan3/pull/2389))
+* Deprecated `seqan3::field::seq_qual`. Use `seqan3::fiel::seq` and `seqan3::field::qual` instead.
+  ([\#2379](https://github.com/seqan/seqan3/pull/2379)). Check out 
+  [SeqAn3 Cookbook - Write Record](https://docs.seqan.de/seqan/3.0.3/cookbook.html)
+  for usage.
 
 #### Search
 
--- a/include/seqan3/io/record.hpp
+++ b/include/seqan3/io/record.hpp
@@ -64,7 +64,8 @@ enum class field
     seq,            //!< The "sequence", usually a range of nucleotides or amino acids.
     id,             //!< The identifier, usually a string.
     qual,           //!< The qualities, usually in Phred score notation.
-    seq_qual,       //!< Sequence and qualities combined in one range.
+    _seq_qual_deprecated, //!< [DEPRECATED] Sequence and qualities combined in one range. Use field::seq and field::qual instead.
+
     offset,         //!< Sequence (seqan3::field::seq) relative start position (0-based), unsigned value.
 
     // Fields unique to structure io ...........................................
@@ -104,11 +105,14 @@ enum class field
     user_defined_8, //!< Identifier for user defined file formats and specialisations.
     user_defined_9, //!< Identifier for user defined file formats and specialisations.
 
+    // deprecated lowercase:
+    seq_qual SEQAN3_DEPRECATED_310 = _seq_qual_deprecated, //!< [DEPRECATED] Sequence and qualities combined in one range. Use field::seq and field::qual instead.
+
     // deprecated uppercase:
     SEQ SEQAN3_DEPRECATED_310 = seq, //!< Please use the field name in lower case.
     ID SEQAN3_DEPRECATED_310 = id, //!< Please use the field name in lower case.
     QUAL SEQAN3_DEPRECATED_310 = qual, //!< Please use the field name in lower case.
-    SEQ_QUAL SEQAN3_DEPRECATED_310 = seq_qual, //!< Please use the field name in lower case.
+    SEQ_QUAL SEQAN3_DEPRECATED_310 = _seq_qual_deprecated, //!< [DEPRECATED] Sequence and qualities combined in one range. Use field::seq and field::qual instead.
     OFFSET SEQAN3_DEPRECATED_310 = offset, //!< Please use the field name in lower case.
     BPP SEQAN3_DEPRECATED_310 = bpp, //!< Please use the field name in lower case.
     STRUCTURE SEQAN3_DEPRECATED_310 = structure, //!< Please use the field name in lower case.
--- a/include/seqan3/io/sequence_file/input.hpp
+++ b/include/seqan3/io/sequence_file/input.hpp
@@ -57,10 +57,6 @@ namespace seqan3
  *
  * \details
  *
- * Note that the alphabet type of the seqan3::field::seq_qual cannot be specified directly, it is always
- * seqan3::qualified<sequence_alphabet, quality_alphabet> and the container type templates for
- * the field are those of seqan3::field::seq.
- *
  * \{
  */
 /*!\typedef using sequence_alphabet
@@ -202,12 +198,11 @@ struct sequence_file_input_default_traits_aa : sequence_file_input_default_trait
  * FastA and FastQ, but some may also be interested in treating SAM or BAM files as sequence
  * files, discarding the alignment.
  *
- * The Sequence file abstraction supports reading four different fields:
+ * The Sequence file abstraction supports reading three different fields:
  *
  *   1. seqan3::field::seq
  *   2. seqan3::field::id
  *   3. seqan3::field::qual
- *   4. seqan3::field::seq_qual (sequence and qualities in one range)
  *
  * The first three fields are retrieved by default (and in that order). The last field may be selected to have
  * sequence and qualities directly stored in a more memory-efficient combined container. If you select the last
@@ -330,10 +325,10 @@ public:
     using stream_char_type      = char;
     //!\}
 
-    /*!\brief The subset of seqan3::field IDs that are valid for this file; order corresponds to the types in
+   /*!\brief The subset of seqan3::field IDs that are valid for this file; order corresponds to the types in
      * \ref field_types.
      */
-    using field_ids            = fields<field::seq, field::id, field::qual, field::seq_qual>;
+    using field_ids            = fields<field::seq, field::id, field::qual, field::_seq_qual_deprecated>;
 
     static_assert([] () constexpr
                   {
@@ -347,7 +342,7 @@ public:
 
     static_assert([] () constexpr
                   {
-                      return !(selected_field_ids::contains(field::seq_qual) &&
+                      return !(selected_field_ids::contains(field::_seq_qual_deprecated) &&
                                (selected_field_ids::contains(field::seq) ||
                                (selected_field_ids::contains(field::qual))));
                   }(),
@@ -367,10 +362,14 @@ public:
     //!\brief The type of field::qual (std::vector <seqan3::phred42> by default).
     using quality_type          = typename traits_type::template quality_container<
                                     typename traits_type::quality_alphabet>;
-    //!\brief The type of field::seq_qual (std::vector <seqan3::dna5q> by default).
+#ifdef SEQAN3_DEPRECATED_310
+    /*!\brief [DEPRECATED] The type of field::seq_qual (std::vector <seqan3::dna5q> by default).
+     * \deprecated This type will be removed in SeqAn-3.1.0. Use sequence_type and quality_type instead.
+     */
     using sequence_quality_type = typename traits_type::
                                     template sequence_container<qualified<typename traits_type::sequence_alphabet,
                                                                           typename traits_type::quality_alphabet>>;
+#endif // SEQAN3_DEPRECATED_310
 
     //!\brief The previously defined types aggregated in a seqan3::type_list.
     using field_types           = type_list<sequence_type, id_type, quality_type, sequence_quality_type>;
@@ -591,7 +590,7 @@ public:
 
     //!\brief The options are public and its members can be set directly.
     sequence_file_input_options<typename traits_type::sequence_legal_alphabet,
-                             selected_field_ids::contains(field::seq_qual)> options;
+                             selected_field_ids::contains(field::_seq_qual_deprecated)> options;
 
 protected:
     //!\privatesection
@@ -650,13 +649,13 @@ protected:
         std::visit([&] (auto & f)
         {
             // read new record
-            if constexpr (selected_field_ids::contains(field::seq_qual))
+            if constexpr (selected_field_ids::contains(field::_seq_qual_deprecated))
             {
                 f.read_sequence_record(*secondary_stream,
                                        options,
-                                       detail::get_or_ignore<field::seq_qual>(record_buffer),
+                                       detail::get_or_ignore<field::_seq_qual_deprecated>(record_buffer),
                                        detail::get_or_ignore<field::id>(record_buffer),
-                                       detail::get_or_ignore<field::seq_qual>(record_buffer));
+                                       detail::get_or_ignore<field::_seq_qual_deprecated>(record_buffer));
             }
             else
             {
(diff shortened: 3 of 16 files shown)


eaasna (commented):
Nice work! 

I'm just  wondering why it was decided you should avoid deprecation warnings instead of using the alternative of `seqan3::views::zip` and `std::views::transform`. Or is that a next PR?

SGSSGene:
> Nice work!
> 
> I'm just wondering why it was decided you should avoid deprecation warnings instead of using the alternative of `seqan3::views::zip` and `std::views::transform`. Or is that a next PR?

I am not sure what you mean by this.

I havn't planned another PR :-)

include/seqan3/io/sequence_file/input.hpp
```diff
@@ -589,9 +592,12 @@ class sequence_file_input
     }
     //!\}
 
+#pragma GCC diagnostic push // avoid deprecation warning of field::seq_qual
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
     //!\brief The options are public and its members can be set directly.
     sequence_file_input_options<typename traits_type::sequence_legal_alphabet,
                              selected_field_ids::contains(field::seq_qual)> options;
```
eaasna, line 599: Here for example the deprecated `field:seq_qual` is being used. It seems to me that it would make more sense to remove the usage and replace it with an alternative (since you are deprecating `field:seq_qual`). But you wrap it to avoid deprecation warnings instead.

And I noticed that the PR description

SGSSGene: Thx! Yes, this doesn't seem correct. I will look into it

eaasna: Okay :) what you did does match the issue description but I also found it a bit strange

include/seqan3/io/record.hpp
```diff
@@ -64,7 +66,7 @@ enum class field
     seq,            //!< The "sequence", usually a range of nucleotides or amino acids.
     id,             //!< The identifier, usually a string.
     qual,           //!< The qualities, usually in Phred score notation.
-    seq_qual,       //!< Sequence and qualities combined in one range.
+    seq_qual SEQAN3_DEPRECATED_310, //!< [DEPRECATED] Sequence and qualities combined in one range.
```
marehr, line 69: I'm wondering whether

```suggestion
    _seq_qual_deprecated, //!< [DEPRECATED] Sequence and qualities combined in one range.
    seq_qual SEQAN3_DEPRECATED_310 = _seq_qual_deprecated, //!< [DEPRECATED] Sequence and qualities combined in one range.
```

include/seqan3/io/record.hpp
```diff
@@ -108,7 +110,7 @@ enum class field
     SEQ SEQAN3_DEPRECATED_310 = seq, //!< Please use the field name in lower case.
     ID SEQAN3_DEPRECATED_310 = id, //!< Please use the field name in lower case.
     QUAL SEQAN3_DEPRECATED_310 = qual, //!< Please use the field name in lower case.
-    SEQ_QUAL SEQAN3_DEPRECATED_310 = seq_qual, //!< Please use the field name in lower case.
+    SEQ_QUAL SEQAN3_DEPRECATED_310 = seq_qual, //!< [DEPRECATED] Please use the field name in lower case.
```
marehr, line 113: and

```suggestion
    SEQ_QUAL SEQAN3_DEPRECATED_310 = _seq_qual_deprecated, //!< [DEPRECATED] Please use the field name in lower case.
```

will enable us to remove the surrondung `#pragma GCC diagnostic ignored "-Wdeprecated-declarations"`

SGSSGene: I believe this is a matter of style and taste. To me both solution are valid solution.
I think that the `pragma` solution is very spot on, but the `_seq_qal_deprecated` is very compact.
You decide :-)

marehr: True, I've made the experience that opening  `#pragma GCC diagnostic ...` can hide errors if you span it around complete classes / structs. So I would prefer a more local solution. Can you change it? :)

include/seqan3/io/sequence_file/input.hpp
```diff
@@ -330,10 +325,13 @@ class sequence_file_input
     using stream_char_type      = char;
     //!\}
 
-    /*!\brief The subset of seqan3::field IDs that are valid for this file; order corresponds to the types in
+#pragma GCC diagnostic push // avoid deprecation warning of field::seq_qual
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+   /*!\brief The subset of seqan3::field IDs that are valid for this file; order corresponds to the types in
      * \ref field_types.
      */
     using field_ids            = fields<field::seq, field::id, field::qual, field::seq_qual>;
```
marehr, line 333:
```suggestion
    using field_ids            = fields<field::seq, field::id, field::qual, field::_seq_qal_deprecated>;
```

We can now also use without adding more `#pragma GCC diagnostic push`

SGSSGene: Yes, that is much prettier! :+1:

include/seqan3/io/record.hpp
SGSSGene: Done :-)

include/seqan3/io/sequence_file/input.hpp
```diff
@@ -650,13 +647,14 @@ class sequence_file_input
         std::visit([&] (auto & f)
         {
             // read new record
-            if constexpr (selected_field_ids::contains(field::seq_qual))
+            //
```
marehr, line 650:
```suggestion
            // read new record
```

include/seqan3/io/sequence_file/input.hpp
```diff
@@ -367,7 +362,9 @@ class sequence_file_input
     //!\brief The type of field::qual (std::vector <seqan3::phred42> by default).
     using quality_type          = typename traits_type::template quality_container<
                                     typename traits_type::quality_alphabet>;
-    //!\brief The type of field::seq_qual (std::vector <seqan3::dna5q> by default).
+    /*!\brief [DEPRECATED] The type of field::seq_qual (std::vector <seqan3::dna5q> by default).
+     * \deprecated This type will be removed in SeqAn-3.1.0. Use sequence_type and quality_type instead.
+     */
     using sequence_quality_type = typename traits_type::
                                     template sequence_container<qualified<typename traits_type::sequence_alphabet,
                                                                           typename traits_type::quality_alphabet>>;
```
marehr, line 370:
```suggestion
#ifdef SEQAN3_DEPRECATED_310
    /*!\brief [DEPRECATED] The type of field::seq_qual (std::vector <seqan3::dna5q> by default).
     * \deprecated This type will be removed in SeqAn-3.1.0. Use sequence_type and quality_type instead.
     */
    using sequence_quality_type = typename traits_type::
                                    template sequence_container<qualified<typename traits_type::sequence_alphabet,
                                                                          typename traits_type::quality_alphabet>>;
#endif // SEQAN3_DEPRECATED_310
```

include/seqan3/io/sequence_file/input_options.hpp
```diff
@@ -18,8 +18,8 @@ namespace seqan3
 {
 /*!\brief The options type defines various option members that influence the behaviour of all or some formats.
  * \tparam sequence_legal_alphabet_ The sequence legal alphabet exposed as type trait to the format.
- * \tparam seq_qual_combined Trait that exposes to the format whether seq and qual arguments are actually the
- * same/combined.
+ * \tparam seq_qual_combined [DEPRECATED] Trait that exposes to the format whether seq and qual arguments are
+ * actually the same/combined. Will be removed in SeqAn-3.1.0.
  */
 template <typename sequence_legal_alphabet, bool seq_qual_combined>
```
marehr, line 24:
```suggestion
#ifdef SEQAN3_DEPRECATED_310
template <typename sequence_legal_alphabet, bool seq_qual_combined>
#else // ^^^ before seqan 3.1 / after seqan 3.1 vvv
template <typename sequence_legal_alphabet>
#endif // SEQAN3_DEPRECATED_310
```

marehr:
Hmm something went wrong with your rebase, can you do it again?

marehr (commented):
small review, I will do the rest after the rebase was successful. I stopped for now, because I had some trouble to see what you or someone else changed.

test/snippet/io/sequence_file/sequence_file_output_fields_trait_1.cpp
```diff
@@ -25,7 +25,7 @@ int main()
 
         // ...
 
-        fout.emplace_back(id, seq_qual);       // note also that the order the argumets is now different, because
+        fout.emplace_back(id, seq_qual);       // note also that the order the arguments is now different, because
```
marehr, line 28:
```suggestion
        fout.emplace_back(id, seq_qual, std::get<1>(seq_qual));       // note also that the order of the arguments is now different, because
```

Otherwise, you don't actually write out the quality.

@eseiler can you double-check the grammar 🙈 ?

include/seqan3/io/sequence_file/output.hpp
```diff
@@ -64,12 +64,11 @@ namespace seqan3
...
- * The Sequence file abstraction supports writing four different fields:
+ * The Sequence file abstraction supports writing three different fields:
  *
  *   1. seqan3::field::seq
  *   2. seqan3::field::id
  *   3. seqan3::field::qual
- *   4. seqan3::field::seq_qual (sequence and qualities in one range)
  *
  * The member functions take any and either of these fields. If the field ID of an argument cannot be deduced, it
  * is assumed to correspond to the field ID of the respective template parameter.
```
marehr, line 74: in line 115 

```cpp
/*
 * If you want to pass a combined object for SEQ and QUAL fields to push_back() / emplace_back(), or if you want
 * to change the order of the parameters, you can pass a non-empty fields trait object to the
 * sequence_file_output constructor to select the fields that are used for interpreting the arguments.
 */
```

The first part must be removed.

test/unit/io/sequence_file/sequence_file_format_test_template.hpp
```diff
@@ -122,20 +122,21 @@ TYPED_TEST_P(sequence_file_read, only_id)
...
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
         EXPECT_RANGE_EQ(seqan3::get<seqan3::field::id>(*it), this->ids[i]);
-        EXPECT_RANGE_EQ(seqan3::get<seqan3::field::seq_qual>(*it) | seqan3::views::convert<seqan3::dna5>,
+        EXPECT_RANGE_EQ(seqan3::get<seqan3::field::_seq_qual_deprecated>(*it) | seqan3::views::convert<seqan3::dna5>,
                         this->seqs[i]);
 #pragma GCC diagnostic pop
 
         EXPECT_RANGE_EQ((*it).id(), this->ids[i]);
     }
+
```
marehr, line 139:
```suggestion
```

CHANGELOG.md
```diff
@@ -56,6 +56,9 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
   to add convenient functions that compute information based on the record itself and to provide better documentation.
   ([\#2340](https://github.com/seqan/seqan3/pull/2340), [\#2380](https://github.com/seqan/seqan3/pull/2380),
   [\#2389](https://github.com/seqan/seqan3/pull/2389))
+* Deprecated `seqan3::field::seq_qual`. Use `seqan3::fiel::seq` and `seqan3::field::qual` instead.
+  ([\#2379](https://github.com/seqan/seqan3/pull/2379)). Checkout [SeqAn3 Cookbook - Write Record](https://docs.seqan.de/seqan/3.0.2/cookbook.html)
+  for usage.
```
eseiler, line 61:
```suggestion
* Deprecated `seqan3::field::seq_qual`. Use `seqan3::fiel::seq` and `seqan3::field::qual` instead.
  ([\#2379](https://github.com/seqan/seqan3/pull/2379)). Check out 
  [SeqAn3 Cookbook - Write Record](https://docs.seqan.de/seqan/3.0.2/cookbook.html)
  for usage.
```

Isn't it wrong to link to the `3.0.2` ? Because this is the old way to do it.
I guess the correct way would be to use `3.0.3` and set the server up to redirect to the `master` documentation until `3.0.3` is released?

cc @marehr

include/seqan3/io/sequence_file/input_options.hpp
```diff
@@ -18,10 +18,14 @@ namespace seqan3
 {
 /*!\brief The options type defines various option members that influence the behaviour of all or some formats.
  * \tparam sequence_legal_alphabet_ The sequence legal alphabet exposed as type trait to the format.
- * \tparam seq_qual_combined Trait that exposes to the format whether seq and qual arguments are actually the
- * same/combined.
+ * \tparam seq_qual_combined [DEPRECATED] Trait that exposes to the format whether seq and qual arguments are
+ * actually the same/combined. Will be removed in SeqAn-3.1.0.
```
eseiler, line 22:
```suggestion
 * \tparam seq_qual_combined [DEPRECATED] Trait that exposes to the format whether seq and qual arguments are
 *                           actually the same/combined. Will be removed in SeqAn-3.1.0.
```

CHANGELOG.md
eseiler: I set up a symlink, so we can just use `3.0.3`

test/snippet/io/sequence_file/sequence_file_output_fields_trait_1.cpp
```diff
@@ -15,18 +16,25 @@ int main()
...
 
-        fout.emplace_back(id, seq_qual);       // note also that the order the argumets is now different, because
-        // or:                                    you specified that ID should be first in the fields template argument
-        fout.push_back(std::tie(id, seq_qual));
+        fout.emplace_back(id, view_on_seq, view_on_qual); // Also note that the order of the arguments is now
+                                                          // different, because you specified that ID should be
+                                                          // first in the fields template argument
+        // or:
+        fout.push_back(std::tie(id, view_on_seq, view_on_qual));
     }
```
SGSSGene, line 39: I think this snippet really confusing. So we declare an `seqan3::sequence_file_output` of type `seqan3::format_fasta` which only includes `id` and `sequence`. I explicit declare that I want to have `field::qual` which doesn't really mean anything, since `format_fasta` will drop this information?

The snippet itself talks about how the `order of the arguments is now different` but I actually don't know what its talking about. It is the same order as everywhere else ?!

I don't think I understand what this snippet is supposed to demonstrate. Any thoughts on this
@marehr @eseiler ?

eseiler: This should probably be fastq output and not fasta.

The ordering refers to the default `seq, id, qual` which should be more clear. (2nd snippet in https://docs.seqan.de/seqan/3-master-user/tutorial_sequence_file.html#autotoc_md213 shows the default order).

SGSSGene: Thx, I didn't get that the default order is different. Also fastq makes a lot more sense. I will change it.

test/snippet/io/sequence_file/sequence_file_output_fields_trait_1.cpp
```diff
@@ -7,26 +7,34 @@
...
+        auto view_on_qual = seqan3::views::get<1>(seq_qual);
 
         // ...
 
-        fout.emplace_back(id, seq_qual);       // note also that the order the argumets is now different, because
-        // or:                                    you specified that ID should be first in the fields template argument
-        fout.push_back(std::tie(id, seq_qual));
+        fout.emplace_back(id, view_on_seq, view_on_qual); // Also note that the order of the arguments is now
+                                                          // different, because you specified that ID should be
+                                                          // first in the fields template argument
```
eseiler, line 36: I analysed the comment too long:


Maybe something along the lines of:

```
Note that the order of the arguments is different from the default `seq, id, qual`, because you specified that ID should be first in the fields template argument.
```

* I removed the `Also` because this is the first thing to note.
* I don't like the use of `now` in the sentence because there is no point of reference as to what it was before.
* For the same reason, I added the default order.
* I also added a period at the end :)
* I think this comment could go to the line above `fout.emplace_back` and use the full length...

eaasna (approved):
Looks good, I don't have anything to add :)

marehr (approved):
Thank you!

PR #1196, opened 2019-07-26
Enum support for the argument parser 

smehringer:

PR #2381, opened 2021-02-16
[FIX] parsing containers of enums not working

eseiler:
In https://github.com/seqan/seqan3/pull/1196, we forgot to allow for, e.g. `std::vector<foo::bar>`, where `foo::bar` is an enum.

### DIFF
--- a/include/seqan3/argument_parser/detail/format_parse.hpp
+++ b/include/seqan3/argument_parser/detail/format_parse.hpp
@@ -317,7 +317,7 @@ private:
      *          seqan3::enumeration_names<option_t> map and otherwise seqan3::option_parse_result::success.
      */
     template <named_enumeration option_t>
-    option_parse_result parse_option_value(option_t & value, std::string_view const in)
+    option_parse_result parse_option_value(option_t & value, std::string const & in)
     {
         auto map = seqan3::enumeration_names<option_t>;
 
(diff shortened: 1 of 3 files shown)


Irallia (approved):
Thank you, for the fast help!
LGFM :)

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -339,15 +339,16 @@ class format_parse : public format_base
...
      *
      * \param[out] value The container that stores the parsed value.
      * \param[in] in The input argument to be parsed.
      * \returns A seqan3::option_parse_result whether parsing was successful or not.
      */
     template <sequence_container container_option_t>
     //!\cond
-        requires input_stream_over<std::istringstream, typename container_option_t::value_type>
+        requires input_stream_over<std::istringstream, typename container_option_t::value_type> ||
+                 named_enumeration<typename container_option_t::value_type>
```
marehr, line 351:
```suggestion
        requires requires (std::ranges::value_type<container_option_t> & container_value, std::string const & in)
        {
           { parse_option_value(container_value, in) } -> std::same_as<option_parse_result>;
        }
```

wouldn't that be more general? If it is a container, try to defer to its value_type (that's exactly what the function does). The only problem would be container that have themselves as value_type xD

Since we also have overloads for bool and other things, I guess `std::vector<bool>` does not work right now?

marehr (commented):
Looks good, but can we generalise it ?

include/seqan3/argument_parser/detail/format_parse.hpp
eseiler: Very good catch, `std::vector<bool>` did not work. The concept was hard to write... tldr: you need an instance of the class if you want to test member functions, but the compiler does not necessarily tell you :)

I opted to use `container_option_t::value_type` instead of the ranges thing:
1. Doesn't seem to be currently included (less includes :) )
2. It currently requires `sequence_container`, which requires `value_type` to be defined

If this is good, you can squash

marehr: Very nice!

marehr (approved):

PR #2391, opened 2021-02-17
[FIX] arithmetic_range_validator should take any integer type

marehr:


eseiler:
Can you check the description for `value_list_validator`? It seems very wrong

marehr:
Sorry, I don't completely understand what you mean. Why do you think it is wrong? 

Am 19. Februar 2021 12:20:02 MEZ schrieb Enrico Seiler <notifications@github.com>:
>Can you check the description for `value_list_validator`? It seems very
>wrong
>
>-- 
>https://github.com/seqan/seqan3/pull/2391#issuecomment-782012100

-

test/unit/argument_parser/format_parse_validators_test.cpp
```diff
@@ -75,7 +75,7 @@ TEST(validator_test, fullfill_concept)
     EXPECT_TRUE(seqan3::validator<seqan3::detail::default_validator<int> &>);
 
     EXPECT_TRUE(seqan3::validator<seqan3::detail::default_validator<std::vector<int>>>);
-    EXPECT_TRUE(seqan3::validator<seqan3::arithmetic_range_validator>);
+    EXPECT_TRUE(seqan3::validator<seqan3::arithmetic_range_validator<int>>);
```
SGSSGene, line 78: Shouldn't this also work for `double`?
```suggestion
    EXPECT_TRUE(seqan3::validator<seqan3::arithmetic_range_validator<int>>);
    EXPECT_TRUE(seqan3::validator<seqan3::arithmetic_range_validator<double>>);
```

SGSSGene (changes requested):

SGSSGene (approved):
looks good to me :+1:

eseiler (changes requested):
Generally LGTM, but can you have a look at the documentation I linked?

marehr:
> Generally LGTM, but can you have a look at the documentation I linked?

I did and corrected here. I think `seqan3::value_list_validator` is still not completely correct, since it does two weird things:

* It supports a range AND a parameter pack
* It uses the first option_type in a parameter pack (std::common_type_t<...> would be better)
* It has special cases for strings, which might not be necessary?

include/seqan3/argument_parser/validators.hpp
```diff
@@ -107,6 +107,7 @@ SEQAN3_CONCEPT validator = std::copyable<std::remove_cvref_t<validator_type>> &&
 /*!\brief A validator that checks whether a number is inside a given range.
  * \ingroup argument_parser
  * \implements seqan3::validator
+ * \tparam option_value_t Must be an arithmetic type.
```
eseiler, line 110:
```suggestion
 * \tparam option_value_t The value type of the range; must model seqan3::arithmetic .
```

include/seqan3/argument_parser/validators.hpp
```diff
@@ -172,17 +174,17 @@ class arithmetic_range_validator
 /*!\brief A validator that checks whether a value is inside a list of valid values.
  * \ingroup argument_parser
  * \implements seqan3::validator
- * \tparam option_value_t Must be a (container of) arithmetic type(s).
+ * \tparam option_value_t \copybrief seqan3::option_value_type::option_value_type
```
eseiler, line 177:
```suggestion
 * \tparam option_value_t \copybrief seqan3::value_list_validator::option_value_type
```

probably?

include/seqan3/argument_parser/validators.hpp
```diff
@@ -172,17 +174,17 @@ class arithmetic_range_validator
...
  * On construction, the validator must receive a range or parameter pack of valid values.
  * The class than acts as a functor, that throws a seqan3::validation_error
  * exception whenever a given value is not in the given list.
  *
- * \note In order to simplify the chaining of validators, the option value type is deduced to `double` for ranges whose
- *       value type models seqan3::arithmetic, and to `std::string` if the ranges value type is convertible to it.
- *       Otherwise, the option value type is deduced to the value type of the range.
+ * \note In order to simplify the chaining of validators, the option value type is deduced to `std::string` if the
+ *       ranges value type is convertible to it. Otherwise, the option value type is deduced to the value type of the
+ *       range.
```
eseiler, line 187:
```suggestion
 * \note In order to simplify the chaining of validators, the option value type is deduced to `std::string` if the
 *       range's value type is convertible to it. Otherwise, the option value type is deduced to the value type of the
 *       range.
```

eseiler (changes requested):

eseiler (approved):

Issue #2393, opened 2021-02-18
Argument Parser cannot have default container arguments


eseiler:
## Description

```cpp
#include <seqan3/argument_parser/argument_parser.hpp>
#include <seqan3/core/debug_stream.hpp>

int main()
{
    std::vector<int> option_values{1, 2, 3};

    const char * argv[] = {"./argument_parser_test", "-i", "2", "-i", "1", "-i", "3"};
    seqan3::argument_parser parser{"test_parser", 7, argv, seqan3::update_notifications::off};
    parser.add_option(option_values, 'i', "int-option", "this is an int option.");

    parser.parse();

    seqan3::debug_stream << option_values << '\n';
}
```

### Expected behaviour

Intuitive:
`[2,1,3]`

### Actual behaviour

Unintuitive:
`[1,2,3,2,1,3]`

marehr:
Can you post how the help page looks like?

eseiler:
> Can you post how the help page looks like?

```console
test_parser
===========

OPTIONS

  Basic options:
    -h, --help
          Prints the help page.
    -hh, --advanced-help
          Prints the help page including advanced options.
    --version
          Prints the version information.
    --copyright
          Prints the copyright/license information.
    --export-help (std::string)
          Export the help page information. Value must be one of [html, man].
    -i, --int-option (List of signed 32 bit integer's)
          this is an int option. Default: [1,2,3].

VERSION
    Last update: 
    test_parser version: 
    SeqAn version: 3.0.3
```

PR #2394, opened 2021-02-19
[FIX] default container arguments

eseiler:
Resolves https://github.com/seqan/seqan3/issues/2393

### DIFF
--- a/include/seqan3/argument_parser/detail/format_parse.hpp
+++ b/include/seqan3/argument_parser/detail/format_parse.hpp
@@ -563,6 +563,9 @@ private:
         auto it = find_option_id(argv.begin(), end_of_options_it, id);
         bool seen_at_least_once{it != end_of_options_it};
 
+        if (seen_at_least_once)
+            value.clear();
+
         while (it != end_of_options_it)
         {
             identify_and_retrieve_option_value(value, it, id);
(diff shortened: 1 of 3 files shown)


Irallia (approved):
LGFM, thank you!

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -736,10 +739,12 @@ class format_parse : public format_base
                                     std::to_string(positional_option_calls.size()) +
                                     "). See -h/--help for more information.");
 
-        if (sequence_container<option_type> && !std::is_same_v<option_type, std::string>) // vector/list will be filled with all remaining arguments
+        if constexpr (sequence_container<option_type> && !std::is_same_v<option_type, std::string>) // vector/list will be filled with all remaining arguments
         {
             assert(positional_option_count == positional_option_calls.size()); // checked on set up.
 
+            value.clear();
```
marehr, line 746: I don't quite understand, why we have two different functions to set a value.

I would have expected to only have one point to change this value.

marehr (approved):

include/seqan3/argument_parser/detail/format_parse.hpp
eseiler: It's once for positional options and once for options

PR #2404, opened 2021-02-24
[DOC] new: seqan3::exposition-only namespace

marehr:


SGSSGene (approved):
Description looks nice!

include/seqan3/core/all.hpp
```diff
@@ -45,6 +45,27 @@
...
+  *
+  * but we still want to express a general look-and-feel of the API in our documentation.
+  *
+  * We therefore use the same trick as the C++ standard that defines entities, s.a. concepts, in a partial defined
+  * state to express a general intend of the API without being explicit about it.
+  *
+  * For example https://eel.is/c++draft/iterator.concept.readable where indirectly-readable-impl describes the general
+  * intention of the concept, but does not name it since it is a helper-entity for the std::indirectly_­readable concept.
+  */
+
```
eseiler, line 68:
```suggestion
 /*!\namespace seqan3::exposition_only
  * \brief A namespace for SeqAn entities that are intended for documentation purposes only.
  *
  * \details
  *
  * We try to guarantee a strong \ref api_stability "API-Stability". This has the downside that we sometimes opt to not
  * document entities, because
  *
  * 1. we do not want to guarantee that the entity will always keep the same name or
  * 2. we need a helper entity to define the actual entity or
  * 3. we are not completely sure if the entity is general enough to mark it stable,
  *
  * but we still want to express a general look-and-feel of the API in our documentation.
  *
  * We therefore use the same trick as the C++ standard that defines entities, s.a. concepts, in a partial defined
  * state to express a general intend of the API without being explicit about it.
  *
  * For example, see https://eel.is/c++draft/iterator.concept.readable where `indirectly-readable-impl` describes the
  * general intention of the concept, but does not name it since it is a helper-entity for the std::indirectly_­readable
  * concept.
  */

```

eseiler (changes requested):

eseiler (approved):

PR #2407, opened 2021-02-24
[TEST] rename test/unit/io/{alignment => sam}_file

marehr:
Part of many small rename PRs, tackles https://github.com/seqan/product_backlog/issues/290.


test/unit/io/sam_file/format_bam_test.cpp
```diff
@@ -238,14 +238,14 @@ struct alignment_file_read<seqan3::format_bam> : public alignment_file_data
 // parametrized tests
 // ---------------------------------------------------------------------------------------------------------------------
 
-INSTANTIATE_TYPED_TEST_SUITE_P(bam, alignment_file_read, seqan3::format_bam, );
-INSTANTIATE_TYPED_TEST_SUITE_P(bam, alignment_file_write, seqan3::format_bam, );
+INSTANTIATE_TYPED_TEST_SUITE_P(bam, sam_file_read, seqan3::format_bam, );
+INSTANTIATE_TYPED_TEST_SUITE_P(bam, sam_file_write, seqan3::format_bam, );
```
joergi-w, line 242: Why is there a template class definition for `sam_file_read<seqan3::format_bam>` but not for `sam_file_write<seqan3::format_bam>`?

test/unit/io/sam_file/sam_file_format_test_template.hpp
```diff
@@ -32,9 +32,9 @@ using seqan3::operator""_tag;
 seqan3::alignment_file_input_options<seqan3::dna5> input_options;
 seqan3::alignment_file_output_options output_options;
 
-struct alignment_file_data : public ::testing::Test
+struct sam_file_data : public ::testing::Test
```
joergi-w, line 35: Please document in a comment that this file/class is used for both sam and bam, which is not obvious any more since you changed the name.

joergi-w (changes requested):

test/unit/io/sam_file/format_bam_test.cpp
marehr: I don't know, I just replaced a name.

I guess that output definition is the same between sam and bam, so that it does not need a special specialisation of that class.

test/unit/io/sam_file/sam_file_format_test_template.hpp
marehr, line 35:
```suggestion
// this class is shared for sam and bam tests
struct sam_file_data : public ::testing::Test
```

marehr: Like so?

marehr, line 36:
```suggestion
// this class is shared between sam and bam tests
struct sam_file_data : public ::testing::Test
```

joergi-w (approved):
Good, thank you!

eseiler (approved):

PR #2416, opened 2021-02-26
[FIX] kmer_hash combined with reverse

### LINKED ISSUE #2415: [BUG] kmer_hash and reverse for a list of one results in wrong result
```cpp
    std::vector<seqan3::dna4> text1{"TAAA"_dna4};
    std::vector<seqan3::dna4> text2{"TTTA"_dna4};
    auto v =  seqan3::views::kmer_hash(text1, seqan3::ungapped{4}) | std::views::reverse;
    auto v2 = text2 | seqan3::views::kmer_hash(seqan3::ungapped{4}) | std::views::reverse;
    seqan3::debug_stream << v << " " << v2<< "\n";
```

Results in: [48] [63]
It should be: [192] [252]

48 would be true for TAA and 63 for TTT. I don't know why the reverse leads to a wrong kmer_hash result. If I remove the std::views::reverse it is working as expected.

eseiler:
Resolves #2415

### DIFF
--- a/include/seqan3/range/views/kmer_hash.hpp
+++ b/include/seqan3/range/views/kmer_hash.hpp
@@ -327,16 +327,15 @@ public:
         assert(std::ranges::size(shape_) > 0);
 
         auto urange_size = std::ranges::distance(it_start, it_end);
-        auto step = (shape_.size() + 1 > urange_size) ? 0 : urange_size - shape_.size() + 1;
+        auto step = (shape_.size() > urange_size + 1) ? 0 : urange_size - shape_.size() + 1;
         text_left = std::ranges::next(it_start, step, it_end);
-        text_right = it_end;
 
         // shape size = 3
         // Text:      1 2 3 4 5 6 7 8 9
         // text_left: ^
         // text_right:    ^
         // distance(text_left, text_right) = 2
-        if (shape_.size() <= std::ranges::distance(text_left, text_right) + 1)
+        if (shape_.size() <= std::ranges::distance(text_left, it_end) + 1)
         {
             roll_factor = pow(sigma, static_cast<size_t>(std::ranges::size(shape_) - 1));
             hash_full();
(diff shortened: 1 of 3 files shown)


include/seqan3/range/views/kmer_hash.hpp
```diff
@@ -327,16 +327,15 @@ class kmer_hash_view<urng_t>::basic_iterator
         assert(std::ranges::size(shape_) > 0);
 
         auto urange_size = std::ranges::distance(it_start, it_end);
-        auto step = (shape_.size() + 1 > urange_size) ? 0 : urange_size - shape_.size() + 1;
+        auto step = (shape_.size() > urange_size + 1) ? 0 : urange_size - shape_.size() + 1;
```
eseiler, line 330: This was the bug

include/seqan3/range/views/kmer_hash.hpp
```diff
@@ -327,16 +327,15 @@ class kmer_hash_view<urng_t>::basic_iterator
...
         text_left = std::ranges::next(it_start, step, it_end);
-        text_right = it_end;
```
eseiler, line 332: This was an unnecessary assignment

MitraDarja (approved):
Thanks for fixing this so fast. 😍

include/seqan3/range/views/kmer_hash.hpp
marehr: Where is text_right defined? Can it be removed?

marehr (commented):
.

include/seqan3/range/views/kmer_hash.hpp
eseiler: It's a class member, cannot be removed. See `\details` of this constructor as to why we have `it_end`.

But what happened here was:
* `text_right = it_end` at this line
* Maybe calling `hash_full` which immediately does `text_right = text_left`
* `text_right = it_end` at the end

marehr (approved):

PR #2418, opened 2021-02-26
[FIX] BAM: Not consuming sequence

### LINKED ISSUE #2417: [BUG] Extracting the alignment from a BAM file referenes ID index not in range.
## Platform

  - SeqAn version: 
  - Operating system: Linux
  - Compiler: g++ 7.5

## Description
When I extract the alignment field of records in an alignment file, at some point it throws the following message:
```terminate called after throwing an instance of 'seqan3::format_error'
  what():  Reference id index '287314191' is not in range of header.ref_ids(), which has size 34.
```

### How to repeat the problem

Code snippet:
```cpp
#include <seqan3/io/alignment_file/input.hpp>   // SAM/BAM support
#include <filesystem>

int main()
{
    std::filesystem::path alignment_path = "test.bam";

    using my_fields = seqan3::fields<seqan3::field::id,
                                     seqan3::field::alignment>;

    seqan3::alignment_file_input alignment_file{alignment_path, my_fields{}};

    for (auto & [ id, alignment ] : alignment_file)
    {
        seqan3::debug_stream << ": " << id << ": " << std::get<1>(alignment) << '\n';
    }

}
```

Test file attached. It's a part of a larger BAM file, the error occurs at the last line of this part.

[test.bam.gz](https://github.com/seqan/seqan3/files/6049807/test.bam.gz)

I wonder if it has to do with the record trying to extract an alignment in an unmapped sequence?

eseiler:
Resolves #2417 

If we do not want to parse the sequence, but still want an alignment, but then have an empty CIGAR, we create a dummy alignment, but we forget to consume the sequence.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,11 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 * The `seqan3::views::kmer_hash` does not return wrong values when combined with `std::views::reverse` on a text of the
   same size as the kmer ([\#2416](https://github.com/seqan/seqan3/pull/2416)).
 
+#### I/O
+
+* Requesting the alignment without also requesting the sequence for BAM files containing empty CIGAR strings does now
+  not result in erroneous parsing ([\#2418](https://github.com/seqan/seqan3/pull/2418)).
+
 ## API changes
 
 #### Alphabet
--- a/include/seqan3/io/sam_file/format_bam.hpp
+++ b/include/seqan3/io/sam_file/format_bam.hpp
@@ -444,6 +444,13 @@ inline void format_bam::read_alignment_record(stream_type & stream,
 
         if constexpr (detail::decays_to_ignore_v<seq_type>)
         {
+            auto skip_sequence_bytes = [&] ()
+            {
+                detail::consume(seq_stream);
+                if (core.l_seq & 1)
+                    std::ranges::next(std::ranges::begin(stream_view));
+            };
+
             if constexpr (!detail::decays_to_ignore_v<align_type>)
             {
                 static_assert(sequence_container<std::remove_reference_t<decltype(get<1>(align))>>,
@@ -486,14 +493,13 @@ inline void format_bam::read_alignment_record(stream_type & stream,
                 }
                 else
                 {
+                    skip_sequence_bytes();
                     get<1>(align) = std::remove_reference_t<decltype(get<1>(align))>{}; // assign empty container
                 }
             }
             else
             {
-                detail::consume(seq_stream);
-                if (core.l_seq & 1)
-                    std::ranges::next(std::ranges::begin(stream_view));
+                skip_sequence_bytes();
             }
         }
         else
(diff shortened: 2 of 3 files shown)


test/unit/io/alignment_file/format_bam_test.cpp
```diff
@@ -516,3 +519,38 @@ TEST_F(bam_format, too_long_cigar_string_write)
...
+    std::istringstream stream{input};
+
+    seqan3::alignment_file_input fin{stream, seqan3::format_bam{}, seqan3::fields<seqan3::field::id,
+                                                                                  seqan3::field::alignment>{}};
+
+    std::vector<seqan3::gapped<seqan3::dna5>> const empty_sequence{};
+
+    // In 2417, the sequence was not consumed. Thus, wrong bytes were read for the following records.
+    // With the chosen `input` this also means that there will be more than 1 record in the alignment file.
+    // Hence, we need the for loop even though there is only 1 record.
```
marehr, line 549: I don't understand this remark.

eseiler: There's only one record in the bam file.
The bug caused the sequence to not be consumed, which also means that we think there is another record because we are not at the end of the stream after reading one record.
The second record will then be gibberish and throw something somewhere while being parsed.

Since I only read `id` and `alignment`, there is not really something I can check that would be parsed incorrectly in this one record (`id` is before the sequence and alignment is an empty dummy).

I could also check if the file iterator is at end after reading one record, but the for-loop is closer to the reported bug/use-case and also kinda easier to parse IMO.

marehr: Okay! Can you add a counter that counts the number of loop-passes? That one should be 1 right?

include/seqan3/io/alignment_file/format_bam.hpp
```diff
@@ -486,6 +486,9 @@ inline void format_bam::read_alignment_record(stream_type & stream,
                 }
                 else
                 {
+                    detail::consume(seq_stream);
+                    if (core.l_seq & 1)
+                        std::ranges::next(std::ranges::begin(stream_view));
```
marehr, line 491: Can you out-source this into a lambda?

test/unit/io/alignment_file/format_bam_test.cpp
```diff
@@ -22,7 +22,10 @@ struct alignment_file_read<seqan3::format_bam> : public alignment_file_data
     // formatted input
     // -----------------------------------------------------------------------------------------------------------------
     // See format_sam_test for the corresponding input in human readable-form.
-    // The byte sequence here are all uncompressed bam files since the file handles compression.
+    // The byte sequence here are all uncompressed (both gzip and bgzf) bam files since the file handles compression.
+    // Conversion can look like this: samtools view -u test.sam | bgzip -d
```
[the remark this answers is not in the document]
eseiler: Nope.
You write what you want in the SAM file. `samtools view -u` returns an uncompressed (=no gzip compression) BAM file, which you then bgzf-decompress.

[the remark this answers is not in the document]
marehr: Okay, I tried it with bam and worked to :)

marehr (approved):

marehr:
rebase :)

eaasna (approved):
Looks good to me :)

PR #2421, opened 2021-03-02
[FIX,TEST] missing DS tag in RG

eseiler:


joergi-w (approved):

marehr (approved):
Thank you even though I don't know why RG needs DS :)

eseiler:
> Thank you even though I don't know why RG needs DS :)

`@RG ID:group1 more info` is not valid SAM
`@RG ID:group1 DS:more info` is :)

Our test file was wrong.

PR #2423, opened 2021-03-03
[FIX] Parsing BAM headers with 64 reference sequences

### LINKED ISSUE #2422: SeqAn parses BAM headers past the value given in l_text
## Platform

  - SeqAn version: Commit df9fd5ff64f59fdb124c4a564a4141d1f9cff22b
  - Operating system: Linux whx-21 4.15.0-88-generic #88~16.04.1-Ubuntu SMP Wed Feb 12 04:19:15 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
  - Compiler: g++-9 (Ubuntu 9.3.0-10ubuntu2~16.04) 9.3.0

## Description

Under some conditions, SeqAn attempts to parse BAM headers past the value given in the l_text header. I believe the bug is usually masked by [a check in format_sam_base.hpp](https://github.com/seqan/seqan3/blob/df9fd5ff64f59fdb124c4a564a4141d1f9cff22b/include/seqan3/io/sam_file/format_sam_base.hpp#L471) which causes the header processing loop to stop. However, if the content of the n_ref field in the BAM headers is e.g. 64, which equals 0x40000000 in little endian or “@” followed by nul characters, SeqAn will treat the “@” as the beginning of the next header and, since the next character is nul, throw an exception about an illegal SAM header.

The number of remaining characters in the passed `stream_view` does not seem to be updated correctly in this case. I believe the reason is that in function [`parse_tag_value`](https://github.com/seqan/seqan3/blob/df9fd5ff64f59fdb124c4a564a4141d1f9cff22b/include/seqan3/io/sam_file/format_sam_base.hpp#L463), a call to `detail::consume` updates the `target_size` of the newly created range (the one passed to `detail::consume`) but not that of the passed `stream_view`. 

### How to repeat the problem

Parse a BAM file with headers such that `parse_tag_value` is called and the value of the n_ref field is e.g. 0x40000000.

### Expected behaviour

File parsed correctly.

### Actual behaviour

SeqAn fails by throwing the exception as a result of parsing the headers outside the range given in the file.

eseiler:
Resolves #2422 

Todo:
- [x] BAM works, SAM does not
- [x] Test
- [x] Changelog
- [x] Check for performance regression

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,7 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 
 * Requesting the alignment without also requesting the sequence for BAM files containing empty CIGAR strings does now
   not result in erroneous parsing ([\#2418](https://github.com/seqan/seqan3/pull/2418)).
+* BAM files with 64 references are now parsed correctly ([\#2423](https://github.com/seqan/seqan3/pull/2423)).
 
 ## API changes
 
--- a/include/seqan3/io/sam_file/format_sam_base.hpp
+++ b/include/seqan3/io/sam_file/format_sam_base.hpp
@@ -493,16 +493,40 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
                                          ref_seqs_type & /*ref_id_to_pos_map*/)
 {
     auto it = std::ranges::begin(stream_view);
+    auto end = std::ranges::end(stream_view);
+    std::vector<char> string_buffer{};
 
-    auto parse_tag_value = [&stream_view, &it, this] (auto & value) // helper function to parse the next tag value
+    auto take_until_predicate = [&it, &string_buffer] (auto const & predicate)
     {
-        detail::consume(stream_view | views::take_until_or_throw(is_char<':'>)); // skip tag name
-        std::ranges::next(std::ranges::begin(stream_view));                     // skip ':'
-        read_field(stream_view | views::take_until_or_throw(is_char<'\t'> || is_char<'\n'>), value);
-        it = std::ranges::begin(stream_view);                                    // make sure iterator is up2date
+        string_buffer.clear();
+        while (!predicate(*it))
+        {
+            string_buffer.push_back(*it);
+            ++it;
+        }
     };
 
-    while (is_char<'@'>(*it))
+    auto skip_tag_name_and_colon = [&it] ()
+    {
+        while (!is_char<':'>(*it))
+            ++it;
+        ++it;
+    };
+
+    auto parse_tag_value = [&] (auto & value) // helper function to parse the next tag value
+    {
+        skip_tag_name_and_colon();
+        take_until_predicate(is_char<'\t'> || is_char<'\n'>);
+        read_field(string_buffer, value);
+    };
+
+    auto parse_until_newline = [&] (auto & value)
+    {
+        take_until_predicate(is_char<'\n'>);
+        read_field(string_buffer, value);
+    };
+
+    while (it != end && is_char<'@'>(*it))
     {
         ++it;                                                           // skip @
 
@@ -552,7 +576,7 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
                 if (is_char<'\t'>(*it))                                 // read rest of the tags
                 {
                     ++it;                                               // skip tab
-                    read_field(stream_view | views::take_until_or_throw(is_char<'\n'>), get<1>(info));
+                    parse_until_newline(get<1>(info));
                 }
                 ++it;                                                   // skip newline
 
@@ -594,7 +618,7 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
                 if (is_char<'\t'>(*it))                                 // read rest of the tags
                 {
                     ++it;
-                    read_field(stream_view | views::take_until_or_throw(is_char<'\n'>), get<1>(tmp));
+                    parse_until_newline(get<1>(tmp));
                 }
                 ++it;                                                   // skip newline
 
@@ -653,7 +677,7 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
                 ++it;                                                   // skip C
                 ++it;                                                   // skip O
                 ++it;                                                   // skip :
-                read_field(stream_view | views::take_until_or_throw(is_char<'\n'>), tmp);
+                parse_until_newline(tmp);
                 ++it;                                                   // skip newline
 
                 hdr.comments.emplace_back(std::move(tmp));
--- a/test/unit/io/sam_file/format_bam_test.cpp
+++ b/test/unit/io/sam_file/format_bam_test.cpp
@@ -180,6 +180,47 @@ struct sam_file_read<seqan3::format_bam> : public sam_file_data
         '\x66', '\x66', '\x66', '\x46', '\x40', '\x7A', '\x7A', '\x5A', '\x73', '\x74', '\x72', '\x00'
     };
 
+    std::string many_refs{[] ()
+    {
+        // X = non-printable character
+        std::array<char, 2> buffer;
+        //                   B       A       M       X      ----------- 1345 -----------     @       H       D
+        std::string result{'\x42', '\x41', '\x4D', '\x01', '\x41', '\x05', '\x00', '\x00', '\x40', '\x48', '\x44',
+        //                   \t      V       N       :       1       .       6       \n
+                           '\x09', '\x56', '\x4E', '\x3A', '\x31', '\x2E', '\x36', '\x0A'};
+        for (char i = '0'; i <= '9'; ++i)
+            //                      @       S       Q       \t      S       N       :       r       e       f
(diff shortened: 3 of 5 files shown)


include/seqan3/io/sam_file/format_sam_base.hpp
```diff
@@ -459,16 +459,34 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
+    };
+
+    auto parse_tag_value = [&] (auto & value) // helper function to parse the next tag value
+    {
+        consume();
+        call_read_field(is_char<'\t'> || is_char<'\n'>, value);
     };
 
-    while (is_char<'@'>(*it))
+    while (it != end && is_char<'@'>(*it))
```
eseiler, line 489:
```suggestion
    while (it != end && is_char<'@'>(*it))
```

`it != end` is used in BAM, the `stream_view` is a `view_take_exactly_or_throw` and we need to stop at the end (the size is known).

`is_char<'@'>(*it)` is used in SAM, the `stream_view` is a view over the file and we stop when we read all lines that start with `@`.

include/seqan3/io/sam_file/format_sam_base.hpp
```diff
@@ -459,16 +459,34 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
+    auto consume = [&it] ()
+    {
+        while (!is_char<':'>(*it))
+            ++it;
+        ++it;
+    };
+
+    auto parse_tag_value = [&] (auto & value) // helper function to parse the next tag value
+    {
+        consume();
```
eseiler, line 485:
```suggestion
    auto take_until_colon_and_consume = [&it] ()
    {
        while (!is_char<':'>(*it))
            ++it;
        ++it;
    };

    auto parse_tag_value = [&] (auto & value) // helper function to parse the next tag value
    {
        take_until_colon_and_consume();
```

include/seqan3/io/sam_file/format_sam_base.hpp
```diff
@@ -459,16 +459,34 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
                                          ref_seqs_type & /*ref_id_to_pos_map*/)
 {
     auto it = std::ranges::begin(stream_view);
+    auto end = std::ranges::end(stream_view);
+    std::string string_buffer{};
 
-    auto parse_tag_value = [&stream_view, &it, this] (auto & value) // helper function to parse the next tag value
+    auto call_read_field = [&it, &string_buffer, this] (auto predicate, auto & value)
```
eseiler, line 465:
```suggestion
    auto call_read_field = [&it, &string_buffer, this] (auto const & predicate, auto & value)
```

test/unit/io/sam_file/format_bam_test.cpp
```diff
@@ -179,6 +179,47 @@ struct sam_file_read<seqan3::format_bam> : public sam_file_data
         '\x66', '\x66', '\x66', '\x46', '\x40', '\x7A', '\x7A', '\x5A', '\x73', '\x74', '\x72', '\x00'
     };
 
+    std::string many_refs{[] ()
```
eseiler, line 182: The BAM would be multiple hundred lines long, that's why there it is generated.

The comments + BAM specification make it feasible to understand.

CHANGELOG.md
```diff
@@ -80,6 +80,7 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 
 * Requesting the alignment without also requesting the sequence for BAM files containing empty CIGAR strings does now
   not result in erroneous parsing ([\#2418](https://github.com/seqan/seqan3/pull/2418)).
+* BAM files with 64 references are now parsed correctly ([\#2423](https://github.com/seqan/seqan3/pull/2423)).
```
eseiler, line 83: Technically it's every number of reference whose last byte is `\x40`, e.g. it's also affecting `320`, but I didn't come up with a concise way to describe this.

include/seqan3/io/sam_file/format_sam_base.hpp
```diff
@@ -459,16 +459,34 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
+        }
+        read_field(string_buffer, value);
+    };
+
+    auto take_until_colon_and_consume = [&it] ()
+    {
+        while (!is_char<':'>(*it))
+            ++it;
+        ++it;
+    };
```
joergi-w, line 481: `views::take_until(is_char<':'>)` doesn't work here?

joergi-w (approved):
I checked this PR regarding the syntax and what I know about I/O, but please get another review from someone who knows the sam/bam specific semantics better than me...

include/seqan3/io/sam_file/format_sam_base.hpp
eseiler: No, see the issue.
We would make a copy of the `stream_view` and hence not update the size of the orignal `stream_view`

CHANGELOG.md
marehr: (mod 64) ?

include/seqan3/io/sam_file/format_sam_base.hpp
```diff
@@ -459,16 +459,34 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
+        string_buffer.clear();
+        while (!predicate(*it))
+        {
+            string_buffer.push_back(*it);
+            ++it;
+        }
+        read_field(string_buffer, value);
+    };
+
+    auto take_until_colon_and_consume = [&it] ()
```
marehr, line 476:
```suggestion
    auto skip_tag_name_and_colon = [&it] ()
```

include/seqan3/io/sam_file/format_sam_base.hpp
```diff
@@ -459,16 +459,34 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
                                          ref_seqs_type & /*ref_id_to_pos_map*/)
 {
     auto it = std::ranges::begin(stream_view);
+    auto end = std::ranges::end(stream_view);
+    std::string string_buffer{};
 
-    auto parse_tag_value = [&stream_view, &it, this] (auto & value) // helper function to parse the next tag value
+    auto call_read_field = [&it, &string_buffer, this] (auto const & predicate, auto & value)
```
marehr, line 465:
```suggestion
    auto take_until_predicate = [&it, &string_buffer, this] (auto const & predicate, auto & value)
```

include/seqan3/io/sam_file/format_sam_base.hpp
```diff
@@ -459,16 +459,34 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
-        std::ranges::next(std::ranges::begin(stream_view));                     // skip ':'
-        read_field(stream_view | views::take_until_or_throw(is_char<'\t'> || is_char<'\n'>), value);
-        it = std::ranges::begin(stream_view);                                    // make sure iterator is up2date
+        string_buffer.clear();
+        while (!predicate(*it))
+        {
+            string_buffer.push_back(*it);
+            ++it;
+        }
+        read_field(string_buffer, value);
```
marehr, line 473:
```suggestion
```

include/seqan3/io/sam_file/format_sam_base.hpp
```diff
@@ -459,16 +459,34 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
+    {
+        while (!is_char<':'>(*it))
+            ++it;
+        ++it;
+    };
+
+    auto parse_tag_value = [&] (auto & value) // helper function to parse the next tag value
+    {
+        take_until_colon_and_consume();
+        call_read_field(is_char<'\t'> || is_char<'\n'>, value);
```
marehr, line 486:
```suggestion
        skip_tag_name_and_colon();
        take_until_predicate(is_char<'\t'> || is_char<'\n'>, value);
        read_field(string_buffer, value);
```

include/seqan3/io/sam_file/format_sam_base.hpp
```diff
@@ -459,16 +459,34 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
                                          ref_seqs_type & /*ref_id_to_pos_map*/)
 {
     auto it = std::ranges::begin(stream_view);
+    auto end = std::ranges::end(stream_view);
+    std::string string_buffer{};
```
marehr, line 463: Please use vector here, because std::string can be quite slower because of the small string optimization

test/unit/io/sam_file/format_bam_test.cpp
marehr: We could have also generated a bam file and added it as file :D

marehr (commented):
thank you :)

CHANGELOG.md
eseiler: Well, it works properly for 128. The first byte has to look like `01000000` or `11000000`, i.e. the first least significant bit to be set is the seventh...

marehr: you are right: (mod 256) == 64

marehr (approved):
very nice!

PR #2427, opened 2021-03-03
alphabet_base expects functions instead of look-up tables.

marehr:
This is part of https://github.com/seqan/product_backlog/issues/295

This changes `rank_to_char` and `char_to_rank` to static member functions.

Blocked by https://github.com/seqan/seqan3/pull/2430.


include/seqan3/alphabet/composite/alphabet_variant.hpp
```diff
@@ -595,6 +595,19 @@ class alphabet_variant : public alphabet_base<alphabet_variant<alternative_types
 
         return char_to_rank;
     }();
+
+    /*!\brief Compile-time generated lookup table which maps the char to rank.
+     *
+     * An map generated at compile time where the key is the char of one of the
```
simonsasse, line 610:
```suggestion
     * A map generated at compile time where the key is the char of one of the
```

simonsasse (changes requested):
Found just one typo :)

simonsasse (commented):
Looks good. Only thing is that when I see `chr`, I always think of `chromosome` and not `character`. Maybe its worth changing the `chr` to `char`. Probably just personal preference.

eseiler:
> Looks good. Only thing is that when I see `chr`, I always think of `chromosome` and not `character`. Maybe its worth changing the `chr` to `char`. Probably just personal preference.

Can you approve and assign core for review?

simonsasse (approved):
✅

include/seqan3/alphabet/quality/quality_base.hpp
```diff
@@ -141,25 +141,11 @@ class quality_base : public alphabet_base<derived_type, size, char>
...
-                else if (i >= derived_type::offset_char + alphabet_size)  // map too-large to highest possible
-                    ret[static_cast<rank_type>(i)] = alphabet_size - 1;
-                else                                                   // map valid range to identity
-                    ret[static_cast<rank_type>(i)] = i - derived_type::offset_char;
-            }
-
-            return ret;
-        }()
-    };
+        int difference = chr - derived_type::offset_char;
```
eseiler, line 146:
```suggestion
        int64_t difference = chr - derived_type::offset_char;
```

If I assume that `char_type` is `char32_t`, I would subtract two `uint32_t` (actually `uint_least32_t`).


Or would `auto` work?
And clamp needs all three parameters to be of the same type, but if all are signed, it should be fine?
But I think `int` is not generally correct.

eseiler (changes requested):
Looks good in general, just some documentation remarks and general questions.

---

This point does not need / should not be addressed in this PR:

I am missing a bit some documentation on what the advantage of this is. If there is a changelog entry, it should shortly describe why this is good?
I think `doc/howto/write_an_alphabet/dna2_derive_from_base.cpp` can also shortly mention that there is an alternative to tables (it does show it, but doesn't explicitly mention it).

include/seqan3/alphabet/quality/quality_base.hpp
marehr: You are right. I hate these integers implicit conversion things. This one will still throw warnings if we had an architecture where `chr` is unsigned.

I guess

```suggestion
        int64_t difference = static_cast<int64_t>(chr) - static_cast<int64_t>(derived_type::offset_char);
```

would be the most correct version.

include/seqan3/alphabet/alphabet_base.hpp
```diff
@@ -156,16 +196,42 @@ class alphabet_base
...
+#   pragma GCC diagnostic push
+#   pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+        using index_t = std::make_unsigned_t<char_type>;
+        index_t const id = static_cast<index_t>(chr);
+        return derived_type::char_to_rank[id];
+#   pragma GCC diagnostic pop
+#else // ^^^ workaround / no workaround vvv
+        using index_t = std::make_unsigned_t<char_type>;
+        return derived_type::char_to_rank[static_cast<index_t>(chr)];
+#endif // SEQAN3_WORKAROUND_GCC_99318
```
marehr, line 230:
```suggestion
        using index_t = std::make_unsigned_t<char_type>;
#if SEQAN3_WORKAROUND_GCC_99318
#   pragma GCC diagnostic ignored "-Wdeprecated-declarations"
#endif // SEQAN3_WORKAROUND_GCC_99318
        return derived_type::char_to_rank[static_cast<index_t>(chr)];
```

include/seqan3/alphabet/aminoacid/aa10li.hpp
```diff
@@ -160,6 +166,13 @@ class aa10li : public aminoacid_base<aa10li, 10>
             return ret;
         }()
     };
+
+    //!\copydoc seqan3::aa27::char_to_rank
+    static constexpr rank_type char_to_rank(char_type const chr)
+    {
+        using index_t = std::make_unsigned_t<char_type>;
+        return char_to_rank_table[static_cast<index_t>(chr)];
+    }
```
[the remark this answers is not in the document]
marehr: That's unfortunately not possible.

include/seqan3/alphabet/aminoacid/aminoacid_base.hpp
```diff
@@ -129,9 +129,14 @@ class aminoacid_base : public alphabet_base<derived_type, size, char>, public am
...
             {
-                ret[         c ] = true;
+                uint8_t c{};
+                if constexpr (detail::has_rank_to_char_table<derived_type>)
+                    c = derived_type::rank_to_char[rank];
+                else
+                    c = derived_type::rank_to_char(rank);
+                ret[c] = true;
                 ret[to_lower(c)] = true;
             }
```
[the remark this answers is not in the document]
marehr: I added a `// ^^^ before 3.1.0 release / after 3.1.0 release vvv` comment.

marehr:
I added / changed documentation, I tried to clarify the snippets.

Otherwise, I used your suggestion not to use the term lookup table, but just a general description.

doc/howto/write_an_alphabet/dna2_derive_from_base.cpp
```diff
@@ -9,15 +9,40 @@
...
+    friend seqan3::alphabet_base<dna2, 2>;
+
     // map 0 => 'S' and 1 => 'W'
-    static constexpr char_type rank_to_char[alphabet_size] {'S', 'W'};
+    static constexpr char_type rank_to_char(rank_type const rank)
+    {
+        // via a lookup table
+        return rank_to_char_table[rank];
+        // or via an arithmetic expression
+        return rank == 1 ? 'W' : 'S';
```
eseiler, line 27: clang with `-Wunreachable-code-return` will complain about an unreachable return, but no strong feelings

eseiler (changes requested):
Please don't force-push during reviews, I have no idea what you changed. Looks like you addressed the points though.

Can you check the code coverage? Maybe you need to rebase on master

doc/howto/write_an_alphabet/dna2_derive_from_base.cpp
marehr: yeah I thought about this too, I wasn't sure how to express this.

Maybe two different implementations?

eseiler (approved):

PR #2428, opened 2021-03-04
[FEATURE] interleaved_bloom_filter::clear

eseiler:
resolves seqan/product_backlog#296

### DIFF
--- a/include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
+++ b/include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
@@ -278,7 +278,7 @@ public:
      *
      * \include test/snippet/search/dream_index/interleaved_bloom_filter_emplace.cpp
      */
-    void emplace(size_t const value, bin_index const bin)
+    void emplace(size_t const value, bin_index const bin) noexcept
     //!\cond
         requires (data_layout_mode == data_layout::uncompressed)
     //!\endcond
@@ -291,6 +291,59 @@ public:
             assert(idx < data.size());
             data[idx] = 1;
         };
+    }
+
+    /*!\brief Clears a specific bin.
+     * \param[in] bin The bin index to clear.
+     *
+     * \attention This function is only available for **uncompressed** Interleaved Bloom Filters.
+     *
+     * \details
+     *
+     * ### Example
+     *
+     * \include test/snippet/search/dream_index/interleaved_bloom_filter_clear.cpp
+     */
+    void clear(bin_index const bin) noexcept
+    //!\cond
+        requires (data_layout_mode == data_layout::uncompressed)
+    //!\endcond
+    {
+        assert(bin.get() < bins);
+        for (size_t idx = bin.get(), i = 0; i < bin_size_; idx += technical_bins, ++i)
+            data[idx] = 0;
+    }
+
+    /*!\brief Clears a range of bins.
+     * \tparam rng_t The type of the range. Must model std::ranges::forward_range and the reference type must be
+     *               seqan3::bin_index.
+     * \param[in] bin_range The range of bins to clear.
+     *
+     * \attention This function is only available for **uncompressed** Interleaved Bloom Filters.
+     *
+     * \details
+     *
+     * ### Example
+     *
+     * \include test/snippet/search/dream_index/interleaved_bloom_filter_clear.cpp
+     */
+    template <typename rng_t>
+    //!\cond
+        requires (data_layout_mode == data_layout::uncompressed)
+    //!\endcond
+    void clear(rng_t && bin_range) noexcept
+    {
+        static_assert(std::ranges::forward_range<rng_t>, "The range of bins to clear must model a forward_range.");
+        static_assert(std::same_as<std::remove_cvref_t<std::ranges::range_reference_t<rng_t>>, bin_index>,
+                      "The reference type of the range to clear must be seqan3::bin_index.");
+#ifndef NDEBUG
+        for (auto && bin : bin_range)
+            assert(bin.get() < bins);
+#endif // NDEBUG
+
+        for (size_t offset = 0, i = 0; i < bin_size_; offset += technical_bins, ++i)
+            for (auto && bin : bin_range)
+                data[bin.get() + offset] = 0;
     }
 
     /*!\brief Increases the number of bins stored in the Interleaved Bloom Filter.
--- a/test/performance/search/dream_index/interleaved_bloom_filter_benchmark.cpp
+++ b/test/performance/search/dream_index/interleaved_bloom_filter_benchmark.cpp
@@ -7,6 +7,7 @@
 
 #include <benchmark/benchmark.h>
 
+#include <seqan3/range/views/to.hpp>
 #include <seqan3/range/views/zip.hpp>
 #include <seqan3/search/dream_index/interleaved_bloom_filter.hpp>
 #include <seqan3/test/performance/sequence_generator.hpp>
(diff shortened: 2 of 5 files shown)


test/performance/search/dream_index/interleaved_bloom_filter_benchmark.cpp
```diff
@@ -20,13 +21,19 @@ inline benchmark::Counter hashes_per_second(size_t const count)
...
+        for (int32_t bits = 15; bits <= 20; bits += 5)
         {
-            for (int32_t hash_num = 2; hash_num < 3/* Increase for more extensive benchmarks*/; ++hash_num)
+            // The bits per bin must fit in an int32_t
+            if (bits - std::countr_zero(static_cast<uint32_t>(bins)) < 32)
             {
-                b->Args({bins, bits/bins, hash_num, 1'000/* Increase for more extensive benchmarks*/});
+                for (int32_t hash_num = 2; hash_num < 3; ++hash_num)
+                {
+                    b->Args({bins, (1LL << bits)/bins, hash_num, 1'000});
```
MitraDarja, line 35: Why did you add the 1LL here? And why did you remove the comment to increase 1000 for more extensive benchmarks?

test/performance/search/dream_index/interleaved_bloom_filter_benchmark.cpp
```diff
@@ -63,6 +70,52 @@ void emplace_benchmark(::benchmark::State & state)
...
+void clear_range_benchmark(::benchmark::State & state)
+{
+    auto && [ bin_indices, hash_values, ibf ] = set_up<ibf_type>(state.range(0),
+                                                                 state.range(1),
+                                                                 state.range(2),
+                                                                 state.range(3));
+    (void) bin_indices;
+    (void) hash_values;
+
+    std::vector<seqan3::bin_index> bin_range = std::views::iota(0u, static_cast<size_t>(state.range(0)))
```
MitraDarja, line 106: Maybe I am blind, but is this not the same as in clear_benchmark()? I don't see what this function does different.

MitraDarja (changes requested):
Thanks, that is a really useful function. I just have two questions, the rest looks good to me. :)

test/performance/search/dream_index/interleaved_bloom_filter_benchmark.cpp
eseiler: Before, `bits` was the overall IBF size; but googlebenchmark only accepts `int32_t` as input for arguments. So we were limited to at most a size of 2^31 basically.
What was worse was that if you put `bits <= 35` (or anything > 31), there was an overflow and the program did not terminate and consumed all memory.

Now you set the exponent, i.e. `2^bits`. `1LL << bits` is equivalent to `2^bits`. We use signed integers to not need to cast again. 
And we can also avoid the problem with `bits > 31`.

I removed the comments because it's basically the same in every benchmark we have: if you want to benchmark more, you need to change parameters.

test/performance/search/dream_index/interleaved_bloom_filter_benchmark.cpp
eseiler: This expression is the same.
Inside the benchmarked loop we use either the clear on a single bin or the clear on a range.

MitraDarja (approved):
Thanks for explaining. :)

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -293,6 +293,59 @@ class interleaved_bloom_filter
...
+     */
+    template <typename rng_t>
+    //!\cond
+        requires (data_layout_mode == data_layout::uncompressed)
+    //!\endcond
+    void clear(rng_t && bin_range) noexcept
+    {
+        static_assert(std::ranges::input_range<rng_t>, "The range of bins to clear must model an input_range.");
+        static_assert(std::same_as<std::remove_cvref_t<std::ranges::range_reference_t<rng_t>>, bin_index>,
+                      "The reference type of the range to clear must be seqan3::bin_index.");
```
eseiler, line 338:
```suggestion
    /*!\brief Clears a range of bins.
     * \tparam rng_t The type of the range. Must model std::ranges::forward_range and the reference type must be
     *               seqan3::bin_index.
     * \param[in] bin_range The range of bins to clear.
     *
     * \attention This function is only available for **uncompressed** Interleaved Bloom Filters.
     *
     * \details
     *
     * ### Example
     *
     * \include test/snippet/search/dream_index/interleaved_bloom_filter_clear.cpp
     */
    template <typename rng_t>
    //!\cond
        requires (data_layout_mode == data_layout::uncompressed)
    //!\endcond
    void clear(rng_t && bin_range) noexcept
    {
        static_assert(std::ranges::forward_range<rng_t>, "The range of bins to clear must model an forward_range.");
        static_assert(std::same_as<std::remove_cvref_t<std::ranges::range_reference_t<rng_t>>, bin_index>,
                      "The reference type of the range to clear must be seqan3::bin_index.");
```
```suggestion
    /*!\brief Clears a range of bins.
     * \tparam rng_t The type of the range. Must model std::ranges::input_range and the reference type must be
     *               seqan3::bin_index.
     * \param[in] bin_range The range of bins to clear.
     *
     * \attention This function is only available for **uncompressed** Interleaved Bloom Filters.
     *
     * \details
     *
     * ### Example
     *
     * \include test/snippet/search/dream_index/interleaved_bloom_filter_clear.cpp
     */
    template <typename rng_t>
    //!\cond
        requires (data_layout_mode == data_layout::uncompressed)
    //!\endcond
    void clear(rng_t && bin_range) noexcept
    {
        static_assert(std::ranges::input_range<rng_t>, "The range of bins to clear must model an input_range.");
        static_assert(std::same_as<std::remove_cvref_t<std::ranges::range_reference_t<rng_t>>, bin_index>,
                      "The reference type of the range to clear must be seqan3::bin_index.");
```

include/seqan3/search/dream_index/interleaved_bloom_filter.hpp
```diff
@@ -293,6 +293,59 @@ class interleaved_bloom_filter
...
+     *
+     * ### Example
+     *
+     * \include test/snippet/search/dream_index/interleaved_bloom_filter_clear.cpp
+     */
+    template <typename rng_t>
+    //!\cond
+        requires (data_layout_mode == data_layout::uncompressed)
+    //!\endcond
+    void clear(rng_t && bin_range) noexcept
```
marehr, line 334: I guess that this function only exists in the case that bins are close to each other and clearing them would be more cache efficient?

marehr: But that would assume sorting on the bins. I guess you collect all bins to be cleared in a vector. Do you collect them in-order?

eseiler: > I guess that this function only exists in the case that bins are close to each other and clearing them in bulk would be more cache efficient?

Yes, has a similar effect as having a range overload for `emplace`.
I did some benchmarks, and you have some cache effects, the more bins you have and the bigger the IBF, the better the speedup (inversely, for small IBFs (a few MiB) and low bin count (<256), the non-range overload is faster). I did not look into how many bins you need to clear at once.
I saw x10 for 4 GiB and 1024 bins for clearing all bins.

> But that would assume sorting on the bins. I guess you collect all bins to be cleared in a vector. Do you collect them in-order?

I kinda assume sorting - I don't know how the bins are collected in an application.
I could:
  * Sort the range (I would then require `std::ranges::random_access_range`)
    * If I do this, I might as well make the range unique?
  * Document that it would be good if the range was sorted (and unique).

marehr: > Yes, has a similar effect as having a range overload for `emplace`.

We don't have one right?

> I kinda assume sorting - I don't know how the bins are collected in an application. I could:

I'm torn because I know that this can have a performance impact. The main problem is that the data layout is in strives and I would assume that most bins (~1024) fit in one cache line and even "few" bins should see some benefit from that even if not sorted (since they fall in the same cache line).

Form an interface perspective a range "for a" clear is somehow strange.

What I'm currently thinking whether we can use the "binning_bitvector" to clear the bins?

eseiler: > We don't have one right?

Not yet, but we should add it soon.

> Form an interface perspective a range "for a" clear is somehow strange.

`erase` would probably be a better name; however `erase` usually also does not have a range interface

> What I'm currently thinking is whether we can use the "binning_bitvector" to clear the bins?

How do you mean exactly?
But now that you mention it, it would probably be better to maybe manipulate words instead of bits.
I.e. erasing bins [1,2,3,63] would be one "word access", but 4 "bit accesses".

marehr: > How do you mean exactly?

I was asking myself: "When do you clear?" Probably because of a "contain" query. So can't we use the same data type?

eseiler: You want to erase a bin as part of an update procedure.
If you build an IBF over a database and, say, `genome_1` is in `bin_1`, but then in an update `genome_1` gets deleted, you will want to erase `bin_1`.

So basically, you need `erase` to do remove and edit of bins affected by some database update.

marehr (approved):

PR #2337, opened 2021-01-14
[MISC] Update writable_alphabet concept

MitraDarja:
Resolves https://github.com/seqan/product_backlog/issues/278

PR #2430, opened 2021-03-05
[FIX] fix internal compiler error on gcc-7

marehr:
Changes in https://github.com/seqan/seqan3/pull/2427 triggered an ICE (in Fact the same one that @MitraDarja had in https://github.com/seqan/seqan3/pull/2337).

This is an attempt to fix this ICE once and for all by using an "easier" way of using is_valid_char at compile time.

Also part of https://github.com/seqan/product_backlog/issues/124, since it throws away some meta-library functionality.

### DIFF
--- a/test/unit/alphabet/composite/alphabet_variant_detail_test.cpp
+++ b/test/unit/alphabet/composite/alphabet_variant_detail_test.cpp
@@ -21,6 +21,7 @@ public:
     using seqan3::alphabet_variant<alphabet_types...>::partial_sum_sizes;
     using seqan3::alphabet_variant<alphabet_types...>::rank_to_char;
     using seqan3::alphabet_variant<alphabet_types...>::char_to_rank;
+    using seqan3::alphabet_variant<alphabet_types...>::first_valid_char_table;
 };
 
 TEST(alphabet_variant_detail_test, partial_sum_sizes)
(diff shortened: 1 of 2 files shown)


test/unit/alphabet/composite/alphabet_variant_detail_test.cpp
```diff
@@ -123,3 +124,48 @@ TEST(alphabet_variant_detail_test, char_to_rank)
...
+    EXPECT_EQ(first_valid_char_table4['A'], 0);
+    EXPECT_EQ(first_valid_char_table4['C'], 0);
+    EXPECT_EQ(first_valid_char_table4['G'], 0);
+    EXPECT_EQ(first_valid_char_table4['N'], 0);
+    EXPECT_EQ(first_valid_char_table4['T'], 0);
+    EXPECT_EQ(first_valid_char_table4['-'], 1);
+    EXPECT_EQ(first_valid_char_table4['A'], 0);
+    EXPECT_EQ(first_valid_char_table4['C'], 0);
+    EXPECT_EQ(first_valid_char_table4['G'], 0);
+    EXPECT_EQ(first_valid_char_table4['T'], 0);
```
Irallia, line 169: Why are these lines duplicated?

test/unit/alphabet/composite/alphabet_variant_detail_test.cpp
```diff
@@ -123,3 +124,48 @@ TEST(alphabet_variant_detail_test, char_to_rank)
...
+{
+    constexpr std::array first_valid_char_table2 = detail_alphabet_variant<seqan3::dna4,
+                                                                           seqan3::gap>::first_valid_char_table;
+    EXPECT_EQ(first_valid_char_table2.size(), 256u);
+    EXPECT_EQ(first_valid_char_table2['A'], 0);
+    EXPECT_EQ(first_valid_char_table2['C'], 0);
+    EXPECT_EQ(first_valid_char_table2['G'], 0);
+    EXPECT_EQ(first_valid_char_table2['T'], 0);
+    EXPECT_EQ(first_valid_char_table2['-'], 1);
+    EXPECT_EQ(first_valid_char_table2['X'], 2); // invalid char
```
Irallia, line 138: Do the number mean, to which alphabet in the table the character belongs to? And if the number is higher than the table dimension, its not a valid character?

test/unit/alphabet/composite/alphabet_variant_detail_test.cpp
```diff
@@ -123,3 +124,48 @@ TEST(alphabet_variant_detail_test, char_to_rank)
...
+    EXPECT_EQ(first_valid_char_table3['A'], 0);
+    EXPECT_EQ(first_valid_char_table3['C'], 0);
+    EXPECT_EQ(first_valid_char_table3['G'], 0);
+    EXPECT_EQ(first_valid_char_table3['N'], 2);
+    EXPECT_EQ(first_valid_char_table3['T'], 0);
+    EXPECT_EQ(first_valid_char_table3['X'], 3); // invalid char
+
+    constexpr std::array first_valid_char_table4 = detail_alphabet_variant<seqan3::dna5,
+                                                                           seqan3::gap,
+                                                                           seqan3::dna4>::first_valid_char_table;
```
Irallia, line 158: Maybe its nicer to check with the following table? Because than you get 'N' with 2?
```suggestion
    constexpr std::array first_valid_char_table4 = detail_alphabet_variant<seqan3::dna4,
                                                                           seqan3::gap,
                                                                           seqan3::dna5>::first_valid_char_table;
```

Irallia (changes requested):
I have some ideas and questions about the tests.

test/unit/alphabet/composite/alphabet_variant_detail_test.cpp
marehr: Yeah, kind of. Except when two or more alphabets have the same "valid" char, it chooses the first one.

test/unit/alphabet/composite/alphabet_variant_detail_test.cpp
marehr: That is already tested by `first_valid_char_table3`

test/unit/alphabet/composite/alphabet_variant_detail_test.cpp
marehr: This lists all valid chars for each alphabet and shows that the table stores the first index.

Irallia (approved):
Thanks for the explanation.

eseiler (approved):

PR #2434, opened 2021-03-09
[MISC] Add data access to IBF

eseiler:


eseiler:
> LGTM, but what is the use case of this?

That you can access the actual memory directly if you want to (e.g., containers (also ours) usually offer this). 

In my specific usecase, I want to know how big the IBF is, and the SDSL offers functionality for its bitvector

marehr (approved):

PR #2435, opened 2021-03-09
[FIX] SeqAn3 cmake not working with -Werror

### LINKED ISSUE #2433: SeqAn3 cannot be configured with -Werror
## Description

We cannot use the `-Werror` flag via `-DCMAKE_CXX_FLAGS`, it will result in failed feature detection.

The problem is as follows (by example of the concept detection):
-DCONCEPTS_FLAG-fconcepts
https://github.com/seqan/seqan3/blob/696b6e3776ea7bb6659335cf5d3f6c545a768f0e/build_system/seqan3-config.cmake#L255-L277

The line
https://github.com/seqan/seqan3/blob/696b6e3776ea7bb6659335cf5d3f6c545a768f0e/build_system/seqan3-config.cmake#L263
will create a compiler call to compile and run a source file with some flags:
```
/usr/bin/g++-9  [...] -Werror -DCONCEPTS_FLAG-fconcepts -Werror -std=c++17 -fconcepts -o [...] -c [...]
```

which will give us

```
error: ISO C++11 requires whitespace after the macro name [-Werror]
```

I.e., `-DCONCEPTS_FLAG-fconcepts` is actually not valid.

(CMake try_compile outputs can be debugged via `--debug-trycompile`)

### How to repeat the problem

Configure anything that includes SeqAn3 with `-DCMAKE_CXX_FLAGS="-Werror"`, including our test suites.

### Expected behaviour

Works :)

### Actual behaviour

Worksn't :(

### Workaround
When using SeqAn3 as dependency:
```cmake
set (WORKAROUND_2433 "${CMAKE_CXX_FLAGS}")
set (CMAKE_CXX_FLAGS "")

list(APPEND CMAKE_PREFIX_PATH "${CMAKE_CURRENT_SOURCE_DIR}/include/seqan3/build_system")
find_package (seqan3 3.0 REQUIRED)

set (CMAKE_CXX_FLAGS "${WORKAROUND_2433} ${CMAKE_CXX_FLAGS}")
```

eseiler:
Resolves #2433

### DIFF
--- a/build_system/seqan3-config.cmake
+++ b/build_system/seqan3-config.cmake
@@ -250,30 +250,33 @@ set (CXXSTD_TEST_SOURCE
     "static_assert (__cpp_concepts >= 201507);
     int main() {}")
 
-check_cxx_source_compiles ("${CXXSTD_TEST_SOURCE}" CONCEPTS_BUILTIN)
+set (SEQAN3_FEATURE_CONCEPT_FLAG_BUILTIN "")
+set (SEQAN3_FEATURE_CONCEPT_FLAG_STD20 "-std=c++20")
+set (SEQAN3_FEATURE_CONCEPT_FLAG_STD2a "-std=c++2a")
+set (SEQAN3_FEATURE_CONCEPT_FLAG_GCC_TS "-fconcepts")
 
-if (CONCEPTS_BUILTIN)
-    seqan3_config_print ("C++ Concepts support:       builtin")
-else ()
-    set (CONCEPTS_FLAG "")
+set (SEQAN3_CONCEPTS_FLAG "")
 
-    foreach (_FLAG -std=c++20 -std=c++2a -fconcepts)
-        set (CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS_SAVE} ${_FLAG}")
+foreach (_FLAG BUILTIN STD20 STD2a GCC_TS)
+    set (CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS_SAVE} ${SEQAN3_FEATURE_CONCEPT_FLAG_${_FLAG}}")
 
-        check_cxx_source_compiles ("${CXXSTD_TEST_SOURCE}" CONCEPTS_FLAG${_FLAG})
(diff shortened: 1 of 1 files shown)


Irallia (approved):
I approve with restriction that the 2 reviewer should look particularly carefully, because I have understood particularly little :/
Is there any way to build tests for such cmake flags and would such a thing make sense?

build_system/seqan3-config.cmake
```diff
@@ -250,30 +250,33 @@ set (CXXSTD_TEST_SOURCE
...
+set (SEQAN3_FEATURE_CONCEPT_FLAG_GCC_TS "-fconcepts")
 
-if (CONCEPTS_BUILTIN)
-    seqan3_config_print ("C++ Concepts support:       builtin")
-else ()
-    set (CONCEPTS_FLAG "")
-
-    foreach (_FLAG -std=c++20 -std=c++2a -fconcepts)
-        set (CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS_SAVE} ${_FLAG}")
+set (CONCEPTS_FLAG "")
```
marehr, line 258:
```suggestion
set (SEQAN3_CONCEPTS_FLAG "")
```

can we prefix this, as it is the final result?

The intermediate variables (as `CONCEPTS_FLAG${_FLAG}`) don't necessarily need to be prefixed.

build_system/seqan3-config.cmake
```diff
@@ -311,31 +314,31 @@ endif ()
...
+set (SEQAN3_FEATURE_FILESYSTEM_LIB "stdc++fs")
 
-if (C++17FS_BUILTIN)
-    seqan3_config_print ("C++ Filesystem library:     builtin")
-else ()
-    set (C++17FS_LIB "")
-
-    foreach (_LIB stdc++fs)
-        set (CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES_SAVE} ${_LIB})
+set (FILESYSTEM_FLAG "")
```
marehr, line 320: same here?

marehr (approved):
ty

PR #2439, opened 2021-03-11
[FIX] gcc-11: error: parameter ‘seed’ may not appear in this context

marehr:
```
seqan3/include/seqan3/range/views/minimiser_hash.hpp:76:49: error: parameter ‘seed’ may not appear in this context
   76 |                               seed const seed = seed{0x8F3F73B5CF1C9ADE}) const
      |                                                 ^~~~
```

### DIFF
--- a/include/seqan3/range/views/minimiser_hash.hpp
+++ b/include/seqan3/range/views/minimiser_hash.hpp
@@ -73,7 +73,7 @@ struct minimiser_hash_fn
     constexpr auto operator()(urng_t && urange,
                               shape const & shape,
                               window_size const window_size,
-                              seed const seed = seed{0x8F3F73B5CF1C9ADE}) const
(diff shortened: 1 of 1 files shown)


include/seqan3/range/views/minimiser_hash.hpp
```diff
@@ -73,7 +73,7 @@ struct minimiser_hash_fn
     constexpr auto operator()(urng_t && urange,
                               shape const & shape,
                               window_size const window_size,
-                              seed const seed = seed{0x8F3F73B5CF1C9ADE}) const
+                              seed const seed = seqan3::seed{0x8F3F73B5CF1C9ADE}) const
```
joergi-w, line 76: It's a bit unfortunate/confusing that the types have the same names as the variables here...

joergi-w (approved):

eseiler (approved):

PR #2444, opened 2021-03-13
[TEST] use EXPECT_RANGE_EQ in to_lower_test...

marehr:
Part of seqan/product_backlog#83.

### DIFF
--- a/test/unit/range/views/to_lower_test.cpp
+++ b/test/unit/range/views/to_lower_test.cpp
@@ -7,55 +7,51 @@
 
 #include <gtest/gtest.h>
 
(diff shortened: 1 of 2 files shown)


MitraDarja (approved):
LGTM, but can you explain to me, what this sv stands for that you use with every string?

marehr:
> LGTM, but can you explain to me, what this sv stands for that you use with every string?

sv = string_view :) https://en.cppreference.com/w/cpp/string/basic_string_view/operator%22%22sv

eseiler (approved):

PR #2445, opened 2021-03-13
[TEST] use EXPECT_RANGE_EQ in get_test...

marehr:
Part of seqan/product_backlog#83.

### DIFF
--- a/test/unit/range/views/get_test.cpp
+++ b/test/unit/range/views/get_test.cpp
@@ -13,15 +13,17 @@
 
 #include <seqan3/alphabet/mask/masked.hpp>
 #include <seqan3/alphabet/quality/aliases.hpp>
+#include <seqan3/core/detail/debug_stream_alphabet.hpp>
 #include <seqan3/range/concept.hpp>
 #include <seqan3/range/views/get.hpp>
 #include <seqan3/range/views/complement.hpp>
 #include <seqan3/range/views/to_char.hpp>
-#include <seqan3/range/views/to.hpp>
 #include <seqan3/range/views/zip.hpp>
 #include <seqan3/std/ranges>
+#include <seqan3/test/expect_range_eq.hpp>
 
 using seqan3::operator""_dna4;
+using seqan3::operator""_phred42;
 
 TEST(view_get, basic)
 {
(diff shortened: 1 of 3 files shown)


test/unit/range/views/get_test.cpp
```diff
@@ -74,80 +59,53 @@ TEST(view_get, advanced)
                                                    {'C'_dna4, seqan3::mask::UNMASKED},
                                                    {'G'_dna4, seqan3::mask::MASKED},
                                                    {'T'_dna4, seqan3::mask::UNMASKED}};
-    std::vector<seqan3::masked<seqan3::dna4>> functor0 = seqan3::views::get<0>(t) | seqan3::views::to<std::vector>;
-    EXPECT_EQ(cmp0, functor0);
 
-    std::vector<seqan3::phred42> cmp1{seqan3::phred42{0}, seqan3::phred42{1}, seqan3::phred42{2}, seqan3::phred42{3}};
-    std::vector<seqan3::phred42> functor1 = seqan3::views::get<1>(t) | seqan3::views::to<std::vector>;
-    EXPECT_EQ(cmp1, functor1);
+    EXPECT_RANGE_EQ(cmp0, seqan3::views::get<0>(t));
```
Irallia, line 63: As there is no `cmp00` anymore you could rename `cmp0` to `cmp` or maybe maybe use a more explanatory name as `functor_compared`.

test/unit/range/views/istreambuf_test.cpp
```diff
@@ -46,39 +45,32 @@ INSTANTIATE_TYPED_TEST_SUITE_P(iterator_fixture, iterator_fixture, test_type, );
 
 TEST(view_istreambuf, basic)
 {
+    using namespace std::literals;
```
Irallia, line 48: Is this for using this `operator""sv` thing?

Irallia (approved):
Looks very good!
However, I noticed that the tests use a lot of meaningless variables like `cmp0`, `t` and `is`. Maybe you want to change something about that?
And I have a question about the using literal stuff. Maybe you want to add a short documentation notice?

I approve either way, you can decide if you want to rename the variables or if that is too much effort.

test/unit/range/views/istreambuf_test.cpp
eseiler: Yes, we also want to do `seqan3::literals`. See https://github.com/seqan/product_backlog/issues/287#issuecomment-775223706 for some reasoning

eseiler (changes requested):
Open point about renaming, otherwise LGTM

test/unit/range/views/istreambuf_test.cpp
Irallia: Ah so we can use it for all operator""... stuff. Nice!

marehr:
> Open point about renaming, otherwise LGTM

😭 I don't want to rename everything.

test/unit/range/views/get_test.cpp
marehr: I renamed the stuff in the get test.

eseiler (approved):

PR #2451, opened 2021-03-13
[MISC] rewrite recursion of seqan3::views::deep

marehr:


SGSSGene (approved):
:+1:

eseiler (approved):
This change was because of ADL trouble with gcc-11, right?
I vaguely remember that you wanted to write up the changes at some point?

marehr:
> This change was because of ADL trouble with gcc-11, right?
> I vaguely remember that you wanted to write up the changes at some point?

This PR does not solve the gcc-11 issue. It just moves the recursion from the `operator()()` into the `impl` function (which will be called by the base class implementation of `operator()()`). That means this PR moves the recursion logic from the deep class into the range adaptor.

PR #2456, opened 2021-03-18
[NOREVIEW] Replace tmp_filename with tmp_directory

SGSSGene:
Replaces tmp_filename with tmp_directory.

The class `tmp_directory` builds on top of `sandboxed_path`. It has a much less surprising API. It provides a temporary directory and does not track any individual files.


SGSSGene:
MacOS seems to have a symlink as the temporary path. The question is: is a path that leads away from the sandbox by a symlink actually leaving the sandbox? I would say yes... But then the whole abstraction doesn't work :-(

This means I need to apply "canonical_weak" for `sandbox_directory`. Since `canonical_weak` doesn't exists for gcc7, I need to implement a more sophisticated version. Currently the version implemented only deals with ".." and "." paths.

eseiler:
> MacOS seems to have a symlink as the temporary path. The question is: is a path that leads away from the sandbox by a symlink actually leaving the sandbox? I would say yes... But then the whole abstraction doesn't work :-(
> 
> This means I need to apply "canonical_weak" for `sandbox_directory`. Since `canonical_weak` doesn't exists for gcc7, I need to implement a more sophisticated version. Currently the version implemented only deals with ".." and "." paths.

Which part doesn't work?
If it's within the ctor, you can weakly canonicalise the `filesystem::path` before you construct the `sandboxed_path`

SGSSGene:
This is dropped in favor of PR https://github.com/seqan/seqan3/pull/2753

PR #2458, opened 2021-03-21
[FIX] compression streams writing wrong format

eseiler:
Documentation will be follow-up PR

### DIFF
--- a/include/seqan3/io/detail/misc_output.hpp
+++ b/include/seqan3/io/detail/misc_output.hpp
@@ -12,11 +12,11 @@
 
 #pragma once
 
+#include <seqan3/std/filesystem>
 #include <iostream>
 #include <string>
 #include <tuple>
 
-#include <seqan3/utility/detail/exposition_only_concept.hpp>
 #ifdef SEQAN3_HAS_BZIP2
     #include <seqan3/contrib/stream/bz2_ostream.hpp>
 #endif
@@ -24,7 +24,8 @@
     #include <seqan3/contrib/stream/bgzf_ostream.hpp>
     #include <seqan3/contrib/stream/gz_ostream.hpp>
 #endif
-#include <seqan3/std/filesystem>
+#include <seqan3/io/exception.hpp>
+#include <seqan3/utility/detail/exposition_only_concept.hpp>
 
 namespace seqan3::detail
 {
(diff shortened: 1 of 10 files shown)


include/seqan3/io/detail/misc_output.hpp
```diff
@@ -46,16 +47,24 @@ inline auto make_secondary_ostream(std::basic_ostream<char_t> & primary_stream,
...
 
-    if ((extension == ".gz") || (extension == ".bgzf") || (extension == ".bam"))
+    if (extension == ".gz")
+    {
+    #ifdef SEQAN3_HAS_ZLIB
+        filename.replace_extension("");
+        return {new contrib::basic_gz_ostream<char_t>{primary_stream}, stream_deleter_default};
+    #else
+        throw file_open_error{"Trying to write a gzipped file, but no ZLIB available."};
+    #endif
```
marehr, line 57:
```suggestion
#ifdef SEQAN3_HAS_ZLIB
        filename.replace_extension("");
        return {new contrib::basic_gz_ostream<char_t>{primary_stream}, stream_deleter_default};
#else
        throw file_open_error{"Trying to write a gzipped file, but no ZLIB available."};
#endif
```

simonsasse (approved):
Looks good :)

test/unit/io/detail/misc_test.cpp
```diff
@@ -70,3 +74,52 @@ TEST(misc, valid_compression_extensions)
...
+    if constexpr (std::same_as<TypeParam, seqan3::detail::bgzf_compression>)
+    {
+        EXPECT_TRUE(TypeParam::validate_header(std::span{magic_header}));
+    }
+    else
+    {
+        EXPECT_TRUE(seqan3::detail::starts_with(magic_header, this->expected_magic_header));
+        EXPECT_FALSE(seqan3::detail::bgzf_compression::validate_header(std::span{magic_header}));
+    }
+}
```
marehr, line 125: I think it would be easier to understand if this would be 3 test cases instead of 1 typed test as they have something in common, but are also quite different.

I mean the original problem was only gzip <=> bgzf

test/unit/io/detail/misc_test.cpp
```diff
@@ -70,3 +74,52 @@ TEST(misc, valid_compression_extensions)
...
+    {
+        auto file_path = this->file_path();
+        std::ofstream filestream{file_path};
+        auto stream_ptr = seqan3::detail::make_secondary_ostream(filestream, file_path);
+        *stream_ptr << std::string(this->biggest_magic_header_size, 'a') << '\n';
+    }
+
+    std::ifstream filestream{this->file_path()};
+    std::array<char, this->biggest_magic_header_size> magic_header{};
+    std::copy_n(std::istreambuf_iterator{filestream}, this->biggest_magic_header_size, magic_header.begin());
```
marehr, line 114: Wouldn't something along the lines

```c++
std::vector<char> file_content{std::istreambuf_iterator{filestream}, std::istreambuf_iterator{}};
```

easier to understand?

test/unit/io/detail/misc_test.cpp
```diff
@@ -70,3 +74,52 @@ TEST(misc, valid_compression_extensions)
         EXPECT_TRUE(std::find(valid_compression.begin(), valid_compression.end(), "zst") != valid_compression.end());
     #endif
 }
+
+template <typename compression_t>
```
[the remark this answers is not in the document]
eseiler: Agreed

test/unit/io/detail/misc_test.cpp
```diff
@@ -70,3 +74,52 @@ TEST(misc, valid_compression_extensions)
...
+    }
+
+private:
+    std::string const tmp_file_name = "io_misc_test.txt" + compression_extension;
+    seqan3::test::tmp_filename const tmp_file{tmp_file_name.c_str()};
+};
+
+using compression_types = ::testing::Types<seqan3::detail::gz_compression,
+                                           seqan3::detail::bgzf_compression,
+                                           seqan3::detail::bz2_compression>;
```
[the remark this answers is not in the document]
eseiler: It did not but now it does

test/unit/io/detail/misc_test.cpp
eseiler: Yes, it's better when it's split.
I did some more convoluted stuff before, and it did save a lot of code to have a fixture.
Then I managed to simplify everything and did not re-evaluate the usage of a fixture

test/unit/io/detail/misc_test.cpp
eseiler: Yup.

I needed to fix some stuff in our span implementation, mainly also stripping cv qualifiers in addition to ref when checking concepts and the deduction guide for ranges.

test/unit/io/detail/misc_test.cpp
```diff
@@ -70,3 +74,52 @@ TEST(misc, valid_compression_extensions)
         EXPECT_TRUE(std::find(valid_compression.begin(), valid_compression.end(), "zst") != valid_compression.end());
     #endif
 }
```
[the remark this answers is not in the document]
eseiler: I left it in `misc_test` (didnt rename it), because it does not use `misc_input.hpp`

marehr (approved):
Very Nice! Looks much better, I didn't double-check whether the span changes do line-up with the standard.

Issue #2464, opened 2021-03-22
argument_parser: unhelpful message if the given options isn't one of enums


marehr:
Another problem is that for invalid input values it will already throw before the validator is called.
It will try to look up the invalid string in the `enumeration_names` and not find it; which results in a throw.
But this is something where we would have to change something in seqan

_Originally posted by @eseiler in https://github.com/seqan/iGenVar/issues/78#issuecomment-791992669_

## Platform

  - SeqAn version: 
  - Operating system: 
  - Compiler: 

### How to repeat the problem

Run this test in the iGenVar project. The problem is the argument `9` for option `m`, which does not exist.
```cpp
TEST_F(detect_breakends, test_unknown_argument)
{
    cli_test_result result = execute_app("iGenVar",
                                         "-j", data(default_alignment_long_reads_file_path),
                                         fasta_out_file_path,
                                         "-m 9");
    std::string expected_err
    {
        "[Error] Value parse failed for -m: Argument 9 could not be parsed as type std::string.\n"
    };
    EXPECT_EQ(result.exit_code, 65280);
    EXPECT_EQ(result.out, std::string{});
    EXPECT_EQ(result.err, expected_err);
}
```

### Expected behaviour

The Validator should handle the error message

### Actual behaviour

In the iGenVar Project, we found this unhelpful error message:
https://github.com/seqan/iGenVar/pull/91/files#r598704725

PR #2465, opened 2021-03-22
[FIX] Fix unhelpful message with help_page_message of the validator

Irallia:
Resolves https://github.com/seqan/seqan3/issues/2464

The resolving test looks like:
```cpp
TEST_F(detect_breakends, test_unknown_argument)
{
    cli_test_result result = execute_app("iGenVar",
                                         "-j", data(default_alignment_long_reads_file_path),
                                         fasta_out_file_path,
                                         "-m 9");
    std::string expected_err
    {
        // Old message: "[Error] Value parse failed for -m: Argument 9 could not be parsed as type std::string.\n"
        "[Error] Validation failed for value 9.\n"
        "Value must be one of (method name or number) [cigar_string,0,split_read,1,read_pairs,2,read_depth,3].\n"
    };
    EXPECT_EQ(result.exit_code, 65280);
    EXPECT_EQ(result.out, std::string{});
    EXPECT_EQ(result.err, expected_err);
}
```

### DIFF
--- a/include/seqan3/argument_parser/detail/format_parse.hpp
+++ b/include/seqan3/argument_parser/detail/format_parse.hpp
@@ -313,8 +313,8 @@ private:
      * \tparam option_t Must model seqan3::named_enumeration.
      * \param[out] value Stores the parsed value.
      * \param[in] in The input argument to be parsed.
-     * \returns seqan3::option_parse_result::error if `in` could not be found in the
-     *          seqan3::enumeration_names<option_t> map and otherwise seqan3::option_parse_result::success.
+     * \throws seqan3::user_input_error if `in` is not a key in seqan3::enumeration_names<option_t>.
+     * \returns seqan3::option_parse_result::success.
      */
     template <named_enumeration option_t>
     option_parse_result parse_option_value(option_t & value, std::string const & in)
@@ -322,9 +322,14 @@ private:
         auto map = seqan3::enumeration_names<option_t>;
 
         if (auto it = map.find(in); it == map.end())
-            return option_parse_result::error;
+        {
+            throw user_input_error{detail::to_string("You have chosen an invalid input value: ", in,
+                                                     ". Please use one of: ", map | std::views::keys)};
+        }
         else
+        {
             value = it->second;
+        }
 
         return option_parse_result::success;
     }
--- a/test/unit/argument_parser/format_parse_test.cpp
+++ b/test/unit/argument_parser/format_parse_test.cpp
@@ -569,78 +569,6 @@ TEST(parse_type_test, parse_error_double_option)
     EXPECT_THROW(parser2.parse(), seqan3::argument_parser_error);
 }
 
-namespace foo
-{
-enum class bar
-{
-    one,
-    two,
-    three
-};
-
-auto enumeration_names(bar)
-{
-    return std::unordered_map<std::string_view, bar>{{"one", bar::one}, {"two", bar::two}, {"three", bar::three}};
-}
-} // namespace foo
-
-namespace Other
-{
-enum class bar
-{
-    one,
-    two
-};
-} // namespace Other
-
-namespace seqan3::custom
-{
-template <>
-struct argument_parsing<Other::bar>
-{
-    static inline std::unordered_map<std::string_view, Other::bar> const enumeration_names
-    {
-        {"one", Other::bar::one}, {"two", Other::bar::two}
-    };
-};
-} // namespace seqan3::custom
-
-TEST(parse_type_test, parse_success_enum_option)
-{
-    {
-        foo::bar option_value{};
-
-        const char * argv[] = {"./argument_parser_test", "-e", "two"};
-        seqan3::argument_parser parser{"test_parser", 3, argv, seqan3::update_notifications::off};
-        parser.add_option(option_value, 'e', "enum-option", "this is an enum option.");
-
-        EXPECT_NO_THROW(parser.parse());
-        EXPECT_TRUE(option_value == foo::bar::two);
-    }
-
-    {
-        Other::bar option_value{};
-
-        const char * argv[] = {"./argument_parser_test", "-e", "two"};
-        seqan3::argument_parser parser{"test_parser", 3, argv, seqan3::update_notifications::off};
-        parser.add_option(option_value, 'e', "enum-option", "this is an enum option.");
-
-        EXPECT_NO_THROW(parser.parse());
-        EXPECT_TRUE(option_value == Other::bar::two);
-    }
-}
-
-TEST(parse_type_test, parse_error_enum_option)
-{
-    foo::bar option_value{};
-
-    const char * argv[] = {"./argument_parser_test", "-e", "four"};
-    seqan3::argument_parser parser{"test_parser", 3, argv, seqan3::update_notifications::off};
-    parser.add_option(option_value, 'e', "enum-option", "this is an enum option.");
-
-    EXPECT_THROW(parser.parse(), seqan3::argument_parser_error);
-}
-
 TEST(parse_test, too_many_arguments_error)
 {
     int option_value;
@@ -911,10 +839,10 @@ TEST(parse_test, subcommand_argument_parser_success)
     {
         const char * argv[]{"./top_level", "-f", "sub1", "-h"};
         seqan3::argument_parser top_level_parser{"top_level",
-                                          4,
-                                          argv,
-                                          seqan3::update_notifications::off,
-                                          {"sub1", "sub2"}};
+                                                 4,
+                                                 argv,
+                                                 seqan3::update_notifications::off,
+                                                 {"sub1", "sub2"}};
         top_level_parser.add_flag(flag_value, 'f', "foo", "foo bar");
 
         EXPECT_NO_THROW(top_level_parser.parse());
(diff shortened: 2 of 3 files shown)


eseiler:
I would propose

```patch
diff --git a/include/seqan3/argument_parser/detail/format_parse.hpp b/include/seqan3/argument_parser/detail/format_parse.hpp
index 753d89e4..ff14cb07 100644
--- a/include/seqan3/argument_parser/detail/format_parse.hpp
+++ b/include/seqan3/argument_parser/detail/format_parse.hpp
@@ -313,6 +313,7 @@ private:
      * \tparam option_t Must model seqan3::named_enumeration.
      * \param[out] value Stores the parsed value.
      * \param[in] in The input argument to be parsed.
+     * \throws seqan3::user_input_error if `in` is not a key in seqan3::enumeration_names<option_t>.
      * \returns seqan3::option_parse_result::error if `in` could not be found in the
      *          seqan3::enumeration_names<option_t> map and otherwise seqan3::option_parse_result::success.
      */
@@ -322,7 +323,7 @@ private:
         auto map = seqan3::enumeration_names<option_t>;
 
         if (auto it = map.find(in); it == map.end())
-            return option_parse_result::error;
+            throw user_input_error{detail::to_string("Invalid input: ", in, " Allowed are: ", map | std::views::keys)};
         else
             value = it->second;
 
diff --git a/include/seqan3/std/ranges b/include/seqan3/std/ranges
index 9ba20774..eb8b2a40 100644
--- a/include/seqan3/std/ranges
+++ b/include/seqan3/std/ranges
@@ -246,7 +246,7 @@ namespace views { using ::ranges::cpp20::views::reverse; }
 namespace views
 {
 // using ::ranges::cpp20::views::elements;
-// using ::ranges::cpp20::views::keys;
+using ::ranges::cpp20::views::keys;
 using ::ranges::cpp20::views::values;
 } // namespace views
 
```

Returning the `option_parse_result::error` does not make sense because the callee cannot handle it properly.
It does kinda replace some of the functionality that a validator may have (if it was just checking for existence), but this will work for all validators.

I think with this, we can get rid of the
`seqan3::value_list_validator{(seqan3::enumeration_names<foo::bar> | seqan3::views::get<1>)` in documentation related to `enumeration_names`, because we will implicitly check this; regardless of whether the input is a container or not.

Irallia:
> In your iGenVar testcase it would look like:
> 
> ```
> [Error] Invalid input: 9 Allowed are: [read_depth,3,read_pairs,2,split_read,1,cigar_string,0]
> ```

Yes, it does! :)

simonsasse (approved):
Looks good :)

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -313,16 +313,16 @@ class format_parse : public format_base
...
+     * \returns seqan3::option_parse_result::success.
      */
     template <named_enumeration option_t>
     option_parse_result parse_option_value(option_t & value, std::string const & in)
     {
         auto map = seqan3::enumeration_names<option_t>;
 
         if (auto it = map.find(in); it == map.end())
-            return option_parse_result::error;
+            throw user_input_error{detail::to_string("Invalid input: ", in, " Allowed are: ", map | std::views::keys)};
```
eseiler, line 325: Can you do a more polished error message?

eseiler, line 325: Can you polish the error message?

eseiler (changes requested):
I think the error message should be more user friendly 

tests are missing
the test should go somehere here, maybe as `issueXXX` test 
https://github.com/seqan/seqan3/blob/c3883c5d5503b95d4def53aa9ee53340fce63450/test/unit/argument_parser/format_parse_test.cpp#L633

the test i linked also should throw an user_input error now, can you check?

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -313,16 +313,17 @@ class format_parse : public format_base
...
     option_parse_result parse_option_value(option_t & value, std::string const & in)
     {
         auto map = seqan3::enumeration_names<option_t>;
 
         if (auto it = map.find(in); it == map.end())
-            return option_parse_result::error;
+            throw user_input_error{detail::to_string("You have choosen an invalid input value: ", in,
+                                                     " Please use one of: ", map | std::views::keys)};
         else
             value = it->second;
```
eseiler, line 328:
```suggestion
        {
            throw user_input_error{detail::to_string("You have choosen an invalid input value: ", in,
                                                     " Please use one of: ", map | std::views::keys)};
        }
        else
        {
            value = it->second;
        }
```

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -313,16 +313,17 @@ class format_parse : public format_base
...
      */
     template <named_enumeration option_t>
     option_parse_result parse_option_value(option_t & value, std::string const & in)
     {
         auto map = seqan3::enumeration_names<option_t>;
 
         if (auto it = map.find(in); it == map.end())
-            return option_parse_result::error;
+            throw user_input_error{detail::to_string("You have choosen an invalid input value: ", in,
+                                                     " Please use one of: ", map | std::views::keys)};
```
eseiler, line 326: Would it look better to have a `.` before Please, i.e. `". Please use one of: "`? Not sure.

test/unit/argument_parser/format_parse_test.cpp
```diff
@@ -641,6 +641,26 @@ TEST(parse_type_test, parse_error_enum_option)
     EXPECT_THROW(parser.parse(), seqan3::argument_parser_error);
 }
 
+TEST(parse_test, issue2464_enum_user_input_error)
+{
+    // see issue https://github.com/seqan/seqan3/issues/2464
```
eseiler, line 646:
```suggestion
// https://github.com/seqan/seqan3/issues/2464
TEST(parse_test, issue2464_enum_user_input_error)
{
```

test/unit/argument_parser/format_parse_test.cpp
```diff
@@ -641,6 +641,26 @@ TEST(parse_type_test, parse_error_enum_option)
...
+    Other::bar option_value{};
+
+    {
+        const char * argv[] = {"./argument_parser_test", "-m", "9"};
+        seqan3::argument_parser parser{"test_parser", 3, argv, seqan3::update_notifications::off};
+        parser.add_option(option_value, 'e', "enum-option", "this is an enum option.");
+        EXPECT_THROW(parser.parse(), seqan3::argument_parser_error);
```
eseiler, line 653:
```suggestion
        EXPECT_THROW(parser.parse(), seqan3::user_input_error);
```

test/unit/argument_parser/format_parse_test.cpp
```diff
@@ -641,6 +641,26 @@ TEST(parse_type_test, parse_error_enum_option)
...
+    }
+    {
+        Other::bar option_value{};
+        const char * argv[] = {"./argument_parser_test", "-m", "nine"};
+        seqan3::argument_parser parser{"test_parser", 3, argv, seqan3::update_notifications::off};
+        parser.add_option(option_value, 'e', "enum-option", "this is an enum option.");
+        EXPECT_THROW(parser.parse(), seqan3::argument_parser_error);
+    }
```
eseiler, line 661: This seems like the same test as before?

Maybe you could test the variant where you parse a vector of enums and then one of the enums has an invalid value.

test/unit/argument_parser/format_parse_test.cpp
Irallia: I tried user_input_error, but it is not thrown.. I dont understand why..

test/unit/argument_parser/format_parse_test.cpp
Irallia: Good idea!

Is there a possibility to test for the Error message aswell. I found no tests about the messages at all.

test/unit/argument_parser/format_parse_test.cpp
eseiler: Can you check what is thrown?
You can comment this line out.
Then just put `parser.parse();` on a new line.
Run the test with `--gtest_catch_exceptions=0` to prevent gtest from catching the exception.

(In general, `some_test_name --help` shows some sometimes useful googletest options)

eseiler: Might well be that it is actually an `argument_parser_error` or some other error that inherits from there, because the `user_input_error` is caught and rethrown somewhere :)

Irallia:
I changed and added tests, but not all of them work as exprected.

eseiler:
> I changed and added tests, but not all of them work as exprected.

I'll have a look and probably push the changes

eseiler:
Should work now, I also moved the enum-related tests to the bottom of the file since they were all over the file.

eseiler (approved):
I approve, but my approval doesn't really count since I was too involved :D

eseiler:
As a follow-up we should:
* Check if we can have a nice default help-page message for enums
* We currently advertise to use enums in conjuntion with `seqan3::value_list_validator`
* * We do not need this anymore for validating the input (this is what this PR does)
* * Becomes very obsolete if we manage to do the first point (Use would then only be to customise the help-page output or do some special things for validation).

test/unit/argument_parser/format_parse_test.cpp
eseiler: I don't think there is functionality for that.
You could use a try-catch block, and check the string equality of `error.what()` and the expected error message (which we then should add as a macro).

But I think we only ever check for error types.

test/unit/argument_parser/format_parse_test.cpp
```diff
@@ -1028,6 +956,118 @@ TEST(parse_test, is_option_set)
...
+}
+
+// https://github.com/seqan/seqan3/issues/2464
+TEST(parse_test, issue2464)
+{
+    using option_t = foo::bar;
+
+    seqan3::value_list_validator enum_validator{(seqan3::enumeration_names<option_t> | std::views::values)};
+    option_t option_value{};
+    std::vector<option_t> option_values{};
```
marehr, line 1038: Can we move these into the `{}` test cases?

test/unit/argument_parser/format_parse_test.cpp
```diff
@@ -1028,6 +956,118 @@ TEST(parse_test, is_option_set)
...
+
+    // Using a non-existing value of foo::bar should throw.
+    {
+        const char * argv[] = {"./argument_parser_test", "-e", "nine"};
+        seqan3::argument_parser parser{"test_parser", 3, argv, seqan3::update_notifications::off};
+        parser.add_option(option_value, 'e', "enum-option", "this is an enum option.");
+        EXPECT_THROW(parser.parse(), seqan3::user_input_error);
```
marehr, line 1045:
```suggestion
        const char * argv[] = {"./argument_parser_test", "-e", "nine"};
        
        option_t option_value{};
        
        seqan3::argument_parser parser{"test_parser", 3, argv, seqan3::update_notifications::off};
        parser.add_option(option_value, 'e', "enum-option", "this is an enum option.");
        EXPECT_THROW(parser.parse(), seqan3::user_input_error);
```

like this?

test/unit/argument_parser/format_parse_test.cpp
```diff
@@ -1028,6 +956,118 @@ TEST(parse_test, is_option_set)
...
+    }
+    {
+        const char * argv[] = {"./argument_parser_test", "-e", "one", "-e", "nine"};
+        seqan3::argument_parser parser{"test_parser", 5, argv, seqan3::update_notifications::off};
+        parser.add_option(option_values, 'e', "enum-option", "this is an enum option.");
+        EXPECT_THROW(parser.parse(), seqan3::user_input_error);
+    }
+
+    // The validator does not handle invalid inputs for enums. (No seqan3::validation_error)
```
marehr, line 1054:
```suggestion
    // Invalid inputs for enums are handled before any validator is evaluated. Thus the exception will be seqan3::user_input_error and not seqan3::validation_error.
```

@eseiler Is that okay? (Minus grammar :))

marehr (commented):
looks good :)

test/unit/argument_parser/format_parse_test.cpp
eseiler: Yup 👍

test/unit/argument_parser/format_parse_test.cpp
Irallia: Yes, I will, but isn't that code duplication?

test/unit/argument_parser/format_parse_test.cpp
```diff
@@ -1028,6 +956,129 @@ TEST(parse_test, is_option_set)
...
+    EXPECT_THROW(parser.parse(), seqan3::user_input_error);
+}
+
+// https://github.com/seqan/seqan3/issues/2464
+TEST(parse_test, issue2464)
+{
+    using option_t = foo::bar;
+
+    seqan3::value_list_validator enum_validator{(seqan3::enumeration_names<option_t> | std::views::values)};
+
```
marehr, line 1037:
```suggestion

```

test/unit/argument_parser/format_parse_test.cpp
```diff
@@ -1028,6 +956,129 @@ TEST(parse_test, is_option_set)
...
+        parser.add_option(option_values, 'e', "enum-option", "this is an enum option.");
+        EXPECT_THROW(parser.parse(), seqan3::user_input_error);
+    }
+
+    // Invalid inputs for enums are handled before any validator is evaluated.
+    // Thus the exception will be seqan3::user_input_error and not seqan3::validation_error.
+    {
+        const char * argv[] = {"./argument_parser_test", "-e", "nine"};
+
+        option_t option_value{};
```
marehr, line 1063:
```suggestion

        seqan3::value_list_validator enum_validator{(seqan3::enumeration_names<option_t> | std::views::values)};
        option_t option_value{};
```

test/unit/argument_parser/format_parse_test.cpp
```diff
@@ -1028,6 +956,129 @@ TEST(parse_test, is_option_set)
...
+        seqan3::argument_parser parser{"test_parser", 3, argv, seqan3::update_notifications::off};
+        parser.add_option(option_value, 'e', "enum-option", "this is an enum option.", seqan3::option_spec::advanced,
+                          enum_validator);
+        EXPECT_THROW(parser.parse(), seqan3::user_input_error);
+    }
+    {
+        const char * argv[] = {"./argument_parser_test", "-e", "one", "-e", "nine"};
+
+        std::vector<option_t> option_values{};
+
```
marehr, line 1074:
```suggestion

        seqan3::value_list_validator enum_validator{(seqan3::enumeration_names<option_t> | std::views::values)};
        std::vector<option_t> option_values{};

```

marehr (approved):

marehr:
@Irallia Not sure if squash merge is okay for you :)

PR #2467, opened 2021-03-23
[FIX] CodeCov having trouble to find commit SHA

eseiler:


MitraDarja (approved):

.github/workflows/ci.yml
```diff
@@ -240,6 +240,7 @@ jobs:
         uses: actions/checkout@v2
         with:
           path: seqan3
+          fetch-depth: 2
```
eseiler, line 243: This should fix the warning about not finding commit SHA (because github actions are run in detached HEAD), but the bash uploader still emitts the warning

.github/workflows/ci.yml
```diff
@@ -385,8 +386,10 @@ jobs:
 
       - name: Submit coverage build
         if: matrix.build == 'coverage'
-        shell: bash
-        run: bash <(curl -s https://codecov.io/bash) -f ./seqan3-build/seqan3_coverage -R ./seqan3 || echo 'Codecov failed to upload'
+        uses: codecov/codecov-action@v1
```
eseiler, line 389: Using the action results in no warning anymore

.github/workflows/ci.yml
```diff
@@ -385,8 +386,10 @@ jobs:
...
+        with:
+          files: ./seqan3-build/seqan3_coverage
+          root_dir: ./seqan3
```
eseiler, line 392: the options are equivalent and this action does not fail CI by default (replaces the echo)

marehr (approved):

PR #2468, opened 2021-03-23
[FIX] sandboxed_path shall not resolve symlinks

SGSSGene:
The old behaviour of `sandboxed_path` resolved symlinks if they existed on
the filesystem. This is not the indented behaviour.
`sandboxed_path` should not change behaviour depending on the physical
existence of symlinks.
This commits also adds a test preventing any regression bugs.

### DIFF
--- a/test/unit/test/sandboxed_path_test.cpp
+++ b/test/unit/test/sandboxed_path_test.cpp
@@ -296,7 +296,7 @@ TEST(sandboxed_path_symbolic_link, symbolic_link)
     }
 
     // create symlink and a sandboxed_path
-    create_directory_symlink(tmp_base_dir, tmp_dir);
+    std::filesystem::create_directory_symlink(tmp_base_dir, tmp_dir);
     sandboxed_path path{tmp_dir};
 
     // check the sandboxed path did not resolve the symlink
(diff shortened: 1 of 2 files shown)


Irallia (approved):
LGFM, thank you!

test/unit/test/sandboxed_path_test.cpp
```diff
@@ -281,3 +281,27 @@ TEST(sandboxed_path_free_operator_append, free_operator_append)
...
+    auto tmp_dir = tmp_base_dir / "seqan3_sandboxed_path_symbolic_link_test";
+
+    // if link already exists, remove it
+    if (std::filesystem::exists(tmp_dir))
+    {
+        std::filesystem::remove_all(tmp_dir);
+    }
+
+    // create symlink and a sandboxed_path
+    create_directory_symlink(tmp_base_dir, tmp_dir);
```
eseiler, line 299:
```suggestion
    std::filesystem::create_directory_symlink(tmp_base_dir, tmp_dir);
```

?

SGSSGene: Yes, I agree!

eseiler (approved):

PR #2470, opened 2021-03-23
[FIX] argument_parser/file_validator_base accept derivatives of filesystem::path

SGSSGene:
I want to hand over `sandboxed_path` to the validators. The
file_validator_base accidentally identifies `sandboxed_path` as a
useful forward_range to iterate over. It is sufficient if the
`std::filesystem::path const&` c'tor accepts the `sandboxed_path`.

### DIFF
--- a/include/seqan3/argument_parser/validators.hpp
+++ b/include/seqan3/argument_parser/validators.hpp
@@ -351,7 +351,8 @@ public:
      */
     template <std::ranges::forward_range range_type>
     //!\cond
-        requires std::convertible_to<std::ranges::range_value_t<range_type>, std::filesystem::path const &>
+        requires (std::convertible_to<std::ranges::range_value_t<range_type>, std::filesystem::path const &>
+                 && !std::convertible_to<range_type, std::filesystem::path const &>)
     //!\endcond
     void operator()(range_type const & v) const
     {


include/seqan3/argument_parser/validators.hpp
```diff
@@ -351,7 +351,8 @@ class file_validator_base
      */
     template <std::ranges::forward_range range_type>
     //!\cond
-        requires std::convertible_to<std::ranges::range_value_t<range_type>, std::filesystem::path const &>
+        requires (std::convertible_to<std::ranges::range_value_t<range_type>, std::filesystem::path const &>
+                 && !std::is_base_of_v<std::filesystem::path, range_type>)
```
marehr, line 355: Without having too much thought about this, but this seems to be a code smell and might need a different solution.

My initial thought is that the `std::filesystem::path` is one of those rare cases that have infinite range depth as each value is itself a `std::filesystem::path`.

The original intend is to allow a range of "std::filesystem::path"-like objects and convert them.

This one is tough...

simonsasse (commented):
I think my review is not important here until this is resolved, please reassign then. :)

include/seqan3/argument_parser/validators.hpp
SGSSGene: Yes, I think there is too much magic going on :/
The solution of using `!std::convertible_to<range_type, std::filesystem::path const &>` is much better than mine.

simonsasse (approved):
Look good :)

marehr (approved):

PR #2473, opened 2021-03-23
[MISC] use new CPO-way to define seqan3::assign_char_strictly_to.

marehr:
Part of https://github.com/seqan/product_backlog/issues/298.

### DIFF
--- a/include/seqan3/alphabet/concept.hpp
+++ b/include/seqan3/alphabet/concept.hpp
@@ -544,39 +544,26 @@ inline constexpr auto char_is_valid_for = detail::adl_only::char_is_valid_for_fn
 namespace seqan3::detail::adl_only
 {
 
-//!\brief Functor definition for seqan3::assign_char_strictly_to.
+//!\brief Function object definition for seqan3::assign_char_strictly_to.
 //!\ingroup alphabet
 struct assign_char_strictly_to_fn
 {
-    //!\brief Operator overload for lvalues.
-    template <typename alph_t>
-    //!\cond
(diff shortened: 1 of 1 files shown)


include/seqan3/alphabet/concept.hpp
```diff
@@ -544,39 +544,26 @@ inline constexpr auto char_is_valid_for = detail::adl_only::char_is_valid_for_fn
...
-        return seqan3::assign_char_to(r, a);
-    }
-
     //!\brief Operator overload for rvalues.
-    template <typename alph_t>
+    template <typename alphabet_t>
+    constexpr decltype(auto) operator()(seqan3::alphabet_char_t<alphabet_t> const chr, alphabet_t && alphabet) const
     //!\cond
-        requires requires (alph_t a, seqan3::alphabet_char_t<alph_t> r)
+        requires requires ()
```
MitraDarja, line 555: Can you explain me, why there are two requires here?

MitraDarja (approved):
LGTM, but could you explain me my question  below? :)

marehr:
Of course :)

The first requires is needed to say the following statement must be true to be considered. Therefore it is the keyword to say that the function has some further constraints.

 https://en.cppreference.com/w/cpp/language/function

The second requires is the same requires you would do in a concept definition.

https://en.cppreference.com/w/cpp/language/constraints#requires_expressions



Am 24. März 2021 13:10:00 MEZ schrieb MitraDarja ***@***.***>:

eseiler (approved):

PR #2488, opened 2021-03-24
[DOC] beginner friendly cmake documentation

### LINKED ISSUE #2017: [DOC] cmake default without -O3 and order of args/options in cmake call
Stefan Kurtz reported the following issue over our seqan-dev list:

1) I followed the description for the Cmake-setup and noticed in the example CMakeList.txt shown at the end of https://docs.seqan.de/seqan/3-master-user/setup.html that compiler optimization options such as -O3 is missing, which will lead to poor performance for users who do not recognize it.
  
    I now use
    ```CMAKE
    add_compile_options(-DNDEBUG -Wall -Wextra -Wunused-parameter -Wunused-variable -pedantic -Werror -O3)
    ```
    in CmakeLists.txt.

2) The cmake-setup specifies that one has to call
     cmake ../source -DCMAKE_CXX_COMPILER=/path/to/executable/g++-7

     The help pages of cmake version 3.16.3 on macOS and cmake version 3.10.2 on Linux state that options come before arguments, but only the macOS version enforces this. So on my macOS system only
     ```bash
     cmake  -DCMAKE_CXX_COMPILER=/path/to/executable/g++-7 ../source
     ```
     works. To simplify the use for macOS, it is probably better use this order in the cmake-setup section (and elsewhere) which works on both platforms.

## Platform

  - SeqAn version: 
  - Operating system: 
  - Compiler:

eaasna:
This PR resolves  #2017

- `cmake` options now come before arguments in the example bash commands
- added explanation about `Release` vs `Debug` builds

### DIFF
--- a/doc/setup/library_tests/index.md
+++ b/doc/setup/library_tests/index.md
@@ -26,11 +26,11 @@ cd /home/me/devel/seqan3-build/debug
 Invoke CMake (this is often referred to as the "Configure" step):
 
 ```bash
-cmake ../../seqan3/test/unit -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_COMPILER=g++-7
+cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_COMPILER=g++-7 ../../seqan3/test/unit
 ```
 
-The build type could be "Release", but stick to "Debug" for now. Specifying the compiler is optional; depending on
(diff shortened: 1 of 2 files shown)


simonsasse (approved):
Looks good and makes a lot of sense to me. I would have less struggles with this setup guide than I had with the old one. :)

marehr (approved):
Awesome!

PR #2492, opened 2021-03-24
[FIX,DOC] escape single quotes in single backticks

eseiler:
See https://github.com/doxygen/doxygen/issues/8441


SGSSGene (approved):
Good catch :+1:

marehr (approved):

PR #2495, opened 2021-03-24
[DOC] Add changelog entry for #2458

### LINKED ISSUE #2455: gz output writes bgzf
## Description

When writing file to gzipped output, we will write bgzf instead of gz.
Output will have, e.g., `1f 8b 08 04 00 00 00 00 00 ff 06 00 42 43 02 00 eb 58` which is a bgzf magic header, instead of the expected `1f 8b 08` magic header.

Even though both compressions are compatible, it's technically incorrect.
The decision happens here:
https://github.com/seqan/seqan3/blob/e0a1a95dabc830b767a6d5aa9087b06d40d98010/include/seqan3/io/detail/misc_output.hpp#L49
We would need a different case for just `".gz"`.

This may lead to problems downstream, because the file is not what it seems.
Parsing such files again with seqan will lead to high CPU usage because of:
https://github.com/seqan/seqan3/blob/f521b69fe98085f90bada1c3af1f6b22b87b771f/include/seqan3/contrib/stream/bgzf_stream_util.hpp#L40
(= use all threads when using bgzf)

Even if you know that we use all CPUs for bgzf, you would not expect this to happen for gzip.

eseiler:
resolves #2455

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,15 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 * Requesting the alignment without also requesting the sequence for BAM files containing empty CIGAR strings does now
   not result in erroneous parsing ([\#2418](https://github.com/seqan/seqan3/pull/2418)).
 * BAM files with 64 references are now parsed correctly ([\#2423](https://github.com/seqan/seqan3/pull/2423)).
+* Writing `gz`-compressed output no longer results in `bgzf`-compressed output. This change may have following effects
+  ([\#2458](https://github.com/seqan/seqan3/pull/2458)).
+  * A noticeable slowdown when writing `gz`-compressed content since, in contrast to `bgzf`, `gz` does not feature
+    parallelisation (magnitude depends on the application and level of parallelisation).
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -96,6 +96,15 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 * Requesting the alignment without also requesting the sequence for BAM files containing empty CIGAR strings does now
   not result in erroneous parsing ([\#2418](https://github.com/seqan/seqan3/pull/2418)).
 * BAM files with 64 references are now parsed correctly ([\#2423](https://github.com/seqan/seqan3/pull/2423)).
+* Writing `gz`-compressed output no longer results in `bgzf`-compressed output. This change may have following effects
+  ([\#2458](https://github.com/seqan/seqan3/pull/2458)).
```
Irallia, line 100:
```suggestion
  ([\#2458](https://github.com/seqan/seqan3/pull/2458)):
```

Irallia (changes requested):
LGFM
However, I would use a colon when starting a bulleted list.

Irallia (approved):

PR #2500, opened 2021-03-26
[FEATURE] implementation of a bloom filter

### LINKED ISSUE #2490: Bloom Filter
Are there any plans to add a simple bloom filter in addition to the interleaved one? This would be very useful for many applications, and simply using the interleaved bloom filter with a single bin doesn't work (efficiently) as it would be extended to 64 technical bins.

tloka:
Resolves #2490.

Implementation of a simple (not interleaved) bloom filter.  
This is a basic data structure, so I think it makes sense to include it to the library (for discussion, see #2490). It can, e.g., be used as an alternative to a k-mer index if the k-mer size needed is too high. In my application, I use the proposed implementation for the filtering of human reads.

# Implementation
The Bloom Filter was implemented based on the existing code for the Interleaved Bloom Filter with the following modifications:
 - All binning-related code was adapted or removed.
 - No agents are used for querying and counting, as the return types are native, thus I think there are no advantages when using agents in this case.
   - For querying a single hash value, use `bool bloom_filter::contains(value)`
   - For counting k-mer hits of an input range, use `std::integral bloom_filter::count(values)` (where the return value defaults to `uint16_t`).

# Notes
Some things to consider during review:
 - I didn't want to touch the IBF for now, but also tried to avoid redundant implementation of underlying types (such as `data_layout`). Thus, for now, I included the IBF to use the data types defined there, until you make a design decision to this. I think, it generally would make sense to remove this dependency / include.
 - As stated above, I decided to not use agents. If you want to have the same design as with the IBF, this probably needs to be changed (but would imply a less convenient usage, I guess).
 - I didn't implement any performance tests yet. If required, these need to be added.
- I am not sure whether it makes sense to have the BF implementation fully separated from the IBF, or if some parts of the code could actually be redesigned for common use (e.g., the hash seeds etc.).

If any changes are requested during review, I am happy to contribute these until finally merged.

### DIFF
--- a/include/seqan3/utility/bloom_filter/all.hpp
+++ b/include/seqan3/utility/bloom_filter/all.hpp
@@ -1,6 +1,6 @@
 // -----------------------------------------------------------------------------------------------------
-// Copyright (c) 2006-2020, Knut Reinert & Freie Universität Berlin
-// Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+// Copyright (c) 2006-2021, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2021, Knut Reinert & MPI für molekulare Genetik
 // This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
 // shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
 // -----------------------------------------------------------------------------------------------------
--- a/include/seqan3/utility/bloom_filter/all.hpp
+++ b/include/seqan3/utility/bloom_filter/all.hpp
@@ -11,7 +11,7 @@
  *
  * \defgroup submodule_bloom_filter Bloom Filter
  * \brief Provides seqan3:bloom_filter.
- * \ingroup search
+ * \ingroup utility
  */
 
  #pragma once
--- a/include/seqan3/utility/bloom_filter/all.hpp
+++ b/include/seqan3/utility/bloom_filter/all.hpp
@@ -9,7 +9,7 @@
  * \author Tobias Loka <tobias.loka AT hpi.de>
  * \brief Meta-header for the Bloom Filter.
  *
- * \defgroup submodule_bloom_filter Bloom Filter
+ * \defgroup utility_bloom_filter Bloom Filter
  * \brief Provides seqan3:bloom_filter.
  * \ingroup utility
  */
@@ -17,3 +17,4 @@
  #pragma once
 
  #include <seqan3/utility/bloom_filter/bloom_filter.hpp>
+ 
--- a/include/seqan3/utility/bloom_filter/all.hpp
+++ b/include/seqan3/utility/bloom_filter/all.hpp
@@ -17,4 +17,3 @@
  #pragma once
 
  #include <seqan3/utility/bloom_filter/bloom_filter.hpp>
- 
--- a/include/seqan3/utility/bloom_filter/all.hpp
+++ b/include/seqan3/utility/bloom_filter/all.hpp
@@ -1,0 +1,19 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2020, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \author Tobias Loka <tobias.loka AT hpi.de>
+ * \brief Meta-header for the Bloom Filter.
+ *
+ * \defgroup submodule_bloom_filter Bloom Filter
+ * \brief Provides seqan3:bloom_filter.
+ * \ingroup search
+ */
+
+ #pragma once
+
+ #include <seqan3/utility/bloom_filter/bloom_filter.hpp>
--- a/include/seqan3/utility/bloom_filter/bloom_filter.hpp
+++ b/include/seqan3/utility/bloom_filter/bloom_filter.hpp
@@ -1,6 +1,6 @@
 // -----------------------------------------------------------------------------------------------------
-// Copyright (c) 2006-2020, Knut Reinert & Freie Universität Berlin
-// Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+// Copyright (c) 2006-2021, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2021, Knut Reinert & MPI für molekulare Genetik
 // This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
 // shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
 // -----------------------------------------------------------------------------------------------------
--- a/include/seqan3/utility/bloom_filter/bloom_filter.hpp
+++ b/include/seqan3/utility/bloom_filter/bloom_filter.hpp
@@ -1,0 +1,385 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2020, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \author Tobias Loka <tobias.loka AT hpi.de>
+ * \brief Provides seqan3::bloom_filter.
+ */
+
+#pragma once
+
+#include <seqan3/std/algorithm>
+#include <seqan3/std/bit>
+#include <seqan3/search/dream_index/interleaved_bloom_filter.hpp>
+
+#include <sdsl/bit_vectors.hpp>
+
+#include <seqan3/core/concept/cereal.hpp>
+#include <seqan3/core/detail/strong_type.hpp>
+
+namespace seqan3
+{
+
+/*!\brief The bloom filter. A data structure that efficiently answers set-membership queries.
+ * \tparam data_layout_mode_ Indicates whether the underlying data type is compressed. See seqan3::data_layout.
+ * \implements seqan3::cerealisable
+ *
+ * \details
+ *
+ * ### Bloom Filter (BF)
+ *
+ * The [Bloom Filter](https://en.wikipedia.org/wiki/Bloom_filter) is a probabilistic data structure.
+ * A Bloom Filter can be thought of as a bitvector of length `n` and `h` hash functions and is used to determine set
+ * membership. To insert data, the data is hashed by the `h` hash functions (returning values in `[0, n)`) and the
+ * corresponding `h` positions in the bitvector are set to `1`. To query data, i.e. to determine whether the query
+ * belongs to the set the Bloom Filter was built for, the query is hashed by the same `h` hash functions and the
+ * corresponding positions are checked. If all `h` positions contain a `1`, the query is (probably) in the data set.
+ * Since the Bloom Filter has variable length, the hashing is not bijective, i.e. it may return true for a set
+ * membership query even though the query was never inserted into the Bloom Filter. Note that the Bloom Filter
+ * will always return `true` if the query was inserted, i.e. there may be false positives, but no false negatives.
+ * 
+ * ### Querying
+ * To query the Bloom Filter for a value, call seqan3::bloom_filter::contains(value) which returns
+ * true if the k-mer hash is present in the index, and false if the hash is not present.
+ * The value is a hash value of the k-mer to check membership for.
+ * 
+ * To query the Bloom Filter for a range of values, call seqan3::bloom_filter::count(values) which returns the
+ * number of k-mer hits in the Bloom Filter for the given range of values.
+ * 
+ * Please note the results are based on a heuristic data structure and with a certain probability (depending on 
+ * the selected size of the bit vector) you may receive a false positive result.
+ *
+ * Compared to the interleaved bloom filter implementation, the bloom filter does not make use of
+ * agents to determine and count membership.
+ *
+ * ### Compression
+ *
+ * The Bloom Filter can be compressed by passing `data_layout::compressed` as template argument.
+ * The compressed `seqan3::bloom_filter<seqan3::data_layout::compressed>` can only be constructed from a
+ * `seqan3::bloom_filter`, in which case the underlying bitvector is compressed.
+ * The compressed Bloom Filter is immutable, i.e. only querying is supported.
+ *
+ * ### Thread safety
+ *
+ * The Bloom Filter promises the basic thread-safety by the STL that all
+ * calls to `const` member functions are safe from multiple threads (as long as no thread calls
+ * a non-`const` member function at the same time).
+ *
+ */
+template <data_layout data_layout_mode_ = data_layout::uncompressed>
+class bloom_filter
+{
+private:
+    //!\cond
+    template <data_layout data_layout_mode>
+    friend class bloom_filter;
+    //!\endcond
+
+    //!\brief The underlying datatype to use.
+    using data_type = std::conditional_t<data_layout_mode_ == data_layout::uncompressed,
+                                         sdsl::bit_vector,
+                                         sdsl::sd_vector<>>;
+
+    //!\brief The size of the underlying bit vector in bits.
+    size_t size_in_bits{};
+    //!\brief The number of bits to shift the hash value before doing multiplicative hashing.
+    size_t hash_shift{};
+    //!\brief The number of hash functions.
+    size_t hash_funs{};
+    //!\brief The bitvector.
+    data_type data{};
+    //!\brief Precalculated seeds for multiplicative hashing. We use large irrational numbers for a uniform hashing.
+    static constexpr std::array<size_t, 5> hash_seeds{13572355802537770549ULL, // 2**64 / (e/2)
+                                                      13043817825332782213ULL, // 2**64 / sqrt(2)
+                                                      10650232656628343401ULL, // 2**64 / sqrt(3)
+                                                      16499269484942379435ULL, // 2**64 / (sqrt(5)/2)
+                                                      4893150838803335377ULL}; // 2**64 / (3*pi/5)
+
+    /*!\brief Perturbs a value and fits it into the vector.
+     * \param h The value to process.
+     * \param seed The seed to use.
+     * \returns A hashed value representing a position within the bounds of `data`.
+     * \sa https://probablydance.com/2018/06/16/
+     * \sa https://lemire.me/blog/2016/06/27
+     */
+    inline constexpr size_t hash_and_fit(size_t h, size_t const seed) const
+    {
+        h *= seed;
+        h ^= h >> hash_shift; // XOR and shift higher bits into lower bits
+        h *= 11400714819323198485ULL; // = 2^64 / golden_ration, to expand h to 64 bit range
+        // Use fastrange (integer modulo without division) if possible.
+#ifdef __SIZEOF_INT128__
+        h = static_cast<uint64_t>((static_cast<__uint128_t>(h) * static_cast<__uint128_t>(size_in_bits)) >> 64);
+#else
+        h %= size_in_bits;
+#endif
+        return h;
+    }
+
+public:
+    //!\brief Indicates whether the Bloom Filter is compressed.
+    static constexpr data_layout data_layout_mode = data_layout_mode_;
+
+    /*!\name Constructors, destructor and assignment
+     * \{
+     */
+    bloom_filter() = default; //!< Defaulted.
+    bloom_filter(bloom_filter const &) = default; //!< Defaulted.
+    bloom_filter & operator=(bloom_filter const &) = default; //!< Defaulted.
+    bloom_filter(bloom_filter &&) = default; //!< Defaulted.
+    bloom_filter & operator=(bloom_filter &&) = default; //!< Defaulted.
+    ~bloom_filter() = default; //!< Defaulted.
+
+    /*!\brief Construct an uncompressed Bloom Filter.
+     * \param size The bit vector size (in bits).
+     * \param funs The number of hash functions. Default 2. At least 1, at most 5.
+     *
+     * \attention This constructor can only be used to construct **uncompressed** Bloom Filters.
+     *
+     * \details
+     *
+     * ### Example
+     *
+     * \include test/snippet/utility/bloom_filter/bloom_filter_constructor.cpp
+     */
+    bloom_filter(seqan3::bin_size size,
+                 seqan3::hash_function_count funs = seqan3::hash_function_count{2u})
+    //!\cond
+        requires (data_layout_mode == data_layout::uncompressed)
+    //!\endcond
+    {
+        size_in_bits = size.get();
+        hash_funs = funs.get();
+
+        if (hash_funs == 0 || hash_funs > 5)
+            throw std::logic_error{"The number of hash functions must be > 0 and <= 5."};
+        if (size_in_bits == 0)
+            throw std::logic_error{"The size of a bloom filter must be > 0."};
+
+        hash_shift = std::countl_zero(size_in_bits);
+        data = sdsl::bit_vector(size_in_bits);
+    }
+
+    /*!\brief Construct a compressed Bloom Filter.
+     * \param[in] bf The uncompressed seqan3::bloom_filter.
+     *
+     * \attention This constructor can only be used to construct **compressed** Bloom Filters.
+     *
+     * \details
+     *
+     * ### Example
+     *
+     * \include test/snippet/utility/bloom_filter/bloom_filter_constructor_compressed.cpp
+     */
+    bloom_filter(bloom_filter<data_layout::uncompressed> const & bf)
+    //!\cond
+        requires (data_layout_mode == data_layout::compressed)
+    //!\endcond
+    {
+        std::tie(size_in_bits, hash_shift, hash_funs) =
+            std::tie(bf.size_in_bits, bf.hash_shift, bf.hash_funs);
+
+        data = sdsl::sd_vector<>{bf.data};
+    }
+    //!\}
+
+    /*!\name Modifiers
+     * \{
+     */
+    /*!\brief Inserts a value into the bloom filter.
+     * \param[in] value The raw numeric value to process.
+     *
+     * \attention This function is only available for **uncompressed** Bloom Filters.
+     *
+     * \details
+     *
+     * ### Example
+     *
+     * \include test/snippet/utility/bloom_filter/bloom_filter_emplace.cpp
+     */
+    void emplace(size_t const value) noexcept
+    //!\cond
+        requires (data_layout_mode == data_layout::uncompressed)
+    //!\endcond
+    {
+        for (size_t i = 0; i < hash_funs; ++i)
+        {
+            size_t idx = hash_and_fit(value, hash_seeds[i]);
+            assert(idx < data.size());
+            data[idx] = 1;
+        };
+    }
+
+    /*!\brief Remove all values from the Bloom Filter by setting all bits to 0.
+     *
+     * \attention This function is only available for **uncompressed** Bloom Filters.
+     *
+     * \details
+     *
+     * While all values are removed from the vector, the size of the Bloom Filter is not changed.
+     * 
+     * ### Example
+     *
+     * \include test/snippet/utility/bloom_filter/bloom_filter_clear.cpp
+     */
+    void clear() noexcept
+    //!\cond
+        requires (data_layout_mode == data_layout::uncompressed)
+    //!\endcond
+    {
+        sdsl::util::_set_zero_bits(data);
+    }
+    //!\}
+    
+    /*!\name Lookup
+     * \{
+     */
+    /*!\brief Check whether a value is present in the Bloom Filter.
+     * \param[in] value The raw numeric value to process.
+     *
+     * \attention This function is only available for **uncompressed** Bloom Filters.
+     *
+     * \details
+     *
+     * ### Example
+     *
+     * \include test/snippet/utility/bloom_filter/bloom_filter_contains.cpp
+     */
+    bool contains(size_t const value) const noexcept
+    {
+        for ( size_t i = 0; i < hash_funs; i++ )
+        {
+            size_t idx = hash_and_fit(value, hash_seeds[i]);
+            assert(idx < data.size());
+            if ( data[idx] == 0 )
+                return false;
+        }
+        return true;
+    }
+    //!\}
+
+    /*!\name Counting
+     * \{
+     */
+    /*!\brief Counts the occurrences in each bin for all values in a range.
+     * \tparam value_t The type of the return value. Must model std::integral. Defaults to uint16_t.
+     * \tparam value_range_t The type of the range of values. Must model std::ranges::input_range. The reference type
+     *                       must model std::unsigned_integral.
+     * \param[in] values The range of values to process.
+     *
+     * \details
+     *
+     * ### Example
+     *
+     * \include test/snippet/utility/bloom_filter/bloom_filter_count.cpp
+     *
+     * ### Thread safety
+     *
+     * Concurrent invocations of this function are thread safe.
+     */
+    template <std::ranges::range value_range_t>
+    size_t count(value_range_t && values) const noexcept
+    {
+        static_assert(std::ranges::input_range<value_range_t>, "The values must model input_range.");
+        static_assert(std::unsigned_integral<std::ranges::range_value_t<value_range_t>>,
+                      "An individual value must be an unsigned integral.");
+
+        size_t result = 0;
+
+        for (auto && value : values)
+            result += contains(value);
+
+        return result;
+    }
+    //!\}
+
+    /*!\name Capacity
+     * \{
+     */
+    /*!\brief Returns the number of hash functions used in the Bloom Filter.
+     * \returns The number of hash functions.
+     */
+    size_t hash_function_count() const noexcept
+    {
+        return hash_funs;
+    }
+
+    /*!\brief Returns the size that the Bloom Filter manages.
+     * \returns The size in bits.
+     */
+    size_t size() const noexcept
+    {
+        return size_in_bits;
+    }
+    //!\}
+
+    /*!\name Comparison operators
+     * \{
+     */
+    /*!\brief Test for equality.
+     * \param[in] lhs A `seqan3::bloom_filter`.
+     * \param[in] rhs `seqan3::bloom_filter` to compare to.
+     * \returns `true` if equal, `false` otherwise.
+     */
+    friend bool operator==(bloom_filter const & lhs, bloom_filter const & rhs) noexcept
+    {
+        return std::tie(lhs.size_in_bits, lhs.hash_shift, lhs.hash_funs, lhs.data) ==
+               std::tie(rhs.size_in_bits, rhs.hash_shift, rhs.hash_funs, rhs.data);
+    }
+
+    /*!\brief Test for inequality.
+     * \param[in] lhs A `seqan3::bloom_filter`.
+     * \param[in] rhs `seqan3::bloom_filter` to compare to.
+     * \returns `true` if unequal, `false` otherwise.
+     */
+    friend bool operator!=(bloom_filter const & lhs, bloom_filter const & rhs) noexcept
+    {
+        return !(lhs == rhs);
+    }
+    //!\}
+
+    /*!\name Access
+     * \{
+     */
+    /*!\brief Provides direct, unsafe access to the underlying data structure.
+     * \returns A reference to an SDSL bitvector.
+     *
+     * \details
+     *
+     * \noapi{The exact representation of the data is implementation defined.}
+     */
+    constexpr data_type & raw_data() noexcept
+    {
+        return data;
+    }
+
+    //!\copydoc raw_data()
+    constexpr data_type const & raw_data() const noexcept
+    {
+        return data;
+    }
+    //!\}
+
+    /*!\cond DEV
+     * \brief Serialisation support function.
+     * \tparam archive_t Type of `archive`; must satisfy seqan3::cereal_archive.
+     * \param[in] archive The archive being serialised from/to.
+     *
+     * \attention These functions are never called directly, see \ref serialisation for more details.
+     */
+    template <cereal_archive archive_t>
+    void CEREAL_SERIALIZE_FUNCTION_NAME(archive_t & archive)
+    {
+        archive(size_in_bits);
+        archive(hash_shift);
+        archive(hash_funs);
+        archive(data);
+    }
+    //!\endcond
+};
+
+} // namespace seqan3
--- a/include/seqan3/utility/bloom_filter/bloom_filter.hpp
+++ b/include/seqan3/utility/bloom_filter/bloom_filter.hpp
@@ -35,27 +35,27 @@ namespace seqan3
  * Since the Bloom Filter has variable length, the hashing is not bijective, i.e. it may return true for a set
  * membership query even though the query was never inserted into the Bloom Filter. Note that the Bloom Filter
  * will always return `true` if the query was inserted, i.e. there may be false positives, but no false negatives.
- * 
+ *
  * ### Querying
- * 
+ *
  * To query the Bloom Filter for a value, call `seqan3::bloom_filter::contains` which returns
  * true if the k-mer hash is present in the index, and false if the hash is not present.
  * The value is a hash value of the k-mer to check membership for.
- * 
+ *
  * To query the Bloom Filter for a range of values, call `seqan3::bloom_filter::count` which returns the
  * number of k-mer hits in the Bloom Filter for the given range of values.
- * 
- * Please note the results are based on a heuristic data structure and with a certain probability (depending on 
+ *
+ * Please note the results are based on a heuristic data structure and with a certain probability (depending on
  * the selected size of the bit vector) you may receive a false positive result.
  *
  * ### Differences to the Interleaved Bloom Filter (IBF)
- * 
+ *
  * While the Bloom Filter provides a single linear bit vector to represent the underlying data, the Interleaved Bloom
  * Filter provides a data structure that combines a set of Bloom Filters to enable efficient queries to multiple
- * fractions of the data. In doing so, the Interleaved Bloom Filter can not only answer the question whether a hash
+ * fractions of the data. In doing so, the Interleaved Bloom Filter can not only answer whether a hash
  * value is present in the data, but also provides information in which fraction of the data it occurs.
- * The design of the Interleaved Bloom Filter is particulary useful when the underlying data is systematically 
- * structured; for example, if each fraction of the data represents a specific set of organisms. Important 
+ * The design of the Interleaved Bloom Filter is particularly useful when the underlying data is systematically
+ * structured; for example, if each fraction of the data represents a specific set of organisms. Important
  * applications of the Interleaved Bloom Filter include taxonomic classification of sequencing data, or prefiltering
  * of specific fractions of an input data set to enable more efficient in-depth analysis.
  * The Bloom Filter, on the other hand, is useful if the database does not contain any underlying structure, or it is
@@ -78,7 +78,7 @@ namespace seqan3
  * a non-`const` member function at the same time).
  *
  * \sa seqan3::interleaved_bloom_filter
- * 
+ *
  */
 template <data_layout data_layout_mode_ = data_layout::uncompressed>
 class bloom_filter
@@ -231,7 +231,7 @@ public:
      * \details
      *
      * While all values are removed from the vector, the size of the Bloom Filter is not changed.
-     * 
+     *
      * ### Example
      *
      * \include test/snippet/utility/bloom_filter/bloom_filter_reset.cpp
@@ -244,7 +244,7 @@ public:
         sdsl::util::_set_zero_bits(data);
     }
     //!\}
-    
+
     /*!\name Lookup
      * \{
      */
--- a/include/seqan3/utility/bloom_filter/bloom_filter.hpp
+++ b/include/seqan3/utility/bloom_filter/bloom_filter.hpp
@@ -276,8 +276,7 @@ public:
     /*!\name Counting
      * \{
      */
-    /*!\brief Counts the occurrences in each bin for all values in a range.
-
+    /*!\brief Counts the occurrences for all values in a range.
      * \tparam value_range_t The type of the range of values. Must model std::ranges::input_range. The reference type
      *                       must model std::unsigned_integral.
      * \param[in] values The range of values to process.
--- a/include/seqan3/utility/bloom_filter/bloom_filter.hpp
+++ b/include/seqan3/utility/bloom_filter/bloom_filter.hpp
@@ -17,12 +17,10 @@
 namespace seqan3
 {
 
-/*!\addtogroup submodule_bloom_filter
- * \{
- */
 /*!\brief The Bloom Filter. A data structure that efficiently answers set-membership queries.
  * \tparam data_layout_mode_ Indicates whether the underlying data type is compressed. See seqan3::data_layout.
  * \implements seqan3::cerealisable
+ * \ingroup utility_bloom_filter
  *
  * \details
  *
@@ -39,6 +37,7 @@ namespace seqan3
  * will always return `true` if the query was inserted, i.e. there may be false positives, but no false negatives.
  * 
  * ### Querying
+ * 
  * To query the Bloom Filter for a value, call `seqan3::bloom_filter::contains` which returns
  * true if the k-mer hash is present in the index, and false if the hash is not present.
  * The value is a hash value of the k-mer to check membership for.
@@ -318,10 +317,10 @@ public:
         return hash_funs;
     }
 
-    /*!\brief Returns the size that the Bloom Filter manages.
-     * \returns The size in bits.
+    /*!\brief Returns the size of the underlying bitvector.
+     * \returns The size in bits of the underlying bitvector.
      */
-    size_t size() const noexcept
+    size_t bit_size() const noexcept
     {
         return size_in_bits;
     }
@@ -391,7 +390,5 @@ public:
     }
     //!\endcond
 };
-
-//!\}
 
 } // namespace seqan3
--- a/include/seqan3/utility/bloom_filter/bloom_filter.hpp
+++ b/include/seqan3/utility/bloom_filter/bloom_filter.hpp
@@ -12,19 +12,15 @@
 
 #pragma once
 
-#include <seqan3/std/algorithm>
-#include <seqan3/std/bit>
 #include <seqan3/search/dream_index/interleaved_bloom_filter.hpp>
-
-#include <sdsl/bit_vectors.hpp>
-
-#include <seqan3/core/concept/cereal.hpp>
-#include <seqan3/core/detail/strong_type.hpp>
 
 namespace seqan3
 {
 
-/*!\brief The bloom filter. A data structure that efficiently answers set-membership queries.
+/*!\addtogroup submodule_bloom_filter
+ * \{
+ */
+/*!\brief The Bloom Filter. A data structure that efficiently answers set-membership queries.
  * \tparam data_layout_mode_ Indicates whether the underlying data type is compressed. See seqan3::data_layout.
  * \implements seqan3::cerealisable
  *
@@ -43,22 +39,35 @@ namespace seqan3
  * will always return `true` if the query was inserted, i.e. there may be false positives, but no false negatives.
  * 
  * ### Querying
- * To query the Bloom Filter for a value, call seqan3::bloom_filter::contains(value) which returns
+ * To query the Bloom Filter for a value, call `seqan3::bloom_filter::contains` which returns
  * true if the k-mer hash is present in the index, and false if the hash is not present.
  * The value is a hash value of the k-mer to check membership for.
  * 
- * To query the Bloom Filter for a range of values, call seqan3::bloom_filter::count(values) which returns the
+ * To query the Bloom Filter for a range of values, call `seqan3::bloom_filter::count` which returns the
  * number of k-mer hits in the Bloom Filter for the given range of values.
  * 
  * Please note the results are based on a heuristic data structure and with a certain probability (depending on 
  * the selected size of the bit vector) you may receive a false positive result.
  *
- * Compared to the interleaved bloom filter implementation, the bloom filter does not make use of
- * agents to determine and count membership.
+ * ### Differences to the Interleaved Bloom Filter (IBF)
+ * 
+ * While the Bloom Filter provides a single linear bit vector to represent the underlying data, the Interleaved Bloom
+ * Filter provides a data structure that combines a set of Bloom Filters to enable efficient queries to multiple
+ * fractions of the data. In doing so, the Interleaved Bloom Filter can not only answer the question whether a hash
+ * value is present in the data, but also provides information in which fraction of the data it occurs.
+ * The design of the Interleaved Bloom Filter is particulary useful when the underlying data is systematically 
+ * structured; for example, if each fraction of the data represents a specific set of organisms. Important 
+ * applications of the Interleaved Bloom Filter include taxonomic classification of sequencing data, or prefiltering
+ * of specific fractions of an input data set to enable more efficient in-depth analysis.
+ * The Bloom Filter, on the other hand, is useful if the database does not contain any underlying structure, or it is
+ * not relevant for the analysis. A typical application is the removal of host sequences or different types of
+ * contamination where it is usually not of interest which part of the database was matched. In such cases, the Bloom
+ * Filter provides a lighter data structure and a more simple interface (for example, the use of agents for determining
+ * and counting membership is not necessary in this case).
  *
  * ### Compression
  *
- * The Bloom Filter can be compressed by passing `data_layout::compressed` as template argument.
+ * The Bloom Filter can be compressed by passing `seqan3::data_layout::compressed` as template argument.
  * The compressed `seqan3::bloom_filter<seqan3::data_layout::compressed>` can only be constructed from a
  * `seqan3::bloom_filter`, in which case the underlying bitvector is compressed.
  * The compressed Bloom Filter is immutable, i.e. only querying is supported.
@@ -69,6 +78,8 @@ namespace seqan3
  * calls to `const` member functions are safe from multiple threads (as long as no thread calls
  * a non-`const` member function at the same time).
  *
+ * \sa seqan3::interleaved_bloom_filter
+ * 
  */
 template <data_layout data_layout_mode_ = data_layout::uncompressed>
 class bloom_filter
@@ -190,7 +201,7 @@ public:
     /*!\name Modifiers
      * \{
      */
-    /*!\brief Inserts a value into the bloom filter.
+    /*!\brief Inserts a value into the Bloom Filter.
      * \param[in] value The raw numeric value to process.
      *
      * \attention This function is only available for **uncompressed** Bloom Filters.
@@ -224,9 +235,9 @@ public:
      * 
      * ### Example
      *
-     * \include test/snippet/utility/bloom_filter/bloom_filter_clear.cpp
+     * \include test/snippet/utility/bloom_filter/bloom_filter_reset.cpp
      */
-    void clear() noexcept
+    void reset() noexcept
     //!\cond
         requires (data_layout_mode == data_layout::uncompressed)
     //!\endcond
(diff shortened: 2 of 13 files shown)


eseiler (commented):
Just some general points.
We would need benchmarks for this, though they can just be copied from the IBF.
I can do this if you want me to

tloka:
Thanks for the quick review @eseiler.
I'll have a look after the weekend. I can also add the benchmarks, if it's basically the same as with the IBF I should be able to manage this.

include/seqan3/search/bloom_filter/bloom_filter.hpp
```diff
@@ -0,0 +1,386 @@
+// -----------------------------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
tloka: I am not sure about this.
The BF has nothing to do with the DI, except that it was derived from the IBF code. But on a conceptual level, it's not part of the DI. Thus, I would find it rather confusing to put it into the DI directory.
However, the final decision is up to you. If you want it in there, I will change it (or you can do it yourself).

include/seqan3/search/bloom_filter/bloom_filter.hpp
```diff
@@ -0,0 +1,386 @@
...
+ * `seqan3::bloom_filter`, in which case the underlying bitvector is compressed.
+ * The compressed Bloom Filter is immutable, i.e. only querying is supported.
+ *
+ * ### Thread safety
+ *
+ * The Bloom Filter promises the basic thread-safety by the STL that all
+ * calls to `const` member functions are safe from multiple threads (as long as no thread calls
+ * a non-`const` member function at the same time).
+ *
+ * Additionally, concurrent calls to `emplace` are safe.
```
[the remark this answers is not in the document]
tloka: You are right. That was too much of modified copy pasting without thinking about it.

include/seqan3/search/bloom_filter/bloom_filter.hpp
```diff
@@ -0,0 +1,386 @@
...
+    bool contains(size_t const value) const noexcept
+    {
+        for ( size_t i = 0; i < hash_funs; i++ )
+        {
+            size_t idx = hash_and_fit(value, hash_seeds[i]);
+            assert(idx < data.size());
+            if ( data[idx] == 0 )
+                return false;
+        }
+        return true;
```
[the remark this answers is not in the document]
tloka: Not sure if this is what you meant, but I checked the following with a 4GBit BF and, depending on the hit ratio, 250,000 - 1,000,000 entries:

```c++
// In bloom_filter.hpp
    bool contains(size_t const value) const noexcept
    {
        for ( size_t i = 0; i < hash_funs; i++ )
        {
            size_t idx = hash_and_fit(value, hash_seeds[i]);
            assert(idx < data.size());
            if ( data[idx] == 0 )
                return false;
        }
        return true;
    }

    bool contains_ibf_version(size_t const value) const noexcept
    {

        std::array<size_t, 5> bloom_filter_indices;
        std::memcpy(&bloom_filter_indices, &hash_seeds, sizeof(size_t) * hash_funs);

        for (size_t i = 0; i < hash_funs; ++i)
            bloom_filter_indices[i] = hash_and_fit(value, bloom_filter_indices[i]);

        for (size_t i = 0; i < hash_funs; ++i)
        {
            assert(bloom_filter_indices[i] < data.size());
            if ( data[bloom_filter_indices[i]] == 0 )
                return false;
        }

        return true;
    }
```

```c++
// In main executable
    size_t BF_SIZE = 4000000000;
    size_t BF_ENTRIES = 1000000;

    // Check use of 2-5 hash functions
    for ( size_t hash_funs = 2; hash_funs < 6; ++hash_funs)
    {
        std::cout << std::to_string(hash_funs) << " hash functions:\n";
        
        // Check hit "probability" of 0.5, 0.66, 0.75 and 1.0 
        for ( size_t prob_divider = 1; prob_divider <= 4; prob_divider++ )
        {
            std::cout << "   Hit probability of " << std::to_string(1.0f/prob_divider) << "\n";
            seqan3::bloom_filter bf(seqan3::bin_size{BF_SIZE}, seqan3::hash_function_count{hash_funs});
            for ( size_t hash = 0; hash < BF_ENTRIES; hash += 1 )
            {
                // Only emplace every 1st, 2nd, 3rd, or 4th entry, based on the prob_divider
                if ( hash % prob_divider == 0 )
                    bf.emplace(hash);
            }

            // Measure with BF version
            auto start = std::chrono::high_resolution_clock::now();

            size_t hits = 0;
            for ( size_t hash = 0; hash < BF_ENTRIES; hash += 1 )
                hits += bf.contains(hash);

            auto finish = std::chrono::high_resolution_clock::now();
            std::chrono::duration<double> elapsed = finish - start;            
            std::cout << "      BF version: Found " << hits << " hits in " << elapsed.count() << " s\n";

            // Measure with IBF version
            start = std::chrono::high_resolution_clock::now();

            hits = 0;
            for ( size_t hash = 0; hash < BF_ENTRIES; hash += 1 )
                hits += bf.contains_ibf_version(hash);

            finish = std::chrono::high_resolution_clock::now();
            elapsed = finish - start;            
            std::cout << "      IBF version: Found " << hits << " hits in " << elapsed.count() << " s\n";
        }
    }
```

Not sure whether I messed something up with the benchark (please check), but compiled in release mode I obtain the following results:
```
2 hash functions:
   Hit probability of 1.000000
      BF version: Found 1000000 hits in 0.282854 s
      IBF version: Found 1000000 hits in 0.322944 s
   Hit probability of 0.500000
      BF version: Found 500000 hits in 0.19024 s
      IBF version: Found 500000 hits in 0.292299 s
   Hit probability of 0.333333
      BF version: Found 333334 hits in 0.160255 s
      IBF version: Found 333334 hits in 0.306781 s
   Hit probability of 0.250000
      BF version: Found 250000 hits in 0.151015 s
      IBF version: Found 250000 hits in 0.358289 s
3 hash functions:
   Hit probability of 1.000000
      BF version: Found 1000000 hits in 0.351634 s
      IBF version: Found 1000000 hits in 0.568406 s
   Hit probability of 0.500000
      BF version: Found 500000 hits in 0.249056 s
      IBF version: Found 500000 hits in 0.404341 s
   Hit probability of 0.333333
      BF version: Found 333334 hits in 0.197615 s
      IBF version: Found 333334 hits in 0.389003 s
   Hit probability of 0.250000
      BF version: Found 250000 hits in 0.181299 s
      IBF version: Found 250000 hits in 0.411997 s
4 hash functions:
   Hit probability of 1.000000
      BF version: Found 1000000 hits in 0.515441 s
      IBF version: Found 1000000 hits in 0.627472 s
   Hit probability of 0.500000
      BF version: Found 500000 hits in 0.296285 s
      IBF version: Found 500000 hits in 0.509962 s
   Hit probability of 0.333333
      BF version: Found 333334 hits in 0.238565 s
      IBF version: Found 333334 hits in 0.453795 s
   Hit probability of 0.250000
      BF version: Found 250000 hits in 0.214064 s
      IBF version: Found 250000 hits in 0.44655 s
5 hash functions:
   Hit probability of 1.000000
      BF version: Found 1000000 hits in 0.590384 s
      IBF version: Found 1000000 hits in 0.845794 s
   Hit probability of 0.500000
      BF version: Found 500000 hits in 0.356562 s
      IBF version: Found 500000 hits in 0.628335 s
   Hit probability of 0.333333
      BF version: Found 333334 hits in 0.279993 s
      IBF version: Found 333334 hits in 0.533912 s
   Hit probability of 0.250000
      BF version: Found 250000 hits in 0.235909 s
      IBF version: Found 250000 hits in 0.489913 s
```

Thus, at least in this setup, the BF is **significantly** faster in every of the chosen setups. I am not sure why the difference is even that big for 100% hits, but maybe the initialization of the array + memcpy takes actually more time than "switching context"!?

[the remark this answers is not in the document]
eseiler: Seems like a missed optimization / premature "optimzation" on our part.
Thanks for testing :)

tloka:
@eseiler 
Do we need a benchmark for the `clear()` function of the BF?  
I don't think it's necessary, as we just perform a reinitialization of the underlying bitvector.  
Thus, the benchmark function currently looks like this, which doesn't make much sense to me (as the bins are missing compared to the IBF):
```c++
template <typename bf_type>
void clear_benchmark(::benchmark::State & state)
{
    auto && [ hash_values, bf ] = set_up<bf_type>(state.range(0),
                                                  state.range(1),
                                                  state.range(2));
    (void) hash_values;

    for (auto _ : state)
    {
        bf.clear();
    }

    state.counters["bfs/sec"] = hashes_per_second(1);
}
```

When knowing this, and knowing whether to change the directory structure (see https://github.com/seqan/seqan3/pull/2500#discussion_r603040223), I think I am ready to provide the requested changes.

include/seqan3/search/bloom_filter/bloom_filter.hpp
```diff
@@ -0,0 +1,386 @@
...
+     * ### Example
+     *
+     * \include test/snippet/search/bloom_filter/bloom_filter_clear.cpp
+     */
+    void clear() noexcept
+    //!\cond
+        requires (data_layout_mode == data_layout::uncompressed)
+    //!\endcond
+    {
+        data = sdsl::bit_vector(size_);
```
eseiler, line 234:
```suggestion
        data.clear();
```

The move assignment probably involves allocating new memory.
Clear should just be linear in the size.

eseiler:
> @eseiler
> Do we need a benchmark for the `clear()` function of the BF?
> I don't think it's necessary, as we just perform a reinitialization of the underlying bitvector.
> Thus, the benchmark function currently looks like this, which doesn't make much sense to me (as the bins are missing compared to the IBF):

I would add it, because it can matter if you change the underlying data structure.
It's mainly a regression test if we change something.
Also, you could use it to test whether reinitialisation or calling `.clear()` is faster (see my review comment).

> 
> ```c++
> template <typename bf_type>
> void clear_benchmark(::benchmark::State & state)
> {
>     auto && [ hash_values, bf ] = set_up<bf_type>(state.range(0),
>                                                   state.range(1),
>                                                   state.range(2));
>     (void) hash_values;
> 
>     for (auto _ : state)
>     {
>         bf.clear();
>     }
> 
>     state.counters["bfs/sec"] = hashes_per_second(1);
> }
> ```

I think we don't need the `state.counters`, because the iterations which are always reported should be enough for this benchmark. Unless it's easier to interpret when `clear` also has a custom counter.

> 
> When knowing this, and knowing whether to change the directory structure (see [#2500 (comment)](https://github.com/seqan/seqan3/pull/2500#discussion_r603040223)), I think I am ready to provide the requested changes.

Directory not decided yet :)

include/seqan3/search/bloom_filter/bloom_filter.hpp
tloka: I don't think this is appropriate, as - according to the behavior of std containers, `data.clear()` would also reset the size of the bitvector to 0. I think for the `bloom_filter::clear()`, we want to set all entries to 0 rather than changing the vector size, don't we!?  
  
Thus, maybe the best method would be using `std::fill`, which should guarantee to prevent reallcation and has also linear time:
`std::fill(data.begin(), data.end(), 0);`  
  
In this case, I don't think we need to benchmark, this should be the better option to reinitialization.

eseiler: > I don't think this is appropriate, as - according to the behavior of std containers - `data.clear()` would also reset the size of the bitvector to 0. I think for the `bloom_filter::clear()`, we want to set all entries to 0 rather than changing the vector size, don't we!?
> 
> Thus, maybe the best method would be using `std::fill`, which should guarantee to prevent reallcation and has also linear time:
> `std::fill(data.begin(), data.end(), 0);`
> 
> In this case, I don't think we need to benchmark, this should be the better option to reinitialization.

Yes, you are right, please use `std::fill`. This is an error I like to make :)

eseiler: When using `std::fill`, a benchmark would still be interesting, because the iterators of the `sdsl::bit_vector` are really slow.

tloka: > 
> 
> When using `std::fill`, a benchmark would still be interesting, because the iterators of the `sdsl::bit_vector` are really slow (I'd figure they are not slower than allocating memory, but who knows...)
> 
> Doesn't have to be very extensive, just running the benchmark you proposed once with each implementation.

You are absolutely right, the `std::fill` version is MUCH slower, even with optimizations turned on (where I'd expect it to be translated to memset). Though, I don't know if a reallocation took place in the reinit version.  
I will try a manual memset version and see whether this improves something.

eseiler: Can you try 
```cpp
    sdsl::util::_set_zero_bits(data);
```

this uses the data pointer, and should be reasonably fast

tloka: I ran some different options, after every change also running the tests to ensure that it does what it should (at least within the tested scope):
```
// manual memset; most efficient, but huge trouble if something is going wrong
memset(data.data(), 0, ((data.bit_size() + 63) >> 3));
---
/32768/2/1000            36.5 ns         36.4 ns     19418859
/1048576/2/1000          2245 ns         2240 ns       310369


// sdsl::util::_set_zero_bits; some loss of efficiency in my setup, but still doing quite well
sdsl::util::_set_zero_bits(data);
---
/32768/2/1000             201 ns          200 ns      3419712
/1048576/2/1000          6202 ns         6036 ns       111505


// std::fill; very inefficient
std::fill(data.begin(), data.end(), 0);
---
/32768/2/1000           39396 ns        39305 ns        18462
/1048576/2/1000       1259445 ns      1235777 ns          560


// Reinitialization; nearly as efficient as using memset
data = sdsl::bit_vector(size_in_bits, 0);
---
/32768/2/1000            62.7 ns         62.6 ns     11083526
/1048576/2/1000          2340 ns         2333 ns       300935


// Creating new bit vector to enforce reallocation; 
// though, the bitvector is pretty small in this example, so there is no huge impact
sdsl::bit_vector data2(size_in_bits, 0);
data = std::move(data2);
---
/32768/2/1000            63.4 ns         63.3 ns     10542822
/1048576/2/1000          2376 ns         2285 ns       309440
```

Thus, I see `memset` or `_set_zero_bits` as the best options, the first being more efficient, the second being more safe to use. If reallocation is not too much of a problem, reinitialization (as my original implementation) could also still be an option. I leave that decision to you.

eseiler: It would be ok for me to use `memset`.
We can also use `_set_zero_bits` and change it in the SDSL (which I'll do anyway).

If you need `clear` in your application, you may want the faster version, otherwise we can use the SDSL version.

tloka: I think, I would go for the `_set_zero_bits` version and wait for the SDSL changes.  
I don't feel THAT comfortable with direct memory access, especially when I am not 100% sure about the details of the underlying data structure (which is the case for the SDSL bit vectors).  
Additionally, I don't expect anyone to call `clear()` too often, it should usually have only minor impact on a global perspective. So I think it's fine to use the SDSL for now and wait for the change there such that someone who is more familiar with the details of the data structure can have a look (I guess, there is a reason why they didn't directly implement the memset there, but only put a comment).

tloka:
@eseiler 
One more question:  
When running the benchmark in Release mode, i get runtimes for `contains()` and `count()` of 0.000n.  
Is this something you would expect because of the lighter data structure than for the IBF, or is there something going wrong? When running in Debug, or for the compressed BF, I get "normal" time stamps:

```sh
# Release
emplace_benchmark<seqan3::bloom_filter<seqan3::data_layout::uncompressed>>/32768/2/1000          3489 ns         3406 ns       198670 hashes/sec=293.568M/s
emplace_benchmark<seqan3::bloom_filter<seqan3::data_layout::uncompressed>>/1048576/2/1000        3557 ns         3546 ns       190546 hashes/sec=282.008M/s
clear_benchmark<seqan3::bloom_filter<seqan3::data_layout::uncompressed>>/32768/2/1000             195 ns          191 ns      3651244
clear_benchmark<seqan3::bloom_filter<seqan3::data_layout::uncompressed>>/1048576/2/1000          5998 ns         5986 ns       114991
contains_benchmark<seqan3::bloom_filter<seqan3::data_layout::uncompressed>>/32768/2/1000        0.000 ns        0.000 ns   1000000000 hashes/sec=5.10204E/s
contains_benchmark<seqan3::bloom_filter<seqan3::data_layout::uncompressed>>/1048576/2/1000      0.000 ns        0.000 ns   1000000000 hashes/sec=4.46429E/s
contains_benchmark<seqan3::bloom_filter<seqan3::data_layout::compressed>>/32768/2/1000           7598 ns         7458 ns        89446 hashes/sec=134.077M/s
contains_benchmark<seqan3::bloom_filter<seqan3::data_layout::compressed>>/1048576/2/1000         7678 ns         7509 ns        93205 hashes/sec=133.179M/s
count_benchmark<seqan3::bloom_filter<seqan3::data_layout::uncompressed>>/32768/2/1000           0.000 ns        0.000 ns   1000000000 hashes/sec=5.15464E/s
count_benchmark<seqan3::bloom_filter<seqan3::data_layout::uncompressed>>/1048576/2/1000         0.000 ns        0.000 ns   1000000000 hashes/sec=4.42478E/s
count_benchmark<seqan3::bloom_filter<seqan3::data_layout::compressed>>/32768/2/1000              7755 ns         7743 ns        91020 hashes/sec=129.147M/s
count_benchmark<seqan3::bloom_filter<seqan3::data_layout::compressed>>/1048576/2/1000            8303 ns         8110 ns        88479 hashes/sec=123.306M/s

#Debug
emplace_benchmark<seqan3::bloom_filter<seqan3::data_layout::uncompressed>>/32768/2/1000         45440 ns        45362 ns        15197 hashes/sec=22.0447M/s
emplace_benchmark<seqan3::bloom_filter<seqan3::data_layout::uncompressed>>/1048576/2/1000       46474 ns        46401 ns        12387 hashes/sec=21.5511M/s
clear_benchmark<seqan3::bloom_filter<seqan3::data_layout::uncompressed>>/32768/2/1000            1082 ns         1053 ns       609288
clear_benchmark<seqan3::bloom_filter<seqan3::data_layout::uncompressed>>/1048576/2/1000         33501 ns        33454 ns        21025
contains_benchmark<seqan3::bloom_filter<seqan3::data_layout::uncompressed>>/32768/2/1000        22034 ns        21498 ns        32797 hashes/sec=46.5156M/s
contains_benchmark<seqan3::bloom_filter<seqan3::data_layout::uncompressed>>/1048576/2/1000      21685 ns        21367 ns        32177 hashes/sec=46.8017M/s
contains_benchmark<seqan3::bloom_filter<seqan3::data_layout::compressed>>/32768/2/1000          74374 ns        72271 ns         9463 hashes/sec=13.8368M/s
contains_benchmark<seqan3::bloom_filter<seqan3::data_layout::compressed>>/1048576/2/1000        73136 ns        72927 ns         9524 hashes/sec=13.7123M/s
count_benchmark<seqan3::bloom_filter<seqan3::data_layout::uncompressed>>/32768/2/1000           21277 ns        20829 ns        33557 hashes/sec=48.0103M/s
count_benchmark<seqan3::bloom_filter<seqan3::data_layout::uncompressed>>/1048576/2/1000         21281 ns        21241 ns        32593 hashes/sec=47.0798M/s
count_benchmark<seqan3::bloom_filter<seqan3::data_layout::compressed>>/32768/2/1000             72760 ns        72627 ns         9514 hashes/sec=13.7689M/s
count_benchmark<seqan3::bloom_filter<seqan3::data_layout::compressed>>/1048576/2/1000           74788 ns        72912 ns         8533 hashes/sec=13.7153M/s
```

If you can't tell without seeing the code, I think this can wait until the next push.

eseiler:
> @eseiler
> One more question:
> When running the benchmark in Release mode, i get runtimes for `contains()` and `count()` of 0.000n.
> Is this something you would expect because of the lighter data structure than for the IBF, or is there something going wrong? When running in Debug, or for the compressed BF, I get "normal" time stamps:
> If you can't tell without seeing the code, I think this can wait until the next push.

I can't tell without the code, but you are probably missing `benchmark::DoNotOptimize`.

`bf.clear();` may be optimised away, `benchmark::DoNotOptimize(bf.clear());` cannot

include/seqan3/search/bloom_filter/bloom_filter.hpp
eseiler: I did this benchmark https://gist.github.com/eseiler/21023865b43b6567a39fe250cf43c027
```
---------------------------------------------------------------------
Benchmark                           Time             CPU   Iterations
---------------------------------------------------------------------
set_to_zero_via_fill<1>        852937 ns       852939 ns          850
set_to_zero_via_fill<8>         11022 ns        11022 ns        62989
set_to_zero_via_fill<16>        23139 ns        23139 ns        30210
set_to_zero_via_fill<32>        49874 ns        49874 ns        14071
set_to_zero_via_fill<64>       127919 ns       127916 ns         5333

set_to_zero_via_memset<1>         999 ns          999 ns       686000
set_to_zero_via_memset<8>       11211 ns        11211 ns        61638
set_to_zero_via_memset<16>      23167 ns        23168 ns        30146
set_to_zero_via_memset<32>      49401 ns        49401 ns        14111
set_to_zero_via_memset<64>     125041 ns       125041 ns         5711

set_to_zero_via_util<1>          5798 ns         5798 ns       127256
set_to_zero_via_util<8>         46142 ns        46143 ns        15099
set_to_zero_via_util<16>        93289 ns        93289 ns         7986
set_to_zero_via_util<32>       190294 ns       190295 ns         3674
set_to_zero_via_util<64>       380571 ns       380572 ns         1692

set_to_zero_via_assign<1>        5818 ns         5818 ns       111745
set_to_zero_via_assign<8>       46895 ns        46895 ns        14814
set_to_zero_via_assign<16>      91859 ns        91859 ns         7355
set_to_zero_via_assign<32>     203704 ns       203704 ns         3648
set_to_zero_via_assign<64>     722713 ns       722716 ns          933
```

tloka: Interesting, so this looks like `std::fill` actually optimizes to `memset`, but only for word size>=8, or even only for multiples of 8 (would be interesting to see something like `set_to_zero_via_fill<30>`, but I guess it's just not possible to optimize this for non-multiples of 8 which actually would make sense). And assignment looks actually worse than util, but mainly for larger words (which would be irrelevant for bitvectors).  
So I guess, the main conclusion would be the same as derived from my benchmarks (using util as secure, or memset as most efficient solution), or is there anything I oversee!?

eseiler: We're doing https://github.com/xxsds/sdsl-lite/pull/86 now,

I guess fill can optimize for multiples of `8`, for `1` it needs to iterate (=slow).
It's also slow for, e.g., `30`.

The other versions don't care about the width, because they work directly on the data.

tloka:
@eseiler  Would you prefer me to already push the current version to check the modified code, or wait for the directory decision being made and push all at once?

eseiler:
> @eseiler Would you prefer me to already push the current version to check the modified code, or wait for the directory decision being made and push all at once?

You can push whenever you want :) 

Regarding the directory: If we had plans regarding how to integrate the BF into the search, we should put it in `search`, however we don't really have plans right now, so it should go to `utility`.

tloka:
Not sure what's with the code coverage, the bloom filter has 100% and the rest should be unaffected!? 
Everything else should be ready for review (moved the BF to utility as discussed).

eseiler:
> Not sure what's with the code coverage, the bloom filter has 100% and the rest should be unaffected!?

Can you rebase on master?

tloka:
> 
> 
> > Not sure what's with the code coverage, the bloom filter has 100% and the rest should be unaffected!?
> 
> Can you rebase on master?

Ok, that was a bit confusing as the failed coverage mentioned comparison to the original base commit. However, now it's clear to me I need to rebase before each push to make the checks pass (even if, as in this case, other code is 100% unaffected - this was why I thought it would not be necessary to rebase).

eseiler:
> > > Not sure what's with the code coverage, the bloom filter has 100% and the rest should be unaffected!?
> > 
> > 
> > Can you rebase on master?
> 
> Ok, that was a bit confusing as the failed coverage mentioned comparison to the original base commit. However, now it's clear to me I need to rebase before each push to make the checks pass (even if, as in this case, other code is 100% unaffected - this was why I thought it would not be necessary to rebase).

It usually doesn't matter, but we recently had some changes in the coverage CI that you need to rebase for.

include/seqan3/utility/bloom_filter/bloom_filter.hpp
```diff
@@ -0,0 +1,385 @@
...
+#pragma once
+
+#include <seqan3/std/algorithm>
+#include <seqan3/std/bit>
+#include <seqan3/search/dream_index/interleaved_bloom_filter.hpp>
+
+#include <sdsl/bit_vectors.hpp>
+
+#include <seqan3/core/concept/cereal.hpp>
+#include <seqan3/core/detail/strong_type.hpp>
```
eseiler, line 22:
```suggestion
#pragma once

#include <seqan3/search/dream_index/interleaved_bloom_filter.hpp>
```

This is enough for now.

Since it is `utility`, we would need to decouple this from anything that is not `seqan3/core`, i.e. we would need to move the `data_layout` from the `interleaved_bloom_filter.hpp` to .... `core`?

But this can be done later.

include/seqan3/utility/bloom_filter/bloom_filter.hpp
```diff
@@ -0,0 +1,385 @@
...
+
+namespace seqan3
+{
+
+/*!\brief The bloom filter. A data structure that efficiently answers set-membership queries.
```
eseiler, line 27:
```suggestion
/*!\brief The Bloom Filter. A data structure that efficiently answers set-membership queries.
```

Can you check that it's always **B**loom **F**ilter and **I**nterleaved **B**loom **F**ilter. While the capitalization of filter is debatable, Bloom is the name of the inventor and is hence always in upper case.

include/seqan3/utility/bloom_filter/bloom_filter.hpp
```diff
@@ -0,0 +1,385 @@
...
+ * membership. To insert data, the data is hashed by the `h` hash functions (returning values in `[0, n)`) and the
+ * corresponding `h` positions in the bitvector are set to `1`. To query data, i.e. to determine whether the query
+ * belongs to the set the Bloom Filter was built for, the query is hashed by the same `h` hash functions and the
+ * corresponding positions are checked. If all `h` positions contain a `1`, the query is (probably) in the data set.
+ * Since the Bloom Filter has variable length, the hashing is not bijective, i.e. it may return true for a set
+ * membership query even though the query was never inserted into the Bloom Filter. Note that the Bloom Filter
+ * will always return `true` if the query was inserted, i.e. there may be false positives, but no false negatives.
+ * 
+ * ### Querying
+ * To query the Bloom Filter for a value, call seqan3::bloom_filter::contains(value) which returns
```
eseiler, line 46:
```suggestion
 * To query the Bloom Filter for a value, call seqan3::bloom_filter::contains which returns
```

It doesn't link, and I think you do not name the parameter.
You can check if it works by building the documentation or visiting the preview build posted by the vercel bot (https://seqan3-git-fork-tloka-bloomfilter-seqan.vercel.app/usr/html/classseqan3_1_1bloom__filter.html in this case)

include/seqan3/utility/bloom_filter/bloom_filter.hpp
```diff
@@ -0,0 +1,385 @@
...
+ * true if the k-mer hash is present in the index, and false if the hash is not present.
+ * The value is a hash value of the k-mer to check membership for.
+ * 
+ * To query the Bloom Filter for a range of values, call seqan3::bloom_filter::count(values) which returns the
```
eseiler, line 50:
```suggestion
 * To query the Bloom Filter for a range of values, call seqan3::bloom_filter::count which returns the
```

include/seqan3/utility/bloom_filter/bloom_filter.hpp
```diff
@@ -0,0 +1,385 @@
...
+ * number of k-mer hits in the Bloom Filter for the given range of values.
+ * 
+ * Please note the results are based on a heuristic data structure and with a certain probability (depending on 
+ * the selected size of the bit vector) you may receive a false positive result.
+ *
+ * Compared to the interleaved bloom filter implementation, the bloom filter does not make use of
+ * agents to determine and count membership.
```
eseiler, line 57: Maybe we can do a `### Differences to the Interleaved Bloom Filter` or something and shortly explain what it is and what the differences are

include/seqan3/utility/bloom_filter/bloom_filter.hpp
```diff
@@ -0,0 +1,385 @@
...
+ * `seqan3::bloom_filter`, in which case the underlying bitvector is compressed.
+ * The compressed Bloom Filter is immutable, i.e. only querying is supported.
+ *
+ * ### Thread safety
+ *
+ * The Bloom Filter promises the basic thread-safety by the STL that all
+ * calls to `const` member functions are safe from multiple threads (as long as no thread calls
+ * a non-`const` member function at the same time).
+ *
+ */
```
eseiler, line 72:
```suggestion
 * a non-`const` member function at the same time).
 *
 * \sa seqan3::interleaved_bloom_filter
 */
```

include/seqan3/utility/bloom_filter/bloom_filter.hpp
```diff
@@ -0,0 +1,385 @@
...
+     *
+     * \details
+     *
+     * While all values are removed from the vector, the size of the Bloom Filter is not changed.
+     * 
+     * ### Example
+     *
+     * \include test/snippet/utility/bloom_filter/bloom_filter_clear.cpp
+     */
+    void clear() noexcept
```
eseiler, line 229: We weren't sure about the `clear` name and here it becomes a bit more obvious that this might not be the best name - because `clear` usually means that `size == 0` afterwards.

Maybe `reset` or `erase` or something along those lines would be better?

This can also be follow-up since we would change the IBF too, but I don't think `clear` is a good name.

Reset would borrow from https://en.cppreference.com/w/cpp/utility/bitset/reset

include/seqan3/utility/bloom_filter/bloom_filter.hpp
```diff
@@ -0,0 +1,385 @@
...
+     *
+     * ### Example
+     *
+     * \include test/snippet/utility/bloom_filter/bloom_filter_count.cpp
+     *
+     * ### Thread safety
+     *
+     * Concurrent invocations of this function are thread safe.
+     */
+    template <std::ranges::range value_range_t>
```
eseiler, line 284:
```suggestion
    /*!\brief Counts the occurrences for all values in a range.
     * \tparam value_range_t The type of the range of values. Must model std::ranges::input_range. The reference type
     *                       must model std::unsigned_integral.
     * \param[in] values The range of values to process.
     *
     * \details
     *
     * ### Example
     *
     * \include test/snippet/utility/bloom_filter/bloom_filter_count.cpp
     *
     * ### Thread safety
     *
     * Concurrent invocations of this function are thread safe.
     */
    template <std::ranges::range value_range_t>
```

test/performance/utility/bloom_filter/bloom_filter_benchmark.cpp
```diff
@@ -0,0 +1,126 @@
...
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+#include <benchmark/benchmark.h>
+
+#include <seqan3/range/views/to.hpp>
+#include <seqan3/range/views/zip.hpp>
+#include <seqan3/utility/bloom_filter/bloom_filter.hpp>
+#include <seqan3/test/performance/sequence_generator.hpp>
```
eseiler, line 13:
```suggestion
#include <seqan3/test/performance/sequence_generator.hpp>
#include <seqan3/utility/bloom_filter/bloom_filter.hpp>
```

test/performance/utility/bloom_filter/bloom_filter_benchmark.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+    return std::make_tuple(hash_values, bf);
+}
+
+template <typename ibf_type>
+void emplace_benchmark(::benchmark::State & state)
+{
+    auto && [ hash_values, bf ] = set_up<ibf_type>(state.range(0),
+                                                    state.range(1),
+                                                    state.range(2));
```
eseiler, line 55:
```suggestion
    auto && [ hash_values, bf ] = set_up<ibf_type>(state.range(0),
                                                   state.range(1),
                                                   state.range(2));
```

test/snippet/utility/bloom_filter/bloom_filter_clear.cpp
```diff
@@ -0,0 +1,35 @@
...
+    // Count all 5-mers of sequence2
+    seqan3::debug_stream << bf.count(sequence2 | hash_adaptor) << '\n'; // 9
+
+    // Clear the Bloom Filter
+    bf.clear();
+
+    // After clearing, no 5-mers are found
+    seqan3::debug_stream << bf.count(sequence2 | hash_adaptor) << '\n'; // 0
+
+}
```
eseiler, line 35:
```suggestion
    seqan3::debug_stream << bf.count(sequence2 | hash_adaptor) << '\n'; // 0
}
```

test/unit/utility/bloom_filter/bloom_filter_test.cpp
```diff
@@ -0,0 +1,144 @@
...
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+#include <gtest/gtest.h>
+
+#include <seqan3/utility/bloom_filter/bloom_filter.hpp>
+#include <seqan3/test/cereal.hpp>
+#include <seqan3/test/expect_range_eq.hpp>
+
```
eseiler, line 13:
```suggestion
#include <seqan3/test/cereal.hpp>
#include <seqan3/test/expect_range_eq.hpp>
#include <seqan3/utility/bloom_filter/bloom_filter.hpp>

```

test/unit/utility/bloom_filter/bloom_filter_test.cpp
```diff
@@ -0,0 +1,144 @@
...
+    seqan3::bloom_filter bf{seqan3::bin_size{1024u},
+                            seqan3::hash_function_count{2u}};
+
+    for (size_t hash : std::views::iota(0u, 64u))
+        bf.emplace(hash);
+
+    // 2. Construct either the uncompressed or compressed bloom_filter and test via hit counts
+    TypeParam bf2{bf};
+    EXPECT_EQ(bf.count(std::views::iota(0u, 64u)), 64u); // test inserted hashes
+}
```
eseiler, line 114: To decouple the tests a bit, I would

* Test `contains` as is
* Test `emplace` and use `contains` to verify
* Test `count` as is

Sorry if I suggested switching to `count` in the `emplace` test before; but otherwise we either assume that `emplace` or `count` just works, and we never explicitly test if `contains` works.

include/seqan3/utility/bloom_filter/all.hpp
```diff
@@ -0,0 +1,19 @@
...
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \author Tobias Loka <tobias.loka AT hpi.de>
+ * \brief Meta-header for the Bloom Filter.
+ *
+ * \defgroup submodule_bloom_filter Bloom Filter
+ * \brief Provides seqan3:bloom_filter.
+ * \ingroup search
```
eseiler, line 14:
```suggestion
 * \ingroup utility
```

include/seqan3/utility/bloom_filter/bloom_filter.hpp
eseiler, line 27:
```suggestion
namespace seqan3
{

/*!\addtogroup submodule_bloom_filter
 * \{
 */
/*!\brief The bloom filter. A data structure that efficiently answers set-membership queries.
```

include/seqan3/utility/bloom_filter/bloom_filter.hpp
```diff
@@ -0,0 +1,385 @@
...
+    {
+        archive(size_in_bits);
+        archive(hash_shift);
+        archive(hash_funs);
+        archive(data);
+    }
+    //!\endcond
+};
+
+} // namespace seqan3
```
eseiler, line 385:
```suggestion
};

//!\}

} // namespace seqan3

```

The `//!\}` closes the `addtogroup submodule_bloom_filter` from the top, and the newline at the end was missing

test/unit/utility/bloom_filter/bloom_filter_test.cpp
```diff
@@ -0,0 +1,144 @@
...
+    seqan3::bloom_filter bf{seqan3::bin_size{1024u}};
+    EXPECT_LE(sdsl::size_in_mega_bytes(bf.raw_data()), 0.001f);
+}
+
+TYPED_TEST(bloom_filter_test, serialisation)
+{
+    TypeParam bf{TestFixture::make_bf(seqan3::bin_size{1024u})};
+    seqan3::test::do_serialisation(bf);
+}
+
```
eseiler, line 144:
```suggestion
    seqan3::test::do_serialisation(bf);
}
```

I assume this is github's way of telling me "two new lines at end"

eseiler (changes requested):
Looks good, just minor stuff

eseiler:
And can you please push the review-changes as new commit? Makes it easier to track changes

include/seqan3/utility/bloom_filter/bloom_filter.hpp
tloka: I think `reset` makes sense. If you agree, I can already change this.

eseiler: You can change it if you want to

include/seqan3/utility/bloom_filter/bloom_filter.hpp
```diff
@@ -0,0 +1,398 @@
...
+        return true;
+    }
+    //!\}
+
+    /*!\name Counting
+     * \{
+     */
+    /*!\brief Counts the occurrences in each bin for all values in a range.
+
+     * \tparam value_range_t The type of the range of values. Must model std::ranges::input_range. The reference type
```
eseiler, line 281:
```suggestion
    /*!\brief Counts the occurrences for all values in a range.
     * \tparam value_range_t The type of the range of values. Must model std::ranges::input_range. The reference type
```

eseiler (approved):

include/seqan3/search/bloom_filter/bloom_filter.hpp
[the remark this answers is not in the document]
rrahn: I think this is not quite comparable, since the BF version drops out of the function ones it found a 0 bit, doesn't it?
But the IBF needs to compute all hashes and do the binary-and operation. One could check for a 0 binning bit vector and terminate early, as well. But this might only be useful if there are actually more queries that are not contained in any bin, I presume? Would it be possible to also try the following version? Just wondering, how this compares to the other when the hit probabilty is 100% (at least uses only one for loop and does not need to initialise the array before, but otherwise should have the same behavior, that is examining all hash functions)
```cpp
    bool contains_ibf_version_2(size_t const value) const noexcept
    {
        bool result = true;
        for ( size_t i = 0; i < hash_funs; ++i)
        {
            size_t idx = hash_and_fit(value, hash_seeds[i]);
            assert(idx < data.size());
            result &= data[idx];
            // could also be just one line:  result &= data[hash_and_fit(value, hash_seeds[i])];
        }
        return result;
    }
```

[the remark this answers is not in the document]
tloka: > I think this is not quite comparable, since the BF version drops out of the function once it found a 0 bit, doesn't it?
> But the IBF needs to compute all hashes and do the binary-and operation. 

Yes, that's why I originally changed the implementation. I thought that @eseiler actually wanted to see whether the influence of memory optimization might be higher than breaking hash value computation. 

I just realized that I actually might have messed something up in the previous benchmark where I had much higher runtimes than now ... I first thought that I maybe mistyped something in the compile parameters making it run in Debug, but even then I get very different numbers. Not sure what was going on there, probably my local setup is simply not ideal for benchmarking (at least, in terms of being comparable between different sessions) ...

Well, however, the key tendencies are still the same as before, though for smaller hit ratios, the difference in the 100% Hits is a bit lower now. Running with 10M hits (10 times more than in the benchmarks showed above, to compensate for the much smaller time stamps), I get the following results. IBF version 3 is the same as IBF version, but adding the hash calculation to the second for loop to also interrupt hash value calculation in that version (though, we are changing context again in this case, so I wouldn't expect it to be really useful in practice):

```
2 hash functions:
   Hit probability of 1.000000
      BF version:    Found 10000000 hits in 0.494258 s
      IBF version:   Found 10000000 hits in 0.564766 s
      IBF version 2: Found 10000000 hits in 0.516285 s
      IBF version 3: Found 10000000 hits in 0.515815 s
   Hit probability of 0.500000
      BF version:    Found 5000036 hits in 0.353638 s
      IBF version:   Found 5000036 hits in 0.434797 s
      IBF version 2: Found 5000036 hits in 0.491515 s
      IBF version 3: Found 5000036 hits in 0.412789 s
   Hit probability of 0.333333
      BF version:    Found 3333349 hits in 0.345686 s
      IBF version:   Found 3333349 hits in 0.395406 s
      IBF version 2: Found 3333349 hits in 0.515044 s
      IBF version 3: Found 3333349 hits in 0.379747 s
   Hit probability of 0.250000
      BF version:    Found 2500013 hits in 0.227909 s
      IBF version:   Found 2500013 hits in 0.426553 s
      IBF version 2: Found 2500013 hits in 0.551206 s
      IBF version 3: Found 2500013 hits in 0.369852 s
3 hash functions:
   Hit probability of 1.000000
      BF version:    Found 10000000 hits in 0.790269 s
      IBF version:   Found 10000000 hits in 0.776725 s
      IBF version 2: Found 10000000 hits in 0.783915 s
      IBF version 3: Found 10000000 hits in 0.831222 s
   Hit probability of 0.500000
      BF version:    Found 5000001 hits in 0.533165 s
      IBF version:   Found 5000001 hits in 0.590132 s
      IBF version 2: Found 5000001 hits in 0.748815 s
      IBF version 3: Found 5000001 hits in 0.510539 s
   Hit probability of 0.333333
      BF version:    Found 3333335 hits in 0.452182 s
      IBF version:   Found 3333335 hits in 0.624311 s
      IBF version 2: Found 3333335 hits in 0.751224 s
      IBF version 3: Found 3333335 hits in 0.484645 s
   Hit probability of 0.250000
      BF version:    Found 2500000 hits in 0.391202 s
      IBF version:   Found 2500000 hits in 0.56284 s
      IBF version 2: Found 2500000 hits in 0.775221 s
      IBF version 3: Found 2500000 hits in 0.408564 s
4 hash functions:
   Hit probability of 1.000000
      BF version:    Found 10000000 hits in 1.02324 s
      IBF version:   Found 10000000 hits in 1.11717 s
      IBF version 2: Found 10000000 hits in 0.993771 s
      IBF version 3: Found 10000000 hits in 1.06993 s
   Hit probability of 0.500000
      BF version:    Found 5000000 hits in 0.62999 s
      IBF version:   Found 5000000 hits in 0.78361 s
      IBF version 2: Found 5000000 hits in 1.00452 s
      IBF version 3: Found 5000000 hits in 0.66989 s
   Hit probability of 0.333333
      BF version:    Found 3333334 hits in 0.501071 s
      IBF version:   Found 3333334 hits in 0.686884 s
      IBF version 2: Found 3333334 hits in 0.964291 s
      IBF version 3: Found 3333334 hits in 0.526065 s
   Hit probability of 0.250000
      BF version:    Found 2500000 hits in 0.443009 s
      IBF version:   Found 2500000 hits in 0.641082 s
      IBF version 2: Found 2500000 hits in 0.891113 s
      IBF version 3: Found 2500000 hits in 0.484015 s
5 hash functions:
   Hit probability of 1.000000
      BF version:    Found 10000000 hits in 1.05258 s
      IBF version:   Found 10000000 hits in 1.27367 s
      IBF version 2: Found 10000000 hits in 1.17618 s
      IBF version 3: Found 10000000 hits in 1.28749 s
   Hit probability of 0.500000
      BF version:    Found 5000000 hits in 0.689612 s
      IBF version:   Found 5000000 hits in 1.03318 s
      IBF version 2: Found 5000000 hits in 1.19449 s
      IBF version 3: Found 5000000 hits in 0.784504 s
   Hit probability of 0.333333
      BF version:    Found 3333334 hits in 0.574388 s
      IBF version:   Found 3333334 hits in 0.845826 s
      IBF version 2: Found 3333334 hits in 1.1951 s
      IBF version 3: Found 3333334 hits in 0.606484 s
   Hit probability of 0.250000
      BF version:    Found 2500000 hits in 0.508206 s
      IBF version:   Found 2500000 hits in 0.890143 s
      IBF version 2: Found 2500000 hits in 1.22659 s
      IBF version 3: Found 2500000 hits in 0.555937 s
```

Thus, the game changer is definitely breaking the loop. I am not sure how expensive a comparison to an empty binning bit vector would be, and if we would expect to actually have this situation very often in the typical applications of the IBF, but without the loop breaking I think the difference between the versions is too small (and probably depends on hardware, system, application and data) to really care about it.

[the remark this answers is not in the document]
rrahn: What is ibf version 3? 
Just for my understanding, when I have 100 % hit probability, then I never break the loop? So the number of loop iterations should be always the same in all versions of this test case? Would be interesting to look into the generated assembly of the different versions.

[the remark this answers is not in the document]
tloka: > What is ibf version 3?

> IBF version 3 is the same as IBF version, but adding the hash calculation to the second for loop to also interrupt hash value calculation in that version

Sorry, was too lazy to put the code, I thought that short sentence would be clear enough.

```c++
    bool contains_ibf_version_3(size_t const value) const noexcept
    {
        std::array<size_t, 5> bloom_filter_indices;
        std::memcpy(&bloom_filter_indices, &hash_seeds, sizeof(size_t) * hash_funs);
        
        for (size_t i = 0; i < hash_funs; ++i)
        {
            bloom_filter_indices[i] = hash_and_fit(value, bloom_filter_indices[i]);
            assert(bloom_filter_indices[i] < data.size());
            if ( data[bloom_filter_indices[i]] == 0 )
                return false;
        }

        return true;
    }
```

> Just for my understanding, when I have 100 % hit probability, then I never break the loop? So the number of loop iterations should be always the same in all versions of this test case?

Yes. That's why there are only minor differences for 100% hit prob.

> Would be interesting to look into the generated assembly of the different versions.

I am really not familiar with this, I am not sure how to do this properly and whether I would actually be able to understand it.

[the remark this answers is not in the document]
rrahn: Ah sorry. I haven't seen that you already explained it. My bad. Ok  so basically IBF3 is like BF but with the overhead of doing unnecessary memcpy before. But ok, then this is fine enough. With the assembly I was just curious, what happens under the hood. Nevermind. But then the BF version it is :) Thanks for exploring all this.

include/seqan3/utility/bloom_filter/bloom_filter.hpp
```diff
@@ -0,0 +1,397 @@
...
+// -----------------------------------------------------------------------------------------------------
+
+/*!\file
+ * \author Tobias Loka <tobias.loka AT hpi.de>
+ * \brief Provides seqan3::bloom_filter.
+ */
+
+#pragma once
+
+#include <seqan3/search/dream_index/interleaved_bloom_filter.hpp>
```
marehr, line 15:
```suggestion
#include <seqan3/search/dream_index/interleaved_bloom_filter.hpp>
```

It would be desirable if we do not depend on search at all.

The question is how much we need to untangle to just include the entities we need.

Entities seen so far:
* seqan3::data_layout
* seqan3::bin_size
* seqan3::hash_function_count
* sdsl::bit_vector

include/seqan3/utility/bloom_filter/bloom_filter.hpp
```diff
@@ -0,0 +1,397 @@
...
+ * membership. To insert data, the data is hashed by the `h` hash functions (returning values in `[0, n)`) and the
+ * corresponding `h` positions in the bitvector are set to `1`. To query data, i.e. to determine whether the query
+ * belongs to the set the Bloom Filter was built for, the query is hashed by the same `h` hash functions and the
+ * corresponding positions are checked. If all `h` positions contain a `1`, the query is (probably) in the data set.
+ * Since the Bloom Filter has variable length, the hashing is not bijective, i.e. it may return true for a set
+ * membership query even though the query was never inserted into the Bloom Filter. Note that the Bloom Filter
+ * will always return `true` if the query was inserted, i.e. there may be false positives, but no false negatives.
+ * 
+ * ### Querying
+ * To query the Bloom Filter for a value, call `seqan3::bloom_filter::contains` which returns
```
marehr, line 42:
```suggestion
 * ### Querying
 *
 * To query the Bloom Filter for a value, call `seqan3::bloom_filter::contains` which returns
```

include/seqan3/utility/bloom_filter/bloom_filter.hpp
marehr: The seqan3 entities could be moved into this utility/dream_index folder (one file per entity) making utility independent again.

include/seqan3/utility/bloom_filter/bloom_filter.hpp
```diff
@@ -0,0 +1,397 @@
...
+     */
+    size_t hash_function_count() const noexcept
+    {
+        return hash_funs;
+    }
+
+    /*!\brief Returns the size that the Bloom Filter manages.
+     * \returns The size in bits.
+     */
+    size_t size() const noexcept
```
marehr, line 324:
```suggestion
    size_t bit_size() const noexcept
```

`size` is kind of a reserved keyword for the number of elements.

Can we use the same description as in:

https://github.com/seqan/seqan3/blob/c7d7f1a60d8946a09cc8ca3405550a88c79d8b12/include/seqan3/search/dream_index/interleaved_bloom_filter.hpp#L479-L485

?

test/snippet/utility/bloom_filter/bloom_filter_contains.cpp
```diff
@@ -0,0 +1,15 @@
...
+
+int main()
+{
+    seqan3::bloom_filter bf{seqan3::bin_size{1024u}};
+    bf.emplace(126);
+    bf.emplace(712);
+    bf.emplace(237);
+
+    // Query the Bloom Filter. Note that there may be false positive results!
+    // A return of `true` indicates the (probable) presence of the query in the Bloom Filter.
```
marehr, line 12:
```suggestion
    // A return of `true` indicates the (probable) presence of the query in the Bloom Filter.
```

Isn't the real strong point of a bloom filter is that a `false` definitely means not contained? And a `true` might actually be a `false`?

test/unit/utility/bloom_filter/bloom_filter_test.cpp
```diff
@@ -0,0 +1,144 @@
...
+    EXPECT_TRUE(std::is_move_assignable_v<TypeParam>);
+    EXPECT_TRUE(std::is_destructible_v<TypeParam>);
+
+    // num hash functions defaults to two
+    TypeParam bf1{TestFixture::make_bf(seqan3::bin_size{1024u})};
+    TypeParam bf2{TestFixture::make_bf(seqan3::bin_size{1024u},
+                                       seqan3::hash_function_count{2u})};
+    EXPECT_TRUE(bf1 == bf2);
+
+     // bin_size parameter is too small
```
marehr, line 48:
```suggestion
    // bin_size parameter is too small
```

test/snippet/utility/bloom_filter/bloom_filter_count.cpp
```diff
@@ -0,0 +1,29 @@
...
+{
+    using namespace seqan3::literals;
+
+    seqan3::bloom_filter bf{seqan3::bin_size{8192u},
+                            seqan3::hash_function_count{2u}};
+
+    auto const sequence1 = "ACTGACTGACTGATC"_dna4;
+    auto const sequence2 = "GTGACTGACTGACTCG"_dna4;
+    auto const sequence3 = "AAAAAAACGATCGACA"_dna4;
+    auto hash_adaptor = seqan3::views::kmer_hash(seqan3::ungapped{5u});
```
marehr, line 17:
```suggestion
    auto 5mers = seqan3::views::kmer_hash(seqan3::ungapped{5u});
```

test/snippet/utility/bloom_filter/bloom_filter_count.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+    // Insert all 5-mers of sequence1
+    for (auto && value : sequence1 | hash_adaptor)
```
marehr, line 20:
```suggestion
    for (auto && value : sequence1 | 5mers)
```

test/snippet/utility/bloom_filter/bloom_filter_count.cpp
```diff
@@ -0,0 +1,29 @@
...
+        bf.emplace(value);
+
+    // Insert all 5-mers of sequence3
+    for (auto && value : sequence3 | hash_adaptor)
```
marehr, line 24:
```suggestion
    for (auto && value : sequence3 | 5mers)
```

test/snippet/utility/bloom_filter/bloom_filter_reset.cpp
```diff
@@ -0,0 +1,34 @@
...
+{
+    using namespace seqan3::literals;
+
+    seqan3::bloom_filter bf{seqan3::bin_size{8192u},
+                            seqan3::hash_function_count{2u}};
+
+    auto const sequence1 = "ACTGACTGACTGATC"_dna4;
+    auto const sequence2 = "GTGACTGACTGACTCG"_dna4;
+    auto const sequence3 = "AAAAAAACGATCGACA"_dna4;
+    auto hash_adaptor = seqan3::views::kmer_hash(seqan3::ungapped{5u});
```
marehr, line 16:
```suggestion
    auto 5mers = seqan3::views::kmer_hash(seqan3::ungapped{5u});
```

test/snippet/utility/bloom_filter/bloom_filter_reset.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+    // Insert all 5-mers of sequence1
+    for (auto && value : sequence1 | hash_adaptor)
```
marehr, line 19:
```suggestion
    for (auto && value : sequence1 | 5mers)
```

test/snippet/utility/bloom_filter/bloom_filter_reset.cpp
```diff
@@ -0,0 +1,34 @@
...
+        bf.emplace(value);
+
+    // Insert all 5-mers of sequence3
+    for (auto && value : sequence3 | hash_adaptor)
```
marehr, line 23:
```suggestion
    for (auto && value : sequence3 | 5mers)
```

test/snippet/utility/bloom_filter/bloom_filter_reset.cpp
```diff
@@ -0,0 +1,34 @@
...
+        bf.emplace(value);
+
+    // Count all 5-mers of sequence2
+    seqan3::debug_stream << bf.count(sequence2 | hash_adaptor) << '\n'; // 9
```
marehr, line 27:
```suggestion
    seqan3::debug_stream << bf.count(sequence2 | 5mers) << '\n'; // 9
```

test/snippet/utility/bloom_filter/bloom_filter_reset.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+    // Reset the Bloom Filter
+    bf.reset();
+
+    // After reset, no 5-mers are found
+    seqan3::debug_stream << bf.count(sequence2 | hash_adaptor) << '\n'; // 0
```
marehr, line 33:
```suggestion
    seqan3::debug_stream << bf.count(sequence2 | 5mers) << '\n'; // 0
```

marehr (commented):
I'm REAAAALLLLY SORRY! I wanted to do this review for a really long time.

The PR itself looks fantastic, and I think it just some small things left.

include/seqan3/utility/bloom_filter/bloom_filter.hpp
```diff
@@ -0,0 +1,397 @@
...
+    {
+        archive(size_in_bits);
+        archive(hash_shift);
+        archive(hash_funs);
+        archive(data);
+    }
+    //!\endcond
+};
+
+//!\}
```
marehr, line 395:
```suggestion
//!\}
```

include/seqan3/utility/bloom_filter/bloom_filter.hpp
```diff
@@ -0,0 +1,397 @@
...
+
+namespace seqan3
+{
+
+/*!\addtogroup submodule_bloom_filter
+ * \{
+ */
+/*!\brief The Bloom Filter. A data structure that efficiently answers set-membership queries.
+ * \tparam data_layout_mode_ Indicates whether the underlying data type is compressed. See seqan3::data_layout.
+ * \implements seqan3::cerealisable
```
marehr, line 25:
```suggestion
/*!\brief The Bloom Filter. A data structure that efficiently answers set-membership queries.
 * \tparam data_layout_mode_ Indicates whether the underlying data type is compressed. See seqan3::data_layout.
 * \implements seqan3::cerealisable
 * \ingroup utility_bloom_filter
```

marehr (commented):
Small thing that I just saw.

tloka:
> 
> 
> I'm REAAAALLLLY SORRY! I wanted to do this review for a really long time.

No problem, I just wanted to konw whether your plans to merge this changed in general, as I would then need to finish a full and clean integration into my own code as I am already using the functionality. No time pressure, though, I just needed to know whether you still plan to merge it at all.

> 
> The PR itself looks fantastic, and I think it just some small things left.

Great, thanks for the quick review now. I will have a look within the next days!

include/seqan3/utility/bloom_filter/bloom_filter.hpp
tloka: I think for `seqan3::data_layout` and `seqan3::hash_function_count`, this would make sense. For `seqan3::bin_size` on the other hand, I have a bit of the feeling that I am "abusing" this here anyways (as we don't have any bins...).  
I am not sure how important it is for your design to work with `strong_type`, but in principle we could either just use a `size_t` here, or define its own `strong_type`!? I guess, the `strong_type` is more important when having more than one numeric parameter, isn't it?

```
bloom_filter(size_t size, // should this be renamed to bit_size as well?
             seqan3::hash_function_count funs = seqan3::hash_function_count{2u})
```

test/snippet/utility/bloom_filter/bloom_filter_contains.cpp
tloka: To be honest, I think I just copied this from the IBF snippet.  
Probably, we could make this a bit more clear. However, it's some more comment lines:

```c++
// Query the Bloom Filter. 
// A return of `false` guarantees the query not being present in the Bloom Filter.
// A return of `true` indicates the (probable) presence of the query in the Bloom Filter. 
// Note that there may be false positive results!
```

include/seqan3/utility/bloom_filter/bloom_filter.hpp
tloka: According to the change below, this should be removed, right? (The suggestion shows replacing it by the same thing...)

tloka:
@marehr Just a quick ping, I had some questions to your comments that I need answered to finish this (not urgent - just to not lose sight of the PR).

include/seqan3/utility/bloom_filter/bloom_filter.hpp
eseiler: This would need to be removed. At the beginning of the file `\addtogroup` -> `\ingroup` is suggested, and then the `//!\}` would be dangling.

test/snippet/utility/bloom_filter/bloom_filter_contains.cpp
eseiler: I think it is ok to be more clear about it.

eseiler (commented):
I answered the questions, I think this is good to go when all open remarks are resolved

test/snippet/utility/bloom_filter/bloom_filter_count.cpp
tloka: I will change this to
```c++
auto kmers = seqan3::views::kmer_hash(seqan3::ungapped{5u});
```
to not start the variable name with a number.

tloka:
I made the requested changes except https://github.com/seqan/seqan3/pull/2500#discussion_r650934476 (as discussed).  
I am happy to deliver these changes once you let me know how exactly you want it to be solved (either still here, or in a separate PR).

marehr (approved):

marehr:
Okay let's do the split of the header later.

tloka:
Thank You! ❤️‍🔥

PR #2502, opened 2021-03-29
[MISC] Move `parse_cigar` from `format_sam_base` into `cigar`.

Irallia:
Resolves first part of https://github.com/seqan/iGenVar/issues/85

We decided to move `parse_cigar` from `seqan3::detail::format_sam_base` into `seqan3/io/sam_file/detail/cigar.hpp`.
This is still detail, but now you normaly just need to include one header. Apon this, it is also easier to get everything to do with `cigar` out of `detail` later.

### DIFF
--- a/include/seqan3/io/sam_file/detail/cigar.hpp
+++ b/include/seqan3/io/sam_file/detail/cigar.hpp
@@ -27,7 +27,6 @@
 #include <seqan3/utility/char_operations/predicate.hpp>
 #include <seqan3/utility/tuple/concept.hpp>
 
-#include <range/v3/view/zip.hpp>
 
 namespace seqan3::detail
 {
--- a/include/seqan3/io/sam_file/detail/cigar.hpp
+++ b/include/seqan3/io/sam_file/detail/cigar.hpp
@@ -27,7 +27,6 @@
 #include <seqan3/utility/char_operations/predicate.hpp>
 #include <seqan3/utility/tuple/concept.hpp>
 
-
 namespace seqan3::detail
 {
 //!\brief Comparator that is able two compare two views
--- a/include/seqan3/io/sam_file/detail/cigar.hpp
+++ b/include/seqan3/io/sam_file/detail/cigar.hpp
@@ -24,6 +24,7 @@
 #include <seqan3/range/views/single_pass_input.hpp>
 #include <seqan3/range/views/take_until.hpp>
 #include <seqan3/range/views/zip.hpp>
+#include <seqan3/utility/char_operations/predicate.hpp>
 #include <seqan3/utility/tuple/concept.hpp>
 
 #include <range/v3/view/zip.hpp>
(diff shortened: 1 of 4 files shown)


include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -95,11 +95,6 @@ class format_sam_base
                                  header_type      & header,
                                  ref_seqs_type    & /*tag*/);
 
-    static void update_alignment_lengths(int32_t & ref_length,
```
simonsasse, line 98: This static void does not appear in the `include/seqan3/io/sam_file/detail/cigar.hpp`. Is it just unnecessary ? :)

Irallia: As the function is not in a `class` anymore, I think the `static` is not necessary.

include/seqan3/io/sam_file/detail/cigar.hpp
```diff
@@ -24,6 +24,7 @@
 #include <seqan3/range/views/single_pass_input.hpp>
 #include <seqan3/range/views/take_until.hpp>
 #include <seqan3/range/views/zip.hpp>
+#include <seqan3/utility/char_operations/predicate.hpp>
 #include <seqan3/utility/tuple/concept.hpp>
 
 #include <range/v3/view/zip.hpp>
```
eseiler, line 30:
```suggestion
```

simonsasse (approved):
Thanks for the explanation. Looks good :)

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -43,7 +43,7 @@
 #include <seqan3/range/views/to.hpp>
 #include <seqan3/range/views/to_char.hpp>
 #include <seqan3/range/views/zip.hpp>
-#include <seqan3/utility/char_operations/predicate.hpp>
+// #include <seqan3/utility/char_operations/predicate.hpp>
```
eseiler, line 46:
```suggestion
```

include/seqan3/io/sam_file/detail/cigar.hpp
```diff
@@ -24,9 +24,9 @@
 #include <seqan3/range/views/single_pass_input.hpp>
 #include <seqan3/range/views/take_until.hpp>
 #include <seqan3/range/views/zip.hpp>
+#include <seqan3/utility/char_operations/predicate.hpp>
 #include <seqan3/utility/tuple/concept.hpp>
```
eseiler, line 29:
```suggestion
```

eseiler (approved):

include/seqan3/io/sam_file/detail/format_sam_base.hpp
eseiler: A `static` member of a class would mean that the member (in this case a member function) is independent of the actual object. Maybe think of it as "there is one (global) instance of this member" and the member exists without any instance of the class existing.

It also means that you do not need an instance of the `format_sam_base` (or any derived) class to invoke this function. You can use `format_sam_base::update_alignment_lengths` instead of `some_instance.update_alignment_lengths`.

PR #2498, opened 2021-03-25
[DOC] Setup guide renewal✨

eseiler:
Resolves https://github.com/seqan/product_backlog/issues/139

Todo:

* [x] Put CMake stuff into an external_project test und use as snippet

PR #2506, opened 2021-04-09
[DOC] Hide BUG_WORKAROUND in Tutorial

h-2:
I realised that some WORKAROUND macros are leaking into our tutorial which they shouldn't. This small change should prevent that.

### DIFF
--- a/test/documentation/seqan3_doxygen_cfg.in
+++ b/test/documentation/seqan3_doxygen_cfg.in
@@ -52,6 +52,7 @@ PREDEFINED             = "CEREAL_SERIALIZE_FUNCTION_NAME=serialize" \
                          "CEREAL_LOAD_MINIMAL_FUNCTION_NAME=load_minimal" \
                          "CEREAL_SAVE_MINIMAL_FUNCTION_NAME=save_minimal" \
                          "SEQAN3_DOXYGEN_ONLY(x)= x" \
+                         "SEQAN3_WORKAROUND_GCC_93983=0" \
                          "${SEQAN3_DOXYGEN_PREDEFINED_NDEBUG}"
 
 EXPAND_AS_DEFINED      = SEQAN3_CPO_IMPL SEQAN3_CPO_OVERLOAD_BODY SEQAN3_CPO_OVERLOAD SEQAN3_DEPRECATED_310


h-2:
Doesn't work. I will investigate this weekend.

eseiler:
I don't know if there is an elegant way, I had some trouble with nested snippets lately, I did it like this: 
https://github.com/seqan/seqan3/pull/2498/files#diff-388fce22ef166b2979f7784016dbe1d0ab7917e6bafdbc6a0540afb2efa6c74cR141

i.e. https://www.doxygen.nl/manual/commands.html#cmddontinclude and similar.

eseiler:
(You pushed this branch directly on the repo instead of your fork)

h-2:
There seems to be some Doxygen bug somewhere, because by default, the `#if` should not appear in the generated documentation:
https://www.doxygen.nl/manual/preprocessing.html

I have tried adding `SEQAN3_WORKAROUND_GCC_93983` to `EXPAND_AS_DEFINED`, this should definitely replace all occurrences of `SEQAN3_WORKAROUND_GCC_93983` with `0` -- independent of whether the `#if` is evaluated or not. But this also doesn't happen. 🤷🏻 

> (You pushed this branch directly on the repo instead of your fork)

Sorry abou this. I will fix it tonight when I have a computer with git and internet access (😑).

eseiler:
I think `snippet` and `include` are verbatim and don't preprocess, and also can't be forced to do so. The macro expansion stuff is for stuff like
```
//!\brief Something.
#if SEQAN3_WORKAROUND_GCC_93983
void foo();
#else
void bar();
#endif
```
within the parsed codebase.

eseiler:
Maybe we should handle workarounds in tutorials differently (i.e. more verbose), because with the workaround the code is usually more complex than it needs to be, and without the workaround might not work if you use the wrong compiler.
Could be something like promoting the non-workaround version but having a collapsible showing the workaround for some compiler versions.

h-2:
> Maybe we should handle workarounds in tutorials differently (i.e. more verbose), because with the workaround the code is usually more complex than it needs to be, and without the workaround it might not work if you use the wrong compiler.
> Could be something like promoting the non-workaround version but having a collapsible showing the workaround for some compiler versions.

I think that workarounds for compiler bugs that are fixed in the currently shipped version of the compiler should not be shown at all. They just make the code difficult to read. OTOH I think that for bugs that affect a current version we should maybe only show the workaround. -> Get rid of the macros inside the tutorial for good?

I am currently looking through things, because I might be giving a presentation or workshop on this at deCODE.

h-2:
> Which means that duplicating the snippet and having the #if/#else/#endif outside the snippet markers might work.

This also does not seem to work. It always picks the first block corresponding to the tags and ignores any #ifs  .

Would you be OK if I just add the `#if SEQAN3_WORKAROUND_...` before the whole file like we currently have the `#if SEQAN3_WITH_CEREAL`? We officially only support the latest patch release of each compiler anyway.

eseiler:
Ah well, if it's some workaround that only applies to some old minor version, you can probably just remove it. I thought it was one of these workarounds that apply to every minor version.

marehr:
SEQAN3_WORKAROUND_GCC_93983 can be removed in the snippets, as gcc 10.2 is out for a while now :)

There might be other workarounds in our snippets that can't be removed.

h-2:
> SEQAN3_WORKAROUND_GCC_93983 can be removed in the snippets, as gcc 10.2 is out for a while now :)

Yep, that's what I meant, too!

> There might be other workarounds in our snippets that can't be removed.

I will go through the tutorial (but not the other doc) and check all the WORKAROUNDS there. I will then make a new PR, so we can close this here now and also delete the branch.

PR #2520, opened 2021-04-19
[INFRA] API-Stability cronjob

eseiler:
Resolves https://github.com/seqan/product_backlog/issues/310

Open:
- [x] Question: Make a seqan/api team that gets assigned, core might be too much, and we can always change the team members. Answer: No assignment
- [x] Should other compilers also run? The issue would be updated for each job failure, but the URL would always be the same, so it is possible. Answer: Nope
- [x] Remove `if` in workflow.
- [x] Remove `matrix` stuff
- [x] only one restore key
- [x] "[CRON]" in title
- [x] Link to api-stab README

See https://github.com/eseiler/seqan3/issues/7 for example

### DIFF
--- a/.github/ISSUE_TEMPLATE/api_cron_template.md
+++ b/.github/ISSUE_TEMPLATE/api_cron_template.md
@@ -1,0 +1,8 @@
+---
+title: '[CRON] API Stability Failure'
+labels: 'bug'
+---
+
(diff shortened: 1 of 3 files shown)


Irallia (approved):
LGFM

marehr (approved):
Very nice! Thank you!

PR #2526, opened 2021-04-19
[MISC] Replace `seqan3::views::join` with `seqan3::views::join_with`

SGSSGene:
- Deprecates `seqan3::views::join` use `seqan3::views::join_with` or `std::views::join` instead.
- Adds `seqan3::views::join_with`
- Move files from seqan3/ranges/views/join.hpp to seqan3/utility/views/join_with.hpp

resolves https://github.com/seqan/product_backlog/issues/351

### DIFF
--- a/include/seqan3/io/sequence_file/format_fasta.hpp
+++ b/include/seqan3/io/sequence_file/format_fasta.hpp
@@ -35,7 +35,6 @@
 #include <seqan3/io/sequence_file/output_options.hpp>
 #include <seqan3/io/stream/detail/fast_ostreambuf_iterator.hpp>
 #include <seqan3/range/detail/misc.hpp>
-#include <seqan3/range/views/join.hpp>
 #include <seqan3/range/views/take.hpp>
 #include <seqan3/range/views/take_exactly.hpp>
 #include <seqan3/range/views/take_line.hpp>
--- a/include/seqan3/range/container/concatenated_sequences.hpp
+++ b/include/seqan3/range/container/concatenated_sequences.hpp
@@ -21,7 +21,6 @@
 #include <seqan3/range/container/concept.hpp>
 #include <seqan3/range/detail/random_access_iterator.hpp>
 #include <seqan3/range/views/as_const.hpp>
-#include <seqan3/range/views/join.hpp>
 #include <seqan3/utility/views/repeat_n.hpp>
 #include <seqan3/utility/views/slice.hpp>
 
@@ -1111,7 +1110,7 @@ public:
             return begin() + pos_as_num;
 
         /* TODO implement views::flat_repeat_n that is like
-         *  views::repeat_n(value, count) | views::join | ranges::views::bounded;
+         *  views::repeat_n(value, count) | std::views::join | ranges::views::bounded;
          * but preserves random access and size.
          *
          * then do
--- a/include/seqan3/utility/views/join_with.hpp
+++ b/include/seqan3/utility/views/join_with.hpp
@@ -7,12 +7,29 @@
 
 /*!\file
  * \author Hannes Hauswedell <hannes.hauswedell AT fu-berlin.de>
- * \brief [DEPRECATED] Provides seqan3::views::join.
+ * \brief Provides seqan3::views::join_with.
  */
 
 #pragma once
 
-#include <seqan3/utility/views/join_with.hpp>
+#include <range/v3/view/join.hpp>
 
-SEQAN3_DEPRECATED_HEADER(
-   "This header is deprecated and will be removed in SeqAn-3.1.0; Please #include <seqan3/utility/views/join_with.hpp> instead.")
+#include <seqan3/core/platform.hpp>
+
+namespace seqan3::views
+{
+
+/*!\brief A join view.
+ * \ingroup views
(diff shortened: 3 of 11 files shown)


test/performance/range/views/view_translate_2D_benchmark.cpp
```diff
@@ -11,22 +11,22 @@
...
-#include <seqan/sequence.h>
 #include <seqan/seq_io.h>
+#include <seqan/sequence.h>
 #include <seqan/translation.h>
 #endif
 
 // Tags used to define the benchmark type
 struct baseline_tag{}; // Baseline where view is applied and only iterating the output range is benchmarked
-struct translate_tag{}; // Benchmark view_translate followed by seqan3::views::join
+struct translate_tag{}; // Benchmark view_translate followed by std::views::join
```
MitraDarja, line 29: Question, we don't want to benchmark with join_with?

MitraDarja (changes requested):
LGTM, just one question. :)

SGSSGene:
Rename file to "join_with_view.hpp"

test/performance/range/views/view_translate_2D_benchmark.cpp
SGSSGene: I think we are comparing the pipe `seqan3::views::translate | std::views::join` against `seqan3::views::translate_join`.
`std::views::join_with` should only be used if you need a delimiter, otherwise `std::views::join` does the job.

MitraDarja (approved):

include/seqan3/range/container/concatenated_sequences.hpp
```diff
@@ -21,9 +21,9 @@
 #include <seqan3/range/container/concept.hpp>
 #include <seqan3/range/detail/random_access_iterator.hpp>
 #include <seqan3/range/views/as_const.hpp>
-#include <seqan3/range/views/join.hpp>
 #include <seqan3/range/views/repeat_n.hpp>
 #include <seqan3/range/views/slice.hpp>
+#include <seqan3/utility/views/join_with.hpp>
```
marehr, line 26: Is this even used in this class?

include/seqan3/io/sequence_file/format_fasta.hpp
```diff
@@ -34,14 +34,14 @@
...
-#include <seqan3/range/views/join.hpp>
-#include <seqan3/range/views/to_char.hpp>
 #include <seqan3/range/views/take.hpp>
 #include <seqan3/range/views/take_exactly.hpp>
 #include <seqan3/range/views/take_line.hpp>
 #include <seqan3/range/views/take_until.hpp>
+#include <seqan3/range/views/to_char.hpp>
 #include <seqan3/utility/char_operations/predicate.hpp>
 #include <seqan3/utility/detail/type_name_as_string.hpp>
+#include <seqan3/utility/views/join_with.hpp>
```
marehr, line 44: Is this even used in this class?

include/seqan3/range/views/join.hpp
```diff
@@ -7,20 +7,12 @@
...
-#include <seqan3/std/ranges>
-
-namespace seqan3::views
-{
-
-//TODO reimplement me and rename to join_with
-using ::ranges::views::join;
-
-} // namespace seqan3::views
+SEQAN3_DEPRECATED_HEADER(
```
marehr, line 17:
```suggestion

namespace seqan3::views
{

/*!\brief A join view.
 * \ingroup views
 * \deprecated Please use std::views::join or seqan3::views::join_with (if a separator is needed)
 */
SEQAN3_DEPRECATED_310 inline constexpr auto join = ::ranges::views::join;

} // namespace seqan3::views

SEQAN3_DEPRECATED_HEADER(
```

include/seqan3/utility/views/join_with.hpp
```diff
@@ -0,0 +1,35 @@
...
+#include <seqan3/core/platform.hpp>
+
+namespace seqan3::views
+{
+
+/*!\brief A join view.
+ * \ingroup views
+ * \deprecated Please use std::views::join or seqan3::views::join_with (if a separator is needed)
+ */
+SEQAN3_DEPRECATED_310 inline constexpr auto join = ::ranges::views::join;
```
marehr, line 26: AHH!

This should be only in the old header :)

marehr (commented):
thank you!

include/seqan3/range/views/join.hpp
marehr: On second thought, your initial solution might be better in this case.

* as we can deprecate `seqan3/range/views/join.hpp` and
* don't use that internally any more
*  `seqan3/range/views/all.hpp` includes `seqan3/utility/views/join_with.hpp` and thus exposes the deprecated `seqan3::views::join`

include/seqan3/range/container/concatenated_sequences.hpp
SGSSGene: I was cold. Needed my computer to heat up ;-)
Yes, not needed at all, no clue why I added it.

marehr (approved):

marehr:
@SGSSGene I'll rebase this for you :)

PR #2527, opened 2021-04-19
[FIX] gcc-11: seqan3::views::deep: std::transform::views can't handle const &&

marehr:
This will make seqan3 compile with gcc-11 (my build is from 20210408).

### DIFF
--- a/include/seqan3/range/views/deep.hpp
+++ b/include/seqan3/range/views/deep.hpp
@@ -130,40 +130,69 @@ public:
      * \tparam    urng_t                Type of the underlying range.
      * \tparam    underlying_adaptor_t_ Same as underlying_adaptor_t with possibly different cref qualification.
      * \param[in] urange                The view's underlying range.
-     * \param[in] adap                  The stored adaptor unwrapped by the base-classes explode()-member.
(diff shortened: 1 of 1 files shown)


include/seqan3/range/views/deep.hpp
```diff
@@ -130,40 +130,69 @@ class deep : public detail::adaptor_base<deep<underlying_adaptor_t>, underlying_
...
-        //             return inner_most_range | adap;
-        //         });
-        //         // ...
-        //     });
-        // });
-
-        // Note: we avoid the (unqualified) pipe|-notation and use function notation here, to avoid problems with ADL.
-        if constexpr (std::ranges::input_range<std::ranges::range_reference_t<urng_t>>)
+        static_assert(std::same_as<underlying_adaptor_t_, underlying_adaptor_t>,
+                      "Internal stored deep-adaptor does not match!");
```
eseiler, line 140:
```suggestion
                      "Internally stored deep-adaptor does not match!");
```

SGSSGene (approved):
Your applied changes look great!

eseiler (approved):

Issue #1201, opened 2019-08-02
BAM unknown reference name


h-2:
Maybe we are too strict?

I am seeing this on random files from the internet:
```
terminate called after throwing an instance of 'seqan3::format_error'
  what():  Unknown reference name 'chr1' found in BAM file header (header.ref_ids():[]).
```
e.g. http://hgdownload.cse.ucsc.edu/goldenPath/hg19/encodeDCC/wgEncodeUwRepliSeq/wgEncodeUwRepliSeqBg02esG1bAlnRep1.bam

```cpp
#include <seqan3/core/debug_stream.hpp>
#include <seqan3/io/sam_file/input.hpp>

int main()
{
    seqan3::sam_file_input fin{"wgEncodeUwRepliSeqBg02esG1bAlnRep1.bam"};
    seqan3::debug_stream << fin.header().format_version << '\n';
}
```

smehringer:
I guess we are. samtools only emits a warning, maybe we should too? Or just debug assert this and ignore in release.

rrahn:
It depends. Is the format specification such that it is an ill-formed BAM file if the reference names are not present? 
I guess, what we need is kind of a property that allows people to parse strict (ensuring the format is correct) and a less strict versions so users can disable the strict parsing on purpose. That could be a simple strong typed bool flag added to the constructor.

h-2:
> I guess, what we need is kind of a property that allows people to parse strict (ensuring the format is correct) and a less strict versions so users can disable the strict parsing on purpose. That could be a simple strong typed bool flag added to the constructor.

We discussed this idea previously in the context of FastA and came to the conclusion that it is not feasible in general. Enforcing a strict format is much more expensive and enforcing it conditionally would mean adding a huge amount of if-checks that are also too expensive. We decided then that it is not important to reject wrong files as long as we accept all correct files. If someone wants strict format handling, they can add an extra format for this.

rrahn:
> Enforcing a strict format is much more expensive and enforcing it conditionally would mean adding a huge amount of if-checks that are also too expensive. We decided then that it is not important to reject wrong files as long as we accept all correct files. If someone wants strict format handling, they can add an extra format for this

[...]

PR #2536, opened 2021-04-20
[FIX] not accepting BAM with empty header

eseiler:
Resolves https://github.com/seqan/seqan3/issues/1201

### DIFF
--- a/include/seqan3/io/sam_file/format_bam.hpp
+++ b/include/seqan3/io/sam_file/format_bam.hpp
@@ -333,34 +333,42 @@ inline void format_bam::read_alignment_record(stream_type & stream,
         if (!std::ranges::equal(stream_view | views::take_exactly_or_throw(4), std::string_view{"BAM\1"}))
             throw format_error{"File is not in BAM format."};
 
-        int32_t tmp32{};
-        read_field(stream_view, tmp32);
+        int32_t l_text{}; // length of header text including \0 character
+        int32_t n_ref{}; // number of reference sequences
+        int32_t l_name{}; // 1 + length of reference name including \0 character
+        int32_t l_ref{}; // length of reference sequence
 
-        bool const header_present{tmp32 > 0};
+        read_field(stream_view, l_text);
 
-        if (header_present) // header text is present
-            read_header(stream_view | views::take_exactly_or_throw(tmp32), header, ref_seqs);
+        if (l_text > 0) // header text is present
+            read_header(stream_view | views::take_exactly_or_throw(l_text), header, ref_seqs);
 
-        int32_t n_ref;
         read_field(stream_view, n_ref);
 
         for (int32_t ref_idx = 0; ref_idx < n_ref; ++ref_idx)
         {
-            read_field(stream_view, tmp32); // l_name (length of reference name including \0 character)
+            read_field(stream_view, l_name);
 
-            string_buffer.resize(tmp32 - 1);
-            std::ranges::copy_n(std::ranges::begin(stream_view), tmp32 - 1, string_buffer.data()); // copy without \0 character
+            string_buffer.resize(l_name - 1);
+            std::ranges::copy_n(std::ranges::begin(stream_view), l_name - 1, string_buffer.data()); // copy without \0 character
             std::ranges::next(std::ranges::begin(stream_view)); // skip \0 character
 
-            read_field(stream_view, tmp32); // l_ref (length of reference sequence)
+            read_field(stream_view, l_ref);
 
-            // If there was no header text, we parse reference sequences block as header information
-            if (!header_present)
+            if constexpr (detail::decays_to_ignore_v<ref_seqs_type>) // no reference information given
             {
-                header.ref_ids().push_back(string_buffer);
-                header.ref_id_info.emplace_back(tmp32, "");
-                header.ref_dict[(header.ref_ids())[(header.ref_ids()).size() - 1]] = (header.ref_ids()).size() - 1;
-                continue;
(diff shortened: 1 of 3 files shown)


MitraDarja (approved):

marehr:
CI fails with:

```cpp
../../../seqan3/include/seqan3/io/sam_file/format_bam.hpp:360:17: error: passing ‘const std::vector<std::__cxx11::basic_string<char> >’ as ‘this’ argument discards qualifiers [-fpermissive]
                 header.ref_ids().push_back(string_buffer);
                 ^~~~~~
```

include/seqan3/io/sam_file/format_bam.hpp
```diff
@@ -352,6 +354,15 @@ inline void format_bam::read_alignment_record(stream_type & stream,
 
             read_field(stream_view, tmp32); // l_ref (length of reference sequence)
 
+            // If there was no header text, we parse reference sequences block as header information
+            if (!header_present)
+            {
+                header.ref_ids().push_back(string_buffer);
+                header.ref_id_info.emplace_back(tmp32, "");
+                header.ref_dict[(header.ref_ids())[(header.ref_ids()).size() - 1]] = (header.ref_ids()).size() - 1;
```
marehr, line 362:
```suggestion
                auto & reference_ids = header.ref_ids();

                // put the length of the reference sequence into ref_id_info
                header.ref_id_info.emplace_back(tmp32, "");
                // put the reference name into reference_ids
                reference_ids.push_back(string_buffer);
                // assign the reference name an ascending reference id (starts at index 0).
                header.ref_dict.try_emplace(reference_ids.back(), reference_ids.size() - 1);
```

not sure if this is more readable?


```cpp
header.ref_id_info.emplace_back(tmp32, "");
```
I don't understand the data structure (std::tuple<int, std::string>) and the help page wasn't helpful either.

eseiler: I used `emplace` because it may move and duplicate entries are not allowed.

The `int` is the reference length, `string` is all tags that might be set in the header (`DS`, etc.)

include/seqan3/io/sam_file/format_bam.hpp
```diff
@@ -336,7 +336,9 @@ inline void format_bam::read_alignment_record(stream_type & stream,
         int32_t tmp32{};
         read_field(stream_view, tmp32);
```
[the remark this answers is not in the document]
eseiler: I added variables for each field

include/seqan3/io/sam_file/format_bam.hpp
```diff
@@ -333,24 +333,43 @@ inline void format_bam::read_alignment_record(stream_type & stream,
...
 
         for (int32_t ref_idx = 0; ref_idx < n_ref; ++ref_idx)
         {
-            read_field(stream_view, tmp32); // l_name (length of reference name including \0 character)
+            read_field(stream_view, l_name);
 
-            string_buffer.resize(tmp32 - 1);
-            std::ranges::copy_n(std::ranges::begin(stream_view), tmp32 - 1, string_buffer.data()); // copy without \0 character
+            string_buffer.resize(l_name - 1);
+            std::ranges::copy_n(std::ranges::begin(stream_view), l_name - 1, string_buffer.data()); // copy without \0 character
```
eseiler, line 353: In theory, this should copy the `\0` character, because l_name is 1 + length including \0.

include/seqan3/io/sam_file/format_bam.hpp
```diff
@@ -333,24 +333,43 @@ inline void format_bam::read_alignment_record(stream_type & stream,
         if (!std::ranges::equal(stream_view | views::take_exactly_or_throw(4), std::string_view{"BAM\1"}))
             throw format_error{"File is not in BAM format."};
 
-        int32_t tmp32{};
-        read_field(stream_view, tmp32);
+        int32_t l_text{}; // length of header text including \0 character
+        int32_t n_ref{}; // number of reference sequences
+        int32_t l_name{}; // 1 + length of reference name including \0 character
+        int32_t l_ref{}; // length of reference sequence
```
eseiler, line 339: Per specification, these should be `uint32_t`. Maybe there was a reason they are not, and maybe we can find that reason.
Just replacing does not work.

eseiler (commented):
Follow-up stuff?

include/seqan3/io/sam_file/format_bam.hpp
marehr: Okay, regarding the duplicate entries, I would have assumed that they shouldn't be duplicate in this case :)

include/seqan3/io/sam_file/format_bam.hpp
marehr: Not sure what you mean :)

Do want to copy that `\0` too?

include/seqan3/io/sam_file/format_bam.hpp
marehr: Yeah, definitely a follow-up!

marehr (approved):
Thank you! This looks great!

include/seqan3/io/sam_file/format_bam.hpp
eseiler: The comment says: We do not copy `\0`. 
The code says: We copy `\0`.

So follow-up would be to double check and if we do not copy the `\0` to add a comment as to why is that. Because `l_name` is one more than the length including `\0`, so `l_name - ` should include the `\0`.

eseiler:
I'll do the follow-up issue once this is merged so that I can do permalinks to the code.

PR #2540, opened 2021-04-20
deprecate seqan3::views::drop

marehr:
fixes https://github.com/seqan/product_backlog/issues/325

### DIFF
--- a/include/seqan3/core/platform.hpp
+++ b/include/seqan3/core/platform.hpp
@@ -333,6 +333,7 @@
 #endif
 
 //!\brief See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100139
+//!       std::views::{take, drop} do not type-erase. This is a defect within the standard lib.
 #ifndef SEQAN3_WORKAROUND_GCC_100139 // not yet fixed
 #   if defined(__GNUC__)
 #       define SEQAN3_WORKAROUND_GCC_100139 1
--- a/include/seqan3/core/platform.hpp
+++ b/include/seqan3/core/platform.hpp
@@ -332,6 +332,15 @@
 #   endif
 #endif
 
+//!\brief See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100139
+#ifndef SEQAN3_WORKAROUND_GCC_100139 // not yet fixed
+#   if defined(__GNUC__)
+#       define SEQAN3_WORKAROUND_GCC_100139 1
+#   else
+#       define SEQAN3_WORKAROUND_GCC_100139 0
+#   endif
+#endif
+
 /*!\brief This is needed to support CentOS 7 or RHEL 7; Newer CentOS's include a more modern default-gcc version making
  *        this macro obsolete.
  *
--- a/include/seqan3/range/views/drop.hpp
+++ b/include/seqan3/range/views/drop.hpp
@@ -170,7 +170,7 @@ namespace seqan3::views
  *
  * ### Example
  *
- * \include test/snippet/range/views/drop.cpp
+ * \include test/snippet/utility/views/drop.cpp
  *
  * \hideinitializer
  *
--- a/include/seqan3/range/views/drop.hpp
+++ b/include/seqan3/range/views/drop.hpp
@@ -54,6 +54,7 @@ struct drop_fn
         // safeguard against wrong size
         if constexpr (std::ranges::sized_range<urng_t>)
         {
+            // in standard
             size_t urange_size = std::ranges::size(urange);
             drop_size = std::min(drop_size, urange_size);
             new_size = urange_size - drop_size;
@@ -62,12 +63,14 @@ struct drop_fn
         // string_view
         if constexpr (is_type_specialisation_of_v<std::remove_cvref_t<urng_t>, std::basic_string_view>)
         {
+            // in standard
             return urange.substr(drop_size);
         }
         // string const &
         else if constexpr (is_type_specialisation_of_v<std::remove_cvref_t<urng_t>, std::basic_string> &&
                            std::is_const_v<std::remove_reference_t<urng_t>>)
         {
+            // not in standard
             return std::basic_string_view{std::ranges::data(urange) + drop_size, new_size};
         }
         // contiguous
@@ -75,6 +78,7 @@ struct drop_fn
                            std::ranges::contiguous_range<urng_t> &&
                            std::ranges::sized_range<urng_t>)
         {
+            // in standard
             return std::span{std::ranges::data(urange) + drop_size, new_size};
         }
         // random_access
@@ -82,6 +86,7 @@ struct drop_fn
                            std::ranges::random_access_range<urng_t> &&
                            std::ranges::sized_range<urng_t>)
         {
+            // not in standard
             return std::ranges::subrange<std::ranges::iterator_t<urng_t>, std::ranges::iterator_t<urng_t>>
             {
                 std::ranges::begin(urange) + drop_size,
@@ -168,8 +173,12 @@ namespace seqan3::views
  * \include test/snippet/range/views/drop.cpp
  *
  * \hideinitializer
+ *
+ * \deprecated Use std::views::drop or seqan3::views::type_reduce | std::views::drop.
  */
-inline constexpr auto drop = detail::drop_fn{};
+#ifdef SEQAN3_DEPRECATED_310
+SEQAN3_DEPRECATED_310 inline constexpr auto drop = detail::drop_fn{};
+#endif // SEQAN3_DEPRECATED_310
 
 //!\}
 
--- a/include/seqan3/range/views/slice.hpp
+++ b/include/seqan3/range/views/slice.hpp
@@ -65,6 +65,7 @@ struct slice_fn
 
         // SEQAN3_WORKAROUND_GCC_100139 == 1 if std::views::{take, drop} does not type reduce (e.g. keep in type
         // std::basic_string_view, std::span, std::ranges::subrange).
+        // See https://github.com/seqan/seqan3/pull/2540/files#r617575294
 #if SEQAN3_WORKAROUND_GCC_100139
         // string_view
         if constexpr (is_type_specialisation_of_v<std::remove_cvref_t<urng_t>, std::basic_string_view>)
--- a/include/seqan3/range/views/slice.hpp
+++ b/include/seqan3/range/views/slice.hpp
@@ -19,6 +19,7 @@
 #include <seqan3/range/views/detail.hpp>
 #include <seqan3/range/views/drop.hpp>
 #include <seqan3/range/views/take.hpp>
+#include <seqan3/range/views/type_reduce.hpp>
 #include <seqan3/std/concepts>
 #include <seqan3/std/iterator>
 #include <seqan3/std/ranges>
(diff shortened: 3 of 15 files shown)


SGSSGene (changes requested):
I think there are missing some adjustments:
`include/seqan3/alignment/pairwise/alignment_algorithm.hpp` line 446
`include/seqan3/search/detail/unidirectional_search_algorithm.hpp` line 226
`include/seqan3/io/sam_file/format_sam.hpp` line 1158
`include/seqan3/io/sam_file/format_sam.hpp` line 1171

marehr:
> I think there are missing some adjustments:
> `include/seqan3/alignment/pairwise/alignment_algorithm.hpp` line 446
> `include/seqan3/search/detail/unidirectional_search_algorithm.hpp` line 226
> `include/seqan3/io/sam_file/format_sam.hpp` line 1158
> `include/seqan3/io/sam_file/format_sam.hpp` line 1171

Very nice catch. The interesting thing is that the library build completely through. I'm currently building seqan3 without the definition of seqan3::views::drop. That should show me potential left-overs :)

marehr:
```c++
 In file included from ../../../seqan3/include/seqan3/search/fm_index/bi_fm_index.hpp:22,
                 from ../../../seqan3/test/unit/search/fm_index_cursor/bi_fm_index_cursor_collection_test.cpp:15:
../../../seqan3/include/seqan3/search/fm_index/bi_fm_index_cursor.hpp: In member function ‘void gtest_suite_bi_fm_index_cursor_collection_test_::serialisation<gtest_TypeParam_>::TestBody() [with gtest_TypeParam_ = seqan3::bi_fm_index_cursor<seqan3::bi_fm_index<seqan3::dna4, seqan3::collection> >]’:
../../../seqan3/include/seqan3/search/fm_index/bi_fm_index_cursor.hpp:627:24: error: ‘it.seqan3::bi_fm_index_cursor<seqan3::bi_fm_index<seqan3::dna4, seqan3::collection> >::_last_char’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
  627 |         sdsl_char_type c = _last_char;
      |                        ^
../../../seqan3/include/seqan3/search/fm_index/bi_fm_index_cursor.hpp: In member function ‘void gtest_suite_bi_fm_index_cursor_collection_test_::serialisation<gtest_TypeParam_>::TestBody() [with gtest_TypeParam_ = seqan3::bi_fm_index_cursor<seqan3::bi_fm_index<seqan3::dna5, seqan3::collection> >]’:
../../../seqan3/include/seqan3/search/fm_index/bi_fm_index_cursor.hpp:627:24: error: ‘it.seqan3::bi_fm_index_cursor<seqan3::bi_fm_index<seqan3::dna5, seqan3::collection> >::_last_char’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
  627 |         sdsl_char_type c = _last_char;
      |                        ^
```

SGSSGene (approved):
:+1:

include/seqan3/core/platform.hpp
```diff
@@ -332,6 +332,15 @@
 #   endif
 #endif
 
+//!\brief See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100139
+#ifndef SEQAN3_WORKAROUND_GCC_100139 // not yet fixed
```
eseiler, line 336:
```suggestion
//!\brief See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100139
//!       std::views::{take, drop} do not type-erase. This is a defect within the standard lib.
#ifndef SEQAN3_WORKAROUND_GCC_100139 // not yet fixed
```

We don't do this for every workaround, but I would be in favor of adding a description to the workaround briefs.

test/performance/range/views/view_drop_benchmark.cpp
```diff
@@ -62,31 +62,66 @@ void sequential_read(benchmark::State & state)
 
 BENCHMARK_TEMPLATE(sequential_read, std::string, void);
 BENCHMARK_TEMPLATE(sequential_read, std::string, decltype(std::views::drop));
+#ifdef SEQAN3_DEPRECATED_310
```
eseiler, line 65: I wonder if it had been easier to have another `seqan3::detail::drop` instance that is not deprecated and to use it in the tests.

include/seqan3/range/views/slice.hpp
```diff
@@ -49,21 +50,65 @@ struct slice_fn
...
+            end_pos = std::min(end_pos, urange_size);
         }
+        position_t target_size = end_pos - begin_pos;
 
         if (end_pos < begin_pos)
             throw std::invalid_argument{"end_pos argument to seqan3::views::slice must be >= the begin_pos argument."};
 
-        // urange | drop | take
-        return views::take(views::drop(std::forward<urng_t>(urange), begin_pos), end_pos - begin_pos);
+        // SEQAN3_WORKAROUND_GCC_100139 == 1 if std::views::{take, drop} does not type reduce (e.g. keep in type
```
eseiler, line 66: Can you explain this to me again?
I would have thought that I need to use `views::type_reduce` if the `std::views` did not type erase; but apparently it's the other way around?

eseiler (changes requested):
Mostly questions

test/performance/range/views/view_drop_benchmark.cpp
marehr: we want to remove our implementation, so this should be the "right way" to do this. remove the benchmark parts. we could argue to remove the whole benchmark.

we could maybe add a local type-reduced drop adaptor just for this benchmark, like this

```cpp
inline constexpr type_reduced_drop = seqan3::views::type_reduce | std::views::drop;
```

eseiler: Mhm, I meant something like:

```cpp
namespace seqan3::detail
{

inline constexpr auto drop = detail::drop_fn{};

}
```

Then you have `seqan3::detail::drop` which is not deprecated, and `seqan3::views::drop` which is deprecated.
Now you could use `seqan3::detail::drop` in tests and benchmarks. When we finally remove it, we can remove all `seqan3::detail::drop` occurrences.
Seems easier than doing `pragma` everywhere.

include/seqan3/range/views/slice.hpp
marehr: Yeah, this is a bit complicated xD

so `std::views::drop(range_t, 2) -> view_t` has the following rules:

1. `range_t = std::span`, returns `view_t = std::span`
2. `range_t = std::string_view`, returns `view_t = std::string_view`
3. `range_t = std::ranges::subrange`, returns `view_t = std::ranges::subrange`

otherwise
* `view_t = std::ranges::drop_view<range_t>``

Currently, gcc 10 and 11 does always return `view_t = std::ranges::drop_view<range_t>`.

seqan3::views::type_reduce has some additional special rules:
1. `range_t is view`, returns view as is
  * this line is problematic if we do `view | type_reduce`,
  * as a `std::ranges::drop_view<range_t>` will be passed through;
  * that means only `type_reduce | view` makes sense
  * one could argue that this rule should be checked in a different order (at least that the borrowed_range ones have priority)
2. `range_t = std::string const &`, returns `view_t = std::string_view`
3. `range_t = std::ranges::contiguous_range && std::ranges::borrowed_range && std::ranges::sized_range`, returns `view_t = std::span`
4. `range_t = std::ranges::random_access_range && std::ranges::borrowed_range && std::ranges::sized_range`, returns `view_t = std::ranges::subrange`

----

Anyway since only `type_reduce | view` makes sense right now, we need that `std::views::drop` stays in type itself, otherwise it will always wrap `std::ranges::drop_view` around any type.

Example

view_t = range_t | seqan3::views::slice(0, 0);

| range_t | view_t if `SEQAN3_WORKAROUND_GCC_100139 == 1` | view_t if `SEQAN3_WORKAROUND_GCC_100139 == 0` | expected view_t |
| --- | --- | --- | --- |
| std::span | std::span | std::span \| type_reduce (1. Rule) \| drop (1. Rule) \| take (1. Rule) = std::span | std::span |
| std::string_view | std::string_view | std::string_view \| type_reduce (1. Rule) \| drop (2. Rule) \| take (2. Rule) = std::string_view | std::string_view |
| std::ranges::subrange | std::ranges::subrange | std::ranges::subrange \| type_reduce (4. Rule?) \| drop (3. Rule) \| take (3. Rule) = std::ranges::subrange | std::ranges::subrange |
| std::string const & | std::string_view | std::string const & \| type_reduce (2. Rule) \| drop (2. Rule) \| take (2. Rule) = std::string_view | std::string_view |
| std::ranges::contiguous_range && std::ranges::borrowed_range && std::ranges::sized_range | std::span | range_t \| type_reduce (3. Rule) \| drop (1. Rule) \| take (1. Rule) = std::span | std::span |
| std::ranges::random_access_range && std::ranges::borrowed_range && std::ranges::sized_range | std::ranges::subrange | range_t \| type_reduce (4. Rule) \| drop (3. Rule) \| take (3. Rule) = std::ranges::subrange | std::ranges::subrange |

test/performance/range/views/view_drop_benchmark.cpp
marehr: this does not work; you can only pipe range_adaptor_closures

eseiler: Why not?
`detail::drop` is the same as `views::drop`.
So you can just use `detail::drop` in tests, instead of `pragma`ing them

marehr: This would work:

```patch
diff --git a/test/performance/range/views/view_drop_benchmark.cpp b/test/performance/range/views/view_drop_benchmark.cpp
index a3e20bcc9..9ad8356aa 100644
--- a/test/performance/range/views/view_drop_benchmark.cpp
+++ b/test/performance/range/views/view_drop_benchmark.cpp
@@ -16,6 +16,7 @@
 
 #include <seqan3/range/views/single_pass_input.hpp>
+#include <seqan3/range/views/type_reduce.hpp>
 
 // ============================================================================
 //  sequential_read
@@ -60,68 +61,41 @@ void sequential_read(benchmark::State & state)
     [[maybe_unused]] volatile uint8_t dummy2 = dummy;
 }
 
+struct type_reduced_drop_fn
+{
+    constexpr auto operator()(int drop_size) const
+    {
+        return seqan3::views::type_reduce | std::views::drop(drop_size);
+    }
+};
+
 BENCHMARK_TEMPLATE(sequential_read, std::string, void);
 BENCHMARK_TEMPLATE(sequential_read, std::string, decltype(std::views::drop));
-#ifdef SEQAN3_DEPRECATED_310
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
-BENCHMARK_TEMPLATE(sequential_read, std::string, decltype(seqan3::views::drop));
-#pragma GCC diagnostic pop
-#endif // SEQAN3_DEPRECATED_310
+BENCHMARK_TEMPLATE(sequential_read, std::string, type_reduced_drop_fn);
 
 BENCHMARK_TEMPLATE(sequential_read, std::vector<uint8_t>, void);
 BENCHMARK_TEMPLATE(sequential_read, std::vector<uint8_t>, decltype(std::views::drop));
```

eseiler: Yes, but I meant:

```cpp
diff --git a/include/seqan3/range/views/drop.hpp b/include/seqan3/range/views/drop.hpp
index d1680ca9..5fd8c42e 100644
--- a/include/seqan3/range/views/drop.hpp
+++ b/include/seqan3/range/views/drop.hpp
@@ -99,6 +99,9 @@ struct drop_fn
     }
 };
 
+//!brief Instance for usage until removed.
+inline constexpr auto drop = drop_fn{};
+
 } // namespace seqan3::detail
 
 // ============================================================================
diff --git a/test/performance/range/views/view_drop_benchmark.cpp b/test/performance/range/views/view_drop_benchmark.cpp
index fe645ef4..6545fec8 100644
--- a/test/performance/range/views/view_drop_benchmark.cpp
+++ b/test/performance/range/views/view_drop_benchmark.cpp
@@ -62,7 +62,7 @@ void sequential_read(benchmark::State & state)
 
 BENCHMARK_TEMPLATE(sequential_read, std::string, void);
 BENCHMARK_TEMPLATE(sequential_read, std::string, decltype(std::views::drop));
-BENCHMARK_TEMPLATE(sequential_read, std::string, decltype(seqan3::views::drop));
+BENCHMARK_TEMPLATE(sequential_read, std::string, decltype(seqan3::detail::drop));
 
 BENCHMARK_TEMPLATE(sequential_read, std::vector<uint8_t>, void);
 BENCHMARK_TEMPLATE(sequential_read, std::vector<uint8_t>, decltype(std::views::drop));

```
(and replacing `seqan3::views::drop` with `seqan3::detail::drop` in the unit test)

And once we remove `seqan3::views::drop`, we can remove the test as a whole and the benchmarks with `seqan3::detail::drop`.

marehr: Ah okay :)

I think I want to keep those benchmarks. Maybe adding type_reduce later would also be cool.

include/seqan3/range/views/slice.hpp
```diff
@@ -49,21 +50,65 @@ struct slice_fn
...
+        // std::basic_string_view, std::span, std::ranges::subrange).
+#if SEQAN3_WORKAROUND_GCC_100139
```
eseiler, line 68:
```suggestion
        // std::basic_string_view, std::span, std::ranges::subrange).
        // See https://github.com/seqan/seqan3/pull/2540/files#r617575294
#if SEQAN3_WORKAROUND_GCC_100139
```

include/seqan3/range/views/drop.hpp
```diff
@@ -112,64 +117,11 @@ namespace seqan3::views
...
- *
- * \include test/snippet/range/views/drop.cpp
- *
- * \hideinitializer
+/*!\deprecated Use std::views::drop or seqan3::views::type_reduce | std::views::drop.
  */
-inline constexpr auto drop = detail::drop_fn{};
+#ifdef SEQAN3_DEPRECATED_310
+SEQAN3_DEPRECATED_310 inline constexpr auto drop = detail::drop_fn{};
+#endif // SEQAN3_DEPRECATED_310
```
eseiler, line 124: Why the `ifdef` ?

include/seqan3/core/platform.hpp
eseiler: open

include/seqan3/range/views/drop.hpp
```diff
@@ -112,64 +117,11 @@ namespace seqan3::views
...
- *
- * Construction time of the returned view is in \f$ O(1) \f$ if the underlying range models at least
- * std::ranges::random_access_range and std::ranges::sized_range; otherwise in \f$ O(drop\_size) \f$.
- *
- * ### Example
- *
- * \include test/snippet/range/views/drop.cpp
- *
- * \hideinitializer
+/*!\deprecated Use std::views::drop or seqan3::views::type_reduce | std::views::drop.
```
eseiler, line 120: Shouldn't we keep the docs until it is actually removed?

eseiler (changes requested):
Small stuff

include/seqan3/range/views/drop.hpp
marehr: Yeah, we could. What do you prefer?

include/seqan3/range/views/drop.hpp
marehr: I want to do a PR that introduces a macro that disables the definition of `SEQAN3_DEPRECATED_310` to emulate the situation after we remove all the deprecated entities.

We can talk about my vision tomorrow :)

include/seqan3/range/views/drop.hpp
eseiler: Would prefer keeping it, but it's not that incredibly important since we will remove it soon anyways

include/seqan3/range/views/drop.hpp
eseiler: Makes sense, but then we should start doing it where applicable (or just do it for the other views afterwards)

eseiler (approved):

PR #2542, opened 2021-04-21
move seqan3::views::slice to seqan3/utility/views/

marehr:
fixes https://github.com/seqan/product_backlog/issues/361


MitraDarja (approved):

test/unit/range/views/slice_test.cpp
```diff
@@ -15,9 +15,9 @@
 #include <vector>
```
eseiler, line 15: Shouldn't this be moved to utility?

eseiler (changes requested):

test/unit/range/views/slice_test.cpp
marehr: 👍🏽 good catch xD

eseiler (approved):

PR #2552, opened 2021-04-22
[INFRA] Split actions

eseiler:
Splits CI in 3 parts: Linux, macOS, misc
- each part can be restarted individually
- Linux + macOS CI have fail_fast enabled, i.e., if one job fails, the workflow gets cancelled
- cancel workflow now uses `all_but_latest`, it will also cancel future runs such that only the latest run stays in the queue (otherwise it will only cancel runs up until the point in time the cancel workflow was enqueued)
- cmake install step is now a script and also sets the correct path for cmake on macOS (had no effect on macOS before as the path was wrong)
- doxygen install is also a script
- we dont cache doxygen/cmake downloads anymore, as it did not work on macos (cmake) and we rarely get a cache hit with these 
- This PR only: CI will show as queued as its expecting runs from the old CI


marehr (approved):

eseiler:
Move snippet/header/performance to SeqAn3 CI

.github/workflows/ci_linux.yml
```diff
@@ -0,0 +1,154 @@
...
+          cmake ../seqan3/test/${{ matrix.build }} -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} -DCMAKE_CXX_FLAGS="${{ matrix.cxx_flags }}" -DSEQAN3_VERBOSE_TESTS=OFF -DSEQAN3_BENCHMARK_MIN_TIME=0.01
+          make gtest_build
+
+      - name: Build tests
+        env:
+          CCACHE_BASEDIR: ${{ github.workspace }}
+          CCACHE_DIR: ${{ github.workspace }}/.ccache
+          CCACHE_COMPRESS: true
+          CCACHE_COMPRESSLEVEL: 6
+          CCACHE_MAXSIZE: 1G
```
SGSSGene, line 136: Is 1G enough? (it was 5G before)

SGSSGene (changes requested):

.github/workflows/ci_linux.yml
eseiler: Yep, coverage (biggest) is 580MiB
Also, ccache won't try to clean up (remove obsolete entries) if it does not reach the limit

SGSSGene (approved):
:+1:

PR #2554, opened 2021-04-23
[MISC] Rename views::get to views::elements

eseiler:
Resolves seqan/product_backlog#326

`std::views::elements` is not implemented in range-v3

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100233


include/seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm_banded.hpp
```diff
@@ -17,8 +17,8 @@
 
 #include <seqan3/alignment/pairwise/detail/pairwise_alignment_algorithm.hpp>
 #include <seqan3/range/views/drop.hpp>
-#include <seqan3/range/views/take.hpp>
 #include <seqan3/range/views/slice.hpp>
+#include <seqan3/range/views/take.hpp>
```
Irallia, line 21: Don't you need to add `#include <seqan3/range/views/elements.hpp>`?

Irallia (approved):
Maybe one include missing, but I'll already aprove.
The rest LGFM, thank you!

marehr (approved):

PR #2555, opened 2021-04-23
[FIX] cmake-3.20: make add_include_dependencies work again

marehr:
This also adds a cmake warning if the header-dependency file is empty to make diagnostics easier if the whole mechanism doesn't work any more.

See introducing PR https://github.com/seqan/seqan3/pull/2261 for more details.


SGSSGene (approved):
*cmake wisdom*

eseiler (changes requested):
Just 💅 

The important change is `CMAKE_DEPENDS_USE_COMPILER`?
And are the other changes important for that or just better style/error handling?

marehr:
> Just nail_care
> 
> The important change is `CMAKE_DEPENDS_USE_COMPILER`?

Yes

> And are the other changes important for that or just better style/error handling?

I also added some changes that makes the processing more robust, as the output format varied to that before (and did not work)

eseiler (approved):

PR #2563, opened 2021-04-26
[MISC] deprecate seqan3::views::move

marehr:
resolve https://github.com/seqan/product_backlog/issues/328


SGSSGene (approved):
:+1:

eseiler (changes requested):
Looks good, but what about

https://github.com/seqan/seqan3/blob/8b4cfe0b66215cdba1afba3b5a9ecfd77bb453d1/doc/tutorial/sequence_file/sequence_file_solution3.cpp#L90

marehr:
> Looks good, but what about
> 
> https://github.com/seqan/seqan3/blob/8b4cfe0b66215cdba1afba3b5a9ecfd77bb453d1/doc/tutorial/sequence_file/sequence_file_solution3.cpp#L90
> 
> Snippets build because it is a snippet with a workaround that is always active on GCC

Ty good catch

eseiler (approved):
I removed the `SEQAN3_WORKAROUND_GCC_96070` workaround from `sequence_file_solution3.cpp`

PR #2570, opened 2021-04-28
Deprecate istreambuf

marehr:
resolves https://github.com/seqan/product_backlog/issues/327


include/seqan3/range/views/istreambuf.hpp
```diff
@@ -6,112 +6,13 @@
 // -----------------------------------------------------------------------------------------------------
 
 /*!\file
+ * \brief [DEPRECATED] Provides seqan3::detail::istreambuf.
  * \author Hannes Hauswedell <hannes.hauswedell AT fu-berlin.de>
```
MitraDarja, line 10: Not sure if necessary, but also add \deprecated

MitraDarja (approved):
LGTM, just one minor thing. 💅

eseiler (approved):

PR #2576, opened 2021-04-28
[FIX] gcc-11: satisfaction of atomic constraint depends on itself

marehr:
Fixes https://github.com/seqan/product_backlog/issues/371

### DIFF
--- a/include/seqan3/range/container/bitcompressed_vector.hpp
+++ b/include/seqan3/range/container/bitcompressed_vector.hpp
@@ -1088,7 +1088,7 @@ public:
      * \details
      * \stableapi{Since version 3.1.}
      */
-    friend constexpr bool operator==(bitcompressed_vector && lhs, bitcompressed_vector && rhs) noexcept
+    friend constexpr bool operator!=(bitcompressed_vector && lhs, bitcompressed_vector && rhs) noexcept
     {
         return lhs.data != rhs.data;
     }
--- a/include/seqan3/range/container/bitcompressed_vector.hpp
+++ b/include/seqan3/range/container/bitcompressed_vector.hpp
@@ -1079,18 +1079,18 @@ public:
      * \details
      * \stableapi{Since version 3.1.}
      */
-    constexpr bool operator==(bitcompressed_vector const & rhs) const noexcept
+    friend constexpr bool operator==(bitcompressed_vector && lhs, bitcompressed_vector && rhs) noexcept
     {
-        return data == rhs.data;
+        return lhs.data == rhs.data;
     }
 
     /*!\brief Checks whether `*this` is not equal to `rhs`.
      * \details
      * \stableapi{Since version 3.1.}
      */
-    constexpr bool operator!=(bitcompressed_vector const & rhs) const noexcept
+    friend constexpr bool operator==(bitcompressed_vector && lhs, bitcompressed_vector && rhs) noexcept
     {
-        return data != rhs.data;
+        return lhs.data != rhs.data;
     }
 
     /*!\brief Checks whether `*this` is less than `rhs`.
--- a/include/seqan3/range/container/bitcompressed_vector.hpp
+++ b/include/seqan3/range/container/bitcompressed_vector.hpp
@@ -1079,7 +1079,7 @@ public:
      * \details
      * \stableapi{Since version 3.1.}
      */
-        constexpr bool operator==(bitcompressed_vector const & rhs) const noexcept
+    constexpr bool operator==(bitcompressed_vector const & rhs) const noexcept
     {
         return data == rhs.data;
     }
--- a/include/seqan3/range/container/bitcompressed_vector.hpp
+++ b/include/seqan3/range/container/bitcompressed_vector.hpp
@@ -1079,18 +1079,18 @@ public:
      * \details
      * \stableapi{Since version 3.1.}
      */
-    friend constexpr bool operator==(bitcompressed_vector && lhs, bitcompressed_vector && rhs) noexcept
+        constexpr bool operator==(bitcompressed_vector const & rhs) const noexcept
     {
-        return lhs.data == rhs.data;
+        return data == rhs.data;
     }
 
     /*!\brief Checks whether `*this` is not equal to `rhs`.
      * \details
      * \stableapi{Since version 3.1.}
      */
-    friend constexpr bool operator!=(bitcompressed_vector && lhs, bitcompressed_vector && rhs) noexcept
+    constexpr bool operator!=(bitcompressed_vector const & rhs) const noexcept
     {
-        return lhs.data != rhs.data;
+        return data != rhs.data;
     }
 
     /*!\brief Checks whether `*this` is less than `rhs`.
--- a/include/seqan3/range/container/bitcompressed_vector.hpp
+++ b/include/seqan3/range/container/bitcompressed_vector.hpp
@@ -200,9 +200,11 @@ public:
      *
      * \experimentalapi{Experimental since version 3.1. This is a non-standard C++ extension.}
      */
-    template <std::ranges::input_range other_range_t>
+    template <typename other_range_t>
     //!\cond
-        requires has_same_value_type_v<other_range_t>
+        requires (!std::same_as<bitcompressed_vector, std::remove_cvref_t<other_range_t>>) &&
+                 std::ranges::input_range<other_range_t> &&
+                 has_same_value_type_v<other_range_t>
     //!\endcond
     explicit bitcompressed_vector(other_range_t && range) :
         bitcompressed_vector{std::ranges::begin(range), std::ranges::end(range)}
--- a/test/unit/range/container/bitcompressed_vector_test.cpp
+++ b/test/unit/range/container/bitcompressed_vector_test.cpp
@@ -42,8 +42,9 @@ TEST(bitcompressed_vector_test, issue1743_view_combinability)
     EXPECT_RANGE_EQ(complement, (seqan3::dna4_vector{'T'_dna4, 'G'_dna4, 'C'_dna4, 'A'_dna4}));
 }
 
+// https://github.com/seqan/product_backlog/issues/371
 TEST(bitcompressed_vector_test, issue371)
-{ // https://github.com/seqan/product_backlog/issues/371
+{ 
     using alphabet_t = seqan3::alphabet_variant<seqan3::dna4, seqan3::dna15>;
     seqan3::bitcompressed_vector<alphabet_t> source{};
     auto it = source.begin();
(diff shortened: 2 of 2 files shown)


eseiler:
Interesting, I could "reduce" it to

```cpp
#include <seqan3/alphabet/adaptation/char.hpp>
#include <seqan3/alphabet/gap/gapped.hpp>
#include <seqan3/range/container/bitcompressed_vector.hpp>

int main()
{
   using alphabet_t = seqan3::alphabet_variant<char, seqan3::dna4>;
   seqan3::bitcompressed_vector<alphabet_t> source{};
   auto it = source.begin();
   auto end = source.end();
   it != end; // This line causes error.
}
```

And could fix is making the comparison operators friends
```patch
diff --git a/include/seqan3/range/container/bitcompressed_vector.hpp b/include/seqan3/range/container/bitcompressed_vector.hpp
index baddc729..44b9ea62 100644
--- a/include/seqan3/range/container/bitcompressed_vector.hpp
+++ b/include/seqan3/range/container/bitcompressed_vector.hpp
@@ -1077,18 +1077,18 @@ public:
      * \details
      * \stableapi{Since version 3.1.}
      */
-    constexpr bool operator==(bitcompressed_vector const & rhs) const noexcept
+    friend constexpr bool operator==(bitcompressed_vector && lhs, bitcompressed_vector && rhs) noexcept
     {
-        return data == rhs.data;
+        return lhs.data == rhs.data;
     }
 
     /*!\brief Checks whether `*this` is not equal to `rhs`.
      * \details
      * \stableapi{Since version 3.1.}
      */
-    constexpr bool operator!=(bitcompressed_vector const & rhs) const noexcept
+    friend constexpr bool operator!=(bitcompressed_vector && lhs, bitcompressed_vector && rhs) noexcept
     {
-        return data != rhs.data;
+        return lhs.data != rhs.data;
     }
 
     /*!\brief Checks whether `*this` is less than `rhs`.

```

marehr:
> I could "reduce" it to
> 
> ```c++
> #include <seqan3/alphabet/adaptation/char.hpp>
> #include <seqan3/alphabet/gap/gapped.hpp>
> #include <seqan3/alphabet/nucleotide/dna4.hpp>
> #include <seqan3/range/container/bitcompressed_vector.hpp>
> 
> int main()
> {
>    using alphabet_t = seqan3::alphabet_variant<char, seqan3::dna4>;
>    seqan3::bitcompressed_vector<alphabet_t> source{};
>    auto it = source.begin();
>    auto end = source.end();
>    it != end; // This line causes error.
> }
> ```
> 
> And could fix it by making the comparison operators friends
> 
> ```diff
> diff --git a/include/seqan3/range/container/bitcompressed_vector.hpp b/include/seqan3/range/container/bitcompressed_vector.hpp
> index baddc729..44b9ea62 100644
> --- a/include/seqan3/range/container/bitcompressed_vector.hpp
> +++ b/include/seqan3/range/container/bitcompressed_vector.hpp
> @@ -1077,18 +1077,18 @@ public:
>       * \details
>       * \stableapi{Since version 3.1.}
>       */
> -    constexpr bool operator==(bitcompressed_vector const & rhs) const noexcept
> +    friend constexpr bool operator==(bitcompressed_vector && lhs, bitcompressed_vector && rhs) noexcept
>      {
> -        return data == rhs.data;
> +        return lhs.data == rhs.data;
>      }
>  
>      /*!\brief Checks whether `*this` is not equal to `rhs`.
>       * \details
>       * \stableapi{Since version 3.1.}
>       */
> -    constexpr bool operator!=(bitcompressed_vector const & rhs) const noexcept
> +    friend constexpr bool operator!=(bitcompressed_vector && lhs, bitcompressed_vector && rhs) noexcept
>      {
> -        return data != rhs.data;
> +        return lhs.data != rhs.data;
>      }
>  
>      /*!\brief Checks whether `*this` is less than `rhs`.
> ```

Nice, I set on this forever xD

Rock, paper, scissor?

(My fix is more general, as it excludes the potential of recursion which I also had trouble with clang, your test case is more simple and shows the actual problem xD)

eseiler:
> Nice, I sat on this forever xD

I had the right hunch so it just took a few minutes (for once)

> Rock, paper, scissor?
>
> (My fix is more general, as it excludes the potential of recursion which I also had trouble with clang, your test case is more simple and shows the actual problem xD)

I would definitely change the testing example, because I think `it != end` is clearer.

The range-based for loops does the same (https://cppinsights.io/s/d1d6f5e9)
But the `c = a` indicates that this may be a problem with assignment which it is not.

Restricting `other_range_t` to be not recursive seems also correct.
IIRC we want to have comparison operators as friends, no?

Then maybe we just do both changes?

FYI gcc11 introduces new template matching: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0522r0.html
Haven't read it yet but maybe related

marehr:
> FYI gcc11 introduces new template matching: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0522r0.html
> Haven't read it yet but maybe related

Nice, I didn't know that there was a paper for this. This bothered my a couple of times already :)

marehr:
> IIRC we want to have comparison operators as friends, no?

I personally like them to be members, especially since C++20 where also the commutative ones are auto-generated.

But, let's just apply that and see if anything breaks :)

eseiler:
> > IIRC we want to have comparison operators as friends, no?
> 
> I personally like them to be members, especially since C++20 where also the commutative ones are auto-generated.
> 
> But, let's just apply that and see if anything breaks :)

I am also fine with just doing the `non-recursive` thing.

test/unit/range/container/bitcompressed_vector_test.cpp
```diff
@@ -39,3 +41,13 @@ TEST(bitcompressed_vector_test, issue1743_view_combinability)
...
 }
+
+TEST(bitcompressed_vector_test, issue371)
+{ // https://github.com/seqan/product_backlog/issues/371
+    using alphabet_t = seqan3::alphabet_variant<seqan3::dna4, seqan3::dna15>;
+    seqan3::bitcompressed_vector<alphabet_t> source{};
+    alphabet_t a{};
+
+    for (auto && c : source)
+        c = a;
```
marehr, line 52:
```suggestion
    using alphabet_t = seqan3::alphabet_variant<seqan3::dna4, seqan3::dna15>;
    seqan3::bitcompressed_vector<alphabet_t> source{};
    auto it = source.begin();
    auto end = source.end();
    it != end; // This line causes error.
```

include/seqan3/range/container/bitcompressed_vector.hpp
```diff
@@ -1077,18 +1079,18 @@ class bitcompressed_vector
...
-        return data == rhs.data;
+        return lhs.data == rhs.data;
     }
 
     /*!\brief Checks whether `*this` is not equal to `rhs`.
      * \details
      * \stableapi{Since version 3.1.}
      */
-    constexpr bool operator!=(bitcompressed_vector const & rhs) const noexcept
+    friend constexpr bool operator==(bitcompressed_vector && lhs, bitcompressed_vector && rhs) noexcept
```
eseiler, line 1091:
```suggestion
    friend constexpr bool operator!=(bitcompressed_vector && lhs, bitcompressed_vector && rhs) noexcept
```

include/seqan3/range/container/bitcompressed_vector.hpp
```diff
@@ -1077,18 +1079,18 @@ class bitcompressed_vector
      * \details
      * \stableapi{Since version 3.1.}
      */
-    constexpr bool operator==(bitcompressed_vector const & rhs) const noexcept
+    friend constexpr bool operator==(bitcompressed_vector && lhs, bitcompressed_vector && rhs) noexcept
```
marehr, line 1082:
```suggestion
        constexpr bool operator==(bitcompressed_vector const & rhs) const noexcept
```

include/seqan3/range/container/bitcompressed_vector.hpp
```diff
@@ -1077,18 +1079,18 @@ class bitcompressed_vector
...
     {
-        return data == rhs.data;
+        return lhs.data == rhs.data;
```
marehr, line 1084:
```suggestion
        return data == rhs.data;
```

include/seqan3/range/container/bitcompressed_vector.hpp
```diff
@@ -1077,18 +1079,18 @@ class bitcompressed_vector
...
     }
 
     /*!\brief Checks whether `*this` is not equal to `rhs`.
      * \details
      * \stableapi{Since version 3.1.}
      */
-    constexpr bool operator!=(bitcompressed_vector const & rhs) const noexcept
+    friend constexpr bool operator!=(bitcompressed_vector && lhs, bitcompressed_vector && rhs) noexcept
```
marehr, line 1091:
```suggestion
    constexpr bool operator!=(bitcompressed_vector const & rhs) const noexcept
```

include/seqan3/range/container/bitcompressed_vector.hpp
```diff
@@ -1077,18 +1079,18 @@ class bitcompressed_vector
...
     {
-        return data != rhs.data;
+        return lhs.data != rhs.data;
```
marehr, line 1093:
```suggestion
        return data != rhs.data;
```

include/seqan3/range/container/bitcompressed_vector.hpp
```diff
@@ -1077,7 +1079,7 @@ class bitcompressed_vector
      * \details
      * \stableapi{Since version 3.1.}
      */
-    constexpr bool operator==(bitcompressed_vector const & rhs) const noexcept
+        constexpr bool operator==(bitcompressed_vector const & rhs) const noexcept
```
marehr, line 1082:
```suggestion
    constexpr bool operator==(bitcompressed_vector const & rhs) const noexcept
```

test/unit/range/container/bitcompressed_vector_test.cpp
```diff
@@ -39,3 +41,12 @@ TEST(bitcompressed_vector_test, issue1743_view_combinability)
...
 }
+
+TEST(bitcompressed_vector_test, issue371)
+{ // https://github.com/seqan/product_backlog/issues/371
+    using alphabet_t = seqan3::alphabet_variant<seqan3::dna4, seqan3::dna15>;
+    seqan3::bitcompressed_vector<alphabet_t> source{};
+    auto it = source.begin();
+    auto end = source.end();
+    it != end; // This line causes error.
+}
```
eseiler, line 52:
```suggestion

// https://github.com/seqan/product_backlog/issues/371
TEST(bitcompressed_vector_test, issue371)
{ 
    using alphabet_t = seqan3::alphabet_variant<seqan3::dna4, seqan3::dna15>;
    seqan3::bitcompressed_vector<alphabet_t> source{};
    auto it = source.begin();
    auto end = source.end();
    it != end; // This line causes error.
}
```

MitraDarja (approved):

eseiler (approved):

PR #2577, opened 2021-04-29
[INFRA] Add GCC 9 and 11 to API

eseiler:
Resolves https://github.com/seqan/product_backlog/issues/368

No ccache because this action won't run very often and hence almost never have cache hits (also, the run needs to succeed for the cache to be stored).


SGSSGene (approved):
This variable notation `gcc${{ matrix.compiler }}` is interesting. This is not yaml specific, is it?

eseiler:
> This variable notation `gcc${{ matrix.compiler }}` is interesting. This is not yaml specific, is it?

The mustache syntax is YAML, but the `$` is custom

marehr (approved):

PR #2582, opened 2021-04-29
[FIX] gcc-10: missing RANGES_DEEP_STL_INTEGRATION

marehr:
resolves https://github.com/seqan/product_backlog/issues/372
resolves https://github.com/seqan/seqan3/issues/2580

part of https://github.com/seqan/product_backlog/issues/124

### DIFF
--- a/doc/tutorial/argument_parser/solution3.cpp
+++ b/doc/tutorial/argument_parser/solution3.cpp
@@ -1,13 +1,11 @@
 //![program]
+#include <seqan3/std/charconv>            // includes std::from_chars
+#include <seqan3/std/filesystem>          // use std::filesystem::path
 #include <fstream>
 #include <numeric>
 
-#include <range/v3/view/split.hpp>
-
 #include <seqan3/argument_parser/all.hpp> // includes all necessary headers
 #include <seqan3/core/debug_stream.hpp>   // our custom output stream
-#include <seqan3/std/charconv>            // includes std::from_chars
-#include <seqan3/std/filesystem>          // use std::filesystem::path
 
 // This is the program!
 // Take a look at it if you are interested in an example of parsing a data file.
--- a/doc/tutorial/sequence_file/sequence_file_snippets.cpp
+++ b/doc/tutorial/sequence_file/sequence_file_snippets.cpp
@@ -1,20 +1,20 @@
 #include <fstream>
 #include <numeric> // std::accumulate
-//![include_ranges_chunk]
-#include <range/v3/view/chunk.hpp>
-//![include_ranges_chunk]
-
-#include <seqan3/utility/char_operations/predicate.hpp>
-//![include]
-#include <seqan3/io/sequence_file/all.hpp>
-//![include]
-//![include_debug_stream]
-#include <seqan3/core/debug_stream.hpp>
-//![include_debug_stream]
-#include <seqan3/range/detail/misc.hpp>
 //![include_ranges]
 #include <seqan3/std/ranges>
 //![include_ranges]
+
+//![include_debug_stream]
+#include <seqan3/core/debug_stream.hpp>
+//![include_debug_stream]
+//![include]
+#include <seqan3/io/sequence_file/all.hpp>
+//![include]
+#include <seqan3/range/detail/misc.hpp>
+#include <seqan3/utility/char_operations/predicate.hpp>
+//![include_ranges_chunk]
+#include <seqan3/utility/views/chunk.hpp>
+//![include_ranges_chunk]
 
 struct write_file_dummy_struct
 {
--- a/include/seqan3/utility/views/interleave.hpp
+++ b/include/seqan3/utility/views/interleave.hpp
@@ -16,14 +16,13 @@
 #include <seqan3/std/concepts>
 #include <seqan3/std/ranges>
 
-#include <range/v3/view/chunk.hpp>
-
 #include <seqan3/core/detail/persist_view.hpp>
 #include <seqan3/range/detail/random_access_iterator.hpp>
 #include <seqan3/range/views/detail.hpp>
 #include <seqan3/range/views/type_reduce.hpp>
 #include <seqan3/utility/type_traits/detail/transformation_trait_or.hpp>
 #include <seqan3/utility/type_traits/pre.hpp>
+#include <seqan3/utility/views/chunk.hpp>
 #include <seqan3/utility/views/join_with.hpp>
 
 namespace seqan3::detail
@@ -294,7 +293,7 @@ struct interleave_fn
         }
         else
         {
-            return std::forward<urng_t>(urange) | ranges::views::chunk(static_cast<size_t>(size))
+            return std::forward<urng_t>(urange) | seqan3::views::chunk(static_cast<size_t>(size))
                                                 | views::join_with(std::forward<inserted_rng_t>(i));
         }
     }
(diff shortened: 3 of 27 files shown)


eseiler (approved):

marehr:
This does not solve the api_stability issue yet, as the first included ranges header (range/v3/view/unique.hpp) does not have `RANGES_DEEP_STL_INTEGRATION` defined.

https://github.com/seqan/seqan3/blob/370673ffd20d790f759aed15771072436a422ec3/test/unit/range/views/view_slice_test.cpp#L8-L26

I'm thinking of adding a diagnostic warning

```c++
#if (defined(RANGES_V3_ITERATOR_ACCESS_HPP) || defined(RANGES_V3_ITERATOR_TRAITS_HPP)) && !RANGES_DEEP_STL_INTEGRATION
#pragma warning "You included a range-v3 header before including `#include <seqan3/std/ranges>`, please be aware that seqan3 might fail to compile if `#define RANGES_DEEP_STL_INTEGRATION 1` isn't provided."
#endif // 
```

eseiler:
> This does not solve the api_stability issue yet, as the first included ranges header (range/v3/view/unique.hpp) does not have `RANGES_DEEP_STL_INTEGRATION` defined.

Is it something that might happen often? I'd figure we could also add `-DRANGED...` to our `CMAKE_FLAGS`. But if it is once in a while like here, I think a diagnostics to `std/ranges` is sufficient.

marehr:
No that doesn't happen often, but it is unexpected if it happens and hard to diagnose. So I think giving a warning makes it easier to put you on the right track. 

Am 30. April 2021 10:32:14 MESZ schrieb Enrico Seiler ***@***.***>:
>> This does not solve the api_stability issue yet, as the first
>included ranges header (range/v3/view/unique.hpp) does not have
>`RANGES_DEEP_STL_INTEGRATION` defined.
>
>Is it something that might happen often? I'd figure we could also add
>`-DRANGES...` to our `CMAKE_FLAGS`. But if it is once in a while like
>here, I think a diagnostics to `std/ranges` is sufficient.

eseiler:
> No that doesn't happen often, but it is unexpected if it happens and hard to diagnose. So I think giving a warning makes it easier to put you on the right track. 

Diagnostics it is then :)
I already updated the wiki about our include order (which is occasionally being consulted) and added a note that you need to `#include <seqan3/std/ranges>` before including any range-v3 header.

test/unit/std/ranges_test.cpp
```diff
@@ -35,3 +35,21 @@ TEST(ranges_test, gcc10bug_rangev3_1514)
         EXPECT_EQ(*std::ranges::begin(iota), 0u);
     }
 }
+
+// https://github.com/seqan/product_backlog/issues/372
+TEST(ranges_test, issue372)
+{
+#ifdef __cpp_lib_ranges // >= gcc-10, ranges-bug
```
marehr, line 42:
```suggestion
#ifdef __cpp_lib_ranges // >= gcc-10, range-v3-bug
```

test/unit/std/ranges_test.cpp
```diff
@@ -35,3 +35,21 @@ TEST(ranges_test, gcc10bug_rangev3_1514)
...
+    std::istringstream istringstream{vec};
+
+    auto v1 = std::ranges::subrange{std::istream_iterator<char>{istringstream}, std::default_sentinel};
+    auto v2 = std::views::take(v1, 1);
+
+    using iterator2_t = std::ranges::iterator_t<decltype(v2)>;
+    EXPECT_TRUE(std::indirectly_readable<iterator2_t>);
+    EXPECT_TRUE(ranges::indirectly_readable<iterator2_t>);
+    EXPECT_TRUE(std::input_iterator<iterator2_t>);
+    EXPECT_TRUE(ranges::input_iterator<iterator2_t>);
```
marehr, line 53:
```suggestion
    EXPECT_TRUE(std::indirectly_readable<iterator2_t>);
    EXPECT_TRUE(ranges::indirectly_readable<iterator2_t>); // this failed
    EXPECT_TRUE(std::input_iterator<iterator2_t>);
    EXPECT_TRUE(ranges::input_iterator<iterator2_t>); // this failed
```

marehr:
> > No that doesn't happen often, but it is unexpected if it happens and hard to diagnose. So I think giving a warning makes it easier to put you on the right track.
> 
> Diagnostics it is then :)
> I already updated the wiki about our include order (which is occasionally being consulted) and added a note that you need to `#include <seqan3/std/ranges>` before including any range-v3 header.

The diagnostics found some issues which is really nice :)

doc/tutorial/argument_parser/solution3.cpp
```diff
@@ -2,8 +2,6 @@
 #include <fstream>
 #include <numeric>
 
-#include <range/v3/view/split.hpp>
-
 #include <seqan3/argument_parser/all.hpp> // includes all necessary headers
 #include <seqan3/core/debug_stream.hpp>   // our custom output stream
 #include <seqan3/std/charconv>            // includes std::from_chars
```
eseiler, line 7: Can you sort? :D

doc/tutorial/sequence_file/sequence_file_snippets.cpp
```diff
@@ -1,5 +1,9 @@
 #include <fstream>
 #include <numeric> // std::accumulate
+//![include_ranges]
+#include <seqan3/std/ranges>
+//![include_ranges]
+
 //![include_ranges_chunk]
 #include <range/v3/view/chunk.hpp>
```
eseiler, line 8:
```suggestion
#include <seqan3/utility/views/chunk.hpp>
```

marehr: I fix that also in the rest of the header xD This PR now turned into https://github.com/seqan/product_backlog/issues/124

include/seqan3/utility/views/interleave.hpp
```diff
@@ -327,7 +326,7 @@ namespace seqan3::views
  * \header_file{seqan3/utility/views/interleave.hpp}
  *
  * This view can be used to insert one range into another range at regular intervals. It behaves essentially like
- * `| std::views::chunk(step_size) | views::join(inserted_range)` except that for input that models
+ * `| seqan3::views::chunk(step_size) | seqan3::views::join_with(inserted_range)` except that for input that models
```
SGSSGene, line 329: `join_with` doesn't exists. This will be fixed #2526 which will change it to `std::views::join`

include/seqan3/utility/views/interleave.hpp
```diff
@@ -352,7 +351,7 @@ namespace seqan3::views
  *
  *
  * If above requirements are not met, this adaptor forwards to
- * `| seqan3::views::chunk(step_size) | std::views::join(inserted_range)`
+ * `| seqan3::views::chunk(step_size) | seqan3::views::join_with(inserted_range)`
```
SGSSGene, line 354: same as above

SGSSGene (changes requested):

include/seqan3/utility/views/interleave.hpp
marehr: Hmm the code still uses join_with. Why would std::views::join be more correct?

marehr: but it will right?

marehr: I set https://github.com/seqan/seqan3/pull/2526 as  blocking for this PR; so if https://github.com/seqan/seqan3/pull/2526 is merged the comment should be correct, right?

SGSSGene: I didn't see the blocking PR, sorry. Yes, in that case it would be correct. But I don't know if  `std::views::join` should be preferred. #2526 replaces every call of `join_with` with `std::views::join` if no delimiter is being used.

marehr: Yeah, but here the semativ is to add a delimiter, right?

SGSSGene: Yes you are right! That means my change in #2526 is wrong :upside_down_face:

SGSSGene (approved):

PR #2588, opened 2021-05-02
[MISC] deprecate phred_base(phred_type) constructor

marehr:
Left over of https://github.com/seqan/product_backlog/issues/358


SGSSGene (approved):
Should work without flaw. But why replace the call to the base call with `assign_phred`. Any advantages or more personal taste?

marehr:
> Should work without flaw. But why replace the call to the base call with `assign_phred`. Any advantages or more personal taste?

To workaround annoying deprecation warnings xD

eseiler (approved):

PR #2589, opened 2021-05-03
Move range/container/*.hpp

marehr:
Resolves https://github.com/seqan/product_backlog/issues/376


test/include/seqan3/test/range/container_test_template.hpp
```diff
@@ -5,19 +5,18 @@
 // shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
 // -----------------------------------------------------------------------------------------------------
```
eseiler, line 7: Why is this now a header?
The templates are usually part of the respective directory of `test`

MitraDarja (approved):

test/include/seqan3/test/range/container_test_template.hpp
marehr: Yeah, here is the problem that it so spread out, that I think that it might be good just move it to the test header.

I thought about this for some time already, and I think that all test templates should be in the test header

This would enable us to use absolute paths `#include <seqan3/test/...>` that will not break when moving test files.

eseiler: Mhm ok, we should also do this for the other templates then.

marehr: I'll do that as follow-up :)

include/seqan3/utility/container/aligned_allocator.hpp
```diff
@@ -0,0 +1,223 @@
...
+    //!\brief The value type of the allocation.
+    using value_type = value_t;
+    //!\brief The pointer type of the allocation.
+    using pointer = value_type*;
+    //!\brief The difference type of the allocation.
+    using difference_type = typename std::pointer_traits<pointer>::difference_type;
+    //!\brief The size type of the allocation.
+    using size_type = std::make_unsigned_t<difference_type>;
+
+    //!\brief Are any two allocators of the same aligned_allocator type always compare equal?
```
eseiler, line 86:
```suggestion
    //!\brief Do any two allocators of the same aligned_allocator type always compare equal?
```

include/seqan3/utility/container/concept.hpp
```diff
@@ -0,0 +1,267 @@
...
+ * \brief Adaptations of concepts from the standard library.
+ * \author Hannes Hauswedell <hannes.hauswedell AT fu-berlin.de>
+ */
+
+#pragma once
+
+#include <initializer_list>
+#include <seqan3/std/iterator>
+#include <seqan3/std/concepts>
+#include <type_traits>
```
eseiler, line 18:
```suggestion

#include <seqan3/std/concepts>
#include <initializer_list>
#include <seqan3/std/iterator>
#include <type_traits>
```

eseiler (approved):

PR #2591, opened 2021-05-03
[INFRA] Add cron for latest libraries

eseiler:
Resolves https://github.com/seqan/product_backlog/issues/373

https://github.com/eseiler/seqan3/actions/runs/807164984

Luckily I got a kill signal, so the issue creation also works: https://github.com/eseiler/seqan3/issues/13


marehr (approved):
Very noice

eseiler:
`view_take_until_benchmark` has some trouble with gcc9 and latest gbenchmark.
Runs out of memory. I tried it locally, and I killed it when it reached 25GiB RAM

SGSSGene (approved):

PR #2601, opened 2021-05-04
Deprecate/take exactly

SGSSGene:
Fixes https://github.com/seqan/product_backlog/issues/331


MitraDarja (approved):

CHANGELOG.md
```diff
@@ -238,6 +238,8 @@ regression test suite and patches at https://github.com/seqan/seqan3/tree/master
   [\#2541](https://github.com/seqan/seqan3/pull/2541).
 * We deprecated `seqan3::views::take_line` and it will be removed in 3.1.0
   [\#2525](https://github.com/seqan/seqan3/pull/2525).
+* We deprecated `seqan3::views::take_exactly`. Please use `std::views::take` or `std::views::counted` instead
+  [\#2601](https://github.com/seqan/seqan3/pull/2601).
```
eseiler, line 242:
```suggestion
  ([\#2601](https://github.com/seqan/seqan3/pull/2601)).
```

eseiler (changes requested):
The idea was that we do not just deprecate it but move it to `detail` because we need it internally, correct?
If so, can you do it like we have done it with the `persist` view? In short, we moved the view itself into detail and did the deprecation of the old header and the old view a bit differently

SGSSGene:
> The idea was that we do not just deprecate it but move it to `detail` because we need it internally, correct?
> If so, can you do it like we have done it with the `persist` view? In short, we moved the view itself into detail and did the deprecation of the old header and the old view a bit differently

Yes, you are right. I changed it accordingly.

PR #2602, opened 2021-05-04
[Alignment] Deprecate `seqan3::alignment_coordinate`

marehr:
resolves https://github.com/seqan/product_backlog/issues/282


include/seqan3/alignment/matrix/all.hpp
```diff
@@ -18,7 +18,7 @@
  * \see alignment
  */
 
-#include <seqan3/alignment/matrix/alignment_coordinate.hpp>
+#include <seqan3/alignment/matrix/detail/advanceable_alignment_coordinate.hpp>
 #include <seqan3/alignment/matrix/alignment_optimum.hpp>
 #include <seqan3/alignment/matrix/debug_matrix.hpp>
```
Irallia, line 23:
```suggestion
#include <seqan3/alignment/matrix/alignment_optimum.hpp>
#include <seqan3/alignment/matrix/debug_matrix.hpp>
#include <seqan3/alignment/matrix/detail/advanceable_alignment_coordinate.hpp>
```

Irallia (changes requested):
LGFM!
The only thing missing is the lexicographical order in the `all.hpp`

include/seqan3/alignment/matrix/all.hpp
```diff
@@ -18,7 +18,7 @@
  * \see alignment
  */
 
-#include <seqan3/alignment/matrix/alignment_coordinate.hpp>
+#include <seqan3/alignment/matrix/detail/advanceable_alignment_coordinate.hpp>
```
marehr, line 21:
```suggestion
```

include/seqan3/alignment/matrix/all.hpp
marehr: I removed it :)

Irallia (approved):

include/seqan3/alignment/matrix/all.hpp
eseiler: the all-header shouldn't include detail stuff?

test/unit/alignment/matrix/alignment_coordinate_test.cpp
```diff
@@ -9,284 +9,10 @@
 
 #include <type_traits>
 
-#include <seqan3/alignment/matrix/detail/advanceable_alignment_coordinate.hpp>
+#include <seqan3/alignment/matrix/alignment_coordinate.hpp>
```
eseiler: Can you make this deprecation-warning safe?

include/seqan3/alignment/matrix/alignment_coordinate.hpp
```diff
@@ -7,320 +7,20 @@
 
 /*!\file
  * \author Marcel Ehrhardt <marcel.ehrhardt AT fu-berlin.de>
- * \brief Provides seqan3::detail::alignment_coordinate.
+ * \brief Provides seqan3::alignment_coordinate.
  */
```
eseiler, line 11: The file itself should be deprecated and there should be a
```
SEQAN3_DEPRECATED_HEADER(
  "This header is deprecated and will be removed along all its content in SeqAn-3.1.0.")
```

Or will something survive in this header?

eseiler (changes requested):
Some things regarding deprecation.
Also, a changelog entry is missing. (or is there already one?)

include/seqan3/alignment/matrix/alignment_coordinate.hpp
marehr: No it won't. Do you know how the tests will handle it if I include a deprecated header?

Normally we would just include the new header, but in this case there is no new header. So I opted to not deprecate the header.

I could move the deprecated alignment_coordinate again into the other header and include that...

include/seqan3/alignment/matrix/all.hpp
marehr: Hmm I already removed that :)

eseiler (approved):

PR #2605, opened 2021-05-05
[FIX] Handle unsupported tags

### LINKED ISSUE #2600: [BUG] SAM does not allow user-defined tags in header
While testing iGenVar with real data (PacBio long reads from https://ftp-trace.ncbi.nlm.nih.gov/ReferenceSamples/giab/data/AshkenazimTrio/HG002_NA24385_son/PacBio_CCS_10kb/alignment/) I ran into some errors:

```bash
terminate called after throwing an instance of 'seqan3::format_error'
  what():  Illegal SAM header tag in @HG starting with: p
```
from header file:
```sam
@HD	VN:1.5	SO:coordinate	pb:3.0.1
```
and without the `pb:3.0.1`
```bash
terminate called after throwing an instance of 'seqan3::format_error'
  what():  Illegal SAM header tag starting with: I
```
from a header line like
```sam
@RG	ID:35690b08	SM:HG002	PU:m54238_180628_014238	DS:READTYPE=CCS;BINDINGKIT=101-365-900;SEQUENCINGKIT=101-309-400;BASECALLERVERSION=5.0.0;FRAMERATEHZ=100.000000	PL:PACBIO	PM:SEQUEL
```

## Platform

  - SeqAn version: 3.0.3
  - Operating system: Debian GNU/Linux 10
  - Compiler: gcc (Debian 8.3.0-6) 8.3.0

### How to repeat the problem

```bash
git clone https://github.com/seqan/iGenVar.git
git submodule update --recursive --init
cmake ../../Repos/iGenVar/ -Weverything
make -j 16

wget https://ftp-trace.ncbi.nlm.nih.gov/ReferenceSamples/giab/data/AshkenazimTrio/HG002_NA24385_son/PacBio_CCS_10kb/alignment/HG002.Sequel.10kb.pbmm2.hs37d5.whatshap.haplotag.RTG.10x.trio.bam -nv

./build/iGenVar/bin/iGenVar -j HG002.Sequel.10kb.pbmm2.hs37d5.whatshap.haplotag.RTG.10x.trio.bam -o output.vcf
```

### Expected behaviour

iGenVar should run without errors

### Actual behaviour

`format_sam_base` throws `seqan3::format_error`

Irallia:
Resolves #2600

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -118,6 +118,10 @@ Note that 3.1.0 will be the first API stable release and interfaces in this rele
   * The processed data should experience no negative effects since `gz` and `bgzf` are **fully compatible**.
   * This bug may also cause unexpected parallelisation when reading `gz`-compressed input. This is the case when the
     `gz`-compressed input was also generated with SeqAn.
+* The following requirements of parsing a SAM/BAM header where relaxed as they were in violation of the specification
+  ([\#2605](https://github.com/seqan/seqan3/pull/2605)):
+  * The order of tags within the header may be arbitrary. Before, necessary tags where required to be first.
+  * Unknown tags will emit a warning. Before, a error was raised.
 
 ## API changes
 
--- a/include/seqan3/io/sam_file/detail/format_sam_base.hpp
+++ b/include/seqan3/io/sam_file/detail/format_sam_base.hpp
@@ -8,6 +8,7 @@
 /*!\file
  * \brief Provides the seqan3::format_sam_base that can be inherited from.
  * \author Svenja Mehringer <svenja.mehringer AT fu-berlin.de>
+ * \author Lydia Buntrock <lydia.buntrock AT fu-berlin.de>
  */
 
 #pragma once
@@ -63,7 +64,7 @@ protected:
     //!\brief A variable that tracks whether the content of header has been written or not.
     bool header_was_written{false};
 
-    //!\brief Tracks whether reference information (\@SR tag) were found in the SAM header
+    //!\brief Tracks whether reference information (\@SQ tag) were found in the SAM header
     bool ref_info_present_in_header{false};
 
     template <typename ref_id_type,
@@ -415,182 +416,355 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
         }
     };
 
-    auto skip_tag_name_and_colon = [&it] ()
+    auto skip_until_predicate = [&it] (auto const & predicate)
     {
-        while (!is_char<':'>(*it))
+        while (!predicate(*it))
             ++it;
-        ++it;
     };
 
     auto parse_tag_value = [&] (auto & value) // helper function to parse the next tag value
     {
-        skip_tag_name_and_colon();
+        skip_until_predicate(is_char<':'>);
+        ++it; // skip :
         take_until_predicate(is_char<'\t'> || is_char<'\n'>);
         read_field(string_buffer, value);
     };
 
-    auto parse_until_newline = [&] (auto & value)
+    auto parse_until_tab_or_newline = [&] (auto & value)
     {
-        take_until_predicate(is_char<'\n'>);
+        take_until_predicate(is_char<'\t'> || is_char<'\n'>);
+        string_buffer.push_back('\t');
         read_field(string_buffer, value);
+    };
+
+    auto print_cerr_of_unspported_tag = [&it] (char const * const header_tag, char const first_letter)
+    {
+        std::cerr << "Unsupported SAM header tag in @" << header_tag << ": " << first_letter << *it << '\n';
     };
 
     while (it != end && is_char<'@'>(*it))
     {
-        ++it;                                                           // skip @
+        ++it; // skip @
 
         switch (*it)
         {
-            case 'H': /* HD (header) tag */
+            case 'H': // HD (header) tag
             {
-                parse_tag_value(hdr.format_version);                    // parse required VN (version) tag
-
-                // The SO, SS and GO tag are optional and can appear in any order
-                while (is_char<'\t'>(*it))
+                ++it; // skip H
+                if (is_char<'D'>(*it))
                 {
-                    ++it;                                               // skip tab
-                    std::string * who = std::addressof(hdr.grouping);
+                    ++it; // skip D
 
-                    if (is_char<'S'>(*it))
+                    // All tags can appear in any order, VN is the only required tag
+                    while (is_char<'\t'>(*it))
                     {
-                        ++it;                                           // skip S
+                        std::string * header_entry{nullptr};
+                        ++it; // skip tab
+                        char const first_letter{*it};
+                        ++it;
 
-                        if (is_char<'O'>(*it))                          // SO (sorting) tag
-                            who = std::addressof(hdr.sorting);
-                        else if (is_char<'S'>(*it))                     // SS (sub-order) tag
-                            who = std::addressof(hdr.subsorting);
+                        switch (first_letter)
+                        {
+                            case 'V':
+                            {
+                                if (is_char<'N'>(*it)) // parse required VN (version) tag
+                                {
+                                    header_entry = std::addressof(hdr.format_version);
+                                    break;
+                                }
+                                [[fallthrough]];
+                            }
+                            case 'S':
+                            {
+                                if (is_char<'O'>(*it)) // SO (sorting) tag
+                                {
+                                    header_entry = std::addressof(hdr.sorting);
+                                    break;
+                                }
+                                if (is_char<'S'>(*it)) // SS (sub-order) tag
+                                {
+                                    header_entry = std::addressof(hdr.subsorting);
+                                    break;
+                                }
+                                [[fallthrough]];
+                            }
+                            case 'G':
+                            {
+                                if (is_char<'O'>(*it)) // GO (grouping) tag
+                                {
+                                    header_entry = std::addressof(hdr.grouping);
+                                    break;
+                                }
+                                [[fallthrough]];
+                            }
+                            default: // unsupported header tag
+                            {
+                                print_cerr_of_unspported_tag("HD", first_letter);
+                            }
+                        }
+
+                        if (header_entry != nullptr)
+                            parse_tag_value(*header_entry);
                         else
-                            throw format_error{std::string{"Illegal SAM header tag: S"} + *it};
+                            skip_until_predicate(is_char<'\t'> || is_char<'\n'>);
                     }
-                    else if (!is_char<'G'>(*it))                        // GO (grouping) tag
-                    {
-                        throw format_error{std::string{"Illegal SAM header tag in @HG starting with: "} + *it};
-                    }
+                    ++it; // skip newline
 
-                    parse_tag_value(*who);
-                }
-                ++it;                                                   // skip newline
-            } break;
-
-            case 'S': /* SQ (sequence dictionary) tag */
-            {
-                ref_info_present_in_header = true;
-                std::ranges::range_value_t<decltype(hdr.ref_ids())> id;
-                std::tuple<int32_t, std::string> info{};
-
-                parse_tag_value(id);                                    // parse required SN (sequence name) tag
-                ++it;                                                   // skip tab or newline
-                parse_tag_value(get<0>(info));                          // parse required LN (length) tag
-
-                if (is_char<'\t'>(*it))                                 // read rest of the tags
-                {
-                    ++it;                                               // skip tab
-                    parse_until_newline(get<1>(info));
-                }
-                ++it;                                                   // skip newline
-
-                /* If reference information was given, the ids exist and we can fill ref_dict directly.
-                 * If not, wee need to update the ids first and fill the reference dictionary afterwards. */
-                if constexpr (!detail::decays_to_ignore_v<ref_seqs_type>) // reference information given
-                {
-                    auto id_it = hdr.ref_dict.find(id);
-
-                    if (id_it == hdr.ref_dict.end())
-                        throw format_error{detail::to_string("Unknown reference name '", id, "' found in SAM header ",
-                                                            "(header.ref_ids(): ", hdr.ref_ids(), ").")};
-
-                    auto & given_ref_info = hdr.ref_id_info[id_it->second];
-
-                    if (std::get<0>(given_ref_info) != std::get<0>(info))
-                        throw format_error{"Provided reference has unequal length as specified in the header."};
-
-                    hdr.ref_id_info[id_it->second] = std::move(info);
+                    if (hdr.format_version.empty())
+                        throw format_error{std::string{"The required VN tag in @HD is missing."}};
                 }
                 else
-                {
-                    static_assert(!detail::is_type_specialisation_of_v<decltype(hdr.ref_ids()), std::deque>,
-                                "The range over reference ids must be of type std::deque such that "
-                                "pointers are not invalidated.");
+                    throw format_error{std::string{"Illegal SAM header tag: @H"} + *it};
 
-                    hdr.ref_ids().push_back(id);
-                    hdr.ref_id_info.push_back(info);
-                    hdr.ref_dict[(hdr.ref_ids())[(hdr.ref_ids()).size() - 1]] = (hdr.ref_ids()).size() - 1;
-                }
-            } break;
+                break;
+            }
 
-            case 'R': /* RG (read group) tag */
+            case 'S': // SQ (sequence dictionary) tag
             {
-                std::pair<std::string, std::string> tmp{};
-
-                parse_tag_value(get<0>(tmp));                           // read required ID tag
-
-                if (is_char<'\t'>(*it))                                 // read rest of the tags
+                ++it; // skip S
+                if (is_char<'Q'>(*it))
                 {
-                    ++it;
-                    parse_until_newline(get<1>(tmp));
-                }
-                ++it;                                                   // skip newline
+                    ++it; // skip Q
 
-                hdr.read_groups.emplace_back(std::move(tmp));
-            } break;
+                    ref_info_present_in_header = true;
+                    std::ranges::range_value_t<decltype(hdr.ref_ids())> id;
+                    std::tuple<int32_t, std::string> info{};
 
-            case 'P': /* PG (program) tag */
-            {
-                typename sam_file_header<ref_ids_type>::program_info_t tmp{};
-
-                parse_tag_value(tmp.id);                                // read required ID tag
-
-                // The PN, CL, PP, DS, VN are optional tags and can be given in any order.
-                while (is_char<'\t'>(*it))
-                {
-                    ++it;                                               // skip tab
-                    std::string * who = &tmp.version;
-
-                    switch (*it)
+                    // All tags can appear in any order, SN and LN are required tags
+                    while (is_char<'\t'>(*it))
                     {
-                        case 'P':
-                            ++it;                                       // skip P
+                        ++it; // skip tab
+                        char const first_letter{*it};
+                        ++it;
 
-                            if (is_char<'N'>(*it))                      // PN (program name) tag
-                                who = &tmp.name;
-                            else                                        // PP (previous program) tag
-                                who = &tmp.previous;
-                            break;
-
-                        case 'C': // CL (command line) tag
-                            who = &tmp.command_line_call;
-                            break;
-
-                        case 'D': // DS (description) tag
-                            who = &tmp.description;
-                            break;
-
-                        case 'V': // VN (version) tag
-                            // already set as default above
-                            break;
-
-                        default:
-                            throw format_error{std::string{"Illegal SAM header tag starting with: "} + *it};
+                        switch (first_letter)
+                        {
+                            case 'S': // parse required SN (sequence name) tag
+                            {
+                                if (is_char<'N'>(*it))
+                                {
+                                    parse_tag_value(id);
+                                    break;
+                                }
+                                [[fallthrough]];
+                            }
+                            case 'L': // parse required LN (length) tag
+                            {
+                                if (is_char<'N'>(*it))
+                                {
+                                    parse_tag_value(get<0>(info));
+                                    break;
+                                }
+                                [[fallthrough]];
+                            }
+                            default: // Any other tag
+                            {
+                                get<1>(info).push_back(first_letter);
+                                parse_until_tab_or_newline(get<1>(info));
+                            }
+                        }
                     }
+                    ++it; // skip newline
 
-                    parse_tag_value(*who);
+                    if (id.empty())
+                        throw format_error{std::string{"The required SN tag in @SQ is missing."}};
+                    if (get<0>(info) == int32_t{})
+                        throw format_error{std::string{"The required LN tag in @SQ is missing or 0."}};
+                    if (is_char<'\t'>(get<1>(info).back())) // See parse_until_tab_or_newline
+                        get<1>(info).pop_back();
+
+                    // If reference information was given, the ids exist and we can fill ref_dict directly.
+                    // If not, we need to update the ids first and fill the reference dictionary afterwards.
+                    if constexpr (!detail::decays_to_ignore_v<ref_seqs_type>) // reference information given
+                    {
+                        auto id_it = hdr.ref_dict.find(id);
+
+                        if (id_it == hdr.ref_dict.end())
+                            throw format_error{detail::to_string("Unknown reference name '", id, "' found in SAM ",
+                                                                 "header (header.ref_ids(): ", hdr.ref_ids(), ").")};
+
+                        auto & given_ref_info = hdr.ref_id_info[id_it->second];
+
+                        if (std::get<0>(given_ref_info) != std::get<0>(info))
+                            throw format_error{"Provided and header-based reference length differ."};
+
+                        hdr.ref_id_info[id_it->second] = std::move(info);
+                    }
+                    else
+                    {
+                        static_assert(!detail::is_type_specialisation_of_v<decltype(hdr.ref_ids()), std::deque>,
+                                      "The range over reference ids must be of type std::deque such that "
+                                      "pointers are not invalidated.");
+
+                        hdr.ref_ids().push_back(id);
+                        hdr.ref_id_info.push_back(info);
+                        hdr.ref_dict[(hdr.ref_ids())[(hdr.ref_ids()).size() - 1]] = (hdr.ref_ids()).size() - 1;
+                    }
                 }
-                ++it;                                                   // skip newline
+                else
+                    throw format_error{std::string{"Illegal SAM header tag: @S"} + *it};
 
-                hdr.program_infos.emplace_back(std::move(tmp));
-            } break;
+                break;
+            }
+
+            case 'R': // RG (read group) tag
+            {
+                ++it; // skip R
+                if (is_char<'G'>(*it))
+                {
+                    std::pair<std::string, std::string> tmp{};
+                    ++it; // skip G
+
+                    // All tags can appear in any order, SN and LN are required tags
+                    while (is_char<'\t'>(*it))
+                    {
+                        ++it; // skip tab
+                        char const first_letter{*it};
+                        ++it;
+
+                        switch (first_letter)
+                        {
+                            case 'I': // parse required ID tag
+                            {
+                                if (is_char<'D'>(*it))
+                                {
+                                    parse_tag_value(get<0>(tmp));
+                                    break;
+                                }
+                                [[fallthrough]];
+                            }
+                            default: // Any other tag
+                            {
+                                get<1>(tmp).push_back(first_letter);
+                                parse_until_tab_or_newline(get<1>(tmp));
+                            }
+                        }
+                    }
+                    ++it; // skip newline
+
+                    if (get<0>(tmp).empty())
+                        throw format_error{std::string{"The required ID tag in @RG is missing."}};
+                    if (is_char<'\t'>(get<1>(tmp).back())) // See parse_until_tab_or_newline
+                        get<1>(tmp).pop_back();
+
+                    hdr.read_groups.emplace_back(std::move(tmp));
+                }
+                else
+                    throw format_error{std::string{"Illegal SAM header tag: @R"} + *it};
+
+                break;
+            }
+
+            case 'P': // PG (program) tag
+            {
+                ++it; // skip P
+                if (is_char<'G'>(*it))
+                {
+                    typename sam_file_header<ref_ids_type>::program_info_t tmp{};
+                    ++it; // skip G
+
+                    // All tags can appear in any order, ID is the only required tag
+                    while (is_char<'\t'>(*it))
+                    {
+                        std::string * program_info_entry{nullptr};
+                        ++it; // skip tab
+                        char const first_letter{*it};
+                        ++it;
+
+                        switch (first_letter)
+                        {
+                            case 'I': // read required ID tag
+                            {
+                                if (is_char<'D'>(*it))
+                                {
+                                    program_info_entry = std::addressof(tmp.id);
+                                    break;
+                                }
+                                [[fallthrough]];
+                            }
+                            case 'P':
+                            {
+                                if (is_char<'N'>(*it)) // PN (program name) tag
+                                {
+                                    program_info_entry = std::addressof(tmp.name);
+                                    break;
+                                }
+                                if (is_char<'P'>(*it)) // PP (previous program) tag
+                                {
+                                    program_info_entry = std::addressof(tmp.previous);
+                                    break;
+                                }
+                                [[fallthrough]];
+                            }
+                            case 'C': // CL (command line) tag
+                            {
+                                if (is_char<'L'>(*it))
+                                {
+                                    program_info_entry = std::addressof(tmp.command_line_call);
+                                    break;
+                                }
+                                [[fallthrough]];
+                            }
+                            case 'D': // DS (description) tag
+                            {
+                                if (is_char<'S'>(*it))
+                                {
+                                    program_info_entry = std::addressof(tmp.description);
+                                    break;
+                                }
+                                [[fallthrough]];
+                            }
+                            case 'V': // VN (version) tag
+                            {
+                                if (is_char<'N'>(*it))
+                                {
+                                    program_info_entry = std::addressof(tmp.version);
+                                    break;
+                                }
+                                [[fallthrough]];
+                            }
+                            default: // unsupported header tag
+                            {
+                                print_cerr_of_unspported_tag("PG", first_letter);
+                            }
+                        }
+
+                        if (program_info_entry != nullptr)
+                            parse_tag_value(*program_info_entry);
+                        else
+                            skip_until_predicate(is_char<'\t'> || is_char<'\n'>);
+                    }
+                    ++it; // skip newline
+
+                    if (tmp.id.empty())
+                        throw format_error{std::string{"The required ID tag in @PG is missing."}};
+
+                    hdr.program_infos.emplace_back(std::move(tmp));
+                }
+                else
+                    throw format_error{std::string{"Illegal SAM header tag: @P"} + *it};
+
+                break;
+            }
 
             case 'C': /* CO (comment) tag */
             {
-                std::string tmp;
-                ++it;                                                   // skip C
-                ++it;                                                   // skip O
-                ++it;                                                   // skip :
-                parse_until_newline(tmp);
-                ++it;                                                   // skip newline
+                ++it; // skip C
+                if (is_char<'O'>(*it))
+                {
+                    std::string tmp;
+                    skip_until_predicate(is_char<'\t'>); // skip CO
+                    ++it; // skip tab
+                    take_until_predicate(is_char<'\n'>);
+                    read_field(string_buffer, tmp);
+                    ++it; // skip newline
+                    hdr.comments.emplace_back(std::move(tmp));
+                }
+                else
+                    throw format_error{std::string{"Illegal SAM header tag: @C"} + *it};
 
-                hdr.comments.emplace_back(std::move(tmp));
-            } break;
+                break;
+            }
 
             default:
                 throw format_error{std::string{"Illegal SAM header tag starting with:"} + *it};
--- a/include/seqan3/io/sam_file/detail/format_sam_base.hpp
+++ b/include/seqan3/io/sam_file/detail/format_sam_base.hpp
@@ -406,6 +406,22 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
     auto end = std::ranges::end(stream_view);
     std::vector<char> string_buffer{};
 
+    auto make_tag = [] (uint8_t char1, uint8_t char2) constexpr
+    {
+        return static_cast<uint16_t>(char1) | (static_cast<uint16_t>(char2) << CHAR_BIT);
+    };
+
+    std::array<char, 2> raw_tag{};
+
+    auto parse_and_make_tag = [&] ()
+    {
+        raw_tag[0] = *it;
+        ++it;
+        raw_tag[1] = *it;
+        ++it;
+        return make_tag(raw_tag[0], raw_tag[1]);
+    };
+
     auto take_until_predicate = [&it, &string_buffer] (auto const & predicate)
     {
         string_buffer.clear();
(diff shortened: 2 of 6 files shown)


include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -589,11 +595,14 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
                             break;
 
                         default:
-                            throw format_error{std::string{"Illegal SAM header tag starting with: "} + *it};
+                            seqan3::debug_stream << "Not supported SAM header tag in @PG starting with: " << *it << '\n';
                     }
 
                     parse_tag_value(*who);
                 }
+                if (!id_tag_given)
```
eseiler, line 603:
```suggestion
                if (tmp.id.empty())
```

Then you do not need the `id_tag_given`

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -22,8 +22,7 @@
 
 #include <seqan3/alphabet/views/char_to.hpp>
 #include <seqan3/alphabet/views/to_char.hpp>
-#include <seqan3/core/debug_stream/detail/to_string.hpp>
-#include <seqan3/core/detail/debug_stream_range.hpp>
+#include <seqan3/core/debug_stream.hpp>
```
eseiler, line 25: Can we get rid of this include?

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -481,9 +480,9 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
                         else
                             throw format_error{std::string{"Illegal SAM header tag: S"} + *it};
                     }
-                    else if (!is_char<'G'>(*it))                        // GO (grouping) tag
+                    else                                                // GO (grouping) tag, pb (PacBio) version, etc
                     {
-                        throw format_error{std::string{"Illegal SAM header tag in @HG starting with: "} + *it};
+                        seqan3::debug_stream << "Not supported SAM header tag in @HD starting with: " << *it << '\n';
```
eseiler, line 485:
```suggestion
                        std::cerr << "Not supported SAM header tag in @HD starting with: " << *it << '\n';
```

? debug_stream is a bit overkill since it includes half the library and we don't need to format stuff here

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -557,16 +556,23 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
             {
                 typename sam_file_header<ref_ids_type>::program_info_t tmp{};
 
-                parse_tag_value(tmp.id);                                // read required ID tag
+                ++it;                                                   // skip P
```
eseiler, line 559: the `RG` tag above has the same problem with `ID`

eseiler:
We also have the problem that we assume that the required tags are the first tags, which is not true.
Since this affects multiple header lines (`HD`, `SQ`, `RG`, and `PG`), we should probably do 2 PRs and just fix the `pb` issue here

include/seqan3/io/sam_file/detail/format_sam_base.hpp
Irallia: Ah, I did not know that. I thought that would be better than cerr, but no problem, I change that directly :) thanks for the tip

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -481,9 +480,9 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
                         else
                             throw format_error{std::string{"Illegal SAM header tag: S"} + *it};
                     }
-                    else if (!is_char<'G'>(*it))                        // GO (grouping) tag
+                    else                                                // GO (grouping) tag, pb (PacBio) version, etc
```
eseiler, line 483: You changed some logic here.

* `GO` tags will now emit warnings even though they shouldn't
* unknown tags are passed to `parse_tag_value(*who);` which might not be correct?
* * Maybe we need to skip the unknown tag with `take_until_predicate` but I am not sure.

Irallia: > - GO tags will now emit warnings even though they shouldn't

But we didn't do anything with the GO tag before, did we? So I thought I would include GO in the warning. But I can also undo that.

> - unknown tags are passed to parse_tag_value(*who); which might not be correct?
        - Maybe we need to skip the unknown tag with take_until_predicate but I am not sure.



`take_until_predicate` just skips everything until the next day? Well, since we don't support them, that sounds like the right way to go. Maybe @eldariont has a hint if it is worth to save them.

Irallia: We can also think about making a specific distinction between upper case and lower case. Because lower case comes from the user and upper case should be in the specification, if I understood correctly.

eseiler: > But we didn't do anything with the GO tag before, did we? So I thought I would include GO in the warning. But I can also undo that.

It was passed to `parse_tag_value(*who);` before.

> `take_until_predicate` just skips everything until the next day? 

`take_until_predicate(is_char<'\t'>)` or similar syntax should work

> Well, since we don't support them, that sounds like the right way to go. Maybe @eldariont has a hint if it is worth to save them.

We should support them at some point, but this would be a new feature (and would need to use the sam_tag_dictionary). For now, we should just ignore them.

eseiler: > We can also think about making a specific distinction between uppercase and lowercase. Because lower case comes from the user and upper case should be in the specification, if I understood correctly.

It's not forbidden to do user-defined uppercase tags, it's just that lowercase are reserved for this, and it's best practice to use lowercase.
I wouldn't make a distinction between lower/uppercase

Irallia: Okay, than I will skip them with your given syntax and open an issue for supporting them later.

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -481,9 +480,9 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
                         else
                             throw format_error{std::string{"Illegal SAM header tag: S"} + *it};
```
eldariont, line 481: We should not throw for unknown tags starting with S (e.g. SL or whatever) but just ignore them.

include/seqan3/io/sam_file/detail/format_sam_base.hpp
eldariont: I have never used any of the HD tags except SO (sorting order). Nevertheless, I agree with @eseiler that seqAn should not emit a warning for tags defined in the specs, like the GO tag.

Irallia:
I have removed a lot of headers as they are probably not needed in the file, but it would be nice if someone verifies this.

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -461,51 +445,125 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
+                            else
+                                throw format_error{std::string{"Illegal SAM header tag: S"} + *it};
+                        } break;
+                        case 'A':                                       // read rest of the tags (AH, AS
+                        case 'D':                                       // DS,
+                        case 'M':                                       // M5,
+                        case 'T':                                       // TP,
+                        case 'U':                                       // UR)
+                        {
+                            parse_until_tab_or_newline(get<1>(info));
```
eseiler, line 549: This will only store the last tag if there are multiple ones because you always parse until tab and overwrite `get<1>(info)`, no?

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -527,84 +585,152 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
+                        case 'B':                                       // read rest of the tags (BC,
+                        case 'C':                                       // CN,
+                        case 'D':                                       // DS, DT,
+                        case 'F':                                       // FO,
+                        case 'K':                                       // KS,
+                        case 'L':                                       // LB,
+                        case 'P':                                       // PG, PI, PL, PM, PU,
+                        case 'S':                                       // SM)
+                        {
+                            parse_until_tab_or_newline(get<1>(tmp));
```
eseiler, line 628: Same problem with overwriting if multiple tags

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -527,84 +585,152 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
-                    parse_tag_value(*who);
+                    if (no_default)
+                        parse_tag_value(*who);
                 }
-                ++it;                                                   // skip newline
-
+                if (tmp.id.empty())
+                    throw format_error{std::string{"The required ID tag is missing."}};
                 hdr.program_infos.emplace_back(std::move(tmp));
+                ++it;                                                   // skip newline
```
eseiler, line 724: Same comments as above

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -21,13 +21,13 @@ struct sam_file_read<seqan3::format_sam> : public sam_file_data
...
+@PG	ID:qc	PN:quality_control	CL:qc -f file1	DS:trim reads with low qual	VN:1.0.0	pb:usr_defined_tag
+@PG	PN:novoalign	ID:novoalign	VN:V3.02.07	CL:novoalign -d /path/hs37d5.ndx -f /path/file.fastq.gz	PP:qc
 @SQ	SN:ref	LN:249250621
-@SQ	SN:ref2	LN:243199373	AS:hs37d5
+@SQ	SN:ref2	LN:243199373	AS:hs37d5	pb:usr_defined_tag
 @RG	ID:U0a_A2_L1	PL:illumina	PU:1	LB:1	SM:NA12878
-@RG	ID:U0a_A2_L2	PL:illumina	SM:NA12878	PU:1	LB:1
+@RG	ID:U0a_A2_L2	PL:illumina	SM:NA12878	PU:1	LB:1	pb:usr_defined_tag
 @CO	Tralalalalalala this is a comment
 )"};
```
eseiler, line 32: Please add new test cases for the individual changes and leave the old ones as they are.
I.e. a two line header for swapping `PN` and `ID` and a two line header for adding a user defined tag

include/seqan3/io/sam_file/detail/format_sam_base.hpp
eseiler: open

include/seqan3/io/sam_file/format_bam.hpp
```diff
@@ -13,34 +13,22 @@
...
 
-#include <seqan3/alphabet/detail/convert.hpp>
 #include <seqan3/alphabet/nucleotide/dna16sam.hpp>
-#include <seqan3/core/debug_stream/detail/to_string.hpp>
-#include <seqan3/core/detail/debug_stream_optional.hpp>
-#include <seqan3/core/detail/template_inspection.hpp>
-#include <seqan3/core/range/type_traits.hpp>
-#include <seqan3/io/detail/ignore_output_iterator.hpp>
-#include <seqan3/io/detail/misc.hpp>
+#include <seqan3/core/debug_stream.hpp>
```
eseiler, line 22: Please include minimal debug_stream detail header instead

include/seqan3/io/sam_file/format_sam.hpp
```diff
@@ -12,36 +12,26 @@
...
 
 #include <seqan3/alphabet/concept.hpp>
-#include <seqan3/alphabet/views/char_to.hpp>
 #include <seqan3/alphabet/views/to_char.hpp>
-#include <seqan3/core/debug_stream/detail/to_string.hpp>
-#include <seqan3/core/detail/template_inspection.hpp>
-#include <seqan3/core/range/type_traits.hpp>
-#include <seqan3/io/detail/ignore_output_iterator.hpp>
-#include <seqan3/io/detail/misc.hpp>
+#include <seqan3/core/debug_stream.hpp>
```
eseiler, line 22: use minimal detail header

eseiler (changes requested):
I commented on things I saw.

I think after amending, the next step would be to add new tests.

Some parsing should not work correctly, i.e. parsing multiple "other" tags as only the last one will be stored.

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -433,23 +416,24 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
         take_until_predicate(is_char<'\t'> || is_char<'\n'>);
         read_field(string_buffer, value);
     };
 
-    auto parse_until_newline = [&] (auto & value)
+    auto parse_until_tab_or_newline = [&] (auto & value)
     {
-        take_until_predicate(is_char<'\n'>);
+        take_until_predicate(is_char<'\t'> || is_char<'\n'>);
+        string_buffer.push_back('\t');
```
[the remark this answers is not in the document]
Irallia: No, without the push_back() we would get unseperated tags:
```bash
header.read_groups[0]
    Which is: ("U0a_A2_L1", "PL:illuminaPU:1LB:1SM:NA12878")
  (std::pair<std::string, std::string>{"U0a_A2_L1", "PL:illumina\tPU:1\tLB:1\tSM:NA12878"})
    Which is: ("U0a_A2_L1", "PL:illumina\tPU:1\tLB:1\tSM:NA12878")
```

include/seqan3/io/sam_file/detail/format_sam_base.hpp
Irallia: No it apends the tags. Thats why I need to `push_back('\t’)`

include/seqan3/io/sam_file/detail/format_sam_base.hpp
Irallia: see above answer

include/seqan3/io/sam_file/detail/format_sam_base.hpp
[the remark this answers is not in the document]
eseiler: Ah.

If I see it correctly, you could just do:

```
take_until_predicate(is_char<'\t'> || is_char<'\n'>);
if (is_char<'\t'>(*it))
    string_buffer.push_back('\t');
read_field(string_buffer, value);
```

After `take_until_predicate` the iterator should point to a tab or newline, and you only need a tab in your buffer if you are not at the end.

Then you can also get rid of `pop_back` further down.

Irallia:
I have refactored a lot of the code, I hope I have seen all your comments and incorporated them correctly.

eseiler:
> I have refactored a lot of the code, I hope I have seen all your comments and incorporated them correctly.

This one is open: https://github.com/seqan/seqan3/pull/2605/files#r631864522

eseiler:
@Irallia I rebased on master and changed the commits a bit. There's one open review point.

include/seqan3/io/sam_file/detail/format_sam_base.hpp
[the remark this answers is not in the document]
Irallia: That would be nice, but this doesn't work eather. When we take the example again, the sam file line looks like:
`@RG	ID:U0a_A2_L1	PL:illumina	PU:1	LB:1	SM:NA12878`
but it could also look like
`@RG	PL:illumina	PU:1	LB:1	SM:NA12878	ID:U0a_A2_L1`
The ID is not a part of the second part of the `read_groups` tuple, thus the `\n` would not be at the end of the elements of this second part and we would never avoid the `if`.

->
```bash
.../seqan3/test/unit/io/sam_file/sam_file_format_test_template.hpp:183: Failure
  header.read_groups[0]
    Which is: ("U0a_A2_L1", "PL:illumina\tPU:1\tLB:1\tSM:NA12878\t")
  (std::pair<std::string, std::string>{"U0a_A2_L1", "PL:illumina\tPU:1\tLB:1\tSM:NA12878"})
    Which is: ("U0a_A2_L1", "PL:illumina\tPU:1\tLB:1\tSM:NA12878")
```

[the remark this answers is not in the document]
eseiler: Thanks for the explanation :)
Can you add this in a condensed form as comment to the lambda?

I'll take another look sometime later, but it doesn't seem like it can be simplified.

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -188,24 +188,36 @@ TEST_F(sam_format, header_errors)
...
+            Not supported SAM header tag in @PG: VB
+            Not supported SAM header tag in @PG: pb
+        */
         std::string header_str
         {
-            "@HD\tVN:1.6\tSO:coordinate\tSS:coordinate:queryname\tGO:none\tpb:usr_defined_tag\n"
-            "@PG\tID:qc\tPN:quality_control\tCL:qc -f file\tDS:trim reads with low qual\tVN:1.0.0\tpb:usr_defined_tag\n"
-            "@SQ\tSN:ref2\tLN:243199373\tAS:hs37d5\tpb:usr_defined_tag\n"
-            "@RG\tID:U0a_A2_L1\tPL:illumina\tPU:1\tLB:1\tSM:NA12878\tpb:usr_defined_tag\n"
+            "@HD\tVN:1.6\tVB:usr_defined_tag\tSB:usr_defined_tag\tGB:usr_defined_tag\tpb:usr_defined_tag\n"
```
eseiler, line 213: Can you check if something like here works?

https://github.com/seqan/seqan3/blob/929c355bd14caeb9761996b25dc9271379da776b/test/unit/argument_parser/format_parse_test.cpp#L24-L26 (but with `EXPECT_EQ([...]GetCapturedStderr(), "Unsupported...")`

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -188,24 +188,36 @@ TEST_F(sam_format, header_errors)
...
+            Not supported SAM header tag in @HD: VB
+            Not supported SAM header tag in @HD: SB
+            Not supported SAM header tag in @HD: GB
+            Not supported SAM header tag in @HD: pb
+            Not supported SAM header tag in @PG: IB
+            Not supported SAM header tag in @PG: PB
+            Not supported SAM header tag in @PG: CB
+            Not supported SAM header tag in @PG: DB
+            Not supported SAM header tag in @PG: VB
+            Not supported SAM header tag in @PG: pb
```
eseiler, line 209: Can you rename it to "Unsupported" ?

Irallia:
If the CI runs through and the codecoverage looks okay again, I think this PR is ready to review.

test/unit/io/sam_file/format_sam_test.cpp
Irallia: jepp that works!

eseiler:
Note to myself

current behaviour:
skip tags in `HD` and `PG` and warn 
do not warn in `RG` and `SQ`, tags are stored as string in tuple

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -536,217 +539,237 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
-                                ++it;                                       // skip I
-                                if (is_char<'D'>(*it))                      // read required ID tag
+                                if (is_char<'D'>(*it))
                                 {
                                     parse_tag_value(get<0>(tmp));
+                                    break;
                                 }
-                                else                                        // parse other user defined tag with D..
-                                {
-                                    get<1>(tmp).push_back('D');
```
eseiler: this was wrong

eseiler (changes requested):
I refactored the code a bit (with using `[[fallthrough]];` to reduce code duplication).

I think after adding tests for the remaining missing coverage, this is good to go

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -125,64 +125,224 @@ struct sam_format : public sam_file_data
...
     {
         std::string header_str
         {
-            "@HD\tVN:1.0\tTT:this is not a valid tag\n"
+            "@HA\tthis is not a valid tag\n"
         };
         std::istringstream istream(header_str);
         seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
-        EXPECT_THROW(fin.begin(), seqan3::format_error);
+        EXPECT_THROW(fin.begin(), seqan3::format_error);    // "Illegal SAM header tag: @HA"
```
eseiler, line 137:
```suggestion
        EXPECT_THROW(fin.begin(), seqan3::format_error);
```

Until we actually test the error messages (which we currently cannot), I wouldn't add the expected error message as they might change and it's quite clear what you are testing (`// unsupported @-tag tests` + `@HA this is not a valid tag`)

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -125,64 +125,224 @@ struct sam_format : public sam_file_data
...
+    }
+    {
+        std::string header_str
+        {
+            "@PA\tthis is not a valid tag\n"
+        };
+        std::istringstream istream(header_str);
+        seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
+        EXPECT_THROW(fin.begin(), seqan3::format_error);    // "Illegal SAM header tag: @PA"
+    }
```
eseiler, line 156:
```suggestion
    { // invalid header record type: @PA
        std::string header_str
        {
            "@PA\tthis is not a valid tag\n"
        };
        std::istringstream istream(header_str);
        seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
        EXPECT_THROW(fin.begin(), seqan3::format_error);
    }
```

etc.

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -125,64 +125,224 @@ struct sam_format : public sam_file_data
...
-            "@HD\tVN:1.0\n"
             "@TT\tthis is not a valid tag\n"
         };
         std::istringstream istream(header_str);
         seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
-        EXPECT_THROW(fin.begin(), seqan3::format_error);
+        EXPECT_THROW(fin.begin(), seqan3::format_error);    // "Illegal SAM header tag starting with:T"
     }
+    // unusual ordering of tags
     {
```
eseiler, line 185:
```suggestion
    { // order of tags does not matter
```

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -125,64 +125,224 @@ struct sam_format : public sam_file_data
...
-            "@PG\tID:prog\tTT:this is not a valid tag\n"
+            "@HD\tGO:none\tSO:coordinate\tVN:1.6\tSS:coordinate:queryname\n"
+            "@PG\tPN:novoalign\tPP:qc\tID:novoalign\tVN:V3.02.07\tCL:novoalign -d /hs37d5.ndx -f /file.fastq.gz\n"
+            "@SQ\tAS:hs37d5\tSN:ref2\tLN:243199373\n"
+            "@RG\tLB:1\tSM:NA12878\tPL:illumina\tPU:1\tID:U0a_A2_L1\n"
         };
         std::istringstream istream(header_str);
         seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
-        EXPECT_THROW(fin.begin(), seqan3::format_error);
+        EXPECT_NO_THROW(fin.begin());                       // Order of tags should not trigger errors
```
eseiler, line 195:
```suggestion
        EXPECT_NO_THROW(fin.begin());
```

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -125,64 +125,224 @@ struct sam_format : public sam_file_data
...
+    }
+    // user defined tags -> cerr of unspported tags
+    {
+        // User defined tags should not trigger errors,but prints in cerr:
```
eseiler:
```suggestion
    { // User defined tags should not trigger errors, but print warnings to cerr
```

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -125,64 +125,224 @@ struct sam_format : public sam_file_data
...
+        std::string header_str
+        {
+            "@HD\tVN:1.6\tVB:usr_defined_tag\tSB:usr_defined_tag\tGB:usr_defined_tag\tpb:usr_defined_tag\n"
+            "@SQ\tSN:ref2\tLN:243199373\tSB:usr_defined_tag\tLB:usr_defined_tag\tpb:usr_defined_tag\n"
+            "@RG\tID:U0a_A2_L1\tIB:usr_defined_tag\tpb:usr_defined_tag\n"
+            "@PG\tID:qc\tIB:usr_defined_tag\tPB:usr_defined_tag\tCB:usr_defined_tag\tDB:usr_defined_tag\tVB:usr_defined_tag\tpb:usr_defined_tag\n"
```
eseiler, line 205: `usr_defined_tag` -> `user_tag`

suggestion won't let me suggest

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -125,64 +125,224 @@ struct sam_format : public sam_file_data
...
+        };
+        std::string cerr_of_unspported_tags
```
eseiler, line 207:
```suggestion
        std::string expected_cerr
```

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -125,64 +125,224 @@ struct sam_format : public sam_file_data
...
+    // @HD tests
+    {
+        std::string header_str
+        {
+            "@HD\n"                 // required VN tag missing
+        };
+        std::istringstream istream(header_str);
+        seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
+        EXPECT_THROW(fin.begin(), seqan3::format_error);    // "The required VN tag in @HD is missing."
+    }
```
eseiler, line 236:
```suggestion
    { // Required tags in HD
        std::string header_str
        {
            "@HD\n" // VN is missing
        };
        std::istringstream istream(header_str);
        seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
        EXPECT_THROW(fin.begin(), seqan3::format_error);
    }
```

etc

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -125,64 +125,224 @@ struct sam_format : public sam_file_data
...
+    {
+        std::string header_str
+        {
+            "@SQ\tSN:ref\n"
+        };
+        std::istringstream istream(header_str);
+        seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
+        EXPECT_THROW(fin.begin(), seqan3::format_error);    // "The required LN tag in @SQ is missing or 0."
     }
     {
```
eseiler, line 256:
```suggestion
    { // Unknown reference name in SQ
```

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -125,64 +125,224 @@ struct sam_format : public sam_file_data
...
-            "@HD\tVN:1.0\n"
-            "@SQ\tSN:unknown_ref\tLN:0\n"
+            "@SQ\tSN:unknown_ref\tLN:1\n"
         };
         std::istringstream istream(header_str);
         seqan3::sam_file_input fin{istream, this->ref_ids, this->ref_sequences, seqan3::format_sam{}};
-        EXPECT_THROW(fin.begin(), seqan3::format_error);
+        EXPECT_THROW(fin.begin(), seqan3::format_error);    // "Unknown reference name 'unknown_ref' found in SAM header
+                                                            //  (header.ref_ids(): [ref])."
+    }
```
eseiler, line 265:
```suggestion
        EXPECT_THROW(fin.begin(), seqan3::format_error);
    }
```

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -125,64 +125,223 @@ struct sam_format : public sam_file_data
...
+    {
+        std::string header_str
+        {
+            "@SQ\tSN:ref\tLN:-10\n"                          // allowed range of LN: [1,2^31−1]
+        };
+        std::istringstream istream(header_str);
+        seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
+        EXPECT_THROW(fin.begin(), seqan3::format_error);    // "Provided reference has unequal length as specified in
+                                                            //  the header."
+    }
```
eseiler, line 294: seems redundant

eseiler (changes requested):
Some more comments regarding test; I added a commit that should enable the LN checks

include/seqan3/io/sam_file/detail/format_sam_base.hpp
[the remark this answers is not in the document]
eseiler: I managed to find a way

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -433,182 +416,362 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
+                                                                 "header (header.ref_ids(): ", hdr.ref_ids(), ").")};
 
-                    if (id_it == hdr.ref_dict.end())
-                        throw format_error{detail::to_string("Unknown reference name '", id, "' found in SAM header ",
-                                                            "(header.ref_ids(): ", hdr.ref_ids(), ").")};
+                        auto & given_ref_info = hdr.ref_id_info[id_it->second];
 
-                    auto & given_ref_info = hdr.ref_id_info[id_it->second];
+                        if (std::get<0>(given_ref_info) != std::get<0>(info))
+                            throw format_error{"Provided reference and reference length provided in header differ."};
```
eseiler, line 599:
```suggestion
                            throw format_error{"Provided and header-based reference length differ."};
```

Irallia:
Thanks for your work! I changed the test documentation, as you mentioned, and added a changelog entry.

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -125,49 +125,133 @@ struct sam_format : public sam_file_data
...
+    { // invalid header record type: @CA
+        std::string header_str
+        {
+            "@CA\tthis is not a valid tag\n"
+        };
+        std::istringstream istream(header_str);
+        seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
+        EXPECT_THROW(fin.begin(), seqan3::format_error);
+    }
+    {{ // invalid header record type: @TT
```
eseiler, line 174:
```suggestion
    { // invalid header record type: @TT
```

Irallia: jepp already saw it :)

CHANGELOG.md
```diff
@@ -106,6 +106,9 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
   * The processed data should experience no negative effects since `gz` and `bgzf` are **fully compatible**.
   * This bug may also cause unexpected parallelisation when reading `gz`-compressed input. This is the case when the
     `gz`-compressed input was also generated with SeqAn.
+* Parsing sam & bam files caused some errors related to their headers Now, as described in the specification, the
+  ordering of the header tags doesn't matter anymore and no error is given for unsupported tags
+  ([\#2605](https://github.com/seqan/seqan3/pull/2605)).
```
eseiler, line 111:
```suggestion
* The following requirements for parsing a SAM/BAM header where relaxed as they where in violation of the specification 
  ([\#2605](https://github.com/seqan/seqan3/pull/2605)):
  * The order of tags within the header may be arbitrary. Before, necessary tags where required to be first.
  * Unknown tags will emit a warning. Before, a error was raised.
```

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -125,49 +125,133 @@ struct sam_format : public sam_file_data
...
+    // @SQ tests
+    {
+        std::string header_str
+        {
+            "@SQ\tLN:1\n"
+        };
+        std::istringstream istream(header_str);
+        seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
+        EXPECT_THROW(fin.begin(), seqan3::format_error);    // "The required SN tag in @SQ is missing."
+    }
```
eseiler, line 241:
```suggestion
    { // missing SN tag in @SQ
        std::string header_str
        {
            "@SQ\tLN:1\n"
        };
        std::istringstream istream(header_str);
        seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
        EXPECT_THROW(fin.begin(), seqan3::format_error);
    }
```

also for the others.

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -125,49 +125,133 @@ struct sam_format : public sam_file_data
...
+        testing::internal::CaptureStderr();
+        EXPECT_NO_THROW(fin.begin());
+        EXPECT_EQ(testing::internal::GetCapturedStderr(), expected_cerr);
+    }
+    { // Required tags in HD
         std::string header_str
         {
-            "@HD\tVN:1.0\n"
-            "@PG\tID:prog\tTT:this is not a valid tag\n"
+            "@HD\n" // VN is missing
```
eseiler, line 226:
```suggestion
            "@HD\n"
```

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -125,49 +125,133 @@ struct sam_format : public sam_file_data
...
+            "Unsupported SAM header tag in @PG: pb\n"
+        };
+        std::istringstream istream(header_str);
+        seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
+
+        testing::internal::CaptureStderr();
+        EXPECT_NO_THROW(fin.begin());
+        EXPECT_EQ(testing::internal::GetCapturedStderr(), expected_cerr);
+    }
+    { // Required tags in HD
```
eseiler, line 223:
```suggestion
    { // missing VN tag in @HD
```

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -125,49 +125,133 @@ struct sam_format : public sam_file_data
...
+    {
+        std::string header_str
+        {
+            "@SQ\tLN:1\n"
+        };
+        std::istringstream istream(header_str);
+        seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
+        EXPECT_THROW(fin.begin(), seqan3::format_error);    // "The required SN tag in @SQ is missing."
+    }
     {
```
eseiler, line 242:
```suggestion
    { // missing LN tag in @SQ
```

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -125,49 +125,133 @@ struct sam_format : public sam_file_data
...
         std::string header_str
         {
-            "@HD\tVN:1.0\n"
-            "@SQ\tSN:unknown_ref\tLN:0\n"
+            "@SQ\tSN:ref\n"
+        };
+        std::istringstream istream(header_str);
+        seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
+        EXPECT_THROW(fin.begin(), seqan3::format_error);    // "The required LN tag in @SQ is missing or 0."
```
eseiler, line 249:
```suggestion
        EXPECT_THROW(fin.begin(), seqan3::format_error);
```

Also, this would raise an error that LN is missing since we handle that case explicitly

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -176,13 +260,72 @@ TEST_F(sam_format, header_errors)
     {
         std::string header_str
         {
-            "@HD\tVN:1.0\n"
-            "@SQ\tSN:ref\tLN:0\n" /*wrong length*/
+            "@SQ\tSN:ref\tLN:0\n"                           // allowed range of LN: [1,2^31−1]
```
eseiler, line 263:
```suggestion
            "@SQ\tSN:ref\tLN:0\n"
```

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -176,13 +260,72 @@ TEST_F(sam_format, header_errors)
     {
```
eseiler, line 260:
```suggestion
    { // LN cannot be 0
```

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -176,13 +260,72 @@ TEST_F(sam_format, header_errors)
...
         std::string header_str
         {
-            "@HD\tVN:1.0\n"
-            "@SQ\tSN:ref\tLN:0\n" /*wrong length*/
+            "@SQ\tSN:ref\tLN:0\n"                           // allowed range of LN: [1,2^31−1]
+        };
+        std::istringstream istream(header_str);
+        seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
+        EXPECT_THROW(fin.begin(), seqan3::format_error);    // "The required LN tag in @SQ is missing or 0."
```
eseiler, line 267:
```suggestion
        EXPECT_THROW(fin.begin(), seqan3::format_error);
```

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -176,13 +260,72 @@ TEST_F(sam_format, header_errors)
...
+    {
+        std::string header_str
+        {
+            "@SQ\tSN:ref\tLN:-1\n"                          // allowed range of LN: [1,2^31−1]
+        };
+        std::istringstream istream(header_str);
+        seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
+        EXPECT_THROW(fin.begin(), seqan3::format_error);    // "Provided reference has unequal length as specified in
+                                                            //  the header."
+    }
```
eseiler, line 278:
```suggestion
    { // LN cannot be negative
        std::string header_str
        {
            "@SQ\tSN:ref\tLN:-1\n"
        };
        std::istringstream istream(header_str);
        seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
        EXPECT_THROW(fin.begin(), seqan3::format_error);
    }
```

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -176,13 +260,72 @@ TEST_F(sam_format, header_errors)
...
+    {
+        std::string header_str
+        {
+            "@SQ\tSN:ref\tLN:2147483647\n"                  // = 2^31−1, allowed range of LN: [1,2^31−1]
+        };
+        std::istringstream istream(header_str);
+        seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
+        EXPECT_NO_THROW(fin.begin());
+    }
```
eseiler, line 287:
```suggestion
    { // maximum LN value is 2^31-1
        std::string header_str
        {
            "@SQ\tSN:ref\tLN:2147483647\n"
        };
        std::istringstream istream(header_str);
        seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
        EXPECT_NO_THROW(fin.begin());
    }
```

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -176,13 +260,72 @@ TEST_F(sam_format, header_errors)
...
+    {
+        std::string header_str
+        {
+            "@SQ\tSN:ref\tLN:2147483648\n"                  // = 2^31, allowed range of LN: [1,2^31−1]
+        };
+        std::istringstream istream(header_str);
+        seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
+        EXPECT_THROW(fin.begin(), seqan3::format_error);    // "[CORRUPTED SAM FILE] Casting '2147483648' into type int
+                                                            //  would cause an overflow."
+    }
```
eseiler, line 297:
```suggestion
    { // LN exceeds maximum value
        std::string header_str
        {
            "@SQ\tSN:ref\tLN:2147483648\n"
        };
        std::istringstream istream(header_str);
        seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
        EXPECT_THROW(fin.begin(), seqan3::format_error);
    }
```

test/unit/io/sam_file/format_sam_test.cpp
eseiler, line 242: Can you cluster all `LN` related tests together?

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -176,13 +260,72 @@ TEST_F(sam_format, header_errors)
...
+    {
```
eseiler, line 298:
```suggestion
    { // Provided and header-based reference length differ
```

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -176,13 +260,72 @@ TEST_F(sam_format, header_errors)
...
+        std::string header_str
+        {
+            "@SQ\tSN:ref\tLN:4\n"
         };
         std::istringstream istream(header_str);
         seqan3::sam_file_input fin{istream, this->ref_ids, this->ref_sequences, seqan3::format_sam{}};
-        EXPECT_THROW(fin.begin(), seqan3::format_error);
+        EXPECT_THROW(fin.begin(), seqan3::format_error);    // "Provided reference has unequal length as specified in
+                                                            //  the header."
```
eseiler, line 306:
```suggestion
        EXPECT_THROW(fin.begin(), seqan3::format_error);
```

(the message is also different now)

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -176,13 +260,72 @@ TEST_F(sam_format, header_errors)
...
+    // @RG tests
+    {
+        std::string header_str
+        {
+            "@RG\n"
+        };
+        std::istringstream istream(header_str);
+        seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
+        EXPECT_THROW(fin.begin(), seqan3::format_error);    // "The required ID tag in @RG is missing."
+    }
```
eseiler, line 317:
```suggestion
    { // missing ID tag in @RG
        std::string header_str
        {
            "@RG\n"
        };
        std::istringstream istream(header_str);
        seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
        EXPECT_THROW(fin.begin(), seqan3::format_error);
    }
```

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -176,13 +260,72 @@ TEST_F(sam_format, header_errors)
...
+    // @PG tests
+    {
+        std::string header_str
+        {
+            "@PG\n"
+        };
+        std::istringstream istream(header_str);
+        seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
+        EXPECT_THROW(fin.begin(), seqan3::format_error);    // "The required ID tag in @PG is missing."
```
eseiler, line 326:
```suggestion
    { // missing ID tag in @PG
        std::string header_str
        {
            "@PG\n"
        };
        std::istringstream istream(header_str);
        seqan3::sam_file_input fin{istream, seqan3::format_sam{}};
        EXPECT_THROW(fin.begin(), seqan3::format_error);
```

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -176,13 +260,72 @@ TEST_F(sam_format, header_errors)
...
     }
+
```
eseiler, line 328:
```suggestion
```

eseiler (changes requested):
Last bit of 💅 :)

Irallia:
I can tidy up the history, after your approve.

eseiler (approved):
LGTM, please rebase/cleanup and then request second review :)

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -433,182 +416,362 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
+                                if (is_char<'N'>(*it)) // parse required VN (version) tag
+                                {
+                                    header_entry = std::addressof(hdr.format_version);
+                                    break;
+                                }
+                                [[fallthrough]];
+                            }
+                            case 'S':
+                            {
+                                if (is_char<'O'>(*it)) // SO (sorting) tag
```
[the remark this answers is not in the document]
marehr: Why even use a switch here? Wouldn't an if case in this case be better?

Why not read in the second letter in a single 2byte word (we could use a switch in that case):

```c++
    // For prove of concept: https://godbolt.org/z/ab36EzYvG
    auto make_tag = [](uint8_t char1, uint8_t char2) constexpr
    {
        return static_cast<uint16_t>(char1) | (static_cast<uint16_t>(char2) << CHAR_BIT);
    };

    char [2] raw_tag{};
    raw_tag[0] = *it;
    ++it;
    raw_tag[1] = *it;
    ++it;
    uint16_t tag = make_tag(raw_tag[0], raw_tag[1]);

    switch (tag)
    {
        case make_tag('V', 'N'): // parse required VN (version) tag
            header_entry = std::addressof(hdr.format_version);
            break;
        case make_tag('S', 'O'): // SO (sorting) tag
            header_entry = std::addressof(hdr.sorting);
            break;
        case make_tag('S', 'S'):  // SS (sub-order) tag
            header_entry = std::addressof(hdr.sorting);
            break;
        // ...
        default:
            rint_cerr_of_unspported_tag("HD", raw_tag[0], raw_tag[1]);
            break;
    }
```

I know that is asking a lot, but can we make this change to simplify this complete section?

marehr (changes requested):
I think we should do the tag handling differently if we touch the file anyway.

I proposed a change to flatten the nested-ness of all if/switch.

CHANGELOG.md
```diff
@@ -106,6 +106,10 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
   * The processed data should experience no negative effects since `gz` and `bgzf` are **fully compatible**.
   * This bug may also cause unexpected parallelisation when reading `gz`-compressed input. This is the case when the
     `gz`-compressed input was also generated with SeqAn.
+* The following requirements of parsing a SAM/BAM header where relaxed as they were in violation of the specification
+  ([\#2605](https://github.com/seqan/seqan3/pull/2605)):
+  * The order of tags within the header may be arbitrary. Before, necessary tags where required to be first.
+  * Unknown tags will emit a warning. Before, a error was raised.
```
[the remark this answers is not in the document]
eseiler: It was an error before, but is only a warning via cerr now.

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -125,64 +125,200 @@ struct sam_format : public sam_file_data
 // since BAM uses the same read header function from SAM, it only needs to be tested once
 TEST_F(sam_format, header_errors)
 {
-    {
+
+    { // invalid header record type: @HA
         std::string header_str
         {
-            "@HD\tVN:1.0\tTT:this is not a valid tag\n"
```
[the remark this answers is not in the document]
Irallia: A header contains header lines starting with @. Each @Header Tag is optional. That means that we dont need the @HD... when we test something else. So I have simplified the example headers to include only what is necessary.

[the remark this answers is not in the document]
Irallia: You can find the information here: https://samtools.github.io/hts-specs/SAMv1.pdf (Section 1.3)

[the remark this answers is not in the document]
Irallia: This means the old test cases are rephrased and simplified.

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -63,7 +64,7 @@ class format_sam_base
     //!\brief A variable that tracks whether the content of header has been written or not.
     bool header_was_written{false};
 
-    //!\brief Tracks whether reference information (\@SR tag) were found in the SAM header
+    //!\brief Tracks whether reference information (\@SQ tag) were found in the SAM header
```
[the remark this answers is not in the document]
Irallia: Yes it was a typo. There are only the following @Header Tags existing: @HD, @SQ, @RG, @PG and @CO. And here we have a switch case for SQ.

test/unit/io/sam_file/format_sam_test.cpp
```diff
@@ -125,64 +125,200 @@ struct sam_format : public sam_file_data
 // since BAM uses the same read header function from SAM, it only needs to be tested once
```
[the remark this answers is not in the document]
Irallia: Yes, we will accept them and warn, that we don't process them.

> This PR will fail the api-stability test:

What does that mean? Why and what can I do about it?

include/seqan3/io/sam_file/detail/format_sam_base.hpp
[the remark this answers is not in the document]
Irallia: Oha, I have not seen a way to check two char's at once. Of course, that would be much clearer. I will try it out.
The fallthrough I also did not know, that came from @eseiler. 
All these new things I get to know :)

test/unit/io/sam_file/format_sam_test.cpp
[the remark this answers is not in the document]
marehr: That means that seqan3/test/api_stability will fail. I'll help you with that 👍🏽

test/unit/io/sam_file/format_sam_test.cpp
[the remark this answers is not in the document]
marehr: Okay, but on the other side we still have `@HD` tests, right?

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -433,182 +422,307 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
+        std::cerr << "Unsupported SAM header tag in @" << header_tag << ": " << raw_tag[0] << raw_tag[1] << '\n';
+    };
+
     while (it != end && is_char<'@'>(*it))
     {
-        ++it;                                                           // skip @
+        ++it; // skip @
+
+        // char[2] raw_at_tag{};
+        std::array<char, 2> raw_at_tag{};
```
Irallia, line 466: `char [2]` did not work, but `std::array<char, 2>` worked.

```bash
.../seqan3/include/seqan3/io/sam_file/detail/format_sam_base.hpp:464:14: error: expected identifier before numeric constant
  464 |         char[2] raw_at_tag{};
      |              ^
```

marehr:
```c++
char raw_at_tag[2]{};
```

is the correct syntax :)

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -430,346 +436,291 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
-    // Some tags are not parsed individually. Instead, these are simply copied into a string.
+    // Some tags are not parsed individually. Instead, these are simply copied into a std::string.
     // Multiple tags must be separated by a `\t`, hence we prepend a tab to the string, except the first time.
     // Alternatively, we could always append a `\t`, but this would have the side effect that we might need to trim a
     // trailing tab after parsing all tags via `pop_back()`.
     // Unfortunately, we do not know when we are parsing the last tag (and in this case just not append a tab),
     // because even though we can check if the line ends in a `\n`, it is not guaranteed that the last tag of the
     // line is passed to this lambda. For example, the line might end with a tag that is properly parsed, such as `ID`.
-    auto parse_until_tab_or_newline = [&] (std::string & value, char const first_letter)
+    auto parse_until_tab_or_newline = [&] (std::string & value, std::array<char, 2> raw_tag)
```
marehr, line 446: this name is somehow wrong, because it adds the raw_tag, it seems like a function that is used only in error cases?

include/seqan3/io/sam_file/detail/format_sam_base.hpp
marehr: but we can keep the std::array

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -433,182 +422,307 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
+    {
+        std::cerr << "Unsupported SAM header tag in @" << header_tag << ": " << raw_tag[0] << raw_tag[1] << '\n';
+    };
+
     while (it != end && is_char<'@'>(*it))
     {
-        ++it;                                                           // skip @
+        ++it; // skip @
+
+        // char[2] raw_at_tag{};
```
marehr, line 465:
```suggestion
```

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -423,6 +406,12 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
     auto end = std::ranges::end(stream_view);
     std::vector<char> string_buffer{};
 
+    // For prove of concept: https://godbolt.org/z/ab36EzYvG
```
marehr, line 409:
```suggestion
```

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -433,182 +422,307 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
-                    parse_until_newline(get<1>(info));
-                }
-                ++it;                                                   // skip newline
+                    ++it; // skip tab
+
+                    std::array<char, 2> raw_tag{};
+                    raw_tag[0] = *it;
+                    ++it;
+                    raw_tag[1] = *it;
+                    ++it;
```
marehr, line 547: maybe we should put that into a function, as we do that multiple times:

```c++
auto read_tag = [](auto it, std::array<char, 2> & tag)
{
    raw_tag[0] = *it;
    ++it;
    raw_tag[1] = *it;
    ++it;
    return it;
}

std::array<char, 2> raw_tag{};
it = read_tag(it, raw_tag);
```

marehr (commented):
That's definitely better! 

I would dry (=don't repeat yourself) reading in the tag and I think this PR is good.

include/seqan3/io/sam_file/detail/format_sam_base.hpp
eseiler: why copy the iterator and not just capture ?

marehr: We can also take it via lvalue. I'm always a bit confused, because the standard often returns an iterator as copy after an operation.

eseiler: I think I would prefer `[&it]` or `[&]` because we do it for other lambdas.
Maybe we could also have a function that directly returns the `uint16_t` tag since we need to do it anyways.

include/seqan3/io/sam_file/detail/format_sam_base.hpp
Irallia: I think I dont understand what you mean.
This function is used to parse tags plus their value into a string, as these tags are not handled properly. So we parse these tags until their end (tab odr newline) into this string of kind of unhandled tags.

test/unit/io/sam_file/format_sam_test.cpp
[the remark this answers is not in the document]
Irallia: yepp, the test of lines 183, 194, 226. And the tests with normal headers which i left untouched (e.g. line 24)

Irallia (commented):
Found one useless code line

include/seqan3/io/sam_file/detail/format_sam_base.hpp
Irallia: So you mean to put these into the lambda function aswell?

```cpp
                    std::array<char, 2> raw_tag{};
                    it = read_tag(it, raw_tag);
                    uint16_t tag = make_tag(raw_tag[0], raw_tag[1]);
```

Irallia: so the idea is:
```cpp
std::array<char, 2> raw_tag{};

uint16_t get_tag = [*it](std::array<char, 2> & raw_tag)
{
    raw_tag[0] = *it;
    ++it;
    raw_tag[1] = *it;
    ++it;
    return make_tag(raw_tag[0], raw_tag[1]); 
}

switch (get_tag())
...
```

eseiler: Yeah basically, the array could also be once in the beginning, and then you could even do

```cpp
std::array<char, 2> raw_tag{}; // <- somwhere in the beginning and used every time when parsing a tag

// the lambda
uint16_t get_tag = [&] ()
{
    raw_tag[0] = *it;
    ++it;
    raw_tag[1] = *it;
    ++it;
    return make_tag(raw_tag[0], raw_tag[1]); 
}

// when parsing a tag
switch (get_tag())
```

but let's see what Marcel thinks

include/seqan3/io/sam_file/detail/format_sam_base.hpp
marehr: I think the name is too general as it does more than that. Why should this lambda append additional stuff from the raw_tag if it is named parse_until_tab_or_newline.

I think something like `parse_unknown_tag_value` would be more appropriate, but I'm not too sure when this function is called, so I need your help what this function actually is intended for.

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -423,6 +406,20 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
+    };
+
+    auto read_tag = [](auto it, std::array<char, 2> & raw_tag)
+    {
+        raw_tag[0] = *it;
+        ++it;
+        raw_tag[1] = *it;
+        ++it;
+        return it;
+    };
```
marehr, line 421:
```suggestion
    auto read_tag = [&it](std::array<char, 2> & raw_tag)
    {
        raw_tag[0] = *it;
        ++it;
        raw_tag[1] = *it;
        ++it;
    };
```

okay let's do enricos suggestion

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -433,182 +430,290 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
+        std::cerr << "Unsupported SAM header tag in @" << header_tag << ": " << raw_tag[0] << raw_tag[1] << '\n';
+    };
+
     while (it != end && is_char<'@'>(*it))
     {
-        ++it;                                                           // skip @
+        ++it; // skip @
+
+        std::array<char, 2> raw_at_tag{};
+        it = read_tag(it, raw_at_tag);
```
marehr, line 474:
```suggestion
        read_tag(raw_at_tag);
```

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -433,182 +430,290 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
-                        if (is_char<'O'>(*it))                          // SO (sorting) tag
-                            who = std::addressof(hdr.sorting);
-                        else if (is_char<'S'>(*it))                     // SS (sub-order) tag
-                            who = std::addressof(hdr.subsorting);
-                        else
-                            throw format_error{std::string{"Illegal SAM header tag: S"} + *it};
-                    }
-                    else if (!is_char<'G'>(*it))                        // GO (grouping) tag
+                    std::array<char, 2> raw_tag{};
+                    it = read_tag(it, raw_tag);
```
marehr, line 488:
```suggestion
                    read_tag(raw_tag);
```

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -433,182 +430,290 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
-                    ++it;                                               // skip tab
-                    parse_until_newline(get<1>(info));
-                }
-                ++it;                                                   // skip newline
+                    ++it; // skip tab
 
-                /* If reference information was given, the ids exist and we can fill ref_dict directly.
-                 * If not, wee need to update the ids first and fill the reference dictionary afterwards. */
+                    std::array<char, 2> raw_tag{};
+                    it = read_tag(it, raw_tag);
```
marehr, line 545:
```suggestion
                    read_tag(raw_tag);
```

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -433,182 +430,290 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
-                if (is_char<'\t'>(*it))                                 // read rest of the tags
+                // All tags can appear in any order, SN and LN are required tags
+                while (is_char<'\t'>(*it))
                 {
-                    ++it;
-                    parse_until_newline(get<1>(tmp));
+                    ++it; // skip tab
+
+                    std::array<char, 2> raw_tag{};
+                    it = read_tag(it, raw_tag);
```
marehr, line 616:
```suggestion
                    read_tag(raw_tag);
```

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -433,182 +430,290 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
-                    std::string * who = &tmp.version;
+                    ++it; // skip tab
+                    std::string * program_info_entry{nullptr};
 
-                    switch (*it)
-                    {
-                        case 'P':
-                            ++it;                                       // skip P
+                    std::array<char, 2> raw_tag{};
+                    it = read_tag(it, raw_tag);
```
marehr, line 652:
```suggestion
                    read_tag(raw_tag);
```

marehr:
Why not? Looks fine to me 😁

Am 18. Mai 2021 14:49:57 MESZ schrieb Enrico Seiler ***@***.***>:
***@***.*** commented on this pull request.

test/unit/io/sam_file/format_sam_test.cpp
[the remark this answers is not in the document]
marehr: (because before something threw an exception, now it doesn't anymore)

include/seqan3/io/sam_file/detail/format_sam_base.hpp
marehr: I like that version, but I would still call it `parse_tag` to be in line with the other names.

test/unit/io/sam_file/format_sam_test.cpp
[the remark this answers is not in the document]
marehr: I added a commit that fixes the api stability

include/seqan3/io/sam_file/detail/format_sam_base.hpp
```diff
@@ -433,182 +432,270 @@ inline void format_sam_base::read_header(stream_view_type && stream_view,
...
 
-    auto parse_until_newline = [&] (auto & value)
+    // Some tags are not parsed individually. Instead, these are simply copied into a std::string.
+    // Multiple tags must be separated by a `\t`, hence we prepend a tab to the string, except the first time.
+    // Alternatively, we could always append a `\t`, but this would have the side effect that we might need to trim a
+    // trailing tab after parsing all tags via `pop_back()`.
+    // Unfortunately, we do not know when we are parsing the last tag (and in this case just not append a tab),
+    // because even though we can check if the line ends in a `\n`, it is not guaranteed that the last tag of the
+    // line is passed to this lambda. For example, the line might end with a tag that is properly parsed, such as `ID`.
+    auto parse_until_tab_or_newline = [&] (std::string & value, std::array<char, 2> raw_tag)
```
marehr, line 456:
```suggestion
    auto parse_unknown_tag_value = [&] (std::string & value, std::array<char, 2> raw_tag)
```

?

marehr (changes requested):

include/seqan3/io/sam_file/detail/format_sam_base.hpp
eseiler: Maybe even `append_tag_to_string` since it does not necessarily need to be unknown, we just use this lambda in 2 places where we "dump" all tags that get no special treatment into a string

marehr: okay, then: parse_and_append_unknown_tag_to_string?

Irallia: for me okay

eseiler: Sure, but maybe we should remove the `unknown`, i.e. `parse_and_append_tag_to_string`; `unhandled` would probably be more correct...
for the really unknown tags we emit a warning, and skip them.

marehr: then `parse_and_append_unhandled_tag_to_string`

marehr (approved):
Big thank you! (Sorry for the nitpicky-ness!)

Do you want to rebase or is squash merge okay?

marehr:
(I merged as it seems you cleaned up the history)

PR #2612, opened 2021-05-07
[MISC] Remove template argument from pipeable_config_element.

marehr:
Also inherits privately from the base class such that a reference or pointer to it cannot be formed.

-----
This is one of the commits that survived the rebase of https://github.com/seqan/seqan3/pull/2198.


eseiler (changes requested):
The configs in `search` still do `public` inheritance. Was this intended?

marehr:
> The configs in `search` still do `public` inheritance. Was this intended?

I guess not; I think it was just missed.

eseiler (approved):

PR #2638, opened 2021-05-11
[MISC] move common_tuple to correct place

marehr:
Part of https://github.com/seqan/product_backlog/issues/270

This move happened after the release 3.0.2, so it is safe to just move the header.


Irallia (approved):
LGFM, but I have a question. Why don't you need to deprecate the old file?

marehr:
> LGFM, but I have a question. Why don't you need to deprecate the old file?

Because this change wasn't released yet.

eseiler (approved):

PR #2642, opened 2021-05-12
[MISC] Fix enum order when throwing parser error

eseiler:
Since `enumeration_names` is a map, the order of keys is not defined and may differ across compilers or operating systems.

This makes it hard to, e.g., compare error messages in apps.
Hence, we should sort the keys. I chose to sort them by the order defined in the enum.

### DIFF
--- a/include/seqan3/argument_parser/detail/format_parse.hpp
+++ b/include/seqan3/argument_parser/detail/format_parse.hpp
@@ -323,8 +323,17 @@ private:
 
         if (auto it = map.find(in); it == map.end())
         {
+            std::vector<std::pair<std::string_view, option_t>> key_value_pairs(map.begin(), map.end());
+            std::ranges::sort(key_value_pairs, [] (auto pair1, auto pair2)
+                {
+                    if constexpr (std::totally_ordered<option_t>)
+                        return pair1.second < pair2.second;
+                    else
+                        return pair1.first < pair2.first;
+                });
+
             throw user_input_error{detail::to_string("You have chosen an invalid input value: ", in,
-                                                     ". Please use one of: ", map | std::views::keys)};
+                                                     ". Please use one of: ", key_value_pairs | std::views::keys)};
         }
         else
         {


Irallia (approved):
Thank you! Thats should help me with my gcc11 problem. :)
❤️

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -323,8 +324,10 @@ class format_parse : public format_base
 
         if (auto it = map.find(in); it == map.end())
         {
+            auto keys = map | std::views::keys | seqan3::views::to<std::vector>;
+            std::ranges::sort(keys, [&map] (auto lhs, auto rhs) { return map[lhs] < map[rhs] ;} );
```
marehr, line 328: Why not sort by key? I think we can only sort by value if the value itself is sortable.

I would want to keep the argument_parser free from (external) ranges stuff.

```suggestion
            auto keys_view = map | std::views::keys;
            std::vector<key_type> keys(keys_view.begin(), keys_view.end());
            std::ranges::sort(keys, [&map] (auto key1, auto key2) { return map[key1] < map[key2] ;} );
```

eseiler: Thanks for getting rid of `views::to`, I somehow always forget about the iterator-pair-constructor.

The value must be an enum, so it's always ordered.
I'd prefer sorting by the enum values because they might have some meaningful order regarding the interface (`low`, `average`, `high` might be in that order when sorted by enum, but not when sorted lexicographically).

E.g. in https://github.com/seqan/iGenVar/blob/master/src/variant_detection/method_enums.cpp there are multiple strings that map to one enum, so it would be nice if the string corresponding to an enum are also next to each other.

marehr: Hmm https://docs.seqan.de/seqan/3-master-user/interfaceseqan3_1_1named__enumeration.html just says:

> A instance of seqan3::enumeration_names<option_type> must exist and be of type std::unordered_map<std::string, option_type>. 

It doesn't state any assumptions about the option_type.

The idea was to also allow stuff like:

```c++
struct my_option1{};
struct my_option2{};
struct my_option3{};
using option_type = std::variant<my_option1, my_option2, my_option3>;
```

And these don't have an order.

include/seqan3/argument_parser/detail/format_parse.hpp
```diff
@@ -323,8 +323,11 @@ class format_parse : public format_base
 
         if (auto it = map.find(in); it == map.end())
         {
+            auto keys_view = map | std::views::keys;
```
marehr, line 326:
```suggestion
            auto keys_view = map | std::views::keys | std::views::common;
```

marehr: (That's the one I always forget xD)

marehr: Maybe we should do the second version and just copy the map as is and sort that.

And extract the keys afterwards `". Please use one of: ", keys | std::views::keys))};`

include/seqan3/argument_parser/detail/format_parse.hpp
eseiler: you're right.

What about using value-order if they are ordered, i.e. `std::totally_ordered<option_type>`.
And otherwise by key?

I'd figure that this will mostly be used for enums

marehr: Sound good, we can just change the functor the compares depending on the property :)

include/seqan3/argument_parser/detail/format_parse.hpp
eseiler: If i'd copied the pairs, i would need to fully specify the `std::vector`, maps return `std::pair<const Key, Value>`, but we would need `std::pair<Key, Value>` to be able to sort

include/seqan3/argument_parser/detail/format_parse.hpp
Irallia: I'll try both in iGenVar and give you a feedback.

include/seqan3/argument_parser/detail/format_parse.hpp
eseiler: Actually, gcc7 cannot handle the ctad, so I'll need to define the vector anyways and might as well just copy the pairs

marehr: 👍🏽

marehr (approved):

PR #2644, opened 2021-05-12
[MISC] improve enum ordering when throwing parser error

eseiler:
follow-up of #2642 
In case multiple values are the same, we should use the (always unique) keys as tie-breaker


marehr:
can we add a test that tests that behaviour?

eseiler:
> can we add a test that tests that behaviour?

I did it with try/catch. I couldn't capture the stderr/stdout via internal gtest things.
There is some functionality (`EXPECT_THAT`) for that, but it is in googlemock

marehr (approved):
thank you!

Irallia (approved):
Ah, very good noted!  👍

PR #2651, opened 2021-05-13
Borrowed range

marehr:
https://eel.is/c++draft/ranges.syn defines more views to be `std::ranges::borrowed_range`s.

This PR adds all of them.

Doing this showed that std::views::take isn't a borrowed range in gcc-10, this PR will define that property also for gcc-10.

gcc-10.3 does this anyway.

### DIFF
--- a/include/seqan3/std/ranges
+++ b/include/seqan3/std/ranges
@@ -64,7 +64,7 @@ inline constexpr bool enable_borrowed_range<::std::ranges::iota_view<W, Bound>> 
 template <class T>
 inline constexpr bool enable_borrowed_range<::std::ranges::ref_view<T>> = true;
 
-// Note: in gcc-10 enable_borrowed_range wasn't defined for the take_view and the other following views.
+// Note: in gcc-10 enable_borrowed_range wasn't defined for the take_view and the other following views. (gcc-10.3 fixed this)
 // Explictly defining this here will have the side-effect that we override the definition of
 // std::ranges::enable_borrowed_range for all those views.
 template <class T>
@@ -132,8 +132,9 @@ inline constexpr bool enable_borrowed_range<T> = true;
 #include <seqan3/std/iterator>
 #include <string_view>
 
-// until range <= 0.11.0 release, std::basic_string_view was not recognised as view or borrowed_range
-#if RANGE_V3_VERSION <= 1100
+// After the range release 0.11.0, std::basic_string_view was not recognised as view or borrowed_range any more for
+// gcc <= 9. Remove this if https://github.com/ericniebler/range-v3/pull/1625 was merged before the next release.
+#if RANGE_V3_VERSION >= 1100
 #if defined(__cpp_lib_string_view) && __cpp_lib_string_view == 201603L
 namespace ranges
 {
@@ -146,7 +147,7 @@ template<class CharT, class Traits>
 inline constexpr bool enable_borrowed_range<std::basic_string_view<CharT, Traits>> = true;
 } // namespace ranges
 #endif // defined(__cpp_lib_string_view) && __cpp_lib_string_view >= 201603L
-#endif // RANGE_V3_VERSION <= 1100
+#endif // RANGE_V3_VERSION >= 1100
 
 // ============================================================================
 //  namespace aliasing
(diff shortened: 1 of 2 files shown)


include/seqan3/std/ranges
```diff
@@ -61,6 +63,27 @@ inline constexpr bool enable_borrowed_range<::std::ranges::iota_view<W, Bound>>
 
 template <class T>
 inline constexpr bool enable_borrowed_range<::std::ranges::ref_view<T>> = true;
+
+// note: in gcc-10 enable_borrowed_range wasn't defined for the take_view and the other following views.
```
MitraDarja, line 67:
```suggestion
// Note: in gcc-10 enable_borrowed_range wasn't defined for the take_view and the other following views.
```

MitraDarja (approved):
LGTM, just one minor thing.

include/seqan3/std/ranges
```diff
@@ -107,6 +130,23 @@ inline constexpr bool enable_borrowed_range<T> = true;
...
+template<typename Char, typename Traits>
+    requires true
+inline constexpr bool enable_view<std::basic_string_view<Char, Traits>> = true;
+
+template<class CharT, class Traits>
+    requires true
+inline constexpr bool enable_borrowed_range<std::basic_string_view<CharT, Traits>> = true;
+} // namespace ranges
+#endif // defined(__cpp_lib_string_view) && __cpp_lib_string_view >= 201603L
+#endif // RANGE_V3_VERSION <= 1100
```
eseiler, line 149: As far as I can see, ranges-v3 0.11.0 does this already:

https://github.com/ericniebler/range-v3/blob/0.11.0/include/range/v3/range/concepts.hpp#L186-L190
and 
https://github.com/ericniebler/range-v3/blob/0.11.0/include/range/v3/range/access.hpp#L43-L47

Did I miss something?

Or did you mean to say `RANGE_V3_VERSION < 1100`? But then: we require 0.11.0

eseiler (changes requested):

include/seqan3/std/ranges
marehr: Interesting: the current master does have an error in it:

https://github.com/ericniebler/range-v3/blob/b0696c75dd8b420432c76e5d90fa133ccb423437/include/range/v3/range/access.hpp#L45-L49

That means the version should be `RANGE_V3_VERSION >= 1100`.

I proposed a patch here https://github.com/ericniebler/range-v3/pull/1625 for current master.

eseiler: Now this makes sense :)
Depends on when this is fixed upstream. `>=` would be the most secure bet for now.

marehr: yep

include/seqan3/std/ranges
```diff
@@ -61,6 +63,27 @@ inline constexpr bool enable_borrowed_range<::std::ranges::iota_view<W, Bound>>
 
 template <class T>
 inline constexpr bool enable_borrowed_range<::std::ranges::ref_view<T>> = true;
+
+// Note: in gcc-10 enable_borrowed_range wasn't defined for the take_view and the other following views.
```
marehr, line 67:
```suggestion
// Note: in gcc-10 enable_borrowed_range wasn't defined for the take_view and the other following views. (gcc-10.3 fixed this)
```

include/seqan3/std/ranges
```diff
@@ -107,6 +130,23 @@ inline constexpr bool enable_borrowed_range<T> = true;
 
 #include <seqan3/std/concepts>
 #include <seqan3/std/iterator>
+#include <string_view>
+
+// until range <= 0.11.0 release, std::basic_string_view was not recognised as view or borrowed_range
```
marehr, line 135:
```suggestion
// after the range release 0.11.0 release, std::basic_string_view was not recognised as view or borrowed_range any more in gcc <= 9
```

include/seqan3/std/ranges
```diff
@@ -107,6 +130,23 @@ inline constexpr bool enable_borrowed_range<T> = true;
...
+#if RANGE_V3_VERSION <= 1100
```
marehr, line 136:
```suggestion
#if RANGE_V3_VERSION >= 1100
```

include/seqan3/std/ranges
marehr, line 149:
```suggestion
#endif // RANGE_V3_VERSION >= 1100
```

eseiler (approved):

PR #2658, opened 2021-05-16
remove ref_seq, evalue, bit_score from sam file output

marehr:
sam_file_output does not need the fields ref_seq, evalue, bit_score.

This commit remove those fields from the default-fields and ensures that custom fields don't contain those fields.

This commit also splits ::push_back and ::emplace_back into four versions to deprecate the usage of the old default-fields case.

### DIFF
--- a/include/seqan3/io/sam_file/output.hpp
+++ b/include/seqan3/io/sam_file/output.hpp
@@ -563,8 +563,10 @@ public:
         push_back_tuple(std::forward<tuple_t>(t));
     }
 
-    // This is a bit of a problem, because to decide whether the new syntax or the old syntax is used is a bit
-    // complicated.
+    // The new syntax enforces via static_assert (see above) that field::ref_seq, field::evalue, and field::bit_score
+    // isn't set. That makes sure that someone how explicitly requested for these fields will get a deprecation warning.
+    // That leaves the case where the user did not provide any fields and just used the default ones. 
+    // This causes a problem, because it is complicated to decide whether the new syntax or the old syntax is used.
     // position | old fields         | new fields
     // -------- | ------------------ | -------------
     //  0       | field::seq         | field::seq
--- a/include/seqan3/io/sam_file/output.hpp
+++ b/include/seqan3/io/sam_file/output.hpp
@@ -145,7 +145,6 @@ template <detail::fields_specialisation selected_field_ids_ =
               fields<field::seq,
                      field::id,
                      field::offset,
-                     field::ref_seq,
                      field::ref_id,
                      field::ref_offset,
                      field::alignment,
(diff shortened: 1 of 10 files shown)


marehr:
TODO: Changelog

MitraDarja (approved):
LGTM, one minor thing to improve doc. :)

include/seqan3/io/sam_file/output.hpp
```diff
@@ -534,11 +552,114 @@ class sam_file_output
...
+    //!\cond
+        // new syntax enforces via static_assert that field::ref_seq, field::evalue, and field::bit_score isn't set
+        requires tuple_like<tuple_t> && (!detail::record_like<tuple_t>) && (!is_default_selected_field_ids)
+    //!\endcond
+    {
+        push_back_tuple(std::forward<tuple_t>(t));
+    }
+
+    // This is a bit of a problem, because to decide whether the new syntax or the old syntax is used is a bit
+    // complicated.
```
marehr, line 567:
```suggestion
    // The new syntax enforces via static_assert (see above) that field::ref_seq, field::evalue, and field::bit_score
    // isn't set. That makes sure that someone how explicitly requested for these fields will get a deprecation warning.
    // That leaves the case where the user did not provide any fields and just used the default ones. 
    // This causes a problem, because it is complicated to decide whether the new syntax or the old syntax is used.
```

include/seqan3/io/sam_file/output.hpp
```diff
@@ -534,11 +552,116 @@ class sam_file_output
...
+    //!\cond
+        // new syntax enforces via static_assert that field::ref_seq, field::evalue, and field::bit_score isn't set
+        requires tuple_like<tuple_t> && (!detail::record_like<tuple_t>) && (!is_default_selected_field_ids)
+    //!\endcond
+    {
+        push_back_tuple(std::forward<tuple_t>(t));
+    }
+
+    // The new syntax enforces via static_assert (see above) that field::ref_seq, field::evalue, and field::bit_score
+    // isn't set. That makes sure that someone how explicitly requested for these fields will get a deprecation warning.
```
marehr, line 567:
```suggestion
    // isn't set. That makes sure that someone who explicitly requested for these fields will get a deprecation warning.
```

include/seqan3/io/sam_file/output.hpp
```diff
@@ -534,11 +540,114 @@ class sam_file_output
...
+    void push_back(tuple_t && t)
+    //!\cond
+        // new syntax enforces via static_assert that field::ref_seq, field::evalue, and field::bit_score isn't set
+        requires tuple_like<tuple_t> && (!detail::record_like<tuple_t>) && (!is_default_selected_field_ids)
+    //!\endcond
+    {
+        push_back_tuple(std::forward<tuple_t>(t));
+    }
+
+    // This is a bit of a problem, because to decide whether the new syntax or the old syntax is used is a bit
```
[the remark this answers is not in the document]
marehr: Thank you!

eseiler (approved):

eseiler:
squash or rebase?

Issue #2620, opened 2021-05-09
/usr/include/c++/11/array:196:9: error: call to non-'constexpr' function 'void std::__failed_assertion()'


sagitter:
Hi all.

## Platform

  - SeqAn version: 52f0b79cfae340e56321ae85fce8fec5b73d8f86
  - Operating system: Fedora 35
  - Compiler: GCC-11.1.1

## Description
```
[ 87%] Building CXX object utility/container/CMakeFiles/small_vector_test.dir/small_vector_test.cpp.o
cd /builddir/build/BUILD/seqan3/build/utility/container && /usr/bin/ccache /usr/lib64/ccache/g++ -DSEQAN3_HAS_BZIP2=1 -DSEQAN3_HAS_ZLIB=1 -DSEQAN3_WITH_CEREAL=1 -DSEQAN3_WITH_LEMON=1 -I/usr/include/gtest -I/builddir/build/BUILD/seqan3/test/include -I/builddir/build/BUILD/seqan3/include -isystem /builddir/build/BUILD/seqan3/submodules/sdsl-lite/include -O2  -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1  -fcommon -m64  -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -std=c++2a -fconcepts -lstdc++fs -DNDEBUG -pedantic -Wall -Wextra -Werror -pthread -MD -MT utility/container/CMakeFiles/small_vector_test.dir/small_vector_test.cpp.o -MF CMakeFiles/small_vector_test.dir/small_vector_test.cpp.o.d -o CMakeFiles/small_vector_test.dir/small_vector_test.cpp.o -c /builddir/build/BUILD/seqan3/test/unit/utility/container/small_vector_test.cpp
In file included from /usr/include/c++/11/cstddef:49,
                 from /usr/include/gtest/gtest.h:55,
                 from /builddir/build/BUILD/seqan3/test/unit/utility/container/small_vector_test.cpp:8:
/usr/include/c++/11/array: In member function 'virtual void small_vector_comparison_Test::TestBody()':
[...]

PR #2662, opened 2021-05-16
[INFRA] add -DCMAKE_BUILD_TYPE=FEDORA to simulate fedora builds

marehr:
Added flags of https://github.com/seqan/seqan3/issues/2620#issuecomment-838482217

### DIFF
--- a/test/seqan3-test.cmake
+++ b/test/seqan3-test.cmake
@@ -27,6 +27,20 @@ option (SEQAN3_TEST_BUILD_OFFLINE "Skip the update step of external projects." O
 option (SEQAN3_BENCHMARK_ALIGN_LOOPS "Pass -falign-loops=32 to the benchmark builds." ON)
 
 # ----------------------------------------------------------------------------
+# Custom Build types
+# ----------------------------------------------------------------------------
+
+# -DCMAKE_BUILD_TYPE=FEDORA; our library did not compile for fedora quite a few times, because of that we created this
+# custom build type to emulate their flag set-up.
+# We omitted:
+#   -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1
+#   -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1
+#   set -mtune=native
+#   and -fcf-protection=check
+# See https://src.fedoraproject.org/rpms/redhat-rpm-config/blob/rawhide/f/buildflags.md for an overview
+set (CMAKE_CXX_FLAGS_FEDORA "-O2 -flto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fstack-protector-strong -m64 -mtune=native -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection=check")
+
+# ----------------------------------------------------------------------------
 # Paths to folders.
 # ----------------------------------------------------------------------------
 


SGSSGene (approved):

test/seqan3-test.cmake
```diff
@@ -26,6 +26,20 @@ option (SEQAN3_TEST_BUILD_OFFLINE "Skip the update step of external projects." O
...
+# Custom Build types
+# ----------------------------------------------------------------------------
+
+# -DCMAKE_BUILD_TYPE=FEDORA; our library did not compile for fedora quite a few times, because of that we created this
+# custom build type to emulate their flag set-up.
+# We omitted:
+#   -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1
+#   -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1
+#   set -mtune=native
+#   and -fcf-protection=check
```
eseiler, line 39: We could actually do a compile check for this one if `-fcf-protection=full` does not work, we use `=check`. 
But I guess this flag doesnt add much

test/seqan3-test.cmake
```diff
@@ -26,6 +26,20 @@ option (SEQAN3_TEST_BUILD_OFFLINE "Skip the update step of external projects." O
...
+# See https://src.fedoraproject.org/rpms/redhat-rpm-config/blob/rawhide/f/buildflags.md for an overview
+set (CMAKE_CXX_FLAGS_FEDORA "-O2 -flto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fstack-protector-strong -m64 -mtune=native -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection=check")
```
eseiler, line 41: Also, to what do the other flags default, nothing? They should probably be debug?
You could "inherit" fromd debug, e.g. `set (CMAKE_C_FLAGS_FEDORA "${CMAKE_C_FLAGS_DEBUG")`, also for the linker flags.

On the nightlies, I use Debug and just add the cxx flags

marehr: Sorry, I don't quite understand that comment. I think we don't need the C flags, because this should be a C++ project.

> On the nightlies, I use Debug and just add the cxx flags

And we could change that to `-DCMAKE_BUILD_TYPE=FEDORA`

test/seqan3-test.cmake
marehr: Yeah true, I think most of the flags are unneeded. I think we should have for a first version have only flags work on multiple platforms. Those could be refined later, if they show hidden errors.

test/seqan3-test.cmake
eseiler: It might be a difference if you build in debug and add some cxxflags; or change the build type. E.g. C flags would be the debug flags, etc.; linker flags would also be debug flags.
Now there are probably no flags for linker or C (gtest uses C, no?); because we do not set them.

marehr: ah; I thought of fedora like a specialised RELEASE build type

eseiler: That's ok, but I'm not sure what with the other flags happens. 
We can either leave it as is, or define the C and linker flags, even if we just alias the release flags.

marehr: the other flags are empty. you don't need to specialise them.

eseiler (approved):

PR #2667, opened 2021-05-18
[INFRA] Ignore apt failures

eseiler:
This should be cherry-picked into master afterwards

:( 

```
Reading package lists...
E: Failed to fetch https://packages.microsoft.com/ubuntu/20.04/prod/dists/focal/main/binary-amd64/Packages.bz2  File has unexpected size (73402 != 73160). Mirror sync in progress? [IP: 40.74.238.15 443]
   Hashes of expected file:
    - Filesize:73160 [weak]
    - SHA512:782bf0315c8de4314049e599009430053b75a5015491b3d383b09239fcf31857d52bac3399cef8e813f7b2131dce09b03eaf89d79c5ebd2849402cee7aa9d228
    - SHA256:dbf4bf81b4a0a17dae53f2aff3edd4cf5bb1dcd5dcd216ce5c5c8ab7c507a26a
    - SHA1:0e1b47276cae4168a08d8fd981891d4773f053b9 [weak]
    - MD5Sum:0fcf3ed335cd4efde65300bb48c05afe [weak]
   Release file created at: Sat, 15 May 2021 00:38:08 +0000
E: Some index files failed to download. They have been ignored, or old ones used instead.
Error: Process completed with exit code 100.
```

### DIFF
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -41,6 +41,7 @@ jobs:
           path: seqan3/submodules/seqan
 
       - name: Configure APT
+        continue-on-error: true
         run: bash ./seqan3/.github/workflows/scripts/configure_apt.sh
 
       - name: Install CMake
--- a/.github/workflows/ci_linux.yml
+++ b/.github/workflows/ci_linux.yml
@@ -92,6 +92,7 @@ jobs:
           path: seqan3/submodules/seqan
 
       - name: Configure APT
+        continue-on-error: true
         run: bash ./seqan3/.github/workflows/scripts/configure_apt.sh
 
       - name: Install CMake
--- a/.github/workflows/ci_misc.yml
+++ b/.github/workflows/ci_misc.yml
@@ -124,6 +124,7 @@ jobs:
           path: seqan3/submodules/seqan
 
       - name: Configure APT
+        continue-on-error: true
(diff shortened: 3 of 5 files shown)


marehr:
No re-try flag for that issue?

marehr (approved):

eseiler:
> No re-try flag for that issue?

I don't think there is, if it's a download issue, we already retry. But this filesize issue is usually not sporadic but for a few hours, so it's best to just move on - retrying would not help.
I don't think we use anything from the package and we could also just remove it from the repository list; but I don't want to fiddle with the repositories more than I need to.

MitraDarja (approved):

.github/workflows/api.yml
```diff
@@ -41,6 +41,7 @@ jobs:
           path: seqan3/submodules/seqan
 
       - name: Configure APT
+        continue-on-error: true
```
marehr, line 44: Maybe put it into the script itself? Otherwise, I like that you can see here that the apt failed.

eseiler: There is either fail or no fail in a step, I cannot skip on failure. So it doesn't matter where I put it. This way, it's at least not hidden inside the script that we do not fail on error

PR #2668, opened 2021-05-18
[FIX] brew script

eseiler:


marehr (approved):

marehr:
Not sure why but `SeqAn3 CI on macOS / Unit gcc10 (pull_request)` seems to be hung up?

image

SGSSGene (approved):

PR #2671, opened 2021-05-19
-DSEQAN3_REMOVE_DEPRECATED_310

marehr:
This adds an option to simulate the removal of deprecated API. Sometimes some entities won't trigger a good deprecation notice and the only option is to see the build fail.

### DIFF
--- a/test/snippet/io/record_2.cpp
+++ b/test/snippet/io/record_2.cpp
@@ -9,7 +9,6 @@
 int main()
 {
     using namespace seqan3::literals;
-    using seqan3::get;
 
     // The order of the types below represent a mapping between the type and the key.
     using types        = seqan3::type_list<seqan3::dna4_vector, std::string, std::vector<seqan3::phred42>>;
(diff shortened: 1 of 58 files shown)


test/snippet/io/record_2.cpp
```diff
@@ -20,9 +19,13 @@ int main()
 
     record_type my_record{};
     get<1>(my_record) = "the most important sequence in the database";            // access via index
+    get<std::string>(my_record) = "the least important sequence in the database"; // access via type
```
eseiler, line 22:
```suggestion
    std::get<1>(my_record) = "the most important sequence in the database";            // access via index
    std::get<std::string>(my_record) = "the least important sequence in the database"; // access via type
```

Irallia (approved):
This `-DSEQAN3_REMOVE_DEPRECATED_310` looks like a really good check 👍 
LGFM, thank you!

eseiler (approved):
Will the cron be a follow-up PR? 
Do you want to rebase?

marehr:
CI in next PR.

PR #2676, opened 2021-05-21
[FIX] some deprecations

eseiler:
Some guards were missing in the benchmarks for `-DSEQAN3_REMOVE_DEPRECATED_310` to work

### DIFF
--- a/test/performance/std/bit_benchmark.cpp
+++ b/test/performance/std/bit_benchmark.cpp
@@ -48,6 +48,7 @@ static void is_power_of_two_arithmetic(benchmark::State & state) {
 }
 BENCHMARK(is_power_of_two_arithmetic);
 
+#ifdef SEQAN3_DEPRECATED_310
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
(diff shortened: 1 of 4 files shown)


marehr (approved):

test/performance/std/bit_benchmark.cpp
```diff
@@ -48,6 +48,7 @@ static void is_power_of_two_arithmetic(benchmark::State & state) {
 }
 BENCHMARK(is_power_of_two_arithmetic);
 
+#ifdef SEQAN3_DEPRECATED_310
```
SGSSGene, line 51: If I see this correctly, all benchmark that are seqan3 related are being removed when SEQAN3_DEPRECATED_310 is not defined. In that case we are only checking std function. May the complete file should be surrounded by a guard, so we remove it completely when we drop deprecated functionalities?

SGSSGene (changes requested):

test/performance/std/bit_benchmark.cpp
eseiler: There are still tests for intrinsics in this benchmark (`__builtin_popcountll`).

SGSSGene (approved):

PR #2686, opened 2021-05-28
Move implementation of seqan3::detail::take_exactly_view 

marehr:


test/unit/utility/views/slice_test.cpp
```diff
@@ -16,6 +16,7 @@
 
 #include <seqan3/test/expect_range_eq.hpp>
 #include <seqan3/test/expect_same_type.hpp>
+#include <seqan3/utility/range/concept.hpp>
```
Irallia, line 19: I  assume that this inlcude was previously given from somewhere else.

Irallia (approved):
And now the removing part starts, right?
LGFM.

test/unit/utility/views/slice_test.cpp
eseiler: Yup, `slice.hpp` included `take_view.hpp` which included `concept.hpp`. Now that `take_view.hpp`, we have to add it here

eseiler (approved):

PR #2687, opened 2021-05-28
[FIX] skip tests that do not work with root rights

eseiler:
Follow-up should be a cron that runs ctest with sudo :)


SGSSGene (approved):

test/unit/test/CMakeLists.txt
```diff
@@ -1,8 +1,9 @@
 seqan3_test(expect_range_eq_test.cpp)
 seqan3_test(expect_same_type_test.cpp)
+seqan3_test(permissions.cpp)
```
marehr, line 3:
```suggestion
seqan3_test(file_access_test.cpp)
```

?

test/unit/test/permissions.cpp
```diff
@@ -0,0 +1,100 @@
...
+// Copyright (c) 2016-2021, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+#include <gtest/gtest.h>
+
+#include <seqan3/test/file_access.hpp>
+#include <seqan3/test/tmp_directory.hpp>
+#include <seqan3/test/tmp_filename.hpp>
```
marehr, line 12: This test is cyclic;

* seqan3/test/file_access.hpp
* tested by file_access_test.cpp
* uses seqan3/test/tmp_directory.hpp
* tested by tmp_directory_test.cpp
* uses seqan3/test/file_access.hpp

Can we write this test in a way that does not need seqan3/test/tmp_directory.hpp and seqan3/test/tmp_filename.hpp?

marehr (changes requested):

test/unit/test/permissions.cpp
eseiler: I could, e.g., use the current working directory and skip the tests if I do not have permissions to write there

marehr: Sounds like a great idea.

marehr (approved):
noice! ty!

PR #2693, opened 2021-05-30
[FIX] ExternalProject_Add using wrong gbenchmark tag

eseiler:
Now that gbenchmark renamed `master` to `main` and deleted `master`, we get errors since the default branch is `master`.


marehr (approved):

SGSSGene (approved):
:+1:

PR #2694, opened 2021-05-30
[INFRA] workflow for RAM-Usage

eseiler:
E.g., https://github.com/eseiler/seqan3/actions/runs/890412988


marehr:
Can you add an example `.tsv` file? Furthermore, I have the feeling that we should start to document what all the different CI's are doing and how to use / trigger them.

Maybe a README.md in `.github/workflows`?

.github/workflows/ram_usage.yml
```diff
@@ -0,0 +1,94 @@
...
+
+      - name: Install CMake
+        run: bash ./seqan3/.github/workflows/scripts/install_cmake.sh
+
+      - name: Install compiler g++-${{ github.event.inputs.compiler }}
+        run: sudo apt-get install --yes g++-${{ github.event.inputs.compiler }}
+
+      - name: Create gcc wrapper
+        run: |
+          echo -e '#!/bin/bash\n{ /usr/bin/time -v /usr/bin/g++-${{ github.event.inputs.compiler }} "$@"; }' > g++.sh
```
marehr, line 54: Just a stupid question, would `-DCMAKE_CXX_COMPILER="/usr/bin/time -v /usr/bin/g++-${{ github.event.inputs.compiler }}"` work, or `-DCMAKE_CXX_COMPILER=g++-${{ github.event.inputs.compiler }} -DCMAKE_CXX_COMPILER_LAUNCHER="/usr/bin/time -v"` ?

.github/workflows/ram_usage.yml
```diff
@@ -0,0 +1,94 @@
...
+        env:
+          CXX: ${{ github.workspace }}/g++.sh
+          CC: gcc-${{ github.event.inputs.compiler }}
+        run: |
+          mkdir seqan3-build
+          cd seqan3-build
+          cmake ../seqan3/test/unit -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS="${{ github.event.inputs.cxx_flags }}" -DSEQAN3_VERBOSE_TESTS=OFF
+
+      - name: Update gcc wrapper
+        run: echo -e '#!/bin/bash\nFILE=$(mktemp ram_usage.XXXXXXXX)\n{ /usr/bin/time -v /usr/bin/g++-${{ github.event.inputs.compiler }} "$@" 2> $FILE; }' > g++.sh
```
marehr, line 67: Huh? Wouldn't the configuration not work otherwise?

eseiler: Will fail to configure the `{i,o}stream_test` in `contrib`, I haven't looked into that.

.github/workflows/ram_usage.yml
eseiler: Maybe, but won't work with the redirect I need further down

eseiler:
> Can you add an example `.tsv` file? Furthermore, I have the feeling that we should start to document what all the different CI's are doing and how to use / trigger them.
> 
> Maybe a README.md in `.github/workflows`?

I wouldn't add an example `csv`, because it is just `test_name, RAM in MiB`, so not very exciting, but I can add a few lines of the output csv to a README.

But I think the README should be follow-up.

.github/workflows/ram_usage.yml
eseiler: I needed to handle `-v` for zlib/bzip2 detection

marehr (approved):

.github/workflows/scripts/ram_usage.py
```diff
@@ -0,0 +1,40 @@
...
+parser.add_argument('output', type=str, help='File to write output to.')
+arguments = parser.parse_args()
+
+file_names = []
+ram_usages = []
+
+with open(arguments.input, 'r') as input_file:
+    parsing_ram_usage = False
+    for line_number, line in enumerate(input_file):
+        if line_number % 23 == 0:
```
MitraDarja, line 27: Can you explain me what this if and the one below (if parsing_ram_usage and ((line_number - 9) % 23) == 0:) are needed for?

MitraDarja (approved):
LGTM, but I have a question. :)

.github/workflows/scripts/ram_usage.py
eseiler: You have something like this

```
        Command being timed: "/usr/bin/g++-11 [...] -c /home/.../seqan3/test/unit/argument_parser/format_parse_test.cpp"
        User time (seconds): 9.21
        System time (seconds): 0.39
        Percent of CPU this job got: 99%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:09.64
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 672856
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 0
        Minor (reclaiming a frame) page faults: 211593
        Voluntary context switches: 9
        Involuntary context switches: 13
        Swaps: 0
        File system inputs: 8
        File system outputs: 44712
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
        Command being timed: "/usr/bin/g++-11 CMakeFiles/format_parse_test.dir/format_parse_test.cpp.o -o format_parse_test [...]"
        User time (seconds): 0.22
        System time (seconds): 0.04
        Percent of CPU this job got: 76%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.34
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 54872
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 0
        Minor (reclaiming a frame) page faults: 15478
        Voluntary context switches: 725
        Involuntary context switches: 0
        Swaps: 0
        File system inputs: 0
        File system outputs: 0
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
```

Every 23 lines there is a new "Command being timed:".
There's the compile step I'm actually interested in which contains the path to the source file (e.g, `seqan3/test/unit/argument_parser/format_parse_test.cpp`.
And then there is the linking step (e.g., `CMakeFiles/format_parse_test.dir/format_parse_test.cpp.o -o format_parse_test`) and it does not contain the path to the source file.

Hence, I can iterate over all the time outputs (blocks of length 23) and check if there is `unit` in the command. If yes, this is something I want the RAM usage of (`parsing_ram_usage`). If I want to parse the RAM usage, and I'm on the 9th line after the beginning of a block, I'm in the line that contains the `Maximum resident set size`.

PR #2695, opened 2021-06-01
[DOC] Add a cookbook entry for an enum validator

Irallia:
After creating my own enum validator in iGenVar, has @eseiler suggested to add this as a cookbook entry into the seqan3 documentation.
I have collected all the necessary changes for this pr and put them together in a single file.


smehringer:
@Irallia Can this be moved to sharg? Also (I quickly skimmed the code), coudn't the same be achieved with a `value_list_validator`?

Irallia:
We would like to move this to the sharg parser: https://github.com/seqan/sharg-parser/issues/87

PR #2734, opened 2021-06-07
[FIX] Removing executable bit from files that can't be executed

SGSSGene:
The executable bit should not be set on files that are not being
executed. Some source files and the README had this bit accidentally set.
This commit removes the bit from these files.

There is no issue associated with this PR.


MitraDarja (approved):

marehr:
Can you rebase on `upstream/release-3.1.0`?

eseiler (approved):

PR #2736, opened 2021-06-07
[FIX] Adjust benchmark for view_take_until

SGSSGene:
This patch addresses the performance regression between 3.0.2 and 3.0.3.

The view_take_until benchmark was refactored after the 3.0.2 release.
The benchmark code was accidentally changed in its semantics, causing
a false indication of increased run time. This patch restores the
benchmark to its original test.

### DIFF
--- a/test/performance/io/detail/view_take_until_benchmark.cpp
+++ b/test/performance/io/detail/view_take_until_benchmark.cpp
@@ -46,7 +46,8 @@ void sequential_read(benchmark::State & state)
         {
             single_pass_or_ref_t single_pass_or_ref{container};
             for (auto elem : single_pass_or_ref)
-                sum += elem;
+                if(sum += elem; elem >= 101)
+                    break;
         }
     }
     else // {seqan3,std}::views::take* adaptor


test/performance/io/detail/view_take_until_benchmark.cpp
```diff
@@ -46,7 +46,13 @@ void sequential_read(benchmark::State & state)
         {
             single_pass_or_ref_t single_pass_or_ref{container};
             for (auto elem : single_pass_or_ref)
+            {
                 sum += elem;
+                if (elem >= 100)
+                {
+                    break;
+                }
+            }
```
eseiler, line 55:
```suggestion
            for (auto elem : single_pass_or_ref)           
                if (sum += elem ; elem >= 101)
                    break;
```

Less code :)

I think it should be `101` (from 3.0.2):

https://github.com/seqan/seqan3/blob/370673ffd20d790f759aed15771072436a422ec3/test/performance/range/views/view_take_until_benchmark.cpp#L42-L44

SGSSGene: Oh yeah, it should be 101....I can't even copy&paste xD

marehr (approved):

eseiler (approved):

PR #2737, opened 2021-06-07
[TEST] Release snippets

marehr:
The cmake change will allow dots (i.e. `.`) in test names, e.g. "foo3.3_test.cpp" or "3.0.1_some-message.cpp". Before this change the test names would have been `foo3` and `3` instead of `foo3.3` and `3.0.1_some-message`.

The main part of the PR adds (updated) snippets of our release notes.

### DIFF
--- a/test/cmake/seqan3_path_longest_stem.cmake
+++ b/test/cmake/seqan3_path_longest_stem.cmake
@@ -1,0 +1,44 @@
+# -----------------------------------------------------------------------------------------------------
+# Copyright (c) 2006-2021, Knut Reinert & Freie Universität Berlin
+# Copyright (c) 2016-2021, Knut Reinert & MPI für molekulare Genetik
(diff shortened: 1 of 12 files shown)


SGSSGene (approved):
LGTM! :+1: 

Could you explain (or point to documentation) of the exact meaning of `//![main]`?
Only parts between these tags will be presented in the final output?

eseiler:
> LGTM! 👍
> 
> Could you explain (or point to documentation) of the exact meaning of `//![main]`?
> Only parts between these tags will be presented in the final output?

It's just a doxygen snippet (https://www.doxygen.nl/manual/commands.html#cmdsnippet) though we are not using them right now. But if you were to include the snippets, only the stuff between the two `//![main]`  would show up

eseiler (approved):

PR #2740, opened 2021-06-08
[FIX] properly guard ZLIB and BZIP2 

marehr:
In https://github.com/seqan/seqan3/issues/2726#issuecomment-856603031 we noticed that the current header tests can't be build via

```
cmake -DSEQAN3_NO_ZLIB=1 -DSEQAN3_NO_BZIP2=1 <seqan3>/test/header
```

This PR fixes that.

### DIFF
--- a/doc/cookbook/compression_threads.cpp
+++ b/doc/cookbook/compression_threads.cpp
@@ -1,4 +1,4 @@
-#if SEQAN3_HAS_ZLIB
+#if defined(SEQAN3_HAS_ZLIB)
 //![example]
 #include <seqan3/io/all.hpp>
 
@@ -15,4 +15,4 @@ int main()
     return 0;
 }
 //![example]
-#endif // SEQAN3_HAS_ZLIB
+#endif // defined(SEQAN3_HAS_ZLIB)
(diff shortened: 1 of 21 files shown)


eseiler:
What's the effect of `#if SEQAN3_HAS_ZLIB` -> `#if defined(SEQAN3_HAS_ZLIB)`?

`#ifdef SEQAN3_HAS_ZLIB` and `#if defined(SEQAN3_HAS_ZLIB)` are the same, we use both versions in the lib, so why was this changed? Just style?

marehr:
> What's the effect of `#if SEQAN3_HAS_ZLIB` -> `#if defined(SEQAN3_HAS_ZLIB)`?
> 
> `#ifdef SEQAN3_HAS_ZLIB` and `#if defined(SEQAN3_HAS_ZLIB)` are the same, we use both versions in the lib, so why was this changed? Just style?

No particular reason, it was just easier to normalise `#if SEQAN3_HAS_ZLIB` and `#ifdef SEQAN3_HAS_ZLIB` to `#if defined(SEQAN3_HAS_ZLIB)` by a simple search and replace.

simonsasse (approved):
looks good. :)

eseiler (approved):

eseiler:
rebase?

marehr:
> rebase?

I didn't as the first commit is the exact same commit.

eseiler:
I just noticed now that this had master as base, so this needs to be cherry-picked into release?

marehr:
dammit

Issue #2747, opened 2021-06-09
[CRON] Latest Libraries


eseiler:

eseiler:
[Resolved upstream](https://github.com/xxsds/sdsl-lite/commit/79199ff15ffac4ae9e5301939d8eddc7378fe0a6)

eseiler:
Fixed upstream: https://github.com/seqan/seqan/pull/2478

eseiler:
Note: bogus warnings with gcc12 in googletest won't be workaround'ed upstream. They should be fixed with gcc12.3

eseiler:
Fixed upstream

seqan-actions:
Failure header on gcc-11
---

See https://github.com/seqan/seqan3/actions/runs/11539362409 for more information.

seqan-actions:
Failure snippet on gcc-12
---

See https://github.com/seqan/seqan3/actions/runs/11539362409 for more information.

seqan-actions:
Failure header on gcc-13
---

See https://github.com/seqan/seqan3/actions/runs/11539362409 for more information.

seqan-actions:
Failure snippet on gcc-11
---

See https://github.com/seqan/seqan3/actions/runs/11539362409 for more information.

seqan-actions:
Failure performance on gcc-11
---

See https://github.com/seqan/seqan3/actions/runs/11539362409 for more information.

seqan-actions:
Failure snippet on gcc-13
---

See https://github.com/seqan/seqan3/actions/runs/11539362409 for more information.

seqan-actions:
Failure performance on gcc-13
---

See https://github.com/seqan/seqan3/actions/runs/11539362409 for more information.

seqan-actions:
Failure unit on clang-17
---

See https://github.com/seqan/seqan3/actions/runs/11539362409 for more information.

seqan-actions:
Failure performance on clang-18
---

See https://github.com/seqan/seqan3/actions/runs/11539362409 for more information.

seqan-actions:
Failure performance on clang-17
---

See https://github.com/seqan/seqan3/actions/runs/11539362409 for more information.

seqan-actions:
Failure snippet on clang-18
---

See https://github.com/seqan/seqan3/actions/runs/11539362409 for more information.

seqan-actions:
Failure unit on gcc-11
---

[...]

Issue #2746, opened 2021-06-09
[CRON] API-Stability


eseiler:

eseiler:
Fixed by https://github.com/seqan/seqan3/pull/3019

PR #2748, opened 2021-06-09
[INFRA] Use one issue for each CRON

eseiler:
This will post comments into an issue and re-open the issue if there should be a failure.

API: https://github.com/seqan/seqan3/issues/2746
LL: https://github.com/seqan/seqan3/issues/2747

Example:
API: https://github.com/eseiler/seqan3/issues/18
LL: https://github.com/eseiler/seqan3/issues/17

In ended up rendering the template with sed by myself because there wasn't an action that worked well for that.


MitraDarja (approved):

marehr (approved):
How does this behave in my local clone?

eseiler:
> How does this behave in my local clone?

Probably not so well

A follow-up could be to query the repo for a specific issue by name.

PR #2753, opened 2021-06-15
[MISC] Use tmp_directory instead of tmp_filename

SGSSGene:
Moves forward https://github.com/seqan/product_backlog/issues/281

This replaces occurrences of `tmp_filename` and replaces them with `tmp_directory`.

This PR doesn't remove `tmp_filename`, because it is unclear how the *.patch files must be updated to not break.

### DIFF
--- a/test/include/seqan3/test/cereal.hpp
+++ b/test/include/seqan3/test/cereal.hpp
@@ -18,7 +18,7 @@
 
 #include <seqan3/core/concept/cereal.hpp>
 #include <seqan3/core/platform.hpp>
-#include <seqan3/test/tmp_filename.hpp>
+#include <seqan3/test/tmp_directory.hpp>
 #include <seqan3/utility/type_traits/basic.hpp>
 
 #if SEQAN3_WITH_CEREAL
@@ -47,17 +47,18 @@ template <cereal_input_archive in_archive_t, cereal_output_archive out_archive_t
 //!\endcond
 void do_cerealisation(value_t && value)
 {
-    tmp_filename filename{"cereal_test"};
+    tmp_directory tmp{};
+    auto filename = tmp.path() / "cereal_test";
 
     {
-        std::ofstream os{filename.get_path(), std::ios::binary};
+        std::ofstream os{filename, std::ios::binary};
         out_archive_t oarchive{os};
         oarchive(value);
     }
 
     {
         std::remove_cvref_t<value_t> value_from_archive{};
-        std::ifstream is{filename.get_path(), std::ios::binary};
+        std::ifstream is{filename, std::ios::binary};
         in_archive_t iarchive{is};
         iarchive(value_from_archive);
         EXPECT_TRUE(value == value_from_archive);
--- a/test/snippet/core/cereal_example.cpp
+++ b/test/snippet/core/cereal_example.cpp
@@ -6,29 +6,30 @@
 #include <cereal/types/vector.hpp>    // includes cerealisation support for std::vector
 
 #include <seqan3/core/debug_stream.hpp>
-#include <seqan3/test/tmp_filename.hpp>
+#include <seqan3/test/tmp_directory.hpp>
 
 // Written for std::vector, other types also work.
-void load(std::vector<int16_t> & data, seqan3::test::tmp_filename & tmp_file)
+void load(std::vector<int16_t> & data, std::filesystem::path const & tmp_file)
 {
-    std::ifstream is(tmp_file.get_path(), std::ios::binary); // Where input can be found.
-    cereal::BinaryInputArchive archive(is);                  // Create an input archive from the input stream.
-    archive(data);                                           // Load data.
+    std::ifstream is(tmp_file, std::ios::binary); // Where input can be found.
+    cereal::BinaryInputArchive archive(is);       // Create an input archive from the input stream.
+    archive(data);                                // Load data.
 }
 
 // Written for std::vector, other types also work.
-void store(std::vector<int16_t> const & data, seqan3::test::tmp_filename & tmp_file)
+void store(std::vector<int16_t> const & data, std::filesystem::path const & tmp_file)
 {
-    std::ofstream os(tmp_file.get_path(), std::ios::binary); // Where output should be stored.
-    cereal::BinaryOutputArchive archive(os);                 // Create an ouput archive from the output stream.
-    archive(data);                                           // Store data.
+    std::ofstream os(tmp_file, std::ios::binary); // Where output should be stored.
+    cereal::BinaryOutputArchive archive(os);      // Create an output archive from the output stream.
+    archive(data);                                // Store data.
 }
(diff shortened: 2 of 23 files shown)


test/snippet/core/cereal_example.cpp
```diff
@@ -6,29 +6,30 @@
...
 
 // Written for std::vector, other types also work.
-void store(std::vector<int16_t> const & data, seqan3::test::tmp_filename & tmp_file)
+void store(std::vector<int16_t> const & data, std::filesystem::path const& tmp_file)
 {
-    std::ofstream os(tmp_file.get_path(), std::ios::binary); // Where output should be stored.
-    cereal::BinaryOutputArchive archive(os);                 // Create an ouput archive from the output stream.
-    archive(data);                                           // Store data.
+    std::ofstream os(tmp_file, std::ios::binary); // Where output should be stored.
+    cereal::BinaryOutputArchive archive(os);      // Create an ouput archive from the output stream.
```
eaasna, line 23:
```suggestion
    cereal::BinaryOutputArchive archive(os);      // Create an output archive from the output stream.
```

test/unit/io/sam_file/sam_file_input_test.cpp
```diff
@@ -406,17 +421,19 @@ std::string input_bgzf
 
 TEST_F(sam_file_input_f, decompression_by_filename_bgzf)
 {
-    seqan3::test::tmp_filename filename{"sam_file_output_test.sam.bgzf"};
+    seqan3::test::tmp_directory tmp{};
+    auto filename = tmp.path() / "sam_file_onput_constructor.sam.bgzf";
```
eaasna, line 425: I wonder what an "onput_constructor" is

test/unit/io/sam_file/sam_file_input_test.cpp
```diff
@@ -459,17 +476,19 @@ std::string input_bz2
 
 TEST_F(sam_file_input_f, decompression_by_filename_bz2)
 {
-    seqan3::test::tmp_filename filename{"sam_file_output_test.sam.bz2"};
+    seqan3::test::tmp_directory tmp{};
+    auto filename = tmp.path() / "sam_file_onput_constructor.sam";
```
eaasna, line 480: same here

eaasna (commented):
Nice work! :+1:

test/unit/io/sam_file/sam_file_input_test.cpp
SGSSGene: oh no, yes that is not correct. The name doesn't matter that much, but I will correct it.

test/snippet/core/cereal_example.cpp
SGSSGene: How did I mess that up :/

eaasna (approved):

test/include/seqan3/test/cereal.hpp
```diff
@@ -47,17 +47,18 @@ template <cereal_input_archive in_archive_t, cereal_output_archive out_archive_t
 //!\endcond
 void do_cerealisation(value_t && value)
 {
-    tmp_filename filename{"cereal_test"};
+    tmp_directory tmp{};
+    auto filename = tmp.path() / "cereal_test";
```
smehringer, line 51: Ok I looked into the issue. Correct me if I am wrong:

* you copied over the beaviour from tmp_filename to tmp_directory (in some other PR)
* now you changed every occurence of tmp_filename to use tmp_directory

Even though in this case I marked, it is clear that one only wants a tmp_filename.

so:
* What about the change in the issue about the tmp_filename? That it only get's a default constructor?
* Would it be better to use it then in appropriate places right away instead of changing everying to tmp_directory and then in some places back to tmp_filename?

smehringer (changes requested):

test/include/seqan3/test/cereal.hpp
SGSSGene: I don't understand the your first point `What about the change in the issue about the tmp_filename? That it only get's a default constructor?`

About the second part "Would it be better to use it then in appropriate places right away instead of changing everying to tmp_directory and then in some places back to tmp_filename?"
You are trying to say that we should keep a `tmp_filename` class that only deals with temporary files? Sure, we could do that.

smehringer: Replying from my phone so I don't have the compete overview right now. hope everything looks right.

I am pretty sure we wanted to keep a `tmp_filename()`too. Doesn't it say so in the issue?

```cpp
// I need just one temp file, I don't care about the name:
seqan3::test::tmp_filename file{};
// If I do care about the extension (e.g. I/O)
file.replace_extension(".fa");. // Like Filesystem::path
// File should be deleted on reconstruction

// If need a directory, multiple files or some other specific behaviour:
seqan3::test::tmp_directory dir{};
// Add other dirs and files
// Everything in the directory should be deleted on deconstruction
```
 While writing this I wonder what happens if I do this
```cpp
{
    seqan3::test::tmp_filename file{};
    std::ofstream ofs{file.get_path()}; // constructs a file;
    file.replace_extension(".txt");
} // Tmp file should be deleted on deconstruction but the filename is different from the constructed file. ?? 
```

SGSSGene: > I am pretty sure we wanted to keep a `tmp_filename()`too. Doesn't it say so in the issue?
> 
Ups 😅! I somehow missed that. Will change everything accordingly.

marehr: > I am pretty sure we wanted to keep a `tmp_filename()`too. Doesn't it say so in the issue?

I think the resolution under that issue https://github.com/seqan/product_backlog/issues/281#issuecomment-766807049 says: 

> For purist reasons we only want `seqan3::test::tmp_directory` with the following features: 

That means no `seqan3::test::tmp_filename` any more. The future way of things should be to only offer a temporary directory where you can do whatever you want. It will warn you if you stepped outside that directory (to protect you from deleting `/tmp/XXXXXX_random_dir/../../`), if you forgot to delete a temporary file, and other safety features. But it will clean that directory no matter what after the destruction of `seqan3::test::tmp_filename`.

We discussed that a `seqan3::test::tmp_filename` makes no sense as you normally want a deterministic file name + extension, but `tmp_filename` implies that you would create a random file name in tmp, i.e. `/tmp/XXXXXXX_random_file_name`. Even if you can specify an extension, it won't be as flexible as the approach of getting the random-ness from a random tmp_directory, i.e. `/tmp/XXXXX_random_folder/` and putting a file in it with a `deterministic_file_name_with.extension`.

Thus, we decided against the `seqan3::test::tmp_filename` altogether in favour for the two-step approach `seqan3::test::tmp_directory`, by creating a random tmp directory first and doing whatever you want in there.

SGSSGene:
Super seeded by #3062

PR #2754, opened 2021-06-22
[MISC] Increase coverage in argument_parser/validator

simonsasse:
Part of: https://github.com/seqan/product_backlog/issues/222 

I was not able to close one coverage gap. I think some code blocks will almost never be executed or only if the OS seriously messes something up.

Here is the Codecov report I used: https://codecov.io/gh/seqan/seqan3/src/master/include/seqan3/argument_parser/validators.hpp

### DIFF
--- a/test/unit/argument_parser/format_parse_validators_test.cpp
+++ b/test/unit/argument_parser/format_parse_validators_test.cpp
@@ -477,6 +477,18 @@ TEST(validator_test, output_directory)
         EXPECT_EQ(path, dir_out_path.string());
     }
 
+    { // create output directory in next level, when parent directory exists.
+        // Parent path exists and is writable
+        seqan3::test::tmp_filename tmp_child_name{"dir/child_dir"};
+        std::filesystem::path tmp_child_dir{tmp_child_name.get_path()};
+        std::filesystem::path tmp_parent_path{tmp_child_dir.parent_path()};
+        
+        std::filesystem::create_directory(tmp_parent_path);
+
+        EXPECT_TRUE(std::filesystem::exists(tmp_parent_path));
+        EXPECT_NO_THROW(seqan3::output_directory_validator{}(tmp_child_dir));
+    }
+
     {
         // get help page message
         std::filesystem::path path;
@@ -525,6 +537,24 @@ TEST(validator_test, inputfile_not_readable)
                                  std::filesystem::perms::owner_read | std::filesystem::perms::group_read |
                                  std::filesystem::perms::others_read,
                                  std::filesystem::perm_options::add);
+}
+
+TEST(validator_test, inputfile_not_regular)
+{
+    seqan3::test::tmp_filename tmp{"my_file.test"};
+    // std::string filename     = "my_file.test";
+    std::filesystem::path filename = tmp.get_path();
+    mkfifo(filename.c_str(), 0644); 
+
+    EXPECT_THROW(seqan3::input_file_validator{}(filename), seqan3::validation_error);
+}
+
+TEST(validator_test, inputdir_not_existing)
+{
+    seqan3::test::tmp_filename tmp_name{"dir"};
+    std::filesystem::path not_existing_dir{tmp_name.get_path()};
+
+    EXPECT_THROW(seqan3::input_directory_validator{}(not_existing_dir), seqan3::validation_error);
 }
 
 TEST(validator_test, inputdir_not_readable)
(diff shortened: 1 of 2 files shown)


SGSSGene:
The fix looks good. Here some inspiration to break some other code lines (Oh yeah! breaking stuff):

```
#include <seqan3/test/tmp_directory.hpp>
#include <sys/stat.h>
....
TEST(validator_test, inputfile_not_regular)
{
    seqan3::test::tmp_directory tmp{};
    auto filename     = tmp.path() / "my_file.test";
    mkfifo(filename.c_str(), 0644);

    EXPECT_THROW(seqan3::input_file_validator{}(filename), seqan3::validation_error);
    tmp.clean();
}
```
This will break https://github.com/simonsasse/seqan3/blob/db7d4d6c612b18c4ffdc3661af6adf8567e3c9c3/include/seqan3/argument_parser/validators.hpp#L419-L421
Since a "fifo" is not a regular file. (The whole code needs some "#if" guards to only be applied under linux system (or use `mknod` to be posix conform).

eseiler:
> The fix looks good. Here some inspiration to break some other code lines (Oh yeah! breaking stuff):
> 
> ```
> #include <seqan3/test/tmp_directory.hpp>
> #include <sys/stat.h>
> ....
> TEST(validator_test, inputfile_not_regular)
> {
>     seqan3::test::tmp_directory tmp{};
>     auto filename     = tmp.path() / "my_file.test";
>     mkfifo(filename.c_str(), 0644);
> 
>     EXPECT_THROW(seqan3::input_file_validator{}(filename), seqan3::validation_error);
>     tmp.clean();
> }
> ```
> 
> This will break https://github.com/simonsasse/seqan3/blob/db7d4d6c612b18c4ffdc3661af6adf8567e3c9c3/include/seqan3/argument_parser/validators.hpp#L419-L421
> Since a "fifo" is not a regular file. (The whole code needs some "#if" guards to only be applied under linux system).

Can't you just create a directory? As far as I've read, a directory is not a regular file because it is classified as ... directory.

simonsasse:
I tried the symlink but it was somehow recognized as a regular file. I couldn't find out why. I'll continue working on this PR later :)

SGSSGene:
> > The fix looks good. Here some inspiration to break some other code lines (Oh yeah! breaking stuff):
> > ```
> > #include <seqan3/test/tmp_directory.hpp>
> > #include <sys/stat.h>
> > ....
> > TEST(validator_test, inputfile_not_regular)
> > {
> >     seqan3::test::tmp_directory tmp{};
> >     auto filename     = tmp.path() / "my_file.test";
> >     mkfifo(filename.c_str(), 0644);
> > 
> >     EXPECT_THROW(seqan3::input_file_validator{}(filename), seqan3::validation_error);
> >     tmp.clean();
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This will break https://github.com/simonsasse/seqan3/blob/db7d4d6c612b18c4ffdc3661af6adf8567e3c9c3/include/seqan3/argument_parser/validators.hpp#L419-L421
> > Since a "fifo" is not a regular file. (The whole code needs some "#if" guards to only be applied under linux system).
> 
> Can't you just create a directory? As far as I've read, a directory is not a regular file because it is classified as ... directory.
> Edit: Ah, this is already caught before :D
> 
> Edit Edit: How about a symlink then? At least this is in the std

Symlink doesn't work. Symlinks are recognized as regular files. This is also good. Since we want to be able to open files that are symlinks. But to be honest, I am now really confused, lets hope std::filesystem is not broken.
https://en.wikipedia.org/wiki/Unix_file_types

eseiler:
> Symlink doesn't work. Symlinks are recognized as regular files. This is also good. Since we want to be able to open files that are symlinks. But to be honest, I am now really confused, lets hope std::filesystem is not broken.
> https://en.wikipedia.org/wiki/Unix_file_types

Ah, that makes sense.
We just call [std::filesystem::is_regular_file](https://en.cppreference.com/w/cpp/filesystem/directory_entry/is_regular_file), which in turn calls [std::filesystem::status](https://en.cppreference.com/w/cpp/filesystem/status).
However, `status` will follow the symlink, hence we get the stats of the target (which is a regular file).

You could use `symlink_status` instead to get the symlink's stats, which would then show symlink instead of regular; however, we want to actually follow the symlink.

But that's the reason symlinks don't work for the test.
fifo seems like the easiest alternative, with some `#if !defined(_WIN32)` I guess.

eseiler:
Also, if you have some `throw`s or branches that really cannot be checked, because they are some contingency plan for an OS failure, you can exclude them from coverage:

```cpp
if (something)
{
    // LCOV_EXCL_START
    throw ...;
    // LCOV_EXCL_STOP
}
```
or
```cpp
if (something)
{
    throw ...; // LCOV_EXCL_LINE
}
```

Note that the `if` is always checked, we just never go inside the `if`.
If it's not obvious why something cannot be checked, add a comment.

SGSSGene (approved):

SGSSGene:
By the power vested in me, I hereby award you the :100: test coverage medal :medal_sports: !

test/unit/argument_parser/format_parse_validators_test.cpp
```diff
@@ -477,6 +477,18 @@ TEST(validator_test, output_directory)
         EXPECT_EQ(path, dir_out_path.string());
     }
 
+    { // create output directory in next level, when parent directory exists.
+        // Parent path exists and is writable
```
eseiler, line 481:
```suggestion
    { // Parent path exists and is writable
```

test/unit/argument_parser/format_parse_validators_test.cpp
```diff
@@ -527,6 +539,24 @@ TEST(validator_test, inputfile_not_readable)
                                  std::filesystem::perm_options::add);
 }
 
+TEST(validator_test, inputfile_not_regular)
+{
+    seqan3::test::tmp_filename tmp{"my_file.test"};
+    // std::string filename     = "my_file.test";
```
eseiler, line 545:
```suggestion
```

eseiler (changes requested):
Only small changes. Did you run the tests with root rights to make sure they work independent of user rights?

eseiler (approved):

PR #2756, opened 2021-07-16
[FIX] Constructing large fm_indices

SGSSGene:
Fixes unsigned vs signed int bug. This causes to limit the fm_index to
work for any text sizes above 2^31-1 base pairs.
This fix now allows up to 2^32-2 which is required for the human genome.
Correction: it should work with 2^63-2 base pairs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,15 @@ The following API changes should be documented as such:
 If possible, provide tooling that performs the changes, e.g. a shell-script.
 -->
 
+# 3.1.0
+
+## Notable Bug-fixes
+
+#### Search
+
+* Resolved an issue that prevented the FM-Index from being constructed correctly for inputs larger than 4 GiB
+  ([\#2756](https://github.com/seqan/seqan3/pull/2756)).
+
 # 3.0.3
 
 Note that 3.1.0 will be the first API stable release and interfaces in this release might still change.
--- a/include/seqan3/search/fm_index/fm_index.hpp
+++ b/include/seqan3/search/fm_index/fm_index.hpp
@@ -293,7 +293,7 @@ private:
         size_t const number_of_texts{text_sizes.size()};
 
         // text size including delimiters
-        size_t const text_size = std::accumulate(text_sizes.begin(), text_sizes.end(), 0) + number_of_texts;
+        size_t const text_size = std::accumulate(text_sizes.begin(), text_sizes.end(), number_of_texts);
 
         if (number_of_texts == text_size)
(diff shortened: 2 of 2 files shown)


Irallia (approved):
LGFM, thanks for the explanation :)

marehr (approved):
Hmm, unfortunately we can't write `std::accumulate<size_t>` to make it more clear as the first argument is the iterator (see definition https://en.cppreference.com/w/cpp/algorithm/accumulate). This seems to be fixed by the standard with a default value version in `std::reduce`.

This is a terrible and hard to see error.

What I'm wondering is that you wrote 

> This fix now allows up to 2^32-2 which is required for the human genome.

But the initial value is `size_t`, so shouldn't the complete 64bit work for the size?

marehr:
Can we add a CHANGELOG entry?

marehr (changes requested):
Changelog is missing; otherwise lgtm

tloka:
> What I'm wondering is that you wrote
> 
> > This fix now allows up to 2^32-2 which is required for the human genome.
> 
> But the initial value is `size_t`, so shouldn't the complete 64bit work for the size?

What is the answer to this? I would probably be close to the 2^32 limit soon, and it wouldn't allow to have forward + reverse sequences in one index for the human genome (which might be useful for some applications). Would be great to have the full 2^64 available.

eseiler:
> > What I'm wondering is that you wrote
> > > This fix now allows up to 2^32-2 which is required for the human genome.
> > 
> > 
> > But the initial value is `size_t`, so shouldn't the complete 64bit work for the size?
> 
> What is the answer to this? I would probably be close to the 2^32 limit soon, and it wouldn't allow to have forward + reverse sequences in one index for the human genome (which might be useful for some applications). Would be great to have the full 2^64 available.

I'd say it's a typo. It should say `2^64`.

tloka:
> I'd say it's a typo. It should say `2^64`.

I can confirm. I just successfully built an index of ~4.8GB with a local fix which wouldn't work for 2^32 (didn't use it for alignment yet though, but at least building finished without error).

SGSSGene:
> > I'd say it's a typo. It should say `2^64`.
> 
> I can confirm. I just successfully built an index of ~4.8GB with a local fix which wouldn't work for 2^32 (didn't use it for alignment yet though, but at least building finished without error).

Thanks for trying it out!. I have only seen the usage of `divsufsort` and read the wrong docu. The SDSL actually is distributed with its own version of divsufsort which handles larger integers.

marehr (approved):

PR #2762, opened 2021-08-11
[FIX,DOC] Escape #

eseiler:
See https://cdash.seqan.de/testDetails.php?test=51999369&build=176475


marehr (approved):

simonsasse (approved):
looks good :) Sorry for the delay!

PR #2769, opened 2021-08-23
[MISC] Adjust default options when outputing FastA-Files

SGSSGene:
Most programs output fasta files in this format:
```
>Test1
AACCGGTT
```
By default `seqan3` puts a whitespace in between '>' and 'Test1'. This default is being changed with this PR to meet the standard of most other programs out there.
seqan3 default:
```
> Test1
AACCGGTT
```

PR #2770, opened 2021-08-23
[FEAT] Option to ignore blanks before ids when reading FastA-Files

SGSSGene:
A simple roundtrip through seqan3 (reading and writing a fasta file) should not introduce any changes with the default options
```
auto fout = seqan3::sequence_file_input{std::istringstream{input}, seqan3::format_fasta{}} |
                  seqan3::sequence_file_output{std::ostringstream{}, seqan3::format_fasta{}};
```

The PR https://github.com/seqan/seqan3/pull/2769 fixes the forced introduction of spaces before the sequence ids.
This PR fixes the removal of whitespaces when reading a fasta file.
This changes the default behavior. To achieve the old behavior a new flag is introduced:
```fin.options.fasta_ignore_blank_before_id``` which is by default `true`.

Example:
File
```
> seq1
ACTG
```
seqan3 now: `id="seq1"`
some others: `id=" seq1"`

### DIFF
--- a/test/unit/io/sequence_file/sequence_file_integration_test.cpp
+++ b/test/unit/io/sequence_file/sequence_file_integration_test.cpp
@@ -47,6 +47,33 @@ TEST(rows, assign_sequence_files)
     EXPECT_EQ(reinterpret_cast<std::ostringstream&>(fout.get_stream()).str(), output_comp);
 }
 
+TEST(rows, assign_sequence_files_read_blanks)
+{
+    std::string const input
+    {
+        ">TEST 1\n"
+        "ACGT\n"
+        "> Test2\n"
+        "AGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGN\n"
+        ">  Test3\n"
+        "GGAGTATAATATATATATATATAT\n"
+    };
+
+    std::string const expected_output = input;
+
+    seqan3::sequence_file_input fin{std::istringstream{input}, seqan3::format_fasta{}};
+    fin.options.fasta_ignore_blanks_before_id = false;
+
+    seqan3::sequence_file_output fout{std::ostringstream{}, seqan3::format_fasta{}};
+    fout.options.fasta_letters_per_line = 0;
+
+    fout = fin;
+
+    fout.get_stream().flush();
+    std::string const output = reinterpret_cast<std::ostringstream&>(fout.get_stream()).str();
+    EXPECT_EQ(output, expected_output);
+}
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: test/unit/io/sequence_file/sequence_file_integration_test.cpp [hunk 1]
```diff
@@ -51,21 +52,29 @@ TEST(integration, assign_sequence_file_pipes)
...
+        ">Test3\n"
         "GGAGTATAATATATATATATATAT\n"
     };
 
     // valid without assignment?
     seqan3::sequence_file_input{std::istringstream{input}, seqan3::format_fasta{}} |
         seqan3::sequence_file_output{std::ostringstream{}, seqan3::format_fasta{}};
 
+    //!TODO this can be changed back to the following lines
+    // when the default of output will change (no blanks before the id
```
MitraDarja, line 68: A closing bracket is missing

MitraDarja (approved):
LGTM, just one minor thing. :)

eseiler:
#2769 is merged, so this can be rebased?

SGSSGene:
> > #2769 is merged, so this can be rebased? (This PR was marked blocked by #2769)
> 
> Its blocked by #2869 

No, this is not blocked by #2869, this can be independently be merged :-)

eseiler (changes requested):
I think we should merge #2869 first, I want to have another look at the predicates afterwards :)

eseiler (approved):
Didn't find an elegant way to avoid code duplication

test/unit/io/sequence_file/sequence_file_integration_test.cpp
```diff
@@ -47,6 +47,40 @@ TEST(rows, assign_sequence_files)
...
+    };
+
+    std::string const output_comp
+    {
+        ">TEST 1\n"
+        "ACGT\n"
+        "> Test2\n"
+        "AGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGNAGGCTGN\n"
+        ">  Test3\n"
+        "GGAGTATAATATATATATATATAT\n"
```
smehringer, line 69: wouldn't it be even more understandable if you use this as input AND output string. So one does not have to compare `input` and `output_comp` when looking at it but directly notices the "perfect roundtrip".

Or do you explicitly want to show that spaces in the sequence are still removed?

smehringer (changes requested):
Although there is no change in behavior, I would add a short changelog entry for this new feature

test/unit/io/sequence_file/sequence_file_integration_test.cpp
SGSSGene: Good point, will adjust. :+1:

smehringer (approved):

PR #2771, opened 2021-08-23
[TEST] repeat_test: use std::views::take

marehr:


test/unit/utility/views/repeat_test.cpp
```diff
@@ -163,9 +162,18 @@ TEST(view, factory)
...
-        EXPECT_EQ(std::ranges::size(v), 3u);
+        auto v = seqan3::views::repeat(str) | std::views::take(3);
+
+        using it_t = std::ranges::iterator_t<decltype(v)>;
+
+        // Note: std::views::take doesn't handle infinite std::ranges::random_access_range's that are not
+        //       std::ranges::sized_range, so we can't use std::ranges::size() to compute the size.
+        // See https://eel.is/c++draft/range.take.view
+        it_t it = v.begin();
+        it_t sentinel = std::ranges::next(v.begin(), v.end()); // unfortunately linear in time
```
SGSSGene, line 173: I can't find any documentation on std::ranges::next:
https://en.cppreference.com/w/cpp/iterator/ranges/next
That only seems to be valid for c++20?

SGSSGene (changes requested):
In general this change seems to drop the support of `seqan3::detail::take_exactly(3)`...
is this really what we want?

test/unit/utility/views/repeat_test.cpp
marehr: Yeah, we "import" those in std-c++17 mode in `include/seqan3/std/iterator` from the ranges-v3 library

> I can't find any documentation on std::ranges::next:
> https://en.cppreference.com/w/cpp/iterator/ranges/next

Yes, it uses the (3.) overload.

```
template< std::input_or_output_iterator I, std::sentinel_for<I> S >
constexpr I next( I i, S bound );
```

> In general this change seems to drop the support of `seqan3::detail::take_exactly(3)`...
> is this really what we want?

Yes, in the long-run we should not provide an own `std::views::take` implementation. In this case the `seqan3::views::repeat` test does not need a dependency on the take exactly view, I removed it.

SGSSGene: Thanks for the explanation!

SGSSGene (approved):

eseiler (approved):

PR #2775, opened 2021-08-23
 [FIX] seqan3::views::single_pass_input::iterator::operator++(int) must return void (real cxx20 std::input_iterator)

marehr:


MitraDarja (approved):

eseiler (approved):
I added a commit to change the documentation (and resolved some conflict with renaming)

PR #2779, opened 2021-08-23
[FIX] clang+gcc9: error: no matching member function for call to 'read_field'

marehr:
```
/seqan3/include/seqan3/io/sam_file/format_bam.hpp:420:5: error: no matching member function for call to 'read_field'
    read_field(stream_view | detail::take_exactly_or_throw(core.l_read_name - 1), id); // field::id
    ^~~~~~~~~~
```

The problem is that clang can't inherit constrained functions, see https://bugs.llvm.org/show_bug.cgi?id=51544.

The fix is to give all `read_field` functions unique names, this arguably makes it also easier to understand which "parsing" algorithm is being called.

### DIFF
--- a/include/seqan3/io/sam_file/detail/format_sam_base.hpp
+++ b/include/seqan3/io/sam_file/detail/format_sam_base.hpp
@@ -86,20 +86,14 @@ protected:
 
     void transfer_soft_clipping_to(std::vector<cigar> const & cigar_vector, int32_t & sc_begin, int32_t & sc_end) const;
 
-    template <typename stream_view_type>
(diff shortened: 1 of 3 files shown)


simonsasse (approved):
looks good :)

eseiler (approved):
Apparently we also never used the `std::optional` overloads?

marehr:
We used it in a single case. But, I could simply replace it. 

Am 24. August 2021 14:36:12 MESZ schrieb Enrico Seiler ***@***.***>:
***@***.*** approved this pull request.

PR #2780, opened 2021-08-23
[FIX] clang+gcc9: seqan3/std/memory: use _GLIBCXX_RELEASE for stdlib version

marehr:
This fixes two issues, one that clang had and another one that our `std::to_address` shim implementation didn't work on gcc 7 - 9.

```
/seqan3/include/seqan3/std/memory:47:15: error: redefinition of 'to_address'
constexpr T * to_address(T * p) noexcept
              ^
```


MitraDarja:
@marehr Can you explain the test file to me? It looks like we are only testing things from the std-library?

eseiler:
> @marehr Can you explain the test file to me? It looks like we are only testing things from the std-library?

We provide an `std::to_address` implementation for gcc<10, but it was faulty.
But it did not work for clang (first commit), and gcc<10 (second commit).

So the tests now basically just check that it works for gcc<10 (and are kinda not required for gcc10/11 CI, but run anyways).

MitraDarja (approved):

eseiler (approved):

PR #2781, opened 2021-08-23
[FIX] __cpp_lib_type_identity should be determined by the standard library version and not the compiler version

marehr:
__cpp_lib_type_identity should be determined by the standard library version and not the compiler version, since clang could use the stdlib from gcc.


SGSSGene (approved):
This is very subtle! nice catch!

eseiler (approved):

PR #2782, opened 2021-08-23
[FIX] clang+gcc9: seqan3/std/bit: use _GLIBCXX_RELEASE for stdlib

marehr:
```
/seqan3/include/seqan3/std/bit:160:33: error: call to 'countl_zero' is ambiguous
    return detail::bits_of<T> - countl_zero(x);
                                ^~~~~~~~~~~
```


Irallia (approved):
LGFM

eseiler (approved):

PR #2784, opened 2021-08-23
[FIX] clang+gcc9: aligned_allocator: clang does not support __cpp_sized_deallocation by default

marehr:
```
/seqan3/include/seqan3/utility/container/aligned_allocator.hpp:180:13: error: no matching function for call to 'operator delete'
            ::operator delete(p, bytes_to_deallocate);
            ^~~~~~~~~~
```

https://github.com/pybind/pybind11/issues/1604#issuecomment-459796241 shows how to enable this.

### DIFF
--- a/include/seqan3/utility/container/aligned_allocator.hpp
+++ b/include/seqan3/utility/container/aligned_allocator.hpp
@@ -18,6 +18,11 @@
 
 #include <seqan3/core/platform.hpp>
 
+// __cpp_aligned_new is a C++17 feature that we use in this allocator and we require it.
+#if __cpp_aligned_new < 201606
+#    pragma GCC warning "Non-C++17 compliant compiler! Please open an issue with your compiler and platform!"
+#endif // __cpp_aligned_new < 201606
+
 namespace seqan3
 {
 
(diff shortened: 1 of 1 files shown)


simonsasse (approved):
lgtm 👍

include/seqan3/utility/container/aligned_allocator.hpp
```diff
@@ -18,6 +18,11 @@
 
 #include <seqan3/core/platform.hpp>
 
+// __cpp_aligned_new is a C++17 feature that we use in this allocator and we require it.
+#if __cpp_aligned_new < 201606
+#    pragma GCC warning "Non-C++17 compliant compiler! Please open an issue with your compiler and platform!"
+#endif // __cpp_aligned_new < 201606
```
eseiler, line 24: Wouldn't this be something that goes into `platform.hpp`? Though I guess it would also prohibit you from using the rest of the library.

include/seqan3/utility/container/aligned_allocator.hpp
```diff
@@ -176,10 +181,24 @@ class aligned_allocator
     void deallocate(pointer const p, size_type const n) const noexcept
     {
         size_t bytes_to_deallocate = n * sizeof(value_type);
+
+        // Clang doesn't have __cpp_sized_deallocation defined by default even though this is a C++14! feature
+        // > In Clang 3.7 and later, sized deallocation is only enabled if the user passes the `-fsized-deallocation`
+        // > flag.
+        // see also https://clang.llvm.org/cxx_status.html#n3778
```
eseiler, line 188: Isn't this something we could configure?

But it also says the following, so I'm not sure.
> The user must supply definitions of the sized deallocation functions, either by providing them explicitly or by using a C++ standard library that does. libstdc++ added these functions in version 5.0, and libc++ added them in version 3.7.

eseiler (changes requested):
2 questions

include/seqan3/utility/container/aligned_allocator.hpp
marehr: Yeah, this should always be true, but on the other side I don't want this to be an annoyance for people who don't use this header in the case that it gives a warning. I don't have any strong feelings about this.

include/seqan3/utility/container/aligned_allocator.hpp
marehr: No we can't within the source code itself. We could add this to our cmake files, but we would when dictate our users to use that flag. And any other build system would not know about this. I think the safest bet is to just use it if it is enabled, because the one who enables it, must make sure that the provided stdlib (in clang it might the one from gcc or clang) provides those functions, and it is not our liability.

eseiler (approved):

PR #2792, opened 2021-08-23
[FIX] clang+gcc9: error: cannot initialize return object of type

marehr:
```
seqan3/include/seqan3/alignment/matrix/detail/trace_iterator_base.hpp:139:16: error: cannot initialize return object of type 'seqan3::detail::trace_iterator_base::pointer' (aka 'seqan3::detail::trace_directions *') with an rvalue of type 'const seqan3::detail::trace_directions *'
        return &current_direction;
               ^~~~~~~~~~~~~~~~~~
```

Part of https://github.com/seqan/product_backlog/issues/127


Irallia (approved):
LGFM

PR #2804, opened 2021-09-06
[TEST] iterator_test_template: should handle contiguous iterator with sentinel

marehr:


Irallia (approved):
LGFM

test/unit/range/iterator_test_template.hpp
```diff
@@ -685,35 +685,43 @@ inline void address_difference_test(it_begin_t it_begin, it_sentinel_t it_end)
 
 TYPED_TEST_P(iterator_fixture, address_of_end)
 {
-    if constexpr (std::derived_from<typename TestFixture::iterator_tag, std::contiguous_iterator_tag>)
+    // sentinel must not be the same as the iterator
+    if constexpr (std::ranges::common_range<decltype(this->test_range)>)
```
eseiler, line 689: Why is that?
Also, code and comment are contradictory.

test/unit/range/iterator_test_template.hpp
```diff
@@ -685,35 +685,43 @@ inline void address_difference_test(it_begin_t it_begin, it_sentinel_t it_end)
...
 }
 
 TYPED_TEST_P(iterator_fixture, address_difference)
 {
     if constexpr (std::derived_from<typename TestFixture::iterator_tag, std::contiguous_iterator_tag>)
     {
-        address_difference_test(std::ranges::begin(this->test_range), std::ranges::end(this->test_range));
+        auto it = std::ranges::begin(this->test_range);
+        auto sentinel_it = std::ranges::next(it, std::ranges::end(this->test_range));
+        address_difference_test(it, sentinel_it);
```
eseiler, line 716: Shouldn't both the old and new version work?

eseiler (changes requested):

test/unit/range/iterator_test_template.hpp
marehr: I have to investigate again :)

marehr: Maybe I should add the reason: The only requirement for contiguous range is the following https://eel.is/c++draft/range.refinements#3

```
🔗
template<class T>
  concept contiguous_­range =
    random_­access_­range<T> && contiguous_­iterator<iterator_t<T>> &&
    requires(T& t) {
      { ranges::data(t) } -> same_­as<add_pointer_t<range_reference_t<T>>>;
    };
```

and

```
Given an expression t such that decltype((t)) is T&, T models contiguous_­range only if to_­address(​ranges​::​begin(t)) == ranges​::​data(t) is true.
```

that means it only required on the iterator (= std::ranges::begin) and not the sentinel.

marehr: This expression here transforms any sentinel type into an iterator type.

test/unit/range/iterator_test_template.hpp
eseiler:
```suggestion
    // `end - begin` must be a valid expression
    using rng_t = decltype(this->test_range);
    if constexpr (std::sentinel_for<std::ranges::sentinel_t<rng_t>, std::ranges::iterator_t<rng_t>>)
```

test/unit/range/iterator_test_template.hpp
```diff
@@ -685,35 +685,43 @@ inline void address_difference_test(it_begin_t it_begin, it_sentinel_t it_end)
...
-    if constexpr (std::derived_from<typename TestFixture::iterator_tag, std::contiguous_iterator_tag>)
+    // sentinel must not be the same as the iterator
+    if constexpr (std::ranges::common_range<decltype(this->test_range)>)
     {
-        using difference_t = std::iter_difference_t<decltype(std::ranges::begin(this->test_range))>;
-        auto it_begin = std::ranges::begin(this->test_range);
-        auto it_end = std::ranges::end(this->test_range);
-        EXPECT_EQ(std::to_address(it_end), std::to_address(it_begin) + static_cast<difference_t>(it_end - it_begin));
-    }
+        if constexpr (std::derived_from<typename TestFixture::iterator_tag, std::contiguous_iterator_tag>)
```
eseiler, line 691: The next `if constexpr` uses the same expression and additionally requires `const_iterable`. Can we simplify this such that we do not check for the `std::contiguous_iterator_tag` twice?

test/unit/range/iterator_test_template.hpp
eseiler: Makes sense :) 
Maybe add a small comment such as 

`The standard only requires the iterator, but not the sentinel, to be contiguous.`

test/unit/range/iterator_test_template.hpp
eseiler: Same for `address_difference`.

I think refactoring would be more clear, otherwise you start to wonder why it's the same expression (i.e., "Can't I read?", "Is this an error?")

test/unit/range/iterator_test_template.hpp
marehr: We talked about this, and it is a property of needing it to be common. We changed the test function declaration to reflect that fact.

eseiler (approved):

PR #2814, opened 2021-09-21
Allow including certain cmake files more than once

marehr:
The main problem with 

```cmake
find_package(SeqAn3)
find_package(SeqAn3)
```

and

```cmake
include (seqan3_require_test.cmake)
include (seqan3_require_test.cmake)
```

is that cmake targets would be re-defined.

This can be fixed by adding an explicit check if the target already exists.


eaasna (approved):
LGTM!

eseiler (approved):

PR #2817, opened 2021-09-21
cmake: Use CMAKE_CURRENT_{BINARY,SOURCE}_DIR

marehr:
This change allows using `add_subdirectory` in an umbrella `CMakeLists.txt` to combine multiple tests into one.

(See: https://stackoverflow.com/questions/15662497/difference-between-cmake-current-source-dir-and-cmake-current-list-dir)


Irallia (approved):
LGFM

eseiler (approved):

PR #2820, opened 2021-09-22
constexpr char literal for cigar

### LINKED ISSUE #2819: Constexpr for the CIGAR char literal
In some contexts, the cigar character needs to be used in constexpr context, e.g.
```cpp
    seqan3::cigar::operation op = 'M'_cigar_operation;
    switch (op.to_rank())
    {
        case 'M'_cigar_operation.to_rank():
            ;
    }
```

I think that it is useful to implement this and I'll provide a PR for it.

joergi-w:
Resolves #2819

### DIFF
--- a/include/seqan3/alphabet/cigar/cigar.hpp
+++ b/include/seqan3/alphabet/cigar/cigar.hpp
@@ -235,7 +235,7 @@ inline namespace literals
  *
  * \stableapi{Since version 3.1.}
  */
-inline cigar::operation operator""_cigar_operation(char const c) noexcept
+constexpr cigar::operation operator""_cigar_operation(char const c) noexcept
 {
     return cigar::operation{}.assign_char(c);
 }
--- a/test/unit/alphabet/cigar/cigar_test.cpp
+++ b/test/unit/alphabet/cigar/cigar_test.cpp
@@ -56,3 +56,10 @@ TEST(cigar, assign_string)
     EXPECT_EQ(uint32_t{223}, seqan3::to_rank(get<0>(c1)));
     EXPECT_EQ('M',           get<1>(c1).to_char());
 }
+
+TEST(cigar, constexpr_char_literal)
+{
+    using seqan3::operator""_cigar_operation;
+    constexpr seqan3::cigar::operation op = 'D'_cigar_operation;
+    EXPECT_EQ(op.to_rank(), 1u);
+}


MitraDarja (approved):

test/unit/alphabet/cigar/cigar_test.cpp
```diff
@@ -56,3 +56,16 @@ TEST(cigar, assign_string)
     EXPECT_EQ(uint32_t{223}, seqan3::to_rank(get<0>(c1)));
     EXPECT_EQ('M',           get<1>(c1).to_char());
 }
+
+TEST(cigar, constexpr_char_literal)
+{
+    using seqan3::operator""_cigar_operation;
+    seqan3::cigar::operation op = 'M'_cigar_operation;
```
eseiler, line 63:
```suggestion
    constexpr seqan3::cigar::operation op = 'M'_cigar_operation;
```

eseiler:
@marehr Should go into release branch?

test/unit/alphabet/cigar/cigar_test.cpp
marehr: I would change from 'M' as it is rank == 0; to 'D' which has rank 1.

```suggestion
    constexpr seqan3::cigar::operation op = 'D'_cigar_operation;
```

test/unit/alphabet/cigar/cigar_test.cpp
```diff
@@ -56,3 +56,16 @@ TEST(cigar, assign_string)
...
+    switch (op.to_rank())
+    {
+        case 'M'_cigar_operation.to_rank():
+            break;
+        default:
+            ;
+    }
```
marehr, line 70: After the change from @eseiler the whole switch statement doesn't add anything. 

```suggestion
    EXPECT_EQ(op.to_rank(), 1u);
```

marehr:
@eseiler Yes, seems like a good idea. Change is small enough :)

include/seqan3/alphabet/cigar/cigar.hpp
```diff
@@ -235,7 +235,7 @@ inline namespace literals
  *
  * \stableapi{Since version 3.1.}
  */
-inline cigar::operation operator""_cigar_operation(char const c) noexcept
+inline constexpr cigar::operation operator""_cigar_operation(char const c) noexcept
```
marehr, line 238:
```suggestion
constexpr cigar::operation operator""_cigar_operation(char const c) noexcept
```

marehr:
@joergi-w Can you rebase in such a way that the PR goes against `upstream/release-3.1.0` ?

joergi-w:
Thank you for the helpful review comments! The switch statement in the test was my use case to detect the issue, but your proposal is surely shorter to test `constexpr`-ness. 

I have rebased on release-3.1.0 and checked that all the other alphabets have constexpr char literal as well.

marehr (approved):

PR #2822, opened 2021-09-23
[INFRA] fix gtest_main target for older cmake version (should propagate include)

marehr:
This PR moves out cmake stuff that belongs to the gtest/gbenchmark fall-back solution.

It also enables that directly using 

```cmake
add_library(my_target PUBLIC gtest)
```

adds the include folder of gtest in case of the fallback solution.


SGSSGene (approved):
:+1:

test/seqan3-test.cmake
```diff
@@ -80,7 +72,7 @@ if (NOT TARGET seqan3::test::performance)
         target_compile_options (seqan3_test_performance INTERFACE "-falign-loops=32")
     endif ()
 
-    target_include_directories (seqan3_test_performance INTERFACE "${SEQAN3_BENCHMARK_CLONE_DIR}/include/")
+
```
smehringer, line 75:
```suggestion
```

smehringer (approved):
LGTM also I still have poblems understanding CMAKE code

smehringer:
I'm a tiny bit hesitant to merge, wondering if I forgot anything. But I guess it is good to go?

PR #2823, opened 2021-09-23
[TEST] snippet: don't link in the gtest main

marehr:
This has the effect that if a snippet does not define a `int main() {}` function, it will use the fallback main-function from `snippet_main`.

The problem right now is, that `gtest_main` will be used and thus adds output that is unrelated to the snippet.

E.g. `./utility/type_traits/function_traits_snippet` has no `main`-function and thus will output `gtest_main`

https://github.com/seqan/seqan3/blob/e438f79b761de8e834bb90e5941d21f1b4455a6a/test/snippet/utility/type_traits/function_traits.cpp#L1-L17

```
[==========] Running 0 tests from 0 test suites.
[==========] 0 tests from 0 test suites ran. (0 ms total)
[  PASSED  ] 0 tests.
```

after the change it will not output anything as it uses `snippet_main`:

```
```

### DIFF
--- a/test/snippet/CMakeLists.txt
+++ b/test/snippet/CMakeLists.txt
@@ -13,6 +13,9 @@ include (../cmake/diagnostics/list_unused_snippets.cmake)
 
 option (seqan3_generate_snippets "Whether seqan3 snippets should be generated and overwritten within the source tree." ON)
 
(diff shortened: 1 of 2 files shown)


Irallia (approved):
LGFM, thank you!

eseiler (approved):

PR #2824, opened 2021-09-23
[TEST] snippet: don't call the version checker on snippet

marehr:
The version checker will output something onto `std::cerr` that makes it "hard" to have deterministic output. This is part of https://github.com/seqan/product_backlog/issues/386.

### DIFF
--- a/test/snippet/CMakeLists.txt
+++ b/test/snippet/CMakeLists.txt
@@ -23,7 +23,11 @@ macro (seqan3_snippet test_name_prefix snippet snippet_base_path)
         RUNTIME_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/${snippet_target_path}"
     )
 
-    add_test (NAME "${test_name_prefix}/${snippet_test_name}_snippet" COMMAND ${target})
+    set (snippet_test_target "${test_name_prefix}/${snippet_test_name}_snippet")
+    add_test (NAME "${snippet_test_target}" COMMAND ${target})
+
+    # disable version checker, as it interferes with comparing the snippet output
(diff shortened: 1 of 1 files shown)


Irallia (approved):
LGFM

test/snippet/CMakeLists.txt
```diff
@@ -23,7 +23,11 @@ macro (seqan3_snippet test_name_prefix snippet snippet_base_path)
         RUNTIME_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/${snippet_target_path}"
     )
 
-    add_test (NAME "${test_name_prefix}/${snippet_test_name}_snippet" COMMAND ${target})
+    set (snippet_test_target "${test_name_prefix}/${snippet_test_name}_snippet")
+    add_test (NAME "${snippet_test_target}" COMMAND ${target})
+
+    # disable version checker, as it interferes with comparing the snippet output
+    set_tests_properties ("${snippet_test_target}" PROPERTIES ENVIRONMENT SEQAN3_NO_VERSION_CHECK=0)
```
smehringer, line 30: So this turns of the version check for all snippets correct? What about the argument parser snippets?

marehr: > So this turns of the version check for all snippets correct?

Yes

> What about the argument parser snippets?

They of course, too.

If you wanted to ask whether we have snippets that "show" the behaviour of the version checker, that isn't the case as far as I can see.

smehringer: > If you wanted to ask whether we have snippets that "show" the behaviour of the version checker, that isn't the case as far as I can see.

Yes. And alright then

smehringer (approved):

PR #2831, opened 2021-09-24
[INFRA] CI Updates

eseiler:
See commit messages.

clang install does work:
https://github.com/eseiler/seqan3/runs/3698326784?check_suite_focus=true
It won't compile, of course :)
In the example, it fails because of gtest, but this is fixed on gtest-master.


Irallia (approved):
LGFM, thank you!

marehr (approved):

PR #2834, opened 2021-09-27
[DOC,FIX] Move minimiser_hash to search

eseiler:


SGSSGene (approved):
Shouldn't the real question be: why are minimizers in search? Couldn't minimizers be used somewhere else?

smehringer (approved):

Issue #2764, opened 2021-08-16
[DOC] Need more documentation for local and global alignment


Irallia:
Especially for the semi-global aligment we need a more detailed description of the gap scheme.
It should be noted that leading and trailing gap settings are connected as an OR and not via AND. This should also be noted again in the tutorial.

The different documentation pages should better link to each other, so that one can find an overview page from everywhere (e.g. https://docs.seqan.de/seqan/3-master-user/structseqan3_1_1align__cfg_1_1free__end__gaps__sequence1__leading.html, https://docs.seqan.de/seqan/3-master-user/classseqan3_1_1align__cfg_1_1method__global.html, https://docs.seqan.de/seqan/3-master-user/classseqan3_1_1align__cfg_1_1gap__cost__affine.html, ...)

marehr:
@Irallia We want to have this for the release 3.1. Do you think you can work on it?

PR #2838, opened 2021-09-29
[DOC] Add detailed documentation to method local and method global …

Irallia:
…and some references to affine gap costs.

Resolves https://github.com/seqan/seqan3/issues/2764

Method local & global:

Bildschirmfoto 2021-09-29 um 12 58 37

Affine gap cost:
Bildschirmfoto 2021-09-29 um 13 02 39


include/seqan3/alignment/configuration/align_config_method.hpp
```diff
@@ -28,12 +28,53 @@ namespace seqan3::align_cfg
...
- * there is no default and this configuration must always be provided for the alignment algorithm.
+ * There are several methods for sequence alignment. We distinguish between \ref seqan3::align_cfg::method_local "local"
+ * and \ref seqan3::align_cfg::method_global "global" alignments and the semi-global alignment in between, which is a
+ * variation of the global alignment.
+ *
+ * **Global Alignment**:\verbatim
+--T--CC-C-AGT--TATGT-CAGGGGACACG-A-GCATGCAGA-GAC
+  |  || |  ||  | | | |||    || | | |  | ||||   |
+AATTGCCGCC-GTCGT-T-TTCAG----CA-GTTATG-T-CAGAT--C\endverbatim
+ * Finding the optimal global alignment of two sequences is solved by the Needleman-Wunsch algorithm a “dynamic program”.
```
Irallia, line 39:
```suggestion
 * Finding the optimal global alignment of two sequences is solved by the **Needleman-Wunsch algorithm** a “dynamic program”.
```

include/seqan3/alignment/configuration/align_config_method.hpp
```diff
@@ -28,12 +28,53 @@ namespace seqan3::align_cfg
...
+ * **Local Alignment** (better suited to find conserved segment):
+ * \verbatim
+                  tccCAGTTATGTCAGgggacacgagcatgcagagac
+                     ||||||||||||
+aattgccgccgtcgttttcagCAGTTATGTCAGatc
+\endverbatim
+ * A \ref seqan3::align_cfg::method_local "local" alignment is effectively a global alignment of two partial sequences,
+ * that is, when two genes from different species are similar in short conserved regions and dissimilar in the remaining
+ * regions. A global alignment would not find the local matching because it would try to align the entire sequence.
+ * This is solved by the Smith-Waterman algorithm.
```
Irallia, line 51:
```suggestion
 * This is solved by the **Smith-Waterman algorithm**.
```

eseiler:
Can you rebase on the release branch?

include/seqan3/alignment/configuration/align_config_method.hpp
```diff
@@ -28,12 +29,53 @@ namespace seqan3::align_cfg
...
  *
- * The alignment algorithm can be categorised in different methods. For example, the
- * \ref seqan3::align_cfg::method_local "local" and the
- * \ref seqan3::align_cfg::method_global "global" alignment are two different methods, while the semi-global alignment
- * is a variation of the global alignment. This differentiation makes it possible to define a subset of configurations
- * that can work with a particular method. Since it is not possible to guess what the desired method for a user is,
- * there is no default and this configuration must always be provided for the alignment algorithm.
+ * There are several methods for sequence alignment. We distinguish between \ref seqan3::align_cfg::method_local "local"
+ * and \ref seqan3::align_cfg::method_global "global" alignments and the semi-global alignment in between, which is a
+ * variation of the global alignment.
```
MitraDarja, line 34:
```suggestion
* and \ref seqan3::align_cfg::method_global "global" alignments. The semi-global alignment is implemented as a variation of the global alignment.
```

include/seqan3/alignment/configuration/align_config_method.hpp
```diff
@@ -28,12 +29,53 @@ namespace seqan3::align_cfg
...
+ *
+ * **Global Alignment**:\verbatim
+--T--CC-C-AGT--TATGT-CAGGGGACACG-A-GCATGCAGA-GAC
+  |  || |  ||  | | | |||    || | | |  | ||||   |
+AATTGCCGCC-GTCGT-T-TTCAG----CA-GTTATG-T-CAGAT--C\endverbatim
+ * Finding the optimal global alignment of two sequences is solved by the **Needleman-Wunsch algorithm** a “dynamic
+ * program”. Hence, the algorithm requires O(nm) time and memory.
```
MitraDarja, line 41: Is it important to mention that it is based on dynamic programing? If so, I would suggest:
```suggestion
Finding the optimal global alignment of two sequences is solved by the dynamic programing algorithm **Needleman-Wunsch**. Hence, the algorithm requires O(nm) time and memory.
```

Or we leave the information out completely, Needleman-Wunsch is so well known that you either know that already or can google it in a few seconds. ;)

MitraDarja: Also, we mention the time complexity here and a few sentences below again, I think it is enough to mention it only once.

include/seqan3/alignment/configuration/align_config_method.hpp
```diff
@@ -28,12 +29,53 @@ namespace seqan3::align_cfg
...
+ *
+ * We need to store (n+1)(m+1) numbers, with a constant number of calculations required for each number (three sums and
+ * one maximum). Therefore, the algorithm requires **O(nm)** time and memory.
+ *
+ * To reduce the time complexity you can use a \ref seqan3::align_cfg::band_fixed_size "banded" alignment.
+ *
+ * --
+ *
+ * By distinguishing between these alignments, it is possible to define a subset of
+ * \ref seqan3::align_cfg "configurations" that can work with a particular method. Since it is not possible to guess
```
MitraDarja, line 64: I am not sure, I understand this paragraph completley, what particular methods can we configurate? Global and local or also something else?

include/seqan3/alignment/configuration/align_config_method.hpp
```diff
@@ -28,12 +29,53 @@ namespace seqan3::align_cfg
...
+ * which method a user wants, there is no default and this \ref seqan3::align_cfg "configuration" must always be
+ * specified for the alignment algorithm.
+ *
+ * These include configurations for a semi-global alignment, such as the overlap alignment when you do not wish to
```
MitraDarja, line 68: If it is not too much work, add an example for an semi-global alignment like the ones above?

MitraDarja (changes requested):
LGTM, but some minor things. :)

include/seqan3/alignment/configuration/align_config_method.hpp
Irallia: Bildschirmfoto 2021-09-30 um 10 42 14

MitraDarja: Hm, now I think I would add the semi-global alignment below the global one? But leave the sentence how to generate it with the configs here. Does that make sense?

include/seqan3/alignment/configuration/align_config_method.hpp
MitraDarja: Hasn't changed?

MitraDarja (changes requested):
Sorry, but still some :nail_care: :)

include/seqan3/alignment/configuration/align_config_method.hpp
Irallia: it has, but not in your comment ^^wierd...
from
Bildschirmfoto 2021-09-30 um 11 24 41
to
Bildschirmfoto 2021-09-30 um 11 23 01

include/seqan3/alignment/configuration/align_config_method.hpp
Irallia: Yes I thought about this, aswell. I will try :)

Irallia: How do you like this:
Bildschirmfoto 2021-09-30 um 11 31 29

MitraDarja: Nice! :+1:

MitraDarja (approved):

include/seqan3/alignment/configuration/align_config_method.hpp
```diff
@@ -28,12 +29,60 @@ namespace seqan3::align_cfg
...
  * \details
  *
- * The alignment algorithm can be categorised in different methods. For example, the
- * \ref seqan3::align_cfg::method_local "local" and the
- * \ref seqan3::align_cfg::method_global "global" alignment are two different methods, while the semi-global alignment
- * is a variation of the global alignment. This differentiation makes it possible to define a subset of configurations
- * that can work with a particular method. Since it is not possible to guess what the desired method for a user is,
- * there is no default and this configuration must always be provided for the alignment algorithm.
+ * There are several methods for sequence alignment. We distinguish between \ref seqan3::align_cfg::method_local "local"
+ * and \ref seqan3::align_cfg::method_global "global" alignments. The semi-global alignment is implemented as a
```
smehringer, line 33:
```suggestion
 * and \ref seqan3::align_cfg::method_global "global" alignments. The *semi-global* alignment is implemented as a
```
:nail_care:

include/seqan3/alignment/configuration/align_config_method.hpp
```diff
@@ -28,12 +29,60 @@ namespace seqan3::align_cfg
...
+ * variation of the global alignment.
+ *
+ * **Global Alignment**:\verbatim
+--T--CC-C-AGT--TATGT-CAGGGGACACG-A-GCATGCAGA-GAC
+  |  || |  ||  | | | |||    || | | |  | ||||   |
+AATTGCCGCC-GTCGT-T-TTCAG----CA-GTTATG-T-CAGAT--C\endverbatim
+ * Finding the optimal global alignment of two sequences is solved by the **Needleman-Wunsch algorithm**.
```
smehringer, line 40: maybe a link to a website describing needleman wunsch?

include/seqan3/alignment/configuration/align_config_method.hpp
```diff
@@ -28,12 +29,60 @@ namespace seqan3::align_cfg
...
+ *
+ * **Local Alignment** (better suited to find conserved segment):
```
smehringer, line 42:
```suggestion
 * **Local Alignment** (better suited to find conserved segments):
```
Now that I see this. Maybe you can add right at the top when you use which alignment method? You currently just state "we differentiate between..." but I think for the user skimming the first few sentences it might be nice to directly read something like  `Global alignments are e.g. typically used when aligning sequence reads to a reference genome. Local alignments are better suited to find conserved segments/regions, e.g. in motiv-finding applications.` If there is something like motiv finding applications.. but isn't MARS one?

include/seqan3/alignment/configuration/align_config_method.hpp
```diff
@@ -28,12 +29,60 @@ namespace seqan3::align_cfg
...
+ * \verbatim
+                  tccCAGTTATGTCAGgggacacgagcatgcagagac
+                     ||||||||||||
+aattgccgccgtcgttttcagCAGTTATGTCAGatc
+\endverbatim
+ * A \ref seqan3::align_cfg::method_local "local" alignment is effectively a global alignment of two partial sequences,
+ * that is, when two genes from different species are similar in short conserved regions and dissimilar in the remaining
```
smehringer, line 49:
```suggestion
 * A \ref seqan3::align_cfg::method_local "local" alignment is effectively a global alignment of two partial sequences.
 * For example when two genes from different species are similar in short conserved regions and dissimilar in the remaining
```
Or is `that is` the same as `for example`? That is.. sounds to me as this is the only case.

include/seqan3/alignment/configuration/align_config_method.hpp
```diff
@@ -28,12 +29,60 @@ namespace seqan3::align_cfg
...
+ * regions. A global alignment would not find the local matching because it would try to align the entire sequence.
+ * This is solved by the **Smith-Waterman algorithm**.
+ *
+ * **Semi-global Alignment** (e.g. overlapping sequences):
+ * \verbatim
+                  TCCCAGTTATGTCAGgggacacgagcatgcagagac
+                  |||||||||||||||
+aattgccgccgtcgttttTCCCAGTTATGTCAG
+\endverbatim
+ * The semi-global alignment is a specially configured global alignment.
```
smehringer, line 59:
```suggestion
 * The semi-global alignment is a specially configured global alignment, namely we do not penalize gaps at the ends of the alignment. Semi-global alignments are often used in genome assembly applications when trying to find matching overlaps.
```
I think :D

include/seqan3/alignment/configuration/align_config_method.hpp
```diff
@@ -28,12 +29,60 @@ namespace seqan3::align_cfg
...
+ *
+ * **Complexity** of both algorithms:
+ *
+ * We need to store (n+1)(m+1) numbers, with a constant number of calculations required for each number (three sums and
+ * one maximum). Therefore, the algorithm requires **O(nm)** time and memory.
```
smehringer, line 64:
```suggestion
 * Since both algorithms are based on dynamic programming, they run in quadratic time and memory **O(nm)** (where  
 * `n` and `m` are the lengths of the respective aligned sequences). 
 ```
 I don't think it is especially helpful to know that we compute three sums and one maximum in the recursion of the DP, is it?

File: include/seqan3/alignment/configuration/align_config_method.hpp [hunk 13]
```diff
@@ -28,12 +29,60 @@ namespace seqan3::align_cfg
...
+ *
+ * To reduce the time complexity you can use a \ref seqan3::align_cfg::band_fixed_size "banded" alignment.
```
smehringer, line 66:
```suggestion
 * To reduce the time complexity you can use a \ref seqan3::align_cfg::band_fixed_size "banded" alignment. It reduces
 * the runtime by a constant although remaining quadratic and limiting the possible solutions slightly.
 *
 * You can speed up the computation significantly if you parallelize (todo reference option parallel) and simdify
 * (reference simd option) your alignment.
```
Is simd possible yet, via public configuration? @marehr

smehringer (changes requested):
Firstly, this helps a lot I think!

I am not done with the very last paragraph because I noticed something now. This documentation now appears only at the local alignment configuration object. But it described all the methods in general. I don't know if that's the perfect spot then?

I think this should either go into the `Pairwise` landing page or the `align_pairwise` documentation. Do you want to chat about it? Because I think there is also the nice way of putting some of your documentation at the respective configuration objects and then copy it to the landing page so it appears at both locations.

include/seqan3/alignment/configuration/align_config_method.hpp
marehr: We have a vectorise config and you need to add some compiler options to enable machine optimisation like `-march=native` or `-mavx2`.

marehr:
> Because I think there is also the nice way of putting some of your documentation at the respective configuration objects and then copy it to the landing page so it appears at both locations.

Just a side note, doxygen supports including markdown fragments for text passages. I would prefer that over copying from entities, as we don't know how they might change.

See https://github.com/seqan/seqan3/tree/master/doc/fragments for some examples. Especially the `cigar_operation_table.md` fragment.

include/seqan3/alignment/configuration/align_config_method.hpp
Irallia: should I link to seqan3::detail::simd_affine_gap_policy or what do you mean?

Irallia: What would be the right link? I get a warning with the one above..
```bash
warning: unable to resolve reference to 'seqan3::simd' for \ref command
```

Irallia:
I have now already included the direct text change requests, the rest we will discuss then.
Bildschirmfoto 2021-10-01 um 12 13 39

include/seqan3/alignment/configuration/align_config_method.hpp
marehr: seqan3::simd is dev-only documented.

include/seqan3/alignment/configuration/align_config_method.hpp
```diff
@@ -28,12 +29,68 @@ namespace seqan3::align_cfg
...
+ * the alignment. Semi-global alignments are often used in genome assembly applications when trying to find matching
+ * overlaps.
+ *
+ * **Complexity** of both algorithms:
+ *
+ * Since both algorithms are based on dynamic programming, they run in quadratic time and memory **O(nm)** (where `n`
+ * and `m` are the lengths of the respective aligned sequences).
+ *
+ * To reduce the time complexity you can use a \ref seqan3::align_cfg::band_fixed_size "banded" alignment. It reduces
+ * the runtime by a constant although remaining quadratic and limiting the possible solutions slightly.
```
marehr, line 71:
```suggestion
 * To reduce the time complexity you can use a \ref seqan3::align_cfg::band_fixed_size "banded" alignment. A fixed band of size `k` can reduce the runtime to **O(kn)**, but be aware that the quality of the sequence alignments will be worse than an optimal sequence alignment. You should use this option when you know that the two sequences are mostly similar.
```

I think we should mention that a band is orthogonal to the alignment method and e.g. works in combination with global, semi-global and local alignment.

Furthermore, we should mention that depending on the configuration that you can decrease the runtime down to `O(nm / w)` where `w` is typical 64. For the edit distance (myers bitvector, https://docs.seqan.de/seqan/3-master-user/group__alignment__configuration.html#ga3070f8ad7b5692e9c798b411109500a3).

I'm wondering if we should direct to https://docs.seqan.de/seqan/3-master-user/group__alignment__pairwise.html as that already has information about the runtime / space complexity :)

We should change that page to give the space / runtime in terms of two sequences: `O(nm)`

smehringer:
> Maybe we should add finer grained sub-modules structure for the configuration to group them more together.

what do you mean exactly? Because the (sub)modules in the documentation refer to the file structure, no?

Irallia:
See the changed pages:
Method global:
https://seqan3-o4x25wiy3-seqan.vercel.app/usr/html/classseqan3_1_1align__cfg_1_1method__global.html#details
Method local:
https://seqan3-o4x25wiy3-seqan.vercel.app/usr/html/classseqan3_1_1align__cfg_1_1method__local.html#details

Alignment page, which will go later to the Pairwise page:
https://seqan3-o4x25wiy3-seqan.vercel.app/usr/html/group__alignment.html#details

eseiler (commented):
Can you try to do it like 
https://github.com/seqan/seqan3/blob/master/doc/fragments/cigar_operation_table.md
and
https://github.com/seqan/seqan3/blob/34f3a7f672b6228d5ae1036d2ff590b244cb098e/include/seqan3/alphabet/cigar/all.hpp#L35

test/snippet/alignment/configuration/align_cfg_method_global.cpp
```diff
@@ -0,0 +1,20 @@
...
+#include <seqan3/alignment/pairwise/align_pairwise.hpp>
+#include <seqan3/alignment/scoring/nucleotide_scoring_scheme.hpp>
+#include <seqan3/alphabet/nucleotide/dna4.hpp>
+#include <seqan3/core/debug_stream.hpp>
+
+using namespace seqan3::literals;
+
+int main()
+{
+    auto min_cfg = seqan3::align_cfg::method_global{} |
```
smehringer, line 12:
```suggestion
    // configure a global alignment for DNA sequences
    auto min_cfg = seqan3::align_cfg::method_global{} |
```

test/snippet/alignment/configuration/align_cfg_method_global.cpp
```diff
@@ -0,0 +1,20 @@
...
+                   seqan3::align_cfg::scoring_scheme{seqan3::nucleotide_scoring_scheme{seqan3::match_score{4},
+                                                                                       seqan3::mismatch_score{-5}}};
+
+    auto seq1 = "TCGT"_dna4;
+    auto seq2 = "ACGA"_dna4;
+    for (auto res : align_pairwise(std::tie(seq1, seq2), min_cfg))
```
smehringer, line 18:
```suggestion
    for (auto res : seqan3::align_pairwise(std::tie(seq1, seq2), min_cfg))
```
funny that adl works..

test/snippet/alignment/configuration/align_cfg_method_global.cpp
```diff
@@ -0,0 +1,20 @@
...
+        seqan3::debug_stream << res.score() << '\n';
```
smehringer, line 19:
```suggestion
        seqan3::debug_stream << res.score() << '\n'; // print out the alignment score
```

smehringer (changes requested):
> Can you try to do it like https://github.com/seqan/seqan3/blob/master/doc/fragments/cigar_operation_table.md and
> 
> https://github.com/seqan/seqan3/blob/34f3a7f672b6228d5ae1036d2ff590b244cb098e/include/seqan3/alphabet/cigar/all.hpp#L35

Es @eseiler suggested. This should be done with a markdown fragment instead of pseudo doxygen documentation. I saw in my old seqan3 branch that it was done like you did it. But in the newest master there are only `.md` files in `doc/fragments/`. Did you check if you are up to date?

Otherwise only :nail_care:

Irallia:
> > Can you try to do it like https://github.com/seqan/seqan3/blob/master/doc/fragments/cigar_operation_table.md and
> > https://github.com/seqan/seqan3/blob/34f3a7f672b6228d5ae1036d2ff590b244cb098e/include/seqan3/alphabet/cigar/all.hpp#L35
> 
> Es @eseiler suggested. This should be done with a markdown fragment instead of pseudo doxygen documentation. I saw in my old seqan3 branch that it was done like you did it. But in the newest master there are only `.md` files in `doc/fragments/`. Did you check if you are up to date?
> 
> Otherwise only 💅

I'am up to date, but as this PR merges into release, there are no md files. I just saw the old hpp files.
I'll change my new files to md. :) But the release maybe needs an update.

smehringer:
> > > Can you try to do it like https://github.com/seqan/seqan3/blob/master/doc/fragments/cigar_operation_table.md and
> > > https://github.com/seqan/seqan3/blob/34f3a7f672b6228d5ae1036d2ff590b244cb098e/include/seqan3/alphabet/cigar/all.hpp#L35
> > 
> > 
> > Es @eseiler suggested. This should be done with a markdown fragment instead of pseudo doxygen documentation. I saw in my old seqan3 branch that it was done like you did it. But in the newest master there are only `.md` files in `doc/fragments/`. Did you check if you are up to date?
> > Otherwise only nail_care
> 
> I'am up to date, but as this PR merges into release, there are no md files. I just saw the old hpp files. I'll change my new files to md. :) But the release maybe needs an update.

good to know!

smehringer (approved):

PR #2839, opened 2021-10-01
[DOC] Make api tags more visible.

smehringer:
Let's see if this works on different browsers.


test/documentation/api-stability.js
```diff
@@ -13,12 +13,28 @@ $(document).ready(function()
...
     {
         if (entry_type == "member")
+        {
             $(this).find('span.mlabels').prepend('<span class="mlabel ' + api_type + '">' + api_type + '</span>');
+
+            // get header above the item, append the api_type tag to the header and change the bg color
+            prev = entry_obj;
+            do prev = prev.previousSibling; while(prev && prev.nodeType !== 1);
+
+            prev.innerText = prev.innerText + '  ' + '[' + api_type + '] ';
```
smehringer, line 26:
```suggestion
            prev.innerContent = prev.innerContent + '  ' + '[' + api_type + '] ';
```

test/documentation/api-stability.js
```diff
@@ -13,12 +13,28 @@ $(document).ready(function()
...
-    var add_labels = function(api_type, entry_type)
+    var add_labels = function(api_type, api_color, entry_obj, entry_type)
     {
         if (entry_type == "member")
+        {
             $(this).find('span.mlabels').prepend('<span class="mlabel ' + api_type + '">' + api_type + '</span>');
+
+            // get header above the item, append the api_type tag to the header and change the bg color
+            prev = entry_obj;
+            do prev = prev.previousSibling; while(prev && prev.nodeType !== 1);
```
marehr, line 24: we seem to have jquery. it should contain an `up` function or something that you could use instead to make this more "robust".

test/documentation/api-stability.js
```diff
@@ -13,12 +13,28 @@ $(document).ready(function()
...
+
+            prev.innerContent = prev.innerContent + '  ' + '[' + api_type + '] ';
```
marehr, line 26: if you can you should avoid something like this, as this needs to recreate all node. You should be able to append a new "<span>" element.

something like this:
```
$(prev).append();
```

test/documentation/api-stability.js
```diff
@@ -13,12 +13,28 @@ $(document).ready(function()
...
+
+            if (api_type != "stable-api") // for now, leave stable api background in blue.
+            {
+                prev.style.backgroundImage = "none";
+                prev.style.backgroundColor = api_color;
+            }
```
marehr, line 32: I would prefer if we added a css class here and make the styling in the css file. That would have the result that you don't need to pass in the color.

test/documentation/api-stability.js
smehringer: but I need to append WITHIN the title (<h2>). `append` would append something after the closing `</h2>` element. So how to I get the inner stuff? I found only this. (And `innerHTML` but that was discouraged and `innerContent was recommended instead`)

test/documentation/api-stability.js
smehringer: more work for something so tiny? :(

test/documentation/api-stability.js
smehringer: I'll check again

test/documentation/api-stability.js
```diff
@@ -13,12 +13,28 @@ $(document).ready(function()
...
+        }
         else if (entry_type == "header")
```
eseiler, line 34:
```suggestion

            // get header above the item, append the api_type tag to the header and change the bg color
            var header = $(entry_obj).prev('h2.memtitle')[0];
            header.append(' [' + api_type + '] ');

            if (api_type != "stable-api") // for now, leave stable api background in blue.
            {
                header.style.backgroundImage = "none";
                header.style.backgroundColor = api_color;
            }
        }
        else if (entry_type == "header")
```

test/documentation/api-stability.js
smehringer: Thanks @eseiler

eseiler (approved):

test/documentation/api-stability.js
```diff
@@ -13,12 +13,21 @@ $(document).ready(function()
         $(this).find('div.memproto').html(new_html);
     }
 
-    var add_labels = function(api_type, entry_type)
+    var add_labels = function(api_type, entry_obj, entry_type)
```
marehr, line 16: @eseiler Last question, can we get rid of `entry_obj` and use `this`?

eseiler: I'll check

smehringer:
It works for me too :)

PR #2844, opened 2021-10-06
[FIX] replace seqan3::views::join_with with a for-loop

marehr:
Part of https://github.com/seqan/product_backlog/issues/402 and related to https://github.com/ericniebler/range-v3/pull/1665


marehr:
@Irallia CI passes now :)

PR #2869, opened 2021-10-27
[FIX]  correct parsing of IDs in fasta files

SGSSGene:
Fasta has the following structure:
```
> MyID
MySeq
```

This should be parsed into 'MyID' and 'MySeq' (This works fine).

If the sequence id has '>' as a character, this might fail:

```
> >MyID
MySeq
```
This will be parsed into 'MyId' and 'MySeq', which does not meet the
specification. It should be parsed to '>MyID' and 'MySeq'.

This PR fixes this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,11 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 
 ## Notable Bug-fixes
 
+#### IO
+
+* Fixed an issue parsing FASTA files containing sequence IDs that start with `>`
+ ([\#2869](https://github.com/seqan/seqan3/pull/2869)).
+
 #### Utility
 
 * `seqan3::views::single_pass_input` cannot propagate the `std::ranges::output_range` property, because it cannot
--- a/test/performance/io/format_fasta_no_performance_benchmark.cpp
+++ b/test/performance/io/format_fasta_no_performance_benchmark.cpp
@@ -1,0 +1,13 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2021, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2021, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/* This file includes the same performance tests as format_fasta_benchmark.cpp
+ * but sets the `SEQAN3_WORKAROUND_VIEW_PERFORMANCE` variable first.
+ * This assures that different definitions of functions are used.
+ */
+#define SEQAN3_WORKAROUND_VIEW_PERFORMANCE 0
+#include "format_fasta_benchmark.cpp"
--- a/test/unit/io/sequence_file/sequence_file_integration_no_performance_test.cpp
+++ b/test/unit/io/sequence_file/sequence_file_integration_no_performance_test.cpp
@@ -1,0 +1,14 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2021, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2021, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/* This file includes the same unit tests as sequence_file_integration_test.cpp
+ * but sets the `SEQAN3_WORKAROUND_VIEW_PERFORMANCE` variable first.
+ * This assures that different definitions of functions are used.
+ */
+
+#define SEQAN3_WORKAROUND_VIEW_PERFORMANCE 0
+#include "sequence_file_integration_test.cpp"
--- a/test/unit/io/sequence_file/sequence_file_integration_test.cpp
+++ b/test/unit/io/sequence_file/sequence_file_integration_test.cpp
@@ -135,3 +135,28 @@ TEST(integration, convert_fastq_to_fasta)
     fout.get_stream().flush();
     EXPECT_EQ(reinterpret_cast<std::ostringstream&>(fout.get_stream()).str(), fasta_out);
 }
+
+TEST(integration, sequence_name_contains_id_tag)
+{
+    // The sequence id is '>TEST 1'.
+    std::string const input
+    {
+        "> >TEST 1\n"
+        "ACGT\n"
+    };
+
+    std::string const expected_output
+    {
+        ">>TEST 1\n"
+        "ACGT\n"
+    };
+
+    seqan3::sequence_file_input fin{std::istringstream{input}, seqan3::format_fasta{}};
+    seqan3::sequence_file_output fout{std::ostringstream{}, seqan3::format_fasta{}};
+    fout = fin;
+
+    fout.get_stream().flush();
+
+    std::string const output = static_cast<std::ostringstream&>(fout.get_stream()).str();
+    EXPECT_EQ(output, expected_output);
+}
--- a/include/seqan3/io/sequence_file/format_fasta.hpp
+++ b/include/seqan3/io/sequence_file/format_fasta.hpp
@@ -194,7 +194,9 @@ private:
             #if SEQAN3_WORKAROUND_VIEW_PERFORMANCE
                 auto it = stream_view.begin();
                 auto e = stream_view.end();
-                for (; (it != e) && (is_id || is_blank)(*it); ++it)
+                ++it; // already checked `is_id`
+
+                for (; (it != e) && (is_blank)(*it); ++it)
                 {}
 
                 bool at_delimiter = false;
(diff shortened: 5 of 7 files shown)


test/unit/io/sequence_file/sequence_file_integration_test.cpp
```diff
@@ -135,3 +135,25 @@ TEST(integration, convert_fastq_to_fasta)
...
+    {
+        ">>>TEST 1\n"
+        "ACGT\n"
+    };
+
+    std::string const output_comp
+    {
+        "> >>TEST 1\n"
+        "ACGT\n"
+    };
```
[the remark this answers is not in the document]
SGSSGene: Yes, this is supposed to look wierd.  :-)
All Documentation I can find on how fasta work states that a sequence ID starts with an '>' and then the actual sequence ID follows. Which means `>>MyID` would be the sequence ID `>MyID`. Only the first '>' should be dropped.
Our implementation also has a second twist in it, (which is also not standard) which makes it even more confusing.
We don't only drop the first '>' but also all following spaces. And when writing we add a space at the beginning. (We are going to change this in another PR, to not write an extra space)
Meaning `>   MyID` would be read as the ID `MyID`.  and `YourID` would be writen out as `> YourID`.
This has the wierd twist that `>>>TEST` will be read as `>>TEST` and be writen out as `> >>TEST`.

Yes, it is a confusing big mess xD. It is a good Idea to add another test.

smehringer:
memo: change your base branch to master

test/unit/io/sequence_file/sequence_file_integration_test.cpp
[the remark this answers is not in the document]
SGSSGene: I have changed the unit test and added a comment. Can you check if this clears things up?

SGSSGene:
> memo: change your base branch to master

Thank you, changed!

Irallia (approved):

Irallia:
Maybe add a changelog entry for this :)

Irallia (approved):

smehringer:
> Maybe add a changelog entry for this :)

yes please

CHANGELOG.md
```diff
@@ -28,6 +28,11 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 
 ## Notable Bug-fixes
 
+#### IO
+
+* Fixed an edge case on parsing FASTA files with sequence ids that start with '>' symbols.
+ [\#2869](https://github.com/seqan/seqan3/pull/2869).
```
eseiler, line 34:
```suggestion
* Fixed an issue parsing FASTA files containing sequence IDs that start with `>`
  ([\#2869](https://github.com/seqan/seqan3/pull/2869)).
```

test/unit/io/sequence_file/sequence_file_integration_test.cpp
```diff
@@ -135,3 +135,27 @@ TEST(integration, convert_fastq_to_fasta)
...
+        "> >TEST 1\n"
+        "ACGT\n"
+    };
+
+    std::string const output_comp
+    {
+        "> >TEST 1\n"
+        "ACGT\n"
+    };
+
```
eseiler, line 154: I think:

* rename: `output_comp` -> `output`

And either:

* Write `std::string const output = input;`, so one doesn't start looking for the difference between the two strings :)
* Just re-use `input`.

I prefer the former, because it's explicit.

eseiler (changes requested):
Can you double-check that the tests work with both `SEQAN3_WORKAROUND_VIEW_PERFORMANCE=0` and `SEQAN3_WORKAROUND_VIEW_PERFORMANCE=1`?

I think we set it to `1`, so you would need to pass `=0` via CMake? Or just change it in the `platform.hpp` for testing purpose.

SGSSGene:
> Can you double-check that the tests work with both `SEQAN3_WORKAROUND_VIEW_PERFORMANCE=0` and `SEQAN3_WORKAROUND_VIEW_PERFORMANCE=1`?
> 
> I think we set it to `1`, so you would need to pass `=0` via CMake? Or just change it in the `platform.hpp` for testing purpose.

Yes, turns out this is more than required! My code was actually not working.
I added another test, to cover the "SEQAN3_WORKAROUND_VIEW_PERFORMANCE=0" case.
It uses a forbidden `#include "*.cpp"` statement. It seemed very practical to avoid code duplication.
Do you have any better ideas?
See: test/unit/io/sequence_file/sequence_file_integration_no_performance_test.cpp

CHANGELOG.md
```diff
@@ -34,6 +34,11 @@ If possible, provide tooling that performs the changes, e.g. a shell-script.
 
 ## Notable Bug-fixes
 
+#### IO
+
+* Fixed an issue parsing FASTA files containing sequence IDs that start with `>` symbols.
+ ([\#2869](https://github.com/seqan/seqan3/pull/2869)).
```
eseiler, line 40:
```suggestion
* Fixed an issue parsing FASTA files containing sequence IDs that start with `>`
 ([\#2869](https://github.com/seqan/seqan3/pull/2869)).
```

test/unit/io/sequence_file/sequence_file_integration_no_performance_test.cpp
```diff
@@ -0,0 +1,14 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2021, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2021, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/* This file includes the same unit tests as sequence_file_integration_test.cpp
+ * but sets the `SEQAN3_WORKAROUND_VIEW_PERFORMANCE` variable first.
+ * This assures that a different definitions of functions are being used
```
eseiler, line 10:
```suggestion
 * This assures that a different definitions of functions are being used.
```

test/unit/io/sequence_file/sequence_file_integration_test.cpp
```diff
@@ -135,3 +135,30 @@ TEST(integration, convert_fastq_to_fasta)
     fout.get_stream().flush();
     EXPECT_EQ(reinterpret_cast<std::ostringstream&>(fout.get_stream()).str(), fasta_out);
 }
+
+TEST(integration, tricky_id)
```
eseiler, line 139:
```suggestion
TEST(integration, sequence_name_contains_id_tag)
```

maybe?

test/unit/io/sequence_file/sequence_file_integration_test.cpp
```diff
@@ -135,3 +135,30 @@ TEST(integration, convert_fastq_to_fasta)
...
+{
+    // The read/written part of the id should start with '> ' for indicating a sequence.
+    // The sequence id is '>TEST 1'.
+    std::string const input
+    {
+        "> >TEST 1\n"
+        "ACGT\n"
+    };
```
eseiler, line 147:
```suggestion
    // The sequence id is '>TEST 1'.
    std::string const input
    {
        "> >TEST 1\n"
        "ACGT\n"
    };
```

Could be removed if test name is changed, otherwise:
```
The sequence name should start with `>`, which usually marks a line as containing the ID.
```

test/unit/io/sequence_file/sequence_file_integration_test.cpp
```diff
@@ -135,3 +135,30 @@ TEST(integration, convert_fastq_to_fasta)
...
+
+    std::string const expected_output
+    {
+        ">>TEST 1\n"
+        "ACGT\n"
+    };
+
```
eseiler, line 154:
```suggestion
```

eseiler (changes requested):

test/performance/io/format_fasta_no_performance_benchmark.cpp
```diff
@@ -0,0 +1,13 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2021, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2021, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/* This file includes the same performance tests as format_fasta_benchmark.cpp
+ * but sets the `SEQAN3_WORKAROUND_VIEW_PERFORMANCE` variable first.
+ * This assures that a different definitions of functions are used.
```
eseiler, line 10:
```suggestion
 * This assures that different definitions of functions are used.
```

test/unit/io/sequence_file/sequence_file_integration_no_performance_test.cpp
```diff
@@ -0,0 +1,14 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2021, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2021, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
+
+/* This file includes the same unit tests as sequence_file_integration_test.cpp
+ * but sets the `SEQAN3_WORKAROUND_VIEW_PERFORMANCE` variable first.
+ * This assures that a different definitions of functions are used.
```
eseiler, line 10:
```suggestion
 * This assures that different definitions of functions are used.
```

eseiler (approved):

PR #2870, opened 2021-10-27
[FIX] the in/out parameter order of convert_through_char_representation

SGSSGene:
This fixes the in/out parameter order. First all input parameters, than all output parameters.


smehringer (changes requested):
Please rebase on master :)

Otherwise LGTM

smehringer (approved):

PR #2877, opened 2021-11-08
[DOC] Update the remote "html-book.tar.xz" to the newest version.

SGSSGene:
Updates the remote "html-book.tar.xz" to the newest version.

(It seems like these download haven't been updated for over a year. so the update is still from the year 2020).


smehringer (commented):
LGTM. What about the manual links to C++20 entities that @eseiler mentioned when we talked about this?

eseiler (approved):

smehringer (approved):

PR #2892, opened 2021-11-15
[INFRA] CI update

eseiler:
Update of googlebenchmark to `1.6.0` fixes some missing includes for gcc11

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ cache:
 addons:
   apt:
     sources:
-      - ubuntu-toolchain-r-test
+      - sourceline: 'ppa:ubuntu-toolchain-r/test'
     packages:
       - g++-11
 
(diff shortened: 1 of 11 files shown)


.travis.yml
```diff
@@ -51,27 +51,30 @@ install:
...
+      header) make -j2 gtest_build gbenchmark_build;;
+    esac
 
 script:
   - export SEQAN3_NO_VERSION_CHECK=1
   - export CCACHE_COMPRESS=true
   - export CCACHE_COMPRESSLEVEL=6
   - export CCACHE_MAXSIZE=500M
-  - make -k -j2
+  - make -k -j2 || make -k
```
eseiler, line 72: The runners with *exotic* architecture have less memory available (around 3 GB). Usually, we can build everything using 2 threads, but if we are unfortunate, something gets killed. In this case, we just build the killed targets with 1 thread.

SGSSGene: Could you make this comment inside the yaml file?

.travis.yml
```diff
@@ -51,27 +51,30 @@ install:
...
   - |
     if [[ "${BUILD}" =~ ^(snippet)$ ]]; then
       ctest . --output-on-failure
     else
-      ctest . -j2 --output-on-failure
+      ctest . -j2 --output-on-failure || ctest . --rerun-failed --output-on-failure
```
SGSSGene, line 77: I assume, the same applies here?

SGSSGene (changes requested):
I like the much better formatting!
much easier to read.

.travis.yml
eseiler: I put it in line 67

SGSSGene: Very nice! thx!

SGSSGene (approved):
Looks Good!

smehringer (approved):
Nice!

PR #2894, opened 2021-11-17
[INFRA] Add diagnostics for GCC 7 and 8

eseiler:
We might wanna think about doing this for all known broken compilers (or at least warn if compiler is not GCC).


smehringer:
> We might wanna think about doing this for all known broken compilers (or at least warn if compiler is not GCC).

I agree

eseiler:
> > We might wanna think about doing this for all known broken compilers (or at least warn if compiler is not GCC).
> 
> I agree

Ok, this'll be a follow-up. We will have to check if all compilers understand `#pragma GCC warning` which we use to emit warnings instead of fatal errors.

SGSSGene (approved):

include/seqan3/core/platform.hpp
```diff
@@ -22,6 +22,14 @@
 #define SEQAN3_STR(x) SEQAN3_STR_HELPER(x)
 //!\endcond
 
+// ============================================================================
+//  Compiler support
+// ============================================================================
+
+#if defined(__GNUC__) && (__GNUC__ == 7 || __GNUC__ == 8)
```
smehringer, line 29: Maybe make it `< 9` ? So the warning also is emitted for acc < 7. 
Or do you want to have a different warning for < 7?

smehringer (approved):

PR #2895, opened 2021-11-17
[MISC] Deprecate std/filesystem

eseiler:


Irallia (approved):

include/seqan3/std/filesystem
```diff
@@ -15,113 +15,7 @@
...
-{
-
-namespace filesystem = std::experimental::filesystem;
-
-} // namespace std
-
-#endif // __has_include(experimental/filesystem)
-//!\endcond
+SEQAN3_DEPRECATED_HEADER(
+    "This header is deprecated and will be removed in SeqAn-3.2.0; Please #include <filesystem> instead.")
```
eseiler, line 21: Actually:

```suggestion
    "This header is deprecated and will be removed in SeqAn-3.3.0; Please #include <filesystem> instead.")
```

no?
Since we only deprecate in 3.2.0

smehringer: Yes. IMO will be removed in 3.3

smehringer (approved):
Just alter the deprecation warning to 3.3 and you can merge

PR #2898, opened 2021-11-22
[feature] views::char_strictly_to

h-2:


include/seqan3/alphabet/views/char_strictly_to.hpp
```diff
@@ -0,0 +1,79 @@
...
+ * \tparam alphabet_t   The alphabet to convert to; must satisfy seqan3::alphabet.
+ * \param[in] urange    The range being processed. [parameter is omitted in pipe notation]
+ * \returns             A range of converted elements. See below for the properties of the returned range.
+ * \ingroup views
+ *
+ * \details
+ *
+ * \header_file{seqan3/alphabet/views/char_strictly_to.hpp}
+ *
+ * This views differs from seqan3::views::chars_to in that it throws an exception if an invalid character conversion
```
Irallia, line 40:
```suggestion
 * This views differ from seqan3::views::chars_to in that they throw an exception if an invalid character conversion
```

include/seqan3/alphabet/views/char_strictly_to.hpp
```diff
@@ -0,0 +1,79 @@
...
+ * happens. See seqan3::char_strictly_to for more details.
+ *
+ * ### View properties
+ *
+ * This view is a **deep view** Given a range-of-range as input (as opposed to just a range), it will apply
```
Irallia, line 45:
```suggestion
 * This view is a **deep view**. Given a range-of-range as input (as opposed to just a range), it will apply
```

include/seqan3/alphabet/views/char_strictly_to.hpp
```diff
@@ -0,0 +1,79 @@
...
+ * | std::ranges::viewable_range      | *required*                            | *guaranteed*                                       |
+ * | std::ranges::view                |                                       | *guaranteed*                                       |
+ * | std::ranges::sized_range         |                                       | *preserved*                                        |
+ * | std::ranges::common_range        |                                       | *preserved*                                        |
+ * | std::ranges::output_range        |                                       | *lost*                                             |
+ * | seqan3::const_iterable_range     |                                       | *preserved*                                        |
+ * |                                  |                                       |                                                    |
+ * | std::ranges::range_reference_t   | seqan3::alphabet_char_t<alphabet_t>   | `alphabet_t`                                       |
```
Irallia, line 63:
```suggestion
 * | Concepts and traits              | `urng_t` (underlying range type)      | `rrng_t` (returned range type) |
 * |----------------------------------|:-------------------------------------:|:------------------------------:|
 * | std::ranges::input_range         | *required*                            | *preserved*                    |
 * | std::ranges::forward_range       |                                       | *preserved*                    |
 * | std::ranges::bidirectional_range |                                       | *preserved*                    |
 * | std::ranges::random_access_range |                                       | *preserved*                    |
 * | std::ranges::contiguous_range    |                                       | *lost*                         |
 * |                                  |                                       |                                |
 * | std::ranges::viewable_range      | *required*                            | *guaranteed*                   |
 * | std::ranges::view                |                                       | *guaranteed*                   |
 * | std::ranges::sized_range         |                                       | *preserved*                    |
 * | std::ranges::common_range        |                                       | *preserved*                    |
 * | std::ranges::output_range        |                                       | *lost*                         |
 * | seqan3::const_iterable_range     |                                       | *preserved*                    |
 * |                                  |                                       |                                |
 * | std::ranges::range_reference_t   | seqan3::alphabet_char_t<alphabet_t>   | `alphabet_t`                   |
```

Irallia (changes requested):
LGFM, just some spelling and style things.

Irallia (approved):

marehr:
I don't know if this was discussed, but is this REALLY necessary? What's the use case, what's the semantic?

This view mangles input validation with transformation, and it should be a niche use-case for most users. Most users will get their sequences by reading in a file by using our IO. Of course our IO should provide some level of diagnostics if the read in characters are wrong. But should this API really be public in the alphabet module?

If you want to use it in IO, please just make it detail for now.

I would even argue that the current functionality is kind of lacking for a user. It just throws an exception, but the user won't be able to recover from that. They don't know where the offending character is, they don't know how the context around the error looks like, they just know that there was somewhere an invalid character.

From a user POV the same can also be achieved by iterating with `std::ranges::all_of` over the sequence and test if the characters are all valid.

I'm pro for some diagnostic "library" functionality to give a user a good reason why a sequence is invalid, like gcc or clang does for parsing a file, but I don't think a VIEW is the right place for that.

h-2:
> I don't know if this was discussed, but is this REALLY necessary?

Yes, from my point of view it is. I don't want to have page-long arguments on Github for adding effectively 6 lines of code to the library. It has got one approval already, and another team member has been automatically assigned for a second review. If you feel that this change is so fundamental that Enrico is not qualified to approve it and that it needs to be discussed by the entire core-team, please bring it up at the next meeting of the core-team.
In that case, please give me a heads-up, so I can send you a detailed reasoning.

eseiler:
> Yes, from my point of view it is. I don't want to have page-long arguments on Github for adding effectively 6 lines of code to the library. It has got one approval already, and another team member has been automatically assigned for a second review. If you feel that this change is so fundamental that Enrico is not qualified to approve it and that it needs to be discussed by the entire core-team, please bring it up at the next meeting of the core-team. In that case, please give me a heads-up, so I can send you a detailed reasoning.

My first question when I saw the PR was `Do we want this?` - and this answer did not add anything useful to ... well, anything.

I wanted to discuss this PR in last week's core meeting, but there were only two members present, so we cancelled it.

I think it is very valid to question the usefulness: "Oh wow, there is a wrong character somewhere in my 20 GiB file"- which also translates to whether this should be API or detail. What are even the use cases? Apparently you have some, but decided not to share them.
However, this point is not crucial to me. That what's `\experimentalapi` is for.

Since from my current understanding, both `assign_char_strictly_to` and `views::char_strictly_to` are at least experimental, and hence open for debate in the next minor release, it might not hurt to provide some insight from your point of view.

In my opinion, this view should either be `\experimentalapi` with some explicit statement of uncertainty, or just in detail for now.
It's OK for me to add this view as-is (either experimental or detail). However, if you have more to say about the whole `strictly_to` mechanism, I would be delighted to read more than _`No description provided.`_

h-2:
> My first question when I saw the PR was (Why) Do we want this? - and this answer did not add anything useful to ... well, anything.

Why didn't you ask me this question?

> I think it is very valid to question the usefulness: "Oh wow, there is a wrong character somewhere in my 20 GiB file"- which also translates to whether this should be API or detail. 

That is not how I am using it.

> What are even the use cases? Apparently you have some, but decided not to share them.

My IO-Code returns records with views into the stream-buffer. The IO does not verify the characters and instead returns a view to the user that converts to the requested alphabet type. There needs to be a user-friendly mechanism of notifying the user when there is an illegal character in the data (user-friendly means not telling the user to run some extra algorithms on the data). Since this is a type that end-users see, it should not be detail, because it means they don't know what is going. Also, this is not some "hidden implementation", it is a very simple view with a 5-line implementation that works exactly like all the other views.

> Since from my current understanding, both assign_char_strictly_to and views::char_strictly_to are at least experimental, and hence open for debate in the next minor release, it might not hurt to provide some insight from your point of view.

I was not aware that you removed these requirements from the concept. I thought that it was rather self-explanatory that one might want to check which characters "are in" the alphabet?

> In my opinion, this view should either be \experimentalapi with some explicit statement of uncertainty, or just in detail for now.

I write external library and application code, so things in detail, NOAPI or experimental are not very helpful to me. In that case, it is easier to just have things in my own code.

eseiler:
> Why didn't you ask me this question?

I wanted to ask in the Core-Meeting, and afterwards it slipped my mind :)

> That is not how I am using it.

> My IO-Code returns records with views into the stream-buffer. The IO does not verify the characters and instead returns a view to the user that converts to the requested alphabet type. There needs to be a user-friendly mechanism of notifying the user when there is an illegal character in the data (user-friendly means not telling the user to run some extra algorithms on the data). Since this is a type that end-users see, it should not be detail, because it means they don't know what is going. Also, this is not some "hidden implementation", it is a very simple view with a 5-line implementation that works exactly like all the other views.

That was my main problem with the PR. There was no description/reasoning, so I was left to wonder what that view might be used for. And as I suspected, you have a use case where this view is useful.

> I was not aware that you removed these requirements from the concept. I thought that it was rather self-explanatory that one might want to check which characters "are in" the alphabet?

The "use" of the view itself is self-explanatory. What I was missing was a use case that highlights how the view solves some problem. Without any use case, we will fall back to implementing "everything that might seem useful", and this did not work out too well.

> I write external library and application code, so things in detail, NOAPI or experimental are not very helpful to me. In that case, it is easier to just have things in my own code.

The only thing I was missing was some context, which you provided.

In general:
* Not marking an entity means `\noapi` (See [review of docs](https://seqan3-git-fork-h-2-viewscharsstrictlyto-seqan.vercel.app/usr/html/namespaceseqan3_1_1views.html#a440d31b2d6b1cb7c6dcacdc14c1dee5f))
* Everything new is `\experimentalapi{Experimental since version 3.2.}`

In this case, `assign_char_strictly_to` is currently experimental, so the view cannot be stable. But that also means that one could argue that if the CPO is stable, the view instantly follows (i.e. stable with 3.2).
Our main problem with `assign_char_strictly_to` was, IIRC, that we didn't have use cases for this and therefore weren't sure if it really should be stable just now (3.1).
Now that you provide use cases, it's (from my POV) unlikely that `assign_char_strictly_to` changes (will remain as is and will be stable with 3.2).

include/seqan3/alphabet/views/char_strictly_to.hpp
```diff
@@ -0,0 +1,80 @@
+// -----------------------------------------------------------------------------------------------------
+// Copyright (c) 2006-2020, Knut Reinert & Freie Universität Berlin
+// Copyright (c) 2016-2020, Knut Reinert & MPI für molekulare Genetik
+// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
+// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
+// -----------------------------------------------------------------------------------------------------
```
eseiler, line 6:
```suggestion
// -----------------------------------------------------------------------------------------------------
// Copyright (c) 2006-2021, Knut Reinert & Freie Universität Berlin
// Copyright (c) 2016-2021, Knut Reinert & MPI für molekulare Genetik
// This file may be used, modified and/or redistributed under the terms of the 3-clause BSD-License
// shipped with this file and also available at: https://github.com/seqan/seqan3/blob/master/LICENSE.md
// -----------------------------------------------------------------------------------------------------
```

include/seqan3/alphabet/views/char_strictly_to.hpp
```diff
@@ -0,0 +1,80 @@
...
+#include <seqan3/utility/type_traits/basic.hpp>
+#include <seqan3/utility/views/deep.hpp>
+
+namespace seqan3::views
+{
+
+/*!\name Alphabet related views
+ * \{
+ */
+
```
eseiler, line 27:
```suggestion
```

include/seqan3/alphabet/views/char_strictly_to.hpp
```diff
@@ -0,0 +1,80 @@
...
+/*!\brief               A view over an alphabet, given a range of characters.
+ * \tparam urng_t       The type of the range being processed. See below for requirements. [template parameter is
+ *                      omitted in pipe notation]
+ * \tparam alphabet_t   The alphabet to convert to; must satisfy seqan3::alphabet.
+ * \param[in] urange    The range being processed. [parameter is omitted in pipe notation]
+ * \returns             A range of converted elements. See below for the properties of the returned range.
```
eseiler, line 33:
```suggestion
 * \returns             A range of converted elements. See below for the properties of the returned range.
 * \throws              seqan3::invalid_char_assignment if an invalid character is encountered.
```

include/seqan3/alphabet/views/char_strictly_to.hpp
```diff
@@ -0,0 +1,80 @@
...
+ * \ingroup views
```
eseiler, line 34:
```suggestion
 * \ingroup alphabet_views
```

include/seqan3/alphabet/views/char_strictly_to.hpp
```diff
@@ -0,0 +1,80 @@
...
+ * | std::ranges::common_range        |                                       | *preserved*                    |
+ * | std::ranges::output_range        |                                       | *lost*                         |
+ * | std::semiregular                 |                                       | *preserved*                    |
+ * | seqan3::const_iterable_range     |                                       | *preserved*                    |
+ * |                                  |                                       |                                |
+ * | std::ranges::range_reference_t   | seqan3::alphabet_char_t<alphabet_t>   | `alphabet_t`                   |
+ *
+ * See the \link views views submodule documentation \endlink for detailed descriptions of the view properties.
+ *
+ */
```
eseiler, line 68:
```suggestion
 * ### Example
 *
 * \include test/snippet/alphabet/views/char_strictly_to.cpp
 * \hideinitializer
 *
 * \experimentalapi{Experimental since version 3.2.}
 */
```

eseiler (changes requested):
Basically just documentation 💅 

Can you add a snippet, please?

Something like this(untested)?

`test/snippet/alphabet/views/char_strictly_to.cpp`

```cpp
#include <seqan3/alphabet/views/char_strictly_to.hpp>

int main()
{
    std::string str{"ACTTTGATAN"};
    auto v = ;
    try
    {
        seqan3::debug_stream << (str | seqan3::views::char_strictly_to<seqan3::dna4>); // ACTTTGATA
    }
    catch (seqan3::invalid_char_assignment)
    {
       seqan3::debug_stream << "\n[ERROR] Invalid char!\n"; // Will throw on parsing 'N'
    }
}
```

`test/snippet/alphabet/views/char_strictly_to.err`

```text
ACTTTGATA
[ERROR] Invalid char!

```

h-2:
> The only thing I was missing was some context, which you provided.

Yes, I had thought that assign_char_strictly was still in the concept, so I thought missing this view was an oversight. I hope the use-case is now clear. If you have any questions, please let me know.

marehr:
@h-2 would it be also an alternative to have a view that just checks whether a character is part of the "domain", as we already have a view that has the converting effect?

Something like this:

```c++
auto assign_char_strictly = views::char_is_valid | views::assign_char;
```

name would be up-to-discussion (`throw_on_invalid_char`, ...)

h-2:
> would it be also an alternative to have a view that just checks whether a character is part of the "domain", as we already have a view that has the converting effect?

Let me think about it!

h-2:
I wouldn't mind splitting the implementation into two separate views, but I would still like to have the combined view (defined as you proposed above) for usability. Would that be agreeable?

eseiler:
I like both proposals, i.e. having a view that checks for validity, and then having the combined view.

I just noticed that we probably need to restrict the view to not just take `alphabet`.
We need to check for (depending on implementation) either:
* `assign_char_strictly_to`
* `assign_char_to` + `char_is_valid_for`

I think a static_assert would be enough.

`assign_char_strictly_to` looks like this:

https://github.com/seqan/seqan3/blob/67873e4f9db65d051a58f07ca38cc681d6b21bbf/include/seqan3/alphabet/concept.hpp#L729-L750

eseiler:
> > I just noticed that we probably need to restrict the view to not just take alphabet
> 
> Unless someone changed it, `char_is_valid_for` has a works-by-default implementation for all alphabets that don't customise it. So no additional requirements are needed from my POV.

You're right, I missed the default implementation in the CPO.

test/snippet/alphabet/views/valid_char_for.err
```diff
@@ -0,0 +1,2 @@
+ACTTTGATA
```
eseiler, line 1: Can you rename this file to <code>valid<b>ate</b>_char_for.err</code>?

eseiler (changes requested):
One filename is wrong.

Only snippets are failing. 
GitHub seems to have trouble serving the Cmake download today, that's why CI sometimes fails.

eseiler (approved):

PR #2899, opened 2021-11-22
[feature] add list_traits::repeat

h-2:
There might be smarter ways to do this, but this does this job.


h-2:
Same thing here: why is coverage in arg parser reduced?

MitraDarja (approved):

smehringer (approved):

PR #2902, opened 2021-11-23
[INFRA] Update CI

eseiler:
* Cancelling running jobs is now natively supported by GitHub
* Use conda for installing `ccache` (apt: 3.7 (not updated), conda: 4.5 (updated))
* Increase compression level
* Increase cache size (was too low and forced cache clean)
* ccache: ignore `-fprofile-abs-path` (not ignoring = no cache hits)
* print verbose ccache statistics


.github/workflows/documentation.yaml
```diff
@@ -7,6 +7,10 @@ on:
   # Enables a manual trigger.
   workflow_dispatch:
 
+concurrency:
+  group: docs-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: false
```
SGSSGene, line 12: Why not abort documentation in progress?

.github/workflows/ci_macos.yml
```diff
@@ -113,13 +117,13 @@ jobs:
           CCACHE_BASEDIR: ${{ github.workspace }}
           CCACHE_DIR: ${{ github.workspace }}/.ccache
           CCACHE_COMPRESS: true
-          CCACHE_COMPRESSLEVEL: 6
-          CCACHE_MAXSIZE: 50M
+          CCACHE_COMPRESSLEVEL: 12
+          CCACHE_MAXSIZE: 75M
         run: |
-          ccache -p || true
+          ccache -z
```
SGSSGene, line 123: Here you wrote `ccache -z` further down (ci_misc.yml) it is `ccache -z || true`.
1. Why are they different?
2. and wouldn't be better to have `ccache -z` and `ccache -p`?

SGSSGene (changes requested):
I need some insights :-)

.github/workflows/documentation.yaml
eseiler: Because it uses rsync to copy the documentation to the server and I would like it not to interrupt that

.github/workflows/ci_macos.yml
eseiler: In the linux/macos CI, we always install `ccache`.
In the misc CI, we only install it if needed (require_ccache in the job matrix)

I never found `ccache -p` useful in the CI, I found it more useful to reset statistics (`-z`) and then print the verbose statistics at the end (`-sv`).

SGSSGene (approved):
Looks good!
And thank your for the insights on the documentation build and the `prog1 || true` syntax. I wasn't aware that that actually works.

smehringer (approved):

PR #2904, opened 2021-11-26
Missing seqan3::field::cigar.

joshuak94:
The field `seqan3::field::cigar` was missing from the list of default_fields in the SAM snippet.


smehringer:
Reviewing after team review. But looks goot to me :)

SGSSGene (approved):
If I understand correctly, the `default_fields` type didn't match the default type given in 
for `seqan3::sam_file_input` in include/seqan3/io/sam_file/input.hpp:238

In that case, this looks good!

smehringer (approved):

PR #2893, opened 2021-11-16
[FIX] signature of store_sse4

eseiler:
See declaration:

https://github.com/seqan/seqan3/blob/66b59a268f5f1645aec1c8bd76b910e06a5a1b18/include/seqan3/utility/simd/detail/simd_algorithm_sse4.hpp#L38

PR #2906, opened 2021-11-29
[INFRA] Add AVX2 cron

eseiler:
This adds a cron job to compiler our tests with AVX2 enabled.
This allows us to notice things like https://github.com/seqan/seqan3/pull/2893


SGSSGene (approved):
AVX2 to the win....or at least making sure it compiles :-)

smehringer (approved):

